PR #6232, opened 2016-09-21
[spaceship] Add basic support for Developer ID profiles

radex:
This extends Spaceship's provisioning profile support for Mac Developer ID distribution:
- If you have any Developer ID profiles, fetching all profiles won't blow up at `Can't find class '#{attrs['distributionMethod']}`
- Developer ID profiles become `Spaceship::Portal::ProvisioningProfile::Direct` objects
- Creating new Developer ID profiles works
- Updating Developer ID profiles works

### DIFF
--- a/spaceship/lib/spaceship/portal/provisioning_profile.rb
+++ b/spaceship/lib/spaceship/portal/provisioning_profile.rb
@@ -29,6 +29,8 @@ module Spaceship
       #     "adhoc"
       # @example Development Profile
       #     "limited"
+      # @example Mac Developer ID Profile
+      #     "direct"
       attr_accessor :distribution_method
 
       # @return (String) The name of this profile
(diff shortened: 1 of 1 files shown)


mfurtak:
Hi @radex - thanks for putting this together. It is looking great! 👍 

Can you take a look at `provisioning_profile_spec.rb` and see how it best makes sense to extend those tests to account for your changes?

Thanks!

mfurtak (changes requested):
Please update tests to account for the new changes and prevent future regressions.

mfurtak (approved):

mfurtak:
On second thought, we're going to try to get a _fastlane_ release out today, so let's adopt these changes as-is, and I'll supplement the tests 👍

KrauseFx:
Looks great, thanks @radex 🚀

radex:
Awesome — thanks @mfurtak!

liftoff

PR #6257, opened 2016-09-22
Fix Sticker category by adding correct Apps prefix

asfalcone:
- The full category for stickers is MZGenre.Apps.Stickers.
- Fix wording in _deliver_ output
- Add "Stickers" to Reference.md

PR #6104, opened 2016-09-13
Update internal dependencies

KrauseFx:
:rocket:

PR #6262, opened 2016-09-22
Update Deliver version and dependency on spaceship

asfalcone:
- Fix Sticker category by adding correct Apps prefix (#6257)
- Update internal dependencies (#6104)


KrauseFx (approved):
👍

Issue #6279, opened 2016-09-23
Move from sonar-runner to sonar-scanner


giacgbj:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Just run the sonar action :-)

```
[23:23:19]: -------------------
[23:23:19]: --- Step: sonar ---
[23:23:19]: -------------------
[23:23:19]: $ cd /Users/jhack/workspaceIOS/projectName && sonar-runner -Dsonar.projectVersion="1.0.18"
[23:23:19]: ▸ WARN: sonar-runner script is deprecated. Please use sonar-scanner instead.
```

It should be extremely easy to change the script to call.

### Configuration Files

Fastfile:
```ruby
sonar
```

### Environment

fastlane version (run `fastlane -v`): 1.104.0

TKBurner:
@giacgbj Awesome! Thanks for your contribution. We'll work with you on your pull request.

Hope you have a great day!

PR #6280, opened 2016-09-23
#6279: Move from sonar-runner to sonar-scanner

giacgbj:
Related to #6279.


KrauseFx:
Thanks for your contribution, this looks great, however we can't just break the API of the action. Could you revert the changes of the public API? 👍

giacgbj:
I mainly created this pull request to avoid the warning, and I know it can be done without breaking the public API.
However I think that sooner or later the public API should be changed to align everything to the new name in order to avoid confusion.
I'm going to revert those changes.

KrauseFx:
The tests seem to be failing

giacgbj:
Fixed.

KrauseFx (approved):

KrauseFx:
This is great, thanks for your contribution @giacgbj, this will be in the next release 🚀

Issue #6277, opened 2016-09-23
[gym] EXPORT FAILED without using legacy build api on CircleCI


nafu:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I found the gym problem on CircleCI, and show one sample solution without using legacy build api (**Solution B** listed below).
It would be better to revise [gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh](https://github.com/fastlane/fastlane/blob/master/gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh). What do you think?

Here is [sample project](https://github.com/ookamiinc/Xcode8BuildOnCircleCI) and [full build log on CircleCI](https://circleci.com/gh/ookamiinc/Xcode8BuildOnCircleCI/10).

#### Problem

Even if rvm and rbenv commands do not exist, bundler set env variables below.
It causes gym command (ipatool) fails.

```diff
+BUNDLE_ORIG_PATH=/usr/local/bin:/usr/local/lib/ruby/gems/2.3.0/bin:/usr/bin:/bin:/usr/sbin:/sbin
+BUNDLE_BIN_PATH=/usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/exe/bundle
+BUNDLE_GEMFILE=/Users/distiller/Xcode8BuildOnCircleCI/Gemfile
+RUBYOPT=-rbundler/setup
+RUBYLIB=/usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib
```

#### Solution A

Of course, we can use legacy build api to solve this.

#### Solution B

This is one sample solution without using legacy build api.
Unset env variables in **xcbuild-safe.sh**.
https://github.com/nafu/fastlane/commit/3d8563e795718185b63bbf348249b886d9a10115#diff-dd20e2475cd26f8228344467473c9b1fR52

```diff
+unset BUNDLE_BIN_PATH
+unset BUNDLE_GEMFILE
+unset GEM_HOME
+unset GEM_PATH
+unset RUBYLIB
+unset RUBYOPT
```

We need to unset GEM_HOME and GEM_PATH on CircleCI, even if we do not use rbenv.
https://github.com/fastlane/fastlane/issues/6277#issuecomment-253102805

##### Complete output when running fastlane, including the stack trace and command used

[...]

Issue #5240, opened 2016-06-28
xcbuild-safe.sh should support chruby (and others?)


aidansteele:
I noticed that [`gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh`](https://github.com/fastlane/fastlane/blob/master/gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh) explicitly checks for the present of either rvm or rbenv - as you can imagine, this fails in the scenario where a user is using chruby.

chruby needs the same treatment as rbenv, e.g.

```bash
if chruby-exists-or-whatever; then
  echo "chruby detected, removing env variables"
  unset RUBYLIB
  unset RUBYOPT
  unset GEM_HOME
  unset GEM_PATH
fi
```

(My bash-fu isn't strong enough to know how to check for chruby's existence.)

That said, is there any reason we don't just execute `xcodebuild` with these env vars unconditionally unset? Is there any scenario where we'd want them to be set? Surely we don't want this script to grow as the number of Ruby managers grows, but I'm out of my depth here.

KrauseFx:
I don't have any background on this, but the original PR can be found here https://github.com/fastlane/fastlane/pull/4625. I'm more than happy to merge a PR adding chruby support. The reason why this was added in the first place is because of the way Xcode handles its Ruby environment.

dflems:
RVM is the invasive one that needs special logic. `rbenv` and `chruby` just need the ruby environment unset.  Aside from the RVM logic, this should work fine as a catch-all for other ruby managers (at least it's worked for me with chruby).

```bash
unset BUNDLE_BIN_PATH
unset BUNDLE_GEMFILE
unset GEM_HOME
unset GEM_PATH
unset RUBYLIB
unset RUBYOPT
unset RUBY_ENGINE
unset RUBY_ROOT
unset RUBY_VERSION
unset _ORIGINAL_GEM_PATH
```

KrauseFx:
Hey everyone, we just pushed a new _gym_ release with [gym 1.11.1](https://github.com/fastlane-old/gym/releases/tag/1.11.1), that should fix an error in `xcodebuild`. Xcode depends on system Ruby being used, and sometimes can't find the system gems. The new release should fix that error. Please disable the `use_legacy_build_api` flag, update to the latest version of _gym_ and give it a try. More information about the fixed error can be found here: https://github.com/fastlane/fastlane/issues/6495

[...]

PR #6301, opened 2016-09-26
added support for chruby

tcurdt:

Before submitting a pull request, we appreciate if you create an issue first to discuss the change :+1: ...  This is related to  

https://github.com/fastlane/fastlane/issues/5240
https://github.com/fastlane/fastlane/issues/6277

 and various other issues potential "device not found" issues.

Anyway - this is what finally fixed many issues for me. Would be great to get this merged.

(Who would have thought that the `xcodebuild` has a dependency to system ruby?!)

### DIFF
--- a/gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh
+++ b/gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh
@@ -49,6 +49,11 @@ if which rbenv > /dev/null; then
   unset GEM_PATH
 fi
 
+if type -t chruby > /dev/null; then
+  echo "chruby detected, forcing to use system ruby"
+  chruby system
+fi
+
 # to help troubleshooting
 # env | sort > /tmp/env.wrapper
 # rvm info >> /tmp/env.wrapper


KrauseFx:
Thanks for your pull request @tcurdt. I don't have a lot of knowledge around this. @nafu Would that solve the issue for you? (via https://github.com/fastlane/fastlane/issues/6277#issuecomment-249353140)

tcurdt:
It did solve a big issue for me - if that counts :)

dflems:
`chruby` is a function, not a binary. Since `xcbuild-safe.sh` is run in a subshell, I don't think it has access to functions declared outside of it unless sourced-in directly. I still think my comments in #5240 are relevant to this discussion:

> RVM is the invasive one that needs special logic. `rbenv` and `chruby` just need the ruby environment unset.  Aside from the RVM logic, this should work fine as a catch-all for other ruby managers (at least it's worked for me with chruby).
>
> ```bash
unset BUNDLE_BIN_PATH
unset BUNDLE_GEMFILE
unset GEM_HOME
unset GEM_PATH
unset RUBYLIB
unset RUBYOPT
unset RUBY_ENGINE
unset RUBY_ROOT
unset RUBY_VERSION
unset _ORIGINAL_GEM_PATH
```

**Comment by tcurdt** [comment]
@dflems I am not sure I am getting your point.

This is how one includes `chruby`:

```
source /usr/local/opt/chruby/share/chruby/chruby.sh
source /usr/local/opt/chruby/share/chruby/auto.sh
chruby ruby-2.3
```

Because it's function it is `type -t chruby`.

I also don't see a good reason why one should use a custom way of unsetting the `ruby` env. It's the contract of `chruby` to restore the env - why would we want to duplicate that?

**Comment by dflems** [comment]
@tcurdt:

chruby is added by sourcing a few scripts in your shell and it becomes available as a function *in your shell*. We shouldn't make the expectation that that function will be available *anywhere* that `xcbuild-safe.sh` is called. It would be one thing if chruby was a binary and lived on the `PATH`, but I feel it's unwise to make the assumption that the function will be available in all cases.

For instance, this doesn't work for me because I use zsh and source the chruby scripts into my `.zshrc`. The shebang in `xcbuild-safe.sh` is `#!/bin/bash --login` which means it runs bash with a login shell. If you use bash, this may work fine for you. Since I use zsh and don't source the chruby scripts into my bash profile, this will not work for me (and potentially many other people who use zsh primarily and don't keep bash in sync with it). Do you follow?

Unsetting the ruby-specific environment variables will cover the general use-case for chruby and rbenv across-the-board (and maybe RVM too). We might be able go go even further by just obliterating *anything* set in the user shell and starting fresh:

```sh
#!/usr/bin/env -i sh
# ^ starts a fresh shell with no environment

# source system-wide profile
source /etc/profile

# xcode should not be polluted :D
xcodebuild "$@"
```

tcurdt:
I am using a `.profile` and no `.bashrc` but that might be beside the point.

Right now the assumption is that if there is a `chruby` function the ruby version could potentially be non-system and if there is no `chruby` function it is safe to assume it's system ruby (or at least not altered by `chruby`). If that assumption is wrong you might have a point.

I don't understand your scenario yet though. If you use `chruby` from zsh - but the `xcbuild-safe.sh` uses bash (via shebang) - why should you have anything but the sytem ruby in your bash?

Resetting the full env sounds like an easy approach on the first glance - but users might expect env variables to be there that suddenly aren't. I am not convinced this will turn out to be the best approach in real life.

milch:
After looking at the code and the discussion, I agree with @dflems that this is a very specific solution that probably wouldn't work for most users. I'm closing this for now, as the underlying issues have been resolved (see [5240#issuecomment-253082825](https://github.com/fastlane/fastlane/issues/5240#issuecomment-253082825)) - that should affect most users. 

If there is still an issue, let us know and we can reopen the discussion! 🚀

PR #6311, opened 2016-09-27
Fail build when asking for user or password on non-interactive shell

KrauseFx:
Until now the build would be hanging when the password was missing


viktorbenei:
Awesome, thanks @KrauseFx ! ;)

mfurtak (approved):

PR #6322, opened 2016-09-28
Preventing actions called from another action to show in summary

### LINKED ISSUE #4546: Incorrect time in summary when using other_action
### New Issue Checklist
### Issue Description

When calling an action from another action using `other_action` (https://github.com/fastlane/fastlane/pull/4441), the inner action shows in the summary, in a way that you'd think that it took the time of the inner action + the action itself (see example).

I see two ways of solving it:
1. Don't show actions called from another action in the summary. When an action needs to call another, that's an implementation detail and the user of the action shouldn't need to know about it.
2. Change the summary to allow nested levels of actions. Something like this:

```
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | opt_out_usage                       | 0           |
| 2    | Verifying required fastlane version | 0           |
| 3    | default_platform                    | 0           |
| 4    | is_ci                               | 0           |
| 5    | bundle_install                      | 0           |
| 6    | is_ci                               | 0           |
| 7    | cocoapods_if_needed                 | 62          |
|      |    > cocoapods                      | 51          |
+------+-------------------------------------+-------------+
```

However this could become ugly, as you could have several levels of actions calling others (although not usual).

I think option 1 is the way to go.

```
$ fastlane pod_install

...
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | opt_out_usage                       | 0           |
| 2    | Verifying required fastlane version | 0           |
| 3    | default_platform                    | 0           |
| 4    | is_ci                               | 0           |
| 5    | bundle_install                      | 0           |
| 6    | is_ci                               | 0           |
| 7    | cocoapods                           | 51          |
| 8    | cocoapods_if_needed                 | 62          |
+------+-------------------------------------+-------------+
```
### Configuration Files

Please copy the complete content of your `Fastfile` and any other configuration files you use below:

Fastfile:

``` ruby
fastlane_version "1.86.1"

default_platform :ios

platform :ios do
  desc "Runs pod install if needed"
  lane :pod_install do
    cocoapods_if_needed
  end
end
```

`cocoapods_if_needed.rb`:

``` ruby
module Fastlane
  module Actions
    class CocoapodsIfNeededAction < Action

      def self.needs_to_run_pod_install
        begin
          !FileUtils.compare_file('Podfile.lock', 'Pods/Manifest.lock')
        rescue  Exception => e
          true
        end
      end

      def self.run(params)
        if needs_to_run_pod_install
          sh 'bundle exec pod repo update'

          other_action.cocoapods
        else
          UI.message "🍫  Skipping pod install because Pods seems updated".blue
        end
      end

      # @!group Documentation

      def self.description
        "Runs pod repo update and pod install if needed"
      end

      def self.details
        "Checks if the sandbox is in sync with the Podfile.lock. If not, runs pod repo " \
        "update and then pod install"
      end

      def self.available_options
        []
      end

      def self.authors
        ["marcelofabri"]
      end

        true
      end
    end
  end
end
```
### Environment

fastlane version (run `fastlane -v`): fastlane 1.86.1

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)? yep

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)? rvm

marcelofabri:

This should fix #4546. We probably should add at least a test to make sure we don't have regressions on this, but just wanted to see it the implementation makes sense first.

Just to make clear, this hides the "inner actions" from the summary:

```
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | opt_out_usage                       | 0           |
| 2    | Verifying required fastlane version | 0           |
| 3    | default_platform                    | 0           |
| 4    | is_ci                               | 0           |
| 5    | bundle_install                      | 0           |
| 6    | is_ci                               | 0           |
| 7    | cocoapods_if_needed                 | 98          |
+------+-------------------------------------+-------------+
```

You can compare to the behavior before this PR below:

```
$ fastlane pod_install

...
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | opt_out_usage                       | 0           |
| 2    | Verifying required fastlane version | 0           |
| 3    | default_platform                    | 0           |
| 4    | is_ci                               | 0           |
| 5    | bundle_install                      | 0           |
| 6    | is_ci                               | 0           |
| 7    | cocoapods                           | 51          |
| 8    | cocoapods_if_needed                 | 62          |
+------+-------------------------------------+-------------+
```

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -109,7 +109,9 @@ module Fastlane
 
     # This is being called from `method_missing` from the Fastfile
     # It's also used when an action is called from another action
-    def trigger_action_by_name(method_sym, custom_dir, *arguments)
+    # @param from_action Indicates if this action is being trigged by another action.
+    #                    If so, it won't show up in summary.
+    def trigger_action_by_name(method_sym, custom_dir, from_action, *arguments)
       # First, check if there is a predefined method in the actions folder
       class_ref = class_reference_from_action_name(method_sym)
       unless class_ref
@@ -130,7 +132,7 @@ module Fastlane
       if class_ref
         if class_ref.respond_to?(:run)
           # Action is available, now execute it
-          return self.execute_action(method_sym, class_ref, arguments, custom_dir: custom_dir)
+          return self.execute_action(method_sym, class_ref, arguments, custom_dir: custom_dir, from_action: from_action)
         else
           UI.user_error!("Action '#{method_sym}' of class '#{class_name}' was found, but has no `run` method.")
         end
@@ -197,7 +199,7 @@ module Fastlane
       end
     end
 
-    def execute_action(method_sym, class_ref, arguments, custom_dir: nil)
+    def execute_action(method_sym, class_ref, arguments, custom_dir: nil, from_action: false)
       if custom_dir.nil?
         custom_dir ||= "." if Helper.test?
         custom_dir ||= ".."
@@ -209,7 +211,10 @@ module Fastlane
 
       begin
         Dir.chdir(custom_dir) do # go up from the fastlane folder, to the project folder
-          Actions.execute_action(class_ref.step_text) do
+          # If another action is calling this action, we shouldn't show it in the summary
+          # (see https://github.com/fastlane/fastlane/issues/4546)
+          action_name = from_action ? nil : class_ref.step_text
+          Actions.execute_action(action_name) do
             # arguments is an array by default, containing an hash with the actual parameters
             # Since we usually just need the passed hash, we'll just use the first object if there is only one
             if arguments.count == 0
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -143,7 +143,7 @@ module Fastlane
 
     # Is used to look if the method is implemented as an action
     def method_missing(method_sym, *arguments, &_block)
-      self.runner.trigger_action_by_name(method_sym, nil, *arguments)
+      self.runner.trigger_action_by_name(method_sym, nil, false, *arguments)
     end
 
     #####################################################
--- a/fastlane/lib/fastlane/other_action.rb
+++ b/fastlane/lib/fastlane/other_action.rb
@@ -22,6 +22,7 @@ module Fastlane
 
       self.runner.trigger_action_by_name(method_sym,
                                          "./fastlane",
+                                         true,
                                          *arguments)
     end
   end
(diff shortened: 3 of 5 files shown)


KrauseFx:
Hey @marcelofabri, sorry for the late reply on this, and thanks for your pull request. I don't think we can merge the pull request as it is now. We'd need some more documentation in the code, and also tests covering this change. 
Is that something you could do?

marcelofabri:
No worries! Where exactly do you want to see more docs? In `trigger_action_by_name` header?

Also, can you point me to where this is being tested or how could I add tests for this? (Since it needs to capture the summary output)

ohwutup:
Hey @marcelofabri - Apologies for the delay here- I think a good place to write a test for this would be in `fastlane_runner_spec.rb`. And Regarding adding more docs to the code, I think if you put a little comment above `trigger_action_by_name`'s definition about what the `from_action` param is doing, it might be helpful. Thanks! 🐙

marcelofabri:
@ohwutup I've rebased this and added a comment in the header, but I'm still having trouble on how to test this without mocking the whole stack. Are there any integration tests? Maybe this would  be better tested in that way.

marcelofabri:
@ohwutup So, I took another look and found `action_spec`. I was able to add tests there in e4ff4d6a. I'm not sure if it's the best approach, but the new tests do fail without 52bfe29 and pass with it. 🎉

fastlane/lib/fastlane/runner.rb
```diff
@@ -109,7 +109,9 @@ def find_alias(action_name)
 
     # This is being called from `method_missing` from the Fastfile
     # It's also used when an action is called from another action
-    def trigger_action_by_name(method_sym, custom_dir, *arguments)
+    # @param from_action Indicates if this action is being trigged by another action.
+    #                    If so, it won't show up in summary.
+    def trigger_action_by_name(method_sym, custom_dir, from_action, *arguments)
```
milch: In the past we've had trouble with adding parameters to existing functions breaking people's setups, so could you give this a default value please? e.g. `def trigger_action_by_name(method_sym, custom_dir, from_action: false, *arguments)`

milch (changes requested):

fastlane/lib/fastlane/runner.rb
marcelofabri: Sure, although I'd say that this method shouldn't really be called by end users 😅

marcelofabri: Actually, it seems that this is not valid syntax because of the `*arguments` later 😬

milch: > Sure, although I'd say that this method shouldn't really be called by end users 😅

That's never stopped someone from doing things before 😉 

> Actually, it seems that this is not valid syntax because of the *arguments later 😬

Damn, you're right. I just double checked that you changed all occurences of trigger_action_by_name, so we should be good.

milch (approved):

PR #6326, opened 2016-09-28
Add `rake update_dependencies` to automatically update internal dependencies

KrauseFx:
We’ve spent enough time doing this manually, time for the bots to take over

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -48,19 +48,17 @@ task :update_dependencies do
     content = File.read(current_gemspec_path)
 
     content.gsub!(regex) do |full_match|
-      matched = full_match.match(regex)
-      gem_name = matched[1]
-      current_version_number = Gem::Version.new(matched[2]) # used to detect if we actually changed something
-
+      gem_name = $1
+      current_version_number = Gem::Version.new($2) # used to detect if we actually changed something
 
       version_path = "./#{gem_name}/lib/#{gem_name}/version.rb"
       if File.exist?(version_path) && gem_name != "screengrab" # internal dependency
         version = Gem::Version.new(File.read(version_path).match(/VERSION.=..(\d+\.\d+\.\d+)./)[1])
-        major_version = Gem::Version.new("#{version.segments[0] + 1}.0.0")
+        next_major_version = Gem::Version.new("#{version.segments[0] + 1}.0.0")
 
         puts "🚀  Updating #{gem_name} from #{current_version_number} to #{version} for #{gem_name}" if version != current_version_number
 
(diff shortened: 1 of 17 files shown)


Rakefile
```diff
@@ -39,6 +39,38 @@ task :rubygems_admins do
...
+
+  regex = %r{spec.add_dependency .(.+).\, .\>\= (\d+\.\d+\.\d+).\, .\< \d+\.\d+\.\d+.}
+
+  Dir["./**/*.gemspec"].each do |current_gemspec_path|
+    content = File.read(current_gemspec_path)
+
+    content.gsub!(regex) do |full_match|
+      matched = full_match.match(regex)
+      gem_name = matched[1]
+      current_version_number = Gem::Version.new(matched[2]) # used to detect if we actually changed something
```
mfurtak: According to the docs for `gsub` you can use `$1` and `$2` to refer to these match values rather than running the regex match again.

Rakefile
```diff
@@ -39,6 +39,38 @@ task :rubygems_admins do
...
+
```
asfalcone: extra newlines here?

Rakefile
```diff
@@ -39,6 +39,38 @@ task :rubygems_admins do
...
+
+      version_path = "./#{gem_name}/lib/#{gem_name}/version.rb"
+      if File.exist?(version_path) && gem_name != "screengrab" # internal dependency
+        version = Gem::Version.new(File.read(version_path).match(/VERSION.=..(\d+\.\d+\.\d+)./)[1])
+        major_version = Gem::Version.new("#{version.segments[0] + 1}.0.0")
```
asfalcone: Should this be called `next_major_version`?

Rakefile
```diff
@@ -39,6 +39,38 @@ task :rubygems_admins do
   end
 end
 
+task :update_dependencies do
+  puts "Updating all internal fastlane dependencies"
+
+  regex = %r{spec.add_dependency .(.+).\, .\>\= (\d+\.\d+\.\d+).\, .\< \d+\.\d+\.\d+.}
```
KrauseFx: Currently only works with 3 parts, which is the notation we've used until now. Should we keep this assumption?

Rakefile
KrauseFx: Oh nice, I didn't know that 👍

KrauseFx:
Thanks, I updated the PR and resolved the merge conflicts after the xcpretty update 👍

Rakefile
mfurtak: I think it is OK, maybe just document the assumption in a comment for the code :+1:

mfurtak (approved):
A comment about version assumptions would be :100:, but approved!

Issue #6320, opened 2016-09-28
No matching provisioning profiles found and can not create a new one because you enabled `readonly`


lukewakeford:
I am experiencing exactly the same issue as discussed here -> #1979
I am running Fastlane version 1.104.0

Here is my match code (which I run twice for two app ids) both for enterprise/InHouse certificates and profiles.

    match(
      type: "appstore",
      git_url: ENV['PROVISIONING_REPO'],
      username: ENV['PRODUCE_USERNAME'],
      app_identifier: ENV['PRODUCE_APP_IDENTIFIER'],
      readonly: true,
      verbose: true
    )

and here is the output

```
[16:34:19]: -------------------
[16:34:19]: --- Step: match ---
[16:34:19]: -------------------

+-----------------------+---------------------------------------------------------+
|                             Summary for match 0.8.0                             |
+-----------------------+---------------------------------------------------------+
| type                  | appstore                                                |
| git_url               | git@xxxxxxxxxxxxxxxx|
| username              | xxxxxxxxxxxxxxxx                                       |
| app_identifier        | xxxxxxxxxxxxxxxx                            |
| readonly              | true                                                    |
| verbose               | true                                                    |
| git_branch            | master                                                  |
| keychain_name         | login.keychain                                          |
| force                 | false                                                   |
| shallow_clone         | false                                                   |
| force_for_new_devices | false                                                   |
| skip_docs             | false                                                   |
+-----------------------+---------------------------------------------------------+

INFO [2016-09-28 16:34:19.67]: Cloning remote git repo...
INFO [2016-09-28 16:34:19.67]: $ GIT_TERMINAL_PROMPT=0 git clone 'xxxxxxxxxxxxxxxx' '/var/folders/jg/679wq2q95q780r5nxgxpt8yh0000gn/T/d20160928-60028-9bbkh8'
INFO [2016-09-28 16:34:19.67]: ▸ Cloning into '/var/folders/jg/679wq2q95q780r5nxgxpt8yh0000gn/T/d20160928-60028-9bbkh8'...
INFO [2016-09-28 16:34:20.83]: ▸ remote: Counting objects: 304, done.
INFO [2016-09-28 16:34:21.02]: ▸ remote: Compressing objects: 100% (295/295), done.
INFO [2016-09-28 16:34:21.42]: ▸ remote: Total 304 (delta 51), reused 0 (delta 0)
[...]

PR #6328, opened 2016-09-28
Add error message to make it more clear enterprise profiles are not supported

KrauseFx:
More issues around enterprise https://github.com/fastlane/fastlane/issues/6320


asfalcone (approved):
👍

Issue #6327, opened 2016-09-28
SIGH_UDID should be SIGH_UUID


maschall:
- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

UDID - Unique Device Identifiers
UUID - Universal Unique Identifier

In the Provision profile Plist the element is UUID, but we are using UDID, which is incorrect.

Anyway we can deprecate the incorrect one and change to UUID?

KrauseFx:
Thanks for letting us know, I'm aware of this, however we'd still have to support both to not break any setups. This would add some complexity, however I'm more than happy to review a PR fixing this 👍

maschall:
@KrauseFx What do you think if we refactored lane_context and ENV, into objects that could redirect UUID to UDID.

Example:
Environment["SIGH_UUID"] would read ENV["SIGH_UUID"] first and then fall back to ENV["SIGH_UDID"]

I'm thinking this should help in the future if we needed to deprecate/change an environment variables again.

TKBurner:
@maschall Thanks for the PR, we need to make sure to not break any setups. We'll work with you over there. Hope that you are having a great week! 👍

PR #6339, opened 2016-09-29
Deprecate SIGH_UDID in favor of corrected SIGH_UUID

maschall:

#6327


fastlane/lib/fastlane/actions/update_project_code_signing.rb
```diff
@@ -39,8 +40,12 @@ def self.available_options
                                        end),
           FastlaneCore::ConfigItem.new(key: :udid,
                                        env_name: "FL_PROJECT_SIGNING_UDID",
-                                       description: "The UDID of the provisioning profile you want to use",
-                                       default_value: ENV["SIGH_UDID"])
+                                       description: "DEPRECATED: see :uuid",
+                                       default_value: ENV["SIGH_UUID"]),
```
maschall: I changed this to UUID, because I believe the intention is for this to default to the result of the sigh action, as opposed to defaulting the Environment variable set by the user.  I could be wrong and happy to revert this line back to UDID.

fastlane/lib/fastlane/actions/sigh.rb
```diff
@@ -22,11 +23,13 @@ def self.run(values)
           Actions.lane_context[SharedValues::SIGH_PROFILE_PATH] = path # absolute path
           Actions.lane_context[SharedValues::SIGH_PROFILE_PATHS] ||= []
           Actions.lane_context[SharedValues::SIGH_PROFILE_PATHS] << path
-          Actions.lane_context[SharedValues::SIGH_UDID] = ENV["SIGH_UDID"] if ENV["SIGH_UDID"] # The UDID of the new profile
+
+          uuid = ENV["SIGH_UUID"] || ENV["SIGH_UDID"] # the UUID of the profile
```
maschall: I feel like we are using ENV["SIGH_UUID"] to pass the uuid from Sigh:Worker.start command, opposed to respecting the Environment variable from the user.

If this is the case, I think we can just remove ENV["SIGH_UDID"] in a few places.

milch:
Hey @maschall, thanks for your contribution! 

The change looks good to me, and I see there was some discussion over at the issue that this is something we might want to have. However, we don't want to break anything for existing users. It seems to me that existing setups should work with this without any problems, is that correct?

maschall:
@milch My testing thus far hasn't come up with any issues.  Majority of developers I've talked to don't override the SIGH_UDID in the fastfile, they use what is set by the sigh action.  Even if they are setting the SIGH_UDID environment variable/lane context we will be still accept that value.

milch:
Okay, thank you, that's what I thought. 

I'm going to accept this, however, there's some stuff left to do before we can merge. Can you move all of the changes to `sigh` into their own PR please? All changes to `fastlane` can stay here.

maschall:
sure thing

milch (approved):
👍

milch:
Thanks for your contribution! 🚀 Also, sorry for the long delay before we could review/address your PR.

PR #6344, opened 2016-09-29
Builds S3 domain using SDK Endpoints.hostname method

lyricsboy:
This fixes an issue with the `s3` action where specifying the AWS region as `us-east-1` would incorrectly build an S3 URL to a domain that doesn't exist.

Instead, it now uses the AWS SDK's built-in support for providing the domain for a given service and region.

I'd appreciate testing from someone who uses a region other than `us-east-1` to ensure it works as expected. Unfortunately the action is not particularly friendly to unit testing as written, and I didn't want to refactor the whole shebang.


ohayon (approved):
This looks great! Thanks @lyricsboy ! 🚀

PR #6347, opened 2016-09-29
Ignore SIGH_PROFILE_PATH when using newer api

maschall:

For Xcode 7+, there is a error "You're using Xcode 7 or above, the `provisioning_profile_path` value will be ignored" that gym is outputting.  In my case, I am using `sigh` before `gym` and seeing this error pop up.  Originally this happened when some other issue appeared and this led me astray for a bit.

So this change will not forward the `SIGH_PROFILE_PATH` onto `gym` if the developer is not using the legacy api.

Note: It will still warn, if the developer explicitly defines `provisioning_profile_path` in the action options.

Lastly: There are no specs for gym action, that I can find.  I'd be happy to write some just wanted to get this up first.

### DIFF
--- a/fastlane/lib/fastlane/actions/gym.rb
+++ b/fastlane/lib/fastlane/actions/gym.rb
@@ -12,7 +12,9 @@ module Fastlane
         begin
           FastlaneCore::UpdateChecker.start_looking_for_update('gym') unless Helper.is_test?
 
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):

KrauseFx:
Great addition, thanks for your contribution @maschall  👍

PR #6355, opened 2016-09-30
Improve docs

KrauseFx:


fastlane/lib/fastlane/actions/clean_build_artifacts.rb
```diff
@@ -35,12 +35,12 @@ def self.available_options
       end
 
       def self.description
-        "Deletes files created as result of running ipa, cert, sigh or download_dsyms"
+        "Deletes files created as result of running gym, cert, sigh or download_dsyms"
```
ohwutup: should this be _gym_?

ohayon (changes requested):
Just that one little change but otherwise 👍

fastlane/lib/fastlane/actions/clean_build_artifacts.rb
KrauseFx: Yes

ohwutup: also should actions names maybe also be italicized?

KrauseFx: The description is printed in the action list also, so we don't always render it in markdown

ohwutup: ohhh! got it!

ohayon (approved):

PR #6363, opened 2016-10-02
Added support to absolute path in action commit_version_bump

### LINKED ISSUE #6238: Actions `increment_version_number` and `commit_version_bump` have different behaviour for `xcodeproj` parameter
### New Issue Checklist
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

Actions `increment_version_number` and `commit_version_bump` have different behaviour for `xcodeproj` parameter

[increment_version_number](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/increment_version_number.rb#L18) accepts full path but [commit_version_bump](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/commit_version_bump.rb#L11) only accepts relative path.

Would be great to user full path for both.
### Environment

fastlane version (run `fastlane -v`):
`fastlane 1.103.0`

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?
Yes

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?
`rbenv`

diogot:
- [X] Run `rspec` for all tools you modified
- [X] Run `rubocop -a` to ensure the code style is valid

Fix #6238

### DIFF
--- a/fastlane/lib/fastlane/actions/commit_version_bump.rb
+++ b/fastlane/lib/fastlane/actions/commit_version_bump.rb
@@ -8,13 +8,15 @@ module Fastlane
         require 'set'
         require 'shellwords'
 
-        xcodeproj_path = params[:xcodeproj] ? File.expand_path(File.join('.', params[:xcodeproj])) : nil
-
         # find the repo root path
         repo_path = Actions.sh('git rev-parse --show-toplevel').strip
         repo_pathname = Pathname.new(repo_path)
 
+        xcodeproj_path = params[:xcodeproj] ? File.expand_path(params[:xcodeproj]) : nil
+
         if xcodeproj_path
+          # if directory don't exists the path may be relative
+          xcodeproj_path = File.expand_path(File.join('.', params[:xcodeproj])) unless File.directory?(xcodeproj_path)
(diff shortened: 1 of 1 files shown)


KrauseFx:
Sorry, I'm not sure if that code makes sense. Are you sure that's the behaviour you want? Is it the same as the other action?

diogot:
Let me try to explain what I tried to do.

First I don't want to break the default behaviour and I want to be able to pass the full path as parameter.

So first I take the parameter and expand, this will allow to use `~/project`, if this directory don't exists, this means that the path is relative and I should fall back to the old behaviour. Add a `./` and expand it.

Does it make sense?

bguidolim:
@diogot I think you could solve it easier using `(Pathname.new xcodeproj_path).absolute?`.

ohwutup:
Hey everyone! I am going to close this issue for now as there has not been much activity on here recently. Please feel free to comment and continue discussion and we can reopen it if necessary!

Thanks!! 🐙

PR #6364, opened 2016-10-02
support signing dylibs located in main bundle path

iMokhles:

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -530,7 +530,29 @@ function resign {
         log "Updating the bundle version (CFBundleVersion) from '$CURRENT_VALUE' to '$BUNDLE_VERSION'"
         PlistBuddy -c "Set :CFBundleVersion $BUNDLE_VERSION" "$APP_PATH/Info.plist"
     fi
-
+    
+    # Check for and resign any dylibs in the main bundle directory 
+    DYLIBS_DIR="$APP_PATH/*dylib"
+    if [ -d "$DYLIBS_DIR" ];
+    then
+        if [ "$TEAM_IDENTIFIER" == "" ];
+        then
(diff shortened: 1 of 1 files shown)


ohwutup:
Hey @iMokhles, thanks for submitting the PR, could you just tell us a little about the use case for this?

iMokhles:
@ohwutup hey thanks for responding 
SO, iOS root known as JAILBREAK inject dylibs into Mach-O to modify codes at runtime through libraries aka ( Tweaks ) so we can also modify apps codes at runtime without JAILBREAK while swizzling it's code by injecting libraries into it'a Mach-O using injection tools ( like optool l, yololib, etc ) so the app code run signed codes which can modify it's functionality, and some people including ( ME ) places those libraries ( dylibs files ) inside the main bundle directory ( so it should be signed with the same way as frameworks to run into inside the app )

Hope it's clear :)

ohwutup:
Hey @iMokhles, I'm not sure if it is totally clear, is this something that only Jailbreak tweak developers are going to get value out of? 🐙

iMokhles:
hey @ohwutup
yes and no ( anyone can inject modifications without jailbreak to swizzle functions ) but for now ( some jailbreak developers including -ME ) doing those stuffs to improve functionality of some AppStore apps ( like I did here https://github.com/iMokhles/Tweaks-for-SideLoaded-Apps ) without jailbreak at all and as I explained before those tweaks ( dylibs files ) need to be signed to work, and sometimes they are placed under the main bundle directory without sub-folders


thanks 😁

ohwutup:
Hey @iMokhles,

So I think that we would rather keep this out of our core functionality in resigning apps with _sigh_. That being said, I think that you could definitely create a plugin (https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md) that could be inserted into your _fastlane_ invocations.

I am going to close this PR for now, but feel free to keep discussion going here if you'd like!
🐙 🚀 🐮

PR #6384, opened 2016-10-04
[match] Don’t hang when git credentials are not there

KrauseFx:
This would be hanging when not running in read-only mode, having a public repo, but not having push access


mfurtak:
This is great! However, it feels like we are putting this magic sauce in a lot of places. It would be good to think a bit about whether we can centralize knowledge about this flag so that we don't need to remember to use it everywhere 👍

mfurtak (approved):

Issue #6387, opened 2016-10-04
Update provisioning profile import for Sierra


KrauseFx:
The error message

> security: SecPolicySetValue: One or more parameters passed to a function were not valid.

is shown for the command

```
security cms -D -i "#{path}"
```

however the command seems to be valid according to the documentation, and works as expected. We need to ignore the error message, as a fellow other project does (https://github.com/DanTheMan827/ios-app-signer/commit/3e4749855f3aab100f8f91ca0d34cf2098976f50)

Related issues:
- https://github.com/fastlane/fastlane/issues/5649
- https://github.com/fastlane/fastlane/issues/6233
- https://github.com/fastlane/fastlane/issues/6151
- https://github.com/fastlane/fastlane/issues/5649

ericlewis:
+1

KrauseFx:
I just pushed https://github.com/fastlane/fastlane/pull/6388

@ericlewis if you have a chance, I'd love to get a +1 on the fix 🚀

ericlewis:
@KrauseFx direct me to the docs necessary to do so an I will.

damirjuretic:
@KrauseFx fix that you have made is not working. 

Message:`security: SecPolicySetValue: One or more parameters passed to a function were not valid.`
Is not piped to STDERR so fix with 2>/dev/null won't work.

A message is piped to STDOUT so the only solution would be analyzing STDOUT and removing the unnecessary message.

TKBurner:
Hi @damirjuretic Can you run `fastlane run update_fastlane` and then share your input and output with this error? Thanks so much!

KrauseFx:
@damirjuretic That's why we're using `2> /dev/null`, so the stderror gets piped there

damirjuretic:
Hi @TKBurner, I have updated fastlane and now is working properly. Thank you @KrauseFx 👍

KrauseFx:
Yeah 🎉

PR #6388, opened 2016-10-04
Don’t show error message when importing provisioning profile

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/6387

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -24,7 +24,7 @@ module FastlaneCore
       def parse(path)
         require 'plist'
 
-        plist = Plist.parse_xml(`security cms -D -i "#{path}"`)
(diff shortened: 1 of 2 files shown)


asfalcone (approved):
This looks pretty good. 

It might be useful to capture the error and display it in the future, if there is some other error that isn't a macOS farse. This code would mask other errors that might some up.

KrauseFx:
I agree. What's the best approach to do so? Writing this to a file, and reading it seems like a lot of work just to do that. Should we merge and release, and then think about showing the error message @asfalcone?

asfalcone:
Capture all of the output to a string, see if the first line is the error, print as a warning if it is, send the rest of the lines to the Plist parser?

Totally fine to revisit later. Just wanted to call out that this could bite us later!

KrauseFx:
Mh, kind of difficult to see when the error ends, and when the plist starts, right? We could also switch to the fully featured Ruby IO thingy, where you have different pipes.

Issue #5649, opened 2016-08-09
match - macOS Sierra keychain error


jmjordan:
### Issue Description

When using macOS Sierra beta I am experiencing an issue with `match`. I receive an error saying that it cannot locate my keychain. It appears that Sierra is saving keychains with the keychain-db extension instead of keychain. If I change the keychain_name value to login.keychain-db it runs successfully.

##### Complete output when running fastlane, including the stack trace and command used

```
$ match adhoc

+-----------------------+---------------------------------------------+
|                       Summary for match 0.6.2                       |
+-----------------------+---------------------------------------------+
| git_url               |  |
| type                  | adhoc                                       |
| git_branch            | master                                      |
| app_identifier        |                         |
| username              |                      |
| keychain_name         | login.keychain                              |
| readonly              | false                                       |
| team_id               |                                  |
| verbose               | false                                       |
| force                 | false                                       |
| shallow_clone         | false                                       |
| force_for_new_devices | false                                       |
| skip_docs   false                                       |
+-----------------------+---------------------------------------------+

[21:11:03]: Cloning remote git repo...
[21:11:10]: 🔓  Successfully decrypted certificates repo
[21:11:10]: Verifying that the certificate and profile are still valid on the Dev Portal...
[21:11:11]: Installing certificate...
[21:11:11]: 🔒  Successfully encrypted certificates repo

Looking for related GitHub issues on fastlane/fastlane...

Found no similar issues. To create a new issue, please visit:
https://github.com/fastlane/fastlane/issues/new

[!] Could not locate the provided keychain. Tried:
	/Users/username/Library/Keychains/login.keychain
	login.keychain```

### Environment

match 0.6.2

**Comment by TKBurner** [comment]
Hi @jmjordan What happens if you unlock the login keychain prior to the match command? Do you still see this error?

**Comment by TKBurner** [comment]
Hi @jmjordan Thanks for following up. Can you add the --verbose flag to your command and share the full output? Thanks!

**Comment by jmjordan** [comment]
```

+-----------------------+---------------------------------------------+
|                       Summary for match 0.6.2                       |
+-----------------------+---------------------------------------------+
| verbose               | true                                        |
| git_url               | git@bitbucket.org:xxxxxxxxxxxxxxx.git |
| type                  | adhoc                                       |
| git_branch            | master                                      |
| app_identifier        | com.xxxxx.xxxxx                         |
| username              | xxxxx@mac.com                     |
| keychain_name         | login.keychain                              |
[...]

PR #6389, opened 2016-10-05
[mach] Fix Keychain handling on macOS Sierra

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/5649

### DIFF
--- a/match/lib/match/utils.rb
+++ b/match/lib/match/utils.rb
@@ -11,10 +11,7 @@ module Match
       # We need to expand each path because File.exist? won't handle directories including ~ properly
       #
       # We also try to append `-db` at the end of the file path, as with Sierra the default Keychain name
-      # has changed for some users:
-      # - https://github.com/fastlane/fastlane/issues/5649
-      # - https://github.com/fastlane/fastlane/issues/6151
-      # - https://github.com/fastlane/fastlane/issues/5649
+      # has changed for some users: https://github.com/fastlane/fastlane/issues/5649
       #
       keychain_paths = [
         File.join(Dir.home, 'Library', 'Keychains', keychain),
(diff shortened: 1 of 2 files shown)


match/spec/utils_spec.rb
```diff
@@ -23,12 +24,25 @@
...
+        expect(File).to receive(:exist?).with("#{Dir.home}/Library/Keychains/my/special.keychain-db").and_return(false)
         expect(File).to receive(:exist?).with('/my/special.keychain').and_return(false)
+        expect(File).to receive(:exist?).with('/my/special.keychain-db').and_return(false)
 
         expect do
           Match::Utils.import('item.path', '/my/special.keychain')
         end.to raise_error(/Could not locate the provided keychain/)
       end
+
+      it "tries to find the macOS Sierra keychain too" do
```
asfalcone: 👍

match/lib/match/utils.rb
```diff
@@ -9,9 +9,18 @@ def self.import(item_path, keychain)
       # as the keychain path.
       #
       # We need to expand each path because File.exist? won't handle directories including ~ properly
+      #
+      # We also try to append `-db` at the end of the file path, as with Sierra the default Keychain name
+      # has changed for some users:
+      # - https://github.com/fastlane/fastlane/issues/5649
```
asfalcone: Do we need to link to all of these issues in the comments? I feel like adding all of these here clutters up the code quite a bit.

asfalcone (approved):

match/lib/match/utils.rb
KrauseFx: I personally think it's good to provide some context and background on why a certain line or method was added. This way, future fastlane team can see why some things were added, including the complete discussion of why this change is necessary, with an easy way to actively reach out to people who had this problem. 
I agree with you that 3 issues might be too many, we can reduce it to the most important one, if you prefer that 👍

KrauseFx: Whops, 2 are duplicates anyway, let me update the PR 👍

KrauseFx:
Updated 👍

PR #6411, opened 2016-10-05
Version Bump

mfurtak:
Changes since release 1.10.0:
- Update internal dependencies
- Improved `rbenv` handling in gym
- Update xcpretty dependency to 0.2.3 (#6323)
- Don’t detect and set provisioning profile if legacy API is disabled (#6253)


asfalcone:
#6253 seems like it might be worthy of a minor version bump instead of a patch. It's a somewhat significant change in behavior

asfalcone (approved):

PR #5958, opened 2016-08-30
Display error when parsing a .strings file returns empty results (#1747)

mihaibabusca:
Display error messages when the parser can not read the strings file or the file contains empty data.
Related to this issue: #1747, fixed as discussed with @mfurtak in issue #5956

PR #4849, opened 2016-05-25
Escape single quotes in frameit title strings

dwhitlow:
Fixes #4851

PR #3989, opened 2016-03-30
Added interline spacing to frameit titles

jvanzummeren:
Added an optional config to Framefile.json to Frameit which allows spacing between lines in case of a multi-lined screenshot title.

PR #6414, opened 2016-10-05
[frameit] Version bump

KrauseFx:
- Update internal dependencies (#6104)
- Add a test for error shown with a UTF-8 encoded strings file (#5960)
- Display error when parsing a .strings file returns empty results fixing #1747 (#5958)
- Add a ROOT path constant for frameit (#5843)
- Escape single quotes in frameit title strings (#4849)
- Added interline spacing to frameit titles (#3989)


ohayon (approved):
👍

PR #6422, opened 2016-10-06
Update snapshot reporter to work with the new iPad Pro

KrauseFx:


asfalcone (approved):

snapshot/lib/snapshot/reports_generator.rb
```diff
@@ -59,7 +59,9 @@ def available_devices
         'iPhone6' => "iPhone6 (4.7-Inch)",
         'iPhone5' => "iPhone5 (4-Inch)",
         'iPhone4' => "iPhone4 (3.5-Inch)",
```
mfurtak: Since we were removing "iPhone 4s" in another PR... is "iPhone4" still a thing?

KrauseFx: People still might use _snapshot_ for iOS 9, so the HTML reporter should still support them. We just removed it from the default `Snapfile`, since new projects will not find it useful

PR #6423, opened 2016-10-06
Fix upload of image assets to iTunes Connect

KrauseFx:
iTunes Connect requires `height`, `width`, `size` and `checksum` when uploading screenshots. There is not a single GitHub issue, or tweet about this problem, I'm unsure if that's related to my account.


hjanuschka:
looking forward to see this - as it blocks current usage of deliver. @KrauseFx

hjanuschka:
as mentioned in https://github.com/fastlane/fastlane/issues/6426

asfalcone (approved):

PR #6434, opened 2016-10-06
Improve output of changelog

KrauseFx:
- Don’t show command output
- Use of colors
- Remove [tool] prefix if there

From 

screenshot 2016-10-06 12 45 08

to

screenshot 2016-10-06 12 45 00


mfurtak (approved):

KrauseFx:
:boom: making the world a better place

PR #6441, opened 2016-10-06
Downgrade more xcpretty dependencies

KrauseFx:


asfalcone:
Did we decide to wait until `xcpretty 0.2.4` was released?

KrauseFx:
Yeah, I think waiting is the better solution

Issue #4301, opened 2016-04-21
team_id does not automatically select correct team when provisioning in iTunes Connect


ssteinerx:
### Issue Description

My Apple ID belongs to two teams.  I have specified the team_id I wish to use in my Appfile.  It seems that there is no matching between the team_id and the iTunes Connect access corresponding to the team.

I searched for some way to indicate the correct iTunes Connect id (which is different from the team_id), but did not find one.
##### Unexpected output when running fastlane

```
Confusingly, I got this prompt even though I specified a team_id in my Appfile:

Multiple iTunes Connect teams found, please enter the number of the team you want to use:

1) "xxxxx" (2*******)
2) "yyyyy" (1******)
```
### Configuration Files

Fastfile is copied directly from doc, only using :provision for the moment; there is one team_id specified  in the Appfile and I'm not doing anything fancy.
### Environment

fastlane 1.81.0

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?

NO

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?
NO

System ruby: ruby 2.0.0p481 (2014-05-08 revision 45883) [universal.x86_64-darwin14]

All gems up to date.

vpolouchkine:
Can you please still share the Fastfile so we're on the same page? Thanks!

ssteinerx:
```ruby
fastlane_version "1.81.0"

default_platform :ios

platform :ios do
  before_all do
    ENV["SLACK_URL"] = "https://hooks.slack.com/services/...
  end

  # 1
  desc "Creating a code signing certificate and provisioning profile"
  # 2
  lane :provision do
    # 3
    produce(
      app_name: 'TheAppName',
      language: 'English',
      app_version: '1.0',
      sku: 'THESKU'
    )
    # 4
    cert
    # 5
    sigh(force: true)
  end

  error do |lane, exception|
    slack(
      message: exception.message,
      success: false
    )
  end
end
```

ssteinerx:
Appfile:

```
# The bundle identifier of your app
app_identifier "com.xxxz.applicationname"
apple_id "ssteinerX@gmail.com" 	# Your Apple email address
team_id "ABCDEFGHIJ"  	 #  Developer Portal Team ID
```

TKBurner:
Hi @ssteinerx! Thanks for following up. Can you pass the team_id as a parameter to produce directly? This should clear up your issue 👍

[...]

PR #6442, opened 2016-10-07
[spaceship] Validate team ID and show nice error message

KrauseFx:
screenshot 2016-10-06 17 18 29

Related to https://github.com/fastlane/fastlane/issues/4301

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -66,6 +66,23 @@ module Spaceship
 
     # Set a new team ID which will be used from now on
     def team_id=(team_id)
+      # First, we verify the team actually exists, because otherwise iTC would return the
+      # following confusing error message
+      # 
+      #     invalid content provider id
+      # 
+      available_teams = teams.collect do |team|
+        team.fetch("contentProvider", {}).fetch("contentProviderId", nil)
+      end
+
+      result = available_teams.find do |available_team_id|
+        team_id.to_s == available_team_id.to_s
+      end
+
+      unless result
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -66,6 +66,23 @@ def team_id
...
+      # following confusing error message
+      #
+      #     invalid content provider id
+      #
+      available_teams = teams.collect do |team|
+        team.fetch("contentProvider", {}).fetch("contentProviderId", nil)
+      end
+
+      result = available_teams.find do |available_team_id|
+        team_id.to_s == available_team_id.to_s
```
asfalcone: Does it make sense to stop iterating once we find the `available_team_id` that matches?

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -66,6 +66,23 @@ def team_id
 
     # Set a new team ID which will be used from now on
     def team_id=(team_id)
+      # First, we verify the team actually exists, because otherwise iTC would return the
+      # following confusing error message
+      #
+      #     invalid content provider id
+      #
+      available_teams = teams.collect do |team|
+        team.fetch("contentProvider", {}).fetch("contentProviderId", nil)
```
mfurtak: I would suggest that we stop using `fetch` because I think that it has surprising behavior. It looks like you can give it a default value, but actually that doesn't prevent the value returned from being `nil` if the hash entry contains the value `nil` (which happens a lot in Apple's responses)

I would suggest we use `(team['contentProvider'] || {})` instead.

KrauseFx:
@mfurtak pushed the changes 👍

asfalcone (approved):

PR #6445, opened 2016-10-07
Add debug information for Appfile configs

KrauseFx:
This will make it easier for people to detect if an Appfile was loaded, and what values were detected

screenshot 2016-10-06 18 07 12

### DIFF
--- a/credentials_manager/lib/credentials_manager/appfile_config.rb
+++ b/credentials_manager/lib/credentials_manager/appfile_config.rb
@@ -2,6 +2,8 @@ module CredentialsManager
   # Access the content of the app file (e.g. app identifier and Apple ID)
   class AppfileConfig
     def self.try_fetch_value(key)
+      # We need to load the file every time we call this method
+      # to support the `for_lane` keyword
       begin
         return self.new.data[key]
       rescue => ex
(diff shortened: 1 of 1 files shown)


credentials_manager/lib/credentials_manager/appfile_config.rb
```diff
@@ -41,12 +43,35 @@ def initialize(path = nil)
...
+      self.class.already_printed_debug_information ||= {}
+      return if self.class.already_printed_debug_information[self.data]
+      # self.class.already_printed_debug_information is a hash, we use to detect if we already printed this data
+      # this is necessary, as on the course of a fastlane run, the values might change, e.g. when using
+      # the `for_lane` keyword.
+
+      puts "Successfully loaded Appfile at path '#{path}'".yellow
+
+      self.data.each do |key, value|
+        puts "- #{key.to_s.cyan}: '#{value.to_s.green}'"
```
asfalcone: Should we use the table that we use for tool params and plugins?

image 
image

KrauseFx: Great idea, I'll update the PR 👍

KrauseFx: Ah, the code is in _credentials_manager_, which has no dependency to `terminal-table`. I don't think it's worth adding the dependency for this

asfalcone: I guess this is only for verbose logging, to 👌

asfalcone (approved):

PR #6475, opened 2016-10-10
Verify docs on pull requests and releases

KrauseFx:
## Ensure that the docs are still up to date and valid

```
fastlane verify_docs
```

By running this on every pull request, the build will fail if someone just removes an option from one of the tools
## Submit a pull request with the updated actions.md

```
fastlane update_docs
```

By running `fastlane update_docs`, this will automatically submit a pull request to the `fastlane/docs` with the latest documentation. This is automatically called right after releasing a new version of any tool.
## Magic

By having those 2 new lanes, that are both called automatically whenever necessary (on every PR, on every release), our documentation is **always** up to date and works with the latest release of any _fastlane_ tool 🚀

Both lanes are called automatically, whenever they're needed, so no need to do anything manually


asfalcone (approved):

KrauseFx:
Tests are passing 🚀 Ready for review 👍

fastlane/fastlane/Fastfile
```diff
@@ -121,21 +121,63 @@ end
...
+  clone_docs do
+    require 'fastlane/documentation/markdown_docs_generator'
+    actions_md_path = File.expand_path("docs/actions.md")
+    Fastlane::MarkdownDocsGenerator.new.generate!(target_path: actions_md_path)
+
+    # Create a new branch
+    sh "git checkout -b 'update-actions-md'"
+
+    Dir.chdir("fastlane") do # this is an assumption of fastlane, that we have to .. when shelling out
+      if `git status -s`.length > 0
```
asfalcone: Can we do this before checking out the branch? So that if there are no changes we don't end up with this branch? https://github.com/fastlane/fastlane/pull/6475/commits/f85fdc76b59fe9a74cac73ca6d4baa0f8001079f#diff-705d2a114305ea865bfec89cb0c77431R145

asfalcone: Can we go back to master and delete that branch locally when done?

KrauseFx: The branch is just locally in a temporary directory, right? We never push it to the remote

KrauseFx: Check out `clone_docs`, it will clone fastlane/docs in a temporary directory, so we don't have to care about what state we leave the directory afterwards, since we delete it anyway

KrauseFx:
I'm more worried about what happens when the branch name is already taken, because we didn't merge the last PR. I'll update this 👍

KrauseFx:
Ready for another review @asfalcone 👍

KrauseFx:
Ready for another review @asfalcone 👍

asfalcone:
LGTM 🚀

fastlane/fastlane/Fastfile
```diff
@@ -112,13 +112,76 @@ lane :release do |options|
...
+
+desc "Update the actions.md on https://docs.fastlane.tools"
+desc "This will also automatically submit a pull request to fastlane/docs"
+lane :update_docs do
+  clone_docs do
+    require 'fastlane/documentation/markdown_docs_generator'
+    actions_md_path = File.expand_path("docs/actions.md")
+    Fastlane::MarkdownDocsGenerator.new.generate!(target_path: actions_md_path)
+
+    if `git status -s`.length == 0
```
mfurtak: A minor suggestion, but I think `--porcelain` is the more common way to check this

fastlane/fastlane/Fastfile
```diff
@@ -112,13 +112,76 @@ lane :release do |options|
...
+      end
+    end
+  end
+end
+
+def clone_docs
+  require 'tmpdir'
+  git_url = "https://github.com/fastlane/docs"
+
+  tmp_path = Dir.mktmpdir("fl_clone")
```
KrauseFx: Nice, I knew there was something, but didn't find it immediately. Will update the code 👍

fastlane/fastlane/Fastfile
KrauseFx: Cool, thanks

PR #6491, opened 2016-10-10
[spaceship] Make parse_response and request method public

KrauseFx:
This is useful when external plugins want to extend spaceship without modifying its core


ohayon (approved):
👍

PR #6492, opened 2016-10-10
[spaceship] Version bump

KrauseFx:
- Make `parse_response` and `request` method public, useful for external plugins


ohayon (approved):
👍

Issue #4906, opened 2016-05-30
Screeengrab problem with arab[ar-...] languages 


VictorJava:
### New Issue Checklist

- [x] Updated fastlane to the latest version

- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Android API 16+ is able to revert arabic string. 
You could see this from video (since 0.16 seconds) but doesn't before.

But if I set locale using screengrab - android doesn't rotate(mirror) it(before 0.16 seconds). 
Only if choose arabic language by click.
It's reasonable to admit that 
android.permission.CHANGE_CONFIGURATION 
granted while screengrab works.
I'm using the latest support library 23.4.0.

So, I've tested ar-AE and ar-SA
from
https://github.com/fastlane/fastlane/issues/2310

You can download video with problem here:
http://ifolder.com.ua/5hrevavv20h4.html
or
http://uafile.com.ua/get/168099/

Could you help me?
Thanks.

##### Complete output when running fastlane, including the stack trace and command used
```
It`s without errors so nothing to put. 
```

### Environment

fastlane version (1.91.0):
screengrab version (0.3.2 ):

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?
 - yes and auto on jenkins.
Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?
- no.

vpolouchkine:
Fairly certain that's not supported at the moment.

mfurtak:
Hi @VictorJava - [_screengrab_ 0.5.5](https://github.com/fastlane/fastlane/releases/tag/screengrab%2F0.5.5) was just released with some improvements for RTL language support! :rocket:

If you want to check it out, upgrade your gem with `gem update screengrab` and then make sure to update your _screengrab_ library dependency in your `build.gradle` to reference version **0.5.5**.

PR #6496, opened 2016-10-11
Set layout direction to match Locale on API 17+

efrohnhoefer:
Problem:
Layout isn't mirrored when setting RTL language.

Solution:
Call Configuration#setLayoutDirection(Locale) when changing the configuration.

Fixes: https://github.com/fastlane/fastlane/issues/4906


asfalcone (approved):
🎉 Thanks @efrohnhoefer !!

Issue #6495, opened 2016-10-11
Codesigning "No applicable devices found"


KrauseFx:
Called [gym](https://fastlane.tools/gym) using
```ruby
gym(scheme: "Themoji", export_method: "ad-hoc", export_team_id: "ABC...")
```

Used OS: macOS Sierra
Used Xcode: 8.0

Used plist file

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>method</key>
	<string>ad-hoc</string>
	<key>teamID</key>
	<string>N8X438SEU2</string>
</dict>
</plist>
```

Generated export command (new lines for better readability)
```
/usr/bin/xcrun /Library/Ruby/Gems/2.0.0/gems/gym-1.11.0/lib/assets/wrap_xcodebuild/xcbuild-safe.sh 
-exportArchive 
-exportOptionsPlist '/var/folders/qq/5l_bf23d7_g0h2sv2k66xq_00000gn/T/gym_config20161010-7967-asw94y.plist' 
-archivePath /Users/felixkrause/Library/Developer/Xcode/Archives/2016-10-10/Themoji\ 2016-10-10\ 20.04.52.xcarchive 
-exportPath '/var/folders/qq/5l_bf23d7_g0h2sv2k66xq_00000gn/T/gym_output20161010-7967-y1r6se' 
```

simplified export command

```
/usr/bin/xcrun /Library/Ruby/Gems/2.0.0/gems/gym-1.11.0/lib/assets/wrap_xcodebuild/xcbuild-safe.sh 
-exportArchive 
-exportOptionsPlist '/tmp/my.plist' 
-archivePath /Users/.../something.xcarchive 
-exportPath '/tmp/output' 
```

**Update 1**
The error message is generated by the bundled `ipatool`:

```
2016-10-11 03:05:13 +0000  /Applications/Xcode.app/Contents/Developer/usr/bin/ipatool:15:in `require': cannot load such file -- cfpropertylist (LoadError)
```

so it's related to the third party dependency [CFPropertyList](https://github.com/ckruse/CFPropertyList). Since Xcode doesn't make use of bundler, it seems like the `ipatool` has a problem loading the third party dependency, depending on the Ruby environment.

**Update 2**
[...]

PR #6497, opened 2016-10-11
Fix build errors due to system Ruby dependency

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/6495


djtarazona:
I haven't tried this out, but based on the related issue and these code changes it looks like this may solve my issue of having to `rvm system` when using Fastlane. 👍

KrauseFx:
Thanks @djtarazona, preparing a new release right now https://github.com/fastlane/fastlane/pull/6511

Issue #6501, opened 2016-10-11
[supply] rollout option ingored when promoting to track "rollout"


belawaeckerlig:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I would like to move an APK from the alpha or beta track to the staged rollout track, but the Google API always returns that the user fraction must be greater than 0.0 and less than 1.0.

As far as I could see in the source code (supply/uploader.rb line 39) when passing the "promote_to" option, the `update_track` method is always called with 1.0 for the user fraction. So setting the rollout option has no effect and the default value is not working, because it can not be 1.0

##### Complete output when running fastlane, including the stack trace and command used

```
[bela@localhost common]$ fastlane changeTrack current:alpha target:rollout rollout:0.1 --verbose
DEBUG [2016-10-11 14:55:03.29]: Checking if there are any plugins that should be loaded...
DEBUG [2016-10-11 14:55:03.30]: Using current: alpha
DEBUG [2016-10-11 14:55:03.30]: Using target: rollout
DEBUG [2016-10-11 14:55:03.31]: Using rollout: 0.1
INFO [2016-10-11 14:55:03.32]: -------------------------------------------------
INFO [2016-10-11 14:55:03.32]: --- Step: Verifying required fastlane version ---
INFO [2016-10-11 14:55:03.32]: -------------------------------------------------
INFO [2016-10-11 14:55:03.32]: Your fastlane version 1.105.1 matches the minimum requirement of 1.105.0  ✅
INFO [2016-10-11 14:55:03.32]: ------------------------------
INFO [2016-10-11 14:55:03.32]: --- Step: default_platform ---
INFO [2016-10-11 14:55:03.32]: ------------------------------
INFO [2016-10-11 14:55:03.34]: Driving the lane 'android changeTrack' 🚀
INFO [2016-10-11 14:55:03.34]: --------------------
INFO [2016-10-11 14:55:03.34]: --- Step: supply ---
INFO [2016-10-11 14:55:03.34]: --------------------

+-------------------------+-----------------------------+
|               Summary for supply 0.7.1                |
+-------------------------+-----------------------------+
| track                   | alpha                       |
| track_promote_to        | rollout                     |
| skip_upload_apk         | true                        |
| skip_upload_metadata    | true                        |
| skip_upload_images      | true                        |
[...]

PR #6502, opened 2016-10-11
make use of the rollout option when promoting from a track to the sta…

belawaeckerlig:

Fixes https://github.com/fastlane/fastlane/issues/6501


belawaeckerlig:
Please don't commit until it was discussed in #6501

belawaeckerlig:
@asfalcone PR ready for review.
I changed the description of the rollout option, its default value and the verify block

supply/lib/supply/uploader.rb
```diff
@@ -34,9 +34,9 @@ def perform_upload
 
     def promote_track
       version_codes = client.track_version_codes(Supply.config[:track])
-      client.update_track(Supply.config[:track], 1.0, nil)
+      client.update_track(Supply.config[:track], 0.1, nil)
```
asfalcone: Should this also use `Supply.config[:rollout]` for consistency?

belawaeckerlig: I wasn't sure about that. the value here does not really matter as long as it is between 0,05 and 0.5. But if we put there the value the user provides, it looks like the value matters. So I simply put some fixed value there.
But with a fixed value it is some kind of "magic" number...
So if you want me to change it, no problem. I'm fine with both ;)
I could also add a comment explaining that it does not matter which value is passed there

asfalcone: A comment is definitely important, probably more so than the actual value.

Please add a great comment and I'll get this merged. 👍

belawaeckerlig: @asfalcone sorry for the late reply. I added a comment

asfalcone (approved):

asfalcone:
Thanks @belawaeckerlig! 🚀

PR #6407, opened 2016-10-05
Version bump dependencies on fastlane_core and spaceship

mfurtak:

PR #6503, opened 2016-10-11
Version Bump

mfurtak:
Changes since release 0.5.4:
- Set layout direction to match Locale on API 17+ (#6496)
- Version bump dependencies on fastlane_core and spaceship (#6407)
- Shellescape apk paths (#6405)


ohayon (approved):
👍

PR #6453, opened 2016-10-07
Fix update_project_team with Xcode 8

gcamp:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:
- [x] Run `rspec` for all tools you modified (didn't have any spec)
- [x] Run `rubocop -a` to ensure the code style is valid
- [x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Xcode now update DEVELOPMENT_TEAM in addition to DevelopmentTeam.

Fixes https://github.com/fastlane/fastlane/issues/6452

PR #6504, opened 2016-10-11
Fix update project team

gcamp:
Fix error from https://github.com/fastlane/fastlane/pull/6453 where `params[:teamid]` would not be used anymore.


KrauseFx:
Whops, nice catch 👍

KrauseFx (approved):

PR #6529, opened 2016-10-13
[gym] provision profile specifier

maschall:

I'd like to propose adding support for PROVISIONING_PROFILE (deprecated in Xcode 8) and PROVISIONING_PROFILE_SPECIFIER (added in Xcode 8).

I believe this will help many code signing issues, but I realize it does go against the fastlane recommended guidelines for code signing.

With these changes, reverting #6347, and removing the deprecated :provisioning_profile_path warning for Xcode 7 package generator the following would happen:

```
lane :beta do
   sigh
   gym
end
```

The provision profile selected by `sigh` would be passed to `gym`, which would explicitly use the UUID (in Xcode 7 and before) or the Name (in Xcode 8 and above).  This guarantees that the more obvious provision profile is used for `gym`, without requiring changes to the Xcode Project files.

This should also work with `match`, but I haven't tested it yet, as our project currently does not implement it.

I am happy to write more tests.

Without the changes described above related to reverting and removing deprecation messages, this PR is still valid and allows developers to do the following and get the same effect:

```
lane :beta do
   sigh
   gym( provisioning_profile_uuid: ENV["SIGH_UDID"] )
end
```

I do recognize that this is currently possible with `xcargs` parameter for `gym`, but I believe it would benefit the community and help alleviate the code signing hydra

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -37,7 +37,7 @@ module FastlaneCore
       def uuid(path)
         parse(path).fetch("UUID")
       end
-      
+
       # @return [String] The Name of the given provisioning profile
       def name(path)
         parse(path).fetch("Name")
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -37,6 +37,11 @@ module FastlaneCore
       def uuid(path)
         parse(path).fetch("UUID")
       end
+      
+      # @return [String] The Name of the given provisioning profile
+      def name(path)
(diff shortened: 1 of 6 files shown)


KrauseFx:
Hey @maschall, thanks for preparing this pull request. I think we've received similar proposals in the past. My main question/concern is: how does this change handle apps with multiple targets? Every day, more and more apps have multiple targets, and it doesn't make a lot of sense to focus on one-target apps anymore

maschall:
@KrauseFx you are absolutely right.  I'm so used to one target schemes, and our project is just ramping up a watch app and extension.

KrauseFx:
Cool, thanks for understanding and looking into this. That's the one and only reason why we try to "force" people to use the multi-target approach, as more and more apps will eventually support at least a second target.

PR #6536, opened 2016-10-13
Prompt for json_key if it and legacy option are both missing

mfurtak:
Previously an error message was displayed, but we would prefer that the user is prompted for a value, since one is required


asfalcone (approved):
It would be nice to be able to completely deprecate `key` and `issuer` so that we could use the build in prompt from `configuration.rb` but this will do for now.

PR #6540, opened 2016-10-14
Add option to clear existing screenshots

giginet:

Formerly, I had been uploaded screenshots of all available sizes via deliver.
Thanks to recent iTunes Connect updates, we have only to upload the largest size.
So I dropped other sizes and submitted via deliver. But these are not removed and old ones published onto App Store.

So I added the new option for deliver to remove all existing screenshots forcibly.

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -159,6 +159,11 @@ module Deliver
                                      env_name: "DELIVER_ITC_PROVIDER",
                                      description: "The provider short name to be used with the iTMSTransporter to identify your team",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :overwrite_screenshots,
+                                     env_name: "DELIVER_OVERWRITE_SCREENSHOTS",
+                                     description: "Clear all previously uploaded screenshots",
+                                     is_string: false,
+                                     default_value: false),
 
         # App Metadata
         # Non Localised
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -161,7 +161,7 @@ module Deliver
                                      optional: true),
         FastlaneCore::ConfigItem.new(key: :overwrite_screenshots,
                                      env_name: "DELIVER_OVERWRITE_SCREENSHOTS",
-                                     description: "Clear all previously uploaded screenshots",
+                                     description: "Clear all previously uploaded screenshots before uploading the new ones",
                                      is_string: false,
                                      default_value: false),
 
--- a/deliver/lib/deliver/upload_screenshots.rb
+++ b/deliver/lib/deliver/upload_screenshots.rb
@@ -10,20 +10,21 @@ module Deliver
       UI.user_error!("Could not find a version to edit for app '#{app.name}'") unless v
 
       UI.message("Starting with the upload of screenshots...")
+      screenshots_per_language = screenshots.group_by(&:language)
 
-      # First, clear all previously uploaded screenshots, but only where we have new ones
-      # screenshots.each do |screenshot|
-      #   to_remove = v.screenshots[screenshot.language].find_all do |current|
-      #     current.device_type == screenshot.device_type
-      #   end
-      #   to_remove.each { |t| t.reset! }
-      # end
-      # This part is not working yet...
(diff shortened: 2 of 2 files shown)


KrauseFx:
I'm not sure it's a great solution to add yet another parameter to _deliver_. Could you provide some more use cases on why this is useful? 👍

hjanuschka:
related to an jurassic issue ? https://github.com/fastlane/fastlane/issues/1529
i love the functionality but i personally would love to see screenshots deleted by default (as the # commented code did back then.) - and only the ones on disk should get uploaded

KrauseFx:
We've had multiple people complain that _deliver_ removed screenshots and they didn't have a copy anywhere, so they lost the screenshots. That's why now _deliver_ only deletes the old screenshots when new ones are provided.

hjanuschka:
good point you r right about it - its really frustrating to loose them.

deliver/lib/deliver/upload_screenshots.rb
```diff
@@ -10,20 +10,21 @@ def upload(options, screenshots)
...
-      #   end
-      #   to_remove.each { |t| t.reset! }
-      # end
-      # This part is not working yet...
+      if options[:overwrite_screenshots]
+        UI.message("Removing all previously uploaded screenshots...")
+        # First, clear all previously uploaded screenshots
+        screenshots_per_language.keys.each do |language|
+          v.screenshots[language].each_with_index do |t, index|
+            v.upload_screenshot!(nil, index, t.language, t.device_type)
```
KrauseFx: I haven't tested this yet, just setting `nil` will delete the screenshot?

deliver/lib/deliver/options.rb
```diff
@@ -159,6 +159,11 @@ def self.available_options
                                      env_name: "DELIVER_ITC_PROVIDER",
                                      description: "The provider short name to be used with the iTMSTransporter to identify your team",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :overwrite_screenshots,
+                                     env_name: "DELIVER_OVERWRITE_SCREENSHOTS",
+                                     description: "Clear all previously uploaded screenshots",
```
KrauseFx: Let's use `Clear all previously uploaded screenshots before uploading the new ones`

KrauseFx (changes requested):

KrauseFx:
Hey @giginet, putting some more thought into this, I think the feature makes sense, and by having to explicitly setting a flag, it's very clear that screenshots get removed, and therefore it won't happen that someone deletes their screenshots by mistake.

I added one request for change, and one question before it's ready to be merged 👍 

Thanks for your contribution

giginet:
OK, I'll add requested changes soon.
Sorry for late reply.

KrauseFx:
Thanks @giginet, no need to say sorry, it's just been a few hours 👍  I appreciate your help

deliver/lib/deliver/upload_screenshots.rb
giginet: Yes. I tried this method following ways.

1. Add 4.7 and 5.5 inch screenshots on Finder
2. Run deliver
3. Remove the 4.7 inch screenshot
4. Run deliver with `overwrite_screenshots` option
5. the 4.7 inch screenshot was removed on iTunes Connect

Then it seems work well.

giginet:
I've fixed a few days ago. However CI is still failing. Could you fix it?

KrauseFx:
There is merge conflicts, could you resolve them?

nolanwilson:
@KrauseFx I see that your latest comment mentioned merge conflicts. I tried a merge with the fastlane master and [was able to do so without conflict](https://github.com/nolanwilson/fastlane/tree/overwrite-screenshots).

Is there action still required for [this task](https://github.com/fastlane/fastlane/pull/6540#pullrequestreview-5517693)?

Would love to see this get in soon since we have several apps per day that get submitted with fewer screenshots than the previous version in the store and the remaining screenshots are still sticking around until this gets merged in :)

KrauseFx (approved):

KrauseFx:
This looks great, thanks everyone involved. The change will be available in the next release  🚀

giginet:
:tada:

PR #6114, opened 2016-09-14
Pass configuration to showBuildSettings command

maschall:
#6113

PR #6542, opened 2016-10-14
Use configuration argument if present when detecting build settings

### LINKED ISSUE #6514: FastlaneCore::Project->build_settings() not using -configuration parameter
### New Issue Checklist
- [ x] Updated fastlane to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

Scan::detect_values.rb->detect_simulator() uses FastlaneCore::Project->build_settings()

which is using command:
xcodebuild -showBuildingSettings -workspace -scheme -project

The issue is that -configuration parameter is not used to list the build settings and it will use the default configuration. This may cause unexpected return if a project has multiple scheme and configuration. e.g different deployment target for different configuration.

fastlane version (run `fastlane -v`):

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?

tadelv:
- [] Run `rspec` for all tools you modified
- [] Run `rubocop -a` to ensure the code style is valid
- [] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

🔑
I'm trying to fix #6514, because it's low hanging fruit.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -195,6 +195,7 @@ module FastlaneCore
       proj << "-workspace #{options[:workspace].shellescape}" if options[:workspace]
       proj << "-scheme #{options[:scheme].shellescape}" if options[:scheme]
       proj << "-project #{options[:project].shellescape}" if options[:project]
+      proj << "-configuration #{options[:configuration].shellescape}" if options[:configuration]
 
       return proj
     end


KrauseFx:
I don't know enough about this, will this have any side effects we should think about? Thanks for submitting the PR 👍

chenxiao0228:
@KrauseFx I raised the issue. This change corrects the function. Because if configuration is not specified, Xcode will pick the default one. And the default configuration may not be the one bind to the scheme. I learned it the hard way :)

milch:
@tadelv Thanks for your contribution! 

However, I will have to close this, as there is already an earlier PR that solves this problem: #6114.

PR #6547, opened 2016-10-14
Remove call to verify_docs during test_all to keep CI speedy

mfurtak:


KrauseFx (approved):
Perfect

KrauseFx:
To be honest, I think it's fair to just remove this from here, and run it only on the actual deployment. That's where we want to be sure the documentation is up to date with the actual fastlane code

Issue #6531, opened 2016-10-13
[gym] for_lane support in Gymfile


maschall:
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

I work on a project with multiple schemes and configurations.  I'd like to be able to specify a configuration for gym in specific lane.

Example:

```
lane :crashlytics do
   sigh
   gym
end

lane :appstore do
   sigh
   gym
end
```

Appfile:

```
for_platform :ios do
  apple_id #{appleId}

  for_lane :crashlytics do
    app_identifier ENV['CRASHLYTICS_APP_ID']
    team_id ENV['CRASHLYTICS_APP_ID_TEAM_ID']
  end

  for_lane :appstore do
    app_identifier ENV['APPSTORE_APP_ID']
    team_id ENV['APPSTORE_TEAM_ID']
  end
end
```

Gymfile:

```
for_platform :ios do
  apple_id #{appleId}

  scheme ENV['SCHEME'

  for_lane :crashlytics do
     configuration "Crashlytics"  
  end

  for_lane :appstore do
    configuration "Release"
  end
end
```

I think this can apply to `Scanfile`s as well, but there could also be a better solution to have `configuration` set-able `for_lane` in another fashion. (Maybe using the `lane_context`?)

KrauseFx:
This looks good, however I think this requires a lot of changes in various code parts, which may lead to unexpected side effects. Did you look into how it's possible to change this?

PR #6549, opened 2016-10-14
for_lane support all configuration files

maschall:
#6531

I'm not sure a good reason for the silent ignoring duplicate values in config files, but if you review AppFile, the example given is no longer valid, you can overwrite the same value with duplicate lines.

Also I started with the tests for Gymfile but realized quickly that the configuration effects Dangerfile, Matchfile, Scanfile, etc...

I can help update docs if this looks good


KrauseFx:
Oh wow, thanks for the pull request, I somehow completely missed that notification. Definitely something we should look into.

fastlane_core/spec/configuration_file_spec.rb
```diff
@@ -64,12 +64,6 @@
         expect(config[:another_boolean]).to be(true)
       end
 
-      it "ignores the same item being set after the first time" do
```
KrauseFx: By changing this, would this break setups where you have the same line in one config file twice, and you are used to them being ignored? Is there a way we can leave that test and make it pass, while still adding this new feature?

fastlane_core/spec/fixtures/ConfigFileForLane
```diff
@@ -0,0 +1,13 @@
...
+  app_identifier "com.forlane.enterprise"
+end
+
+for_platform "ios" do
+  app_identifier "com.forplatform.ios"
+  
+  for_lane "release" do
+    app_identifier "com.forplatformios.forlanerelease"
+  end
+end
```
KrauseFx: Can you add a new line here

KrauseFx (changes requested):
Great tests, great change, I love it 👍 

Please add some documentation somewhere, probably in Advanced.md for now, and we're good to go. 

Sorry it took so long, and thanks again for working on this :rocket:

fastlane_core/spec/configuration_file_spec.rb
maschall: I was wondering about this, I'll take another stab at it, but I believe (and may be wrong) is that Appfile is not behaving in this same manner, and that was why I looked at removing this.

I'll see what I can do to get this back in there.

asfalcone:
@maschall It looks like there are a few changes still needed for this PR.

fastlane_core/spec/fixtures/ConfigFileForLane
mfurtak: Done

mfurtak:
Since it seems like we were still pretty keen on getting this merged, I did some more work on it in #7859 to reconcile what I perceived to be the blocking problems here.

mfurtak:
Carried forward and merged as part of #7859 👍

PR #6080, opened 2016-09-12
Match: support for supplying multiple app identifiers at once

jverdi:
🔑 Allows you to supply multiple app identifiers to match at once and it will only perform (1) clone of your repo, and (1) check for your associated certificate.

As an action:

```
match(type: "appstore", 
      app_identifier: ["tools.fastlane.app", "tools.fastlane.app2", "tools.fastlane.app3"])
```

I made sure to also continue supporting the old syntax, so this doesn't break existing configurations:

```
match(type: "appstore", app_identifier: "tools.fastlane.app")
```

Multiple apps via command-line:

```
match appstore -a tools.fastlane.app,tools.fastlane.app2,tools.fastlane.app3
```

This addresses performance issues discussed in #4100

I don't see the need for a full matrix-like solution. Generally speaking, CI builds will be building for one environment at a time. The real pain point has been the proliferation of app ids for extensions. A matrix also doesn't translate well for command-line usage.

PR #6554, opened 2016-10-14
Match: support for supplying multiple app identifiers at once

hjanuschka:
this  is a rebased version of https://github.com/fastlane/fastlane/pull/6080 - as there is no activity for many days. - original commit is preserved.
i would like to see this feature in match.


hjanuschka:
ok closed this - as i saw the original  PR just has not yet been reviewed - sorry

PR #6555, opened 2016-10-14
[fastlane_core] Add an `os_type` property to the `Device` class

0xced:

---

This will be required for an upcoming change to `scan` for handling projects supporting multiple platforms with a single scheme as described in http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/

### DIFF
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -40,7 +40,7 @@ module FastlaneCore
             end
 
             if match && !match[4] && (os_type == requested_os_type || requested_os_type == "")
-              @devices << Device.new(name: match[1], os_version: os_version, udid: match[2], state: match[3], is_simulator: true)
+              @devices << Device.new(name: match[1], os_type: os_type, os_version: os_version, udid: match[2], state: match[3], is_simulator: true)
             end
           end
         end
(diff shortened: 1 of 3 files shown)


KrauseFx (approved):
This looks good to me, great addition to _fastlane_core_. I'd love to have a second review before we merge this 👍

0xced:
I was very impressed that adding a new parameter in the middle of the `Device` constructor did not break dependencies (such as scan) which don't know about that parameter!

Or I did not actually test scan with my local copy of fastlane_core. 😄

0xced:
I just rebased, I hope it can be part of fastlane_core 0.52.4. ☺️

asfalcone (approved):

mfurtak:
Do I understand correctly that this parameter doesn't actually reflect the OS of the devices we find, but rather, the OS of the devices we ask for? Meaning, if we don't specify a `requested_os_type`, then all `Device`s come back with a `os_type` of `''`?

0xced:
Actually, the `connected_devices(requested_os_type)` function returns early if `requested_os_type` is not `iOS` or `tvOS`, so the `os_type` attribute will always be valid.

mfurtak:
Cool. Is the same true for `simulators`?

0xced:
For the simulators, `os_type` (the one parsed from `xcrun simctl list`) is used instead of `requested_os_type`, so it will also be always valid.

mfurtak:
Just got around to pulling the branch and giving it a try. It seems great - nice job! 👍

PR #6561, opened 2016-10-15
add support for provisioning_profile_specifier

hjanuschka:
this one adds the ability to set the new `PROVISIONING_PROFILE_SPECIFIER`

quick test:

```
bundle exec fastlane run update_project_provisioning provisioning_profile_specifier:asda xcodeproj:/Users/hja/Desktop/fastlane/fastlane-plugin-update_project_codesigning/demo-project/demo/demo.xcodeproj
```

### DIFF
--- a/fastlane/lib/fastlane/actions/update_project_provisioning.rb
+++ b/fastlane/lib/fastlane/actions/update_project_provisioning.rb
@@ -12,36 +12,36 @@ module Fastlane
 
         # assign folder from parameter or search for xcodeproj file
         folder = params[:xcodeproj] || Dir["*.xcodeproj"].first
+        unless params[:provisioning_profile_specifier]
+          # validate folder
+          project_file_path = File.join(folder, "project.pbxproj")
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/update_project_provisioning.rb
```diff
@@ -108,6 +111,10 @@ def self.available_options
                                        env_name: "FL_PROJECT_PROVISIONING_PROFILE_TARGET_FILTER",
                                        description: "A filter for the target name. Use a standard regex",
                                        optional: true),
+          FastlaneCore::ConfigItem.new(key: :provisioning_profile_specifier,
+                                       env_name: "FL_PROJECT_PROVISIONING_PROFILE_SPECIFIER",
+                                       description: "A filter for the target name. Use a standard regex",
```
hjanuschka: you are right, many thx 👍

milch:
@hjanuschka Thanks for your contribution! 🚀 

However, since `update_project_provisioning` is effectively deprecated, we don't want to add any new features to it. I'm going to close this now. If you feel that some users would benefit from this change, you can always create a plugin instead of changing this action!

PR #6566, opened 2016-10-16
add aws_device_farm to plugin list

hjanuschka:
let me know if we can add it to the list already.
🤒  i even got the "iOS" right.


KrauseFx (approved):

KrauseFx:
iOS, yeah!  🚀 

There is a script to automatically generate this file https://github.com/fastlane/fastlane/blob/master/fastlane/Rakefile#L16

Issue #481, opened 2015-08-07
[Feature] Add overlay image on icons action


marcelofabri:
I created an action to apply overlay images on icons as a part of our build process. 

The result can be seen here:

![](https://www.dropbox.com/s/ar1wcm49s7o7tlz/playkids-icons.png?dl=1)

I posted the action on a Gist (https://gist.github.com/marcelofabri/ac3884d6170ee6414257) to gather feedback before creating a PR. It's not very generic yet because we don't need it now.

KrauseFx:
This looks nice :+1: 

Some notes:
- You can remove `ADD_ICON_OVERLAY_CUSTOM_VALUE`
- `mini_magick` should be added as dependency for `fastlane` in the `fastlane.gemspec`
- If the path to `appiconset_path` is wrong, this will modify and replace all the `png` files there, right? 

How do you use it? Do you store the resulting image in git? Or do you use this action and then do a `git reset` afterwards?

marcelofabri:
> If the path to appiconset_path is wrong, this will modify and replace all the png files there, right?
Yep. Any ideas to make it better?

> How do you use it? Do you store the resulting image in git? Or do you use this action and then do a git reset afterwards?
I don't store them, doing a git reset after. This is because each lane can use a different overlay, so it doesn't make sense for us to commit. Only the "clean" version is versioned.

iainsmith:
I was hoping someone had already built this. 😀

It can also be very helpful to add either the first few characters of the sha-1 or the build number. That way everyone can easily see exactly what build they have.

coreyfloyd:
@marcelofabri Awesome work!

As for improvements, I agree with @iainsmith - It would be great to pass arbitrary text and maybe a color so we can customize the overlay at build time.

I currently have something like this in my Xcode build phases. We use image magic to overlay the build number and the build type (beta/alpha/etc…). 

It looks something like this:

```
convert -background $overlay_color -fill white -gravity center -size ${overlay_width}x${overlay_height}\
caption:"${CONFIGURATION}\n(${build})"\
"${source_icon_path}" +swap -gravity south -composite "${target_icon_path}"
```

KrauseFx:
@coreyfloyd agreed, instead of providing multiple images, passing a color + text and maybe a font should be the way to go :+1:

vinnybad:
[...]

PR #6568, opened 2016-10-16
Add check for ImageMagick before running the badge action

AliSoftware:
This allows to check for the presence of ImageMagick before trying to run the `badge` action, which requires ImageMagick to be installed in order to work properly without erroring.

This intends to implement what has been discussed in https://github.com/fastlane/fastlane/issues/481#issuecomment-147425621 which seems not to have been implemented in the end. This has also been discussed with @KrauseFx directly in the [Fastlane Slack](https://fastlane-tools.slack.com/archives/general/p1476433989000193).

---
### Specs

I'm not sure how to write specs for this, as we'd have to fake various binary installations to check that properly. Besides, it seems that other helpers (like `git_helper.rb` etc) don't have specs either, so I'm not sure you're expecting them for helpers anyway?


mfurtak:
Hi @AliSoftware - thanks for the contribution! 👍 

Is there a reason that we need to require that homebrew is installed? Shouldn't we only need to verify that ImageMagick is installed?

AliSoftware:
@mfurtak there isn't a real reason to only check if ImageMagick is installed via homebrew rather than in any other method. I just don't really know what feature (lib? Binary? Other?) `badge` specifically needs from ImageMagick and how it uses or invoke it to know exactly what to check instead of it being installed by homebrew.

Besides, the author of the action seem themselves to suggest to install ImageMagick via homebrew (and it's the official way to install it at least for what the action needs) so I figured it should be the best way to check, but if you know a better way I'd love to adapt the PR.


The main purpose of this PR is also, in case ImageMagick isn't actually installed on the mac, to actually give the proper instructions to install it (at least in the form requested/used by the `badge` action), so in case of error, we'll mention `brew` anyway in those installation instructions.

But if you have a better way to suggest for the ImageMagick detection we call still change the detection code while still suggestion to install/use `brew` only in case ImageMagick is needed?

mfurtak:
Cool. While I like the precision from using `brew` to make the check, I think it is better to recognize that they have ImageMagick even if they don't have Homebrew. 👍 

The `convert` binary is at the core of ImageMagick so checking for that [like _frameit_ does](https://github.com/fastlane/fastlane/issues/481#issuecomment-147425621) should be fine.

Suggesting to install via Homebrew is good for the most part, but we should also be aware that some folks run _fastlane_ on non-Mac machines, and telling them about `brew` won't help. :smile:

If you want to try adding some tests, one way to mock the result from shelling-out can be seen [here](https://github.com/fastlane/fastlane/blob/master/fastlane_core/spec/project_spec.rb#L382). That intercepts the call to the backticks, but I suspect you want something closer to:

```ruby
expect(Fastlane::Actions).to receive(:`).and_return('')
```

AliSoftware:
@mfurtak I've refactored the check to now check for the `convert` binary and use the same check and error message as `frameit` does.

Let me know if that better suits you. If so, I might try to add the associated specs for this new code!

mfurtak:
Looks good! 🙌

AliSoftware:
I'm afraid the only time window I had was this weekend and today and won't have time to write specs for that until not next weekend (busy) but the one after…

So probably best to merge this now, and I'll reopen a PR later to write specs when I have more free time for that 👍

mfurtak (approved):

mfurtak:
Great, thanks for the improvement!

Issue #6570, opened 2016-10-16
[Internal] Pin rubocop version to avoid inconsistencies between machines


AliSoftware:
In order to avoid inconsistencies when running `rubocop -a` from different machines (especially a contributor's Mac vs. the CI) we should ensure everybody uses the same version of `rubocop`

### What Happened

* Earlier, before submitting a new PR, I ran `rubocop -a` locally on my code changes, and there were no offenses detected.
* But then when I submitted the PR and Circle-CI started checking the code, it found 2 offenses on the remote machine
* If though the reason was because my `rubocop` installation might be quite old and too different from the one used by Circle-CI, so I ran `gem update rubocop` then ran `rubocop -a` locally again and this time it found… 634 offenses (mostly not in my code but elsewhere, mainly "Block has too many lines" offenses but not only)

### Suggestion

I think the behavior difference is clearly because of a difference in `rubocop` versions between the CI and contributor's local machines.

**The easy solution for that would be pin the rubocop version, probably in the `Gemfile`**

```ruby
group :test do
  gem 'rubocop', '<version_requirement>'
end
```
 
### Environment

After my latest `rubocop` update, I'm currently using `rubocop 0.44.1` on my Mac.

_I couldn't find the version of `rubocop` on the CI in Circle-CI's logs though to compare_

KrauseFx:
Hey @AliSoftware, thanks for submitting and describing this issue. I agree the version has to be locked, and I thought we did so with https://github.com/fastlane/fastlane/blob/master/fastlane/fastlane.gemspec#L73.
Did you run `rubocop` using `bundle exec rubocop`? 

However we should also update to the latest version of `rubocop`, I missed their new releases.

AliSoftware:
Ah maybe I forgot to `be`, was late; might explain 😉

AliSoftware:
Actually now that I think about it I'm pretty sure I didn't use `bundle exec` as I usually don't use the `-a` flag myself on other ruby projects, so if I did us it this time it's because I followed the instructions in `CONTRIBUTING.md` to the letter… and they don't mention `bundler` for that step I believe  😜

So _that_ might be the core problem to address then 😉

AliSoftware:
Closing this issue as the version is indeed already pinned (see also #6573) and the problem was in fact the incomplete `PULL_REQUEST_TEMPLATE.md` instructions — which are gonna be fixed by my PR #6576

KrauseFx:
[...]

PR #6576, opened 2016-10-17
[PR Template] Add 'bundle exec' to instructions

AliSoftware:
This is to ensure people use `bundle exec` when running `rspec` and `rubocop` when validating their PR, to be sure everybody use the same version.

Related to https://github.com/fastlane/fastlane/issues/6570#issuecomment-254134586


KrauseFx:
This only works when the user is using bundler though, which not everybody does

AliSoftware:
@KrauseFx But if people are not using `bundler`, how do they ensure they use the proper version of `rspec`, `rubocop`, and all, to ensure the checks are meaningful?

It seemed to me that the whole development process in fastlane kinda required `bundler` anyway, as even the `Rakefile` and `rake test_all` requires `bundler` anyway?

KrauseFx (approved):

KrauseFx:
Good point, thanks 👍

PR #6584, opened 2016-10-17
[Gym] Add reminder of the full log path when erroring

AliSoftware:
## The problem

Lately I have had multiple cases of build failed where the `xcpretty` output wasn't really helpful as it filtered out some important info
- Last time I encountered this, `xcpretty` filtered out some code signing errors, making it hard to figure out what was wrong (it has since been fixed by @mgrebenets in supermarin/xcpretty#247)
- Today I got an "Interaction is not allowed" error when driving our build via a headless CI, but was only able to see that error and understand it was the cause by reading the unfiltered log as well

Reading the full log unfiltered by `xcpretty` instead of the filtered stdout helped me a lot figuring out what the real problem was, so I think it's useful to mention it more clearly in case of failure.
## Solution

This PR aims at reminding the path where you can read the full log (not filtered by `xcpretty`) at the end of a `gym` failure, so users are incited to read it to get more info about the error, and have a quick way to find its path.

gym_failure

### DIFF
--- a/gym/lib/gym/error_handler.rb
+++ b/gym/lib/gym/error_handler.rb
@@ -6,7 +6,7 @@ module Gym
       # @param [String] The output of the errored build
       # This method should raise an exception in any case, as the return code indicated a failed build
       def handle_build_error(output)
-        # The order of the handling below is import
+        # The order of the handling below is important
         case output
         when /Your build settings specify a provisioning profile with the UUID/
           print "Invalid code signing settings"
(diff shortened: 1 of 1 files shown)


asfalcone (approved):

asfalcone:
Thanks @AliSoftware 🚀

KrauseFx:
What a great idea!!! 👍

Issue #6553, opened 2016-10-14
Connection reset by peer 


thecb4:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

##### Complete output when running fastlane, including the stack trace and command used

```
+-----------------------+-----------------------------------------------+
|                        Summary for match 0.8.1                        |
+-----------------------+-----------------------------------------------+
| git_url               | git@gitlab.com:thecb4/XXXXXX.git |
| app_identifier        | io.thecb4.XXXXXX                |
| type                  | appstore                                      |
| verbose               | true                                          |
| git_branch            | master                                        |
| username              | cavelle@XXXXX                            |
| keychain_name         | login.keychain                                |
| readonly              | false                                         |
| team_id               | XXXXXX                                    |
| force                 | false                                         |
| shallow_clone         | false                                         |
| force_for_new_devices | false                                         |
| skip_docs             | false                                         |
+-----------------------+-----------------------------------------------+
```

```
INFO [2016-10-14 16:52:47.88]: Cloning remote git repo...
INFO [2016-10-14 16:52:47.89]: $ GIT_TERMINAL_PROMPT=0 git clone 'git@gitlab.com:thecb4/FastlaneMatchTheCB4.git' '/var/folders/yh/rbjzzgk50b7f5g8pnyrntf8r0000gn/T/d20161014-9928-1ycj0u6'
INFO [2016-10-14 16:52:47.91]: ▸ Cloning into '/var/folders/yh/rbjzzgk50b7f5g8pnyrntf8r0000gn/T/d20161014-9928-1ycj0u6'...
INFO [2016-10-14 16:52:49.50]: ▸ remote: Counting objects: 57, done.
INFO [2016-10-14 16:52:49.69]: ▸ remote: Compressing objects: 100% (53/53), done.
INFO [2016-10-14 16:52:49.84]: ▸ remote: Total 57 (delta 4), reused 0 (delta 0)
INFO [2016-10-14 16:52:49.84]: ▸ Receiving objects: 100% (57/57), 102.94 KiB | 0 bytes/s, done.
INFO [2016-10-14 16:52:49.84]: ▸ Resolving deltas: 100% (4/4), done.
INFO [2016-10-14 16:52:49.85]: ▸ Checking connectivity... done.
[...]

PR #6593, opened 2016-10-17
Update error handling for recent TLS 1.2 problems talking to Apple servers

mfurtak:
Help users towards the solution for #6553 by providing better error messaging when Ruby can't make an SSL connection to Apple's servers.

I needed to move the local Faraday error class creation outside of the `Commander` module in order to have the correct shadowing happen if/when `Faraday` is later defined.

I knew I was on the right track with that when Ruby complained that the `ConnectionFailed` class supertypes did not match (I originally started with `StandardError`, like in the previous code). [Looking at the Faraday code](https://github.com/lostisland/faraday/blob/a712938071bb0997410b5de4a4f2517d30fc4d3d/lib/faraday/error.rb), those extra super classes are now also defined locally so that the class hierarchies can match correctly.

image

### DIFF
--- a/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
+++ b/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
@@ -1,17 +1,19 @@
+unless Object.const_defined?("Faraday")
+  # We create these empty error classes if we didn't require Faraday
+  # so that we can use it in the rescue block below even if we didn't
+  # require Faraday or didn't use it
+  module Faraday
+    class Error < StandardError; end
+    class ClientError < Error; end
+    class SSLError < ClientError; end
+    class ConnectionFailed < ClientError; end
+  end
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
```diff
@@ -1,17 +1,20 @@
+unless Object.const_defined?("Faraday")
+  puts '**** defining empty classes ****'
```
KrauseFx: Debugging code?

KrauseFx (approved):

jogu:
This kind of makes an assumption that people are using homebrew :-)

In case anyone else is already using MacPorts, here are the commands to get this working:

```
sudo port install ruby23
sudo port select --set ruby ruby23
sudo gem install fastlane
```

Issue #6362, opened 2016-10-01
Frameit is unable to use iPhone 6S frames


ayushgoel:
- [x] Updated fastlane to the latest version
Yes, using `fastlane run update_fastlane`.

- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

1. iPhone 7 device frames are not being used currently.
2. `frameit` even though able to find the device frame for iPhone 6S, is not using it.

##### Complete output when running fastlane, including the stack trace and command used

```
DEBUG [2016-10-01 17:50:32.62]: Looking for iPhone-6s-Plus-Silver-vertical and found 2
DEBUG [2016-10-01 17:50:33.10]: Looking for iPhone-6s-Plus-Silver-vertical and found 2
DEBUG [2016-10-01 17:50:33.10]: Parsing config file '.../fastlane/screenshots/Framefile.json'
DEBUG [2016-10-01 17:50:37.03]: Falling back to default text as there was nothing specified in the .strings file
ERROR [2016-10-01 17:50:37.04]: `Helper.gem_path` is deprecated. Use the `ROOT` constant from the appropriate tool module instead.
DEBUG [2016-10-01 17:50:37.66]: Using .../fastlane/screenshots/OpenSans.ttf as font the title of ./fastlane/screenshots/en-US/1.png
DEBUG [2016-10-01 17:50:37.66]: Adding text 'TEXT'
INFO [2016-10-01 17:50:42.64]: Added frame: '.../fastlane/screenshots/en-US/1_framed.png'
```

### Configuration Files

Framefile:
```ruby
{
  "default": {
    "keyword": {
      "font": "OpenSans.ttf"
    },
    "title": {
      "fonts": [
        {
          "font": "OpenSans.ttf",
          "supported": ["en-US"]
        }
      ],
      "color": "#545454"
    },
    "background": "./red-background.jpg",
    "padding": 50
  },
  "data": [
    {
      "filter": "LandingScreen",
      "keyword": {
        "color": "#d21559"
      }
    }
  ]
}
```

### Environment

fastlane version (run `fastlane -v`):

* fastlane 1.104.0
* frameit 2.7.0

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?
No

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?
Yes, rvm
```
ruby --version
[...]

Issue #5306, opened 2016-07-02
frameit: Apple renamed images


ogezue:
I think apple renamed the images of their devices that frameit need. So frameit needs to be updated.

e.g. Missing file: 'iPad-Pro-Space-Gray-vertical.png'

PR #6597, opened 2016-10-17
Re-write frameit to support new device frames

KrauseFx:
This PR will allow the remote distribution of new device frames, including their offsets, making _frameit_ much more stable and easier to maintain

This will fix multiple issues, including
- https://github.com/fastlane/fastlane/issues/6362
- https://github.com/fastlane/fastlane/issues/5306

This will also add support for Rose Gold, Gold, iPad Pro and more 👍 

---
- Rewrote the whole framing process, so that you don't have to do anything manually any more
- _frameit_ now automatically downloads and installs all the frames the first time you use it
- Support for Rose Gold, Gold and the new Matte Black
- Support for  the new iPhone 7
- Full support for portrait and landscape mode
- You can lock a specific version of frames in your `Framefile.json` if you want to, to have consistent builds in the future

**Rose Gold**
screenshot 2016-10-21 15 11 23
**Gold**
screenshot 2016-10-21 15 14 30


SRowley90:
I'm not too familiar with Ruby but have got the offsets for the iPad Air 2, iPad Pro and the 2 iPhone 7 devices based of the screenshots that are downloaded, i'm not sure how much helkp this is to you but this is my offsets.json file:

> {
  "portrait": {
    "iPhone SE": {
      "offset": "+63+235",
      "width": 641
    },
    "iPhone 5": {
      "offset": "+63+235",
      "width": 641
    },
    "iPhone 6s": {
      "offset": "+60+225",
      "width": 752
    },
    "iPhone 6s Plus": {
      "offset": "+102+378",
      "width": 1250
    },
    "iPhone 7": {
      "offset": "+62+206",
      "width": 750
    },
    "iPhone 7 Plus": {
      "offset": "+106+345",
      "width": 1242
    },
    "iPad Pro": {
      "offset": "+120+217",
      "width": 2048
    },
    "iPad Air 2": {
      "offset": "+113+225",
      "width": 1536
    }
  }
}

frameit/lib/frameit/device_types.rb
```diff
@@ -1,11 +1,13 @@
 module Frameit
   module Color
-    BLACK = "SpaceGray"
-    SILVER = "Slvr"
+    BLACK = "Space Gray"
```
tadelv: Just a quick note: the iPhone 7 has only 'Black' listed as color, there is no more Space Gray

KrauseFx: Nice catch, thanks 👍

KrauseFx: Currently frameit doesn't support iPhone 7 anyway, right? It will just use the iPhone 6 assets

KrauseFx: @tadelv I'm still thinking: how would you want to select if you want an iPhone 7 or an iPhone 6? The screenshots are the same size

KrauseFx:
This is ready for a review 👍

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,76 @@
...
+  puts "Move the content of the zip file into './raw'"
+  FileUtils.mkdir("raw") unless File.directory?("raw")
+  system("open ./raw")
+  confirm
+
+  # Let's find all the image resources we need and copy it into the correct directory
+  current_time = Time.now.to_i.to_s
+  output = File.join("output", current_time)
+  sh "rm -rf output"
+  FileUtils.mkdir_p(output) unless File.directory?(output)
```
mfurtak: Rake has `rm_rf` and `mkdir_p` helper methods built-in. 👍

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,76 @@
...
+]
+
+task :generate_device_frames do
+  puts "Download and extract the latest devices from Facebook"
+  puts ""
+  puts "  http://facebook.design/devices"
+  puts ""
+  confirm
+  puts "Move the content of the zip file into './raw'"
+  FileUtils.mkdir("raw") unless File.directory?("raw")
```
mfurtak: Rake has `mkdir` as a built-in helper method 👍

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,76 @@
...
+  File.write(File.join(output, "files.json"), image_assets.to_json)
+  puts "Wrote names of available files to files.json"
+
+  File.write(File.join(output, "version.txt"), current_time)
+  puts "Wrote the current time stamp to version.txt"
+
+  FileUtils.cp("offsets.json", File.join(output, "offsets.json"))
+  puts "Copied over offsets.json"
+
+  FileUtils.cp_r(output, File.join("output/latest"))
```
mfurtak: Rake has the helper method `cp_r` built-in. 👍

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,76 @@
...
+
+  raise "Could not find any images, make sure to copy all extracted files into ./raw" if image_assets.count < 10
+
+  File.write(File.join(output, "files.json"), image_assets.to_json)
+  puts "Wrote names of available files to files.json"
+
+  File.write(File.join(output, "version.txt"), current_time)
+  puts "Wrote the current time stamp to version.txt"
+
+  FileUtils.cp("offsets.json", File.join(output, "offsets.json"))
```
mfurtak: Rake has the helper method `cp` built-in. 👍

frameit/frames_generator/Rakefile
mfurtak: `File.join("output", "latest")` or just use the string?

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,76 @@
...
+
+  system("open ./output")
+  puts "Done writing all the things to './output', please upload both the 'latest' and the time stamp '#{current_time}' folder to GitHub".green
+  puts "To do so, clone https://github.com/fastlane/frameit-frames, make sure to check out the `gh-pages` branch".green
+  puts "and overwrite the 'latest' folder, and copy the time stamp folder into the root.".green
+  puts "Push the changes, and run `frameit update_frames` to ensure everything worked as expected".green
+end
+
+def confirm
+  puts "Confirm when complete with the last step".yellow
```
mfurtak: Suggestion: "Press any key when you're ready to continue"

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,76 @@
...
+  gets
+end
+
+# This will remove the white space around the actual frame
+def convert_image_resource(path)
+  MiniMagick.configure do |config|
+    config.validate_on_create = false
+  end
```
mfurtak: If this configuration only needs to be done once, can we extract it up to an earlier code location?

frameit/lib/frameit/editor.rb
```diff
@@ -37,17 +42,31 @@ def prepare_image
...
+        frame.rotate(-90)
+        @image.rotate(-90)
+      end
+
       @image = frame.composite(image, "png") do |c|
         c.compose "Over"
         c.geometry offset['offset']
       end
+
+      # We have to revert the state to be landscape screenshots
```
mfurtak: "revert... to be portrait"?

frameit/lib/frameit/frame_downloader.rb
```diff
@@ -0,0 +1,69 @@
...
+      UI.message "non-infringement of intellectual property rights."
+      UI.message "----------------------------------------"
+    end
+
+    private
+
+    def download_file(path, txt: "file")
+      url = File.join(HOST_URL, Frameit.frames_version, path)
+      UI.message("Downloading #{txt} from '#{url}' ...")
+      body = Excon.get(url.gsub(' ', '%20')).body
```
mfurtak:
```ruby
require 'uri'

url = URI.join(HOST_URL, Frameit.frames_version, URI.escape(path))
```

And then you can remove the manual escaping. 👍

mfurtak: For the future, we should try to stream the downloaded data directly to disk instead of storing it entirely in memory first.

mfurtak (changes requested):
Really awesome changes overall 🙌 

However... for many changes, many comments 😉

frameit/frames_generator/Rakefile
KrauseFx: Just a string should be fine

frameit/lib/frameit/editor.rb
KrauseFx: This comment is correct, those are landscape screenshots, which we only temporarly rotate to be portrait for the imagemagick math. Should I update the comments?

frameit/lib/frameit/frame_downloader.rb
KrauseFx: The escaping part is great, however the URI.join doesn't work with 3 components.

```ruby
URI.join(HOST_URL, Frameit.frames_version, URI.escape(path))
# => => #<URI::HTTPS https://fastlane.github.io/version.txt>
```

I keep using `File.join`, but use the proper escape method 👍

KrauseFx:
Thanks for the feedback, ready for another review 👍

mfurtak (approved):

frameit/frames_generator/README.md
```diff
@@ -0,0 +1,12 @@
+# Prepare and upload new device frames
+
+Since October 2016 we host the device frame resources since Apple removed lots of old device frame images.
```
i2amsam:
```
Starting in October 2016 we are now using the (Facebook frameset)[http://facebook.design/devices] and hosting the device frame resources since Apple sometimes has removed / changed some of the old images making framing a moving target.

To run the wizard to prepare new device assets, just run `rake` in this directory.

You will need to run this wizard every time you want update the device assets

This will generate 2 directories:

- The timestamp based one which is used if the user specifies a particular version in their Framefile
- The the most recent one which will be used by default
```

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,75 @@
...
+  puts ""
+  confirm
+  puts "Move the content of the zip file into './raw'"
+  mkdir("raw") unless File.directory?("raw")
+  system("open ./raw")
+  confirm
+
+  # Let's find all the image resources we need and copy it into the correct directory
+  current_time = Time.now.to_i.to_s
+  output = File.join("output", current_time)
```
i2amsam: Instead of current_time, what do you think about using the hash of the zip?

frameit/frames_generator/Rakefile
```diff
@@ -0,0 +1,75 @@
...
+  cp("offsets.json", File.join(output, "offsets.json"))
+  puts "Copied over offsets.json"
+
+  cp_r(output, "output/latest")
+
+  system("open ./output")
+  puts "Done writing all the things to './output', please upload both the 'latest' and the time stamp '#{current_time}' folder to GitHub".green
+  puts "To do so, clone https://github.com/fastlane/frameit-frames, make sure to check out the `gh-pages` branch".green
+  puts "and overwrite the 'latest' folder, and copy the time stamp folder into the root.".green
+  puts "Push the changes, and run `frameit update_frames` to ensure everything worked as expected".green
```
i2amsam: This looks great!  A useful next step here once the frames are downloaded might be to generate screenshots for all the combinations to see if the offsets need to be updated.

frameit/lib/frameit/template_finder.rb
```diff
@@ -1,84 +1,32 @@
 module Frameit
   # Responsible for finding the correct device
   class TemplateFinder
-    class FilenameTransform
-      def initialize(device_name)
-        @device_name = device_name
-      end
-    end
-
-    # Example: iPhone_5s_Vert_SpaceGray_sRGB
```
i2amsam: Wooo!  Gross code gone!

i2amsam (commented):
This is a very nice change 👍 made a couple of minor suggestions about possible enhancements, but this looks good to me!

frameit/frames_generator/Rakefile
mfurtak: I like a directory name that sorts well and has some meaning. Maybe append a hash to the end?

KrauseFx: Interesting - why would we want to add a hash? What's the benefit of it?

KrauseFx:
💥

Issue #6585, opened 2016-10-17
[Internal] Add specs for `badge` action


AliSoftware:
The `badge` action doesn't have any specs yet

See also https://github.com/fastlane/fastlane/pull/6568#issuecomment-254102564

mfurtak:
Implemented in #6605 👍

PR #6605, opened 2016-10-18
Provide a test for badge action ImageMagick verification

mfurtak:
Follow-up to #6568

/cc @AliSoftware

### DIFF
--- a/fastlane/lib/fastlane/actions/badge.rb
+++ b/fastlane/lib/fastlane/actions/badge.rb
@@ -143,17 +143,16 @@ module Fastlane
       end
 
       def self.check_imagemagick!
-        unless `which convert`.include? 'convert'
-          UI.error("You have to install ImageMagick to use `badge`")
-          UI.error("")
-          UI.error("Install it using:")
-          UI.command("brew update && brew install imagemagick")
-          UI.error("")
-          UI.error("If you don't have homebrew, visit http://brew.sh")
+        return if `which convert`.include?('convert')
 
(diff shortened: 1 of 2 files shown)


AliSoftware:
Cool! This will also allow to close #6585 I opened to be sure not to forget :wink:

asfalcone (approved):

AliSoftware (commented):
Shouldn't we also write a passing test? (if ImageMagick is present, ensure we don't fail)

mfurtak:
Yeah, that would be good, I agree. However, it's a bit trickier to do because once the action continues on past the ImageMagick check, it immediately wants to require and use the Badge gem.

In order to turn the success case into a test that can actually run to completion, I think I would have to mock out almost 100% of the logic. At that point, I'm not sure that I'm really testing very much.

One option is to remove the private method restriction on `check_imagemagick!` and test it more directly. What do folks think about that?

AliSoftware:
Ah, I see your point.

I think it would be better and more logical to directly test `check_imagemagick!` instead of testing `badge`.

We can still keep it private in the implementation, and use `BadgeAction.send(:check_imagemagick!)` from the specs to workaround the private accessibility 😉

AliSoftware (changes requested):
Test `BadgeAction.send(:check_imagemagick!)` directly instead, to be able to provide both success & failure tests

mfurtak:
I'm going to keep the top-level action test, as I think it's good that every action has at least one test that tries to use it as part of a Fastfile evaluation. 👍

KrauseFx:
👍

Issue #6595, opened 2016-10-17
Screengrab: Screenshots not copied to local filesystem from android device after test failure


lalunamel:
### Issue Description

When running screengrab in our Android UI Automation tests, Screengrab fails to copy screenshots taken from the android device to the local filesystem.
##### Complete output when running fastlane, including the stack trace and command used

```
./gradlew assembleDebug assembleAndroidTest

...NORMAL BUILD OUTPUT...

screengrab

+-----------------------------+-------------------------------------------------+
|                         Summary for screengrab 0.5.5                          |
+-----------------------------+-------------------------------------------------+
| app_package_name            | com.appname.debug                        |
| app_apk_path                | app/build/outputs/apk/app-debug.apk             |
| tests_apk_path              | app/build/outputs/apk/app-debug-androidTest.apk |
| clear_previous_screenshots  | true                                            |
| output_directory            | /Users/lalunamel/project   |
| tests_package_name          | com.appname.debug.test                   |
| android_home                | /Users/lalunamel/Library/Android/sdk              |
| locales                     | ["en-US"]                                       |
| skip_open_summary           | false                                           |
| test_instrumentation_runner | android.support.test.runner.AndroidJUnitRunner  |
| ending_locale               | en-US                                           |
| device_type                 | phone                                           |
+-----------------------------+-------------------------------------------------+

[15:12:41]: Limiting the test classes run by `screengrab` to just those that generate screenshots can make runs faster.
[15:12:41]: Consider using the :use_tests_in_classes or :use_tests_in_packages option, and organize your tests accordingly.
[15:12:41]: Clearing phoneScreenshots within /Users/lalunamel/appname
[15:12:42]: $ adb devices -l
[15:12:42]: ▸ List of devices attached
[15:12:42]: ▸ emulator-5554          device product:sdk_google_phone_x86_64 model:Android_SDK_built_for_x86_64 device:generic_x86_64
[15:12:42]: $ adb -s emulator-5554 shell echo \$EXTERNAL_STORAGE
[15:12:42]: ▸ /sdcard
[15:12:42]: Cleaning screenshots on device
[15:12:42]: $ adb -s emulator-5554 shell rm -rf /sdcard/com.appname.debug/screengrab
[15:12:42]: Validating app APK
[15:12:42]: $ /Users/lalunamel/Library/Android/sdk/build-tools/24.0.3/aapt dump permissions app/build/outputs/apk/app-debug.apk
[...]

PR #6606, opened 2016-10-18
[#6595] Add exit_on_test_failure option to allow copying screenshots …

lalunamel:


screengrab/lib/screengrab/runner.rb
```diff
@@ -225,7 +225,11 @@ def run_tests(device_serial, test_classes_to_use, test_packages_to_use)
                                       print_all: true,
                                       print_command: true)
 
-        UI.user_error!("Tests failed", show_github_issues: false) if test_output.include?("FAILURES!!!")
+        if @config[:exit_on_test_failure]
```
lalunamel: I would wrap this block in the `if test_output.include?("FAILURES!!!")` but rubocop then complains I need to include a `next` to skip the current iteration of the loop (which is not necessary)

screengrab/spec/runner_spec.rb
```diff
@@ -8,12 +8,60 @@
     @runner = Screengrab::Runner.new(mock_executor, config, mock_android_environment)
   end
 
-  def set_mock_adb_response(command, mock_response)
+  def mock_adb_response_for_command(command, mock_response)
```
lalunamel: Again, rubocop complains that I can't have a method with the word 'set' in it that's a setter. (even though that existed before this commit)

lalunamel:
https://github.com/fastlane/fastlane/issues/6595

asfalcone (approved):

PR #6613, opened 2016-10-18
Set FORK_XCODE_WRITING env variable to work around xcodeproj chdir bug

### LINKED ISSUE #6278: update_app_identifier and update_project_provisioning both change current directory, but do not change back when done
### New Issue Checklist

### Issue Description

If you run the `update_app_identifier` or the `update_project_provisioning` action, they change directories from `${SRCROOT}/fastlane` to `${SRCROOT}`, but they never change back when they are done. This causes other code that looks for files in a specific path to always fail to find the file, and fastlane fails at that point. We run fastlane on multiple apps at once in a loop, so we're running these actions multiple times per lane execution. This only started happening recently, somewhere between fastlane 1.101 and 1.104.

```
Prefer not to include, as it's sensitive. See example lane instead that reproduces issue
```

### Configuration Files

Please copy the complete content of your `Fastfile` and any other configuration files you use below:

Fastfile:
```ruby
  # lane that reproduces issue:
  lane :bug do
      puts File.absolute_path(".")
      # first time it's ${SRCROOT}/fastlane, second time it's ${SRCROOT}.
      # It should always be one or the other (preferrably the first as that's how it worked before).

      # replace with your own project info
      proj = 'WayfairApp.xcodeproj'
      plist = 'WayfairApp/Info.plist'
      identifier = 'com.wayfair.WayfairApp'
      target_pattern = '^WayfairApp$'

      # problem action 1
      update_app_identifier(xcodeproj: proj,
                            plist_path: plist,
                            app_identifier: identifier)
      # problem action 2
      cert
      sigh(app_identifier: identifier)
      update_project_provisioning(xcodeproj: proj,
                                  target_filter: target_pattern)

      puts File.absolute_path(".")
  end
```

### Environment

fastlane version (run `fastlane -v`): fastlane 1.104.0

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)? no

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)? rbenv, but happens on machines without it

asfalcone:
Fixes #6278

### DIFF
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -38,6 +38,12 @@ module Fastlane
 
       platform, lane = choose_lane(ff, platform) unless lane
 
+      # xcodeproj has a bug in certain versions that causes it to change directories
+      # and not return to the original working directory
+      # https://github.com/CocoaPods/Xcodeproj/issues/426
+      # Setting this environment variable causes xcodeproj to work around the problem
+      ENV["FORK_XCODE_WRITING"] = "true" unless platform == 'android'
+
       load_dot_env(env)
 
       started = Time.now


KrauseFx:
Why is this change in `lane_manager`? So that we know if it's a non-Android lane?

asfalcone:
@KrauseFx It seemed like the best top-level place to put it so that we could add it before any lane runs. Where else do you think it could live? Open to suggestions.

KrauseFx:
Users can also run fastlane actions as one-off actions, without a lane, I think the code block is not called. I think the best way to add it is probably in `fastlane.rb` inside the `Fastlane` module, this way we can be sure this value is set every time.

asfalcone:
If they ran individual actions the bug that this fixes wouldn't come into play because the working directory after the `update_app_identifier` action wouldn't matter. The bug is really about using actions in combination in a lane, I think.

KrauseFx:
Good point, I'm okay either way. Putting it to `fastlane.rb` might be easier to find than `lane_manager`, as it's not specific to lanes. 
We can leave it as is though

KrauseFx (approved):

PR #6573, opened 2016-10-17
Update rubocop and update styling rules

KrauseFx:
- Update code to fit new rules

Fixes https://github.com/fastlane/fastlane/issues/6570

PR #6614, opened 2016-10-18
[fastlane_core] Version bump

KrauseFx:
- Print out summary of config files to make debugging easier (#6572)
- Update rubocop and update styling rules (#6573)


mfurtak (approved):
:gem: ✨

PR #6635, opened 2016-10-19
Match: support for supplying multiple app identifiers at once

KrauseFx:
Taken from https://github.com/fastlane/fastlane/pull/6080


match/lib/match/runner.rb
```diff
@@ -12,25 +12,43 @@ def run(params)
...
+        app_identifiers = params[:app_identifier]
+      else
+        app_identifiers = params[:app_identifier].split(/\s*,\s*/).uniq
+      end
+
       # Verify the App ID (as we don't want 'match' to fail at a later point)
-      spaceship.bundle_identifier_exists(params) if spaceship
+      if spaceship
+        app_identifiers.each do |app_identifier|
+          params[:app_identifier] = app_identifier
```
mfurtak: I don't really like the pattern of mutating the `params` `Configuration` object for every iteration over the `app_identifiers` that we do. If these `SpaceshipEnsure` methods don't actually need `Configuration` objects, can we just provide them hashes with the right keys instead?

match/lib/match/runner.rb
```diff
@@ -12,25 +12,43 @@ def run(params)
...
       cert_id = fetch_certificate(params: params)
       spaceship.certificate_exists(params, cert_id) if spaceship
 
-      # Provisioning Profile
-      uuid = fetch_provisioning_profile(params: params,
-                                certificate_id: cert_id)
-      spaceship.profile_exists(params, uuid) if spaceship
+      # Provisioning Profiles
+      app_identifiers.each do |app_identifier|
+        params[:app_identifier] = app_identifier
```
mfurtak: Same thought about mutating the `Configuration` object in the loop.

match/lib/match/runner.rb
```diff
@@ -12,25 +12,43 @@ def run(params)
...
       # Done
       if self.changes_to_commit and !params[:readonly]
         message = GitHelper.generate_commit_message(params)
         GitHelper.commit_changes(params[:workspace], message, params[:git_url], params[:git_branch])
       end
 
-      TablePrinter.print_summary(params)
+      # Print a summary table for each app_identifier
+      app_identifiers.each do |app_identifier|
+        params[:app_identifier] = app_identifier
```
mfurtak: Same thought about mutating the `Configuration` object in the loop.

mfurtak (changes requested):

match/lib/match/runner.rb
KrauseFx: I agree, I'll update the PR 👍

KrauseFx:
@mfurtak Perfect, thanks for your feedback. I refactored _match_ to properly pass the `app_identifier` around, great feedback - let me know what you think about the updated code

mfurtak (approved):

mfurtak:
🍕 🎉

jcampbell05:
Awesome !

PR #6578, opened 2016-10-17
fastlane env

hjanuschka:
- [x] atleast on unit test
- [x] documentation on how to use this feature, mostly in the ISSUE_TEMPLATE.md
- [x] Output should be nice markdown output, easy to copy and paste to GitHub
- [x] Instructions in the terminal on how to copy & paste to GitHub
- [x] Listing all the used Ruby gem versions (something similar to the Gemfile.lock, but has to be   fetched from the run time). It's easy to fetch the version of all gems that are being loaded
- [x] Xcode path (xcode-select -p)
- [x] Version of the Xcode installation on xcode-select -p
- [x] Print out the stack similar to CocoaPods (see below)
- [x] Executable path (similar to CocoaPods)
- [x] Loaded plugins (if any)

i took over https://github.com/fastlane/fastlane/pull/6252 - hope it is ok @KrauseFx 

Sample Output:
### Stack

| Key | Value |
| --- | --- |
| Fastlane | 1.105.2 |
| OS | 10.12 |
| Ruby | 2.2.1 |
| Xcode | /Applications/Xcode.app/Contents/Developer |
| Xcode Version | Xcode 8.0 Build version 8A218a |
| Git | git version 2.10.0 |
| Installation Source | /Users/hjanuschka/.rvm/gems/ruby-2.2.1/bin/fastlane |
| Host | Mac OS X 10.12 (16A323) |
| Ruby Lib Dir | /Users/hjanuschka/.rvm/rubies/ruby-2.2.1/lib |
### Fastlane Files:

`Fastfile` located in ./fastlane/Fastfile

```
lane :demo do
    puts "my fastfile"
end
```

`Appfile` located in ./fastlane/Appfile

```
my_app_file=1
```
### Fastlane Gems

| Gem | Version | Update-Status |
| --- | --- | --- |
| credentials_manager | 0.16.2 | 😇 Up-To-Date |
| fastlane_core | 0.52.1 | 😇 Up-To-Date |
| spaceship | 0.36.0 | 😇 Up-To-Date |
| cert | 1.4.3 | 😱 Update availaible |
| deliver | 1.14.2 | 😇 Up-To-Date |
| frameit | 2.8.0 | 😇 Up-To-Date |
| gym | 1.11.2 | 😇 Up-To-Date |
| sigh | 1.11.2 | 😇 Up-To-Date |
| match | 0.8.1 | 😇 Up-To-Date |
| pem | 1.3.2 | 😇 Up-To-Date |
[...]

PR #6649, opened 2016-10-20
fastlane env

KrauseFx:
Taken from https://github.com/fastlane/fastlane/pull/6578

screenshot 2016-10-20 11 17 06

### DIFF
--- a/fastlane/lib/fastlane/commands_generator.rb
+++ b/fastlane/lib/fastlane/commands_generator.rb
@@ -234,7 +234,7 @@ module Fastlane
           require "fastlane/environment_printer"
           env_info = Fastlane::EnvironmentPrinter.get
           puts env_info
-          if UI.confirm("🙄 Wow, that's a lot of markdown text... should fastlane put it into your clipboard, so you can easily paste it on GitHub?")
+          if UI.confirm("🙄  Wow, that's a lot of markdown text... should fastlane put it into your clipboard, so you can easily paste it on GitHub?")
             Fastlane::EnvironmentPrinter.copy_to_clipboard(env_info)
             UI.success("Successfully copied markdown into your clipboard 🎨")
           end
--- a/fastlane/lib/fastlane/commands_generator.rb
+++ b/fastlane/lib/fastlane/commands_generator.rb
@@ -234,6 +234,11 @@ module Fastlane
           require "fastlane/environment_printer"
           env_info = Fastlane::EnvironmentPrinter.get
           puts env_info
+          if UI.confirm("🙄 Wow, that's a lot of markdown text... should fastlane put it into your clipboard, so you can easily paste it on GitHub?")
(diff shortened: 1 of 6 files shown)


KrauseFx:
This will fail until we have a new fastlane_core release 👍

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -0,0 +1,192 @@
...
+      env_output << "\n\n"
+      env_output
+    end
+
+    # Copy a given string into the clipboard
+    # Make sure to ask the user first, as some people don't
+    # use a clipboard manager, so they might lose something important
+    def self.copy_to_clipboard(string)
+      require 'tmpdir'
+      tmp_file = File.join(Dir.mktmpdir, "temporary") # so that we don't have problems with escaping the string
```
mfurtak: What do you think about:
```ruby
require 'tempfile'
Tempfile.create('environment_printer') do |tmp_file|
  File.write(tmp_file, string)
  `cat '#{tmp_file}' | pbcopy`
end
```

fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -224,6 +227,21 @@ def run
...
       end
 
+      command :env do |c|
+        c.syntax = 'fastlane env'
+        c.description = 'Print your fastlane environment, use this when you submit an issue on GitHub'
+        c.action do |args, options|
+          require "fastlane/environment_printer"
+          env_info = Fastlane::EnvironmentPrinter.get
+          puts env_info
+          if UI.confirm("🙄  Wow, that's a lot of markdown text... should fastlane put it into your clipboard, so you can easily paste it on GitHub?")
```
mfurtak: `pbcopy` will only work for Mac users, so should we avoid showing it to people on other platforms?

mfurtak (commented):
Seems great overall - I left two comments.

fastlane/lib/fastlane/commands_generator.rb
KrauseFx: Mh, we don't currently support any other platforms, right? But I agree, we probably want to show this only for mac users.

mfurtak: We have definitely tried to make things better for Linux users before 😄

hjanuschka: i think its done with `xclip` on linux  - but has to be checked if installed.

fastlane/lib/fastlane/environment_printer.rb
KrauseFx: Thanks, I updated the code. Note that `tmp_file` is not the path, but a `File` object, I had to change the code to use `tmp_file.path` at least on the shelling out part

KrauseFx:
Ready for another review 👍

mfurtak (approved):

PR #6651, opened 2016-10-20
[deliver] Print out absolute path if file can't be found

KrauseFx:
This way it's clear in what path _fastlane_ is running

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -31,7 +31,7 @@ module Deliver
                                      description: "Path to your ipa file",
                                      default_value: Dir["*.ipa"].first,
                                      verify_block: proc do |value|
-                                       UI.user_error!("Could not find ipa file at path '#{value}'") unless File.exist?(value)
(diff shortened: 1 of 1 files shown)


mfurtak:
`File.exist?` has trouble with certain kinds file paths. Specifically, I've noticed in the past that it can't handle paths that start with '~'.

```
$ irb
>> File.exist? '~/Documents'
=> false
>> File.exist? File.expand_path('~/Documents')
=> true
```

So, for all of these checks, I think it may be better to capture the `File.expand_path` value first, and then check it in `File.exist?`, and then use it in the message if needed.

mfurtak (changes requested):
`File.exist?` should ideally only be called on expanded paths.

KrauseFx:
@mfurtak Yeah, that's true, and we do it in some tools, not everywhere yet though. Let's merge this PR that just improves the error message to make it more clear on how _fastlane_ accesses the files, and have a separate PR that improves file checking and loading across all tools 👍

mfurtak (approved):

PR #6659, opened 2016-10-21
[WIP] Unify fastlane environment variables

AliSoftware:
## Current state: the inconsistency

Environment variables used to configure various fastlane actions are inconsistent.

Some are prefixed with `ACTIONNAME_`, some are prefixed with `FL_ACTIONNAME_`, some are not prefixed at all or wrongly prefixed. Examples include:
- `scan` using `SLACK_URL` instead of `SCAN_SLACK_URL`
- `gym` using `XCPRETTY_FORMATTER` or `SUPPRESS_OUTPUT`
- `match` using `FASTLANE_TEAM_ID` instead of `MATCH_TEAM_ID`
  …

This PR aims to add consistency in all the env variables. I know this might break some people's configurations (they'll have to rename their environment variables if they used the old inconsistent names we are changing) but I think it's important, for reasons stated below.
## The problem it creates

More than just a consistency matter, not having those environment variables prefixed can cause multiple issues:
- Conflicts with an environment variables used and set by another tool totally different from `fastlane`
- Side effects with settings from one action having effects on other actions.

I recently hit one of these side effects while setting `ENV['SLACK_URL']` in my `Fastfile` in order to call the `slack` action manually after a successful OTA being published on Crashlytics Beta (at the end of my `ota` lane). Setting that env var, which I thought was only used by the `slack` action, I was surprised to see my slack spammed with a lot of messages coming from `scan`, as `scan` uses the `SLACK_URL` env var instead of `SCAN_SLACK_URL` (all other env vars used by `scan` are correctly prefixed, this one was the exception)
## Proposed solutions

I added a `spec` to test all actions and test all of their available options' env var names. It appears that there are a LOT of actions not namespacing their env vars correctly to avoid collision.
1. Currently, the spec allows both `ACTIONNAME` and `FL_ACTIONNAME` as valid env var name prefixes, as I've seen both being used. I wonder if that would be a good idea to force all actions to be `FL_ACTIONNAME`, but that would mean rename a lot more of env variables, so that might be too big a change?
2. I allowed some actions to have a custom prefix (see `prefix_map` in the spec). That makes sense for some actions, but then:
   - I wonder if that custom prefix shouldn't be stored in the `Fastlane::Actions` class itself (the parent class defaulting to `name.upcase`) instead of having such a hash in the specs.
   - If we do so, it would be cleaner to make every call to `FastlaneCore::ConfigItem.new` in `Fastlane::Actions.available_options` to use that new `env_var_prefix` we'd introduced, and I'm not sure how to do that. Explicitly call `env_var_prefix + "URL"` when building each `FastlaneCore::ConfigItem.new`? would be cumbersome. Add a convenience method in `Fastlane::Actions` to build a `FastlaneCore::ConfigItem` and wrap that prefixing? Maybe better
## Current status

:warning: This PR is a Work in Progress. I've added the spec to highlight all actions where env var names are inconsistent, but haven't fixed them all, as **I'm waiting for your input about what's the most consistent naming convention you all prefer and in which direction we're going**.

_Also, I'm not sure how I could make `rspec` run the tests with the local code, in order to take into account the changes I made e.g. in `scan`'s source code: when the spec is running it seems to load the `scan` gem installed on my machine instead of the local `scan` source where the env var has been fixed already?_
## Vote!

1️⃣ Add `Fastlane::Action.env_var_prefix` so that the prefix is stored in each action instead of having the `prefix_map` in the specs [YES/NO]
2️⃣ Add the convenience wrapper `config_item` which will prefix the `env_var` for us using that `env_var_prefix` instead of having the `FastlaneCore::ConfigItem.new` with explicit full name [YES/NO]
3️⃣ Enforce all env variables of every _action_ other than a fastlane tool (so `slack` or `get_version_number` for example, but not `gym` or `scan`) to start with `FL_` [YES/NO]
4️⃣ Enforce all env variables of tools (like `scan`, `gym`, to also start with `FL_` (like `FL_GYM_…`) [YES/NO]

_My personal vote is yes for 1️⃣ & 3️⃣ , no for 2️⃣ & 4️⃣ ._


AliSoftware:
I just started implementing the 2nd suggestion (convenience wrapper to build config_items) in fc85748bb3380b497ff5d534af66e88c8e8666e0, but only on `scan`, so that you can see what I had in mind and so we can discuss if that's the direction we want to move to (and I should continue this way for other actions) or not (and I'll revert that commit).

AliSoftware:
Here are the current mismatched env names detected by the spec (only the first mismatching env name is detected since it makes the spec fail as soon as it encounter the first failed expectation)

| Action | Env Var | Expected Start | Accepted Start |
| ------ | ------- | -------------- | -------------- |
| `adb` | `FL_ANDROID_SERIAL` | `FL_ADB` | `ADB` |
| `adb_devices` | `FL_ADB_PATH` | `FL_ADB_DEVICES` | `ADB_DEVICES` |
| `appstore` | `DELIVER_USERNAME` | `FL_APPSTORE` | `APPSTORE` |
| `clear_derived_data` | `DERIVED_DATA_PATH` | `FL_CLEAR_DERIVED_DATA` | `CLEAR_DERIVED_DATA` |
| `commit_version_bump` | `FL_COMMIT_BUMP_MESSAGE` | `-` | `FL_VERSION_BUMP` |
| `create_pull_request` | `GITHUB_PULL_REQUEST_API_TOKEN` | `FL_CREATE_PULL_REQUEST` | `CREATE_PULL_REQUEST` |
| `get_build_number_repository` | `USE_HG_REVISION_NUMBER` | `-` | `FL_BUILD_NUMBER_REPO` |
| `get_github_release` | `FL_GITHUB_RELEASE_SERVER_URL` | `FL_GET_GITHUB_RELEASE` | `GET_GITHUB_RELEASE` |
| `gradle` | `FL_ANDROID_SERIAL` | `FL_GRADLE` | `GRADLE` |
| `gym` | `SUPPRESS_OUTPUT` | `FL_GYM` | `GYM` |
| `hg_commit_version_bump` | `FL_COMMIT_BUMP_MESSAGE` | `-` | `FL_VERSION_BUMP` |
| `import_certificate` | `KEYCHAIN_NAME` | `FL_IMPORT_CERTIFICATE` | `IMPORT_CERTIFICATE` |
| `increment_build_number` | `FL_BUILD_NUMBER_BUILD_NUMBER` | `FL_INCREMENT_BUILD_NUMBER` | `INCREMENT_BUILD_NUMBER` |
| `increment_version_number` | `FL_VERSION_NUMBER_BUMP_TYPE` | `FL_INCREMENT_VERSION_NUMBER` | `INCREMENT_VERSION_NUMBER` |
| `install_on_device` | `FL_IOD_EXTRA` | `FL_INSTALL_ON_DEVICE` | `INSTALL_ON_DEVICE` |
| `install_xcode_plugin` | `FL_XCODE_PLUGIN_URL` | `FL_INSTALL_XCODE_PLUGIN` | `INSTALL_XCODE_PLUGIN` |
| `latest_testflight_build_number` | `FASTLANE_APP_IDENTIFIER` | `FL_LATEST_TESTFLIGHT_BUILD_NUMBER` | `LATEST_TESTFLIGHT_BUILD_NUMBER` |
| `match` | `FASTLANE_TEAM_ID` | `FL_MATCH` | `MATCH` |
| `nexus_upload` | `FL_NEXUS_FILE` | `FL_NEXUS_UPLOAD` | `NEXUS_UPLOAD` |
| `onesignal` | `ANDROID_TOKEN` | `-` | `ONE_SIGNAL` |
| `push_git_tags` | `FL_PUSH_GIT_FORCE` | `-` | `PUSH_GIT` |
| `push_to_git_remote` | `FL_GIT_PUSH_LOCAL_BRANCH` | `FL_PUSH_TO_GIT_REMOTE` | `PUSH_TO_GIT_REMOTE` |
| `recreate_schemes` | `XCODE_PROJECT` | `FL_RECREATE_SCHEMES` | `RECREATE_SCHEMES` |
| `register_devices` | `DELIVER_USER` | `FL_REGISTER_DEVICES` | `REGISTER_DEVICES` |
| `reset_git_repo` | `FL_RESET_GIT_FILES` | `FL_RESET_GIT_REPO` | `RESET_GIT_REPO` |
| `resign` | `FL_DISPLAY_NAME` | `FL_RESIGN` | `RESIGN` |
| `s3` | `""` | `FL_S3` | `S3` |
| `scan` | `SLACK_URL` | `FL_SCAN` | `SCAN` |
| `scp` | `FL_SSH_USERNAME` | `FL_SCP` | `SCP` |
| `set_build_number_repository` | `USE_HG_REVISION_NUMBER` | `FL_SET_BUILD_NUMBER_REPOSITORY` | `SET_BUILD_NUMBER_REPOSITORY` |
| `set_changelog` | `FASTLANE_APP_IDENTIFIER` | `FL_SET_CHANGELOG` | `SET_CHANGELOG` |
| `set_github_release` | `FL_GITHUB_RELEASE_SERVER_URL` | `FL_SET_GITHUB_RELEASE` | `SET_GITHUB_RELEASE` |
| `set_info_plist_value` | `FL_SET_INFO_PLIST_PARAM_NAME` | `FL_SET_INFO_PLIST_VALUE` | `SET_INFO_PLIST_VALUE` |
| `setup_jenkins` | `KEYCHAIN_PATH` | `FL_SETUP_JENKINS` | `SETUP_JENKINS` |
| `testflight` | `PILOT_USERNAME` | `FL_TESTFLIGHT` | `TESTFLIGHT` |
| `twitter` | `FL_TW_CONSUMER_KEY` | `FL_TWITTER` | `TWITTER` |
| `update_app_group_identifiers` | `FL_UPDATE_APP_GROUP_IDENTIFIER_ENTITLEMENTS_FILE_PATH` | `FL_UPDATE_APP_GROUP_IDENTIFIERS` | `UPDATE_APP_GROUP_IDENTIFIERS` |
| `update_fastlane` | `FL_TOOLS_TO_UPDATE` | `FL_UPDATE_FASTLANE` | `UPDATE_FASTLANE` |
| `update_info_plist` | `FL_UPDATE_PLIST_PROJECT_PATH` | `FL_UPDATE_INFO_PLIST` | `UPDATE_INFO_PLIST` |
| `update_project_code_signing` | `FL_PROJECT_SIGNING_PROJECT_PATH` | `FL_UPDATE_PROJECT_CODE_SIGNING` | `UPDATE_PROJECT_CODE_SIGNING` |
| `update_project_provisioning` | `FL_PROJECT_PROVISIONING_PROJECT_PATH` | `FL_UPDATE_PROJECT_PROVISIONING` | `UPDATE_PROJECT_PROVISIONING` |
| `update_project_team` | `FL_PROJECT_SIGNING_PROJECT_PATH` | `FL_UPDATE_PROJECT_TEAM` | `UPDATE_PROJECT_TEAM` |
| `update_urban_airship_configuration` | `URBAN_AIRSHIP_PLIST_PATH` | `FL_UPDATE_URBAN_AIRSHIP_CONFIGURATION` | `UPDATE_URBAN_AIRSHIP_CONFIGURATION` |
| `upload_symbols_to_crashlytics` | `CRASHLYTICS_API_TOKEN` | `FL_UPLOAD_SYMBOLS_TO_CRASHLYTICS` | `UPLOAD_SYMBOLS_TO_CRASHLYTICS` |
| `upload_symbols_to_sentry` | `SENTRY_HOST` | `FL_UPLOAD_SYMBOLS_TO_SENTRY` | `UPLOAD_SYMBOLS_TO_SENTRY` |
| `version_get_podspec` | `FL_VERSION_PODSPEC_PATH` | `FL_VERSION_GET_PODSPEC` | `VERSION_GET_PODSPEC` |
| `xcode_install` | `FL_XCODE_VERSION` | `FL_XCODE_INSTALL` | `XCODE_INSTALL` |
| `xcversion` | `FL_XCODE_VERSION` | `FL_XCVERSION` | `XCVERSION` |

KrauseFx:
Hey @AliSoftware, just came across this PR again, great work on researching and preparing this change. I'd love to unify the env variables, and have a better overview of what variables are available and have them have a consistent naming. We can't really introduce breaking changes right now, so even if we rename the variables, we'd still need to be backwards compatible. We could potentially then deprecate the old naming soon, and remove them with a 3.0 release. 

What do you think?

AliSoftware:
@KrauseFx Yeah, that seems like a good plan. Keeping backward compatibility with depreciations and removing in 3.0 is ok to me.

I won't probably have time to go forward with this PR though tbh, as I already have a lot of work pending on SwiftGen (especially AliSoftware/SwiftGen#240 and migrating SwiftGen to its own org), so feel free to take over.

Especially I don't know what direction you prefer amongst the ones I suggested in my original comment, especially which would be the best way to implement that so it's easy to migrate, still supports multiple variants for each env var so that we can have backwards compatibility (and keep a clean architecture with a good design pattern choice, i.e. convenience method vs wrapper vs a type variable per `Action`)

mfurtak:
Thanks for the good work and research here @AliSoftware - we can pick this up again in the future 👍

PR #6669, opened 2016-10-21
[frameit] Version bump

KrauseFx:
- Rewrote the whole framing process, so that you don't have to do anything manually any more
- _frameit_ now automatically downloads and installs all the frames the first time you use it
- Support for Rose Gold, Gold and the new Matte Black
- Support for  the new iPhone 7
- Full support for portrait and landscape mode
- You can lock a specific version of frames in your `Framefile.json` if you want to, to have consistent builds in the future

**Rose Gold**
screenshot 2016-10-21 15 11 23
**Gold**
screenshot 2016-10-21 15 14 30

### DIFF
--- a/frameit/lib/frameit/version.rb
+++ b/frameit/lib/frameit/version.rb
@@ -1,3 +1,3 @@
 module Frameit
-  VERSION = "2.8.0".freeze
+  VERSION = "3.0.0".freeze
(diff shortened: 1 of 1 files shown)


mfurtak (approved):
💎 ✨

PR #6708, opened 2016-10-24
Match enterprise type mismatch

milch:
There was a mismatch in comparing the types of provisioning profiles when `type: "appstore"` was used with the `MATCH_FORCE_ENTERPRISE`. Generating the profiles would work, but when cloning on a different machine it would try to generate a new certificate or error out in `readonly` mode. 

The error now includes more useful information on why no provisioning profile was found. Conversion of type strings to symbols is now centralized, so the above error doesn't happen anymore.

### DIFF
--- a/match/lib/match.rb
+++ b/match/lib/match.rb
@@ -29,4 +29,21 @@ module Match
   def self.enterprise?
     ENV["MATCH_FORCE_ENTERPRISE"]
   end
+
+  def self.type_is_enterprise?(type)
+    Match.enterprise? && type != "development"
+  end
+
+  def self.profile_type_sym(type)
+    return :enterprise if self.type_is_enterprise? type
+    return :adhoc if type == "adhoc"
+    return :appstore if type == "appstore"
+    return :development
+  end
+
+  def self.cert_type_sym(type)
+    return :enterprise if self.type_is_enterprise? type
+    return :development if type == "development"
+    return :distribution
+  end
 end
--- a/match/lib/match.rb
+++ b/match/lib/match.rb
@@ -26,10 +26,12 @@ module Match
     return envs
   end
 
+  # @return [Boolean] returns true if the unsupported enterprise mode should be enabled
   def self.enterprise?
     ENV["MATCH_FORCE_ENTERPRISE"]
   end
 
+  # @return [Boolean] returns true if match should interpret the given [certificate|profile] type as an enterprise one
   def self.type_is_enterprise?(type)
     Match.enterprise? && type != "development"
   end
--- a/match/lib/match/generator.rb
+++ b/match/lib/match/generator.rb
@@ -36,7 +36,7 @@ module Match
     def self.generate_provisioning_profile(params: nil, prov_type: nil, certificate_id: nil, app_identifier: nil)
       require 'sigh'
 
-      prov_type = :enterprise if Match.enterprise? && ENV["SIGH_PROFILE_ENTERPRISE"] && !params[:type] == "development"
+      prov_type = Match.profile_type_sym(params[:type])
 
       profile_name = ["match", profile_type_name(prov_type), app_identifier].join(" ")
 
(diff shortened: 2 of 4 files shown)


match/lib/match.rb
```diff
@@ -29,4 +29,21 @@ def self.environments
   def self.enterprise?
     ENV["MATCH_FORCE_ENTERPRISE"]
   end
+
+  def self.type_is_enterprise?(type)
+    Match.enterprise? && type != "development"
```
KrauseFx: So running _match_ with type `adhoc` or `appstore` will both use the `enterprise` type? Is that by design?

match/lib/match.rb
```diff
@@ -29,4 +29,21 @@ def self.environments
   def self.enterprise?
     ENV["MATCH_FORCE_ENTERPRISE"]
   end
+
+  def self.type_is_enterprise?(type)
```
KrauseFx: Could you add a comment above this method and the `self.enterprise?` to describe what the method returns?

match/lib/match/generator.rb
```diff
@@ -36,7 +36,7 @@ def self.generate_certificate(params, cert_type)
     def self.generate_provisioning_profile(params: nil, prov_type: nil, certificate_id: nil, app_identifier: nil)
       require 'sigh'
 
-      prov_type = :enterprise if Match.enterprise? && ENV["SIGH_PROFILE_ENTERPRISE"] && !params[:type] == "development"
+      prov_type = Match.profile_type_sym(params[:type])
```
KrauseFx: 😍

match/lib/match/nuke.rb
```diff
@@ -47,7 +47,7 @@ def run(params, type: nil)
     # Collect all the certs/profiles
     def prepare_list
       UI.message "Fetching certificates and profiles..."
-      cert_type = type.to_sym
+      cert_type = Match.cert_type_sym type
```
KrauseFx: Please add `( )` around `type`

match/lib/match/runner.rb
```diff
@@ -60,9 +60,7 @@ def run(params)
     end
 
     def fetch_certificate(params: nil)
-      cert_type = :distribution
-      cert_type = :development if params[:type] == "development"
-      cert_type = :enterprise if Match.enterprise? && params[:type] == "enterprise"
+      cert_type = Match.cert_type_sym params[:type]
```
KrauseFx: Please add `( )` for method calls

match/lib/match/runner.rb
```diff
@@ -92,10 +90,11 @@ def fetch_certificate(params: nil)
 
     # @return [String] The UUID of the provisioning profile so we can verify it with the Apple Developer Portal
     def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier: nil)
-      prov_type = params[:type].to_sym
+      prov_type = Match.profile_type_sym params[:type]
```
KrauseFx: Please add `( )` for method calls

match/lib/match/runner.rb
```diff
@@ -105,7 +104,15 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
       end
 
       if profile.nil? or params[:force]
-        UI.user_error!("No matching provisioning profiles found and can not create a new one because you enabled `readonly`") if params[:readonly]
+        if params[:readonly]
+          all_profiles = Dir.entries(base_dir).reject { |f| f.start_with? "." }
+          UI.error "No matching provisioning profiles found for '#{profile_name}'"
+          UI.error "A new one cannot be created because you enabled `readonly`"
+          UI.error "Certificates in your repo for type `#{prov_type}`:"
```
KrauseFx: This should be `Provisioning profiles` instead of `Certificates`

KrauseFx (changes requested):
This is awesome, I love how we move the type detection out to its central place, and how we now show additional debug information when something fails  🚀
I added some notes before it's ready to be merged

match/lib/match.rb
milch: Previously there were two different checks: `type != "development"` and `type == "enterprise"`. I chose to go with the more permissive one.

KrauseFx (approved):

KrauseFx:
Great changes, thanks 👍

Issue #6711, opened 2016-10-25
Spaceship::TunesClient::ITunesConnectError: invalid request body


summera:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Receiving `Spaceship::TunesClient::ITunesConnectError: invalid request body` when calling `Spaceship::Tunes.select_team`. The request to set the team id on [line 88](https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/tunes/tunes_client.rb#L88) is returning HTTP 400. Here is the response body:

```ruby
{
  "data"=>nil,
  "messages"=>{"warn"=>nil, "error"=>["invalid request body"], "info"=>nil}, 
  "statusCode"=>"ERROR"
}
```

I did some digging with switching teams in ITC and it looks like Spaceship needs the `dsId` param in the `ra/v1/session/webSession`. Once I added this param to the request, I received a 200 and everything worked as expected. 

```ruby
response = request(:post) do |req|
  req.url "ra/v1/session/webSession"
  req.body = { contentProviderId: team_id, dsId: 'mydsid' }.to_json
  req.headers['Content-Type'] = 'application/json'
end
```

Note: by the time i got this posted there was a duplicate: #6710 . sorry about that

ldiqual:
@summera I'm sure the fastlane team will react promptly, but in the meantime is there a workaround besides patching the gem?

summera:
I'm not quite sure what the dsId is yet because I can't even load ITC right now haha. Maybe its the user's ID? So I'm not sure if Spaceship provides a way to obtain this ID already. As far as I can tell this needs to be in the request for it to work now. So immediate fix would be patching.

chakming:
@summera ITC seems updated, have a new login page layout now

summera:
@chakming yea ITC was telling me it was down for a few and now login page looks a little different. I think everything else seems to be the same and error still exists though.

dsId is different for each user. If you go to your Personal Details for your account it's the id at the end of that URL... https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/ng/users_roles/#{dsid}

chakming:
@summera found one interesting thing: if I use another `dsId` which is not refer to my user account, it still works

summera:
[...]

Issue #6710, opened 2016-10-25
Deliver fails with `invalid request body`


ldiqual:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

When trying to deliver my app to iTunes connect, I get the following error:

```
invalid request body
```

Deliver is used as part of a lane, and I'm retrying using `bundle exec deliver`.

##### Complete output when running fastlane, including the stack trace and command used

```
[ldiqual@ldiqual-mbp-2 app]$ bundle exec deliver --verbose
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
Successfully loaded Appfile at path '/Users/ldiqual/Company/app/fastlane/Appfile'
-------
INFO [2016-10-24 19:25:42.01]: Successfully loaded '/Users/ldiqual/Company/app/fastlane/Deliverfile' 📄

+-----------------------+-----------------------+
| Detected Values from './fastlane/Deliverfile' |
+-----------------------+-----------------------+
| skip_screenshots      | true                  |
| skip_metadata         | true                  |
| force                 | true                  |
+-----------------------+-----------------------+

INFO [2016-10-24 19:25:42.01]: Login to iTunes Connect (lois@company.com)
Looking for iTunes Connect Team with ID 123456

Looking for related GitHub issues on fastlane/fastlane...
Search query: The request could not be completed because:
    invalid request body

URL: https://api.github.com/search/issues?q=The%20request%20could%20not%20be%20completed%20because:%0A%09invalid%20request%20body+repo:fastlane/fastlane
➡️  pilot crashes on upload
   https://github.com/fastlane/fastlane/issues/4821 [open] 23 💬
   3 days ago

➡️  Unable to log in to iTunes Connect - returns HTML
[...]

PR #6715, opened 2016-10-25
Send dsId in body when setting team_id

summera:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:
- [ ] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [ ] Run `bundle exec rubocop -a` to ensure the code style is valid
- [ ] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Before submitting a pull request, we appreciate if you create an issue first to discuss the change :+1:

Fixes https://github.com/fastlane/fastlane/issues/6711

PR #6716, opened 2016-10-25
Implement dsIS to iTunes Connect team selection

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/6711
Fixes https://github.com/fastlane/fastlane/issues/6710

Thanks @summera for providing a solution for the problem

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -87,7 +87,10 @@ module Spaceship
 
       response = request(:post) do |req|
         req.url "ra/v1/session/webSession"
-        req.body = { contentProviderId: team_id }.to_json
+        req.body = {
+          contentProviderId: team_id,
+          dsId: Time.now.to_i.to_s # https://github.com/fastlane/fastlane/issues/6711
+        }.to_json
         req.headers['Content-Type'] = 'application/json'
       end
 
--- a/spaceship/lib/spaceship/tunes/user_detail.rb
+++ b/spaceship/lib/spaceship/tunes/user_detail.rb
@@ -2,9 +2,11 @@ module Spaceship
   module Tunes
     class UserDetail < TunesBase
       attr_accessor :content_provider_id
+      attr_accessor :ds_id
 
       attr_mapping(
-        'contentProviderId' => :content_provider_id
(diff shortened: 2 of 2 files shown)


webpolis:
@KrauseFx you rock!

milch (approved):
👍

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -87,7 +87,10 @@ def team_id=(team_id)
 
       response = request(:post) do |req|
         req.url "ra/v1/session/webSession"
-        req.body = { contentProviderId: team_id }.to_json
+        req.body = {
+          contentProviderId: team_id,
+          dsId: Time.now.to_i.to_s # https://github.com/fastlane/fastlane/issues/6711
```
summera: @KrauseFx don't you want to call `user_detail_data.ds_id` here

KrauseFx: @summera it works with a random number too, right? By fetching the actual ds_id the request is a slower, as it has to fetch all the Team IDs.

KrauseFx: What do you think?

KrauseFx:
Moved to https://github.com/fastlane/fastlane/pull/6715

spaceship/lib/spaceship/tunes/tunes_client.rb
milch: Let's keep an eye on this, the fact that it works with a random number might point to a bug or a workaround at 's end.

summera: @milch my thoughts exactly. may work now, but not in future. i think it's worth doing right if it doesn't slow things down too much, which I don't think it does. @KrauseFx merged in #6715 which uses the dsId from user details

KrauseFx: Great think, I merged your PR, much cleaner. Thanks for your contribution. Are you on Twitter?

summera: Of course! Glad to help. Thanks for the quick response and fast merge to fix. 

I am: https://twitter.com/summ3ra

Issue #5824, opened 2016-08-21
Allow gym's build and export to each run in isolation


richardszalay:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

🔑 I'd like to propose a change to gym that would allow build and export to run at different times.

It would involve two change:
1. Adding a skip_export option
2. Allowing the output type (iOS, tv) to be passed in as an option (ie so there's no dependency on the project still being around)

TKBurner:
Hi @richardszalay Thanks for reaching out about this. I marked this as a feature request to see if the community is interested in this functionality. Thanks!

richardszalay:
Thanks @TKBurner - I'm happy to look at doing a PR for it.

If you're interested in the usecase, checkout the [blog series I just published](https://blog.richardszalay.com/2016/08/22/ios-deploy-pipeline-1-introduction/) (though I don't mention this issue until the "Closing Thoughts" in part 4.

TKBurner:
@richardszalay Very cool! I think it makes sense to add these. If you are interested go for it. We are a little behind on merging PRs so I want to set the expectation that it might be slow going once you prepare something. By making sure you have good test coverage we can hopefully merge faster!

richardszalay:
I'm working on the PR :)

TKBurner:
@richardszalay Awesome! 👍 We'll work with you there .

nubbel:
It is pretty common to build the archive once and export multiple ipas, i.e. one for adhoc, one for in-house, and one for App Store distribution.
I would love to have to separate actions for that `gym_archive` and `gym_export`.

Better yet, gym would detect that the archive has already been built and skips that step. This would allow me to define the following sequence of actions:
```ruby
gym output_name: 'adhoc', export_method: 'ad-hoc'
gym output_name: 'inhouse', export_method: 'enterprise'
gym output_name: 'appstore', export_method: 'appstore'
```
Only the first invocation of `gym` would build the archive, the others would just export ipas from it.

@richardszalay, @TKBurner: What do you think?

richardszalay:
[...]

PR #6726, opened 2016-10-25
gym now supports a skip_export option

richardszalay:

Relates to #5824. I ended up only adding the skip_archive option. The project dependency was just too ingrained to take out for such a small feature.

### DIFF
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -127,7 +127,21 @@ module Gym
                                      env_name: "GYM_SKIP_BUILD_ARCHIVE",
                                      description: "Export ipa from previously build xarchive. Uses archive_path as source",
                                      is_string: false,
-                                     optional: true),
+                                     optional: true,
+                                     conflicting_options: [:skip_export],
+                                     conflict_block: proc do |value|
+                                       UI.user_error!("'#{value.key}' must be false to use 'skip_build_archive'")
+                                     end),
+        FastlaneCore::ConfigItem.new(key: :skip_export,
+                                     env_name: "GYM_SKIP_EXPORT",
+                                     description: "Builds and generates an archive, but does not export",
+                                     is_string: false,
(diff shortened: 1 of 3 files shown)


KrauseFx:
Hey @richardszalay, thanks for the pull request. Could you quickly clarify on why this feature is useful for fastlane users? 👍

richardszalay:
It's useful if you need to build once and then export a number of times later with different configuration (think changing theme, or using a different API base url, etc)

Previously I was achieving this via unsigned IPAs, but since the XCode 7 export+plist tooling doesn't appear to support that I'm looking into using xcarchive as a build artifact instead.

KrauseFx:
@richardszalay and how do you export the archive later on? Do you use the `xcodebuild` action? If so, you could also use the `xcodebuild` action for building your app, right?

KrauseFx:
From the project page

> gym builds and packages iOS and macOS apps for you. It takes care of all the heavy lifting and makes it super easy to generate a signed ipa or app file 💪

So the primary goal is to generate signed ipa files, that's why I'm a little concerned with adding this functionality.

richardszalay:
That's fair, but gym couples both archive and export steps and only allows skipping the archive. There's no fastlane action that simplifies the archive part of the process, and the easiest way around that is to use gym and skip the export step.

richardszalay:
You use gym again at a future time, passing `skip_archive`.

ohwutup:
Hey @richardszalay, thanks for the PR! We have made a bunch of changes to this area of the code in _gym_ in addition to the fact that we have also bundled all of the gems into one, so some of the behavior has changed. I am going to close this issue for now as it is a bit outside of the spectrum of supported use cases for _gym_ and the conflicts that are now present. Feel free to submit a new PR or comment on here to continue conversation if I overlooked something! Thanks again for your contribution! 🚀

PR #6735, opened 2016-10-25
Added exporting to PKCS12

punksta:
Cert stores private key with `.p12` extension. But `.p12` is not used to store raw private keys. p12 means PKCS12. It's secure archive which stores private key and cert. https://en.wikipedia.org/wiki/PKCS_12

I think that cert can make 3 files: private key and certificate for importing to key-chain, storing and pkcs12. The last one for secure sending to another mac, storing in a cloud like s3 or some public storage. PKCS12 is common standard of sending private keys and digital certificate in the web.

### DIFF
--- a/cert/lib/cert/options.rb
+++ b/cert/lib/cert/options.rb
@@ -54,7 +54,14 @@ module Cert
                                      verify_block: proc do |value|
                                        value = File.expand_path(value)
                                        UI.user_error!("Keychain not found at path '#{value}'") unless File.exist?(value)
-                                     end)
+                                     end),
+
+        FastlaneCore::ConfigItem.new(key: :p12_password,
+                                     short_option: "-p",
+                                     env_name: "CERT_P12_PASSWORD",
+                                     description: "The password that will be used in PKCS #12 keypair container",
+                                     optional: true,
+                                     default_value: "")
       ]
     end
   end
--- a/cert/lib/cert/runner.rb
+++ b/cert/lib/cert/runner.rb
@@ -160,7 +160,7 @@ module Cert
 
       # Generate p12 file
       p12_key_path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.p12"))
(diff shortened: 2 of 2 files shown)


KrauseFx:
Could you clarify on why it's useful for users to have a password? Just want to be sure this will add benefits for majority of the users

punksta:
@KrauseFx You recommends to use cert when user wants to manage certificates by itself. When he has `.p12` container secured with password, he can send it to another mac or in a cloud without extra encryption. Mac keychain [can](https://www.digicert.com/ssl-support/p12-import-export-mac-mavericks-server.htm) export and import p12 files and it recommends to use password. Now you gives certificate and private key without encryption.

`cert` create next files in my pl:
- cert_id.p12 encrypted pair (added)
- cert_id.pem - private key (changes from .p12)
- cert_id.cer - certificate
- request.pem - request

You can reject this feature. But difficult not to agree that storing private keys with `.p12` extension makes confusion. Store private key with `pem` or `der` extension for compliance standards.

milch:
I'm not sure we want to merge this. While I recognize that it might be useful, it changes how `cert` has worked basically from beginning and might break existing setups.

jpokrzyk:
Just chiming in to vote this one up. I would find this useful. I'm using visual studio mobile center for CI and distribution and it needs a p12 file. It doesn't seem like there's a current way of getting one besides keychain.

milch:
Hey everyone, I'll be closing this PR for now since our concerns have not been addressed and the author has not responded in a while. Furthermore, we think this change would impact how a large group of existing users use `cert` in a way they would not expect, while only providing benefits to a small subset.

If you think this is a mistake please feel free to chime in, or even better creating a new issue to enable us to discuss this.

punksta:
In understand you. I just hope what I was clearly explained the problem: `.p12` file generated by `cert` is just raw private key, not PKCS12 container. Only you uses `p12` extension for files with raw private key.

PR #6737, opened 2016-10-25
[deliver] Fix string appending on setup when no metadata is available

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/6711#issuecomment-256044291


ohayon (approved):
👍

Issue #5662, opened 2016-08-09
Spaceship::Tunes.login Permission denied @ dir_s_mkdir - /home/webapp


summera:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

Getting `Permission denied @ dir_s_mkdir - /home/webapp` when logging into iTunes with 

``` ruby
Spaceship::Tunes.login(ENV['ITUNES_USER'], ENV['ITUNES_PASS'])
```

I'm running this command on AWS Elasticbeanstalk as the webapp user. From what I can tell, it looks like Spaceship wants to [store the cookie in a directory](https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L160) that does not exist, `/home/webapp`. 

To test this out:

``` bash
su -s /bin/bash -c "ls ~/" webapp
ls: cannot access /home/webapp/: No such file or directory
```

Stack trace:

```
/opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:252:in `mkdir': Permission denied @ dir_s_mkdir - /home/webapp (Errno::EACCES)
    from /opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:252:in `fu_mkdir'
    from /opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:226:in `block (2 levels) in mkdir_p'
    from /opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:224:in `reverse_each'
    from /opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:224:in `block in mkdir_p'
    from /opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:210:in `each'
    from /opt/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:210:in `mkdir_p'
    from gems/spaceship-0.30.0/lib/spaceship/client.rb:162:in `persistent_cookie_path'
    from gems/spaceship-0.30.0/lib/spaceship/two_step_client.rb:77:in `load_session_from_file'
    from gems/spaceship-0.30.0/lib/spaceship/client.rb:243:in `send_shared_login_request'
[...]

PR #6759, opened 2016-10-26
spaceship selects cookie path from set of preferred

summera:
Fixes https://github.com/fastlane/fastlane/issues/5662.
@KrauseFx Confirmed that this works in an AWS Elasticbeanstalk environment.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -155,6 +155,7 @@ module Spaceship
 
     def store_cookie(path: nil)
       path ||= persistent_cookie_path
+      FileUtils.mkdir_p(File.expand_path("..", path))
 
       # really important to specify the session to true
       # otherwise myacinfo and more won't be stored
@@ -162,9 +163,21 @@ module Spaceship
       return File.read(path)
     end
 
+    # Returns preferred path for storing cookie
+    # for two step verification.
     def persistent_cookie_path
-      path = File.expand_path(File.join("~", ".spaceship", self.user, "cookie"))
-      FileUtils.mkdir_p(File.expand_path("..", path))
+      if ENV["SPACESHIP_COOKIE_PATH"]
+        path = File.expand_path(File.join(ENV["SPACESHIP_COOKIE_PATH"], "spaceship", self.user, "cookie"))
+      else
+        ["~/.spaceship", "/var/tmp/spaceship", "#{Dir.tmpdir}/spaceship"].each do |dir|
+          dir_parts = File.split(dir)
+          if directory_accessible?(dir_parts.first)
+            path = File.expand_path(File.join(dir, self.user, "cookie"))
+            break
+          end
+        end
+      end
+
       return path
     end
 
@@ -405,6 +418,10 @@ module Spaceship
     end
 
     private
+
+    def directory_accessible?(path)
+      Dir.exist?(File.expand_path(path))
+    end
 
     def do_login(user, password)
       @loggedin = false
--- a/spaceship/spec/client_spec.rb
+++ b/spaceship/spec/client_spec.rb
@@ -115,4 +115,37 @@ describe Spaceship::Client do
       expect(subject.send(:log_response, :get, TestClient.hostname, response)).to be_truthy
     end
   end
+
+  describe "#persistent_cookie_path" do
+    before do
+      subject.login("username", "password")
+    end
+
+    after do
+      ENV.delete("SPACESHIP_COOKIE_PATH")
+    end
+
+    it "uses $SPACESHIP_COOKIE_PATH when set" do
+      ENV["SPACESHIP_COOKIE_PATH"] = "/custom_path"
+      expect(subject.persistent_cookie_path).to eq("/custom_path/spaceship/username/cookie")
+    end
+
+    it "uses home dir by default" do
+      allow(subject).to receive(:directory_accessible?).with("~").and_return(true)
+      expect(subject.persistent_cookie_path).to eq(File.expand_path("~/.spaceship/username/cookie"))
+    end
+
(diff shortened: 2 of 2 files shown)


KrauseFx (dismissed):

KrauseFx:
@mfurtak just pointed out, that storing this in /tmp is not a great solution. We store the cookie for 2 step verification in there, separate by account. This way, if you have 2 factor enabled, you're not asked to verify using your phone every time you run _fastlane_. We have to store it somewhere, where the cookie doesn't get lost, as it's valid for over a month, so it should persist even restarts on a Mac.

summera:
@KrauseFx @mfurtak that makes sense. How about `/var/tmp/spaceship/#{self.user}` ?

KrauseFx:
What's `/var/tmp`? Does it also get deleted when restarting the computer?

summera:
It will not get deleted when restarting. It is similar to `/tmp` with the main difference being that it is preserved across reboots. Per the [Filesystem hierarchy standard (FHS)](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard):  "Temporary files to be preserved between reboots."

There is a more detailed discussion here: http://unix.stackexchange.com/questions/30489/what-is-the-difference-between-tmp-and-var-tmp

mfurtak:
Would it make sense to delegate the responsibility of choosing the temp folder location to Ruby?

On my Mac:
```
>> require 'tmpdir'
=> true
>> Dir.tmpdir
=> "/var/folders/w6/ff6y4wtj7dx3trvb_0v1mwvc0000gp/T"
```

That folder currently contains ~300 items from many sources on my machine.

summera:
Good thinking. I tried this on beanstalk and it returned `/tmp`. Looking at the [source](https://ruby-doc.org/stdlib-2.2.0/libdoc/tmpdir/rdoc/Dir.html) it uses `/tmp` as a backup if the various ENV vars for setting the temp directory do not exist. So that may work on some systems like Mac but may not in a linux server environment. I think this would have the same problem as using `/tmp` because the [$TMPDIR](https://en.wikipedia.org/wiki/TMPDIR) env variable is meant to specify temporary storage that may be cleaned up at regular intervals and not guaranteed to persist after reboots.

I did notice that a file I wrote to `/tmp` on beanstalk remained after a deploy. The cleanup time for `/tmp` varies per system and I'm not sure what it is on beanstalk.

`/var/tmp` is pretty standard on unix systems for storing files that need to be preserved over reboots. Another option would be to allow the user to configure the cookie path as a last resort.

mfurtak:
Thanks for checking that out in your environment, @summera 👍 

Is it possible for you to set the `$TMPDIR` env variable for your environment before invoking _fastlane_/Ruby? In that case, if we use `Dir.tmpdir`, can you get that control over where things wind up?

summera:
@mfurtak No problem. Using `/tmp` as my cookie dir isn't an issue for me specifically because I'm not using two-factor. The original issue for me was that the home dir was not available in my environment. 

For those users who are using two-factor, I don't think setting $TMPDIR to a location that is not ephemeral just for fastlane to work is a great solution given that $TMPDIR, by definition, is ephemeral and other programs may depend on this.

mfurtak:
Cool. I agree with the thoughts around $TMPDIR and its meaning.

So my feeling is that the bug we try to fix here shouldn't be about selecting a single different directory, but about increasing our resiliency to not having write access to our preferred storage locations.

To that end, what do folks think about writing a `Dir.tmpdir`-like method that checks a preferred set of places, and returns a known-writable location that this data can be stored?

Another angle that needs to be considered for any change here is how we'll prevent currently authenticated users from being "signed out" by the update 👍 

Sorry to make things complicated! 😅

summera:
@mfurtak sounds reasonable to me. That way we can prevent currently authenticated users from being signed out by keeping the current path as the first default we try in the set. 

No worries at all! Better to hash this out now then merge a problematic change 👌 .

So I guess the next question is what our preferred set of locations should be.  What do you think about the following order?

1. "~/.spaceship/#{self.user}" <-- keep current path as first option
2. "/var/tmp/spacehship/#{self.user}"
3. "#{ENV['SPACESHIP_COOKIE_PATH']}/#{self.user}"
4. "#{Dir.tmpdir}/spaceship/#{self.user}"  <-- if all else fails

mfurtak:
That order sounds good to me!

summera:
@mfurtak @KrauseFx Just pushed an update. Sorry for the delay on this!

summera:
@mfurtak Just checking back in on this. Does the latest update look good to you?

mfurtak:
Thanks for the ping. This is looking good! I'm inclined to merge it. What do you think @KrauseFx?

KrauseFx:
Looks good to me 👍

mfurtak (approved):

mfurtak:
🎉 🎊  Yay! Thanks for the great contribution @summera 👍

lafiosca:
@summera @mfurtak Not sure if this is the place to mention this or if it's worth opening another issue, but when running in an environment that has no "HOME" variable set (such as AWS Lambda), it seems the expansion of "~" in step 2 fails and crashes out the program without ever getting to step 3:

```"/var/task/ruby_app/lib/vendor/ruby/2.3.0/gems/fastlane-2.5.0/spaceship/lib/spaceship/client.rb:426:in `expand_path': couldn't find HOME environment -- expanding `~' (ArgumentError)",```

I'm able to work around this by setting `SPACESHIP_COOKIE_PATH` (or `HOME`) in the Lambda environment.

KrauseFx:
@lafiosca Please submit a new issue for this 👍

lafiosca:
Sorry for the delay on this. I hadn't seen your comment until today. The issue has been submitted as #8465. Thank you.

Issue #1873, opened 2016-03-01
Pilot fails even though it finishes uploading the Build


fastlane-bot-helper:
<table>
      <tr>
        <td>
          <img src='https://avatars0.githubusercontent.com/u/11356621?v=3&s=70' width='35'>
        </td>
        <td>
          Original issue by @Ybrin - Imported from <a href='https://github.com/fastlane/pilot/issues/157'>fastlane/pilot#157</a>
        </td>
      </tr>
    </table>

I am currently using pilot together with fastlane in travis-ci and everything works fine, my builds are being uploaded to TestFlight and published to the internal Testers as it should.

But the thing I don't understand is, fastlane finishs with exit code 1 and says "You must provide an answer for this question" without even asking something.

I think this error is pilot related and has nothing to do with fastlane so I am posting this here.

This is the weird output:

```
[12:29:40]: Successfully finished processing the build
[12:29:40]: You can now tweet: 
[12:29:40]: iTunes Connect #iosprocessingtime 15 minutes
[12:29:40]: Distributing new build to testers
[12:29:43]: Variable Dump:
[12:29:43]: {:DEFAULT_PLATFORM=>:ios, :PLATFORM_NAME=>:ios, :LANE_NAME=>"ios beta_ci", :SIGH_PROFILE_TYPE=>"app-store", :LATEST_TESTFLIGHT_BUILD_NUMBER=>3, :BUILD_NUMBER=>"4", :IPA_OUTPUT_PATH=>"/Users/travis/build/KWStudios/Walk-iOS/Pods.ipa", :DSYM_OUTPUT_PATH=>"/Users/travis/build/KWStudios/Walk-iOS/Pods.app.dSYM.zip", :XCODEBUILD_ARCHIVE=>"/Users/travis/Library/Developer/Xcode/Archives/2016-03-01/Pods 2016-03-01 12.06.19.xcarchive"}
[12:29:43]: You must provide an answer for this question.
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 3    | cocoapods                           | 110         |
| 4    | match                               | 27          |
| 5    | match                               | 4           |
| 6    | latest_testflight_build_number      | 8           |
| 7    | increment_build_number              | 2           |
[...]

PR #6766, opened 2016-10-27
[spaceship] Show instructions on how to solve common TestFlight error

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/1873
Fixes https://github.com/fastlane/fastlane/issues/4002
and more

### DIFF
--- a/spaceship/lib/spaceship/tunes/build_train.rb
+++ b/spaceship/lib/spaceship/tunes/build_train.rb
@@ -151,7 +151,7 @@ module Spaceship
             # https://github.com/fastlane/fastlane/issues/1873
             # https://github.com/fastlane/fastlane/issues/4002
             error_message = [""] # to have a nice new-line in the beginning
-            error_message << "TestFlight requires you to provide the answer to the encryption question"]
+            error_message << "TestFlight requires you to provide the answer to the encryption question"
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/tunes/build_train.rb
```diff
@@ -141,7 +141,28 @@ def update_testing_status!(new_value, testing_type, build = nil)
           end
         end
 
-        result = client.update_build_trains!(application.apple_id, testing_type, data)
+        begin
+          result = client.update_build_trains!(application.apple_id, testing_type, data)
+        rescue Spaceship::TunesClient::ITunesConnectError => ex
+          if ex.to_s.include?("You must provide an answer for this question")
```
mfurtak: Is it possible that this text matches other "questions" than the encryption question? Is there any text more specific to the `ITSAppUsesNonExemptEncryption` key we can match on instead?

mfurtak (approved):
This seems good, but the text that we match on is very broad. I'd feel better if we could have more specific error text to match on, but if not, :ship: it!

KrauseFx:
@mfurtak Taking a look at this now, great feedback 👍

KrauseFx:
It took a while, and I can confirm that the error message doesn't contain more information about what value is missing.

PR #6773, opened 2016-10-27
'titleFontSize' key for fontSize of frame title

Tomoy:


Tomoy:
It adds the possibility to change the size of the font of the title, because if not is scaling it all the time and the different screens will have a different font size. Now if the user defines the titleFontSize key, it will use that one, if the total width with the fontsize is greater than the image width, will still be scaled as before.

Tomás Moyano from my iPhone

ohwutup:
Hey @Tomoy I am going to close this PR for now because we have updated a bunch of tools and CI is broken, that being said, I would be happy to merge the change, but it would be great to change `titleFontSize` to `title_font_size` to fit the rest of our convention, and just remove any unnecessary new lines/white space that causes rubocop to freak out. Thanks for being involved and making changes to make fastlane better! 
🚀

Issue #6673, opened 2016-10-22
Is there a way to deliver .app file for MacOS to iTunes Connect?


nickeyzzz:
### New Issue Checklist
- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

For me gym makes .app file for MacOS by default, no PKG file.
This issue is explained here https://github.com/fastlane/fastlane/issues/1234

As far, as I understand (am I mistaken?) there are two ways to upload app to iTunes
1. APP file itself, no installer
2. PKG file with app installer

As there are two kinds of certificate for Mac OS X:
1. Mac App Distribution
This certificate is used to code sign your app and configure a Distribution Provisioning Profile for submission to the Mac App Store.
2. Mac Installer Distribution
This certificate is used to sign your app's Installer Package for submission to the Mac App Store.

For me first way is preferable and APP file is enough.
Is "deliver" able to upload APP files instead of PKG files?

yokoe:
You can not upload .app file itself to iTunes and unfortunately gym still doesn't support exporting .pkg file.

At least in Xcode 8, we can create .pkg file from xcarchive in the same way exporting ipa file from iOS project using xcodebuild with "--exportArchive" option.
So I think very small change may add support for exporting pkg file.
In fast, I could confirm .pkg file exported by just adding "package_app" in `runner.rb`.

```
module Gym
  class Runner
    # @return (String) The path to the resulting ipa
    def run
      ...
      if Gym.project.ios? || Gym.project.tvos?
        ...
      elsif Gym.project.mac?
        package_app
        compress_and_move_dsym
        copy_mac_app
        # need_some_additional_things_here
      end
    end
```

Hope somebody to create PR, or I may make it some time later.

TKBurner:
@yokoe We'd love to see a PR from you. Let me know if you need any help getting started 👍

yokoe:
I created a small PR #6774 . This PR just adds support for exporting pkg file to gym. 
Some additional PRs for deliver or fastlane will be required to support entire pkg upload process in fastlane though.

TKBurner:
@yokoe Awesome! We'll work with you there. To set an expectation, we have had a backlog of PRs lately so it might take a little while, but we will definitely address it. I really appreciate you contributing 👍

triplef:
[...]

PR #6774, opened 2016-10-27
[gym] Add support for exporting pkg file.

yokoe:
- [ ] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [ ] Run `bundle exec rubocop -a` to ensure the code style is valid
- [ ] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

---

Related to #6673. This PR adds missing steps for generating pkg file in macOS project to gym. The steps are almost same to that for iOS project.

We can generate pkg file from xcarchive in the same way exporting ipa file in iOS project using xcodebuild with "--exportArchive" option.


ohayon (dismissed):
Thanks for the submission @yokoe!

gym/lib/gym/runner.rb
```diff
@@ -28,6 +28,8 @@ def run
 
         path
       elsif Gym.project.mac?
+        package_app
```
ohwutup: I don't think that we want to do this here for every app. There are times where this is not going to be desired behavior. Instead, I think there should be some argument passed to _gym_ to make this happen.

ohayon (changes requested):

ohwutup:
Thanks for the submission @yokoe, I think the approach is a bit too heavy handed to generate the `pkg` for every project built for macOS. Let me know what you think! 🐙

yokoe:
Thanks for the review. 

I think the steps should be linked to the `export_method` option. I've updated the source code, and now the steps never run unless `export_method` was specified explicitly.
I suppose `export_method` option is not specified in most macOS app projects because it has no effect currently, and without specifying it, gym will run in the same way, so I think this change will not do much harm.

I also considered adding a new option but it seems redundant.

gym/lib/gym/runner.rb
```diff
@@ -28,8 +28,10 @@ def run
 
         path
       elsif Gym.project.mac?
-        package_app
-        move_pkg
+        if Gym.config[:export_method]
```
ohwutup: should we at least check that the export method is of a certain kind?

yokoe:
I modified the source code and now packaging steps run only if export_method is 'app-store' or 'developer-id'.

Thanks to @triplef
https://github.com/fastlane/fastlane/issues/6673#issuecomment-256959890

ohwutup:
Hey @yokoe, apologies for the delay here... I think I am missing a bit of knowledge about this situation: Is there a case in which someone would want to generate both a `.pkg` and a `.app` in the same execution of _gym_? 

Also, regarding the `export_method` option, I think this is using the option a little bit differently than how it is used for iOS builds as this option implies the configuration in which the app should be built (i.e. Debug or Release).

ohwutup:
Hey @yokoe, I am going to close this PR as there hasn't been activity for a while and we have changed around some things now that we are only publishing one gem, _fastlane_. Feel free to make some changes as mentioned above and fix the merge conflicts and we can reopen this or you can open a new PR.

Thanks again for your contribution! 🚀

PR #6776, opened 2016-10-27
Improve grammar for gym message, and provide sample usage

mfurtak:
A fastlane usability tester ran into this problem on their very first run, and wasn't sure what to do about the error. Hopefully providing examples will help in the future 👍

### DIFF
--- a/gym/lib/gym/error_handler.rb
+++ b/gym/lib/gym/error_handler.rb
@@ -83,14 +83,22 @@ module Gym
         when /expected one of \{\}/
           print "It seems like you ran into this radar"
           print "https://openradar.appspot.com/radar?id=4952000420642816"
-          print "You can temporary use the :use_legacy_build_api option to get the build to work again"
+          print "You can temporarily use the :use_legacy_build_api option to work around the issue:"
+          print "In your Fastfile:"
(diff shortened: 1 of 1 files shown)


asfalcone (approved):

gym/lib/gym/error_handler.rb
```diff
@@ -83,14 +83,22 @@ def handle_package_error(output)
         when /expected one of \{\}/
           print "It seems like you ran into this radar"
           print "https://openradar.appspot.com/radar?id=4952000420642816"
-          print "You can temporary use the :use_legacy_build_api option to get the build to work again"
+          print "You can temporarily use the :use_legacy_build_api option to work around the issue:"
+          print "In your Fastfile:"
+          print "    gym(use_legacy_build_api: true)"
```
KrauseFx: Instead of `print`, which is just `UI.error` (see below), you could use `UI.command_output`, which would color-code the code sample.

Usually it looks great to have a space before and after the code sample.

Same for the message below. 

This of course is optional, if you feel like `UI.error` looks nice enough we can keep it 👍

mfurtak: I'll keep the spacing as-is for now. I'm not opposed to future changes though 👍

Issue #6770, opened 2016-10-27
Find the currently published build number on App Store


fotiDim:
### New Issue Checklist
- [Y] Updated fastlane to the latest version
- [Y] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

This is a feature request I suppose. I need a way to be able to get the **currently published build number** on App Store. `Latest_testflight_build_number` returns the latest build number from my uploaded builds but I need the build number of the latest published one (the one being sold on the App Store). Is there any way to achieve that?

hjanuschka:
ok i am not 100% sure if this is the most elegant way to solve it but i tested it on 5 apps apps.

```
lane :get_prod_build_nr do
    Spaceship::Tunes.login
    app = Spaceship::Tunes::Application.find("krone.at-fivexfive")
    version = app.live_version
    latest_prod_build_nr = version.raw_data["preReleaseBuild"]["buildVersion"]
    puts "#{latest_prod_build_nr}"
end

```

fotiDim:
@hjanuschka you sir are wonderful. Worked like a charm. I only had to alter `Spaceship::Tunes.login("email@icloud.com")` to include my ITC email and not ask for login interactively.

I really think this should be a fastlane feature.

hjanuschka:
nice to know it helps - you can set the right env vars to get arround the fixed email@icloud.com - fastlane should named them

hjanuschka:
@KrauseFx  i am not sure on how to add this to spaceship - so `app.live_version`  has the key's set - any idea where to start from? - after it is in spaceship - i think it would be easy  to add an action like `latest_live_build_number` - or modify `Latest_testflight_build_number` to be also capable to return the production build nr.

hjanuschka:
ok got it -  here is the PR to add spaceship support for the build_number https://github.com/fastlane/fastlane/pull/6785

TKBurner:
@fotiDim The only way to do this is through spaceship directly. Let me know if you run into any issues once you try that 👍 Have a great weekend!

TKBurner:
@hjanuschka Nice! Thanks so much for your continued help with this 👍

hjanuschka:
@fotiDim there we go here is the PR to include a fastlane action for this 👍 
https://github.com/fastlane/fastlane/pull/6790

fotiDim:
[...]

PR #6785, opened 2016-10-28
[spaceship] get current build_number from AppVersion 🚀🤘

hjanuschka:
return build_number on AppVersion.

as supposed to be required here: https://github.com/fastlane/fastlane/issues/6770
if this one is merged - i ll add a custom action to retrieve the current build number.

adds method `current_build_number` to appVersion class.

e.g.:

``` ruby
app = Spaceship::Tunes::Application.find("krone.at-fivexfive")
[2] pry(main)> version = app.live_version.current_build_number
=> "7"
[3] pry(main)> version = app.edit_version.current_build_number
=> "14"

```

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -30,6 +30,9 @@ module Spaceship
       # @return (String) App Status (e.g. 'readyForSale'). You should use `app_status` instead
       attr_accessor :raw_status
 
+      # @return (String) Build Version
+      attr_accessor :build_version
+
       # @return (Bool)
       attr_accessor :can_reject_version
 
@@ -133,6 +136,7 @@ module Spaceship
         'largeAppIcon.value.url' => :app_icon_url,
         'releaseOnApproval.value' => :release_on_approval,
         'status' => :raw_status,
+        'preReleaseBuild.buildVersion' => :build_version,
         'supportsAppleWatch' => :supports_apple_watch,
         'versionId' => :version_id,
         'version.value' => :version,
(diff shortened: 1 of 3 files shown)


spaceship/lib/spaceship/tunes/app_version.rb
```diff
@@ -215,6 +219,14 @@ def create_languages(languages)
         nil
       end
 
+      def current_build_number
+        if self.is_live?
+          build_version
+        else
+          candidate_builds.first.build_version
```
mfurtak: Will this fail if there are no `candidate_builds`?

mfurtak:
This looks great - I left one question 👍

spaceship/lib/spaceship/tunes/app_version.rb
hjanuschka: no it will not fail - it will return `nil` - tested it, is that ok? - i am not sure what else would fit as a return value?

```
[30] pry(main)> app = Spaceship::Tunes::Application.find("krone.at-krone.at")
=> <Spaceship::Tunes::Application 
	apple_id="633492143", 
	name="Krone", 
	vendor_id="krone.at-krone.at", 
	bundle_id="krone.at-krone.at", 
	last_modified=1477673411000, 
	issues_count=0, 
	app_icon_preview_url="https://is3-ssl.mzstatic.com/image/thumb/Purple71/v4/90/a2/5e/90a25ecc-b333-6827-094d-1927a1404da8/mzl.hzffjpxt.png/340x340bb-80.png">
[31] pry(main)> version = app.edit_version.build_version
=> nil
[32] pry(main)> 
```

mfurtak: Ah, sorry - I could have been more clear.

If `candidate_builds` returns an empty array, `first` will return `nil`, which will then cause an exception, right?

hjanuschka: ok got it - pushed and rebased - hope this will do it 👍  thx for your help.

mfurtak (approved):

mfurtak:
Sweet, thanks! 🚀

hjanuschka:
my first spaceship PR ! feels good 💯  thx

PR #6788, opened 2016-10-28
[scan] Properly handle report generation failures

0xced:

---

A failing report generation (e.g. when given a non-writable path) was erroneously reported as successful.

Now, an error is properly reported and the scan process exits with a non 0 status code.

### DIFF
--- a/scan/lib/scan/report_collector.rb
+++ b/scan/lib/scan/report_collector.rb
@@ -16,8 +16,11 @@ module Scan
 
       commands = generate_commands(path)
       commands.each do |output_path, command|
-        system(command)
-        UI.success("Successfully generated report at '#{output_path}'")
+        if system(command)
+          UI.success("Successfully generated report at '#{output_path}'")
+        else
+          UI.user_error!("Failed to generate report at '#{output_path}'")
+        end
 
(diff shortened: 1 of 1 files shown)


scan/lib/scan/report_collector.rb
```diff
@@ -14,16 +14,22 @@ def initialize(open_report, output_types, output_directory, use_clang_report_nam
...
+      success = true
       commands = generate_commands(path)
       commands.each do |output_path, command|
-        system(command)
-        UI.success("Successfully generated report at '#{output_path}'")
+        if system(command)
+          UI.success("Successfully generated report at '#{output_path}'")
+        else
+          UI.error("Failed to generate report at '#{output_path}'")
+          success = false
```
KrauseFx: Couldn't we just add `UI.user_error!` here?

0xced: Yes!

The only advantage of this method is that it reports all the errors for different output types. But we can stop at the first report generation error, this will be less code.

KrauseFx (approved):

KrauseFx:
This is a great change, thanks 👍

PR #6790, opened 2016-10-28
[fastlane] add current_build_number action 🎯

hjanuschka:
as mentioned  in https://github.com/fastlane/fastlane/issues/6770 returns the current build number of either live_version or edit_version  - works only with spaceship-master (as this PR is required: https://github.com/fastlane/fastlane/pull/6785)

following ui:

``` ruby
lane :demo do
  # query the appstore version (ready for sale)
  build_number = current_build_number
  puts build_number

  #query the latest testflight upload
  build_number = current_build_number(live: false)
  puts build_number
end
```

edit: i think this is a pretty common feature, therefore my PR with a new action, instead of a plugin

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -1,0 +1,149 @@
+module Fastlane
+  module Actions
+    module SharedValues
+      LATEST_BUILD_NUMBER = :LATEST_BUILD_NUMBER
+    end
+
+    class AppStoreBuildNumberAction < Action
+      def self.run(params)
+        require 'spaceship'
+
+        UI.message("Login to iTunes Connect (#{params[:username]})")
+        Spaceship::Tunes.login(params[:username])
+        Spaceship::Tunes.select_team
+        UI.message("Login successful")
+
+        app = Spaceship::Tunes::Application.find(params[:app_identifier])
+        if params[:live]
+          UI.message("Fetching the latest build number for live-version")
+          build_nr = app.live_version.current_build_number
+        else
+          version_number = params[:version]
+          unless version_number
+            # Automatically fetch the latest version in testflight
+            begin
+              testflight_version = app.build_trains.keys.last
+            rescue
+              testflight_version = params[:version]
+            end
+
+            if testflight_version
+              version_number = testflight_version
+            else
+              version_number = UI.input("You have to specify a new version number")
+            end
+
+          end
+
+          UI.message("Fetching the latest build number for version #{version_number}")
+
+          begin
+            train = app.build_trains[version_number]
+            build_nr = train.builds.map(&:build_version).map(&:to_i).sort.last
+          rescue
+            UI.user_error!("could not find a build on iTC - and 'initial_build_number' option is not set") unless params[:initial_build_number]
+            build_nr = params[:initial_build_number]
+          end
+        end
+        UI.message("Latest upload is build number: #{build_nr}")
+        Actions.lane_context[SharedValues::LATEST_BUILD_NUMBER] = build_nr
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Returns the current build_number of either live or edit version"
+      end
+
+      def self.available_options
+        user = CredentialsManager::AppfileConfig.try_fetch_value(:itunes_connect_id)
+        user ||= CredentialsManager::AppfileConfig.try_fetch_value(:apple_id)
+        [
+          FastlaneCore::ConfigItem.new(key: :initial_build_number,
+                                       env_name: "INITIAL_BUILD_NUMBER",
+                                       description: "sets the build number to given value if no build is in current train",
+                                       default_value: 1,
+                                       is_string: false),
+          FastlaneCore::ConfigItem.new(key: :app_identifier,
+                                       short_option: "-a",
+                                       env_name: "FASTLANE_APP_IDENTIFIER",
+                                       description: "The bundle identifier of your app",
+                                       default_value: CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier)),
+          FastlaneCore::ConfigItem.new(key: :username,
+                                       short_option: "-u",
+                                       env_name: "ITUNESCONNECT_USER",
+                                       description: "Your Apple ID Username",
+                                       default_value: user),
+          FastlaneCore::ConfigItem.new(key: :team_id,
+                                       short_option: "-k",
+                                       env_name: "APPSTORE_BUILD_NUMBER_LIVE_TEAM_ID",
+                                       description: "The ID of your iTunes Connect team if you're in multiple teams",
+                                       optional: true,
+                                       is_string: false, # as we also allow integers, which we convert to strings anyway
+                                       default_value: CredentialsManager::AppfileConfig.try_fetch_value(:itc_team_id),
+                                       verify_block: proc do |value|
+                                         ENV["FASTLANE_ITC_TEAM_ID"] = value.to_s
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :live,
+                                       short_option: "-l",
+                                       env_name: "APPSTORE_BUILD_NUMBER_LIVE",
+                                       description: "Query the live version (ready-for-sale)",
+                                       optional: true,
+                                       default_value: true),
+          FastlaneCore::ConfigItem.new(key: :version,
+                                       env_name: "LATEST_VERSION",
+                                       description: "The version number whose latest build number we want",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :team_name,
+                                       short_option: "-e",
+                                       env_name: "LATEST_TESTFLIGHT_BUILD_NUMBER_TEAM_NAME",
+                                       description: "The name of your iTunes Connect team if you're in multiple teams",
+                                       optional: true,
+                                       default_value: CredentialsManager::AppfileConfig.try_fetch_value(:itc_team_name),
+                                       verify_block: proc do |value|
+                                         ENV["FASTLANE_ITC_TEAM_NAME"] = value.to_s
+                                       end)
+        ]
+      end
+
+      def self.output
+        [
+          ['LATEST_BUILD_NUMBER', 'The latest build number of either live or testflight version']
+        ]
+      end
+
+      def self.details
+        "Returns the current build number of either the live or testflight version - it is useful for getting the build_number of the current or ready-for-sale app version, and it also works on non-live testflight version. If you need to handle more build-trains please see `latest_testflight_build_number`"
+      end
+
+      def self.example_code
+        [
+          'app_store_build_number',
+          'app_store_build_number(
+            app_identifier: "app.identifier",
+            username: "user@host.com"
+          )',
+          'app_store_build_number(
+            live: false,
+            app_identifier: "app.identifier",
+            version: "1.2.9"
+          )'
+        ]
+      end
+
+      def self.authors
+        ["hjanuschka"]
+      end
+
+      def self.category
+        :misc
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
+      end
+    end
+  end
+end
--- a/fastlane/lib/fastlane/actions/latest_testflight_build_number.rb
+++ b/fastlane/lib/fastlane/actions/latest_testflight_build_number.rb
@@ -8,44 +8,7 @@ module Fastlane
 
     class LatestTestflightBuildNumberAction < Action
       def self.run(params)
-        require 'spaceship'
-
-        credentials = CredentialsManager::AccountManager.new(user: params[:username])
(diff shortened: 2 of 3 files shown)


KrauseFx:
```ruby
  #query the latest testflight upload
  build_number = current_build_number(live: false)
  puts build_number
```

Looking at this `current_build_number(live: false)` is the same as `latest_testflight_build_number`? Or is there any differences? 

Other than that I like the action and I feel like this is useful to many users 👍

fastlane/docs/Actions.md
```diff
@@ -6458,6 +6458,47 @@ Key | Description
 
 
 
+### current_build_number
+
+Fetches the current build number from either production or testflight train.
```
KrauseFx: Should be `TestFlight`

fastlane/lib/fastlane/actions/current_build_number.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def self.example_code
+        [
+          'current_build_number'
+        ]
+      end
+
+      def self.authors
+        # So no one will ever forget your contribution to fastlane :) You are awesome btw!
```
KrauseFx: You can remove this line

fastlane/lib/fastlane/actions/current_build_number.rb
```diff
@@ -0,0 +1,102 @@
...
+        Actions.lane_context[SharedValues::LATEST_BUILD_NUMBER] = build_nr
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Returns the current build_number of either live or edit version"
+      end
```
KrauseFx: Can you also provide `self.details` with more details, in particular when to use this and when to use `latest_testflight_build_number`

fastlane/spec/unused_options_spec.rb
```diff
@@ -1,7 +1,7 @@
 describe Fastlane do
   describe Fastlane::Action do
     describe "No unused options" do
-      let (:all_exceptions) { %w(pilot appstore cert deliver gym match pem produce scan sigh snapshot supply testflight mailgun testfairy ipa import_from_git hockey deploygate crashlytics artifactory appledoc slather screengrab download_dsyms notification frameit set_changelog register_devices latest_testflight_build_number) }
+      let (:all_exceptions) { %w(pilot appstore cert deliver gym match pem produce scan sigh snapshot supply testflight mailgun testfairy ipa import_from_git hockey deploygate crashlytics artifactory appledoc slather screengrab download_dsyms notification frameit set_changelog register_devices latest_testflight_build_number current_build_number) }
```
KrauseFx: Why does this action have an unused action?

KrauseFx (dismissed):

hjanuschka:
yeah its pretty much the same return value as in `latest_testflight_build_number` -first i tried to modify the `latest_testflight_build_number` - to return the build_number from live_version.
but it would be too much spagetti. so i decided to do a new action, and found that it maybe usefull to query both variations, as the spaceship extension is on `AppVersion` 👍

KrauseFx:
How about we call the new action `current_build_number` from the TestFlight action so that we have only one code to maintain?

fastlane/lib/fastlane/actions/current_build_number.rb
hjanuschka: done

fastlane/docs/Actions.md
hjanuschka: done

fastlane/spec/unused_options_spec.rb
hjanuschka: the `:team_id` is only used in the verify_block, wich is not detected by the spec.

```
verify_block: proc do |value|
                                         ENV["FASTLANE_ITC_TEAM_NAME"] = value.to_s
                                       end

```

hjanuschka:
thats a great idea, i am working on it!

hjanuschka:
@KrauseFx  - ok  i am  a bit unsure of the refactor, as the testflight variation uses the build_trains - and i am a bit scared that it will break too much.

hjanuschka:
@KrauseFx  ok after a few hours of sleeping, i think i got a solution, will refactor both actions tonight - to share as much code as possible - without breaking BC

hjanuschka:
@KrauseFx  ok refactored the `latest_testflight_build_number` to be basically like an alias (with a forced `live: false`, and a bc sharedValue) - i think it should'nt break anything.


```
lane :demo do
  b = current_build_number(app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com")
  puts b

  b = current_build_number(live: false, app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com")
  puts b

  latest_testflight_build_number(app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com")
  
  b = current_build_number(live: false, app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com", version: "1.2.9")
  puts b
  
  latest_testflight_build_number(app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com", version: "1.2.9")

end

```

tested with the lane above, and it did work :)

fastlane/lib/fastlane/actions/current_build_number.rb
```diff
@@ -0,0 +1,142 @@
...
+
+      def self.output
+        [
+          ['LATEST_BUILD_NUMBER', 'The latest build number of either live or testflight version']
+        ]
+      end
+
+      def self.details
+        "Returns the current build number of either the live or testflight version - it is usefull for getting the build_number of the current, ready-for-sale, app version, it also works on non-live testflight version. If you need to handle more build-trains please see `latest_testflight_build_number`"
```
milch: `usefull` -> `useful`

milch: `current, ready-for-sale, app version, it also works` -> `current or ready-for-sale app version, and it also works`

fastlane/lib/fastlane/actions/current_build_number.rb
```diff
@@ -0,0 +1,142 @@
...
+              testflight_version = app.build_trains.keys.last
+            rescue
+              testflight_version = params[:version]
+            end
+
+            if testflight_version
+              version_number = testflight_version
+            else
+              UI.message("You have to specify a new version number: ")
+              version_number = STDIN.gets.strip
```
milch: Use UI.input [as described here](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/UI.md)

fastlane/lib/fastlane/actions/current_build_number.rb
```diff
@@ -0,0 +1,142 @@
...
+      def self.authors
+        ["hjanuschka"]
+      end
+
+      def self.category
+        :misc
+      end
+
+      def self.is_supported?(platform)
+        true
```
milch: `is_supported?` is supposed to return true for the action target's platform - so in this case `[:ios, :mac].include?(platform)` is probably more appropriate

milch (changes requested):
In general, I don't like the name `current_build_number`. Just from that name I'd guess that it's only fetching the current build number from the Xcode project or whatever.

fotiDim:
Maybe in order to be consistent with `latest_testflight_build_number` it should be named `latest_app_store_build_number` or `latest_live_build_number`. 

Just a small question though... What if there is a build in `pending developer release` state? Would this build number be returned or the public one?

hjanuschka:
* @milch feedback done! ( i love the gone `Actions.md` 🎉 )
  * rebased current master and squased all commits

about the name how about `app_store_build_number` - as it is not always the latest! (you can set `:version`  for a specific train)

@KrauseFx  can we get a core-team decision about the name? - so i can refactor it?


@fotiDim if your release is pending dev release - it will return the build number from published one, as of the `state` - the "pending dev release` is not live

milch:
@hjanuschka Thanks! 

I like `app_store_build_number`, it makes it pretty clear that this is not something that is read from a local file or something.

fastlane/lib/fastlane/actions/current_build_number.rb
```diff
@@ -0,0 +1,141 @@
+module Fastlane
+  module Actions
+    module SharedValues
+      LATEST_BUILD_NUMBER = :LATEST_BUILD_NUMBER
+    end
+
+    class CurrentBuildNumberAction < Action
+      def self.run(params)
+        require 'spaceship'
+        credentials = CredentialsManager::AccountManager.new(user: params[:username])
```
KrauseFx: This is not necessary, check out the other actions like `DownloadDsymsAction`. I just checked, and this code has been here before, but let's get rid of it and do like `DownloadDsymsAction`

fastlane/lib/fastlane/actions/latest_testflight_build_number.rb
```diff
@@ -69,6 +33,12 @@ def self.available_options
         user ||= CredentialsManager::AppfileConfig.try_fetch_value(:apple_id)
 
         [
+          FastlaneCore::ConfigItem.new(key: :live,
```
KrauseFx: This value gets overwritten in any case, so that's not good

fastlane/lib/fastlane/actions/latest_testflight_build_number.rb
```diff
@@ -69,6 +33,12 @@ def self.available_options
...
+                                       short_option: "-l",
+                                       env_name: "CURRENT_BUILD_NUMBER_LIVE",
+                                       description: "Query the live version (ready-for-sale)",
+                                       optional: true,
+                                       default_value: true),
```
KrauseFx: We overwrite the default value?

KrauseFx: Let's do it like this: Make the default value `false`, and not set it on line 11 (just remove line 11)

KrauseFx (dismissed):

KrauseFx:
Thanks everyone, this looks great already, I added some more comments before this is ready 👍

hjanuschka:
what a PR! 😵  - thx for all your help! 🎉 

  * rebased current master, squashed commits
  * @KrauseFx  done the feedback
  * renamed action to `app_store_build_number`


test lanes:

```ruby
lane :demo do
  b = app_store_build_number(app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com")
  puts b

  b = app_store_build_number(live: false, app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com")
  puts b

  latest_testflight_build_number(app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com")
  
  b = app_store_build_number(live: false, app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com", version: "1.2.9")
  puts b
  
  latest_testflight_build_number(app_identifier: "krone.at-fivexfive", username: "helmut@januschka.com", version: "1.2.9")

end

```

fastlane/lib/fastlane/actions/latest_testflight_build_number.rb
hjanuschka: done  - with default value `false` and not overwriting it, basically the `latest_testflight_build_number` - should be deprecated - but should work as a BC shim!

milch (approved):

fastlane/lib/fastlane/actions/app_store_build_number.rb
```diff
@@ -0,0 +1,141 @@
...
+        ]
+      end
+
+      def self.details
+        "Returns the current build number of either the live or testflight version - it is useful for getting the build_number of the current or ready-for-sale app version, and it also works on non-live testflight version. If you need to handle more build-trains please see `latest_testflight_build_number`"
+      end
+
+      def self.example_code
+        [
+          'appstore_build_number'
```
KrauseFx: There should be a second example with more complex parameters, similar to what we do in most other actions: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/sigh.rb#L71-L80

KrauseFx (changes requested):

fastlane/lib/fastlane/actions/app_store_build_number.rb
```diff
@@ -0,0 +1,141 @@
...
+      def self.run(params)
+        require 'spaceship'
+
+        UI.message("Login to iTunes Connect (#{params[:username]})")
+        Spaceship::Tunes.login(params[:username])
+        Spaceship::Tunes.select_team
+        UI.message("Login successful")
+
+        app = Spaceship::Tunes::Application.find(params[:app_identifier])
+        build_nr = nil
```
KrauseFx: I don't think we need to assign this to `nil`, right?

hjanuschka:
@KrauseFx issues have been addressed.

KrauseFx (approved):

milch:
Thanks 👍

fotiDim:
Great work guys! When do we expect a new fastlane release that includes those changes?

hjanuschka:
@fotiDim  1.107.0 has it -> https://github.com/fastlane/fastlane/releases/tag/1.107.0 🎉

PR #6800, opened 2016-10-31
[spaceship] Add update check to ensure people are running the latest release

KrauseFx:
screenshot 2016-10-30 15 31 17

We’ve had many issues around users not using the latest version of spaceship.
While we show update messages in the tools themselves, spaceship didn’t show a warning before. This way we also had to push new releases for all tools depending on spaceship

We don't use the update manager from _fastlane_core_ as _spaceship_ doesn't have a dependency to it.

### DIFF
--- a/spaceship/lib/spaceship/update_checker.rb
+++ b/spaceship/lib/spaceship/update_checker.rb
@@ -18,7 +18,7 @@ module Spaceship
 
       show_update_message(Spaceship::VERSION, version)
     rescue => ex
-      puts "#{ex}" if $verbose
+      puts ex.to_s if $verbose
       puts "Couldn't verify that spaceship is up to date"
     end
 
--- a/spaceship/lib/spaceship/update_checker.rb
+++ b/spaceship/lib/spaceship/update_checker.rb
@@ -31,8 +31,8 @@ module Spaceship
       puts ""
       puts "or `bundle update` if you use bundler."
       puts ""
-      puts "You're on spaceship: #{local_version}".yedllow
-      puts "Latest spaceship   : #{live_version}".yellow
+      puts "You're on spaceship version: #{local_version}".yellow
+      puts "Latest spaceship version  : #{live_version}".yellow
       puts ""
     end
   end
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/update_checker.rb
```diff
@@ -0,0 +1,39 @@
...
+      return if defined?(SpecHelper) # is this running via tests
+      return if ENV["FASTLANE_SKIP_UPDATE_CHECK"]
+
+      require 'faraday'
+      require 'json'
+
+      response = Faraday.get(UPDATE_URL)
+      return if response.nil? || response.body.to_s.length == 0
+
+      version = JSON.parse(response.body)["version"]
```
ohwutup: Just curious here: why don't we check RubyGems for the most up to date version here instead of _refresher_?

KrauseFx:
Update: this needs to work better when we have no internet connection

KrauseFx:
This is ready now  🚀

spaceship/lib/spaceship/update_checker.rb
```diff
@@ -0,0 +1,42 @@
...
+      puts "---------------------------------------------".red
+      puts "You're using an old version of spaceship"
+      puts "To ensure spaceship and fastlane works"
+      puts "update to the latest version."
+      puts ""
+      puts "Run `[sudo] gem update spaceship`"
+      puts ""
+      puts "or `bundle update` if you use bundler."
+      puts ""
+      puts "You're on spaceship: #{local_version}".yedllow
```
milch: `.yedllow`

spaceship/lib/spaceship/update_checker.rb
```diff
@@ -0,0 +1,42 @@
...
+      puts "Latest spaceship   : #{live_version}".yellow
```
milch: Also might be worthwhile to say "[Latest|You're on] spaceship version" for these two lines, doesn't hurt to be verbose.

milch (changes requested):
⛔

KrauseFx:
@milch: Perfect, thanks. Fixed 👍

milch (approved):

Issue #5758, opened 2016-08-15
Full support for incrementing version number in Settings.bundle


jdee:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

This is a feature request:

It would be nice if fastlane provided full support for including the current version and build numbers in a settings bundle. I don't find any support for this in the increment_build_number or increment_version_number actions. There is a bit of support for including Settings.bundle/Root.plist in commit_version_bump (https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/commit_version_bump.rb#L93), but it requires the Settings.bundle to reside at the top of the repo. It also requires force: true, since the Root.plist is not added to expected_changes until after validation.

Ideally:
- Allow specification of the user defaults key or keys used for the version and build numbers
- Allow specification of the format used if using a single key, e.g., 1.0.2. (113)
- ~~Allow specification of the location of Settings.bundle~~ Take location of Settings.bundle from project file

At the moment, I can accomplish what I need using the plist gem. This would be a nice option for the stock actions mentioned above.

vpolouchkine:
Thanks for the FR, keeping this open for upvotes.

jdee:
Actually, I suppose the Settings.bundle location needs to come from the project file. I've updated the issue accordingly.

If I can find some time, I'm tempted to submit a PR for this.

TKBurner:
@jdee That'd be super interesting!

dombarnes:
If it helps, I've got this in my Fastfile
````
def update_root_plist(params)
    require 'plist'
    begin
      path = File.expand_path(params[:path])
      plist = Plist.parse_xml(path)
      # Set the XML position of the Version item
      plist["PreferenceSpecifiers"][0]["DefaultValue"] = params[:value]
      new_plist = plist.to_plist
      File.write(path, new_plist)
      return params[:value]
    rescue => ex
      Helper.log.error ex
      Helper.log.error "Unable to set value to plist file at '#{path}'".red
    end
  end
````
and I run 
````
update_root_plist(
      path: 'Resources/Settings.bundle/Root.plist', 
      value: friendly_version
    )
````
to update the file before running gym.

jdee:
[...]

PR #6807, opened 2016-11-01
Update settings bundle

jdee:

- [ x] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [ x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x ] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x ] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Please see https://github.com/fastlane/fastlane/issues/5758

This seems like a candidate for core Fastlane functionality, closely related to increment_version_number, increment_build_number and commit_version_bump. It basically just extends agvtool. It writes a new parameter to the lane context for the benefit of commit_version_bump. I had the impression from the discussion in the issue that this should be a new action for Fastlane, or possibly even an option to increment_version_number and/or increment_build_number. So while I understand you're not accepting new actions, I'm opening this PR anyway. Please close it if this is not the right thing to do, and I'll set it up as a plugin.

This just supports some basic functionality:

```ruby
update_settings_bundle path: "Resources/Settings.bundle/Root.plist, setting_key: "CurrentAppVersion"
```

See the spec example.

It takes the current version and build numbers from the lane context and formats them as "#{version} (#{build})", updating the key in the specified plist file. It also records the supplied path in the SharedValues::SETTINGS_PLIST_PATH variable in the lane context.

Still to do:

[ ] Modify the commit_version_bump action to read the SETTINGS_PLIST_PATH from the lane context and include it in expected changes.
[ ] Finish off unit tests.

Some possible improvements:

- Use the xcodeproj gem to get the Settings.bundle location from the project file.
- Support other formats or possibly just allow overriding the values from the lane context, like:

```ruby
version = Actions.lane_context[Actions::SharedValues::VERSION_NUMBER]
build = Actions.lane_context[Actions::SharedValues::BUILD_NUMBER]
setting_value = "#{version}-#{build}"
update_settings_bundle path: ".../Root.plist", key: "CurrentAppVersion", value: setting_value
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ rspec_logs.json
 test_output/
 .DS_STORE
 .keys
+*.swp
 
 ## Specific to Android
 .gradle/
--- a/fastlane/docs/Actions.md
+++ b/fastlane/docs/Actions.md
@@ -2345,10 +2345,37 @@ Key | Description
 
 </details>
 
+### update_settings_bundle
 
+Updates a key in a plist file in the Settings.bundle to the current version
+and build number.
 
+> Run after increment_version_number and/or increment_build_number. Takes the updated version number and modifies the app's settings bundle.
 
+update_settings_bundle |
+-----|----
+Supported platforms | ios
+Author | @jdee
 
+<details>
+<summary>1 Example</summary>
+
+```ruby
+udpate_settings_bundle(
+  path: "Resources/Settings.bundle/Root.plist",
+  setting_key: "CurrentAppVersion"
+)
+```
+</details>
+
+<details>
+<summary>Parameters</summary>
+Key | Description
+----|------------
+  `path` | The path to the plist file in the Settings.bundle. Does not have to be the Root.plist.
+  `setting_key` | The identifier of the preference specifier to update
+
+</details>
 
 # Code Signing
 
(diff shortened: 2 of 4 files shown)


jdee:
I think the xcodeproj integration will be necessary or at least helpful. The lane context approach for version and build numbers doesn't work well unless you update both. The only thing that makes sense is to read the current versions from the project and/or Info.plist after updating. So you can have:

```ruby
update_settings_bundle path: ".../Root.plist", info_plist: ".../Info.plist", setting_key: "CurrentAppVersion"
```

Or just:

```ruby
update_settings_bundle path xcodeproj: "MyApp.xcodeproj", setting_key: "CurrentAppVersion"
```

I think it makes most sense to pull this all out of the project using xcodeproj, rather than having the user pass the paths in for each file. I don't think that's a lot of trouble.

jdee:
In fact, if the update_settings_bundle action is just going to get the Settings.bundle location from the project file, then commit_version_bump can do the same thing, and there's no need to pass the information in the lane context. This could strengthen the case for making this a plugin. I can see arguments for doing it either way. I'll wait for some feedback from community/maintainers before making any further changes.

milch:
Thank you for your contribution! 

However, I feel like adding this as an action would violate the `We only accept actions that are generally usable for a big majority of developers` rule. Please make a plugin instead.

Issue #6136, opened 2016-09-15
Pilot - Is there a way to add a tester to a Group?


bpham711:
### New Issue Checklist
- [ x] Updated fastlane to the latest version
- [x ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

We have many different groups and I'd like a way to add new testers to specific groups in TestFlight. Is there a way to do that with Pilot?

TKBurner:
@bpham711 _pilot_ doesn't currently have support for groups. I'm marking this as a feature request so we can see if there is interest. Thanks!

wacheski:
I am interested. Just tried to include groups in the import csv :)

TKBurner:
@wacheski Have you checked out spaceship? You might be able to get it working there.

wacheski:
@bpham711 Thanks. I don't see group support yet, but will have a closer look.

daneov:
👍

wacheski:
A beautiful thing. Works great. Thanks!

PR #6812, opened 2016-11-01
Ability to associate new testers with groups

### LINKED ISSUE #1911: User groups
<table>
      <tr>
        <td>
          
        </td>
        <td>
          Original issue by @ahknight - Imported from <a href='https://github.com/fastlane/pilot/issues/46'>fastlane/pilot#46</a>
        </td>
      </tr>
    </table>

We use the user groups functionality in iTunes Connect for our beta tests a lot.  I don't see support for that in pilot at the moment.  Am I missing it?

What we do is recruit a set of testers, add them to the iTunes account in a user group ("MyApp Beta 1") and then add the group to the app.  This way we know which tests a user is on, and also why they're on the account in the first place (we have many, many apps in the account).

Is there a way to specify a group when importing a CSV so that the user is also added to that group (and have it created if it doesn't exist)?

kobim:

Adding the ability to associate new testers with a group (either by name or id).
(Depends on #6829, Resolves #1911 and #6136)

There are 2 options for defining a group - either by command line or by additional column in the import CSV.

Currently I didn't add a support for "on-the-fly" group adding, as I'm not sure there's a common use case to that (at least not the one I'm trying to resolve, and what's reported in the above tickets).


milch (commented):
Looks good to me in principle. Does it change anything for existing setups?

kobim:
The group parameter is optional, hence it has no affect on existing setups.

pilot/lib/pilot/tester_importer.rb
```diff
@@ -13,8 +13,10 @@ def import_testers(options)
       tester_manager = Pilot::TesterManager.new
       imported_tester_count = 0
 
+      group_name = options[:group]
+
       CSV.foreach(file, "r") do |row|
-        first_name, last_name, email = row
+        first_name, last_name, email, testing_group = row
```
milch: It seems to me that this change will not be compatible with the exporter. This is how the CSV is generated:

```csv << [tester.first_name, tester.last_name, tester.email, tester.devices.count, group_names, #...```

File: pilot/lib/pilot/tester_exporter.rb [hunk 2]
```diff
@@ -23,14 +23,7 @@ def export_testers(options)
...
         testers.each do |tester|
-          groups = tester.raw_data.get("groups")
-
-          group_names = ""
-          if groups && groups.length > 0
-            names = groups.map { |group| group["name"]["value"] }
-            group_names = names.join(';')
-          end
-
+          group_names = tester.groups_list || ""
```
KrauseFx: Why do we change this here?

KrauseFx:
In case we want to go forward with this change, this would need to be split into 2 separate PRs, so we can push a _spaceship_ release before doing a _pilot_ release 👍

kobim:
@milch - Just changed and made sure both the import and export formats will be the same (groups separated by a semicolon)
@KrauseFx - I removed spaceship's modifications from this PR and opened #6829 for that.

milch (approved):

milch:
Thanks! 🚀

Issue #6784, opened 2016-10-28
gym: incompatible character encodings: ASCII-8BIT and UTF-8 


edwardlafoy:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

I keep getting encoding errors whenever I run a lane that has `gym`:

```
    gym(
      scheme: @scheme,
      clean: true
    )
```
##### Complete output when running fastlane, including the stack trace and command used

``````
INFO [2016-10-27 22:32:42.85]: ▸ Generating 'RxCocoa.framework.dSYM'
INFO [2016-10-27 22:32:46.15]: ▸ Copying _RX.h
INFO [2016-10-27 22:32:46.18]: ▸ Copying _RXDelegateProxy.h
INFO [2016-10-27 22:32:46.22]: ▸ Copying _RXKVOObserver.h
INFO [2016-10-27 22:32:46.22]: ▸ Copying _RXObjCRuntime.h
INFO [2016-10-27 22:32:46.22]: ▸ Copying RxCocoa-umbrella.h
INFO [2016-10-27 22:32:46.23]: ▸ Copying RxCocoa.h
INFO [2016-10-27 22:32:46.23]: ▸ Touching RxCocoa.framework
INFO [2016-10-27 22:32:46.31]: ▸ Building Pods/SwiftyBeaver [Release]
INFO [2016-10-27 22:32:46.31]: ▸ Check Dependencies
INFO [2016-10-27 22:32:46.32]: ▸ Processing Info.plist
INFO [2016-10-27 22:32:46.32]: ▸ Compiling AES256CBC.swift
INFO [2016-10-27 22:32:46.32]: ▸ Compiling BaseDestination.swift
INFO [2016-10-27 22:32:46.32]: ▸ Compiling ConsoleDestination.swift
INFO [2016-10-27 22:32:46.32]: ▸ Compiling FileDestination.swift
INFO [2016-10-27 22:32:46.32]: ▸ Compiling Filter.swift
INFO [2016-10-27 22:32:46.32]: ▸ Compiling SBPlatformDestination.swift
INFO [2016-10-27 22:32:46.32]: ▸ Compiling SwiftyBeaver.swift
INFO [2016-10-27 22:32:46.33]: ▸ Compiling AES256CBC.swift
[...]

PR #6835, opened 2016-11-02
Fix for plist encoding error

edwardlafoy:

Fix for: #6784

It looks like the `config_content` method is expecting to use the `plist` gem, and is actually getting the `CFPropertyList` gem due to a conflicting `Hash#to_plist` monkey patch.

`CFPropertyList` is used by `xcodeproj`, which is a Fastlane dependency.

### DIFF
--- a/gym/lib/gym/generators/package_command_generator_xcode7.rb
+++ b/gym/lib/gym/generators/package_command_generator_xcode7.rb
@@ -174,7 +174,7 @@ module Gym
 
         to_plist(hash)
       end
-      
+
       # Avoids a Hash#to_plist conflict between CFPropertyList and plist gems
       def to_plist(hash)
         Plist::Emit.dump(hash, true)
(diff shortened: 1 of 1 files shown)


gym/lib/gym/generators/package_command_generator_xcode7.rb
```diff
@@ -168,11 +168,15 @@ def config_content
...
-          UI.command_output(hash.to_plist)
+          UI.command_output(to_plist(hash))
           UI.command_output("-----------------------------------------")
         end
 
-        hash.to_plist
+        to_plist(hash)
+      end
+
+      def to_plist(hash)
```
KrauseFx: This looks great. Could you add a comment here on why we have this as a separate method and using the more complex way? This way we'll keep using this like it in the future 👍

KrauseFx (dismissed):

KrauseFx:
Hey @edwardlafoy, thanks for taking the time to dive into this issue and finding the source of it. This looks great and it seems to work perfectly. Can you just address this one comment before it's ready to be merged.

Thanks 🚀

edwardlafoy:
@KrauseFx Great, comment added!

milch (approved):
👍

milch:
Brilliant change, thank you!

KrauseFx:
🚀

PR #6847, opened 2016-11-03
Add `pod update` functionality to cocoapods before `pod install`

ghost:

- [X] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.

I ran the recommended tests.  My apologies for not creating an issue.

All of my mobile projects require running `pod update` before `pod install` to pull in various assets.  Currently, I handle that task before initiating fastlane.  I have also used the fastlane shell commands to drive the pod update function.  I felt it would be best to integrate it into the cocoapods action incase others have the same needs as me.


fastlane/lib/fastlane/actions/cocoapods.rb
```diff
@@ -15,6 +15,7 @@ def self.run(params)
         end
 
         cmd << ['bundle exec'] if File.exist?('Gemfile') && params[:use_bundle_exec]
+        cmd << ['pod update &&'] if params[:use_pod_update]
```
milch: When both `use_pod_update` and `use_bundle_exec` are set, this will cause only pod update to be run using bundle exec, because the whole command will be:

```
bundle exec pod update && pod install
```

fastlane/lib/fastlane/actions/cocoapods.rb
```diff
@@ -68,6 +69,11 @@ def self.available_options
                                        description: "Use bundle exec when there is a Gemfile presented",
                                        is_string: false,
                                        default_value: true),
+          FastlaneCore::ConfigItem.new(key: :use_pod_update,
+                                       env_name: "FL_COCOAPODS_USE_POD_UPDATE",
+                                       description: "Use 'pod update' before 'pod install'",
```
milch: I'm not sure if this option is such a good idea in general, as just always running `pod update` could cause your build to break (e.g. because of API changes in your dependencies). 

I'm not opposed to adding this for power users who know what they are doing, but at the very least the description here needs to somehow reflect that this could be dangerous and is not recommended for Testflight/AppStore or enterprise builds.

milch (changes requested):

fastlane/lib/fastlane/actions/cocoapods.rb
KrauseFx: I agree with @milch on this, it seems risky and not something people should do.

PetrosIchor:
Interesting.  I agree with the bundle exec problem, but not sure i understand the 'risks'.  From my perspective, this is professional tooling.  It should be expected that users are educated in the software they are using.

I left the option as a toggle so if an application doesn't need it, the user doesn't have to turn it on.   

Milch.  I work in the enterprise space with multiple mobile product lines.   The largest platform  component has a pod update requirement.  We want and expect to have decencies updated everyday.  As we try to move closer to a countinous process, store deployments become more frequent (almost daily). 

Thanks for the discussion though!  I'll think of a better way to approach it as well as language to the user.

milch:
Thanks for taking the time to do this right! 

You're right, in a professional environment it's to be expected that people know what they are doing. However, many beginners and amateur programmers are also using `fastlane`, and we have to be inclusive with them too. We also need to be mindful of introducing changes that might result in increased support-load in the future, which is why I'd rather err on the side of caution here. 

I think if we want to merge this, then these things need to be done:

- Make sure both `pod update` and `pod install` are run with `bundle exec` if that option is set
- Change the description from `Use 'pod update' before 'pod install'` to `Use 'pod update' before 'pod install'. This is an option for advanced users. Use with caution, as this might break your build if your dependencies' APIs change!`
- I can see the value in this for non-external CocoaPods, because that's what you have control over. I think it would be valuable to have a list of pods to update, e.g. if I have `ReactiveSwift` in my `Podfile` in addition to my own modules, but don't want to update `ReactiveSwift` because of potentially breaking API changes. One way to implement this could be changing the option from a bool-flag to a list, with some syntactic sugar for updating all pods (e.g. passing `[:update_all_pods]` as a param)

beloso:
Pardon my intrusion.

But if you don't want to update a given pod you should lock its version in your podfile. That's the correct way I think.

I do concur that `pod update` is not something that should be done by default in a CI setting. Unless you have internal (private) pods that you need to be sure you have up to date.

ohwutup:
Hey everyone! I am going to close this PR for now as there hasn't been any activity on it in over a month. If you all would like to continue discussion and/or make the changes mentioned above that would be great, and we can reopen this or a new PR to move forward. Thanks!! 🐙

PR #6849, opened 2016-11-03
[fastlane] Add error message on `fastlane init` when fastlane directory or file already exists

KrauseFx:


fastlane/lib/fastlane/fastlane_folder.rb
```diff
@@ -28,6 +28,8 @@ def self.setup?
 
     def self.create_folder!(path = nil)
       path = File.join(path || '.', FOLDER_NAME)
+      return if File.directory?(path) # directory is already there
```
milch: This check is probably useless

milch (approved):
🐑

PR #6858, opened 2016-11-04
Implement `fastlane_require` to automatically install gems into bundle

KrauseFx:
screenshot 2016-11-03 18 45 20

### DIFF
--- a/fastlane/lib/fastlane/actions/cocoapods.rb
+++ b/fastlane/lib/fastlane/actions/cocoapods.rb
@@ -14,7 +14,9 @@ module Fastlane
           cmd << ["cd '#{podfile_folder}' &&"]
         end
 
-        cmd << ['bundle exec'] if File.exist?('Gemfile') && params[:use_bundle_exec]
+        if File.exist?('Gemfile') && params[:use_bundle_exec] && !Helper.contained_fastlane?
+          cmd << ['bundle exec']
+        end
         cmd << ['pod install']
 
         cmd << '--no-clean' unless params[:clean]
--- a/fastlane/lib/fastlane/actions/cocoapods.rb
+++ b/fastlane/lib/fastlane/actions/cocoapods.rb
@@ -14,7 +14,7 @@ module Fastlane
           cmd << ["cd '#{podfile_folder}' &&"]
         end
 
-        cmd << ['bundle exec'] if shell_out_should_use_bundle_exec?
+        cmd << ['bundle exec'] if params[:use_bundle_exec] && shell_out_should_use_bundle_exec?
         cmd << ['pod install']
 
         cmd << '--no-clean' unless params[:clean]
--- a/fastlane/lib/fastlane/actions/cocoapods.rb
+++ b/fastlane/lib/fastlane/actions/cocoapods.rb
@@ -14,9 +14,7 @@ module Fastlane
           cmd << ["cd '#{podfile_folder}' &&"]
         end
 
-        if File.exist?('Gemfile') && params[:use_bundle_exec] && !Helper.contained_fastlane?
-          cmd << ['bundle exec']
-        end
+        cmd << ['bundle exec'] if shell_out_should_use_bundle_exec?
         cmd << ['pod install']
 
         cmd << '--no-clean' unless params[:clean]
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -188,7 +188,7 @@ module Fastlane
     end
 
     def fastlane_require(gem_name)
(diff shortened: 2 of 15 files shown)


fastlane/lib/fastlane/fast_file.rb
```diff
@@ -24,6 +24,13 @@ def initialize(path = nil)
                 'you should turn off smart quotes in your editor of choice.'
       end
 
+      content.scan(/.*require (.*)/).each do |current|
+        current = current.first
+        next if current.include?(".") # these are local gems
+        next if current.include?("/")
```
KrauseFx: This could be improved

ohwutup: What is this looking for exactly?

fastlane/lib/fastlane/fast_file.rb
```diff
@@ -24,6 +24,13 @@ def initialize(path = nil)
                 'you should turn off smart quotes in your editor of choice.'
       end
 
+      content.scan(/.*require (.*)/).each do |current|
```
ohwutup: Are we looking for `.*require` because we want to check both `fastlane_require` and regular `require`?

ohayon (approved):
I think this looks really awesome! 💯

hjanuschka:
@KrauseFx will this work  on Fastfiles imported with `import_from_git` too? (would be awesome!)

fastlane/lib/fastlane/fast_file.rb
```diff
@@ -24,6 +24,14 @@ def initialize(path = nil)
                 'you should turn off smart quotes in your editor of choice.'
       end
 
+      content.scan(/(.*require) (.*)/).each do |current|
+        require_command = current.first
+        gem_name = current.last
+        next if require_command.include?("fastlane_require")
```
milch: Use a regex that only matches require and not `fastlane_require`, like `^\s*(require) (.*)`

fastlane/lib/fastlane/fast_file.rb
```diff
@@ -179,6 +187,10 @@ def desc_collection
       @desc_collection ||= []
     end
 
+    def fastlane_require(gem_name)
+      FastlaneRequire.install_gem(gem_name: gem_name, require_gem: true)
```
milch: `install_gem` should probably be called `install_gem_if_needed`

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -39,6 +39,12 @@ def self.bundler?
       return false
     end
 
+    # Do we run from a bundled fastlane, which contains Ruby and OpenSSL?
+    # Usually this means the fastlane directory is ~/.fastlane/bin/
+    def self.contained_fastlane?
+      ENV["FASTLANE_IS_BUNDLED"].to_s.length > 0
```
milch: Maybe do an explicit check for `== "true"` here or something, or things will break if someone sets `ENV["FASTLANE_IS_BUNDLED"]="false"` in the future

milch (changes requested):

fastlane_core/lib/fastlane_core/helper.rb
milch: Note: I've left this comment to link to in the other PR, this should be removed if we split this into 2 PR's

KrauseFx:
@milch Thanks for the feedback, I updated the PR 👍

fastlane/lib/fastlane/actions/cocoapods.rb
```diff
@@ -14,7 +14,7 @@ def self.run(params)
           cmd << ["cd '#{podfile_folder}' &&"]
         end
 
-        cmd << ['bundle exec'] if File.exist?('Gemfile') && params[:use_bundle_exec]
+        cmd << ['bundle exec'] if params[:use_bundle_exec] && shell_out_should_use_bundle_exec?
```
milch: Still not 100% satisfied with this. I'd go for 

```ruby
def self.shell_bundle_exec_prefix
    shell_out_should_use_bundle_exec ? "bundle exec" : ""
end
```

and then do 

```ruby
cmd << shell_bundle_exec_prefix if params[:use_bundle_exec]
```

everywhere (or something).

milch: Although I'm willing to conceded that that would be the cherry  on top, more or less, so if the `contained_fastlane?` thing is fixed I'd be okay to merge.

KrauseFx: Yeah, let's try to get the first version working and refactor this later on.

milch: Makes sense

milch (approved):

PR #6637, opened 2016-10-20
[scan] Handle single scheme Xcode projects

0xced:
- [x] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)
- [x] #6555 is merged (and this PR is rebased)
- [x] A new version of `fastlane_core` is released

Single scheme projects are explained here: http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/

With a single scheme project, it is possible that both `Scan.project.ios?` and `Scan.project.tvos?` return false even though iOS and tvOS are supported platforms for the project. So instead of detecting simulators if the project supposedly support iOS or tvOS, simulators are detected when the --device or --devices option is passed to scan.

PR #6864, opened 2016-11-04
[scan] Fix crash when running scan on macOS projects

0xced:

---
scan would crash if the --device option is not specified. This regression was introduced in #6637, sorry. It would also crash for single scheme projects when no --device option is specified.


asfalcone (approved):
👍

PR #6870, opened 2016-11-04
add OS helper

hjanuschka:
adds more OS detections to Helper class

required for: https://github.com/fastlane/fastlane/pull/6871

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -48,16 +48,13 @@ module FastlaneCore
       return false
     end
 
-    def self.unix?
-      !windows?
-    end
-
     def self.windows?
+      # taken from: http://stackoverflow.com/a/171011/1945875
       (/cygwin|mswin|mingw|bccwin|wince|emx/ =~ RUBY_PLATFORM) != nil
     end
 
     def self.linux?
-      unix? and !mac?
+      (/linux/ =~ RUBY_PLATFORM) != nil
     end
 
     # Is the currently running computer a Mac?
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -48,6 +48,18 @@ module FastlaneCore
       return false
     end
 
+    def self.unix?
+      !windows?
+    end
+
+    def self.windows?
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -48,6 +48,18 @@ def self.ci?
       return false
     end
 
+    def self.unix?
```
milch: Not sure what this is useful for in this day and age, I think we should remove it.

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -48,6 +48,18 @@ def self.ci?
...
+      !windows?
+    end
+
+    def self.windows?
+      (/cygwin|mswin|mingw|bccwin|wince|emx/ =~ RUBY_PLATFORM) != nil
+    end
+
+    def self.linux?
+      unix? and !mac?
```
milch: Let's not define linux by it being a unix while not being a mac, I'm sure some people would get a heart attack reading this 😅 

I just checked, `RUBY_PLATFORM` is `<arch>-linux` on linux. Maybe a check like that would be more appropriate.

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -48,6 +48,18 @@ def self.ci?
       return false
     end
 
+    def self.unix?
+      !windows?
+    end
+
+    def self.windows?
+      (/cygwin|mswin|mingw|bccwin|wince|emx/ =~ RUBY_PLATFORM) != nil
```
milch: I see that this code is taken from an answer in http://stackoverflow.com/questions/170956/how-can-i-find-which-operating-system-my-ruby-program-is-running-on ... Please add a comment stating where it was taken from.

milch (changes requested):

fastlane_core/lib/fastlane_core/helper.rb
hjanuschka: absolutly right about that.

hjanuschka:
@milch feedback addressed.

milch (approved):

milch:
Thanks for your contribution! 🚀

PR #6871, opened 2016-11-04
Fix 'fastlane env' crash on non-macOS OS

hjanuschka:
second part of: https://github.com/fastlane/fastlane/pull/6870
tests will fail until the first part is merged.

### DIFF
--- a/fastlane/lib/fastlane/environment_printer.rb
+++ b/fastlane/lib/fastlane/environment_printer.rb
@@ -173,20 +173,51 @@ module Fastlane
       require "openssl"
 
       env_output = "### Stack\n\n"
-      product, version, build = `sw_vers`.strip.split("\n").map { |line| line.split(':').last.strip }
+      product, version, build = "Unknown"
+      os_version = "UNKNOWN"
+
+      if Helper.mac?
+        product, version, build = `sw_vers`.strip.split("\n").map { |line| line.split(':').last.strip }
+        os_version = version
+      end
+
+      if Helper.linux?
+        # this should work on pretty much all linux distros
+        os_version = `uname -a`.strip
+        version = ""
+        build = `uname -r`.strip
+        product = `cat /etc/issue.net`.strip
+
+        distro_guesser = {
+          fedora: "/etc/fedora-release",
+          debian_based: "/etc/debian_version",
+          suse: "/etc/SUSE-release",
+          mandrake: "/etc/mandrake-release"
+        }
+
+        distro_guesser.each do |dist, vers|
+          os_version = "#{dist} " + File.read(vers).strip if File.exist?(vers)
+          version = os_version
+        end
+      end
+
       table_content = {
         "fastlane" => Fastlane::VERSION,
-        "OS" => `sw_vers -productVersion`.strip,
+        "OS" => os_version,
         "Ruby" => RUBY_VERSION,
         "Bundler?" => Helper.bundler?,
-        "Xcode Path" => Helper.xcode_path,
-        "Xcode Version" => Helper.xcode_version,
         "Git" => `git --version`.strip.split("\n").first,
         "Installation Source" => $PROGRAM_NAME,
         "Host" => "#{product} #{version} (#{build})",
         "Ruby Lib Dir" => RbConfig::CONFIG['libdir'],
         "OpenSSL Version" => OpenSSL::OPENSSL_VERSION
       }
+
+      if Helper.mac?
+        table_content["Xcode Path"] = Helper.xcode_path
+        table_content["Xcode Version"] = Helper.xcode_version
+      end
+
       table = ["| Key | Value |"]
       table += table_content.collect { |k, v| "| #{k} | #{v} |" }
 


fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,10 +132,24 @@ def self.print_system_environment
       require "openssl"
 
       env_output = "### Stack\n\n"
-      product, version, build = `sw_vers`.strip.split("\n").map { |line| line.split(':').last.strip }
+      product, version, build = "Unkown"
```
milch: `Unkown` -> `Unknown`

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,10 +132,24 @@ def self.print_system_environment
...
+      os_version = "UNKOWN"
```
milch: `UNKOWN` -> `UNKNOWN`

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,10 +132,24 @@ def self.print_system_environment
...
+
+      if FastlaneCore::Helper.mac?
+        product, version, build = `sw_vers`.strip.split("\n").map { |line| line.split(':').last.strip }
+        os_version = `sw_vers -productVersion`.strip
```
milch: Why not use the version from the line above? Those strings should be the same, I think.

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,10 +132,24 @@ def self.print_system_environment
...
+      end
+
+      if FastlaneCore::Helper.linux?
```
milch: Linux doesn't have a single way to check for version information unfortunately, but some are described here:  http://unix.stackexchange.com/questions/88644/how-to-check-os-and-version-using-a-linux-command

It would be nice if we could fall back on multiple types of ways of checking the version information on Linux.

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,10 +132,24 @@ def self.print_system_environment
...
+        os_version = `uname -a`.strip
+        version = ""
+        build = ""
+        product = `cat /etc/issue`.strip
```
milch: Use `/etc/issue.net` instead to receive nicer output.

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,10 +132,24 @@ def self.print_system_environment
...
+
+      if FastlaneCore::Helper.mac?
+        product, version, build = `sw_vers`.strip.split("\n").map { |line| line.split(':').last.strip }
+        os_version = `sw_vers -productVersion`.strip
+      end
+
+      if FastlaneCore::Helper.linux?
+        os_version = `uname -a`.strip
+        version = ""
+        build = ""
```
milch: We could show the Kernel version here using `uname -r`

milch (changes requested):

hjanuschka:
@milch  - feedback addressed - also fixed xcode-select problem on linux 😭 


linux output:

| Key                 | Value                            |
| ------------------- | -------------------------------- |
| OS                  | 8.6                              |
| Ruby                | 2.1.5                            |
| Bundler?            | true                             |
| Git                 | git version 2.1.4                |
| Installation Source | /var/lib/gems/2.1.0/bin/fastlane |
| Host                | Debian 8.6 (3.16.0-4-amd64)      |
| Ruby Lib Dir        | /usr/lib                         |
| OpenSSL Version     | OpenSSL 1.0.1t  3 May 2016       |



mac output:

| Key                 | Value                                               |
| ------------------- | --------------------------------------------------- |
| OS                  | 10.12.1                                             |
| Ruby                | 2.2.1                                               |
| Bundler?            | false                                               |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/         |
| Xcode Version       | 8.1                                                 |
| Git                 | git version 2.10.0                                  |
| Installation Source | /Users/hjanuschka/.rvm/gems/ruby-2.2.1/bin/fastlane |
| Host                | Mac OS X 10.12.1 (16B2555)                          |
| Ruby Lib Dir        | /Users/hjanuschka/.rvm/rubies/ruby-2.2.1/lib        |
| OpenSSL Version     | OpenSSL 1.0.2h  3 May 2016                          |

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -132,20 +132,47 @@ def self.print_system_environment
...
+      if FastlaneCore::Helper.linux?
+        # this should work on pretty much all linux distros
+        os_version = `uname -a`.strip
+        version = ""
+        build = `uname -r`.strip
+        product = `cat /etc/issue.net`.strip
+
+        # see if we can get a bit more details using 'lsb_release'
+        # check if it is availaible
+        lsb_release = `which lsb_release`.strip
```
milch: This is good, but doesn't work well on Ubuntu. Something like this gives you more information from `/etc/lsb-release` (which is available on Ubuntu) that we should use: 

```ruby
Hash[File.read("/etc/lsb-release").lines.map { |t| t.strip.split("=") }]
```

hjanuschka: in fact /etc/lsb-release is not existing by default on debian only with some weird `edu-package` - and contains:

```
DISTRIB_DESCRIPTION="DebianEdu/Skolelinux"
```

see my latest commit - with a different try.

hjanuschka:
latest commit changes to:


| Key                 | Value                                               |
| ------------------- | --------------------------------------------------- |
| OS                  | debian_based8.6                                     |
| Ruby                | 2.1.5                                               |
| Bundler?            | true                                                |
| Git                 | git version 2.1.4                                   |
| Installation Source | /var/lib/gems/2.1.0/bin/fastlane                    |
| Host                | Debian GNU/Linux 8 debian_based 8.6 (3.16.0-4-amd64) |
| Ruby Lib Dir        | /usr/lib                                            |
| OpenSSL Version     | OpenSSL 1.0.1t  3 May 2016                          |

milch (approved):

milch:
Just tested it on Ubuntu, much better output. 

Can you rebase master now please so the tests pass, then we can merge!

hjanuschka:
rebased - tests still fail because of - https://github.com/fastlane/fastlane/pull/6870 (merged) but missing `fastlane_core` release - because test fail out of `linux?`  undefined....

@milch can you do the fastlane_core release

hjanuschka:
ok tests should pass now.

hjanuschka:
i don't get it - rebased and pushed, and fastlane_core seems to be released - but still no `linux?`

milch:
Hmm, curious. I have no idea, I just tried it on my machine and it seems to work. Try removing the `FastlaneCore::` prefix from the Helper class on lines 143 and 138, it should work without (see line 175).

hjanuschka:
@milch god damn it - all tests pass but circle still read?

hjanuschka:
@milch  i got it green!!! with the last rebase.

milch:
Awesome, thanks for your patience! 🚀

Issue #6880, opened 2016-11-06
cert supports path to custom keychain, but create_keychain does not


GoodMirek:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
cert supports path to custom keychain, but create_keychain does not.
Use case is to create a temporary keychain in working directory by CI process. The working directory is always cleaned up, thus assuring maximum reproducibility and no dependency/interference with files in ~/Library/Keychains

Is it desirable to add the support for custom keychain path into create_keychain?

hjanuschka:
@GoodMirek - i love the "maximum reproducibility" - see this PR: https://github.com/fastlane/fastlane/pull/6887 - maybe you +1 it for me.

GoodMirek:
Solved via #6887, thanks!

PR #6887, opened 2016-11-07
[fastlane] support keychains in custom paths and fix non-existing default keychain crash🔐

hjanuschka:
as discussed here: https://github.com/fastlane/fastlane/issues/6880

adds parameter `path` - so `create_keychain` either accepts `path` (full path to keychain)  - or `name` (name inside `~/Libarary/Keychains/`)

  * also fixes issue with `set-default` - when there is no default keychain

sample:
```
lane :demo do
        create_keychain(path: "/tmp/11131.keychain")
end
```

### DIFF
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -14,7 +14,7 @@ module Fastlane
           escaped_name = params[:name].shellescape
           keychain_path = "~/Library/Keychains/#{escaped_name}"
         else
-          keychain_path = params[:path]
+          keychain_path = params[:path].shellescape
         end
 
         if keychain_path.nil?
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -8,30 +8,44 @@ module Fastlane
 
     class CreateKeychainAction < Action
       def self.run(params)
-        escaped_name = params[:name].shellescape
         escaped_password = params[:password].shellescape
 
-        commands = []
-        commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{escaped_name}", log: false)
-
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -48,6 +57,11 @@ def self.available_options
                                        env_name: "KEYCHAIN_NAME",
                                        description: "Keychain name",
                                        optional: false),
+          FastlaneCore::ConfigItem.new(key: :is_path,
```
mpirri: Rather than setting -is_path and passing in a keychain via -name, why not create a -path option and pass the keychain path in that way? Seems simpler than having to set a boolean value on the command and then use a inaccurately name parameter ("name"). Thoughts?

hjanuschka:
@mpirri you are right - i did the changes.

  * rebased to one commit
  * refactored to support either :path or :name

hjanuschka:
fixed tests + added some for :path param

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -8,30 +8,44 @@ module SharedValues
...
     class CreateKeychainAction < Action
       def self.run(params)
-        escaped_name = params[:name].shellescape
         escaped_password = params[:password].shellescape
 
+        if params[:name]
+          escaped_name = params[:name].shellescape
+          keychain_path = "~/Library/Keychains/#{escaped_name}"
+        else
+          keychain_path = params[:path]
```
mpirri: Is this missing a `.shellescape' ?

mpirri (changes requested):
This is looking really good - thanks for adding those tests. Just let me know what you think about that `.shellescape`.

hjanuschka:
@mpirri  added shellescape! thx

mpirri (approved):

Issue #6878, opened 2016-11-06
Set 'log: true' while calling 'Fastlane::Actions.sh' if CLI option --verbose


GoodMirek:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I found it difficult to troubleshoot errors like:
<br>`ERROR [2016-11-06 03:34:20.16]: Shell command exited with exit status 48 instead of 0.`<br>
which are returned while running action e.g. `create_keychain`
Currently, shell command is called like this:
<br>`Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{escaped_name}", log: false)`<br>

Would you agree to set `log: true` if command line option `--verbose` is specified?
Or do you prefer another way for setting the `log: true`?
Please may you let me know so I can submit a pull request?

KrauseFx:
Yeah, I like the idea of having log being true if not specified otherwise 👍 
Do you think there will be any side-effects?

KrauseFx:
For actions like the keychain, we can't just print the output, as it also contains the password. This shouldn't show up in build logs, even for `verbose` mode

hjanuschka:
how about adding `--troubleshoot` - wich enforces `true` on the `sh` action and also sets `verbose` - so you can manually choose it.

GoodMirek:
I like the `--troubleshoot` option.

If even more safety was desired, we could impose interactive user confirmation asking the user whether they really desire to display sensitive info, like the confirmation required in case of `match nuke`.
Also, the `--troubleshoot` option might be ignored if is_ci is true.

milch:
> If even more safety was desired, we could impose interactive user confirmation asking the user whether they really desire to display sensitive info, like the confirmation required in case of match nuke.

Yes, I very much agree. Some users might not realize, enable it on CI (or wherever) and leak information. There's also the issue of people wanting to be helpful and copying the output with `--troubleshoot` to their issue/PR. 

- When running on CI, I wouldn't ignore the option, but fail and let the user know that this option needs to be disabled on CI
- Otherwise there should be an interactive confirmation telling the user not to copy/paste the output anywhere w/o double checking and removing all passwords from the text and wether they are sure they really want to run with `--troubleshoot`

GoodMirek:
Solved via #6889.

PR #6889, opened 2016-11-07
[WIP] add troubleshoot option 🔫

hjanuschka:
as proposed here - to help debugging/troubleshooting.
https://github.com/fastlane/fastlane/issues/6878

adds `--troubleshoot`  enables `log:true`  on all `.sh` calls.
this should not be run on CI, existing setups using `--verbose` on CI - should not be affected, therefore not passwords in ci log

### DIFF
--- a/fastlane/lib/fastlane/commands_generator.rb
+++ b/fastlane/lib/fastlane/commands_generator.rb
@@ -48,6 +48,13 @@ module Fastlane
     end
 
     def self.confirm_troubleshoot
+      if Helper.is_ci?
+        UI.error "---"
+        UI.error "You are trying to use '--troubleshoot' on CI"
+        UI.error "this option is not usable in CI, as it is insecure"
+        UI.error "---"
+        UI.user_error!("Do not use --troubleshoot in CI")
+      end
       # maybe already set by 'start'
       return if $troubleshoot
       UI.error "---"
--- a/fastlane/lib/fastlane/commands_generator.rb
+++ b/fastlane/lib/fastlane/commands_generator.rb
@@ -12,6 +12,7 @@ module Fastlane
       # however we do want to log verbose information in the PluginManager
       $verbose = true if ARGV.include?("--verbose")
       $capture_output = true if ARGV.include?("--capture_output")
+      $troubleshoot = true if ARGV.include?("--troubleshoot")
 
       if $capture_output
         # Trace mode is enabled
@@ -56,6 +57,10 @@ module Fastlane
 
       global_option('--verbose') { $verbose = true }
       global_option('--capture_output', 'Captures the output of the current run, and generates a markdown issue template') { $capture_output = true }
+      global_option('--troubleshoot', 'Enables extended verbose mode, beware this eventually includes passwords, should not be run on CI') do
+        $verbose = true
+        $troubleshoot = true
+      end
 
       always_trace!
 
--- a/fastlane/lib/fastlane/commands_generator.rb
+++ b/fastlane/lib/fastlane/commands_generator.rb
@@ -58,9 +58,9 @@ module Fastlane
       # maybe already set by 'start'
       return if $troubleshoot
       UI.error "---"
-      UI.error "Are you sure to enable '--troubleshoot'."
+      UI.error "Are you sure you want to enable '--troubleshoot'?"
       UI.error "All commmands will run in full unfiltered output mode."
-      UI.error "Beware that there maybe passwords printed to the console/log."
+      UI.error "Sensitive data, like passwords, could be printed to the log."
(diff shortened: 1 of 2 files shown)


milch:
I think this needs some more discussion, see my comments in #6878. I'll add [WIP] to the title for now to reflect this.

hjanuschka:
@milch i did a confirm arround the option.

`fastlane demo --troubleshoot`

bildschirmfoto 2016-11-08 um 07 57 18

fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -42,6 +47,26 @@ def self.start
       end
     end
 
+    def self.confirm_troubleshoot
+      if Helper.is_ci?
+        UI.error "---"
+        UI.error "You are trying to use '--troubleshoot' on CI"
+        UI.error "this option is not usable in CI, as it is insecure"
+        UI.error "---"
+        UI.user_error!("Do not use --troubleshoot in CI")
```
KrauseFx: Do we really want to force people not to do that? Showing a warning is fine, but I think we should still let them do their thing.

fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -42,6 +47,26 @@ def self.start
...
+      end
+      # maybe already set by 'start'
+      return if $troubleshoot
+      UI.error "---"
+      UI.error "Are you sure to enable '--troubleshoot'."
+      UI.error "All commmands will run in full unfiltered output mode."
+      UI.error "Beware that there maybe passwords printed to the console/log."
+      UI.error "---"
+      if agree("(y/n)", true)
```
KrauseFx: Please use `UI.confirm` instead, which should add the `y/n` for you 👍

fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -56,6 +81,7 @@ def run
 
       global_option('--verbose') { $verbose = true }
       global_option('--capture_output', 'Captures the output of the current run, and generates a markdown issue template') { $capture_output = true }
+      global_option('--troubleshoot', 'Enables extended verbose mode, beware this eventually includes passwords, should not be run on CI')
```
KrauseFx: Besides here, do we want to document this anywhere?

fastlane/lib/fastlane/commands_generator.rb
milch: See the discussion here: #6878

hjanuschka: both are right  - please just let me know if we should keep or remove it.

fastlane/lib/fastlane/commands_generator.rb
hjanuschka: where would be a good place?

hjanuschka:
@KrauseFx  feedback adressed

fastlane/lib/fastlane/commands_generator.rb
KrauseFx: I wasn't aware of the previous discussion, let's keep it then

fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -42,6 +47,26 @@ def self.start
...
+        UI.error "---"
+        UI.error "You are trying to use '--troubleshoot' on CI"
+        UI.error "this option is not usable in CI, as it is insecure"
+        UI.error "---"
+        UI.user_error!("Do not use --troubleshoot in CI")
+      end
+      # maybe already set by 'start'
+      return if $troubleshoot
+      UI.error "---"
+      UI.error "Are you sure to enable '--troubleshoot'."
```
milch: `Are you sure you want to enable '--troubleshoot'?`

fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -42,6 +47,26 @@ def self.start
...
+      UI.error "All commmands will run in full unfiltered output mode."
+      UI.error "Beware that there maybe passwords printed to the console/log."
```
milch: `Sensitive data, like passwords, could be printed to the log.`

fastlane/lib/fastlane/commands_generator.rb
milch: For the description: `Enables extended verbose mode. Use with caution, as this even includes ALL sensitive data. Cannot be used on CI.`

milch (changes requested):

hjanuschka:
@milch  feedback adressed

milch (approved):

milch:
Looks good, thanks!

PR #6886, opened 2016-11-07
[match] support force on nuke 💣

hjanuschka:
support `--force` on `match nuke` - to support non interactive shells.

PR #6894, opened 2016-11-07
[match] fix/rework for #6886

hjanuschka:
this is a rework of #6886.
reverts the use of `:force` to the state before.

adds the option `--unattended`. so the combination of `match --force --nuke`  - ignites the nuke - without human interaction.


match/lib/match/nuke.rb
```diff
@@ -27,15 +27,15 @@ def run(params, type: nil)
       end
 
       if (self.certs + self.profiles + self.files).count > 0
-        unless params[:force]
+        unless params[:unattended]
```
mpirri: Documentation seems to indicate that both `-unattended` and `-force` are required, but it looks like only checking `params[:unattended]` here.

match/lib/match/options.rb
```diff
@@ -80,6 +80,11 @@ def self.available_options
                                      description: "Renew the provisioning profiles every time you run match",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :unattended,
```
mpirri: I'm trying to think of other attribute names we can use here (`force` seems like the best choice, but it isn't worth the argument confusion). How about something like `-skip_confirmation`?  Or we could do `-force_nuke` to allow us to stick the the better `force` naming, but not overloading the actual `nuke` argument. What do you think?

mpirri (changes requested):

hjanuschka:
@mpirri  done, i like `skip_confirmation`

match/lib/match/options.rb
```diff
@@ -80,6 +80,11 @@ def self.available_options
                                      description: "Renew the provisioning profiles every time you run match",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :skip_confirmation,
+                                     env_name: "MATCH_SKIP_CONFIRMATION",
+                                     description: "Disables confirmation dialogs, answering them with yes",
```
KrauseFx: Which currently is only used in `nuke`, we should mention that

hjanuschka:
@KrauseFx feedback addressed - wording is not my strength - not even in my motherlanguage

match/lib/match/options.rb
```diff
@@ -80,6 +80,11 @@ def self.available_options
                                      description: "Renew the provisioning profiles every time you run match",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :skip_confirmation,
+                                     env_name: "MATCH_SKIP_CONFIRMATION",
+                                     description: "Disables confirmation dialogs during nuke, answering them with yes",
```
mpirri: Sorry to keep bothering you with wording changes, but lets use "prompts" rather than "dialogs" here. So:
```Disables confirmation prompts during nuke, answering them with yes```

hjanuschka:
@mpirri  wording done.

mpirri (approved):

PR #6902, opened 2016-11-07
Add contained_fastlane? to Helper.rb

KrauseFx:
This is for https://github.com/fastlane/fastlane/pull/6858

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -43,7 +43,7 @@ module FastlaneCore
     # Usually this means the fastlane directory is ~/.fastlane/bin/
     # We set this value via the environment variable `SELF_CONTAINED`
     def self.contained_fastlane?
-      ENV.key?("SELF_CONTAINED")
+      ENV["SELF_CONTAINED"].to_s == "true"
     end
 
     # @return [boolean] true if building in a known CI environment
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -39,6 +39,12 @@ def self.bundler?
       return false
     end
 
+    # Do we run from a bundled fastlane, which contains Ruby and OpenSSL?
+    # Usually this means the fastlane directory is ~/.fastlane/bin/
+    def self.contained_fastlane?
+      ENV["FASTLANE_IS_BUNDLED"].to_s.length > 0
```
milch: See [this comment](https://github.com/fastlane/fastlane/pull/6858/files#r86898616)

milch (changes requested):

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -39,6 +39,13 @@ def self.bundler?
       return false
     end
 
+    # Do we run from a bundled fastlane, which contains Ruby and OpenSSL?
+    # Usually this means the fastlane directory is ~/.fastlane/bin/
+    # We set this value via the environment variable `SELF_CONTAINED`
+    def self.contained_fastlane?
+      ENV.key?("SELF_CONTAINED")
```
milch: I'd rather we explicitly check `ENV["SELF_CONTAINED"].to_s == "true"` here

milch (changes requested):

fastlane_core/lib/fastlane_core/helper.rb
KrauseFx: In general
```
ENV["SELF_CONTAINED"].to_s == "true
```
doesn't work when the user uses
```
export SELF_CONTAINED="1"
```
however for this case it's good enough

milch (approved):

milch:
🔥

PR #6903, opened 2016-11-08
[fastlane_core] Version bump

KrauseFx:
* Add verbose output if environment variables are used (#6895)
* Enhance env printing + add  `--capture_output`  option
* Add new Helper methods


milch (dismissed):
🐐

PR #5061, opened 2016-06-09
[Match] added installed certificate info logging

punksta:
Added feature https://github.com/fastlane/fastlane/issues/5014

```
[13:21:49]: Installing certificate...
+-------------------+-----------------------------------------------+
|                    Installed Code Certificate                     |
+-------------------+-----------------------------------------------+
| User ID           | **********                                    |
| Certificate Name  | iPhone Developer: Stas Shakirov (***********) |
| Organisation Unit | **********                                    |
| Organisation Name | ********* *****                               |
| Country           | US                                            |
| Start Datetime    | Jun  3 11:36:38 2016 GMT                      |
| End Datetime      | Jun  3 11:36:38 2017 GMT                      |
+-------------------+-----------------------------------------------+
[13:21:51]: Installing provisioning profile...
```

PR #6907, opened 2016-11-08
[match] print certificate info ℹ️

hjanuschka:
this one redoes https://github.com/fastlane/fastlane/pull/5061 - credit goes to @punksta
i had to rework it - as merging/rebasing was to heavy.

bildschirmfoto 2016-11-08 um 15 10 20

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -85,6 +85,10 @@ module Match
         # Import the private key
         # there seems to be no good way to check if it's already installed - so just install it
         Utils.import(keys.last, params[:keychain_name], password: params[:keychain_password])
+
+        # Get and print info of certificate
+        info = Utils.get_cert_info(cert_path)
+        TablePrinter.print_certificate_info(cert_info: info)
       end
 
       return File.basename(cert_path).gsub(".cer", "") # Certificate ID
--- a/match/lib/match/utils.rb
+++ b/match/lib/match/utils.rb
@@ -45,6 +45,35 @@ module Match
       (base_environment_variable_name(app_identifier: app_identifier, type: type) + ["profile-name"]).join("_")
     end
 
+    def self.get_cert_info(cer_certificate_path)
+      command = "openssl x509 -inform der -in #{cer_certificate_path.shellescape} -subject -dates -noout"
+      command << " &" # start in separate process
+      output = Helper.backticks(command, print: $verbose)
+
+      # openssl output:
+      # subject= /UID={User ID}/CN={Certificate Name}/OU={Certificate User}/O={Organisation}/C={Country}\n
+      # notBefore={Start datetime}\n
+      # notAfter={End datetime}
+      cert_info = output.gsub(/\s*subject=\s*/, "").tr("/", "\n")
+      out_array = cert_info.split("\n")
+      openssl_keys_to_readable_keys = {
+           'UID' => 'User ID',
+           'CN' => 'Common Name',
+           'OU' => 'Organisation Unit',
+           'O' => 'Organisation',
+           'C' => 'Country',
+           'notBefore' => 'Start Datetime',
+           'notAfter' => 'End Datetime'
+       }
+
+      return out_array.map { |x| x.split(/=+/) if x.include? "=" }
+                      .compact
+                      .map { |k, v| [openssl_keys_to_readable_keys.fetch(k, k), v] }
+    rescue => ex
+      UI.error(ex)
+      return {}
+    end
+
     def self.base_environment_variable_name(app_identifier: nil, type: nil)
       ["sigh", app_identifier, type]
     end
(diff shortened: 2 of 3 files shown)


match/lib/match/utils.rb
```diff
@@ -45,6 +45,55 @@ def self.environment_variable_name_profile_name(app_identifier: nil, type: nil)
       (base_environment_variable_name(app_identifier: app_identifier, type: type) + ["profile-name"]).join("_")
     end
 
+    # Fill identity(name) of certificate to env variable
+    def self.fill_environment_cert(params, identity)
```
milch: This seems to be unused, in that case, remove please.

match/lib/match/runner.rb
```diff
@@ -85,6 +85,14 @@ def fetch_certificate(params: nil)
         # Import the private key
         # there seems to be no good way to check if it's already installed - so just install it
         Utils.import(keys.last, params[:keychain_name], password: params[:keychain_password])
+
+        # Get and print info of certificate
+        info = Utils.get_cert_info(cert_path)
+        cert_name = info.select { |key, value| key == 'Common Name' }
+        if cert_name.length > 0
```
milch: Why is this check needed? When would the length be 0?

match/lib/match/utils.rb
```diff
@@ -45,6 +45,55 @@ def self.environment_variable_name_profile_name(app_identifier: nil, type: nil)
...
+      # notAfter={End datetime}
+
+      # removing subject= in first line
+      subject_value = output[0][output[0].index('=') + 2, output[0].length]
+
+      out_array = subject_value.split('/')
+      out_array << output[1]
+      out_array << output[2]
+
+      oppenssl_keys_to_readable_keys = {
```
milch: typo: `oppenssl_keys_to_readable_keys` -> `openssl_keys_to_readable_keys`

match/lib/match/utils.rb
```diff
@@ -45,6 +45,55 @@ def self.environment_variable_name_profile_name(app_identifier: nil, type: nil)
...
+      output = Helper.backticks(command, print: $verbose).split("\n")
+
+      # openssl output:
+
+      # subject= /UID={User ID}/CN={Certificate Name}/OU={Certificate User}/O={Organisation}/C={Country}\n
+      # notBefore={Start datetime}\n
+      # notAfter={End datetime}
+
+      # removing subject= in first line
+      subject_value = output[0][output[0].index('=') + 2, output[0].length]
```
milch: Do `.gsub(/\s*subject=\s*/, "")` and `.gsub("/", "\n")` before running `.split("\n")` instead, that's a bit more clear and doesn't introduce 2 temporaries.

match/lib/match/utils.rb
```diff
@@ -45,6 +45,55 @@ def self.environment_variable_name_profile_name(app_identifier: nil, type: nil)
...
+           'C' => 'Country',
+           'notBefore' => 'Start Datetime',
+           'notAfter' => 'End Datetime'
+       }
+
+      # collect openssl answer to structure [[key1, v1], [key2, v2], ...]
+      key_value_pairs = out_array.map { |x| x.split(/=+/) if x.include? "=" }.compact
+
+      # change keys to readable
+      table_data = key_value_pairs.map { |k, v| [oppenssl_keys_to_readable_keys.fetch(k, k), v] }
```
milch: We don't need to introduce temporaries here, with the changes from above you can do:

```ruby
       output.map { |x| x.split(/=+/) if x.include? "=" }
                  .compact
                  .map { |k, v| [oppenssl_keys_to_readable_keys.fetch(k, k), v] }
```

milch (changes requested):
Just tried it, I really like this change, thanks! 🚀  Needs some simplification, then we can merge.

hjanuschka:
@milch feedback addressed.

about the `cert_info.length > 0` check on `common name` - i took it from the original PR - i think its because of "in X.509 the Common Name is not mandatory." - but i think its ok to skip/remove that check, as in ios world - common name is always set.

  * rebased current master

milch (approved):

milch:
@hjanuschka Alright, this looks good to me then. Thanks for your contribution! 🚀 

PS: In the future, when addressing reviews and/or rebasing on master, please don't `--amend` and force push. It makes it easier to just quickly scan over just the requested changes, instead of having to go through all the code again. We squash anyway before merging 😉

hjanuschka:
@milch - ok thx for the info, will do so in the future - just a bad habbit (from back in the days when it was not possible to squash) to submit only one commit!

KrauseFx:
Is the country always US?

hjanuschka:
@KrauseFx seems to be always US - not sure why - either apple or sigh (in the crt) ?

hjanuschka:
@KrauseFx  infact i think its defaulting to US - because  fastlane/spaceship - does not set a country code.


https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/certificate.rb#L195 the country should be included in the subject of the CSR.


sample on how to set country code:
https://gist.github.com/mitfik/1922961#file-ruby_csr_example-rb-L31

Issue #6901, opened 2016-11-07
Sigh: Add option to update provision profile with a single UDID


craigho:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I'm using fastlane (sigh) to manage a large list of UDIDs. It is easier for my purposes to have sigh add a single UDID to a provision profile, rather than pulling in all available UDIDs.

I am proposing a feature: a flag that works similar to/overrides --force functionality. The profile is renewed, but only a single UDID is added. (Existing UDIDs will remain). This can be accomplished by setting `profiles.devices` to a single device rather than `Spaceship.device.all_for_profile_type(profile.type)`. The device can be found by using `Spaceship.device.find_by_udid`

Please let me know if I should submit a pull request for this functionality. https://github.com/craigho/fastlane/tree/device-list

##### Complete output when running fastlane, including the stack trace and command used
(--add is the newly proposed flag for this functionality)

> sigh -a com.company.app.id -u abc@xyz.com --skip_install --skip_certificate_verification --add 1234567890123456789012345678901234567890 --development
> [14:11:46]: Starting login with user 'abc@xyz.com'
> [14:11:48]: Successfully logged in
> [14:11:48]: Fetching profiles...
> [14:11:49]: Found 1 matching profile(s)
> [14:11:50]: Attempting to add 1234567890123456789012345678901234567890 to the provision profile
> [14:11:57]: Downloading provisioning profile...
> [14:11:57]: Successfully downloaded provisioning profile...

TKBurner:
@craigho This is an interesting idea. I can understand the use case. If you are interested in a submitting a PR we'd work with you.

craigho:
Thanks @TKBurner, I have submitted my PR.

milch:
I'm going to close this, as discussed in https://github.com/fastlane/fastlane/pull/6911#issuecomment-270947910

PR #6911, opened 2016-11-08
Add option to update provision profile with a single UDID (Issue #6901)

craigho:

- [X] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.

---
See: https://github.com/fastlane/fastlane/issues/6901

The purpose of this enhancement is to allow the renewal of an adhoc/development provision profile while adding only a single new UDID, rather than pulling in all available UDIDs. This is useful if you have a single developer.apple.com account but there are multiple projects and users. Some devices will be used to develop one app, while a different set of devices might be used to develop a different app. Adding all devices to all provisioning profiles might allow app leakage across testing teams.


KrauseFx:
I'm not 100% sold on this feature, this seems to be a very specific use-case. Couldn't you solve this by providing the name instead?

craigho:
@krausefx thanks for taking the time to review. 

I apologize, but I don't quite follow your question. I'm not aware of any functionality in sigh that allows you to add a single device to a profile without pulling in all available devices - either by name or UDID. If there is such an option, I feel foolish having overlooked it!

craigho:
@krausefx I'd like to merge this in if possible.... lets continue the discussion on this one.

craigho:
Happy to resubmit if necessary?

milch:
Hi @craigho, sorry for the huge delay with your PR/issue. 

I'm not 100% sold on the idea - is there any reason/benefit of doing this as part of running `sigh`, instead of adding the device before and then running `sigh` with `force:true`? I'd rather not repurpose the tool for something it's not intended for and that's not beneficial to most users.

I think it'd be better if we added a flag for renewing, but keeping the same device list. That way one could add as many devices as one wants, e.g. 

```ruby
lane :add_new_devices do
    register_devices(devices_file: "devices.txt")

    profile = Sigh.Runner.new.fetch_profiles.first
    #... extract new devices from "devices.txt" into device_list
    profile.devices << device_list    
    profile.update!

    sigh(
        keep_device_list: true, 
        force: true,
        # ... 
    ) 
end
```

craigho:
Your suggestion seems workable. I do agree that my solution is fairly specific to my use-case. I guess the flip side of the argument is that if you did want to add only a single device, my workflow is a single stage and easier from a user standpoint.

If we want to go in this direction, I can test out the workflow and submit a new PR. But I won't be able to do so until March.

milch:
> Your suggestion seems workable. I do agree that my solution is fairly specific to my use-case.

Alright, thanks for understanding! We generally only want to support features with fastlane that are useful to a majority of users. I can see the value in keeping the device list of a profile constant, both for small developers and large, so that is what I'd rather support - especially if it also solves this issue. 

>  I guess the flip side of the argument is that if you did want to add only a single device, my workflow is a single stage and easier from a user standpoint.

There is always the option to make a more easy-to-use action out of this, and submit it as a [plugin](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md#plugins)! Plugins are first class citizens in fastlane, the only difference is that they are user maintained. If you decide to create a plugin and it turns out a lot of users are using it, we can then always revisit adding those features back into the main fastlane repo. 

> If we want to go in this direction, I can test out the workflow and submit a new PR. But I won't be able to do so until March.

No worries, no one is in a rush here. I'm going to close this PR for now, let me know if I should reopen it!

Issue #6914, opened 2016-11-08
Add language settings to env output


milch:
We often receive issues that have problems with the language settings, see #6891. Users receive errors like `�[33mencode': "\xE2" on US-ASCII (Encoding::InvalidByteSequenceError)�[0m`, or other, similar ones. To be able to quickly triage/solve these issues, the value of these environment vars needs to be added to the output of `fastlane env`:

- LANG
- LANGUAGE
- LC_ALL

The value of all of these should be `en_US.UTF-8` (or some other type of UTF-8, I guess), and the output should show some sign if they aren't (e.g. ✅  and ⛔️ respectively)

hjanuschka:
ready for review: https://github.com/fastlane/fastlane/pull/6915

PR #6915, opened 2016-11-08
[fastlane] add locale envs

hjanuschka:
as proposed here: https://github.com/fastlane/fastlane/issues/6914

i actually added my own LANG as ✅  because getting ⛔️ just because not an american is too hard for me.

### DIFF
--- a/fastlane/lib/fastlane/environment_printer.rb
+++ b/fastlane/lib/fastlane/environment_printer.rb
@@ -15,6 +15,7 @@ module Fastlane
       require "fastlane/markdown_table_formatter"
       env_output = ""
       env_output << print_system_environment
+      env_output << print_system_locale
       env_output << print_fastlane_files
       env_output << print_loaded_fastlane_gems
       env_output << print_loaded_plugins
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -15,6 +15,7 @@ def self.get
       require "fastlane/markdown_table_formatter"
       env_output = ""
       env_output << print_system_environment
+      env_output << print_system_locale
```
KrauseFx: Isn't that something we could fix, instead of just showing in the output?

hjanuschka:
yeah i am not sure what happens once you enforce en_US.utf8 - on a system wich hasn't installed the locale.

for debugging/troubleshooting, having it in `fastlane env` is nice anyway.
(not 100% sure about the emojis, there are a lot of charsets without problems - that are not US)

milch:
I agree, I'm not sure about all the things that could possibly break if we just set the locale for the user. At least now we know the error message, it's easy to debug, and if we spot that the locale is set incorrectly in the `env` output (after merging this PR) it'll be easy to close issues that have this problem as well.

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -128,6 +129,25 @@ def self.print_loaded_gems
       return env_output
     end
 
+    def self.print_system_locale
+      env_output = "### System Locale\n\n"
+      table = "| Variable | Value |  |\n"
+      table << "|-----|---------|----|\n"
+      ["LANG", "LC_ALL", "LANGUAGE"].each do |e|
+        env_icon = "⛔️"
+        if ["en_US.UTF-8", "de_DE.UTF-8"].include?(ENV[e]) || !ENV[e]
```
milch: How about we just test that it is ending in `.UTF-8` instead? I've read reports from users using it with all kinds of UTF-8 locales and it fixed their issues.

milch (changes requested):

milch:
I think we should merge this for now since I think it's going to be useful, we can always revert it later if we decide to just set the environment ourselves. There's some other code we'd need to remove too, anyway.

hjanuschka:
yeah i think it is not bad to have locale info in issues - once we decide to set it on startup our self,
i am happy to do a PR and remove the code or rework it - to print all `FASTLANE_|FL_|DELIVER....` env (only if they exists, without value for sure 👍 )


@milch ok i redid it with `end_with?` - UTF8

and removed the broken sign on undefined envs.
and also check if atleast one of the vars has UTF8 otherwise output this table:

### System Locale

milch (approved):

milch:
Thanks, looks good!

PR #6925, opened 2016-11-09
Add Project.supported_platforms method

0xced:

---
Single-scheme projects may support multiple platforms, as explained in [Multiplatform, Single-scheme Xcode Projects](http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/)

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -173,21 +173,25 @@ module FastlaneCore
     end
 
     def mac?
-      # Some projects have different values... we have to look for all of them
-      return true if build_settings(key: "PLATFORM_NAME") == "macosx"
-      return true if build_settings(key: "PLATFORM_DISPLAY_NAME") == "macOS"
-      return true if build_settings(key: "PLATFORM_DISPLAY_NAME") == "OS X"
-      false
+      supported_platforms.include?(:macOS)
     end
 
     def tvos?
-      return true if build_settings(key: "PLATFORM_NAME").to_s.include? "appletv"
-      return true if build_settings(key: "PLATFORM_DISPLAY_NAME").to_s.include? "tvOS"
-      false
+      supported_platforms.include?(:tvOS)
     end
 
     def ios?
-      !mac? && !tvos?
+      supported_platforms.include?(:iOS)
+    end
+
+    def supported_platforms
+      supported_platforms = build_settings(key: "SUPPORTED_PLATFORMS").split
+      platforms = []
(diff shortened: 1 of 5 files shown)


fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -173,21 +173,25 @@ def app_name
...
 
     def ios?
-      !mac? && !tvos?
+      supported_platforms.include?(:iOS)
+    end
+
+    def supported_platforms
+      supported_platforms = build_settings(key: "SUPPORTED_PLATFORMS").split
+      platforms = []
+      platforms << :macOS   if supported_platforms.include?("macosx")
```
milch: What about the others that were previously in the `mac?` method? Or does the `SUPPORTED_PLATFORMS` key always just contain `macosx`? 

```
return true if build_settings(key: "PLATFORM_DISPLAY_NAME") == "macOS"
return true if build_settings(key: "PLATFORM_DISPLAY_NAME") == "OS X"
```

0xced: Yes. The _display_ name changed with Xcode versions but the underlying platform never changed, just like iOS platform is still `iphoneos` and `iphonesimulator` even though iPhone OS was renamed iOS 6 years ago.

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -173,21 +173,25 @@ def app_name
...
+      platforms << :iOS     if (supported_platforms & ["iphonesimulator", "iphoneos"]).any?
+      platforms << :watchOS if (supported_platforms & ["watchsimulator", "watchos"]).any?
+      platforms << :tvOS    if (supported_platforms & ["appletvsimulator", "appletvos"]).any?
+      return platforms
     end
```
mfurtak: Working around Ruby's lack of Sets is always a pain... What do you think about this:

```ruby
supported_platforms = build_settings(key: "SUPPORTED_PLATFORMS").split
supported_platforms.reduce([]) do |acc, val|
  acc << case val
         when "macosx" then :macOS
         when "iphonesimulator", "iphoneos" then :iOS
         when "watchsimulator", "watchos" then :watchOS
         when "appletvsimulator", "appletvos" then :tvOS
         end
end.uniq!
```

mfurtak:
Hi @0xced - there was a change to `faraday` that has caused some of our tests to break. This is fixed on master, now. Could you please rebase onto that to get CI happy again? Thanks!

mfurtak (approved):
Thanks for the improvement! 👍

PR #6676, opened 2016-10-22
Fix commit_version_bump could not find a .xcodeproj, this closes #6675

wangjiejacques:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:
- [x] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Before submitting a pull request, we appreciate if you create an issue first to discuss the change :+1:

…pository's working directory when the current repository is under something like xxxPods/myProjectName

PR #6929, opened 2016-11-09
[fastlane] Version bump

KrauseFx:
* Implement `fastlane_require` to automatically install gems into bundle (#6858)
* Add new `fastlane update_fastlane` action (#6896)
* Support keychains in custom paths and fix non-existing default keychain crash 🔐 (#6887)
* Hide options with no description in docs (#6919)
* Fix commit_version_bump could not find a .xcodeproj (#6676)


milch:
Note: This depends on #6940 and #6939

milch (approved):
🌮

PR #6933, opened 2016-11-09
[fastlane_core] Version bump

KrauseFx:
- Keychain prompts for permission to use private key regardless of access control settings on sierra (#6932)
- Handle iTMSTransporter failure/retry loop (#6926)


milch (approved):
:godmode:

PR #6935, opened 2016-11-09
Fix match to use fastlane_core keychain_importer

milch:
This will fix #6866.

Depends on #6932 being released.

PR 3 of 3 🎉

PR #6856, opened 2016-11-03
Provide more context for choosing your git URL

spanage:
The user will receive an auth error if he/she uses the incorrect type of Git URL, i.e. if he/she is using SSH to authenticate but uses an HTTPS url. This change adds a note reminding the user of this.

PR #6940, opened 2016-11-09
[match] Version Bump

milch:
* Fix hang on CI on macOS Sierra (#6935)
* New `keychain_password` parameter to avoid system keychain prompt 
* support force on nuke (#6886, #6894)
* Provide more context for choosing your git URL (#6856)
* Update internal dependencies


KrauseFx (approved):
🎷

KrauseFx:
In the changelog, as the second item, it would be cool to have the parameter there too

PR #6950, opened 2016-11-10
Fix `fastlane env` command for bundled fastlane

KrauseFx:
Also rename `x` to `current_gem` cc @hjanuschka

Before that, it didn't actually show all fastlane gems, since they don't get required at that point yet
screenshot 2016-11-10 12 50 48

### DIFF
--- a/fastlane/lib/fastlane/environment_printer.rb
+++ b/fastlane/lib/fastlane/environment_printer.rb
@@ -78,6 +78,16 @@ module Fastlane
       env_output
     end
 
+    # We have this as a separate method, as this has to be handled
+    # slightly differently, depending on how fastlane is being called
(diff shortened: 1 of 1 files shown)


hjanuschka:
lgtm, i didn't realize the troubles with the gems not required at this point 😭

milch:
Looks good, but it would be nice if we could also say "This is the bundled version of fastlane" or something somewhere.

KrauseFx:
@milch: Yeah, but this would be a separate PR as it would be another row being added. For now, this fixes the issue with listing the gems. 

Currently we can already see if it's bundled as part of the Ruby path 👍

milch:
Alright, then

milch (approved):
🍕

PR #6952, opened 2016-11-11
[fastlane] Update `fastlane_version` action to show new update message

KrauseFx:
Related to https://github.com/fastlane/fastlane/pull/6953, but no dependency

screenshot 2016-11-10 17 19 21
screenshot 2016-11-10 17 17 41
screenshot 2016-11-10 17 17 25
screenshot 2016-11-10 17 16 39

PR #6953, opened 2016-11-11
[fastlane_core] Move update command to its own method

KrauseFx:
- Better support for bundled fastlane update command
- Little refactoring to make the code nicer
- Don’t show `gem cleanup` message for bundled fastlane
- Add tests for that

Related to https://github.com/fastlane/fastlane/pull/6952, but no dependency

screenshot 2016-11-10 17 19 21
screenshot 2016-11-10 17 17 41
screenshot 2016-11-10 17 17 25
screenshot 2016-11-10 17 16 39

### DIFF
--- a/fastlane_core/lib/fastlane_core/update_checker/changelog.rb
+++ b/fastlane_core/lib/fastlane_core/update_checker/changelog.rb
@@ -19,7 +19,8 @@ module FastlaneCore
           break
         end
 
-        puts "\nPlease update using `#{UpdateChecker.update_command(gem_name: gem_name)}`".green if did_show_changelog
+        puts ""
+        puts "Please update using `#{UpdateChecker.update_command(gem_name: gem_name)}`".green if did_show_changelog
       rescue
         # Something went wrong, we don't care so much about this
       end
--- a/fastlane_core/lib/fastlane_core/update_checker/changelog.rb
+++ b/fastlane_core/lib/fastlane_core/update_checker/changelog.rb
@@ -4,18 +4,22 @@ module FastlaneCore
   class Changelog
     class << self
       def show_changes(gem_name, current_version)
+        did_show_changelog = false
+
         self.releases(gem_name).each_with_index do |release, index|
           next unless Gem::Version.new(release['tag_name']) > Gem::Version.new(current_version)
           puts ""
           puts release['name'].green
           puts release['body']
+          did_show_changelog = true
 
           next unless index == 2
           puts ""
           puts "To see all new releases, open https://github.com/fastlane/#{gem_name}/releases".green
           break
         end
-        puts "\nUpdate using 'sudo gem update #{gem_name.downcase}'".green
+
+        puts "\nPlease update using `#{UpdateChecker.update_command(gem_name: gem_name)}`".green if did_show_changelog
       rescue
         # Something went wrong, we don't care so much about this
       end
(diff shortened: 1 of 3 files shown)


fastlane_core/spec/update_checker_spec.rb
```diff
@@ -36,6 +36,31 @@
       end
     end
 
+    describe "#update_command" do
```
milch: ❤️ Tests

fastlane_core/lib/fastlane_core/update_checker/update_checker.rb
```diff
@@ -47,27 +47,45 @@ def self.show_update_status(gem_name, current_version)
...
         puts '#######################################################################'.green
         puts "# Run `sudo gem cleanup` from time to time to speed up fastlane".green
       end
       puts '#######################################################################'.green
       Changelog.show_changes(gem_name, current_version) unless ENV["FASTLANE_HIDE_CHANGELOG"]
     end
 
+    # The command that the user should use to update their mac
+    def self.update_command(gem_name: nil)
+      gem_name ||= "fastlane"
```
milch: Why use `nil` as default and then set it to fastlane if it's nil?

fastlane_core/lib/fastlane_core/update_checker/changelog.rb
```diff
@@ -4,18 +4,22 @@ module FastlaneCore
...
+          did_show_changelog = true
 
           next unless index == 2
           puts ""
           puts "To see all new releases, open https://github.com/fastlane/#{gem_name}/releases".green
           break
         end
-        puts "\nUpdate using 'sudo gem update #{gem_name.downcase}'".green
+
+        puts "\nPlease update using `#{UpdateChecker.update_command(gem_name: gem_name)}`".green if did_show_changelog
```
milch: `puts "", ...` instead of the `\n` at the beginning of the line

milch (changes requested):

fastlane_core/lib/fastlane_core/update_checker/update_checker.rb
KrauseFx: That's how I did it until now across the fastlane code base. Should I just include it inside the brackets?

fastlane_core/lib/fastlane_core/update_checker/changelog.rb
KrauseFx: So I can do the following? 
```ruby
puts "", "Please update using `#{UpdateChecker.update_command(gem_name..."
```
I feel like this is more confusing, especially for non-Ruby developers, as you have to know how Ruby handles this. `\n` is universally understood by developers.

fastlane_core/lib/fastlane_core/update_checker/update_checker.rb
milch: Yeah, I think that'd make more sense.

fastlane_core/lib/fastlane_core/update_checker/changelog.rb
milch: Then how about this:

```ruby
  puts ""
  puts "Please update using `#{UpdateChecker.update_command(gem_name..."
```

Reason is consistency, we almost never use explicit '\n' in the code

fastlane_core/spec/update_checker_spec.rb
KrauseFx: The best

KrauseFx:
Updated 👍

milch (approved):

milch:
🎷

PR #6955, opened 2016-11-11
[fastlane] output contained in env

hjanuschka:
as proposed here: https://github.com/fastlane/fastlane/pull/6950#issuecomment-259813816

### DIFF
--- a/fastlane/lib/fastlane/environment_printer.rb
+++ b/fastlane/lib/fastlane/environment_printer.rb
@@ -186,7 +186,7 @@ module Fastlane
         "Host" => "#{product} #{version} (#{build})",
         "Ruby Lib Dir" => RbConfig::CONFIG['libdir'],
         "OpenSSL Version" => OpenSSL::OPENSSL_VERSION,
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -185,7 +185,8 @@ def self.print_system_environment
         "Installation Source" => $PROGRAM_NAME,
         "Host" => "#{product} #{version} (#{build})",
         "Ruby Lib Dir" => RbConfig::CONFIG['libdir'],
-        "OpenSSL Version" => OpenSSL::OPENSSL_VERSION
+        "OpenSSL Version" => OpenSSL::OPENSSL_VERSION,
+        "Contained Version" => Helper.contained_fastlane?.to_s
```
milch: `"Is contained"` would be better, since the value will be `true` or `false`

milch (changes requested):

hjanuschka:
@milch feedback addressed.

milch (approved):

Issue #6900, opened 2016-11-07
[match] Import code signing certificate to match repo


fabiomassimo:
# What

As `match` user I love that all my certificates and provisioning profiles can be automated, although when using `match` it is not possible to import an existing code signing certificate to the repo. 
The suggested solution is to use `match nuke` to remove all certificate not managed by `match`, since creation, and recreate them. Although this is a valid solution, that aims to keep consistency in the repo, it can easily be not ideal for some projects.

A command like `match import {certificate}` makes it easy to import a P12 certificate (for instance exported from the Keychain) to a `match` repo.

# Why

1. The team has distributed apps via TestFlight and it has reached the number of iOS Distribution certificate that can be issued. Since they want to start using `match` they soon find out that they should run `match nuke` because no new distribution certificate can be issued. This solution, would make all the distributed TestFlight builds stop working. 
Thanks to `match import` the team can easily move the project to start using `match` by importing the iOS Distribution certificate they already own in their Keychain.

2. https://fastlane-tools.slack.com/archives/questions/p1478178628000681

3. https://fastlane-tools.slack.com/archives/questions/p1475762422000253

# TODO

The `import` feature must take care of:

- make sure the certificate to import exists in the Apple Developer Portal.
- the certificate type is properly identified such that can be added to the repo in the correct folder tree structure.
- the certificate to import gets renamed with the id assigned by the Apple Developer Portal

# Spoiler alert

I have implemented this functionality already but before opening a PR I'd love to have your opinion here.
For more details: https://github.com/fabiomassimo/fastlane/tree/match_import

TKBurner:
@fabiomassimo Thanks for reaching out. I'm bumped this with the core team to see what the interest is. I am really happy to see what you built 👍

milch:
@fabiomassimo Thanks for your detailed issue! I am very open to this feature as I've run into this before myself. 

However, we always have to be careful when implementing new features. This especially affects core features that could have a big impact on how users are using fastlane, like this one. 

In my experience, while it is a nuisance to send out another build that testers need to update to, fastlane does make it very easy. Review times being as low as they are right now also means that the new build would be available to testers VERY soon. As such, I am not entirely convinced yet that this would fulfill the need of a large amount of users. 

What do you think?

fabiomassimo:
@milch You are right, this is not an easy feature and I'd like to gather as more feedbacks as possible. Thank you for chiming in.

[...]

PR #6972, opened 2016-11-12
[match] Import P12 codesigning certificate to match's repo.

fabiomassimo:
# What

Implementation for #6900

# Notes

This is my first draft for implementing the `match import` feature.
In parallel, I'm working on a PR for fastlane.docs, to explain to the user how to export a code signing certificate to P12 and properly use this feature.

Before this gets merged it is required to review the related fastlane docs page and update the `match`'s __README.md__ to include a link to the fastlane docs for the `import` feature.


match/lib/match/spaceship_ensure.rb
```diff
@@ -50,6 +50,34 @@ def certificate_exists(username: nil, certificate_id: nil)
...
+      UI.message "Looking up availabe code signing certificates in the Apple Developer Portal..."
+
+      matching_certificates_on_portal = Spaceship.certificate.all.find_all do |cert|
+        (cert.expires == pkcs12.certificate.not_after)
+      end
+
+      if !matching_certificates_on_portal.nil? and matching_certificates_on_portal.count > 1
+        UI.warning "Found more than one eligible certificate in the Developer Portal that matches the certificate to import, trying to match on `owner_id` too..."
+
+        cert_cn = pkcs12.certificate.subject.to_s.split("CN=")[1].split("/")[0]
```
milch: This seems a bit hacky to me ... Maybe instead of trying to compare `owner_id`s (which should be the same for each org) we should download the matching certs here and compare serials?

match/lib/match/utils.rb
```diff
@@ -53,5 +53,20 @@ def self.environment_variable_name_profile_name(app_identifier: nil, type: nil)
...
+    # Load a P12 certificate file as OpenSSL:PKCS12 object.
+    def self.load_pkcs12_file(path, password = nil)
+      UI.user_error!("Certificate #{path} couldn't be found") unless File.exist?(path)
+
+      begin
+        p12 = OpenSSL::PKCS12.new(File.read(path), password)
+      rescue OpenSSL::PKCS12::PKCS12Error
+        UI.user_error!("The certificate password is either incorrect or missing")
+      rescue StandardError => e
+        UI.user_error!("Certificate #{path} couldn't be loaded: #{e.message}")
```
milch: This is probably not a user error, so let's use `UI.crash!` here instead

milch (dismissed):
Hi @fabiomassimo, we are very sorry for the huge delay in reviewing your PR 🙁

I just looked through everything, your docs changes seem great, and other than that there are just a couple small changes I would make. On top of that, you'd also need to rebase to current master before we can merge this 👍 

Thanks for your patience with this, we really appreciate it!

bernikowich-cedoni:
Waiting for this feature.

biscuitehh:
@mfurtak @milch would it be best to continue working on this PR or start a new one? I'm willing to get this fixed up as it would be super useful for my organization

mkko:
Just curious, this only covers importing existing certs. Any possibility of adding provisioning profile import here as well? At least the encryption process would be the same.

From my point of view the most valuable use for this feature is if you don't happen to have control over the certs or provisioning profiles and you still want to share them with the team.

biscuitehh:
@mkko it looks like the PR only covers certs but I definitely need it to handle provisioning profiles as well.

fabiomassimo:
@biscuitehh  I'd advice to continue this one since you can take advantage of some convenience methods and helpers that I introduced. The migration to fastlane 2.0 shouldn't be painful and I think the ideas shared by @milch can be easily processed. Unfortunately I don't think I can work on this until next month. If you want to take ownership on this feel free to and let me know such that we can close this and move forward, otherwise I'll jump on it as soon as I'll have more time for it.

@mkko the process of encrypt and un-crypt is the same for mobile provisioning profiles but it needs to be verified with a different flow against the Apple Developer Portal API. My implementation specifically queries only certificates. 
I'd put more thoughts on this since mobile provisioning profiles ca be easily updated while certificates have a combination of public and private key that is hard to manage. Plus, if a certificate is revoked many apps, signed by using that certificate, can crash on launch that's why I wanted to focus on certificates more than mobile provisioning profiles.

biscuitehh:
Thanks, @fabiomassimo - one question, though: is it possible to match the certificate with what's already in the repo as an option alongside the Apple Developer Portal API? While Match heavily relies on that API, my use case involves manually creating certificates and managing them as my company's clients do not always give us access to their developer accounts.

mfurtak:
I think this PR is stalled out, so I'm going to close it. @biscuitehh, if you wanted to adapt it into a new PR please feel free 👍

razum2um:
@mfurtak I rebased that branch on current master and was able to import my certificates, do you wish me to open PR?

MajinRaph:
This feature is so much needed :+1: :pray:

okonon:
this would be so cool

razum2um:
@mfurtak please, proceed with https://github.com/fastlane/fastlane/pull/9752

@fabiomassimo thanks a lot for the basis ❤️ , I slightly updated matching logic as requested in comments to be sure in the match.

Issue #6482, opened 2016-10-10
commit_version_bump settings doesn't read SRC_ROOT


dombarnes:
### New Issue Checklist
- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

When using the param `settings: true` in a fastfile, it does not appear to include ${SRC_ROOT} from Xcode in the file path, so fails to locate if this is not in the root of the git repo
See [commit_version_bump.rb#L94](https://github.com/fastlane/fastlane/blob/6769d5c930d292bf31ab4be9ec80a7c7ccd30899/fastlane/lib/fastlane/actions/commit_version_bump.rb#L94)
##### Complete output when running fastlane, including the stack trace and command used

Note: /Users/domster83/workprojects/TrilbyTV-iOS/TrilbyTVPlayer is my Xcode project root.

```
[16:55:32]: fastlane finished with errors

[!] Exit status of command 'git add /Users/domster83/workprojects/TrilbyTV-iOS/TrilbyTVPlayer/TrilbyTVPlayer.xcodeproj/project.pbxproj /Users/domster83/workprojects/TrilbyTV-iOS/TrilbyTVPlayer/TrilbyTVPlayer/TrilbyTVPlayer-Info.plist /Users/domster83/workprojects/TrilbyTV-iOS/TrilbyTVPlayer/TrilbyTVPlayer/TrilbyTVPlayer-Dev-Info.plist /Users/domster83/workprojects/TrilbyTV-iOS/Settings.bundle/Root.plist' was 128 instead of 0.
fatal: pathspec '/Users/domster83/workprojects/TrilbyTV-iOS/Settings.bundle/Root.plist' did not match any files
```
### Configuration Files

Please copy the complete content of your `Fastfile` and any other configuration files you use below:

Fastfile:

``` ruby
ommit_version_bump(
      message: "Deploying version #{friendly_version}.",
      xcodeproj: 'TrilbyTVPlayer.xcodeproj',
      force: true,
      settings: true
    )
```
### Environment

fastlane version (run `fastlane -v`):
1.105.0
Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?
no
Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?
rbenv

TKBurner:
[...]

PR #6997, opened 2016-11-14
Allow commit_version_bump to find the Settings.bundle in the project

jdee:

- [x ] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
 Some tests are failing in master and continue to fail here. This action actually has no tests. I may add some.
- [x ] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x ] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x ] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Please see issue #5758.

This change allows the `commit_version_bump` action to infer the location of the Settings.bundle from the project. It also adds a new optional :settings_file parameter that allows the user to specify a plist file in the Settings.bundle other than Root.plist.

### DIFF
--- a/fastlane/lib/fastlane/actions/commit_version_bump.rb
+++ b/fastlane/lib/fastlane/actions/commit_version_bump.rb
@@ -67,6 +67,14 @@ module Fastlane
         expected_changed_files = []
         expected_changed_files << pbxproj_path
         expected_changed_files << info_plist_files
+
+        if params[:settings]
+          file = params[:settings_file]
+
+          settings_file_pathname = Pathname.new settings_bundle_file_path(xcodeproj_path, file)
+          expected_changed_files << settings_file_pathname.relative_path_from(repo_pathname).to_s
+        end
+
         expected_changed_files.flatten!.uniq!
 
         # get the list of files that have actually changed in our git workdir
@@ -89,10 +97,6 @@ module Fastlane
             ].join("\n")
             UI.user_error!(error)
           end
-        end
-
-        if params[:settings]
-          expected_changed_files << 'Settings.bundle/Root.plist'
         end
 
         # get the absolute paths to the files
@@ -148,6 +152,12 @@ module Fastlane
                                        optional: true,
                                        default_value: false,
                                        is_string: false),
+          FastlaneCore::ConfigItem.new(key: :settings_file,
+                                       env_name: "FL_COMMIT_SETTINGS_FILE",
+                                       description: "A plist file in Settings.bundle to include if :settings is true",
+                                       optional: true,
+                                       default_value: "Root.plist",
+                                       type: String),
           FastlaneCore::ConfigItem.new(key: :ignore,
                                        description: "A regular expression used to filter matched plist files to be modified",
                                        optional: true,
@@ -192,6 +202,19 @@ module Fastlane
 
       def self.category
         :source_control
+      end
+
+      class << self
+        private
+        def settings_bundle_file_path(xcodeproj_path, settings_file_name)
+          require 'xcodeproj'
+          project = Xcodeproj::Project.open xcodeproj_path
+          settings_bundle = project.files.find { |f| f.path =~ /Settings.bundle/ }
+          raise "No Settings.bundle in project" if settings_bundle.nil?
+
+          project_parent = File.dirname xcodeproj_path
+          File.join project_parent, settings_bundle.path, settings_file_name
+        end
       end
     end
   end
--- a/fastlane/lib/fastlane/actions/commit_version_bump.rb
+++ b/fastlane/lib/fastlane/actions/commit_version_bump.rb
@@ -69,10 +69,10 @@ module Fastlane
         expected_changed_files << info_plist_files
 
         if params[:settings]
-          file = params[:settings_file]
-
-          settings_file_pathname = Pathname.new settings_bundle_file_path(xcodeproj_path, file)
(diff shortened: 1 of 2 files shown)


jdee:
This is also related to #6482.

jdee:
There are probably not too many cases where an app might include multiple settings plist files in a version bump, but it doesn't seem too far-fetched to think that someone might want to update their `Root.plist` as well as an `About.plist` and so on.

A common Ruby idiom is to allow different types to be passed to an argument, e.g.

```ruby
commit_version_bump xcodeproj: "MyProject.xcodeproj", settings: true # Root.plist
commit_version_bump xcodeproj: "MyProject.xcodeproj", settings: "About.plist"
commit_version_bump xcodeproj: "MyProject.xcodeproj", settings: [ "Root.plist", "About.plist" ]
```

This might be simpler than adding multiple new arguments like `settings_file:` (currently in this branch) and `settings_files:`. But I'm not sure if the `ConfigItem` can support this. I suppose I could check. Either way, it wouldn't be hard to support including multiple files. 

Please let me know if further work is needed for this PR to be merged. Thanks.

ohwutup:
Hey @jdee ! Thanks for the PR, apologies for the delay in reviewing! 

This looks good... I think it would be great to have everything in one option (`ConfigItem`). I don't think you should have a problem allowing it to be multiple types by not specifying the type and then just inspecting the type at runtime.

Let me know if you have any questions!

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -66,6 +66,14 @@ def self.run(params)
         expected_changed_files = []
         expected_changed_files << pbxproj_path
         expected_changed_files << info_plist_files
+
+        if params[:settings]
+          file = params[:settings_file]
```
ohwutup: Would be great to combine these.

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -147,6 +151,12 @@ def self.available_options
                                        optional: true,
                                        default_value: false,
                                        is_string: false),
+          FastlaneCore::ConfigItem.new(key: :settings_file,
+                                       env_name: "FL_COMMIT_SETTINGS_FILE",
+                                       description: "A plist file in Settings.bundle to include if :settings is true",
+                                       optional: true,
+                                       default_value: "Root.plist",
+                                       type: String),
```
ohwutup: Just skip the type specification here.

ohayon (changes requested):

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -192,6 +202,20 @@ def self.example_code
...
+        private
+
+        def settings_bundle_file_path(xcodeproj_path, settings_file_name)
+          require 'xcodeproj'
+          project = Xcodeproj::Project.open xcodeproj_path
+          settings_bundle = project.files.find { |f| f.path =~ /Settings.bundle/ }
+          raise "No Settings.bundle in project" if settings_bundle.nil?
+
+          project_parent = File.dirname xcodeproj_path
+          File.join project_parent, settings_bundle.path, settings_file_name
```
KrauseFx: Please surround this with `(` and `)`

jdee:
Hi @ohwutup. Sorry for the delay. I had an app to get out. Thanks for the feedback. I think this is much nicer now. In fact these are all now equivalent:

```ruby
commit_version_bump xcodeproj: "MyApp.xcodeproj", settings: true
commit_version_bump xcodeproj: "MyApp.xcodeproj", settings: "Root.plist"
commit_version_bump xcodeproj: "MyApp.xcodeproj", settings: ["Root.plist"]
```

I also added some tests for the methods I added. Please let me know if there's anything left to be done. Thanks.

ohwutup:
Thanks for making those changes @jdee ! If you don't mind, can we just change one more little thing about this? I think that `settings` is a bit of an ambiguous name for the option. Can we change that to be something a bit more descriptive?

jdee:
Hi @ohwutup. I'd be happy to change that. Any suggestions? Maybe `include_settings`, `settings_plist(s)`, `settings_file(s)`, `include_settings_file(s)`, `include_settings_plists(s)`?

Also, what if someone is using the current syntax. I know it was broken, but in a couple of apps at my last job, I was using `settings: true, force: true` and I just put the Settings.bundle at the top of the repo. Maybe the action could report an error if the old argument is supplied and say, "Please use the new syntax." Or both could be supported for now. Or just forget about it and change it?

ohwutup:
Oh! My bad @jdee! Was there already a flag called `settings`? Sorry, there are so many options for different tools I must have gotten mixed up! If there was already a `settings` we should just go with that! Really sorry for the back and forth!

jdee:
No worries, @ohwhtup. I thought that might be the case. I'll revert that commit and submit a PR to the docs repo. Thanks.

ohayon (approved):

ohwutup:
Thanks for the PR and working through all the changes @jdee !! 🚀

jdee:
My pleasure. Fastlane is a nice tool. I'm glad to be able to contribute.

Issue #6986, opened 2016-11-14
Resign can not find provision profile at path '~/Library/MobileDevice/Provisioning Profiles/'


ruslanfutureuniverse:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Resign fails to find provisioning profile at path '~/Library/MobileDevice/Provisioning Profiles/something.mobileprovision'. But '/Users/user/Library/MobileDevice/Provisioning Profiles/something.mobileprovision' works. And this is fragile.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

bundle exec fastlane resign_ipa
[14:11:00]: -------------------------------------------------
[14:11:00]: --- Step: Verifying required fastlane version ---
[14:11:00]: -------------------------------------------------
[14:11:00]: Your fastlane version 1.108.0 matches the minimum requirement of 1.107  ✅
[14:11:00]: ------------------------------
[14:11:00]: --- Step: default_platform ---
[14:11:00]: ------------------------------
[14:11:00]: Driving the lane 'ios resign_ipa' 🚀
[14:11:00]: --------------------
[14:11:00]: --- Step: resign ---
[14:11:00]: --------------------
[14:11:00]: Error setting value '~/Library/MobileDevice/Provisioning Profiles/f100dffe-dca0-4a73-9363-9bec3de7aa19.mobileprovision' for option 'provisioning_profile'
[14:11:00]: You passed invalid parameters to 'resign'.
[14:11:00]: Check out the error below and available options by running `fastlane action resign`
[14:11:00]: Variable Dump:
[14:11:00]: {:DEFAULT_PLATFORM=>:ios, :PLATFORM_NAME=>:ios, :LANE_NAME=>"ios resign_ipa"}
[14:11:00]: Couldn't find provisiong profile at path '~/Library/MobileDevice/Provisioning Profiles/f100dffe-dca0-4a73-9363-9bec3de7aa19.mobileprovision'

+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 💥    | resign                              | 0           |
[...]

PR #7004, opened 2016-11-15
[sigh] expand_path on resign

hjanuschka:
expand_path on resign so `~/Downloads/1.mobileprovision` will work.

as discussed here: https://github.com/fastlane/fastlane/issues/6986

### DIFF
--- a/sigh/lib/sigh/resign.rb
+++ b/sigh/lib/sigh/resign.rb
@@ -27,7 +27,10 @@ module Sigh
       # validate that we have valid values for all these params, we don't need to check signing_identity because `find_signing_identity` will only ever return a valid value
       validate_params(resign_path, ipa, provisioning_profiles)
       entitlements = "-e #{entitlements.shellescape}" if entitlements
-      provisioning_options = provisioning_profiles.map { |fst, snd| "-p #{[fst, snd].compact.map(&:shellescape).join('=')}" }.join(' ')
+      provisioning_options = provisioning_profiles.map do |fst, snd|
+        fst = File.expand_path(fst)
+        "-p #{[fst, snd].compact.map(&:shellescape).join('=')}"
+      end.join(' ')
       version = "-n #{version}" if version
       display_name = "-d #{display_name.shellescape}" if display_name
       short_version = "--short-version #{short_version}" if short_version
--- a/sigh/lib/sigh/resign.rb
+++ b/sigh/lib/sigh/resign.rb
@@ -35,7 +35,7 @@ module Sigh
       #     "at.fastlane.today" => "/folder/mobile.mobileprovision"
       #   }
       # or an array
-      # ( resign.sh also takes "-p /folder/mobile.mobileprovision" as a param
+      # (resign.sh also takes "-p /folder/mobile.mobileprovision" as a param)
       #   [
       #        "/folder/mobile.mobileprovision"
       #   ]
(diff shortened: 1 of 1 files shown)


sigh/lib/sigh/resign.rb
```diff
@@ -27,7 +27,10 @@ def resign(ipa, signing_identity, provisioning_profiles, entitlements, version,
       # validate that we have valid values for all these params, we don't need to check signing_identity because `find_signing_identity` will only ever return a valid value
       validate_params(resign_path, ipa, provisioning_profiles)
       entitlements = "-e #{entitlements.shellescape}" if entitlements
-      provisioning_options = provisioning_profiles.map { |fst, snd| "-p #{[fst, snd].compact.map(&:shellescape).join('=')}" }.join(' ')
+      provisioning_options = provisioning_profiles.map do |fst, snd|
```
milch: Do you have any idea what first and second are supposed to be here? If yes, can we make the name more explicit?

hjanuschka: @milch  i am not sure, in my test cases this was empty, but i am afraid of removing it, probably it breaks something?

milch: Yeah I'm not sure it ever does something, looking through the code the `provisioning_profiles` var is:

- Is a single string at first [(here)](https://github.com/hjanuschka/fastlane/blob/07aa9e4346d765d30776181e305dff384c73ef04/sigh/lib/sigh/resign.rb#L75)
- Gets passed to the main method [here](https://github.com/hjanuschka/fastlane/blob/07aa9e4346d765d30776181e305dff384c73ef04/sigh/lib/sigh/resign.rb#L8)
- Is then wrapped in an array [here](https://github.com/hjanuschka/fastlane/blob/07aa9e4346d765d30776181e305dff384c73ef04/sigh/lib/sigh/resign.rb#L24)
- As a result, `snd` [here](https://github.com/hjanuschka/fastlane/blob/07aa9e4346d765d30776181e305dff384c73ef04/sigh/lib/sigh/resign.rb#L30) should always be nil, since at this point `provisioning_profiles` is an array of single values

What do you think? Does that sound correct?

hjanuschka: @milch  ok got you - i think it is right so it could be simplified to:

```ruby
provisioning_options = provisioning_profiles.map do |profile|
        profile = File.expand_path(profile).shellescape
        "-p #{profile}"
end.join(' ')
```

i am just concerned about the `join("=")`  - where and why is the `=`  required? (some commandline argument, passing stuff?)

in a sample it works:
```ruby
require "shellwords"
provisioning_profiles = ["a", "b", "~/Downloads/c"]
provisioning_options = provisioning_profiles.map do | profile |
        profile = File.expand_path(profile).shellescape
        "-p #{profile}"
end.join(' ')

puts provisioning_options
```

hjanuschka: @milch - sorry my reply was to fast.
after looking at `resign.sh` - wich actually gets called, there is a documentation about the whole `-p` paramater.

it can be passed either:

  * `-p file.mobileprovision`
  *  `-p com.nested-app=nested-app.mobileprovision`

so i think the `snd` param is only set, once you resign your extensions, like today-extension.


sounds correct?

milch: Ah, interesting. I get it now, thanks for investigating! Any idea how we could make this a bit clearer? Maybe a comment before the code block?

hjanuschka:
@milch - added a comment to clarify passed param.
also renamed `fst, `snd` -> `app_id`, `app_id_prov`

milch:
I'm not sure that comment is always correct, is it really always a hash? I think it's either a hash or an array, though I'm not a 100% on that.

hjanuschka:
@milch  updated comment

milch (approved):

PR #7011, opened 2016-11-15
Adding bitrise.io git_branch ENV variable check

ivanbruel:
The git_branch action seems to return an empty string when run in bitrise.io.
I added the ENV check for BITRISE_GIT_BRANCH (listed in BITRISE_GIT_BRANCH).


milch:
Good catch, thanks!  

Also, congrats on your first fastlane PR, we are looking forward to your future contributions! 🚀

milch (approved):
🤖

Issue #6995, opened 2016-11-14
Retry app store upload (pilot/deliver)


kluepfel:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

The App Store is pretty flaky recently. Submitting from Xcode or using fastlane sometimes just fails for no particular reason. Just retrying usually does the trick. 

Is there a way to make the `pilot` and `deliver` commands retry once or twice on failure?

KrauseFx:
Could you share the full output and stack trace when that happens? _fastlane_ should already cover some of those issues.

kluepfel:
There's no stack trace. I've been trying to reproduce the error with --verbose, but so far it succeeds every time I used --verbose :/

Here's the full output:
```
Chimp:appName kluepfel$ fastlane release
[22:21:27]: -------------------------------------------------
[22:21:27]: --- Step: Verifying required fastlane version ---
[22:21:27]: -------------------------------------------------
[22:21:27]: Your fastlane version 1.108.0 matches the minimum requirement of 1.107.0  ✅
[22:21:27]: ------------------------------
[22:21:27]: --- Step: default_platform ---
[22:21:27]: ------------------------------
[22:21:27]: Driving the lane 'ios release' 🚀
[22:21:27]: -----------------
[22:21:27]: --- Step: gym ---
[22:21:27]: -----------------
[22:21:27]: Using legacy build system - waiting for radar to be fixed: https://openradar.appspot.com/radar?id=4952000420642816
[22:21:27]: Successfully loaded '/Users/kluepfel/companyName/appName-iOS/appName/fastlane/Gymfile' 📄
[22:21:27]: No values defined in './fastlane/Gymfile'
[22:21:27]: Installing provisioning profile...
[22:21:27]: Successfully loaded '/Users/kluepfel/companyName/appName-iOS/appName/fastlane/Gymfile' 📄
[22:21:27]: No values defined in './fastlane/Gymfile'
[22:21:27]: $ xcodebuild -list -project ./appName.xcodeproj
[22:21:29]: $ xcodebuild clean -showBuildSettings -scheme appName -project ./appName.xcodeproj

+---------------------------+----------------------------------------------------------------------+
|                                      Summary for gym 1.12.0                                      |
+---------------------------+----------------------------------------------------------------------+
| scheme                    | appName                                                          |
[...]

PR #7012, opened 2016-11-15
[deliver,pilot,fastlane_core] itunes transporter retry ♻️

hjanuschka:
as proposed here: https://github.com/fastlane/fastlane/issues/6995

```ruby
deliver(retry: 3)
pilot(retry: 3)
```

will retry it 3 times.


KrauseFx:
I'm not a fan of adding a new option for something that should just work. How about we just default to up to 3 retries when it fails? However I can see this being very inefficient when the upload fails because of a user error (e.g. build was already uploaded). 
I don't think we should go ahead with this. 

Thanks for taking the time to work on this though 👍

hjanuschka:
thx anyway - you are right - it will be a pain when it comes to ITMS errors.
maybe we should suggest the user in the issue, that he should do a `begin` `rescue` `until` arround pilot/deliver - so he atleast gets his troubles sorted?


i guess it is mostly of itc api having troubles since a few weeks.

PR #7035, opened 2016-11-15
[gym] Improve build error message with more debug information

KrauseFx:
screenshot 2016-11-15 15 10 34

Every time the build fails, it will now show a summary of the build environment, in particular the Xcode path and the SDK, since many people don't have the Xcode path set correctly. 

I don't know if we should show all that information about how to set the path though, since it's shown on every build failure, since we can't detect if it was caused by a wrong SDK or something else.


Sam0711er:
Ha, I have no clue where the idea for that could've arisen from 😊❤️👏

milch (approved):
Looks good 👍

PR #5982, opened 2016-09-02
Added the ability to suppress echo when entering sensitive parameters…

SandyChapman:
… such as passwords.

PR #7037, opened 2016-11-16
Anonymize paths in `fastlane env` output

0xced:

---

This way, the username is not leaked. Also, it makes it easy to visually spot when a path is in the user’s home directory.

### DIFF
--- a/fastlane/lib/fastlane/environment_printer.rb
+++ b/fastlane/lib/fastlane/environment_printer.rb
@@ -207,15 +207,15 @@ module Fastlane
         "Ruby" => RUBY_VERSION,
         "Bundler?" => Helper.bundler?,
         "Git" => `git --version`.strip.split("\n").first,
-        "Installation Source" => $PROGRAM_NAME,
+        "Installation Source" => anonymized_path($PROGRAM_NAME),
         "Host" => "#{product} #{version} (#{build})",
-        "Ruby Lib Dir" => RbConfig::CONFIG['libdir'],
+        "Ruby Lib Dir" => anonymized_path(RbConfig::CONFIG['libdir']),
         "OpenSSL Version" => OpenSSL::OPENSSL_VERSION,
         "Is contained" => Helper.contained_fastlane?.to_s
       }
 
       if Helper.mac?
-        table_content["Xcode Path"] = Helper.xcode_path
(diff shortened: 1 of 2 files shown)


hjanuschka:
i really like that. recently was thinking about doing a `UI.input`  before copy&paste - and asking the user for strings that they wan't to be removed? like (app identifier ...) - what do you think about it @0xced

0xced:
@hjanuschka This is a good idea, but it should be done in another pull request IMHO. I wonder how much data we could anonymize automatically, without needing `UI.input`.

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -270,6 +270,10 @@ def self.print_fastlane_files
       env_output
     end
 
+    def self.anonymized_path(path)
+      return ENV['HOME'] ? path.gsub(/^#{ENV['HOME']}/, '~') : path
```
milch: In the super edgy edge case that the user's home dir is the prefix of other folders this is gonna output garbage. Example: `$HOME=/work`, `anonymized_path("/workspace/whatever/fastlane")` => `~space/whatever/fastlane`. 

Might be more applicable to linux where the user's home dir is not generally inside the `/Users/` folder.

milch (changes requested):

hjanuschka:
@milch @0xced 

how about this PR here -> https://github.com/fastlane/fastlane/pull/5982
we could refactor all options and set the flags on all critical parameters, and use this in some way to anonymize?

milch:
@hjanuschka Yeah, I've thought about something like that before. Sounds good to me, @0xced do you have any thoughts about that?

0xced:
I updated the code, and added a [spec](https://github.com/0xced/fastlane/blob/8a93790d8ecee25e9b6b8e83499592b6e7443c41/fastlane/spec/env_spec.rb#L39-L42).

With $HOME == /Users/john
`/Users/john/.fastlane/bin/bundle/bin/fastlane` → `~/.fastlane/bin/bundle/bin/fastlane`
`/Users/john` → `~`
`/Users/john/` → `~/`

With $HOME == /work
`/workspace/project/test` → `/workspace/project/test`

milch (approved):
Looks good now! Thanks for your contribution 🚀

Issue #7046, opened 2016-11-16
The unlock_keychain action is unaware of the new Sierra keychain location


jverdi:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
In `fastlane/fastlane/lib/fastlane/actions/unlock_keychain.rb`, the new Sierra keychain path of `~/Library/Keychains/login.keychain-db` is not included in the possible keychain locations. This results in failure when the command is run:

`Could not find the keychain file in: ["login", "~/Library/Keychains/login", "~/Library/Keychains/login.keychain"]`

### Environment

| Gem                 | Version | Update-Status |
| ------------------- | ------- | ------------- |
| credentials_manager | 0.16.2  | ✅ Up-To-Date  |
| fastlane_core       | 0.56.0  | ✅ Up-To-Date  |
| spaceship           | 0.37.0  | ✅ Up-To-Date  |
| deliver             | 1.15.0  | ✅ Up-To-Date  |
| snapshot            | 1.16.3  | ✅ Up-To-Date  |
| frameit             | 3.0.0   | ✅ Up-To-Date  |
| pem                 | 1.4.0   | ✅ Up-To-Date  |
| cert                | 1.4.4   | ✅ Up-To-Date  |
| sigh                | 1.11.2  | ✅ Up-To-Date  |
| produce             | 1.3.0   | ✅ Up-To-Date  |
| gym                 | 1.12.0  | ✅ Up-To-Date  |
| pilot               | 1.12.0  | ✅ Up-To-Date  |
| scan                | 0.14.1  | ✅ Up-To-Date  |
| supply              | 0.7.1   | ✅ Up-To-Date  |
| match               | 0.11.0  | ✅ Up-To-Date  |
| fastlane            | 1.109.0 | ✅ Up-To-Date  |

hjanuschka:
on my machines i symlinked the keychain 😢  - anyway here is the PR - for adding `-db`  to availaible paths.

https://github.com/fastlane/fastlane/pull/7048

jverdi:
Yup, as a workaround, a symlink works, as does supplying the full path to the new keychain to your Fastfile's `unlock_keychain` action. But, this should work out of the box.

milch:
Linking back from here to the other 3 PRs -> #7075, #7076, #7077

[...]

PR #6460, opened 2016-10-07
[match] make match pass the keychain name to cert

iCarambaa:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:
- [ x] Run `rspec` for all tools you modified
- [ x] Run `rubocop -a` to ensure the code style is valid
- [ x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

This fixes #6375 . Match looks up the keychain path by its name and passes it to cert. "== 0" has been replaced with ".zero?" by rubocop.

PR #7048, opened 2016-11-17
[fastlane] sierra keychain path

hjanuschka:
as proposed here: https://github.com/fastlane/fastlane/issues/7046


milch:
Hmm, we discussed here #6460 if we should put the keychain-finding into the core so we can access it from everywhere. Since this is now the second time this has come up, maybe now would be the time to do this.

hjanuschka:
@milch - ok so we should continue with: #6460, if so lets close this PR  - as long as the underlying issue is solved, i am fine with it, right now users might have troubles using `unlock_keychain` on a fresh deployed sierra machine.

milch:
@hjanuschka Yeah I think the path will be: merge #6460 -> new PR w/ keychain lookup in fastlane_core -> new PR for this issue

hjanuschka:
@milch  as #6460  has been merged, i did the 3 follow up PR's to solve #7046

Issue #7050, opened 2016-11-17
Pem: Add option to download push certificate if it exists


revolter:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Trying to create an already existing push certificate throws there warnings:

> [11:11:32]: You already have a push certificate, which is active for more than 30 more days. No need to create a new one
> [11:11:32]: If you still want to create a new one, use the --force option when running PEM.

It would be great if there was an option that would let you only download it, if it exists. So something like:

> [11:11:32]: If you only want to download it, use the --download option when running PEM.

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                        |
| ------------------- | -------------------------------------------- |
| OS                  | 10.11.6                                      |
| Ruby                | 2.0.0                                        |
| Bundler?            | true                                         |
| Git                 | git version 2.9.3 (Apple Git-75)             |
| Installation Source | /Users/Iulian/.gems/bin/fastlane             |
| Host                | Mac OS X 10.11.6 (15G1004)                   |
| Ruby Lib Dir        | /Users/Iulian/.rbenv/versions/2.0.0-p353/lib |
| OpenSSL Version     | OpenSSL 1.0.2e 3 Dec 2015                    |
| Is contained        | false                                        |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/  |
| Xcode Version       | 8.1                                          |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     |             |   |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE |             |   |

### fastlane gems

| Gem                 | Version | Update-Status |
| ------------------- | ------- | ------------- |
| credentials_manager | 0.16.2  | ✅ Up-To-Date  |
| fastlane_core       | 0.56.0  | ✅ Up-To-Date  |
| spaceship           | 0.37.0  | ✅ Up-To-Date  |
| cert                | 1.4.4   | ✅ Up-To-Date  |
| deliver             | 1.15.0  | ✅ Up-To-Date  |
| frameit             | 3.0.0   | ✅ Up-To-Date  |
[...]

Issue #4309, opened 2016-04-21
Download P12 certs using PEM action in every build. 


marcomorain:
I'd love to be able to download exiting P12 certs using `pem`. 

Currently we use PEM to create new certs, and we upload the cert to our server when a new cert is created. If the upload fails for any reason, the cert is created, but we don't have a copy on our server. I then have to manually download the cert from the developer portal. 

This would be awesome if PEM would download the cert if it doesn't create one, so that I can upload the cert to our server on every build. 

This would allow me to build a self-healing system whereby if a build fails and I don't get the cert uploaded, I can be confident that the next time I push I will get the cert. 

Essentially I want the new_profile callback to fire on each invocation of the PEM action.

TKBurner:
@marcomorain Thanks for reaching out. I'll let the team know about this idea. Hope that you have a great weekend 👍

marcomorain:
👍👍😎
 
--
Marc O'Morain
marc@marcomorain.com
 
 
 
On Fri, 22 Apr 2016, at 21:44, Todd Burner wrote:
> @marcomorain[1] Thanks for reaching out. I'll let the team know about
> this idea. Hope that you have a great weekend 👍
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub[2]
>

Links:

  1. https://github.com/marcomorain
  2. https://github.com/fastlane/fastlane/issues/4309#issuecomment-213582705

KrauseFx:
Hey @marcelofabri, sorry for the late reply. That's actually not possible, since the certificate is private and not stored on Apple's server. There is no way to retrieve an existing push certificate (p12 file) from Apple. If there is, please let me know 👍

PR #7052, opened 2016-11-17
[pem] download certificates if they exist 🎟

hjanuschka:
as described here -> https://github.com/fastlane/fastlane/issues/7050
let me know if this is ok for you guys!

### DIFF
--- a/pem/lib/pem/manager.rb
+++ b/pem/lib/pem/manager.rb
@@ -22,12 +22,37 @@ module PEM
             else
               UI.success "You already have a push certificate, which is active for more than 30 more days. No need to create a new one"
               UI.success "If you still want to create a new one, use the --force option when running PEM."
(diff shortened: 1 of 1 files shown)


asfalcone:
@hjanuschka Is this different than what @KrauseFx was mentioning in this issue? https://github.com/fastlane/fastlane/issues/4309#issuecomment-221153594

hjanuschka:
we'll just try it - the p12 downloads and seems to be valid.
i ll setupa sample app and verify it

KrauseFx:
As far as I know there is no way to download an existing private key file from the Apple Developer Portal

hjanuschka:
@KrauseFx  ok tested it - it seems to be a valid p12 cert but it doesnt work so i guess i ll close this PR now

PR #7075, opened 2016-11-18
[fastlane_core] move keychain_path to core *1/3*

hjanuschka:
This series of PR tries to solve following issue:
   * https://github.com/fastlane/fastlane/pull/7048
   
# This is PR 1/3
   
as described the path is to move `keychain_path` to `fastlane_core`:

  * PR 1/3 -> move `keychain_path` to `fastlane_core` #7075 
  * PR 2/3 -> refactor `match` to use fastlane_core `keychain_path` #7076 
  * PR 3/3 -> refactor `unlock_keychain` to use fastlane_core `keychain_path` #7077 
     
     
needs to be merged in the right order

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -151,6 +151,32 @@ module FastlaneCore
       return File.join(self.itms_path, 'bin', 'iTMSTransporter')
     end
 
+    def self.keychain_path(name)
+      # Existing code expects that a keychain name will be expanded into a default path to Libary/Keychains
+      # in the user's home directory. However, this will not allow the user to pass an absolute path
+      # for the keychain value
+      #
+      # So, if the passed value can't be resolved as a file in Library/Keychains, just use it as-is
+      # as the keychain path.
+      #
+      # We need to expand each path because File.exist? won't handle directories including ~ properly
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -151,6 +151,32 @@ def self.transporter_path
...
+      # So, if the passed value can't be resolved as a file in Library/Keychains, just use it as-is
+      # as the keychain path.
+      #
+      # We need to expand each path because File.exist? won't handle directories including ~ properly
+      #
+      # We also try to append `-db` at the end of the file path, as with Sierra the default Keychain name
+      # has changed for some users: https://github.com/fastlane/fastlane/issues/5649
+      #
+
+      keychain_paths = [
```
milch: It might be worthwhile to append `.keychain` to the name here if it doesn't already include it, what do you think?

milch:
Linking back here to the original issue: #7046

fastlane_core/lib/fastlane_core/helper.rb
asfalcone: Let's stick with the original behavior from _match_ for this reactor

asfalcone (approved):

PR #7005, opened 2016-11-15
[gym] support Commandline apps 💪💻

hjanuschka:
as discussed here: https://github.com/fastlane/fastlane/issues/6944

trys to add support for commandline apps / commandline librarys.

the sample project, is for further unit tests.

demo lane:

```ruby
lane :commandline do
	 gym(scheme: "commandline", output_name: "commandline", project: "gym/examples/commandline/commandline.xcodeproj", clean: false)	
end
```

resulting in a binary: `commandline` - in my folder 👍 

sample output:

```
[09:44:47]: You require a gem, please call `fastlane_require 'spaceship'` before to ensure the gem is installed
[09:44:47]: Driving the lane 'commandline' 🚀
[09:44:47]: -----------------
[09:44:47]: --- Step: gym ---
[09:44:47]: -----------------
[09:44:48]: $ xcodebuild -list -project gym/examples/commandline/commandline.xcodeproj
[09:44:50]: $ xcodebuild clean -showBuildSettings -scheme commandline -project gym/examples/commandline/commandline.xcodeproj

+----------------------+------------------------------------------------+
|                        Summary for gym 1.12.0                         |
+----------------------+------------------------------------------------+
| scheme               | commandline                                    |
| output_name          | commandline                                    |
| project              | gym/examples/commandline/commandline.xcodeproj |
| clean                | false                                          |
| destination          | generic/platform=macOS                         |
| output_directory     | .                                              |
| silent               | false                                          |
| use_legacy_build_api | false                                          |
| buildlog_path        | ~/Library/Logs/gym                             |
| xcode_path           | /Applications/Xcode.app                        |
+----------------------+------------------------------------------------+

[09:44:51]: $ set -o pipefail && xcodebuild -scheme commandline -project gym/examples/commandline/commandline.xcodeproj -destination 'generic/platform=macOS' -archivePath /Users/hjanuschka/Library/Developer/Xcode/Archives/2016-11-15/commandline\ 2016-11-15\ 09.44.51.xcarchive archive | tee /Users/hjanuschka/Library/Logs/gym/App-commandline.log | xcpretty
[...]

PR #7090, opened 2016-11-18
[fastlane_core] support Commandline apps 💪💻  1/2

hjanuschka:
split of #7005

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -172,70 +172,60 @@ module FastlaneCore
       return "App" # default value
     end
 
-    def build_for?
-      return :ios_app if ios_app?
-      return :ios_framework if ios_framework?
-      return :ios_library if ios_library?
-      return :mac_app if mac_app?
-      return :mac_library if mac_library?
-      return :mac_framework if mac_framework?
-      return :command_line_tool if command_line_tool?
-    end
-
     def dynamic_library?
-      return true if build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.library.dynamic"
+      (build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.library.dynamic")
     end
 
     def static_library?
-      return true if build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.library.static"
+      (build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.library.static")
     end
 
     def library?
-      return true if static_library? || dynamic_library?
+      (static_library? || dynamic_library?)
     end
 
     def framework?
-      return true if build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.framework"
+      (build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.framework")
     end
 
     def application?
-      return true if build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.application"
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -172,6 +172,14 @@ def app_name
       return "App" # default value
     end
 
+    def command_line_library?
+      return true if build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.library.dynamic"
```
ohwutup: `"com.apple.product-type.library.dynamic"` is this the product type when Xcode is building a dynamic framework also? I'm wondering if we should only worry about supporting `"com.apple.product-type.tool"` below?

hjanuschka:
@ohwutup updated!

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -172,6 +172,72 @@ def app_name
...
+    def ios_framework?
+      return true if framework? && build_settings(key: "PLATFORM_NAME") == "iphoneos"
+    end
+
+    def ios_app?
+      return true if application? && build_settings(key: "PLATFORM_NAME") == "iphoneos"
+    end
+
+    def produces_archive?
+      return true unless framework? || static_library? || dynamic_library?
```
ohwutup: Frameworks and libraries don't create archives?

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -172,6 +172,72 @@ def app_name
       return "App" # default value
     end
 
+    def build_for?
```
ohwutup: Where is this getting used? I couldn't find a usage of this in https://github.com/fastlane/fastlane/pull/7091/files either?

fastlane_core/lib/fastlane_core/project.rb
hjanuschka: nope they don't. they actually produce xcarchive files, but they are empty.

fastlane_core/lib/fastlane_core/project.rb
hjanuschka: true - not used, i initially thought i'd use this as a classifcation, but then it worked without, anyway i thought it might be nice to keep, or should be removed?

fastlane_core/lib/fastlane_core/project.rb
ohwutup: 👍

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -172,6 +172,72 @@ def app_name
...
+      return :ios_framework if ios_framework?
+      return :ios_library if ios_library?
+      return :mac_app if mac_app?
+      return :mac_library if mac_library?
+      return :mac_framework if mac_framework?
+      return :command_line_tool if command_line_tool?
+    end
+
+    def dynamic_library?
+      return true if build_settings(key: "PRODUCT_TYPE") == "com.apple.product-type.library.dynamic"
```
ohwutup: One more thing for all of these `?` methods... would you mind just returning the boolean expression rather than `true if boolean-expression`?

ohayon (dismissed):

milch:
@ohwutup Merging this so we can merge #7207 - since as far as I can see, your reviews have been addressed.

milch (approved):

PR #6563, opened 2016-10-15
add SharetheMeal Plugin

hjanuschka:

PR #7093, opened 2016-11-18
[docs] regenerate AvailaiblePlugins

hjanuschka:
as told in #6563

### DIFF
--- a/fastlane/docs/AvailablePlugins.md
+++ b/fastlane/docs/AvailablePlugins.md
@@ -2,13 +2,13 @@
 
 To get an up to date list of all available plugins run
 
-```
+```no-highlight
 fastlane search_plugins
 ```
 
 To search for a specific plugin
 
-```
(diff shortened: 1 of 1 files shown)


milch:
That AvailablePlugins.md is woefully out of date, I think we just haven't kept it updated because it's automatically generated on docs.fastlane.tools anyway. 

Maybe instead of manually fixing that file, could you regenerate it using the rake task? https://github.com/fastlane/fastlane/blob/master/fastlane/Rakefile

hjanuschka:
@milch  done

milch (approved):

milch:
Great, thanks! 🚀

KrauseFx:
We actually kind of moved this file to [docs.fastlane.tools](https://docs.fastlane.tools). Should we also deprecate the docs and link to the new page? 👍

Issue #7094, opened 2016-11-18
Supply should support a dry run option


tmtrademarked:
I recently had an issue where I managed to break my master build on CI due to a failure to upload to Google Play. (I had stupidly left an APK as debuggable - d'oh!) This wasn't caught by my tests, since non-master branches don't run my deploy step.

The Google Play Publisher API supports a "validate edit" feature. This can be used to verify that the edit is valid without actually changing the data. This would have been super helpful for me.

From what I understand, this should be pretty trivial to implement. Thoughts?

tmtrademarked:
I'm a terrible person, and decided to just create a sample pull request to see what it would take. But my ruby is very weak - so please feel free to ignore this if it's not a good way to fix this problem!

milch:
> I'm a terrible person

Aww, don't be too hard on yourself. We love all our contributors equally! ❤️ 

> The Google Play Publisher API supports a "validate edit" feature. This can be used to verify that the edit is valid without actually changing the data. This would have been super helpful for me.

That does indeed sound like a very good idea. I'm gonna take a look at your PR now 🚀

tmtrademarked:
Quick question here - what's the process look like for getting this into a release? No urgency, just curious about the flow of releases for the various tools.

milch:
@tmtrademarked 

Unless there is an emergency bugfix or a large feature merged, we usually wait until there's a small number of PRs merged before we release a new version of one of the tools.

PR #7095, opened 2016-11-18
Adding support for a dry-run option

tmtrademarked:

- [ X] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [ X] Run `bundle exec rubocop -a` to ensure the code style is valid
- [ X] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ X] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Proposed fix for https://github.com/fastlane/fastlane/issues/7094

This pull request adds support for a validate_only option. This will only validate the commit with Google Play, rather than actually uploading the data.

This is useful for testing your changes before actually commiting them. For example, in a CI environment, this can be used to make sure that all changes would successfully upload to Play before merging to master.

### DIFF
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -125,7 +125,7 @@ module Supply
         FastlaneCore::ConfigItem.new(key: :validate_only,
                                      env_name: "SUPPLY_VALIDATE_ONLY",
                                      optional: true,
-                                     description: "If true, changes will only be validated with Google Play rather than actually uploaded",
+                                     description: "Indicate that changes will only be validated with Google Play rather than actually uploaded",
                                      is_string: false,
                                      default_value: false)
 
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -125,7 +125,7 @@ module Supply
         FastlaneCore::ConfigItem.new(key: :validate_only,
                                      env_name: "SUPPLY_VALIDATE_ONLY",
                                      optional: true,
-                                     description: "Indicate that changes will only be validated with Google Play rather than actually uploaded",
+                                     description: "Indicate that changes will only be validated with Google Play rather than actually published",
                                      is_string: false,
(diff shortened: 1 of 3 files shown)


supply/lib/supply/options.rb
```diff
@@ -121,7 +121,13 @@ def self.available_options
                                      verify_block: proc do |value|
                                        available = valid_tracks
                                        UI.user_error! "Invalid value '#{value}', must be #{available.join(', ')}" unless available.include? value
-                                     end)
+                                     end),
+        FastlaneCore::ConfigItem.new(key: :validate_only,
+                                     env_name: "SUPPLY_VALIDATE_ONLY",
+                                     optional: true,
+                                     description: "If true, changes will only be validated with Google Play rather than actually uploaded",
```
milch: `Indicate that changes will only be validated with Google Play rather than actually published` would sound better. Also I think that your data is still uploaded - just the publishing step is skipped?

milch (changes requested):

milch:
Hey @tmtrademarked, congrats on your first PR in the fastlane repo, and thanks a lot for your contribution! 🚀 

Your PR is looking pretty good, we just need to change the wording around a little bit.

supply/lib/supply/options.rb
tmtrademarked: Thanks for the review! Updated the comment per your suggestion.

As I understand the validate functionality, it uploads the data but doesn't change anything. So if you don't call the commit function, nothing about your listing will change. (See the documentation here: https://developers.google.com/android-publisher/api-ref/edits/validate) So I think with the new wording, we're reasonably accurate.

milch (approved):

PR #7103, opened 2016-11-20
Add option to change provisioning style of targets between automatic …

klundberg:
…and manual

I need to be able to change provisioning styles in fastlane. We have automatic provisioning set during development so that developers can build to their device with minimal effort, but during a beta or appstore build we need to tightly control the signing process. We currently don't use match since it doesn't support enterprise builds according to the docs, so we set up signing with cert and sigh and update_project_provisioning, and we need to be able to change provisioning style here too to make sure everything works as expected.

I'm open to the idea of moving this option to a different action, perhaps update_development_team or something that is agreed on as a better alternative, but putting it here seems like a reasonable place to me.

If there's any other considerations to make here, I'm open to them.

Thanks!
 
---

### DIFF
--- a/fastlane/lib/fastlane/actions/update_project_provisioning.rb
+++ b/fastlane/lib/fastlane/actions/update_project_provisioning.rb
@@ -39,6 +39,7 @@ module Fastlane
 
         target_filter = params[:target_filter] || params[:build_configuration_filter]
         configuration = params[:build_configuration]
+        provisioning_style = params[:provisioning_style]
 
         # manipulate project file
         UI.success("Going to update project '#{folder}' with UUID")
@@ -52,6 +53,8 @@ module Fastlane
             UI.important("Skipping target #{target.product_name} as it doesn't match the filter '#{target_filter}'")
             next
           end
+
+          project.root_object.attributes["TargetAttributes"][target.uuid]["ProvisioningStyle"] = provisioning_style unless provisioning_style.nil?
 
           target.build_configuration_list.build_configurations.each do |build_configuration|
             config_name = build_configuration.name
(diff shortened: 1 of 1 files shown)


hjanuschka:
i once did this as a plugin: https://github.com/hjanuschka/fastlane-plugin-update_project_codesigning

```ruby
 update_project_codesigning(path: "demo-project/demo/demo.xcodeproj", use_automatic_signing: true)
```

ohwutup:
Hey @klundberg and @hjanuschka! 

Thanks for the changes and feedback! One thing I am a bit curious about: @klundberg - how does this work without setting a team as well? In my experience with Automatic code signing in Xcode, a team needs to be set and its possible to have more than one available in Preferences>Accounts.

Same question for you, @hjanuschka, I noticed in your plugin you have a param for team ID, but in your example above, you don't mention that. Isn't it necessary?

Thanks!

klundberg:
@ohwutup Currently we use `update_developer_team` in a different part of our fastfile, so we're setting that at that point. If you feel that this option is better served there, then that makes sense to me, but you still need to set provisioning info if you set it to manual at that point (which `update_developer_team` does not do). It seems to be different tradeoffs either way.

ohwutup:
Hey @klundberg - ah, that makes sense that you are using that action first! Have you given @hjanuschka's plugin a try? It seems like it could do the trick for you.

Would love to hear how that goes either way in order to hopefully avoid duplicating something that exists as a plugin. 😄

klundberg:
I haven’t yet. I’ll give it a shot and see if it will work for us, and update here when I can. Thanks!
--
Kevin Lundberg

hjanuschka:
@klundberg let me know if have issues 👍

ohwutup:
Thanks @klundberg! I am going to close this PR for now and we can definitely revisit it if you are having issues with @hjanuschka's plugin.

🚀 🐙

PR #7112, opened 2016-11-21
Fix spaceship retry tests

mfurtak:
It seems like some exception handling behavior has changed in faraday. Possibly in the 0.10.0 release on Nov 11?

A thrown `Errno::EPIPE` is now always wrapped in a `Faraday::Error::ConnectionFailed`


ohayon (approved):
👍

Issue #7086, opened 2016-11-18
[Feature Request] Add support for build-for-testing and test-without-building


samskiter:
### Issue Description

XCode 8 added build-for-testing and test-without-building commands to xcodebuild - we could potentially use these to greatly speed up our CI system. It would be great if fastlane had a native way to support this.

hjanuschka:
how about using the `xcodebuild` action?

```ruby
xcodebuild(
  clean: true,
  workspace: 'FiveXFive.xcworkspace',
  scheme: 'UITests',
  destination: 'generic/platform=iOS',
  configuration: 'Development',
  xcargs: " build-for-testing"
)
```

hjanuschka:
see -> https://github.com/fastlane/fastlane/pull/7130

samskiter:
@hjanuschka Looks good. Build for testing available too?

hjanuschka:
@samskiter with this pr scan supports the use of a already build test app - just running it without building it.

what exactly do you wan't to do?

as described here: https://github.com/fastlane/fastlane/issues/7086#issuecomment-261608999  you already can - "just build" the test binary?


for snapshot  i also did a PR: https://github.com/fastlane/fastlane/pull/7131 - to only build once and run snapshot on different devices, using build-for-testing, test-without-building.

samskiter:
I would like to build the artifacts, distribute them to different build machines and run them there.

The idea of scan is to be easy to use right?
"scan uses the latest APIs and tools to make running tests plain simple and offer a great integration into your existing workflow, like fastlane or Jenkins."

I'd like to use the same simple api that scan provides rather than trying to use xcodebuild if possible

hjanuschka:
@samskiter ok got you!
updated my pr.

now following thing should work:

```ruby
  scan(derived_data_path: "my_folder", build_for_testing: true)
# copy the whole folder 'my_folder' to another machine

# run this on other machine
  scan(derived_data_path: "my_folder", test_without_building: true, skip_build: true)
```

samskiter:
@hjanuschka looks great, thank-you!

hjanuschka:
@samskiter please +1 the PR if you mind.

PR #7130, opened 2016-11-22
support for test-without-building, build-for-testing

hjanuschka:
this one adds support for `test-without-building` using derived data path
  
  ```ruby
  #Build For Testing
  scan(derived_data_path: "my_folder", build_for_testing: true)
  # run tests using derived data from prev. build
  scan(derived_data_path: "my_folder", test_without_building: true)
  # or run it from an existing xctestrun package
  scan(xctestrun: "/path/to/mytests.xctestrun")
  ```
  
related issues:
  * https://github.com/fastlane/fastlane/issues/7086


samskiter:
+1

scan/lib/scan/test_command_generator.rb
```diff
@@ -31,8 +31,9 @@ def options
 
         options = []
         options += project_path_array
-        options << "-sdk '#{config[:sdk]}'" if config[:sdk]
-        options << destination # generated in `detect_values`
+        options << "-sdk '#{config[:sdk]}'" if config[:sdk] && !config[:build_for_testing]
+        options << destination unless config[:build_for_testing] # generated in `detect_values`
+        options << "  -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO" if config[:build_for_testing]
```
KrauseFx: Why do we potentially set sdk twice?

KrauseFx:
Thanks for the PR. This would also require some mentioning in the documentation

scan/lib/scan/test_command_generator.rb
hjanuschka: if we build for testing we need `iphonesimulator` as a sdk - in the `config[:sdk]` user can supply one, for testing, but as we want to build only once, its required to have `iphonesimulator` as sdk.

macduy:
+1

samskiter:
@hjanuschka @KrauseFx what documentation is needed and where?

samskiter:
Really keen to start using this. Can I help?

hjanuschka:
@KrauseFx  can you help me here, where should i add some documentation for it?

scan/lib/scan/test_command_generator.rb
```diff
@@ -49,7 +50,9 @@ def actions
         actions = []
         actions << :clean if config[:clean]
         actions << :build unless config[:skip_build]
-        actions << :test
+        actions << "build-for-testing" if config[:build_for_testing]
+        actions << :test unless config[:test_without_building] || config[:build_for_testing]
```
lyricsboy: The way these actions are being managed is a little cumbersome. I'd be inclined to design it such that `scan` supports a new `action` option, which defaults to "test". Other possible values would be "build-for-testing" or "test-without-building" to match the actions passed to xcodebuild.

It seems wise to move closer to the 🍎  rather than have scan's UI diverge further from what's happening under the hood.

hjanuschka: @lyricsboy you are right this looks like a good thing. in fact we still need some more detection of this, as with build-for-testing/test-without-building we need to switch parameters here: https://github.com/fastlane/fastlane/pull/7130/files#diff-ef49192ad6b55e081d8efc1ddabd1ee8R36

i ll happily adopt to proposed solution - but i think there is still no core-team decision on how to proceed with the whole , `build-for-testing` thing. @KrauseFx  any idea?

@mfurtak probably 8.2 brings a speed improvement - should we retest it?

mfurtak: @hjanuschka Retesting sounds like a good idea! I won't be able to do that this week, but I'm interested to hear what you find :+1:

Jules2010:
I'm willing to help in someway @hjanuschka , but I've no clue how to checkout out the pr and build it.

What needs to happen for this to be approved @mfurtak , there seems to be some mention of the documentation and where changes should be made @KrauseFx

hjanuschka:
You can run it as described below:

# Option 1
get this script into your project folder, 
https://gist.github.com/hjanuschka/3040fc6f371f46acff72a5e67f649b27

This requires you to push changes first to github and then call `try_pr.sh` again, 
but its an easy way to try out PR's
```bash
cd /path/to/your/project/
# this only has to be done once.
curl -o try_pr.sh https://gist.githubusercontent.com/hjanuschka/3040fc6f371f46acff72a5e67f649b27/raw/9600a34188949061a96f77c82505efd99ad6f5e4/try_pr.sh
chmod a+x  try_pr.sh


# then call it with (which in this case should show the help screen of your supply):
./try_pr.sh hjanuschka  testing_scan fastlane your_lane
```


# Option 2
Totally local.:

```bash
cd /path/to/your/fastlane-clone/
# ensure your branch is checked out
export FASTLANE_LOCAL_DEV=1
bundle update # only once
bundle install #only once
bundle exec fastlane
```


> just to be sure, unsupported, unofficial, use at your own risk! - and review the shell script before you run it.

Jules2010:
Thanks I will take a look.

What does the other pull request do, it looks like the same as this?

hjanuschka:
@Jules2010 one PR adds the test-without-building/build-for-testing to scan and the other for snapshot

samskiter:
Could someone please clarify what needs to happen for this to go in. @mfurtak @KrauseFx  ?

ohwutup:
@samskiter - I think that the issue is that we are not seeing any speed improvements to testing times by doing this.

Jules2010:
@ohwutup over multiple devices ?

ohwutup:
@Jules2010, so I'm not so familiar with the new feature in `xcodebuild` tbh. But I don't believe that should matter given the fact that running it on the same machine in multiple simulators showed no speed improvements at all. In fact, the runs were slightly slower. See this thread for reference: https://github.com/fastlane/fastlane/pull/7131#issuecomment-262290506

Jules2010:
@ohwutup I'm confused I thought the idea was that it built once then used that build without the other simulators and this lack of subsequent build time was the saving. Maybe I've misunderstood.

ohwutup:
@Jules2010 we were equally confused 🤔 haha. We thought the same thing as you but as we put it through testing we didn't notice any improvements in speed.

fsaragoca:
I would love to see `build-for-testing` & `test-without-building` integrated with scan – I'm currently using `xcodebuild` but I miss some scan features like jUnit reports. Although the discussion here is related to snapshot, I can see many other use cases like testing different OS versions, locales and screen sizes.

This is my current setup:

screen shot 2016-12-20 at 7 29 07 pm

I'm running `xcodebuild` with `build-for-testing` in build step, them uploading results to a cloud storage. In test steps, I download results from previous step and run `xcodebuild` with `test-without-building`.

Please let me know if there is anything I can do to help.

ohwutup:
@fsaragoca we didn't notice any time savings when running tests on simulators, in fact it got a bit longer. Are you noticing speed enhancements? https://github.com/fastlane/fastlane/pull/7131#issuecomment-262290506

samskiter:
@fsaragoca Actually I think this PR is about scan. 

The other (https://github.com/fastlane/fastlane/pull/7131) is about snapshot

Are others here getting the two mixed up? 

I don't know why you would see speedup for snapshot. build-for-testing and test-without-building are about creating all the compiled binaries and object files (and a description of the tests to be run) and then being able to distribute those onto multiple machines to actually run.

I've previously done some tests with xcodebuild. On related compiles for testing, I remember seeing a speedup. It's been a while, but I will try to get some numbers tomorrow.

Regardless of being able to measure speedups or not - this is an interface onto xcodebuild that is currently missing from scan, so why hold up exposing it?

fsaragoca:
@ohwutup yes I am. Here is a comparison between `xcodebuild` and `scan`:

#### Test A:
-  Running xcodebuild & test-without-building
- Derived data contains results from xcodebuild & build-without-testing

```
+------+----------------------------------------+-------------+
|                      fastlane summary                       |
+------+----------------------------------------+-------------+
| Step | Action                                 | Time (in s) |
+------+----------------------------------------+-------------+
| 1    | Verifying required fastlane version    | 0           |
| 2    | default_platform                       | 0           |
| 3    | setup_environment_variables            | 0           |
| 4    | Switch to ios prepare_for_testing lane | 0           |
| 5    | check_build_cache_workspace            | 0           |
| 6    | clear_derived_data                     | 0           |
| 7    | unarchive_derived_data                 | 1           |
| 8    | xcodebuild                             | 21          |
+------+----------------------------------------+-------------+
```

#### Test B:
- Running scan
- Derived data contains results from xcodebuild & build-without-testing

```
+------+----------------------------------------+-------------+
|                      fastlane summary                       |
+------+----------------------------------------+-------------+
| Step | Action                                 | Time (in s) |
+------+----------------------------------------+-------------+
| 1    | Verifying required fastlane version    | 0           |
| 2    | default_platform                       | 0           |
| 3    | setup_environment_variables            | 0           |
| 4    | Switch to ios prepare_for_testing lane | 0           |
| 5    | check_build_cache_workspace            | 0           |
| 6    | clear_derived_data                     | 0           |
| 7    | unarchive_derived_data                 | 1           |
| 8    | scan                                   | 281         |
+------+----------------------------------------+-------------+
```

#### Test C:
- Running scan
- Derived data contains results from previous scan build

```
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 3    | setup_environment_variables         | 0           |
| 4    | scan                                | 37          |
+------+-------------------------------------+-------------+
```

#### Summary for `scan`:

```
+-----------------------+----------------------------------+
|                 Summary for scan 0.14.2                  |
+-----------------------+----------------------------------+
| scheme                | Project Test                     |
| clean                 | false                            |
| skip_build            | true                             |
| sdk                   | iphonesimulator                  |
| derived_data_path     | fastlane/build_cache/DerivedData |
| workspace             | ./Project.xcworkspace            |
| code_coverage         | false                            |
| address_sanitizer     | false                            |
| output_directory      | ./fastlane/test_output           |
| output_types          | html,junit                       |
| buildlog_path         | ~/Library/Logs/scan              |
| open_report           | false                            |
| skip_slack            | false                            |
| slack_only_on_failure | false                            |
| use_clang_report_name | false                            |
| fail_build            | true                             |
| xcode_path            | /Applications/Xcode.app          |
+-----------------------+----------------------------------+
```

#### xcodebuild parameters:

```
xcodebuild(
  scheme: 'Project Test',
  destination: 'platform=iOS Simulator,name=iPhone 6,OS=10.2',
  derivedDataPath: ENV['BUILD_CACHE_DERIVED_DATA_PATH'],
  xcargs: '-sdk iphonesimulator test-without-building'
)
```

I also noticed that `scan` derived data is not reliable, sometimes nothing changes and it recompiles everything. `xcodebuild` + `build-for-testing` derived data can be cached and shared between different CI machines.

fsaragoca:
@samskiter yes, I agree with you

scan/lib/scan/test_command_generator.rb
samskiter: I think we still need a destination here right? the idea is everything should be the same as if you called `test`

See http://devstreaming.apple.com/videos/wwdc/2016/409jh83sf1h8dqrt00q/409/409_advanced_testing_and_continuous_integration.pdf

ohwutup:
Thanks for all that info @fsaragoca! And you're definitely right @samskiter that we are not currently supporting the entire interface into `xcodebuild`. That being said, our goal with _scan_ or any of our tools for that matter is not to be a direct interface into underlying tools, rather, to provide tools that meet the needs of common app development and deployment tasks and we need to be careful introducing new functionality and options so as to make sure if things change down the line we will not break people's configurations. Now, with *that* being said, I do think that it is worthwhile to support the `test-without-building` and `build-for-testing` options. It would be awesome if @hjanuschka and company could address the open feedback regarding passing the destination option that @samskiter mentioned and retesting in Xcode 8.2 and then we can work on getting it merged in.

samskiter:
Thanks for the reply. Just in the process of getting some testing metrics
together with the 'destination' set, to show the advantage you get
compiling for multiple targets (I.e. reusing .o files for the common
files). We have a small cluster of machines that a Jenkins instance can
control and we are hoping to see a big speedup by centralising 'compile' on
one fast machine and distributing 'run' to several slower machines.

To be clear, the speedup from this feature will only be seen when running
with multiple build agents.

scan/lib/scan/test_command_generator.rb
hjanuschka: @samskiter correct me if i am wrong but during creating this PR - i discovered that once you set destination you get troubles building for different architectures and ios versions.


so i it ended up to switch from `-destination`   to `-sdk iphonesimulator ONLY_ACTIVE_ARCH=NO`  - found the note somewhere (i think in aws device farm docs)

samskiter:
Yes, but that's not the point right - you should test with the same device
you build for right?

ohwutup:
@samskiter, not sure I follow. Would you mind elaborating a bit? 😄

samskiter:
Sorry, github e-mail reply fail.

hjanuschka:
@samskiter  i think that was meant to be a codeline reply :)

you can specify multiple devices - like `['iPhone 6', 'iPad Air']`  this would potentially be an ARCH problem - therefore build for iphonesimulator with all arch's. or am i totally wrong - at least what i saw during testing.

samskiter:
@hjanuschka So are you saying when you tried to specify multiple device in the build-for-test phase, it then didn't work with those same devices in the test-without-building phase?

hjanuschka:
@samskiter that what i ran into! - my first try was just to swap out "build"  for "build-for-testing"  - and it just didn't work. and especially if you wan't to build-for-testing, and later on copy those files to a remote machine, you can specify different devices/versions - as it is not bound to special arch.

samskiter:
Aren't you supposed to swap out "test" for "build-for-testing" ? Not sure if theres a difference in what you'd specify.

scan/lib/scan/test_command_generator.rb
nikolaykasyanov: @samskiter isn't the idea of building without testing is to actually be able to run build product on different simulators?

samskiter:
@hjanuschka I've looked into the multiple devices thing.

Firstly, AFAICT scan doesn't really support testing from multiple architectures - it only really supports simulators through the `devices` - you can pass `destination` to get it running on physical devices, but that's for 'pro' users AND you can't pass multiple `destination` arguments to scan. So I actually don't see how anyone would currently achieve creating a scan run that will be compiling for multiple architectures. My code signing is bust right now so I've been a bit limited in my testing.


I've also tried using xcodebuild directly with multiple devices (one simulator and real device) and found I could build-for-testing and then build-without-testing without any issues. For reference, using XCode 8.2 and the following commands:

```
set -o pipefail && env NSUnbufferedIO=YES xcodebuild -workspace MyProject.xcworkspace -scheme MyProject -destination 'platform=iOS Simulator,id=<UDID>' -destination 'platform=iOS,id=<UDID>' -derivedDataPath 'derivedData' build-for-testing | tee '/Users/samskiter/Library/Logs/scan/MyProject-MyProject.log' | xcpretty
```
```
set -o pipefail && env NSUnbufferedIO=YES xcodebuild -workspace MyProject.xcworkspace -scheme MyProject -destination 'platform=iOS Simulator,id=<UDID>' -destination 'platform=iOS,id=<UDID>' -derivedDataPath 'derivedData' test-without-building | tee '/Users/samskiter/Library/Logs/scan/MyProject-MyProject.log' | xcpretty
```

If you do have trouble with multiple architectures, that seems like a bug with xcodebuild. Forcing the multiple architecture compile will increase compile times, which is the exact opposite of what this is supposed to achieve. You should either specify the device up-front in the build-for-testing phase, or, if you really need to be able to compile for everything, I think it would be a better solution so add an option to scan to manually turn on multi-arch compile.

From my reading of the presentation I posted earlier, the point of build-for-testing isn't to allow you to be able to test on any device you might like sometime in the future, you should know ahead of time, what you will be running on and can build accordingly, then later run the test with *exactly* the same arguments.

Re: this MR - I've tested quite a bit on my fork (which just removes the ACTIVE_ARCH stuff) and found it works great :)

scan/lib/scan/test_command_generator.rb
samskiter: @nikolaykasyanov Different sims wouldn't require different ARCH would they? it's all x86_64 right?

This option will slow down compile times, rather than speed them up. Which sort of defeats the point.

If you want multiple ARCHS, you would specify them in the `device` argument to scan.

samskiter: Just to clarify, there's the use case of "I want to run the same testing target on loads of different devices" that I think you are alluding to. But there's also the use case of "I want to do all my compiling for my multiple testing targets on one machine and reap the speedup reward of not recompiling shared dependencies, but then actually run the tests on different machines". I'm interested in the latter.

hjanuschka:
@samskiter  thank you for your detailed rexplanation.
is it ok for you if i just cherry-pick your 2 commits into this branch, or should we continue in a new PR directly from your branch?

samskiter:
@hjanuschka Yea, fine, really they are just reverting your change, but feel free to adjust this PR as needed

samskiter (approved):

scan/lib/scan/test_command_generator.rb
nikolaykasyanov: @samskiter iPhone 4s or 5 simulators are i386, for example.

nikolaykasyanov:
@samskiter 

>I think it would be a better solution so add an option to scan to manually turn on multi-arch compile.

Sounds good! 👍

scan/lib/scan/test_command_generator.rb
```diff
@@ -49,7 +49,9 @@ def actions
...
```
KrauseFx: unless in combination with the `or` is pretty confusing, and caused a major issue in gym already. Let's rewrite that line so it's easier to read and more clear

scan/lib/scan/test_command_generator.rb
```diff
@@ -49,7 +49,9 @@ def actions
...
+        actions << "test-without-building" if config[:test_without_building]
```
KrauseFx: Okay, there is not many alternatives. Maybe just splitting it into more lines makes it easier 👍 
```
if config[:build_for_testing]
  actions << "build-for-testing"
end

unless config[:test_without_building] || config[:build_for_testing])
  actions << :test
end

if config[:test_without_building]
  actions << "test-without-building"
end
```

hjanuschka: you are right - i splitted it, and well - i learned my lesson - will use unless not to often anymore 🙏

hjanuschka (commented):
feedback addressed

hjanuschka:
@mfurtak i simplified the build mode detection and made it more exclusive.

also added a few specs to prevent regressions.

scan/lib/scan/test_command_generator.rb
```diff
@@ -48,9 +48,20 @@ def actions
...
+          config[:skip_build] = true
+        end
+
+        if config[:test_without_building]
+          building_mode = "test-without-building"
+          config[:skip_build] = true
+        end
+
+        actions << :build unless config[:skip_build]
+        actions << building_mode
```
ohwutup: Hey @hjanuschka, I think ideally we would be able to avoid manipulating the `config` object. Would it be okay to re-write to be a bit more like so:

```ruby
if config[:build_for_testing]
  actions << "build-for-testing"
elsif config[:test_without_building]
  actions << "test-without-building"
else
  actions << :build unless config[:skip_build]
  actions << :test
end
```

ohayon (dismissed):

hjanuschka:
@ohwutup feedback addressed, i tried to avoid `elsif` - but its better not to change the config params.

ileitch:
@hjanuschka This PR needs to support the `-xctestrun` option. This option is especially helpful when the build is executed on one machine, packaged up and then transferred to another machine for the testing phase.

You can look at my PR (which was closed in favor of this one) as an example: https://github.com/fastlane/fastlane/pull/7728

hjanuschka:
@ileitch  ok just to check if i got it right.

`build_for_testing` is ok as it is in this PR.
and `test_without_building` should by default use the `derived_data` approach.

and i should add a option: `xctestrun` - where you can set a path to a xctestrun package, and if this is supplied it should be added to the test command?


i am happy to add this - if you wan't you also could do a PR against the `testing_scan` branch  on my fork adding this functionallity.

ileitch:
> build_for_testing is ok as it is in this PR.
Correct.

> and test_without_building should by default use the derived_data approach.
Correct.

> and i should add a option: xctestrun - where you can set a path to a xctestrun package, and if this is supplied it should be added to the test command?
Correct.

`-xctestrun` implies `test_without_building`, so I'm not sure how you want to handle that from a Fastlane API perspective. Require both options be explicitly provided, or enable `test_without_building` if `xctestrun` is present?

hjanuschka:
@ileitch  see 41cc468 i added a option for that `:xctestrun`(`-X`) and a test for it, can you try this PR?

edited initial PR body and show how to use it.

scan/lib/scan/options.rb
```diff
@@ -135,6 +135,13 @@ def self.available_options
                                      conflicting_options: [:test_without_building],
                                      is_string: false,
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :xctestrun,
+                                     short_option: "-X",
+                                     env_name: "SCAN_XCTESTRUN",
+                                     description: "Run tests using the provided .xctest binary",
```
hjanuschka: thx - fixed

samskiter:
Can this go yet?

KrauseFx (approved):
Looks great, thanks for working on this @hjanuschka 👍

hjanuschka:
many thx on all involved guys, this is a great example of team work!!!

lyndsey-ferguson:
Thanks @hjanuschka !

Issue #6992, opened 2016-11-14
Build once ?


Jules2010:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

When using screenshot it builds for each device, surely it should only need to be built once and copied to the next simulator ?

##### Complete output when running fastlane, including the stack trace and command used

N/A

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                                          |
| ------------------- | -------------------------------------------------------------- |
| OS                  | 10.12.1                                                        |
| Ruby                | 2.0.0                                                          |
| Bundler?            | false                                                          |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version       | 8.1                                                            |
| Git                 | git version 2.3.0                                              |
| Installation Source | /usr/local/bin/fastlane                                        |
| Host                | Mac OS X 10.12.1 (16B2555)                                     |
| Ruby Lib Dir        | /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib |
| OpenSSL Version     | OpenSSL 0.9.8zc 19 Mar 2015                                    |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_GB.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

**No Fastfile found**

**No Appfile found**

### fastlane gems

| Gem                 | Version | Update-Status |
| ------------------- | ------- | ------------- |
| spaceship           | 0.37.0  | ✅ Up-To-Date  |
| credentials_manager | 0.16.2  | ✅ Up-To-Date  |
| fastlane_core       | 0.55.0  | ✅ Up-To-Date  |
| deliver             | 1.15.0  | ✅ Up-To-Date  |
| snapshot            | 1.16.3  | ✅ Up-To-Date  |
| frameit             | 3.0.0   | ✅ Up-To-Date  |
| pem                 | 1.4.0   | ✅ Up-To-Date  |
| cert                | 1.4.4   | ✅ Up-To-Date  |
| sigh                | 1.11.2  | ✅ Up-To-Date  |
| produce             | 1.3.0   | ✅ Up-To-Date  |
| gym                 | 1.12.0  | ✅ Up-To-Date  |
[...]

PR #7131, opened 2016-11-22
[snapshot] support for built-for-testing, test-without-building

hjanuschka:
This one refactors snapshot to use `build-for-testing` and `test-without-building`
to massive improve snapshot runtime.

run with (in fastlane clone):

`bundle exec snapshot --project snapshot/example/Example.xcodeproj -s Example`

```
+----------------------+---------+
|        snapshot results        |
+----------------------+---------+
| Device               | en-US   |
+----------------------+---------+
| iPhone 5             |  💚      |
| iPhone 6             |  💚      |
| iPhone 6 Plus        |  💚      |
| iPhone 7             |  💚      |
| iPhone 7 Plus        |  💚      |
| iPhone SE            |  💚      |
| iPad Air             |  💚      |
| iPad Pro (9.7-inch)  |  💚      |
| iPad Pro (12.9-inch) |  💚      |
+----------------------+---------+
```

Time with all simulators (all sims, atleast one time booted, on my machine):  **8m25.228s**

without this PR:  **9m47.999s**

(depending on project-size i think this will have a huge performance boost)

related issues:
  * https://github.com/fastlane/fastlane/issues/6992

### DIFF
--- a/snapshot/lib/snapshot/runner.rb
+++ b/snapshot/lib/snapshot/runner.rb
@@ -32,7 +32,7 @@ module Snapshot
 
       UI.success "Building and running project - this might take some time..."
 
-      build_for_testing(Snapshot.config[:devices].first)
+      build_for_testing(Snapshot.config[:devices].first) if FastlaneCore::Feature.enabled?('FASTLANE_SNAPSHOT_BUILD_FOR_TESTING')
 
       self.number_of_retries_due_to_failing_simulator = 0
       self.collected_errors = []
--- a/snapshot/lib/snapshot/runner.rb
+++ b/snapshot/lib/snapshot/runner.rb
@@ -32,6 +32,8 @@ module Snapshot
 
       UI.success "Building and running project - this might take some time..."
 
+      build_for_testing(Snapshot.config[:devices].first)
+
       self.number_of_retries_due_to_failing_simulator = 0
       self.collected_errors = []
       results = {} # collect all the results for a nice table
@@ -118,6 +120,14 @@ module Snapshot
       puts ""
     end
 
+    def build_for_testing(device_type = nil)
+      build_command = TestCommandGenerator.generate(device_type: device_type, build_type: "build-for-testing")
+      # build-for-testing
+      FastlaneCore::CommandExecutor.execute(command: build_command,
+                                          print_all: true,
+                                      print_command: true)
+    end
+
     # Returns true if it succeded
     def launch(language, locale, device_type, launch_arguments)
       screenshots_path = TestCommandGenerator.derived_data_path
@@ -153,8 +163,7 @@ module Snapshot
       add_media(device_type, :video, Snapshot.config[:add_videos]) if Snapshot.config[:add_videos]
 
       open_simulator_for_device(device_type)
-
-      command = TestCommandGenerator.generate(device_type: device_type)
+      test_command = TestCommandGenerator.generate(device_type: device_type, build_type: "test-without-building")
 
       if locale
         UI.header("#{device_type} - #{language} (#{locale})")
@@ -171,7 +180,8 @@ module Snapshot
         }
       ]
 
-      FastlaneCore::CommandExecutor.execute(command: command,
+      # test-without-building
+      FastlaneCore::CommandExecutor.execute(command: test_command,
                                           print_all: true,
                                       print_command: true,
                                              prefix: prefix_hash,
--- a/snapshot/lib/snapshot.rb
+++ b/snapshot/lib/snapshot.rb
@@ -14,6 +14,7 @@ require 'snapshot/fixes/simulator_zoom_fix'
 require 'snapshot/fixes/hardware_keyboard_fix'
 
 require 'fastlane_core'
+require 'snapshot/features'
 
 require 'open3'
 
--- a/snapshot/lib/snapshot/features.rb
+++ b/snapshot/lib/snapshot/features.rb
@@ -1,0 +1,2 @@
+FastlaneCore::Feature.register(env_var: 'FASTLANE_SNAPSHOT_BUILD_FOR_TESTING',
+                               description: 'Use Build-For-Testing instead of build on every device used in snapshot')
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -2,13 +2,14 @@ module Snapshot
   # Responsible for building the fully working xcodebuild command
   class TestCommandGenerator
     class << self
-      def generate(device_type: nil)
+      def generate(device_type: nil, build_type: "build-for-testing")
         parts = prefix
         parts << "xcodebuild"
         parts += options
-        parts += destination(device_type)
+        parts += destination(device_type) if build_type != "build-for-testing"
+        parts += ['  -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO'] if build_type == "build-for-testing"
         parts += build_settings
-        parts += actions
+        parts += actions(build_type)
         parts += suffix
         parts += pipe
 
@@ -46,11 +47,11 @@ module Snapshot
         build_settings
       end
 
-      def actions
+      def actions(build_type)
         actions = []
-        actions << :clean if Snapshot.config[:clean]
-        actions << :build # https://github.com/fastlane/snapshot/issues/246
-        actions << :test
+        actions << :clean if Snapshot.config[:clean] and build_type != "test-without-building"
+        # actions << :build # https://github.com/fastlane/snapshot/issues/246
+        actions << build_type
 
         actions
       end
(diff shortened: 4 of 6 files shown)


greencoppercompany:
That's a really good improvement for timing. Hope this PR will be merged soon :)

samskiter:
+1

mfurtak:
Hi @hjanuschka - thanks for looking into this promising opportunity! 🙌  While the approach here seems like it should be a big win, I'm not seeing the same improvement when I run your test.

Runs:

master | This PR
------- | -------
5m30.245s | 5m50.148s
5m36.411s | 5m49.829s
5m37.685s | 5m48.498s

Average time in seconds:

master | This PR
------- | -------
334.78s | 349.49s

I think before we could make a change like this become the new default, we'd want to allow people to opt-in to the changes. That way people can try them out and report back about speed improvements, or any problems, without immediately affecting all of our users. 👍 

Check out 
https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/features.rb and https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/itunes_transporter.rb#L295 for an example of how to define and check a feature switch to allow people to opt-in to behavior changes.

mfurtak (changes requested):
A change to the core of how building and testing works should be protected behind a feature switch, so that we can get community feedback before affecting all users.

hjanuschka:
ok i am trying to find out how to feature flag it.

how big was your test project, because i think if you have a lot of pods and so on, it should get a lot bigger from point of improvement.

samskiter:
Just my 2 cents - if this get's held up going into snapshot, I'd really love to have the feature available as a flag to `scan` 👍  🙏

mfurtak:
@hjanuschka I ran it against `snapshot/example` to try to replicate your results :+1:

hjanuschka:
@mfurtak  ok feature flagged it - sadly you first have to merge this one https://github.com/fastlane/fastlane/pull/7135 - or apply it locally before running snapshot, otherwise the env is ignored.

mfurtak:
@hjanuschka Ah, I should have been more specific, sorry! Each tool can and should have a `features.rb` that sets up the feature flags for itself.

Please add `lib/snapshot/features.rb` with that line that you'd added in _fastlane_core_. Then, require it after `require 'snapshot/version'` in `lib/snapshot.rb` 👍

snapshot/lib/snapshot/runner.rb
```diff
@@ -153,8 +163,7 @@ def launch(language, locale, device_type, launch_arguments)
       add_media(device_type, :video, Snapshot.config[:add_videos]) if Snapshot.config[:add_videos]
 
       open_simulator_for_device(device_type)
-
-      command = TestCommandGenerator.generate(device_type: device_type)
+      test_command = TestCommandGenerator.generate(device_type: device_type, build_type: "test-without-building")
```
hjanuschka: it is prepared to be not feature flagged, and ignored/forced-old-values in test_command_generator.rb

mfurtak: Oh, I see. 👍

mfurtak:
OK, looking pretty good! Are you interested in updating `test_command_generator_spec` to cover these new possibilities?

hjanuschka:
@mfurtak i am not sure how to feature flag inside a spec, you have a sample ?

mfurtak:
Great question. Here's what I would suggest:

This is from `fastlane_core/spec/spec_helper.rb`

```ruby
# Executes the provided block after adjusting the ENV to have the
# provided keys and values set as defined in hash. After the block
# completes, restores the ENV to its previous state.
def with_env_values(hash)
  old_vals = ENV.select { |k, v| hash.include?(k) }
  hash.each do |k, v|
    ENV[k] = hash[k]
  end
  yield
ensure
  hash.each do |k, v|
    ENV.delete(k) unless old_vals.include?(k)
    ENV[k] = old_vals[k]
  end
end
```

So far, I don't think we have a great solution for sharing helper code between tests in different projects... so I've been copying this over where I need it. 😅 

That helper will let you switch on the ENV variable that controls the feature switch for the duration of your test (and then turn it back off afterwards). See:

https://github.com/fastlane/fastlane/blob/master/fastlane_core/spec/itunes_transporter_spec.rb#L130

hjanuschka:
@mfurtak thx for the sample - doing test cases tomorrow, even if the build times are not really faster than with current `build test` approach, i would like to see this as a feature-flag option.

and hopefully apple improves the test-without-building

samskiter:
OOI does this look like it's going to go in?

hjanuschka:
@mfurtak  spec added  - thx for the env helper!

@samskiter we tested it on different apps (different size) multiple runs (different machines), and in fact it never was faster than with the traditional snapshot.

anyway i think we should get this into snapshot (behind feature flag, until we are sure it works)

samskiter:
@hjanuschka it will be faster for us - we aren't using this for snapshot - we need it for `scan` so we can build once and then run on multiple machines

samskiter:
ah, sorry i didn't realise this was different from https://github.com/fastlane/fastlane/pull/7130

my bad

Jules2010:
Is the time on failed ci run, saying 41 mins ago correct @hjanuschka I'm sure nothing had changed on this thread for a couple of weeks.

It would be great to get this feature.

Cheers.

hjanuschka:
@Jules2010  i just rebased master, so that the PR's stays mergeable.

@mfurtak - any ideas on how the core team proceeds on the whole `build-for-testing` stuff?

batesyuk:
@hjanuschka love this idea but how do you get this to work now it is a 'feature' with fastlane? tried setting `--env FASTLANE_SNAPSHOT_BUILD_FOR_TESTING` but it still compiles for every iteration. What am I missing?

hjanuschka:
you need to set the env before you run it (and for sure, you need this PR checked out 👍 )

```
FASTLANE_SNAPSHOT_BUILD_FOR_TESTING=1 fastlane ...
```

or for snapshot

```
FASTLANE_SNAPSHOT_BUILD_FOR_TESTING=1 snapshot
```

Jules2010:
@hjanuschka 

How do I go about getting this to run on my machine, presumably the code needs building?

I use git a work on windows, but I'm not sure about connecting to remote server from the command line.

I've found your branch, I think...
https://github.com/hjanuschka/fastlane/tree/snapshot_testing/snapshot

batesyuk:
I might be mis-understanding the concept of build-to-test but isn't it meant to compile once and then use that to run the tests? When I use this it still seems to compile the same amount of times as before?

Jules2010:
That's what I asked in my original question which was said to be a duplicate

Jules2010:
@batesyuk are you implying that, that isn't the functionality that @hjanuschka has provided ?

Jules2010:
@batesyuk I think you mean this one, I got confused too https://github.com/fastlane/fastlane/pull/7130

Jules2010:
@hjanuschka can the build once feature from fastlane be introduced into snapshot now?

mfurtak:
This has merge conflicts and seems to be stalled, so I'm going to close this for now.

ergunkocak:
Hi,

So do we have this feature or not? I could not see in documentation (fastlane action snapshot)

In my opinion this is a very helpful feature to have. Also after first run if a test fails and we just want to rerun that would be great to have a switch for not to compile bu just use the one installed or compiled already.

Cheers

samskiter:
@ergunkocak Great point about retrying - we'll have to do the same. I believe we do have this feature now (see: https://github.com/fastlane/fastlane/issues/7130)

Another interesting thing worth considering is Bluepill: https://github.com/linkedin/bluepillI I wonder if we could add support for this into fastlane

ergunkocak:
So how can i test this feature? What shall i do in Snapfile?

PS: Tnx for fast response and bluepill link :)

samskiter:
@ergunkocak so it should be available on scan commands (you can see examples here https://github.com/fastlane/fastlane/pull/7130)

There is a great WWDC talk about this option here here http://devstreaming.apple.com/videos/wwdc/2016/409jh83sf1h8dqrt00q/409/409_advanced_testing_and_continuous_integration.pdf

ergunkocak:
@samskiter when i run scan (if i am not wrong) it does not save screenshots?

By the way i could not find a comparison chart for scan vs snapshot. What are most important differences between scan and snapshot?

Cheers

samskiter:
hey @ergunkocak you might be jumping in a bit at the deep end with this PR - have a read of these: https://github.com/fastlane/fastlane#fastlane-toolchain

`scan` is actually for running tests

potmo:
Is someone working on this PR? I'd really like to have this feature. I have a very slow build that I would like to run on multiple devices and multiple languages and it's super slow to recompile every for each entry in that matrix

Jules2010:
+1

Issue #7150, opened 2016-11-23
[Spaceship] Allow auto releasing no earlier than a specific date


enozero:
### Issue Description
Using the `auto_release_date` field, we can set for an app to be released no earlier than a certain date and time once the app is approved. The field is an epoch time (GMT) value.

enozero:
I'm already working on this feature.

mpirri:
#7151 has been merged to master, closing this issue. Thanks for the contribution, @enozero !

PR #7151, opened 2016-11-23
Add Automatic Release Date for Tunes Submission

andyvanosdale:
Implements https://github.com/fastlane/fastlane/issues/7150

Exposes the auto_release_date field in AppVersion. The field takes in
a millisecond value. Setting this will supercede the release_on_approval
field, so this field must be nil for release_on_approval to be used.

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -45,6 +45,11 @@ module Spaceship
       # @return (Bool) Should the app automatically be released once it's approved?
       attr_accessor :release_on_approval
 
+      # @return (Fixnum) Milliseconds for releasing in GMT (e.g. 1480435200000 = Tue, 29 Nov 2016 16:00:00 GMT).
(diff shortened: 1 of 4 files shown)


milch:
Hi @enozero, thanks for your contribution and sorry for the late reply! 

It seems like there is no user facing change to use this feature, so someone will probably have to pick this up later and implement it then if they so desire. Apparently there are some conditions on using this feature that you mention in the commit message, but should also be included in the code, e.g.:

```
Setting this will supercede the release_on_approval field, so this field must be nil for release_on_approval to be used.
```

The tests should also reflect this requirement.

KrauseFx:
Hey @enozero, have you had the time to look into @milch's comment?

enozero:
I have but haven't been able to update the change yet and will do so soon!

enozero:
@KrauseFx @milch I've updated the PR with the requested changes. Please take another look and let me know if you have any questions. Thanks!

mpirri (approved):
Look good, thanks for the contribution!

Issue #7149, opened 2016-11-23
Allow rejecting app versions


enozero:
### Issue Description
App versions can be developer rejected while waiting for export compliance, waiting for review, and in review. Checking the `can_reject_version` flag in `AppVersion`, we can then call the `reject` `POST` API to have the app version rejected.

enozero:
I'm already starting to work on this

hjanuschka:
love this one - rejecting an submited but not reviewed build - is the last manual step in my workflow (luckily doesn't happen that often), let me know if i can help you anywhere!

enozero:
Implementation was pretty simple. PR is available

mfurtak:
Merged #7149 👍

PR #7152, opened 2016-11-23
Allow rejecting app versions

andyvanosdale:
Implements https://github.com/fastlane/fastlane/issues/7149

New method reject! on AppVersion. The version must be in a rejectable state,
as notified in the canRejectVersion field.


mfurtak (approved):
Fantastic PR - thanks! 🚀

Issue #7155, opened 2016-11-24
[sigh] Generated `-p` option of resign is strange


yukiasai:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

When Hash is given to provisioning_profile to resign, `-p` option is wrong

Hash

```
{
  "at.fastlane" => "/path/to/provision/at.fastlane.mobileprovision",
  "at.fastlane.today" => "/path/to/provision/at.fastlane.today.mobileprovision",
}
```

Generated `-p` option

```
-p /path/to/project/at.fastlane=/path/to/provision/at.fastlane.mobileprovision -p /path/to/project/at.fastlane.today=/path/to/provision/at.fastlane.today.mobileprovision
```

AppID (Key of Hash) is expanded to absolute path. This is not fulfilled the test case written in the comment.
https://github.com/fastlane/fastlane/blob/master/sigh/lib/sigh/resign.rb#L31

hjanuschka:
@yukiasai seems to be introduced with #7004  - i did a pr with a fix: https://github.com/fastlane/fastlane/pull/7157

can you give the PR a try?

hjanuschka:
@yukiasai  saw you did a PR too - closed mine, your's seems a bite nicer with the test case too! 🙌

yukiasai:
@hjanuschka hi! sorry for confusing you. 🙇

hjanuschka:
@yukiasai  no problem - thx for the quick 'n nice PR

PR #7154, opened 2016-11-24
[sigh] Not fulfilled the test case written in the comment when provisioning_profiles is Hash

xai3:
Implements https://github.com/fastlane/fastlane/issues/7155

- [x] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

I also implemented a test case!
thank you.

PR #7157, opened 2016-11-24
[sigh] fix resign of nested apps

hjanuschka:
Fix a bug introduced with: https://github.com/fastlane/fastlane/pull/7004
as reported here: https://github.com/fastlane/fastlane/issues/7155


hjanuschka:
there was a another PR - i like the one more as it features a spec https://github.com/fastlane/fastlane/pull/7154

PR #7169, opened 2016-11-24
sensitive flag on config items 

hjanuschka:
as i really like to have the `sensitive` flag - lets push this a bit!
(so we can start modify existing options code to set sensitive to true.; and filter sensitive options out of `env`)

e.g.: unlock_keychain -> password

this is a rebased version of: https://github.com/fastlane/fastlane/pull/5982
original commit is preserved 🍻  @SandyChapman! - 

changes:
  * merge conflict solve
  *  use `UI.password` instead of `ask`

cc @mfurtak

### DIFF
--- a/fastlane_core/spec/configuration_spec.rb
+++ b/fastlane_core/spec/configuration_spec.rb
@@ -134,6 +134,18 @@ describe FastlaneCore do
         end
       end
 
+      describe "#sensitive flag" do
+        it "should set the sensitive flag", now: true do
+          config_item = FastlaneCore::ConfigItem.new(key: :foo,
+                                                     description: 'foo',
+                                                     type: Array,
+                                                     optional: true,
+                                                     sensitive: true,
+                                                     default_value: ['5', '4', '3', '2', '1'])
+          expect(config_item.sensitive).to eq(true)
+        end
+      end
+
       describe "arrays" do
         it "returns Array default values correctly" do
           config_item = FastlaneCore::ConfigItem.new(key: :foo,
--- a/fastlane_core/spec/configuration_spec.rb
+++ b/fastlane_core/spec/configuration_spec.rb
@@ -135,6 +135,11 @@ describe FastlaneCore do
       end
 
       describe "#sensitive flag" do
+        before(:each) do
+          allow(FastlaneCore::Helper).to receive(:is_test?).and_return(false)
+          allow(FastlaneCore::Helper).to receive(:interactive?).and_return(true)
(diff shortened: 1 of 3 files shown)


SandyChapman:
Looks good to me. Thanks for updating with the new input framework being used.

mfurtak:
This looks good. Would you be interested in establishing a `config_item_spec.rb` to test this new functionality?

hjanuschka:
@mfurtak  there are already ConfigItem specs in configuration_specs.rb
added it there.

or do you mean - i should move the item related to a own file? (maybe in a follow up PR?)

fastlane_core/spec/configuration_spec.rb
```diff
@@ -134,6 +134,46 @@
...
+          config_item = FastlaneCore::ConfigItem.new(key: :foo,
+                                                     description: 'foo',
+                                                     type: String,
+                                                     is_string: true,
+                                                     optional: false,
+                                                     sensitive: true)
+          config = FastlaneCore::Configuration.create([config_item], {})
+          allow(FastlaneCore::Helper).to receive(:is_test?).and_return(false)
+          allow(FastlaneCore::Helper).to receive(:interactive?).and_return(true)
+          allow(FastlaneCore::Helper).to receive(:ci?).and_return(false)
```
hjanuschka: configuration.rb -> `fetch`  disables itself, if `is_test?` - non `interactive?`  or `ci?`
to test if plain or password is used, i mock them - hope this is ok.

mfurtak: Yup, this is OK 👍 

To remove the repetition from each test, you can define these lines in a `before(:each)` block at the right nesting level (in your sensitive flag grouping), and remove those lines from the tests themselves.

```ruby
describe "#sensitive flag" do
  before(:each) do
    allow(FastlaneCore::Helper).to receive(:is_test?).and_return(false)
    allow(FastlaneCore::Helper).to receive(:interactive?).and_return(true)
    allow(FastlaneCore::Helper).to receive(:ci?).and_return(false)
  end

  # ...
end
```

hjanuschka:
@mfurtak  awesome, spec updated, i learned alot with ruby from hacking on fastlane  - thx to all of you for your guidance and patience 😃

mfurtak (approved):

mfurtak:
Thanks @SandyChapman and @hjanuschka  👍

KrauseFx:
That's a great change 👍

PR #7171, opened 2016-11-25
Ensure correct line containing Xcode version is used when other outpu…

danielbowden:

____

Currently the action `ensure_xcode_version` takes the first line of output to match the Xcode version.

This has been working fine for a while until today it fails:

```sh
[14:57:02]: ----------------------------------
[14:57:02]: --- Step: ensure_xcode_version ---
[14:57:02]: ----------------------------------
[14:57:02]: $ xcversion selected | head -1 | xargs echo -n
[14:57:05]: ▸ xargs: unterminated quote

[!] Exit status of command 'xcversion selected | head -1 | xargs echo -n' was 1 instead of 0.
xargs: unterminated quote
```

This is because the output of `xcversion selected` is currently:

```sh
Couldn't verify that spaceship is up to date
Xcode 8.0
Build version 8A218a
```

Ideally, spaceship would be up to date so this wouldn't occur but this caused failing builds on several projects.
This PR fixes this issue by instead taking the first line that starts with 'Xcode'

### DIFF
--- a/fastlane/lib/fastlane/actions/ensure_xcode_version.rb
+++ b/fastlane/lib/fastlane/actions/ensure_xcode_version.rb
@@ -8,7 +8,7 @@ module Fastlane
       def self.run(params)
         required_version = params[:version]
 
-        selected_version = sh "xcversion selected | head -1 | xargs echo -n"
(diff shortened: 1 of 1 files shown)


mfurtak (approved):
Thanks for the fix! 🚀

PR #7173, opened 2016-11-25
Fix rbenv install message

hotchpotch:
```rbenb install ruby-2.3.1``` is failed. change to ```2.3.1```.


mfurtak (approved):
Thanks for the fix! :rocket:

Issue #7178, opened 2016-11-25
Spaceship login problem


drakar4ik:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
When I run my ruby script, it fails on Spaceship::Tunes.login(login, password) command.

##### Complete output when running fastlane, including the stack trace and command used
Cannot run command: /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/client.rb:172:in `persistent_cookie_path': undefined method `tmpdir' for Dir:Class (NoMethodError)
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/two_step_client.rb:81:in `load_session_from_file'
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/client.rb:261:in `send_shared_login_request'
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/tunes/tunes_client.rb:164:in `send_login_request'
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/client.rb:428:in `do_login'
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/client.rb:243:in `login'
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/client.rb:90:in `login'
	from /Library/Ruby/Gems/2.0.0/gems/spaceship-0.38.0/lib/spaceship/tunes/spaceship.rb:22:in `login'
	from get_app_status.rb:11:in '\<main\>'

"require 'tmpdir'" is need there to fix the problem.

mfurtak:
Thanks for the report and fix 👍

mfurtak:
Your fix is released as part of [_spaceship_ 0.38.1](https://github.com/fastlane-old/spaceship/releases/tag/0.38.1) 🚀

PR #7179, opened 2016-11-25
fix "`persistent_cookie_path': undefined method `tmpdir' for Dir:Class (NoMethodError)" for client.rb 

drakar4ik:
fix for issue #7178


mfurtak (approved):
Thanks for the fix! 🚀

Issue #6810, opened 2016-11-01
Call update_info_plist follow by set_info_plist_value result in bplist plist file


sainttail:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

When I run actions 

```
update_info_plist(
      display_name: 'new name',
      plist_path: plist_path,
      xcodeproj: xcodeproj
    )
set_info_plist_value(
    key: 'CFBundleVersion',
    value: '2222',
    path: params[:plist_path]
  )
```
Somehow it changes normal xml `plist` to `bplist` and causes the consequence action that related to reading `plist` file broken

If I either run only one of the above action or run `set_info_plist_value` first, the `plist` is update correctly.

### You can find the test project [here](https://github.com/sainttail/TestFastlane)

I don't know this is related or not but on my machine I have 2 versions of xcode (7.3.1 and 8.0)

[INSERT OUTPUT HERE]

Here is the output everything look normal. However, if I add a action `set_build_number(plist_path: plist_path)` after plist become bplist file. It will show `invalid byte sequence in UTF-8`

```
~
~ ❯❯❯ cd De...
~ ❯❯❯ cd Desktop/per...
~ ❯❯❯ cd Desktop/personal-project/Test...
~ ❯❯❯ cd Desktop/personal-project/TestFastlane/
~ ❯❯❯ cd Desktop/personal-project/TestFastlane
~ ❯❯❯ cd Desktop/personal-project/TestFastlane
~/D/p/TestFastlane git:master ❯❯❯ fastlane beta                               ✱
[09:27:22]: ------------------------------
[09:27:22]: --- Step: default_platform ---
[09:27:22]: ------------------------------
[09:27:22]: ---------------------------
[09:27:22]: --- Step: opt_out_usage ---
[09:27:22]: ---------------------------
[09:27:22]: Disabled upload of used actions
[09:27:22]: Driving the lane 'ios beta' 🚀
[09:27:22]: -------------------------------
[09:27:22]: --- Step: update_info_plist ---
[09:27:22]: -------------------------------
[09:27:22]: Updated TestFastlane/Info.plist 💾.
[09:27:22]: ----------------------------------
[09:27:22]: --- Step: set_info_plist_value ---
[09:27:22]: ----------------------------------

+------+----------------------+-------------+
|             fastlane summary              |
[...]

PR #7188, opened 2016-11-27
Fix call set_info_plist_value resulting in bplist file

sainttail:
Refer to #6810 

Since, I'm not too familiar with ruby and this issue happens on some specific machines (which happens to be me). So I try to fix it.

This PR seem like it solve that problem but I don't know why. Can someone explain this to me?

Thanks.


mfurtak (approved):
I can confirm that this change fixes things relative to your sample project that exhibits the problem.

The two calls _should_ be equivalent, so I think this change is fine. The fact that they're resulting in different outcomes is confounding 😫

PR #7191, opened 2016-11-27
[gym] Uses `output_name` to attempt to find the correct .app package

### LINKED ISSUE #7190: [gym] Only one(/wrong) .app artefact is copied over
Hi!

I have a Mac app which contains another .app as a dependency & build artefact (it's a helper app to allow the app to boot on startup, [documented here](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLoginItems.html).

The `BuildProductsPath/Release` directory listing looks like:

```
Foo.app
Foo.app.dSYM
Foo.swiftmodule
FooHelper.app
FooHelper.app.dSYM
FooHelper.swiftmodule
... // other frameworks
```

So far, so good...

### What happens

Currently, `copy_mac_app()` only [copies the last found .app in the folder](https://github.com/fastlane/fastlane/blob/7a83e79d56deed259dfc68af446b5984d4c9c3d3/gym/lib/gym/runner.rb#L160) (in this case it's `FooHelper.app`) during a build.

My CLI output is:

```
[12:29:02]: Successfully exported and compressed dSYM file
[12:29:02]: Successfully exported the .app file:
[12:29:02]: ./FooHelper.app
```

### What should happen

I'd expect `Foo.app` to be available at the end of the build. It's `.dSYM` is, so that's a start :)

### Solutions/discussion

There are a couple of solutions I can think of to this:

 - Copy all `.app`s. This is probably not good (since you probably never need only the `*Helper.app` on its own), but it's easiest to implement
 - Specify an env var to state which `.app` to copy
 - Look into the chosen scheme's `Info.plist` to find the `CFBundleExecutable`, and copy only that (defaulting to the last .app if not found?)

I'd say the last one is the best way to do this. I'll take a look at how easy that is, but would appreciate input!

notjosh:
Fixes #7190.

I can see some circumstances where this might not be 100% the best fix, but I can't see it causing any regressions.

As a sidebar, I noticed there are two different ways this is being generated ([here](https://github.com/fastlane/fastlane/blob/11cf73824a176496b15f55902bd3c3e11435231b/gym/lib/gym/generators/package_command_generator_legacy.rb#L50-L55) and [here](https://github.com/fastlane/fastlane/blob/11cf73824a176496b15f55902bd3c3e11435231b/gym/lib/gym/runner.rb#L160) - one using `.first` and one using `.last`). I didn't bother touching the `*_legacy` (for obvious reasons), but there will potentially be inconsistencies for legacy projects (which were likely already there though, so not a big problem for this PR).

---

### DIFF
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -157,7 +157,7 @@ module Gym
 
     # Copies the .app from the archive into the output directory
     def copy_mac_app
-      app_path = Dir[File.join(BuildCommandGenerator.archive_path, "Products/Applications/*.app")].last
+      app_path = find_app_path
       UI.crash!("Couldn't find application in '#{BuildCommandGenerator.archive_path}'") unless app_path
 
       FileUtils.cp_r(app_path, File.expand_path(Gym.config[:output_directory]), remove_destination: true)
@@ -165,6 +165,18 @@ module Gym
 
       UI.success "Successfully exported the .app file:"
       UI.message app_path
+      app_path
+    end
+
+    # Finds the most likely candidate for this target's .app
+    def find_app_path
+      # guess the correct .app name from the config
+      output_path = File.join(BuildCommandGenerator.archive_path, "Products/Applications", Gym.config[:output_name] + ".app")
+      app_path = output_path if File.exist?(output_path)
+
+      # default to finding an arbitrary .app file, since we can't know any better
+      app_path ||= Dir[File.join(BuildCommandGenerator.archive_path, "Products/Applications/*.app")].last
+
       app_path
     end
 
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -60,10 +60,11 @@ module Gym
         FastlaneCore::ConfigItem.new(key: :output_name,
                                      short_option: "-n",
                                      env_name: "GYM_OUTPUT_NAME",
-                                     description: "The name of the resulting ipa file",
+                                     description: "The name of the resulting ipa/app file",
                                      optional: true,
                                      verify_block: proc do |value|
                                        value.gsub!(".ipa", "")
+                                       value.gsub!(".app", "")
                                        value.gsub!(File::SEPARATOR, "_")
                                      end),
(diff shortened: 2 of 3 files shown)


gym/lib/gym/runner.rb
```diff
@@ -168,6 +168,18 @@ def copy_mac_app
       app_path
     end
 
+    # Finds the most likely candidate for this target's .app
+    def find_app_path
+      # guess the correct .app name from the config
+      output_path = File.join(BuildCommandGenerator.archive_path, "Products/Applications", Gym.config[:output_name] + ".app")
+      app_path = output_path if File.exist?(output_path)
```
milch: We should show a `UI.user_error!` here if the `Gym.config[:output_name]` value is given but no file is found. Consider the following example:

```bash
$ gym --output_name "TetsApp.app" --scheme "TestApp" #...snip
# Produces TestApp.app and HelperForTestApp.app, the command will seemingly work correctly 
$ gym  --output_name "HelperForTetsApp.app" --scheme "TestApp" #...snip
# Produces TestApp.app and HelperForTestApp.app, the command will copy the wrong .app ... User starts pulling their hair out because they can't figure out why it works sometimes and not other times
```

milch (changes requested):

milch:
Hey @notjosh, thanks for your contribution, and congrats on your first fastlane PR! 🚀  We hope you'll be inspired to submit many more.

Your PR looks good to me, so once you've made the changes I've requested in my review I'll be happy to merge.

Since this is your first PR, when you make changes related to a review, please submit them as separate commits (as opposed to amending to a previous commit and force pushing). That makes it easier for us to review just the changes, and we'll squash the commits anyway before merging. Thanks again for your help!

KrauseFx:
Hey @notjosh, have you had the time to look into the proposed changes and update the PR? 👍

ohwutup:
Hey @notjosh, thanks for the PR! I am going to close this for now as the PR has been inactive for a while now. Feel free to ping us on here when you have had a chance to make the requested changes and fix the merge conflicts! Thank you again! 🚀 👍

Issue #7192, opened 2016-11-27
Fastlane Action reset_simulators doesn't support ios parameter


fawkeswei:
Running `snapshot reset_simulators --ios 8.1,10.0` resets and creates simulators with iOS 8.1 & iOS 10.0, but the fastlane action used in fastfile `reset_simulators` doesn't respect the `--ios` parameter given.

Fastfile is like this
```
platform :ios do
  lane :test do
    reset_simulators(
        ios: ["8.1", "10.0"]
    )
  end
end
```

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
[13:48:38]: -------------------------------------------------
[13:48:38]: --- Step: Verifying required fastlane version ---
[13:48:38]: -------------------------------------------------
[13:48:38]: Your fastlane version 1.106.2 matches the minimum requirement of 1.25.0  ✅
[13:48:38]: ------------------------------
[13:48:38]: --- Step: default_platform ---
[13:48:38]: ------------------------------
[13:48:39]: Driving the lane 'ios test' 🚀
[13:48:39]: ------------------------------
[13:48:39]: --- Step: reset_simulators ---
[13:48:39]: ------------------------------
[13:48:39]: Resetting iPhone 5
[13:48:39]: Resetting iPhone 7
[13:48:40]: Resetting iPhone 7 Plus
[13:48:40]: Simulators reset

+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 3    | reset_simulators                    | 1           |
+------+-------------------------------------+-------------+

[13:48:40]: fastlane.tools finished successfully 🎉
```

hjanuschka:
see PR: https://github.com/fastlane/fastlane/pull/7194

milch:
@fawkeswei Hey, thanks for submitting your issue! Could you go into a bit more detail of why this would be useful, or how this would benefit your workflow?

[...]

PR #7194, opened 2016-11-27
[fastlane, action] reset_simulator - add :ios version flag

hjanuschka:
as requested here: https://github.com/fastlane/fastlane/issues/7192
initially i tried: `FastlaneCore::Simulator::reset(udid: nil, name: nil, os_version: nil)`  but it depends on either udid, or name, and i think in this case its ok to get those few more lines in the action itself.

```ruby
lane :sim do
        reset_simulators(
                ios: ["8.1", "10.1"]
        )
        reset_simulators
end
```

### DIFF
--- a/fastlane/lib/fastlane/actions/reset_simulators.rb
+++ b/fastlane/lib/fastlane/actions/reset_simulators.rb
@@ -2,7 +2,17 @@ module Fastlane
   module Actions
     class ResetSimulatorsAction < Action
       def self.run(params)
-        FastlaneCore::Simulator.reset_all
+        if params[:ios]
+          params[:ios].each do |os_version|
+            FastlaneCore::Simulator.all.each do |dev|
+              if dev.os_version == os_version
+                dev.reset
+              end
+            end
+          end
+        else
+          FastlaneCore::Simulator.reset_all
+        end
         UI.success('Simulators reset')
       end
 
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/reset_simulators.rb
```diff
@@ -2,7 +2,17 @@ module Fastlane
   module Actions
     class ResetSimulatorsAction < Action
       def self.run(params)
-        FastlaneCore::Simulator.reset_all
+        if params[:ios]
```
mgrebenets: could we potentially make it part of `Simulator` class `device_manager.rb`?
The functionality seems to be reusable and can sit in the core.
There could be another function that reuses `def reset(udid: nil, name: nil, os_version: nil)` under the hood and uses `os_version` parameter.

mgrebenets: Something along the lines of 

```
def reset_versions(versions)
  match = all.detect { |device| device.os_version == os_version }
  match.reset if match
end
```

hjanuschka:
@mgrebenets 
continued in:
  * 1/2 -> https://github.com/fastlane/fastlane/pull/7196
  * 2/2 -> https://github.com/fastlane/fastlane/pull/7197

PR #7196, opened 2016-11-28
[fastlane_core] FastlaneCore::Simulator reset by version 1/2

hjanuschka:
# PR 1/2
Split of: https://github.com/fastlane/fastlane/pull/7194

as requested here: #7192

```ruby
lane :sim do
        reset_simulators(
                ios: ["8.1", "10.1"]
        )
        reset_simulators
end
```

  * 1/2 -> https://github.com/fastlane/fastlane/pull/7196
  * 2/2 -> https://github.com/fastlane/fastlane/pull/7197

### DIFF
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -186,7 +186,7 @@ module FastlaneCore
       end
 
       def reset_all_by_version(os_version: nil)
-        return false until os_version
+        return false unless os_version
         version_sim = all.select { |device| device.os_version == os_version }
         version_sim.each(&:reset) if version_sim
       end
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -185,6 +185,12 @@ module FastlaneCore
         all.each(&:reset)
       end
 
+      def reset_all_by_version(os_version: nil)
+        return false until os_version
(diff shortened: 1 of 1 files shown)


milch (changes requested):

fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -185,6 +185,12 @@ def reset_all
         all.each(&:reset)
       end
 
+      def reset_all_by_version(os_version: nil)
+        return false until os_version
```
hjanuschka: @milch  holy crap! - thx for the catch! - changed it

fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -185,6 +185,12 @@ def reset_all
         all.each(&:reset)
       end
 
+      def reset_all_by_version(os_version: nil)
+        return false unless os_version
+        version_sim = all.select { |device| device.os_version == os_version }
+        version_sim.each(&:reset) if version_sim
```
milch: I think the `if version_sim` is unnecessary here, as `select` should always at least return an empty array. If it is, Let's just do this on one line without creating a temporary var:

`all.select { |device| device.os_version == os_version }.each(&:reset)`

hjanuschka:
@milch  awesome! thx 
feedback addressed.

milch (approved):

milch:
Great, thanks! 🚀

Issue #7206, opened 2016-11-28
Fastlane Produce doesn't work with non latin chars in App Name


Poliuk:
I'm having the same issue described here: https://github.com/fastlane/fastlane/issues/5236

My app name is in Japanese: 名倉堂

Apple accepts non-latin chars on Apps when using the itunes connect web.

Issue #5236, opened 2016-06-27
app name not in english


hula1:
I'm trying to create a new application with fastlane/produce by running the following command
produce -u "$user" -a "$bundle_id" -q "$app_name" -z "$version" -y "$bundle_identifier" -m English

(as you can see I run it with bash)

but unfortunately when app_name is not in english I receive the following error

Creating new app '  ' on the Apple Dev Center
[31m
[!] Apple provided the following error info:
	There were errors in the data supplied. Please correct and re-submit.
	No value was provided for the parameter &#x27;appIdName&#x27;.[0m

is it possible to specify app_name in the language other than english?

truth3:
@hula1 Is this the first application being created for this iTunes Connect account?

hula1:
No, it's not the first time I create new application (my account already has some apps), but previously I used app names in english, when I try to use fastlane/produce with app name specified in another language then I receive that error

hula1:
Is it possible at all to use fastlane/produce with a non-english language?

vpolouchkine:
Which language are you using? I'll try it on my end.

hula1:
I use hebrew

jiyee:
same problem in using Simplified Chinese

TKBurner:
Hi! Can you give an example of an app name that is affected? Can you add the --verbose flag and share the full input and output? Thanks!

ofrioster:
I'm also using hebrew and is receive and error. 
An app example can be: "דוגמה לשם "
Do you have solution?

TKBurner:
@GT-3 Thanks for following up on this. I looped in the core team and they'll take a look. I'm also going to mark this as help wanted in case anyone wants to submit a PR. Thanks!

mfurtak:
I can reproduce this locally. But based on the code, removing all non-latin characters appears to be intentional behavior. See: https://github.com/fastlane/fastlane/blob/master/produce/lib/produce/developer_center.rb#L37-L40

@KrauseFx can you recall the history of this code and why it is important?

TKBurner:
Hi, we need to look into this on the core team before we can accept a PR. Removing the help wanted label. Thanks!

hjanuschka:
this has been fixed.  with #7210  and #7211

PR #7210, opened 2016-11-28
[spaceship] non-ASCII app name in produce 1/2

hjanuschka:
# PR 1/2

The Problem with non ASCII characters support in Produce is infact a `portal` problem.
`tunes` supports all utf8 characters (as they get converted with `.to_json`)

This one trys to fix this, allowing non-ASCII characters in app_name.

> the app name is in fact still be latinzed but, to prevent an empty string as app_name
> i added a md5 of the string

```ruby
lane :produce_demo do
  produce(app_name: "名倉堂áásaa1")
end
```
  * Branch with all PRS:
    * https://github.com/hjanuschka/fastlane/tree/produce_non_latin_full
  * PRS: 
    * 1/2 -> https://github.com/fastlane/fastlane/pull/7210
    * 2/2 -> https://github.com/fastlane/fastlane/pull/7211
    
  * Issue:
    * https://github.com/fastlane/fastlane/issues/7206
    * https://github.com/fastlane/fastlane/issues/5236

(When it was introduced)
 
  * History:
    * https://github.com/fastlane-old/produce/pull/17/files

### DIFF
--- a/spaceship/lib/spaceship/portal/portal_client.rb
+++ b/spaceship/lib/spaceship/portal/portal_client.rb
@@ -133,7 +133,8 @@ module Spaceship
     end
 
     def valid_name_for(input)
-      latinazed = input.gsub(/[^0-9A-Za-z\d\s]/, '') # remove non-valid characters
+      latinazed = input.to_slug.transliterate
+      latinazed = latinazed.gsub(/[^0-9A-Za-z\d\s]/, '') # remove non-valid characters
       # it may result in totaly empty strings that only contain chinese symbols.
       if latinazed != input
         latinazed << " "
--- a/spaceship/lib/spaceship/portal/portal_client.rb
+++ b/spaceship/lib/spaceship/portal/portal_client.rb
@@ -132,6 +132,17 @@ module Spaceship
       details_for_app(app)
     end
 
+    def valid_name_for(input)
+      latinazed = input.to_slug.transliterate.to_s # remove accents
+      latinazed = latinazed.gsub(/[^0-9A-Za-z\d\s]/, '') # remove non-valid characters
+      # it may result in totaly empty strings that only contain chinese symbols.
+      if latinazed != input
+        latinazed << " "
+        latinazed << Digest::MD5.hexdigest(input)
+      end
+      latinazed
+    end
+
     def create_app!(type, name, bundle_id, mac: false)
       # We moved the ensure_csrf to the top of this method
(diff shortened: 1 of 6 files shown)


spaceship/lib/spaceship/portal/portal_client.rb
```diff
@@ -132,6 +132,17 @@ def associate_groups_with_app(app, groups)
       details_for_app(app)
     end
 
+    def valid_name_for(input)
+      latinazed = input.to_slug.transliterate.to_s # remove accents
+      latinazed = latinazed.gsub(/[^0-9A-Za-z\d\s]/, '') # remove non-valid characters
+      # it may result in totaly empty strings that only contain chinese symbols.
+      if latinazed != input
```
milch: The only reason the md5 would need to be added would be if the latinized version of the app name is empty, right? If so, let's change the condition to `latinazed.strip.empty?`

milch (changes requested):

spaceship/lib/spaceship/portal/portal_client.rb
hjanuschka: the thing is, when we start `gsub` it may result in a app_name that already exists.
and it would be hard to explain to the users to choose another name, without a non-ascii char.
therefore i thought that checking if it has been modified, and if so, append md5

hjanuschka:
@milch added a spec, and removed the dependcy to babosa, also commented on your review

spaceship/lib/spaceship/portal/portal_client.rb
milch: Right, that makes sense. If the user looks into the developer portal manually they'll probably be surprised either way to see that their app is named differently. How about we show a `UI.warning` to the user saying something like `Your app name includes non-ASCII characters, which are not supported by the Apple Developer Portal. To fix this a unique (internal) name "#{app_name}" has been created for you. Your app's real name "#{original_app_name}" will still show up correctly on iTunes Connect and the App Store`. 

We still need the `transliterate` though so that `ÜberApp` will be created as `UberApp` instead of `berApp <md5>` on the portal.

milch: Yep, sounds good, thanks! 👍

hjanuschka:
@milch  feedback addressed

milch (approved):

Poliuk:
Thanks a lot guys

PR #7217, opened 2016-11-29
[fastlane] support aliases in actions

hjanuschka:
In order to refactor actions, and easier deprecate actions - but not breaking existing fastfiles. 
let me know what you think.

PR includes 
  * a sample in `puts.rb` action to add aliases for `println`, `echo`
  * specs to test all different types of actions - using `alias_used`

**alias_used** is called before run, so that actions may tweak options. (e.g: latest_testflight_number, app_store_version) - values are mutable

Sample use-case would be here: 
  * https://github.com/fastlane/fastlane/issues/7192#issuecomment-263499724
  * testflight, pilot - are basically the same

inside actions:
```ruby
def self.aliases
   ["println", "echo"]
end
def self.alias_used(action_alias, params)
        # only action with options
        params[:option]="overriden"
        # on string based actions
        params.replace("override")
        UI.important("#{action_alias} called please use 'puts' instead!")
end
```

Fastfile:
```ruby
lane :alias_test do
        println "asdf"
        puts "asdf"
        echo "asdf"
end
```

bildschirmfoto 2016-11-29 um 10 23 07

### DIFF
--- a/fastlane/lib/fastlane/actions/actions_helper.rb
+++ b/fastlane/lib/fastlane/actions/actions_helper.rb
@@ -1,4 +1,3 @@
-# rubocop:disable ModuleLength
 module Fastlane
   module Actions
     module SharedValues
@@ -69,36 +68,14 @@ module Fastlane
     # Returns the class ref to the action based on the action name
     # Returns nil if the action is not aailable
     def self.action_class_ref(action_name)
-      alias_found = find_alias(action_name)
-      orig_action = action_name
-      if alias_found
-        action_name = alias_found
-      end
       class_name = action_name.to_s.fastlane_class + 'Action'
       class_ref = nil
       begin
         class_ref = Fastlane::Actions.const_get(class_name)
-        if alias_found
-          # notify action that it has been used by alias
-          if class_ref.respond_to?(:alias_used)
-            class_ref.alias_used(orig_action)
-          end
-        end
       rescue NameError
         return nil
       end
       return class_ref
-    end
-
-    # lookup if an alias exists
-    def self.find_alias(action_name)
-      alias_actions.each do |key, v|
-        next unless alias_actions[key]
-        if alias_actions[key].include?(action_name)
-          return key
-        end
-      end
-      return nil
     end
 
     # load aliases of actions
--- a/fastlane/lib/fastlane/actions/actions_helper.rb
+++ b/fastlane/lib/fastlane/actions/actions_helper.rb
@@ -6,8 +6,19 @@ module Fastlane
       ENVIRONMENT = :ENVIRONMENT
     end
 
+    def self.reset_aliases
+      @alias_actions = nil
+    end
+
     def self.alias_actions
-      @alias_actions ||= {}
+      unless @alias_actions
+        @alias_actions = {}
+        ActionsList.all_actions do |action, name|
+          next unless action.respond_to?(:aliases)
+          @alias_actions[name] = action.aliases
+        end
+      end
+      @alias_actions
     end
 
     def self.executed_actions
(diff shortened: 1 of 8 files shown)


KrauseFx:
I didn't look at the code yet, however the idea is nice. We already have some aliases (e.g. [appstore](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/appstore.rb), or `testflight`).

How does this affect documentation on [docs.fastlane.tools](https://docs.fastlane.tools/actions)?

hjanuschka:
@KrauseFx right now the alias'es don't show up in all actions, which is good i think (support but not promote older names), also the doc. should be updated, and existing actions which are copie's should be reimplimented as aliases, and therefore be removed from documentation.

KrauseFx:
Maybe the aliases should be listed in the docs somewhere though? Maybe another row in the action table on docs.fastlane.tools?

KrauseFx:
Either way, I agree this would be a separate PR

hjanuschka:
@KrauseFx  yeah a follow up PR. lets get this one reviewed/tweaked/merged - and then continue with docs

milch (approved):
Looks good, thanks for the PR! 🚀

fastlane/lib/fastlane/actions/actions_helper.rb
```diff
@@ -74,6 +78,15 @@ def self.action_class_ref(action_name)
       return class_ref
     end
 
+    # load aliases of actions
+    def self.load_action_aliases
+      ActionsList.all_actions do |action, name|
+        if action.respond_to?(:aliases)
```
KrauseFx: Use `next unless action.respond_to?(:aliases)` instead 👍

fastlane/lib/fastlane/actions/actions_helper.rb
```diff
@@ -74,6 +78,15 @@ def self.action_class_ref(action_name)
       return class_ref
     end
 
+    # load aliases of actions
+    def self.load_action_aliases
```
KrauseFx: Could we move the whole block from here to `def self.alias_actions` instead and cache it there in the `@alias_actions` variable? This way we can lazy init and can remove the code from `fastlane.rb` 👍

fastlane/spec/action_spec.rb
```diff
@@ -14,6 +14,39 @@
       end
     end
 
+    describe "can call alias action" do
+      it "redirects to the correct class and method" do
+        result = Fastlane::FastFile.new.parse("lane :test do
+          println \"alias\"
+        end").runner.execute(:test)
+      end
+      it "alias can override option" do
```
KrauseFx: Please add spaces after an `end`, not sure how to tell `rubocop` to check this.

fastlane/spec/action_spec.rb
```diff
@@ -14,6 +14,39 @@
       end
     end
 
+    describe "can call alias action" do
```
KrauseFx: Great tests ❤️

fastlane/lib/fastlane/runner.rb
```diff
@@ -97,11 +97,31 @@ def class_reference_from_action_name(method_sym)
       nil
     end
 
+    # lookup if an alias exists
+    def find_alias(action_name)
+      Actions.alias_actions.each do |key, v|
+        next unless Actions.alias_actions[key]
+        if Actions.alias_actions[key].include?(action_name)
```
KrauseFx: How about also using a `next unless` like on line 103?

KrauseFx:
Hey @hjanuschka, I took a closer look at the implementation, overall this is really great, however I'd like some smaller changes to be adapted before we merge it 👍  Thanks again for all your patience and contribution

hjanuschka:
@KrauseFx feedback addressed. i also had to add a method `self.reset_aliases`  which invalidates the now lazy loaded `action_aliases` - so that stuff like `import_from_git` works too.

let me know if you have any further improvements i can implement!

milch:
I've merged it for now, since @KrauseFx's review was addressed. Thanks for your PR! 🚀

fastlane/lib/fastlane/actions/actions_helper.rb
```diff
@@ -6,6 +6,21 @@ module SharedValues
       ENVIRONMENT = :ENVIRONMENT
     end
 
+    def self.reset_aliases
+      @alias_actions = nil
+    end
+
+    def self.alias_actions
+      unless @alias_actions
```
KrauseFx: Beautiful ❤️

Issue #7229, opened 2016-11-29
Pilot broken after update to spaceship 0.38.2


lkraider:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
pilot --help
/usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- unicode (LoadError)
	from /usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/lib/ruby/gems/2.3.0/gems/babosa-1.0.2/lib/babosa/utf8/unicode_proxy.rb:1:in `<top (required)>'
	from /usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/lib/ruby/gems/2.3.0/gems/babosa-1.0.2/lib/babosa/identifier.rb:41:in `<class:Identifier>'
	from /usr/local/lib/ruby/gems/2.3.0/gems/babosa-1.0.2/lib/babosa/identifier.rb:31:in `<module:Babosa>'
	from /usr/local/lib/ruby/gems/2.3.0/gems/babosa-1.0.2/lib/babosa/identifier.rb:2:in `<top (required)>'
	from /usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /usr/local/lib/ruby/gems/2.3.0/gems/babosa-1.0.2/lib/babosa.rb:16:in `<top (required)>'
	from /usr/local/Cellar/ruby/2.3.1_2/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
[...]

PR #7231, opened 2016-11-29
[spaceship] unicode hotfix

hjanuschka:
see https://github.com/fastlane/fastlane/issues/7229


lkraider (approved):

KrauseFx:
This would also require to revert the revert commit, right?

hjanuschka:
@KrauseFx this PR was done before the revert, team is investigating how this could have worked in the past (as same babosa code was used in produce).

PR #7253, opened 2016-11-29
Fix the tests' allow for interactive? to be for UI, not Helper

mfurtak:
I suspect that this is causing test failures when trying to ship _fastlane_core_ with `rake push`


milch (approved):
🐝

PR #7265, opened 2016-11-30
Update information about https://developerservices2.apple.com API

KrauseFx:
This API doesn't return the certificates and devices any more


milch (approved):
📝

PR #4449, opened 2016-05-02
[Pilot] Adding in platform support for pilot

bassrock:
This is an attempt at adding in basic platform support for iTunes Connect TestFlight Build Trains.

This is the third of 3 pull requests to fix: #4427

Issue #4427, opened 2016-05-01
[Pilot] Multiple Platform Build Support/tvOS Universal Apps


bassrock:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
When uploading builds to a universal tvOS app, I constantly run into the issue: 
```ruby
pilot/lib/pilot/build_manager.rb:91:in `wait_for_processing_build': [!] undefined method `train_version' for nil:NilClass (NoMethodError)
```

When traced down I believe a contributing factor is I can submit two ipas, with the same build number, version and app id in order for it to be a universal app on the App Store. Currently Pilot only filters for processing builds with the App Version, Build Number and App Id, but it should also be filtering builds by platform, which is available in each respective build data.

Another contributing issue to this could be that when I run:
```shell
bundle exec pilot builds
```
Not all of my builds show. For example here is the iTunes Connect Data and the respective pilot output:
![screen shot 2016-05-01 at 2 37 27 am](https://cloud.githubusercontent.com/assets/1010384/14941120/ad31e054-0f45-11e6-90cc-9049dc34e52d.png)
![screen shot 2016-05-01 at 2 38 03 am](https://cloud.githubusercontent.com/assets/1010384/14941122/c07f204a-0f45-11e6-908f-2d5cb04df713.png)

It looks as though the ```pilot list``` command is only showing one of the build trains which in this case is the tvOS one. When I currently encounter the above issue it is when I try to upload the iOS version so I am wondering if it is because pilot is only recognizing the tvOS build train.

bassrock:
Ok so I think I found the line that is causing the issue:
https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/tunes/build_train.rb#L54

In my app I like to keep the Major, Minor version the same on iOS and on tvOS so the above line is overriding the selected app train, not taking into account the applications platform.

bassrock:
And actually I think it goes down to adding a platform attribute to the tunes/application.rb

TKBurner:
@bassrock Always great to hear from you. Thanks for the PRs we'll work on them with you to get them merged 👍

vfog:
[...]

PR #7267, opened 2016-12-01
Teach pilot about apps with multiple platforms

tmm1:
It is possible now in iTunes to create an app which contains both iOS and tvOS binaries. Such an app confuses pilot immensely, and it is not possible to upload or distribute tvOS builds to testflight because the code assumes you are uploading an iOS app.

With this PR, you can use `pilot builds --platform appletvos` etc to perform testflight actions on a specific platform only.


spaceship/lib/spaceship/tunes/build_train.rb
```diff
@@ -118,11 +118,11 @@ def latest_build
 
       # @param (testing_type) internal or external
       def update_testing_status!(new_value, testing_type, build = nil)
-        data = client.build_trains(self.application.apple_id, testing_type, platform: self.application.platform)
-
         build ||= latest_build if testing_type == 'external'
         testing_key = "#{testing_type}Testing"
 
+        data = client.build_trains(self.application.apple_id, testing_type, platform: build.platform)
```
KrauseFx: Why is this change necessary? Should we add some code comments so this doesn't break in the future?

pilot/lib/pilot/options.rb
```diff
@@ -134,6 +134,13 @@ def self.available_options
                                      verify_block: proc do |value|
                                        ENV["FASTLANE_TEAM_ID"] = value.to_s
                                      end),
+        FastlaneCore::ConfigItem.new(key: :platform,
+                                     env_name: "PILOT_PLATFORM",
+                                     description: "The optional platform for which to retrieve builds (defaults to all platforms)",
+                                     optional: true,
+                                     verify_block: proc do |value|
+                                       UI.user_error!("The platform can only be ios, appletvos") unless %('ios', 'appletvos').include? value
```
KrauseFx: Can you surround `value` with `()`

spaceship/lib/spaceship/tunes/build_train.rb
tmm1: Using `self.application.platform` doesn't make any sense, because the application has binaries for both platforms. Instead we use `build.platform` so that build trains for the correct platform is updated.

pilot/lib/pilot/options.rb
tmm1: fixed.

tmm1:
Looks like @bassrock has been working on something similar in https://github.com/fastlane/fastlane/pull/4449 for https://github.com/fastlane/fastlane/issues/4427

bassrock:
Actually it just got approved yesterday.

tmm1:
@mpirri Would appreciate some review on here.

pilot/lib/pilot/build_manager.rb
```diff
@@ -81,12 +82,8 @@ def list(options)
         config[:app_identifier] = UI.input("App Identifier: ")
       end
 
-      if config[:app_platform].to_s.length == 0
-        config[:app_platform] = ask("App Platform (ios, appletvos, osx): ")
-      end
-
-      UI.user_error!("App Platform must be ios, appletvos, or osx") unless ['ios', 'appletvos', 'osx'].include? config[:app_platform]
-      builds = app.all_processing_builds(platform: config[:app_platform]) + app.builds(platform: config[:app_platform])
+      platform = fetch_app_platform(required: false)
```
mpirri: ❤️ Nice reduction in code duplication.

mpirri:
Looks 💯 , thanks for this!  :+1:

bassrock:
Thanks @tmm1! My code was pulled from another branch I did and with how fast fast lane changes, looks like I missed some stuff!!

Issue #7058, opened 2016-11-17
Upload Android APK to S3


jerome-cp:
At Chauffeur Privé, we are big fans of fastlane tools so we propose a simple plugin to be able to upload an APK to amazon S3.

The current s3 upload plugin is very helpful but unfortunately it doesn't handle very well APKs.. so we decided to create a S3 APK plugin inspired by the actual s3 plugin.

The plugin is available here : 
https://github.com/transcovo/fastlane-plugin-s3-android

Feel free to download it, use it, comment it as needed :-)

I hope this will help some of us using fastlane with Android to build better apps

Cheeers

asfalcone:
Thanks @jerome-cp 👍 

I'm going to close this issue since there isn't action we can take here. We'll be sure to include this plugin in the available plugins list on our next release 🚀

jerome-cp:
Great ! 👍

PR #7275, opened 2016-12-01
Add a s3_android action to upload an Android APK (tests included)

jerome-gr:
Hi, 

I know that you don't want to add new actions and you prefer to use plugins but..

The current s3 upload action is very helpful but unfortunately it doesn't handle very well APKs.. so we decided to create a S3 APK plugin inspired by the actual s3 plugin.

The purpose of this PR is to **integrate** the action "s3_android" to fastlane because i think it should be included in fastlane, exactly **like iOS**.
This "s3_android" action has the same **test coverage** as the s3 native action.

I hope this will help some of us using fastlane with Android to build better apps

Feel free if you have comments 👍 

Note that i've already presented a plugin here : https://github.com/fastlane/fastlane/issues/7058

Cheers,

### DIFF
--- a/fastlane/lib/fastlane/actions/s3_android.rb
+++ b/fastlane/lib/fastlane/actions/s3_android.rb
@@ -1,0 +1,269 @@
+require 'fastlane/erb_template_helper'
+
+module Fastlane
+  module Actions
+    module SharedValues
+      S3_APK_OUTPUT_PATH = :S3_APK_OUTPUT_PATH
+      S3_APK_HTML_OUTPUT_PATH = :S3_APK_HTML_OUTPUT_PATH
+    end
+
+    S3_ANDROID_ARGS_MAP = {
+      apk: '-f',
+      access_key: '-a',
+      secret_access_key: '-s',
+      bucket: '-b',
+      region: '-r',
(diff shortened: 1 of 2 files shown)


mfurtak:
Hi @jerome-cp - thanks for putting this together. The plugin looks great! 🚀 

As an Android engineer, I definitely feel where you're coming from. Because _fastlane_ started as a tool for iOS first, the amount of built-in things for that platform is definitely higher. However, adding actions on the basis of having Android reach parity would sign us up for a lot more code!

We also know that people want to be able to write new, better actions, and improve everyone's workflow by sharing them. However, every action that gets merged into _fastlane_ becomes something that the team has to take responsibility for, keep working in the long term, etc.

We're excited to offer a [plugin system](https://fabric.io/blog/introducing-fastlane-plugins/) because we think it strikes the right balance between growth and flexibility.

It's definitely possible for plugins to merge back into _fastlane_ proper, but we will be intentionally conservative about that process. Community ownership of fastlane plugins is a great thing, and we want to let the community show us what's truly indispensable by their usage. 📈 

It's _awesome_ to have you building great stuff for the Android _fastlane_ community. Please keep it up! 👍

PR #7077, opened 2016-11-18
[fastlane] fix unlock_keychain sierra *3/3*

hjanuschka:
This series of PR tries to solve following issue:
   * https://github.com/fastlane/fastlane/pull/7048
   
# This is PR 3/3
   
as described the path is to move `keychain_path` to `fastlane_core`:

  * PR 1/3 -> move `keychain_path` to `fastlane_core` #7075 
  * PR 2/3 -> refactor `match` to use fastlane_core `keychain_path` #7076 
  * PR 3/3 -> refactor `unlock_keychain` to use fastlane_core `keychain_path` #7077 
     
     
needs to be merged in the right order

PR #7276, opened 2016-12-01
Try more combinations of keychain paths

milch:
As described here: https://github.com/fastlane/fastlane/pull/7077#issuecomment-264185745 not all combinations that previously worked still do after we migrated `keychain_paths` to `fastlane_core`.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -165,12 +165,16 @@ module FastlaneCore
       # has changed for some users: https://github.com/fastlane/fastlane/issues/5649
       #
 
+      # Remove the ".keychain" at the end of the name
+      name.sub!(/\.keychain$/, "")
+
       keychain_paths = [
         File.join(Dir.home, 'Library', 'Keychains', name),
-        File.join(Dir.home, 'Library', 'Keychains', "#{name}-db"),
-        name,
-        "#{name}-db"
+        name
       ].map { |path| File.expand_path(path) }
+
+      # Transforms ["thing"] to ["thing", "thing-db", "thing.keychain", "thing-db.keychain"]
+      keychain_paths = keychain_paths.product(["", "-db"], ["", ".keychain"]).map(&:join)
 
       keychain_path = keychain_paths.find { |path| File.exist?(path) }
       UI.user_error!("Could not locate the provided keychain. Tried:\n\t#{keychain_paths.join("\n\t")}") unless keychain_path
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -168,13 +168,19 @@ module FastlaneCore
       # Remove the ".keychain" at the end of the name
       name.sub!(/\.keychain$/, "")
 
-      keychain_paths = [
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -165,13 +165,17 @@ def self.keychain_path(name)
...
       keychain_paths = [
         File.join(Dir.home, 'Library', 'Keychains', name),
-        File.join(Dir.home, 'Library', 'Keychains', "#{name}-db"),
-        name,
-        "#{name}-db"
+        name
       ].map { |path| File.expand_path(path) }
 
+      # Transforms ["thing"] to ["thing", "thing-db", "thing.keychain", "thing-db.keychain"]
+      keychain_paths = keychain_paths.product(["", "-db"], ["", ".keychain"]).map(&:join)
```
KrauseFx: Such Ruby magic 🎩

ohwutup: This is definitely some sweet sweet functional awesomeness, but given that there are only 8 options, I'm wondering if it would be more explicit for contributors to come in and understand the code if we just spelled out each option in the array above?

ohayon (approved):
This looks great, just one comment about maybe making more explicit to make it more readable that you can take or leave. 🚀

fastlane_core/lib/fastlane_core/helper.rb
milch: IMO adding/changing one line in the `keychain_paths` is less error prone than modifying multiple lines (in the future, not in this PR), and we have the comment there explaining what happens. Otherwise, something like this will happen inevitably which is hard to spot and will only lead to more headaches:

```ruby
keychain_paths = [
         File.join(Dir.home, 'Library', 'Keychains', name),
          #...
          name,
          "#{name}-db",
          "#{name}-db.keychain",
          "#{name}.keychain",
          File.join(Dir.home, 'Library', 'Keychains', new_option),
          File.join(Dir.home, 'Library', 'Keychains', "#{new_option}-db"),
          File.join(Dir.home, 'Library', 'Keychains', "#{new_option}.keychain"),
].map { |path| File.expand_path(path) }
```

Do you think we should expand more on the comment in L176 to make it more obvious what happens? Or should I just leave it like that?

ohwutup: I think that having a list like you put in the example above is the best solution because it makes it explicit where we are looking for the keychain, and I think that changing that one line with the functional operators will cause more fallout than changing a line in the list above, also if it becomes more complicated, the `product` and `map($:join)` calls will be much harder to reason about, but I think if you would really like to keep that code in there, then yeah, a more descriptive comment would be helpful. 

I just worry about a case where using `product` doesn't make sense, like if there are permutations of a path that do not conveniently match up as they do now. For example, it could be the case that `#{name}-keychain.something` but not `#{name}.something` is possible and then using `product` will produce erroneous paths.

All this being said, I think its fine to leave it and deal with it down the road if anything comes up.

milch:
@KrauseFx @ohwutup I modified the `product` code to be easier to read/understand - let me know if it's alright now 👍

KrauseFx (approved):

PR #7277, opened 2016-12-01
Added comment to keep track when the babosa fix can be removed again.

sebastianludwig:


milch:
Good idea, thanks! I subscribed myself to the thread in https://github.com/norman/babosa/pull/42 so I'll be notified once that happens just in case.

milch (approved):
🔖

PR #7280, opened 2016-12-02
Add docs to help the community respond to issues/PRs

milch:
To make it easier for contributors to interact with our users on GitHub 🚀

### DIFF
--- a/RespondingToIssuesAndPullRequests.md
+++ b/RespondingToIssuesAndPullRequests.md
@@ -33,21 +33,28 @@ Each tool has it’s own label, e.g. `fastlane`, `fastlane_core` and `gym`.
 
 ### Recommended setup for testing the code
 
-Clone the _fastlane_ repository by running  `git clone git@github.com:fastlane/fastlane.git` in your terminal
-For each PR you want to review, make sure to add the user’s fork as a remote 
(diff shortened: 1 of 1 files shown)


hjanuschka:
@milch  how about adding this: https://gist.github.com/hjanuschka/3040fc6f371f46acff72a5e67f649b27
to easy try out PR's for users?

asfalcone:
I like what @hjanuschka is suggesting! 

I believe people with commit access can do this more easily with 
```
git fetch origin pull/1234/head:pr-1234
git checkout pr-1234
```

mfurtak (approved):
Merge when ready. 👍 I think we should chat about whether we care about naming consistency afterwards.

milch:
@hjanuschka Not a huge fan of that shell script to be honest - while it makes it easy to *try* the code, it hides the repo in a `/tmp` folder so that actually looking and reviewing the code is not as simple. However, Felix and I were talking about making a change to the fastlane bot so that it comments something along the lines of "If you want to try out this PR to see if it fixes your issue / works for you, run this code: ..." - this could be very useful for that! 

@asfalcone Didn't know that that worked, thanks! I'll add it.

PR #7284, opened 2016-12-02
Improve boolean prompt action

KrauseFx:
Now the user doesn’t have to click “Enter” to confirm boolean values any more, using our FastlaneCore UI instead


mfurtak (approved):
👍

PR #7296, opened 2016-12-04
Update Gemfile setup to use local copy by default

KrauseFx:
When someone wants to modify the fastlane code base (e.g. _fastlane_core_), it is really hard for them to actually try the changes and see them in effect. So many users submit PRs without actually running and testing the code.

With this change, users can just clone the repo, run `bundle update` in e.g. the _sigh_ sub-directory and test the changes in _fastlane_core_ when running _bundle exec sigh_ 👍 

While writing the tooling community ownership documents we noticed that before it was very tedious to get things up and running.

This will use one central `Gemfile` instead of one per tool. The `Gemfile` links to each tool using a local path, and links to _fastlane_ using the `gemspec` attribute, so that we also have all development dependencies

So all bundler-based fastlane execution now uses source code, instead of deployed Gems

### DIFF
--- a/fastlane/Gemfile
+++ b/fastlane/Gemfile
@@ -1,24 +1,3 @@
-source 'https://rubygems.org'
-
-# Specify your gem's dependencies in .gemspec
-gemspec
-
-if `cd ..; git remote -v`.include?("countdown")
-  gem "fastlane_core", path: "../fastlane_core"
-  gem "credentials_manager", path: "../credentials_manager"
-  gem "spaceship", path: "../spaceship"
-  gem "deliver", path: "../deliver"
-  gem "snapshot", path: "../snapshot"
-  gem "frameit", path: "../frameit"
-  gem "pem", path: "../pem"
-  gem "cert", path: "../cert"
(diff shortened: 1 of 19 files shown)


KrauseFx:
brb, fixing the tests

KrauseFx:
✅

fastlane/Gemfile
```diff
@@ -1,24 +1,3 @@
-source 'https://rubygems.org'
-
-# Specify your gem's dependencies in .gemspec
-gemspec
-
-if `cd ..; git remote -v`.include?("countdown")
```
KrauseFx: @mfurtak we kind of already had this, in a quite hacky way before, but only for the _fastlane_ tool

mfurtak: Right, I remember it...

mfurtak (approved):
I think the end result for ease of local development is a good one. The change also behaves more like what I would expect for a "monorepo".

We should pay extra attention and look for build & ship behavior that we're not used to, especially on CI once this goes in.

hjanuschka:
this is a great change! 🚗

PR #7301, opened 2016-12-05
[all] remove all $ globals

hjanuschka:
lets try to get rid of $ variables.
feedback welcome. - i ll split the PR once its reviewed.

### DIFF
--- a/fastlane_core/lib/fastlane_core/globals.rb
+++ b/fastlane_core/lib/fastlane_core/globals.rb
@@ -1,7 +1,7 @@
 module FastlaneCore
   class Globals
     def self.captured_output
-      @captured_output || = ""
+      @captured_output ||= ""
     end
 
     class << self
--- a/fastlane_core/lib/fastlane_core/globals.rb
+++ b/fastlane_core/lib/fastlane_core/globals.rb
@@ -14,7 +14,8 @@ module FastlaneCore
     end
 
     def self.capture_output?
-      return @capture_output
+      return nil unless @capture_output
+      return true
     end
 
     def self.captured_output?
@@ -25,7 +26,8 @@ module FastlaneCore
     end
 
     def self.verbose?
-      return @verbose
+      return nil unless @verbose
+      return true
     end
   end
 end
--- a/fastlane_core/lib/fastlane_core/globals.rb
+++ b/fastlane_core/lib/fastlane_core/globals.rb
@@ -14,8 +14,7 @@ module FastlaneCore
     end
 
     def self.capture_output?
-      return false unless @capture_output
-      return true
+      return @capture_output
     end
 
     def self.captured_output?
@@ -26,8 +25,7 @@ module FastlaneCore
     end
 
     def self.verbose?
-      return false unless @verbose
-      return true
+      return @verbose
     end
   end
 end
--- a/fastlane_core/lib/fastlane_core/globals.rb
+++ b/fastlane_core/lib/fastlane_core/globals.rb
@@ -1,0 +1,31 @@
+module FastlaneCore
+  class Globals
+    def self.captured_output
+      unless @captured_output
+        @captured_output = ""
+      end
+      @captured_output
+    end
+
+    class << self
+      attr_writer :captured_output
+      attr_writer :capture_output
+      attr_writer :verbose
+    end
+
+    def self.capture_output?
+      return @capture_output
+    end
+
+    def self.captured_output?
+      if @capture_output && @captured_output.length > 0
+        return true
+      end
+      return false
+    end
+
+    def self.verbose?
+      return @verbose
+    end
+  end
+end
--- a/fastlane_core/lib/fastlane_core/globals.rb
+++ b/fastlane_core/lib/fastlane_core/globals.rb
@@ -14,7 +14,8 @@ module FastlaneCore
     end
 
     def self.capture_output?
-      return @capture_output
+      return false unless @capture_output
+      return true
     end
 
     def self.captured_output?
@@ -25,7 +26,8 @@ module FastlaneCore
     end
 
     def self.verbose?
-      return @verbose
+      return false unless @verbose
+      return true
     end
   end
 end
--- a/fastlane_core/lib/fastlane_core/globals.rb
+++ b/fastlane_core/lib/fastlane_core/globals.rb
@@ -16,7 +16,7 @@ module FastlaneCore
     end
 
     def self.captured_output?
-       @capture_output && @captured_output.length > 0
+      @capture_output && @captured_output.length > 0
     end
 
     def self.verbose?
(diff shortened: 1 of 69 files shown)


milch:
Is there a good reason for why we need this?

KrauseFx:
There has been a discussion around this somewhere by @hjanuschka, I can't find the issue. In general this is a good idea, since it allows everyone to see what variables are available, and it's easier and cleaner to write tests 👍

hjanuschka:
well i actually think having global variables is not the best thing.
internally we have a penalty box where you put in some € for introducing new globals to some code 👍 

and i brought atleast two into fastlane, `$capture_output`  - `$captured_output`
i think the goal of this PR should be to encapsulate the fastlane global state.


imagine another gem that sets $verbose?!

milch:
Hmm true, hadn't thought about other gems changing the globals. `verbose` is a pretty common name so the chances of that happening aren't even that slim. 

Anyway, there are loads of errors on the tests, if we want to merge this then we'd need to get rid of those first.

hjanuschka:
@milch  going to work on this pr in the next few days. i ll cc you once i get a green test state ;)

hjanuschka:
@milch  tests pass now, i also added a explicit test for the globals class.

BUT i had to revert the changes in `credentials_manager` as it does not depend on fastlane_core. 😢 

let me know if we are ready for a PR split 👍

milch:
> BUT i had to revert the changes in credentials_manager as it does not depend on fastlane_core. 😢

Hmm, that is interesting... Could that present any problems since the globals could be out of sync?

hjanuschka:
@milch i think it would be good to also set the $verbose - what is the reason why we cannot introduce fastlane_core as a dep. to `credentials_manager`  ?

milch:
@hjanuschka 

> i think it would be good to also set the $verbose


> what is the reason why we cannot introduce fastlane_core as a dep. to credentials_manager ?

The reason is that fastlane_core depends on credentials_manager, so we would be introducing a circular dependency. @KrauseFx and I are currently working on #7436 which will solve that, so let's maybe just defer this PR until that is merged.

hjanuschka:
@milch ok lets defer it monogem FTW!

hjanuschka:
fixed merge conflicts, ready for another review round! since monogem, cred_manager should work again.

fastlane_core/lib/fastlane_core/globals.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    class << self
+      attr_writer :captured_output
+      attr_writer :capture_output
+      attr_writer :verbose
+    end
+
+    def self.capture_output?
+      return nil unless @capture_output
+      return true
```
mfurtak: Why have the extra check here? Is there a reason we want to return `nil` when the value is `false`?

fastlane_core/lib/fastlane_core/globals.rb
```diff
@@ -0,0 +1,33 @@
+module FastlaneCore
+  class Globals
+    def self.captured_output
+      unless @captured_output
+        @captured_output = ""
+      end
+      @captured_output
```
mfurtak: Is there a reason we need to assign `@captured_output = ""` as part of this? Would 

```ruby
def self.captured_output
  @captured_output || ""
end
```

be OK?

fastlane_core/lib/fastlane_core/globals.rb
```diff
@@ -0,0 +1,33 @@
...
+    def self.capture_output?
+      return nil unless @capture_output
+      return true
+    end
+
+    def self.captured_output?
+      if @capture_output && @captured_output.length > 0
+        return true
+      end
+      return false
```
mfurtak: Is there a need to conditional checks and return boolean values directly?

```ruby
def self.captured_output?
  @capture_output && @captured_output.length > 0
end
```

fastlane_core/lib/fastlane_core/globals.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def self.verbose?
+      return nil unless @verbose
+      return true
```
mfurtak: Why have the extra check here? Is there a reason we want to return nil when the value is false?

mfurtak:
I believe that _spaceship_ shouldn't be made to depend on _fastlane_core_, so that part of this refactor is problematic. /cc @KrauseFx

hjanuschka:
@mfurtak @KrauseFx  how about checking if fastlane-core is loaded/available and then use the global class, and if not found/available - fallback to `$verbose` ?

if we drop the globals class in spaceship (and keep using `$` ) - i think we have a similar situation as on the "before-monogem" time - where we had troubles with `credentials_manager`  and deferred the PR.

btw.: once we have a decision on how to deal with spaceship - i am going to work on the open comments 👍

DanToml:
@hjanuschka that seems like it should work to me :+1:

fastlane_core/lib/fastlane_core/globals.rb
lacostej: or 
`@captured_output ||= ""`
if you need to assign it

hjanuschka: thx fixed

fastlane_core/lib/fastlane_core/globals.rb
hjanuschka: there are many places where it is explcitily required to be nil, and `false`  would change the behaviour.

fastlane_core/lib/fastlane_core/globals.rb
hjanuschka: thx!! thats awesome.

fastlane_core/lib/fastlane_core/globals.rb
hjanuschka: same as above

hjanuschka (commented):
@DanToml @mfurtak feedback addressed. spaceship is no indpendant of fastlane-core, but if it is loaded, it will respect the verbose state from fastlane-core.

master rebased, conflicts solved.
really looking forward to get this done.

fastlane_core/lib/fastlane_core/globals.rb
DanToml: This seems bad, but outside the scope of this or to fix. This is something that we should fix tho

hjanuschka: yeah but i ll preferr to do this in a seperate PR - once this is merge, return a real false, and then fix the broken things (mostly test cases that are failing) - i just want to keep the mistake-surface as low as possible - as this PR already changes a lot.

endocrimes (approved):

KrauseFx (approved):
Exciting 👍

spaceship/lib/spaceship/globals.rb
```diff
@@ -0,0 +1,12 @@
+module Spaceship
+  class Globals
+    # if spaceship is run with a FastlaneCore available respect the global state there
```
KrauseFx: Nice 👍

Issue #7105, opened 2016-11-21
[sigh] Even if using wildcard provisioning profile, WKAppBundleIdentifier is rewritten


yukiasai:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I'm using app groups with iOS and Today extension. so I'm building with following configuration.

* iOS app
  * bundle identifier is `com.my.app`
  * use `com.my.app` provisioning profile. (Same as bundle identifier)
* Today extension
  * bundle identifier is `com.my.app.today`
  * use `com.my.app.today` provisioning profile. (Same as bundle identifier)
* Watch app
  * bundle identifier is `com.my.app.watch`
  * use `com.my.*` provisioning profile. (**Different form bundle identifier**)
* Watch app extension
  * bundle identifier is `com.my.app.watch.extension`
  * use `com.my.*` provisioning profile. (**Different form bundle identifier**)

In this case, when resign, the `WKAppBundleIdentifier` in `Info.plist` of Watch app extension becomes `com.my.*`.

When trying to installing application from deploygate etc., the following error is output to device log and installation will fail.

```
installd(MobileSystemServices)[52] <Notice>: 0x16dfdb000 -[MIBundle validatePluginMetadataWithError:]: 552: WatchKit 2 app extension's NSExtension.NSExtensionAttributes.WKAppBundleIdentifier value does not match WatchKit app's bundle ID (found "com.my.*"; expected "com.my.app.watch").
```

### Idea

Submited a pull request #7110

thank you.

TKBurner:
@yukiasai Thanks for contributing! We are a little behind on pull requests but will work with you soon over there. I hope you have a wonderful week!

asfalcone:
@yukiasai The PR looks good but it would be great to get some verification from the community that this meets their signing needs before merging. 👍

mgrebenets:
@yukiasai, @asfalcone 
Hey there ~

I'm not sure what's the process, but this fix actually breaks our resigning.
I'm looking into more details, but once I have identified actual reason, what do I do next?
Do I reopen this issue or open a new one?

mgrebenets:
I'm not sure on this code:

```
if [ "$REF_BUNDLE_ID" != "$NEW_REF_BUNDLE_ID" && ! "$NEW_REF_BUNDLE_ID" =~ \* ];
[...]

PR #7302, opened 2016-12-05
Fix error in resign.sh bash syntax

mgrebenets:
This is a bash syntax fix for #7105 closed with PR #7110.

There is an error in bash syntax that causes the following condition to be always false:

```
if [ "$REF_BUNDLE_ID" != "$NEW_REF_BUNDLE_ID" && ! "$NEW_REF_BUNDLE_ID" =~ \* ];
```

Here's an example of running this code in bash 3.2:

```
bash-3.2$ REF_BUNDLE_ID="app.prod" NEW_REF_BUNDLE_ID="app.*"; if [ "$REF_BUNDLE_ID" != "$NEW_REF_BUNDLE_ID" && ! "$NEW_REF_BUNDLE_ID" =~ \* ]; then echo "YES"; else echo "NO"; fi
bash: [: missing `]'
NO
bash-3.2$ REF_BUNDLE_ID="app.prod" NEW_REF_BUNDLE_ID="app"; if [ "$REF_BUNDLE_ID" != "$NEW_REF_BUNDLE_ID" && ! "$NEW_REF_BUNDLE_ID" =~ \* ]; then echo "YES"; else echo "NO"; fi
bash: [: missing `]'
NO
bash-3.2$
```

In both cases we get an error and end up in "false" branch.

The fix is to update shell syntax.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -564,8 +564,8 @@ function resign {
             REF_PROVISION=`provision_for_bundle_id $REF_BUNDLE_ID`
             # Map to the new bundle id
             NEW_REF_BUNDLE_ID=`bundle_id_for_provison "$REF_PROVISION"`
-            # Change if not the same
-            if [ "$REF_BUNDLE_ID" != "$NEW_REF_BUNDLE_ID" && ! "$NEW_REF_BUNDLE_ID" =~ \* ];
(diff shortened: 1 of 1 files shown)


yukiasai:
👍

asfalcone (approved):

asfalcone:
Thanks @mgrebenets! You are our bash hero! 🎉

PR #7322, opened 2016-12-05
Make it easier for users to test spaceship

KrauseFx:
Let's just include the content by default

### DIFF
--- a/ToolsAndDebugging.md
+++ b/ToolsAndDebugging.md
@@ -185,6 +185,8 @@ task :debug do
 end
 ```
 
+Optionally, add `Spaceship.select_team` right after the `.login` call, if you're part of multiple teams
+
 To run the newly created script, `cd` into the _spaceship_ directory and run
 
 ```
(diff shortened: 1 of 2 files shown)


ToolsAndDebugging.md
```diff
@@ -162,24 +162,31 @@ SPACESHIP_DEBUG=1 spaceship
...
   apps = Spaceship::Tunes::Application.all
+
   require 'pry'
   binding.pry
+
+  puts "All done"
 end
 ```
 
+Optionally, add `Spaceship.select_team` right after the `.login` call, if you're part of multiple teams
```
KrauseFx: I think there are 2 different team selection methods, right? Depending on if iTC or Portal

milch: Hmm, then which one gets called when you just run `Spaceship.select_team` ?

milch: Just checked, it's an alias for `Spaceship::Portal.select_team`. In that case this needs to be updated to tell users to use the correct one 👍

ohwutup:
Hey @KrauseFx and @milch! I am going to close this for now as the monogem work removed the `Rakefile` from _spaceship_ and I figured this would be easy to copy over to a better place (I just didn't know where that better place would be right now 😛 )

KrauseFx:
Great point @ohwutup 👍

PR #7324, opened 2016-12-05
[spaceship] Add missing require block

KrauseFx:
This would cause an error if spaceship isn't used from within fastlane


milch (approved):
💯

ohwutup:
Hey @KrauseFx, I'm going to close this PR as it looks like since the monogem work its no longer necessary? Obviously feel free to reopen it if needed! 🚀

KrauseFx:
Great point @ohwutup, there is no more update checker in spaceship with the mono gem 👍

PR #7325, opened 2016-12-05
[spaceship] Show update message on spaceship failure

KrauseFx:
Since the initial update message is shown on top, we really want the user to see the message if something fails


milch (approved):

KrauseFx:
Restoring this branch now

KrauseFx:
Oh alright, we don't need this any more, as we don't have a separate _spaceship_ gem any more 👍

KrauseFx:
cc @ohwutup

PR #7327, opened 2016-12-06
[WIP] Enable danger

KrauseFx:


KrauseFx:
Doesn't work currently... still work in progress

KrauseFx:
Closing, as Circle isn't stable enough for _danger_ yet 😢

hjanuschka:
@KrauseFx  can't we use travis for danger only? or some other CI?

KrauseFx:
It's a problem with Circle CI, it seems like the combination of danger and Circle isn't stable enough yet

KrauseFx:
It might be a good idea to run it on a separate CI, e.g. travis

hjanuschka:
this way it even could run in parallel - only thing is about test results, or do you think we should run the tests on travis too? 
or we could ignore test results, from danger point of view, as github already shows the red X on failed tests.

KrauseFx:
No, we don't want to introduce Travis, as the mac containers are too slow for us

PR #7328, opened 2016-12-06
Add platform parameter to spaceship iTC client's find method

milch:


KrauseFx (approved):

hjanuschka:
@milch was that introduced by the produce mac support PRs? :'(

milch:
@hjanuschka Yeah, it was missing a parameter. I just need to be more careful in the future with methods that introduce new params and check if their usage really was updated everywhere.

hjanuschka:
@milch  sorry for the troubles, i need to be more carefull too!!
anyway we can automate this? 😄

milch:
@hjanuschka Hmm, I don't think so, at least not without some terrible magicks. 

Well, I guess we could at least use danger to catch method definitions that have changed and then output a warning or something.

milch:
@KrauseFx is working on getting danger to run again here: #7327

KrauseFx:
Wow, great idea @milch 👍

hjanuschka:
@milch great idea. 
@KrauseFx  any ideas when danger will work?

KrauseFx:
@hjanuschka no, I'm working on it

PR #7334, opened 2016-12-06
[frameit] bug fixes and new design options

futuretap:
Provides fixes for the following issues:

- Want percentage padding (#7343)
- Want same font size for all shots (#7345)
- Margin between title and keyword should match a standard space (#7346)
- Bugfix: offset is changing from shot to shot (#7347)

------

### DIFF
--- a/frameit/lib/frameit/config_parser.rb
+++ b/frameit/lib/frameit/config_parser.rb
@@ -91,6 +91,10 @@ module Frameit
               UI.user_error! "padding must be type integer or pair of integers of format 'AxB' or a percentage" unless value.kind_of?(Integer) || value.split('x').length == 2 || (value.end_with?('%') && value.to_f > 0)
             end
           end
+
+          if key == 'font_scale_factor'
+            UI.user_error! "font_scale_factor must be numeric" unless value.kind_of?(Numeric)
+          end
         end
       end
     end
--- a/frameit/lib/frameit/config_parser.rb
+++ b/frameit/lib/frameit/config_parser.rb
@@ -64,11 +64,10 @@ module Frameit
         if value.kind_of?(Hash)
           validate_values(value) # recursive call
         else
-          if key == 'font'
+          case key
+          when 'font'
             UI.user_error!("Could not find font at path '#{File.expand_path(value)}'") unless File.exist?(value)
-          end
-
-          if key == 'fonts'
+          when 'fonts'
             UI.user_error!("`fonts` must be an array") unless value.kind_of?(Array)
 
             value.each do |current|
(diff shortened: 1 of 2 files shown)


frameit/lib/frameit/config_parser.rb
```diff
@@ -64,32 +64,27 @@ def validate_values(values)
...
+          when 'color'
             UI.user_error!("Invalid color '#{value}'. Must be valid Hex #123123") unless value.include?("#")
-          end
-
-          if key == 'padding'
-            unless value.kind_of?(Integer) || value.split('x').length == 2
-              UI.user_error!("padding must be type integer or pair of integers of format 'AxB'")
+          when 'padding'
+            unless value.kind_of?(Integer) || value.split('x').length == 2 || (value.end_with?('%') && value.to_f > 0)
+              UI.user_error! "padding must be type integer or pair of integers of format 'AxB' or a percentage"
```
mpirri: Perhaps this error message could indicate that this is a percentage of screen size? Also, not necessarily required right now, but is there any value for a pair of percentages to specify separate width/height multipliers?

frameit/lib/frameit/editor.rb
```diff
@@ -238,12 +245,13 @@ def put_title_into_background(background)
...
     def actual_font_size
-      [@image.width / 10.0].max.round
+      font_scale_factor = fetch_config['font_scale_factor'] || 0.1
+      [@image.width * font_scale_factor].max.round
     end
 
     # The space between the keyword and the title
     def keyword_padding
-      (actual_font_size / 2.0).round
+      (actual_font_size / 3.0).round
```
mpirri: I'm a bit worried about changing this out from under existing `frameit` users. What if we kept this value as 2.0 by default and added a `keyword_padding_factor` configuration item to allow custom values to be specified?

frameit/lib/frameit/editor.rb
```diff
@@ -106,19 +106,23 @@ def should_add_title?
...
+        background = put_title_into_background(background)
+      end
+
       if self.frame # we have no frame on le mac
         resize_frame!
         put_into_frame
 
         # Decrease the size of the framed screenshot to fit into the defined padding + background
         frame_width = background.width - horizontal_frame_padding * 2
+        frame_height = background.height - top_space_above_device - vertical_frame_padding
```
mpirri: What is the motivation for this height computation change (and the one a few lines below)?

mpirri (commented):
These are good changes for the most part. I think I'd prefer that each of the changes got its own pull request, since it makes the PR a bit harder to follow this way and we cannot roll-back just one part of the change if we wanted.

Also, is there any way to add any automated testing around these changes? Specifically, around adding new configuration options and new types for existing options.

Thanks for the submission!

frameit/lib/frameit/editor.rb
futuretap: The temporary frame_height variable is needed for the check in lines 123-125. This fixes bug #1725 where the image height doesn't fit into the frame and was cut off because of this.

frameit/lib/frameit/config_parser.rb
futuretap: Will add the " of screen size" string. Right now, only a single percentage is supported.

frameit/lib/frameit/editor.rb
futuretap: Honestly, I see the too-small spacing as a bug (#7346). Since this PR fixes all kinds of other layout bugs, there will be a layout change (to the better) anyway. It's your code so you decide of course, but I'd advise against adding just another layout config parameter. It's hard to explain and I expect that most users are just fine with the current setting. I think the scope of frameit is not to invent another language like PDF, isn't it?

futuretap:
I've submitted separate pull requests. I also looked briefly into the tests but noticed that there are barely any. I'd be fine adding coverage for just the added options (by merely copying&pasting existing tests and just changing them slightly) but can't afford right now to catch up on all your homework ;)

mpirri:
Closing this PR since this has been superseded by individual PRs for each of the changes added here.

PR #7335, opened 2016-12-06
[fastlane] anon. the sensitive options

hjanuschka:
this one try's to replace sensitive option values from a `captured_output` run.

anyone has a idea how to find sensitive options in `fastlane env` - without capture output?

### DIFF
--- a/fastlane/lib/fastlane/environment_printer.rb
+++ b/fastlane/lib/fastlane/environment_printer.rb
@@ -2,6 +2,12 @@ module Fastlane
   class EnvironmentPrinter
     def self.output
       env_info = get
+
+      # Remove sensitive option values
+      FastlaneCore::Configuration.sensitive_strings.compact.each do |sensitive_element|
+        env_info.gsub!(sensitive_element, "#########")
+      end
+
       puts env_info
       if FastlaneCore::Helper.mac? && UI.interactive? && UI.confirm("🙄  Wow, that's a lot of markdown text... should fastlane put it into your clipboard, so you can easily paste it on GitHub?")
         copy_to_clipboard(env_info)
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -51,9 +51,8 @@ module FastlaneCore
       # those will be later - replaced by ####
       if $capture_output
         available_options.each do |element|
-          if element.sensitive
-            self.class.sensitive_strings << values[element.key]
-          end
+          next unless element.sensitive
+          self.class.sensitive_strings << values[element.key]
         end
       end
 
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -35,9 +35,27 @@ module FastlaneCore
     # @!group Setting up the configuration
     #####################################################
 
+    # collect sensitive strings
+    def self.sensitive_strings
+      unless @sensitive_strings
+        @sensitive_strings = []
+      end
+      @sensitive_strings
+    end
+
     def initialize(available_options, values)
       self.available_options = available_options || []
       self.values = values || {}
+
+      # if we are in captured output mode - keep a array of sensitive option values
+      # those will be later - replaced by ####
+      if $capture_output
+        available_options.each do |element|
+          if element.sensitive
+            self.class.sensitive_strings << values[element.key]
+          end
+        end
+      end
 
       verify_input_types
       verify_value_exists
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -37,10 +37,7 @@ module FastlaneCore
 
     # collect sensitive strings
     def self.sensitive_strings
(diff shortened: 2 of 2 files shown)


fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -35,10 +35,28 @@ def self.create(available_options, values)
     # @!group Setting up the configuration
     #####################################################
 
+    # collect sensitive strings
+    def self.sensitive_strings
+      unless @sensitive_strings
```
KrauseFx: Use
```ruby
def self.sensitive_strings
  @sensitive_strings ||= []
end
```

fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -35,10 +35,28 @@ def self.create(available_options, values)
...
+
     def initialize(available_options, values)
       self.available_options = available_options || []
       self.values = values || {}
 
+      # if we are in captured output mode - keep a array of sensitive option values
+      # those will be later - replaced by ####
+      if $capture_output
+        available_options.each do |element|
+          if element.sensitive
```
KrauseFx: use
```ruby
next unless element.sensitive
self.class.sen....
```

KrauseFx (dismissed):

KrauseFx:
> anyone has a idea how to find sensitive options in fastlane env - without capture output?

what do you mean?

hjanuschka:
@KrauseFx  feedback addressed.
with this PR - we can filter out all sensitive options - when run with `--capture_output` - as it triggers the `Configuration.initialize`  - atleast on all actions that will be run (before an exception).


but it does not satisfy me 100% - as in the case of normal `fastlane env` - the `Fastfile` is loaded.
and i am looking for a "thing" to parse the Fastfile, get all actions, parse the options (matching to the action) - checking if it is sensitive and put it to the `sensitive_strings`.


hope you understand what i mean.

KrauseFx:
There is no good way to do that

milch:
@hjanuschka Yeah I don't think that'll be easy to do - one way would be to construct regexes like `/<action_name>\s*\([^\)]*<param_name>\s*(:|=>)\s*([^,\n]*),[^\)]*\)\s*\n/gm` (and a second one for the case where the param is the last parameter in the list) for each sensitive option and run them on the Fastfile contents. Depending on the length of the Fastfile and the amount of sensitive options, this might take a while though. 

Doing this might just be a bit too much magic though 🎩 ✨

milch:
Another thing that would be possible would be to create a simple Fastfile parser that'd just get the actions and all their parameters from the Fastfile. I've thought about this before so that we could make a kind of linter and/or dry-run option for fastlane `*`, but a parser like that could also be used for anonymizing the Fastfile. 

`*`: i.e. a thing that runs through the Fastfile and doesn't execute anything, but just tell you things like "this action doesn't have a parameter like that", "this parameter expects a bool but you passed a String", etc... I've ran fastlane too often and had it error out after building for 10 minutes only to tell me that I had misspelled a configuration parameter.

hjanuschka:
@milch https://gist.github.com/hjanuschka/fa7241a2462d36cb35ded66021188604 there you go, i talked about this to @KrauseFx  - in fact it seems to be to much effort.

the GIST is a working sample - which parses the fastfile - detects all actions, shims them through a `face_action` - which loads the available options from the original actions, looping `*args`  to see if this value has to be removed.


let me know what you think, i know it may be "propesed" to not use sensitive data inside fastfile (e.g: use ENV) - but in fact users are not used to this, i use api-keys myself on a few projects directly in the fastfile 👎

milch:
@hjanuschka Hmm, that looks pretty good! Can you open up an issue where we can discuss this instead of here? Then we can merge this and discuss that other code snippet in your issue 👍

fastlane_core/lib/fastlane_core/configuration/configuration.rb
milch: +1

milch (changes requested):

fastlane_core/lib/fastlane_core/configuration/configuration.rb
hjanuschka: awesome - thx!

hjanuschka (commented):
@milch, @KrauseFx  feedback addressed

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -2,6 +2,12 @@ module Fastlane
   class EnvironmentPrinter
     def self.output
       env_info = get
+
+      # Remove sensitive option values
+      FastlaneCore::Configuration.sensitive_strings.compact.each do |sensitive_element|
+        env_info.gsub!(sensitive_element, "#########")
```
milch: I don't think it matters to be honest, we could also put `"<REDACTED>"` in there to make it more clear that the value was automatically replaced.

milch (approved):

fastlane/lib/fastlane/environment_printer.rb
hjanuschka: i also think it does not matter if # or * - can we keep the #?

milch: cc @KrauseFx

Issue #7326, opened 2016-12-06
Move environment variable checking into fastlane_core


milch:
As discussed in https://github.com/fastlane/fastlane/pull/7064#issuecomment-264584574 - we should provide a method in fastlane_core to check if a boolean environment variable was set to a "falsy" value. Right now, we use many checks of the form 

```ruby
if ENV["MATCH_FORCE_ENTERPRISE"]
  # ... do work
end

# or

do_work("value) if ENV["MATCH_FORCE_ENTERPRISE"]
```

Users falsely assume that they can enable and disable boolean environment variables by setting them to truthy/falsy values (e.g. by putting `ENV["VAR"] = "0"` at the top of their Fastfile), which is not the case right now. An implementation of this could look like 

```ruby
if FastlaneCore::feature_enabled?("MATCH_FORCE_ENTERPRISE")
  # ... do work
end

# or even better, fancy dynamic method version

if FastlaneCore::env_match_force_enterprise?
  # ... do work
end
```

but I'm open to other ideas.

KrauseFx:
How about prefixing it with a class? 

```
FastlaneCore::Env.enabled?("MATCH_FORCE")
```

I'd prefer not to have dynamic methods here, as it might not immediately be clear for contributors when looking at the code on where the value comes from

milch:
> How about prefixing it with a class?

Sounds good to me! 

> I'd prefer not to have dynamic methods here, as it might not immediately be clear for contributors when looking at the code on where the value comes from

If it is implemented in the prefixed class, the implementation for the lookup will be in there.

hjanuschka:
awesome! ok when i go for it? with @KrauseFx  suggestion ?

PR #7337, opened 2016-12-06
[ALL] Fastlane env class

hjanuschka:
as discussed here: https://github.com/fastlane/fastlane/issues/7326

once reviewed, i am happily splitting it into seperated PR's!

i hope i didn't miss any parts 

the actual class is here: https://github.com/fastlane/fastlane/pull/7337/files#diff-4f6fae6b89f5d847cff92d730e83d6bf

with: `ENV["SKIP_ENV_HANDLER"]` set - it should behave like before!

/cc @KrauseFx @milch


credentials_manager/lib/credentials_manager.rb
```diff
@@ -1,3 +1,4 @@
+require 'fastlane_core'
```
milch: I'm not sure we can actually just pull in fastlane_core as a dependency to credentials_manager, as it'll create a circular dependency. When we publish updates and update dependencies they'd bump each other continuously. @KrauseFx can you weigh in on this, or am I talking nonsense?

credentials_manager/lib/credentials_manager/appfile_config.rb
```diff
@@ -154,7 +154,7 @@ def for_lane(lane_name)
         # the lane name will be verified below
       end
 
-      if ENV["FASTLANE_LANE_NAME"] == lane_name.to_s
+      if FastlaneCore::Env.enabled?("FASTLANE_LANE_NAME") == lane_name.to_s
```
milch: This condition will always be false 😉

fastlane/lib/fastlane/environment_printer.rb
```diff
@@ -146,7 +146,7 @@ def self.print_system_locale
       env_table = ""
       ["LANG", "LC_ALL", "LANGUAGE"].each do |e|
         env_icon = "🚫"
-        if ENV[e] && ENV[e].end_with?("UTF-8")
+        if FastlaneCore::Env.enabled?(e) && ENV[e].end_with?("UTF-8")
```
milch: I think the intent of this line was to check if `ENV[e]` was `nil`, not to see if it was enabled.

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -81,7 +81,7 @@ def self.last_git_commit_message
 
     # Returns the current git branch - can be replaced using the environment variable `GIT_BRANCH`
     def self.git_branch
-      return ENV['GIT_BRANCH'] if ENV['GIT_BRANCH'].to_s.length > 0 # set by Jenkins
+      return ENV['GIT_BRANCH'] if FastlaneCore::Env.enabled?('GIT_BRANCH').to_s.length > 0 # set by Jenkins
```
milch: Will always be true, the old line was correct

fastlane_core/lib/fastlane_core/env.rb
```diff
@@ -0,0 +1,12 @@
+module FastlaneCore
+  class Env
+    def self.enabled?(env)
+      return false unless ENV[env]
+      unless ENV["SKIP_ENV_HANDLER"]
+        return false if ["no", "false", "off", "0"].include?(ENV[env].to_s)
+      end
+      # if SKIP_ENV_HANDLER is set this should fallback to previous behaviour
+      return ENV[env]
```
milch: Convert to a bool please

fastlane_core/lib/fastlane_core/env.rb
```diff
@@ -0,0 +1,12 @@
+module FastlaneCore
+  class Env
+    def self.enabled?(env)
+      return false unless ENV[env]
+      unless ENV["SKIP_ENV_HANDLER"]
```
milch: Is there a reason we need this? I mean, it's kind of weird that setting `ENV["SKIP_ENV_HANDLER"] = "0"` would actually disable this.

fastlane_core/lib/fastlane_core/feature/feature.rb
```diff
@@ -11,8 +11,7 @@ def register(env_var: nil, description: nil)
       def enabled?(env_var)
         feature = @features[env_var]
         return false if feature.nil?
-        return false if ENV[feature.env_var] == '0' # '0' is a non-falsey value that should indicate feature switch disabled
-        return ENV[feature.env_var] # rely on the truthiness of the value set in the environment variable
+        FastlaneCore::Env.enabled?(feature.env_var)
```
milch: Use explicit `return` please

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -84,17 +84,17 @@ def self.is_mac?
...
     # Do we want to disable the colored output?
     def self.colors_disabled?
-      ENV["FASTLANE_DISABLE_COLORS"]
+      FastlaneCore::Env.enabled?("FASTLANE_DISABLE_COLORS")
     end
 
     # Does the user use the Mac stock terminal
     def self.mac_stock_terminal?
-      !!ENV["TERM_PROGRAM_VERSION"]
+      !!FastlaneCore::Env.enabled?("TERM_PROGRAM_VERSION")
```
milch: Remove `!!`

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -84,17 +84,17 @@ def self.is_mac?
...
     end
 
     # Does the user use iTerm?
     def self.iterm?
-      !!ENV["ITERM_SESSION_ID"]
+      !!FastlaneCore::Env.enabled?("ITERM_SESSION_ID")
```
milch: Same here

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -444,7 +444,7 @@ def parsed_info
     # If scheme not specified, do we want the scheme
     # matching project name?
     def automated_scheme_selection?
-      !!ENV["AUTOMATED_SCHEME_SELECTION"]
+      !!FastlaneCore::Env.enabled?("AUTOMATED_SCHEME_SELECTION")
```
milch: Remove `!!`

fastlane_core/spec/env_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe FastlaneCore do
+  describe FastlaneCore::Env do
```
milch: A test for nil would be great too, since an unset variable is different from a nil variable in ruby

match/lib/match.rb
```diff
@@ -28,12 +28,7 @@ def self.environments
 
   # @return [Boolean] returns true if the unsupported enterprise mode should be enabled
   def self.enterprise?
-    force_enterprise = ENV["MATCH_FORCE_ENTERPRISE"]
-
-    return false if (force_enterprise.kind_of?(String) || force_enterprise.kind_of?(Numeric)) &&
-                    (force_enterprise.to_s == "0")
-    return false if force_enterprise.kind_of?(String) &&
-                    (force_enterprise == "" || force_enterprise.casecmp("false") == 0 || force_enterprise.casecmp("no") == 0)
+    force_enterprise = FastlaneCore::Env.enabled?("MATCH_FORCE_ENTERPRISE")
```
milch: return here directly, without the `!!`

spaceship/lib/spaceship.rb
```diff
@@ -1,3 +1,4 @@
+require 'fastlane_core'
```
milch: Also not entirely sure here about the dependencies, although it should be less of a problem /cc @KrauseFx

milch (changes requested):
Thanks for your contribution Helmut! Seems like your regex was a bit overzealous at times 😉 

This is a big PR and we don't want anything breaking that's currently working, so can we get some more eyes on this? 👀  /cc @KrauseFx @mfurtak @ohwutup

credentials_manager/lib/credentials_manager/appfile_config.rb
hjanuschka: thx - in my initial implementation i had `enabled?` returning the value.

fastlane/lib/fastlane/environment_printer.rb
hjanuschka: reverted

fastlane/lib/fastlane/helper/git_helper.rb
hjanuschka: fixed

fastlane_core/lib/fastlane_core/env.rb
hjanuschka: ok i dropped the idea. initially i thought of encapsulate all `ENV` calls - and there would be a way to use an explicit `0` as a value.

but since its a true/false only thing now, i think its good to remove.

fastlane_core/lib/fastlane_core/env.rb
hjanuschka: done

fastlane_core/lib/fastlane_core/feature/feature.rb
hjanuschka: done

fastlane_core/lib/fastlane_core/helper.rb
hjanuschka: done

fastlane_core/lib/fastlane_core/helper.rb
hjanuschka: done

fastlane_core/lib/fastlane_core/project.rb
hjanuschka: done

fastlane_core/spec/env_spec.rb
hjanuschka: done

match/lib/match.rb
hjanuschka: done

credentials_manager/lib/credentials_manager.rb
hjanuschka: without this - circle failed. - not sure too here.

hjanuschka (commented):
thx for the review - i address almost all comments.

and YES please as many 👀 as possible on this PR please!

@milch  - i didn't regex this, did it by hand 😂 - thats why there are soo many errors.

and well about the `!!`  no clue why this was there before.

fastlane_core/lib/fastlane_core/env.rb
KrauseFx: To do so, use `!!ENV[env]` I think

credentials_manager/lib/credentials_manager.rb
KrauseFx: _credentials_manager_ doesn't have a dependency to _fastlane_core_. Let's not use the new env manager in _credentials_manager_ for now 👍

match/lib/match.rb
```diff
@@ -28,13 +28,7 @@ def self.environments
...
   # @return [Boolean] returns true if the unsupported enterprise mode should be enabled
   def self.enterprise?
-    force_enterprise = ENV["MATCH_FORCE_ENTERPRISE"]
-
-    return false if (force_enterprise.kind_of?(String) || force_enterprise.kind_of?(Numeric)) &&
-                    (force_enterprise.to_s == "0")
-    return false if force_enterprise.kind_of?(String) &&
-                    (force_enterprise == "" || force_enterprise.casecmp("false") == 0 || force_enterprise.casecmp("no") == 0)
-    return !!force_enterprise
+    return FastlaneCore::Env.enabled?("MATCH_FORCE_ENTERPRISE")
```
KrauseFx: ❤️

spaceship/lib/spaceship.rb
KrauseFx: Same here, spaceship doesn't know about fastlane_core and shouldn't do so

hjanuschka: @KrauseFx  just to be sure, just revert all spaceship changes,remove require, and keep the old `ENV[]` stuff.?

milch: @KrauseFx How about maintaining their own implementation so we have consistent behaviour in all the tools? 

@hjanuschka Otherwise, yeah, that's exactly what you'd need to do.

hjanuschka: i am for the "duplication" in this case
@KrauseFx  your decision

hjanuschka:
@KrauseFx @milch spaceship, and credentials_manager reverted.

KrauseFx:
Where did we stand on combining this with the feature switches?

hjanuschka:
i think we should not combine it, as features are required to be pre-defined in the features.rb.
or any other idea?

ohwutup:
I think it would be great to change `enabled?` to `truthy?` or something similar to explain exactly what the method is checking. To say something is `enabled` is to say that something is either "on" or "off" rather than something being "true" or "false" which is what the logic is expressing IMO. 😄

ohwutup:
For example, the following code:
```ruby
class PluginUpdateManager
      def self.start_looking_for_updates
        return if FastlaneCore::Env.enabled?("FASTLANE_SKIP_UPDATE_CHECK")
.....
```
I would say that `FASTLANE_SKIP_UPDATE_CHECK` can not be enabled or disabled per se, rather it is true or false.

milch (approved):

milch:
@ohwutup From my POV this is good to go, do you have anything more to add?

ohwutup:
@milch nope! this looks good to me! 

Moving forward, I wonder if there is any way for us to enforce the use of this class over `ENV`. Can we add a custom rule into rubocop or something to look for instances of `ENV` and prompt the coder to replace it with `FastlaneCore::Env`?

milch:
@ohwutup We don't really want every usage of ENV replaced, just the ones that are used as booleans - figuring that out is gonna be pretty hard. If we had `danger` running, just outputting a warning everytime the changes include ENV would work but @KrauseFx evaluated danger and there still seem to be a couple problems with danger running on Circle

ohwutup:
@milch that point is 💯

milch:
@hjanuschka Now that it's ready, you can prepare the individual PRs. It'd be great though if you could wait until Circle fixes their issues so that we aren't flooded by too many PRs now that we can't merge them.

hjanuschka:
@milch ok i ll do so!

hjanuschka:
continued in seperated PR's

PR #7348, opened 2016-12-06
[fastlane_core] add support for reporting installation method to refresher

ohayon:


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -43,7 +43,15 @@ def self.bundler?
     # Usually this means the fastlane directory is ~/.fastlane/bin/
     # We set this value via the environment variable `FASTLANE_SELF_CONTAINED`
     def self.contained_fastlane?
-      ENV["FASTLANE_SELF_CONTAINED"].to_s == "true"
+      ENV["FASTLANE_SELF_CONTAINED"].to_s == "true" && !self.is_homebrew?
+    end
+
+    def self.is_homebrew?
```
KrauseFx: Should be called `self.homebrew?` as rubocop will complain otherwise. We just have the `is_something?` for legacy reasons around, calling the actual method 👍

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -43,7 +43,15 @@ def self.bundler?
...
+      ENV["FASTLANE_INSTALLED_VIA_HOMEBREW"].to_s == "true"
+    end
+
+    def self.is_gem?
```
KrauseFx: Same here, no `is`

KrauseFx (dismissed):

fastlane_core/lib/fastlane_core/update_checker/update_checker.rb
```diff
@@ -103,11 +103,17 @@ def self.fetch_latest(url)
...
     def self.finished_running(gem_name)
-      time = (Time.now - @start_time).to_i
+      return if ENV["FASTLANE_OPT_OUT_USAGE"]
 
+      time = (Time.now - @start_time).to_i
       url = UPDATE_URL + "time/#{gem_name}"
       url += "?time=#{time}"
       url += "&ci=1" if Helper.is_ci?
+      url += "&gem=1" if Helper.is_gem?
+      url += "&bundler=1" if Helper.bundler?
```
KrauseFx: Is this true when using either the homebrew or standalone version?

KrauseFx:
When the user uses the Fabric app version of _fastlane_ it would also be `standalone=1`, right?

KrauseFx:
Nice 👍

fastlane_core/lib/fastlane_core/helper.rb
ohwutup: i think that `gem` is a reserved word in ruby no?

KrauseFx: Great point! `installed_using_gem?`? maybe?

ohwutup:
About the Mac app version of the installation, I think I might need to add in a separate check for that. I will look into that next!

KrauseFx (approved):

KrauseFx:
```
undefined method `is_homebrew?' for FastlaneCore::Helper:Module
```

`update_checker.rb` needs to be updated for the new method too 👍

ohwutup:
Hey @KrauseFx, would you mind taking another look at this? I added the migration and also added a column for the mac app. Thanks! 🎉

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -43,7 +43,19 @@ def self.bundler?
...
+
+    def self.mac_app?
+      ENV["FASTLANE_SELF_CONTAINED"].to_s == "false"
+    end
+
+    def self.homebrew?
+      ENV["FASTLANE_INSTALLED_VIA_HOMEBREW"].to_s == "true"
+    end
+
+    def self.rubygems?
```
KrauseFx: Great name, could you add a comment above each of these methods, similar to what we do for `def self.contained_fastlane?`

asfalcone (approved):

PR #7359, opened 2016-12-07
Fixed parallel upload locked due to temp file in ipa_file_analyser

philipp-heyse:
Ran into https://github.com/fastlane/fastlane/issues/1625

Would love to see this fixed in master.


milch:
Hey @philipp-heyse, thanks for your contribution, and congrats on your first fastlane PR! 🚀  We hope you'll be inspired to submit many more.

I think this is probably a good idea, although while we're working on that piece of code - I think it'd probably be better to put the plist into a temporary directory created by `Dir.mktmpdir` instead of appending an MD5 hash and placing it in `/tmp`. Subsequent invocations of that method will always return a different temporary directory so should suit this use-case perfectly 👍

philipp-heyse:
Thanks. I adapted the code by your suggestions.

philipp-heyse:
Hey @milch, can you have another look at this?

milch (approved):

milch:
@philipp-heyse Thanks, looks good now. We had a couple issues with Circle end of last week, but it should be working now - can you rebase to current master and push please, so that the tests will be triggered?

milch:
@philipp-heyse Nevermind that, I just did it for you 👍

philipp-heyse:
Thank you for your help!

Issue #7356, opened 2016-12-07
deliver fails with: [!] undefined method `description' for nil:NilClass  


Tibbs:
When I run "deliver --skip_screenshots" it fails with the following error:

```
/Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/upload_metadata.rb:40:in `block (2 levels) in upload': [!] undefined method `description' for nil:NilClass (NoMethodError)
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/upload_metadata.rb:37:in `each'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/upload_metadata.rb:37:in `block in upload'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/upload_metadata.rb:28:in `each'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/upload_metadata.rb:28:in `upload'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/runner.rb:59:in `upload_metadata'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/runner.rb:21:in `run'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/deliver-1.16.1/lib/deliver/commands_generator.rb:63:in `block (2 levels) in run'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/commander-4.4.0/lib/commander/command.rb:178:in `call'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/commander-4.4.0/lib/commander/command.rb:153:in `run'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/commander-4.4.0/lib/commander/runner.rb:444:in `run_active_command'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/fastlane_core-0.58.0/lib/fastlane_core/ui/fastlane_runner.rb:38:in `run!'
	from /Users/user/.rvm/gems/ruby-2.3.0@global/gems/commander-4.4.0/lib/commander/delegates.rb:15:in `run!'
[...]

PR #7360, opened 2016-12-07
[deliver] fix platform

hjanuschka:
issue: https://github.com/fastlane/fastlane/issues/7356


hjanuschka:
infact the issue was resolved by just setting the platform to `osx` - but i think the change in this PR should also be there.

KrauseFx (approved):

KrauseFx:
Would you mind pushing again, so that we re-trigger Circle?

hjanuschka:
@KrauseFx  done

milch (approved):
🎷

Issue #6736, opened 2016-10-25
scan: report generation takes a rather long time


AndyIsaacson:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

When I run `scan`, the final step to generate test reports take a very long time to complete.  My project is complicated to compile, but (unfortunately) there are not a ton of tests.  To give an example, compilation takes ~20m on CircleCI, the 174 tests run in 13s, and then generating a junit report takes another 8m.  I turned off HTML report generation, as it was taking even longer to run.

Is there anything I can do to speed this up?  Can I (should I?) turn it off entirely?  My builds are slow enough as it is - I'd love to shave off those extra minutes.  TBH, I'm not even sure if I've set it up properly for Circle to see the results.
##### Complete output when running fastlane, including the stack trace and command used

Just the end of the scan command, happy to provide more context if it would be useful:

[08:35:31](▸): ▸      Executed 174 tests, with 0 failures (0 unexpected) in 13.240 (13.369) seconds

[08:35:35]: ▸ Test Succeeded
+--------------------+-------+
|        Test Results        |
+--------------------+-------+
| Number of tests    | 174   |
| Number of failures | 0     |
+--------------------+-------+

[08:43:05]: Successfully generated report at '/Users/distiller/project-ios/fastlane/test_output/report.junit'
### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
Redacted the Fastfile - literally just running `scan` 

<details><summary>

✅ fastlane environment ✅</summary>

### Stack

| Key | Value |
| --- | --- |
| OS | 10.11.6 |
| Ruby | 2.2.5 |
| Bundler? | true |
| Xcode Path | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version | 8.0 |
| Git | git version 2.8.4 (Apple Git-73) |
| Installation Source | /Users/andyi/.rvm/gems/ruby-2.2.5/bin/fastlane |
| Host | Mac OS X 10.11.6 (15G1004) |
| Ruby Lib Dir | /Users/andyi/.rvm/rubies/ruby-2.2.5/lib |
| OpenSSL Version | OpenSSL 1.0.2e 3 Dec 2015 |
### fastlane gems

| Gem | Version | Update-Status |
| --- | --- | --- |
[...]

PR #5428, opened 2016-07-18
[scan] Add option to specify JUnit report output file name

PycKamil:
Scan junit report file had .junit extension, so it was not possible to use with some CI, fe. Bamboo. 

It was bugging me for a long time, in the end it was the reason to write my first line of ruby code ever. Hope it didn't breake anything 😁 

This solves #1942

Issue #1942, opened 2015-10-31
Add support for output_name parameter


fastlane-bot-helper:
<table>
      <tr>
        <td>
          <img src='https://avatars0.githubusercontent.com/u/2593278?v=3&s=70' width='35'>
        </td>
        <td>
          Original issue by @sguillope - Imported from <a href='https://github.com/fastlane/scan/issues/27'>fastlane/scan#27</a>
        </td>
      </tr>
    </table>

It would be nice if it was possible to specify the name of the output file (similar to `gym`'s `output_name` parameter).
We are running our tests on Bamboo but the [Junit Parser plugin](https://confluence.atlassian.com/bamboo/junit-parser-289277056.html) only seems to parse files with an `.xml` extension. We currently have to manually rename the output file `report.junit` to have the `.xml` extension.

nfalco79:
Mee too.
Jenkins SonarQube and other tools that parse junit report wants xml.
Using xml file for junit avoid a lot of setup.

fabioknoedt:
+1 for this feature

mikecole20:
I would also like to see this. CircleCI also expects an xml file. I was able to work around it using `mv $CIRCLE_TEST_REPORTS/xcode/report.junit $CIRCLE_TEST_REPORTS/xcode/results.xml`

TKBurner:
@fabioknoedt Thanks for the plus 1. We are going to consider this over here. Have a great day!

PycKamil:
I have added new flag for specifying full custom report name (custom_report_file_name). scan version 0.10.0 includes this change and this should solve this issue 👍

jacobvanorder:
Thank you @PycKamil!

KrauseFx:
Thanks for updating the issue and contributing to fastlane @PycKamil 🚀

kirillzh:
`--custom_report_file_name STRING Sets custom full report file name`
Is this report filename for HTML or XML report? Also, why is it still `report.junit` and not `report.xml`? 😕

TKBurner:
@kirillzh I believe it impacts both.

PR #7365, opened 2016-12-07
[scan] Reduce run time by removing report generation command as separate process

keatongreve:

Fixes https://github.com/fastlane/fastlane/issues/6736.

One of the biggest changes I wanted reviewed was removing `:custom_report_file_name` from `Scan.available_options`. There was a bit of an oversight in the addition of this parameter, as it causes issues for those who would want to provide custom report names for _multiple_ reports. See https://github.com/fastlane/fastlane/pull/5428. The resulting defect is that all files will be generated with the same name, therefore some files are lost. I don't know the process around removing action options, but I imagine the proper version bump for `scan` would be sufficient.

I replaced it with `:output_files`, which is a list of files that will all be generated under the same `:output_directory`. The values should correspond to what the user specifies for `:output_types`. I also considered making something like a hash to easily pair report types and paths, but I have never seen a fastlane action that takes a dictionary, so I didn't want to try something new (I'm sure there is one somewhere).

Also, feel free to tell me if I overdid it with all the tests I wrote. Just wanted to cover all my bases.


keatongreve:
If anyone's curious, here's the impact on our builds.
### Before
```
[16:39:46][Step 7/8] [16:39:46]: ▸ Test Succeeded
[16:43:38][Step 7/8] [16:43:38]: Successfully sent Slack notification
[16:43:38][Step 7/8] +--------------------+-------+
[16:43:38][Step 7/8] |        Test Results        |
[16:43:38][Step 7/8] +--------------------+-------+
[16:43:38][Step 7/8] | Number of tests    | 432   |
[16:43:38][Step 7/8] | Number of failures | 0     |
[16:43:38][Step 7/8] +--------------------+-------+
[16:43:38][Step 7/8] 
[16:47:20][Step 7/8] [16:47:20]: Successfully generated report at '/Applications/buildAgent/work/8214c3be3567f444/fastlane/test_output/report.junit'
[16:47:21][Step 7/8] [16:47:21]: 
[16:47:21][Step 7/8] [16:47:21]: ---------------------
[16:47:21][Step 7/8] [16:47:21]: --- Step: slather ---
[16:47:21][Step 7/8] [16:47:21]: ---------------------
// Slather output removed for brevity
[16:48:06][Step 7/8] 
[16:48:06][Step 7/8] +------+---------+-------------+
[16:48:06][Step 7/8] |       fastlane summary       |
[16:48:06][Step 7/8] +------+---------+-------------+
[16:48:06][Step 7/8] | Step | Action  | Time (in s) |
[16:48:06][Step 7/8] +------+---------+-------------+
[16:48:06][Step 7/8] | 1    | scan    | 1331        |
[16:48:06][Step 7/8] | 2    | slather | 45          |
[16:48:06][Step 7/8] +------+---------+-------------+
[16:48:06][Step 7/8] 
[16:48:06][Step 7/8] [16:48:06]: fastlane.tools just saved you 23 minutes!
```

### After
```
[17:35:10][Step 6/7] [17:35:10]: ▸ Test Succeeded
[17:35:10][Step 6/7] [17:35:10]: Successfully sent Slack notification
[17:35:10][Step 6/7] +--------------------+-------+
[17:35:10][Step 6/7] |        Test Results        |
[17:35:10][Step 6/7] +--------------------+-------+
[17:35:10][Step 6/7] | Number of tests    | 432   |
[17:35:10][Step 6/7] | Number of failures | 0     |
[17:35:10][Step 6/7] +--------------------+-------+
[17:35:10][Step 6/7] 
[17:35:11][Step 6/7] [17:35:11]: 
[17:35:11][Step 6/7] [17:35:11]: ---------------------
[17:35:11][Step 6/7] [17:35:11]: --- Step: slather ---
[17:35:11][Step 6/7] [17:35:11]: ---------------------
// Slather output removed for brevity
[17:36:01][Step 6/7] 
[17:36:01][Step 6/7] +------+---------+-------------+
[17:36:01][Step 6/7] |       fastlane summary       |
[17:36:01][Step 6/7] +------+---------+-------------+
[17:36:01][Step 6/7] | Step | Action  | Time (in s) |
[17:36:01][Step 6/7] +------+---------+-------------+
[17:36:01][Step 6/7] | 1    | scan    | 903         |
[17:36:01][Step 6/7] | 2    | slather | 49          |
[17:36:01][Step 6/7] +------+---------+-------------+
[17:36:01][Step 6/7] 
[17:36:01][Step 6/7] [17:36:00]: fastlane.tools just saved you 16 minutes!
```

milch:
Hi @keatongreve, thanks a lot for your contribution! 🚀 

This looks amazing, and wow that is a pretty significant speedup! Not sure if we really want to merge this, or our `developer hours saved` stats are going to drop too much 😉 

Rest assured, I'm going to take a look and test everything out with multiple projects. I'll also get the code reviewed then, but we might want to have a couple more eyes look at this since it's a fairly big PR. 

>  I don't know the process around removing action options, but I imagine the proper version bump for scan would be sufficient.

Breaking changes would mean a major version release, and I don't think we're up for that just now. The process in this case would be to deprecate the old parameter, both in the documentation and by outputting a warning that it is going to be removed. The old parameter should be kept around as an alias so that we can easily remove it at a later time.

keatongreve:
> test everything out with multiple projects to see if the savings are applicable to a wider range of projects

If there's any way I can help out here with data gathering, I'd be glad to. I think in general, most users should see the benefits of this, since raw xcodebuild logs are massive, regardless of project size. Just want to also note, running `clean` builds is the best measure of accuracy here, otherwise there's significantly less build output preceding the test output in the log file.

> The process in this case would be to deprecate the old parameter, both in the documentation and by outputting a warning that it is going to be removed. 

100% agree with that approach. I'm sure that I've seen that elsewhere in fastlane, but if you could point to an example, that would be greatly appreciated and I will get to work on making that change here as soon as I can.

milch:
> Just want to also note, running clean builds is the best measure of accuracy here, otherwise there's significantly less build output preceding the test output in the log file.

Alright, thanks for that tip!

>  I'm sure that I've seen that elsewhere in fastlane, but if you could point to an example, that would be greatly appreciated and I will get to work on making that change here as soon as I can.

I did a quick search but couldn't really come up with something. Maybe @KrauseFx can weigh in on the process we want to use here to make it as seamless as possible.

hjanuschka:
@keatongreve you can set `:deprecated` on the ConfigItem - then there will be a deprecation notice.

also for further deprecation notes you should use `UI.deprecated` i think

samples:
  * https://github.com/fastlane/fastlane/blob/master/gym/lib/gym/options.rb#L180
  * https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb#L42

keatongreve:
Thanks @hjanuschka! I also added a `conflicting_options` and that appears to be working as well.

keatongreve:
Any update on this from potential reviewers? Let me know if there's anything I can do to help with the process, since I'm fairly new to big PRs.

Also, for what it's worth, the company I work for has been using my fork internally for about a week now, and no issues from `scan` so far.

milch:
@keatongreve Sorry for the delay, we had to ship fastlane 2.0 last week, so that took up most of my time. 

I just tested your PR on a project that includes around ~400 tests, however, I wasn't able to see any significant speedup. When comparing averages between runs I only see about a 7% speedup which seems inconsistent with what you are getting. 

| run (#) | time (s) | branch |
| --- | --- | --- |
| 1 | 290 | master |
| 2 | 308 | master |
| 3 | 288 | master |
| **Avg** | **295** | **master** |
|  |  |  |
| 1 | 280 | scan-speedup |
| 2 | 276 | scan-speedup |
| 3 | 270 | scan-speedup |
| **Avg** | **275** | **scan-speedup** |

Here's the lane I used to test for reference: 

```ruby
lane :test do
        scan(
            output_types: "junit",
            custom_report_file_name: "test.xml",
            clean: true
        )
end
```

Could you confirm that my testing was done correctly? Is there anything else I would've needed to do to make use of this PR?

keatongreve:
I think you did everything correctly, but the main issue with the scan "delay" is not that it scales with number of tests, but the overall size of the build log, which is mostly impacted by project size.

In my example, the primary app I work on at my company is pretty big. ~2.5k source files, and ~50 targets from our CocoaPods setup as well. I don't know if you have any examples of projects like that, but it would be pretty easy to throw together a new Xcode project and add a lot of CocoaPod dependencies, at least.

When I get a chance, I'll get the file size of the raw xcodebuild log for my project.

milch:
Alright, the app I had access to with the most source files had only about 1k, so that might've been the issue. 

Since everything seems to continue working I'm very inclined to merge this anyway after I look over the code.

milch:
I'm not 100% sure what's going on here, so excuse me for asking stupid questions: can you go into a bit of detail on how you achieved the speedup? What was originally slowing down the report creation, and why is it faster now?

keatongreve:
No stupid questions at all! This was definitely a weird issue and I was definitely convinced for a while it was an xcpretty issue. I'll paraphrase my findings from my comment in https://github.com/fastlane/fastlane/issues/6736.

What scan currently does:
1. Runs `xcodebuild test [options] | xcpretty`. This xcpretty process is just for the terminal output with formatting, no reports generated. https://github.com/fastlane/fastlane/blob/master/scan/lib/scan/test_command_generator.rb#L91
2. Once the tests are done, we do `cat xcodebuild.log | xcpretty --report junit --output junit_report` to generate a tempt Junit report. this is used to print out a fancy test pass/fail table in the terminal. We use a tempfile because the user may not request a junit report themselves. source: https://github.com/fastlane/fastlane/blob/e721b30fdae69ad0f5de572431cd0eb4f422331d/scan/lib/scan/runner.rb#L52-L63
3. After that, for *every* report type requested by the user, we do another `cat xcodebuild.log | xcpretty --report #{report_type} --output #{output_name}`. source: https://github.com/fastlane/fastlane/blob/e721b30fdae69ad0f5de572431cd0eb4f422331d/scan/lib/scan/runner.rb#L83


in short, the long pauses that the original users (including myself) was seeing was waiting on a very long `cat | xcpretty` process. At least 1, even if you didn't request reports.

Example: a user runs `scan`, requesting an `html` report and a `junit` report at the end. As a result,  all of these commands would be run:
```
xcodebuild test [options] | xcpretty
cat xcodebuild.log | xcpretty --report junit --output /tmp/junit_report // temp file, discarded
cat xcodebuild.log | xcpretty --report html --output report.html // requested by user
cat xcodebuild.log | xcpretty --report junit --output report.junit // requested by me
```

This PR would convert it into a single `xcpretty` without re-processing the log file with `cat`:
```
xcodebuild test [options] | xcpretty --report junit --output /tmp/junit_report --report html --output report.html --report junit --output report.junit
```

`xcpretty` supports multiple sets of `--report` and `--output` as an intentional feature.

milch (dismissed):
Looks good to me, didn't find any big issues... Just some nitpicking. 

Since this PR is pretty big, I want to have some more 👀 on this. /cc @ohwutup @KrauseFx @joshdholtz @mpirri

milch:
@keatongreve That makes a lot of sense, thanks for that detailed explanation!

lacostej:
Does the user really need to specify a report file name for each output? Is it really useful?

In the case of multiple outputs, what about using the `custom_report_file_name` as prefix (without extension) for the report file name?

The change would be something like:
```ruby
    def determine_output_file_name(type)
       if @use_clang_report_name && type == "json-compilation-database"
         "compile_commands.json"
       elsif !@custom_report_file_name.nil?
         "@custom_report_file_name.#{type}"
       else
         "report.#{type}"
       end
     end
```

Some warnings could be added to detect/remove file extension when they are specified in the `custom_report_file_name` and if multiple outputs are used.

We would keep constructing a single call to `xcodebuild test` but without requiring the user to specify one file name per output. And the scan options wouldn't change.

keatongreve:
That would be ideal, unfortunately the reason `:custom_report_file_name` was added in the first place was because a user wanted their JUnit report's file extension to be `.xml` and not `.junit`: https://github.com/fastlane/fastlane/issues/1942. And you can see in the last comments that it did impact generation of multiple reports, so I don't know if we want to keep this option as is.

All in all, I agreed with @KrauseFx's old [comment](https://github.com/fastlane/fastlane/issues/1942#issuecomment-193428586) that `mv`ing would have sufficed, but now we support a custom file name (but not multiple) and it's hard to go back from that without introducing some breaking changes.

KrauseFx:
Hey @keatongreve, great work on preparing the change, I understand why this will make _fastlane_ faster, however I'm not a big fan of the large code changes. How much of a speed increase does this result in? 

In the past we've had issues where big PRs would break existing setups, which is something we try to avoid. 

Just trying to get a better sense of if it's worth the risk 👍

keatongreve:
Hey @KrauseFx, sorry for the delay. holidays and all that.

Like I mentioned to @milch before, the "hanging" (and speed gain from these changes) would be more noticeable against a large xcodebuild log file. The user who filed the original issue mentioned about 8 minutes to generate one report, and his log was ~250 MB (though I can't tell if that includes the report to print the terminal table). I haven't checked my build log in a while, but the delay is about 4 minutes for me, so I imagine my build logs are ~100 MB.

I think you can make the call whether this has more widespread value than just a handful of users with very large projects. If it ever comes up again in greater numbers, at least we understand the underlying issue.

Side note: Another fix I brainstormed was changing `xcpretty` to somehow split up the pre-prettified build and test output to separate files during the first run, to significantly speed up the re-parsing of the test results without going through the build output again. Or you could do it with the post-prettified output, but I imagine `xcpretty` isn't designed to parse its own output, only `xcodebuild`'s.
I imagine that the first option would be possible, but I've only skimmed the source of the project.

milch:
@krausefx I think it'd make sense for this project to be merged, since it won't make things worse for anyone - users with small projects should see negligible improvements, but users with large projects should see huge improvements. The code I reviewed also seemed to stay at least partly the same (as compared to before this PR), it was just refactored and moved around a bit.

KrauseFx:
I'm okay with merging, however there is 2 files being completely replaced, and I'd need to review these changes first.

mgrebenets:
I have notices significant slow downs of builds after moving to `fastlane` for our main app (10s of Ks of LOC).

Our builds logs are 21Mb and 36Mb on average (we have 2 targets).
For example, unit tests CI job went from 4-6 mins on average to 7-11 mins on average - quite an increase.

I'll definitely try to build gem from this branch and give it a go to see how it impacts our builds times!

mgrebenets:
@keatongreve 

I've ran some tests.
Out build log is 22Mb and the speedup is essential:

|clear derived data|speedup|
|-|-|
|NO|18%|
|YES|9%|

Can't really say why clearing derived data actually results into less speedup...

keatongreve:
Hey @mgrebenets, glad to see you got some improvements out of these changes. Not sure why your results would vary for clearing vs. not clearing DD. Maybe you could also compare the size of the xcodebuid log files and also report the actual time saved?

I'm not aware of any speed issues for `gym`, but these changes wouldn't fix that. The bottleneck I addressed here was related to running `xcpretty` post-build, which `gym` does not do.

mgrebenets:
@keatongreve Added more data to the table. For example with total run time of 257s, speedup was 41s.

keatongreve:
I'm looking into the failed tests.

AndyIsaacson:
@KrauseFx @milch Any chance this PR could get some more love?  I'm estimating this change would shave off more than 5 minutes from my builds running on CircleCI VMs

matthewellis:
+1 This issue is causing a delay of 5 or more minutes per report on travis, this is happening across two projects consistently.

milch:
@AndyIsaacson @matthewellis 

I agree, I'm still pro-merging this. There are a lot of tests, plus it has now been tested by a couple outside projects and seems to be safe. When I tested it, it worked as-advertised. 

I'll do a re-review soon, however, I still think for a large change like this we should have more 👀 on this than just one. @KrauseFx and @mfurtak what do you think?

lacostej (commented):
Some comments I wrote in December and forgot to send...

scan/lib/scan/xcpretty_reporter_options_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+      # adds another junit reporter in case the user does not specify one
+      # this will be used to generate a results table and then discarded
+      require 'tempfile'
+      Scan.cache[:temp_junit_report] = Tempfile.new("junit_report").path
+      reporter << "--report junit"
+      reporter << "--output #{Scan.cache[:temp_junit_report]}"
+      return reporter
+    end
+
+    def determine_output_file_name(type)
```
lacostej: this function should be made private. It's implementation assumes that `@output_types` has been checked and only contains valid output types. If not used that way, it could return a `nil` `index`, raising an exception when searching for the `file`

scan/lib/scan/options.rb
```diff
@@ -90,8 +90,13 @@ def self.available_options
         FastlaneCore::ConfigItem.new(key: :output_types,
                                      short_option: "-f",
                                      env_name: "SCAN_OUTPUT_TYPES",
-                                     description: "Comma separated list of the output types (e.g. html, junit)",
+                                     description: "Comma separated list of the output types (e.g. html, junit, json-compilation-database)",
                                      default_value: "html,junit"),
+        FastlaneCore::ConfigItem.new(key: :output_files,
+                                     env_name: "SCAN_OUTPUT_FILES",
```
lacostej: we could add that if `use_clang_report_name` is set, it has precedence over the name set here.

scan/spec/xcpretty_reporter_options_generator_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+        generator = Scan::XCPrettyReporterOptionsGenerator.new(false, "html", "report.html", "test_output", false)
+        reporter_options = generator.generate_reporter_options
+        temp_junit_report = Scan.cache[:temp_junit_report]
+
+        expect(temp_junit_report).not_to be_nil
+        expect(reporter_options).to include("--report junit")
+        expect(reporter_options).to include("--output #{temp_junit_report}")
+      end
+
+      it "generates options for a custom junit report with default file name" do
```
lacostej: I haven't seen a test for the following case 
1. default output types are set (html,junit)
2. `custom_report_file_name` is set but not `output_files`

this should trigger the warning of the number of elements in `output_types` and `output_files` are not the same and use the default `report.#{type}` file name, if I understood the intent.

mfurtak:
Just checking in on the status of this PR - can its comments and conflicts be resolved?

keatongreve:
I signed it.

scan/spec/xcpretty_reporter_options_generator_spec.rb
keatongreve: I wasn't sure the best way to approach writing this test, since `XCPrettyReporterOptionsGenerator` never had knowledge of `Scan.config`. `test_command_generator.rb` itself contains [the logic](https://github.com/fastlane/fastlane/pull/7365/files#diff-ef49192ad6b55e081d8efc1ddabd1ee8R106) for defaulting to `custom_report_file_name`

I added a class method, `generate_from_scan_config`, to be used by that unit test. See [here](https://github.com/fastlane/fastlane/pull/7365/files#diff-8725c22c4c270712c2f08f44769bdd77R5).

keatongreve: If you prefer I solve this a different way, that's totally fine.

keatongreve:
@mfurtak Ready for another review.

mfurtak:
@lacostej can you continue this review?

palusik86:
Also I'm looking forward to this merged and release. Currently in my project report test takes 30 minutes.

RishabhTayal:
Looking forward to get this merged for next release. 💯

KrauseFx:
This will not get merged for the next release unfortunately, as we're busy migrating the remaining spaceship parts to work with the most recent TestFlight update 👍

huy-le:
This merge is very important. As the build take 20 mins, we try really hard to make it faster every second, but the generated report step take 4 mins.

palusik86:
This PR need some attention ;-).

matthewellis:
@KrauseFx Any idea on a ETA on this now the spaceship work is merged?

KrauseFx (approved):
I just did a review and tested it on my machine, and things work well. Thanks @keatongreve for preparing this PR. 

@palusik86 @matthewellis @huy-le @RishabhTayal Could you please verify that the branch works for you for your setup by doing the following

- Create a `Gemfile` with the content
```ruby
gem "fastlane", git: "https://github.com/keatongreve/fastlane", branch: "scan-speedup"
```
- Run `bundle update`
- Run `bundle exec fastlane [lane]`

And see if that is faster and works as expected for your setup 👍

KrauseFx:
I'll merge as soon as you can confirm that things work as expected for your setups 🚀 Thanks everyone

RishabhTayal:
@KrauseFx it works for me and reduced build time from 6 mins to 4 mins. 👌

matthewellis:
@KrauseFx Tested the branch with a few projects, seems to work great. 👍

KrauseFx:
Awesome, thanks everybody involved, this is great ❤

KrauseFx:
@keatongreve: Please drop me a short email (you can find it on my GitHub profile)

lacostej:
Thanks a lot to @keatongreve for this improvement!

PR #7379, opened 2016-12-08
[gym] Remove promotion for use_legacy_build_api

hjanuschka:
lets get rid of the legacy_build_api

### DIFF
--- a/gym/lib/gym/detect_values.rb
+++ b/gym/lib/gym/detect_values.rb
@@ -21,6 +21,11 @@ module Gym
 
       config[:use_legacy_build_api] = true if Xcode.pre_7?
 
+      if config[:use_legacy_build_api]
+        UI.deprecated("the use_legacy_build_api option is deprecated")
+        UI.deprecated("it should not be used anymore - e.g.: if you use app-extensions")
+      end
+
       detect_scheme
       detect_platform # we can only do that *after* we have the scheme
       detect_configuration
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -95,14 +95,14 @@ module Gym
                                      is_string: false,
                                      optional: true),
         FastlaneCore::ConfigItem.new(key: :use_legacy_build_api,
-                                     deprecated: 'Do not use this option anymore.',
+                                     deprecated: "Don't use this option any more, as it's deprecated by Apple",
                                      env_name: "GYM_USE_LEGACY_BUILD_API",
-                                     description: "Don't use the new API because of https://openradar.appspot.com/radar?id=4952000420642816",
+                                     description: "Don't use this option any more, as it's deprecated by Apple",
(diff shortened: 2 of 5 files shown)


ohwutup:
Hey @hjanuschka, was there some discussion around this that you could reference?

KrauseFx:
I talked in the fastlane slack about this with Helmut, since there are so many people still asking about using the legacy build API, and it's not something we should recommend, since it's deprecated for over 2 years now. It's causing issues with Swift, WatchKit and apps with extensions.

KrauseFx:
We should try to get as little number of legacy build API users as possible and solve the underlying issue instead.

gym/lib/gym/detect_values.rb
```diff
@@ -21,6 +21,11 @@ def self.set_additional_default_values
 
       config[:use_legacy_build_api] = true if Xcode.pre_7?
 
+      if config[:use_legacy_build_api]
+        UI.deprecated("the use_legacy_build_api option is deprecated")
+        UI.deprecated("it should not be used anymore - e.g.: if you use app-extensions")
```
KrauseFx: Don't we have this message somewhere else already?

gym/lib/gym/options.rb
```diff
@@ -95,6 +95,7 @@ def self.plain_options
                                      is_string: false,
                                      optional: true),
         FastlaneCore::ConfigItem.new(key: :use_legacy_build_api,
+                                     deprecated: 'Do not use this option anymore.',
```
KrauseFx: Let's replace this with 
```
Don't use this option any more, as it's deprecated by Apple
```

KrauseFx (changes requested):

KrauseFx:
Maybe we should also make the warnings bigger and more prominent? I think we already show the message somewhere.

gym/lib/gym/options.rb
hjanuschka: done

gym/lib/gym/detect_values.rb
hjanuschka: i don't find it anywhere - do you have a hint?

hjanuschka (commented):
@KrauseFx feedback addressed
do you have any idea where we can put to this to address the "more prominent" thing?

gym/lib/gym/detect_values.rb
KrauseFx: Never mind

KrauseFx (approved):
Let's ship this 🎷

KrauseFx:
Could you resolve the 2 merge conflicts, and I can merge 👍

hjanuschka:
@KrauseFx should be mergable 🐙

KrauseFx:
This is a great change, thanks for preparing this  🚀

PR #7383, opened 2016-12-08
Gemfile fix Dir.pwd - with 

hjanuschka:
fixes  the fact that you cannot use `BUNDLE_GEMFILE` as used in
https://gist.github.com/hjanuschka/3040fc6f371f46acff72a5e67f649b27


KrauseFx:
Build is failing, and this might not be necessary with mono gem 👍

PR #7390, opened 2016-12-08
[deliver] no screenshot information - reword

hjanuschka:
be a bit more precise on the error message.
since screenshots won't be deleted by default anymore (only when `overwrite_screenshots` is set) 

as proposed by @KrauseFx


KrauseFx:
Thanks for preparing this change. Did you check if that's actually how _deliver_ behaves?

hjanuschka:
@krausefx - checked it - and if i am not wrong it should be how deliver behaves

KrauseFx (approved):

PR #7391, opened 2016-12-08
Make terminal table more responsive

hjanuschka:
this trys to fix the ugliness of terminal table on small terminals - by making tables fit to terminal width.
in addition to the variable size, i also did a middle_truncate.

so instead of `very long va...`  now `very ... value`  is displayed.

# Large
large

# Medium
medium

# Small
small

> Ruby 1.9.3 added 'io/console' to the standard library.
the additional require should not be a problem.

to quickly test it use this ruby script:

```ruby
require "fastlane_core"
FastlaneCore::PrintTable.print_values(config:
  { wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
      wwwwwwwwwwwwwwwwwwwwwwwwxw: "ccc bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" },

                                   hide_keys: [],
                                       title: "Summary SUM")

```

### DIFF
--- a/fastlane_core/lib/fastlane_core/print_table.rb
+++ b/fastlane_core/lib/fastlane_core/print_table.rb
@@ -4,7 +4,7 @@ module FastlaneCore
       # This method prints out all the user inputs in a nice table. Useful to summarize the run
       # You can pass an array to `hide_keys` if you don't want certain elements to show up (symbols or strings)
       # You can pass an array to `mask_keys` if you want to mask certain elements (symbols or strings)
-      def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
+      def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [], transform: :newline)
         require 'terminal-table'
 
         options = {}
@@ -24,7 +24,14 @@ module FastlaneCore
         rows = self.collect_rows(options: options, hide_keys: hide_keys.map(&:to_s), mask_keys: mask_keys.map(&:to_s), prefix: '')
 
         params = {}
-        params[:rows] = limit_row_size(rows)
+        transform = false if FastlaneCore::Env.truthy?("FL_SKIP_TABLE_TRANSFORM")
+
+        if transform
+          params[:rows] = transform_output(rows, transform)
+        else
+          params[:rows] = rows
+        end
+
         params[:title] = title.green if title
 
         puts ""
@@ -34,16 +41,60 @@ module FastlaneCore
         return params
       end
 
-      def limit_row_size(rows, max_length = 100)
-        require 'fastlane_core/string_filters'
-
-        max_key_length = rows.map { |e| e[0].length }.max || 0
-        max_allowed_value_length = max_length - max_key_length - 7
-        rows.map do |e|
-          value = e[1]
-          value = value.to_s.truncate(max_allowed_value_length) unless [true, false].include?(value)
-          [e[0], value]
+      def colorize_array(array, colors)
+        value = ""
+        array.each do  |l|
+          colored_line = l
+          colored_line = "#{colors.first[0]}#{l}#{colors.last[0]}" if colors.length > 0
+          value << colored_line
+          value << "\n"
         end
+        return value
+      end
+
+      def transform_output(rows, transform)
+        require 'fastlane_core/string_filters'
+        require 'tty-screen'
+
+        return_array = []
+        tcols = TTY::Screen.width
+
+        col_count = rows.map(&:length).first || 1
+
+        # -4 per column - as tt adds "| " and " |"
+        terminal_table_padding = 4
+        max_length = tcols - (col_count * terminal_table_padding)
+
+        max_value_length = (max_length / col_count)
+
+        rows.map do |row|
+          new_row = []
+          row.each do |col|
+            value = col.to_s.dup
+            if transform == :truncate_middle
+              value = value.middle_truncate(max_value_length)
+            elsif transform == :newline && value
+              # remove all fixed newlines as it may mess up the output
+              value.tr!("\n", " ") if value.kind_of?(String)
+              if value.length >= max_value_length
+                colors = value.scan(/(\e\[.*?m)/)
+                if colors && colors.length > 0
+                  colors.each do |color|
+                    value.delete!(color.first)
+                    value.delete!(color.last)
+                  end
+                end
+                lines = value.wordwrap(max_value_length)
+                value = colorize_array(lines, colors)
+              end
+            end
+            new_row << value
+          end
+          return_array << new_row
+          return_array << :separator
+        end
+        return_array.pop
+        return return_array
       end
 
       def collect_rows(options: nil, hide_keys: [], mask_keys: [], prefix: '', mask: '********')
--- a/fastlane_core/lib/fastlane_core/string_filters.rb
+++ b/fastlane_core/lib/fastlane_core/string_filters.rb
@@ -31,4 +31,20 @@ class String
 
     "#{self[0, stop]}#{omission}"
   end
+
+  # Base taken from: https://www.ruby-forum.com/topic/57805
+  def wordwrap(length = 80)
+    self.gsub!(/(\S{#{length}})(?=\S)/, '\1 ')
+    self.scan(/.{1,#{length}}(?:\s+|$)/)
+  end
+
+  # Base taken from: http://stackoverflow.com/a/12202205/1945875
+  def middle_truncate(length = 20, options = {})
+    omission = options[:omission] || '...'
+    return self if self.length <= length + omission.length
+    return self[0..length] if length < omission.length
+    len = (length - omission.length) / 2
+    s_len = len - length % 2
+    self[0..s_len] + omission + self[self.length - len..self.length]
+  end
 end
--- a/fastlane_core/spec/print_table_spec.rb
+++ b/fastlane_core/spec/print_table_spec.rb
@@ -48,15 +48,15 @@ describe FastlaneCore do
 
       value = FastlaneCore::PrintTable.print_values(config: @config, title: title, hide_keys: [:a_sensitive])
       expect(value[:title]).to eq(title.green)
-      expect(value[:rows]).to eq([['cert_name', "asdf"], ['output', '..'], ["a_bool", true]])
+      expect(value[:rows]).to eq([["cert_name", "asdf"], :separator, ["output", ".."], :separator, ["a_bool", "true"]])
     end
     it "automatically masks sensitive options" do
       value = FastlaneCore::PrintTable.print_values(config: @config)
-      expect(value[:rows]).to eq([["cert_name", "asdf"], ["output", ".."], ["a_bool", true], ["a_sensitive", "********"]])
+      expect(value[:rows]).to eq([["cert_name", "asdf"], :separator, ["output", ".."], :separator, ["a_bool", "true"], :separator, ["a_sensitive", "********"]])
     end
     it "supports mask_keys property with symbols and strings" do
       value = FastlaneCore::PrintTable.print_values(config: @config, mask_keys: [:cert_name, 'a_bool'])
-      expect(value[:rows]).to eq([["cert_name", "********"], ["output", ".."], ["a_bool", "********"], ["a_sensitive", "********"]])
+      expect(value[:rows]).to eq([["cert_name", "********"], :separator, ["output", ".."], :separator, ["a_bool", "********"], :separator, ["a_sensitive", "********"]])
     end
 
     it "supports hide_keys property with symbols and strings" do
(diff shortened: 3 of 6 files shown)


fastlane_core/lib/fastlane_core/string_filters.rb
```diff
@@ -31,4 +31,14 @@ def truncate(truncate_at, options = {})
 
     "#{self[0, stop]}#{omission}"
   end
+
+  # Base taken from: http://stackoverflow.com/a/12202205/1945875
+  def middle_truncate(length = 20, options = {})
```
KrauseFx: Is there a good way to detect the initial terminal width?

hjanuschka: @krausefx
this is just a generic function.
in the table printer itself - the terminal width is used

tcols = IO.console.winsize[1]
max_length = tcols

milch:
This is a really good idea, however, I would also like an option to disable this (e.g. when running on CI) so that users have at least an option to see the whole thing for debugging purposes.

hjanuschka:
@milch  ok done - 

```ruby
truncate = false if Helper.is_ci? || FastlaneCore::Env.truthy?("DO_NOT_TRUNCATE_TABLES")
```

sorry for the force push, as i had to rebase for the `truthy?` stuff.

milch:
> sorry for the force push

No worries, I haven't reviewed yet 👍 

> `FastlaneCore::Env.truthy?("DO_NOT_TRUNCATE_TABLES")`

This is a bit nitpicky but mostly we'd like to prefix ENV-vars. Can we call it `FL_NO_TRUNCATE_TABLES` (or something)?

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -18,7 +18,10 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
         rows = self.collect_rows(options: options, hide_keys: hide_keys.map(&:to_s), mask_keys: mask_keys.map(&:to_s), prefix: '')
 
         params = {}
-        params[:rows] = limit_row_size(rows)
+        truncate = false if Helper.is_ci? || FastlaneCore::Env.truthy?("DO_NOT_TRUNCATE_TABLES")
```
KrauseFx: Instead of `DO_NOT`, we've used `SKIP_...` until now

fastlane_core/lib/fastlane_core/string_filters.rb
KrauseFx: Ah yeah, cool 👍

KrauseFx:
This looks super promising, ideally it would also be used for the action table: https://github.com/fastlane/fastlane/issues/7811

hjanuschka:
@KrauseFx feedback addressed.
should i do the action table thing in this PR or in a seperate?

hjanuschka:
@KrauseFx ok i updated it to also work with newlines instead of the `..` stuff (defaulting to: `....`)


here is a sample to test:

```ruby

require "fastlane_core"



FastlaneCore::PrintTable.print_values(config:
  { wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
      wwwwwwwwwwwwwwwwwwwwwwwwxw: "ccc bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" },

                                   hide_keys: [],
                                       title: "Summary SUM")


FastlaneCore::PrintTable.print_values(config:
                                         { wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
                                             wwwwwwwwwwwwwwwwwwwwwwwwxw: "ccc bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbb bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" },

                                                                          hide_keys: [],
                                                                              title: "Summary SUM",
                                                                              transform: :newline)

```

bildschirmfoto 2017-01-11 um 20 54 17




and i also added support for this in the `actions_list.rb`:



```
fastlane actions
```

bildschirmfoto 2017-01-11 um 20 55 15



and in detailed action descr.:

```
fastlane action adb
```


bildschirmfoto 2017-01-11 um 20 56 22

jinjorge:
is there a way to make the line wraps happen at word boundaries?

from the screenshot above there are several words that are broken up from one line to the next

hjanuschka:
@jinjorge  good point, but i think there are cases where really long words would not be wrapped and resulting in a ugly table, you have a idea on how to solve that?

jinjorge:
@hjanuschka It's a numbers game - there are a lot more sub 5-6 letter words than really long words. Probably best to optimize for the majority and see how things fall out.

To you question on how to solve it - I don't know, though I will do some research.

hjanuschka:
test's pass now, ready for new review! @jinjorge  i am unsure on how to solve it without breaking the output.

jinjorge:
a regular expression match (`\b`) for word boundary.

hjanuschka:
@jinjorge  got it working !!!  respecting whole word, but if too long forcing to break it! ;)

bildschirmfoto 2017-01-24 um 08 03 29


```ruby
  # Base taken from: https://www.ruby-forum.com/topic/57805
  def wordwrap(length  = 80)
     self.gsub!(/(\S{#{length}})(?=\S)/, '\1 ')
     self.scan(/.{1,#{length}}(?:\s+|$)/)
  end
```

jinjorge:
👏 👏

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,61 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
+      end
+
+      def transform_output(rows, transform)
         require 'fastlane_core/string_filters'
+        require 'io/console'
+
+        return_array = []
+        tcols = IO.console.winsize[1]
+
+        col_count = rows.map(&:length).first || 1
```
milch: Why is 1 being used as default here?

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,61 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
         rows.map do |e|
-          value = e[1]
-          value = value.to_s.truncate(max_allowed_value_length) unless [true, false].include?(value)
-          [e[0], value]
+          new_row = []
+          e.each do |col|
+            value = col.to_s.dup
+            if transform == :truncate_middle
+              value = value.middle_truncate(max_value_length)
+            end
```
milch: Use `elsif` here to more clearly show intent

milch (dismissed):
Looks good to me aside from a couple minor things. I really think we should merge this soon

fastlane_core/lib/fastlane_core/print_table.rb
hjanuschka: i used to use 0 - but well we have always atleast one column, otherwise it crashes later on due to a division by zero 💃

fastlane_core/lib/fastlane_core/print_table.rb
hjanuschka: good point, going to adress this once back in private life 👍

hjanuschka: ahhh sorry to fast reply, i think there are cases where transform is `false`  (e.g: CI run)
so its neither :newline nor :truncate_middle - any other ideas?

hjanuschka:
@milch  feedback addressed, as we have the super awesome full-word split now i enabled `:newline` by default, and therefore the `elsif` worked out pretty good 👍

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -24,7 +24,14 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
         rows = self.collect_rows(options: options, hide_keys: hide_keys.map(&:to_s), mask_keys: mask_keys.map(&:to_s), prefix: '')
 
         params = {}
-        params[:rows] = limit_row_size(rows)
+        transform = false if FastlaneCore::Env.truthy?("SKIP_TRANSFORM_TABLES")
```
DanToml: `FL_SKIP_TABLE_TRANSFORM`?

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
+
+        # -4 per column - as tt adds "| " and " |"
+        terminal_table_padding = 4
+        max_length = tcols - (col_count * terminal_table_padding)
+
+        max_value_length = (max_length / col_count)
 
-        max_key_length = rows.map { |e| e[0].length }.max || 0
-        max_allowed_value_length = max_length - max_key_length - 7
         rows.map do |e|
```
DanToml: `rows.map do |row|`? - single letter variable made it pretty difficult to infer when scan reading

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
+        return_array = []
+        tcols = IO.console.winsize[1]
+
+        col_count = rows.map(&:length).first || 1
+
+        # -4 per column - as tt adds "| " and " |"
+        terminal_table_padding = 4
+        max_length = tcols - (col_count * terminal_table_padding)
+
+        max_value_length = (max_length / col_count)
```
DanToml: you don't need parens here

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
-        max_key_length = rows.map { |e| e[0].length }.max || 0
-        max_allowed_value_length = max_length - max_key_length - 7
         rows.map do |e|
-          value = e[1]
-          value = value.to_s.truncate(max_allowed_value_length) unless [true, false].include?(value)
-          [e[0], value]
+          new_row = []
+          e.each do |col|
+            value = col.to_s.dup
+            if transform == :truncate_middle
```
DanToml: would be nice to break this out into a transform_row/column function, because then appending to the row will be in a single place and be easier to maintain and test

i.e
```
def transform_row(transform, row)
  row.map do |column|
    transform_column(transform, column)
  end
end
```

```
def transform_column(transform, column)
  if transform == :truncate_middle
    ...
    value
  elseif transform == :newline
    ...
    value
end
```

endocrimes (dismissed):
I like where this is going :+1:

fastlane_core/lib/fastlane_core/print_table.rb
hjanuschka: done

fastlane_core/lib/fastlane_core/print_table.rb
hjanuschka: can we keep them?, i like it a bit more from point of readability.

fastlane_core/lib/fastlane_core/print_table.rb
hjanuschka: true point, fixed

fastlane_core/lib/fastlane_core/print_table.rb
hjanuschka: i am not 100% comfortable with this, as there is bunch a logic going on in :newline, colorization for example - lets keep it as it is, and try to finish this PR - as its a really long running one, and do the modularization in a seperate PR. (i just broke way too much with last minute quick-changes :sad:, last PRs)

hjanuschka (commented):
feedback adressed, master rebased and tty-screen used 👍

fastlane_core/lib/fastlane_core/print_table.rb
DanToml: I tend to only prefer them when there are multiple expressions to make precedence easier to read, but sure.

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -24,7 +24,14 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
         rows = self.collect_rows(options: options, hide_keys: hide_keys.map(&:to_s), mask_keys: mask_keys.map(&:to_s), prefix: '')
 
         params = {}
-        params[:rows] = limit_row_size(rows)
+        transform = false if FastlaneCore::Env.truthy?("FL_SKIP_TABLE_TRANSFORM")
```
DanToml: rather than overloading this, I'd maybe extract out a `transform?` method that is `!FastlaneCore::Env.truthy?...` and have the if statement be `if transform? ...` - this means you don't need to trace through mutating state.

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
+
+        # -4 per column - as tt adds "| " and " |"
+        terminal_table_padding = 4
+        max_length = tcols - (col_count * terminal_table_padding)
+
+        max_value_length = (max_length / col_count)
+
+        rows.map do |row|
+          new_row = []
+          row.each do |col|
```
DanToml: as we only return one value from this, we could do `new_row = row.map do |col|`

DanToml: might also be easier to read if we break the contents of this block into a `transform_column(column, truncation_style)` method?

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
         return params
       end
 
-      def limit_row_size(rows, max_length = 100)
+      def colorize_array(array, colors)
+        value = ""
```
giginet: you can write this logic more clearly.

```ruby
array.map { |l| "#{colors.first[0]}#{l}#{colors.last[0]}" unless colors.empty? }.join('\n')
```

giginet (dismissed):

fastlane_core/lib/fastlane_core/print_table.rb
mfurtak: I think this may not be equivalent, because it can create `nil` entries

mfurtak:
Just checking in on the status of this PR - can the feedback be addressed? Everyone seems pretty positive about getting it in! 👍

KrauseFx:
Improving all fastlane tables is definitely something we should fix, it's a big issue currently. @hjanuschka, is that something you want to finish up? 🚀

hjanuschka:
@KrauseFx yeah i want to see this in.
however i worked alot on this, and the open feedback, is are just nit-picks, and i am afraid i break everything to resolve the feedback (which as a personal point of view, is not mission critical in this case)

so, how about merging this, and later on the feedback suggesters go on with it?

hjanuschka:
@googlebot thx for this, i accidentally commited with corporate email - as soon as we know how to proceed with this PR i am going to rebase then it should be fine 👍

hjanuschka:
master rebased, cla stuff sorted out.

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
         return params
       end
 
-      def limit_row_size(rows, max_length = 100)
+      def colorize_array(array, colors)
```
KrauseFx: This method is never called, should it be part of this PR?

fastlane_core/lib/fastlane_core/print_table.rb
```diff
@@ -34,16 +41,60 @@ def print_values(config: nil, title: nil, hide_keys: [], mask_keys: [])
...
+                  end
+                end
+                lines = value.wordwrap(max_value_length)
+                value = colorize_array(lines, colors)
+              end
+            end
+            new_row << value
+          end
+          return_array << new_row
+          return_array << :separator
```
KrauseFx: Why do we have this here? This will result in a new line of dashes inbetween every row, making the table a lot bigger

KrauseFx:
screen shot 2017-04-06 at 7 06 57 pm

I think not having the separators for every row looks better 👍

fastlane_core/lib/fastlane_core/print_table.rb
KrauseFx: updated 👍

fastlane_core/lib/fastlane_core/print_table.rb
KrauseFx: Whops, never mind, it is used

fastlane_core/spec/print_table_spec.rb
```diff
@@ -48,15 +48,15 @@
 
       value = FastlaneCore::PrintTable.print_values(config: @config, title: title, hide_keys: [:a_sensitive])
       expect(value[:title]).to eq(title.green)
-      expect(value[:rows]).to eq([['cert_name', "asdf"], ['output', '..'], ["a_bool", true]])
+      expect(value[:rows]).to eq([["cert_name", "asdf"], :separator, ["output", ".."], :separator, ["a_bool", "true"]])
```
KrauseFx: It's clearly visible that the `separator` changes the existing behavior

KrauseFx:
Moved to https://github.com/fastlane/fastlane/pull/8809 (by mistake), I addressed all of your feedback, thanks. Also some more tests and code style changes.

PR #7398, opened 2016-12-09
[deliver] support app_icons via metadata

hjanuschka:
This one adds support for upload/download of app_icon/apple_watch_icon.

icons are stored int `metdata/app_icon.png` `metdata/watch_icon.png`

# Upload
as there is an option to explicit set a path to an app_icon, and to not break existing usage of this flags, the option will be used if set. otherwise if unset, and a file exists in the metdata folder root this file will be used

bildschirmfoto 2016-12-09 um 08 31 43

bildschirmfoto 2016-12-09 um 08 41 47

# Download
`download_metadata`  also downloads existing app icons.

bildschirmfoto 2016-12-09 um 12 23 49

> App icon redacted 👍

### DIFF
--- a/deliver/lib/assets/summary.html.erb
+++ b/deliver/lib/assets/summary.html.erb
@@ -107,7 +107,7 @@
               margin-left: 10px;
               margin-right: 10px;
           }
-          .app-icons img{
+          .app-icons img {
             width: 150px;
           }
           .app-icons .app-icon {
--- a/deliver/lib/assets/summary.html.erb
+++ b/deliver/lib/assets/summary.html.erb
@@ -107,10 +107,32 @@
               margin-left: 10px;
               margin-right: 10px;
           }
+          .app-icons img{
+            width: 150px;
+          }
+          .app-icons .app-icon {
+              float: left;
+              margin-right: 20px;
+          }
         </style>
     </head>
     
     <body>
+      <div class="app-icons">
+        
+        <% if @options[:app_icon] %>
+        <div class="app-icon">
+          Large App Icon:<br>
+          <img src="<%= @options[:app_icon] %>">
+        </div>
+        <% end %>
+        <% if @options[:apple_watch_app_icon] %>
+        <div class="watch-icon">
+          Watch App Icon:<br>
+          <img src="<%= @options[:apple_watch_app_icon] %>">
+        </div>
+        <% end %>
+      </div>
       <% @languages.each do |language| %>
         <% if @options[:name] %>
           <div class="app-name">
--- a/deliver/lib/deliver/runner.rb
+++ b/deliver/lib/deliver/runner.rb
@@ -52,6 +52,10 @@ module Deliver
       UploadMetadata.new.load_from_filesystem(options)
       UploadMetadata.new.assign_defaults(options)
 
+      # Handle app icon / watch icon
+
+      prepare_app_icons(options)
+
       # Validate
       validate_html(screenshots)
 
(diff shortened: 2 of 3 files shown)


deliver/lib/assets/summary.html.erb
```diff
@@ -107,10 +107,32 @@
               margin-left: 10px;
               margin-right: 10px;
           }
+          .app-icons img{
```
JustinJiaDev: I think you forget a space before `{`?

KrauseFx: 👍

deliver/lib/deliver/runner.rb
```diff
@@ -52,6 +52,10 @@ def upload_metadata
       UploadMetadata.new.load_from_filesystem(options)
       UploadMetadata.new.assign_defaults(options)
 
+      # Handle app icon / watch icon
+
```
KrauseFx: Please remove this space here

deliver/lib/deliver/setup.rb
```diff
@@ -67,6 +67,18 @@ def generate_metadata_files(v, path)
       end
 
       UI.success("Successfully created new configuration files.")
+
+      # get App icon + watch icon
+      if v.large_app_icon.asset_token
+        app_icon_path = File.join(path, "app_icon.png")
+        File.write(app_icon_path, open(v.large_app_icon.url).read)
```
KrauseFx: So if there is no app icon available yet, the `if` block before would detect that, right?

deliver/lib/deliver/runner.rb
```diff
@@ -62,6 +66,19 @@ def upload_metadata
       UploadAssets.new.upload(options) # e.g. app icon
     end
 
+    def prepare_app_icons(options = {})
```
KrauseFx: Could you add 1-2 lines of comment above this method on what it does

deliver/lib/deliver/runner.rb
```diff
@@ -62,6 +66,19 @@ def upload_metadata
...
+      app_icon_path = options[:app_icon] if options[:app_icon]
+      app_icon_metadata = File.join(options[:metadata_path], "app_icon.png")
```
KrauseFx: Will this crash if `metadata_path` isn't set? Is `metadata_path` optional? It seems to be set in `detect_values.rb`

deliver/lib/deliver/runner.rb
```diff
@@ -62,6 +66,19 @@ def upload_metadata
...
+      if app_icon_path.nil? && File.exist?(app_icon_metadata)
+        options[:app_icon] = app_icon_metadata
+      end
+      watch_icon_path = options[:apple_watch_app_icon] if options[:apple_watch_app_icon]
```
KrauseFx: Both code blocks seem overly complicated, can we use

```ruby
default_app_icon_path = File.join(options[:metadata_path], "app_icon.png")
options[:app_icon] ||= default_app_icon_path if File.exist?(default_app_icon_path)
```
This should have the same effect, right?

KrauseFx: Same for watch

KrauseFx (changes requested):

deliver/lib/assets/summary.html.erb
hjanuschka: done - thx!

deliver/lib/deliver/runner.rb
hjanuschka: done

deliver/lib/deliver/runner.rb
hjanuschka: done

deliver/lib/deliver/runner.rb
hjanuschka: in fact it is optional, but in `detect_values.rb`

`options[:metadata_path] ||= File.join(containing, 'metadata')` this one is called everytime so it shouldn't be nil, i guess.

i added a return if not set - just in case.

deliver/lib/deliver/runner.rb
hjanuschka: awesome ! thx - stil impressed by ruby! 👍

deliver/lib/deliver/setup.rb
hjanuschka: yep tested on one of my apps which has no icon - and `v.large_app_icon.asset_token`  is nil

hjanuschka (commented):
thx guys @KrauseFx @JustinJiaDev 
feedback addressed.

KrauseFx (approved):

Issue #7366, opened 2016-12-07
deliver doesn't download iMessage screenshots into `iMessage` subdirectory


hborders:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I just ran `deliver download_screenshots`, and it dutifully downloaded all my app's screenshots. However, it didn't put iMessage screenshots into the `./screenshots/iMessage` directory [as is required for iMessage screenshot uploading](https://github.com/fastlane/fastlane/pull/6830). Instead all of my screenshots are mixed together in their respective localization folders.

hborders:
<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                 | Value                                                          |
| ------------------- | -------------------------------------------------------------- |
| OS                  | 10.11.6                                                        |
| Ruby                | 2.0.0                                                          |
| Bundler?            | false                                                          |
| Git                 | git version 2.8.3                                              |
| Installation Source | /usr/local/bin/fastlane                                        |
| Host                | Mac OS X 10.11.6 (15G1108)                                     |
| Ruby Lib Dir        | /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib |
| OpenSSL Version     | OpenSSL 0.9.8zc 19 Mar 2015                                    |
| Is contained        | false                                                          |
| Xcode Path          | /Applications/Xcode-8.1.app/Contents/Developer/                |
| Xcode Version       | 8.1                                                            |


### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE | en_US.UTF-8 | ✅ |


### fastlane files:

**No Fastfile found**


**No Appfile found**


### fastlane gems

| Gem                 | Version | Update-Status       |
| ------------------- | ------- | ------------------- |
| credentials_manager | 0.16.2  | ✅ Up-To-Date        |
| spaceship           | 0.38.5  | ✅ Up-To-Date        |
| fastlane_core       | 0.58.0  | ✅ Up-To-Date        |
| deliver             | 1.16.1  | ✅ Up-To-Date        |
| snapshot            | 1.16.4  | ✅ Up-To-Date        |
[...]

PR #7399, opened 2016-12-09
[spaceship] identify imessage screenshots

hjanuschka:
# PR 1/2

issue: https://github.com/fastlane/fastlane/issues/7366

# Spaceship
 * adds a attr `platform` - sets it to `imessage` if screenshot is imessage.

# deliver
 * download imessage screens into the right folder `iMessage`
 
 
 
  * 1/2 -> https://github.com/fastlane/fastlane/pull/7399 (spaceship)
  * 2/2 -> https://github.com/fastlane/fastlane/pull/7400 (deliver)

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -703,7 +703,7 @@ module Spaceship
             data = {
                 device_type: display_family['name'],
                 language: row["language"],
-                platform: "imessage" # to identify imessage screenshots later on (e.g: during download)
+                is_imessage: true # to identify imessage screenshots later on (e.g: during download)
             }.merge(screenshot_data)
             result << Tunes::AppScreenshot.factory(data)
           end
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/tunes/app_version.rb
```diff
@@ -702,7 +702,8 @@ def setup_messages_screenshots_for(row)
             screenshot_data = screenshot["value"]
             data = {
                 device_type: display_family['name'],
-                language: row["language"]
+                language: row["language"],
+                platform: "imessage" # to identify imessage screenshots later on (e.g: during download)
```
milch: As far as I can tell from this and the other PR, the platform is only ever set to `"imessage"`. In that case I'd prefer to have a boolean flag, e.g. `is_imessage?`

milch (changes requested):

hjanuschka:
@milch  updated

milch (approved):

PR #7406, opened 2016-12-09
Fix xcodebuild action specs

mfurtak:
As of 2016-12-7, CircleCI has started setting the `FL_BUILDLOG_PATH` environment variable, which changes the output location for build logs.

Many specs were assuming that the generated commands would include the default log path, rather than using the `buildlog_path` helper function that takes into account the environment variable.

This fixes those specs to use the helper and get the right path.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,6 @@
 machine:
+  environment:
+    FL_BUILDLOG_PATH: ~/Library/Logs
   xcode:
     version: "7.3"
 dependencies:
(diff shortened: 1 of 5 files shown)


ohayon (approved):
💯

DanToml:
@mfurtak Looks good to me as a change.

One potentially unintended side effect is that you'll now be sending the test test logs to CI artefacts tho.

mfurtak:
@DanToml Because this path that Circle sets is collected as artifacts automatically? Would you suggest that we set `FL_BUILDLOG_PATH` for our project env vars to be somewhere else if we don't want this?

DanToml:
@mfurtak Yeah. We pointed it at artifacts to help with debugging builds. - For the case of fastlane/fastlane, you probably want to override it in your circle.yml

endocrimes (approved):
🚢

mfurtak:
Thanks for the review!

PR #7410, opened 2016-12-09
[fastlane/action] get_version_number, improve target detection/notice doc.

hjanuschka:
as discussed here https://github.com/fastlane/fastlane/issues/7361
this PR try's to improve the output once a specified target - is not found, by listing all available targets (as those are not 100% always the same names as in xcode target)

e.g.: Xcode Target: "DemoAs" Plist: "DemoA-Info.plist"

i think it happens once you rename a target in xcode! as the plist name does not change.

bildschirmfoto 2016-12-09 um 21 20 16

### DIFF
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -104,7 +104,7 @@ module Fastlane
         targets_available = self.parse_targets(results)
         if params[:target] && !targets_available.include?(params[:target])
           UI.important("Could not find specified target: #{params[:target]}")
-          UI.important("Availaible targets:")
+          UI.important("Available targets:")
           targets_available.each do |t|
             UI.important("\t #{t}")
           end
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -80,11 +80,36 @@ module Fastlane
         version_number = line.partition('=').last
         return version_number if Helper.is_test?
 
+        targets_available = self.parse_targets(results)
+        self.validate_target(params, targets_available)
+
         # Store the number in the shared hash
         Actions.lane_context[SharedValues::VERSION_NUMBER] = version_number
       rescue => ex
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -80,13 +80,38 @@ def self.run(params)
         version_number = line.partition('=').last
         return version_number if Helper.is_test?
 
+        targets_available = self.parse_targets(results)
+        self.validate_target(params, targets_available)
```
ohwutup: Can `self.validate_target` be renamed to `self.validate_targets` and it we could just pass it the `results` and it could handle calling `parse_targets`? Just to condense this down into one line that is validating targets.

ohayon (changes requested):
Hey @hjanuschka, this looks great, just thought we might be able to abstract it one tiny level further with the one comment I left.

Totally up for debate so if you disagree, feel free to let me know and we can merge away! 🚀

hjanuschka:
@ohwutup thx feedback adressed.

fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -80,13 +80,38 @@ def self.run(params)
...
+          end
+        end
+        return targets_available
+      end
+
+      def self.validate_targets(params, results)
+        targets_available = self.parse_targets(results)
+        if params[:target] && !targets_available.include?(params[:target])
+          UI.important("Could not find specified target: #{params[:target]}")
+          UI.important("Availaible targets:")
```
mfurtak: Availaible -> Available

KrauseFx:
Should the same method be applied to `get_build_number` and other similar methods?

hjanuschka:
@mfurtak  thx for the hint, my english is really bad - it even sucks more than my native-writing skills (and those are really bad 😒)

@KrauseFx  i ll have a look - at other `target` related places, and maybe its worth moving it into fastlane core.

mfurtak:
This approach doesn't seem to have support from the original reporters, so I'm going to close this for now. We can always reopen/revisit if needed 👍

PR #7414, opened 2016-12-10
[spaceship] Automatically re-try “UnauthorizedAccessError” errors

KrauseFx:
Those errors would randomly occur sometimes due to a server error. The full stack trace when this happens:

```
**Unauthorized Access**
/app/local_fastlane/spaceship/lib/spaceship/client.rb:469:in `block in send_request'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:362:in `with_retry'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:463:in `send_request'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:400:in `request'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:306:in `send_shared_login_request'
/app/local_fastlane/spaceship/lib/spaceship/tunes/tunes_client.rb:164:in `send_login_request'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:430:in `do_login'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:245:in `login'
/app/local_fastlane/spaceship/lib/spaceship/client.rb:92:in `login'
/app/local_fastlane/spaceship/lib/spaceship/tunes/spaceship.rb:22:in `login'
```

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -360,7 +360,7 @@ module Spaceship
 
     def with_retry(tries = 5, &_block)
       return yield
-    rescue Faraday::Error::ConnectionFailed, Faraday::Error::TimeoutError, AppleTimeoutError, UnauthorizedAccessError, Errno::EPIPE => ex # New Faraday version: Faraday::TimeoutError => ex
+    rescue Faraday::Error::ConnectionFailed, Faraday::Error::TimeoutError, AppleTimeoutError, Errno::EPIPE => ex # New Faraday version: Faraday::TimeoutError => ex
       unless (tries -= 1).zero?
         logger.warn("Timeout received: '#{ex.message}'. Retrying after 3 seconds (remaining: #{tries})...")
         sleep 3 unless defined? SpecHelper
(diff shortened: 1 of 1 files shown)


KrauseFx:
Whops, found that we now catch the exception twice

```ruby
    rescue UnauthorizedAccessError => ex
      if @loggedin && !(tries -= 1).zero?
        msg = "Auth error received: '#{ex.message}'. Login in again then retrying after 3 seconds (remaining: #{tries})..."
        puts msg if $verbose
        logger.warn msg
        do_login(self.user, @password)
        sleep 3 unless defined? SpecHelper
        retry
      end
```

looking into it

KrauseFx:
since the original stack trace shows that we also get those error message on the initial login, we should re-login at that point as well.

snatchev:
@KrauseFx I would like to add a test-case for this behavior. Do you know what the response (or response code) is when the server error occurs?

snatchev:
I am unclear on how this change will resolve the symptoms. According to 
https://github.com/fastlane/fastlane/blob/059a289b8bdeb2981277b3f67d46fe81c4458c66/spaceship/lib/spaceship/client.rb#L471-L474

The only time we raise `UnauthorizedAccessError` is when the response is a 401. Are we seeing that the service responds with a 401 status, but it is a server error?

snatchev (changes requested):
I would be more confident about merging this if there was a test-case that shows how this changes fixes the problem. Would be happy to collaborate on this.

KrauseFx:
Actually this issue isn't relevant any more 👍

PR #7422, opened 2016-12-11
[fastlane] ENV class

hjanuschka:
# PR 2/7

Split PR's of initial PR -> https://github.com/fastlane/fastlane/pull/7337
 original discussed here: #7326

# Series
  * 1/7 #7421 (fastlane_env_class_fastlane_core)
  * 2/7 #7422 (fastlane_env_class_fastlane)
  * 3/7 #7423 (fastlane_env_class_gym)
  * 4/7 #7424 (fastlane_env_class_match)
  * 5/7 #7425 (fastlane_env_class_scan)
  * 6/7 #7426 (fastlane_env_class_snapshot)
  * 7/7 #7427 (fastlane_env_class_supply)

 # All in One 
  * https://github.com/fastlane/fastlane/compare/master...hjanuschka:fastlane_env_class

__Generated by fastlane-toolbelt__

PR #7421, opened 2016-12-11
[fastlane_core] Env Class

hjanuschka:
# PR 1/7

Split PR's of initial PR -> https://github.com/fastlane/fastlane/pull/7337
 original discussed here: #7326

# Series
  * 1/7 #7421 (fastlane_env_class_fastlane_core)
  * 2/7 #7422 (fastlane_env_class_fastlane)
  * 3/7 #7423 (fastlane_env_class_gym)
  * 4/7 #7424 (fastlane_env_class_match)
  * 5/7 #7425 (fastlane_env_class_scan)
  * 6/7 #7426 (fastlane_env_class_snapshot)
  * 7/7 #7427 (fastlane_env_class_supply)

 # All in One 
  * https://github.com/fastlane/fastlane/compare/master...hjanuschka:fastlane_env_class

__Generated by fastlane-toolbelt__

PR #7424, opened 2016-12-11
[match] ENV class

hjanuschka:
# PR 4/7

Split PR's of initial PR -> https://github.com/fastlane/fastlane/pull/7337
 original discussed here: #7326

# Series
  * 1/7 #7421 (fastlane_env_class_fastlane_core)
  * 2/7 #7422 (fastlane_env_class_fastlane)
  * 3/7 #7423 (fastlane_env_class_gym)
  * 4/7 #7424 (fastlane_env_class_match)
  * 5/7 #7425 (fastlane_env_class_scan)
  * 6/7 #7426 (fastlane_env_class_snapshot)
  * 7/7 #7427 (fastlane_env_class_supply)

 # All in One 
  * https://github.com/fastlane/fastlane/compare/master...hjanuschka:fastlane_env_class

__Generated by fastlane-toolbelt__

### DIFF
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -52,7 +52,7 @@ module Match
              encrypt: false)
         rescue
           UI.error "Couldn't decrypt the repo, please make sure you enter the right password!"
-          UI.user_error!("Invalid password passed via 'MATCH_PASSWORD'") if ENV["MATCH_PASSWORD"]
+          UI.user_error!("Invalid password passed via 'MATCH_PASSWORD'") if FastlaneCore::Env.truthy?("MATCH_PASSWORD")
           clear_password(git_url)
           decrypt_repo(path: path, git_url: git_url)
           return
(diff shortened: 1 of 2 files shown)


match/lib/match/encrypt.rb
```diff
@@ -52,7 +52,7 @@ def decrypt_repo(path: nil, git_url: nil, manual_password: nil)
              encrypt: false)
         rescue
           UI.error "Couldn't decrypt the repo, please make sure you enter the right password!"
-          UI.user_error!("Invalid password passed via 'MATCH_PASSWORD'") if ENV["MATCH_PASSWORD"]
+          UI.user_error!("Invalid password passed via 'MATCH_PASSWORD'") if FastlaneCore::Env.truthy?("MATCH_PASSWORD")
```
milch: Haha, this is an interesting one ... If the user passes a password like `"false"` this will not be called. I think keeping it how it was is fine in that case (or change it to `unless ENV["MATCH_PASSWORD"].nil?` to make it more explicit)

milch (changes requested):

match/lib/match/encrypt.rb
milch: That's also why the tests fail, I think.

hjanuschka:
@milch thx! feedback addressed

milch (approved):

Issue #7412, opened 2016-12-09
Change fastlane to be a mono gem


KrauseFx:
We want fastlane to be just one gem instead of 10. People would install _fastlane_, and get all the other tools with that. We want to deprecate the other tools as standalone gems and make them part of fastlane

**Why?**

- Make releases much simpler, just one gem to ship
- No more internal dependency updates
- No more version conflicts if we add some methods to some sub-gems
- 1 PR could modify multiple tools and tests are still passing (doesn’t mean we have to, but we can for example to update _spaceship_ and _deliver_ at the same time if necessary)

Are there any concerns or disadvantages we should consider?

hjanuschka:
i really like this idea - no more PR splitting!!!

KrauseFx:
> we found some issues with having spaceship as standalone gem, as this would also require _credentials_manager_ as standalone gem
> this way, we’d have some sub-folders being there own gems, and some not
making the whole setup much more complex
> we looked at libraries.io, the only big open source project using spaceship directly is xcode-install, which is owned by me, and could be part of the fastlane mono repo
> let us know if you have any concerns around this

hjanuschka:
so for people who wan't to use spacehip only they would have to install fastlane gem, and then `require "fastlane/spaceship"` ?

KrauseFx:
@hjanuschka they can just `require 'spaceship'` and it just works ™ but yeah, they need to _fastlane_ as dependency

supermarin:
Disadvantages - we can't use just pilot now. we need to add 50 gems only to use pilot.
I believe there's better way to pack gems rather than merging all into one

supermarin:
To elaborate, this adds further potential gem collisions if users have been using different versions of gems `fastlane` is depending on

ikonst:
We woke up to a morning of breaking CI, since:
a) `credentials_manager-1.6.0` and `credentials_manager-1.6.2` were yanked out of Rubygems
b) we were using `pilot-1.13.1` which depends on `credentials_manager>=1.6.0`

All Pilot users are now forced to either downgrade their Pilot to 1.10.0 (all versions above are unusable due to missing deps), or install Fastlane 2.0.

fabb:
We always upgrade to the newest gym/pilot/deliver versions in our CI script. Packing all into one gem increases the unnecessary dependencies a lot. Isn't there a better way to install only the necessary tools? Something like subspecs or as plugins to fastlane core?

[...]

PR #7436, opened 2016-12-12
Migrate fastlane to be a mono gem

KrauseFx:
Implements https://github.com/fastlane/fastlane/issues/7412


fastlane/lib/fastlane.rb
```diff
@@ -26,7 +26,6 @@ module Fastlane
   Helper = FastlaneCore::Helper # you gotta love Ruby: Helper.* should use the Helper class contained in FastlaneCore
   UI = FastlaneCore::UI
   ROOT = Pathname.new(File.expand_path('../..', __FILE__))
-  DESCRIPTION = "The easiest way to automate beta deployments and releases for your iOS and Android apps"
```
KrauseFx: Let's leave this in `fastlane.rb`, it makes more sense to be there, as it's not version specific 👍

milch (approved):
🦄🚀🤖

PR #7452, opened 2016-12-12
Update gemspecs for fastlane 2.0

KrauseFx:
Note: To merge this we actually need to release fastlane 2.0 first so that we can make a final 1.x release for each tool that is dependent on fastlane 2.0


milch:
Note: To merge this we actually need to release `fastlane 2.0` first so that we can make a final 1.x release for each tool that is dependant on fastlane 2.0

milch (approved):
💯

PR #7500, opened 2016-12-15
Allow android platform for testfairy action

stevenyap:
Testfairy API for both iOS and Android is exactly the same so the same code can be used for Android uploading too. This PR added android to the supported platform. This is tested on my testfairy account and I am able to upload an Android APK successfully.

### DIFF
--- a/fastlane/lib/fastlane/actions/testfairy.rb
+++ b/fastlane/lib/fastlane/actions/testfairy.rb
@@ -136,7 +136,7 @@ module Fastlane
                                        end),
           FastlaneCore::ConfigItem.new(key: :ipa,
                                        env_name: 'TESTFAIRY_IPA_PATH',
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/testfairy.rb
```diff
@@ -136,7 +136,7 @@ def self.available_options
                                        end),
           FastlaneCore::ConfigItem.new(key: :ipa,
                                        env_name: 'TESTFAIRY_IPA_PATH',
-                                       description: 'Path to your IPA file. Optional if you use the _gym_ or _xcodebuild_ action',
+                                       description: 'Path to your IPA file for iOS or APK for Android',
```
KrauseFx: Does that work? Have you tried it locally?

JAGerMANY:
A good news! I'll try it

KrauseFx (approved):

Issue #7519, opened 2016-12-16
Circular dependency between fastlane and deliver.


shammatime:
We started getting this error earlier today and am not sure how to solve the problem.

The command run 'bundle exec update'

Output:
Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem 'deliver' or gem 'fastlane' and try again.

Our Gemfile :
source "https://rubygems.org"

gem 'fastlane' , '1.111.0'
gem 'curb'
gem 'rubyzip' , '1.1.7'
gem 'cocoapods' , '1.1.1'

We do not want to update to 2.0.2. Any help would be great.

milch:
@shammatime You should upgrade to 2.0.2 to fix this issue or you won't get future updates for version 1. There are minimal breaking changes for our users (other than how the gem and internal dependencies are structured), so the change should be seamless. 

Barring that you would need to pin the deliver version in your Gemfile to `1.16.2`, which was the last version that supported fastlane `1.x`.

Let me know if you have any further problems.

KrauseFx:
Is there any specific reason why you don't want to update to 2.0?

shammatime:
We tried 2.0.2 and having issues with gym in 2.0.2. Here is our error:
bundler: failed to load command: fastlane (/private/tmp/workspace_build/workspace/MOBILE_READER_IOS_BUILD/I/IosReader/vendor/bundle/ruby/2.3.0/bin/fastlane)
20:07:05 Zip::Error: [31m[!] File /var/folders/q8/ntxbg8dn2cv3jmdrbxbd6w3c0000gq/T/gym_output20161215-30066-hkixcp/IonicIosSdk.ipa not found[0m
20:07:05   /private/tmp/workspace_build/workspace/MOBILE_READER_IOS_BUILD/I/IosReader/vendor/bundle/ruby/2.3.0/gems/rubyzip-1.2.0/lib/zip/file.rb:85:in `initialize'
20:07:05   /private/tmp/workspace_build/workspace/MOBILE_READER_IOS_BUILD/I/IosReader/vendor/bundle/ruby/2.3.0/gems/rubyzip-1.2.0/lib/zip/file.rb:99:in `new'
20:07:05   /private/tmp/workspace_build/workspace/MOBILE_READER_IOS_BUILD/I/IosReader/vendor/bundle/ruby/2.3.0/gems/rubyzip-1.2.0/lib/zip/file.rb:99:in `open'
[...]

PR #7525, opened 2016-12-16
Losen up rubyzip dependency

KrauseFx:


mfurtak:
Can you say a bit about why this is needed?

KrauseFx:
We've had multiple reports around version conflicts with rubyzip with other libarires:

e.g. https://github.com/fastlane/fastlane/issues/7519

This is due to external fastlane plugins, I submitted PRs to them, but this might help. We don't actually have any specific rubyzip version requirements, so it's better to have it loser I think

mfurtak:
Sounds good. Just wanted to get some context so we can remember why we did this later :smile: 👍

mfurtak (approved):

milch (approved):

PR #7527, opened 2016-12-16
Disable auto-install and show message when user uses bundler and missing gem

KrauseFx:


KrauseFx:
Before that, it would automatically install the gem, which doesn't make sense for gems

mfurtak (approved):

PR #7528, opened 2016-12-16
Add fastlane `rake push` task to release new fastlane releases

KrauseFx:


mfurtak (approved):
The title is a little misleading, since the rake task appears to be called `push`

KrauseFx:
nice catch, updated

PR #7546, opened 2016-12-16
[fastlane] more tests for aliases

hjanuschka:
test for https://github.com/fastlane/fastlane/pull/7543 

as reported in: #7542

### DIFF
--- a/fastlane/spec/action_spec.rb
+++ b/fastlane/spec/action_spec.rb
@@ -44,6 +44,14 @@ describe Fastlane do
           somealias_no_param('PARAM')
         end").runner.execute(:test)
       end
+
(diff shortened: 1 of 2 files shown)


fastlane/spec/action_spec.rb
```diff
@@ -44,6 +44,14 @@
           somealias_no_param('PARAM')
         end").runner.execute(:test)
       end
+
+      it "alias does not crash - when 'alias_used' not defined", now: true do
```
KrauseFx: `now: true`

hjanuschka:
@KrauseFx thx - done - i hope i ll not forget it in the future.

KrauseFx (approved):

Issue #7535, opened 2016-12-16
set_info_plist_value should fail build on error


zantoku:
We're setting some Info.plist values using set_info_plist_value(). Due to a typo in the Fastfile, set_info_plist_value() ended up being called with a key, but a null value, like this:

```
set_info_plist_value(
    path: options[:plist],
    key: 'FacebookAppID',
    value: options[:facebook_app_id]
)
```

It's structured like this because we need to set different FB AppIDs for different build variants. The build proceeded with this warning message:

```
[11:41:54]: ----------------------------------
[11:41:54]: --- Step: set_info_plist_value ---
[11:41:54]: ----------------------------------
[11:41:54]: No value found for 'value'
[11:41:54]: Unable to set value to plist file at '/jenkins/workspace/[redacted]/Info.plist'

```

However, we missed the warning message, and the build went through and was marked with SUCCESS. Unfortunately, we ended up with a non-functioning/crashing FB login on the App Store. :-(

I believe that a failed set_info_plist_value should fail the build.

ENVIRONMENT

We're not super up to date but I haven't found any reported issues related to this, so I expect current behaviour is the same.

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                 | Value                                              |
| ------------------- | -------------------------------------------------- |
| OS                  | 10.11.6                                            |
| Ruby                | 2.3.1                                              |
| Bundler?            | true                                               |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/        |
| Xcode Version       | 8.1                                                |
| Git                 | git version 2.9.3 (Apple Git-75)                   |
| Installation Source | /Users/[redacted]/.rbenv/versions/2.3.1/bin/fastlane |
| Host                | Mac OS X 10.11.6 (15G1108)                         |
| Ruby Lib Dir        | /Users/[redacted]/.rbenv/versions/2.3.1/lib          |
| OpenSSL Version     | OpenSSL 1.0.2j  26 Sep 2016                        |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
fastlane_version "1.39.0"

default_platform :ios

[content removed, contains sensitive info]

```
</details>

<details><summary>`./fastlane/Appfile`</summary>

```ruby
app_identifier "" # The bundle identifier of your app
apple_id "[redacted]" # Your Apple email address
```
</details>

### fastlane gems

| Gem                 | Version | Update-Status       |
| ------------------- | ------- | ------------------- |
[...]

PR #7547, opened 2016-12-16
[fastlane,actions] set_info_plist_value - user_error if plist error

hjanuschka:
as reported here: https://github.com/fastlane/fastlane/issues/7535

i think if someone wan'ts to silently fail the action - it should be done in fastfile.


KrauseFx (approved):

KrauseFx:
Nice catch, and great fix 👍

PR #7559, opened 2016-12-17
Not all specs are enabled since the move to monogem

lacostej:
This reveals lots of failure. Not yet looked into.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,7 +15,7 @@ task :rubygems_admins do
 end
 
 task :test_all do
-  sh "rspec --pattern */spec/*_spec.rb"
+  sh "rspec --pattern ./**/*_spec.rb"
 end
 
 # Overwrite the default rake task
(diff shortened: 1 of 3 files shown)


KrauseFx:
Oh wow, nice catch. This needs also to be updated in `circle.yml`, so that the CI build fails as well. I see there are 40 failures with that change.

KrauseFx:
Oh and this would only run 998 tests, that's not all

KrauseFx:
```rspec --pattern ./**/*_spec.rb```
this seems to be the pattern that would include all the tests 👍

KrauseFx:
Please update the pattern at both location to the one I posted, and merge https://github.com/lacostej/fastlane/pull/4 into your branch 👍 This will make all 3394 tests green 💚

KrauseFx (approved):

KrauseFx:
Thanks for updating the PR 👍

PR #7560, opened 2016-12-17
Logging is disabled while running tests. This makes it harder than necessary to disable.

lacostej:
Not really sure why, so I add this environment variable

### DIFF
--- a/ToolsAndDebugging.md
+++ b/ToolsAndDebugging.md
@@ -19,10 +19,10 @@ As debugging with pry requires the development dependencies, make sure to execut
 bundle exec fastlane beta --verbose
 ```
 
-If you need the breakpoint when running tests, make sure to have the `DEBUG` mode enabled, as the default test runner will remove all output from stdout, and therefore not showing the output of `pry`:
+If you need the breakpoint when running tests, make sure to have the `DEBUG` environment variable set, as the default test runner will remove all output from stdout, and therefore not showing the output of `pry`:
 
 ```
-DEBUG=1 bundle exec rspec
+DEBUG= bundle exec rspec
 ```
 
 You will then jump into an interactive debugger that allows you to print out variables, call methods and [much more](https://github.com/pry/pry/wiki).
--- a/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
+++ b/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
@@ -7,7 +7,8 @@ module FastlaneCore
 
       $stdout.sync = true
 
-      if Helper.is_test?
+      if Helper.is_test? && !ENV.key?('DEBUG')
+        $stdout.puts "Logging disabled while running tests. Force them by setting the DEBUG environment variable"
         @log ||= Logger.new(nil) # don't show any logs when running tests
       else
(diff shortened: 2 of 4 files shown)


KrauseFx:
I've used `DEBUG=1 bundle exec rspec` until now. Can you try if that works for you and let's use that in the documentation instead? I like the message 👍

lacostej:
I tested and it works. It's not perfect though:

* The message is printed to the standard $stdout, which might be /tmp/spaceship_tests
* there are 2 places with the following code:
```
unless ENV["DEBUG"]
  $stdout = File.open("/tmp/spaceship_tests", "w")
end
```

and the above code doesn't say what it does. It might be better to make it more explicit if we also added a message here before overriding $stdout. That way the user knows what is happening. Hence my second commit.

spaceship/spec/spec_helper.rb
```diff
@@ -6,6 +6,7 @@
 require_relative 'du/du_stubbing'
 
 unless ENV["DEBUG"]
```
KrauseFx: Wait, this whole block should actually be removed from the spaceship spec helper. I don't have commit access to your fork, can you update the PR?

fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
```diff
@@ -7,7 +7,8 @@ def log
 
       $stdout.sync = true
 
-      if Helper.is_test?
+      if Helper.is_test? && !ENV.key?('DEBUG')
+        $stdout.puts "Logging disabled while running tests. Force them by setting the DEBUG mode"
```
KrauseFx: `Force them by setting the `DEBUG` environment variable`

KrauseFx (dismissed):

fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
lacostej: I had used that wording because of the one I found in the doc. I changed it, and changed the doc.

spaceship/spec/spec_helper.rb
lacostej: sure

lacostej:
Note that I renamed the output file to `/tmp/fastlane_tests`

ToolsAndDebugging.md
```diff
@@ -19,10 +19,10 @@ As debugging with pry requires the development dependencies, make sure to execut
 bundle exec fastlane beta --verbose
 ```
 
-If you need the breakpoint when running tests, make sure to have the `DEBUG` mode enabled, as the default test runner will remove all output from stdout, and therefore not showing the output of `pry`:
+If you need the breakpoint when running tests, make sure to have the `DEBUG` environment variable set, as the default test runner will remove all output from stdout, and therefore not showing the output of `pry`:
 
 ```
-DEBUG=1 bundle exec rspec
+DEBUG= bundle exec rspec
```
KrauseFx: Why no more 1?

milch:
Seems like everything is clear now. Thanks for your contribution, @lacostej 🚀

milch (approved):

Issue #7513, opened 2016-12-15
match asks password even in CI mode


lacostej:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I tried running a fastlane job that uses match for the first time on a new build slave.

The job went into a loop asking for the password, even if the build was not attended.

I suspect we lack a `UI.interactive?` here

https://github.com/fastlane/fastlane/blob/485199c8757113f2fdd1eee3448a0e0f3cd6d4e7/match/lib/match/encrypt.rb#L17-L23

or 

https://github.com/fastlane/fastlane/blob/485199c8757113f2fdd1eee3448a0e0f3cd6d4e7/match/lib/match/change_password.rb#L16-L27

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
[21:52:52]: Couldn't decrypt the repo, please make sure you enter the right password!
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
[21:52:52]: Enter the passphrase that should be used to encrypt/decrypt your certificates
[21:52:52]: This passphrase is specific per repository and will be stored in your local keychain
[21:52:52]: Make sure to remember the password, as you'll need it when you run match on a different machine
[21:52:52]: Passphrase for Git Repo: 
[21:52:53]: Couldn't decrypt the repo, please make sure you enter the right password!
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
[21:52:53]: Enter the passphrase that should be used to encrypt/decrypt your certificates
[21:52:53]: This passphrase is specific per repository and will be stored in your local keychain
[21:52:53]: Make sure to remember the password, as you'll need it when you run match on a different machine
[21:52:53]: Passphrase for Git Repo: 
[21:52:53]: Couldn't decrypt the repo, please make sure you enter the right password!
```
### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

N/A

hjanuschka:
@lacostej  did you set the env `MATCH_PASSWORD`?

lacostej:
@hjanuschka no I didn't. But the code shouldn't go into a loop when the build is `!UI.interactive?`

It should do

```
search in MATCH_PASSWORD
[...]

PR #7561, opened 2016-12-17
#7513 fixes non-interactive shell password_ask loop in match

lacostej:
I have 2 questions / potential improvements

* we could replace the `UI.error "No password found neither in environment nor in local keychain. Bailing out as in non interactive mode."` with a `UI.user_error!`. If not the current code path relies on the caller to do something with it. But I am not fond off `user_error` being so down the call stack. I don't like returning `nil` either. Preference?

* we could use `UI.crash!` instead of `raise`. Not sure what's the standard when finding invalid API use

### DIFF
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -15,11 +15,15 @@ module Match
       end
 
       unless password
-        UI.important "Enter the passphrase that should be used to encrypt/decrypt your certificates"
-        UI.important "This passphrase is specific per repository and will be stored in your local keychain"
-        UI.important "Make sure to remember the password, as you'll need it when you run match on a different machine"
-        password = ChangePassword.ask_password(confirm: true)
-        store_password(git_url, password)
+        if !UI.interactive?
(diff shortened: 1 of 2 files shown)


asfalcone (approved):

asfalcone:
Thanks @lacostej 💯

match/lib/match/encrypt.rb
```diff
@@ -15,11 +15,15 @@ def password(git_url)
       end
 
       unless password
-        UI.important "Enter the passphrase that should be used to encrypt/decrypt your certificates"
-        UI.important "This passphrase is specific per repository and will be stored in your local keychain"
-        UI.important "Make sure to remember the password, as you'll need it when you run match on a different machine"
-        password = ChangePassword.ask_password(confirm: true)
-        store_password(git_url, password)
+        if !UI.interactive?
+          UI.error "No password found neither in environment nor in local keychain. Bailing out as in non interactive mode."
```
KrauseFx: This should have been `UI.user_error!` to raise an exception, right?

PR #7207, opened 2016-11-28
[gym] support Commandline apps 💪💻 2/2

hjanuschka:
@ohwutup i don't know how i managed to smash the previous PR - i had a bad rebase, pushed force (thinking being on a different branch) and :boom: PR was closed, i didn't find a way to reopen it

BIG SORRY FOR THE TROUBLES - stupid me!

split of #7005 

test it with embeded `commandline` project.

```ruby
lane :powerlift do
	cmd_tool
	cmd_lib
	cmd_framework
	macapp
	touch_library
	touch_framework
	touch_app
        tvos
end

lane :tvos do
	 gym(scheme: "tvos",project: "gym/examples/commandline/commandline.xcodeproj", clean: false, output_directory: "out")	
end


lane :cmd_tool do
	 gym(scheme: "cmd-tool",project: "gym/examples/commandline/commandline.xcodeproj", clean: false, output_directory: "out")	
end

lane :cmd_lib do
	 gym(scheme: "cmd-library",project: "gym/examples/commandline/commandline.xcodeproj", clean: false, output_directory: "out")	
end


lane :cmd_framework do
	 gym( output_directory: "out", scheme: "cmd-framework",derived_data_path: "asdf", project: "gym/examples/commandline/commandline.xcodeproj", clean: false)	
end

lane :macapp do
	 gym(scheme: "macapp",project: "gym/examples/commandline/commandline.xcodeproj", clean: false)	
end

lane :touch_library do
	 gym( output_directory: "out", scheme: "touch-library",project: "gym/examples/commandline/commandline.xcodeproj", clean: false)	
end

lane :touch_framework do
	 gym( output_directory: "out", scheme: "touch-framework",project: "gym/examples/commandline/commandline.xcodeproj", clean: false)	
end

lane :touch_app do
	 gym(  output_directory: "out",scheme: "touch-app",project: "gym/examples/commandline/commandline.xcodeproj", clean: false)	
end
```

PR #7591, opened 2016-12-20
[gym] temp. disable `produces_archives?` to fix user issues.

hjanuschka:
as discussed here: #7581 

reverts changes introduced by: https://github.com/fastlane/fastlane/pull/7207 - which causes archive/ipa-move not done when having multiple targets in build phase in a scheme.

bildschirmfoto 2016-12-20 um 09 31 46

` xcodebuild clean -showBuildSettings -workspace Fotoplus-iOS/cewe_myphotos.xcworkspace -scheme CeweMyphotosJenkins -configuration Jenkins`

returns build  settings for app and pods target.
which results in a false/positive return in  `build_settings`  for PRODUCT_TYPE key. (returns from the first, POD, type == framework)

### DIFF
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -11,7 +11,7 @@ module Gym
         clear_old_files
         build_app
       end
-      verify_archive if Gym.project.produces_archive?
+      verify_archive
       FileUtils.mkdir_p(File.expand_path(Gym.config[:output_directory]))
 
       if Gym.project.ios? || Gym.project.tvos?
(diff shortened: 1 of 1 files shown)


gym/lib/gym/runner.rb
```diff
@@ -16,10 +16,10 @@ def run
 
       if Gym.project.ios? || Gym.project.tvos?
         fix_generic_archive # See https://github.com/fastlane/fastlane/pull/4325
-        package_app if Gym.project.produces_archive?
```
KrauseFx: What about `verify_archive`?

gym/lib/gym/runner.rb
```diff
@@ -16,10 +16,10 @@ def run
...
+        package_app
         fix_package
         compress_and_move_dsym
-        path = move_ipa if Gym.project.produces_archive?
+        path = move_ipa
```
KrauseFx: Should we also remove or comment out the `if Gym.project.library? || Gym.project.framework?` part?

KrauseFx: The whole block

hjanuschka:
@KrauseFx  feedback addressed, also removed the `marked by gym` early return

KrauseFx (approved):

Issue #6805, opened 2016-10-31
Application specific password seems wrong


suzhaev:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I turned on 2fa auth for my apple id.
I tried running my fastlane script this evening and received this error:

##### Complete output when running fastlane, including the stack trace and command used

```
[19:36:37]: ▸ [19:36:37]: Ready to upload new build to TestFlight (App: 1161154268)...
[19:36:37]: ▸ [19:36:37]: Going to upload updated app to iTunes Connect
[19:36:37]: ▸ [19:36:37]: This might take a few minutes. Please don't interrupt the script.
[19:36:41]: ▸ 
[19:36:42]: ▸ [19:36:42]: Transporter transfer failed.
[19:36:42]: ▸ [19:36:42]: 
[19:36:42]: ▸ 
[19:36:42]: ▸ [19:36:42]: Application specific password seems wrong
[19:36:42]: ▸ [19:36:42]: Please make sure to follow the instructions
[19:36:42]: ▸ security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
[19:36:42]: ▸ [19:36:42]: Your account has 2 step verification enabled
[19:36:42]: ▸ [19:36:42]: Please go to https://appleid.apple.com/account/manage
[19:36:42]: ▸ [19:36:42]: and generate an application specific password for
[19:36:42]: ▸ [19:36:42]: the iTunes Transporter, which is used to upload builds
[19:36:42]: ▸ [19:36:42]: Going to upload updated app to iTunes Connect
[19:36:42]: ▸ [19:36:42]: This might take a few minutes. Please don't interrupt the script.
[19:36:46]: ▸ 
[19:36:46]: ▸ [19:36:46]: Transporter transfer failed.
[19:36:46]: ▸ [19:36:46]: 
[19:36:46]: ▸ 
[19:36:46]: ▸ [19:36:46]: Application specific password seems wrong
[19:36:46]: ▸ [19:36:46]: Please make sure to follow the instructions
[19:36:46]: ▸ security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
[19:36:46]: ▸ [19:36:46]: Your account has 2 step verification enabled
[...]

Issue #7562, opened 2016-12-18
Failure while using two factor authentication 


chipsnyder:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- Using fastlane 2.0.5
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Spaceship can't find method `store_cookie` and failing while trying to authenticate using 2-factor.
Result says it failed to log in however I am seeing the prompt on my phone and computer to allow the connection and provide the code. However the flow crashes before I can respond. 

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you
```
spaceauth -u redacted_email --capture_output
Logging into to iTunes Connect (redacted_email)...
Could not login to iTunes Connect...
/Users/csnyder/.rvm/gems/ruby-2.3.0/gems/spaceship-0.39.0/bin/spaceauth:25:in `<top (required)>': undefined method `store_cookie' for nil:NilClass (NoMethodError)
	from /Users/csnyder/.rvm/gems/ruby-2.3.0/bin/spaceauth:23:in `load'
	from /Users/csnyder/.rvm/gems/ruby-2.3.0/bin/spaceauth:23:in `<main>'
	from /Users/csnyder/.rvm/gems/ruby-2.3.0/bin/ruby_executable_hooks:15:in `eval'
	from /Users/csnyder/.rvm/gems/ruby-2.3.0/bin/ruby_executable_hooks:15:in `<main>'
```
### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                       |
| ------------------- | ------------------------------------------- |
| OS                  | 10.12.2                                     |
| Ruby                | 2.3.0                                       |
| Bundler?            | true                                        |
| Git                 | git version 2.11.0                          |
| Installation Source | ~/.rvm/gems/ruby-2.3.0/bin/fastlane         |
| Host                | Mac OS X 10.12.2 (16C67)                    |
| Ruby Lib Dir        | ~/.rvm/rubies/ruby-2.3.0/lib                |
| OpenSSL Version     | OpenSSL 1.0.2j  26 Sep 2016                 |
| Is contained        | false                                       |
[...]

PR #7600, opened 2016-12-20
Limit which cookies are stored to avoid bad datatypes.

ohayon:


milch:
We have some 2FA issues open, this seems like it would fix https://github.com/fastlane/fastlane/issues/6805 and https://github.com/fastlane/fastlane/issues/7562

milch (approved):

PR #7601, opened 2016-12-20
Add new xcodebuild options for skip-testing and only-testing to scan

arifken:

- [X] Run `bundle exec rspec` from the root directory.

This PR adds options to Scan to specify specific test bundles/test suites/test cases to be run using the new -only-testing and -skip-testing options in Xcode 8's version of xcodebuild. More info on the actions can be found in the wwdc talk on [Advanced Testing and Continuous Integration](https://developer.apple.com/videos/play/wwdc2016/409/)

### DIFF
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -148,6 +148,34 @@ module Scan
                                      verify_block: proc do |value|
                                        UI.user_error!("File not found at path '#{File.expand_path(value)}'") unless File.exist?(value)
                                      end),
+        FastlaneCore::ConfigItem.new(key: :only_testing,
+                                     env_name: "SCAN_ONLY_TESTING",
+                                     description: "Array of strings matching Test Bundle/Test Suite/Test Cases to run",
+                                     optional: true,
+                                     is_string: false,
+                                     verify_block: proc do |value|
+                                       if value.kind_of?(Array)
+                                         value.each do |test_path|
+                                           UI.user_error!("'only_testing' array should only contain strings, but found: '#{test_path}' which is of type '#{value.class.name}'") unless test_path.kind_of?(String)
+                                         end
+                                       elsif !value.kind_of?(String)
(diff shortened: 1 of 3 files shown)


ohwutup:
Hey @arifken! Thanks for submitting the PR, as a matter of fact, we had a bunch of discussion about this a little while back on another issue https://github.com/fastlane/fastlane/pull/7131 and we did not notice any speed increases or efficiency using the new options.

arifken:
@ohwutup no worries! This is actually for different new options (I did see the issue #7131 ). That issue discusses the build-for-testing or test-without-building options. this is only-testing and skip-testing which are for when you want to run scan, but exclude some tests that you don't want to run/report on. I find this useful if we have a job that just runs one particular suite of tests to generate a slather report just for a single component. But if you don't see value in adding this now, that's totally fine too

scan/lib/scan/options.rb
```diff
@@ -148,6 +148,34 @@ def self.available_options
...
+                                     optional: true,
+                                     is_string: false,
+                                     verify_block: proc do |value|
+                                       if value.kind_of?(Array)
+                                         value.each do |test_path|
+                                           UI.user_error!("'only_testing' array should only contain strings, but found: '#{test_path}' which is of type '#{value.class.name}'") unless test_path.kind_of?(String)
+                                         end
+                                       elsif !value.kind_of?(String)
+                                         UI.user_error!("Incorrect parameter type for 'only_testing': '#{value}'")
+                                       end
```
ohwutup: Could you pull all of this logic into a method to call in both verify_blocks? It looks to be the same right?

ohayon (changes requested):
Hey @arifken! Apologies for mixing up the purpose of the two PRs! xcodebuild has so many options!

There is definitely value in getting this merged in. Currently rubocop is failing because the available options array is too complex, I believe that is because of the verify blocks.

Would you mind trying to clean that up a bit and we can get this merged in! 🚀

mfurtak:
We carried this forward in #8002 👍

PR #7604, opened 2016-12-20
[spaceship] fix path to referenced fastfile

ohayon:


mpirri:
Looks good to me, but should probably do this for all of the Fastfile's for all of the other tools.

milch:
What do we need this for?

ohwutup:
@milch if you try to run `fastlane x` inside of a tool's directory, it spits out an error because the path to _fastlane_'s Fastfile is currently invalid. This fixes that issue.

mpirri (commented):
✨ 👍 ✨

milch:
@ohwutup Can we remove the files instead?

ohwutup:
Thought about that. A couple (_screengrab_ and _PEM_) have extra things in there.

milch:
Hmm, we don't ship single gems anymore, so maybe we should delete them and move the lanes from pem and screengrab into the main Fastfile

ohwutup:
they are specific to to those tools though, putting them in the fastlane dir smells a bit funny, but happy to do that if you think thats what's good!

milch (approved):

PR #7607, opened 2016-12-20
Setting --default-path for rspec and made all testing directions accu…

joshdholtz:
…rate

## Problem 😱 
- Running `bundle exec rspec` did not run all of the tests. 
- Wrong instructions in `ToolsAndDebugging.md`
  - It said to run `bundle exec rspec` in root directory which resulted in 0 tests being run
- Different instructions in `RespondingToIssuesAndPullRequests.md`
  - It said to run `bundle exec rspec` from `fastlane` root directory
- The `circle.yml` file did used a pattern of `./**/*_spec.rb` to run all the files

- `rspec` uses a default pattern of `**{,/*/**}/*_spec.rb` but uses a default path of `spec`
  - This was causing `bundle exec rspec` to not work because it was looking in a `spec` directory which didn't have all the things ❌ 
  - This was causing `bundle exec rspec <directory>` to work because it used the default pattern in a directory that had tests ✅ 
  - This was causing the `circle.yml` file to work (but to be a bit hacky) because the default pattern included the current directoy ❌ ✅ 

## Solution 🚀 
- Set the `--default-path` in the `.rspec` file to `./`
  - `bundle exec rspec` works because it looks in all of the directory
  - `bundle exec rspec <directory>` still works because it overrides that default directory
  - `cirlce.yml` can remove the pattern because of `bundle exec rspec` just working
- Modified doc so everything is correct everywhere

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -2,6 +2,7 @@ require 'rspec/core/rake_task'
 
 RSpec::Core::RakeTask.new(:spec)
 
+--default-path ./
 --require ./spec_helper
 --color
 --format d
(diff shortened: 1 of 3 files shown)


milch (approved):
> so everything is correct everywhere

I wish I heard that more often

PR #7609, opened 2016-12-20
Deprecated actions will display deprecated notes in actions list, action info, while running, and in docs

joshdholtz:
- [x] Show deprecated actions in `fastlane actions`
- [x] Show deprecated action in `fastlane action <name>` 
- [x] Show deprecated action while executing a lane
- [x] Show deprecated actions in docs
- [x] add notes to `notify`
- [x] add notes to `update_project_code_signing`
- [x] deprecate `ipa`
- [x] deprecate and add notes to `s3`
- [x] deprecate and add notes to `upload_symbols_to_sentry`

## How things currently work...

**deprecated** actions have lived in `fastlane` already but have never been _really_ shown or notified to the user in a consistent way. And with the introduction of plugins, some actions _should_ be migrating away from being inside of `fastlane`. The two currently deprecated actions simply have some `UI.important` and `UI.message` calls in them to notify the user but these do not stand out enough and also don't show anything to the user when the action is queried with `fastlane action <name>`.

`notify` and `update_project_code_signing` have been define by the following...
```rb
def self.category
  :deprecated
end
```

## How things WILL work...
This PR will introduce `deprecated_notes` for an action which will inform the user (in a consistent manner) on how to migrate off of these **deprecated** actions.

### Notify (use a different internal action instead)
```rb
def self.category
  :deprecated
end

def self.deprecated_notes
  "It's recommended to use the new 'notification' method instead of 'notify'"
end
```

### S3 (install a plugin)
```rb
def self.category
  :deprecated
end

def self.deprecated_notes
  "Please use the `aws_s3` plugin instead.\n" +
  "Install using `fastlane add_plugin aws_s3`."
end
```

### fastlane actions
screen shot 2016-12-20 at 5 25 46 pm

### fastlane action (without `deprecated_notes`)
screen shot 2016-12-20 at 5 22 51 pm

### fastlane action name (with `deprecated_notes`)
screen shot 2016-12-20 at 5 23 03 pm

### fastlane (while running a lane)
screen shot 2016-12-20 at 5 24 08 pm


hjanuschka:
i really like that one.

fastlane/spec/fixtures/fastfiles/README.md
```diff
@@ -0,0 +1,44 @@
+fastlane documentation
```
joshdholtz: Ha, yup 😇

joshdholtz:
@KrauseFx @milch I think this is ready for a review 😇  Tear it apart ✂️

fastlane/lib/fastlane/documentation/actions_list.rb
```diff
@@ -56,6 +61,13 @@ def self.show_details(filter: nil)
         print_output_variables(action, filter)
         print_return_value(action, filter)
 
+        if Fastlane::Actions.is_class_action?(action) && action.category == :deprecated
+          puts "==========================================".deprecated
+          puts "This plugin (#{filter}) is deprecated".deprecated
```
milch: Calling this "plugin" could be confusing, I'd rather have it say "action"

fastlane/lib/fastlane/documentation/actions_list.rb
```diff
@@ -56,6 +61,13 @@ def self.show_details(filter: nil)
         print_output_variables(action, filter)
         print_return_value(action, filter)
 
+        if Fastlane::Actions.is_class_action?(action) && action.category == :deprecated
+          puts "==========================================".deprecated
```
milch: Instead of `puts` this should be `UI.deprecated`

fastlane/lib/fastlane/documentation/actions_list.rb
```diff
@@ -56,6 +61,13 @@ def self.show_details(filter: nil)
...
+          puts "This plugin (#{filter}) is deprecated".deprecated
+          puts "\n#{action.deprecated_notes}".deprecated if action.deprecated_notes
```
milch: Don't include `\n` in the string please

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
```diff
@@ -82,6 +82,12 @@ def all_actions_from_enhancer
       require 'faraday'
       require 'json'
 
+      if ENV['USE_ENHANCE_TEST_DATA']
```
milch: For testing bool environment vars we now have `FastlaneCore::Env.truthy?("var_name")`. It has some additional features, e.g. it allows the env var to be set to `"0"` or `"false"` to disable it.

milch: Also, please add a short comment above here please as to why this is needed (e.g. only Fabric team members have access to the enhancer instance, so this can be used to check docs changes for everyone else - or something like that)

fastlane/lib/fastlane/runner.rb
```diff
@@ -223,6 +223,13 @@ def execute_action(method_sym, class_ref, arguments, custom_dir: nil)
               UI.user_error!("You have to call the integration like `#{method_sym}(key: \"value\")`. Run `fastlane action #{method_sym}` for all available keys. Please check out the current documentation on GitHub.")
             end
 
+            if Fastlane::Actions.is_class_action?(class_ref) && class_ref.category == :deprecated
+              puts "\n==========================================".deprecated
```
milch: Same here, re: `\n` and `puts`

fastlane/spec/fixtures/plist/Info.plist
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
milch: Revert this please.

Also, out of interest, what's your Xcode version?

fastlane/Fastfile
```diff
@@ -147,6 +147,14 @@ lane :verify_docs do
   end
 end
 
+desc "Generates the actions.md that will get used on https://docs.fastlane.tools"
+desc "This will NOT also automatically submit a pull request to fastlane/docs"
+desc "This SHOULD only be used for testing changes to the docs"
+lane :generate_docs do
```
milch: In accordance with the description above the lane, please rename this to `test_generate_docs`

milch (dismissed):

fastlane/lib/fastlane/documentation/actions_list.rb
joshdholtz: Everything else in here uses `puts`... the UI methods add timestamps which don't go with the rest of the actions list stuff

milch: Ah, right, sorry about that. I'm a simple man, I see `puts` I say "Use `UI` instead".

fastlane/spec/fixtures/plist/Info.plist
joshdholtz: I'm running 8.2

joshdholtz:
@milch Can I get a re-review on this? ☺️

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
```diff
@@ -82,6 +82,14 @@ def all_actions_from_enhancer
       require 'faraday'
       require 'json'
 
+      # Only Fabric team members have access to the enhancer instance
+      # This can be used to check doc changes for everyone else
+      if FastlaneCore::Env.truthy?('USE_ENHANCE_TEST_DATA')
```
KrauseFx: 👍

KrauseFx:
Hey @joshdholtz, this looks good to me, thank you so much for the work you put into this. Could you also add 2-3 tests covering this, so it won't break in the future? 👍

joshdholtz:
@KrauseFx Yes sir, I will work on adding tests 👌

joshdholtz:
@milch @KrauseFx Can I get a re-review on this? 😊

fastlane/lib/assets/Actions.md.erb
```diff
@@ -30,6 +30,8 @@ import './path/to/other/Fastfile'
 
 <%= action.description %>
 
+<%= "> #{action.deprecated_notes.gsub("\n\n", "\n")}" unless action.deprecated_notes.to_s.empty? %>
```
KrauseFx: Could we even highlight this to be red? Maybe using the super fancy colored boxes?

fastlane/lib/fastlane/action.rb
```diff
@@ -129,5 +129,10 @@ def self.shell_out_should_use_bundle_exec?
     def self.other_action
       return OtherAction.new(self.runner)
     end
+
+    # Describes how the user should handle deprecated an action if its deprecated
```
KrauseFx: Maybe mention that this should return a string

fastlane/lib/fastlane/actions/actions_helper.rb
```diff
@@ -134,5 +134,14 @@ def self.load_external_actions(path)
     def self.formerly_bundled_actions
       ["xcake"]
     end
+
+    def self.is_class_action?(class_ref)
```
KrauseFx: Please add a comment on what this method is used for

fastlane/lib/fastlane/actions/actions_helper.rb
```diff
@@ -134,5 +134,14 @@ def self.load_external_actions(path)
...
+      return false if class_ref.nil?
+      class_ref < Fastlane::Action || false
```
KrauseFx: That's a little confusing, can you use a regular if else block to make it easier to understand for contributors?

fastlane/lib/fastlane/actions/notify.rb
```diff
@@ -31,6 +31,10 @@ def self.example_code
       def self.category
         :deprecated
       end
+
+      def self.deprecated_notes
+        "It's recommended to use the new 'notification' method instead of 'notify'"
```
KrauseFx: Maybe even say: "`notify` is deprecated, please use `notification` instead`

fastlane/lib/assets/Actions.md.erb
joshdholtz: @KrauseFx How do I do fancy colored red boxes?

fastlane/lib/fastlane/actions/notify.rb
joshdholtz: Can do! Just copied/pasted this from higher up in this file 😇

fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
```diff
@@ -121,3 +121,9 @@ def verify_interactive!(message)
     end
   end
 end
+
+class String
```
KrauseFx: Is it a good idea to define String here? If somebody doesn't use `shell` as output (in the future), this won't work. As we now call .deprecated across the code base. Why do we use `puts "str".deprecated` in the code base instead of using `UI.deprecated`?

KrauseFx (changes requested):
Hey @joshdholtz, great work on this, I added some small comments before it's ready to be merged 👍

fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
joshdholtz: @KrauseFx `UI.deprecated` puts a timestamp in the front of the line which makes the output on `fastlane actions` and `fastlane action <action_name>` look super weird since nothing else as a timestamp in front of it 😉  

I also wanted `deprecated` as a common method (since its combination of a color and a bold style) so I could use it here - https://github.com/fastlane/fastlane/pull/7609/files#diff-43650ff42a4b937717605d848a2d8f7aR18

KrauseFx: Okay that makes sense, but let's fix it for when we don't include `shell.rb`

fastlane/lib/fastlane/actions/notify.rb
KrauseFx: 👍

fastlane/lib/assets/Actions.md.erb
joshdholtz: @KrauseFx Ping 😊  I am not sure how to make this red 😱

joshdholtz:
@KrauseFx I think we are ready for another review 😊

KrauseFx (approved):

Issue #7592, opened 2016-12-20
Screengrab crashes on Android 7+


SemenovAlexander:
Screengrab uses deprecated file access mode here
https://github.com/fastlane/fastlane/blob/master/screengrab/screengrab-lib/src/main/java/tools.fastlane.screengrab/FileWritingScreenshotCallback.java#L70

obviously, run on Android 7+ device causes Screengrab to crash with 
`java.lang.SecurityException: MODE_WORLD_READABLE no longer supported`

See details:
https://developer.android.com/reference/android/content/Context.html#MODE_WORLD_READABLE

We should think for some other method to store screenshots files in Android 7, screenshots from latest os version is a critical for such tool.

hjanuschka:
@SemenovAlexander can you checkout the PR here -> https://github.com/fastlane/fastlane/issues/7616

according to the docs changing WORLD_READABLE to PRIVATE fixes it - can you give it a try?

SemenovAlexander:
@hjanuschka yeap, just tried, and it works like a charm for me!

However, I have no option to check this on real device, tried only on emulator.

mfurtak:
Hi @SemenovAlexander - thanks for the report!

I'll look into this, and @hjanuschka's proposed fix 👍

mfurtak:
@SemenovAlexander Can you please share with me your `androidTestCompile` statement that sets up `screengrab` for your project?

Also, what is the output of `screengrab -v`?

Thanks!

mfurtak:
@SemenovAlexander I have tried to reproduce this on both an API 25 (Android 7.1.1 Google APIs) emulator, and a Nexus 6p running Nougat. In both cases things are working without issue.

Can you please provide more context around this problem, such as the command line output from running _screengrab_ and/or any stack traces you see from the Android side?

Thanks!

SemenovAlexander:
@mfurtak sorry for late answer, I've missed the reminder somehow. Sure:
`androidTestCompile 'tools.fastlane:screengrab:0.5.6'`
And here is output, when I call `Screengrab.screenshot("test")`:

```
java.lang.RuntimeException: Unable to capture screenshot.
at tools.fastlane.screengrab.FileWritingScreenshotCallback.screenshotCaptured(FileWritingScreenshotCallback.java:57)
at tools.fastlane.screengrab.UiAutomatorScreenshotStrategy.takeScreenshot(UiAutomatorScreenshotStrategy.java:39)
[...]

PR #7616, opened 2016-12-21
[screengrab] handle android 7+ deprecated File Context flag

hjanuschka:
as discussed here -> https://github.com/fastlane/fastlane/issues/7592
waiting for @SemenovAlexander


mfurtak:
So far I'm not able to reproduce the problem that this addresse, so I'm going to close this until we can get more information 👍

mfurtak:
Using `Context.MODE_WORLD_READABLE` is still important for other scenarios on older devices, so we need to keep it for those. The fix in #7838 instead makes sure that devices on >= N can't enter this particular fall-back block. 👍

PR #7612, opened 2016-12-21
Update Gemfile for monogem

milch:
Fixes #7610

PR #7605, opened 2016-12-20
iCell mailgun

ohayon:
@iCell Just moving this PR over here so I could rebase on to master after some major changes we made!

https://github.com/fastlane/fastlane/pull/7374

PR #7570, opened 2016-12-18
[snapshot] add additional ipad classifiers.

hjanuschka:
fixes: https://github.com/fastlane/fastlane/issues/6822

proof:
bildschirmfoto 2016-12-18 um 23 00 36

PR #7632, opened 2016-12-21
Version bump to 2.2.0

### LINKED ISSUE #7630: Gym fails to archive because Fastlane configuration picks the wrong target for settings.
### New Issue Checklist

### Issue Description

Gym is failing to archive my app due to multiple targets inside my workspace. It seems that FastlaneCore is grabbing the wrong target settings when showBuildSettings runs against my workspace.

Snippet from my Fastfile: 
```
      gym(
        workspace: "myworkspace.xcworkspace",
        scheme: "myscheme",
        clean: true,
        configuration: "myconfig",
        output_directory: "../builds",
        output_name: MY_BUILDNAME,
        disable_xcpretty: false
        )
```
It looks like Gym is running: 

```
set -o pipefail && xcodebuild -workspace myworkspace.xcworkspace -scheme myscheme -configuration myconfig -destination 'generic/platform=iOS' -archivePath /my/archive/path/archive.xcarchive clean build | tee /Users/me/Library/Logs/gym/target1-name.log | xcpretty
```

Which later fails because it fails to find the xcarchive created. The xcarchive fails to be created because fastlane is passing `build` instead of `archive` to xcodebuild. It is passing xcodebuild `build` because the build command generator is checking the `Gym.project.library` or `Gym.project.framework` and coming to the conclusion that my target is actually a library or framework. (in gym/generators/build_command_generator.rb ->  def actions line 54)

It's coming to this conclusion because when FastlaneCore attempts to pull the build settings using xcodebuild: 
```
xcodebuild clean -showBuildSettings -workspace myworkspace.xcworkspace -scheme myscheme -configuration myconfig
```

But that returns the build settings for every target in my scheme, right now myworkspace has 7 targets.  It is grabbing the first target (alphabetically), which is a framework, and using the settings for that for the rest of my archiving. 

So what looks like is happening is build_settings in fastlane_core/lib/fastlane_core/project.rb is getting fed in the output from the showBuildSettings command and taking the first values it finds, not the ones for the actual target I want to build. 

If I run: 
```
xcodebuild clean -showBuildSettings -project myproject.xcodeproj -configuration myconfig -target mytarget
```
I see only one output, and that output is for my target. However, I need to run archive against my workspace.

milch:
Auto-generated by fastlane 🤖

**Changes since release '2.1.3':**

* Revert library support of gym, fixing archive errors, fixes #7630 (#7631)
* Update Gemfile for monogem (#7612)
* Setting --default-path for rspec and made all testing directions accurate (#7607)
* [spaceship] fix path to referenced fastfile (#7604)
* iCell mailgun (#7605)
* anon. the sensitive options (#7335)
* [snapshot] add additional ipad classifiers. (#7570)

### DIFF
--- a/fastlane/lib/fastlane/version.rb
+++ b/fastlane/lib/fastlane/version.rb
@@ -1,4 +1,4 @@
 module Fastlane
-  VERSION = '2.1.3'.freeze
+  VERSION = '2.2.0'.freeze
   DESCRIPTION = "The easiest way to automate beta deployments and releases for your iOS and Android apps".freeze
 end


ohayon (approved):
👍

PR #7642, opened 2016-12-22
Add note about version in cocoapods

KrauseFx:
We've already received 3 PRs adding this parameter, this should avoid it in the future.

### DIFF
--- a/fastlane/lib/fastlane/actions/cocoapods.rb
+++ b/fastlane/lib/fastlane/actions/cocoapods.rb
@@ -77,6 +77,9 @@ module Fastlane
                                          UI.user_error!("Could not find Podfile") unless File.exist?(value) || Helper.test?
                                        end)
         ]
+        # Please don't add a version parameter to the `cocoapods` action. If you need to specify a version when running
+        # `cocoapods`, please start using a Gemfile and lock the version there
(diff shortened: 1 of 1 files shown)


endocrimes (approved):
👍👍👍👍👍👍👍👍👍👍👍👍

ohwutup:
Will using a Gemfile work when using the bundled _fastlane_?

ohwutup:
Should we add a note saying that this will not work if _fastlane_ is installed via the zip file or homebrew?

milch:
@ohwutup 

If the fastlane bundle was installed this could still work, since we're shelling out and could call pod using bundle exec. Maybe we should change the line https://github.com/fastlane/fastlane/blob/7908e2af585ce859312972bc2bd9e361f4229b86/fastlane/lib/fastlane/actions/cocoapods.rb#L17 here and remove the `shell_out_should_use_bundle_exec?` and just honor the `use_bundle_exec` param. This is a bit confusing for the user anyway as to why setting `use_bundle_exec` seemingly doesn't do anything.

KrauseFx:
I'm ok either way, however we should merge this PR with the warning for now to avoid getting further PRs adding the version number. Feel free to merge directly after approving the PR 👍

milch (approved):

PR #7667, opened 2016-12-24
Refactor Fastfile changelog methods and open GitHub diff page

KrauseFx:
This will now also open the GitHub diff page, which makes it much easier to see the changes since the last release

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -51,13 +51,6 @@ lane :bump do |options|
   sh "git checkout master"
 end
 
-desc "Display a changelog for fastlane since its last release."
-lane :show_changelog do
-  old_version = current_version
-
-  print_release_changelog(old_version: old_version)
-end
-
 desc "Does everything that's needed for a release"
 desc "This includes running tests and verifying the GitHub release"
 lane :release do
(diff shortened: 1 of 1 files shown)


asfalcone (approved):

fastlane/Fastfile
```diff
@@ -261,14 +254,15 @@ private_lane :no_binary do
 end
 
 desc "Print out the changelog since the last tagged release"
```
mpirri: Update desc to not say "Print out"?  (since it is displaying a web page)

fastlane/Fastfile
```diff
@@ -261,14 +254,15 @@ private_lane :no_binary do
...
+lane :show_changelog do |options|
+  old_version = options[:old_version] || current_version
 
-  changes = Actions.sh("git log --pretty='* %s' #{tool_name}/#{old_version}...HEAD --no-merges ..", log: $verbose).gsub("\n\n", "\n")
-  changes.gsub!("[#{tool_name}] ", "") # some commits are pre-fixed with the tool name in [ ]
+  changes = sh("git log --pretty='* %s' #{old_version}...HEAD --no-merges ..", log: $verbose).gsub("\n\n", "\n")
   changes.gsub!("[WIP]", "") # sometimes a [WIP] is merged
 
+  github_diff_url = "https://github.com/fastlane/fastlane/compare/#{old_version}...master"
+  sh "open #{github_diff_url}"
```
mpirri: Question: what happens if we execute `open` on a CI box (where this might be running on a shell independent from the graphical logged-in user session)?

KrauseFx: I don't think we currently call it from a CI machine

fastlane/Fastfile
KrauseFx: good point

fastlane/Fastfile
mpirri: :+1:

mpirri (approved):
:+1:

PR #7674, opened 2016-12-26
fix json in ruby 2.4

hjanuschka:
should fix problems with ruby 2.4 - as reported here: https://github.com/fastlane/fastlane/issues/7673

PR #7695, opened 2016-12-28
[Regression] Remove `json` gem dependency

mrcljx:
Hey @KrauseFx,

first, thanks again for writing `fastlane` - hope you have a great time at 33C3!

The change #7674 in 2.3.1 forced a downgrade of `activesupport` from `4.2.7.1` to `4.0.13`. Reason: `fastlane` depends on `activesupport < 5` and the latest `4.2.7.1` depends on `json ~> 1.7`. To solve that conflict `gem` reverts to an earlier version that didn't specify a `json` as a dependency.

Also there are quite a few other gems (at least in my list of dependencies) which have the `json` dependency set to `1.7`.

I think the only solution forward is to use the native extension mechanism to install the `json` dependency if it is needed (at least based on the comment in the `.gemspec` the only intention seems to be to 'install it if missing' which this should still do).

---

- [x] Ran `gem build fastlane.gemspec` with Ruby 2.2
- [x] Ran `gem install fastlane-2.3.0.gemspec` under clean Ruby 2.2
- [x] Ran `gem install fastlane-2.3.0.gemspec` under clean Ruby 2.4

---


KrauseFx:
Hey @sirlantis, thanks for the PR. I'm not sure if we really want this separate file just for the dependency. Do you think there is an alternative?

sirlantis:
I'm skeptical but will give it another thought tomorrow (jet-lag tiredness got me again just now).

Do you know why you [added the dependency](https://github.com/fastlane-old/fastlane_core/commit/8e4eb5b33487197902ba9d6fb897e71dcc520b9e) in the first place? It should be [present in all 2.0+ rubies](http://ruby-doc.org/stdlib-2.0.0/libdoc/json/rdoc/JSON.html)... we could just dodge it and remove the dependency.

sirlantis:
Closing this to discuss it further in #7700

sirlantis:
P.S. My proposed change didn't solve the issue after proper testing (`gem uninstall -aIx` is your friend).

PR #7702, opened 2016-12-29
Create dependencyci.yml

KrauseFx:


dependencyci.yml
```diff
@@ -1,6 +1,4 @@
 # don't run unmaintained test on any dependencies in Gemfile
```
mpirri: Is this comment out of date now?  (since we don't specifically call out the `Gemfile` in the configuration...)

KrauseFx: It's still a Gemfile, that references the gemspec 👍

mpirri: Okay, sounds good!

mpirri (approved):
:+1:

PR #7714, opened 2017-01-02
Add sign option to add_git_tag action

Vratislav:
This PR adds an ability to invoke `add_git_tag` with `-s` flag. 

This can be useful when releasing a new version of some library/SDK to verify that the tagged release comes from trusted source.


asfalcone (approved):

asfalcone:
Thanks 👍

PR #7722, opened 2017-01-03
If running --version do not install plugins

ohayon:
This makes sure that running `fastlane -v` does not have the side effect of installing plugins.


mpirri (approved):
:+1:

Issue #7723, opened 2017-01-03
DeviceManager.connected_devices fails parsing USB data


kdubb:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

DeviceManager.connected_devices fails when parsing the USB device information.  It's expecting keys in the hash that are not there.

### Captured Output

Command Used: `build_and_test`
<details><summary>Output/Log</summary>

```

[11:20:05]: Driving the lane 'build_and_test' 🚀
[11:20:05]: ----------------------------------
[11:20:05]: --- Step: ensure_xcode_version ---
[11:20:05]: ----------------------------------
[11:20:05]: $ xcversion selected
[11:20:05]: ▸ Xcode 8.2.1
[11:20:05]: ▸ Build version 8C1002
[11:20:05]: Selected Xcode version is correct: 8.2.1
[11:20:05]: -------------------
[11:20:05]: --- Step: is_ci ---
[11:20:05]: -------------------
+---------------+----------------+
|          Lane Context          |
+---------------+----------------+
| PLATFORM_NAME |                |
| LANE_NAME     | build_and_test |
+---------------+----------------+
[11:20:06]: undefined method `each' for nil:NilClass

+------+----------------------+-------------+
|             fastlane summary              |
+------+----------------------+-------------+
| Step | Action               | Time (in s) |
+------+----------------------+-------------+
| 1    | ensure_xcode_version | 0           |
| 2    | is_ci                | 0           |
+------+----------------------+-------------+

[11:20:06]: fastlane finished with errors

Looking for related GitHub issues on fastlane/fastlane...

➡️  [!] Undefined method `description' for nil:NilClass (NoMethodError)
   https://github.com/fastlane/fastlane/issues/7686 [closed] 4 💬
   5 days ago

➡️  deliver fails with: [!] undefined method `description' for nil:NilClass  
   https://github.com/fastlane/fastlane/issues/7356 [closed] 5 💬
   4 weeks ago

➡️  undefined method `<<' for nil:NilClass
   https://github.com/fastlane/fastlane/issues/6731 [closed] 7 💬
   26 Oct 2016

and 68 more at: https://github.com/fastlane/fastlane/search?q=undefined%20method%20%60each'%20for%20nil&type=Issues&utf8=✓

```

</details>

[...]

Issue #7675, opened 2016-12-26
FastlaneCore::DeviceManager.connected_devices is incorrect when using USB hub


kdubb:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

The `DeviceManager::connected_devices` does not report devices connected to an external USB hub. In [device_manager.rb](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/device_manager.rb#L74l) the comment refers to "looping" in case it has more than 1 controller. 

Unfortunately USB is a tree, for this code to work properly it needs to be a recursive search to find devices connected to a hub, that are connected to a hub, ad infinitum.

##### Complete output when running fastlane, including the stack trace and command used

As this isn't a failure there isn't any stack trace produce.  The `connected_devices` method finishes without exception; it just returns an incorrect list of devices.

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                                          |
| ------------------- | -------------------------------------------------------------- |
| OS                  | 10.12.2                                                        |
| Ruby                | 2.0.0                                                          |
| Bundler?            | false                                                          |
| Git                 | git version 2.10.1 (Apple Git-78)                              |
| Installation Source | /usr/local/bin/fastlane                                        |
| Host                | Mac OS X 10.12.2 (16C67)                                       |
| Ruby Lib Dir        | /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib |
| OpenSSL Version     | OpenSSL 0.9.8zc 19 Mar 2015                                    |
| Is contained        | false                                                          |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version       | 8.2                                                            |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
fastlane_require 'yaml'

workspace = "MessagesKit.xcworkspace"
edge_compositions = {
  'test' => "fastlane/edge-test-compose.yml",
  'dev' => "fastlane/edge-dev-compose.yml"
}


# Build Configuraion

[...]

PR #7725, opened 2017-01-03
Fix USB data parsing problem for DeviceManager

mfurtak:
Addresses #7723

`DeviceManager.connected_devices` fails to parse some returned USB data when some expected keys are not present in all items.

This fix removes the assumption that the `_items` key will be present for all USB items entries.


asfalcone (approved):
Yay tests 😍

mfurtak:
I'll follow up by looking into #7675 next

PR #7727, opened 2017-01-03
Fix USB device discovery to account for USB hubs

mfurtak:
Addresses #7675

USB device discovery is now performed recursively to account for the tree-like nature of USB and USB hubs


fastlane_core/spec/device_manager_spec.rb
```diff
@@ -216,6 +217,24 @@
       expect(devices).to be_empty
     end
 
+    it "properly finds devices in system_profiler output when connected via USB hubs" do
```
KrauseFx: ❤

KrauseFx (approved):

PR #7742, opened 2017-01-04
Detect if session was set via env variable if session is invalid, and show better error message

KrauseFx:
screenshot 2017-01-04 19 22 58

This bit me a lot in the past


ohayon (approved):
👍

Issue #7715, opened 2017-01-02
How to secure the json file for fastlane supply


paulononaka:
I am automating my CI with fastlane supply. Supply requires a json file that is created through Google Play to give fastlane access to upload my APK. Isn't unsafe to keep this json file in my CI? Looks like there is no way to pass the credentials through arguments, only through this file. If I could pass through arguments I could pass environment variables. I'm not expert in the subject, so sorry if I'm saying bullshit.

VenkataMutyala:
@paulononaka @fastlane-bot @KrauseFx I actually have the same issue.

It looks like only a local file is supported at this time.
Per: https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb line 58 to 66  

I think it would help in terms of security by being able to pass in the credentials to supply as an environment variable or a parameter. This way we don't have to keep the .json file stored as plaintext on disk.




Here is my fastlane env output:

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                       |
| ------------------- | ------------------------------------------- |
| OS                  | 10.12.2                                     |
| Ruby                | 2.3.1                                       |
| Bundler?            | false                                       |
| Git                 | git version 2.10.1 (Apple Git-78)           |
| Installation Source | /usr/local/bin/fastlane                     |
| Host                | Mac OS X 10.12.2 (16C67)                    |
| Ruby Lib Dir        | /usr/local/Cellar/ruby/2.3.1_2/lib          |
| OpenSSL Version     | OpenSSL 1.0.2j  26 Sep 2016                 |
| Is contained        | false                                       |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version       | 8.2.1                                       |


### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |


### fastlane files:

**No Fastfile found**


**No Appfile found**


### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
| fastlane | 2.5.0   | ✅ Up-To-Date  |


### Loaded fastlane plugins:

**No plugins Loaded**


<details><summary><b>Loaded gems</b></summary>

| Gem                       | Version      |
| ------------------------- | ------------ |
| did_you_mean              | 1.0.0        |
| slack-notifier            | 1.5.1        |
[...]

PR #7748, opened 2017-01-05
[supply] Added support for google service account credentials via environment variable

venkatamutyala:
[supply] Added support for google service account credentials via environment variable

Existing github issue: https://github.com/fastlane/fastlane/issues/7715

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -7,7 +7,7 @@ require 'net/http'
 
 module Supply
   class Client
-    #environment variable that can be used to set credentials
+    # Environment variable that can be used to set credentials
     SUPPLY_JSON_GOOGLE_CREDS = "SUPPLY_JSON_GOOGLE_CREDS"
 
     # Connecting with Google
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -7,6 +7,9 @@ require 'net/http'
 
 module Supply
   class Client
+    #environment variable that can be used to set credentials
+    SUPPLY_JSON_GOOGLE_CREDS = "SUPPLY_JSON_GOOGLE_CREDS"
+
     # Connecting with Google
     attr_accessor :android_publisher
 
@@ -22,7 +25,7 @@ module Supply
 
     # instantiate a client given the supplied configuration
     def self.make_from_config
-      unless Supply.config[:json_key] || (Supply.config[:key] && Supply.config[:issuer])
+      unless Supply.config[:json_key] || ENV[SUPPLY_JSON_GOOGLE_CREDS] || (Supply.config[:key] && Supply.config[:issuer])
         UI.important("To not be asked about this value, you can specify it using 'json_key'")
         Supply.config[:json_key] = UI.input("The service account json file used to authenticate with Google: ")
       end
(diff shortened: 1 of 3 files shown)


asfalcone:
@VenkataMutyala Thanks for the PR! I am hesitant to add an undocumented environment variable for this. I'd like to see another [option](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb) added to _supply_ that would automatically support an environment variable and conflicting option detection. You could name it something like `:json_key_data` to distinguish that this param contains the key data not the file path. What do you think?

VenkataMutyala:
@asfalcone I agree that is a good idea! I just updated this PR so its now an option. Please let me know what you think.

lacostej:
An argument isn't in my mind much safer than a file on disk. It will be visible a on `ps` output.

The whole concept of protecting secrets is complicated to do right. It depends a lot on your concerns. There are many areas to look at and it is something that in my mind should be carefully looked into. For example, fastlane doesn't sanitize environment variables in sub processes.

I am not aware of a simple solution to this to implement a safe secret manager accros all fastlane tools. The CredentialsManager is a good step but is Mac only right now. As for CI integration, I think every system has (or should have) some kind of mechanism to pass the secrets safely to the build machine. A third party solution, there are interesting projects like https://www.vaultproject.io/ but I didn't delve into any of them.

VenkataMutyala:
@lacostej I agree protecting secrets is really complicated and its unlikely there is a simple solution. As you mentioned it does depend on the security concerns and in the case of this PR supply no longer requires the service account json to be stored at rest.

FWIW, I have seen a demo on VaultProject and its **really cool** and it probably has the most comprehensive feature. A member of the Hashicorp team told me their biggest internal use case is using it as a transient backend where they just encrypt/decrypt data on the fly for their web apps. A few other interesting solutions I have come across for secret management are:
https://square.github.io/keywhiz/
https://lyft.github.io/confidant/

supply/lib/supply/client.rb
```diff
@@ -22,25 +22,28 @@ class Client
...
+                        path_to_service_account_json: Supply.config[:json_key], service_account_json: Supply.config[:json_key_data])
     end
 
     # Initializes the android_publisher and its auth_client using the specified information
+    # @param service_account_json: The raw service account Json data
     # @param path_to_service_account_json: The path to your service account Json file
     # @param path_to_key: The path to your p12 file (@deprecated)
     # @param issuer: Email addresss for oauth (@deprecated)
-    def initialize(path_to_key: nil, issuer: nil, path_to_service_account_json: nil)
+    def initialize(path_to_key: nil, issuer: nil, path_to_service_account_json: nil, service_account_json: nil)
```
asfalcone: Could this be simplified to only accept the `service_account_json` and do the `File.open` in `make_from_config`?

asfalcone (changes requested):

asfalcone:
This is looking great. I left 1 more comment about simplifying the code!

supply/lib/supply/client.rb
VenkataMutyala: sounds good

VenkataMutyala:
@asfalcone please let me know what you think.

asfalcone:
@VenkataMutyala 💯 This is great! Thank you for the contribution!

asfalcone (approved):

Issue #6987, opened 2016-11-14
frameit (Landscape detection not working as expected)


psamhaber:
**Problem:**
Landscape screenshots of iPhone and iPad frames are not framed correctly, see attached.
It seems the landscape detection part doesn't seem to understand the offset/widths correctly?

**Specs:**
- Running frameit directly from the console
- Version: frameit 3.0.0;
- Screenshot sizes are correct (iPhone 750x1334px, iPad 2048x1536px)
- I'm using the latest frames version

![ipad-horz-max-00_framed](https://cloud.githubusercontent.com/assets/4225828/20266459/766300a4-aa76-11e6-8df4-25feb30e2295.png)

![digitalcontent-01_test_framed](https://cloud.githubusercontent.com/assets/4225828/20266440/640dcd80-aa76-11e6-9c46-f113deee0a37.png)

psamhaber:
Can anyone reproduce this?

TKBurner:
@psamhaber I think I've seen another report elsewhere so we are going to test on our end. Thanks!

futuretap:
Yes, I can reproduce this when no background is set.

psamhaber:
@TKBurner any update on this? I would fix it myself but my ruby skills are basic at best. 

Looking at editor.rb (line 23ff) I guess the problem lies in `width = offset['width']`. If you check the resulting screens I attached above, it looks like the width value is taken from before the frame is flipped...

mfurtak:
Thanks for the report! I've reproduced this locally, and will look into it. 👍

mfurtak:
Hi folks - I've merged a change in #7759 that seems to fix this problem. If possible, could you give it a try before we ship it in the next release?

Thanks! 🙌

psamhaber:
Hi Michael, it's working again, awesome! Thanks a bunch!

mfurtak:
Great - glad to hear it! 🙌 

This has shipped as part of [_fastlane_ 2.6.0](https://github.com/fastlane/fastlane/releases/tag/2.6.0) :rocket:

PR #7759, opened 2017-01-05
Fix frameit 'easy mode' problem in landscape

mfurtak:
Addresses #6987 by removing some resizing code in the "easy" mode, that does not appear to be necessary.


KrauseFx (approved):

KrauseFx:
Mh, I wrote the initial code, which seems to be wrong, so this looks fine to me 👍

mfurtak:
As discussed with @KrauseFx out-of-band, the resizing appears to not be necessary here. Removing it fixes the problem.

KrauseFx:
🚀

PR #7762, opened 2017-01-06
Automatic-retry for iTC OPERATION_FAILED error messages in spaceship

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/6419

### DIFF
--- a/spaceship/lib/spaceship/tunes/build_train.rb
+++ b/spaceship/lib/spaceship/tunes/build_train.rb
@@ -51,9 +51,8 @@ module Spaceship
         # @param app_id (String) The unique Apple ID of this app
         def all(application, app_id, platform: nil)
           trains = []
-          # Array for when there is no valid server response
-          trains += Array(client.build_trains(app_id, 'internal', platform: platform)['trains'])
-          trains += Array(client.build_trains(app_id, 'external', platform: platform)['trains'])
+          trains += client.build_trains(app_id, 'internal', platform: platform)['trains']
+          trains += client.build_trains(app_id, 'external', platform: platform)['trains']
 
           result = {}
           trains.each do |attrs|
--- a/spaceship/lib/spaceship/tunes/build_train.rb
+++ b/spaceship/lib/spaceship/tunes/build_train.rb
@@ -51,8 +51,9 @@ module Spaceship
         # @param app_id (String) The unique Apple ID of this app
         def all(application, app_id, platform: nil)
           trains = []
-          trains += client.build_trains(app_id, 'internal', platform: platform)['trains']
-          trains += client.build_trains(app_id, 'external', platform: platform)['trains']
+          # Array for when there is no valid server response
+          trains += Array(client.build_trains(app_id, 'internal', platform: platform)['trains'])
+          trains += Array(client.build_trains(app_id, 'external', platform: platform)['trains'])
 
           result = {}
           trains.each do |attrs|
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -574,8 +574,9 @@ module Spaceship
     # @!group Build Trains
     #####################################################
 
+    # rubocop:disable Metrics/BlockNesting
     # @param (testing_type) internal or external
-    def build_trains(app_id, testing_type, platform: nil, tries: 5)
+    def build_trains(app_id, testing_type, tries = 5, platform: nil)
       raise "app_id is required" unless app_id
       url = "ra/apps/#{app_id}/trains/?testingType=#{testing_type}"
       url += "&platform=#{platform}" unless platform.nil?
@@ -585,16 +586,18 @@ module Spaceship
       # Build trains fail randomly very often
       # we need to catch those errors and retry
       # https://github.com/fastlane/fastlane/issues/6419
-      error_content = ex.to_s
-      if error_content == "{\"data\"=>nil, \"messages\"=>{\"warn\"=>nil, \"error\"=>[\"ITC.response.error.OPERATION_FAILED\"], \"info\"=>nil}, \"statusCode\"=>\"ERROR\"}"
-        unless (tries -= 1).zero?
+      if ex.to_s.include?("ITC.response.error.OPERATION_FAILED")
+        tries -= 1
+        if tries > 0
           logger.warn("Received temporary server error from iTunes Connect. Retrying the request...")
           sleep 3 unless defined? SpecHelper
           retry
         end
       end
-      return {} # better than nothing
+
+      raise Spaceship::Client::UnexpectedResponse, "Temporary iTunes Connect error: #{ex}"
     end
+    # rubocop:enable Metrics/BlockNesting
 
     def update_build_trains!(app_id, testing_type, data)
       raise "app_id is required" unless app_id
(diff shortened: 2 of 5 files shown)


spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -575,12 +575,22 @@ def candidate_builds(app_id, version_id)
     #####################################################
 
     # @param (testing_type) internal or external
-    def build_trains(app_id, testing_type, platform: nil)
+    def build_trains(app_id, testing_type, platform: nil, tries: 5)
```
KrauseFx: The `tries` value isn't working yet, it seems like it's not used because we're using `:` instead of the `=` notation. Looking into this

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -575,12 +575,25 @@ def candidate_builds(app_id, version_id)
...
       url += "&platform=#{platform}" unless platform.nil?
       r = request(:get, url)
-      parse_response(r, 'data')
+      return parse_response(r, 'data')
+    rescue Spaceship::Client::UnexpectedResponse => ex
+      # Build trains fail randomly very often
+      # we need to catch those errors and retry
+      # https://github.com/fastlane/fastlane/issues/6419
+      error_content = ex.to_s
+      if error_content == "{\"data\"=>nil, \"messages\"=>{\"warn\"=>nil, \"error\"=>[\"ITC.response.error.OPERATION_FAILED\"], \"info\"=>nil}, \"statusCode\"=>\"ERROR\"}"
```
mfurtak: This feels very precise, perhaps overly so. I worry that we'll stop doing this behavior if the output format changes even a little bit.

Might it make more sense to just match on a subset of the response? Say, `ITC.response.error.OPERATION_FAILED`

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -575,12 +575,25 @@ def candidate_builds(app_id, version_id)
...
+        unless (tries -= 1).zero?
```
mfurtak: This is a personal preference, (and somewhat nitpicky) but I strongly dislike conditional checks that include a side-effect.

The reason is that it becomes possible to accidentally prevent the side-effect from happening in later edits, by introducing a short-circuiting operation to the conditional. So, later on someone edits it to be:

`unless other_thing && (tries -= 1).zero?`

and if `other_thing` causes a short-circuited evaluation, now your counting is possibly wrong, or infinitely looping.

So my preference is to do the side-effect operation separately, then do the boolean expression on it. :+1:

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: Agreed. Initially I wanted to parse the actual content via objects, but we don't have access to the raw JSON here any more. Your approach sounds good to me.

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: Cool. I'll update. I took that code from client.rb

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -575,12 +575,25 @@ def candidate_builds(app_id, version_id)
...
+          logger.warn("Received temporary server error from iTunes Connect. Retrying the request...")
+          sleep 3 unless defined? SpecHelper
+          retry
+        end
+      end
+      return {} # better than nothing
```
lacostej: this means that if we keep failing after retrying, we return empty result. Don't we want to fail instead?

lacostej: This scenario lacks a test as well

spaceship/lib/spaceship/tunes/tunes_client.rb
lacostej: can this error happens elsewhere i.e. not just for build_trains.  Given #1631, it might then once again make more sense to move this special retry under `with_tunes_retry`

lacostej (changes requested):
It looks like the retry code would benefit from being implemented in the `with_tunes_retry`. The swallowing of the error after many retries is also a bit weird.

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: Oh yeah, great point, I wasn't thinking this through. Maybe we should even show a custom error message, that this might be a temporary server error 👍

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: Good question, I asked about this on https://github.com/fastlane/fastlane/issues/6419#issuecomment-270919290, and it seems like it only affects build trains, we can extend this to more endpoints as we go.

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: Interesting, how does it relate to `with_reply`? I see that `with_tunes_retry` is only used once, however it seems like it is used together, since the actual request is always sent with `with_reply`, correct? @lacostej

KrauseFx: I'd need to update the `with_tunes_retry` to also catch that error

KrauseFx:
@mfurtak @lacostej thanks for all your feedback, I addressed it all, and also updated the tests to cover all the different cases 👍

spaceship/lib/spaceship/tunes/build_train.rb
```diff
@@ -51,8 +51,9 @@ def factory(attrs)
         # @param app_id (String) The unique Apple ID of this app
         def all(application, app_id, platform: nil)
           trains = []
-          trains += client.build_trains(app_id, 'internal', platform: platform)['trains']
-          trains += client.build_trains(app_id, 'external', platform: platform)['trains']
+          # Array for when there is no valid server response
```
lacostej: @KrauseFx given that we never return an empty array, is this still needed?

lacostej (approved):

spaceship/lib/spaceship/tunes/build_train.rb
KrauseFx: Great point, I'm going to remove this 👍

spaceship/spec/tunes/server_failures_spec.rb
```diff
@@ -0,0 +1,33 @@
+describe Spaceship::TunesClient do
+  describe "Random Server Failures" do
+    before { Spaceship::Tunes.login }
+    let(:client) { Spaceship::Application.client }
+    let(:app) { Spaceship::Application.all.first }
+
+    describe "#build_trains failing" do
+      it "automatically re-tries the request when getting a ITC.response.error.OPERATION_FAILED when receive build trains" do
```
mfurtak: ❤️

KrauseFx:
Tests are failing after removing the `Array`, looking into this tomorrow

lfarah:
I didn't get any timeout or errors. Just left it there for a lot of hours
screen shot 2017-01-07 at 2 34 04 pm
screen shot 2017-01-07 at 2 34 10 pm

KrauseFx:
Updated the PR to fix the last remaining test, ready to be merged from my side 👍

PR #7771, opened 2017-01-07
Fix FastlaneRequire#gem_installed? to handle already-loaded libs

### LINKED ISSUE #7770: fastlane_require fails on dependencies fastlane has already loaded
### New Issue Checklist

### Issue Description

When trying to `require` a dependency (in this case `base64`) from the Ruby standard library, I got a warning to use `fastlane_require` instead. While the `require` worked just fine, I switched over to `fastlane_require` and was met with an error telling me to add `base64` to my gemfile. This seemed odd since `base64` isn't a gem, so I submitted #7769.

Looking into it more, it seems the real issue is that `fastlane_require`'s `gem_installed?` check is misinterpreting a `false` return value from [Kernel#require](http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-require). False in this context means that the library has already been loaded, not that the `require` failed.

reidab:
`fastlane_require`'s `gem_installed?` check seems to be misinterpreting a `false` return value from [Kernel#require](http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-require).

The `require` returns true when that call to require is the one that loads the library and false if the library was already loaded. The `gem_installed?` check is treating a false return as a failure to find/load the requested library.

Fixes #7770

---

### DIFF
--- a/fastlane/lib/fastlane/fastlane_require.rb
+++ b/fastlane/lib/fastlane/fastlane_require.rb
@@ -45,7 +45,7 @@ module Fastlane
         # See https://github.com/fastlane/fastlane/issues/6951
         fork do
           begin
-            exit(1) unless require name
+            require name
           rescue LoadError
             exit(1)
           end


KrauseFx (approved):

KrauseFx:
Thank you so much for your contribution, the change makes sense to me. @milch any concerns? 👍

milch:
@krausefx Makes sense!

PR #7763, opened 2017-01-06
Update gemspec to require correct version of Play API

tmtrademarked:
The deobfuscation upload feature in supply requires 0.9.2+ of the google-api-client gem. I didn't realize this because I had a version of that gem on my system already.

bundle exec rspec is failing with a bunch of errors that seem unrelated to this change, but I'm not sure how best to diagnose them. (They also fail on master on my machine) Most likely this is because I don't have Xcode installed. I'll try to get it installed, but didn't want to block uploading this PR on that.

Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:

- [ ] Run `bundle exec rspec` from the root directory.
- [ X ] Run `bundle exec rubocop -a` to ensure the code style is valid
- [ X ] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ X ] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Before submitting a pull request, we appreciate if you create an issue first to discuss the change :+1:

PR #7743, opened 2017-01-04
Store spaceship and frameit config in ~/.fastlane

KrauseFx:

PR #7774, opened 2017-01-08
Version bump to 2.6.0

KrauseFx:
Auto-generated by fastlane 🤖

**Changes since release '2.5.0':**

* Make waiting for build processing more stable by retrying failed requests (#7762)
* Fix FastlaneRequire#gem_installed? to handle already-loaded libs (#7771)
* [supply] Added support for google service account credentials via environment variable (#7748)
* Update in-code documentation tool calls to use new fastlane syntax (#7751)
* Update screengrab tooling and dependency versions (#7766)
* Update gemspec to require correct version of Play API (#7763)
* Fix frameit 'easy mode' problem in landscape (#7759)
* frameit: support resuming of interrupted downloads of frames (#7750)
* Store spaceship and frameit config in ~/.fastlane (#7743)
* Detect if session was set via env variable if session is invalid, and show appropriate error message (#7742)

### DIFF
--- a/fastlane/lib/fastlane/version.rb
+++ b/fastlane/lib/fastlane/version.rb
@@ -1,4 +1,4 @@
 module Fastlane
-  VERSION = '2.5.0'.freeze
+  VERSION = '2.5.1'.freeze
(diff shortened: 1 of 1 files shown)


mfurtak (approved):
👍

Issue #7787, opened 2017-01-09
Spacechip - uninitialized constant Spaceship::Client::FastlaneCore


igorstimoli:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I have a little ruby script that starts with

```ruby
#!/usr/bin/env ruby
require "spaceship"

itunes_username = ARGV[0]
itunes_password = ARGV[1]
app_identifier = ARGV[2]

client = Spaceship::Tunes.login(itunes_username, itunes_password)
```
since I updated to fastlane 2.6.0 It breaks at login with the following error
```
NameError: uninitialized constant Spaceship::Client::FastlaneCore
```

##### Complete output when running fastlane, including the stack trace and command used

```
NameError: uninitialized constant Spaceship::Client::FastlaneCore
Did you mean?  Fastlane
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/client.rb:175:in `persistent_cookie_path'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/two_step_client.rb:81:in `load_session_from_file'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/client.rb:264:in `send_shared_login_request'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/tunes/tunes_client.rb:164:in `send_login_request'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/client.rb:437:in `do_login'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/client.rb:246:in `login'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/client.rb:93:in `login'
	from /usr/local/lib/ruby/gems/2.3.0/gems/fastlane-2.6.0/spaceship/lib/spaceship/tunes/spaceship.rb:22:in `login'
	from (irb):2
	from /usr/local/bin/irb:11:in `<main>'
```
### Environment
[...]

PR #7789, opened 2017-01-09
Fix FastlaneCore method not available in spaceship

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/7787


mfurtak (approved):
👍

PR #7804, opened 2017-01-10
Add support to copy simulator device logs to output directory

lyndsey-ferguson:
This is useful to help debug problems that occur in the simulated app
when running UI tests on remote CI machines.


lyndsey-ferguson:
**Note**: where should I put an automated test for this?

scan/lib/scan/options.rb
```diff
@@ -97,6 +97,12 @@ def self.available_options
                                      env_name: "SCAN_BUILDLOG_PATH",
                                      description: "The directory were to store the raw log",
                                      default_value: "#{FastlaneCore::Helper.buildlog_path}/scan"),
+        FastlaneCore::ConfigItem.new(key: :include_simulator_logs,
+                                     env_name: "SCAN_INCLUDE_SIMULATOR_LOGS",
+                                     description: "Should the logs of the app on the Simulator be written to the output_directory",
```
KrauseFx: What logs are they?

scan/lib/scan/options.rb
```diff
@@ -97,6 +97,12 @@ def self.available_options
...
+                                     type: TrueClass,
```
KrauseFx: Does this work like this?

scan/lib/scan/options.rb
lyndsey-ferguson: These are the logs provided by the iOS app itself. For example, if you write `NSLog` statements in your app, these get written to the simulated device's system.log. You'll never see those log statements in your test output.

Super useful if you're trying to debug an issue that only occurs on a remote iOS CI slave and you absolutely need to know what is happening in the app.

lyndsey-ferguson: [This](https://github.com/soffes/SAMKeychain/issues/52) is such an issue that I solved by modifying a copy of `scan` on our Jenkins CI slave to produce the logs.

KrauseFx: Sounds good, do you mind clarifying the option's description to mention "the app's output (e.g. NSLog)" or similar 👍

scan/lib/scan/options.rb
lyndsey-ferguson: Do you mean does this ConfigItem actually work? It should as I used the same code for [import_certificate](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/import_certificate.rb#L35), but I have not created any tests against it and would like to know where the best place would be to write such a test (I don't see specific tests for options in `scan`)

lyndsey-ferguson:
@KrauseFx code review comments addressed. Can you take another look? Also, is there a place that I can add a test for this, or should I add a new spec file?

lyndsey-ferguson:
@KrauseFx bump

lyndsey-ferguson:
@KrauseFx I improved the code to only try and copy when the simulator device log exists. I also added unit tests around this new functionality in a new file, `runner_spec.rb`.

If it looks good, can you merge? If not, let me know and I'll fix it up.

lyndsey-ferguson:
@KrauseFx bump

lyndsey-ferguson:
@KrauseFx bump

hjanuschka (approved):
nice! thx for your contribution 🚀

lyndsey-ferguson:
Thank you @hjanuschka !

hjanuschka:
@lyndsey-ferguson do you have any plan's on adding this to `snapshot` too? (as i do my integration tests, wherease the sim log would be awesome too )

lyndsey-ferguson:
@hjanuschka I could definitely. We don't use it yet so my knowledge of how `snapshot` works is not as deep. Give me a bit 'o time and I'll look into it.

lyndsey-ferguson:
Hm. It may be almost the same as `scan`; I could put it in the runner's handle results method

PR #5159, opened 2016-06-18
Scan Support for connected devices

matthewellis:
Relates to #5160 

This adds support for testing using connected devices for scan.

I refactored the device selection to use DeviceManager so physical devices can be tested by adding the device name to list of devices.

PR #7808, opened 2017-01-10
Revert "Scan Support for connected devices"

mpirri:
Reverts fastlane/fastlane#5159, due to concerns about naming ambiguity.


ohayon (approved):
👍

PR #7810, opened 2017-01-10
Performance: Make .env Dir.glob much less greedy

reidab:
I noticed that even the most basic `fastlane` command was hanging for a long time between the version check and the start of running a lane, so I started poking around with `ruby-prof`. The majority of the time was being taken up in `Dir.glob` inside `LaneManager.load_dot_env`.

The glob in question was searching recursively for files matching `*.env*` in all subdirectories, but then only loading `.env` files from either the root or the Fastlane folder (if defined).

The project directory where I ran across this contains an obscene number of files (~690,000), so I was seeing a good 20 second lag while it chewed through them all.

This PR changes the glob in that guard statement to only search the directories where `.env` files will be loaded from, instead of recursing through the entire directory structure.

---

### DIFF
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -162,19 +162,20 @@ module Fastlane
     end
 
     def self.load_dot_env(env)
-      return if Dir.glob("**/*.env*", File::FNM_DOTMATCH).count == 0
+      base_path = FastlaneCore::FastlaneFolder.path || '.'
+      return if Dir.glob(File.join(base_path, '*.env*'), File::FNM_DOTMATCH).count == 0
       require 'dotenv'
 
       Actions.lane_context[Actions::SharedValues::ENVIRONMENT] = env if env
 
       # Making sure the default '.env' and '.env.default' get loaded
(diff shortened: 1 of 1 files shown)


KrauseFx:
Oh wow, nice catch, thanks for analysing this 👍

KrauseFx (approved):

KrauseFx:
This looks good to me, thanks for the PR. The only thing is that people might have their .env files in the root of the project or in the `./fastlane` dir, so we should keep supporting both 👍 
Also I'd like @joshdholtz to do the final review before we go ahead  🚀

joshdholtz (approved):
This makes so much sense 👍  Great find!

KrauseFx:
> This looks good to me, thanks for the PR. The only thing is that people might have their .env files in the root of the project or in the `./fastlane` dir, so we should keep supporting both 👍 

Did this get addressed? It's very important to not break any existing setups @joshdholtz @reidab

lacostej:
@KrauseFx I tested with 2.8.0 and I think it didn't. I submitted #7872

reidab:
I'd meant to check that case, but missed it before the merge went in. 😢 The approach in #7872 looks reasonable to me. 👍

PR #7824, opened 2017-01-11
Update screengrab aar version

asfalcone:
Updating screengrab version to match the version of screengrab currently in [rubygems](https://rubygems.org/gems/screengrab/versions). I'll deploy the aar after this commit.

Going forward we'll come up with a sensible strategy for versioning the aar and making sure it is deployed accordingly.

### DIFF
--- a/screengrab/version.properties
+++ b/screengrab/version.properties
@@ -1,3 +1,3 @@
-major=0
-minor=6
-patch=1
+major=1
+minor=0
+patch=0


mfurtak (approved):

KrauseFx:
Whops, sorry, seems like that's something we should have done with the mono gem release?

asfalcone:
Yes, _screengrab_ needs to have the aar deployed independently of the _fastlane_ gem using `fastlane deploy_aar`

Our main question is how to version the aar now. In the past we kept it updated with the _screengrab_ gem even if there were no code changes. I think we have 2 options:

1) Keep the _screegrab_ aar at the same version as _fastlane_ and update the aar every time we release. Maybe makine `deploy_aar` part of the main deploy script. This would be easy to understand from a versioning perspective but would mean very frequent updates and fast moving version numbers even when there is no code changed.
2) Allow the aar to have it's own version number that is only bumped when changes are made to the client side code. This is more difficult to remember when deploying and users have to know a _fastlane_ gem version and a _screengrab_ aar version, but keeps the version number meaningful.

Thoughts? It would be great to hear from folks using _screengrab_ as well. What would be best for your workflow?

KrauseFx:
From my side ideally this would happen automatically when we push a new fastlane release, so that we don't forget

mfurtak:
I think that as a consumer of the Android library, I would be confused/upset if the version number was changing rapidly without any difference in the library code.

So I'd vote for option 2. That would mean keeping the `version.properties` file around to manage the separate AAR version number.

Then, ideally we could include `deploy_aar` as part of the main deploy script, but only run it if the current version is not already released?

PR #7838, opened 2017-01-12
Prevent users on Android N from attempting to use internal storage

mfurtak:
### Checklist
- [x] I've updated the documentation if neccessary.

### Description
As reported in #7592, the existing code will encounter a SecurityException on Android N, if the code attempts to use the internal storage location for screenshots.

This change prevents that fall-back from being tried on Android N and above.

### DIFF
--- a/screengrab/example/fastlane/Appfile
+++ b/screengrab/example/fastlane/Appfile
@@ -1,3 +1,1 @@
-issuer "asd" # Follow https://github.com/fastlane/supply#setup to get one
-keyfile "sdf" # Path to the keyfile
 package_name "tools.fastlane.localetester"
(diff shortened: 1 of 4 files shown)


screengrab/example/fastlane/Appfile
```diff
@@ -1,3 +1 @@
-issuer "asd" # Follow https://github.com/fastlane/supply#setup to get one
-keyfile "sdf" # Path to the keyfile
```
mfurtak: These values were not used, and were causing deprecation warnings

screengrab/example/fastlane/Screengrabfile
```diff
@@ -1,4 +1,4 @@
 locales ['en-US', 'fr-FR', 'ja-JP']
 clear_previous_screenshots true
-tests_apk_path 'build/outputs/apk/example-debug-androidTest-unaligned.apk'
```
mfurtak: Newer build tools are not building an unaligned APK

asfalcone (approved):

PR #7845, opened 2017-01-12
Fix for multiple Gradle tasks in single command

araratispiroglu:
### Checklist
- [X] I've updated the documentation if neccessary.

### Description
I've changed gradle action to allow multiple tasks in single command. Simply If Anyone wants to execute multiple gradle tasks with single command and the command doesn't starts with assemble task currently fastlane will not give output path. And the change-set will fix this.

### Motivation and Context
I've tried to run gradle(task:"clean assemble") in lane, It always generates new .apk file but never give result for SharedValues::GRADLE_APK_OUTPUT_PATH parameter from the lane_context. Then I've realised there was a task.start_with?('assemble') control and I've changed the string control to the regex match. If it finds any word starts with assemble in tasks apk output path parameter will not be nil.


asfalcone (approved):
perfect, thanks 👍

PR #7855, opened 2017-01-13
Increase ulimit to prevent issues with too many open files

asfalcone:
### Description

Sometimes the Circleci build fails with 
`/Users/distiller/.gem/ruby/2.3.1/gems/coveralls-0.8.17/lib/coveralls/output.rb:63:in `require': Too many open files - getcwd (Errno::EMFILE)`

### Motivation and Context

Increasing from the default of 250 to 400 should give us enough headroom to run without failing but not increase too high to hide actual problems

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,6 +7,8 @@ dependencies:
   override:
     - bundle install --jobs=4 --retry=3
 test:
+  pre:
+    - ulimit -n 400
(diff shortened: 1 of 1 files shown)


endocrimes (approved):

DanToml:
:shipit:

PR #7859, opened 2017-01-14
for_lane, for_platform blocks in configurations

mfurtak:
A continuation of work on #6549 

This reconciles the new configuration blocks (`for_lane`, `for_platform`) and their desire to be able to override configuration values, with the existing rules about "first assignment wins" for configurations.

It does this by having the new configuration blocks create a fresh configuration context every time you enter one. This way, each block gets a fresh application of the "first assignment wins" rule. As each block exits, any values set during the block are merged with any configuration context that was present at the time the block started. Thus, we can override values in these blocks, but only once, and still build a single comprehensive configuration!

This is supported with new `push_values!` and `pop_values!` methods on `Configuration`

If we like the solution, we can always add documentation as a follow-up.

/cc @maschall

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
@@ -80,13 +80,13 @@ module FastlaneCore
       end
     end
 
-    # Override configuration for a specific lane.
+    # Override configuration for a specific lane. If received lane name does not
+    # match the lane name available as environment variable, no changes will
+    # be applied.
     #
-    # lane_name  - Symbol representing a lane name.
-    # block - Block to execute to override configuration values.
+    # @param lane_name Symbol representing a lane name.
+    # @yield Block to run for overriding configuration values.
     #
-    # Discussion If received lane name does not match the lane name available as environment variable, no changes will
-    #             be applied.
     def for_lane(lane_name)
       if ENV["FASTLANE_LANE_NAME"] == lane_name.to_s
         with_a_clean_config_merged_when_complete do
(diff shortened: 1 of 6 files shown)


asfalcone (approved):
LGTM but definitely think @KrauseFx should review as well

fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
```diff
@@ -8,6 +8,12 @@ class ConfigurationFile
     # @param path [String] The path to the configuration file to use
     def initialize(config, path, block_for_missing)
       self.config = config
+
+      # Setting this variable to be true allows a one-time overwrite of a particular
+      # configuration value. This is used by the override block methods to allow
+      # giving a more specific config value for a particular circumstance
+      @allow_overwite = false
```
mfurtak: Whoops, this should be removed. Is leftover from my first try/approach 😦

fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
```diff
@@ -78,5 +79,48 @@ def method_missing(method_sym, *arguments, &block)
         end
       end
     end
+
+    # Override configuration for a specific lane.
+    #
+    # lane_name  - Symbol representing a lane name.
+    # block - Block to execute to override configuration values.
+    #
+    # Discussion If received lane name does not match the lane name available as environment variable, no changes will
```
KrauseFx: What's the Discussion block?

fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
```diff
@@ -78,5 +79,48 @@ def method_missing(method_sym, *arguments, &block)
...
+          yield
+        end
+      end
+    end
+
+    # Allows a configuration block (for_lane, for_platform) to get a clean
+    # configuration for applying values, so that values can be overridden
+    # (once) again. Those values are then merged into the surrounding
+    # configuration as the block completes
+    def with_a_clean_config_merged_when_complete
```
KrauseFx: This is great

fastlane_core/spec/configuration_file_spec.rb
```diff
@@ -119,6 +119,60 @@
           end
         end)
       end
+
+      describe "for_lane and for_platform support" do
```
KrauseFx: Perfect, just new tests, without modifying any of the existing ones 👍

KrauseFx (approved):
This looks great @mfurtak 👍 I don't know for sure where we want to add this to the documentation, besides `Advanced.md`. Do you have a better idea?

fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
mfurtak: Hm, this was from the original PR. I can make the comments look more like our usual documentation. 👍

mfurtak:
@KrauseFx https://github.com/fastlane/docs/pull/256

Issue #7072, opened 2016-11-18
Can't specify a custom dotenv .env file location


tristanbbq:
According to [docs](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Advanced.md), I can specify custom .env files to load when running fastlane, but unfortunately it does not seem possible to specify a custom .env file **location**.

For example, I already have a `.env` file at the root of my project. I'd like to load that exact same file instead of having a duplicate in my /fastlane folder.

TKBurner:
@tristanbbq Thanks for reaching out. Right now this isn't possible but we'd be interested in seeing this added. Let me know if you are interested in submitting a PR. 

Also, have you tried using imports? Thanks!

tristanbbq:
@TKBurner: it'd be happy to contribute - not very familiar with the ruby world but I'll definitely give it a shot if it's not fixed anytime soon. Please keep me updated if it gets added in the meantime.

PR #7872, opened 2017-01-16
Search for dotenvs in either fastlane's or current dir

lacostej:
This should allow the .env files to be located one level up the fastlane directory (as requested by @KrauseFx  in #7810) and partially solve #7072

**Details**

this has been merged, but I would be happy if someone could get some more eyes on it.

My goal was to have our dotenvs in the parent directory of fastlane. Today we have links so that they work properly, but this is not fun to maintain at all. This parent directory is the current directory when it comes to running fastlane.

My understanding is that it was requested (#7072) and it was maybe a regression (#7810). I couldn't confirm that this feature was working before. I made the following test project
```
$ cat .env 
BUILD_VERSION=1.0.2
$ cat Gemfile
ruby "2.2.4"
source 'https://rubygems.org'

gem 'fastlane', "2.7.0"
$ cat fastlane/Fastfile 
lane :test do
  UI.message("BUILD_VERSION: #{ENV['BUILD_VERSION']}")
end

$ bundle install
...
$ bundle exec fastlane test
[07:22:25]: Driving the lane 'test' 🚀
[07:22:25]: BUILD_VERSION: 
[07:22:25]: fastlane.tools finished successfully 🎉
```

and tested with various versions (1.80, 2.2.0, 2.6.0, 2.7.0. 2.8.0) and couldn't see any version that displays the environment variable. I checked the dotenv code as well and it didn't change since 2015 until the performance fixes in the past weeks.

The drawbacks I can see in my change
* it could affect people who had no dotenv under fastlane folder before, but a dotenv in the current directory. Their dotenv might now be loaded.
* is that it will only find it if the dotenvs are in the current directory.
If you run the command from within the `fastlane` directory, then the dotenv in the parent aren't read
* one could instead want to load explicitly from fastlane's parent instead of current, which might be more consistent/repeatable

Note that if you had dotenvs in both fastlane folder and current, only fastlane's are loaded.

I am happy to add tests if needed, but would like to have some feedback on the change.

### DIFF
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -162,8 +162,12 @@ module Fastlane
     end
 
     def self.load_dot_env(env)
-      base_path = FastlaneCore::FastlaneFolder.path || '.'
-      return if Dir.glob(File.join(base_path, '*.env*'), File::FNM_DOTMATCH).count == 0
+      # search for dotenvs in either fastlane or the current dir
+      search_paths = [FastlaneCore::FastlaneFolder.path, '.'].compact
+      base_path = search_paths.find do |dir|
+        Dir.glob(File.join(dir, '*.env*'), File::FNM_DOTMATCH).count > 0
+      end
+      return unless base_path
       require 'dotenv'
 
       Actions.lane_context[Actions::SharedValues::ENVIRONMENT] = env if env


hjanuschka:
@lacostej  i did the rubocop freeze as a standalone PR https://github.com/fastlane/fastlane/pull/7873
once this is merged , could you rebase and remove the lock in this PR  (to keep it seperated)

hjanuschka (approved):
🚀

PR #7881, opened 2017-01-16
mask all sensitive options by default when printing summary table

hjanuschka:
mask all sensitive options when printing summary tables.
required in: https://github.com/fastlane/fastlane/pull/7880

bildschirmfoto 2017-01-16 um 16 00 48


mfurtak (changes requested):
Can we cover this with a test to prevent regressions?

hjanuschka:
@mfurtak good idea, done.

mfurtak (approved):

KrauseFx (approved):
Beautiful, thanks @hjanuschka 👍

PR #7883, opened 2017-01-16
fix and bump rubocop

hjanuschka:
fixes new issues with rubocop => 0.47 👮

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,6 +38,10 @@ Style/NumericLiteralPrefix:
 Style/TernaryParentheses:
   Enabled: false
 
+# sometimes it is usefull to have those empty methods
+Style/EmptyMethod:
+  Enabled: false
+
 # It's better to be more explicit about the type
 Style/BracesAroundHashParameters:
   Enabled: false
--- a/fastlane/lib/fastlane/actions/appaloosa.rb
+++ b/fastlane/lib/fastlane/actions/appaloosa.rb
@@ -83,7 +83,11 @@ module Fastlane
       def self.get_screenshots(screenshots_path, locale, device)
         get_env_value('screenshots').nil? ? locale = '' : locale.concat('/')
         device.nil? ? device = '' : device.concat('-')
-        !screenshots_path.strip.empty? ? screenshots_list(screenshots_path, locale, device) : nil
+        if !screenshots_path.strip.empty?
+          return screenshots_list(screenshots_path, locale, device)
+        else
+          return nil
+        end
       end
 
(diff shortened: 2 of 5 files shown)


.rubocop.yml
```diff
@@ -38,6 +38,10 @@ Style/NumericLiteralPrefix:
 Style/TernaryParentheses:
   Enabled: false
 
+# sometimes it is usefull to have those empty methods
```
lacostej: Here I think rubocop just wants us to do something like:

```
      def self.return_value; end
```

I am not against it. It's a few files to change.

fastlane/lib/fastlane/actions/appaloosa.rb
```diff
@@ -83,7 +83,11 @@ def self.get_screenshots_links(api_key, store_id, screenshots_path, locale, devi
       def self.get_screenshots(screenshots_path, locale, device)
         get_env_value('screenshots').nil? ? locale = '' : locale.concat('/')
         device.nil? ? device = '' : device.concat('-')
-        !screenshots_path.strip.empty? ? screenshots_list(screenshots_path, locale, device) : nil
+        if !screenshots_path.strip.empty?
```
lacostej: I think it meant it preferred to invert the condition into something like
````
screenshots_path.strip.empty? ? nil : screenshots_list(screenshots_path, locale, device)
````

**Review by lacostej — COMMENTED** [review]
@hjanuschka see my comments! Thanks for the PR

File: fastlane/lib/fastlane/actions/ipa.rb [hunk 3]
```diff
@@ -1,5 +1,4 @@
 # rubocop:disable Lint/AssignmentInCondition
-# rubocop:disable Style/Next
```
mfurtak: Remove matching tag from end of file

.rubocop.yml
hjanuschka: yeah `-a` would fix it anyway 👍  i am not a fan of it, and i thought that i saw some other rules  like that.

mfurtak: I am in favor of having this check disabled. :+1: I don't find the one-line style to be major win.

fastlane/lib/fastlane/actions/ipa.rb
hjanuschka: done - thx

fastlane/lib/fastlane/actions/appaloosa.rb
hjanuschka: 👍  thx - fixed

mfurtak (approved):

PR #7659, opened 2016-12-23
Added support for providing utf flag to xcpretty via gym/xcodebuild

ghost:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:

- [✔︎] Run `bundle exec rspec` from the root directory.
- [✔︎] Run `bundle exec rubocop -a` to ensure the code style is valid
- [✔︎] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [✔︎] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

This change adds support for passing the 'utf' flag to xcpretty, which is useful when you don't have strict control over your bash environment to apply the usual encoding fix in .profile.

PR #7891, opened 2017-01-17
CONTINUE: Added support for providing utf flag to xcpretty via gym/xcodebuild

hjanuschka:
continuation of: https://github.com/fastlane/fastlane/pull/7659
rebased master, commits preserved.

original-PR body:

-----------

This change adds support for passing the 'utf' flag to xcpretty, which is useful when you don't have strict control over your bash environment to apply the usual encoding fix in .profile.

-----------

/cc @mfurtak @mgrebenets @lyptt


mfurtak (approved):
👍

PR #7904, opened 2017-01-18
Snapshot store screenshots on failure

mbalex:
### Checklist

### Description
The flag store_screenshots_on_failure was added to allow screenshots to be collected even if tests failed. Instead of assuming the error code 65 means that the build failed, we now assume that the tests failed and therefore allow to get the screenshots that were made. 

If the flag store_screenshots_on_failure is not set, the behavior of snapshot is the same as before.
If the flag store_screenshots_on_failure is set, we collect and generate the report before raising an error.

Due to complexity issues rubocop reported (after my additions), the "launch" methods was split into several smaller ones.

### Motivation and Context
We wanted to make screenshots of failing tests. We added snapshot to a XCTest teardown and even if a test failed we now have a screenshot of the screen that the error occurs in.

Our test was adding a snapshot call to tearDown of a XCTestCase and then calling XCTFail in a test method. With the flag on, it behaved as before and no screenshots were collected. With the flag on, screenshots will get collected and a report generated and still the snapshot step will raise an exception.


ohwutup:
Hey @mbalex! Thank you so much for the PR! I am a bit hesitant to approve this PR because it seems like a lot of code change to add functionality that is a bit out of scope of _snapshot_ given that the tool's intent is to generate screenshots for distribution to the AppStore rather than capturing failed test output. I think we can work towards this if we take smaller, more simple steps. I am going to close this for now, but we are definitely open to discussion about this so please do not hesitate to keep discussion flowing! 🚀 🐙

PR #7909, opened 2017-01-19
Parametrize ARGV to remove spurious warnings while running tests

lacostej:
I hate it when the output is like this:

image

Found a much cleaner way


hjanuschka (approved):
we should probably start moving all other shared spec code into that new class :)

Issue #7907, opened 2017-01-18
Feature request: `launch_arguments` equivalent for screengrab


hblumberg:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
We have been working on automating screenshots of the Khan Academy Android app using screengrab. It would be very helpful to be able to provide additional arguments to instrumentation tests. This feature seems to be available for the screenshot tool with `launch_arguments`.

Ideally, we would add something like:
`launch_arguments ['-username hblumberg']`
to our `Screengrabfile` and would be able to access those arguments in our tests with something like:
`final String username = InstrumentationRegistry.getArguments().getString("username");`

Thank you so much for creating the screengrab tool - we're so excited to use it!

hjanuschka:
@hblumberg thats a good idea, i did a PR: https://github.com/fastlane/fastlane/pull/7911 maybe you can give it a try and confirm that it works?

KrauseFx:
Yeah 👍

hblumberg:
Thank you so much for the incredibly quick response and PR! We look forward to using this feature!

PR #7911, opened 2017-01-19
add support for user supplied custom launch_arguments

hjanuschka:
as requested in: https://github.com/fastlane/fastlane/issues/7907
support for user specified launch arguments, i think the feature is really usefull in many use-cases (and goes along with `snapshot`)

set launch arguments via Screengrabfile:
```ruby
launch_arguments: ["username hjanuschka", "build_type hero"]
```

get it in your test class

```java
{
    Bundle extras = InstrumentationRegistry.getArguments();
    String peerID = null;

    if ( extras != null ) {
        if ( extras.containsKey ( "username" ) ) {
            username = extras.getString("username");
            System.out.println("Username: " + username);
        } else {
            System.out.println("No username in extras");
        }
    } else {
        System.out.println("No extras");
    }
}
```

i got the sample from here: http://stackoverflow.com/a/34726783/1945875

tested the map/join with:

```ruby
launch_arguments = ["username asdfasf", "somethingelse asdf"]
puts launch_arguments.map { |item| '-e ' + item + ' ' }.join.to_s
exit
```

@asfalcone, @mfurtak  could you help me with review?


screengrab/lib/screengrab/runner.rb
```diff
@@ -219,6 +219,7 @@ def run_tests(device_serial, test_classes_to_use, test_packages_to_use)
                               "-e endingLocale #{@config[:ending_locale].tr('-', '_')}"]
         instrument_command << "-e class #{test_classes_to_use.join(',')}" if test_classes_to_use
         instrument_command << "-e package #{test_packages_to_use.join(',')}" if test_packages_to_use
+        instrument_command << launch_arguments.map { |item| '-e ' + item + ' ' }.join.to_s if launch_arguments
```
lacostej: Nitpick: wouldn't thi work as well?
```ruby
instrument_command << launch_arguments.map { |item| '-e ' + item }.join(' ') if launch_arguments
```

hjanuschka: feedback addressed

lacostej (approved):
:+1:

screengrab/spec/runner_spec.rb
```diff
@@ -24,6 +24,20 @@ def mock_adb_response(mock_response)
...
     let(:test_packages_to_use) { nil }
 
+    context "when launch arguments are specified" do
+      before do
+        config[:launch_arguments] = ["username hjanuschka", "build_type x500"]
+        config[:locales] = %w(en-US)
+        config[:ending_locale] = 'en-US'
+      end
+      it 'sets custom launch_arguments' do
+        # expect(FastlaneCore::CommandExecutor).to receive(:execute).with(/hjanuschka/)
```
mfurtak: Please remove dead code 👍

mfurtak:
Nice feature! It would be great if we could add information about it to the [docs](https://github.com/fastlane/docs/blob/master/docs/getting-started/android/screenshots.md) so that people can learn about it

hjanuschka:
@mfurtak removed the dead code in #7917, and i ll do a docs PR

Issue #7899, opened 2017-01-18
Fastlane has a misleading error when unable to find a variable


lacostej:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I am setting up a variable in a lane, and trying to access it in the `error` handler. It fails with a misleading error:
```
/Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane_core/lib/fastlane_core/ui/interface.rb:145:in `user_error!': [!] Could not find action or lane 'foo'. Check out the README for more details: https://github.com/fastlane/fastlane/tree/master/fastlane (FastlaneCore::Interface::FastlaneError)
	from /Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane_core/lib/fastlane_core/ui/ui.rb:1
```
##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

### Captured Output

Command Used: `test1`
<details><summary>Output/Log</summary>

```

DEBUG [2017-01-18 09:02:17.22]: Checking if there are any plugins that should be loaded...
INFO [2017-01-18 09:02:17.23]: Driving the lane 'test1' 🚀
DEBUG [2017-01-18 09:02:17.23]: /Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane/lib/fastlane/runner.rb:142:in `trigger_action_by_name'
/Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane/lib/fastlane/fast_file.rb:146:in `method_missing'
Fastfile:12:in `block in parsing_binding'
/Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane/lib/fastlane/runner.rb:64:in `call'
/Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane/lib/fastlane/runner.rb:64:in `block in execute'
/Users/lacostej/.rvm/gems/ruby-2.2.2/gems/fastlane-2.9.0/fastlane/lib/fastlane/runner.rb:59:in `chdir'
[...]

PR #7915, opened 2017-01-19
Reword the missing reference error to handle variables (#7899). Also add a missing require to be able to run a spec standalone

lacostej:


fastlane/lib/fastlane/plugins/plugin_manager.rb
```diff
@@ -1,3 +1,5 @@
+require "fastlane/cli_tools_distributor"
```
hjanuschka: is that intended to be here?

hjanuschka (approved):

PR #7916, opened 2017-01-19
[sigh] #7330 - Ignore blacklisted keys when using --use-app-entitlements

### LINKED ISSUE #798: Store submission fails with iCloud entitlements
When a provisioning profile has iCloud entitlements enabled, the buck signed binary will not pass store submission. A typical iCloud provisioning profile will have entitlements similar to:

```lang=xml
<key>Entitlements</key>
<dict>
    <key>keychain-access-groups</key>
    <array>
        <string>4ZX4Z3MVHG.*</string>
    </array>
    <key>application-identifier</key>
    <string>4ZX4Z3MVHG.com.appleseedinc.MyProject</string>
    <key>com.apple.developer.ubiquity-kvstore-identifier</key>
    <string>4ZX4Z3MVHG.*</string>
    <key>com.apple.developer.icloud-services</key>
    <string>*</string>
    <key>com.apple.developer.icloud-container-environment</key>
    <array>
        <string>Development</string>
        <string>Production</string>
    </array>
    <key>com.apple.developer.icloud-container-identifiers</key>
    <array>
        <string>iCloud.com.appleseedinc.MyProject</string>
        <string>iCloud.com.appleseedinc.container1</string>
    </array>
    <key>com.apple.developer.icloud-container-development-container-identifiers</key>
    <array>
        <string>iCloud.com.appleseedinc.MyProject</string>
        <string>iCloud.com.appleseedinc.container1</string>
    </array>
    <key>com.apple.developer.ubiquity-container-identifiers</key>
    <array>
        <string>iCloud.com.appleseedinc.MyProject</string>
        <string>iCloud.com.appleseedinc.container1</string>
    </array>
    <string>4ZX4Z3MVHG</string>
    <key>aps-environment</key>
    <string>production</string>
</dict>
```

The default entitlements handling in buck will combine the specified entitlements file with the entitlements in the provisioning profile. This will result in both the `com.apple.developer.icloud-container-development-container-identifiers` and `com.apple.developer.ubiquity-kvstore-identifier` keys being added to the signing entitlements. Trying to submit a binary with either of these present will result in the following iTunes store errors:

```
ERROR ITMS-90045: "Invalid Code Signing Entitlements. Your application bundle's signature contains code signing entitlements that are not supported on iOS. Specifically, key com.apple.developer.icloud-container-development-container-identifiers in Payload/Blah.app is not supported."
```

and:

```
ERROR ITMS-90211: Invalid Code Signing Entitlements. The signature for your app bundle contains entitlement values that are not supported. For the com.apple.developer.ubiquity-kvstore-identifier entitlement, the value must start with the prefix provided by Apple in the provisioning profile, followed by characters that are uppercase or lowercase Roman letters [A-Z, a-z], the digits 0 through 9, dot, or hyphen, and not contain any wildcard characters. Specifically, value 4ZX4Z3MVHG.* for the key com.apple.developer.ubiquity-kvstore-identifier in Payload/Blah.app is not supported.
```
 
I'm not sure where the behaviour of merging the entitlements came from, is this mirroring what Xcode would do? It seems more correct to not merge the entitlements, would this result in a different error though, do we always need to merge certain keys from the provisioning profile?

### LINKED ISSUE #7330: [sigh] Blacklisted keys are aded to resigned app when using app entitlements
### New Issue Checklist

### Issue Description

sigh version: `0.12.1`

When we use `sigh resign` with `--use-app-entitlements` option we end up copying some blacklisted keys from provisioning profile into the final app entitlements.

I have faced this issue just now with an app using iCloud.

Facebook's `buck` tool also has faced the same issue: https://github.com/facebook/buck/issues/798
They went into all the details, looking into underlying private Apple framework to figure out the hidden logic. So they managed to get the list of the keys that shouldn't be copied from provisioning profile to resigned app's entitlements: https://github.com/facebook/buck/pull/802/files.

Here's the full list:
```
+        "com.apple.developer.icloud-container-development-container-identifiers",
 +        "com.apple.developer.icloud-container-environment",
 +        "com.apple.developer.icloud-container-identifiers",
 +        "com.apple.developer.icloud-services",
 +        "com.apple.developer.restricted-resource-mode",
 +        "com.apple.developer.ubiquity-container-identifiers",
 +        "com.apple.developer.ubiquity-kvstore-identifier",
 +        "inter-app-audio",
 +        "com.apple.developer.homekit",
 +        "com.apple.developer.healthkit",
 +        "com.apple.developer.in-app-payments",
 +        "com.apple.developer.associated-domains",
 +        "com.apple.security.application-groups",
 +        "com.apple.developer.maps",
 +        "com.apple.developer.networking.vpn.api",
 +        "com.apple.external-accessory.wireless-configuration"
```

So we should update `resigh.sh`'s _use app entitlements_ path with fix to avoid copying over blacklisted keys.

That or recommend people to use `buck` for resigning? ;)

P.S.

I know that `--use-app-entitlements` is a last resort option and ideally the explicit entitlements file should be used for resigning, but it seems that it is still used by people in the iOS community. The fact that Facebook had to go into lots of digging  to fix issue also says something.
I am one of the unlucky ones who has to stick with this way of resigning for now 😕

mgrebenets:
option for sign resign

### Checklist

### Description
Add a blacklist of keys which should not be kept while merging entilements from new provisioning profile and from app's embedded provisioning profile.

Apparently that's exactly what Apple's private development frameworks are doing: _merging entitlements_. Thanks to Facebook engineers we know what to fix: https://github.com/facebook/buck/issues/798.

### Motivation and Context
Fixes #7330
We have an internal app which uses iCloud. While resigning with `--use-app-entitlements` keys we couldn't upload resigned .ipa because it would have "invalid" keys. Using this modified resigner fixes the problem.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -715,6 +715,33 @@ function resign {
         log "Replacing old bundle ID '$OLD_BUNDLE_ID' with new bundle ID '$NEW_BUNDLE_ID' in patched entitlements"
         sed -i .bak "s/$OLD_BUNDLE_ID/$NEW_BUNDLE_ID/g" "$PATCHED_ENTITLEMENTS"
 
+        log "Removing blacklisted keys from patched profile"
+        # See https://github.com/facebook/buck/issues/798 and https://github.com/facebook/buck/pull/802/files
+        BLACKLISTED_KEYS=(\
+            "com.apple.developer.icloud-container-development-container-identifiers" \
+            "com.apple.developer.icloud-container-environment" \
+            "com.apple.developer.icloud-container-identifiers" \
+            "com.apple.developer.icloud-services" \
+            "com.apple.developer.restricted-resource-mode" \
+            "com.apple.developer.ubiquity-container-identifiers" \
+            "com.apple.developer.ubiquity-kvstore-identifier" \
+            "inter-app-audio" \
+            "com.apple.developer.homekit" \
+            "com.apple.developer.healthkit" \
+            "com.apple.developer.in-app-payments" \
+            "com.apple.developer.associated-domains" \
+            "com.apple.security.application-groups" \
+            "com.apple.developer.maps" \
+            "com.apple.developer.networking.vpn.api" \
+            "com.apple.external-accessory.wireless-configuration"
+        )
+
+        # Blacklisted keys must not be included into new profile, so remove them from patched profile
+        for KEY in ${BLACKLISTED_KEYS[@]}; do
+            log "Removing blacklisted key: $KEY"
+            PlistBuddy -c "Delete $KEY" "$PATCHED_ENTITLEMENTS" 2>/dev/null
+        done
+
         log "Resigning application using certificate: '$CERTIFICATE'"
         log "and patched entitlements:"
         log "$(cat "$PATCHED_ENTITLEMENTS")"


hjanuschka (approved):
this is a greate change, i think 🚀  
and nice explanation!

mfurtak:
Nice find!

Do you have any thoughts about the other pieces of what was described in that Buck ticket?

> Otherwise we just work out what it is doing and replicate the logic, but it is unfortunately doing quite a lot more than you would expect, including adding in defaults, replacing wildcards, blacklisting a whole bunch of keys (which presumably change with every tooling upgrade) and doing some verification of some of the keys.

Namely, "adding in defaults" and "replacing wildcards"? Are those things we need to be aware of and doing?

mgrebenets:
@mfurtak 

I don't have much thoughts about "adding in defaults", will try to look into Buck source to see what they mean.

With regards to "replacing wildcards" I think I have some idea.
That probably refers to the fact that in profile we always have wildcards for keys like `keychain-access-group`, but in the app entitlements we will have specific values, for example:

```
// profile
	<array>
		<string>TEAMID.au.com.company.*</string>
	</array>

// app
	<array>
		<string>TEAMID.au.com.company.app</string>
	</array>

```

This is the reason why `--use-app-entitlements` option was added a while ago - we had an issue with keychain access group messed up during resign.

mgrebenets:
Just a side-note, it would make sense to rewrite `resign.sh` in Ruby, but it would be too much of a breaking change. Maybe one day :)

hjanuschka:
@mgrebenets  i was thinking of porting the sh script to ruby a few days ago, and started with it, but hell i stopped it, its too much i think, atleast for a quick-win 👍  but in the end of the day it would be awesome to have this done in ruby.

mgrebenets:
@mfurtak do you mind if I merge this one and follow up with improvements in the future (if any found by then)?

mgrebenets:
I'll get it in and then investigate more to see what else can be improved.

PR #7923, opened 2017-01-19
Add reinstall_app config to screengrab to uninstall the APKs before run it

gersonmdesouza:
### Checklist

### Description
Adds config to screengrab called `reinstall_app` to uninstall APKs before running tests for each `locale` configured in the `Screengrabfile`.

### Motivation and Context
We have some configurations that are set the first time that the app runs and is locale bound. Even if we change the language of the device, most of the configurations are already set by the first run and will never change.

Following the configurations of iOS' `Snapshot`, that already features this configuration, we believe that it would be a good addition to `Screengrab` to have that as well.

### DIFF
--- a/screengrab/lib/screengrab/runner.rb
+++ b/screengrab/lib/screengrab/runner.rb
@@ -63,11 +63,7 @@ module Screengrab
 
       validate_apk(app_apk_path)
 
-      install_apks(device_serial, app_apk_path, tests_apk_path)
-
-      grant_permissions(device_serial)
-
-      run_tests(device_serial, test_classes_to_use, test_packages_to_use, @config[:launch_arguments])
+      run_tests(device_serial, app_apk_path, tests_apk_path, test_classes_to_use, test_packages_to_use, @config[:launch_arguments])
 
       number_of_screenshots = pull_screenshots_from_device(device_serial, device_screenshots_paths, device_type_dir_name)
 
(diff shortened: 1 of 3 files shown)


screengrab/lib/screengrab/runner.rb
```diff
@@ -210,27 +220,41 @@ def grant_permissions(device_serial)
...
-                                      print_command: true)
-
-        if @config[:exit_on_test_failure]
-          UI.user_error!("Tests failed", show_github_issues: false) if test_output.include?("FAILURES!!!")
-        else
-          UI.error("Tests failed") if test_output.include?("FAILURES!!!")
+        if @config[:reinstall_app]
+          uninstall_apks(device_serial, @config[:app_package_name], @config[:tests_package_name])
+          install_apks(device_serial, app_apk_path, tests_apk_path)
+          grant_permissions(device_serial)
```
hjanuschka: could this get bit simplified?

like trying to get rid of the `unless` above?

could this work (and removing the unless + the if here?):

```ruby
uninstall_apks(device_serial, @config[:app_package_name], @config[:tests_package_name]) if @config[:reinstall_app]

```

hjanuschka:
i really like this - lets wait for circle to work again (they seem to 503 right now)

gersonmendes:
@hjanuschka It's weird, CircleCI didn't run yet, can I trigger it manually?

hjanuschka (approved):

hjanuschka:
this is awesome! many thx for your contribution 🚀

Issue #7945, opened 2017-01-21
[scan] Add support for thread sanitizer


mgrebenets:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

This is a feature request.
`xcodebuild` has new `-enableThreadSanitizer` option available since Xcode 8 for Swift 3.0 and above projects.

```
-enableThreadSanitizer YES|NO       turn the thread sanitizer on or off
```

PR #7946, opened 2017-01-21
[scan] Add support for thread sanitizer

mgrebenets:
### Checklist

### Description
Add support for thread sanitizer
- Thread sanitizer can't be enabled together with address sanitizer
- Thread sanitizer is available only for Swift 3.x and above projects
  - We don't need to handle that though, xcodebuild will do it for us and xcpretty will parse the error

### Motivation and Context
It's a new feature as requested in #7945

### DIFF
--- a/fastlane/spec/actions_specs/xcodebuild_spec.rb
+++ b/fastlane/spec/actions_specs/xcodebuild_spec.rb
@@ -27,6 +27,7 @@ describe Fastlane do
             destination_timeout: 240,
             dry_run: true,
             enable_address_sanitizer: true,
+            enable_thread_sanitizer: false,
             enable_code_coverage: true,
             export_archive: true,
             export_format: 'ipa',
@@ -65,7 +66,7 @@ describe Fastlane do
           "-hideShellScriptEnvironment -jobs \"5\" -parallelizeTargets OTHER_CODE_SIGN_FLAGS=\"--keychain /path/to/My.keychain\" -project " \
           "\"MyApp.xcodeproj\" -resultBundlePath \"/result/bundle/path\" -scheme \"MyApp\" -sdk \"iphonesimulator\" -skipUnavailableActions -target " \
           "\"MyAppTarget\" -toolchain \"toolchain name\" -workspace \"MyApp.xcworkspace\" -xcconfig \"my.xcconfig\" -newArgument YES -enableAddressSanitizer " \
-          "\"YES\" -enableCodeCoverage \"YES\" | tee 'mypath/xcodebuild.log' | xcpretty --color --test"
+          "\"YES\" -enableThreadSanitizer \"NO\" -enableCodeCoverage \"YES\" | tee 'mypath/xcodebuild.log' | xcpretty --color --test"
         )
       end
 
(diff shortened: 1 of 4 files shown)


fastlane/spec/actions_specs/xcodebuild_spec.rb
```diff
@@ -27,6 +27,7 @@
             destination_timeout: 240,
             dry_run: true,
             enable_address_sanitizer: true,
+            enable_thread_sanitizer: true,
```
lacostej: this use cases seems a bit weird as if I understand `thread`  and `address sanitizer` can't be enabled together.

scan/lib/scan/test_command_generator.rb
```diff
@@ -37,6 +37,7 @@ def options
         options << "-resultBundlePath '#{result_bundle_path}'" if config[:result_bundle]
         options << "-enableCodeCoverage YES" if config[:code_coverage]
         options << "-enableAddressSanitizer YES" if config[:address_sanitizer]
```
lacostej: not an xcode expert, but would there be a use case where wants to use the CL to override the project's config, in which case we should not also pass the value NO when the config of `address_sanitizer` or `thread_sanitizer ` explicitly pass the value `false` ?

lacostej (changes requested):
A few corner cases that need clarification from my POV, otherwise nice PR!

scan/lib/scan/test_command_generator.rb
mgrebenets: Good point! Do you think same should apply to thread sanitizer and code coverage options too?

fastlane/spec/actions_specs/xcodebuild_spec.rb
mgrebenets: Agree. I'll just remove it from here, there are separate tests for address and thread sanitizer options anyways.

scan/lib/scan/test_command_generator.rb
lacostej: I would say that as a rule all options should reflect the user's explicit configuration. This should be a rule for all tools/wrappers.

fastlane/spec/actions_specs/xcodebuild_spec.rb
lacostej: 👍  Alternative it to have it set to false :)

scan/lib/scan/test_command_generator.rb
```diff
@@ -35,8 +35,9 @@ def options
         options << destination # generated in `detect_values`
         options << "-derivedDataPath '#{config[:derived_data_path]}'" if config[:derived_data_path]
         options << "-resultBundlePath '#{result_bundle_path}'" if config[:result_bundle]
-        options << "-enableCodeCoverage YES" if config[:code_coverage]
-        options << "-enableAddressSanitizer YES" if config[:address_sanitizer]
+        options << "-enableCodeCoverage #{config[:code_coverage] ? 'YES' : 'NO'}" unless config[:code_coverage].nil?
```
mgrebenets: Not sure if we need a helper like `yesno` mixin for boolean type or `humanize`-like method...
So far `predicate ? 'YES' : 'NO'` seems to be the way to "stringify" booleans.

fastlane/spec/actions_specs/xcodebuild_spec.rb
mgrebenets: 👍 Done, `false` it is.

scan/lib/scan/test_command_generator.rb
mgrebenets: Done. Removed default `false` value for all 3 options.
If users don't specify this option, we simply add no `-enable***` flag to `xcodebuild` and it uses whatever is already set in Xcode project or scheme.

mgrebenets: Small follow up, had to make all 3 optional too.

lacostej (approved):

lacostej:
@mgrebenets thanks a lot for the PR! 🚀

PR #7952, opened 2017-01-21
[Request for comment] Update gemspec to not support Ruby 2.4.0

KrauseFx:
I was able to install fastlane in Ruby 2.4.0 using the source code, but I'm unsure if everything is fixed already. This should be merged if we really don't support 2.4.0

I'm not sure how to best test this tbh

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://fastlane.tools"
   spec.license       = "MIT"
 
-  spec.required_ruby_version = '>= 2.0.0'
+  spec.required_ruby_version = ['>= 2.2.0', '< 2.4.0']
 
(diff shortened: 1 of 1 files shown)


lacostej:
> I'm unsure if everything is fixed already

what do you mean by that? Was there any breakages during the upgrade to 2.3.?

I think that everybody upgrading the runtime platform should be aware of the risks.

I see 2 issues applyinh this:

* it will prevent users from installing newer versions of fastlane and testing for potential issues 
* not sure if this change is retroactive or not (i.e. affects the whole gem or that particular version). If it affects the version, it might restrict users of ruby 2.4 to download fastlane version 2.9.0 or earlier. Causing more problems.

lacostej (commented):
See above.

KrauseFx:
Sounds legit, let's close it. There was an issue with the JSON gem, but it was fixed according to @orta 👍

PR #7953, opened 2017-01-22
Add support for --quiet option for swiftlint action

mgrebenets:
### Checklist

### Description
Add support for `swiftlint`'s `--quiet` option.

### Motivation and Context
Allow users to have quiet `swiftlint` runs to reduce console output.

### DIFF
--- a/fastlane/lib/fastlane/actions/swiftlint.rb
+++ b/fastlane/lib/fastlane/actions/swiftlint.rb
@@ -18,7 +18,7 @@ module Fastlane
         command << supported_option_switch(params, :quiet, "0.9.0", true)
 
         if params[:files]
-          if version < Gem::Version.new('0.5.1') and !Helper.test?
+          if version < Gem::Version.new('0.5.1')
             UI.user_error!("Your version of swiftlint (#{version}) does not support list of files as input.\nUpdate swiftlint using `brew update && brew upgrade swiftlint`")
           end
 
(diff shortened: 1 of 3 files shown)


hjanuschka (dismissed):

mgrebenets:
@hjanuschka Thanks for approving! I've added one last touch to it - check that `--quiet` is supported (needs to be at least `0.9.0`) and display important message otherwise.

hjanuschka:
like putting this:

```
if params[:quiet]
          if version < Gem::Version.new('0.9.0') and !Helper.test?
            UI.important("Your version of swiftlint (#{version}) does not support '--quiet' option.\nThe option will be ignored.\nUpdate swiftlint to 0.9.0 or above using `brew update && brew upgrade swiftlint`")
          else
            command << " --quiet"
          end
        end
```

into a method - returning "--quiet" or nothing, and do a `command << new_method_name(params)`.

this should work around the rubocop `PerceivedComplexity` problem

mgrebenets:
@hjanuschka Added `supported_option_switch` function, that now handles `--strict` and `--quiet` switches.

hjanuschka:
@mgrebenets that looks really nice, i think you have to add the action as an exception to the unused options spec here -> https://github.com/fastlane/fastlane/blob/master/fastlane/spec/unused_options_spec.rb#L4 to get CI to green.

hjanuschka (approved):

mfurtak:
This is great! If you're feeling up to it, it would be awesome to get a test case covering the "unhappy path" where the option is omitted. 👍

mgrebenets:
@mfurtak I have added 2 more specs for `strict: false` and `quiet: false` uses.
Is that what you mean?

It seems for the case where `quiet` or `strict` are omitted there is `"default use case"` spec.

mfurtak:
Ah, sorry for not being more specific... I was thinking about the case when you have asked for `--quiet`, but the current version of the lib is not new enough.

mfurtak (approved):
👍

Issue #7963, opened 2017-01-23
Gemspec version requirement for Bundler


summera:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Wondering if there is a specific reason for having bundler version set to `~>1.12` in the gemspec? Default bundler for my server environment is `1.9.6`. So curious if it's possible to lower this requirement if it's not necessary. I see that it was changed in https://github.com/fastlane/fastlane/pull/4722. Thanks!

KrauseFx:
Nice catch, thanks, just submitted https://github.com/fastlane/fastlane/pull/7964 👍

summera:
@KrauseFx thanks for the quick response! i'm wondering if it's possible to go even lower than `1.12`. For example: `spec.add_dependency 'bundler', '~> 1.0'`

KrauseFx:
Let's continue discussion on the PR 👍

PR #7964, opened 2017-01-23
Update bundler dependency to be more flexible

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/7963

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -53,7 +53,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'security', '= 0.1.3' # macOS Keychain manager, a dead project, no updates expected
   spec.add_dependency 'xcpretty-travis-formatter', '>= 0.0.3'
   spec.add_dependency 'dotenv', '>= 2.1.1', '< 3.0.0'
-  spec.add_dependency 'bundler', "~> 1.12" # Used for fastlane plugins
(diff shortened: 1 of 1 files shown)


endocrimes (approved):
LGTM

summera:
@KrauseFx possible to go lower than `1.12`? For example: `~> `1.0`?

KrauseFx:
@summera Why? We haven't tested it with a lower version, we use the bundler API directly when the user has plugins enabled for their project.

summera:
Got it. So is `>= 1.12` required for fastlane plugins to work or you're unsure? 

I'm in a situation where the default bundler version in my AWS Elastic Beanstalk environment is lower than `1.12` (1.9.6). So I'm wondering if it's possible to make the requirement less strict by allowing lower versions of bundler for people in similar situations.

KrauseFx:
@summera I'm unsure, I haven't tested it below that version, that's why we require it. If you can try using it with an older version, and make sure plugin installation and adding new plugins work, I'm ok with reducing the minimum version more 👍

KrauseFx:
We can merge this for now, as it's already an improvement like this 👍

summera:
@KrauseFx Cool sounds good 👍 . I'll get back to you on reducing the min version. Thanks again!

KrauseFx:
Thanks 👍

Issue #7965, opened 2017-01-24
ensure_xcode_version requires the xcode-install gem


streeter:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Installing Fastlane and not having the `xcode-install` gem installed as well, fails with the error:

```
fastlane-2.10.0/fastlane/lib/fastlane/helper/sh_helper.rb:36:in `popen': [!] No such file or directory - xcversion (Errno::ENOENT)
```

streeter:
Here's the error:

```
➜ /tmp/fastlane fastlane version
[17:38:56]: -------------------------------------------------
[17:38:56]: --- Step: Verifying required fastlane version ---
[17:38:56]: -------------------------------------------------
[17:38:56]: Your fastlane version 2.10.0 matches the minimum requirement of 1.106.2  ✅
[17:38:56]: ------------------------------
[17:38:56]: --- Step: default_platform ---
[17:38:56]: ------------------------------
[17:38:56]: Driving the lane 'ios version' 🚀
[17:38:56]: ----------------------------------
[17:38:56]: --- Step: ensure_xcode_version ---
[17:38:56]: ----------------------------------
[17:38:56]: $ xcversion selected
+------------------+-------------+
|          Lane Context          |
+------------------+-------------+
| DEFAULT_PLATFORM | ios         |
| PLATFORM_NAME    | ios         |
| LANE_NAME        | ios version |
+------------------+-------------+
[17:38:56]: No such file or directory - xcversion

+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 💥   | ensure_xcode_version                | 0           |
+------+-------------------------------------+-------------+

[17:38:56]: fastlane finished with errors
/Users/streeter/.rvm/gems/ruby-2.3.3@fastlane/gems/fastlane-2.10.0/fastlane/lib/fastlane/helper/sh_helper.rb:36:in `popen': [!] No such file or directory - xcversion (Errno::ENOENT)
[...]

PR #7972, opened 2017-01-24
ensure xcode-install gem is there

hjanuschka:
Fixes: https://github.com/fastlane/fastlane/issues/7965


KrauseFx (approved):
🎷

lacostej:
Cool. I wasn't aware of that feature.

Some questions:
* would it be useful to standardize on the required dependencies declaratively, e.g. having a `def required_gems` or similar on the plugin interface? 
* if yes would it be useful to run the dependency check earlier (before `run`)? E.g. at `Fastfile` parsing. This would cause early installation/failures. Easy if first point implemented.

hjanuschka:
@lacostej i like the idea of having a `required_gems` not sure how much work has to be done, and how to scan current codebase for places and gems.

Issue #7971, opened 2017-01-24
Nested resigning not working since 2.10.0


viktorasl:
Hello, something went wrong with resigning in 2.10.0 release.
Fastfile contains simple resigning code which resigns from AdHoc distribution to Development:
```ruby
resign(
    ipa: "AnApp.ipa",
    signing_identity: "Developer identity",
    provisioning_profile: {
      "an.app.main" => "path_to_an_app_provisioning_profile",
      "an.app.imessage.extension" => "path_to_an_app_imessage_extension_provisioning_profile"
    }
)
```
that worked fine till fastlane version 2.10.0 now it produces
```
No such file or directory
Encountered an error, aborting!
```
error.

salabaha:
I have the same issue. Here is my environment. 

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                       |
| ------------------- | ------------------------------------------- |
| OS                  | 10.12.2                                     |
| Ruby                | 2.3.3                                       |
| Bundler?            | false                                       |
| Git                 | git version 2.10.0                          |
| Installation Source | ~/.rvm/rubies/ruby-2.3.3/bin/fastlane       |
| Host                | Mac OS X 10.12.2 (16C67)                    |
| Ruby Lib Dir        | ~/.rvm/rubies/ruby-2.3.3/lib                |
| OpenSSL Version     | OpenSSL 1.0.2j  26 Sep 2016                 |
| Is contained        | false                                       |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version       | 8.2.1                                       |


### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |


### fastlane files:

**No Fastfile found**


**No Appfile found**


### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
| fastlane | 2.10.0  | ✅ Up-To-Date  |


### Loaded fastlane plugins:

**No plugins Loaded**


<details><summary><b>Loaded gems</b></summary>

| Gem                       | Version      |
| ------------------------- | ------------ |
| did_you_mean              | 1.0.0        |
| executable-hooks          | 1.3.2        |
| bundler-unload            | 1.0.2        |
| rubygems-bundler          | 1.4.4        |
| bundler                   | 1.14.2       |
[...]

PR #7975, opened 2017-01-24
Fix 'No such file or directory' error in sigh resign

mgrebenets:
Fix for #7971 

Unnecessary quotes around `${KEYCHAIN_FLAG}` were causing an issue
instead of passing to `codesign` as `--keychain $KEYCHAIN_PATH` it was passed as `"--keychain $KEYCHAIN_PATH"` causing an issue.

An unfortunate side-effect of fixing shellcheck issues.
I have revisited all other changes in that shellcheck update and ran resigning to make sure it works.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -764,7 +764,9 @@ function resign {
         log "Resigning application using certificate: '$CERTIFICATE'"
         log "and entitlements from provisioning profile: $NEW_PROVISION"
         cp -- "$TEMP_DIR/newEntitlements" "$APP_PATH/archived-expanded-entitlements.xcent"
-        /usr/bin/codesign ${VERBOSE} "${KEYCHAIN_FLAG}" -f -s "$CERTIFICATE" --entitlements "$TEMP_DIR/newEntitlements" "$APP_PATH"
+        # Must not qote KEYCHAIN_FLAG because it needs to be unwrapped and passed to codesign with spaces
+        # shellcheck disable=SC2086
+        /usr/bin/codesign ${VERBOSE} ${KEYCHAIN_FLAG} -f -s "$CERTIFICATE" --entitlements "$TEMP_DIR/newEntitlements" "$APP_PATH"
         checkStatus
     fi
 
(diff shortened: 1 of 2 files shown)


hjanuschka (approved):

mgrebenets:
ha :) so now shellcheck comlains abuout that change asking to double quote it :)

will add disable pragma

sigh/lib/assets/resign.sh
```diff
@@ -764,7 +764,9 @@ function resign {
         log "Resigning application using certificate: '$CERTIFICATE'"
         log "and entitlements from provisioning profile: $NEW_PROVISION"
         cp -- "$TEMP_DIR/newEntitlements" "$APP_PATH/archived-expanded-entitlements.xcent"
-        /usr/bin/codesign ${VERBOSE} "${KEYCHAIN_FLAG}" -f -s "$CERTIFICATE" --entitlements "$TEMP_DIR/newEntitlements" "$APP_PATH"
+        # Must not qote KEYCHAIN_FLAG because it needs to be unwrapped and passed to codesign with spaces
+        # shellcheck disable=SC2086
+        /usr/bin/codesign ${VERBOSE} ${KEYCHAIN_FLAG} -f -s "$CERTIFICATE" --entitlements "$TEMP_DIR/newEntitlements" "$APP_PATH"
```
mfurtak: Is `"$CERTIFICATE"` going to have the same problem?

asfalcone:
Does https://github.com/fastlane/fastlane/pull/7975/files#diff-7b2948da13b0a146580ebc7c06533c87L554 need to change as well?

mgrebenets:
@asfalcone You mean the change to `sigh/lib/sigh/resign.rb`?

That one puts ipa name to be the last. That's more safe for parsing arguments in shell.
I looks for all the `a` `--aaa` flags and if none works - then assumes that's the output file name.
So I've put it to be last.

mgrebenets:
@asfalcone I'm running more tests on all kinds of edge cases for resigning. Will update you soon. To be confident about releasing 2.10.1 patch.

mgrebenets:
@asfalcone Saw it! One more quoting missing. PR is on the way.

PR #7980, opened 2017-01-24
Fix the type of option reinstall_app in Screengrab tool

gersonmdesouza:
### Checklist

### Description
Add the property `is_string` with value `false` to `reinstall_app` Screengrab options that need to have it.

### Motivation and Context
This Screengrab option is a flag and need to be boolean type. If the property `is_string` is not set to `false`, Fastlane doesn't recognize it as boolean and crash with the error:
>`'reinstall_app' value must be a String! Found TrueClass instead.`

### DIFF
--- a/screengrab/lib/screengrab/options.rb
+++ b/screengrab/lib/screengrab/options.rb
@@ -116,7 +116,8 @@ module Screengrab
         FastlaneCore::ConfigItem.new(key: :reinstall_app,
                                      env_name: 'SCREENGRAB_REINSTALL_APP',
                                      description: "Enabling this option will automatically uninstall the application before running it",
-                                     default_value: false)
+                                     default_value: false,
+                                     is_string: false)
       ]
     end
   end


gersonmendes:
@hjanuschka Hey, I open a Pull Request [(#7923)](https://github.com/fastlane/fastlane/pull/7923) that adds `reinstall_app` flag to Screengrab. In the file `options.rb` I copied the block of code of the flag `exit_on_test_failure` and change it to fit my new flag. But now, when I run Fastlane Screengrab, there is an error with this flag.
>'reinstall_app' value must be a String! Found TrueClass instead.

So, I open this Pull Request adding `is_string` with value `false`. Can you tell me why the option `exit_on_test_failure` doesn't have it, and why I need to put it on `reinstall_app`?

mfurtak:
@gersonmendes Thanks for putting this together.

Your fix looks correct to me, and my belief is that the other option is incorrect as well. As to the question of why the other is not causing a problem... I'm not sure without investigating more deeply.

I'm going to accept your change, and then we can make a new PR for the other option. 👍

mfurtak (approved):

gersonmendes:
@mfurtak Thanks for the reply.

I will try to investigate `exit_on_test_failure` flag, but I agree with you, the other option seems to be incorrect as well. I can create a new PR fixing the other option if you want, just let me know :)

hjanuschka:
i think `exit_on_test_failure` is faulty too - but it works (seems to) as it has a `default_value:true` 
but a pr to improve it would be nice.

mfurtak:
I've adjusted the other option here: https://github.com/fastlane/fastlane/pull/7981

PR #7982, opened 2017-01-24
[sigh] Fix shell script quoting issue - follow up

mgrebenets:
Fix for #7971 (follow up)

Unnecessary quotes around `${KEYCHAIN_FLAG}` were causing an issue
instead of passing to `codesign` as `--keychain $KEYCHAIN_PATH` it was passed as `"--keychain $KEYCHAIN_PATH"` causing an issue.

An unfortunate side-effect of fixing shellcheck issues.
I have revisited all other changes in that shellcheck update and ran resigning to make sure it works.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -551,7 +551,10 @@ function resign {
         do
             if [[ "$framework" == *.framework || "$framework" == *.dylib ]]
             then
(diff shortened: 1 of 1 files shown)


hjanuschka (approved):

mgrebenets:
@asfalcone Thanks for picking it up!
I did miss one more quoting issue in the https://github.com/fastlane/fastlane/pull/7975 fix.

mgrebenets:
To be confident (this time) I ran the script on our main app, which has both embedded apps and extensions, as well as embedded Swift frameworks.

PR #7994, opened 2017-01-25
Fix screengrab to doesn’t exit if the app could not be uninstalled

gersonmdesouza:
### Checklist

### Description
Using the new Screengrab flag `reinstall_app`, Fastlane will uninstall the APKs for each locale configured in the Screengrabfile. Fastlane should not exit if APKs could not be uninstalled, so I remove the code that the runner checks the output of `adb uninstall` command.

### Motivation and Context
At the first run of Screengrab, the APKs may not be installed yet, so the uninstall command will fail.
After the first run, the APK will be installed and the uninstall command will success normally.

### DIFF
--- a/screengrab/lib/screengrab/runner.rb
+++ b/screengrab/lib/screengrab/runner.rb
@@ -187,16 +187,14 @@ module Screengrab
 
     def uninstall_apks(device_serial, app_package_name, tests_package_name)
       UI.message 'Uninstalling app APK'
-      apk_uninstall_output = run_adb_command("adb -s #{device_serial} uninstall #{app_package_name}",
-                                             print_all: true,
-                                             print_command: true)
-      UI.user_error! "App APK could not be uninstalled" if apk_uninstall_output.include?("Failure [")
+      run_adb_command("adb -s #{device_serial} uninstall #{app_package_name}",
+                      print_all: true,
+                      print_command: true)
 
       UI.message 'Uninstalling tests APK'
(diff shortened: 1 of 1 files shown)


screengrab/lib/screengrab/runner.rb
```diff
@@ -187,16 +187,14 @@ def install_apks(device_serial, app_apk_path, tests_apk_path)
...
+      run_adb_command("adb -s #{device_serial} uninstall #{app_package_name}",
+                      print_all: true,
+                      print_command: true)
 
       UI.message 'Uninstalling tests APK'
-      apk_uninstall_output = run_adb_command("adb -s #{device_serial} uninstall -r #{tests_package_name}",
-                                             print_all: true,
-                                             print_command: true)
-      UI.user_error! "Tests APK could not be uninstalled" if apk_uninstall_output.include?("Failure [")
+      run_adb_command("adb -s #{device_serial} uninstall -r #{tests_package_name}",
```
mfurtak: Why is this command `uninstall -r` while the other is just `uninstall`?

gersonmendes: The option `-r` is wrong. This code is a copy-and-paste from the function that install the APKs. In `install` command, the option `-r` is to reinstall the APK keeping the data. In `uninstall` command, the option `-r` doesn't exists. I will amend my commit fixing this.

mfurtak (approved):

mfurtak:
Thanks!

PR #8000, opened 2017-01-25
Enable enterprise in-house profiles for match

KrauseFx:
Lots of people have already been using _match_ for enterprise profiles, and things have been working really well. Let's enable it, and users can explicitly say they want to use an "enterprise" profile. This PR also updated the documentation to mention to be sure the repo is private and the encryption uses a safe password.
Additionally we're now showing an error if the user tries to use _match_ for an enterprise profile for an account that doesn't support in-house

### DIFF
--- a/match/lib/match.rb
+++ b/match/lib/match.rb
@@ -21,30 +21,18 @@ module Match
   DESCRIPTION = "Easily sync your certificates and profiles across your team using git"
 
   def self.environments
-    envs = %w(appstore adhoc development)
-    envs << "enterprise" if self.enterprise?
-    return envs
-  end
-
-  # @return [Boolean] returns true if the unsupported enterprise mode should be enabled
-  def self.enterprise?
-    return FastlaneCore::Env.truthy?("MATCH_FORCE_ENTERPRISE")
-  end
-
-  # @return [Boolean] returns true if match should interpret the given [certificate|profile] type as an enterprise one
-  def self.type_is_enterprise?(type)
-    Match.enterprise? && type != "development"
+    return %w(appstore adhoc development enterprise)
   end
 
   def self.profile_type_sym(type)
-    return :enterprise if self.type_is_enterprise? type
+    return :enterprise if type == "enterprise"
     return :adhoc if type == "adhoc"
     return :appstore if type == "appstore"
     return :development
   end
 
   def self.cert_type_sym(type)
-    return :enterprise if self.type_is_enterprise? type
+    return :enterprise if type == "enterprise"
     return :development if type == "development"
     return :distribution
   end
--- a/match/lib/match/nuke.rb
+++ b/match/lib/match/nuke.rb
@@ -172,7 +172,7 @@ module Match
     def certificate_type(type)
       cert_type = Spaceship.certificate.production
       cert_type = Spaceship.certificate.development if type == :development
-      cert_type = Spaceship.certificate.in_house if Match.enterprise? && Spaceship.client.in_house?
+      cert_type = Spaceship.certificate.in_house if type == :enterprise
 
       cert_type
     end
@@ -180,7 +180,7 @@ module Match
     # The kind of provisioning profile we're interested in
     def profile_type(type)
       profile_type = Spaceship.provisioning_profile.app_store
-      profile_type = Spaceship.provisioning_profile.in_house if Match.enterprise? && Spaceship.client.in_house?
+      profile_type = Spaceship.provisioning_profile.in_house if type == :enterprise
       profile_type = Spaceship.provisioning_profile.ad_hoc if type == :adhoc
       profile_type = Spaceship.provisioning_profile.development if type == :development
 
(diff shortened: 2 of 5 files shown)


match/lib/match/nuke.rb
```diff
@@ -172,15 +172,15 @@ def delete_files!
     def certificate_type(type)
       cert_type = Spaceship.certificate.production
       cert_type = Spaceship.certificate.development if type == :development
-      cert_type = Spaceship.certificate.in_house if Match.enterprise? && Spaceship.client.in_house?
+      cert_type = Spaceship.certificate.in_house if type == :enterprise
```
hjanuschka: i am a bit confused is enterprise and in_house the same - if so, should all in_house things getting refactored and unified to enterprise ? 

and keep the `certificate.in_house` too, inside spaceship - for bc reasons?

KrauseFx: I assume that's because _spaceship_ uses the official names being used by the Apple Developer Portal. I don't feel like it's something we need to change, from the user's perspective it seems pretty solid.

hjanuschka: ok you are right.

asfalcone (approved):

match/lib/match.rb
```diff
@@ -21,30 +21,18 @@ module Match
   DESCRIPTION = "Easily sync your certificates and profiles across your team using git"
 
   def self.environments
-    envs = %w(appstore adhoc development)
-    envs << "enterprise" if self.enterprise?
-    return envs
-  end
-
-  # @return [Boolean] returns true if the unsupported enterprise mode should be enabled
-  def self.enterprise?
```
lacostej: there's no risk someone using these deleted methods from the outside?

match/lib/match.rb
```diff
@@ -21,30 +21,18 @@ module Match
...
-
-  # @return [Boolean] returns true if match should interpret the given [certificate|profile] type as an enterprise one
-  def self.type_is_enterprise?(type)
-    Match.enterprise? && type != "development"
+    return %w(appstore adhoc development enterprise)
   end
 
   def self.profile_type_sym(type)
-    return :enterprise if self.type_is_enterprise? type
+    return :enterprise if type == "enterprise"
```
lacostej: can we use `type.to_sym` now?

match/lib/match.rb
```diff
@@ -21,30 +21,18 @@ module Match
...
     return :adhoc if type == "adhoc"
     return :appstore if type == "appstore"
     return :development
   end
 
   def self.cert_type_sym(type)
-    return :enterprise if self.type_is_enterprise? type
+    return :enterprise if type == "enterprise"
```
lacostej: same

lacostej (commented):
Some comments

match/lib/match/nuke.rb
lacostej: and here we could `map` now.

match/lib/match/nuke.rb
```diff
@@ -172,15 +172,15 @@ def delete_files!
...
 
       cert_type
     end
 
     # The kind of provisioning profile we're interested in
     def profile_type(type)
       profile_type = Spaceship.provisioning_profile.app_store
-      profile_type = Spaceship.provisioning_profile.in_house if Match.enterprise? && Spaceship.client.in_house?
+      profile_type = Spaceship.provisioning_profile.in_house if type == :enterprise
       profile_type = Spaceship.provisioning_profile.ad_hoc if type == :adhoc
```
lacostej: Same here we could `map` now.

lacostej (commented):
a couple more

match/lib/match.rb
KrauseFx: Great point, however _match_ isn't designed as a tool to access methods and classes directly from the outside, as opposed to _spaceship_

match/lib/match.rb
KrauseFx: We could

match/lib/match/nuke.rb
KrauseFx: What do you mean with `.map`

lacostej: Something like
```ruby
{ 
  enterprise => Spaceship.provisioning_profile.in_house, ...
}[type]
```

KrauseFx:
Thanks @lacostej, agreed that we can improve the code style, however I merged this PR, so we can have a separate PR for the code style changes

lacostej:
was re-reading the PR, and noticed that in line

https://github.com/fastlane/fastlane/blob/39625edfd1bfed631b994b63a18e196483fccbbc/match/lib/match/nuke.rb#L54

It looks like we are going to collect  `:development` `prov_types` for `:enterprise` ones.

@KrauseFx ?

KrauseFx:
Just pushed https://github.com/fastlane/fastlane/pull/8065

Issue #8011, opened 2017-01-26
[Issue] Tried to load unspecified class: HTTP::Cookie when running fastlane spaceauth -u


dimmdesign:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

**`Tried to load unspecified class: HTTP::Cookie`**
```
/Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/class_loader.rb:97:in `find': [!] Tried to load unspecified class: HTTP::Cookie (Psych::DisallowedClass)
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/class_loader.rb:28:in `load'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:396:in `resolve_class'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:208:in `visit_Psych_Nodes_Mapping'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/visitor.rb:16:in `visit'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/visitor.rb:6:in `accept'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:32:in `accept'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:330:in `block in register_empty'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:330:in `each'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:330:in `register_empty'
	from /Users/jenkins/.rbenv/versions/2.3.1/lib/ruby/2.3.0/psych/visitors/to_ruby.rb:141:in `visit_Psych_Nodes_Sequence'
[...]

PR #8012, opened 2017-01-26
Fix spaceauth YAML spaceship cookie loading

hjanuschka:
introduced by:
  *  https://github.com/fastlane/fastlane/pull/7883 (rubocop rules update)

issues: 
  * https://github.com/fastlane/fastlane/issues/8011

### DIFF
--- a/spaceship/lib/spaceship/spaceauth_runner.rb
+++ b/spaceship/lib/spaceship/spaceauth_runner.rb
@@ -33,7 +33,7 @@ module Spaceship
       # Example:
       # name: DES5c148586daa451e55afb017aa62418f91
       # value: HSARMTKNSRVTWFlaF/ek8asaa9lymMA0dN8JQ6pY7B3F5kdqTxJvMT19EVEFX8EQudB/uNwBHOHzaa30KYTU/eCP/UF7vGTgxs6PAnlVWKscWssOVHfP2IKWUPaa4Dn+I6ilA7eAFQsiaaVT
-      cookies = YAML.safe_load(itc_cookie_content)
+      cookies = YAML.safe_load(itc_cookie_content, [HTTP::Cookie])
 
       # We remove all the un-needed cookies
       cookies.select! do |cookie|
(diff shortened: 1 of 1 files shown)


mfurtak:
When testing `spaceauth` from the command line, I was not able to reproduce the problem. I had to start an `irb` session and effectively run the `SpaceauthRunner` code in order to see it.

```
$ irb
>> require 'spaceship'
=> true
>> require 'spaceship/spaceauth_runner'
=> true
>> Spaceship::SpaceauthRunner.new(username: 'your@email.address').run
Logging into to iTunes Connect (michael.furtak@gmail.com)...
Successfully logged in to iTunes Connect
```

I added a commit that adjusts the `safe_load` call to account for all of the settings needed to make the call work for me.

mfurtak:
I should say that the account that I used has two-step auth turned on, but I don't think that's a requirement for testing this?

KrauseFx:
Code-wise this looks fine to me, however I'll have to have a way to reproduce the issue to say for sure that's a valid fix. Seems like @dimmdesign mentioned it didn't work for him?

hjanuschka:
i am pretty sure the fix is ok - here is stripped down sample to reproduce.


```
require 'http-cookie'

stored = "1.cookie"


# try to redo the things going on in client.rb
jar = HTTP::CookieJar.new
cookie = HTTP::Cookie.new("uid", "u12345", domain: 'example.org',
                                            for_domain: true,
                                            path: '/',
                                            max_age: 7*86400)

jar.add(cookie)
jar.save(stored, :yaml, session: true)


cookie_content = File.read(stored)

# load it -> redo things going on in spaceauth_runner
begin
  deserialized = YAML.safe_load(cookie_content)
  puts "Non whitelist works (UNEXPECTED)"
  puts deserialized.inpsect
rescue => ex
  puts "Non Whitelist does not work (EXPECTED)"
end


deserialized = YAML.safe_load(
         cookie_content,
        [HTTP::Cookie, Time], # classes whitelist
        [],                   # symbols whitelist
        true                  # allow YAML aliases
       )


puts deserialized.inspect
puts "with whitelist it works"
```

mfurtak (approved):

dimmdesign:
@KrauseFx @mfurtak @hjanuschka — 
I tried this again and it seems to be working with this fix. I didn't use the local version of fastlane correctly before, my bad. Confirming that `master` branch (since this PR merged) can create a session.

KrauseFx:
Cool, thanks for confirming @dimmdesign, we're about to ship a new release for this fix 👍

PR #8019, opened 2017-01-26
Allow our custom Spaceship::Base#inspect to work with object graphes containing circular references

lacostej:
for @snatchev

### DIFF
--- a/spaceship/lib/spaceship/base.rb
+++ b/spaceship/lib/spaceship/base.rb
@@ -217,20 +217,39 @@ module Spaceship
     #####################################################
 
     def inspect
-      inspectables = self.attributes
+      # To avoid circular references, we keep track of the references
+      # of all objects already inspected from the first call to inspect
+      # in this call stack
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/base.rb
```diff
@@ -217,20 +217,42 @@ def attributes
...
+      # in this call stack
+      # We use a Thread local storage for multi-thread friendliness
+      thread = Thread.current
+      tree_root = thread[:inspected_objects].nil?
+      thread[:inspected_objects] = Set.new if tree_root
 
-      value = inspectables.map do |k|
+      if thread[:inspected_objects].include? self
+        # already inspected objects have a default value
+        value = "~~DUPE~~"
```
snatchev: Ruby's convention for displaying a circular reference is
`#<Object ...>`

snatchev (approved):
Just a really small comment on the inspection value. 
This is a really nice solution. Thank you!

PR #8044, opened 2017-01-29
add attachment support to Mailgun action

igor-makarov:
### Checklist
- [✅ ] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [✅] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [✅] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [✅] I've updated the documentation if necessary.

### Description
Allow using `:attachment` param to add files to email.

### Motivation and Context
As a developer, I wanted to receive an attachment that contains my UI test results.

### DIFF
--- a/fastlane/lib/fastlane/actions/mailgun.rb
+++ b/fastlane/lib/fastlane/actions/mailgun.rb
@@ -86,7 +86,12 @@ module Fastlane
                                       env_name: "MAILGUN_REPLY_TO",
                                       description: "Mail Reply to",
                                       optional: true,
-                                      is_string: true)
+                                      is_string: true),
+          FastlaneCore::ConfigItem.new(key: :attachment,
+                                      env_name: "MAILGUN_ATTACHMENT",
+                                      description: "Mail Attachments",
+                                      optional: true,
+                                      is_string: false)
 
         ]
       end
@@ -114,6 +119,13 @@ module Fastlane
         unless options[:reply_to].nil?
           params.store(:"h:Reply-To", options[:reply_to])
         end
+
+        unless options[:attachment].nil?
+          attachment_filenames = [*options[:attachment]]
+          attachments = attachment_filenames.map { |filename| File.new(filename, 'rb') }
+          params.store(:attachment, attachments)
+        end
+
         RestClient.post "https://api:#{options[:apikey]}@api.mailgun.net/v3/#{sandbox_domain}/messages", params
         mail_template(options)
       end
@@ -158,7 +170,8 @@ module Fastlane
             message: "Mail Body",
             app_link: "http://www.myapplink.com",
             ci_build_link: "http://www.mycibuildlink.com",
-            template_path: "HTML_TEMPLATE_PATH"
+            template_path: "HTML_TEMPLATE_PATH",
+            attachment: "build/changelog.txt"
           )'
         ]
       end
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/mailgun.rb
```diff
@@ -86,7 +86,12 @@ def self.available_options
                                       env_name: "MAILGUN_REPLY_TO",
                                       description: "Mail Reply to",
                                       optional: true,
-                                      is_string: true)
+                                      is_string: true),
+          FastlaneCore::ConfigItem.new(key: :attachment,
+                                      env_name: "MAILGUN_ATTACHMENT",
+                                      description: "Mail Attachments",
```
KrauseFx: Could you update the description to describe better on what to pass? An array? One item as string? The data or the file path(s)?

fastlane/lib/fastlane/actions/mailgun.rb
```diff
@@ -114,6 +119,13 @@ def self.mailgunit(options)
         unless options[:reply_to].nil?
           params.store(:"h:Reply-To", options[:reply_to])
         end
+
+        unless options[:attachment].nil?
+          attachment_filenames = [*options[:attachment]]
+          attachments = attachment_filenames.map { |filename| File.new(filename, 'rb') }
```
KrauseFx: Why the rb?

fastlane/lib/fastlane/actions/mailgun.rb
```diff
@@ -158,7 +170,8 @@ def self.example_code
             message: "Mail Body",
             app_link: "http://www.myapplink.com",
             ci_build_link: "http://www.mycibuildlink.com",
-            template_path: "HTML_TEMPLATE_PATH"
+            template_path: "HTML_TEMPLATE_PATH",
+            attachment: "build/changelog.txt"
```
KrauseFx: Why build/changelog?

KrauseFx (changes requested):
Thanks for the contribution. File attachment seems like a great feature for fastlane :+1:

fastlane/lib/fastlane/actions/mailgun.rb
igor-makarov: I'm not very familiar with the distinction, but this is how it appears in [rest-client's docs](https://github.com/rest-client/rest-client#multipart) so I didn't think too much of it. 

From documentation it seems like the `b` option causes no CRLF conversion to be performed, which is obviously the case for attachments.

fastlane/lib/fastlane/actions/mailgun.rb
igor-makarov: You're right, this is less clear. Will revise.

fastlane/lib/fastlane/actions/mailgun.rb
igor-makarov: It's filename or array of filenames. I'll update.

igor-makarov:
I updated the requested stuff.

fastlane/lib/fastlane/actions/mailgun.rb
hjanuschka: its "read-binary", i think they do it this way in the documentation as if you don't use "rb" you get in troubles with non character files.

igor-makarov: So do you agree that it's right in this case?

hjanuschka (approved):

KrauseFx (approved):

KrauseFx:
Awesome, thanks you both! 👍

PR #8051, opened 2017-01-30
Allow ConfigItems to auto-convert Hash & Array as String inputs

AliSoftware:
This is especially useful for ConfigItems like xcargs

### Description

`ConfigItems` now allows a new pseudo-type `:shell_string`, which will behave like the standard `String` type, but:

* If passed an `Array` as input, items of the `Array` will be `to_s.shellescape` then joined using a space
* If passed a `Hash` as input, keys + values of the `Hash` will now be `to_s.shellescape`, key and value will be joined by "=" and those pairs then joined using a space

For now the only `ConfigItem` using `:shell_string` are the ones for `xcargs`.

### Motivation and Context

This is especially useful for `ConfigItems` like `xcargs`. This will now allow passing `Arrays` and `Hashes` to `xcargs`, like this:

```
lane :beta_prod do
  args = {
    :BUNDLE_APP_ID => 'com.mycompany.myapp.prod',
    :WEBSERVICE_PLATFORM => 'prod',
    :VERSION_NAME => 'Demo for Production',
    :APP_PROV_PROFILE => 'MyApp Prod',
    :WATCHAPP_PROV_PROFILE => 'MyApp WatchApp Prod',
    :WATCHEXT_PROV_PROFILE => 'MyApp WatchExt Prod',
  }
  gym(
    xcargs: args
  )
  crashlytics
end
```

I personally use `xcargs` a lot to help configure my Xcode project with User-Defined Build Settings and override them using Fastlane depending on the configuration I want to build.
Having to write a long and manually-escaped String for that is quite error-prone, while having a nicely formatted Hash instead helps clarity a lot!

### Alternate Solutions

* I considered applying the rule for all `ConfigItem` expecting a type `String`, instead of creating a new pseudo-type `:shell_string`, but then realised that some `ConfigItems` were not meant to be converted to shell strings (like `device` for example), so we should definitely only apply that conversion if explicitly specified in the `ConfigItem`.
* I considered adding a `allow_shell_conversion` boolean parameter to the `ConfigItem` initializer, instead of using the `type`  parameter and special symbol `:shell_string`, but then Rubocop told me that there were already too many parameters for this constructor. Hence the solution of `:shell_string` instead (see intermediate commits)

### Open Question

Are there any other `ConfigItem` for which this kind of conversion would make sense (for which `allow_shell_conversion` should be set to `true`), besides `xcargs`?


fastlane/spec/fixtures/plist/Info.plist
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
hjanuschka: this change always comes in once some newer xcode opens them. should we keep it?

fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -52,12 +52,13 @@ def initialize(key: nil, env_name: nil, description: nil, short_option: nil, def
       @default_value = default_value
       @verify_block = verify_block
       @is_string = is_string
-      @data_type = type
+      @data_type = (type == :shell_string) ? String : type
```
hjanuschka: how about making this a bit easier to read?

like 

```ruby
@data_type = type
if type == :shell_string
  @data_type = String
end
```

hjanuschka (changes requested):
that is super awesome! i left a few comments if you mind, lets discuss those points.

fastlane/spec/fixtures/plist/Info.plist
AliSoftware: I hesitated to commit it, but figured that if it was changed every time I ran tests, better commit it once and for all to avoid it on each PR?

fastlane_core/lib/fastlane_core/configuration/config_item.rb
AliSoftware: Will change it to this seem:

```
@data_type = type
@data_type = String if type == :shell_string
```

fastlane/spec/fixtures/plist/Info.plist
hjanuschka: yes let as commit it once.

fastlane_core/lib/fastlane_core/configuration/config_item.rb
hjanuschka: thats cool ;)

AliSoftware:
@hjanuschka Do you see any other `ConfigItem` which could benefit from this new `:shell_string` pseudo-type, besides `xcargs`?

I also wondered about naming it `:ShellString` to be more type-alike, rather than `:shell_string` which doesn't feel like a pseudo-type when written in snake-lowercase… even if it seems the ruby convention for symbols is to use snake_lower_case, for this specific case semantically equivalent to a pseudo-type, I feel it would still read better as `:ShellString`?

hjanuschka:
@AliSoftware hmmm - there is also `export_xcargs` on gym - which would profit from this change.
there comes an idea to my mind, how about doing a class `ShellString` and do the construction of the string in `.to_s`  - and later on pass `ShellString` instead of `:shell_string`.

if that is too hard / too much work - i think we should stick to snake case symbols.,

AliSoftware:
> how about doing a class ShellString and do the construction of the string in .to_s - and later on pass ShellString instead of :shell_string.

@hjanuschka I actually just thought of exactly the same, and was giving it a try.

But then I wondered, how would one pass that as an argument? I mean, wouldn't it mean that, at call-site, the user would have to wrap the parameter into a `ShellString` constructor?

I want it to be possible to write:
```ruby
gym(xcargs: { :a => 'b' })
```

But having a dedicated `ShellString` type, although elegant, would mean that to call `gym` with that I'd need to call:
```ruby
gym(xcargs: ShellString({ :a => 'b' }))
```
That doesn't seem as nice at the call site 😢 

Any workaround in mind?

AliSoftware:
@hjanuschka Ah, actually got it, see 2d233233545edad4a3f8ecb5c41d46bf6effab48 👍

hjanuschka:
@AliSoftware 🚀 that looks nice. any chances we can get a spec for the commandline parameters (e.g.: force them to be treated as strings?)

AliSoftware:
Hmmm… not sure what you mean (and even if I knew, where those kind of specs are in fastlane_core, as  assume there already are some existing ones for the thing you mention?)

hjanuschka:
@AliSoftware  seems to be in: `fastlane_core/spec/commander_generator_spec.rb`

AliSoftware:
Ok, reverted the last 4 commits given that introducing a new `ShellString` type would mean too much risk for the rest of the code not handling non-standard types. Reverting to the solution with the `:shell_string` symbol instead.

fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -1,6 +1,6 @@
 module FastlaneCore
   class ConfigItem
-    attr_accessor :key, :env_name, :description, :short_option, :default_value, :verify_block, :optional, :conflicting_options, :conflict_block, :deprecated, :sensitive
+    attr_accessor :key, :env_name, :description, :short_option, :default_value, :verify_block, :optional, :conflicting_options, :conflict_block, :deprecated, :sensitive, :allow_shell_conversion
```
KrauseFx: Since there are so many now, would you mind splitting each attribute to a new line, and adding a comment above each of these? (you can copy those from below)

KrauseFx (commented):
Looks good to me, one question/concern: Is this going to affect any existing setups, that already provide the e.g. `xcargs` parameter? Do they have to change anything?

AliSoftware:
>  one question/concern: Is this going to affect any existing setups, that already provide the e.g. xcargs parameter? Do they have to change anything?

Nope, and actually that's why it took multiple commits (and some force-pushed reverts you don't see here anymore), as I wanted to ensure retro-compatibility.

* Any `ConfigItem` that still uses `type: String` works same as before
* Any `ConfigItem` that now uses `type: :shell_string` is just the same as a `ConfigItem` using a `type: String`, + setting an additional internal flag to tell that providing an `Array` or `Hash` is also valid, and thus escaped as a string during the `auto_convert_value` call. That's actually exactly how it was implemented at first, but rubocop didn't allow me to add one more parameter to `initialize` given that there were already too many.

So any setup using `xcargs` can still provide them as a `String`. That PR only allows to alternatively provide a `Hash` or `Array` but still allows `String` as the primary type for that argument.

fastlane_core/lib/fastlane_core/configuration/config_item.rb
AliSoftware: nvm, those are different, as some parameters in `initialize` are not exposed as `attr_accessor` and vice-versa, so we'll have to keep both anyway.

Done in 085ae69f2119ea3677bbffc3ceeb1ffcd3fce8b5

KrauseFx (approved):
Beautiful, thank you @AliSoftware 🚀 This is amazing

fastlane_core/spec/commander_generator_spec.rb
```diff
@@ -79,6 +79,45 @@
         end
       end
 
+      describe ':shell_string flags' do
```
hjanuschka: ❤️

fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -1,6 +1,43 @@
...
+
+    # [String] Set if the option is deprecated. A deprecated option should be optional and is made optional if the parameter isn't set, and fails otherwise
+    attr_accessor :deprecated
+
+    # [Boolean] Set if the variable is sensitive, such as a password or API token, to prevent echoing when prompted for the parameter
+    attr_accessor :sensitive
+
+    # [Boolean] Set if the variable is to be converted to a shell-escaped String when provided as a Hash or Array
+    # Allows items expected to be strings used in shell arguments to b alternatively provided as a Hash or Array for better readability and auto-escaped for us.
+    attr_accessor :allow_shell_conversion
```
hjanuschka: ❤️

hjanuschka (approved):
this is a great change!!!

PR #8065, opened 2017-01-31
Disable `match nuke` for enterprise accounts

KrauseFx:
Also: Avoid asking for app identifier for `match nuke`

### DIFF
--- a/match/lib/match/nuke.rb
+++ b/match/lib/match/nuke.rb
@@ -13,7 +13,7 @@ module Match
 
       params[:workspace] = GitHelper.clone(params[:git_url], params[:shallow_clone], skip_docs: params[:skip_docs], branch: params[:git_branch])
 
-      had_app_identifier = self.params[:app_identifier]
+      had_app_identifier = self.params.fetch(:app_identifier, ask: false)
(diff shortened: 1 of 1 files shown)


match/lib/match/nuke.rb
```diff
@@ -13,7 +13,7 @@ def run(params, type: nil)
 
       params[:workspace] = GitHelper.clone(params[:git_url], params[:shallow_clone], skip_docs: params[:skip_docs], branch: params[:git_branch])
 
-      had_app_identifier = self.params[:app_identifier]
+      had_app_identifier = self.params.fetch(:app_identifier, ask: false)
```
lacostej: is this related to the change?

hjanuschka: i think its intended to disable the auto ask for the user. so in case `app_identifier` is not set - set `had_app_identifier` to nil instead of prompting user to set a app_identifier

hjanuschka (approved):

match/lib/match/nuke.rb
hjanuschka: the pr body states that this is to avoid asking for app_identifier

PR #8066, opened 2017-01-31
Cleanup match after #8000. Also add proper nuking of enterprise profiles (related #8065)

lacostej:
Note that the cleanup reveals what appears to be a broken test

### DIFF
--- a/match/lib/match/nuke.rb
+++ b/match/lib/match/nuke.rb
@@ -51,8 +51,10 @@ module Match
       UI.message "Fetching certificates and profiles..."
       cert_type = Match.cert_type_sym(type)
 
-      prov_types = [:development]
+      prov_types = []
+      prov_types = [:development] if cert_type == :development
       prov_types = [:appstore, :adhoc] if cert_type == :distribution
+      prov_types = [:enterprise] if cert_type == :enterprise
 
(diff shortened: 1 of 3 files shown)


match/lib/match/nuke.rb
```diff
@@ -170,21 +172,21 @@ def delete_files!
...
-      profile_type = Spaceship.provisioning_profile.development if type == :development
-
-      profile_type
+    def profile_type(prov_type)
+      {
+        appstore:    Spaceship.provisioning_profile.app_store,
+        development: Spaceship.provisioning_profile.development,
+        enterprise:  Spaceship.provisioning_profile.in_house,
+        adhoc:       Spaceship.provisioning_profile.ad_hoc
+      }[prov_type] ||= raise "Unknown prov. type #{prov_type}"
```
KrauseFx: Let's write `provisioning type`

hjanuschka (approved):

hjanuschka:
@lacostej see felix last comment i think it slipped through

lacostej:
@hjanuschka where? I think github was a bit slow to update the PR state.

hjanuschka:
yep - seemed to be github cache or whatever.

PR #8067, opened 2017-01-31
fix error reporting missing html decoding

hjanuschka:
should fix apple error html encoding.

from:
bildschirmfoto 2017-01-31 um 22 54 16

to:

bildschirmfoto 2017-01-31 um 23 34 12


KrauseFx (approved):
Oh yeah, this is a great improvement :rocket:

PR #8068, opened 2017-01-31
[spaceship] app, application finding case insensitive

### LINKED ISSUE #8039: Error caused by case sensitive bundle_id
```
+----------------+------------------------------+
|           Summary for produce 2.12.0          |
+----------------+------------------------------+
| username       | xxx@xxxx.com                 |
| app_identifier | com.xxx.catalog              |
| team_id        | 8G28592XXX                   |
| app_name       | app                          |
| skip_itc       | true                         |
| sku            | 1485526837                   |
| platform       | ios                          |
| language       | English                      |
| skip_devcenter | false                        |
+----------------+------------------------------+

[15:20:41]: Creating new app 'app' on the Apple Dev Center

Looking for related GitHub issues on fastlane/fastlane...

➡️  whitescreen after signing the app
   https://github.com/electron-userland/electron-osx-sign/issues/88 [open] 13 💬
   07 Nov 2016

➡️  Build Fail -- Ensure that the application's Info.plist contains a value for CFBundleIdentifier.
   https://github.com/fbsamples/f8app/issues/123 [open] 7 💬
   13 Oct 2016

➡️  Can't access app path ../ios/build/device/name.app : No such file or directory
   https://github.com/driftyco/ionic-cli/issues/1197 [open] 45 💬
   4 weeks ago

[!] Apple provided the following error info:
	An App ID with Identifier &#x27;com.xxx.catalog&#x27; is not available. Please enter a different string.
```
I have an App ID with bundle_id `com.xxx.Catalog` in my Developer Portal.
If I understand correctly bundle_ids could not be case sensitive. To avoid this error the `app_exists?` mehtod or even `Spaceship.app.find` has to remove upper and lowercase.

hjanuschka:
this PR makes app finding by bundle_id case insensitive.

  * Tunes -> application.find
  * Portal -> app.find

fixes: https://github.com/fastlane/fastlane/issues/8039#issuecomment-276304079

test:

```
# creates new app
bundle exec fastlane produce  --app_identifier some.new.app

# says it already exists, without PR it will produce a exception
bundle exec fastlane produce  --app_identifier some.nEW.app
```

i tested it with about 10 apps and it works - adp shows them with the casing used on first create, but validation if bundle_id is uniqe seems to be done case insensitive on apple side.

i also added specs for this.

### DIFF
--- a/spaceship/spec/portal/app_spec.rb
+++ b/spaceship/spec/portal/app_spec.rb
@@ -66,7 +66,7 @@ describe Spaceship::Portal::App do
       expect(app.app_id).to eq("B7JBD8LHAA")
       expect(app.is_wildcard).to eq(false)
     end
-    
+
     it "works with specific App IDs even with diffrent case" do
       app = Spaceship::Portal::App.find("net.sunaPPs.151")
       expect(app.app_id).to eq("B7JBD8LHAA")
--- a/spaceship/spec/portal/app_spec.rb
+++ b/spaceship/spec/portal/app_spec.rb
@@ -66,6 +66,12 @@ describe Spaceship::Portal::App do
       expect(app.app_id).to eq("B7JBD8LHAA")
       expect(app.is_wildcard).to eq(false)
     end
+    
+    it "works with specific App IDs even with diffrent case" do
+      app = Spaceship::Portal::App.find("net.sunaPPs.151")
+      expect(app.app_id).to eq("B7JBD8LHAA")
+      expect(app.is_wildcard).to eq(false)
+    end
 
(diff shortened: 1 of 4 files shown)


milch (approved):
Looks good to me 👍

spaceship/spec/portal/app_spec.rb
```diff
@@ -67,6 +67,12 @@
       expect(app.is_wildcard).to eq(false)
     end
 
+    it "works with specific App IDs even with diffrent case" do
```
KrauseFx: Spelling mistake here

Issue #8052, opened 2017-01-30
Pilot find returns unexpected results


prnewman:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I have tested the 'fastlane pilot find' command with valid TestFlight external testers (i.e., users returned by the 'fastlane pilot list' command). The results vary: some users are found, some are not found, and some generate an error.

I have been unable to determine why some email addresses are found, and others fail to be found.

**Example 1: Error**

```
[10:31:20] NYPNewmanMAC:nightly-build paul.newman$ fastlane pilot find joe.user@example.com
[10:31:40]: Login to iTunes Connect (login@example.com)
[10:31:43]: Login successful
[10:31:46]: [joe.user@example.com]: undefined method `length' for nil:NilClass

[!] Some operations failed: [joe.user@example.com]: undefined method `length' for nil:NilClass
```

**Example 2: Not found**

```
[10:26:18] NYPNewmanMAC:nightly-build paul.newman$ fastlane pilot find jane@example.com
[10:27:55]: Login to iTunes Connect (login@example.com)
[10:27:58]: Login successful
[10:28:01]: [jane@example.com]: Tester jane@example.com not found

[!] Some operations failed: [jane@example.com]: Tester jane@example.com not found
```

**Example 3: Found**

```
[10:28:01] NYPNewmanMAC:nightly-build paul.newman$ fastlane pilot find alex.user@example.com
[10:31:12]: Login to iTunes Connect (login@example.com)
[10:31:16]: Login successful
+---------------------+---------------------------------+
|            alex.user@example.com            |
+---------------------+---------------------------------+
| First name          | Alex                            |
| Last name           | User                         |
| Email               | alex.user@example.com |
| Latest Version      | 4.15.0 (4)                      |
| Latest Install Date | 10/06/16 12:59                  |
| 1 Devices           | • iPhone 5s, iOS 10.0.2         |
+---------------------+---------------------------------+

```

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                                          |
| ------------------- | -------------------------------------------------------------- |
| OS                  | 10.11.6                                                        |
[...]

PR #8071, opened 2017-02-01
[WIP] fix crash on tester-id = nil

hjanuschka:
hopefully fixes issue: https://github.com/fastlane/fastlane/issues/8052
i am waiting for user to confirm

### DIFF
--- a/spaceship/lib/spaceship/tunes/tester.rb
+++ b/spaceship/lib/spaceship/tunes/tester.rb
@@ -90,7 +90,7 @@ module Spaceship
         # @param identifier (String) (required): Value used to filter the tester, case insensitive
         def find(identifier)
           all.find do |tester|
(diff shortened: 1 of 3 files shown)


spaceship/lib/spaceship/tunes/tester.rb
```diff
@@ -90,7 +90,7 @@ def all
         # @param identifier (String) (required): Value used to filter the tester, case insensitive
         def find(identifier)
           all.find do |tester|
-            (tester.tester_id.to_s.casecmp(identifier.to_s).zero? or tester.email.to_s.casecmp(identifier.to_s).zero?)
+            ((tester.tester_id && tester.tester_id.to_s.casecmp(identifier.to_s).zero?) or tester.email.to_s.casecmp(identifier.to_s).zero?)
```
KrauseFx: `nil.to_s` => `""`, so that shouldn't be the issue

hjanuschka:
ok seems to be not the root of the issue. closing it.

PR #8080, opened 2017-02-01
[fastlane] local actions should be imported before their Fastfile otherwise importing the Fastfile might fail

lacostej:


KrauseFx (approved):

fastlane/spec/fixtures/fastfiles/import1/actions/hello.rb
```diff
@@ -0,0 +1,35 @@
...
+      def self.output
+        [
+        ]
+      end
+
+      def self.return_value
+      end
+
+      def self.authors
+        ["lacostej/jerome@wewanttoknow.com"]
```
KrauseFx: Just the username should be good enough

KrauseFx:
This makes lots of sense, thanks for this. Have you tested if the tests fail when you revert the actual change?

lacostej:
@KrauseFx yes I started writing the test first then squashed. Next time I push separately so you get the circleci failure first if you prefer.

PR #8082, opened 2017-02-02
[Sigh] download_all name file using profile UUID instead of application bundle identifier

lgaches:
### Checklist

### Description
download_all action needs to name every mobile provision with an unique name.

### Motivation and Context
A team can have several mobile provision for the same application bundle identifier but with a different name. Naming the file with the application bundle identifier doesn't allow to download all the mobile provision from a specific team because the last one downloaded will replace all the previous. By using the profile UUID, we'll be sure that all mobile provision are downloaded.


sigh/lib/sigh/download_all.rb
```diff
@@ -23,7 +23,7 @@ def download_profile(profile)
       type_name = profile.class.pretty_type
       type_name = "AdHoc" if profile.is_adhoc?
 
-      profile_name = "#{type_name}_#{profile.app.bundle_id}.mobileprovision" # default name
+      profile_name = "#{type_name}_#{profile.uuid}.mobileprovision" # default name
```
hjanuschka: can you please add the   `profile.app.bundle_id` too -  so that we see the bundle id in the filename too.

hjanuschka (changes requested):

hjanuschka (approved):
nice change, thx for your contribution! 🚀

elianfish:
We have the following command to create a profile file

Fastlane sigh -a com.ysl.apple -u ysl@163.com -n apple_appstore

Finally, the local storage name is InHouse_com.ysl.apple.mobileprovision

But when we updated the profile and execute the download command:

Fastlane sigh download_all -u ysl@163.com

Its local storage name becomes InHouse_41939ba6-6190-4024-a7d7-e69922752778_com.ysl.apple.mobileprovision

Excuse me, create a new profile and re-download the profile, the storage name can be consistent?

PR #8104, opened 2017-02-03
html encode all error messages

hjanuschka:


mgrebenets (approved):

KrauseFx (approved):
I wonder why there are no tests that need to be updated?

hjanuschka:
i was shocked about no test failing too - i expected atleast a couple of them to fail.

KrauseFx:
@hjanuschka can you take a look on why this didn't cause any test failures?

hjanuschka:
in fact there are only 1 spec using the `userInfo`, `resultString`.
`portal_permission_spec.rb` -> mocking a result, but not using html encoded things.


and a grep over the whole repo, just shows that `resultString` and `userInfo` are not really used.
one case is `(content["resultString"] || "").include?("NotAllowed")` -> should not be affected by the encoding.

so i am confident it is ok.


not sure if a spec with a mocked html response, checking if it gets escaped makes sense?

PR #8110, opened 2017-02-03
[actions/hockey] Improvements: allow to proxy, improve error messages, do not upload unnecessary extra files when sending non ipas, support create/update mechanism

lacostej:
Features
* allow to proxy
* improve error messages
* do not upload unnecessary extra files when sending non ipas
* remove empty fields in request.
* support create/update mechanism

@AliSoftware any feedback on the new ENV variables?

Even if we don't have a standard inside fastlane today, could we have at least something to aim for, so that we don't add to the mess.

### DIFF
--- a/fastlane/lib/fastlane/actions/hockey.rb
+++ b/fastlane/lib/fastlane/actions/hockey.rb
@@ -257,7 +257,7 @@ module Fastlane
       end
 
       def self.author
-        ["KrauseFx", "modzelewski"]
+        ["KrauseFx", "modzelewski", "lacostej"]
       end
 
       def self.is_supported?(platform)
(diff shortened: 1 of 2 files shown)


AliSoftware:
I'm ok for the ENV var name (`SPACESHIP_DEBUG`), so for this part I'm 👍 

Otherwise, I don't know about `Faraday` and never used actions like `HockeyApp`, so not sure I'm the right person to review the rest of that PR.

lacostej:
@KrauseFx I made #8124 . If that one is accepted I'll remove the change in `supported_platform?`

hjanuschka (approved):

Issue #8113, opened 2017-02-03
[Regression] "fastlane match nuke distribution" fails


rhcpfan:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

After running `fastlane match dstribution`, I get ```/Users/Administrator/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.14.1/match/lib/match.rb:35:in `cert_type_sym': [!] Unknown cert type: 'distribution' (RuntimeError)```

I think (but I'm not sure) that you can fix this by changing line 34 from `match/lib/match.rb` from 
```ruby
return :distribution if type == "adhoc" || type == "appstore"
```
to
```ruby
return :distribution if type == "adhoc" || type == "appstore" || type == "distribution"
```

I guess it's a bug from the latest release 😕

KrauseFx:
Thanks for submitting this, super helpful 👍  We just prepared https://github.com/fastlane/fastlane/pull/8119 for this

rhcpfan:
Awesome! Glad I could help 😉

PR #8119, opened 2017-02-03
Fix `match nuke` for distribution

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/8113

### DIFF
--- a/match/lib/match.rb
+++ b/match/lib/match.rb
@@ -31,7 +31,7 @@ module Match
   def self.cert_type_sym(type)
     return :enterprise if type == "enterprise"
     return :development if type == "development"
-    return :distribution if type == "adhoc" || type == "appstore"
+    return :distribution if ["adhoc", "appstore", "distribution"].include?(type)
     raise "Unknown cert type: '#{type}'"
   end
 end


endocrimes (approved):

DanToml:
We should also add a test for this, but merge now to get it into the release

KrauseFx:
Absolutely. _match_ is a bad fastlane citizen and has almost no tests compared to the other tools unfortunately 😢

lacostej:
@KrauseFx sorry about that regression. Another thing that is rather confusing in match is that there are plenty of methods to convert from on type to another and all the inputs are strings, or sometimes symbols. Hard to know what the potential inputs are and test all cases. That's also why I wanted those conversion methods to be more explicits.

KrauseFx:
@lacostej I agree, it's not ideal

lacostej:
@KrauseFx the good thing is that with the new code pattern, there's no implicit conversion and the issue is very easily identified and fixed. You avoid having issues like the one where nuking enterprise would nuke development certificates. This is a code pattern we should follow more.

KrauseFx:
@lacostej what code pattern do you mean?

PR #8124, opened 2017-02-03
[fastlane] Relax supported platforms

lacostej:
Platform support is used in several ways:
* for fastlane to hardcode the list of supported platforms (See `SupportedPlatforms`)
* for actions to list the platforms they support (See `self.is_supported?(platform)` under `fastlane/lib/fastlane/action.rb`)
* for fastlane to list the actions available for a particular platform:  (see `fastlane actions --platform ios`)

The main question is: is there any value in strongly restricting the list of platforms fastlane can support?

I personally think not. In fact we today hack around this to support for the `:windows` platform ( see https://gist.github.com/lacostej/7ac093cf035eb2a87a1559cc81fe8742)

So we propose to:
1. relax it and turn the error throwing into warnings. Fastlane would continue to have a list of supported platforms, but users could define new ones without fastlane preventing them to do so.
2. add the ability for users to register supported platforms per project, to reduce the amount of warnings displayed.

This PR currently adds support for both ideas. The ideas are implemented in different commits. I also added an action `add_extra_platforms` to register the new platform.

One issue to think about is the complexity of the CL. The non determinism of the type of the first argument (lane, action, platform etc) could cause issues if platforms added later conflict with lanes/action names.

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -262,12 +262,11 @@ module Fastlane
 
     def verify_supported_os(name, class_ref)
       if class_ref.respond_to?(:is_supported?)
-        if Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
-          # This value is filled in based on the executed platform block. Might be nil when lane is in root of Fastfile
-          platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
-
+        # This value is filled in based on the executed platform block. Might be nil when lane is in root of Fastfile
+        platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
+        if platform
           unless class_ref.is_supported?(platform)
-            UI.user_error!("Action '#{name}' doesn't support required operating system '#{platform}'.")
+            UI.important("Action '#{name}' isn't known to support operating system '#{platform}'.")
           end
         end
       end
--- a/fastlane/lib/fastlane/supported_platforms.rb
+++ b/fastlane/lib/fastlane/supported_platforms.rb
@@ -1,11 +1,21 @@
 module Fastlane
   class SupportedPlatforms
+    class << self
+      attr_accessor :extra
+      attr_reader :default
+
+      def extra=(value)
+        value ||= []
+        UI.important("Setting '#{value}' as extra SupportedPlatforms")
+        @extra = value
+      end
+    end
+
+    @default = [:ios, :mac, :android]
+    @extra = []
+
     def self.all
-      [
-        :ios,
-        :mac,
-        :android
-      ]
+      (@default + @extra).flatten
     end
 
     # this will log a warning if the passed platform is not supported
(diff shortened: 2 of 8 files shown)


fastlane/spec/fast_file_spec.rb
```diff
@@ -178,6 +177,13 @@
         File.delete("/tmp/after_all.txt")
       end
 
+      it "allows the user to invent a new platform", focus: true do
```
KrauseFx: Why do we have to focus? I don't think it's needed

fastlane/lib/fastlane/runner.rb
```diff
@@ -258,12 +258,11 @@ def execute_flow_block(block, current_platform, lane, parameters)
 
     def verify_supported_os(name, class_ref)
       if class_ref.respond_to?(:is_supported?)
-        if Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
```
KrauseFx: I understand that instead of raising an exception we show a warning. But why are the other changes necessary here?

fastlane/spec/fast_file_spec.rb
lacostej: Nah, was for me running tests. It slipped :)

KrauseFx: 👍

fastlane/lib/fastlane/runner.rb
lacostej: if by other changes you mean 

* those in this file, then it's just reordering the code to reduce code duplication

* if it is about the other changes, they allow the user to add a new (unsupported) platform in her Fastfile without having the platform to crash. See the overall PR description to see how we use it today at my work place.

hjanuschka (approved):
that is such a cool change! was looking for something for a long time. 🎉

fastlane/lib/fastlane/supported_platforms.rb
```diff
@@ -1,17 +1,27 @@
...
-        :ios,
-        :mac,
-        :android
-      ]
+      (@default + @extra).flatten
     end
 
-    # this will throw an exception if the passed platform is not supported
+    # this will log a warning if the passed platform is not supported
     def self.verify!(platform)
```
lacostej: one could consider removing the trailing slash here. Or keeping it for Backwards Compatibility only and adding a new method without it.

lacostej:
Would love some feedback from @KrauseFx as well or any other Googler. Also comment on the `AddExtraPlatformsAction`. Do we want something like that in core? If so I can add it and make a PR for doc.

hjanuschka:
@lacostej i saw the sample with `:windows` getting added on-demand, if that is the case, the purpose of the action is to suppress the warning - for me it sounds good to ship a action for that.?

lacostej:
@hjanuschka with the proposed PR

* if the user doesn't add his extra platforms somehow to fastlane (e.g. using a CustomAction like the one I pasted in the description), every invocation of `verify!` will trigger  warning. So if you run `fastlane windows something` you will get some warnings in the logs. Possibly several

* If you register the extra platforms, you will get a single warning, at registration time. The warning was not removed to make sure people understood this is a somewhat non official setup. The warning could also be removed.

hjanuschka:
+1 for shipping the action then! thx for detailed explanation

KrauseFx (approved):

PR #8126, opened 2017-02-04
Catch and print exceptions occuring in error block

KrauseFx:
This has been annoying, and really hard to debug: when the error block raised an exception, it basically replaced the actual reason of error. The error block is usually used to notify about the failure, but the user is interested in the *actual* error message.

This PR also adds a test covering exactly that

screenshot 2017-02-03 16 02 55

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -58,10 +58,14 @@ module Fastlane
       rescue => ex
         Dir.chdir(path_to_use) do
           # Provide error block exception without colour code
-          error_ex = ex.exception(ex.message.gsub(/\033\[\d+m/, ''))
-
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/runner.rb
```diff
@@ -58,10 +58,14 @@ def execute(lane, platform = nil, parameters = nil)
       rescue => ex
         Dir.chdir(path_to_use) do
           # Provide error block exception without colour code
-          error_ex = ex.exception(ex.message.gsub(/\033\[\d+m/, ''))
```
mfurtak: Is OK that we've lost this message color removal?

KrauseFx: Yes, forgot to mention that in the PR body: this is not needed any more, as we now use `UI.user_error!` or `UI.crash!` instead of `"error_message".red` (which was the case in the early fastlane days

mfurtak (approved):
👍

PR #8137, opened 2017-02-06
[WIP]: change terminal-table to tty-table

hjanuschka:
do not merge.

todo:
  - [ ]  fix resize crash (if terminal is to tiny)
  - [ ] find a way to support "title"


lacostej:
Wonder if `limit_row_size` disappeared

hjanuschka:
@lacostej  this is not even close to ready for review - and well ttty-table should do the resize (as i read through the docs). don't mind pushing to that branch if you find a fix 👍

hjanuschka:
right now its a bit too hard to replace current terminal-table implemenation, and not on-board alot of issues.  - lets move on to #7391 - which adresses the ugliness of tables on small terminals.

PR #8147, opened 2017-02-06
Skip loading and printing plugins when running `--help`

KrauseFx:
You can reproduce using `bundle exec fastlane --help | cat`


hjanuschka (approved):

hjanuschka:
why the `cat` at the end?

KrauseFx:
Because otherwise it's using `less`, you'd still see the plugins, but only after exiting 👍

ohwutup:
💯

PR #7386, opened 2016-12-08
Centralize BuildWatcher for iTunes Connect builds

hjanuschka:
this PR's try's to de-duplicate the various blocks that check if a build is finished processing.
by moving the logic to fastlanecore.

bildschirmfoto 2016-12-08 um 21 49 00

PR #8152, opened 2017-02-07
Fix pilot stuck on older processing build

### LINKED ISSUE #6069: `pilot` upload waits forever for processing of an *older* build if that is stuck in processing
### New Issue Checklist
### Issue Description

When a TestFlight build is "processing", and a new build is uploaded with `pilot` without turning off waiting for processing (the default), then `pilot` will wait for the _old_ build to finish processing even after the actual (new) build it just uploaded has already finished processing.

This is quite a nuisance when the old build is stuck in processing for a long time, e.g. happened to me 2 days ago when uploading with bad internet connectivity resulted in a "successful" upload which however was somehow damaged. The build, with build number 66, displayed as uploaded in iTunes Connect, but was "processing" all the time without option to click its details (and without app icon displayed). It took more than 24 hours for this damaged build to get removed from iTunes Connect. When uploading a subsequent, new build, with build number 67, `pilot` displayed _"Waiting for iTunes Connect to finish processing the new build (1.0.0 - 66)"_ (i.e. the old build, not the one that was uploaded by that invocation of fastlane).

Please note that there is no option to remove such a stalled old build from iTunes Connect on one's own. I contacted iTunes Support with request to remove the build, but even they are unable to remove the build as they responded:

> The system will eventually cleanup builds stuck in processing states, but **we are unable to remove them manually** at this time.
> 
> If you are the developer of your integration system, you may consider modifying the system to account for situations like this. Otherwise you might want to report the issue to the developer of this integration system so they are aware of the issue.

I suggest fastlane accounts for this by checking for processing status specifically of the build it uploaded, not [checking in general whether any build is processing](https://github.com/fastlane/fastlane/blob/master/pilot/lib/pilot/build_manager.rb#L127-L128).

No longer have it, the stalled build has since then been removed so cannot reproduce it.
### Configuration Files

using `pilot` with `skip_waiting_for_build_processing` left unspecified, defaulting to `false`:

```
...
pilot distribute_external: false, changelog: changelog
...
```
### Environment

fastlane version (run `fastlane -v`): **`fastlane 1.102.0`**

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)? **yes**

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)? **rbenv (ruby 2.0.0p648)**
### Note

A fix for this will obviously be hard to test 😄. Reproducing a successful upload of a damaged build is probably hard. However it is possible to at least test that normal uploads continue working.

negebauer:
### Checklist

### Description
This makes pilot only wait for the latest upload when waiting for processing, instead of checking for all builds that might be processing

### Motivation and Context
Fixes the problem where pilot would get stuck waiting for an older build to finish processing by iTunes Connect
This could happen, for example, with a faulty upload. The only way of fixing this currently is by waiting till an Apple script removes the faulty build

This should fix #6069


pilot/lib/pilot/build_manager.rb
```diff
@@ -142,10 +143,14 @@ def wait_for_processing_build(options, platform)
         #  have a nil build
         if app.build_trains(platform: platform).count == 0
           UI.message("New application; waiting for build train to appear on iTunes Connect")
-        else
+        elsif !found_build
           builds = app.all_processing_builds(platform: platform)
           break if builds.count == 0
           latest_build = builds.last
+          found_build = true
+        else
```
lacostej: would this alternative code work? (one liner)

* remove the found_build variable
* use
```ruby
    else
      builds = app.all_processing_builds(platform: platform)
      break if builds.count == 0
      latest_build = builds.last		
      next unless builds.include? latest_build # skip broken builds and retry
```

KrauseFx:
We've actually tweaked this a lot in the past, and don't have a central place for it yet. @hjanuschka did some great work on https://github.com/fastlane/fastlane/pull/7386 https://github.com/fastlane/fastlane/pull/7387 and https://github.com/fastlane/fastlane/pull/7388. I'd very much be in favor of merging the 3 PRs into one (now that we're on a mono-gem) and add the changes you proposed to that if they make sense. 

@hjanuschka what are your thoughts on it? 👍 The PRs have been open for a while, but I'd love to get this resolved 🚀

KrauseFx (dismissed):
.

negebauer:
Seems good to me
Better to first make the unification of the wait build function and then think about this problem

hjanuschka:
@KrauseFx  i cleaned up the PR's closing all - only keeping #7386  - still waiting for tests to pass, but it should be ready for review.

@negebauer  maybe you can have a look/test it, and tell me if it works for you.

negebauer:
@hjanuschka I'm kind of new here 😅 
On can I run fastlane on your PR?

lacostej:
@negebauer @hjanuschka's PR was merged so we can resume yours.

I have a question: can you explain how repeating the fetching of the processing builds makes sure that we wait for the right build? Is the intent that, in the time frame the build was sent, it will be added to the list of processing builds?

Shouldn't we instead for example:
* wait for a very specific build number (the one we just uploaded)
* check the processing builds before up uploading and comparing before starting to wait

etc.

negebauer:
@lacostej 
The repeating part was how it was done before. I understood from the comments of `wait_for_processing_build ` [here](https://github.com/negebauer/fastlane/blob/dd22761b0f7aaf8b0e20ce5d301f442282e1ca30/pilot/lib/pilot/build_manager.rb#L141) that only when it's the first build of a new app it doesn't go into processing right away. So the last should be the one that was just updated.
What I propose to change is to only wait for one build (the latest) instead of checking if all builds finished processing

negebauer:
@KrauseFx @hjanuschka @lacostej  I had this problem again today
captura de pantalla 2017-03-03 a las 19 46 34
captura de pantalla 2017-03-03 a las 19 48 08

I uploaded the build number 75 binary but pilot is still stuck waiting for build  73

Is it good for merging?

hjanuschka:
@negebauer any updates on this PR? the build watcher has been reverted

negebauer:
@hjanuschka  I haven't looked at it
So there a multiple watchers now?

KrauseFx:
Unfortunately the build watcher change introduced multiple regressions so we had to revert the changes for now 😢

Do we still need this PR then?

negebauer:
@KrauseFx According to [this](https://github.com/fastlane/fastlane/blob/master/pilot/lib/pilot/build_manager.rb#L149)

```ruby
    break if builds.count == 0
    latest_build = builds.last
```
This PR would still be needed

If you upload a build with pilot, cancel it before it's done and then upload a second one pilot will be stuck waiting for the first one to be processed, which will never happen (apple will actually delete that build at some point)

We would just need to change that for this
```ruby
    latest_build = builds.last unless latest_build
    break unless builds.include?(latest_build)
```

negebauer:
@KrauseFx @hjanuschka I actually deleted the fork from where this PR came (yeah, silly me)
I opened a new PR #8640 with the change in the `pilot` build watcher

Issue #8154, opened 2017-02-07
Broken plugins break fastlane entirely


lacostej:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

If you have a broken plugin in your project, you can't do simple things like `fastlane env` or `fastlane update_plugins`anymore.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
/Users/lacostej/.rvm/gems/ruby-2.2.4/bundler/gems/fastlane-plugin-jenkins_job_config-0202463333b7/lib/fastlane/plugin/jenkins_job_config.rb:15:in `require': /Users/lacostej/.rvm/gems/ruby-2.2.4/bundler/gems/fastlane-plugin-jenkins_job_config-0202463333b7/lib/fastlane/plugin/jenkins_job_config/actions/jenkins_job_config_action.rb:31: syntax error, unexpected ')', expecting '}' (SyntaxError)
/Users/lacostej/.rvm/gems/ruby-2.2.4/bundler/gems/fastlane-plugin-jenkins_job_config-0202463333b7/lib/fastlane/plugin/jenkins_job_config/actions/jenkins_job_config_action.rb:105: syntax error, unexpected end-of-input, expecting keyword_end
	from /Users/lacostej/.rvm/gems/ruby-2.2.4/bundler/gems/fastlane-plugin-jenkins_job_config-0202463333b7/lib/fastlane/plugin/jenkins_job_config.rb:15:in `block in <top (required)>'
	from /Users/lacostej/.rvm/gems/ruby-2.2.4/bundler/gems/fastlane-plugin-jenkins_job_config-0202463333b7/lib/fastlane/plugin/jenkins_job_config.rb:14:in `each'
	from /Users/lacostej/.rvm/gems/ruby-2.2.4/bundler/gems/fastlane-plugin-jenkins_job_config-0202463333b7/lib/fastlane/plugin/jenkins_job_config.rb:14:in `<top (required)>'
	from /Users/lacostej/.rvm/gems/ruby-2.2.4/gems/fastlane-2.14.2/fastlane/lib/fastlane/fastlane_require.rb:10:in `require'
[...]

PR #8166, opened 2017-02-07
[fastlane] Properly catch all kinds of errors, including parsing errors, while loading plugins #8154

lacostej:
Before fix:
image

After fix:
image


KrauseFx (approved):
Looks great, nice catch 👍

hjanuschka (approved):

Issue #6292, opened 2016-09-26
Providing a detailed error message to slack


jamyc:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

I'm forcing a failing build and get a generic exception in Slack:
`Error building the application - see the log above`

A couple lines above that (in the Jenkins console output) it shows the error perfectly fine:

`[31m[x] [0m/Users/[USER]/.jenkins/workspace/Fastlanedev/Fastlane-integration/ViewController.swift:16:9: [31muse of unresolved identifier 'prnt'[0m`

Is there any possible way to either:
- Provide a link to the failing Jenkins build
- Provide a more detailed message
### Configuration Files

Fastfile:

``` ruby
  desc "Experimental dev lane"
  lane :dev do
    gym(scheme: "Fastlane-integration", use_legacy_build_api: true)
slack(
  message: "Build succeeded",
  channel: "#fastlane",
  success: true,
  payload: {
    'Build Date' => Time.new.to_s,
    'Built by' => 'Jenkins',
  },
  default_payloads: [:git_branch, :git_author]
)
  end

  error do |lane, exception|
     slack(
       message: exception.to_s,
       success: false
     )
  end
```
##### fastlane 1.104.0

Running the dev lane with a build step in Jenkins:

TKBurner:
@jamyc Great to hear from you! I think this could be better too. Are you interested in submitting a PR? This seems like a great place to contribute. Thanks!

thiagofelix:
@jamyc you might be able to get the link to the Jenkins build using the environment variable `BUILD_URL` . 

More information on the [Jenkins documentation](https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-JenkinsSetEnvironmentVariables)

jamyc:
@TKBurner Unfortunately, I won't be able to get to a PR request to implement any of this.

@thiagofelix Thank you for your suggestion, that worked out perfectly fine!

netbe:
@jamyc and @TKBurner  was there a PR done/merged?

jamyc:
@netbe Not that I know of. I used the suggestion of @thiagofelix to include the BUILD_URL in the slack message.

thiagofelix:
From the original request
- [x] Provide a link to the failing Jenkins build
[...]

PR #8173, opened 2017-02-08
Provide detailed error message when gym fails #6292

thiagofelix:
### Checklist

### Description

I am adding the field `error_info` to FastlaneError class intended to contain detailed information about the error along side with the error message itself. 
The new field can be specified using the UI.user_error `options` parameter, same way the `show_github_issues` works.

### Motivation and Context

Currently if you have a problem with the gym execution the `error` block will receive only a friendly message saying that you build or package command didn't work. The idea here is to extend the error class so the user can get access to the whole xcodebuild output and use that information when reporting the problem. 

One use case would be sending a slack message with the build output as attachment so everyone in the room can quickly scan and see what went wrong.

### Usage Example

**Badly behaving lane:**

```ruby
lane :build  do
    gym(..)
end

error do |lane, exception|
    slack(
       # message with short human friendly message
       message: exception.to_s, 
       success: false, 
       # Output containing extended log output
       payload: { "Output" => exception.error_info.to_s } 
    )
end
```

**Slack Message:**
image

It would be even better if xcodebuild given us exactly what broke the build (file, line number, etc), this way the error_info would be more precise showing the problem versus just dumping the whole process output.

The original issue/discussion can be found here #6292 

EDIT: Forgot to mention the changes on documentation, I add some information to both fastlane/Readme and Gym specific documentation.

### DIFF
--- a/fastlane/docs/README.md
+++ b/fastlane/docs/README.md
@@ -80,7 +80,7 @@ end
 ### `error` block
 
 This block will get executed when an error occurs, in any of the blocks (*before_all*, the lane itself or *after_all*).
-You can get more information about the error using the field `error_info`.
+You can get more information about the error using the `error_info` property.
 
 ```ruby
 error do |lane, exception|
@@ -91,7 +91,6 @@ error do |lane, exception|
   )
 end
 ```
-
 
 ## Extensions
 
--- a/fastlane/docs/README.md
+++ b/fastlane/docs/README.md
@@ -80,15 +80,18 @@ end
 ### `error` block
 
 This block will get executed when an error occurs, in any of the blocks (*before_all*, the lane itself or *after_all*).
+You can get more information about the error using the field `error_info`.
 
 ```ruby
 error do |lane, exception|
   slack(
     message: "Something went wrong with the deployment.",
-    success: false
+    success: false,
+    payload: { "Error Info" => exception.error_info.to_s } 
   )
 end
 ```
+
 
 ## Extensions
 
--- a/fastlane_core/lib/fastlane_core/ui/interface.rb
+++ b/fastlane_core/lib/fastlane_core/ui/interface.rb
@@ -143,7 +143,6 @@ module FastlaneCore
     # Basically this should be used when you actively catch the error
     # and want to show a nice error message to the user
     def user_error!(error_message, options = {})
-      options = { show_github_issues: false, error_info: nil }.merge(options)
       raise FastlaneError.new(options), error_message.to_s
     end
 
--- a/fastlane_core/lib/fastlane_core/ui/interface.rb
+++ b/fastlane_core/lib/fastlane_core/ui/interface.rb
@@ -118,9 +118,11 @@ module FastlaneCore
     # raised from user_error!
     class FastlaneError < StandardError
       attr_reader :show_github_issues
+      attr_reader :error_info
 
-      def initialize(show_github_issues: false)
+      def initialize(show_github_issues: false, error_info: nil)
         @show_github_issues = show_github_issues
+        @error_info = error_info
       end
     end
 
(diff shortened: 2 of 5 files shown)


fastlane_core/lib/fastlane_core/ui/interface.rb
```diff
@@ -118,9 +118,11 @@ class FastlaneCrash < StandardError
     # raised from user_error!
     class FastlaneError < StandardError
       attr_reader :show_github_issues
+      attr_reader :error_info
```
lyndsey-ferguson: Just a thought, but as we will receive an 'exception', could the instance variable could just be `:info`?

thiagofelix: I agree with you, *error_*  is redundant inside the exception, good catch.
The reason behind error_Info was to follow with [SpaceShip similar error class](https://github.com/fastlane/fastlane/blob/e02b5d15ec6606bad0ad954b0fb62fea144f39b2/spaceship/lib/spaceship/client.rb#L72) which also expose property named `error_info`

gym/README.md
```diff
@@ -192,8 +192,22 @@ lane :beta do
   gym(scheme: "MyApp")
   crashlytics
 end
+
+# error block is executed when a error occurs
+error do |lane, exception|
+    slack(
```
KrauseFx: This should be 2 spaces 👍

thiagofelix: Fixed =) tks

gym/README.md
```diff
@@ -195,13 +195,13 @@ end
...
 error do |lane, exception|
-    slack(
-       # message with short human friendly message
-       message: exception.to_s, 
-       success: false, 
-       # Output containing extended log output
-       payload: { "Output" => exception.error_info.to_s } 
-    )
+  slack(
+      # message with short human friendly message
```
KrauseFx: That still needs to be updated 👍

thiagofelix: Ops sorry, fixed again 😂

fastlane_core/lib/fastlane_core/ui/interface.rb
DanToml: I'd keep the `error_info` property as is, for consistency, and because you're showing an underlying error.

fastlane_core/lib/fastlane_core/ui/interface.rb
```diff
@@ -141,8 +143,8 @@ def crash!(exception)
     # Basically this should be used when you actively catch the error
     # and want to show a nice error message to the user
     def user_error!(error_message, options = {})
-      options = { show_github_issues: false }.merge(options)
-      raise FastlaneError.new(show_github_issues: options[:show_github_issues]), error_message.to_s
+      options = { show_github_issues: false, error_info: nil }.merge(options)
```
DanToml: You should be able to pass `options` straight through, as both of these are the same as the default values.

fastlane/docs/README.md
```diff
@@ -80,16 +80,19 @@ end
...
   slack(
     message: "Something went wrong with the deployment.",
-    success: false
+    success: false,
+    payload: { "Error Info" => exception.error_info.to_s } 
   )
 end
 ```
 
+
```
DanToml: rm extra new line

fastlane/docs/README.md
```diff
@@ -80,16 +80,19 @@ end
 ### `error` block
 
 This block will get executed when an error occurs, in any of the blocks (*before_all*, the lane itself or *after_all*).
+You can get more information about the error using the field `error_info`.
```
DanToml: maybe "using the `error_info` property."?

gym/README.md
```diff
@@ -192,8 +192,22 @@ lane :beta do
...
+    message: exception.to_s, 
+    success: false, 
+    # Output containing extended log output
+    payload: { "Output" => exception.error_info.to_s } 
+  )
+end
 ```
 
+When gym raises an error the field `error_info` will contain the process output
+in case you want to display the error in 3rd party tools such as slack.
```
DanToml: "the `error_info` property" and "Slack"

endocrimes (changes requested):
A few points but this is a nice change :+1:

fastlane_core/lib/fastlane_core/ui/interface.rb
thiagofelix: Good call Dan, thanks!

fastlane/docs/README.md
thiagofelix: Fixed =)

endocrimes (approved):

DanToml:
This is great, thank you @thiagofelix!

PR #8196, opened 2017-02-10
[CI] Run using Xcode 8.2

endocrimes:
This updates fastlane CI to Xcode 8.2, with the advantages being:
- Testing on the Xcode version that many Fastlane users will be using
- Validating Sierra support (Xcode 7 is funky on Sierra sometimes)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   environment:
     FL_BUILDLOG_PATH: ~/Library/Logs
   xcode:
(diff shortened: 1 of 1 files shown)


lacostej:
Sounds to be we should run a matrix of jobs with various configurations, at least once in a while.

KrauseFx (approved):
Looks great to me 🎷 Completely forgot about bumping this

PR #7868, opened 2017-01-15
respect file extension based on users uploaded one.

hjanuschka:
fixes: #7857

i verified the globbing and end_with via this little sample.:
having a `app_icon.jpg` and `app_icon.png` in a folder `demo/`

```ruby
u = "http://Wwwwwww.........jpg"
puts u.end_with? "jpg"
files = Dir["demo/app_icon.{png,jpg}"]
puts files.inspect
exit

```

PR #8225, opened 2017-02-14
[deliver] Fix uncaught nil value + watch icon option name

reidab:
### Description

- Fixes an uncaught nil value in cases where no app icon was found, introduced in #7868 
- While fixing that ☝️, I noticed that watch icons were being set to `options[:app_icon]` instead of `options[:apple_watch_app_icon]`

### Motivation and Context

My app's `deliver` step started failing with `[!] no implicit conversion of nil into String (TypeError)` after updating.

### DIFF
--- a/deliver/lib/deliver/runner.rb
+++ b/deliver/lib/deliver/runner.rb
@@ -76,7 +76,7 @@ module Deliver
       options[:app_icon] ||= default_app_icon_path if default_app_icon_path && File.exist?(default_app_icon_path)
 
       default_watch_icon_path = Dir[File.join(options[:metadata_path], "watch_icon.{png,jpg}")].first
-      options[:app_icon] ||= default_watch_icon_path if default_watch_icon_path && File.exist?(default_watch_icon_path)
+      options[:apple_watch_app_icon] ||= default_watch_icon_path if default_watch_icon_path && File.exist?(default_watch_icon_path)
     end
 
     # Upload the binary to iTunes Connect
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Looks good to me, thanks for contributing @reidab. Did that cause a crash in previous versions? Should we ship a release for this change? cc @hjanuschka

reidab:
In previous versions, `File.exist?` would always have been called with a string and would have returned appropriately. It seems reasonable to assume that `File.exist?` would just return `false` when called with `nil`, but it doesn't 😞.

deliver/lib/deliver/runner.rb
```diff
@@ -73,10 +73,10 @@ def prepare_app_icons(options = {})
       return unless options[:metadata_path]
 
       default_app_icon_path = Dir[File.join(options[:metadata_path], "app_icon.{png,jpg}")].first
-      options[:app_icon] ||= default_app_icon_path if File.exist?(default_app_icon_path)
+      options[:app_icon] ||= default_app_icon_path if default_app_icon_path && File.exist?(default_app_icon_path)
 
       default_watch_icon_path = Dir[File.join(options[:metadata_path], "watch_icon.{png,jpg}")].first
-      options[:app_icon] ||= default_watch_icon_path if File.exist?(default_watch_icon_path)
+      options[:apple_watch_app_icon] ||= default_watch_icon_path if default_watch_icon_path && File.exist?(default_watch_icon_path)
```
KrauseFx: Whops, should that stay `options[: app_icon]` instead? Did you want to change this too?

hjanuschka:
@reidab many thx.🚀
@krausefx i missed the ping on that - thx forbthebhotfix release

PR #8232, opened 2017-02-14
Test code samples at build time

lacostej:


fastlane/actions/test_sample_code.rb
```diff
@@ -0,0 +1,105 @@
...
+          end
+        end
+
+        UI.error("Found errors in the documentation, more information above") unless errors.empty?
+        errors.each do |ex|
+          UI.error(ex)
+        end
+
+        ENV.delete("CI")
+        UI.user_error!("Found at #{errors.count} errors in the documentation") unless errors.empty?
```
mfurtak: Found at -> Found

lacostej:
@mfurtak updated

What about the duplication with the docs project? Should I make it so this action becomes reusable for external projects. It could be useful for plugin writers I guess.

KrauseFx:
The reason why we don't do that by default, is the additional build time, did you track how many minutes this will add?

lacostej:
On my 2011 Macbook Pro:

```xml
      <testcase classname="fastlane.lanes" name="10: Switch to test_generate_docs lane" time="0.003411">
      </testcase>
      <testcase classname="fastlane.lanes" name="11: Switch to ensure_code_samples lane" time="0.001983">
      </testcase>
      <testcase classname="fastlane.lanes" name="12: test_sample_code" time="0.236161">
```

lacostej:
@KrauseFx if you want I can try to optimize one unit test,  to compensate :)

KrauseFx:
I really thought cloning the docs repo takes longer, if it's just a few seconds, I'm totally onboard with it. However we should not duplicate that code, but instead use the `import_from_git` action to import a remote Fastfile

mfurtak (approved):

lacostej:
@KrauseFx I am not cloning the docs repo, just running the check sample action on the newly generated `Actions.md`. This goes very fast.

KrauseFx:
That's smart 👍

PR #8235, opened 2017-02-14
[reset_git_repo] only do a full reset when paths is nil

koenpunt:
### Checklist

### Description

only do a full reset when paths is nil, so no accidental full-resets can occur when accidentally specifying an empty array

### DIFF
--- a/fastlane/lib/fastlane/actions/reset_git_repo.rb
+++ b/fastlane/lib/fastlane/actions/reset_git_repo.rb
@@ -10,7 +10,7 @@ module Fastlane
 
           return paths if Helper.is_test?
 
-          if (paths || []).count == 0
+          if paths.nil?
             Actions.sh('git reset --hard HEAD')
 
             clean_options = ['q', 'f', 'd']


mfurtak:
While the behavior you're going for seems reasonable to me, the change alters existing behavior that some people may be depending on.

Can you say more about why you thing this is important? We should try not to alter existing behavior that people may be using unless absolutely necessary 👍

koenpunt:
I added a glob to the files option, which didn't match any files, and as a result the version change in the xcode project was reverted too..

```rb
reset_git_repo files: Dir['ios/**/AppIcon.appiconset/*.png']
```

mfurtak:
Cool, thanks for the example use-case. Your approach was reasonable, and the results were surprising (and destructive) so, let's do it.

mfurtak (approved):

mfurtak:
Thanks for the contribution 👍

KrauseFx:
Great change 👍

PR #8242, opened 2017-02-14
WIP: try to fix ADP issues

hjanuschka:
endpoint seems to have moved. suggested/initial finding by @shinkitsune 🍻


DavidBertet (changes requested):
The url that works for me is : https://developer.apple.com/services-account/QH65B2/account/listTeams.action  (you missed /account/ at the end)

hjanuschka:
@DavidBertet THX - ok i think i have sorted it out now. (force pushed)

PR #8250, opened 2017-02-15
[spaceship] Fix AppDetails's multilingual fields

jasonkit:
LanguageItem expects the languages hashes have 'language' key for the
locale code, however for AppDetails, the languages hashes are using
'localeCode' for locale code.


asfalcone:
@jasonkit Thanks for the PR. Can you add a test and fixture data for this change?

mfurtak:
I'm going to close this for now. We can always reopen it if more forward progress can be made 👍

KrauseFx:
@jasonkit I'd be curios on when this would fail when being used? How can we reproduce the error you ran into?

PR #8129, opened 2017-02-04
Fix option conflicts in sigh

mfurtak:
Adjust command option handling to stop using `global_option` for Commander options that are
generated for the default command (`renew`). Instead, they are now generated as normal options underneath that command.

The only major (negative?) change in behavior affects how help information is generated/shown by Commander.

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [X] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [X] I've updated the documentation if necessary.

### Description

<details><summary><b>Current option handling in sigh</b></summary>

Current help output:
```
$ fastlane sigh -h
  sigh

  CLI for 'sigh' - Because you would rather spend your time building stuff than fighting provisioning

  Commands:
    download_all Downloads all valid provisioning profiles
    help         Display global or [command] help documentation
    manage       Manage installed provisioning profiles on your system.
    renew        Renews the certificate (in case it expired) and outputs the path to the generated file
    repair       Repairs all expired or invalid provisioning profiles
    resign       Resigns an existing ipa file with the given provisioning profile

  Global Options:
    --verbose
    --adhoc [VALUE]      Setting this flag will generate AdHoc profiles instead of App Store Profiles (SIGH_AD_HOC)
    --development [VALUE] Renew the development certificate instead of the production one (SIGH_DEVELOPMENT)
    --skip_install [VALUE] By default, the certificate will be added on your local machine. Setting this flag will skip this action (SIGH_SKIP_INSTALL)
    -f, --force [VALUE]  Renew provisioning profiles regardless of its state - to automatically add all devices for ad hoc profiles (SIGH_FORCE)
    -a, --app_identifier STRING The bundle identifier of your app (SIGH_APP_IDENTIFIER)
    -u, --username STRING Your Apple ID Username (SIGH_USERNAME)
    -b, --team_id STRING The ID of your Developer Portal team if you're in multiple teams (SIGH_TEAM_ID)
    -l, --team_name STRING The name of your Developer Portal team if you're in multiple teams (SIGH_TEAM_NAME)
    -n, --provisioning_name STRING The name of the profile that is used on the Apple Developer Portal (SIGH_PROVISIONING_PROFILE_NAME)
[...]

PR #8254, opened 2017-02-15
Refactor frameit CommanderGenerator usage

mfurtak:
### Checklist

### Description

Carry the work and lessons learned from #8129 over to _frameit_

### Motivation and Context

_frameit_ does not currently have any option conflicts, but this refactoring is still generally more correct and safe.

### DIFF
--- a/frameit/spec/commands_generator_spec.rb
+++ b/frameit/spec/commands_generator_spec.rb
@@ -1,0 +1,49 @@
+require 'frameit/commands_generator'
+
+describe Frameit::CommandsGenerator do
+  def expect_runner_run
+    fake_runner = "runner"
+    expect(Frameit::Runner).to receive(:new).and_return(fake_runner)
+    expect(fake_runner).to receive(:run)
+  end
+
+  describe ":run options handling" do
+    it "can use the use_legacy_iphone5s flag from tool options" do
+      # leaving out the command name defaults to 'run'
+      stub_commander_runner_args(['--use_legacy_iphone5s', 'true'])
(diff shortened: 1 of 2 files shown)


frameit/spec/commands_generator_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(Frameit.config[:use_legacy_iphone5s]).to be(true)
+    end
+  end
+
+  describe ":rose_gold options handling" do
+    it "can use the use_legacy_iphone5s flag from tool options" do
+      stub_commander_runner_args(['rose_gold', '--use_legacy_iphone5s', 'true'])
+      expect_runner_run
+      Frameit::CommandsGenerator.start
+      expect(Frameit.config[:use_legacy_iphone5s]).to be(true)
```
ohayon: Looks like these lines (42-43) are the same in each test, any chance we can roll them up into the `expect_runner_run` method or into a separate one if thats not a good fit?

ohayon (approved):
Looks great! Just one little comment about code reuse but this is 💯

frameit/spec/commands_generator_spec.rb
mfurtak: I think I like that less than the other kinds of code reuse we've been working towards in these tests. In my opinion it's important for a test to be clear about its trigger and its assertions. Having helpers for setting up a bunch of the repeated preconditions helps readability, but I think hiding/extracting the trigger and assertions, hurts test understandability.

PR #8256, opened 2017-02-15
[supply] only set apk path when none is given

koenpunt:
### Checklist

### Description

only set apk path when none is given

### Motivation and Context

User options should always have precedence, and if for some reason multiple apks are generated, it should still be possible to only use one.

### DIFF
--- a/fastlane/lib/fastlane/actions/supply.rb
+++ b/fastlane/lib/fastlane/actions/supply.rb
@@ -7,11 +7,13 @@ module Fastlane
 
         FastlaneCore::UpdateChecker.start_looking_for_update('supply') unless Helper.is_test?
 
-        all_apk_paths = Actions.lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS] || []
-        if all_apk_paths.length > 1
-          params[:apk_paths] ||= all_apk_paths
-        else
-          params[:apk] ||= Actions.lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]
+        unless params[:apk_paths].present? || params[:apk].present?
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/supply.rb
```diff
@@ -7,11 +7,13 @@ def self.run(params)
 
         FastlaneCore::UpdateChecker.start_looking_for_update('supply') unless Helper.is_test?
 
-        all_apk_paths = Actions.lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS] || []
-        if all_apk_paths.length > 1
-          params[:apk_paths] ||= all_apk_paths
-        else
-          params[:apk] ||= Actions.lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]
+        unless params[:apk_paths].present? || params[:apk].present?
```
lacostej: I think the change is OK, but we should really consider
* adding tests. There are too many ways to set up the paths.
* go towards removing the use of `params[:apk]` at least internally. For example we could do something like

```ruby
params[: apk_paths] = [ Actions.lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH] ] if Actions.lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]
```

lacostej (commented):
@koenpunt do you think you could try adding some tests?

mfurtak:
@koenpunt Do the comments from @lacostej make sense? Do you have a clear path forward for this?

koenpunt:
@mfurtak I tried to write some tests but walked into some difficulties. I can push what I have, and maybe then you can give me some pointers?

mfurtak (changes requested):
`present?` is not a method in core Ruby, only in Rails 😄 

I'll work on adapting this and adding tests 👍

PR #8280, opened 2017-02-17
[match] Set Provisioning Profile path for environment variables

giginet:
### Checklist

### Description

Using match, Added environment variable which indicates fetched Provisioning Profile installed path.

match

### Motivation and Context

I'd like to use Provisioning Profile path to fight against code signing.
This value can be passed to other actions. (e.g. `update_project_provisioning`)

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -62,7 +62,7 @@ module FastlaneCore
           end
         end
 
-        true
+        destination
       end
 
       private
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -134,7 +134,7 @@ module Match
         self.changes_to_commit = true
       end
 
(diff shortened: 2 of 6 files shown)


fastlane_core/lib/fastlane_core/provisioning_profile.rb
```diff
@@ -62,7 +62,7 @@ def install(path)
           end
         end
 
-        true
+        destination
```
giginet: return value is changed.
However, it seems not to be used anywhere.

https://github.com/fastlane/fastlane/blob/master/gym/lib/gym/detect_values.rb#L59
https://github.com/fastlane/fastlane/blob/master/sigh/lib/sigh/manager.rb#L41

mfurtak:
This looks good. 👍 

Could you extend the test coverage in `runner_spec` to provide a regression test for this new behavior? Thanks!

giginet:
@mfurtak Hi, I added tests that match will set environment variables.
Please check it.

hjanuschka (approved):
looking good to me, many THX 🚀

PR #8289, opened 2017-02-17
Adding GitLab CI git_branch ENV variable check

gpg0:
The git_branch action seems to return always HEAD when running inside GitLab CI. I added the env variable CI_BUILD_REF_NAME for GitLab CI.


mfurtak (approved):
Looks good, thanks! 👍

PR #8295, opened 2017-02-18
[Regression] Fixed wrong keys (with dots) inserted when resigning IPA

djsison:
### Checklist

 I tried to run `bundle exec rspec`, but it fails even though without changes yet.

### Description
Using the resign action, wrong keys are inserted as new entitlements when the key contains dots.
For example, if key is "com.apple.developer.associated-domains", it is incorrectly inserted as "com\.apple.developer\.associated-domains".

This issue was brought by the change from the use of \`\` to $().
\`\` replaces some of the escaped characters, but $() treats the contents as the command itself.
For this reason, we also need to change the contents of $() to take into account unnecessary escapes.

### Motivation and Context

> Why is this change required? What problem does it solve?

Resign action succeeds, but user will not be able to install the IPA because entitlements do not match.

> If it fixes an open issue, please link to the issue here.

n/a

> Please describe in detail how you tested your changes.

I tested by resigning an app with Associated Domains capability enabled.

1. Run resign action before the change. This will produce an IPA that can't be installed.
1. Inspect the entitlements of the IPA using `codesign -d --entitlements :- <path to .app>`. It will have a line like this: `<key>com\.apple\.developer\.associated-domains</key>`.
1. Run resign action again after the change. It will now produced an IPA that I can install.

I tried running `bundle exec rspec`, but it fails even though without changes yet.

### Existing problems
I created this PR to seek possible help on how to handle another problem.
With the change in my current PR, entitlement keys like `com.apple.developer.associated-domains` will be correctly inserted to the patched entitlements. But it will also soon be removed because of `BLACKLISTED_KEYS`. The IPA is installable, but Universal Links will not work because the `com.apple.developer.associated-domains` entitlement has been removed.

I am not sure what the reason for adding things like `com.apple.developer.associated-domains` in the `BLACKLISTED_KEYS` is, but it's weird that it is included in `ENTITLEMENTS_TRANSFER_RULES` but it is deleted immediately after because of blacklisting.

By the way, I am trying to resign an AppStore IPA into an AdHoc IPA.
Any ideas?

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -703,7 +703,7 @@ function resign {
             # otherwise it interprets they key path as nested keys
             # TODO: Should be able to replace with echo ${KEY//\./\\\\.} and remove shellcheck disable directive
             # shellcheck disable=SC2001
-            PLUTIL_KEY=$(echo "$KEY" | sed 's/\./\\\\./g')
+            PLUTIL_KEY=$(echo "$KEY" | sed 's/\./\\\./g')
             plutil -insert "$PLUTIL_KEY" -xml "$ENTITLEMENTS_VALUE" "$PATCHED_ENTITLEMENTS"
 
             # Patch the ID value if specified


lacostej (approved):
Tested using

```sh
KEY=com.apple.developer.associated-domains
PLUTIL_KEY=`echo "$KEY" | sed 's/\./\\\\./g'`
echo $PLUTIL_KEY

PLUTIL_KEY=$(echo "$KEY" | sed 's/\./\\\\./g')
echo $PLUTIL_KEY

PLUTIL_KEY=$(echo "$KEY" | sed 's/\./\\\./g')
echo $PLUTIL_KEY
```

output

```
lacostej@Jeromes-MacBook-Pro ~/Code/OSS/Fastlane/fastlane (supply_tests) bash run
com\.apple\.developer\.associated-domains
com\\.apple\\.developer\\.associated-domains
com\.apple\.developer\.associated-domains
lacostej@Jeromes-MacBook-Pro ~/Code/OSS/Fastlane/fastlane (supply_tests) sh run
com\.apple\.developer\.associated-domains
com\.apple\.developer\.associated-domains
com\.apple\.developer\.associated-domains
```

lacostej:
Thanks @djsison !

djsison:
Any idea about the still existing problem?

mgrebenets:
@djsison 

This is the original PR that introduced blacklisted keys: https://github.com/fastlane/fastlane/issues/7330

The other 2 links which were used as a reference:
https://github.com/facebook/buck/issues/798
https://github.com/facebook/buck/pull/802/files

So my assumption would be that resigning your IPA with `buck` should have the same issue.
I'm only about to dive deeper into the `buck`, so I've never used their resigner yet.

mgrebenets:
@djsison 
So if I'm reading `buck` source code correctly, they have the steps like this:

- Select best matching profile from profile store
  - Assuming they manage some kind of profile store
  - the "best matching" is decided by number of criteria, such as presence/absence of certain keys (more here: https://github.com/facebook/buck/blob/aa106d8cf5672c1470482594eed054aa2ba816f6/src/com/facebook/buck/apple/ProvisioningProfileStore.java, line 133)
- Get only _mergeable_ entitlements form the profile
  - This is where blacklisted keys are filtered out
- Proceed using these filtered out entitlements for further resigning

What happens right now in `sigh resign ... --use-app-entitlements` is this:

1. Get entitlements from provisioning profile
2. Extract entitlements from the app binary
3. Merge and patch (aka "transfer") the app binary entitlements with profile entitlements
4. Remove blacklisted entries _from patched entitlements_

I believe the order of steps is what matters.
What seems to be the correct approach is to remove blacklisted keys from the profile entitlements _before_ merging with app binary entitlements.
The assumption here that app binary entitlements are correct and on of its keys should be removed in the process.

I'd say this is the correct flow:

1. Get entitlements from provisioning profile
2. Remove blacklisted entries _from provisioning profile entitlements_
3. Extract entitlements from the app binary
3. Merge and patch (aka "transfer") the app binary entitlements with profile entitlements

This way entitlements like `com.apple.developer.associated-domains` will be preserved, because those were part of app entitlements, thus will not be blacklisted.

PR #8282, opened 2017-02-17
Use eval_gemfile to install plugins

giginet:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Description

Adding plugins, `PluginManager` inserts `eval`. However, it is a bit dangerous.
This PR make using `eval_gemfile` instead.

### Motivation and Context

We prefer using `eval_gemfile` to load `Gemfile` recursively.
This way is more safety and recommended by bundler.

https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L36

PR #8299, opened 2017-02-19
Change eval to eval_gemfile for Pluginfile

nafu:
### Checklist

### Description

Apply #8282 changes to other places

### Motivation and Context

Now `code_to_attach` in `plugin_manager.rb` has changed to use `eval_gemfile` instead of `eval`.
That's why it would be better that related codes and documents also use `eval_gemfile`.


mfurtak (approved):
Thanks! 👍

PR #8320, opened 2017-02-21
Fix git_branch action issue

chuganzy:
### Description

* `git_branch` action does not follow its description. It says "If no branch could be found, this action will return nil" / "Returns the name of the current git branch", but
  * If no branch could be found, it raises an exception.
  * It does not return the actual branch name - returns ENV value first.

### What I did in this PR

* Fixed all issues above.
  * Return actual branch name first
  * Make it not to raise an exception.

### DIFF
--- a/fastlane/lib/fastlane/actions/git_branch.rb
+++ b/fastlane/lib/fastlane/actions/git_branch.rb
@@ -5,9 +5,14 @@ module Fastlane
 
     class GitBranchAction < Action
       def self.run(params)
-        env_vars = %w(GIT_BRANCH BRANCH_NAME TRAVIS_BRANCH BITRISE_GIT_BRANCH CI_BUILD_REF_NAME)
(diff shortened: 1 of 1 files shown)


asfalcone:
Hi @chuganzy. Can you briefly explain the issue you're fixing here? I was about to do a release with this code but will wait if this is a critical bug fix. 👍 Let me know!

chuganzy:
@asfalcone I've added the description! Please check it.

mfurtak:
Hi @chuganzy. Thanks for looking into this 👍 

Your change makes it so that Git is preferred over ENV, but the current behavior is that ENV is preferred over Git.

Unfortunately I don't think we can accept this because we need to maintain the behavior that users of this action will currently expect.

I have reproduced the crash you described, though, and I will submit a fix. I will also try to improve the documentation to describe the actual behavior. Thank you!

Issue #6806, opened 2016-10-31
Match can't find foreign branch when shallow clone option used


xNekOIx:
fastlane 1.106.2

### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Generate new signings using Match in the not default branch.
`bundle exec match appstore --git_branch not_default_branch`

Install them with `--shallow_clone` option.
`bundle exec match appstore --git_branch not_default_branch --readonly --shallow_clone --verbose`

Installation faults with the following stack trace:
```
bundler: failed to load command: match (/Users/some-user/.rbenv/versions/2.3.0/bin/match)
FastlaneCore::Interface::FastlaneCrash: [!] No code signing identity found and can not create a new one because you enabled `readonly`
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/fastlane_core-0.53.0/lib/fastlane_core/ui/interface.rb:132:in `crash!'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/fastlane_core-0.53.0/lib/fastlane_core/ui/ui.rb:14:in `method_missing'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/match-0.10.0/lib/match/runner.rb:72:in `fetch_certificate'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/match-0.10.0/lib/match/runner.rb:30:in `run'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/match-0.10.0/lib/match/commands_generator.rb:53:in `block (3 levels) in run'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/commander-4.4.0/lib/commander/command.rb:178:in `call'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/commander-4.4.0/lib/commander/command.rb:153:in `run'
  /Users/some-user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/commander-4.4.0/lib/commander/runner.rb:444:in `run_active_command'
[...]

PR #8341, opened 2017-02-22
fix Match foreign branch shallow clone issue

xNekOIx:
### Checklist

### Description
Fixes issues when try to use Match to shallow clone not default branch (when work with different teams). 
More here #6806

### DIFF
--- a/match/lib/match/git_helper.rb
+++ b/match/lib/match/git_helper.rb
@@ -6,7 +6,7 @@ module Match
       @dir = Dir.mktmpdir
 
       command = "git clone '#{git_url}' '#{@dir}'"
-      command << " --depth 1" if shallow_clone
+      command << " --depth 1 --no-single-branch" if shallow_clone
 
       UI.message "Cloning remote git repo..."
       begin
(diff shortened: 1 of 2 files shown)


lacostej:
Fine for me!

To confirm, @xNekOIx you ran that code in production? :)

lacostej (approved):

xNekOIx:
@lacostej yeap. checked. worked for my project.

KrauseFx:
Could you provide a little more context on what the `--no-single-branch` flag does and why this should be part of _match_?

xNekOIx:
@KrauseFx when you do a shallow clone with `git clone --depth 1` git truncates history only to the current branch default branch (master). Therefore when you do `git branch --list origin/#{branch.shellescape} --no-color -r` for your other team branch it returns nothing and you got the error described in the initial issue.
There is an option `--no-single-branch` which can only be used with `--depth` which solves this issue.
from the git docs:
_--depth <depth>
Create a shallow clone with a history truncated to the specified number of commits. Implies `--single-branch` unless `--no-single-branch` is given to ***fetch the histories near the tips of all branches***._
This way we will still do a shallow clone and will have an information about remote branches.

mfurtak:
Thanks for the contribution 🚀

Issue #8263, opened 2017-02-16
2.17.0 sigh fetching profiles duration


danielbowden:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I've just updated to `2.17.0` which has solved the Apple response issues from previous days. All builds are now passing however the runtime of `sigh` fetching profiles has greatly increased.

I've tried running my lane 9 times now across two machines, on two internet connections in two different cities, both based in Australia. Each time the lane has completed successfully but each run of `sigh` within my fastfile has consistently taken exactly 484s (~8 mins) to complete.

This has blown out our CI run times from 3-4 minutes to 26 minutes.

```
[13:25:32]: Successfully logged in
[13:25:32]: Fetching profiles...
[13:33:51]: Found 1 matching profile(s)
[13:33:51]: Downloading provisioning profile...
[13:33:55]: Successfully downloaded provisioning profile...
```

Happy that our builds are back passing again but just wanted to raise this issue. Suspect it could be more connection problems to Apple, but unsure.
 I've also tried removing `spaceship_itc_service_key.txt` before running.

### Environment
<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                       |
| ------------------- | ------------------------------------------- |
| OS                  | 10.12.3                                     |
| Ruby                | 2.2.2                                       |
| Bundler?            | true                                        |
| Git                 | git version 2.10.1 (Apple Git-78)           |
| Installation Source | ~/.rbenv/versions/2.2.2/bin/fastlane        |
| Host                | Mac OS X 10.12.3 (16D32)                    |
| Ruby Lib Dir        | ~/.rbenv/versions/2.2.2/lib                 |
| OpenSSL Version     | OpenSSL 1.0.2j  26 Sep 2016                 |
| Is contained        | false                                       |
| Is homebrew         | false                                       |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version       | 8.2.1                                       |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_AU.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |
</details>

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
[...]

Issue #8261, opened 2017-02-15
Deprecation of xcode managed profiles in 2.17.0 


nonemoreblack:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I noticed that the ability to retrieve xcode managed profiles was deprecated in 2.17.0. From the comments in https://github.com/fastlane/fastlane/pull/8251, it would appear this is due to Apple no longer returning them in the latest API. This is a big problem for the way we work, and we won't be able to update as a result (fortunately, as of this morning, 2.16.0 started working again).

It would seem that these profiles must still be fetchable in some way, as xcode must fetch them. The old API is still returning them in 2.16.0. Could retaining this capability be investigated further? If there is no way going forward, we will be forced to change our Apple signing processes, which will be no small task.

Thanks for the quick work on 2.17.0, btw.

asfalcone:
@nonemoreblack Thanks for letting us know that this is an issue for you. We'll try to figure out if there is a way to reliably access the xcode managed provisioning profiles via API. 👍

asfalcone:
This should be fixed in _faslane_ 2.19.0. We reverted to the old behavior and introduced the `SPACESHIP_AVOID_XCODE_API` env variable. You'll want to use the default behavior to retreive the xcode managed profiles but if this API has issues again, you can set SPACESHIP_AVOID_XCODE_API=1 and at least have some functionality 👍. Please give it a try and let me know, thanks!

nonemoreblack:
I installed 2.19.1 and it works perfectly. Thanks so much for fixing this!

KrauseFx:
Great to hear, thanks for confirming @nonemoreblack 👍

PR #8251, opened 2017-02-15
remove developerservices2 in favor of developer.apple.com

snatchev:
This PR is a _work-in-progress_ to address issues in #8209

Migrated to developer.apple.com:
* provisioning profiles
* list_teams

removed provisioning profiles CSRF token dance since we are using 1 cookie

TODO:
- [x] update all of the stubs and failing tests.
- [x] update `create_device` to accept device type parameter
- [ ] determine if any other endpoints are affected

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Description
* Change URLs and response parsing to use new API
* Fixtures had to change to reflect the new responses from the API.

### Motivation and Context
Apple seems to have deprecated `developerservices2.apple.com`. This PR migrates to using `developer.apple.com` API. See issue #8209.

### Thanks
@lacostej  @asfalcone @hjanuschka for their help.

PR #8359, opened 2017-02-24
use Xcode API for listing provisioning profiles

snatchev:
### Checklist

### Description
This PR reverts some of the behavior from #8251. Spaceship will now use the Xcode API endpoint
`https://developerservices2.apple.com/services/QH65B2/ios/listProvisioningProfiles.action` for getting a list of provisioning profiles.

This addresses issues #8263 #8261

We will use an environment variable `SPACESHIP_AVOID_XCODE_API` to disable this behavior should it have problems again in the future. If the Xcode API is not used, we will perform the lookups necessary to get all of the `appId`s for each provisioning profile.

### Motivation and Context
PR #8251 was introduced to work around long-running stability issues with the API. This change introduced an N+1 query problem for finding provisioning profiles by their app's bundle id. The problem is that the developer portal API does not return `appId` along with a provisioning profile and its search feature does not work for searching by `appId` (even though `appId` can be found in the search query 🤔 )


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

spaceship/lib/spaceship/portal/portal_client.rb
```diff
@@ -465,6 +465,22 @@ def provisioning_profiles(mac: false)
       end
     end
 
+    ##
+    # this endpoint is used by Xcode to fetch provisioning profiles.
+    # The response is an xml plist but has the added benefit of containing the appId of each provisioning profile.
+    def provisioning_profiles_via_xcode_api(mac: false)
```
lacostej: Should `SPACESHIP_AVOID_XCODE_API` be managed in `provisioning_profile` or here in the `provisioning_profile_details` implementation, delegated to 2 different methods?

spaceship/lib/spaceship/portal/portal_client.rb
```diff
@@ -465,6 +465,22 @@ def provisioning_profiles(mac: false)
...
+      req = request(:post) do |r|
+        r.url "https://developerservices2.apple.com/services/#{PROTOCOL_VERSION}/#{platform_slug(mac)}/listProvisioningProfiles.action"
+        r.params = {
+          teamId: team_id,
+          includeInactiveProfiles: true,
+          onlyCountLists: true
+        }
+      end
+
+      parse_response(req, 'provisioningProfiles')
```
lacostej: if I recall there was some weird cs_rf  code management in the former implementation right? Is this gone for good?

snatchev: My understanding is that the csrf code was only necessary for modifying provisioning profiles. Since the create/delete/repair requests go to the developer portal domain, we don't need it for those requests. I will double check to make sure my understanding is correct.

I will test more thoroughly

spaceship/lib/spaceship/portal/portal_client.rb
snatchev: I had `SPACESHIP_AVOID_XCODE_API` in the client initially but decided against it.
My thinking was this: The client is just a library of valid api requests and their parsed responses. Each method makes 1 request and shouldn't need to know much about the state of the application.
As such, it should be up to the caller (i.e. `ProvisioningProfile` class) to contain the logic of deciding which request to make.

Does that seem reasonable?

asfalcone: I like this design where the client just has a bunch of methods for other callers to use 👍

spaceship/lib/spaceship/portal/portal_client.rb
asfalcone: The get all methods (`provisioning_profiles`, etc) shouldn't need to make the csrf call, since calling them is what gets the token.

spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -260,16 +260,19 @@ def create!(name: nil, bundle_id: nil, certificate: nil, devices: [], mac: false
         #  If you're calling this from a subclass (like AdHoc), this will
         #  only return the profiles that are of this type
         # @param mac (Bool) (optional): Pass true to get all Mac provisioning profiles
-        # @param xcode DEPRECATED
+        # @param xcode (Bool) (optional): Pass true to include Xcode managed provisioning profiles
```
asfalcone: Should we mention this won't work if you have set `SPACESHIP_AVOID_XCODE_API` and add the warning output back in for that case?

spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -484,9 +487,13 @@ def certificates
       end
 
       def app
-        fetch_details
+        if raw_data.has_key?('appId')
```
asfalcone: Nice abstraction to keep the `app` method 👍

spaceship/lib/spaceship/portal/portal_client.rb
```diff
@@ -465,6 +465,22 @@ def provisioning_profiles(mac: false)
       end
     end
 
+    ##
+    # this endpoint is used by Xcode to fetch provisioning profiles.
+    # The response is an xml plist but has the added benefit of containing the appId of each provisioning profile.
```
asfalcone: Add something to the comment to the effect of "Using this method is more efficient than `provisioning_profiles` because no secondary api calls are necessary" or comment on `provisioning_profiles` talking about it being less efficient.

snatchev: comment added.

asfalcone (approved):

asfalcone:
LGTM! Add some documentation about the ENV variable as discussed and squash and merge! 🎉

spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -260,16 +263,19 @@ def create!(name: nil, bundle_id: nil, certificate: nil, devices: [], mac: false
         #  If you're calling this from a subclass (like AdHoc), this will
         #  only return the profiles that are of this type
         # @param mac (Bool) (optional): Pass true to get all Mac provisioning profiles
-        # @param xcode DEPRECATED
+        # @param xcode (Bool) (optional): Pass true to include Xcode managed provisioning profiles
         def all(mac: false, xcode: false)
-          profiles = client.provisioning_profiles(mac: mac).map do |profile|
-            self.factory(profile)
+          if ENV['SPACESHIP_AVOID_XCODE_API']
```
lacostej: We have a`FastlaneCore::Env.truthy?` method we can use here for consistency

lacostej: i.e. if we want "no" to mean `false`

spaceship/spec/provisioning_profile_spec.rb
```diff
@@ -54,6 +54,30 @@
     end
   end
 
+  describe '#all via xcode api' do
+    around(:all) do |example|
+      switch = ENV['SPACESHIP_AVOID_XCODE_API']
+      example.run
```
lacostej: do we want to introduce a `rescue` after the `example.run`?

lacostej (changes requested):
a couple of nitpicks

spaceship/spec/provisioning_profile_spec.rb
snatchev: Could you explain this further? I'm not clear on what case a `rescue` will satisfy.

lacostej: I mean to ensure we set back the `ENV[]` to its original state. Otherwise it breaks the next test as well.

spaceship/lib/spaceship/portal/provisioning_profile.rb
snatchev: Cool, did not know about that. Will update.

spaceship/spec/provisioning_profile_spec.rb
lacostej: if for some reason the `example.run` fails hard.

snatchev: Ah, good catch. The docs indicate `around` handles this gracefully:
https://www.relishapp.com/rspec/rspec-core/docs/hooks/around-hooks#an-around-hook-continues-to-run-even-if-the-example-throws-an-exception

lacostej: Ah cool feature

lacostej (approved):

spaceship/lib/spaceship/portal/provisioning_profile.rb
snatchev: I can't confirm this in test, but it possible to use `spaceship` without `fastlane_core`? I do not see a `require 'fastlane_core'` but I may be missing something.

mfurtak: `spaceship` is still intended to avoid `fastlane_core` dependencies as far as I know. 👍

koenpunt:
Is it possible that fastlane doesn't use this new API by default? Because for me downloads and even listing the profiles can take up to 20 minutes...

PR #8363, opened 2017-02-25
[spaceship] refactor out fetch_details to allow users of the API to access the details without having to call a separate method

lacostej:
### Checklist

### Description
This PR adjusts the behaviour from #8251. 

The PP class has a `device_details` field that isn't populated by default as it causes a request. The `fetch_details` method is there to fill it but isn't public. 

It would make more sense to make the `device_details` accessor fill on demand.

It would also make the implementation changes in `provisioning_profile` proposed by #8359 slightly more in line with other methods of the same class (i.e. direct use of `device_details`)


mfurtak:
How does this line up with the work being done in #8359?

lacostej:
@mfurtak thought of this change while reading #8359, added a better description

snatchev (approved):
I like this refactor. 👍

PR #8369, opened 2017-02-27
Beautify Logging output

hjanuschka:
Wraps line according to terminal size.

turns this:
bildschirmfoto 2017-02-27 um 10 23 33

into this:
bildschirmfoto 2017-02-27 um 10 22 05

test script:

```ruby
require "fastlane_core"

FastlaneCore::UI.important("a"*2048)
```

### DIFF
--- a/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
+++ b/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
@@ -21,12 +21,9 @@ module FastlaneCore
         timestamp = format_string(datetime, severity)
 
         colors = msg.scan(/(\e\[.*?m)/)
-        if colors && colors.length > 0
-          colors.each do |color|
-            msg.delete!(color.first)
-            msg.delete!(color.last)
-          end
-        else
+        msg = FastlaneCore::Helper.strip_ansi_colors(msg)
+
+        unless colors && colors.length > 0
           colors = ["", ""]
         end
 
--- a/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
+++ b/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
@@ -17,7 +17,27 @@ module FastlaneCore
       end
 
       @log.formatter = proc do |severity, datetime, progname, msg|
-        "#{format_string(datetime, severity)}#{msg}\n"
+        terminal_length = TTY::Screen.width
+        timestamp = format_string(datetime, severity)
+
+        colors = msg.scan(/(\e\[.*?m)/)
+        if colors && colors.length > 0
+          colors.each do |color|
+            msg.delete!(color.first)
+            msg.delete!(color.last)
+          end
+        else
+          colors = ["", ""]
+        end
+
+        max_msg_length = terminal_length - timestamp.length - 5 # 5 = padding
+        msg.gsub!(/(\S{#{max_msg_length}})(?=\S)/, '\1 ')
+        components = msg.scan(/.{1,#{max_msg_length}}(?:\s+|$)/)
+        final_msg = ""
+        components.each do |choped|
+          final_msg << timestamp << "#{colors.first[0]}#{choped}#{colors.last[0]}\n"
+        end
+        final_msg
       end
 
       @log


mfurtak:
Hm, I think I would not be in favor of this change. With this conversion, you lose information about the boundaries of individual log statements, making things tougher to interpret in my opinion. There is a benefit to knowing that every time there is a timestamp, a new piece of output was produced.

Also, if there are tools out there that interpret fastlane logs, changing the line break patterns is likely to mess them up.

KrauseFx:
@mfurtak: I agree with you, this information is crucial. How about we show it like the following instead?

```
[10:22:44]: Unfortunately we can't connect to the
            server, please try again later or file
            a new issue on GitHub
[10:22:44]: An error occurred
```

or

```
[10:22:44]: Unfortunately we can't connect to the
             server, please try again later or file
             a new issue on GitHub
[10:22:44]: An error occurred
```

or


```
[10:22:44]: Unfortunately we can't connect to the
              server, please try again later or file
              a new issue on GitHub
[10:22:44]: An error occurred
```

mfurtak:
@KrauseFx I still think I would be against it for the second reason, which is that we shouldn't change our output format to suddenly contain more line breaks and whitespace.

One more reason: If you manually add line breaks and spacing for a particular terminal width, this output will look bad if displayed at any other width. This could happen in the same terminal window, or when copied & pasted to another program/editor.

IMO it is better to let the lines be long and be wrapped automatically by the container they are displayed in.

KrauseFx:
I agree with you that the correct approach shouldn't include new-lines. However we'd add new-lines only in the `shell.rb` class, which means, that if we ever have other outputs (e.g. app) this wouldn't be included.

Currently we use the following code in many areas in the code base, e.g.
```ruby
      UI.error("An error occured while verifying your certificates and profiles with the Apple Developer Portal.")
      UI.error("If you already have your certificates stored in git, you can run `fastlane match` in readonly mode")
      UI.error("to just install the certificates and profiles without accessing the Dev Portal.")
      UI.error("To do so, just pass `readonly: true` to your match call.")
```

or

```
      ui.error ""
      ui.error "The best solution is to use the self-contained fastlane version."
      ui.error "Which ships with a bundled OpenSSL,ruby and all gems - so you don't depend on system libraries"
      ui.error " - Use One-Click-Installer:"
      ui.error "    - download fastlane at https://download.fastlane.tools"
      ui.error "-----------------------------------------------------------"
      ui.error "    - extract the archive and double click the `install`"
      ui.error "-----------------------------------------------------------"
      ui.error " - Use Homebrew"
      ui.error "    - update brew with `brew update`"
      ui.error "    - install fastlane:"
      ui.error "-----------------------------------------------------------"
      ui.error "      - 🚀 `brew cask install fastlane` 🚀"
      ui.error "-----------------------------------------------------------"
      ui.error "for more details on ways to install fastlane please refer the documentation:"
      ui.error "-----------------------------------------------------------"
      ui.error "        🚀       https://docs.fastlane.tools          🚀   "
      ui.error "-----------------------------------------------------------"
      ui.error ""
      ui.error "You can also install a new version of Ruby"
      ui.error ""
      ui.error "- Make sure OpenSSL is installed with Homebrew: `brew update && brew upgrade openssl`"
      ui.error "- If you use system Ruby:"
      ui.error "  - Run `brew update && brew install ruby`"
      ui.error "- If you use rbenv with ruby-build:"
      ui.error "  - Run `brew update && brew upgrade ruby-build && rbenv install 2.3.1`"
      ui.error "  - Run `rbenv global 2.3.1` to make it the new global default Ruby version"
      ui.error "- If you use rvm:"
      ui.error "  - First run `rvm osx-ssl-certs update all`"
      ui.error "  - Then run `rvm reinstall ruby-2.3.1 --with-openssl-dir=/usr/local`"
      ui.error ""
      ui.error "If that doesn't fix your issue, please google for the following error message:"
      ui.error "  '#{e}'"
      ui.error "-----------------------------------------------------------------------"
```

All of this just to make the output look nicer (ok, the second output is a list, but still)

Copy and pasting isn't ideal, I agree with that, but the main use-case is displaying text and I feel like we should optimize for that, as we don't really break the other behavior, but it's just less ideal in the favor of having an improve main use-case.

mfurtak:
Here is what output looks like on Circle CI. I think its "terminal" settings may be requesting a specific (small) column width.

image

Compare to before:

image

Also, I think there may be a problem with the text processing, as `m`s from the output are missing 😅

hjanuschka:
@mfurtak  😂 about the m's - going to find them
@DanToml  what terminal width is set ?

mfurtak:
Overall I remain against this behavior. If we decide to bring it in, I would strongly suggest that it be opt-in so that we don't affect people with the change by accident.

lacostej:
I am not very fond of the feature. I spend lots of time in the terminal and if fastlane wraps the output of a command and prints it, I expect it to not change it. Otherwise it will modify the output redirected to a file where I wouldn't have the column limitation, etc. Even CI builds might look weird. :-1:

asfalcone:
I am going to pile on and say that I don't think this feature is right for _fastlane_, for all of the reasons mentioned above by @mfurtak @lacostej . Thanks for submitting the PR @hjanuschka and I'm glad we could discuss its merits as a team :)

hjanuschka:
ok so just close it 👍

Issue #8379, opened 2017-02-27
Fastlane assumes ./fastlane as integration folder and doesn't check for ./.fastlane


mgrebenets:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Some fastlane tools assume that fastlane folder is `./fastlane` after checking for fastlane integration.

For example, this is from `scan/lib/scan/options.rb`
```ruby
containing = Helper.fastlane_enabled? ? './fastlane' : '.'

# Later down the file - default output path
default_value: File.join(containing, "test_output"))
```

When fastlane is set up in `.fastlane` dot-prefixed hidden folder, the default output will point to folder that doesn't exist.

The same issue will happen for `deliver`, `screengrab` and `snapshot` tools.

This is also very similar to https://github.com/fastlane/fastlane/pull/8233/files

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                       |
| ------------------- | ------------------------------------------- |
| OS                  | 10.12.2                                     |
| Ruby                | 2.2.2                                       |
| Bundler?            | true                                        |
| Git                 | git version 2.10.1 (Apple Git-78)           |
| Installation Source | ~/.rvm/gems/ruby-2.2.2/bin/fastlane         |
| Host                | Mac OS X 10.12.2 (16C68)                    |
| Ruby Lib Dir        | ~/.rvm/rubies/ruby-2.2.2/lib                |
| OpenSSL Version     | OpenSSL 1.0.2j  26 Sep 2016                 |
| Is contained        | false                                       |
| Is homebrew         | false                                       |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version       | 8.2.1                                       |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_AU.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
skip_docs

desc "Runs the unit tests and other verifications for the fastlane repo"
lane :test do |options|
  validate_repo
end

desc "Increment the version number of this gem"
lane :bump do |options|
  ensure_git_branch(branch: "master")
  ensure_git_status_clean

[...]

PR #8380, opened 2017-02-27
Fix hardcoded fastlane paths

mgrebenets:
### Checklist

### Description
Change hard-coded "./fastlane" to use `FastlaneCore::FastlaneFolder.path` in `deliver`, `scan`, `screengrab` and `snapshot`.

### Motivation and Context
Fix for https://github.com/fastlane/fastlane/issues/8379
I had to modify number of specs to stub `FastlaneCore::FastlaneFolder.path` before execution.

### DIFF
--- a/scan/spec/test_command_generator_spec.rb
+++ b/scan/spec/test_command_generator_spec.rb
@@ -78,7 +78,7 @@ describe Scan do
 
     describe "Supports toolchain" do
       it "should fail if :xctestrun and :toolchain is set" do
-        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
+        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(".")
         expect do
           Fastlane::FastFile.new.parse("lane :test do
             scan(
@@ -429,7 +429,7 @@ describe Scan do
                                      ])
       end
       it "should raise an exception if two build_modes are set" do
-        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
+        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(".")
         expect do
           Fastlane::FastFile.new.parse("lane :test do
             scan(
--- a/deliver/lib/deliver/commands_generator.rb
+++ b/deliver/lib/deliver/commands_generator.rb
@@ -124,7 +124,7 @@ module Deliver
           options = FastlaneCore::Configuration.create(deliverfile_options(skip_verification: true), options.__hash__)
           options.load_configuration_file("Deliverfile")
           Deliver::Runner.new(options, skip_version: true) # to login...
-          containing = FastlaneCore::Helper.fastlane_enabled? ? './fastlane' : '.'
+          containing = FastlaneCore::Helper.fastlane_enabled? ? FastlaneCore::FastlaneFolder.path : '.'
           path = options[:screenshots_path] || File.join(containing, 'screenshots')
           Deliver::DownloadScreenshots.run(options, path)
         end
@@ -140,7 +140,7 @@ module Deliver
           options = FastlaneCore::Configuration.create(deliverfile_options(skip_verification: true), options.__hash__)
           options.load_configuration_file("Deliverfile")
           Deliver::Runner.new(options) # to login...
-          containing = FastlaneCore::Helper.fastlane_enabled? ? './fastlane' : '.'
+          containing = FastlaneCore::Helper.fastlane_enabled? ? FastlaneCore::FastlaneFolder.path : '.'
           path = options[:metadata_path] || File.join(containing, 'metadata')
           res = ENV["DELIVER_FORCE_OVERWRITE"]
           res ||= UI.confirm("Do you want to overwrite existing metadata on path '#{File.expand_path(path)}'?")
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -9,7 +9,7 @@ module Scan
     end
 
     def self.available_options
-      containing = Helper.fastlane_enabled? ? './fastlane' : '.'
+      containing = Helper.fastlane_enabled? ? FastlaneCore::FastlaneFolder.path : '.'
 
       [
         FastlaneCore::ConfigItem.new(key: :workspace,
--- a/screengrab/lib/screengrab/commands_generator.rb
+++ b/screengrab/lib/screengrab/commands_generator.rb
@@ -48,7 +48,7 @@ module Screengrab
 
         c.action do |args, options|
           require 'screengrab/setup'
-          path = (Screengrab::Helper.fastlane_enabled? ? './fastlane' : '.')
+          path = Screengrab::Helper.fastlane_enabled? ? FastlaneCore::FastlaneFolder.path : '.'
           Screengrab::Setup.create(path)
         end
       end
--- a/snapshot/lib/snapshot/commands_generator.rb
+++ b/snapshot/lib/snapshot/commands_generator.rb
@@ -44,7 +44,7 @@ module Snapshot
 
         c.action do |args, options|
           require 'snapshot/setup'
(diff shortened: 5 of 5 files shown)


mfurtak:
This seems like a good change to me 👍 

Aside from the changes to allow the existing tests to pass, how can we ensure that this doesn't break things for our users?

fastlane/spec/unused_options_spec.rb
```diff
@@ -4,10 +4,12 @@
...
 
       Fastlane::ActionsList.all_actions do |action, name|
-        next unless action.available_options.kind_of?(Array)
-        next unless action.available_options.last.kind_of?(FastlaneCore::ConfigItem)
-
         it "No unused parameters in '#{name}'" do
+          allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return("./fastlane")
+
+          next unless action.available_options.kind_of?(Array)
+          next unless action.available_options.last.kind_of?(FastlaneCore::ConfigItem)
```
mgrebenets: The call to `available_options` will use variable called `containing`, which is defined as `containing = Helper.fastlane_enabled? ? FastlaneCore::FastlaneFolder.path : '.'` after the change.

So in this case the `FastlaneCore::FastlaneFolder.path` has to be mocked with `allow` before call to `available_options`.

I tried to move `allow` to outer scope. I have put it in a `before do/end` block right inside `describe Fastlane`, `describe Fastlane::Action` as well as `describe "No unused options"`. 

For all those cases the `before do/end` block wasn't called 😕  and `available_options` would get `nil` when resolving `containing` variable...

The block would only be called when put inside `it` spec, but in that case I had to move the `available_options` call inside the `it` block too.

I'm sure I'm missing something here, can't figure out why the `before do/end` block would not get called when put inside `describe` block.

scan/spec/test_command_generator_spec.rb
```diff
@@ -1,5 +1,8 @@
 describe Scan do
+  require "rspec/mocks/standalone"
```
mgrebenets: Without this line I was getting these errors:

```
  1) Scan Scan::TestCommandGenerator raises an exception when project path wasn't found
     Failure/Error: allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(".")
       The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported.
     # ./scan/spec/test_command_generator_spec.rb:4:in `block (2 levels) in <top (required)>'
```

After googling around I've found 3 possible solutions. 2 of those would need a `require` line and the third one was involving use of `RSpec::Mocks.with_temporary_scope do/end` wrapper, but it didn't work for me.

http://stackoverflow.com/questions/21235269/method-stubbing-on-beforeall

mgrebenets:
@mfurtak I'll have more look into adding more tests to ensure that current functionality is not impacted.

scan/spec/test_command_generator_spec.rb
mgrebenets: I've also moved it right inside before(:all) block to make its intent more clear.

fastlane/spec/unused_options_spec.rb
mgrebenets: Tried to put`before(:all/:each) do/end` in different scopes - same results - not getting called. There must be some fundamental thing I'm missing about this particular spec.

mfurtak: OK, I see what's happening. The looping done by `Fastlane::ActionsList.all_actions do |action, name|` happens during spec definition time, well before the `before` blocks get evaluated during execution. That's why the calls to `next` need to be moved inside of the body of the spec example... otherwise they are evaluated before the mocking can take place.

This seems like a reasonable compromise 👍

scan/spec/test_command_generator_spec.rb
mfurtak: Got it. This is pretty unusual, so putting a comment before the `require` to explain why its necessary would be good for our future selves 👍

giginet (approved):

fastlane/spec/action_metadata_spec.rb
```diff
@@ -3,6 +3,10 @@
 describe Fastlane::Action do
   Fastlane::ActionsList.all_actions do |action, name|
     describe name do
+      before do
+        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return("./fastlane")
```
nafu: @mgrebenets 

This is required because of below behavior of `FastlaneCore::FastlaneFolder.path`, right?

https://github.com/fastlane/fastlane/blob/37fb600751bfc17d49da1d4f4369cee0adab771b/fastlane_core/lib/fastlane_core/fastlane_folder.rb#L12
```ruby
value = nil if Helper.is_test? # this is required, as the tests would use the ./fastlane folder otherwise
```

I think this behavior should be deleted. What do you think?
I have an idea, so I'll PR for this.
After that, I'd like to discuss more about the way we test at this part.

nafu (changes requested):

fastlane/spec/action_metadata_spec.rb
nafu: @mgrebenets 

I created the PR!
Could you take a look at this? https://github.com/fastlane/fastlane/pull/8484

mgrebenets: @nafu Right, this all boils down to that line of code.
After applying my changes, I've reaslized that I only mock `./fastlane` path, but I should also mock the `fastlane` path, so this PR definitely needs improvements in terms of rspecs.

nafu: @mgrebenets https://github.com/fastlane/fastlane/pull/8484 is merged, so you don't need these mock anymore😄 Could you rebase and update this PR?

mgrebenets:
Lovely! 👍. You mean I don't have to be mocking things? That's cool! Will
update first thing tomorrow.

mgrebenets:
@googlebot I signed it!

hjanuschka:
@mgrebenets has this been solved/implemented by https://github.com/fastlane/fastlane/pull/8484 ?

nafu:
@hjanuschka No, this resolve different problems. 🙂

nafu (approved):
@mgrebenets
Looks good to me🚀

hjanuschka (approved):

hjanuschka:
@nafu thx for clarification ;) lgtm

mgrebenets:
@hjanuschka, @nafu 
I guess I still have a small tech debt with regards to adding some new specs.
Would you guys have any suggestions in that regards?
I wasn't able to come with some new test scenarios so far, I mostly made sure the correct path is used and refactored unit tests after the change - haven't added new tests.

Also, CI failed due to expired signing identity... I hope rebuild will fix that.

mgrebenets:
Still fails the CI...

```
Your provisioning profile (a287f946-0c82-434d-b034-d59682b15a42) has expired.
```

I'm not sure where those profiles even come from.

mfurtak:
These profiles are being used in various test and example directories, but it seems like a new occurrence that Circle would find and try to install them. I just put in a change to see what would happen with the build for my branch, and I didn't run into this problem 😢 

This CI problem seems somewhat isolated at the moment, can we try some Git rebasing, etc. to try to fool the system? Maybe bring this in as a new PR if absolutely necessary?

mgrebenets:
@mfurtak
Rebasing fixed the problem 👍 
That should've been the first thing to do - needs to be like a habit :)

mfurtak:
Nice! I think @DanToml flicked some switches on your behalf as well 😉 👍

mfurtak (approved):

nafu:
@mgrebenets @mfurtak @DanToml 🚀

mgrebenets:
OMG, I feel like we've just achieved singularity 😀

We wouldn't need release engineers if we had a bot like this one where I work 😂

KrauseFx:
@mgrebenets the bot keeps coming back 🎉

PR #8394, opened 2017-02-28
Sigh allows members to download their own certs and  doesn't error ou…

### LINKED ISSUE #8385: Member account can't download provisioning profiles with sigh
### New Issue Checklist

- [-] Updated fastlane to the latest version
- [-] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Running `sigh` as a Member (not Admin) doesn't allow fastlane to download any existing provisioning profiles.
I'm aware this is the "expected" behaviour based on several issues I've read already. Given the permission tree at that Apple has on developer portal this seems not appropriate. 
As a Member I am able to download any existing provisioning profiles (given that I have installed the private keys and certificates, which I "manually" install). And so I'd expect fastlane to be able to do the same.

Let me know if this is unrealistic or has any limitation I don't see straight (maybe API limitation).
Also happy to help with anything I can.

(does not apply)

### Environment
(doesn't apply either)

joshdholtz:
…t if the user does not have enough permissions to download certificate

Fixes #8385 

## The Problem

The issue is happening here - https://github.com/fastlane/fastlane/blob/master/sigh/lib/sigh/runner.rb#L100-L114

1. Downloads and filters profile(s)
2. Iterates over all certificates in all the profiles
3. Attempts to download all the certificates to see if they are installed
4. Only installs the profile if the local machine can use it (so if the certificates match)

**Step 3** is where things are going :boom: for you.

An **Admin** has the ability to download all certificates on all the profiles. A **Member** can only download their own certificate. I am not sure if this is recent change or not. An **Admin** and a **Member** has the ability to download all profiles on all the profiles (event if the certificates are installed for that profile).

So, you are correct in saying that a **Member** should be able to download all profile even if they aren't able to use them. 

## The Solution

The solution here is to put a begin/rescue are part of the code I have linked above. It is still a valid error/warning since a user may not have access to that certificate but it should not halt the fastlane process as a valid cert may still exist in the array somewhere.

```sh
[10:19:07]: Starting login with user 'josh+iosmember@rokkincat.com'
[10:19:10]: Successfully logged in
[10:19:10]: Fetching profiles...
[10:19:19]: Verifying certificates...
[10:19:19]: Cannot download cert 5T69GDU72Q - User josh+iosmember@rokkincat.com (Team ID WPBN76YKX2) doesn't have enough permission for the following action: download_certificate (You are not permitted to download this certificate.)
[10:19:20]: Cannot download cert Y284AATB34 - User josh+iosmember@rokkincat.com (Team ID WPBN76YKX2) doesn't have enough permission for the following action: download_certificate (You are not permitted to download this certificate.)
[10:19:20]: Cannot download cert L8F463V45L - User josh+iosmember@rokkincat.com (Team ID WPBN76YKX2) doesn't have enough permission for the following action: download_certificate (You are not permitted to download this certificate.)
[10:19:21]: Cannot download cert PBL2B7BD7M - User josh+iosmember@rokkincat.com (Team ID WPBN76YKX2) doesn't have enough permission for the following action: download_certificate (You are not permitted to download this certificate.)
[10:19:21]: Cannot download cert TE4B367B7M - User josh+iosmember@rokkincat.com (Team ID WPBN76YKX2) doesn't have enough permission for the following action: download_certificate (You are not permitted to download this certificate.)
[10:19:23]: Found 1 matching profile(s)
[10:19:23]: Downloading provisioning profile...
[10:19:24]: Successfully downloaded provisioning profile...
[10:19:25]: Installing provisioning profile...
```

### DIFF
--- a/sigh/lib/sigh/runner.rb
+++ b/sigh/lib/sigh/runner.rb
@@ -97,12 +97,28 @@ module Sigh
 
       return results if Sigh.config[:skip_certificate_verification]
 
+      UI.message "Verifying certificates..."
       return results.find_all do |a|
-        # Also make sure we have the certificate installed on the local machine
         installed = false
-        a.certificates.each do |cert|
+
+        # Attempts to download all certificats from this profile
+        # for checking if they are installed.
+        # `cert.download_raw` can fail if the user is a
+        # "member" and not an a "admin"
+        raw_certs = a.certificates.map do |cert|
+          begin
+            raw_cert = cert.download_raw
+          rescue => error
+            UI.important("Cannot download cert #{cert.id} - #{error.message}")
+            raw_cert = nil
+          end
+          raw_cert
+        end.compact
+
+        # Makes sure we have the certificate installed on the local machine
+        raw_certs.each do |raw_cert|
           file = Tempfile.new('cert')
-          file.write(cert.download_raw)
+          file.write(raw_cert)
           file.close
           if FastlaneCore::CertChecker.installed?(file.path)
             installed = true


joshdholtz:
Hmm, now that I think about this... I'm not sure _if_ this is the best solution yet. `SIGH_SKIP_CERTIFICATE_VERIFICATION` can be set to get around this issue but that is not a super transparent solution to the user.

Verification is on by default so it failing _does_ make sense but I think maybe a more elegant error message (that still fails) could be shown to the user saying to enable `SIGH_SKIP_CERTIFICATE_VERIFICATION`.

Thoughts?

mfurtak (approved):
LGTM from a code point of view. 👍

mfurtak:
Bounced the CI job now that the Internet works again 😉

Issue #6818, opened 2016-11-01
Pilot : submit for external review crashes when it's currently in review or already approved


gcamp:
I'm using `Pilot::BuildManager.new.distribute(config)` directly in my Fastfile. Not sure if this is bad practice but it was my best way to reuse code from Pilot and not re-implement everything myself using spaceship. 

There's a problem with how pilot is handling the submit for external review part. Here is the current timeline of events : 
- Submit a beta for a new build train. 
  - The call for `client.submit_testflight_build_for_review!` is working. 
  - However the call `build_train.update_testing_status!` throws an error "This build could not be used for external testing because the build is not approved." This is a valid error, however it should probably be catched in some way and shown to the user as a warning.
- The previous beta that was sent to review is either currently in review or approved.
  - The call for `client.submit_testflight_build_for_review!` is not working and instead a error is thrown "ITC.response.error.INVALID_REQUEST" (can't submit a build to review if it's already in review)
  - The call `build_train.update_testing_status!` is not called at all since an error is thrown before. 

Proposed solution : 
- Add attributes in Spaceship::Tunes::Build for `betaState` and `internalState`. Current values I've seen are "inactive", "submitForReview", "waiting", "active".  I'm missing the in review and approved state. 
- pilot should take into account this value to either not send the build to review, or not send to beta testers if the build is not approved. 

I can do a pull request. Just wanted to discuss it first.

TKBurner:
@gcamp I think these make sense as changes. Can you submit a lightweight PR to address this first in spaceship? And then ask in that PR about Pilot? We really appreciate you working with us 👍 Thanks!

joeydong:
I've also run into this workflow problem. The beta app review can take 24-48 hours so it's unreasonable to keep `pilot` running in the background like we currently do while waiting for the build to be processed.

I agree once we have the right information coming from Spaceship, we can provide a cleaner exit code for `pilot`. Then it will be up to the user to run `pilot distribute` once they get an email notification that their build has been approved by beta app review.

On the other hand, nothing is really stopping us from keeping `pilot` running for a day or so while waiting for the review so maybe we can make this an advanced feature for those who are running `pilot` on a build machine that stays connected all the time anyway.

gcamp:
Still on my todo list, sorry for the delay

BroderickLee:
When will this issue be resolved? I really hope it can be fixed as soon as possible. Thanks!!!

cpk1:
[...]

PR #8395, opened 2017-02-28
Pilot external testing

cpk1:
### Checklist

### Description

This is to address https://github.com/fastlane/fastlane/issues/6818
I've added the "betaState" attribute to Spaceship::Tunes:Build
In Pilot::BuildManager I've added the "betaState" status to the builds output.
Additionally I've added the parameter "external_available" to "distribute_build" this will be checked before calling "uploaded_build.build_train.update_testing_status!" 

### Motivation and Context

When building from jenkins and using a post-build action with "fastlane pilot upload distribute_external TRUE" pilot will always return an iTC error because it is impossible to turn on a build for external testing before beta app review, this in turn will cause the job to be marked as failed despite everything occurring as expected - build was uploaded and submitted for beta app review successfully.

https://github.com/fastlane/fastlane/issues/6818

I've uploaded and attempted to distribute for external testers at the same time (fastlane pilot upload distribute_external TRUE) and tried to distribute internally and externally builds that were already uploaded and approved and not approved for external distribution ("fastlane pilot distribute distribute_external TRUE" and "fastlane pilot distribute")


pilot/lib/pilot/build_manager.rb
```diff
@@ -203,9 +215,17 @@ def distribute_build(uploaded_build, options)
...
       # Submit for beta testing
-      type = options[:distribute_external] ? 'external' : 'internal'
-      uploaded_build.build_train.update_testing_status!(true, type, uploaded_build)
-      return true
+      # We need to check if the user is attempting to turn on external builds and the status for external builds
+      # If we attempt to make an unapproved external build available iTC will return an error
+      if external_available || !options[:distribute_external]
+        type = options[:distribute_external] ? 'external' : 'internal'
+        uploaded_build.build_train.update_testing_status!(true, type, uploaded_build)
+        work_status = true
```
giginet: I think you should return value here.

giginet (changes requested):
Are there tests in this situation?

This logic seems to be complicated. So, you should add some specs.

gcamp:
@cpk1 Are you planning to update your pull request? I would be available to fix it if you don't plan on doing it.

cpk1:
I'm trying to take a look this week. Sorry for the accidental push!

cpk1:
Oh I've noticed the change :). It looks like Pilot was affected quite a bit by todays update.

KrauseFx:
@cpk1 that's ok, we're in the works on updating spaceship 👍

ohayon:
I am going to close this because I think these changes have been superseded by the changes we needed to make when Apple rolled out changes to TestFlight a little while ago. thanks for this though @cpk1 !! 🚀

PR #7648, opened 2016-12-22
deliver: respect the '--platform' command-line setting

elliot-nelson:
### GOAL
Allow users to successfully upload AppleTV apps to iTunes Connect (using `--platform appletvos`).  See issue https://github.com/fastlane/fastlane/issues/7623.

### CHANGES
* Modify `Deliver::Runner` so that it correctly passes the value of `options[:platform]` (valid values are `ios`, `appletvos`, and `osx`) to the various backend tools.
* Modify `Deliver::DetectValues` to guess the platform correctly when possible (the platform check now comes first, before the other method calls which look at `options[:platform]`).
* Add some basic tests for `Deliver::Runner#upload_binary`.

### TESTS
- [x] Run `bundle exec rspec` from the root directory.
- [x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Issue #6480, opened 2016-10-10
Support for Universal Purchases in Pilot/Spaceship/Deliver/Fastlane


dombarnes:
### New Issue Checklist
- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

Currently, managing builds for apps with a Universal Purchase (iOS app and tvOS app using the same bundle ID) causes a number of conflicting results during use.
#### Setup

Parent folder has two sub-project folders, one for iOS app, one for tvOS app (named AppName and AppNameTVOS). Both share the same bundle ID, and so are both installable once one is purchased. Each project outputs its own independent IPA file, has its own TestFlight management, and has its own screenshots. Each project has its own fastlane directory, with screenshots, meta data. etc. 
#### Sample Fastfile

```
lane :beta do  
    ensure_git_branch(  
      branch: 'develop'  
      )  
    cocoapods    
    increment_build_number({  
      build_number:  latest_testflight_build_number({  
        initial_build_number:  get_build_number(  
          xcodeproj: 'TrilbyTVPlayerTVOS.xcodeproj'  
        ).to_i  
      }) + 1  
        })  
    gym(  
      scheme: "TrilbyTV Player",   
      workspace: "TrilbyTVPlayerTVOS.xcworkspace",   
      clean: true  
    )  
    # TODO: snapshot  
    sigh  
    pilot(  
      skip_submission: false,   
      distribute_external: false  
      )  
```

**Issues**
1. latest_testflight_build_number returns build number of iOS app, not tvOS when inside tvOS project folder
In my instance, the latest tvOS testflight build number is 10. It returns 124 based on my iOS app.
2. agvtool fails to update Info.plist's
I understand this isn't a fastlane issue, just noting it. Incorrect build number passed to agvtool though.

```
agvtool new-version -all 124
Setting version of project TrilbyTVPlayerTVOS to:
    124. 
Also setting CFBundleVersion key (assuming it exists)
Updating CFBundleVersion in Info.plist(s)...
$(SRCROOT)/TrilbyTVPlayerTVOS/TrilbyTVPlayer-Info.plist
Cannot find "$(SRCROOT)/TrilbyTVPlayerTVOS/TrilbyTVPlayer-Info.plist"
$(SRCROOT)/TrilbyTVPlayerTVOS/TrilbyTVPlayerDev-Info.plist
[...]

PR #8414, opened 2017-03-02
Improve tvos support in deliver

tmm1:
Previously, finding builds on a universal app would not work.

cc #6480 #7648

### DIFF
--- a/fastlane/lib/fastlane/actions/deliver.rb
+++ b/fastlane/lib/fastlane/actions/deliver.rb
@@ -44,7 +44,7 @@ module Fastlane
       end
 
       def self.is_supported?(platform)
-        [:ios, :mac].include?(platform)
+        [:ios, :appletvos, :mac].include?(platform)
       end
 
       def self.example_code
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -29,7 +29,7 @@ module Deliver
 
     def select_build(options)
       app = options[:app]
-      v = app.edit_version
+      v = app.edit_version(platform: options[:platform])
 
       if options[:build_number] and options[:build_number] != "latest"
         UI.message("Selecting existing build-number: #{options[:build_number]}")
@@ -55,7 +55,7 @@ module Deliver
       start = Time.now
 
       loop do
-        build = find_build(app.latest_version.candidate_builds)
+        build = find_build(app.latest_version(platform: options[:platform]).candidate_builds)
         return build if build.processing == false
 
         UI.message("Waiting iTunes Connect processing for build #{build.train_version} (#{build.build_version})... this might take a while...")


ohayon:
Thanks for this PR @tmm1! Has this been tested for all of the supported platforms? 🐙

tmm1:
I only tested uploading a tvos app, but I have no reason to believe it wouldn't work everywhere.

lacostej:
@tmm1 I see other places in deliver where we don't pass the `platform` option.

E..g `verify_available_languages!`, `UploadScreenshots#upload` for `edit_version` and more for `latest_version`

Have you had a chance to test all those features? (I don't do tvos apps, so can't help there)

lacostej:
See:
```sh
$ git grep -n -E "(edit|latest)_version" deliver | grep -v ":platform"
deliver/lib/deliver/commands_generator.rb:150:          v = options[:app].latest_version
deliver/lib/deliver/download_screenshots.rb:13:      v = options[:app].latest_version
deliver/lib/deliver/html_generator.rb:41:      @languages ||= options[:app].latest_version.description.languages
deliver/lib/deliver/setup.rb:27:      v = options[:app].latest_version
deliver/lib/deliver/submit_for_review.rb:32:      v = app.edit_version
deliver/lib/deliver/submit_for_review.rb:58:        build = find_build(app.latest_version.candidate_builds)
deliver/lib/deliver/upload_metadata.rb:153:      v = options[:app].edit_version
deliver/lib/deliver/upload_screenshots.rb:37:        v = app.edit_version
deliver/spec/commands_generator_spec.rb:207:      expect(fake_app).to receive(:latest_version).and_return('1.0.0')
deliver/spec/setup_spec.rb:9:      # expect(app).to receive(:latest_version).and_return(version)
```

tmm1:
I've used those features in the past with a tvos-only app, but I imagine they would not work as expected with a universal app.

I actually don't have a universal app myself, so if someone else wants to take over this effort I would prefer that.

fastlane/lib/fastlane/actions/deliver.rb
```diff
@@ -44,7 +44,7 @@ def self.author
       end
 
       def self.is_supported?(platform)
-        [:ios, :mac].include?(platform)
+        [:ios, :appletvos, :mac].include?(platform)
```
giginet: Is this right? In other implementations use `:tvos`.

giginet (changes requested):

ohayon:
Hey @tmm1, just checking in to see if you had seen @giginet 's requested changes, also, have you tested this since the new changes that Apple rolled out to iTC? Would be good to know if it is still a working change.

ohayon:
Hey @tmm1 - Wanted to bump this PR again just to see if you had a chance to look at @giginet's comments? Since this is a pretty old PR we should try to get this in or closed in the near future 😄 🚀 🐙

PR #8415, opened 2017-03-02
Add missing header

KrauseFx:
Looks like we used `##` and `###` for the same level. Also it looked like the friendliness was part of the dependencies


hjanuschka (approved):
great change ❤️

Issue #6325, opened 2016-09-28
ERROR - Crashlytics Developer Tools error. java.io.FileNotFoundException: /tmp/changelog20160928-32826-b1gtf0 (No such file or directory)


spirosoik:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
##### Complete output when running fastlane, including the stack trace and command used

```
ERROR - Crashlytics Developer Tools error.
java.io.FileNotFoundException: /tmp/changelog20160928-32826-b1gtf0 (No such file or directory)
    at java.io.FileInputStream.open0(Native Method)
    at java.io.FileInputStream.open(FileInputStream.java:195)
    at java.io.FileInputStream.<init>(FileInputStream.java:138)
    at com.google.common.io.Files$FileByteSource.openStream(Files.java:126)
    at com.google.common.io.Files$FileByteSource.openStream(Files.java:116)
    at com.google.common.io.ByteSource$AsCharSource.openStream(ByteSource.java:435)
    at com.google.common.io.CharSource.read(CharSource.java:162)
    at com.google.common.io.Files.toString(Files.java:367)
    at com.crashlytics.tools.android.DeveloperTools.getPropertyValueOrValueFromPropertyPath(DeveloperTools.java:743)
    at com.crashlytics.tools.android.DeveloperTools.processProperties(DeveloperTools.java:662)
    at com.crashlytics.tools.android.DeveloperTools.processArgsInternal(DeveloperTools.java:348)
    at com.crashlytics.tools.android.DeveloperTools.main(DeveloperTools.java:273)
Exception in thread "main" com.crashlytics.tools.android.exception.PluginException: Crashlytics Developer Tools error.
    at com.crashlytics.tools.android.DeveloperTools.processArgsInternal(DeveloperTools.java:356)
    at com.crashlytics.tools.android.DeveloperTools.main(DeveloperTools.java:273)
Caused by: java.io.FileNotFoundException: /tmp/changelog20160928-32826-b1gtf0 (No such file or directory)
    at java.io.FileInputStream.open0(Native Method)
    at java.io.FileInputStream.open(FileInputStream.java:195)
    at java.io.FileInputStream.<init>(FileInputStream.java:138)
    at com.google.common.io.Files$FileByteSource.openStream(Files.java:126)
[...]

PR #8416, opened 2017-03-02
Enable to pass `notes_path` using the file path of parent directoryf

tanukiti1987:
### Checklist

### Description

When we publish to android app with`crashlytics` method, we can specify `notes_path`.
But this `notes_path` can only pass below.

- the file path of current directory
- the file path of UNDER current directory
- the absolute path

So if we'd like to pass the file path of parent directories, we have to do `cd ../to/parent_directory`  and call `crashlytics` method, like this. https://github.com/fastlane/fastlane/issues/6325#issuecomment-278862019

This PR will any `notes_path` convert to the absolute path with `File.expand_path`.
Then we can specify `notes_path` with any file path expression.

### DIFF
--- a/fastlane/lib/fastlane/helper/crashlytics_helper.rb
+++ b/fastlane/lib/fastlane/helper/crashlytics_helper.rb
@@ -51,7 +51,7 @@ module Fastlane
 
           # Optional
           command << "-betaDistributionEmails '#{params[:emails]}'" if params[:emails]
-          command << "-betaDistributionReleaseNotesFilePath '#{params[:notes_path]}'" if params[:notes_path]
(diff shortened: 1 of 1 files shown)


hjanuschka (approved):
thats a great catch! thank you very much 🚀  and congrats to your first PR 🎉

mfurtak:
Thanks for this improvement! :+1: Are there tests we can add to make sure that this stays working in the future?

tanukiti1987:
> Are there tests we can add to make sure that this stays working in the future?

I concerned for writing the tests before sending PR.  But CrashlyticsHelper does not have any test currently. So I'll add specs on next PR :)

Issue #8402, opened 2017-03-01
The format of app_icon


Xwoder:
### New Issue Checklist

- [!] Updated fastlane to the latest version
- [!] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

##### Complete output when running fastlane, including the stack trace and command used

```
$ fastlane init 
[20:32:46]: Detected iOS/Mac project in current directory...
[20:32:46]: This setup will help you get up and running in no time.
[20:32:46]: fastlane will check what tools you're already using and set up
[20:32:46]: the tool automatically for you. Have fun! 
[20:32:46]: Created new folder './fastlane'.
...emit...
[20:34:04]: Successfully finished setting up fastlane
```

After all these done. I can find a icon file named `app_icon.jpg` in `./fastlane/metadata`

Then I use `fastlane deliver` to upload ipa, but fastlane exit and output
`[!] './fastlane/metadata/app_icon.jpg' doesn't seem to be a png file`

This is strange. I didn't rename or convert the icon to JPG format.

So I have to convert the icon to PNG format myself.

Is this a bug? I think that the icon image which fastlane downloaded for me should be in PNG format.

Thanks.

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                       |
| ------------------- | ------------------------------------------- |
| OS                  | 10.11.6                                     |
| Ruby                | 2.4.0                                       |
| Bundler?            | false                                       |
| Git                 | git version 2.12.0                          |
| Installation Source | /usr/local/bin/fastlane                     |
| Host                | Mac OS X 10.11.6 (15G1217)                  |
| Ruby Lib Dir        | /usr/local/Cellar/ruby/2.4.0/lib            |
| OpenSSL Version     | OpenSSL 1.0.2k  26 Jan 2017                 |
| Is contained        | false                                       |
| Is homebrew         | false                                       |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version       | 8.2.1                                       |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | zh_CN.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

[...]

Issue #8401, opened 2017-03-01
fastlane deliver error


santoshshetty2612:
Hi,

I am all of a sudden getting the below error using deliver.

./metadata/app_icon.jpg' doesn't seem to be a png file

Looks like some check has changed on the Apple end.

As a workaround, I tried ignoring metadata but still it tries to upload metadata and then throws the error.

fastlane deliver --skip_metadata --skip_screenshots --force

mfurtak:
Do you have a `Deliverfile`? If so, can you show me its contents?

santoshshetty2612:
Hi I don't have much content in the Deliverfile apart from the below two lines.

app_identifier "myapp.company.com" # The bundle identifier of your app
username "build.ci@company.com" # your Apple ID user

A metadata folder is created which contains app_icon.jpg

The error is that the icon should be in png format and not jpg.

hjanuschka:
seems to be related to #8402  i did a PR here: https://github.com/fastlane/fastlane/pull/8418 that should enable/allow jpg files - could you give it a try?

santoshshetty2612:
Hi @hjanuschka, I updated fast lane but still facing the issue.

santoshshetty2612:
My fast lane version is 2.19.1 and deliver version 2.0.1

Error is same /metadata/app_icon.jpg' doesn't seem to be a png file

santoshshetty2612:
I can understand that your PR is not yet merged into master.

CedricSoubrie:
2.19.3 version of deliver. Still the same problem

PR #8418, opened 2017-03-02
[deliver] support JPG icons for metadata

hjanuschka:
itunes also supports JPG for app/watch icon in the store entry.

addresses: https://github.com/fastlane/fastlane/issues/8402, #8401

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -189,16 +189,18 @@ module Deliver
                                      optional: true,
                                      short_option: "-l",
                                      verify_block: proc do |value|
+                                       supported_picture_endings = [".png", ".jpg", ".jpeg"]
                                        UI.user_error!("Could not find png file at path '#{File.expand_path(value)}'") unless File.exist?(value)
-                                       UI.user_error!("'#{value}' doesn't seem to be a png file") unless value.end_with?(".png")
+                                       UI.user_error!("'#{value}' doesn't seem to be one of the supported files. supported: #{supported_picture_endings.join(',')}") unless supported_picture_endings.include?(File.extname(value))
                                      end),
         FastlaneCore::ConfigItem.new(key: :apple_watch_app_icon,
                                      description: "Metadata: The path to the Apple Watch app icon",
                                      optional: true,
                                      short_option: "-q",
                                      verify_block: proc do |value|
+                                       supported_picture_endings = [".png", ".jpg", ".jpeg"]
                                        UI.user_error!("Could not find png file at path '#{File.expand_path(value)}'") unless File.exist?(value)
-                                       UI.user_error!("'#{value}' doesn't seem to be a png file") unless value.end_with?(".png")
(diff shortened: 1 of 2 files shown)


deliver/lib/deliver/options.rb
```diff
@@ -189,16 +189,18 @@ def self.available_options
                                      optional: true,
                                      short_option: "-l",
                                      verify_block: proc do |value|
+                                       supported_picture_endings = [".png", ".jpg", ".jpeg"]
```
KrauseFx: We define the same array multiple times now?

hjanuschka: i thought it is better twice declard than having it globally in a class? - any other idea, or sample, where stuff like this should go?

KrauseFx: It could probably just be a variable inside the method, just like here https://github.com/fastlane/fastlane/pull/8418/files#diff-cec46a69540d57cbb9c8631d43a7a89bR7

giginet: I think `Deliver::UploadAssets` should have them as constants.

giginet (changes requested):

deliver/lib/deliver/options.rb
```diff
@@ -189,16 +189,18 @@ def self.available_options
...
                                      end),
         FastlaneCore::ConfigItem.new(key: :apple_watch_app_icon,
                                      description: "Metadata: The path to the Apple Watch app icon",
                                      optional: true,
                                      short_option: "-q",
                                      verify_block: proc do |value|
+                                       supported_picture_endings = [".png", ".jpg", ".jpeg"]
                                        UI.user_error!("Could not find png file at path '#{File.expand_path(value)}'") unless File.exist?(value)
-                                       UI.user_error!("'#{value}' doesn't seem to be a png file") unless value.end_with?(".png")
+                                       UI.user_error!("'#{value}' doesn't seem to be a png file") unless supported_picture_endings.include?(File.extname(value))
```
giginet: This checking should be case insensitive.

`supported_picture_endings.include?(File.extname(value).downcase)`

hjanuschka:
@KrauseFx @giginet thx for review, feedback addressed.

deliver/lib/deliver/upload_assets.rb
```diff
@@ -1,5 +1,6 @@
 module Deliver
   class UploadAssets
+    SUPPORTED_ICON_EXTENSIONS = [".png", ".jpg"]
```
giginet: Is this OK without `jpeg` extension?

hjanuschka:
@giginet eagle-eye - thx i missed the `jpeg` when moving to the constant

giginet (approved):
LGTM

santoshshetty2612:
Thanks @giginet for your work.
@KrauseFx Can we please create a release for the same so that this deliver issue is fixed.

PR #12, opened 2015-01-22
Upload dSYM file with build

vytis:
At the moment crashes cannot be symbolicated properly because symbols are not uploaded together with the `.ipa`.

This fix adds support for uploading dSYM.zip if it is found in the current directory (this is provided if Shenzen is used to create `.ipa`) or you can optionally supply a parameter with dSYM path.

PR #8440, opened 2017-03-04
[cert] Add generate_p12 & p12_password options to export p12 file

0x5e:
### Checklist

### Description
Add `generate_p12` and `p12_password` options for cert.
The code was copied from `pem/lib/pem/manager.rb`.
Changed private key extension  to `pkey` (not `p12` actually).

### Motivation and Context
Solved issue #1805


cert/lib/cert/options.rb
```diff
@@ -13,6 +13,16 @@ def self.available_options
...
+        FastlaneCore::ConfigItem.new(key: :generate_p12,
+                                     env_name: "CERT_GENERATE_P12_FILE",
+                                     description: "Generate a p12 file additionally to a PEM file",
+                                     is_string: false,
+                                     default_value: true),
+        FastlaneCore::ConfigItem.new(key: :p12_password,
+                                     env_name: "CERT_P12_PASSWORD",
+                                     sensitive: true,
+                                     description: "The password that is used for your p12 file",
+                                     default_value: ""),
```
hjanuschka: since `:generate_p12` is `true` by default wouldn't that fail/corrupt the p12 file because of `blank password` ?

hjanuschka (changes requested):
left a comment - i like the idea, but to prevent regressions could you please add some tests for the added options.

cert/lib/cert/options.rb
0x5e: According to #2814, `p12` file with blank password is valid, although Apple's Keychain Access refuse to import it.
`openssl pkcs12 -info -in /path/to/my.p12` still works.

0x5e:
OK, I will try :)

cert/lib/cert/options.rb
hjanuschka: we just have to be carefull - to not break existing setups, as one of fastlane goals is to have sane defaults!  so once you have specs for this PR - i am going to give it a testrun on a couple of projects.

cert/lib/cert/runner.rb
```diff
@@ -160,11 +160,18 @@ def create_certificate
       request_path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.certSigningRequest"))
       File.write(request_path, csr.to_pem)
 
-      private_key_path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.p12"))
+      private_key_path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.pkey"))
```
KrauseFx: We can't do that, as this would be a breaking change for setups that assume the `.p12` file extension. Or does https://github.com/fastlane/fastlane/pull/8440/files#diff-b4b9b8ba735c52a8fe2212a8f0981c4dR171 write the file on the same output location?

KrauseFx:
Before properly reviewing the PR, could you provide some more context on why this change is useful for _fastlane_ users? 👍

cert/lib/cert/runner.rb
0x5e: Yes, https://github.com/fastlane/fastlane/pull/8440/files#diff-b4b9b8ba735c52a8fe2212a8f0981c4dR171 white the same location, with private_key and certificate bundled in.

0x5e:
My motivation is to manage clients' certificate & private key & mobileprovision files on my server, not only saved in a single local machine. We have about hundred clients, and already have a website to manage this. 
Then I found cert export `p12` file which was actually not a PKCS #12 file. Just a private key.
Sorry for my poor english :)

0x5e:
Hi, I added some tests, please let me know if more test is required.

0x5e:
I signed it!

mfurtak:
Hi @0x5e, @hjanuschka  - do we know what is needed to make a determination on this?

hjanuschka (approved):

hjanuschka:
@mfurtak  lgtm - @KrauseFx still on comment from you left, i think.

hjanuschka (changes requested):
ahh - damn - too fast approved, i am still not 100% clear about the default `true`  of `:generate_p12` - and default `""`  of `:p12_password` as this leads to p12 files without passwords, which are not importable (as you stated) - can you clarify that? @0x5e 


i think the better approach would be to default `:generate_p12` to `false` (this way it should not impact any existing user.)

also if `""`  is not a valid password, how about `UI.input` ?

0x5e:
Hi @hjanuschka , macOS keychain requires to input a non-empty password when import p12 file, so if a p12 without password, it can't be import.

I have no idea about `:generate_p12` default value, just because `PEM_GENERATE_P12_FILE` default is `true`, so I think maybe it's good to have the same default value.
I just need to export a p12 file, each default value are acceptable for me :)

hjanuschka (approved):

hjanuschka:
looking good from my side, @fastlane/external-contributors can someone help out here for a quick double-check!?

cert/lib/cert/runner.rb
```diff
@@ -160,11 +160,18 @@ def create_certificate
...
       File.write(private_key_path, pkey)
 
       cert_path = store_certificate(certificate)
 
+      if Cert.config[:generate_p12]
+        p12_cert_path = File.join(Cert.config[:output_path], "#{certificate.id}.p12")
+        p12 = OpenSSL::PKCS12.create(Cert.config[:p12_password], type_name, pkey, certificate.download)
+        File.write(p12_cert_path, p12.to_der)
+        UI.message("p12 certificate: ".green + Pathname.new(p12_cert_path).realpath.to_s)
```
milch: Usually we use `File.expand_path` instead of `Pathname.new(...).realpath.to_s`. Unless that does something I don't know about?

milch (changes requested):

milch:
@hjanuschka I looked over it, just saw a small inconsistency with how we usually do things.

cert/lib/cert/runner.rb
milch: Also, instead of `UI.message(<string>.green ...` maybe `UI.success(<string>)` would be more appropriate?

0x5e: I don't know... copied from `\pem\lib\pem\manager.rb`. Still new learner in ruby, I'll modify it later

0x5e: okay

0x5e:
@milch Updated

milch (approved):

milch:
@0x5e Alright, thanks!

@hjanuschka I'd say this is good now 👍

hjanuschka:
@0x5e thx for your contribution 🚀  
@milch awesome THX - for the help here.!

KrauseFx:
Unfortunately this broke _match_ (https://github.com/fastlane/fastlane/issues/8810) 

I added a comment that it would break things via https://github.com/fastlane/fastlane/pull/8440#discussion_r104380017, and the PR was still merged without addressing my comment. 

_match_ assumes the `p12` extension, which now isn't available by default, I'm looking into this, but it's probably something we need to revert or fix.

cert/lib/cert/options.rb
KrauseFx: Yeah...

cert/lib/cert/options.rb
```diff
@@ -13,6 +13,16 @@ def self.available_options
                                      description: "Create a development certificate instead of a distribution one",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :generate_p12,
+                                     env_name: "CERT_GENERATE_P12_FILE",
+                                     description: "Generate a p12 file additionally to a PEM file",
+                                     is_string: false,
+                                     default_value: false),
```
KrauseFx: Changing the default to `true` should make it work, right/

PR #8447, opened 2017-03-06
fix right rounded corner - and beautify svg

hjanuschka:
https://github.com/fastlane/fastlane/pull/8447/files?short_path=acaf668#diff-acaf6681c66def9713b4ec25b92131a3


mfurtak (approved):
:100:

Issue #8432, opened 2017-03-03
Summary not reporting all devices


zarghol:
### Issue Description
Hi ! 
I ran fastlane snapshot with 3 devices et 5 languages, all worked, but the summary only contains 2 devices...

I will run again the command to check if the bug persist, but even if it works, there's a random bug :-)

Thanks for your help !
##### Output

[Output fastlane.txt](https://github.com/fastlane/fastlane/files/816815/Output.fastlane.txt)
[Snapfile.txt](https://github.com/fastlane/fastlane/files/816819/Snapfile.txt)
[screenshots.html.txt](https://github.com/fastlane/fastlane/files/816827/screenshots.html.txt)

### Environment
<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                                          |
| ------------------- | -------------------------------------------------------------- |
| OS                  | 10.12.2                                                        |
| Ruby                | 2.0.0                                                          |
| Bundler?            | false                                                          |
| Git                 | git version 2.10.1 (Apple Git-78)                              |
| Installation Source | /usr/local/bin/fastlane                                        |
| Host                | Mac OS X 10.12.2 (16C67)                                       |
| Ruby Lib Dir        | /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib |
| OpenSSL Version     | OpenSSL 0.9.8zc 19 Mar 2015                                    |
| Is contained        | false                                                          |
| Is homebrew         | false                                                          |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version       | 8.2.1                                                          |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | fr_FR.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

**No Fastfile found**

**No Appfile found**

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
| fastlane | 2.19.1  | ✅ Up-To-Date  |

### Loaded fastlane plugins:

**No plugins Loaded**

<details><summary><b>Loaded gems</b></summary>

| Gem                       | Version      |
| ------------------------- | ------------ |
| slack-notifier            | 1.5.1        |
| rouge                     | 1.11.1       |
| xcpretty                  | 0.2.4        |
| terminal-notifier         | 1.7.1        |
| unicode-display_width     | 1.1.3        |
[...]

PR #8451, opened 2017-03-06
Test ReportsGenerator available_devices

mfurtak:
### Checklist

### Description

Since the `Collector` creates file names where the device names are stripped of any spaces we should give ourselves a safety check on `available_devices` that ensures that we don't introduce keys with spaces in the future.

### Motivation and Context

#8432 discovered a mistake with the device name key for the iPhone SE in `ReportsGenerator`


ohayon (approved):
:100:

Issue #8456, opened 2017-03-07
undefined local variable or method `cert' for #<Sigh::Runner:...>


kas-kad:
### New Issue Checklist

- [yes] Updated fastlane to the latest version
- [no] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
$ fastlane sigh --development

+-------------------------------------+-------------------------------------+
|                          Summary for sigh 2.19.2                          |
+-------------------------------------+-------------------------------------+
| development                         | true                                |
| adhoc                               | false                               |
| skip_install                        | false                               |
| force                               | false                               |
| app_identifier                      | com.myapp 							|
| team_id                             | T6U248P7YM                          |
| ignore_profiles_with_different_name | false                               |
| skip_fetch_profiles                 | false                               |
| skip_certificate_verification       | false                               |
| platform                            | ios                                 |
+-------------------------------------+-------------------------------------+

[11:51:44]: To not be asked about this value, you can specify it using 'username'
Your Apple ID Username: mymail@mymail.com
[11:51:51]: Starting login with user 'mymail@mymail.com'
[11:51:54]: Successfully logged in
[11:51:54]: Fetching profiles...
[11:51:55]: Verifying certificates...

Looking for related GitHub issues on fastlane/fastlane...

➡️  undefined local variable or method `cert' for #<Sigh::Runner:...>
   https://github.com/fastlane/fastlane/issues/8456 [open] 0 💬
   17 minutes ago

➡️  undefined method `[]=' for nil:NilClass for sigh manage
   https://github.com/fastlane/fastlane/issues/6751 [closed] 15 💬
   4 weeks ago

➡️  Undefined method `fetch` for nil:NilClass (NoMethodError) again?
   https://github.com/fastlane/fastlane/issues/7307 [closed] 4 💬
   3 weeks ago

and 34 more at: https://github.com/fastlane/fastlane/search?q=undefined%20%60cert'%20for%20Sigh::Runner&type=Issues&utf8=✓

🔗  You can ⌘ + double-click on links to open them directly in your browser.
[...]

PR #8457, opened 2017-03-07
[sigh] Fix Regression undefined variable

hjanuschka:
should address: https://github.com/fastlane/fastlane/issues/8456

also gets rid of a few uggly loop vars.


mfurtak:
Is there a test we could add that would have helped us detect this regression sooner?

kas-kad (approved):
Tried this version of runner.rb on my machine, the original issue(https://github.com/fastlane/fastlane/issues/8456) didn't appear.

mfurtak (approved):
Approved for hotfix - let's get tests in place soon!

PR #8458, opened 2017-03-07
Change "download zip" URL to point to latest version in README

BalestraPatrick:
### Checklist

### Description
Change "download zip" URL to point to latest version in README. I mistakenly installed an old version by download the zip file using the URL. I think the zip file hosted on the server is not being kept up to date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ xcode-select --install
 </tr>
 <tr> 
 <td width="33%"><code>brew cask install fastlane</code></td>
-<td width="33%"><a href="https://download.fastlane.tools/fastlane.zip">Download the zip file</a>. Then double click on the <code>install</code> script (or run it in a terminal window).</td>
+<td width="33%"><a href="https://download.fastlane.tools/">Download the zip file</a>. Then double click on the <code>install</code> script (or run it in a terminal window).</td>
 <td width="33%"><code>sudo gem install fastlane -NV</code></td>
 </tr>
 </table>


hjanuschka (approved):
great catch 🚀

hjanuschka:
@krausefx should we remove the old zip file from s3?

mfurtak:
I believe the files downloaded between those two URL are the same. Am I mistaken?

BalestraPatrick:
@mfurtak I tried twice to download from the old URL on two machines and both times I ended up with an old version. I was thinking to point to the release/latest page here in GitHub but after removing the direct link to the fastlane website, it seems to download the latest one.

KrauseFx:
They should be the same, but maybe @ohayon knows more. If we merge this change, there are about 5 other locations where we link to this, so we'd need to update the URL there too.

hjanuschka:
@KrauseFx  i am going to merge this, and do a follow up pr with hunted zip urls 👍

mfurtak:
I think that what @balestrapatrick is talking about is how the version of fastlane that comes in the package is not always up-to-date. This is a known issue, and is not related to the URL

hjanuschka:
@mfurtak ohh well thats sad, should we revert this one then?

BalestraPatrick:
@mfurtak Actually I downloaded the zip file and it installed version 2.5.0. Could you please try to check if it's really downloading the latest version with the old URL?

ohayon:
I believe @mfurtak is correct here. Those URLs should download the same thing. There is some weirdness around versioning such that the package itself is not tied to a version of _fastlane_. @BalestraPatrick - this is why it might appear as though you are downloading an old version, but upon installation or running `fastlane update_fastlane` you can update _fastlane_ to the most recent version of the gem.

hjanuschka:
@ohayon  but using the link without `fastlane.zip` is also correct? (so no revert?)

ohayon:
Yup @hjanuschka ! No need to revert 😄

mfurtak:
Either link is correct (identical), but Felix had mentioned consistency for all the places we link to it

hjanuschka:
ok - i am fine with a revert however i had immediately after felix's comment prepared a PR here: https://github.com/fastlane/docs/pull/298 and https://github.com/fastlane/fastlane/pull/8459 that should make them consistent

mfurtak:
OK that's fine too

Xwoder:
After I download the zip file and unzip it, I found that the content of file `.../fastlane/fastlane_lib/VERSION` still showed `1.7`.

The `Download the zip file` links to `https://kits-crashlytics-com.s3.amazonaws.com/fastlane/standalone/fastlane.zip`.

mfurtak:
Hi @Xwoder 👋 

The version number noted in the fastlane installer package represents the version of the installer package. The version of _fastlane_ that gets installed will be different. Once you install _fastlane_, it will keep itself up-to-date, and you can run `fastlane -v` to verify your version. 👍 

I'm going to lock this issue for now to cut down on notifications for everyone.

PR #8462, opened 2017-03-07
Check for ENV vars that interfere with local testing

mfurtak:
Help us discover local state problems that interfere with releases

### Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -228,18 +228,6 @@ end
 
 desc "Verifies all tests pass and the current state of the repo is valid"
 private_lane :validate_repo do
-  # Ensure that no ENV vars which interefere with testing are set
-  #
-  set_auth_vars = [
-    'FASTLANE_SESSION',
-    'FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD',
-    'FASTLANE_PASSWORD'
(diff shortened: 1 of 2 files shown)


lacostej:
Could we unset them instead (and maybe warn the user?)

fastlane/Fastfile
```diff
@@ -228,6 +228,18 @@ end
 
 desc "Verifies all tests pass and the current state of the repo is valid"
 private_lane :validate_repo do
+  # Ensure that no ENV vars which interefere with testing are set
```
KrauseFx: Is the `Fastfile` the right spot to put this code? Shouldn't this be part of the test suite instead?

mfurtak:
@lacostej Either is probably fine, but I think my preference is make our mistakes noisy, rather than ignorable.

fastlane/Fastfile
mfurtak: I suppose? I guess I look at my normal testing process to be more fluid. Maybe I'm setting ENV variables to explore things... In comparison, the release process should be pretty locked down.

Does that make sense? If not, do you have a sense of where you would put it?

KrauseFx: I think `spec_helper` is the perfect place for this 👍

KrauseFx (approved):
A great update 👍

PR #8484, opened 2017-03-09
Revise FastlaneCore::FastlaneFolder.path

nafu:
### Checklist

### Description

Remove the code below from FastlaneCore::FastlaneFolder.path.
```ruby
value = nil if Helper.is_test? # this is required, as the tests would use the ./fastlane folder otherwise
```

### Motivation and Context

`value = nil if Helper.is_test?` should not be there, because it's only for testing.
We can handle this in rspec side.

I tried to make this change as simple as possible because the main target of this change is #8380.
I also plan to remove other `Helper.is_test?` after this get merged.

### DIFF
--- a/spec_helper.rb
+++ b/spec_helper.rb
@@ -30,10 +30,6 @@ end
 my_main = self
 RSpec.configure do |config|
   config.before(:each) do |current_test|
-    if current_test.id =~ %r{(fastlane\/spec\/(?!lane_manager))|(scan\/spec\/test_command_generator)}
-      allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
-    end
-
     tool_name = current_test.id.match(%r{\.\/(\w+)\/})[1]
     method_name = "before_each_#{tool_name}".to_sym
     begin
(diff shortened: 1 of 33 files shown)


spec_helper.rb
```diff
@@ -30,6 +30,10 @@ module SpecHelper
 my_main = self
 RSpec.configure do |config|
   config.before(:each) do |current_test|
+    if current_test.id =~ %r{(fastlane\/spec\/(?!lane_manager))|(scan\/spec\/test_command_generator)}
+      allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
+    end
```
KrauseFx: I agree, this is not ideal.

nafu: @mfurtak @KrauseFx Thank you for your feedback.
I'll put it to each test that needs the mock 👍

nafu:
@mfurtak @KrauseFx 
I updated this PR! Could you review?

mfurtak (approved):
Looks good! 👍

nafu:
@mfurtak Thanks 🚀

Issue #8384, opened 2017-02-28
Add carthage --cache-builds option


alphatroya:
Carthage released version 0.20 with builds cache [flag](https://github.com/Carthage/Carthage/releases/tag/0.20.0), I think it should be added to the fastlane action

joshdholtz:
@alphatroya Thanks for the suggestion! I will add this for you :blush:

Rekubot:
👍

alphatroya:
I think it implemented, thanks

PR #8504, opened 2017-03-13
Adding Carthage cache-builds flag

LukeDurrant:
### Checklist

### Description

Enabled cache-builds flag

By default Carthage will rebuild a dependency regardless of whether it's the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can. See https://github.com/Carthage/Carthage/blob/master/Documentation/VersionFile.md

### Motivation and Context
Carthage has a new cache-builds flag

### DIFF
--- a/fastlane/lib/fastlane/actions/carthage.rb
+++ b/fastlane/lib/fastlane/actions/carthage.rb
@@ -1,6 +1,7 @@
 module Fastlane
   module Actions
     class CarthageAction < Action
+      # rubocop:disable Metrics/PerceivedComplexity
       def self.run(params)
         validate(params)
 
@@ -26,9 +27,11 @@ module Fastlane
         cmd << "--derived-data #{params[:derived_data].shellescape}" if params[:derived_data]
         cmd << "--toolchain #{params[:toolchain]}" if params[:toolchain]
         cmd << "--project-directory #{params[:project_directory]}" if params[:project_directory]
+        cmd << "--cache-builds" if params[:cache_builds] == true
 
         Actions.sh(cmd.join(' '))
       end
+      # rubocop:enable Metrics/PerceivedComplexity
 
       def self.validate(params)
         command_name = params[:command]
@@ -127,6 +130,11 @@ module Fastlane
                                            UI.user_error!("Please pass a valid platform. Use one of the following: #{available_platforms.join(', ')}") unless available_platforms.map(&:downcase).include?(platform.downcase)
                                          end
                                        end),
+          FastlaneCore::ConfigItem.new(key: :cache_builds,
+                                       env_name: "FL_CARTHAGE_CACHE_BUILDS",
+                                       description: "By default Carthage will rebuild a dependency regardless of whether it\'s the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can",
+                                       is_string: false,
+                                       default_value: false),
           FastlaneCore::ConfigItem.new(key: :frameworks,
                                        description: "Framework name or names to archive, could be applied only along with the archive command",
                                        default_value: [],
@@ -175,6 +183,7 @@ module Fastlane
             verbose: false,                                 # Print xcodebuild output inline
             platform: "all",                                # Define which platform to build for (one of ‘all’, ‘Mac’, ‘iOS’, ‘watchOS’, ‘tvOS‘, or comma-separated values of the formers except for ‘all’)
             configuration: "Release",                       # Build configuration to use when building
+            cache_builds: true,                             # By default Carthage will rebuild a dependency regardless of whether it\'s the same resolved version as before.
             toolchain: "com.apple.dt.toolchain.Swift_2_3"   # Specify the xcodebuild toolchain
           )'
         ]
--- a/fastlane/lib/fastlane/actions/carthage.rb
+++ b/fastlane/lib/fastlane/actions/carthage.rb
@@ -27,7 +27,7 @@ module Fastlane
         cmd << "--derived-data #{params[:derived_data].shellescape}" if params[:derived_data]
         cmd << "--toolchain #{params[:toolchain]}" if params[:toolchain]
         cmd << "--project-directory #{params[:project_directory]}" if params[:project_directory]
-        cmd << "--cache-builds" if params[:cache_builds] == true
+        cmd << "--cache-builds" if params[:cache_builds]
 
         Actions.sh(cmd.join(' '))
       end
@@ -132,7 +132,7 @@ module Fastlane
                                        end),
           FastlaneCore::ConfigItem.new(key: :cache_builds,
                                        env_name: "FL_CARTHAGE_CACHE_BUILDS",
-                                       description: "By default Carthage will rebuild a dependency regardless of whether it\'s the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can",
+                                       description: "By default Carthage will rebuild a dependency regardless of whether it's the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can",
                                        is_string: false,
                                        default_value: false),
           FastlaneCore::ConfigItem.new(key: :frameworks,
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -127,6 +128,14 @@ def self.available_options
                                            UI.user_error!("Please pass a valid platform. Use one of the following: #{available_platforms.join(', ')}") unless available_platforms.map(&:downcase).include?(platform.downcase)
                                          end
                                        end),
+          FastlaneCore::ConfigItem.new(key: :cache_builds,
+                                       env_name: "FL_CARTHAGE_CACHE_BUILDS",
+                                       description: "By default Carthage will rebuild a dependency regardless of whether it's the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can. See https://github.com/Carthage/Carthage/blob/master/Documentation/VersionFile.md",
```
KrauseFx: Let's try to avoid having URLs in the description and add them to the `self.details` instead

fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -127,6 +128,14 @@ def self.available_options
...
+                                       is_string: false,
+                                       optional: true,
+                                       verify_block: proc do |value|
+                                         UI.user_error!("Please pass a valid value for cache_builds. Use one of the following: true, false") unless value.kind_of?(TrueClass) || value.kind_of?(FalseClass)
```
KrauseFx: I think we already have some kind of solution for boolean verification in _fastlane_

mcfedr:
Fixes https://github.com/fastlane/fastlane/issues/8384

fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -127,6 +130,11 @@ def self.available_options
                                            UI.user_error!("Please pass a valid platform. Use one of the following: #{available_platforms.join(', ')}") unless available_platforms.map(&:downcase).include?(platform.downcase)
                                          end
                                        end),
+          FastlaneCore::ConfigItem.new(key: :cache_builds,
+                                       env_name: "FL_CARTHAGE_CACHE_BUILDS",
+                                       description: "By default Carthage will rebuild a dependency regardless of whether it\'s the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can",
+                                       is_string: false,
+                                       optional: true),
```
giginet: You should set `default` is `false` rather than `optional` is `true`.

giginet: You should set `default` is `false` rather than `optional` is `true`.

giginet (changes requested):
You should avoid using `optional` parameter.

fastlane/lib/fastlane/actions/carthage.rb
LukeDurrant: Why is that?
I simply followed the rest of the params haven't seen anywhere where it does this?

hjanuschka: @giginet i think `optional: true` in combination with `is_string: false`  has the same effect, and is used accross the whole codebase. @KrauseFx  am i wrong?

KrauseFx: I think it's the same, yes 👍

LukeDurrant:
@giginet?

LukeDurrant:
I signed it!

giginet (approved):

fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -26,9 +27,11 @@ def self.run(params)
         cmd << "--derived-data #{params[:derived_data].shellescape}" if params[:derived_data]
         cmd << "--toolchain #{params[:toolchain]}" if params[:toolchain]
         cmd << "--project-directory #{params[:project_directory]}" if params[:project_directory]
+        cmd << "--cache-builds" if params[:cache_builds] == true
```
KrauseFx: Please remove the `==true` code

fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -127,6 +130,11 @@ def self.available_options
                                            UI.user_error!("Please pass a valid platform. Use one of the following: #{available_platforms.join(', ')}") unless available_platforms.map(&:downcase).include?(platform.downcase)
                                          end
                                        end),
+          FastlaneCore::ConfigItem.new(key: :cache_builds,
+                                       env_name: "FL_CARTHAGE_CACHE_BUILDS",
+                                       description: "By default Carthage will rebuild a dependency regardless of whether it\'s the same resolved version as before. Passing the --cache-builds will cause carthage to avoid rebuilding a dependency if it can",
```
KrauseFx: Why not just `'` instead of `\'`

fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -175,6 +183,7 @@ def self.example_code
             verbose: false,                                 # Print xcodebuild output inline
             platform: "all",                                # Define which platform to build for (one of ‘all’, ‘Mac’, ‘iOS’, ‘watchOS’, ‘tvOS‘, or comma-separated values of the formers except for ‘all’)
             configuration: "Release",                       # Build configuration to use when building
+            cache_builds: true,                             # By default Carthage will rebuild a dependency regardless of whether it\'s the same resolved version as before.
```
KrauseFx: Why the `\'`

fastlane/spec/actions_specs/carthage_spec.rb
```diff
@@ -423,6 +443,20 @@
...
+          Fastlane::FastFile.new.parse("lane :test do
+            carthage(
+              use_ssh: true,
+              use_submodules: true,
+              use_binaries: false,
+              cache_builds: true,
+              platform: 'iOS'
+            )
+          end").runner.execute(:test)
+        end.not_to raise_error
```
KrauseFx: `not_to raise_error` is very generic, can we actually test something here maybe?

KrauseFx (dismissed):
Hey @LukeDurrant, thanks for the PR, this is a great change. I added some notes before it's ready

fastlane/spec/actions_specs/carthage_spec.rb
LukeDurrant: Suggestion?

fastlane/lib/fastlane/actions/carthage.rb
LukeDurrant: Whys that?

LukeDurrant:
Accidentally closed reopening!

fastlane/lib/fastlane/actions/carthage.rb
LukeDurrant: Because it needs escaping?

fastlane/lib/fastlane/actions/carthage.rb
LukeDurrant: Doesn't need escaping here

adomanico:
Love this, thanks for setting this up.

Going to speed up our daily builds a lot!

mcfedr:
Yea, been using the for about a week, build times have been halved, its great!

LukeDurrant:
Yeah build servers across the world will have a .... `sigh` of relief

netbe:
Can't wait to use it too. Just curious how you handle versions (let's say someone using new fastlane version with this `cache-builds` flag but with older carthage version)

mcfedr:
In our case, fastlane crashes, and someone asks in the the group chat why its broken, and I say, update carthage :)

fastlane/spec/actions_specs/carthage_spec.rb
hjanuschka: i think the prev. test also covers this. so i am in favor of removing this particular test case.
as i would like to see this PR merged, @KrauseFx  whats your point?

mcfedr:
Is there still something holding this up?

fastlane/spec/actions_specs/carthage_spec.rb
mfurtak: This is consistent with the tests around it, so we should ship it, IMO. If folks feel strongly, we can always take it back out. 👍

mfurtak (approved):

mfurtak:
Thanks for the contribution, @LukeDurrant, and for your testing @mcfedr 🚀

PR #8509, opened 2017-03-13
Fix error installing the AppleWWDRCA certificate on multiuser systems

mcfedr:
### Checklist

### Description

Download AppleWWDRCA.cer as a Tempfile rather than to a specific filename.
Using Tempfile also means that ruby automatically removes the file from the system.

### Motivation and Context

It can happen that the file /tmp/AppleWWDRCA.cer already exists, this
causes the install to fail because the file cannot be created as the
current user. Using a random temp file prevents this issue.

This happens on my CI server, which is a single mac with multiple users as
separate runners, the first to create the file `/tmp/AppleWWDRCA.cer` works,
all the other runners fail to run `match`.

### DIFF
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -1,3 +1,5 @@
+require 'tempfile'
+
 module FastlaneCore
   # This class checks if a specific certificate is installed on the current mac
   class CertChecker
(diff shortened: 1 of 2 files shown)


ohayon:
Thanks for the PR @mcfedr, this looks like a good change. Currently these changes are causing the tests to break. If you could make sure that the tests are passing we will be able to get this merged in! 🐙

mcfedr:
@ohayon yea, ill look at that, rspec didnt want to run locally so i pushed and hoped :)

giginet (approved):
LGTM
Please fix CI.

mcfedr:
@giginet All greened up!

hjanuschka (approved):

PR #8511, opened 2017-03-13
Find and delete keychains with -db extension

mcfedr:
### Checklist

### Description
When deleting keychains, search for files with `-db` at the end of the file name.

### Motivation and Context
Since macOS Sierra keychain creates files with an extra -db extension, which delete_keychain action would fail to find.

### DIFF
--- a/fastlane/lib/fastlane/actions/delete_keychain.rb
+++ b/fastlane/lib/fastlane/actions/delete_keychain.rb
@@ -6,19 +6,16 @@ module Fastlane
       def self.run(params)
         original = Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN]
 
-        search_paths = []
-        search_paths << File.expand_path(params[:name]) if params[:name]
-        search_paths << File.expand_path(File.join("~", "Library", "Keychains", params[:name])) if params[:name]
-        search_paths << File.expand_path(params[:keychain_path]) if params[:keychain_path]
-
-        if search_paths.empty?
+        if params[:keychain_path]
+          if File.exist?(params[:keychain_path])
+            keychain_path = params[:keychain_path]
+          else
+            UI.user_error!("Unable to find the specified keychain.")
+          end
+        elsif params[:name]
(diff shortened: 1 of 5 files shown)


KrauseFx:
Thanks for the pull request @mcfedr 👍 There have been quite some changes around the keychain in the past, I'd like @milch's approval on this before merging.

fastlane/lib/fastlane/actions/delete_keychain.rb
```diff
@@ -7,9 +7,16 @@ def self.run(params)
         original = Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN]
 
         search_paths = []
-        search_paths << File.expand_path(params[:name]) if params[:name]
-        search_paths << File.expand_path(File.join("~", "Library", "Keychains", params[:name])) if params[:name]
-        search_paths << File.expand_path(params[:keychain_path]) if params[:keychain_path]
+        if params[:name]
+          search_paths << File.expand_path(params[:name])
```
milch: We recently introduced a new [helper](https://github.com/fastlane/fastlane/blob/5eed4da1559136ff4c490029ebf0b794c826ded2/fastlane_core/lib/fastlane_core/helper.rb#L180) that handles all of this. It searches all the possible locatiosn that you are adding to the search_paths (and more) so you should be able to use that helper instead of this code.

milch (dismissed):
Hey @mcfedr, thanks for your PR! 

We've moved all keychain searching code to a central location, so please use that method instead of manually building the search list. It seems that we forgot `delete_keychains` when we did that refactoring, so thanks for taking care of it!

mcfedr:
Hi @milch I have changed things around to use the helper method now, that seems to make a lot more sense!

I noticed that the import_certificate action also had its own search, so and changed that to use the helper.

For a moment I thought the helper wasnt working completely correctly, so there are a could of specs added for it, I assume that isnt a problem

mfurtak:
This appears to fix #8543, and LGTM. Is this ready to go in your eyes, @milch?

mfurtak (approved):
Thanks for this fix!

PR #8512, opened 2017-03-13
Add missing `ENV.delete` statement after generating test data

KrauseFx:
This would cause the `USE_ENHANCE_TEST_DATA` environment variable to be set while verifying the repo is valid, however it's still being active *after* pushing the fastlane release, when generating the final docs, causing the order of the actions to be wrong, and the plugins to be hidden


KrauseFx:
cc @joshdholtz

joshdholtz (approved):
Legit 👌

PR #8517, opened 2017-03-14
Error when description is less than 10 chars

jetersen:
### Checklist

### Description
Give Warning when Beta Description is less than 10 characters or greater than 4000

### Motivation and Context
Bumper to get denied after the upload is successful.
If I could be pointed in the right direction for a test case of pilot update build info

```
[16:32:36]: Successfully finished processing the build
[16:32:36]: You can now tweet: 
[16:32:36]: iTunes Connect #iosprocessingtime 10 minutes
*snip lane context*
[16:32:37]: [English]: Beta App Description must be at least 10 characters long. Beta App Description must be at least 10 characters long.
```

### DIFF
--- a/pilot/lib/pilot/options.rb
+++ b/pilot/lib/pilot/options.rb
@@ -46,7 +46,11 @@ module Pilot
                                      short_option: "-d",
                                      optional: true,
                                      env_name: "PILOT_BETA_APP_DESCRIPTION",
-                                     description: "Provide the beta app description when uploading a new build"),
+                                     description: "Provide the beta app description when uploading a new build",
+                                     verify_block: proc do |value|
+                                       UI.user_error!("Beta App Description must be at least 10 characters long.") unless value.to_s.length >= 10
+                                       UI.user_error!("Beta App Description must be less than 4000 characters long.") unless value.to_s.length <= 4000
+                                     end),
         FastlaneCore::ConfigItem.new(key: :beta_app_feedback_email,
                                      short_option: "-n",
                                      optional: true,


pilot/lib/pilot/build_manager.rb
```diff
@@ -67,6 +67,7 @@ def distribute(options, build = nil)
 
       unless config[:update_build_info_on_upload]
         if should_update_build_information(options)
+          UI.user_error!("Beta App Description must be at least 10 characters long.") if options[:beta_app_description].to_s.length < 10 && options[:beta_app_description].to_s.length > 0
```
yukiasai: There are other lines called `update_build_information!` method. If it fails in all cases, I think we should validate with `verify_block` in `options.rb`.

casz: Sure thing, I am still new to ruby 👍

yukiasai: @casz No probs. It's nice PR😊

pilot/lib/pilot/options.rb
```diff
@@ -46,7 +46,10 @@ def self.available_options
                                      short_option: "-d",
                                      optional: true,
                                      env_name: "PILOT_BETA_APP_DESCRIPTION",
-                                     description: "Provide the beta app description when uploading a new build"),
+                                     description: "Provide the beta app description when uploading a new build",
+                                     verify_block: proc do |value|
+                                       UI.user_error!("Beta App Description must be at least 10 characters long.") unless options[:beta_app_description].to_s.length >= 10
```
ohayon: Can you check the `value.to_s.length` rather than `options[:beta_app_description]`?

ohayon (changes requested):
This is a good change, just one little comment above. 🐙

pilot/lib/pilot/options.rb
casz: Copy paste error, my bad!

casz:
Added both minimum and maximum requirement 👍

KrauseFx (changes requested):
We try to avoid having user input checks from within _fastlane_ that actually happen on iTunes Connect or the Developer Portal, as those requirements might change in the future. Many of the user checks on Apple's side can even be changed on a developer-account level, meaning there are different requirement/numbers per account. 

Thanks for the contribution, but we'll not go ahead with this. Instead, to make the error message more clear, I'd be all for showing the iTunes Connect error message in a nicer way, if it's not the case yet 👍

PR #8531, opened 2017-03-15
Handle exit status of error_callback

giginet:
### Checklist

### Description

If executing external commands is failed and recover in error callbacks, lane should be succeed.
However, `user_error!` is always called.

This PR make if callbacks return exit code 0, then `user_error!` is not called. In other words, lane become succeed.

```ruby
sh "./failure", error_callback: lambda { |result|
  sh "./success"
  next 0
}
```

If no value is returned from the callback. `sh` action should be failed. (same as until now.)

### DIFF
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -51,8 +51,10 @@ module Fastlane
                     end
           message += "\n#{result}" if print_command_output
 
-          error_callback.call(result) if error_callback
-          UI.user_error!(message)
+          callback_exit_status = error_callback.call(result) if error_callback
+          if callback_exit_status != 0
+            UI.user_error!(message)
+          end
         end
       end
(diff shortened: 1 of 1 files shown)


giginet:
I signed it!

mfurtak:
Hm... I like what you are trying to do here, but needing to return `0` to make things work feels a little too specific and magical. I don't immediately have a better idea, but let's see what folks think!

giginet:
Hm, I agreed with your suggestion. However, I have no idea....

mfurtak:
I think @ohayon was thinking about a way to approach this...?

ohayon:
I had a thought that because the `error_callback` is a `lambda` we could just `return sh './success'`, but I'm not sure if this is the best solution?

mfurtak:
Hi everyone, I think we need to think and plan some more about this before we bring in a solution. I'm going to close this for now. 👍

Issue #8543, opened 2017-03-16
[Regression] delete_keychain(name:) in 2.20.0 causing problems


pangers:
### New Issue Checklist

- [yes ] Updated fastlane to the latest version
- [yes ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I just upgraded from 2.19.3 -> 2.20.0 and the delete_keychain(name:) action seems to be failing now (Fastfile is unchanged from 2.19.3 to 2.20.0).

I create a keychain using:

```
  create_keychain(
    name: project_name, # This is just a normal string, not a path
    default_keychain: true,
    unlock: true,
    timeout: 3600,
    lock_when_sleeps: true,
    password: project_name
  )
``` 

and delete the keychain like so:

```
delete_keychain(name: project_name)
```

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

The relevant output looks like:

```
[10:46:07]: -----------------------------
[10:46:07]: --- Step: delete_keychain ---
[10:46:07]: -----------------------------
[10:46:07]: An error occured while executing the `error` block:
[10:46:07]: Shell command exited with exit status 161 instead of 0.
+---------------------------+----------------------------------------------------------------------+
|                                           Lane Context                                           |
+---------------------------+----------------------------------------------------------------------+
| DEFAULT_PLATFORM          | ios                                                                  |
| PLATFORM_NAME             | ios                                                                  |
| LANE_NAME                 | ios build_only                                                       |
| ORIGINAL_DEFAULT_KEYCHAIN | "/Users/james.pang/Library/Keychains/login.keychain"                 |
| CERT_FILE_PATH            | /Users/james.pang/Desktop/Projects/iOS/FastlaneExample/CertsAndPr... |
| CERT_CERTIFICATE_ID       | Y2528PM4JX                                                           |
| PRODUCE_APPLE_ID          |                                                                      |
| SIGH_PROFILE_PATH         | /Users/james.pang/Desktop/Projects/iOS/FastlaneExample/CertsAndPr... |
| SIGH_PROFILE_PATHS        | ["/Users/james.pang/Desktop/Projects/iOS/FastlaneExample/CertsAnd... |
| SIGH_UDID                 | c268e55a-3c8f-4651-8c0f-949b7449aaa1                                 |
[...]

PR #8467, opened 2017-03-07
Fix regression with delete_keychain(name:)

mfurtak:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Description

This attempts to allow both the `:name` and `:keychain_path` style calls to work.

### Motivation and Context

#8003 introduced a regression for some users that were already using `delete_keychain` with a `:name` that was a fully-qualified path.

PR #8553, opened 2017-03-16
Address delete_keychain regression for -db keychains

mfurtak:
### Checklist

### Description

For the keychain name 'example', `create_keychain` makes a keychain at:

```
/Users/mfurtak/Library/Keychains/example-db
```

The `-db` suffix is a new wrinkle in recent macOS versions. We've been fixing places where this has broken our keychain handling, but these fixes had not yet reached `delete_keychain`.

So, in the previous version, for the keychain name 'example', `delete_keychain` would generate the command:

```
security delete-keychain /Users/mfurtak/Library/Keychains/example
```

Assuming you ran `create_keychain` and got the keychain as described above, this path doesn't exist. However, `security delete-keychain` doesn't care, and helpfully deletes `example-db` for you. So, the previous version was working somewhat by coincidence.

The most recent version of `delete_keychain` attempted to be better about searching in more places for the keychain to delete, but it also added checks to make sure that the keychains exist before trying to ask `security` to delete them. This broke the working situation above.

My proposed fix is to make `delete_keychain` use the shared utility `FastlaneCore::Helper.keychain_path` to take advantage of its shared keychain locating checks.

### Motivation and Context

#8543 reported a regression introduced by #8467 while fixing the regression caused by #8003

### DIFF
--- a/fastlane/lib/fastlane/actions/delete_keychain.rb
+++ b/fastlane/lib/fastlane/actions/delete_keychain.rb
@@ -4,25 +4,17 @@ module Fastlane
   module Actions
     class DeleteKeychainAction < Action
       def self.run(params)
-        original = Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN]
-
-        search_paths = []
-        search_paths << File.expand_path(params[:name]) if params[:name]
-        search_paths << File.expand_path(File.join("~", "Library", "Keychains", params[:name])) if params[:name]
-        search_paths << File.expand_path(params[:keychain_path]) if params[:keychain_path]
-
-        if search_paths.empty?
+        if params[:name].nil? && params[:keychain_path].nil?
           UI.user_error!("You either have to set :name or :keychain_path")
         end
 
-        keychain_path = search_paths.find { |path| File.exist?(path) }
+        original = Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN]
 
-        if keychain_path.nil?
-          UI.user_error!("Unable to find the specified keychain. Looked in:\n\t" + search_paths.join("\n\t"))
-        end
+        keychain = params[:name] || params[:keychain_path]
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/delete_keychain.rb
```diff
@@ -4,25 +4,17 @@ module Fastlane
   module Actions
     class DeleteKeychainAction < Action
       def self.run(params)
-        original = Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN]
-
-        search_paths = []
-        search_paths << File.expand_path(params[:name]) if params[:name]
-        search_paths << File.expand_path(File.join("~", "Library", "Keychains", params[:name])) if params[:name]
-        search_paths << File.expand_path(params[:keychain_path]) if params[:keychain_path]
```
lyndsey-ferguson: Before, this would look for all the possible names of the keychain. Could you not add another search path with the '-db' filename? The `search_paths.find` will be sure to find the one path

mfurtak: We could, but the realization I made is that this is what the `Helper` method does. It felt better to keep this name discovery logic consistent in that one place.

Does that make sense, or am I misunderstanding?

lyndsey-ferguson: No, makes perfect sense and it is doing just what this used to do. Great change, thanks!

lyndsey-ferguson: For reference, [here's the relevant code](https://github.com/fastlane/fastlane/blob/f7a5071122aad789d7c1451509cbbd1efea494e8/fastlane_core/lib/fastlane_core/helper.rb#L203) in the helper method

mcfedr:
This appears to be a duplicate of https://github.com/fastlane/fastlane/pull/8511

mfurtak:
Ah, thanks for that @mcfedr - I missed this earlier work :+1:

mfurtak:
Closed in favor of #8511

PR #8558, opened 2017-03-16
Add support for fish tab completions

JosephDuffy:
### Checklist
- [x] I've updated the documentation if necessary (I couldn't find anything that would need updating)

### Description
If the user’s default shell is [fish](https://fishshell.com/) when `fastlane enable_auto_complete` is ran a completion file will be copied to `~/.config/fish/completions`, rather than the user being prompted to add a shell script to their rc file.

### Motivation and Context
Prior to this PR fastlane did not support tab completions with the fish shell, meaning that when running `fastlane enable_auto_complete` users would be provided with instructions that will not work in their shell, and would be left without a way of enabling tab completion.

To test these changes I ran `bundle exec fastlane enable_auto_complete` with fish as my default shell. A `fastlane.fish` file was copied to `~/.config/fish/completions`. I then changed my default shell to bash using `chsh -s /bin/bash`, loaded a new session, and ran `bundle exec fastlane enable_auto_complete` again. This time the old behaviour happened.

### DIFF
--- a/fastlane/lib/assets/completions/completion.fish
+++ b/fastlane/lib/assets/completions/completion.fish
@@ -1,5 +1,5 @@
 # This function was taken from https://github.com/Carthage/Carthage/blob/master/Source/Scripts/carthage-fish-completion
-function __fish_prog_needs_subcommand
+function __fish_fastlane_needs_subcommand
   set cmd (commandline -opc)
   if [ (count $cmd) -eq 1 -a $cmd[1] = 'fastlane' ]
     return 0
@@ -34,4 +34,4 @@ for line in $commands
   end
 end
 
-complete -c fastlane -n '__fish_prog_needs_subcommand' -a (string trim $commands_string) -f
+complete -c fastlane -n '__fish_fastlane_needs_subcommand' -a (string trim $commands_string) -f
--- a/fastlane/lib/assets/completions/completion.fish
+++ b/fastlane/lib/assets/completions/completion.fish
@@ -17,7 +17,7 @@ else
   exit 1
 end
 
-set commands (string match --regex '.*lane\ \:([^\s]*)\ do' (cat $file))
+set commands (string match --regex '.*lane\ \:(?!private_)([^\s]*)\ do' (cat $file))
 
 set commands_string
 
--- a/fastlane/lib/assets/completions/completion.fish
+++ b/fastlane/lib/assets/completions/completion.fish
@@ -1,0 +1,37 @@
+# This function was taken from https://github.com/Carthage/Carthage/blob/master/Source/Scripts/carthage-fish-completion
+function __fish_prog_needs_subcommand
+  set cmd (commandline -opc)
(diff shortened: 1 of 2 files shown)


JosephDuffy:
I signed it!

fastlane/lib/assets/completions/completion.fish
```diff
@@ -0,0 +1,37 @@
+# This function was taken from https://github.com/Carthage/Carthage/blob/master/Source/Scripts/carthage-fish-completion
+function __fish_prog_needs_subcommand
```
milch: IIRC fish doesn't deal gracefully with multiple functions being defined with the same name. I'd rename the function to something like `__fish_fastlane_needs_subcommand` to be sure.

fastlane/lib/assets/completions/completion.fish
```diff
@@ -0,0 +1,37 @@
...
+  set file "Fastfile"
+else if test -e "fastlane/Fastfile"
+  set file "fastlane/Fastfile"
+else if test -e ".fastlane/Fastfile"
+  set file ".fastlane/Fastfile"
+else
+  exit 1
+end
+
+set commands (string match --regex '.*lane\ \:(?!private_)([^\s]*)\ do' (cat $file))
```
milch: Is there a reason this is done differently than in the other completion files? (i.e. [zsh](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/assets/completions/completion.zsh#L17) and [bash](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/assets/completions/completion.bash#L18))

milch (changes requested):
Hi, thanks for your PR! I left a couple questions and comments in the review, please take a look.

fastlane/lib/assets/completions/completion.fish
JosephDuffy: I was not aware of this, thanks! I will address this in a commit later

fastlane/lib/assets/completions/completion.fish
JosephDuffy: No specific reason, although I did write the regex before I knew of the other scripts. From what I can see the other scripts don't check for the `do` suffix and will include private lanes. Should I update anything?

milch: Hmm, I think this is fine for now as it'll require some more thinking. I don't like that we have essentially duplicate logic across 3 files right now. 

Just rebase to the current master please, then we can merge your PR

JosephDuffy: I agree that this could be improved. Let me know if you decide on any changes and I will create a new PR.

My branch is now up to date with fastlane's master :)

mfurtak:
Hi folks - any other changes needed to get this in?

milch (approved):

milch:
@mfurtak Nope, it was ready. Sorry for the delay!

mfurtak:
No worries, thanks folks! 🙌

PR #8565, opened 2017-03-17
Revise action alias handling

nafu:
### Checklist

### Description

Revise action alias handling if no params are passed.

### Motivation and Context

When you use params inside `alias_used` but not passed any params, that cause NoMethodError.

```
Failures:

  1) Fastlane Fastlane::Action can call alias action alias does not crash with no param
     Failure/Error:
         somealias
       end").runner.execute(:test)

     NoMethodError:
       undefined method `[]=' for nil:NilClass
```

You can check this behavior with [this commit](https://github.com/nafu/fastlane/commit/36c1e6c144ec79670cccbc97f5c46e82161911c1).

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -101,6 +101,17 @@ module Fastlane
       nil
     end
 
+    # Pass a action alias symbol (e.g. :enable_automatic_code_signing)
+    # and this method will return a reference to the action class
+    # if it exists. In case the action with this alias can't be found
+    # this method will return nil.
+    def class_reference_from_action_alias(method_sym)
+      alias_found = find_alias(method_sym.to_s)
+      return nil unless alias_found
+
+      class_reference_from_action_name(alias_found.to_sym)
+    end
+
     # lookup if an alias exists
     def find_alias(action_name)
       Actions.alias_actions.each do |key, v|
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/runner.rb
```diff
@@ -119,14 +130,12 @@ def trigger_action_by_name(method_sym, custom_dir, from_action, *arguments)
...
+        class_ref = class_reference_from_action_alias(method_sym)
+        # notify action that it has been used by alias
+        if class_ref.respond_to?(:alias_used)
           orig_action = method_sym.to_s
-          class_ref = class_reference_from_action_name(alias_found.to_sym)
-          # notify action that it has been used by alias
-          if class_ref.respond_to?(:alias_used)
-            class_ref.alias_used(orig_action, arguments.first)
-          end
+          arguments = [{}] if arguments.empty?
```
hjanuschka: sorry to ask the dumb question, what does the `[{}]` do? intialize an array of hashes? (ruby magic 😵 )

hjanuschka (approved):
that is awesome, many thx!

fastlane/lib/fastlane/runner.rb
nafu: @hjanuschka Thanks for reviewing.

It initializes an array of its first element as hash. 👍 

```ruby
irb(main):001:0> Array.new(1, Hash.new)
=> [{}]
```

hjanuschka: thx @nafu - never stop learning 👍 
i would have done it like

```ruby
arguments = [] 
arguments << {}
```

hjanuschka:
@nafu  see https://github.com/fastlane/docs/pull/302#discussion_r106592025 i think it is another problem.

nafu:
@hjanuschka 👍  thx

Here is another thought about `TestSampleCodeAction`.
`TestSampleCodeAction` should test if all actions works correctly.
That's why it would be better to mock `trigger_action_by_name` ideally and `alias_used` should also be called in this case.

Actually, I found the bugs by mocking [a part of `trigger_action_by_name`](https://github.com/fastlane/fastlane/blob/91980e55806b165571614c5129ebc91e63007e15/fastlane/lib/fastlane/runner.rb#L120-L131) in `TestSampleCodeAction`.

PR #8577, opened 2017-03-19
Add feature to set env variables on Fastfile

giginet:
### Checklist

### Description

Introduce new feature to set environment variables on `Fastfile`.

You can use `env` to define environment variables on DSL.

```ruby
env 'FASTLANE_TEST_ENV_GLOBAL', 'foobar'

lane :echo do
  puts 'global'
end

platform :ios do
  env 'FASTLANE_TEST_ENV_PLATFORM', 'ios'

  lane :echo do
    puts 'ios'
  end
end

platform :android do
  env 'FASTLANE_TEST_ENV_PLATFORM', 'android'

  lane :echo do
    puts 'android'
  end
end
```

### Motivation and Context

If we want to set environment variables, we should use `before_all`.

```ruby
before_all do
  ENV["FASTLANE_HIDE_CHANGELOG"] = "1"
end
```

It is not readable.
For example, you can set the environment variable like following.

```ruby
env 'FASTLANE_HIDE_CHANGELOG', '1'
```

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -116,6 +116,11 @@ module Fastlane
       self.current_platform = nil
     end
 
+    # Define environment variables
+    def env(name, value)
+      @runner.add_env(self.current_platform, name, value)
+    end
+
     # Is executed before each test run
     def before_all(&block)
       @runner.set_before_all(@current_platform, block)
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -9,6 +9,9 @@ module Fastlane
     # @return [Hash] All the lanes available, first the platform, then the lane
     attr_accessor :lanes
 
+    # @return [Hash] All the environment variables to set, first the platform, then the variable name
+    attr_accessor :environment_variables
+
     def full_lane_name
       [current_platform, current_lane].reject(&:nil?).join(' ')
     end
@@ -43,6 +46,10 @@ module Fastlane
       parameters ||= {}
       begin
         Dir.chdir(path_to_use) do # the file is located in the fastlane folder
+          environment_variables[current_platform] ||= {}
+          environment_variables[current_platform].each do |name, key|
+            ENV[name] = key
+          end
           execute_flow_block(before_all_blocks, current_platform, current_lane, parameters)
           execute_flow_block(before_each_blocks, current_platform, current_lane, parameters)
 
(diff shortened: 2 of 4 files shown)


mfurtak:
This is an interesting idea. 👍  What happens if someone uses `env` inside of a lane?

fastlane/lib/fastlane/fast_file.rb
```diff
@@ -116,6 +116,11 @@ def platform(platform_name)
       self.current_platform = nil
     end
 
+    # Define environment variables
+    def env(name, value)
+      @runner.add_env(self.current_platform, name, value)
```
KrauseFx: Should we do some input verifications here? Maybe define `env` with no parameters and print out an error that tells the user how to provide parameters?

fastlane/lib/fastlane/runner.rb
```diff
@@ -43,6 +46,10 @@ def execute(lane, platform = nil, parameters = nil)
       parameters ||= {}
       begin
         Dir.chdir(path_to_use) do # the file is located in the fastlane folder
+          environment_variables[current_platform] ||= {}
+          environment_variables[current_platform].each do |name, key|
+            ENV[name] = key
```
KrauseFx: Let's also `UI.verbose` the `name` and `key` so that in verbose mode the user can see that the environment variable is being set 👍

fastlane/spec/fixtures/fastfiles/FastfileEnv
```diff
@@ -0,0 +1,21 @@
...
+  end
+end
+
+platform :android do
+  env 'FASTLANE_TEST_ENV_PLATFORM', 'android'
+
+  lane :echo do
+    puts 'android'
+  end
+end
```
KrauseFx: I'd love to have the tests also cover the case where the user adds the `env` call inside a lane

KrauseFx (changes requested):
Great idea, I added some comments. Also this needs to be documented somewhere too, where do you think is the best spot to so?

KrauseFx:
Hey @giginet, I was just talking with @mfurtak that we really like this change, let us know when you had some time to update the PR 👍 Thanks

KrauseFx:
If you want, I can write the documentation for it 👍

KrauseFx:
Hey @giginet, what's the status on this PR? Do you want me to write the documentation for this feature, or do you do it? 👍

KrauseFx:
Also, did you have time to address `What happens if someone uses env inside of a lane?` by @mfurtak?

KrauseFx:
I'll close this PR for now, but feel free to re-open whenever you're ready 👍

PR #8582, opened 2017-03-20
Fix actions install on device

rafaelsachetto:
### Checklist

### Description
Escape ipa path to compile using ios-deploy at action install on device

### Motivation and Context
My app name consists of two words and it was necessary to escape the same to use the function (action) install on device

### DIFF
--- a/fastlane/lib/fastlane/actions/install_on_device.rb
+++ b/fastlane/lib/fastlane/actions/install_on_device.rb
@@ -11,7 +11,7 @@ module Fastlane
           "ios-deploy",
           params[:extra],
           "--bundle",
-          params[:ipa]
+          params[:ipa].gsub(/ /, '\ ')
         ]
         taxi_cmd << "--no-wifi" if params[:skip_wifi]
         taxi_cmd << ["--id", params[:device_id]] if params[:device_id]
--- a/fastlane/lib/fastlane/actions/install_on_device.rb
+++ b/fastlane/lib/fastlane/actions/install_on_device.rb
@@ -11,7 +11,7 @@ module Fastlane
           "ios-deploy",
           params[:extra],
           "--bundle",
-          params[:ipa].gsub(/ /, '\ ')
+          params[:ipa].shellescape
         ]
         taxi_cmd << "--no-wifi" if params[:skip_wifi]
         taxi_cmd << ["--id", params[:device_id]] if params[:device_id]
--- a/fastlane/spec/actions_specs/install_on_device_spec.rb
+++ b/fastlane/spec/actions_specs/install_on_device_spec.rb
@@ -7,6 +7,13 @@ describe Fastlane do
         end").runner.execute(:test)
         expect(result).to eq("ios-deploy  --bundle demo.ipa")
       end
+
+      it "generates a valid command using app name with space" do
+        result = Fastlane::FastFile.new.parse("lane :test do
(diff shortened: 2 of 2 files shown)


fastlane/lib/fastlane/actions/install_on_device.rb
```diff
@@ -11,7 +11,7 @@ def self.run(params)
           "ios-deploy",
           params[:extra],
           "--bundle",
-          params[:ipa]
+          params[:ipa].gsub(/ /, '\ ')
```
giginet: Use `shellescape` instead

`params[:ipa].shellescape`

giginet: `ipa` param seems to be optional.

So you have to check nil.

https://github.com/fastlane/fastlane/pull/8582/files#diff-cec084ffdd8e2b1d290fbc3ba9bacb06R55

fastlane/spec/actions_specs/install_on_device_spec.rb
```diff
@@ -7,6 +7,13 @@
         end").runner.execute(:test)
         expect(result).to eq("ios-deploy  --bundle demo.ipa")
       end
+
+      it "generates a valid command using app name with space" do
+        result = Fastlane::FastFile.new.parse("lane :test do
+          install_on_device(ipa: 'demo app.ipa')
+        end").runner.execute(:test)
+        expect(result).to eq("ios-deploy  --bundle demo\\ app.ipa")
```
giginet: Why does it have two spaces?

I think this behavior is strange. it should be fixed.

giginet (changes requested):
Thank you contribution.

Your suggestion is good. however, some codes should be fixed.

fastlane/lib/fastlane/actions/install_on_device.rb
rafaelsachetto: But it is not, it will always pass the bundle once it is needed for compilation.

fastlane/spec/actions_specs/install_on_device_spec.rb
rafaelsachetto: I tried to solve this "problem" but I saw that it would be an unnecessary disorder. Feel free to settle.

fastlane/lib/fastlane/actions/install_on_device.rb
KrauseFx: Please replace this with shellescape as instructed by @giginet 👍

fastlane/spec/actions_specs/install_on_device_spec.rb
KrauseFx: It seems like the double space issue is not related to this PR, but has something to do with a `nil` value in `params[:extra]`, maybe something that should be fixed in a separate PR?

KrauseFx (approved):
Thanks for the pull request and also adding a test 👍

fastlane/spec/actions_specs/install_on_device_spec.rb
rafaelsachetto: @KrauseFx I have an idea how to fix, I'll create a new PR to clean this.

giginet (approved):

KrauseFx:
Looks great, thanks @rafaelsachetto 👍

Issue #8532, opened 2017-03-15
No `Podfile' found in the project directory.


pronebird:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Just copied Fastfile from Fabric.io and tried `fastlane beta` and it didn't work.

```
+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 💥   | cocoapods                           | 0           |
+------+-------------------------------------+-------------+

[14:09:58]: fastlane finished with errors

[!] Exit status of command 'pod install' was 1 instead of 0.
[!] No `Podfile' found in the project directory.
```

Podfile is right there:

```
➜  XXX git:(master) ✗ ls -la | grep file            
-rw-r--r--   1 pronebird  staff   813 Mar 15 14:08 Fastfile
-rw-r--r--   1 pronebird  staff    76 Mar 15 14:01 Gemfile
-rw-r--r--   1 pronebird  staff  4763 Mar 15 14:01 Gemfile.lock
-rw-r--r--@  1 pronebird  staff   455 Mar  9 23:58 Podfile
-rw-r--r--   1 pronebird  staff  1326 Mar  9 23:59 Podfile.lock
```

pronebird:
I just ran `bundle exec pod install` and it works. Is there any way to know the location where it runs `pod install`? Can't be that hard to force it run in current directory?

KrauseFx:
This should work by default, sorry you ran into this issue. Would you mind sharing a screenshot of your project directory, containing both the fastlane folder, Fastfile and the Podfile? I assume you didn't put the `Fastfile` inside a `fastlane` directory, which is required by _fastlane_ 👍  Let me know if that's the case

pronebird:
Hi @KrauseFx,

You are right, I forgot to put Fastfile into `fastlane` folder and when I did everything worked like a charm. 

Could you please update https://fabric.io/features/distribution?utm_campaign=fastlane.tools to reflect that? I skimmed through that page many times but see no mention of subfolder requirement, even downloaded ZIP does not have `fastlane/` folder inside.

Maybe something like this as a hint:

[...]

PR #8593, opened 2017-03-21
Fix using the Fastfile if parent dir isn't named fastlane

KrauseFx:
Add tests for all those cases, and ensured that tests fail without this patch

Fixes https://github.com/fastlane/fastlane/issues/8532

### DIFF
--- a/fastlane_core/spec/fastlane_folder_spec.rb
+++ b/fastlane_core/spec/fastlane_folder_spec.rb
@@ -1,0 +1,45 @@
+describe FastlaneCore::FastlaneFolder do
+  let (:fastfile_name) { "Fastfile" }
+  describe "#path" do
+    it "returns the fastlane path if it exists" do
+      expect(FastlaneCore::FastlaneFolder.path).to eq("./fastlane/")
+    end
+
+    it "returns nil if the fastlane path isn't available" do
+      Dir.chdir("/") do
+        expect(FastlaneCore::FastlaneFolder.path).to eq(nil)
+      end
+    end
+
+    it "returns the path if fastlane is being executed from the fastlane directory" do
+      Dir.chdir("fastlane") do
+        expect(FastlaneCore::FastlaneFolder.path).to eq('./')
+      end
+    end
+  end
+
+  describe "#fastfile_path" do
+    it "uses a Fastfile that's inside a fastlane directory" do
+      expect(FastlaneCore::FastlaneFolder.fastfile_path).to eq(File.join(".", "fastlane", fastfile_name))
+    end
+
+    it "uses a Fastfile that's inside a fastlane directory if cd is inside a fastlane dir" do
+      Dir.chdir("fastlane") do
+        expect(FastlaneCore::FastlaneFolder.fastfile_path).to eq(File.join(".", fastfile_name))
+      end
+    end
+
+    it "doesn't try to use a Fastfile if it's not inside a fastlane folder" do
+      Dir.chdir("sigh") do
+        File.write(fastfile_name, "")
+        begin
+          expect(FastlaneCore::FastlaneFolder.fastfile_path).to eq(nil)
+        rescue RSpec::Expectations::ExpectationNotMetError => ex
+          raise ex
+        ensure
+          File.delete(fastfile_name)
+        end
+      end
+    end
+  end
+end
--- a/fastlane_core/lib/fastlane_core/fastlane_folder.rb
+++ b/fastlane_core/lib/fastlane_core/fastlane_folder.rb
@@ -13,7 +13,9 @@ module FastlaneCore
 
     # Path to the Fastfile inside the fastlane folder. This is nil when none is available
     def self.fastfile_path
-      path = File.join(self.path || '.', 'Fastfile')
+      return nil if self.path.nil?
+
+      path = File.join(self.path, 'Fastfile')
       return path if File.exist?(path)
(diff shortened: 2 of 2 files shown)


fastlane_core/spec/fastlane_folder_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+    end
+  end
+
+  describe "#fastfile_path" do
+    it "uses a Fastfile that's inside a fastlane directory" do
+      expect(FastlaneCore::FastlaneFolder.fastfile_path).to eq(File.join(".", "fastlane", fastfile_name))
+    end
+
+    it "uses a Fastfile that's inside a fastlane directory if cd is inside a fastlane dir" do
```
nafu: I prefer it looks same if it has same context.

> it "returns the path if fastlane is being executed from the fastlane directory" do

fastlane_core/spec/fastlane_folder_spec.rb
```diff
@@ -0,0 +1,45 @@
+describe FastlaneCore::FastlaneFolder do
+  let (:fastfile_name) { "Fastfile" }
+  describe "#path" do
```
nafu: It's class method, so it should be `describe ".path"`.

fastlane_core/spec/fastlane_folder_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    it "returns the path if fastlane is being executed from the fastlane directory" do
+      Dir.chdir("fastlane") do
+        expect(FastlaneCore::FastlaneFolder.path).to eq('./')
+      end
+    end
+  end
+
+  describe "#fastfile_path" do
```
nafu: It's class method, so it should be `describe ".fastfile_path"`.

nafu (approved):
Looks good to me 👍 
I left some comments for better specs.

fastlane_core/spec/fastlane_folder_spec.rb
KrauseFx: Interesting, I thought it would be `#`. We don't use the `.` across the code base currently. Is that something you'd want to change? I don't feel like it's worth the effort.

nafu: `#` indicates instance method in ruby, so this is confusing at least to me.
If we have some guidelines for this, it would be ok 👍 

### [In ruby documentation](http://ruby-doc.org/documentation-guidelines.html)

- `.` (or `::`) for class method
- `#` for instance method

e.x. [`class File`](https://docs.ruby-lang.org/en/2.4.0/File.html)

buret:
Hi, why having added this restriction on the folder name? I have different git repositories for my 'fastlane' tasks: one using 'match' to manage all my provisioning profiles, one with a Fastfile to build all my projects, etc. They all have diffrent names and contain a Fastfile. I think it's a pity to have this naming constraint :(

EskiMag:
What's the point of this change? All my fastlane based builds are now broken. This change is communicated nowhere. I was just using `Fastfile` placed at the root of my repository everywhere. Now I have to move it to `fastlane` folder and propagate this change into all my branches etc. Am I right? :-(

KrauseFx:
Hey @EskiMag, we didn't want to do that, and we want to fix this. Really sorry for introducing this. Can you share more about your current fastlane setup? Ideally a screenshot of your project's root and the Fastfile and fastlane folder (if you have one). 

We've always forced people to have a fastlane folder that contains a Fastfile, however it broke a few months ago. This PR reverted the regressions, but it looks like maybe we should just allow the Fastfile to be there.

@buret Can you also share a screenshot of your current directory structure

EskiMag:
Hi @KrauseFx  I will be able to put all `Fastfile` into `fastlane` folders, but it was quite a surprise for me, since I did the update with good intentions and the result was exactly the opposite. :-) Nevermind, I was just little bit upset, sorry about that. Still, Fastlane is the best thing since sliced bread. :-D
As for my setup. I have a few different ones. But everywhere I had `Fastfile` directly in the root of my repository, without the `fastlane` folder. All lanes are specified in the `Fastfile`, but I am somehow just starting to use `fastlane` as a tool, so it will be growing probably, so putting everything into the `fastlane` folder will be a good thing to do anyway. Unfortunately, I can't do a screenshot for you. For me, this change is not a problem anymore, thanks for clarification and help.

KrauseFx:
Hey @EskiMag, that's great to hear, thanks for letting us know. The problem with having a Fastfile directly in the root without a `fastlane` folder as parent, is that many actions will behave differently, and will be executed in the wrong directory. 

Please do move the file into the `fastlane` directory, and make sure everything works as expected, you might have to update `sh` calls inside your Fastfile

mikocot:
I have similar feeling to @EskiMag and @buret. We usually copy the fastfile together with other resources through SSH and run it temporary folders that reflect current publishing request. We should be able to adapt the new structure requirement, but this development comes as a big surprise. We rarely touch those setups, as the requirements don't change, but then we are always afraid to update the tools for exactly this reason - simple minor version update breaks the whole, previously working process. This is always a pain.
It was also quite hard to get here, from the error message you get.

KrauseFx:
Really sorry about this @mikocot. This shouldn't have worked in the past 😢

mikocot:
well, it's still the best product of this kind on the market :)

PR #8605, opened 2017-03-22
Improve react-native onboarding

KrauseFx:
- Add custom error handling for react-native projects with no bundle identifier set
- Tell user to navigate to platform subfolder on `fastlane init` for react-native projects
- This PR also moves out 2 method calls that shouldn't be part of the `begin rescue` block

### DIFF
--- a/fastlane/lib/fastlane/setup/setup.rb
+++ b/fastlane/lib/fastlane/setup/setup.rb
@@ -14,6 +14,11 @@ module Fastlane
       elsif is_android?
         UI.message("Detected Android project in current directory...")
         platform = :android
+      elsif is_react_native?
+        UI.important("Detect react-native app. To set up fastlane, please run")
+        UI.command("fastlane init")
+        UI.important("in the sub-folder for each platform (\"ios\" or \"android\")")
+        UI.user_error!("Please navigate to the platform subfolder and run `fastlane init` again")
       else
         UI.message("Couldn't automatically detect the platform")
         val = agree("Is this project an iOS project? (y/n) ".yellow, true)
@@ -35,6 +40,10 @@ module Fastlane
 
     def is_android?
       Dir["*.gradle"].count > 0
+    end
+
+    def is_react_native?
+      SetupIos.new.project_uses_react_native?(path: "./ios")
     end
 
     def show_analytics
--- a/fastlane/lib/fastlane/setup/setup.rb
+++ b/fastlane/lib/fastlane/setup/setup.rb
@@ -15,7 +15,7 @@ module Fastlane
         UI.message("Detected Android project in current directory...")
         platform = :android
       elsif is_react_native?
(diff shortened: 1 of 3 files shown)


mfurtak:
It would be good to have tests to cover this new functionality, and ensure it stays working :+1:

KrauseFx:
@mfurtak added tests for the 2 cases I added in this PR 👍

fastlane/lib/fastlane/setup/setup_ios.rb
```diff
@@ -21,9 +21,11 @@ def run(user: nil)
       FastlaneCore::FastlaneFolder.create_folder! unless Helper.is_test?
       is_manual_setup = false
 
+      setup_project
+      react_native_pre_checks
+      ask_for_apple_id
```
giginet: Why did you move them from inside `begin` block?

KrauseFx: Because we don't access iTunes Connect or the Dev Portal at this point yet, so it's better ot not be inside the `begin`, see also the PR body 👍

giginet (approved):

fastlane/lib/fastlane/setup/setup.rb
```diff
@@ -14,9 +14,14 @@ def run(user: nil)
       elsif is_android?
         UI.message("Detected Android project in current directory...")
         platform = :android
+      elsif is_react_native?
+        UI.important("Detectedj react-native app. To set up fastlane, please run")
```
mfurtak: Typo in "Detected"

fastlane/lib/fastlane/setup/setup_ios.rb
```diff
@@ -70,6 +73,29 @@ def try_manual_setup
...
+        error_message << "in the `General` section of your project settings."
+        error_message << "Restart `fastlane init` once you're done!"
+        UI.user_error!(error_message.join(" "))
+      end
+    end
+
+    def self.project_uses_react_native?(path: Dir.pwd)
+      package_json = File.join(path, "..", "package.json")
+      return unless File.basename(path) == "ios"
+      return unless File.exist?(package_json)
```
mfurtak: I think it would be better if this method always returned `true` or `false`, instead of sometimes returning `nil`, for these checks. :+1:

mfurtak (changes requested):
Looking good! Just a few small things, then we can :ship:!

fastlane/lib/fastlane/setup/setup_ios.rb
KrauseFx: Ah, great point 👍

KrauseFx:
@mfurtak Addressed the feedback 👍

KrauseFx:
Also resolved merge confclits

mfurtak (approved):
👍

PR #8608, opened 2017-03-22
react-native: tell user to navigate to platform subfolder on `fastlane init`

KrauseFx:
screenshot 2017-03-21 06 12 16

This will show an appropriate error message when running `fastlane init` in the root of the react-native project. The Xcode and the Android project are in the subfolders.

### DIFF
--- a/fastlane/lib/fastlane/setup/setup.rb
+++ b/fastlane/lib/fastlane/setup/setup.rb
@@ -14,6 +14,11 @@ module Fastlane
       elsif is_android?
         UI.message("Detected Android project in current directory...")
         platform = :android
+      elsif is_react_native?
+        UI.important("Detect react-native app. To set up fastlane, please run")
+        UI.command("fastlane init")
+        UI.important("in the sub-folder for each platform (\"ios\" or \"android\")")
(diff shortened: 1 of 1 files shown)


KrauseFx:
I'm not 100% happy yet about this, as ideally people would use a shared Fastfile across the platforms, but at least the behavior is better that what we used to do

KrauseFx:
Still, this is ready for review and to be merged

fastlane/lib/fastlane/setup/setup.rb
```diff
@@ -37,6 +42,10 @@ def is_android?
       Dir["*.gradle"].count > 0
     end
 
+    def is_react_native?
+      SetupIos.new.project_uses_react_native?(path: "./ios")
```
mfurtak: Does this method already exist? I don't see it in the codebase...

KrauseFx: This is actually a PR to another branch of an outstanding PR (https://github.com/fastlane/fastlane/pull/8605), we need to merge this one first 👍

mfurtak (approved):

PR #8627, opened 2017-03-23
Make scan's slack_url ignore empty strings

mfurtak:
### Checklist

### Description

Adjust _scan_'s handling of the `slack_url` parameter, to treat empty strings as it would a `nil` value. i.e. it will skip sending Slack notifications.

### Motivation and Context

[This issue](https://github.com/fastlane/fastlane/issues/8583) showed confusion about the correct way to specify that _scan_ should not send notifications. While the code is behaving as expected for the way that it is written, I think this change makes it a little more tolerant for people's intent.

### DIFF
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -223,7 +223,9 @@ module Scan
                                      description: "Create an Incoming WebHook for your Slack group to post results there",
                                      optional: true,
                                      verify_block: proc do |value|
-                                       UI.user_error!("Invalid URL, must start with https://") unless value.start_with? "https://"
+                                       if !value.to_s.empty? && !value.start_with?("https://")
+                                         UI.user_error!("Invalid URL, must start with https://")
+                                       end
                                      end),
         FastlaneCore::ConfigItem.new(key: :slack_channel,
                                      short_option: "-e",
--- a/scan/spec/slack_poster_spec.rb
+++ b/scan/spec/slack_poster_spec.rb
@@ -1,0 +1,122 @@
+require 'scan'
+require 'slack-notifier'
+
+describe Scan::SlackPoster do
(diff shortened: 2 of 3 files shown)


scan/lib/scan/options.rb
```diff
@@ -223,7 +223,9 @@ def self.available_options
                                      description: "Create an Incoming WebHook for your Slack group to post results there",
                                      optional: true,
                                      verify_block: proc do |value|
-                                       UI.user_error!("Invalid URL, must start with https://") unless value.start_with? "https://"
+                                       if !value.to_s.empty? && !value.start_with?("https://")
```
KrauseFx: Is that also an issue for the fastlane slack action?

scan/spec/slack_poster_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+          })
+
+          expect(Slack::Notifier).not_to receive(:new)
+
+          Scan::SlackPoster.new.run({ tests: 0, failures: 0 })
+        end
+      end
+    end
+
+    describe "with the SLACK_URL ENV var set to empty string" do
```
KrauseFx: Beautiful, did you test if the tests are red without the patch?

KrauseFx (approved):

scan/spec/slack_poster_spec.rb
mfurtak: Yes! 👍

scan/lib/scan/options.rb
mfurtak: Hm, yes. I'll address it in a new PR

mfurtak: In further conversation, we decided to leave the action's implementation as it is 👍

PR #8632, opened 2017-03-24
Update fastlane tool collector to work with mono repo change

KrauseFx:
This fixes an issue where the version number of a sub tool wasn't properly recorded when calling `fastlane scan ...` after the mono-gem change

### DIFF
--- a/fastlane_core/lib/fastlane_core/tool_collector.rb
+++ b/fastlane_core/lib/fastlane_core/tool_collector.rb
@@ -146,9 +146,15 @@ module FastlaneCore
     end
 
     def self.determine_version(name)
-      require 'fastlane'
-      return Fastlane::VERSION if Fastlane::ActionsList.find_action_named(name.to_s)
+      unless name.to_s.start_with?("fastlane-plugin")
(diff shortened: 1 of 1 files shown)


KrauseFx:
I was looking into how to add tests to this, however couldn't find a way to do so. We already have tests in https://github.com/fastlane/fastlane/blob/master/fastlane/spec/action_collector_spec.rb#L34-L37, and they're passing as they should, however they passed before. 

The reason why the previous implementation wasn't correct is that when you call a tool with `fastlane scan ...` we might have never called `require 'fastlane'` and therefore don't have the actions loaded to check if it's a built-in action. It's safe to use the _fastlane_ version as we're a mono-gem now, with the exception of plugins, where we want to load the actual plugin version.

I'd be ok with merging it as is, but if anybody has a better idea of how to add tests, I'm more than happy to update the PR. Basically this would require us to somehow unload the `fastlane` library I think

mfurtak:
I think those existing specs cover this functionality well :+1:

mfurtak (approved):

PR #8633, opened 2017-03-24
Update tweet text to use release URL instead of tag URL

KrauseFx:
This way the message is markdown formatted. We used to use tags as they would work better with pre-mono gem releases

I really thought I submitted this exact PR a few days ago, but can't find it 😕


nafu (approved):
🚀

PR #8634, opened 2017-03-24
Fix NameError when running single rspec

nafu:
### Checklist

### Description

Fix error: `NameError: uninitialized constant Fastlane::LaneList`.

### Motivation and Context

Running below command cause `NameError: uninitialized constant Fastlane::LaneList`.

```
rspec fastlane/spec/private_public_fastfile_spec.rb
```

### DIFF
--- a/fastlane/spec/private_public_fastfile_spec.rb
+++ b/fastlane/spec/private_public_fastfile_spec.rb
@@ -25,6 +25,7 @@ describe Fastlane do
       end
 
       it "doesn't expose the private lanes in `fastlane lanes`" do
+        require 'fastlane/lane_list'
         result = Fastlane::LaneList.generate(path)
         expect(result).to include("such smooth")
         expect(result).to_not include("private call")


fastlane/spec/private_public_fastfile_spec.rb
```diff
@@ -25,6 +25,7 @@
       end
 
       it "doesn't expose the private lanes in `fastlane lanes`" do
+        require 'fastlane/lane_list'
```
giginet: I think it is better to load it from head of this file.

giginet (approved):
🚀

fastlane/spec/private_public_fastfile_spec.rb
nafu: @giginet Thanks for the review.
What is going to be better in this case?

giginet: I think it is better to put it before `describe Fastlane do`.

nafu: Yes 👍 I mean what is the benefit in this case? Why do you think it's better? I'm just curious.

giginet: I thought if someone adds specs, they face up to the same situation.

However, according to other specs, there is no rule.  So either is fine. 👍

Issue #8482, opened 2017-03-09
upload_metadata.rb - string not matched


parleer:
Within Fastfile:

```
 lane :publish_appstore do |options|
    deliver(
      username: "#{ENV['FASTLANE_APPLE_ID']}",
      name: "#{ENV['FASTLANE_APP_NAME']}",
      team_id: "#{ENV['FASTLANE_ITC_TEAM_ID']}", 
      app_identifier: "#{ENV['FASTLANE_APP_IDENTIFIER']}", 
      force: true,
      skip_screenshots: true
    )
  end
```
This generates the error:

```
/opt/local/lib/ruby2.2/gems/2.2.0/gems/fastlane-2.19.3/deliver/lib/deliver/upload_metadata.rb:189:in `[]=': [!] string not matched (IndexError)
```

Line 189 of upload_metadata.rb is indicated below:

```
    # Loads the metadata files and stores them into the options object
    def load_from_filesystem(options)
      return if options[:skip_metadata]

      # Load localised data
      Loader.language_folders(options[:metadata_path]).each do |lng_folder|
        language = File.basename(lng_folder)
        (LOCALISED_VERSION_VALUES + LOCALISED_APP_VALUES).each do |key|
          path = File.join(lng_folder, "#{key}.txt")
          next unless File.exist?(path)

          UI.message("Loading '#{path}'...")
          options[key] ||= {}
***189***          options[key][language] ||= File.read(path)
        end
      end
```

I added UI.message debug statements (not included above) and found that the error is thrown after loading metadata/name.txt, and trying to assign `options[name][en-US] = {contents of name.txt}`.

The reason this fails is because `options[name]` is already set to the name of the app. So the statement `options[key] ||= {}` does nothing, since `options[key]` already exist. But the value of `options[key]` is NOT an object, so the assignment to `options[key][language]` fails with "string not matched" error. 

My solution was to change line 188 from

`options[key] ||= {}`

to

`options[key] = {}`

This ensures that `options[key]` is always an object and setting `options[key][language]` always works.

PR #8688, opened 2017-03-28
Resolves failure when deliver() called with name: argument and name.txt exists in locale-based metadata.

parleer:
### Checklist

### Description
Resolves an issue that prevents fastlane from submitting to AppStore. Specifically, when metadata/[language]/name.txt exists AND a name is passed to deliver(), fastlane errors with 'string not matched'. Other scenarios may also reproduce this error, but this is mine. 

### Motivation and Context

In `load_from_filesystem() ` in `upload_metadata.rb`, the line `options[key][language] ||= File.read(path)` attempts to assign the value of a file to the `options[key][language]` variable. But if `options[key]` already exists, then the prior line, `options[key] ||= {}`, never gets executed and `options[key]` is not an object, thus failing with the error 'string not matched'. There is no doubt a more elegant way to resolve this issue, but this worked for me.

https://github.com/fastlane/fastlane/issues/8482

This change was required in order to deliver my app to the AppStore.

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -184,7 +184,7 @@ module Deliver
           next unless File.exist?(path)
 
           UI.message("Loading '#{path}'...")
-          options[key] ||= {}
+          options[key] = {}
           options[key][language] ||= File.read(path)
         end
       end


parleer:
I signed it.

KrauseFx:
Hey @parleer, thanks for the PR. Please update the PR title to describe what this PR changes

parleer:
I'm not sure why the circleci tests are failing. Appears unrelated to my code change?

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -184,7 +184,7 @@ def load_from_filesystem(options)
           next unless File.exist?(path)
 
           UI.message("Loading '#{path}'...")
-          options[key] ||= {}
+          options[key] = {}
```
giginet: I think your change is not so good.

According to the documentation, `options[key]` should be hash by languages.

https://github.com/fastlane/fastlane/blob/master/deliver/Deliverfile.md#metadata

```
description({
  'en-US' => "English Description here",
  'de-DE' => "Deutsche Beschreibung hier"
})
```

So, if you got the error, your `Deliverfile` would be wrong.

Your approach can't work fine in some situations.

In your Deliverfile.

```
description({
  'en-US' => "English Description here",
  'de-DE' => "Deutsche Beschreibung hier"
})
```

and there are two metadata for `en-US` and `JP`.

In above situation, these values were merged.
However, your implementation will cause loss of `de-DE`.

giginet (changes requested):
Thank you for your contribution!
Unfortunately, I think this patch should not be merged. 😞

deliver/lib/deliver/upload_metadata.rb
parleer: I agree that this PR still needs some work. 

In my case I don't have a Deliverfile. I'm calling `deliver()` from Fastfile and passing in a `name:` argument, which is set based on environment variables set during `--env` processing. I need this functionality because my app has a different name for alpha, beta, and release builds.

For now I'll just delete metadata/en-US/name.txt so that it doesn't conflict.

But there is still a fundamental problem when calling `deliver() `and passing in the `name:` argument. Fastlane shouldn't fail. I see a few possible behaviors for this condition:

1. The passed in `name:` overrides all languages, in which case a warning should be printed.

2. The passed in `name:` is not used at all, in which case a warning should be printed.

3. Fastlane fails with an error stating that a `name:` argument cannot be specified while simultaneously including a locale specific name.txt.

4. `name:` argument overrides default locale only, all others use locale-specific name.txt.

giginet: @parleer Sorry, I can't understand your situation. Could you explain in detail?
What the value you passed to `name:`?

parleer: My Fastfile contains the following call to `deliver()`

```
deliver(
      username: "#{ENV['FASTLANE_APPLE_ID']}",
      name: "#{ENV['FASTLANE_APP_NAME']}",
      team_id: "#{ENV['FASTLANE_ITC_TEAM_ID']}", 
      app_identifier: "#{ENV['FASTLANE_APP_IDENTIFIER']}", 
      force: true,
      skip_screenshots: true
    )
```
So I'm passing a specific `name:` value to `deliver()`. Internally, Fastlane is setting `options['name'] = {the value I passed as name:}`.
 
I also have the following folders/files:
```
metadata/
  en-US/
    description.txt
    keywords.txt
    name.txt
    ...
```

When Fastlane calls `load_from_filesystem(options)` (in upload_metadata.rb), it passes in the `options` hash and attempts to augment `options` with data read from the ./metadata files. 

However, as it reads the ./metadata files, it does the following:
```
options[key][language] ||= File.read(path)
```

It's this line that fails when `key` == "name" because `options['name']` is already set to a string, as described above. Since `options['name']` is a string, trying to assign `options['name']['en-US']` fails.

ohayon:
Hey @parleer thanks for submitting this PR and apologies for the long time since a response. I was wondering if you could take a look at https://github.com/fastlane/fastlane/pull/9337 and see if it solves some of the same problems you were having? We have made a couple other changes to how languages are handled recently too. Let us know if this is still a PR you are trying to get through and maybe it should be rebased on #9337 when that gets merged in? Thanks! 🚀

parleer:
David, the latest fastlane 2.38.1 appears to be working for my specific
issue. I do see the following error in the log, but it appears to be
working.

[11:49:30]: Activating language en-US...

[11:50:03]: Error with provided 'name'. Must be a hash, the key being the
language.

[11:50:03]: Uploading metadata to iTunes Connect

[11:50:05]: Successfully uploaded set of metadata to iTunes Connect

[11:50:07]: Uploading binary to iTunes Connect

[11:50:07]: Going to upload updated app to iTunes Connect

ohayon:
Hey @parleer thanks again for the PR! I am going to close this for now as it seems like some other changes fixed your issue. Let us know if there is anything else we can do to help or if there are any other changes you would want to submit 😄  🐙  🚀 !

PR #8696, opened 2017-03-29
Add custom error message when receive an internal server error

KrauseFx:
Unfortunately I couldn't find a way to test this. I did receive the following stack traces:

**error message (body)**
```
**<HTML><HEAD>
<TITLE>Internal Server Error</TITLE>
</HEAD><BODY>
<H1>Internal Server Error - Read</H1>
The server encountered an internal error or misconfiguration and was unable to
complete your request.<P>
Reference&#32;&#35;3&#46;7a6533b8&#46;1490604875&#46;742473b8
</BODY></HTML>
```
**stack trace**
```
fastlane-2.19.3/spaceship/lib/spaceship/client.rb:453:in 
fastlane-2.19.3/spaceship/lib/spaceship/tunes/tunes_client.rb:339:in 
fastlane-2.19.3/spaceship/lib/spaceship/tunes/app_version.rb:183:in 
fastlane-2.19.3/spaceship/lib/spaceship/tunes/application.rb:111:in `edit_version'
```

I'd be ok with merging this without having a way to reproduce this locally, and see if that fixes the issue, and see if still get people reporting this issue. 

By showing a more clear error message, it should make sense for the user that the issue isn't related to their setup, but has something to do with the web-services.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -97,6 +97,9 @@ module Spaceship
     # Raised when 401 is received from portal request
     class UnauthorizedAccessError < BasicPreferredInfoError; end
 
+    # Raised when 500 is received from iTunes Connect
+    class InternalServerError < BasicPreferredInfoError; end
+
     # Authenticates with Apple's web services. This method has to be called once
     # to generate a valid session. The session will automatically be used from then
     # on.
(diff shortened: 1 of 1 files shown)


mfurtak (changes requested):
We should make a different custom error type to represent a 500 response. This is different than a timeout error, IMO.

KrauseFx:
Okay, besides that the change would be ok from your side?

KrauseFx:
I was surprised to see that we didn't have a TemporaryServerError yet

mfurtak:
Yep, fine with me otherwise :+1: timeout just feels too specific

mfurtak (approved):

KrauseFx:
@mfurtak I added the custom error type 👍

mfurtak:
:+1:

PR #8697, opened 2017-03-29
Italicize instances of fastlane in README template

ohayon:
Update the plugin README template to read: _fastlane_ instead of `fastlane`.


KrauseFx (approved):
✅

PR #8701, opened 2017-03-29
[spaceship] rework reviews  📢

hjanuschka:
reworks interface for retreiving app store reviews.

```ruby
require "spaceship"
app = Spaceship::Application.find("krone.at-fivexfive")


# get all reviews from AT
reviews = app.ratings.reviews("AT")

reviews.each do | review |
  puts "#{review.rating}"
  if review.responded? 
    puts "Already been responed to"
  end
end

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_ratings.rb
+++ b/spaceship/lib/spaceship/tunes/app_ratings.rb
@@ -50,7 +50,7 @@ module Spaceship
       # @return (Array) of Review Objects
       def reviews(store_front, versionId = '')
         raw_reviews = client.get_reviews(application.apple_id, application.platform, store_front, versionId)
-        reviews = raw_reviews.map do |review|
+        raw_reviews.map do |review|
           review["value"]["application"] = self.application
           AppReview.factory(review["value"])
         end
@@ -129,6 +129,7 @@ module Spaceship
 
       def responded?
         return true if raw_developer_response
+        false
       end
     end
 
--- a/spaceship/lib/spaceship/tunes/app_ratings.rb
+++ b/spaceship/lib/spaceship/tunes/app_ratings.rb
@@ -50,12 +50,10 @@ module Spaceship
       # @return (Array) of Review Objects
       def reviews(store_front, versionId = '')
         raw_reviews = client.get_reviews(application.apple_id, application.platform, store_front, versionId)
-        reviews = []
-        raw_reviews.each do |review|
+        reviews = raw_reviews.map do |review|
           review["value"]["application"] = self.application
-          reviews << AppReview.factory(review["value"])
+          AppReview.factory(review["value"])
         end
-        reviews
       end
     end
 
@@ -77,11 +75,11 @@ module Spaceship
       })
 
       def create!(text)
-        client.create_developer_response(app_id: application.apple_id, platform: application.platform, review_id: review_id, response: text)
+        client.create_developer_response!(app_id: application.apple_id, platform: application.platform, review_id: review_id, response: text)
       end
 
-      def update(text)
-        client.update_developer_response(app_id: application.apple_id, platform: application.platform, review_id: review_id, response_id: id, response: text)
+      def update!(text)
+        client.update_developer_response!(app_id: application.apple_id, platform: application.platform, review_id: review_id, response_id: id, response: text)
       end
     end
 
@@ -102,10 +100,10 @@ module Spaceship
       attr_accessor :developer_response
 
       attr_mapping({
-        'id'       => :id,
-        'rating'   => :rating,
-        'title'    => :title,
-        'review'   => :review,
+        'id' => :id,
+        'rating' => :rating,
+        'title' => :title,
+        'review' => :review,
         'nickname' => :nickname,
         'storeFront' => :store_front,
         'appVersionString' => :app_version,
@@ -121,10 +119,7 @@ module Spaceship
         def factory(attrs)
           obj = self.new(attrs)
           response_attrs = {}
-
-          if obj.raw_developer_response
-            response_attrs = obj.raw_developer_response
-          end
+          response_attrs = obj.raw_developer_response if obj.raw_developer_response
           response_attrs[:application] = obj.application
           response_attrs[:review_id] = obj.id
           obj.developer_response = DeveloperResponse.factory(response_attrs)
@@ -132,9 +127,8 @@ module Spaceship
         end
       end
 
-      def response?
+      def responded?
         return true if raw_developer_response
-        return false
       end
     end
 
--- a/spaceship/lib/spaceship/tunes/app_ratings.rb
+++ b/spaceship/lib/spaceship/tunes/app_ratings.rb
@@ -105,6 +105,7 @@ module Spaceship
         'id'       => :id,
         'rating'   => :rating,
         'title'    => :title,
+        'review'   => :review,
         'nickname' => :nickname,
         'storeFront' => :store_front,
         'appVersionString' => :app_version,
--- a/spaceship/lib/spaceship/tunes/app_ratings.rb
+++ b/spaceship/lib/spaceship/tunes/app_ratings.rb
@@ -47,9 +47,93 @@ module Spaceship
         instance_variable_set(:@store_fronts, unfolded_store_fronts)
       end
 
-      # @return (Array) of raw hashes representing user reviews for the given store front (and optional versionId)
+      # @return (Array) of Review Objects
       def reviews(store_front, versionId = '')
-        client.get_reviews(application.apple_id, application.platform, store_front, versionId)
+        raw_reviews = client.get_reviews(application.apple_id, application.platform, store_front, versionId)
+        reviews = []
+        raw_reviews.each do |review|
+          review["value"]["application"] = self.application
+          reviews << AppReview.factory(review["value"])
+        end
+        reviews
+      end
+    end
+
+    class DeveloperResponse < TunesBase
+      attr_reader :id
+      attr_reader :response
+      attr_reader :last_modified
+      attr_reader :hidden
+      attr_reader :state
+      attr_accessor :application
+      attr_accessor :review_id
+
+      attr_mapping({
+        'responseId' => :id,
+        'response' => :response,
+        'lastModified' => :last_modified,
+        'isHidden' => :hidden,
+        'pendingState' => :state
+      })
+
+      def create!(text)
+        client.create_developer_response(app_id: application.apple_id, platform: application.platform, review_id: review_id, response: text)
+      end
+
+      def update(text)
+        client.update_developer_response(app_id: application.apple_id, platform: application.platform, review_id: review_id, response_id: id, response: text)
+      end
+    end
+
+    class AppReview < TunesBase
+      attr_accessor :application
+      attr_reader :rating
+      attr_reader :id
+      attr_reader :title
+      attr_reader :review
+      attr_reader :nickname
+      attr_reader :store_front
+      attr_reader :app_version
+      attr_reader :last_modified
+      attr_reader :helpful_views
+      attr_reader :total_views
+      attr_reader :edited
+      attr_reader :raw_developer_response
+      attr_accessor :developer_response
+
+      attr_mapping({
+        'id'       => :id,
+        'rating'   => :rating,
+        'title'    => :title,
+        'nickname' => :nickname,
+        'storeFront' => :store_front,
+        'appVersionString' => :app_version,
+        'lastModified' => :last_modified,
+        'helpfulViews' => :helpful_views,
+        'totalViews' => :total_views,
+        'developerResponse' => :raw_developer_response,
+        'edited' => :edited
+      })
+      class << self
+        # Create a new object based on a hash.
+        # This is used to create a new object based on the server response.
+        def factory(attrs)
+          obj = self.new(attrs)
+          response_attrs = {}
+
+          if obj.raw_developer_response
+            response_attrs = obj.raw_developer_response
+          end
+          response_attrs[:application] = obj.application
+          response_attrs[:review_id] = obj.id
+          obj.developer_response = DeveloperResponse.factory(response_attrs)
+          return obj
+        end
+      end
+
+      def response?
+        return true if raw_developer_response
+        return false
       end
     end
 
--- a/spaceship/lib/spaceship/tunes/app_ratings.rb
+++ b/spaceship/lib/spaceship/tunes/app_ratings.rb
@@ -73,14 +73,6 @@ module Spaceship
         'isHidden' => :hidden,
         'pendingState' => :state
       })
-
-      def create!(text)
-        client.create_developer_response!(app_id: application.apple_id, platform: application.platform, review_id: review_id, response: text)
-      end
-
-      def update!(text)
-        client.update_developer_response!(app_id: application.apple_id, platform: application.platform, review_id: review_id, response_id: id, response: text)
-      end
     end
 
     class AppReview < TunesBase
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: spaceship/lib/spaceship/tunes/app_ratings.rb [hunk 1]
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
         instance_variable_set(:@store_fronts, unfolded_store_fronts)
       end
 
-      # @return (Array) of raw hashes representing user reviews for the given store front (and optional versionId)
+      # @return (Array) of Review Objects
       def reviews(store_front, versionId = '')
-        client.get_reviews(application.apple_id, application.platform, store_front, versionId)
+        raw_reviews = client.get_reviews(application.apple_id, application.platform, store_front, versionId)
+        reviews = []
```
giginet: I think you can use `map`.

```ruby
reviews = raw_reviews.map do |review|
  review["value"]["application"] = self.application
  reviews << AppReview.factory(review["value"])
end
```

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+      attr_reader :app_version
+      attr_reader :last_modified
+      attr_reader :helpful_views
+      attr_reader :total_views
+      attr_reader :edited
+      attr_reader :raw_developer_response
+      attr_accessor :developer_response
+
+      attr_mapping({
+        'id'       => :id,
```
giginet: I think you don't need to align columns.

```ruby
attr_mapping({
  'id' => :id,
  'rating' => :rating,
....
```

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+          obj = self.new(attrs)
+          response_attrs = {}
+
+          if obj.raw_developer_response
+            response_attrs = obj.raw_developer_response
+          end
+          response_attrs[:application] = obj.application
+          response_attrs[:review_id] = obj.id
+          obj.developer_response = DeveloperResponse.factory(response_attrs)
+          return obj
```
giginet: You can remove `return`.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,46 @@ def get_rating_summary(app_id, platform, versionId = '')
...
+        req.headers['Content-Type'] = 'application/json'
+      end
+      parse_response(r, 'data')
+    end
+
     def get_reviews(app_id, platform, storefront, versionId = '')
-      r = request(:get, "ra/apps/#{app_id}/reviews?platform=#{platform}&storefront=#{storefront}&versionId=#{versionId}")
-      parse_response(r, 'data')['reviews']
+      index = 0
+      per_page = 100 # apple default
```
giginet: It should be constants.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,46 @@ def get_rating_summary(app_id, platform, versionId = '')
...
+      all_fetched = false
```
giginet: You can get rid of this variable to reduce side-effects.
It is better to `break` when the loop condition is filled.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,46 @@ def get_rating_summary(app_id, platform, versionId = '')
       parse_response(r, 'data')
     end
 
+    def create_developer_response(app_id: nil, platform: nil, review_id: nil, response: "")
```
giginet: `app_id`, `platform`, `review_id` and `response` are required.
So this interface is not good.
You don't pass the default value to arguments.

```ruby
def create_developer_response(app_id, platform, review_id, response)
```

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,46 @@ def get_rating_summary(app_id, platform, versionId = '')
...
+      }
+      r = request(:post) do |req|
+        req.url "ra/apps/#{app_id}/platforms/#{platform}/reviews/#{review_id}/responses"
+        req.body = data.to_json
+        req.headers['Content-Type'] = 'application/json'
+      end
+      parse_response(r, 'data')
+    end
+
+    def update_developer_response(app_id: nil, platform: "ios", review_id: nil, response_id: nil, response: "")
```
giginet: Same as above.
Required arguments must not have the default arguments.

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+      attr_reader :id
+      attr_reader :response
+      attr_reader :last_modified
+      attr_reader :hidden
+      attr_reader :state
+      attr_accessor :application
+      attr_accessor :review_id
+
+      attr_mapping({
+        'responseId' => :id,
```
giginet: In recent Ruby community, Hash rockets (`=>`) are not recommended.
I think you should use instead `{responseId: :id}`.

How about the coding rule of fastlane?

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+      end
+    end
+
+    class DeveloperResponse < TunesBase
+      attr_reader :id
+      attr_reader :response
+      attr_reader :last_modified
+      attr_reader :hidden
+      attr_reader :state
+      attr_accessor :application
```
giginet: Why are these values as `attr_accessor`?
These values should not be able to set from outside the classes.

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+      })
+      class << self
+        # Create a new object based on a hash.
+        # This is used to create a new object based on the server response.
+        def factory(attrs)
+          obj = self.new(attrs)
+          response_attrs = {}
+
+          if obj.raw_developer_response
+            response_attrs = obj.raw_developer_response
```
giginet: `response_attrs = obj.raw_developer_response or {}`

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+          response_attrs[:application] = obj.application
+          response_attrs[:review_id] = obj.id
+          obj.developer_response = DeveloperResponse.factory(response_attrs)
+          return obj
+        end
+      end
+
+      def response?
+        return true if raw_developer_response
+        return false
```
giginet: You can remove `return`.

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+            response_attrs = obj.raw_developer_response
+          end
+          response_attrs[:application] = obj.application
+          response_attrs[:review_id] = obj.id
+          obj.developer_response = DeveloperResponse.factory(response_attrs)
+          return obj
+        end
+      end
+
+      def response?
```
giginet: Do you mean this method?

If you mean this app review is already responded from developers.
I think it is better `responded?`

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,46 @@ def get_rating_summary(app_id, platform, versionId = '')
...
+
     def get_reviews(app_id, platform, storefront, versionId = '')
-      r = request(:get, "ra/apps/#{app_id}/reviews?platform=#{platform}&storefront=#{storefront}&versionId=#{versionId}")
-      parse_response(r, 'data')['reviews']
+      index = 0
+      per_page = 100 # apple default
+      all_fetched = false
+      all_reviews = []
+      until all_fetched
+        r = request(:get, "ra/apps/#{app_id}/platforms/#{platform}/reviews?storefront=#{storefront}&versionId=#{versionId}&index=#{index}")
```
giginet: Is this OK these requests are executed synchronously?

Isn't it needed duration between each request?

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,94 @@ def unfold_store_fronts(attrs)
...
+        'lastModified' => :last_modified,
+        'isHidden' => :hidden,
+        'pendingState' => :state
+      })
+
+      def create!(text)
+        client.create_developer_response(app_id: application.apple_id, platform: application.platform, review_id: review_id, response: text)
+      end
+
+      def update(text)
```
giginet: I think it should also have bang suffix. because, this change is destructive.

`update!`

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,46 @@ def get_rating_summary(app_id, platform, versionId = '')
       parse_response(r, 'data')
     end
 
+    def create_developer_response(app_id: nil, platform: nil, review_id: nil, response: "")
+      data = {
+        responseText: response,
```
giginet: [Question] If responseText is default value (empty string).
will API pass the value?

giginet (changes requested):
Great work! 👍 
I'm excited at this new feature.

I pointed some issues.
Let's discuss them.

spaceship/lib/spaceship/tunes/app_ratings.rb
hjanuschka: ok tests fail after this, i changed it to:

```ruby
response_attrs = {}
response_attrs = obj.raw_developer_response if obj.raw_developer_response
```

with your code above. it failes 
```ruby
undefined method `[]=' for nil:NilClass
```

spaceship/lib/spaceship/tunes/app_ratings.rb
hjanuschka: ok after testing, removing return here is faulty, and thats exactly the case why i like explicit returns, in this case a `DeveloperResponse` (which is in a attr of obj) is returned, where as it is intended to return a `AppReview` (which is hold in `obj`)

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: i am more a fan of named parameters, makes it easier to refactor later on. (but non-optional named params introduced some problems on older ruby versions, on prev. changes)

but you are right  - @KrauseFx @mfurtak can you jump in here what is the preferred way?

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: i bet it will fail 👍

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: see above

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: this would mean a constant of TunesClient? - i dont like that.

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: changed to `loop do`

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: they are synchronously, i tested it on a quiet large app - with a few hundreds of reviews and i haven't had a issue - the web interface also pages one after another.

hjanuschka (commented):
@giginet thx for your time, i addressed almost all of the feedback.

spaceship/lib/spaceship/tunes/app_ratings.rb
```diff
@@ -47,9 +47,88 @@ def unfold_store_fronts(attrs)
...
+          response_attrs = obj.raw_developer_response if obj.raw_developer_response
+          response_attrs[:application] = obj.application
+          response_attrs[:review_id] = obj.id
+          obj.developer_response = DeveloperResponse.factory(response_attrs)
+          return obj
+        end
+      end
+
+      def responded?
+        return true if raw_developer_response
```
giginet: If `raw_developer_response` is nil, `responded` don't returns `false` but `nil` .

I meant dropping `return` only.  :bowing_man:

```ruby
def responded?
  return true if raw_developer_response
  false
end
```

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,9 +323,43 @@ def get_rating_summary(app_id, platform, versionId = '')
       parse_response(r, 'data')
     end
 
+    def create_developer_response!(app_id: nil, platform: nil, review_id: nil, response: "")
```
giginet: Can `app_id`, `platform` and `review_id` be `nil`?

According to L332, these seem to be required.
If these values are required, default arguments should not be passed.
They have to be nonnull.

giginet (approved):
Thank you to fix them!

I have a few concerns yet.
If they would be fixed, LGTM.

spaceship/lib/spaceship/tunes/app_ratings.rb
hjanuschka: omg, thx for that, feeling a bit dumb now, fixed

hjanuschka: omg, thx for that, feeling a bit dumb now, fixed

spaceship/lib/spaceship/tunes/tunes_client.rb
hjanuschka: i know something like 

`def def create_developer_response!(app_id:, platform: , review_id: , response: "") `

but it does not work on some older ruby's - i did a similar mistake on the IAP-Implementation, where we reverted to default values.

hjanuschka (commented):
feedback addressed.

spaceship/lib/spaceship/tunes/tunes_client.rb
giginet: These values must not be `nil`.
I think you can drop argument labels.

```ruby
def create_developer_response!(app_id, platform, review_id, response)
end
```

Labels are needed?

hjanuschka: i personally preffer labels, in case of any refactoring its easier to handle.

eg adding a param in the middle.

`def create_developer_response!(app_id, platform,new_param, review_id, response)`

is sort a nightmare to refactor without labels, with labels it is no big deal.

giginet: OK, I understood reasons.
Instead, how about adding nil checking?

hjanuschka (commented):
@giginet i addressed the handling of the nil params, let me know if this is ready to be merged now 👍

giginet (approved):
🚀

hjanuschka:
@KrauseFx  could you  review this too, about the BC break (see pr body)

KrauseFx (changes requested):
Please wait with merging

KrauseFx:
Hey @hjanuschka, sorry for not coming back earlier. I'd be much in favor to merge the `fetching the app reviews` first. I personally would prefer not to include the `replying to reviews` part into _spaceship_ itself. Do you think we could separate those 2?

hjanuschka:
@KrauseFx  write ops - have been removed.

KrauseFx (approved):
Wonderful 🎷

nafu:
Rock! 🚀

hjanuschka:
and roll 🎶

PR #8703, opened 2017-03-30
[frameit] Added font_height_factor

### LINKED ISSUE #6116: Need fixed title height parameter for frameit

Hello everyone. 
I've tried `frameit` for its direct purpose and everything went good (after an hour figuring out parameters and possibilities) except for one thing. I can't set the parameters to get devices frames laid out identically on all screenshots for specific device type. The height of the title is always different so device frame can end up slightly (or hugely) shifted vertically relatively to other screenshots. 

My configuration is simple:

```
{
  "default": {
    "title": {
      "color": "#545454"
    },
    "background": "./background.jpg",
    "padding": 70
  },
  "data": []
}
```

If someone can help me setting this I would greatly appreciate that.

But in the meantime I implemented this possibility myself. I decided that the problem can be resolved introducing fixed title height. With fixed title height device frame will always be placed in the same position for one device type.

So before making a pull request I would like to discuss these points:
- Is it possible to conveniently achieve this with the current tool?
- I noticed that the parameter `offset` can be described in `Framefile.json` and it can contain `offset`, `width`, and `titleHeight` entries. `titleHeight` looks like everything I need but it's currently used only in `mac_editor`. Also you can't specify only `titleHeight` entry in `offset` because without `offset` entry in `offset` entry (sorry, this is actually how they're called) the latter is considered invalid:

``` ruby
if @offset_information and (@offset_information['offset'] or @offset_information['offset'])
  return @offset_information
end
UI.user_error! "Could not find offset_information for '#{screenshot}'"
```

So the question is: can I use `titleHeight` entry to specify fixed title height or it's better to introduce new entry in `Framefile.json`? 
If existing one is ok, we  should consider merging offset specified in `Framefile.json` and offset from `Offsets.image_offset(screenshot)` because we don't want to lose `offset` and `width` calculated for current screenshot.
- I don't want to actually resize title to fit fixed title height. It can become too short or too long. Instead I propose adjusting title paddings so that top space above device will always be 

```
top_space_above_device = fixed_title_height + 2 * padding
```

Again, described solution is already implemented and tested. I just want to confirm these points before making a pull request.

Thanks for attention!

chandlerhuff:
Added `font_height_factor` that allows the title and keyword heights to be a set value so that the devices line up across shots of the same device. Should fix #6116.

### Checklist

### Description
Added `font_height_factor` which if defined in the config uses a factor of the calculated font size as the text height instead of calculating the text height which makes the text height equal across all shots of the same device.

### Motivation and Context
If a title that has a lower case g/q/y/j or some other characters is used in one shot but not in another 
 even when using `font_scale_factor` to ensure the same font sizes it results in the example below where the devices don't line up like in issue #6116.
 
example

### DIFF
--- a/frameit/lib/frameit/config_parser.rb
+++ b/frameit/lib/frameit/config_parser.rb
@@ -59,6 +59,7 @@ module Frameit
     end
 
     # Make sure the paths/colors are valid
+    # rubocop:disable Metrics/PerceivedComplexity
     def validate_values(values)
       values.each do |key, value|
         if value.kind_of?(Hash)
@@ -87,9 +88,12 @@ module Frameit
             UI.user_error! "show_complete_frame must be a Boolean" unless [true, false].include?(value)
           when 'font_scale_factor'
             UI.user_error!("font_scale_factor must be numeric") unless value.kind_of?(Numeric)
+          when 'font_height_factor'
+            UI.user_error!("font_height_factor must be numeric") unless value.kind_of?(Numeric)
           end
         end
       end
     end
+    # rubocop:enable Metrics/PerceivedComplexity
   end
 end
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -217,12 +217,12 @@ module Frameit
       vertical_padding = vertical_frame_padding
       keyword_top_space = vertical_padding
 
-      spacing_between_title_and_keyword = (title.height / 2)
-      title_top_space = vertical_padding + keyword.height + spacing_between_title_and_keyword
+      spacing_between_title_and_keyword = (text_height(title) / 2)
+      title_top_space = vertical_padding + text_height(keyword) + spacing_between_title_and_keyword
       title_left_space = (background.width / 2.0 - title_width / 2.0).round
       keyword_left_space = (background.width / 2.0 - keyword_width / 2.0).round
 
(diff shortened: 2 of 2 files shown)


frameit/lib/frameit/config_parser.rb
```diff
@@ -87,9 +88,12 @@ def validate_values(values)
             UI.user_error! "show_complete_frame must be a Boolean" unless [true, false].include?(value)
           when 'font_scale_factor'
             UI.user_error!("font_scale_factor must be numeric") unless value.kind_of?(Numeric)
+          when 'font_height_factor'
+            UI.user_error!("font_height_factor must be numeric") unless value.kind_of?(Numeric)
           end
         end
       end
     end
+    # rubocop:enable Metrics/PerceivedComplexity
```
ohayon: Is there anything we can do to avoid having to disable rubocop around here?

ohayon (changes requested):
Hey @chandlerhuff this is a great fix. Just a little question about rubocop. 🚀

frameit/lib/frameit/config_parser.rb
chandlerhuff: @ohayon I don't think so adding the additional switch case increased the complexity to 19.

AndriyGo:
Is there any chance this feature could find its way into _frameit_? I am facing the exact same problem and I think that the lack of fixed height titles is really bad. If we look at the screenshots of the best apps for iOS they all have devices positioned at exactly the same spot regardless of the text in them.

PR #8708, opened 2017-03-30
[Snapshot] Clean on first build

andyvanosdale:
A new option, clean_on_first_build, that allows you to clean the build only on the first run.

A project could be configured where some build artifacts remain from prior builds but want to keep your current artifactswith the first build for screenshotting.


enozero:
I signed it!

________________________________

KrauseFx:
Could you clarify on the exact use case on when you don't want to use `:clean`, when to use `:clean` and when to use `:clean_on_first_build`. It feels pretty confusing, and ideally _snapshot_ would just to the right thing for the user 👍

enozero:
Great question! What I found when using `:clean` is it applies to the each build. For the subsequent builds, they don't necessarily need to be cleaned before the build, unless they are changing their binary due to other inputs. You still may want to clean the first build if you had previous builds. `:clean_on_first_build` will only clean the first build and allow subsequent builds to use most of the already built binaries. This could be resolved if there was an option to just `clean` instead of clean and build.

ohayon:
I agree with @KrauseFx here that this does seem a bit confusing. Is there a way that you could work around this by having an environment variable that you evaluate in your Fastfile to set or unset the `:clean` option. While I understand conceptually why an option to clean on first builds only would be wanted, I think this option introduces some complexity that can be avoided by moving this type of logic higher up in the tool.

I am going to close this PR for now, but feel free to continue discussion if you have any more thoughts! 🐙 🚀

enozero:
Forked discussion to https://github.com/fastlane/fastlane/issues/8770

PR #8714, opened 2017-03-30
Show information about how to share schemes

KrauseFx:
- Also show the message on CI systems
- Include information that the change should be included in version control

Fixes @b3ll


b3ll:
But I don't need fixing :(

hjanuschka (approved):

PR #4441, opened 2016-05-01
Make it easier to call another action from an action

KrauseFx:
Also updated the rocket action to return a :rocket: emoji

PR #8719, opened 2017-03-30
Modernize action to action call to resolve default values

tbrannam:
### Checklist
- [x] I've read the [Contribution Guidelines]

### Description
Modernize the Action to Action call to fix action default parameters.   Without the change `AppetizeAction` is called without evaluating of the `available_options`.   

Updating to the new newer syntax of `other_action.appetize(path: zipped_bundle, ...` resolves these problems.

### Motivation and Context

When `BuildAndUploadToAppetizeAction` invokes `AppetizeAction`, it loses the context of the calling lane, as well as bypasses the default mappings of `AppetizeAction::available_options`.  

As such, there is no method to assert `APPETIZE_PUBLICKEY`, or `APPETIZE_NOTE` to the underlying `AppetizeAction`'s implementation.

For an iOS build, it can be observed that calling the `BuildAndUploadToAppetizeAction`, that a console message will be emitted by `AppetizeAction` indicating that an APK is being uploaded, even if the target platform is iOS.   This is caused by the available_options not setting the default_platform setting.

[relates to #4441](https://github.com/fastlane/fastlane/pull/4441)

### DIFF
--- a/fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
+++ b/fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
@@ -18,7 +18,7 @@ module Fastlane
         zipped_bundle = Actions::ZipAction.run(path: app_path,
                                         output_path: File.join(tmp_path, "Result.zip"))
 
-        Actions::AppetizeAction.run(path: zipped_bundle,
+        other_action.appetize(path: zipped_bundle,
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
This is perfect, thanks 👍 Can you follow the instructions from the Google bot? Then we can merge

tbrannam:
I signed it!

PR #8721, opened 2017-03-31
Add skip_package_ipa option to Gym

taquitos:
Sometimes you don’t want an ipa but rather, just the xcarchive.

### Checklist

### Description
Add skip_package_ipa option to Gym that skips the rest of the packaging process after generating an xcarchive. When Gym returns, it returns the full path to the xcarchive.

### Motivation and Context
I need to build just an xcarchive that I can submit to a company-run code signing system.


gym/lib/gym/options.rb
```diff
@@ -82,6 +82,11 @@ def self.plain_options
                                      env_name: "GYM_CODE_SIGNING_IDENTITY",
                                      description: "The name of the code signing identity to use. It has to match the name exactly. e.g. 'iPhone Distribution: SunApps GmbH'",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :skip_package_ipa,
+                                     env_name: "GYM_SKIP_PACKAGE_IPA",
+                                     description: "Should we skip packaging the ipa?",
+                                     is_string: false,
+                                     optional: true),
```
KrauseFx: Instead of `optional: true`, use `default_value: false`. Both work fine, but it's more clear I think 👍

KrauseFx: Or both

gym/lib/gym/runner.rb
```diff
@@ -12,28 +12,31 @@ def run
...
       if Gym.project.ios? || Gym.project.tvos?
         fix_generic_archive # See https://github.com/fastlane/fastlane/pull/4325
+        if Gym.config[:skip_package_ipa]
+          return BuildCommandGenerator.archive_path
+        end
         package_app
         fix_package
         compress_and_move_dsym
-        path = move_ipa
+        output_path = move_ipa
```
KrauseFx: Why was this changed?

gym/lib/gym/runner.rb
```diff
@@ -12,28 +12,31 @@ def run
...
       end
       verify_archive
-      FileUtils.mkdir_p(File.expand_path(Gym.config[:output_directory]))
+      output_path = File.expand_path(Gym.config[:output_directory])
+      FileUtils.mkdir_p(output_path)
 
       if Gym.project.ios? || Gym.project.tvos?
         fix_generic_archive # See https://github.com/fastlane/fastlane/pull/4325
+        if Gym.config[:skip_package_ipa]
+          return BuildCommandGenerator.archive_path
```
KrauseFx: This is good like it is, but you for short one line statements, you can use Ruby:
```ruby
return BuildCommandGenerator.archive_path if Gym.config[:skip_package_ipa]
```
which makes it very natural to read the code

KrauseFx (changes requested):

gym/lib/gym/options.rb
taquitos: Great point

gym/lib/gym/runner.rb
taquitos: Oh, right. Thanks!

PR #8725, opened 2017-03-31
Snapshot: Split xcodebuild logs by device name and language

liamnichols:
### Checklist

### Description
Updated the `xcodebuild_log_file` method to include the device name, language and locale where required. This now means that there is a log file generated per run to prevent logs being overwritten.

#### Before
```
➜  ls -l ~/Library/Logs/snapshot 
-rw-r--r--  1 liamnichols  staff  25912  1 Apr 15:54 ExampleApp-ExampleScheme.log
```

#### After
```
➜  ls -l ~/Library/Logs/snapshot 
-rw-r--r--  1 liamnichols  staff  25912  1 Apr 15:54 ExampleApp-ExampleScheme-iPhone 6-en-US.log
-rw-r--r--  1 liamnichols  staff  25912  1 Apr 15:54 ExampleApp-ExampleScheme-iPhone 6-pt-pt_BR.log
-rw-r--r--  1 liamnichols  staff  25912  1 Apr 15:54 ExampleApp-ExampleScheme-iPhone SE-en-US.log
-rw-r--r--  1 liamnichols  staff  25912  1 Apr 15:54 ExampleApp-ExampleScheme-iPhone SE-pt-pt_BR.log
```

### Motivation and Context
I am currently running snapshot on an array of devices however my ui tests are failing randomly in some runs but not always the final device run. I try to look at the console output from xcpretty however it's not actually giving me anything useful. 

Instead I've tried looking at the output in `~/Library/Logs/snapshot/` however because the `xcodebuild` and in turn the `tee` command gets run multiple times, the log file is overwritten each time the tests run. This means that when the last run succeeds and the failure is before that, I have no way of working out what is going on :/. 

By having a unique filename for each `xcodebuild` command performed, i'm able to look back through the raw logs once `fastlane snapshot` has completed running and the important data wouldn't have been overwritten. This is also favoured over the original approach of appending `-a` to the `tee` command because that approach didn't have any way of keeping the log sizes reasonable.

I've tested this in a sample iOS project with a Snapfile that looks like the following:

```
# A list of devices you want to take the screenshots from
 devices([
   "iPhone 6",
   "iPhone SE"
 ])

languages([
  "en-US",
  ["pt", "pt_BR"] # Portuguese with Brazilian locale
])

# The name of the scheme which contains the UI Tests
 scheme "Example"
```

There are also a couple of extra tests to make sure things are working in all scenarios

-----

<details><summary>Original PR</summary>

### Description
Added the `-a` flag to the `tee` command in order to append the xcodebuild logs to the existing file when running snapshot in the `test_command_generator.rb`.

### Motivation and Context
I am currently running snapshot on an array of devices however my ui tests are failing randomly in some runs but not always the final device run. I try to look at the console output from xcpretty however it's not actually giving me anything useful. 

Instead I've tried looking at the output in `~/Library/Logs/snapshot/` however because the `xcodebuild` and in turn the `tee` command gets run multiple times, the log file is overwritten each time the tests run. This means that when the last run succeeds and the failure is before that, I have no way of working out what is going on :/. 

Adding the `-a` argument to `tee` means that the stream is appended to the end of the file rather than overwriting it each time so this will fix my issue and allow me to see the complete logs and debug further. 

I do have slight concerns about the size of the logs if it is always appending? Would this be an issue? I don't know how cleanup operations work here? I guess another option could be to append the device name to `xcodebuild_log_path` so that there are multiple files created instead however my Ruby knowledge doesn't span much further than being able to add an extra three characters into a string unfortunately. 

In terms of how I have tested it, I haven't yet. I'm not sure how to setup my machine to run from source code so if anybody can help that would be great (ruby/gems aren't my strong point). Thanks.
</details>

### DIFF
--- a/snapshot/lib/snapshot/runner.rb
+++ b/snapshot/lib/snapshot/runner.rb
@@ -171,7 +171,7 @@ module Snapshot
 
       open_simulator_for_device(device_type)
 
-      command = TestCommandGenerator.generate(device_type: device_type)
+      command = TestCommandGenerator.generate(device_type: device_type, language: language, locale: locale)
 
       if locale
         UI.header("#{device_type} - #{language} (#{locale})")
@@ -208,7 +208,7 @@ module Snapshot
                                                 end
                                               end)
 
-      raw_output = File.read(TestCommandGenerator.xcodebuild_log_path)
+      raw_output = File.read(TestCommandGenerator.xcodebuild_log_path(device_type: device_type, language: language, locale: locale))
 
       dir_name = locale || language
 
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -63,7 +63,7 @@ module Snapshot
       end
 
       def pipe
-        ["| tee -a #{xcodebuild_log_path.shellescape} | xcpretty #{Snapshot.config[:xcpretty_args]}"]
+        ["| tee #{xcodebuild_log_path.shellescape} | xcpretty #{Snapshot.config[:xcpretty_args]}"]
       end
 
       def find_device(device_name, os_version = Snapshot.config[:ios_version])
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -108,8 +108,13 @@ module Snapshot
         return ["-destination '#{value}'"]
       end
 
-      def xcodebuild_log_path
-        file_name = "#{Snapshot.project.app_name}-#{Snapshot.config[:scheme]}.log"
+      def xcodebuild_log_path(device_type: nil, language: nil, locale: nil)
+        name_components = [ Snapshot.project.app_name, Snapshot.config[:scheme] ]
+        name_components << device_type if device_type
+        name_components << language if language
+        name_components << locale if locale
+        file_name = "#{name_components.join('-')}.log"
+
         containing = File.expand_path(Snapshot.config[:buildlog_path])
         FileUtils.mkdir_p(containing)
 
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -2,7 +2,7 @@ module Snapshot
   # Responsible for building the fully working xcodebuild command
   class TestCommandGenerator
     class << self
-      def generate(device_type: nil)
+      def generate(device_type: nil, language: nil, locale: nil)
         parts = prefix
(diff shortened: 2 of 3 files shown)


liamnichols:
I signed it!

ohayon:
Hey @liamnichols! Thanks for the PR! I do think that appending the device name on to the log file path would be a safer way to handle this, and we can definitely help get that going. 

As far as running your local version of this code the recommended way would be to:
- in your project directory, add a `Gemfile` with `gem 'fastlane', path: '/path/to/your/fastlane/clone'` in the file
- make sure you have `bundler` installed by doing `gem install bundler` on your command line
- inside of your project directory, run `bundle install`
- finally, execute your local _fastlane_ by running `bundle exec fastlane <lane_name>`

Let us know if this is helpful! 🚀 🐙

liamnichols:
@ohayon: I've taken your suggestion and updated the PR so that this change is now to include the device name and language (and locale) in the log filename so that there is one log per each run.

I've edited the title and description of the original PR slightly but let me know if it needs further updates. There are a few new tests to check the behaviour and i've also tried it out on my own simple project to ensure it works.

Ps thanks for helping me get it running locally, really useful!

ohayon:
Hey @liamnichols, thanks for making those changes! This looks pretty good!

There is one issue that we want to be careful to avoid introducing: given that users are accustomed to having the log paths consistent, we should try to avoid breaking changes that would change where users might be looking for logs programmatically. 

It looks like in `snapshot/runner.rb` the extra params are always being sent through to the `TestCommandGenerator.generate` call. I think if we want to include your change, we should make sure that the params only get passed to the `TestCommandGenerator` if the user indicates somehow that they want name-spaced log files. Maybe we could add an option to _snapshot_ that would be something like `:namespace_logs`?

Curious to hear your thoughts. 🐙

KrauseFx:
Hey @liamnichols, have you had time to read and address @ohayon's comment? 👍

liamnichols:
@ohayon, @KrauseFx: hey! sorry I haven't done anything with this yet, been pretty busy lately!  

I actually don't really need this myself anymore but am keen to see if you think it's still a worthwhile change? Reason being is that i'm not great with Ruby so it'll take me some time to get these changes made myself. If you think it'll be worthwhile to add still then I'll find some time to work on it but otherwise I might potentially abandon it? WDYT?

KrauseFx:
Hey @liamnichols, thanks for the PR again, I took some time to do more testing and add a new flag, you can see it here https://github.com/fastlane/fastlane/pull/9117. Could you briefly confirm the Google CLA on the new PR so we can merge and ship the change? 👍  Thanks

PR #8734, opened 2017-03-31
Add skip_package_ipa option to Gym

taquitos:
Sometimes you don’t want an ipa but rather, just the xcarchive.

### Checklist

### Description
Add skip_package_ipa option to Gym so after building the xcarchive we don't try to package up the ipa. Gym will return the path to the archive.

### Motivation and Context
Some company build systems don't give access to certificates and they want to package up an ipa their way. This way you can output an xcarchive and hand it off to the internal build system.


gym/lib/gym/options.rb
```diff
@@ -82,6 +82,11 @@ def self.plain_options
                                      env_name: "GYM_CODE_SIGNING_IDENTITY",
                                      description: "The name of the code signing identity to use. It has to match the name exactly. e.g. 'iPhone Distribution: SunApps GmbH'",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :skip_package_ipa,
+                                     env_name: "GYM_SKIP_PACKAGE_IPA",
+                                     description: "Should we skip packaging the ipa?",
+                                     is_string: false,
```
ohayon: Should we have this be `optional: true`?

ohayon (dismissed):
Just one little thing about optionality. But otherwise this is 💯

mfurtak:
Are there tests we can add to ensure that this stays working in the future?

gym/lib/gym/options.rb
taquitos: I was thinking about that. Instead of optional, I just put ```default_value: false``` since having both seems a bit redundant.
I figure it's never optional if there is always a ```default_value```

KrauseFx: I actually provided the opposite feedback, however both work: https://github.com/fastlane/fastlane/pull/8721

KrauseFx:
@mfurtak we currently don't have any tests for `runner.rb` 😢

gym/lib/gym/options.rb
ohayon: NVM! Realized that there is the `default_value: false` there! 👍

mfurtak:
@KrauseFx @taquitos Could we? 😄

taquitos:
@mfurtak 🤔 
I don't see why not. I'll work on that, maybe Monday 👍

KrauseFx (approved):
The code is great, and it's something other people asked for too https://github.com/fastlane/fastlane/issues/8944

We can either add tests now, or create a task for this, your call ☺

taquitos:
I think I'd like to create a task to do the testing. It seems rather complicated to get the runner test up and running, and this is a trivial update.

EskiMag:
Do I understand it correctly, that now I can do something like this?
```ruby
gym(skip_packaging_ipa: true)
gym(archive_path: "<WHAT_TO_PUT_HERE?>", export_method: 'development')
gym(archive_path: "<WHAT_TO_PUT_HERE?>", export_method: 'app-store')
```
And what to use instead of `"<WHAT_TO_PUT_HERE?>"`? :-)

EskiMag:
I am sorry to be annoying, but do somebody know the answer to my question (few posts above) please? :-)

KrauseFx:
@EskiMag Please ask questions by submitting a new issue for this repo. Also, it seems like your question is not related to this PR.

PR #8744, opened 2017-04-03
Print correct message when session cookie has been expired.

tbrand:
### Checklist

### Description
Correct error message when the CD session cookie has been expired.

### Motivation and Context
When we do CD in our CI servers and the session cookie has been expired, currently we see the incorrect message like "If you're running this in a non-interactive session (e.g. server or CI)...". Because we already know how to set up it, but don't know the expiration. So I’ve corrected error message for this.


tbrand:
I signed it!

Issue #8672, opened 2017-03-27
No such file or directory @ rb_sysopen on Xcode 8.3 - Gym


Bunn:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
gym started to fail after Xcode (Version 8.3 (8E162)) installation

```
+------------------+----------+
|        Lane Context         |
+------------------+----------+
| DEFAULT_PLATFORM | ios      |
| PLATFORM_NAME    | ios      |
| LANE_NAME        | ios beta |
+------------------+----------+
[20:43:42]: No such file or directory @ rb_sysopen - /Applications/Xcode.app/Contents/Developer//Platforms/iPhoneOS.platform/Developer/usr/bin/PackageApplication

+------+------------------+-------------+
|           fastlane summary            |
+------+------------------+-------------+
| Step | Action           | Time (in s) |
+------+------------------+-------------+
| 1    | update_fastlane  | 7           |
| 2    | default_platform | 0           |
| 💥   | gym              | 48          |
+------+------------------+-------------+

[20:43:42]: fastlane finished with errors

Looking for related GitHub issues on fastlane/fastlane...

Found no similar issues. To create a new issue, please visit:
https://github.com/fastlane/fastlane/issues/new
Run `fastlane env` to append the fastlane environment to your issue
/Users/bunn/.fastlane/bin/bundle/lib/ruby/2.2.0/digest.rb:49:in `initialize': [!] No such file or directory @ rb_sysopen - /Applications/Xcode.app/Contents/Developer//Platforms/iPhoneOS.platform/Developer/usr/bin/PackageApplication (Errno::ENOENT)
	from /Users/bunn/.fastlane/bin/bundle/lib/ruby/2.2.0/digest.rb:49:in `open'
	from /Users/bunn/.fastlane/bin/bundle/lib/ruby/2.2.0/digest.rb:49:in `file'
	from /Users/bunn/.fastlane/bin/bundle/lib/ruby/2.2.0/digest.rb:34:in `file'
	from /Users/bunn/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.23.0/gym/lib/gym/xcodebuild_fixes/package_application_fix.rb:23:in `block in patch_package_application'
[...]

PR #8717, opened 2017-03-30
Catch Xcode 8.3 legacy API while patching Xcode PackageApplication script

KrauseFx:
- This will show an appropriate error message for Xcode 8.3 users that still have the legacy build API enabled. We want to migrate 100% of the users away from it, as it was now officially removed from Xcode
- Also link to the release that contains the migration information

PR #8747, opened 2017-04-03
[gym] hard error on xcode 8.3 with legacy api

hjanuschka:
hard error if xcode >= 8.3 is used and `use_legacy_build_api` is set.

error message modeled after this https://github.com/fastlane/fastlane/issues/8672#issuecomment-290544806

bildschirmfoto 2017-04-03 um 08 34 50

> originally i wanted to remove all code paths with `use_legacy_build_api` but maybe there are some folks still using xode8.2 or below.


nafu:
These changes are not enough?
https://github.com/fastlane/fastlane/pull/8692 https://github.com/fastlane/fastlane/pull/8717

hjanuschka:
@nafu  they only trigger on packaging errors, not if for example an ipa is uploaded and later on rejected on apple servers. - i think we should start (with 8.3) to hard disallow legacy api, to protect users from following errors, and therefore keeping issue impact low.

gym/lib/gym/options.rb
```diff
@@ -9,6 +9,23 @@ def self.available_options
...
+      UI.important "and update the gym call to include the export method like this:"
+      UI.important "== App Store Builds =="
+      UI.error '     gym(scheme: "MyScheme", export_method: "app-store")'
+      UI.important "==  Ad Hoc Builds =="
+      UI.error '     gym(scheme: "MyScheme", export_method: "ad-hoc")'
+      UI.important "== Development Builds =="
+      UI.error '     gym(scheme: "MyScheme", export_method: "development")'
+      UI.important "== In-House Enterprise Builds =="
+      UI.error '     gym(scheme: "MyScheme", export_method: "enterprise")'
+      UI.important "If you run into a code signing error, please check out our troubleshooting guide and common issue and how to solve them."
```
KrauseFx: Should be

> If you run into a code signing error, please check out our troubleshooting guide for more information on how to solve the most common issues

KrauseFx (changes requested):
This is a great change for sure, could you add a test for this too? Just that `use_legacy_build_api` causes a user error. Should be simple to add 👍

gym/lib/gym/options.rb
hjanuschka: done

hjanuschka:
@KrauseFx  spec's added

gym/lib/gym/options.rb
```diff
@@ -104,6 +122,9 @@ def self.plain_options
                                        if value
                                          UI.important "Don't use this option any more, as it's deprecated by Apple"
                                        end
+                                       if Gym::Xcode.legacy_api_deprecated?
+                                         Gym::Options.legacy_api_note
```
KrauseFx: At fastlane very often we use `method_name!` with a `!` to clarify that this is either modifying the object, or raising an exception if something goes wrong, could you rename the method?

KrauseFx (approved):
Looks great, one small things, and feel free to merge. Thank you ❤️

nafu:
Nice! 🚀

PR #8748, opened 2017-04-03
Make Ipa package in random tmpdir

mcfedr:
### Checklist

### Description
Don't put the ipa upload package just in `/tmp`, put just `mktmpdir` to make a unique temp name.

### Motivation and Context
On a multiuser system, one user creates the package, then pilot will fail for other users because it cannot read/write to the file.

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -1,3 +1,5 @@
+require 'tmpdir'
+
 module Pilot
   class BuildManager < Manager
     def upload(options)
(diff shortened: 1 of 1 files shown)


ohayon:
Hey @mcfedr, thanks for the PR! Just to clarify your motivation for this: are you looking to run the build as one user and then complete the _pilot_ upload as another user on the same machine? Thanks! 🐙

mcfedr:
Hi, @ohayon . I have a CI machine, with several users each running a gitlab-runner so they can run concurrently.

KrauseFx (approved):
That's a great change that makes sense for sure. However fastlane and xcodebuild doesn't allow running multiple instances afaik, so make sure to test everything

KrauseFx:
Have you tested this on your local machine?

KrauseFx:
To clarify: because you can only run one simulator per machine/user make sure that everything works. _fastlane_ itself might have problems running multiple times at the same time

mcfedr:
Yes, I have tested it on my local machine and on my CI server, and it solves the problem I was having.
Yes, only one of the users on the CI machine can run the simulator, one account runs in a UI session and the other runners can just do builds without running tests.

KrauseFx:
Awesome, thanks @mcfedr 👍

PR #8751, opened 2017-04-03
[snapshot] add option to skip helper version check

### LINKED ISSUE #8749: [Regression] Snapshot doesn't work with Xcode 7.3.1 (w/ workaround)
### New Issue Checklist

- [🚀] Updated fastlane to the latest version
- [🚀] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
The Swift syntax of SnapshotHelper.swift has been upgraded to the newer Swift language, which breaks compatibility with older versions of Xcode (7.3.1.).

This would be fine but the snapshot command fails if you're using an older version of the SnapshotHelper.swift.

### Workaround 
Manually changing the `// SnapshotHelperVersion [1.x]` to 1.3, but with the 1.2 helper syntax bypasses the check and the screenshots run.

### Suggested Fix
If an older version of Xcode is being used don't say the 1.2 Snapshot helper is out of date which currently fails the build.

Appreciate that Xcode 7.3.1 won't work forever with snapshot - but it'd be nice to have this one fixed in the core.

hjanuschka:
fixes: https://github.com/fastlane/fastlane/issues/8749

setting: `skip_helper_version_check`  to true, would disable the SwiftHelper Version check


hjanuschka:
rubocop fixes in seperate PR: https://github.com/fastlane/fastlane/pull/8752

asfalcone (approved):

Issue #8722, opened 2017-03-31
Action version_bump_podspec failure


Antondomashnev:
🔑

### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

The issue happens with the action `bump_podspec_version` when the version in podspec has addition to semver `PATCH` number.
For example:
```ruby
Pod::Spec.new do |s|
  ...
  s.version      = '2.4.4.1'
  ...
end
```

While parsing this Podspec the fastlane crashes with error: `Could not find version in podspec content 'Pod::Spec.new do...`.

I would love to contribute with the fix, but I just want to check first if it makes sense to support this kind of versions in Podspec: `x.y.z.q`? Seems like CocoaPods supports it, so maybe Fastlane should as well?

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.4                                     |
| Ruby                        | 2.3.1                                       |
| Bundler?                    | true                                        |
| Git                         | git version 2.11.0 (Apple Git-81)           |
| Installation Source         | ~/.rvm/gems/ruby-2.3.1/bin/fastlane         |
| Host                        | Mac OS X 10.12.4 (16E195)                   |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.3.1/lib                |
| OpenSSL Version             | OpenSSL 1.0.2g  1 Mar 2016                  |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 8.3                                         |

### System Locale

| Error                       |
| --------------------------- |
| No Locale with UTF8 found 🚫 |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
require 'fileutils'

fastlane_version "1.104.0"

default_platform :ios

platform :ios do

  ############################## PRE ##############################

  before_all do
    update_fastlane
    import_shared_fastfile
    load_environment_variables
  end

  ######################### PUBLIC LANES ##########################

[...]

PR #8762, opened 2017-04-03
Bump podspec version fix

Antondomashnev:
### Checklist
- [ x I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid

### Description
The change is a update of regex in `VersionBumpPodspec` action to support non semantic versions such as `1.2.3.4.5`. The more detailed description is given here #8722 

### Motivation and Context
It solves the problem of setting a new version of Podspec while the previous version was not a semantic. 

The linked issue is #8722 

The spec was updated to test the action against a not semantic version.

### DIFF
--- a/fastlane/lib/fastlane/actions/version_bump_podspec.rb
+++ b/fastlane/lib/fastlane/actions/version_bump_podspec.rb
@@ -15,7 +15,7 @@ module Fastlane
         if params[:version_number]
           new_version = params[:version_number]
         elsif params[:version_appendix]
-          new_version = version_podspec_file.set_version_appendix(params[:version_appendix])
+          new_version = version_podspec_file.update_version_appendix(params[:version_appendix])
         else
           new_version = version_podspec_file.bump_version(params[:bump_type])
         end
--- a/fastlane/lib/fastlane/actions/version_bump_podspec.rb
+++ b/fastlane/lib/fastlane/actions/version_bump_podspec.rb
@@ -14,6 +14,8 @@ module Fastlane
 
         if params[:version_number]
           new_version = params[:version_number]
+        elsif params[:version_appendix]
+          new_version = version_podspec_file.set_version_appendix(params[:version_appendix])
         else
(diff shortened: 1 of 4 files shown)


Antondomashnev:
I signed it!

ohayon (changes requested):
Hey @Antondomashnev, this is a great PR! Thanks for adding/modifying all of the tests! One question: Should we have a test for bumping the newly-added `appendix` number as well?

Antondomashnev:
> Should we have a test for bumping the newly-added appendix number as well?

Hey, @ohayon thanks for the review it's a good point, but I'm not sure if this is a valuable addition. Because `appendix` is supposed to be any suffix from in version after the `PATCH` part. So imagine the edge case example: `1.2.3.4.5.6` and here appendix is `4.5.6`, to my mind the developer will unlikely decide to change it or if she wants eventually I think it's easier to just set the version instead of bump. Don't you agree? 😄

ohayon:
@Antondomashnev I do agree for sure, but I also think that because we are exposing the appendix in the regex, we should make sure that if someone does try to do it programmatically that they are successful, maybe its less of a "bump" and more of a setting of the appendix. Curious to hear your thoughts! 🐙

Antondomashnev:
@ohayon I'm afraid if we eventually make it's less clear for the developers if it's not there yet 😄 
What if we add an exception when you want to `bump` an `appendix` and add an additional method to set an `appendix`?

ohayon:
@Antondomashnev thats a great idea! 🚀

Antondomashnev:
@ohayon you can check, now we have a `set_version_appendix` which basically allows a developer to change appendix, but it was developed with the `bump_version` logic in mind, so the developer has to call `update_podspec` to apply changes. It seems better now 😉

ohayon (approved):

ohayon:
Thanks for the PR @Antondomashnev, this is great!

PR #8764, opened 2017-04-04
Corrected an error message which is shown when a CD session cookie is expired.

tbrand:
### Checklist

### Description
Corrected an error message which is shown when a CD session cookie is expired.

### Motivation and Context
When we do CD on our CI servers while a session cookie is expired, we currently see an error message like "If you're running this in a non-interactive session (e.g. server or CI)..." which appears to be incorrect. So I’ve corrected the error message.

### DIFF
--- a/spaceship/lib/spaceship/two_step_client.rb
+++ b/spaceship/lib/spaceship/two_step_client.rb
@@ -48,6 +48,14 @@ module Spaceship
       puts "If you're running this in a non-interactive session (e.g. server or CI)"
       puts "check out #{two_factor_url}"
 
+      if !File.exist?(persistent_cookie_path) && self.class.spaceship_session_env.to_s.length.zero?
+        puts "If you're running this in a non-interactive session (e.g. server or CI)"
+        puts "check out #{two_factor_url}"
+      else
+        # If the cookie is set but still required, the cookie is expired
+        puts "Your session cookie has been expired."
+      end
+
       security_code = response.body["securityCode"]
       # {"length"=>6,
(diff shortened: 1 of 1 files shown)


ohayon (approved):
Thanks for the PR @tbrand! 🚀

spaceship/lib/spaceship/two_step_client.rb
```diff
@@ -45,8 +45,14 @@ def handle_two_step(response)
     def handle_two_factor(response)
       two_factor_url = "https://github.com/fastlane/fastlane/tree/master/spaceship#2-step-verification"
       puts "Two Factor Authentication for account '#{self.user}' is enabled"
-      puts "If you're running this in a non-interactive session (e.g. server or CI)"
-      puts "check out #{two_factor_url}"
+
+      if !File.exist?(persistent_cookie_path) && self.class.spaceship_session_env.to_s.length.zero?
```
janpio: I know it has been a long time, but:
Does this line make any sense?

I read it as:

```
If `persistent_cookie_path` does not exist AND session env var is 0 characters long (= empty)
   show information
else  
   "Your session cookie has been expired."
```

Doesn't this also mean that as soon as the session env var is set, you _always_ get "Your session cookie has been expired." here? Or how does the checking for an expired session actually work?

janpio: I don't this works at all - so reverted in https://github.com/fastlane/fastlane/pull/13743 where I also added a hopefully working check.

PR #8769, opened 2017-04-04
Use all commits instead of current HEAD in `number_of_commits`.

### LINKED ISSUE #5376: number_of_commits issue when using branches
Would number_of_commits be better using `git rev-list --all --count` rather than `git rev-list HEAD --count`?

I find the current version does not work so well when using branches.

samuelbeek:
### Checklist

### Description
Fixes https://github.com/fastlane/fastlane/issues/5376. 

### Motivation and Context
The documentation described the `number_of_commits` command as `The total number of all commits in current git repo`. But it was the total number of commits on the current branch. This PR fixes that.

### DIFF
--- a/fastlane/lib/fastlane/actions/number_of_commits.rb
+++ b/fastlane/lib/fastlane/actions/number_of_commits.rb
@@ -10,7 +10,10 @@ module Fastlane
 
       def self.run(params)
         if is_git?
-          command = 'git rev-list --all --count'
+          if params[:all]
+            command = 'git rev-list --all --count'
+          else
+            command = 'git rev-list HEAD --count'
         else
           UI.user_error!("Not in a git repository.")
         end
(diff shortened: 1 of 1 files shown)


samuelbeek:
I signed it!

mfurtak:
I think changing the behavior of the action isn't something we can do at this point, as some number of people are likely already dependent on it. Would it make sense to correct the documentation instead?

KrauseFx:
I agree with @mfurtak, adding a parameter to this action might be a good idea. The key is to not change any default behavior, which would break existing setups.

samuelbeek:
That makes sense, I will update my PR accordingly.

hjanuschka (changes requested):
can you please add the `available_options` method to the action with the `:all` option listed?

`optional` => `true`
`is_string` => `false`

example:
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/adb.rb#L27

samuelbeek:
@hjanuschka thanks for the comments, it's fixed now!

hjanuschka (approved):

hjanuschka:
@samuelbeek  many thx.


tested with:
`fastlane run number_of_commits` (expected behaviour on current users)
results in: `git rev-list HEAD --count`


`fastlane run number_of_commits all:true`
results in: `git rev-list --all --count`


many thx for your contribution 🚀

PR #8776, opened 2017-04-05
[pilot] tester management on less-priviliged accounts

hjanuschka:
supports adding testers on accounts below admin role
(developer, app-manager)

Fixes: #8728

### DIFF
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -46,6 +46,8 @@ module Pilot
              UI.success("Existing tester #{tester.email}")
            else
              attrs = {}
+             config[:first_name] ||= ""
+             config[:last_name] ||= ""
              attrs["emailAddress"] = { "value" => config[:email] }
              attrs["firstName"] = { "value" => config[:first_name] }
              attrs["lastName"] = { "value" => config[:last_name] }
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -19,35 +19,56 @@ module Pilot
       end
 
       begin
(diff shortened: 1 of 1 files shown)


hjanuschka:
@snatchev this solution seems to work, however i am not really happy with the factory thing on externalTesters, can you have a look, maybe there is a better solution. the root cause seems that per-se spaceship is adding testers to "overall" tester pool, where you have no access to if you are below  `admin` role.

however adding it to the app_id pool directly works.

bencollier:
Just testing this - added a tester perfectly 💪 but if no first name / last name is set the user appears as **null null**.

They're added fine, but on an Admin account it just shows them as empty instead of null. Cheers!

hjanuschka:
@bencollier awesome, thx - null value's fixed with last commit https://github.com/fastlane/fastlane/pull/8776/commits/e26a80ad52fce39d7be8dce92b8decf2bb4346c3

bencollier:
@hjanuschka looks good! Thanks :) I've not tested it as part of a CSV import, should that use the same logic?

hjanuschka:
@bencollier i think so, atleast it takes the same code-path 👍

tzm41:
When will this be merged? Seeing the error when I use the web app.

KrauseFx:
@hjanuschka This has to be re-written with the updated `pilot`, right? @taquitos actually works on the same thing for _pilot_, so you two might want to sync on this

tzm41:
Thanks for following up!

taquitos:
Update: I'm currently working through feedback on: https://github.com/fastlane/fastlane/pull/9046

KrauseFx:
Alright, cool, should we close this PR for now then?

hjanuschka:
yeah i think!

PR #8789, opened 2017-04-05
Recommend actions when name can't be found

Lutzifer:
### Checklist

### Description
When running 'fastlane actions xy' without a hit, there now is an additional line at the end with similar actions, e.g. for 'fastlane actions xy':

> Available actions with matching name: hg_add_tag, hg_commit_version_bump, hg_ensure_clean_status, hg_push

### Motivation and Context
This makes it easier to find e.g. all "tag" or "git" related actions, providing a nice shortcut for the user. I like it easy.

### DIFF
--- a/fastlane/lib/fastlane/documentation/actions_list.rb
+++ b/fastlane/lib/fastlane/documentation/actions_list.rb
@@ -81,7 +81,7 @@ module Fastlane
           all_actions(nil) do |a, action_name|
             candidates << [action_name] if action_name.include? filter
           end
-          puts "Available actions with similar name: #{candidates.join(', ')}".green unless candidates.empty?
+          puts "Available actions with matching name: #{candidates.join(', ')}".green unless candidates.empty?
         end
 
       end
--- a/fastlane/lib/fastlane/documentation/actions_list.rb
+++ b/fastlane/lib/fastlane/documentation/actions_list.rb
@@ -77,11 +77,21 @@ module Fastlane
         print_all # show all available actions instead
 
         if !filter.nil? && filter.length > 1
-          candidates = []
+          action_names = []
           all_actions(nil) do |action_ref, action_name|
-            candidates << action_name if action_name.include? filter
+            action_names << action_name
           end
-          puts "Available actions with matching name: #{candidates.join(', ')}".green unless candidates.empty?
+
+          corrections = []
+
+          if !Gem.loaded_specs["did_you_mean"].nil? && Gem.loaded_specs["did_you_mean"].version >= Gem::Version.new('1.1.0')
+            spell_checker = DidYouMean::SpellChecker.new(dictionary: action_names)
+            corrections << spell_checker.correct(filter).compact
+          end
+
+          corrections << action_names.select { |name| name.include? filter }
+
+          puts "Did you mean: #{corrections.flatten.uniq.join(', ')}".green unless corrections.flatten.empty?
         end
 
       end
--- a/fastlane/lib/fastlane/documentation/actions_list.rb
+++ b/fastlane/lib/fastlane/documentation/actions_list.rb
@@ -95,7 +95,7 @@ module Fastlane
 
         corrections << action_names.select { |name| name.include? filter }
 
-        puts "Did you mean: #{corrections.flatten.uniq.join(', ')} ?".green unless corrections.flatten.empty?
+        puts "Did you mean: #{corrections.flatten.uniq.join(', ')}?".green unless corrections.flatten.empty?
       end
     end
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/documentation/actions_list.rb
```diff
@@ -73,7 +73,17 @@ def self.show_details(filter: nil)
       else
         puts "Couldn't find action for the given filter.".red
         puts "==========================================\n".red
+
         print_all # show all available actions instead
+
+        if !filter.nil? && filter.length > 1
+          candidates = []
+          all_actions(nil) do |a, action_name|
+            candidates << [action_name] if action_name.include? filter
```
mfurtak: It feels strange to me that we're making an Array of Arrays each containing a String. Is this correct/necessary?

Lutzifer: ah, that was a leftover from previous code, fixed it as it was no longer necessary.

fastlane/lib/fastlane/documentation/actions_list.rb
```diff
@@ -79,7 +79,7 @@ def self.show_details(filter: nil)
         if !filter.nil? && filter.length > 1
           candidates = []
           all_actions(nil) do |a, action_name|
```
hjanuschka: can we use something different than `a` here, i guess its meant to be the hash key?

Lutzifer:
"a" was formerly known as action, but rubocop went all "bad cop" on it for shadowing the existing action variable 😉

Von meinem iPhone gesendet

hjanuschka:
@Lutzifer it seems to be not the action name but the ref to the instance, how about changing `a` to `action_ref`? only `a` is just too confusing

hjanuschka:
@Lutzifer ruby >= 2.3 ships with the "DidYouMean" gem.


```ruby
a = DidYouMean::SpellChecker.new(dictionary: ['action1', 'action2'])
corrections =  a.correct("acton")
puts "Did you mean: #{corrections.join(',')}"
```

results in:
`Did you mean: action2,action1`


i think following paths are available:
  *  detect ruby >2.3
  * add https://github.com/yuki24/did_you_mean/ as dependency
    * it seems to be MIT


what do you guys think?

Lutzifer:
That looks nice. The public interface in your example is new to v1.1.0 though (https://github.com/yuki24/did_you_mean/releases/tag/v1.1.0) and needs at least ruby 2.4.0

KrauseFx:
`did_you_mean` is included in the default Ruby installation now?

hjanuschka:
> Ruby 2.3 and later ships with this gem and it will automatically be required when a Ruby process starts up. No special setup is required.


however my sample uses 2.4 version (got it from their repo), so we'd need to find a way to do it with the 2.3 interface.

Lutzifer:
as this is currently working for all versions and has additional recommendations on 2.4: can we get it merged or is the different behaviour a no-go?

Lutzifer:
todo: add recommendations for "fastlane run" command

fastlane/lib/fastlane/documentation/actions_list.rb
```diff
@@ -73,7 +73,29 @@ def self.show_details(filter: nil)
...
+    def self.print_suggestions(filter)
+      if !filter.nil? && filter.length > 1
+        action_names = []
+        all_actions(nil) do |action_ref, action_name|
+          action_names << action_name
+        end
+
+        corrections = []
+
+        if !Gem.loaded_specs["did_you_mean"].nil? && Gem.loaded_specs["did_you_mean"].version >= Gem::Version.new('1.1.0')
```
KrauseFx: Do we need `1.1.0`? I have `1.0.0` on my current Ruby installation.

KrauseFx: Ok, seems like it doesn't work with `1.0.0`

```
bundler: failed to load command: fastlane (/Users/fkrause/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bin/fastlane)
NameError: [!] uninitialized constant DidYouMean::SpellChecker
```

KrauseFx (commented):
Thanks for working this, this change looks awesome, I'll take over and finish some smaller things up so we can merge it 👍

KrauseFx (approved):
Just tried this, looks great, made one small change, it's ready now 👍  Thanks for your contribution

KrauseFx:
Could you confirm the CLA with the Google bot?

Lutzifer:
great. how do i do that exactly?

KrauseFx:
@Lutzifer Just say that you're ok with the CLA agreement and that this PR can be merged as a comment. This is just because the bot is confused as the PR is by multiple authors 👍

Lutzifer:
I am ok with the cla and the pr can be merged!

KrauseFx:
Beautiful, thanks 👍

Issue #8787, opened 2017-04-05
[Regression] Fails to pass launch parameters when run on Simulator


wryphonedev:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

When attempting to run associated test suite in Xcode or snapshot from CLI, progress fails because launch arguments including device language and locale are not passed.

Console output indicates that a resource was not loaded because a proper scheme was not specified. 

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

Snapshot console output:
Couldn't detect/set language...
Couldn't detect/set locale...
Couldn't detect/set launch_arguments...

General stack trace with error printed:
2017-04-05 15:30:11.087 XCTRunner[86739:2074888] CFURLCopyResourcePropertyForKey failed because it was passed an URL which has no scheme
Printing description of error:
Error Domain=NSCocoaErrorDomain Code=256 "The file “language.txt” couldn’t be opened." UserInfo={NSURL=/Users/ischmidt15/Library/Caches/tools.fastlane/language.txt}

I was able to modify the SnapshotHelper.swift method -pathPrefix to return a Url initialized as a fileURL with the existing path, which resolved this issue. I can submit a PR.

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.12.3                                                        |
| Ruby                        | 2.0.0                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.10.1 (Apple Git-78)                              |
| Installation Source         | /usr/local/bin/fastlane                                        |
| Host                        | Mac OS X 10.12.3 (16D32)                                       |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib |
| OpenSSL Version             | OpenSSL 0.9.8zc 19 Mar 2015                                    |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/                    |
[...]

PR #8791, opened 2017-04-05
Fixes failure to pass launch arguments: Adds scheme to fileURL in SnapshotHelper running on Simulator

wryphonedev:
Returns a fileURL with the existing path rather then returning the or…iginal URL, which lacks a scheme.

### Checklist

### Description

In Snapshot's SnapshotHelper.swift, I modified the pathPrefix method to return a file scheme URL rather then a url without a scheme when running on the simulator. This fixes an issue where required launch argument resources could failed to be located in the filesystem when running on the simulator.

### Motivation and Context

After upgrading my Fastlane Snapshot today, I found that none of my associated UITests would complete. I isolated the and described the issue in #8787.

This changes solves the inability to run snapshot in the simulator with launch arguments.

#8787 

Snapshot now runs properly in my project.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -145,7 +145,7 @@ open class Snapshot: NSObject {
                 print("Can't prepare environment. Simulator home location is inaccessible. Does \(simulatorHostHome) exist?")
                 return nil
             }
-            homeDir = homeDirUrl
+            homeDir = URL(fileURLWithPath: homeDirUrl.path)
         #endif
         return homeDir.appendingPathComponent("Library/Caches/tools.fastlane")
     }


wryphonedev:
Signed CLA

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -145,7 +145,7 @@ open class Snapshot: NSObject {
                 print("Can't prepare environment. Simulator home location is inaccessible. Does \(simulatorHostHome) exist?")
                 return nil
             }
-            homeDir = homeDirUrl
+            homeDir = URL(fileURLWithPath: homeDirUrl.path)
```
ohayon: Can we change line 144 to say `guard let homeDirUrl = URL(fileURLWithPath: simulatorHostHome)...` instead of making a URL here with another URL's path?

ohayon (dismissed):
This is a good find, but I think we could make it a bit cleaner with the comment I left. Let me know your thoughts!

snapshot/lib/assets/SnapshotHelper.swift
wryphonedev: What you describe would have been my preference as well, the problem is that `fileURLWithPath `doesn't return an optional, so we loose the safety.

wryphonedev:
@ohayon Any further thoughts on this? Would love to get it pulled in as I'm being prompted to update my helper.

KrauseFx (approved):
Awesome, thanks for your contribution @wryphonedev 
🎷

gagaJay:
Please update the snapshotHelper file version also, otherwise we wouldn't be aware of that. Except for that, nice fix!

KrauseFx:
Ah yeah, nice catch @gagaJay, thanks 👍

PR #8799, opened 2017-04-06
[WIP] Improve device matching so it includes sims with (xyz) in the names

lyndsey-ferguson:
### Checklist

### Description

Change the regex for device manager so that it can find simulators that have parenthesis `(` in the name.

### Motivation and Context

We were naming our simulators on our CI slaves, `iPhone 5s (9.1)`, `iPhone 7 (10.2)`, etc. so we could differentiate between them. `scan` was unable to find any devices because none of those matched in device manager's regex.

I did not open an issue.

_pending_: I tested by specifying my branch of fastlane in our build system and running `scan` with the simulators in the name.


lyndsey-ferguson:
I signed it!

lyndsey-ferguson:
Okay, I cannot do this unless I specifically allow simulators with parenthesis in their names when `scan` tries to detect the values and that feels too unreliable.

PR #8825, opened 2017-04-08
Add configuration option for slather action

thasegaw:
### Checklist

### Description
[slather-2.4.1](https://github.com/SlatherOrg/slather/releases/tag/v2.4.1) is added `configuration` option. 

### Motivation and Context
Support slather-2.4.1


mgrebenets:
@thasegaw 

I'm not sure how, but is it possible to have some kind of version check for this new option?
I means it should check that `slather` is installed first (which I believe is already the case), but then it needs to make sure that version is at least `2.4.1` if `configuration` option is used.

Probably that could be part of configuration option verify block.

thasegaw:
@mgrebenets 
I added the verification whether slather version is at least 2.4.1 or not. I added it in` self.validate_params!` because I have to check `use_bundle_exec` parameter. Colud you let me know, If you have good idea. Thanks

thasegaw:
@KrauseFx Would you check this RP? I want to use this option.

fastlane/lib/fastlane/actions/slather.rb
```diff
@@ -52,7 +53,22 @@ def self.has_config_file
         File.file?('.slather.yml')
       end
 
+      def self.slather_version(params)
+        command = []
```
KrauseFx: Instead of shelling out to receive the version number, can't we do something like
```ruby
require 'slather'
Slather::VERSION
```

KrauseFx (approved):
Awesome, thanks for updating the PR, and thanks for working on this 🚀 This will be in the next release 👍

PR #8809, opened 2017-04-07
Make all tables in fastlane more responsive to the terminal width

KrauseFx:
Moved from https://github.com/fastlane/fastlane/pull/7391 by mistake

The design on the top table is the one that's used
screen shot 2017-04-06 at 7 06 57 pm

PR #8830, opened 2017-04-08
Make all tables in fastlane more responsive to the terminal width

hjanuschka:
moved from: https://github.com/fastlane/fastlane/pull/8809 - which moved from: https://github.com/fastlane/fastlane/pull/7391  and now fixes tests.

screen shot 2017-04-06 at 7 06 57 pm


KrauseFx (approved):
That's awesome, thanks for fixing the tests, let's see if that works 🚀

PR #8831, opened 2017-04-08
Fix LoadError catching when it's not related to CommandsGenerator

KrauseFx:
We rescue `LoadError` so that we tell users to run the tool directly if CommandsGenerator is not available. However since we call `.start` inside that `begin` `rescue` block we'd catch all LoadErrors from within fastlane, and then show the confusing error message of

```
pilot can't be called via `fastlane pilot`, run 'pilot' directly instead
```

### DIFF
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -48,13 +48,14 @@ module Fastlane
             require File.join(tool_name, "commands_generator")
 
             # Call the tool's CommandsGenerator class and let it do its thing
-            Object.const_get(tool_name.fastlane_module)::CommandsGenerator.start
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/cli_tools_distributor.rb
```diff
@@ -48,13 +48,14 @@ def take_off
             require File.join(tool_name, "commands_generator")
 
             # Call the tool's CommandsGenerator class and let it do its thing
-            Object.const_get(tool_name.fastlane_module)::CommandsGenerator.start
+            commands_generator_ref = Object.const_get(tool_name.fastlane_module)::CommandsGenerator
```
taquitos: nit: adding "_ref" seems redundant

taquitos (approved):
Tiny name nit, but 👍

hjanuschka (approved):
🚀

PR #8835, opened 2017-04-09
Update Actions.md plugin generation to include plugins with not high usage numbers

KrauseFx:
With a recent change in enhancer we excluded many actions to only show the top actions. However for generating the docs, we want to include all the plugins


nafu (approved):
Looks good to me 🤖 

Related enhancer changes for reference:
https://github.com/fastlane/enhancer/pull/52

KrauseFx:
Thanks @nafu 👍

PR #8836, opened 2017-04-09
Setup fastlane train for fastlane deployment

KrauseFx:
Show realtime deployment in Slack room, built via plugin: https://github.com/KrauseFx/fastlane-plugin-slack_train

This allows us to see when someone in the team has a deployment process running (which takes about 10 minutes), so that no one else merges anything into the master branch. If something is merged while something is deployed, it might not show up in the changelog.

If you feel like it pollutes the Fastfile too much, we can also remove the PR, or simplify it to be less steps.

screenshot 2017-04-09 04 20 29

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -59,18 +59,24 @@ end
 desc "Does everything that's needed for a release"
 desc "This includes running tests and verifying the GitHub release"
 lane :release do
+  slack_train_start(distance: 13,
+                       train: "🚀",
+           reverse_direction: true,
+                        rail: "✨")
+
   update_fastlane
+  slack_train
 
   # Git verification
   #
-  ensure_git_status_clean
+  # ensure_git_status_clean
   ensure_git_branch(branch: 'master')
   git_pull
 
-  # Verifying RubyGems version
-  #
   validate_repo
 
+  # Verifying RubyGems version
+  #
   require "../fastlane/lib/fastlane/version.rb"
   version = Fastlane::VERSION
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -59,18 +59,24 @@ end
...
+           reverse_direction: true,
+                        rail: "✨")
+
   update_fastlane
+  slack_train
 
   # Git verification
   #
-  ensure_git_status_clean
+  # ensure_git_status_clean
```
mfurtak: Is this intended to be commented out?

mfurtak:
13 consecutive messages into the Slack room feels like it will be very noisy. Into what room were you imagining this would go?

fastlane/Fastfile
KrauseFx: Ha, nice catch 👍

KrauseFx:
@mfurtak more than happy to update it to 2 or 3 steps, is that something we want to do? 👍 We can post it into #action room on Slack

mfurtak (approved):
I am happy to try it as-is, but let's follow up with the folks in that room and get some feedback :+1:

fastlane/Fastfile
mfurtak: Feel free to ship once returned ↩️

KrauseFx:
@mfurtak I updated the PR to only print out 2 messages, instead of thousands :)

mfurtak (approved):

PR #8888, opened 2017-04-14
push_to_git_remote default tags to `false`

d-date:
### Checklist

### Description
the default `tags` parameter which is one of the `push_to_git_remote` was changed from `true` to `false`.

### Motivation and Context

Sometimes `push_to_git_remote` is used for version bumping to upload adhoc binary to Fabric/Crashlytics with different build version. But when run `push_to_git_remote` with no option, the `tag` will be pushed automatically for remote. It causes action failed when same tag is already exist on remote repo, and It's difficult to notice the `tag` option is false on default for beginners.

You can check this change on running `push_to_git_remote` with no `tags` parameter.


hjanuschka (changes requested):
hey 🤚 
thx for your contribution, and nice catch, however i think we cannot simply change a default parameter - as it may break or lead to unexpected behaviour on users already using this action.

do you see any other solution, like catching the error and printing a nice description text on how to solve it by setting the value of `tags`  to `false`?

d-date:
@hjanuschka 
Hi 😀
Thank you for your comment, and I see your point.
I think it is natural with no tag pushing and it is happened no error with no tag pushing,
so how about notify to user who is no using `tags` param like below?

```
"[Info] When you want to push to git with tag,`push_to_git_remote` requires `tags: true` on your Fastfile from v2.xx.x." 
```
 (I'm Japanese and not good at English, so if you have better comment, please tell me.)

fastlane/lib/fastlane/actions/push_to_git_remote.rb
```diff
@@ -52,9 +52,9 @@ def self.available_options
                                        default_value: false),
           FastlaneCore::ConfigItem.new(key: :tags,
                                        env_name: "FL_PUSH_GIT_TAGS",
-                                       description: "Wether tags are pushed to remote. Defaults to true",
+                                       description: "Wether tags are pushed to remote. Defaults to false",
```
giginet: I found typo 👀 

`Whether`.

giginet (commented):
I have no idea that should we modify default behavior 🤔 

I think we can merge this as long as we inform this changes.

fastlane/lib/fastlane/actions/push_to_git_remote.rb
d-date: @giginet thx, so I made another PR. https://github.com/fastlane/fastlane/pull/8889

d-date:
@hjanuschka  @giginet OK. Thank you for your review, so I'll show messege for user who are using `tags` as default. 

```
"[INFO] If you don't want to push with tag, please specify `tags` to `false` on your Fastfile."

Then, we should discuss about default behavior on issue (I'll make it later)

How about this one?

giginet:
@KrauseFx How do you think about?

nafu (changes requested):
Thank you @d-date for contribution 👋 

The default behaivior of tags is at least before monorepo.
Changing the default behavior has big impact on existing users.

I think it would  be better to

1. Output message to inform 'tags: false' if tags error occur (same as @hjanuschka said)
2. Push again with 'tags: false' if the user want

hjanuschka:
@nafu but keep in mind that "if user wants" needs to be a solution that works even if the user does not have a interactive terminal (e.g: CI)

nafu:
@hjanuschka Rightt 👍

KrauseFx (changes requested):
I'm also skeptical on changing the default behavior, we should try to avoid this. I'd be ok with showing a message when the command fails that the parameter is available 👍

d-date:
@KrauseFx @hjanuschka @nafu @giginet
OK. Thank you for your comment.
I'lol try to show message💪

ohayon:
Hey @d-date, wanted to check in on the status of this PR, it looks like there was a merge commit that caused a lot of files to change that are not relevant to your change. Would you mind cleaning that up so it is just your change and that way we can have a more clear discussion and decide if it should be merged? Thanks! 🚀 🐙

KrauseFx:
Hey @d-date, thanks for your contribution 👍  I'll close the PR for now, as there are some changes included that probably shouldn't be part of this PR. Feel free to submit a new PR any time ❤️

PR #8892, opened 2017-04-15
Correct usage of screengrab

vanniktech:
Just `screengrab` is not found.


mfurtak (approved):
Thanks for this fix :+1:

Issue #8729, opened 2017-03-31
Gym archives are placed in hardcoded archive path


gradha:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Originated as a [StackOverflow question](http://stackoverflow.com/questions/43114922/how-can-i-add-fastlane-builds-to-xcodes-archive-organizer/43124229#43124229), archives created by gym are placed under `~/Library/Developer/Xcode/Archives/[date]`. This doesn't work for users who have Xcode configured to a custom archive path, the organizer won't show the fastlane archives.

In the meantime I've resorted to using a symbolic link to the custom path, since I don't want to store `build_path` or `archive_path` values in the git repository specific to this machine.

KrauseFx:
Hey @gradha, that's correct. However _gym_ should still work as expected, it will just not show the archive. The problem is that we couldn't find a way to detect the user's config for the archives. If you know about a CLI option to get that, it would be a great thing to include into _fastlane_.

gradha:
Haven't seen any option either. The path can be obtained running `/usr/libexec/PlistBuddy -c "Print IDECustomDistributionArchivesLocation" ~/Library/Preferences/com.apple.dt.Xcode.plist`. When the user has not configured any the error `Print: Entry, "IDECustomDistributionArchivesLocation", Does Not Exist` is printed instead.

Obviously hacky, I haven't tracked across xcode versions how stable this private setting is.

KrauseFx:
Mh, that looks good. Are you interested in contributing to _fastlane_ to submit a PR with this change? We have some good guides on how to get started: https://github.com/fastlane/fastlane/blob/master/YourFirstPR.md

gradha:
Highly unlikely, I don't grok ruby.

KrauseFx:
Okay, no worries. Maybe someone else is interested in putting up a PR for this change 👍

KrauseFx:
We're working on this now 👍

PR #8898, opened 2017-04-18
Read archive path from Xcode preferences if set

taquitos:
- Detect when user has Xcode preferences and load the IDECustomDistributionArchivesLocation

### Checklist

### Description
Respects your Xcode preferences for Archive path

### Motivation and Context
Fixes https://github.com/fastlane/fastlane/issues/8729


gym/lib/gym/detect_values.rb
```diff
@@ -1,3 +1,4 @@
+require 'cfpropertylist'
```
KrauseFx: Looks like that gem is actually not a direct dependency of fastlane, but of `xcodeprojc`, we should add it to `fastlane.gemspec` to be more explicit in case the dependency is dropped one day.

gym/lib/gym/detect_values.rb
```diff
@@ -33,9 +34,34 @@ def self.set_additional_default_values
 
       config[:output_name] ||= Gym.project.app_name
 
+      if Gym.config[:build_path].nil? # maybe we manually passed it in somewhere?
+        config[:build_path] = archive_path_from_local_xcode_preferences
```
KrauseFx: Or replace all those lines with
```ruby
config[:build_path] ||= archive_path_from_local_xcode_preferences
```

gym/lib/gym/detect_values.rb
```diff
@@ -33,9 +34,34 @@ def self.set_additional_default_values
...
+      path = File.expand_path(Gym.config[:xcode_preference_plist_path]) if Gym.config[:xcode_preference_plist_path].length > 0
+
+      unless File.exist?(path.to_s) # this file only exists when you edit the Xcode preferences to set custom values
+        return archive_path
+      end
+
+      xcode_preferences_dictionary = CFPropertyList.native_types(CFPropertyList::List.new(file: path).value)
+      custom_archive_path = xcode_preferences_dictionary['IDECustomDistributionArchivesLocation']
+
+      unless custom_archive_path.to_s.length > 0
```
KrauseFx: Could be
```ruby
return archive_path if custom_archive_path.to_s.length == 0
```

gym/lib/gym/options.rb
```diff
@@ -228,6 +228,12 @@ def self.plain_options
                                      verify_block: proc do |value|
                                        UI.user_error!("File not found at path '#{File.expand_path(value)}'") unless File.exist?(value)
                                      end),
+        FastlaneCore::ConfigItem.new(key: :xcode_preference_plist_path,
```
KrauseFx: Why do we need this option? We usually try to avoid adding options as much as possible, and only add ones that are helpful to a majority of our users 👍

KrauseFx (changes requested):
This is great, I added some comments before it's ready 👍

gym/lib/gym/detect_values.rb
taquitos: Good catch

gym/lib/gym/detect_values.rb
taquitos: 👍

gym/lib/gym/detect_values.rb
taquitos: 👍

gym/lib/gym/options.rb
taquitos: We don't! I didn't realize you could do all this wonderful stubbing in Ruby so we don't need to inject anything via the options.
#rubyNoob

fastlane.gemspec
```diff
@@ -36,6 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'terminal-notifier', '>= 1.6.2', '< 2.0.0' # macOS notifications
   spec.add_dependency 'terminal-table', '>= 1.4.5', '< 2.0.0' # Actions documentation
   spec.add_dependency 'plist', '>= 3.1.0', '< 4.0.0' # Needed for set_build_number_repository and get_info_plist_value actions
+  spec.add_dependency 'CFPropertyList', '~> 2.3.3' # Needed to be able to read binary plist format
```
KrauseFx: I wouldn't be too strict about this, maybe have `'>= 2.3', '< 3.0.0'` instead. If `xcodeproj` bumps their dependency, it would result in unresolvable dependencies otherwise

taquitos: 👍

KrauseFx (approved):
Wohoo 🎷

PR #8913, opened 2017-04-19
-showBuildSettings does not require to call clean

### LINKED ISSUE #8099: Clean parameter ignored during build settings check
`xcodebuild -showBuildSettings -workspace (workspace) -scheme (scheme)`

was recently replaced with:

`xcodebuild clean -showBuildSettings -workspace (workspace) -scheme (scheme)`

in https://github.com/fastlane/fastlane/pull/5626

This seems to clean the target even if the argument -clean false is passed in as a parameter. I'm not sure what a solution would look like. Perhaps the retry in https://github.com/fastlane/fastlane/pull/5188 is sufficient?

Thanks to all contributors!

punty:
### Checklist
- [x ] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [ x] I've updated the documentation if necessary.

### Description
In show building settings, don't call xcodebuild clean, because it's not a necessary workaround anymore. 

### Motivation and Context
This change will allow the user to do not clean the build (unless specified in the following commands). 
Previously for example event setting clean: false, will result in callling clean because of the generated  
xcodebuild clean -showBuildSettings

As stated in the comment this was a workaround for an xcodebuild bug. But this bug has being fixed in 8.3 so it's not necessary anymore. https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html
As huge benefit, I can keep using fastlane with my CI, and for example, caching pod builds, and reduce my build time significantly. 
 
This pull request also fixes https://github.com/fastlane/fastlane/issues/8099

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -277,11 +277,7 @@ module FastlaneCore
     #####################################################
 
     def build_xcodebuild_showbuildsettings_command
-      # We also need to pass the workspace and scheme to this command.
-      #
-      # The 'clean' portion of this command is a workaround for an xcodebuild bug with Core Data projects.
-      # See: https://github.com/fastlane/fastlane/pull/5626
-      command = "xcodebuild clean -showBuildSettings #{xcodebuild_parameters.join(' ')}"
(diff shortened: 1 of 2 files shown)


punty:
I signed it!

punty:
I signed it!

PR #8919, opened 2017-04-19
[match] add `disable_encryption` option to skip encryption

hjanuschka:
when used with many many app's 1000+ match seems to be slow - and git repo's grow way to much - per run.

for the normal user this should never be used, but from a enterprise point of view - it seems to be a requirement to atleast have the option to turn off encryption.

Related: #8793 and #8917

call with:

```ruby
match(disable_encryption: true)
```

```bash
fastlane match decrypt -r ssh://git@gitlab.xx.at/xx/IOS-xx.git

# [07:47:21]: Cloning remote git repo...
# [07:47:26]: 🔓  Successfully decrypted certificates repo
# [07:47:26]: Repo is at: '/var/folders/x0/jrjwfzy91k584k7jxr4x17p1m2727j/T/d20170420-91866-19p6baz'

# See the Path that match prints out
cd /var/folders/x0/jrjwfzy91k584k7jxr4x17p1m2727j/T/d20170420-91866-19p6baz

# for easy revertability  - tag current state
git tag before_decrypt

# commit decrypted files.
git commit -a -m "decrypted it 💣"

# push decrypted repo
git push

# push backup tag
git push --tags
```

### Using Current repo state
as there is no built in mechanism in match to encrypt an un-enctrypted repo, put `crypt_helper.rb` somewhere.

crypt_helper.rb:
```ruby
require 'security'
require 'tmpdir'
require 'shellwords'
module Match
  class CryptHelper
    def encrypt_repo(path: nil, password: nil)
      iterate(path) do |current|
        crypt(path: current,
              password: password,
              encrypt: true)
        puts "🔒  Encrypted '#{File.basename(current)}'"
      end
      puts '🔒  Successfully encrypted certificates repo'
    end

    def iterate(source_path)
      Dir[File.join(source_path, '**', '*.{cer,p12,mobileprovision}')].each do |path|
        next if File.directory?(path)
        yield(path)
      end
    end

    def crypt(path: nil, password: nil, encrypt: true)
      tmpfile = File.join(Dir.mktmpdir, 'temporary')
      command = ['openssl aes-256-cbc']
      command << "-k #{password.shellescape}"
      command << "-in #{path.shellescape}"
      command << '-a'
      command << '-d' unless encrypt
      command << '&> /dev/null'

      raise "Error decrypting '#{path}'" unless success
      FileUtils.mv(tmpfile, path)
    end
  end
end

helper = Match::CryptHelper.new
helper.encrypt_repo(path: ARGV[0], password: ARGV[1])
```

```
# clone it
git clone ssh://git@gitlab.xx.at/xx/IOS-xx.git

# cd into it
cd IOS-xx

# crypt it again
ruby crypt_helper.rb $PWD password_to_use # QUADRUPLE CHECK THIS IT that it is the right directory you are in 

# commit it
git commit -a -m "crypted again 🔐"

# push it back
git push
```

```bash
git clone  ssh://git@gitlab.xx.at/xx/IOS-xx.git
cd IOS-xx
git reset --hard before_decrypt
git push -f # 👊 force push! 
```

testing this PR:

```ruby
mkdir new_repo
cd new_repo
git init --bare
cd fastlane
bundle exec fastlane match development -r /path/to/new_repo/ -a krone.at-stickertest
bundle exec fastlane match development -r /path/to/new_repo/ -a krone.at-stickertest --disable_encryption # fails :)

mkdir new_repo
cd new_repo
git init --bare
cd fastlane
bundle exec fastlane match development -r /path/to/new_repo/ -a krone.at-stickertest --disable_encryption
bundle exec fastlane match development -r /path/to/new_repo/ -a krone.at-stickertest --disable_encryption # works :)


```


KrauseFx:
Interesting... before doing that, we should ask the user that reported the problem with the repo size if that would be a feasible solution to them, and if they'd use it.

adamtester:
@hjanuschka nice!
@KrauseFx for us, it would be fine, however large corporate companies would probably not be allowed to use it

I'll test it today and see how well it works and report back!

hjanuschka:
@adamtester thx for testing *BUT PLEASE KEEP IN MIND TO DO A BACKUP OF YOUR REPO BEFORE*

adamtester:
Looks good, didn't test on the CI server but created a similar setup on my mac and used a copy of the certificates repo.

GandjaFuzz:
When release?

hjanuschka:
@KrauseFx you see a chance of this getting merged?

match/lib/match/options.rb
```diff
@@ -55,6 +55,12 @@ def self.available_options
                                      description: "Only fetch existing certificates and profiles, don't generate new ones",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :disable_encryption,
+                                     env_name: "MATCH_DISABLE_ENCRYPTION",
+                                     short_option: "-d",
+                                     description: "Disable encryption, certificates including keys and profiles won't be encrypted! be sure that your git repo is safe!",
```
KrauseFx: `Be` should be upper case. Could you extend the description to mention here that this can't be changed once _match_ is set up?

KrauseFx: In general, I'm wondering if we can show an error message if you run _match_ without encryption, but for an encrypted repo, and the other way around. How does _match_ currently behave in that situation?

KrauseFx (changes requested):

hjanuschka (commented):
@KrauseFx typo's fixed.
and i added a way to detect encryption state and print out notice's if run is impossible due to encryption enabled or disabled.

also added specs for this. and updated PR body to provide a sample to test this pr without destroying a real repo 👍

match/lib/match/git_helper.rb
```diff
@@ -21,6 +23,29 @@ def self.clone(git_url, shallow_clone, manual_password: nil, skip_docs: false, b
         UI.user_error!("Error cloning certificates git repo, please make sure you have access to the repository - see instructions above")
       end
 
+      # if there is a encryption state marker file
+      if File.exist?(File.join(@dir, "match_crypted.txt"))
+        # repo is crypted but user requested to run match in disabled_encryption mode
+        # files cannot be read.
+        if GitHelper.crypted?(@dir) && disable_encryption == true
```
KrauseFx: Why the `== true`?

match/lib/match/git_helper.rb
```diff
@@ -21,6 +23,29 @@ def self.clone(git_url, shallow_clone, manual_password: nil, skip_docs: false, b
...
+          remote_is_crypted!
+        end
+
+        # repo is not crypted, but match is run with enabled encryption
+        if !GitHelper.crypted?(@dir) && disable_encryption == false
```
KrauseFx: Why the `== false` instead of `!disable_encryption`

match/lib/match/git_helper.rb
```diff
@@ -43,6 +68,12 @@ def self.clone(git_url, shallow_clone, manual_password: nil, skip_docs: false, b
       return @dir
     end
 
+    def self.remote_is_crypted!
+      UI.error("Encryption disabled, but remote repository is currently crypted.")
+      UI.error("See: https://github.com/fastlane/fastlane/pull/8919 for details on how to convert your existing repository")
+      UI.user_error!("remote_encrypted")
```
KrauseFx: `Encryption disabled, but remote repository is currently encrypted`

match/lib/match/git_helper.rb
```diff
@@ -54,6 +85,19 @@ def self.generate_commit_message(params)
       ].join(" ")
     end
 
+    def self.crypted?(workspace)
```
KrauseFx: `encrypted?`

KrauseFx (dismissed):

GandjaFuzz:
Do we guys ever see a release?)) Git grows very very much

KrauseFx:
Hey @GandjaFuzz, we're working on it, but first we want to be sure everything works perfectly 👍 
@hjanuschka Thanks for updating the PR so fast. There is 2 more comments from my side above that need to addressed before we can merge 🚀  Thanks

hjanuschka:
@KrauseFx for whatever reason i missed the comments, changes made - thx!

GandjaFuzz:
=)

taquitos (commented):
This looks great to me. I'd like one more person to review, though, as I'm not as familiar with this code. 
Great job at catching all the complexities around migration! 🥇

hjanuschka:
@taquitos thx, at first i thought we should ship the migration code, but i guess its to much code to maintain further on - and too large regression surface..

GandjaFuzz:
up

match/lib/match/encrypt.rb
```diff
@@ -48,7 +52,17 @@ def encrypt_repo(path: nil, git_url: nil)
       UI.success "🔒  Successfully encrypted certificates repo"
     end
 
-    def decrypt_repo(path: nil, git_url: nil, manual_password: nil)
+    def no_encryption_note
```
ohayon: nitpick: Can this be called something like `show_no_encryption_note`?

match/lib/match/git_helper.rb
```diff
@@ -43,6 +45,29 @@ def self.clone(git_url,
...
+          UI.user_error!("remote_decrypted")
+        end
+      else
+        # existing repo has `match_version.txt` - but no encryption state marker but unencrypted request
+        if File.exist?(File.join(@dir, "match_version.txt")) && disable_encryption
+          UI.error("You requested to disable encryption on a repo that is not up-to-date")
+          UI.error("Please run match atleast once with the current version")
+          remote_is_encrypted!
+        end
+      end
```
ohayon: This method is pretty long with a lot of special cases, could we extract some stuff out into separate, methods that could be tested? That might also allow us to remove the rubocop perceived complexity warning suppression as well.

match/lib/match/git_helper.rb
```diff
@@ -43,6 +45,29 @@ def self.clone(git_url,
         UI.user_error!("Error cloning certificates git repo, please make sure you have access to the repository - see instructions above")
       end
 
+      # if there is a encryption state marker file
+      if File.exist?(File.join(@dir, "match_crypted.txt"))
+        # repo is crypted but user requested to run match in disabled_encryption mode
```
ohayon: "repo is *en*crypted"

match/lib/match/git_helper.rb
```diff
@@ -43,6 +45,29 @@ def self.clone(git_url,
...
+        # files cannot be read.
+        if GitHelper.encrypted?(@dir) && disable_encryption
+          remote_is_encrypted!
+        end
+
+        # repo is not crypted, but match is run with enabled encryption
```
ohayon: "*en*crypted"

match/lib/match/git_helper.rb
```diff
@@ -43,6 +45,29 @@ def self.clone(git_url,
...
+        if !GitHelper.encrypted?(@dir) && !disable_encryption
+          UI.error("Encryption enabled, but remote repository is currently decrypted.")
+          UI.error("See: https://github.com/fastlane/fastlane/pull/8919 for details on how to convert your existing repository")
+          UI.user_error!("remote_decrypted")
```
ohayon: Can we just call `UI.user_error!" here with the message above?

match/lib/match/git_helper.rb
```diff
@@ -60,10 +85,17 @@ def self.clone(git_url,
...
+      Encrypt.new.decrypt_repo(path: @dir, git_url: git_url, manual_password: manual_password, disable_encryption: disable_encryption)
 
       return @dir
     end
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    def self.remote_is_encrypted!
+      UI.error("Encryption disabled, but remote repository is currently encrypted.")
+      UI.error("See: https://github.com/fastlane/fastlane/pull/8919 for details on how to convert your existing repository")
+      UI.user_error!("remote_encrypted")
```
ohayon: I think the message can just be put into this `UI.user_error!` call.

match/lib/match/git_helper.rb
```diff
@@ -76,19 +108,34 @@ def self.generate_commit_message(params)
       ].join(" ")
     end
 
+    def self.encrypted?(workspace)
+      path = File.join(workspace, "match_crypted.txt")
```
ohayon: Could this file be called `match_encrypted.txt`?

match/lib/match/git_helper.rb
```diff
@@ -76,19 +108,34 @@ def self.generate_commit_message(params)
...
+      # if file does not exist -> return true (default match behaviour)
+      return true unless File.exist?(path)
+      is_crypted = File.read(path).chomp
```
ohayon: `is_encrypted`

match/lib/match/git_helper.rb
```diff
@@ -76,19 +108,34 @@ def self.generate_commit_message(params)
...
-    def self.commit_changes(path, message, git_url, branch = "master")
+    def self.commit_changes(path, message, git_url, branch = "master", disable_encryption = false)
       Dir.chdir(path) do
         return if `git status`.include?("nothing to commit")
 
-        Encrypt.new.encrypt_repo(path: path, git_url: git_url)
+        Encrypt.new.encrypt_repo(path: path, git_url: git_url, disable_encryption: disable_encryption)
         File.write("match_version.txt", Fastlane::VERSION) # unencrypted
+        # store the state of encryption
+        File.write("match_crypted.txt", (!disable_encryption).to_s)
```
ohayon: `match_encrypted.txt`

ohayon (changes requested):
This is great @hjanuschka! I would still love to get @KrauseFx's 👀  on here one more time before merging. I left a bunch of comments mostly about spelling and some style. Looking forward to hearing your thoughts and getting this merged in!

GandjaFuzz:
up

dtweston:
@hjanuschka: Is there any help I can offer to get this PR finished up?

KrauseFx:
Okay okay, hey everybody, we'll talk more about this, since this is a problem we'll have to fix. Instead of disabling encryption all-together, maybe it makes more sense to just don't re-encrypt all files on every run? We could easily detect what files were modified, and only encrypt those again? This way, we're fully backwards compatible, and the repo is still fully encrypted. 

What do y'all think?

GandjaFuzz:
That sounds good

GandjaFuzz:
and need don't need decrypt and encrypt all files. need decrypt and encrypt only files needed for project. it is really make ?

KrauseFx:
Hey everyone, sounds good, I'll take on this task for this week, so I'll work on a cleaner solution that doesn't require un-encrypted repos :)

PR #8921, opened 2017-04-19
showBuildSettings does not require to call clean

### LINKED ISSUE #8099: Clean parameter ignored during build settings check

punty:
### Checklist
- [x ] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [ x] I've updated the documentation if necessary.

### Description
In show building settings, don't call xcodebuild clean, because it's not a necessary workaround anymore. 

### Motivation and Context
This change will allow the user to do not clean the build (unless specified in the following commands). 
Previously for example event setting clean: false, will result in callling clean because of the generated  
xcodebuild clean -showBuildSettings

As stated in the comment this was a workaround for an xcodebuild bug. But this bug has being fixed in 8.3 so it's not necessary anymore. https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html
As huge benefit, I can keep using fastlane with my CI, and for example, caching pod builds, and reduce my build time significantly. 
 
This pull request also fixes https://github.com/fastlane/fastlane/issues/8099

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -214,10 +214,10 @@ module FastlaneCore
     end
 
     # @return true if XCode version is higher than 8.3
-    def self.xcode_higher_than_8_3?
+    def self.xcode_atleast?(version)
       FastlaneCore::UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
       v = xcode_version
-      Gem::Version.new(v) >= Gem::Version.new('8.3.0')
+      Gem::Version.new(v) >= Gem::Version.new(version)
     end
 
     # @return the full path to the iTMSTransporter executable
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -214,7 +214,7 @@ module FastlaneCore
     end
 
     # @return true if XCode version is higher than 8.3
-    def self.xcode_atleast?(version)
+    def self.xcode_at_least?(version)
       FastlaneCore::UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
       v = xcode_version
       Gem::Version.new(v) >= Gem::Version.new(version)
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -213,6 +213,13 @@ module FastlaneCore
       keychain_path
     end
 
+    # @return true if XCode version is higher than 8.3
+    def self.xcode_higher_than_8_3?
+      FastlaneCore::UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
+      v = xcode_version
+      Gem::Version.new(v) >= Gem::Version.new('8.3.0')
+    end
+
     # @return the full path to the iTMSTransporter executable
     def self.itms_path
       return ENV["FASTLANE_ITUNES_TRANSPORTER_PATH"] if FastlaneCore::Env.truthy?("FASTLANE_ITUNES_TRANSPORTER_PATH")
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -277,11 +277,11 @@ module FastlaneCore
     #####################################################
 
     def build_xcodebuild_showbuildsettings_command
-      # We also need to pass the workspace and scheme to this command.
-      #
-      # The 'clean' portion of this command is a workaround for an xcodebuild bug with Core Data projects.
-      # See: https://github.com/fastlane/fastlane/pull/5626
-      command = "xcodebuild clean -showBuildSettings #{xcodebuild_parameters.join(' ')}"
+      if FastlaneCore::Helper.xcode_higher_than_8_3?
+        command = "xcodebuild -showBuildSettings #{xcodebuild_parameters.join(' ')}"
+      else
+        command = "xcodebuild clean -showBuildSettings #{xcodebuild_parameters.join(' ')}"
+      end
       command += " 2> /dev/null" if xcodebuild_suppress_stderr
       command
     end
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -277,7 +277,7 @@ module FastlaneCore
     #####################################################
 
     def build_xcodebuild_showbuildsettings_command
-      if FastlaneCore::Helper.xcode_higher_than_8_3?
+      if FastlaneCore::Helper.xcode_atleast?('8.3')
         command = "xcodebuild -showBuildSettings #{xcodebuild_parameters.join(' ')}"
       else
         command = "xcodebuild clean -showBuildSettings #{xcodebuild_parameters.join(' ')}"
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -277,7 +277,12 @@ module FastlaneCore
     #####################################################
 
     def build_xcodebuild_showbuildsettings_command
-      if FastlaneCore::Helper.xcode_atleast?('8.3')
+      # We also need to pass the workspace and scheme to this command.
+      #
+      # The 'clean' portion of this command was a workaround for an xcodebuild bug with Core Data projects.
+      # This xcodebuild bug is fixed in Xcode 8.3 so 'clean' it's not necessary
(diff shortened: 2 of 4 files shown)


fastlane_core/spec/project_spec.rb
```diff
@@ -299,7 +299,12 @@ def within_a_temp_dir
       it "SUPPORTED_PLATFORMS should be iphonesimulator iphoneos" do
         options = { project: "./fastlane_core/spec/fixtures/projects/Example.xcodeproj" }
         @project = FastlaneCore::Project.new(options, xcodebuild_list_silent: true, xcodebuild_suppress_stderr: true)
-        expect(FastlaneCore::Project).to receive(:run_command).with("xcodebuild clean -showBuildSettings -project ./fastlane_core/spec/fixtures/projects/Example.xcodeproj 2> /dev/null", { timeout: 10, retries: 3, print: false }).and_return(File.read("./fastlane_core/spec/fixtures/projects/build_settings_with_toolchains"))
+        if FastlaneCore::Helper.xcode_higher_than_8_3?
```
hjanuschka: i think this should be splitted up into 2 tests. with mocked return of `xcode_atleast` on time with `false` and one time with `true`  to be sure that it works on old and new xcodes.

the current variation is depending on the local xcode installation.

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -213,6 +213,13 @@ def self.keychain_path(name)
       keychain_path
     end
 
+    # @return true if XCode version is higher than 8.3
+    def self.xcode_higher_than_8_3?
```
hjanuschka: can we rename this to: `self.xcode_atleast?(version)`

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -213,6 +213,13 @@ def self.keychain_path(name)
...
+      FastlaneCore::UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
+      v = xcode_version
+      Gem::Version.new(v) >= Gem::Version.new('8.3.0')
```
hjanuschka: and use it here: `Gem::Version.new(v) >= Gem::Version.new(version)`

gym/lib/gym/xcode.rb
```diff
@@ -18,9 +18,7 @@ def pre_7?
       end
 
       def legacy_api_deprecated?
-        UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
-        v = xcode_version
-        Gem::Version.new(v) >= Gem::Version.new('8.3.0')
+        FastlaneCore::Helper.xcode_higher_than_8_3?
```
hjanuschka: then use `FastlaneCore::Helper.xcode_atleast?("8.3")`

hjanuschka (changes requested):
🚀  looking promising, left a couple of comments
many thx for your contribution @punty

punty:
@hjanuschka I implemented the changes you requested! Much better!

fastlane_core/spec/project_spec.rb
```diff
@@ -296,12 +296,24 @@ def within_a_temp_dir
...
         options = { project: "./fastlane_core/spec/fixtures/projects/Example.xcodeproj" }
         @project = FastlaneCore::Project.new(options, xcodebuild_list_silent: true, xcodebuild_suppress_stderr: true)
-        expect(FastlaneCore::Project).to receive(:run_command).with("xcodebuild clean -showBuildSettings -project ./fastlane_core/spec/fixtures/projects/Example.xcodeproj 2> /dev/null", { timeout: 10, retries: 3, print: false }).and_return(File.read("./fastlane_core/spec/fixtures/projects/build_settings_with_toolchains"))
+        allow(FastlaneCore::Helper).to receive(:xcode_atleast?).and_return(true)
+        command = "xcodebuild -showBuildSettings -project ./fastlane_core/spec/fixtures/projects/Example.xcodeproj 2> /dev/null"
+        expect(FastlaneCore::Project).to receive(:run_command).with(command.to_s, { timeout: 10, retries: 3, print: false }).and_return(File.read("./fastlane_core/spec/fixtures/projects/build_settings_with_toolchains"))
         expect(@project.build_settings(key: "SUPPORTED_PLATFORMS")).to eq("iphonesimulator iphoneos")
       end
+      unless FastlaneCore::Helper.xcode_atleast?('8.3')
```
hjanuschka: can we get rid of the unless, and mock the result of xcode_atleast to 8.1?
removing the unless and turning the `allow` to an `expect` should do it, the allow should also be changed in the above test. so we always no matter which local xcode is installed verify it works on 8.1 and 8.3

or i am i wrong?

thx for your patience.

hjanuschka:
@punty awesome, looking solid 👍  thx for your work and patience, one comment left.

fastlane_core/spec/project_spec.rb
punty: I am not sure about this. According to the official Apple documentation, the XCode bug has being fixed in 8.3, so if we are running this test, by mocking the call to return true, we will end up calling showBuildSettings without the clean command (even for XCode < 8.3). So in this case there is a risk this command will hangs? 
According to Apple this happens only if the project contains Core Data. But I still think we should include test for scenarios that are not possible.

punty: Allow is used for Mocking, https://www.relishapp.com/rspec/rspec-mocks/v/2-14/docs/method-stubs  so when I call that method i return true or false

hjanuschka: the command is not really run in the tests.
`expect(FastlaneCore::Project).to receive(:run_command).with(command.to_s, { timeout: 10, retries: 3, print: false }).and_return(File.read("./fastlane_core/spec/fixtures/projects/build_settings_with_toolchains"))` mocks it.

the goal should be to test the following two scenarios.

## user has xcode < 8.3
  *  force `xcode_atleast?("8.3") ` to return `false`
  * expect the resulting command, to include `clean`

## user has xcode >= 8.3
  *  force `xcode_atleast?("8.3") ` to return `true`
  * expect the resulting command, to NOT include `clean`

by forcing the result we bypass the local installation interfering the unit test.

punty: thanks for the clarification. I will have a look at that!

punty:
@hjanuschka thanks again for your suggestions! I implemented the changes you requested

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -213,6 +213,13 @@ def self.keychain_path(name)
       keychain_path
     end
 
+    # @return true if XCode version is higher than 8.3
+    def self.xcode_atleast?(version)
```
KrauseFx: `xcode_at_least?`

fastlane_core/spec/project_spec.rb
```diff
@@ -296,10 +296,21 @@ def within_a_temp_dir
     end
 
     describe "build_settings() can handle empty lines" do
-      it "SUPPORTED_PLATFORMS should be iphonesimulator iphoneos" do
+      it "SUPPORTED_PLATFORMS should be iphonesimulator iphoneos on XCode >= 8.3" do
```
KrauseFx: `Xcode`

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -277,11 +277,11 @@ def xcodebuild_parameters
     #####################################################
 
     def build_xcodebuild_showbuildsettings_command
-      # We also need to pass the workspace and scheme to this command.
-      #
-      # The 'clean' portion of this command is a workaround for an xcodebuild bug with Core Data projects.
-      # See: https://github.com/fastlane/fastlane/pull/5626
-      command = "xcodebuild clean -showBuildSettings #{xcodebuild_parameters.join(' ')}"
+      if FastlaneCore::Helper.xcode_atleast?('8.3')
```
KrauseFx: Could you reference in a command on why this was added, when it was fixed, etc.

KrauseFx (dismissed):

hjanuschka (approved):

punty:
@KrauseFx thanks for your feedback. I just implemented the changes you requested.

hjanuschka:
@punty 🚀  merged, many thx for your contribution 🎉

Issue #8895, opened 2017-04-16
[Regression] snapshot is not changing language of Simulator


tobihagemann:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I can't really tell if this issue is new with Xcode 8.3 or some fastlane update. The last time it worked was with Xcode 8.2.1 and fastlane 2.22.0, as far as I can remember. The language of the Simulator stays `en-US` in my case and doesn't change at all for snapshots that are supposed to be done in other languages.

I tried resetting all simulators with `fastlane snapshot reset_simulators`, but that didn't help.

##### Complete output when running fastlane, including the stack trace and command used
```
$ fastlane snapshot
[14:19:54]: Successfully loaded '/Users/<redacted>/fastlane/Snapfile' 📄

+----------------------------+---------------------------------+
|          Detected Values from './fastlane/Snapfile'          |
+----------------------------+---------------------------------+
| scheme                     | <redacted>UITests               |
| devices                    | ["iPhone 7 Plus", "iPad Air 2"] |
| languages                  | ["en-US", "de-DE", "nl", "ru"]  |
| output_directory           | ./fastlane/screenshots          |
| clear_previous_screenshots | false                           |
+----------------------------+---------------------------------+

[14:19:54]: $ xcodebuild -list -workspace ./<redacted>.xcworkspace
[14:19:55]: $ xcodebuild clean -showBuildSettings -workspace ./<redacted>.xcworkspace -scheme <redacted>UITests

+----------------------------+----------------------------------------------------------------+
|                                 Summary for snapshot 2.27.0                                 |
+----------------------------+----------------------------------------------------------------+
| scheme                     | <redacted>UITests                                              |
| devices                    | ["iPhone 7 Plus", "iPad Air 2"]                                |
| languages                  | ["en-US", "de-DE", "nl", "ru"]                                 |
| output_directory           | /Users/<redacted>/fastlane/screenshots                         |
| clear_previous_screenshots | false                                                          |
| workspace                  | ./<redacted>.xcworkspace                                       |
| launch_arguments           | [""]                                                           |
| output_simulator_logs      | false                                                          |
| skip_open_summary          | false                                                          |
| skip_helper_version_check  | false                                                          |
| reinstall_app              | false                                                          |
| erase_simulator            | false                                                          |
[...]

PR #8928, opened 2017-04-20
Fix snapshot creating file path url for simulatorHostHome

Torsph:
### Checklist

### Description

The issue is that `URL(string: simulatorHostHome)` will only generate for example `/users/myuser`. It is therefor  missing `file://` that it is getting from `URL(fileURLWithPath: simulatorHostHome)`.

If you add `URL(string: "file://\(simulatorHostHome)")` it will work as expected. But I will recommend to use `URL(fileURLWithPath: simulatorHostHome)`. That way we don't need to handle the `file://` our self.

I have added a small test to check if the `simulatorHostHome` actually exists. Since there was added a check for this in the previous commit.

### Motivation and Context

Fixes issue described in #8895

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -141,11 +141,11 @@ open class Snapshot: NSObject {
                 print("Couldn't find simulator home location. Please, check SIMULATOR_HOST_HOME env variable.")
                 return nil
             }
-            guard let homeDirUrl = URL(string: simulatorHostHome) else {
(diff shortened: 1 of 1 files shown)


wryphonedev:
Duplicates #8791

KrauseFx:
I just merged https://github.com/fastlane/fastlane/pull/8791, is that ok, or do we need another change? Thanks for the PR 👍

Torsph:
@KrauseFx that sounds good 👍  Didn't see the other PR :)

PR #8943, opened 2017-04-21
Check for default external group and support :groups option

ohayon:
Add support for the `:groups` option in pilot when adding testers. Also this checks to see if there is a group marked as `isDefaultExternalGroup` from the iTC API to add to that if available.

### DIFF
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -51,7 +51,7 @@ module Pilot
 
       return if config[:groups].nil?
 
-      groups = TestFlight::Group.filter_groups(app_id: uploaded_build.app_id) do |group|
+      groups = Spaceship::TestFlight::Group.filter_groups(app_id: app.apple_id) do |group|
         config[:groups].include?(group.name)
       end
 
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -17,7 +17,7 @@ module Pilot
           tester = Spaceship::Tunes::Tester::External.create!(email: config[:email],
                                                               first_name: config[:first_name],
                                                               last_name: config[:last_name])
-          UI.success("Successfully invited tester: #{tester.email}")
+          UI.success("Successfully added tester: #{tester.email} to your account")
         end
 
         app_filter = (config[:apple_id] || config[:app_identifier])
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -19,24 +19,24 @@ module Pilot
                                                               last_name: config[:last_name])
           UI.success("Successfully added tester: #{tester.email} to your account")
         end
-
-        app_filter = (config[:apple_id] || config[:app_identifier])
-        if app_filter
-          begin
-            app = Spaceship::Application.find(app_filter)
-            UI.user_error!("Couldn't find app with '#{app_filter}'") unless app
-
-            add_tester_to_groups(tester: tester, app: app, groups: config[:groups])
-
-            UI.success("Successfully added tester to app #{app_filter}")
-          rescue => ex
-            UI.error("Could not add #{tester.email} to app: #{ex}")
-            raise ex
-          end
-        end
       rescue => ex
         UI.error("Could not create tester #{config[:email]}")
         raise ex
+      end
+
+      app_filter = (config[:apple_id] || config[:app_identifier])
+      if app_filter
+        begin
+          app = Spaceship::Application.find(app_filter)
+          UI.user_error!("Couldn't find app with '#{app_filter}'") unless app
+
+          add_tester_to_groups(tester: tester, app: app, groups: config[:groups])
+
+          UI.success("Successfully added tester to app #{app_filter}")
+        rescue => ex
+          UI.error("Could not add #{tester.email} to app: #{app.name}")
+          raise ex
+        end
       end
     end
 
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -43,8 +43,8 @@ module Pilot
     def add_tester_to_groups(tester: nil, app: nil)
       default_external_group = app.default_external_group
       if default_external_group.nil? && config[:groups].nil?
-        UI.error("The app #{app.name} does not have a default external group.")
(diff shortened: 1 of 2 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pilot/lib/pilot/tester_manager.rb
```diff
@@ -43,6 +40,27 @@ def add_tester(options)
       end
     end
 
+    def add_tester_to_groups(tester: nil, app: nil)
+      default_external_group = app.default_external_group
+      if default_external_group.nil? && config[:groups].nil?
+        UI.error("The app #{app.name} does not have a default external group.")
+        UI.error("Please make sure to pass group names to the `:groups` option.")
+        return
```
KrauseFx: Should this be `UI.user_error!`? As otherwise we'd show a success message later on I think?

pilot/lib/pilot/tester_manager.rb
```diff
@@ -43,6 +40,27 @@ def add_tester(options)
...
+      end
+
+      default_external_group.add_tester!(tester)
```
KrauseFx: What if `default_external_group` is nil, but the `config[:groups]` is set?

pilot/lib/pilot/tester_manager.rb
ohayon: Yeah this could definitely be `user_error!`

pilot/lib/pilot/tester_manager.rb
ohayon: If `config[:groups]` is set, it should pass the return statement on the next line and then go into adding the tester to all the groups.

KrauseFx: Right, but calling `nil.default_external_group!` will raise an exception I think

ohayon: There is a check on line 45 to see if it is nil.

pilot/lib/pilot/tester_manager.rb
```diff
@@ -43,6 +40,27 @@ def add_tester(options)
       end
     end
 
+    def add_tester_to_groups(tester: nil, app: nil)
+      default_external_group = app.default_external_group
+      if default_external_group.nil? && config[:groups].nil?
+        UI.user_error!("The app #{app.name} does not have a default external group.")
+        UI.user_error!("Please make sure to pass group names to the `:groups` option.")
```
KrauseFx: `user_error!` raises the exception, this code line won't be reached. Try the following instead:
```ruby
UI.user_error!("The app #{app.name} does not have a default external group. Please make sure to pass group names to the `:groups` option.")
```

pilot/lib/pilot/tester_manager.rb
KrauseFx: Right, but:

> What if default_external_group is nil, but the config[:groups] is set?

pilot/lib/pilot/tester_manager.rb
```diff
@@ -43,6 +40,26 @@ def add_tester(options)
...
     end
 
+    def add_tester_to_groups(tester: nil, app: nil)
+      default_external_group = app.default_external_group
+      if default_external_group.nil? && config[:groups].nil?
+        UI.user_error!("The app #{app.name} does not have a default external group. Please make sure to pass group names to the `:groups` option.")
+        return
+      end
+
+      default_external_group.add_tester!(tester) unless default_external_group.nil?
```
KrauseFx: 💯

hjanuschka: @ohayon  does this only happen if user has 0 groups? as i get it, it seems the first created is also the default.
can't we create a "External" group if no default is found - and do a `UI.success`reporting it?

ohayon: @hjanuschka We have observed that apps added to TestFlight _after_ the changes from last week do not have a group that is marked as the `isDefaultExternalGroup` in the iTC API. We _could_ create a group called `External Testers` but it still wont have that flag set.

pilot/lib/pilot/tester_manager.rb
```diff
@@ -43,6 +40,26 @@ def add_tester(options)
       end
     end
 
+    def add_tester_to_groups(tester: nil, app: nil)
+      default_external_group = app.default_external_group
+      if default_external_group.nil? && config[:groups].nil?
```
mpirri: Is `add_tester_to_groups` meant to be a private method of TesterManager? I only ask because it takes a `tester` and an `app` but also references the `config` hash. The pattern in this file seems to be that public methods reference `config` and take an `options` argument. Private methods take specific parameters but don't reference `config`.

My thought: mark this method as private, and pass in the `groups` as an argument.

mpirri:
A possible issue - I have an app with no default external group.  When I add `pilot add` with no group name, it works great.  When I `pilot add` to a group name that exists, it works great. But, when I `pilot add` to a group name that DOES NOT exist, I get no error, and the tester is added to the account-level testers, but that is all.  Seems like this should be a warning or an error in this case.

snatchev (approved):
Code LGTM.

Was not able to test with real data.

pilot/lib/pilot/tester_manager.rb
seivan: @KrauseFx @ohayon Sorry, I actually managed to find this commit because I was searching form this in the code base. I don't agree that this should be an error because nothing actually fails.

We still get a build with an email. It could be a warning instead, and also how to remedy this without setting `:groups`.  For our case, we have two applications and one of them fails because for some reason, `app.default_external_group` returns nil, but not for the other app. We would like both to have external groups attached, but not sure how to remedy this either through fastlane or itunes connect. 

Just my two cents, hoping I can convince you to make this a warning/important and not an error as we're getting now in our CI even though it actually builds and submits. Thanks!

seivan: Also, this commit indicates that there is no longer a concept of default external groups for new applications since the recent Testflight update; unless I misunderstood.

https://github.com/fastlane/fastlane/pull/8943/files/bf03c6463a38763bf1c4a940b24c14375636a5ae#r112680533

PR #8951, opened 2017-04-21
Fix string handling in print_table wrapping

### LINKED ISSUE #8947: Typo in Carthage documentation
Sorry not to post my environment etc.
In version 2.28.2 there is a typo.
If you trigger the Carthage documentation via `fastlane action carthage`, the one key is called "fraeworks", not "frameworks"

mfurtak:
### Checklist

### Description

The fix replaces use of some string manipulation methods that have surprising behavior, with use of ones that are easier to use and understand. It also simplifies the data a little by removing nested arrays.

The main problem is that `delete!` does not act like a simple string replacement. It actually acts by remove all characters found in the input string from the target. This behavior can be seen by examining some of the code in question

```
irb(main):003:0> require 'colored'
=> true
irb(main):004:0> value = 'team_name'.yellow
=> "\e[33mteam_name\e[0m"
irb(main):005:0> value.scan(/(\e\[.*?m)/)
=> [["\e[33m"], ["\e[0m"]]
irb(main):006:0> colors = value.scan(/(\e\[.*?m)/)
=> [["\e[33m"], ["\e[0m"]]
irb(main):007:0> if colors && colors.length > 0
irb(main):008:1>   colors.each do |color|
irb(main):009:2*     value.delete!(color.first)
irb(main):010:2>     value.delete!(color.last)
irb(main):011:2>   end
irb(main):012:1> end
=> [["\e[33m"], ["\e[0m"]]
irb(main):013:0> value
=> "tea_nae"
```

Another simplification I've made is a small change to the usage of `scan`. As written, this returns an array of arrays, this makes later handling a little trickier.

```
irb(main):029:0> value = 'team_name'.yellow
=> "\e[33mteam_name\e[0m"
irb(main):030:0> colors = value.scan(/(\e\[.*?m)/)
=> [["\e[33m"], ["\e[0m"]]
```

According to the documentation, if we omit the capture group from the regex, then we won't get the array of arrays, instead getting a flat array of strings.

```
irb(main):039:0> value = 'team_name'.yellow
=> "\e[33mteam_name\e[0m"
irb(main):040:0> colors = value.scan(/\e\[.*?m/)
=> ["\e[33m", "\e[0m"]
```

I've also disabled a RuboCop check that I think gives advice that could easily lead to a bad decision.

### Motivation and Context

This fixes #8947 

The current implementation is causing 'm' chars to get dropped from the table values whenever newline wrapping occurs.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 Style/PercentLiteralDelimiters:
   Enabled: false
-  
+
 # kind_of? is a good way to check a type
 Style/ClassCheck:
   EnforcedStyle: kind_of?
@@ -16,6 +16,12 @@ Style/SafeNavigation:
 Performance/RegexpMatch:
   Enabled: false
 
+# This suggests use of `tr` instead of `gsub`. While this might be more performant,
+# these methods are not at all interchangable, and behave very differently. This can
+# lead to people making the substitution without considering the differences.
+Performance/StringReplacement:
+  Enabled: false
+
 # .length == 0 is also good, we don't always want .zero?
 Style/NumericPredicate:
   Enabled: false
@@ -23,7 +29,7 @@ Style/NumericPredicate:
 # this would cause errors with long lanes
 Metrics/BlockLength:
   Enabled: false
-  
+
 # this is a bit buggy
 Metrics/ModuleLength:
   Enabled: false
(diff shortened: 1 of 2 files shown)


ohayon (commented):
👍

endocrimes (approved):

PR #8962, opened 2017-04-21
Add tests for TestFlight methods in spaceship and fastlane_core

snatchev:
Future of spaceship testing:

* No fixtures: these are too brittle to maintain
* Use sinatra app to construct client responses near the test site
* More focus on each Unit isolation. Heavier use of mock objects

This is a work in progress

### Checklist


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hjanuschka:
wow! ❤️  i am in love with the sinatra + json sheme approach! 👍

spaceship/lib/spaceship/test_flight/build.rb
```diff
@@ -104,7 +104,7 @@ def self.latest(app_id: nil, platform: nil)
     #
     # Note: this will overwrite any non-saved changes to the object
     #
-    # @return (Spaceceship::Base::DataHash) the raw_data of the build.
+    # @return (Spaceship::Base::DataHash) the raw_data of the build.
```
ohayon: Nice find!

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -75,53 +157,25 @@ def delete_tester_from_group(group_id: nil, tester_id: nil, app_id: nil)
...
-        req.headers['Content-Type'] = 'application/json'
-      end
-      handle_response(response)
-    end
+    ##
+    # @!group Testinfo
+    ##
 
-    def get_groups(app_id: nil)
+    def put_testinfo(app_id: nil, testinfo: nil)
```
ohayon: Do we call this anywhere that it needs to get updated? 

Also, just curious why this is changed to know about the model object rather than passing in the fields?

snatchev: Ah, good question. We currently don't use it. I did use it at one point in a debugging phase to try to fix the problem with our shared test account.

I think it's okay to pass in objects that look like hashes (or at least respond to `to_json`) as long as they don't need any mutation to result in a successful response. Here, the client doesn't really know anything about `testinfo`.

spaceship/lib/spaceship/test_flight/build.rb
```diff
@@ -71,12 +71,12 @@ class Build < Base
...
     }
 
-    # Find a Build by `build_id`. Returns `nil` if can't find it.
+    # Find a Build by `build_id`.
     #
     # @return (Spaceship::TestFlight::Build)
     def self.find(app_id: nil, build_id: nil)
       attrs = client.get_build(app_id: app_id, build_id: build_id)
-      self.new(attrs) if attrs
+      self.new(attrs)
```
ohayon: Will this be safe to have this return a `Build` object without any attributes? Might cause an issue if we are checking for its presence at any point right?

spaceship/lib/spaceship/test_flight/client.rb
ohayon: Just read the comment above about passing in large objects. Got it!

spaceship/lib/spaceship/test_flight/build.rb
snatchev: `initialize` always expects a hash. If `nil` is passed in, it throws an exception. So far the pattern has been that if an error response is received, an `UnexpectedResponse` exception is thrown. I think throwing exceptions is consistent in this case.

spaceship/spec/test_flight/tester_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
+
+describe Spaceship::TestFlight::Tester do
+  let(:tester) { Spaceship::TestFlight::Tester.new }
```
ohayon: @mpirri Is this being used?

fastlane_core/spec/build_watcher_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+      expect(UI).to receive(:success).with("Successfully finished processing the build #{export_compliance_required_build.train_version} - #{export_compliance_required_build.build_version}")
+      found_build = FastlaneCore::BuildWatcher.wait_for_build_processing_to_be_complete(app_id: 'some-app-id', platform: :ios)
+
+      expect(found_build).to eq(export_compliance_required_build)
+    end
+
+    it 'waits when a build is still processing' do
+      expect(Spaceship::TestFlight::Build).to receive(:all_processing_builds).and_return([processing_build])
+      expect(Spaceship::TestFlight::Build).to receive(:builds_for_train).and_return([processing_build], [ready_build])
+      expect(FastlaneCore::BuildWatcher).to receive(:sleep).with(10)
```
mfurtak: suggestion: change this to `expect(FastlaneCore::BuildWatcher).to receive(:sleep)`

That way, if we ever change how long we want it to sleep, we don't have to go back and update all the tests where we expected a particular number

spaceship/lib/spaceship/test_flight/base.rb
```diff
@@ -4,6 +4,20 @@ def self.client
       @client ||= Client.client_with_authorization_from(Spaceship::Tunes.client)
     end
 
+    ##
+    # Have subclasses inherit the client from their superclass
+    #
+    # Essentially, we are making a class-inheritable-accessor as described here:
+    # https://apidock.com/rails/v4.2.7/Class/class_attribute
+    #
+    # TODO[snatchev]: move this into spaceship/base.rb once all the tests pass
```
mfurtak: TODONE?

mfurtak (approved):
This looks awesome - I'd definitely like a run-down on how it all works, though :wink:

Might it make sense to update some spaceship developer docs to explain the strategy?

spaceship/lib/spaceship/test_flight/base.rb
mfurtak: Sorry, that was a cryptic comment - are we now in a place where this TODO could be resolved?

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -1,21 +1,37 @@
...
       platform ||= "ios"
       response = request(:get, "providers/#{team_id}/apps/#{app_id}/platforms/#{platform}/trains")
       handle_response(response)
     end
 
     def get_builds_for_train(app_id: nil, platform: nil, train_version: nil)
       assert_required_params(__method__, binding)
-      platform ||= "ios"
 
+      platform ||= "ios"
```
snatchev: `platform` will never be nil because of the assertion above. Perhaps we should set `platform: 'ios'` in the parameter names, or get rid of this.

snatchev:
I agree with @mfurtak about adding some documentation. Where should we add some developer docs?

* at the top of some classes
* a README in the relevant folder (i.e. `spaceship/spec`)?
* in another designated documentation location?

ohayon:
I think a README.md in the `spaceship/spec` directory would make sense. @mfurtak thoughts on where the best place for this would be is?

spaceship/docs/TestFlightTesting.md
```diff
@@ -9,40 +9,61 @@ To run the tests, in your terminal run:
...
 ### Models
+
+Since the data models expect the client to return JSON data as a Ruby hash, we can reasonably mock the client response using RSpec doubles. We should *not* rely on HTTP fixtures because they are brittle, introduce global state, are not easily decomposable and are difficult to maintain. Instead, use the helper method `mock_client_response` to set up the expected data returned by the API. This design also leads the client to be as thin and logic-free as possible.
+
+Defining the response near the test site makes it easy to understand and maintain. Try not to include any more data than is necessary for the spec to pass.
+
 **Examples:**
 
-Mocked responses:
+At the top of your data model, set the client to be a `mock_client`:
```
mfurtak: "At the top of your data model _spec_"?

spaceship/docs/TestFlightTesting.md
```diff
@@ -9,40 +9,61 @@ To run the tests, in your terminal run:
 bundle exec rspec spaceship/spec
 ```
 
+## Overview
+
+Spaceship wraps various APIs using the following pattern:
+
+A simple `client` and various data models, usually suclassed from a `Base` model (e.g. Spaceship::TestFlight::Base)
```
mfurtak: "suclassed" -> "subclassed"

spaceship/docs/TestFlightTesting.md
```diff
@@ -9,40 +9,61 @@ To run the tests, in your terminal run:
...
-	  }
-	]
-  }
+Now, anytime we use a data model that is a subclass of `Spaceship::TestFlight::Base`, it has a `client` that is our mock.
+
+We then configure the response for a given client method like this:
+
+```ruby
+before do
+  mock_client_response(:get_tester, with: { tester_id: 1 }) do
```
mfurtak: Where does this method come from? Is there something that has to be required in the spec to make it work?

spaceship/docs/TestFlightTesting.md
```diff
@@ -9,40 +9,61 @@ To run the tests, in your terminal run:
...
+    {
+      id: 1,
+      name: 'Mr. Tester'
+    }
+  end
 end
 ```
 
+The first parameter is the name of the method we are mocking, and `with:` parameter specifies some required parameter to that method. If you don't give it a `with:`, the mock will accept any parameter. The block is the return value of calling `client.get_tester`.
```
mfurtak: Can this be more than one parameter? If so, is it more clear to say "required parameters" and "any parameters"?

mfurtak (approved):
Looking great, thanks for the docs :+1: Just left a few questions - merge when ready

spaceship/lib/spaceship/test_flight/build.rb
```diff
@@ -130,6 +130,10 @@ def export_compliance_missing?
       external_state == BUILD_STATES[:export_compliance_missing]
     end
 
+    def self.processed?
```
tmm1: This should be `def processed?`, not `def self.processed?`

```
NoMethodError: [!] undefined method `processed?' for #<Spaceship::TestFlight::Build:0x007fb066700ee0>
gems/fastlane-2.28.6/fastlane_core/lib/fastlane_core/build_watcher.rb:15:in `block in wait_for_build_processing_to_be_complete'
```

mpirri: Thanks so much for bringing this to our attention! I can't believe with all the unit tests we added that we missed this...

Anyhow, here is a PR to fix it: https://github.com/fastlane/fastlane/pull/9025
As soon as we can get a 🚢 on it, we will release a new build.

PR #8971, opened 2017-04-23
Generic Github API action `github_api` - existing github api actions updated to use shared actions

tommeier:
### Checklist
- [x] I've updated the documentation if necessary. (not sure about this - is there a way to generate or is the existing auto-docs ok?)

### Motivation and Context

Hi, apologies for the size of the PR, I may have got a little carried away. 

Originally, I was simply looking for a way to commit a github file directly on the (github enterprise) server. Realising that you aren't taking new actions, I built my own personal one using parts of the other examples (specifically `set_github_release` and `get_github_release`). After building this, and using in production for a while, I delved a bit deeper and noticed how generic the github API is (can be) and the duplicated logic across several actions.

As such - I thought it might help future users and contributors if there was a completely generic endpoint for accessing the Github API. Thus, allowing anyone to build a custom action for _any_ github api endpoint easily. I've then proceeded to update all existing github API actions, removed all duplicated logic and all now use the one shared `GithubApiAction`, this should pay dividends going forward if the API changes and only one spot needs to be changed (or the Excon dependency is removed - for example) - or a terminal progress bar is implemented.

I've updated the existing actions with some minor improvements, but I've been conscious of avoiding _any_ API changes (though i'd love to make them all uniform some day - each seems to have a different parameter names)

For the two new actions and all existing actions I've checked and added tests, ensuring existing tests exhibit the same behaviour. I've also manually tested all endpoints. Some logic (like redirects) i've pushed into the base generic action instead of being present on only one action (`get_github_release`). All existing actions should behave identically as before when errors are handled and returned results given - except `create_pull_request` will now return the JSON result like other actions instead of nil as before.

Hope its helpful! Thanks for a great project! 🎉 

#### New actions

  * `github_api` - completely generic action for calling any Github API endpoint and parsing the resulting json.
  * `commit_github_file` - commit github file directly (one file -> one commit). I mostly kept this in as another example of using the generic `github_api` action.

#### Updated actions

  * `create_pull_request`
  * `get_github_release`
  * `set_github_release`
  * `install_xcode_plugin`


tommeier:
CLA signed

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,261 @@
...
+        end
+
+        Actions.lane_context[SharedValues::GITHUB_API_STATUS_CODE] = result[:status]
+        Actions.lane_context[SharedValues::GITHUB_API_RESPONSE] = result[:body]
+        Actions.lane_context[SharedValues::GITHUB_API_JSON] = result[:json]
+
+        return result
+      end
+
+      def self.parse_headers(api_token, overrides)
```
giginet: these methods should be `private`

fastlane/spec/actions_specs/set_github_release_spec.rb
```diff
@@ -4,7 +4,7 @@
       it "sends and receives the right content from GitHub" do
         stub_request(:post, "https://api.github.com/repos/czechboy0/czechboy0.github.io/releases").
           with(body: "{\"tag_name\":\"tag33\",\"name\":\"Awesome Release\",\"body\":\"Bunch of new things :+1:\",\"draft\":false,\"prerelease\":false,\"target_commitish\":\"test\"}",
-          headers: { 'Authorization' => 'Basic MTIzNDVhYmNkZQ==', 'Host' => 'api.github.com:443', 'User-Agent' => 'fastlane-set_github_release' }).
+          headers: { 'Authorization' => 'Basic MTIzNDVhYmNkZQ==', 'Host' => 'api.github.com:443', 'User-Agent' => 'fastlane-github_api' }).
```
giginet: Are there any problems that User agents are modified?
(I think there are no problems.)

Or you can add `user_agent` option to `github_api` action to set it for just in case.

giginet (approved):
Awesome!

I want to use this new action. So I'd like to merge this ASAP.

fastlane/lib/fastlane/actions/github_api.rb
tommeier: yup definitely - thanks will fix

fastlane/spec/actions_specs/set_github_release_spec.rb
tommeier: Great idea, I wasn't sure if it would be a problem - I assumed it was just because it was a required attribute for the Github API, and not for any other reason, see: https://developer.github.com/v3/#user-agent-required

But I can add an option if people will actually use it?

tommeier:
@giginet thanks for reviewing! Question - I've just realised why I needed to have defaults doubled up with my `GithubApiAction.run` internal commands, as I'm not calling it with the full `FastlaneCore::Configuration` object.

It would probobly be best to change to the full calls with `Configuration` object internally yeah? - Still catching up on some of the patterns within the project.

fastlane/spec/actions_specs/set_github_release_spec.rb
giginet: I just got worried.
Maybe, it isn't needed. Thanks.

tommeier: No its a very good point, especially if github were to throttle requests based on the header, I've just made some changes to the tests to show you can override it on the `GithubApiAction`. So if you use the optional `headers:` parameter, you can set `User-Agent` to anything you want. By default it will set to `github_api`. I guess if people find the need we can pass `headers` along from any source github api action as a simple future change.

fastlane/lib/fastlane/actions/github_api.rb
tommeier: updated as private

tommeier:
> as I'm not calling it with the full FastlaneCore::Configuration object.

I can see some actions do this internally, and some that don't, so I'm not sure what the best way would be to keep in fastlane conventions.

tommeier:
After further looking around the fastlane code I can't see any examples of the `.available_options`. I thought perhaps I should call `GithubApiAction.run(config)` with `config = FastlaneCore::Configuration.create(GithubApiAction.available_options, params)` instead of a raw hash. Can change it to this if you'd like, but I'm not seeing any other patterns where this occurs already. The benefit of doing this would ensure the defaults are applied correctly on the wrapping actions (e.g. `SetGithubRelease`) - but as they have their own defaults it doesn't really apply. Let me know if you want me to change all calls to `GithubApiAction.run` send a `Configuration` object but otherwise I think this is ok.

ohayon:
Hey @tommeier this is really awesome. We had some conversation about this PR yesterday and I wanted to give you a little update: Some of us are going to do some local testing just to make sure things work on some of our setups and then we would love to get this merged and released early on in a week so that we can watch it closely once it gets out in the wild and make sure that everything is working for all of our users. 🚀  Hopefully we can queue that up for early next week.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+        headers = { 'User-Agent' => 'fastlane-github_api' }
+        headers['Authorization'] = "Basic #{Base64.strict_encode64(api_token)}" if api_token
+        headers.merge(overrides || {})
+      end
+      private_class_method :parse_headers
+
+      def self.parse_url(server_url, path, url)
+        return_url = path ? File.join(server_url, path) : url
+
+        UI.user_error!("Please provide either 'path' or full 'url' for github api endpoint") unless return_url
```
ohayon: What happens if there is no `server_url`?

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+        return result
+      end
+
+      def self.parse_headers(api_token, overrides)
+        require 'base64'
+        headers = { 'User-Agent' => 'fastlane-github_api' }
+        headers['Authorization'] = "Basic #{Base64.strict_encode64(api_token)}" if api_token
+        headers.merge(overrides || {})
+      end
+      private_class_method :parse_headers
```
ohayon: Is there a reason to mark each method private this way? Another option is to wrap all of these class methods in a `class << self` block and them mark them as `private in the same way instance methods are set to private:
```
class << self 
  private
  def some_method
  end
end
```

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+        require 'excon'
+        opts = {
+          secure: true,
+          debug: false
+        }.merge(params)
+
+        Excon.defaults[:ssl_verify_peer] = opts[:secure]
+        middlewares = Excon.defaults[:middlewares] + [Excon::Middleware::RedirectFollower] # allow redirect in case of repo renames
+
+        UI.message("#{http_method} : #{url}")
```
ohayon: Can we use `UI.verbose` here?

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+                                       is_string: false,
+                                       default_value: {},
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :secure,
+                                     env_name: "FL_GITHUB_API_SECURE",
+                                     description: "Optionally disable secure requests (ssl_verify_peer)",
+                                     is_string: false,
+                                     default_value: true,
+                                     optional: true),
+          FastlaneCore::ConfigItem.new(key: :debug,
```
ohayon: As mentioned above, I think we should get rid of this option and just use our verbose logging for this.

fastlane/lib/fastlane/actions/commit_github_file.rb
```diff
@@ -0,0 +1,184 @@
...
+          FastlaneCore::ConfigItem.new(key: :message,
+                                       env_name: "FL_COMMIT_GITHUB_FILE_MESSAGE",
+                                       description: "The commit message. Defaults to the file name",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :secure,
+                                     env_name: "FL_COMMIT_GITHUB_FILE_SECURE",
+                                     description: "Optionally disable secure requests (ssl_verify_peer)",
+                                     is_string: false,
+                                     default_value: true,
+                                     optional: true),
```
ohayon: Spacing here is off from the rest.

fastlane/lib/fastlane/actions/commit_github_file.rb
```diff
@@ -0,0 +1,184 @@
...
+      def self.run(params)
+        repo_name = params[:repository_name]
+        branch = params[:branch] ||= 'master'
+        commit_message = params[:message]
+
+        file_path = params[:path]
+        file_name = File.basename(file_path)
+        expanded_file_path = File.expand_path(file_path)
+
+        UI.important("Creating commit on #{repo_name} on branch \"#{branch}\" for file \"#{file_path}\"")
```
ohayon: Curious to hear your thoughts about making some of these important and message lines use `UI.verbose` instead to keep noise down unless the user asks for it?

fastlane/lib/fastlane/actions/commit_github_file.rb
```diff
@@ -0,0 +1,184 @@
...
+                                       env_name: "FL_COMMIT_GITHUB_FILE_MESSAGE",
+                                       description: "The commit message. Defaults to the file name",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :secure,
+                                     env_name: "FL_COMMIT_GITHUB_FILE_SECURE",
+                                     description: "Optionally disable secure requests (ssl_verify_peer)",
+                                     is_string: false,
+                                     default_value: true,
+                                     optional: true),
+          FastlaneCore::ConfigItem.new(key: :debug,
```
ohayon: Would be great to not include this option and instead use the `UI.verbose` method that we have. I think they are serving the same kind of purpose?

fastlane/lib/fastlane/actions/get_github_release.rb
```diff
@@ -7,43 +7,38 @@ module SharedValues
...
+          server_url: params[:server_url],
+          api_token: params[:api_token],
+          http_method: 'GET',
+          path: "repos/#{params[:url]}/releases",
+          errors: {
+            404 => proc do |result|
+              UI.error("Repository #{params[:url]} cannot be found, please double check its name and that you provided a valid API token (if it's a private repository).")
+              return nil
+            end,
+            401 => proc do |result|
```
ohayon: Do we think this could be handled lower down in the `GithubApiAction` itself because 401's should always be invalid tokens (right?)?

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+
+        UI.message("#{http_method} : #{url}")
+
+        connection = Excon.new(url)
+        connection.request(
+          method: http_method,
+          headers: headers,
+          middlewares: middlewares,
+          body: body,
+          debug_request: opts[:debug],
```
ohayon: This can check `FastlaneCore::Globals.verbose?` instead of passing this param (but I could be missing the point of what the `debug_request` and `debug_response` params do in Excon

fastlane/lib/fastlane/actions/set_github_release.rb
```diff
@@ -11,70 +11,83 @@ def self.run(params)
...
+            422 => proc do |result|
+              UI.error(result[:body])
+              UI.error("Release on tag #{tag_name} already exists!")
+              return nil
+            end,
+            404 => proc do |result|
+              UI.error(result[:body])
+              UI.user_error!("Repository #{repo_name} cannot be found, please double check its name and that you provided a valid API token (GITHUB_API_TOKEN)")
+            end,
+            401 => proc do |result|
```
ohayon: Maybe lets handle these in the action itself?

ohayon (changes requested):
Hey @tommeier, again, this is an amazing PR. Thank you so much for putting in all this work. The design is great, I really love the error handling blocks. 

I left a bunch of notes, most of which are style things that you can feel free to disagree with and we can have some more conversation around them. 

It's a big PR so while I spent a long time reviewing it line-by-line, I may have missed some things so apologies in advance if I come back in and add some more comments.

fastlane/lib/fastlane/actions/create_pull_request.rb
```diff
@@ -6,34 +6,35 @@ module SharedValues
...
-        require 'base64'
-
         UI.message("Creating new pull request from '#{params[:head]}' to branch '#{params[:base]}' of '#{params[:repo]}'")
 
-        url = "#{params[:api_url]}/repos/#{params[:repo]}/pulls"
-        headers = { 'User-Agent' => 'fastlane-create_pull_request' }
-        headers['Authorization'] = "Basic #{Base64.strict_encode64(params[:api_token])}" if params[:api_token]
-
-        data = {
+        request_body = {
```
tommeier: agreed! from scanning the code it looks like a good pattern would be `payload` for request json, as `json` is used as the response parameter in most other places. I've changed all other areas in this PR to match too.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+
+    class GithubApiAction < Action
+      def self.run(params)
+        require 'json'
+
+        http_method = (params[:http_method] || 'GET').to_s.upcase
+        url = parse_url(params[:server_url], params[:path], params[:url])
+        headers = parse_headers(params[:api_token], params[:headers])
+        request_body = parse_body(params[:body], params[:raw_body])
+        handled_errors = params[:errors] || {}
```
tommeier: yup, much better, I think I was going for it being succinct, but yeah far better to be descriptive. All usage changed to `error_handlers` and associated variable name.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+
+        status_code = response[:status]
+        result = {
+          status: status_code,
+          body: response.body || "",
+          json: parse_json(response.body) || {}
+        }
+
+        if status_code.between?(200, 299)
+          if params[:debug]
```
tommeier: ah - thanks, I forgot that was auto-logged on debug mode - updated. I wasn't aware of the global for verbose, so yup that will work perfectly in this case too (https://github.com/fastlane/fastlane/pull/8971#discussion_r122054377). I've removed the debug parameter for two of the actions and will now use this to discover when to print request/response bodies, much cleaner.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+        end
+
+        Actions.lane_context[SharedValues::GITHUB_API_STATUS_CODE] = result[:status]
+        Actions.lane_context[SharedValues::GITHUB_API_RESPONSE] = result[:body]
+        Actions.lane_context[SharedValues::GITHUB_API_JSON] = result[:json]
+
+        return result
+      end
+
+      def self.parse_headers(api_token, overrides)
```
tommeier: much better thanks, changed to `construct_` for all similar ones too.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+        require 'base64'
+        headers = { 'User-Agent' => 'fastlane-github_api' }
+        headers['Authorization'] = "Basic #{Base64.strict_encode64(api_token)}" if api_token
+        headers.merge(overrides || {})
+      end
+      private_class_method :parse_headers
+
+      def self.parse_url(server_url, path, url)
```
tommeier: Updated to `construct_` as per others

fastlane/lib/fastlane/actions/github_api.rb
tommeier: was following other examples in the code base, but I agree this would be much cleaner, especially in a file like this with so many private methods, wrapping them all up.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+        return_url = path ? File.join(server_url, path) : url
+
+        UI.user_error!("Please provide either 'path' or full 'url' for github api endpoint") unless return_url
+        return_url
+      end
+      private_class_method :parse_url
+
+      def self.parse_body(body, raw_body)
```
tommeier: Updated to construct_ as per others

fastlane/lib/fastlane/actions/github_api.rb
tommeier: Shouldn't ever happen as the parameter has a default value for `https://api.github.com`, but i've added a catch there for server URL to be clear in case the default is ever removed.

fastlane/lib/fastlane/actions/github_api.rb
tommeier: done :)

fastlane/lib/fastlane/actions/github_api.rb
tommeier: fantastic idea! updated and removed the parameter from 2 of the actions, can be based off this global now.

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,266 @@
...
+                                       description: "The endpoint path. e.g. '/repos/:owner/:repo/readme'",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :url,
+                                       env_name: "FL_GITHUB_API_URL",
+                                       description: "The complete full url - used instead of path. e.g. 'https://uploads.github.com/repos/fastlane...'",
+                                       optional: true,
+                                       verify_block: proc do |value|
+                                         UI.user_error!("Please include the protocol in the url, e.g. https://uploads.github.com") unless value.include? "//"
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :errors,
```
tommeier: Updated, much better.

fastlane/lib/fastlane/actions/github_api.rb
tommeier: :+1:

fastlane/lib/fastlane/actions/commit_github_file.rb
tommeier: I'm ok with either way, but when testing or in our own setup, I would expect to see this information when _not_ running debug/verbose mode as it is the minimum information to know exactly what has occured. Can change if you like or to just 'info' but I think its pretty important here?

fastlane/lib/fastlane/actions/commit_github_file.rb
tommeier: Yup - burned, thanks!

fastlane/lib/fastlane/actions/get_github_release.rb
tommeier: Yeah - I agree - my thinking here was to ensure no change to existing setups _at all_. So anyone that is relying on this error message, will continue to see this error message. Perhaps users are regexing for this message before trying something new (i have no idea).

Happy to change it, but the aim of this PR was always to ensure every existing setup would continue to work. Let me know if you'd like me to change it to the base github api action. For instance, the base github api action catch all for all errors will display much more information.

fastlane/lib/fastlane/actions/set_github_release.rb
tommeier: Same reason as https://github.com/fastlane/fastlane/pull/8971#discussion_r122365260 - let me know if you'd like me to change it. I definitely would like to, thats the whole reason for having a base class so we can respond to these generic shared errors across the board.

tommeier:
Hi @ohayon, thanks for the review! great suggestions, applied them all, please let me know if you have any further suggestions and whether we should push the 401 generic error to the GithubApiAction (at the risk of breaking existing setups).

tommeier:
@ohayon any way I can assist with this?

ohayon:
Hey @tommeier we are going to merge and release this on Monday of next week! Putting it on our todo list now! 🚀

ohayon (approved):

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,258 @@
...
+
+          return result
+        end
+
+        #####################################################
+        # @!group Documentation
+        #####################################################
+
+        def description
+          "Call a Github API endpoint and get the resulting JSON response"
```
mpirri: "Github" -> "GitHub".

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,258 @@
...
+        end
+
+        def details
+          "Calls any Github API endpoint. You must provide your GitHub Personal token (get one from https://github.com/settings/tokens/new).
```
mpirri: "Github" -> "GitHub".

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,258 @@
...
+                                         default_value: true,
+                                         optional: true)
+          ]
+        end
+
+        def output
+          [
+            ['GITHUB_API_STATUS_CODE', 'The status code returned from the request'],
+            ['GITHUB_API_RESPONSE', 'The full response body'],
+            ['GITHUB_API_JSON', 'The parsed json returned from Github']
```
mpirri: "Github" -> "GitHub".

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -0,0 +1,258 @@
...
+          headers['Authorization'] = "Basic #{Base64.strict_encode64(api_token)}" if api_token
+          headers.merge(overrides || {})
+        end
+
+        def construct_url(server_url, path, url)
+          UI.user_error!("Please provide server URL, eg: https://api.github.com") unless server_url
+
+          return_url = path ? File.join(server_url, path) : url
+
+          UI.user_error!("Please provide either 'path' or full 'url' for github api endpoint") unless return_url
```
mpirri: "github" -> "GitHub".

fastlane/lib/fastlane/actions/commit_github_file.rb
```diff
@@ -0,0 +1,178 @@
...
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "This will commit a file directly on GitHub via the API"
+      end
+
+      def self.details
+        "Commits a file directly to Github. You must provide your GitHub Personal token
```
mpirri: "Github" -> "GitHub".

mpirri (approved):
A few minor GitHub capitalization errors in result messages, but this looks :100: otherwise!

PR #8982, opened 2017-04-24
Directory ordering of default values for .ipa and . pkg files

tache:
### Checklist
- [x] I've read the [Contribution Guidelines]

### Motivation and Context
When a project has multiple targets, an error could occur when there are .ipa files present for different targets having different bundle ids. The Ruby DIR command in some fastlane command default settings might not select the last .ipa file created. for example, when this occurred with pilot, it would result with the wrong .ipa file being uploaded to ITC, which would fail with an error stating that the bundle id changed. The same was noted for the .pkg option for deliver.

### Description
Replaced the default DIR ordering that was not guaranteed with a time ordered sort, which plucks the most recent ipa or pkg file for the default value.

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -37,7 +37,7 @@ module Deliver
                                      optional: true,
                                      env_name: "DELIVER_IPA_PATH",
                                      description: "Path to your ipa file",
-                                     default_value: Dir["*.ipa"].first,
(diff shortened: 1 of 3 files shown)


tache:
I signed it!

tache:
This still needs to be applied. Issue is still present in the 2.28.5 version

KrauseFx (approved):
I like the idea, nice patch 👍

KrauseFx:
Thanks for your contribution @tache, and sorry for the delayed reply 🚀

PR #8984, opened 2017-04-24
Don't submit non-externally distributed builds for review

mpirri:
### Checklist

### Description

Currently `pilot` is submitting all builds for review. Changed to only submit externally distributed (i.e., build where `distribute_external = true`) for review.  For non-external builds, we will save the build info if export compliance needs to be updated.

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -110,10 +110,10 @@ module Pilot
 
       # This is where we could add a check to see if encryption is required and has been updated
       uploaded_build.export_compliance.encryption_updated = false
-      uploaded_build.beta_review_info.demo_account_required = false
-      uploaded_build.submit_for_testflight_review!
 
       if options[:distribute_external]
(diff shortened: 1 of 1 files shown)


pilot/lib/pilot/build_manager.rb
```diff
@@ -110,25 +110,29 @@ def distribute_build(uploaded_build, options)
...
         if external_group.nil? && options[:groups].nil?
           UI.user_error!("You must specify at least one group using the `:groups` option to distribute externally")
         end
 
-      end
-
-      if options[:groups]
-        groups = Spaceship::TestFlight::Group.filter_groups(app_id: uploaded_build.app_id) do |group|
-          options[:groups].include?(group.name)
+        if options[:groups]
```
mfurtak: Is groups a thing that only applies to external distribution? This used to get evaluated in all cases before...

mpirri: Correct: groups only apply to external distribution.  There is only a single internal distribution group.

mfurtak (approved):

PR #9005, opened 2017-04-26
Fix `pilot distribute`

tmm1:
This makes `fastlane pilot distribute --distribute_external` work again on my app.

```
ArgumentError: [!] wrong number of arguments (given 1, expected 2)
gems/ruby/2.3.0/gems/fastlane-2.28.5/pilot/lib/pilot/build_manager.rb:43:in `distribute'
gems/ruby/2.3.0/gems/fastlane-2.28.5/pilot/lib/pilot/commands_generator.rb:72:in `block (2 levels) in run'
gems/ruby/2.3.0/gems/commander-fastlane-4.4.4/lib/commander/command.rb:178:in `call'
```

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -37,10 +37,10 @@ module Pilot
       UI.message("If you want to skip waiting for the processing to be finished, use the `skip_waiting_for_build_processing` option")
       latest_build = FastlaneCore::BuildWatcher.wait_for_build_processing_to_be_complete(app_id: app.apple_id, platform: platform)
 
-      distribute(options, latest_build)
+      distribute(options, build: latest_build)
     end
 
-    def distribute(options, build = nil)
+    def distribute(options, build: nil)
       start(options)
       if config[:apple_id].to_s.length == 0 and config[:app_identifier].to_s.length == 0
         config[:app_identifier] = UI.input("App Identifier: ")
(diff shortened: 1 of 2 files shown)


tmm1:
@googlebot I signed the CLA

taquitos (changes requested):
This is looking great so far, could you update the PR to use the PR template: https://github.com/fastlane/fastlane/blob/master/.github/PULL_REQUEST_TEMPLATE.md
I'd like to understand more about motivation, testing, etc.

tmm1:
Sorry, but I'm not sure what else there is to say.

I encountered a bug preventing me from distributing builds and I fixed it. I provided context as to the source of the bug in my PR and commit messages.

pilot/lib/pilot/build_manager.rb
```diff
@@ -40,12 +40,17 @@ def upload(options)
       distribute(options, latest_build)
     end
 
-    def distribute(options, build)
+    def distribute(options, build = nil)
```
ohayon: Would be great to change this method over to have named parameters rather than just setting the default value!

taquitos (approved):

pilot/lib/pilot/build_manager.rb
tmm1: Sure.

ohayon:
Thanks @tmm1 !!! 🚀

PR #9007, opened 2017-04-27
Provided an option to restrict fastlane version to a fixed version in addition to minimum version check

karthik-krishnan:
Provided an option to restrict fastlane version to a fixed version in addition to minimum version check

### Checklist

### Motivation and Context

Currently we can mandate minimum version of fastlane required for the project.  However, if the teams want to be able to mandate a very specific version of fastlane in the environment, there is no way out for them.

### Description
I have now enhanced the fastlane_version action to allow an option to specific if fastlane should lock to a specific version.   For instance, instead of 

> fastlane_version 1.3

one can mention 
> fastlane_version 1.3, :lock => true

I have updated the unit tests as well.

### DIFF
--- a/fastlane/lib/fastlane/actions/fastlane_version.rb
+++ b/fastlane/lib/fastlane/actions/fastlane_version.rb
@@ -7,33 +7,42 @@ module Fastlane
       def self.run(params)
         params = nil unless params.kind_of?(Array)
         value = (params || []).first
+        options = (params and params.size > 1) ? params.last : {}
+        lock = options[:lock] || false
         defined_version = Gem::Version.new(value) if value
 
-        UI.user_error!("Please pass minimum fastlane version as parameter to fastlane_version") unless defined_version
-
-        if Gem::Version.new(Fastlane::VERSION) < defined_version
-          FastlaneCore::UpdateChecker.show_update_message('fastlane', Fastlane::VERSION)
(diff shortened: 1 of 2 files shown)


karthik-krishnan:
I signed it!

karthik-krishnan:
Build as well as rubocop is successful locally.  How do I go about re-trigerring the build now?

KrauseFx (changes requested):
Thanks for taking the time to submit this PR, great work. However, I don't think that's the right approach, to have more control over the _fastlane_ version, you should make use of a Gemfile, check out our [docs on how to use a Gemfile](https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile) 👍  Let me know if that helps

karthik-krishnan:
Thanks Felix.  I was also thinking to use Bundler and gemfile.lock eco
system to control the gem version treating fastlane gem like any other gem
in the environment.

On another note, I am not sure what happens with the "update_fastlane" call
inside the script.  My understanding is update_fastlane is happening inside
of the script execution context and was not sure if the subsequent script
code is running on the latest version of fastlane(which could be greater
than the gemfile.lock version)


Best regards,
Karthik Krishnan.
http://karthiks.in

KrauseFx:
If you use a Gemfile, you should probably not use the `update_fastlane` action, but run `bundle update` from time to time instead. This way your builds are reproducible 👍

karthik-krishnan:
Got it Thanks Felix.

Best regards,
Karthik Krishnan.
http://karthiks.in

Issue #9014, opened 2017-04-27
FASTLANE_DISABLE_COLORS not working with pilot anymore


tmm1:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Using `pilot` to add testers and perform other operations now outputs with colors, even if colors are disabled.

##### Complete output when running fastlane, including the stack trace and command used

Output looks identical when copy/pasted, but is colored even though I set `FASTLANE_DISABLE_COLORS=1`. This used to work as expected, so it seems like a regression since the recent testflight changes.

KrauseFx:
Thanks for reporting this, I can reproduce the issue. I'll look into resolving it.

KrauseFx:
Just pushed https://github.com/fastlane/fastlane/issues/9016 👍

PR #9016, opened 2017-04-28
Fix not disabling colors properly when using the `FASTLANE_DISABLE_COLORS` env variable

KrauseFx:
This was a little difficult to debug, but here we go 🚀

Fixes https://github.com/fastlane/fastlane/issues/9014


mfurtak:
Can you say a bit about what the problem was, and why this fixes it? Any idea why this started being a problem now?

KrauseFx:
I don't know why this issue was introduced recently, and when it was introduced. I can only say that this issue is easy to reproduce, and that this PR fixes it.

mfurtak (approved):

tmm1:
@KrauseFx :heart: thanks for digging in. I knew it would be a doozy.

PR #9027, opened 2017-04-29
Replace stub with allow to fix deprecation warnings

nafu:
### Checklist

### Motivation and Context

There is Deprecation Warnings about using `stub`.

### Description

When you run:
```
be rspec spaceship/spec/test_flight/client_spec.rb
```

The message are shown without this patch.
This patch remove below warnings.

```
Deprecation Warnings:

Using `stub` from rspec-mocks' old `:should` syntax without explicitly
enabling the syntax is deprecated. Use the new `:expect` syntax or
explicitly enable `:should` instead. Called from
/Users/nafu/Projects/fastlane/spaceship/spec/test_flight/client_spec.rb:32:in
`block (3 levels) in <top (required)>'.

If you need more of the backtrace for any of these deprecations to
identify where to make the necessary changes, you can configure
`config.raise_errors_for_deprecations!`, and it will turn the
deprecation warnings into errors, giving you the full backtrace.

1 deprecation warning total
```

### DIFF
--- a/spaceship/spec/test_flight/client_spec.rb
+++ b/spaceship/spec/test_flight/client_spec.rb
@@ -29,7 +29,7 @@ describe Spaceship::TestFlight::Client do
   context '#handle_response' do
     it 'handles successful responses with json' do
       response = double('Response', status: 200)
-      response.stub(:body).and_return({ 'data' => 'value' })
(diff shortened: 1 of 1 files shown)


mpirri (approved):
Thanks for catching this and making the change!

nafu:
@mpirri Thanks 🚀

Issue #8969, opened 2017-04-22
Scan failing for macOS target


tiferrei:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
fastlane fails to run scan for macOS target, even though it runs perfectly (and passes tests) on Xcode. iOS target (`fastlane ios test`) runs perfectly. The issue seems to be incorrect simulator picked, I tried adding `device: "",` to the mac tests and it defaults to the iOS SDK, that would explain this: `No architectures to compile for (ARCHS=x86_64, VALID_ARCHS=armv7s armv7 arm64).` However, I do not know which device to pick for macOS targets, I do not want to put my machine name in it because this would be running in multiple machines. The code is [open source](https://github.com/tiferrei/UCLKit) just in case someone needs to take a look at the xcode project.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

–> `bundle exec fastlane mac test --capture_output`
### Captured Output

Command Used: `mac test`
<details><summary>Output/Log</summary>

```

DEBUG [2017-04-23 00:06:18.18]: Checking if there are any plugins that should be loaded...
INFO [2017-04-23 00:06:18.19]: -------------------------------------------------
INFO [2017-04-23 00:06:18.19]: --- Step: Verifying required fastlane version ---
INFO [2017-04-23 00:06:18.19]: -------------------------------------------------
INFO [2017-04-23 00:06:18.19]: Your fastlane version 2.28.3 matches the minimum requirement of 2.14.2  ✅
INFO [2017-04-23 00:06:18.20]: ------------------------------
INFO [2017-04-23 00:06:18.20]: --- Step: default_platform ---
INFO [2017-04-23 00:06:18.20]: ------------------------------
INFO [2017-04-23 00:06:18.20]: Driving the lane 'mac test' 🚀
INFO [2017-04-23 00:06:18.20]: --------------------------------
INFO [2017-04-23 00:06:18.20]: --- Step: clear_derived_data ---
INFO [2017-04-23 00:06:18.20]: --------------------------------
INFO [2017-04-23 00:06:18.20]: Derived Data path located at: /Users/tiferrei/Library/Developer/Xcode/DerivedData
INFO [2017-04-23 00:06:18.27]: Successfully cleared Derived Data ♻️
[...]

PR #9032, opened 2017-05-01
Remove mention of macOS from main README

KrauseFx:
Most tools actually don't really support macOS, so we shouldn't list it in our docs as supported platform. Even though partly it might work, it's not our main focus for now, as we want to focus on providing the best tooling for mobile

Triggered by https://github.com/fastlane/fastlane/issues/8969


taquitos (approved):
👍

PR #9036, opened 2017-05-02
[spaceship] add methods to create/edit developer response

hjanuschka:
Introduces methods to create and edit dev. responses.

```ruby
require "spaceship"
app = Spaceship::Application.find("krone.at-fivexfive")


# get all reviews from AT
reviews = app.ratings.reviews("AT")

# iterate over each review - thx the 4+ and apologize all below
reviews.each do | review |
    if review.rating >= 4
      response_message = "Many Thx for your kind feedback and the awesome rating 🎉"
   else
     response_message = "We are really sorry 💔 that you have troubles with the app - feel free to contact us directly at app@domain.com - to get technical assistance!"
    end
     # which not yet have gotten a dev response
     unless review.responded?
         # respond to this user
        review.developer_response.create!(response_message)
     end
end

# here is a sample of editing a existing developer response
if reviews.first.responded?
     reviews.first.developer_response.update!("First you where awesome 👏 but now you are 🚀")
end

```

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -323,7 +323,7 @@ module Spaceship
       parse_response(r, 'data')
     end
 
-    def create_developer_response(app_id: nil, platform: nil, review_id: nil, response: "")
+    def create_developer_response!(app_id: nil, platform: nil, review_id: nil, response: "")
       data = {
         responseText: response,
         reviewId: review_id
@@ -336,7 +336,7 @@ module Spaceship
       parse_response(r, 'data')
     end
 
-    def update_developer_response(app_id: nil, platform: "ios", review_id: nil, response_id: nil, response: "")
+    def update_developer_response!(app_id: nil, platform: "ios", review_id: nil, response_id: nil, response: "")
       data = {
         responseText: response
       }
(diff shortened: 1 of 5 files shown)


spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -323,6 +323,38 @@ def get_rating_summary(app_id, platform, versionId = '')
       parse_response(r, 'data')
     end
 
+    def create_developer_response!(app_id: nil, platform: "ios", review_id: nil, response: nil)
+      raise "app_id is required" unless app_id
+      raise "review_id is required" unless review_id
+      raise "response is required" unless response
```
snatchev: 👍 on asserting the required parameters.

Perhaps we can move https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/test_flight/client.rb#L191-L211 into `Spaceship::Client` so that it can be re-used here.

hjanuschka: thats a good point 👍  - not sure if i should mix this into this PR?
as @KrauseFx mentioned that there might be concerns about supporting dev response posting in general.

i think we should decide that first.

KrauseFx:
Yes, we have to put developer responses to reviews on hold for now.

dzzh:
Could you please briefly describe the concerns you have about adding this feature or point me to some discussion?

KrauseFx:
@hjanuschka Please update the PR title to not include `stuff`. 
@dzzh Sure! Basically this would allow developers to automatically respond to app reviews, meaning lots of developers would use it to reply with very generic responses about contacting the support, or thanking the user for the review. This is probably something Apple doesn't want developers to do. With _fastlane_ the goal is to automate the app deployment, from code signing, to screenshots, to uploading the binary.

dzzh:
Ok, it's clear. Thanks a lot for quick feedback.

getaaron:
@KrauseFx How would you feel about keeping this but maybe add in some rate limiting? I was thinking of using this to integrate reviews into SalesForce so that the support team could individually respond to any review as a support ticket.

KrauseFx:
@getaaron Not sure if a local rate limit would avoid this use-case I'd say.

ohayon:
Hey everyone. I am going to go ahead and close this PR because we want to be very careful not to step on Apple's toes. Apologies for any inconvenience, feel free to continue discussion! 🚀

PR #9046, opened 2017-05-02
Enables App Managers to be able to add external testers to apps

taquitos:
### Checklist

### Motivation and Context
Only admins can currently add testers, with this patch, we can now add testers while using an account that only has the "app manager" role

### DIFF
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -10,8 +10,11 @@ module Pilot
       app = find_app(app_filter: config[:apple_id] || config[:app_identifier])
       UI.user_error!("You must provide either a Apple ID for the app (with the `:apple_id` option) or app identifier (with the `:app_identifier` option)") unless app
 
-      tester = find_or_create_tester(email: config[:email], first_name: config[:first_name], last_name: config[:last_name], app: app)
-
+      tester = find_app_tester(email: config[:email], app_id: app.apple_id)
+      tester ||= create_tester(email: config[:email],
+                          first_name: config[:first_name],
+                           last_name: config[:last_name],
+                                 app: app)
       begin
         groups = Spaceship::TestFlight::Group.add_tester_to_groups!(tester: tester, app: app, groups: config[:groups])
         if tester.kind_of?(Spaceship::Tunes::Tester::Internal)
@@ -95,7 +98,7 @@ module Pilot
       nil
     end
 
-    def find_or_create_tester(email: nil, first_name: nil, last_name: nil, app: nil)
+    def find_app_tester(email: nil, app_id: nil)
       current_user = Spaceship::Members.find(Spaceship::Tunes.client.user)
       if current_user.admin?
         tester = Spaceship::Tunes::Tester::Internal.find(email)
@@ -104,16 +107,23 @@ module Pilot
         unless app
           UI.user_error!("Account #{current_user.email_address} is only an 'App Manager' and therefore you must also define what app this tester (#{email}) should be added to")
         end
-        tester = Spaceship::Tunes::Tester::Internal.find_by_app(app.apple_id, email)
-        tester ||= Spaceship::Tunes::Tester::External.find_by_app(app.apple_id, email)
+        tester = Spaceship::Tunes::Tester::Internal.find_by_app(app_id, email)
+        tester ||= Spaceship::Tunes::Tester::External.find_by_app(app_id, email)
       else
         UI.user_error!("Account #{current_user.email} doesn't have a role that is allowed to administer app testers, current roles: #{current_user.roles}")
         tester = nil
       end
 
       if tester
-        UI.success("Existing tester #{email}")
-      elsif current_user.admin?
+        UI.success("Found existing tester #{email}")
+      end
+
+      return tester
+    end
+
+    def create_tester(email: nil, first_name: nil, last_name: nil, app_id: nil, app: nil)
+      current_user = Spaceship::Members.find(Spaceship::Tunes.client.user)
+      if current_user.admin?
         tester = Spaceship::Tunes::Tester::External.create!(email: email,
                                                        first_name: first_name,
                                                         last_name: last_name)
@@ -123,8 +133,8 @@ module Pilot
                                                           first_name: first_name,
                                                            last_name: last_name,
                                                                email: email)
-        tester = Spaceship::Tunes::Tester::External.find_by_app(app.apple_id, email)
-        UI.success("Successfully added tester: #{email} to your app #{app.name}")
+        tester = Spaceship::Tunes::Tester::External.find_by_app(app_id, email)
+        UI.success("Successfully added tester: #{email} to app: #{app.name}")
       end
 
       return tester
--- a/pilot/lib/pilot/tester_manager.rb
+++ b/pilot/lib/pilot/tester_manager.rb
@@ -6,15 +6,16 @@ module Pilot
   class TesterManager < Manager
     def add_tester(options)
       start(options)
-
       app = find_app(app_filter: config[:apple_id] || config[:app_identifier])
       UI.user_error!("You must provide either a Apple ID for the app (with the `:apple_id` option) or app identifier (with the `:app_identifier` option)") unless app
 
-      tester = find_app_tester(email: config[:email], app_id: app.apple_id)
-      tester ||= create_tester(email: config[:email],
-                          first_name: config[:first_name],
-                           last_name: config[:last_name],
-                                 app: app)
+      tester = find_app_tester(email: config[:email], app: app)
+      tester ||= create_tester(
+        email: config[:email],
+        first_name: config[:first_name],
+        last_name: config[:last_name],
+        app: app
+      )
       begin
         groups = Spaceship::TestFlight::Group.add_tester_to_groups!(tester: tester, app: app, groups: config[:groups])
         if tester.kind_of?(Spaceship::Tunes::Tester::Internal)
@@ -98,7 +99,7 @@ module Pilot
       nil
     end
 
-    def find_app_tester(email: nil, app_id: nil)
+    def find_app_tester(email: nil, app: nil)
       current_user = Spaceship::Members.find(Spaceship::Tunes.client.user)
       if current_user.admin?
         tester = Spaceship::Tunes::Tester::Internal.find(email)
@@ -107,8 +108,8 @@ module Pilot
         unless app
           UI.user_error!("Account #{current_user.email_address} is only an 'App Manager' and therefore you must also define what app this tester (#{email}) should be added to")
         end
-        tester = Spaceship::Tunes::Tester::Internal.find_by_app(app_id, email)
-        tester ||= Spaceship::Tunes::Tester::External.find_by_app(app_id, email)
+        tester = Spaceship::Tunes::Tester::Internal.find_by_app(app.apple_id, email)
+        tester ||= Spaceship::Tunes::Tester::External.find_by_app(app.apple_id, email)
       else
         UI.user_error!("Account #{current_user.email} doesn't have a role that is allowed to administer app testers, current roles: #{current_user.roles}")
         tester = nil
@@ -121,7 +122,7 @@ module Pilot
       return tester
     end
 
-    def create_tester(email: nil, first_name: nil, last_name: nil, app_id: nil, app: nil)
+    def create_tester(email: nil, first_name: nil, last_name: nil, app: nil)
       current_user = Spaceship::Members.find(Spaceship::Tunes.client.user)
       if current_user.admin?
         tester = Spaceship::Tunes::Tester::External.create!(email: email,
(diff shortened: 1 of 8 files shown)


pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
     end
 
-    def find_or_create_tester(email: nil, first_name: nil, last_name: nil)
-      tester = Spaceship::Tunes::Tester::Internal.find(config[:email])
-      tester ||= Spaceship::Tunes::Tester::External.find(config[:email])
+    def find_or_create_tester(email: nil, first_name: nil, last_name: nil, app: nil)
+      current_user = Spaceship::Members.find(Spaceship::Tunes.client.user)
+      tester = nil
+
+      if current_user.roles.include?("admin")
```
ohayon: Any chance this could be a `admin?` method on the `Spaceship::Member` object?

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
+        tester = Spaceship::Tunes::Tester::Internal.find(email)
+        tester ||= Spaceship::Tunes::Tester::External.find(email)
+      elsif current_user.roles.include?("appmanager")
```
ohayon: Similarly, could this be a `app_manager?` method on the `Member` object?

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
+        raise "Account #{current_user.email_address} is only an 'App Manager' and therefore you must also define what app this tester (#{email}) should be added to" unless app
```
ohayon: This should be a `UI.user_error!("...` call I think. (if `UI` is available)

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
+        tester = Spaceship::Tunes::Tester::Internal.find_by_app(app.apple_id, email)
+        tester ||= Spaceship::Tunes::Tester::External.find_by_app(app.apple_id, email)
+      else
+        raise "Account #{current_user.email} doesn't have a role that is allowed to administer app testers, current roles: #{current_user.roles}"
```
ohayon: Same for this about `UI.user_error!`

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
 
       if tester
-        UI.success("Existing tester #{tester.email}")
-      else
-        tester = Spaceship::Tunes::Tester::External.create!(email: config[:email],
-                                                            first_name: config[:first_name],
-                                                            last_name: config[:last_name])
-        UI.success("Successfully added tester: #{tester.email} to your account")
+        UI.success("Existing tester #{email}")
+      elsif current_user.roles.include?("admin")
```
ohayon: `admin?`

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
+        tester = Spaceship::Tunes::Tester::External.create!(email: email, first_name: first_name, last_name: last_name)
+        UI.success("Successfully added tester: #{email} to your account")
+      elsif current_user.roles.include?("appmanager")
```
ohayon: `app_manager`?

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
...
+        Spaceship::TestFlight::Base.client.create_app_level_tester(app_id: app.apple_id, first_name: first_name, last_name: last_name, email: email)
```
ohayon: So what is this doing exactly? Because the tester needs to be added to a group in order to be able to test an app right?

ohayon: With that said, I think the thing to do is make sure that a group is available for it to be added to.

ohayon (dismissed):

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -112,15 +112,21 @@ def add_group_to_build(app_id: nil, group_id: nil, build_id: nil)
     ##
 
     def post_tester(app_id: nil, tester: nil)
-      assert_required_params(__method__, binding)
+      return create_app_level_tester(app_id: app_id,
```
ohayon: I think this could live on `Spaceship::TestFlight::Tester` rather than adding this method to the client, that way the client is really just an interface into the iTC API rather than having semantics different from the request specs.

pilot/lib/pilot/tester_manager.rb
ohayon: This could be something like 
```
test_flight_tester = Spaceship::TestFlight::Tester.new(attrs)
test_flight_tester.create!
...
```

ohayon: But this is all ugly because we didnt make a clean cut over from the Tunes namespaced testers because those are still represented in the Users and Roles view of iTC

pilot/lib/pilot/tester_manager.rb
KrauseFx: I like the idea too @ohayon 👍

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,35 @@ def find_app(app_filter: nil)
       nil
     end
 
-    def find_or_create_tester(email: nil, first_name: nil, last_name: nil)
-      tester = Spaceship::Tunes::Tester::Internal.find(config[:email])
-      tester ||= Spaceship::Tunes::Tester::External.find(config[:email])
+    def find_or_create_tester(email: nil, first_name: nil, last_name: nil, app: nil)
+      current_user = Spaceship::Members.find(Spaceship::Tunes.client.user)
+      tester = nil
```
KrauseFx: You don't need to init this with `nil`, if you set the value inside an `if/else` block

pilot/lib/pilot/tester_manager.rb
KrauseFx: I'd prefer the `unless app` on the line before, as it's really hard to see, since the line is so long already.

pilot/lib/pilot/tester_manager.rb
KrauseFx: I don't think we're supposed to call `client` here, right @ohayon?

KrauseFx (commented):
Great work @taquitos, this was a pretty tricky task. Glad you find a good solution to solve the problem 👍

ohayon:
I agree with @KrauseFx this is awesome work for sure! 

@KrauseFx - you're right, @taquitos and i talked in chat about putting those calls to the client inside of `Spaceship::TestFlight::Tester` instead of calling it directly

pilot/lib/pilot/tester_manager.rb
taquitos: LOVE IT

pilot/lib/pilot/tester_manager.rb
taquitos: 👍

pilot/lib/pilot/tester_manager.rb
taquitos: 💯

pilot/lib/pilot/tester_manager.rb
taquitos: 💯

pilot/lib/pilot/tester_manager.rb
taquitos: 👍

pilot/lib/pilot/tester_manager.rb
taquitos: 👍

spaceship/lib/spaceship/test_flight/client.rb
taquitos: I think having this api in the client makes sense, it's the same business logic that was there, I just added a new API method that allows for passing in separate args instead of just a Tester object and then I had the old API (which has a Tester object as the arg) just call into this new API. So basically, it's just a proxy to the old logic. I do think calling the client directly is bad form as you pointed out, and I should probably have a method in `Spaceship::TestFlight::Tester` proxy to the client instead.

pilot/lib/pilot/tester_manager.rb
taquitos: 👍 to that @KrauseFx I shouldn't be calling client here.

pilot/lib/pilot/tester_manager.rb
taquitos: Agreed! 💯

taquitos:
I ended up putting the call to `create_app_level_tester` in `Spaceship::TestFlight::Tester` instead of `Spaceship::Tunes::Tester` because that logic already existed in it's client (the API that exists in `Spaceship::TestFlight::Client` was just renamed and the old name/parameters proxied to the new name). We could move that logic from `Spaceship::TestFlight::Client` to `Spaceship::Tunes::Client` but I think there's a larger question around "Where should all this stuff go? `Tunes::Client` or `TestFlight::Client` " Thoughts?

KrauseFx:
Also adding @snatchev to the conversation

spaceship/lib/spaceship/test_flight/client.rb
snatchev: I like the change below where the client does not need to know about testers, and just deals with parameters or data hashes, instead of model objects.

I would also agree that it should be the responsibility of `Spaceship::TestFlight::Tester` to call the method with the correct parameters. The proxy method is be unnecessary. I assume this will require changing some tests 👍

spaceship/lib/spaceship/tunes/member.rb
```diff
@@ -24,6 +24,14 @@ def roles
...
       end
 
+      def admin?
+        roles.include?("admin")
+      end
+
+      def app_manager?
+        roles.include?("appmanager")
+      end
+
```
snatchev: `roles` looks like they could be an enumeration. Let's have them extracted like this: https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/test_flight/build.rb#L67-L73

pilot/lib/pilot/tester_manager.rb
```diff
@@ -91,21 +95,41 @@ def find_app(app_filter: nil)
       nil
     end
 
-    def find_or_create_tester(email: nil, first_name: nil, last_name: nil)
-      tester = Spaceship::Tunes::Tester::Internal.find(config[:email])
-      tester ||= Spaceship::Tunes::Tester::External.find(config[:email])
+    def find_or_create_tester(email: nil, first_name: nil, last_name: nil, app: nil)
```
snatchev: This method is quite long with lots of branching. I think it would be quite a pain to test. I think start writing some tests for it, it would uncover how we can decompose it into smaller pieces.

snatchev (dismissed):
Just a few comments. Most of all, would love to see some tests added for `tester_manager.rb`

pilot/lib/pilot/tester_manager.rb
taquitos: Totes agree. I started to decompose it early and then stuck it all back together when it stopped working. I think it's totally a good thing to do though.

spaceship/lib/spaceship/tunes/member.rb
taquitos: 👍

ahmedgharib:
Hey @taquitos, will this fix also allow non admins to remove testflight users?

pilot/lib/pilot/tester_manager.rb
```diff
@@ -6,19 +6,27 @@ module Pilot
...
         if tester.kind_of?(Spaceship::Tunes::Tester::Internal)
           UI.success("Successfully added tester to app #{app.name}")
         else
-          group_names = groups.map(&:name).join(", ")
-          UI.success("Successfully added tester to app #{app.name} in group(s) #{group_names}")
+          if config[:groups]
+            group_names = groups.map(&:name).join(", ")
+            UI.success("Successfully added tester to group(s): #{group_names} in app: #{app.name}")
+          else
+            UI.success("Successfully added tester to the default tester group in app: #{app.name}")
```
KrauseFx: Where is that being done? Can you add a comment where it explains where that is?

pilot/spec/tester_manager_spec.rb
```diff
@@ -104,5 +63,191 @@
...
+        expect(Spaceship::TestFlight::Tester).to_not receive(:create_app_level_tester)
+        expect(Spaceship::TestFlight::Group).to receive(:add_tester_to_groups!).and_return([custom_tester_group])
+        expect(FastlaneCore::UI).to receive(:success).with('Successfully added tester to group(s): Test Group in app: My Fake App')
+
+        tester_manager.add_tester(default_add_tester_options_with_group)
+      end
+    end
+
+    # Data and mocks
+    let(:tester_manager) { Pilot::TesterManager.new }
```
KrauseFx: In the _fastlane_ code base we usually have all `let` blocks at the top of the `_spec` files 👍

KrauseFx (approved):
So beautiful, more tests than code ❤️

pilot/spec/tester_manager_spec.rb
taquitos: Ohhh, will fix. I like having that stuff at the bottom so I can see the test logic. That's usually what I care about the most.

pilot/lib/pilot/tester_manager.rb
taquitos: 👍

taquitos:
@snatchev and @ohayon, I've updated to add many tests and address all the comments. Let me know if there is anything else I can do to make this 💯 👍

andydrizen:
Thanks for this!

snatchev:
@taquitos looks good. Great tests!

PR #9069, opened 2017-05-04
Show line when password was fetched from env variable

KrauseFx:
Print out a line of information when password is fetched from environment to avoid user confusion


nafu (approved):
Looks good to me 👍

PR #9070, opened 2017-05-04
Improve deliver setup detection

KrauseFx:
If a user has deliver setup without a Deliverfile, and has a `metadata` folder, until now we didn't find it when seeing if deliver is setup for the current project.

### DIFF
--- a/deliver/lib/deliver/commands_generator.rb
+++ b/deliver/lib/deliver/commands_generator.rb
@@ -28,6 +28,7 @@ module Deliver
       return available_options
     end
 
+    # rubocop:disable Metrics/PerceivedComplexity
     def run
       program :name, 'deliver'
       program :version, Fastlane::VERSION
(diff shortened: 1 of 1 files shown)


hjanuschka (approved):

deliver/lib/deliver/commands_generator.rb
```diff
@@ -28,6 +28,7 @@ def deliverfile_options(skip_verification: false)
       return available_options
     end
 
+    # rubocop:disable Metrics/PerceivedComplexity
```
hjanuschka: is this required? too many nested if's? or why is it firing?

KrauseFx: `rubocop` told me about it

hjanuschka: ok seems to be too many "if" statements even if they are not nested. 😵

Issue #9076, opened 2017-05-05
Better log message for "Due to the limit of the Google Play API supply can't download your existing screenshots..."


ndbroadbent:
Same as #3375

I was confused by this message, and I thought there might be something wrong with how I set up the account (something to do with access limits). I agree with @mingle2 and @krispypen in #3375.

"limit" needs to be changed to "limitation", to make it clear that it has nothing to do with API access rate limiting. I think the message should also make it clear that it's impossible to download screenshots from the Google Play API, and that you can't do anything about it. Perhaps something like:

> Due to a limitation of the Google Play API, there is no way to download your existing screenshots. Please copy your screenshots into fastlane/metadata/android/en-US/images/

mfurtak:
Hi @ndbroadbent - that sounds like a good change to me. I'd be happy to review such a PR if you want to make one :+1:

ndbroadbent:
@mfurtak ok thanks, I opened a PR: #9083

mfurtak:
Looks like it got merged - thanks! 🎉

KrauseFx:
Thanks for this 👍

PR #9083, opened 2017-05-05
Better log message for supply not being able to download screenshots

ndbroadbent:
See: #9076

I think this log message would clear up some confusion about the API limitation.


KrauseFx (approved):
Nice improvement, thanks 👍

PR #9085, opened 2017-05-05
Fix iTunesConnect error when not setting changelog

mpirri:
### Checklist

When running `pilot` without a changelog, feedback email, or app description and no beta information set on iTunesConnect, iTC was returning an error. This fixes that issue.

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -121,7 +121,7 @@ module Pilot
     end
 
     def should_update_build_information(options)
-      options[:changelog].to_s.length
+      options[:changelog].to_s.length > 0
     end
 
     def should_update_app_test_information(options)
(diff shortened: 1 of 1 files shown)


pilot/lib/pilot/build_manager.rb
```diff
@@ -121,7 +121,7 @@ def describe_build(build)
     end
 
     def should_update_build_information(options)
```
KrauseFx: If you have some extra minutes, maybe you can update the method name to follow the Ruby convention of ending with a question mark for methods that return boolean values => `should_update_build_information?` 👍

KrauseFx (approved):
Looks great, added one comment that would be a separate PR 👍

ohayon (approved):
I agree with @KrauseFx 's comment, but this is 💯  good find!

KrauseFx (approved):
🏣

PR #9089, opened 2017-05-05
Fix accessing fastlane_core from within spaceship

KrauseFx:
Also adds tests so that never happens again

with the recent TestFlight changes, multiple `UI.something` calls went into _spaceship_ causing crashes if that part of the code is reached. Currently _spaceship_ doesn't have access to _fastlane_core_. I'm updating our build scripts to automatically detect those errors in the future :+1:

### DIFF
--- a/spaceship/lib/spaceship/test_flight/group.rb
+++ b/spaceship/lib/spaceship/test_flight/group.rb
@@ -84,7 +84,7 @@ module Spaceship::TestFlight
       if groups.nil?
         default_external_group = app.default_external_group
         if default_external_group.nil?
-          UI.user_error!("The app #{app.name} does not have a default external group. Please make sure to pass group names to the `:groups` option.")
+          raise "The app #{app.name} does not have a default external group. Please make sure to pass group names to the `:groups` option."
         end
         test_flight_groups = [default_external_group]
       else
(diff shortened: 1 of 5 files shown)


taquitos (approved):
🤦‍♂️  This is great @KrauseFx, thank you! 🐑 it!

spaceship/lib/spaceship/test_flight/group.rb
```diff
@@ -84,15 +84,15 @@ def self.perform_for_groups_in_app(app: nil, groups: nil, &block)
       if groups.nil?
         default_external_group = app.default_external_group
         if default_external_group.nil?
-          UI.user_error!("The app #{app.name} does not have a default external group. Please make sure to pass group names to the `:groups` option.")
+          raise "The app #{app.name} does not have a default external group. Please make sure to pass group names to the `:groups` option."
```
mpirri: My feeling is that we need to be sure that we surface this failure as a user-error of some sort up the chain. Failing to include a group name or mistyping a group name (as in line 95 below) is not really a crash, yet if we do not rescue this error and re-raise as a user error, it will be logged as a crash by _fastlane_'s `runner.rb`.

mpirri (approved):

PR #9090, opened 2017-05-05
Fix spaceship launcher client login for push certificate

atreat:
🔑

### Checklist

### Motivation and Context

I found what I believe is a bug when using `Spaceship::Launcher` and trying to create push certificates for an app. What I've found is that the call to `launcher.certificate.development_push.create!` would throw a RuntimeError. 

After inspecting the code I realized this is because the `Spaceship::Certificate.create!` method is  checking for the app in the portal, to do this it calls out to the `Spaceship::App` portal client. Because this method does not pass the client along, and the default `Spaceship::Portal.client` is nil because I'm using `Spaceship::Launcher`, I receive a RuntimeError telling me to log in. This would only affect Push Certificates as they require the app to exist.

#### Related Github Issue
I didn't find any open issues and am going to forego opening one myself as I believe the fix is pretty small (and more descriptive of the issue)

#### How I tested my changes
I tested my changes by patching my local fastlane gem with the fix. Unfortunately, I tried to add a spec in launcher_spec to describe what I was seeing but I'm running into some issues with WebMock. Rspec will fail for this PR currently and am **looking for help to resolve this testing issue**. 

What I believe is happening is the `Spaceship::Client#request` method is encoding the body (headers, params) of POST requests. This doesn't seem to match the stubbed request bodies. I'm unsure why this doesn't affect any other POST request specs.

### Description

I've added a call to `with_client` to the `create!` method. This passes along the client object from the `Spaceship::Portal::Certificate`. 

Another approach I tried was to set the `Spaceship::Portal.client` inside the `Spaceship::Launcher` initializer, but this is incorrect because it'd break the ability to have two spaceships running side by side.

### DIFF
--- a/spaceship/lib/spaceship/portal/certificate.rb
+++ b/spaceship/lib/spaceship/portal/certificate.rb
@@ -287,7 +287,7 @@ module Spaceship
 
           # look up the app_id by the bundle_id
           if bundle_id
-            app = Spaceship::App.find(bundle_id)
+            app = Spaceship::App.set_client(client).find(bundle_id)
             raise "Could not find app with bundle id '#{bundle_id}'" unless app
             app_id = app.app_id
           end
--- a/spaceship/spec/launcher_spec.rb
+++ b/spaceship/spec/launcher_spec.rb
@@ -59,6 +59,18 @@ describe Spaceship do
         clean_launcher.login(username, password)
         expect(clean_launcher.provisioning_profile.all.count).to eq(6)
       end
+
+      it "shouldn't fail if trying to create new apns_certificate before app is invoked" do
+        bundle_id = 'net.sunapps.151'
+
+        clean_launcher = Spaceship::Launcher.new
+        clean_launcher.login(username, password)
+        csr = PortalStubbing.adp_read_fixture_file('certificateSigningRequest.certSigningRequest')
+
+        expect {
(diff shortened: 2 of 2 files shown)


atreat:
I signed it!

KrauseFx (approved):
Nice catch, thanks 👍

KrauseFx:
Oh, tests are failing, can you fix them? Thanks 👍

atreat:
Hey @KrauseFx, I was having trouble with properly mocking the certificate signing request. I have it written to check to make sure an error is not thrown when using the Portal stubbed certificate signing request, but it seems there is a difference in how WebMock and the `Spaceship::Client.request` encodes the HTTP body. I believe this is causing the mismatch described in the failing spec (its the test I've added).

```
expected no Exception, got #<WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: POST ...esources/controllers/login_cntrl.js")
```

I was wondering if anyone on the fastlane team could suggest the proper mechanism to stub out the certificate request? I'll keep reading through to see if other specs are doing this better than my current approach.

Thanks for the approval!

atreat:
Nevermind! There's a good example in the portal/certificate_spec that I somehow overlooked in my initial attempt.

snatchev:
LGTM

PR #9096, opened 2017-05-07
Update lane scope whenever changing lanes

tommeier:
### Checklist

### Motivation and Context

I'm not sure if this is an issue for others, or if it would cause more problems for the majority of cases, but this is the scenario I've faced today (feel free to close and :fire: this PR - I have a strong feeling this was locked for a reason).

For one client, we have to package two completely different apps at the same time, with the majority of the same code, but two targets. As such, I've wrapped it into one lane `client_release` which calls two other lanes `build_target_one` + `build_target_two`. The wrapping lane then can use all the usual shared lane contexts from the likes of `gym` to manage what to next.

Unfortunately, when switching to `build_target_one` the lane context remains as the `client_release` lane, so any calls to things such as `CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier)` will fail, as it is looking for the default, and not using the `AppFile` scope for `build_target_one` that I would expect (this could be designed behaviour for a reason though).

I don't believe there is a way for private/sub lanes to scope `AppFileConfig` calls to the particular lane context it is being written in.

### Description

Whenever `Fastlane::Runner` assigns a new `current_lane` the scope should be altered for the ENV and the shared lane context. 

With this value updated on every assignment of `current_lane`, any dependencies such as `CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier)` will always use the `AppFile` reference relevent to the inner context of the lane (or private lane) it is in.

The drawbacks of this is any existing projects relying on the outer wrapping lane context for their `AppFile` references, this will not work, they will need additional references in the `AppFile` to default for even private lanes. I imagine this could be an issue? I'm not sure how to solve this, and really wanted to propose this solution and see if there were any other alternatives thought about.

In the meantime, my workaround is to be very specific in the sub/private lanes to specify every attribute and not rely on default values gleaned from `Appfile`.

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -1,13 +1,21 @@
 module Fastlane
   class Runner
     # Symbol for the current lane
-    attr_accessor :current_lane
+    attr_reader :current_lane
 
     # Symbol for the current platform
     attr_accessor :current_platform
 
     # @return [Hash] All the lanes available, first the platform, then the lane
     attr_accessor :lanes
+
+    # Update lane scope whenever current changes
+    def current_lane=(value)
+      @current_lane = value
+      ENV["FASTLANE_LANE_NAME"] = value.to_s
+      Actions.lane_context[Actions::SharedValues::LANE_NAME] = full_lane_name
+      @current_lane
+    end
 
     def full_lane_name
       [current_platform, current_lane].reject(&:nil?).join(' ')


KrauseFx:
Thanks for submitting this. I'll take a look at this PR over the next few days, as I'll have to make sure there are no unexpected side effects 👍

tommeier:
Thanks @KrauseFx , 

Just to be clear, it would definitely have issues for existing users (I'm just not sure how many, and whether its really a _bad_ thing). 

For example:

```
lane :client_release do
  build_target_one
  action_with_config
end

lane build_target_one do 
  action_with_config
end
```

So in the current world, when we run `fastalane client_release`, the lane switches to `build_target_one` but the lane context remains the same, so `action_with_config` triggers with the lane context of `client_release`. Meaning `Appfile` will load `client_release`. 

With the new changes this would scope to `build_target_one` when the lane switches.

To me, this makes sense, if the lane has switched, either full lane or private, I'd expect the scope to now be for that lane. This just may be a hiccup for existing users, who currently rely on the 'first' lane to be the context for the sub lanes. I guess we could make the transition easier, by perhaps making `AppFile` accept an array of lanes (duplicated config).

Before merging I can make some tests, but just wanted to clarify the potential 🐉 's for existing users.

ohayon:
Hey @tommeier we totally appreciate this PR and it is a good change. Unfortunately for now, this is going to break too many people's existing setups. I am going to close this PR, but please feel free to have discussion on here. Thanks again! 🚀

tommeier:
No worries @ohayon - is definitely a risky one for existing setups - I guess in future if more people hit this issue it could be a global config variable to allow slow migration to it (default: off).

PR #9102, opened 2017-05-07
[match] Add new `clone_branch_directly` option to clone faster

KrauseFx:
When an app developer used their main repo, and selected a different branch for match to store the certificates, it becomes really slow very fast if the `master` branch has bigger resources. In my case it took many minutes already.

This PR adds a parameter that tells match that it should just clone this one branch. Also we'll now show a message that tells users about this new option

screenshot 2017-05-08 01 24 54

### DIFF
--- a/match/lib/match/git_helper.rb
+++ b/match/lib/match/git_helper.rb
@@ -1,14 +1,23 @@
 module Match
   class GitHelper
-    def self.clone(git_url, shallow_clone, manual_password: nil, skip_docs: false, branch: "master", git_full_name: nil, git_user_email: nil)
+    def self.clone(git_url, shallow_clone, manual_password: nil, skip_docs: false, branch: "master", git_full_name: nil, git_user_email: nil, clone_branch_directly: false)
       return @dir if @dir
 
       @dir = Dir.mktmpdir
 
       command = "git clone '#{git_url}' '#{@dir}'"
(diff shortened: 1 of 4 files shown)


match/lib/match/git_helper.rb
```diff
@@ -1,14 +1,23 @@
...
+      if shallow_clone
+        command << " --depth 1 --no-single-branch"
+      elsif clone_branch_directly
+        command += " -b #{branch.shellescape} --single-branch"
+      end
 
       UI.message "Cloning remote git repo..."
+
+      if branch && !clone_branch_directly
+        UI.message("If cloning the repo takes too long, you can use the `clone_branch_directly` option in match.")
```
taquitos: 👍

taquitos (approved):
This is awesome, should cut down on clone time for lots of people! 🎉
Also really like the formatting changes you made so the method calls aren't just one long line 👍

KrauseFx:
Thanks @taquitos 👍 I was thinking if there is a way to automatically create the branch if it doesn't exist yet, but with the current architecture of the `git_helper` class, I don't think that's easily possible. Do you have a better idea of how we can solve this?

taquitos:
I think that's tricky. IMO, the current branch would need to be clean, then git checkout -b new_branch; git push origin new_branch;
That being said, making an assumption that the current branch is the correct branch to base the new branch off seems dangerous too.

PR #6991, opened 2016-11-14
Remove Shenzhen dependency

sebastianludwig:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:

- [x] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

This PR is based on the discussion in fastlane/shenzhen#11 and aims to remove the `Shenzhen` dependency.

The actions were pretty straight forward, all unit tests are green. However I haven't set up test accounts where possible yet to test if the actual upload still works as expected.

The `ipa` action is more interesting. It's deprecated at least since March. It uses [a command](https://github.com/fastlane/shenzhen/blob/master/lib/shenzhen/xcodebuild.rb#L65) which [hangs on Xcode 8](https://forums.developer.apple.com/thread/50372). Not sure if this affects everybody but the `ipa` action is currently definitely broken. I only did a quick search, but didn't find any open issues concerning the ipa action. 

This boils down to: How long do you wish to keep the deprecated action around and put more effort into maintaining it? With gym as a perfect(?) successor, I'd suggest to get rid of it. If you agree, I'd be happy to go forward with this PR.

PR #9135, opened 2017-05-11
[WIP] Remove babosa workaround since it has been fixed upstream

milch:
Since https://github.com/norman/babosa/pull/42 has been merged in Babosa, as soon as a new version of the gem is released this PR can be merged. 

This PR removes the workaround that we needed because of the unicode-display_width gem that was introduced in #6991 (see also https://github.com/janlelis/unicode-display_width/issues/4).

@mfurtak Could you look over this to make sure it takes care of everything, since you introduced the workaround in #7243?


hjanuschka:
awesome, got the notification today that babosa PR was merged, nice to see this hack being removed.! 🚀

spaceship/lib/spaceship/client.rb
```diff
@@ -3,7 +3,6 @@
 require 'faraday_middleware'
 require 'fastlane/version'
 require 'logger'
-require 'spaceship/babosa_fix'
```
mfurtak: Do we need to replace this with a `require 'babosa'` if we're taking out the fix class?

parndt:
Open source is fun.. I use fastlane for app deployments and needed to use `unicode-display_width` to determine the correct length of non-English strings, which is where I saw the bug filed against babosa which I help to maintain to merging the PR made a lot of sense. Enjoy, and thanks for fastlane! 👍

spaceship/lib/spaceship/client.rb
milch: Right, good point. Thanks!

KrauseFx:
Hey @milch, that's an awesome PR, thanks 👍 What's the best way to test it? To create Chinese apps, or apps with emojis using _produce_? I heard @hjanuschka has some Chinese apps on his account now.

hjanuschka:
@KrauseFx i have like 100+ chinese app's 😨 - but it totally would give it a try and create another one. but since the last babosa release is from 2015 i am not really sure if this will ever be released?

once done, please ping me - and i ll create another chinese app 🎉

KrauseFx:
Alright, let's leave this WIP and blocked for now, if the status doesn't change within the next few weeks, we should probably just close the PR for now

ohayon:
Hey everyone, looks like babosa is not pushing any updates these days. I am going to close this PR for now and we can update down the road if necessary. Thanks everyone!

PR #9138, opened 2017-05-11
Tell users about new SnapshotHelper.swift

KrauseFx:
- By incrementing the verson number, we notify people about the new version
- As mentioned in https://github.com/fastlane/fastlane/pull/8791#issuecomment-300692592

screenshot 2017-05-11 01 48 59


taquitos (approved):
🐑

PR #9153, opened 2017-05-12
Enhancing app_store_build_number action’s build number sort

### LINKED ISSUE #1178: latest_testflight_build_number issue
On iTunes Connect I have an app at version 3.0.2
I have two builds uploaded as 
`3.0.2.0`
`3.0.2.1`

When I run latest_testflight_build_number(version: 3.0.2)
the response I get back is
```
Fetching the latest build number for version 3.0.2
[07:58:20]: Latest upload is build number: 3
[07:58:20]: FETCHED BUILD NUMBER: 3
```

I was expecting to receive a `1` back for a build number, or is my understanding of the 4th digit different than the community?

sinoru:
### Checklist

### Motivation and Context

I'm using build number as x.y format, and app_store_build_number doesn't support that.
So... This will fixes #1178

### Description

I use ~~Gem::Version~~ custom type to sort build numbers as semver, after that it will convert to string.

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -5,6 +5,50 @@ module Fastlane
     end
 
     class AppStoreBuildNumberAction < Action
+      class BuildNumber
+        protected
+
+        attr_reader :build_numbers
+
+        public
+
+        VERSION_PATTERN = /^(\.?[0-9]+)+$/
+
+        def self.correct?(version)
+          (version.to_s !~ VERSION_PATTERN).nil?
+        end
+
+        def initialize(version)
+          unless self.class.correct?(version)
+            raise ArgumentError, "Malformed version number string #{version}"
+          end
+
+          @build_numbers = version.split(".").map { |s| Integer(s) }
+        end
+
+        def to_s
+          @build_numbers.join(".")
+        end
+
+        def to_i
+          @build_numbers.first
+        end
+
+        def ==(other)
+          self.to_s == other.to_s
+        end
+
+        def <=>(other)
+          return unless other.instance_of?(BuildNumber)
+
+          @build_numbers <=> other.build_numbers
+        end
+
+        def +(other)
+          return build_numbers.first + other
+        end
+      end
+
       def self.run(params)
         require 'spaceship'
 
@@ -39,7 +83,7 @@ module Fastlane
 
           begin
             train = app.build_trains[version_number]
-            build_nr = train.builds.map(&:build_version).map { |s| Gem::Version.new(s) }.sort.last.to_s
+            build_nr = train.builds.map(&:build_version).map { |s| BuildNumber.new(s) }.sort.last
           rescue
             UI.user_error!("could not find a build on iTC - and 'initial_build_number' option is not set") unless params[:initial_build_number]
             build_nr = params[:initial_build_number]
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -15,7 +15,7 @@ module Fastlane
         VERSION_PATTERN = /^(\.?[0-9]+)+$/
 
         def self.correct?(version)
-          (version.to_s !~ VERSION_PATTERN).nil?
+          version.to_s.match(VERSION_PATTERN)
         end
 
         def initialize(version)
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -30,6 +30,10 @@ module Fastlane
           @build_numbers.join(".")
         end
 
+        def to_f
+          @build_numbers.first(2).join(".").to_f
+        end
+
         def to_i
           @build_numbers.first
         end
@@ -45,7 +49,7 @@ module Fastlane
         end
 
         def +(other)
-          return build_numbers.first + other
+          return self.to_i + othser
         end
       end
 
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -12,7 +12,7 @@ module Fastlane
 
         public
 
-        VERSION_PATTERN = /^(\.?[0-9]+)+$/
+        VERSION_PATTERN = /^([0-9])+(\.[0-9]+)*$/
 
         def self.correct?(version)
           version.to_s.match(VERSION_PATTERN)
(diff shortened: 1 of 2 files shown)


sinoru:
@googlebot I signed it!

hjanuschka (changes requested):
@sinoru many thx for your contribution, this is super awesome 🚀  ❤️ .

can you please add a spec for the action so that it verify's that normal integer build numbers are still sorted correctly.

and non semver, non-integer builds atleast not crash?

let me know if you need help on anything.

sinoru:
@hjanuschka I would be happy to add a spec for the action, but I don't know how to do with action which should get information from iTunes Connect.. I can't find any other action specs to reference...

Anyway, This must be work in normal integer build numbers which is worked with Gem::Version which is used to solving gemspecs.

And according to [apple docs](https://developer.apple.com/library/content/technotes/tn2420/_index.html#//apple_ref/doc/uid/DTS40016603-CH1-NUMBERING_CONVENTIONS) build number must be integer or/and `dot` which should compliance semver(-ish).

sinoru:
Uhm.. On the other hand, Combination of `latest_testflight_build_number` + `increment_build_number` which will call `app_store_build_number` may be has issue due to now it returns string than integer.
e.g:
```ruby
increment_build_number({
  build_number: latest_testflight_build_number + 1
})
```

Should I return as integer if it can representable in integer, and otherwise return as string?

hjanuschka:
yeah this is why i thought a spec for the change would help.

i am right now using sorta this code on all my production apps.:

```
   last_tf = latest_testflight_build_number
   increment_build_number(build_number: last_tf + 1)
```


i have the following concerns:

  * return type should be INTEGER - otherwise current setups will definitily break, and i bet most of the users use integer based build numbers instead of semver style. so the regression seems to be not worth it.

  *  we'd need to find a way to convert a semver string (according to docs the simple ones like `x.y.z`- without the full semver features like `build` etc..) into a reliable integer? 


  * verify - via spec - that sorting with the help of `Gem::Version.new` does really work on simple-integers.


i really would love to see this merged, but we'd need to keep an eye on existing setups.
  

let me know if you'd need some help, i am trying to find a spec sample. but it'll take me some time :/

sinoru:
I thinks.. It is hard to represent semver as integer. According to Apple docs, `CFBundleVersion` can be any dots and integer combination in eight characters. So some people can use it as x.y.z.f like that..

So, Uhm..
And, even x.y can not be represented in integer.. It can only in float.. Uhm..

So I have some candidates to solve this problem:
1. If just integer return as integer, If version is float than return as float, If version only can presented in string, than return as string.
2. Get options to return always in string, like (string: true), and otherwise only return as integer.

How about those candidates?

hjanuschka:
@sinoru here is a sample on testing the action.:


`fastlane/spec/actions_specs/app_store_build_number_spec.rb`
```ruby
require 'spec_helper'

describe Fastlane::Actions::AppStoreBuildNumberAction do
  it "Should return latest build" do

    # Mock Builds
    build1 = OpenStruct.new
    build1.build_version = "3"

    build2 = OpenStruct.new
    build2.build_version = "2"

    builds = OpenStruct.new
    builds.builds = [build1, build2]

    # Mock App with Buildtrain
    spec_app = OpenStruct.new
    spec_app.build_trains = {
      "1.0" => builds
    }

    # Mock Spaceship
    expect(Spaceship::Tunes).to receive(:login).and_return(true)
    expect(Spaceship::Tunes).to receive(:select_team).and_return(true)
    expect(Spaceship::Application).to receive(:find).and_return(spec_app)

      app_store_build_number(username:'demo@demo.com', app_identifier: 'demo.app', live: false, version: '1.0')
    end").runner.execute(:test)

    expect(result).to eq(3)
  end
end

```


run it with: `rspec fastlane/spec/actions_specs/app_store_build_number_spec.rb` (with `bundle exec` if you are using bundler)

this should do it as a boilerplate to test new concepts 👍 




how about returning a custom type (shipped inside of app_store_build_number.rb?)
and check if it is a simple integer or a `x.y.z`  and provide `+` && `to_s` method for easy incrementing and outputing?

```ruby

module Fastlane
  module Actions
  class AppStoreBuildNumberType
      attr_reader :version
      attr_reader :semver
      def initialize(value)
        if value.include?(".")
          @version = value
          @type = :semver
        else
          @version = value
          @type = :integer
        end
      end
      def to_s
        @version.to_s
      end
      def +(other)
          if @type == :integer
            return @version.to_i+other
          else
            components = @version.split(".")
            components[-1] = components[-1].to_i+other
            @version = components.join(".")
            return @version
          end
      end
  end
end
end


# Create a Semver style AppStoreBuildNumberType
semver_style = Fastlane::Actions::AppStoreBuildNumberType.new("1.1.1")

# increment it by 2 
incremented = semver_style+2

# results in "1.1.3"
puts incremented


# Use the usual integer build names
simple_style = Fastlane::Actions::AppStoreBuildNumberType.new("100")

# increment by 3
incremented = simple_style+3

# results in "103"
puts incremented


```


let me know what you think.

KrauseFx:
Thanks for the review @hjanuschka 👍  Thanks for the PR @sinoru 👍

sinoru:
How about this?
I add custom type for that, which follows Apple's rule (only dots and number), convertible to int, string, can be sumed and can be sorted.

sinoru:
Additional fixing is done.

hjanuschka:
@sinoru ohh wow thats a huge thing! - still trying to fully understand it.

what is in the case there is a build `433`  (latest uploaded)  - and an exsiting version `4.4.4` (older upload) - would `433`be returend, or is that a case that does not work?

sinoru:
@hjanuschka 433 should be returned.
Actually, this split it by dots, compares as array.
e.g:
```ruby
irb(main):001:0> [[443], [4,4,4]].sort
=> [[4, 4, 4], [443]]
```

sinoru:
I added more specs just for Custom types that should works well.

ohayon:
Hey @sinoru - sorry I am late to this party, but would it be possible to map the values to `Gem::Version.new` rather than introducing a new class which does almost the same thing? 💎  🚀

sinoru:
@ohayon I'm sorry about I also late to respond.
Uhm.. Gem::Version can be used.... But for legacy, Lot's of people uses like:
```
build = app_store_build_number + 1
```


But, If return as Gem::Version, This will breaking changes, and should be used as:
```
build = app_store_build_number.to_s.to_i + 1
```
which is not that good.

ohayon:
Hmmmm @sinoru do you think we could subclass or monkey patch `Gem::Version` to add the addition function?

ohayon:
Hey @sinoru, just saw your comment on [this issue](https://github.com/fastlane/fastlane/issues/9392) related to this PR. Did you have a chance to consider my comment above?

ohayon:
Hey everyone, I am going to close this PR for now as it hasnt been active for quite some time. @sinoru - please feel free to continue discussion and/or reopen this or another PR, but I think it would be great to talk through the changes so that we can make sure we are all on the same page 😄 ! 🚀  🐙

PR #9156, opened 2017-05-12
Use new `build_failure!` method to report gym failures

mpirri:
These failures will not be counted by metrics or recorded by crash reporter

### Checklist

### Motivation and Context
Currently, code compilation errors are reported via `user_error!`, even though these sorts of errors are not actually the result of a user misusing _fastlane_ but just natural errors in software development lifecycle.  Since we do not want to report these sorts of errors to our crash reporter (and don't want to count them as user failures in metrics), we have introduced a new failure type to handle these exceptions without any sort of logging.

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -252,11 +252,13 @@ module Fastlane
             class_ref.run(arguments)
           end
         end
+      rescue \
+        FastlaneCore::Interface::FastlaneBuildFailure,    # build_failure!
+        FastlaneCore::Interface::FastlaneTestFailure => e # test_failure!
+        raise e
       rescue FastlaneCore::Interface::FastlaneError => e # user_error!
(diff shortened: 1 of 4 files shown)


ohayon (approved):
LGTM!

taquitos:
I think this is a great idea, I wonder if the method name could be improved to better communicate the intent behind this method. As I understand it, this is more for situations where we want to say "hey, this is not fastlane's fault, but rather, something that just comes up and the developer/xcode is probably doing something weird here". Maybe something even as verbose as `non_fastlane_related_error` or something like that. 
Thoughts?

mpirri:
We had a like 20 minute conversation trying to think up a good name. In the end we decided that, because we already had FastlaneTestFailure (used by `scan` for when a test fails), that it was okay to continue to use specific failure types for this exception.  I think if we start to get a lot more of these it might be worth combining them all into an exception type that is a bit more generic (and a LOT more hard to name!)

PR #9159, opened 2017-05-12
zip action did not return output path with zip if not given with one

joshdholtz:
`zip` action now will always return an output path with the `.zip` extension 😊

### DIFF
--- a/fastlane/lib/fastlane/actions/zip.rb
+++ b/fastlane/lib/fastlane/actions/zip.rb
@@ -4,9 +4,12 @@ module Fastlane
       def self.run(params)
         UI.message "Compressing #{params[:path]}..."
 
-        params[:output_path] ||= "#{params[:path]}.zip"
+        params[:output_path] ||= params[:path]
 
         absolute_output_path = File.expand_path(params[:output_path])
+
+        # Appends ".zip" if path does not end in ".zip"
+        absolute_output_path = absolute_output_path.gsub(/(?<!.zip)$/, ".zip")
 
         absolute_output_dir = File.expand_path("..", absolute_output_path)
         FileUtils.mkdir_p(absolute_output_dir)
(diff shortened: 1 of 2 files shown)


joshdholtz:
I signed it!

fastlane/lib/fastlane/actions/zip.rb
```diff
@@ -4,10 +4,13 @@ class ZipAction < Action
       def self.run(params)
         UI.message "Compressing #{params[:path]}..."
 
-        params[:output_path] ||= "#{params[:path]}.zip"
+        params[:output_path] ||= params[:path]
 
         absolute_output_path = File.expand_path(params[:output_path])
 
+        # Appends ".zip" if path does not end in ".zip"
+        absolute_output_path = absolute_output_path.gsub(/(?<!.zip)$/, ".zip")
```
hjanuschka: i love regex ❤️

hjanuschka (approved):
saw you fighting against that on stream (awesome btw) - LGTM 🚀

fastlane/lib/fastlane/actions/zip.rb
joshdholtz: An awesome viewer helped me with that one 😇

KrauseFx: Why do we use a regex here, instead of `absolute_output_path.end_with?(".zip")` or even `File.extname(absolute_output_path)`? 👍

KrauseFx (approved):
Nice work, just one comment 👍

fastlane/lib/fastlane/actions/zip.rb
joshdholtz: @KrauseFx It was the cleanest/more fun single line approach 😇  yolo

KrauseFx:
@joshdholtz can you submit a follow-up PR that removes the regex? It's a lot harder to read for developers 👍

Issue #9154, opened 2017-05-12
missing team id in non interactive mode still blocks


doapp-ryanp:
Commenting for https://github.com/fastlane-old/pem/issues/76 is locked so started a new issue here.

Running PEM 2.26.0 in a jenkins non-interactive environment.  The issue above was to request a new feature that allowed a non-interactive program to detect when team ID is not found, and to take corrective action.

I just ran into the problem in v2.26.0:

```
INFO [2017-05-11 14:56:53.68]: Starting login with user 'me@me.com'
Couldn't find team with ID 'XXXXXXXXXX'
Multiple teams found on the Developer Portal, please enter the number of the team you want to use: 
...
```
PEM then blocks my program as it is waiting for input.

I would like PEM to blow up in this scenario and have a return code so that I can take corrective action.  

The original commit to address this (https://github.com/fastlane-old/fastlane_core/commit/a5f24e46cc2168c46a9051c3b522663e01fc0ace) does not work. According to @KrauseFx

> The reason why this doesn't work yet is that spaceship doesn't have access to fastlane_core. However that's something we want to fix.

hjanuschka:
hey @doapp-ryanp 🤚 i did a PR -> #9160  maybe you can give it a try, it should now raise an error when team selection is triggered under non-interactive environments.

doapp-ryanp:
Thanks @hjanuschka I know very little about Ruby - where do all the .rb files live so that I can hand patch them and test?  running `which PEM` gives me `/Users/dbuilder/.rvm/rubies/ruby-2.2.2/bin/PEM'

hjanuschka:
@doapp-ryanp  you can try it with this helper script: https://gist.github.com/hjanuschka/3040fc6f371f46acff72a5e67f649b27


running it with `./try_pr.sh hjanuschka spaceship_interactive fastlane pem` - should run it from the PR branch.

doapp-ryanp:
@hjanuschka how do I roll it back? Don't want to test until I have a sure fire way to get back to where I was in case of problems.

hjanuschka:
@doapp-ryanp it doesn't install anything, the `.sh` scripts clones the repo into `tmp` - checks out the branch and runs it.

so if you run it with `fastlane pem` again it will run the real installed version.

doapp-ryanp:
@hjanuschka tried your PR, it crashes:

```
INFO [2017-05-15 09:25:06.62]: Sending crash report...
[...]

PR #9160, opened 2017-05-12
[spaceship] detect non-interactive terminals (team selection)

hjanuschka:
there is an  endless loop in spaceship when team selection is done, and running in non interactive mode. (aka CI) that almost all tools suffer.

this PR trys to fix this, by providing the 2 missing pieces (`ci?`, `interactive?`)

i know this 2 methods are simple copies from fastlane_core, but as we still have no inter dep. from core to spaceship and vice-a-verse - i think its ok to stay with the couple of duplicated lines. and solving the infinity loop.

can we assume that in fastlane_core, spaceship is available? (then we could remove the methods from fastlane_core, and call the `Spaceship::Client::UserInterface` - or are there any cases where `fastlane_core` is used without `spaceship`?

issues:
#9154, #2798

### DIFF
--- a/spaceship/lib/spaceship/portal/ui/select_team.rb
+++ b/spaceship/lib/spaceship/portal/ui/select_team.rb
@@ -32,14 +32,14 @@ module Spaceship
       #   ]
 
       def self.ci?
-        if Object.const_defined?("FastlaneCore")
+        if Object.const_defined?("FastlaneCore") && FastlaneCore.const_defined?("Helper")
           return FastlaneCore::Helper.ci?
         end
         return false
       end
 
       def self.interactive?
-        if Object.const_defined?("FastlaneCore")
+        if Object.const_defined?("FastlaneCore") && FastlaneCore.const_defined?("Helper")
           return FastlaneCore::Helper.interactive?
         end
         return true
@@ -85,7 +85,7 @@ module Spaceship
           teams.each_with_index do |team, i|
             puts "#{i + 1}) #{team['teamId']} \"#{team['name']}\" (#{team['type']})"
           end
-          raise "multiple_teams_no_interactive_shell"
+          raise "Multiple Teams found; unable to choose, terminal not ineractive!"
         end
 
         # User Selection
--- a/spaceship/lib/spaceship/portal/ui/select_team.rb
+++ b/spaceship/lib/spaceship/portal/ui/select_team.rb
@@ -31,6 +31,21 @@ module Spaceship
       #     {...}
       #   ]
 
+      def self.ci?
+        # Check for Jenkins, Travis CI, ... environment variables
+        ['JENKINS_HOME', 'JENKINS_URL', 'TRAVIS', 'CIRCLECI', 'CI', 'TEAMCITY_VERSION', 'GO_PIPELINE_NAME', 'bamboo_buildKey', 'GITLAB_CI', 'XCS'].each do |current|
+          return true if ENV.key?(current)
+        end
+        return false
+      end
+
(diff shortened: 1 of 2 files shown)


KrauseFx:
Hey @hjanuschka, this is great! 👍  I'd have to look into this closer, but shouldn't the following work: inside _spaceship_ we check if the `FastlaneCore` module is available, and if `FastlaneCore::Helper` is defined. If so we use, otherwise we don't verify if user interaction is enabled. 

From my understanding, this should make it work for all the _fastlane_ tools, and only doesn't if someone's using _spaceship_ directly without `fastlane_core`. Does that make sense?

hjanuschka:
@KrauseFx i thought of this, makes sense, but still ends in an endless loop if used only with `spaceship` - you think this is ok? - if, i ll refactor the PR.

KrauseFx:
@hjanuschka I think it's better than having duplicate code that's harder to maintain I think. A majority of the users will use _spaceship_ via _fastlane_ 👍 
Do you want to update the PR or should I help?

hjanuschka:
@KrauseFx updated! thx!

spaceship/lib/spaceship/portal/ui/select_team.rb
```diff
@@ -31,6 +31,20 @@ class UserInterface
...
 
+      def self.ci?
+        if Object.const_defined?("FastlaneCore")
+          return FastlaneCore::Helper.ci?
+        end
+        return false
+      end
+
+      def self.interactive?
+        if Object.const_defined?("FastlaneCore")
```
KrauseFx: Just to be save, maybe this should be
```ruby
if Object.const_defined?("FastlaneCore") && FastlaneCore.const_defined?("Helper")
```
Because maybe someone defines a `FastlaneCore` module, without actually importing `fastlane_core`

spaceship/lib/spaceship/portal/ui/select_team.rb
```diff
@@ -64,6 +78,16 @@ def select_team
...
         return teams[0]['teamId'] if teams.count == 1 # user is just in one team
 
+        unless self.class.interactive?
+          puts "Multiple teams found on the Developer Portal, Your Terminal is running in non-interactive mode! Cannot continue from here."
+          puts "Please check that you set FASTLANE_TEAM_ID or FASTLANE_TEAM_NAME to the right value."
+          puts "Available Teams:"
+          teams.each_with_index do |team, i|
+            puts "#{i + 1}) #{team['teamId']} \"#{team['name']}\" (#{team['type']})"
+          end
+          raise "multiple_teams_no_interactive_shell"
```
KrauseFx: Why a format like this, and not a human-readable text?

spaceship/lib/spaceship/portal/ui/select_team.rb
hjanuschka: good point, thx! 👍

spaceship/lib/spaceship/portal/ui/select_team.rb
hjanuschka: i am no human 🤖  -> fixed!

hjanuschka (commented):
@KrauseFx feedback addressed

KrauseFx (approved):

KrauseFx:
Tests are red, but I can't open the test output...

hjanuschka:
fixed the test - should be green soon

KrauseFx:
Wohooo, thank you @hjanuschka, this is awesome 👍

PR #9170, opened 2017-05-15
Record scan build failures as just that rather than user_errors

ohayon:
This is to make sure that we are recording build failures properly rather than blaming the user 😄


mpirri (approved):

KrauseFx (approved):
🎷

PR #9171, opened 2017-05-15
Make sure that valid? raises rather than verify

ohayon:
Saw a crash in Stackdriver that showed a lot of failures in the `verify` method, but it calls into `valid?` which should be responsible for raising exceptions.

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/config_item.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/config_item.rb
@@ -103,7 +103,7 @@ module FastlaneCore
     end
 
     def verify!(value)
-      true if valid?(value)
+      valid?(value)
     end
 
     # Make sure, the value is valid (based on the verify block)
--- a/fastlane_core/lib/fastlane_core/configuration/config_item.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/config_item.rb
@@ -102,13 +102,12 @@ module FastlaneCore
       @allow_shell_conversion = (type == :shell_string)
     end
 
-    # This will raise an exception if the value is not valid
     def verify!(value)
       true if valid?(value)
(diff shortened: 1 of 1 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -102,14 +102,12 @@ def initialize(key: nil, env_name: nil, description: nil, short_option: nil, def
       @allow_shell_conversion = (type == :shell_string)
     end
 
-    # This will raise an exception if the value is not valid
     def verify!(value)
-      UI.user_error!("Invalid value '#{value}' for option '#{self}'") unless valid?(value)
-      true
+      true if valid?(value)
```
mpirri: I think I find this a little bit confusing, I think - we have a method ending with a "!" that doesn't explicitly have a raise in it, and a method ending with a "?" that returns either true or else raises an exception.

Feels like the entire body of `valid?` should be moved into `verify!`, and the `true` return value should be eliminated (since it is always true if not raising).

Or am I missing something?  😄

KrauseFx: Yes, technically the correct thing to do would be to raise the exception here (method with the `!`), and only return booleans in `valid?`. However this would also mean that we can't show specific error messages that are helpful for the user to understand. Maybe we should just have one method that raises the exception?

KrauseFx (approved):

Issue #9121, opened 2017-05-09
Cannot retrieve correct prices for auto-renewable subscriptions via Spaceship


klappradla:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

First of all, thanks a ton for providing this awesome set of tools 🙏 

I'm trying to retrieve the actual prices _(in their respective currency)_ for in app purchases of the type `Auto-Renewable Subscription`. The `pricing_intervals` contain entries like this:

```ruby
{:tier=>85, :begin_date=>nil, :end_date=>nil, :grandfathered=>{"value"=>"FUTURE_NONE", "isEditable"=>false, "isRequired"=>false, "errorKeys"=>nil}, :country=>"MT"},
{:tier=>78, :begin_date=>nil, :end_date=>nil, :grandfathered=>{"value"=>"FUTURE_NONE", "isEditable"=>false, "isRequired"=>false, "errorKeys"=>nil}, :country=>"GB"}
```

This matches Apple's [Territory Specific Pricing](https://developer.apple.com/app-store/subscriptions/) and its [Prices for Subscriptions](https://developer.apple.com/app-store/subscriptions/App-Store-Subscription-Pricing.pdf).

However, the `tier` specified here does not relate to the `tier_stem` in an `Spaceship::Tunes::PricingTier`, but rather the row-index of the CVS file I can download on iTunes Connect's *"All Prices and Currencies - Auto-Renewable Subscriptions"* page: 
```
https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/ng/app/<app-id>/pricingMatrix/recurring
```

If I now append a `.json` to the previous url:
```
https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/ng/app/<app-id>/pricingMatrix/recurring.json
```
I get a completely different pricing matrix, now including a `Tier` row. This matrix is what `Spaceship::Tunes` returns for `pricing_tiers`.

#### My Question now
Does `Spaceship` also provide the means to retrieve the prices for *"Auto-Renewable Subscriptions"*?
I'd be happy to work on a PR for this 🎉 , however, I could not yet find a place where Apple themselves provide this information *(besides the CSV file)*.

When trying to *abuse* the iTunes Connect Angular app, I can get the CSV content as JSON via this URL:
```
[...]

PR #7834, opened 2017-01-12
🛍 In-App Purchase support in spaceship 👾

hjanuschka:
* Feature Complete:
    * 100% ██████████ 🎉🎉🎉
  * Specs:
    * 100% ██████████ 🎉🎉🎉
bildschirmfoto 2017-01-19 um 23 04 20
  * Polish for Review
      * 100% ██████████ 🎉🎉🎉
  * Documentation:
    * Inline
      * 100% ██████████ 🎉🎉🎉

# Add Support for In-App-Purchases to Spaceship

Todo:
 - [ ]  ~purchases - submit for review~ -> will do a standalone PR on this, once merged.

## Credits
i took the base of  type/status mapping from this PR: https://github.com/fastlane/fastlane/pull/7853
thx @kobim 

## API:

*BASE*:  

```ruby
app = Spaceship::Application.find("hjanuschka.sticker-test")
```

### Find IAPS:

```ruby
  app.in_app_purchases.all
```

*Returns*: Array of `Spaceship::Tunes::IAPList` 's 

### Delete IAP:

```ruby
purchase = app.in_app_purchases.all.first
purchase.delete!
```
a
### Create IAP:
  * Types:
    * CONSUMABLE
    * NONCONSUMABLE
    * NON_RENEW_SUBSCRIPTION

```ruby
app.in_app_purchases.create!(
  type: Spaceship::Tunes::IAPType::CONSUMABLE, 
  versions: {
    'en-US': {
      name: "test name1",
      description: "Description has at least 10 characters"
    },
    'de-DE': {
      name: "test name german1",
      description: "German has at least 10 characters"
    },
    'da': {
      name: "test name danish",
      description: "German has at least 10 characters"
    }
  },
  reference_name: "locallizeddemo11111",
  product_id: "x.a.a.b.b.c.d.x.y.f",
  cleared_for_sale: true,
  review_notes: "Some Review Notes here bla bla bla",
  review_screenshot: "/Users/hjanuschka/Desktop/review.jpg", 
  pricing_intervals: 
    [
      {
        country: "WW",
        begin_date: nil,
        end_date: nil,
        tier: 1
      }
    ] 
)
```

## Modify IAP

```ruby
purchase = app.in_app_purches.all.first.edit
purchase.versions = {
      'en-US': {
        name: "Edit It",
        description: "Description has at least 10 characters"
      },
      'de-DE': {
        name: "test name german1",
        description: "German has at least 10 characters"
      },
      'da': {
        name: "test name danish",
[...]

PR #9180, opened 2017-05-16
Allow to retrieve the actual prices for in-app purchases

klappradla:
### Checklist

### Motivation and Context

This addresses (and fixes) https://github.com/fastlane/fastlane/issues/9121.

### Description

This PR adds a `pricing_info` method to `IAPDetail` which allows to retrieve the actual prices for in-app purchases. Due to Apple's [Introduction of Territorial Pricing](https://developer.apple.com/app-store/subscriptions/) the return value depends on the type of in-app purchase.
There are two possible scenarios:
1. The prices are defined "world wide" via a single "pricing tier" (e.g. for one-off purchases)
2. The prices are defined individually for each country (= **territorial pricing**) (e.g. for auto-renewable subscriptions)

For the latter case, I also introduced a new type of pricing tier, `IAPSubscriptionPricingTier` *(naming was kind of hard, but that's essentially what this entity is about...)*.

#### Changes to the client

Since retrieving the matrix for auto renewable subscription price tiers is quite computational expensive, it is cached on the client. I introduced the same for the "normal" pricing tiers in order to align the behavior here.

#### Open questions

I added the required yard documentation for the new methods I introduced, staying with the style it generally was set up in the files I worked with. I would however be happy to add even more information, I just did not want to break the style of inline documentation. So please tell me if there are places where you think I could add some more information.

#### Pricing Matrix for Auto-Renewable Subscriptions

Use the `TunesClient` to retrieve all available prices for territorial pricing:

```ruby
app_id = "123456" # any app
Spaceship::Tunes.client.subscription_pricing_tiers(app_id)
```
Returns an array of `Spaceship::Tunes::IAPSubscriptionPricingTier`s. The data is the same for any `app_id`, Apple's endpoint still requires one though 🙈. It resembles the prices listed in this matrix in iTunes Connect 👇

screen shot 2017-05-17 at 12 44 31

#### Pricing Info for Auto-Renwable Subscriptions

Retrieve the pricing info for a subscription based in-app purchase.

##### Prerequisites

Retrieve a subscription based in-app purchase:

```ruby
app_id = "123456" # app offering subscription based IAPs
iap = Spaceship::Tunes::Application
  .find(app_id)
  .in_app_purchases
  .all
  .select { |i| i.type == "Auto-Renewable Subscription" }
  .first
iap.type # => "Auto-Renewable Subscription"
iap.class # => Spaceship::Tunes::IAPList
```

##### Pricing Info

Retrieve the actual prices:

```ruby
# get the IAPDetail object
iap_detail = iap.edit
iap_detail.class # => Spaceship::Tunes::IAPDetail

# the iap needs to be "cleared for sale", otherwise it's not possible to retrieve any pricing information
iap_detail.cleared_for_sale # => true

# retrieve its prices
iap_detail.pricing_info # => [Spaceship::Tunes::IAPSubscriptionPricingInfo]
```

Returns an array of `Spaceship::Tunes::IAPSubscriptionPricingInfo`s, containing the prices for each country.

#### Pricing Info for "one-off" purchases

Retrieve the pricing info for a non-subscription based in-app purchase using world wide pricing *(defined via choosing a single price tier in iTunes Connect)*.

##### Prerequisites

Retrieve a non-subscription based in-app purchase:

```ruby
app_id = "123456" # app offering non-subscription IAPs
iap = Spaceship::Tunes::Application
  .find(app_id)
  .in_app_purchases
  .all
  .select { |i| i.type == "Non-Consumable" }
  .first
iap.type # => "Auto-Renewable Subscription"
iap.class # => Spaceship::Tunes::IAPList
```

##### Pricing Info

Retrieve the actual prices:

```ruby
# get the IAPDetail object
iap_detail = iap.edit
iap_detail.class # => Spaceship::Tunes::IAPDetail

# the iap needs to be "cleared for sale", otherwise it's not possible to retrieve any pricing information
iap_detail.cleared_for_sale # => true

# retrieve its prices
iap_detail.pricing_info # => [Spaceship::Tunes::PricingInfo]
```

Returns an array of `Spaceship::Tunes::PricingInfo`s, containing all prices from the specified price tier.

#### No Pricing Info available

If an in-app purchase is not *"cleared for sale"* yet, it's not possible to retrieve any information related to pricing:

```ruby
iap_detail.cleared_for_sale # => false

# no pricing intervals available
iap_detail.pricing_details # => []

# no pricing info available either
iap_detail.pricing_info # => []
```

It returns an empty array for both, `pricing_intervals` and `pricing_info`.


klappradla:
Hey @googlebot 👋 , I signed it!

klappradla:
Hey @hjanuschka *(I thought I could request a review by you, but the button is disabled for me...)*,

here are the changes we talked about, fixing what I described in #9121 

**Tiny sidenote:** I would have loved to memorize `pricing_intervals` in `IAPDetail` (would have made the code so much cleaner), but I was not sure if this is a pattern you use or not. If that's ok to do, I'd love to update `IAPDetail` respectively.

hjanuschka:
hey @klappradla awesome PR - going to review this tomorrow.

i don't really get the sidenot question. you mean using the `@` cache directly in iap_detail?

klappradla:
Yes, sorry, could have made this more obvious.

**Changes to `pricing_intervals`**

So I mean changing this method in `IAPDetail`:
```ruby
def pricing_intervals
  parsed_intervals = []
  raw_data["pricingIntervals"].each do |interval|
    parsed_intervals << {
      tier: interval["value"]["tierStem"].to_i,
      begin_date: interval["value"]["priceTierEffectiveDate"],
      end_date: interval["value"]["priceTierEndDate"],
      grandfathered: interval["value"]["grandfathered"],
      country: interval["value"]["country"]
    }
  end
  return parsed_intervals
end
```

To first of all use `map` (makes it easier to use memorization afterwards): 👇 
```ruby
def pricing_intervals
  raw_data["pricingIntervals"].map do |interval|
    {
      tier: interval["value"]["tierStem"].to_i,
      begin_date: interval["value"]["priceTierEffectiveDate"],
      end_date: interval["value"]["priceTierEndDate"],
      grandfathered: interval["value"]["grandfathered"],
      country: interval["value"]["country"]
    }
  end
end
```

And then next up one could memorize the value for future calls on the instance: 👇 
```ruby
def pricing_intervals
  @pricing_intervals ||= raw_data["pricingIntervals"].map do |interval|
    {
      tier: interval["value"]["tierStem"].to_i,
      begin_date: interval["value"]["priceTierEffectiveDate"],
      end_date: interval["value"]["priceTierEndDate"],
      grandfathered: interval["value"]["grandfathered"],
      country: interval["value"]["country"]
    }
  end
end
```

Doing so, I could then for instance spare to carry over the `pricing_intervals` though all my subsequent calls (a thing I do in order to not have to do the mapping of the values over and over, but only once)

**Result on my new code**
Doing so, I could change my new methods to this: *(docs omitted for better readability here)* 👇 
```ruby
def pricing_info
  intervals = pricing_intervals
  return world_wide_pricing(pricing_intervals) if world_wide_pricing?(pricing_intervals)
  territorial_pricing(pricing_intervals)
end

private

def world_wide_pricing?
  pricing_intervals.first[:country] == "WW"
end

def world_wide_pricing
  client
    .pricing_tiers
    .find { |p| p.tier_stem == pricing_intervals.first[:tier].to_s }
    .pricing_info
end

def territorial_pricing
  pricing_matrix = client.subscription_pricing_tiers(app_id: application.apple_id)
  pricing_intervals.map do |interval|
    pricing_matrix
      .find { |p| p.tier_stem == interval[:tier].to_s }
      .pricing_info
      .find { |i| i.country_code == interval[:country] }
  end
end
```

This ☝️  would have been the way I would have initially written this code. But I then wondered if not-memorizing the `pricing_intervals` was done on purpose, e.g. to allow fetching changes from the API without initializing a new object. I personally would see this as non-best-practise use case and therefore not encourage users to do this.
I just wasn't sure about your intention with this code, so I wrote the ugly, more defensive version for now and just ask you if I can change it to the memorization version ✌️

hjanuschka:
top notch explanation 👍 tbh. i did this "ugly, more defensive" (best word) - just out of lack of knowledge - id be fine with your implementation it looks solid. and aslong as tests-pass i am definitily on your side of the change.

> if i get it right, it doesn't really change (from end-user point of view) the `pricing_intervals` ?

so you'd update the PR. - and later on i ll give it a test run with my existing IAP's 👍 

also would be nice if you include full examples of the new api (in the PR body) - like i did it in the #7834 - that the PR is sort a documentation - where we can copy&paste samples in the future, to troubleshoot, if there are any regressions.

klappradla:
👍  sounds like a plan, I'll update accordingly.

And no, no changes from end-user point of view, I cannot imagine someone relying on re-computing the `pricing_intervals` hash on every method call.

klappradla:
Alright @hjanuschka, I updated the PR accordingly:

**memorization**
Updated the `pricing_intervals` method to memorize its value

**examples**
Updated the PR description with an *example* section. Just tell me if you're still missing something there

**defensiveness**
As you may have seen when implementing `pricing_intervals`, Apple's API does not return any pricing-related information unless the IAP is "cleared for sale". I added an additional guard clause to `pricing_info` to quickly return an empty collection in this case in order to spare a useless call to Apple's API for fetching the price tier matrix ✌️

hjanuschka:
@klappradla that is one awesome "first-time contribution" 💯 

on the subscription pricing it seems that country is always nil, whereas on non-consumables it is not nil.

```ruby
<Spaceship::Tunes::IAPSubscriptionPricingInfo 
        country=nil, 
        country_code="TM", 
        currency_symbol="$", 
        currency_code="USD", 
        wholesale_price=10.5, 
        wholesale_price2=12.74, 
        retail_price=14.99, 
        f_retail_price="$14.99", 
        f_wholesale_price="$10.50", 
        f_wholesale_price2="$12.74">,
```

beside that, PR looks good. going to test it on a few older ruby's (just in case) , will report tomorrow


i tested it with the following:

```ruby
require "spaceship"
app_id = "at.krone.epaper" # app offering subscription based IAPs
iap = Spaceship::Tunes::Application
  .find(app_id)
  .in_app_purchases
  .find("at.krone.epaper.ABO.monatlich")

iap_detail = iap.edit

require "pry"
binding.pry


iap = Spaceship::Tunes::Application
  .find(app_id)
  .in_app_purchases
  .find("at.krone.epaper.einzelausgabe")

iap_detail = iap.edit

binding.pry

puts ""

```

klappradla:
Cool 👍 

Yes, I've seen the inconsistency with `country` as well. Just for a quick explanation where this comes from:

**non-consumables**

`non-consumables` use world wide pricing, therefore their `pricing_info` is basically the `pricing_info` array from the respective `PriceTier` coming from this API endpoint on Apple's side:

```
https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/apps/pricing/matrix
```

The payload returned looks like this:
```js
{
  "data": {
    "pricingTiers": [
      {
        "tierStem": "0",
        "tierName": "Free",
        "sortOrder": null,
        "pricingInfo": [
          {
            "countryCode": "US",
            "currencyCode": "USD",
            "wholesalePrice": 0,
            "wholesalePrice2": 0,
            "retailPrice": 0,
            "country": "United States",
            "currencySymbol": "$",
            "fRetailPrice": "$0.00",
            "fWholesalePrice": "$0.00",
            "fWholesalePrice2": "$0.00"
          },
          {
            "countryCode": "CA",
            "currencyCode": "CAD",
            "wholesalePrice": 0,
            "wholesalePrice2": 0,
            "retailPrice": 0,
            "country": "Canada",
            "currencySymbol": "$",
            "fRetailPrice": "$0.00",
            "fWholesalePrice": "$0.00",
            "fWholesalePrice2": "$0.00"
          },
         // etc.
```
Here, `country` is usually the full name for the country the price is used for. So this is also part of the resulting `Spaceship::Tunes::PricingInfo` objects then returned when retrieving the `pricing_info` for a non-consumable IAP.

**auto-renwable subscriptions**

`auto-renewable-subscriptions` use territorial pricing, so the `pricing_info` is retrieved from the respective `IAPSubscriptionPriceTier`s coming from this API endpoint:

```
https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/apps/<app-id>/iaps/pricing/matrix/recurring
```

The payload returned looks like this:

```js
{
  "data": {
    "pricingTiers": [
      {
        "tierStem": "1",
        "tierName": "ITC.addons.pricing.tier.1",
        "sortOrder": null,
        "pricingInfo": [
          {
            "countryCode": "PY",
            "currencyCode": "USD",
            "wholesalePrice": 0.35,
            "wholesalePrice2": 0.42,
            "retailPrice": 0.49,
            "country": null,
            "currencySymbol": "$",
            "fRetailPrice": "$0.49",
            "fWholesalePrice": "$0.35",
            "fWholesalePrice2": "$0.42"
          },
          {
            "countryCode": "BJ",
            "currencyCode": "USD",
            "wholesalePrice": 0.35,
            "wholesalePrice2": 0.42,
            "retailPrice": 0.49,
            "country": null,
            "currencySymbol": "$",
            "fRetailPrice": "$0.49",
            "fWholesalePrice": "$0.35",
            "fWholesalePrice2": "$0.42"
          }
         // etc.
```
So `country` is always `null`. I honestly don't know why Apple includes it then - it just makes the payload bigger...

I kept it as an attribute for the `Spaceship::Tunes::IAPSubscriptionPricingInfo` in order to stay as close to `Spaceship::Tunes::PricingInfo` as possible.

However, I would be more then happy to remove the attribute from `Spaceship::Tunes::IAPSubscriptionPricingInfo`, since it a) does not give any value to the object and b) may even be irritating (since `country_code` is set) for future developers.

hjanuschka:
@klappradla  ok then lets do it, since its always `nil` remove it 👍 

i finished my tests, and as expected it fails on system ruby (which is 2.0.0 😱 )
the version does not support optional named parameters. and we HAVE to still support system ruby.

error:

```
SyntaxError: /fastlane/spaceship/lib/spaceship/tunes/tunes_client.rb:1033: syntax error, unexpected ')'

```

changing it works - here is the diff:

```diff
diff --git a/spaceship/lib/spaceship/tunes/tunes_client.rb b/spaceship/lib/spaceship/tunes/tunes_client.rb
index b018c55..3bb6837 100644
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -1030,7 +1030,7 @@ module Spaceship
     #
     # @param params (Hash)
     # @option params (String) :app_id
-    def subscription_pricing_tiers(app_id:)
+    def subscription_pricing_tiers(app_id:nil)
       @subscription_pricing_tiers ||= begin
         r = request(:get, "ra/apps/#{app_id}/iaps/pricing/matrix/recurring")
         data = parse_response(r, "data")["pricingTiers"]
```

hjanuschka:
looking good so far, let me know when you are ready.
one last nitpick, nameing convention for spec's is `iap_subscription_pricing_tier_spec.rb` instead of `iap_subscription_pricing_tier.spec.rb` (i missed that on initial review) - sorry

klappradla:
Ah ok, thanx for pointing out @hjanuschka 🙏 

**kwargs**
I knew Fastlane would be targeting pre-2.3 Ruby versions, but I did not know it would be so far back into the past 😉  I honestly never in my life used the system Ruby, but I just looked it up, it's something along 2.0.x, so yeah, makes total sense: KWARGS were introduced in 2.0, but "required KWARGS" *(the syntax I was using: `def method(arg:)` )* without default value were only introduced in 2.1 💡 

However, setting the default value for `app_id` to `nil` does not make too much sense here in my opinion, since Apple's API endpoint requires it as part of the URL... So having it default to `nil` would just end up as 404 on Apple's side with a non-descriptive error for the Fastlane user if he or she does not pass an `app_id` parameter. Therefore I just made it an "normal parameter", so Spaceship will scream at the developer if it's not passed as an argument. Fits the use case better here in my opinion :v:

**country**
I removed the `country` attribute, 4300b5c741c37e7912a241aee0294e0d147c9f6c explains the background.

klappradla:
> iap_subscription_pricing_tier.spec.rb
This was a typo 😂 , thanx for pointing out 🙏

hjanuschka (approved):
🚀 many thx @klappradla this is a huge "first-time contribution" PR - really awesome work! thx for your patience and invested effort. ❤️

klappradla:
Thanx @hjanuschka, was a pleasure 🥂

KrauseFx:
You all, this is great, thanks @klappradla and @hjanuschka 🇦🇹🇦🇹🇦🇹

ShelinShelin:
@hjanuschka   @KrauseFx  
 hi, what the tier string about the standby tier,  such as CNY6 is tier 1, and what about CNY8? and what is the difference between  tier 1 and  standby tier 1?

PR #9181, opened 2017-05-16
Handle case of Info.plist path prepended by $(SRCROOT)

ayushgoel:
### Checklist

### Motivation and Context
The Fastlane action is unable to find the Info.plist for my newly created target.

### Description
The action when checking xcodeproj for validity of given plist path, would now handle the case of `$(SRCROOT)` prepended to it.

### DIFF
--- a/fastlane/lib/fastlane/actions/update_app_identifier.rb
+++ b/fastlane/lib/fastlane/actions/update_app_identifier.rb
@@ -23,7 +23,7 @@ module Fastlane
           configs = project.objects.select { |obj| obj.isa == 'XCBuildConfiguration' && !obj.build_settings[identifier_key].nil? }
           UI.user_error!("Info plist uses $(#{identifier_key}), but xcodeproj does not") unless configs.count > 0
 
-          configs = configs.select { |obj| obj.build_settings[info_plist_key] == params[:plist_path] }
+          configs = configs.select { |obj| obj.build_settings[info_plist_key] == params[:plist_path] || obj.build_settings[info_plist_key] == "$(SRCROOT)/"+params[:plist_path] }
           UI.user_error!("Xcodeproj doesn't have configuration with info plist #{params[:plist_path]}.") unless configs.count > 0
 
           # For each of the build configurations, set app identifier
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/update_app_identifier.rb
```diff
@@ -23,7 +23,7 @@ def self.run(params)
           configs = project.objects.select { |obj| obj.isa == 'XCBuildConfiguration' && !obj.build_settings[identifier_key].nil? }
           UI.user_error!("Info plist uses $(#{identifier_key}), but xcodeproj does not") unless configs.count > 0
 
-          configs = configs.select { |obj| obj.build_settings[info_plist_key] == params[:plist_path] }
+          configs = configs.select { |obj| obj.build_settings[info_plist_key] == params[:plist_path] || obj.build_settings[info_plist_key] == "$(SRCROOT)/" + params[:plist_path] }
```
KrauseFx: This line now is pretty complex, can you split it up to multiple?

ayushgoel:
@KrauseFx Updated. I hope the extraction of `possible_plist_paths` makes the method expandable too. 

Also, I was thinking if instead we could check for the suffix to be `params[:plist_path]`.

KrauseFx:
Thanks for your contribution @ayushgoel, I have to say I don't have enough knowledge about this, it's really hard to judge if this PR should be merged. It's key, that no existing setups will break when introducing this change. I'd love to get a :+1: from some other production setups before merging it, and ideally have 2 tests covering this change. 

Sorry that we can't proceed faster, but in the past we've broke things quite often when we merged PRs we didn't fully understand.

ohayon:
Hey @ayushgoel, I am also a bit confused by this change. Is it possible you could add a couple comments and/or (ideally) tests and maybe a better description about why this change is helpful/necessary? 😄  Thanks!

ayushgoel:
The `Info.plist`'s path for our project is prepended by `$(SRCROOT)` in the build configuration. This essentially makes it a relative path to the source folder. But this action fails as it is unable to spot the plist path in the build configuration. We have tried and tested it by making a copy of the `UpdateAppIdentifierAction` locally and using this change.

I am new to ruby so do not know much about how testing is done here.
I guess the test file for the changed action is -> https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/update_app_identifier_spec.rb
Will try to update the test and report here. 
Thank you for looking at this change and giving your valuable inputs. 👍

ohayon:
Hey @ayushgoel - thanks again for this PR, if it's okay with you, I am going to close this for now as it doesnt seem like an issue that is affecting many other users. We would be happy to reevaluate this if it comes up more often. Also if you would want to get this merged in in the future, it would be great to have some tests for it just so we know we wont be breaking anyone else's setups. Thanks again! 🚀

PR #9194, opened 2017-05-17
Add missing API doc for spaceship::portal

huang-kun:
I think it's better to add details to let user know that this is the way to get real app details.


huang-kun:
I signed it!

mfurtak (approved):
Thanks for the contribution! 🚀

PR #9199, opened 2017-05-17
Move error types into their own files

ohayon:
As we add more error types and functionality to them, we should not keep adding to `interface.rb` and instead give them the files they deserve. 😄


mpirri (approved):
👍

PR #9200, opened 2017-05-17
Add shell_error! to help clean up places where we do poor error handling

mpirri:
### Checklist

### Motivation and Context
Currently shell errors are all attributed in the crash reporter to `sh_control_output`, particularly when the caller has not done any of their own error handling. This makes it difficult to identify where the errors are actually coming from. As such, we are introducing a new exception type that, when reporting to the crash reporter, does not show any methods in sh_helper, allowing us to see the actual root cause for the issue.


ohayon (approved):
👍

PR #9208, opened 2017-05-18
Stop reporting interruptions as crashes

ohayon:
We don't want to be recording failures that come from users interrupting/cancelling their _fastlane_ executions.


mpirri (approved):
👍

PR #9220, opened 2017-05-19
Check to see if frame is method_missing before dropping

mpirri:
### Checklist

### Motivation and Context
We drop 2 frames when exception is raised by `error!` or `user_error!`, but there are times when these methods are called from within the interface class and do not use the `method_missing` dispatch. As such, we now check to see if the second stack from contains `method_missing` and don't drop unless it is present.


ohayon (approved):
💯

PR #9240, opened 2017-05-23
Skip error message when no simulator is provided

KrauseFx:
We've shown error messages for cases that aren't actually errors


taquitos (commented):
🐑 'it

taquitos (approved):

PR #9245, opened 2017-05-23
Rescue exceptions when crash reporter cannot write out status file

mpirri:
### Checklist

### Motivation and Context
The crash reporter notifies the user that it is sending a crash report on the first crash; it writes a file to the `.fastlane` directory after doing this so that it doesn't overwhelm the user with this message after each crash. However, if the `.fastlane` directory is not writable, the crash reporter currently crashes, and records the file write error as the crash. This is not ideal, as it hides the actual cause of the original crash. So, this change `rescue`s that exception on writing the file.

The downside is that for users in this scenario they will see the crash report message over and over, but we will address that issue if it becomes a problem for the relatively small number of users seeing this issue.

### DIFF
--- a/fastlane_core/lib/fastlane_core/crash_reporter/crash_reporter.rb
+++ b/fastlane_core/lib/fastlane_core/crash_reporter/crash_reporter.rb
@@ -62,11 +62,7 @@ module FastlaneCore
 
         return did_show if did_show
 
-        begin
-          File.write(path, '1')
-        rescue
-          # don't show the user a message, since there is no functionality lost when this happens
-        end
+        File.write(path, '1') rescue nil
         false
       end
 
--- a/fastlane_core/lib/fastlane_core/crash_reporter/crash_reporter.rb
+++ b/fastlane_core/lib/fastlane_core/crash_reporter/crash_reporter.rb
@@ -62,7 +62,11 @@ module FastlaneCore
 
         return did_show if did_show
 
-        File.write(path, '1')
+        begin
+          File.write(path, '1')
+        rescue
+          # don't show the user a message, since there is no functionality lost when this happens
+        end
         false
       end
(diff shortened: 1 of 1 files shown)


lyndsey-ferguson:
Can a check be added to see if we can write and include that information in the rescue section?

mpirri:
My feeling was that if we do that we take an already verbose message and make it even more verbose. Messages of the sort of "Cannot save state that you have already seen the message that we just showed you" are confusing and not super-actionable.  Our hypothesis is that, most likely, this error is happening on CI boxes where the fastlane process doesn't have access to the user's home directory.

lyndsey-ferguson:
Oh!

fastlane_core/lib/fastlane_core/crash_reporter/crash_reporter.rb
```diff
@@ -62,7 +62,11 @@ def did_show_message?
 
         return did_show if did_show
 
-        File.write(path, '1')
+        begin
+          File.write(path, '1')
+        rescue
+          # don't show the user a message, since there is no functionality lost when this happens
```
ohayon: How would you feel about changing this to be something like `File.write(path, '1') rescue nil`?

ohayon (approved):

fastlane_core/lib/fastlane_core/crash_reporter/crash_reporter.rb
mpirri: 🔥

mpirri:
When running with `--verbose`, this is now what is displayed.  Without `--verbose`, the 2nd and 3rd red lines are omitted.

image

joshdholtz (approved):
I like that verbose output 😉

PR #9253, opened 2017-05-24
Remove duplicate auth code in spaceship

KrauseFx:
I didn't run a `git blame` on when and why this was introduced, but I could imagine that this happened during the TestFlight migrations about a month ago. Tests are passing, and some local tests seem to indicate things work fine :)

Also we have 489 tests on spaceship already :rocket:

I moved the code to `client.rb` as it also seems to be used as part of `testflight/client.rb`

In related question @ohayon @snatchev: Why is the name of the TestFlight client `client.rb` and not `testflight_client.rb`, like it's the case for `portal_client` and `tunes_client`? Is that something we can change? 

Thanks everyone 👋

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -137,6 +137,50 @@ module Spaceship
       end
     end
 
+    # Fetch the general information of the user, is used by various methods across spaceship
+    # Sample return value
+    # => {"associatedAccounts"=>
+    #   [{"contentProvider"=>{"contentProviderId"=>11142800, "name"=>"Felix Krause", "contentProviderTypes"=>["Purple Software"]}, "roles"=>["Developer"], "lastLogin"=>1468784113000}],
+    #  "sessionToken"=>{"dsId"=>"8501011116", "contentProviderId"=>18111111, "expirationDate"=>nil, "ipAddress"=>nil},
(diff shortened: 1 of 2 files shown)


ohayon:
Hey @KrauseFx this looks great! And there is no reason in particular that the file is called `client.rb` instead of `testflight_client.rb` that can definitely be changed to match the pattern we have used for other clients.

KrauseFx:
Cool, thanks for confirming. Get a get an approval for this? 👍

snatchev:
@KrauseFx @ohayon my motivation for naming it simply `client.rb` is because it's already namespaced by the folder it is, as well as the module name in code.

KrauseFx:
@snatchev If that's the case, we should rename the other client class too to be consistent, right?

snatchev (approved):

KrauseFx:
Thanks for the review @snatchev 👍

snatchev:
It would be trickier to rename `portal_client.rb` because there is no `Spaceship::Portal` namespace like there is for `Spaceship::TestFlight`. Not a huge code change, but I am not sure.

I am more inclined to "fix it forward" as new APIs get created and old ones get deprecated. Let me

KrauseFx:
Yeah, let's leave it like is, no big deal

PR #9258, opened 2017-05-24
fix method return value is the error log issue

truebit:
### Checklist
- [ x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [ x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Motivation and Context
bug

### Description
when not set apple versioning system following https://developer.apple.com/library/content/qa/qa1827/_index.html and user passed "build_number" argument, the shell command would return 0 instead of 6. but the command to get `build_number` would return the error log string with "There does not seem to be a CURRENT_PROJECT_VERSION key set for this project.  Add this key to your target's expert build settings."

### DIFF
--- a/fastlane/lib/fastlane/actions/increment_build_number.rb
+++ b/fastlane/lib/fastlane/actions/increment_build_number.rb
@@ -45,6 +45,12 @@ module Fastlane
           # Store the new number in the shared hash
           build_number = `#{command_prefix} agvtool what-version`.split("\n").last.strip
 
+          # if agvtool not enabled and user specified build_number, which means only modify plist build number
+          if $?.exitstatus != 0 and params[:build_number]
+            UI.important("only modify plist build number to specifed '#{params[:build_number]}'")
+            build_number = params[:build_number]
+          end
+          
           return Actions.lane_context[SharedValues::BUILD_NUMBER] = build_number
         end
       rescue => ex
@@ -60,7 +66,7 @@ module Fastlane
         [
           FastlaneCore::ConfigItem.new(key: :build_number,
                                        env_name: "FL_BUILD_NUMBER_BUILD_NUMBER",
-                                       description: "Change to a specific version",
+                                       description: "Change to a specific version. If you did not set Versioning System in Xcode, it would only modify build number in .plist file",
                                        optional: true,
                                        is_string: false),
           FastlaneCore::ConfigItem.new(key: :xcodeproj,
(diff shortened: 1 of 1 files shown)


truebit:
I signed it

ohayon:
Hey @truebit, thanks for the PR! I am hesitant to merge this because in order to use the `increment_build_number` action, we do require that the user is using Apple Generic Versioning. 

Could you explain a bit more about why this is helpful for you and what the use case is?

truebit:
`agvtool new-version -all <your_specific_version>` would NOT throw exception (exit code would be 0) even if you did not enable Apple Generic Versioning.  And this command would also set all plist version number. 

But when executing `agvtool what-version`, it would exit with code 6 instead of 0. and the build_number would be set to the stderr string.

You could try this out in any Xcode project which did not enable Apple Generic Versioning:

```
build_num = increment_build_number(build_number: Time.now.strftime("%Y%m%d") +'.'+ number_of_commits.to_s)
puts build_num
```
`build_num` would be `There does not seem to be a CURRENT_PROJECT_VERSION key set for this project.  Add this key to your target's expert build settings.` instead of expected `20170614.123`

ohayon:
Hey @truebit, I was just able to reproduce this on my machine, and what you pointed out is happening for me too, but I think the proper approach here would be to make sure that Apple Generic Versioning is enabled and that the `CURRENT_PROJECT_VERSION` key is present before allowing this action to continue. If you want to set your build number to something specific and you are not using AGV, I think a better approach would be to use the `set_plist_value` action to set the `CFBundleVersion` key to the value you want. I just submitted https://github.com/fastlane/fastlane/pull/9474 to make sure that this action exits early if AGV is not enabled.

I am going to close this issue for now, but please feel free to leave comments here and continue discussion! 🚀

truebit:
The action `set_info_plist_value` needs to specify plist file path, that would be troublesome when two or even more targets exists (e.g. widget and watchos targets besides the main target).
But `agvtool` would set all of them automatically

ohayon:
@truebit, that makes sense for sure, but it I believe it is really just unexpected behavior that the build number gets set. With the incrementing version actions we really want to support projects that do use Apple Generic Versioning.

truebit:
This is the intended behavior as per `man agvtool`:
> new-version [-all version]
>            Sets the version numbers of all targets to the given version.
> 
>            The CURRENT_PROJECT_VERSION and the DYLIB_CURRENT_VERSION will be updated.  **The -all option will also update the CFBundleVersion Info.plist key.**
> 
>            If CVS support is enabled, the modified project file will be committed.

ohayon:
This assumes that the project is using Apple Generic Versioning no?

PR #9263, opened 2017-05-24
Make sure the crash reporter doesnt crash itself

ohayon:
Fixes https://github.com/fastlane/fastlane/issues/9197 by wrapping the whole crash reporter execution into a `begin/rescue` block to handle any issue that could come up including not being able to reach servers from certain places in the world.

PR #9249, opened 2017-05-23
mute `[!] Original Error` console output from test

mpirri:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
When one runs `bundle exec rspec`, as one does before each and every commit, one expects to get a very neat line of green `.`s.  Except, NO! Without warning, a single, glaring, bright-red `[!] Original error` breaks the peaceful continuity of passing tests.

image
The HORROR!

This shall not stand!

Using an rspec mock as our lever, we shall move the world[*](https://www.brainyquote.com/quotes/quotes/a/archimedes101761.html) to trap this pesky crimson affront to common sense. We then devour it before it has the opportunity to offend our orderly dot sensibilities any further.

PR #9246, opened 2017-05-23
Create log archive directory in case it doesn’t exist

lyndsey-ferguson:
Fixes #8909: the `xcrun` command will fail if the `—output` directory does not exist.  We create it to ensure success.

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Scan can fail when collecting logs when the directory that the logs are being expanded to does not exist. See Issue #8909.

### Description

This change uses the mkdir -p command via `FileUtils` to ensure that the destination directory exists. Tested locally with a job that was failing due to this issue.

PR #9265, opened 2017-05-24
Version bump to 2.34.0

KrauseFx:
Auto-generated by fastlane 🤖

**Changes since release '2.33.0':**

* Add opt-in for nightly builds using update_fastlane(nightly: true) (#8720) via Felix Krause
* Catch crashes in the crash reporter (#9263) via David Ohayon
* TestFairy option checking only checked for symbols, not strings (#9257) via Joshua Liebowitz
* [scan] Reduce run time by removing report generation command as separate process (#7365) via Keaton Greve
* Add a EOF newline in the pluginfile content if its not already there (#8942) via Josh Lesch
* Rescue exceptions when crash reporter cannot write out status file (#9245) via Mark Pirri
* mute `[!] Original Error` console output from test (#9249) via Mark Pirri
* Create log archive directory in case it doesn’t exist (#9246) via Lyndsey Ferguson
* [slack] Fix crash when defining FL_SLACK_DEFAULT_PAYLOADS (#9217) via Liam Nichols
* Skip error message when no simulator wasn't provided (#9240) via Felix Krause

### DIFF
--- a/fastlane/lib/fastlane/version.rb
+++ b/fastlane/lib/fastlane/version.rb
@@ -1,4 +1,4 @@
 module Fastlane
-  VERSION = '2.33.0'.freeze
+  VERSION = '2.34.0'.freeze
   DESCRIPTION = "The easiest way to automate beta deployments and releases for your iOS and Android apps".freeze
 end


taquitos (approved):
Best version bump in the history of version bumps

PR #9277, opened 2017-05-25
Don't report crashes from plugins

ohayon:
We are reporting crashes to our crash reporter that come from plugins. We should avoid doing that to keep crash reports to just things that we can manage.

### DIFF
--- a/fastlane/lib/fastlane/actions/badge.rb
+++ b/fastlane/lib/fastlane/actions/badge.rb
@@ -20,6 +20,7 @@ module Fastlane
           Badge::Runner.new.run(params[:path], options)
         rescue => e
           # We want to catch this error and raise our own so that we are not counting this as a crash in our metrics
+          UI.verbose(e.backtrace.join("\n"))
           UI.user_error!("Something went wrong while running badge: #{e}")
         end
       end
--- a/fastlane/lib/fastlane/actions/badge.rb
+++ b/fastlane/lib/fastlane/actions/badge.rb
@@ -16,7 +16,11 @@ module Fastlane
           shield_gravity: params[:shield_gravity],
           shield_no_resize: params[:shield_no_resize]
         }
-        Badge::Runner.new.run(params[:path], options)
+        begin
+          Badge::Runner.new.run(params[:path], options)
+        rescue => e
+          UI.user_error!("Something went wrong while running badge: #{e}")
+        end
       end
 
       #####################################################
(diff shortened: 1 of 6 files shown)


fastlane/lib/fastlane/actions/badge.rb
```diff
@@ -16,7 +16,11 @@ def self.run(params)
           shield_gravity: params[:shield_gravity],
           shield_no_resize: params[:shield_no_resize]
         }
-        Badge::Runner.new.run(params[:path], options)
+        begin
+          Badge::Runner.new.run(params[:path], options)
+        rescue => e
+          UI.user_error!("Something went wrong while running badge: #{e}")
```
KrauseFx: Why do we catch and then do a user_error? Why don't we re-raise the exception? It would be cool to have some code comments here on this 👍

fastlane_core/lib/fastlane_core/crash_reporter/crash_report_generator.rb
```diff
@@ -27,6 +35,12 @@ def format_crash_report_message(exception: nil)
         message + backtrace
       end
 
+      def crash_came_from_plugin?(exception: nil)
+        return false if exception.nil?
+        plugin_frame = exception.backtrace.find { |frame| frame.include?('fastlane-plugin-') }
+        plugin_frame.nil? ? false : true
```
KrauseFx: We could do `!plugin_frame.nil` as well

KrauseFx (approved):
This is great 👍

fastlane/lib/fastlane/actions/badge.rb
ohayon: I think we don't want to re-raise the exception here because we shouldn't be counting this as a 'crash' in our metrics. I can definitely add a comment in there!

KrauseFx: Okay, understood. However eating the exception will make it difficult to debug the problem. Do you think we can also print out the stack trace here? We've needed this a few times now, maybe it's time to have a `Helper` method that prints the original error + stack trace? In `fastlane init` we should start using that as well 👍

KrauseFx (approved):
🎷

fastlane/lib/fastlane/actions/badge.rb
```diff
@@ -16,7 +16,13 @@ def self.run(params)
           shield_gravity: params[:shield_gravity],
           shield_no_resize: params[:shield_no_resize]
         }
-        Badge::Runner.new.run(params[:path], options)
```
KrauseFx: Seems like this is a duplicate change from https://github.com/fastlane/fastlane/pull/9295/files?

KrauseFx (approved):
Nice refactor 👍

PR #9279, opened 2017-05-25
Fix puts crash on linux

ohayon:
Fix a crash happening while printing tables on linux machines.

### DIFF
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -233,11 +233,13 @@ module Fastlane
         [key, content.to_s]
       end
 
-      require 'terminal-table'
-      puts Terminal::Table.new({
-        title: "Lane Context".yellow,
-        rows: FastlaneCore::PrintTable.transform_output(rows)
-      })
+      unless Helper.linux?
+        require 'terminal-table'
+        puts Terminal::Table.new({
+          title: "Lane Context".yellow,
+          rows: FastlaneCore::PrintTable.transform_output(rows)
+        })
+      end
     end
   end
 end
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -239,8 +239,8 @@ module Fastlane
           title: "Lane Context".yellow,
           rows: FastlaneCore::PrintTable.transform_output(rows)
         })
-      rescue => e
-        os = Helper.linux? : 'linux' ? 'mac'
+      rescue
+        os = Helper.linux? ? 'linux' : 'mac'
         UI.crash!("Something went wrong trying to print the lane context on #{os}")
       end
     end
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -233,12 +233,15 @@ module Fastlane
         [key, content.to_s]
       end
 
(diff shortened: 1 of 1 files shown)


KrauseFx:
Is that the only line that's crashing with this error?

ohayon:
According to our crash reporter its happening in a few more places but not nearly as frequently. Maybe we should make this change to make sure it stops this issue and then we can go in and fix the others if this works?

hjanuschka:
whats the actual problem in terminal table (is there a stacktrace?) just hiding tables does not sound like a good plan, for all the other places of the terminal-tables.

ohayon:
@hjanuschka here is the stack trace from our crash reporter
```
[EXCEPTION]: Errno::EIO: Input/output error @ io_write - <STDOUT>
[fastlane_path]/fastlane/lib/fastlane/lane_manager.rb:237:in `write'
[fastlane_path]/fastlane/lib/fastlane/lane_manager.rb:237:in `puts'
[fastlane_path]/fastlane/lib/fastlane/lane_manager.rb:237:in `puts'
[fastlane_path]/fastlane/lib/fastlane/lane_manager.rb:237:in `print_lane_context'
[fastlane_path]/fastlane/lib/fastlane/lane_manager.rb:58:in `rescue in cruise_lane'
[fastlane_path]/fastlane/lib/fastlane/lane_manager.rb:51:in `cruise_lane'
[fastlane_path]/fastlane/lib/fastlane/command_line_handler.rb:30:in `handle'
[fastlane_path]/fastlane/lib/fastlane/commands_generator.rb:104:in `block (2 levels) in run'
[gem_home]/gems/commander-fastlane-4.4.4/lib/commander/command.rb:178:in `call'
[gem_home]/gems/commander-fastlane-4.4.4/lib/commander/command.rb:153:in `run'
[gem_home]/gems/commander-fastlane-4.4.4/lib/commander/runner.rb:476:in `run_active_command'
[fastlane_path]/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb:38:in `run!'
[gem_home]/gems/commander-fastlane-4.4.4/lib/commander/delegates.rb:15:in `run!'
[fastlane_path]/fastlane/lib/fastlane/commands_generator.rb:303:in `run'
[fastlane_path]/fastlane/lib/fastlane/commands_generator.rb:42:in `start'
[fastlane_path]/fastlane/lib/fastlane/cli_tools_distributor.rb:66:in `take_off'
[fastlane_path]/bin/fastlane:20:in `<top (required)>'
/usr/local/bin/fastlane:23:in `load'
/usr/local/bin/fastlane:23:in `<main>'
```

hjanuschka:
awesome, just for the record, seems to be a write access to stdout that is already closed, but no clue when it comes down to such a situation (spaceship on lambda in a multihreaded env?)

hjanuschka (approved):

KrauseFx:
I'm not super happy about applying a patch without knowing why we run into this error. But it seems like we don't have a way to actually reproduce the problem on our side, so 👍 from my side

ohayon:
Totally agree with you @KrauseFx, my thought was, by adding the OS into the crash report, we could at least get some idea as to where this is coming from.

lacostej:
@ohayon shouldn't the environment be part of the crash information all the time? :)

KrauseFx:
@lacostej we currently don't track any environment data with the crash reporter.

PR #9282, opened 2017-05-25
Catch errors with app names

ohayon:
Invalid app names should be reported as user errors rather than crashes.

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -122,6 +122,9 @@ module Deliver
         details.save!
         UI.success("Successfully uploaded set of metadata to iTunes Connect")
       rescue Spaceship::ITunesConnectError => e
+        # This makes sure that we log invalid app names as user errors
+        # If another string needs to be checked here we should
+        # figure out a more generic way to handle these cases.
         if e.message.include?('App Name cannot be longer than 50 characters') || e.message.include?('The app name you entered is already being used')
           UI.user_error!(e.message)
         else
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -124,6 +124,8 @@ module Deliver
       rescue Spaceship::ITunesConnectError => e
         if e.message.include?('App Name cannot be longer than 50 characters') || e.message.include?('The app name you entered is already being used')
           UI.user_error!(e.message)
+        else
+          raise e
         end
       end
     end
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -118,8 +118,14 @@ module Deliver
 
       UI.message("Uploading metadata to iTunes Connect")
       v.save!
-      details.save!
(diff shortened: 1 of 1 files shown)


deliver/lib/deliver/upload_metadata.rb
```diff
@@ -118,8 +118,14 @@ def upload(options)
...
       v.save!
-      details.save!
-      UI.success("Successfully uploaded set of metadata to iTunes Connect")
+      begin
+        details.save!
+        UI.success("Successfully uploaded set of metadata to iTunes Connect")
+      rescue Spaceship::ITunesConnectError => e
+        if e.message.include?('App Name cannot be longer than 50 characters') || e.message.include?('The app name you entered is already being used')
+          UI.user_error!(e.message)
+        end
```
KrauseFx: Could you provide some more context for this change? In particular, why do we not re-raise the exception if the error message isn't one of the 2 messages we check for?

KrauseFx (changes requested):

deliver/lib/deliver/upload_metadata.rb
lacostej: missing `raise` when not matching?

lacostej (dismissed):

deliver/lib/deliver/upload_metadata.rb
ohayon: Was moving too fast. should have done that!

lacostej: :)

KrauseFx:
This change makes a lot of sense, however I'm worried adding more of those checks to the _fastlane_ code base. We can keep adding those `if/else`, which results in noise/smell across the classes. How do we decide on when we should make a specific check? In the example of spaceship, almost every interaction with iTunes Connect / Developer Portal might fail because of one of these reasons. 

Can we think of a simpler approach that would still help us solve the problem? Would it help if we tag all spaceship response exceptions somehow automatically?

ohayon:
I think as of now, we are just making decisions about when to add these checks based on frequent crashes that we are seeing in the crash reporter, for example, this crash has happened 102 times in the past week, so it feels like something we should be guarding against. You do make a good point that we don't want to add conditional checks all over the place, but we do want to make sure that we know when there is a crash because of a user error versus a crash because iTC is down.

lacostej:
I there a way to better distinguish the errors at a different place, maybe in the client.

E.g. ITC down should definitively result in a class or errors different from user errors.

I don't remember the API and how we wrap it though. But turning everything into a `ITunesConnectError` sounds like we might be losing some information.

ohayon:
@lacostej I think it would be a good chunk of work to introduce different error types on the client. I totally agree that iTC being down should have a different type of error, I think the major issue is that its hard for us to determine why a call to iTC is failing (is it a bad request, is their service down?) because they often don't properly use error codes to represent failures.

Also, I do think we should be handling the error in `upload_metadata.rb` as it feels more appropriate to use `UI` in that class versus in the `AppDetails` model object or the client itself.

KrauseFx (approved):

PR #9284, opened 2017-05-25
Warn user when before_all, after_all or error block overwrite each other

KrauseFx:
We don't want to fail, as people probably abuse the system somehow and overwrite those blocks. To not introduce a breaking change, but still help users detect this if they did that by mistake, let's show an error message for those cases.

screenshot 2017-05-25 11 43 21


ohayon (approved):
This is a great idea!

Issue #9275, opened 2017-05-25
download_dsyms - Could not find app with bundle identifier


brunotrox:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### download_dsyms - Could not find app with bundle identifier.
But it seems that `upload_symbols_to_crashlytics` works even after `download_dsyms` error.

#
       [10:16:53]: --- Step: download_dsyms ---
       [10:16:53]: ----------------------------
       [10:16:53]: To not be asked about this value, you can specify it using 'username'
       [10:16:53]: Your Apple ID Username for iTunes Connect: <my_account>
       [10:17:21]: Login to iTunes Connect (<my_account>)
       [10:17:23]: Login successful
       [10:17:23]: Looking for dSYM files for

       [10:17:24]: Could not find app with bundle identifier '' <my_account>

       +------+-------------------------------------+-------------+
       |                     fastlane summary                     |
       +------+-------------------------------------+-------------+
       | Step | Action                              | Time (in s) |
       +------+-------------------------------------+-------------+
       | 1    | Verifying required fastlane version | 0           |
       | 2    | default_platform                    | 0           |
       | 3    | prep                                | 0           |
       | 4    | cocoapods                           | 5           |
       | 5    | gym                                 | 239         |
       | 6    | changelog_from_git_commits          | 0           |
       | 7    | pilot                               | 1536        |
       | 💥   | download_dsyms                      | 30          |
       +------+-------------------------------------+-------------+

       [10:17:24]: fastlane finished with errors

       [!] Could not find app with bundle identifier '' on account <my_account>

### Environment

      Upload to TestFlight.
      pilot(
          wait_processing_interval: 60,
          beta_app_feedback_email: <my_account>,
          distribute_external: true,
          groups: ["Internal Nightly Build Users", "Testlio"]
	)

	download_dsyms                  # Download dSYM files from iTunes Connect
	upload_symbols_to_crashlytics   # Upload them to Crashlytics
	clean_build_artifacts           # Delete the local dSYM files

        <details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.12.4                                                        |
| Ruby                        | 2.0.0                                                          |
| Bundler?                    | false                                                          |
[...]

PR #9288, opened 2017-05-25
Update download_dsyms.rb

taquitos:
Makes app identifier a required option, so we catch stuff like https://github.com/fastlane/fastlane/issues/9275


KrauseFx (approved):
👍

KrauseFx:
Don't forget to delete the branch 😉

PR #9291, opened 2017-05-26
Automatic code signing error message not clear

taquitos:
Update some copy so the errors are a little more clear


KrauseFx (approved):
Nice catch ☺

PR #9297, opened 2017-05-26
Fix supply uploading screenshot html

### LINKED ISSUE #9292: Avoid uploading screenshots.html when running fastlane supply
### New Issue Checklist

- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

After doing lots of screenshots on lots of terminals with lots of locales, we tried to deploy the screenshots using `fastlane supply`. After 15 minutes uploading things, it crashes cos it doesn't know what to do with `screenshots.html`, the report that fastlane screengrab generates to report the screenshots taken. The key point is that supply could be smart enough to ignore sending that file to Google Play.

The error launched by fastlane:
`/Users/roc/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/google-api-client-0.9.28/lib/google/apis/core/http_command.rb:211:in `check_status': [!] unsupportedListingLanguage: Listing language screenshots.html is not supported. (Google::Apis::ClientError)`

Thanks for your work! :+1:

joshdholtz:
Fixes #9292 

`all_languages` was searching meta directory for all folders **AND** files. This fix now filters `all_languages` only for directories

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -162,7 +162,7 @@ module Supply
       end
     end
 
-    # returns only language directories from metadata_path 
+    # returns only language directories from metadata_path
     def all_languages
       Dir.entries(metadata_path)
          .select { |f| File.directory? File.join(metadata_path, f) }
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -162,8 +162,12 @@ module Supply
       end
     end
 
+    # returns only language directories from metadata_path 
     def all_languages
-      Dir.foreach(metadata_path).sort { |x, y| x <=> y }
+      Dir.entries(metadata_path)
+         .select { |f| File.directory? File.join(metadata_path, f) }
+         .reject { |f| f.start_with?('.') }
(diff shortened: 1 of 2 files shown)


supply/lib/supply/uploader.rb
```diff
@@ -162,8 +162,12 @@ def update_track(apk_version_codes)
       end
     end
 
+    # returns only language directories from metadata_path
     def all_languages
-      Dir.foreach(metadata_path).sort { |x, y| x <=> y }
+      Dir.entries(metadata_path)
```
KrauseFx: Does this change have any side? Would that break potential setups?

mfurtak (approved):
Looks good. Thanks for the fix 👍

supply/lib/supply/uploader.rb
joshdholtz: @KrauseFx This _should not_ have any negative side effect. The existing setup was already breaking since this was picking up an HTML file 😉

rocboronat: :+1:

PR #9299, opened 2017-05-27
Add a nil check before asking for the shell-escaped version of path in create_keychain

orta:
Someone, _myself_, set up the ENV vars wrong on our CI. More importantly, this passed nil into the values for create_keychain. Which crashes the action as it assumes non-nil values. I ended up digging around in a bunch of the wrong places before deciding to consult the code instead.

So, I gave it a quick look over for the potential of `nil` values 👍 

### Checklist
- [x] I've once read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [n/a] I've updated the documentation if necessary.

### DIFF
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -8,13 +8,17 @@ module Fastlane
 
     class CreateKeychainAction < Action
       def self.run(params)
+        if params[:password].nil?
+          UI.user_error!("You have to pass a non-nil value to :password")
+        end
+
         escaped_password = params[:password].shellescape
 
         if params[:name]
           escaped_name = params[:name].shellescape
           keychain_path = "~/Library/Keychains/#{escaped_name}"
         else
-          keychain_path = params[:path].shellescape
+          keychain_path = params[:path] && params[:path].shellescape
         end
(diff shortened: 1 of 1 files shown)


KrauseFx:
Ha, nice catch. Actually the fact that this `ConfigItem` is non-optional, there should **never** be a `nil` value there. It seems like that's a bug inside the `config_item.rb` or `configuration.rb` then. How can you pass a nil value via an environment variable? I thought the "worst" thing you can do is to pass an empty string?

azav:
How can you pass a nil var in this case?  I'm going to assume either by omitting it, not assigning it a value or by using the wrong var name.

Just my guesses. I've never (cough, cough) done the same (cough).

Alex Zavatone

orta:
Yeah, I also typo'd it, ideally that would have neem raised at the action-handling level rather than at this level of implmentation

ohayon:
Hey @orta - I'm just working through some old PR's and was wondering if we still wanted to merge this? Thanks! 🚀

orta:
I'm 50/50 - I think the precautions are enough as-is ATM

ohayon:
Thanks for the reply @orta ! 🐙

PR #9303, opened 2017-05-28
Fix scan’s custom report name not working with multiple output types

lyndsey-ferguson:
Fixes Issue #9293 where `scan`'s custom report filename was being used just for the `html` file and not for the `xml` file as it used to. This ensures that someone will not get a nasty surprise after `scan` completes and provides the report files.

One should use `:output_files` instead and this makes it obvious.

### Checklist

### Motivation and Context

Using the `scan` option without specifying the `output_types` will, by default, try and provide a `html` and `junit` report file. Combined with the `custom_report_file_name` used to yield the `xml` `junit` file with the given name. With a recent update, that stopped working and the `html` file was yielded with the custom name: that is confusing when you want it for the `xml` file.

This change makes it so that if you provide multiple output types, and use this deprecated option, you'll get an error and not accidentally archive the `html` file thinking that you get a `junit` `xml` file.

### Description

This adds a check for the options when generating them to ensure that one is not using options that will cause you problems later on.

### DIFF
--- a/scan/lib/scan/detect_values.rb
+++ b/scan/lib/scan/detect_values.rb
@@ -9,6 +9,8 @@ module Scan
       # First, try loading the Scanfile from the current directory
       config.load_configuration_file(Scan.scanfile_name)
 
+      prevalidate
+
       # Detect the project
       FastlaneCore::Project.detect_projects(config)
       Scan.project = FastlaneCore::Project.new(config)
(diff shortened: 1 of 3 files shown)


lyndsey-ferguson:
@mfurtak @hjanuschka can I get a review? If it looks good, I could use a merge too.

mfurtak (approved):

mfurtak:
🚀

Issue #6875, opened 2016-11-05
deliver download_metadata always prompts to overwrite the folder


barnaclebarnes:
### New Issue Checklist

- [* ] Updated fastlane to the latest version
- [* ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

running `deliver download_metadata` always prompts to overwrite the folder even when it doesn't exist. This means that you cannot run the command unattended. It will always ask to overwrite the folder.

##### Complete output when running fastlane, including the stack trace and command used

```
➜  mytours-ios git:(experiment/fastlane) ✗ deliver download_metadata
[17:33:31]: Successfully loaded '/Users/barnacle/Documents/mytours/mytours-ios/Deliverfile' 📄

+------------------+-----------------------------+
|      Detected Values from './Deliverfile'      |
+------------------+-----------------------------+
| app_identifier   | com.example.app1036      |
| username         | xxx@example.com |
| metadata_path    | downloads/metadata/1036     |
| screenshots_path | downloads/screenshots/1036  |
+------------------+-----------------------------+

[17:33:31]: Login to iTunes Connect (xxx@example.com)
[17:33:33]: Login successful

+-----------------------+-----------------------------+
|               deliver 1.15.0 Summary                |
+-----------------------+-----------------------------+
| app_identifier        | com.mytoursapp.app1036      |
| username              | xxx@example.com |
| metadata_path         | downloads/metadata/1036     |
| screenshots_path      | downloads/screenshots/1036  |
| skip_binary_upload    | false                       |
| skip_screenshots      | false                       |
| skip_metadata         | false                       |
| force                 | false                       |
| submit_for_review     | false                       |
| automatic_release     | false                       |
| overwrite_screenshots | false                       |
+-----------------------+-----------------------------+

Do you want to overwrite existing metadata on path '/Users/barnacle/Documents/mytours/mytours-ios/downloads/metadata/1036'? (y/n)
```
### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                 | Value                                              |
| ------------------- | -------------------------------------------------- |
| OS                  | 10.12.1                                            |
| Ruby                | 2.3.1                                              |
| Bundler?            | true                                               |
| Xcode Path          | /Applications/Xcode.app/Contents/Developer/        |
| Xcode Version       | 8.1                                                |
| Git                 | git version 2.10.1                                 |
[...]

Issue #9306, opened 2017-05-29
deliver download_metadata requires to force overwrite even if folder didn't exist


lacostej:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

````
rm -r fastlane/metadata/
fastlane deliver download_metadata --metadata_path "fastlane/metadata" 
[...]
[09:27:55]: ▸ [09:27:55]: Do you want to overwrite existing metadata on path '/Users/lacostej/Code/WWTK/12+/fastlane/metadata'?
````

This is caused by
````
DetectValues.ensure_folders_created
DetectValues.run!
Runner.new
CommandsGenerator command :download_metadata
````

joshdholtz:
@lacostej So are you looking for a `force` flag of some sort on `download_metadata` then? 😇

joshdholtz:
@ohayon Thanks for the reference 😉

lacostej:
@joshdholtz I would like ensure_folders_created to not be created before we check if we can overwrite them...

joshdholtz:
@lacostej Makes sense! Didn't know you had PR started for this but @ohayon linked me to it so I got it after that 😇

ohayon:
Hey @joshdholtz and @lacostej cool if I close this as it looks like its been resolved? Let me know if thats an issue! 🚀

PR #9307, opened 2017-05-29
#6875 deliver: also use --force as a way to overwrite metadata

lacostej:
This is related to #6875 

Deliver requires a specific environment variable to overwrite metadata
Users expect --force to work. We add support for it, while keeping backward compatibility.
Documentation isn't updated as the env. variable wasn't made public.

### DIFF
--- a/deliver/lib/deliver/commands_generator.rb
+++ b/deliver/lib/deliver/commands_generator.rb
@@ -28,6 +28,13 @@ module Deliver
       return available_options
     end
 
+    def self.force_overwrite_metadata?(options, path)
+      res = options[:force]
+      res ||= ENV["DELIVER_FORCE_OVERWRITE"] # for backward compatibility
+      res ||= UI.confirm("Do you want to overwrite existing metadata on path '#{File.expand_path(path)}'?") if UI.interactive?
+      res
+    end
+
     # rubocop:disable Metrics/PerceivedComplexity
     def run
       program :name, 'deliver'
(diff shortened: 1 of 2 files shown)


KrauseFx (approved):
Looks good to me, thanks @lacostej. Have you tested this locally?

KrauseFx:
@lacostej :)

ohayon:
This PR is related to https://github.com/fastlane/fastlane/issues/9306 right?

lacostej:
@ohayon #9306 is another issue that I intend to look at at some point.

lacostej:
@KrauseFx I had managed to have a bug in it... I now fixed it and tested it on a live project. Also added more tests.

deliver/lib/deliver/commands_generator.rb
```diff
@@ -28,6 +28,13 @@ def deliverfile_options(skip_verification: false)
       return available_options
     end
 
+    def self.force_overwrite_metadata?(options, path)
+      res = options[:force]
+      res ||= ENV["DELIVER_FORCE_OVERWRITE"] # for backward compatibility
+      res ||= UI.confirm("Do you want to overwrite existing metadata on path '#{File.expand_path(path)}'?")
```
ohayon: Can we add an `if UI.interactive?` to the end of this line to make sure it doesn't crash in CI?

ohayon (changes requested):

deliver/lib/deliver/commands_generator.rb
lacostej: Fixed. Good catch. I thought `confirm` was doing nothing when in `interactive` but it crashes instead :)

deliver/lib/deliver/commands_generator.rb
```diff
@@ -28,6 +28,13 @@ def deliverfile_options(skip_verification: false)
       return available_options
     end
 
+    def self.force_overwrite_metadata?(options, path)
+      res = options[:force]
+      res ||= ENV["DELIVER_FORCE_OVERWRITE"] # for backward compatibility
+      res ||= UI.confirm("Do you want to overwrite existing metadata on path '#{File.expand_path(path)}'?") unless UI.interactive?
```
ohayon: I think this should be `if` not `unless` right?

lacostej: Doh....

lacostej:
@ohayon finally revised the patch with the proper boolean logic (and fixed the tests)

lacostej:
@ohayon ping! :)

ohayon (approved):

PR #9313, opened 2017-05-30
Add raise_shell_error option for sh action

giginet:
### Checklist

### Motivation and Context

If a shell command is failure, I'd like to recover by its callback.
However, in current, failing command raises `shell_error` and whole lane would be failed.

```ruby
sh "./failure", error_callback: lambda { |result|
  sh "./success"
} # it raises shell_error, lane would be failure.
```

### Description

I added the new option `raise_shell_error` which indicates whether raising shell_error or not.
Default value is `true`.

```ruby
sh "./failure", error_callback: lambda { |result|
  sh "./success"
}, raise_shell_error: false
```

This feature related with #8531.
@mfurtak Do you think how about this solution?
Could you tell me your opinion?

### DIFF
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -6,19 +6,26 @@ module Fastlane
     # When running this in tests, it will return the actual command instead of executing it
     # @param log [Boolean] should fastlane print out the executed command
     # @param error_callback [Block] a callback invoked with the command output if there is a non-zero exit status
-    def self.sh(command, log: true, error_callback: nil)
-      sh_control_output(command, print_command: log, print_command_output: log, error_callback: error_callback)
+    def self.sh(command, log: true, error_callback: nil, raise_shell_error: true)
+      sh_control_output(command,
+                        print_command: log,
+                        print_command_output: log,
+                        error_callback: error_callback,
+                        raise_shell_error: raise_shell_error)
     end
 
-    def self.sh_no_action(command, log: true, error_callback: nil)
-      sh_control_output(command, print_command: log, print_command_output: log, error_callback: error_callback)
+    def self.sh_no_action(command, log: true, error_callback: nil, raise_shell_error: true)
+      sh_control_output(command, print_command: log,
+                        print_command_output: log,
+                        error_callback: error_callback,
+                        raise_shell_error: raise_shell_error)
     end
 
     # @param command [String] The command to be executed
     # @param print_command [Boolean] Should we print the command that's being executed
     # @param print_command_output [Boolean] Should we print the command output during execution
     # @param error_callback [Block] A block that's called if the command exits with a non-zero status
-    def self.sh_control_output(command, print_command: true, print_command_output: true, error_callback: nil)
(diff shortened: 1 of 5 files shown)


mfurtak:
Hi @giginet :wave: 

@ohayon has been working more closely on defining how the new exception types should be used, so I will defer to him on this 👍

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -52,7 +59,7 @@ def self.sh_control_output(command, print_command: true, print_command_output: t
           message += "\n#{result}" if print_command_output
 
           error_callback.call(result) if error_callback
-          UI.shell_error!(message)
+          UI.shell_error!(message) if raise_shell_error
```
joshdholtz: Should we still print out the error message if `raise_shell_error` is set to true? I think it may be beneficial to at least be able to see the error in the logs still even if it won't bubble the error up 😇  Thoughts?

ohayon:
Hey @giginet. I think that instead of adding another option to determine if the error should be raised, that we should not raise the error if an error block is provided. You're totally right that you should be able to rescue the error in the block, and if that's the case the error should not get raised.

I think that just changing it to `UI.shell_error!(message) unless error_callback` might be a bit nicer. What do you think?

fastlane/lib/fastlane/helper/sh_helper.rb
giginet: You're right.
I didn't think deeply.

I'll use instead `UI.error(message)` when this options is disabled.

giginet:
@ohayon 

> I think that just changing it to UI.shell_error!(message) unless error_callback might be a bit nicer. What do you think?

This change would work for me. However, I have no idea how other users use `error_callback`.
This change may brake backward compatibility.
Is there no problem?

ohayon:
@giginet, I think the intended use of the `error_callback` is to catch and handle the error at the call site, so having the error still be thrown seems like a bug to me. I don't think it should break compatibility. 🚀

giginet:
@ohayon Hi, I tried your suggestion.
How do you think?

joshdholtz (changes requested):
This looks solid ✊ Are there any tests that can be written around this to make sure this always works as intended? 😇

giginet:
@joshdholtz 

> Are there any tests that can be written around this to make sure this always works as intended?

No. I think tests should be added too.

However, it is difficult to write tests for sh_helper.
According to I tried, all external command executions aren't called on tests at all 😇 

https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/helper/sh_helper.rb#L32

giginet:
Another solution to test `sh_helper`, we introduce like `Helper.sh_enabled?` and use it instead of `Helper.test?`.
Stub this method only in case of testing `sh_helper`.

```ruby
module Helper
  def self.sh_enabled?
    !self.test?
  end
end
```

giginet:
I added specs for sh_helper.
https://github.com/fastlane/fastlane/pull/9313/commits/25054f4be907b4f0f3ca7c25291f2d91f131ae0d

joshdholtz (approved):
💪

giginet:
Thanks 😄

PR #9320, opened 2017-05-30
Wrap the call to puts Terminal::Table in a rescue block

ohayon:
I fixed a crash just like this last week and it looks like there is another one. This just wraps it in a rescue block so we can get more information about the crash as it happens https://github.com/fastlane/fastlane/pull/9279

### DIFF
--- a/fastlane/lib/fastlane/lane_manager.rb
+++ b/fastlane/lib/fastlane/lane_manager.rb
@@ -111,13 +111,19 @@ module Fastlane
         rows << [index, name, current[:time].to_i]
       end
 
-      puts ""
-      puts Terminal::Table.new(
-        title: "fastlane summary".green,
-        headings: ["Step", "Action", "Time (in s)"],
-        rows: FastlaneCore::PrintTable.transform_output(rows)
-      )
-      puts ""
+      begin
+        require 'terminal-table'
+        puts ""
+        puts Terminal::Table.new(
+          title: "fastlane summary".green,
+          headings: ["Step", "Action", "Time (in s)"],
+          rows: FastlaneCore::PrintTable.transform_output(rows)
+        )
+        puts ""
+      rescue
+        os = Helper.linux? ? 'linux' : 'mac'
+        UI.crash!("Something went wrong trying to print the lane context on #{os}")
+      end
     end
 
     # Lane chooser if user didn't provide a lane


fastlane/lib/fastlane/lane_manager.rb
```diff
@@ -111,13 +111,19 @@ def self.print_table(actions)
...
 
-      puts ""
-      puts Terminal::Table.new(
-        title: "fastlane summary".green,
-        headings: ["Step", "Action", "Time (in s)"],
-        rows: FastlaneCore::PrintTable.transform_output(rows)
-      )
-      puts ""
+      begin
+        require 'terminal-table'
```
joshdholtz: This looks to be a duplicate require from above? ^

fastlane/lib/fastlane/lane_manager.rb
```diff
@@ -111,13 +111,19 @@ def self.print_table(actions)
...
+        puts ""
+        puts Terminal::Table.new(
+          title: "fastlane summary".green,
+          headings: ["Step", "Action", "Time (in s)"],
+          rows: FastlaneCore::PrintTable.transform_output(rows)
+        )
+        puts ""
+      rescue
+        os = Helper.linux? ? 'linux' : 'mac'
```
joshdholtz: Is windows not an option at all?

joshdholtz (commented):
@ohayon Two questions for ya ❓ ❓

KrauseFx:
Do we have to potentially add this to every table we print? Is that something we should fix in the table directly instead?

lacostej:
I am looking at #9279, and we now added a `crash` under `print_lane_context`. This means we won't get the original exception printed out, the one from `fastlane/lib/fastlane/lane_manager.rb:58` meaning we might lose some valuable information as to why this happens.

KrauseFx:
Ohh, nice catch @lacostej, we should probably revert #9279 maybe? cc @ohayon

ohayon:
We will only miss the crash if the crash happens due to table printing though right? We can definitely revert it if you think its better, we should look into a better fix for these crashes in the table printing anyways I think.

KrauseFx:
I'd be in favor of reverting to be honest, and try to find out the underlying issue of what's happening. Do I remember correctly that we didn't get a GitHub issue from anyone about this?

ohayon:
Yeah, I would be in favor of reverting it as well. Better to find the root.

ohayon:
Here is a PR to revert the first one of these: https://github.com/fastlane/fastlane/pull/9450

PR #9329, opened 2017-05-31
Action to check Swift version for   platforms

Shashikant86:
I read that you don't accept new actions, you can reject this PR straight away  !!

### Checklist

### Motivation and Context

Swift is becoming major programming language for developing and iOS and macOS apps. It would be useful to know what version of Swift we are using. Apple 

New change 

Using this action for few month and thought it could come from fastlane 

### Description

This action would be useful to determine what version of Swift we are using for building an app. It would be great idea to check this in the `before_all` step in our build process. 

We can use this action 

` check_swift_version(version: "Apple Swift version 3.0.2")`

You can replace with your version. Check yours using 
` swift -version`

### DIFF
--- a/fastlane/lib/fastlane/actions/check_swift_version.rb
+++ b/fastlane/lib/fastlane/actions/check_swift_version.rb
@@ -1,0 +1,51 @@
+module Fastlane
+  module Actions
+    module SharedValues
+    end
+
+    class CheckSwiftVersionAction < Action
+      def self.run(params)
+        required_version = params[:version]
+        selected_version = sh("swift -version")
+        if selected_version.include?(required_version)
+          UI.success("Selected Swift version is correct: #{selected_version}")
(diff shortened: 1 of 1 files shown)


ohayon:
Hey @Shashikant86! Thanks for submitting the action. This looks like it could be a great plugin. Have you checked out the documentation about how to make plugins here? https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md I am going to close this issue for now as we do try to limit the new actions that we accept, but I would encourage you to publish this as a plugin! 🚀  🔌

Shashikant86:
@ohayon Thanks for the feedback .. I have wrote a plugin before so I am aware of the process however people prefer to use `actions` than `plugins` to avoid additional dependency in `Gemfile` No worries. I will publish it as plugin !

PR #9333, opened 2017-06-01
Provide support for macos on match/sigh/cert

lacostej:
My objective is to use the same pipeline for mac os x as we use for ios, i.e. `match`. I am open to find alternatives if this isn't accepted.

**Details**

There are two types of macstore certificates, app ditribution and installer distribution.
We add parameter type_distribution where appropriate to find it.
Note that there are no prov. profile it seems associated with installer distribution.

We store the cert and profiles under `$ROOT/osx/` in git.

Tested using
```
fastlane produce -u xxxx -a com.YYYY --skip_itc --app_name YYYY --platform mac

fastlane match appstore -a com.YYYY --platform osx --distribution_type app 
fastlane match appstore -a com.YYYY --platform osx --distribution_type installer 
```

**PR status**

Most of the code should be rather intuitive, except for the rework in the runner I guess.

* I cleaned the patches and squashed them into logical grouping (except for the last fixes). Some could be removed (e.g. related to verbosity).
* I didn't yet add unit tests for the new functionality.

* *update* there's a few known issues in the code. E.g. the environment variable name for the mac distribution cert doesn't take into account the distribution_type. Meaning if you import both app and installer in a row, you only expose one. 

**Things that could be discussed**

* whether we want to use that layout (`match/lib/match.rb`)
* the new arguments (`distribution_type` for `match`/`cert`/`sigh`)
* the change in `fastlane_core/lib/fastlane_core/cert_checker.rb` could be done differently
* the change that returns multiple certs "match: return and import multiple certs"  (`match/lib/match/runner.rb) - could it break something?
* Suffix for provisioning profiles for Mac should be different

I will test signing integration now.


KrauseFx (changes requested):
Let's hold off for a few more days to discuss this 👍 Thanks for the PR

ohayon:
We had some convo off line about this PR, I am going to close it for now as it's a bit old, but @lacostej, please ping us whenever you'd like to talk through this a bit more! 😄 🚀

PR #9349, opened 2017-06-02
Allow for case insensitive screenshot directory names

ohayon:
If directories containing screenshots are named with improper cases in the language codes, this will correct them based on languages available in iTC.

This should be followed up with similar logic in the `upload_metadata` situation where languages are being activated.

### DIFF
--- a/deliver/lib/deliver/upload_screenshots.rb
+++ b/deliver/lib/deliver/upload_screenshots.rb
@@ -32,7 +32,11 @@ module Deliver
         lng_text = "language"
         lng_text += "s" if enabled_languages.count != 1
         UI.message("Activating #{lng_text} #{enabled_languages.join(', ')}...")
-        v.save!
+        begin
+          v.save!
+        rescue ITunesConnectError
+          UI.user_error!("Unable to activate languages: #{langs}. Please verify that your language codes are available in iTunesConnect. See https://developer.apple.com/library/content/documentation/LanguagesUtilities/Conceptual/iTunesConnect_Guide/Chapters/AppStoreTerritories.html for more information.")
+        end
         # This refreshes the app version from iTC after enabling a localization
         v = app.edit_version
       end
@@ -75,6 +79,10 @@ module Deliver
       screenshots = []
       extensions = '{png,jpg,jpeg}'
 
+      available_languages = Spaceship::Tunes.client.available_languages.each_with_object({}) do |lang, lang_hash|
+        lang_hash[lang.downcase] = lang
+      end
+
       Loader.language_folders(path).each do |lng_folder|
         language = File.basename(lng_folder)
 
@@ -91,7 +99,9 @@ module Deliver
 
         UI.important("Framed screenshots are detected! 🖼 Non-framed screenshot files may be skipped. 🏃") if prefer_framed
 
-        language = File.basename(lng_folder)
+        language_dir_name = File.basename(lng_folder)
+        language = available_languages[language_dir_name.downcase] || language_dir_name
+
         files.each do |file_path|
           is_framed = file_path.downcase.include?("_framed.")
           is_watch = file_path.downcase.include?("watch")
--- a/deliver/lib/deliver/upload_screenshots.rb
+++ b/deliver/lib/deliver/upload_screenshots.rb
@@ -32,11 +32,7 @@ module Deliver
         lng_text = "language"
         lng_text += "s" if enabled_languages.count != 1
         UI.message("Activating #{lng_text} #{enabled_languages.join(', ')}...")
-        begin
-          v.save!
-        rescue ITunesConnectError
-          UI.user_error!("Unable to activate languages: #{langs}. Please verify that your language codes are available in iTunesConnect. See https://developer.apple.com/library/content/documentation/LanguagesUtilities/Conceptual/iTunesConnect_Guide/Chapters/AppStoreTerritories.html for more information.")
-        end
+        v.save!
         # This refreshes the app version from iTC after enabling a localization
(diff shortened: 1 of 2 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

deliver/lib/deliver/upload_screenshots.rb
```diff
@@ -32,7 +32,11 @@ def upload(options, screenshots)
         lng_text = "language"
         lng_text += "s" if enabled_languages.count != 1
         UI.message("Activating #{lng_text} #{enabled_languages.join(', ')}...")
-        v.save!
+        begin
+          v.save!
+        rescue ITunesConnectError
+          UI.user_error!("Unable to activate languages: #{langs}. Please verify that your language codes are available in iTunesConnect. See https://developer.apple.com/library/content/documentation/LanguagesUtilities/Conceptual/iTunesConnect_Guide/Chapters/AppStoreTerritories.html for more information.")
```
mpirri: Are there any other errors that ITunesConnectError might indicate here that this message is not applicable to?

mpirri (approved):
Looks awesome, one small question only.

deliver/lib/deliver/upload_screenshots.rb
ohayon: Its definitely possible that this could be something else, but this is the most likely.

joshdholtz (approved):

deliver/lib/deliver/upload_screenshots.rb
```diff
@@ -91,7 +99,9 @@ def collect_screenshots_for_languages(path)
 
         UI.important("Framed screenshots are detected! 🖼 Non-framed screenshot files may be skipped. 🏃") if prefer_framed
 
-        language = File.basename(lng_folder)
+        language_dir_name = File.basename(lng_folder)
+        language = available_languages[language_dir_name.downcase] || language_dir_name
```
taquitos: instead of using `language_dir_name` if `available_languages[language_dir_name.downcase]` returns `nil`, would it make sense to raise an error here that you aren't using a supported language?

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -523,6 +523,12 @@ def supported_countries
       parse_response(r, 'data')
     end
 
+    def available_languages
+      r = request(:get, "ra/apps/storePreview/regionCountryLanguage")
+      response = parse_response(r, 'data')
+      response.flat_map { |region| region["storeFronts"] }.flat_map { |storefront| storefront["supportedLocaleCodes"] }.uniq
```
taquitos: formatting question, I usually prefer:
```
response.flat_map { |region| region["storeFronts"] }
        .flat_map { |storefront| storefront["supportedLocaleCodes"] }
        .uniq
```

what's the preferred formatting?

deliver/lib/deliver/upload_screenshots.rb
mpirri: Do we know that the absence of a language from that endpoint's list means the language isn't supported by Apple? I think I like the idea of falling back to previous behavior in this case (and not blocking the user), but there is an argument to be made for logging a warning here, maybe?

spaceship/lib/spaceship/tunes/tunes_client.rb
ohayon: Very RAC-swifty of you! Will change that for sure 😄

deliver/lib/deliver/upload_screenshots.rb
ohayon: Yeah!

spaceship/lib/spaceship/tunes/tunes_client.rb
taquitos: Hah, I got that from @JaviSoto

deliver/lib/deliver/upload_screenshots.rb
taquitos: Ohh, good point, I like that idea. Warn here instead of causing an error.

deliver/lib/deliver/upload_screenshots.rb
```diff
@@ -100,7 +96,12 @@ def collect_screenshots_for_languages(path)
         UI.important("Framed screenshots are detected! 🖼 Non-framed screenshot files may be skipped. 🏃") if prefer_framed
 
         language_dir_name = File.basename(lng_folder)
-        language = available_languages[language_dir_name.downcase] || language_dir_name
+
+        if available_languages[language_dir_name.downcase].nil?
+          UI.user_error!("#{language_dir_name} is not an available language. Please verify that your language codes are available in iTunesConnect. See https://developer.apple.com/library/content/documentation/LanguagesUtilities/Conceptual/iTunesConnect_Guide/Chapters/AppStoreTerritories.html for more information.")
```
taquitos: do we want to start with a warning first? @mpirri mentioned we might want to default to old behavior here, but just show a warning.

ohayon: not sure i understand what you mean? showing a warning would still result in a crash right?

taquitos: I was thinking somehow presenting just a warning but allowing deliver to continue. If we're 100% sure we can't continue past this point if the language isn't available, then cool `UI.user_error!` is totally 💯

PR #9368, opened 2017-06-05
Fix spaceship app_version to work with updated iTC

KrauseFx:
Fixes a bug that was introduced with the new version of iTunes Connect

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -811,7 +811,9 @@ module Spaceship
         result = []
 
         display_families.each do |display_family|
-          trailer_data = display_family.fetch("trailer", {}).fetch("value")
+          trailer_raw = display_family["trailer"]
+          next if trailer_raw.nil?
+          trailer_data = trailer_raw["value"]
           next if trailer_data.nil?
           data = {
             device_type: display_family['name'],


taquitos (approved):
🐑

spaceship/lib/spaceship/tunes/app_version.rb
```diff
@@ -811,7 +811,9 @@ def setup_trailers_for(row)
         result = []
 
         display_families.each do |display_family|
-          trailer_data = display_family.fetch("trailer", {}).fetch("value")
+          trailer_raw = display_family["trailer"]
```
lacostej: The problem was in the second part as we could have written:
```
trailer_data = display_family.fetch("trailer", {})["value"]
or
trailer_data = display_family.fetch("trailer", {}).fetch("value", nil)
```

right?

matteomartinelli:
Does not seem to work, I have monkey patched the last fastlane to make it works!

Since iOS11 you can upload more than one app preview, you should also change `app_trailer.rb` a bit (for instance, what about `sortPosition`?), am I wrong?

Moreover, what about:
```
      def trailer_data_for_language_and_device(language, device)
        container_data_for_language_and_device("appTrailers", language, device)
      end
```

It looks like now is:
```
      def trailer_data_for_language_and_device(language, device)
        container_data_for_language_and_device("trailers", language, device)
      end
```

PR #9372, opened 2017-06-05
Add support for enabling/disabling phased release

KrauseFx:
This PR adds support for toggling phased rollouts for releases in _spaceship_, next step is to add support for it as part _deliver_ 🚀

Also it looks like we have even more control about the rollout plan, like the # of days of the rollout

screen shot 2017-06-05 at 3 43 57 pm

screen shot 2017-06-05 at 3 44 35 pm


thebugcode:
Awesome work felix. The play store has had this for some time, so I'm happy to see it in ITC as well. Ultra-fast development speed man

taquitos (commented):
Trying to think of a way we can test this. I don't think there is, but what do you think?

taquitos (approved):
🐑 it

PR #9388, opened 2017-06-06
Update app only if we uploaded a new icon

vronin:


vronin:
I signed it!

vronin:
BTW. I think it solves another problem too. Right now, each time when delivery happens it will update version of an application prepared for submission.

KrauseFx (approved):
Nice, this is a great improvement 👍 Saving time 🚀

vronin:
@KrauseFx Thank you :)

PR #9406, opened 2017-06-07
Fix 9402 import cert not escaping special chars

### LINKED ISSUE #9402: import certificate can fail if there are special characters in keychain_password
### New Issue Checklist

### Issue Description

When a keychain password has special characters, it isn't shell-escaped and it is not counted as the same password. The command, `security set-key-partition-list` will fail with:
```
security: SecKeychainItemSetAccessWithPassword: The user name or passphrase you entered is not correct.
```

N/A

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

__I won't be posting the output here.__

The fix is fairly simple, but I want to add some debug-ability and update the code to fail when the command fails. To fix this, `FastlaneCore::KeychainImporter.import_file` must `shellescape` the `keychain_password` when passing it to the `security set-key-partition-list` command. I will open up a PR shortly, this is for a fastlane FYI

lyndsey-ferguson:
### Checklist

### Motivation and Context

This fixes https://github.com/fastlane/fastlane/issues/9402

### Description

I updated the code to run `shellescape` on the `keychain_password` so that the password used to create the keychain (which uses `shellescape`) is the same as the password used to import the certificate into the keychain.

I also added a bit of code to aid debugging by printing out the commands when _creating_ the keychain.

### DIFF
--- a/fastlane/spec/actions_specs/import_certificate_spec.rb
+++ b/fastlane/spec/actions_specs/import_certificate_spec.rb
@@ -1,6 +1,10 @@
 describe Fastlane do
   describe Fastlane::FastFile do
     describe "Import certificate Integration" do
+      before(:each) do
+        allow(FastlaneCore::Helper).to receive(:backticks).with('security -h | grep set-key-partition-list', print: false).and_return('    set-key-partition-list               Set the partition list of a key.')
+      end
+
       it "works with certificate and password" do
         cert_name = "test.cer"
         keychain = 'test.keychain'
@@ -10,7 +14,7 @@ describe Fastlane do
         expected_command = "security import #{cert_name} -k '#{keychain_path}' -P #{password} -T /usr/bin/codesign -T /usr/bin/security &> /dev/null"
 
         # this command is also sent on macOS Sierra and we need to allow it or else the test will fail
-        allowed_command = "security set-key-partition-list -S apple-tool:,apple: -k \"\" #{keychain_path.shellescape} &> /dev/null"
+        allowed_command = "security set-key-partition-list -S apple-tool:,apple: -k '' #{keychain_path.shellescape} &> /dev/null"
 
         allow(File).to receive(:exist?).and_return(false)
         allow(File).to receive(:exist?).with(keychain_path).and_return(true)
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/import_certificate.rb
```diff
@@ -23,6 +23,7 @@ def self.available_options
                                        env_name: "FL_IMPORT_CERT_KEYCHAIN_PASSWORD",
                                        description: "The password for the keychain. Note that for the login keychain this is your user's password",
                                        sensitive: true,
+                                       default_value: "",
```
lyndsey-ferguson: Otherwise, `nil` is passed into `keychain_importer`

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -22,31 +24,31 @@ def self.run(params)
...
-            Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN] = Fastlane::Actions.sh("security default-keychain", log: false).strip
+            Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN] = Fastlane::Actions.sh("security default-keychain", log: should_log).strip
           rescue
           end
-          commands << Fastlane::Actions.sh("security default-keychain -s #{keychain_path}", log: false)
+          commands << Fastlane::Actions.sh("security default-keychain -s #{keychain_path}", log: should_log)
         end
 
-        commands << Fastlane::Actions.sh("security unlock-keychain -p #{escaped_password} #{keychain_path}", log: false) if params[:unlock]
+        commands << Fastlane::Actions.sh("security unlock-keychain -p #{escaped_password} #{keychain_path}", log: should_log) if params[:unlock]
```
mfurtak: Shell commands that contain passwords should ideally never have the possibility of making it into logs since often users can't control security for those logs (on hosted CI, for example).

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -10,6 +10,8 @@ class CreateKeychainAction < Action
       def self.run(params)
         escaped_password = params[:password].shellescape
 
+        should_log = FastlaneCore::Globals.verbose?
```
mfurtak: I have some concerns about logging the commands and responses to `security`, as they could wind up as part of logs on a third party service.

If the purpose of this logging is to aid in debugging, I think we may want to control this logging via a more specific control than `--verbose`. My fear is that people are already using `--verbose` and suddenly start logging sensitive information that they were not before.

Does it make sense to trigger this with a specific environment variable, perhaps?

mfurtak (changes requested):
The fix is great, but I have some questions/thoughts about the surrounding changes :+1:

fastlane/lib/fastlane/actions/import_certificate.rb
lyndsey-ferguson: You are spot on, I did not know about the `to_s` beauty

fastlane/lib/fastlane/actions/create_keychain.rb
lyndsey-ferguson: Oh, of course

lyndsey-ferguson:
@mfurtak comments addressed. If everything else is okay, ready to merge.

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -22,31 +24,31 @@ def self.run(params)
         end
 
         commands = []
-        commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{keychain_path}", log: false)
+        commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{keychain_path}", log: should_log)
```
mfurtak: I'd recommend using `Fastlane::Actions.sh_control_output(command, print_command: false, print_command_output: true)` here, just to always bias towards not logging the password.

Feel free to prepend that with a special call to `UI.command(command)` with the password parameter removed (when the  env var is present) if you want to get the rest of the command logged.

fastlane/lib/fastlane/actions/create_keychain.rb
mfurtak: I'd recommend using `Fastlane::Actions.sh_control_output(command, print_command: false, print_command_output: true)` here, just to always bias towards not logging the password.

Feel free to prepend that with a special call to `UI.command(command)` with the password parameter removed (when the  env var is present) if you want to get the rest of the command logged.

mfurtak:
Last comments were typed directly into GitHub - please verify in practice :smile:

lyndsey-ferguson:
@mfurtak comments addressed, ready to go!

mfurtak (approved):
🙌  Thanks!

lyndsey-ferguson:
@mfurtak tests pass, you may want to review the change I made to allow it to `expect` the `expected_set_key_partition_list_command`

PR #9421, opened 2017-06-09
LanguageItem keys might be under `localeCode` too

taquitos:
### Checklist

Some properties that use LanguageItem aren't properly returning their :keys because they are stored under `localeCode` instead of `languages`


KrauseFx (commented):
The change looks nice. Is there a way to add a test that would fail without this change?

taquitos:
@KrauseFx 👍

KrauseFx (approved):
Beautiful tests 👍

Issue #9341, opened 2017-06-02
commit_version_bump broken when a project's Settings.bundle is in a group


jdee:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I recently had to [fix a problem](https://github.com/jdee/settings-bundle/pull/13) in the settings_bundle plugin that breaks `commit_version_bump` because the same poor assumptions are reflected in code I'm responsible for there: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/commit_version_bump.rb#L215. I'll submit a PR to fix, but I'd like to do something better than just repeat the plugin logic and avoid having to modify Fastlane proper in the future for a plugin error. Also, it's possible for settings bundles to have nonstandard names, like MySettings.bundle, which I also recently implemented in the plugin. While fixing this issue it might be best to think more generally.

I have two proposals:

1. Add an option to `commit_version_bump` like `:include` or `:extra_files` to allow anything else to be included explicitly as part of the version bump.
2. Pass the modified files in the `lane_context` from any action that modifies files under SCM, e.g. `SharedValues::MODIFIED_FILES`. This could be done across the board for actions including `increment_version_number` and `increment_build_number`, which would simplify the implementation of `commit_version_bump`. Or this state could simply be used by plugins without changes to any other actions.

Personally I think I prefer option 2 and in the absence of any feedback will soon submit a PR along those lines. I'm delaying the next release of the plugin for this purpose.

##### Complete output when running fastlane, including the stack trace and command used

The lane used to test:

```ruby
lane :test do
  xcodeproj = "examples/SettingsBundleExample/SettingsBundleExample.xcodeproj"

  increment_build_number xcodeproj: xcodeproj
  update_settings_bundle xcodeproj: xcodeproj,
                         key: "CurrentAppVersion",
                         value: ":version (:build)"
  update_settings_bundle xcodeproj: xcodeproj,
                         key: "BuildDate",
                         value: Time.now.strftime("%B %-d, %Y")
  commit_version_bump xcodeproj: xcodeproj, settings: true
end
```

```
[jdee@Jimmy-Dees-MacBookPro settings-bundle (tmp)]$ bundle exec fastlane test
+---------------------------------+---------+------------------------+
|                            Used plugins                            |
+---------------------------------+---------+------------------------+
[...]

PR #9423, opened 2017-06-10
Commit version bump bug fix and enhancements

jdee:
### Checklist

### Motivation and Context

This is motivated by Issue #9341. The problem is that the `:settings` option to `commit_version_bump` doesn't work if the Settings.bundle is in a group in the Xcode project. This option has been there for some time, but it was broken when I first tried to use it. I "fixed" that bug at some point, resulting in the faulty code that's there now. This originally showed up as a bug in my settings_bundle plugin. That has been fixed but not yet released, pending resolution of the issue with `commit_version_bump`.

(_Edited_) To produce the bug:

```
git clone https://github.com/jdee/settings-bundle
cd settings-bundle
git checkout commit-version-bump-bug
bundle install
bundle exec fastlane group_test
```

To test the fix:

```
git checkout -- .
git checkout commit-version-bump-bug-fixed
bundle install
bundle exec fastlane group_test
```

Note that `bundle exec fastlane test` will succeed in both branches. The difference between SettingsBundleExample and SettingsBundleGroupExample is that in the former, the Settings.bundle is at the root of the project in Xcode, but in the latter it's under the SettingsBundleGroupExample group. The `#path` method used [here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/commit_version_bump.rb#L220) returns a value relative to the group, not the project root.

But a bigger issue is that this logic belongs in the plugin, not in this action. I suggested two options in the issue. I've implemented both. There are three stages to the fix, represented by different commits.

1. Fix the broken logic in `settings_bundle_file_path`.
2. Add an `:include` option to specify arbitrary files to include.
3. Add `lane_context[SharedValues::MODIFIED_FILES]` and `Actions::add_modified_files`.

If only some of these changes should be merged, they can easily be reverted or removed.

### Description

1. Fix the broken logic in `settings_bundle_file_path`:
  Use the `#real_path` method, which returns the absolute path to the file. This fixes the original
  bug with a tiny change.
2. Add an `:include` option to specify arbitrary files to include.
  This allows explicit inclusion of an array of relative file paths. This has the virtue that there is no
  mystery if a file is included, but more parameters are required, and the correct path has to be
  explicitly supplied. There is already an `:ignore` parameter. This parameter seems to complement
  it.
3. Add `lane_context[SharedValues::MODIFIED_FILES]` and `Actions::add_modified_files`.
  This allows any action that modifies files to be included in a commit to pass them directly to
  the `commit_version_bump` action via the lane_context. It allows the context to accumulate
  the unique results of all actions. It eliminates the need for any additional parameters to support
  plugins that make modifications to include in a version bump. It could easily be used to support
  Android as well, rewrite `commit_version_bump` and obviate the `:xcodeproj` and `:settings`
  parameters.

I mentioned in the issue that it's possible for a settings bundle to have a custom name other than Settings.bundle. This is and isn't true. It's possible to create such a thing, but it doesn't show up in the Settings app unless the name is Settings.bundle. So that's not a concern, and the first commit is enough to fix the issue and probably put the question of settings bundles to rest.

But it seems plausible that there may be other cases where a lane might modify files other than the Xcode project and Info.plist as part of a version bump. Or I might have to fix another bug in my plugin and have to change `commit_version_bump` again. A more extensible approach seems like a good idea. The other two options are presented for consideration.

(_Edited_) To try out these options, use the `commit-version-bump-bug-fixed` branch in the plugin repo above. See the [Fastfile](https://github.com/jdee/settings-bundle/blob/commit-version-bump-bug-fixed/fastlane/Fastfile) there, which has three lanes that demonstrate the three different methods.

Using the `:include` option:
```
bundle exec fastlane test
```

Using the corrected `:settings` option:
```
bundle exec fastlane group_test
```

Using the lane_context (no `:include` or `:settings` option):
```
bundle exec fastlane library_test
```

Note: `git reset --hard HEAD^` erases the last commit in the current branch and is useful for testing this action.

### DIFF
--- a/fastlane/lib/fastlane/actions/commit_version_bump.rb
+++ b/fastlane/lib/fastlane/actions/commit_version_bump.rb
@@ -63,12 +63,8 @@ module Fastlane
           end
         end
 
-        # returns an array (empty if params[:include] is nil) or nil for invalid argument
-        extra_files = include_list_from_param params[:include]
-        UI.user_error! ":include must specify a string array or a comma-separated string" if extra_files.nil?
-
         # create our list of files that we expect to have changed, they should all be relative to the project root, which should be equal to the git workdir root
-        expected_changed_files = extra_files
+        expected_changed_files = []
         expected_changed_files << pbxproj_path
         expected_changed_files << info_plist_files
 
@@ -160,11 +156,6 @@ module Fastlane
                                        description: "A regular expression used to filter matched plist files to be modified",
                                        optional: true,
                                        default_value: nil,
-                                       is_string: false),
-          FastlaneCore::ConfigItem.new(key: :include,
-                                       description: "A list of extra files to be included in the version bump (string array or comma-separated string)",
-                                       optional: true,
-                                       default_value: nil,
                                        is_string: false)
         ]
       end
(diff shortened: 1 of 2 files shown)


jdee:
@googlebot I signed it.

jdee:
There are now two branches in the settings-bundle repo: `commit-version-bump-bug` and `commit-version-bump-bug-fixed`. The new options are used in the second branch. The original instructions above have been updated.

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -201,23 +234,46 @@ def self.category
...
+          end
         end
 
         def settings_bundle_file_path(project, settings_file_name)
           settings_bundle = project.files.find { |f| f.path =~ /Settings.bundle/ }
           raise "No Settings.bundle in project" if settings_bundle.nil?
 
-          project_parent = File.dirname project.path
-          File.join(project_parent, settings_bundle.path, settings_file_name)
+          File.join settings_bundle.real_path, settings_file_name
```
ohayon: Could you put parentheses in here?

ohayon (changes requested):
Hey @jdee, thank you so much for this PR! I would really like to merge just the first commit/change that you made and then address the second and third in other PRs if that is possible. Would you mind opening up other PRs as those kind of change/add some functionality where it appears that the first really just fixes the bug? Let us know if you have any questions! 🚀

jdee:
@ohayon: Done. The first commit broke a unit test, fixed by the third. I also fixed the parentheses and added examples of the current usage of the `:settings` option, which I neglected to do the last time around. Would you prefer one PR or two for the enhancements?

ohayon (approved):
Thanks for making those changes @jdee !! 🚀

ohayon:
@jdee - I think one PR for the enhancements would be great! Thanks so much for being flexible!

Issue #9410, opened 2017-06-08
Nuke 'entreprise'?


NikKovIos:
Good day, thanks for such a cool tool!

When i type `fastlane match nuke` in console. There is an error:

    [!] Please run `fastlane match nuke [type], allowed values: distribution, enterprise and development. For the 'adhoc' type, please use 'distribution' instead.

distribution and development works well, but enterprise doesn't work. Is it a legacy error text?

Fluzzarn:
For match to work with enterprise, you need to set the environment variable MATCH_FORCE_ENTERPRISE to 1

KrauseFx:
Thanks @Fluzzarn, actually `MATCH_FORCE_ENTERPRISE` isn't needed any more.

> distribution and development works well, but enterprise doesn't work. Is it a legacy error text?

Could you provide more context on `doesn't work`?

NikKovIos:
When i write `fastlane match nuke enterprise` it throws the same error

    [!] Please run `fastlane match nuke [type], allowed values: distribution, enterprise and development. For the 'adhoc' type, please use 'distribution' instead.

Fluzzarn:
if you run If you run fastlane match nuke enterprise --verbose, what does the additional output say?

NikKovIos:
fastlane match nuke enterprise --verbose                                  ✓  1040  20:29:57
/Users/nikkov/.rvm/gems/ruby-2.4.0/gems/fastlane-2.37.0/fastlane_core/lib/fastlane_core/ui/interface.rb:131:in `user_error!': [!] Please run `fastlane match nuke [type], allowed values: distribution, enterprise and development. For the 'adhoc' type, please use 'distribution' instead. (FastlaneCore::Interface::FastlaneError)
	from /Users/nikkov/.rvm/gems/ruby-2.4.0/gems/fastlane-2.37.0/fastlane_core/lib/fastlane_core/ui/ui.rb:14:in `method_missing'
	from /Users/nikkov/.rvm/gems/ruby-2.4.0/gems/fastlane-2.37.0/match/lib/match/commands_generator.rb:112:in `block (2 levels) in run'
	from /Users/nikkov/.rvm/gems/ruby-2.4.0@global/gems/commander-fastlane-4.4.4/lib/commander/command.rb:178:in `call'
	from /Users/nikkov/.rvm/gems/ruby-2.4.0@global/gems/commander-fastlane-4.4.4/lib/commander/command.rb:153:in `run'
[...]

PR #9427, opened 2017-06-12
Fix match error message

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/9410

We don't support `match nuke` for enterprise environment as it's too risky


taquitos (approved):
Low risk is best risk.

PR #9428, opened 2017-06-12
[match] Add improved error message on match failure when branch is specified

KrauseFx:
As the `git clone` command will fail if we want to directly clone a branch that doesn't exist.


taquitos (approved):
Better error messaging is best.

PR #9430, opened 2017-06-12
Fix crash in crash reporter if backtrace isn't accessible

KrauseFx:
There is cases where the backtrace can't be accessed. We don't want to crash the crash reporter, which would then shadow the actual stack trace + error message

Crash + Stack trace with this patch applied
screenshot 2017-06-11 07 52 15

Without this patch, we shadow the actual crash.
screenshot 2017-06-11 07 52 20

I'm wondering if we can design/refactor the crash reporter in a way that it can't shadow the actual crash in any case?

### DIFF
--- a/fastlane_core/lib/fastlane_core/ui/errors/fastlane_error.rb
+++ b/fastlane_core/lib/fastlane_core/ui/errors/fastlane_error.rb
@@ -31,7 +31,7 @@ class Exception
   end
 
   def fastlane_crash_came_from_plugin?
-    plugin_frame = backtrace && backtrace.find { |frame| frame.include?('fastlane-plugin-') }
+    plugin_frame = exception && exception.backtrace && exception.backtrace.find { |frame| frame.include?('fastlane-plugin-') }
     !plugin_frame.nil?
   end
 
--- a/fastlane_core/lib/fastlane_core/ui/errors/fastlane_error.rb
+++ b/fastlane_core/lib/fastlane_core/ui/errors/fastlane_error.rb
@@ -26,12 +26,12 @@ end
 
 class Exception
   def fastlane_crash_came_from_custom_action?
-    custom_frame = exception.backtrace.find { |frame| frame.start_with?('actions/') }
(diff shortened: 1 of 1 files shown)


KrauseFx:
The crash was an error on my side while working on some fastlane things.

KrauseFx:
What surprises me actually is that I have `FASTLANE_OPT_OUT_CRASH_REPORTING` set, and still it's landing in the crash reporter code. Is that expected behavior? 👍

ohayon:
@KrauseFx how far into the crash reporter code is it getting? If you have that env var set, it should not end up posting a crash report at all.

fastlane_core/lib/fastlane_core/ui/errors/fastlane_error.rb
```diff
@@ -26,12 +26,12 @@ def could_contain_pii?
 
 class Exception
   def fastlane_crash_came_from_custom_action?
-    custom_frame = exception.backtrace.find { |frame| frame.start_with?('actions/') }
```
ohayon: This is inside of the `Exception` class, so It seems like this fix should maybe just be
```
custom_frame = backtrace.find { .... } if backtrace
```
right?

KrauseFx: Mh, I don't know, I think to 100% be sure this is available we'd need all 3 checks

ohayon: Where does the `exception` object come from here though? Isn't it just `self` because it is within a monkeypatch?

KrauseFx: I don't know where it comes from. It happens when you have a syntax error in a dynamically loaded file, e.g. a fastlane plugin. I remember I ran into the same thing when I built danger

ohayon: Oh thats super interesting. Lets give it a go!

ohayon (approved):
This is great, but should we change the implementation of `fastlane_crash_came_from_plugin?` below as well?

Issue #9380, opened 2017-06-06
Xcode 9.0 (beta) support


dennisfrett:
Currently, it seems like Xcode 9.0 beta version is not supported:

> 2017-06-06 14:02:41.579 xcodebuild[36176:463668] [MT] IDEDistribution: Step failed: <IDEDistributionSigningAssetsStep: 0x7fff3691e600>: Error Domain=IDEDistributionSigningAssetStepErrorDomain Code=0 "Locating signing assets failed." UserInfo={NSLocalizedDescription=Locating signing assets failed., IDEDistributionSigningAssetStepUnderlyingErrors=(
> "Error Domain=IDEProvisioningErrorDomain Code=9 \"\"FoodTracker.app\" requires a provisioning profile.\" UserInfo={NSLocalizedDescription=\"FoodTracker.app\" requires a provisioning profile., NSLocalizedRecoverySuggestion=Select a provisioning profile for the \"Default\" build configuration in the project editor.}"
> )}
> error: exportArchive: "FoodTracker.app" requires a provisioning profile.
> 
> Error Domain=IDEProvisioningErrorDomain Code=9 ""FoodTracker.app" requires a provisioning profile." UserInfo={NSLocalizedDescription="FoodTracker.app" requires a provisioning profile., NSLocalizedRecoverySuggestion=Select a provisioning profile for the "Default" build configuration in the project editor.}

Provisioning profile is set correctly. If I do `xcode-select -s` to Xcode 8.3.2, it finishes without a problem.

** EXPORT FAILED **
[14:02:41]: Exit status: 70

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                            |
| --------------------------- | ------------------------------------------------ |
| OS                          | 10.12.4                                          |
| Ruby                        | 2.2.4                                            |
| Bundler?                    | false                                            |
| Git                         | git version 2.12.2                               |
| Installation Source         | ~/.fastlane/bin/bundle/bin/fastlane              |
| Host                        | Mac OS X 10.12.4 (16E195)                        |
| Ruby Lib Dir                | ~/.fastlane/bin/bundle/lib                       |
| OpenSSL Version             | OpenSSL 1.0.2g  1 Mar 2016                       |
| Is contained                | false                                            |
| Is homebrew                 | true                                             |
| Is installed via Fabric.app | false                                            |
| Xcode Path                  | /Applications/Xcode-beta.app/Contents/Developer/ |
[...]

PR #9432, opened 2017-06-12
Add support for Xcode 9 by automatically using the right profile for each bundle identifier

KrauseFx:
If you're using _match_ and _gym_ together with _fastlane_, you can already use Xcode 9 out of the box 🚀

I still want a fallback for this, if the user just uses gym (why would you do that even....), because right now we're failing (see https://github.com/fastlane/fastlane/issues/9380), instead we should try to automatically determine the profile to use based on the Xcode project settings. Using match + fastlane is the recommended way as we can be 100% certain the profiles are correct, but we should still support other ways too

Fixes https://github.com/fastlane/fastlane/issues/9380

### DIFF
--- a/fastlane/lib/fastlane/actions/gym.rb
+++ b/fastlane/lib/fastlane/actions/gym.rb
@@ -19,7 +19,7 @@ module Fastlane
         values[:export_method] ||= Actions.lane_context[SharedValues::SIGH_PROFILE_TYPE]
 
         if Actions.lane_context[SharedValues::MATCH_PROVISIONING_PROFILE_MAPPING]
-          # Since Xcode 9 you need to provide the explicit provisioning profile per app target
+          # Since Xcode 9 you need to explicitly provide the provisioning profile per app target
           # If the user is smart and uses match and gym together with fastlane, we can do all
           # the heavy lifting for them
           values[:export_options] ||= {}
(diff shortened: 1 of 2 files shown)


KrauseFx:
Note: I don't know yet of what's the best way to receive the set provisioning profiles from the Xcode project

fastlane/lib/fastlane/actions/gym.rb
```diff
@@ -12,11 +12,19 @@ def self.run(values)
...
-          sigh_path = Actions.lane_context[Actions::SharedValues::SIGH_PROFILE_PATH] || ENV["SIGH_PROFILE_PATH"]
+          sigh_path = Actions.lane_context[SharedValues::SIGH_PROFILE_PATH] || ENV["SIGH_PROFILE_PATH"]
           values[:provisioning_profile_path] = File.expand_path(sigh_path) if sigh_path
         end
 
-        values[:export_method] ||= Actions.lane_context[Actions::SharedValues::SIGH_PROFILE_TYPE]
+        values[:export_method] ||= Actions.lane_context[SharedValues::SIGH_PROFILE_TYPE]
+
+        if Actions.lane_context[SharedValues::MATCH_PROVISIONING_PROFILE_MAPPING]
+          # Since Xcode 9 you need to provide the explicit provisioning profile per app target
```
taquitos: Since Xcode 9 you need to _explicitly provide_ the provisioning profile per app target

taquitos (approved):
🐑  it

taquitos:
@googlebot yeah, include the commit 🤦‍♂️

casz:
This is an awesome change 💯

PR #9440, opened 2017-06-12
Add test to ensure correct export-type in gym

KrauseFx:
Via https://github.com/fastlane/fastlane/pull/9429

I ensured that without #9429 this test fails


taquitos (approved):
AMAZING

PR #9447, opened 2017-06-13
adjusted find build method to exit early if we don't have candidate builds and show a better user error

carolanitz:
before:
screen shot 2017-06-12 at 6 46 29 pm
after:
screen shot 2017-06-12 at 7 04 12 pm


taquitos (approved):
Always a fan of better messaging. Thank you! 🐑 it!

PR #9450, opened 2017-06-13
Revert #9279.

ohayon:
We should look into the root cause of the crash happening when `Terminal::Table` prints out in the `LaneManager`


KrauseFx (approved):
Yeah 🎷

Issue #8674, opened 2017-03-28
SwiftLint action doesn't work when SwiftLint is installed from CocoaPods


kevinrenskers:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I'm using SwiftLint with the Fastlane action to run it automatically on build time. Now that I've switched SwiftLint from Homebrew to Cocoapods, Fastlane's SwiftLint action no longer works (it can't find SwiftLint). 

How can I point to the new bin location in the pods folder?

kevinrenskers:
<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.4                                     |
| Ruby                        | 2.3.0                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.11.0 (Apple Git-81)           |
| Installation Source         | ~/.rvm/gems/ruby-2.3.0/bin/fastlane         |
| Host                        | Mac OS X 10.12.4 (16E195)                   |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.3.0/lib                |
| OpenSSL Version             | OpenSSL 1.0.1u  22 Sep 2016                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 8.3                                         |


### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |


### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
opt_out_usage

fastlane_version "1.111.0"

default_platform :ios

platform :ios do
  desc "Fetches the provisioning profiles so you can build locally and deploy to your device"
  lane :certs do
    match(type: "development")
  end

  desc "Submit a new Beta Build to Apple TestFlight"
  desc "This will also make sure the profile is up to date"
  lane :beta do
    ensure_git_status_clean
    cocoapods

    # Test for linting errors (warnings are ok)
    #swiftlint

    # Increment the build number (not the version number)
    increment_build_number

[...]

PR #9451, opened 2017-06-13
Add option for the path to the swiftlint executable

ohayon:
This fixes the case mentioned in #8674 where users may have `swiftlint` installed via CocoaPods or other means rather than homebrew


KrauseFx (approved):
👍

PR #9452, opened 2017-06-13
Only use fastlane as the action for the crash reporter

ohayon:
We have been hitting limits in our crash reporter for having too many versions and actions being recorded. This PR makes sure that all errors/crashes are being logged under the _fastlane_ action. This is okay because the stack traces show from which action the error is coming anyways.


fastlane_core/spec/crash_reporter/crash_report_generator_spec.rb
```diff
@@ -27,12 +27,6 @@
       allow(Time).to receive(:now).and_return(Time.utc(0))
     end
 
-    it 'sets service to the action that crashed' do
```
ohayon: This test is no longer needed because it's always going to be `fastlane` here.

KrauseFx (approved):
I haven't tested it, but the code looks 💯

Issue #9462, opened 2017-06-13
Helper.keychain_path heuristic can cause it to return directories


lacostej:
### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

We use a custom keychain and named it `fastlane`.

Our `Fastfile` contains the name without the `.keychain` extension
```
    unlock_keychain(path: "fastlane")
```
and we expect fastlane to find it under `~/Library...`

Our project contains a `fastlane` directory.

If you haven't created the `fastlane.keychain` before running your build, the `Helper.keychain_path` call fails as

1. the method will expand the `path` argument  here https://github.com/fastlane/fastlane/blame/e2bcfe7/fastlane_core/lib/fastlane_core/helper.rb#L200
The specified path is then interpreted as relative to the current directory

2. and this check here https://github.com/fastlane/fastlane/blame/e2bcfe7/fastlane_core/lib/fastlane_core/helper.rb#L211 will return `true` as the expanded possible location matches the relative `fastlane` directory

```
"Shell command exited with exit status 181 instead of 0."
security: SecKeychainUnlock /Users/alx/Documents/Projets/gym/fastlane: UNIX[Is a directory]
```

(same symptom as for #9073)

**Workaround**:

Replacing the call to `.exist?` with `.file?` filters out this specific case and the error is now what we expect:

```
ERROR [2017-06-13 23:40:19.11]: Could not locate the provided keychain. Tried:
	/Users/alx/Library/Keychains/fastlane-db
	/Users/alx/Library/Keychains/fastlane.keychain-db
	/Users/alx/Library/Keychains/fastlane
	/Users/alx/Library/Keychains/fastlane.keychain
	/Users/alx/Documents/Projets/gym/fastlane-db
	/Users/alx/Documents/Projets/gym/fastlane.keychain-db
	/Users/alx/Documents/Projets/gym/fastlane
	/Users/alx/Documents/Projets/gym/fastlane.keychain
```

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

N/A

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

N/A

lacostej:
Fixed in #9464

PR #9464, opened 2017-06-13
Helper.keychain_path heuristic should only search for file to avoid returning directories as false positives (#9462)

lacostej:
See #9462 for details

### DIFF
--- a/fastlane/lib/fastlane/actions/unlock_keychain.rb
+++ b/fastlane/lib/fastlane/actions/unlock_keychain.rb
@@ -22,6 +22,8 @@ module Fastlane
         escaped_path = keychain_path.shellescape
         escaped_password = params[:password].shellescape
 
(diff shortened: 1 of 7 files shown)


ohayon (approved):
Great catch @lacostej !! 🚀

KrauseFx (approved):
🐑  is

lacostej:
@KrauseFx some (maybe all) of these tests fail because the internal of the `Helper` implementation are exposed to the test. E.g.

https://github.com/fastlane/fastlane/blob/e55dc348d95591def083a84bad025242c4faa985/fastlane/spec/actions_specs/delete_keychain_spec.rb#L10-L12

Should I fix those tests? E.g. by changing its implementation to fake the result to `keychain_path` instead of its implementation?

KrauseFx:
Nice 👍

PR #9465, opened 2017-06-13
automatic_code_signing defaults to nearest xcodeproj file

jetersen:
### Checklist

### Motivation and Context
This seems like a good change, type less kind of thing.

### Description
Allows you to do the as the example shows, running enable_automatic_code_signing and disable_automatic_code_signing without parameters


ohayon:
Hey @casz, thanks for the PR! I definitely agree that this could be a great change, but I am a bit nervous to merge it as changing default options is always a bit risky for current setups. What is the circumstance that you are using this for?

casz:
well it won't break any current implementation, but the documentation clearly states you should be able to just call without parameters: https://docs.fastlane.tools/actions/#automatic_code_signing
Fastfile that still defines the xcodeproj will still work and those who discover the doc won't be confused 👍 

image

ohayon (approved):
Oh, awesome! Thanks for citing those docs! 🚀

casz:
Thanks! And your welcome 🚀

PR #9468, opened 2017-06-14
Show message that Xcode 9 isn't support in snapshot yet

KrauseFx:
screenshot 2017-06-13 06 58 02


ohayon (approved):
This is great! Love the message! It will be interesting to track how many people are hitting this in the crash reporter too. 🚀

PR #9469, opened 2017-06-14
Allow to skip updating of an app version under 'Prepare for submission'

vronin:


KrauseFx:
Hey @vronin, thanks for this PR. I'm not sure I 100% understand what this option does. Could you clarify a bit more, and maybe update the docs accordingly? Thanks

vronin:
@KrauseFx I updated the docs.

What happens right now is that each time when we do deliver, it will update app version under "Prepare for submission".

This is convenient if you 'deliver' rarely and always submit the latest version to AppStore, because your "Prepare for submission" will have automatically the right version.

However, if you deliver often (dozens of times a day) and submit some older builds then it's counterproductive. 

As an example, I am preparing a version 1 for submission and at that moment new build happens, which does deliver which goes and update version for submission to 2.

Also, this is kind of a continuation of https://github.com/fastlane/fastlane/pull/9388
Deliver requires 'admin' role for a user which uploads a build. The first problem was that it tried to update an app when we explicitly skip icon upload. This was fixed for #9388. However, I missed second API call when it goes and tries to update "Prepare for submission" version.

If this new flag is set with a combination of flags to not upload icons then having a 'developer' role will be enough to just upload new builds to iTunes connect.

ohayon:
Hey @vronin - I'm curious, what is the use case for running _deliver_ so many times per day?

vronin:
@ohayon 
I think we kind of end up in this place, because of a couple of thoughts.
- We started with iOS. 
- We have feature branches+pull requests and when they are reviewed they are merged into master
- iOS builds from master branch are pushed to TestFlight. Beta testers are told which builds needs to be tested. And the useful thing that it doesn't require any additional manual steps
- As soon as we are happy with some build (tested for some period of time) we send this build to AppStore review.
- We moved to macOS next. The process is pretty much modeled on iOS. The only difference is that we use 3rd party tool instead of TestFlight, but we still push builds to iTunes connect, so we can send it to AppStore as soon as testing is finished.
- The probability of new build coming while somebody is editing a"Prepare for submission" is small, but it exists. 

It looks like other companies try to completely automate submission to AppStore and have some manual trigger (as an example tagging or branching) for that in the pipeline. We have an opposite view. We try completely automate a pipeline and have a submission to AppStore as a manual step.

ohayon:
Hey @vronin, so you submit every version for review? And when a new one gets posted for review you reject the one before?

ohayon:
Also, are you bumping the marketing version each time?

vronin:
@ohayon Actually. it's opposite. We send each build to iTunes connect. So all of them are there (just sitting there). This is done automatically. However, we submit only some of them for App review and this is done manually.

vronin:
@ohayon I think the confusion comes from a way how we think about submission process. I think about it as a three step process:
1) Upload a binary to iTunes Connect
2) Update information in "Prepare for submission" including selecting a build + version to send to AppReview.
3) Submit for App Review.

These three steps are stand alone for me. We do 1 automatically and 2 and 3 manually.

Deliver, currently, doesn't allow to separate 1 from 2. This change will allow to do 1 and skip 2.

vronin:
@KrauseFx @ohayon Hey guys. Did you have a chance to review this? Please let me know if you need more info or want to discuss this more (have to jump to the call if it will be simpler).

deliver/Deliverfile.md
```diff
@@ -38,6 +38,20 @@ Optional, as it is usually automatically detected. Specify the version that shou
...
 
+##### skip_app_version_update
+
+In the case if `deliver` uploads your application to iTunes Connect it will automatically update "Prepare for submission" app version (which could be found on iTunes Connect->My Apps->App Store page)
+
+The option allows uploading your app without updating "Prepare for submission" version. 
+
+This could be useful in the case if you are generating a lot of uploads while not submitting the latest build for Apple review.
+
+It's optional. The default value is false.
```
KrauseFx: Let's replace this line with just `The default value is false`

KrauseFx (approved):
Looks 💯 The use case makes sense, thanks for clarifying! One little note, before it's ready before it's ready to be merged

vronin:
@KrauseFx Thank you. I updated Deliverfile.md

KrauseFx:
Alright, thanks for contributing @vronin, this will be in the next release 🚀

PR #5541, opened 2016-07-28
Add ignore option to commit_version_bump action.

guidomb:
Fixes #5510 

The ignore option is an optional regular expression that, if
provided, is used to filter the list of plist files to be
committed.

PR #9483, opened 2017-06-15
Commit version bump enhancements

jdee:
### Checklist

### Motivation and Context

This PR follows #9423, which fixed a bug in the `commit_version_bump` action involving the `:settings` option. This PR is intended to fix the reasons for the bug, which results from logic that need not be in `commit_version_bump`. These changes make the action more extensible and flexible.

### Description

There are two parts to this change, represented by different commits.

1. Add an optional `:include` parameter. There is already an `:ignore` parameter. This gives devs a back door to add any other configuration files (e.g., package.json) that might also be included in a version bump by explicitly specifying them. This option takes either an array of strings or a comma-separated string.

2. Add `SharedValues::MODIFIED_FILES` and `Actions::add_modified_files`. Passing an array of either relative or absolute paths to `add_modified_files` adds them to the list of files to be expected in a commit by `commit_version_bump`. This is stored in the `lane_context` as a `Set` to avoid duplication. Multiple actions or invocations of the same action can add to the list. The aggregate list will be added to the expected changes in `commit_version_bump`. No extra arguments to `commit_version_bump` are required to enable this feature.

It can be useful for validation to spell out all the files expected to be included in a commit, but paths can be lengthy, and this is error-prone. However, the `:include` option would be useful for anyone updating a custom configuration file by shelling out to a custom script, e.g.

Using the lane context makes the action a lot more flexible and requires nothing special from the user. This approach could allow `commit_version_bump` to support Android, for example. The second approach could also be adopted in the existing `increment_build_number` and `increment_version_number` actions to avoid having to specify the `:xcodeproj` parameter in `commit_version_bump`. Modified files could be passed in the lane context, and `commit_version_bump` could become much simpler.

To try it out:

```bash
git clone https://github.com/jdee/settings-bundle
cd settings-bundle
git checkout commit-version-bump-enhancements
bundle install
bundle exec test # uses the :include parameter
bundle exec group_test # uses the current :settings parameter
bundle exec library_test # uses the lane context (no :settings or :include parameter)
```

And see the example [Fastfile](https://github.com/jdee/settings-bundle/blob/commit-version-bump-enhancements/fastlane/Fastfile) there.


jdee:
This is hardly urgent, but I had another thought about this:

If there's sufficient interest, I could just contribute the `update_settings_bundle` action to Fastlane itself and sunset the `settings_bundle` plugin. The plugin has 100% test coverage and has been through a few releases and bug fixes from the field. The `SettingsBundleHelper` is very solid and could just be used in `commit_version_bump` instead of the method in `CommitVersionBumpAction` to avoid having the same code in two places in different repos.

Of course, not every iOS app has a settings bundle, and not every app that has a settings bundle adds static title entries like the version number. So maybe that action isn't sufficiently general for Fastlane. But in that case, settings bundles will continue to be updated by a plugin, and there's this method in Fastlane that will continue to have to be maintained in parallel. That probably won't happen too often, so maybe it's not a huge problem, but it seems like there's room for improvement.

I still think the approach in this PR is nicer and can make `commit_version_bump` a lot more general. But there are plenty of options.

fastlane/spec/actions_specs/commit_version_bump_action_spec.rb
```diff
@@ -8,8 +8,8 @@
     end
 
     it 'returns the param if an Array' do
-      settings_plists = action.settings_plists_from_param ["About.plist", "Root.plist"]
-      expect(settings_plists).to eq ["About.plist", "Root.plist"]
```
ohayon: Just curious why this change is in here?

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -227,6 +275,16 @@ def settings_bundle_file_path(project, settings_file_name)
...
         end
+
+        def modified_files_relative_to_repo_root(repo_root)
+          return [] if Actions.lane_context[SharedValues::MODIFIED_FILES].nil?
+
+          root_pathname = Pathname.new repo_root
+          Actions.lane_context[SharedValues::MODIFIED_FILES].map do |path|
+            next path unless path =~ %r{^/}
+            Pathname.new(path).relative_path_from(root_pathname).to_s
+          end.uniq
```
ohayon: Could we separate this out into another line just to make sure that it is totally clear whats being returned here?

ohayon (changes requested):
Hey @jdee - apologies for reviewing this so late, it's definitely a cool change, I just left a couple of comments. Let me know what you think! 🚀

fastlane/spec/actions_specs/commit_version_bump_action_spec.rb
jdee: Thanks for the feedback, @ohayon. Maybe I changed this at some point and then reverted it by hand. I can revert this change if desired. I don't have very strong opinions about style. I leave that to RuboCop. Though the `%w` notation is always a little bit shorter.

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -208,15 +242,29 @@ def self.category
...
-            return param
+            # commit_version_bump settings: ["Root.plist", "About.plist"]
+            param
+          else
+            # commit_version_bump settings: true # Root.plist
+            ["Root.plist"]
           end
+        end
+
+        def include_list_from_param(param)
```
jdee: This is unnecessary. This should just be an Array param.

jdee:
I hadn't looked at this in a long time. The `include_list_from_param` method was unnecessary because I had not yet discovered the joys of `FastlaneCore::ConfigurationItem#auto_convert_value`. This is one reason I want to add some stuff to the docs on using built-in configuration tools and so on. I'll see what I can come up with this weekend.

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -208,15 +239,15 @@ def self.category
...
+            # commit_version_bump settings: "About.plist"
+            [param]
           elsif param.kind_of? Array
-            # commit_version_bump xcodeproj: "MyProject.xcodeproj", settings: [ "Root.plist", "About.plist" ]
-            return param
+            # commit_version_bump settings: ["Root.plist", "About.plist"]
+            param
+          else
+            # commit_version_bump settings: true # Root.plist
+            ["Root.plist"]
```
ohayon: Can we put back the explicit return statements here?

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -225,6 +256,17 @@ def settings_bundle_file_path(project, settings_file_name)
...
+
+        def modified_files_relative_to_repo_root(repo_root)
+          return [] if Actions.lane_context[SharedValues::MODIFIED_FILES].nil?
+
+          root_pathname = Pathname.new repo_root
+          all_modified_files = Actions.lane_context[SharedValues::MODIFIED_FILES].map do |path|
+            next path unless path =~ %r{^/}
+            Pathname.new(path).relative_path_from(root_pathname).to_s
+          end
+          all_modified_files.uniq
```
ohayon: can we add a return statement in here?

ohayon (changes requested):
Hey @jdee this is great! thanks for responding to that feedback. I just left a few more comments. We are trying to get in the habit of adding explicit returns so its as clear as possible about what's happening in methods. Thanks !! 🚀

jdee:
@ohayon Done. We can discuss it elsewhere, but I'm not sure I'm in favor of the idea. Then again, I still prefer `$:.unshift`. Anyway, good to keep things uniform.

fastlane/lib/fastlane/actions/commit_version_bump.rb
```diff
@@ -61,8 +78,11 @@ def self.run(params)
           end
         end
 
+        extra_files = params[:include]
+        extra_files += modified_files_relative_to_repo_root repo_path
```
ohayon: Ah! sorry I missed this before @jdee - can you just wrap this parameter in parens?

ohayon (changes requested):
Sorry, @jdee I missed on thing with the parens and then I realized something else. It looks like there is now an `ignore` option, but I'm not seeing where it gets used?

jdee:
@ohayon The `:ignore` param (almost) predates my acquaintance with Fastlane. It's used [here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/commit_version_bump.rb#L56). I think it was added in this [PR](https://github.com/fastlane/fastlane/pull/5541).

ohayon (approved):

Issue #9490, opened 2017-06-15
Gym failing after upgrade to 2.38.1 "no implicit conversion of Symbol into Integer"


bes:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

After updating to fastlane-2.38.1 from 2.37.0, the GYM step of our build is failing

This patch might have something to do with it? https://github.com/fastlane/fastlane/commit/3c597616d1d5b170d2e2b2d54446079b32e1b8de (XCode 9 support)

As you can see match is run before gym, which the patch notes say should be enough.

Will roll back to 2.37.0 and await more info.

##### Complete output when running fastlane, including the stack trace and command used

```
~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/actions/gym.rb:26:in `[]=': [!] no implicit conversion of Symbol into Integer (TypeError)
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/actions/gym.rb:26:in `run'
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/runner.rb:252:in `block (2 levels) in execute_action'
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/actions/actions_helper.rb:50:in `execute_action'
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/runner.rb:230:in `block in execute_action'
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/runner.rb:226:in `chdir'
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/runner.rb:226:in `execute_action'
        from ~/.rvm/gems/ruby-2.3.4@my_app/gems/fastlane-2.38.1/fastlane/lib/fastlane/runner.rb:148:in `trigger_action_by_name'
[...]

PR #9493, opened 2017-06-15
Fix gym action failure if user passed a custom export_options plist file path

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/9490
Introduced by https://github.com/fastlane/fastlane/pull/9432


taquitos (approved):
🐑

PR #9524, opened 2017-06-19
Add last_git_commit_hash to git_helper

jvaladez:
### Checklist

### Motivation and Context

Adding the last git commit to the slack message provides useful information to those seeing the message. It identifies the exact commit by its identifier than just the commit message (ex: a1b2c3 vs "hot fix")

### Description

1. I added a last_git_commit_hash function to the git_helper. If given a boolean true it returns the short hash of the last commit, if not, it returns the full hash. 
2. Used the last_git_commit_hash to add the short hash to the default slack function message.

### DIFF
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -230,7 +230,7 @@ module Fastlane
         if Actions.last_git_commit_hash(true) && should_add_payload[:last_git_commit_hash]
           slack_attachment[:fields] << {
             title: 'Git Commit Hash',
-            value: Actions.last_git_commit_hash(true),
+            value: Actions.last_git_commit_hash(short: true),
             short: false
           }
         end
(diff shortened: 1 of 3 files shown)


jvaladez:
I signed it!

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -226,6 +226,15 @@ def self.generate_slack_attachments(options)
           }
         end
 
+        # last_git_commit_hash
+        if Actions.last_git_commit_hash(true) && should_add_payload[:last_git_commit_hash]
+          slack_attachment[:fields] << {
+            title: 'Git Commit Hash',
+            value: Actions.last_git_commit_hash(true),
```
getaaron: I think it would be more readable to call this with the named argument:

```ruby
Actions.last_git_commit_hash(short: true)
```

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -86,6 +86,17 @@ def self.last_git_commit_message
       nil
     end
 
+    # Get the hash of the last commit
+    def self.last_git_commit_hash(short)
```
getaaron: I think this is a bit more readable:

```ruby
format_specifier = short ? '%h' : '%H'
string = last_git_commit_formatted_with(format_specifier).to_s
return string unless string.empty?
return nil
```

getaaron (changes requested):

jvaladez:
@getaaron made the requested changes. Thanks!

mpirri (approved):
This looks great!  Thanks so much for your contribution! 🚀

PR #9557, opened 2017-06-22
Precheck copy correction

RishabhTayal:
Here is a minor PR to correct the copyright friendly name copy change for `precheck`.


mpirri (changes requested):
Thanks for the fix, @RishabhTayal . I suspect it could be squashed/merged as is, but there seem to be a lot of random commits in there in addition to the change for the string.  Can you rebase your branch from _fastlane_'s `master` to help clean up the commits?

Thanks!  🚀

RishabhTayal:
@mpirri not sure how to do it? Could you please elaborate?

mpirri (approved):

mpirri:
Actually, never mind!  It merged cleanly, no need to make any changes.

Thanks so much for your submission!  🚀

RishabhTayal:
Thanks, but could you explain what should I have done future reference?

mpirri:
So it looks like the `master` branch in your repo was not a clean copy of the `fastlane/master` branch. You can see that because of the various other commits, followed by the `Merge remote-tracking branch 'fastlane/master'` commits in the branch. The end result of your branch is only the copy change, made by the last of the 8 commits.

An easy way to clean this up would be to check out your feature branch and type:
`git rebase fastlane/master`
This would remove all the unnecessary commits and only keep the new commit.

You also might want to `checkout` your `master` branch and run the same command.

If for some reason that doesn't work, you can remove the `master` branch from your own repository via:
`git branch -d master && git fetch && git checkout master`

I hope this is helpful - let us know if you have any questions!  💥

RishabhTayal:
When I run `git rebase fastlane/master` I am getting merge conflicts.

mpirri:
You might have to `git rebase --skip` over the first few commits, until you get to commit `e729e80`.  Basically, you want `git rebase` to ignore those first 7 commits (since you don't really need them), so use `git rebase --skip` until you get to that last commit.

PR #9566, opened 2017-06-23
Pass app_identifier through to precheck

taquitos:
explicitly pass app_identifier through to recheck when passed as an arg to deliver


KrauseFx (approved):
Best fix 👍

Issue #9398, opened 2017-06-07
[regression] Crashlytics always succeeds while it shouldn't


Sega-Zero:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Crashlytics action succeed the upload, but cannot add testers due to inactive app id. In 2.36.0 version this was totally fine and crashlytics action were throwing an error, that could be handled by script. I was using this to launch the app in simulator via `scan`, so app could be activated on crashlytics:
```ruby
    tries = 3
    begin
      crashlytics(
		ipa_path: './App.ipa',
		groups: "test-group",
		notifications: false
	  )
    rescue Exception => e
      tries -= 1
      if tries > 0
        scan(
		  workspace: "App.xcworkspace",
		  scheme: "Release",
		  clean: false
		)
        retry
      else
        raise e
      end
    end
```
This is what it looked like on 2.36.0:
```shell
[17:32:25]: -------------------------
[17:32:25]: --- Step: crashlytics ---
[17:32:25]: -------------------------
[17:32:26]: Uploading the build to Crashlytics Beta. Time for some ☕️.
[17:33:19]: 2017-06-01 17:32:26.425 submit Crashlytics: Crashlytics.framework/submit 1.3.5 (18)
2017-06-01 17:33:19.144 submit Crashlytics: Unable to add testers: Project some-inactive-appid is inactive
2017-06-01 17:33:19.144 submit Crashlytics: Failed to add testers/groups to the release
[17:33:20]: Sending crash report...
[17:33:20]: The stacktrace is sanitized so no personal information is sent.
[17:33:20]: To see what we are sending, look here: /Users/travis/.fastlane/latest_crash.json
[17:33:20]: Learn more at https://github.com/fastlane/fastlane#crash-reporting
[17:33:20]: You can disable crash reporting by adding `opt_out_crash_reporting` at the top of your Fastfile
[17:33:20]: ------------------
[17:33:20]: --- Step: scan ---
[17:33:20]: ------------------
```

This is what it looks like now:
```shell
[09:48:59]: -------------------------
[09:48:59]: --- Step: crashlytics ---
[09:48:59]: -------------------------
[09:49:00]: Uploading the build to Crashlytics Beta. Time for some ☕️.
[09:50:51]: Shell command exited with exit status 1 instead of 0.
[...]

PR #9571, opened 2017-06-23
Make sure crashlytics action fails when submit binary returns failure

mpirri:
### Checklist

### Motivation and Context
A recent change to how `sh` and its `error_callback` worked means that a fatal error (i.e., `shell_error!`) is not thrown if an `error_callback` was provided.  Since the `crashlytics` action makes no attempts to rescue in the event of an error and only uses the callback to print out a sanitized version of the error message, we need to be sure the callback reports fatal error in the event of failure.


mpirri:
Fixes #9398

ohayon (approved):

PR #9600, opened 2017-06-28
Sort certificates to always verify the validity of the last

ypelud:
### Motivation and Context
In case I have a certificate expiring in less than 30 days and an expiring certificate in 300 days. Pem tries to create a certificate while the maximum number of certificates is reached and a certificate valid for more than 30 days.
It would be sufficient to verify that the certificate that expires last is still valid.

### Description
Sort certificates by descending expiration dates to check which expires last.


ypelud:
I signed it!

taquitos (approved):
Awesome

PR #9609, opened 2017-06-28
Add verification of Xcode version when running fastlane

KrauseFx:
- This PR will skip verification if no Xcode can be found at all
- This PR will show a nice error message if the user uses an old version of Xcode

screen shot 2017-06-28 at 3 42 17 pm


taquitos (approved):
🏆

fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
```diff
@@ -34,7 +34,24 @@ def run!
...
+        end
+      rescue
+        # We don't care about exceptions here
+        # We'll land here if the user doesn't have Xcode at all for example
+        # which is fine for someone who uses fastlane just for Android project
+        # What we *do* care about is when someone links an old version of Xcode
+      end
+
+      begin
+        if xcode_outdated
```
giginet: You should implement this out of the `begin` block.
I think `UI.user_error!` raises exception and it would be caught [here](https://github.com/fastlane/fastlane/pull/9609/files#diff-afee814ded0e79042832f3de1d681a53R109). Is this right?

giginet (changes requested):

fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
KrauseFx: > +          # We have to raise that error within this `begin` block to show a nice user error without a stack trace

Mh, yes, I think that's good though? This way we'll show a nice error message without stack trace.

giginet (approved):

fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
giginet: I got. There are no problems 😄

Issue #9616, opened 2017-06-29
Fails to upload OSX app to HockeyApp


satyavrat:
### New Issue Checklist

- [YES] Updated fastlane to the latest version
- [YES] I have read the [Contribution Guidelines]

### Issue Description
I am trying to build a OSX app and upload it to hockey for QA & beta distribution 
I was able to build the app and generate .dsym too. it fails at the stage of uploading to Hockeyapp. this is the error I get:

 **File exists @ sys_fail2 - /Users/satyavrat/macos/MyApp/MyApp.app/Contents/Frameworks/HockeySDK.framework/Resources**

##### Complete output when running fastlane, including the stack trace and command used

Attaching the whole --capture_output

INFO [2017-06-29 14:44:03.40]: ▸ Copying HockeySDK-Mac/HockeySDK.framework
INFO [2017-06-29 14:44:03.55]: ▸ Signing /Users/satyavrat/Library/Developer/Xcode/DerivedData/MyApp-ednhhgtcuyyhvobegbfadhfhvkcd/Build/Intermediates/ArchiveIntermediates/MyApp/InstallationBuildProductsLocation/Applications/MyApp.app/Contents/Frameworks/HockeySDK.framework
INFO [2017-06-29 14:44:03.69]: ▸ Touching MyApp.app
INFO [2017-06-29 14:44:03.78]: ▸ Signing /Users/satyavrat/Library/Developer/Xcode/DerivedData/MyApp-ednhhgtcuyyhvobegbfadhfhvkcd/Build/Intermediates/ArchiveIntermediates/MyApp/InstallationBuildProductsLocation/Applications/MyApp.app
INFO [2017-06-29 14:44:03.94]: ▸ Touching MyApp.app.dSYM
INFO [2017-06-29 14:44:04.15]: ▸ Archive Succeeded
DEBUG [2017-06-29 14:44:04.49]: Stored the archive in: /Users/satyavrat/Library/Developer/Xcode/Archives/2017-06-29/MyApp 2017-06-29 14.43.37.xcarchive
INFO [2017-06-29 14:44:04.50]: Compressing 1 dSYM(s)
INFO [2017-06-29 14:44:04.50]: $ cd '/Users/satyavrat/Library/Developer/Xcode/Archives/2017-06-29/MyApp 2017-06-29 14.43.37.xcarchive/dSYMs' && zip -r '/Users/satyavrat/moktale/macos/MyApp/MyApp.app.dSYM.zip' *.dSYM
[...]

Issue #9614, opened 2017-06-29
- hockey() command error on fastlane 2.42.0


DavidArayan:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

The hockey() command on the latest fastlane 2.42.0 produces the following error. Using the exact same input on 2.41.0 is actually fine. We use CircleCI with XCode 8.3, also tested using XCode 8.2 with no difference.

We also tried purging our CircleCI caches to make sure nothing weird was getting in the way.

hockey(
        api_token: 'our token',
        owner_id: 'our owner',
        notes: 'changelog',
        create_update: false,
        strategy: 'replace'
 )

It's also interesting that the build actually gets uploaded into hockey-app when this error occurs which leads me to believe there is some kind of error during the response from hockey-app.

For now, we have rolled back to 2.41.0 to work around this problem.

Specific error is: TypeError: [!] no implicit conversion of nil into String

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

[INSERT OUTPUT HERE]

TypeError: [!] no implicit conversion of nil into String
  /Users/distiller/app-ios/vendor/bundle/ruby/2.3.0/gems/fastlane-2.42.0/fastlane/lib/fastlane/actions/hockey.rb:174:in `+'
  /Users/distiller/app-ios/vendor/bundle/ruby/2.3.0/gems/fastlane-2.42.0/fastlane/lib/fastlane/actions/hockey.rb:174:in `run'
  /Users/distiller/app-ios/vendor/bundle/ruby/2.3.0/gems/fastlane-2.42.0/fastlane/lib/fastlane/runner.rb:252:in `block (2 levels) in execute_action'
  /Users/distiller/app-ios/vendor/bundle/ruby/2.3.0/gems/fastlane-2.42.0/fastlane/lib/fastlane/actions/actions_helper.rb:50:in `execute_action'
  /Users/distiller/app-ios/vendor/bundle/ruby/2.3.0/gems/fastlane-2.42.0/fastlane/lib/fastlane/runner.rb:230:in `block in execute_action'
  /Users/distiller/app-ios/vendor/bundle/ruby/2.3.0/gems/fastlane-2.42.0/fastlane/lib/fastlane/runner.rb:226:in `chdir'
[...]

PR #9619, opened 2017-06-29
Fix the generation of download url

wolffan:
### Checklist

### Motivation and Context
My apologies for failing many people's builds

### Description
This fixes the issue (generated by the previous merge) #9614
Generating crashes un upload https://github.com/fastlane/fastlane/issues/9616

### DIFF
--- a/fastlane/lib/fastlane/actions/hockey.rb
+++ b/fastlane/lib/fastlane/actions/hockey.rb
@@ -171,18 +171,8 @@ module Fastlane
         case response.status
         when 200...300
           url = response.body['public_url']
-          version_url = url
-
-          if response.body['version']
-            version_url = url + '/app_versions/' + response.body['version']
-          elsif response.body['config_url']
-            version_url = url + '/app_versions/' + response.body['config_url'].split('/').last
-          end
-
-          Actions.lane_context[SharedValues::HOCKEY_DOWNLOAD_LINK] = version_url
+          self.puts_download_url(response.body)
           Actions.lane_context[SharedValues::HOCKEY_BUILD_INFORMATION] = response.body
-
-          UI.message("Version Download URL: #{version_url}") if version_url
           UI.message("Public Download URL: #{url}") if url
           UI.success('Build successfully uploaded to HockeyApp!')
         else
@@ -192,6 +182,14 @@ module Fastlane
             UI.user_error!("Error when trying to upload file(s) to HockeyApp: #{response.status} - #{response.body}")
           end
         end
+      end
+
+      def self.puts_download_url(body)
+        version_url = body['public_url']
+        version_url = body['public_url'] + '/app_versions/' + body['config_url'].split('/').last if body['config_url']
+        version_url = body['public_url'] + '/app_versions/' + response.body['version'] if body['version']
+        Actions.lane_context[SharedValues::HOCKEY_DOWNLOAD_LINK] = version_url
+        UI.message("Version Download URL: #{version_url}") if version_url
       end
 
       def self.description
--- a/fastlane/lib/fastlane/actions/hockey.rb
+++ b/fastlane/lib/fastlane/actions/hockey.rb
@@ -186,7 +186,7 @@ module Fastlane
         end
       end
 
-      def self.puts_download_url(body)
+      def self.download_url(body)
         version_url = body['public_url']
         version_url = body['public_url'] + '/app_versions/' + body['config_url'].split('/').last if body['config_url']
         version_url = body['public_url'] + '/app_versions/' + response.body['version'] if body['version']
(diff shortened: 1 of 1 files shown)


wolffan:
@lexusilia this one

wolffan:
I get this
```
Bundler could not find compatible versions for gem "commander-fastlane":
  In Gemfile:
    fastlane was resolved to 2.42.0, which depends on
      commander-fastlane (< 5.0.0, >= 4.4.5)

Could not find gem 'commander-fastlane (< 5.0.0, >= 4.4.5)', which is required by gem 'fastlane', in any of the sources.
```

Any idea?

wolffan:
Ok so I get this:
````
fastlane/lib/fastlane/actions/hockey.rb:124:7: C: Metrics/PerceivedComplexity: Perceived complexity for run is too high. [21/18]
      def self.run(options)
````
Unrelated with the changes I've done... afaik

**Comment by wolffan** [comment]
Tested locally again, works...

**Review by jetersen — APPROVED** [review]
🐑 it, need it working 🚀

File: fastlane/lib/fastlane/actions/hockey.rb [hunk 1]
```diff
@@ -187,6 +184,14 @@ def self.run(options)
         end
       end
 
+      def self.puts_download_url(body)
+        version_url = body['public_url']
+        version_url = body['public_url'] + '/app_versions/' + body['config_url'].split('/').last if body['config_url']
```
Voxar: These lines has more potential nil errors
As this method only prints convenient information, perhaps result parsing problems could be captured and printed instead of raising fatal errors?

fastlane/lib/fastlane/actions/hockey.rb
```diff
@@ -187,6 +184,14 @@ def self.run(options)
...
+        version_url = body['public_url'] + '/app_versions/' + response.body['version'] if body['version']
+        Actions.lane_context[SharedValues::HOCKEY_DOWNLOAD_LINK] = version_url
```
Voxar: Too focused on nil checks to notice it also sets this shared value. As the method name starts with `puts_` I would not expect side effects

wolffan:
Related also to this https://github.com/fastlane/fastlane/issues/9614

fastlane/lib/fastlane/actions/hockey.rb
wolffan: I had to create a method to avoid complaints on `rubocop` and then compress maximum to avoid line class length :/
That's what it seems to do to much, the other option will be to return the version_url and asign the var outside.

fastlane/lib/fastlane/actions/hockey.rb
wolffan: Not super handy on ruby... but in this case the config_url is there always.
In case is empty should split break? (if no '/' are found?)
I asume last will work even if 1 element is present

Voxar: if `body` is a normal Hash returning nil if a key is missing, and nil behaves as normal, you will end up with
```
2.3.3 :001 > {}['hello'].split('/')
NoMethodError: undefined method `split' for nil:NilClass
```

wolffan:
@Voxar reworked a bit based on comments

fastlane/lib/fastlane/actions/hockey.rb
```diff
@@ -187,6 +186,13 @@ def self.run(options)
         end
       end
 
+      def self.puts_download_url(body)
```
Voxar: Not sure how picky one should be but now the method name is not doing what it says :) 

Good update otherwise 👍

wolffan: lol fair point 1 sec

fastlane/lib/fastlane/actions/hockey.rb
```diff
@@ -171,12 +171,11 @@ def self.run(options)
         case response.status
         when 200...300
           url = response.body['public_url']
-          version_url = url + "/app_versions/" + response.body['version']
-          Actions.lane_context[SharedValues::HOCKEY_DOWNLOAD_LINK] = version_url
+          app_url = self.puts_download_url(response.body)
```
Voxar: Oh man I can feel your stress, but yeah maybe also change this line to reflect the method name change.

wolffan: omg, yes in a rush :/.
Actually the change works on my project :).
But i'm very worried of other people's

Voxar (approved):

andrejunges:
do you guys intend to deploy this fix soon?

KrauseFx (approved):
Thanks for the hotfix

KrauseFx:
Whops, should have changed the title of the PR to include the name `hockey`

PR #9632, opened 2017-06-29
Detect number of tests for fastlane plugins and score accordingly

KrauseFx:
This is awesome 👍

### DIFF
--- a/fastlane/helper/plugin_scores_helper.rb
+++ b/fastlane/helper/plugin_scores_helper.rb
@@ -96,7 +96,10 @@ module Fastlane
                                    value: (self.data[:github_issues].to_i * -1)),
             FastlanePluginRating.new(key: :downloads,
                              description: "More downloads = more users have been using the plugin for a while",
-                                   value: (self.data[:downloads].to_i / 250))
+                                   value: (self.data[:downloads].to_i / 250)),
+            FastlanePluginRating.new(key: :tests,
+                             description: "The more tests a plugin has, the better",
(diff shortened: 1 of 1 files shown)


fastlane/helper/plugin_scores_helper.rb
```diff
@@ -96,7 +96,10 @@ def ratings
                                    value: (self.data[:github_issues].to_i * -1)),
             FastlanePluginRating.new(key: :downloads,
                              description: "More downloads = more users have been using the plugin for a while",
-                                   value: (self.data[:downloads].to_i / 250))
+                                   value: (self.data[:downloads].to_i / 250)),
+            FastlanePluginRating.new(key: :tests,
+                             description: "The more tests a plugin has, the better",
+                                   value: self.data[:tests].to_i * 3)
```
taquitos: I feel like this should be the ratio of test files to code files

taquitos: and have a maximum, like if if the ratio is 1 test file to 1 code file, that is a score of 20, if the ratio is 1 to 2, the score is 10, and so on. You never get better than a certain number maybe?

KrauseFx: Ugh, that sounds like work...

taquitos: Work is the worst.

taquitos (approved):
💇‍♂️

PR #9633, opened 2017-06-29
Add information in docs plugin page when plugin has no GitHub URL

KrauseFx:
- This way we show that the code might not be hosted somewhere
- This way we kind of explain why this score is lower (e.g. we couldn't verify the README or the # of tests)

screen shot 2017-06-29 at 2 15 44 pm


taquitos (approved):
great idea 🔢

PR #9643, opened 2017-06-30
Revert all changes applied to the hockey action recently

KrauseFx:
Reverts all recently applied patches of the hockey action


ohayon (approved):
👍

PR #9564, opened 2017-06-23
Rubocop fixes and update

jdee:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Fix #8958.

`bundle exec rubocop` fails on a newly-created plugin. There are two issues:

- There were some rubocop failures because it wanted the gemspec file to be snake_case,
  and it objects to unnecessary use of `%q{}` in the template gemspec. Those have been fixed. The snake_case check was disabled for gemspecs, and the `%q{}` literals in the template gemspec have been changed to simple quotes.
- The plugin template uses the .rubocop.yml from the root of this repo. The `fastlane.gemspec`
  pins rubocop to 0.48.1, which is not current. The plugin template gemspec places no limitation
  on the version of rubocop. Some Cops have changed namespaces in 0.49, generating warnings 
  (not offenses) in a new plugin using the latest version. Because Fastlane proper uses a different version with the same .rubocop.yml, rubocop runs cleanly in the repo.

### Description

There are several commits that implement different stages. Some could be reverted if not desired.

1. The simplest fix is to limit rubocop to 0.48.1 in the plugin template to avoid any changes in the main codebase for this issue. The issue with the gemspec filename and the `%q{}` literals still had to be fixed. `git checkout b8b02e1` for everything working with 0.48.1 (except perhaps a failing unit test).
2. Update Fastlane proper to 0.49.1. After fixing the snake_case and `%q{}` issues, this required:
  - Updating fastlane.gemspec.
  - Changing the Style namespace to Layout for a number of Cops in the .rubocop.yml and code comments.
  - ~~Running `rubocop -a` and making manual fixes for certain complaints.~~ Adding comments at various places in the code to eliminate certain failures temporarily until the authors can review them.
  The HEAD of this branch has everything updated to 0.49.1 and quietly passing both in the repo and new plugins. But this entailed a handful of code changes throughout the repo.
[...]

PR #9650, opened 2017-07-02
Fix broken unit tests for plugin generation

jdee:
### Checklist

### Motivation and Context

See the discussion from #9564.

Unit tests for rake, rspec  and rubocop in new plugins are all executed by passing a block to `Bundler.setup`, but they don't appear to be executed at all, and actual failures are not reported. I wasn't familiar with `Bundler.setup`, and I can't find docs for that method, but judging from the source code:

https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L92
https://github.com/bundler/bundler/blob/master/lib/bundler/runtime.rb#L13

It doesn't look like the method takes a block. I think it wants an array of symbols for groups from the Gemfile, e.g. `Bundler.setup [:dev, :test]`. I suspect that it is called long before execution of these examples, and these calls just short circuit and return `@setup` without doing anything else.

### Description

It wasn't hard to use `RuboCop::CLI` to run RuboCop programmatically from the example. Rake was hard enough that I took another crack at shelling out to `bundle exec`, which turns out to work well.

These tests will now report failures when appropriate. For example, remove the blank line after `# coding: utf-8` in fastlane/lib/fastlane/plugins/template/%gem_name%.gemspec.erb and run

```bash
bundle exec rspec fastlane/spec/plugins_specs/plugin_generator_spec.rb
```

This will show a RuboCop failure.

### DIFF
--- a/fastlane/spec/plugins_specs/plugin_generator_spec.rb
+++ b/fastlane/spec/plugins_specs/plugin_generator_spec.rb
@@ -273,36 +273,37 @@ describe Fastlane::PluginGenerator do
     end
 
     describe "All tests and style validation of the new plugin are passing" do
+      before (:all) do
+        # let(:gem_name) is not available in before(:all), so pass the directory
+        # in explicitly once instead of making this a before(:each)
+        plugin_sh 'bundle install', 'fastlane-plugin-tester_thing'
+      end
+
       it "rspec tests are passing" do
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
This looks great, thank you so much for your contribution 👍 ❤️

PR #9656, opened 2017-07-03
Fix using the wrong team ID/Name in precheck from Appfile

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/9655

Thanks @setoff


ohayon (approved):
👍

PR #9659, opened 2017-07-03
Plugin scores: follow redirects

KrauseFx:
This happens a lot when GitHub repos are being moved over to a new account


mpirri (approved):
🚢

PR #9680, opened 2017-07-06
Catch OpenSSL SSL error as well as Faraday to show instructions

KrauseFx:
Related: https://github.com/fastlane/fastlane/issues/9669 didn't show the instructions, so all code that doesn't use Faraday didn't show those instructions until now


mpirri (approved):
💯

Issue #9661, opened 2017-07-03
GRADLE_ALL_APK_OUTPUT_PATHS is empty when use Android Gradle flavours with dimensions


mitrofany4:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

GRADLE_ALL_APK_OUTPUT_PATHS is empty when use Android Gradle flavours with dimensions. It happens because fastlane only search at the ./app/build/outputs/apk folder.
```
flavorDimensions "icon", "api"
    productFlavors {
        dev {
            dimension "api"
            applicationIdSuffix '.testing'
        }

        uat {
            dimension "api"
            applicationIdSuffix '.uat'
        }

        prod {
            dimension "api"
        }

        black {
            dimension "icon"
        }

        white {
            dimension "icon"
        }
    }
```
It happens because fastlane only search at the ./app/build/outputs/apk folder.
But should search at subfolders too. For example,
./app/build/outputs/apk/whiteDev/release
./app/build/outputs/apk/whiteProd/release
etc.

-----------------------------+-----------------+
|                 Lane Context                  |
+-----------------------------+-----------------+
| DEFAULT_PLATFORM            | android         |
| PLATFORM_NAME               | android         |
| LANE_NAME                   | android betaUAT |
| GRADLE_ALL_APK_OUTPUT_PATHS | []              |
+-----------------------------+-----------------+
[18:20:57]: You have to either pass an ipa or an apk file to the Crashlytics action

+------+---------------------+-------------+
|             fastlane summary             |
+------+---------------------+-------------+
| Step | Action              | Time (in s) |
+------+---------------------+-------------+
| 1    | update_fastlane     | 10          |
| 2    | Verifying required  | 0           |
|      | fastlane version    |             |
| 3    | default_platform    | 0           |
| 4    | slack               | 0           |
| 5    | git_pull            | 1           |
| 6    | gradle              | 37          |
| 7    | is_ci               | 0           |
| 💥   | crashlytics         | 0           |
| 9    | slack               | 0           |
+------+---------------------+-------------+

[18:20:57]: fastlane finished with errors

[!] You have to either pass an ipa or an apk file to the Crashlytics action

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

[...]

PR #9688, opened 2017-07-06
Recursively search through `apk` directory in gradle action

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/9661


mpirri (approved):
👍

Issue #9684, opened 2017-07-06
Gym failing with "no implicit conversion of Symbol into Integer" in 2.45.0 when using sigh for profile selection


sbusch-mobivention:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Using the newest version i always get the same error in gym. I tried using it in different projects and the result is always the same

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

### Captured Output

Command Used: `hockeymobi`
<details><summary>Output/Log</summary>

```

DEBUG [2017-07-06 13:10:23.17]: Checking if there are any plugins that should be loaded...
INFO [2017-07-06 13:10:23.32]: -------------------------------------------------
INFO [2017-07-06 13:10:23.32]: --- Step: Verifying required fastlane version ---
INFO [2017-07-06 13:10:23.32]: -------------------------------------------------
INFO [2017-07-06 13:10:23.32]: Your fastlane version 2.45.0 matches the minimum requirement of 1.99.0  ✅
INFO [2017-07-06 13:10:23.32]: ------------------------------
INFO [2017-07-06 13:10:23.32]: --- Step: default_platform ---
INFO [2017-07-06 13:10:23.32]: ------------------------------
ERROR [2017-07-06 13:10:23.32]: Name of the lane 'hockey' is already taken by the action named 'hockey'
ERROR [2017-07-06 13:10:23.32]: Name of the lane 'appstore' is already taken by the action named 'appstore'
INFO [2017-07-06 13:10:23.32]: Driving the lane 'ios hockeymobi' 🚀
INFO [2017-07-06 13:10:23.33]: -----------------------
INFO [2017-07-06 13:10:23.33]: --- Step: cocoapods ---
INFO [2017-07-06 13:10:23.33]: -----------------------
INFO [2017-07-06 13:10:23.59]: gem 'cocoapods' is already installed
INFO [2017-07-06 13:10:23.59]: $ pod install
INFO [2017-07-06 13:10:29.21]: ▸ Analyzing dependencies
INFO [2017-07-06 13:10:29.21]: ▸ Downloading dependencies
INFO [2017-07-06 13:10:29.21]: ▸ Using ActionSheetPicker-3.0 (2.2.0)
[...]

PR #9701, opened 2017-07-07
Fix gym action and implementation when user provide path instead of hash

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/9684
The user can provide a path to a plist instead of a hash


mpirri (approved):
:shipit:

Issue #9704, opened 2017-07-07
gym fails to export IPA scheme including WatchKit App


iDaniel89:
### New Issue Checklist

- [v] Updated fastlane to the latest version
- [v] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I am running gym from Fastfile in terminal and I keep getting this error "no implicit conversion of nil into String". Works fine in Xcode.

/Users/Administrador/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.45.0/fastlane/lib/fastlane/actions/gym.rb:44:in `expand_path': [!] **no implicit conversion of nil into String** (TypeError)

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

COMMAND
------------------------------------------------------------------
fastlane crashlyticsbeta --verbose

OUTPUT
------------------------------------------------------------------
Fastlane update available
Please run `brew cask reinstall fastlane`
DEBUG [2017-07-07 11:08:30.43]: Checking if there are any plugins that should be loaded...
INFO [2017-07-07 11:08:30.62]: ------------------------------
INFO [2017-07-07 11:08:30.62]: --- Step: default_platform ---
INFO [2017-07-07 11:08:30.62]: ------------------------------
INFO [2017-07-07 11:08:30.62]: Driving the lane 'ios crashlyticsbeta' 🚀
INFO [2017-07-07 11:08:30.63]: -----------------
INFO [2017-07-07 11:08:30.63]: --- Step: gym ---
INFO [2017-07-07 11:08:30.63]: -----------------
DEBUG [2017-07-07 11:08:30.81]: Parsing project file '/Users/Administrador/.jenkins/workspace/Mobile/app/appIOSQA/appNameApp.xcodeproj' to find selected provisioning profiles
INFO [2017-07-07 11:08:31.43]: Detected provisioning profile mapping: {"com.company.app"=>"company app Distribution", "com.company.app.watchkit"=>"company app Watchkit Distribution", "com.company.app.watchkit.extension"=>"company app Watchkit Extension Distribution"}
INFO [2017-07-07 11:08:31.43]: $ xcodebuild -list -workspace ./appNameApp.xcworkspace -configuration Release
[...]

Issue #9260, opened 2017-05-24
Gym fails IPA export in schema including WatchKit App


fukazayo:
### New Issue Checklist

- [v] Updated fastlane to the latest version
- [v] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I created a simple iOS App with WatchKit App project and try to build using gym.
But gym fails IPA export in a schema including WatcKit App.
(Gym succeeds IPA export in a schema without WatchKit App.)

### Captured Output

Command Used: `debug`
<details><summary>Output/Log</summary>

```

DEBUG [2017-05-24 15:27:58.52]: Checking if there are any plugins that should be loaded...
INFO [2017-05-24 15:27:58.52]: -------------------------------------------------
INFO [2017-05-24 15:27:58.52]: --- Step: Verifying required fastlane version ---
INFO [2017-05-24 15:27:58.52]: -------------------------------------------------
INFO [2017-05-24 15:27:58.52]: Your fastlane version 2.33.0 matches the minimum requirement of 2.33.0  ✅
INFO [2017-05-24 15:27:58.52]: ------------------------------
INFO [2017-05-24 15:27:58.52]: --- Step: default_platform ---
INFO [2017-05-24 15:27:58.52]: ------------------------------
ERROR [2017-05-24 15:27:58.52]: Name of the lane 'debug' is already taken by the action named 'debug'
INFO [2017-05-24 15:27:58.52]: Driving the lane 'ios debug' 🚀
INFO [2017-05-24 15:27:58.53]: -----------------
INFO [2017-05-24 15:27:58.53]: --- Step: gym ---
INFO [2017-05-24 15:27:58.53]: -----------------
INFO [2017-05-24 15:27:58.75]: $ xcodebuild -list -project ./WatchKitTest.xcodeproj -configuration Debug
INFO [2017-05-24 15:27:59.41]: $ xcodebuild -showBuildSettings -scheme WatchKitTest\ WatchKit\ App -project ./WatchKitTest.xcodeproj -configuration Debug

+----------------------+--------------------------------------+
|                   Summary for gym 2.33.0                    |
+----------------------+--------------------------------------+
| configuration        | Debug                                |
| export_method        | development                          |
| scheme               | WatchKitTest WatchKit App            |
| project              | ./WatchKitTest.xcodeproj             |
| destination          | generic/platform=iOS                 |
[...]

PR #9706, opened 2017-07-07
Update gym action to work if no ipa path is returned

KrauseFx:
Related to https://github.com/fastlane/fastlane/issues/9704 and https://github.com/fastlane/fastlane/issues/9260


mpirri (approved):
:+1:

PR #9726, opened 2017-07-12
Rename error_handler to error_handler_spec to run tests again

KrauseFx:
Also fix tests that broke after recent migration


KrauseFx:
Merged with the other PR

PR #9736, opened 2017-07-13
Check superseded tracks for rollout in supply

JacobMuchow:
### Checklist

### Motivation and Context

In the current version of fastlane, it isn't possible to supersede tracks on Google Play when deploying to the rollout track. I don't know if this was intentional, but it doesn't appear to be as it worked for me in testing. For my test, I had multiple APKs in the Alpha channel on Google play and did a Rollout to 25% using fastlane with new APKs. Before this change, I would get the multishadow APK error from the Google Developer API. With this change, it worked successfully like it does for the other tracks.

### Description

* Moved the check for superseded tracks to before if/else chain for update_track. Previously it would only check for production, beta and alpha.

* Added "rollout" to the tracks list for checking supersession.

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -155,10 +155,11 @@ module Supply
 
     def update_track(apk_version_codes)
       UI.message("Updating track '#{Supply.config[:track]}'...")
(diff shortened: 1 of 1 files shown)


JacobMuchow:
I signed it!

KrauseFx (approved):
Nice, thanks for the PR 👍

Issue #9690, opened 2017-07-06
May need to change max supply rollout percentage


muchow-hello:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Since the latest update to the Google Play Developer Console, it has been easier to specify rollout percentages. While it states [in the documentation](https://developers.google.com/android-publisher/tracks) that the range is from 5-50%, I think this may be an out of date claim because you can actually set the rollout percentage on the Console to something like 75% for example.

As a further example, I got through this flow with one of my clients:

* Original Deploy: 25% rollout

* 1 Hour later: 50% rollout

* 2 Hours later: 75% rollout

* 3 Hours later: 100% rollout (so production track)

I am working on setting up some Fastlanes to handle this flow for me so I don't have to worry about it anymore, but I am limited to 50% rollout by [supply/options.rb](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb). I will probably just ignore the 75% step for now and go straight from 50% to 100%, but I thought I would let you know this may be something to look into.

I have submitted some feedback on the Google Developer website about the potentially out of date documentation, so I hope they will fix it. It might be worth testing a higher max percentage to see how the API responds. 

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

[12:25:24]: fastlane finished with errors

[!] Invalid value '0.75', must be between 0.01 and 0.5

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

Don't think this is important to the issue, but I can edit it in if desired. There is some NDA stuff I will have to edit out.

KrauseFx:
Nice catch, thanks for letting us know. Are you interested in submitting a PR to change this? :)

muchow-hello:
@KrauseFx If I find some time this week, I will try to test the change and make a pr.

KrauseFx:
Awesome, thanks again for the PR 👍

muchow-hello:
You're welcome! Happy to help

PR #9737, opened 2017-07-13
Updated supply rollout number verification: range is 0.0 - 1.0.

JacobMuchow:
### Checklist

### Motivation and Context

Related issue: https://github.com/fastlane/fastlane/issues/9690

The range when this feature was added to fastlane was 0.05 - 0.5. I believe this was updated on the Google Play Developer Console in their recent update an I'm able to set the rollout percentage to anywhere from 0.0 - 1.0 on the console.

After testing, I was able to determine that the Google API considers numbers <= 0.0 invalid and numbers >= 1.0 invalid. Everything between that works. I tested this by updating a production app in the rollout track. This is the build step I'm using:

```
desc "bump production build to x% rollout in google play"
lane :updateRollout do |options|
# update rollout percentage on Google Play
supply(
  track: 'rollout',
  track_promote_to: 'rollout',
  rollout: options[:pct],
  metadata_path: 'fastlane/prod',
  skip_upload_apk: true,
  skip_upload_metadata: true,
  skip_upload_images: true,
  skip_upload_screenshots: true,
  check_superseded_tracks: true
)
end
```

I confirmed that numbers such as 0.75 and 0.90 work, while 1.0 does not work. 

### Description

I changed the min/max numbers to 0.0 and 1.0. The `between` check is inclusive, so I changed to to `value > min && value < max`.

### DIFF
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -26,9 +26,9 @@ module Supply
                                      description: "The percentage of the user fraction when uploading to the rollout track",
                                      default_value: '0.1',
                                      verify_block: proc do |value|
-                                       min = 0.01
-                                       max = 0.5
-                                       UI.user_error! "Invalid value '#{value}', must be between #{min} and #{max}" unless value.to_f.between?(min, max)
+                                       min = 0.0
+                                       max = 1.0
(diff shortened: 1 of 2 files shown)


KrauseFx (approved):
Awesome, thanks for your contribution @JacobMuchow 👍

Issue #9748, opened 2017-07-13
 bundle update fastlane is not updating to latest build


jlubeck:
### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I got a message from fastlane saying
```
#######################################################################
# fastlane 2.46.1 is available. You are on 2.37.0.
# You should use the latest version.
# Please update using `bundle update fastlane`.
#######################################################################
```
So I ran that command. It did the following

```
Fetching gem metadata from https://rubygems.org/...........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies.....
Using CFPropertyList 2.3.5
Using i18n 0.8.4
Using minitest 5.10.2
Using thread_safe 0.3.6
Using public_suffix 2.0.5
Using aws-sigv4 1.0.0
Using jmespath 1.3.1
Using babosa 1.0.2
Using bundler 1.14.2
Using claide 1.0.2
Using fuzzy_match 2.0.4
Using nap 1.1.0
Using cocoapods-deintegrate 1.0.1
Using cocoapods-downloader 1.1.3
Using cocoapods-search 1.0.0
Using cocoapods-stats 1.0.0
Using netrc 0.7.8
Using cocoapods-try 1.1.0
Using colored 1.2
Using escape 0.0.4
Using fourflusher 2.0.1
Using gh_inspector 1.0.3
Using molinillo 0.5.7
Using ruby-macho 0.2.6
Using colored2 3.1.2
Using nanaimo 0.2.3
Using highline 1.7.8
Using unf_ext 0.0.7.4
Using dotenv 2.2.1
Installing excon 0.57.1 (was 0.56.0)
Using multipart-post 2.0.0
Using fastimage 2.1.0
Using jwt 1.5.6
Using little-plugger 1.1.4
Using multi_json 1.12.1
Installing memoist 0.16.0 (was 0.15.0)
Using os 0.9.6
Using httpclient 2.8.3
Using hurley 0.2
Using mime-types 2.99.3
Using uber 0.0.15
Using retriable 2.1.0
Using json 1.8.6
Using mini_magick 4.5.1
Using multi_xml 0.6.0
Using plist 3.3.0
Using rubyzip 1.2.1
Using security 0.1.3
Using slack-notifier 1.5.1
[...]

PR #9756, opened 2017-07-14
Add warning if fastlane is executed without `bundle exec`

KrauseFx:
This PR also updates the links to our new docs

Related https://github.com/fastlane/fastlane/issues/9748
Related https://twitter.com/orta/status/885543142405140480

### DIFF
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -114,7 +114,7 @@ module Fastlane
         end
         UI.important "For more information, check out https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile"
 
-        sleep 1
+        sleep 2 # napping is life, otherwise the user might not see this message
       end
 
       # Returns an array of symbols for the available lanes for the Fastfile
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -18,8 +18,8 @@ module Fastlane
         require "fastlane" # this might take a long time if there is no Gemfile :(
 
         # We want to avoid printing output other than the version number if we are running `fastlane -v`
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/cli_tools_distributor.rb
```diff
@@ -18,8 +18,8 @@ def take_off
         require "fastlane" # this might take a long time if there is no Gemfile :(
 
         # We want to avoid printing output other than the version number if we are running `fastlane -v`
-        if Time.now - before_import_time > 3 && !running_version_command?
-          print_slow_fastlane_warning
+        unless running_version_command?
```
ohayon: I'm not sure I understand this change, if a user has _fastlane_ installed in their gemset, and in their bundle with different versions, not using `bundle exec` will print a different version than will be executed when running with `bundle exec` right?

ohayon (changes requested):
Is there something we could do in a Fastfile to say that `bundle exec` needs to be used when executing? Otherwise, if i understand this PR correctly, this change will get in the way of people using _packaged-fastlane_, or people that just don't want to have a Gemfile, right?

fastlane/lib/fastlane/cli_tools_distributor.rb
KrauseFx: Sorry, I don't understand the question.

KrauseFx:
> Is there something we could do in a Fastfile to say that bundle exec needs to be used when executing? Otherwise, if i understand this PR correctly, this change will get in the way of people using packaged-fastlane, or people that just don't want to have a Gemfile, right?

We've shown the warning for a while when fastlane is slow starting up, except when it's a bundled fastlane (homebrew, mac app). We encourage people to use a `Gemfile`. This PR will change the behavior that it will warn people that don't use `bundle exec`, even though they have a `Gemfile`, to use `bundle exec`.

ohayon:
Ah! Got it!

ohayon (approved):

PR #9460, opened 2017-06-13
Xcode 9: Automatically detect selected provisioning profiles of the Xcode project

KrauseFx:
This way fastlane will just keep working with Xcode 9 even for setups that don't use _match_ (yet).

Xcode 9 requires a mapping of bundle identifier and what provisioning profile to use. We use `xcodeproj` to find those values, but in a way that _gym_ doesn't break if we can't detect those things

I also verified that things work fine if you pass non-existent bundle identifier/profiles, Xcode will just ignore them 👍

Fixes https://github.com/fastlane/fastlane/issues/9380 even more

Output when running it:
screen shot 2017-06-13 at 2 46 41 pm

Error message if it's not auto-detected, catching Xcode error, and linking to the new docs :) 
screen shot 2017-06-13 at 2 41 55 pm

PR #9760, opened 2017-07-14
Resolve PRODUCT_BUNDLE_IDENTIFIER and PROVISIONING_PROFILE_SPECIFIER for Xcode 9 export settings

UnsafePointer:
### Checklist

### Motivation and Context
I ran into the issue some users are reporting in https://github.com/fastlane/fastlane/pull/9460#issuecomment-312931119. I think while implementing #9380 in #9460, projects that use xcconfig files to override `PRODUCT_BUNDLE_IDENTIFIER` and `PROVISIONING_PROFILE_SPECIFIER` build settings were not considered.

### Description

If a user project uses xcconfig files to set up these build settings, `XCBuildConfiguration#resolve_build_setting` from `xcodeproj` properly resolves the value in case of variable substitution.

This PR depends on https://github.com/CocoaPods/Xcodeproj/pull/501

### Sample project

Get it from: https://github.com/Ruenzuo/fastlane_xcconfig_xcode_9

xcconfig file using variable substitution looks like:

```
PRODUCT_BUNDLE_IDENTIFIER_Release = com.cocoapods.app
PRODUCT_BUNDLE_IDENTIFIER_Debug = com.cocoapods.app.dev
PRODUCT_BUNDLE_IDENTIFIER = $(PRODUCT_BUNDLE_IDENTIFIER_$(CONFIGURATION))

PROVISIONING_PROFILE_SPECIFIER_Release = Some Provisioning Profile
PROVISIONING_PROFILE_SPECIFIER_Debug = Some Other Provisioning Profile
PROVISIONING_PROFILE_SPECIFIER = $(PROVISIONING_PROFILE_SPECIFIER_$(CONFIGURATION))
```

Run on master (using latest fastlane), logs show:

```
Detected provisioning profile mapping: {"$(PRODUCT_BUNDLE_IDENTIFIER_$(CONFIGURATION))"=>"$(PROVISIONING_PROFILE_SPECIFIER_$(CONFIGURATION))"}
```

Run on fix_xcconfig_files (using this and https://github.com/CocoaPods/Xcodeproj/pull/501), logs show:

```
Detected provisioning profile mapping: {"com.cocoapods.app.dev"=>"Some Other Provisioning Profile", "com.cocoapods.app"=>"Some Provisioning Profile"}
```

### DIFF
--- a/gym/lib/gym/detect_values.rb
+++ b/gym/lib/gym/detect_values.rb
@@ -94,10 +94,8 @@ module Gym
           project = Xcodeproj::Project.open(project_path)
           project.targets.each do |target|
             target.build_configuration_list.build_configurations.each do |build_configuration|
-              current = build_configuration.build_settings
-
-              bundle_identifier = current["PRODUCT_BUNDLE_IDENTIFIER"]
-              provisioning_profile_specifier = current["PROVISIONING_PROFILE_SPECIFIER"]
+              bundle_identifier = build_configuration.resolve_build_setting("PRODUCT_BUNDLE_IDENTIFIER")
+              provisioning_profile_specifier = build_configuration.resolve_build_setting("PROVISIONING_PROFILE_SPECIFIER")
               next if provisioning_profile_specifier.to_s.length == 0
 
               provisioning_profile_mapping[bundle_identifier] = provisioning_profile_specifier
--- a/gym/lib/gym/detect_values.rb
+++ b/gym/lib/gym/detect_values.rb
@@ -94,8 +94,8 @@ module Gym
           project = Xcodeproj::Project.open(project_path)
           project.targets.each do |target|
             target.build_configuration_list.build_configurations.each do |build_configuration|
-              bundle_identifier = build_configuration.resolve_build_setting("PRODUCT_BUNDLE_IDENTIFIER")
-              provisioning_profile_specifier = build_configuration.resolve_build_setting("PROVISIONING_PROFILE_SPECIFIER")
+              bundle_identifier = build_configuration.resolve_build_setting("PRODUCT_BUNDLE_IDENTIFIER", target)
+              provisioning_profile_specifier = build_configuration.resolve_build_setting("PROVISIONING_PROFILE_SPECIFIER", target)
               next if provisioning_profile_specifier.to_s.length == 0
 
               provisioning_profile_mapping[bundle_identifier] = provisioning_profile_specifier


Ruenzuo:
I signed it!

Ruenzuo:
@mjholgate @igor-makarov could you please verify if the proposed change fix the problems you mentioned on https://github.com/fastlane/fastlane/pull/9460#issuecomment-312931119?

igor-makarov:
Thanks for the fix. I'm sorry I can't test it right away. Will look at it on Sunday.

mpirri:
If a user does not have the latest version of `xcodeproj` (i.e., the one that will contain the required change to make this PR work as expected), will this new code work the same as it does now? I.e., does the currently available version of `xcodebuild` return the same value for `resolve_build_setting("foo")` as it does for `build_settings["foo"]`?

If it does not, we will need to either conditionalize this code to work in both cases, or else force everyone to the newest version of `xcodebuild` in the `fastlane.gemspec`.

Also, any way you can think to add testing around this code?

Thanks for the submission! 🚀

Ruenzuo:
No, for this to work we need https://github.com/CocoaPods/Xcodeproj/pull/501 merged and a release of `xcodeproj`, so I think a gemspec bump would be appropriate.

I couldn't find test in the original implementation, so I assumed maybe there's an integration suite that contains the test for this? I'd love to add a case there if someone points me where is it. Maybe @KrauseFx knows.

KrauseFx:
Mh, I didn't add tests to it back then. If you tested this for your local setup, I can do some testing on my local machine for my projects, and we can merge. If you have ideas on how to best test it, feel free to add this to this PR 👍

igor-makarov:
Tried my example with the proposed change. It fails with 
```
undefined local variable or method `key' for #<Xcodeproj::Project::Object::XCBuildConfiguration:0x007f9990107cb0>
```
[Travis Build](https://travis-ci.org/igor-makarov/DemoWorkspaceWithXCConfig/builds/254337845)

Ruenzuo:
@igor-makarov thanks for checking this! I'll have a look today.

Ruenzuo:
@igor-makarov I identified the problem: I didn't consider that variable substitution used in an xcconfig file could reference build settings applied in a different level. It should work now if you update `xcodeproj` from my fork.

@KrauseFx I'll work on adding a more tests on the `xcodeproj` side, there are many cases for this 😭  so I want to get it as right as I can. I'll see if I can find the right place to add some sort of "integration" test to fastlane to check that this works, but I believe it makes more sense to add the tests on `xcodeproj` side.

igor-makarov:
@Ruenzuo Let me know if you need anything from me. I'm in GMT+3 FYI.

Ruenzuo:
@igor-makarov did the new version of my fork worked for you?

igor-makarov:
I'll check now.

igor-makarov:
The profile got set correctly this time! 👍

mjholgate:
@Ruenzuo apologies for taking so long to get back to you. I tried it yesterday, and I'm seeing one entry in the provisioningProfile dictionary and one unexpanded one. I'll try and dig into it more today and get back to you with what I find.

Ruenzuo:
@mjholgate a sample project where I can reproduce and debug the issue would be really appreciated 😊

mjholgate:
Hi @Ruenzuo please see the attached sample project. We have quite a complicated setup due to the fact that our app can be configured in multiple dimensions:

 - brand (it's a whitelabel app that can be rebranded)
 - production and development backends
 - enterprise / app store signed
 - various configurations for UI testing etc.

I did a little bit of digging and made a couple of observations, but I didn't understand it enough to fix it :)

1) I think the regex ` /\$[\(|{]([^inherited].*)[\)|}]/"` in `resolve_variable_substitution` is wrong:
 a) because `[^inherited]` matches any variable beginning with either i,n,h,e etc. rather than matching the string "inherited"
 b) the regex captures too much if there are multiple variable substitutions in a line, e.g. $(FOO).$(BAR). Making the regex non greedy with `?` addressed, although I suspect this might break the nested case which I *think* this code currently supports e.g. $($(BAR))

2) Our xcconfig files weren't originally working because we had no spaces around the assignment operator, i.e. `VAR="value"` rather than `VAR = "value"`. In the end I just changed our xcconfigs to be correct, so I don't think is a problem.

3) As `resolve_build_setting` recurses, it seems that `base_configuration_reference`/`config` become null, so variables defined in the configuration file aren't resolved.

Many thanks for all your help on this, it's really appreciated. Let me know if we can assist in any way.





[ExpandingVariables.zip](https://github.com/fastlane/fastlane/files/1155330/ExpandingVariables.zip)

Ruenzuo:
@mjholgate thank you for the awesome feedback ❤️ 

I've updated my fork with some updates. Some comments on your points:

1.a. I'm not a regex expert, my intention there is just to exclude `inherited` from the capture block so would love if you could point me how do this better.
1.b. You're right. Variable substitution was not working at all for appending build settings because of this reason. I modified the code in a way that the regex is lazy and nested values are still supported, also added tests for this.

2. Agree.

3. Yes, this is the exit condition of the recursion. `base_configuration_reference` means the attached xcconfig file and config is just the parsed values. If those are null, it means either the target or the project don't have a xcconfig file attached.

Thanks to your feedback I noticed another problem with my approach. The recursion starts from the target build settings and goes up to the project. If the recursion reaches to a point where the lookup happens at the project level, it is not possible to go back to the target level (at least I don't know how, will ask the CocoaPods team for help). In practice, this means that this method will fail if someone attempts to look for a build setting declared at a project level that resolves values at target level, which is the case of the sample project you sent me. If you move the value of `PRODUCT_BUNDLE_IDENTIFIER` to the target level this works.

Ruenzuo:
@mjholgate I think I got an ugly but working solution this time. Could you check the latest master on my forks?

KrauseFx:
I have no strong opinion about this, with those kinds of changes, very often it will break some setups. As soon as multiple people can confirm that this works, we can merge it 👍

Ruenzuo:
@KrauseFx thanks for the feedback. I'm waiting for @mjholgate response on the last change, then I'll try to get the Xcodeproj PR merged. We would need a released version first and a gemspec update before merging this PR.

mjholgate:
@Ruenzuo sorry for the delay, I'll test this today!

mjholgate:
@Ruenzuo that works perfectly, nice one, thanks so much!

KrauseFx (approved):
Alright, thanks everyone, let's give this a try 👍

Issue #9761, opened 2017-07-14
Precheck reports other_platforms violation for partial matches


FelixII:
### New Issue Checklist

- [x ] Updated fastlane to the latest version
- [x ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I'm using the German word "Notizen" (notes) in my app description and get a platform violation for "tizen". It would be great if precheck could avoid this match or at least let me configure a whitelist.

This issue relates to #9567, not sure if it is exactly the same problem.

##### Complete output when running fastlane, including the stack trace and command used

    $ fastlane precheck
    [17:16:56]: Successfully loaded '/Users/xxx/Developer/xxx/fastlane/Precheckfile' 📄
    [17:16:56]: No values defined in './fastlane/Precheckfile'
    
    +--------------------+----------------------------+
    |           Summary for precheck 2.46.1           |
    +--------------------+----------------------------+
    | app_identifier     | xxx                        |
    | username           | xxx                        |
    | default_rule_level | error                      |
    +--------------------+----------------------------+
    
    [17:16:56]: Starting login with user 'xxx'
    [17:16:59]: Successfully logged in
    [17:16:59]: Checking app for precheck rule violations
    [17:17:07]: ✅  Passed: No negative  sentiment
    [17:17:07]: ✅  Passed: No placeholder text
    [17:17:07]: 😵  Failed: No mentioning  competitors-> mentioning other platforms, like Android or Blackberry
    [17:17:07]: ✅  Passed: No future functionality promises
    [17:17:07]: ✅  Passed: No words indicating test content
    [17:17:07]: ✅  Passed: No curse words
    [17:17:07]: ✅  Passed: Incorrect, or missing copyright date
    [17:17:10]: ✅  Passed: No broken urls
    +----------------------+----------------+
    |          Potential problems           |
    +----------------------+----------------+
    | Field                | Failure reason |
    +----------------------+----------------+
    | keywords: (de-DE)    | found: tizen   |
    | description: (de-DE) | found: tizen   |
    | app name: (de-DE)    | found: tizen   |
    +----------------------+----------------+
    
    [!] precheck found one or more potential problems that must be addressed before submitting to review

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.5                                     |
[...]

PR #9766, opened 2017-07-14
Remove tizen from platforms rule, it's not needed and causes issues

taquitos:
Fix for #9761 
Longer-term, we could consider splitting words instead of just checking if each word is contained in the text.


mpirri (approved):
:shipit:

PR #9819, opened 2017-07-20
Set TargetRubyVersion: 2.0 in .rubocop.yml to avoid unwanted autocorrect with Ruby 2.0+

nafu:
### Checklist

### Motivation and Context

There are some rubocop errors with Ruby 2.4.x.
If we run `bundle exec rubocop -a` with Ruby 2.4.x, it automatically corrects some errors.
But we don't want it, fastlane supports 2.0+.

### Description

Update rubocop configuration and avoid unwanted autocorrect by running `bundle exec rubocop -a`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -217,6 +217,7 @@ Layout/SpaceAroundOperators:
     - '**/spec/actions_specs/xcodebuild_spec.rb'
 
 AllCops:
+  TargetRubyVersion: 2.4
   Include:
     - '**/fastlane/Fastfile'
   Exclude:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -217,7 +217,7 @@ Layout/SpaceAroundOperators:
     - '**/spec/actions_specs/xcodebuild_spec.rb'
 
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 2.0
   Include:
     - '**/fastlane/Fastfile'
   Exclude:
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -217,6 +217,7 @@ Layout/SpaceAroundOperators:
     - '**/spec/actions_specs/xcodebuild_spec.rb'
 
 AllCops:
+  TargetRubyVersion: 2.4
```
giginet: Is this true?

We are still supporting System Ruby (2.0.0).
I think this setting isn't suitable.

giginet (changes requested):

.rubocop.yml
nafu: @giginet 

Thanks. You're right 👍 
Because default ruby version is 2.4.x in my development env, the PR instruction cause autocorrection by rubocop with ruby 2.4.x.

> bundle exec rubocop -a to ensure the code style is valid

Should we define `TargetRubyVersion: 2.0` in .rubocop.yml?

giginet: According to the official document, it should be set minimum support version.

https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md#setting-the-target-ruby-version


```
If .ruby-version exists in the directory RuboCop is invoked in, RuboCop will use the version specified by it. Otherwise, users may let RuboCop know the oldest version of Ruby which your project supports with:
```

So we should set this to `2.0`. 👍

nafu: @giginet 

👍 

I updated branch and issue details!

KrauseFx (approved):
Nice catch, great change, thanks @nafu ❤️

giginet (approved):

nafu:
Thanks for your review 👍  @giginet @KrauseFx

PR #9830, opened 2017-07-22
Update empty archive error message for more instructions on how to solve the issue

KrauseFx:
Via https://github.com/artsy/emission-nebula/commit/705a32d610a94831e049ca218b5e94d0890ab28b cc @orta

### DIFF
--- a/gym/lib/gym/error_handler.rb
+++ b/gym/lib/gym/error_handler.rb
@@ -104,10 +104,11 @@ module Gym
         print "Also, make sure to have a valid code signing identity and provisioning profile installed"
         print "Follow this guide to setup code signing https://docs.fastlane.tools/codesigning/GettingStarted/"
         print "If your intention was only to export an ipa be sure to provide a valid archive at the archive path."
-        print "This error might also happen if your workspace/project file is not in the root directory of your project"
-        print "To workaround that issue, you can add `Dir.chdir('../path/to/dir/containing/proj') do`"
-        print "Check out https://github.com/artsy/emission-nebula/blob/77a516e9a84116d509c1f570c2525f05b0749f2e/fastlane/Fastfile#L34"
-        print "for example usage on how to navigate into the subfolder."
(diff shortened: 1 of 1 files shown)


gym/lib/gym/error_handler.rb
```diff
@@ -104,6 +104,10 @@ def handle_empty_archive
         print "Also, make sure to have a valid code signing identity and provisioning profile installed"
         print "Follow this guide to setup code signing https://docs.fastlane.tools/codesigning/GettingStarted/"
         print "If your intention was only to export an ipa be sure to provide a valid archive at the archive path."
+        print "This error might also happen if your workspace/project file is not in the root directory of your project"
+        print "To workaround that issue, you can add `Dir.chdir('../path/to/dir/containing/proj') do`"
```
taquitos: add this to what, and where?

KrauseFx: Yah, that's why I linked the code.... not sure how to do it better

orta: I wonder If it makes more sense for me to just answer this inside the stack overflow that this links to

taquitos: @KrauseFx  I was wondering where you would add `Dir.chdir` because it's not really clear here. I assume it's in the Fastfile.

@orta yeah, I think that's a really great idea too. I'd also leave in the specific commit that was mentioned here about checking out your example.

orta: I think I could fake a commit and push it up which would just be the `dir.chdir` + comments

KrauseFx: 👍

orta: https://github.com/artsy/emission-nebula/commit/44fe51a7fea8f7d52f0f77d6c3084827fe5dd59e

taquitos:
@googlebot it's ok.

taquitos (approved):
🐑

PR #9832, opened 2017-07-23
Properly rescue around writing the report.xml to the fastlane directory

mpirri:
### Checklist

### Description

If the `.fastlane` directory is not writable (running on CI, for instance), writing the `report.xml` file will fail with an EACCESS exception and cause _fastlane_ to crash.

This wraps the file access to that directory in a `rescue` block that will log a non-terminating error if this problem occurs.


taquitos (approved):
Sweet. 🚀

PR #9846, opened 2017-07-25
Remove donate_food plugin for now from fastlane setup

KrauseFx:
The plugin doesn't seem to work and I don't currently have the time to look closer into it


mpirri (approved):
😢

PR #9855, opened 2017-07-26
Set info plist output

dcordero:
### Checklist

### Motivation and Context
From the same plist file as template, I want to generate different plist files with different configurations without doing any change at all in the template.

### Description
Add a new output parameter to set_info_plist to allow setting a different output for the resultant plist file.

The changes are backward compatible. The new parameter is optional and the current behaviour is kept in case of not providing it.


dcordero:
I signed it!

NewStorm (approved):

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -55,7 +56,11 @@ def self.available_options
                                        optional: false,
                                        verify_block: proc do |value|
                                          UI.user_error!("Couldn't find plist file at path '#{value}'") unless File.exist?(value)
-                                       end)
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :output,
```
KrauseFx: Can we rename this to `output_file_name`?

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -21,7 +21,8 @@ def self.run(params)
             plist[params[:key]] = params[:value]
           end
           new_plist = Plist::Emit.dump(plist)
-          File.write(path, new_plist)
+          output = params[:output] || path
```
KrauseFx: This might be inside a subfolder that doesn't exist yet, right? Maybe something we should add here? 
```ruby
FileUtils.mkdir_p(...)
```

dcordero: Good point.
I included the folder creation and now also expanding the output path, just in case...

fastlane/lib/fastlane/actions/set_info_plist_value.rb
dcordero: done  ✅

dcordero:
Documentation added to fastlane/docs in this pull request: 
https://github.com/fastlane/docs/pull/371

dcordero:
@KrauseFx Thanks a lot for your feedback and your time reviewing this PR. The requested changes were applied to the code.

taquitos (approved):
Thanks for making the changes 👍  Also the tests you added are 💯

Issue #9782, opened 2017-07-17
[regression] fastlane changed default value for `export_method` in gym


Sega-Zero:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
After upgrading to 2.47.0 all my build starts failing in Travis-CI with ruby-2.3.0 via RVM saying my certificates are incorrect. Meanwhile all the same config and certificates were totallly fine in 2.46.1. See error log below.

calling gym like this:
```ruby
scheme = (options[:beta] ? "Release" : "Appstore")
	
    gym(
		workspace: 'Project.xcworkspace',
		scheme: scheme,
		clean: true, 
		output_directory: ".",
		output_name: "Project.ipa",
		silent: true,
		xcargs: options[:xcargs]
	   )
```

Using XCode 8. No XCode 9 installed.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```bash
[16:13:04]: ---------------------
[16:13:04]: --- Step: produce ---
[16:13:04]: ---------------------
+----------------+----------------------------------------+
|               Summary for produce 2.47.0                |
+----------------+----------------------------------------+
| app_identifier | [secure]								  |
| app_name       | [secure]								  |
| skip_itc       | true                                   |
| username       | [secure]                       |
| sku            | [secure]                             |
| platform       | ios                                    |
| language       | English                                |
| skip_devcenter | false                                  |
| team_id        | [secure]                             |
| team_name      | [secure]								  |
| itc_team_id    | [secure]								  |
+----------------+----------------------------------------+
[16:13:06]: [DevCenter] App '[secure]' already exists, nothing to do on the Dev Center
[16:13:07]: --------------------------------------
[16:13:07]: --- Step: increment_version_number ---
[16:13:07]: --------------------------------------
[16:13:12]: $ cd /Users/travis/build/[secure]/iOS/Project && agvtool new-marketing-version 1.1.0
[16:13:13]: ▸ Setting CFBundleShortVersionString of project Project to:
[16:13:13]: ▸ 1.1.0.
[16:13:13]: ▸ Updating CFBundleShortVersionString in Info.plist(s)...
[16:13:13]: ▸ Updated CFBundleShortVersionString in "Project.xcodeproj/../Project/Resources/Info.plist" to 1.1.0
[...]

PR #9863, opened 2017-07-26
Update provisioning profile mapping to only show mismatch error message if it's actually true

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/9782


mpirri (approved):
:shipit:

PR #9875, opened 2017-07-27
Allow multiple paths with wildcards in git_add action.

thii:
### Checklist

### Motivation and Context

Git doesn't have seperate concepts between path and pathspec when using
`git add` command, anything follows `git add` is accepted.

### Description

This combines `path` and `pathspec` parameters and allow multiple paths
with wildcards.

### DIFF
--- a/fastlane/lib/fastlane/actions/git_add.rb
+++ b/fastlane/lib/fastlane/actions/git_add.rb
@@ -63,7 +63,9 @@ module Fastlane
           'git_add(path: "./version.txt")',
           'git_add(path: ["./version.txt", "./changelog.txt"])',
           'git_add(path: "./Frameworks/*")',
-          'git_add(path: ["*.h", "*.m"])'
+          'git_add(path: ["*.h", "*.m"])',
+          'git_add(pathspec: "./Frameworks/*")',
+          'git_add(pathspec: "*.txt")'
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/git_add.rb
+++ b/fastlane/lib/fastlane/actions/git_add.rb
@@ -2,14 +2,11 @@ module Fastlane
   module Actions
     class GitAddAction < Action
       def self.run(params)
-        if params[:pathspec]
-          paths = params[:pathspec]
-          UI.success("Successfully added from \"#{paths}\" 💾.")
-        elsif params[:path]
+        if params[:path]
           if params[:path].kind_of?(String)
-            paths = params[:path].shellescape
+            paths = params[:path]
           else
-            paths = params[:path].map(&:shellescape).join(' ')
+            paths = params[:path].join(' ')
           end
           UI.success("Successfully added \"#{paths}\" 💾.")
         else
@@ -32,23 +29,8 @@ module Fastlane
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :path,
-                                       description: "The file you want to add",
+                                       description: "The file(s) and path(s) you want to add",
                                        is_string: false,
-                                       conflicting_options: [:pathspec],
-                                       optional: true,
-                                       verify_block: proc do |value|
-                                         if value.kind_of?(String)
-                                           UI.user_error!("Couldn't find file at path '#{value}'") unless File.exist?(value)
-                                         else
-                                           value.each do |x|
-                                             UI.user_error!("Couldn't find file at path '#{x}'") unless File.exist?(x)
-                                           end
-                                         end
-                                       end),
-          FastlaneCore::ConfigItem.new(key: :pathspec,
-                                       description: "The pathspec you want to add files from",
-                                       is_string: true,
-                                       conflicting_options: [:path],
                                        optional: true)
         ]
       end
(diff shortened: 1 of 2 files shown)


thii:
Circle CI is failing at `brew install shellcheck`.

ohayon:
Hey @thii, thanks for submitting this PR! I am hesitant to merge this as is because it removes the `pathspec` option which might break existing configurations. Would you mind just making sure that this will stay backwards compatible? Maybe by checking to see if the `pathspec` option is present and then combining it in?

Thanks!! 🐙  🚀

ohayon (changes requested):
Just want to make sure we arent breaking people's current setups.

thii:
@ohayon It should be good to go now.

fastlane/lib/fastlane/actions/git_add.rb
```diff
@@ -70,8 +62,8 @@ def self.example_code
           'git_add',
           'git_add(path: "./version.txt")',
           'git_add(path: ["./version.txt", "./changelog.txt"])',
-          'git_add(pathspec: "./Frameworks/*")',
-          'git_add(pathspec: "*.txt")'
+          'git_add(path: "./Frameworks/*")',
+          'git_add(path: ["*.h", "*.m"])'
```
ohayon: We still need to call `git_add` with the `pathspec` option though right?

ohayon (changes requested):
Thanks for adding this back and adding the deprecated flag, but we want to make sure that people will still be able to have this working without changing any code, would you mind calling `git_add` with the `pathspec` param still for now?

fastlane/lib/fastlane/actions/git_add.rb
```diff
@@ -7,9 +7,9 @@ def self.run(params)
           UI.success("Successfully added from \"#{paths}\" 💾.")
         elsif params[:path]
```
ohayon: Hey @thii sorry for the back and forth... If I'm understanding the purpose of this PR correctly, shouldn't we be combining the pathspec and the path params to add all the files specified by those params instead of checking with `elsif` here?

Let me know if I am just misunderstanding the PR! 🐙

thii: Hi @ohayon you're right. The reason of keeping the`elseif` here is because these params are conflicting with each other. I'll remove the conflicting option since we don't need it anymore.

thii: @ohayon Btw, since we are deprecating the `pathspec` parameter, shouldn't we not include it in the example code?

thii:
@ohayon Hi again. The original motivation was to move the function of `pathspec` into `path` and deprecate the `pathspec`, not "combine" those (I was using the wrong word), so I think it would be better to keep the current implementation of this PR. 

Since both parameters are conflicting with each other, none of existing users should be using both in their config. And we should not allow doing this for a deprecated option. What do you think? 🤔

fastlane/lib/fastlane/actions/git_add.rb
```diff
@@ -7,9 +7,9 @@ def self.run(params)
...
           if params[:path].kind_of?(String)
-            paths = params[:path].shellescape
+            paths = params[:path]
           else
-            paths = params[:path].map(&:shellescape).join(' ')
+            paths = params[:path].join(' ')
```
ohayon: Can we make sure that this is an array here?

ohayon (dismissed):
Hey @thii - apologies for taking so long to get back to you on this PR, just one change I mentioned above and then we should be good to 🚢

fastlane/lib/fastlane/actions/git_add.rb
```diff
@@ -7,9 +7,9 @@ def self.run(params)
           UI.success("Successfully added from \"#{paths}\" 💾.")
         elsif params[:path]
           if params[:path].kind_of?(String)
-            paths = params[:path].shellescape
-          else
-            paths = params[:path].map(&:shellescape).join(' ')
+            paths = params[:path]
```
ohayon: Shouldn't we still be shellescaping this?

fastlane/lib/fastlane/actions/git_add.rb
```diff
@@ -7,9 +7,9 @@ def self.run(params)
...
+          elsif params[:path].kind_of?(Array)
+            paths = params[:path].join(' ')
```
ohayon: Shouldn't we still be shellescaping this?

ohayon (commented):
Hey @thii! Apologies for the huge delay reviewing this. Just two more little comments, let me know your thoughts and we will get it in right away! 🚀

minuscorp:
Hi @thii! Is this PR still alive? You have requested changes pending and some failing tests on the way! 🚀

janpio:
The test failure is not because of the actual code, but because this PR is quite old and is using a now unsupported CI configuration. Rebasing on `master` should fix this quite easily so the tests can test the actual code.

getaaron:
@janpio @thii @ohayon I rebased these changes onto `master` and the new tests are running now

joshdholtz:
@getaaron @ohayon I added a commit which adds a new option called `shell_escape` which defaults to true (for backwards compatability) but can be turned off if using wildcard

getaaron (approved):

getaaron:
@joshdholtz this looks good to merge to me 👍

thii:
Thanks! 👍

PR #9894, opened 2017-07-28
Adds a div around the actions so that CSS can be applied to it

orta:
### Checklist
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)  🔑

### Motivation and Context

I felt like the actions page could use a bit more hierarchy 

screen shot 2017-07-28 at 12 30 30

### Description

So this PR adds two divs for styling `.category-actions` and `.action` - then the docs can use these to apply CSS to those elements

I also added comments to the `end`s, so I could keep track of which was which.

### DIFF
--- a/fastlane/lib/assets/Actions.md.erb
+++ b/fastlane/lib/assets/Actions.md.erb
@@ -20,10 +20,14 @@ import './path/to/other/Fastfile'
 <%- end -%>
 - [Plugins](https://docs.fastlane.tools/plugins/available-plugins)
 
-
 <%- @categories.each do |category, actions| %>
 # <%= category %>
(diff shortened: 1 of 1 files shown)


orta:
Tests are still running locally

ohayon (approved):
Thanks for this @orta!

ohayon:
Before we merge this, should we add some indentation to make it a bit more clear what elements are nested in others?

orta:
I initially thought about that, but you're using markdown, and so indentation is important to the end result - you would end up with really illogical indentation

ohayon:
Thanks @orta !

PR #9896, opened 2017-07-28
[WIP] mark required named parameters

snatchev:
### Checklist

### Motivation and Context

fastlane uses Ruby 2.0's named parameters, but all named parameters in Ruby 2.0 must have a default value. Some parameters are not optional, and we must assert that they have values. Using this module, we can do that via the following pattern:

If the module `RequiredParamters` is mixed into your class, any method can be annotated with parameter requirements like this:

```ruby
def some_method(foo: nil, bar: nil, baz: nil) requires_params(:foo, :bar)
  .
  .
  .
end
```

when `some_method()` is called, `foo` and `bar` must be none-nil

### DIFF
--- a/spaceship/lib/spaceship/helper/required_parameters.rb
+++ b/spaceship/lib/spaceship/helper/required_parameters.rb
@@ -1,0 +1,49 @@
+module RequiredParameters
+  ##
+  # RequiredParameters a mixin to provide a few methods for asserting the presense of named parameter.
+  #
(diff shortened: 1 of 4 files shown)


taquitos:
I love this idea.

snatchev:
@taquitos I briefly looked into giving special rubocop rules for this, but I am not sure that's a great idea.

Would it be better if the `requires_params` was just the first line of the method definition?

KrauseFx:
Since this has been open for about half a year now, I'm gonna go ahead and close this for now. Feel free to re-open 👍

PR #9649, opened 2017-07-01
[match] add template parameter

hermanolsson:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
For some users of Match, the lack of support for the `template` (entitlements) parameter when creating provisioning profiles is a deal breaker. For the apps my team is working on, we rely on specific entitlements configured in the dev portal when creating/updating a provisioning profile.

This has been discussed in a few issues before, but we cannot find a pull request with this change. The issue we used as guidance is #4548. We know this was kind of dismissed back then, but hopefully this PR will make sense, since it breaks nothing for regular usage, and it will be of great benefit to the ones actually using this `template` parameter.

The changes made in this pull request have been tested with our own provisioning profiles both for development and distribution.

### Description
We have made changes to Match, Sigh and Spaceship.

#### Match
* Added a template parameter and option

#### Sigh
* Added template parameter and option

#### Spaceship
* Use the template parameter in the dev portal calls

#### Tests
* Updated tests to handle the new template parameter

I'm not an every day ruby developer, but I have tried to follow all guidances for making changes to fastlane and a pull request.

Issue #9558, opened 2017-06-22
Including Special Entitlements during Provisioning Profile creation


Fluzzarn:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Currently when creating Provisioning Profiles (either automatically using match or manually using sigh). Profiles are created with "Default" entitlements, which work for a majority of developers. However, for the few that rely on using Special Entitlements  (Such as the Network Extension Entitlement), we have to avoid fastlane regenerating provisioning profiles. 

After some experimentation, modifying the params sent in create_provisioning_profile! in portal_client, I was able to create a provisioning profile with the correct entitlements.

As seen in multiple issues ( #9227 ,  #3471 , #4548 ) this seems like it would be a welcomed change to fastlane

taquitos:
This looks like a reasonable addition profile generation, is that something you'd want to take on? It sounds like you already have some code as a proof of concept, maybe it wouldn't be too difficult to turn that into a PR? We have a sweet getting started guide [on your FirstPR ](https://github.com/fastlane/fastlane/blob/master/YourFirstPR.md)if you haven't submitted before

joeybaker:
psss… @fastlane-bot comments aren't the only things that count as activity ;)

thelvis4:
The functionality was implemented in #9912 and released in version `2.55.0`.
Give it a try! 😉

KrauseFx:
That's so awesome, thanks again everyone @thelvis4 

@joeybaker hehe, yah :) the bot didn't know about the PRs :P

Issue #9227, opened 2017-05-19
sigh --force generate provisioning profile with incorrect Custon/Special Entitlements options


chongj615:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
When you regenerate a provisioning profile using the force flag to add all new devices, the regenerated provisioning profile doesn't take into account the Entitlements options, instead always come back as Default.

This is the entitlement in provisioning profile I refer to:
![image](https://cloud.githubusercontent.com/assets/4042482/26266711/6e6557c8-3c9b-11e7-9aaa-c83345769afa.png)

I believe this is similar to this issue
https://github.com/fastlane/fastlane/issues/4548

fastlane sigh --adhoc -a com.nnn.nnn -u chong@nnn.com --team_id NNNNNNN --provisioning_name=adhoc-ios --force 

+-------------------------------------+---------------------+
|                  Summary for sigh 2.32.1                  |
+-------------------------------------+---------------------+
| adhoc                               | true                |
| app_identifier                      | com.nnn.nnn |
| username                            | chong@nnn.com  |
| team_id                             | nnnnnnn         |
| provisioning_name                   | adhoc-ios           |
| force                               | true                |
| development                         | false               |
| skip_install                        | false               |
| ignore_profiles_with_different_name | false               |
| skip_fetch_profiles                 | false               |
| skip_certificate_verification       | false               |
| platform                            | ios                 |
+-------------------------------------+---------------------+

[13:53:32]: Starting login with user 'chong@nnn.com'
[13:53:34]: Successfully logged in
[13:53:34]: Fetching profiles...
[13:53:35]: Verifying certificates...
[13:53:36]: Found 1 matching profile(s)
[13:53:36]: Recreating the profile
[13:53:37]: Creating new provisioning profile for 'com.nnnn.nnn' with name 'adhoc-ios' for 'ios' platform
[13:53:38]: Downloading provisioning profile...
[13:53:38]: Successfully downloaded provisioning profile...
[13:53:38]: Installing provisioning profile...
/Users/chong/AdHoc_com.nnn.nnn.mobileprovision

### Environment
<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.4                                     |
[...]

Issue #4548, opened 2016-05-06
Support custom entitlements


wilrnh:
### New Issue Checklist
- [ x] Updated fastlane to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
### Issue Description

This is probably more of a feature request. Really, I'd like match to be able to create profiles with custom entitlements.

However, I'd be OK with being able to force match to download profiles, without re-generating them. That way, I'd be able to create profiles in the apple dev console, with custom entitlements, myself. Then run match, which would simply download the new profile, and commit it to git for distribution. Is there anyway this may be possible? Or maybe a different work-around?

Thanks!
##### Complete output when running fastlane, including the stack trace and command used

```
INSERT OUTPUT HERE
```
### Configuration Files

Please copy the complete content of your `Fastfile` and any other configuration files you use below:

Fastfile:

``` ruby
# INSERT CONTENT HERE
```
### Environment

fastlane version (run `fastlane -v`):

Do you use bundler to execute fastlane (i.e. `bundle exec fastlane`)?

Do you use a Ruby environment manager (e.g. `chruby`, `rbenv`, `rvm`)?

wilrnh:
#3471

wilrnh:
Does match do a full delete and create of the profile when re-generating a profile? If not, I could probably create my own profile with custom entitlements, using the naming format that match uses (match Adhoc com.myappid) and from then on, match would only ever use that profile, regenerating off of it when new devices are added, or a cert is regenerated.

possible?

wilrnh:
Seems like match actually checks if the profile it currently has is in the dev console, and if it isn't creates a new one. Also, if a profile already exsists with the format match uses (match AdHoc com.myappid) it just creates a new one with a unix-timestamp appended :(

So, no-go on that workaround...

wilrnh:
Went down the rabbithole on this one:

Seems that match uses sigh to create the profiles, and sigh uses spaceship to interact with the dev console. So this feature would require changes across spaceship, sigh, and match.

https://github.com/fastlane/fastlane/blob/master/match/lib/match/generator.rb#L43
https://github.com/fastlane/fastlane/blob/master/sigh/lib/sigh/runner.rb#L113
https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/portal_client.rb#L329

[...]

PR #9912, opened 2017-08-01
Add support for provisioning profile templates

thelvis4:
### Checklist

### Motivation and Context
Some developer accounts (not all) have assigned templates for provisioning profiles. Because not all accounts have templates assigned, there is just a niche of Fastlane users that will need this feature, but they will need it _badly_. Without it, they cannot generate, repair or update template-based provisioning profiles.
**What is a provisioning profile template?**
I couldn't find any official documentation about templates, but from my understanding, a template defines a set of extended entitlements (entitlements that couldn't be set from Xcode or by managing an App ID in Developer Portal) the app will support.
**How does this look in Developer Portal?**
For the accounts that have assigned provisioning profile templates, when creating or editing a provisioning profile, there will be an additional field called *Entitlements*.
provisioning_profile_template

#### Issues and other PRs
Issues: https://github.com/fastlane/fastlane/issues/9558 (thanks to https://github.com/Fluzzarn for the idea behind this solution), https://github.com/fastlane/fastlane/issues/9227, https://github.com/fastlane/fastlane/issues/4548, https://github.com/fastlane/fastlane/issues/3471

Other PR: https://github.com/fastlane/fastlane/pull/9649
I noticed there is another PR that addressed the same problem when I almost finished the work on this. I decided to open this PR because it adds support for repairing and updating (additionally to generating) template based provisioning profiles.

#### Testing
This was tested by creating and repairing provisioning profiles for iOS and tvOS apps using *Sigh*.
Not tested for macOS (not even sure if it's applicable).
**If someone could test this in other contexts (using Match, for example), it will be greatly appreciated!**

### Description
This PR touches Spaceship, Sigh and Match.
**Spaceship**: Added `template` to the list of parameters sent to Dev Portal API when creating and updating a provisioning profile.
If the data about a template-based profile doesn't contain the data about the template (usually it's a dictionary with only `nil`s), we have to make an additional API request and fetch all the data about the profile. If the profile contains needed data, that data is used to initialize provisioning profile template model. This logic was also documented in the code.
**Sigh**, **Match**: Added the optional `template_name` option. Other than that, Sigh and Match just redirect the calls to Spaceship, which does all the work

**Tests**: Tests were updated to handle the new feature.


thelvis4:
@KrauseFx Is there anything I could help with in order to progress with this PR?

@Fluzzarn, @chongj615, @wilrnh If you have an Apple Developer account supporting templates (extended entitlements), it would be very helpful if could give a try to this PR and let us know whether this works for your setup or not.
Thank you!

KrauseFx (approved):
Hey @thelvis4, thanks for this great PR, I love the updated docs & all the tests. I haven't tested this on my actual account, but the code looks great, let's see where this is going 🚀 

Thanks for your contribution ❤️

chongj615:
Thank you, confirm fastlane sigh is now working with the template options!

PR #9914, opened 2017-08-01
App review submission fails

aronpammer:
### Checklist

### Motivation and Context

We were trying to automate the app submission for review for itunes.

This is the code we were using:

      itunes_app = get_itunes_app(options[:app])
      version = itunes_app.edit_version
      builds = version.candidate_builds
      version.select_build(builds.first)
      version.save!

      UI.success("Successfully selected build")

      submission = itunes_app.create_submission

      # Set app submission information
      submission.add_id_info_uses_idfa = true
      submission.add_id_info_serves_ads = true
      submission.add_id_info_tracks_install = true
      submission.add_id_info_tracks_action = true
      submission.add_id_info_limits_tracking = true

      # Finalize app submission
      submission.complete!

but the `submission.complete!` part failed with the following message: `Problem processing review submission.`
We found two issues here. 

The first one was that the url fastlane was using to send the app to review was incorrect.
Fastlane's version: `ra/apps/<app_id>/version/submit/complete`
The correct one: `ra/apps/<app_id>/versions/<version>/submit/complete`

The second, bigger issue came up when we weren't setting the `content_rights_has_rights` and the `content_rights_contains_third_party_content` variables to either true or false. In this case, iTunes connect was expecting the `contentRights` variable to be `null`, but fastlane sent the following instead:
`
"contentRights": {
   "containsThirdPartyContent": {
      "value": null
      },
      "hasRights": {
         "value": null
      }
}
`
and this caused the submission failure.

We tested these changes on one of our applications. We were able to send it to review successfully with these changes.
### Description

app_submission.rb:
In the complete method I manually set the `contentRights` variable to nil in case either the `content_rights_has_rights` or the `content_rights_contains_third_party_content` is nil. (both of them has to have a value to work properly)
I also passed the version_id to the send_app_submission method as a parameter. (see below)

base.rb:
One other thing I found while debugging was that fastlane was sending an AppVersion object in the payload to iTunes Connect (as `version`). I am not sure if any other classes use DataHash, therefore I also made a type check there to not to delete anything from the payload accidentally. 

tunes_client.rb:
Here version_id was necessary in the url, because the original url was broken.

### DIFF
--- a/spaceship/lib/spaceship/base.rb
+++ b/spaceship/lib/spaceship/base.rb
@@ -59,6 +59,9 @@ module Spaceship
       def to_json(*a)
         h = @hash.dup
         h.delete(:application)
+        if h.key?(:version) && h[:version].is_a?(Spaceship::Tunes::AppVersion)
+          h.delete(:version)
+        end
         h.to_json(*a)
       end
 
--- a/spaceship/lib/spaceship/base.rb
+++ b/spaceship/lib/spaceship/base.rb
@@ -59,7 +59,7 @@ module Spaceship
       def to_json(*a)
         h = @hash.dup
         h.delete(:application)
-        if h.key?(:version) && h[:version].is_a?(Spaceship::Tunes::AppVersion)
+        if h.key?(:version) && h[:version].kind_of?(Spaceship::Tunes::AppVersion)
           h.delete(:version)
         end
         h.to_json(*a)
--- a/spaceship/lib/spaceship/base.rb
+++ b/spaceship/lib/spaceship/base.rb
@@ -43,6 +43,10 @@ module Spaceship
         ref[last] = value
       end
 
+      def delete(key)
+        @hash.delete(key)
+      end
+
       def lookup(keys)
         head, *tail = *keys
         if tail.empty?
@@ -59,9 +63,6 @@ module Spaceship
       def to_json(*a)
         h = @hash.dup
         h.delete(:application)
-        if h.key?(:version) && h[:version].kind_of?(Spaceship::Tunes::AppVersion)
-          h.delete(:version)
-        end
         h.to_json(*a)
       end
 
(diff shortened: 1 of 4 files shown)


velvetz7:
I signed it!

spaceship/lib/spaceship/base.rb
```diff
@@ -59,6 +59,9 @@ def each(&block)
       def to_json(*a)
         h = @hash.dup
         h.delete(:application)
+        if h.key?(:version) && h[:version].kind_of?(Spaceship::Tunes::AppVersion)
```
ohayon: I wonder if we can handle this logic somewhere that is a bit more obvious, like in the app submission itself? Just to make sure it's clear for anyone else coming into this code?

ohayon (dismissed):
Hey @velvetz7! This is a great PR thank you so much for making this fix!

I just left one little note about maybe moving up that version check so it isn't so hidden in the abstraction and so we don't keep any logic in the `to_json` function. Definitely open to hearing your thoughts on that though! 

Thanks again! 🐙

spaceship/lib/spaceship/base.rb
velvetz7: I had the same thoughts as you, and initially I put this logic into the complete! method, however deleting the version variable there could be a bit confusing as well, because then after calling the complete method I am not able to access the version variable. There are a lot of ways this could be implemented, I chose the easiest/clearest one. (not the prettiest, I admit) I am open to suggestions though, which method would you move this logic to? 

Do you know any other classes that is a subclass of this base class and needs to send the version variable to itunes connect? If not, then we could just simply delete the version variable alongside with the application variable.

ohayon: I think maybe the best move would be to copy the `raw_data` in the `complete!` method, remove the version from that copy, and then use that copied instance to pass to `send_app_submission` on the client. Does that make sense? 🐙

taquitos (approved):
Yay ❤️  🐐

PR #9924, opened 2017-08-02
Add pem option active_days_limit

luikore:
So user can specify how many days before current push certificate
expires should fastlane pem generate a new one.

### Checklist

### Motivation and Context
I have many push certificates to renew.
And I need them renew before 32 days of expiry (just more than the longest month).

### Description
Added an option `current_active_days_limit` and the default value is same as previous hard-coded 30 days.

### DIFF
--- a/pem/lib/pem/options.rb
+++ b/pem/lib/pem/options.rb
@@ -23,7 +23,7 @@ module PEM
                                      description: "If the current certificate is active for less than this number of days, generate a new one. Default value is 30 days",
                                      default_value: 30,
                                      verify_block: proc do |value|
-                                       UI.user_error!("Value of active_days_limit must be a positive integer or left blank") if value.to_s !~ /\A\d*\z/
+                                       UI.user_error!("Value of active_days_limit must be a positive integer or left blank") if value.to_i <= 0
                                      end),
         FastlaneCore::ConfigItem.new(key: :force,
                                      env_name: "PEM_FORCE",
--- a/pem/lib/pem/options.rb
+++ b/pem/lib/pem/options.rb
@@ -18,9 +18,16 @@ module PEM
                                      description: "Generate a p12 file additionally to a PEM file",
                                      is_string: false,
                                      default_value: true),
+        FastlaneCore::ConfigItem.new(key: :current_active_days_limit,
+                                     env_name: "PEM_CURRENT_ACTIVE_DAYS_LIMIT",
+                                     description: "If current one is active for no more than this number of days, generate a new one. Default value is 30",
+                                     default_value: 30,
+                                     verify_block: proc do |value|
+                                       UI.user_error!("Value of current_active_days_limit must be a positive integer or left blank") if value.to_s !~ /\A\d*\z/
+                                     end),
         FastlaneCore::ConfigItem.new(key: :force,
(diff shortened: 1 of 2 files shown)


pem/lib/pem/options.rb
```diff
@@ -18,9 +18,16 @@ def self.available_options
                                      description: "Generate a p12 file additionally to a PEM file",
                                      is_string: false,
                                      default_value: true),
+        FastlaneCore::ConfigItem.new(key: :current_active_days_limit,
+                                     env_name: "PEM_CURRENT_ACTIVE_DAYS_LIMIT",
+                                     description: "If current one is active for no more than this number of days, generate a new one. Default value is 30",
```
ohayon: Could we change this description to say "If the current certificate is active for less than this number of days, generate a new one. Default value is 30 days."

pem/lib/pem/options.rb
```diff
@@ -18,9 +18,16 @@ def self.available_options
                                      description: "Generate a p12 file additionally to a PEM file",
                                      is_string: false,
                                      default_value: true),
+        FastlaneCore::ConfigItem.new(key: :current_active_days_limit,
```
ohayon: Can we change this key name to be something like `:active_days_limit` or something else that is a bit more self-explanatory?

ohayon (changes requested):
Hey @luikore, thanks so much for the PR! Welcome to the _fastlane_ community! ❤️  This is a good change I think. I just left a couple comments about the language we should use and what we should call the key and environment variable. Would love to hear your thoughts! 🚀  🐙

pem/lib/pem/options.rb
luikore: Sure. but, you know, the trailing `.` is not allowed in description.

pem/lib/pem/options.rb
luikore: OK, but I don't have a better name than `:active_days_limit`

ohayon (approved):

pem/lib/pem/options.rb
```diff
@@ -18,9 +18,16 @@ def self.available_options
                                      description: "Generate a p12 file additionally to a PEM file",
                                      is_string: false,
                                      default_value: true),
+        FastlaneCore::ConfigItem.new(key: :active_days_limit,
+                                     env_name: "PEM_ACTIVE_DAYS_LIMIT",
+                                     description: "If the current certificate is active for less than this number of days, generate a new one. Default value is 30 days",
+                                     default_value: 30,
+                                     verify_block: proc do |value|
+                                       UI.user_error!("Value of active_days_limit must be a positive integer or left blank") if value.to_s !~ /\A\d*\z/
```
ohayon: Can we check here `if value <= 0` instead of doing the string conversion and comparing against a regex?

ohayon (changes requested):
Hey @luikore, sorry I missed one thing in my initial review of this PR, just want to compare the value as a number rather than converting it to a string in the verify block. Thanks! 👑

pem/lib/pem/options.rb
luikore: OK

ohayon (approved):
Thanks @luikore 👍  🚀 !

Issue #9394, opened 2017-06-07
Failure after all screenshots taken, html not created


ko0f:
### New Issue Checklist

- [✅] Updated fastlane to the latest version
- [✅] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

After taking all screenshots, an exception is thrown, screenshots.html not created.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you
```

[10:22:49]: Collecting screenshots...
[10:22:49]: Found 5 screenshots...
[10:22:49]: Copying '/Users/...-d41d8cd98f00b204e9800998ecf8427e.png'...
[10:22:49]: Copying '/Users/...-d41d8cd98f00b204e9800998ecf8427e.png'...
[10:22:49]: Copying '/Users/...-d41d8cd98f00b204e9800998ecf8427e.png'...
[10:22:49]: Copying '/Users/...-d41d8cd98f00b204e9800998ecf8427e.png'...
[10:22:49]: Copying '/Users/...-d41d8cd98f00b204e9800998ecf8427e.png'...

Looking for related GitHub issues on fastlane/fastlane...

Found no similar issues. To create a new issue, please visit:
https://github.com/fastlane/fastlane/issues/new
Run `fastlane env` to append the fastlane environment to your issue
/Users/.../.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.37.0/fastlane_core/lib/fastlane_core/string_filters.rb:38:in `wordwrap': [!] upper is smaller than lower in repeat range: /.{1,0}(?:\s+|$)/ (RegexpError)
	from /Users/.../.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.37.0/fastlane_core/lib/fastlane_core/print_table.rb:75:in `transform_row'
	from /Users/.../.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.37.0/fastlane_core/lib/fastlane_core/print_table.rb:102:in `block (2 levels) in transform_output'
	from /Users/.../.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.37.0/fastlane_core/lib/fastlane_core/print_table.rb:101:in `map'
[...]

PR #9935, opened 2017-08-02
Fix crash while trying to wordwrap

ohayon:
Fix https://github.com/fastlane/fastlane/issues/9394

### DIFF
--- a/fastlane_core/spec/wordwrap_spec.rb
+++ b/fastlane_core/spec/wordwrap_spec.rb
@@ -1,0 +1,20 @@
+require 'fastlane_core/string_filters'
+
+describe WordWrap do
+  context 'wordwrapping will return an array of strings' do
+    let(:test_string) { 'some big long string with lots of characters i think there should be more than eighty carachters here!' }
+
+    it 'will return an empty array if the length is zero' do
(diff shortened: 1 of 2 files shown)


KrauseFx (approved):
Nice catch 👍

KrauseFx (approved):
Tests ❤️

fastlane_core/spec/wordwrap_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    it 'will return an empty array if the length is zero' do
+      result = test_string.wordwrap(0)
+      expect(result).to eq([])
+    end
+
+    it 'will return an array of strings, each being <= the length passed to wordwrap' do
+      wrap_length = 5
+      result = test_string.wordwrap(wrap_length)
+      string_lengths = result.map { |string| string.length }
+      expect(result).to all( be_a(String) )
```
taquitos: is not being a String an option?

ohayon: not really? my thought was it might be nil if the code was changed?

taquitos: Oh no, you can have a nil in an array 😶

PR #8047, opened 2017-01-29
Portal Person MGMT into spaceship 👩‍👩‍👦

hjanuschka:
adds support for person mgmt on portal via spaceship.

as requested in
  * https://github.com/fastlane/fastlane/issues/7903

related:
 * https://github.com/fastlane/fastlane/pull/7993

bildschirmfoto 2017-01-30 um 10 59 17

# API

## List all persons:
(returns array of `Spaceship::Portal::Person`

```ruby
Spaceship::Portal::Persons.all
```

## Find by email:

```ruby
person = Spaceship::Portal::Persons.find("helmut@januschka.com")
```

## Create/Invite a person

```ruby
Spaceship::Portal::Persons.invite("helmut1@januschka.com", "admin")
```

## Delete a person

```ruby
person = Spaceship::Portal::Persons.find("helmut@januschka.com")
person.remove!
```

## Change role of member.

```ruby
person = Spaceship::Portal::Persons.find("helmut@januschka.com")
person.change_role("member")
```

PR #9938, opened 2017-08-02
Add support for listing invites (aka pending members)

abl:
### Checklist

### Motivation and Context
It's currently possible to send invites using Portal but there's no way to view the status of invites that are in flight. Combined with a group of administrators onboarding new users this can lead to a certain amount of duplication.

These changes were tested against a live account; basic rspec coverage was added as well.

### Description
This PR, building on #8047, wraps the `/getInvites` endpoint and emits `Invite` objects.

I purposely kept invited users separate from actual users; an argument could be made that invited users should be shown in `Spaceship::Portal::Persons.all` but the available metadata for an `Invite` differs from the metadata of a `Person`. It might make sense to move this out to `Spaceship::Portal::Invites` (along with `Spaceship::Portal::Persons.invite`) but I wanted to get feedback first.

dateJoined in Person is a formatted string; dateCreated/dateExpires comes back as a timestamp. For consistency, this PR alters Person to parse a Time object as well and updates the relevant specs.


spaceship/lib/spaceship/portal/invite.rb
```diff
@@ -0,0 +1,43 @@
...
+        'inviterName' => :inviter_name,
+        'recipientEmail' => :email_address,
+        'recipientRole' => :type,
+        'dateCreated' => :created,
+        'dateExpires' => :expires
+      )
+
+      class << self
+        def factory(attrs)
+          # rubocop:disable Style/RescueModifier
```
KrauseFx: It would be great if we can avoid using inline `rescue` without an error type specification. Is `attrs['dateCreated']` `nil` sometimes? If so, can we check of `nil` values instead, or catch only the exception type we expect here?

spaceship/lib/spaceship/portal/person.rb
```diff
@@ -42,6 +42,15 @@ def remove!
       def change_role(role)
         client.team_set_role(team_member_id, role)
       end
+
+      class << self
+        def factory(attrs)
+          # rubocop:disable Style/RescueModifier
```
KrauseFx: Same here, ideally we don't want to disable that `rubocop` rule

spaceship/spec/portal/invite_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      "inviteId" => "A0B4C6D8EF",
+      "recipientRole" => "ADMIN",
+      "inviterName" => "Felix Krause",
+      "dateCreated" => created,
+      "recipientEmail" => "abl@g00gle.com",
+      "dateExpires" => expires
+    }
+    invited = Spaceship::Portal::Invite.factory(attrs)
+    # Roles are downcased to match Person
+    expect(invited.type).to eq("admin")
```
KrauseFx: Great tests ❤️

KrauseFx (changes requested):
Hey @abl, thanks for the pull request, this looks great 👍 The only thing that we'd love to have changed before merging the change is to remove all inline `rescue` blocks. Let me know if my comments make sense.

spaceship/spec/portal/fixtures/inviteList.json
```diff
@@ -0,0 +1,32 @@
...
+      "recipientRole": "MEMBER",
+      "inviterName": "David Ohayon",
+      "dateCreated": 1500664007000,
+      "recipientEmail": "kaipi@g00gle.com",
+      "dateExpires": 1503273599000
+    },
+    {
+      "inviteId": "A2B4C6D8EF",
+      "recipientRole": "MEMBER",
+      "inviterName": "Joshua Liebowitz",
```
taquitos: best inviter 😄

taquitos:
Woooo this is awesome!

abl:
Thanks :)

I copied the rescue pattern from [certificate.rb:231](https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/certificate.rb#L231-L235) - in this case, the intent is that if Apple gives us back a new date format we can still return the raw API value.

I've rewritten the inline rescues as standard rescues and added specs to cover - it feels like I'm missing a better way to write this kind of capture, though?

abl:
Meta - I'm looking at the raw patches and I'm not seeing any issues on authorship. @KrauseFx, you were already in the Google organization when you committed, right? If so, we should point the Googlebot team at this PR.

abl:
(also, thanks for the typo fix!)

abl:
Sorry, had to rebase and pull this branch forward (we're using it internally) and it looks like that broke the review.

KrauseFx (approved):
This is great, thanks 👍

PR #9944, opened 2017-08-03
Revise SSL verification handling for debugging

nafu:
### Checklist

### Motivation and Context

The execution of `OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE` should be revised.

### Description

Use Faraday option to use `OpenSSL::SSL::VERIFY_NONE` instead of using `OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE`.

### DIFF
--- a/fastlane/lib/fastlane/actions/hockey.rb
+++ b/fastlane/lib/fastlane/actions/hockey.rb
@@ -1,9 +1,3 @@
-if ENV["SPACESHIP_DEBUG"]
-  require 'openssl'
-  # this has to be on top of this file, since the value can't be changed later
-  OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE
-end
-
 module Fastlane
   module Actions
     module SharedValues
@@ -26,15 +20,6 @@ module Fastlane
           builder.response :json, content_type: /\bjson$/
           builder.use FaradayMiddleware::FollowRedirects
           builder.adapter :net_http
-          if ENV['SPACESHIP_DEBUG']
-            # for debugging only
-            # This enables tracking of networking requests using Charles Web Proxy
-            builder.proxy "https://127.0.0.1:8888"
-          end
-
-          if ENV["DEBUG"]
-            puts "To run _spaceship_ through a local proxy, use SPACESHIP_DEBUG"
-          end
         end
       end
 
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -12,12 +12,6 @@ require 'cgi'
 
 Faraday::Utils.default_params_encoder = Faraday::FlatParamsEncoder
 
-if ENV["SPACESHIP_DEBUG"]
-  require 'openssl'
-  # this has to be on top of this file, since the value can't be changed later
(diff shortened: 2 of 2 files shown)


ohayon (commented):
Hey @nafu - I'm not sure I totally understand this PR, is this going to set `VERIFY_NONE` for all requests because we are not checking for the `SPACESHIP_DEBUG` flag anymore?

nafu:
Hi @ohayon We still use SPACESHIP_DEBUG flag and set VERIFY_NONE only for debugging.

https://github.com/fastlane/fastlane/pull/9944/files#diff-a9086da2d7e8a66b647fdabdb0a7cca7R257

Now SPACESHIP_DEBUG setup is in one place like below. 👍 

```
if ENV['SPACESHIP_DEBUG']
    # for debugging only
    # This enables tracking of networking requests using Charles Web Proxy
    c.proxy "https://127.0.0.1:8888"
    c.ssl[:verify_mode] = OpenSSL::SSL::VERIFY_NONE
end
```

ohayon (approved):
Ah! My bad! I missed that if statement. Thanks for the information! 🚀

fastlane/lib/fastlane/actions/hockey.rb
```diff
@@ -26,15 +20,6 @@ def self.connection(options)
           builder.response :json, content_type: /\bjson$/
           builder.use FaradayMiddleware::FollowRedirects
           builder.adapter :net_http
-          if ENV['SPACESHIP_DEBUG']
-            # for debugging only
-            # This enables tracking of networking requests using Charles Web Proxy
-            builder.proxy "https://127.0.0.1:8888"
```
ohayon: Do we need to set `builder.ssl[:verify_mode] = OpenSSL::SSL::VERIFY_NONE ` Here also? It feels like we are deleting something here that we might want?

fastlane/lib/fastlane/actions/hockey.rb
```diff
@@ -26,15 +20,6 @@ def self.connection(options)
...
-          end
-
-          if ENV["DEBUG"]
-            puts "To run _spaceship_ through a local proxy, use SPACESHIP_DEBUG"
```
ohayon: How come we are deleting this here?

ohayon (commented):
Sorry, @nafu I missed a couple of things there that I just have questions about! 😄

fastlane/lib/fastlane/actions/hockey.rb
nafu: I think this isn't a part of spaceship. I assume these might be copy and pasted from spaceship faraday code. So I delete this from hockey action.

fastlane/lib/fastlane/actions/hockey.rb
nafu: Same reason above, but if I should revert this let me know 👍

ohayon:
I think you're probably right for both of those things @nafu ! Lets go ahead and merge! 🚀  Do you want to press merge?

nafu:
👍 @ohayon Thanks. I'll press 🚀

PR #9957, opened 2017-08-04
Readme: Group toolchain list

janpio:
### Checklist

### Motivation and Context
As a new user it took a bit of time to wrap my head around all the different tools available in fastlane's awesome toolchain. I was missing a list of what belongs where and is used for what, especially for tools that I don't need right now but could be useful down the lane (*ka-ching).

### Description
I took the list that was already part of the readme and grouped it by OS, then by "subject" of the tools. This way it is a lot easier for newcomers to "classify" all the tools of the fastlane toolchain.

If this is not wanted for the README, I think it should be added somewhere else, possibly the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,22 +127,42 @@ If you'd like to take a look at a project already using `fastlane` check out [fa
 
 In addition to `fastlane`'s commands, you also have access to these `fastlane` tools:
 
+### iOS
+
+#### App Store
+- [`produce`](https://github.com/fastlane/fastlane/tree/master/produce): Create new iOS apps on iTunes Connect and Dev Portal using the command line
+- [`precheck`](https://github.com/fastlane/fastlane/tree/master/precheck): Check your app using a community driven set of App Store review rules to avoid being rejected
 - [`deliver`](https://github.com/fastlane/fastlane/tree/master/deliver): Upload screenshots, metadata, and your app to the App Store
-- [`supply`](https://github.com/fastlane/fastlane/tree/master/supply): Upload your Android app and its metadata to Google Play
+
+#### Screenshots
 - [`snapshot`](https://github.com/fastlane/fastlane/tree/master/snapshot): Automate taking localized screenshots of your iOS and tvOS apps on every device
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,22 +127,42 @@ If you'd like to take a look at a project already using `fastlane` check out [fa
...
 - [`boarding`](https://github.com/fastlane/boarding): The easiest way to invite your TestFlight beta testers
+
+#### Building
 - [`gym`](https://github.com/fastlane/fastlane/tree/master/gym): Building your iOS apps has never been easier
-- [`match`](https://github.com/fastlane/fastlane/tree/master/match): Easily sync your certificates and profiles across your team using Git
-- [`scan`](https://github.com/fastlane/fastlane/tree/master/scan): The easiest way to run tests for your iOS and Mac apps
-- [`precheck`](https://github.com/fastlane/fastlane/tree/master/precheck): Check your app using a community driven set of App Store review rules to avoid being rejected
+
+#### Plumbing
+- [`spaceship`](https://github.com/fastlane/fastlane/tree/master/spaceship): Ruby library to access the Apple Dev Center and iTunes Connect
```
ohayon: Can we keep this one up with certificates and provisioning profiles for now?

ohayon (changes requested):
Hey @janpio, thanks for this PR! This is a great organizational change. Just one comment about the position of _spaceship_, but otherwise 💯

README.md
janpio: Sure - felt "wrong" here anyway. (Although: Spent most time of the PR thinking about a headline for this... :1st_place_medal: )

KrauseFx (changes requested):
.

ohayon:
Hey @janpio, after a bit of discussion with the team, we are going to hold off on merging this change as we are going to be doing some work on our docs in the near future and going to take this type of thing into consideration 😄  Thank you so much for your contribution though and we are really happy to have you in the community! 🚀  🐙  🏎

janpio:
Ok. If I am confused again I can look at the readme at my `patch-1` now ;)

KrauseFx:
Hehe :)

janpio:
Did anything happen in the last month regarding this topic?

PR #9963, opened 2017-08-04
Readme: inform about limitations of `supply`

janpio:
### Checklist

### Motivation and Context
`supply` doesn't do some things that people might want it to do. by adding this information to the README they won't waste their time searching for a solution.

### Description
added a new headline and 2 short list items that explain that `supply` can't create new apps and doesn't support some metadata fields and both has to be taken care of manually for now by the developer.


ohayon (changes requested):
Hey @janpio, instead of adding this limitation section, maybe we could just add a *Note* to the `Uploading an APK` section mentioning that new apps have to be created manually. Also, I'm not sure it's worth comparing it to _produce_ as many of our Android users are not familiar with that tool. 

Regarding the metadata, instead of mentioning this as a limitation, we would love to have this functionality and should aim to getting a change into _supply_ that will support those fields! 👨‍🎤 💯

Issue #9922, opened 2017-08-01
Uploading assets with a GitHub release fails in `set_github_release`


provTheodoreNewell:
### New Issue Checklist

- [x ] Updated fastlane to the latest version
- [x ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Using the `set_github_release` action in Fastlane 2.50.0 with a valid array of paths for the `upload_assets` and relying on the default value for `git_url` (https://api.github.com) fails during the upload step, printing the error:
> Please provide server URL, eg: https://api.github.com

This seems to be a regression introduced with the new `github_api` action at https://github.com/fastlane/fastlane/pull/8971.

### Proposed Solution

In `github_api.rb::construct_url`, valid combinations of parameters should be either:
1. `server_url` in conjunction with `path` that can be joined to create a full url
or
2. `url` as the full url

I will prepare a pull request with this solution.

### Fastlane Version

</details>

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
| fastlane | 2.50.0  | ✅ Up-To-Date  |

</details>

ohayon:
Thanks for raising the issue @provTheodoreNewell ! We would be happy to review that PR ! 🚀

JohnEstropia:
We're still experiencing a problem with the new Github actions and it seems to be related to content-types:
```
[16:56:03]: Creating release of <redacted> on tag "<redacted>" with name "<redacted>".
[16:56:03]: Will also upload assets ["<redacted>.ipa", "<redacted>.app.dSYM.zip"].
[16:56:04]: Successfully created release at tag "<redacted>" on GitHub
[16:56:04]: See release at "https://github.com/<redacted>/releases/tag/<redacted>"
[16:56:05]: Uploading <redacted>.ipa
[16:56:32]: GitHub responded with 400:{"message":"Invalid Content-Type: ''","request_id":"CC20:3C83:6E76CA:755512:59842897"}
+-------------------------------+-------------------------------------------------------------------------------------------------------------------------------+
|                                                                         Lane Context                                                                          |
+-------------------------------+-------------------------------------------------------------------------------------------------------------------------------+
| DEFAULT_PLATFORM              | ios                                                                                                                           |
| PLATFORM_NAME                 | ios                                                                                                                           |
| LANE_NAME                     | ios deploy                                                                                                                    |
[...]

PR #9965, opened 2017-08-04
Fix for #9922, Part 2: content type

provTheodoreNewell:
### Checklist

### Motivation and Context

As commented on #9922, an additional regression with the introduction of the new `github_api` action at #8971 causes file uploads to fail in the `set_github_release` action when the filename does not include ".zip".  

### Description

This PR restores the pre-existing behavior of using the `application/zip` content type header for all uploaded binary files.


ohayon (approved):
AMAZING! Thanks for making these fixes @provTheodoreNewell !! 🚀

provTheodoreNewell:
😄

ganjarpanji:
is this not released yet? could we make the release faster? as our teams need to attach the binaries immediately 😄

ganjarpanji:
ah nevermind, its still possible with .zip though. thanks for the fix anyway @provTheodoreNewell

KrauseFx:
Hehe, we generally avoid shipping new releases during a weekend 😉

PR #9976, opened 2017-08-07
Remove `--login` from xcbuild-safe.sh as per suggestion

KrauseFx:
I don't know enough about it, but here we go: https://github.com/fastlane/fastlane/issues/9804

Seems like more and more people run into this, in particular with Xcode 9 it seems to take a lot longer. How can we avoid a problem once the Xcode 9 upgrade is happening?


viniciushisao (approved):

ohayon (approved):
I dont know too much about it either, but yeah I've seen a few people run into this. So I would think its at least worth trying out in the wild!

PR #9986, opened 2017-08-08
Limit user access when creating new App in iTunes

javibm:
### Checklist

### Motivation and Context

I have the need to limit the user access when creating a new Application in iTunes.

I have created an app using spaceship with and without limit user access. I also used Produce to test both scenarios.

### Description

I have created a new option in Produce to pass an array of emails for limiting the user access. If any email is registered as user in iTunes, the new app is limited for these ones.


produce/lib/produce/options.rb
```diff
@@ -65,6 +65,13 @@ def self.available_options
                                      description: "Skip the creation of the app on iTunes Connect",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :itc_users,
+                                     short_option: "-s",
+                                     env_name: "ITC_USERS",
+                                     optional: true,
+                                     type: Array,
+                                     description: "Array with Itunes Connect Users for Limit User Access",
```
KrauseFx: This text isn't clear directly, can we rephrase to something like `Array of iTunes Connect users that have access to newly created app? If provided, no other users will have access` (at least that's how I understood it, is that correct?)

KrauseFx:
Nice change, thanks for the contribution, I added one comment. Also, can you make sure you write `iTunes` everywhere?

javibm:
I have updated the description to be more accurate. Also, I have checked that i write in all the create application for `iTunes`.

KrauseFx (approved):
Nice, thanks for updating the PR, this is a great change 👍

Issue #9695, opened 2017-07-07
fastlane requires a minimum version of Xcode 7.0


lkraider:
### New Issue Checklist

- [ X ] Updated fastlane to the latest version
- [ X ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Fastlane refuses to run any command when Xcode is set to an older version.

We have some CIs that build on old Xcode versions that use fastlane sigh to download the provisioning profiles. Since upgrading from fastlane 2.37 to 2.45 fastlane will not run anymore and complain about Xcode version.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
DEVELOPER_DIR=/Applications/Xcode-4.5.2.app/Contents/Developer/

21:35:46 + sigh --adhoc --username XXX --team_id XXX --app_identifier 'XXX' --skip_install

21:35:48 Fastlane update available
21:35:48 Please update fastlane by downloading an updated bundle from
21:35:48 https://download.fastlane.tools

21:35:51 [WARNING] You are calling sigh directly. Usage of the tool name without the `fastlane` prefix is deprecated in fastlane 2.0
21:35:51 Please update your scripts to use `fastlane sigh --adhoc --username XXX --team_id XXX --app_identifier XXX --skip_install` instead.

21:35:53 
21:35:53 [!] fastlane requires a minimum version of Xcode 7.0, please upgrade and make sure to use `sudo xcode-select -s /Applications/Xcode.app`
```

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

```
$ fastlane env
Fastlane update available
Please update fastlane by downloading an updated bundle from
https://download.fastlane.tools

[!] fastlane requires a minimum version of Xcode 7.0, please upgrade and make sure to use `sudo xcode-select -s /Applications/Xcode.app`
```

lkraider:
Breakage was introduced by https://github.com/fastlane/fastlane/pull/9609

lkraider:
Editing `MINIMUM_XCODE_RELEASE = "4.0"` fixes everything in:
`~/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.45.0/fastlane/lib/fastlane/version.rb`

KrauseFx:
[...]

PR #9998, opened 2017-08-08
Do not attempt to retrieve Xcode version on non-Mac platform

sander-m:
### Checklist
- [x] I've updated the documentation if necessary. (N/A)

### Motivation and Context

Running any Fastlane command on a Linux host will print the following warning:

`No such file or directory - xcode-select`

This is confusing for end users. It doesn't fail the process.

Related:
https://github.com/fastlane/fastlane/issues/9695
https://github.com/fastlane/fastlane/pull/9609

### Description

To fix this, I added one line to immediately return `nil` in the method `xcode_version` when invoked on a non-Mac platform. This prevents Fastlane from attempting to run `xcode-select` to retrieve the Xcode path on a Linux host.
From what I understand,  [fastlane-runner](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb#L51) will properly rescue the exception.
Alternatively this check could be removed entirely from the fastlane-runner itself on non-Mac platforms.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -140,6 +140,7 @@ module FastlaneCore
 
     # @return The version of the currently used Xcode installation (e.g. "7.0")
     def self.xcode_version
+      return nil unless self.is_mac?
       return @xcode_version if @xcode_version
 
       begin


sander-m:
Signed the CLA

KrauseFx (approved):
Nice, that's a great change, thanks for your contribution @sander-m 👍

Issue #9939, opened 2017-08-03
Error "No provisioning profile provided" using Xcode 9


andreaspfurtscheller:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Since we are developing our project using Xcode 9 beta, codesigning for testflight won't work as it did when we were developing with Xcode 8.
 - using **match** + **gym**
 - Xcode's and fastlane's version are the only things that have changed.
 - I have looked at the threads on the issues #9782, #9909 and #9890, but in my case fastlane selects the expected profile, so I think it might be another issue.
 - Building and exporting using Xcode works flawlessly, so the project settings are ok, I guess.
 - I have also re-configured the code signing settings following the Xcode 9 upgrade guide.
 - Also the provisioning profiles and certificates are valid and installed according to match.
 
At that point I don't know how to fix this by myself, maybe you can help me with this.

### Captured Output

Command Used: `beta`
<details><summary>Output/Log</summary>

```

Successfully loaded Appfile at path '/Users/andreas/Documents/project/fastlane/Appfile'
- app_identifier: 'com.company.project'
- apple_id: '■■■■'
- team_id: '■■■■'
-------

+----------------+--------------------------------------+
|      Detected Values from './fastlane/Matchfile'      |
+----------------+--------------------------------------+
| git_url        | git@github.com:■■■■                  |
|                | /project-certificates.git            |
| app_identifier | ["com.company.project"]              |
| username       | ■■■■                                 |
+----------------+--------------------------------------+


+-----------------------+--------------------------------------+
|                   Summary for match 2.50.1                   |
+-----------------------+--------------------------------------+
| type                  | appstore                             |
| readonly              | true                                 |
| git_url               | git@github.com:■■■■                  |
|                       | /project-certificates.git            |
| app_identifier        | ["com.company.project"]              |
| username              | ■■■■                                 |
| git_branch            | master                               |
| keychain_name         | login.keychain                       |
| team_id               | ■■■■                                 |
| verbose               | false                                |
| force                 | false                                |
| skip_confirmation     | false                                |
| shallow_clone         | false                                |
| clone_branch_directly | false                                |
| force_for_new_devices | false                                |
| skip_docs             | false                                |
| platform              | ios                                  |
+-----------------------+--------------------------------------+


+-------------------+-----------------------------------------------+
[...]

Issue #9890, opened 2017-07-28
Gym auto selects provisioning profile mapping wrong in an Xcode project with multiple configs


dral3x:
### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Running the lane `deploy_testflight`, gym generate a plist file with the wrong provisioning profile 
```
[13:27:07]: Generated plist file with the following values:
[13:27:07]: ▸ -----------------------------------------
[13:27:07]: ▸ {
[13:27:07]: ▸   "provisioningProfiles": {
[13:27:07]: ▸     "com.xxx.yyy${BUNDLE_ID_SUFFIX}": "xxx Dev AdHoc Distribution"
[13:27:07]: ▸   },
[13:27:07]: ▸   "method": "app-store"
[13:27:07]: ▸ }
[13:27:07]: ▸ -----------------------------------------
```
`BUNDLE_ID_SUFFIX` is a user define config variabile I have in my project.
It's value, for a TestFlight build (`ProdRelease` config) is empty btw.

The problem is that the chosen provisioning profile is wrong. To deploy the app on TestFlight, the AppStore distribution profile must be used.
The Xcode project is correctly configured to do so

---

Edit by @joshdholtz: This method is the causing this issue when Xcode project have multiple configurations - https://github.com/fastlane/fastlane/blob/6a82b16373568b0d2cd793998dd6164503831aef/gym/lib/gym/detect_values.rb#L89-L120. It sometimes selects the wrong profile and makes things go 💥 

---

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.5                                     |
| Ruby                        | 2.4.0                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.11.0                          |
| Installation Source         | /usr/local/bin/fastlane                     |
| Host                        | Mac OS X 10.12.5 (16F73)                    |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/2.4.0/lib            |
| OpenSSL Version             | OpenSSL 1.0.2j  26 Sep 2016                 |
| Is contained                | false                                       |
[...]

PR #10020, opened 2017-08-09
[match] Add support for provisioning profile mapping for multiple app identifiers

KrauseFx:
Previously we didn't consider the use-case of having multiple app identifiers there

Fixes https://github.com/fastlane/fastlane/issues/9939


joshdholtz:
@KrauseFx Does this also possibly fix #9890? 🤔

KrauseFx:
Mh, not sure, what do you think?

joshdholtz:
@KrauseFx It looks like it might 😊  I will test out tomorrow morning 💪

ohayon (approved):
🎸

Issue #10028, opened 2017-08-10
[regression] can't dup NilClass in latest release (2.53.0)


Sega-Zero:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

After upgrading to the latest fastlane, got the error "can't dup NilClass". Seems like changes in #9952 regressed the things for Xcode < 9. Really need a fix ASAP.

##### Complete output when running fastlane, including the stack trace and command used

```ruby
[17:13:07]: -----------------
[17:13:07]: --- Step: gym ---
[17:13:07]: -----------------
[17:13:07]: Sending crash report...
[17:13:07]: The stacktrace is sanitized so no personal information is sent.
[17:13:07]: To see what we are sending, look here: /Users/travis/.fastlane/latest_crash.json
[17:13:07]: Learn more at https://github.com/fastlane/fastlane#crash-reporting
[17:13:07]: You can disable crash reporting by adding `opt_out_crash_reporting` at the top of your Fastfile
[17:13:10]: ▸ [17:13:09]: Open https://github.com/fastlane/fastlane/issues/new to submit a new issue ✅
+------------------------------------+---------------------------------------------------------+
|                                         Lane Context                                         |
+------------------------------------+---------------------------------------------------------+
| DEFAULT_PLATFORM                   | ios                                                     |
| PLATFORM_NAME                      | ios                                                     |
| LANE_NAME                          | ios auto                                                |
| PRODUCE_APPLE_ID                   |                                                         |
| SIGH_PROFILE_TYPE                  | app-store                                               |
| MATCH_PROVISIONING_PROFILE_MAPPING | {"com.project.id"=>"match AppStore com.project.id"} |
| VERSION_NUMBER                     | 1.0.1                                              |
| BUILD_NUMBER                       | 1.0.1                                               |
+------------------------------------+---------------------------------------------------------+
[17:13:10]: can't dup NilClass
+------+---------------------------------------------------------------+-------------+
|                                  fastlane summary                                  |
+------+---------------------------------------------------------------+-------------+
| Step | Action                                                        | Time (in s) |
+------+---------------------------------------------------------------+-------------+
[...]

PR #10031, opened 2017-08-10
Fix dup crash on nil for gym

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/10028

Works fine on my machine,  I assume the behavior changed with one the recent Ruby updates


taquitos (approved):
🐳

PR #10037, opened 2017-08-11
ensure_xcode_version: clarify description

janpio:
### Checklist

### Motivation and Context
When skimming the description you could get the impression it was talking about a beta version of the app for Testflight which gives a wrong picture what the action actually does.

### Description
Description now contains "of Xcode" to make sure it is clear about what version it is talking about.


taquitos (approved):
🏆 thanks!

PR #10039, opened 2017-08-11
change action `tryouts` to "beta" category

janpio:
### Checklist

### Motivation and Context
While reading through the available default actions I noticed this `tryouts` action in the "misc" category. As Tryouts.io is a (beta) distribution service, this is not really correct.

### Description
I changed the category from "misc" to "beta".


taquitos (approved):
👍

Issue #10075, opened 2017-08-17
fastlane run add_git_tag: undefined method `delete' for nil:NilClass


ened:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

```
$ fastlane run add_git_tag 
[00:04:18]: For more information, check out https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[00:04:20]: -------------------------
[00:04:20]: --- Step: add_git_tag ---
[00:04:20]: -------------------------

Looking for related GitHub issues on fastlane/fastlane...

➡️  fastlane supply fails with error "NoMethodError: [!] undefined method `map' for nil:NilClass"
    https://github.com/fastlane/fastlane/issues/10033 [open] 3 💬
    18 hours ago

➡️  [!] undefined method `sub' for nil:NilClass (NoMethodError)
    https://github.com/fastlane/fastlane/issues/8426 [closed] 5 💬
    06 Jun 2017

➡️  [!] undefined method `native_targets' for #<Xcodeproj::Project:[MEMORY_INSTANCE]>
    https://github.com/fastlane/fastlane/issues/8935 [closed] 2 💬
    4 weeks ago

and 37 more at: https://github.com/fastlane/fastlane/search?q=undefined%20method%20%60delete'%20for%20nil&type=Issues&utf8=✓

🔗  You can ⌘ + double-click on links to open them directly in your browser.
/usr/local/var/rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/fastlane-2.53.1/fastlane/lib/fastlane/actions/add_git_tag.rb:6:in `run': [!] undefined method `delete' for nil:NilClass (NoMethodError)
	from /usr/local/var/rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/fastlane-2.53.1/fastlane/lib/fastlane/runner.rb:252:in `block (2 levels) in execute_action'
	from /usr/local/var/rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/fastlane-2.53.1/fastlane/lib/fastlane/actions/actions_helper.rb:50:in `execute_action'
[...]

PR #10080, opened 2017-08-17
Prevent calling method on nil when lane_name is nil in add_git_tag

taquitos:
Fix for #10075


KrauseFx (approved):
Nice catch, looks great. Might be worth adding one line of comment, because it might not immediately clear on why this might be `nil`

taquitos:
Such comment. So communicative.

ohayon:
wow.

PR #10083, opened 2017-08-18
Add retry logic for new exception: ITunesConnectPotentialServerError

taquitos:
Sometimes you get a bad response from whatever API you're calling and it's not your fault. If you use `handle_itc_response` from `TunesClient` you can now pass `flaky_api_call: true` and it will auto-retry 3 times, with 10 second waits in-between. 

This motivation came after reading #9859, I thought "gee whiz, this sure looks like flaky API if all you have to do is retry your upload script" So, let's do it 😜

### Checklist
(https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -10,6 +10,10 @@ module Spaceship
     class ITunesConnectTemporaryError < ITunesConnectError
     end
 
+    # raised if the server failed to save, and it might be caused by an invisible server error
+    class ITunesConnectPotentialServerError < ITunesConnectError
+    end
+
     attr_reader :du_client
 
     def initialize
@@ -108,7 +112,9 @@ module Spaceship
     end
 
     # rubocop:disable Metrics/PerceivedComplexity
-    def handle_itc_response(raw)
+    # If the response is coming from a flaky api, set flaky_api_call to true so we retry a little.
+    # Patience is a virtue.
+    def handle_itc_response(raw, flaky_api_call: false)
       return unless raw
       return unless raw.kind_of? Hash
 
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -108,7 +112,9 @@ def send_login_request(user, password)
     end
 
     # rubocop:disable Metrics/PerceivedComplexity
-    def handle_itc_response(raw)
+    # If the response is coming from a flaky api, set flaky_api_call to true so we retry a little.
+    # Patience is a virtue.
+    def handle_itc_response(raw, flaky_api_call: false)
```
KrauseFx: This has to be the best parameter in the whole code base ❤️

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -108,7 +112,9 @@ def send_login_request(user, password)
     end
 
     # rubocop:disable Metrics/PerceivedComplexity
-    def handle_itc_response(raw)
+    # If the response is coming from a flaky api, set flaky_api_call to true so we retry a little.
```
KrauseFx: Should we add that to more areas in the code base?

KrauseFx (approved):
So perfect ❤️

spaceship/lib/spaceship/tunes/tunes_client.rb
taquitos: yes, we totally should, I just wanted to be surgical about it. I wasn't 100% sure which apis needed this yet.

janpio:
1 year later we have https://github.com/fastlane/fastlane/issues/13832 and I think I figured out what is _really_ going on - a data upload is failing in the background, causing this strange error message when you try to save the app. Will post updates in #13832.

But that raises the question: Does this logic actually work and help in some cases?

Issue #8759, opened 2017-04-03
Automate recording video using Snapshot?


CraigSiemens:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
As of Xcode 8.2, the simulator has added support for recording video using `xcrun simctl io booted recordVideo <filename>.<file extension>`

It would be great if snapshot could be used to record videos. Combined with using UITests to automate input, this would make it easy to create App Preview or training videos.

KrauseFx:
Hey @CraigSiemens, sorry for not coming back earlier - this is a great idea, I'd love it if _snapshot_ could generate app videos for you

laullon:
I would like to take a shot on implementing this, is it possible?

KrauseFx:
Yeah, that would be awesome, happy to review & merge a pull request adding this functionality. Check out our [contributing.md](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md) docs with instructions on how to get the development environment up and running 👍

laullon:
ok, working on it ;)

laullon:
I just made a basic implementation:

https://github.com/laullon/fastlane/commit/413df988bb06e71e1573ba33ee98ee32e1c263d0

Can you take a quick look?
I still need to do some work and tests.

KrauseFx:
Hey that's great, do you think we can do that without adding another dependency to _fastlane_?

laullon:
sure

PR #10121, opened 2017-08-23
[snapshot] Ability to record video

laullon:
### Checklist

### Motivation and Context
This change add Snapshot to the ability to record videos of iOS applications, using the same method used to take screenshots.
There is a bug open for this: https://github.com/fastlane/fastlane/issues/8759
This was tested using:
* Xcode 8.3
* Xcode 9.0 (Beta 6)

### Description
I have created a light/simple HTTP server that is used by `SnapshotHelper.swift` to send the `startRecording` and `stopRecording` commands to `snapshot.rb` that will execute\kill the `xcrun simctl io booted recordVideo`


laullon:
I signed it!

laullon:
I need to do some changes to apply recent changes on Snapshot

laullon:
done, all is working now

fastlane_core/lib/fastlane_core/command_executor.rb
```diff
@@ -30,7 +30,7 @@ def which(cmd)
       # @param prefix [Array] An array containing a prefix + block which might get applied to the output
```
joshdholtz: Can we get the `pidCreated` added to this comment/doc here? ^ 😇

joshdholtz (changes requested):
This is super cool! Requested one change to add method doc on `execute` for new parameter.

Also, is there anyway to add tests for this? 😇

fastlane_core/lib/fastlane_core/command_executor.rb
laullon: sure

laullon: done

KrauseFx:
Hey @laullon, that's so awesome, thanks for your contribution, I love the idea to use snapshot to create videos. Could you tell us a little more about why we need a HTTP server to communicate? I thought it should be possible to do that without a web server.

brunomunizaf:
I need this PR to be approved so much... Really appreciate the work @laullon !

laullon:
I added the "HTTP Server" because I need a way to communicate the XCTests running on the simulator and the Snapshot process, so it knows when start and stop the recording. I did it that way because:

1. Is very simple to implement on the Snapshot side, just a TCPServer listening on a random port (for security)
2. On the simulator side is very simple to send the command, just a simple URLSession
3. Supports multiples clients (which came handy for Xcode 9 multi simulator support)

This doesn't represent a security problem because this server only support 2 get URLs, any other URL or HTTP command is ignored, any HTTP header is also ignored, and it listen on a different port every time.

KrauseFx:
Yeah, I'm not too worried about the security in this case, but about the complexity that this will add to the code base

laullon:
yeah, "HTTP Server" sounds complex, but is really very simple...

Anyway, if you have a better way in mind let me know, I'll try it.

LordNali:
This is a great addition to snapshot - especially since Apple is pushing towards the usage of videos in the AppStore. 

What is the current position on this PR? Does is need some more testing or is the http server a problem? The http server might be handy for further additions / actions to control the simulator.

KrauseFx:
Yeah, I thought it's possible using
```
xcrun simctl io booted recordVideo
```

laullon:
I use "xcrun simctl io" to record the video, but I don't use the "booted" option because in Xcode 9 you can have multiple simulator running, so it had to called with a divide ID.

KrauseFx:
Yeah, but that should still work, right? First, launch up a simulator, get its UDID, and then use that to record the video. For video recording in _snapshot_ (or a new action/tool?), I feel like it would totally be okay to require Xcode 9 to use the feature 👍

laullon:
That's what I do... and I works on Xcode 8 and 9

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -32,6 +32,27 @@ func snapshot(_ name: String, waitForLoadingIndicator: Bool = true) {
...
 
+func startRecording(name: String) {
+    sendCommand(commnad: "startRecording",args: "name=\(name)")
+}
+
+func stopRecording() {
+    sendCommand(commnad: "stopRecording")
+}
+
+func sendCommand(commnad: String,args: String = "") {
```
ffittschen: There is a typo in the attribute name: `commnad`

fastlane_core/lib/fastlane_core/command_executor.rb
```diff
@@ -26,11 +26,12 @@ def which(cmd)
       # @param command [String] The command to be executed
       # @param print_all [Boolean] Do we want to print out the command output while running?
       # @param print_command [Boolean] Should we print the command that's being executed
+      # @param pidCreated [Block] A block that's called when the new process is spawned and we have a PID, it can be used later on to control the new process
       # @param error [Block] A block that's called if an error occurs
       # @param prefix [Array] An array containing a prefix + block which might get applied to the output
       # @param loading [String] A loading string that is shown before the first output
       # @return [String] All the output as string
-      def execute(command: nil, print_all: false, print_command: true, error: nil, prefix: nil, loading: nil)
+      def execute(command: nil, print_all: false, print_command: true, pidCreated: nil, error: nil, prefix: nil, loading: nil)
```
ffittschen: The other attributes of this function all use snake_case instead of camelCase. This should be consistent

ffittschen:
I agree with @KrauseFx on this. There should be an simpler way to pass the command from Xcode to fastlane other than using a http server. Maybe it just has to try it from another direction: don't start to record from inside of the UITest, but start the relevant parts of the UITest from outside of Xcode (before/after starting to record).

I could imagine something like `snapshot record_videos:true` will internally look similar to this execution of `scan`
```ruby
scan(only_testing: "TestBundleA/TestSuiteVideo")
```

directly followed this part of your code
```ruby
Thread.new do
    FastlaneCore::CommandExecutor.execute(command: "xcrun simctl io booted recordVideo #{language_folder}/#{name}.mp4",
                                          print_all: true,
                                          print_command: true,
                                          loading: "Recording video...",
                                          pidCreated: proc do |pid|
                                            @recording_pid = pid
                                          end,
                                          error: proc do |output, return_code|
                                            ErrorHandler.handle_test_error(output, return_code)
                                            UI.error "Caught error... #{return_code}"
                                            UI.error "Caught output... #{output}"
                                          end) 
end
```

fastlane_core/lib/fastlane_core/command_executor.rb
laullon: solved

snapshot/lib/assets/SnapshotHelper.swift
laullon: solved

ohayon:
Hey folks! I am just going through some older PRs and checking out the status of them. I wanted to bump this in case there was any more discussion to be had and see if we still thought this should get merged in. @laullon did you have a chance to consider the feedback of the http server? 🚀

KrauseFx:
Thanks everyone, I'm gonna go ahead and close the PR for now. I think we should be able to implement video recording feature in _snapshot_ without the need of setting up a local HTTP server 👍 

Thanks for everyone's contribution on this, and sorry for having to close this PR at this time

Jules2010:
What a shame :(

Jules2010:
Oh no, it doesn't provide my video in landscape.
Snapshot does rotate for me, 'simctl' doesn't obey this :(

Jules2010:
Any ideas how to get rid of the 1 second of black screen at the start of the video :(

Jules2010:
Any ideas how to get rid of the 1 second of black screen at the start of the video :(

Edit: this could be several seconds. Also it delayed my first view controller in the video. I’m pretty sure the snapshot erase simulator was the cause.

@laullon any chance you can merge in recent changes from master? I tried on another branch and couldn’t get it to work, I reckon your more familiar with your changes.

One suggestion I do have is that the success of the snapshot still requires a screenshot, also putting the video in the results page would be good.

However this all works for me, good job.

For anyone else reading, I was confused over the http listener stuff but this is all contained within the swift / fastlane, you don’t need to do anything in your app to use this.

laullon:
I'm not working on this right now... don't have time to it...
But there is another PR in progress: https://github.com/fastlane/fastlane/pull/11744

Jules2010:
Yeah thats me :)

laullon:
ohhh!!! right ;)

PR #10129, opened 2017-08-24
More options for download_dsyms action

addbrick:
### Checklist

### Motivation and Context

The ability to specify where the dSYMs are saved when being downloaded from iTunes Connect.
Also the ability to only download the latest build's dSYM instead of having to look up the latest version before calling `download_dsyms`.

I integrated this into my build system and have run it multiple times to check it grabs the latest, grabs a specific version, and all versions of build when downloading dSYMs.

### Description

- Add `save_directory` option to `download_dsyms` action. Allows specifying where the downloaded dSYMs are saved. Useful when you don't want dSYMs in the directory the command is run in.
- Adds keyword value `'latest'` to `version` option in `download_dsyms` action. Forces download of only the latest build's dSYM. Useful when you only need the dSYM for the latest version of the app. Having it be an option means the user doesn't have to figure out, or keep updating, the latest version.

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -14,16 +14,31 @@ module Fastlane
         Spaceship::Tunes.select_team
         UI.message("Login successful")
 
+        # Get App
+        app = Spaceship::Application.find(params[:app_identifier])
+        unless app
+          UI.user_error!("Could not find app with bundle identifier '#{params[:app_identifier]}' on account #{params[:username]}")
+        end
+
+        # Process options
         version = params[:version]
         build_number = params[:build_number]
         platform = params[:platform]
         save_directory = params[:save_directory]
+
+        # Set version if it is latest
+        if version == 'latest'
+          latest_build = latest_build(app.latest_version.candidate_builds)
+          version = latest_build.train_version
+          build_number = latest_build.build_version
+        end
 
         # Make sure save_directory has a slash on the end
         if save_directory && !save_directory.end_with?('/')
           save_directory += '/'
         end
 
+        # Write a nice message
         message = []
         message << "Looking for dSYM files for #{params[:app_identifier]}"
         if version
@@ -35,11 +50,8 @@ module Fastlane
         end
 
         UI.message(message.join(" "))
-        app = Spaceship::Application.find(params[:app_identifier])
-        unless app
-          UI.user_error!("Could not find app with bundle identifier '#{params[:app_identifier]}' on account #{params[:username]}")
-        end
 
+        # Loop through all app versions and download their dSYM
         app.all_build_train_numbers(platform: platform).each do |train_number|
           if version && version != train_number
             next
@@ -83,6 +95,21 @@ module Fastlane
         http.use_ssl = (uri.scheme == "https")
         res = http.get(uri.request_uri)
         res.body
+      end
+
+      def self.latest_build(candidate_builds)
+        if (candidate_builds || []).count == 0
+          UI.user_error!("Could not find any available candidate builds on iTunes Connect to submit")
+        end
+
+        build = candidate_builds.first
+        candidate_builds.each do |b|
+          if b.upload_date > build.upload_date
+            build = b
+          end
+        end
+
+        return build
       end
 
       #####################################################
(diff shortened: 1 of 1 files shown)


addbrick:
I signed it!

addbrick:
This started to fail `bundle exec rubocop -a` because of `Perceived complexity for run is too high [19/18]`. It started to happen after this commit https://github.com/fastlane/fastlane/pull/10129/commits/916d1268435871b7d1e92dbd094f7d0bbf96b829.

Removing the `UI.user_error!` call reduces the complexity enough, so if we don't get a version, the script will naturally error out when calling a function on `nil`.

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -14,10 +14,32 @@ def self.run(params)
...
         build_number = params[:build_number]
         platform = params[:platform]
+        save_directory = params[:save_directory]
+
+        # Set version if it is latest
+        if version == 'latest'
+          # Try to grab the live version first, else fallback to edit version
+          latest_version = app.live_version(platform: platform) || app.edit_version(platform: platform)
+          version = latest_version.version
+          build_number = latest_version.build_version
```
ohayon: Where is this being used besides printing it below?

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -141,12 +163,17 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :version,
                                        short_option: "-v",
                                        env_name: "DOWNLOAD_DSYMS_VERSION",
-                                       description: "The app version for dSYMs you wish to download",
+                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs",
```
ohayon: ❤️

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -141,12 +163,17 @@ def self.available_options
...
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :build_number,
                                        short_option: "-b",
                                        env_name: "DOWNLOAD_DSYMS_BUILD_NUMBER",
                                        description: "The app build_number for dSYMs you wish to download",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :save_directory,
+                                       short_option: "-s",
+                                       env_name: "DOWNLOAD_DSYMS_SAVE_DIRECTORY",
```
ohayon: Could we change this to be called `output_directory` to match some of our other tools with similar options?

ohayon (commented):
Hey @addbrick ! Thank you so much for the PR! This is definitely a good change. I just left a couple of questions above. 

I really like the idea of having a `latest` option in there, but I'm unclear on how its being used as it still looks like it is looping over everything.

I could definitely be missing something here so let me know if I'm off base! 🚀 🐙

fastlane/lib/fastlane/actions/download_dsyms.rb
addbrick: This is used to output build number, and as a check to grab the correct build when looping through them https://github.com/fastlane/fastlane/pull/10129/files/3e9a29f87fd1df907a08115b6868e97873c18c3e#diff-65dfcad8f098c3c96688110dade15ed7L42. I just hooked into the existing loop instead of short-circuiting

fastlane/lib/fastlane/actions/download_dsyms.rb
addbrick: updated 👍

addbrick:
Hey @ohayon thanks for the review. 

The `latest` option does still loop through all the train versions and builds for that train. It just hooks into the existing code by overriding the existing `version` and `build_number` options. 

I looked into short-circuiting when passing in latest, but could not find a function which would give me a `build` object by passing in a version and build_number (eg. `app.build(version, build_number)`). All I could find were `all_build_train_numbers` and `all_builds_for_train` which still require looping and comparing strings. If you can point me to the correct method I can easily short-circuit.

fastlane/lib/fastlane/actions/download_dsyms.rb
addbrick: 🐘

addbrick:
and now I pushed the rename of `save_directory` to the `output_directory`

ohayon (approved):
👍

ohayon:
Thanks @addbrick !! Welcome to the _fastlane_ community ! 🚀

addbrick:
Thank you @ohayon! When should I expect this to be released. I know fastlane does a lot of releases, but not sure what the timeline is for them.

ohayon:
Hey @addbrick, I'm not totally sure when our next release will be, but likely sometime this week or early next week at the latest! 🐙 🚀

Issue #10136, opened 2017-08-24
Supply now requires 0.13.0 of google-api-client, but doesn't enforce this


tmtrademarked:
### New Issue Checklist

- [ X ] Updated fastlane to the latest version
- [ X ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ X ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ X ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Supply appears to be broken with the latest release. It looks like https://github.com/fastlane/fastlane/pull/10051 allowed the google-api-client dependency to be dynamic (>= 0.12.0, < 1.0.0) and https://github.com/fastlane/fastlane/pull/10120 changed supply to use APIs from the 0.13.0 version of the gem.

The google-api-client is NOT API stable. Every time they upgrade the minor build version, the API can potentially change. So it is *not* safe to let this dynamically upgrade. I believe the correct fix is to set to requirement for the google-api-client gem to >= 0.13.0, < 0.14.0 to pin the version to 0.13.x releases.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

[INSERT OUTPUT HERE]
+------+---------------------------------------------+-------------+
|                         fastlane summary                         |
+------+---------------------------------------------+-------------+
| Step | Action                                      | Time (in s) |
+------+---------------------------------------------+-------------+
| 1    | update_fastlane                             | 0           |
| 2    | Verifying required fastlane version         | 0           |
| 3    | default_platform                            | 0           |
| 4    | Switch to android deploy_alpha_mapping lane | 0           |
| 5    | gradle                                      | 326         |
| 💥   | supply                                      | 0           |
+------+---------------------------------------------+-------------+

[10:44:28]: fastlane finished with errors

Looking for related GitHub issues on fastlane/fastlane...

Found no similar issues. To create a new issue, please visit:
https://github.com/fastlane/fastlane/issues/new
Run `fastlane env` to append the fastlane environment to your issue
[...]

PR #10137, opened 2017-08-24
Require 0.13.1+ of the google-api-client

tmtrademarked:
### Checklist
- [ X ] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [ X ] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [ X ] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ X ] I've updated the documentation if necessary.

### Motivation and Context

Supply was recently updated to require the 0.13.0 branch of the google-api-client gem, but the gemspec didn't update to enforce this. See https://github.com/fastlane/fastlane/issues/10136 for more.

### Description

This PR pins the google-api-client version to the 0.13.x version of the gem to prevent drift, along with adding comments which should help to reduce inadvertent churn here. It also corrects what appears to be a missed name in one of the supply APIs (deleteall_image doesn't appear to exist, and should be delete_all_images instead)


mpirri (approved):
This is a great catch; thanks for explaining it so well in the comment, as well.  🚀

Issue #10103, opened 2017-08-22
[match] force_for_new_devices always recreates App Store profiles


Boerworz:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

We use `match` on our CI server when making AdHoc and App Store builds, and specify `force_for_new_devices: true` for both of these types since we figured we'd always want to recreate the profiles when there were new devices in the developer portal. However, we recently noticed that the App Store provisioning profile was _always_ recreated, bloating our match repository.

After looking through the source code for `match`, we realized that this is due to how `match` checks if there are new devices in the developer portal. What it does is that it retrieves the (App Store) provisioning profile from the developer portal and compares the number of devices in the profile with the total number of devices in the developer portal. This works well for Development and Ad Hoc provisioning profiles, but since App Store provisioning profiles don't contain any devices, the [`device_count_different?`](https://github.com/fastlane/fastlane/blob/master/match/lib/match/runner.rb#L180) method always returns true (if there are more than 0 devices in the developer portal, which we can probably safely assume 🙂).

We can avoid this issue by specifying `false` for `force_for_new_devices`, but since the documentation for `force_for_new_devices` states that the provisioning profile should be renewed if _the device count on the developer portal has changed_, we believe this is a bug. We propose that the `force_for_new_devices` flag be ignored when invoking `match` for App Store profiles. If you want to force the App Store profiles to be renewed, the `force` flag should be used.

If this is something you agree with, I'd be happy to create a pull request that implements this change.

Boerworz:
Oops, I pressed Enter before I was done writing the issue 😬 I'm editing it now.

mpirri:
This sounds like an amazing contribution - we would love to review a PR to add this functionality! Thanks so much for reaching out! 🚀

Boerworz:
Cool, I'll get started on a PR! 😃

JagCesar:
This has been fixed in https://github.com/fastlane/fastlane/pull/10141. Kill it with fire 🔥🚒

KrauseFx:
Awesome, thanks for confirming 👍

PR #10141, opened 2017-08-24
[match] Ignore `force_for_new_devices` for App Store profiles

Boerworz:
### Checklist

### Motivation and Context
This pull request implements the change detailed in issue #10103.

I've tested my changes by running `fastlane match appstore --force_for_new_devices true --app_identifier "com.coolcompany.foo"` and verified both that the App Store provisioning profile is not renewed and that the new warning message is displayed.

If the _fastlane_ maintainers feel that this should be covered by an `rspec` test, I would appreciate some help in writing one (having no prior experience with `rspec`) 🙂

### Description
This pull request changes the behaviour of Match to ignore `force_for_new_devices` when running Match for App Store profiles. Before this change, invoking Match for `appstore` profiles with `force_for_new_devices` set would _always_ renew the App Store provisioning profiles, even if there were no new devices on the developer portal.

Even if there would have been new devices on the developer portal, renewing the App Store provisioning profile would be unnecessary since they don't contain any device information.

 Since the App Store profiles don’t contain any devices, the `device_count_different?` method would always return `true` (if there were at least 1 device on the developer portal), resulting in the observed behaviour. This method is no longer invoked for App Store profiles.

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -120,7 +120,14 @@ module Match
       profile = profiles.last
 
       if params[:force_for_new_devices] && !params[:readonly]
-        params[:force] = device_count_different?(profile: profile) unless params[:force]
+        if prov_type != :appstore
+          params[:force] = device_count_different?(profile: profile) unless params[:force]
+        else
+          # App Store provisioning profiles don't contain device identifiers and
+          # thus shouldn't be renewed if the device count has changed.
+          UI.important "Warning: `force_for_new_devices` is set but is ignored for App Store provisioning profiles."
+          UI.important "You can safely stop specifying `force_for_new_devices` when running Match for type 'appstore'."
+        end
       end
 
       if profile.nil? or params[:force]
--- a/match/README.md
+++ b/match/README.md
@@ -274,6 +274,8 @@ end
 
 By using the `force_for_new_devices` parameter, `match` will check if the device count has changed since the last time you ran `match`, and automatically re-generate the provisioning profile if necessary. You can also use `force: true` to re-generate the provisioning profile on each run.
 
+_**Important:** The `force_for_new_devices` parameter is ignored for App Store provisioning profiles since they don't contain any device information._
+
 If you're not using `fastlane`, you can also use the `force_for_new_devices` option from the command line:
 
 ```
--- a/match/lib/match/options.rb
+++ b/match/lib/match/options.rb
@@ -125,7 +125,7 @@ module Match
                                      optional: true),
         FastlaneCore::ConfigItem.new(key: :force_for_new_devices,
                                      env_name: "MATCH_FORCE_FOR_NEW_DEVICES",
-                                     description: "Renew the provisioning profiles if the device count on the developer portal has changed",
+                                     description: "Renew the provisioning profiles if the device count on the developer portal has changed. Ignored for profile type 'appstore'",
                                      is_string: false,
                                      default_value: false),
         FastlaneCore::ConfigItem.new(key: :skip_docs,


Boerworz:
The CLA has been signed! 🙂

match/lib/match/runner.rb
```diff
@@ -120,7 +120,14 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
       profile = profiles.last
 
       if params[:force_for_new_devices] && !params[:readonly]
-        params[:force] = device_count_different?(profile: profile) unless params[:force]
+        if prov_type != :appstore
+          params[:force] = device_count_different?(profile: profile) unless params[:force]
+        else
+          # App Store provisioning profiles don't contain device identifiers and
+          # thus shouldn't be renewed if the device count has changed.
+          UI.important "Warning: `force_for_new_devices` is set but is ignored for App Store provisioning profiles."
```
casz: Is this really necessary? Why not just silently ignore it?
Other then that LGTM.

jetersen (approved):

match/lib/match/runner.rb
ohayon: I agree that it might not be totally necessary, but I do like that it lets people know not to add no-op params to tools because if something changes down the road we will always have to be aware of that. 

That being said, I think this could be moved into a `verify_block` on the option itself and it can print the same message. 🚀

ohayon (changes requested):
This looks great @Boerworz !! Just one little note if we could move the print outs into a verify block in order to avoid the nested conditionals in there.

Thanks! We are excited to get this merged in! 🚀

match/lib/match/runner.rb
Boerworz: @ohayon: Cool, I didn't know about `verify_block`! However, I'm not entirely sure of how I can access the specified provisioning type from the block. Any pointers? 👉

Boerworz: Also, if it's moved into the `verify_block`, should it still be a `UI.important` or should it be a `UI.user_error!`? I believe making it an error is required if we want to avoid the nested conditionals (like you mentioned below), but I guess it would also make this a breaking change compared to previous versions of Match? 💣

ohayon (dismissed):
This looks great @Boerworz !! Just one little note if we could move the print outs into a verify block in order to avoid the nested conditionals in there.

Thanks! We are excited to get this merged in! 🚀

match/lib/match/runner.rb
casz: Please don't make it error! 💣 That will break our lanes for updating provision profiles.
Since then you forcing us to also care about the conditional parameter, which we shouldn't have to.

```
  lane :update do |options|
    match(
      type: options[:type],
      git_url: 'ssh://git@bitbucket.widex.com/swd/certificates.git',
      git_branch: 'widex',
      app_identifier: [
        'com.widex.app',
        'com.widex.app.watchkitapp',
        'com.widex.app.watchkitapp.watchkitextension',
        'com.widex.otherapp',
        'com.widex.otherapp.watchkitapp',
        'com.widex.otherapp.watchkitapp.watchkitextension',
        'com.widex.tinnitus'
      ]
    )
  end

  lane :updateAll do
    update(type: 'development')
    update(type: 'adhoc')
    update(type: 'appstore')
  end
```

Boerworz: @casz: Yeah, I'm with you. Making it an error feels a bit strict 🙂 Hopefully @ohayon agrees 😃

Boerworz:
@ohayon: Sorry to poke you, but do you have any thoughts on what @casz and I discussed in their [approval](https://github.com/fastlane/fastlane/pull/10141#pullrequestreview-58863481)? TL;DR: To avoid the nested conditional in `runner.rb` we'd need to use `UI.user_error!` in the `verify_block`. Is having one less nested conditional worth a breaking change? Neither @casz nor I think so 🙂

KrauseFx:
@ohayon 

> I agree that it might not be totally necessary, but I do like that it lets people know not to add no-op params to tools because if something changes down the road we will always have to be aware of that.
> That being said, I think this could be moved into a verify_block on the option itself and it can print the same message. 🚀

I don't think it can, as we can't check for other values of other `ConfigItem`s in a `verify_block`, so this part of the code should be ok to use 👍 

As we fill in the values for each config item, and then immediately call the `verify` block, it means that we can't know if the other value is already provided, so in other tools we also check for this somewhere in the `runner` code or similar :)

KrauseFx (approved):
Looks good to me now, thanks for your contribution @Boerworz 👍

PR #10190, opened 2017-08-30
Add readonly option to sigh

nicolasbraun:
### Motivation and Context

Because fastlane - especially match - does not support entitlements yet (see #9912)
We have to generate our provisioning profile by hand and then use cert + sigh to retrieve it.

For some reason those profile sometimes become invalid on iTunes Connect (still have to figure out why!). In that case we do not want sigh to generate a profile because the codesign will fail in gym and it floodes our Apple Dev account

### Description
Simply add a reasonly option to sigh and crash if it is enabled and no provisioning profile is found.

### Questions

The readonly option on match works differently, it only checks in the local repo for a matching profile. This is a bit different since it could be on the repo and invalid on Apple Dev I believe.


sigh/lib/sigh/runner.rb
```diff
@@ -30,6 +30,7 @@ def run
           profile = create_profile!
         end
       else
+        UI.crash!("No matching provisioning pofile found and can not create a new one because you enabled `readonly`") if Sigh.config[:readonly]
```
KrauseFx: Please change this to `UI.user_error!` as we don't need the full stack trace here 👍

KrauseFx (approved):
This is awesome, thanks for the fix, please update the PR to address this small thing, then we can merge and ship 🚀

sigh/lib/sigh/runner.rb
nicolasbraun: @KrauseFx thanks for the review! its done

ohayon:
This is great @nicolasbraun ! Would you mind just fixing up the merge conflict and we will be good to go! 🚀

nicolasbraun:
@KrauseFx @ohayon  it got me thinking, what is the behavior of match if the provisioning profile on the portal is invalid for some reason, does it renew it or will it create another one with the timestamp suffix?

PR #10194, opened 2017-08-30
Ignore empty password for FASTLANE_PASSWORD

kouki-dan:
### Checklist

### Motivation and Context

I found this input field when I use match in spite of storing password in Keychain.
```
-------------------------------------------------------------------------------------
Please provide your Apple Developer Program account credentials
The login information you enter will be stored in your macOS Keychain
You can also pass the password using the `FASTLANE_PASSWORD` environment variable
More information about it on GitHub: https://github.com/fastlane/fastlane/tree/master/credentials_manager
-------------------------------------------------------------------------------------
Password (for yourmail@yourdomain): 
```
I noticed that my env file includes this line `FASTLANE_PASSWORD = ''` 
If empty password exists in env file, the user cannot use password from keychain and cannot store password to keychain. I think the user wants to get the password from keychain in this case.

### Description

If FASTLANE_PASSWORD is empty in env, ignore it.

### DIFF
--- a/credentials_manager/lib/credentials_manager/account_manager.rb
+++ b/credentials_manager/lib/credentials_manager/account_manager.rb
@@ -34,7 +34,7 @@ module CredentialsManager
 
     def fetch_password_from_env
       password = ENV["FASTLANE_PASSWORD"] || ENV["DELIVER_PASSWORD"]
-      password unless password.to_s.length == 0
+      return password unless password.to_s.length == 0
     end
 
     def password(ask_if_missing: true)
--- a/credentials_manager/lib/credentials_manager/account_manager.rb
+++ b/credentials_manager/lib/credentials_manager/account_manager.rb
@@ -33,7 +33,8 @@ module CredentialsManager
     end
 
     def fetch_password_from_env
-      ENV["FASTLANE_PASSWORD"] || ENV["DELIVER_PASSWORD"]
+      password = ENV["FASTLANE_PASSWORD"] || ENV["DELIVER_PASSWORD"]
+      password unless password.to_s.length == 0
     end
 
     def password(ask_if_missing: true)
--- a/credentials_manager/spec/account_manager_spec.rb
+++ b/credentials_manager/spec/account_manager_spec.rb
@@ -52,6 +52,20 @@ describe CredentialsManager do
       ENV.delete('FASTLANE_USER')
     end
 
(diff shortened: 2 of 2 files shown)


ohayon:
Hey @kouki-dan! Thanks for the PR! I am curious about one thing. Is there a reason why you wouldn't remove that environment variable all together instead of setting it to be an empty string? 🐙

credentials_manager/lib/credentials_manager/account_manager.rb
```diff
@@ -33,7 +33,8 @@ def user
     end
 
     def fetch_password_from_env
-      ENV["FASTLANE_PASSWORD"] || ENV["DELIVER_PASSWORD"]
+      password = ENV["FASTLANE_PASSWORD"] || ENV["DELIVER_PASSWORD"]
+      password unless password.to_s.length == 0
```
KrauseFx: Please add return statement here to be more explicit. Also it would be good to have an explicit `nil` return also, just to make sure it's clear for future developers

KrauseFx:
@ohayon Ohhh, interesting idea. What are the positive effects removing the env variable for the current process would achieve? Does it have any side effects also?

kouki-dan:
That's right. I also remove this environment variable in my .env file. 

However an apple accounts' password must not be empty. If empty password is set to env, I think it may be mistakes.
I got the mistake and got the error of description. I type a my password and get a below error in this case.
```
Password (for xxxxx@xxxxxx): ********
security: SecKeychainAddInternetPassword <NULL>: The specified item already exists in the keychain.
Could not store password in keychain
```
I cannot use keychain's password and cannot reuse the password from keychain although it may be mistakes.

credentials_manager/lib/credentials_manager/account_manager.rb
kouki-dan: Do you mean like this?
```
if password.to_s.length == 0
    return nil
else
    return password
end
```

ohayon:
Ah, okay. @kouki-dan we can definitely merge this in, but yeah, if you could just add a return statement like @KrauseFx mentioned, maybe just `return password unless password.to_s.length == 0` that would be 💯 !

kouki-dan:
Thank you for your review and I pushed a fixed commit!

KrauseFx (approved):
Beautiful, thanks for your contribution @kouki-dan and updating the PR 👍

PR #10245, opened 2017-09-03
Make Carthage action support dependencies with build command

joihelgi:
### Checklist

### Motivation and Context

With Carthage you can specify which dependencies should be built, e.g. you can run `carthage build Alamofire` and it will only build Alamofire and it's dependencies. The Carthage action ignores the `dependencies` parameter if the command is not `update`. This pr adds support for dependencies with the build command.

I used the same tests as for dependencies with the update command, only changed the command to `build` instead of `update`.


KrauseFx (approved):
Awesome, that's great, thanks for your contribution @joihelgi 👍

Issue #10247, opened 2017-09-04
Pilot uploaded a copied ipa with hash name, not original one


Sulfkain:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I see that on iTunes, the name of the uploaded ipa was a hash, not the original one, which when uploaded manually was correct.

![captura de pantalla 2017-09-04 a las 12 58 26](https://user-images.githubusercontent.com/17082246/30028956-77448788-9187-11e7-94d1-1cb64953413a.png)

I have found that this is the blame part:
https://github.com/fastlane/fastlane/blob/60bac7e4a2a6c1e19d1cd20eb83201e7d0cb9e48/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb#L37

I have modified this 2 lines with:
```
      ipa_file_name = File.basename(ipa_path)
      resulting_path = File.join(self.package_path, "#{ipa_file_name}")
```

With this I achive to get the right name on iTunes connect.
I need this field to some step on CI, so it's important.

The question is:
Why you need to copy the ipa? Why don't do everything on the original ipa? 
If it's needed, would be much better to set the same name or a suffix/prefix on the name that a hash name, which is useless for us. Don't you think?

I will open a Pull Request with this changes, if you agree, merge it, if not, please say how we can get the right name on iTunes connect.

Thanks!

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

Doesn't matter the env. I have the last version, and I'm looking on the master branch on repo.

Sulfkain:
This will be solved on the new version

PR #10248, opened 2017-09-04
Set the original name for the uploaded ipa

Sulfkain:
### Checklist

### Motivation and Context

With this change you get on iTunes connect the right ipa name, not a unusable hash

Issue related: 
https://github.com/fastlane/fastlane/issues/10247

### DIFF
--- a/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
+++ b/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
@@ -34,8 +34,8 @@ module FastlaneCore
     private
 
     def copy_ipa(ipa_path)
-      ipa_file_name = Digest::MD5.hexdigest(ipa_path)
-      resulting_path = File.join(self.package_path, "#{ipa_file_name}.ipa")
+      ipa_file_name = File.basename(ipa_path) 
+      resulting_path = File.join(self.package_path, "#{ipa_file_name}")
       FileUtils.cp(ipa_path, resulting_path)
 
       return resulting_path
--- a/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
+++ b/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
@@ -34,8 +34,8 @@ module FastlaneCore
     private
 
     def copy_ipa(ipa_path)
-      ipa_file_name = File.basename(ipa_path) 
-      resulting_path = File.join(self.package_path, "#{ipa_file_name}")
+      ipa_file_name = File.basename(ipa_path)
+      resulting_path = File.join(self.package_path, ipa_file_name)
       FileUtils.cp(ipa_path, resulting_path)
 
       return resulting_path
--- a/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
+++ b/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
@@ -34,7 +34,7 @@ module FastlaneCore
     private
 
     def copy_ipa(ipa_path)
-      ipa_file_name = "#{File.basename(ipa_path, '.ipa')}_#{Digest::SHA256.file(ipa_path).hexdigest}.ipa"
+      ipa_file_name = "#{File.basename(ipa_path, '.ipa')}(#{Digest::SHA256.file(ipa_path).hexdigest}).ipa"
       resulting_path = File.join(self.package_path, ipa_file_name)
       FileUtils.cp(ipa_path, resulting_path)
 
--- a/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
+++ b/fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
@@ -34,7 +34,7 @@ module FastlaneCore
     private
 
     def copy_ipa(ipa_path)
-      ipa_file_name = File.basename(ipa_path)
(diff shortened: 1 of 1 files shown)


Sulfkain:
Seems the spaces breaks the tests ¿?, I have tried to replace my tabs with the same "tabs" from original file, but seems no change on file and I can't commite it.
What can we do about that?

Sulfkain:
I'm sure I have accepted CLA

fastlane_core/lib/fastlane_core/ipa_upload_package_builder.rb
```diff
@@ -34,8 +34,8 @@ def generate(app_id: nil, ipa_path: nil, package_path: nil, platform: nil)
     private
 
     def copy_ipa(ipa_path)
-      ipa_file_name = Digest::MD5.hexdigest(ipa_path)
-      resulting_path = File.join(self.package_path, "#{ipa_file_name}.ipa")
+      ipa_file_name = File.basename(ipa_path)
```
KrauseFx: Thanks for the PR, I like the idea, however I think there was a reason we have the ipa file's name be a hash, to avoid conflicts. What's the reason you need the original file name on iTC? Can you use a workaround maybe?

Sulfkain: Yes, we create 2 types of ipas, and in Itunnes, the only way to know what kind of ipa is, is looking the fileName.
There is no other reference on iTunes related with the file, nor tag to mark this compilation or so.

I was looking and the ipa is copied on a temporary directory (with Dir.mktmpdir).
I'm check and consecutive calls to this method produce, 2 different tmp directories
Ex:
 /var/folders/p6/dh7nfy296777vf7xf37kstq80000gp/T/d20170905-14814-1i7s8ir/var/folders/p6/dh7nfy296777vf7xf37kstq80000gp/T/d20170905-14814-1ramowk

So the colission is impossible.
Anyway, if you are not sure about this, suffixing the name of ipa with that hash do the job and we still getting the original name on iTunes connect, wich is much better than nothing.
I think is better to get the original name only ( for other developers), but this could be a workaround, for us is not a problem getting a hash as suffix.

Thanks!

Sulfkain: I just realized, that even with this md5 digest, it's not a random one, if you have a file with the same name, and same path, the md5 will be the same.
How this could avoid a colission of 2 ipas with the same name?

Sulfkain:
Hi @KrauseFx, any new about my comments?
We are waiting to the approvement of this PR to continue with our deployment scripts.
Is everything ok?

ohayon:
Hey @Sulfkain - apologies for jumping in here without much context, but if it is collisions that we are worried about, could we hash the file at the path, that way it is guaranteed to be unique and then append the file name? Let me know if that is a workable solution? Thanks! 🐙 🚀

Sulfkain:
Hey @ohayon, yes this will be a solution of us as I commented before on this thread.
But I was wondering, a MD5 Hash of a file, will not avoid a collision, it's a hash, not a random.
Please read the comment that I explain the problem/solution (because I prefer don't mess this thread with copy/paste).
I hope this will be great for you.

Thanks!

ohayon:
Hey @Sulfkain, sorry I dont think I was very clear. We could maybe use the SHA of the file, that way it is different if they are different binaries right? Sorry if I am still missing some context here! 🐒

Sulfkain:
Aha! Ok, that will avoid collisions and will be enought for us.
Anyway as I told, when you create a temporary directory, you will get different directories so seems collision is impossible.
But if you prefer double factor, for us is not a problem getting a hash in suffix name of IPA.

I modify my PR or do you do the change?

Thanks!

ohayon (approved):
Thanks for being responsive to that feedback and making those changes @Sulfkain !! 🚀

Sulfkain:
Perfect 🤜🏻 🤛🏻

Thanks @ohayon

mithiljadhav:
Adding the original name for the uploaded ipa causes the following transporter error **if the file name contains spaces**:

    ERROR ITMS-3000: "Line 7 column 111: character content of element "file_name" invalid; must be a 
    string matching the regular expression "[^/: ]+" at XPath 
    /package/software_assets/asset/data_file/file_name"

I think the spaces should be replaced with underscore.

My ipa file has a space in it "XYZ abc.ipa", so it is failing  after updating fastlane to 2.57.0+. Right now the workaround is to replace the file name to "XYZ_abc.ipa" and run deliver/pilot again.

I think this needs to be fixed since it is a regression.

Sulfkain:
I'm so sorry , I didn't want to break fastlane :'(.
My first thought (and commit) was to do with '_' but I thought would be better for readability a kit kat, but I don't test that last change on iTC, sorry again.

KrauseFx:
@Sulfkain No worries, this happens to everybody - we were able to react quickly :)

Sulfkain:
There is no way to pass the name as originally this was openned?? 
Only a hash of a ipa it's useless for us :/

Sulfkain:
@KrauseFx Hi, any answer about this? 
Maybe some middle solution between the actual state and what I Pull requested.
If not pass some regexp (which Apple has), then fall to the actual method, else do the name+hash.

Even so, I think the actual md5 digest it's wrong because it's gets the hash from the path, but mine get the hash from the content of the file itself (almost impossible to collide).

PR #10252, opened 2017-09-05
Support for updating roles on iTunes Connect members

MortenGregersen:
### Checklist

### Motivation and Context
Spaceship has support for creating new members in iTunes Connect, but once they are created, you can change the roles. We want to use fastlane/spaceship to set the 'App Manager' role on some members on multiple accounts temporary and revoke it afterwards.
I've tested this with my local branch as a local gem in Gemfile. I've changed the roles on our members from 'Developer' and 'Finance' to 'App Manager' and 'Finance' and back.

### Description
This change adds support for updating the roles of a member. The given member object isn't updated. A new request to get the member is required to get an updated list of roles.


MortenGregersen:
I signed it!

KrauseFx:
That's awesome, thanks for the PR. Should we add some docs also to `DeveloperPortal.md`? Also it seems like the Google CLA bot isn't happy, because the git commit seems to not be associated with your username? Could you confirm that the code change is from you, and that you agree with the CLA? :) 
Thanks

MortenGregersen:
D'oh! I used the wrong mail address when I made the commit. I don't want to have this merged with this mail address. I will create a new pull request with the same commit but with my github mail address.

I will reference it here when the pull request is ready.

KrauseFx:
Alright, perfect, thanks 👍

MortenGregersen:
I've created a new pull request with the right mail address: #10257

PR #10257, opened 2017-09-05
Support for updating roles on iTunes Connect members

MortenGregersen:
### Checklist

### Motivation and Context
Spaceship has support for creating new members in iTunes Connect, but once they are created, you can change the roles. We want to use fastlane/spaceship to set the 'App Manager' role on some members on multiple accounts temporary and revoke it afterwards.
I've tested this with my local branch as a local gem in Gemfile. I've changed the roles on our members from 'Developer' and 'Finance' to 'App Manager' and 'Finance' and back.

### Description
This change adds support for updating the roles of a member. The given member object isn't updated. A new request to get the member is required to get an updated list of roles.


KrauseFx (approved):
Awesome, thanks ❤

PR #10267, opened 2017-09-06
Deliver: Add supply's `android` folder to list of excluded directories for language check

### LINKED ISSUE #10265: [deliver] unsupported languages check breaks for people also using `supply` at the same time because of `android` folder
In #9351 a check for unsupported languages in `fastlane/metadata` was added. 
In #10122 an additional option to ignore this check was added because the feature broke custom file and folder structures.

Unfortunately it also breaks when using both `supply` and `deliver` in the same project (as e.g. multi platform projects built with Cordova or Ionic might want to do). Using `supply` to download Android metadata creates an `metadata/android` folder that now causes this error when using `deliver` again:

```
[!] Unsupported directory name(s) for screenshots/metadata in './fastlane/metadata': android
Valid directory names are: ["da", "de-DE", "el", "en-AU", "en-CA", "en-GB", "en-US", "es-ES", "es-MX", "fi", "fr-CA", "fr-FR", "id", "it", "ja", "ko", "ms", "nl-NL", "no", "pt-BR", "pt-PT", "ru", "sv", "th", "tr", "vi", "zh-Hans", "zh-Hant", "appleTV", "iMessage", "default"]

Enable 'ignore_language_directory_validation' to prevent this validation from happening
```

So now one can only use the 'ignore_language_directory_validation' option and lose the check completely.

I suggest that the `android` folder is added to a list of exceptions that are allowed to exist next to the language folders as it is a default folder when using only fastlane tools.

(Of course this is only because of the historical folder structure where Android is smashed into the existing folder structure - but as this won't change any time soon, we somehow have to handle this now)

janpio:
Currently the check for invalid languages also triggers when using `supply` as it creates an "android" folder that this check doesn't like. This adds this folder to the EXCEPTION_DIRECTORIES list so this doesn't happen any more

fixes #10265

(Please note that I have no ruby experience at all - bear with me if my suggested changes are not perfect. Feel free to fix everything I messed up, or let me know so I can)

### DIFF
--- a/deliver/lib/deliver/loader.rb
+++ b/deliver/lib/deliver/loader.rb
@@ -13,7 +13,7 @@ module Deliver
     # If the user also uses `supply` in the same project, an 'android' folder might exist
     SUPPLY_DIR_NAME = "android".freeze
 
-    EXCEPTION_DIRECTORIES = (UploadMetadata::ALL_META_SUB_DIRS.map(&:downcase) + SUPPLY_DIR_NAME).freeze
+    EXCEPTION_DIRECTORIES = (UploadMetadata::ALL_META_SUB_DIRS.map(&:downcase) << SUPPLY_DIR_NAME).freeze
 
     def self.language_folders(root, ignore_validation)
       folders = Dir.glob(File.join(root, '*'))
--- a/deliver/lib/deliver/loader.rb
+++ b/deliver/lib/deliver/loader.rb
@@ -10,7 +10,11 @@ module Deliver
 
     SPECIAL_DIR_NAMES = [APPLE_TV_DIR_NAME, IMESSAGE_DIR_NAME, DEFAULT_DIR_NAME].freeze
 
-    EXCEPTION_DIRECTORIES = UploadMetadata::ALL_META_SUB_DIRS.map(&:downcase).freeze
+    # If the user also uses `supply` in the same project, an 'android' folder might exist
+    SUPPLY_DIR_NAME = "android".freeze
+
+    EXCEPTION_DIRECTORIES = (UploadMetadata::ALL_META_SUB_DIRS.map(&:downcase) + SUPPLY_DIR_NAME).freeze
+
 
     def self.language_folders(root, ignore_validation)
       folders = Dir.glob(File.join(root, '*'))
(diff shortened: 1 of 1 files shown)


janpio:
I saw the [tests](https://github.com/fastlane/fastlane/blob/e5e394918641ec735bcb933066549c5380f34647/deliver/spec/loader_spec.rb) but am not really sure how I should add this. Suggestions? Or doesn't this have to be tested anyway?

KrauseFx (approved):
Awesome, nice catch, thanks for your contribution. Can you send me a short email so I have your email address? :) It's on my profile

KrauseFx:
Oh, the tests are failing 😢

janpio:
CircleCI didn't actually load for me yesterday, now it does. I think I can fix that.

KrauseFx:
Whops :) also don't forget sending me an email

janpio:
From robothingie to real test failure - I will have to debug this it seems and actually see if I can run it locally. Will report back.

KrauseFx:
Awesome, thanks @janpio 👍

janpio:
Ok.  So I installed RVM and a new ruby, then `gem install bundler`, then `bundle install` in my checkout. Then `bundle exec fastlane test`which ran tests and gave me this error:

```
/Users/sujan/fastlane/deliver/lib/deliver/loader.rb:16:in `<module:Loader>': no implicit conversion of String into Array (TypeError)
```

which makes sense - seems I am using the wrong "method" here. (And turns out `bundle exec rspec deliver` is a much faster way to run the same failing tests)

Some googling later I know [how to actually add something to an array in ruby](https://ruby-doc.org/core-2.2.0/Array.html#3C-3C-method) (`<<`, not `+` which combines two arrays).

`bundle exec rspec deliver` and `bundle exec fastlane test` (Oh my, this is even slower when it doesn't fail with broken tests right at the beginning...) again to confirm the tests actually pass now - they do!

So commit the changes to the branch so it gets added here. Hope CircleCI agrees.

Did I miss anything a real ruby dev would have done?

janpio:
(Can I still rename my branch or will this break this PR?)

janpio:
Ok, followed the documentation on testing changes locally: https://github.com/janpio/fastlane/blob/master/YourFirstPR.md#test-the-changes-for-your-application
And it actually worked :)

So, good to merge I think.

ohayon:
Thanks @janpio !!! 🚀

PR #10272, opened 2017-09-06
Improve mismatch message between export_method and selected profiles

taisukeh:
### Checklist

### Motivation and Context

Unfortunately our organization uses a **enterprise** profile named `MyOrg InHouse AdHoc`, then the following message is displayed every time after a build error. 

```
ERROR [2017-09-06 20:51:16.17]: There seems to be a mismatch between your provided `export_method` in gym
ERROR [2017-09-06 20:51:16.17]: and the selected provisioning profiles. You passed the following options:
WARN [2017-09-06 20:51:16.18]:   export_method:      enterprise
WARN [2017-09-06 20:51:16.18]:   Bundle identifier:  my.org.app
WARN [2017-09-06 20:51:16.18]:   Profile name:       MyOrg InHouse AdHoc
WARN [2017-09-06 20:51:16.18]:   Profile type:       ad-hoc
ERROR [2017-09-06 20:51:16.18]: Make sure to either change the `export_method` passed from your Fastfile or CLI
ERROR [2017-09-06 20:51:16.18]: or select the correct provisioning profiles by updating your Xcode project
ERROR [2017-09-06 20:51:16.18]: or passing the profiles to use by using match or manually via the `export_options` hash
```

The printed profile type is wrong, and printed regardless of the cause of the error. 😞

### Description

1. Add `inhouse` key to available_export_types
  - Probably, the enterprise profiles are often named `inhouse`.
2. The enterprise export type take precedence over `ad-hoc`
3. Print message only once
4. Once it matches, don't check subsequent items

I think that this change has little negative impact on other people.

### DIFF
--- a/gym/lib/gym/error_handler.rb
+++ b/gym/lib/gym/error_handler.rb
@@ -237,10 +237,13 @@ module Gym
             "app-store" => "app-store",
             "app store" => "app-store",
             "appstore" => "app-store",
+            "enterprise" => "enterprise",
+            "in-house" => "enterprise",
+            "in house" => "enterprise",
+            "inhouse" => "enterprise",
             "ad-hoc" => "ad-hoc",
             "adhoc" => "ad-hoc",
             "ad hoc" => "ad-hoc",
-            "enterprise" => "enterprise",
             "development" => "development"
           }
 
@@ -259,7 +262,7 @@ module Gym
               # As seen above, there is obviously a mismatch, the user selected an App Store
               # profile, but the export method that's being passed to Xcode is "enterprise"
 
-              next if matching_type.to_s == selected_export_method
+              break if matching_type.to_s == selected_export_method
               UI.message("")
               UI.error("There seems to be a mismatch between your provided `export_method` in gym")
               UI.error("and the selected provisioning profiles. You passed the following options:")
@@ -271,6 +274,7 @@ module Gym
               UI.error("or select the correct provisioning profiles by updating your Xcode project")
               UI.error("or passing the profiles to use by using match or manually via the `export_options` hash")
               UI.message("")
+              break
             end
           end
         end
--- a/gym/spec/error_handler_spec.rb
+++ b/gym/spec/error_handler_spec.rb
@@ -54,5 +54,35 @@ Code signing is required for product type 'Application' in SDK 'iOS 11.0'
 
       Gym::ErrorHandler.handle_build_error(code_signing_output)
     end
+
+    it "prints mismatch between the export_method and the selected profiles only once" do
+      mock_gym_path(@output)
+      expect(UI).to receive(:build_failure!).with("Error building the application - see the log above", error_info: @output)
+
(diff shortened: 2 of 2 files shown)


gym/lib/gym/error_handler.rb
```diff
@@ -259,7 +262,7 @@ def print_environment_information
               # As seen above, there is obviously a mismatch, the user selected an App Store
               # profile, but the export method that's being passed to Xcode is "enterprise"
 
-              next if matching_type.to_s == selected_export_method
+              break if matching_type.to_s == selected_export_method
```
KrauseFx: I think this should be `next` and not `break`, right?

gym/lib/gym/error_handler.rb
```diff
@@ -237,10 +237,13 @@ def print_environment_information
             "app-store" => "app-store",
             "app store" => "app-store",
             "appstore" => "app-store",
+            "enterprise" => "enterprise",
+            "in-house" => "enterprise",
+            "in house" => "enterprise",
+            "inhouse" => "enterprise",
```
KrauseFx: Nice 👍

KrauseFx:
Thanks for your contribution @taisukeh, I added 2 comments 👍

gym/lib/gym/error_handler.rb
taisukeh: The `next` or `break` makes sense when the profile name matches two different export types.

When profile name is `Ad-Hoc Development` and export_method is `ad-hoc`, the printed messages are as follows.

#### `next`

```
[02:32:17]: There seems to be a mismatch between your provided `export_method` in gym
[02:32:17]: and the selected provisioning profiles. You passed the following options:
[02:32:17]:   export_method:      ad-hoc
[02:32:17]:   Bundle identifier:  my.org.app
[02:32:17]:   Profile name:       Ad-Hoc development
[02:32:17]:   Profile type:       development
````

1. `Ad-Hoc` matches `ad-hoc` and is equal to the export method, then `next` to check subsequent types
2. `Development` matches `development`, then print the message

#### `break`

The mismatch message does not printed.

1. `Ad-Hoc` matches `ad-hoc` and is equal to the export method, then `break`

----
I think that there is a high possibility that the detected profile type after the `next` is incorrect, so I changed it to `break`. But of course, I'll fix it to `next` if you think it is better to print the message.

taisukeh:
@KrauseFx Thanks for your review.
I added a reply https://github.com/fastlane/fastlane/pull/10272#discussion_r137343839, please check when you're available.

ohayon:
Hey @taisukeh, apologies for the delay getting back to you here, do you think you'd still want to get this merged in? We have made a lot of changes to the way code signing works with all the changes since Xcode 9. Is this still necessary?

KrauseFx (approved):
Sorry for the delay, seems like you have a good understanding of this - let's give it a try 👍  Thanks for your contribution 🚀

PR #10275, opened 2017-09-06
Add details to error when Apple services are down

raylillywhite:
### Checklist

### Motivation and Context
When Apple developer services go down "Apple 302 detected" doesn't explain that it's a temporary error, and leaves the user wondering how they can fix it, instead of recognizing that they just have to wait.

### Description

I tried to word the message in a way that suggests to check if any Apple services are down without guaranteeing that that's the cause, since I'm not sure if there are other reasons this error might be encountered.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -686,7 +686,7 @@ module Spaceship
         end
 
         if response.body.to_s.include?("<title>302 Found</title>")
-          raise AppleTimeoutError.new, "Apple 302 detected"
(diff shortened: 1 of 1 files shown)


raylillywhite:
I signed it!

KrauseFx:
Hey @raylillywhite, thanks for the PR, I love the idea. Maybe instead of 
```
Apple 302 detected. Check https://developer.apple.com/system-status/ to see if Apple services are down
```
we could phrase it differently, as the status page is not always up to date
```
Apple 302 detected - this might be temporary server error, check https://developer.apple.com/system-status/ to see if there is a known downtime
```

What do you think?

raylillywhite:
Yes, that message is much better :) I just pushed the updated message

KrauseFx (approved):
Awesome, thanks for updating, and submitting the initial PR 👍

PR #10278, opened 2017-09-07
Make curl command fail when wwdc certificate fails to download

mcfedr:
### Checklist
- [N/A] I've updated the documentation if necessary.

### Motivation and Context

My builds are randomly failing because WWDR certificate is failing to install. It appears that its failing to download, but the `curl` command doesn't report this by default, and so `security` tries to import the file, that presumably has some html error message in it.

This is my current log output
```
[11:32:51]: Installing certificate...
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   710  100   710    0     0    627      0  0:00:01  0:00:01 --:--:--   627
security: SecKeychainItemImport: Unknown format in import.
```

### Description

Adding `-f` makes `curl` exit code non-zero for non 200 http response codes. It also causes a message to be printed with the issue. (Something like `curl: (22) The requested URL returned error: 404 Not Found`). This should make debugging more clear and prevent `security` trying to import and invalid cert.

### DIFF
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -61,7 +61,7 @@ module FastlaneCore
       filename = file.path
       keychain = wwdr_keychain
       keychain = "-k #{keychain.shellescape}" unless keychain.empty?
-      Helper.backticks("curl -o #{filename} #{url} && security import #{filename} #{keychain}", print: FastlaneCore::Globals.verbose?)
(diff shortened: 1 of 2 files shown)


KrauseFx (approved):
Awesome, nice catch, thanks for the PR 👍

Issue #8973, opened 2017-04-23
PROVISIONING_PROFILE_SPECIFIER and PROVISIONING_PROFILE changes in Xcode 8.3 causing white labelled app to fail to sign


jdouglas-nz:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
Since the migration from XCode 8.2.1 -> 8.3, our white labelled app (which we parameterize app identifier and app name among other things) now fails to sign. 

The relevant piece of our _Fastfile_ is this: 

``` 
        # ensure there is a private key and signing cert available for the app.
        match(
            app_identifier: newAppId,
            git_url: 'ssh://some/path/to/git/repo.git',
            type: 'appstore',
            username: '[redacted]'
        )

        # ensure there is a provisioning profile available to build and sign the app with
        sigh(
            app_identifier: newAppId,
            provisioning_name: "match AppStore #{newAppId}"
        )

        # update app identifier so can sign with newly obtained provisioning profile
        update_app_identifier(
            xcodeproj: Paths::PROJECT_PATH,
            plist_path: Paths::PLIST_FILE_PATH,
            app_identifier: newAppId
        )

        # set the project provisoning profile UUID to the mobile provisioning profile obtained from sigh
        update_project_provisioning(
            xcodeproj: Paths::PROJECT_PATH,
            target_filter: '.*',
            profile: "./AppStore_#{newAppId}.mobileprovision"
        )

        # finally, build app
        gym(
            scheme: Schemes::APP_SCHEME,
            workspace: Paths::WORKSPACE_PATH
        )
```

after getting to the end of gym, and starting to sign, there are a bunch of these spat out:

> ❌  Code signing is required for product type 'Application' in SDK 'iOS 10.3'

XCode 8.3 and Apple's related build tooling now seems to use **BOTH** the `PROVISIONING_PROFILE_SPECIFIER` and `PROVISIONING_PROFILE` **TOGETHER** to determine if a particular provisioning profile can be used. 

We have confirmed this by after running the script up until `gym`, we have manually changed the `PROVISIONING_PROFILE_SPECIFIER` as well. After this change, code signing works as it did. 

[...]

PR #10298, opened 2017-09-10
Fix 'update_project_provisioning' to set PROVISIONING_PROFILE_SPECIFIER

col:
Update the 'update_project_provisioning' action to set PROVISIONING_PROFILE_SPECIFIER (Name) rather than just the PROVISIONING_PROFILE (UUID)

### Checklist

### Motivation and Context

Xcode 8 uses PROVISIONING_PROFILE_SPECIFIER in preference over PROVISIONING_PROFILE but this action is only setting PROVISIONING_PROFILE. This change adds better support for the Xcode 8+

[https://github.com/fastlane/fastlane/issues/8973](https://github.com/fastlane/fastlane/issues/8973)

Manually tested using both a small test app and a full scale production app. 
I looked into writing a unit test for this change but the existing behavior (setting the PROVISIONING_PROFILE) is not unit tested and it was difficult to add a specific unit test for this change.

### Description

Set PROVISIONING_PROFILE_SPECIFIER so that the correct provisioning profile is used when building with Xcode 8+.

### DIFF
--- a/fastlane/lib/fastlane/actions/update_project_provisioning.rb
+++ b/fastlane/lib/fastlane/actions/update_project_provisioning.rb
@@ -64,6 +64,7 @@ module Fastlane
             end
 
             build_configuration.build_settings["PROVISIONING_PROFILE"] = data["UUID"]
+            build_configuration.build_settings["PROVISIONING_PROFILE_SPECIFIER"] = data["Name"]
           end
         end
 


col:
I signed it!

col:
I signed it!

col:
I signed it!

KrauseFx (approved):
Nice change, thanks for your contribution @col 👍

col:
Thanks for merging that so fast. 

Would you mind also giving me your thoughts on the new action I suggested [here](https://github.com/fastlane/fastlane/issues/8973). See: `update_code_signing_options`

If you think that would be a viable change I'm happy to put another PR together for it.

PR #10307, opened 2017-09-11
Fix, action calld by other_action not run in the root of the project

taisukeh:
### Checklist

### Motivation and Context

The current directory for the action called from `other_action` is incorrect.

For example.

`print_pwd_wrapper` action
```
    def self.run(params)
      other_action.print_pwd()
    end
```

`print_pwd` action
```
    def self.run(params)
      UI.message "pwd: #{Dir.pwd}"
    end
```

Then call them.
```
% cat fastlane/Fastfile
lane :test do
  print_pwd()
  print_pwd_wrapper()
end

% fastlane test
[23:39:19]: -----------------------
[23:39:19]: --- Step: print_pwd ---
[23:39:19]: -----------------------
[23:39:19]: pwd: /Users/tahori/workspace/fast_test
[23:39:19]: -------------------------------
[23:39:19]: --- Step: print_pwd_wrapper ---
[23:39:19]: -------------------------------
[23:39:19]: pwd: /Users/tahori/workspace/fast_test/fastlane
```

The current directory in action should not be the Fastfile's directory...

### Description

The comment in `OtherAction#method_missing` seems to be outdated. I jus passed the corrent directory.

### DIFF
--- a/fastlane/lib/fastlane/other_action.rb
+++ b/fastlane/lib/fastlane/other_action.rb
@@ -11,17 +11,8 @@ module Fastlane
 
     # Allows the user to call an action from an action
     def method_missing(method_sym, *arguments, &_block)
-      # We have to go inside the fastlane directory
-      # since in the fastlane runner.rb we do the following
-      #   custom_dir = ".."
-      #   Dir.chdir(custom_dir) do
-      # this goes one folder up, since we're inside the "fastlane"
-      # folder at that point
-      # Since we call an action from an action we need to go inside
-      # the fastlane folder too
-
       self.runner.trigger_action_by_name(method_sym,
-                                         FastlaneCore::FastlaneFolder.path,
+                                         File.expand_path('..', FastlaneCore::FastlaneFolder.path),
                                          true,
                                          *arguments)
     end
--- a/fastlane/spec/action_spec.rb
+++ b/fastlane/spec/action_spec.rb
@@ -64,14 +64,14 @@ describe Fastlane do
 
     describe "Call another action from an action" do
       it "allows the user to call it using `other_action.rocket`" do
-        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
+        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(Dir.pwd)
         Fastlane::Actions.load_external_actions("./fastlane/spec/fixtures/actions")
         ff = Fastlane::FastFile.new('./fastlane/spec/fixtures/fastfiles/FastfileActionFromAction')
         Fastlane::Actions.executed_actions.clear
 
         response = {
           rocket: "🚀",
-          pwd: Dir.pwd
+          pwd: File.expand_path('..', Dir.pwd)
         }
         expect(ff.runner.execute(:something, :ios)).to eq(response)
         expect(Fastlane::Actions.executed_actions.map { |a| a[:name] }).to eq(['from'])


KrauseFx:
Hey @taisukeh, thanks for submitting this PR. This behavior is documented [over here](https://docs.fastlane.tools/advanced/#directory-behavior) - while we'd love to change this, it would break all existing setups that assume the folder location, from my understanding, so it's not something we can do

taisukeh:
@KrauseFx This PR is not about the behavior, but about the action called by **other_action**.
The action called by **other_action** does not runs in the root of the project, this behavier seems to be a bug 😞

KrauseFx:
Ah, right sorry. However I do thing that would break setups, no? In particular we even had a test covering this. The number 1 priority should be to not break any existing setups for the user

taisukeh:
Ok, I understand the priority. But there are many actions depend on the current working directory, they do not work properly when called from `other_action`.  This means `other_action` does work as expected in many cases.

For example, there are 43 actions that use File.expand_path, they are depend on the current working directory.
<details><summary>Actions that use File.expand_path </summary>
```
% find . -name '*.rb' | xargs -n 1 grep -m 1 -Hn 'File.expand_path' | awk '{ print $1 }'
./actions_helper.rb:73:
./appledoc.rb:65:
./apteligent.rb:37:
./automatic_code_signing.rb:8:
./backup_xcarchive.rb:24:
./clear_derived_data.rb:5:
./commit_github_file.rb:17:
./commit_version_bump.rb:11:
./copy_artifacts.rb:8:
./crashlytics.rb:100:
./create_keychain.rb:48:
./dotgpg_environment.rb:31:
./download_dsyms.rb:75:
./dsym_zip.rb:14:
./get_build_number.rb:14:
./get_info_plist_value.rb:12:
./get_ipa_info_plist_value.rb:9:
./get_version_number.rb:14:
./gradle.rb:27:
./gym.rb:71:
./hg_commit_version_bump.rb:11:
./hockey.rb:130:
./increment_build_number.rb:19:
./increment_version_number.rb:22:
./ipa.rb:33:
./nexus_upload.rb:25:
./oclint.rb:54:
./pem.rb:20:
./pilot.rb:12:
./register_devices.rb:30:
./screengrab.rb:17:
./set_build_number_repository.rb:51:
./set_info_plist_value.rb:11:
./setup_jenkins.rb:54:
./snapshot.rb:16:
./sonar.rb:9:
./splunkmint.rb:47:
./typetalk.rb:20:
./update_urban_airship_configuration.rb:8:
./upload_symbols_to_crashlytics.rb:22:
./verify_build.rb:27:
./verify_xcode.rb:127:
./xcodebuild.rb:172:
./zip.rb:9:
```
</details>
 <br />
 
I encountered this problem when I tried to make a wrapper action for my organization.
I think the `other_action` is not used so much at the moment, so the impact to existing setups is Influence is extremely small.

ohayon:
Hey @taisukeh, I think I agree with @KrauseFx here. While this would be great to change, many people have setups that are configured with this behavior in mind. Would it be feasible for you to do a `Dir.chdir` before the action to ensure you are in the right place? I know it's not ideal, but it could work? 🚀

taisukeh:
@ohayon Thank you for your comment.

> do a Dir.chdir before the action to ensure you are in the right place

`Dir.chdir` not work because the current working directory is changed to `"./fastlane"`( ==`FastlaneCore::FastlaneFolder.path` )  in `trigger_action_by_name`. Actions should work in project root directory as described in the [directory behavior](https://docs.fastlane.tools/advanced/#directory-behavior).

https://github.com/fastlane/fastlane/blob/c8f08826629cc1a2efbe82d192ea1b661b160d63/fastlane/lib/fastlane/other_action.rb#L23-L26

I think that it is closer to a bug fix than a behavior change. But of course, I'll close this PR if you think it shoud not be changed after all !

ohayon:
Ahh I see your point there. Unfortunately I dont think we will want to go ahead with this change at the moment because I do feel like it is going to break some existing people's configurations. Apologies for the inconvenience! 🚀

Issue #10308, opened 2017-09-11
Actions running from `cocoapods` `error_callback` block are running from wrong working directory


SiarheiFedartsou:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Simple lane:

```ruby
lane :test_cocoapods do
    cocoapods(error_callback: lambda  { |result|
        cocoapods(repo_update: true)
})
end
```
will produce(if first `cocoapods` call was failed):
```
➜  TestCocoapods bundle exec fastlane test_cocoapods --verbose
DEBUG [2017-09-11 22:01:54.88]: Checking if there are any plugins that should be loaded...
INFO [2017-09-11 22:01:54.99]: Driving the lane 'test_cocoapods' 🚀
INFO [2017-09-11 22:01:54.99]: -----------------------
INFO [2017-09-11 22:01:54.99]: --- Step: cocoapods ---
INFO [2017-09-11 22:01:54.99]: -----------------------
INFO [2017-09-11 22:01:55.20]: gem 'cocoapods' is already installed
INFO [2017-09-11 22:01:55.20]: $ bundle exec pod install
INFO [2017-09-11 22:01:56.17]: ▸ Analyzing dependencies
INFO [2017-09-11 22:01:56.17]: ▸ [!] The dependency `FooBarBarBar` is not used in any concrete target.
ERROR [2017-09-11 22:01:56.18]: Exit status of command 'bundle exec pod install' was 1 instead of 0.
Analyzing dependencies
[!] The dependency `FooBarBarBar` is not used in any concrete target.

DEBUG [2017-09-11 22:01:56.18]: X3
DEBUG [2017-09-11 22:01:56.18]: /Users/siarhei.fedartsou/TestCocoapods
INFO [2017-09-11 22:01:56.19]: -----------------------
INFO [2017-09-11 22:01:56.19]: --- Step: cocoapods ---
INFO [2017-09-11 22:01:56.19]: -----------------------
INFO [2017-09-11 22:01:56.39]: gem 'cocoapods' is already installed
INFO [2017-09-11 22:01:56.39]: $ pod install --repo-update
[...]

PR #10309, opened 2017-09-11
Fixed wrong working directory in cocoapods.error_callback call

SiarheiFedartsou:
### Checklist

### Motivation and Context
See the issue: https://github.com/fastlane/fastlane/issues/10308

### Description
I'm just changing directory to default fastlane folder before calling `error_callback`. Also I'm not sure: maybe it should be fixed at global level in `sh_helper.rb:sh_control_output` method?

### DIFF
--- a/fastlane/lib/fastlane/actions/cocoapods.rb
+++ b/fastlane/lib/fastlane/actions/cocoapods.rb
@@ -24,7 +24,15 @@ module Fastlane
         cmd << '--verbose' if params[:verbose]
         cmd << '--no-ansi' unless params[:ansi]
 
-        Actions.sh(cmd.join(' '), error_callback: params[:error_callback])
+        if params[:error_callback]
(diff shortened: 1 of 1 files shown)


SiarheiFedartsou:
I signed it!

ohayon (approved):
Thanks @SiarheiFedartsou ! This is a great change. While it would be great to put this in the `sh_control_output` method, I'm a bit worried that it will break some people's existing configurations and we dont really hear of this being an issue very often with other actions so I think we should roll with this change. Thanks again! 🚀 🐙

ohayon (changes requested):
Ah, before we merge this, should we check to make sure the `params[:error_callback]` is not nil before `call`ing it?

SiarheiFedartsou:
@ohayon Thanks for review! Just fixed the issue.

ohayon (approved):

ohayon:
Thanks @SiarheiFedartsou !! 🚀 !!

PR #10313, opened 2017-09-12
Show prettier error message when user has insufficient permissions in iTunes Connect

MortenGregersen:
### Checklist

### Motivation and Context
A iTunes Connect user with 'Developer' role trying to update meta data for an app update will receive an error response like this:
```ruby
{"data"=>nil, "messages"=>{"warn"=>nil, "info"=>nil, "error"=>["insufficient privileges"]}, "statusCode"=>"ERROR"}`
```
This error is not handled and the user will get the above response printed in the log.

### Description
This change will print a prettier message when the above happens.

To ensure that the action/`calling_method_name` was correct in this case, I had to add a new option to `raise_insuffient_permission_error!`. Without this option the error would say that the user don't have permission to `parse_reponse`. I'm not sure this is the right way to do it.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -608,6 +608,8 @@ module Spaceship
       # Check if the failure is due to missing permissions (iTunes Connect)
       if body["messages"] && body["messages"]["error"].include?("Forbidden")
         raise_insuffient_permission_error!
+      elsif body["messages"] && body["messages"]["error"].include?("insufficient privileges")
+        raise_insuffient_permission_error!(caller_location: 3)
       elsif body.to_s.include?("Internal Server Error - Read")
         raise InternalServerError, "Received an internal server error from iTunes Connect / Developer Portal, please try again later"
       elsif (body["resultString"] || "").include?("Program License Agreement")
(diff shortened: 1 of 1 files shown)


ohayon (commented):
Hey @MortenGregersen thanks for this change! It would be really nice to have a better message for that. I am a bit nervous about just checking this string as Apple is known to change things around pretty often. Is there any other way we can detect this error? 🐙

MortenGregersen:
@ohayon The only information we get back from Apple is the response I wrote above. If we should check for anything else we should look at the roles of the user.

ohayon:
I think that might be a little safer because I think at least those are less likely to change? Thoughts? 🐒

MortenGregersen:
But it will require that we make a new request. I don't think we have information about the user at this point.

Furthermore we need to analyze how the roles relates to the different actions the user can perform.

spaceship/lib/spaceship/client.rb
```diff
@@ -608,6 +608,8 @@ def detect_most_common_errors_and_raise_exceptions(body)
       # Check if the failure is due to missing permissions (iTunes Connect)
       if body["messages"] && body["messages"]["error"].include?("Forbidden")
         raise_insuffient_permission_error!
+      elsif body["messages"] && body["messages"]["error"].include?("insufficient privileges")
+        raise_insuffient_permission_error!(caller_location: 3)
```
taquitos: I think specifying `caller_location` is ok here, but lets add some comments around why you're passing 3, and what the effect is.

taquitos (changes requested):
Just add some comments and 👍

MortenGregersen:
@taquitos I've added comments about the `caller_location`

taquitos (approved):

taquitos:
🎈🐐

PR #10321, opened 2017-09-13
Fix creation of certificate when a keychain is specified.

francoisferrand:
In this case, the certificate is properly created in the specified
keychain, but the certificate validation used to fail since it kept
searching in the keychain search list.

This patch ensures the verification is performed using the specified
keychain file.

### DIFF
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -3,10 +3,10 @@ require 'tempfile'
 module FastlaneCore
   # This class checks if a specific certificate is installed on the current mac
   class CertChecker
-    def self.installed?(path)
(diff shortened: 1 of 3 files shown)


Typz:
I signed it!

ohayon (approved):
This looks great! Thanks @Typz !! 🚀  Welcome to the fastlane community!

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -3,10 +3,10 @@
 module FastlaneCore
   # This class checks if a specific certificate is installed on the current mac
   class CertChecker
-    def self.installed?(path)
+    def self.installed?(path, keychain = nil, password = nil)
```
casz: This line is causing a regression on match

PR #9559, opened 2017-06-22
[snapshot] Add support for Xcode 9

ohayon:
In Xcode 9, Apple decided to use `Filename` rather than Xcode 8's `FileName` as a key in the plist generated by tests. This change just looks for that key if using Xcode 9.

PR #10326, opened 2017-09-13
[reset_simulator_contents] Soften xcode 9 action failure to a warning

jaleksynas:
### Checklist

### Motivation and Context
When upgrading xcode or switching between Xcode 8.X and xcode 9 it is very difficult to continually remove this action from the Fastfile to avoid a hard failure.  For backward compatibility and easier forward compatibility, a warning and continue seems a more favorable behavior.

Executed the task with several versions of xcode to ensure that it would message and act properly.

### Description
Change the message from user_error! to important.  Return from the run action immediately so as to not attempt the non-functional operations.

### DIFF
--- a/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
+++ b/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
@@ -3,7 +3,8 @@ module Fastlane
     class ResetSimulatorContentsAction < Action
       def self.run(params)
         if Helper.xcode_at_least?("9")
-          UI.user_error!("resetting simulators currently doesn't work with Xcode 9, stay tuned as we are working to add support for all new tools.")
+          UI.important("Resetting simulators currently doesn't work with Xcode 9, stay tuned as we are working to add support for all new tools.")
+          return
         end
 
         if params[:ios]


benasher44:
I looked at #9559, and it's not clear why this is here. All of the same `xcrun simctl` commands required by the `reset` method in `FastlaneCore` appear to be present in Xcode 9. Why is there a message here at all? It might be nice to document this issue (or just link to it again if I missed it) for someone who's willing to fix it :)

benasher44:
FWIW, I just tried running `FastlaneCore::Simulator.reset_all` on my machine with Xcode 9 GM, and it ran without issue. It's only one data point, but maybe we can remove this? 😅

jaleksynas:
@ohayon This PR could use your input as the author of the code in question.

ohayon:
Thanks for this PR @jaleksynas ! You're totally right, this should be a warning not an error. and @benasher44 - this has not been working in beta versions of Xcode that is why the message is still there. I am going to merge this PR and then we can do some testing to make sure that the action works for all simulators in all versions of Xcode. Thank you both again! 🚀

ohayon (approved):

Issue #10033, opened 2017-08-11
fastlane supply fails with error "NoMethodError: [!] undefined method `map' for nil:NilClass"


trevoriancox:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

This command worked yesterday but fails today. 

```
`bundler exec fastlane supply --track beta --json_key ./appazur-build.json --skip_upload_screenshots --package_name com.... --apk .../com.....apk --skip_upload_apk`

+-------------------------+----------------------------------------------------------------------+
|                                   Summary for supply 2.53.1                                    |
+-------------------------+----------------------------------------------------------------------+
| track                   | beta                                                                 |
| json_key                | ./appazur-build.json                                                 |
| skip_upload_screenshots | true                                                                 |
| package_name            | com.appazur...                                         |
| apk                     | .../bin/Release/com.appazur.....apk |
| skip_upload_apk         | true                                                                 |
| metadata_path           | ./fastlane/metadata/android                                          |
| skip_upload_metadata    | false                                                                |
| skip_upload_images      | false                                                                |
| validate_only           | false                                                                |
| check_superseded_tracks | false                                                                |
+-------------------------+----------------------------------------------------------------------+

[12:22:49]: Preparing to upload for language 'en-US'...
[12:22:49]: Uploading image file ./fastlane/metadata/android/en-US/images/featureGraphic.png...
[12:22:51]: Uploading image file ./fastlane/metadata/android/en-US/images/icon.png...
[12:22:53]: Uploading image file ./fastlane/metadata/android/en-US/images/promoGraphic.png...

Looking for related GitHub issues on fastlane/fastlane...

➡️  Crash on registering device UDID
    https://github.com/fastlane/fastlane/issues/8106 [closed] 3 💬
    30 May 2017

➡️  fastlane reporting apple developer website is down
    https://github.com/fastlane/fastlane/issues/5918 [closed] 19 💬
    06 Dec 2016

➡️  deliver fails to save screenshots with "must be a .jpeg, .jpg, or .png (...)"
    https://github.com/fastlane/fastlane/issues/6426 [closed] 17 💬
    04 Jan 2017

[...]

PR #10330, opened 2017-09-14
[supply] Fix crash in apks_version_codes

ohayon:
Fix https://github.com/fastlane/fastlane/issues/10033


KrauseFx (approved):
Array all the things

Issue #10319, opened 2017-09-13
Slather command "source-files" is wrong


Xcode1994:
[Slather/Command.rb](https://github.com/SlatherOrg/slather/blob/master/lib/slather/command/coverage_command.rb)

Line 32
You can see `option ["--source-files"], "SOURCE_FILES", "A Dir.glob compatible pattern used to limit the lookup to specific source files. Ignored in gcov mode.", :multivalued => true`

So the option --source-files is Array not String

But in your source code [lib/fastlane/actions/slather.rb](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/slather.rb)

Line 240-243
it means you think the option --source-files is string

hope you guys can fix this problem
thank you

ohayon:
Hey @Xcode1994! Thanks for writing in an identifying the fix. Is there any way you would want to submit a PR with that change? We would love to review it! 🚀

Xcode1994:
@ohayon hahahahahhhh, I am not familiar with Ruby, But I will try to fix it

KrauseFx:
Merged with https://github.com/fastlane/fastlane/pull/10333 :+1: Thanks everyone

PR #10333, opened 2017-09-15
[Fix slather.rb] option source_files type is Array not String

QIYA0130:
### Checklist

### Motivation and Context

You can check https://github.com/fastlane/fastlane/issues/10319 
source-files type was wrong, I fix it with this PR
### Description

When initialize ```source_files``` add ```is_string: false``` to define type is not string

### DIFF
--- a/fastlane/lib/fastlane/actions/slather.rb
+++ b/fastlane/lib/fastlane/actions/slather.rb
@@ -240,6 +240,7 @@ Slather is available at https://github.com/SlatherOrg/slather
           FastlaneCore::ConfigItem.new(key: :source_files,
                                        env_name: "FL_SLATHER_SOURCE_FILES",
                                        description: "A Dir.glob compatible pattern used to limit the lookup to specific source files. Ignored in gcov mode",
+                                       is_string: false,
                                        default_value: false,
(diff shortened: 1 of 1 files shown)


ohayon:
So I do believe that this parameter actually should be a string as it is used in a `Dir.glob` call rather than being an array of files. Does that make sense?

DEMO0130:
In my project, I should calculate code coverage in specific files, submit code coverage for a new feature to project manager, so there are a series of file path need to be call. So It should be Array.

Just like this
```
slather(
         html: true,
         build_directory: 'derivedData',
         output_directory: "fastlane/test_output/CodeCoverage",
         source_files: ['X/App/Invoice/*', 'Pods/FRWSearchKit/*'],
         proj: "X.xcodeproj",
         workspace: "X.xcworkspace",
         scheme: "X")
```

For your source code, it will checkout this parameter should be String, I can not do what I want.
Can u understand what I mean ??

Or is there other way to do what I want to do, u can tell me, I will try.

Thank u.

KrauseFx (approved):
Awesome, this looks great, thanks for your contribution 👍

Issue #10134, opened 2017-08-24
Show team name next to id in deliver error message


revolter:
### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I ran deliver while having an incorrect ID in the `Appfile` and I got this error:

    [16:14:24]: Could not set team ID to '123', only found the following available teams: 456, 789

It would be nice if the team name would be next to the IDs like:

    [16:14:24]: Could not set team ID to '123', only found the following available teams: 456 (Team 1), 789 (Team 2)

for easier solving of this.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

[INSERT OUTPUT HERE]

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

[INSERT OUTPUT HERE]

mpirri:
This is a really good idea - any chance you might be interested in preparing a pull request to add this? We are always looking for awesome PRs from the _fastlane_ community, and this would be a great change.

Check out [this document](https://github.com/fastlane/fastlane/blob/master/YourFirstPR.md) for more information on contributing to _fastlane_. 🚀

revolter:
Is there any way to mock the developer account? As I wouldn't want to mess with the work account.

ohayon:
Hey @revolter, you should be able to look at some of the tests in `spaceship/specs` and see if anything makes sense there for mocking out the account. 😄  🐙

RishabhTayal:
Created a PR for this. 👍

PR #10339, opened 2017-09-15
[deliver] shows team name in the error message

RishabhTayal:
### Checklist

### Motivation and Context
Shows team name in the error message when the wrong `team_id` is provided in the `Appfile`.
Fixes: https://github.com/fastlane/fastlane/issues/10134

Tested locally by setting up `fastlane` in a test project. Entered right `username` and wrong values for `team_id` in the `Appfile`.


spaceship/lib/spaceship/client.rb
```diff
@@ -202,15 +202,18 @@ def team_id=(team_id)
...
       #
       available_teams = teams.collect do |team|
-        (team["contentProvider"] || {})["contentProviderId"]
+        {
+          team_id: (team["contentProvider"] || {})["contentProviderId"],
+          team_name: (team["contentProvider"] || {})["name"]
+        }
       end
 
       result = available_teams.find do |available_team_id|
```
ohayon: Now that `available_teams` is an array of hashes is this going to work appropriately? Is there any chance you could add a test for this?

ohayon (commented):
Thanks @RishabhTayal, I left a comment on the PR and am curious to hear your thoughts. 😄

spaceship/lib/spaceship/client.rb
RishabhTayal: Yeah I don't see this variable being used anywhere else. The scope of the variable is local so it shouldn't affect any code outside the function.

ohayon: Line 212 here is going to fail because it is trying to compare the `team_id` to the `available_team_id` but with this change, `available_team_id` is a hash right?

RishabhTayal: Good catch. I fixed it.

RishabhTayal: @ohayon do you know how can I get rid of all those commits in my branch?

ohayon: You should be able to do a `git rebase -i HEAD~28` and then select fixup for all but the important commit. Also, maybe an easier way would be to just pull master, checkout a new branch and remake your changes. I think that will be the faster option.

RishabhTayal: @ohayon Thanks, I will do that for the next pull request.

ohayon:
Hey @RishabhTayal, is there any chance you may be able to add a test or two for this? I just want to make sure that we can edit this code moving forward and are not missing any problems like the one mentioned above 😄  🚀

RishabhTayal:
Sure. Could you help me with some pointer? Not very familiar with ruby unit tests. I am adding in `spaceship/spec/client_spec.rb` file.

ohayon:
That is definitely the right place to start 🚀

RishabhTayal:
@ohayon I am not able to come up with a unit test for this. Not sure how to go about it. Will appreciate if can get some help.

MortenGregersen (approved):
Looks good to me. Not sure either how to make unit tests for this.

ZevEisenberg:
+1 would love to see this merged. This fix would have saved me a ton of time today.

KrauseFx (approved):
Awesome, thanks 👍 I pushed a small change so that the output has correct spaces inbetween commas, it's ready now

revolter:
Thank you too for the PR, @RishabhTayal.

Issue #10347, opened 2017-09-18
Snapshot do not supports multiple Apple TV


juli1quere:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
In snapshot: as soon as we have multiple Apple TV devices in the `devices` list, it trig this error: 
```
[!] All devices provided to snapshot should run the same operating system
```
For example, the following example trig this error: 
```ruby
devices([
  "Apple TV 1080p",
  "Apple TV 4K"
])
```

But if we run `snapshot` with only one device (`Apple TV 1080p` or `Apple TV 4K`), everything works perfectly. 

This issue seems to be related to this method in [TestCommandGenerator](https://github.com/fastlane/fastlane/blob/f410f2a6d9d20ae8e6747db978cae7142a03d781/snapshot/lib/snapshot/test_command_generator.rb#L50): 
```ruby
      def verify_devices_share_os(devices)
        # Check each device to see if it is an iOS device
        all_ios = devices.map do |device|
          device = device.downcase
          device.start_with?('iphone', 'ipad')
        end
        # Return true if all devices are iOS devices
        return true unless all_ios.include?(false)
        # There should only be more than 1 device type if
        # it is iOS, therefore, if there is more than 1
        # device in the array, and they are not all iOS
        # as checked above, that would imply that this is a mixed bag
        return devices.count == 1
      end
```

Indeed, if we have more than 1 device which is not on iOS: the method returns false. It was good by the past (since we had only one available non-iOS device: `Apple TV`). But now, we have to handle multiple Apple TVs. 

I'll try to submit a PR in the next days.
### Environment
fastlane 2.56.0 :)

ohayon:
Thanks @juli1quere! We would love to review that PR!

PR #10350, opened 2017-09-18
Allow snapshot to handle multiple tvOS devices

jlnquere:
### Checklist

### Motivation and Context
This PR is relative to #10347.  Now, we can use `snapshots` with multiple Apple TV devices. Like: 
```ruby
devices([
	"Apple TV 1080p", 
	"Apple TV 4K", 
	"Apple TV 4K (at 1080p)"
])
```

### Description
I edited `verify_devices_share_os()` method to do so. I also added some tests that checks if `verify_devices_share_os`:
- Still return true if there is only iOS devices. 
- Still return false if there is iOS and tvOS devices.
- Return true if there is only tvOS devices (and more than 1).

I also checked that the error mentioned in #10347 is not trigged anymore.

### DIFF
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -55,9 +55,17 @@ module Snapshot
         end
         # Return true if all devices are iOS devices
         return true unless all_ios.include?(false)
+
+        all_tvos = devices.map do |device|
+          device = device.downcase
+          device.start_with?('apple tv')
+        end
+        # Return true if all devices are iOS devices
+        return true unless all_tvos.include?(false)
+
         # There should only be more than 1 device type if
-        # it is iOS, therefore, if there is more than 1
-        # device in the array, and they are not all iOS
+        # it is iOS or tvOS, therefore, if there is more than 1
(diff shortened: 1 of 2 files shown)


snapshot/lib/snapshot/test_command_generator.rb
```diff
@@ -55,9 +55,17 @@ def verify_devices_share_os(devices)
         end
         # Return true if all devices are iOS devices
         return true unless all_ios.include?(false)
+
+        all_tvos = devices.map do |device|
+          device = device.downcase
+          device.start_with?('apple tv')
```
KrauseFx: Doesn't the `.map` method use the return value, so in this case the `all_tvos` array would be full of `[true, false, true, true]`?

juli1quere: Indeed, `all_tvos` is an array of boolean (`true` if a device starts with `apple tv`).  After that, the method will return `true` if `all_tvos` is full of `true` (i.e. all devices are `apple tv`).

KrauseFx (approved):
Sounds good, thanks for clarifying :+1:

Issue #10358, opened 2017-09-18
[Regression] "no implicit conversion of nil into String" in match during cert install after updating to 2.57 


nbrice88:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

After updating to version 2.57 the `match` step now fails with the error `no implicit conversion of nil into String` when trying to install certificates. Running the same command in 2.56 succeeds.

From the debug output it looks to be choking when installing the Apple Worldwide Developer Relations Certification Authority.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

### Captured Output

Command Used: `refreshCertificates --env DEV`
<details><summary>Output/Log</summary>

```

DEBUG [2017-09-18 13:22:54.06]: Checking if there are any plugins that should be loaded...
INFO [2017-09-18 13:22:54.18]: -------------------------------------------------
INFO [2017-09-18 13:22:54.18]: --- Step: Verifying required fastlane version ---
INFO [2017-09-18 13:22:54.18]: -------------------------------------------------
INFO [2017-09-18 13:22:54.18]: Your fastlane version 2.57.0 matches the minimum requirement of 2.56.0  ✅
INFO [2017-09-18 13:22:54.19]: ------------------------------
INFO [2017-09-18 13:22:54.19]: --- Step: default_platform ---
INFO [2017-09-18 13:22:54.19]: ------------------------------
INFO [2017-09-18 13:22:54.19]: Loading from './fastlane/.env.DEV'
INFO [2017-09-18 13:22:54.19]: Driving the lane 'ios refreshCertificates' 🚀
INFO [2017-09-18 13:22:54.20]: -------------------
INFO [2017-09-18 13:22:54.20]: --- Step: match ---
INFO [2017-09-18 13:22:54.20]: -------------------
DEBUG [2017-09-18 13:22:54.21]: Taking value for 'git_url' from environment variable 'MATCH_GIT_URL'
DEBUG [2017-09-18 13:22:54.21]: Taking value for 'git_branch' from environment variable 'MATCH_GIT_BRANCH'
[...]

Issue #9177, opened 2017-05-16
Can't run tests with bundle exec rspec


joelklabo:
### New Issue Checklist

- [x ] Updated fastlane to the latest version
- [x ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Having trouble getting the tests to run. Using `bundle exec rspec` in root directory.

Main error I am seeing: 
`Could not load native IDN implementation.
LoadError: cannot load such file -- rack/mount`

FULL OUTPUT

`$>
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
Changing stdout to /tmp/fastlane_tests, set `DEBUG` environment variable to print to stdout (e.g. when using `pry`)
/Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/addressable-2.5.1/spec/addressable/idna_spec.rb:145: warning: already initialized constant LONG
/Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/addressable-2.5.1/spec/addressable/idna_spec.rb:29: warning: previous definition of LONG was here
Could not load native IDN implementation.
LoadError: cannot load such file -- rack/mount
  /Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/addressable-2.5.1/spec/addressable/rack_mount_compat_spec.rb:21:in `require'
  /Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/addressable-2.5.1/spec/addressable/rack_mount_compat_spec.rb:21:in `<top (required)>'
  /Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.5.4/lib/rspec/core/configuration.rb:1435:in `load'
  /Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.5.4/lib/rspec/core/configuration.rb:1435:in `block in load_spec_files'
  /Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.5.4/lib/rspec/core/configuration.rb:1433:in `each'
  /Users/joelklabo/src/fastlane/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.5.4/lib/rspec/core/configuration.rb:1433:in `load_spec_files'
[...]

PR #10359, opened 2017-09-19
Check for null keychain path before attempting to use it

nbrice88:
### Checklist
- [ ] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass. Can not run in any directory because of https://github.com/fastlane/fastlane/issues/9177. Happy to run if I can get it working.

### Motivation and Context
Upgrading to 2.57 broke the match command for my workflow as referenced in this issue: https://github.com/fastlane/fastlane/issues/10358. 
At @snelson21 's suggestion I have submitted this fix based on his comments.
As I am not currently able to run tests via rspec, I ran my updated fork in my app's repo (by pointing my Gemfile at my local fork) and have confirmed that the `self.list_available_identities(keychain, password)` no longer fails on a null `keychain` parameter.

### Description
I have added an `if keychain` check to the `self.list_available_identities(keychain, password)` method. Previously it called `keychain = File.expand_path(keychain)`, but the `keychain` value may be null, which would cause the `match` fastlane command to fail in my workflow. I'm happy to make any changes or add tests as seen fit.

### DIFF
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -45,7 +45,7 @@ module FastlaneCore
     end
 
     def self.list_available_identities(keychain, password)
-      keychain = File.expand_path(keychain)
+      keychain = File.expand_path(keychain) if keychain
       locked = keychain and !system("security show-keychain-info #{keychain} >/dev/null 2>/dev/null")
       `security unlock -p #{password} #{keychain}` if locked
       `security find-identity -v -p codesigning #{keychain}`


nbrice88:
I signed it!

agfa555 (approved):

KrauseFx (approved):
Awesome, thanks for the PR :+1:

KrauseFx:
Actually, I'm gonna go with a rollback due to https://github.com/fastlane/fastlane/issues/10358 instead, and let's have a separate PR to add this feature again. Sorry about that.

PR #10364, opened 2017-09-19
Add test to make sure uploaded IPAs do not contain special chars

ohayon:
This PR adds a test to make sure that we don't allow changes that introduce special characters into ipa file names.


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane_core/spec/ipa_upload_package_builder_spec.rb
```diff
@@ -0,0 +1,26 @@
+describe FastlaneCore do
+  describe FastlaneCore::IpaUploadPackageBuilder do
+    let(:ipa) { 'iOSAppOnly' }
+    let(:path) { File.expand_path("../fixtures/ipas/#{ipa}.ipa", __FILE__) }
+    let(:uploader) { FastlaneCore::IpaUploadPackageBuilder.new }
+    let(:unique_path) { uploader.unique_ipa_path(path) }
+
+    context 'unique IPA file name' do
+      it 'does not contain any special characters' do
+        special_chars = %w[! @ # $ % ^ & * ( ) + = [ ] " ' ; : < > ? / \ | { } , ~ `]
```
KrauseFx: Ha, nice 👍 Wouldn't it be easier to do it the other way around in this case and use a regex like `[\w\-\.\_]+`?

KrauseFx: Just thinking out loud here, this test would be green if the file name contained `‘` or `ø`, or an emoji

ohayon: Ah good call @KrauseFx, yeah I think a regex would be the better option!

KrauseFx: Let's add a test for the test to see if it's covering emojis 😂

KrauseFx (changes requested):
Hey @ohayon, this PR looks really good, and we had only one small change we need to get this merged 👍 

Let me know if you have the time to update the PR, otherwise I'm also happy to jump in and take over

KrauseFx (approved):
Works for me, let's 🚀

janpio:
Don't you want to harden the test a bit with a regex any more?

janpio:
(You can ignore the failing test on AppVeyor, this is because newer changes to the tests and test setup were not merged in here. Should be good after merge to `master`)

twodave:
FYI this change broke ipa's with spaces in them. Need to scrub for those still (as the previous code was doing). There are a couple closed issues on this as the bug has been reintroduced before.

PR #10386, opened 2017-09-20
[Gym] code_signing_mapping ignore target with unset bundle identifier

nicolasbraun:
### Motivation and Context
In some projects we might have target that do not have a bundle_identifier, specifically targets that are built as .a lib and used as target dependencies for "real targets".
Those target cause the tentative to match profile and bundles to fail.


KrauseFx (approved):
Nice catch, thanks :+1:

PR #10417, opened 2017-09-26
Comment out carthage, cocoapods actions instead of removing

d-date:
### Checklist

### Motivation and Context

When I running `fastlane init`, a awesome Fastfile is generated, but I found missing deleting spaces with not contain `cocoapods` and `carthage`, so Fastfile should be more beautiful.

screen shot 2017-09-26 at 21 47 42

### Description

remove pre 4 spaces from Fastfile when gsub cocoapods and carthage with ''.

### DIFF
--- a/fastlane/lib/fastlane/setup/setup_ios.rb
+++ b/fastlane/lib/fastlane/setup/setup_ios.rb
@@ -276,8 +276,8 @@ module Fastlane
       end
 
       template.gsub!('snapshot', '# snapshot') unless self.tools[:snapshot]
-      template.gsub!('    cocoapods', '') unless self.tools[:cocoapods]
-      template.gsub!('    carthage', '') unless self.tools[:carthage]
+      template.gsub!('cocoapods', '# cocoapods') unless self.tools[:cocoapods]
+      template.gsub!('carthage', '# carthage') unless self.tools[:carthage]
       template.gsub!('[[FASTLANE_VERSION]]', Fastlane::VERSION)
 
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/setup/setup_ios.rb
```diff
@@ -276,8 +276,8 @@ def generate_fastfile(manually: false)
       end
 
       template.gsub!('snapshot', '# snapshot') unless self.tools[:snapshot]
-      template.gsub!('cocoapods', '') unless self.tools[:cocoapods]
-      template.gsub!('carthage', '') unless self.tools[:carthage]
+      template.gsub!('    cocoapods', '') unless self.tools[:cocoapods]
+      template.gsub!('    carthage', '') unless self.tools[:carthage]
```
nafu: I think adding `# ` would be better like snapshot. @KrauseFx what do you think?

nafu (commented):
Thanks @d-date for contribution 🚀

fastlane/lib/fastlane/setup/setup_ios.rb
KrauseFx: Oh yeah, I'd be ok with adding a `#` instead of removing it 👍  Thanks y'all :)

d-date:
Thanks @nafu!, so I fixed as follow you 👍

nafu (approved):
Great 🎉  @d-date nice catch!

PR #10451, opened 2017-09-28
[spaceship] Add pagination support to Client#request

UnsafePointer:
### Checklist

### Motivation and Context

We're using parts of the spaceship API to interact with TestFlight and we noticed that the method `get_builds_for_train` only returns the last 20 builds.

### Description

This PR adds an option to `Client#request` to auto-paginate results (inspired a bit by [Octokit](https://github.com/octokit/octokit.rb/blob/8404007e9e11d3695df0907adcb3514decd23287/lib/octokit/connection.rb#L87-L94)). This could be used in more methods besides `get_builds_for_train`, but it will add quite a lot of overhead. 

In the case of `get_builds_for_train`, it does make sense to return all builds available.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -6,6 +6,7 @@ require 'logger'
 require 'spaceship/babosa_fix'
 require 'spaceship/helper/net_http_generic_request'
 require 'spaceship/helper/plist_middleware'
+require 'spaceship/helper/rels_middleware'
 require 'spaceship/ui'
 require 'tmpdir'
 require 'cgi'
@@ -259,6 +260,7 @@ module Spaceship
         c.response :xml, content_type: /\bxml$/
         c.response :plist, content_type: /\bplist$/
         c.use :cookie_jar, jar: @cookie
+        c.use FaradayMiddleware::RelsMiddleware
         c.adapter Faraday.default_adapter
 
         if ENV['SPACESHIP_DEBUG']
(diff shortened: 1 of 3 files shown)


spaceship/lib/spaceship/client.rb
```diff
@@ -700,6 +704,24 @@ def send_request(method, url_or_path, params, headers, &block)
...
     end
 
+    def send_request_auto_paginate(method, url_or_path, params, headers, &block)
+      response = send_request(method, url_or_path, params, headers, &block)
+      return response if response.body.nil? || response.body['data'].class != Array
+      page = response
+      loop do
+        link = page.headers['link']
+        break if link.nil?
+        link_match = link.match(/<(.*)>/)
```
KrauseFx: uhhhhh

KrauseFx (changes requested):
I think we should add paging, but it looks like there is HTML parsing involved?

spaceship/lib/spaceship/client.rb
Ruenzuo: ehhhhh

Ruenzuo:
This is how the "Link" response header looks like: 
screen shot 2017-10-02 at 10 26 13

Not sure if that qualifies as [parsing HTML with regex](https://stackoverflow.com/a/1732454). I looked for a better way to do this on the faraday API but I couldn't find anything, although [Octokit has a nice API](https://github.com/octokit/octokit.rb/blob/8404007e9e11d3695df0907adcb3514decd23287/lib/octokit/connection.rb#L88) for it. Will try to see how they're doing that and propose a better solution for this.

Ruenzuo:
I had a deeper look at how Octokit does this. Octokit uses [Sawyer](https://github.com/lostisland/sawyer) (based on Faraday) for hypermedia processing so I'm not sure if changing that in Spaceship is an alternative. Sawyer [uses regex to parse the response header](https://github.com/lostisland/sawyer/blob/master/lib/sawyer/response.rb#L43-L49).

Instead, I created a Faraday middleware that expands env a and adds a field to store the rels processing. Rels processing is now using the same method as Sawyer.

KrauseFx (approved):
Alright, this looks great, thanks for your work on this. Were you thinking about submitting a separate PR to make use of the new parameters/methods?

Ruenzuo:
I updated `Spaceship::TestFlight::Client#get_builds_for_train` to make use of the new parameter, because this is how I found out that fastlane was missing this feature.
By default, Apple API returns 20 items per request. This is probably configurable by a parameter in the query. For some methods like `get_builds_for_train`, it makes sense to retrieve all builds, but I don't think this should be the default behaviour in the rest of the API, as this potentially comes with a big overhead if we leave the default query on 20 items.
Not really sure what's the best way of doing this, and also I don't know if there are other people who have complaint about this already. My suggestion would be to make use of the new parameter when it makes sense, only to solve problems reported by users.

KrauseFx:
Alright, beautiful, thanks for your contribution ❤️

KrauseFx:
Also @Ruenzuo, can you drop me a quick email, would love to add you to the fastlane contributor slack :)

PR #10455, opened 2017-09-28
Fix match enterprise profile mapping

nicolasbraun:
When trying to find matching profiles in `merge_profile_mapping` we are searching from profiles in the match response that contains the export-method (case 3.2).

match enterprise profile contain the InHouse word instead of enterprise. Fixing the `app_identifier_contains?` helper to better map match profile.

```
(byebug) primary_mapping
{:"com.applidium.keepcool-dev"=>"match InHouse com.applidium.keepcool-dev"} # match output
(byebug) secondary_mapping
{:"com.applidium.keepcool-dev"=>"PROFILE_UDID"} # xcode detected
```
should result in {:"com.applidium.keepcool-dev"=>"match InHouse com.applidium.keepcool-dev"}

### DIFF
--- a/gym/lib/gym/code_signing_mapping.rb
+++ b/gym/lib/gym/code_signing_mapping.rb
@@ -69,7 +69,7 @@ module Gym
     # We do some `gsub`bing, because we can't really know the profile type, so we'll just look at the name and see if it includes
     # the export method (which it usually does, but with different notations)
     def app_identifier_contains?(str, contains)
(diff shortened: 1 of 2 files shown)


taquitos (approved):
🎈:goat:

nicolasbraun:
An immediate release would be great indeed. All our CI is broken now as we deliver all client with enterprise codesign :)

PR #10504, opened 2017-10-04
Add `try_repo_update_on_error` option to `cocoapods` action

SiarheiFedartsou:
### Checklist

### Motivation and Context
A lot of CocoaPods users from time to time face issue when after updating some pod locally they have CocoaPods error on CI side, because it is needed to go there and run something like `bundle exec pod repo update`(because CI doesn't know that dependency has some new version). So, I propose to add `try_repo_update_on_error` option to `cocoapods` action, which will try to execute pod command one more time with `--repo-update` in case of error. 
Also, yes, I know about already implemented `repo_update` option. But while it is really worth to run in 5% of cases, it take significant time to wait while it performing every time(and also gives additional load on GitHub servers).
Thanks!

### DIFF
--- a/fastlane/lib/fastlane/actions/cocoapods.rb
+++ b/fastlane/lib/fastlane/actions/cocoapods.rb
@@ -106,7 +106,7 @@ module Fastlane
                                        description: 'Retry with --repo-update if action was finished with error',
                                        optional: true,
                                        is_string: false,
-                                       default_value: false,
+                                       default_value: true,
                                        conflicting_options: [:repo_update])
         ]
         # Please don't add a version parameter to the `cocoapods` action. If you need to specify a version when running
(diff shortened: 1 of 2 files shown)


KrauseFx:
Great idea, I like to fact that this would update the repo only if it needs to be updated (probably one of the most common error cases of the `cococapods` fastlane action. Would love a confirmation from @orta that this is a good behavior with no side effects. We could even enable that option by default, right?

orta:
Yeah, seems fine to me 👍

SiarheiFedartsou:
@KrauseFx yes, maybe we can enable it by default, but some of users already `repo_update` option enabled(which is not compatible with new `try_repo_update_on_error`) and we need to think about behavior in this case.
Maybe we can enable `try_repo_update_on_error` by default, but disable it for users who pass `repo_update: true` option?
And also: do someone really need `repo_update` option(which takes significant time) if we have this one? Maybe we can deprecate it?

KrauseFx:
Ah alright, let's leave it as is then. Is it ready to be reviewed & merged from your side @SiarheiFedartsou?

Thanks for confirming @orta

SiarheiFedartsou:
@KrauseFx yes, I think :)

KrauseFx (approved):
Awesome, thanks for your contribution @SiarheiFedartsou 👍

PR #10514, opened 2017-10-05
[set_info_plist_value] add option to merge with map

jetersen:
### Checklist

### Motivation and Context
Felt set_info_plist_value a bit overbearing, wanted to update values as a hashmap.

Further motivation was this was the only action that did not require an Xcode project file to change the plist...
Yaaa __Xamarin__!

### Description
Added parameter that can merge plist with hashmap to overwrite any values you desire.


KrauseFx:
Hey @casz, thanks for the PR 👍 How much complexity would it add to our existing action? We usually try to avoid adding new actions to the code base for now :)

casz:
No trouble at all 👍

casz:
Basically this is the special code:
https://github.com/casz/fastlane/blob/9b7ebab16ebbace1178fc6f6d1f1b232372e5419/fastlane/lib/fastlane/actions/set_info_plist_map.rb#L12-L15

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -45,7 +48,7 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :key,
                                        env_name: "FL_SET_INFO_PLIST_PARAM_NAME",
                                        description: "Name of key in plist",
-                                       optional: false),
+                                       optional: true),
```
casz: optional unless key :map is null. is that possible?

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -79,8 +88,34 @@ def self.is_supported?(platform)
...
+          )',
+          'set_info_plist_map(
+            path: "./Info.plist",
+            map: {
+              "CFBundleIdentifier" => "com.example.fastlane",
+              "CFBundleShortVersionString" => "1.1.1",
+              "CFBundleVersion" => "9999",
+            }
+          )',
+          'set_info_plist_map(
```
casz: Should perhaps just show one example

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -79,8 +88,34 @@ def self.is_supported?(platform)
...
+          )',
+          'set_info_plist_value(
+            path: "./Info.plist",
+            map: {
+              "CFBundleIdentifier" => "com.example.fastlane",
+              "CFBundleShortVersionString" => "1.1.1",
+              "CFBundleVersion" => "9999",
+            }
+          )',
+          'set_info_plist_value(
```
casz: reduce to one example? do we care that it can handle symbols?

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -10,7 +10,10 @@ def self.run(params)
         begin
           path = File.expand_path(params[:path])
           plist = Plist.parse_xml(path)
-          if params[:subkey]
+          if params[:map]
+            stringified_hash = Hash[params[:map].collect { |k, v| [k.to_s, v] }]
```
casz: `plist` from `Plist.parse_xml` is stringified, and merge cannot merge symbols and strings it duplicates them.
So we have to convert it to keyed strings.

casz: `plist` from `Plist.parse_xml` is stringified, and `Hash.merge`  cannot merge symbols and strings, rather it duplicates them.
So we have to convert it to keyed strings.

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: optional unless key :map is null. is that possible?

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: reduce to one example? do we care that it can handle symbols?

casz:
@KrauseFx PTAL

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: perhaps an 
```ruby
unless hash.keys.all? { |k| k.is_a? String }
```
Would do good

casz:
How can help, move this forward :)

casz:
What are the chances for this to be the next release?

casz:
#10670 why u no include? 😐

casz:
I still think the use case for this is valid and would be useful to all the Xamarin, instead of repeating the same set_info_plist_value(key: x, path: y) in a Fastfile

ohayon:
Hey @casz - I think this is an awesome feature to have. Is it possible you would add a test or two? Thanks!!! 🚀  🐙

casz:
I give it a shot 👍

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -10,7 +10,10 @@ def self.run(params)
...
+            plist = plist.merge!(stringified_hash)
```
casz: Should perhaps return plist to be able to validate test? And some might want it to return the plist? To further update the map?

casz: @ohayon WDYT?

casz:
I realized my solution is naive, it doesn't handle multi hashes 👎

casz:
I wanted to use deep_merge to merge `dictionary` that is noted by plist as [{}] ruby object.
Was basically writing a reimplementation of deep merge at the end 👍 
Decided to find and replace others usages of deep merge.

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -45,16 +72,32 @@ def self.available_options
...
-                                       optional: false),
+                                       conflicting_options: [:map],
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :map,
+                                       env_name: "FL_SET_INFO_PLIST_PARAM_MAP",
+                                       description: "Map of keys and values in plist to be merged",
+                                       type: Hash,
+                                       conflicting_options: [:key, :subkey, :value],
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :replace,
```
casz: One might want to replace the entire `plist` with another.

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -29,13 +37,32 @@ def self.run(params)
...
 
+          return plist if params[:map]
           return params[:value]
         rescue => ex
           UI.error(ex)
           UI.user_error!("Unable to set value to plist file at '#{path}'")
         end
       end
 
+      def self.deep_stringify(obj)
```
casz: Don't want to fight `plist` usage of string but I do want to support dev using `symbols`

fastlane/spec/actions_specs/set_info_plist_value_spec.rb
```diff
@@ -45,12 +51,44 @@
...
+        allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
+        ret = Fastlane::FastFile.new.parse("lane :test do
+          set_info_plist_value(
+            path: '#{plist_path}',
+            map: {
+              CFBundleIdentifier: 'com.krausefx.app123',
+              CFBundleShortVersionString: '9.9.9',
+              UIRequiredDeviceCapabilities: ['arm64'],
+              CFBundleURLTypes: [{
+                'CFBundleURLName'=>'com.krausefx.app123',
```
casz: This is just to show we can handle the deep merging of `dictionary`

fastlane/spec/fixtures/plist/Info-expected.plist
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
```
casz: added a new plist since it was easier to validate against

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -155,7 +152,10 @@ def self.details
       # @!group Helper
       #####################################################
 
+      # rubocop:disable Metrics/PerceivedComplexity
```
casz: class had already reached `Perceived Complexity` as my run of `bundle exec rubocop` as it was saying 19 out 18.

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -236,21 +236,9 @@ def self.generate_slack_attachments(options)
         end
 
         # merge additional properties
-        deep_merge(slack_attachment, options[:attachment_properties])
-      end
-
-      # Adapted from https://stackoverflow.com/a/30225093/158525
-      def self.deep_merge(a, b)
```
casz: 🚀 we can get rid of this ugly hash/array deep merge

frameit/lib/frameit.rb
```diff
@@ -41,12 +41,3 @@ def self.frames_version
     return @frames_version
   end
 end
-
-# rubocop:disable all
-class ::Hash
-  def fastlane_deep_merge(second)
```
casz: 🚀 we can get rid of this ugly hash deep merge

jetersen (commented):
🤖

ohayon:
Hi @casz - we are trying to be super careful these days of introducing new dependencies into the codebase, is there any way to achieve this without adding `deep_merge` as a dependency?

casz:
@ohayon I would be but that would add another method that needs some kind of deep merge but in a different flavour.

Look at the already removed code that implemented two different deep merge.
One was nested hashes and the other was nested hashmap and array.
Mine would be a third with nested hashmap, nested array and nested array of hashmaps...

I feel like https://github.com/danielsdeleo/deep_merge has a good test base already
https://github.com/danielsdeleo/deep_merge/blob/master/test/test_deep_merge.rb

casz:
I feel dirty :(

```ruby
      # Adapted from https://stackoverflow.com/a/30225093/158525		
      def self.deep_merge(a, b)		
        merger = proc do |key, v1, v2|		
          Hash === v1 && Hash === v2 ?		
                 v1.merge(v2, &merger) : Array === v1 && Array === v2 ?		
                   array_deep_merge(v1, v2) : [:undefined, nil, :nil].include?(v2) ? v1 : v2		
        end		
        a.merge(b, &merger)
      end

      def self.array_deep_merge(source, dest)
        source_all_hashes = source.all? { |i| i.kind_of?(Hash) }
        dest_all_hashes = dest.all? { |i| i.kind_of?(Hash) }
        if source_all_hashes && dest_all_hashes
          # merge hashes in lists
          list = []
          dest.each_index do |i|
            list[i] = deep_merge(source[i] || {}, dest[i])
          end
          list += source[dest.count..-1] if source.count > dest.count
          dest = list
        else
          dest = source | dest
        end
        dest
      end
```

casz:
@ohayon There, no deep_merge dep 😭 

🐑 🎈

casz:
In any case, amazing what unit test can reveal, it's like opening a can of worms 🤣

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -5,12 +5,19 @@ module SharedValues
...
     class SetInfoPlistValueAction < Action
       def self.run(params)
-        require "plist"
+        require 'plist'
 
         begin
           path = File.expand_path(params[:path])
           plist = Plist.parse_xml(path)
-          if params[:subkey]
+          if params[:map]
```
ohayon: This if statement is getting a bit unruly. Could we break some of these things out into their own methods to not have so much nested code?

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -29,13 +36,66 @@ def self.run(params)
...
+        end
+        if obj.kind_of?(Array)
+          stringified_array ||= []
+          obj.each do |v|
+            stringified_array << deep_stringify(v)
+          end
+          return stringified_array
+        end
+        obj
+      end
```
ohayon: Thanks for not including the extra dependency, can we move these into their own class or maybe monkey patches of Array and/or Hash?

fastlane/lib/fastlane/actions/set_info_plist_value.rb
```diff
@@ -45,16 +105,32 @@ def self.available_options
...
+                                       type: Hash,
+                                       conflicting_options: [:key, :subkey, :value],
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :replace,
+                                       env_name: "FL_SET_INFO_PLIST_PARAM_REPLACE",
+                                       description: "Replace plist with map instead of merging",
+                                       is_string: false,
+                                       default_value: false,
+                                       conflicting_options: [:key, :subkey, :value],
+                                       optional: true),
```
ohayon: Should there be a verify block here to make sure that `:map` is passed in?

Also, sort of related, is there a better word than `map` as that is not a commonly used word in ruby nor iOS programming? Might make more sense as `hash` or `dictionary`? Curious if you have any thoughts about that?

fastlane/spec/fixtures/plist/Info-expected.plist
ohayon: How come?

casz: Did because `get_info_plist_value` action only returns the value, and I was lazy 😐 

I'll try and rewrite the test to use expected objects instead of an entire plist.

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: Not too familiar with Ruby, so unsure 🤔 
But sounds like a good idea.

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: I been thinking the same thing :|
To me hash sounds better since we are working with ruby objects on our end.

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: If we move it into it's own class we can reuse it on slack and frameit, and hopefully we won't break anything 👍

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: Could you give an example on how to write a verify_block to check another config_item. I cannot find it anywhere else in the code base...

fastlane/lib/fastlane/actions/set_info_plist_value.rb
casz: @ohayon any suggestion on what this util class would be called and where to place it? so we can reuse it

casz:
@ohayon any chance this can get unstuck 🙁

getaaron:
@casz there are a 2 minor merge conflicts here, would you be able to resolve them?

Would like to move the merging stuff into a common class, maybe `helper.rb`?

casz:
@getaaron I need to dig into it again. But I'll definitely give it a try 🤣

fastlane/lib/fastlane/actions/set_info_plist_value.rb
janpio: Some monkey patches for `String` are at `fastlane\fastlane_core\lib\fastlane_core\core_ext\string.rb` for example. Would probably make sense to move these over there or in a similar file that is included in a similar way.

janpio:
Hey @casz, I just tried to get a handle on this PR as it is the oldest one here at fastlane right now ;) A big problem of understanding all this are the many comments on the PR and on the code. Could you maybe go through them and see if they are all still relevant and maybe delete the ones that aren't? Also the PR description could use some work to make it easier to understand what problem this solves and how it does that exactly. Thanks. (Had a quick look at the conflicts as well, those should not really be a problem I think)

casz:
I still think the PR is relevant since it allows you to change plist in a simpler fashion.

Instead of multiple `set_info_plist_value` calls you can do it in one.

Instead of this
```
set_info_plist_value(
  path: "./Info.plist",
  key: "CFBundleIdentifier",
  value: "com.example.fastlane",
)
set_info_plist_value(
  path: "./Info.plist",
  key: "CFBundleShortVersionString",
  value: "1.1.1",
)
set_info_plist_value(
  path: "./Info.plist",
  key: "CFBundleVersion",
  value: "9999",
)
```

You can do this:
```
set_info_plist_value(
  path: "./Info.plist",
  hash: {
    CFBundleIdentifier: "com.example.fastlane",
    CFBundleShortVersionString: "1.1.1",
    CFBundleVersion: "9999",
  }
)
```

joshdholtz:
Heyyy 👋 I'm first wanted to say thanks for the time you spent on this contribution! I'm going to close this PR for now since bulk assignment is actually possible already using https://docs.fastlane.tools/actions/update_plist/#update_plist. We appreciate the time and effort though ❤️

PR #10515, opened 2017-10-05
Sets the installed profile name from sigh in an environment variable …

joshdholtz:
…and lane context

## Why
Because we used to need the `UUID` in the past for really ugly reasons but now Xcode likes the have the name so yolo

## Example Usage
```rb
sigh
profile_name = Actions.lane_context[SharedValues::SIGH_NAME]

fastlane_require 'xcodeproj'
project_path = File.absolute_path File.join(path, "MyProject.xcodeproj")
xcproj = Xcodeproj::Project.open(project_path)
xcproj.targets.each do |target|
  target.build_configuration_list.set_setting 'PROVISIONING_PROFILE_SPECIFIER', profile_name
end
xcproj.save
```

### DIFF
--- a/fastlane/lib/fastlane/actions/sigh.rb
+++ b/fastlane/lib/fastlane/actions/sigh.rb
@@ -5,6 +5,7 @@ module Fastlane
       SIGH_PROFILE_PATHS = :SIGH_PROFILE_PATHS
       SIGH_UDID = :SIGH_UDID # deprecated
       SIGH_UUID = :SIGH_UUID
+      SIGH_NAME = :SIGH_NAME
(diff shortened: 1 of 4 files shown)


jdee (approved):
:+1:

Only comment: There's a convenient method in xcodeproj that can make your example in the comment a little simpler. You can just do:

```Ruby
target.build_configuration_list.set_setting 'PROVISIONING_PROFILE_SPECIFIER', profile_name
```

That sets for all configurations.

joshdholtz:
@jdee That's so cool! Had no idea 😇

PR #10518, opened 2017-10-06
Show verbose information for `fastlane snapshot reset_simulators`

KrauseFx:
Show the command that is being run of `--verbose`


ohayon (approved):
🍠

PR #10519, opened 2017-10-06
Tell user about env variable for `fastlane snapshot reset_simulators`

KrauseFx:
Tell user about env variable for `fastlane snapshot reset_simulators


ohayon (approved):
💥

PR #10521, opened 2017-10-06
Fix an issue where new docs files/changes weren't committed

KrauseFx:
This would result in not running `git add`, so no new files were added

- Fix an issue where new docs (`.md`) files/changes weren't committed
- Add support for automatic copying and committing of custom assets for docs


ohayon (approved):
➕

PR #10523, opened 2017-10-06
Remove pre-mono-gem code of deployment pipeline

KrauseFx:
This code wasn't used any more after the mono-gem migration


ohayon (approved):
👴

PR #10531, opened 2017-10-06
Run docs tests when running fastlane tests also

KrauseFx:
We didn't generate new docs when running tests, but only ran the ones from `master` from `fastlane/docs`, that's not really useful, so now it's better 👍


janpio:
Now that the docs check is split into its own container, time needed to execute this is not a problem any more.

KrauseFx:
That's awesome @janpio, what do we have to do to get it running? I can't believe I never got a review on this one 😢

janpio:
Not sure. Probably just resolve the conflicts, and specify a bit more what actually changed here (instead of just moved).

PR #10537, opened 2017-10-07
Add verbose output for upload_dsym

taquitos:
be able to see what command is failing when verbose output is on

It's difficult to debug why this would fail if we can't get verbose output.

NOTE: This will spit out an actual api token, so maybe we should think about how to protect that?


ohayon (approved):
🐫

PR #10541, opened 2017-10-08
Fix waiting during network activity for iPhone X

### LINKED ISSUE #10539: [snapshot] NetworkActivityIndicator detection broken for iPhone X
### New Issue Checklist

### Issue Description

Since the iPhone X has a new status bar and a new network activity indicator, the current [SnapshotHelper.swift](https://github.com/fastlane/fastlane/blob/master/snapshot/lib/assets/SnapshotHelper.swift) won't be able to wait for the network indicator to disappear. It just thinks, that there is no network activity indicator, because it compares the frame of elements in the status bar to `CGSize(width: 10, height: 20)`. According to my explorations, the new network activity indicator has a width of `46.3333333333333` and a height of `2.66666666666667` in the simulator at 100% scale. (see attachments)

Sadly the elements in the new status bar don't have accessibility values set, so the size is still the only way to find the network activity indicator. I filed a radar regarding the accessibility: [rdar://34872220](http://openradar.appspot.com/34872220)

Nevertheless, I'll try to get my fix working for both status bar versions and then I'll create a PR.

#### Screenshot
screen shot 2017-10-07 at 23 58 26

#### Attachements
The attachments contain the output of `debugPrint(XCUIApplication())` for the iPhone 8 and iPhone X simulator as described in the linked radar.

[iPhone8_Simulator_debugPrint_Output.txt](https://github.com/fastlane/fastlane/files/1366464/iPhone8_Simulator_debugPrint_Output.txt)
[iPhoneX_Simulator_debugPrint_Output.txt](https://github.com/fastlane/fastlane/files/1366465/iPhoneX_Simulator_debugPrint_Output.txt)

ffittschen:
### Checklist

### Motivation and Context
This PR resolves #10539. 

As described in the issue, the current SnapshotHelper.swift will not wait for the network activity indicator to disappear, because the size of the indicator changed.

I tested my changes using Xcode 9 with an iPhone 8 and an iPhone X simulator. I tested both cases where the network activity indicator was active and inactive when `snapshot("screenshot name")` was called.

### Description
* I extended the `isLoadingIndicator` computed property to not only look for `CGSize(width: 10, height: 20)`, but also for the new one, which is oddly `CGSize(width: 46.3333333333333, height: 2.66666666666667)`.
* I also added a new computed property that helps to find the status bar, because on the iPhone X, the status bar is of type `other` and can't be found with `XCUIApplication().statusBars`.
* Finally, I replaced the while loop that calls `sleep(1)` while looking for a disappearing element of `CGSize(width: 10, height: 20)` with the new XCTWaiter that evaluates a predicate.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -19,17 +19,24 @@ import XCTest
 var deviceLanguage = ""
 var locale = ""
 
-@available(*, deprecated, message: "use setupSnapshot: instead")
-func setLanguage(_ app: XCUIApplication) {
-    setupSnapshot(app)
-}
-
 func setupSnapshot(_ app: XCUIApplication) {
     Snapshot.setupSnapshot(app)
 }
 
-func snapshot(_ name: String, waitForLoadingIndicator: Bool = true) {
-    Snapshot.snapshot(name, waitForLoadingIndicator: waitForLoadingIndicator)
+@available(*, deprecated, renamed: "snapshot(_:timeWaitingForIdle:)")
+func snapshot(_ name: String, waitForLoadingIndicator: Bool) {
+    if waitForLoadingIndicator {
+        Snapshot.snapshot(name)
+    } else {
+        Snapshot.snapshot(name, timeWaitingForIdle: 0)
+    }
+}
+
+/// - Parameters:
+///   - name: The name of the snapshot
+///   - timeout: Amount of seconds to wait until the network loading indicator disappears. Pass `0` if you don't want to wait.
+func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20) {
+    Snapshot.snapshot(name, timeWaitingForIdle: timeout)
 }
 
 enum SnapshotError: Error, CustomDebugStringConvertible {
@@ -116,9 +123,9 @@ open class Snapshot: NSObject {
         }
     }
 
-    open class func snapshot(_ name: String, waitForLoadingIndicator: Bool = true) {
-        if waitForLoadingIndicator {
-            waitForLoadingIndicatorToDisappear()
+    open class func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20) {
+        if timeout > 0 {
+            waitForLoadingIndicatorToDisappear(within: timeout)
         }
 
         print("snapshot: \(name)") // more information about this, check out https://github.com/fastlane/fastlane/tree/master/snapshot#how-does-it-work
@@ -140,17 +147,14 @@ open class Snapshot: NSObject {
         #endif
     }
 
-    class func waitForLoadingIndicatorToDisappear() {
+    class func waitForLoadingIndicatorToDisappear(within timeout: TimeInterval) {
         #if os(tvOS)
             return
         #endif
 
-        let query = XCUIApplication().statusBars.children(matching: .other).element(boundBy: 1).children(matching: .other)
-
-        while (0..<query.count).map({ query.element(boundBy: $0) }).contains(where: { $0.isLoadingIndicator }) {
-            sleep(1)
-            print("Waiting for loading indicator to disappear...")
-        }
+        let networkLoadingIndicator = XCUIApplication().otherElements.deviceStatusBars.networkLoadingIndicators.element
+        let networkLoadingIndicatorDisappeared = XCTNSPredicateExpectation(predicate: NSPredicate(format: "exists == false"), object: networkLoadingIndicator)
+        XCTWaiter.wait(for: [networkLoadingIndicatorDisappeared], timeout: timeout)
     }
 
     class func pathPrefix() throws -> URL? {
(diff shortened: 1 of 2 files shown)


ffittschen:
I signed it!

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -19,17 +19,24 @@ import XCTest
...
-}
-
 func setupSnapshot(_ app: XCUIApplication) {
     Snapshot.setupSnapshot(app)
 }
 
-func snapshot(_ name: String, waitForLoadingIndicator: Bool = true) {
-    Snapshot.snapshot(name, waitForLoadingIndicator: waitForLoadingIndicator)
+@available(*, deprecated, renamed: "snapshot(_:timeWaitingForIdle:)")
+func snapshot(_ name: String, waitForLoadingIndicator: Bool) {
```
getaaron: Even if we're deprecating this method,`waitForLoadingIndicator` should have the same default value to avoid changing behavior in existing code, right?

Also, I'm not 100% sure we should deprecate this. Seems like we could just keep both methods as two different interfaces.

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -140,17 +147,14 @@ open class Snapshot: NSObject {
...
 
-        let query = XCUIApplication().statusBars.children(matching: .other).element(boundBy: 1).children(matching: .other)
-
-        while (0..<query.count).map({ query.element(boundBy: $0) }).contains(where: { $0.isLoadingIndicator }) {
-            sleep(1)
-            print("Waiting for loading indicator to disappear...")
-        }
+        let networkLoadingIndicator = XCUIApplication().otherElements.deviceStatusBars.networkLoadingIndicators.element
+        let networkLoadingIndicatorDisappeared = XCTNSPredicateExpectation(predicate: NSPredicate(format: "exists == false"), object: networkLoadingIndicator)
+        XCTWaiter.wait(for: [networkLoadingIndicatorDisappeared], timeout: timeout)
```
getaaron: 👍

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -180,16 +184,65 @@ open class Snapshot: NSObject {
...
-            return false
+private extension XCUIElementAttributes {
+    var isNetworkLoadingIndicator: Bool {
+        let oldLoadingIndicatorSize = CGSize(width: 10, height: 20)
+        
+        let hasOldLoadingIndicatorSize = frame.size == oldLoadingIndicatorSize
+        let hasNewLoadingIndicatorSize = frame.size.width.isBetween(46, and: 47) && frame.size.height.isBetween(2, and: 3)
+        
+        return !hasWhiteListedIdentifier && (hasOldLoadingIndicatorSize || hasNewLoadingIndicatorSize)
+    }
```
getaaron: I think this might be a bit more readable with `guard` and `switch`, let me know what you think:

```swift
var isNetworkLoadingIndicator: Bool {
    guard !hasWhiteListedIdentifier else { return false }

    switch frame.size {
    case CGSize(width: 10, height: 20):
        // old loading indicator
        return true
    case _ where frame.size.width.isBetween(46, and: 47) && frame.size.height.isBetween(2, and: 3):
        // iPhone X loading indicator
        return true
    default:
        return false
    }
}
```

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -180,16 +184,65 @@ open class Snapshot: NSObject {
...
+            return true
         }
-        return self.frame.size == CGSize(width: 10, height: 20)
+        
+        let positionedAtTopLeft = frame.origin == .zero
+        let hasOldStatusBarSize = frame.size == CGSize(width: deviceWidth, height: 20)
+        let hasNewStatusBarSize = frame.size == CGSize(width: deviceWidth, height: 44)
+        
+        return positionedAtTopLeft && (hasOldStatusBarSize || hasNewStatusBarSize)
+    }
```
getaaron: I think this one could be simplified with `guard` as well:

```swift
func isStatusBar(_ deviceWidth: CGFloat) -> Bool {
    guard elementType != .statusBar else { return true }
    guard frame.origin == .zero else { return false }

    return [CGSize(width: deviceWidth, height: 20),
            CGSize(width: deviceWidth, height: 44)]
        .contains(frame.size)
}
```

snapshot/lib/assets/SnapshotHelper.swift
ffittschen: If there are two methods with `_ name: String` as first parameter and a (different) second parameter which both have a default value, calling `snapshot("test")` will result in an error stating "Ambiguous use of 'snapshot'". That's why I removed this default value.

The behavior for people who just call `snapshot("test")` will not change, it will still wait. But only the default value of 20 seconds instead of indefinitely.

Yes we could keep it. It just seemed cleaner to me to have just one interface. Shall I remove the deprecation?

snapshot/lib/assets/SnapshotHelper.swift
ffittschen: I agree with the early exit through a `guard` or `if`, then I don't even need to negate `hasWhiteListedIdentifier`. That's a good catch

But the switch has a cyclomatic complexity of 4 and needs comments, whereas when I write 
```swift
let hasOldLoadingIndicatorSize = frame.size == CGSize(width: 10, height: 20)
let hasNewLoadingIndicatorSize = frame.size.width.isBetween(46, and: 47) && frame.size.height.isBetween(2, and: 3)

return hasOldLoadingIndicatorSize || hasNewLoadingIndicatorSize
```
it has a cyclomatic complexity of 2 and it doesn't need comments because of named constants ;)

snapshot/lib/assets/SnapshotHelper.swift
getaaron: Ah makes sense. I would leave it the way it is. Thanks for the explanation.

snapshot/lib/assets/SnapshotHelper.swift
getaaron: Fair :+1:

taquitos:
Love the change @ffittschen, question: were you able to test upgrading a project that uses `SnapshotHelperVersion [1.5]` to your version (1.6) for both Swift and ObjC? It looks like you addressed all the backwards compatibility issues, but you didn't mention if you tested this path.

ffittschen:
I actually used the `XCTWaiter` and timeout part of this PR for a few months in an edited version of SnapshotHelper 1.5 in one of my projects in production. But I can reset that one and upgrade to 1.6 to test the other parts.

I don't have any ObjC projects, but I'm sure I can find some example projects from pods to try it out.

taquitos (approved):
🎈:goat: This is really great, thank you for testing it out so thoroughly.

PR #10544, opened 2017-10-09
 update `:update_docs` commit and Slack message

janpio:
`:update_docs` does more than just updating `actions.md`


KrauseFx (approved):
Ha, nice catch, thanks @janpio 👍

PR #10551, opened 2017-10-09
Ci build number

SiarheiFedartsou:
### Checklist

### Motivation and Context
In each of our projects we have common code in lanes like this:
```
increment_build_number(build_number: ENV['BUILD_NUMBER'])
```
Currently we use Jenkins CI and it sets current build number to build number from our Jenkins job. But if we will decide to move to another CI system we will have to change it. I think it will be much better to have more or less CI independent way to get current build number from CI.
So, this pull request introduces new action `ci_build_number`, which returns current build number retrieved from CI using environment variables. It supports Jenkins, Travis CI, Circle CI, TeamCity, GoCD, Bamboo, Gitlab CI and Xcode Server. And also adds new option to `increment_build_number` action `use_ci_build_number` which allows to automatically set build number using value returned by `ci_build_number`.

### DIFF
--- a/fastlane/lib/fastlane/actions/ci_build_number.rb
+++ b/fastlane/lib/fastlane/actions/ci_build_number.rb
@@ -1,0 +1,64 @@
+module Fastlane
+  module Actions
+    class CiBuildNumberAction < Action
+      def self.run(params)
+        if ENV.key?('JENKINS_HOME') || ENV.key?('JENKINS_URL')
+          return ENV['BUILD_NUMBER']
+        end
+
+        if ENV.key?('TRAVIS')
+          return ENV['TRAVIS_BUILD_NUMBER']
+        end
+
+        if ENV.key?('CIRCLECI')
+          return ENV['CIRCLE_BUILD_NUM']
+        end
+
+        if ENV.key?('TEAMCITY_VERSION')
+          return ENV['BUILD_NUMBER']
+        end
+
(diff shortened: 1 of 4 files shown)


taquitos:
@SiarheiFedartsou thank you for your contribution! I've been thinking about this. One of the goals of `fastlane` is to make it so that you can call `fastlane` from anywhere, whether it is from ci, local dev, or wherever. So, for me, if you make an action that can only be used in ci, then it means you can only run certain lanes on ci. 

I think it makes more sense to me that you would create a specific plugin for this. Have you seen our documentation on how to do that? https://docs.fastlane.tools/plugins/create-plugin/

SiarheiFedartsou:
@taquitos Thanks for your response! At the beginning I was thinking about adding it to my plugin https://github.com/SiarheiFedartsou/fastlane-plugin-versioning . But then I decided to make PR to main repo because I thought that this is very common problem for fastlane users. 
So, I think it will be good update for my plugin :)
Thanks.

taquitos:
Your plugin is still sitting at number 1 on our plugins page 😄  https://docs.fastlane.tools/plugins/available-plugins/

SiarheiFedartsou:
And now it even better with new `ci_build_number` action  😄

lacostej:
I think this can be closed as the code was moved to the external plugin. Reopen if I was mistaken.

PR #10590, opened 2017-10-13
Keychain created with incorrect path

jetersen:
### Checklist

### Motivation and Context

Should use the standard set in sierra when creating keys.

Without this, keychain is literally called "name-db" and you get unexpected results 👍

### DIFF
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -12,7 +12,7 @@ module Fastlane
 
         if params[:name]
           escaped_name = params[:name].shellescape
-          keychain_path = "~/Library/Keychains/#{escaped_name}"
+          keychain_path = "~/Library/Keychains/#{escaped_name}.keychain-db"
         else
           keychain_path = params[:path].shellescape
         end


fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -12,7 +12,7 @@ def self.run(params)
 
         if params[:name]
           escaped_name = params[:name].shellescape
-          keychain_path = "~/Library/Keychains/#{escaped_name}"
+          keychain_path = "~/Library/Keychains/#{escaped_name}.keychain-db"
```
mpirri: Thanks for the submission @casz! A question - I assume this doesn't work on pre-Sierra OS's, right? Is it possible to add an OS check so that it does the right thing in case the user is running an older version of the OS? Thanks so much!

casz: It's a lot worse then initially thought :| The keychain is created as `~/Library/Keychains/` inside the `$HOME/Libray/Keychains/` when using `keychain_path` in the action. So it needs to be completely rewritten for High Sierra. And name was missing keychain-db

Issue #10200, opened 2017-08-30
Creating Pass Type ID Certificate


proorlov:
Hello!

I have created Pass Type ID and it has successfully created:

```
passbook = Spaceship::Portal.passbook.create!(bundle_id: 'pass.com.my.pb', name: 'Fastlane Passbook')
```

Now I try to create Pass Type ID Certificate:

```
[1] pry(#<Spaceship::Playground>)> csr, pkey = Spaceship.certificate.create_certificate_signing_request
=> [#<OpenSSL::X509::Request:0x007f8885868890>, #<OpenSSL::PKey::RSA:0x007f88858689a8>]

[2] pry(#<Spaceship::Playground>)> Spaceship.certificate.passbook.create!(csr: csr, bundle_id: "pass.com.my.pb")
RuntimeError: Could not find app with bundle id 'pass.com.my.pb'
from /Library/Ruby/Gems/2.0.0/gems/fastlane-2.54.4/spaceship/lib/spaceship/portal/certificate.rb:291:in `create!'
```
And it failed. It search for app not for passbook.
What do i wrong and how I should create Pass Type ID Certificate?
Thank you!

##### Complete output when running fastlane

```
mac:Develop user$ fastlane spaceship
/Library/Ruby/Gems/2.0.0/gems/fastimage-2.1.0/lib/fastimage.rb:67: warning: already initialized constant URI::DEFAULT_PARSER
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/uri/common.rb:545: warning: previous definition of DEFAULT_PARSER was here
[19:57:19]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
WARN: Unresolved specs during Gem::Specification.reset:
      terminal-table (< 2.0.0, >= 1.4.5)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
Username: xxx@gmail.com
Logging into to iTunes Connect (xxx@gmail.com)...
Successfully logged in to iTunes Connect

Logging into the Developer Portal (xxx@gmail.com)...
Successfully logged in to the Developer Portal

---------------------------------------
| Welcome to the spaceship playground |
---------------------------------------

Enter docs to open up the documentation
Enter exit to exit the spaceship playground
Enter _ to access the return value of the last executed command

Just enter the commands and confirm with Enter
[1] pry(#<Spaceship::Playground>)> 
```

[...]

PR #6943, opened 2016-11-10
[spaceship] Webpush Id support added for spaceship.

ilhanadiyaman:
Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes:

- [x] Run `bundle exec rspec` from the subdirectory of each tool you modified. Alternatively, run `rake test_all` from the root directory.
- [x] Run `bundle exec rubocop -a` to ensure the code style is valid
- [x] Read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] We currently don't accept new actions, please publish a plugin instead, more information in [Plugins.md](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Plugins.md)

Before submitting a pull request, we appreciate if you create an issue first to discuss the change :+1:

# API

## List all WebsitePush IDs:
returns array of Spaceship::Portal::WebsitePush
```ruby
Spaceship::Portal::WebsitePush.all
```
## Find by Website Push ID:
```ruby
website_push = Spaceship::Portal::WebsitePush.find("web.com.example.two")
```
## Create a WebsitePush ID:
```ruby
Spaceship::Portal::WebsitePush.create!(bundle_id: 'web.com.fastlane.example', name: 'Fastlane Website Push')
```
## Delete a WebsitePush ID:
```ruby
website_push = Spaceship::Portal::WebsitePush.find("web.com.example.two")
website_push.delete!
```

PR #10597, opened 2017-10-13
[pem] Create webpush certificates

jhanggi:
### Checklist

### Motivation and Context
Spaceship has [just about everything](https://github.com/fastlane/fastlane/pull/6943) it needs to support provisioning Web Push certificates, but you couldn't provision them because `Spaceship::App.all` does not include Web Push identifiers.

All that is really needed to make this work is to search `Spaceship::WebsitePush` instead when you're trying to provision a `Certificate::WebsitePush`. 

This should take care of https://github.com/fastlane/fastlane/pull/6943

Example usage of the PEM changes: `fastlane pem --website_push --app_identifier web.com.example.testing_app`

The addition of `specialIdentifierDisplayId` in the `submitCertificateRequest.action` request _might_ help with this issue: https://github.com/fastlane/fastlane/issues/10200 It probably won't completely address it, but hopefully it helps make some progress.

This was tested using `fastlane pem` as well as within my app, which also uses Spaceship to provision the push ID.

I'm open to suggestions on how to make this better and support more of the certificate types in the future.


jhanggi:
I signed it!

janpio:
Hey @jhanggi, sorry that you didn't get _any_ reaction until now. 

I don't know anything about Webpush, is this for this?
https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/NotificationProgrammingGuideForWebsites/PushNotifications/PushNotifications.html#//apple_ref/doc/uid/TP40013225-CH3-SW1

It would be really nice if you could take care of the merge conflicts and merge onto current `master`. This might unfortunately be a bit of work as during that timeframe we changed our coding style guide so that all method calls use parentheses - hope this is not too much work.

jhanggi:
That's exactly what it's for. It's for sending push notifications to Safari. 

Most of the portal pieces were already there in `spaceship/lib/spaceship/portal/webpush.rb`. You could already create the webpush id, just not the certificate. For the cert, the cli options and the fact that the parameter for the id is called `specialIdentifierId` rather than `apiIdId`  were the main things missing.

I've merged in from latest `master`. I could rebase if you'd prefer (this PR is so old, the actual commits might get lost in distant history). 

The only significant conflict was the move of the extra namespace in `Spaceship::App` to `Spaceship::Portal::App` (and then `Spaceship::WebsitePush` to `Spaceship::Portal::WebsitePush`.

Thanks for finally following up!

pem/lib/pem/options.rb
```diff
@@ -15,6 +15,11 @@ def self.available_options
                                      description: "Renew the development push certificate instead of the production one",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :website_push,
```
janpio: this new param conflicts with `development`, right? I think there is a way to note that.

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -289,7 +294,7 @@ def create!(csr: nil, bundle_id: nil)
 
           # look up the app_id by the bundle_id
           if bundle_id
-            app = Spaceship::Portal::App.set_client(client).find(bundle_id)
+            app = portal_type.set_client(client).find(bundle_id)
```
janpio: maybe I missed it in the code, but where does this come from when it is not a website_push? For that case you added the definition above in line 133 - but for production or development?

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -303,6 +308,12 @@ def create!(csr: nil, bundle_id: nil)
           response['certificateTypeDisplayId'] = response['certificateType']['certificateTypeDisplayId']
           self.new(response)
         end
+
+        # Default portal class to use when finding by bundle_id
```
janpio: maybe move this up so the definition is before the first usage of it?
(honestly don't know if this matters for the actual code, but I just wondered where this came from in line 297, wrote a comment, and then just found this here when scrolling down later... so having it up would have a benefit)

spaceship/spec/portal/portal_stubbing.rb
```diff
@@ -192,7 +192,7 @@ def adp_stub_certificates
       stub_request(:get, "https://developer.apple.com/services-account/QH65B2/account/ios/certificate/downloadCertificateContent.action?certificateId=XC5PH8DAAA&teamId=XXXXXXXXXX&type=R58UK2EAAA").
         to_return(status: 200, body: adp_read_binary_fixture_file('aps_development.cer')) # note: binary!
       stub_request(:post, "https://developer.apple.com/services-account/QH65B2/account/ios/certificate/submitCertificateRequest.action").
-        with(body: { "appIdId" => "2HNR359G63", "csrContent" => adp_read_fixture_file('certificateSigningRequest.certSigningRequest'), "type" => "BKLRAVXMGM", "teamId" => "XXXXXXXXXX" }).
```
janpio: is there no value in keeping the old test as well?

janpio (commented):
Nah, merging was just fine - same result. (We will squash and merge later to create a new commit with the changes.)

I added a few comments trying to understand how this works.

I will also try to get someone else to have a look at the code for a review.

How confident are you in the code in the current state? Tested locally, also the functionality it touches (production and development push certificates)?

joshdholtz:
Would like some 👀 on my two latest commits 😊 I tested this with both web push, dev, and dist certs and things were 👌

getaaron (approved):

janpio (approved):
Looks good!

janpio:
There is a report that this is causing issues in the Swift variation of Fastlane: https://github.com/fastlane/fastlane/issues/14427

PR #10601, opened 2017-10-15
Resign info variable

maschall:
### Checklist

### Motivation and Context
I noticed a few variables that could be refactored for different plists.  

Additionally I wanted to refactor how we set plist keys, because I had plans for a future feature to extend the ability to change any key in the Info.plist (example: Facebook App IDs, Adobe Analytics keys, environment variables if that's your thing)

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -458,13 +458,14 @@ function resign {
 
     # Replace the embedded mobile provisioning profile
     log "Validating the new provisioning profile: $NEW_PROVISION"
-    security cms -D -i "$NEW_PROVISION" > "$TEMP_DIR/profile.plist"
+    PROFILE_PLIST="$TEMP_DIR/profile.plist"
+    security cms -D -i "$NEW_PROVISION" > "$PROFILE_PLIST"
     checkStatus
 
-    APP_IDENTIFIER_PREFIX=$(PlistBuddy -c "Print :Entitlements:application-identifier" "$TEMP_DIR/profile.plist" | grep -E '^[A-Z0-9]*' -o | tr -d '\n')
+    APP_IDENTIFIER_PREFIX=$(PlistBuddy -c "Print :Entitlements:application-identifier" "$PROFILE_PLIST" | grep -E '^[A-Z0-9]*' -o | tr -d '\n')
     if [ "$APP_IDENTIFIER_PREFIX" == "" ];
     then
-        APP_IDENTIFIER_PREFIX=$(PlistBuddy -c "Print :ApplicationIdentifierPrefix:0" "$TEMP_DIR/profile.plist")
+        APP_IDENTIFIER_PREFIX=$(PlistBuddy -c "Print :ApplicationIdentifierPrefix:0" "$PROFILE_PLIST")
         if [ "$APP_IDENTIFIER_PREFIX" == "" ];
         then
             error "Failed to extract any app identifier prefix from '$NEW_PROVISION'"
@@ -478,10 +479,10 @@ function resign {
     # Set new app identifer prefix if such entry exists in plist file
     PlistBuddy -c "Set :AppIdentifierPrefix $APP_IDENTIFIER_PREFIX." "$INFO_PLIST" 2>/dev/null
 
-    TEAM_IDENTIFIER=$(PlistBuddy -c "Print :Entitlements:com.apple.developer.team-identifier" "$TEMP_DIR/profile.plist" | tr -d '\n')
+    TEAM_IDENTIFIER=$(PlistBuddy -c "Print :Entitlements:com.apple.developer.team-identifier" "$PROFILE_PLIST" | tr -d '\n')
     if [ "$TEAM_IDENTIFIER" == "" ];
     then
-        TEAM_IDENTIFIER=$(PlistBuddy -c "Print :TeamIdentifier:0" "$TEMP_DIR/profile.plist")
+        TEAM_IDENTIFIER=$(PlistBuddy -c "Print :TeamIdentifier:0" "$PROFILE_PLIST")
         if [ "$TEAM_IDENTIFIER" == "" ];
         then
             warning "Failed to extract team identifier from '$NEW_PROVISION', resigned ipa may fail on iOS 8 and higher"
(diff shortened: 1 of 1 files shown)


sigh/lib/assets/resign.sh
```diff
@@ -499,43 +515,13 @@ function resign {
...
 
-    # Update bundle version in the Info.plist if provided
-    if [[ -n "$BUNDLE_VERSION" ]];
-    then
-        CURRENT_VALUE="$(PlistBuddy -c "Print :CFBundleVersion" "$APP_PATH/Info.plist")"
-        # Even if the old value is same - just update, less code, less debugging
-        log "Updating the bundle version (CFBundleVersion) from '$CURRENT_VALUE' to '$BUNDLE_VERSION'"
-        PlistBuddy -c "Set :CFBundleVersion $BUNDLE_VERSION" "$APP_PATH/Info.plist"
-    fi
+    change_plist_value "$INFO_PLIST" "CFBundleVersion" "$SHORT_VERSION"
```
maschall: This may look a little weird, to set after setting before, but the validation steps in the beginning of the script will restrict that VERSION_NUMBER will not be set if SHORT_VERSION and BUNDLE_VERSION are set.

sigh/lib/assets/resign.sh
```diff
@@ -499,43 +515,13 @@ function resign {
     # Replace embedded provisioning profile with new file
     cp -f "$NEW_PROVISION" "$APP_PATH/embedded.mobileprovision"
 
-    #if the current bundle identifier is different from the new one in the provisioning profile, then change it.
```
janpio: could we keep this (and the follow 2) comments so it is clear what these new `change_plist_value` lines do?

sigh/lib/assets/resign.sh
```diff
@@ -721,11 +707,9 @@ function resign {
         done
 
         # Replace old bundle ID with new bundle ID in patched entitlements
-        # Read old bundle ID from the old Info.plist which was saved for this purpose
-        OLD_BUNDLE_ID="$(PlistBuddy -c "Print :CFBundleIdentifier" "$TEMP_DIR/oldInfo.plist")"
```
janpio: any idea why this was done this way? this PR now just uses `CURRENT_BUNDLE_IDENTIFIER` further down instead of this "copy" - is this safe?

sigh/lib/assets/resign.sh
```diff
@@ -447,25 +470,18 @@ function resign {
...
-            log "Changing display name from '$CURRENT_NAME' to '$DISPLAY_NAME'"
-            PlistBuddy -c "Set :CFBundleDisplayName $DISPLAY_NAME" "$APP_PATH/Info.plist"
-        fi
-    fi
+    change_plist_value "$INFO_PLIST" "CFBundleDisplayName" "$DISPLAY_NAME"
 
     # Replace the embedded mobile provisioning profile
     log "Validating the new provisioning profile: $NEW_PROVISION"
-    security cms -D -i "$NEW_PROVISION" > "$TEMP_DIR/profile.plist"
+    PROFILE_PLIST="$TEMP_DIR/profile.plist"
```
janpio: no `local` as with the other vars? (I have no idea of bash...)

sigh/lib/assets/resign.sh
```diff
@@ -380,6 +380,30 @@ function add_provision {
...
+}
+
+# changes key in plist
+# change_plist_key $PLIST $KEY $VALUE
+function change_plist_value {
+  local PLIST=$1
+  local KEY=$2
+  local VALUE=$3
+  
+  if [[ -n "$VALUE" ]];
```
janpio: is this identical to `if [ "${VALUE}" != "" ];`?

sigh/lib/assets/resign.sh
```diff
@@ -499,43 +515,13 @@ function resign {
...
-        log "Updating the bundle identifier from '$CURRENT_BUNDLE_IDENTIFIER' to '$BUNDLE_IDENTIFIER'"
-        PlistBuddy -c "Set :CFBundleIdentifier $BUNDLE_IDENTIFIER" "$APP_PATH/Info.plist"
-        checkStatus
-    fi
-
-    # Update the version number properties in the Info.plist if a version number has been provided
-    if [ "$VERSION_NUMBER" != "" ];
-    then
-        CURRENT_VERSION_NUMBER=$(PlistBuddy -c "Print :CFBundleVersion" "$APP_PATH/Info.plist")
-        if [ "$VERSION_NUMBER" != "$CURRENT_VERSION_NUMBER" ];
```
janpio: did you intentionally remove this equality (old vs. new) check for all of them?

janpio (commented):
Wow, so first bash script I have ever reviewed ;)

First part of the PR is extracting some values into variables, second one is to extract the "change value in plist" thing into its own method. Correct?

If so, this looks good - I just left some comments for some minor details (mostly me now knowing bash I guess...).

Is this file and its functionality fully covered by tests?
Did you test this manually?

sigh/lib/assets/resign.sh
```diff
@@ -380,6 +380,30 @@ function add_provision {
...
+
+# changes key in plist
+# change_plist_key $PLIST $KEY $VALUE
+function change_plist_value {
+  local PLIST=$1
+  local KEY=$2
+  local VALUE=$3
+  
+  if [[ -n "$VALUE" ]];
+  then
```
lyndsey-ferguson: Nit: can we put the `then` on the same line as the `if` to keep consistency?

sigh/lib/assets/resign.sh
lyndsey-ferguson: I would stick with `local` as this variable is not used outside of the function. At my company, we keep local variables lowercase to make it more obvious.

sigh/lib/assets/resign.sh
```diff
@@ -380,6 +380,30 @@ function add_provision {
...
+      CURRENT_VALUE="$(plist_value "$PLIST" $KEY)"
+      log "Updating the $KEY from '$CURRENT_VALUE' to '$VALUE'"
+      PlistBuddy -c "Set :$KEY $VALUE" "$PLIST"
```
lyndsey-ferguson: What do you think of *not* setting the `$VALUE` if it is equal to `$CURRENT_VALUE`? Saves a small bit of time.

sigh/lib/assets/resign.sh
lyndsey-ferguson: I suppose there is not harm in setting the value if it is the same as what is there. If my suggestion to not set it if it is already that value in the `change_plist_value` function is accepted, then this is fine.

lyndsey-ferguson (changes requested):
Left some comments.

joshdholtz:
Heyyy 👋 Same thing here! Thanks for the time spent on making this PR! This PR seems to have gone stale and `resigh.sh` has changed a lot since then. Feel free to let me know if you want to reopen this or create a new PR for this and we can have another look!

PR #10563, opened 2017-10-10
Migrate config to 2.0

endocrimes:
This PR migrates the existing configuration to CircleCI 2.0, in a follow up PR, I would like to break out the linting and integration steps into their own jobs. In doing so we can probably ~half CI run time.

PR #10603, opened 2017-10-15
Convert plugin template to Circle CI 2.0

jdee:
### Checklist

### Motivation and Context

Related to #10563. This updates the plugin template to use Circle CI 2.0, including test reporting.

### Description

Replaced circle.yml with .circleci/config.yml.

To enable test reporting, added `rspec_junit_formatter` as a dev dependency. RSpec now generates both doc-style output and a JUnit-style report.xml. The latter is used by CI for test reporting.

I played around with a number of options until I found all the right info in various places in the Circle docs to produce results like this:

screen shot 2017-10-15 at 9 21 35 am

https://circleci.com/docs/2.0/configuration-reference/#store_test_results

Linked from that page ("classic CircleCI test metadata directory layout"):

https://circleci.com/docs/1.0/test-metadata/
https://circleci.com/docs/1.0/test-metadata/#rspec

### DIFF
--- a/fastlane/lib/fastlane/plugins/template/.circleci/config.yml
+++ b/fastlane/lib/fastlane/plugins/template/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "Gemfile.lock" }}
+          - v1-dependencies-{{ checksum "Gemfile" }}
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
@@ -29,7 +29,7 @@ jobs:
       - save_cache:
           paths:
             - ./venv
-          key: v1-dependencies-{{ checksum "Gemfile.lock" }}
+          key: v1-dependencies-{{ checksum "Gemfile" }}
 
       # run tests!
       - run:
--- a/fastlane/lib/fastlane/plugins/template/.circleci/config.yml
+++ b/fastlane/lib/fastlane/plugins/template/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
 
       - run:
           name: install dependencies
-          command: bundle install --jobs=4 --retry=3 --path vendor/bundle
+          command: bundle check || bundle install --jobs=4 --retry=3 --path vendor/bundle
 
       - save_cache:
           paths:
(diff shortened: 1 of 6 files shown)


KrauseFx:
Awesome, thanks for your contribution on this @jdee, awesome idea. Let's see what @DanToml says about it :)

jdee:
One thought I had after submitting this: There's also Travis. That is, there's also a .travis.yml in the template. So devs have the option of not using what's here, and they may not use the `rspec_junit_formatter` gem. Of course, this is just a template, and they can modify whatever they like afterward. (They probably won't use both the .travis.yml and .circleci/config.yml, but they're both present, and one or both will probably be deleted.) It's nice for things to be set up well by default. And of course test reporting can also be set up for Travis. I think JUnit-style reporting is supported by most CI systems. I've used this gem with Jenkins before. So I don't think there's any harm including this dependency by default. But open to other opinions.

fastlane/lib/fastlane/plugins/template/.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "Gemfile" }}
+          # fallback to using the latest cache if no exact match is found
+          - v1-dependencies-
+
+      - run:
+          name: install dependencies
+          command: |
+            bundle install --jobs=4 --retry=3 --path vendor/bundle
```
DanToml: this should be `command: bundle check || bundle install --jobs=4 --retry=3 --path vendor/bundle`

fastlane/lib/fastlane/plugins/template/.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - save_cache:
+          paths:
+            - ./venv
```
DanToml: I think we want the vendor folder here

fastlane/lib/fastlane/plugins/template/.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+          key: v1-dependencies-{{ checksum "Gemfile" }}
+
+      # run tests!
+      - run:
+          name: run tests
+          command: |
```
DanToml: no need for multi line command here either

endocrimes (changes requested):
Good start!

jdee:
@dantoml Thanks. I think I addressed your concerns, particularly `bundle check` before `bundle install` and caching the vendor folder. (I think I was working from an old Circle template and never noticed that. All my gems thank you for catching it.)

endocrimes (approved):

PR #10611, opened 2017-10-16
[sigh] add option to not create a profile if none exists

maschall:
### Checklist

### Motivation and Context
Our team uses manually created provision profiles and periodically they become "Invalid" in the developer portal.  When this happens, `sigh` is automatically creating new profiles, but we'd prefer it to just fail.

### Description
I've added a `--skip-profile-creation` flag and an error message if no profiles are found in the developer portal.

I did not add a short-option for this, because I don't think it would be a normal operation for majority of `sigh` users.


janpio (approved):
Sorry this is only now getting attention, looks like a useful addition (although clearly a not so common use case). I would prefer another review though before merging.

sigh/lib/sigh/runner.rb
```diff
@@ -29,6 +29,8 @@ def run
           profile.delete!
           profile = create_profile!
         end
+      elsif Sigh.config[:skip_profile_creation]
+        UI.user_error!("No existing profiles found, that match the certificates you have installed locally!")
```
janpio: Maybe add an indication in the error message that this is ok as indicated with a param/option?

getaaron:
@maschall @janpio Can't this be accomplished with the sigh `readonly` option?

getaaron:
Looks like `readonly` wasn't available when this PR was initially written. So I'm closing this since that option is now available.

PR #10612, opened 2017-10-17
Import from git versioning

### LINKED ISSUE #10189: import_from_git versioning
### New Issue Checklist

### Issue Description

We're using Fastlane at our company for a lot of iOS projects to build and deliver our apps.
To make it all maintainable, we're using a hosted Fastfile that has everything in place to build the apps.
We make use of `import_from_git` to do that. This had lots of benefits, for example: pushing an update to the hosted Fastfile repo had direct impact to all our project. All changes to it have to be compatible with the projects of-course. 

Recently we started working on a new, better version of our hosted Fastfile, but we want to avoid problems after finishing up that release. Once merged, all projects should adapt the new Fastfile immediately. But at our scale, that's not an option. 

So it would be nice for us if we could use some kind of versioning with the import function.

For example:
`import_from_git(url: 'git@example.com:project.git', version:'1.*')`
or  
`import_from_git(url: 'git@example.com:project.git', version:'~> 1.0')`
or
`import_from_git(url: 'git@example.com:project.git', tag:'1.*')`
etc.

I'm aware that we can use git tags, but it's currently not possible allow new versions being used within a specific version range.

Is this something we can add to Fastlane, or are there any other/better ideas to fix this problem?

Thanks

bramseynhaeve:
### Checklist

### Motivation and Context

We're using the import_from_git method to use a shared Fastfile in our company. One Fastlane file to manage all our different projects. All changes to this file were always backwards compatible, but we wanted to create a new and better one that isn't backwards compatible anymore. To prevent all our builds from failing, we should be able to select a version range when importing the shared Fastfile.

This fixes https://github.com/fastlane/fastlane/issues/10189

We tested this by adding the version parameter in one of our projects, and using a local version of Fastlane by modifying the Gemfile of our iOS project.
```
import_from_git(url: 'git@bitbucket.org:------------/----------.git', version: '~> 2.0')
```

### Description
To solve this, we updated the already implemented `import_from_git` function. We added a new parameter `version` next to `branch`. Both do almost the same, except that version will use all git tags to find a matching version. When a version is found, it will checkout just like you would have used a branch.

Optimistic parameters are added to support auto updates with backwards compatible code.


bramseynhaeve:
I signed it!

fastlane/lib/fastlane/fast_file.rb
```diff
@@ -235,26 +236,72 @@ def import_from_git(url: nil, branch: 'HEAD', path: 'fastlane/Fastfile')
...
+              matching_git_tags = git_tags.select do |tag|
+                tag == version_number
+              end
+
+              UI.user_error!("The specified version \"#{version_number}\" doesn't exist") if matching_git_tags.count == 0
+              checkout_param = matching_git_tags.last
+            else
+              UI.user_error!("The specified operator \"#{operator}\" in \"#{version}\" is unknown. Please use one of these '~> ->'")
+            end
+          end
```
ohayon: Any chance we could pull this into it's own method? Might be nice to just not have so much nested code right here in line.

ohayon (changes requested):
Hey @bramseynhaeve - this is a really cool use case, idea, and PR! I just have one comment in the review, let me know your thoughts!

bramseynhaeve:
Thanks @ohayon for the nice words and feedback! 
Sure, I moved the logic to separate functions to solve your PR remark.

ohayon:
Thanks @bramseynhaeve ! That looks great 🌟 ! I'm not sure those failing tests are related to your code, would you mind just taking a look at those to make sure that thats the case?

bramseynhaeve:
Thanks. I'm not having these failed test local and they showed up after I merged master into my PR. But I'll take a look.

ohayon:
Thanks @bramseynhaeve! Also, we are wondering if we could tweet about this change. We think this change is really awesome!

bramseynhaeve:
@ohayon A tweet would be nice, go ahead! A mention would be awesome. ;-)

friederbluemle:
Thanks @bramseynhaeve This looks very useful! 👍

ohayon (approved):

ohayon:
Thanks again for this @bramseynhaeve ! We are going to tweet about it at some point soon 🐦 ! This is an awesome first-time contribution! We hope to see more from you!

bramseynhaeve:
🙌🏻

lacostej:
@bramseynhaeve super feature! We were discussing the need to add this when I found out you already implemented it. Thanks a lot! 🍻

PR #10619, opened 2017-10-17
Forward match keychain password into cert

joshdholtz:
Not forwarding the password was not allowing `security set-key-partition-list` to fail with my keychain name/password combo since the password was always empty string https://github.com/fastlane/fastlane/blob/1ad696f360b937e1881b5ca63e340c86ebe61cf8/fastlane_core/lib/fastlane_core/keychain_importer.rb#L16-L24

Using this in a way to prevent the "codesign allow" popup from showing when importing a new cert with a custom keychain but not passing in the password was preventing this from actually happening.

### DIFF
--- a/match/lib/match/generator.rb
+++ b/match/lib/match/generator.rb
@@ -11,7 +11,8 @@ module Match
         force: true, # we don't need a certificate without its private key, we only care about a new certificate
         username: params[:username],
         team_id: params[:team_id],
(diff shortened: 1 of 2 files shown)


joshdholtz:
I fixed the tests and pushed my commit but not seeing it come through 😔 Github probably having issues

mpirri (approved):
Awesome!  💯

PR #10636, opened 2017-10-18
Adds new precheck rule to check for iPhone X.

farkasseb:
### Checklist
  - I've run it but there are some failures with `AnalyticsSession` that seems completely unrelated to this change.

### Motivation and Context
Tonight we had a rejection with an app of ours with the following reason.

> Please remove all references to pre-release versions of Apple software and hardware from your app and its metadata.
>
> Specifically, your metadata states “Improved iPhone X support.”

This PR aims to add a new rule which could be updated regularly with pre-release Apple software and hardware names.

### Description
I've added a new rule subclassing the existing `AbstractTextMatchRule`. I saw that tests already exist for `AbstractTextMatchRule` and there are no separate test for `TestWordsRule`, so I didn't write any new ones.

### DIFF
--- a/precheck/lib/assets/PrecheckfileTemplate
+++ b/precheck/lib/assets/PrecheckfileTemplate
@@ -25,4 +25,5 @@
 # test_words(level: :error)
 # unreachable_urls(level: :error)
 # custom_text(data: ["fabric"], level: :warn)
+# pre_release_apple_sw_hw(level: warn)
 
--- a/precheck/lib/precheck/options.rb
+++ b/precheck/lib/precheck/options.rb
@@ -15,7 +15,8 @@ module Precheck
         FreeStuffIAPRule,
         CustomTextRule,
         CopyrightDateRule,
-        UnreachableURLRule
+        UnreachableURLRule,
+        PreReleaseAppleSoftwareHardware
       ].map(&:new)
     end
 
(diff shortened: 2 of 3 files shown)


KrauseFx:
Interesting, thanks for the PR, this is great. The only thing is that the iPhone X comes out rather soon, so the question is when they'll allow mentioning the name? We could disable that rule automatically based on the dates? What do you think?

farkasseb:
Seems like an excellent idea. Do you have a plan or pointer in mind or is there anything similar in the current codebase? If not I will come up with something on my own.

taquitos:
@farkasseb, @KrauseFx, this makes me think we should consider having “temporary” precheck rules. Maybe we can host a static file here in the repo that contains an array of words that will get you rejected, and during precheck runs, we can check on the `master` branch for any updates. That way we don't have to wait for _fastlane_ releases to update it.

farkasseb:
Well here is an interesting update for [ESPN](https://itunes.apple.com/us/app/espn-live-sports-scores/id317469184?mt=8).

> [+] Optimized for iOS 11 and the new iPhone X to have us running as smoothly as Saquon Barkley through the hole

It seems to be highly dependent on the reviewer person.

So what should be the conclusion? Should I try to implement a "temporary" rule based on @taquitos's idea? Do you have any pointers for me?

janpio:
Hey @farkasseb, sorry this took so long to get any response.

I think we all agree that adding these strings as static checks is not the best idea, as it also depends on the user updating their fastlane installation to unblock after a release.

But the general idea of having a way to save against using a list of words, for whatever reason, isn't a bad one. Maybe look into creating a new PR for a check that gets the blocked words from a text file? Then the user themselves of course would have to maintain the list, but the check would do the actual work of checking.

PR #10621, opened 2017-10-17
[BuildWatcher] Handle parse error while waiting for build processing

UnsafePointer:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

This PR should fix the issue described in https://github.com/fastlane/fastlane/issues/10097 and https://github.com/fastlane/fastlane/issues/10420.

### Description

The suggested approach by @thibaudrobelain in https://github.com/fastlane/fastlane/issues/10097#issuecomment-334335475 might be too generic. As far as I can tell, this only occurs during the loop in [`Build_watcher#wait_for_build_processing_to_be_complete `](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/build_watcher.rb#L10-L19), so I added a rescue statement inside the loop that should handle this error up to 10 times, as suggested by @mpirri.

I'm also happy to move this into [`Client#with_retry`](https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L528-L531).

Issue #10420, opened 2017-09-26
Hit HTML parsing error in Pilot


danieljvdm:
During the following Pilot output: `Waiting for iTunes Connect to finish processing the new build` this crash happened below. 

```
.fastlane/bin/bundle/lib/ruby/2.2.0/json/common.rb:155:in `parse': [!] 776: unexpected token at '<!DOCTYPE html><html lang="en"> (Faraday::ParsingError)
<head>
<title></title>
<link rel="stylesheet" href="/styles/main.css">
<style type='text/css'>
    /*include ../../dist/styles/main.css*/
</style>
</head>

<body><div id="wrapper" class="j42">
<h2>Internal Server Error</h2></div>

<div class="footer">
    <p>Copyright &copy; <span id='year'>2015</span> Apple Inc. All rights reserved.</p>
    <ul><li><a href="http://www.apple.com/legal/internet-services/itunes/us/terms.html">Terms of Service</a></li><li><a href="http://www.apple.com/legal/privacy">Privacy Policy</a></li><li><a href="http://www.apple.com/support/">Contact Us</a></li></ul>
</div>


<script type='text/javascript'>(function() {
    var isIos = navigator.userAgent.match(/(iPod|iPhone|iPad)/);
    document.body.setAttribute( 'class', (!isIos) ? 'otherBrowser' : 'isMobileSafari' )
    var yearEl = document.getElementById('year');
    if (yearEl) { yearEl.textContent = new Date().getFullYear(); }
})();
</script>

</body>
</html>
```

jasonhocker:
I too have the same issue.  Here is my env

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.6                                     |
| Ruby                        | 2.2.4                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.2.1                           |
| Installation Source         | ~/.fastlane/bin/bundle/bin/fastlane         |
| Host                        | Mac OS X 10.12.6 (16G29)                    |
| Ruby Lib Dir                | ~/.fastlane/bin/bundle/lib                  |
| OpenSSL Version             | OpenSSL 1.0.2g  1 Mar 2016                  |
| Is contained                | false                                       |
| Is homebrew                 | true                                        |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
[...]

Issue #10097, opened 2017-08-21
[pilot] Faraday::ParsingError: [!] 765: unexpected token at '<!DOCTYPE html><html lang="en">


Sega-Zero:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Just caught build failure when building for Testflight. The build went fine, ipa uploaded succesfully, but then pilot failed while waiting for build to appear in itunes connect. When I opened itunesconnect web interface, I see the processing build fine.

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```bash
[20:35:11]: Successfully exported and compressed dSYM file
[20:35:11]: Successfully exported and signed the ipa file:
[20:35:11]: /Users/travis/build/company/iOS/Project/Project.ipa
[20:35:13]: -------------------------
[20:35:13]: --- Step: crashlytics ---
[20:35:13]: -------------------------
[20:35:14]: Uploading the build to Crashlytics Beta. Time for some ☕️.
[20:44:08]: iTunes Transporter successfully finished its job
[20:44:08]: --------------------------------------------------------------------
[20:44:08]: Successfully uploaded package to iTunes Connect. It might take a few minutes until it's visible online.
[20:44:08]: --------------------------------------------------------------------
[20:44:08]: Successfully uploaded the new binary to iTunes Connect
[20:44:08]: If you want to skip waiting for the processing to be finished, use the `skip_waiting_for_build_processing` option
[20:44:23]: Waiting for iTunes Connect to finish processing the new build (1.11.1111 - 1.11.1111)
[20:44:54]: Waiting for iTunes Connect to finish processing the new build (1.11.1111 - 1.11.1111)
[20:45:24]: Waiting for iTunes Connect to finish processing the new build (1.11.1111 - 1.11.1111)
[20:45:55]: Waiting for iTunes Connect to finish processing the new build (1.11.1111 - 1.11.1111)
[20:46:25]: Waiting for iTunes Connect to finish processing the new build (1.11.1111 - 1.11.1111)
[...]

PR #10646, opened 2017-10-19
[Client] Handle Faraday::ParsingError as part of Client#with_retry logic

UnsafePointer:
### Checklist

### Motivation and Context

This is related:

https://github.com/fastlane/fastlane/issues/10097: [pilot] Faraday::ParsingError: [!] 765: unexpected token at `'<!DOCTYPE html><html lang="en">`
https://github.com/fastlane/fastlane/issues/10420: Hit HTML parsing error in Pilot
https://github.com/fastlane/fastlane/pull/10621: [BuildWatcher] Handle parse error while waiting for build processing 
https://github.com/fastlane/fastlane/issues/10088: Pilot fails after successful upload: "Could not set changelog"

### Description

After using the fix proposed in https://github.com/fastlane/fastlane/pull/10621 for a couple of days we ran into https://github.com/fastlane/fastlane/issues/10088, so I think a more general solution might be needed for this, as proposed in https://github.com/fastlane/fastlane/issues/10097#issuecomment-334335475.

The whole reason why this happens is because iTunes Connect is not respecting `Content-Type` and returning HTML in the response, so I'm also not sure if this is something fastlane should handle.

https://github.com/fastlane/fastlane/pull/10621 might not be needed anymore if we decide to handle the error with this instead, because using both fixed would mean retry a lot during `wait_for_build_processing_to_be_complete`.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -527,6 +527,7 @@ module Spaceship
     rescue \
         Faraday::Error::ConnectionFailed,
         Faraday::Error::TimeoutError,
+        Faraday::ParsingError, # <h2>Internal Server Error</h2> with content type json
         AppleTimeoutError,
         InternalServerError => ex # New Faraday version: Faraday::TimeoutError => ex
       tries -= 1
--- a/spaceship/spec/client_spec.rb
+++ b/spaceship/spec/client_spec.rb
@@ -42,7 +42,8 @@ describe Spaceship::Client do
   describe 'retry' do
     [
       Faraday::Error::TimeoutError,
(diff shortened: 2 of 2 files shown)


Ruenzuo:
@mpirri do you have any thoughts on this? I'm also happy to revert #10621 as part of this PR.

mpirri:
Whoa. This is amazing. Makes so much sense.  Do you mind making a PR to revert #10621? Then we can merge these together?

Thanks!  🚀

Ruenzuo:
@mpirri here: https://github.com/fastlane/fastlane/pull/10647

mpirri (approved):
🚀

PR #10650, opened 2017-10-20
Unify param loading

milch:
Mostly did some refactoring. Two main things: 

1. Decoupled the `Configuration` and `ConfigurationFile` classes so that after parsing all parameters that were successfully set are now available as a Hash in `ConfigurationFile`
2. Split the parameters in `Configuration` so that it is possible to differentiate between a parameter passed via CLI (in `values`) and one passed via the `ConfigurationFile` (in `config_file_options`)

This makes it easier and straight forward to model the actual parameter loading order from the [docs](https://docs.fastlane.tools/advanced/#priorities-of-parameters-and-options) in the `fetch` function - it's basically a 1:1 mapping between that list in the docs and the actual code now.

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -208,11 +208,11 @@ module FastlaneCore
       option = verify_options_key!(key)
 
       # Same order as https://docs.fastlane.tools/advanced/#priorities-of-parameters-and-options
-      value = if @values.key? key and !@values[key].nil?
+      value = if @values.key?(key) and !@values[key].nil?
                 @values[key]
               elsif option.env_name and !ENV[option.env_name].nil?
                 ENV[option.env_name].dup
-              elsif self.config_file_options.key? key
+              elsif self.config_file_options.key?(key)
                 self.config_file_options[key]
               else
                 option.default_value
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -14,6 +14,9 @@ module FastlaneCore
     # @return [String] The name of the configuration file (not the path). Optional!
     attr_accessor :config_file_name
 
+    # @return [Hash] Options that were set from a config file using load_configuration_file. Optional!
+    attr_accessor :config_file_options
+
     def self.create(available_options, values)
       UI.user_error!("values parameter must be a hash") unless values.kind_of?(Hash)
       v = values.dup
@@ -43,6 +46,7 @@ module FastlaneCore
     def initialize(available_options, values)
       self.available_options = available_options || []
       self.values = values || {}
+      self.config_file_options = {}
 
       # used for pushing and popping values to provide nesting configuration contexts
       @values_stack = []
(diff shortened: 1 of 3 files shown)


fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -171,8 +171,28 @@ def load_configuration_file(config_file_name = nil, block_for_missing = nil, ski
       return if paths.count == 0
 
       path = paths.first
-      configuration_file = ConfigurationFile.new(self, path, block_for_missing, skip_printing_values)
+      begin
```
milch: I implemented it this way so that it would pass the test [here](https://github.com/fastlane/fastlane/blob/master/fastlane_core/spec/configuration_file_spec.rb#L201)... I don't think it's actually a good idea though, since it basically crashes anyway after that point. Could any actual setups use this "feature" or could it be removed entirely?

fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -185,33 +205,22 @@ def load_configuration_file(config_file_name = nil, block_for_missing = nil, ski
...
-        # however we don't print the actual value, as it may contain sensitive information
-        # The user can easily find the actual value by print out the environment
-        UI.verbose("Taking value for '#{key}' from environment variable '#{option.env_name}'")
-
-        value = option.auto_convert_value(ENV[option.env_name].dup)
-        option.verify!(value) if value
-      end
-
-      value = option.default_value if value.nil?
       value = nil if value.nil? and !option.string? # by default boolean flags are false
```
milch: I have no idea what this line is supposed to do so I kept it in

fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -185,33 +205,22 @@ def load_configuration_file(config_file_name = nil, block_for_missing = nil, ski
     def fetch(key, ask: true)
       UI.user_error!("Key '#{key}' must be a symbol. Example :app_id.") unless key.kind_of?(Symbol)
 
-      option = option_for_key(key)
-      UI.user_error!("Could not find option for key :#{key}. Available keys: #{@available_options.collect(&:key).join(', ')}") unless option
+      option = verify_options_key!(key)
 
-      value = @values[key]
+      # Same order as https://docs.fastlane.tools/advanced/#priorities-of-parameters-and-options
+      value = if @values.key? key and !@values[key].nil?
```
jdee: This is not only more readable, but more easily testable as well. By storing the config. file state on the Configuration object, you can mock a config. file by just setting the @config_file_options in tests and not even mocking the ConfigurationFile class.

fastlane_core/lib/fastlane_core/configuration/configuration.rb
jdee: I think the answer is it does nothing, since it says, effectively, `value = nil if value == nil`. I wonder if the intention was `value = false if value.nil? and !option.string?`. Since Ruby has no Boolean type, you cannot specify `type: Boolean` for these flags, but usually specify `is_string: false`. So if the end result is nil, it would set them to false. As long as they're not strings. But this would then also happen for things like Array params (false if none specified), which seems dubious. You wouldn't be able to do `if params[:array_param].nil?` in an action if it were false instead of nil.

jdee (approved):
This looks good. Are there already tests for the order that weren't working right but are now? If this broke before, the order should probably be tested to keep it from happening again.

milch:
@jdee I've added tests to check that the parameter order is correct now

milch:
Also rebased from master so the tests pass again (hopefully)

jdee:
Hooray for tests! :shipit:

fastlane_core/lib/fastlane_core/configuration/configuration.rb
KrauseFx: Brackets for method calls please :)

KrauseFx (approved):
This looks alright, thanks for taking the time @milch. We definitely should make sure to monitor for any regressions when we roll it out

fastlane_core/lib/fastlane_core/configuration/configuration.rb
milch: done

Issue #10632, opened 2017-10-18
[Deliver] ["ITC.apps.idfa.errorMessage.agreeoptout"] missing in additional review options?


minuscorp:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When using Deliver to upload and submitting into review a build which uses IDFA identifier, it fails with
**Something wrong with your Ad ID information: ["ITC.apps.idfa.errorMessage.agreeoptout"].** because in the formulary is required to fill the agreement as a developer (a checkbox).

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
    [09:18:40]: Selecting build 1.1.7 (232.171018)...
    [09:18:42]: Successfully selected build
    [09:18:42]: Submitting the app for review...
    [09:19:09]: Setting 'add_id_info_tracks_install' to 'true'...
    [09:19:09]: Setting 'export_compliance_uses_encryption' to 'false'...
    [09:19:09]: Setting 'export_compliance_encryption_updated' to 'false'...
    [09:19:09]: Setting 'add_id_info_uses_idfa' to 'true'...
    [09:19:13]: Sending crash report...
    [09:19:13]: The stack trace is sanitized so no personal information is sent.
    [09:19:13]: To see what we are sending, look here: /Users/buddybuild/.fastlane/latest_crash.json
    [09:19:13]: Learn more at https://github.com/fastlane/fastlane#crash-reporting
    [09:19:13]: You can disable crash reporting by adding `opt_out_crash_reporting` at the top of your Fastfile
    +------------------+----------------+
    |           Lane Context            |
    +------------------+----------------+
    | DEFAULT_PLATFORM | ios            |
    | PLATFORM_NAME    | ios            |
    | LANE_NAME        | ios buddybuild |
    +------------------+----------------+
    [09:19:13]: Something wrong with your Ad ID information: ["ITC.apps.idfa.errorMessage.agreeoptout"].
    [09:19:13]: Sending Crash/Success information. Learn more at https://github.com/fastlane/fastlane#metrics
[...]

PR #10653, opened 2017-10-20
[Deliver] Deprecates add_id_info_limits_tracking in favor of add_id_info_uses_idfa

minuscorp:
### Checklist

### Motivation and Context
This change borns in the context of the discussion in #10632, where I discovered that the submission key `add_id_info_limits_tracking` should be always the same value as the `add_id_info_uses_idfa`, just like the iTunesConnect portal requires. If their values are different the submission for review process will fail for sure. This change deprecates the `add_id_info_limits_tracking` in favor of `add_id_info_uses_idfa`.

### Description
I've just mark as deprecated the `add_id_info_limits_tracking` property in favor of `add_id_info_uses_idfa` and when the parameters are being created for the review submission request, the value is substituted by whichever value `add_id_info_uses_idfa` has, as they have to be the same in order to work properly.

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_submission.rb
+++ b/spaceship/lib/spaceship/tunes/app_submission.rb
@@ -16,7 +16,9 @@ module Spaceship
 
       # To pass from the user
 
+      # @deprecated Setted automatically by <tt>add_id_info_uses_idfa</tt> usage
       # @return (Boolean) Ad ID Info - Limits ads tracking
+      # <b>DEPRECATED:</b> Use <tt>add_id_info_uses_idfa</tt> instead.
       attr_accessor :add_id_info_limits_tracking
 
       # @return (Boolean) Ad ID Info - Serves ads
@@ -69,7 +71,7 @@ module Spaceship
 
       attr_mapping({
         # Ad ID Info Section
-        'adIdInfo.limitsTracking.value' => :add_id_info_limits_tracking,
+        'adIdInfo.limitsTracking.value' => :add_id_info_uses_idfa,
         'adIdInfo.servesAds.value' => :add_id_info_serves_ads,
         'adIdInfo.tracksAction.value' => :add_id_info_tracks_action,
         'adIdInfo.tracksInstall.value' => :add_id_info_tracks_install,


minuscorp:
I signed it!

minuscorp:
One test is failing which has nothing to do with the change I made to the code... And does not fail in my machine... 😔 Any clue?

KrauseFx:
Maybe it was some caching issue, I triggered the build again 👍

mpirri:
@minuscorp Looks like we have an issue in our unit tests - I'm going to make a PR to fix the unit tests, and once that is merged, you can rebase off of master and we can try this again.  So sorry for the inconvenience with this!

minuscorp:
Don't worry! I wait for your comment then! Thank you!

mpirri:
@minuscorp Okay, the fix is merged to `master` - do you mind rebasing and re-pushing?  Thanks!

minuscorp:
Seems to be working! 😊

mpirri (approved):
Great job fixing this!  👍

mmenu-mantano:
Hi,

This fix doesn't seem to work for me (fastlane 2.62.1). Now the `adIdInfo.limitsTracking.value` is always `null` in the request sent to iTunes Connect. Regardless of whether I set the `add_id_info_limits_tracking` property in the Deliverfile or not.

It fails with this error: **Something wrong with your Ad ID information: ["ITC.apps.idfa.errorMessage.agreeoptout"]**.

Here's my submission information, in the Deliverfile:

```
submission_information({
  'content_rights_has_rights' => true,
  'export_compliance_available_on_french_store' => true,
  'add_id_info_uses_idfa' => true,
  'add_id_info_tracks_install' => false,
  'add_id_info_serves_ads' => true,
  'add_id_info_tracks_action' => false,
  'export_compliance_encryption_updated' => false,
  'export_compliance_contains_proprietary_cryptography' => true,
  'content_rights_contains_third_party_content' => true,
  'export_compliance_is_exempt' => true,
  'export_compliance_contains_third_party_cryptography' => true,
  'export_compliance_uses_encryption' => true,
})
```


And here's the data that is sent to Apple by fastlane:
```
"limitsTracking": { "value": null, "isEditable": false, "isRequired": false, "errorKeys": null }
```

minuscorp:
As far as i tested adding this to your `Deliverfile`:

```Ruby
submission_information({
  add_id_info_tracks_install: true,
  export_compliance_uses_encryption: false,
  export_compliance_encryption_updated: false,
  add_id_info_uses_idfa: true,
  content_rights_contains_third_party_content: false,
  content_rights_has_rights: true
})
```

Works fine. It can be a syntax error?

mmenu-mantano:
I don't think so, I was able to submit the app by reverting the changes of this pull request in my local repo, and adding back `add_id_info_limits_tracking: true` to my Deliverfile.

Did you check if `limitsTracking` has the proper value? Add a `puts data.to_json` in tunes_client.rb/send_app_submission()

minuscorp:
The pull request just ensures that the same value from `uses_idfa` is set to `limits_tracking`. 
It should not affect the result of the attribute mapping because it does not change anything on it. I don't know what to say...🤔

mmenu-mantano:
From looking at base.rb/mapping_module(), it looks like a new setter is created for each value in the `attr_mapping` dictionary. These setters are used in turn to set the matching raw value for the JSON request to Apple.

So I believe that using the same value twice for two different JSON keys can't work, as one will get its setter overwritten.

minuscorp:
So I think I messed up with the fix and created an issue instead... Is there a workaround to set the same value of another property in the `attr_mapping`?

mmenu-mantano:
I have no idea :/

As an alternative fastlane could issue a warning if `add_id_info_limits_tracking` is `false` when `add_id_info_uses_idfa` isn't. Especially since on iTunes Connect the checkbox is presented as a digital signature:

> I, Mr. X, confirm that this app, and any third party that interfaces with this app, uses the Advertising Identifier checks and honors a user's Limit Ad Tracking setting in iOS and, when it is enabled by a user, this app does not use Advertising Identifier, and any information obtained through the use of the Advertising Identifier, in any way other than for "Limited Advertising Purposes" as defined in the iOS Developer Program License Agreement.

minuscorp:
I was reading more about `attr_mapping`:

```ruby
      # Defines the attribute mapping between the response from Apple and our model objects.
      # Keys are to match keys in the response and the values are to match attributes on the model.
      #
      # Example of using `attr_mapping`
      #
      #   class Widget < Spaceship::Base
      #     attr_accessor :id, :name, :foo_bar, :wiz_baz
      #     attr_mapping({
      #       'name' => :name,
      #       'fooBar' => :foo_bar,
      #       'wizBaz' => :wiz_baz
      #     })
      #   end
```
So the key and the value should be separated, as the key is defined by `'adIdInfo.usesIdfa.value'` and the value is the `:add_id_info_uses_idfa` value...

mmenu-mantano:
Yes I think this solution could work when mapping from iTunes Connect to the model object, but it is not working in the other direction, to JSON.

minuscorp:
I'm working in a workaround that deletes the `:add_id_info_limits_tracking` from the `attr_mapping` and adds to the final mapping the required keys and values depending on the presence of the `'adIdInfo.usesIdfa.value'` in the json.

mmenu-mantano:
It didn't work unfortunately.

From what I gathered with the logs, `factory` was called before `add_id_info_uses_idfa` had a chance to be set.

You should be able to test your code without having a testing app, if you just check what is the out-going JSON. It should have a value for the key `limitsTracking`.

I edited tunes_client.rb and added `puts data.to_json` in `send_app_submission()`, I guess you can return early to avoid making the network call to iTunes Connect.

minuscorp:
Can we move the discussion to the new PR? Maybe someone can give me a hint to repair the mess and solve both issues. If we can't the only way is to revert the changes...

GandjaFuzz:
I have sam issue 
but 1 more
```Something wrong with your Ad ID information: ["ITC.apps.idfa.errorMessage.agreeoptout", "ITC.apps.idfa.errorMessage.binaryviolatesterms"].```

my delivery submissions
```			submission_information: {
				export_compliance_available_on_french_store: false,
				export_compliance_contains_proprietary_cryptography: false,
				export_compliance_contains_third_party_cryptography: false,
				export_compliance_is_exempt: false,
				export_compliance_uses_encryption: false,
				export_compliance_app_type: nil,
				export_compliance_encryption_updated: false,
				export_compliance_compliance_required: false,
				export_compliance_platform: "ios",
				content_rights_contains_third_party_content: true,
				content_rights_has_rights: true,
				add_id_info_serves_ads: false,
				add_id_info_uses_idfa: true,
				add_id_info_tracks_action: true,
				add_id_info_tracks_install: true
			},
```

minuscorp:
You can use the version we're working on in #10715 and let me know if it solves your issue!

GandjaFuzz:
I change file from you case - spaceship/lib/spaceship/tunes/app_submission.rb
and receive all ok

marcomorain:
Same here - this change has broken our submission process :(

mmenu-mantano:
@marcomorain Could you check the pull request here #10715 and give some feedback? It should fix the issue.

marcomorain:
I've reverted to `2.62.0` for now, I don't have to test it out I'm afraid :(

minuscorp:
I'm sorry for the inconvenience, we're waiting for the fix to be reviewed and pushed into production as soon as possible.

GandjaFuzz:
after update to fastlane 2.63 - problem return

minuscorp:
I'll request to be patched as soon as possible. PR was just merged.

PR #10655, opened 2017-10-20
[fastlane core] fix unit tests when Xcode 9.0 not installed

mpirri:
### Checklist

### Motivation and Context

Several unit tests for Analytics Session have an implicit assumption that Xcode 9.0 is installed; after upgrading to Xcode 9.0.1 this has caused 3 test failures. This change decouples these 3 tests from the actual Xcode version installed.

### Description

To do this, we mock out the `FastlaneCore::Helper` `xcode_version` method to always return the exact string that is in our test fixture.


ohayon (approved):
This is 💯 for now. I wonder if there is a way we can improve this so that we don't have to update these tests every time a new Xcode comes out?

KrauseFx (approved):
David, this is already stubbed now, so this should work with all future Xcode versions :+1:

PR #10662, opened 2017-10-21
fix warning for unused result

mhmiles:
Fixed a compiler warning due to an unused result in SnapshotHelper.swift


mhmiles:
I signed it!

ohayon (approved):
Not sure how we let this slip by! Thanks for the fix and welcome to the _fastlane_ community @mhmiles !!! 🚀

Issue #9917, opened 2017-08-01
plugins not available as dependencies of other plugins


jdee:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

If you add a plugin gem to another plugin as a (non-development) dependency in the gemspec, Fastlane does not detect it as an available plugin, and its actions are not available in the other_action context. The gem will be installed, and it is possible to call the action class directly, but it doesn't always work as expected.

##### Complete output when running fastlane, including the stack trace and command used

With `spec.add_dependency 'fastlane-plugin-patch'` in the gemspec:

```
[jdee@Jimmy-Dees-MacBookPro fastlane-plugin-branch (source-code-patching)]$ bundle exec fastlane lanes
+------------------------+---------+---------------------------------------+
|                               Used plugins                               |
+------------------------+---------+---------------------------------------+
| Plugin                 | Version | Action                                |
+------------------------+---------+---------------------------------------+
| fastlane-plugin-branch | 0.1.0   | setup_branch validate_universal_links |
+------------------------+---------+---------------------------------------+


--------- general---------
----- fastlane update
Add Branch keys and Universal Link domains to BranchPluginExample

----- fastlane update_and_commit
Add Branch keys and Universal Link domains to BranchPluginExample and commit to Git

----- fastlane validate
Validate Universal Link settings for a project

Execute using `fastlane [lane_name]`
```

With `gem 'fastlane-plugin-patch'` in the Pluginfile:

```
[jdee@Jimmy-Dees-MacBookPro fastlane-plugin-branch (source-code-patching)]$ bundle exec fastlane lanes
+------------------------+---------+---------------------------------------+
|                               Used plugins                               |
+------------------------+---------+---------------------------------------+
| Plugin                 | Version | Action                                |
+------------------------+---------+---------------------------------------+
| fastlane-plugin-branch | 0.1.0   | setup_branch validate_universal_links |
| fastlane-plugin-patch  | 0.1.0   | apply_patch revert_patch              |
+------------------------+---------+---------------------------------------+


--------- general---------
----- fastlane update
Add Branch keys and Universal Link domains to BranchPluginExample

----- fastlane update_and_commit
Add Branch keys and Universal Link domains to BranchPluginExample and commit to Git

----- fastlane validate
Validate Universal Link settings for a project

[...]

PR #10675, opened 2017-10-23
Document how to call another action in an action

### LINKED ISSUE #442: Document `other_action`
`other_action` to call an "other action" in an action exists and is very useful. Unfortunately it is not documented in the new docs, where I am creating this issue.

My usecase is that I theoretically wanted to to use plugin in a plugin, which is not possible. So I was advised to copy the action over to the second plugin and use it there. To do that, I have to go from one action to another.

---

Background from Slack:

> janpio [7:48 PM] 
> Can anyone think of a plugin that has at least 2 actions and one of these uses the other one? I need an  example how to best implement this.
> Hmm probably this: https://github.com/fastlane/fastlane/pull/4441
> Which means this: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/other_action.rb
> Correct?
> Is this documented anywhere?
> Seems it was, but isn't any more in the new docs or anything accessible:
> https://github.com/fastlane/fastlane/blob/master/fastlane/docs/README.md#extensions
> 
> josh [10:36] 
> `other_action.your_other_action_name`
> ie `other_action.<action_name>`

janpio:
Text cobbled together from https://github.com/fastlane/fastlane/pull/4441 and https://github.com/fastlane/fastlane/blob/master/fastlane/docs/README.md#extensions

 🎩💁‍♂️  @joshdholtz 

fixes https://github.com/fastlane/docs/issues/442

### DIFF
--- a/fastlane/docs/Plugins.md
+++ b/fastlane/docs/Plugins.md
@@ -154,3 +154,14 @@ gem "fastlane-plugin-[plugin_name]", git: "https://github.com/[user]/[plugin_nam
 #### Multiple actions in one plugin
 
 Let's assume you work on a `fastlane` plugin for project management software. You could call it `fastlane-plugin-pm` and it may contain any number of actions and helpers, just add them to your `actions` folder. Make sure to mention the available actions in your plugin's `README.md`.
+
+##### Calling another action in an action
+
+Unless there is a very good reason for it, you should not do this. But, to call another action from within your action, you can use the same code you would use in a `Fastfile` and prepend the action name with `other_action`:
+
+```ruby
+other_action.deliver(text: "Please input your password:", 
+                      key: 123)
+```
+
+In general, think twice before you do this, most of the times, the action should be separate and only called by the user's `Fastfile`.


jdee:
I think I would at least change the wording. There's nothing wrong with calling another action in an action. I think from your issue it sounds like you may be better off refactoring your code so that all your actions can share the same helper code and don't depend on each other. But there are plenty of times you may want to call built-in actions from a plugin or other action.

I do quite a bit of this these days:

https://github.com/BranchMetrics/fastlane-plugin-branch/blob/master/lib/fastlane/plugin/branch/actions/setup_branch_action.rb#L275
https://github.com/BranchMetrics/fastlane-plugin-branch/blob/master/lib/fastlane/plugin/branch/actions/setup_branch_action.rb#L300
https://github.com/BranchMetrics/fastlane-plugin-branch/blob/master/lib/fastlane/plugin/branch/actions/setup_branch_action.rb#L49
https://github.com/BranchMetrics/fastlane-plugin-branch/blob/master/lib/fastlane/plugin/branch/actions/setup_branch_action.rb#L94

I'd rather use the built-in actions and leverage the work that's been done there than shell out in all those cases. Calling a built-in action gives you nice output, easy error handling and reporting, all the options to that action, and so on. It's also not that easy to use `other_action` from a helper. You only get the `other_action` context within a subclass of Action, so in a helper you have to do something like:

```ruby
def do_something(context, ...)
  # ...
  # Do something in the helper
  # ...

  # Now call the action
  context.some_action
end
```

And then in the action:
```ruby
Helper.do_something other_action
```

The `other_action` attribute is just a convenience, however. Any time you have an action, you always have the option of just requiring the Action subclass and calling it, e.g.:

https://github.com/BranchMetrics/fastlane-plugin-branch/blob/master/lib/fastlane/plugin/branch/helper/ios_helper.rb#L474

```ruby
require "fastlane/plugin/patch"

        Actions::PatchAction.run(
          files: podfile_path,
          regexp: /^(\s*)pod\s*/,
          text: "\n\\1pod \"Branch\"\n",
          mode: :prepend,
          offset: 0
        )
```

(You can see I also have plenty of refactoring to do in this codebase.) You can do this with most built-in actions as well, though there are a few, like `sh`, that have fancy implementations without a `run` method.

This is the sort of thing I'll be sure to explain in the guide to building actions and plugins that I plan to write soon. Thanks for bringing up the issue.

janpio:
There is no issue from me, so I am not sure what you are referring to. Nothing in this PR is original content from me, just text pieces of former official documentation and pull requests, which served as quasi-documentation, that I linked in the PR.

I just wanted to access an action in another action and couldn't find any documentation on it. This lead me to asking in Slack and finding the linked URLs via Google.

jdee:
https://github.com/fastlane/docs/issues/442, the one you linked to in this PR.

"Can anyone think of a plugin that has at least 2 actions and one of these uses the other one?"

That's what I meant. If you have multiple actions in the same plugin, and one has to call the other, you could also think about factoring it out so that they can both call the same helper code.

I just wouldn't go so far as to say that you should not call an action from another action.

janpio:
Ahhhh, this was a quote from Slack - I should have made that clearer ;)

Background information on this: I have 2 Fastlane plugins. One should use an action from the other. As there are no dependencies from plugins on plugins (I was told), I was advised to just copy the action from one plugin to the other and use it. (Rewriting in a helper wouldn't help, as I still had to copy over a file)

janpio:
> I just wouldn't go so far as to say that you should not call an action from another action.

I don't disagree.

@KrauseFx, this is based on a quote from the PR you wrote. What is the current stance on `other_action`? Should I remove the "disclaimers" in this PR? Make it less drastic?

jdee:
Here's the problem with having one plugin depend on another: https://github.com/fastlane/fastlane/issues/9917.

I do exactly that in the branch plugin, using the patch plugin as a dependency. Then `other_action` does not work for the reasons mentioned in that issue. But then I can just call the gem like any other. That's a case where the patch action is basically just a utility action, one that probably isn't very useful in a Fastfile, but may be used by other actions. In fact, I factored it out into a separate gem so I could also use it in non-Fastlane code, and eventually I'll refactor the branch plugin so that it doesn't use the action.

Someone also built a plugin on top of my plugin in order to reuse the helper code. https://github.com/col/fastlane-plugin-remove_setting

It would also be nice to fix that dependency issue.

I like the wording in the current README: "Think twice."

ohayon:
Hey folks, I am a bit hesitant to merge this in as I agree with one of @jdee's points that if this is necessary, it probably means that the actions should be distinct. Also, given that we dont see many requests of this type, I think we should hold off on making this change until we realize that its in high demand from the community. Curious to hear your thoughts on that before closing the PR 🐙

janpio:
The functionality and documentation for this is already there. This only adds it to the new docs.

ohayon:
Hey @janpio, could you point me in the direction of the existing documentation?

casz:
@ohayon linked in the OP 😆  https://github.com/fastlane/fastlane/blob/master/fastlane/docs/README.md#extensions
image

jdee:
Admittedly the use case I have for this is outside the main focus of Fastlane, since I'm mostly building SDKs and tools, not apps. There may not ordinarily be that many cases where you need to call an action from another action.

If you have a plugin that updates a custom configuration file, and you want to optionally commit the changes, it's not hard to shell out to `git commit`, but it also seems reasonable to think you could call the built-in `git_commit` action. (And if you shell out, you might want to use the `sh` action to do it.) I definitely would rethink calling an action from another action in the same plugin except in some cases, e.g., you've deprecated one action and replaced it with another. It's probably an indication that it's time to refactor. There can also be issues with config. files in actions that have them (like gym, deliver, scan), depending how you call them. In the long run it might be best to distinguish a number of cases. For now, the README seems fine.

I'll take a crack at the promised guide to building plugins and actions, probably this weekend, and we can discuss whether/how to improve on current advice.

PR #10687, opened 2017-10-24
Pin rubocop in new plugins to the same version as the main repo.

jdee:
### Checklist

### Motivation and Context

Whenever a new minor version of RuboCop is released, new Cops are added, and Cops often change namespaces, generating warnings and errors from RuboCop. This is avoided in the main repo by pinning to a specific version of RuboCop, but the plugin template does not include the same version requirement. The same .rubocop.yml is used in both places. The result is often that in a new plugin, the most recent version of RuboCop is installed, and the generated code doesn't pass with the .rubocop.yml from the repo. The simplest solution seems to be pinning new plugins to the same version as the repo.

### Description

Added a `Fastlane::RUBOCOP_REQUIREMENT` constant, used in `fastlane.gemspec`, plugin generator tests and the plugin template itself. This is probably the simplest solution.

Originally I was pulling it out of the gemspec, but that is not present when installed via RubyGems.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -81,7 +81,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'yard', '~> 0.8.7.4'
   spec.add_development_dependency 'webmock', '~> 2.3.2'
   spec.add_development_dependency 'coveralls', '~> 0.8.13'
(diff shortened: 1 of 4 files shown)


ohayon (approved):
This is 🍠 ! Thanks @jdee !! 🚀

PR #10706, opened 2017-10-26
In two-stage Hockey action, do not set the status to 2 (downloadable) until after the app is uploaded

srubin:
### Checklist

### Motivation and Context

When using the two-stage Hockey upload process (create a new version and then upload app assets to the newly created version), you should not set the version's status to "2" -- which means it's downloadable -- until you've uploaded the app.

My understanding from my [Hockey support request](https://support.hockeyapp.net/discussions/problems/83559) is that  Hockey will only fire "New Release" webhooks when a version's status is first set to "2" _iff_ there's an app to download.

I've tested these changes with my app, and Hockey correctly fires the "Release" webhook. Before this change, that webhook doesn't fire.

### Description

Before creating the new version, we save the fastlane user's requested `status` and set the status to `"1"` ([Don't allow users to download or install the version](https://support.hockeyapp.net/kb/api/api-versions#-u-post-api-2-apps-app_id-app_versions-new-u-)). Then, when uploading the app, set `status` back to the user's requested status.

### DIFF
--- a/fastlane/lib/fastlane/actions/hockey.rb
+++ b/fastlane/lib/fastlane/actions/hockey.rb
@@ -82,8 +82,8 @@ module Fastlane
 
         # https://support.hockeyapp.net/discussions/problems/83559
         # Should not set status to "2" (downloadable) until after the app is uploaded.
-        final_status = options[:status]
-        options[:status] = "1"
+        update_status = options[:status]
+        options[:status] = options[:create_status]
 
         response = connection.get do |req|
           req.url("/api/2/apps/#{app_id}/app_versions/new")
@@ -105,7 +105,7 @@ module Fastlane
           options[:dsym] = dsym_io
         end
 
-        options[:status] = final_status
+        options[:status] = update_status
 
         connection.put do |req|
           req.options.timeout = options.delete(:timeout)
@@ -245,6 +245,10 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :status,
                                        env_name: "FL_HOCKEY_STATUS",
                                        description: "Download status: \"1\" = No user can download; \"2\" = Available for download (only possible with full-access token)",
+                                       default_value: "2"),
+          FastlaneCore::ConfigItem.new(key: :create_status,
+                                       env_name: "FL_HOCKEY_CREATE_STATUS",
+                                       description: "Download status for initial version creation when create_update is true: \"1\" = No user can download; \"2\" = Available for download (only possible with full-access token)",
                                        default_value: "2"),
           FastlaneCore::ConfigItem.new(key: :notes_type,
                                       env_name: "FL_HOCKEY_NOTES_TYPE",
--- a/fastlane/lib/fastlane/actions/hockey.rb
+++ b/fastlane/lib/fastlane/actions/hockey.rb
@@ -81,7 +81,8 @@ module Fastlane
         end
 
         # https://support.hockeyapp.net/discussions/problems/83559
-        # Should not set status to "2" (downloadable) until after the app is uploaded.
+        # Should not set status to "2" (downloadable) until after the app is uploaded, so allow the caller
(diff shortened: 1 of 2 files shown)


srubin:
I signed it!

ohayon:
Hey @srubin - this PR looks great! I'm a bit concerned to merge it however as this is the first time that I personally have seen this be an issue raised up by the community and given how often the hockey action is used, I think I would have expected to see this sooner. Is there any sort of test we can write to make sure this is always the case? Or, is there some sort of override we could provide for people to not use this flow? Thanks! 🚀

srubin:
Understood about it being odd that it's the first time you've seen this issue raised. Seeing this bug relies on 2 properties of a build that I (maybe wrongly) assume are less common: (1) the two-phase version creation -- which is only required when you're uploading an app that's not a .app or .ipa or .apk (in our case, we switched to uploading a .dmg), and (2) monitoring the "Release" webhook for some purpose. As far as I can tell, there's no real bug on Hockey's end other than that the webhook doesn't fire. I'm not sure I'd even call that a bug.

So, regarding testing -- testing whether the webhook fired seems out of the scope of the _fastlane_ tests because it's essentially testing the way a third-party service behaves. What are your thoughts there, @ohayon? 

I'd be happy to add an option to to skip this flow, if you'd feel more comfortable with it. I can't currently think of an advantage of not using this new flow, but without understanding the internals of Hockey, I can't be sure.

ohayon:
Ah! Thanks for the explanation @srubin that makes total sense, maybe we can just add a check for the file type that is being uploaded and then use this new flow if it isn't a standard .app, .ipa, or .apk file?

srubin:
Well, it's already an opt-in flow: we only call the `create_and_update_build` function when the user specifies the `create_update = true` option when executing the hockey action.

ohayon:
Sorry, I'm not so familiar with the way that hockey works, but is it the case that someone could be using that flag and still be uploading a standard file type and would then enter this new code?

srubin:
Yes, someone could set that flag while still uploading a standard file type, but I'm not sure why they would.

- According to [the documentation for the version update endpoint](https://support.hockeyapp.net/kb/api/api-versions#update-version), this can be used:
  - to upload the mapping.txt without the .apk file for Android apps, or
  - to upload additional .dSYM files for iOS and OS X apps, or
  - to upload .pkg or .dmg files for OS X apps, or
  - to upload .msi files for Windows apps, or
  - to upload .zip files for custom apps.

And there's no indication that I can find anywhere in the HockeyApp docs that suggests any other advantages of using the version update endpoint.

I thought of another more explicit way to deal with this issue: we could add a parameter to the Hockey action called `create_status` which we would pass as the `status` when we _create_ the version, and then we could use `status` (possibly mirrored by a new action parameter `update_status`) as the `status` when we _update_ the version. We could default `create_status` to `"1"` and `update_status` to `"2"` -- then, at least, it would be clear what's going on from the documentation. And it would be easy to override this behavior.

ohayon:
Hey @srubin thanks for the explanation and apologies if I am being cautious here, the reason is that we have a lot of users that rely heavily on this action and I really want to make sure that we are not changing any existing functionality. I feel like the most reliable way to make sure that we are not breaking anyone's flow is to keep things how they are and use your new option to create a different code path even though that might not be the most elegant solution. Thoughts?

srubin:
Understood. Can you take a look at my latest changes? I added a `create_status` parameter that defaults to "2". That way, there's no change in the existing behavior unless the user changes `create_status` when calling their hockey action.

The change is only a few lines of code, so it's pretty easy to see that there's no logical difference (by default) between this and the existing behavior. What do you think about this? Or, when you said we should create a different code path, did you mean something more extreme (like branching on an optional parameter)?

ohayon (approved):
Thats great! Thanks @srubin !! 🚀

Issue #10422, opened 2017-09-26
Feature Request: Reset iOS Summary Rating


pparmaei:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### in iTC a new field has been added:

- 'Reset iOS Summary Rating'

- You can reset your app’s summary rating for all territories when you release this version. Keep in mind that once this version is released, you won’t be able to restore the rating. Your app’s existing customer reviews will still appear on the App Store. View Ratings and Reviews

With two options:

- Keep existing rating

- Reset rating when this version is released 

Can this field be added to the fields that can be updated using Fastlane?

Thanks

adellibovi:
@KrauseFx, I would like to work on this issue if it’s something we want to add, which I suppose yes, but please confirm.
Also would you mind to point me out how to approach the PR?
I.e.: do we discuss the naming of parameters ahead or I could go with my suggestions and discuss it in the PR? This would be part of “deliver” right? Thanks 😊

KrauseFx:
Sounds great, thanks you two, [CONTRIBUTING.md](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md) is a great starting point and for all `spaceship` related changes, check out https://github.com/fastlane/fastlane/blob/master/ToolsAndDebugging.md 👍

adellibovi:
@KrauseFx pushed the PR 😄 🎉

pparmaei:
@adellibovi cheers!

PR #10708, opened 2017-10-26
Add support for reset rating

adellibovi:
### Checklist

### Motivation and Context

This fixes issue #10422 
### Description

As title, it add support of the recent feature for resetting or not the rating of an app in iTunesConnect

### DIFF
--- a/spaceship/docs/iTunesConnect.md
+++ b/spaceship/docs/iTunesConnect.md
@@ -129,6 +129,7 @@ attr_accessor :can_reject_version
 attr_accessor :can_prepare_for_upload
 attr_accessor :can_send_version_live
 attr_accessor :release_on_approval
+attr_accessor :ratings_reset
 attr_accessor :can_beta_test
 attr_accessor :supports_apple_watch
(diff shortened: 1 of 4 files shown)


ohayon (commented):
Hey @adellibovi ! Thanks for the PR! This would definitely be a great addition!

I just have one question here: in the fixture JSON, it looks like `ratingsReset` is a hash, but in the code it is just a Bool, have you tested this to see if its working? 🐙

adellibovi:
@ohayon thanks for checking it out!
The magic happens in  `attr_mapping` and `'ratingsReset.value' => :ratings_reset,` where we get only `value` and not the entire hash 😄 .
Also I added an expectation in the test to check if it's properly initialised to `false`, in this particular fixture.

ohayon (approved):
Ohhh! My bad! I missed the `.value` on that `attr_mapping`. Sweet! Lets do it! 🚀

adellibovi:
Thanks @ohayon, are you going to close #10422?

Issue #10713, opened 2017-10-27
Supporting multiple review screenshot resolutions with in app purchases


alaincaltieri:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
iTunes connect supports multiple screenshot sizes when dealing with review screenshot assets for in app purchases. Spaceship instead requires to use 640x960 resolution (iPhone 4s) which is a bit tedious since the app may not be optimised for that screen resolution and aspect ratio as less and less people use those devices.

It would be nice to have Spaceship support all valid screen resolutions and I'd be happy to do that. I'd need a bit of introduction on how to approach this maybe, since I don't know that much private APIs of iTC. Just pointing out in general how it works and where this call is performed should be enough for getting me started :)

This has been referenced in https://github.com/fastlane/fastlane/issues/9390 and still looks open with the latest fastlane release (tested locally)

What do you think about this?

alaincaltieri:
I actually found my way through, was pretty easy.

Will create a PR soon. I hope this is useful and will find it's way to master 😄

earien:
👍 It would be superuseful... being forced to upload an iphone 4s screenshot is pretty annoying 😄

PR #10716, opened 2017-10-27
[Spaceship] Support all review screenshot resolutions - iOS In App Purchases

alaincaltieri:
### Checklist

### Motivation and Context
In App Purchases in Spaceship required always to provide iPhone4 review screenshots, which was cumbersome to the user. This PR introduces support for all supported devices and exposes a way to retrieve the correct Apple validation rule given the screenshot file. 

It fixes this issue: https://github.com/fastlane/fastlane/issues/10713

I've updated the unit tests and tested the changes locally with the most common screen resolutions, including iPhone 6 plus, iPhone 6, iPad and iPad Pro. I also tested using screens without the status bar area (normally 40px less in height) and in portrait modes.

### Description
I've introduced a new public method in du_client.rb:  `get_picture_type(upload_file)` which provides Apple's picture_type code for screenshot validations (used also for uploading app screenshots). This method computes the correct validation rule starting from the UploadFile object, by leveraging Spaceship::Utilities  module to get the resolution of the screenshot file and map that to the proper device type. It then uses that to get the validation code with the already existing methods.

I've also slightly refactored the way iAP review screenshots data was retrieved by the itunes_client. This used to be the raw response, but then required to expose some information that only the du_client should have (namely the Apple's picture_type validation code). To avoid this, and reuse the code that was previously duplicated in iap_details and itunes_client itself, I wrapped the definition of the screenshot hash data to be sent to iTunesConnect to save a screenshot in itunes_client.rb's `upload_purchase_review_screenshot` method.

This change should not affect in any way the public apis of Spaceship and should thereby be transparent to the user. Users can simply provide the review screenshots with much more flexibility.

Hope this helps the community and gets merged soon! Let me know if I need to provide further details 😄

### DIFF
--- a/spaceship/lib/spaceship/du/du_client.rb
+++ b/spaceship/lib/spaceship/du/du_client.rb
@@ -44,11 +44,11 @@ module Spaceship
 
     def get_picture_type(upload_file)
       resolution = Utilities.resolution(upload_file.file_path)
-      result = device_resolution_map.find { |key, resolutions| 
+      result = device_resolution_map.find do |key, resolutions|
         resolutions.include? resolution
-      }
-      raise "Unknown device for screen resolution #{resolution}" unless result != nil
-      
+      end
+      raise "Unknown device for screen resolution #{resolution}" if result.nil?
+
       picture_type_map[result[0]]
     end
 
--- a/spaceship/lib/spaceship/du/du_client.rb
+++ b/spaceship/lib/spaceship/du/du_client.rb
@@ -47,7 +47,7 @@ module Spaceship
       result = device_resolution_map.find { |key, resolutions| 
         resolutions.include? resolution
       }
-      raise "Unknown picture type for resolution #{resolution}" unless result != nil
+      raise "Unknown device for screen resolution #{resolution}" unless result != nil
       
       picture_type_map[result[0]]
     end
--- a/spaceship/lib/spaceship/du/du_client.rb
+++ b/spaceship/lib/spaceship/du/du_client.rb
@@ -19,7 +19,7 @@ module Spaceship
     end
 
     def upload_purchase_review_screenshot(app_id, upload_file, content_provider_id, sso_token_for_image)
-      upload_file(app_id: app_id, upload_file: upload_file, path: '/upload/image', content_provider_id: content_provider_id, sso_token: sso_token_for_image, du_validation_rule_set: 'MZPFT.SortedScreenShot')
+      upload_file(app_id: app_id, upload_file: upload_file, path: '/upload/image', content_provider_id: content_provider_id, sso_token: sso_token_for_image, du_validation_rule_set: get_picture_type(upload_file))
     end
 
     def upload_large_icon(app_version, upload_file, content_provider_id, sso_token_for_image)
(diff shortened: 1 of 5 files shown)


alaincaltieri:
@hjanuschka If I remember correctly you're the master mind behind the in app purchases support in Spaceship, am I right? Can you have a look at this? :smile:

matteomartinelli:
I was looking for the same feature since I'm forced to use 640x960 screenshot while in iTC is possible to have different resolution.

+1

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -661,9 +661,21 @@ def upload_watch_icon(app_version, upload_image)
...
+              "assetToken" => data["token"],
+              "sortOrder" => 0,
+              "type" => du_client.get_picture_type(upload_image),
+              "originalFileName" => upload_image.file_name,
+              "size" => data["length"],
+              "height" => data["height"],
+              "width" => data["width"],
+              "checksum" => data["md5"]
+          }
+      }
```
ohayon: This method looks like it is calling into the client to upload something but then has a side affect of returning a hash? Is it possible that we can separate out these two things? I'm not sure I understand the change and why this was extracted from `iap_detail.rb`?

spaceship/lib/spaceship/du/du_client.rb
```diff
@@ -111,6 +121,21 @@ def messages_picture_type_map
...
+        ipad:         [[1024, 748], [1024, 768], [2048, 1496], [2048, 1536], [768, 1004], [768, 1024], [1536, 2008], [1536, 2048]],
+        ipadPro:      [[2048, 2732], [2732, 2048]],
+        iphone6:      [[750, 1334], [1334, 750]],
+        iphone6Plus:  [[1242, 2208], [2208, 1242]],
+        iphone4:      [[640, 1096], [640, 1136], [1136, 600], [1136, 640]],
+        iphone35:     [[640, 960], [640, 920], [960, 600], [960, 640]],
+        appleTV:      [[1920, 1080]],
+        desktop:      [[1280, 800], [1440, 900], [2560, 1600], [2880, 1800]]
+      }
+    end
```
ohayon: Would it be possible to have all these 2 dimensional arrays be hashes with keys "width" and "height" to make this a bit more clear?

ohayon (commented):
Hey @alaincaltieri thanks for the PR ! This is definitely a great change, just had a couple questions in the review. Looking forward to hearing your thoughts! 🚀

spaceship/lib/spaceship/du/du_client.rb
alaincaltieri: I was concerned with the verbosity of that option, and this looked a bit more natural to me since it's the same format `fastimage` uses to express resolutions (and is also the same format `Utilities.resolution(upload_file.file_path)` returns.

I can easily change this of course. Let me know what do you think

spaceship/lib/spaceship/tunes/tunes_client.rb
alaincaltieri: Hi @ohayon! Thanks for the review!
This code creating the hash was actually duplicated in both `iap_detail.rb` and in `tunes_client.rb`, which made iAP modification and creation behave in potentially different ways. So my primary goal was to reduce duplication and manage this in only one place. `itunes_client.rb` was the best one I could think of. 

Also since this method in `tunes_client.rb` is specific to the iAP functionality and was just relaying the `du_client.rb` functionality, I thought it could have been the perfect place to generate the hash too.

An alternative would be to keep the duplicated code and just use the new way to find the apple screen validation in both places. Or we can expose a new method in tunes_client that builds the hash in order to avoid the duplication and that leverages the old `upload_purchase_review_screenshot` method? Would that be better?

alaincaltieri:
@ohayon Any news on this? Did you have a chance to have a look at the comments above? :)

KrauseFx (approved):
Looks great, thanks for your contribution @alaincaltieri, and sorry for the delay

Issue #10686, opened 2017-10-24
deliver fails when app is "Waiting for Review"


RishabhTayal:
Reopening issue: https://github.com/fastlane/fastlane/issues/6010 to reinitiate the conversation.

When the app is in "Waiting for review" status on iTunes Connect, `deliver` will fail with error `Problem processing review submission.`. According to the issue #6010 this is an expected behavior. I just want to understand if this is still the case. Is there any other way to reject the version on the iTunes connect and resubmit with `deliver` automatically?

minuscorp:
I'd like to help, can you provide some more information about this issue, maybe you can run in `verbose` the command and add it to the issue. 🚀

RishabhTayal:
@minuscorp Just add `deliver(submit_for_review: true, force: true)` to your lane and run it when the latest version is in "Waiting for Review" status in iTunes Connect.

RishabhTayal:
correct. Maybe there could be a parameter which will reject the current build on iTunes Connect and submit the new one.

ohayon:
Hey folks! I think thats a really great idea @RishabhTayal, it would be nice to have that option. We would definitely take a look at a PR with that added. Let us know if we can help! 🚀

minuscorp:
Reading the stub that produces the error: [this](https://github.com/fastlane/fastlane/blob/master/spaceship/spec/tunes/fixtures/app_submission/complete_failed.json), I'm wondering if the only difference between a successful review and one that doesn't is the `error` key with `null` content... If there's no certain message that Apple uses to really confirm that the review was errored because it's being reviewed... How can we do this?

RishabhTayal:
@ohayon where can I start with the implementation? I know about the YOUR_FIRST_PR.md. I am talking about the technical details about the feature?

ohayon:
Hey @RishabhTayal - I think [this file](https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/submit_for_review.rb) would maybe be a good place to start tracing through some of the code. Also, if you're not familiar, you can use [pry](https://github.com/nixme/pry-debugger) to help with debugging and stepping through code too 😄 .

RishabhTayal:
@ohayon I will check that. What's a good way to mock the app store submission while development, since I don't want to keep on submitting a build for dev purpose.

ohayon:
@RishabhTayal I think we have some fixture json files that have mocked responses, you should be able to use WebMock and return those as necessary 👍

[...]

PR #10722, opened 2017-10-27
deliver - option for submitting when waiting for review

RishabhTayal:
Adding new option in `deliver` which will reject the submission on iTunes connect if it's in "Waiting for review" status and submit the new build.

Fixes: #10686


deliver/lib/deliver/submit_for_review.rb
```diff
@@ -22,7 +22,11 @@ def submit!(options)
       end
 
       # Finalize app submission
-      submission.complete!
+      if options[:reject_if_waiting_for_review]
```
getaaron: Will this conditional always evaluate to `true` since you set a `default_value` in `options.rb`?

If you, you can just call `submission.complete(reject_if_waiting_for_review)`

spaceship/lib/spaceship/tunes/app_submission.rb
```diff
@@ -124,7 +124,7 @@ def create(application, version)
       end
 
       # Save and complete the app submission
-      def complete!
+      def complete(reject_if_waiting_for_review)
```
getaaron: Why remove the `!`?

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -208,6 +207,8 @@ def handle_itc_response(raw, flaky_api_call: false)
           raise_insuffient_permission_error!
         elsif flaky_api_call
           raise ITunesConnectPotentialServerError.new, errors.join(' ')
+        elsif errors.count == 1 and errors.first.first.include?("Problem processing review submission.")
```
getaaron: Did you mean `errors.first.first`? The other checks just use `errors.first`. If `errors.first` is a `String` you will get `NoMethodError`

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -971,6 +972,18 @@ def send_app_submission(app_id, version, data)
 
       handle_itc_response(r.body)
 
+      messages = r.body.fetch("messages", {})
+      errors = messages.fetch('error', nil)
+
+      if !errors.nil? and !errors.empty? and errors.first.include?("Problem processing review submission.")
```
getaaron: I think you could use the Ruby safe navigation operator here to avoid the explicit nil checks:

```ruby
if errors&.first&.include?("Problem processing review submission.")
```

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -987,6 +1000,22 @@ def send_app_submission(app_id, version, data)
     end
 
     #####################################################
+    # @!group Reject App Submission
+    #####################################################
+    def reject_app_submission(app_id, version)
+      raise "app_id is required" unless app_id
```
getaaron: Do we want to check for `version` here also?

spaceship/spec/tunes/app_submission_spec.rb
```diff
@@ -54,7 +54,7 @@
       submission.add_id_info_uses_idfa = false
 
       expect do
-        submission.complete!
+        submission.complete(false)
       end.to raise_exception("Problem processing review submission.")
       expect(submission.submitted_for_review).to eq(false)
     end
```
getaaron: We should add a test for `complete(true)`

spaceship/lib/spaceship/tunes/tunes_client.rb
taquitos: Wow, TIL: safe nav operator.
That's a great suggestion, but unfortunately we're tied to Ruby 2.0.0 😢 and it looks like the safe nav operator was introduced in 2.3.0

RishabhTayal: based on @taquitos's comment, keeping the existing code. Not making requested change.

spaceship/lib/spaceship/tunes/app_submission.rb
janpio: This is making the tests failing btw.

spaceship/lib/spaceship/tunes/tunes_client.rb
RishabhTayal: I tested this with mock flow. What kind of change are you expecting me to add?

spaceship/lib/spaceship/tunes/app_submission.rb
RishabhTayal: Fixed the test case.

RishabhTayal:
Can someone help me with test cases for this PR? I would like to get this PR merged. What do I need to do?

minuscorp:
So I'm seeing in the CircleCI, that your code does not pass the Rubocop step due to several indentation issues and so on. Check out the log of the build to see the affected lines!

janpio:
Next step is to rebase on a recent version of `master` so your code is closer to `master` when merging.

RishabhTayal:
@janpio merged `master` to my branch. @minuscorp Fixed the formatting. 

The test case is however failing.

minuscorp:
One offense found in Rubocop

RishabhTayal:
@minuscorp fixed now.

minuscorp:
So... I'm seeing this test case error:
```
Real HTTP connections are disabled. Unregistered request: GET https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/apps/898536088/versions/812106519/submit/summary with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Content-Type'=>'application/json', 'User-Agent'=>'Spaceship 2.76.0'}
```
And in the list of stubbed requests there's only one match with your query:
```
stub_request(:get, "https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/apps/898536088/versions/812106519/submit/summary").
  with(:headers => {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Content-Type'=>'application/json', 'User-Agent'=>'Spaceship 2.70.3'}
```
The `'User-Agent`' value is getting different versions from your request than from the stubbed one.
The reason is:
```ruby
stub_request(:get, "https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/apps/898536088/versions/812106519/submit/summary").
    with(headers: { 'Accept' => '*/*', 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Content-Type' => 'application/json', 'User-Agent' => 'Spaceship 2.70.3' }).
    to_return(status: 200, body: "", headers: {})
```
Hard-coding those parameters risk your tests to fail in future releases.

RishabhTayal:
@minuscorp Removed header from the stubbing but still test case is failing.

```
Spaceship::Tunes::AppSubmission
  successfully creates a new app submission
    successfully submits app for review if `reject_if_waiting_for_review` option is true (FAILED - 1)

Failures:

  1) Spaceship::Tunes::AppSubmission successfully creates a new app submission successfully submits app for review if `reject_if_waiting_for_review` option is true
     Failure/Error: submission = app.create_submission
     Spaceship::UnexpectedResponse:
     # ./spaceship/lib/spaceship/client.rb:592:in `parse_response'
     # ./spaceship/lib/spaceship/tunes/tunes_client.rb:988:in `prepare_app_submissions'
     # ./spaceship/lib/spaceship/tunes/app_submission.rb:119:in `create'
     # ./spaceship/lib/spaceship/tunes/application.rb:317:in `create_submission'
     # ./spaceship/spec/tunes/app_submission_spec.rb:67:in `block (3 levels) in <top (required)>'

Finished in 1.31 seconds (files took 6.69 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spaceship/spec/tunes/app_submission_spec.rb:62 # Spaceship::Tunes::AppSubmission successfully creates a new app submission successfully submits app for review if `reject_if_waiting_for_review` option is true
```

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -999,6 +1000,18 @@ def send_app_submission(app_id, version, data)
 
       handle_itc_response(r.body)
 
+      messages = r.body.fetch("messages", {})
+      errors = messages.fetch('error', nil)
+
+      if !errors.nil? and !errors.empty? and errors.first.include?("Problem processing review submission.")
+        if reject_if_waiting_for_review
+          reject_app_submission(app_id, version)
```
minuscorp: Did you check that this is actually being called and that it behaves as expected?

janpio:
Before you start deleting parts from the tests, make sure that these test failures are actually triggered by changes you made. If not, you should not start messing with tests.

spaceship/lib/spaceship/tunes/tunes_client.rb
RishabhTayal: @minuscorp I have tested the reject app submission option and it's working as intended. Not sure how to fix the test case though.

RishabhTayal: @minuscorp Finally figured out the test case. Fixed it.

RishabhTayal:
All the test cases are passed, code formatting complete. Pull request is ready to be reviewed again.

deliver/lib/deliver/submit_for_review.rb
janpio: +1

spaceship/spec/tunes/app_submission_spec.rb
```diff
@@ -22,7 +22,7 @@
       submission.content_rights_contains_third_party_content = true
       submission.content_rights_has_rights = true
       submission.add_id_info_uses_idfa = false
-      submission.complete!
+      submission.complete!(false)
```
janpio: Is there a way to set `false` as the default of no param is supplied to this method? Would be much neater that way.

RishabhTayal: Good idea. Added that.

RishabhTayal:
@janpio This is ready to be reviewed again.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -212,7 +212,6 @@ def handle_itc_response(raw, flaky_api_call: false)
       # e.g. {"warn"=>nil, "error"=>["operation_failed"], "info"=>nil}
       different_error = raw.fetch('messages', {}).fetch('error', nil)
       errors << different_error if different_error
-
```
janpio: why remove this empty line?

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -999,6 +1000,19 @@ def send_app_submission(app_id, version, data)
 
       handle_itc_response(r.body)
 
+      messages = r.body.fetch("messages", {})
+      errors = messages.fetch('error', nil)
+
+      if !errors.nil? and !errors.empty? and errors.first.include?("Problem processing review submission.")
```
janpio: What other reasons could there be for that besides "Waiting for review"? Isn't it kind of dangerous to just reject the app even if the error doesn't directly

spaceship/lib/spaceship/tunes/tunes_client.rb
RishabhTayal: Added the empty line back again.

spaceship/lib/spaceship/tunes/tunes_client.rb
RishabhTayal: @janpio This error message is returned when the app is already submitted for review. The status could be "Waiting for review", "In Review", "Pending developer release". In all three cases it will reject the app and submit again.

RishabhTayal: I think we should change the param name from `reject_if_waiting_for_review` to something more generic like `reject_if_already_submitted` or `reject_if_previously_submitted`. I am open to suggestions if you have any other name. We will have to change the description of the param accordingly.

minuscorp: I agree with @janpio that assuming that if the error received matches that string the error is due to the review process is already going on. The response does not include any other error that ensures that the error is produced by the application being already sent to review?

RishabhTayal: Is there other solution for this problem? Is there a better way to check if the app is only in the above mentioned states?

RishabhTayal:
@KrauseFx Can you please review this pr?

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: Yes, can confirm that this will eventually crash with `NoMethodError`, as you don't know how many errors are there, could be an empty array

spaceship/lib/spaceship/tunes/tunes_client.rb
KrauseFx: But I think we could use `Array(errors).include?("something")`, right?

KrauseFx (commented):
Thanks for the work on this PR, to be honest, it feels a little fragile at first, we have to make sure we don't introduce any regressions for existing setups 👍

gerbil802:
How is this different than this? https://github.com/fastlane/fastlane/issues/7149

joshdholtz:
@gerbil802 Thanks for linking that! It looks like that PR (https://github.com/fastlane/fastlane/pull/7152) _actually_ only added the method in _spaceship_ to reject (it did not actually get tied it into _deliver_).

The best approach here might be to use the `reject!` defined method in `app_version.rb` since that already prechecks with `can_reject_version`? So maybe instead of monitoring error statuses that get returned (since that does seem a bit fragile) that this PR gets changed up to leverage existing `reject!` logic? 🤔

RishabhTayal:
@joshdholtz, @KrauseFx This is ready to review again.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -1014,6 +1027,22 @@ def send_app_submission(app_id, version, data)
       parse_response(r, 'data')
     end
 
+    #####################################################
+    # @!group Reject App Submission
+    #####################################################
+    def reject_app_submission(app_id, version)
```
joshdholtz: @RishabhTayal This method is still a duplicate of `reject!` https://github.com/RishabhTayal/fastlane/blob/f8b922f7848f2981a742bbad8dd611f1fbf7ee66/spaceship/lib/spaceship/tunes/tunes_client.rb#L1316-L1328

spaceship/spec/tunes/app_submission_spec.rb
```diff
@@ -58,5 +58,19 @@
       end.to raise_exception("Problem processing review submission.")
       expect(submission.submitted_for_review).to eq(false)
     end
+
+    it "successfully submits app for review if `reject_if_waiting_for_review` option is true" do
```
joshdholtz: Don't need this if you remove your duplicate method

spaceship/spec/tunes/tunes_stubbing.rb
```diff
@@ -154,6 +154,22 @@ def itc_stub_app_submissions_already_submitted
         to_return(status: 200, body: itc_read_fixture_file('app_submission/complete_failed.json'), headers: { 'Content-Type' => 'application/json' })
     end
 
+    def itc_stub_app_submission_reject_submission
```
joshdholtz: Don't need this if you delete your duplicate method

spaceship/lib/spaceship/tunes/app_submission.rb
```diff
@@ -142,7 +142,8 @@ def complete!
           )
         end
 
-        client.send_app_submission(application.apple_id, application.edit_version.version_id, raw_data_clone)
+        client.send_app_submission(application.apple_id, application.edit_version.version_id, raw_data_clone, reject_if_waiting_for_review)
```
joshdholtz: There isn't a reason to keep passing `reject_if_waiting_for_review` into here (as this is more business logic and less API client logic). I would take your logic you put in `send_app_submission` and actually put it into its own method in this class called `reject_if_waiting_for_review!`

joshdholtz (changes requested):
This is looking much better! I got a few more suggestions. Let's try to not touch `tunes_client.rb` at all. Instead,

1) Create new method `reject_if_waiting_for_review ` in `app_submission` with the reject logic
2) Remove `reject_if_waiting_for_review` parameter from `app_submission.rb`
3) Call your `submission.reject_if_waiting_for_reivew!` from `submit_for_review.rb`

This is more action based logic and less client based logic so I want to keep this as close to the action as possible - it makes it easier to maintain in the future 🙃

spaceship/spec/tunes/app_submission_spec.rb
RishabhTayal: Are you sure we won't need the test cases?

joshdholtz: I don't think we need this exact test if you remove all the logic out of the `complete` method and make a new method called `reject_if_waiting_for_review ` but we may need a new test case for that method

RishabhTayal: @joshdholtz  I am not very comfortable with removing the logic out of `complete` and creating a new method. I am not sure if I would be able to do that. I am going to keep the code as it is. Feel free to modify my branch and update it.

KrauseFx (approved):
Looks nice 👍

RishabhTayal:
👍 So excited about this PR. Can't wait for others to use this option. 😍😎.

Thanks @KrauseFx and @joshdholtz for your support!

RishabhTayal:
@KrauseFx I am testing the changes pushed by @joshdholtz with my test project. I think there is some issue with the logic. It's not working as expected. Will provide more details soon. For now don't merge this.

spaceship/lib/spaceship/tunes/application.rb
```diff
@@ -236,6 +236,15 @@ def ensure_version!(version_number, platform: nil)
         end
       end
 
+      def reject_version_if_possible!
+        can_reject = edit_version.can_reject_version
+        if can_reject
```
RishabhTayal: @joshdholtz What should be the right behaviour if we set `reject_if_waiting_for_review` as `true` but the app still can not be rejected because it's in 'can not reject` state? There is an intermediate state where the app is 'waiting for review' state but Apple doesn't let you reject the app.

deliver/lib/deliver/options.rb
```diff
@@ -114,6 +114,11 @@ def self.available_options
                                      description: "Submit the new version for Review after uploading everything",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :reject_if_waiting_for_review,
+                                     env_name: "DELIVER_REJECT_IF_WAITING_FOR_REVIEW",
```
RishabhTayal: @joshdholtz @KrauseFx  Do you think we should change the option name from `reject_if_waiting_for_review` to `reject_version_if_possible`? The app can be rejected in 3 states, 'waiting for review', 'in review' and 'Pending Release'. Naming it `reject_if_waiting_for_review` would be misleading to other devs.

spaceship/lib/spaceship/tunes/application.rb
RishabhTayal: I believe we will still get error saying `Problem processing review submission.` if the app is not in 'rejectable' state.

deliver/lib/deliver/runner.rb
```diff
@@ -45,8 +45,9 @@ def run
 
       UI.success("Finished the upload to iTunes Connect") unless options[:skip_binary_upload]
 
-      precheck_success = precheck_app
+      reject_version_if_possible if options[:reject_if_waiting_for_review]
 
+      precheck_success = precheck_app
       submit_for_review if options[:submit_for_review] && precheck_success
```
RishabhTayal: When we call the `deliver` action without `reject_if_waiting_for_review`, and the app is in `can_reject` state, do you think we should show a message suggesting they can use this option to reject the version?

spaceship/lib/spaceship/tunes/application.rb
RishabhTayal: Is that ok?

gerbil802:
@RishabhTayal If the app cannot be rejected right now:
1. On one hand the script _must_ fail, because it's supposed to reject the version
2. Failing sucks and beats the purpose of Fastlane, so failing is not an option

My idea is to wait (just like Deliver and Pilot wait) until the binary is ejectable, then reject it!

RishabhTayal:
@gerbil802 

> My idea is to wait (just like Deliver and Pilot wait) until the binary is ejectable, then reject it!

I will leave that upto Felix and Josh.

deliver/lib/deliver/options.rb
joshdholtz: I think this makes sense ^

spaceship/lib/spaceship/tunes/application.rb
joshdholtz: I think this is okay the way it is now for that situation ^

deliver/lib/deliver/runner.rb
joshdholtz: I will change this and give it a shot ^

joshdholtz (approved):

PR #10732, opened 2017-10-29
[Deliver] Add functionality to set available territories from deliver

matthewellis:
### Checklist

### Motivation and Context
Deliver is currently missing the feature to manage territory availability.

### Description
This PR adds the feature to manage availability from deliver using a territory.yml file or options passed. Deliver setup now downloads the current availability from iTunes connect and dumps it to file, a download_availability command has also been added to trigger this.


jetersen (approved):
This is awesome :)
Think this PR requires a docs PR too :)
Would love a list of country codes, and an example
https://github.com/fastlane/docs/blob/master/docs/actions/deliver.md

deliver/lib/deliver/options.rb
```diff
@@ -205,6 +205,20 @@ def self.available_options
                                      is_string: false,
                                      default_value: :warn),
 
+        # Territory availability
+        FastlaneCore::ConfigItem.new(key: :availability_all_territories,
+                                     description: "Set availability to all territories",
+                                     is_string: false,
+                                     default_value: false),
+        FastlaneCore::ConfigItem.new(key: :availability_territories,
```
ohayon: Does this option and the one below confilct with `availability_all_territories`?

ohayon (commented):
This is really sweet @matthewellis !! I just left one question in the review. Let me know your thoughts! 😄

deliver/lib/deliver/options.rb
matthewellis: Indeed it does, this will take preference over the lists, a validation function was added to the Availability class that will warn the user if any conflicts are detected when uploading.

If you have any ideas on how to make this more clear to users i would be happy to make changes.

ohayon: I think you can pass `conflicting_options: [:available_all_territories]` to the ConfigItem initializer?

jdee:
My only concern here is maintaining a large list of territories. I like the options to the action: `availability_all_territories`, `availability_exclude_territories`, `availability_territories`. I'm worried that someone may run `download_territories` to get a list of all of them and then remove a few rather than using the `availability_exclude_territories` option, which can lead to trouble if new App Stores are added, or if they accidentally change something, which is hard to detect in a large list. If it were me, I'd probably make the `territories.yml` file informational only, since the list is too large to display at the terminal, but I'd ignore the contents when running `deliver`. Whatever the contents of the file, they can always be specified in the Deliverfile using the `availability_territories` option (even using something like `YAML.load_file` from within the Deliverfile), and then you don't have the risk of conflict between the file contents and other options.

Of course, I may be missing a key use case that this satisfies.

jdee:
I guess I have one more thought: What is the current behavior? And what will the behavior be if none of these options is present, i.e. no `availability_all_territories`, `availability_territories` or `availability_exclude_territories` and no `territories.yml`? I would expect the behavior to be unchanged.

I guess I'm not entirely clear on whether this submits to multiple App Stores or just limits distribution when you submit to one App Store. If it's the latter, then I think the default behavior is probably just all territories. If that's correct, I would at least remove the `availability_all_territories` option, since that's the default. In fact you _could_ get away with a single `availability` option, e.g.:

```Ruby
availability: :all # default, but might be useful at other times
```

```Ruby
availability: [:all, except: %w{US CA}] # this is one of those times
```

```Ruby
availability: %w{GB DE FR JP}
```

```Ruby
availability: {file: "territories.yml"}
```

And in the future (if it makes sense) you might even be able to do things like:

```Ruby
availability: [[:americas, except: %w{US MX}], :pacrim]
```

```Ruby
availability: [:all, except: :emea]
```

This is just to illustrate how you could flexibly reuse a single option, which could make it easier to use, though you have more code to write to interpret the value.

The main concern is just default behavior and avoiding unnecessary, possibly confusing and conflicting options.

matthewellis:
@jdee Hey sorry for the delay my aim for this feature was to make it work in a very similar way to the metadata upload, due to this the options are given higher priority than the territories.yml file, if neither are provided the upload is skipped to maintain the current behaviour.  I do prefer including loading for this file as i think its valuable to keep use of this clean and consistent with using metadata.

The use case i had in mind is being able to commit the list of availability much like metadata so it can too be version controlled, the 3 options came from the 3 most likely scenarios i could think of for this. 
1. The developer just wants to make all available
2. The developer has a small list of territories the app is available
3. The developer wants to release to all but a few territories

I do agree that the list could become quite large but some care has been taken to avoid this, if the user downloads the current availability via setup or the manual command the smallest list will be created. This is done by detecting first if all the possible territories are selected resulting in the all_territories option being used. 

Otherwise if a list must be used exclude will be used by inverting the input list if the number of territories to enable is less then half of a total, otherwise the list is used unmodified as availability_territories. 

Also just to clarify this feature is just to manage where the app listing is available, it does not interact with the upload of a app. I do like the idea of using only one parameter but this is quite different than anything else i have seen in fastlane and would be very interested to see if anyone else in the team has some thoughts on that.

Hope this answers most of your questions 🚀

jdee:
Hi @matthewellis.

If I understand the argument correctly, the idea is that if the Deliverfile has `availability_exclude_territories: "US"`, when you go back to look at a tag, you don't know exactly which territories it was distributed to, since you don't know the full list at that time. I think that's enough to convince me that this has some value. I still think there's some risk, but I guess it's up to the user.

I'm still not sure of the current default behavior, but I suspect it's something like submit to one App Store but allow distribution to all available territories. In that case, the `availability_all_territories` option wouldn't seem to do anything. If it's true, false or absent, the behavior would be the same.

If you decide to go with a single option, this would kind of be moot. You could support `availability: :all` to be explicit or to allow an `:except` clause without having a separate option.

jdee:
One thought that might help improve the metadata approach would be to have something like a diff option to your `download_availability` command, which would download the list and instead of populating the file, show any territories not present in the file. That might help make the file more maintainable.

matthewellis:
Closing for now, will reopen when feature is complete

PR #10768, opened 2017-11-01
Add auto_release_date option to deliver

taquitos:
We have `auto_release_date` on the app version so let's create an option so we can pass it through 🎈

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -111,6 +111,15 @@ module Deliver
                                      description: "Should the app be automatically released once it's approved?",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :auto_release_date,
+                                     env_name: "DELIVER_AUTO_RELEASE_DATE",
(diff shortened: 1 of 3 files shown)


deliver/lib/deliver/options.rb
```diff
@@ -111,6 +111,15 @@ def self.available_options
                                      description: "Should the app be automatically released once it's approved?",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :auto_release_date,
+                                     env_name: "DELIVER_AUTO_RELEASE_DATE",
+                                     description: "Time to release when release on approval is used in milliseconds GMT",
```
ohayon: Not sure I fully understand what this description is supposed to mean. Could we change it to say something like "Date in milliseconds on which to automatically release pending approval"?

taquitos: yeah, I basically copy/pasted a comment from the app_version class. Your text is much nicer :)

ohayon (approved):

PR #10809, opened 2017-11-06
Fixed instruction for update plugin

RishabhTayal:
### Motivation and Context
If you are using plugins in your `fastlane` setup and the used plugin have a new version available, it will show the message `To update all plugins, just run fastlane update_plugins` to update the plugin. From my understanding, when using the plugin we always need to use `bundle exec` since the `Gemfile` is to be used.

The instructions should change to `To update all plugins, just run bundle exec fastlane update_plugins`. This PR changes the instructions shown to the user to update the plugin.


KrauseFx (approved):
Awesome, nice catch, thanks 👍

PR #10823, opened 2017-11-07
prevent nil crasher in associated_groups

jcarroll-mediafly:
Key may be present while value is nil, in these cases this method will crash on nil

### Checklist

### Motivation and Context
When running SpaceShip::Portall.app.all and using output lane would crash as associated_groups key was present but value was nil

### Description
Simply preventing nil crasher


KrauseFx (approved):
Yeah, great catch, thanks @jcarroll-mediafly 👍

Issue #10812, opened 2017-11-06
503 returned when calling produce


jcarroll-mediafly:
### New Issue Checklist

- [√] Updated fastlane to the latest version
- [√] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [√] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [√] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Unsure if this is an apple issue or fastlane. But I'm constantly getting 503 when using `produce`. Had my team members run the lane as well and same issue. I found an older ticket https://github.com/fastlane/fastlane/issues/8209 with similar problems, and it seems unclear if its an apple thing or fastlane even. Of course apple services are all green at the moment.

I will say we have ~50 app ids and periodically we run this lane to ensure the extensions for each app id has the correct app id, services enabled etc turned on / configured correctly. Basically its a lane to ensure dev portal is up to snuff without having to click 5000 times or make a human mistake of reading or changing something incorrectly.

Anyone else have 503 issues? Our latest guess is apple is possibly blocking / rate limiting us.

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.6                                     |
| Ruby                        | 2.2.3                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.15.0                          |
| Installation Source         | ~/.rvm/gems/ruby-2.2.3/bin/fastlane         |
| Host                        | Mac OS X 10.12.6 (16G1036)                  |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.2.3/lib                |
| OpenSSL Version             | OpenSSL 1.0.2h  3 May 2016                  |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 8.3.3                                       |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
[...]

Issue #10793, opened 2017-11-03
iTunes account blocked/corrupted by fastlane sigh download


elorz007:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

After using `fastlane sigh download` to manually download provisioning profiles, the itunes account that was used gets somehow blocked. When signing in to appleid.apple.com I get the error "**Failed to verify your identity. Try again**" and the page stays with the loading indicator forever. 

![screen shot 2017-11-03 at 18 36 32](https://user-images.githubusercontent.com/3601038/32388801-1bda896c-c0c9-11e7-95c5-45314b59091c.png)

With the same account I can log in to developer.apple.com but **not** to itunesconnect.apple.com

After some time or some action that I cannot pinpoint, the account works again (asks for the security questions every time you log in but it works). Then `fastlane sigh download` works a couple of times and right after that the accounts gets blocked again.

I tried with a complete new apple id, registered everything in the development portal, etc. Then I used `fastlane sigh download` couple of times, it worked and then the **new account also got blocked**. 

Needless to say, I tried resetting the password (successfully), different computers, etc. Didn't help.

When the account is in this blocked state, when using `fastlane sigh download` I get the following error:

```
[18:28:16]: ▸ [18:28:16]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[18:28:18]: ▸ +-------------------------------------+-------------------------------------------+
[18:28:18]: ▸ |                             Summary for sigh 2.63.0                             |
[18:28:18]: ▸ +-------------------------------------+-------------------------------------------+
[18:28:18]: ▸ | development                         | true                                      |
[18:28:18]: ▸ | provisioning_name                   | <Redacted>      
[18:28:18]: ▸ | username                            | <Redacted>                    
[18:28:18]: ▸ | app_identifier                      | <Redacted> 
[18:28:18]: ▸ | skip_certificate_verification       | true                                      |
[18:28:18]: ▸ | team_id                             | <Redacted>                                |
[18:28:18]: ▸ | adhoc                               | false                                     |
[...]

PR #10826, opened 2017-11-07
Fix login rate limiting

taquitos:
implemented new endpoints for login
Fix for #10793, also #10812

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -444,7 +444,8 @@ module Spaceship
           req.url "https://idmsa.apple.com/IDMSWebAuth/authenticate"
           req.body = "appIdKey=891bd3417a7776362562d2197f89480a8547b108fd934911bcbea0110d07f757&appleId=#{user}&accountPassword=#{password}"
           req.headers['Content-Type'] = 'application/x-www-form-urlencoded'
-          req.headers['Accept-Encoding'] = 'deflate, gzip'
+          # req.headers['Accept-Encoding'] = 'deflate, gzip'
+          req.headers['Accept'] = 'application/json, text/javascript'
           req.headers["Cookie"] = modified_cookie if modified_cookie
         end
       rescue UnauthorizedAccessError
@@ -456,10 +457,17 @@ module Spaceship
       case response.status
       when 403
         raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
-      when 200, 302
+      when 200
+        # 2 factor requires HTML parsing now...
+        # When 2 factor is needed, it's a 200 instead of a 302 redirect
+        handle_two_step(response)
+        fetch_olympus_session
+        return response
+      when 302
         fetch_olympus_session
         return response
       when 409
+        # Can probably remove this
         # 2 factor is enabled for this account, first handle that
         # and then get the olympus session
         handle_two_step(response)
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -420,12 +420,6 @@ module Spaceship
       # If this is a CI, the user can pass the session via environment variable
       load_session_from_env
 
-      data = {
-        accountName: user,
-        password: password,
-        rememberMe: true
-      }
-
       begin
         # The below workaround is only needed for 2 step verified machines
         # Due to escaping of cookie values we have a little workaround here
@@ -447,12 +441,10 @@ module Spaceship
         end
 
         response = request(:post) do |req|
-          req.url "https://idmsa.apple.com/appleauth/auth/signin"
-          req.body = data.to_json
-          req.headers['Content-Type'] = 'application/json'
-          req.headers['X-Requested-With'] = 'XMLHttpRequest'
-          req.headers['X-Apple-Widget-Key'] = self.itc_service_key
-          req.headers['Accept'] = 'application/json, text/javascript'
+          req.url "https://idmsa.apple.com/IDMSWebAuth/authenticate"
+          req.body = "appIdKey=891bd3417a7776362562d2197f89480a8547b108fd934911bcbea0110d07f757&appleId=#{user}&accountPassword=#{password}"
+          req.headers['Content-Type'] = 'application/x-www-form-urlencoded'
+          req.headers['Accept-Encoding'] = 'deflate, gzip'
           req.headers["Cookie"] = modified_cookie if modified_cookie
         end
       rescue UnauthorizedAccessError
@@ -464,7 +456,7 @@ module Spaceship
       case response.status
       when 403
         raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
-      when 200
+      when 200, 302
         fetch_olympus_session
         return response
       when 409
@@ -488,13 +480,17 @@ module Spaceship
 
     # Get the `itctx` from the new (22nd May 2017) API endpoint "olympus"
     def fetch_olympus_session
-      response = request(:get, "https://olympus.itunes.apple.com/v1/session")
+      response = request(:get, "https://itunesconnect.apple.com/olympus/v1/session")
       if response.body
         user_map = response.body["user"]
         if user_map
           self.user_email = user_map["emailAddress"]
         end
       end
+    end
+
+    def logout
+      request(:get, "https://itunesconnect.apple.com/logout")
     end
 
     def itc_service_key
@@ -504,7 +500,7 @@ module Spaceship
       itc_service_key_path = "/tmp/spaceship_itc_service_key.txt"
       return File.read(itc_service_key_path) if File.exist?(itc_service_key_path)
 
-      response = request(:get, "https://olympus.itunes.apple.com/v1/app/config?hostname=itunesconnect.apple.com")
+      response = request(:get, "https://itunesconnect.apple.com/olympus/v1/app/config?hostname=itunesconnect.apple.com")
       @service_key = response.body["authServiceKey"].to_s
 
       raise "Service key is empty" if @service_key.length == 0
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -440,11 +440,15 @@ module Spaceship
           modified_cookie.gsub!(unescaped_important_cookie, escaped_important_cookie)
         end
 
+        require 'uri'
+        encoded_password = URI.encode(password)
+        encoded_apple_id = URI.encode(user)
+
         response = request(:post) do |req|
           req.url "https://idmsa.apple.com/IDMSWebAuth/authenticate"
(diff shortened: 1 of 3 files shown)


KrauseFx (approved):

spaceship/lib/spaceship/client.rb
```diff
@@ -447,12 +441,10 @@ def send_shared_login_request(user, password)
...
 
         response = request(:post) do |req|
-          req.url "https://idmsa.apple.com/appleauth/auth/signin"
-          req.body = data.to_json
-          req.headers['Content-Type'] = 'application/json'
-          req.headers['X-Requested-With'] = 'XMLHttpRequest'
-          req.headers['X-Apple-Widget-Key'] = self.itc_service_key
-          req.headers['Accept'] = 'application/json, text/javascript'
+          req.url "https://idmsa.apple.com/IDMSWebAuth/authenticate"
+          req.body = "appIdKey=891bd3417a7776362562d2197f89480a8547b108fd934911bcbea0110d07f757&appleId=#{user}&accountPassword=#{password}"
```
casz: How did we arrive on this magic appIdKey ❓

taquitos: Great question, I cleared my history and navigated to developer.apple.com
then I clicked on `account` which took me to the login page `https://idmsa.apple.com/IDMSWebAuth/login?appIdKey=891bd3417a7776362562d2197f89480a8547b108fd934911bcbea0110d07f757&path=%2Faccount%2F&rv=1`
I then tracked the `POST` request and that `appIdKey` was included.
I resent the request without the appIdKey and it failed, so I added it back in and the request succeeded.
My guess is that Apple is using SSO now, and each service they operate will have an "appIdKey" associated with it.

taquitos: We might be able to derive this `appIdKey` from somewhere, but I didn't find it yet.

joshdholtz (approved):
Might be good to test with 2FA before this 🚢 s but code looks 👌

jetersen (approved):
🚀

taquitos:
I'm going to do some 2FA testing real quick

spaceship/lib/spaceship/client.rb
lacostej: If you google for it, you will find plenty of references on other's people scripts :)

taquitos:
2FA doesn't work, looking into that

joshdholtz (changes requested):
2FA 😇

spaceship/lib/spaceship/client.rb
lacostej: If you get https://developer.apple.com/account/, you get a 302 to the login page you hardcoded.

youtux: Does this mean we can’t use passwords that contain `=` or `&`?

taquitos:
I'm going to hold off on merging this until:

- We fix urlencoding of username/password
- 2FA is fixed

2FA is important because if we ship this right now, it will break everybody with it 😢 
It's a bit complicated though because we have to switch over from ITC 2FA to the Apple SSO 2FA which is significantly different. I'm about 50% of the way there, but it could take another day.

spaceship/lib/spaceship/two_step_client.rb
```diff
@@ -131,36 +83,17 @@ def select_device(r, device_id)
...
-      # Send token back to server to get a valid session
+      # Send code back to server to get a valid session
       r = request(:post) do |req|
-        req.url "https://idmsa.apple.com/appleauth/auth/verify/device/#{device_id}/securitycode"
-        req.body = { "code" => code.to_s }.to_json
-        req.headers['Content-Type'] = 'application/json'
-
-        update_request_headers(req)
+        req.url "https://idmsa.apple.com/IDMSWebAuth/validateSecurityCode"
+        req.body = "digit1=#{code[0]}&digit2=#{code[1]}&digit3=#{code[2]}&digit4=#{code[3]}&scnt=#{scnt}"
```
casz: why 🍎, would go away from a JSON post
kek https://github.com/amyworrall/QuickRadar/blob/8326a52ba49f47c1faad59117ead6935987f91e0/QuickRadar/QRRadarSubmissionService.m#L247
https://github.com/clayfreeman/IDMSWebAuth/blob/f5e3a6ed8be4bc76f53f7280209faaaa8d40823d/include/IDMSWebAuth.php#L147
seems like you should handle between 4 or 6 digit.

Could we do some kind of array and then just join on '&'

spaceship/lib/spaceship/client.rb
```diff
@@ -447,11 +441,10 @@ def send_shared_login_request(user, password)
         end
 
         response = request(:post) do |req|
-          req.url "https://idmsa.apple.com/appleauth/auth/signin"
-          req.body = data.to_json
-          req.headers['Content-Type'] = 'application/json'
-          req.headers['X-Requested-With'] = 'XMLHttpRequest'
-          req.headers['X-Apple-Widget-Key'] = self.itc_service_key
+          req.url "https://idmsa.apple.com/IDMSWebAuth/authenticate"
```
lacostej: @taquitos WRT to the hardcoded appID, could we add this as documentation?

# GET https://developer.apple.com/account/ returns a 302 redirect to idmsa.apple.com...

lacostej: Also there was a question from @youtux WRT encoding passwords.

https://github.com/fastlane/fastlane/pull/10826/files/8bf8bb3d3c4232356a035d4e8209756ec6a70c1f#r149506222

spaceship/lib/spaceship/client.rb
casz: I assume we have to url encode the password

```ruby
require 'uri'
URI.encode(password)
```

youtux: and the username as well

taquitos: Thank you, updated!

spaceship/lib/spaceship/two_step_client.rb
```diff
@@ -116,66 +63,54 @@ def self.spaceship_session_env
...
 
-    def select_device(r, device_id)
+    def select_device(scnt, device_id)
       # Request Token
-      r = request(:put) do |req|
-        req.url "https://idmsa.apple.com/appleauth/auth/verify/device/#{device_id}/securitycode"
-        update_request_headers(req)
+      response = request(:post) do |req|
+        req.url "https://idmsa.apple.com/IDMSWebAuth/generateSecurityCode"
+        req.body = "deviceIndex=#{device_id}&scnt=#{scnt}"
```
joshdholtz: @taquitos Should we uri encode `device_id` and `scnt` as well? I don't know _exactly_ what these values look like but better safe than sorry if we don't have full control over this input?

spaceship/lib/spaceship/two_step_client.rb
```diff
@@ -116,66 +63,54 @@ def self.spaceship_session_env
...
-          return select_device(r, device_id)
-        end
-
-        raise ex
+      request_body << "scnt=#{scnt}"
+
+      # Send code back to server to get a valid session
+      response = request(:post) do |req|
+        req.url "https://idmsa.apple.com/IDMSWebAuth/validateSecurityCode"
+        req.body = body.join('&')
```
joshdholtz: Insert another dumb uri encode question here 😇

joshdholtz (commented):
A few thoughts on this great work of yours 💪

taquitos:
Posting for visibility, I updated the issue: https://github.com/fastlane/fastlane/issues/10793#issuecomment-342906293

spaceship/lib/spaceship/two_step_client.rb
```diff
@@ -116,66 +63,54 @@ def self.spaceship_session_env
...
-      # we use `Spaceship::TunesClient.new.handle_itc_response`
-      # since this might be from the Dev Portal, but for 2 step
-      Spaceship::TunesClient.new.handle_itc_response(r.body)
+      # if the hidden rate limit message isn't there, that means it is being shown, so we have an error
+      hidden_rate_limit_regex = /<div\s*id="tooManyCodesSentError"\s*style="display:\snone">/
+      hidden_rate_limit_error_matches = response.body.scan(hidden_rate_limit_regex)[0]
+      raise "Too many codes sent, enter the last code you received, use a different device, or try again later" if hidden_rate_limit_error_matches.nil?
 
       puts "Successfully requested notification"
       code = ask("Please enter the 4 digit code: ")
```
casz: Should the 4 digit not be updated?

spaceship/lib/spaceship/client.rb
```diff
@@ -446,12 +440,15 @@ def send_shared_login_request(user, password)
...
+
         response = request(:post) do |req|
-          req.url "https://idmsa.apple.com/appleauth/auth/signin"
-          req.body = data.to_json
-          req.headers['Content-Type'] = 'application/json'
-          req.headers['X-Requested-With'] = 'XMLHttpRequest'
-          req.headers['X-Apple-Widget-Key'] = self.itc_service_key
+          req.url "https://idmsa.apple.com/IDMSWebAuth/authenticate"
+          # appIdKey is hardcoded because that seems to the be an internal SSO service identifier.
+          req.body = "appIdKey=891bd3417a7776362562d2197f89480a8547b108fd934911bcbea0110d07f757&appleId=#{encoded_apple_id}&accountPassword=#{encoded_password}"
```
casz: for readability could we put the apiIdKey into a variable?

taquitos:
While the PR fixes this for the simple login case, it only addresses SMS 2FA, and it employs web scraping, which is super flakey. 

Furthermore, the authentication logic we already have addresses so many corner cases, and is battle tested. Replacing that with this flakey update could cause a lot of pain for everybody.

Instead, we think the right thing moving forward is to assume Apple isn’t going to fix this obvious issue, and then focus our energy on improving our session caching, so that between various actions, we don’t login multiple times unless explicitly called for.

This will take some time to properly address, but we’ll have a plan soon and we’ll share that out as soon as we do 🚀. In the meantime, we’re going to keep this PR alive if it is helping to unblock people.

We have some workarounds whose success seem to vary:

- Login using a different account from same IP
- Change IP address (VPN/Proxy)
- Login from a different IP, it unblocks your original IP.
- Login in developer.apple.com and then navigate to itunesconnect.apple.com

spaceship/lib/spaceship/client.rb
bolom: it's possible to use
encoded_password = CGI.escape(password)
 encoded_apple_id = CGI.escape(user)

when you have "+" in user. it's not working

spaceship/lib/spaceship/client.rb
KrauseFx: I don't think we're actually allowed to have that API key as part of the fastlane code base

KrauseFx:
Hey everyone, thanks again for looking into this issue. Sorry it took so long, I finally found some time to look further into this issue, and was able to reproduce the issue and prepare a fix. You can check out the PR over here https://github.com/fastlane/fastlane/pull/11108

I'd love to ship this as a new release tomorrow morning 👍

PR #10876, opened 2017-11-10
Fastlane core action aliases

mpirri:
### Checklist

### Motivation and Context

New users to fastlane are often perplexed by the core action names in Fastfiles (e.g., `gym`, `cert`, `sigh`). This PR adds new verb/noun-style aliases for each of the core actions. Over time we will change over documentation to use these new aliases for the core actions.

### Description

Core action aliases are implemented as simple subclasses of the core action. Changes include more clear descriptions and a simple message when running the actions to help users understand that the core action/tool is being invoked to complete the step.

Unit tests were changed to omit the option check from these new aliases and also to allow multi-level subclassing within actions.


getaaron:
What do you think about the inverse approach? Rename the core actions, and create aliases to continue supporting the old names.

I think that would be less confusing for new users, and also would make it easier for us to eventually deprecate the old names by printing deprecation messages in the subclass.

jetersen (approved):
Even as one year in, user.
I can appreciate the change of names, and the inverse sounds to me like the better option. I never liked trying to explain gym, deliver or snapshot tool to any one. Also pilot did not help much either.

However there is also case of Match which always been easy to explain. Sometimes words just fit :)

Pilot has also been challenging cause people get confused. Because they were already confusing enough with the whole TestFairy and TestFlight name confusion, so glad we switched to HockeyApp.

mpirri:
@getaaron Interesting thought - from the standpoint of the user, is there an advantage of moving the code to the new aliases and then back-aliasing to the original actions? These seem functionally equivalent, but I could very well be missing something.  🤔 

My expectation is that the original core actions will stick around for a really long time, since removing them would be enough of a disruption to existing users so as not to be worth it. As such, we (probably) don't need to plan for deprecation at this point. And we can always change/move around the code later, of course.

The advantage that we saw to the subclassing approach was that it would help users of these new aliases understand what is going on under the hood (upload_to_testflight == pilot, build_app == gym, etc) which still seems like a necessary skill to help users level-up with fastlane later. As we make other changes to make the original tool names less important, we can revisit this implementation decision later.

Thoughts?

mpirri:
@getaaron Okay, I've changed my mind - I really like your suggestion. I think it will position us well for the future, and the core actions are pretty confusing, in my opinion (more so than the actual command-line tools, since they only do one of the commands of the tool). So, I'm going to make the following changes:
* change the name/class of each of the core actions to the new readable alias name
* create new subclasses back-aliasing each of the new actions to the old core action name

I'll post a new commit in the next few hours - thanks again for the awesome suggestion! 🚀

mpirri:
@getaaron Okay, I pushed up the refactor. Is this what you had in mind?

fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

getaaron:
@mpirri this looks great. I added a few minor suggestions / ideas to consider

mpirri:
@getaaron Thanks for the feedback!

Quick update - I'm going to rebuild this PR in a way that will make GitHub realize that what I"m doing is a rename/edit rather than an add/delete that is making this hard to read.

mpirri:
I've created a new PR (hopefully more readable!) at #10939.

PR #10886, opened 2017-11-13
fix the checking of URLs containing non-ASCII characters and a fragment identifier

tattn:
### Checklist

### Motivation and Context
This change correctly checks whether you can access that URLs that contain non-ASCII characters and fragment identifiers. e.g. https://foobar.com/テスト#anchor

### Description
Before:   

use `URI.encode` only

```ruby
uri = URI.encode('https://foobar.com/テスト#anchor') # dummy url
uri  # => https://foobar.com/%E3%83%86%E3%82%B9%E3%83%88%23anchor
request = Faraday.new(uri) do |connection|
  connection.use FaradayMiddleware::FollowRedirects
  connection.adapter :net_http
end
request.head.status == 200 # => false
```

After:  

remove a fragment identifier

```ruby
uri = Addressable::URI.parse('https://foobar.com/テスト#anchor')
uri.fragment = nil # remove a fragment identifier
uri = URI.encode(uri.to_s)
uri # => https://foobar.com/%E3%83%86%E3%82%B9%E3%83%88
request = Faraday.new(uri) do |connection|
  connection.use FaradayMiddleware::FollowRedirects
  connection.adapter :net_http
end

request.head.status == 200 # => true
```

The reason I used `Addressable::URI.parse` is that` URI.parse` is not support for non-ASCII characters.

```ruby
url = 'https://foobar.com/テスト#anchor'
URI.parse(url)
# => /.../uri/rfc3986_parser.rb:21:in `split': URI must be ascii only "https://foobar.com/\u30C6\u30B9\u30C8#anchor" (URI::InvalidURIError)
```

### DIFF
--- a/precheck/lib/precheck/rules/unreachable_urls_rule.rb
+++ b/precheck/lib/precheck/rules/unreachable_urls_rule.rb
@@ -24,7 +24,9 @@ module Precheck
         return RuleReturn.new(validation_state: Precheck::VALIDATION_STATES[:failed], failure_data: "empty url") if url.empty?
 
         begin
-          request = Faraday.new(URI.encode(url)) do |connection|
+          uri = Addressable::URI.parse(url)
+          uri.fragment = nil
+          request = Faraday.new(URI.encode(uri.to_s)) do |connection|
             connection.use FaradayMiddleware::FollowRedirects
             connection.adapter :net_http
           end


tattn:
I signed it!

getaaron (approved):
Nice catch, looks good to me. Do we need to make a similar change anywhere else in fastlane?

For anyone interested, more discussion is available [on this related Ruby issue](https://bugs.ruby-lang.org/issues/12852).

taquitos:
@tattn do you have any sample code that could be used in an rspec to catch this in the future?

getaaron:
I think a good test could use the URL from [this test](https://github.com/fastlane/fastlane/blob/41e84cb96c63995e35fe10508bf30bf88a6bb897/precheck/spec/rules/unreachable_urls_rule_spec.rb#L29-L36) but append `#anchor` to the URL

taquitos (approved):
🎈:goat:

tattn:
thanks :smile:

I will send PRs again and add a test for a fragment identifier if I find something else.

PR #10892, opened 2017-11-13
[WIP] Add verification of Xcode project as part of precheck

KrauseFx:
To give this a try, do the following

- Use a sample Xcode project
- Add a build phase with the content, replace the path to your `bundle` binary (TODO, this should not be necessary)
```
FASTLANE_DISABLE_COLORS=1 /Users/fkrause/.rbenv/shims/bundle exec fastlane verify_firebase
```
- Update your `Gemfile` to point to  your local fastlane code base, with this branch checked out and run `bundle update`
- Currently we only check if we can find the plist file, and if it has some content (at least one row) and that it can be parsed, check out the [full class here](https://github.com/davidair/fastlane/blob/master/precheck/lib/precheck/rules/verify_plist_rule.rb#L40)
- To run it manually from outside of Xcode (useful for debugging reasons) you can use the following, with of course your actual Plist file path and Xcode project path. This way you can iterate faster, as it doesn't require you to re-compile the every every single time
```
PROJECT_FILE_PATH=./Themoji.xcodeproj INFOPLIST_FILE =./Themoji/Info.plist  bundle exec fastlane verify_firebase
```


taquitos:
so clean

davidair:
The environment variable to specify the plist is actually INFOPLIST_FILE. Can you please update the description?

taquitos:
@davidair our naming conventions generally favor `INFOPLIST_PATH` when talking about a path to a file

KrauseFx:
Updated 👍

Gemfile
```diff
@@ -2,6 +2,8 @@ source "https://rubygems.org"
 
 gem "xcode-install", ">= 2.2.1" # needed for running xcode-install related tests
 
+gem "xcodeproj", ">= 1.5.3"
```
KrauseFx: This should already be a dependency of fastlane I think, check out `fastlane.gemspec` 👍

precheck/spec/fixtures/Common/verify-firebase.sh
```diff
@@ -0,0 +1,23 @@
...
+  # Then try to load from a root install
+  source "/usr/local/rvm/scripts/rvm"
+else
+  printf "ERROR: An RVM installation was not found.\n"
+  exit 128
+fi
+
+# Actually run fastlane from the fastlane root directory.
+cd ../../../
+bundle exec fastlane verify_firebase
```
KrauseFx: In our experience, lots of people just use system Ruby and don't have rvm, and also some people use rbenv over rvm, do we need to add code to cover those also?

kapfab:
`gym` seems to expect a project file in its prechecks even when it is requested to export from an existing archive.
As a workaround, we had to add a dumb Xcode project file to our lanes in order to able to export IPAs.
We will report an issue so that `skip_build_archive` option does not look for a useless project file, but I understand this PR will make things even boring for our case until this issue is fixed…

joshdholtz:
@KrauseFx You still working on this? 😊

KrauseFx:
Ok I don't think we'll go ahead with this, let's close it

PR #10898, opened 2017-11-13
When building plugin scores, we should ignore repos that are private

taquitos:
sets GIT_TERMINAL_PROMPT=0 when we build plugin scores


KrauseFx (approved):
So good, how did you find out about `GIT_TERMINAL_PROMPT`?

taquitos:
@KrauseFx it just came to me in a vision while I was sleeping last night.

KrauseFx:
So visionary

Issue #9826, opened 2017-07-21
Can't save IAP details


jgongo:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

I'm trying to use Spaceship to update IAP products. Every time that I try I get a `ITC.response.error.OPERATION_FAILED`. I'm a Ruby newbie, so maybe I'm missing anything! I try to do the following:

```
require 'Spaceship'
Spaceship::Tunes.login('########', '#########')
iap = Spaceship::Tunes::Application.all.first.in_app_purchases.all.first.edit
iap.save!
```

The iTunes user I'm using to login has the App Manager role, so it should be ok, no?

##### Complete output when running fastlane, including the stack trace and command used
> You can use: `--capture_output` as the last commandline argument to get that collected for you

```
$ bundle exec ruby scripts/updateIAP.rb 
/Users/jose.gonzalez/Documents/projects/rakuten/ios/ios-v3/vendor/bundle/ruby/2.4.0/gems/fastlane-2.47.0/spaceship/lib/spaceship/tunes/tunes_client.rb:162:in `handle_itc_response': ITC.response.error.OPERATION_FAILED (Spaceship::TunesClient::ITunesConnectError)
	from /Users/jose.gonzalez/Documents/projects/rakuten/ios/ios-v3/vendor/bundle/ruby/2.4.0/gems/fastlane-2.47.0/spaceship/lib/spaceship/tunes/tunes_client.rb:964:in `block in update_iap!'
	from /Users/jose.gonzalez/Documents/projects/rakuten/ios/ios-v3/vendor/bundle/ruby/2.4.0/gems/fastlane-2.47.0/spaceship/lib/spaceship/tunes/tunes_client.rb:1296:in `with_tunes_retry'
	from /Users/jose.gonzalez/Documents/projects/rakuten/ios/ios-v3/vendor/bundle/ruby/2.4.0/gems/fastlane-2.47.0/spaceship/lib/spaceship/tunes/tunes_client.rb:958:in `update_iap!'
	from /Users/jose.gonzalez/Documents/projects/rakuten/ios/ios-v3/vendor/bundle/ruby/2.4.0/gems/fastlane-2.47.0/spaceship/lib/spaceship/tunes/iap_detail.rb:211:in `save!'
	from scripts/updateIAP.rb:26:in `<main>'

```

[...]

Issue #10743, opened 2017-10-30
Modifying In-App Purchase fails with ITC.response.error.OPERATION_FAILED - spaceship


wufire:
### Issue Description

Following the same steps found on this issue:  https://github.com/fastlane/fastlane/issues/10216

I still see the OPERATION_FAILED itc response.

```
irb(main):005:0> p = app.in_app_purchases.all.first.edit
=> <Spaceship::Tunes::IAPDetail
...
>
irb(main):006:0> p.save!
Spaceship::TunesClient::ITunesConnectError: ITC.response.error.OPERATION_FAILED
	from /usr/local/lib/ruby/gems/2.4.0/gems/fastlane-2.62.1/spaceship/lib/spaceship/tunes/tunes_client.rb:212:in `handle_itc_response'
	from /usr/local/lib/ruby/gems/2.4.0/gems/fastlane-2.62.1/spaceship/lib/spaceship/tunes/tunes_client.rb:1074:in `block in update_iap!'
	from /usr/local/lib/ruby/gems/2.4.0/gems/fastlane-2.62.1/spaceship/lib/spaceship/tunes/tunes_client.rb:1406:in `with_tunes_retry'
	from /usr/local/lib/ruby/gems/2.4.0/gems/fastlane-2.62.1/spaceship/lib/spaceship/tunes/tunes_client.rb:1068:in `update_iap!'
	from /usr/local/lib/ruby/gems/2.4.0/gems/fastlane-2.62.1/spaceship/lib/spaceship/tunes/iap_detail.rb:205:in `save!'
	from (irb):6
	from /usr/local/bin/irb:11:in `<main>'
```

#### Environment
<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.6                                     |
| Ruby                        | 2.4.2                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.9.2                           |
| Installation Source         | /usr/local/bin/fastlane                     |
| Host                        | Mac OS X 10.12.6 (16G29)                    |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/2.4.2/lib            |
| OpenSSL Version             | OpenSSL 1.0.2l  25 May 2017                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
[...]

Issue #9630, opened 2017-06-29
Retrieve existing IAP review screenshot


kcornbower:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

It appears that updating IAPs always requires re-uploading the original review screenshot anytime anything is updated. Is it possible to retrieve the existing IAP screenshot and reuse it? We have hundreds of existing IAP screenshots and no longer have those files. The ```review_screenshot``` property always appears to return ```nil```.

PR #10903, opened 2017-11-14
[Spaceship] Ability to use existing IAP screenshots

ecornbower:
### Checklist

### Motivation and Context

Fixes issues #9826, #9630 and possibly #10743.

The current code does not allow editing an iOS IAPs (in-app purchases) without uploading a new screenshot as well. This change uses the existing uploaded screenshots by default unless the user specifies that they wish to upload a new one.

This can be tested by retrieving an existing IAP and immediately re-saving it. The current code will fail with an ```OPERATION_FAILED``` response due to the missing screenshots and a bug with the ```contentHosting``` variable missing. The new code fixes this bug.

```
app = Spaceship::Application.find("XXX")
iap = app.in_app_purchases.all.first.edit
iap.save!
```

### Description

I've added a public accessor ```review_screenshot``` to allow users to retrieve the previously uploaded screenshot data. The new code now re-uses the retrieved values for the review screenshot by default, as well as reuses the existing ```contentHosting``` values which were previously missing when editing and causing saving errors.

### DIFF
--- a/spaceship/lib/spaceship/tunes/iap_detail.rb
+++ b/spaceship/lib/spaceship/tunes/iap_detail.rb
@@ -133,6 +133,12 @@ module Spaceship
         Tunes::IAPStatus.get_from_string(raw_data["versions"].first["status"])
       end
 
+      # @return (Hash) Hash containing existing review screenshot data
+      def review_screenshot
+        return nil unless raw_data && raw_data["versions"] && raw_data["versions"].first && raw_data["versions"].first["reviewScreenshot"] && raw_data['versions'].first["reviewScreenshot"]["value"]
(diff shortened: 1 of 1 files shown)


kcornbower:
@googlebot I signed it!

kcornbower:
Looks like some sort of environment error on CircleCI caused the build to fail, will need an admin to try and rebuild it.

```
Unexpected environment preparation error: error establishing an SSH session ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

taquitos:
Looks like circleci is having issues today

ohayon (approved):
📸

PR #10914, opened 2017-11-16
Allow device-specific data

bvirlet:
Before this, the Framefile.json could have defaults, and screenshot-specific overrides. Here we can add device-specific overrides.

### Checklist

### Motivation and Context

In some instances, you may want the screenshots to render a bit differently for difference device types.

I have reprocessed my screenshots with and without the changes. Also the change is falling back to the default, so only affects if it's explicitly added in the Framefile.json. This should eliminate any potential regressions.

### Description

A typical usage would be to be able to use `show_complete_frame` for the iPad.

To achieve this, I introduce a new filter: `device_filter`.

```
{
  "device_frame_version": "latest",
  "default": {
    "title": {
      "color": "#ffffff"
    },
    "background": "./Frameit-background.png",
    "padding": "3%",
    "show_complete_frame": false,
    "stack_title" : true,
    "title_below_image": false,
    "font_scale_factor": 0.05,
    "title_padding": "5%x15%"
  },
   "data": [
    {
      "filter": "scan"
    },
    {
      "filter": "edit"
    },
    {
      "filter": "multipage_pdf"
    },
    {
      "filter": "organize"
    },
    {
      "filter": "export"
    },
    {
      "device_filter" : "iPad",
      "show_complete_frame" : true,
      "font_scale_factor": 0.035
    }
  ]
}
```

### Visually, without and with the filter

This lets modify the iPad screenshot without modifying the iPhone's

**Before**
before

**After**
after

### DIFF
--- a/frameit/lib/frameit/config_parser.rb
+++ b/frameit/lib/frameit/config_parser.rb
@@ -23,12 +23,15 @@ module Frameit
 
     # Fetches the finished configuration for a given path. This will try to look for a specific value
     # and fallback to a default value if nothing was found
-    def fetch_value(path)
+    def fetch_value(path, device_name = nil)
+      device_specific = @data['data'].find { |a| a['device_filter'] && device_name.include?(a['device_filter']) } unless device_name.nil?
+
       specific = @data['data'].find { |a| path.include? a['filter'] }
 
       default = @data['default']
 
-      values = default.fastlane_deep_merge(specific || {})
+      values = default.fastlane_deep_merge(device_specific || {})
+      values = values.fastlane_deep_merge(specific || {})
 
       change_paths_to_absolutes!(values)
       validate_values(values)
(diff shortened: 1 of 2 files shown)


bvirlet:
Regarding documentation, not sure about the process to update since it's in a different repo.

bvirlet:
Closing as I have reopened a better PR under #11014

janpio:
> Regarding documentation, not sure about the process to update since it's in a different repo.

Action documentation is generated from the file in this repo, so changing it here will propagate over to /docs.

Other documentation can be changed directly in fastlane/docs. Just mention that it depends on your other PR, and we will make sure to no merge it before the actual PR is merged and deployed.

(Great PR by the way!)

bvirlet:
@janpio Thanks, I found this and updated the doc in my new PR: #11014

Issue #9574, opened 2017-06-23
Google Api Error: multiApkShadowedActiveApk: Version 4145 of this app can not be downloaded...


GeoffreyPlitt:
Our deploys were working fine for a long time and suddenly started failing with this error. I don't understand what's wrong, or how to fix.

```
fastlane supply --apk ./android/app/build/outputs/apk/app-release.apk --json_key ./creds_android/playstore_jsonkey.json --track beta --package_name com.stardust

+-------------------------+-------------------------------------------------+
|                         Summary for supply 2.41.0                         |
+-------------------------+-------------------------------------------------+
| apk                     | ./android/app/build/outputs/apk/app-release.apk |
| json_key                | ./creds_android/playstore_jsonkey.json          |
| track                   | beta                                            |
| package_name            | com.stardust                                    |
| rollout                 | 0.1                                             |
| skip_upload_apk         | false                                           |
| skip_upload_metadata    | false                                           |
| skip_upload_images      | false                                           |
| skip_upload_screenshots | false                                           |
| validate_only           | false                                           |
| check_superseded_tracks | false                                           |
+-------------------------+-------------------------------------------------+

[15:01:06]: Preparing apk at path './android/app/build/outputs/apk/app-release.apk' for upload...
[15:01:10]: Updating track 'beta'...
[15:01:10]: Uploading all changes to Google Play...
[15:01:13]: Sending crash report...
[15:01:13]: The stacktrace is sanitized so no personal information is sent.
[15:01:13]: To see what we are sending, look here: /Users/distiller/.fastlane/latest_crash.json
[15:01:13]: Learn more at https://github.com/fastlane/fastlane#crash-reporting
[15:01:13]: You can disable crash reporting by adding `opt_out_crash_reporting` at the top of your Fastfile

[!] Google Api Error: multiApkShadowedActiveApk: Version 4145 of this app can not be downloaded by any devices as they will all receive APKs with higher version codes.

fastlane supply --apk ./android/app/build/outputs/apk/app-release.apk --json_key ./creds_android/playstore_jsonkey.json --track beta --package_name com.stardust returned exit code 1

Action failed: fastlane supply --apk ./android/app/build/outputs/apk/app-release.apk --json_key ./creds_android/playstore_jsonkey.json --track beta --package_name com.stardust
```

GeoffreyPlitt:
Sorry, I thought the above was enough. Here it is:
```
distiller-build-59:~ distiller$ fastlane env
[...]

PR #10927, opened 2017-11-16
[Supply] Apk check_superseded_tracks improvements

capyvara:
### Checklist

### Motivation and Context
Improvements to Supply check_superseded_tracks, try to fix and help debugging some users has having, such as in #9574 

### Description
- Added some UI.verbosity messages
- Fixes a bug in which a superseded track may not be computed
- Added tests

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -192,6 +192,7 @@ module Supply
         end
 
         next if removed_version_codes.empty?
+
         keep_version_codes = track_version_codes - removed_version_codes
         max_tracks_version_code = keep_version_codes[0] unless keep_version_codes.empty?
         client.update_track(track, 1.0, keep_version_codes)
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -168,7 +168,7 @@ module Supply
     #  - Lesser than the greatest of any later (i.e. production) track
     #  - Or lesser than the currently being uploaded if it's in an earlier (i.e. alpha) track
     def check_superseded_tracks(apk_version_codes)
-      UI.message("Checking superseded tracks...")
+      UI.message("Checking superseded tracks, uploading '#{apk_version_codes}' to '#{Supply.config[:track]}'...")
       max_apk_version_code = apk_version_codes.max
       max_tracks_version_code = nil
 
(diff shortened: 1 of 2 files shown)


andreyslyusar:
When are you going to merge these changes ?

janpio:
> Fixes a bug in which a superseded track may not be computed
Could you explain that one a bit so a reviewer knows what to look for?

capyvara:
@janpio 
If you merge only the rspec tests the bug will happen (on the remove lesser than the currently being uploaded if it is in an earlier (i.e. alpha) track), the case was that the first found track wasn't considered a candidate for deprecation, fixed by changing a if-else to a if-end.

andreyslyusar:
Can somebody please make a review so we can merge it?

KrauseFx:
Hey everyone, sorry I don't have enough Android knowledge to do a full review of this PR. Maybe someone from @fastlane/android can help out here?

haxxpop:
I really want this one !!

supply/lib/supply/uploader.rb
```diff
@@ -168,34 +168,34 @@ def update_track(apk_version_codes)
...
       config_track_index = tracks.index(Supply.config[:track])
 
       tracks.each_index do |track_index|
-        next if track_index.eql? config_track_index
         track = tracks[track_index]
-
         track_version_codes = client.track_version_codes(track).sort
+        UI.verbose("Found '#{track_version_codes}' on track '#{track}'")
+
+        next if track_index.eql? config_track_index
```
janpio: This was moved down here so the verbose output would still happen?
What exactly does this check? (Maybe add as comment?)

janpio:
@andreyslyusar @haxxpop What do you think of these changes, do they make sense to you?

supply/lib/supply/uploader.rb
```diff
@@ -168,34 +168,34 @@ def update_track(apk_version_codes)
...
-            keep_version_codes = track_version_codes - removed_version_codes
-            max_tracks_version_code = keep_version_codes[0] unless keep_version_codes.empty?
-            client.update_track(track, 1.0, keep_version_codes)
-            UI.message("Superseded track '#{track}', removed '#{removed_version_codes}'")
-          end
+        removed_version_codes = track_version_codes.take_while do |v|
+          v < max_tracks_version_code || (v < max_apk_version_code && track_index > config_track_index)
         end
+
+        next if removed_version_codes.empty?
```
janpio: empty line after this for clarity that the next line is a new thing?

supply/lib/supply/uploader.rb
andreyslyusar: We need verbose output in order to get more info why it does failed, more description is in #9574.

supply/lib/supply/uploader.rb
capyvara: Not sure, rubocop doesn't complain, feel free to change since it doesn't modify the behavior.

janpio: done.

janpio:
(That message is expected and not a problem, we know that both people (@capyvara and me) have signed the CLA and so can just merge the changed when ready)

janpio (approved):
Oh, just now I figured out that you @capyvara wrote the initial implementation of this! This makes merging this change a lot easier of course, I expect you know what you were doing back then ;)

PR #10937, opened 2017-11-17
TestFlight made changes to how adding/removing testers works

taquitos:
- update to use new APIs

### DIFF
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -213,7 +213,7 @@ module Spaceship::TestFlight
         req.body = [{
           "email" => email,
           "firstName" => first_name,
-          "lastName" => first_name
+          "lastName" => last_name
         }].to_json
         req.headers['Content-Type'] = 'application/json'
       end
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -202,7 +202,7 @@ module Spaceship::TestFlight
       handle_response(response)
     end
 
-    def put_tester_to_group(app_id: nil, email: nil, first_name: nil, last_name: nil, group_id: nil)
+    def post_tester_to_group(app_id: nil, email: nil, first_name: nil, last_name: nil, group_id: nil)
       assert_required_params(__method__, binding)
 
       # Then we can add the tester to the group that allows the app to test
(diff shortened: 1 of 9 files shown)


spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -105,6 +105,44 @@ def add_group_to_build(app_id: nil, group_id: nil, build_id: nil)
...
+      return @cached_groups if @cached_groups
+
+      r = request(:get, "/testflight/v2/providers/#{self.provider.provider_id}/apps/#{app_id}/groups")
+      @cached_groups = parse_response(r, 'data')
+    end
+
+    #####################################################
+    # @!group Testers
+    #####################################################
+    def testers(tester)
```
ohayon: Not sure I understand what this is doing? Passing a tester to get the URL?

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -105,6 +105,44 @@ def add_group_to_build(app_id: nil, group_id: nil, build_id: nil)
...
+      url = tester.url[:index]
+      r = request(:get, url)
+      parse_response(r, 'data')['users']
+    end
+
+    def testers_by_app(tester, app_id, group_id: nil)
```
ohayon: Same comment as above, but also, should we have `app_id` be a named parameter?

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -164,18 +202,19 @@ def create_app_level_tester(app_id: nil, first_name: nil, last_name: nil, email:
       handle_response(response)
     end
 
-    def put_tester_to_group(app_id: nil, tester_id: nil, group_id: nil)
+    def put_tester_to_group(app_id: nil, tester: nil, group_id: nil)
```
ohayon: I think there was some convention around not passing in model objects into the client. But I am totally cool with this happening.

ohayon (approved):
Couple tiny things that definitely dont need to be changed. Nice work!

spaceship/lib/spaceship/test_flight/client.rb
taquitos: yeah, it's copypasta from `Tunes` into `TestFlight` since `Tunes` should be deprecated/removed.

spaceship/lib/spaceship/test_flight/client.rb
taquitos: same as above, copypasta

spaceship/lib/spaceship/test_flight/client.rb
taquitos: Yeah, I think that makes sense.

PR #10938, opened 2017-11-17
[gradle] example how to sign and zipalign app

janpio:
I looked for a way to sign and zipalign an app using Fastlane. Copypaste ready actions exist, but I hoped for a way to do this with `gradle`. And after some googling I indeed found a way to do that by supplying some properties to the gradle build process.

This PR:
- adds a code example how to sign and zipalign the app with `gradle`
- replaces wrong " with correct (and escaped) \'


taquitos (approved):
🎈:goat:

PR #10939, opened 2017-11-17
Fastlane core action rename & aliases

mpirri:
**Note for reviewers!** `git` handles file renaming followed by file creation at the old file name (e.g., rename `gym.rb` to `build_app.rb` and then create a new file at `gym.rb`) badly. As such, this will be a difficult PR to review by looking at the overall PR diff.  Instead, it is recommended that you review by looking at each of the sub-commits.

### Checklist

### Motivation and Context

New users to fastlane are often perplexed by the core action names in Fastfiles (e.g., gym, cert, sigh). This PR renames core actions to use a new verb/noun-style style. Old action names become aliases pointing to the newly renamed actions. Over time we will change over documentation to use these new aliases for the core actions.

### Description

This PR consists of four steps (each a commit):
- Original core actions are renamed to the new noun/verb style
- Old aliases (`testflight` and `appstore`) are recreated as simple subclasses of core actions
- Update the unit tests to omit the option check from these new aliases and also to allow multi-level subclassing within actions.
- New aliases are created with the old core action names

### Renamed core actions

| old name | new name | description |
| ---- | ---- | ---- |
| cert |  get_certificates | same as `fastlane cert create` |
| deliver | upload_to_app_store | same as `fastlane deliver run` |
| frameit | frame_screenshots | same as `fastlane frameit run` |
| gym | build_ios_app, build_app | same as `fastlane gym build` |
| match | sync_code_signing | same as `fastlane match run` |
| pem |  get_push_certificate | same as `fastlane pem renew` |
| pilot | upload_to_testflight | same as `fastlane pilot upload` |
| precheck | check_app_store_metadata | same as `fastlane precheck` |
| produce |  create_app_online | same as `fastlane produce create` |
| scan | run_tests | same as `fastlane scan` |
| screengrab | capture_android_screenshots | same as `fastlane screengrab run` |
| sigh |  get_provisioning_profile | same as `fastlane sigh renew` |
| snapshot | capture_ios_screenshots, capture_screenshots | same as `fastlane snapshot run` |
| supply | upload_to_play_store | same as `fastlane play_store run` |

A new alias `build_android_app` is being created for the `gradle` action as well.

[*] `build_app` and `capture_screenshots` current are aliased to their iOS equivalents (`build_ios_app` and `capture_ios_screenshots`), but we will eventually make these context-sensitive actions that will invoke the correct action for the current platform being targeted.


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lib/fastlane/actions/build_app.rb
```diff
@@ -5,7 +5,7 @@ module SharedValues
       DSYM_OUTPUT_PATH = :DSYM_OUTPUT_PATH
     end
 
-    class GymAction < Action
+    class BuildAppAction < Action
```
alexsinger: How does renaming action files impact the doc generation? If somebody searches for `gym` will gym docs be available? We say "via gym", "via snapshot" - will users be able to understand what these are?

mpirri: Excellent question - looking into this now...

fastlane/lib/fastlane/actions/synchronize_provisioning.rb
```diff
@@ -4,7 +4,7 @@ module SharedValues
       MATCH_PROVISIONING_PROFILE_MAPPING = :MATCH_PROVISIONING_PROFILE_MAPPING
     end
 
-    class MatchAction < Action
+    class SynchronizeProvisioningAction < Action
```
alexsinger: Wondering if `SynchronizeCodeSigning` is more appropriate?

mpirri: @getaaron also had a great suggestion - using the word `sync` rather than `synchronize` because it is easier to spell. Thoughts?

fastlane/lib/fastlane/actions/build_app.rb
alexsinger: Seems like we might be OK because of the changes in commit `d0b5408` ?

getaaron (approved):
Two suggested pre-merge changes:

- [ ] Change the verb in `renew_push_certificate` to `create` to indicate it can be used for first-time creation (is `create_push_certificate_if_necessary` too long?)
- [ ] Rename `synchronize_provisioning` to `sync_provisioning` (spelling is hard)

Overall, the changes look great and I'm OK with merging even if you don't make these changes

fastlane/lib/fastlane/actions/synchronize_provisioning.rb
alexsinger: `sync` sounds good

mpirri:
@getaaron Agreed about `renew` not being the right word here - it feels like `cert`, `sigh`, and `pem` all have a similar verb problem (and all use different verbs!). "create", "renew", "generate" are all too over-specific (and sometimes inaccurate) about what they do.

What about if we used something like `get`?  (`get_provisoining_profiles`, `get_certificates`, `get_push_certificates`) It is a more generic word that implies that it will make sure the thing is there, but doesn't specify how it does it. Not sure I love it, though - it doesn't really convey the idea that nothing happens if the item is already present.

With regards `synchronize_provisioning`,  @alexsinger had recommended we use `code_signing` instead of `provisioning` and I agree.  I'm a bit worried about using `sync` INSTEAD OF `synchronize` just because none of the other aliases use truncated words. What if we had both?  Too confusing?  (`sync_code_signing` as an alias for `synchronize_code_signing`)

jetersen (approved):
This is looking awesome 👍

milch:
@mpirri I think `get` would indeed work best here. Logically what it does is an [`upsert`](https://wiki.postgresql.org/wiki/UPSERT) but I doubt a lot of people are familiar with this kind of terminology 😄 

One more issue is that `produce` creates apps on both iTunes Connect and the Developer Portal so `create_itunes_connect_app` is not quite accurate. It would feel strange to call `create_itunes_connect_app(skip_itc: true)` to actually create an app on the Developer Portal... So maybe `create_app_online` or something?

mpirri:
I've updated the PR's description with the new names.

Specifically:

* `create_app_online` as suggested by @milch (I wish there was a more elegant way of phrasing this, but, as outlined, it isn't easy)
* `get_certificates`, `get_push_certificate`, `get_provisioning_profiles` (as per discussion with @getaaron and @alexsinger)
* BOTH `sync_code_signing` and `sychronize_code_signing` (as per discussion with @getaaron and @alexsinger)

Code changes coming soon...

Thoughts?

fastlane/lib/fastlane/actions/build_app.rb
```diff
@@ -0,0 +1,142 @@
...
+
+        Actions.lane_context[SharedValues::DSYM_OUTPUT_PATH] = absolute_dsym_path if File.exist?(absolute_dsym_path)
+        Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE] = Gym::BuildCommandGenerator.archive_path
+        ENV[SharedValues::DSYM_OUTPUT_PATH.to_s] = absolute_dsym_path if File.exist?(absolute_dsym_path)
+
+        return absolute_ipa_path
+      end
+
+      def self.description
+        "Easily build and sign your app (via gym)"
```
janpio: What would the Android equivalent be here?

mpirri: Hmm, good question - perhaps this maps to the `gradle` action?  In which case, we should probably have `build_ios_app` and `build_android_app` (a la screenshots), and we will introduce a `build_app` action that maps to the correct behavior based on platform?

KrauseFx:
This looks good! We can update the docs on `fastlane/docs` after this is merged and shipped, as this is fully backwards compatible, right?

fastlane/lib/fastlane/actions/build_app.rb
casz: This sounds AWESOME 👍

mpirri: I'm a little bit wary of having `build_android_app` do the same thing as `gradle`, since `gradle` does more than just build apps. What do you think, @janpio ?

janpio: I actually didn't know that - until now I thought `gradle` was also only for building.
Is `gym` limited to absolutely only building?

mpirri:
@KrauseFx Yeah, the next step will be to update the docs to reference these new actions. Everything should be backwards compatible, so no expected breakages for existing Fastfiles.

mpirri:
One more change added:
`build_app` is now an alias to `build_ios_app`.
`build_android_app` is now an alias to `gradle` (as per discussion with @janpio)
`capture_screenshots` is now an alias to `capture_ios_screenshots`.

The intention is to replace `build_app` and `capture_screenshots` with (coming soon!) context-sensitive actions that will properly invoke the `ios` or `android` version of the action when invoked.

fastlane/lib/fastlane/actions/create_itunes_connect_app.rb
```diff
@@ -4,7 +4,7 @@ module SharedValues
       PRODUCE_APPLE_ID = :PRODUCE_APPLE_ID
     end
 
-    class CreateItunesConnectAppAction < Action
+    class CreateAppOnlineAction < Action
```
ohayon: `Online` here means iTunesConnect right? Is having this be `CreateAppOnITunesConnect` too verbose?

mpirri: Check out @milch 's comments from about a week ago - his point was that `produce` creates both iTC and DeveloperPortal apps. I don't love this "online" naming style, but it is a sticky problem.

ohayon:
This all looks awesome @mpirri !! 

I dont want to add another cook to the kitchen or derail anything, but I wonder if we could do something with the `build_app` commands that instead of having to have two separate ones, we could do something like rake where `build_app` would default to ios, and you can specify by doing `build_app:android` or something similar. I know commander (at least used to be able to) can handle syntax like: `command :'build_app:android' do...` or something to that effect.

Definitely dont want to add any complexity here though! 🚀

fastlane/lib/fastlane/actions/create_itunes_connect_app.rb
milch: Yeah it's not a good name, I'd prefer something more verbose if it's more expressive. `create_app_on_itunes_connect_or_apple_developer_portal` might be a bit *too* verbose, though 😄

mpirri:
@ohayon I do think that we will do something like this very soon with just having the `build_app` and `capture_screenshots` commands call the appropriate platform code. These commands are just placeholders for now.

The current thought is to look at the platform block from which the action is called and call the correct platform-method based on that. If the platform block is not present, we might require the user to specify a `:android` or `:ios` argument or else use the `build_XXX_app` action.

milch:
@ohayon I wouldn't want to do something like `build_app:android` since it'd conflict with param syntax on the command line: 

```
fastlane beta version:1.0
```

fastlane/spec/unused_options_spec.rb
```diff
@@ -1,7 +1,7 @@
...
+      let (:all_exceptions) { %w(pilot appstore cert deliver gym match pem produce scan sigh snapshot precheck supply testflight mailgun testfairy ipa import_from_git hockey deploygate crashlytics artifactory appledoc slather screengrab download_dsyms notification frameit set_changelog register_device register_devices latest_testflight_build_number app_store_build_number sh swiftlint plugin_scores google_play_track_version_codes modify_services build_app build_android_app build_ios_app capture_screenshots capture_android_screenshots capture_ios_screenshots check_app_store_metadata get_certificates create_app_online frame_screenshots get_provisioning_profile get_push_certificate run_tests submit_build_to_app_store sync_code_signing upload_to_app_store upload_to_play_store upload_to_testflight) } # rubocop:disable Metrics/LineLength
```
casz: Anyone against making this multiline? This is getting very hard to read 😭

mpirri: @casz Done! Much better!

getaaron:
This looks incredible and I request merging it quickly so I can rewrite my fastfile asap 🤠

fastlane/lib/fastlane/actions/create_itunes_connect_app.rb
mpirri: Fortunately, creating these aliases (and keeping old ones around) is pretty easy & cheap, so we can always create a new alias later (once we have become smarter and can think of a better name 🤣 ).

ohayon: got it. this makes sense, sorry for missing that before!

ohayon (approved):
🐇 🎈

@taquitos am i doing that right? 

🎉 🚀

casz:
💣 💥 all the PRs

KrauseFx:
Follow-up PR https://github.com/fastlane/fastlane/pull/11505

PR #10946, opened 2017-11-18
Allow shell conversion when type is :shell_string

jdee:
### Checklist

### Motivation and Context

There is an attr_accessor called `:allow_shell_conversion` in `FastlaneCore::ConfigItem`. When I was writing up recent docs, I documented it but then discovered it didn't work, at least not as I expected. (So I deleted it from the docs.)  And in particular, the `sh` action does not allow array arguments. I expected to be able to do this:

```Ruby
sh ["git", "commit", "-m", message, *changes]
```

Which would result in all arguments being shell-escaped. The fact that this doesn't work is quite surprising. This seems like the action this processing was made for.

These changes make both those things work, though there are other ways to make them work as well, and perhaps some of these choices are deliberate.

### Description

The reason the `allow_shell_conversion` option doesn't seem to work is because:

- It cannot be passed to the initializer.
- The initializer sets it to true if `type` is `:shell_string`.
- The `allow_shell_conversion` option is only consulted in `#auto_convert_value` if `type` is `String`. But then, of course it will be false because `type != :shell_string`.

It is actually possible to enable it currently this way:

```Ruby
config_item = FastlaneCore::ConfigItem.new(
  key: :shell_command,
  type: String
)
config_item.allow_shell_conversion = true
```

This is because it is an `attr_accessor`. But this is a little inconvenient when declaring an array of literals.

I like the idea of using a symbol for `type`. This would also be a nicer way to support Boolean options than using `is_string: false`, e.g. `type: :boolean`.

But there are other ways of doing this, such as passing `allow_shell_conversion` directly to the initializer.

Also updated `Actions.sh` to split and shell-escape arrays so you can now run the command in the description as indicated from anywhere this functionality is supported. This includes the `sh` "action" (not really any action, just a method in `FastFile`) and the `Actions.sh` method, which is available within all actions by just calling `sh`. (This is preferable to calling `other_action.sh`, which entails a directory change.)

To make this more testable, also updated the implementation of `Actions.sh` to use `Open3.popen2e`, which doesn't rely on the global variable `$?` to determine the process exit status. This makes it easy to mock the return value in tests without actually executing any subshell in tests.

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -175,7 +175,7 @@ module Fastlane
 
     # Execute shell command
     def sh(command, log: true, error_callback: nil)
-      command = command.map(&:to_s).map(&:shellescape).join(" ") if command.kind_of?(Array)
+      command = Shellwords.join(command) if command.kind_of?(Array)
       command_header = log ? command : "shell command"
       Actions.execute_action(command_header) do
         Actions.sh_no_action(command, log: log, error_callback: error_callback)
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -175,6 +175,7 @@ module Fastlane
 
     # Execute shell command
     def sh(command, log: true, error_callback: nil)
+      command = command.map(&:to_s).map(&:shellescape).join(" ") if command.kind_of?(Array)
       command_header = log ? command : "shell command"
       Actions.execute_action(command_header) do
         Actions.sh_no_action(command, log: log, error_callback: error_callback)
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -39,7 +39,6 @@ module Fastlane
             UI.command_output(line.strip) if print_command_output
             result << line
           end
-          io.close
           exit_status = thread.value.exitstatus
         end
 
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -25,7 +25,7 @@ module Fastlane
       Encoding.default_external = Encoding::UTF_8
       Encoding.default_internal = Encoding::UTF_8
 
-      command = command.join(' ') if command.kind_of?(Array) # since it's an array of one element when running from the Fastfile
+      command = command.map(&:to_s).map(&:shellescape).join(' ') if command.kind_of?(Array) # since it's an array of one element when running from the Fastfile
       UI.command(command) if print_command
 
       result = ''
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -27,7 +27,7 @@ module Fastlane
       Encoding.default_external = Encoding::UTF_8
       Encoding.default_internal = Encoding::UTF_8
 
-      command = command.map(&:to_s).map(&:shellescape).join(' ') if command.kind_of?(Array) # since it's an array of one element when running from the Fastfile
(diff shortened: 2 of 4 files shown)


fastlane/lib/fastlane/fast_file.rb
```diff
@@ -175,6 +175,7 @@ def actions_path(path)
 
     # Execute shell command
     def sh(command, log: true, error_callback: nil)
+      command = command.map(&:to_s).map(&:shellescape).join(" ") if command.kind_of?(Array)
```
jdee: This repeats what is in the `Actions.sh` method, but it makes the `command_header` come out right.

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -25,20 +27,20 @@ def self.sh_control_output(command, print_command: true, print_command_output: t
...
 
-      command = command.join(' ') if command.kind_of?(Array) # since it's an array of one element when running from the Fastfile
+      command = command.map(&:to_s).map(&:shellescape).join(' ') if command.kind_of?(Array) # since it's an array of one element when running from the Fastfile
       UI.command(command) if print_command
 
       result = ''
       if Helper.sh_enabled?
         exit_status = nil
-        IO.popen(command, err: [:child, :out]) do |io|
+        Open3.popen2e(command) do |stdin, io, thread|
```
jdee: The reason for using `Open3` is to avoid the read-only global `$?`, which makes testing hard. Otherwise, there is no change here of any substance. With `IO.popen`, stdin and stdout (and I think stderr) are mapped to the single `io` in the block. With `popen2e`, `io` is the merged stdout and stderr. Standard input is mapped to a separate IO. In neither case is it used. The behavior of `sh` is unchanged with regard to input.

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -25,20 +27,20 @@ def self.sh_control_output(command, print_command: true, print_command_output: t
...
           io.sync = true
           io.each do |line|
             UI.command_output(line.strip) if print_command_output
             result << line
           end
           io.close
```
milch: Do you have to close stdout explicitly here? The docs for `popen2e` seem to suggest that closing is not necessary in the block form

fastlane/spec/helper/sh_helper_spec.rb
```diff
@@ -1,14 +1,23 @@
 describe Fastlane::Actions do
   describe "#sh" do
+    let (:mock_input) { double :input }
+    let (:mock_output) { double :output }
+    let (:mock_status) { double :status }
+    let (:mock_thread) { double :thread, value: mock_status }
+
     before do
       allow(FastlaneCore::Helper).to receive(:sh_enabled?).and_return(true)
+      allow(mock_output).to receive(:sync=).with(true)
```
milch: Is there a way to just ignore these? Personally I'm not a huge fan of tests that test the internal implementation instead of the expected behaviour. The important thing is that it does shell out and the command is correct, whether it calls `each` on the standard output should be irrelevant

fastlane/lib/fastlane/fast_file.rb
milch: Can you refactor this so that the conversion is a separate method? As far as I can see the same logic is repeated here, once again at `fastlane/lib/fastlane/helper/sh_helper.rb:30` and one more time at `fastlane_core/lib/fastlane_core/configuration/config_item.rb:161`

milch (changes requested):
Looking good, I'd just like to see the tests changed a little bit and some light refactoring

jdee:
@milch Thanks for the review. I've made changes along the lines you requested.

In principle I agree regarding tests, but it can be hard to test certain sorts of code without any knowledge of the implementation. Still, it's best to avoid irrelevant conditions in tests just to make them pass, and just opening an empty fixture file was an improvement. There are lots of tests for Fastlane that actually call certain commands and should probably be mocked out.

milch (approved):

milch:
Thanks for the changes, looks good to me now!

Issue #10952, opened 2017-11-20
fastlane init attempts to create an app for framework in workspaces


GRiMe2D:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

#### Project Structure
```
xcodeworkspace
├ moneyapp-ios (App)
├ privateframework-ios (Dynamic Framework)
└ ReactiveObjC-iOS (Dynamic framework)
```

`fastlane init` tries to register apps for frameworks in iTunesConnect and Apple Developer. Unable to skip. 

##### Complete output when running fastlane, including the stack trace and command used
```[16:23:47]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[16:23:50]: Detected iOS/Mac project in current directory...
[16:23:50]: This setup will help you get up and running in no time.
[16:23:50]: fastlane will check what tools you're already using and set up
[16:23:50]: the tool automatically for you. Have fun! 
[16:23:50]: Created new folder './fastlane'.
[16:23:50]: $ xcodebuild -list -workspace ./moneyapp.xcworkspace
[16:23:52]: $ xcodebuild -showBuildSettings -workspace ./moneyapp.xcworkspace -scheme ReactiveObjC-iOS
[16:23:54]: Your Apple ID (e.g. fastlane@krausefx.com):
```

janpio:
Could you please update your issue with output that shows the problem you are reporting?

GRiMe2D:
```
$ fastlane init
[17:39:42]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[17:39:45]: Detected iOS/Mac project in current directory...
[17:39:45]: This setup will help you get up and running in no time.
[17:39:45]: fastlane will check what tools you're already using and set up
[17:39:45]: the tool automatically for you. Have fun! 
[17:39:45]: $ xcodebuild -list -workspace ./moneyapp.xcworkspace
[17:39:47]: $ xcodebuild -showBuildSettings -workspace ./moneyapp.xcworkspace -scheme ReactiveObjC-iOS
[...]

PR #10955, opened 2017-11-20
Allow the user to select their scheme during fastlane init

milch:
See #10952. This will 

* Try to automatically select a scheme that matches the project name instead of the first one in the list, ask for the scheme otherwise
* Warn the user if there is more than one scheme that the selected one might be wrong

### DIFF
--- a/fastlane/lib/fastlane/setup/setup_ios.rb
+++ b/fastlane/lib/fastlane/setup/setup_ios.rb
@@ -34,6 +34,9 @@ module Fastlane
           detect_if_app_is_available
         end
         print_config_table
+        if self.project.schemes.count > 1
+          UI.important("Note: If the values above are incorrect, it is possible the wrong scheme was selected")
+        end
         if UI.confirm("Please confirm the above values")
           default_setup
         else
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/setup/setup_ios.rb
```diff
@@ -34,6 +34,8 @@ def run(user: nil)
           detect_if_app_is_available
         end
         print_config_table
+        UI.important("Note: If the values above are not correct, it is possible the wrong scheme was selected by fastlane")
+        UI.important("As a workaround, you can drag your scheme to the top of the scheme editor in Xcode (Product > Scheme > Manage Schemes...)")
```
janpio: Maybe this edge case should "only" be documented in the docs and these two lines be replaced by URL? This is shown to _all_ `fastlane init` users.

milch: Huh, it seems something was messed up there. I previously did a force push since I thought of the same thing, so now it only shows it to people with multiple schemes

milch: I re-pushed, now it shows the correct code 👍🏼

KrauseFx (approved):
Awesome, nice catch, and great PR, thanks 👍

Issue #10960, opened 2017-11-20
undefined method `[]' for nil:NilClass during Match


Fluzzarn:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [X] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [X] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [X] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Over the weekend, interacting with the dev center using match errors out. After drilling down, it looks like the providers for the supplied developer account are nil/null and the terms aren't accepted. I vaguely remember there was a warning in fastlane that new terms were needed to be accepted on the Developer Center. However, on the Dev Center there's no new terms update, and I can do all normal functions on the dev center with this account (regen/create provisioning profiles, add/remove team members, make new bundle ID's).

The JSON Blob printed out in the output is the body of the request to https://olympus.itunes.apple.com/v1/session made in fetch_olympus_session

(please excuse the mess of output, it's taken from Bamboo who maims fastlane's output)

##### Complete output when running fastlane, including the stack trace and command used
INFO [2017-11-20 14:26:15.65]: [32m-------------------[0m
INFO [2017-11-20 14:26:15.65]: [32m--- Step: match ---[0m
INFO [2017-11-20 14:26:15.65]: [32m-------------------[0m
INFO [2017-11-20 14:26:15.65]: [32mSuccessfully loaded '/Users/buildserver/Documents/bamboo-home/bamboo_agent_5.11.1.1/xml-data/build-dir/HHUS-CIQE315-JOB1/ios/Fastlane/Matchfile' ?[0m

+---------+----------------------------------------------------------------------------------------+
|                           Detected Values from './Fastlane/Matchfile'                            |
+---------+----------------------------------------------------------------------------------------+
| git_url | [[redacted]]                                                                           |
| type    | enterprise                                                                             |
+---------+----------------------------------------------------------------------------------------+

+-----------------------+----------------------------------------------------------------------------------------+
|                                            [32mSummary for match 2.65.0[0m                                            |
+-----------------------+----------------------------------------------------------------------------------------+
| type                  | enterprise                                                                             |
| force                 | true                                                                                   |
| git_url               | [[redacted]]                                                                           |
| git_branch            | master                                                                                 |
| template_name         | [[Redacted]]                                                                           |
[...]

PR #10972, opened 2017-11-21
Hotfix for TestFlight related crashes

milch:
See #10960. For some people the provider accessed in `fetch_olympus_session` is nil. 

Predictably, this kills the fastlane run. I've rummaged around on iTunes Connect and looked at the requests/responses and it seems that `provider_id` is always equal to the `team_id` for me. As a result this hotfix returns the `provider_id` if it is available and the `team_id` otherwise, and ignores the provider during initialisation of the client if it is not available.


taquitos (approved):
🎈:goat:

PR #10987, opened 2017-11-21
Update limits for external testers

maxkomarychev:
references: https://developer.apple.com/news/?id=07312017a
https://developer.apple.com/testflight/

Update docs according to recent changes in testflight.


mpirri (approved):
Good catch!

PR #10993, opened 2017-11-22
[copy_artifacts] additional code example

janpio:
adds an example how to use a shared value of `gym` to automatically get the path of the created .ipa file and copy it (and also "teaches" a user how to dynamically fill the array used to define the patterns)


milch (approved):
🚀

PR #10994, opened 2017-11-22
Separate concurrent simulators by OS version

milch:
See: #10969 

If Xcode tries to concurrently run multiple simulators with different versions it fails running some of them. This PR separates the devices by OS version so that when the devices are launched, only devices of the same version are launched together.


mpirri (approved):
This is awesome - much more elegant fix than I was expecting!  😍

PR #10999, opened 2017-11-22
Fix incorrect auto complete options in zsh

revolter:
Before this, the "update_fastlane" option was being appended to
the last found option (lane name). So, having a lane named "test"
and typing "fastlane tes" and pressing the "tab" key auto
completed it to "fastlane test\ update_fastlane", where the space
character is escaped by a backslash.

The safest way to append a newline and the custom option that
should work in every zsh version is by adding a literal newline
in the string as described here:
http://zsh.sourceforge.net/FAQ/zshfaq03.html#313

### Checklist


KrauseFx (approved):
Beautiful ❤️

PR #11001, opened 2017-11-22
Check if the bundle id is already populated in the list of applications

paramv17:
### Checklist
- :white_check_mark: I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- :white_check_mark: I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- :white_check_mark: I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- :white_check_mark: I've updated the documentation if necessary.

### Motivation and Context
Creating the application  using Tunes returns failure sometimes - and later on the applications are magically populated. I believe this issue happens when the appId is not available in the itunesconnect and we try to use it to create the app. at any rate, this is a nice check to have.

### Description
Adding a check to see available bundle ids in TunesClient create_application - and throwing an exception if the available bundle ids don't contain the bundle id requested. Updating the tests to run with this.

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -293,7 +293,7 @@ module Spaceship
 
       # check if the bundle id exists in the list of bundle ids
       available_bundle_ids = data['bundleIds']
-      raise 'requested bundle id not available' unless available_bundle_ids.keys.include? bundle_id
+      raise 'requested bundle id not available' unless available_bundle_ids.keys.include?(bundle_id)
 
       # Now fill in the values we have
       # some values are nil, that's why there is a hash
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -291,6 +291,10 @@ module Spaceship
       r = request(:get, "ra/apps/create/v2/?platformString=#{platform}")
       data = parse_response(r, 'data')
 
+      # check if the bundle id exists in the list of bundle ids
+      available_bundle_ids = data['bundleIds']
+      raise 'requested bundle id not available' unless available_bundle_ids.keys.include? bundle_id
+
       # Now fill in the values we have
       # some values are nil, that's why there is a hash
       data['name'] = { value: name }
--- a/spaceship/spec/tunes/application_spec.rb
+++ b/spaceship/spec/tunes/application_spec.rb
@@ -79,7 +79,7 @@ describe Spaceship::Application do
           Spaceship::Tunes::Application.create!(name: "My Name",
                                                 sku: "SKU123",
                                                 bundle_id: "net.sunapps.100")
-        end.to raise_error "You must choose a primary language. You must choose a primary language."
+        end.to raise_error("You must choose a primary language. You must choose a primary language.")
       end
 
       it "raises an error if bundle is wildcard and bundle_id_suffix has not specified" do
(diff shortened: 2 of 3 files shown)


pav1790:
I can clean this up a bit. I guess it's a bit sloppy. I d be more than happy to discuss my motivations with making this change.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -274,6 +274,10 @@ def create_application!(name: nil, primary_language: nil, version: nil, sku: nil
       r = request(:get, "ra/apps/create/v2/?platformString=#{platform}")
       data = parse_response(r, 'data')
 
+      # check if the bundle id exists in the list of bundle ids
+      available_bundle_ids = data['bundleIds']
+      raise 'requested bundle id not available' unless available_bundle_ids.keys.include? bundle_id
```
KrauseFx: Please add `( )` around `bundle_id` for method calls. Also, can we change the error message to `Bundle identifier provided is already taken and not available any more. Please choose a different one`

KrauseFx (commented):
Awesome, nice change, thanks. Does this affect how _produce_ works? I think in _produce_ we're already checking if the app identifier is taken, so it should be fine, right?

pav1790:
@KrauseFx I checked `produce` - but I do not think this is a check that produce conducts. As of now, fastlane has no way of doing it. 

Here's a problem I am trying to fix : 
1. We create app entry in developer portal
2. This entry is not enumerated in ITC yet. 
3. We try to create an app with that bundle id 

Here I see that the `create!` fails but the application appears in the ITC a while later. 

So, I am trying to avoid this race condition - so that we can achieve determinism when creating an application. Since creating several applications is a serious usage of apple's resources, I am hesitant to conduct extensive experiments around it - or I'd try to fix the root cause assuming it were possible. 

Alternatively, I could just create a method in `TunesClient` to return a list of available bundle ids. That would clean up this mess a lot - and we can use that in `produce` and people using spaceship can use that separately.

spaceship/spec/tunes/application_spec.rb
```diff
@@ -139,30 +149,30 @@
         mock_client_response(:get_builds_for_train, with: hash_including(train_version: '1.0')) do
           [
             {
-              id: 1,
-              appAdamId: 10,
-              trainVersion: '1.0',
-              uploadDate: '2017-01-01T12:00:00.000+0000',
-              externalState: 'testflight.build.state.export.compliance.missing'
+                id: 1,
```
janpio: why these formatting changes?

janpio (changes requested):
Please see the comment and suggested changes @pav1790. Thanks.

janpio:
Oh and resolve the conflicts please (several added `(...)`) and rebase on current master please so we get current CI setup. Thanks again.

joshdholtz:
@pav1790 Hey! Are you still interested in getting this merged in? If so, would you be able to address these concerns? 😊

pav1790:
@KrauseFx yeah, sorry i kinda got busy with other things. Let me address these changes.

spaceship/spec/tunes/application_spec.rb
pav1790: rubocop induced is my guess. I ran rubocop while uploading a patch and it was still kept as is. I can reverse it, if you like @janpio

joshdholtz (approved):
Thanks for making the requested changes! This looks good 😊

joshdholtz:
@janpio Can you give this another review when you get a minute? 😊

janpio:
Seems I didn't see that notification.

Why did you close @heartofagoof ? Not relevant any more or just fed up with waiting?

heartofagoof:
@janpio it's alright. i created a different PR that had less disruptive changes - which helps me achieve the same thing.

PR #11002, opened 2017-11-22
[WIP] Enable simple, single word params that default to `true`

janpio:
### Checklist

### Motivation and Context

Right now the only way to supply parameters when executing lanes is in the `key:value` format. It would be neat to also be able to just call it with a single word that gets interpreted as a `true` bool variable.

### Description

Right now all params with `:` are by default used as params, those without for the platform selection and lane name. This PR just takes the leftover entries from that array and adds them to the params array with value `true`.

### DIFF
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -26,12 +26,12 @@ module Fastlane
         lane = platform_lane_info.delete_at(0)
       end
 
-      # convert remaining platform_lane_info values into parameters
+      # convert remaining platform_lane_info values into `:true` parameters
       platform_lane_info.each do |key|
-        UI.verbose("Using #{key}: true (!)")
+        UI.verbose("Using (!) #{key}: true")
         lane_parameters[key.to_sym] = true
       end
-      
+
       dot_env = Helper.is_test? ? nil : options.env
       Fastlane::LaneManager.cruise_lane(platform, lane, lane_parameters, dot_env)
     end
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -19,13 +19,19 @@ module Fastlane
       end
 
       platform = nil
-      lane = platform_lane_info[1]
+      lane = platform_lane_info.delete_at(1)
       if lane
-        platform = platform_lane_info[0]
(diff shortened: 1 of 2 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lib/fastlane/command_line_handler.rb
```diff
@@ -19,11 +19,17 @@ def self.handle(args, options)
       end
 
       platform = nil
-      lane = platform_lane_info[1]
```
milch: This code block is a bit complicated, maybe we could simplify it to something like 

```ruby
 platform, lane = if platform_lane_info.count > 1
                    platform_lane_info.shift(2)
                  else
                    [nil, platform_lane_info.shift]
                  end
```

Maybe it's just me but I was honestly kind of confused with the ordering of the`delete_at`s and that `nil`-dodging `if`

janpio: Probably because it evolved out of what was there before.

janpio: Yes indeed, this causes it to try to execute a `foo bool` lane, where `foo` is the platform and `bool` the lane name :(

janpio (commented):
Current state doesn't work with `fastlane foo bool`, calling a lane without specifying a platform, as this causes the lane name to be interpreted as the platform and the parameter as the lane name.

janpio:
[`Fastlane::SupportedPlatforms.all` exists](https://github.com/fastlane/fastlane/blob/65a0e08151f9c97ceae8b360018a8b8a15e8b376/fastlane/lib/fastlane/supported_platforms.rb) and contains a list of supported platforms.

Unfortunately [add_extra_platforms](https://docs.fastlane.tools/actions/add_extra_platforms/) also exists, which can be used in the `Fastfile` to modify the list of platforms. But the command line handler gets executed before that...

Only solution I can think of right now:
Use the default platform list in the command line handler when parsing the command. If `add_extra_platforms` is used later, parse the command again with this new knowledge to fix the platform/lane/platform settings.

Better ideas?

janpio:
I'll close this for now as it is not really _that_ important.

PR #11004, opened 2017-11-22
Extended tab auto complete to work with bundler and custom commands

revolter:
### Checklist

### Motivation and Context

This enabled tab auto complete for `bundle exec fastlane` and allows passing custom commands like `fl` to enable tab auto complete for `./fl`.

Works with:

    bundle exec fastlane enable_auto_complete
    bundle exec fastlane enable_auto_complete --custom first
    bundle exec fastlane enable_auto_complete --custom first,second

and:

    sh('bundle exec fastlane enable_auto_complete --custom first')

in a lane in the `Fastfile`.


fastlane/lib/fastlane/commands_generator.rb
```diff
@@ -218,10 +218,11 @@ def run
       command :enable_auto_complete do |c|
         c.syntax = 'fastlane enable_auto_complete'
         c.description = 'Enable tab auto completion'
+        c.option '-c STRING[,STRING2]', '--custom STRING[,STRING2]', String, 'Add custom command(s) to use for auto complete'
```
milch: Could you expand a bit on this comment? From this description it sounds a bit unclear to me what this does

milch (commented):
I like it, although I do think the possibility for it completing `bundle lane_name` could be a bit of an issue. Do we think that only advanced users will install shell completion anyway, so these users would not be confused by this? It looks like the auto completion for bundle is also added when calling `fastlane enable_auto_complete` without the `bundle exec`. Maybe we could work around any confusion by only adding the completion if it is called with `bundle exec`?

fastlane/lib/fastlane/commands_generator.rb
revolter: > Add custom command(s) for which tab auto complete should work too

revolter:
Indeed, installing completion generally means that you are accustomed to the terminal, and if you need to use the `bundle` command, chances that you know how it works are pretty high. In the worst case scenarios, `bundler` would print `Could not find command "deploy".` and `bundler: command not found: deploy`. Having no auto complete at all seems a worse solution, and there doesn't seem to be any way for it to work when typing the entire `bundle exec fastlane` string.

fastlane/lib/fastlane/commands_generator.rb
milch: Yup, sounds better to me! Maybe `should be enabled too` instead of `should work too`

milch (approved):

PR #11005, opened 2017-11-22
note: only `bundle exec ...` uses local fastlane

janpio:
Highlight that you have to use `bundle exec ...` for your local fastlane code checkout to be used.

### DIFF
--- a/YourFirstPR.md
+++ b/YourFirstPR.md
@@ -75,7 +75,7 @@ bundle show fastlane
 
 which should print out the path to your local development environment.
 
-From now on, every time you introduce a change to your local _fastlane_ code base, you can immediately test it by running  (`bundle exec fastlane …`. (Note that just using `fastlane …` without `bundle exec` will **not** use your local _fastlane_ code base!)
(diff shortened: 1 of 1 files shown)


milch (dismissed):
™️

YourFirstPR.md
```diff
@@ -75,7 +75,7 @@ bundle show fastlane
 
 which should print out the path to your local development environment.
 
-From now on, every time you introduce a change to your local _fastlane_ code base, you can immediately test it by running `bundle exec fastlane …`.
+From now on, every time you introduce a change to your local _fastlane_ code base, you can immediately test it by running  (`bundle exec fastlane …`. (Note that just using `fastlane …` without `bundle exec` will **not** use your local _fastlane_ code base!)
```
milch: I think you snuck an extra `( ` in there at the beginning 😅

milch (changes requested):

YourFirstPR.md
janpio: Good catch 🎣 Thanks.

janpio:
Great to see that I am not the only one confused by the review interface. I always open reviews without wanting it and submit my comments to early. Strange.

milch:
Yeah I was just confused because I typed in the comment above, and then wanted to submit it right away with the `Request Changes` button. Turns out it doesn't work that way 😅

milch (approved):

PR #11023, opened 2017-11-25
Docs: Fix (special) actions logo size

janpio:
Right now most of the tools / actions with logo's logos are too big in the default view. Example: https://docs.fastlane.tools/actions/scan/

An exception is `deliver`, which looks better: https://docs.fastlane.tools/actions/deliver/

I adapted all other actions to use the same html code.


KrauseFx (approved):
Nice catch, thanks @janpio 👍

KrauseFx:
Since this is a small change, I'll just go ahead and merge it for now 👍

PR #11043, opened 2017-11-28
Fix FastlaneCore::Helper.xcode_path for non-Mac environments

janpio:
`FastlaneCore::Helper.xcode_path` finds out if it is inside a test by checking for if `SpecHelper` is defined:
https://github.com/fastlane/fastlane/blob/5019acb53fc2adf050b4227c3e3216e365ba36ec/fastlane_core/lib/fastlane_core/helper.rb#L23-L25
If the module is only loaded after the check is first executed, it internally never realizes it is in a test and return the wrong value on non-Mac environments. Only on subsequent calls does it return the correct value.

Moving the SpecHelper module code before the first call to `FastlaneCore::Helper.xcode_path` fixes the problem.

### DIFF
--- a/spec_helper.rb
+++ b/spec_helper.rb
@@ -1,3 +1,7 @@
+# This module is only used to check the environment is currently a testing env
+module SpecHelper
+end
+
 require "coveralls"
 Coveralls.wear! unless ENV["FASTLANE_SKIP_UPDATE_CHECK"]
 
@@ -6,10 +10,6 @@ WebMock.disable_net_connect!(allow: 'coveralls.io')
 
 require "fastlane"
 UI = FastlaneCore::UI
-
-# This module is only used to check the environment is currently a testing env
-module SpecHelper
-end
 
 unless ENV["DEBUG"]
   $stdout.puts "Changing stdout to /tmp/fastlane_tests, set `DEBUG` environment variable to print to stdout (e.g. when using `pry`)"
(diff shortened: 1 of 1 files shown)


jdee (approved):
Looks good.

spec_helper.rb
```diff
@@ -12,6 +12,10 @@
   $stdout = File.open("/tmp/fastlane_tests", "w")
 end
 
+# This module is only used to check the environment is currently a testing env
```
milch: Why not add it at the very beginning?

janpio: Because I have noe idea what other strange side effects it could have.

Do you think and advice that I should just move it up to the top?

milch: Yeah I think it's a good idea. If ever someone else comes across this file and tries to add something they'll more likely add it below the module definition if it's at the beginning. If it has any adverse effects, we should see tests start to fail

janpio: Moved it further up, under the requires. Or even above those?

milch: Hmm, I think right at the beginning would be best, since other code is already called with the other requires

milch: By the way, since I saw you have a merge commit: you can avoid merge commits by doing `git pull --rebase origin master` to rebase your changes on top of the master branch

janpio: No rebase with the Git web UI ;)

jdee:
Usually coverage tools want to be required as the very first line of the spec helper. I don't think moving this up will break anything major, but I thought it was best after the requires.

jdee:
For example, this is how it is in the plugin template, and I think this is right.

https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/plugins/template/spec/spec_helper.rb.erb#L8

PR #11045, opened 2017-11-28
spec_helper.rb: Only check for valid Xcode path on Mac

### LINKED ISSUE #11044: Running tests on non-Mac OS impossible
These lines make running the tests (`bundle exec rake test_all`) on a non-Mac environments (where Xcode does not exist) impossible:
https://github.com/fastlane/fastlane/blob/b1e01920882fa2d05d022ba998d1edf780431a72/spec_helper.rb#L16-L23

Is there a better/other way to give this information to a user?

Would it be a valid workaround to give this output only if one is actually executing the tests on Mac?

janpio:
### Motivation and Context
See #11044 

### Description
Added an additional if around the check that makes sure the code is only executed if the platform is Mac.

fixes #11044

### DIFF
--- a/spec_helper.rb
+++ b/spec_helper.rb
@@ -12,8 +12,7 @@ unless ENV["DEBUG"]
   $stdout = File.open("/tmp/fastlane_tests", "w")
 end
 
-is_mac = FastlaneCore::Helper.is_mac?
-if is_mac
+if FastlaneCore::Helper.is_mac?
   xcode_path = FastlaneCore::Helper.xcode_path
   unless xcode_path.include?("Contents/Developer")
     UI.error("Seems like you didn't set the developer tools path correctly")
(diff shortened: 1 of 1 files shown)


spec_helper.rb
```diff
@@ -12,14 +12,17 @@
...
 
-xcode_path = FastlaneCore::Helper.xcode_path
-unless xcode_path.include?("Contents/Developer")
-  UI.error("Seems like you didn't set the developer tools path correctly")
-  UI.error("Detected path '#{xcode_path}'") if xcode_path.to_s.length > 0
-  UI.error("Please run the following on your machine")
-  UI.command("sudo xcode-select -s /Applications/Xcode.app")
-  UI.error("Adapt the path if you have Xcode installed/named somewhere else")
-  exit(1)
+is_mac = FastlaneCore::Helper.is_mac?
```
KrauseFx: It feels like assigning this value, only to use it one is overhead, we can just have that right next to the `if` keyword? :)

janpio: Sure we can - leftover from my wrangling with `unless`

KrauseFx (approved):
🎷

PR #11053, opened 2017-11-29
[Docs template] Actions: Linkable headline instead of bolded text

janpio:
Right now the "Parameters" table is not linkable with `#parameters` on action docs pages. 

This changes the headline above from a bolded text to a semantically correct headline - with the sideeffect to also generate a linkable id, similar to the main headline.


KrauseFx (approved):
Yeah 👍 Nice catch

PR #11068, opened 2017-11-30
Further improvements to sh

jdee:
### Checklist

**Updated**

### Motivation and Context

This change permits complete passthrough of a variadic argument list to `Open3.popen2e`, which interprets the the list in the same was as `Process.spawn` and `Kernel#system`.

```Ruby
sh "ls", "-la"
sh "cat", path, error_callback: ->(result) { ... }
sh { "FOO" => "Hello" }, "echo $FOO"
```

This makes `sh` much more flexible and powerful, while at the same time simplifying the code and removing any need for argument handling to enable this.

See http://ruby-doc.org/core-2.4.2/Process.html#method-c-spawn for details on the arguments that can be passed.

### Description

The variadic argument list is simply passed through from every variant of `sh` directly to `popen2e`. And that's it.

To construct a command to display, `Actions.shell_command_from_args` was introduced. This uses shell escaping and joining to produce a shell command equivalent to what is being executed. This method is well tested. It is also largely cosmetic, so issues can reasonably be addressed later if they arise.

It's worth mentioning one important change that applies to all variadic methods that handle commands, like `system`. If the command to be executed (ignoring any optional leading environment hash) is a single string, the string is passed to the shell to be executed. In all other cases, the arguments are passed directly to exec as the argv. No shell is invoked. There is no shell escaping and no shell to convert the escaped arguments back to the argv. This also saves a child process, since a shell forks each command.

This change also breaks the Array handling that was introduced in 2.68.0 but not yet advertised or documented. The reason is that `system` and friends allow overriding `argv[0]`. For example:

```Ruby
sh [ "/usr/local/bin/foo", "foo" ], "-x"
```

This executes the binary `/usr/local/bin/foo` but passes "foo" as `argv[0]`. This is how `popen2e` and `system` will interpret any array passed. I think it's just as well to proceed with this change, since the Array support is a half-measure. I don't think anyone will miss the brackets.

There is one syntactic corner case that may somehow be made to work and is probably not worth supporting if it does not. However, it's difficult. Because `sh` accepts its own keywords, e.g.
`error_callback`, it is difficult to pass a trailing options Hash. Compare:

```Ruby
system "ls", "-la", chdir: "/tmp"
```

This changes directory to `/tmp` and runs the command.

```Ruby
sh "ls", "-la", chdir: "/tmp"
```

This generates `no such keyword: chdir`. It's difficult and maybe impossible to pass this hash to a method with keyword arguments. I think I have a fix, but this may not be worth supporting. Other options can be used to redirect output, e.g., which would at least be confusing and possibly cause errors. For now, I've ignored any handling of a trailing options hash to `popen2e`.

This change makes `sh` much more like the `system` call. Here's a comparison:

|sh|system|
|--|-------|
|Raises Errno::ENOENT if command to be executed is not found|Raises Errno::ENOENT if command to be executed is not found|
|Output is returned on success. An exception is raised on error.|true is returned on success. false is returned on error.|
|$? is set to the Process::Status of the command|$? is set to the Process::Status of the command|
|calls error_callback on error||
|does not accept a trailing options Hash|accepts a trailing options Hash|
|outputs the command and formatted output via UI|does not output the command or format the output|
|shorter method name|longer method name|

### DIFF
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -70,6 +70,7 @@ module Fastlane
     end
 
     def self.command_from_args(*args)
+      raise ArgumentError, "sh requires at least one argument" unless args.count > 0
       if args.count == 1
         command = args.first
         command = command.shelljoin if command.kind_of?(Array)
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -27,7 +27,7 @@ module Fastlane
       Encoding.default_external = Encoding::UTF_8
       Encoding.default_internal = Encoding::UTF_8
 
-      command = command_from_args(*command)
+      command = shell_command_from_args(*command)
       UI.command(command) if print_command
 
       result = ''
@@ -69,7 +69,7 @@ module Fastlane
       Encoding.default_internal = previous_encoding.last
     end
 
-    def self.command_from_args(*args)
+    def self.shell_command_from_args(*args)
       raise ArgumentError, "sh requires at least one argument" unless args.count > 0
       if args.count == 1
         command = args.first
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -8,26 +8,26 @@ module Fastlane
     # When running this in tests, it will return the actual command instead of executing it
     # @param log [Boolean] should fastlane print out the executed command
     # @param error_callback [Block] a callback invoked with the command output if there is a non-zero exit status
-    def self.sh(command, log: true, error_callback: nil)
-      sh_control_output(command, print_command: log, print_command_output: log, error_callback: error_callback)
+    def self.sh(*command, log: true, error_callback: nil)
+      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback)
     end
 
-    def self.sh_no_action(command, log: true, error_callback: nil)
-      sh_control_output(command, print_command: log, print_command_output: log, error_callback: error_callback)
+    def self.sh_no_action(*command, log: true, error_callback: nil)
+      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback)
     end
 
-    # @param command [String] The command to be executed
+    # @param command [String, Array] The command to be executed
     # @param print_command [Boolean] Should we print the command that's being executed
     # @param print_command_output [Boolean] Should we print the command output during execution
     # @param error_callback [Block] A block that's called if the command exits with a non-zero status
-    def self.sh_control_output(command, print_command: true, print_command_output: true, error_callback: nil)
+    def self.sh_control_output(*command, print_command: true, print_command_output: true, error_callback: nil)
       print_command = print_command_output = true if $troubleshoot
       # Set the encoding first, the user might have set it wrong
       previous_encoding = [Encoding.default_external, Encoding.default_internal]
       Encoding.default_external = Encoding::UTF_8
       Encoding.default_internal = Encoding::UTF_8
 
-      command = Shellwords.join(command) if command.kind_of?(Array) # since it's an array of one element when running from the Fastfile
+      command = command_from_args(*command)
       UI.command(command) if print_command
 
       result = ''
@@ -67,6 +67,16 @@ module Fastlane
     ensure
       Encoding.default_external = previous_encoding.first
       Encoding.default_internal = previous_encoding.last
+    end
+
+    def self.command_from_args(*args)
+      if args.count == 1
+        command = args.first
+        command = command.shelljoin if command.kind_of?(Array)
+      else
+        command = args.shelljoin
+      end
+      command
     end
   end
 end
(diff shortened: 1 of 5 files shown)


fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -68,5 +68,15 @@ def self.sh_control_output(command, print_command: true, print_command_output: t
       Encoding.default_external = previous_encoding.first
       Encoding.default_internal = previous_encoding.last
     end
+
+    def self.command_from_args(*args)
+      if args.count == 1
+        command = args.first
+        command = command.shelljoin if command.kind_of?(Array)
```
milch: Is this to account for nested arrays? i.e. someone calls `sh [["git", "commit", "-m", "my message"]]` this is what makes it work?

jdee: @milch You're probably familiar with variadic Ruby methods, but for anyone else looking, here's how it works:

```Ruby
def sh(*command, log: true, error_callback: nil)
```

defines a variadic method that can take any number of arguments as well as named options at the end. The `command` arg is always received by the method as an Array, regardless how many args are passed. There is special handling for the case of one element in the array. This applies when you have:

```Ruby
sh "git commit -m #{message.shellescape}" # line 74
sh ["git", "commit", "-m", message] # line 75
```

Without the special handling for one element (if we called `args.shelljoin` for all input), things like `"git commit"` would be shell escaped to `"git\ commit"`, which would break

If the array has multiple elements, the array is shelljoined, so the case for `count != 1` handles this:

```Ruby
sh "git", "commit", "-m", message # line 77
```

The case with one element is important mainly for current Fastfiles and anything else that is building a shell-escaped string (or doesn't require a shell escape because no special chars, spaces in args, etc.). There can also be times when you have a string from somewhere that is a full command. Maybe the user types it into a config. file or something. Then you'd have to shellsplit the string in order to pass it into this method if it didn't handle a single string argument. 

The case where `command.count == 1 && command.first.kind_of?(Array)` handles `sh ["git", "commit", ...]`. Since support for this was only introduced yesterday, it's less critical, but there can be times where you build your command as an array and just want to pass the array. (Otherwise you'd have to pass `*array`.) And it can be nice to use delimiters to keep track of a lot of args:

```Ruby
sh ["git", "commit", "-m", message, path1, path2, path3], error_callback: ->(result) { ... }
```

There is one small thing I realized while writing this, which is when `command.count == 0`, you end up with a blank string. If you just write:

```Ruby
sh
```

It will attempt to execute `""`. This is probably not a major problem, but it would perhaps be best to raise `ArgumentError` if nothing is passed. I'll add a commit for that.

milch: Alright, thanks for the in-depth explanation. I wasn't sure wether the `*` operator would also wrap a single array argument in a second array, but from your explanation it seems like it does. I'll take another look once you've added the commit that handles the case when there are no arguments

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -68,5 +68,15 @@ def self.sh_control_output(command, print_command: true, print_command_output: t
       Encoding.default_external = previous_encoding.first
       Encoding.default_internal = previous_encoding.last
     end
+
+    def self.command_from_args(*args)
```
milch: Some nitpicking: maybe rename this from `command_from_args` to `shell_command_from_args`. Even though it is in the shell_helper file it can be called from outside via `Fastlane::Actions` so this should make it more explicit what is happening

jdee:
This is the right idea, but the wrong implementation. One inspiration for this was rake, whose `sh` method works this way. I finally dug in to see how rake is doing it, and I found that this is actually just basic system Ruby command handling.

https://ruby-doc.org/core-2.4.2/Process.html#method-c-spawn

The `Open3.popen2e` method, along with `Kernel#system` and quite a few others directly honor this variadic form. So the simplest way to do this is not by using `Shellwords.join` but by just passing the argument list directly to `Open3.popen2e`: `Open3.popen2e(*command)`. Then you can also take advantage of things like the optional `env` Hash as the first argument, e.g.:

```Ruby
sh({ "FOO" => "Hello" }, "echo $FOO")
```

(You can already do this using `system`.)

This simplifies and improves everything and I think is finally the right solution.

However, it makes the `shell_command_from_args` method more complicated. In order to echo the correct command to the terminal (e.g. with an `env` Hash as above), that method has to handle everything, or I'll have to find a built-in Ruby method to do it.

I have everything but that working locally. I'll get that together and push later.

milch:
@jdee Huh, interesting. Didn't know you could do that with `system`. 

I looked through the Ruby source to see if they have a method that they use to evaluate the arguments. The `system`, `exec`, `spawn` (and so on) commands are implemented in C and use [these methods](https://github.com/ruby/ruby/blob/trunk/process.c#L4334) to parse the arguments. Unfortunately, it doesn't seem like they are exported anywhere, so there is probably no builtin way to get that (from ruby). Building a C extension seems possible, but I wouldn't do that to our users

jdee:
@milch Any further thoughts after the last round of changes? I updated the description. I think this is basically the right thing to do.

milch (approved):
@jdee I just looked through it again, looks good to me! 👍

Issue #4675, opened 2016-05-16
Supply tries to upload `app-debug.apk` to instead of `app-release.apk`


mmazzarolo:
Hello everyone!

### Issue Description
Supply tries to upload `app-debug.apk` instead of `app-release.apk` (I guess it checks for `apk_paths` instead of `apk` if running gradle and supply in the same lane... but why is `apk_paths` not empty?).

The interested line in the log output is:
`
[14:48:08]: Preparing apk at path '/Users/matteo/dev/tony-app-react/android/app/build/outputs/apk/app-debug.apk' for upload...`

##### Complete output when running fastlane, including the stack trace and command used

```
matteo@Mini-di-Patrick ~/d/tony-app-react (master)> fastlane android alpha
[14:46:55]: -------------------------------------------------
[14:46:55]: --- Step: Verifying required fastlane version ---
[14:46:55]: -------------------------------------------------
[14:46:55]: fastlane version valid
[14:46:55]: ------------------------------
[14:46:55]: --- Step: default_platform ---
[14:46:55]: ------------------------------
[14:46:55]: Driving the lane 'android alpha' 🚀
[14:46:55]: --------------------
[14:46:55]: --- Step: gradle ---
[14:46:55]: --------------------
[14:46:55]: $ /Users/matteo/dev/tony-app-react/android/gradlew assembleRelease -p android/
[14:46:57]: ▸ :app:preBuild UP-TO-DATE
[14:46:57]: ▸ :app:preReleaseBuild UP-TO-DATE
[14:46:57]: ▸ :app:checkReleaseManifest
[14:46:57]: ▸ :app:preDebugBuild UP-TO-DATE
[14:46:57]: ▸ :react-native-vector-icons:compileLint
[14:46:57]: ▸ :react-native-vector-icons:copyReleaseLint UP-TO-DATE
[14:46:57]: ▸ :react-native-vector-icons:preBuild UP-TO-DATE
[14:46:57]: ▸ :react-native-vector-icons:preReleaseBuild UP-TO-DATE
[14:46:57]: ▸ :react-native-vector-icons:checkReleaseManifest
[14:46:57]: ▸ :react-native-vector-icons:preDebugAndroidTestBuild UP-TO-DATE
[14:46:57]: ▸ :react-native-vector-icons:preDebugBuild UP-TO-DATE
[14:46:57]: ▸ :react-native-vector-icons:preDebugUnitTestBuild UP-TO-DATE
[...]

PR #11086, opened 2017-12-01
Android fastfile template: clean before build

slorber:
This ensures that newcomers don't encounter this issue: https://github.com/fastlane/fastlane/issues/4675

### DIFF
--- a/fastlane/lib/assets/FastfileTemplateAndroid
+++ b/fastlane/lib/assets/FastfileTemplateAndroid
@@ -27,8 +27,7 @@ platform :android do
 
   desc "Submit a new Beta Build to Crashlytics Beta"
   lane :beta do
-    gradle(task: 'clean')
-    gradle(task: "assembleRelease")
+    gradle(task: "clean assembleRelease")
     crashlytics
 
     # sh "your_script.sh"
@@ -37,8 +36,7 @@ platform :android do
 
   desc "Deploy a new version to the Google Play"
   lane :deploy do
-    gradle(task: 'clean')
-    gradle(task: "assembleRelease")
+    gradle(task: "clean assembleRelease")
     upload_to_play_store
   end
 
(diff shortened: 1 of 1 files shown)


slorber:
I signed it!

janpio:
@slorber PR looks good.

Just to make sure this is actually the correct solution: Why is #4675 actually happening? Is `supply` just getting whatever is in the `build` folder? Or _why_ is the `clean` necessary?

slorber:
@janpio as far as I understand I built my app with Android Studio which generated a debug apk and the default Fastlane android template did try assemble release (so there are now debug + release apk) and upload debug + release.

In my case it failed because the versionCode of the debug apk was already existing, because it was not incremented.

I think it's decent to cleanup build folder before assembling new release.

Other solution could be to make sure the release build is uploaded.

Don't really know if it makes sense to upload a debug apk to google play store, I'm a newbie Android developer (React Native actually)

janpio:
Ok, this might point to a deeper problem with `supply` (now called `upload_to_play_store`) in the way it picks up APKs. Will look into this (or at least try to understand what it does).

(I am not qualified to decide if a `clean` before each build is useful. Hope someone else will look into this.)

ohayon:
Oh, hai, @fastlane/android, any chance you folks could take a peek at this? 🤖

fastlane/lib/assets/FastfileTemplateAndroid
```diff
@@ -27,6 +27,7 @@ platform :android do
 
   desc "Submit a new Beta Build to Crashlytics Beta"
   lane :beta do
+    gradle(task: 'clean')
     gradle(task: "assembleRelease")
```
casz: You could make into one Gradle action `gradle(task: "clean assembleRelease")`

slorber:
@casz thanks I've updated the pr

janpio:
Hi @slorber, sorry that you had to wait so long for a reaction. 

Unfortunately CI is failing with a failing test, maybe something caused by the additional clean? Also there is a merge conflict now that has to be taken care of. Could you please rebase onto `master`? (This will also retrigger a CI run with current CI setup which might give more information.)

slorber:
@janpio I've created a new PR https://github.com/fastlane/fastlane/pull/11873 because the code has been refactored since last PR.

PR #11117, opened 2017-12-05
Show plugin actions on the plugin page

milch:
To help with plugin discovery and give a better preview of what each plugin does, this change lists all actions that a plugin provides, the description and category.

Preview:

screen shot 2017-12-05 at 19 55 46

### DIFF
--- a/fastlane/helper/plugin_scores_helper.rb
+++ b/fastlane/helper/plugin_scores_helper.rb
@@ -15,6 +15,20 @@ module Fastlane
         end
       end
 
+      class FastlanePluginAction
+        attr_accessor :name
+        attr_accessor :category
+        attr_accessor :description
+        attr_accessor :details
+
+        def initialize(name: nil, category: nil, description: nil, details: nil)
+          self.name = name
+          self.category = category
+          self.description = description
+          self.details = details
+        end
+      end
+
       class FastlanePluginScore
         attr_accessor :name
         attr_accessor :downloads
@@ -44,7 +58,8 @@ module Fastlane
             has_github_page: has_github_page,
             has_mit_license: includes_license?("MIT"),
             has_gnu_license: includes_license?("GNU") || includes_license?("GPL"),
-            major_release: Gem::Version.new(hash["version"]) >= Gem::Version.new("1.0.0")
+            major_release: Gem::Version.new(hash["version"]) >= Gem::Version.new("1.0.0"),
+            actions: []
           }
 
           if has_github_page
@@ -154,6 +169,12 @@ module Fastlane
                 tests_count += File.read(spec_file).scan(/ it /).count
               end
               self.data[:tests] = tests_count
+
+              actions = Dir["**/actions/*.rb"].map do |action_file|
+                FastlaneActionFileParser.new.parse_file(action_file)
+              end
+              # Result of the above is an array of arrays, this merges all of them into data[:actions]
+              self.data[:actions].concat(*actions)
             end
           end
         end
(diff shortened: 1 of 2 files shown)


fastlane/helper/plugin_scores_helper.rb
```diff
@@ -201,6 +222,121 @@ def append_github_data
...
+            puts "Last string statement: #{last_string_statement}, last method name: #{last_method_name}"
+          end
+          actions
+        end
+
+        def is_block?(line)
+          # Matches beginning block statement from the list of keywords
+          # `do` has to be handled separately because it can be preceded by a function call
+          line.match?(/^(?:[\s\w]*=)?\s*(if|unless|case|while|loop|for|until|begin)/) \
+          || line.match?(/^(?:[\s\w]*=)?\s*.*\s+do\s*(?:\|.*\|)?\s*(?:#.*)?$/)
```
milch: For the record this regex is not that accurate so might misfire in some cases... I previously used this regex: 

```regex
^(?:[\s\w]*=)?\s*(?:(?:[\w\.]+(?:\[.*\])*)+(?:\(.*\))?)*\s+do\s*(?:\|.*\|)?\s*(?:#.*)?$
```

Which works wonderfully in JS and matches exactly one variable assignment, a variable number of method calls including subscripts, and then a trailing do block. In ruby it hangs the interpreter though so I had to go with the current solution

jdee: Is there any way to actually require each plugin and use reflection if necessary to find the action classes it implements? That might end up being cleaner or simpler than the regexp approach, but there would be issues with dependencies.

jdee (approved):
I wished for this in Slack some time ago. Nice work. I made a comment about an alternate approach, but that could be an iteration--if it's even feasible.

fastlane/helper/plugin_scores_helper.rb
milch: I thought about 3 options when coming up with this PR: 

1. Parse the ruby file into an AST using a 3rd party plugin, grab values from AST 
2. `require` the file and grab the metadata we want to show via ruby directly
3. Parse a subset of ruby that includes the information we need and grab values like that

I decided against option 1 because I thought bringing in a 3rd party plugin would introduce additional maintenance burden (if the plugin API changes we need to update that code) and I guessed that the number of lines to parse the file and get the values that we need out of the AST would be comparable in length to option 3

I also decided against option 2 because of security considerations, since we're downloading arbitrary code from the internet. IIRC even if you just require a file top-level code is run, so we would have to build in some safeguards to make sure that no malicious code can be run like that. Even then, the stakes are high and in a worst case scenario if we have a bug in our code an attacker could get access to contributors' computers or CI systems. With options 1 & 3 the worst case scenario is that for some plugins no actions are shown, which is an acceptable trade-off

milch:
Note: Usually I would add tests for something like this, but the actions fastlane itself uses for deployment aren't tested anywhere, and I don't necessarily want to add these to the main fastlane tests

fastlane/helper/plugin_scores_helper.rb
jdee: I suppose there is a potential security concern no matter how you handle it, but it seems like the most sensible approach, and I wonder if there's some way to mitigate the security concern, but I don't have one right away. For what it's worth, it will also produce arbitrary text. Someone could produce a plugin with libelous content for the description, e.g., and Fastlane would end up rendering it on the docs site.

milch: Aye that's true, but that was always possible by just giving the plugin an obscene name or something. At least in the action description it's hidden away until a user un-collapses it

fastlane/helper/plugin_scores_helper.rb
```diff
@@ -154,6 +169,12 @@ def append_git_data
                 tests_count += File.read(spec_file).scan(/ it /).count
               end
               self.data[:tests] = tests_count
+
+              actions = Dir["**/actions/*.rb"].map do |action_file|
+                FastlaneActionFileParser.new.parse_file(action_file)
+              end
+              # Result of the above is an array of arrays, this merges all of them into data[:actions]
+              self.data[:actions].concat(*actions)
```
revolter: @milch, Any idea why I'm getting:

```
wrong number of arguments (given 8, expected 1)
...
bundler: failed to load command: fastlane (/Users/revolt/.gems/bin/fastlane)
ArgumentError: [!] wrong number of arguments (given 8, expected 1)
  /Users/revolt/Development/GitHub/revolter/fastlane/fastlane/helper/plugin_scores_helper.rb:177:in `concat'
  /Users/revolt/Development/GitHub/revolter/fastlane/fastlane/helper/plugin_scores_helper.rb:177:in `block (2 levels) in append_git_data'
  /Users/revolt/Development/GitHub/revolter/fastlane/fastlane/helper/plugin_scores_helper.rb:150:in `chdir'
...
```

when running `bundle exec fastlane update_docs`?

revolter: It looks like it works with ruby 2.4 but not 2.3

PR #11096, opened 2017-12-03
[WIP] [frameit] Ensure both texts use same baseline

funnel20:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
Output: 
```
Finished in 6 minutes 46 seconds (files took 3.7 seconds to load)
4675 examples, 0 failures
```
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
Output: 
```
973 files inspected, no offenses detected
```
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.
Not applicable

### Motivation and Context

This solves issue https://github.com/fastlane/fastlane/issues/10440

The current implementation applies a `.trim` to both images for Keyword and Title.
Due to typography differences (for example the "L" that needs more vertical space than an "e") the generic baseline between 2 texts is lost. This can be seen in the excellent analysis of @janpio in this comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348010988

_Before_
iphone6-01_remotetab_framed

This solution requests the calculated trim box and offset without applying it.
By comparing the vertical top offsets of both trim boxes, the vertical offset between them can be calculated. The box with the biggest offset will be compensated to equal the same of the other, so their text baseline will be equal.
Finally a `.crop` will be applied with the (compensated) trim boxes.

_After_
iphone6-01_remotetab_framed

Test results can be found in issue comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348760923

### Description

The changes can be found in issue comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348760717

Please note that:

- I've no Ruby skills, so this might not be proper Production code. But the description and the code show the intended behaviour.
- All `print` commands are for debugging, so can be removed.
- Hence @janpio offered to help, so I added [WIP] to the title of the PR.
- The current code aligns the trimmed/cropped images to the top. To make this fix robust for any future changes (for example align to bottom), the height of both trim boxes (parameter 2) should be made equal: the largest after the current vertical top offset compensations is leading.

Issue #10440, opened 2017-09-27
[Frameit] The keyword and title sometimes have a vertical offset 


funnel20:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I use Frameit with pleasure to frame my screenshots and add a localised keyword and title to it.
The majority of text is aligned correctly:
![iphone5s-03_keyboarddiscovered-589996837d813aace56d9a0cf7b764a0_framed](https://user-images.githubusercontent.com/2139844/30935231-52fefd12-a3d0-11e7-9cba-b786f8fc268f.png)

Some of the framed screenshots show a vertical offset between the keyword and title. The offset might be a few pixels:
![iphone5s-02_smarttab-589996837d813aace56d9a0cf7b764a0_framed](https://user-images.githubusercontent.com/2139844/30935245-62b46080-a3d0-11e7-89e6-2a3233ddd49b.png)

Or massive:
![iphone5s-01_remotetab-589996837d813aace56d9a0cf7b764a0_framed](https://user-images.githubusercontent.com/2139844/30935287-7a33d628-a3d0-11e7-8223-5e31a7bda185.png)

My Framefile:
```json
{
    "default": {
        "keyword": {
            "font": "./fonts/SF-UI-Display-Semibold.otf",
            "color": "#007AFF"
        },
        "title": {
            "font": "./fonts/SF-UI-Display-Regular.otf",
            "color": "#002955"
        },
        "background": "background.png",
        "padding": 25,
        "show_complete_frame": true,
        "stack_title" : false
    },
    "data": [
             {
             "filter": "01"
             },
             {
             "filter": "02"
             },
             {
             "filter": "03"
             },
             {
             "filter": "04"
             },
             {
             "filter": "05"
             }
             ]
}
```

These are French, but I have got it in other languages too.
The ones with a small offset, have a special character in either the keyword or title. For example in German the "ä", but also the apostrophe (') in a Dutch word.
But I can't explain the massive offset in the example above, where no special character is used. The behaviour seems somehow random.

Here are the *.strings files for the above French screenshots:

_keyword.strings_
```strings

"01" = "Familier";
"02" = "Logos";
[...]

PR #11121, opened 2017-12-06
[WIP] [frameit] Ensure both texts use same baseline

funnel20:
### Checklist
Output: 
```
Finished in 6 minutes 46 seconds (files took 3.7 seconds to load)
4675 examples, 0 failures
```
Output: 
```
973 files inspected, no offenses detected
```
Not applicable

### Motivation and Context

This solves issue https://github.com/fastlane/fastlane/issues/10440

The current implementation applies a `.trim` to both images for Keyword and Title.
Due to typography differences (for example the "L" that needs more vertical space than an "e") the generic baseline between 2 texts is lost. This can be seen in the excellent analysis of @janpio in this comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348010988

_Before_
iphone6-01_remotetab_framed

This solution requests the calculated trim box and offset for each image **without applying it**.
By comparing the vertical top offsets and heights of both trim boxes, the vertical offset between them can be calculated. The box with the biggest offset will be compensated to equal the same of the other, so their text baseline will be equal. After this the height is made equal, so they share the same bottom offset.
Finally a `.crop` will be applied with the (compensated) trim boxes.

_After_
iphone6-01_remotetab_framed

Test results can be found in issue comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348760923

Here the difference is clear, while highlighting the trim box in yellow:

Original code with `.trim`:
iphone6-01_remotetab_framed trim

My initial commit with `.crop` to align top:
iphone6-01_remotetab_framed vertical top

My robust solution with `.crop` for both align top and align bottom:
iphone6-01_remotetab_framed vertical bottom

It also works when only one of the images needs to be adjusted both on the top **and** bottom side. In this case for "en":
iphone6-01_remotetab_framed

### Description

The changes can be found in issue comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348760717

Please note that:

- This will supersede PR #11096 as I committed that anonymously.
- I've no Ruby skills, so this might not be proper Production code. But the description and the code show the intended behaviour.
- Hence @janpio offered to help, so I added [WIP] to the title of the PR.

### DIFF
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -319,7 +319,8 @@ module Frameit
       words = [:keyword, :title].keep_if { |a| fetch_text(a) } # optional keyword/title
       results = {}
       trim_boxes = {}
-      min_vertical_trim_offset = Float::INFINITY # Init at a large value, as the code will search for a minimal value.
+      top_vertical_trim_offset = Float::INFINITY # Init at a large value, as the code will search for a minimal value.
+      bottom_vertical_trim_offset = 0
       words.each do |key|
         # Create empty background
         empty_path = File.join(Frameit::ROOT, "lib/assets/empty.png")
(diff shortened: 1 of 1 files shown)


funnel20:
This PR is obsolete and superseded by #11124

PR #11124, opened 2017-12-06
[frameit] Ensure both texts use same baseline (issue #10440)

funnel20:
### Checklist
Output: 
```
Finished in 6 minutes 46 seconds (files took 3.7 seconds to load)
4675 examples, 0 failures
```
Output: 
```
973 files inspected, no offenses detected
```
Not applicable

### Motivation and Context

This solves issue https://github.com/fastlane/fastlane/issues/10440

The current implementation applies a `.trim` to both images for Keyword and Title.
Due to typography differences (for example the "L" that needs more vertical space than an "e") the generic baseline between 2 texts is lost. This can be seen in the excellent analysis of @janpio in this comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348010988

_Before_
iphone6-01_remotetab_framed

This solution requests the calculated trim box and offset for each image **without applying it**.
By comparing the vertical top offsets and heights of both trim boxes, the vertical offset between them can be calculated. The box with the biggest offset will be compensated to equal the same of the other, so their text baseline will be equal. After this the height is made equal, so they share the same bottom offset.
Finally a `.crop` will be applied with the (compensated) trim boxes.

_After_
iphone6-01_remotetab_framed

Test results can be found in issue comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348760923

Here the difference is clear, while highlighting the trim box in yellow:

Original code with `.trim`:
iphone6-01_remotetab_framed trim

My initial commit with `.crop` to align top:
iphone6-01_remotetab_framed vertical top

My robust solution with `.crop` for both align top and align bottom:
iphone6-01_remotetab_framed vertical bottom

It also works when only one of the images needs to be adjusted both on the top **and** bottom side. In this case for "en":
iphone6-01_remotetab_framed

### Description

The changes can be found in issue comment: https://github.com/fastlane/fastlane/issues/10440#issuecomment-348760717

Please note that:

- This will supersede PR #11096 as I committed that anonymously.
- I've no Ruby skills, so this might not be proper Production code. But the description and the code show the intended behaviour.
- Hence @janpio offered to help, so I added [WIP] to the title of the PR.

### DIFF
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -359,25 +359,21 @@ module Frameit
           b.format("%@") # CALCULATED: trim bounding box (without actually trimming), see: http://www.imagemagick.org/script/escape.php
         end
 
-        # The .identify with format("%@") returns a string with "<width>x<height>+<offset_x>+<offset_y>". Extract these 4 parameters into an integer array, by using multiple string separators: the whitespace "\s", "x" and "+":
-        trim_values = calculated_trim_box.split(/[\sx+]/).map(&:to_i)
-
-        # Get the height and vertical offset:
-        trim_box_height = trim_values[1]
-        trim_box_offset_y = trim_values[3]
+        # Create a Trimbox object from the MiniMagick .identify string with syntax "<width>x<height>+<offset_x>+<offset_y>":
+        trim_box = Frameit::Trimbox.new(calculated_trim_box)
 
         # Get the minimum top offset of the trim box:
-        if trim_box_offset_y < top_vertical_trim_offset
-          top_vertical_trim_offset = trim_box_offset_y
+        if trim_box.offset_y < top_vertical_trim_offset
+          top_vertical_trim_offset = trim_box.offset_y
         end
 
         # Get the maximum bottom offset of the trimbox, this is the top offset + height:
-        if (trim_box_offset_y + trim_box_height) > bottom_vertical_trim_offset
-          bottom_vertical_trim_offset = trim_box_offset_y + trim_box_height
+        if (trim_box.offset_y + trim_box.height) > bottom_vertical_trim_offset
+          bottom_vertical_trim_offset = trim_box.offset_y + trim_box.height
         end
 
         # Store for the crop action:
-        trim_boxes[key] = trim_values
+        trim_boxes[key] = trim_box
       end
 
       # Crop images based on top_vertical_trim_offset and bottom_vertical_trim_offset to maintain text baseline:
@@ -385,34 +381,27 @@ module Frameit
         # Get matching trim box:
         trim_box = trim_boxes[key]
 
-        # Get the height and vertical offset:
-        trim_box_height = trim_box[1]
-        trim_box_offset_y = trim_box[3]
-
         # Determine the trim area by maintaining the same vertical top offset based on the smallest value from all trim boxes (top_vertical_trim_offset).
         # When the vertical top offset is larger than the smallest vertical top offset, the trim box needs to be adjusted:
-        if trim_box_offset_y > top_vertical_trim_offset
+        if trim_box.offset_y > top_vertical_trim_offset
           # Increase the height of the trim box with the difference in vertical top offset:
-          trim_box_height += trim_box_offset_y - top_vertical_trim_offset
+          trim_box.height += trim_box.offset_y - top_vertical_trim_offset
           # Change the vertical top offset to match that of the others:
-          trim_box_offset_y = top_vertical_trim_offset
+          trim_box.offset_y = top_vertical_trim_offset
 
           UI.verbose("Trim box for key \"#{key}\" is adjusted to align top: #{trim_box}\n")
         end
 
         # Check if the height needs to be adjusted to reach the bottom offset:
-        if (trim_box_offset_y + trim_box_height) < bottom_vertical_trim_offset
+        if (trim_box.offset_y + trim_box.height) < bottom_vertical_trim_offset
           # Set the height of the trim box to the difference between vertical bottom and top offset:
-          trim_box_height = bottom_vertical_trim_offset - trim_box_offset_y
+          trim_box.height = bottom_vertical_trim_offset - trim_box.offset_y
 
           UI.verbose("Trim box for key \"#{key}\" is adjusted to align bottom: #{trim_box}\n")
         end
 
-        # Convert trim box parameters to string for crop method (format: "<width>x<height>+<offset_x>+<offset_y>"):
-        crop_input = "#{trim_box[0]}x#{trim_box_height}+#{trim_box[2]}+#{trim_box_offset_y}"
-
-        # Crop image:
-        results[key].crop(crop_input)
+        # Crop image with adjusted trim box parameters in MiniMagick string format:
+        results[key].crop(trim_box.string_format)
       end
 
       results
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -351,58 +351,65 @@ module Frameit
           i.fill fetch_config[key.to_s]['color']
         end
 
-        # Trimming the image will result in the loss of the common baseline between the text in all images.
-        # title_image.trim # remove white space
-
         results[key] = title_image
 
-        # Hence retrieve the calculated trim bounding box:
-        # CALCULATED: trim bounding box (without actually trimming), see: http://www.imagemagick.org/script/escape.php
-        calculated_trim_box = title_image.identify { |b| b.format("%@") }
+        # Natively trimming the image with .trim will result in the loss of the common baseline between the text in all images.
+        # Hence retrieve the calculated trim bounding box without actually trimming:
+        calculated_trim_box = title_image.identify do |b|
+          b.format("%@") # CALCULATED: trim bounding box (without actually trimming), see: http://www.imagemagick.org/script/escape.php
+        end
 
-        # Get vertical top offset from the trim box by converting it from string to an integer array:
-        trim_values = calculated_trim_box.split(/[\sx+]/).map(&:to_i) # The format is "wxh+X+Y", so use multiple string separators: the whitespace "\s", "x" and "+"
+        # The .identify with format("%@") returns a string with "<width>x<height>+<offset_x>+<offset_y>". Extract these 4 parameters into an integer array, by using multiple string separators: the whitespace "\s", "x" and "+":
+        trim_values = calculated_trim_box.split(/[\sx+]/).map(&:to_i)
+
+        # Get the height and vertical offset:
+        trim_box_height = trim_values[1]
+        trim_box_offset_y = trim_values[3]
 
         # Get the minimum top offset of the trim box:
-        if trim_values[3] < top_vertical_trim_offset
-          top_vertical_trim_offset = trim_values[3]
+        if trim_box_offset_y < top_vertical_trim_offset
+          top_vertical_trim_offset = trim_box_offset_y
         end
 
         # Get the maximum bottom offset of the trimbox, this is the top offset + height:
-        if (trim_values[3] + trim_values[1]) > bottom_vertical_trim_offset
-          bottom_vertical_trim_offset = trim_values[3] + trim_values[1]
+        if (trim_box_offset_y + trim_box_height) > bottom_vertical_trim_offset
+          bottom_vertical_trim_offset = trim_box_offset_y + trim_box_height
         end
 
         # Store for the crop action:
         trim_boxes[key] = trim_values
       end
 
-      # Crop images based on top_vertical_trim_offset to maintain text baseline:
+      # Crop images based on top_vertical_trim_offset and bottom_vertical_trim_offset to maintain text baseline:
       words.each do |key|
         # Get matching trim box:
         trim_box = trim_boxes[key]
 
+        # Get the height and vertical offset:
+        trim_box_height = trim_box[1]
+        trim_box_offset_y = trim_box[3]
+
         # Determine the trim area by maintaining the same vertical top offset based on the smallest value from all trim boxes (top_vertical_trim_offset).
-        # The 4th parameter is the vertical top offset. When it's larger than the smallest vertical top offset, the trim box needs to be adjusted:
-        if trim_box[3] > top_vertical_trim_offset
-          # Increase the height of the trim box (2nd parameter) with the difference in vertical top offset:
-          trim_box[1] += trim_box[3] - top_vertical_trim_offset
+        # When the vertical top offset is larger than the smallest vertical top offset, the trim box needs to be adjusted:
+        if trim_box_offset_y > top_vertical_trim_offset
+          # Increase the height of the trim box with the difference in vertical top offset:
+          trim_box_height += trim_box_offset_y - top_vertical_trim_offset
           # Change the vertical top offset to match that of the others:
-          trim_box[3] = top_vertical_trim_offset
+          trim_box_offset_y = top_vertical_trim_offset
 
           UI.verbose("Trim box for key \"#{key}\" is adjusted to align top: #{trim_box}\n")
         end
 
-        # Repeat the same for the bottom offset:
-        if (trim_box[3] + trim_box[1]) < bottom_vertical_trim_offset
-          # Set the height of the trim box (2nd parameter) to the difference between vertical bottom and top offset:
-          trim_box[1] = bottom_vertical_trim_offset - trim_box[3]
+        # Check if the height needs to be adjusted to reach the bottom offset:
+        if (trim_box_offset_y + trim_box_height) < bottom_vertical_trim_offset
+          # Set the height of the trim box to the difference between vertical bottom and top offset:
+          trim_box_height = bottom_vertical_trim_offset - trim_box_offset_y
 
           UI.verbose("Trim box for key \"#{key}\" is adjusted to align bottom: #{trim_box}\n")
         end
 
-        # Convert trim box parameters to string for crop method (format: wxh+X+Y):
-        crop_input = "#{trim_box[0]}x#{trim_box[1]}+#{trim_box[2]}+#{trim_box[3]}"
+        # Convert trim box parameters to string for crop method (format: "<width>x<height>+<offset_x>+<offset_y>"):
+        crop_input = "#{trim_box[0]}x#{trim_box_height}+#{trim_box[2]}+#{trim_box_offset_y}"
 
         # Crop image:
         results[key].crop(crop_input)
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -318,6 +318,9 @@ module Frameit
     def build_title_images(max_width, max_height)
       words = [:keyword, :title].keep_if { |a| fetch_text(a) } # optional keyword/title
       results = {}
+      trim_boxes = {}
+      top_vertical_trim_offset = Float::INFINITY # Init at a large value, as the code will search for a minimal value.
+      bottom_vertical_trim_offset = 0
       words.each do |key|
         # Create empty background
         empty_path = File.join(Frameit::ROOT, "lib/assets/empty.png")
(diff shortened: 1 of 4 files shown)


janpio:
Remaining test failure is discussed here:
https://github.com/fastlane/fastlane/pull/11096#issuecomment-349593731
https://github.com/fastlane/fastlane/pull/11096#issuecomment-349602194

janpio:
> The reason why I ended up with .identify was that I tried to do it with MiniMagick::Tool::Identify.new, but couldn't find a way to let that act on the image variable title_image instead of a file name.

Question is if `MiniMagick::Tool::Identify.new` would also use GraphicsMagick or ImageMagick... I don't really understand how it decides that :/ Only makes sense to switch if it does _not_ use GraphicsMagick which seems to be missing as a dependency.

Interesting enough `imagemagick` also doesn't seem to be defined as an dependency:
https://github.com/fastlane/fastlane/search?utf8=%E2%9C%93&q=imagemagick&type=
Does it maybe just work by accident (as it is maybe installed on macOS by default)?

janpio:
I misinterpreted the error message:

` ImageMagick/GraphicsMagick is not installed`

means that the one of the both it tried to use is not installed.

funnel20:
> Question is if MiniMagick::Tool::Identify.new would also use GraphicsMagick or ImageMagick... I don't really understand how it decides that :/ Only makes sense to switch if it does not use GraphicsMagick which seems to be missing as a dependency.

Line 327 (existing code before my edits) uses `title_image = MiniMagick::Image.open(empty_path)` to create the image. That did pass the tests before.

funnel20:
> Interesting enough imagemagick also doesn't seem to be defined as an dependency:
> https://github.com/fastlane/fastlane/search?utf8=%E2%9C%93&q=imagemagick&type=
> Does it maybe just work by accident (as it is maybe installed on macOS by default)?

I have looked in my Terminal history. When I followed instructions from https://github.com/fastlane/fastlane/blob/master/YourFirstPR.md#checking-out-the-fastlane-repo
I needed to install Bundler and later in the local Fastlane repo run: `bundle update`
One of many output lines was:
`Using mini_magick 4.5.1`

janpio:
Yes, but mini_magick is just a wrapper around ImageMagick and GraphicsMagick. 
But this is in the docs: https://docs.fastlane.tools/getting-started/ios/screenshots/#dependencies

Still strange why it didn't fail before...

I pulled @milch into this, he is looking into it.

funnel20:
Maybe the test script is altered?

milch:
@funnel20 I found the issue and fixed the tests directly. It should work now! We'll let it run for now to see if it does. 

It works locally because we all probably have imagemagick installed on our machines, which the CI doesn't have

funnel20:
@milch Awesome, I see in your commit that the script just fakes some output. Thanks for your super fast assistance.

funnel20:
I'm glad to see that the test passed.
@milch But someone here doesn't have a CLA...

janpio:
That is something different and we can ignore - it just means more than 1 person commited and the maintainers have to check manually of it's okay to be merged. As @milch is the second person, it's all good :)

Can you remove the [WIP] from the title to indicate it is good to be reviewed?

milch (approved):
@funnel20 Great PR, thanks for your help 👍 

Sorry that you were having issues with the CLA in the previous PR. I definitely understand that the whole CLA business is frustrating for a first-time contributor. I hope now that everything is sorted out it won't discourage you from helping out on fastlane other Google-owned open source projects

To satisfy our @googlebot overlord, @funnel20 has agreed to have their code incorporated by submitting this PR and I, the only other commit author in this PR, obviously also agree to have my commit merged

milch:
@funnel20 As @janpio said we can ignore the bot in this case. We both signed the CLAs and it is obvious that I agree to having the commit I made merged as well. 

The [WIP] was just because the tests didn't pass, right? The code looks good to me so if you confirm that we'll merge

funnel20:
I have removed the WIP. It was because I have no Ruby skills and did the review together with @janpio in an earlier PR: https://github.com/fastlane/fastlane/pull/11096#pullrequestreview-80704510

janpio (approved):

funnel20:
About the CLA. My blood pressure was finally a bit down. 
**But what did just happen?**
Now there is a failing CLA check and we can just override/ignore it by commenting that we all agree.
Or worse, @milch writes this on my behalf. Sorry?
This is unbelievable, unacceptable and probably illegal.

To satisfy the googlebot I would strongly recommend to make a valid, transparent and LEGAL option to let this check pass. Legal because I personally approved for God knows what when signing my CLA.
I expect that each submitter comments a similar command as `I signed it!`, for instance` I agree this PR` and the bot does the math if everyone has done so.

The whole idea of checks is that they should pass. During my endeavour of solving this issue, I hit 3 failing checks:

1. CLA
2. Rubocop error max. amount of Class lines
3. circle failing check.

I did really have to stretch my perseverance by counting till 10 a dozen of times.
But now the last check is just ignored. It's still not green. Or should I say RED.

janpio:
Read what @googlebot writes in https://github.com/fastlane/fastlane/pull/11124#issuecomment-349645734

Background is that @googlebot is made for PRs with 1 committer, which covers 99% of PRs that come in. Only when someone with commit rights to the fastlane repo (as @milch and me for example) comes along and decide to commit to your PR, @googlebot bails out and posts this message.

It explicitly mentions that "It's up to you to confirm consent of the commit author(s) and merge this pull request when appropriate.". This is what we can now do to merge this PR: 

1. As @milch is a maintainer, I know he has a signed CLA. 
2. We know that you also have one, as @googlebot said so above.

So technically we could have merged your PR even before you had a valid CLA. But the message from @googlebot told us that this was not okay. Now he tells us, that we can use judgement and thinking to decide if we want to ignore the `x` the check provides right now.

No need to get riled up ;)

> The whole idea of checks is that they should pass.

Actually, the idea is that they check something. And depending on how they are configured, the can block a merge - or just inform about their state (and so optionally be ignored).

funnel20:
So many checks, procedures and rules.
And the last one can stay red.
Seems a bad excuses to not have a proper procedure for this 1% use-case.

frameit/lib/frameit/editor.rb
```diff
@@ -347,10 +350,64 @@ def build_title_images(max_width, max_height)
           i.interline_spacing interline_spacing if interline_spacing
           i.fill fetch_config[key.to_s]['color']
         end
-        title_image.trim # remove white space
+
+        # Trimming the image will result in the loss of the common baseline between the text in all images.
+        # title_image.trim # remove white space
```
ohayon: Should we just remove these lines instead of leaving it commented?

frameit/lib/frameit/editor.rb
```diff
@@ -347,10 +350,64 @@ def build_title_images(max_width, max_height)
...
+
+        # Hence retrieve the calculated trim bounding box:
+        # CALCULATED: trim bounding box (without actually trimming), see: http://www.imagemagick.org/script/escape.php
+        calculated_trim_box = title_image.identify { |b| b.format("%@") }
+
+        # Get vertical top offset from the trim box by converting it from string to an integer array:
+        trim_values = calculated_trim_box.split(/[\sx+]/).map(&:to_i) # The format is "wxh+X+Y", so use multiple string separators: the whitespace "\s", "x" and "+"
+
+        # Get the minimum top offset of the trim box:
+        if trim_values[3] < top_vertical_trim_offset
```
ohayon: This seems a bit magical to know that we should index in to `3` and `1` below, is there any explanation we could add here to make this a bit more clear?

frameit/lib/frameit/editor.rb
```diff
@@ -347,10 +350,64 @@ def build_title_images(max_width, max_height)
...
+        # Get matching trim box:
+        trim_box = trim_boxes[key]
+
+        # Determine the trim area by maintaining the same vertical top offset based on the smallest value from all trim boxes (top_vertical_trim_offset).
+        # The 4th parameter is the vertical top offset. When it's larger than the smallest vertical top offset, the trim box needs to be adjusted:
+        if trim_box[3] > top_vertical_trim_offset
+          # Increase the height of the trim box (2nd parameter) with the difference in vertical top offset:
+          trim_box[1] += trim_box[3] - top_vertical_trim_offset
+          # Change the vertical top offset to match that of the others:
+          trim_box[3] = top_vertical_trim_offset
```
ohayon: it seems like we use index 1 and 3 quite a bit, could we stick them into their own, more descriptive variables?

ohayon (commented):
This is a great change for sure. @funnel20 I just left a few comments that are definitely open for discussion. Nothing is too pressing and we totally understand if you just want to get this in quickly.

frameit/lib/frameit/editor.rb
funnel20: I left it in for the code review as it was the root cause for the issue.
I agree that the commented `.trim` can be removed.
I suggest to merge the comment "Trimming the image will result..." in the comment a few lines below that starts with "Hence retrieve the calculated...", since it's the rational for the crop solution.

janpio: Maybe make the comment "Naively trimming the image with `.trim` will..."?

ohayon: That's cool @funnel20, you can make that whatever you'd like, just wanted to make sure you thought it was necessary.

frameit/lib/frameit/editor.rb
funnel20: What about this extended comment:
```Ruby
# The .identify with format("%@") returns a string with "<width>x<height>+<offset_x>+<offset_y>". Extract these 4 parameters into an integer array, by using multiple string separators: the whitespace "\s", "x" and "+".
```

To also answer your next comment. Sure, we can define these indexes.
Do you have any naming convention for these kind of constants?
Example:
```Ruby
index_trim_box_width = 0
index_trim_box_height = 1
index_trim_box_offset_x = 2
index_trim_box_offset_y = 3
```

frameit/lib/frameit/editor.rb
funnel20: See previous comment

funnel20: Although..... Adding these 4 parameters would trigger the 320 lines Class limit again.
Or is there an overall Class with constants in the project?

frameit/lib/frameit/editor.rb
funnel20: @janpio 👍

funnel20: # Natively trimming the image with .trim will result in the loss of the common baseline between the text in all images. Hence retrieve the calculated trim bounding box without actually trimming. See: http://www.imagemagick.org/script/escape.php

ohayon: thats great @funnel20 !

frameit/lib/frameit/editor.rb
ohayon: I think code that is self explanatory is better than comments, I would say that you could do something like: 

```
trim_box_width = trim_values[0]
trim_box_height = trim_values[1]
trim_box_offset_x = trim_values[2]
trim_box_offset_y = trim_values[3]
```

frameit/lib/frameit/editor.rb
ohayon: Lets not worry about the 320 lines for now. Lets make the code the best we can and we'll worry about trimming it down on our end.

frameit/lib/frameit/editor.rb
funnel20: Understood.

frameit/lib/frameit/editor.rb
funnel20: Do you mind if I restore the 1 line that you changed earlier to keep within the 320 limit? We both agreed that didn't result in the best code.

ohayon: Thats cool! I plan to go in and make some changes before merging to keep the lines down. 😄

funnel20: Can I just commit my changes simultaneously?

frameit/lib/frameit/editor.rb
funnel20: Basically only need trim_box_height and trim_box_offset_y, this saves 2 lines already 👊

frameit/lib/frameit/editor.rb
ohayon: Sure!

frameit/lib/frameit/editor.rb
funnel20: Hmm, unfortunately I have to define them twice, because the trim box is used in 2 loops:
1. The loop where the image is created. To get the minimum top and maximum bottom offset of all images.
2. The loop where the trim boxes are compensated based on the earlier minimum top and maximum bottom offsets, to maintain the text baseline.

So now I have this with 2 times `# Get the height and vertical offset`:
```Ruby
        # The .identify with format("%@") returns a string with "<width>x<height>+<offset_x>+<offset_y>". Extract these 4 parameters into an integer array, by using multiple string separators: the whitespace "\s", "x" and "+":
        trim_values = calculated_trim_box.split(/[\sx+]/).map(&:to_i)
                            
        # Get the height and vertical offset:
        trim_box_height = trim_values[1]
        trim_box_offset_y = trim_values[3]

        # Get the minimum top offset of the trim box:
        if trim_box_offset_y < top_vertical_trim_offset
          top_vertical_trim_offset = trim_box_offset_y
        end

        # Get the maximum bottom offset of the trimbox, this is the top offset + height:
        if (trim_box_offset_y + trim_box_height) > bottom_vertical_trim_offset
          bottom_vertical_trim_offset = trim_box_offset_y + trim_box_height
        end

        # Store for the crop action:
        trim_boxes[key] = trim_values
      end

      # Crop images based on top_vertical_trim_offset to maintain text baseline:
      words.each do |key|
        # Get matching trim box:
        trim_box = trim_boxes[key]

        # Get the height and vertical offset:
        trim_box_height = trim_box[1]
        trim_box_offset_y = trim_box[3]
```

frameit/lib/frameit/editor.rb
janpio: I actually really meant "Naively", as in "using the most obvious function that comes to mind to do the job of trimming". But yours probably works as well ;)

funnel20: @janpio I see :-) 
It's a thinker though, as one might suspect a typo (like I did)

frameit/lib/frameit/editor.rb
ohayon: Okay. So we have a couple options. I don't want you to keep having to make changes as you've already put in a great deal of work here. I am cool to merge it as is, and we can go back in and clean it up later.

The other option would be to create some sort of Ruby object that encapsulates these box dimensions and then have an initializer where it could be created from the array of `trim_boxes` that way it could be created and used a lot more efficiently *and* have the extra benefit of being abstracted into another class and therefor maybe help with the 320 line restriction.

What are your thoughts?

funnel20: Option 2 is definitely preferred, as it improves on 3 subjects:
- readability
- 320 line check
- re-use

It would be great if you can make it, as I lack the Ruby skills. Or show me a similar file in the repo that I can use as example.

My functional thoughts:
1. The initializer uses the raw string from the .identify as input and retrieves it's 4 contents to create the trim box object with 4 populated properties: width, height, offset_x and offset_y
2. The current Class can read and change these object properties 
3. A method that returns the string format for the `.crop` method.

ohayon: Yeah that sounds awesome and we are definitely happy to help with that!

ohayon:
@funnel20 disregard that @googlebot message! Apologies for the noise with that.

funnel20:
@ohayon No problem, I think it's my mistake as SourceTree committed the earlier commit of @milch again.

funnel20:
The failing circle is expected, as I exceed the 320 lines (it's 325).
I want to fix that with the new trim box class, but first wanted to make this commit as the final solution in 1 class.

frameit/lib/frameit/editor.rb
funnel20: For me just the outline of a class with an init method, a property and 1 method that returns a string is enough. Then I can populate it all myself.

What do you prefer as Class name?
Although currently everything is titled "trim box", maybe "crop box", "crop frame" or "crop canvas" are more appropriate?

ohayon:
Sounds good. Let me know when you are in a place where you want us to take over and handle the refactor into that class! 🐰  🚀

funnel20:
As mentioned in the (collapsed) review comment:

> If you can write (or point to) the outline of a class with an init method, a property and 1 method that returns a string is enough. Then I can populate it all myself.
> 
> What do you prefer as Class name?
> Although currently everything is titled "trim box", maybe "crop box", "crop frame" or "crop canvas" are more appropriate?

I'd like to start it myself, as this code generation gives me finally some satisfaction over all administrative hassle

funnel20:
I guess this is a good candidate: https://github.com/fastlane/fastlane/blob/master/frameit/lib/frameit/screenshot.rb

ohayon:
Ah got it, sorry I missed that message! 

Yeah, that screenshot class you linked to is a perfect candidate. I was just drafting up something like this:

```
class TrimBox
  attr_accessor :width
  attr_accessor :height
  attr_accessor :x_offset
  attr_accessor :y_offset

  def initialize(idenetify_string)
    @width = ...
    
  end

  def crop
    return "some string"
  end
end
```

funnel20:
Awesome, try that tomorrow locally.
Two Ruby newbie questions:
1. In the screenshot class I sometimes see the @ before the variable, and sometimes not. What's the rule?
2. How to call the init from the Editor class?

ohayon:
Those are two great questions: 

1. the `@` indicates that its an instance variable. Also worth noting that when you have an `attr_accessor` defined with the same name, any references to `variable_name` are really just getting out the value from `@variable_name`
2. just calling `new` on a class will call it's initializer (i.e. `TrimBox.new("some string")` calls through to `TrimBox#initialize`

🚀

funnel20:
@ohayon I committed the new Trimbox class and the modified code in the Editor class.
Can you please verify it?

frameit/lib/frameit/trim_box.rb
```diff
@@ -0,0 +1,33 @@
...
+  # Represents the MiniMagick trim bounding box for cropping a text image
+  class Trimbox
+    attr_accessor :width # width of the trim box
+    attr_accessor :height # height of the trim box
+    attr_accessor :offset_x # horizontal offset from the canvas to the trim box
+    attr_accessor :offset_y # vertical offset from the canvas to the trim box
+
+    # identify_string: A string with syntax "<width>x<height>+<offset_x>+<offset_y>". This is returned by MiniMagick when using function .identify with format("%@"). It is also required for the MiniMagick .crop function.
+    def initialize(identify_string)
+      UI.user_error!("Trimbox can not be initialised with an empty 'identify_string'.") unless identify_string.length > 0
```
funnel20: I copied this logic from another Class.
When testing with faulty input, I see the Error in Terminal, but Frameit just continues.
It does not write a framed screenshot.

Is this intended behaviour?

frameit/lib/frameit/trim_box.rb
```diff
@@ -0,0 +1,33 @@
...
+      @width = trim_values[0]
+      @height = trim_values[1]
+      @offset_x = trim_values[2]
+      @offset_y = trim_values[3]
+    end
+
+    # Get the trimbox parameters in the MiniMagick string format
+    def string_format
+      # Convert trim box parameters to string with syntax: "<width>x<height>+<offset_x>+<offset_y>":
+      return "#{@width}x#{@height}+#{@offset_x}+#{@offset_y}"
```
funnel20: For readability I might guess the following is more clear:
`return "%dx%d+%d+%d" % [@width, @height, @offset_x, @offset_y]`

Do you agree?

ohayon: I think what you have here looks great already! Using `"#{variable}` is definitely the preferred way to do string interpolation. 🚀

frameit/lib/frameit/trim_box.rb
ohayon: It's possible this error is getting rescued somewhere and continuing execution. I would say leave it as is, this class looks 💯 !

ohayon (approved):

ohayon:
@funnel20 - this is awesome. Thank you so much for putting in the time and effort and dealing with some of the stress around the process, I think I speak for our whole team that we have learned a lot and will work to make this easier and a better experience for contributors. We always strive to make contributing a pleasant experience for folks. 

We would love to see more from you. Please don't hesitate to reach out to us with more questions or for help implementing new features or fixing bugs. 🚀 🐙

funnel20:
Excellent. 
Thanks to all for your patience with me and bringing me up to speed with Ruby.

PR #11144, opened 2017-12-07
Extend ErbTemplateHelper with trim_mode parameter

krin-san:
### Checklist

### Motivation and Context
Current implementation of **ErbTemplateHelper** doesn't allow to pass **trim_mode** parameter to **Erb** initializer. It makes this class impossible to use with more or less advanced erb files eliminating empty lines. There is a workaround - Erb trim_mode could be defined globally, but it's not the way I'd like to follow.

A simple test is added to check if trim_mode parameter are actually used. Other tests in the same file was slightly changed to keep file syntax consistent.

### Description
**ErbTemplateHelper** now have a new parameter - **trim_mode** - which is passed to **ErbalT** object and then passed to **Erb** initializer.

### DIFF
--- a/fastlane/lib/fastlane/erb_template_helper.rb
+++ b/fastlane/lib/fastlane/erb_template_helper.rb
@@ -13,13 +13,18 @@ module Fastlane
       File.read(template_filepath)
     end
 
-    def self.render(template, template_vars_hash)
-      Fastlane::ErbalT.new(template_vars_hash).render(template)
+    def self.render(template, template_vars_hash, trim_mode = nil)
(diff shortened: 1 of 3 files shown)


janpio:
Hi @krin-san, looks good. Totally naive question: Where is this code actually used? When would you actually use this new option?

krin-san:
@janpio, already using it on few our projects to generate email messages for release candidates. Without this trim_mode option it looks impossible to keep template file readable while preserving empty lines where they needs to be.

_p.s.: I even forgot about this PR - so old it is_

janpio (approved):
Yeah, then let's just merge this.

janpio (approved):
Yeah, then let's just merge this.

Issue #11140, opened 2017-12-07
`sh` has strange, repeating error output and unclear stacktrace at the end


janpio:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I have a lane that uses https://github.com/ionic-zone/fastlane-plugin-ionic to build an Ionic project. The plugin uses `sh` to run a CLI command. This command is failing and throwing an error. Unfortunately this causes quite strangem repeating output of said error and a stacktrace at the end that doesn't help at all.

Here is the code that is calling `sh`: 
https://github.com/ionic-zone/fastlane-plugin-ionic/blob/59f895897ea6cde2e339746b6698403fd279f5b9/lib/fastlane/plugin/ionic/actions/ionic_action.rb#L91

This is the console output I get when running my lane:

```
Virtuals-Mac:ionic-fastlane foo$ fastlane ios build_debug
...
[02:39:14]: -------------------
[02:39:14]: --- Step: ionic ---
[02:39:14]: -------------------
...
[02:40:15]: $ ionic cordova compile ios --debug --device --prod -- --packageType=development --developmentTeam=4Z338HBBX3
> cordova compile ios --packageType=development --developmentTeam=4Z338HBBX3 --debug --device
Building project: /Users/foo/ionic-fastlane/platforms/ios/FastlaneExample.xcworkspace
...
[02:40:37]: ▸ Check dependencies
Code Signing Error: No profiles for 'zone.ionic.fastlane' were found:  Xcode couldn't find any iOS App Development provisioning profiles matching 'zone.ionic.fastlane'. Automatic signing is disabled and unable to generate a profile. To enable automatic signing, pass -allowProvisioningUpdates to xcodebuild.
Code Signing Error: Code signing is required for product type 'Application' in SDK 'iOS 11.2'
** ARCHIVE FAILED **
[02:40:37]: ▸ The following build commands failed:
[02:40:37]: ▸ Check dependencies
[02:40:37]: ▸ (1 failure)
[...]

PR #11152, opened 2017-12-07
Add support for a postfix block with sh

jdee:
### Checklist

### Motivation and Context

Partly this is motivated by @janpio's [issue](https://github.com/fastlane/fastlane/issues/11140) and partly by spending quality time with Rake.

Working with Procs can be awkward. To have a multiline `error_callback`, you have choices like:

```Ruby
callback = Proc.new do |result|
  # multiline error handler
end

sh command, error_callback: callback
```

or

```Ruby
sh(command, error_callback: ->(result) { handle_error(result) }) # call a method
```

Postfix blocks are easier to use syntactically and more idiomatic. Rake supports something like this in its `sh` method. I've often thought it would be nice to allow blocks with actions, but it would require some work in the runner context. However, `sh` is not an action, so this is fairly trivial.

With this change:

```Ruby
sh command do |status, result|
  if status.success?
    UI.success "Command succeeded"
  else
    UI.user_error "Command failed with status #{status.exitstatus}"
  end
end
```

### Description

This just passes blocks from all variants of `sh` down to `Actions.sh_control_output`, which yields if `block_given?`.

This also replaces the empty fixture file in the unit tests with a more general `StringIO` that allows for mocking output.

### DIFF
--- a/fastlane/lib/fastlane/helper/sh_helper.rb
+++ b/fastlane/lib/fastlane/helper/sh_helper.rb
@@ -8,19 +8,22 @@ module Fastlane
     # When running this in tests, it will return the actual command instead of executing it
     # @param log [Boolean] should fastlane print out the executed command
     # @param error_callback [Block] a callback invoked with the command output if there is a non-zero exit status
-    def self.sh(*command, log: true, error_callback: nil)
-      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback)
+    def self.sh(*command, log: true, error_callback: nil, &b)
+      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback, &b)
     end
 
-    def self.sh_no_action(*command, log: true, error_callback: nil)
-      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback)
+    def self.sh_no_action(*command, log: true, error_callback: nil, &b)
+      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback, &b)
     end
 
     # @param command The command to be executed (variadic)
     # @param print_command [Boolean] Should we print the command that's being executed
     # @param print_command_output [Boolean] Should we print the command output during execution
     # @param error_callback [Block] A block that's called if the command exits with a non-zero status
-    def self.sh_control_output(*command, print_command: true, print_command_output: true, error_callback: nil)
+    # @yield [status, result] The return status of the command and all output from the command
+    # @yieldparam [Process::Status] status A Process::Status indicating the status of the completed command
+    # @yieldparam [String] result The complete output to stdout and stderr of the completed command
+    def self.sh_control_output(*command, print_command: true, print_command_output: true, error_callback: nil, &b)
       print_command = print_command_output = true if $troubleshoot
       # Set the encoding first, the user might have set it wrong
       previous_encoding = [Encoding.default_external, Encoding.default_internal]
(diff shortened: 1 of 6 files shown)


ohayon (commented):
Hey @jdee - is adding the `&b` param to those methods necessary? Can't ruby handle the block param for any method and it doesnt need to be explicitly listed in the method definition?

jdee:
@ohayon I don't think it's necessary in `Actions.sh_control_output`. I don't think it's passed through multiple method calls without passing it as an arg. I can review. Less is more.

jdee:
It's necessary to support blocks in all the variants, but it can be removed from `sh_control_output`. Here's a little test code to demonstrate:

```Ruby
def f1(arg)
  f2 arg
end

def f2(arg)
  yield arg if block_given?
end

f2 "foo" do |arg|
  puts "result from f2: #{arg.inspect}"
end

f1 "foo" do |arg|
  puts "result from f1: #{arg.inspect}"
end
```

```bash
[jdee@Jimmy-Dees-MacBookPro tmp]$ ruby test.rb
result from f2: "foo"
```

The block is not invoked when calling `f1`. Then if i change `f1`:

```Ruby
def f1(arg, &b)
  f2 arg, &b
end
```

```bash
[jdee@Jimmy-Dees-MacBookPro tmp]$ ruby test.rb
result from f2: "foo"
result from f1: "foo"
```

I can remove it from `sh_control_output`.

jdee:
I was also thinking of using `Forwardable` if possible to remove some of the wrappers methods here and there, like `Action.sh`.

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -62,14 +70,15 @@ def self.sh_control_output(*command, print_command: true, print_command_output:
...
-          else
+          elsif !block_given?
             UI.shell_error!(message)
           end
         end
       else
         result << shell_command # only for the tests
       end
 
+      return block_return if block_given?
```
ohayon: Will this still end up reporting the error if it skips that on line 73 if there is a block?

ohayon (approved):
This looks really awesome @jdee ! Just one question above, but might not require any changes. So go ahead and merge when ready! 🚀

jdee:
@ohayon Thanks for the approval. In the cut that you reviewed, an error message was not directly presented to the user via UI, unlike the other invocation methods. This seems best for consistency, though of course the block will get sufficient information to show the same information.

I made that change and then also moved the yield until after the error is logged, or you'll get the command output after a custom error message from the block.

Finally, I also added the constructed shell command to the yield, for ease of logging when using a gnarly Ruby expression for the argument. So now you can easily do this:

```Ruby
  sh "ls", "-la", "foobar" do |status, result, command|
    UI.error %(Command #{command} (pid #{status.pid}) failed with status #{status.exitstatus}) unless status.success?
    UI.message "Result is #{result.inspect}"
  end
```

and generate output like

```
[09:01:07]: ---------------------------
[09:01:07]: --- Step: ls -la foobar ---
[09:01:07]: ---------------------------
[09:01:07]: $ ls -la foobar
[09:01:07]: ▸ ls: foobar: No such file or directory
[09:01:07]: Exit status of command 'ls -la foobar' was 1 instead of 0.
ls: foobar: No such file or directory

[09:01:07]: Command ls -la foobar (pid 94338) failed with status 1
[09:01:07]: Result is "ls: foobar: No such file or directory\n"
```

The last two lines come from the block.

I'll merge this later, but if you have any further thoughts, please weight in.

ohayon:
No more thoughts this is 💯 !

PR #11161, opened 2017-12-09
Add option to force device frame orientation (fixes #10928)

### LINKED ISSUE #10928: Frameit: Need to turn device to left in landscape mode
### New Issue Checklist

### Issue Description

I need to turn the iPhones (especially iPhone X) to left in landscape mode instead to right. Your example screens are showing exactly the mode I need but I can't find any option to set correct orientation. Is there anything I can do?

milch:
Currently for landscape the orientation is always landscape right. This makes it possible to configure a block that gets the filename and decides for the user which orientation to use. 

By default, screenshot filenames ending in `force_landscapeleft` or `force_landscaperight` (all case-insensitive) will be forced to the specified rotation

### DIFF
--- a/frameit/lib/frameit/options.rb
+++ b/frameit/lib/frameit/options.rb
@@ -39,7 +39,18 @@ module Frameit
                            env_name: "FRAMEIT_USE_LEGACY_IPHONE_6_S",
                            is_string: false,
                            description: "Use iPhone 6s frames instead of iPhone 7 frames",
-                           default_value: false)
+                           default_value: false),
+        FastlaneCore::ConfigItem.new(key: :force_orientation_block,
+                           type: Proc,
+                           description: "[Advanced] A block to customize your screnshots' device orientation",
+                           default_value: proc do |filename|
+                             f = filename.downcase
+                             if f.end_with?("force_landscapeleft")
+                               :landscape_left
+                             elsif f.end_with?("force_landscaperight")
+                               :landscape_right
+                             end
+                           end)
       ]
     end
   end
--- a/frameit/lib/frameit/options.rb
+++ b/frameit/lib/frameit/options.rb
@@ -43,6 +43,7 @@ module Frameit
         FastlaneCore::ConfigItem.new(key: :force_orientation_block,
                            type: Proc,
                            description: "[Advanced] A block to customize your screnshots' device orientation",
+                           display_in_shell: false,
                            default_value: proc do |filename|
                              f = filename.downcase
                              if f.end_with?("force_landscapeleft")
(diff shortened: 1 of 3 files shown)


IchHabRecht:
Also the tests fail, I can confirm it works in production

milch:
Yeah, the tests are failing for an unrelated reason. @jdee was aiming to fix it in his PR #11163, and once he merges that, I will rebase it onto my branch so that the tests pass again

janpio:
#11163 has been merged and released.

milch:
@janpio Ah right, totally forgot about this, sorry. Just rebased master, will merge ASAP

frameit/lib/frameit/options.rb
```diff
@@ -39,7 +39,18 @@ def self.available_options
                            env_name: "FRAMEIT_USE_LEGACY_IPHONE_6_S",
                            is_string: false,
                            description: "Use iPhone 6s frames instead of iPhone 7 frames",
-                           default_value: false)
+                           default_value: false),
+        FastlaneCore::ConfigItem.new(key: :force_orientation_block,
+                           type: Proc,
+                           description: "[Advanced] A block to customize your screnshots' device orientation",
```
janpio: What does "a block" mean here?

milch: In ruby this construct is called a block, usually passed in as a trailing parameter:

```ruby
[1, 2, 3].each do |x|
  puts x
end
```

(the `do ... end` is the actual block)

When you want to pass it as a parameter explicitly you need to wrap it in a `Proc`, though, because you can't assign blocks to variables/parameters:

```ruby
block = proc do |x|
  puts x
end
```

janpio (approved):
Only semi understand what one would actually set as this new option, but the code looks good.

milch:
@janpio The purpose is so that you can customise what each screenshot should be oriented as. e.g.

```ruby
force_orientation_block: proc do |filename|
  case filename
    when "1Login" 
      :landscape_left
    when "2Profile"
      :landscape_right
    # and so on
  end
end
```

Reason that I made it possible to pass a custom block to decide this behaviour is that by default it'll force left landscape for filenames ending in `force_landscapeleft` and right landscape for filenames ending in `force_landscaperight`, which is kind of long and might not be the desired naming scheme people use for their apps' screenshots. I didn't want to use anything simpler or it might change behaviour for currently working setups. Some people might not want to rename their screenshots at all so they have the freedom to customise this however they want with the code they can pass in here

janpio:
Ok, I would have expected this to be an "array" (map? whatever its called in ruby) with just filename => orientation combinations. My PHP influence I guess :)

IchHabRecht:
Just wanted to share my call for frameit with the new feature:

```
frameit(silver: true, path: './fastlane/screenshots', force_orientation_block: Proc.new do |filename|
      f = filename.downcase
      if f.include?("landscape")
        :landscape_left
      end
end)
```

PR #11163, opened 2017-12-09
Fix conflict between CFPropertyList and plist in gym (was: Move require 'cfpropertylist' to limit its scope.)

jdee:
### Checklist

### Motivation and Context

A test is sporadically failing in CI because of an issue with CFPropertyList. This could also happen in other environments.

### Description

**Updated**

This has been through a number of iterations. Here's the summary:

1. Introduced `fastlane_core/core_ext/cfpropertylist` (this could move).
2. When required, it requires `cfpropertylist` and `plist` and then restores the extensions from `plist` (or adds them, if `plist` was not previously required).
3. It adds a `#to_binary_plist` method to `Array`, ~~`Enumerator`~~ and `Hash` equivalent to `#to_plist` from CFPropertyList. (CFPropertyList adds `Enumerator#to_plist`, but plist doesn't, so that is not touched.)
4. ~~This is only required inside the `xcode_preferences_dictionary` method, so these changes only occur once this method is called.~~ This didn't work as well. I think a require at the top of the source file works better. With the nested require, tests pass locally for me but fail in CI.
5. Added to the IPAFileAnalyzer to avoid a call to `plutil`.

~~This may interfere with generation of binary plists with `CFPropertyList` in some cases (doesn't happen in Fastlane). It should be possible to work around this if necessary or improve on this change.~~ The `#to_plist` method is not called anywhere in the CFPropertyList gem. It is only provided as a convenience. So this breaks nothing.

It's also possible to do this:

```Ruby
require 'fastlane_core/core_ext/cfpropertylist'
File.write path, obj.to_binary_plist
```

Without this change there is unpredictable interference between these two gems. This solution may not be the final one, but it controls the interference and makes it manageable.

The only plugin that uses `CFPropertyList` is `acknowledgements`. It only uses `#save`, not `#to_plist`, so it will not be affected.


janpio (approved):
Also solves the problem of missing `plutil` for Ubuntu/WSL 🥇

jdee:
One more thought on this: I think the CFPropertyList gem supports XML as well as binary. It might also be possible to get rid of plist in favor of CFPropertyList. But that would require a bit of work. This change might be a stopgap or more permanent.

KrauseFx:
Wow, awesome work @jdee 🎉

robertjpayne:
There is a regression in this, in `package_command_generator_xcode7.rb` line `188` there is still a `to_plist` that now fails.

jdee:
@robertpayne Thanks for reporting the problem. A fix (actually #11205) was released with 2.69.1. Hopefully that addresses the issue for you.

Issue #11027, opened 2017-11-27
Fastlane Precheck fails with 'Failed: No broken urls-> unreachable URLs in app metadata'


sheinix:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I'm running precheck as always, and since two days ago started to fail with unreachable url, for support, marketing and privacy URL's. The urls are ok, they don't redirect and they are giving a 200 OK status.

### Captured Output

Command Used: `check_metadata`
<details><summary>Output/Log</summary>

```

Successfully loaded Appfile at path '/Users/juannuvreni/Desktop/Development/Mascoteros/mascoteros-ios/Mascoteros/fastlane/Appfile'
- apple_id: 'apple@mascoteros.com'
- team_id: '2279G3HUYX'
-------

+--------------------------+----------------------+
|           Summary for precheck 2.66.2           |
+--------------------------+----------------------+
| username                 | apple@mascoteros.com |
| default_rule_level       | error                |
| include_in_app_purchases | true                 |
+--------------------------+----------------------+

Looking for iTunes Connect Team with ID 118116013

+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 3    | cocoapods                           | 20          |
| 💥   | precheck                            | 5           |
+------+-------------------------------------+-------------+



```

</details>

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.13.1                                     |
| Ruby                        | 2.4.0                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.13.6 (Apple Git-96)           |
| Installation Source         | ~/.rvm/gems/ruby-2.4.0/bin/fastlane         |
| Host                        | Mac OS X 10.13.1 (17B48)                    |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.4.0/lib                |
[...]

PR #11171, opened 2017-12-11
Added require for 'addressable'

taquitos:
Fixes issue found in https://github.com/fastlane/fastlane/issues/11027

### DIFF
--- a/precheck/lib/precheck/rules/unreachable_urls_rule.rb
+++ b/precheck/lib/precheck/rules/unreachable_urls_rule.rb
@@ -1,4 +1,5 @@
 require 'precheck/rule'
+require 'addressable'
 
 module Precheck
   class UnreachableURLRule < URLRule


KrauseFx (approved):
Always good to require more stuff

ohayon (approved):

funnel20:
@KrauseFx Aren't we going to fast here? The root cause is not clear, because the issue doesn't occur at each setup. See https://github.com/fastlane/fastlane/issues/11027#issuecomment-350546217

taquitos:
Using classes from `addressable` only works if we've loaded that requirement. Since we don't actually explicitly load it in by requiring it in the `Precheck` module, it's most likely not going to be loaded when we run that rule. 

I don't think fixing issues by `requiring` a certain dependency that we have listed in our gemspec is going too fast. We want an explicit require because it's generally accepted as the right thing to do. It prevents strange flaky behavior (like that comment points out) that can vary depending on ruby version, and environment. The key here is making sure you don't implicitly depend on something.

funnel20:
@taquitos Please don't should the messenger, I only repeat what @milch mentioned.

There was a 2nd solution by removing the dependency for `addressable`:

> Replace Addressable::URI by URI

Wouldn't that be a cleaner solution?

taquitos:
@funnel20, not shooting the messenger at all. I just wanted to clarify why that `require` seems reasonable to me. 

Replacing `Addressable::URI` seems pretty subjective to me. On one hand, URI is provided by the system, on the other, `Addressable::URI` supposedly is a better URI implementation (but probably doesn't matter for us?). I prefer to leave things as they are if they work. In this case, since we've had `Addressable` as a dependency for a while, I feel comfortable leaving it in there now, but just making sure it doesn't break in the meantime for people where the implicit reference doesn't work right.

KrauseFx:
Ah sorry for confirming so fast @funnel20, I didn't have all the context and just assumed adding a missing `require` is okay. Is that something we should revert and re-consider you think?

funnel20:
@taquitos The current code (and also your PR) uses both `Addressable::URI` and `URI` within 3 lines:
https://github.com/fastlane/fastlane/blob/91f7f5e6271dfc7e725b778529a2b95ea91dd6c2/precheck/lib/precheck/rules/unreachable_urls_rule.rb#L28

https://github.com/fastlane/fastlane/blob/91f7f5e6271dfc7e725b778529a2b95ea91dd6c2/precheck/lib/precheck/rules/unreachable_urls_rule.rb#L30

This is not consistent.

These URL prechecks did work until a few weeks ago, while the require `Addressable` was never in this file. So maybe it was defined in a higher level Fastlane file.
It might be removed for whatever reason, so precheck fails and issue #11027 is created. But maybe other issues are caused by this removal too. 
Hence I think it's important to find the root cause, and not immediately apply for a (local) solution. That's why the discussion was still pending.

funnel20:
@KrauseFx Hi Felix, I made my first Fastlane contribution last week (see #10440) and I encountered quite some proceses and procedures. Initially annoying, but of course I can understand why these are in place; to create robust and maintainable code. I'm a Ruby newbie, so luckily @ohayon helped me a lot to deliver the best solution: https://github.com/fastlane/fastlane/pull/11124#issuecomment-350068913.

I'm a bit puzzled why we are having this conversation in this PR, because this PR shouldn't exist (yet).
The discussion in issue #11027 is still pending, and several solutions were provided. Out of the blue someone who is not involved in the discussion (@taquitos) creates a PR, which is approved by you, based on assumptions and no context.

This doesn't look like a solid procedure. It seems best to put this PR on hold, and we go back to the original issue #11027 to continue the discussion. Correct me if I'm wrong.

taquitos:
@funnel20 I think we're saying the same thing: I agree with you that it's important to find the root cause, and that's totally great that the conversation is ongoing. It sounds like you and @milch made some great progress, and a more thorough solution found. We want to keep the conversation rolling and in the original place.

The conversation I was hoping to keep in this PR is just around why we decided to merge this PR. Not at all meant to stop any other conversations around the root cause. Just this: let's get something in master that works for people and let you and the other contributors continue working towards a better solution. I apologize for merging something so quickly without jumping into that thread and discussing why we felt it was important.

Now, out of the blue isn't so out of the blue 😄 I sit next to @KrauseFx in SF, and occasionally we go through all issues and prs to make fixes in order to unblock people who are crashing or to help move things along. 

Just to clarify on your point:

> Out of the blue someone who is not involved in the discussion (@taquitos) creates a PR, which is approved by you, based on assumptions and no context.

I built the precheck tool, so I have quite a bit of context 😄 In the past I've seen similar errors which were fixed by requiring the missing dependency.

PR #11177, opened 2017-12-11
Fix simulator_path

janpio:
Building the path of the simulator.app works most of the time.
Use `File.join` so it works all the time.


KrauseFx (approved):
👍

PR #11178, opened 2017-12-11
Limit testing of Timeout-Process-Bug to macOS only

janpio:
disable unkilled-process-after-timeout-tests for non-macOS platforms.
this was causing tests to fail on other platforms/environments, where this bug does not exist.
(Background: https://github.com/fastlane/fastlane/issues/1504)

### DIFF
--- a/fastlane_core/spec/project_spec.rb
+++ b/fastlane_core/spec/project_spec.rb
@@ -430,18 +430,20 @@ describe FastlaneCore do
         text = "FOOBAR" # random text
         count = count_processes(text)
         cmd = "ruby -e 'sleep 3; puts \"#{text}\"'"
-        # this doesn't work
         expect do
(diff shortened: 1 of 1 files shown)


fastlane_core/spec/project_spec.rb
```diff
@@ -430,18 +430,20 @@ def count_processes(text)
         text = "FOOBAR" # random text
         count = count_processes(text)
         cmd = "ruby -e 'sleep 3; puts \"#{text}\"'"
-        # this doesn't work
```
KrauseFx: What a great comment we had in here, haha

KrauseFx (approved):

fastlane_core/spec/project_spec.rb
janpio: it's not wrong...

PR #11180, opened 2017-12-12
Workaround hardcoded OS names in fixtures

janpio:
Some text fixtures have a hardcoded OS `macOS`. This of course makes these tests only work on this platform.

This PR overwrites the fields with the actual OS the test is running in _while_ the test is running, at runtime.

### DIFF
--- a/fastlane_core/spec/analytics/analytics_session_spec.rb
+++ b/fastlane_core/spec/analytics/analytics_session_spec.rb
@@ -186,7 +186,7 @@ describe FastlaneCore::AnalyticsSession do
           expected_final_array[2]['primary_target']['detail'] = `uname`.strip
           expected_final_array[11]['primary_target']['detail'] = `uname`.strip
         end
-        
+
         parsed_events.zip(expected_final_array).each do |parsed, fixture|
           expect(parsed).to eq(fixture)
         end
--- a/fastlane_core/spec/analytics/analytics_session_spec.rb
+++ b/fastlane_core/spec/analytics/analytics_session_spec.rb
@@ -46,6 +46,11 @@ describe FastlaneCore::AnalyticsSession do
 
         session.action_launched(launch_context: launch_context)
 
+        # Modify Fixture if not on Mac
+        unless FastlaneCore::Helper.is_mac?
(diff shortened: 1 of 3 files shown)


fastlane_core/spec/analytics/analytics_session_spec.rb
```diff
@@ -46,6 +46,11 @@
 
         session.action_launched(launch_context: launch_context)
 
+        # Modify Fixture if not on Mac
+        unless FastlaneCore::Helper.is_mac?
+          fixture_data[2]['primary_target']['detail'] = `uname`.strip
```
janpio: Is there a better way to get the correct string (`Linux` e.g.) than shelling out to `uname`?

revolter: https://stackoverflow.com/a/21082972/865175 ?

janpio: I didn't want to introduce an additional dependency because of that.

taquitos: We do have `class CPUInspector`, maybe there is something in there that could be used to help identify the correct platform, and we can hardcode platform names based on environment?

janpio: That didn't pan out, but it made me remember that the Analytics thing actually has to get the data from somewhere itself. This is it:

janpio: Alternative idea: We could also make this method always return `macOS`, not matter where we run the tests. Then the generated file will match the static fixtures. I think this should work similar to https://github.com/fastlane/fastlane/blob/563ec322458c3e1ecada354b99d893430bea7389/fastlane_core/spec/analytics/analytics_session_spec.rb#L12-L13.

KrauseFx (approved):
🍺

janpio:
Damn, turns out this method to replace a string was a bad idea: 

My branch that I used to test this on non-Mac was missing a commit that added an entry to the `.json` fixture file that is used to test here, so it now fails. This means that any test to these files, would require a change in this code as well - which is not a good idea.

I will create another PR removing this code and implementing the other method that mocks the operating system value.

PR #11182, opened 2017-12-12
Workaround for missing/delayed file after `system("openssl ...")`

janpio:
This is a curious one:

The `FileUtils.mv` can fail because `tmpfile` is not immediately available after `system("openssl ...")` returns on some systems. This causes an exception like this:

```
 1) Match Match::Encrypt encrypt
     Failure/Error: @e.encrypt_repo(path: @directory, git_url: @git_url)

     Errno::ENOENT:
       No such file or directory @ rb_file_s_rename - (/tmp/d20171210-12500-1a0jemh/temporary, /tmp/d20171210-12500-1a0jemh/randomFile.mobileprovision)
     # ./match/lib/match/encrypt.rb:106:in `crypt'
     # ./match/lib/match/encrypt.rb:45:in `block in encrypt_repo'
     # ./match/lib/match/encrypt.rb:76:in `block in iterate'
     # ./match/lib/match/encrypt.rb:74:in `each'
     # ./match/lib/match/encrypt.rb:74:in `iterate'
     # ./match/lib/match/encrypt.rb:44:in `encrypt_repo'
     # ./match/spec/encrypt_spec.rb:17:in `block (3 levels) in <top (required)>'
```

Read this StackOverflow question about how I debugged this and found out it is a timing problem:
https://stackoverflow.com/questions/47735995/fileutils-mv-fails-if-run-directly-after-systemopenssl-on-a-file-thi

### DIFF
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -93,7 +93,11 @@ module Match
       success = system(command.join(' '))
 
       UI.crash!("Error decrypting '#{path}'") unless success
-      sleep 0.1 until File.exist?(tmpfile)
+
+      # On non-Mac systems it might take some time for the file to actually be there
+      if FastlaneCore::Helper.is_mac?
+        sleep 0.1 until File.exist?(tmpfile)
+
       FileUtils.mv(tmpfile, path)
     end
   end
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -93,9 +93,7 @@ module Match
       success = system(command.join(' '))
 
       UI.crash!("Error decrypting '#{path}'") unless success
-      until File.exist?(tmpfile)
-        sleep 0.1
-      end 
+      sleep 0.1 until File.exist?(tmpfile)
       FileUtils.mv(tmpfile, path)
     end
   end
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -97,6 +97,7 @@ module Match
       # On non-Mac systems it might take some time for the file to actually be there
       unless FastlaneCore::Helper.is_mac?
         sleep 0.1 until File.exist?(tmpfile)
+      end
 
       FileUtils.mv(tmpfile, path)
     end
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -93,6 +93,9 @@ module Match
       success = system(command.join(' '))
 
       UI.crash!("Error decrypting '#{path}'") unless success
+      until File.exist?(tmpfile)
+        sleep 0.1
+      end 
       FileUtils.mv(tmpfile, path)
     end
   end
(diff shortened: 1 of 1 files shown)


match/lib/match/encrypt.rb
```diff
@@ -93,6 +93,7 @@ def crypt(path: nil, password: nil, encrypt: true)
       success = system(command.join(' '))
 
       UI.crash!("Error decrypting '#{path}'") unless success
+      sleep 0.1 until File.exist?(tmpfile)
```
taquitos: Is this something that we could do "if not mac, then sleep"?
Also, is just 0.1 enough or should it be something insanely long, like .250?

janpio: Hm, only executing it on non-Mac systems seems like a good idea. I encountered this while testing fastlane on WSL, so it probably is not happening on Mac itself.

> Also, is just 0.1 enough or should it be something insanely long, like .250?

This code should be repeated until the file exists. (Although I realize this might not be the best idea as well... how can I do this better? I know that it just takes some time for the file to be where it should be :/ )

janpio:
Note that only half of @taquitos's comment is outdated, although Github thinks otherwise.

match/lib/match/encrypt.rb
taquitos: WHoops, didn't catch the `until`
Spinning is great for stuff like this, but maybe timeout after X seconds (is 2 seconds long enough?) and throw an error? 
-Josh

janpio: I honestly have no idea how long it takes - I don't even understand _why_ this is happening in the first place.

taquitos: I was thinking something like just setting a variable `count = 0` and then looping + incrementing count up to `count == 20 || File.exist?(tmpfile)` (20 * .1  is 1 second)
So whatever happens first, either 20 loops of waiting 0.1 seconds, or if the file exists, the loop will break

janpio: I think/hope I implemented just that.

match/lib/match/encrypt.rb
```diff
@@ -93,6 +93,17 @@ def crypt(path: nil, password: nil, encrypt: true)
       success = system(command.join(' '))
 
       UI.crash!("Error decrypting '#{path}'") unless success
+
+      # On non-Mac systems it might take some time for the file to actually be there (#11182)
```
KrauseFx: Interesting, is that an issue in the `openssl` command? Is there a flag to make it a blocking call?

janpio: I have no idea tbh. Found the solution by looking at the follow up problem.Should I investigate this further instead of this workaround?

janpio: (As this is only run on non-Mac systems I would prefer to just have this merged and investigate later, maybe also switch to an openssl gem [that I think is already in use in other places])

KrauseFx (approved):
Alright, thanks @janpio, green light from my side. However before merging could you add 2-3 lines of comments to the code part describing why this is necessary and provide some context? This is to fix flakey tests on Linux based CIs?

janpio:
Of course. Did so and also made an entry in my todo list to revisit this some time later. Much later.

PR #11185, opened 2017-12-12
Option descriptions cannot include newlines

jdee:
### Checklist

### Motivation and Context

https://docs.fastlane.tools/actions/get_version_number/#parameters

Parameter listing is screwed up.

### Description

Remove newlines from the description field.

### DIFF
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -115,10 +115,10 @@ module Fastlane
                              end),
           FastlaneCore::ConfigItem.new(key: :scheme,
                              env_name: "FL_VERSION_NUMBER_SCHEME",
-                             description: "Specify a specific scheme if you have multiple per project, optional.
-                                          This parameter is deprecated and will be removed in a future release.
-                                          Please use the 'target' parameter instead. The behavior of this parameter
-                                          is currently undefined if your scheme name doesn't match your target name",
+                             description: "Specify a specific scheme if you have multiple per project, optional. " \
+                                          "This parameter is deprecated and will be removed in a future release. " \
+                                          "Please use the 'target' parameter instead. The behavior of this parameter " \
+                                          "is currently undefined if your scheme name doesn't match your target name",
                              optional: true,
                              deprecated: true),
           FastlaneCore::ConfigItem.new(key: :target,


fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -115,10 +115,10 @@ def self.available_options
                              end),
           FastlaneCore::ConfigItem.new(key: :scheme,
                              env_name: "FL_VERSION_NUMBER_SCHEME",
-                             description: "Specify a specific scheme if you have multiple per project, optional.
-                                          This parameter is deprecated and will be removed in a future release.
-                                          Please use the 'target' parameter instead. The behavior of this parameter
-                                          is currently undefined if your scheme name doesn't match your target name",
+                             description: "Specify a specific scheme if you have multiple per project, optional. " \
```
revolter: [It looks like](https://github.com/fastlane/fastlane/blob/91f7f5e6271dfc7e725b778529a2b95ea91dd6c2/fastlane/lib/fastlane/actions/swiftlint.rb#L95-L96) it should work without quotes.

jdee: The problem with that is that it leaves a bunch of spaces in the string. I think these are ignored by markdown/HTML, but you can get other weirdness.

```
+--------------------+------------------------------------------------------------------------------+---------+---------+
|                                                   swiftlint Options                                                   |
+--------------------+------------------------------------------------------------------------------+---------+---------+
| Key                | Description                                                                  | Env Var | Default |
+--------------------+------------------------------------------------------------------------------+---------+---------+
| mode               | SwiftLint mode: :lint                                                        |         | lint    |
|                    | (default) or :autocorrect;                                                   |         |         |
|                    | default is :lint                                                             |         |         |
| output_file        | Path to output SwiftLint                                                     |         |         |
|                    | result                                                                       |         |         |
| config_file        | Custom configuration file of                                                 |         |         |
|                    | SwiftLint                                                                    |         |         |
| strict             | Fail on warnings?                                                            |         | false   |
|                    | (true/false)                                                                 |         |         |
| files              | List of files to process                                                     |         |         |
| ignore_exit_status | Ignore the exit status of                                                    |         | false   |
|                    | the SwiftLint command, so                                                    |         |         |
|                    | that serious violations                                                      |         |         |
|                    | don't fail the build                                                         |         |         |
|                    | (true/false)                                                                 |         |         |
| reporter           | Choose output reporter                                                       |         |         |
| quiet              | Don't print status logs like                                                 |         | false   |
|                    | 'Linting <file>' & 'Done                                                     |         |         |
|                    | linting'                                                                     |         |         |
| executable         | Path to the `swiftlint`                                                      |         |         |
|                    | executable on your machine                                                   |         |         |
+--------------------+------------------------------------------------------------------------------+---------+---------+
```

vs

```
+-----------+-------------------------------+---------------------------+---------+
|                           get_version_number Options                            |
+-----------+-------------------------------+---------------------------+---------+
| Key       | Description                   | Env Var                   | Default |
+-----------+-------------------------------+---------------------------+---------+
| xcodeproj | optional, you must specify    | FL_VERSION_NUMBER_PROJECT |         |
|           | the path to your main Xcode   |                           |         |
|           | project if it is not in the   |                           |         |
|           | project root directory        |                           |         |
| scheme    | [DEPRECATED!] true - Specify  | FL_VERSION_NUMBER_SCHEME  |         |
|           | a specific scheme if you      |                           |         |
|           | have multiple per project,    |                           |         |
|           | optional. This parameter is   |                           |         |
|           | deprecated and will be        |                           |         |
|           | removed in a future release.  |                           |         |
|           | Please use the 'target'       |                           |         |
|           | parameter instead. The        |                           |         |
|           | behavior of this parameter    |                           |         |
|           | is currently undefined if     |                           |         |
|           | your scheme name doesn't      |                           |         |
|           | match your target name        |                           |         |
| target    | Specify a specific target if  | FL_VERSION_NUMBER_TARGET  |         |
|           | you have multiple per         |                           |         |
|           | project, optional             |                           |         |
+-----------+-------------------------------+---------------------------+---------+
```

I don't think it matters very much, but I think this is better than the other way. It would probably be best if the doc tools (`fastlane action` and the markdown generator, at least) just did something like `description.gsub(/\n/, '')`. For now, this just fixes one page.

revolter: So should the `swiftlint.rb` be edited too?

jdee: Arguably it should, since the value is not quite right.

```
2.4.2 :006 > Fastlane::Actions::SwiftlintAction.available_options.find { |o| o.key == :ignore_exit_status }.description.gsub(/\s/, '.')
 => "Ignore.the.exit.status.of.the.SwiftLint.command,.so.that.serious.violations.....................................................don't.fail.the.build.(true/false)" 
2.4.2 :007 > Fastlane::Actions::GetVersionNumberAction.available_options.find { |o| o.key == :scheme }.description.gsub(/\s/, '.')
 => "[DEPRECATED!].true.-.Specify.a.specific.scheme.if.you.have.multiple.per.project,.optional..This.parameter.is.deprecated.and.will.be.removed.in.a.future.release..Please.use.the.'target'.parameter.instead..The.behavior.of.this.parameter.is.currently.undefined.if.your.scheme.name.doesn't.match.your.target.name" 
```

But there's nothing fundamentally wrong with the documentation for `swiftlint`. The docs page for `get_version_number` is confusing because of the formatting problem. This was just intended as a quick, targeted fix for that page before my work day starts, and I have no more time for Fastlane.

Make it work. Make it right. Make it fast.

This just makes it work in hopes of fixing that page. I think we can make it right later, probably by adjusting the docs code instead of hunting down little issues in a huge number of actions.

revolter (approved):

PR #11188, opened 2017-12-12
`openssl` exit status is "wrong" on non-macOS platforms, fix code to work anyway

janpio:
The old code 
```
UI.crash!("Error decrypting '#{path}'") unless success
``` 
depends on the `command` (which is `openssl ...` in this case) to correctly return an "exit status" to initate the expected crash when decrypting fails:

> system returns true if the command gives zero exit status, false for non zero exit status. Returns nil if command execution fails.
https://ruby-doc.org/core-2.4.0/Kernel.html#method-i-system

For the macOS implementation of `openssl` this seems to be true. 
Unfortunately, it is not for the Ubuntu/WSL implementation where even if decryption fails, the exit status is `true` (but an error output is generated, which `system` didn't surface.).

Using `Open3.capture3` enables us to look at or for that error output and "fix" `success` to `false` if something is returned, which indicates a failure.

(First line of PR also fixes a tiny typo)

### DIFF
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -91,10 +91,12 @@ module Match
       command << "-d" unless encrypt
       command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show show an error message if something goes wrong
 
-      _out, err, _st = Open3.capture3(command.join(' '))
-      if err.to_s == ''
-        success = true
-      else
+      _out, err, st = Open3.capture3(command.join(' '))
+      success = st.success?
+
+      # Ubuntu `openssl` does not fail on failure
+      # but at least outputs an error message
+      if err.to_s != ''
         success = false
       end
 
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -89,8 +89,14 @@ module Match
       command << "-out #{tmpfile.shellescape}"
       command << "-a"
       command << "-d" unless encrypt
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show show an error message is something goes wrong
-      success = system(command.join(' '))
+      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show show an error message if something goes wrong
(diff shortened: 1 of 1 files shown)


janpio:
Alternatively I could wrap this new code in a conditional `is_mac` and leave the Mac code path unchanged.

match/lib/match/encrypt.rb
```diff
@@ -89,8 +89,16 @@ def crypt(path: nil, password: nil, encrypt: true)
       command << "-out #{tmpfile.shellescape}"
       command << "-a"
       command << "-d" unless encrypt
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show show an error message is something goes wrong
-      success = system(command.join(' '))
+      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show show an error message if something goes wrong
```
revolter: `show` is also duplicated, if you're at it.

revolter (changes requested):

revolter (approved):

match/lib/match/encrypt.rb
```diff
@@ -89,8 +89,16 @@ def crypt(path: nil, password: nil, encrypt: true)
...
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show show an error message is something goes wrong
-      success = system(command.join(' '))
+      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show an error message if something goes wrong
+
+      _out, err, st = Open3.capture3(command.join(' '))
+      success = st.success?
+
+      # Ubuntu `openssl` does not fail on failure
+      # but at least outputs an error message
+      if err.to_s != ''
```
giginet: you should use `unless err.ro_s.empty?`

giginet (approved):

match/lib/match/encrypt.rb
janpio: Thanks, I'm pretty noob in ruby. Changed it.

Issue #11206, opened 2017-12-13
Rename `fastlane_version` to `min_fastlane_version`


janpio:
`fastlane_version` ...

> [v]erifies the minimum fastlane version required

https://docs.fastlane.tools/actions/fastlane_version/

If you don't know anything about fastlane, you right now can misunderstand `fastlane_version "1.23.4"` to mean that your fastlane setup will use version 1.23.4, which is not true.

Example: https://github.com/fastlane/fastlane/issues/6006#issuecomment-245193151

Renaming the action would solve this problem.

TODO:
- Copy action to new naming
- Provide old naming as alias for backwards compatibility
- Change `Fastfile` template(s) to use new naming

revolter:
Another problem I found with that is that it looked 100% like a variable, instead of a function/action that does something, so adding a verb would be great! Examples: `require_fastlane_version`, `require_min_fastlane_version`.

jdee:
This could presumably be done using aliases, making the `fastlane_version` action an alias for `min_fastlane_version` the way `gym` is now an alias for `build_app`. Because I'm sure @mpirri needs something to do.

milch:
I agree with @revolter, I think `require_fastlane_version` would make the most sense

jdee:
That sounds reasonable. It's fairly trivial to change the names. I'll leave the PR as is until a consensus is reached.

janpio:
`require_fastlane_version` has the exact same problem as only `fastlane_version` - it can be misunderstood to require that exact version.

If you want the `require` in there it would have to be `require_min_fastlane_version` or `require_fastlane_min_version`.

milch:
Some more bike shedding: `min_fastlane_version_required` then?

janpio:
`required` is probably better than `require` as it matches better to the "passive" nature of the already present installation of fastlane vs. the active act of "requiring" something right now.

(I personally would leave it off in my own code, but if majority agrees I don't object of course)

jdee:
I guess one issue is the general disconnect when mobile app devs use a Ruby tool and aren't familiar with general dependency management using a Gemfile and the Bundler. It would be easy to think that something like this would work:

```Ruby
required_fastlane_version "~> 2.69"
```

[...]

PR #11209, opened 2017-12-13
Added min_fastlane_version action and fastlane_version alias.

jdee:
### Checklist

### Motivation and Context

Address #11206.

### Description

Renamed `fastlane_version` to `min_fastlane_version`. Added `fastlane_version` as an alias.

**Updated:**

~~Added a `settings` method to the Fastfile DSL that allows setting min_fastlane_version and default_platform as attrs on an object, e.g.:~~

This is moving to a new PR. This PR only adds min_fastlane_version and makes
fastlane_version an alias. It also updates the default Ruby Fastfile templates.

```Ruby
settings do |s|
  s.min_fastlane_version = "2.69.3"
  s.default_platform = :ios
end

lane :do_something do
  # something
end
```

Or

```Ruby
settings.min_fastlane_version = "2.69.3"
settings.default_platform = :ios

lane :do_something do # etc.
```

This does not remove support for the `fastlane_version` or `min_fastlane_version`. It only provides a new mechanism for passing single-parameter settings to actions in a way that may be more natural and idiomatic. This is one of the concerns in the linked issue, and it jibes with the Swift approach. In a Swift Fastfile you can just:

```Swift
let minFastlaneVersion = "2.69.3"
```

(Though in all the templates it's a computed property instead of a let.)

It's not hard to support

```Ruby
self.min_fastlane_version = "2.69.3"
```

But it can't be done without the self because of a Ruby scoping issue. That seems out of place here. I like the idea of using a method that yields or returns an object with setters. I'm open to suggestions on what the name should be.


janpio (approved):
🥇

KrauseFx (approved):
📯

nafu (approved):
💛

jdee:
Thanks for all the approvals. I spent about five minutes on this and would like to review it a bit before merging, to see what I might have missed in tests, docs, etc. There has been some discussion of what the new name should be as well, so I've been dragging my feet until that is resolved. I should be able to finish this off this evening (PST).

jdee:
@KrauseFx, @nafu This requires a re-review.

I think @revolter and @janpio may like the new Fastfile syntax I introduced.

I'll deal with the Swift end of the method-name change as soon as I can touch base with @taquitos.

revolter:
Great way to bundle them together!

jdee:
I think there's a bigger win here than just the syntax. As @revolter said, these things are now bundled together, which is to say they have been relegated to a certain scope or namespace. Fastlane's current approach simply does not scale. A flat namespace for method names guarantees eventual collisions. The Actions module, which is a reflection of that same namespace, is bursting at the seams. There are 207 source files in *one folder* in the repo (fastlane/lib/fastlane/actions). *And* the same module/namespace is extended by 233 plugins. These numbers are obviously growing (231 plugins a couple days ago) and there's no reason to expect that growth to slow.

This is an alternative approach for seven actions. These actions can be deprecated and ultimately removed from the global namespace by adding a mechanism for extending the base Fastlane DSL other than the simple action model. And there's no reason this settings mechanism can't be extended in a way similar to actions.

I propose (in a separate PR yet to come) to add a new `Fastlane::Setting` base class and allow settings to be defined something like this:

```Ruby
module Fastlane
  module Settings
    class MinFastlaneVersionSetting < Setting
      class << self
        def apply(value)
          # This one is all validation
          # The team_id setting will set env. vars. here
        end

        def valid?(value)
          valid = Gem::Version.new(Fastlane::VERSION) >= Gem::Version.new(value)
          UI.error "Current Fastlane version does not meet specified minimum"
          valid
        end

        def value_type
          String
        end

        def description
          "Set the minimum Fastlane version"
        end
      end
    end
  end
end
```

Then plugins can also add their own settings. _And_ they can be scoped to the plugin, so you could have something in your Fastfile like:

```Ruby
settings do |s|
  s.minimum_fastlane_version = "2.69.3"
  s.my_plugin do |p|
    p.plugin_setting = true
  end

  s.another_plugin do |p|
    # Look, Ma, no conflicts!
    p.plugin_setting = true
  end
end
```

And then I think we'll also want a general way to externalize all settings in one or more config files.

I'll take a crack at this idea soon.

milch:
I looked through the code, and I think your settings-scoping thing is a good idea, but I think at this point we should split up the PR. Then we can merge the `min_fastlane_version` support first and have the settings + plugin settings support separate from that

milch:
Or, maybe instead of splitting these it might be a good idea to invert the flow: instead of having the `FastlaneConfig` call the `Action` have the `Action` set the config value in `FastlaneConfig`. The verification could be done by giving a block or something. Something like this: 

```ruby
# Somewhere in fastlane
FastlaneConfig.add_setting("minimum_fastlane_version",
  scope: Fastlane,
  description: "Verifies the minimum fastlane version required",
  verify_block: Proc.new do |version|
    # ... do the verification 
  end
) 

# Somewhere in my plugin 
FastlaneConfig.add_setting("plugin_setting",
  scope: MyPlugin,
  description: "Makes foo behave like bar",
  verify_block: Proc.new do |value|
    UIvalue.end_with? "png"
  end
) 

# In fastlane/lib/fastlane/actions/min_fastlane_version.rb 
# Replaces the method in Fastlane::Actions::MinFastlaneVersionAction
def self.run(params)
  FastlaneConfig.config.set_value("minimum_fastlane_version", scope: Fastlane, value: params[:value])
end

# In my Fastfile (same as @jdee's example)
settings do |s|
  s.minimum_fastlane_version = "2.69.3"
  s.my_plugin do |p|
    p.plugin_setting = "bla.png"
  end
end
```

jdee:
I was thinking of splitting the PR anyway, since there's a clean break a few commits back, and that part was basically already approved by everyone. There is the question of Swift and updating the default Swift Fastfile to use `minFastlaneVersion` by default. I was digging through the generator last night but couldn't quite find what I was looking for. But the Swift stuff will still work with the alias change, and I can work with @taquitos to deal with that separately. I'll separate them and open a new PR, and we can work through the settings implementation issues. I like the idea of integrating this with FastlaneCore::Configuration, but I'll save those thoughts for a new PR.

taquitos:
I ❤️ ❤️ ❤️ this idea. Swift support can come later since we don't actually enforce anything right now.

Issue #11197, opened 2017-12-12
Pass app identifier to match function in Fastfile


eman6576:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
 When calling `match` in a Fastfile, the parameter type for `app_identifier` has changed to `Array`. I tried passing in something like this `Array[fabric_bundle_identifier]`, but I still get `'app_identifier' value must be a Array! Found String instead.`. What is the correct value to pass to the parameter?

### Captured Output

Command Used: `update_test_devices`
<details><summary>Output/Log</summary>

```

Successfully loaded Appfile at path '/Users/emanuel/Desktop/SilverLogic/Bitbucket/iOS/baseapp-ios-v2/fastlane/Appfile'
- app_identifier: 'com.silverlogic.BaseAppV2'
- apple_id: 'apple@tsl.io'
- team_id: '4SB3F2S2K3'
-------

+----------------+----------------------------------------------------------------------------------------+
|                               Detected Values from './fastlane/Matchfile'                               |
+----------------+----------------------------------------------------------------------------------------+
| git_url        | ssh://git@bitbucket.org/silverlogic/baseapp-ios-provisioning-profiles-certificates.git |
| app_identifier | com.silverlogic.BaseAppV2                                                              |
| username       | apple@tsl.io                                                                           |
+----------------+----------------------------------------------------------------------------------------+


+------+-------------------------------------+-------------+
|                     fastlane summary                     |
+------+-------------------------------------+-------------+
| Step | Action                              | Time (in s) |
+------+-------------------------------------+-------------+
| 1    | Verifying required fastlane version | 0           |
| 2    | default_platform                    | 0           |
| 3    | register_devices                    | 3           |
| 💥   | match                               | 0           |
| 5    | slack                               | 0           |
+------+-------------------------------------+-------------+



```

</details>

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.6                                     |
| Ruby                        | 2.3.1                                       |
| Bundler?                    | false                                       |
[...]

PR #11212, opened 2017-12-13
Fix match requiring an Array for app_identifier

taquitos:
- match accepts both, but type: Array is defined in options
- disable type checking until we have a better solution
fix for https://github.com/fastlane/fastlane/issues/11197


KrauseFx (approved):
🏣

PR #11215, opened 2017-12-13
Xcode is not required for all actions

icecrystal23:
### Checklist

### Motivation and Context

We use "sigh" by itself; it does not require Xcode.  A new version of fastlane was failing with the following, which was annoying since this command does not require Xcode.
```
sh: /Library/Developer/CommandLineTools//usr/bin/xcodebuild: No such file or directory
ERROR [2017-12-13 12:22:40.21]: undefined method `split' for nil:NilClass
sh: /Library/Developer/CommandLineTools//usr/bin/xcodebuild: No such file or directory
ERROR [2017-12-13 12:22:40.24]: undefined method `split' for nil:NilClass
/Users/jenkins/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.69.2/fastlane_core/lib/fastlane_core/ui/interface.rb:131:in `user_error!': [!] Error detecting currently used Xcode installation, please ensure that you have Xcode installed and set it using `sudo xcode-select -s [path]` (FastlaneCore::Interface::FastlaneError)
```

### Description

Return nil as the xcode_version if usr/bin/xcodebuild does not exist at the xcode-select path

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -142,6 +142,7 @@ module FastlaneCore
     def self.xcode_version
       return nil unless self.is_mac?
       return @xcode_version if @xcode_version && @developer_dir == ENV['DEVELOPER_DIR']
+      return nil unless File.exist?("#{xcode_path}/usr/bin/xcodebuild")
 
       begin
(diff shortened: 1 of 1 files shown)


icecrystal23:
I signed the CLA and corrected my commit to use the same email address.

taquitos (approved):
🎈:goat:
Thank you, great catch. Super annoying to crash in this case!

PR #11216, opened 2017-12-14
Enable to specify the tag to push to remote

chuganzy:
### Checklist

### Motivation and Context

Want to push a specific tag.

### Description

Add an option to `push_git_tags` to specify tag to push to remote.

### DIFF
--- a/fastlane/lib/fastlane/actions/push_git_tags.rb
+++ b/fastlane/lib/fastlane/actions/push_git_tags.rb
@@ -5,14 +5,12 @@ module Fastlane
         command = [
           'git',
           'push',
-          '--tags'
+          params[:remote],
+          params[:tag] || '--tags'
         ]
 
         # optionally add the force component
         command << '--force' if params[:force]
-
-        # optionally add the remote component
-        command << params[:remote] if params[:remote]
 
         result = Actions.sh(command.join(' '))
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/push_git_tags.rb
```diff
@@ -5,15 +5,13 @@ def self.run(params)
         command = [
           'git',
           'push',
-          '--tags'
+          params[:remote],
```
KrauseFx: Is it an issue that this is potentially `nil`? Until now we only added the `remote` to the array if there is actually a value

chuganzy: I just followed `push_to_git_remote`'s spec to make this logic simpler, but totally fine with keeping it optional (=not providing the default value) if you want!

https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/push_to_git_remote.rb#L61

taquitos: I think somebody could set `params[:remote]` to nil, and that would override the default, but that being said, if you do that, you deserve the crash 😆

taquitos (approved):
🎈:goat:

PR #11222, opened 2017-12-14
Add missing `highline` import in spaceship library

KrauseFx:
Fixes https://github.com/KrauseFx/xcode-install/issues/249

### DIFF
--- a/spaceship/lib/spaceship.rb
+++ b/spaceship/lib/spaceship.rb
@@ -4,6 +4,9 @@ require 'spaceship/client'
 require 'spaceship/provider'
 require 'spaceship/launcher'
 
+# For basic user inputs
+require 'highline/import'
+
 # Dev Portal
 require 'spaceship/portal/portal'
 require 'spaceship/portal/spaceship'


jdee (approved):
:shipit:

jdee:
@KrauseFx I just had a thought. I'm sure you tested this, but is it actually `highline/import` that you need?

jdee:
And testing may not solve the problem, since commander is probably bringing in `highline/import`. I suspect that you do need `highline/import` or you'll have to `include` a module where `ask` is used.

taquitos (approved):
🎈:goat: always good to require more stuff

KrauseFx:
@jdee I didn't do a lot of testing, but in this case it seems like when someone requires spaceship, but not highline, it will fail, as that's where the `ask` method is implemented

jdee:
I guess this is what I mean:

```
2.4.2 :001 > require 'highline'
 => true 
2.4.2 :002 > answer = ask "question"
NoMethodError: undefined method `ask' for main:Object
	from (irb):2
	from /Users/jdee/.rvm/rubies/ruby-2.4.2/bin/irb:11:in `<main>'
2.4.2 :003 > require 'highline/import'
 => true 
2.4.2 :004 > answer = ask "question"
question
42
 => "42" 
```

I don't think require 'highline' is enough to bring in the method in spaceship. But I'd have to do some testing to prove it.

jdee:
Wait, you already had `highline/import`? I guess I'm just blind. Don't mind me.

KrauseFx:
So the PR looks good @jdee? :)

PR #11237, opened 2017-12-15
Settings method

jdee:
### Checklist

### Motivation and Context

See #11206 #11209 

Ruby syntax can be confusing for Fastlane users. Using parentheses with an action such as `min_fastlane_version("2.69.3")` makes things a little clearer. However, there are a number of actions, like `default_platform` and `team_id` that amount to setting a parameter. Even in Ruby this is more common:

```Ruby
Gem::Specification.new do |s|
  s.name = "fastlane" # not s.lane "fastlane"
```

Using methods in a DSL is common, e.g. in config files like Gymfile: `project "MyProject.xcodeproj"`. it's actually not possible because of a Ruby scoping issue to allow that to become `project = "MyProject.xcodeproj"`. Hence the approach taken in gemspecs and a lot of other Ruby configuration.

This change introduces a `settings` method that returns or yields and object that will accept setter names for the following actions:

- default_platform
- fastlane_version
- min_fastlane_version
- ruby_version
- team_id
- team_name

For example:

```Ruby
settings.min_fastlane_version = "2.69.3"
```

or

```Ruby
settings do |s|
  s.min_fastlane_version = "2.69.3"
  s.default_platform = :ios
  s.ruby_version = "2.4.2"
  s.team_id = "ABCDEFG"
end
```

This is also a first step toward scoping certain actions to avoid conflicts in a global namespace.

For previous discussion and context see particularly:

- https://github.com/fastlane/fastlane/pull/11209#issuecomment-352011828
- https://github.com/fastlane/fastlane/pull/11209#issuecomment-352026011

### Description

Added `settings` method to the DSL in `fast_file.rb`. This just creates and yields or returns a `Fastlane::Config` object, which intercepts setters in `#method_missing` and maps them to action calls. It also converts the default Ruby Fastfile templates to use this syntax.

There will be some documentation effort to indicate this alternate usage for each of these actions.

There are plenty of questions around the implementation and extensibility to be worked out as well.


jdee:
Another thought here is that this could be extended to include all `available_options` for any action. Here's something I'd like to improve:

https://github.com/jdee/settings-bundle/blob/master/fastlane/Fastfile#L40

I don't really think I want to add support for a `SettingsBundlefile`. I could have a .env file. But it might be nice to be able to do this in some cases:

```Ruby
settings do |s|
  s.update_settings_bundle.xcodeproj = "examples/SettingsBundleLibraryExample/SettingsBundleLibraryExample.xcodeproj"
end
```

jdee:
This [gist](https://gist.github.com/jdee/70616a91e9a5d62754a7adcad5716f9b) offers a possible Swift API. In both cases, an action/method style call will be supported as well, so in Ruby you could have:

```Ruby
settings({
  min_fastlane_version: "2.69.3",
  team_id: "ABCDEFG",
  my_plugin: {
    plugin_setting: true
  }
})
```

janpio:
I am closing this now as it is conflicted, and there is quite a lot of work needed to ever get this merged. But I added it to a Milestone "Far future Fastlane" so it doesn't get forgotten and lost. Some day hopefully will take the time to look at this and decide to finish it.

PR #11242, opened 2017-12-16
Circle CI: Run in Ubuntu and macOS environments, in parallel

janpio:
This PR takes the current Circle CI configuration for the macOS test run and creates a copy for Ubuntu, makes all the necessary changes (some commands have to be replaced, a dependency that is always present on macOS has to be manually installed, the caches need better names) and then adds a "workflow" that makes both test, macOS and Ubuntu, run at the same time.


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

janpio:
Success on first try:

image

👍

janpio:
Although Github UI still shows this:
image
Bug or did I miss something?

janpio:
Found something, seems to be a limitation of Github/Circle integration:
https://circleci.com/docs/2.0/workflows/#workflows-waiting-for-status-in-github
We can change that when this PR is merged.

KrauseFx:
Wow 🚀

janpio:
PR is now ready for review.

KrauseFx:
screenshot 2017-12-19 08 56 43

Looks like CI is green, but the overall status is missing or something?

KrauseFx:
This looks awesome, and I'm really excited to have Linux CI running also. Since this is a very Circle heavy PR, I'd love to get @DanToml's input on this before we merge. I know she's on vacation right now, so if it's cool from your side, let's wait for her to come back and get her input?

janpio:
Absolutely, looking forward to it - this was my first contact with Circle CI.

Further work that benefits from Ubuntu or cross platform testing can just use this branch as base instead of `master`, so no harm done.

janpio:
Oh, missed your second comment with the screenshot: Yes, that what I talked about in https://github.com/fastlane/fastlane/pull/11242#issuecomment-352146780 - you (or some other admin) would have to change a setting here in Github for this to be correct.

KrauseFx:
Yeah, seems like it's something we should wait for until after the merge

janpio:
Yes, as we know how to solve it (and CircleCI is aware and has this explicit instructions) I don't think this will become a problem to solve, we just shouldn't forget to do it when we merged this.

.circleci/config.yml
```diff
@@ -23,7 +23,61 @@ jobs:
...
+            gem update --system
+            bundle check || bundle install --jobs=4 --retry=3 --path .bundle
+            sudo apt-get install shellcheck
+      - run:
+          name: Install additional dependency xar
+          command: |
+            cd /tmp
+            mkdir download
+            cd download
+            wget https://github.com/downloads/mackyle/xar/xar-1.6.1.tar.gz
```
lacostej: We might want to provide a pre-built binary to avoid having to do that every time and maybe host it here: https://launchpad.net/ubuntu/+ppas

You could of course host it on github first.

.circleci/config.yml
```diff
@@ -23,7 +23,61 @@ jobs:
...
+            tar -xzf xar-1.6.1.tar.gz
+            cd xar-1.6.1
+            ./autogen.sh --noconfigure
+            ./configure
+            make
+            sudo make install
+      - save_cache:
+          key: fastlane-ubuntu-gems-{{ checksum "Gemfile" }}
```
lacostej: does /Users/distiller work on Linux? (store_test_results section)

janpio: afaik it doesn't even work on macOS - or I couldn't find out how it should work, what it should do etc. Do you have more insight?

.circleci/config.yml
janpio: That sounds like a good idea. I will look into that.

janpio: Fascinating, there are actually 4 sources for ready built `xar`s there I didn't find googling before:
https://launchpad.net/ubuntu/+ppas?name_filter=xar
Unfortunately none of them are v1.6.1, so I will see if I can build this myself.

janpio: https://launchpad.net/ubuntu/+ppas links to https://help.launchpad.net/Packaging/PPA links to https://help.launchpad.net/Packaging/PPA/Uploading links to https://help.launchpad.net/Packaging/PPA/BuildingASourcePackage links to http://packaging.ubuntu.com/html/ links to http://packaging.ubuntu.com/html/packaging-new-software.html - and is a bit complicated.

I think I will move this into a new issue and work on it independently from this PR. 

Do you think deploying it like this would be fine for the beginning @lacostej ?

janpio: https://github.com/fastlane/fastlane/issues/11280 now exists.

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 
 jobs:
-  build:
+  "macOS":
```
DanToml: no need to wrap these in quotes

.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: v1-gems-{{ checksum "Gemfile" }}
+          key: fastlane-macos-gems-{{ checksum "Gemfile" }}
```
DanToml: `v1-macos-gems-{{ checksum "Gemfile" }}` would be a better cache key here, because we need to be able to invalidate the cache key easily

.circleci/config.yml
DanToml: `v1-ubuntu-gems-{{ checksum "Gemfile" }}` would be a better cache key here, because we need to be able to invalidate the cache key easily

.circleci/config.yml
DanToml: We should build a docker container that includes these (based on the circleci/ruby:2.3 container)

.circleci/config.yml
DanToml: /Users/distiller will not work on linux

.circleci/config.yml
```diff
@@ -23,7 +23,61 @@ jobs:
...
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: fastlane-ubuntu-gems-{{ checksum "Gemfile" }}
+      - run:
+          name: Setup Build
+          command: |
+            mkdir -p $CIRCLE_TEST_RESULTS
+            echo "2.3" > .ruby-version
```
DanToml: No need to write .ruby-version on linux

.circleci/config.yml
```diff
@@ -23,7 +23,61 @@ jobs:
...
+            gem update --system
```
DanToml: No need to update system rubygems on linux either, the circle image builds nightly

.circleci/config.yml
```diff
@@ -23,7 +23,61 @@ jobs:
...
+            bundle check || bundle install --jobs=4 --retry=3 --path .bundle
+            sudo apt-get install shellcheck
```
DanToml: Another candidate for the build docker container.
Ideally we should seperate shellcheck linting out and run it in its own job.

.circleci/config.yml
```diff
@@ -23,7 +23,61 @@ jobs:
...
+  
+  "Ubuntu":
+    environment:
+      CIRCLE_TEST_RESULTS: $HOME/test-results
+      LC_ALL: C.UTF-8
+      LANG: C.UTF-8
+      FASTLANE_ITUNES_TRANSPORTER_PATH: .bundle
+    docker:
+      - image: circleci/ruby:2.3
+    shell: /bin/bash --login -eo pipefail
```
DanToml: Probably don't need a login shell on linux?

endocrimes (changes requested):
This is a good start!

.circleci/config.yml
janpio: I wanted to indicate that these are chosen strings vs. stuff that from a list Circle expects. Bad idea?

.circleci/config.yml
janpio: Will look into what this actually does/means. Stupid copy/paste from macOS for now.

.circleci/config.yml
janpio: Got it.

.circleci/config.yml
janpio: Got it.

.circleci/config.yml
janpio: `v1` refers to what here? Is this v1 of your cache file? I was unclear if this references Circle CI 1.0 maybe somehow.

KrauseFx:
Thanks for your help here @DanToml ❤️

.circleci/config.yml
janpio: (Why do we need this on macOS?)

janpio: Read up on it:
Don't think we need that.

https://circleci.com/docs/2.0/configuration-reference/#default-shell-options


(Why do we need this on macOS?)

janpio: Read up on it:
Don't think we need that.

https://circleci.com/docs/2.0/configuration-reference/#default-shell-options

(Why do we need this on macOS?)

janpio: Read up on it: https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html#Bash-Startup-Files
Don't think we need that.

(Why do we need this on macOS?)

When `--login` is gone, all that is left is actually the default: https://circleci.com/docs/2.0/configuration-reference/#default-shell-options
So this line can be completely removed 👍

.circleci/config.yml
janpio: Found it: https://circleci.com/docs/2.0/caching/#clearing-cache

.circleci/config.yml
janpio: Ok, looked into this a bit:

This step is currently failing on our normal builds:
https://circleci.com/gh/fastlane/fastlane/12636
```
Unable to save test results from /Users/distiller/test-results
Error stat /Users/distiller/test-results: no such file or directory
Found no path with test results, skipping
```

But test results should actually be written to somewhere: https://github.com/fastlane/fastlane/blob/3830c51f153875ec0b7350af359709e32616cfab/Rakefile#L18

I will create an issue for this first, when it is fixed the result can be incorporated here.

janpio: This is now #11303 and PR #11304

.circleci/config.yml
janpio: See #11302

.circleci/config.yml
janpio: See #11302

.circleci/config.yml
janpio: PR #11304 is ready for review (and merge). It fixes this step and will probably work unchanged on Ubuntu as well, so can be copied over.

.circleci/config.yml
```diff
@@ -23,10 +23,62 @@ jobs:
...
+            - .bundle
+
+      - run:
+          name: Check PR Metadata
+          command: bundle exec danger || echo "danger failed"
+
+      - run: bundle exec fastlane test
+
+      - store_test_results:
+          path: /Users/distiller/test-results
```
janpio: replace with result from #11302

janpio:
Update:
- Waiting for review and merge of #11302 
- Dependency stuff that should be solved via a Docker container was "outsourced" to #11302

Besides that I hope I covered all comments.

janpio:
#11302 is merged, I will incorporate this here now.

janpio:
Rebased, and changes from #11304 adapted for the second half of the file. Circle CI now correctly collects the test reports of rspec and display them:

> Your build ran 4541 tests in RSpec with 0 failures
Slowest test: fastlane_core.spec.build_watcher_spec FastlaneCore::BuildWatcher.wait_for_build_processing_to_be_complete returns specified build when strict watch is enabled (took 10.00 seconds).

I think this is now "good" again and all relevant, outstanding feedback was addressed. Can I get another look at it please?

(As said before, I think the dependency and Docker stuff will be handled in a follow up issue: #11302)

.circleci/config.yml
DanToml: They're both the same when they get parsed anyway :woman_shrugging:, as much as I wish YAML required quotes everywhere, stylistically, I'm not a fan of them when they're not required, as it makes me assume something weird is going on.

.circleci/config.yml
DanToml: Yeah - We don't have a good way to clear caches through the UI/API yet, so the hack is to bump a version string in the yaml

.circleci/config.yml
DanToml: We need it on macOS jobs so that we source the `chruby` init scripts, to use our chosen ruby, and not system ruby

.circleci/config.yml
DanToml: :+1:

.circleci/config.yml
```diff
@@ -1,45 +1,107 @@
...
           name: Post Test Results to GitHub
           command: bundle exec danger || echo "danger failed"
           when: always # Run this even when tests fail
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - "macOS"
+      - "Ubuntu"
```
DanToml: Add a newline at the end and this is good to go

endocrimes (approved):

DanToml:
@KrauseFx On merge, this will also require changing the required status' for merging, as the old `ci/circleci` named status doesn't get reported in workflows (because we get a status per job)

janpio:
Oh, I actually **can't** merge this before @KrauseFx changes that setting as this is one of the expected checks ;) 

See here for instructions: https://circleci.com/docs/2.0/workflows/#workflows-waiting-for-status-in-github

PR #11244, opened 2017-12-16
Require fixes and optimizations

milch:
The aim of this PR is to fix all requires in all fastlane files. The reason why this is necessary is that *many* parts of fastlane are architected like this:

```ruby
# in tool/lib/tool/fileA.rb

require 'spaceship'

module Tool
  class A
    def use()
      Spaceship.login(...)
      # ...
    end
  end
end

# in tool/lib/tool/fileB.rb

module Tool
  class B
    def use()
      Spaceship.login(...)
      # ...
    end
  end
end

# in tool/lib/tool.rb

require 'fileA'
require 'fileB'

module Tool
# ...
end
```

To a casual observer this might seem fine. However, observe that the code breaks if e.g. the order of the files in `tool/lib/tool.rb` is reversed, if `fileA.rb` is removed, or if in some other file someone tries to `require 'tool/fileB'`. This is also super easy to miss when you review a PR. 

 This ultimately leads to issues such as #11027 and many others (searching for `NameError` brings up over 200, some of which are caused by this, I'm sure there are many others) which are hard to debug and seem to come up at random for some users: depending on the order in which some tools are called dependencies are pulled in that make unrelated tools work. 

In developing this PR I've come up with a rubocop linter that automatically finds missing requires. As I finish this PR I'm going to keep updating that and then submit it to the rubocop repo. That'll make sure that once this is merged and that cop is incorporated, from that point on all dependencies will be correctly required whenever a change is made.

Fixing this additionally provides some advantages "for free":
- Every file will be individually require-able
  - This makes debugging/iterating with `pry` super simple: just require the file you're working on 
  - In many places we have `require 'tool'` when we really only need some specific sub-system of that tool (`fastlane_core` when we really only need `fastlane_core/ui` happens super often)
- It is much easier to see whether you have `require`d the right thing. Sometimes it is hard to judge whether `require 'tool/subsystem/subsubsystem'` will pull in all required dependencies for the stuff you're using. The cop makes this much simpler because it'll tell you, so `require 'tool'` will be a thing of the past
- I'm also trying to replace fastlane-internal requires by `require_relative`
  - This should fix a class of issues that I've seen pop up from time to time where people had weird bugs because `require 'tool'` was pulling in an older version, e.g. the standalone gem they had installed from years ago, instead of the bundled version
  - It should also speed up startup time by quite a bit: `require` always  searches the entire `$LOAD_PATH` (155 directories on my machine with a standard fastlane `Gemfile` install) vs `require_relative` which only needs to search one directory. 
  - `require_relative` for most tools should happen much later (where the dependency is actually used) than the previous pattern of having `require 'other_tool'` in `tool/lib/tool.rb` further improving the startup time 

Disadvantages to this solution:
- For a beginner contributor it might be hard to manage requires 
  - Once the rubocop linter is live this should be easier, since it'll tell you if you're missing something
  - We the reviewers can obviously also help out there if it fails 
- With all dependencies in each file, sometimes cycles can happen
  - Currently these fail the cop, but there is no good way to actually detect it (I think). Will continue investigating 
  - After fixing 5 tools and 1 module I've only had a single cycle though (between `loader.rb` and `upload_metadata.rb` in deliver) so these shouldn't happen too often in the wild 
  - I've avoided other cycles by splitting up the base module definition into it's own file that is separate from the tool entrypoint (i.e. `tool/lib/tool.rb` and `tool/lib/tool/module.rb`)


milch:
Note: Not sure how the test failures are related here, but I'll fix them at the end

milch:
With `fastlane_core`, `deliver`, `sigh`, `cert`, `precheck` and `credentials_manager` optimised/fixed I've done some preliminary statistics and already startup time is ~16% faster on average

fastlane_core/lib/fastlane_core/test_parser.rb
```diff
@@ -55,7 +66,7 @@ def initialize(path)
 
     # Returns the JUnit report as String
     def to_junit
-      JunitGenerator.new(self.data).generate
+      Fastlane::JUnitGenerator.new(self.data).generate
```
milch: Don't know if this even worked before since the class is called `JUnitGenerator`, not `JunitGenerator`

fastlane_core/lib/fastlane_core/test_parser.rb
```diff
@@ -32,7 +41,9 @@ def self.auto_convert(config)
           to_path = path.gsub(".plist", config[:extension])
         end
 
+        # rubocop:disable Lint/MissingRequireStatement
         tp = Trainer::TestParser.new(path)
```
milch: This is weird, this file depends on a third party plugin? Help @KrauseFx

KrauseFx: Uh, that's really weird, I'm pretty sure this doesn't work. Looks like this code is called from SimulatorLauncher?

KrauseFx:
That's awesome! Great work. So when do you use `require` and when `require_relative`?

KrauseFx:
Also, interesting 2 spots you found there

milch:
Currently using `require` for third party gems and for other fastlane tools since it's easier to iterate when I don't have to look up relative depth to the other tools. `require_relative` is currently only used for the same tool. 

However, I'm planning to write a script that'll convert all `require` statements for other fastlane tools to `require_relative`. Should be pretty simple because it only needs to look up the depth of each file in the directory tree and then prepend `../`

snapshot/lib/snapshot/dependency_checker.rb
```diff
@@ -1,49 +1,54 @@
...
 
       if Snapshot::LatestOsVersion.ios_version.to_f < 9 # to_f is bad, but should be good enough
-        UI.error '#############################################################'
-        UI.error "# Your xcode-select Xcode version is below 7.0"
-        UI.error "# To use snapshot 1.0 and above you need at least iOS 9"
-        UI.error "# Set the path to the Xcode version that supports UI Tests"
-        UI.error "# or downgrade to versions older than snapshot 1.0"
-        UI.error '#############################################################'
-        UI.user_error!("Run 'sudo xcode-select -s /Applications/Xcode-beta.app'")
+        FastlaneCore::UI.error '#############################################################'
```
lacostej: why isn't this kept as `UI` ? missing the `require module`?

lacostej (commented):
Great work. #FeelingInspired :)

snapshot/lib/snapshot/dependency_checker.rb
milch: The problem here is that there is a circular dependency here, with the module requiring the dependency checker (which it uses inline) and the dependency checker requiring the module. That's also the reason why `latest_os_version` is only required in this method - because that in turn also requires the module.

.rubocop.yml
```diff
@@ -261,3 +261,6 @@ Style/IfInsideElse:
 # Sometimes we just want to 'collect'
 Style/CollectionMethods:
   Enabled: false
+
+Lint/UnneededDisable:
```
milch: This'll be able to be removed later once we upgrade to a RuboCop version that includes my Lint/MissingRequireStatement checker ... I'll submit a PR to their repo soon

fastlane_core/lib/fastlane_core/update_checker/changelog.rb
```diff
@@ -3,7 +3,7 @@
 module FastlaneCore
   class Changelog
     class << self
-      def show_changes(gem_name, current_version)
+      def show_changes(gem_name, current_version, update_command: "bundle update")
```
milch: This was to break the circular dependency between `changelog.rb` and `update_checker.rb`

spaceship/lib/spaceship/client.rb
```diff
@@ -37,71 +42,16 @@ class Client
...
+    # legacy support
+    BasicPreferredInfoError = Spaceship::BasicPreferredInfoError
+    InvalidUserCredentialsError = Spaceship::InvalidUserCredentialsError
+    NoUserCredentialsError = Spaceship::NoUserCredentialsError
+    ProgramLicenseAgreementUpdated = Spaceship::ProgramLicenseAgreementUpdated
+    InsufficientPermissions = Spaceship::InsufficientPermissions
+    UnexpectedResponse = Spaceship::UnexpectedResponse
+    AppleTimeoutError = Spaceship::AppleTimeoutError
+    UnauthorizedAccessError = Spaceship::UnauthorizedAccessError
+    InternalServerError = Spaceship::InternalServerError
```
milch: All the errors have been refactored to their own `errors.rb` file and live in the `Spaceship` module now, with these wrappers provided for legacy support

spaceship/lib/spaceship/portal/app_service.rb
```diff
@@ -30,85 +30,44 @@ def initialize(service_id, value)
...
-          self::Hotspot
-        end
-
-        def multipath
-          self::Multipath
-        end
-
-        def nfc_tag_reading
-          self::NFCTagReading
+      def self.new_service(id, values: { on: true, off: false })
```
milch: This is a bit of a larger refactor, for two reasons:

1. The require checker can't cope with the constants that are being defined at the end of the file being accessed here
2. It was kinda ugly, having to define the same methods over and over again, when we can have ruby easily do it for us, so we saved some 250 lines

KrauseFx (approved):
This is the most amazing thing ever 🎉

fastlane_core/lib/fastlane_core/update_checker/changelog.rb
lacostej: Alternative is to move the `UpdateChecker.update_command` method into a new class `UpdateCheckerUtils` and make both class call it. That way, no API change.

If you keep it like you proposed, maybe rename the parameter `update_gem_command` ?

frameit/lib/frameit/runner.rb
```diff
@@ -1,6 +1,11 @@
 require 'deliver'
```
lacostej: any reason we require the full deliver here?

milch: Good question, I probably did `require 'deliver'` and wanted to replace it with a more specific version later, then forgot

lacostej: In case that helps, here are occurences of similar large requires:

`Dir.glob("*/lib").map{|s| s.split('/')[0]}.each {|tool| command = "git grep \"require '#{tool}'\""; puts `#{command}`}`

(to run in a irb shell at root)

hjanuschka:
🥛  power move, that's some awesome stuff!!! looking forward to it.

fastlane_core/lib/fastlane_core/update_checker/changelog.rb
milch: Hmm, I think I'll keep it, but do the rename. I saw there's a bunch of tests associated with the method as well, which would require more refactoring. From a usage perspective the API doesn't change anyway, because the parameter has an optional default value

frameit/lib/frameit/runner.rb
milch: Upon further checking, deliver is not actually used in this file at all 😅 I assume the file was probably just copied way back when frameit was made, or something, and that was left over. I'm sure there are some unnecessary requires over all of fastlane. 

I think my MissingRequireStatement cop could even warn for those, since `require` returns `false` if the file was already loaded... hmm...

KrauseFx (approved):
Let's do it :rocket:

PR #11254, opened 2017-12-18
Update code gen default value ignore filters for new method names

taquitos:
Code generation relies on default values set in the config items. Some things should be ignored like local paths that are specific to the developer's workstation. 
Updated the filters to include the new name alias actions.


KrauseFx:
Can we add all GitHub tokens and AWS tokens as well?

taquitos:
@KrauseFx which ones are missing?

KrauseFx (approved):

KrauseFx:
GitHub and AWS tokens, you can find all of them by searching for `GITHUB_API_` for all GitHub related actions, and `s3.rb` for the AWS credentials

PR #11264, opened 2017-12-18
Choose `find` syntax depending on OS (no `find -E` on Ubuntu)

janpio:
`find -E` only works on macOS, all the other unix like OS have no `-F` but use `-regextype`.


revolter (approved):

KrauseFx:
This only affects the fastlane `Fastfile`, right? Do we ever plan not to deploy from a Mac?

janpio:
This affects testing as this is run during/after tests. 
I was surprised as well.

PR #11236, opened 2017-12-15
[WIP] Run tests on Ubuntu (2)

janpio:
~~Theoretically tests on Ubuntu could work - locally they do. Let's see if I can figure out how to configure this. Hope I don't break Circle along the way.~~

Tests run on Ubuntu, proven in #11223. But it takes some modifications. 
This PR includes the changes as individual commits, minus all the experimentation in #11223. 
It will probably never be merged, but be the base for a new branch.

PR #11265, opened 2017-12-18
non-macOS: always return emptry string as xcode_path

janpio:
Currently this method only returns an empty string if it a) is run not on a Mac and b) it is run during a test. 

This doesn't make any sense, why should it only return this during testing? Xcode always will be missing, so the path can always be returned as empty.

(This code has already been tested on macOS and Ubuntu via #11236 and passed)

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -141,7 +141,7 @@ module FastlaneCore
     # @return the full path to the Xcode developer tools of the currently
     #  running system
     def self.xcode_path
-      return "" if self.is_test? and !self.is_mac?
+      return "" unless self.is_mac?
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Ahh, at first I thought it's similar to https://github.com/fastlane/fastlane/commit/5019acb53fc2adf050b4227c3e3216e365ba36ec#diff-6c7c4ae2979901f56b7f8a626b9ced92 and https://github.com/fastlane/fastlane/commit/d6c767b83bb513e15788d5581b99932c36a771e7#diff-6c7c4ae2979901f56b7f8a626b9ced92, but seems like that's not the case.

janpio:
Yeah, all the `xcod_path` stuff is pretty complicated - and so many special cases to handle.

When we get to cross platform support for that, I suspect we will have to dive a bit deeper and hopfefully can clean it up a bit.

PR #11268, opened 2017-12-18
[WIP] Install iTMSTransporter on Circle/Ubuntu

janpio:
This is not really a PR, I am more looking for feedback.

The relevant part is this:

```
     - run:
          name: Install additional dependency iTMSTransporter
          command: |
            cd /tmp
            wget https://github.com/janpio/fastlane-ubuntu-foo/raw/master/iTMSTransporter_installer_linux_1.9.7.sh
            sudo sh iTMSTransporter_installer_linux_1.9.7.sh --target itms --noexec
            cd itms
            sudo sed -e '15,43d' -i install_script.sh
            sudo ./install_script.sh
```

This is a step on Circle CI that is run on Ubuntu, more or less a Shell script.

Problem with this are:
1. I downloaded the `iTMSTransporter_installer_linux_1.9.7.sh` from the developer portal of Apple and put it on Github as a login is required to download.
2. `sudo sed -e '15,43d' -i install_script.sh` deletes parts of the install script that requires paging though some Terms of Services and typing "yes" twice.

Is there a better way to get this dependency fulfilled on a CI server that I am not seeing?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: v1-gems-{{ checksum "Gemfile" }}
+          key: fastlane-ubuntu-gems-{{ checksum "Gemfile" }}
       - run:
           name: Setup Build
           command: |
@@ -23,7 +23,7 @@ jobs:
             brew install shellcheck
             bundle check || bundle install --jobs=4 --retry=3 --path .bundle
       - save_cache:
-          key: v1-gems-{{ checksum "Gemfile" }}
+          key: fastlane-v2-gems-{{ checksum "Gemfile" }}-b
           paths:
             - .bundle
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ jobs:
       CIRCLE_TEST_RESULTS: $HOME/test-results
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
+      FASTLANE_ITUNES_TRANSPORTER_PATH: .bundle
     docker:
       - image: circleci/ruby:2.3
     shell: /bin/bash --login -eo pipefail
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,8 +4,8 @@ jobs:
   build:
     environment:
       CIRCLE_TEST_RESULTS: $HOME/test-results
-      LC_ALL: en_US.UTF-8
-      LANG: en_US.UTF-8
+      LC_ALL: C.UTF-8
+      LANG: C.UTF-8
     docker:
       - image: circleci/ruby:2.3
     shell: /bin/bash --login -eo pipefail
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,19 @@ jobs:
             gem update --system
             bundle check || bundle install --jobs=4 --retry=3 --path .bundle
             sudo apt-get install shellcheck
+      - run:
+          name: Install additional dependency xar
+          command: |
+            cd /tmp
+            mkdir download
+            cd download
+            wget https://github.com/downloads/mackyle/xar/xar-1.6.1.tar.gz
+            tar -xzf xar-1.6.1.tar.gz
+            cd xar-1.6.1
+            ./autogen.sh --noconfigure
+            ./configure
+            make
+            sudo make install
       - save_cache:
           key: fastlane-v2-gems-{{ checksum "Gemfile" }}-b
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,8 @@ jobs:
             mkdir -p $CIRCLE_TEST_RESULTS
             echo "2.3" > .ruby-version
             gem update --system
-            brew install shellcheck
             bundle check || bundle install --jobs=4 --retry=3 --path .bundle
+            sudo apt-get install shellcheck
       - save_cache:
           key: fastlane-v2-gems-{{ checksum "Gemfile" }}-b
           paths:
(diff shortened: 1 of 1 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lacostej:
1. Looks like the official URL for linux is https://itunesconnect.apple.com/WebObjects/iTunesConnect.woa/ra/resources/download/Transporter__linux/bin/ (I had a hard time finding it...) We could download it automatically using fastlane as we already support itunesconnect

2. not sure modifying the script is the proper way to accept the user agreement. Are there CLI options to answer yes? If not does it show a GUI or is in text based? Otherwise you could try with the expect tool to script the answers.

3. I believe one would want a docker setup file to avoid having to do this like that as part of the build

janpio:
Re 1: Yes, problem is that this only works when logged in on https://developer.apple.com/account/

Re 2: No CLI options that would skip the licence and prompts. This is what is available:

```
foo@foo-VirtualBox:~/Downloads$ sudo sh iTMSTransporter_installer_linux_1.9.7.sh --yes
[sudo] password for foo: 
Unrecognized flag : --yes
Makeself version 2.1.4
 1) Getting help or info about iTMSTransporter_installer_linux_1.9.7.sh :
  iTMSTransporter_installer_linux_1.9.7.sh --help   Print this message
  iTMSTransporter_installer_linux_1.9.7.sh --info   Print embedded info : title, default target directory, embedded script ...
  iTMSTransporter_installer_linux_1.9.7.sh --lsm    Print embedded lsm entry (or no LSM)
  iTMSTransporter_installer_linux_1.9.7.sh --list   Print the list of files in the archive
  iTMSTransporter_installer_linux_1.9.7.sh --check  Checks integrity of the archive
 
 2) Running iTMSTransporter_installer_linux_1.9.7.sh :
  iTMSTransporter_installer_linux_1.9.7.sh [options] [--] [additional arguments to embedded script]
  with following options (in that order)
  --confirm             Ask before running embedded script
  --noexec              Do not run embedded script
  --keep                Do not erase target directory after running
			the embedded script
  --nox11               Do not spawn an xterm
  --nochown             Do not give the extracted files to the current user
  --target NewDirectory Extract in NewDirectory
  --tar arg1 [arg2 ...] Access the contents of the archive through the tar command
  --                    Following arguments will be passed to the embedded script
```

Text based interface, here are screenshots: 
image
image
...
image
image
image

Here is the actual script that is executed in the background:
```
#!/bin/sh 
#PATH=/usr/bin:/bin
#umask 022
#NODOT_VERSION=142_12
#javahome=j2sdk1.4.2_12
#PACKED_JARS="lib/tools.jar jre/lib/rt.jar jre/lib/jsse.jar jre/lib/charsets.jar jre/lib/ext/localedata.jar jre
#/lib/plugin.jar jre/javaws/javaws.jar"
#LINUX_RPM=
#JAVAWS_BIN=jre/javaws/javaws
#libthread_path=
#localinstall=$1
#diskSpaceRequired=104668
#MIME_TYPE=application/x-java-jnlp-file
 
more License.txt
agreed=
while [ x$agreed = x ]; do
    echo
    echo "Do you agree to the above license terms? [yes or no] "
    read reply leftover
    case $reply in
        y* | Y*)
            agreed=1;;
        n* | N*)
    echo "You must agree to the license agreement to install this software!";
    exit 1;;
    esac
done

cat Welcome.txt
agreed=
while [ x$agreed = x ]; do
    echo
    echo "Continue? [yes or no] "
    read reply leftover
    case $reply in
        y* | Y*)
            agreed=1;;
        n* | N*)
    echo "Installation aborted by user";
    exit 1;;
    esac
done

if [ -d /usr/local/itms/ ]
 then
  rm -r /usr/local/itms
fi
if [ -d ~/.itmstransporter ]; then rm -rf ~/.itmstransporter; fi
sudo cp -r itms /usr/local/
sudo chmod 555 /usr/local/itms/bin/iTMSTransporter

#use awk to keep finding the parent pid until the user is
#is not root or we reach the top process
FILE_OWNER=`ps -e -o "user ppid pid" | awk -v proc=$$ '{
	user_a[row] = $1
	parent_a[row] = $2
	child_a[row] = $3
#print $1,$2,$3
#print row
	row++
}


END {
	p=proc
        # building one list of users.
        # the user of the current process is the
        # first item.  the user of its parent process
        # is the second item. the user of its grandparent
        # process is the third item.  etc..
        # "pi++<50" is there as a safety to avoid
        # infinite loops
	while (p!=0 && pi++<50){
		user=userForChild(p)
#print user, p
#print "::",user,":",p,":",pi,":"
		p=parentForChild(p)
		userList[pi]=user

	}
        # go through the list of users until we find
        # the first non-root user.  this is the user
        # that su-ed to root.
	i=0
	do {
                i++
 		user=userList[i]
#print "user=:",user,":"
#print i, pi
#print "pi=",pi
#print "i=",i
	}
	while (i<pi && user=="root")
	print userList[i]
}

function userForChild(l_proc) {
	rval=""
	for(i=0;rval=="" && i<row;i++) {
		if (child_a[i]==l_proc) {
			rval=user_a[i]
		}
	}
	return rval
}

function parentForChild(l_proc) {
	rval=""
	for(i=0;rval=="" && i<row;i++) {
		if (child_a[i]==l_proc) {
			rval=parent_a[i]
		}
	}
	return rval
}'`

#chown $SUDO_USER /usr/local/itms/share/iTMSTransporter.woa/Contents/Resources/id_dsa
#echo "FILE_OWNER: "$FILE_OWNER
```

Re 3: Ah, so just taking the current image and adding iTMSTransporter would be an option. Interesting.

janpio:
See #11331

Issue #11266, opened 2017-12-18
Return a string from getBuildNumber func? 


altyus:
https://github.com/fastlane/fastlane/blob/3aaababca9498872e3fd631a379a5b766a08f1ed/fastlane/swift/Fastlane.swift#L1396

Wouldn't you want this to return the build number?

taquitos:
🤦‍♂️ thank you.

We use code generation that depends on a few things, and it looks like I missed adding a `return_type` to that action.

Fixing that right now

KrauseFx:
Beautiful, this will be part of the next release, thanks @altyus 👍

acegreen:
How about getVersionNumber()?

taquitos:
@acegreen you want the world, don't you 😜
PR incoming

acegreen:
@taquitos I changed it on my end, was going to do a PR. No rush, if you don't get to it, I'll do it

taquitos:
@acegreen thanks, the fix just got merged. Let me know if you find anything else, I had a feeling a few of these might slip through.

acegreen:
@taquitos I see a bunch of get....() methods that have no return in that file. I would return String for all of them

taquitos:
That's a great point. One thing that prevented me from doing that is some of those only store a file somewhere, or they save an object in the lane context for later use. Maybe in those cases we could return a path, or update the action to return the value too.

acegreen:
Ah true, just to keep it simple for now, maybe for the obvious ones return String. And in the future, more use-cases it will be clear what to do with the exceptions you mentioned

On a side note, I noticed fast lane swift doesn't have a deploy_to_fabric equivalent, I thought it would be something like uploadToFabric(...)

PR #11269, opened 2017-12-19
Fix for #11266, get_build_number missing return type

taquitos:
Missing return type `:string` added, fix for #11266


jdee (approved):
:+1: 🍏

KrauseFx (approved):

PR #11277, opened 2017-12-19
[Fastlane.swift] Fix gitBranch() has no return type

thii:
Generated Swift API of `git_branch` action had no return type.


KrauseFx (approved):
Nice catch, thanks for your contribution 👍

PR #11289, opened 2017-12-19
Deploying fastlane should bump and commit fastlane swift files

taquitos:
Version bumping fastlane should version bump and commit swift files too


KrauseFx (approved):
I'm scared this will push my GitHub API token or AWS credentials to GitHub... Did we make sure none of those parameters are included in the Swift default values any more?

PR #11291, opened 2017-12-20
Expose the swift Snapshot class to Objective-C

0xced:
Explicitly annotate the Snapshot class with `@objcMembers` since Swift 3 `@objc` inference is deprecated in Swift 4.

This is required for Objective-C developers to be able to use [Snapshot snapshot:@"Screenshot" timeWaitingForIdle:20];

It is not required to add `@objcMembers` in the SnapshotHelperXcode8Version.swift file since Xcode < 9 does not support Swift 4.


KrauseFx (approved):
Nice catch, thanks @0xced 👍

PR #11300, opened 2017-12-20
[snapshot] Fixes failure on physical device due to snapshot()

umverteiler:
### Checklist

### Motivation and Context
Running any XCUITest that contains a snapshot() call on a **physical device** will fail. That's because the app property inside setupSnaphot() is set after (!) the setting of cacheDirectory. Unfortunately cacheDirectory can't be computed using pathPrefix() on a physical device due to the missing SIMULATOR_HOST_HOME property. So pathPrefix() will fail and by that always skip the setting of the app property as well (unnecessarily).

### Description
- Makes app and cacheDirectory properties in SnapshotHelper.swift optional (like they should be)
- Sets app before setting cacheDirectory
- On a physical device the snapshot() still not works, but at least it doesn't fails / crashes the test anymore

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -60,17 +60,19 @@ enum SnapshotError: Error, CustomDebugStringConvertible {
 
 @objcMembers
 open class Snapshot: NSObject {
-    static var app: XCUIApplication!
-    static var cacheDirectory: URL!
+    static var app: XCUIApplication?
+    static var cacheDirectory: URL?
     static var screenshotsDirectory: URL? {
-        return cacheDirectory.appendingPathComponent("screenshots", isDirectory: true)
+        return cacheDirectory?.appendingPathComponent("screenshots", isDirectory: true)
     }
 
     open class func setupSnapshot(_ app: XCUIApplication) {
+        
+        Snapshot.app = app
+
         do {
             let cacheDir = try pathPrefix()
             Snapshot.cacheDirectory = cacheDir
-            Snapshot.app = app
             setLanguage(app)
             setLocale(app)
             setLaunchArguments(app)
@@ -80,6 +82,11 @@ open class Snapshot: NSObject {
     }
 
     class func setLanguage(_ app: XCUIApplication) {
+        guard let cacheDirectory = self.cacheDirectory else {
+            print("CacheDirectory is not set - probably runnig a physical device?")
+            return
+        }
+        
         let path = cacheDirectory.appendingPathComponent("language.txt")
 
         do {
@@ -92,6 +99,11 @@ open class Snapshot: NSObject {
     }
 
     class func setLocale(_ app: XCUIApplication) {
+        guard let cacheDirectory = self.cacheDirectory else {
+            print("CacheDirectory is not set - probably runnig a physical device?")
+            return
+        }
+        
         let path = cacheDirectory.appendingPathComponent("locale.txt")
 
         do {
@@ -107,6 +119,11 @@ open class Snapshot: NSObject {
     }
 
     class func setLaunchArguments(_ app: XCUIApplication) {
+        guard let cacheDirectory = self.cacheDirectory else {
+            print("CacheDirectory is not set - probably runnig a physical device?")
+            return
+        }
+        
         let path = cacheDirectory.appendingPathComponent("snapshot-launch_arguments.txt")
         app.launchArguments += ["-FASTLANE_SNAPSHOT", "YES", "-ui_testing"]
 
@@ -135,6 +152,12 @@ open class Snapshot: NSObject {
         #if os(OSX)
             XCUIApplication().typeKey(XCUIKeyboardKeySecondaryFn, modifierFlags: [])
         #else
+            
+            guard let app = self.app else {
+                print("XCUIApplication is not set")
+                return
+            }
+            
             let screenshot = app.windows.firstMatch.screenshot()
             guard let simulator = ProcessInfo().environment["SIMULATOR_DEVICE_NAME"], let screenshotsDir = screenshotsDirectory else { return }
             let path = screenshotsDir.appendingPathComponent("\(simulator)-\(name).png")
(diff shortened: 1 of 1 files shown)


LordNali:
I signed it!

KrauseFx:
Hey @LordNali, thanks for your contribution on this. I didn't know you can run _snapshot_ on real devices, how does it work?

LordNali:
You're right - you can't run snapshot on physical device - but we have some testcases that are using snapshot as a 'helper' to create screenshots for debugging / documentation during test execution. Theses testcases all fail on a physical device due to snapshot, even though the actual test would run perfectly on a physical device.

So in the end this fix doesn't help running snapshot on physical device, but prevent testcases that are using snapshot from crashing on physical devices ;-)

KrauseFx:
Ah, alright, cool. Considering that, it probably doesn't make sense to merge this to the _fastlane_ code base, as it adds complexity, to cover a use-case we don't support. Is it cool to close this PR and you run your own fork for now?

LordNali:
The implicit unwrapping of the properties app and cacheDirectory isn't good design. Furthermore setting property app only in case the cache directory has been set, doesn't seem to be good idea either.

I'm not sure what speaks against (merging) this PR, because in the end we're loosing nothing here, but will gain an error-free run in case someone runs tests that include a snapshot call on a physical device.

The purpose of the PR is not to enhance snapshot in terms of functionality, but to ensure that testcases that are using snapshot still run on a physical device.

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -80,6 +82,11 @@ open class Snapshot: NSObject {
     }
 
     class func setLanguage(_ app: XCUIApplication) {
+        guard let cacheDirectory = self.cacheDirectory else {
+            print("CacheDirectory is not set - probably runnig a physical device?")
```
KrauseFx: `probably running on a physical device` <- added on

KrauseFx:
Ahh, right, that makes sense. Sorry for asking stupid questions, I haven't actually learned Swift yet, so I appreciate you pointing out why we do this 👍

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -135,6 +152,12 @@ open class Snapshot: NSObject {
         #if os(OSX)
             XCUIApplication().typeKey(XCUIKeyboardKeySecondaryFn, modifierFlags: [])
         #else
+            
+            guard let app = self.app else {
+                print("XCUIApplication is not set")
```
KrauseFx: This error message could be a little clearer so that the user knows what they have to do to fix the error 👍

LordNali:
> ... I haven't actually learned Swift yet ...

And yet you're the author of one of the greatest tool(chain) for Swift Developers  :-)

I added the missing 'on' to the print message and extended the Snapshot error by a specific one for 'running on a physical device'. Finally I added the additional explanation in regard to fixing 'XCUIApplication is not set' error message.

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -83,7 +86,7 @@ open class Snapshot: NSObject {
 
     class func setLanguage(_ app: XCUIApplication) {
         guard let cacheDirectory = self.cacheDirectory else {
-            print("CacheDirectory is not set - probably runnig a physical device?")
+            print("CacheDirectory is not set - probably runnig on a physical device?")
```
janpio: runnig => running

milch (commented):
You also need to increment the version number in this file, see the comment at the beginning:

```
 // -----------------------------------------------------
 // IMPORTANT: When modifying this file, make sure to
 //            increment the version number at the very
 //            bottom of the file to notify users about
 //            the new SnapshotHelper.swift
 // -----------------------------------------------------
```

LordNali:
Okay, updated it to 1.8.1 (assuming you're using some Semantic Versioning?).

janpio:
Interesting, there seems to be a test for this that is now failing:

```
  1) Snapshot Snapshot::Runner Parses embedded SnapshotHelper.swift finds the current embedded version
     Failure/Error: expect(helper_version).to match(/^SnapshotHelperVersion \[\d.\d\]$/)

       expected "SnapshotHelperVersion [1.8.1]" to match /^SnapshotHelperVersion \[\d.\d\]$/
       Diff:
       @@ -1,2 +1,2 @@
       -/^SnapshotHelperVersion \[\d.\d\]$/
       +"SnapshotHelperVersion [1.8.1]"
     # ./snapshot/spec/runner_spec.rb:11:in `block (4 levels) in <top (required)>'
```

LordNali:
Well, I'm not sure, but do we need to specify the new SnapshotHelper version anywhere else?

janpio:
So I looked at this test a bit:
https://github.com/fastlane/fastlane/blob/c7b074dc4b564bea9f35814466734a704bcc20d4/snapshot/spec/runner_spec.rb#L6-L13
It seems to expect the version to be x.y, not x.y.z. So this indeed seems not to be semver, but Major.Minor. 

Solutions seems to be to change it to 1.9 then I guess?

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -135,6 +155,12 @@ open class Snapshot: NSObject {
         #if os(OSX)
             XCUIApplication().typeKey(XCUIKeyboardKeySecondaryFn, modifierFlags: [])
         #else
+            
+            guard let app = self.app else {
```
janpio: Unrelated change?

janpio (approved):
Thanks, tests seem to like 1.9 :)

One last question, but I think this is good to merge no matter what you reply.

snapshot/lib/assets/SnapshotHelper.swift
LordNali: Not really - since property app has become optional (like cacheDirectory) we need to unwrap it before using it.

janpio:
Question answered, so let's merge this 🥇 Thanks for the PR @LordNali.

PR #11301, opened 2017-12-20
Make swift code generation more aware of sensitive default values

taquitos:
Code generation uses actions' `config_item`s
Some of these items have `default_value`s that read from `ENV` or other sensitive areas. This PR audits that use and adds in properties to look for `code_gen_sensitive` and ignore the `default_value`.
Also added `code_gen_default_value` if you really want a default value there and the normal default is sensitive.


KrauseFx (dismissed):
Need to think more about it

taquitos:
@KrauseFx any other thoughts? I think this is a better solution than what we currently have. I was thinking about:
- When we add new params, we should define all properties about that param in the config_item. 

One possible solution I dismissed was creating a separate configuration specifically for code gen. I didn't like that because people might not know to add metadata about a parameter somewhere other than in the config item.

Any solution that splits config information about a config_item away from the config_item is probably bad for discoverability.

fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -15,6 +15,9 @@ class ConfigItem
     # the value which is used if there was no given values and no environment values
     attr_accessor :default_value
 
+    # the value which is used during code generation
```
KrauseFx: This might require some more comments, what's `code generation`, for new developers it might not be clear it's Swift related

fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -33,8 +36,12 @@ class ConfigItem
     attr_accessor :deprecated
 
     # [Boolean] Set if the variable is sensitive, such as a password or API token, to prevent echoing when prompted for the parameter
+    # If a default value exists, it won't be used during code generation as default values can read from environment variables.
     attr_accessor :sensitive
 
+    # [Boolean] Set if the default value should never be used during code generation for Swift
```
KrauseFx: Same here, regarding more comments

KrauseFx (approved):
Looking at the code, this looks solid, I like it 👍 

I couldn't help but notice there is not a test covering this, would it be easy to add? Also 2 pieces that will require some more comments

fastlane_core/lib/fastlane_core/configuration/config_item.rb
taquitos: I might have added more comments

KrauseFx (approved):

taquitos:
@KrauseFx I couldn't help but add more unit tests and refactor where default values were being filled out #simplify

PR #11305, opened 2017-12-20
[test_sample_code] Respect skip_type_validation when testing sample code.

mpirri:
### Checklist

### Motivation and Context

fastlane ConfigItem's can be marked as having their type checking disabled (`skip_type_validation`). This is done for legacy reasons - some ConfigItems have traditionally received data of various types and then switch their behavior based on the type of data received. However, we were not respecting this config item flag when test-running sample code in the test_sample_code action.

### Description

Disable the type-check when skip_type_validation is set for a ConfigItem.


taquitos (approved):
🎈:goat:

PR #11306, opened 2017-12-20
If using system ruby, and json isn't installed, do the right thing

taquitos:
- use quirks mode for strings/bools/ints
Fixes #11276

PR #11296, opened 2017-12-20
[supply] replace hardcoded string `en-US` with actual language

janpio:
Error messages currently always output `en-US`.
This replaces it with the actual language being handled at the moment.

(fixes #9962)

PR #11270, opened 2017-12-19
Fix version selection with import_from_git

jdee:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

`checkout_param_twiddle_wakka`, we hardly knew ye.

Fix #11252.

Use RubyGems API for tag resolution in `import_from_git`. This also now allows things like:

```Ruby
import_from_git(url: repo_url, version: [">= 1.1.0", "< 1.7.0"])
```

Which can be a bit more specific than an optimistic match.

### Description

Use `Gem::Requirement` and `Gem::Version` to determine which tags match a version. The `version` param is passed directly to the [Gem::Version.new](http://docs.seattlerb.org/rubygems/Gem/Requirement.html) method. Pass a string or an array.

Removed all the "Version helpers" methods except `fetch_remote_tags`, which was simplified and corrected as well.

PR #11251, opened 2017-12-18
Verify all env variables are set for deployment

KrauseFx:
Until now we checked for those variables too late during the deployment, this PR fixes it

PR #11312, opened 2017-12-21
Version bump to 2.70.3

taquitos:
Auto-generated by fastlane 🤖

**Changes since release '2.70.2':**

* Update support for `latest` for `download_dsyms` action (#10828) via Shai Mishali
* Verify all env variables are set for deployment (#11251) via Felix Krause
* Update yard development dependency (#11308) via Felix Krause
* Fix version selection with import_from_git (#11270) via Jimmy Dee
* If using system ruby, and json isn't installed, do the right thing (#11306) via Joshua Liebowitz
* Respect skip_type_validation when testing sample code. (#11305) via Mark Pirri
* [supply] replace hardcoded string `en-US` with actual language (#11296) via Jan Piotrowski

### DIFF
--- a/fastlane/lib/fastlane/version.rb
+++ b/fastlane/lib/fastlane/version.rb
@@ -1,5 +1,5 @@
 module Fastlane
-  VERSION = '2.70.2'.freeze
+  VERSION = '2.70.3'.freeze
   DESCRIPTION = "The easiest way to automate beta deployments and releases for your iOS and Android apps".freeze
   MINIMUM_XCODE_RELEASE = "7.0".freeze
   RUBOCOP_REQUIREMENT = '0.49.1'.freeze
--- a/fastlane/swift/Deliverfile.swift
+++ b/fastlane/swift/Deliverfile.swift
@@ -11,4 +11,4 @@ class Deliverfile: DeliverfileProtocol {
 
 
 
-// Generated with fastlane 2.70.1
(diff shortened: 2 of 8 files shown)


ohayon (approved):
Are those versions in the swift files bumped automatically? 🌟  :nice: 🚀

taquitos:
@ohayon they sure are 🎉

jdee:
I noticed that in this 2.70.3 release all the generated Swift files say generated by 2.70.2. I wonder if that could lead someone to think their generated API is out of date. I assume the version is incremented after the Swift API is generated. I imagine that's easy to change, but maybe you made a conscious decision to do it this way.

I've developed a habit of actually doing the version bump (without a tag) immediately after each release in my repos, so that unreleased code in master always has a different version from what's released to prod. The tag is added only when the release is done (by `rake release` in a gem), and each version bump comes immediately after the preceding tag. I think this can have certain benefits. It _has_ led at least one user to conclude that I forgot to release something in the past, and had to be explained.

Anyway, if the versions at the end of each Swift source file should be changed, that would be another way to do it.

taquitos:
I was thinking about this stuff today, and you bring up a good point. My thought process was "Well, technically, it was generated on the previous version because the next version hasn't been released yet". That being said... I think the point about user confusion is more important than being `technically` correct. In the future, I think we should increment the version, then generate the swift. Thanks for pointing that out @jdee

PR #11320, opened 2017-12-21
Tests: Mark skipped tests as pending (with reason)

janpio:
Skip examples that don't work on non-macOS platforms instead of filtering them (which marks and reports them as pending). also offer reasons for skipping.

(Implementation details via discussion at http://groups.google.com/forum/#!topic/rspec/5qeKQr_7G7k)

Gives output for Ubuntu:

> [17:45:21]: ▸ 4704 examples, 0 failures, 175 pending

No change in test output on macOS of course, so a bit hard to review. Please just note that it doesn't break macOS and everything is fine ;)

### DIFF
--- a/scan/spec/xcpretty_reporter_options_generator_spec.rb
+++ b/scan/spec/xcpretty_reporter_options_generator_spec.rb
@@ -1,5 +1,5 @@
 describe Scan do
-  describe Scan::XCPrettyReporterOptionsGenerator do
+  describe Scan::XCPrettyReporterOptionsGenerator, requires_xcodebuild: true do
     before(:all) do
       options = { project: "./scan/examples/standard/app.xcodeproj" }
       Scan.config = FastlaneCore::Configuration.create(Scan::Options.available_options, options)
@@ -7,7 +7,7 @@ describe Scan do
     end
 
     describe "xcpretty reporter options generation" do
-      it "generates options for the junit tempfile report required by scan", requires_xcodebuild: true do
+      it "generates options for the junit tempfile report required by scan" do
         generator = Scan::XCPrettyReporterOptionsGenerator.new(false, "html", "report.html", "/test_output", false)
         reporter_options = generator.generate_reporter_options
         temp_junit_report = Scan.cache[:temp_junit_report]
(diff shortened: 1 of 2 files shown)


milch (commented):
Since the person from the Google groups has mentioned some caveats, e.g. not being able to "un-skip" individual samples, is it somehow possible to verify this is not being done? e.g. by going through the test examples metadata and seeing if any have `skip` set explicitly to false?

janpio:
Technically this would be possible with e.g. rubocop or any other process that scans all the files I think. 

But in my opinion this currently is not a thing we have to care too much about as a) this PR itself introduces the concept of skipping via example/group metadata into the project, and b) the override is limited to non-macOS platforms. 

If one _would_ add a test into a rspec file that is affected by this override, and define it as `skip:false`, they would hopefully notice that the test is not executed but continues to be listed as "pending" (which indicates that it is still skipped). A solution then would be to extract that one test e.g. into a group that has is not skipped and has the same `before(:all)`.

If we really feel this "magic" solution is to intransparent, I had a solution that explicitly added a "skip" to the  `before(:all)` of that one test (only for non-macOS). I replaced with this fancy version as this is more general. (Again, not _really_ relevant as this concerns just one test file at the moment, but I felt it was better not to have the explicit "skip" in the `before(:all)` as this could be forgotten when removing the tags from the examples/group and then could cause unexpected results as the skipping of the `before(:all)` is _not_ reported as the skipping of the test example (as pending test)).

`before(:all)` really are a strange thing in RSpec, as Myron said in the mailing list thread.

milch:
Hmm, interesting. Well it is documented, so it should probably be ok. 

My concern wasn't with how this is used right now, but now that it has been introduced we have to expect people to possibly use this technique, and then it's easy to forget during review and we'll have strange errors or flaky tests later on. But I think you're right, it's probably fine as is. 

I like the solution overall (that we have separate groups that are disabled because of dependencies on certain tools), but `before(:all)` really seems like it should be avoided completely in the future. I suspect most specs would probably continue working if we replaced `before(:all)` with a `before(:each)` since they mostly just set up expects or the variables anyway

janpio:
Yes, I though about that as well. But running the same setup code e.g. >20 times (multiplied by the many hundred test files we have) instead of just once probably could have a pretty hefty performance impact, so again I chose to keep it the way it is until we notice any real benefit.

(I am not sure we will keep this "skipping" mechanism for the long term anyway, as #windows will work towards fixing this stuff in the actual code being run somehow. If we are really successful, we can just rip it out again.)

(Also: Each newly added test that uses one of these tags to skip tests will have to be looked at at least twice as this really shouldn't be the normal case.)

KrauseFx (approved):
Ha, that's perfect, nice work 👍

PR #11321, opened 2017-12-21
Testflight: New method to create group

paramv17:
### Checklist
- ✅  I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- ✅  I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- ✅  I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- ✅  I've updated the documentation if necessary.

### Motivation and Context
This new api will allow users to create a testing group in testflight.

I tested following scenarios on testflight : 
1. Created a group. - successfully created. 
2. Left parameters empty while calling this method. - error stating the parameter not provided.
3. Created a group with the name already present. - This returns a group that is already present. The post API creates a new group with the same name but different id and throws no warnings/errors, which might be an issue - so I am handling it here and returning the existing group.

From the above scenarios, I added spec tests for 1 & 3. Scenario 2 is common behaviour, which fastlane handles well as it is. 

### Description
I created a client method that takes the app id and group name and creates a group. I used this method in Spaceship::TestFlight::Group to implement a method called create!. I added spec tests where applicable.

### DIFF
--- a/spaceship/spec/test_flight/client_spec.rb
+++ b/spaceship/spec/test_flight/client_spec.rb
@@ -17,7 +17,6 @@ describe Spaceship::TestFlight::Client do
   subject { TestFlightTestClient.new(current_team_id: 'fake-team-id') }
   let(:app_id) { 'some-app-id' }
   let(:platform) { 'ios' }
-
   context '#assert_required_params' do
     it 'requires named parameters to be passed' do
       expect do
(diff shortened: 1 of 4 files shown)


pav1790:
I must have forgotten to run rubocop -a after making some quick changes. :( 

Making another commit just as soon as the specs finish running.

pav1790:
Thanks @janpio! :) Should it not be "Spaceship" instead of "action"?

spaceship/spec/test_flight/client_spec.rb
```diff
@@ -17,7 +17,6 @@ def handle_response(response)
   subject { TestFlightTestClient.new(current_team_id: 'fake-team-id') }
   let(:app_id) { 'some-app-id' }
   let(:platform) { 'ios' }
-
```
janpio: accidental?

janpio:
You are of course right. I just read testflight and didn't really look at the code. Sorry ;)

spaceship/spec/test_flight/client_spec.rb
pav1790: Either I removed it or my ide. In any case, I can add it back. : )

pav1790:
Thank you!

KrauseFx (approved):
This is beautiful, this is how a perfect PR looks like, tests, descriptions and everything 👍 ❤️

KrauseFx:
@pav1790 can you send me your email via email github@krausefx.com?

PR #11325, opened 2017-12-22
Update reference to age rating template

max-ott:
### Checklist

### DIFF
--- a/spaceship/docs/iTunesConnect.md
+++ b/spaceship/docs/iTunesConnect.md
@@ -102,6 +102,7 @@ v.set_rating({
 })
 # Available values:
 # https://github.com/fastlane/fastlane/blob/master/deliver/assets/example_rating_config.json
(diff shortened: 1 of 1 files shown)


max-ott:
I signed it!

janpio:
Hmm, this is actually also available in the docs:
https://docs.fastlane.tools/actions/deliver/#available-age-rating-groups

But you first have to open the "Reference" `<details>` element:
https://docs.fastlane.tools/actions/deliver/#reference
(Then click the first link again)

This includes better documentation on possible values.
But should probably link to the file you added here, as an actual "example" of what the file should look like is missing there.

KrauseFx:
Nice catch @janpio 👍

janpio:
Javascript in the docs pages would probably be the way to go:
https://stackoverflow.com/a/37033774/252627
But this is not exactly what we need... will open an issue on fastlane/docs.

@max-ott I think for now it would be best to include the link you used and the #reference one.

janpio (approved):
Added the second link for completeness, now this is good to merge.

PR #11329, opened 2017-12-22
Improve error handling when parse plist file

v9n:
The error is mis-leading. The plist file may exisits but in a format that we cannot parse(such as binary plist) leads to error like `invalid byte sequence in UTF-8`.

We need to handle this properly and output the correct error so user won't confuse.

### Checklist

### DIFF
--- a/fastlane/lib/fastlane/actions/get_info_plist_value.rb
+++ b/fastlane/lib/fastlane/actions/get_info_plist_value.rb
@@ -10,7 +10,8 @@ module Fastlane
 
         begin
           path = File.expand_path(params[:path])
-          plist = Plist.parse_xml(path)
+
+          plist = File.open(path) { |f| Plist.parse_xml(f) }
 
           value = plist[params[:key]]
           Actions.lane_context[SharedValues::GET_INFO_PLIST_VALUE_CUSTOM_VALUE] = value
@@ -18,7 +19,6 @@ module Fastlane
           return value
         rescue => ex
           UI.error(ex)
-          UI.error("Unable to find plist file at '#{path}'")
         end
       end
 


kureikain:
I signed it!

fastlane/lib/fastlane/actions/get_info_plist_value.rb
```diff
@@ -10,6 +10,12 @@ def self.run(params)
 
         begin
           path = File.expand_path(params[:path])
+          
+          unless File.exist?(path)
+            UI.error("Unable to find plist file at '#{path}'")
+            return
+          end
+          
           plist = Plist.parse_xml(path)
```
jdee: Thanks for catching this. I usually do this:

```Ruby
plist = File.open(path) { |f| Plist.parse_xml(f) }
```

The problem is actually here: https://github.com/patsplat/plist/blob/master/lib/plist/parser.rb#L66. If the file does not exist, the input is assumed to be XML. Parsing fails because it's a path, and the method returns nil.

I prefer the approach above because `File.open` raises `Errno::ENOENT` in this case, so you get consistent exception handling instead of the custom error message for this one case, and it makes the code simpler.

jdee (commented):
Please see inline comment.

jdee:
Also it looks like you need to run `bundle exec rubocop -a` or manually get rid of some whitespace.

kureikain:
thanks @jdee. Great review. I lean something today.

KrauseFx (approved):
Nice and simple fix, thanks @kureikain 👍

KrauseFx:
@jdee since you did the initial review, feel free to merge whenever you think it's ok 👍

jdee:
Thanks for this fix @kureikain.

PR #11331, opened 2017-12-22
Add docker config for circle 

### LINKED ISSUE #11302: [Ubuntu/CircleCI] Create Docker container with our dependencies
Quoting @DanToml in [this PR comment](https://github.com/fastlane/fastlane/pull/11242/files#r158092394):

> We should build a docker container that includes these (based on the circleci/ruby:2.3 container)

"these" refers to:
* `sudo apt-get install shellcheck`
* #11280, `xar`
* probably #11268, `iTMSTransporter`

milch:
As a followup to @janpio's enabling of concurrent Linux & macOS tests on Circle, dependencies are refactored into a separate docker container here 

closes #11302


milch:
See #11302

.circleci/Dockerfile
```diff
@@ -0,0 +1,77 @@
+FROM debian:jessie as itms_installer
+
+ENV ITMS_VERSION "1.9.7"
+USER root
+
+WORKDIR /tmp
+
+# Install iTMSTransporter
+ADD https://github.com/janpio/fastlane-ubuntu-foo/raw/master/iTMSTransporter_installer_linux_$ITMS_VERSION.sh .
```
KrauseFx: Is there a chance for us to move this to the fastlane org, to minimize the external dependencies? Also a 50MB `.sh` file?!

.circleci/config.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       LANG: C.UTF-8
       FASTLANE_ITUNES_TRANSPORTER_PATH: .bundle
     docker:
-      - image: circleci/ruby:2.3
+      - image: milch/fastlane-base
```
KrauseFx: Best 🥛 🍼

.circleci/Dockerfile
milch: cc @janpio, this seems to be your repo 

I think the file is kind of like a self-extracting zip file, except it's a tar archive and sh script at the same time. Pretty weird

.circleci/config.yml
milch: If you create a fastlane org on hub.docker.com we can obviously move it there, this was just more convenient for me to test 😄

milch:
@janpio This seems to work, all tests on Ubuntu pass 👍🏼

.circleci/Dockerfile
janpio: This file is offered by Apple, but only when logged in into developer.apple.com/account. For my PoC I just downloaded it and added it manually.

.circleci/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+
+ENV PATH $PATH:/usr/local/itms/bin
+
+# Java versions to be installed
+ENV JAVA_VERSION 8u131
+ENV JAVA_DEBIAN_VERSION 8u131-b11-1~bpo8+1
+ENV CA_CERTIFICATES_JAVA_VERSION 20161107~bpo8+1
+
+# Needed for fastlane to work
+ENV LANG C.UTF-8
```
janpio: Do you know what the `C` here means and why it is not `en_US` as on macOS? I still have an open support topic at Circle for that: https://discuss.circleci.com/t/difference-between-en-us-utf-8-and-c-utf-8/18835

.circleci/Dockerfile
janpio: Because of the possible terms problems the current circle/config.yml does NOT actually install iTMSTransporter for now!

.circleci/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+ENV LC_ALL C.UTF-8
+
+# Required for iTMSTransporter to find Java
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/jre
+
+COPY --from=itms_installer /usr/local/itms /usr/local/itms
```
janpio: Is this necessary? Really running the iTMS installer should automatically move the files there.

.circleci/config.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       LANG: C.UTF-8
       FASTLANE_ITUNES_TRANSPORTER_PATH: .bundle
```
janpio: If you really install iTMS, this path should be updated to the actual installation path of iTMS.

janpio (commented):
Looking good!

As I wrote, iTMS is actually _not_ installed in the current setup as it was not clear how I should handle a) hosting the file and b) the licence acceptance stuff (` sed -e '15,43d' -i install_script.sh` just deletes the line that are responsible for that...) and c) the licence/terms situation for iTMSTransporter in general.

(As I know nothing about how Docker works I thought we would build the image once locally, which would take care of problems a) and b) and only leave c) to be solved - but as this is a script now as well, this doesn't really seem to be the case. Will have to educate myself a bit here)

So it might be better to get rid of that here as well for now, the tests aren't actually using iTMSTransporter anyway - they just need an existing folder as it's path and are happy.

(This will change in the future when we get further with the tests that are actually executed on Ubuntu, and then we will indeed have to find a solution for that of course.)

.circleci/Dockerfile
milch: I believe for fastlane, it makes no difference, since it is language/region independent. 

IIRC the regular C locale is just ASCII as it is used by the C language, C.UTF-8 is therefore ASCII extended by UTF-8. I think generally both en_US.UTF-8 (or any other two letter language and two letter region combo) and C.UTF-8 should work for us. It is simply what is installed by default on Debian/Ubuntu, so rather than installing a separate locale I chose that one

.circleci/Dockerfile
milch: Each `FROM` line in the Dockerfile essentially creates separate VMs (called stages), so this line is necessary to copy over the folder from the other stage.

Having separate "stages" is generally a good idea because it has implications on final image size and the layer cache. Each line in the Dockerfile for each of the stages is essentially a separate layer in the virtual file system. If this was all a single stage, updating iTMS would mean that essentially: 

1. The whole cache would need to be discarded
    - XAR would also need to be rebuilt 
2. Most of the final image would change completely, which would mean uploading hundreds of MB on a rebuild

It is generally a good idea to keep the amount of layers to a minimum - which is also why you see the pattern of having one `RUN` statement with every command joined by `&&` as opposed to having them in separate `RUN` statements. Anything that you clean up within one such statement won't result in increased image size. 

It also saves some space to discard any temporary directories or dependencies that are not needed anymore after building by limiting them to their own stage. For example, to compile `xar` we need to have a complete compiler collection installed (hundreds of MB), whereas to actually install it we only need `make` (a couple MB). The final image created by this is indeed only about 180MB larger than its' base, most of which is because of Java. With Java disabled it's only 14MB. 

Finally, when the image changes and circle (or any other client using it) pulls it, only the changed/new layers will need to be pulled, so the more layers can be cached the better. 

Which also reminds me, I should move the `COPY` lines to the end. This means that if ITMS or XAR is updated all the dependencies won't have to be reinstalled

milch:
> As I wrote, iTMS is actually _not_ installed in the current setup as it was not clear how I should handle a) hosting the file and b) the licence acceptance stuff (` sed -e '15,43d' -i install_script.sh` just deletes the line that are responsible for that...) and c) the licence/terms situation for iTMSTransporter in general.

@KrauseFx Can you work out these licensing issues that Jan brought up (whether we can accept automatically in the script like that) with the Google lawyers? Note that as I describe later, this would only need to be run by us so we can update the fastlane docker image, not by any user or anyone pulling the image. 

> (As I know nothing about how Docker works I thought we would build the image once locally, which would take care of problems a) and b) and only leave c) to be solved - but as this is a script now as well, this doesn't really seem to be the case. Will have to educate myself a bit here)

@janpio It does, I just included it in here so it is part of the fastlane repo. We could obviously also move it to a separate repository, so that it can automatically be rebuilt on CI, which might be a better idea anyway. In the README of that repo we can include the iTMSTransporter terms and clarify that they are accepted by the script automatically. For your `sed` line, that could in that case probably be replaced by using the `yes` command 

> So it might be better to get rid of that here as well for now, the tests aren't actually using iTMSTransporter anyway - they just need an existing folder as it's path and are happy.

I'll just comment it out for now. Once we've got the licensing issues addressed we can uncomment those lines.

milch:
Alright I removed the Dockerfile from here and put it into a separate repo. That also allows automated builds to run on [Docker Hub](https://hub.docker.com). 

I created a fastlanetools org there (fastlane was taken) and set up the automatic builds to point to my [milch/fastlane-docker](https://github.com/milch/fastlane-docker) repo for now. Once the config is ready we can just move that repo into the fastlane org to fastlane/docker or something

.circleci/Dockerfile
janpio: During the transition to Ubuntu I tried `en_US.` and got error messages as I didn't specify anything at all, changing to `C.` fixed this. But `en_US` not being installed makes sense.

janpio:
> For your sed line, that could in that case probably be replaced by using the yes command

I added the `sed` as I couldn't make it work with `yes` -  but something in that direction properly applied would probably work.

----

Tests on Ubuntu are green and the build is indeed using the container @milch built:

```
Starting container fastlanetools/fastlane
  image cache not found on this host, downloading fastlanetools/fastlane
latest: Pulling from fastlanetools/fastlane
f49cf87b52c1: Already exists
7b491c575b06: Already exists
b313b08bab3b: Already exists
51d6678c3f0e: Already exists
b7e583189d20: Already exists
4e01508081ad: Already exists
7a3173838c97: Already exists
27c911efce60: Pulling fs layer
614dfcfd6425: Pulling fs layer
e50c27c0cd52: Pulling fs layer
7a54e429ad55: Pulling fs layer
0a88a94a8259: Pulling fs layer
c6000836df46: Pulling fs layer
32da2c7af650: Pulling fs layer
ff54cf1b9d1e: Pulling fs layer
33a0e8ee7d6b: Pulling fs layer
25892d4cec43: Pulling fs layer
e57901e7d380: Pulling fs layer
f42ab7bd5910: Pulling fs layer
c6000836df46: Waiting
0a88a94a8259: Waiting
32da2c7af650: Waiting
f42ab7bd5910: Waiting
25892d4cec43: Waiting
ff54cf1b9d1e: Waiting
e57901e7d380: Waiting
33a0e8ee7d6b: Waiting
27c911efce60: Verifying Checksum
27c911efce60: Download complete
e50c27c0cd52: Download complete
27c911efce60: Pull complete
7a54e429ad55: Verifying Checksum
7a54e429ad55: Download complete
c6000836df46: Verifying Checksum
c6000836df46: Download complete
32da2c7af650: Verifying Checksum
32da2c7af650: Download complete
ff54cf1b9d1e: Verifying Checksum
ff54cf1b9d1e: Download complete
0a88a94a8259: Download complete
614dfcfd6425: Verifying Checksum
614dfcfd6425: Download complete
e57901e7d380: Verifying Checksum
e57901e7d380: Download complete
f42ab7bd5910: Download complete
25892d4cec43: Verifying Checksum
25892d4cec43: Download complete
33a0e8ee7d6b: Verifying Checksum
33a0e8ee7d6b: Download complete
614dfcfd6425: Pull complete
e50c27c0cd52: Pull complete
7a54e429ad55: Pull complete
0a88a94a8259: Pull complete
c6000836df46: Pull complete
32da2c7af650: Pull complete
ff54cf1b9d1e: Pull complete
33a0e8ee7d6b: Pull complete
25892d4cec43: Pull complete
e57901e7d380: Pull complete
f42ab7bd5910: Pull complete
Digest: sha256:2d7fc8b57f7418d553c9f3cc9bb6bd82a17ceaeb5a1e4b9d041b30b9407d17d9
Status: Downloaded newer image for fastlanetools/fastlane:latest
  using image fastlanetools/fastlane@sha256:2d7fc8b57f7418d553c9f3cc9bb6bd82a17ceaeb5a1e4b9d041b30b9407d17d9
```

The "Install xar" step is also gone, so this look perfect!

janpio (approved):

janpio:
Oh wait, the `danger` steps are failing because of a missing ENV variable:

```
bundle exec danger || echo "danger failed"

Could not set up API to Code Review site for Danger

For your GitHub repo, you need to expose: DANGER_GITHUB_API_TOKEN
You may also need: DANGER_GITHUB_HOST, DANGER_GITHUB_API_BASE_URL
```

Why? Is this normal in all PRs?

milch:
Hmm weird, it seems the API token is set if you check the Circle config: https://circleci.com/gh/fastlane/fastlane/edit#env-vars

janpio:
Same problem on the macOS run.
Test runs of other branches don't have that problem though.

Maybe because you are merging from /milch/fastlane? I always work with branches in fastlane/fastlane...

janpio:
Yep, probably this setting on https://circleci.com/gh/fastlane/fastlane/edit#advanced-settings: 
> Pass secrets to builds from forked pull requests
> Run builds for fork pull request changes with this project's configuration, environment variables, and secrets.

It is set to off, as it should be.

sshaw:
>> For your sed line, that could in that case probably be replaced by using the yes command

> I added the sed as I couldn't make it work with yes - but something in that direction properly applied would probably work.

Try this (untested), it will cause `more` to skip the first `9999...`  lines:

```shell
yes | MORE=-99999999 ./install_script.sh
```

And if that doesn't work (maybe something else is reading from `stdin`), try this: 

```shell
echo -e "y\ny\n" | MORE=-99999999 ./install_script.sh
```

milch:
@janpio @KrauseFx Thoughts? Should we merge this?

milch:
@sshaw I ended up doing something similar, i.e.

```bash
cat License.txt
yes | MORE=-V ./install_script.sh
```

Unfortunately `more` grabs the stdin whether it displays lines or not, so the only way to circumvent this was to have it output version information with `-V` and show the licence separately. We don't need it right now anyway though, because we have to clear whether this is actually ok to do first

janpio:
✅ from my side.

milch:
Alright, rebased master, will merge this once tests pass

sshaw:
So there were no concerns about packaging `iTMSTransporter` in the Docker image?  The license states: 

> Licensee may not modify, sublicense, rent, lease, lend or distribute the Apple Software and may not create derivative works based upon the Apple Software.

Just curious ⚖

janpio:
There is no iTMSTransporter in the Docker image at all. `FASTLANE_ITUNES_TRANSPORTER_PATH` points to an unrelated directory, which is enough for now. The tests that actually use iTMSTransporter are skipped on all non-Mac platforms.

We will tackle this later, probably by finding a better way to install it or finding out if this licence actually only applies to the final software, and not the installation package, which would be a valid and logical thing to think - so better ask a lawyer if logic applies ;)

PR #11333, opened 2017-12-22
Flatten application_id in match

taquitos:
- workaround for #11324
Sometimes it's an array of arrays ¯\\_(ツ)_/¯


taquitos:
As noted in #11324 
As noted: 
It's just a temporary fix. A better fix is to address ConfigItems that have parameters with multiple types , for example, 'app_identifier' can be a String or Array.

I want to unblock folks though, so I'm noting it in the code and on this issue. I wouldn't normally do this since I believe in fixing it right, the first time, but I just don't want to leave people hanging.

KrauseFx (approved):

PR #11334, opened 2017-12-23
Enable rubocop cache on Circle

milch:
Rubocop has a cache which is used to speed up repeat processing of unchanged files. See their [manual entry](https://github.com/bbatsov/rubocop/blob/master/manual/caching.md) on caching.

From there:

> Later runs will be able to retrieve this information and present the
stored information instead of inspecting the file again. This will be
done if the cache for the file is still valid, which it is if there
are no changes in:
> * the contents of the inspected file
> * RuboCop configuration for the file
> * the options given to `rubocop`, with some exceptions that have no
>   bearing on which offenses are reported
> * the Ruby version used to invoke `rubocop`
> * version of the `rubocop` program (or to be precise, anything in the
>   source code of the invoked `rubocop` program)

So this cache should be easily reusable even between different branches, and would hopefully speed up Circle runs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,8 +14,6 @@ jobs:
 
       - restore_cache:
           key: v1-macos-gems-{{ checksum "Gemfile" }}
-      - restore_cache:
-          key: v1-{{ arch }}-rubocop
       - run:
           name: Setup Build
           command: |
@@ -32,6 +30,9 @@ jobs:
       - run:
           name: Check PR Metadata
           command: bundle exec danger || echo "danger failed"
+
+      - restore_cache:
+          key: v1-{{ arch }}-rubocop
 
       - run: bundle exec fastlane test
 
@@ -63,8 +64,6 @@ jobs:
 
       - restore_cache:
           key: v1-ubuntu-gems-{{ checksum "Gemfile" }}
-      - restore_cache:
-          key: v1-{{ arch }}-rubocop
       - run:
           name: Setup Build
           command: |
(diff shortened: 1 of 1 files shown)


milch:
Looking at the Ubuntu tests it seems like it saved around 40s of build time

.circleci/config.yml
```diff
@@ -14,6 +14,8 @@ jobs:
 
       - restore_cache:
           key: v1-macos-gems-{{ checksum "Gemfile" }}
+      - restore_cache:
```
janpio: Move that down to before the test run maybe? No need to do it this early.

janpio (approved):
Looks good, see comment for possible change but is good to merge anyway.

.circleci/config.yml
milch: Possible, although I put it next to the other restore_cache for consistency

janpio: Thought behind that placement was that it "wraps" the step that is creating/using that cache.

milch: Hmm, good point. I changed it!

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             echo "2.3" > .ruby-version
             gem update --system
             brew install shellcheck
-            bundle check || bundle install --jobs=4 --retry=3 --path .bundle
+            bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
```
janpio: meaning `bundle check` could never have worked with the `--path` used for `bundle install`?

milch: Yeah I don't think so. That's why it would seemingly always return "Bundler can't satisfy dependencies"

jetersen (approved):
Awesomesauce :rocket:

.circleci/config.yml
janpio: Ok, Probably got broken in the transition from Circle CI 1.0 to 2.0. My guess is `--path .bundle` was added to the install but not the check.

Great you caught that!

PR #11339, opened 2017-12-24
Use provisioning profile name instead of UUID for .mobileprovision

Panajev:
### Checklist

### Motivation and Context

By default fastlane match generates the profiles with a descriptive name with Profile type (App Store, Enterprise, Development), and bundle id. 
Whenever profiles are recreated, which can happen for a multitude of reasons, such as adding new devices to the portal, a new entry appears inside the user's ~/"Library/MobileDevices/Provisioning Profiles" folder as each profile will have a different filename currently based on the profile UUID. With enough applications and enough variations of the profiles (One team for InHouse Distribution can have its enterprise, adhoc, and development profiles) which may easily lead to a folder growing a fair bit each time.

For the same reasons we use full manual signing with match or with our own match like setups, being able to have a single copy of a certain provisioning profile and not having to rely on Xcode or other tools to use the latest profiles created with all the devices you have just added, for example, it would be nice to have clear names for the profiles we install on our computer and replace the older one with the contents of the new one we have recreated. 

I am not sure if this Xcode behaviour was always there or if it occurred when the support for Provisioning Profile names instead of their UUID in your Xcode project signing settings landed (Xcode 8.x), but I have noticed that simply copying the files I downloaded from the Developer Portal in my ~/"Library/MobileDevices/Provisioning Profiles" folder ended up working just fine from Xcode's point of view. I also do realise that fastlane is a tool used by people that not only may have an appreciation for automation and DevOps-y stuff in general, but that have scaling needs (tons of apps, several Apple Developer Teams, mixing company and personal work, etc...).

This pull requests aims to have match copy the filename it has created as is, but adding a factor of uniqueness that should satisfy the needs we have. This would make the apparent structure of the final destination filename to be the following: 

> **<Apple_Dev_Team_ID>\_**<InHouse/Development/AppStore>\_<App_Bundle_ID>.mobileprovision

Part in bold is what this PR adds, the rest is the name match was already using.

The filename is calculated with the Team ID and the original name Fastlane created for the profile, helping the profile be unique (containing the risk of collisions), but not in such a way that we are providing protection from collision from itself, i.e. we allow by design the new profile name to be equal to the old profile name for the same Team ID, the same profile type, and the same bundle ID as before and we delete the older file if present before copying the new one.
In order to prevent duplicates, i.e. if the sync process installs a profiles that is already installed and has not been regenerated just with the new name instead of the old UUID style based name, the old style filename is calculated and if the file is available on the destination folder it is deleted before copying the new file.

In order to improve backward compatibility, this behaviour is only activated when either the **FASTLANE_MOBILE_PROVISION_FILES_NAMES_MODE** environment variable is set or when the **-m** flag is passed to fastlane match.

### Testing

Ran the unit tests, followed the instructions in https://github.com/fastlane/fastlane/blob/master/YourFirstPR.md#test-the-changes-for-your-application and used the local fastlane modified version  to sync all the provisioning profiles from the App Store, Enterprise, and Development provisioning profiles types 
```bash
#!/bin/bash

BUNDLE_IDS=("com.company.corp.app1" "com.company.corp.app2" "com.company.corp.app3" "com.company.corp.app4" "com.company.corp.app5" "com.company.corp.app6" "com.company.corp.app7" "com.company.corp.app8" "com.company.corp.app9" "com.company.corp.app10" "com.company.corp.app11" "com.company.corp.app12" "com.company.corp.app13" "com.company.corp.app14" "com.company.corp.app15")

if [ "${1}" == "PROFILES_NAME" ]; then
	optionProfiles="${1}"
	echo "optionProfiles = ${optionProfiles}"
	fastlane match development --readonly -a $BUNDLE_IDS --team_name "<CompanyName>" --skip_docs -m ${optionProfiles}
else
	fastlane match development --readonly -a $BUNDLE_IDS --team_name "<Company Name>" --skip_docs
fi

[...]
```

```ruby
bundle exec rspec Match
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
Changing stdout to /tmp/fastlane_tests, set `DEBUG` environment variable to print to stdout (e.g. when using `pry`)

[...]

Spaceship::Tunes::Members
  members
    should return a list with members
    finds one member by email
    creates a new member
      role: admin, apps: all
      role: developer apps: all
      role: appmanager, apps: 898536088
    updates roles and apps for an existing member
      role: admin, apps: all
      role: developer apps: all
      role: appmanager, apps: 898536088

Spaceship::Tunes::Territory
  supported_territories
    inspect works
    correctly creates all territories
    correctly parses the territories

Spaceship::Tunes::SandboxTester
  Sandbox testers
    listing
      loads sandbox testers correctly
    creation
      creates sandbox testers correctly
    deletion
      deletes a user
      deletes all users

Spaceship::TunesClient
  #login
    raises an exception if authentication failed
  client
    exposes the session cookie
  Logged in
    stores the username
    #hostname
    #handle_itc_response
      raises an exception if something goes wrong
      does nothing if everything works as expected and returns the original data
      identifies try again later responses
  CI
    crashes when running in non-interactive shell

Spaceship::Tunes
  InsufficientPermissions
    raises an appropriate iTunes Connect error when user doesn't have enough permission to do something

Supply
  Supply::Client
    AndroidPublisher
      has all the expected Google API methods

Supply::CommandsGenerator
  :run options handling
    can use the skip_upload_metadata flag from tool options
  :init options handling
    can use the package_name short flag from tool options

Supply
  Supply::Uploader
    #find_obbs
      finds no obb when there's none to find
      skips unrecognized obbs
      finds one match and one patch obb
      finds zero obb if too main mains
      finds zero obb if too many patches
    metadata encoding
      prints a user friendly error message if metadata is not UTF-8 encoded
    all_languages
      only grabs directories
    check superseded tracks
      remove lesser than the greatest of any later (i.e. production) track
      remove lesser than the currently being uploaded if it is in an earlier (i.e. alpha) track
      combined case

Finished in 5 minutes 55 seconds (files took 6.12 seconds to load)
4745 examples, 0 failures
```

Robocop detected no issues:
```bash
bundle exec rubocop -a
Inspecting 1018 files
................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................./Users/goff.marocchi/Programming/workspace/gitGamesys/GitHub/Signing/fastlaneSigning/fastlane/spaceship/lib/spaceship/ui.rb: Naming/MethodName has the wrong namespace - should be Style
/Users/goff.marocchi/Programming/workspace/gitGamesys/GitHub/Signing/fastlaneSigning/fastlane/spaceship/lib/spaceship/ui.rb: Naming/MethodName has the wrong namespace - should be Style
.........................................................................................................................

1018 files inspected, no offenses detected
```

When running all tests on my machine I have 4 failures in portions of the code I have not touched and also occur if I do a clean clone of the official fastlane repo's master branch, install dependencies with bundler, and execute the tests. These are the results:
```bash
[...]

Failures:

  1) FastlaneCore::TagVersion correct? returns true for versions supported by Gem::Version
     Failure/Error: expect(FastlaneCore::TagVersion.correct?(tag)).to be true

       expected true
            got #<Fixnum:1> => 0
     # ./fastlane_core/spec/tag_version_spec.rb:20:in `block (3 levels) in <top (required)>'

  2) FastlaneCore::TagVersion correct? returns true for tags that can be converted to a Gem::Version using version_number_from_tag
     Failure/Error: expect(FastlaneCore::TagVersion.correct?(tag)).to be true

       expected true
            got #<Fixnum:1> => 0
     # ./fastlane_core/spec/tag_version_spec.rb:25:in `block (3 levels) in <top (required)>'

  3) FastlaneCore::TagVersion correct? returns false for tags that are not versions
     Failure/Error: expect(FastlaneCore::TagVersion.correct?(tag)).to be false

       expected false
            got nil
     # ./fastlane_core/spec/tag_version_spec.rb:30:in `block (3 levels) in <top (required)>'

Finished in 4 minutes 41.1 seconds (files took 2.83 seconds to load)
4735 examples, 3 failures

Failed examples:

rspec ./fastlane_core/spec/tag_version_spec.rb:18 # FastlaneCore::TagVersion correct? returns true for versions supported by Gem::Version
rspec ./fastlane_core/spec/tag_version_spec.rb:23 # FastlaneCore::TagVersion correct? returns true for tags that canbe converted to a Gem::Version using version_number_from_tag
rspec ./fastlane_core/spec/tag_version_spec.rb:28 # FastlaneCore::TagVersion correct? returns false for tags that are not versions
```

This is my gem env
```bash
RubyGems Environment:
  - RUBYGEMS VERSION: 2.6.14
  - RUBY VERSION: 2.3.3 (2016-11-21 patchlevel 222) [x86_64-darwin16]
  - INSTALLATION DIRECTORY: /Users/goff.marocchi/.rvm/gems/ruby-2.3.3
  - USER INSTALLATION DIRECTORY: /Users/goff.marocchi/.gem/ruby/2.3.0
  - RUBY EXECUTABLE: /Users/goff.marocchi/.rvm/rubies/ruby-2.3.3/bin/ruby
  - EXECUTABLE DIRECTORY: /Users/goff.marocchi/.rvm/gems/ruby-2.3.3/bin
  - SPEC CACHE DIRECTORY: /Users/goff.marocchi/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /Users/goff.marocchi/.rvm/rubies/ruby-2.3.3/etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-darwin-16
  - GEM PATHS:
     - /Users/goff.marocchi/.rvm/gems/ruby-2.3.3
     - /Users/goff.marocchi/.rvm/gems/ruby-2.3.3@global
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /Users/goff.marocchi/.rvm/gems/ruby-2.3.3/bin
     - /Users/goff.marocchi/.rvm/gems/ruby-2.3.3@global/bin
```

### Changes Description

The major change in the code is really in the file that manages the provisioning profiles (provisioning_profiles.rb in fastlane_core):
```ruby
      # @return [String] The UUID of the given provisioning profile
      def uuid(path)
        parse(path).fetch("UUID")
      end

      # @return [String] The Name of the given provisioning profile
      def name(path)
        parse(path).fetch("Name")
      end

      # @return [String] The Apple TeamIdentifier of the given provisioning profile
      def team_identifier(path)
        parse(path).fetch("Entitlements").fetch("com.apple.developer.team-identifier")
      end

      def profiles_path
        path = File.expand_path("~") + "/Library/MobileDevice/Provisioning Profiles/"
        # If the directory doesn't exist, create it first
        unless File.directory?(path)
          FileUtils.mkdir_p(path)
        end

        return path
      end

      # Installs a provisioning profile for Xcode to use
      def install(path, provisioning_profiles_mode_option = nil)
        UI.message("Installing provisioning profile...")
        profile_filename = uuid(path) + ".mobileprovision"
        destination = File.join(profiles_path, profile_filename)

        if path != destination

          if ENV['FASTLANE_MOBILE_PROVISION_FILES_NAMES_MODE'] == "PROFILES_NAME" || provisioning_profiles_mode_option == "PROFILES_NAME"
            # Remove previous generation provisioning profile name
            #  to avoid installing the same file twice with different file names
            FileUtils.remove_entry(destination, force: true)

            # Copy to Xcode provisioning profile directory and delete previous file if present
            profile_filename = team_identifier(path) + "_" + File.basename(path)
            destination = File.join(profiles_path, profile_filename)

            FileUtils.remove_entry(destination, force: true)
            FileUtils.copy(path, destination)
          else
            FileUtils.copy(path, destination)
            unless File.exist?(destination)
              UI.user_error!("Failed installation of provisioning profile at location: '#{destination}'")
            end
          end
        end
```

The filename is calculated with the Team ID and the original name Fastlane created for the profile, helping the profile be unique (containing the risk of collisions), but not in such a way that we are providing protection from collision from itself, i.e. we allow by design the new profile name to be equal to the old profile name for the same Team ID, the same profile type, and the same bundle ID as before and we delete the older file if present before copying the new one.
In order to prevent duplicates, i.e. if the sync process installs a profiles that is already installed and has not been regenerated just with the new name instead of the old UUID style based name, the old style filename is calculated and if the file is available on the destination folder it is deleted before copying the new file.

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -44,6 +44,11 @@ module FastlaneCore
         parse(path).fetch("Name")
       end
 
+      # @return [String] The Apple TeamIdentifier of the given provisioning profile
+      def team_identifier(path)
+        parse(path).fetch("Entitlements").fetch("com.apple.developer.team-identifier")
+      end
+
       def profiles_path
         path = File.expand_path("~") + "/Library/MobileDevice/Provisioning Profiles/"
         # If the directory doesn't exist, create it first
@@ -55,16 +60,29 @@ module FastlaneCore
       end
 
       # Installs a provisioning profile for Xcode to use
-      def install(path)
+      def install(path, provisioning_profiles_mode_option = nil)
         UI.message("Installing provisioning profile...")
         profile_filename = uuid(path) + ".mobileprovision"
         destination = File.join(profiles_path, profile_filename)
 
         if path != destination
-          # copy to Xcode provisioning profile directory
-          FileUtils.copy(path, destination)
-          unless File.exist?(destination)
-            UI.user_error!("Failed installation of provisioning profile at location: '#{destination}'")
+
+          if ENV['FASTLANE_MOBILE_PROVISION_FILES_NAMES_MODE'] == "PROFILES_NAME" || provisioning_profiles_mode_option == "PROFILES_NAME"
+            # Remove previous generation provisioning profile name
+            #  to avoid installing the same file twice with different file names
+            FileUtils.remove_entry(destination, force: true)
+
+            # Copy to Xcode provisioning profile directory and delete previous file if present
+            profile_filename = team_identifier(path) + "_" + File.basename(path)
+            destination = File.join(profiles_path, profile_filename)
+
+            FileUtils.remove_entry(destination, force: true)
+            FileUtils.copy(path, destination)
+          else
+            FileUtils.copy(path, destination)
+            unless File.exist?(destination)
+              UI.user_error!("Failed installation of provisioning profile at location: '#{destination}'")
+            end
           end
         end
 
--- a/match/lib/match/options.rb
+++ b/match/lib/match/options.rb
@@ -154,7 +154,12 @@ module Match
                                      env_name: "MATCH_PROVISIONING_PROFILE_TEMPLATE_NAME",
                                      description: "The name of provisioning profile template. If the developer account has provisioning profile templates, template name can be found by inspecting the Entitlements drop-down while creating/editing a provisioning profile",
                                      optional: true,
-                                     default_value: nil)
+                                     default_value: nil),
+        FastlaneCore::ConfigItem.new(key: :provisioning_profile_names_mode,
+                                     short_option: "-m",
+                                     env_name: "FASTLANE_MOBILE_PROVISION_FILES_NAMES_MODE",
+                                     description: "Using regular UUID based names for provisioning profile names (default value for the option is 'UUID') or use a name that is unique, but will reduce the number of profiles stored when they are recreated (provide the following value for the option 'PROFILES_NAME')",
+                                     default_value: "UUID")
       ]
     end
   end
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -167,7 +167,7 @@ module Match
         self.files_to_commmit << profile
       end
 
-      installed_profile = FastlaneCore::ProvisioningProfile.install(profile)
+      installed_profile = FastlaneCore::ProvisioningProfile.install(profile, params[:provisioning_profile_names_mode])
 
       parsed = FastlaneCore::ProvisioningProfile.parse(profile)
       uuid = parsed["UUID"]
--- a/match/spec/runner_spec.rb
+++ b/match/spec/runner_spec.rb
@@ -6,6 +6,7 @@ describe Match do
       allow(ENV).to receive(:[]).and_call_original
       allow(ENV).to receive(:[]).with('MATCH_KEYCHAIN_NAME').and_return(keychain)
       allow(ENV).to receive(:[]).with('MATCH_KEYCHAIN_PASSWORD').and_return(nil)
+      allow(ENV).to receive(:[]).with('FASTLANE_MOBILE_PROVISION_FILES_NAMES_MODE').and_return('UUID')
     end
 
     it "creates a new profile and certificate if it doesn't exist yet", requires_security: true do
@@ -29,7 +30,7 @@ describe Match do
                                                                             prov_type: :appstore,
                                                                        certificate_id: "something",
                                                                        app_identifier: values[:app_identifier]).and_return(profile_path)
-      expect(FastlaneCore::ProvisioningProfile).to receive(:install).with(profile_path).and_return(destination)
+      expect(FastlaneCore::ProvisioningProfile).to receive(:install).with(profile_path, "UUID").and_return(destination)
       expect(Match::GitHelper).to receive(:commit_changes).with(
         repo_dir,
         "[fastlane] Updated appstore and platform ios",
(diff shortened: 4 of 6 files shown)


KrauseFx:
Is this gonna be fully backwards compatible for all users already using fastlane match?

Panajev:
I would imagine that for users that have match already, like it happened with the test profiles the unit tests install, nothing would change although they would have the same exact provisioning profile twice with the same effective UUID (unless they are recreating the profiles in which case they would have the profiles with the new name in addition to the old ones... you could take the new name as another different identifier like the UUID was).

Would you think that adding a check where if the same provisioning profile already existed with the old <UUID>.mobileprovision name (instead of <team id>_<buildType>_<bundle id>.mobileprovision one) we delete it? 

We can compute how the old name would have looked and delete the file before copying the new one... that would not be destructive and would avoid Xcode from having two copies of the same file with same contents but just a different file name and creation date. 

Do you have other backwards compatibility checks in mind?

Panajev:
Hey @KrauseFx, I have added a change to address backwards compatibility concerns, let me know if I am still away from the mark, as I was writing in my last comment and rebased on top of master. Thank you for the feedback :).

KrauseFx (changes requested):
Hey @Panajev, thanks for your pull request, really appreciate. I don't fully understand the benefit of this yet. Please note that this change is a major change on how _match_ works and it involves quite some risk breaking existing and new setups. Those kind of changes we usually try to avoid to make sure things are super solid.

I don't fully understand why having more provisioning profiles in the local profiles folder is an issue. Does Xcode select the wrong profile? If so, you can specify the one to use using _fastlane_. Or does it slow down the build process?

Panajev:
Hey @KrauseFX, no worries... I did not think you just walk a patch into fastlane... or was it walking into Mordor you could not just easily do? Never mind that ;).

#Backwards compatibility musings
Would it help if this behaviour were hidden behind an environment variable, behaviour added to the docs, and letting this be battle tested (after some more review comments here of course if more eyes want to offer comments and suggest other potential issues I can address)? The change set is small enough that I could reasonably keep rebasing it on top of master, but I would rather be able to learn enough and improve this PR to land it into fastlane :D. I would like to help mitigate risks beyond shielding them behind an env variable (which would provide the old behaviour of the variable is not set)... might need some help with the unit tests and testing of this environment variable, but it is fine :).

#Additional Risk Mitigation
Could you and/or other core maintainers please expand on the risks so I can have a bit of a think and try my best to address them (maybe over time if the environment variable approach is acceptable) mitigating risks as much as possible?
 I am not sure how my PR could break existing setups the way it is and I do not think there is a likely chance of files collisions, but my lack of experience with fastlane core as well as the wildly different setups people may have would make my assessment not as solid as I think it is.

##Rational wise
We had had issues in the past where multiple versions of the same profile (or certificate in other cases) differing only by content (devices, entitlements, or other) could confuse Xcode and while I have not measured exact speed impact due to the number of profiles installed Xcode needs to do needless work to find the profiles that match the certificate and bundle id and then select the ones that matches best. I would like to have a solution that works with Xcode but does not need to track the uuid of the profile (I do not want to go back to those “dark ages” pre Xcode 8.x :)).

The spirit of fastlane, which  embraced IMHO by making it easier and better to do fully automated signing, is to automate users’ workflows with less and less dependencies on Xcode magic (hence match instead of relying on Xcode fully automated signing) and thus provide a more resilient solution: if we can avoid duplicating provisioning profile files and giving less work to Xcode there is less potential issues that can be dependent on third party code and not our fastlane/match/project setup.

KrauseFx:
Yeah, totally, if we find a good way to clean old provisioning profiles from the local machine, I'm all for it. It might be a little tricky, as we don't want to delete profiles that are still being used.

Panajev:
Hey @KrauseFx , so... I agree in the same sprit as this PR, cleaning up the old profiles if you decide to use the new option (uuid less) makes complete sense :). Just be patient with my questions as I still find some things here daunting hehe :).

One approach that should be safe, but maybe a bit slow, would be to read the **profile name** and **Team ID** (would profile name be enough) from the current provisioning profile that match is trying to install, then loop through the .mobileprovision files inside the MobileDevices/Provisioning Profiles folder and delete each of the profiles which have the same **profile name** and **Team ID** before installing the new profile.

> we don't want to delete profiles that are still being used
Fully agreed, but that would have to be a profile we already sync over match, a profile we are not installing a synced profile stored in the private match repo else why would it be a problem if we deleted it and replaced it with the profile we are installing at this time? 

I guess that the worry would be that some projects are using match and yet they are still specifying the provisioning profile by UUID inside Xcode or their .xcconfig files instead of using the profile display name. Is this an officially supported fastlane setup considering that 
1. Xcode supports the use of display name avoiding a lot of changes everytime profiles are recreated
2. If you were using the new uuid-less naming option for the provisioning profiles should be explicitly for people that are indeed using the profile display name in their projects.
3. In the "solution" I mentioned above we would not delete provisioning profiles you have not already migrated to fastlane match.

What do you think? I have a lingering fear of oversiplifying this, so please any feedback is welcomed! Thanks.

KrauseFx:
My feeling is is that it's not something that's worth the extra work. From my understanding, this still work nicely without this patch. Adding this logic adds quite some complexity and will probably end up causing other issues, that we don't think of yet.

Panajev:
Hey @KrauseFx , I have added the -m option to preserve backward compatibility with older setups. If you want to opt in the new behaviour introduced by this PR you would either use the -m flag or setup the corresponging environment variable (FASTLANE_MOBILE_PROVISION_FILES_NAMES_MODE).

I agree that the extra login to loop through all profiles by profile name and Team ID to remove older ones, but not if they have not been re-created with the new naming convention, would be a bit overkill compared to the benefits it brings.

Panajev:
Oh hello @fastlane-bot  :).

Panajev:
@KrauseFx rerunning the build again... sometimes I see odd issues on the Ubuntu agent... tested the build locally with ruby 2.5.0 on macOS High Sierra and the macOS agent (which uses ruby 2.3.3) also works... bit puzzled... hopefully it passes on both now...

KrauseFx:
Hey @Panajev, sorry to be the bad guy here, but I'd prefer not to go ahead with this change so far... thanks for your contributions and your ideas.

If you really really want this change, you can run and maintain your own fork, but I would recommend against it, as it will be hard to keep up. 

Again sorry we're not going ahead with this one 😢

Panajev:
@KrauseFx I understand :), no hard feelings. 

In the end for controlled environments one could go the way of deleting all the existing profiles, syncing with fastlane as normal, etc... I knew it would not be an easy thing to add and I have gotten some understanding of why this could bring problems (would like some more feedback if it is possible in terms of problems that could happen to people opting in this approach and just always using it consistently as it would give me a better understanding on the impact of my changes.... safe to learn as they say :)). 

Thanks anyways, will close the PR, but keep running the macOS test on my branch for as long as I can reasonably keep it up to date without including additional risks :).

PR #11340, opened 2017-12-25
Update Swift API generation to handle default values for Procs correctly

taquitos:
- Don't handle them, set them nil (best handling)


KrauseFx (approved):
👍

PR #11352, opened 2017-12-27
Add loading indicator for fastlane startup time

KrauseFx:
- This adds a light-weight dependeny, part of the trusted `tty` tools
- I was thinking of adding more and more loading indicators as we go, but this one is the most obvious, as for people without a `Gemfile` wait for multiple seconds here without any output at all

GitHub doesn't support video attachments unfortunately

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -44,7 +44,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'public_suffix', '~> 2.0.0' # https://github.com/fastlane/fastlane/issues/10162
 
   # TTY dependencies
-  spec.add_dependency 'tty-screen', '~> 0.6.3' # detect the terminal width
+  spec.add_dependency 'tty-screen', '>= 0.6.3', '< 1.0.0' # detect the terminal width
+  spec.add_dependency 'tty-spinner', '>= 0.7.0', '< 1.0.0' # loading indicators
 
   spec.add_dependency 'babosa', '>= 1.0.2', "< 2.0.0"
   spec.add_dependency 'colored' # colored terminal output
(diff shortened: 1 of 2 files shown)


RishabhTayal:
You can upload video to Dropbox or Google Drive and attach a link here.

dgyesbreghs (approved):

nafu (approved):
Looks great to me 🚀 

fastlane-spinner

(Created gif from mov that @KrauseFx uploaded)

KrauseFx:
Beautiful, thanks @nafu 🚀

fxfactorial:
Spinners and eye candy are nice, but ad-hoc adding all kind of things that ad-hoc change the output UI with no ENV variable to buy out of this behavior, its not a great addition IMO.

janpio:
What env variable would you expect to control this?
And why exactly?

fxfactorial:
I already run `FASTLANE_DISABLE_COLORS` and `FASTLANE_SKIP_UPDATE_CHECK` to have a better experience. Its just better as a UNIX tool to do less ad-hoc things and make programs that are easy for other programs to consume and built on top of.

janpio:
Ok, as these don't match there would have to be a new one. `FASTLANE_DISABLE_ANIMATION` maybe?

> Its just better as a UNIX tool to do less ad-hoc things and make programs that are easy for other programs to consume and build on top of.

How is the added loading indicator a problem for that goal? (Honestly curious as I didn't see this as a problem)

fxfactorial:
I think it would be even better to have like `FASTLANE_BARE_BONES`, so that CI environments could just use one environment.

KrauseFx:
I'd be ok with adding `FASTLANE_DISABLE_ANIMATION`, since we have loading indicators in a central spot anyway, we can then easily print out the raw text instead of the loading indicator if necessary. 

Is that something you'd be interested in submitting a PR for?

fxfactorial:
Sure, I know enough ruby to do that. I'll try to bang it out later today

KrauseFx:
Awesome, thanks @fxfactorial, really appreciate it 👍

Issue #11348, opened 2017-12-27
Running the fastlane tests will cause one of the fixture Xcode projects to be modified


KrauseFx:
Run
```
bundle exec rspec
```
and run 
```
git diff
```

This should be fixed by either not changing the project, or reverting those changes

pav1790:
@KrauseFx Can you see if the fix above would be okay?

KrauseFx:
Awesome, thanks @pav1790

PR #11358, opened 2017-12-27
Clean up fixture for automatic_code_signing_spec after tests are done

paramv17:
### Checklist
- ✅ I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- ✅ I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- ✅  I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- ✅  I've updated the documentation if necessary.

### Motivation and Context
This change fixes the issue raised here : https://github.com/fastlane/fastlane/issues/11348
I ran the tests by `bundle exec rspec` and checked the `git diff` and the file was no longer modified at the end of the tests.

### Description
I added an `after :all` block - which runs after the tests are done. In it, I deleted the entries added by the rspec tests. This was necessary, because the project needs to be modified in order for the successful testing. So the only way to have a clean project afterwards is to delete those entries.

### DIFF
--- a/fastlane/spec/actions_specs/automatic_code_signing_spec.rb
+++ b/fastlane/spec/actions_specs/automatic_code_signing_spec.rb
@@ -7,6 +7,16 @@ describe Fastlane do
       allow(FastlaneCore::FastlaneFolder).to receive(:path).and_return(nil)
     end
 
+    after :all do
+      project = Xcodeproj::Project.open("./fastlane/spec/fixtures/xcodeproj/automatic_code_signing.xcodeproj")
+      project.targets.each do |target|
+        target.build_configurations.each do |configuration|
+          configuration.build_settings.delete('CODE_SIGN_STYLE')
+        end
+      end
+      project.save
+    end
+
     it "enable_automatic_code_signing" do
       allow(UI).to receive(:success)
       expect(UI).to receive(:success).with("Successfully updated project settings to use ProvisioningStyle 'Automatic'")


janpio (commented):
This fixes the symptom instead of the problem, doesn't it?

pav1790:
I had the same concerns, if you see the PR description. The issue is that, this action introduces a setting which is not present beforehand in the xcodepro. Thus modification of xcodeproj is unavoidable unless, I were to rewrite the specs that test the given action - assuming it's possible to test it another way. 

Or could create another xcodeproj and just have the setting populated in there - so that it need not be deleted?

janpio:
> Or could create another xcodeproj and just have the setting populated in there - so that it need not be deleted?

I think we had similar pull requests in the last few days/weeks.

On the other hand: Why is this happening now? Is this dependant on the Xcode version installed locally?

pav1790:
The test cases follow this structure : 

1. Run action by loading the fastfile populated within the test - which specifies the test case being conducted. 
2. This would enable/disable automatic code signing - either way, this introduces a new setting in the xcodeproj the was not there before.
3. Verify correct behavior this by checking it.

This is not dependent on the Xcode version - though you do raise valid point : why did we not see it before?  Because, I see that the change that would have manifested this behavior was merged about 12 days ago. So that's when the behavior would have started to show, at the earliest. :)

pav1790:
@janpio what d u think?

janpio:
I didn't further look into this as I don't know enough about these test fixtures and signing stuff in general. Hopefully someone else can give some input.

PR #11374, opened 2017-12-29
Improve support for Ruby 2.0.0

jdee:
### Checklist

### Motivation and Context

OS X Sierra shipped Ruby 2.0.0, and lots of users may not have updated to High Sierra.

`gem install fastlane` can fail on 2.0 because `jwt` 2.x requires Ruby 2.2. This comes in through `googleauth` and `signet`, both of which can accept 1.5.6. This doesn't affect users with Gemfiles. The Bundler accounts for the Ruby version when installing. Otherwise anyone running `gem install fastlane` using Ruby 2.0 has to first run `gem install jwt -v '~> 1.5'`. This ensures the same version of `jwt` across Ruby versions.

Some `spaceship` specs fail with 2.0 because the `Array#to_h` method was introduced in 2.1.

### Description

~~Pinned `jwt` to `~> 1.5` in the gemspec.~~ This solution is problematic. There may be no good solution to guarantee that `gem install fastlane` will always succeed without a Gemfile. If there is, it can be a separate PR.

Added `fastlane_core/core_ext/array` to supply ~~the to_h method if not otherwise present~~ a method called `fl_to_h` that calls `to_h` if present and otherwise provides an implementation based on the C implementation here: http://ruby-doc.org/core-2.5.0/Array.html#method-i-to_h ("click to toggle source").

With the latter change, spaceship specs pass on 2.0.0 and 2.4.2.


fastlane.gemspec
```diff
@@ -71,6 +71,9 @@ Gem::Specification.new do |spec|
   # If you upgrade this gem, make sure to upgrade the users of it as well.
   spec.add_dependency 'google-api-client', '>= 0.13.1', '< 0.14.0' # Google API Client to access Play Publishing API
 
+  # jwt 2.x requires Ruby 2.1. This is ultimately pulled in by the google-api-client.
+  spec.add_dependency 'jwt', '~> 1.5'
```
KrauseFx: Instead of using `~>` could we use `'>= 1.5', < 3.0.0` to allow 1.5, but use a 2.x.x version if possible/available?

jdee: The requirements from googleauth and signet are already similar:

https://rubygems.org/gems/googleauth
https://rubygems.org/gems/signet

This is what happens currently:

```
[jdee@Jimmy-Dees-MacBookPro ~]$ rvm use 2.0.0
Using /Users/jdee/.rvm/gems/ruby-2.0.0-p648
[jdee@Jimmy-Dees-MacBookPro ~]$ gem uninstall jwt

You have requested to uninstall the gem:
	jwt-1.5.6

googleauth-0.6.2 depends on jwt (< 3.0, >= 1.4)
signet-0.8.1 depends on jwt (< 3.0, >= 1.5)
If you remove this gem, these dependencies will not be met.
Continue with Uninstall? [yN]  y
Successfully uninstalled jwt-1.5.6
[jdee@Jimmy-Dees-MacBookPro ~]$ gem install fastlane
Fetching: jwt-2.1.0.gem (100%)
ERROR:  Error installing fastlane:
	The last version of jwt (< 3.0, >= 1.5) to support your Ruby & RubyGems was 1.5.6. Try installing it with `gem install jwt -v 1.5.6` and then running the current command again
	jwt requires Ruby version >= 2.1. The current ruby version is 2.0.0.
```

I'm afraid the gemspec actually has to pin, which is unfortunate. I'd love to find a better way to do it.

jdee: Now I'm skeptical about this filter on the `jwt` version. The problem is that someone may have a newer version of `jwt` installed with Ruby >= 2.1. Then with this `gem install fastlane` may fail because RubyGems will refuse to active 1.5.6 because a newer version is already active.

There may be no graceful solution to the `gem install` issue. It may be a matter of documenting:

1. Use Ruby >= 2.1, preferably with a version manager.
2. Use a Gemfile whenever possible.
3. If you can't do either of these things, then with 2.0, run `gem install jwt -v 1.5.6` before `gem install fastlane`.

I think 2.0 is just becoming harder to support.

janpio:
Semi OT: Should we start testing different ruby versions? Shouldn't be too hard on Circle...

KrauseFx:
@janpio I don't think that's possible without adding significant time to each test run 😢

jdee:
The gemspec says Fastlane supports `>= 2.0.0`, but I don't imagine there's need to test every single version. The most important ones are probably 2.0.0, shipped with both El Capitan and Sierra and 2.3.3, shipped with High Sierra. @janpio On Linux and Windows what versions are you using? Are you using a system Ruby in each case or a version manager?

janpio:
I think we are testing 2.3.x on both macOS and Ubuntu:
https://github.com/fastlane/fastlane/blob/master/.circleci/config.yml#L21
https://github.com/fastlane/fastlane/blob/master/.circleci/config.yml#L61

For Windows I locally am using 2.4.3 right now (`ruby 2.4.3p205 (2017-12-14 revision 61247) [x64-mingw32]` to be exact), AppVeyor (Windows CI) is proposed as `Ruby 2.4.2-2 x64`:
https://github.com/fastlane/fastlane/pull/11388/files#diff-180360612c6b8c4ed830919bbb4dd459R11

milch:
Can't we drop 2.0.0 at some point 😢 So many fixes in between then and now...

jdee:
I guess we can drop 2.0.0 once we can drop the system Ruby on macOS Sierra. Since High Sierra only shipped a few months ago, seems like 2.0.0 is necessary for a while.

High Sierra ships 2.3.3. I get 2.3.1 from `apt-get` on Ubuntu 16.04, the LTS release. This is the version used for both operating systems in the `.circleci/config.yml`. Maybe we should target 2.3 on Windows as well for consistency.

It's likely that devs will use later versions, so it will probably just work on 2.4 and 2.5. But since devs aren't likely to use 2.0, CI for 2.0 is probably valuable until we can ignore Sierra. If it's possible to set up a parallel job, that seems like a good idea.

KrauseFx (approved):
Let's try shipping this?

trusk89:
@KrauseFx should we expect a release in the next few hours?

KrauseFx:
Maybe :) I'll try my best

trusk89:
@KrauseFx You da real mvp!

spaceship/lib/spaceship/tunes/iap_list.rb
```diff
@@ -64,6 +66,22 @@ def edit
       def delete!
         client.delete_iap!(app_id: application.apple_id, purchase_id: self.purchase_id)
       end
+
+      private
+
+      def build_iap
```
KrauseFx: This method doesn't exist or isn't used, am I missing something?

jdee: This diff looks strange. I see you pushed some commits from master. I didn't realize you could push commits to my fork. TIL. I have a feeling somehow that's affecting the diff. There were three sets of changes here: 1. Pin jwt dependency in gemspec and then revert that change. 2. Accidentally committed a fixture that gets modified in unit tests and then reverted that. 3. (The only diff that should remain) Add the core_ext for Array with the fl_to_h method. Require that here at the top and then call it below. Not sure why it's showing up this way, but I think that's what you see on the commit before this.

KrauseFx: Yah, something seems wrong here. I resolved some merge conflicts, maybe it happened because of this?

trusk89:
No update on the merge?

KrauseFx:
@trusk89 even with this PR, it probably wouldn't fix the issue you're having. Could you try a different installation method? Either `brew cask install fastlane`, or upgrade your Ruby installation.

trusk89:
@KrauseFx for now I've added jwt 1.5.6 to my gemfile as a workaround, in all my project, but I can't update the ruby installation jenkins is using https://stackoverflow.com/questions/48153140/macos-jenkins-uses-system-ruby-instead-of-installed-ruby

joshdholtz:
@jdee @KrauseFx What do we want to do about this PR? Is this something we still want to try to merge 🤔

janpio:
This PR in its current state doesn't make any sense any more. 

It adds a `Array.fl_to_h` that is then used only in some methods that are also added in this PR. Looking at the individual commits, `to_h` used to be used in the file and was replaced with `fl_to_h`, but later changes on that file got rid of that usage.

If 2.0.0 compat is still an issue, we can look at the individual commits again later.

Issue #11356, opened 2017-12-27
GYM output filename is not correct.


taboulot:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I recently update fastlane to the last version (2.71.0).

It seems that gym doesn't create an IPA with the right GYM_OUTPUT_NAME.
If I have this in my .env file :
```
GYM_OUTPUT_NAME='app.ipa'
IOS_IPA_PATH='./dist/app.ipa'
```
Then, the output file is named : app.api.api and the last version is not deploy on hockey app.

I have the impression that '.ipa' is no longer remove from the GYM_OUTPUT_NAME when gym generate the IPA file of my app.

### Environment

```
| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.13.2                                     |
| Ruby                        | 2.3.4                                       |
| Bundler?                    | true                                        |
| Git                         | git version 2.14.3 (Apple Git-98)           |
| Installation Source         | ~/.rvm/gems/ruby-2.3.4/bin/fastlane         |
| Host                        | Mac OS X 10.13.2 (17C88)                    |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.3.4/lib                |
| OpenSSL Version             | OpenSSL 1.0.2l  25 May 2017                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 9.2                                         |


### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_GB.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
| fastlane | 2.71.0  | ✅ Up-To-Date  |


### Loaded fastlane plugins:

| Plugin                                 | Version | Update-Status |
| -------------------------------------- | ------- | ------------- |
[...]

PR #11377, opened 2017-12-29
Fix gym `GYM_OUTPUT_NAME` with `.ipa` extension causes duplicate `.ipa` extension

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/11356


jdee (approved):
:+1:

PR #11382, opened 2017-12-31
Add Helper.is_windows?

janpio:
Adds a convenience method to check if we are on Windows

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -120,6 +120,10 @@ module FastlaneCore
       self.mac?
     end
 
+    def self.is_windows?
+      self.windows?
+    end
+
     # Do we want to disable the colored output?
     def self.colors_disabled?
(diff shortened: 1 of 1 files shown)


hjanuschka (approved):
🏠

KrauseFx (approved):
👍

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -120,6 +120,10 @@ def self.is_mac?
       self.mac?
     end
 
+    def self.is_windows?
+      self.windows?
```
giginet: you can also use `alias_method`

giginet (approved):

fastlane_core/lib/fastlane_core/helper.rb
KrauseFx: Oh, great idea, and we could use the same for other methods also, like `is_mac?`, etc., right? :)

PR #11385, opened 2017-12-31
Work on copy of variable so mocks don't get retroactively changed

janpio:
This helper figures out a list of locations to look for a file.
The variable it changes, was used to specify a mock in a test before - so this can mess with stuff.
Don't change the variable, but create a copy that is used in here.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -206,9 +206,8 @@ module FastlaneCore
       #
       # We also try to append `-db` at the end of the file path, as with Sierra the default Keychain name
       # has changed for some users: https://github.com/fastlane/fastlane/issues/5649
-      #
 
-      # Remove the ".keychain" at the end of the name
+      # Remove the ".keychain" at the end of the keychain name
       name = keychain_name.sub(/\.keychain$/, "")
 
       possible_locations = [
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -209,7 +209,7 @@ def self.keychain_path(name)
       #
 
       # Remove the ".keychain" at the end of the name
-      name.sub!(/\.keychain$/, "")
+      name = name.sub(/\.keychain$/, "")
```
hjanuschka: ahh i see the prev. code changed the reference variable.
however, how about changing the local name of `name`  instead of turning the input variable into a local variable?

janpio: Suggestion?

KrauseFx: Yeah, just using a new variable instead of assigning it to the same one should solve the issue 👍

hjanuschka: `local_name`, `keychain_name`  i am not 100% sure 😵

janpio: I renamed the method param to `keychain_name` as this is actually what is goven to the method I think, then by removing the `.keychain` it is only a name we ued to expand to `keychain_paths` again.

hjanuschka (approved):

hjanuschka:
as a bonus, maybe adding a test to verify that incomming parameter is untouched after method call? 🥇

janpio:
Didn't have the time to do this right now - and also not enough knowledge about testing to be honest - but as this sounds useful I noted it in #11399 and will do it later. Thanks for the suggestion @hjanuschka

PR #11388, opened 2018-01-01
Add basic configuration for AppVeyor Windows CI

janpio:
This PR adds a minimal configuration for AppVeyor CI:

- better version number/string
- use ruby environment
- disable default build and run fastlane test script
- change default git checkout newlines to Linux style
- set ENV variables that will be needed by the test script

The tests at AppVeyor are failing, but that is expected. We will need several additional PRs to make the test suite actually pass. (The failing checks are not blocking merges, so this is more a cosmetic problem.)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 # https://www.appveyor.com/docs/appveyor-yml/
 
-version: {build}-{branch}
+version: {build}
 
 init:
   - git config --global core.autocrlf true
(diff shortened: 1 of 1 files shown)


janpio:
Tests at AppVeyor are a) failing and b) the check here had to be manually triggered by me (via a curl request), but that is to be expected as quite some changes have to be implemented. But this PR enables us to enable b) and get future PRs automatically tested on Windows.

janpio:
Interesting that this fails only _inside_ `bundle exec fastlane test`, locally I had to comment out a few dependencies to make it get that far. I will investigate this. (Different ruby version?)

KrauseFx (approved):

PR #11389, opened 2018-01-01
get home directory in a cross platform way

janpio:
`~` is the home dir in most OS, but not all.

https://ruby-doc.org/core-2.2.0/Dir.html#method-c-home


KrauseFx:
Don't we use `~` in more places in the fastlane code base?

janpio:
Oh sure, but those didn't pop up with failing tests yet. Most of these cases are also not as clear cut as this one. (Many of them are in the macOS specific code though, so no need and use to fix those anyway) Just start with this one, and then see what comes up.

KrauseFx (approved):

PR #11404, opened 2018-01-03
[Snapshot] Add test-without-building capability

igor-makarov:
### Checklist

### Motivation and Context
Allow to build test suite once and run multiple times, to save time on large language sets.

### Description
Add `test_without_building` option that is emitted to the xcodebuild command.


snapshot/lib/snapshot/options.rb
```diff
@@ -137,6 +137,13 @@ def self.available_options
                                      description: "Should the project be cleaned before building it?",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :test_without_building,
+                                     short_option: "-T",
+                                     env_name: "SNAPSHOT_TEST_WITHOUT_BUILDING",
+                                     description: "Test without building, requires a derived data path",
```
KrauseFx: Can we add a check for this in `snapshot/detect_values.rb` and show a nice `UI.user_error!("...")` if that's not the case, with instructions on how to set it up?

KrauseFx (commented):
This is great, thanks for your contribution @igor-makarov 👍 

One small comment also. And, can we update `snapshot.md` to mention this and how to use it somewhere?

igor-makarov:
I've added the @KrauseFx requested. 

However, I'm having trouble figuring out where in `snapshot.md` to mention this. I think this should go into docs website. Do I need to edit [this page](https://docs.fastlane.tools/actions/capture_ios_screenshots/) somehow, or is it updated automatically?

KrauseFx:
Ah, sorry, I meant this file: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/snapshot.md which is used to auto-generated the rest of the docs 👍

KrauseFx:
It's then rendered to https://docs.fastlane.tools/actions/snapshot/

snapshot/lib/snapshot/detect_values.rb
```diff
@@ -19,6 +19,10 @@ def self.set_additional_default_values
         end
       end
 
+      if !config[:test_without_building].nil? && config[:derived_data_path].nil?
+        UI.user_error! "Cannot use test_without_building option without a derived_data_path!"
```
KrauseFx: Please add `(` around the method call

snapshot/lib/snapshot/detect_values.rb
```diff
@@ -19,6 +19,10 @@ def self.set_additional_default_values
         end
       end
 
+      if !config[:test_without_building].nil? && config[:derived_data_path].nil?
```
KrauseFx: Use `config[:derived_data_path].to_s.length == 0` instead, as an empty string should also show this error message

igor-makarov:
I checked the file you mentioned, and there is no param list there. The params are auto-generated from the options.

KrauseFx (approved):
Thanks for updating 👍

Issue #11243, opened 2017-12-16
Match could not find App ID


EvgeniyGushchin:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I'm trying to generate certificates for my app using `match` but `match` doesn't able to find App ID
(see output below). 
I have tried to register App ID manually and using `produce` command. Both times `match` failed to find App ID
##### Complete output when running fastlane, including the stack trace and command used

bundle exec fastlane match appstore

+-----------------------+----------------+
|        Summary for match 2.69.3        |
+-----------------------+----------------+
| type                  | appstore       |
| git_branch            | master         |
| keychain_name         | login.keychain |
| readonly              | false          |
| verbose               | false          |
| force                 | false          |
| skip_confirmation     | false          |
| shallow_clone         | false          |
| clone_branch_directly | false          |
| force_for_new_devices | false          |
| skip_docs             | false          |
| platform              | ios            |
+-----------------------+----------------+

[13:00:00]: To not be asked about this value, you can specify it using 'git_url'
[13:00:00]: URL to the git repo containing all the certificates: git@github.com:myapp/ios-certificates.git
[13:00:11]: Cloning remote git repo...
[13:00:11]: If cloning the repo takes too long, you can use the `clone_branch_directly` option in match.
[13:00:18]: 🔓  Successfully decrypted certificates repo
[13:00:18]: To not be asked about this value, you can specify it using 'username'
[13:00:18]: Your Apple ID Username: 
[13:00:37]: Verifying that the certificate and profile are still valid on the Dev Portal...
[13:00:44]: To not be asked about this value, you can specify it using 'app_identifier'
[13:00:44]: The bundle identifier(s) of your app (comma-separated): com.bsl.bsl
[13:00:50]: 
[13:00:50]: ==========================================
[13:00:50]: Could not find App ID with bundle identifier '["com.bsl.bsl"]'
[13:00:50]: You can easily generate a new App ID on the Developer Portal using 'produce':
[13:00:50]: 
[13:00:50]: fastlane produce -u myuser@com.br -a ["com.bsl.bsl"] --skip_itc
[13:00:50]: 
[...]

PR #11407, opened 2018-01-03
Updated handling of Arrays that are passed as Strings

taquitos:
Handle more combination of inputs


taquitos:
I'm pretty sure this should address #11243

KrauseFx (approved):
Good tests 👍  in the future it might make sense to split the `expects` into different test cases, so that the test cases reflect one case for each test :)

taquitos:
@KrauseFx yeah, that's probably a good idea.

PR #11416, opened 2018-01-03
Wrap loading animation with condition of env variable

fxfactorial:
### Checklist

Re: https://github.com/fastlane/fastlane/pull/11352#issuecomment-355043978

### DIFF
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -15,24 +15,27 @@ module Fastlane
       def take_off
         before_import_time = Time.now
 
-        # Usually in the fastlane code base we use
-        #
-        #   Helper.show_loading_indicator
-        #   longer_taking_task_here
-        #   Helper.hide_loading_indicator
-        #
-        # but in this case we haven't required FastlaneCore yet
-        # so we'll have to access the raw API for now
-        require "tty-spinner"
-        require_fastlane_spinner = TTY::Spinner.new("[:spinner] 🚀 ", format: :dots)
-        require_fastlane_spinner.auto_spin
+        if !FastlaneCore::Env.truthy?("FASTLANE_DISABLE_ANIMATION")
+          # Usually in the fastlane code base we use
+          #
+          #   Helper.show_loading_indicator
+          #   longer_taking_task_here
+          #   Helper.hide_loading_indicator
+          #
+          # but in this case we haven't required FastlaneCore yet
+          # so we'll have to access the raw API for now
+          require "tty-spinner"
+          require_fastlane_spinner = TTY::Spinner.new("[:spinner] 🚀 ", format: :dots)
+          require_fastlane_spinner.auto_spin
 
-        # this might take a long time if there is no Gemfile :(
-        # That's why we show the loading indicator here also
-        require "fastlane"
+          # this might take a long time if there is no Gemfile :(
+          # That's why we show the loading indicator here also
+          require "fastlane"
 
-        require_fastlane_spinner.success
-
+          require_fastlane_spinner.success
+        else
+          require "fastlane"
+        end
         # We want to avoid printing output other than the version number if we are running `fastlane -v`
         unless running_version_command?
           print_bundle_exec_warning(is_slow: (Time.now - before_import_time > 3))
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -15,7 +15,7 @@ module Fastlane
       def take_off
         before_import_time = Time.now
 
-        if !FastlaneCore::Env.truthy?("FASTLANE_DISABLE_ANIMATION")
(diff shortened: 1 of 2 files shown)


fxfactorial:
I signed it!

fastlane/lib/fastlane/cli_tools_distributor.rb
```diff
@@ -15,24 +15,27 @@ def running_help_command?
...
-        require "tty-spinner"
-        require_fastlane_spinner = TTY::Spinner.new("[:spinner] 🚀 ", format: :dots)
-        require_fastlane_spinner.auto_spin
-
-        # this might take a long time if there is no Gemfile :(
-        # That's why we show the loading indicator here also
-        require "fastlane"
-
-        require_fastlane_spinner.success
+        if !FastlaneCore::Env.truthy?("FASTLANE_DISABLE_ANIMATION")
```
hjanuschka: i think this is a chicken-egg problem, we haven't loaded fastlane_core at this point-in-code - however you are using `FastlaneCore::`- so i guess you'd need to check the env with raw ruby `ENV[]` style.

this also should fix tests, i hope!

hjanuschka (changes requested):

hjanuschka:
@fxfactorial congrats on your first PR 🚀  🎉   and welcome to the fastlane project.
please see my comments that need to be addressed, and once we have the ci green it should be good to be merged🐙

fastlane/lib/fastlane/cli_tools_distributor.rb
fxfactorial: pushed, thanks.

KrauseFx: Nice catch @hjanuschka 👍

KrauseFx (commented):
Awesome, thanks @fxfactorial. The startup time is one place where we show a loading indicator, and then for the rest of the code base, we're gonna use the `Helper` method

So we'll need a check over here as well: https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/helper.rb#L277-L284

The first line of each method could just be
```ruby
return if FastlaneCore::Env.truthy?("FASTLANE_DISABLE_ANIMATION")
```

fxfactorial:
Okay, so I am not clear about what you want me to do now then? Is there anything for me to change since your linked file seems to imply it has already been done

hjanuschka:
@fxfactorial i think @KrauseFx  means that in addition to your current changes, the check for the ENV should also be done here: https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/helper.rb#L277

as you can assume that fastlane_core is already loaded at this point-in-code.
you can add 

```
return if FastlaneCore::Env.truthy?("FASTLANE_DISABLE_ANIMATION")
```

the `truthy?`  has some nice extras like `FASTLANE_DISABLE_ANIMATION=false` is not `truthy?`

fxfactorial:
Cool, thanks @hjanuschka and I didn't know about that neat ruby syntax feature. Rebased and pushed.

carlosefonseca:
Could `FASTLANE_DISABLE_ANIMATION` be set to true if `is_ci?` is true?

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -275,6 +275,7 @@ def self.gem_path(gem_name)
 
     # Show/Hide loading indicator
     def self.show_loading_indicator(text = "🚀")
+      return if FastlaneCore::Env.truthy?("FASTLANE_DISABLE_ANIMATION")
       @require_fastlane_spinner = TTY::Spinner.new("[:spinner] #{text} ", format: :dots)
       @require_fastlane_spinner.auto_spin
     end
```
hjanuschka: one last nitpick, `self.hide_loading_indicator`  also needs to be worked on other wise it crashes as @require_fastlane_spinner  might not be initialized.

hjanuschka (changes requested):

fastlane_core/lib/fastlane_core/helper.rb
fxfactorial: okay, i think i did that just now.

KrauseFx:
@fxfactorial hey, hey, sorry, totally my bad, I wasn't thinking. Even if not showing the loading indicator, we'll still have to show the text, as it might contain important information.

Sorry for the back and forth - I'm happy to do the extra work if you don't feel like updating the PR, I'd totally understand 👍

fxfactorial:
Sure, np. If you're offering to continue it & finish it, then I'm happy about that.

carlosefonseca:
@fxfactorial Maybe switch it to an `if` and add `UI.message(text)`.
Like what I was doing here https://github.com/fastlane/fastlane/pull/11442/files

KrauseFx:
Hey everyone, sorry, couldn't push on your branch, and ended up in a new PR, you can find it here https://github.com/fastlane/fastlane/pull/11447

Thanks for your contribution @fxfactorial and @carlosefonseca 👍

PR #11417, opened 2018-01-03
Replace `cp ...` with (cross-platform) FileUtils.cp ... 

janpio:
`cp` as a command line tool is not available on Windows (except when somehow installed via MSYS, Cygwin or similar), so this fails on new systems. `FileUtils.cp` works all the time.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -6,7 +6,7 @@ require 'fastlane/version'
 
 # Copy over the latest .rubocop.yml style guide
 rubocop_config = File.expand_path('../.rubocop.yml', __FILE__)
-`cp #{rubocop_config} #{lib}/fastlane/plugins/template/.rubocop.yml`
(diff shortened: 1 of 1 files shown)


janpio:
(❌ for AppVeyor is because of problem further down the line, not because of a problem with this PR. I can and will fix this in a later PR when this is merged.)

hjanuschka (approved):
are there more places where we are using this  kind of cp? ❓

milch (approved):
I agree with @hjanuschka, if there are other places where just cp is used that should be replaced as well 👍🏼Can be done in a separate PR/issue though

janpio:
As far as I could see, only `frameit`'s `Rakefile` and `resign.sh` also use `cp` directly, all others use `FileUtils.sh` already. Didn't want to touch those as they are most probably macOS only anyway.

KrauseFx:
Nice 👍

PR #11384, opened 2017-12-31
FastlaneCore::PTY to replace PTY

janpio:
In https://github.com/fastlane/fastlane/pull/7202 @hjanuschka proposed a way to replace PTY for platforms where it is not available. This PR used this work and applies it to the current codebase.

PR #11418, opened 2018-01-03
Replace `fork` usage with cross-platform alternative

janpio:
`fork` is not available on Windows. Fortunately almost all uses don't rely on the result anyway, so `Thread.new` is a appropriate replacement.

For the one case where it actually matters what happens in the forked process, I couldn't find a way to properly trigger the code. As it is an edge case anyway (as I understood it, requiring a gem that is already preinstalled) and not executed in tests in any way, I made the judgement call to "disable" the code on these platforms by checking for the existence of `Process.fork`.

### DIFF
--- a/fastlane/lib/fastlane/actions/appium.rb
+++ b/fastlane/lib/fastlane/actions/appium.rb
@@ -37,7 +37,7 @@ module Fastlane
 
       def self.invoke_appium_server(params)
         appium = detect_appium(params)
-        fork do
+        Thread.new do
           Process.exec("#{appium} -a #{params[:host]} -p #{params[:port]}")
         end
       end
--- a/fastlane/lib/fastlane/actions/appium.rb
+++ b/fastlane/lib/fastlane/actions/appium.rb
@@ -37,9 +37,7 @@ module Fastlane
 
       def self.invoke_appium_server(params)
         appium = detect_appium(params)
-        Thread.new do
-          Process.exec("#{appium} -a #{params[:host]} -p #{params[:port]}")
-        end
+        Process.spawn("#{appium} -a #{params[:host]} -p #{params[:port]}")
       end
 
       def self.detect_appium(params)
--- a/fastlane/lib/fastlane/fastlane_require.rb
+++ b/fastlane/lib/fastlane/fastlane_require.rb
@@ -39,24 +39,26 @@ module Fastlane
       end
 
       def gem_installed?(name, req = Gem::Requirement.default)
-        # In special cases a gem is already preinstalled, e.g. YAML.
-        # To find out we try to load a gem with that name in a child process
-        # (so we don't actually load anything we don't want to load)
-        # See https://github.com/fastlane/fastlane/issues/6951
-        if Process.respond_to?(:fork)
-          fork do
-            begin
-              require name
-            rescue LoadError
-              exit(1)
+        installed = Gem::Specification.any? { |s| s.name == name and req =~ s.version }
+        unless installed
+          # In special cases a gem is already preinstalled, e.g. YAML.
+          # To find out we try to load a gem with that name in a child process
+          # (so we don't actually load anything we don't want to load)
+          # See https://github.com/fastlane/fastlane/issues/6951
+          if Process.respond_to?(:fork)
+            fork do
+              begin
+                require name
+              rescue LoadError
+                exit(1)
+              end
             end
+            _pid, status = Process.wait2
+            return true if status.exitstatus == 0
+            # else OS doesn't support `fork` (e.g. Windows)
           end
-          _pid, status = Process.wait2
-          return true if status.exitstatus == 0
-          # else OS doesn't support `fork` (e.g. Windows)
         end
-
-        Gem::Specification.any? { |s| s.name == name and req =~ s.version }
+        return installed
       end
 
       def find_gem_name(user_supplied_name)
--- a/fastlane/lib/fastlane/fastlane_require.rb
+++ b/fastlane/lib/fastlane/fastlane_require.rb
@@ -40,22 +40,21 @@ module Fastlane
 
       def gem_installed?(name, req = Gem::Requirement.default)
         installed = Gem::Specification.any? { |s| s.name == name and req =~ s.version }
-        unless installed
-          # In special cases a gem is already preinstalled, e.g. YAML.
-          # To find out we try to load a gem with that name in a child process
-          # (so we don't actually load anything we don't want to load)
-          # See https://github.com/fastlane/fastlane/issues/6951
-          require_tester = <<-RB.gsub(/^ */, '')
-            begin
-              require ARGV.first
-            rescue LoadError
-              exit(1)
-            end
-          RB
-          system(RbConfig.ruby, "-e", require_tester.lines.map(&:chomp).join("; "), name)
-          return true if $?.success?
-        end
-        return installed
+        return true if installed
+
+        # In special cases a gem is already preinstalled, e.g. YAML.
+        # To find out we try to load a gem with that name in a child process
+        # (so we don't actually load anything we don't want to load)
+        # See https://github.com/fastlane/fastlane/issues/6951
+        require_tester = <<-RB.gsub(/^ */, '')
+          begin
+            require ARGV.first
+          rescue LoadError
+            exit(1)
+          end
+        RB
+        system(RbConfig.ruby, "-e", require_tester.lines.map(&:chomp).join("; "), name)
+        return true if $?.success?
       end
 
       def find_gem_name(user_supplied_name)
--- a/fastlane/lib/fastlane/fastlane_require.rb
+++ b/fastlane/lib/fastlane/fastlane_require.rb
@@ -39,19 +39,22 @@ module Fastlane
       end
 
       def gem_installed?(name, req = Gem::Requirement.default)
-        # We fork and try to load a gem with that name in the child process so we
-        # don't actually load anything we don't want to load
-        # This is just to test if the gem is already preinstalled, e.g. YAML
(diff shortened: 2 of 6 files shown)


janpio:
(AppVeyor/Windows ❌ is because #11384 is not merged yet, not indication that this PR is bad)

KrauseFx:
One thing that's really important, is that we don't introduce any regressions with this. That should be our #1 priority. Maybe we can have `if else` blocks to keep the existing way for all macOS systems, and fallback to something else for Windows? Maybe we can abstract things out in a central class?

janpio:
Makes sense, especially as this affects 3 very sensitive areas: Loading of plugins, sending of analytics and checking for updates.

The `if Process.respond_to?(:fork)` in `fastlane_require.rb` already is checking if `Process.fork` is implemented, which is the most reliable way to find out if `fork` does work or not. 

I could definitely copy that over to the other cases and only `Thread.new` if `fork` is definitely not available. Then this only introduces new code to platforms that didn't work before at all and the existing and working platforms run the same code.

Or does anyone have a better idea how to solve this maybe?

fastlane/lib/fastlane/fastlane_require.rb
```diff
@@ -39,19 +39,22 @@ def install_gem_if_needed(gem_name: nil, require_gem: true)
...
+        # In special cases a gem is already preinstalled, e.g. YAML.
+        # To find out we try to load a gem with that name in a child process
+        # (so we don't actually load anything we don't want to load)
         # See https://github.com/fastlane/fastlane/issues/6951
-        fork do
-          begin
-            require name
-          rescue LoadError
-            exit(1)
+        if Process.respond_to?(:fork)
```
milch: You could instead do something like the following:

```ruby
require_tester = <<-RB.gsub(/^ */, '')
  begin
    require ARGV.first
  rescue LoadError
    exit(1)
  end
RB
system(RbConfig.ruby, "-e", require_tester.lines.map(&:chomp).join("; "), name)
return true if $?.success?
```

Which should work on all of Linux, Windows and macOS

milch (approved):
Looks good to me 👍🏼 I'm not sure though why some of those `fork`s were used in the first place instead of `Thread.new` so might be that I'm overlooking something here (in `update_checker.rb`, `tool_collector.rb` and `analytics_ingester_client.rb`)

fastlane/lib/fastlane/actions/appium.rb
```diff
@@ -37,7 +37,7 @@ def self.run(params)
 
       def self.invoke_appium_server(params)
         appium = detect_appium(params)
-        fork do
+        Thread.new do
           Process.exec("#{appium} -a #{params[:host]} -p #{params[:port]}")
```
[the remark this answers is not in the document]
janpio: Oh, good point. I will replace this. Less code is always good I think.

fastlane/lib/fastlane/fastlane_require.rb
janpio: Ah, that's what you meant in that comment of the commit where this was added ;)

Without having tried it, this looks like it could do what it should do.

I am still unsure how to actually test this... I am really hesitant on mess with this (hopefully working) code without knowing how I can replicate this situation or best even write a test that makes sure this code does what it should do. Ideas?

milch: Sure, the whole purpose of this code was just to test whether a module comes preinstalled, like `time` or `yaml`. If you run `Fastlane::FastlaneRequire.gem_installed?("yaml")` it return `true` if this code works and `false` if it doesn't.

The reason this was added was that the line below `Gem::Specification.any? #...` returns true *only* for gems that were installed with RubyGems, not anything else that would still be a valid `require` statement. 

For example, fire up `bundle exec pry` and try out the following statements:

```ruby
[1] pry(main)> Gem::Specification.any { |s| s.name == "yaml" }
=> false
[2] pry(main)> require "yaml"
=> true
[3] pry(main)> require "blabla"
LoadError: cannot load such file -- blabla
from (pry):2:in `require'
[4] pry(main)> require_relative 'fastlane/lib/fastlane/fastlane_require'
=> true
[5] pry(main)> Fastlane::FastlaneRequire.gem_installed?("yaml")
=> true
[6] pry(main)> Fastlane::FastlaneRequire.gem_installed?("blabla")
=> false
```

janpio: Ah ok.
On which systems a module is preinstalled? Is this default for all ruby installations?

milch: There's a couple that come default for all ruby installations ... why they don't get required I have no idea however. I don't know if there's a list anywhere either

janpio: Thanks for the explanation, I finally understood what is happening here and wrote some tests for the method that show it working on Ubuntu and macOS (and failing locally on Windows). Now I can continue to work on the code and try what you posted above.

janpio: Done. I also moved the "normal" case with the simple check before the special, edge case handling. 

Works perfectly :)

janpio:
@KrauseFx I wrote some tests that cover the functionality of that method to make sure the current state works 100% identical to what it was before.

(These are the first tests I wrote for fastlane. Is the grouping and example naming ok? Don't really know much about that)

I used a suggestion of @milch to achieve this, where instead of forking a process we run `ruby -e` with a "file" we build on the fly that does the same check. Additionally I moved the "normal" case before this edge case handling, so the code doesn't even need to be run in 99% of the cases.

(Tested locally on Windows to confirm the new tests pass there as well)

fastlane/lib/fastlane/fastlane_require.rb
```diff
@@ -39,21 +39,23 @@ def install_gem_if_needed(gem_name: nil, require_gem: true)
...
-        # don't actually load anything we don't want to load
-        # This is just to test if the gem is already preinstalled, e.g. YAML
-        # See https://github.com/fastlane/fastlane/issues/6951
-        fork do
-          begin
-            require name
-          rescue LoadError
-            exit(1)
-          end
+        installed = Gem::Specification.any? { |s| s.name == name and req =~ s.version }
```
milch: Good call moving this up, that's probably a good idea anyway. Small nitpick, instead of the 

```
unless installed
  # ...
end
```

Maybe just do an early return

milch:
@janpio Just re-reviewed, still looks good 👍🏼

fastlane/lib/fastlane/fastlane_require.rb
janpio: `return installed if installed`?

milch: Yeah or just `return true if installed` (since installed can only be true anyway, and to make that explicit)

lacostej:
For future maintenance, are there ways to detect (with rubocop?) that `fork` is used? (cc @milch)

milch:
@lacostej Sure, that's definitely possible

fastlane/lib/fastlane/fastlane_require.rb
```diff
@@ -39,21 +39,22 @@ def install_gem_if_needed(gem_name: nil, require_gem: true)
...
             exit(1)
           end
-        end
-        _, status = Process.wait2
-        return true if status.exitstatus == 0
-
-        Gem::Specification.any? { |s| s.name == name and req =~ s.version }
+        RB
+        system(RbConfig.ruby, "-e", require_tester.lines.map(&:chomp).join("; "), name)
+        return true if $?.success?
```
lacostej: why not `return $?.success?` ?

janpio: That is leftover from when this code run first, I will fix it.

janpio: And fixed. Should be good now.

milch:
@lacostej I've added a RuboCop cop that checks this 👍 /cc @janpio

janpio:
@milch rubocop is unhappy:
```
[07:00:14]: ▸ Offenses:
[07:00:14]: ▸ fastlane.gemspec:10:15: C: Security/YAMLLoad: Prefer using YAML.safe_load over YAML.load.
[07:00:14]: ▸ config = YAML.load(open(rubocop_config))
[07:00:14]: ▸ ^^^^
[07:00:14]: ▸ 1021 files inspected, 1 offense detected
```
Also conflicts to master now - could you also take a look?

milch:
@janpio Have you tried just replacing the `load` with `safe_load`? Seems like it should work. I can take a look into rebasing on master later

janpio:
I was not sure if this was valid as this is your code, not mine. But now I changed it, seems to work.
Also resolved the merge conflicts introduced by this and will now also rebase on master again to maybe get successful builds.

janpio:
@milch Unfortunately I landed in merge hell and had to "clean up" a bit, during which because of conflicting merges the authorship of your got lost for your commits. I "saved" them (caution: and a bit of junk) to https://github.com/fastlane/fastlane/commit/860ce291e1ca4b6db6d55268e0f93e6e061b7885 so you can a) commit the code here again or b) create an additional PR. Let me know how you want to proceed and I will merge this, wait for your commit and/or delete that branch again.

milch:
Hahaha, can't say that hasn't happened to me before. You can merge if you want, I'll just do a separate PR

janpio:
Done.

Seriously, how do I handle merge conflicts during rebase that were made from another account/person? Also, how do I keep the commit message? Source tree just goes into "conflict resolution mode" with an empty commit message input field and with my user...

milch:
Here: #11502 

I have no idea to be honest

PR #11423, opened 2018-01-04
Fix not auto-converting values after user input

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/11243


taquitos (approved):
🎈🐐

PR #11424, opened 2018-01-04
Revert 'Updated handling of Arrays that are passed as Strings'

KrauseFx:
I'm not sure if we want to revert this, it's your call @taquitos. Just wanted to have a PR ready so we can hit the merge button if we decide to revert


taquitos (approved):
Yup, better to keep things simple if an easier fix is available! ☂️🐐

PR #11427, opened 2018-01-04
Replace hard coded test paths (e.g. /tmp)

janpio:
- On macOS and Linux systems you always have `/tmp`. On Windows you don't.
- Also, on macOS and other systems you can use `/path` and also get back `/path` from commands. On Windows you get back `C:/path`, which breaks tests that check for `/path`.

This PR fixes all the tests that assumed such things for paths to also work on Windows. 

Mainly I created a variable that got a tmp path via `Dir.mktmpdir` or similar, extracted the paths used in the test into a variable that then used that tmp path and replaced the hardcoded paths in the tests with that second variable.

(I am pretty sure there are more tests and code that assume hardcoded paths, but those didn't fail any tests on Windows and so are out of scope for this PR. Later one could go through the whole codebase and look for those, add tests so the code gets actually exercided and fix them as well.)


lacostej:
Everytime I run my tests, ruby will create new files/directories (we don't use the block version of the call most of the time).

2 issues
* you might run into issues if you like me do not reboot often and have the directories fill up
* you sometimes want to check why something fail and you have lots of directories to look.

A solution would be to create a top temp dir in spec_helper using mkdtmpdir and then provide an API for the specs to create tmpdir under that folder.

That way we can clean up all temp test resources by removing the top tmpdir if the tests passed.

WDYT?

janpio:
Sounds wonderful. But...

I thought about it briefly, after looking through the code base how many changes that would mean I decided for now to not tackle this and just get it into a working state. Reworking how the tests work with temp directories is definitely out of scope ;)

(But I would definitely give input into an issue collecting ideas about this. Making `/tmp` [or better: `C:/tmp/`] non-writeable is on my #windows TODO as well, and I guess this will popup there again)

janpio:
Cleaned up all the conflicts that happened in the mean time. Could I please have a review?

janpio:
Cleaned up all the conflicts that happened in the mean time.
Could I please have a review?

KrauseFx (approved):
Looks great, nice work 👍

Issue #10166, opened 2017-08-28
Same file error


florianhenonin:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When I try to generate ipa with fastlane I always get that error : 
(Real app names are replaced)

```
+------+-----------------------+-------------+
|              fastlane summary              |
+------+-----------------------+-------------+
| Step | Action                | Time (in s) |
+------+-----------------------+-------------+
| 1    | Verifying required    | 0           |
|      | fastlane version      |             |
| 2    | default_platform      | 0           |
| 3    | xcversion             | 1           |
| 4    | clean_build_artifacts | 0           |
| 5    | clear_derived_data    | 8           |
| 💥   | gym                   | 1285        |
+------+-----------------------+-------------+

[10:05:36]: fastlane finished with errors

Looking for related GitHub issues on fastlane/fastlane...

/Users/xxx/.fastlane/bin/bundle/lib/ruby/2.2.0/fileutils.rb:1569:in `block in fu_each_src_dest': [!] same file: /var/folders/kb/3qmqmk_569v9ggc23nnc5btm0000gn/T/gym_output20170828-57245-1owcwhu/xx xx xx - xx.ipa and /var/folders/kb/3qmqmk_569v9ggc23nnc5btm0000gn/T/gym_output20170828-57245-1owcwhu/xx xx xx - xx.ipa (ArgumentError)
```

This is my fastline file :
```
fastlane_version "2.54.2"

default_platform :ios

platform :ios do
    lane :xxx_test do
        xcversion(version: "7.3.1") # Selects Xcode version

        clean_build_artifacts
        clear_derived_data

        gym(scheme: "xx xx xx - xx", workspace: "XXX.xcworkspace", include_bitcode: false, clean: true)
        testflight
    end
end

```

In XCode Organizer, even after this error I got my ipa generated after few minutes...
Manual archive works fine.

Any ideas ?

florianhenonin:
Fastlane env :
`<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                        |
| --------------------------- | -------------------------------------------- |
[...]

PR #11432, opened 2018-01-04
add helper methods for detecting fs_insensitivty and comparing paths

hjanuschka:
trying  to fix: 
  - https://github.com/fastlane/fastlane/issues/11392
  - https://github.com/fastlane/fastlane/issues/10166

waiting for user confirmation.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -113,16 +113,27 @@ module FastlaneCore
       self.test?
     end
 
+    # Compare given paths if they are equal, respecting filesystem case sensitivity.
+    # Some systems have insensitive filesystems (e.g.: macOS AFPS)
+    #   https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/APFS_Guide/FAQ/FAQ.html
+    # Issues:
+    #   - https://github.com/fastlane/fastlane/issues/11392
+    #   - https://github.com/fastlane/fastlane/issues/10166
     def self.compare_paths(path1, path2)
       compare_path1 = File.expand_path(path1)
       compare_path2 = File.expand_path(path2)
       if self.fs_is_insensitive?
-        # FS is insensitive, there fore the file `fastlane`  is also accessable via `FASTLANE``
-        return compare_path1.downcase == compare_path2.downcase
+        # FS is insensitive, so /path/fastlane is equal to /path/FASTLANE
+        # return the compare result by ignoring case sensitivity
+        return compare_path1.casecmp(compare_path2).zero?
       end
+      # FS is configured with sensitive filenames
+      # /tmp/fastlane is not /tmp/FASTLANE
+      # return result of plain compare.
       return compare_path1 == compare_path2
     end
 
+    # Detect if filesystem is insenstive or not.
     def self.fs_is_insensitive?
       is_insensitive = false
       Dir.mktmpdir("foo") do |dir|
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -135,12 +135,13 @@ module FastlaneCore
 
     # Detect if filesystem is insenstive or not.
     def self.fs_is_insensitive?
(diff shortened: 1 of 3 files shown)


hjanuschka:
user confirmed: https://github.com/fastlane/fastlane/issues/11392#issuecomment-355401903

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -113,6 +113,25 @@ def self.is_test?
       self.test?
     end
 
+    def self.compare_paths(path1, path2)
```
KrauseFx: Can you add some comments here on what this does, why it's necessary?

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -113,6 +113,25 @@ def self.is_test?
...
+      compare_path1 = File.expand_path(path1)
+      compare_path2 = File.expand_path(path2)
+      if self.fs_is_insensitive?
+        # FS is insensitive, there fore the file `fastlane`  is also accessable via `FASTLANE``
```
KrauseFx: I don't understand that comment, also double space after fastlane, also `therefore` I think?

KrauseFx:
Looks awesome, I understand the issue, but still needs some comments. You can even link the GitHub issues in the code base 👍

hjanuschka:
thx @KrauseFx feedback addressed

fastlane_core/spec/helper_spec.rb
```diff
@@ -58,6 +58,30 @@
...
+      it "detects fs sensitivity" do
+        expect(File).to receive(:exist?).and_return(true)
+        expect(FastlaneCore::Helper.fs_is_insensitive?).to be true
+        expect(File).to receive(:exist?).and_return(false)
+        expect(FastlaneCore::Helper.fs_is_insensitive?).to be false
+      end
+    end
+
+    describe "#compare_paths?" do
+      it "Compare Path on sensitive system" do
```
lacostej: from a rspec style perspective, you could use `context` to avoid having to `expect` multiple times and make the intent clearer.

```
      context 'when the file system is insentive' do
        before(:each) do
          allow(FastlaneCore::Helper).to receive(:fs_is_insensitive?) { true }
        end
        it 'ignores case while comparing' do
             expect(FastlaneCore::Helper.compare_paths("FASTLANE", "fastlane")).to be true
        end
        it 'matches equal strings' do
             expect(FastlaneCore::Helper.compare_paths("fastlane", "fastlane")).to be true
        end
```

Not a big deal here, but might be useful in other "contextes" :)

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -113,6 +113,36 @@ def self.is_test?
...
+      end
+      # FS is configured with sensitive filenames
+      # /tmp/fastlane is not /tmp/FASTLANE
+      # return result of plain compare.
+      return compare_path1 == compare_path2
+    end
+
+    # Detect if filesystem is insenstive or not.
+    def self.fs_is_insensitive?
+      is_insensitive = false
```
lacostej: ideally we could / should cache this value, especially as we create file+dir to resolve it.

hjanuschka:
@lacostej thx for your review, added a cache of the tmp file creation. and changed to the context in rspec, also added a bunch of new tests for the cache.

janpio:
Test failure on Ubuntu can be ignored as this was just a race condition thing with the documentation. Should be gone when rebasing on `master` though as we improved our testing setup.

Conflict should also go away.

@hjanuschka Can you rebase on master please? Then I think this is good to merge, @lacostej @KrauseFx ?

antoinelamy:
Bump, still experiencing the issue this PR aims to fix. @lacostej ? @KrauseFx ?

janpio:
This needs a rebase and conflict resolution before anyone can think about merging. You could possibly do so in a new branch and PR, I assume @hjanuschka just doesn't have the time right now.

antoinelamy:
@janpio done -> https://github.com/fastlane/fastlane/pull/13479

Issue #11441, opened 2018-01-05
Loading Indicator and Jenkins


carlosefonseca:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Recently (#11378) Deliver was updated with loading indicators for some tasks but this kind of stuff sucks when running on Jenkins (see the image). Could a`is_ci?` check be added to not animate this indicators on CI systems?

![](https://dsc.cloud/carlosefonseca/Screen-Shot-2018-01-05-15-46-14-LFqeM.png)

fxfactorial:
This is exactly what I was talking about in my comments on the PR that introduced this feature originally.

carlosefonseca:
Yeah, I saw the PR but missed the Issue and the discussion. I'm gonna close this one. See #11416.

PR #11442, opened 2018-01-05
Added CI checks to show/hide_loading_indicator

carlosefonseca:
See #11441.


carlosefonseca:
@googlebot I signed it!

janpio:
See #11416 for work done on the same code with a different goal. but shows another location probably relevant. You might also want to make sure your code works with the other PR's code.

carlosefonseca:
I can't do `is_ci?` on the start up animation since it's not yet available and I'd rather avoid duplicating the code. The other PR is probably better.

janpio:
Adding an `is_ci?` check still seems useful to me as it doesn't require manually setting a ENV var, even if it won't work for all spinners.

PR #11472, opened 2018-01-08
Check for missing requires 

milch:
I included my require checker as a rubocop extension: https://github.com/milch/rubocop-require_tools

Checking this automatically for PRs is definitely preferred 👍🏼

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -278,6 +278,3 @@ Style/IfInsideElse:
 # Sometimes we just want to 'collect'
 Style/CollectionMethods:
   Enabled: false
-
-Lint/UnneededDisable:
-  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+require: 
+  - rubocop/require_tools
+
 Style/MultipleComparison:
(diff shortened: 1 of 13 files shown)


janpio:
Oh, that Ubuntu failure is interesting: https://circleci.com/gh/fastlane/fastlane/13647 `xcode-select` is not installed and failing. I wonder how you could guard against that :/

milch:
Yeah seems to me like this just doesn't work. It'd probably also fail on Windows?

KrauseFx (approved):
Perfect! You saw the build failure, besides that, it's all 💯

janpio:
@milch Yeah, probably not - at least not in an easy way. You could implement one of the comments that can be used to manipulate/turn off rubocop rules that is platform dependant. Then "tag" all the requires this applies to.

milch:
@janpio Fixed ... since that code is not meant to be run on Linux anyway I just early-exit if it is 👍🏼 

Beautiful, now macOS and Ubuntu fail for the same reason

KrauseFx (approved):
Wow, sweeeet

milch:
@KrauseFx Much thank, very wow

PR #11489, opened 2018-01-09
Cache the homebrew repo on Circle 

milch:
This should speed up the build by like a minute, hopefully

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,8 @@ jobs:
 
       - restore_cache:
           key: v1-macos-gems-{{ checksum "Gemfile" }}
+      - restore_cache:
+          key: v1-homebrew
       - run:
           name: Setup Build
           command: |
@@ -22,6 +24,10 @@ jobs:
             gem update --system
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
+      - save_cache:
+          key: v1-homebrew
+          paths:
+            - /usr/local/Homebrew
       - save_cache:
           key: v1-macos-gems-{{ checksum "Gemfile" }}
           paths:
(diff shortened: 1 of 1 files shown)


janpio:
`brew install` uses this?

milch:
Yeah if you check macOS tests, you can see that `brew install shellcheck` takes ~1min. Most of that is because brew since a couple versions ago updates the entire homebrew repo before install, which takes loooooong. This caches that

milch:
@janpio There, force pushed  - the Setup Bundle step went from 1:23 to 14s

.circleci/config.yml
```diff
@@ -14,6 +14,9 @@ jobs:
 
       - restore_cache:
           key: v1-macos-gems-{{ checksum "Gemfile" }}
+      - restore_cache:
+          key: v1-homebrew-{{ .BuildNum }}
```
janpio: This will never really match, correct?

.circleci/config.yml
```diff
@@ -14,6 +14,9 @@ jobs:
...
+          key: v1-homebrew
```
janpio: And that will never change. Isn't this a source for problems if something in `/usr/local/Homebrew` would/should change?

milch: From my previous testing with Circle partial matches produce a cache hit. So even though e.g. in the previous build it will be saved as `v1-homebrew-1234` it will match `v1-homebrew`. I believe we could potentially remove the line above but it's in their example, so I'd rather leave it in. I think I do need to upgrade the `v1` to `v2` though since I previously used the exact key `v1-homebrew` for `save_cache`

`/usr/local/homebrew` is a git repository that brew does `git pull` on when you run it, so if anything ever changes the worst that can happen is that it'll get slower again over time

janpio:
Maybe @DanToml can give some more input how we should cache something like that?

janpio (approved):

PR #11491, opened 2018-01-09
We're all one team ❤

KrauseFx:
**Proposed changes:**
- Already changed the GitHub teams, so that all people (Google and non-Google) have the same permissions
- No more difference between Google employed engineers and core contributors, we're all one team ❤️ 
- Have a list of all core contributors in a machine readable format to use for other things below
- List the whole team on the GitHub main README, the [docs page](https://docs.fastlane.tools) and the new [fastlane.tools website](https://fastlane.tools). Those can be auto generated based on the `.json` file
- Using that list, we could even automate adding new peeps to GitHub teams
- Automatically update them `.gemspec` to list all authors

^ All points above are proposed with this PR here, and not decided on

**Gonna do a separate PR for:**
- Grant **push access to RubyGems** to all authors, based on this list and replace the [old script](https://github.com/fastlane/fastlane/blob/master/Rakefile#L7-L14)


milch:
Awesome 💯 Maybe we should also add the GitHub username, since GH and Twitter won't be equal for everyone (like me) 

> Grant push access to RubyGems to all authors, based on this list and replace the old script

I think this would be great since we won't depend on Google employees to push anymore. However, we also have to think about the chances of a compromised core contributor potentially pushing rogue updates. If we could instead have it so that CircleCI automatically pushes an update when a Version Bump commit is pushed to master, that would be even better, I think 👍 

> No more difference between Google employed engineers and core contributors, we're all one team ❤️

❤️

lacostej:
> If we could instead have it so that CircleCI automatically pushes an update when a Version Bump
> commit is pushed to master, that would be even better, I think

Go for it!

hjanuschka:
> No more difference between Google employed engineers and core contributors, we're all one team ❤️
:heart: super awesome.!!! best team!


i also would prefer a CI based release/push - like pushing to a tag `create_release`  or stuff like that.  in that case tests can run, and the publish procedure can run afterwards, without depending on my local machine, should reduce failure rate.

KrauseFx:
Just verified: it's possible to add new owners using username also, so no worries re email. Seems like we want to go with the CI approach anyway, which we could do, as we already have the nightly builds set up

KrauseFx:
The problem with deploying from a server is mostly 2 things:

1) During the `fastlane release` lane, you'll be asked for changelog, which mostly is just the commit messages, but usually we're trying to order them by importance and prettify some changes to make it easy for users to understand. Also we don't mention things that are internal only
2) We usually tweet the release (which could be automated using the `twitter` action, but we append a screenshot of the release notes)

We already have [fastlane/nightly](https://github.com/fastlane/nightly), so we already have the foundation needed to automate the release. 

I see 3 potential solutions:
1) Grant RubyGems push access to people who want to have it
2) Have the automation ready, but only use it for hotfix releases
3) Let everybody do a release, but the actual `gem push` happens from a centralized server. This requires some engineering work, but something we could do. We could use the same code as used in `nightly`, and do just the `gem push`, and the rest of the deployment happens on your local machine. This way we don't have to distribute push access to RubyGems to everybody.

KrauseFx:
> Awesome 💯 Maybe we should also add the GitHub username, since GH and Twitter won't be equal for everyone (like me)

@milch I thought the `key` of the JSON is the GitHub user :)

lacostej:
@googlebot I do!

KrauseFx:
screen shot 2018-01-10 at 12 37 14 pm

fastlane/Fastfile
```diff
@@ -105,6 +105,9 @@ lane(:bump) do |options|
   # Add version file path to change set
   paths_for_commit << version_file_path
 
+  sh "bundle exec rake generate_team_table"
```
taquitos: I love this 🎈:balloon::goat:

Rakefile
```diff
@@ -25,6 +25,36 @@ task(:push) do
...
 
+task(:generate_team_table) do
+  require 'json'
+  content = ["<table>"]
+
+  contributors = JSON.parse(File.read("team.json"))
+  counter = 0
+  number_of_rows = 5
+
+  contributors.keys.shuffle.each do |github_user|
```
lacostej: I ❤️ the shuffle!

README.md
```diff
@@ -26,6 +26,90 @@ If the above doesn't help, please [submit an issue](https://github.com/fastlane/
 
 **Note**: If you want to report a regression in _fastlane_ (something that has worked before, but broke with a new release), please mark your issue title as such using `[Regression] Your title here`. This enables us to quickly detect and fix regressions.
 
+## _fastlane_ team
+
+<table>
+<tr>
+<td>
+<img src='https://github.com/lacostej.png?size=200' width=140>
+<h4 align='center'><a href='https://twitter.com/'>Jerome Lacoste</a></h4>
```
lacostej: where did my accents go?

Note that I only added them to see what would happen with them ;)

README.md
```diff
@@ -26,6 +26,90 @@ If the above doesn't help, please [submit an issue](https://github.com/fastlane/
 
 **Note**: If you want to report a regression in _fastlane_ (something that has worked before, but broke with a new release), please mark your issue title as such using `[Regression] Your title here`. This enables us to quickly detect and fix regressions.
 
+## _fastlane_ team
+
+<table>
+<tr>
+<td>
+<img src='https://github.com/lacostej.png?size=200' width=140>
```
revolter: Why not use https://github.com/lacostej.png?size=140 and drop the `width` attribute?

Rakefile
```diff
@@ -25,6 +25,36 @@ task(:push) do
...
+    user_content = contributors[github_user]
+
+    content << "<tr>" if counter % number_of_rows == 0
+    content << "<td>"
+    content << "<img src='https://github.com/#{github_user}.png?size=200' width=140>"
```
revolter: Why not use https://github.com/lacostej.png?size=140 and drop the `width` attribute?

fastlane/Fastfile
revolter: I think that the CI wants this to be `sh("bundle exec rake generate_team_table")`

Rakefile
```diff
@@ -25,6 +25,36 @@ task(:push) do
...
+  content = ["<table>"]
+
+  contributors = JSON.parse(File.read("team.json"))
+  counter = 0
+  number_of_rows = 5
+
+  contributors.keys.shuffle.each do |github_user|
+    user_content = contributors[github_user]
+
+    content << "<tr>" if counter % number_of_rows == 0
```
milch: This looks really good on desktop/tablet but it's a bit small on mobile. Does anyone know enough html/css magic to make it look pretty on both? 😊

revolter: What exactly is small? Could you post a screenshot? As it looks ok for me.

milch: e.g. here the rightmost 2 columns are cut off:

screen shot 2018-01-11 at 11 37 44

milch: If this was iOS I'd be using a `UICollectionView` but somehow I doubt html has a `<collection-view>` tag 😄

revolter: It has if you want (meaning you can pun any string and it doesn't crash), but it won't have the functionality too. 😂  

I'm not sure how would this be achievable, as the HTML is obviously static and the DOM is constructed with multiple columns and I don't think you can change it with CSS. Except maybe if it uses flexbox :-?

milch: Yeah I did some quick testing, modifying the `generate_team_table` task like this: 

```ruby
task(:generate_team_table) do
  require 'json'
  content = ["<div id='team-table' style='display:flex; flex-wrap:wrap; flex-direction:row;' width='100%'>"]

  contributors = JSON.parse(File.read("team.json"))
  contributors.keys.shuffle.each do |github_user|
    user_content = contributors[github_user]

    content << "<div style='padding: 6px 13px; border: 1px solid #dfe2e5;'>"
    content << "<img src='https://github.com/#{github_user}.png?size=200' width=140>"
    content << "<h4 align='center'><a href='https://twitter.com/#{user_content['twitter']}'>#{user_content['name']}</a></h4>"
    # content << "<p align='center'>#{user_content['slogan']}</p>" if user_content['slogan'].to_s.length > 0

    content << "</div>"
  end
  content << "</div>"

  readme = File.read("README.md")
  readme.gsub!(%r{\<div id='team-table'.*\<\/div\>}m, content.join("\n"))
  File.write("README.md", readme)
  puts("All done")
end
```

Seems to make it work without hard-coding the row/column count, and just uses the space available

revolter: Something like [this][1], though it should be tested on GitHub by actually committing it, because the preview adds extra elements which breaks this.

  [1]: http://jsbin.com/zusudar/edit?html,output

janpio: Yep, in the "old days" we would have `float:left` on all those `div`s ;)

hjanuschka: does the github markdown allow div tags with inline styles?

milch: Ha, no, turns out you can't: https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb#L40

Boooooo

hjanuschka: have inline styles 👍 

how about using https://github.com/minimagick/minimagick and rendering a single image?, well thats drifting off a bit, don't think it should be too complicated...sorry

revolter: Well, yes, the best solution is: rotate your phone 90 degrees 😆

KrauseFx: So anything we can do, or should just leave it and ignore mobile? On docs.fastlane.tools we can have it nicer anyway, also on fastlane.tools

milch: Yeah not much we can do I think, since GitHub is blocking everything

README.md
KrauseFx: PR? :)

README.md
```diff
@@ -26,6 +26,90 @@ If the above doesn't help, please [submit an issue](https://github.com/fastlane/
...
+<img src='https://github.com/giginet.png?size=200' width=140>
+<h4 align='center'><a href='https://twitter.com/giginet'>Kohki Miki</a></h4>
+</td>
+<td>
+<img src='https://github.com/snatchev.png?size=200' width=140>
+<h4 align='center'><a href='https://twitter.com/snatchev'>Stefan Natchev</a></h4>
+</td>
+<td>
+<img src='https://github.com/AliSoftware.png?size=200' width=140>
+<h4 align='center'><a href='https://twitter.com/'>Olivier Halligon</a></h4>
```
janpio: Link to Twitter homepage here.

README.md
```diff
@@ -26,6 +26,90 @@ If the above doesn't help, please [submit an issue](https://github.com/fastlane/
...
+</td>
+<td>
+<img src='https://github.com/revolter.png?size=200' width=140>
+<h4 align='center'><a href='https://twitter.com/Revolt666'>Iulian Onofrei</a></h4>
+</td>
+</tr>
+<tr>
+<td>
+<img src='https://github.com/janpio.png?size=200' width=140>
+<h4 align='center'><a href='https://twitter.com/'>Jan Piotrowski</a></h4>
```
janpio: README is generated? Maybe do once more so will release with current data (except if this is done on release anyway)

team.json
```diff
@@ -0,0 +1,73 @@
+{
+  "KrauseFx": {
+    "name": "Felix Krause",
+    "twitter": "KrauseFx",
+    "slogan": "Professional iOS Code Signing Issue Resolver"
```
milch: Felix with his fancy slogan

janpio (approved):
Awesome.

Note I added 2 last comments re the Twitter links, but doesn't block merging.

README.md
revolter: @KrauseFx, In the JSON they exist, so the README only needs to be re-generated

team.json
KrauseFx: Yep, not being used yet though

README.md
KrauseFx: Yep, auto-generated, part of this PR

nafu (approved):
Love this ❤️  Best team 🚀

team.json
nafu: ✨

giginet:
:heart:

Issue #11481, opened 2018-01-08
[Regression] method `app_id' not defined in Spaceship::TestFlight::Build (NameError)


michaelgmcd:
In versions 2.74.* I get the following error:
spaceship/lib/spaceship/base.rb:137:in `remove_method': method `app_id' not defined in Spaceship::TestFlight::Build (NameError)

### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When running any fastlane command, I get the following error, claiming that `method `app_id' not defined in Spaceship::TestFlight::Build (NameError)`

Note that I am not able to include information other than a screenshot due to the fact that every fastlane command throws this same error.

##### Complete output when running fastlane, including the stack trace and command used
![screen shot 2018-01-08 at 3 44 48 pm](https://user-images.githubusercontent.com/8161781/34691548-e876ad08-f48a-11e7-95ae-da2b9857e2f8.png)

taquitos:
I'm trying to repro this now

taquitos:
@michaelgmcd I can't repro on my setup, maybe you can try:
`rvm cleanup all` and followed by 
`rvm all-gemsets do gem cleanup`
this might be caused by stale gems.

michaelgmcd:
Interesting. After running your commands, I still get the error. This has occurred on two mac machines and after installing in an ubuntu docker image.

taquitos:
@michaelgmcd could you provide some simple docker setup instructions on how I can repro this? I'm a docker newb.

itruf:
Same problem right now 👎 

MBP-Ivan-2:client itruf$ fastlane env
[⠋] 🚀 /Users/itruf/.rvm/rubies/ruby-2.3.3/lib/ruby/gems/2.3.0/gems/tty-screen-0.6.3/lib/tty/version.rb:3: warning: already initialized constant TTY::Screen::VERSION
/Users/itruf/.rvm/gems/ruby-2.3.3@global/gems/tty-screen-0.6.3/lib/tty/version.rb:3: warning: previous definition of VERSION was here
[...]

PR #11493, opened 2018-01-09
Fix spaceship crash on some Ruby environments

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/11481


milch (approved):
Hmm, will be interesting to see if this works 👍

KrauseFx:
Yah, I'm a little worried, this will then break some spaceship features later in the fastlane lifecycle...

PR #11500, opened 2018-01-09
Hide loading indicator for CI systems

KrauseFx:
This doesn't include the first rocket that's shown on startup, as it would have to have acess to `FastlaneCore`, which it currently doesn't.


janpio (approved):
nice one

PR #11504, opened 2018-01-09
Be more specific with gem cache key

milch:
fastlane dependencies are spread out over these three files


janpio (approved):

milch:
Hmm, looks like this only works for top-level files. That's ok, I suppose the Pluginfile never actually changes anyway

PR #11506, opened 2018-01-09
Improve behavior of cfpropertylist extension

jdee:
### Checklist

### Motivation and Context

This extension sometimes causes problems when used with RVM gemsets.

### Description

Be careful about removing the `#to_plist` method in case it was already removed.

### DIFF
--- a/fastlane_core/lib/fastlane_core/core_ext/cfpropertylist.rb
+++ b/fastlane_core/lib/fastlane_core/core_ext/cfpropertylist.rb
@@ -12,7 +12,9 @@ require "plist"
 # CFPropertyList also adds Enumerator#to_plist, but there is no such
 # method from Plist, so leave it.
 [Array, Hash].each do |c|
-  c.send(:alias_method, :to_binary_plist, :to_plist)
-  c.send(:remove_method, :to_plist) if c.method_defined?(:remove_method)
+  if c.method_defined?(:to_plist)
+    c.send(:alias_method, :to_binary_plist, :to_plist)
+    c.send(:remove_method, :to_plist)
+  end
   c.module_eval("include Plist::Emit")
 end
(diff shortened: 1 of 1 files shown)


jdee:
@milch @krausefx @snatchev Lost my Slack. I think this might help.

milch:
Yeah Slack is dead I think. Looks like you're checking whether `:remove_method` exists rather than `:to_plist` though?

KrauseFx (approved):
🚀

KrauseFx:
Do we also need the `begin` `rescue` blocks here?

KrauseFx:
Similar to https://github.com/fastlane/fastlane/commit/a879476819c18c896ee64b39225405f64686ba75 I mean

jdee:
Even that doesn't actually fix the problem, which I can still produce. It raises at line 17, despite the check for the existence of the method.

jdee:
That works when rescuing `NameError`. But I really have no idea why it's happening.

I also still see this:

```
[⠋] 🚀 /Users/jdee/.rvm/rubies/ruby-2.4.2/lib/ruby/gems/2.4.0/gems/tty-screen-0.6.4/lib/tty/version.rb:3: warning: already initialized constant TTY::Screen::VERSION
/Users/jdee/.rvm/gems/ruby-2.4.2@global/gems/tty-screen-0.6.4/lib/tty/version.rb:3: warning: previous definition of VERSION was here
[⠦] 🚀 /Users/jdee/.rvm/gems/ruby-2.4.2@global/gems/fastlane-2.75.1/spaceship/lib/spaceship/test_flight/build.rb:87: warning: already initialized constant Spaceship::TestFlight::Build::BUILD_STATES
/Users/jdee/.rvm/rubies/ruby-2.4.2/lib/ruby/gems/2.4.0/gems/fastlane-2.75.1/spaceship/lib/spaceship/test_flight/build.rb:87: warning: previous definition of BUILD_STATES was here
[✔] 🚀 
```

But this seems to fix the crash.

janpio:
@jdee Is this good to merge in your opinion?

bguest:
Can we get this merged in, I'm sick of copying this code into my local gem everytime it brakes, seems like a no brainer...

PR #11507, opened 2018-01-09
[WIP] 2nd batch of require work: require_relative all the things

milch:
Part 2 of #11244

In this episode we:

- Fixed up `supply`
- Wrote a helper that, for each tool, gives a helper `from_tool` that returns the relative path from the file it is being called from to that tool
- Wrote a custom cop for fastlane that does `require 'tool/module'` -> `require_relative from_tool/'module'`
- Ran that thing on all of fastlane

TODO:

- [ ] Appease the `Style/SpacesAroundOperators` cop because `require_relative from_fastlane/'runner'` looks better than `require_relative from_fastlane / 'runner'`
- [ ] Reenable Require/MissingRequireStatement once I figure out how to make this work with that


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

cert/lib/cert/commands_generator.rb
```diff
@@ -1,7 +1,7 @@
 require 'commander'
-require 'fastlane/version'
-require 'fastlane_core/configuration/configuration'
-require 'fastlane_core/globals'
+require_relative from_fastlane/'version'
```
KrauseFx: `from_fastlane`?

milch: Well  I thought using a function to do this was nicer than `require_relative '../../../fastlane/lib/fastlane/version'`. Ideally I would just name the functions the tool name but that would interfere with the actions in a Fastfile, I believe. So I thought this reads pretty nicely: "require relative from fastlane". We can obviously rename it though. 

Could also replace `require_relative from_fastlane/'version'` (using `Pathname`) with a function `require_relative from_fastlane('version')`.

In practice you can just do `require 'fastlane/version'` anyway and it gets automatically replaced to the require_relative version with the cop: https://github.com/fastlane/fastlane/blob/0a3a16d510c28c44a50e9cea4c9dccf0518bc6d1/rubocop/require_relative.rb#L40

milch: Note that one of the advantages of using `require_relative` for everything is that it minimises cross installation pollution. Since the code is 1/2 converted and 1/2 not converted right now on master it seems to happen more often on broken setups, but once this is done and merged it is impossible for a broken installation (e.g. other versions of fastlane, multiple rubies, old versions of non-monogem gems, ...) to affect fastlane

milch:
Did some more preliminary testing, seeing huge speedups with this enabled. 30-40% faster loading... the spinner is basically unnecessary since it disappears instantly

cert/lib/cert/commands_generator.rb
janpio: FYI: I would never understand that this is a function by myself.

milch: @janpio So you would prefer the `from_fastlane('version')` variant?

milch: The helper with the definitions is here by the way: https://github.com/fastlane/fastlane/pull/11507/files#diff-25e7ffc319c07bcabbd06d6ecc53f9eb

janpio: Yes, definitely. I didn't even notice this was not part of the string, and even if I did I would wonder how it works and be confused. Maybe `require_relative_fastlane` to replace `require_relative` in these cases? That would be clearest to me tbh.

milch: So `require_relative_fastlane 'version'`, `require_relative_spaceship 'portal'`, etc. or `require_relative_fastlane 'fastlane/version'`, `require_relative_fastlane 'spaceship/portal'`? 

Maybe something like `require_internal 'spaceship/portal'` would be more clear as well?

janpio: Yep, that sounds good to me. Then the only thing I could miss is that this is not a native ruby function but something fastlane defines, but actually wouldn't care too much because it is close enough to the native thing in functionality (Is it?).

milch: You'll still be able to do `require 'fastlane/version'`, it would just automatically be replaced by the rubocop when you do `rubocop -a` before submitting, so you don't really need to worry about anything 😄 

Really it just uses the native `require_relative` anyway, but computes the relative path to the other tool for you. Otherwise it can be quite cumbersome to do this manually, and results in long paths like `../../../../fastlane_core/lib/fastlane_core/ui/ui`. This wouldn't be as much of an issue if fastlane didn't have the old directory structure back from when everything used to be separate gems, but right now we have these deep folder trees. 

Maybe once this is complete we can actually do away with the old hierarchy, since changing it will amount to only having to rewrite one function.

janpio: > so you don't really need to worry about anything 😄

Well, I still have to (maybe, at least a bit) understand the code I (or `rubocop -a`) write and commit there...

KrauseFx (approved):
Let's do this thing 🚀

KrauseFx:
Oh tests 😢

KrauseFx:
> the spinner is basically unnecessary since it disappears instantly

OMG 😍

milch:
I have no idea what happened here... I'll just close and reopen this one.

janpio:
Uh @milch, was this ever reopened or completely abandoned?

milch:
Hmm, I remember reopening at some point, but never merging it due to ... issues ... 

In any case, I'm now clear to pick this up again. I'll take a look this weekend!

janpio:
As I'm experimenting with performance optimization stuff right now I rebased your branch onto `master`, went through the merge conflicts and created a new PR to see if it actually succeeds in tests: https://github.com/fastlane/fastlane/pull/13458 It does not, and I also left several comment of missing or additional `require`s.

PR #11512, opened 2018-01-10
Enable Swift FastlaneRunner to be updated automatically 🎉

taquitos:
Currently, when you make changes to fastlane API and we regenerate `Fastlane.swift` there is no way to actually get that file into users' `FastlaneRunner` projects.

Not only this file, but also these files (currently):
`Fastlane.swift`,`DeliverfileProtocol.swift`,`GymfileProtocol.swift`,`MatchfileProtocol.swift`,`PrecheckfileProtocol.swift`,`ScanfileProtocol.swift`,`ScreengrabfileProtocol.swift`,`SnapshotfileProtocol.swift`,`LaneFileProtocol.swift`,`ControlCommand.swift`,`RubyCommand.swift`,`RubyCommandable.swift`,`Runner.swift`,`SocketClient.swift`,`SocketClientDelegateProtocol.swift`,`SocketResponse.swift`,`ArgumentProcessor.swift`,`main.swift`, and `RunnerArgument.swift`

These belong to different groups, and as we make changes/fixes we'll be adding other files, and possibly other groups to the Xcode project.

This PR automates the process of updating users' projects so that new files/groups are added, and file changes are applied as well.

This currently only supports overwriting a set of files that are generated, and will not remove/update any of the users' own files.

- Updated deployment script to generate a required Swift file manifest

Current logic is as follows:
- find all groups in existing project
  - if a group is missing, add it
    - add all files for group into new group, build target, and compile phase
- iterate through existing groups
  - update all files needing updating by checking version values first, then checking actual contents
- iterate through existing groups
  - if a file from the manifest is new, add it to group, build target, and compile phase
- Save and return true if any action was taken to modify any file (project included)


fastlane/Fastfile
```diff
@@ -150,6 +154,33 @@ lane(:generate_swift_api) do |options|
...
 
+desc("Generate the FastlaneSwiftRunner project manifest")
+lane(:generate_fastlane_swift_runner_manifest) do |options|
+  # Don't attempt to upgrade config files, like Fastfile.swift, Deliverfile.swift, etc...
+  non_upgrade_set = TOOL_CONFIG_FILES.map { |tool_name| "#{tool_name}.swift" }.to_set
+
+  swift_folder = File.expand_path("swift")
+  runner_project_path = File.expand_path(File.join(swift_folder, "/FastlaneSwiftRunner/FastlaneSwiftRunner.xcodeproj"))
+  project = Xcodeproj::Project.open(runner_project_path)
+  swift_files = project.files.select do |file|
```
KrauseFx: Maybe this piece is easier to read if we have multiple `.select do` blocks? It would take longer to run, but this should be done in less than a second anyway, right?

fastlane/lib/fastlane/swift_lane_manager.rb
```diff
@@ -54,6 +62,11 @@ def self.cruise_lane(lane, parameters = nil, env = nil)
       duration = ((Time.now - started) / 60.0).round
 
       finish_fastlane(nil, duration, e, skip_message: skip_message)
+
+      if display_upgraded_message
+        UI.message("We updated your FastlaneRunner project during this run to make it compatible with your current version of fastlane.")
+        UI.message("Please make sure to check the changes into source control.")
```
KrauseFx: What happens after that? That's during a normal _fastlane_ run, right? Will the user be able to see this, or does it pretty much immediately disappear because the actual build output will be printed out?

KrauseFx:
This looks great. I don't think I fully understand the user flow here. 

User runs `fastlane beta` with outdated Swift support files. What happens? From my understanding, unless this flag is used, we automatically upgrade the files, telling the user to check them into git, but as part of the build output, so probably something they might miss.

**Questions:**
- how does this work on CI? Should we always skip the upgrade when `Helper.ci?` is true? As the user will never check it in from there. Also CI shouldn't modify source code
- Instead of automatically upgrading the files for the user, could we show a message at the very end of the _fastlane_ output, telling them to run `fastlane upgrade_swift_files` or similar?

With _fastlane_ we often try to avoid modifying files on behalf of the user automatically, we usually either tell them what command to run, or how to modify certain files/projects to get them where they want to be

taquitos:
@KrauseFx you bring up a great point about modifying the user's projects. Traditionally we don't and I totally agree with that.

Here's why I think it's ok for us to do it in this circumstance:
- `FastlaneSwiftRunner.xcodeproj` is a project we supply
  - Any changes will not impact the user's app
  - This project builds a core piece of communication and is 100% required to be updated sometimes
- We have really strong messaging around what files may be overwritten, and what not to do:
  - https://github.com/fastlane/fastlane/pull/11512/files#diff-d4bddcd1602a1495ce67bd9b9faf06d2 is included in every group
- We have an upgrade manifest file that includes the list of files we're going to update, and it's very targeted.
  - Never update configuration files (`Toolnamefile.swift`)
  - Never delete a file
  - Never delete a group
  - Only update files we've generated
- If the user doesn't update, fastlane will likely fail.

I'll respond to your other comments inline

fastlane/lib/fastlane/swift_lane_manager.rb
taquitos: This happens at the end of the run 🎉 

screen shot 2018-01-10 at 10 25 28 am

taquitos:
> User runs fastlane beta with outdated Swift support files. What happens? 

It could work, but if defaults were changed, params moved, or updated, it will fail on the ruby side. 
The worst case is that we change our protocol and make it backwards incompatible (like I did in 1.74.0). Then the client-server comms would just fail entirely.

> how does this work on CI? 

I forgot about CI, I agree with you:

> always skip the upgrade when Helper.ci? is true

because these changes shouldn't happen at CI time, it should happen in a developer's workflow. I'm going to update it so that we don't do it during CI.

Also, I think I'll add a confirm step to the upgrade with more messaging and with a default of "yes"

revolter:
Worst case scenario would be if we change some default values and it __doesn't__ fail but does incorrect things instead, IMO.

KrauseFx (approved):

fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/swift/Fastlane.swift
```diff
@@ -460,7 +460,7 @@ func buildApp(workspace: String? = nil,
               scheme: String? = nil,
               clean: Bool = false,
               outputDirectory: String = ".",
-              outputName: String? = nil,
+              outputName: String = "TACOS",
```
taquitos: ugh. test code

Issue #11464, opened 2018-01-07
[frameit] Title has very small font and aligned to right


jkleroy:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I recently updated fastlane on my mac and ever since frameit has issues.
I am trying to update screenshots for the App Store but the title now appears tiny and aligned to right.
I attached two screenshots, one produced a few months ago, another with the latest version of fastlane.

Current output: 
![1_packing_framed](https://user-images.githubusercontent.com/7473932/34655306-9ad6e94c-f407-11e7-81c1-a7029fc6396a.png)

Output a few months ago: 
![1_packing_framed](https://user-images.githubusercontent.com/7473932/34655315-b5ee289e-f407-11e7-8009-d72b45a65f5c.png)

##### Complete output when running fastlane, including the stack trace and command used

```
frameit silver
[✔] 🚀 
[00:02:00]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[00:02:00]: Using device frames version 'latest'
[00:02:00]: Framing screenshot './no/1_Packing.png'
Forced orientation: 
Forced orientation: 
identify: geometry does not contain image `/var/folders/_c/_4wt3z4d0wndqw9kc6yxk1kr0000gn/T/mini_magick20180108-28651-16ehgel.png' @ warning/attribute.c/GetImageBoundingBox/240.
mogrify: geometry does not contain image `/var/folders/_c/_4wt3z4d0wndqw9kc6yxk1kr0000gn/T/mini_magick20180108-28651-16ehgel.png' @ warning/transform.c/CropImage/640.
composite: invalid profile length `/var/folders/_c/_4wt3z4d0wndqw9kc6yxk1kr0000gn/T/mini_magick20180108-28651-lk83s.png' @ warning/png.c/MagickPNGWarningHandler/1705.
Forced orientation: 
Forced orientation: 
Forced orientation: 
[...]

PR #11517, opened 2018-01-10
Improve "Testing" part of YourFirstPR.md

janpio:
### Motivation and Context
Testing instructions right now only prominently show running the whole test suite which takes quite some time which I noticed when replying to https://github.com/fastlane/fastlane/issues/11464#issuecomment-356604527

### Description
- Add instruction on how to run tests from one specific file
- Add subheadlines to Testing part for more structure
- Change code formatting so alternative testing commands are more prominent

### DIFF
--- a/YourFirstPR.md
+++ b/YourFirstPR.md
@@ -42,13 +42,31 @@ You can also run those steps independently or on a more fine grained way.
 
 Make sure to run the automated tests using `bundle exec` to ensure you’re running the correct version of `rspec` and `rubocop`
 
-First, navigate into the root of the _fastlane_ project and run unit tests using
+#### All unit tests
+
+First, navigate into the root of the _fastlane_ project and run all unit tests using
 
 ```
 bundle exec rspec
 ```
 
-If you want to run tests only for one tool, use `bundle exec rspec [tool_name]`
+#### Unit tests for one specific tool
+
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Nice 🎉

YourFirstPR.md
```diff
@@ -42,13 +42,31 @@ You can also run those steps independently or on a more fine grained way.
...
+```
+bundle exec rspec [tool_name]
+```
+
+#### Unit tests in one specific test file
+
+If you know exactly which `_spec.rc` file you want to run, use
+
+```
+bundle exec rspec .\fastlane\spec\fastlane_require_spec.rb
```
revolter: Hmmm, Windows' back slashes, right?

janpio:
In Windows both work, I think this happened via auto completion. Doesn't work for you?

YourFirstPR.md
janpio: In Windows both work, I think this happened via auto completion.

revolter: No, on macOS and Linux, the separator is `/`, so I think it completely removes `\`:

> LoadError: cannot load such file -- /Users/iulian/Development/GitHub/fastlane/fastlane/.fastlanespecfastlane_require_spec.rb

PR #11522, opened 2018-01-10
Enable nuking of Enterprise account and warn the user beforehand

vburojevic:
### Checklist

### Motivation and Context

Nuking of the enterprise account is currently disabled. This PR will enable it but user will have to confirm this action beforehand.

### Description

Issue reference: #11463

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
+++ b/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
@@ -292,6 +292,7 @@ To revoke all certificates and provisioning profiles for a specific environment:
 ```no-highlight
 fastlane match nuke development
 fastlane match nuke distribution
+fastlane match nuke enterprise
 ```
 
 <img src="/img/actions/match_nuke.gif" width="550" />
--- a/match/lib/match/commands_generator.rb
+++ b/match/lib/match/commands_generator.rb
@@ -122,11 +122,11 @@ module Match
         c.syntax = "fastlane match nuke"
         c.description = "Delete all certificates and provisioning profiles from the Apple Dev Portal"
         c.action do |args, options|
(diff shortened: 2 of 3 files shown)


vburojevic:
I signed it!

KrauseFx (approved):
That's awesome, thanks for your contribution @vburojevic. Did you test this on your local machine?

vburojevic:
@KrauseFx I've changed the code in the latest commits to warn the user only if it's a distribution enterprise nuke and modified the message accordingly. I think it's not necessary when nuking development because it doesn't affect in-house apps. I've also improved the code to exit with user error if the user doesn't confirm. Seems ok to you?

I've tested the warning prompt code but didn't go through with nuking, only printed the tables because I stumbled upon an issue.

It seems that current code only fetches ad-hoc certs and profiles on Enterprise acccount when running with distribution parameter. It doesn't include enterprise certificates and provisioning profiles. But there are only 2 viable parameters now when running nuke: `development` or `distribution`, defined in commands_generator.rb.

So, I've added a 3rd option for enterprise nuke that will fetch and nuke enterprise certificates and provisioning profiles, and modified the documentation accordingly. I hope that's fine, let me know what you think.

Also, if you think we can merge this, I'll nuke our account over the weekend. So I can test everything with actual nuking just to confirm everything runs smoothly and report back here before the merge.

vburojevic:
But maybe it's better to include enterprise provisioning profiles and certificates as part of `match nuke distribution` and leave only `development` and `distribution` options? 🤔

janpio:
Adding the third option seems like a good idea to make sure people know what they are doing.

KrauseFx (approved):
Looks legit, nice change. Thanks for your contribution @vburojevic 👍

Issue #11524, opened 2018-01-10
Improve setup command timeout to increase odds of success


danielpunkass:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

(I tried to use capture_output but I don't think it's pertinent because my case has to do with fastlane init).

Upon trying fastlane for the first time against one of my Xcode projects, it failed to list schemes by timing out repeatedly with the 10 second default timeout.

Although it helpfully suggested I could increase the timeout by setting an env variable, I thought it was not helpful that the timeout remained the same, limited to 10 seconds, on each successive retry.

I poked around and ended up coming up with a proposed solution that I think will make fastlane more likely to succeed on the first try for more users.

Forthcoming pull request includes a fix that starts with much lower default timeout values, but doubles them on each successive iteration.

Sorry I didn't write the bug first. I was just hacking on it and thought I would propose the change in case you find it reasonable. If not, not sweat!

I'm also open to making changes if you think you might want to take the fix, but aren't happy with how I've approached it.

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.13.2                                                        |
| Ruby                        | 2.3.3                                                          |
| Bundler?                    | true                                                           |
| Git                         | git version 2.14.3 (Apple Git-98)                              |
| Installation Source         | /Library/Ruby/Gems/2.3.0/bin/fastlane                          |
| Host                        | Mac OS X 10.13.2 (17C88)                                       |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib |
| OpenSSL Version             | LibreSSL 2.2.7                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | ~/Applications/Xcode/Xcode.app/Contents/Developer/             |
| Xcode Version               | 9.2                                                            |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
[...]

PR #11526, opened 2018-01-10
Improve command timeouts

danielpunkass:
### Checklist

### Motivation and Context
This proposed change improves the timeout behavior when executing commands that extract schemes and build settings from Xcode projects.

By changing from a timeout strategy where a fixed timeout is applied repeatedly to the same command (likely to fail by timing out repeatedly), to one in which the timeout is progressively increased with each iteration, it is more likely that users of fastlane will have a successful initialization of a new project.

This pull request fixes issue #11524.

I tested these changes first by manually confirming that the timeout value increases as expected while repeatedly iterating on one of my more complicated Xcode project files, then by setting an artificially low timeout value for each of FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT and FASTLANE_XCODE_LIST_TIMEOUT, and observing that after exhausting increasingly greater timeouts, the commands fail with as similar failure mode to before.

Finally I updated existing unit tests covering retry and timeout behavior so that they pass in the context of the new behavior.

### Description

1. The base timeout defaults for FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT and FASTLANE_XCODE_LIST_TIMEOUT are each lowered from 10 to 3.

2. On each recurring try to run a command (4 tries by default, unchanged), the timeout is doubled so that by default a command will run with timeouts of 3, 6, 12, and 24 seconds.

3. The changed strategy increases _maximum_ default timeout that will be tried from 10 seconds to 24 seconds, while increasing the _total_ elapsed time from exhausting 4 tries in 40 seconds to 45 seconds.

Typical users will likely see fastlane init succeed in the 3, 6, or 12 second timeout iteration, while users with more complex projects are now much more likely to succeed without having to specify a custom environment variable value.

The choice of a new default "3" is made with some consideration to wanting to preserve the same rough total elapsed time before erroring out (40 seconds), while improving the chances that users whose projects are only moderately complex will reach a successful command completion sooner.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -394,8 +394,8 @@ module FastlaneCore
         retries = FastlaneCore::Project.xcode_list_retries
         @raw = FastlaneCore::Project.run_command(command, timeout: timeout, retries: retries, print: !silent)
       rescue Timeout::Error
-      UI.user_error!("xcodebuild -list timed out after #{retries + 1} retries with a base timeout of #{timeout}. You might need to recreate the user schemes." \
-          " You can override the base timeout value with the environment variable FASTLANE_XCODE_LIST_TIMEOUT")
+        UI.user_error!("xcodebuild -list timed out after #{retries + 1} retries with a base timeout of #{timeout}. You might need to recreate the user schemes." \
+            " You can override the base timeout value with the environment variable FASTLANE_XCODE_LIST_TIMEOUT")
       end
 
       UI.user_error!("Error parsing xcode file using `#{command}`") if @raw.length == 0
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -317,8 +317,8 @@ module FastlaneCore
             UI.error("Could not read build settings. Make sure that the scheme \"#{options[:scheme]}\" is configured for running by going to Product → Scheme → Edit Scheme…, selecting the \"Build\" section, checking the \"Run\" checkbox and closing the scheme window.")
           end
         rescue Timeout::Error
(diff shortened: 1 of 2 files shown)


danielpunkass:
@googlebot I signed the Google CLA.

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -438,22 +438,27 @@ def self.run_command(command, timeout: 0, retries: 0, print: true)
...
       rescue Timeout::Error
         try_limit_reached = try >= total_tries
 
-        message = "Command timed out after #{timeout} seconds on try #{try} of #{total_tries}"
-        message += ", trying again..." unless try_limit_reached
+        # Try harder on each iteration
+        next_timeout = try_timeout * 2
+
+        message = "Command timed out after #{try_timeout} seconds on try #{try} of #{total_tries}"
+        message += ", trying again with #{next_timeout} second timeout..." unless try_limit_reached
```
lacostej: typo "seconds" ?

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -438,22 +438,27 @@ def self.run_command(command, timeout: 0, retries: 0, print: true)
 
       total_tries = retries + 1
       try = 1
+      try_timeout = timeout
```
lacostej: maybe document the method with the fact that the timeout will be doubled at each retry?

fastlane_core/lib/fastlane_core/project.rb
danielpunkass: Thanks for taking a look. I updated the method's documentation with details about the timeout doubling behavior. I don't know if you thought "second" should be "seconds", but I think it would sound awkward to say e.g. "trying again with 30 seconds timeout". I tweaked it to put "a" before the number for e.g. "trying again with a 30 second timeout". Maybe that is more clear?

KrauseFx (approved):
Solid, thank you for your contribution @danielpunkass 
 👍

PR #11535, opened 2018-01-11
Issue #11497: [screengrab] Adding Falcon as a ScreenshotStrategy

### LINKED ISSUE #11497: [screengrab] Adding Falcon as a ScreenshotStrategy
### New Issue Checklist

### Issue Description

I have been using fastlane+screengrab to take screenshots from my android app. I have been using `UiAutomatorScreenshotStrategy` and some of my screenshot tests are flaky. Some screenshots are taken before the whole ui is ready even though I am using Espresso onView and checks.

I haven't dig into how UiAutomator takes the screenshot but there is a race condition somewhere. Adding a small delay before taking the screenshots fixes all my issues.

While looking for other libraries to take screenshots I found https://github.com/jraska/Falcon
Falcon is a 400 lines dependency which takes screenshots in the main thread.

I am looking forward to add Falcon as a screengrab dependency and create a `ScreenshotStrategy`.

Would that be ok?

Macarse:
Fixes #11497

### DIFF
--- a/screengrab/example/src/androidTest/java/tools/fastlane/localetester/FalconScreenshots.java
+++ b/screengrab/example/src/androidTest/java/tools/fastlane/localetester/FalconScreenshots.java
@@ -1,0 +1,61 @@
+package tools.fastlane.localetester;
+
+import android.support.test.rule.ActivityTestRule;
+
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import tools.fastlane.screengrab.FalconScreenshotStrategy;
+import tools.fastlane.screengrab.Screengrab;
+import tools.fastlane.screengrab.locale.LocaleTestRule;
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+@RunWith(JUnit4.class)
+public class FalconScreenshots {
+
+    @ClassRule
+    public static final LocaleTestRule localeTestRule = new LocaleTestRule();
+
+    @Rule
+    public ActivityTestRule<MainActivity> activityRule = new ActivityTestRule<>(MainActivity.class, false, false);
+
+
+    @Test
+    public void testTakeScreenshot() {
+        activityRule.launchActivity(null);
+        Screengrab.setDefaultScreenshotStrategy(new FalconScreenshotStrategy(activityRule.getActivity()));
+
+        onView(withId(R.id.greeting)).check(matches(isDisplayed()));
+
+        Screengrab.screenshot("falcon_beforeFabClick");
+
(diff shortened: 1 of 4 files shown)


screengrab/example/src/androidTest/java/tools/fastlane/localetester/FalconScreenshots.java
```diff
@@ -0,0 +1,61 @@
...
+    public static final LocaleTestRule localeTestRule = new LocaleTestRule();
+
+    @Rule
+    public ActivityTestRule<MainActivity> activityRule = new ActivityTestRule<>(MainActivity.class, false, false);
+
+
+    @Test
+    public void testTakeScreenshot() {
+        activityRule.launchActivity(null);
+        Screengrab.setDefaultScreenshotStrategy(new FalconScreenshotStrategy(activityRule.getActivity()));
```
Macarse: @janpio I can't see a way around setting the default strategy on every test, since Falcon requires the activity.

janpio: Interesting, How does this actually work inside? Does it just expect a `this.activity` to be present?

janpio:
So is the result as hoped and expected in #11497?

@fastlane/android, please take a look at this and its parent issue.

screengrab/example/src/androidTest/java/tools/fastlane/localetester/FalconScreenshots.java
Macarse: here's what they do:
https://github.com/jraska/Falcon/blob/master/falcon/src/main/java/com/jraska/falcon/Falcon.java#L186

Macarse:
@janpio any news?

janpio:
If you say this is working, I think we can just go ahead and merge it.

But: Could you add some documentation somewhere? Otherwise nobody will know about it.

Macarse:
@janpio I would prefer that someone else review and test. There are no android devs maintaining fastlane?

taquitos:
@Macarse we're hoping to find some contributors who have more experience around Android than we currently have 😞 
fastlane traditionally has been very iOS focused since that ecosystem has had the largest challenges.

screengrab/screengrab-lib/build.gradle
```diff
@@ -51,4 +51,5 @@ dependencies {
     compile 'com.android.support.test:rules:0.5'
     compile 'com.android.support.test.espresso:espresso-core:2.2.2'
     compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
+    compile 'com.jraska:falcon:1.0.4'
```
janpio: Hey @Macarse, is there maybe a way to only require this if the Falcon thing is actually used, so do it on runtime? Would This is the only change that will affect users not using this strategy, and would make merging much easier of course.

Macarse: @janpio I think fastlane should include falcon as a dependency. If Falcon is not added, using the uiautomator implementation will have race condition bugs.

Macarse: Honestly, I would drop the other 2 implementations and force everyone to use Falcon.

joshdholtz:
@Macarse Looking into this for you! 👀

joshdholtz (approved):
This is fantastic 💯 I tested it out with both the UiAutomater and Falcon and both worked. Going to merge this, get a new build out to bintray(hopefully this week), and then I will update the docs once that's out 👍

PR #11548, opened 2018-01-12
match/sigh/sync_code_signing: Improving documentation on profile templates.

### LINKED ISSUE #11049: [match] Support Custom Entitlements - Apple Pay Pass Suppression

We have 8 Instances of a white labeled app on the appstore, the next version will be supporting apple pay pass suppression, this is a custom entitlement that can only be added to the provisioning profile.

Our current fastlane flow uses match to provision all our apps during CI. Going forward I will need to use a work around to manually create my provisioning profiles at least once per year and then encrypt them and upload them to our provisioning repos as mentioned here: #4548

Would it be possible to add a flag to match to allow custom entitlement support?

Thanks.

lukewakeford:
Improving documentation on profile templates where also known as custom entitlements.

### Checklist

### Motivation and Context
Struggled for over a month trying to add custom entitlements to a fastlane ci setup, only to find out that it was already supported just under a different name.

Resolves #11049 

### Description
Updated docs to mention 'custom entitlements' so that the `template_name` key is searchable using 'custom entitlements'.

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
+++ b/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
@@ -248,6 +248,16 @@ match(app_identifier: ["tools.fastlane.app", "tools.fastlane.app.today_widget"],
 
 _match_ can even use the same one Git repository for all bundle identifiers.
 
+##### Templates (aka: custom entitlements)
+
+Match can generate profiles that contain custom entitlements by passing in the entitlement's name with the `template_name` parameter.
+
+```
+match(git_url: "https://github.com/fastlane/certificates",
+      type: "development",
+      template_name: "Apple Pay Pass Suppression Development")
+```
+
 ### Setup Xcode project
 
 [Docs on how to set up your Xcode project](/codesigning/xcode-project/)
--- a/match/lib/match/options.rb
+++ b/match/lib/match/options.rb
@@ -152,7 +152,7 @@ module Match
                                      end),
         FastlaneCore::ConfigItem.new(key: :template_name,
                                      env_name: "MATCH_PROVISIONING_PROFILE_TEMPLATE_NAME",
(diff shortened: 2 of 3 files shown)


lukewakeford:
I signed it!

janpio:
Did you find anything about these "profile templates" online? Maybe this was only the internal name for it of the person that contributed that code. Then we should probably fully change it to the "official" naming.

How is it used? How can I submit multiple?
Maybe also add an usage example with this so it is clearer?

lukewakeford:
I've searched for usage of 'profile template' in apple documentation etc but with no luck. 
The only place it is used is in the name of the select element in the dev portal `<select name="template">`

The title next to the select tag says 'Entitlements' so I don't think apple know which is the official naming

Entitlements

It's not possible to submit multiple, not from the developer portal (single select element) or the current Fastlane implementation.

I've added a usage example to the docs also to show the value I am using, but I think keeping 'templates' and 'custom entitlements' will help people find this feature. myself included :laughing:

janpio:
Nice.

Do you think it would make sense to add it to the list of examples defined here?
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/sync_code_signing.rb#L85

Also, maybe somewhere in the text of that page? 
https://docs.fastlane.tools/actions/match/

janpio:
ping @lukewakeford

lukewakeford:
Hey! Sorry for the delay, yes I think it should be somewhere in the text of the Match page.
Where does the list of examples get used?

janpio:
I think here: https://docs.fastlane.tools/actions/match/#4-examples

lukewakeford:
Ah cool, yeah I'll take a look at this.

lukewakeford:
I don't think it needs to be in the examples as it's not a very common use case, I have added it to body of that page tho, in the **fastlane** section underneath **Multiple Targets**.

janpio (approved):
💯

PR #11554, opened 2018-01-13
Improve team section in README

revolter:
Related to #11491.

With this:
- Clicking on the anchor next to a name, correctly jumps to the top of the image instead of the name
- Removes conflicting image size (`200` vs `140`)
- Replaces automatic link (added by GitHub) to the image file with a link to the GitHub profile

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,7 +40,7 @@ task(:generate_team_table) do
 
     content << "<tr>" if counter % number_of_rows == 0
     content << "<td id='#{github_user_id}'>"
-    content << "<img src='https://github.com/#{github_user}.png?size=200' width=140>"
+    content << "<img src='https://github.com/#{github_user}.png?size=140'>"
     if user_content['twitter']
       content << "<h4 align='center'><a href='https://twitter.com/#{user_content['twitter']}'>#{github_user_name}</a></h4>"
     else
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -35,14 +35,19 @@ task(:generate_team_table) do
...
   contributors.keys.shuffle.each do |github_user|
     user_content = contributors[github_user]
+    github_user_name = user_content['name']
+    github_user_id = github_user_name.downcase.gsub(' ', '-')
+    github_profile_url = "https://github.com/#{github_user}"
 
     content << "<tr>" if counter % number_of_rows == 0
-    content << "<td>"
-    content << "<img src='https://github.com/#{github_user}.png?size=200' width=140>"
+    content << "<td id='#{github_user_id}'>"
```
lacostej: you could use github_user instead of constructing the id from the name.

revolter: It doesn't work, because the anchor is automatically created by GitHub with that id as `href="#id"`. You can temporarily see this [here](https://github.com/revolter/fastlane/blob/693af3a54cedc2eb6830e5b59abfd6474b30e42a/README.md#j%C3%A9r%C3%B4me-lacoste) _(until I delete that branch)_.

revolter: I also tried [here](https://github.com/revolter/fastlane/blob/656a4f005c134faad003e5a0980bf1ef8a709574/README.md#j%C3%A9r%C3%B4me-lacoste) to create that anchor myself, but GitHub still creates its own anchor.

KrauseFx (approved):
Awesome, thanks 👍

PR #11555, opened 2018-01-13
[deliver] Fix for activating languages

sobakasu:
### Checklist

### Motivation and Context

With `fastlane deliver upload_metadata`, I get the following error on a old iOS project that has never had languages added to it before:
    
    [⠇] Activating languages de-DE, en-US, es-ES, fr-FR, it, ja, ko, ms, pt-PT, ru, sv, zh-Hans... Potential server error received: 'Unexpected Error'. Retrying after 10 seconds (remaining: 5)...

Note that this error doesn't occur when I delete an existing language from ITC and then use `fastlane deliver` to recreate it.  The current code to activate languages works when a language has been previously used, but not on fresh projects.

I tested my changes on a project in ITC that only had an en-US language (primary). The current version of fastlane (2.75.1) failed with the error above. The changed code successfully added the languages.
There doesn't appear to be any existing spec tests for the deliver metadata code.

### Description

The changed code uses the edit details client instead of the edit version client to add languages.

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -235,7 +235,7 @@ module Deliver
       v.languages.each do |language_data|
         lang = language_data["localeCode"]
         enabled_languages[lang] = language_data
-        primary_data = language_data if !language_data["canDeleteLocale"]
+        primary_data = language_data unless language_data["canDeleteLocale"]
       end
 
       UI.user_error!("could not find primary language data for app '#{options[:app].name}'") unless primary_data
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -226,30 +226,47 @@ module Deliver
       # Collect all languages we need
       # We only care about languages from user provided values
       # as the other languages are on iTC already anyway
-      v = options[:app].edit_version
-      UI.user_error!("Could not find a version to edit for app '#{options[:app].name}', the app metadata is read-only currently") unless v
+      v = options[:app].details
+      UI.user_error!("Could not find details for app '#{options[:app].name}'") unless v
 
-      enabled_languages = options[:languages] || []
+      # find languages currently enabled on ITC (enabled_languages)
+      enabled_languages = {}  # map of locale code to language data hash
+      primary_data = nil      # the primary language data, e.g. english name/subtitle
+      v.languages.each do |language_data|
+        lang = language_data["localeCode"]
+        enabled_languages[lang] = language_data
+        primary_data = language_data if !language_data["canDeleteLocale"]
+      end
+
+      UI.user_error!("could not find primary language data for app '#{options[:app].name}'") unless primary_data
+
+      # find list of languages we have data for (local_languages)
+      local_languages = options[:languages] || []
       LOCALISED_VERSION_VALUES.each do |key|
         current = options[key]
         next unless current && current.kind_of?(Hash)
         current.each do |language, value|
-          language = language.to_s
-          enabled_languages << language unless enabled_languages.include?(language)
+          lang = language.to_s
+          next if lang == "default"
+          local_languages << lang unless local_languages.include?(lang)
         end
       end
 
-      # Reject "default" language from getting enabled
-      # because "default" is not an iTC language
-      enabled_languages = enabled_languages.reject do |lang|
-        lang == "default"
-      end.uniq
+      UI.verbose("ITC languages: #{enabled_languages.keys}")
+      UI.verbose("local languages: #{local_languages}")
+      # determine languages to add
+      add_languages = local_languages - enabled_languages.keys
 
-      if enabled_languages.count > 0
-        v.create_languages(enabled_languages)
+      if add_languages.count > 0
+        # we need to add languages to ITC
+        # update v.languages
+        add_languages.each do |locale|
+          v.languages << primary_data.merge("localeCode" => locale, "canDeleteLocale" => true)
+        end
+
         lng_text = "language"
-        lng_text += "s" if enabled_languages.count != 1
-        Helper.show_loading_indicator("Activating #{lng_text} #{enabled_languages.join(', ')}...")
+        lng_text += "s" if add_languages.count != 1
+        Helper.show_loading_indicator("Activating #{lng_text} #{add_languages.join(', ')}...")
         v.save!
         Helper.hide_loading_indicator
       end


sobakasu:
signed cla

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -226,30 +226,47 @@ def verify_available_languages!(options)
       # Collect all languages we need
       # We only care about languages from user provided values
       # as the other languages are on iTC already anyway
-      v = options[:app].edit_version
-      UI.user_error!("Could not find a version to edit for app '#{options[:app].name}', the app metadata is read-only currently") unless v
+      v = options[:app].details
```
janpio: why the switch from `.edit_version` to `.details` here?

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -226,30 +226,47 @@ def verify_available_languages!(options)
...
+      UI.user_error!("Could not find details for app '#{options[:app].name}'") unless v
+
+      # find languages currently enabled on ITC (enabled_languages)
+      enabled_languages = {}  # map of locale code to language data hash
+      primary_data = nil      # the primary language data, e.g. english name/subtitle
```
janpio: `primary_language` or `main_language` maybe?
(or `primary_language_data`)

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -226,30 +226,47 @@ def verify_available_languages!(options)
...
+        lang = language_data["localeCode"]
+        enabled_languages[lang] = language_data
+        primary_data = language_data unless language_data["canDeleteLocale"]
+      end
 
-      enabled_languages = options[:languages] || []
+      UI.user_error!("could not find primary language data for app '#{options[:app].name}'") unless primary_data
+
+      # find list of languages we have data for (local_languages)
+      local_languages = options[:languages] || []
```
janpio: what does `local` refer to here?

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -226,30 +226,47 @@ def verify_available_languages!(options)
...
 
-      # Reject "default" language from getting enabled
-      # because "default" is not an iTC language
-      enabled_languages = enabled_languages.reject do |lang|
-        lang == "default"
-      end.uniq
+      UI.verbose("ITC languages: #{enabled_languages.keys}")
+      UI.verbose("local languages: #{local_languages}")
+      # determine languages to add
+      add_languages = local_languages - enabled_languages.keys
```
janpio: this removes all the languages already in iTC from the "local" list?
ruby sure is nice sometimes...

janpio: but why the `.keys` here? do `local_languages` and `enabled_languages` have a different structure?

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -226,30 +226,47 @@ def verify_available_languages!(options)
...
+
+      if add_languages.count > 0
+        # we need to add languages to ITC
+        # update v.languages
+        add_languages.each do |locale|
+          v.languages << primary_data.merge("localeCode" => locale, "canDeleteLocale" => true)
```
janpio: huh what does this do exactly? `primary_data` is unexpected here.

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -226,30 +226,47 @@ def verify_available_languages!(options)
...
+        end
 
-      if enabled_languages.count > 0
-        v.create_languages(enabled_languages)
```
janpio: Where did this method come from? Why don't we need to call it any more?

janpio (commented):
Hey, nice first PR!

Left some pretty noob comments on the review trying to understand what it does and why. Would be great if you could reply to them @sobakasu

deliver/lib/deliver/upload_metadata.rb
sobakasu: I think I looked at what the ITC web interface did when you added a new language, and the request seemed to correspond with what .details was doing, rather than .edit_version .. maybe. sorry my memory is only reliable up to 7 days max

deliver/lib/deliver/upload_metadata.rb
sobakasu: i think primary data is a hash of information representing the settings for the primary language/locale

deliver/lib/deliver/upload_metadata.rb
sobakasu: here i was using `local` to refer to the list of languages defined under .fastlane/metadata

deliver/lib/deliver/upload_metadata.rb
sobakasu: looks like `local_languages` is a list of locale/languages and `enabled_languages` is a hash with locale/language keys. the result `add_languages` is a list of locale/languages that need to be added

deliver/lib/deliver/upload_metadata.rb
sobakasu: `primary_data` is a hash with settings e.g. for english. this code just copies the primary data to the new language, and changes locale code to the target locale/language. the data is later overwritten by the normal fastlane upload process

deliver/lib/deliver/upload_metadata.rb
sobakasu: i think the code I have written that updates `v.languages` has the effect of creating the new languages. not sure what the old code does exactly or why it wasn't working for me.

deliver/lib/deliver/upload_metadata.rb
janpio: ah, so you use `primary_data` as a kind of "template" where you add/change two values for the language to be added, correct?

sobakasu: yes that's right

minuscorp (changes requested):
My main concerns with this PR is the lack of tests. I think that the process of activating new languages is _mockable_ doing the requests needed in the iTC board. Normally I wouldn't mind in approve this additive PRs, but I think this PR modify pieces of code sensitive enough to have new tests covering the expected behavior as well as maintaining the existing one. I'm not confident enough that this will work without the proper tests being added. Can you @sobakasu try to mock some requests in this regard and adding them as tests?

janpio:
Hey @sobakasu, I am currently cleaning out our open PRs and will close this one as a) the code is pretty opaque and complicated (see my comments), b) this part of the code is mostly untested, c) there were no other requests for these changes in the meantime (that I am aware of) and d) there is now a merge conflict.

This doesn't mean your PR was bad in any way, just that it is easier for us right now to close this than to make the effort to integrate it 🙏 

So many thanks for your efforts. Please feel free to create a new issue if you still want to get this fixed. Your branch will be visible here, so someone can definitely build on these changes again.

PR #11582, opened 2018-01-16
Automatic creation + updating of Gemfile on `fastlane init`

### LINKED ISSUE #11546: [proposal] Create a Gemfile for the user when running `fastlane init`
Users using system Ruby without a `Gemfile` cause a lot of our incoming support tickets.

@milch has some good opinions (some would even say the best opinions)

Currently we support the following installation methods (the % is the number of executions done by each installation method)

### Using a `Gemfile` (49%)

Currently we classify this as a more advanced setup, and don't recommend this out of the box

### `gem install fastlane` (30%)

This doesn't use a proper dependency manager, causing lots of setups to break

### `brew cask install fastlane` (17%)

This is great, however requires `homebrew` to be installed, which not all iOS developers have

### `download.fastlane.tools` (5%)

This is basically `homebrew`, but without homebrew, and a few extra manual steps

### Using the Fabric Mac app (0.1%)

This is used by less than 0.1% of the user base

## Main proposal

The idea is to simplify the installation methods and make it easier for the user. As during onboarding most people will run `fastlane init`, the proposal is to also create a `Gemfile` for the user, so they can check it into version control together with the `Fastfile`.

This requires a few things:
1) During `fastlane init`, check if a `Gemfile` already exists, if not create one, if yes, append `fastlane` to it and verify `rubygems.org` is the defined source
2) `fastlane update_fastlane` should use `bundle update fastlane` (which I think it might already do)
3) Automatically run `Bundler.load` (or similar) on _fastlane_ if there is a `Gemfile`

## Secondary proposal

Secondly, we should think about dropping support for hardly used installation methods in the future. In particular, the Fabric Mac app, and `download.fastlane.tools`. Both use basically the `homebrew` zip file, so it's not a big deal to maintain. But at least we probably don't want to mention them in the docs. We're gonna leave it around for a bit, and see if they cause problems in the future.

`gem install fastlane` is "supported" by default, but we're just not gonna mention it in the docs

KrauseFx:
Fixes #11546

**Done**
- If fastlane already set up, but no Gemfile, offer to create one
- More testing

**Follow-up PR**
- We have code that mentions `To get started with a Gemfile, run`, we should unify this, and make it all as part of `fastlane init`

Related docs https://github.com/fastlane/fastlane/issues/11586

screen shot 2018-01-16 at 2 52 19 pm

### DIFF
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -12,6 +12,10 @@ module Fastlane
         ARGV.include?('-h') || ARGV.include?('--help')
       end
 
+      def running_init_command?
+        ARGV.include?("init")
+      end
+
       def take_off
         before_import_time = Time.now
 
@@ -37,7 +41,7 @@ module Fastlane
           require "fastlane"
         end
         # We want to avoid printing output other than the version number if we are running `fastlane -v`
-        unless running_version_command?
+        unless running_version_command? || running_init_command?
           print_bundle_exec_warning(is_slow: (Time.now - before_import_time > 3))
         end
 
--- a/fastlane/lib/fastlane/setup/setup.rb
+++ b/fastlane/lib/fastlane/setup/setup.rb
@@ -42,7 +42,7 @@ module Fastlane
         UI.message("")
 
         setup_ios = self.new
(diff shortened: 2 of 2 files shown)


taquitos (approved):
I ❤️ this
🎈:goat:

PR #11588, opened 2018-01-16
[Proposal] Run `Bundler.setup` for the user

KrauseFx:
This worked on my local machine, but I probably didn't consider everything. Here are some thoughts:
1. Will this work with plugins?
1. Will this actually make it possible for people to use _fastlane_ without using `bundle exec` ever time? 
1. Does this change then actually resolve the issues we had with conflicting versions? 
1. Does this have any side effects?

### DIFF
--- a/bin/fastlane
+++ b/bin/fastlane
@@ -4,6 +4,17 @@ if RUBY_VERSION < '2.0.0'
   abort("fastlane requires Ruby 2.0.0 or higher")
 end
 
+# We recommend people use bundler together with a Gemfile
+# but we can't really force them. To make it easier for them
+# we don't want to tell every user to run every fastlane command
+# using `bundle exec`, so instead we'll try to load it for them
+require 'bundler'
+begin
+  Bundler.setup
+rescue => ex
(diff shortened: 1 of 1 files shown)


KrauseFx:
Ok tried this, and it has side effects and doesn't actually work.

```
 ~  Developer  fastlane  KrauseFx-patch-4  5✎  %  fastlane lanes
/Users/fkrause/.rbenv/versions/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `require': cannot load such file -- tty-spinner (LoadError)
	from /Users/fkrause/.rbenv/versions/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `rescue in require'
	from /Users/fkrause/.rbenv/versions/2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:40:in `require'
	from /Users/fkrause/Developer/fastlane/fastlane/lib/fastlane/cli_tools_distributor.rb:27:in `take_off'
```

PR #11593, opened 2018-01-16
Fix broken and old links

revolter:
### Checklist

### DIFF
--- a/VISION.md
+++ b/VISION.md
@@ -20,22 +20,22 @@ With these challenges in mind, [_fastlane_ plugin system](https://fabric.io/blog
 
 Each _fastlane_ tool has a specific purpose and should be kept focused on the functionality required for that task.
 
-* [deliver](https://github.com/fastlane/fastlane/tree/master/deliver): Upload screenshots, metadata, and your app binary to the App Store
-* [supply](https://github.com/fastlane/fastlane/tree/master/supply): Upload your Android app and its metadata to Google Play
-* [snapshot](https://github.com/fastlane/fastlane/tree/master/snapshot): Automate taking localized screenshots of your iOS apps on every device
-* [screengrab](https://github.com/fastlane/fastlane/tree/master/screengrab): Automate taking localized screenshots of your Android app on every device
-* [frameit](https://github.com/fastlane/fastlane/tree/master/frameit): Quickly put your screenshots into the right device frames
-* [pem](https://github.com/fastlane/fastlane/tree/master/pem): Automatically generate and renew your push notification certificates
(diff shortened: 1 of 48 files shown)


VISION.md
```diff
@@ -20,22 +20,22 @@ With these challenges in mind, [_fastlane_ plugin system](https://fabric.io/blog
 
 Each _fastlane_ tool has a specific purpose and should be kept focused on the functionality required for that task.
 
-* [deliver](https://github.com/fastlane/fastlane/tree/master/deliver): Upload screenshots, metadata, and your app binary to the App Store
```
janpio: Tools and actions are technically not the same thing... I know before those had the same names, but now the tool named actions are only aliases. (I personally like that the "code collections" that are the tools still exist)

fastlane/lib/fastlane/actions/docs/sync_code_signing.md
```diff
@@ -56,7 +56,7 @@ Before starting to use _match_, make sure to read the [codesigning.guide](https:
 💥  | Easily reset your existing profiles and certificates if your current account has expired or invalid profiles
 ♻️  | Automatically renew your provisioning profiles to include all your devices using the `--force` option
 👥  | Support for multiple Apple accounts and multiple teams
-✨ | Tightly integrated with [fastlane](https://fastlane.tools) to work seamlessly with [gym](https://docs.fastlane.tools/actions/gym) and other build tools
+✨ | Tightly integrated with [fastlane](https://fastlane.tools) to work seamlessly with [gym](https://docs.fastlane.tools/actions/gym/) and other build tools
```
janpio: General question: Is this a recent change that the action docs all end with a slash?

VISION.md
revolter: I don't quite understand what should be changed here.

fastlane/lib/fastlane/actions/docs/sync_code_signing.md
revolter: I think it's from the beginning. If you visit https://docs.fastlane.tools/actions/gym, a trailing slash (ugly IMHO) will be added, so I decided to add it too to links missing it. Also because they were not consistent, some of them had it, some not.

VISION.md
janpio: The tool `deliver` (and all the others) and its code still live at Github. Only the documentation of its action moved to the docs.

revolter: So should `VISION.md` be completely reverted? If yes, are there other files in this case?

janpio: I personally would revert these, yes.

janpio: I think these links go to source code of tools, not documentation of actions.

KrauseFx (approved):
Beautiful 😍

Issue #10059, opened 2017-08-15
REDO #9405 cert checker aborts execution if the desired certificate is already present


BanzaiMan:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

`fastlane match` will fail if WWDRCA certificate is already installed in the keychain.

In https://github.com/fastlane/fastlane/blob/1240fe39eaf236eb9774ec985a437b123cb10aa7/fastlane_core/lib/fastlane_core/cert_checker.rb#L64-L65, you can see that fastlane attempts to download https://developer.apple.com/certificationauthority/AppleWWDRCA.cer and install it in a keychain that the user has access to. It is assumed here that the nonzero exit status means the *failure* to import the certificate, but it does not account for the possibility the nonzero exit because the key already exists.

I believe the error manifests now on Travis CI, because we added the WWDRCA certificate to our login keychain.

See https://github.com/travis-ci/travis-ci/issues/7870#issuecomment-306892418 for more context.
 

### Environment

Sorry, I don't use fastlane to produce this.

BanzaiMan:
I'm just redoing https://github.com/fastlane/fastlane/issues/9405.

taquitos:
Given your output from the closed issue: 
```sh-session
Traviss-Mac-270:~ travis$ curl -O https://developer.apple.com/certificationauthority/AppleWWDRCA.cer
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1062  100  1062    0     0    853      0  0:00:01  0:00:01 --:--:--   854
Traviss-Mac-270:~ travis$ security import AppleWWDRCA.cer /Users/travis/Library/Keychains/login.keychain-db
1 certificate imported.
Traviss-Mac-270:~ travis$ security import AppleWWDRCA.cer /Users/travis/Library/Keychains/login.keychain-db
security: SecKeychainItemImport: The specified item already exists in the keychain.
Traviss-Mac-270:~ travis$ echo $?
1
Traviss-Mac-270:~ travis$ security delete-certificate -c "Apple Worldwide Developer Relations Certification Authority"
Traviss-Mac-270:~ travis$ security import AppleWWDRCA.cer /Users/travis/Library/Keychains/login.keychain-db
1 certificate imported.
[...]

PR #11594, opened 2018-01-16
Fix for Travis CI not installing WWDRC how we expect it

taquitos:
- Travic CI doesn't seem to install the WWDRC cert how we expect
 so attempt to install it, and if we get a failure during install
 check the reason. If it's because the cert exists, then skip.

This is a workaround for #10059

### DIFF
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -63,8 +63,22 @@ module FastlaneCore
       filename = file.path
       keychain = wwdr_keychain
       keychain = "-k #{keychain.shellescape}" unless keychain.empty?
-      Helper.backticks("curl -f -o #{filename} #{url} && security import #{filename} #{keychain}", print: FastlaneCore::Globals.verbose?)
-      UI.user_error!("Could not install WWDR certificate") unless $?.success?
+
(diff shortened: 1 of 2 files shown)


fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -63,8 +63,22 @@ def self.install_wwdr_certificate
...
+      stdout, stderr, status = Open3.capture3("curl -f -o #{filename} #{url} && security import #{filename} #{keychain}")
+      if FastlaneCore::Globals.verbose?
+        UI.command_output(stdout)
+        UI.command_output(stderr)
+      end
+
+      unless status.success?
+        UI.verbose("Failed to install WWDR Certificate, checking output to see why")
+        # Check the command output, WWDR might already exist
+        unless /The specified item already exists in the keychain./ =~ stderr
```
KrauseFx: Missing a link to the issue with some context

KrauseFx (approved):
Looks legit 👍

fastlane_core/lib/fastlane_core/cert_checker.rb
taquitos: Link to issue is in PR description and commit 🤔

PR #11602, opened 2018-01-17
Sort candidate_builds by upload_date in Spaceship::Tunes::AppVersion

### LINKED ISSUE #11420: Spaceship::Tunes::AppVersion.current_build_number is incorrect
AppVersion `current_build_number` doesn't return the latest submitted build.
Specifically it takes the very first candidate build for edit version and candidate builds are not upload-time sorted (it turns out).

https://github.com/fastlane/fastlane/blob/7171725879ab8d4e78d501fbc663d0e59a0490eb/spaceship/lib/spaceship/tunes/app_version.rb#L315-L323

probably something like:
```ruby
def current_build_number 
   if self.is_live? 
     build_version 
   else 
     if candidate_builds.length > 0 
       candidate_builds.sort_by(&:upload_date).last.build_version
     end 
   end 
 end 
```

ambientlight:
### Checklist

### Motivation and Context
Fixes #11420.

### Description
Unlike https://github.com/fastlane/fastlane/blob/master/spaceship/spec/tunes/fixtures/candiate_builds.json which are time-sorted iTunes Connect doesn't necessarily returns the candidate builds in a time-sorted order, thus sorting by upload_date is required to retrieve the latest version.

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -328,7 +328,7 @@ module Spaceship
           build_version
         else
           if candidate_builds.length > 0
-            candidate_builds.first.build_version
+            candidate_builds.sort_by(&:upload_date).last.build_version
           end
         end
       end


ambientlight:
I signed it!

spaceship/lib/spaceship/tunes/app_version.rb
```diff
@@ -328,7 +328,7 @@ def current_build_number
           build_version
         else
           if candidate_builds.length > 0
-            candidate_builds.first.build_version
+            candidate_builds.sort_by(&:upload_date).last.build_version
```
janpio: my ruby is ... lacking: is there a way to reverse sort this so you don't have to use  `.last` here?

janpio (approved):
Looks like this should work 🥇

spaceship/lib/spaceship/tunes/app_version.rb
ambientlight: my rube is ... same. will something like https://stackoverflow.com/a/2642196/2380455 go fine?

ambientlight: `sort_by().last` on the other side is the simplest and clearest 😃

PR #11609, opened 2018-01-18
Don't override demo_account_required before submit_for_testflight_review!

### LINKED ISSUE #11056: Add groups to pilot distribute_build without disabling demo account
In pilot's `distribute_build` when `groups` or `distribute_external` are set then`uploaded_build.beta_review_info.demo_account_required` is set to `false`.

https://github.com/fastlane/fastlane/blob/680420d25f96c255a3e9e3d7fdbc9a1f59088e88/pilot/lib/pilot/build_manager.rb#L135-L170

Is this necessary? I'd like to automatically add the build to our list of internal testers but a demo account is still required for Apple, there is no sign up page in this business app.

taquitos:
Only set to false if no value is set

fixes #11056

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -145,7 +145,7 @@ module Pilot
       uploaded_build.export_compliance.encryption_updated = false
 
       if options[:groups] || options[:distribute_external]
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
👍

PR #11621, opened 2018-01-19
Fix `fastlane init` crash when having no iOS or Android project

KrauseFx:
This was structured in a weird way, because we wanted to ask if the user wanted to use Swift. Since we don't do that any more, we don't need the code. 

Previously it would crash because we didn't call `choose_swift` from `setup.rb` when no project was found.


KrauseFx:
Feel free to merge if ok

janpio (approved):

PR #11625, opened 2018-01-19
[Fastlane.swift] Support printing out console log from Swift

thii:


janpio:
2 tests are failing because of these changes.

What is the motivation behind these changes, what could this be used for?

fastlane/lib/fastlane/actions/puts.rb
```diff
@@ -13,6 +13,17 @@ def self.description
         "Prints out the given text"
       end
 
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :message,
```
taquitos: I think in ` def self.run(params)` you're going to want to inspect `params` for `params[:message]`
if it exists, basically, something like:
```
unless params[:message].nil?
  UI.message(params[:message]) 
  return
end
 UI.message(params.join(' '))

I think this would fix the test errors too

**Review by taquitos — CHANGES_REQUESTED** [review]

**Comment by taquitos** [comment]
@janpio I recently came across: https://github.com/fastlane/fastlane/issues/11537
I think this PR would fix that issue, is that correct @thii ?

File: fastlane/spec/actions_specs/puts_spec.rb [hunk 2]
```diff
@@ -6,6 +6,11 @@
           puts 'hi'
         end").runner.execute(:test)
       end
+      it "works in Swift" do
```
taquitos: Woo, love that you added a test!

taquitos:
@thii I just made a couple updates, check em out 👀

thii (commented):
👍 Thanks

taquitos (dismissed):

KrauseFx (approved):
Looks solid 👍

KrauseFx:
Thanks everyone 🚀

taquitos:
New PR incoming to create action aliases

Issue #11189, opened 2017-12-12
IAPs pricing not working after Introductory Pricing


alaincaltieri:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Starting since yesterday, on iTunes connect the new `Introductory Pricing` feature has been deployed. This refactors the management of both Pricing and trial period for Auto renewable subscriptions and causes the current version of Spaceship to fail in both retrieving and saving pricing information for auto renewable subscriptions.

Let me know if you'd like some help on this, I'd be happy to contribute.

taquitos:
@alaincaltieri that would be amazing! We have some information on how we usually go about fixing issues like this: https://github.com/fastlane/fastlane/blob/master/ToolsAndDebugging.md
I'm not sure when we'll be able to fix this issue in fastlane, but if you feel up to the challenge of updating the api, we can provide some guidance 👍

alaincaltieri:
Awesome! Thanks for the doc. Will keep you posted with my progress

ahmgeek:
is there any progress on this guys?

alaincaltieri:
Had no chance to have a look at it yet, sorry :(

ahmgeek:
I already fixed id 🤗  will send a PR tomorrow.

untitaker:
@aht's PR partially fixes this by using the new API endpoint where the pricing information moved to. However, we've noticed another change wrt how trials are represented in ITC.

Spaceship's IAPDetails currently maps the `freeTrialDurationType` key to `subscription_free_trial`, but since everything is an initial offer now, this key is no longer used and always set to `null`.

I don't see an easy way to map the new model to the old model and therefore suggest to remove this attribute for now and work on a proper implementation of introductory pricing in a separate issue.

byohay:
I'm working on a fix that will send the request to the new API, both when getting and updating pricing. Will open the PR soon :)

luongm:
The PR made by @byohay should have fixed this issue. Although free trial is still broken because it was migrated after Introductory Pricing too, but that can be a separate issue

PR #11627, opened 2018-01-19
Spaceship: support getting and setting pricing of recurring subscriptions.

byohay:
### Checklist

### Motivation and Context
Since Apple introduced the introductory prices, the API for getting and setting the pricing for auto-renewable has changed. In order to set or get the pricing, a request to a new URL is needed.
This PR solves this issue by adding functionality to send requests to the new URL.
The changes were tested by creating new auto-renewable subscription using Spaceship, and getting the pricing of existing auto-renewable subscription. In addition, I verified that creating a consumable products still works as before.

Solves #11189.

### Description
- `tunes_client` was added two methods: one for setting and another for getting auto-renewable subscriptions pricing.
- The API of `iap`, `iap_detail` and `iap_list` hasn't changed. If there is a need to set or get the pricing information of an auto-renewable subscription, these classes perform an additional call to the methods added to `tunes_client`. That way the user of these classes will not have to take extra steps to get or set the pricing, as it was before Apple changed the API.


spaceship/lib/spaceship/tunes/iap.rb
```diff
@@ -75,6 +75,17 @@ def create!(type: "consumable",
                            family_id: family_id,
                            subscription_duration: subscription_duration,
                            subscription_free_trial: subscription_free_trial)
+
+        # Update pricing for a recurring subscription.
+        if type == Spaceship::Tunes::IAPType::RECURRING && pricing_intervals
+          # There are cases where the product that was just created is not immediately found,
+          # and in order to update its pricing the purchase_id is needed. Therefore polling is done
+          # until it is found.
+          product = find(product_id) until product
```
janpio: I don't know the codebase, but are there other parts of the code that use the same method for similar problems?

byohay:
I signed it!

spaceship/lib/spaceship/tunes/iap.rb
byohay: Not that I know of, which sounds weird to me because creating a product and updating it immediately afterwards seems like a common scenario.
Another solution I came up with is to sleep for a few seconds. I tried it, but the amount needed to sleep might change from one connection to another. Note that `find(product_id)` itself takes a second or so, so there won't be so many URL requests and the polling isn't that consuming.
BTW just wanted to make sure: I assumed that if the product's creation failed an exception will be raised (otherwise the polling will be stuck forever).

janpio: No idea, I know nothing about that part of the codebase, but that was one of the things I was wondering as well. Or what happens if the product just doesn't appear because of whatever.

byohay: You can see in `handle_itc_response` that if there was an error an exception is raised. I created a few dozen products with the new Spaceship API already and it never got stuck: I always got an exception or it was created successfully. Do you think I should add a timeout?

janpio: Nah I think this is good until it really breaks because of that. No unneeded complexity - this is all hard enough already.

luongm (changes requested):
Looks good, a few comments.

Also, can you take a look at the free trial change as well? looks like they moved it from Free Trial dropdown to the Introductory Pricing page.

spaceship/lib/spaceship/tunes/iap.rb
untitaker: If this method is called with a nonexistent `product_id`, wouldn't this result in an infinite loop since the product is always not there?

spaceship/lib/spaceship/tunes/iap_list.rb
```diff
@@ -58,6 +58,12 @@ def status
       def edit
         attrs = client.load_iap(app_id: application.apple_id, purchase_id: self.purchase_id)
         attrs[:application] = application
+
+        if attrs["addOnType"] == Spaceship::Tunes::IAPType::RECURRING
+          attrs["pricingIntervals"] = client.load_recurring_iap_pricing(app_id: application.apple_id,
```
untitaker: * We could do this lazily inside of `IAPDetail#pricing_intervals`
* could we avoid throwing away the rest of the response and store it in an attribute called like `raw_pricing_data`? (I think this is reasonably within scope of this PR as opposed to fully dealing with introductory pricing)

spaceship/lib/spaceship/tunes/iap.rb
```diff
@@ -75,6 +75,17 @@ def create!(type: "consumable",
...
+          client.update_recurring_iap_pricing!(app_id: self.application.apple_id,
```
luongm: btw, I found one little bug here, pricing_intervals passed in this method should be the full pricing intervals for all countries. This list is generated in `iap_detail.rb` before saving (here https://github.com/fastlane/fastlane/pull/11627/files#diff-0ad09501b52f13ed83e0709d8a5b1df3L180)

spaceship/spec/tunes/fixtures/iap_create_recurring.json
```diff
@@ -0,0 +1 @@
..."isRequired":false,"errorKeys":null},"bonusPeriodDurationType":{"value":null,"isEditable":true,"isRequired":false,"errorKeys":null},"versions":[{"id":null,"details":{"value":[{"value":{"description":{"value":"Description has at least 10 characters"},"name":{"value":"test name2"},"localeCode":"en-US"}},{"value":{"description":{"value":"German has at least 10 characters"},"name":{"value":"test name german2"},"localeCode":"de-DE"}}],"isEditable":true,"isRequired":false,"errorKeys":null},"contentHosting":null,"contentHostingData":null,"reviewNotes":{"value":"Some Review Notes here bla bla bla"},"reviewScreenshot":{"value":{"assetToken":null,"sortOrder":null,"type":null,"size":null,"width":null,"height":null,"checksum":null,"url":null,"thumbNailUrl":null,"originalFileName":null},"isEditable":true,"isRequired":false,"errorKeys":null},"merch":{"images":[{"id":null,"image":{"value":{"assetToken":null,"originalFileName":null,"width":null,"height":null,"checksum":null},"isEditable":true,"isRequired":false,"errorKeys":null},"status":null}],"showByDefault":true,"isActive":false},"status":null,"canBeSubmitted":false}],"missingRequiredPrivacyPolicyData":false,"missingRequiredFamilyDetail":false}
```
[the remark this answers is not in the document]
byohay: @luongm unfortunately no, stubbing of an HTTP request is done by matching the string in this file to the one that the test sends. If there was a match the stubbed response is returned. Otherwise the test fails because it is not allowed to send requests in tests.

spaceship/spec/tunes/fixtures/iap_update_recurring.json
```diff
@@ -0,0 +1 @@
...ull,"referenceName":{"value":"created by spaceship","isEditable":true,"isRequired":true,"errorKeys":null,"maxLength":64,"minLength":2},"productId":{"value":"x.a.a.b.b.c.d.x.y.z","isEditable":false,"isRequired":true,"errorKeys":null,"maxLength":100,"minLength":2},"clearedForSale":{"value":true,"isEditable":true,"isRequired":false,"errorKeys":null},"pricingDurationType":null,"pricingIntervals":[{"value":{"tierStem":4,"priceTierEffectiveDate":null,"priceTierEndDate":null,"country":"WW","grandfathered":null}}],"ungrandfatheredIntervals":null,"freeTrialDurationType":null,"bonusPeriodDurationType":null,"versions":[{"reviewNotes":{"value":null},"contentHosting":null,"details":{"value":[{"value":{"description":{"value":"German has at least 10 characters"},"name":{"value":"test name german1"},"localeCode":"de-DE"}},{"value":{"description":{"value":"German has at least 10 characters"},"name":{"value":"test name danish"},"localeCode":"da"}},{"value":{"description":{"value":"Description has at least 10 characters"},"name":{"value":"Edit It"},"localeCode":"en-US"}}]},"id":"95811535","reviewScreenshot":{"value":null}}],"missingRequiredPrivacyPolicyData":false,"missingRequiredFamilyDetail":false}
```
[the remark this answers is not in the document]
byohay: Same as above.

spaceship/lib/spaceship/tunes/iap.rb
byohay: You're right, I fixed it. For now I copy the code from `iap_detail.rb`. In the future I suggest to have one place that does this transformation, I think `tunes_client.rb` is a good place for this.   It is already done in `create_iap!` method.

spaceship/lib/spaceship/tunes/iap_list.rb
byohay: 1. I suggest against it, `IAPDetail#pricing_intervals` is a simple getter, and if I were to fetch it lazily it would have to make an HTTP request which will take longer. `product.edit` makes a request anyway so from that perspective the request should take only a while longer.
2. It's a good idea, but it makes things a bit more difficult:

- I have to maintain both this and `raw_data["pricingIntervals"]` when the user changes the pricing intervals, and whenever `raw_data["pricingIntervals"]` is accessed, I might have to use `raw_pricing_data["subscriptions"]` make it so that `raw_pricing_data["subscriptions"]` is used instead.
- `raw_pricing_data` will have to go into `IAPDetail`, and it has to be public even thought its kinda implementation detail. I would have added it as a private field and initialize it once in `initialize`, but I can't because the `Base` class takes over the initializer.

Then I thought that `raw_pricing_data` can be put inside `raw_data` itself. But since `update_iap!` uses it as a whole, it will send `raw_pricing_data` which will cause the request to fail. So I have to remove it from `raw_data` before `update_iap!` is called. What do you think of this solution?

spaceship/lib/spaceship/tunes/iap.rb
byohay: What do you mean? This method creates a new product with `product_id`. From what I saw, every time the creation failed an exception was raised. Though I agree that a retry counter should be here just in case.

byohay:
@untitaker @luongm thanks for the comments! After we finish resolving the `raw_pricing_data` issue I'll update the PR. 
@luongm I think that handling intro prices should be in another PR.

spaceship/lib/spaceship/tunes/iap.rb
untitaker: Consider this for example:

* Process A creates product
* Process B deletes same product
* Process A hits this line and hangs up

spaceship/lib/spaceship/tunes/iap_list.rb
untitaker: That's why I suggested moving the entire logic into IAPDetails, such that it would fetch pricing intervals and introductory offers on its own, and store *both* only in `raw_pricing_data` (not adding keys to `raw_data` at all)

I can make a separate PR for this if you want

>it has to be public even thought its kinda implementation detail

this is crucial, "we" (our company) need access to the raw data as long as introductory offers are not fully implemented.

spaceship/lib/spaceship/tunes/iap.rb
byohay: You will run into an exception if you try to create a product with the same `product_id`, even if the product was deleted. Anyway, I will add a retry counter.

untitaker: I didn't say that two products were created at all. All process A does it to try to create the product with the lines above this one, and while Process A is still in the same execution of this method (or suspended inbetween due to high load), process B deletes the product. Then process A would hit an infinite loop.

byohay: Right, in this case process A will hit an infinite loop.

spaceship/lib/spaceship/tunes/iap_list.rb
byohay: After some thought I agree with you, we should not add keys to `raw_data` because it is used in many places and sending a product update request with the new keys will fail.
So I added a `raw_pricing_data` field to `IAPDetail`. I initialize it from the pricing data in `raw_data` that I add beforehand, and then remove the superfluous data from `raw_data`.

byohay:
I pushed a fix with the review fixes.

janpio:
This looks pretty good to me. What do you think @untitaker @luongm?

untitaker (approved):
good stuff, thanks!

spaceship/lib/spaceship/tunes/iap_list.rb
```diff
@@ -58,6 +58,13 @@ def status
       def edit
         attrs = client.load_iap(app_id: application.apple_id, purchase_id: self.purchase_id)
         attrs[:application] = application
+
+        if attrs["addOnType"] == Spaceship::Tunes::IAPType::RECURRING
+          raw_pricing_data = client.load_recurring_iap_pricing(app_id: application.apple_id,
+                                                               purchase_id: self.purchase_id)
+          attrs["pricingData"] = raw_pricing_data
```
untitaker: This is not very fancy, I suppose changing the constructor of IAPDetails to take two distinct hashes is not doable?

luongm (approved):
LGTM

is there a way to escalate this PR to one of the people with write access?

untitaker:
No, you just wait :)

janpio (approved):
I take your word for it - and merge it. 
Thanks @byohay  🥇

spaceship/lib/spaceship/tunes/iap_list.rb
byohay: I didn't like it either, but in `Base::initialize` it says: ``Do not override `initialize` in your own models.``

PR #11628, opened 2018-01-19
Add missing action aliases

thii:
`fastlane/lib/fastlane/actions/puts.rb` lists `println` and `echo` as action aliases but these actions are missing.


taquitos (approved):
Thank you!
🎈:goat:

PR #11639, opened 2018-01-20
Fix plugin search only returns plugins which have search query as name prefix.

thii:
`fastlane search_plugins cache` currently returns no plugins. With this
fix, it will return `carthage_cache`, `rome`, `build_cache` and
`carthage_cache_ftps`.

### DIFF
--- a/fastlane/lib/fastlane/plugins/plugin_fetcher.rb
+++ b/fastlane/lib/fastlane/plugins/plugin_fetcher.rb
@@ -12,7 +12,7 @@ module Fastlane
       page = 1
       plugins = []
       loop do
-        url = "https://rubygems.org/api/v1/search.json?query=#{PluginManager.plugin_prefix}#{search_query}&page=#{page}"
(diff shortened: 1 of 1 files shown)


janpio:
Looks good. Can you explain the change a bit? How and why does it work?

thii:
Because Ruby Gems API doesn't seem to support regex in search query, instead of sending the search query `fastlane-plugin-#{search_query}` directly to Ruby Gems API, it searches for all available plugins and then filters out based on search query.

janpio:
https://rubygems.org/api/v1/search.json?query=fastlane-plugin-ionic

Seems the implementation for now was searching for exact matches in the plugin name, in your example `fastlane-plugin-cache`.

I think your changes are indeed better, but maybe there was a reason for the previous behavior?

thii:
I don't know exactly. Maybe because it's faster.

KrauseFx:
I guess, this will cause longer search times because more pages are required, but I think it's still ok 👍

KrauseFx (approved):
Beautiful 👍

PR #11640, opened 2018-01-20
[deliver] Download screenshots: add `use_live_version` option

andreacipriani:
### Checklist
   - I haven't, because if I'm not wrong there is no documentation about the `download_screenshots` options

### Motivation and Context

Imagine you have two versions on iTunes connect: `1` (live) and `2` (edit). When you created version `2`, the screenshots were not carried over from `1` (it did happen to me for the last two releases).
So you might want to download screenshots from `1`, but by default, if there is an edit version, `download_screenshots` will always use that version. References to code: [here](https://github.com/fastlane/fastlane/blob/83f9895f557293862a9f7546e0ff06cd4b3c1440/deliver/lib/deliver/download_screenshots.rb#L15) and [here](https://github.com/fastlane/fastlane/blob/1a79353a6d918eddca61274e97962a5f2770210e/spaceship/lib/spaceship/tunes/application.rb#L130).

The PR adds the possibility to specify a `--use_live_version` option to overcome this problem.

I tested that I didn't break the existing behavior when not using the new option.
I tested that when using the option: `bundle exec fastlane deliver download_screenshots --use_live_version` it uses the live version also when there is an edit version.

### Description
- Adding the new option
- When the option is set, use the already defined `live_version` [function](https://github.com/fastlane/fastlane/blob/1a79353a6d918eddca61274e97962a5f2770210e/spaceship/lib/spaceship/tunes/application.rb#L118)


andreacipriani:
I signed it!

janpio (approved):
What a simple but nice change. Looks good!

You did test it locally and it is working as expected? Both downloading latest and live version screenshots?

andreacipriani:
@janpio yep!

janpio:
🥇

Issue #10306, opened 2017-09-11
Fastlane using system version of OpenSSL instead of the one in fastlane env


Cyberbeni:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
`fastlane env` shows `OpenSSL Version | OpenSSL 1.0.2l  25 May 2017`
I manually installed OpenSSL 1.1.0f following [this guide](http://mac-dev-env.patrickbougie.com/openssl/) and added the exports in bash_profile, so `which openssl` gives me this:
```
openssl is /usr/local/openssl/bin/openssl
openssl is /usr/bin/openssl
```
And `openssl version -a` this:
```
OpenSSL 1.1.0f  25 May 2017
built on: reproducible build, date unspecified
platform: darwin64-x86_64-cc
compiler: cc -DDSO_DLFCN -DHAVE_DLFCN_H -DNDEBUG -DOPENSSL_THREADS -DOPENSSL_NO_STATIC_ENGINE -DOPENSSL_PIC -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DRC4_ASM -DMD5_ASM -DAES_ASM -DVPAES_ASM -DBSAES_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DPADLOCK_ASM -DPOLY1305_ASM -DOPENSSLDIR="\"/usr/local/mac-dev-env/openssl-1.1.0f/ssl\"" -DENGINESDIR="\"/usr/local/mac-dev-env/openssl-1.1.0f/lib/engines-1.1\"" 
OPENSSLDIR: "/usr/local/mac-dev-env/openssl-1.1.0f/ssl"
ENGINESDIR: "/usr/local/mac-dev-env/openssl-1.1.0f/lib/engines-1.1"
```
If I try to use `fastlane match` I get an error:
```
Couldn't decrypt the repo, please make sure you enter the right password!
[...]

Issue #9542, opened 2017-06-21
[match] OpenSSL 1.1.0 breaks backwards compatibility with old encrypted keys, etc


pepasflo:
### Summary:

It looks like OpenSSL 1.1.0 changed something about the default settings they use to encrypt / decrypt files.

One possible work-around for this issue would be to modify `match` to do the following:
- Try to decrypt the p12 file
- If that fails, try it again using the older openssl behavior (by appending `-md md5` to the openssl command)
- If that also fails, treat it like an actual failure and continue with the existing (failure) codepath.

### Details:

Here's a stackoverflow post which describes the issue: https://unix.stackexchange.com/a/344586

> The default hash used by openssl enc for password-based key derivation changed in 1.1.0 to SHA256 versus MD5 in lower versions.

On a box with older openssl installed, `match` will work correctly.  You can test this by manually decrypting one of match's p12 files using this command:

```
openssl aes-256-cbc -k password-goes-here -in ABCDE12345.p12 -out /dev/null -a -d
```

On a box with openssl 1.1.0+ installed, `match` will fail:

```
WARN [2017-06-19 16:49:50.76]: Enter the passphrase that should be used to encrypt/decrypt your certificates
WARN [2017-06-19 16:49:50.76]: This passphrase is specific per repository and will be stored in your local keychain
WARN [2017-06-19 16:49:50.76]: Make sure to remember the password, as you'll need it when you run match on a different machine
 [2017-06-19 16:49:50.76]: Passphrase for Git Repo: ********
 [2017-06-19 16:49:54.30]: Type passphrase again: ********
bad decrypt
140736683381696:error:06065064:digital envelope routines:EVP_DecryptFinal_ex:bad decrypt:crypto/evp/evp_enc.c:535:
ERROR [2017-06-19 16:49:56.67]: Couldn't decrypt the repo, please make sure you enter the right password!
```

However, if you run the above `openssl` command and append `-md md5` (which forces the old behavior), it will work, you can correctly decrypt a p12 file which was encrypted using the older version of `openssl`.

The new behavior of openssl is the equivalent of appending `-md sha256`.

### New Issue Checklist

- [❌ ] Updated fastlane to the latest version -- **I am currently using 2.37.0**
- [✅ ] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Environment

I am using fastlane 2.37.0.

<details><summary>🚫 fastlane environment 🚫</summary>

[...]

PR #11641, opened 2018-01-20
[match] using ruby openssl instead of terminal one

Cyberbeni:
### Checklist

### Motivation and Context
https://github.com/fastlane/fastlane/issues/10306
https://github.com/fastlane/fastlane/issues/9542

My current solution does exactly the same as the master on macOS 10.12.6

We should also change any other terminal based openssl usage to the ruby version to avoid these kind of conflicts.

We should also deprecate the current encryption of certificates to be more safer:
-randomly generate an IV instead of calculating it from the password and provide it in the certs repo unencrypted
-use higher iteration count for generating the key
https://crypto.stackexchange.com/questions/2280/why-is-the-iv-passed-in-the-clear-when-it-can-be-easily-encrypted
https://stackoverflow.com/questions/1949640/does-iv-work-like-salt
https://ruby.github.io/openssl/OpenSSL/Cipher.html

### Description
Changed terminal openssl to ruby openssl in fastlane match

### DIFF
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -99,7 +99,6 @@ module Match
     end
 
     def decrypt(path: nil, password: nil)
-      stored_data = ""
       stored_data = Base64.decode64(File.read(path))
       salt = stored_data[8..15]
       data_to_decrypt = stored_data[16..-1]
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -93,6 +93,8 @@ module Match
       encrypted_data = "Salted__" + salt + cipher.update(data_to_encrypt) + cipher.final
 
       File.write(path, Base64.encode64(encrypted_data))
+    rescue FastlaneCore::Interface::FastlaneError => user_error
+      raise
     rescue => error
       UI.error(error.to_s)
       UI.crash!("Error encrypting '#{path}'")
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -81,6 +81,12 @@ module Match
       end
     end
 
+    # We encrypt with MD5 because that was the most common default value in older fastlane versions which used the local OpenSSL installation
+    # A more secure key and IV generation is needed in the future
+    # IV should be randomly generated and provided unencrypted
+    # salt should be randomly generated and provided unencrypted (like in the current implementation)
+    # key should be generated with OpenSSL::KDF::pbkdf2_hmac with properly chosen parameters
+    # Short explanation about salt and IV: https://stackoverflow.com/a/1950674/6324550
     def encrypt(path: nil, password: nil)
       UI.user_error!("No password supplied") if password.to_s.strip.length == 0
 
@@ -100,6 +106,8 @@ module Match
       UI.crash!("Error encrypting '#{path}'")
     end
 
+    # The encryption parameters in this implementations reflect the old behaviour which depended on the users' local OpenSSL version
+    # 1.0.x OpenSSL and earlier versions use MD5, 1.1.0c and newer uses SHA256, we try both before giving an error
     def decrypt(path: nil, password: nil, hash_algorithm: "MD5")
       stored_data = Base64.decode64(File.read(path))
       salt = stored_data[8..15]
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -93,7 +93,7 @@ module Match
       encrypted_data = "Salted__" + salt + cipher.update(data_to_encrypt) + cipher.final
 
       File.write(path, Base64.encode64(encrypted_data))
-    rescue FastlaneCore::Interface::FastlaneError => user_error
+    rescue FastlaneCore::Interface::FastlaneError
       raise
     rescue => error
       UI.error(error.to_s)
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -100,21 +100,26 @@ module Match
       UI.crash!("Error encrypting '#{path}'")
     end
 
-    def decrypt(path: nil, password: nil)
+    def decrypt(path: nil, password: nil, hash_algorithm: "MD5")
       stored_data = Base64.decode64(File.read(path))
       salt = stored_data[8..15]
       data_to_decrypt = stored_data[16..-1]
 
       decipher = OpenSSL::Cipher.new('AES-256-CBC')
       decipher.decrypt
-      decipher.pkcs5_keyivgen(password, salt, 1, "MD5")
+      decipher.pkcs5_keyivgen(password, salt, 1, hash_algorithm)
 
       decrypted_data = decipher.update(data_to_decrypt) + decipher.final
 
       File.write(path, decrypted_data)
     rescue => error
-      UI.error(error.to_s)
-      UI.crash!("Error decrypting '#{path}'")
+      fallback_hash_algorithm = "SHA256"
+      if hash_algorithm != fallback_hash_algorithm
+        decrypt(path, password, fallback_hash_algorithm)
+      else
+        UI.error(error.to_s)
+        UI.crash!("Error decrypting '#{path}'")
+      end
     end
   end
 end
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -3,9 +3,11 @@ require_relative 'change_password'
 
 module Match
   class Encrypt
+    require 'base64'
+    require 'openssl'
+    require 'securerandom'
     require 'security'
     require 'shellwords'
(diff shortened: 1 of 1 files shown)


Cyberbeni:
I signed it!

KrauseFx:
Awesome, thanks, same question as https://github.com/fastlane/fastlane/pull/11652#issuecomment-359449856

Cyberbeni:
Yes, there are also tests that fail if it wouldn't work correctly.

KrauseFx:
Tests are great, but they don't cover everything 😉

Cyberbeni:
Yeah, but I also tested this one thoroughly because at first I had some problems figuring out what to use to get the exact same result as the openssl that's preinstalled on macOS 10.12.6

match/lib/match/encrypt.rb
```diff
@@ -81,43 +81,45 @@ def iterate(source_path)
...
+      decipher.decrypt
+      decipher.pkcs5_keyivgen(password, salt, 1, "MD5")
 
-      FileUtils.mv(tmpfile, path)
+      decrypted_data = decipher.update(data_to_decrypt) + decipher.final
+
+      File.open(path, "w") do |file|
+        file.write(decrypted_data)
+      end
+    rescue => error
```
lacostej: * any reason we catch errors only in decrypt and not in encrypt?
* why logging the error only in verbose?

match/lib/match/encrypt.rb
```diff
@@ -3,9 +3,11 @@
 
 module Match
   class Encrypt
+    require 'base64'
+    require 'openssl'
```
lacostej: if we start depending on a new gem, we should modify the Gemfile. If we want to enforce a way to use it, maybe we should provide a wrapper.

match/lib/match/encrypt.rb
```diff
@@ -81,43 +81,45 @@ def iterate(source_path)
...
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show an error message if something goes wrong
-
-      _out, err, st = Open3.capture3(command.join(' '))
-      success = st.success?
-
-      # Ubuntu `openssl` does not fail on failure
-      # but at least outputs an error message
-      unless err.to_s.empty?
-        success = false
+      data_to_encrypt = ""
```
lacostej: shorter with `data_to_encrypt = File.read(path)`

match/lib/match/encrypt.rb
```diff
@@ -81,43 +81,45 @@ def iterate(source_path)
...
-      # On non-Mac systems (more specific Ubuntu Linux) it might take some time for the file to actually be there (see #11182).
-      # To try to circumvent this flakyness (in tests), we wait a bit until the file appears (max 2s) (usually only 0.1 is actually waited)
-      unless FastlaneCore::Helper.is_mac?
-        count = 0
-        # sleep until file exists or 20*0.1s (=2s) passed
-        until File.exist?(tmpfile) || count == 20
-          sleep(0.1)
-          count += 1
-        end
+      File.open(path, "w") do |file|
```
lacostej: shorter with `File.write..`.

match/lib/match/encrypt.rb
Cyberbeni: Which new gem? OpenSSL? It's already used in most places where we want to work with SSL related stuff and we want to use the latest version possible, so we have the latest security updates.
We don't want to enforce the way it's used in this current pull request because that's deprecated, it's just the way how the command line openssl command worked.
After this PR is merged, we can work out a better way of encrypting stuff (generate a random IV instead of generating it from the password, like the key, and provide it unencrypted. Reasons why this is more secure is explained in the links in the PR) and do all new encrypting that way, while still supporting decrypting with the old algorithm if the users' match repo haven't been encrypted in the new way yet.
The main goal for this PR is to solve the error for people who installed OpenSSL 1.1.x on their machine (and work with people who didn't) because that changes the default hashing algorithm and it wasn't implicitly told to the old openssl command to use MD5.

Cyberbeni:
Okay, I really should try out RubyMine instead of Atom while I still have my all product license going. Feels like some of my newbie mistakes could've made an instant warning instead of having to run the tests.

janpio:
@Cyberbeni Could you please create an account at appveyor.com, add your fork of `fastlane` as a project and invite use "janpio" to it? I am in the progres of adding Windows CI to fastlane, but couldn't find a way to run the tests on your fork+branch from my side :/

(If you don't want to do this, we can also wait until this is merged and see if it breaks stuff on Windows CI for fastlane and **then** fix it here with another PR)

Cyberbeni:
@janpio It asks for e-mail when I want to add collaborator for the team.

janpio:
piotrowski+github@gmail.com should work.

janpio:
9 failures, but unrelated to your changes. This is good to merge from a Windows viewpoint then.

```
Finished in 2 minutes 34.3 seconds (files took 3.7 seconds to load)
5213 examples, 9 failures, 180 pending
Failed examples:
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:10 # Fastlane Automatic Code Signing enable_automatic_code_signing
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:33 # Fastlane Automatic Code Signing disable_automatic_code_signing for specific target
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:55 # Fastlane Automatic Code Signing disable_automatic_code_signing
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:75 # Fastlane Automatic Code Signing sets team id
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:102 # Fastlane Automatic Code Signing sets code sign identity
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:134 # Fastlane Automatic Code Signing sets profile name
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:166 # Fastlane Automatic Code Signing sets profile uuid
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:198 # Fastlane Automatic Code Signing sets bundle identifier
rspec ./fastlane/spec/actions_specs/automatic_code_signing_spec.rb:230 # Fastlane Automatic Code Signing targets not found notice
```

janpio:
I really like how you are removing two really ugly hacks I had to use here by the way :)

match/lib/match/encrypt.rb
janpio: @lacostej Any work necessary here or is this good?

janpio:
Note to myself:
Also merge https://github.com/fastlane/fastlane/pull/11720 and https://github.com/fastlane/fastlane/pull/11652 when this gets merged.

match/lib/match/encrypt.rb
lacostej: @Cyberbeni you're right. It's not new. It's just not yet depended upon in the Gemfile. Weird.

Question: are we 100% sure no one managed to encrypt data with openssl 1.1 and its new hashing algorithm and thus wouldn't be able to decrypt it with the previous hash? If so, shouldn't we use a decrypt mechanism that tries with both hash methods?

Cyberbeni: @lacostej it doesn't hurt to fallback to SHA256 if decrypting with MD5 fails.
could only edit in browser, hopefully no errors.

lacostej: @Cyberbeni it looks OK. I would probably add a few notes in the code

* that we use MD5 as primary for encryption even if known to not be the latest/most secure and that this whole encrypt/decrypt algorithm is meant to be improved
* that we fallback to decryption using SHA256 in case people encrypted data using openssl 1.1.0c or later with fastlane <= 2.80.0

or something like that.

Last question: given that the `pkcs5_keyivgen` method is deprecated (https://apidock.com/ruby/v1_9_3_125/OpenSSL/Cipher/pkcs5_keyivgen), was there a non deprecated way to achieve the same results?

Cyberbeni: @lacostej They recommend using OpenSSL::KDF::pbkdf2_hmac instead which has almost the same parameters and only 1 output. pkcs5_keyivgen generates both the key(16 length) and IV(32 length), I tried with same inputs and 16 length and it resulted in a different key, so my guess is that there is no other, non-deprecated method that keeps us backwards compatible with older fastlane versions.

I'll add comments then.

lacostej (approved):
Thanks a lot @Cyberbeni for taking time to explain all the inner workings and trying various approaches.

Cyberbeni:
@lacostej no problem. A coworker and me spent a good hour or two trying to find out why match didn't work on my MacBook, just wanted to help others with OpenSSL 1.1 avoid that.

janpio:
🥇

Issue #11434, opened 2018-01-05
765: Unexpected token at 'Bad Request' when calling to provisioningProfile.delete!


cfis:
Fastlane 2.72.0
This is likely related to #11162

Basic code:

profile = ::Spaceship::Portal.provisioning_profile.find_by_bundle_id(bundle_id: app.app_id).first
if profile.status == 'Invalid'
  profile.delete!
end

The root cause is Spaceship::PortalClient#provisioning_profiles_via_xcode_api

Line 590 in portal_client_id.rb:

      csrf_cache[Spaceship::ProvisioningProfile] = self.csrf_tokens

The problem is that the api response does not include and csrf_tokens, so self.csrf_tokens is an empty hash.  But since empty is not checked, then:

    def delete_provisioning_profile!(profile_id, mac: false)
      ensure_csrf(Spaceship::ProvisioningProfile) do
        fetch_csrf_token_for_provisioning
      end

The call the ensure_csrf checks for nil but not an empty hash:

    def ensure_csrf(klass)
      if csrf_cache[klass]
        self.csrf_tokens = csrf_cache[klass]
        return
      end

Thus the request is made with no tokens and you get a bad request.

This error just started happening for me recently.  I suspect the difference is the Apple account id I am using is now associated with multiple teams.  Previously it was not, and Fastlane was working (version 268).   Alternatively, maybe the  provisioning_profiles_via_xcode_api api no longer returns csfr tokens?

My quick fix is change:

      csrf_cache[Spaceship::ProvisioningProfile] = self.csrf_tokens

To:

      csrf_cache[Spaceship::ProvisioningProfile] = self.csrf_tokens unless self.csrf_tokens.empty?

fullsailor:
I encountered this problem as well. It does seem like the Xcode API used in `PortalClient#provisioning_profiles_via_xcode_api` does not return csrf tokens, so the empty hash incorrectly bypasses the check in `ensure_csrf`.

Issue #11162, opened 2017-12-09
765: unexpected token at 'Bad Request' when calling to provisioningProfile.update!


pbeast:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
The following code crashes with

> /Users/pbeast/.gems/gems/json-2.1.0/lib/json/common.rb:156:in 'parse': **765: unexpected token at 'Bad Request'** (Faraday::ParsingError)
	from /Users/pbeast/.gems/gems/json-2.1.0/lib/json/common.rb:156:in 'parse'
	from /Users/pbeast/.gems/gems/faraday_middleware-0.12.2/lib/faraday_middleware/response/parse_json.rb:11:in 'block in <class:ParseJson>'
	from /Users/pbeast/.gems/gems/faraday_middleware-0.12.2/lib/faraday_middleware/response_middleware.rb:51:in 'parse'
	from /Users/pbeast/.gems/gems/faraday_middleware-0.12.2/lib/faraday_middleware/response_middleware.rb:40:in 'process_response'
	from /Users/pbeast/.gems/gems/faraday_middleware-0.12.2/lib/faraday_middleware/response_middleware.rb:33:in 'block in call'
	from /Users/pbeast/.gems/gems/faraday-0.13.1/lib/faraday/response.rb:61:in 'on_complete'
	from /Users/pbeast/.gems/gems/faraday_middleware-0.12.2/lib/faraday_middleware/response_middleware.rb:31:in 'call'
	from /Users/pbeast/.gems/gems/faraday-0.13.1/lib/faraday/rack_builder.rb:141:in 'build_response'
	from /Users/pbeast/.gems/gems/faraday-0.13.1/lib/faraday/connection.rb:387:in vrun_request'
	from /Users/pbeast/.gems/gems/faraday-0.13.1/lib/faraday/connection.rb:174:in 'post'
	from /Users/pbeast/.gems/gems/fastlane-2.68.2/spaceship/lib/spaceship/client.rb:724:in 'block in send_request'
	from /Users/pbeast/.gems/gems/fastlane-2.68.2/spaceship/lib/spaceship/client.rb:559:in 'with_retry'
[...]

PR #11650, opened 2018-01-22
Enforcing CSRF token fetching when working with provisioning profile

pbeast:
### Checklist

### Motivation and Context
When working with provisioning profiles (`portal_client.rb`) it looks like CSRF tokens may become obsolete sporadically what leads to error responses.

### Description
The solution is to enforce CSRF token update before making following requests:

- /createProvisioningProfile.action
- /downloadProfileContent
- /deleteProvisioningProfile.action
- /regenProvisioningProfile.action


KrauseFx:
This will cause every spaceship request to send an additional request. Can we find a way to still use that caching, while also fix the situation you described?

KrauseFx:
What's the best way to reproduce

pbeast:
Hi,

The issue #11162 contains reproduction scenario. The reproduction doesn’t depends on time you are connected to the systems, but on number of calls to the same endpoint.

pbeast:
Hi, just noted another open issue - #11434

lacostej:
Just curious, 
* have we looked at the response headers of the response previous to the one that cause errors to see if there are hints that something is going to go wrong?
* alternatively, could we do a retry with fetching the CSRF upon 421 error ?

julianoterres (approved):

janpio:
Hey @pbeast, is this still something that should be merged or did the problem go away in the meantime?

pbeast:
Hi, the issue was resolved and can be closed. Thanks

Regards,
Pavel Yankelevich

________________________________

PR #11675, opened 2018-01-23
Support checking remote git tags existence

chuganzy:
### Checklist

### Motivation and Context

This PR enables to check remote tag existence with `git_tag_exists` action.

### Description

Also switch to use `error_callback` to check existence since `git` commands outputs std logs in some cases like warnings and I thought it's more robust to check using status code.

### DIFF
--- a/fastlane/lib/fastlane/actions/git_tag_exists.rb
+++ b/fastlane/lib/fastlane/actions/git_tag_exists.rb
@@ -34,12 +34,14 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :tag,
                                        description: "The tag name that should be checked"),
           FastlaneCore::ConfigItem.new(key: :remote,
-                                       description: "Defaults to `false`",
+                                       description: "Whether to check remote. Defaults to `false`",
                                        type: Boolean,
-                                       default_value: false),
+                                       default_value: false,
+                                       optional: true),
           FastlaneCore::ConfigItem.new(key: :remote_name,
                                        description: "The remote to check. Defaults to `origin`",
-                                       default_value: 'origin')
+                                       default_value: 'origin',
+                                       optional: true)
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/git_tag_exists.rb
+++ b/fastlane/lib/fastlane/actions/git_tag_exists.rb
@@ -2,8 +2,21 @@ module Fastlane
   module Actions
     class GitTagExistsAction < Action
       def self.run(params)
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/git_tag_exists.rb
```diff
@@ -2,8 +2,19 @@ module Fastlane
   module Actions
     class GitTagExistsAction < Action
       def self.run(params)
-        result = Actions.sh("git rev-parse -q --verify refs/tags/#{params[:tag].shellescape} || true", log: FastlaneCore::Globals.verbose?).chomp
-        !result.empty?
+        tag_ref = "refs/tags/#{params[:tag].shellescape}"
+        if params[:remote]
```
janpio: wouldn't it be simpler to just use the existance of `remote_name` to decide if a remote should be checked?

chuganzy: Good point!

chuganzy: Hmm, but in that case we cannot provide default `remote_name` value..
It's fine to me but most of other `git remote` related actions have default value for `remote`.

janpio: Also true, right.
What other actions are there? Quick google only found me https://docs.fastlane.tools/actions/push_git_tags/

chuganzy: There is push_to_git_remote I found.
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/push_to_git_remote.rb

janpio: Hmm... problem in comparing those is that both `push_git_tags` and `push_to_git` by default interact with a remote, so having a `remote` param with default value `origin` makes total sense. But this one works locally by default, and only with this change can does remote things, right?

janpio: (Which of course doesn't help us solve the problem at all :p)

chuganzy: Correct, and it totally makes sense. Had a great discussion here, thanks! 
Will fix it only to use `remote_name`!

janpio: That wasn't really my conclusion though. I think you are good with how it is right now if you want to.

getaaron (approved):

nafu:
@chuganzy Thank you for your contribution ✨

PR #11678, opened 2018-01-23
Read files as UTF8 or binary explicitly

janpio:
(as this makes a difference on Windows)

### DIFF
--- a/spaceship/spec/portal/portal_stubbing.rb
+++ b/spaceship/spec/portal/portal_stubbing.rb
@@ -1,7 +1,7 @@
 class PortalStubbing
   class << self
     def adp_read_fixture_file(filename)
-      File.binread(File.join('spaceship', 'spec', 'portal', 'fixtures', filename))
+      File.read(File.join('spaceship', 'spec', 'portal', 'fixtures', filename))
     end
 
     def adp_read_binary_fixture_file(filename)
--- a/spaceship/spec/portal/portal_stubbing.rb
+++ b/spaceship/spec/portal/portal_stubbing.rb
@@ -1,12 +1,11 @@
 class PortalStubbing
   class << self
     def adp_read_fixture_file(filename)
-      if filename == "aps_development.cer"
-        file = File.binread(File.join('spaceship', 'spec', 'portal', 'fixtures', filename))
-      else
-        file = File.read(File.join('spaceship', 'spec', 'portal', 'fixtures', filename))
-      end
-      file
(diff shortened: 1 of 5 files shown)


spaceship/spec/portal/portal_stubbing.rb
```diff
@@ -1,7 +1,12 @@
 class PortalStubbing
   class << self
     def adp_read_fixture_file(filename)
-      File.read(File.join('spaceship', 'spec', 'portal', 'fixtures', filename))
```
lacostej: I am not that fond of the hardcoded file name in the `adp_read_fixture_file` implementation. Alternatives are maybe to have
* a different function for bin reading ( `adp_binread_fixture_file`)
* an explicit parameter (`adp_read_fixture_file(file, binary: false)`)

lacostej (commented):
There are many more places where we call File.read without specifying the encoding. Will they create an issue?

janpio:
> There are many more places where we call File.read without specifying the encoding. Will they create an issue?

That is possible and also very probable - it will pop up when we actually start using fastlane on Windows and testing the individual tools and actions.

My hope for the long term is there is a better solution than to provide `encoding: "utf-8"` for each `File.read` that needs it (like configuring a system wide default or similar), but for the short term this was the easiest/simplest way to get the tests to pass. So this is definitely an issue that will be tackled later on.

(If it will not appear during normal fastlane usage, then I think it is good I didn't invest more time into finding that "bigger" solution. But I honestly don't expect that to happen.)

spaceship/spec/portal/portal_stubbing.rb
janpio: Yes, that is one ugly piece of code.

I like the idea of two different functions (param is not obvious enough imho) and will implement this.

janpio: This is now fixed.

spaceship/spec/portal/portal_stubbing.rb
```diff
@@ -1,7 +1,11 @@
...
+      File.binread(File.join('spaceship', 'spec', 'portal', 'fixtures', filename))
```
lacostej: both are binread now?

lacostej (changes requested):

spaceship/spec/portal/portal_stubbing.rb
janpio: Ahem, while this of course works this was not the goal. 
That's why we have review - sorry about that.
Fixed.

lacostej (approved):
great

PR #11679, opened 2018-01-23
Don't check for executable files on Windows

janpio:
No executable files on Windows, so we just skip that check there and assume success.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -325,7 +325,7 @@ module FastlaneCore
 
     # checks if a given path is an executable file
     def self.executable?(cmd_path)
-      cmd_path && (File.executable?(cmd_path) || (Helper.is_windows? && File.exist?(cmd_path))) && !File.directory?(cmd_path)
+      cmd_path && (File.executable?(cmd_path) || (Helper.windows? && File.exist?(cmd_path))) && !File.directory?(cmd_path)
     end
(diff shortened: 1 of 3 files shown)


lacostej (commented):
Any reason we implemented this differently in the 2 calls?

Maybe we want to move the check under Helper

`Helper.executable?(path)`

janpio:
Both valid points, will implement it this way.

janpio:
Implemented, much cleaner. Thanks for the nudge.

lacostej (approved):
Great!

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -322,5 +322,10 @@ def self.hide_loading_indicator
         @require_fastlane_spinner.success
       end
     end
+
+    # checks if a given path is an executable file
+    def self.executable?(cmd_path)
+      cmd_path && (File.executable?(cmd_path) || (Helper.is_windows? && File.exist?(cmd_path))) && !File.directory?(cmd_path)
```
lacostej: Last nitpick: I think the call should be `Helper.windows?`

lacostej: `is_windows?` is an alias for backwards compatibility

janpio: Oh good point. 
I have to reorder that file, getting to confusing.

PR #11680, opened 2018-01-23
Resolve more (strange) Windows test failures

janpio:
These commits are fixing more surprising and some even pretty strange things that made the tests on Windows fail:

- Tests were checking for the presence of `-x` in calculated commands that also include temp paths. These temp paths now can include `-x` in strings, so the matchers were suffixed with a space ` ` that to `-x ` which is actually the param that should (not) be present on these commands so the test is now actually better.
- "Verifying the --help command" didn't work on Windows as setting an env variable via `PAGER=cat cmd` doesn't work and results in an `No such file or directory - PAGER=cat bin/fastlane --help` error. So I now set the command depending on the platform and use one for Windows that works there and does the job.
- `which` was checking for executable files on Windows only by using the preset `ENV['PATHEXT']` list, but especially for the tests it also has to look at files with no file extension - so this was manually added to the list
- `echo "HO"` actually echos `"HO"` on Windows instead of only `HO` on Linux and macOS - so the command was changed to `echo HO` which matches the expectation of the test - which is only checking if commands can run (and not the implementation details of `echo`)
- A backtick command needs `"` in Windows instead of `'` which is also valid on macOS/Linux - so this was replaced
- In `encrypt.rb` the implementation of some error output if `verbose` is true was actually broken after previous changes (when `system` was replaced by `Open3.capture3`) - so I repaired that and made it cross platform at the same time.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -47,12 +47,11 @@ lane(:execute_tests) do
   version = local_version # has to be outside of the `Dir.chdir`
 
   # Verifying the --help command
-  unless Helper.windows?
-    Dir.chdir("..") do
-      content = sh("PAGER=cat bin/fastlane --help")
-      ["--version", "https://fastlane.tools", "fastlane"].each do |current|
-        UI.user_error!("--help missing information: '#{current}'") unless content.include?(current)
-      end
+  Dir.chdir("..") do
+    cmd = (Helper.windows?) ? "cd bin && fastlane --help" : "PAGER=cat bin/fastlane --help"
+    content = sh(cmd)
+    ["--version", "https://fastlane.tools", "fastlane"].each do |current|
+      UI.user_error!("--help missing information: '#{current}'") unless content.include?(current)
     end
   end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -47,11 +47,12 @@ lane(:execute_tests) do
   version = local_version # has to be outside of the `Dir.chdir`
 
   # Verifying the --help command
-  #
-  Dir.chdir("..") do
-    content = sh("PAGER=cat bin/fastlane --help")
-    ["--version", "https://fastlane.tools", "fastlane"].each do |current|
-      UI.user_error!("--help missing information: '#{current}'") unless content.include?(current)
+  unless Helper.windows?
+    Dir.chdir("..") do
+      content = sh("PAGER=cat bin/fastlane --help")
+      ["--version", "https://fastlane.tools", "fastlane"].each do |current|
+        UI.user_error!("--help missing information: '#{current}'") unless content.include?(current)
+      end
     end
   end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -48,7 +48,7 @@ lane(:execute_tests) do
 
   # Verifying the --help command
   Dir.chdir("..") do
-    cmd = (Helper.windows?) ? "cd bin && fastlane --help" : "PAGER=cat bin/fastlane --help"
+    cmd = Helper.windows? ? "cd bin && fastlane --help" : "PAGER=cat bin/fastlane --help"
     content = sh(cmd)
     ["--version", "https://fastlane.tools", "fastlane"].each do |current|
       UI.user_error!("--help missing information: '#{current}'") unless content.include?(current)
--- a/fastlane_core/lib/fastlane_core/command_executor.rb
+++ b/fastlane_core/lib/fastlane_core/command_executor.rb
@@ -15,8 +15,8 @@ module FastlaneCore
       def which(cmd)
         # PATHEXT contains the list of file extensions that Windows considers executable, semicolon separated.
         # e.g. ".COM;.EXE;.BAT;.CMD"
-        exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
-        exts << ''
+        exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : [] 
+        exts << '' # Always have an empty string (= no file extension)
 
         ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
           exts.each do |ext|
--- a/fastlane_core/lib/fastlane_core/command_executor.rb
+++ b/fastlane_core/lib/fastlane_core/command_executor.rb
@@ -15,7 +15,7 @@ module FastlaneCore
       def which(cmd)
         # PATHEXT contains the list of file extensions that Windows considers executable, semicolon separated.
         # e.g. ".COM;.EXE;.BAT;.CMD"
-        exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : [] 
+        exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : []
         exts << '' # Always have an empty string (= no file extension)
 
         ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
--- a/fastlane_core/lib/fastlane_core/command_executor.rb
+++ b/fastlane_core/lib/fastlane_core/command_executor.rb
@@ -16,6 +16,7 @@ module FastlaneCore
         # PATHEXT contains the list of file extensions that Windows considers executable, semicolon separated.
         # e.g. ".COM;.EXE;.BAT;.CMD"
         exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+        exts << ''
 
         ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
           exts.each do |ext|
--- a/fastlane_core/spec/project_spec.rb
+++ b/fastlane_core/spec/project_spec.rb
@@ -419,7 +419,7 @@ describe FastlaneCore do
       end
 
       it "runs simple commands" do
-        cmd = 'echo "HO"'
+        cmd = 'echo HO'
         expect(FastlaneCore::Project.run_command(cmd)).to eq("HO\n")
       end
 
--- a/fastlane_core/spec/project_spec.rb
+++ b/fastlane_core/spec/project_spec.rb
@@ -419,7 +419,7 @@ describe FastlaneCore do
       end
 
       it "runs simple commands" do
-        cmd = 'echo HO'
+        cmd = 'echo HO' # note: this command is deliberately not using `"` around `HO` as `echo` would echo those back on Windows
         expect(FastlaneCore::Project.run_command(cmd)).to eq("HO\n")
       end
 
--- a/frameit/lib/frameit/strings_parser.rb
+++ b/frameit/lib/frameit/strings_parser.rb
@@ -10,7 +10,7 @@ module Frameit
       result = {}
 
       # A .strings file is UTF-16 encoded. We only want to deal with UTF-8
-      content = `iconv -f UTF-16 -t UTF-8 '#{path}' 2>&1`
+      content = `iconv -f UTF-16 -t UTF-8 "#{path}" 2>&1`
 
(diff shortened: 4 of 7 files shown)


fastlane_core/lib/fastlane_core/command_executor.rb
```diff
@@ -15,11 +15,12 @@ def which(cmd)
         # PATHEXT contains the list of file extensions that Windows considers executable, semicolon separated.
         # e.g. ".COM;.EXE;.BAT;.CMD"
         exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+        exts << ''
```
lacostej: what does this do?

fastlane_core/lib/fastlane_core/command_executor.rb
```diff
@@ -15,11 +15,12 @@ def which(cmd)
...
 
         ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
           exts.each do |ext|
             cmd_path = File.join(path, "#{cmd}#{ext}")
-            return cmd_path if File.executable?(cmd_path) && !File.directory?(cmd_path)
+            return cmd_path if (File.executable?(cmd_path) || (Helper.is_windows? && File.exist?(cmd_path))) && !File.directory?(cmd_path)
```
lacostej: THis is part of #11679

fastlane_core/spec/project_spec.rb
```diff
@@ -419,7 +419,7 @@ def count_processes(text)
       end
 
       it "runs simple commands" do
-        cmd = 'echo "HO"'
+        cmd = 'echo HO'
```
lacostej: Part of this integration test was probably to check that running the program with those quotes worked as the platform is was supposed to be supported on. Right?

frameit/lib/frameit/strings_parser.rb
```diff
@@ -10,7 +10,7 @@ def self.parse(path)
       result = {}
 
       # A .strings file is UTF-16 encoded. We only want to deal with UTF-8
-      content = `iconv -f UTF-16 -t UTF-8 '#{path}' 2>&1`
+      content = `iconv -f UTF-16 -t UTF-8 "#{path}" 2>&1`
```
lacostej: Wouldn't it be better to use a cross platform way of escaping things?

This is how we solved it in u3d: https://github.com/DragonBox/u3d/blob/master/lib/u3d_core/core_ext/string.rb#L27-L33

match/lib/match/encrypt.rb
```diff
@@ -93,7 +93,9 @@ def crypt(path: nil, password: nil, encrypt: true)
       command << "-out #{tmpfile.shellescape}"
       command << "-a"
       command << "-d" unless encrypt
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show an error message if something goes wrong
+      unless Helper.is_windows?
+        command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show an error message if something goes wrong
```
lacostej: I believe you want `"2> nul"` in Windows to achieve the same thing

screengrab/lib/screengrab/android_environment.rb
```diff
@@ -84,7 +85,7 @@ def find_aapt(build_tools_path)
     end
 
     def executable_command?(cmd_path)
-      cmd_path && File.executable?(cmd_path) && !File.directory?(cmd_path)
+      cmd_path && (File.executable?(cmd_path) || Helper.is_windows?) && !File.directory?(cmd_path)
```
lacostej: again part of #11679

fastlane_core/lib/fastlane_core/command_executor.rb
janpio: It adds "no file extension" to the list. This was needed for a specific test... that I just now can't remember or pinpoint. Will try to find out again.

fastlane_core/lib/fastlane_core/command_executor.rb
janpio: Ouch, right. Will remove and repush.

fastlane_core/spec/project_spec.rb
janpio: The test name or expect doesn't indicate that the quotes were relevant. I don't think this was meant to test the `echo` implementation (which is what is different on Windows and macOS), but that `run_command` indeed runs a command in the expected way, right?

Alternatively to my change, we could also change the `.to eq()` to `should include()` to achieve the same thing. Would that be better?

frameit/lib/frameit/strings_parser.rb
janpio: Is this an escaping problem?

I thought it was just an `iconv` implementation detail that required the path to be wrapped in `"` on Windows instead of `"` _or_ `'` on Linux or macOS. Did I get that wrong?

match/lib/match/encrypt.rb
janpio: Thanks, will check that out.

janpio: Googling confirmed, so I implemented this. 
(Hope I got the `unless`es right...)

fastlane_core/lib/fastlane_core/command_executor.rb
janpio: Looking at the code again I think I just wrote that in a suoptimal way.
`exts` should always contain empty string as a value, but get the result of the `.split` on windows additionally. 
How would I write this in way that is more obvious and not confusing?

frameit/lib/frameit/strings_parser.rb
lacostej: Maybe. Not sure!

match/lib/match/encrypt.rb
```diff
@@ -93,7 +93,14 @@ def crypt(path: nil, password: nil, encrypt: true)
       command << "-out #{tmpfile.shellescape}"
       command << "-a"
       command << "-d" unless encrypt
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show an error message if something goes wrong
+      # to show an error message if something goes wrong
+      unless FastlaneCore::Globals.verbose?
+        if !Helper.is_windows?
```
lacostej: nitpick:

```
if windows
  ...
else
  ...
```

frameit/lib/frameit/strings_parser.rb
lacostej: I just think that it is the way Windows requires its parameters to be passed. Single quotes have no meaning there. Not specific to `iconv` in my mind.

match/lib/match/encrypt.rb
janpio: I did not do that as the code always is "by default assume macOS / Linux" - so I thought it would be better to have that code come first. Disagree?

frameit/lib/frameit/strings_parser.rb
janpio: Ok. What would be the best solution here then?

match/lib/match/encrypt.rb
```diff
@@ -93,7 +93,14 @@ def crypt(path: nil, password: nil, encrypt: true)
       command << "-out #{tmpfile.shellescape}"
       command << "-a"
       command << "-d" unless encrypt
-      command << "&> /dev/null" unless FastlaneCore::Globals.verbose? # to show an error message if something goes wrong
+      # to show an error message if something goes wrong
+      unless FastlaneCore::Globals.verbose?
+        if !Helper.windows?
+          command << "&> /dev/null"
+        else
+          command << "2> nul"
```
janpio: Ouch, that is not working on Windows @lacostej :

janpio: Tried all the obvious results from Google searches and Stackoverflow results - always "unknown option" in one form or another.

janpio:
As all other PRs are merged I rebased this one as well onto current `master` and reran the tests on Windows: 

> 5208 examples, 0 failures, 180 pending

Meaning: As soon as this PR is merged our Windows build will pass by default 🥇

fastlane_core/spec/project_spec.rb
lacostej: echo behaves differently on Windows and Unix. Let's keep that test testing running programs only, and if needed we will add per platform tests that check quoting. Maybe add a comment that the command is written in a way that it works the same on all platforms.

frameit/lib/frameit/strings_parser.rb
lacostej: I think a cross platform way of quoting things would be ideal on the long run for running generic commands. The fact that double quote works on both is probably acceptable here given that the command is explicit and not built at runtime. Maybe comment that double quotes required to work on LMW.

match/lib/match/encrypt.rb
```diff
@@ -104,6 +103,12 @@ def crypt(path: nil, password: nil, encrypt: true)
         success = false
       end
 
+      # to show an error message if something goes wrong
```
lacostej: you meant this inside the previous `unless`?

fastlane/Fastfile
```diff
@@ -47,11 +47,12 @@ lane(:execute_tests) do
   version = local_version # has to be outside of the `Dir.chdir`
 
   # Verifying the --help command
-  #
-  Dir.chdir("..") do
-    content = sh("PAGER=cat bin/fastlane --help")
-    ["--version", "https://fastlane.tools", "fastlane"].each do |current|
-      UI.user_error!("--help missing information: '#{current}'") unless content.include?(current)
+  unless Helper.windows?
+    Dir.chdir("..") do
```
lacostej: I believe PAGER is set for commander.

Here: https://github.com/commander-rb/commander/blob/927d6418fbb027fc8e0dcfd6db48deea73f15168/lib/commander/user_interaction.rb#L274-L300

Have you tried having a different command on windows, e.g. just `bin/fastlane --help`?

Given that Windows doesn't have tty, paging should not be needed.

lacostej (commented):
Some more comments!

fastlane_core/lib/fastlane_core/command_executor.rb
janpio: 1 it is - that is 100% clear and readable. Thanks!

fastlane_core/spec/project_spec.rb
janpio: added the comment

frameit/lib/frameit/strings_parser.rb
janpio: added a comment

match/lib/match/encrypt.rb
janpio: no, that is the old comment of the `dev/null` stuff that is now the code below this line - it makes sure if the `openssl` command fails with an error, that this error message is output to the command line of the user if `verbose` is set. Should I rewrite this somehow?

fastlane/Fastfile
janpio: I could make it do something with `cd bin && fastlane --help`, `bin/fastlane` it doesn't like. Let's try this.

janpio: Yes, that works! Thanks for the idea. Is the code I used to differentiate the commands ok?

match/lib/match/encrypt.rb
lacostej: but you are printing this when on verbose and regardless if there was an error.

janpio: Oh. My brain...

janpio: So: Yes, this was meant to go into the `unless`. Moved it there and also moved the comments around to match.

lacostej (approved):
Fine for me!

taquitos:
Awesome work @janpio, and really thorough review @lacostej, y'all are the best! 🎈:goat:

PR #11683, opened 2018-01-24
Reorder methods of Helper class in logical chunks

janpio:
This PR reorders the methods of the Helper class in logical chunks so similar methods are next to each other.


KrauseFx (approved):
Great idea

janpio:
Triggered by @lacostej in another PR review comment. I wasn't aware that `is_windows` is technically "deprecated" because it wasn't clearly visible in the file. Now it is :)

PR #11684, opened 2018-01-24
Replace usage of deprecated Helper methods

janpio:
Several of the Helper methods are "deprecated" and should not be used any more internally. Replaced all usages of those methods with their current counterparts.


KrauseFx (approved):
Amazing 👍 

We probably should still keep the methods around, in case plugins or Fastfiles use it, but maybe show a `UI.important` whenever it's called to stop encouraging their use

janpio:
For now they got their own "chunk" in #11683 at the bottom.

PR #11685, opened 2018-01-24
Instructions on how to run specific unit test or unit test group

janpio:
Some more instruction on how to run specific unit tests or unit test groups.

### DIFF
--- a/YourFirstPR.md
+++ b/YourFirstPR.md
@@ -66,7 +66,17 @@ If you know exactly which `_spec.rb` file you want to run, use
 bundle exec rspec ./fastlane/spec/fastlane_require_spec.rb
 ```
 
-(replace `./fastlane/spec/fastlane_require_spec.rb` with the path of your test file of course)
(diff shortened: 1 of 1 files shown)


YourFirstPR.md
```diff
@@ -66,7 +66,17 @@ If you know exactly which `_spec.rb` file you want to run, use
...
+
+#### Specific unit test (group) in a specific test file
+
+If you know the specific unit test or unit test group you want to run, use
+
+```
+bundle exec rspec ./fastlane/spec/fastlane_require_spec.rb:17
+```
+
+The number is the line number of the unit test (`it ... do`) or unit test group (`describe ... do`) you want to run.
```
KrauseFx: Instead of using the line number, maybe we should recommend using the `it "something", now: true` notation and then running `bundle exec rspec -t now`

janpio: The line number is returned on failing tests - that's how I learned about it :)

janpio: Done.

KrauseFx (approved):
Beautiful, improving all the docs 👍

PR #11689, opened 2018-01-24
Disable pager explicitly for  'git branch --list' command

robinkunde:
### Checklist

### Motivation and Context
Starting with v2.16.0, git by default runs the output of `git branch --list` through its pager. (https://github.com/git/git/commit/0ae19de74f6f5d6c6f9c80899e1ecd611c5b9827)
This causes FastlaneCore::CommandExecutor.execute() to return a string of invisible control characters in Match::GitHelper.branch_exists?(), where previously it would return an empty string. This in turn causes match to not create a branch when needed.

### Description
Fixed the issue by passing "--no-pager" option to git. I tested this option with some older versions of git as well, including v2.5, which is over 2 years old, and it doesn't seem to cause any problems.

### DIFF
--- a/match/lib/match/git_helper.rb
+++ b/match/lib/match/git_helper.rb
@@ -181,7 +181,7 @@ module Match
       return unless @dir
 
       result = Dir.chdir(@dir) do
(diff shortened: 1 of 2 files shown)


KrauseFx (approved):
Beautiful, nice catch, and thanks for your contribution 👍

PR #11691, opened 2018-01-25
[WIP] FixPty: alternative implementation to 11688

lacostej:
Pushed to trigger builds on Windows and verify it works.


janpio:
Hm, we have AppVeyor on manual right now (as the tests still fail by default) so you would probably have to create an account there yourself and create your fork to run the tests. Or you create a branch here in `fastlane/fastlane`, then I can manually start the test run on this. What do you prefer?

PR #11695, opened 2018-01-25
Allow to use lane name contains `:'

Watson1978:
### Checklist

### Motivation and Context
I'd like to have a namespace by using `:` in lane name in `Fastfile` like `Rakefile`.

### Description

This patch will allow to use `:` in lane name like a following:

```
fastlane_version "2.55.0"

default_platform :ios

platform :ios do
  lane :"update:version:minor" do
    puts "** Update minor version"
  end

  lane :"update:version:patch" do
    puts "** Update patch version"
  end
end
```

### DIFF
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -5,7 +5,13 @@ module Fastlane
     # @param [Array] args A hash of all options (e.g. --env NAME)
     def self.handle(args, options)
       lane_parameters = {} # the parameters we'll pass to the lane
-      platform_lane_info = [] # the part that's responsible for the lane/platform definition
+      platform_lane_info = args.slice!(0, 1) # the part that's responsible for the lane/platform definition
+      platform_or_lane = platform_lane_info[0]
+      if platform_or_lane && SupportedPlatforms.all.include?(platform_or_lane.to_sym)
+        # keep the lane name if platform name was given
+        platform_lane_info << args.shift
+      end
+
       args.each do |current|
         if current.include?(":") # that's a key/value which we want to pass to the lane
           key, value = current.split(":", 2)
--- a/fastlane/spec/command_line_handler_spec.rb
+++ b/fastlane/spec/command_line_handler_spec.rb
@@ -18,5 +18,12 @@ describe Fastlane do
                                                                   nil)
       Fastlane::CommandLineHandler.handle(["ios", "deploy", "key:true", "key2:false"], {})
     end
+
+    it "allows lane name contains `:'" do
+      expect(Fastlane::LaneManager).to receive(:cruise_lane).with("ios", "update:version:major",
+                                                                  { key: true },
+                                                                  nil)
+      Fastlane::CommandLineHandler.handle(["ios", "update:version:major", "key:true"], {})
+    end
   end
 end


fastlane/lib/fastlane/command_line_handler.rb
```diff
@@ -5,7 +5,8 @@ class CommandLineHandler
     # @param [Array] args A hash of all options (e.g. --env NAME)
     def self.handle(args, options)
       lane_parameters = {} # the parameters we'll pass to the lane
-      platform_lane_info = [] # the part that's responsible for the lane/platform definition
+      platform_lane_info = args.slice!(0, 2) || [] # the part that's responsible for the lane/platform definition
```
KrauseFx: I'm a little confused, usually `args` is nil here? Only when calling a lane with `:` that's not the case? How do you call a lane with `:`?

KrauseFx (commented):
Thanks for the PR @Watson1978 👍 Really appreciate it. I'm a little skeptical on the importance of this change vs the chances of it breaking something, but I totally see where you're coming from. I'm ok with going ahead, just added one question

fastlane/lib/fastlane/command_line_handler.rb
Watson1978: Ah, it is my mistake.
`args.slice!(0, 2)` always returns Array object. So, `|| []` was unnecessary.

fastlane/lib/fastlane/command_line_handler.rb
```diff
@@ -5,7 +5,8 @@ class CommandLineHandler
     # @param [Array] args A hash of all options (e.g. --env NAME)
     def self.handle(args, options)
       lane_parameters = {} # the parameters we'll pass to the lane
-      platform_lane_info = [] # the part that's responsible for the lane/platform definition
+      platform_lane_info = args.slice!(0, 2) # the part that's responsible for the lane/platform definition
```
janpio: what happens if we call just `fastlane lanename key1:value1 key2:value2`, without explicitly adding the platform (because the one we are using is defined as default)?

janpio (changes requested):
I think this won't work, see comment.

janpio:
(The test failure on Ubuntu would probably be resolved by rerunning the tests or rebasing on `master` where the underlying cause for this false fail is solved)

fastlane/lib/fastlane/command_line_handler.rb
Watson1978: if you will omit platform name at the command-line, seem that default platform name will be passed in here.

janpio: I don't think so, see here below your additions:

Watson1978: Ah, I understood your review.
I fixed my mistake at https://github.com/fastlane/fastlane/pull/11695/commits/b551fe3b279bfe580e16123dab88e457068c649f

Thank you.

fastlane/lib/fastlane/command_line_handler.rb
```diff
@@ -5,7 +5,13 @@ class CommandLineHandler
     # @param [Array] args A hash of all options (e.g. --env NAME)
     def self.handle(args, options)
       lane_parameters = {} # the parameters we'll pass to the lane
-      platform_lane_info = [] # the part that's responsible for the lane/platform definition
+      platform_lane_info = args.slice!(0, 1) # the part that's responsible for the lane/platform definition
+      platform_name = platform_lane_info[0]
+      if platform_name && SupportedPlatforms.all.include?(platform_name.to_sym)
```
janpio: Does `SupportedPlatforms.all` include the platforms that were added at runtime via https://docs.fastlane.tools/actions/add_extra_platforms/?

Watson1978: Hmm, seems `SupportedPlatforms.all` does not return extra platforms in here :(

Watson1978:
I realized that it is too hard to fix this issue by design. 

I will close this. Thank you for your review.

janpio:
Yeah, sorry I didn't have better news. 

A bit ago I had created a PR to this same code that wanted to add simple "flag" parameters (`key` instead of `key:true`) and also had to give up because of this. See here: https://github.com/fastlane/fastlane/pull/11002

Do you have an idea how we could improve this functionality?

Watson1978:
Sorry, I have no idea so far...

Issue #11700, opened 2018-01-25
[Regression] version_bump_podspec regex is broken with the introduction of swift_version in CocoaPods 1.4.0


tschob:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

CocoaPods 1.4.0 introduced a new podspec configuration field `swift_version`. As the version Regex from `version_bump_podspec` matches "version =..." (simplified), the Swift version is bumped instead of the Pod version.

Right field to bump: `spec.version`
Wrong field to bump: `spec.swift_version`

### Before `version_bump_podspec(bump_type: "minor")`

```ruby
  spec.version       = "1.2.3"
  spec.swift_version = "4.0"
```

### After `version_bump_podspec(bump_type: "minor")`

```ruby
  spec.version       = "1.2.3"
  spec.swift_version = "4.1.0"
```

### Expected

```ruby
  spec.version       = "1.3.0"
  spec.swift_version = "4.0"
```

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.13.1                                                        |
| Ruby                        | 2.3.3                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.14.3 (Apple Git-98)                              |
| Installation Source         | /usr/local/bin/fastlane                                        |
| Host                        | Mac OS X 10.13.1 (17B48)                                       |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib |
| OpenSSL Version             | LibreSSL 2.2.7                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | /Applications/Xcode-9.2.app/Contents/Developer/                |
| Xcode Version               | 9.2                                                            |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | de_DE.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

[...]

PR #11702, opened 2018-01-25
Update podspec CocoaPods version regex to work with new `swift_version` in CocoaPods 1.4.0

KrauseFx:
Fixes https://github.com/fastlane/fastlane/issues/11700


taquitos (approved):
Would it make sense to add a test to make sure it doesn't increment any other version fields?

KrauseFx:
Yep, great idea, I can do that another day

taquitos:
🎈:goat:

PR #11718, opened 2018-01-27
Fix duplicate loading of Gymfile. Was loaded twice when current dir == project root.

thii:
`load_configuration_file` is already trying to look for Gymfile in `fastlane`, `.fastlane` and the project folders.

https://github.com/fastlane/fastlane/blob/0c23d5e64f3c193ac013e9383551149ac6e8d88c/fastlane_core/lib/fastlane_core/configuration/configuration.rb#L163-L168

https://github.com/fastlane/fastlane/blob/0c23d5e64f3c193ac013e9383551149ac6e8d88c/fastlane_core/lib/fastlane_core/configuration/configuration.rb#L196-L204


lacostej (approved):
Nice catch

PR #11652, opened 2018-01-22
changed cert checker to use ruby openssl to generate sha1 fingerprint

Cyberbeni:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Changed the used openssl in cert_checker to generate sha1 fingerprint from the one used in terminal to the ruby version that's used in most places, because that's the one printing the version number in fastlane env.
https://github.com/fastlane/fastlane/pull/11641

https://stackoverflow.com/a/16529867/6324550

### Description

Using ruby openssl instead of terminal one in cert_checker.

PR #11720, opened 2018-01-29
[match/utils] using ruby openssl instead of terminal version

Cyberbeni:
### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/pull/11641
This function also worked incorrectly if the user had OpenSSL 1.1.x installed because output for subject had different delimiters.

### Description

Using ruby openssl instead of the locally installed terminal openssl to avoid errors from inconsistencies between versions.


Cyberbeni:
I think this is the last place where we are using terminal OpenSSL.
What does that appveyor.yml do? Should we also remove "- which openssl" from it, since we are only using the ruby openssl anymore?

match/lib/match/utils.rb
```diff
@@ -54,6 +51,8 @@ def self.get_cert_info(cer_certificate_path)
       return out_array.map { |x| x.split(/=+/) if x.include?("=") }
                       .compact
                       .map { |k, v| [openssl_keys_to_readable_keys.fetch(k, k), v] }
+                      .append([openssl_keys_to_readable_keys.fetch("notBefore"), cert.not_before])
+                      .append([openssl_keys_to_readable_keys.fetch("notAfter"), cert.not_before])
```
minuscorp: hmmmmm, souldn't be this be: `.append([openssl_keys_to_readable_keys.fetch("notAfter"), cert.not_after])`?

Cyberbeni: 😱 I'm blind. Fixed.

janpio:
`appveyor.yml` is for coming Windows CI, the `which openssl` output the path of the openssl used to debug problems. If the terminal openssl is not used anywhere any more, this can also removed.

Cyberbeni:
Ok, I removed it then. Once https://github.com/fastlane/fastlane/pull/11641 https://github.com/fastlane/fastlane/pull/11652 and this PR gets merged, we will only use ruby OpenSSL in fastlane

janpio (approved):

janpio:
🥇

PR #11731, opened 2018-01-29
Auto-collapse debug output in the Issue template

revolter:
This auto-collapses the (sometimes verbose) output using the
`details` tag, and also separates it with a gray background using
the `pre` tag.


taquitos (approved):
🎈:goat:
Great idea!

PR #11737, opened 2018-01-30
Fixed slack action in Swift.

antonino-u:
The default `slack` action doesn't work even if you override the two parameters that are broken. This is due to Ruby recognizing the default `"{}"` string as a string and not as a hash.

(I'm not a Ruby dev at all so if this isn't the right way to fix this please help me!)

### Checklist

### Motivation and Context

Slack action failling to send any messages.


antonino-u:
I signed it!

antonino-u:
I signed it!

taquitos:
Good news/bad news, 
Good news: that fix would work, so thank you for looking into it! 
Bad news: That file is actually autogenerated 

@thii has a PR open to fix the code generator 🎉  https://github.com/fastlane/fastlane/pull/11749 so once that is updated and merged, `Fastlane.swift` will be updated at the next deployment 🎈

antonino-u:
@taquitos I thought it might be autogenerated, but I wasn't 100% sure hehe. Thanks for looking at this, I'll close the PR now and wait for the one you mentioned to be merged. Cheers!

PR #11749, opened 2018-01-31
[Fastlane.swift] Fix wrong data type when convert Hash-type options to Swift

thii:
Options with `Hash` type were converted to `String` type in Swift.


fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -106,6 +106,8 @@ def get_type(param: nil, default_value: nil, optional: nil, param_type_override:
         type = "Bool"
       elsif default_value.kind_of?(Array)
         type = "[String]"
+      elsif default_value.kind_of?(Hash)
```
taquitos: Great catch

taquitos (changes requested):
This is great, thank you. 
If you can remove the changes to the `fastlane/swift/Fastlane.swift` file, we can get this merged.
Most of those `.swift` files are automatically generated during deployment time 🎉

taquitos (approved):

taquitos:
🎈:goat:

antonino-u:
@taquitos curious: this has apparently made it into 2.80.0 but the autogenerated files haven't been updated. There are also other fixes which I'm sure have been made that aren't in Fastlane.swift. Is there a way for me to force the project to regenerate them? Does someone have to do something to increment FastlaneRunnerAPIVersion?

taquitos:
You're right, it should have made it into this: https://github.com/fastlane/fastlane/pull/11760/files

If you check out the fastlane repo, you can run `fastlane generate_swift_api` and that will generate the files. It works for me, so I wonder if our deploy script doesn't work right.

taquitos:
Ahh, figured it out, we merged this PR after we version bumped. The version bump does the fastlane swift generation.

PR #11761, opened 2018-02-01
Allow CFPropertyList==3.0.0

untitaker:
### Checklist

### Motivation and Context

We use Spaceship not as a development tool, but on a server running JRuby. Unfortunately all versions in the allowed version range for CFPropertyList are breaking Enumerable under JRuby so fundamentally that even unrelated code crashes. The reason is a monkeypatch in CFPropertyList that enables support for MRI 1.8.

CFPropertyList 3.0.0 removed that monkeypatch, which means it drops support for MRI 1.8 while restoring compatibility with JRuby: https://github.com/ckruse/CFPropertyList/issues/52

### Description

This patch allows not only 2.x, but also 3.x versions of CFPropertyList to be used with fastlane. Either ombination works fine with MRI 2.4 in our tests, the combination with CFPropertyList 3.0.0 additionally works fine with JRuby.

Note that this patch technically still doesn't prevent people from using fastlane with MRI 1.8, because users who need the support can simply pin CFPropertyList to an older 2.x version.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('terminal-notifier', '>= 1.6.2', '< 2.0.0') # macOS notifications
   spec.add_dependency('terminal-table', '>= 1.4.5', '< 2.0.0') # Actions documentation
   spec.add_dependency('plist', '>= 3.1.0', '< 4.0.0') # Needed for set_build_number_repository and get_info_plist_value actions
-  spec.add_dependency('CFPropertyList', '>= 2.3', '< 3.0.0') # Needed to be able to read binary plist format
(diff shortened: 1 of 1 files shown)


untitaker:
I signed it!

KrauseFx:
Hey @untitaker, thanks for your contribution, I understand why this change is needed, however I'm worried this will have unexpected side effects and break something else. Does the `3.0` release have any breaking changes in their API as it's a major version bump?

untitaker:
@KrauseFx nope, see https://github.com/ckruse/CFPropertyList/commits/master, it's literally the only change in 3.0.0. And minor/patch versions after 3.0.0 shouldn't break anything (at least by semver rules)

untitaker:
(I think the reason why it's a major version bump is just because it's removing support for MRI 1.8)

KrauseFx (approved):
Alright, let's give this a shot 👍

Issue #11790, opened 2018-02-06
Swift wrapper can't create push certificates


J-Bossi:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Running the getPushCertificate action from FastlaneSwift is not working. The Swift wrapper passes a String instead of an Int for the activeDaysLimit parameter.

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>DEBUG [2018-02-06 11:30:13.75]: received command:#<Fastlane::ActionCommand:0x007f86b6220d28 @method_name="get_push_certificate", @class_name=nil, @command_id="", @args=[named argument: development, value: true, type: , named argument: generate_p12, value: true, type: , named argument: active_days_limit, value: 30, type: , named argument: force, value: true, type: , named argument: save_private_key, value: true, type: , named argument: app_identifier, value: com.xxxx.xxxx, type: , named argument: username, value: xxxx@xxxx.xx, type: , named argument: p12_password, value: xxxxxx, type: , named argument: output_path, value: ., type: ]>
INFO [2018-02-06 11:30:13.80]: ----------------------------------
INFO [2018-02-06 11:30:13.80]: --- Step: get_push_certificate ---
INFO [2018-02-06 11:30:13.80]: ----------------------------------
ERROR [2018-02-06 11:30:13.95]: Error setting value '30' for option 'active_days_limit'
ERROR [2018-02-06 11:30:13.95]: You passed invalid parameters to 'get_push_certificate'.
ERROR [2018-02-06 11:30:13.95]: Check out the error below and available options by running `fastlane action get_push_certificate`
[...]

PR #11791, opened 2018-02-06
Fastlane.swift: Fix getPushCertificate type

J-Bossi:
Changing type of activeDaysLimit to Int to creathe pushcertificates

### Checklist

### Motivation and Context
This change fixes a bug in the Swift getPushCertificate wrapper. The type for the activedaysLimit needs to be an Int
This solves issue #11790

I tested it locally and it works

### DIFF
--- a/fastlane/swift/Fastlane.swift
+++ b/fastlane/swift/Fastlane.swift
@@ -1515,7 +1515,7 @@ func getProvisioningProfile(adhoc: Bool = false,
 }
 func getPushCertificate(development: Bool = false,
                         generateP12: Bool = true,
-                        activeDaysLimit: String = "30",
+                        activeDaysLimit: Int = 30,
                         force: Bool = false,
                         savePrivateKey: Bool = true,
                         appIdentifier: String,


J-Bossi:
I signed it!

J-Bossi:
I really signed it

minuscorp (approved):
This should be the second part of #11793, so It looks good to me.

janpio (approved):

thii:
I think Fastlane.swift shouldn't be committed because it's autogenerated.

janpio:
Ohoh. @taquitos?

taquitos:
@thii good catch, that's right fastlane.swift is autogenerated based off the ConfigItems. I just made a PR for the root of the issue: https://github.com/fastlane/fastlane/pull/11845

During fastlane deployment, we autogenerate all the Swift stuff, so while you don't see any swift changes now, you will when we do our version bump pr 🎉

janpio:
(no need to revert this then?)

taquitos:
That's right, no need to revert, there shouldn't be a conflict when we update the swift api.

Issue #9840, opened 2017-07-24
Unable to upload symbols (macOS High Sierra)


c-st:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description
I am running Fastlane v2.48 under macOS High Sierra. When attempting to submit my dSYMs to Crashlytics via `upload_symbols_to_crashlytics` it fails:

```
❯ fastlane beta --capture_output
[00:16:45]: For more information, check out https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
DEBUG [2017-07-25 00:16:48.27]: Checking if there are any plugins that should be loaded...
INFO [2017-07-25 00:16:48.40]: -------------------------------------------------
INFO [2017-07-25 00:16:48.40]: --- Step: Verifying required fastlane version ---
INFO [2017-07-25 00:16:48.40]: -------------------------------------------------
INFO [2017-07-25 00:16:48.40]: Your fastlane version 2.48.0 matches the minimum requirement of 2.48  ✅
INFO [2017-07-25 00:16:48.41]: ------------------------------
INFO [2017-07-25 00:16:48.41]: --- Step: default_platform ---
INFO [2017-07-25 00:16:48.41]: ------------------------------
INFO [2017-07-25 00:16:48.41]: Driving the lane 'ios beta' 🚀
INFO [2017-07-25 00:16:48.41]: -------------------------------------------
INFO [2017-07-25 00:16:48.41]: --- Step: upload_symbols_to_crashlytics ---
INFO [2017-07-25 00:16:48.41]: -------------------------------------------
INFO [2017-07-25 00:16:48.57]: Extracting '/Users/chris/Development/TheApp/TheApp/TheApp .app.dSYM.zip'...
INFO [2017-07-25 00:16:48.57]: $ unzip -qo /Users/chris/Development/TheApp/TheApp/TheApp .app.dSYM.zip
INFO [2017-07-25 00:16:49.37]: Uploading 'RealmSwift.framework.dSYM'...
ERROR [2017-07-25 00:16:49.42]: Shell command exited with exit status 8 instead of 0.
INFO [2017-07-25 00:16:49.42]: Uploading '34DB257B-AC09-337D-8211-B3AD89434965.dSYM'...
[...]

PR #11798, opened 2018-02-08
Option to upload symbols to Firebase Crashlytics via GoogleService-Info.plist

jonasrottmann:
### Checklist

### Motivation and Context
The `upload_symbols_to_crashlytics` Action doesn't yet support passing the _GoogleService-Info.plist_ when wanting to upload to Firebase Crashlytics. This can now be done with the `-gsp` parameter like shown in the [Firebase Crashlytics documentation](https://firebase.google.com/docs/crashlytics/get-deobfuscated-reports).
This also addresses https://github.com/fastlane/fastlane/issues/9840#issuecomment-361908666.

### Description
The action has now a new option with the `:gsp_path` (env var name `GOOGLE_SERVICES_INFO_PLIST_PATH`). The action now only throws an error if neither one of Fabric API key or _GoogleService-Info.plist_ is given.

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -8,6 +8,11 @@ module Fastlane
 
         find_binary_path(params)
         find_api_token(params)
+        find_gsp_path(params)
+
+        if !params[:api_token] && !params[:gsp_path]
+          UI.user_error!('Either Fabric API key or path to Firebase Crashlytics GoogleService-Info.plist must be given.')
+        end
 
         dsym_paths = []
         dsym_paths << params[:dsym_path] if params[:dsym_path]
@@ -79,7 +84,8 @@ module Fastlane
         UI.message("Uploading '#{path}'...")
         command = []
         command << File.expand_path(params[:binary_path]).shellescape
-        command << "-a #{params[:api_token]}"
+        command << "-a #{params[:api_token]}" if params[:api_token]
+        command << "-gsp #{params[:gsp_path]}" if params[:gsp_path]
         command << "-p #{params[:platform] == 'appletvos' ? 'tvos' : params[:platform]}"
         command << File.expand_path(path).shellescape
         begin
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
```diff
@@ -148,6 +162,15 @@ def self.available_options
                                        verify_block: proc do |value|
                                          UI.user_error!("No API token for Crashlytics given, pass using `api_token: 'token'`") if value.to_s.length == 0
                                        end),
+          FastlaneCore::ConfigItem.new(key: :gsp_path,
+                                      env_name: "GOOGLE_SERVICES_INFO_PLIST_PATH",
+                                      sensitive: true,
```
taquitos: I think this would be better served as `code_gen_sensitive` instead of `sensitive`. Sensitive is usually used for API keys or passwords so that if we prompt for it, you'll get `****` instead of plain text. `code_gen_sensitive` prevents our code generation from pulling in this value when we generate the Swift API.

jonasrottmann: Overlooked that one 🙈 Changed it and fixed the indentation 🙂

taquitos: 🎉

taquitos (approved):
🎈:goat:
Thank you!

niveus:
Would it be possible to remove this check?
```ruby
if !params[:api_token] && !params[:gsp_path]
  UI.user_error!('Either Fabric API key or path to Firebase Crashlytics GoogleService-Info.plist must be given.')
end
```
The script actually checks your Info.plist and gets the API token, so the error is not correct.

janpio:
(Best create a new issue where you link to this one. If this is blocking you and worked before, including "Regression" in the title to get the attention of a developer more quickly.)

jonasrottmann:
@niveus I don't get why you want to remove the whole check. The method [`find_api_token(params)`](https://github.com/jonasrottmann/fastlane/blob/422620e7bc13d28327c77676a67ee901c997052a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb#L100) looks for the token and sets it if found. The message is a bit misleading and we should maybe add something like "...or the key could not be found". Without the check you will just see the actual Fabric upload script to fail and I'd rather give the user feedback as soon as possible.
I only tested my changes with the _GoogleService-Info.plist_ but I don't see why this would break anything, or did I miss something?

niveus:
I’ve never used the parameter or the googleservice-info.plist. The old style is just to put it in your Info.plist where the script found it. Now fastlane dies with that UI.error that’s been inserted

jonasrottmann:
The action is still able to find the token in the _Info.plist_ which happens in the `find_api_token(params)` method which has not been touched and which is called before this check. But like I said: I only tested with the _GoogleService-Info.plist_.

Could be that the Fabric script `upload-symbols` itself also looks for the token but as far as I know this behavior isn't documented and I wouldn't trust it. 🤔

If your error still exists please open a new issue and reference this thread. 👍

PR #11804, opened 2018-02-08
testfairy: removed watermark, added defaults to auto-update

yairba:
### Checklist


janpio:
Why did you remove the watermark option?

yairba:
It is an old feature that is not supported any more.
Here is the updated list: https://docs.testfairy.com/API/Upload_API.html

taquitos (commented):
Before merging, can you also remove the code that makes use of the `icon_watermark` value in `testfairy.rb`?

taquitos (approved):
🎈:goat:

PR #11838, opened 2018-02-14
Fix getting certificate info in match/utils.rb

kkhaidukov:
### Checklist

### Motivation and Context

After the latest update `match` fails with ```undefined method `append' for #<Array:0x007f8cbf8fee88>```.
### Description

This PR fixes the issue by replacing `append`, which is not a valid array method, with `push`.
I tested the changes on my local setup to make sure the issue with match is gone.


kkhaidukov:
I signed it!

minuscorp (approved):
Nice catch! Thank God this is only used to print the table information of `match`!

kkhaidukov:
@minuscorp yeah, but in our case it broke our build script because match exited with non-0

kkhaidukov:
@minuscorp @janpio can we have a release?

PR #11875, opened 2018-02-19
Spaceship: support subscription_pricing_target when creating IAP.

byohay:
### Checklist

### Motivation and Context
`IAPDetails` has a field called `subscription_pricing_target` that, that can be set with price tier and currency. If the property is set when the product is `save!`d, `pricing_intervals` will be populated with all the countries, and each country will have the price tier that has the closest price to the price set in `subscription_pricing_target` (it will not be necessarily have the same price tier).
Previous to this PR, it was not possible to create a recurring product with `pricing_intervals` using the `subscription_pricing_target`. One needed to create the product, wait for the creation to finish, and then update it using `IAPDetails`.
In addition, there are several places where the `pricing_intervals` that is received from the user of `spaceship` is transformed to raw `pricing_intervals` sent to Apple. This PR removes the duplicates and puts the transformation in one place.

I tested it by creating a new product and updating an existing product.

### Description
A new method that returns the transformed `pricing_intervals` was added to `tunes_client`. In addition the method `IAP::create!` was added a `subscription_pricing_target` argument.


minuscorp:
Hi @byohay! I think this PR is great, just some comments, would you be able to add a test case of stubbed creation request-response? That'd be great, as this code changes existing code. 😄

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -544,6 +544,41 @@ def update_price_tier!(app_id, price_tier)
       handle_itc_response(r.body)
     end
 
+    def transform_to_raw_pricing_intervals(app_id = nil, purchase_id = nil, pricing_intervals = nil, subscription_price_target = nil)
```
minuscorp: Can you explain me why this method belongs to the a client instance instead of a class method in  `iap_detail`?

byohay: Sure. Even though `iap.rb` requires `iap_detail.rb`, the class `IAP` doesn't depend on `IAPDetail`. And it makes sense to me because I don't see an obvious connection between those classes.
The only place I could think that can be accessed by both classes was `TunesClient`. But then again I'm not that familiar with the codebase, is there a better place to put it?

byohay:
Hi @minuscorp, I added tests and replied to you comment :)

minuscorp:
Hey @joshdholtz, this seems legit to me, should we merge it?

joshdholtz:
@minuscorp Code and tests look 💯  but I'm going to try and run this first 😊

joshdholtz:
@byohay Question! I'm getting an error when running this - see below for my lane and response. Am I creating the purchase right? The purchase ends up showing up in iTC like 1 minute later. It seems like iTC might be running slow (today) which is causing a fetch/retry to not work which is causing this to fail?

Any ideas? 😊 

```rb
lane :create_iap_sub_price_target do
  require 'spaceship'
  Spaceship::Tunes.login("email")
  app = Spaceship::Tunes::Application.find("bundle")

  app.in_app_purchases.create!(
  type: Spaceship::Tunes::IAPType::RECURRING, 
  versions: {
    "en-US" => {
      name: "sub2",
      description: "This is the sub2 description"
    }
  },
  reference_name: "sub2",
  family_id: "famfam",
  product_id: "a.b.c.d",
  cleared_for_sale: true,
  review_notes: "A note for a reviewer",
  review_screenshot: "test_nonconsume1.png", 
  subscription_duration: "1m", 
  subscription_price_target: [{ currency: "EUR", tier: 1 }]
  )
end
```

gets me

```sh
[!] The request could not be completed because:
	Failed to find the product with id=com.joshdholtz.FastlaneTest.sub2. This can be caused either by a server error or due to the removal of the product.
```

byohay:
Hi @joshdholtz, I ran your code multiple times, it did occur quite a lot. AFAIS The only thing we can do is increase the retry count.

BTW there are two problems with the parameters you pass to `create!`:
- `family_id` should be the id of the subscription group,.
- `subscription_price_target` accepts a hash, not an array, like so: `{ currency: "EUR", tier: 1 }`.

joshdholtz:
@byohay Ah, thanks for the feedback! I will give things another try 😊

LUJYM:
@byohay  How to solve this problem? I also encountered such a problem. and it show "unknown keyword: subscription_price_target"

spaceship/lib/spaceship/tunes/iap.rb
```diff
@@ -78,30 +88,20 @@ def create!(type: "consumable",
...
 
         # Update pricing for a recurring subscription.
-        if type == Spaceship::Tunes::IAPType::RECURRING && pricing_intervals
+        if type == Spaceship::Tunes::IAPType::RECURRING &&
+           (pricing_intervals || subscription_price_target)
           # There are cases where the product that was just created is not immediately found,
           # and in order to update its pricing the purchase_id is needed. Therefore polling is done
           # for 4 times until it is found. If it's not found after 4 tries, a PotentialServerError
           # exception is raised.
           product = find_product_with_retries(product_id, 4)
```
joshdholtz: Can we bump this retry to maybe 6? (also update comments too). This seems to be a decently slow request on Apple's side 🙃

joshdholtz (changes requested):
Woot, I got it working once I manually bumped up the retry number and changed the thing to a hash 👍 
```rb
app.in_app_purchases.create!(
  type: Spaceship::Tunes::IAPType::RECURRING, 
  versions: {
    "en-US" => {
      name: "sub#{num}",
      description: "This is the sub#{num} description"
    }
  },
  reference_name: "sub#{num}",
  family_id: "XXXXXXX",
  product_id: "com.joshdholtz.FastlaneTest.sub#{num}",
  cleared_for_sale: true,
  review_notes: "A note for a reviewer",
  review_screenshot: "test_nonconsume1.png", 
  subscription_duration: "1m", 
  subscription_price_target: { currency: "EUR", tier: 1 }
)
```

Only request I think it is to bump the number of retries

joshdholtz:
@LUJYM Did you download this branch?

spaceship/lib/spaceship/tunes/iap.rb
```diff
@@ -52,6 +52,16 @@ def families
...
+      # @param subscription_price_target (Hash) Only used on RECURRING purchases, used to set the
+      # price of all the countries to be roughly the same as the price calculated from the price
+      # tier and currency given as input.
+      # @example:
+      #  [
+      #    {
+      #      currency: "EUR",
+      #      tier: 2
+      #    }
+      #  ]
```
joshdholtz: Oh, this is where I got the array from ^ Should this example really just be a hash?

spaceship/lib/spaceship/tunes/iap.rb
byohay: Sure, does it work all the time with 6?

spaceship/lib/spaceship/tunes/iap.rb
byohay: Oh right! Fixed, my bad 👍

byohay:
@joshdholtz updated to 6 retries, does it work every time now?

joshdholtz:
@byohay Did you push your changes? 🙃 But yup! When I changed my local branch to 6 retries yesterday things worked every time 🎊 It was low still but it worked 🎉

byohay:
Yep, changes were pushed.

joshdholtz:
@byohay Oh mannnnn, I was totally looking for another commit 😛 That was my bad! I will take another look at this PR then, test it out again, and then it will probably be 👍 to merge

superandrew213:
@joshdholtz what is the status on this?

byohay:
@joshdholtz checks are finally passing 😄

joshdholtz:
@byohay Will review and test this again today!

joshdholtz (approved):
We are merging this! 💪 😊

PR #11882, opened 2018-02-20
Add option to disable any kind of formatting on the output

pepicrft:
### Checklist

### Motivation and Context
The CI platform that we are using, Buildkite, supports grouping if the logs have the following format:

```bash
--- this is a group
# then something else is printed
# and another thing
--- and this is another group
# with some logs inside
```
However, when we use Fastlane, our groups lines get prefixed with the `▸` character and coloured with a magenta color. As a result, the grouping doesn't work.

### Description
I've added a new environment variables, `FASTLANE_DISABLE_OUTPUT_FORMAT` that disables any kind of formatting on the output.

### DIFF
--- a/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
+++ b/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
@@ -72,8 +72,12 @@ module FastlaneCore
     def command_output(message)
       actual = (message.split("\r").last || "") # as clearing the line will remove the `>` and the time stamp
       actual.split("\n").each do |msg|
-        prefix = msg.include?("▸") ? "" : "▸ "
-        log.info(prefix + "" + msg.magenta)
+        if FastlaneCore::Env.truthy?("FASTLANE_DISABLE_OUTPUT_FORMAT")
+          log.info(msg)
+        else
+          prefix = msg.include?("▸") ? "" : "▸ "
+          log.info(prefix + "" + msg.magenta)
+        end
       end
     end
 
--- a/fastlane_core/README.md
+++ b/fastlane_core/README.md
@@ -20,9 +20,10 @@ This gem contains all shared classes and code:
 
 You can hide the inline changelog by setting the `FASTLANE_HIDE_CHANGELOG` environment variable
 
-## Timestamps
+## Output environment variables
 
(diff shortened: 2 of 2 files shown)


nafu (approved):
Hi @pepibumur 
Thank you for your contribution 😊 
This looks great to me 👍

janpio:
Nice PR. 
Aren't there already some ENV variables or config options to disable colors and such?

nafu:
@janpio 

> colors

Yes, we have for colors. `FASTLANE_DISABLE_COLORS`

pepibumur:
Do you think it makes sense to group the two flags or should we use a new flag for this?

fastlane_core/lib/fastlane_core/ui/implementations/shell.rb
```diff
@@ -72,8 +72,12 @@ def command(message)
     def command_output(message)
       actual = (message.split("\r").last || "") # as clearing the line will remove the `>` and the time stamp
       actual.split("\n").each do |msg|
-        prefix = msg.include?("▸") ? "" : "▸ "
-        log.info(prefix + "" + msg.magenta)
+        if FastlaneCore::Env.truthy?("FASTLANE_DISABLE_OUTPUT_FORMAT")
+          log.info(msg)
```
nafu: @pepibumur 

> Do you think it makes sense to group the two flags or should we use a new flag for this?

Grouping the two flags means using `log.info(msg.magenta)` instead?

If so, only disabling prefix looks better to me 👍 
And it would be better rename `FASTLANE_DISABLE_OUTPUT_FORMAT` to `FASTLANE_DISABLE_COMMAND_OUTPUT_PREFIX` or something.

pepibumur: Thinking about it I think disabling any kind of formatting/prefixing makes more sense. Instead of adding flags for each individual element, it's just one flag that disables everything.

pepibumur:
Is there anything missing to get it merged?

KrauseFx (approved):
Looks good from my side also, thanks @pepibumur. @nafu do you want to merge?

nafu:
@pepibumur This is awesome 🎉

Thanks @janpio @KrauseFx for the review.

PR #11889, opened 2018-02-21
Setting keychain path to decode provisioning profiles

### LINKED ISSUE #11814: [match] Getting error when default keychain is not set
### New Issue Checklist

### Issue Description

I'm running the Fastlane on Jenkins, in order to use the same Mac for more than one build at time and as I'm using enterprise and company account, I need to set the keychain that I want to use for each build.

To achieve that, I'm using the follow action on my Fastfile:

`setup_jenkins(set_default_keychain: false, add_keychain_to_search_list: true)`

After other stuffs, I'm calling `match`, using `Matchfile` to set my keychain path.

When `match` is trying to install the provisioning profile, I'm getting this error:

<details>
<pre>
[16:35:15]: Installing provisioning profile...
[16:35:15]: Failure to decode /var/folders/r3/xnw05cjx61sb6s4yf46bz6w00000gn/T/d20180209-67857-ambn3w/profiles/appstore/AppStore_com.enfore.pos.mobileprovision. Exit: 1: security: cert import failed: A default keychain could not be found.
security: problem decoding
</pre>
</details>

After some research on Fastlane code, I found the issue, but I don't know how to fix that in a good way. So I would like to introduce to someone that can fix that easily. It's a simple fix, add `-k` parameters with the current keychain path on this line below:

https://github.com/fastlane/fastlane/blob/83f9895f557293862a9f7546e0ff06cd4b3c1440/fastlane_core/lib/fastlane_core/provisioning_profile.rb#L82

As the parameter `-k` is not in use, it's trying to use the default keychain, but as I said, I'm not using a default keychain.

For now I set my Jenkins to execute one build at time to avoid problems, but would be nice to have that fix soon.

Thanks.

bguidolim:
### Checklist

### Motivation and Context

This PR fix #11814 

I configured my CI (Jenkins) to run different jobs using the same project and machine specifying the directories dynamically to support parallels builds using Fastlane.

In my case, I have `enterprise` and `appstore` distribution and even forcing the keychain name and/or path, `match` action cannot decode provisioning profiles if there is no default keychain set in the machine. This PR allows to set a specific keychain to use with no need to set a default keychain.

### Test cases

I have tested these cases below:

 - `match` with no default keychain;
 - `match` with default keychain;
 - `match` with 2 or more different keychains;
 - `match` with 2 or more different keychains in a different order; (required to get WWDR certificate)

### Description

This PR adds `keychain_path` parameter for some methods in [Provisioning Profile](https://github.com/fastlane/fastlane/blob/73bffaaac9f16ac34a129819f50edec49dda608d/fastlane_core/lib/fastlane_core/provisioning_profile.rb) class, if no value is provided, it will use default keychain or throw an exception if a default keychain is not set (current behaviour).

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -23,10 +23,10 @@ module FastlaneCore
       #   "TimeToLive"=>185,
       #   "UUID"=>"1752e382-53bd-4910-a393-aaa7de0005ad",
       #   "Version"=>1}
-      def parse(path)
+      def parse(path, keychain_path = nil)
         require 'plist'
 
-        plist = Plist.parse_xml(decode(path))
+        plist = Plist.parse_xml(decode(path, keychain_path))
         if (plist || []).count > 5
           plist
         else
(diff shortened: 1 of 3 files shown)


taquitos (approved):
This is awesome. Thank you!
🎈:goat:

PR #11893, opened 2018-02-22
[WIP] Operating System compatibility of tools and actions 

### LINKED ISSUE #11859: [Discuss] Operating System compatibility of tools and actions
The next step in cross-platform (Windows and Linux) support is to find a way to handle code and actions that can not work on a specific platform in a better way than crashing 💣 

For the test suite, we for now solved this by disabling the tests with tags. The tests are not run ("pending test") and the test suite as a whole can pass. But this of course only works for tests - there is no similar mechanism that can warn a user when they use an action in their Fastfile - yet.

--- 

My current idea is to add a list of "operating system compatibility" to each action: 

- known good,
- known bad
- or unknown. 

By default (and if this "compatibility mapping" is missing) all actions are assumed to be "known good" for macOS (they were developed on and for macOS after all) and "unknown" for all other operating systems. 

After manual testing and confirmation, this can manually (Pull Request to the action source code) be changed to "known good" or "known bad" in the code.

Before an action is run, a global "before" method that is run for all tools/actions can check this mapping and output a warning in the console or even block the execution of the tool/action with an error message.

---

This concept is quite similar to how the platform support (on which an action can be used in the `Fastfile`) is defined:

```
  [:ios, :mac].include?(platform)
end
```

This is used to generate the list of supported platforms in the docs: 

https://github.com/fastlane/fastlane/blob/8a2e328892c34ff5d1d450a759a480cb325c5008/fastlane/lib/assets/ActionDetails.md.erb#L22

But also to output an error if the platform is not listed: 

https://github.com/fastlane/fastlane/blob/8a2e328892c34ff5d1d450a759a480cb325c5008/fastlane/lib/fastlane/runner.rb#L294-L304

---

The definition of compatibility could look like this (pseudocode):

```
is_compatible(operating_system)
  compatibility = { 
    "macos": true, 
    "linux": false, 
    "windows": null 
  }
  return compatibility[operating_system]
end
```

This method, then, could be used in similar ways than `is_supported`: generate documentation on operating system compatibility of tools/actions and error/warning generation at runtime.

(I think it might also be possible to switch the "pending tests" from a static tag written in the test file to also use this compatibility mapping on runtime to decide if a test should be run on the platform or not.)

---

What do you think?

--- 

PS: An additional useful feature could be to be able to define a "message" regarding a "false" value: 

> This action uses Xcode which is not available for this operating system.

The pending tests already do something like that, so it would be neat to keep that. The `compatibility` definition would have to be different then, of course.

janpio:
### Checklist

### Description

closes #11859

### DIFF
--- a/fastlane/lib/fastlane/action.rb
+++ b/fastlane/lib/fastlane/action.rb
@@ -151,7 +151,7 @@ module Fastlane
     end
 
     def self.is_compatible?(operating_system, action_compatibility = {})
-      # TODO find way to motivate developer to do so implement `is_compatible` in action
+      # TO DO find way to motivate developer to do so implement `is_compatible` in action
       # e.g. throw warning/error message/whatever when `is_compatible` is not implemented in child class (actual action) that is being called
       # similar to what `is_supported?` does:
       # if action_compatibility == {}
--- a/fastlane/lib/fastlane/action.rb
+++ b/fastlane/lib/fastlane/action.rb
@@ -150,16 +150,16 @@ module Fastlane
       nil
     end
 
-    def self.is_compatible?(operating_system, compatibility = nil)
-      # UI.crash!("Implementing `is_compatible?` for all actions is mandatory. Please update #{self}")
+    def self.is_compatible?(operating_system, action_compatibility = {})
       # TODO find way to throw warning/error message when method is not implemented in child class that is being called
-      if compatibility.nil?
-        compatibility = {
-          "MacOS"   => true,  # We assume macOS to be good 
-          "Linux"   => nil,
-          "Windows" => nil
-        }
-      end
+      default_compatibility = {
+        #                   # Unless specified otherwise...
+        "MacOS"   => true,  # we assume action on macOS will work
+        "Linux"   => nil,   # we assume action on Linux might be ok
+        "Windows" => nil,   # we assume action on Windows might be ok
+        "Unkown"  => false  # we assume action on other OS will not work
+      }
+      compatibility = default_compatibility.merge(action_compatibility)
       return compatibility[operating_system]
     end
   end
--- a/fastlane/lib/fastlane/action.rb
+++ b/fastlane/lib/fastlane/action.rb
@@ -162,7 +162,7 @@ module Fastlane
         "MacOS"   => true,  # we assume action on macOS will work
         "Linux"   => nil,   # we assume action on Linux might be ok
         "Windows" => nil,   # we assume action on Windows might be ok
-        "Unkown"  => false  # we assume action on other OS will not work
+        "Unknown" => false  # we assume action on other OS will not work
       }
       compatibility = default_compatibility.merge(action_compatibility)
       return compatibility[operating_system]
--- a/fastlane/lib/fastlane/action.rb
+++ b/fastlane/lib/fastlane/action.rb
@@ -151,7 +151,12 @@ module Fastlane
     end
 
     def self.is_compatible?(operating_system, action_compatibility = {})
-      # TODO find way to throw warning/error message when method is not implemented in child class that is being called
+      # TODO find way to motivate developer to do so implement `is_compatible` in action
+      # e.g. throw warning/error message/whatever when `is_compatible` is not implemented in child class (actual action) that is being called
+      # similar to what `is_supported?` does:
+      # if action_compatibility == {}
+      #   UI.crash!("Implementing `is_compatible?` for all actions is mandatory. Please update #{self}")
+      # end
       default_compatibility = {
         #                   # Unless specified otherwise...
         "MacOS"   => true,  # we assume action on macOS will work
--- a/fastlane/lib/fastlane/action.rb
+++ b/fastlane/lib/fastlane/action.rb
@@ -149,5 +149,18 @@ module Fastlane
     def self.deprecated_notes
       nil
     end
+
+    def self.is_compatible?(operating_system, compatibility = nil)
+      # UI.crash!("Implementing `is_compatible?` for all actions is mandatory. Please update #{self}")
+      # TODO find way to throw warning/error message when method is not implemented in child class that is being called
+      if compatibility.nil?
+        compatibility = {
+          "MacOS"   => true,  # We assume macOS to be good 
+          "Linux"   => nil,
+          "Windows" => nil
+        }
+      end
+      return compatibility[operating_system]
+    end
   end
 end
--- a/fastlane/lib/fastlane/action.rb
+++ b/fastlane/lib/fastlane/action.rb
@@ -45,7 +45,7 @@ module Fastlane
     end
 
     def self.details
-      nil # this is your change to provide a more detailed description of this action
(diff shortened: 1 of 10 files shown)


janpio:
Open questions:
- How to properly test this?
- How to stop code execution if `is_compatible` returns `false`? (output, exception, crash, return false?)
- How to motivate developers to implement `is_compatible` in their actions? (`is_supported` uses an `UI.crash!`)

Open TODOs:
- Implement default `is_compatible` in plugin action template.
- Check generated action documentation.

revolter:
> How to stop code execution if is_compatible returns false? (output, exception, crash, return false?)
(See fastlane/fastlane/pull/11893/files#diff-394215798b66f7b366c9d2d7ecee29e2R316)

I think that this call:

https://github.com/fastlane/fastlane/blob/7792837cf2e6b781f546689706793208a38f1345/fastlane/lib/fastlane/runner.rb#L315

could be replaced with a call to `user_error`, that, IIRC, automatically stops the executing lane.

janpio:
> > 1. How to properly test this?
>
> Do we need to have unit tests for this 

Definitely - but my question referred to the code I wrote for `is_compatible` to work. I want to make sure that this thing is super stable as it will be executed before each and every action is executed.

> ... or wait for people to use the respective actions on a Linux machine?
> Having a unit test passing on Linux for an action means that we can mark it with true?

But those are good questions as well. In general we assume that the fastlane codebase is fully covered with tests, and as almost all tests are passing on Linux and Windows we could assume that all will be good.

But I don't fully trust that yet, which is why I prefer marking both new platforms (Linux, Windows) as `nil` meaning "this will probably work, but we don't know for sure" yet and only change this if it is manually set to `true` in the action. 

Question then is what is enough to set an action to true. Unit tests as you suggest? Actual usage of all parameters? I don't know yet. Suggestions?


> > 2. How to stop code execution if is_compatible returns false? (output, exception, crash, return false?)
>
> I think that this call [...] could be replaced with a call to user_error, that, IIRC, automatically stops the executing lane.

That's a good idea. From fastlane_core README:

> fastlane "crash" because of a user error everything that is caused by the user and is not unexpected

Here is the actual implementation:

https://github.com/fastlane/fastlane/blob/41cc2955e38cb568f0caa6b96d6dcf28d67b0511/fastlane_core/lib/fastlane_core/ui/interface.rb#L124-L134

This might be a nice alternative, that doesn't blame the user for the problem (and probably also doesn't pollute analytics):

https://github.com/fastlane/fastlane/blob/41cc2955e38cb568f0caa6b96d6dcf28d67b0511/fastlane_core/lib/fastlane_core/ui/interface.rb#L167-L177

>> 3. How to motivate developers to implement is_compatible in their actions? (is_supported uses an UI.crash!)
>
> I'd say that using crash! is a correct approach, but don't take my word for it.

Yeah, that is to harsh in this case. Actions _need_ an `is_supported`. But `is_compatible` has a default setting, so it also works without an implementation. It just would be nice if developers would test their actions on all platforms and make that compatibility explicit by implementing the method and setting the hash.

revolter:
I'd argue that calling an action on Linux that doesn't support this platform is a user error. It's something that the user is doing while he/she isn't supposed to do so. That's why we also fail early instead of letting it crash.
Also, maybe we do want to know how many people are (or maybe will be) calling actions on unsupported platforms.

janpio:
> Also, maybe we do want to know how many people are (or maybe will be) calling actions on unsupported platforms.

Good point, will keep that in mind and do some research what is possible there.

janpio:
`user_error!` does indeed stop the execution, which can be looked at with the first failing test: https://circleci.com/gh/fastlane/fastlane/16028

fastlane/lib/fastlane/action.rb
```diff
@@ -149,5 +149,23 @@ def self.other_action
...
+      # similar to what `is_supported?` does:
+      # if action_compatibility == {}
+      #   UI.crash!("Implementing `is_compatible?` for all actions is mandatory. Please update #{self}")
+      # end
+      default_compatibility = {
+        #                   # Unless specified otherwise...
+        "MacOS"   => true,  # we assume action on macOS will work
+        "Linux"   => nil,   # we assume action on Linux might be ok
+        "Windows" => nil,   # we assume action on Windows might be ok
+        "Unkown"  => false  # we assume action on other OS will not work
```
joshdholtz: Should be `Unknown`

fastlane/lib/fastlane/action.rb
```diff
@@ -149,5 +149,23 @@ def self.other_action
     def self.deprecated_notes
       nil
     end
+
+    def self.is_compatible?(operating_system, action_compatibility = {})
+      # TODO find way to motivate developer to do so implement `is_compatible` in action
```
joshdholtz: We could probably increase plugin rating if this is defined ^

fastlane/lib/fastlane/actions/adb_devices.rb
```diff
@@ -68,6 +68,15 @@ def self.authors
...
         platform == :android
       end
+
+      def self.is_compatible?(operating_system)
+        compatibility = {
+          "MacOS"   => true,
+          "Linux"   => false,
+          "Windows" => true
+        }
+        super(operating_system, compatibility)
```
joshdholtz: Is it possible to do this without calling `super`? I don't think we use super anywhere else so not sure if we want to try and make things consistent when people try to implement them? But I do also like this approach too 🤔 

Is it possible to also just do something like...
```rb
def self.is_compatible?(operating_system)
  [:macos, :windows].include?(operating_system)
end
```

fastlane/lib/fastlane/action.rb
janpio: knew that looked wonky somehow...

fastlane/lib/fastlane/actions/adb_devices.rb
janpio: > Is it possible to do this without calling super?

I don't think so as I wanted it to have a "default compatibility" that is present for all actions by default - which would currently be all of them. 

(The `is_supported?` which we already have and is a bit similar doesn't need that as it was required from the beginning and calling an action without it actually stops execution with an error message)

I actually like this better as this way each action doesn't "replicate" the logic to check - but of course this is only important here as the checking logic is a bit more complicated with the defaults.

> Is it possible to also just do something like...
> `[:macos, :windows].include?(operating_system)`

I think using just an array only works for yes/no situations. But be have yes/no/unknown to express that it 1) works, 2) doesn't work or 3) hasn't been explicitly set/tested.

fastlane/lib/fastlane/action.rb
janpio: That's a great idea.

janpio:
After not actively thinking about this for half a year but testing fastlane on Windows and Ubuntu, I think it might make more sense to simplify this to only add a "setting" to the actions that definitely do not work on a platform. No `default_compatibility` and just an optional `is_not_compatible` list of platforms where an action definitely does not work (yet).

Issue #11913, opened 2018-02-23
Disable metrics collection


taquitos:
Making changes to fastlane metrics infrastructure. For now, we'll disable all collection until we decide what metrics, if any, would help us serve our users better.

PR #11914, opened 2018-02-23
Disable metrics collection

taquitos:
For #11913

Disable all metrics collection for the time being.


KrauseFx (approved):
This is great 💯

PR #11923, opened 2018-02-25
YES/NO auto convert fix

sovcharenko:
### Checklist

### Motivation and Context
I'm developing a plugin that updates xcconfig files and noticed that `'YES'/'NO'` or `'true'/'false'` strings passed as action's config gets automatically converted to `true/false` (boolean) despite config's type is explicitly set to `String`. This is causing issues to me as my code is unable to get the initial value that was set by the user - was it `YES` or `true`?

### Description
I've updated the code that auto converts `YES/NO/true/false` strings into booleans so that it check config's type before performing conversion. If type was explicitly set to `String` auto conversion will not happen.

Hope that makes sense.


KrauseFx (approved):
Nice catch, and great solution, thanks @sovcharenko 👍 Let's hope this doesn't break anything else

PR #11924, opened 2018-02-25
Add plugin scores caching

revolter:
The "update_docs" lane went down from 10 minutes to 1 and a half
minutes, and the "plugin_scores" action went down from 9:40
minutes to a little over 1 minute 🎉

The other advantage is that there are fewer requests to the GitHub API made ❤️

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -338,6 +338,7 @@ lane :update_docs do
   verify_env_variables
 
   template_path = File.expand_path("./assets/render_plugin.md.erb")
+  plugin_scores_cache_path = File.expand_path("./plugin_scores_cache.yml")
   clone_docs do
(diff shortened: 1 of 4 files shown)


milch (approved):
LGTM 👍🏼 One nitpick maybe: would it make more sense to use the SHA as hash key and delete entries that are unused maybe? Using the name makes intuitive sense, but figuring out that the cache does the right thing takes quite a bit more time because you have to go through a bunch of code

revolter:
@milch, I see what you mean, but I think that not having the name as the key would make the file hard to read in case you need to debug it. And wouldn't the delete unused entries part add extra complexity too?

milch:
Right, deleting unused entries would definitely complicate things a bit, but it's a nice thing to have regardless 

The name could obviously still be included in the hash. But like I said, I'm fine with merging this either way

PR #11935, opened 2018-02-27
Some actions fail if true is passed as one of their params

sovcharenko:
### Checklist

This PR fixes the following:

1.  **Some actions fail if 'true' is passed as one of their params.**

If action's option is expected to receive boolean and it's `type` or `is_string: false` not set  - the action will fail with the following error:

```
'xxx' value must be a String! Found TrueClass instead.
```
| Action | Params | Try it yourself |
| ------------- | ------------- | ------------- | 
| artifactory  | ssl_verify  |  fastlane run artifactory ssl_verify:true |
| nexus_upload |  ssl_verify |  fastlane run nexus_upload ssl_verify:true | 
| version_bump_podspec |  require_variable_prefix |  fastlane run version_bump_podspec require_variable_prefix:true |
|  version_get_podspec | require_variable_prefix | fastlane run version_get_podspec require_variable_prefix:true |
| zip | verbose | fastlane run zip verbose:true |

Fix: Set type to be non-string either with `type: Boolean` or `is_string: false` (I've used `is_string` in some files despite it's deprecated just to match the existing style)

2. **Option validation routine is not executed for `false` values.**

I've found that validation routine (`verify_block` and `type` validation) is not executed for `false` values. The reason - Fastlane doesn't do validation for `nil` values, however, there is a bug in the code (I don't think that was done intentially) that skips validation for falsey values instead of just for `nil`s.

Fix: Skip validation just for `nil`s.

This fix broke `import_certificate`, `scan` and `snapshot` actions and they begin to crash with error similar to the first one (`false` started to go through validation routine). I've updated these actions as well and set their types to Boolean instead of TrueClass (which I think is not correct anyway and actually they were the only ones that used True/FalseClass as option type).

As a side note: I came across the second issue when I was building my custom action that require boolean option validation. After diving into the problem I've found the first one.


janpio:
What's the difference between `is_string: false` and `type: Boolean`?

sovcharenko:
It doesn't look like regression to me. I was playing around with verify_block and noticed that verify_block doesn't get called for `false` values that confused me a little. Further investigations led me to an existing issues with some actions. I guess, those params are not frequently used with `true` values therefore these issues remained unnoticed for a while.

`is_string: false`: will turn off type checks.
`type: XXXX`: if set, value will checked to be a kind of XXXX.

```ruby
# @param is_string *DEPRECATED: Use `type` instead* (Boolean) is that parameter a string? Defaults to true. If it's true, the type string will be verified.
# @param type (Class) the data type of this config item. Takes precedence over `is_string`. Use `:shell_string` to allow types `String`, `Hash` and `Array` that will be converted to shell-escaped strings
```

And this is the relation between the two:
```ruby
# Determines the defined data type of this ConfigItem
def data_type
  if @data_type.kind_of?(Symbol)
    nil
  elsif @data_type
    @data_type
  else
    (@is_string ? String : nil)
  end
end
```

If none of them set, the value will be checked to be a `String`, which obviously fails for booleans. At the moment it only fails for `true` values because of the second issue and that is showcased in the first one.

KrauseFx:
Great question @janpio, as we did have a change in boolean values with yesterday's fastlane release

sovcharenko:
I highly doubt that this is connected. I’m  sure that I was seeing this issue in the release before the last one when I was working on #11923

KrauseFx (approved):
Looks good to me, thanks for solving this issue, and thanks for confirming it's unrelated to the other change 👍

sovcharenko:
Just for the record. Issue with `false` values not being validated introduced in 70c815b33080d2d996a997cf59e654c8f7646f0c back in 2015.

PR #11939, opened 2018-02-27
Delete keychain checks the keychain list before deleting

ugenlik:
### Checklist

### Motivation and Context

Our CI pipeline includes multiple apps, each has a different Distribution Certificate. CI pipeline's macmini slave throws errors when other app's Distribution Certificate's is in the Keychain.

Suggested solution for above problem is to create then delete a keychain in fastfile. This works until your build/tests fails.... 

Let's say you create a keychain

```
    create_keychain(
        name: ENV["KEYCHAIN_NAME"],
        unlock: true,
        timeout: false,
        lock_when_sleeps: false,
        password: ENV["MATCH_PASSWORD"]
    ) 
 
    //some tests
    // some builds

    delete_keychain(name: ENV["KEYCHAIN_NAME"])
```
if your app fails before delete_keychain and next time when you run the fastlane file create_keychain won't check if there is already a keychain exists with ` ENV["KEYCHAIN_NAME"]` and it will try to create it anyway. which will fail and throw and error. 

So then you need to go an manually delete the keychain you created previously. 

It is time consuming and a terrible practise, it works against CI mindset.... 

So I thought either create_keychain should check if the keychain exists then don't create it again. 
Or delete check if keychain exists if so deletes the keychain to overcome errors.

VERY VERY IMPORTANT added optional `throw_error` make sure if you pass false , fastlane doesn't throw an error, default is set to true so previous versions can use the `delete_keychain`  without changing it  

```
delete_keychain(name: ENV["KEYCHAIN_NAME"],throw_error:false)
```

So the new lane would look like

```
    delete_keychain(name: ENV["KEYCHAIN_NAME"],throw_error:false)  # call before create so if keychain exists, it deletes it 
    create_keychain(
        name: ENV["KEYCHAIN_NAME"],
        unlock: true,
        timeout: false,
        lock_when_sleeps: false,
        password: ENV["MATCH_PASSWORD"]
    ) 
 
    //some tests
    // some builds

    delete_keychain(name: ENV["KEYCHAIN_NAME"])
```

**Tests:**

- works with keychain name found locally
- works with keychain name found in ~/Library/Keychains
- works with keychain name found in ~/Library/Keychains with -db" (sometimes make creates the keychain as -db so make sure to handle that situation)
- works with absolute keychain path
- shows an error message if the keychain can't be found (optional)
- shows an error message if neither :name nor :keychain_path is given 

**PS: I have never developed in Ruby and rspec, so thake this code with a grain of salt**


ugenlik:
I signed it!

ugenlik:
I signed it!

Baumchen:
> if your app fails before delete_keychain and next time when you run the fastlane file create_keychain won't check if there is already a keychain exists with ENV["KEYCHAIN_NAME"] and it will try to create it anyway. which will fail and throw and error.

Just out of curiosity: Why don't you use `begin / rescue / finally` to ensure that the keychain is deleted after your build?

http://crodrigues.com/try-catch-finally-equivalent-in-ruby/
```
begin
    some_code
rescue
     handle_error
ensure
    this_code_is_always_executed
end
```

Or `before all` / `before each` / `after all` / `after each`? https://docs.fastlane.tools/advanced/

Couldn't that solve your problem?

ugenlik:
@Baumchen that is an excellent question, I think it is because I am not proficient in Ruby, so I didn't know if such thing existed.

`before all` would solve my problem but I'd like the keep the fastfile simple as possible and let the fastlane handle details.

I think rather than just trying to delete a keychain bluntly ,  it'd be nice to check if it exists before and not throw an error when it doesn't exist is a decent request.

joshdholtz:
@ugenlik Thanks for the work on this PR! However, I don't think that introducing this new option into `delete_keychain` is the direction we would like to go at this moment as there are other solutions that will work for this problem.

I think the proper solution would be using fastlane `after_all` `error` block... You could also just run `rescue` after the `delete_keychain` action if you don't care if it fails

## Solution 1
```rb
# This `rescue nil` will eat the error since we don't care if it fails
delete_keychain(name: ENV["KEYCHAIN_NAME"]) rescue nil
create_keychain(
    name: ENV["KEYCHAIN_NAME"],
    unlock: true,
    timeout: false,
    lock_when_sleeps: false,
    password: ENV["MATCH_PASSWORD"]
) 
```

## Solution 2: Using fastlane blocks

```rb
# Gets called after successful run of all the lanes 
after_all do |lane, options|
  delete_keychain(name: ENV["KEYCHAIN_NAME"])
end

# Gets called after an error occurs
error do |lane, exception, options|
  delete_keychain(name: ENV["KEYCHAIN_NAME"])
end
```

Thanks again for your work put into this PR ❤️

PR #11947, opened 2018-02-28
Add use_bundle_exec option to pod_push action

nicolasbraun:
### Motivation and Context

The `pod_lib_lint` action offer an action to use (or not) the local Gemfile
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/pod_lib_lint.rb#L7

The `pod_push` does not which can cause sometimes the lib lint to pass and then  the push to fail

### Description
Added the same option to the `pod_push` action

I choose to default this to false, for backward compatibility but it could be defaulted to true for consistency with `pod_lib_lint` (and it seems to be the legit expectation if the project has a Gemfile)

### DIFF
--- a/fastlane/lib/fastlane/actions/pod_push.rb
+++ b/fastlane/lib/fastlane/actions/pod_push.rb
@@ -2,7 +2,7 @@ module Fastlane
   module Actions
     class PodPushAction < Action
       def self.run(params)
-        command = ''
+        command = []
 
         command << "bundle exec" if params[:use_bundle_exec] && shell_out_should_use_bundle_exec?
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/pod_push.rb
```diff
@@ -2,11 +2,15 @@ module Fastlane
   module Actions
     class PodPushAction < Action
       def self.run(params)
+        command = ''
+
+        command << "bundle exec" if params[:use_bundle_exec] && shell_out_should_use_bundle_exec?
```
KrauseFx: Mh, are you sure this works? Looks like there is no `&&` inbetween the commands, so this should crash? Did it work when you tested it locally?

nicolasbraun: Hello @KrauseFx 

Sorry I must confess I do not have a pod project to test again, I will wait for a prod release and test.
As for your remark I think I do not need && since i want to run `bundle exec pod ...` instead of `pod ..` 

Still I got confused because `command` is a collection for the lib lint action and here a simple string. I think I am just missing a space after the word exec, do you agree ?

nicolasbraun:
@KrauseFx fixed and tested! I aligned the coding style with the other actions that constructs shell commands

joshdholtz:
@nicolasbraun I have totes needed this fix before 🙌 The code looks good but can you add a test to `pod_push_spec.rb` for this 😊

nicolasbraun:
@joshdholtz test added 🚀 
What do you think of defaulting it to true, see me remark in the PR initial message

joshdholtz:
@nicolasbraun I think default this to false for backward compatibility (like you stated in the description) is probably for the best 🤔

nicolasbraun:
Allright, if test are ok we are good to go then :)

nicolasbraun:
@joshdholtz any chance this gets in the next release?

joshdholtz (approved):

PR #11953, opened 2018-02-28
Update example code to match real code.

ZevEisenberg:
### Checklist

### Motivation and Context
I noticed this discrepancy in the docs at https://docs.fastlane.tools/actions/slack/. The sample code references a `thumb_url`, but it looks like this key has changed to `icon_url` and the sample code wasn't updated.

### Description
I changed `thumb_url` to `icon_url` to make the sample code match the real code.


janpio:
I `thumb` and `icon` really the same thing?

Icon was added 2 years ago: https://github.com/fastlane/fastlane/commit/ce0b80e95dca9cfae4f7d02f7e10a48d1d33bd86

The thumb example code "only" a year ago:
https://github.com/fastlane/fastlane/commit/261115f73bef21b11273843261bc05e5cb1ec682

ZevEisenberg:
Oh, maybe they’re not? I just saw the mismatch in the sample code and I guess I assumed that one had supplanted the other.

janpio:
Yeah, I did as well at first but then started to investigate and now I am not sure any more :/ 

Actually, on https://api.slack.com/docs/message-attachments there is a `thumb_url` so I think the docs are actually correct right now. Do you agree?

ZevEisenberg:
Ah, yes, it looks like it's a Slack property, not a fastlane property. Good catch! Maybe the fastlane docs should link to the Slack doc you posted so you can learn more about which properties are valid.

janpio:
It does, I got the link from the description of `attachment_properties` - it is just not a clickable link.

ZevEisenberg:
well I'll be

PR #11964, opened 2018-03-01
Shellscape java_home ENV var in Crashlytics helper

taquitos:
Fix for #11255

This fix makes it so people can use spaces in their `JAVA_HOME` path again.


KrauseFx (approved):
Nice catch 👍

Issue #11094, opened 2017-12-01
download_dsyms action crashes when downloading from multiple iTunesConnect teams


faithfracture:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I am part of several different Apple Developer / iTunesConnect teams for different variations of the same(ish) app. I have a `refresh_dsyms` lane in my Fastfile that will connect to and download the dsyms for all my apps across all my teams. For multiple apps in the same team, this is working fine, but as soon as it switches to a different team, everything fails. I can go and manually switch the order in which the dsyms are downloaded so that a different team's apps are downloaded first and that will work for that team, so it's not a problem with a specific team, it's switching from one team to another. This lane _used_ to work. Last time I ran it was a couple months ago, so I can't say exactly when it stopped working.

### Captured Output

Command Used: `refresh_dsyms iosVersion:1.4.0 buildNumber:406 beta:true`

I've edited out the sensitive information (I hope...), but the important stuff is still there.

<details><summary>Output/Log</summary>

```

DEBUG [2017-12-01 13:53:11.94]: Checking if there are any plugins that should be loaded...
DEBUG [2017-12-01 13:53:11.94]: Loading 'fastlane-plugin-update_provisioning_profile_specifier' plugin
INFO [2017-12-01 13:53:11.98]: gem 'fastlane-plugin-update_provisioning_profile_specifier' is already installed
DEBUG [2017-12-01 13:53:11.99]: Loading 'fastlane-plugin-badge' plugin
INFO [2017-12-01 13:53:12.02]: gem 'fastlane-plugin-badge' is already installed
+-------------------------------------------------------+---------+---------------------------------------+
|                                              Used plugins                                               |
+-------------------------------------------------------+---------+---------------------------------------+
| Plugin                                                | Version | Action                                |
+-------------------------------------------------------+---------+---------------------------------------+
| fastlane-plugin-update_provisioning_profile_specifier | 1.3.0   | update_provisioning_profile_specifier |
| fastlane-plugin-badge                                 | 1.0.0   | add_badge                             |
+-------------------------------------------------------+---------+---------------------------------------+

[...]

PR #11970, opened 2018-03-02
Reinitialize Spaceship::TestFlight.client if Spaceship::Tunes.client is auth with different team id

mnespor:
🔑

### Checklist

### Motivation and Context

This change allows the download_dsyms action to be run with multiple team IDs in the same session.

This addresses Github issue #11094  

First I tested the TestFlight client in `irb`:
```> require 'spaceship'
> Spaceship::Tunes.login
> ENV['FASTLANE_ITC_TEAM_ID'] = '12345'
> Spaceship::Tunes.select_team
> Spaceship::Application.all.first.all_build_train_numbers
> Spaceship::Tunes.login
> ENV['FASTLANE_ITC_TEAM_ID'] = '67890'
> Spaceship::Tunes.select_team
> Spaceship::TestFlight::Base.invalidate_client
> Spaceship::Application.all.first.all_build_train_numbers
```

After that started working properly, I pointed my iPhone project's Gemfile at my local fastlane clone and ran my dsym refresh lane:

```
  lane :refresh_dsyms do
    flavors = Flavors.flavors.values.reject { |flavor| flavor.nonprod? }
    flavors.each do |flavor|
      download_dsyms(username: flavor.team.tunes_username,
                     app_identifier: flavor.bundle_id,
                     team_id: flavor.team.tunes_id,
                     version: 'latest')
    end

    upload_symbols_to_crashlytics(api_token: CRASHLYTICS_API_TOKEN)
    clean_build_artifacts
  end
```

### Description

This PR adds a new method, `Spaceship::TestFlight::Base.invalidate_client`, which nils the client out. After that, the next call to `client` forces a fresh clone from the iTunes Connect client.

The download_dsyms action now invalidates the TestFlight client after logging in. The TestFlight client can get stuck with an outdated cookie if it's not invalidated.

Thanks for your time!

### DIFF
--- a/spaceship/lib/spaceship/test_flight/base.rb
+++ b/spaceship/lib/spaceship/test_flight/base.rb
@@ -5,10 +5,16 @@ module Spaceship
   module TestFlight
     class Base < Spaceship::Base
       def self.client
-        # Initialize new client if none or team changed
-        if Spaceship::Tunes.client && (@client.nil? || @client.team_id != Spaceship::Tunes.client.team_id)
-          @client = Client.client_with_authorization_from(Spaceship::Tunes.client)
+        # Verify there is a client that can be used
+        if Spaceship::Tunes.client
+          # Initialize new client if new or if team changed
+          if @client.nil? || @client.team_id != Spaceship::Tunes.client.team_id
+            @client = Client.client_with_authorization_from(Spaceship::Tunes.client)
+          end
         end
+
+        # Need to handle not having a client but this shouldn't ever happen
+        raise "Please login using `Spaceship::Tunes.login('user', 'password')`" unless @client
 
         @client
       end
--- a/spaceship/lib/spaceship/test_flight/base.rb
+++ b/spaceship/lib/spaceship/test_flight/base.rb
@@ -8,6 +8,10 @@ module Spaceship
         @client ||= Client.client_with_authorization_from(Spaceship::Tunes.client)
       end
 
+      def self.invalidate_client
+        @client = nil
+      end
+
       ##
       # Have subclasses inherit the client from their superclass
       #
--- a/spaceship/lib/spaceship/test_flight/base.rb
+++ b/spaceship/lib/spaceship/test_flight/base.rb
@@ -6,10 +6,10 @@ module Spaceship
     class Base < Spaceship::Base
       def self.client
         # Initialize new client if none or team changed
-        if @client.nil? || @client.team_id != Spaceship::Tunes.client.team_id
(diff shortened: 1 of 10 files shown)


mnespor:
I signed it!

joshdholtz:
@mnespor Thanks for idea! I improved upon your thought a bit to re-auth in `self.client` if the `team_id` changed 😊 Teamwork makes the dreamwork

mnespor:
@joshdholtz I love this <3 Thanks for the review!

spaceship/lib/spaceship/test_flight/base.rb
```diff
@@ -5,7 +5,12 @@ module Spaceship
   module TestFlight
     class Base < Spaceship::Base
       def self.client
-        @client ||= Client.client_with_authorization_from(Spaceship::Tunes.client)
+        # Initialize new client if none or team changed
+        if Spaceship::Tunes.client && (@client.nil? || @client.team_id != Spaceship::Tunes.client.team_id)
```
KrauseFx: What if `Spaceship::Tunes.client` is `nil`? Can it ever be `nil`? We handled it differently before this change I guess?

KrauseFx (commented):
Interesting, the PR title isn't accurate I think? This PR changes all of _spaceship_, right? So all methods support multiple teams now, and it would automatically re-login.

All of this is a little dangerous, that's just because it's a very foundational change, on how _spaceship_ works, and it might have some side effects. I'm a little unsure of what other areas this could potentially break

spaceship/lib/spaceship/test_flight/base.rb
joshdholtz: I guess before this change it would have crashed?
```rb
@client ||= Client.client_with_authorization_from(Spaceship::Tunes.client)
```
Should we error out if its nil??

joshdholtz:
@KrauseFx Yeah, I put some commits in here along with @mnespor so I should probably change the PR title 😇 

But I think it was more dangerous before this PR 😬 It was strongly caching the client before and not allowing for a new `Spaceship::TestFlight.client` if there was a new `Spaceship::Tunes.client`. This meant that `Spaceship::Tunes.client` and `Spaceship::TestFlight.client` were out of sync which was causing a bit 💥 

And this isn't **all** of spaceship - just the TestFlight parts.

spaceship/lib/spaceship/test_flight/base.rb
taquitos: I think we should error out, this would be a bug in fastlane if `@client` is nil by the end of this.

taquitos (approved):
Night and thorough. This should help a lot of people encountering credential issues when switching teams!
🌞:goat:

PR #11978, opened 2018-03-03
Add Support for Registering Mac Devices on Apple Developer Portal

Baumchen:
### Checklist

### Motivation and Context
Currently the `register_devices` action does not support to add Mac devices to the Apple Developer Portal. `Spaceship::Device` does support adding Mac devices though.
Changes have been tested running the action with the Mac parameter set and adding actual Mac devices to Apples Developer Portal.

### Description
A new parameter `platform` is introduced.
The default value of `platform` will be taken from `Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]` if available, "ios" otherwise.

For Mac we need a different regular expression for validating UDIDs (in the form of `12345678-1234-1234-1234-123456789012`).
The `Spaceship::Device.all` and `Spaceship::Device.create` methods both take a parameter `mac: Bool` which we're using now.
The flag is set simply based on comparing the `platform` parameter listed in `params`.

Additionally we're now fetching all devices before adding any:
The documentation states:
          "This is an optimistic action, in that it will only ever add new devices to the member center, and never remove devices. If a device which has already been registered within the member center is not passed to this action, it will be left alone in the member center and continue to work."
The current implementation doesn't check if a devices UDID is listed already when passing devices via the `devices` parameter, not the `devices_file` parameter.

### DIFF
--- a/fastlane/lib/fastlane/actions/register_devices.rb
+++ b/fastlane/lib/fastlane/actions/register_devices.rb
@@ -3,10 +3,11 @@ require 'credentials_manager'
 module Fastlane
   module Actions
     class RegisterDevicesAction < Action
-      UDID_REGEXP = /^\h{40}$/
+      UDID_REGEXP_IOS = /^\h{40}$/
+      UDID_REGEXP_MAC = /^[\h\-]{36}$/
 
       def self.is_supported?(platform)
-        platform == :ios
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Tested this out and this is fantastic! Thank you 😊

Issue #11642, opened 2018-01-20
Failing test with Xcode's DerivedData handling set to relative to workspace


Cyberbeni:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
1 test failing when I run `bundle exec rspec` because my Xcode is set to store DerivedData relative to the workspace.

##### Complete output when running fastlane, including the stack trace and command used

`bundle exec rspec`

```
  1) Scan Scan::DetectValues Xcode config handling fetches the path from the Xcode config
     Failure/Error: expect(derived_data).to match(%r{/.*Xcode/DerivedData/app-\w*$})

       expected "/Users/cyberbeni/Desktop/iOS/fastlane/scan/examples/standard" to match /\/.*Xcode\/DerivedData\/app-\w*$/
       Diff:
       @@ -1,2 +1,2 @@
       -/\/.*Xcode\/DerivedData\/app-\w*$/
       +"/Users/cyberbeni/Desktop/iOS/fastlane/scan/examples/standard"

     # ./scan/spec/detect_values_spec.rb:12:in `block (4 levels) in <top (required)>'
```

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.6                                     |
| Ruby                        | 2.5.0                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.16.0                          |
| Installation Source         | /usr/local/bin/fastlane                     |
| Host                        | Mac OS X 10.12.6 (16G1114)                  |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/2.5.0/lib            |
| OpenSSL Version             | OpenSSL 1.0.2n  7 Dec 2017                  |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 9.2                                         |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
[...]

PR #12013, opened 2018-03-08
Derived data test

Cyberbeni:
### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/issues/11642

### Description

Changed `Scan::DetectValues` test to include workspace relative DerivedData location via `WorkspaceSettings.xcsettings` and expect project path as derived data path


scan/spec/detect_values_spec.rb
```diff
@@ -3,13 +3,15 @@
     describe 'Xcode config handling' do
       before do
         options = { project: "./scan/examples/standard/app.xcodeproj" }
+        FileUtils.mkdir_p("./scan/examples/standard/app.xcodeproj/project.xcworkspace/xcuserdata/#{ENV['USER']}.xcuserdatad/")
+        FileUtils.copy("./scan/examples/standard/WorkspaceSettings.xcsettings", "./scan/examples/standard/app.xcodeproj/project.xcworkspace/xcuserdata/#{ENV['USER']}.xcuserdatad/WorkspaceSettings.xcsettings")
```
joshdholtz: Is this `ENV['USER`]` needed in these tests? Is it possible for us to put a static path there instead?

joshdholtz (commented):
One question!

scan/spec/detect_values_spec.rb
Cyberbeni: These settings can't be in the shared xcsettings, so we need to use the current user.

joshdholtz: 👌

joshdholtz (approved):
💪

PR #12014, opened 2018-03-08
Added update block to update_url_schemes action

### LINKED ISSUE #12012: Is it possible to add new url_scheme keeping existing one? "Add" instead of "update all"?
### New Issue Checklist

### Issue Description

Hi. Thanks for a great tool.
I wonder is it possible to add new url_scheme to info.plist file? 
As I see "update_url_schemes" rewrite all existing schemes.
So is it possible to just add new url_scheme to the array of schemes?
Or at least manually get all existing schemes? And add them again with the new one?  
Thanks.

joshdholtz:
Fixes #12012

 `update_url_schemes` only allowed users to override existing schemes (without seeing what was being overridden. Now there is a `update_url_schemes` block that can be passed in (instead of `url_schemes`) which passes in the existing schemes

### Option 1: With block (new way)
```rb
update_url_schemes(
  update_url_schemes: proc do |schemes|
    puts "existing schemes: #{schemes}"
    schemes + ["yooooo"]
  end
)
```

### Option 2: With string array (old way)
```rb
update_url_schemes(
  url_schemes: ["heyyyy", "hoooo", "heeeeee"],
)
```

### DIFF
--- a/fastlane/lib/fastlane/actions/update_url_schemes.rb
+++ b/fastlane/lib/fastlane/actions/update_url_schemes.rb
@@ -6,10 +6,42 @@ module Fastlane
       def self.run(params)
         path = params[:path]
         url_schemes = params[:url_schemes]
+        update_url_schemes = params[:update_url_schemes]
 
         hash = Plist.parse_xml(path)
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
📮

PR #12015, opened 2018-03-08
Update google-api-client dependency

minuscorp:
In order to have the ability to operate with the Google Cloud Storage API in `fastlane.ci`, we have to bump the dependency of the gem in _fastlane_.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -92,7 +92,7 @@ Gem::Specification.new do |spec|
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
   # If you upgrade this gem, make sure to upgrade the users of it as well.
-  spec.add_dependency('google-api-client', '>= 0.13.1', '< 0.14.0') # Google API Client to access Play Publishing API
+  spec.add_dependency('google-api-client', '>= 0.19.0', '< 0.20.0') # Google API Client to access Play Publishing API
 
   # Development only
   spec.add_development_dependency('rake', '< 12')


joshdholtz:
@minuscorp What does this mean? `# If you upgrade this gem, make sure to upgrade the users of it as well.` Do you know? 😇

minuscorp:
I really don't know 🤷‍♂️ @KrauseFx

joshdholtz:
I think this PR will require **A LOT** of manually testing. I vaguely remember that writing stubbed out tests against this gem wasn't easy at all so our tests alone _might_ not show all breaks.

Example: The last minor upgrade had a breaking change - https://github.com/fastlane/fastlane/pull/10137

This is a few minor upgrades so might have to look further into what all possibly could have changed? 🤔

KrauseFx:
Yah, I'm scared also - but it feels like we're eventually gonna have to upgrade to the new version, otherwise we collect more and more technical debt. When the update broke something, we never actually went ahead and fixed the root of the issue, but only reverted the change. @joshdholtz I guess now things are still super busy, but this might be a very impactful project for you to take on once things get a little more quiet?

joshdholtz:
Yeah yeah, I can definitely start chipping away at this. I'm slowly getting a nice suite of tests for Android tools built up to test things which should make testing upgrades like this a bit easier 😊

But I don't think I feel comfortable yolo merging this yet though until all the bits are heavily tested 😇

minuscorp:
Of course! This is not blocking _fastlane.ci_ at all (as for now we can just downgrade the dep) but in a future  revisit this PR to ensure we can provide both _fastlane_ and _fastlane.ci_ latest updates as possible! Thank you!!

joshdholtz:
I completely agree! Being a few minor versions behind doesn't feel great. I actually did run a few things with newest `google-api-client` yesterday when debugging some `supply` issues and it didn't crash which was good - but that definitely wasn't enough testing 😏

vanniktech:
This has already been updated by @joshdholtz recently when he added support for app bundles.

KrauseFx:
👍

joshdholtz:
Going to close this one since already updated for bundles

PR #12031, opened 2018-03-11
Ignore "be" test method for MethodCallWithArgsParentheses cop

revolter:
The actual reason for this is that the same config file is used for new plugins, and I needed for `be` to be (pun intended) ignored.

If it doesn't make sense, or there are many more test methods that I don't know of and we don't want to add them all, feel free to close this.


KrauseFx (approved):
💯

PR #12037, opened 2018-03-12
get_version_number uses xcodeproj as fallback when can't find "target"

### LINKED ISSUE #11645: get_version_number retrieves version number of test target
My issue is same reported [here](https://github.com/fastlane/fastlane/issues/1490)
According to this closed issue, it was fixed in [#3944](https://github.com/fastlane/fastlane/pull/3944) but it is not working.

My fastlane version is 2.77.1.

My xcode project has two targets and output is as below for my lane.

```
lane :testvar do

		version     = get_version_number(xcodeproj: "abc.xcodeproj")
		build       = get_build_number(xcodeproj: "abc.xcodeproj")
		puts " Current Version is *#{version}* Build(#{build})"
	end
```
```
[20:51:54]: Driving the lane 'ios testvar' 🚀
[20:51:54]: --------------------------------
[20:51:54]: --- Step: get_version_number ---
[20:51:54]: --------------------------------
[20:51:54]: $ cd /Users/xxx/Documents/Developer/XCode\ Projects && agvtool what-marketing-version -terse
[20:51:55]: ▸ "abc.xcodeproj/../abcUITests/Info.plist"=1.0
[20:51:55]: ▸ "abv.xcodeproj/../Resources/Info.plist"=0.2.0
[20:51:55]: ------------------------------
[20:51:55]: --- Step: get_build_number ---
[20:51:55]: ------------------------------
[20:51:55]: $ cd /Users/xxx/Documents/Developer/XCode\ Projects && agvtool what-version -terse
[20:51:56]: ▸ 8
[20:51:56]:  Current Version is *1.0* Build(8)
```

I tried these but doesnt work.
`version     = get_version_number(xcodeproj: "abc.xcodeproj", target:"abc")  `
`version     = get_version_number(xcodeproj: "abc.xcodeproj", scheme:"abc")  `

What could be wrong?

Thanks,
Ashish

joshdholtz:
Fixes #11645

## What was happening
`get_version_number` has an option for `target` but this wasn't actually looking for the target - it was doing string comparisons on path to find the name (or parent directory) of the **plist file** (which is most likely named after the target). Because of the comparison on the plist file name or parent directory, the results of this action were sometimes not finding the version number even though the version number **should** have been found

## Solution
- Use `xcodeproj` as a fallback when the legacy comparisons fail
 - Generates map of target name (key) and absolute plist file path (value)
 - When can't find version legacy way, keys for target in map (making sure plist path matches) and finds correct version that `agvtool` prints out
- Also changed tests to use actual Xcode project
- Changed action to use backticks instead of `Action.sh` so that tests could properly be run on `agvtool`
  - `Action.sh` doesn't run in tests properly (it returns what was passed in instead)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/spec/actions_specs/get_version_number_action_spec.rb
```diff
@@ -3,62 +3,71 @@
     describe "Get Version Number Integration" do
       require 'shellwords'
 
+      path = File.absolute_path("./fastlane/spec/fixtures/actions/get_version_number/get_version_number/")
+
       it "gets the correct version number for 'TargetA'" do
         result = Fastlane::FastFile.new.parse("lane :test do
-          get_version_number(xcodeproj: '.xcproject', target: 'TargetA')
+          get_version_number(xcodeproj: '#{path}', target: 'TargetA')
```
revolter: You can replace this with just `path` instead of `'#{path}'`

fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -4,132 +4,80 @@ class GetVersionNumberAction < Action
...
+        target_name = params[:target]
+        configuration = params[:configuration]
 
-        command_prefix = [
-          'cd',
-          File.expand_path(folder).shellescape,
-          '&&'
-        ].join(' ')
+        # Get version_number 
+        project = get_project!(folder)
```
revolter: Aren't `!` used for methods that change the objects? https://stackoverflow.com/a/612196/865175

joshdholtz: @revolter Also...
>bang is also used for methods that raise an exception when the method without does not

`!` means dangerous. Modifying is dangerous but so is raising an exception 😊

revolter: Awesome, didn't know 👍

KrauseFx (approved):
👍

PR #12040, opened 2018-03-12
Coerce plugin output key to a string

philihp:
This line will fail when a plugin specifies its output as an array map of symbols to descriptions. What I usually see are plugins mapping string-to-strings, but since those are constants elsewhere, why not use them there too?

In other words, let the developer say

```ruby
module SharedValues
  SOME_VAR = :SOME_VAR
end

def self.output
  [
    [SharedValues::SOME_VAR, 'Some output variasble']
  ]
end
```

### Checklist

### Motivation and Context

I had a plugin, and it wouldn't work when I did a `bundle exec fastlane action my_action`, because of this:

```
https://github.com/fastlane/fastlane/issues/new
Run `fastlane env` to append the fastlane environment to your issue
bundler: failed to load command: fastlane (/Users/phil_busby/vendor/bundle/ruby/2.3.0/bin/fastlane)
NoMethodError: [!] undefined method `yellow' for :FOO:Symbol
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/documentation/actions_list.rb:149:in `block in print_output_variables'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/documentation/actions_list.rb:149:in `map'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/documentation/actions_list.rb:149:in `print_output_variables'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/documentation/actions_list.rb:61:in `show_details'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/documentation/actions_list.rb:6:in `run'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/commands_generator.rb:234:in `block (2 levels) in run'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/commander-fastlane-4.4.6/lib/commander/command.rb:178:in `call'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/commander-fastlane-4.4.6/lib/commander/command.rb:153:in `run'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/commander-fastlane-4.4.6/lib/commander/runner.rb:476:in `run_active_command'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb:75:in `run!'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/commander-fastlane-4.4.6/lib/commander/delegates.rb:15:in `run!'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/commands_generator.rb:333:in `run'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/commands_generator.rb:42:in `start'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/fastlane/lib/fastlane/cli_tools_distributor.rb:98:in `take_off'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/gems/fastlane-2.85.0/bin/fastlane:20:in `<top (required)>'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/bin/fastlane:23:in `load'
  /Users/phil_busby/vendor/bundle/ruby/2.3.0/bin/fastlane:23:in `<top (required)>'
```

### DIFF
--- a/fastlane/lib/fastlane/documentation/actions_list.rb
+++ b/fastlane/lib/fastlane/documentation/actions_list.rb
@@ -146,7 +146,7 @@ module Fastlane
       puts(Terminal::Table.new(
              title: "#{name} Output Variables".green,
              headings: ['Key', 'Description'],
-             rows: FastlaneCore::PrintTable.transform_output(output.map { |key, desc| [key.yellow, desc] })
+             rows: FastlaneCore::PrintTable.transform_output(output.map { |key, desc| [key.to_s.yellow, desc] })
       ))
       puts("Access the output values using `lane_context[SharedValues::VARIABLE_NAME]`")
       puts("")


joshdholtz:
@philihp Hey, thanks for the contribution! This is a small PR but I want to make sure this would be getting merged in for the correct reasons 😊 Is there a specific reason it won't work to define your `output` with strings as keys? This is method is for documentation purposes (to be read) so it makes sense for everything to be a string already.

The custom action template shows it as a string as well. 
```
      def self.output
        # Define the shared values you are going to provide
        # Example
        [
          ['[[NAME_UP]]_CUSTOM_VALUE', 'A description of what this value contains']
        ]
      end
```

Forcing these keys to be strings is best for quality as we know exactly what will be outputted otherwise if its a key it will be determined at runtime - which could be a it more fragile 😇

philihp:
@joshdholtz Of course. It seems to me that if a constant is defined, then we should use it... otherwise what's the point? Lemme take for example [gradle.rb](https://github.com/fastlane/fastlane/blob/c03a1ef64742796f80042d6147dfdb8d979cdcac/fastlane/lib/fastlane/actions/gradle.rb). It sets an output lane_context variable using the constant determined at runtime.

```ruby
Actions.lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH] = File.expand_path(last_apk_path) if last_apk_path
```
So it should report 
```ruby
def self.output
  [
    ['GRADLE_APK_OUTPUT_PATH', 'The path to the newly generated apk file. Undefined in a multi-variant assemble scenario']
  ]
end
```
should be
```ruby
def self.output
  [
    [SharedValues::GRADLE_APK_OUTPUT_PATH, 'The path to the newly generated apk file. Undefined in a multi-variant assemble scenario']
  ]
end
```

As far as I could see, *every* `SharedValues::FOO` defined to `:FOO`, so I don't think this adds much more cognitive load, but would also protect against the case where the reported string here actually falls out of sync with what the actual defined value is.

Additionally, someone reading the source would see this and know _hey, if i want to use this value, i need to do_ `lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]`, rather than `lane_context['GRADLE_APK_OUTPUT_PATH']`.

philihp:
Also, I'm down to go through all the actions and make that change... because consistency is also a good thing.

joshdholtz:
@philihp Thanks for taking the time to explain that! You really put up a good viewpoint 😊 However, I think we would really like to keep the keys of the hash a string as that field is meant to be static - since its for static documentation.

Our documentation for fastlane (docs.fastlane.tools) gets generated off of these static metadata action/plugin fields like 'output`, `return_value`, `authors`, etc). So for consistency and quality sake, I think we would really prefer to keep the `output` keys as a static string as well so we know _exacly_ what will be generated for documentation 🙃 

Your proposed changed isn't something _super_ dynamic but its dynamic enough where we might need to do some sort of validation when documentation gets generated to make sure that all keys generated to their proper string version.

Also, running `fastlane action gradle` shows out to use the output variable names already so stringing the key would add a little redundancy 😇

screen shot 2018-03-15 at 8 45 30 am

philihp:
I understand where you're coming from. Could I ask how the documentation is generated? Can I run that locally? I'd like to see how this method is used aside from printing to the console.

joshdholtz:
@philihp Turns out... it the docs don't actually look at `output` yet but I will be adding that tonight 😬 - https://github.com/fastlane/fastlane/blob/master/fastlane/lib/assets/ActionDetails.md.erb

joshdholtz:
Created issue #12061 to show that cause it really should

Issue #11578, opened 2018-01-16
FASTLANE_USER is ignored for app_store_build_number


revolter:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Running `bundle exec fastlane run app_store_build_number` asks for username, even if `FASTLANE_USER` is defined in `<root>/fastlane/.env`:

```
FASTLANE_USER=my.email@address.com
```

##### Complete output when running fastlane, including the stack trace and command used

`bundle exec fastlane --capture_output run app_store_build_number` silently fails with exit code `1`.

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.13.2                                                        |
| Ruby                        | 2.3.3                                                          |
| Bundler?                    | true                                                           |
| Git                         | git version 2.15.1                                             |
| Installation Source         | ~/.gems/bin/fastlane                                           |
| Host                        | Mac OS X 10.13.2 (17C88)                                       |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib |
| OpenSSL Version             | LibreSSL 2.2.7                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version               | 9.2                                                            |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     |             |   |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE |             |   |

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
| fastlane | 2.75.1  | ✅ Up-To-Date  |

### Loaded fastlane plugins:

| Plugin                 | Version | Update-Status |
| ---------------------- | ------- | ------------- |
| fastlane-plugin-aws_s3 | 1.1.1   | ✅ Up-To-Date  |

<details><summary><b>Loaded gems</b></summary>

[...]

PR #12047, opened 2018-03-13
Apply FASTLANE_DISABLE_COLORS globally (with bonus earlier dotenv loading)

### LINKED ISSUE #10498: Do not display title of PrintTable colored if disabled via environment variable
I disabled colored output by setting the environment variable `FASTLANE_DISABLE_COLORS` to `1`. Unfortunately, the title of `PrintTable` is still printed in green:

```
+-------------------+-------------------------+
|           [32mSummary for PEM 2.59.0[0m            |
+-------------------+-------------------------+
| team_id           | ***              |
| app_identifier    | de.***.app |
| pem_name          | ***                |
| development       | false                   |
| generate_p12      | true                    |
| active_days_limit | 30                      |
| force             | false                   |
| save_private_key  | true                    |
| username          | fastlane@***.com      |
| p12_password      | ********                |
| output_path       | certs                   |
+-------------------+-------------------------+
```

I believe that the fix is quite simple and should be located somewhere around this line: https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/print_table.rb#L34

joshdholtz:
Fixes #10498

## The Tale of the Unset Environment Variable
Written by: @joshdholtz

**fastlane community:** What is happening here?
**josh:** You see, `FASTLANE_DISABLE_COLORS` is supposed to disable all colors in the output but was only disabling the colors when `log` was being called for the first time.
**fastlane community:** Interesting
**josh:** Indeed! Now disable colors will get triggered as soon as the fastlane process starts so everything will be colorless.
**fastlane community:** But what does this have to do with `dotenv`? Why do you always have to bring `dotenv` into everything you do, Josh?
**josh:** That's a funny question, fastlane community. You see, I was working on an issue before this for `FASTLANE_HIDE_CHANGELOG` by running `fastlane lanes`. I was setting this environment variable in my `.env` file but it wasn't being loaded for some reason. I thought I was just being a dingus this morning.
**fastlane community:** ...
**josh:** Well, it turns out that `FASTLANE_HIDE_CHANGELOG` wasn't ever being loaded because 'fastlane lanes` never ran a lane which means my `.env. never got loaded. This was confusing to me as I was expecting my `.env` file to be applied.
**fastlane community:** ...
**josh:** Exactly! fastlane should load environment variables everywhere exactly like it does with lanes. And now it does!
**fastlane community:** 💃 🕺 🎊

### DIFF
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -56,6 +56,15 @@ module Fastlane
 
         FastlaneCore::UpdateChecker.start_looking_for_update('fastlane')
 
+        # Loading any .env files before any lanes are called since
+        # variables like FASTLANE_HIDE_CHANGELOG and FASTLANE_DISABLE_COLORS
+        # need to be set early on in execution
+        require 'fastlane/helper/dotenv_helper'
+        Fastlane::Helper::DotenvHelper.load_dot_env(nil)
+
+        # Disabling colors if environment variable set
+        require 'fastlane_core/ui/disable_colors' if FastlaneCore::Helper.colors_disabled?
+
         ARGV.unshift("spaceship") if ARGV.first == "spaceauth"
         tool_name = ARGV.first ? ARGV.first.downcase : nil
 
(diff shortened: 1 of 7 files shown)


taquitos (approved):
Very smart refactor, I like it 🎈:goat:

revolter:
@joshdholtz, Does this also fix https://github.com/fastlane/fastlane/issues/11578#issuecomment-362680123?

PR #12048, opened 2018-03-13
Add filename option for cert

jhanzo:
### Checklist
- [x] ~I've updated the documentation if necessary~ (not necessary?)

### Motivation and Context

As described in issue https://github.com/fastlane/fastlane/issues/8499, there is no option for adding a custom filename to generated certificate.

Based on `filename` option in `sigh`, I have added the same option in `cert`.

I test on my own : 
- [x] running `cert` with an env var named `CERT_FILE_NAME``
- [x] running `cert` in Fastfile **with** `filename` option
- [x] running `cert` in Fastfile **without** `filename` option
- [x] running `cert` in Fastfile with an incorrect `filename` option (without `.cer`)

### Description

Just add an option `filename` for cert command in the same way `sigh` action does.

### DIFF
--- a/cert/lib/cert/options.rb
+++ b/cert/lib/cert/options.rb
@@ -48,6 +48,12 @@ module Cert
                                      verify_block: proc do |value|
                                        ENV["FASTLANE_TEAM_NAME"] = value.to_s
                                      end),
+        FastlaneCore::ConfigItem.new(key: :filename,
+                                     short_option: "-q",
+                                     env_name: "CERT_FILE_NAME",
+                                     optional: true,
+                                     description: "The filename of certificate to store",
+                                     is_string: true),
         FastlaneCore::ConfigItem.new(key: :output_path,
                                      short_option: "-o",
                                      env_name: "CERT_OUTPUT_PATH",
--- a/cert/lib/cert/runner.rb
+++ b/cert/lib/cert/runner.rb
@@ -89,7 +89,7 @@ module Cert
           next
         end
 
-        path = store_certificate(certificate)
+        path = store_certificate(certificate, Cert.config[:filename])
         private_key_path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.p12"))
 
         if FastlaneCore::CertChecker.installed?(path)
@@ -170,7 +170,7 @@ module Cert
       private_key_path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.p12"))
       File.write(private_key_path, pkey)
 
-      cert_path = store_certificate(certificate)
+      cert_path = store_certificate(certificate, Cert.config[:filename])
 
       # Import all the things into the Keychain
       keychain = File.expand_path(Cert.config[:keychain_path])
@@ -187,8 +187,10 @@ module Cert
       return cert_path
     end
 
-    def store_certificate(certificate)
-      path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.cer"))
+    def store_certificate(certificate, filename = nil)
+      cert_name = filename ? filename : certificate.id
+      cert_name = "#{cert_name}.cer" unless File.extname(cert_name) == ".cer"
+      path = File.expand_path(File.join(Cert.config[:output_path], cert_name))
       raw_data = certificate.download_raw
       File.write(path, raw_data)
       return path
--- a/cert/spec/runner_spec.rb
+++ b/cert/spec/runner_spec.rb
@@ -1,3 +1,5 @@
+require 'tmpdir'
+
 describe Cert do
   describe Cert::Runner do
     before do
@@ -72,6 +74,59 @@ describe Cert do
 
       Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, keychain_path: ".")
       Cert::Runner.new.revoke_expired_certs!
+    end
+
+    describe ":filename option handling" do
+      filename = ""
+      let(:temp) { Dir.tmpdir }
+      let(:certificate) { stub_certificate }
+      let(:filepath) do
+        filename_ext = File.extname(filename) == ".cer" ? filename : "#{filename}.cer"
+        File.extname(filename) == ".cer" ? "#{temp}/#{filename}" : "#{temp}/#{filename}.cer"
+      end
+
+      before do
+        allow(Spaceship).to receive(:login).and_return(nil)
+        allow(Spaceship).to receive(:client).and_return("client")
+        allow(Spaceship).to receive(:select_team).and_return(nil)
+        allow(Spaceship.client).to receive(:in_house?).and_return(false)
+        allow(Spaceship.certificate.production).to receive(:all).and_return([certificate])
+        allow(FastlaneCore::CertChecker).to receive(:installed?).and_return(true)
+      end
+
+      let(:generate) do
+        options = { output_path: temp, filename: filename }
+        Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, options)
+        Cert::Runner.new.launch
+      end
+
+      context 'with filename flag' do
+        it "can forget the file extension" do
+          filename = "cert_name"
+          expect(File.exist?(filepath)).to be false
+          generate
+          expect(File.exist?(filepath)).to be true
+          File.delete(filepath)
+        end
+
+        it "can use the file extension" do
+          filename = "cert_name.cer"
+          expect(File.exist?(filepath)).to be false
+          generate
+          expect(File.exist?(filepath)).to be true
+          File.delete(filepath)
+        end
+      end
+
+      context 'without filename flag' do
+        it "can generate certificate" do
+          filename = "cert_name.cer"
+          expect(File.exist?(filepath)).to be false
+          generate
+          expect(File.exist?(filepath)).to be true
+          File.delete(filepath)
+        end
+      end
     end
   end
 end
--- a/cert/spec/commands_generator_spec.rb
+++ b/cert/spec/commands_generator_spec.rb
@@ -60,4 +60,17 @@ describe Cert::CommandsGenerator do
       expect(Cert.config[:output_path]).to eq('output/path')
     end
   end
+
+  describe ":filename option handling" do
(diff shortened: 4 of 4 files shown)


cert/lib/cert/options.rb
```diff
@@ -48,6 +48,15 @@ def self.available_options
                                      verify_block: proc do |value|
                                        ENV["FASTLANE_TEAM_NAME"] = value.to_s
                                      end),
+        FastlaneCore::ConfigItem.new(key: :filename,
+                                     short_option: "-q",
+                                     env_name: "CERT_FILE_NAME",
+                                     optional: true,
+                                     description: "The filename of certificate to store (must include .cer)",
```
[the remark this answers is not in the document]
KrauseFx: Instead of the `verify_block`

cert/lib/cert/runner.rb
```diff
@@ -187,8 +187,9 @@ def create_certificate
       return cert_path
     end
 
-    def store_certificate(certificate)
-      path = File.expand_path(File.join(Cert.config[:output_path], "#{certificate.id}.cer"))
+    def store_certificate(certificate, filename = nil)
+      cert_name = filename ? filename : "#{certificate.id}.cer"
```
KrauseFx: Ah we already auto-append, right?

cert/lib/cert/options.rb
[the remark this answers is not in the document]
jhanzo: Yes we could 👍 But for consistency I have kept arguments written in `sigh` options : [sigh#L102-L109](https://github.com/fastlane/fastlane/blob/master/sigh/lib/sigh/options.rb#L102-L109).

cert/lib/cert/runner.rb
jhanzo: There is no auto-append for now

cert/lib/cert/options.rb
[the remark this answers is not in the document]
jhanzo: But if we auto append file extension, we should keep the `verify_block` for checking filename doesn't contain a file extension, right ?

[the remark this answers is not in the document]
joshdholtz: @jhanzo I like your thought on consistency so if anything I would be down for also changing `sigh` in the future to auto append the extension if needs it (since its easier on end users for us to auto append what we are looking for)

In this case, I agree with @KrauseFx the we should move the `must include '.cer'` from the description and also remove the `verify_block` 👍 This will make this addition a lot more forgiving for new users 🙃

cert/lib/cert/runner.rb
joshdholtz: Here we need to choose a base filename and then append the `cer` extension (if not already on there) so I recommend a code change like the following...

```rb
cert_name = filename or certificate.id
cert_name = "#{cert_name}.cer" unless File.extname(base_name) == "cer"
```

⚠️ I wrote this code without testing it so it is subject to breakage 😇 But this should get you close to where we need to be

joshdholtz (changes requested):

cert/lib/cert/runner.rb
jhanzo: Of course ! Minor updates : 
```ruby
cert_name = filename ? filename : certificate.id
cert_name = "#{cert_name}.cer" unless File.extname(cert_name) == ".cer"

Comment (jhanzo) [reply cert/lib/cert/runner.rb]: FYI, I rebase commits for keeping a clean history

**Comment by jhanzo** [comment]
FYI, I rebase commits for keeping a clean history

**Review by joshdholtz — CHANGES_REQUESTED** [review]
@jhanzo These code changes look good 👍 Are you able to write some tests to verify that the cert filename gets created correctly with and without this new `filename` option?

File: cert/spec/runner_spec.rb [hunk 3]
```diff
@@ -73,5 +73,48 @@
...
+        certificate = stub_certificate
+
+        allow(Spaceship).to receive(:login).and_return(nil)
+        allow(Spaceship).to receive(:client).and_return("client")
+        allow(Spaceship).to receive(:select_team).and_return(nil)
+        allow(Spaceship.client).to receive(:in_house?).and_return(false)
+        allow(Spaceship.certificate.production).to receive(:all).and_return([certificate])
+        allow(FastlaneCore::CertChecker).to receive(:installed?).and_return(true)
+
+        expect(File.exist?(filename_ext)).to be false
```
joshdholtz: Are you able to remove the `expect`s from this `let(:generate)` and move them into the `its` below? the `let` should really only be for creating, not testing 😊 

Other than that this looks really good 😊

joshdholtz (changes requested):

cert/spec/runner_spec.rb
jhanzo: OoOoh I thought that you will tell me that 😁I did it for being identical to [fastlane/certs/commands_generator_spec#L4-L8](https://github.com/fastlane/fastlane/blob/master/cert/spec/commands_generator_spec.rb#L4-L8).

I have updated code, have a look 😉 FYI, I had to move an extra method for avoiding to many code duplicates.

Commits rebased : 🆗

cert/spec/runner_spec.rb
```diff
@@ -73,5 +73,55 @@
...
+        Cert::Runner.new.launch
+      end
+
+      context 'with filename flag' do
+        it "can forget the file extension" do
+          filename = "cert_name"
+          expect(File.exist?(filename_ext)).to be false
+          generate
+          expect(File.exist?(filename_ext)).to be true
+          File.delete(filename_ext)
```
joshdholtz: One last question! Where do these files get created (and then deleted) from right now? Will these get generated **in** the specs folder? If so, can we tell cert to create them in a temp directory?

jhanzo: Files are created where command is executed. 

Thus I can create a dedicated folder for this, but how should I manage it? Should I create and delete it each time a test is done? 

Moreover, maybe a global rule would be relevant for having all specs run into a tmp folder? this would be consistent and reusable.

How do you feel about it ?

joshdholtz: `cert` takes an option for `path` so I think pass a `Dir.tmpdir()` (or similar) into the `Cert.config`? 😊

joshdholtz: The reason I think this should be a temporary directory because right now, if the test fails, that file won't be deleted and will remain around

jhanzo:
Before rebasing, please let me know if changes are okay for you !

ie : https://github.com/fastlane/fastlane/pull/12048/commits/624790832694395e30e700ad9313ef5b7c8965ea

jhanzo:
@joshdholtz 

Before rebasing, please let me know if changes are okay for you ! 😉 

ie : https://github.com/fastlane/fastlane/pull/12048/commits/624790832694395e30e700ad9313ef5b7c8965ea

cert/spec/runner_spec.rb
```diff
@@ -73,5 +75,58 @@
       Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, keychain_path: ".")
       Cert::Runner.new.revoke_expired_certs!
     end
+
+    describe ":filename option handling" do
+      filename = ""
```
joshdholtz: This all looks muuuccchhhh better 🙌 The one last thing I would recommend is staying away from putting a global-ish variable like this that gets mutated by the tests. I would probably attempt the following...

```rb
let (:temp) { Dir.tmpdir }
let (:certificate) { stub_certificate }
before do
  allow(Spaceship).to receive(:login).and_return(nil)
  allow(Spaceship).to receive(:client).and_return("client")
  allow(Spaceship).to receive(:select_team).and_return(nil)
  allow(Spaceship.client).to receive(:in_house?).and_return(false)
  allow(Spaceship.certificate.production).to receive(:all).and_return([certificate])
  allow(FastlaneCore::CertChecker).to receive(:installed?).and_return(true)
end

def make_filepath(filename)
  filename_ext = File.extname(filename) == ".cer" ? filename : "#{filename}.cer"
  File.extname(filename) == ".cer" ? "#{temp}/#{filename}" : "#{temp}/#{filename}.cer"
end

context 'with filename flag' do
  it "can forget the file extension" do
    filename = "cert_name"
    filepath = make_filepath(filename)
    expect(File.exist?(filepath)).to be false

    options = { output_path: temp, filename: filename }
    Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, options)
    Cert::Runner.new.launch

    expect(File.exist?(filepath)).to be true
    File.delete(filepath)
  end
end
```

This does a few things...
1) `generate` right now are being used as methods instead of variables (they should really be used as variables. By moving the "allows" to a `before` block and moving the `certificate` to a "let", we can get rid of `generate`
2) 'filepath` is also being used as a method. We can actually define this as method and pass a filename to it to generate the `filepath` which allows us to get rid of a global-ish `filename` variable that we are mutating

Let me know if this doesn't work 🙃

joshdholtz (changes requested):

cert/spec/runner_spec.rb
jhanzo: I am agree with you, but about your first note, you delete `generate` method but it implies to duplicate "Runner" code as many times you need to test, are you sure ? A convenient method named "generate" would do this job : 
```ruby
      let(:generate) do
        options = { output_path: temp, filename: filename }
        Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, options)
        Cert::Runner.new.launch
      end

Comment (joshdholtz) [reply cert/spec/runner_spec.rb]: @jhanzo I'm okay with a little bit of duplicate runner code 😊 Its okay, IMO, for tests to have some duplicate code. You could create a method instead of doing `let(:generate)` if you want...
```rb
def generate(output_path, filename)
  options = { output_path: temp, filename: filename }
  Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, options)
   Cert::Runner.new.launch
end
```
but this isn't totally necessary 😇

jhanzo:
@joshdholtz done

joshdholtz (approved):
This is great! Thanks for going through all my requests 😊 ❤️

PR #12049, opened 2018-03-13
Proposal to remove and/or from conditionals with Mr. Rubocop 🚨

joshdholtz:
## Why
Ruby style guide explains how `&&`, `||`, `and`, and `or` should be used - https://github.com/bbatsov/ruby-style-guide#no-and-or-or

>The and and or keywords are banned. The minimal added readability is just not worth the high probability of introducing subtle bugs. For boolean expressions, always use && and || instead. For flow control, use if and unless; && and || are also acceptable but less clear. 

## What
`rubocop` changed about 160ish instances of `and` and `or` used for boolean expressions. This change **does** touch many places of the codebase but **shouldn't** have any side effects (unless there were side effects from `and` and `or` that we were somehow depending on)

### DIFF
--- a/fastlane/lib/fastlane/swift_lane_manager.rb
+++ b/fastlane/lib/fastlane/swift_lane_manager.rb
@@ -7,8 +7,8 @@ module Fastlane
     # @param parameters [Hash] The parameters passed from the command line to the lane
     # @param env Dot Env Information
     def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false)
-      UI.user_error!("lane must be a string") unless lane.kind_of?(String) or lane.nil?
-      UI.user_error!("parameters must be a hash") unless parameters.kind_of?(Hash) or parameters.nil?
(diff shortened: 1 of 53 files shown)


fastlane/lib/fastlane/swift_lane_manager.rb
```diff
@@ -7,8 +7,8 @@ class SwiftLaneManager < LaneManagerBase
     # @param parameters [Hash] The parameters passed from the command line to the lane
     # @param env Dot Env Information
     def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false)
-      UI.user_error!("lane must be a string") unless lane.kind_of?(String) or lane.nil?
```
KrauseFx: uhh, did this ever work?

KrauseFx (approved):
I'm very much for this change, we'll add the same rule to fastlane.ci 👍 

Did you manually migrate it or did rubocop do it? Can we be sure this doesn't break anything?

joshdholtz:
@KrauseFx Rubocop did all the work. I just enjoyed its success 😊

PR #12052, opened 2018-03-13
Fix plugin scores updating crash

revolter:
I don't know exactly how to reproduce it, but sometimes it crashed with
"wrong number of arguments (given X, expected 1)". Using flatten seems
to fix it and looks safer and cleaner.


joshdholtz (approved):
This looks much better 💪

PR #12059, opened 2018-03-15
[spaceship] Raise a meaningful error if the device name is too long

thii:
Apple Developer Portal does not allow to register a device with name
longer than 50 characters, but `spaceship` is currently returning an
unknown error if we try to do that.

### DIFF
--- a/spaceship/lib/spaceship/portal/device.rb
+++ b/spaceship/lib/spaceship/portal/device.rb
@@ -148,7 +148,7 @@ module Spaceship
             raise "You cannot create a device without a device_id (UDID) and name"
           end
 
-          raise "Device name must be 50 characters or less. \"#{name}\" has a #{name.length}-character length." if name.length > 50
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/portal/device.rb
```diff
@@ -148,6 +148,8 @@ def create!(name: nil, udid: nil, mac: false)
             raise "You cannot create a device without a device_id (UDID) and name"
           end
 
+          raise "Device name must be 50 characters or less. \"#{name}\" has a #{name.length}-character length." if name.length > 50
```
joshdholtz: This is a bit nit picky but can we change `#{name.length}-character` to `#{name.length} character` 🙃

joshdholtz (changes requested):
Small nit pick but this is great 😊

joshdholtz (approved):
Thanks for the contribution!

PR #12062, opened 2018-03-15
Add output variables to match with profiles names and UUIDs

nicolasbraun:
### Motivation and Context
On some boilerplate projects the bundle id of the apps and extensions might be changed dynamically.
For this reason it is not possible to hardcode the value avec the profile specifier using this technique https://docs.fastlane.tools/codesigning/xcode-project/#xcode-7-and-lower so we use a generic evar like `$PROFILE_UUID `and setup it's value to `ENV['SIGH_NAME']`.

Since this evar is note set up when `match` uses a local profile, we currently have to force a `sigh` run (using the `force` option of `match`).

### Description
This PR simply adds output variables to `match`: collections with the UUID / name for each profile generated (since match can generate multiple profiles at once).

### Side notes
I first intented to use evar since the lane_context was not available in the `match` action but since there might be multiple values an array seemed better than a coma-separated string.

Sigh sets both
https://github.com/fastlane/fastlane/blob/055f325725ac58d006ba46f0d85c6bb74e98adbc/fastlane/lib/fastlane/actions/get_provisioning_profile.rb#L25
Is there any good practice here? 

### Test

`bundle exec fastlane test` fails on my laptop but it also does on master so I suspect I might have a local issue

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -2,7 +2,6 @@ require 'fastlane_core/cert_checker'
 require 'fastlane_core/provisioning_profile'
 require 'fastlane_core/print_table'
 require 'spaceship/client'
-require 'fastlane/actions/lane_context'
 require_relative 'generator'
 require_relative 'git_helper'
 require_relative 'module'
@@ -172,8 +171,6 @@ module Match
       installed_profile = FastlaneCore::ProvisioningProfile.install(profile, keychain_path)
       parsed = FastlaneCore::ProvisioningProfile.parse(profile, keychain_path)
       uuid = parsed["UUID"]
-      Fastlane::Actions.lane_context[:MATCH_PROFILES_UUID] = Fastlane::Actions.lane_context[:MATCH_PROFILES_UUID].to_a << uuid
-      Fastlane::Actions.lane_context[:MATCH_PROFILES_NAME] = Fastlane::Actions.lane_context[:MATCH_PROFILES_NAME].to_a << parsed["Name"]
 
       if spaceship && !spaceship.profile_exists(username: params[:username], uuid: uuid)
         # This profile is invalid, let's remove the local file and generate a new one
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -2,6 +2,7 @@ require 'fastlane_core/cert_checker'
 require 'fastlane_core/provisioning_profile'
 require 'fastlane_core/print_table'
 require 'spaceship/client'
+require 'fastlane/actions/lane_context'
 require_relative 'generator'
 require_relative 'git_helper'
 require_relative 'module'
@@ -171,6 +172,8 @@ module Match
       installed_profile = FastlaneCore::ProvisioningProfile.install(profile, keychain_path)
       parsed = FastlaneCore::ProvisioningProfile.parse(profile, keychain_path)
       uuid = parsed["UUID"]
+      Fastlane::Actions.lane_context[:MATCH_PROFILES_UUID] = Fastlane::Actions.lane_context[:MATCH_PROFILES_UUID].to_a << uuid
+      Fastlane::Actions.lane_context[:MATCH_PROFILES_NAME] = Fastlane::Actions.lane_context[:MATCH_PROFILES_NAME].to_a << parsed["Name"]
 
       if spaceship && !spaceship.profile_exists(username: params[:username], uuid: uuid)
         # This profile is invalid, let's remove the local file and generate a new one
--- a/fastlane/lib/fastlane/actions/sync_code_signing.rb
+++ b/fastlane/lib/fastlane/actions/sync_code_signing.rb
@@ -2,6 +2,8 @@ module Fastlane
   module Actions
     module SharedValues
       MATCH_PROVISIONING_PROFILE_MAPPING = :MATCH_PROVISIONING_PROFILE_MAPPING
+      MATCH_PROFILES_UUIDS = :MATCH_PROFILES_UUIDS
+      MATCH_PROFILES_NAMES = :MATCH_PROFILES_NAMES
     end
 
     class SyncCodeSigningAction < Action
@@ -13,6 +15,7 @@ module Fastlane
 
         define_profile_type(params)
         define_provisioning_profile_mapping(params)
+        define_shared_values(params)
       end
 
       def self.define_profile_type(params)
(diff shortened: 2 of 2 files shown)


nicolasbraun:
Hey @joshdholtz 

So this is the PR we talked about on Slack.
This is a bit weird since locally my rubocop run fails but with less warning and one offense that I also have on master (7555c12f404d0932d947658dbaa235c9b8f1b64f)


On master
```
➜  fastlane git:(master) bundle exec rubocop -D
Inspecting 1013 files
....................................................................................................C................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Offenses:

match/lib/match/commands_generator.rb:46:11: C: Require/MissingRequireStatement: Match::Runner not found, you're probably missing a require statement or there is a cycle in your dependencies.
          Match::Runner.new.run(params)
          ^^^^^^^^^^^^^

1013 files inspected, 1 offense detected
```

On my branch 
```
➜  fastlane git:(match_name_and_uuid_lane_context) bundle exec rubocop -D                        
Inspecting 1013 files
....................................................................................................C................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Offenses:

match/lib/match/commands_generator.rb:46:11: C: Require/MissingRequireStatement: Match::Runner not found, you're probably missing a require statement or there is a cycle in your dependencies.
          Match::Runner.new.run(params)
          ^^^^^^^^^^^^^

1013 files inspected, 1 offense detected```

**Comment by nicolasbraun** [comment]
Even weirder when running bundle exec fastlane test, i have no violations

```
[...]
[01:03:48]: grep -RE ' UI\.' '/Users/nicolasbraun/Sources/fastlane/credentials_manager/lib' --include=\*.rb --exclude-dir .bundle
[01:03:48]: No debug code found in code base 🐛
[01:03:48]: ---------------------
[01:03:48]: --- Step: rubocop ---
[01:03:48]: ---------------------
[01:03:48]: $ bundle exec rubocop -D
[01:03:50]: ▸ Inspecting 1013 files
[01:03:50]: ▸ .....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
[01:03:50]: ▸ 1013 files inspected, no offenses detected
[01:03:50]: -----------------------------------------------
```

**Comment by joshdholtz** [comment]
@nicolasbraun What version of Ruby are you running? I think the lint tests use `2.3.0` 🙃

**Comment by nicolasbraun** [comment]
`2.4.1`, i'll try to run the tests in 2.3.0 during the weekend.

Still, I have no idea how to fix this, any input @joshdholtz ?

Thanks 🙏

**Comment by nicolasbraun** [comment]
@joshdholtz  With ruby 2.3.0 I have one offense but also have it on master :/
Still do you have any suggestion on how to expose the lane_context in match ? not sure this is the right way

``` 3 files changed, 44 insertions(+), 15 deletions(-)
➜  fastlane git:(master) bundle exec rubocop -D
Inspecting 1013 files
....................................................................................................C................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Offenses:

match/lib/match/commands_generator.rb:46:11: C: Require/MissingRequireStatement: Match::Runner not found, you're probably missing a require statement or there is a cycle in your dependencies.
          Match::Runner.new.run(params)
          ^^^^^^^^^^^^^

1013 files inspected, 1 offense detected
➜  fastlane git:(master) ruby -v
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin17]
➜  fastlane git:(master) git checkout match_name_and_uuid_lane_context  
Switched to branch 'match_name_and_uuid_lane_context'
Your branch is up-to-date with 'origin/match_name_and_uuid_lane_context'.
➜  fastlane git:(match_name_and_uuid_lane_context) bundle exec rubocop -D                        
Inspecting 1013 files
....................................................................................................C................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Offenses:

match/lib/match/commands_generator.rb:46:11: C: Require/MissingRequireStatement: Match::Runner not found, you're probably missing a require statement or there is a cycle in your dependencies.
          Match::Runner.new.run(params)
          ^^^^^^^^^^^^^

1013 files inspected, 1 offense detected
➜  fastlane git:(match_name_and_uuid_lane_context) ```

File: match/lib/match/runner.rb [hunk 1]
```diff
@@ -171,6 +172,8 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
       installed_profile = FastlaneCore::ProvisioningProfile.install(profile, keychain_path)
       parsed = FastlaneCore::ProvisioningProfile.parse(profile, keychain_path)
       uuid = parsed["UUID"]
+      Fastlane::Actions.lane_context[:MATCH_PROFILES_UUID] = Fastlane::Actions.lane_context[:MATCH_PROFILES_UUID].to_a << uuid
+      Fastlane::Actions.lane_context[:MATCH_PROFILES_NAME] = Fastlane::Actions.lane_context[:MATCH_PROFILES_NAME].to_a << parsed["Name"]
```
joshdholtz: `Fastlane::Actions.lane_context` shouldn't get set in the `match/running.rb`. We should find a way to bring these values outside of runner so they can be set in the match action instead. I would start with looking how other actions/tools do this 😊 This should get rid of your rubocop error as well.

joshdholtz (changes requested):

match/lib/match/runner.rb
nicolasbraun: @joshdholtz I think i got it, i pushed changes. Thanks a lot 🙏

joshdholtz (commented):
@nicolasbraun Thanks for making that change! This implementation looks much better 👌 Thoughts on making this a single map looks something like...
```rb
{
  "com.bundle.id1: {
    "uuid": "1234",
    "name": "thename.provisioninprofile"
  },
     "com.bundle.id2: {
    "uuid": "1234",
    "name": "thename.provisioninprofile"
  }
}
```

I think something like that may make it a little more usable? ^ Thoughts?

joshdholtz:
Waaaiiiittt... follow up to my review... Isn't my thought pretty much what `MATCH_PROVISIONING_PROFILE_MAPPING` creates? 🤔

nicolasbraun:
@joshdholtz oh yes, you are right I could totally use the `MATCH_PROVISIONING_PROFILE_MAPPING` and search for the profile name for my current identifier.

I think passing the uuid to Xcode might be a little bit more robust (since you can have many profile with the same name).
According to `xcodebuild` doc it should work, any thoughts?

```
	provisioningProfiles : Dictionary

		For manual signing only. Specify the provisioning profile to use for each executable in your app. Keys in this dictionary are the bundle identifiers of executables; values are the provisioning profile name or UUID to use.
```

joshdholtz:
@nicolasbraun The name is actually the preferred way these days. Passing in the UUID is deprecated 😬  So making use of the `MATCH_PROVISIONING_PROFILE_MAPPING` might be the way to go which makes these changes probably not needed 🙃

PR #12063, opened 2018-03-15
Add git diff option to ensure_git_status_clean

nicolasbraun:
### Motivation and Context
Some people do not have full access on our CI system so displaying the precisely why the repo is dirty helps

### Description
Simple add an option to ensure_git_status_clean to display full git diff


joshdholtz:
@nicolasbraun Are you able to add testes to this so we can both make sure it works and test/verify what the output of new error message will look like? 😊

nicolasbraun:
@joshdholtz  done 🚀

joshdholtz (approved):

PR #12064, opened 2018-03-16
Add the git_submodule_update action

nicolasbraun:
### Description

Add an action to handle git submodule updates


joshdholtz (changes requested):
Can we get some tests with this action? 😊 It may not seem like it adds a lot of value for a small action but it makes for strong confidence if/when somebody else would come in an add features or debug this action

nicolasbraun:
@joshdholtz  done 🚀

joshdholtz (approved):

PR #12065, opened 2018-03-16
Allow actions with zero options and properly generate JSON for payload

### LINKED ISSUE #12043: Adding last_git_commit to Fastlane Swift
Hey guys, love fastlane and don't know what I would do without it.

Recently I have been trying to convert our current fastlane setup into Swift and I ran across a missing command that I would like to add `last_git_commit` to `Fastlane.swift`

Let me know what you think!

### New Issue Checklist

### Issue Description
I want to add the command last_git_commit that is missing from Fastlane swift.

```
@discardableResult func lastGitCommit() -> [String : String] {
    let command = RubyCommand(commandID: "", methodName: "last_git_commit", className: nil, args: [])
    return parseDictionary(fromString: runner.executeCommand(command))
}
```

joshdholtz:
Fixes #12043

# What was the problem?
- Actions that has zero options were not getting swifty
- JSON was unsafely being generated 
  - Was erring out when payload objects
  - Replaced with `JSON.generate` 

```sh
[20:29:53]: -----------------------------
[20:29:53]: --- Step: last_git_commit ---
[20:29:53]: -----------------------------
[20:29:53]: Commit: ["abbreviated_commit_hash": "4e8505b", "message": "Heyyy\n", "author": "Josh Holtz", "commit_hash": "4e8505bd660824458feb81ba079ba5271c043f94"]
```

### DIFF
--- a/fastlane/lib/fastlane/server/socket_server.rb
+++ b/fastlane/lib/fastlane/server/socket_server.rb
@@ -194,11 +194,18 @@ module Fastlane
           return_value: closure_arg,
           return_value_type: :string # always assume string for closure error_callback
         )
-        closure_arg = ', "closure_argument_value": ' + closure_arg
       end
 
       Thread.current[:exception] = nil
-      return '{"payload":{"status":"ready_for_next", "return_object":' + return_object + closure_arg + '}}'
+
+      payload = {
+        payload: {
+          status: "ready_for_next",
+          return_object: return_object,
+          closure_argument_value: closure_arg
+        }
+      }
+      return JSON.generate(payload)
     rescue StandardError => e
       Thread.current[:exception] = e
 
@@ -210,7 +217,14 @@ module Fastlane
         exception_array << "cause: #{e.class}"
         exception_array << e.backtrace
       end
-      return "{\"payload\":{\"status\":\"failure\",\"failure_information\":#{exception_array.flatten}}}"
+
+      payload = {
+        payload: {
+          status: "failure",
+          failure_information: exception_array.flatten
+        }
+      }
+      return JSON.generate(payload)
     end
   end
 end
--- a/fastlane/lib/fastlane/swift_fastlane_api_generator.rb
+++ b/fastlane/lib/fastlane/swift_fastlane_api_generator.rb
@@ -321,10 +321,7 @@ func parseInt(fromString: String, function: String = #function) -> Int {
     end
 
     def process_action(action: nil)
-      unless action.available_options
-        return nil
-      end
-      options = action.available_options
+      options = action.available_options or []
 
       action_name = action.action_name
       keys = []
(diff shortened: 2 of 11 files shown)


fastlane/lib/fastlane/server/socket_server.rb
```diff
@@ -194,11 +194,18 @@ def execute_action_command(command: nil)
...
-      return '{"payload":{"status":"ready_for_next", "return_object":' + return_object + closure_arg + '}}'
+
+      payload = {
+        payload: {
+          status: "ready_for_next",
+          return_object: return_object,
+          closure_argument_value: closure_arg
+        }
+      }
+      return JSON.generate(payload)
```
taquitos: 🌀I was meaning to do this when things got complicated enough...

fastlane/lib/fastlane/swift_fastlane_api_generator.rb
```diff
@@ -321,10 +321,7 @@ def ignore_param?(function_name: nil, param_name: nil)
     end
 
     def process_action(action: nil)
-      unless action.available_options
-        return nil
-      end
-      options = action.available_options
+      options = action.available_options or []
```
taquitos: OR? lol.

taquitos:
Make sure you leave out all the autogenerated swift files
Those will get regenerated during deployment 🐐

fastlane/swift/Gymfile.swift
```diff
@@ -18,4 +18,4 @@ class Gymfile: GymfileProtocol {
 
 
 
-// Generated with fastlane 2.85.0
```
taquitos: Weird that it is stripping out the version entirely, let's see what the `version bump` pr does with it. Might be a bug somewhere. I think I've seen this before

taquitos (approved):
Great stuff here 🎈:goat:

I forgot to mention, if you didn't already figure it out, you can put `fastlane` into a socket server mode where it will accept connections and automatically reset itself when a connection drops. That way you can actually debug stuff by running the FastlaneRunner from xcode.

fastlane/lib/fastlane/swift_fastlane_api_generator.rb
joshdholtz: Yeah yeah! `or` is totally okay when its not in a conditional I learned 🙃 That is the only case when its supposed to be used... when its not in boolean logic 😉

taquitos: In this case, it might be clearer to just use `||` thoughts?
I think since it's so uncommon to see `or` it might be better to use the OG `||`

fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -214,7 +214,6 @@ def return_statement
 
     def implementation
       args = build_argument_list
-
```
taquitos: Eh. I liked this extra line. But #yolo

taquitos (approved):
One comment about `or` otherwise 🐐it!

Issue #10424, opened 2017-09-26
Provided xcconfig is not used for provisioning profile selection


pkostenko:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

If xcconfig is provided from the command line, it's not being used during the provisioning profiles search.

##### Complete output when running fastlane, including the stack trace and command used

`fastlane gym --export_method "enterprise" --xcconfig "Config.xcconfig" MyProject.xcproj`

```
DEBUG [2017-09-27 00:27:12.27]: Parsing project file 'MyProject.xcproj' to find selected provisioning profiles
DEBUG [2017-09-27 00:27:12.27]: Finding provision profiles for 'Release'
DEBUG [2017-09-27 00:27:12.63]: Merging provisioning profile mappings
DEBUG [2017-09-27 00:27:12.63]: -------------------------------------
DEBUG [2017-09-27 00:27:12.63]: Primary provisioning profile mapping:
DEBUG [2017-09-27 00:27:12.63]: {}
DEBUG [2017-09-27 00:27:12.63]: Secondary provisioning profile mapping:
DEBUG [2017-09-27 00:27:12.63]: {:"<REDACTED_BUNDLE_ID>"=>"<REDACTED_PROFILE_NAME>"}
DEBUG [2017-09-27 00:27:12.63]: Resulting in the following mapping:
DEBUG [2017-09-27 00:27:12.63]: {:"<REDACTED_BUNDLE_ID>"=>"<REDACTED_PROFILE_NAME>"}
INFO [2017-09-27 00:27:12.63]: Detected provisioning profile mapping: {:"<REDACTED_BUNDLE_ID>"=>"<REDACTED_PROFILE_NAME>"}
```
Where <REDACTED_BUNDLE_ID> and <REDACTED_PROFILE_NAME> are being the ones set in the project and not in the provided xcconfig file.

The contents of the xcconfig:

```
PRODUCT_BUNDLE_IDENTIFIER = <MyBundleId>
PROVISIONING_PROFILE_SPECIFIER = <MyProfileName>
```

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
[...]

PR #12067, opened 2018-03-16
Take xcconfig option into account when generating showBuildSettings command

chrispcampbell:
### Checklist

### Motivation and Context

As described in #10424 `gym` takes the xcconfig option into account when generating the main `xcodebuild` command, but it doesn't take it into account when generating the preliminary `xcodebuild -showBuildSettings` command, which causes various problems later.

I verified this fix in my own project (tested the fix against `scan` and `gym` specifically).  Also ran all the automated tests.

### Description

I started by finding that `build_xcodebuild_showbuildsettings_command` uses the shared `xcodebuild_parameters` function, and that function didn't previously take `xcconfig` into account, so I added it there.

Then I grep'd and found that a few other commands (`scan`, `gym`, and `snapshot`) were using `xcodebuild_parameters` as well.  `scan` and `gym` were previously manually handling the `xcconfig` option, so now that I added that to `xcodebuild_parameters` I could remove that per-action handling.

`snapshot` on the other hand previously didn't list `xcconfig` as a supported option, so the tests actually failed.  I see no reason why `xcconfig` shouldn't be supported for `snapshot` as it is for the other commands, so I added it as a newly supported option, and now all the tests are passing.

I also added a simple unit test to verify that the correct `showBuildSettings` command is generated when `xcconfig` is included in the options.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -309,6 +309,7 @@ module FastlaneCore
       proj << "-scheme #{options[:scheme].shellescape}" if options[:scheme]
       proj << "-project #{options[:project].shellescape}" if options[:project]
       proj << "-configuration #{options[:configuration].shellescape}" if options[:configuration]
+      proj << "-xcconfig #{options[:xcconfig].shellescape}" if options[:xcconfig]
 
       return proj
     end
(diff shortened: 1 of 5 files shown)


chrispcampbell:
I signed it!

joshdholtz (approved):
This is a really good find 🙌 Also, thank you so much for that PR description that you provided. It made it very easy to follow why you did what you did 😊 

This looks good to me 👍

Issue #12088, opened 2018-03-19
update_info_plist fails when plist file extension is different


RishabhTayal:
As of now the `update_info_plist` action will fail [here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/update_info_plist.rb#L87) if the extension of the file passed in `plist_path`  parameter is anything other than '.plist'. I believe we should be able to support other file extensions as well for e.g. `.entitlements` since `.entitlements` is also a plist file.

I would be more than happy to open a PR to make this change.

elitalon:
@RishabhTayal I don't know what the core team thinks about this, but the interface of the action assumes you're  actually editing an "Info.plist" file, rather than some generic `.plist` file. Allowing other extensions in the same action would change the semantics.

Maybe it makes sense to add some generic `update_plist` or an specific `update_entitlements` action. Which in turn could be used internally to refactor [update_app_group_identifiers](https://docs.fastlane.tools/actions/update_app_group_identifiers) and [update_icloud_container_identifiers](https://docs.fastlane.tools/actions/update_icloud_container_identifiers).

RishabhTayal:
Closing in favor of https://github.com/fastlane/fastlane/issues/12188

PR #12091, opened 2018-03-19
Disabled .plist extension verification

RishabhTayal:
Fixes: #12088

### DIFF
--- a/fastlane/lib/fastlane/actions/update_info_plist.rb
+++ b/fastlane/lib/fastlane/actions/update_info_plist.rb
@@ -82,10 +82,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :plist_path,
                                        env_name: "FL_UPDATE_PLIST_PATH",
                                        description: "Path to info plist",
-                                       optional: true,
-                                       verify_block: proc do |value|
-                                         UI.user_error!("Invalid plist file") unless value[-6..-1].casecmp(".plist").zero?
-                                       end),
+                                       optional: true),
           FastlaneCore::ConfigItem.new(key: :scheme,
                                        env_name: "FL_UPDATE_PLIST_APP_SCHEME",
                                        description: "Scheme of info plist",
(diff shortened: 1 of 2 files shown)


joshdholtz (changes requested):
Can you add tests around handling when this path contains invalid plist content? 🙃

RishabhTayal:
@joshdholtz test case is added. Ready for review again. 👍

joshdholtz:
After thinking about this comment - https://github.com/fastlane/fastlane/issues/12088#issuecomment-374511298 -  I think that the name of this action _may_ be a bit misleading now 🤔 

I **think** it makes the most sense to pull this logic out into a helper and then...
- Update `update_info_plist` to use the helper
- Update `update_app_group_identifiers` to use the helper
- Update `update_icloud_container_identifers` to use the helper
- Add new action `update_entitlements`to use the helper (what this PR is really for)

^ This will make it easy for users to look for things they want to do while using shared plist code

This is probably a decent amount of work so I can take over from here if you'd like 🙃

RishabhTayal:
@joshdholtz I don't think we should restrict the file extensions to be only `.plist`. The `.entitlements` file is a valid plist file so I don't see a reason why we would restrict the file extensions.

joshdholtz:
@RishabhTayal I don't think we should either but the name of this action is `update_info_plist` and entitlements isn't an "Info.plist" which is where things start to get weird

RishabhTayal:
I agree too. I know you are not going to like hearing this but how about we change it to `update_plist`.

RishabhTayal:
@joshdholtz Where do we stand with the decision on what to do? Are we creating a new helper function and new action for this?

joshdholtz:
@RishabhTayal We can't change this action _specifically_ to be `update_plist` since that would break soooooooo many other systems. I would _could_ create a new action called `update_plist` which `update_info_plist` would use. I think that would be okay 👍

RishabhTayal:
@joshdholtz Sounds like a plan to me. Do you think this is something I could take, If it's not too complicated?

RishabhTayal:
You may close this PR now.

joshdholtz:
@RishabhTayal I would be down for a PR for this if you have time 🙃 If not, I totally understand 😉

PR #9742, opened 2017-07-13
[gym] Refactor and improve build failure output when using an incorrect export_method in gym

KrauseFx:
- Automatically detect mismatches between the provided `export_method` and the selected provisioning profiles
- On build failure, show `export_method` and warn about using :development
- Don't use `:development` export type when running match before _gym_

----

**Screenshot 1:**
- Shows the `export_method` in the summary
- Show that `development` was defined as `export_method` which almost never makes sense
screen shot 2017-07-13 at 4 55 27 pm

**Screenshot 2**
- Show information about a mismatch of `export_method` and provisioning profiles for each app target
screen shot 2017-07-13 at 5 31 32 pm

PR #12092, opened 2018-03-19
Prompt user  instead of not setting the export_method when codesign env = developmen

nicolasbraun:
### Motivation and Context

Currently fastlane code does not set export_method when codesign = development
https://github.com/fastlane/fastlane/blob/f4e516c3a6f4a17088bcd099feb02ffa1a6fe9c7/fastlane/lib/fastlane/actions/build_ios_app.rb#L12

It comes from an old commit from @KrauseFx https://github.com/fastlane/fastlane/pull/9742 that assumes exporting using gym in development is usually not a legit usecase.

Currently when using match(type: development), gym tries to exports as app-store and fails.
Actually this use case is legit. For example when testing recurring In App Purchases. 

### Description

The export_method is correctly set and if the UI is interactive we prompt the user for confirmation since it might be a configuration error.

PS: Not even sure that prompt is legit

### DIFF
--- a/fastlane/lib/fastlane/actions/build_ios_app.rb
+++ b/fastlane/lib/fastlane/actions/build_ios_app.rb
@@ -9,9 +9,12 @@ module Fastlane
       def self.run(values)
         require 'gym'
 
-        unless Actions.lane_context[SharedValues::SIGH_PROFILE_TYPE].to_s == "development"
-          values[:export_method] ||= Actions.lane_context[SharedValues::SIGH_PROFILE_TYPE]
+        if Actions.lane_context[SharedValues::SIGH_PROFILE_TYPE].to_s == "development" && UI.interactive?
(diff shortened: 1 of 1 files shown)


joshdholtz:
@nicolasbraun I think the solution to your problem wouldn't be the change proposed in this PR. I would suggest just manually setting your `export_type` to `development` instead.

nicolasbraun:
@joshdholtz thanks for stepping in.
Thats what i currently do and it's woking fine. I just don't get why we do this. This is totally unnatural to the user since his build is destined to fail, gym will inevitably fail if trying to export with method `app-store`  and match was used using `development`

am I missing something here ?

joshdholtz:
@nicolasbraun I think the reason this is done is because 99% of the time, people will be building for `app-store` so this makes it easier for that. If someone is doing something a little bit more custom, they will look at options and explicitly set the export type 🙃

Issue #10931, opened 2017-11-17
Couldn't decrypt the repo, please make sure you enter the right password! - running match


mberthos:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
When running command for generate and management certified : 
- fastlane match development

Require a password for encrypt/decrypt, but this password not exists, i tried all passwords, but not success.

##### Complete output when running fastlane, including the stack trace and command used
INFO [2017-11-17 05:32:16.88]: Cloning remote git repo...
INFO [2017-11-17 05:32:16.88]: If cloning the repo takes too long, you can use the `clone_branch_directly` option in match.
INFO [2017-11-17 05:32:16.88]: $ GIT_TERMINAL_PROMPT=0 git clone 'git@bitbucket.org:tecsinapse/ios-certificates.git' '/var/folders/jw/tmzdrc590w7f2m3tsg5v5hb00000gp/T/d20171117-30393-1xhuu4d'
INFO [2017-11-17 05:32:16.89]: ▸ Cloning into '/var/folders/jw/tmzdrc590w7f2m3tsg5v5hb00000gp/T/d20171117-30393-1xhuu4d'...
INFO [2017-11-17 05:32:17.17]: ▸ remote: Counting objects: 12, done.
INFO [2017-11-17 05:32:17.17]: ▸ remote: Compressing objects: 100% (11/11), done.
INFO [2017-11-17 05:32:17.17]: ▸ remote: Total 12 (delta 0), reused 0 (delta 0)
INFO [2017-11-17 05:32:17.17]: ▸ Receiving objects: 100% (12/12), 7.55 KiB | 0 bytes/s, done.
INFO [2017-11-17 05:32:17.18]: ▸ Checking connectivity... done.
WARN [2017-11-17 05:32:17.22]: Enter the passphrase that should be used to encrypt/decrypt your certificates
WARN [2017-11-17 05:32:17.22]: This passphrase is specific per repository and will be stored in your local keychain
WARN [2017-11-17 05:32:17.22]: Make sure to remember the password, as you'll need it when you run match on a different machine
 [2017-11-17 05:32:17.22]: Passphrase for Git Repo: *******
[...]

Issue #11858, opened 2018-02-16
Couldn't decrypt the repo when using swift in syncCodeSigning


pballartzinio:
### New Issue Checklist

- [x ] Updated fastlane to the latest version
- [x ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I have been using fastlane for some months and have correctly setup match for code signing. Now I'm trying to upgrate to the new swift language so I'm translating the ruby script to be in Swift and use The FastlaneRunner and the generate xcode project and so.
The issue I'm having is that when running syncCodeSigning function it throws an error saying that the password is wrong when i'm sure is correct. Then I run plain 'fastlane run match' and introduce the same params when asked (git url, password) and it successfully decrypts the certificates and everything works fine. The only difference I can see is that one is run from old-way ruby and the other is from swift and goes through all the RubyCommand.Argument stuff.

daneov:
I've just encountered this behaviour myself.
Match seemed to work correctly in `fastlane 2.80.0`, and started displaying this behaviour in `2.81.0`.

I've started porting to Swift as well, but did so on `2.80.0`, where this worked. 

I'm fairly certain if you downgrade to `2.80.0` and regenerate the Swift runner (deleting the old one, moving your fastlane folder to `fastlane.bak`, running `fastlane init swift` and then `mv fastlane/swift fastlane.bak/`), that it'll work. Not in the slightest because I just did the same 😉 

On a sidenote: I created #11854  to know if there's a way to simplify this regeneration

daneov:
Just from the changelogs in `2.81.0`, I'd assume something goes wrong with either of these 
* [match/utils] using ruby openssl instead of terminal version (#11720) via Cyberbeni
* [match] using ruby openssl instead of terminal one (#11641) via Cyberbeni

Cyberbeni:
@pballartzinio what is your `openssl version`?

pballartzinio:
I had the contained version installed 2.81.0 which was using OpenSSL 1.0.2g  1 Mar 2016 
Now I removed it and installed version 2.80.0 via gem and match works correctly but now I can see OpenSSL 1.0.2l  25 May 2017.
I hope that helps. Meanwhile I will stick with version 2.80.0

[...]

PR #12112, opened 2018-03-20
Write Match encrypted files in binary mode to avoid UndefinedConversi…

emes:
…on error

### Checklist

### Motivation and Context
Writing encrypted files from a Match git repo resulted in a conversion error when values could not be converted from ASCII-8BIT to UTF-8.

For example: 
```
Exception in decrypt Encoding::UndefinedConversionError "\x82" from ASCII-8BIT to UTF-8: 
...match/lib/match/encrypt.rb:123:in `write'
```

(Fastlane might want to consider Encoding.default_external = "ASCII-8BIT" to avoid the myriad changes from read/write to binread/binwrite.)

### Description
Open the file for write in binary mode to avoid the UTF-8 conversion.

This fixes at least some of the issues reported in #11858, but I'm not using Swift, so others will need to verify.

Specifically, I no longer get the following (misleading ;-) error:

```
ERROR [2018-03-06 16:31:52.48]: Couldn't decrypt the repo, please make sure you enter the right password!
ERROR [2018-03-06 16:31:52.94]: Invalid password passed via 'MATCH_PASSWORD'
```

### DIFF
--- a/match/lib/match/encrypt.rb
+++ b/match/lib/match/encrypt.rb
@@ -119,7 +119,7 @@ module Match
 
       decrypted_data = decipher.update(data_to_decrypt) + decipher.final
 
-      File.write(path, decrypted_data)
+      File.binwrite(path, decrypted_data)
     rescue => error
       fallback_hash_algorithm = "SHA256"
       if hash_algorithm != fallback_hash_algorithm


joshdholtz:
@emes This is a really good find 💪🔍 So you were seeing this issue in Ruby and this fixed it for you, right?

Might also fix #10931

emes:
Correct.  I tested extensively.  No issues at all once I started binwriting :-)  I had a very similar issue with dsyms a while back, so I'm regretting not having thought of this sooner...

joshdholtz:
Ahhh, we do do a `binwrite` on dSYMs 🤔 I like this fix 👍

joshdholtz (approved):

PR #12124, opened 2018-03-21
Make sure the schemes list includes workspace contained schemes

### LINKED ISSUE #12113: xcodeproj cannot see schemes in the workspace container (error is "Couldn't find specified scheme")
In 2.85.0 everything worked for me. When I updated to 2.86.0, gym could not find my schemes.

I debugged and I see that the root cause is xcodeproj's Workspace.schemes does not load xcscheme files that are in the workspace container

I made a simple ruby script

```
require 'xcodeproj'
project_path = './Trellis.xcworkspace/'
workspace = Xcodeproj::Workspace.new_from_xcworkspace(project_path)
workspace.load_schemes(project_path)
print workspace.schemes
```

I ran it

```
{"Pods"=>"/Users/lfranco/src/ios/Trellis.xcworkspace/Pods/Pods.xcodeproj", "TrelloUIKit"=>"/Users/lfranco/src/ios/Trellis.xcworkspace/TrelloUIKit/TrelloUIKit.xcodeproj", "Trellis"=>"/Users/lfranco/src/ios/Trellis.xcworkspace/Trellis.xcodeproj"}
```

But, I have all of these schemes:

```
20:27 $ find Trellis.xcworkspace -name '*.xcscheme'
Trellis.xcworkspace/xcshareddata/xcschemes/Trello Today Widget.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trellis.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trello Watch App.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trello Share Extension.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trello Watch Notification.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trellis Ad Hoc.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trello Notification Service Extension.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/Trello Stickers.xcscheme
Trellis.xcworkspace/xcshareddata/xcschemes/TrelloEnterprise.xcscheme
```

I realize the bug is in xcodeproj, but I think you should think of reverting.  I am going to report to them as well, but it will take time to update, I think.

As an alternative, you could load the workspace schemes and add them to the list generated in project.rb `schemes` method.

To make this a little easier to find in searches, this is the error I get from gym:

```
Couldn't find specified scheme 'Trellis Ad Hoc'.
-- | --
Multiple schemes found but you haven't specified one.
Since this is a CI, please pass one using the `scheme` option
Couldn't find any schemes in this project, make sure that the scheme is shared if you are using a workspace
Open Xcode, click on `Manage Schemes` and check the `Shared` box for the schemes you want to use
Afterwards make sure to commit the changes into version control
```

loufranco:
### Checklist

### Motivation and Context
Fixes #12113 
The xcodeproj library does not see schemes that are in the .xcworkspace (as opposed to a project), so when we get a list of schemes, it doesn't include them. This causes issues downstream as that list is checked against for any scheme that you pass in.

### Description
I added a new method to get the workspace schemes and then I called that in project.schemes to append them on.

I added a .xcworkspace with this kind of setup and added tests to make sure its schemes are read correctly. We already had tests for workspaces without workspace schemes and they still pass.

I believe that it's possible that the xcodeproj project will add a fix (I reported to them), but I don't know the exact way they will do that -- their scheme list is a map from schemes to project files -- so they might decide to have a new method with the workspace ones.   If they make their Workspace.schemes method show these, then the code in the PR will be wrong (but the test will catch that).

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -112,12 +112,21 @@ module FastlaneCore
       @project ||= Xcodeproj::Project.open(path)
     end
 
+    # xcodeproj doesn't include schemes that are under the .xcworkspace folder.
+    # So, get a list of them to include.
+    # This is based on xcodeproj's Project::schemes method.
+    def workspace_contained_schemes
+      Dir[File.join(path, 'xcshareddata', 'xcschemes', '*.xcscheme')].map do |scheme|
+        File.basename(scheme, '.xcscheme')
+      end
+    end
+
     # Get all available schemes in an array
     def schemes
       @schemes ||= if workspace?
                      workspace.schemes.reject do |k, v|
                        v.include?("Pods/Pods.xcodeproj")
-                     end.keys
+                     end.keys + self.workspace_contained_schemes
                    else
                      Xcodeproj::Project.schemes(path)
                    end
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -115,6 +115,8 @@ module FastlaneCore
     # xcodeproj doesn't include schemes that are under the .xcworkspace folder.
     # So, get a list of them to include.
     # This is based on xcodeproj's Project::schemes method.
+    # To Do: When Xcodeproj merges https://github.com/CocoaPods/Xcodeproj/issues/557, we
+    # should remove this code and make sure to call Workspace.load_schemes
     def workspace_contained_schemes
       Dir[File.join(path, 'xcshareddata', 'xcschemes', '*.xcscheme')].map do |scheme|
         File.basename(scheme, '.xcscheme')
--- a/fastlane_core/spec/fixtures/projects/workspace_schemes/WorkspaceSchemes.xcworkspace/contents.xcworkspacedata
+++ b/fastlane_core/spec/fixtures/projects/workspace_schemes/WorkspaceSchemes.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "group:WorkspaceSchemesFramework/WorkspaceSchemesFramework.xcodeproj">
+   </FileRef>
+   <FileRef
+      location = "group:WorkspaceSchemesApp/WorkspaceSchemesApp.xcodeproj">
+   </FileRef>
+</Workspace>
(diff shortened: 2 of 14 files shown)


Statusgman:
This PR fixed my issue. All schemes from workspace are visible for "scan" now! 👍

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -112,12 +112,21 @@ def project
       @project ||= Xcodeproj::Project.open(path)
     end
 
+    # xcodeproj doesn't include schemes that are under the .xcworkspace folder.
```
DanToml: Could you add a TODO with a link to https://github.com/CocoaPods/Xcodeproj/issues/557 if this is merged before we bump Xcodeproj please?

endocrimes (approved):
This lgtm, if we need to ship this _now_ the go ahead, but I'd rather ship an updated xcodeproj. (Can cut a release tomorrow if the PR is in)

fastlane_core/lib/fastlane_core/project.rb
loufranco: It looks like it might take a day for my job to get me on the CLA, so I'll either put the TODO or just do the bump and call to load_schemes (depending on where xcodeproj is)

loufranco:
@googlebot I signed. I am covered under Atlassian (using my Atlassian email -- which is on the commits)

fastlane_core/lib/fastlane_core/project.rb
loufranco: @DanToml I added a "To Do" (this project's rubocop disallows TODO).  This is ready to merge -- I also added a PR to Xcodeproj -- if you would prefer to close this PR, and merge the xcode one -- I'd be happy to bump Xcodeproj here and add the load_schemes call.

loufranco:
@DanToml I added the comment and then appveyor failed -- looks like it had some kind of network issue getting GTX -- is there a way to re-run it?

loufranco:
@joshdholtz is there a way to re-run appveyor? All I did was add a comment -- and its log looks like it had a network issue.

joshdholtz:
@loufranco I can look into it! But I think I would like to see if we can get your `xcodeproj` PR pushed out today (https://github.com/CocoaPods/Xcodeproj/pull/559) so we can change the dependency on that instead since 😊

Statusgman:
I have one more issue in this PR's branch:
scan now lists schemes with dublicated one:
```
1. App
2. AppCore
3. AppAnalytics
1. App
```
I don't know if this issue is caused by changes in PR.

loufranco:
@Statusgman this code is going to change today.  Xcodeproj is being updated to include the proper behavior and that version does use a map for schemes, so the keys will be deduped.  I'll ping you when a PR to fastlane with that is ready so you can take a look.

PR #12128, opened 2018-03-21
[supply] Added internal track

### LINKED ISSUE #12117: [supply] Add support for internal test
### New Issue Checklist

### Issue Description

Google introduced internal test as a new way to distribute apks to testers.
So I tried to setup supply to upload my apk to intern test track, but supply doesn't allow track which are not in [this](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/uploader.rb#L184) array.

Maybe it is enough to expand the array and add the right track name?

I couldn't find the correct track name in the [google developer api docs](https://developers.google.com/android-publisher/api-ref/edits/tracks) , but the docs was last updated on 15 February 2018.

It would be great if someone could find a way to bring internal testing support to supply :)

joshdholtz:
Fixes #12117

- Allowed for new `internal` track for supply

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -181,7 +181,7 @@ module Supply
       max_apk_version_code = apk_version_codes.max
       max_tracks_version_code = nil
 
-      tracks = ["production", "rollout", "beta", "alpha"]
+      tracks = ["production", "rollout", "beta", "alpha", "internal"]
       config_track_index = tracks.index(Supply.config[:track])
 
       tracks.each_index do |track_index|
--- a/supply/spec/uploader_spec.rb
+++ b/supply/spec/uploader_spec.rb
@@ -158,27 +158,28 @@ describe Supply do
 
       it 'remove lesser than the greatest of any later (i.e. production) track' do
         allow(client).to receive(:track_version_codes) do |track|
-          next [103] if track.eql?('production')
-          next [102] if track.eql?('rollout')
-          next [101] if track.eql?('beta')
+          next [104] if track.eql?('production')
+          next [103] if track.eql?('rollout')
+          next [102] if track.eql?('beta')
+          next [101] if track.eql?('alpha')
           []
         end
 
         allow(client).to receive(:update_track) do |track, rollout, apk_version_code|
-          expect(track).to eq('beta').or(eq('rollout'))
+          expect(track).to eq('beta').or(eq('rollout')).or(eq('alpha'))
           expect(rollout).to eq(1.0)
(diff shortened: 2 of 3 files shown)


supply/spec/uploader_spec.rb
```diff
@@ -158,27 +158,28 @@ def find_obbs
 
       it 'remove lesser than the greatest of any later (i.e. production) track' do
         allow(client).to receive(:track_version_codes) do |track|
-          next [103] if track.eql?('production')
-          next [102] if track.eql?('rollout')
-          next [101] if track.eql?('beta')
+          next [104] if track.eql?('production')
```
janpio: I don't understand how this code works, but that the number <-> track combination is now different doesn't mean anything?

joshdholtz: Yeah yeah, so I bumped up the versions so that `production` has highest version and `internal` has lowest version. `internal` needs to be the lowest so I just bumped up the other ones so that I could insert `internal` as the lowest.

joshdholtz: This tests tried to remove any older versions than the latest production version

nekator:
@joshdholtz So internal is the right track name? Is it documented somewhere? I don't doubt it but I'm interested in how you figured it out.

mandybess:
@nekator yupp 

documented here: https://developers.google.com/apis-explorer/#search/Google%20Play%20Developer%20API/androidpublisher/v2/androidpublisher.edits.tracks.update

nekator:
@mandybess thanks for the link.

joshdholtz:
@nekator Yup, internal is the correct link! I uploaded an APK to the internal track and verified it via Google Play dashboard 😊 

screen shot 2018-03-22 at 10 42 04 am

nekator:
@joshdholtz thanks. 
Now I can't wait until this pr will get merged 🤩

supply/lib/supply/uploader.rb
```diff
@@ -181,7 +181,7 @@ def check_superseded_tracks(apk_version_codes)
       max_apk_version_code = apk_version_codes.max
       max_tracks_version_code = nil
 
-      tracks = ["production", "rollout", "beta", "alpha"]
+      tracks = ["production", "rollout", "beta", "alpha", "internal"]
```
minuscorp: why not `%w(production beta alpha internal rollout)`?

KrauseFx: Either way is fine, we can leave it as is

snatchev (approved):
LGTM

PR #12134, opened 2018-03-22
Use correct spelling of icloud service in product/service.rb

jakubknejzlik:
These changes are already done in produce/developer_center.rb

https://github.com/fastlane/fastlane/pull/11861

### Checklist

### Motivation and Context
This extends this fix #11861 

### Description
Use the correct method Spaceship::Portal::AppService.cloud because the constant is named Cloud instead of iCloud. The latter would cause that the Ruby-style method would be named i_cloud. That doesn't look better either.

### DIFF
--- a/produce/lib/produce/developer_center.rb
+++ b/produce/lib/produce/developer_center.rb
@@ -98,7 +98,7 @@ module Produce
             enabled_clean_options[app_service.cloud.on.service_id] = app_service.cloud.on
             enabled_clean_options[app_service.cloud_kit.xcode5_compatible.service_id] = app_service.cloud_kit.xcode5_compatible
           when SERVICE_CLOUDKIT
-            enabled_clean_options[app_service.i_cloud.on.service_id] = app_service.i_cloud.on
+            enabled_clean_options[app_service.cloud.on.service_id] = app_service.cloud.on
             enabled_clean_options[app_service.cloud_kit.cloud_kit.service_id] = app_service.cloud_kit.cloud_kit
           end
         else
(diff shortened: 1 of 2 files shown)


jakubknejzlik:
@googlebot I signed it!

joshdholtz (commented):
These changes look good! Just one quick question... Do you know if we need to change this line at all? https://github.com/fastlane/fastlane/blob/master/produce/lib/produce/service.rb#L153

jakubknejzlik:
@joshdholtz oh, that's very good point! As always "it worked on our side" even with the service removals, but judging from current version of app_service.rb...
https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/app_service.rb#L51

...it should be updated. Unfortunately I'm not sure what's the desired behaviour, disable both services or even with the cloudkit?
```
app.update_service(Spaceship.app_service.cloud.off)
app.update_service(Spaceship.app_service.i_cloud.off)
```

joshdholtz:
@jakubknejzlik I would assume both would make sense 😊

jakubknejzlik:
Me too, but the fact that this area isn't covered by unit tests seems like very thin ice 😄

joshdholtz:
@snatchev 👋 Do you have any thoughts on the above? 😇

joshdholtz:
I signed it!

joshdholtz:
@jakubknejzlik Hope its okay that I pushed a commit to your PR 😊 

1) "off" should use `cloud.off` 
2) `i_cloud` was actually wrong (from previous PR) and should just be `cloud.on`

joshdholtz:
Also, my CLA is signed so we are all good, @googlebot 👍

taquitos:
All CLAs are fine 👍

joshdholtz (approved):

PR #12075, opened 2018-03-17
Ignore FastlaneSwiftRunner.xcodeproj for detecting ios projects of users

nafu:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Resolve https://github.com/fastlane/fastlane/issues/11606

### Description

`fastlane init` always detect the projects are ios projects because fastlane has FastlaneSwiftRunner.xcodeproj inside.
That is why I Ignore FastlaneSwiftRunner.xcodeproj for detecting ios projects of users.

PR #12142, opened 2018-03-23
Ignore *.xcodeproj & *.xcworkspace in gem paths for detecting ios projects

chiastolite:
### Checklist

### Motivation and Context
`fastlane init` always ios projects when install the fastlane by bundle with `--(project_dir)/vendor/bundle` option, because some gems(include fastlane)  has *.xcodeproj or *.xcworkspace inside.

This fix related with #12075

### Description

Ignore *.xcodeproj & *.xcworkspace files in Gem paths.

### DIFF
--- a/fastlane/lib/fastlane/setup/setup.rb
+++ b/fastlane/lib/fastlane/setup/setup.rb
@@ -54,6 +54,9 @@ module Fastlane
       spinner.auto_spin
 
       ios_projects = Dir["**/*.xcodeproj"] + Dir["**/*.xcworkspace"]
+      ios_projects.delete_if do |path|
+        Gem.path.any? { |gem_path| File.expand_path(__dir__, path).start_with?(gem_path) }
+      end
       ios_projects.delete_if { |path| path.match("fastlane/swift/FastlaneSwiftRunner/FastlaneSwiftRunner.xcodeproj") }
       android_projects = Dir["**/*.gradle"]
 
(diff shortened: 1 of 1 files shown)


chiastolite:
I signed it!

nafu (approved):
Nice catch 👍 

Thank you for your contribution 💚

fastlane/lib/fastlane/setup/setup.rb
```diff
@@ -54,6 +54,9 @@ def self.start(user: nil, is_swift_fastfile: false)
       spinner.auto_spin
 
       ios_projects = Dir["**/*.xcodeproj"] + Dir["**/*.xcworkspace"]
+      ios_projects.delete_if do |path|
+        Gem.path.any? { |gem_path| File.expand_path(__dir__, path).start_with?(gem_path) }
```
nafu: @chiastolite 

[Q] Where do you want to refer with `__dir__`?

I think this should be `Dir.pwd`. I'd appreciate if you could check this again and make sure this works as you intended.

nafu (changes requested):
I'm sorry I've once approved but I have one question to make sure this works correctly.

fastlane/lib/fastlane/setup/setup.rb
nafu: [IMO] `File.expand_path(path)` would be your solution 👍

chiastolite: > [IMO] File.expand_path(path) would be your solution 👍

Oh, you’re right! I will fix it.

nafu (approved):
@chiastolite This is great 🚀 Thank you!

PR #12157, opened 2018-03-24
Add tests for verify build action

mgrebenets:
### Checklist

### Motivation and Context
`verify_build` action is lacking any test coverage. This PR adds tests for `verify_build` action.

### Description
2 new ipa files added to test fixtures
- One properly signed `very-capable-ipa.ipa`, which also happens to have all capabilities enabled available at the moment, which will come handy when extending `verify_build` action with more functionality
- One incorrect zip archive `not-an-ipa.ipa` which is just a text file


mgrebenets:
cla signed

fastlane/spec/actions_specs/verify_build_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      describe "Missing ipa file" do
+        it "raises an error if ipa file path set via lane context does not exist" do
+          Fastlane::Actions.lane_context[Fastlane::Actions::SharedValues::IPA_OUTPUT_PATH] = no_such_file
+
+          expect(FastlaneCore::UI).to receive(:user_error!).with("Unable to find ipa file '#{no_such_file}'.").and_call_original
+          expect do
+            Fastlane::FastFile.new.parse("lane :test do
+                verify_build(
+                    ipa_path: '#{no_such_file}'
```
getaaron: Since this test is testing `lane_context` then I think we should just call `verify_build` without passing `ipa_path` here, is that right? If we pass `ipa_path` explicitly that will override the `lane_context` value.

getaaron (changes requested):
@mgrebenets Thanks for the PR, this looks great. Just had one question, otherwise this is is good to merge

fastlane/spec/actions_specs/verify_build_spec.rb
mgrebenets: Yep, copy-paste error :) thanks for picking it up :)

mgrebenets: oh, does it loose context when I squash commits and force push?
I am pretty sure this thread of comments was on line 41 before...

getaaron:
Looks good now but I think you force pushed commits with the wrong author tag :)

mgrebenets:
@getaaron yeah, typed the wrong email, fixed it, CLA should be OK now.

getaaron (approved):

PR #12162, opened 2018-03-24
Default skip_build to true in Scanfile template

### LINKED ISSUE #12050: 'scan' should not run the 'build' action by default
### New Issue Checklist

### Issue Description

The default behavior of the `scan` action is to run `xcodebuild [options] build test`. The problem here is if the scheme specifies a different configuration for Run vs Test, this ends up building the project twice, once with the Run configuration and then again with the Test configuration.

I can't think of any reason why it should run the `build` action at all. The `test` action already builds what's necessary, so all this does is slow down the scan. If I want to do a debug build first, I should be using a separate action for that. As it is, the current behavior is surprising and penalizes anyone who uses the default configuration.

Reading through the code I noticed there's a config flag to skip the `build` action. So obviously someone realized the `build` was pointless, but they didn't go far enough. The config flag should be inverted so it only inserts `build` if explicitly instructed to. Since it looks like the config flags have default values, this could be accomplished simply by switching the default value of `:skip_build` to `true`.

getaaron:
As @kballard notes, _scan_ should not run the 'build' action by default because its output is typically unused.

We don't want to change the default option to avoid breaking existing setups, but we can at least opt new users into skipping the unnecessary build by setting the appropriate option in the default `Scanfile`. Fixes #12050.

### DIFF
--- a/scan/lib/assets/ScanfileTemplate
+++ b/scan/lib/assets/ScanfileTemplate
@@ -11,3 +11,6 @@
 # open_report(true)
 
 # clean(true)
+
+# Enable skip_build to skip debug builds for faster test performance
+skip_build(true)


revolter (approved):
LGTM 👍🏻

Issue #12140, opened 2018-03-23
Element Window cannot request screenshot data because it has an empty frame SnapshotHelper.swift:164


sirdharma:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I am trying to set up Snapshot and have followed the Snapshot installation steps.

I have a single UI test that's pretty straightforward:

```
class XxxUITests: XCTestCase {
        
    override func setUp() {
        super.setUp()
        continueAfterFailure = false
        let app = XCUIApplication()
        setupSnapshot(app)
        app.launch()
    }
    
    override func tearDown() {
        super.tearDown()
    }
    
    func testOnboarding() {
        let app = XCUIApplication()

        let getStartedButton = app.buttons["signin.button.get started"]
        let nextButton = app.buttons["signin.button.next"]
        let joinButton = app.buttons["signin.button.join"]

        XCTAssert(getStartedButton.exists)
        snapshot("01Onboarding")
        getStartedButton.tap()

        XCTAssert(nextButton.exists)
        snapshot("02Onboarding")
        nextButton.tap()

        XCTAssert(nextButton.exists)
        snapshot("03Onboarding")
        nextButton.tap()

        XCTAssert(joinButton.exists)
        snapshot("04Onboarding")
    }
    
}
```

- The tests run fine when I comment out the calls to `snapshot()`.
- The tests run fine when I leave the calls to `snapshot()` but comment out the call to `setupSnapshot()` (I get a warning in the console that says to call `setupSnapshot`, as is expected).
- I get the following error `Element Window cannot request screenshot data because it has an empty frame SnapshotHelper.swift:164` when I run the tests otherwise (with Xcode or with `fastlane snapshot`).

My `Snapfile` is as follows:

```
devices([
  "iPhone SE",
])
clear_previous_screenshots(true)
concurrent_simulators(false)
```

Any hint towards the right direction would be much appreciated. Thank you!

##### Complete output when running fastlane, including the stack trace and command used

`fastlane snapshot --capture_output` returns an error

<details>
  <pre>invalid option: --capture_output</pre>
</details>

### Environment

<details>
[...]

PR #12174, opened 2018-03-26
[snapshot] Make sure matched window has a non-empty frame

sirdharma:
### Checklist

### Motivation and Context

When finding a window to call `screenshot()` on, the provided code in `SnapshotHelper.swift` doesn't ensure that said window has a non-empty frame. This results in the call to `screenshot()` raising the following error: `Element Window cannot request screenshot data because it has an empty frame SnapshotHelper.swift:164`

This change solves https://github.com/fastlane/fastlane/issues/12140.

`SnapshotHelper.swift` does not appear to be tested at all. My apologies if this is the case and I just missed it.

### Description

Rather than using `firstMatch` on `app.windows`, we make sure that the selected window has a non-empty frame by using `allElementsBoundByIndex.first(where: { $0.frame.isEmpty == false })`.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -278,4 +278,4 @@ private extension CGFloat {
 
 // Please don't remove the lines below
 // They are used to detect outdated configuration files
-// SnapshotHelperVersion [1.9]
+// SnapshotHelperVersion [1.10]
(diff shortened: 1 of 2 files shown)


sirdharma:
@googlebot I signed it!

janpio:
You would have to increase the version in the file in the last line to get this merged and released (1.9 => 1.10).

getaaron (approved):

getaaron:
Oh, I guess our version regex doesn't support two-digit version numbers

getaaron:
Thanks for the PR @sirdharma 👍

PR #12176, opened 2018-03-26
Fix crashlytics to not autoload gsp_path if api_token is set

### LINKED ISSUE #12132: Upload to Crashlytics plugin has no way to disable Firebase plist upload
### New Issue Checklist

### Issue Description
Recently, UploadSymbolsToCrashlyticsAction was updated to support the relatively new feature of uploading symbols to Fabric using a GoogleServices plist file (#11798). While this is listed as an optional setting, if there is a GoogleServices.plist file anywhere in the directory structure, the action will now attempt to upload using the plist.
Unfortunately, this causes problems in multi-organization scenarios for Fabric. Firebase doesn't currently support having different configurations within the same app (to support things like enterprise versions or internal versions), so there is only one plist file. At the same time, uploading to Fabric with the plist overrides any value used for the Fabric App ID.
What this results in is that the dSYM files can be uploaded to the wrong Fabric Organization, which will result in crashes being unable to be symbolicated.

This action should be update to truly make the GoogleServices plist optional and not pull it in automatically if found.

We have created a workaround by replicating the previous version of the UploadSymbolsToCrashlyticsAction as a local action and using the local action instead of the real action, but we would prefer using the official action, if possible.

joshdholtz:
Fixes #12132 

## Wut
- `upload_symbols_to_crashlytics` would automatically use a `gsp_path` (by searching for it in project directory) even if an `api_token` was being used
  - only one of the setting needs to be used
  - should only auto look for `gsp_path` if no `api_token`
  - using the `gsp_path` is optional but would ALWAYS be used if a `GoogleServices-Info.plist` existed in the project (even if not wanted)

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -110,6 +110,8 @@ module Fastlane
       end
 
       def self.find_gsp_path(params)
+        return if params[:api_token]
+
         if params[:gsp_path].to_s.length > 0
           params[:gsp_path] = File.expand_path(params[:gsp_path])
         else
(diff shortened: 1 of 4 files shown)


getaaron (approved):
LGTM 👍🏼

PR #12178, opened 2018-03-27
Use PrintTable instead of puts in validate_build action

mgrebenets:
### Checklist

### Motivation and Context
Old implementation was using `puts` to print out a table.
This wasn't inline with the rest of the code base and couldn't be tested easily either.

### Description
This change will use Fastlane's helper to print out table, plus it adds tests to check that the output is correct.


getaaron:
Are we able to remove `require 'terminal-table'` at the top now? Or do we need that for something else?

mgrebenets:
I guess it's no longer needed, removed it 👍

fastlane/spec/actions_specs/verify_build_spec.rb
```diff
@@ -5,6 +5,17 @@
...
       let(:incorrectly_signed_ipa) { File.expand_path("./fastlane_core/spec/fixtures/ipas/ContainsWatchApp.ipa") }
+      let(:expected_title) { "Summary for verify_build #{Fastlane::VERSION}" }
+      let(:expected_app_info) do
+        {
+          "bundle_identifier" => "org.fastlane.very-capable-app",
+          "provisioning_type" => "distribution",
+          "team_identifier" => "5UKXVKJP8C",
+          "app_name" => "XC org fastlane very-capable-app",
+          "provisioning_uuid" => "ef6962fa-f6b4-4203-90a4-11480af19ec7",
+          "team_name" => "Maksym Grebenets"
```
joshdholtz: Is this your actual `team_identifier`, `provisioning_uuid`, and `team_name`? If so, do you want to scrub this at all? 🙃

joshdholtz (changes requested):

fastlane/spec/actions_specs/verify_build_spec.rb
mgrebenets: well... that's the tricky part...

It is my real team ID and name, but how can I test code signing otherwise? 
I need a valid IPA properly packaged by Apple tools and properly signed, as well as having all the correct/matching entitlements and profiles. I could potentially stub all the code that reads entitlements and code-signing info, but then what would I be testing in that case :)

Using real developer account is the only way I could get such an IPA so far.
The app is just a dummy app and doesn't even have an iTunes Connect entry, just the app ID and stuff.
The Team ID is not a secret and is easily obtainable by anyone afaik. 
Same thing for profile UDID.

In general I was reluctant using my personal team ID, but I couldn't come up with better alternative. Not sure if this is the best way. 

Do you think we could have a bunch of IPA fixtures signed by some special designated Fastlane-owned iOS developer account? There's quite a few actions in Fastlane that involve signing and re-signing and so far we haven't had good tests for those actions.

mgrebenets: I'll do some research if it's possible to create an iOS Signing Identity without even having developer account.

Such an identity would be useless for any kind of distribution, but would be a perfect test fixture.

mgrebenets: Hey @joshdholtz 

I've took care of it.
The fixture `ipa` is now a completely fake ipa, with fake team id and all that stuff, signed by fake signed identity.

These are the values:

```ruby
        {
          "bundle_identifier" => "org.fastlane.very-capable-app",
          "team_identifier" => "TestFixture",
          "app_name" => "very-capable-app",
          "provisioning_uuid" => "12345678-1234-1234-1234-123456789012",
          "team_name" => "TestFixture"
        }
```

mgrebenets:
@joshdholtz I still can't get used to the fact that force pushing to my branch makes review comments "invisible".

The update is that I've managed to create a completely fake, but properly signed `ipa`, which is using fake self-signed certificate and signing identity, fake team name, profile UUID and everything else.

I think I will just upload that whole fake signing identity as a fixture some time soon.
It will be essential for finally having tests for `sigh resign` action.

For this change, the app info looks like this:

```ruby
        {
          "bundle_identifier" => "org.fastlane.very-capable-app",
          "team_identifier" => "TestFixture",
          "app_name" => "very-capable-app",
          "provisioning_uuid" => "12345678-1234-1234-1234-123456789012",
          "team_name" => "TestFixture"
        }
```

joshdholtz (approved):

PR #12027, opened 2018-03-10
Use xcodeproj to introspect Xcode projects

endocrimes:
This should be faster, and stops us from needing to maintain a parser for
the command line output of Xcode.

It also allows more tests to not require a full Xcode installation, and should
speed things up a little.

PR #12182, opened 2018-03-27
Improve error message when specified scheme is not found

0xced:
Since #12027, schemes are read with Xcodeproj rather than parsed from `xcodebuild -list` output. Unlike `xcodebuild -list`, Xcodeproj reads only **shared** schemes (from [project](https://github.com/CocoaPods/Xcodeproj/blob/50e70d32d1176d14c57807c8b5adaa6f42ef38ac/lib/xcodeproj/project.rb#L780-L793)
and also from [workspace](https://github.com/CocoaPods/Xcodeproj/blob/50e70d32d1176d14c57807c8b5adaa6f42ef38ac/lib/xcodeproj/workspace.rb#L199-L204)).

As the specified scheme might exist on the user machine but not be shared, the error message is improved to suggest that the scheme must be shared with a link to Apple's documentation.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -131,7 +131,7 @@ module FastlaneCore
       if options[:scheme].to_s.length > 0
         # Verify the scheme is available
         unless schemes.include?(options[:scheme].to_s)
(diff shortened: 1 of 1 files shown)


getaaron (approved):

getaaron:
Thanks for the PR, @0xced 👍

PR #12185, opened 2018-03-27
[match] Check OpenSSL version on attempt to decrypt certificates repo

### LINKED ISSUE #12149: [Match] Check OpenSSL version on attempt to decrypt certificates repo
### New Issue Checklist

### Issue Description
This is rather suggestion, than an issue. Can match be updated to store the OpenSSL version that was used during certificates encryption in something like a .openssl-version file in the repo and show warning on attempt to decrypt certificates with different OpenSSL version?
Right now if there is a mismatch between OpenSSL versions used to encrypt and decrypt files, an "Incorrect password" error will be shown even if password is correct, which is very misleading. I've experienced such situation during attempt to decrypt the repo, which was encrypted with LibreSSL 2.2.7, with OpenSSL 1.1.0e.
If such check will be integrated, it will eliminate issues like https://github.com/fastlane/fastlane/issues/7127. Thanks.

viktorasl:
### Checklist

### Motivation and Context
Fixes #12149.

### Description
Checks if repo was encrypted with the currently used OpenSSL version if OpenSSL version file exists. Backwards compatible as missing file is ignored.
Introduces step to create the OpenSSL version file on commit+push action.

Written some unit tests, they're quite messy I guess as I don't have much experience working with ruby & RSpec. So any advice is more than welcomed!
In addition I've tested it in mocked environment, works fine.

Questionable:
* checks for exact OpenSSL version (e.g `OpenSSL 1.0.2k  26 Jan 2017`) instead of developer (?) (e.g `OpenSSL`)
* ignores missing OpenSSL version file making this backwards compatible
* raises OpenSSL mismatch issue instead of doing something clever
* unit tests have some hardcoded values

### DIFF
--- a/match/lib/match/git_helper.rb
+++ b/match/lib/match/git_helper.rb
@@ -6,6 +6,7 @@ require_relative 'encrypt'
 module Match
   class GitHelper
     MATCH_VERSION_FILE_NAME = "match_version.txt"
+    OPENSSL_VERSION_FILE_NAME = "openssl_version.txt"
 
     def self.clone(git_url,
                    shallow_clone,
@@ -59,6 +60,16 @@ module Match
 
       checkout_branch(branch) unless branch == "master"
 
+      # makes it backwards compatible allowing checkout and decryption of repo
+      # if open ssl version file not exist
+      openssl_version_file = File.join(@dir, OPENSSL_VERSION_FILE_NAME)
+      if File.exist?(openssl_version_file)
+        repo_openssl_version = File.read(openssl_version_file)
+        if repo_openssl_version != Encrypt.openssl_version.to_s
+          UI.user_error!("Repository was encrypted using '#{repo_openssl_version}' and currently used version is '#{Encrypt.openssl_version}'")
+        end
+      end
+
       if !Helper.test? && GitHelper.match_version(@dir).nil? && manual_password.nil? && File.exist?(File.join(@dir, "README.md"))
         UI.important("Migrating to new match...")
         ChangePassword.update(params: { git_url: git_url,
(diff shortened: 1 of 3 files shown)


match/lib/match/git_helper.rb
```diff
@@ -59,6 +60,16 @@ def self.clone(git_url,
 
       checkout_branch(branch) unless branch == "master"
 
+      # makes it backwards compatible allowing checkout and decryption of repo
+      # if open ssl version file not exist
+      openssl_version_file = File.join(@dir, OPENSSL_VERSION_FILE_NAME)
+      if File.exist?(openssl_version_file)
+        repo_openssl_version = File.read(openssl_version_file)
+        if repo_openssl_version != Encrypt.openssl_version.to_s
+          UI.user_error!("Repository was encrypted using '#{repo_openssl_version}' and currently used version is '#{Encrypt.openssl_version}'")
```
joshdholtz: Is it possible that decrypting with different OpenSSL versions _may_ still work though? I have a small fear that this may error out on the user for a minor/patch OpenSSL version where decrypting may still work resulting in a false failure. How would a user handle this error when it occurs?

Which makes me think this may be better but after a decrypt fails **but** then it may not be due to OpenSSL versions but could also be due to password 🤔

Would it be better for us to store the method of encrypting instead of storing the OpenSSL version?

joshdholtz (changes requested):
This is a great PR 🎊 🎉 Thanks for doing this! Just got a quick question regarding this causing possible false failures in decrypting.

I'm thinking that it may be good to store the OpenSSL version but may be better / more backwards compatible to store the method of encryption that was used?

See https://github.com/fastlane/fastlane/issues/9542 for more info on what the root cause of the error was 😊 

Old behavior was `-md md5` and new behavior is `-md sha256`. So maybe we _also_ store `md5` or `sha256` and check based on that?

match/lib/match/git_helper.rb
viktorasl: I’m not sure if it’s possible that two different OpenSSL versions could produce different hashes, probably not likely. Not sure about OpenSSL and LibreSSL producing different hashes though.

Method of encryption does not change that frequently (i guess?). During the transiton it’s useful info, but in general not really

viktorasl:
Yeah, probably this PR does not make much sense, what do you think @joshdholtz, can it be transformed into something useful? Will it be useful to store the digest?

joshdholtz:
I think this PR could be transformed to **ALSO** store the digest 👌 I think storing the `match` version is solid but we shouldn't do stop anything because of it 😊

joshdholtz:
@viktorasl I would say put the digest in a separate file?

For openssl mismatch: I think we display a message warning the user just so they know
For digest mismatch: I don't think there is a way for us to really tell besides the decryption failing 🙃

viktorasl:
Sorry to bother @joshdholtz 
As I can see [encrypt.rb#L102](https://github.com/viktorasl/fastlane/blob/984a275a224c9e463959060fa1e7e87c99038578/match/lib/match/encrypt.rb#L102) explicitly sets the encryption method and [encrypt.rb#L127](https://github.com/viktorasl/fastlane/blob/984a275a224c9e463959060fa1e7e87c99038578/match/lib/match/encrypt.rb#L128) fallbacks to to `sha256` fixing any possible problem caused by that OpenSSL version bump.

So the effect of storing the digest would be beneficial if some day that explicit digest would be changed without any backwards compatibility. Then after failed decoding we could check if saved digest do not match.

Is it worth doing that? 🤔 Currently (and probably for a long time to the future) that digest file would not be beneficial.

viktorasl:
Yep, agree :)

PR #12187, opened 2018-03-27
Make enable_automatic_code_signing more robust to xcodeproj conf

nicolasbraun:
### Motivation and Context
In some cases, and it happens when automatic codesign was enabled, disabled and the reenabled the xcodeproj specifies the code_sign_identity for specific SDK. (tested on a clean project)

screen shot 2018-03-27 at 22 54 31

In that case the `disable_automatic_code_signing(code_sign_identity: 'iPhone Distribution')` (which is usually what you need) will not work since it will set the Distribution value for the high level paramter but not for the specific SDK

### Description
The PR enforce the requested code_sign_identity for any supported SDK (except macOS and simulators) to make it more robust toward xcodeproj configurations

### DIFF
--- a/fastlane/lib/fastlane/actions/automatic_code_signing.rb
+++ b/fastlane/lib/fastlane/actions/automatic_code_signing.rb
@@ -40,9 +40,12 @@ module Fastlane
           end
           if params[:code_sign_identity]
             build_configuration_list.set_setting("CODE_SIGN_IDENTITY", params[:code_sign_identity])
-            build_configuration_list.set_setting("CODE_SIGN_IDENTITY[sdk=*]", params[:code_sign_identity])
-            build_configuration_list.set_setting("CODE_SIGN_IDENTITY[sdk=iphoneos*]", params[:code_sign_identity])
-            build_configuration_list.set_setting("CODE_SIGN_IDENTITY[sdk=appletvos*]", params[:code_sign_identity])
+
+            # We also need to update the value if it was overriden for a specific SDK
+            build_configuration_list.build_configurations.each do |build_configuration|
+              codesign_build_settings_keys = build_configuration.build_settings.keys.select { |key| key.to_s.match(/CODE_SIGN_IDENTITY.*/) }
+              codesign_build_settings_keys.map { |key| build_configuration_list.set_setting(key, params[:code_sign_identity]) }
+            end
             UI.important("Set Code Sign identity to: #{params[:code_sign_identity]} for target: #{found_target[:name]}")
           end
           if params[:profile_name]
--- a/fastlane/lib/fastlane/actions/automatic_code_signing.rb
+++ b/fastlane/lib/fastlane/actions/automatic_code_signing.rb
@@ -40,6 +40,9 @@ module Fastlane
           end
           if params[:code_sign_identity]
             build_configuration_list.set_setting("CODE_SIGN_IDENTITY", params[:code_sign_identity])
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/automatic_code_signing.rb
```diff
@@ -40,6 +40,9 @@ def self.run(params)
           end
           if params[:code_sign_identity]
             build_configuration_list.set_setting("CODE_SIGN_IDENTITY", params[:code_sign_identity])
+            build_configuration_list.set_setting("CODE_SIGN_IDENTITY[sdk=*]", params[:code_sign_identity])
+            build_configuration_list.set_setting("CODE_SIGN_IDENTITY[sdk=iphoneos*]", params[:code_sign_identity])
+            build_configuration_list.set_setting("CODE_SIGN_IDENTITY[sdk=appletvos*]", params[:code_sign_identity])
```
joshdholtz: Is there a way to get/set these dynamically wit `xcodeproj` instead of hardcoding `sdk=*`, `sdk=iphoneos*`, and `sdk=appletvos*`?

nicolasbraun:
Not really sure what you mean josh, this relies on xcodeproj already doesn’t it ?

Sent on the road. Please excuse any typo or missing punctuation.

fastlane/lib/fastlane/actions/automatic_code_signing.rb
nicolasbraun: Hey @joshdholtz , my response by mail went in the bad thread.

Sadly I think it is not. The `build_settings` [method](http://www.rubydoc.info/gems/xcodeproj/Xcodeproj/Project/Object/XCConfigurationList#build_settings-instance_method) of the gem only works for a specific build configuration.
What we could do is 
• Loop on the configurations
• Call build_settings
• Search the ones with `CODE_SIGN_IDENTITY`
• Replace if exists

This seems a bit too much.

joshdholtz: I _think_ lopping through all configurations might be the correct approach though 🙃 This will make this more stable if/when a new platform is added and _should_ also make it work for Mac apps (which I don't think the current solution would account for with the things being hardcoded?)

nicolasbraun:
Hey @joshdholtz ,

Now looping on configs as suggested.
I was not even aware fastlane supported Mac OS :)

fastlane/lib/fastlane/actions/automatic_code_signing.rb
```diff
@@ -40,6 +40,12 @@ def self.run(params)
           end
           if params[:code_sign_identity]
             build_configuration_list.set_setting("CODE_SIGN_IDENTITY", params[:code_sign_identity])
+
+            # We also need to update the value if it was overriden for a specific SDK
+            build_configuration_list.build_configurations.each do |build_configuration|
+              codesign_build_settings_keys = build_configuration.build_settings.keys.select { |key| key.to_s.match(/CODE_SIGN_IDENTITY.*/) }
+              codesign_build_settings_keys.map { |key| build_configuration_list.set_setting(key, params[:code_sign_identity]) }
```
joshdholtz: Can you change this `map` to an `each`? This doesn't need to be map since nothing is being returned or modified on `codesign_build_settings_keys` 🙃

joshdholtz (changes requested):
One more small request 🙃

nicolasbraun:
@joshdholtz done! Thanks for 👀

joshdholtz (approved):

PR #12190, opened 2018-03-28
new action update_plist

### LINKED ISSUE #12188: [action] New action for `update_plist`
This is a follow up issue for https://github.com/fastlane/fastlane/issues/12088. 

After having a discussion with @joshdholtz we decided to add a new action called `update_plist`. The existing action `update_info_plist` is going to use this action internally.

RishabhTayal:
- [x] Added new action `update_plist`
- [x] Use the new action `update_plist` from the existing action `update_info_plist`.

fixes: #12188


RishabhTayal:
@joshdholtz could you let me know if i am moving in the right direction?

fastlane/lib/fastlane/actions/update_info_plist.rb
```diff
@@ -34,18 +34,15 @@ def self.run(params)
...
-          plist['CFBundleIdentifier'] = params[:app_identifier] if params[:app_identifier]
-          plist['CFBundleDisplayName'] = params[:display_name] if params[:display_name]
-          params[:block].call(plist) if params[:block]
-
-          # Write changes to file
-          Xcodeproj::Plist.write_to_path(plist, info_plist_path)
-
-          UI.success("Updated #{params[:plist_path]} 💾.")
-          File.read(info_plist_path)
+          other_action.update_plist(
```
joshdholtz: Instead of `other_action.update_plist`, call `UpdatePlistAction.run`. I'm fairly certain that calling `other_action` displays some console output that shows its calling another action which we will a bit confusing. We want to make this a nice clean alias to the users 👍

joshdholtz (changes requested):
Code is looking pretty good to me so far 😊

fastlane/lib/fastlane/actions/update_info_plist.rb
RishabhTayal: got it.

RishabhTayal:
@joshdholtz its ready to review 👀.

fastlane/lib/fastlane/actions/update_info_plist.rb
```diff
@@ -34,18 +34,15 @@ def self.run(params)
...
-          params[:block].call(plist) if params[:block]
-
-          # Write changes to file
-          Xcodeproj::Plist.write_to_path(plist, info_plist_path)
-
-          UI.success("Updated #{params[:plist_path]} 💾.")
-          File.read(info_plist_path)
+          UpdatePlistAction.run(
+            plist_path: info_plist_path,
+            block: lambda { |plist|
```
joshdholtz: Can we change this to use `proc do`? This is how other actions like `pem` do this so just for consistency purposes 😊 
```rb
block: proc do |plist|
  plist['CFBundleIdentifier'] = params[:app_identifier] if params[:app_identifier]
  plist['CFBundleDisplayName'] = params[:display_name] if params[:display_name]
  params[:block].call(plist) if params[:block]
end
```

fastlane/lib/fastlane/actions/update_plist.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      def self.author
+        'rishabhtayal'
+      end
+
+      def self.example_code
+        [
+          'update_plist( # Advanced processing: find URL scheme for particular key and replace value
+            plist_path: "path/to/Info.plist",
+            block: lambda { |plist|
```
joshdholtz: And also change this example to use `proc do`?

fastlane/lib/fastlane/actions/update_plist.rb
```diff
@@ -0,0 +1,87 @@
...
+        [
+
+          FastlaneCore::ConfigItem.new(key: :plist_path,
+                                       env_name: "FL_UPDATE_PLIST_PATH",
+                                       description: "Path to plist file",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :block,
+                                       is_string: false,
+                                       description: 'A block to process plist with custom logic',
+                                       optional: true)
```
joshdholtz: Should this actually be required for this action? Since it won't do anything without `block`? 🙃

joshdholtz (changes requested):
This looks really good 💯 Just a few minor comments/suggetions 😊

RishabhTayal:
@joshdholtz Updated the pr with requested changes.

RishabhTayal:
@joshdholtz This ready for 👀 again.

fastlane/lib/fastlane/actions/update_plist.rb
```diff
@@ -0,0 +1,86 @@
...
+  module Actions
+    module SharedValues
+    end
+
+    class UpdatePlistAction < Action
+      def self.run(params)
+        require 'xcodeproj'
+
+        # Check if parameters are set
+        if params[:block]
```
joshdholtz: Can we remove this since if since `block` is required now?

joshdholtz (changes requested):
Okay, ONE last change 😊 And can we also remove `require 'xcodeproj'` from `update_info_plist.rb` now since we are no longer needing it?

joshdholtz:
@RishabhTayal One bug I found was the old action errors out when only a block is set (even though it shouldn't)
```rb
update_info_plist(
      xcodeproj: 'Test.xcodeproj',
      plist_path: 'Test/Info.plist',
      block: proc do |plist|
        plist["SupDude2"] = "hellow dude2"
      end
    )
```

The error I get is...
```sh
[10:22:53]: -------------------------------
[10:22:53]: --- Step: update_info_plist ---
[10:22:53]: -------------------------------
[10:22:53]: block parameter can not be specified with app_identifier or display_name
```

🤔

joshdholtz (approved):
This is great! Thanks for doing this 😊

PR #12191, opened 2018-03-28
[WIP] remove emoji from changelog

koenpunt:
### Checklist

### Motivation and Context

When trying to submit a changelog containing emoji, the request fails with a response similar to:

```
ERROR-1: What to Test can't contain the following characters: ?.; ERROR-1: What to Test can't contain the following characters: ?.; ERROR-1: What to Test can't contain the following characters: ?.
```

### Description

Apple doesn't allow emoji to be used in the changelog value

WIP because the regex has to be updated to apply to a broader range of emoji, or maybe there's a gem that can do the heavy lifting for us.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -95,6 +95,8 @@ Gem::Specification.new do |spec|
   # If you upgrade this gem, make sure to upgrade the users of it as well.
   spec.add_dependency('google-api-client', '>= 0.13.1', '< 0.14.0') # Google API Client to access Play Publishing API
 
+  spec.add_dependency('emoji_regex', '~> 0.1') # Used to scan for Emoji in the changelog
+
   # Development only
   spec.add_development_dependency('rake', '< 12')
   spec.add_development_dependency('rspec', '~> 3.5.0')
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -131,13 +131,17 @@ module Pilot
       changelog
     end
 
-    def self.sanitize_changelog(changelog)
-      changelog = truncate_changelog(changelog)
+    def self.strip_emoji(changelog)
       if changelog && changelog =~ EmojiRegex::Regex
         changelog.gsub!(EmojiRegex::Regex, "")
         UI.important("Emoji symbols have been removed from the changelog, since they're not allowed by Apple.")
       end
       changelog
+    end
+
+    def self.sanitize_changelog(changelog)
+      changelog = strip_emoji(changelog)
+      truncate_changelog(changelog)
     end
 
     private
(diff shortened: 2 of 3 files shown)


getaaron (changes requested):
I don't think we should arbitrarily change the changelog content. How about something like:

```ruby
UI.user_error!("Disappointingly, emojis aren't allowed in the changelog 🙄") if options[:changelog].match(/[\u{1F600}-\u{1F6FF}]/)
```

joshdholtz:
@koenpunt I agree with @getaaron on this one - we tend to stay away from automatically changing things on the use. **BUT**, since you said that this was from a changeling generated from git commands, it maybe be best to add a cleanse/sanitize emojis option in the `changelog_from_git_commits` action?

koenpunt:
Since emoji cause the problem for Apple, the flag should be with the `set_changelog` and `upload_to_testflight` actions IMO.
But since we're already truncating the changelog based on size, I don't really see a problem with just replacing the emoji, especially since it otherwise doesn't work. 
We could of course still warn the user about the fact that those symbols have been removed, which is what is what also happens when the changelog is truncated.

joshdholtz:
@koenpunt That is a fair point ^ I think if you can check if there is an emoji and then warn that user (with your current implementation) that be 👌

koenpunt:
Alright, I've updated the implementation

getaaron (approved):

getaaron:
We might want to add a unit test to [build_manager_spec.rb](https://github.com/fastlane/fastlane/blob/6649d0daa6d8c51105c33d380aa526b657f12d6f/pilot/spec/build_manager_spec.rb)?

e.g.

```
  it "removes emoji" do
    changelog = "I'm 🦇B🏧an!"
    changelog = Pilot::BuildManager.truncate_changelog(changelog)
    expect(changelog).to eq("I'm Ban!")
  end

  it "truncates spaces after removing emoji at the end" do
    changelog = "Why doesn't Apple support Emoji? 🤷‍♀️"
    changelog = Pilot::BuildManager.truncate_changelog(changelog)
    expect(changelog).to eq("Why doesn't Apple support Emoji?")
  end
```

…and now that I think about it, maybe `truncate_changelog` should come after the emojis are stripped?

joshdholtz (dismissed):
Yeah yeah, I agree with @getaaron. Can we add some tests to this and call truncate after sanitize?

joshdholtz:
@koenpunt Heyyy 👋 Are you still working on this PR? 😊

getaaron:
Thanks, @koenpunt. Great work!

koenpunt:
> Thanks, @koenpunt. Great work!

You're welcome! 🎉

PR #12138, opened 2018-03-22
get_latest_version handles $(SRCROOT) and tries to get target that isn't test

joshdholtz:
## Needed to add a few more smart things to get_latest_version
- ignores `$(SRCROOT)` since `$(SRCROOT)` is directory of where the xcodeproject is
- doesn't look at test targets when trying to auto-selected target when no target name given

PR #12212, opened 2018-04-02
[action] Make sure get_version_number targets respond to test_target_type?

joshdholtz:
Fixes https://github.com/fastlane/fastlane/pull/12138#issuecomment-377970544

It turns out not all targets respond to `test_target_type?` 😇 

- Added aggregate target to test project
- Tests run against this new target and filter it out when looking for a test target


getaaron (approved):
LGTM, although this feels like a hack around a problem in xcodeproj - it should return `false` in whatever situations it's throwing an error

PR #12219, opened 2018-04-03
Fix spec in automatic_code_signing action

nicolasbraun:
Suggested values are incorrect for  development


joshdholtz (approved):
I'll accept this change 😉

PR #12239, opened 2018-04-05
Add new parameter to allow custom Fastfile path for LaneManager

KrauseFx:
Needed for https://github.com/fastlane/ci/issues/571


joshdholtz (commented):
Do we want to verify the existence of `fastfile_path`? And does this need any tests? 🙃

KrauseFx:
For fastlane.ci we don't need to verify, as we already do that https://github.com/fastlane/ci/blob/master/app/features/build_runner/fastlane_build_runner.rb#L70

No time for tests for now 🚀

joshdholtz (approved):
I will do a follow up PR that adds tests 😉

PR #12259, opened 2018-04-08
[fastlane_core][pilot][deliver] Improved transporter logging to pilot

### LINKED ISSUE #11902: Make upload_to_testflight FastlaneCore::Interface::FastlaneError messages useful
The exception message just says `Error uploading ipa file, for more information see above` which makes it hard to do any specific error handling.

cpk1:
### Checklist

#### Why is this change required? What problem does it solve?

When running Pilot to upload builds to Apple, if there is a failure with transporter there is no meaningful error message in `UI.user_error` in Pilot, this provides the errors from transporter instead of giving a generic "there was an error, look at the logs"

#### If it fixes an open issue, please link to the issue here.

Resolves https://github.com/fastlane/fastlane/issues/11902

#### Please describe in detail how you tested your changes.

I've tested uploading builds with build numbers that already existed on iTC, this correctly passed through the iTMSTransporter error message. I also tested uploading builds that had never been uploaded and everything appeared to work normally.
I have not gotten all the rspec tests to pass yet, but wanted to make sure that this was an appropriate change before moving forward

#### Describe your changes in detail

I think I've come up with a way that should get some useful messaging to pilot from transporter errors and before trying to wrap this up wanted to get the PR started to get feedback that this change would be OK, let me know!

I've added a new global in `itunes_transporter.rb` that is used to collect all the errors from transporter, this then eventually gets returned to pilot and is used as part of the message in `UI.user_error`. Also I changed some of the logging behavior as it looked like messages were getting printed twice in a row which was unnecessary.

I also changed the matching text for redundant binary upload to what apple appears to have changed it to, although we might want to consider just matching against the error code for that as that is the least likely to change over time


fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -42,6 +42,7 @@ def execute(command, hide_output)
       end
 
       @errors = []
+      @returnederrors = []
```
taquitos: `@returnederrors` -> `@returned_errors`

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -125,7 +129,7 @@ def parse_line(line, hide_output)
             CredentialsManager::AccountManager.new(user: @user).invalid_credentials
             UI.error("Please run this tool again to apply the new password")
           end
-        elsif $1.include?("Redundant Binary Upload. There already exists a binary upload with build")
+        elsif $1.include?("Redundant Binary Upload. You've already uploaded a build with build")
```
taquitos: `Redundant Binary Upload. You've already uploaded a build with build` -> `Redundant Binary Upload. You've already uploaded a build with this build number (<maybe build number here>)`

pilot/lib/pilot/build_manager.rb
```diff
@@ -34,10 +34,12 @@ def upload(options)
                                                                       platform: platform)
 
       transporter = FastlaneCore::ItunesTransporter.new(options[:username], nil, false, options[:itc_provider])
-      result = transporter.upload(app.apple_id, package_path)
+      result, transportererrors = transporter.upload(app.apple_id, package_path)
```
taquitos: `transportererrors`-> `transporter_errors`

taquitos:
Yeah, I think this approach makes sense. I don't think we use multiple return objects much, but this makes sense to me.

joshdholtz:
I'm use multiple return objects occasionally but this one I would like to be careful of since this is modifying/breaking any existing implementations and we need to make sure everything is being changed/handled properly

I search through the codebase and I think the following lines would (at least) need to be fixed...
- https://github.com/cpk1/fastlane/blob/dd7d67e4e241c42b397b7ac180accf1b334adc0d/fastlane_core/lib/fastlane_core/itunes_transporter.rb#L346
- https://github.com/cpk1/fastlane/blob/dd7d67e4e241c42b397b7ac180accf1b334adc0d/deliver/lib/deliver/runner.rb#L160

I feel like the "cleanest" solution (in our case) would  just be to **not** return the errors like this an just access them through a variable: `transporter.returned_errors`. I would probably feel much safer releasing a change like this since its not modifying any existing behaviors.

joshdholtz:
@cpk1 Are you still interested in working on this PR? 😊

cpk1:
Yes! Sorry for the delay, the warm weather has been too enticing for me :). I'll revisit and see what I can do as this is definitely functionality we are very much interested in having released.

joshdholtz:
@cpk1 No problem last all! Just wanted to make sure

I just got warm weather here so I totally understand 🙃

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -389,7 +397,11 @@ def upload(app_id, dir)
         handle_error(@password)
       end
 
-      result
+      return result
+    end
+
+    def returned_errors
```
joshdholtz: You mentioned that you would like to get rid of `@errors_to_show`. Could we rename this to `displayable_errors` and have it do something like...
```rb
@transporter_executor.map({ |error| "[Transporter Error Output]: #{error}" }).join("\n").gsub!(/"/, "")
```

cpk1: Thanks Josh, I've made the changes for this. This gets useful messaging out to slack now, however I think the logging/messaging is a bit too verbose and we are essentially duping error logs because `UI.crash!("Return status of iTunes Transporter was #{$1}: #{@errors.join('\n')}")` is essentially dumping the same message into @errors that is already there

fastlane_core/lib/fastlane_core/itunes_transporter.rb
cpk1: @taquitos Not sure if you mean the user friendly error message should have the build number? I've added that in now though. If you wanted to check specifically for the build number that is in the IPA we uploaded I'm not sure if Fastlane is unpacking the IPA to get that info at all before uploading it via transporter?

taquitos: Yeah, that message didn't make any sense to me `You've already uploaded a build with build`  would it make more sense as `You've already uploaded a build with the build action` or is it `You've already uploaded a build with a the same build number X`

janpio: Error message is still broken here @cp1. Maybe you didn't push a commit you made?

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -125,9 +129,10 @@ def parse_line(line, hide_output)
             CredentialsManager::AccountManager.new(user: @user).invalid_credentials
             UI.error("Please run this tool again to apply the new password")
           end
-        elsif $1.include?("Redundant Binary Upload. There already exists a binary upload with build")
+        elsif $1.include?("Redundant Binary Upload. You've already uploaded a build with build")
           UI.error($1)
-          UI.error("You have to change the build number of your app to upload your ipa file")
+          /build number (?<build_number>'.*') for version number (?<version_number>'.*')./ =~ $1
+          UI.error("You need to upload a binary with a build number greater than #{build_number}. Or upload a binary with a version greater than #{version_number}.")
```
janpio: This error message is not very clear. Couldn't this be combined?
```suggestion
          UI.error("You need to upload a binary with a build number greater than #{build_number} or a version greater than #{version_number}.")
```
Or isn't this correct then?

fastlane_core/lib/fastlane_core/itunes_transporter.rb
janpio: The actual error message is below, this just seems to match on one of the original error messages here to display the error message below.

janpio:
Tests are passing now, what further work is needed here @cpk1? Can we help?

cpk1:
Thanks for bringing this back up Jan! I'm pretty slammed at work but will take a look this month to see if everything is still working correctly.

janpio:
(Different name and email address on the last commit, you can probably fix this by adding that email address to the CLA)

fastlane_core/lib/fastlane_core/itunes_transporter.rb
cpk1: I'm not too hung up on exact wording but do want to make sure we don't leave any confusion for the user.
This error happens because for a given version - CFBundleShortVersionString (https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-111349) the user has uploaded a binary with a build number - CFBundleVersion (https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).

Would it be worth pointing them to the docs at all to help people understand how the two strings work?

cpk1:
I signed the CLA with the committer address.
I still need to check that this is still generating the expected output, hopefully everything is still working correctly!

janpio:
Hey @cpk1, could you take another look at this? I went through all the comments and resolved the ones that were taken care of in your commits - nothing left really. So we just need you to have another look at it and then remove the `[WIP]` from the PR so we can see if we can merge this. Thanks!

cpk1:
Thanks for following up on this Janpio. I've tested locally and the output from Fastlane is now the expected duplicate binary messaging.
I'm having some trouble running this locally testing that the slack messages get passed forward as expected but am working on testing that as well.

cpk1:
Output is as follows:
`INFO [2019-05-16 12:15:27.92]: iTunes Transporter output above ^
ERROR [2019-05-16 12:15:27.92]: ERROR ITMS-90189: "Redundant Binary Upload. You've already uploaded a build with build number '34' for version number '1.0'. Make sure you increment the build string before you upload your app to App Store Connect. Learn more in Xcode Help (http://help.apple.com/xcode/mac/current/#/devba7f53ad4)."
Return status of iTunes Transporter was 1: ERROR ITMS-90189: "Redundant Binary Upload. You've already uploaded a build with build number '34' for version number '1.0'. Make sure you increment the build string before you upload your app to App Store Connect. Learn more in Xcode Help (http://help.apple.com/xcode/mac/current/#/devba7f53ad4)."
The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure.`

Not sure if we would want to reduce this to not be so repetitive?

janpio:
It's not super pretty, but better than not showing it at all. I would be good with it, if you don't have an obvious fix in mind that you can implement quickly.

cpk1:
I think ideally we would prevent the error from making it to here? https://github.com/cpk1/fastlane/blob/cpk1-transporterLogging/fastlane_core/lib/fastlane_core/itunes_transporter.rb#L81-L82

That way the user just sees the " iTunes Transporter output above ^" and then maybe "The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure." (which looking at it now should technically not be part of the transporter error message that is getting generated).

I think ideally the log would look as follows:

> ...transporter output...
> [11:22:22]: iTunes Transporter output above ^
> The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure.
> 
> [!] Error uploading ipa file: 
> [Transporter Error Output]: Return status of iTunes Transporter was 1: ERROR ITMS-90189: Redundant Binary Upload. You've already uploaded a build with build number '34' for version number '1.0'. Make sure you increment the build string before you upload your app to App Store Connect. Learn more in Xcode Help (http://help.apple.com/xcode/mac/current/#/devba7f53ad4).

The current treatment is definitely not ideal but will at least get the message across. Could make another issue to clean up the messaging?

janpio:
Of course! Just create the issue and describe what you want to improve, then submit a PR when you are done with it. Anything that improves the error output is more than welcome.

cpk1:
Sounds good. Let me know if anything else needs to be done to get this merged in.

janpio (approved):
Good from my side.

What do you think @taquitos and @joshdholtz? Safe to merge?

cpk1:
Happy Friday @taquitos and @joshdholtz . Wanted to see if we can still get this merged in? I was a little worried that all the app store connect API changes might make this less useful but it doesn't seem like the way builds get uploaded will change any time soon so I think this still adds value for the foreseeable future.

janpio:
Looking into this again right now and as a first action cleaned up the comments a bit. Now on to having another look at the code.

janpio (changes requested):
Ooookay @cpk1, this comes super late and I hope this doesn't come over as unnecessary, buuuut it would be super great if you could split the changes at https://github.com/fastlane/fastlane/pull/12259/files#diff-03ad193b71c4d81874c287fcd4bedcc5L120-R126 into their own PR. 

Those are not really related to this PR, and as both parts high risk ones (very possible we do break existing behavior someone depends on here) I would like to have these as separate commits in the main codebase (which means they have to be separate PRs) so we can revert as needed.

With these changes extracted to their own PR, I am pretty sure we can just merge both and see what happens <3

cpk1:
@janpio I know I'm coming back into this super late, but ready to do the work to release this now, are you saying you want lines 120-126 extracted to their own PR?

janpio:
Yes, I think that is what I meant - it would move this high risk change into its own PR that we can merge first to make sure nothing breaks. Do you agree with that?

cpk1:
Totally makes sense. I'm still trying to make sure I'm set up to test correctly but it looks like this is working other than the ITMS error double printing.

Also, I think the apple message is plenty helpful on its own that we might not need to provide any extra insight?

AlphaJuliettOmega:
@janpio can you make the changes? it's been a long time since you asked @cpk1

cpk1:
@AlphaJuliettOmega thanks for the nudge. I'll take a look at it today.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
@cpk1 Thank you _sooooooo_ much for making this PR! My apologies on taking so long to get this merged 😔 I ended up adding a commit to also add this logic to `deliver`. I really appreciate this contribution ❤️

cpk1:
Thank you! I promise I won't let my next PR languish for so long!

PR #12272, opened 2018-04-09
Fix action info table not rendering correctly in docs

revolter:
Before:

screen shot 2018-04-09 at 23 16 18

After:

screen shot 2018-04-09 at 23 16 26

I didn't find a way to make the header cell span 2 columns.

### DIFF
--- a/fastlane/lib/fastlane/actions/commit_github_file.rb
+++ b/fastlane/lib/fastlane/actions/commit_github_file.rb
@@ -153,7 +153,7 @@ module Fastlane
         [
           "A hash containing all relevant information for this commit",
           "Access things like 'html_url', 'sha', 'message'"
-        ].join("<br>")
+        ].join("\n")
       end
 
       def self.authors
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/commit_github_file.rb
```diff
@@ -153,7 +153,7 @@ def self.return_value
         [
           "A hash containing all relevant information for this commit",
           "Access things like 'html_url', 'sha', 'message'"
-        ].join("\n")
+        ].join("<br>")
```
joshdholtz: This needs to stay a `\n` for terminal output 🙃 

screen shot 2018-04-10 at 5 34 17 am

fastlane/lib/fastlane/actions/set_github_release.rb
```diff
@@ -236,7 +236,7 @@ def self.return_value
         [
           "A hash containing all relevant information of this release",
           "Access things like 'html_url', 'tag_name', 'name', 'body'"
-        ].join("\n")
+        ].join("<br>")
```
joshdholtz: Same here 😇

joshdholtz (changes requested):

fastlane/lib/fastlane/actions/commit_github_file.rb
revolter: Damn :( Thanks for catching this! This means I need to `gsub` `\n` with `<br>` in the `erb`, right? I'll test it tonight.

joshdholtz (approved):
💯 💪

PR #12274, opened 2018-04-09
[scan] suppress xcode output and also open report if tests fail

revolter:
This PR:

- adds these options to the _scan_ action:
  - `suppress_xcode_output` that defaults to `false` and prevents printing the build's output when set to `true`
  - `app_identifier` and `reinstall_app` that defaults to `false` and reinstalls the app with the specified, auto-detected or typed `app_identifier` when set to `true`
- opens the report (if the `open_report` option is active) when tests fail too


scan/lib/scan/options.rb
```diff
@@ -292,7 +297,19 @@ def self.available_options
...
-                                     is_string: true)
+                                     is_string: true),
+        FastlaneCore::ConfigItem.new(key: :app_identifier,
+                                     env_name: 'SCAN_APP_IDENTIFIER',
+                                     optional: true,
+                                     description: "The bundle identifier of the app to uninstall (only needed when enabling reinstall_app)",
+                                     code_gen_sensitive: true,
+                                     default_value: ENV["SCAN_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
+                                     default_value_dynamic: true),
+        FastlaneCore::ConfigItem.new(key: :reinstall_app,
```
joshdholtz: Instead of making `app_identifier` and `reinstall_app`, can we make this one option? Can we call it something like `reinstall_with_app_identifier`? This makes usage, doc, and code a bit cleaner to combine this into one (especially if `app_identifier` is only to be used with `reinstall_app`)

scan/lib/scan/runner.rb
```diff
@@ -39,11 +39,20 @@ def test_app
         }
       ]
       exit_status = 0
+
+      Scan.devices.each do |device|
+        app_identifier = Scan.config[:app_identifier]
+        app_identifier ||= UI.input("App Identifier: ")
```
joshdholtz: This `UI.input` would end up prompting for 1 to many times depending on the number of devices that are being iterated through here which probably isn't ideal. But I think this can be solved with the comment about combining the two new options into `reinstall_with_app_identifier` (mentioned above)

joshdholtz (changes requested):
Left some thoughts/suggestions 😊

scan/lib/scan/options.rb
revolter: I copied this from _snapshot_:

https://github.com/fastlane/fastlane/blob/665a23e6351702bbd6980fe1f7ec75a6f1f7bfa6/snapshot/lib/snapshot/options.rb#L119-L126

https://github.com/fastlane/fastlane/blob/665a23e6351702bbd6980fe1f7ec75a6f1f7bfa6/snapshot/lib/snapshot/options.rb#L104-L108

The question is if we want to live it like this to be consistent, make it simpler like you said, and/or make it like this for _snapshot_ too in the future which will get us hundreds of issues.

scan/lib/scan/runner.rb
revolter: This can be solved by me not being stupid 😂

scan/lib/scan/options.rb
joshdholtz: @revolter Oh maaannnn, I didn't know _snapshot_ had it formatted like that 🤔 I don't think we can change _snapshot_ without it breaking some existing setups so I maybe we should keep this similar for consistency 😬

scan/lib/scan/options.rb
```diff
@@ -292,7 +297,19 @@ def self.available_options
...
                                      conflicting_options: [:output_files],
                                      optional: true,
-                                     is_string: true)
+                                     is_string: true),
+        FastlaneCore::ConfigItem.new(key: :app_identifier,
+                                     env_name: 'SCAN_APP_IDENTIFIER',
+                                     optional: true,
+                                     description: "The bundle identifier of the app to uninstall (only needed when enabling reinstall_app)",
+                                     code_gen_sensitive: true,
+                                     default_value: ENV["SCAN_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
```
joshdholtz: Is `ENV["SCAN_APP_IDENTITIFER"]` spelled wrong and in `default_value` for a specific reason? 🤔

scan/lib/scan/runner.rb
```diff
@@ -39,11 +39,20 @@ def test_app
         }
       ]
       exit_status = 0
+
+      app_identifier = Scan.config[:app_identifier]
+      app_identifier ||= UI.input("App Identifier: ")
+
+      Scan.devices.each do |device|
+        FastlaneCore::Simulator.uninstall_app(app_identifier, device.name, device.udid) if Scan.config[:reinstall_app]
```
joshdholtz: This `if Scan.config[:reinstall_app]` should go before the `Scan.devices.each` so we don't have to iterate over devices if we don't have to 😊

joshdholtz (changes requested):
Tested these changes out! Things work great 👍 Just two small comments on things 🙃

scan/lib/scan/options.rb
revolter: Because copy paste :)) I copied it from here https://github.com/fastlane/fastlane/blob/2d08ade8e28670b49b1f8abfe6121affb4905092/snapshot/lib/snapshot/options.rb#L125 I'll send another PR to fix that.

revolter: Done: https://github.com/fastlane/fastlane/pull/13093

scan/lib/scan/options.rb
```diff
@@ -292,7 +297,19 @@ def self.available_options
...
                                      conflicting_options: [:output_files],
                                      optional: true,
-                                     is_string: true)
+                                     is_string: true),
+        FastlaneCore::ConfigItem.new(key: :app_identifier,
+                                     env_name: 'SCAN_APP_IDENTIFIER',
+                                     optional: true,
+                                     description: "The bundle identifier of the app to uninstall (only needed when enabling reinstall_app)",
+                                     code_gen_sensitive: true,
+                                     default_value: ENV["SCAN_APP_IDENTIFIER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
```
revolter: @joshdholtz, After the conversation from the other PR, this can be removed, right?

scan/lib/scan/options.rb
```diff
@@ -303,7 +303,6 @@ def self.available_options
                                      optional: true,
                                      description: "The bundle identifier of the app to uninstall (only needed when enabling reinstall_app)",
                                      code_gen_sensitive: true,
-                                     default_value: ENV["SCAN_APP_IDENTIFIER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
```
joshdholtz: @revolter Did you not want to keep the `CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier)`? 🤔

scan/lib/scan/options.rb
```diff
@@ -146,6 +146,11 @@ def self.available_options
                                      type: Boolean,
                                      default_value: false,
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :suppress_xcode_output,
+                                     env_name: "SUPPRESS_OUTPUT",
```
joshdholtz: Can we change this to be `SUPPRESS_XCODE_OUTPUT` to make it match the key? 😊

joshdholtz (changes requested):
Onnnneeeee more small request 😉

scan/lib/scan/options.rb
revolter: Most probably I copied it from another place and forgot to update it 😃

revolter:
@joshdholtz, Bump 😃

scan/lib/scan/options.rb
```diff
@@ -146,6 +146,11 @@ def self.available_options
                                      type: Boolean,
                                      default_value: false,
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :suppress_xcode_output,
+                                     env_name: "SUPPRESS_XCODE_OUTPUT",
```
joshdholtz: This `env_name` should be prefix with "SCAN_" like the rest of the envs 😊 so I think it should be `SCAN_SUPPRESS_XCODE_OUTPUT`

joshdholtz (changes requested):
Okay, this last thing should do it! 😊

scan/lib/scan/options.rb
revolter: Nice catch, thanks!

jineshqa:
Bump

joshdholtz (approved):
We gooooood! ❤️ Thanks for the PR 😊

PR #12286, opened 2018-04-11
[new action] - app_store_build_info

### LINKED ISSUE #12221: app_store_build_number should return version number too
### New Issue Checklist

### Issue Description

The `app_store_build_number` action has an output variable for the build number, but it would also be great if it had an output variable for the version number. This wouldn't be useful when you pass a version number to the action, but it would be useful when using `live: true`.

The context here is I'd like to have a lane that creates a git tag for the current live version (we already have a way to go from build number -> commit), and this obviously requires knowing what the version number is. The idea is we'd run this after publishing a new version.

RishabhTayal:
fixes: #12221


joshdholtz:
@RishabhTayal Can you remove the `[WIP]` if this is no longer a WIP so that we know we can review it? 😊

RishabhTayal:
Will do when its complete. It's still wip.

RishabhTayal:
@joshdholtz This is ready for review.

RishabhTayal:
@joshdholtz can you tell why I am getting `Action 'app_store_build_info' doesn't use the option :team_id` error on test cases?

joshdholtz:
@RishabhTayal Can you run rubocop on this?

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -0,0 +1,195 @@
...
+          'app_store_build_info(
+            live: false,
+            app_identifier: "app.identifier",
+            version: "1.2.9"
+          )'
+        ]
+      end
+
+      def self.authors
+        ["rishabhtayal"]
```
joshdholtz: Can you included `hjanuschka` in here as well? Since that is where this code was originally derived from? 😊

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -0,0 +1,195 @@
...
+              build_nr = params[:initial_build_number]
+            end
+          rescue
+            UI.user_error!("Could not find a build on iTC - and 'initial_build_number' option is not set") unless params[:initial_build_number]
+            build_nr = params[:initial_build_number]
+          end
+        end
+        UI.message("Latest upload for version #{version_number} is build: #{build_nr}")
+
+        build_nr
```
joshdholtz: Probably should return a dictionary of info here? Build number, version number, etc?

joshdholtz (changes requested):

fastlane/lib/fastlane/actions/app_store_build_info.rb
RishabhTayal: Just returning the same value as before. Other values can be accessed via SharedValues.

joshdholtz: I'm thinking since this is called `app_store_build_info` it should return a dictionary of all the info and then `app_store_build_number` should pull the build number from that. Having `app_store_build_info` only returning the build number just doesn't seem right 😇

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -180,7 +180,7 @@ def self.example_code
       end
 
       def self.authors
-        ["rishabhtayal"]
+        ["hjanuschka"]
```
joshdholtz: You can keep your name here too since you changed stuff! But just wanted to keep `hjanuschka` in there too 😉

joshdholtz (changes requested):
Left some comments!

fastlane/lib/fastlane/actions/app_store_build_info.rb
RishabhTayal: done.

RishabhTayal:
@joshdholtz Ready for review.

fastlane/lib/fastlane/actions/app_store_build_number.rb
```diff
@@ -1,76 +1,13 @@
...
-          end
-        end
-        UI.message("Latest upload for version #{version_number} is build: #{build_nr}")
-
-        build_nr
-      end
-
-      def self.order_versions(versions)
-        versions.map(&:to_s).sort_by { |v| Gem::Version.new(v) }
+        build_number = AppStoreBuildInfoAction.run(params)
```
joshdholtz: `AppStoreBuildInfoAction` now returns a hash to this needs to look for `latest_build_number` within that hash 🙃

joshdholtz (changes requested):

fastlane/lib/fastlane/actions/app_store_build_number.rb
RishabhTayal: Good catch. Made this change.

joshdholtz:
@RishabhTayal Can you rebase this from master? There have been some changes in tests that should fix appveyor from failing 🙃

purswanihemant:
@joshdholtz done. The tests are passing now. ✅

RishabhTayal:
@joshdholtz done. The tests are passing now. ✅

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -0,0 +1,198 @@
...
+    end
+
+    class AppStoreBuildInfoAction < Action
+      def self.run(params)
+        require 'spaceship'
+
+        build_nr = get_build_number(params)
+
+        # Convert build_nr to int (for legacy use) if no "." in string
+        if build_nr.kind_of?(String) && !build_nr.include?(".")
```
joshdholtz: @RishabhTayal `build_nr` is a hash here so this logic should probably get reworked to lookup from the hash properly

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -0,0 +1,198 @@
...
+          build_nr = build_nr.to_i
+        end
+
+        Actions.lane_context[SharedValues::LATEST_BUILD_NUMBER] = build_nr
+      end
+
+      def self.get_build_number(params)
```
joshdholtz: This now gets build into so it should probably get renamed to `get_build_info`?

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -0,0 +1,198 @@
...
+        UI.message("Login successful")
+
+        platform = params[:platform]
+
+        app = Spaceship::Tunes::Application.find(params[:app_identifier])
+        if params[:live]
+          UI.message("Fetching the latest build number for live-version")
+          live_version = app.live_version
+          build_nr = live_version.current_build_number
+          Actions.lane_context[SharedValues::LATEST_VERSION_NUMBER] = live_version.version
```
joshdholtz: Also, `version_number` only gets set in the `else` in this if statement. We probably need to add a line in this first part of this app that is `version_number = live_version.version`

joshdholtz (changes requested):
Found some things 😊

fastlane/lib/fastlane/actions/app_store_build_info.rb
RishabhTayal: added.

fastlane/lib/fastlane/actions/app_store_build_info.rb
RishabhTayal: renamed

fastlane/lib/fastlane/actions/app_store_build_info.rb
RishabhTayal: looking up from the hash now.

RishabhTayal:
@joshdholtz Could you please take a look at the failing test case. Not sure why it's failing. 🤔

fastlane/spec/actions_specs/app_store_build_info_spec.rb
```diff
@@ -1,42 +1,42 @@
...
         versions = [3, '1', '2.3', 9, '6.5.4', '11.4.6', 5.6]
-        result = Fastlane::Actions::AppStoreBuildNumberAction.order_versions(versions)
+        result = Fastlane::Actions::AppStoreBuildInfoAction.order_versions(versions)
 
         expect(result).to eq(['1', '2.3', '3', '5.6', '6.5.4', '9', '11.4.6'])
       end
 
       it "returns value as string (with build number as version string)" do
-        allow(Fastlane::Actions::AppStoreBuildNumberAction).to receive(:get_build_number).and_return("1.2.3")
+        allow(Fastlane::Actions::AppStoreBuildInfoAction).to receive(:get_build_number).and_return("1.2.3")
```
joshdholtz: Test are failing because `get_build_number` was changed to `get_build_info`

fastlane/spec/actions_specs/app_store_build_info_spec.rb
```diff
@@ -1,42 +1,42 @@
...
-          app_store_build_number(username: 'name@example.com', app_identifier: 'x.y.z')
+          app_store_build_info(username: 'name@example.com', app_identifier: 'x.y.z')
         end").runner.execute(:test)
 
         expect(result).to eq("1.2.3")
       end
 
       it "returns value as integer (with build number as version number)" do
-        allow(Fastlane::Actions::AppStoreBuildNumberAction).to receive(:get_build_number).and_return("3")
+        allow(Fastlane::Actions::AppStoreBuildInfoAction).to receive(:get_build_number).and_return("3")
```
joshdholtz: Same here - Test are failing because `get_build_number` was changed to `get_build_info`

joshdholtz (changes requested):

RishabhTayal:
@joshdholtz this is ready for review again. 🤞

joshdholtz (changes requested):
Can actually add some of these tests back in? Like the last 3 so that we know that this actually still works like it was supposed to 🤔

joshdholtz (changes requested):
Found a few more things 😊

fastlane/lib/fastlane/actions/app_store_build_info.rb
```diff
@@ -0,0 +1,206 @@
...
+              build_nr = params[:initial_build_number]
+            end
+          rescue
+            UI.user_error!("Could not find a build on iTC - and 'initial_build_number' option is not set") unless params[:initial_build_number]
+            build_nr = params[:initial_build_number]
+          end
+        end
+        UI.message("Latest upload for version #{version_number} is build: #{build_nr}")
+
+        [
```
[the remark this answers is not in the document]
RishabhTayal: swift habits 🤷‍♂️

RishabhTayal:
Closing this. Added this as a plugin: https://github.com/RishabhTayal/fastlane-plugin-appstore_build_info

PR #12290, opened 2018-04-12
Fix force output files on trade_representative_contact_information

mofneko:
### Checklist

### Description

In my case, when creating a new application, iTC had my developer profile entered by default. However, because postal code was not registered in the item of trade_representative_contact_information, only postal_code.txt was not created in the process of deliver init and download_metadata.
If some information is missing like this, the generated file will also become fragmented, so I modified it.


joshdholtz (approved):
Thanks for making this fix! 💪 The other metadata sections behave the same by making an empty file so I think this is the correct way to go for this as well 👍

mofneko:
You’re welcome😊!
I am also pleased to contribute to your great product🍻

PR #12311, opened 2018-04-15
[spaceship] add preorder to availability

### LINKED ISSUE #12283: update_availability is not working. Apple has changed mandatory fields.
### New Issue Checklist

### Issue Description
Getting ** ~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/tunes_client.rb:227:in `handle_itc_response': Your app must be marked as cleared for pre-order to modify available date. (Spaceship::Tunes::Error)** while trying to change available countries.

Code
```ruby
	app = Spaceship::Tunes::Application.find(bundle_id)

	availability = app.availability
	availability.include_future_territories = true
	availability = Spaceship::Tunes::Availability.from_territories(["FR", "DK", "GB", "IE", "UA", "SE", "DE", "AT"])
	app.update_availability!(availability)
```

<details>
  <pre>~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/tunes_client.rb:227:in `handle_itc_response': Your app must be marked as cleared for pre-order to modify available date. (Spaceship::Tunes::Error)
[12:36:16]: ▸ from ~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/tunes_client.rb:611:in `update_availability!'
[12:36:16]: ▸ from ~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/application.rb:261:in `update_availability!'
[12:36:16]: ▸ from usefull_scripts/submit_to_review.rb:31:in `<main>'
[12:36:16]: Exit status of command 'ruby usefull_scripts/submit_to_review.rb com.xxx.xxx' was 1 instead of 0.
[12:35:45]: Login to iTunes Connect (iwaiterdeveloper@gmail.com)
~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/tunes_client.rb:227:in `handle_itc_response': Your app must be marked as cleared for pre-order to modify available date. (Spaceship::Tunes::Error)
	from ~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/tunes_client.rb:611:in `update_availability!'
	from ~/.rvm/gems/ruby-2.4.1/gems/fastlane-2.89.0/spaceship/lib/spaceship/tunes/application.rb:261:in `update_availability!'
	from usefull_scripts/submit_to_review.rb:31:in `<main>'</pre>
</details>

Apple added `"preOrder"=>{"clearedForPreOrder"=>{"value"=>false, "isEditable"=>false, "isRequired"=>false, "errorKeys"=>nil}` to "ra/apps/#{app_id}/pricing/intervals" so I fixed this error by adding 
`data["preOrder"] = {"clearedForPreOrder"=>{"value"=>false, "isEditable"=>false, "isRequired"=>false, "errorKeys"=>nil}}` to update_availability! method of TunesClient but it's good to have it updated in fastlane. Thanks.

joshdholtz:
Fixes #12283

screen shot 2018-04-15 at 8 42 00 am

## Issue
- Apple add `preorder` to to the `pricing/intervals` response/request
- It was breaking for users that never "used" it before
  - If `clearedForPreorder` was false for all new users
  - But `appAvailableDate` _could_ have the date of the late release
  - This was causing our requests to 💣 since we were the response from `pricing/intervals` as the next `request`

## Fixes
- Allows user to set `cleared_for_preorder` and `app_available_date`
- `appAvailableDate` will always be set to `null` if `cleared_for_preorder` is `false`
  - Because this is what the iTC API expects (comments for this are in code)


spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -26,12 +32,33 @@ def self.from_territories(territories = [], params = {})
...
       def territories
         @territories ||= raw_data['countries'].map { |info| Territory.new(info) }
       end
+
+      def cleared_for_preorder
+        return @cleared_for_preorder unless @cleared_for_preorder.nil?
+
+        value = false
+        if (pre_order = raw_data['preOrder']) && (hash = pre_order['clearedForPreOrder'])
+          value = hash['value']
```
snatchev: Could this `value` lookup leverage the `attr_mapping` ability to drill down to the value as shown here? https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/tunes/app_details.rb#L47

spaceship/spec/tunes/availability_spec.rb
```diff
@@ -68,6 +68,28 @@
         expect(availability.territories.length).to eq(1)
         expect(availability.territories[0].code).to eq("US")
       end
+
+      it "correctly sets preorder with no app available date" do
+        TunesStubbing.itc_stub_set_preorder_cleared
```
snatchev: Nice. We're using the older-style stubbing here, but it's close to the test site 👍

snatchev (approved):
Looks good to merge.

I would recommend you leverage the `attr_mapping` to grab the `preOrder. clearedForPreOrder.value` for `cleared_for_preorder` and similar for `app_available_date`. Maybe in another refactor PR?

If you need to perform some logic like making sure a nil value returns false then you can override the getter like:

```ruby
def cleared_for_preorder
  super || false
end
```

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -605,6 +605,17 @@ def update_availability!(app_id, availability)
...
 
+      # InitializespreOrder (if needed)
+      data["preOrder"] ||= {}
+
+      # Sets app_available_date to nil if cleared_for_preorder if false
+      # This is need for apps that have never set either of these before
+      # API will error out if cleared_for_preorder is false and app_available_date has a date
+      cleared_for_preorder = availability.cleared_for_preorder
+      app_available_date = cleared_for_preorder ? availability.app_available_date : nil
+      data["preOrder"]["clearedForPreOrder"] = { "value" => cleared_for_preorder, "isEditable" => true, "isRequired" => true, "errorKeys" => nil }
```
snatchev: I believe if you use the `attr_mapping` you don't make to create these hashes, they already exist on the model in this format.

joshdholtz:
@snatchev I don't think we can _actually_ use the `attr_mapping` for this object because the current implementation of `Availability` requires us to modify an attribute for when we "update" it. If we use `attr_mapping`, our setter gets removed (https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/base.rb#L146) and we cannot use the existing implementation to modify an attribute (as easily) for updating.

Let me know if I'm misunderstanding something but I think we need to keep this same implementation **for now** but I will make sure to put comments as to why we aren't using `attr_mapping` 😬

snatchev:
@joshdholtz it's a bit subtle but that method removal happens to any method defined *before* `attr_mapping` gets called (in practice, this is the methods created by attr_accessor, et al.)

If we define a method *after* `attr_mapping`, then `super` will reference the method created by the mapping.

joshdholtz:
@snatchev Okay, fixed it! This was the thing I was missing that was making the tests fail 😱 https://github.com/fastlane/fastlane/pull/12311/commits/03c36aaeb6534a8ab0bbdcd991e87374bdfae31e#diff-149054f84fcd8657fb1c37faab9dcc02R28

PR #12338, opened 2018-04-18
Move FastlaneCore::UpdateChecker to after dot_env load

### LINKED ISSUE #10163: setting FASTLANE_SKIP_UPDATE_CHECK env variable does not prevent update check
### New Issue Checklist

### Issue Description

[INSERT OUTPUT HERE]

### Environment

Please run `fastlane env` and copy the output below. This will help us help you :+1:
If you used `--capture_output` option please remove this block - as it is already included there.

[INSERT OUTPUT HERE]

joshdholtz:
Fixes #10163

🙂

### DIFF
--- a/fastlane/lib/fastlane/cli_tools_distributor.rb
+++ b/fastlane/lib/fastlane/cli_tools_distributor.rb
@@ -54,13 +54,14 @@ module Fastlane
           end
         end
 
-        FastlaneCore::UpdateChecker.start_looking_for_update('fastlane')
-
         # Loading any .env files before any lanes are called since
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Do you think this will have any side effects?

joshdholtz:
@KrauseFx It has a positive side effect of allowing `FASTLANE_SKIP_UPDATE_CHECK` to be used in a `.env` file 😀 But I couldn't find anyway when testing or looking through the code. I almost felt like there would have been a weird side effect before this change due to environment variables being loaded while `FastlaneCore::UpdateChecker` was doing some forking 🙃

PR #12364, opened 2018-04-23
Add support for hidding Ruby backtraces

UnsafePointer:
### Checklist

### Motivation and Context
This is probably going to be a bit controversial, but for teams where not everyone is a Ruby developer, the Ruby backtrace has very little value and makes reading the output and intimidating task. I'd like to hide the Ruby backtrace from CI output, so developers can feel less intimidated by Ruby while focusing on the problems with their iOS projects.

### Description
Setting FASTLANE_HIDE_BACKTRACE environment variable will hide all Ruby exceptions backtraces.

### DIFF
--- a/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
+++ b/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
@@ -269,7 +269,8 @@ module Commander
     end
 
     def reraise_formatted!(e, message)
-      raise e, "[!] #{message}".red, e.backtrace
+      backtrace = FastlaneCore::Env.truthy?("FASTLANE_HIDE_BACKTRACE") ? [] : e.backtrace
+      raise e, "[!] #{message}".red, backtrace
     end
 
     def show_github_issues(message_or_error)


lacostej:
If we still use commander at the base, and if commander is tracing, commander has a feature to disable tracing 

https://github.com/commander-rb/commander/blob/master/lib/commander/runner.rb#L64

Ruenzuo:
@lacostej I noticed that, but we are overriding the `run!` method and we don't provide support for that feature there: https://github.com/Ruenzuo/fastlane/blob/68e1a9b438ee2a68637f0929c542b01e4169e1ea/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb#L29-L53, so I guess this wouldn't work? That's why I went for an environment variable and the central place where we format crashes.

Ruenzuo:
@KrauseFx hey Felix, any thoughts on this? 😄

KrauseFx:
cc @joshdholtz who will make the decision on this one 👍

joshdholtz (approved):
I can be cool with this 😊 Tested this out works as expected for both on and off 👌 Thanks for the contribution!

PR #12370, opened 2018-04-24
[verify_build] App build path support

mgrebenets:
### Checklist

### Motivation and Context
Verifying an `.ipa` file is default use case, but it is often handy to verify unpackaged `.app` file as well.
Adding support for more generic "build path" allows to add support for validating `.xcarchive` files in the future without API changes.

### Description
Add new `build_path` option and use it to pass in path to `.app` package.


fastlane/lib/fastlane/actions/verify_build.rb
```diff
@@ -21,14 +21,18 @@ def self.run(params)
...
+        build_path = File.expand_path(build_path)
 
-        `unzip '#{ipa_path}' -d #{dir}`
-        UI.user_error!("Unable to unzip ipa") unless $? == 0
+        if File.extname(build_path) =~ /\.(ipa|zip)/
+          `unzip #{build_path} -d #{dir} -x '__MACOSX/*' '*.DS_Store'`
+          UI.user_error!("Unable to unzip ipa") unless $? == 0
 
-        app_path = File.expand_path("#{dir}/Payload/*.app")
+          app_path = Dir["#{dir}/**/Payload/*.app"].first
```
joshdholtz: This change is breaking my test that worked on master 😱 

### Before `app_path` was
```
/var/folders/19/63rs8ddd675gw909cd4y67qc0000gn/T/d20180425-99925-3w00ed/Payload/*.app
```

### With this change `app_path` is
```
app_path: /var/folders/19/63rs8ddd675gw909cd4y67qc0000gn/T/d20180425-99961-eojm3n/Payload/Some Test App.app
```

fastlane/lib/fastlane/actions/verify_build.rb
```diff
@@ -21,14 +21,18 @@ def self.run(params)
...
-        UI.user_error!("Unable to find ipa file '#{ipa_path}'.") unless File.exist?(ipa_path)
-        ipa_path = File.expand_path(ipa_path)
+        build_path = params[:ipa_path] || params[:build_path] || Actions.lane_context[SharedValues::IPA_OUTPUT_PATH] || ''
+        UI.user_error!("Unable to find file '#{build_path}'") unless File.exist?(build_path)
+        build_path = File.expand_path(build_path)
 
-        `unzip '#{ipa_path}' -d #{dir}`
-        UI.user_error!("Unable to unzip ipa") unless $? == 0
+        if File.extname(build_path) =~ /\.(ipa|zip)/
+          `unzip #{build_path} -d #{dir} -x '__MACOSX/*' '*.DS_Store'`
```
joshdholtz: Can we re-add single quotes around `#{build_path}`? This breaks if there is a space in the name/path

fastlane/lib/fastlane/actions/verify_build.rb
joshdholtz: The reason this broke my existing test case is because of the spaces in the name. The fix is either to revert back to using `File.expand_path` or add single quotes around the path component of these commands...

https://github.com/mgrebenets/fastlane/blob/8107884814c963c6c3d9f2ca5860ca5b660a8a90/fastlane/lib/fastlane/actions/verify_build.rb#L41

and

https://github.com/mgrebenets/fastlane/blob/8107884814c963c6c3d9f2ca5860ca5b660a8a90/fastlane/lib/fastlane/actions/verify_build.rb#L64

joshdholtz (changes requested):
I like the addition of this here! Just found a few issues with test cases that involve names with spaces 🙃 I'm a monster 🦖

fastlane/lib/fastlane/actions/verify_build.rb
mgrebenets: I see, I'll add a test case where `.app` path inside `.ipa` contains spaces 👍

mgrebenets: So did it work before because `*` was carried over and was then expanded when used in the shell scripts?
I guess that's what got me confused...

Looks like it was the case. It would "think" there is an `.app` file because `app_path` wasn't `nil`, but then when trying to run `codesign` on it, it would fail mostly because I forgot to shellescape the argument 😀 

Now when I'm adding proper shell escaping I get errors like "can't call `shellescape` on `nil`"...

Anyways, fixing it now 👍

joshdholtz (approved):
OMG, those tests for IPA and App with spaces 😱 Love it! This is 💯

fastlane/lib/fastlane/actions/verify_build.rb
```diff
@@ -21,20 +21,24 @@ def self.run(params)
...
-
-        app_path = File.expand_path("#{dir}/Payload/*.app")
+        build_path = params[:ipa_path] || params[:build_path] || Actions.lane_context[SharedValues::IPA_OUTPUT_PATH] || ''
+        UI.user_error!("Unable to find file '#{build_path}'") unless File.exist?(build_path)
+        build_path = File.expand_path(build_path)
+
+        if File.extname(build_path) =~ /\.(ipa|zip)/
+          `unzip #{build_path.shellescape} -d #{dir.shellescape} -x '__MACOSX/*' '*.DS_Store'`
+          UI.user_error!("Unable to unzip ipa") unless $? == 0
+          # Adding extra ** for edge-case ipas where Payload directory is nested.
```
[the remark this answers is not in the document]
mgrebenets: Honestly, I don't think it's a thing either...
For some reason the `ContainsWatchApp.ipa` has `ContainsWatchApp/Payload/` inside.
I now suspect that it's just incorrectly zipped fixture `.ipa` file, I'll fix it up later.

PR #12378, opened 2018-04-25
[scan] add option for zipping build productions

joshdholtz:
## Motivation
- Adding option to zip build products that is often gets sent to remove services to test

## Usage
New option `: should_zip_derived_data` will place a `build_products.zip` (a zip of `<derived_data/Products/Build>` in the outputs directory 

### CLI
```sh
$: fastlane scan -Z
```

### Fastfile
```rb
lane :test do
  scan(should_zip_derived_data: true)
end
```

## To Test This Change
Add the following to your Gemfile and then run `bundle install`
```rb
gem 'fastlane', :git => 'https://github.com/fastlane/fastlane.git', :branch => 'joshdholtz-scan-zip-derived-data'
```

Run CLI command with `bundle exec fastlane scan -Z`


hjanuschka:
💯  looks awesome, should we add this to `gym`  and|or `xcodebuild` too?

joshdholtz:
@hjanuschka I'm down for moving this into a helper! And is this a thing that `gym` or `xcodebuild` could actually use? I'm actually don't know 🙃

hjanuschka:
@joshdholtz 
how about: 
  - move it to a helper
  - find places where user might be able to set `derived_data_path` - those are the potential places where a zipping might make sense?

maybe just move it to a helper and use it in scan, with this PR - and do the rest in separate pr's 😀

joshdholtz:
@hjanuschka Updated the codes!

hjanuschka:
awesome!

hjanuschka (approved):

PR #12382, opened 2018-04-25
[swift] find swift lane names with option parameters

joshdholtz:
## Motivation
Be able to run a lane with options and use the options in swift
```sh
$: fastlane test some_option:"I want to see this in swift"
```

## Issue
- fastlane swift technically supports `func testLane(withOptions options: [String: String]?)` but fastlane ruby wasn't able to find this as a lane due to regex being too strict
- `Fastlane::LaneList` regex for a finding a lane name swift function wasn't allowing for a line like `func testLane(withOptions options: [String: String]?)`
  - The regex was looking only for `\s*` between the parentheses but really needed `(.*)`

## Fix
- Switched `\s*` for `(.*)` in the swift function regex 
- Added tests for `lane_name_from_swift_line`
- Also added extra check to **not set** not try to add a lane name if `lane_name_from_swift_line` returns `nil`

## Result 🎈 

```sh
$: fastlane test some_option:"I want to see this in swift"
```

```swift
func testLane(withOptions options:[String: String]?) {
	puts(message: "options: \(options)")
}
```

```sh
➜  test-swift git:(master) ✗ fl test some_option:'hey'
[✔] 🚀
[07:32:16]: loading manifest: /Users/josh/Projects/fastlane/fastlane/fastlane/swift/upgrade_manifest.json
[07:32:16]: FastlaneSwiftRunner project is up-to-date
[07:32:16]: $ ./fastlane/FastlaneRunner lane test some_option hey > /dev/null
[07:32:16]: options: Optional(["some_option": "hey"])
```

### DIFF
--- a/fastlane/lib/fastlane/lane_list.rb
+++ b/fastlane/lib/fastlane/lane_list.rb
@@ -52,7 +52,6 @@ module Fastlane
 
     def self.lane_name_from_swift_line(potential_lane_line: nil)
       function_name_match = SWIFT_FUNCTION_REGEX.match(potential_lane_line)
-
       unless function_name_match
         return nil
       end
--- a/fastlane/lib/fastlane/lane_list.rb
+++ b/fastlane/lib/fastlane/lane_list.rb
@@ -1,7 +1,7 @@
 module Fastlane
   class LaneList
     # Print out the result of `generate`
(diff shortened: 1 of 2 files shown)


KrauseFx (approved):
So many tests 😀

joshdholtz:
@KrauseFx Safety first!

PR #12383, opened 2018-04-25
Screengrab + Snapshot: Generate ids onto languages headers and device name sections.

vanniktech:
### Checklist

### Motivation and Context

I want to have some links that I can open quickly and jump to the correct language / language device name. I've got 17 different languages and this feature makes it easier to navigate and handle them.

### Description

Added the ids. Here's the diff:

```diff
-<h1 class="language">en-US</h1>
+<h1 id="en-US" class="language">en-US</h1>
<hr>
<table>
  <tr>
    <th colspan="36">
-      <span class="deviceName">phoneScreenshots</span>
+      <span id="en-US-phoneScreenshots" class="deviceName">phoneScreenshots</span>
    </th>
  </tr>
```

### DIFF
--- a/screengrab/lib/screengrab/page.html.erb
+++ b/screengrab/lib/screengrab/page.html.erb
@@ -65,12 +65,12 @@
     </style>
   </head>
   <body><% image_counter = 0 %><% @data.each do |language, content| %>
-    <h1 class="language"><%= language %></h1>
+    <h1 id="<%= language %>" class="language"><%= language %></h1>
     <hr>
     <table><% content.each do |device_name, screens| %>
       <tr>
         <th colspan="<%= screens.count %>">
-          <span class="deviceName"><%= device_name %></span>
+          <span id="<%= language %>-<%= device_name %>" class="deviceName"><%= device_name %></span>
         </th>
       </tr>
       <tr><% screens.each do |screen_path| %><% next if screen_path.include?"_framed.png" %>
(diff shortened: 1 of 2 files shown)


vanniktech:
Can someone with the proper rights restart one of the builds?

> Assigned VM request id: 11806683
  provisioning: ........................................
We had an unexpected error preparing a VM for this build, potentially due to our infrastructure or cloud provider.  Please retry the build in a few minutes
Unexpected environment preparation error: VM creation failed

joshdholtz (commented):
@vanniktech  This is a cool idea 💡 I wonder if we could take this one step further and turn the `<span>` into a `<a>` so its easier to get the link?

```html
<a id="en-US-iPad Pro (12.9-inch)" class="deviceName" href="#en-US-iPad Pro (12.9-inch)">iPad Pro (12.9-inch)</a>
```

vanniktech:
How about supporting both? So the id on the span as well as on links?

joshdholtz:
@vanniktech ids can only be in one place 🙃 Is there a reason why changing the `<span>` to `<a>` won't work?

Or, I guess instead of changing the whole thing from a `<span>` to an `<a>`, we just maybe put a "link" image next to span? I can add onto this PR if you would prefer 😊

joshdholtz (approved):
Thanks for making that change! 😊

PR #12387, opened 2018-04-25
Add skip_bundle_update option to the update_docs lane

revolter:
Running:

```
bundle exec update_docs skip_plugin_scores:true skip_bundle_update:true
debug:true
```

finished in about 5-7 seconds which helps a lot when modifying and
testing docs changes.


joshdholtz (approved):
💯

PR #12389, opened 2018-04-25
[slack] add link names arg

sujithvm:
Motivation and Context
Mentioning a developer or pinging a slack channel in a slack message is often a nice feature in cases where a build/deployment fails. Currently it does not link them in a slack message.

Description
Slack API has link_names arg that can be use to link usernames and channels.

### DIFF
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -33,6 +33,8 @@ module Fastlane
 
         notifier = Slack::Notifier.new(options[:slack_url], channel: channel, username: username)
 
+        link_names = options[:link_names]
+
         icon_url = options[:use_webhook_configured_username_and_icon] ? nil : options[:icon_url]
 
         slack_attachment = generate_slack_attachments(options)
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -124,6 +124,12 @@ def self.available_options
                                        description: "Should an error sending the slack notification cause a failure? (true/false)",
                                        optional: true,
                                        default_value: true,
+                                       is_string: false),
+          FastlaneCore::ConfigItem.new(key: :link_names,
+                                       env_name: "FL_SLACK_LINK_NAMES",
+                                       description: "Find and link channel names and usernames. (true/false)",
```
joshdholtz: Can you remove the `.` from the description? 🙃

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -40,7 +40,7 @@ def self.run(options)
         return [notifier, slack_attachment] if Helper.test? # tests will verify the slack attachments and other properties
 
         begin
-          results = notifier.ping('', icon_url: icon_url, attachments: [slack_attachment])
+          results = notifier.ping('', link_names: options[:link_names], icon_url: icon_url, attachments: [slack_attachment])
```
joshdholtz: Can you also set the option to a variable above and use that variable here? Since that is what it seems like this action tends to do 😊

joshdholtz (changes requested):
This is a good addition! Just a few small nit picks 😀 Can you also sign the CLA too?

sujithvm:
@joshdholtz Thank you very much for reviewing my PR. I have signed CLA. Let me know if any other change is required

joshdholtz (approved):
This is great! Thanks for the contribution 😊

PR #12394, opened 2018-04-26
[verify_build] Add support for xcarchive files

mgrebenets:
- Add xcarchive fixture and use it instead of .app file
- Add tests for xcarchive support and for invalid input edge-cases
- Ensure that app_path exists

### Checklist

### Motivation and Context
This is a follow up for #12370.
Since `xcarchive` files are also a build product, it makes sense to have an option to verify `xcarchive` as well. 
This change allows future improvements, such as:
- Not only verify the `.app` file inside an Xcode archive, but also verify the integrity of the archive itself, e.g. make sure that dSYMs and bitcode symbols match the binary, verify archive's `Info.plist`, make sure that `SwiftSupport` contains correct list of libraries, etc.

### Description
Initial support for `xcarchive` files will only verify `.app` file found in the archive. The validity of `Info.plist` is not checked yet.


fastlane/lib/fastlane/actions/verify_build.rb
```diff
@@ -30,10 +30,13 @@ def self.app_path(params, dir)
           UI.user_error!("Unable to unzip ipa") unless $? == 0
           # Adding extra ** for edge-case ipas where Payload directory is nested.
           app_path = Dir["#{dir}/**/Payload/*.app"].first
+        elsif File.extname(build_path) == ".xcarchive"
+          app_path = Dir["#{build_path}/Products/Applications/*.app"].first
         else
-          app_path = build_path
+          app_path = build_path # Assume that input is an app file.
```
revolter: You could replace this with a `switch`:

```ruby
case File.extname(build_path)
when ".ipa", ".zip"
  ...
when ".xcarchive"
  ...
else
  ...
end
```

mgrebenets: 👍

joshdholtz (approved):
So good 👍

PR #12409, opened 2018-04-30
added secure text input option

### LINKED ISSUE #12408: [action] prompt - secure text entry
Currently there seems to be no support for secure text entry for the action `prompt`. If the user wishes to ask for password, the password will be displayed in run time. There could be an option which should hide the entered text.

RishabhTayal:
fixes #12408


joshdholtz (approved):
Looks good 💪

Issue #32, opened 2015-02-05
Fastlane init: snapshots directory is messed probably


artem-alexandrov:
I've had SnapFile setup properly before I decided to move to Fastlane.
on init step I've seen such lines:
```
Successfully created new Deliverfile at './fastlane//Deliverfile'
INFO [2015-02-05 17:08:43.24]: The 'screenshots' folder inside the 'deliver' folder will not be used.
INFO [2015-02-05 17:08:43.24]: Instead the 'screenshots' folder inside the 'fastlane' folder will be used.
INFO [2015-02-05 17:08:43.24]: Click Enter to confirm
```
But I've found that snapshots directory remains ./deliver/snapshots instead of .snapshots when I open Deliverfile.

I'm not sure whenever it is a bug or not, but looks like this line should be changed.

KrauseFx:
Can you please share your `Snapfile`?

artem-alexandrov:
```
# Uncomment the lines below you want to change by removing the # in the beginning

# A list of devices you want to take the screenshots from
devices([
  "iPhone 6",
  "iPhone 6 Plus",
  "iPhone 5",
  "iPhone 4s",
  "iPad Air"
])

languages([
  'en-US'
])

# Where should the resulting screenshots be stored?
screenshots_path "./screenshots"

clear_previous_screenshots # remove the '#'' to clear all previously generated screenshots before creating new ones

# JavaScript UIAutomation file
js_file './snapshot.js'

# The name of the project's scheme
scheme 'Application'

# Where is your project (or workspace)? Provide the full path here
#project_path '../NIXProject.xcworkspace'

# By default, the latest version should be used automatically. If you want to change it, do it here
# ios_version '8.1'

# Custom Callbacks

# setup_for_device_change do |device| 
#   puts "Preparing device: #{device}"
# end

# setup_for_language_change do |lang, device|
#   puts "Running #{lang} on #{device}"
#   system("./popuplateDatabase.sh")
# end

# teardown_language do |lang, device|
#   puts "Finished with #{lang} on #{device}"
# end

# teardown_device do |device|
#   puts "Cleaning device #{device}"
#   system("./cleanup.sh")
# end

```

KrauseFx:
Thanks, fixed with https://github.com/KrauseFx/fastlane/commit/a5914ff7db65f17c20cd6bca826e62dcb3e5c34e

artem-alexandrov:
Man, you are awesome!

[...]

PR #12414, opened 2018-04-30
[frameit] accepts utf8 and utf16 .strings files

### LINKED ISSUE #12127: frameit: Why do the localised .strings file need to be in UTF-16 BE as the code converts them to UTF-8?
### New Issue Checklist

### Issue Description
The frameit docu at https://docs.fastlane.tools/actions/frameit/ shows:

> Note: These .strings files MUST be utf-16 encoded (UTF-16 BE with BOM). They also must begin with an empty line. If you are having trouble see issue #1740

However, the first thing that Class _strings_parser.rb_ does is converting the file to UTF-8 encoding:
https://github.com/fastlane/fastlane/blob/48ad2fae84e10c3c89402b702f8dd6bdc2f71660/frameit/lib/frameit/strings_parser.rb#L12-L13

We've got a struggle to maintain UTF-16 BE encoding for these .strings files in Xcode over commits.
Often it falls back to some format (WINDOWS-1252 according to [uchardet](https://github.com/BYVoid/uchardet) ) that destroys special characters in locales, like French.
We don't have this problem with the app specific locale .strings files, because they all have default UTF-8 encoding.

So what is the reason these frameit specific .strings files should have UTF-16 BE encoding?
What is needed to support UTF-8 encoding?

joshdholtz:
Fixes #12127

- Accepts utf8, us-acii (which is a subset of utf8), and utf16 .strings files
- Suggested by @KrauseFx to work with both - https://github.com/fastlane-old/frameit/issues/32#issuecomment-125682831


KrauseFx (approved):
Good ol' #32 on `frameit-old` 👍

PR #12416, opened 2018-04-30
Make sure xcov action's parameters are generated in the docs

revolter:
Having `xcov` added as a development dependency ensures that the computer on which the `update_docs` lane is run has `xcov` installed so the `available_options` can be returned from:

https://github.com/fastlane/fastlane/blob/d56f70d416ae63c60172b84d4763be97f837779e/fastlane/lib/fastlane/actions/xcov.rb#L34

Before:

screen shot 2018-05-01 at 01 17 09

After:

screen shot 2018-05-01 at 01 17 15

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -114,4 +114,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency('rest-client', '>= 1.8.0')
   spec.add_development_dependency('fakefs', '~> 0.8.1')
   spec.add_development_dependency('sinatra', '~> 1.4.8')
+  spec.add_development_dependency('xcov', '~> 1.4.1')
 end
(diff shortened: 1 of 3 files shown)


revolter:
@joshdholtz, Fixed the tests, could you take a look? 😃

fastlane.gemspec
```diff
@@ -114,4 +114,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency('rest-client', '>= 1.8.0')
   spec.add_development_dependency('fakefs', '~> 0.8.1')
   spec.add_development_dependency('sinatra', '~> 1.4.8')
+  spec.add_development_dependency('xcov', '~> 1.4.1')
```
joshdholtz: Can we put a comment here to why we are requiring this as a dev dependency? (since its kind of an outlier) 😊

joshdholtz (changes requested):

joshdholtz (approved):

PR #12420, opened 2018-04-30
Adding support for b2b vpp

paramv17:
### Checklist

### Motivation and Context

I wished to provide the support to enable vpp and add the vpp email addresses. I have deliberately not provided a way to remove them or disable vpp because i do not want to introduce changes that may mess up the system.

I have tested adding b2b users when existing users present. also enabled b2b distribution when it was disabled. I have also ensured it by making sure existing spec tests are working - which is a good test in this case since the specs for availability were quite extensive.

### Description
Created a class `b2b_user` for parsing the user details
Provided parsing for `b2bAppEnabled`, `b2bAppFlagDisabled` and `b2bUsers`
Updated `TunesClient`, `AvailabilitySpec`, `TunesStubbing` for this change and added a couple of fixtures for tests.

### DIFF
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -36,11 +36,10 @@ module Spaceship
       )
 
       # Create a new object based on a set of territories.
+      # This will override any values already set for cleared_for_preorder, app_available_date, b2b_unavailable,
+      # b2b_app_enabled, and educational_discount
       # @param territories (Array of String or Spaceship::Tunes::Territory objects): A list of the territories
       # @param params (Hash): Optional parameters (include_future_territories (Bool, default: true) Are future territories included?)
-      # This method has serious implications on vpp + educational discount but i don't want to break
-      # existing support. please be very careful when using this and provide appropriate params.
-      # In future, we should create availability from territories as an instance method not as a class method.
       def self.from_territories(territories = [], params = {})
         # Initializes the DataHash with our preOrder structure so values
         # that are being modified will be saved
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -22,6 +22,8 @@ module Spaceship
       # @return (Bool) b2b available for distribution
       attr_accessor :b2b_unavailable
 
+      # @return (Array of Spaceship::Tunes::B2bUser objects) A list of users set by user - if not
+      # then the b2b user list that is currently set
       attr_accessor :b2b_users
 
       attr_mapping(
@@ -82,21 +84,18 @@ module Spaceship
       # Sets the b2b flag. If you call Save on app_details without adding any b2b users
       # it will result in an error.
       def enable_b2b_app!
-        print("b2b : " + b2b_unavailable.to_s)
         raise "Not possible to enable b2b on this app" if b2b_unavailable
         @b2b_app_enabled = true
-        return self
+        self
       end
 
-      # just adds to the availability, You will still have to call update_availabilty
+      # just adds users to the availability, You will still have to call update_availabilty
       def add_b2b_users(user_list = [])
         raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
-        b2b_user_array = []
-        user_list.each do |user|
-          b2b_user_array.push(B2bUser.from_username(user))
+        @b2b_users = user_list.map do |user|
+          B2bUser.from_username(user)
         end
-        @b2b_users = b2b_user_array
-        return self
+        self
       end
     end
   end
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -1,6 +1,5 @@
 require_relative 'territory'
 require_relative 'b2b_user'
-
 module Spaceship
   module Tunes
     class Availability < TunesBase
@@ -19,6 +18,9 @@ module Spaceship
       # @return (Bool) app enabled for b2b users
       attr_accessor :b2b_app_enabled
 
+      # @return (Bool) app enabled for educational discount
+      attr_accessor :educational_discount
+
       # @return (Bool) b2b available for distribution
       attr_accessor :b2b_unavailable
 
@@ -30,13 +32,15 @@ module Spaceship
         'theWorld' => :include_future_territories,
         'preOrder.clearedForPreOrder.value' => :cleared_for_preorder,
         'preOrder.appAvailableDate.value' => :app_available_date,
-        'b2bAppEnabled' => :b2b_app_enabled,
         'b2BAppFlagDisabled' => :b2b_unavailable
       )
 
       # Create a new object based on a set of territories.
       # @param territories (Array of String or Spaceship::Tunes::Territory objects): A list of the territories
       # @param params (Hash): Optional parameters (include_future_territories (Bool, default: true) Are future territories included?)
+      # This method has serious implications on vpp + educational discount but i don't want to break
+      # existing support. please be very careful when using this and provide appropriate params.
+      # In future, we should create availability from territories as an instance method not as a class method.
       def self.from_territories(territories = [], params = {})
         # Initializes the DataHash with our preOrder structure so values
         # that are being modified will be saved
@@ -66,6 +70,7 @@ module Spaceship
         obj.app_available_date = params.fetch(:app_available_date, nil)
         obj.b2b_unavailable =  params.fetch(:b2b_unavailable, false)
         obj.b2b_app_enabled =  params.fetch(:b2b_app_enabled, false)
+        obj.educational_discount = params.fetch(:educational_discount, true)
         return obj
       end
 
@@ -77,6 +82,14 @@ module Spaceship
         @b2b_users ||= raw_data['b2bUsers'].map { |user| B2bUser.new(user) }
       end
 
+      def b2b_app_enabled
+        @b2b_app_enabled.nil? ? raw_data['b2bAppEnabled'] : @b2b_app_enabled
+      end
+
+      def educational_discount
+        @educational_discount.nil? ? raw_data['educationalDiscount'] : @educational_discount
+      end
+
       def cleared_for_preorder
         super || false
       end
@@ -86,6 +99,8 @@ module Spaceship
       def enable_b2b_app!
         raise "Not possible to enable b2b on this app" if b2b_unavailable
         @b2b_app_enabled = true
+        # need to set the educational discount to false
+        @educational_discount = false
         self
       end
 
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -1,4 +1,5 @@
 require_relative 'territory'
+require_relative 'b2b_user'
 
 module Spaceship
   module Tunes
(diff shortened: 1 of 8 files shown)


spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -15,10 +16,20 @@ class Availability < TunesBase
       # @return (String) App available date in format of "YYYY-MM-DD"
       attr_accessor :app_available_date
 
+      # @return (Bool) app enabled for b2b users
+      attr_accessor :b2b_app_enabled
+
+      # @return (Bool) b2b available for distribution
+      attr_accessor :b2b_unavailable
+
+      attr_accessor :b2b_users
```
joshdholtz: Can you add a comment above this one?

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -51,16 +62,42 @@ def self.from_territories(territories = [], params = {})
...
+      end
+
       def cleared_for_preorder
         super || false
       end
+
+      # Sets the b2b flag. If you call Save on app_details without adding any b2b users
+      # it will result in an error.
+      def enable_b2b_app!
+        print("b2b : " + b2b_unavailable.to_s)
```
joshdholtz: Can you 💣 this `print`?

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -51,16 +62,42 @@ def self.from_territories(territories = [], params = {})
...
+
+      # just adds to the availability, You will still have to call update_availabilty
+      def add_b2b_users(user_list = [])
+        raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
+        b2b_user_array = []
+        user_list.each do |user|
+          b2b_user_array.push(B2bUser.from_username(user))
+        end
+        @b2b_users = b2b_user_array
+        return self
```
joshdholtz: This method looks like its "setting" users instead of "adding" users. Should this method be renamed? Also, this method could be cleaned up to look like the follow if it is actually just "setting"...

```rb
def add_b2b_users(user_list = [])
  raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
  @b2b_users = map do |user|
    B2bUser.from_username(user)
  end
  return self
end
```

joshdholtz (changes requested):
This looks good! Just a few small comments 🙃

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: @joshdholtz its peculiar but when i examined the request body, only the ones that were being added or removed were included in the request body. so this is indeed "adding" instead of "setting". But your comment about "cleanup" stands. I will make the change.

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: oops! :/

heartofagoof:
@joshdholtz see if this looks good!

heartofagoof:
Oops I think I found a bug! Do not merge. I m working on a fix.

heartofagoof:
Fixed. 😥

joshdholtz:
@heartofagoof I will pull this down and test it later today / over this weekend! Would be able to provide me with some code examples I can play with for all the functionalities that you added? 😊 I can replace with all of my app ids and stuff but nice to have an example to base it off of 💪

heartofagoof:
@joshdholtz Something like this should work. I tried it out. 

```
require 'spaceship'  
Spaceship::Tunes.login()
Spaceship::Tunes.select_team 
app = Spaceship::Tunes::Application.find("")
a = app.availability
a = a.enable_b2b_app!
a = a.add_b2b_users([""])
app.update_availability!(a)
```
And then try to just add b2b users when the b2b is enabled. That should work too. 

What should fail is this : 

- if it is not possible to enable b2b, it should raise an exception
- ifit is not enabled when trying to add b2b users it should also raise an exception.

heartofagoof:
how's it looking @joshdholtz ?

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -15,15 +15,32 @@ class Availability < TunesBase
...
+        'preOrder.appAvailableDate.value' => :app_available_date,
+        'b2BAppFlagDisabled' => :b2b_unavailable
       )
 
       # Create a new object based on a set of territories.
       # @param territories (Array of String or Spaceship::Tunes::Territory objects): A list of the territories
       # @param params (Hash): Optional parameters (include_future_territories (Bool, default: true) Are future territories included?)
+      # This method has serious implications on vpp + educational discount but i don't want to break
+      # existing support. please be very careful when using this and provide appropriate params.
+      # In future, we should create availability from territories as an instance method not as a class method.
```
joshdholtz: Can we remove this comment block after the `@param` and move it to a new line between the `Create` and the first `@param` saying something like...
```
Create a new object based on a set of territories.
This will override any values already set for cleared_for_preorder, app_available_date, b2b_unavailable, b2b_app_enabled, and educational_discount
# @param territories (Array of String or Spaceship::Tunes::Territory objects): A list of the territories
# @param params (Hash): Optional parameters (include_future_territories (Bool, default: true) Are future territories included?)
```

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -51,16 +68,50 @@ def self.from_territories(territories = [], params = {})
...
+      def enable_b2b_app!
+        raise "Not possible to enable b2b on this app" if b2b_unavailable
+        @b2b_app_enabled = true
+        # need to set the educational discount to false
+        @educational_discount = false
+        self
+      end
+
+      # just adds users to the availability, You will still have to call update_availabilty
+      def add_b2b_users(user_list = [])
```
joshdholtz: Does this actually "set" the users and not "add" users? 🤔

joshdholtz: If this does set/override the `@b2b_users`, could we maybe combine `enable_b2b_app!` with this one and just pass the user list into the `enable_b2_app!` method?

joshdholtz (changes requested):

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: It does not set or override. It may look like this but the way the requests are ( if you observe the object ) only the email ids either being added or removed are sent over.

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: Done!

spaceship/lib/spaceship/tunes/availability.rb
joshdholtz: @heartofagoof Interesting 🤔 Good to know! So this _could_ also remove the user if the user was already set?

heartofagoof: No. For that you'd have to set removed = true. Which we are not setting right now. I didn't want to implement anything destructive. :)

heartofagoof: I can of course change the implementation to something like this if it helps : 

`update_b2b_users(users_to_add: [], users_to_remove: [])`

and update the signature to: 

`from_username(email_id: [], add_or_remove:)`

heartofagoof: This way we provide the ability to add and remove b2b teams.

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -51,16 +67,50 @@ def self.from_territories(territories = [], params = {})
...
+
+      def educational_discount
+        @educational_discount.nil? ? raw_data['educationalDiscount'] : @educational_discount
+      end
+
       def cleared_for_preorder
         super || false
       end
+
+      # Sets the b2b flag. If you call Save on app_details without adding any b2b users
```
joshdholtz: Can we change this comment to...
```rb
# Sets `b2b_app_enabled` to true and `educational_discount` to false
# Requires users to be added with `add_b2b_users` otherwise `update_availability` will error
```

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -51,16 +67,50 @@ def self.from_territories(territories = [], params = {})
...
+      # it will result in an error.
+      def enable_b2b_app!
+        raise "Not possible to enable b2b on this app" if b2b_unavailable
+        @b2b_app_enabled = true
+        # need to set the educational discount to false
+        @educational_discount = false
+        self
+      end
+
+      # just adds users to the availability, You will still have to call update_availabilty
```
joshdholtz: And...
```rb
# Adds users for b2b enabled apps
```

joshdholtz (changes requested):
Just a few nitpicks on the comments and then I think we will be good 😊

joshdholtz (approved):
Looks good! 👍 But waiting for one more set of 👀

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -613,8 +611,11 @@ def update_availability!(app_id, availability)
       # API will error out if cleared_for_preorder is false and app_available_date has a date
       cleared_for_preorder = availability.cleared_for_preorder
       app_available_date = cleared_for_preorder ? availability.app_available_date : nil
+      data["b2bAppEnabled"] = availability.b2b_app_enabled
+      data["educationalDiscount"] = availability.educational_discount
       data["preOrder"]["clearedForPreOrder"] = { "value" => cleared_for_preorder, "isEditable" => true, "isRequired" => true, "errorKeys" => nil }
       data["preOrder"]["appAvailableDate"] = { "value" => app_available_date, "isEditable" => true, "isRequired" => true, "errorKeys" => nil }
+      data["b2bUsers"] = availability.b2b_app_enabled ? availability.b2b_users.map { |user| { "value" => { "add" => user.add, "delete" => user.delete, "dsUsername" => user.ds_username } } } : []
```
snatchev: I believe you can convert the `b2b_users` back into data by doing: `b2b_users.raw_data.to_h`. Ideally, the model and the `attr_mapping` should be the only place that has to know how to convert to and from the JSON data.

snatchev (approved):
I _think_ we can take advantage of the attr_mapping to get the data back out in the right format when we make the request. This is more of a nice-to-have if it works correctly (which I haven't tested).

Otherwise, it looks good to me. Nice specs :)

spaceship/lib/spaceship/tunes/tunes_client.rb
heartofagoof: Okay digging into the metaprogramming ya'll have to make sense of it. I tried to use the mapping but it was not working quite as expected - it might be the limitation of the model. 

I am more than happy to give it a proper go, however. If it is limited in scope, I can make the changes here but if it is extensive, hopefully be I can make it in another PR.

snatchev: Hey @heartofagoof, if it doesn't work as expected out of the box, I wouldn't bother too much trying to hack it to work

joshdholtz: Yeah, no worries at all! We will merge what you got here 😊

PR #12425, opened 2018-05-01
Specify source filename in pilot import docs

getaaron:
I ran into an issue where I incorrectly ran:

```bash
bundle exec fastlane import updated_testers.csv
```

I should've used the `-c` flag:

```bash
bundle exec fastlane import -c updated_testers.csv
```

Instead of either (1) importing the specified file, or (2) failing from invalid input, fastlane imported the `testers.csv` which was leftover from the last time I ran the `pilot export` command.

This PR attempts to clarify usage in the docs; a later PR will improve input validation.

### DIFF
--- a/pilot/lib/pilot/commands_generator.rb
+++ b/pilot/lib/pilot/commands_generator.rb
@@ -151,7 +151,7 @@ module Pilot
 
       command :import do |c|
         c.syntax = "fastlane pilot import"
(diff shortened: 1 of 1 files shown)


taquitos (approved):
Much clearer now, 🎷:goat:

KrauseFx (approved):

PR #12429, opened 2018-05-01
Changelog from git commits path option

### LINKED ISSUE #12397: changelog_from_git_commits
i want to write commit logs to WHAT TO TEXT while uploading testflight, then i saw "changelog_from_git_commits action, but there is no PATH parameter. My Fastlane directory and git repo are different paths.

RishabhTayal:
fixes #12397


fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
```diff
@@ -83,6 +87,12 @@ def self.available_options
                                        verify_block: proc do |value|
                                          UI.user_error!(":commits_count must be >= 1") unless value.to_i >= 1
                                        end),
+          FastlaneCore::ConfigItem.new(key: :path,
+                                       env_name: 'FL_CHANGELOG_FROM_GIT_COMMITS_PATH',
+                                       description: 'Path of the git repository',
+                                       optional: true,
+                                       default_value: './'
```
RishabhTayal: Not sure if this should be the value for `default_value`. Can someone suggest what would be the correct value here?

joshdholtz (approved):
Good fix 👍

PR #12450, opened 2018-05-04
Add xcpretty_output option to spm action

taisukeh:
### Checklist

### Motivation and Context

I use `spm`( swift package manager ) action to run tests of a SwiftPM library. But the output is difficult to see.

image

### Description

The xcpretty works well to summarize the `swift test` output. So I add `xcpretty_output` option to the `spm` action to enable xcpretty.

image

### DIFF
--- a/fastlane_core/lib/fastlane_core/command_executor.rb
+++ b/fastlane_core/lib/fastlane_core/command_executor.rb
@@ -47,8 +47,6 @@ module FastlaneCore
           UI.command_output(loading)
         end
 
-        return command unless Helper.sh_enabled? # only for the tests
-
         begin
           FastlaneCore::FastlanePty.spawn(command) do |command_stdout, command_stdin, pid|
             begin
--- a/fastlane_core/lib/fastlane_core/command_executor.rb
+++ b/fastlane_core/lib/fastlane_core/command_executor.rb
@@ -47,6 +47,8 @@ module FastlaneCore
           UI.command_output(loading)
         end
 
+        return command unless Helper.sh_enabled? # only for the tests
+
         begin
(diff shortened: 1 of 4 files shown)


fastlane_core/lib/fastlane_core/command_executor.rb
```diff
@@ -47,6 +47,8 @@ def execute(command: nil, print_all: false, print_command: true, error: nil, pre
           UI.command_output(loading)
         end
 
+        return command unless Helper.sh_enabled? # only for the tests
```
joshdholtz: I know the `sh` action works like this but I would like to keep any `Helper.sh_enabled?` outside of main code as much as possible (and since `FastlaneCommand::Executor` isn't `sh`) 😇 

So for now I think the best solution is here to probably do something like this in your tests...
```rb
before do
  # Force FastlaneCore::CommandExecutor.execute to return command
  allow(FastlaneCore::CommandExecutor).to receive(:execute).and_wrap_original do |m, *args|
    args[0][:command]
  end
end
```

joshdholtz (changes requested):
Small nit pick 😊

fastlane_core/lib/fastlane_core/command_executor.rb
taisukeh: @joshdholtz Fixed. Your code worked perfectly! Thank you for the good solution 🙇

joshdholtz (approved):
I don't have anything to run this with myself but the tests look good and pass so let's merge this one 👍

joshdholtz:
@taisukeh Thanks for the contribution!

PR #12453, opened 2018-05-04
Enable variables in git path

skrabacz-michal:
### Checklist

### Motivation and Context
This change allows using environment variables in git paths (this is very useful or even mandatory to use a repository with certificates on GitLab).

### Description
Change is quite simple - I just replaced single quotes with double ones.

### DIFF
--- a/match/lib/match/git_helper.rb
+++ b/match/lib/match/git_helper.rb
@@ -25,7 +25,7 @@ module Match
 
       @dir = Dir.mktmpdir
 
-      command = "git clone '#{git_url}' '#{@dir}'"
+      command = %(git clone "#{git_url}" '#{@dir}')
       if shallow_clone
         command << " --depth 1 --no-single-branch"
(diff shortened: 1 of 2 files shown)


match/lib/match/git_helper.rb
```diff
@@ -25,7 +25,7 @@ def self.clone(git_url,
 
       @dir = Dir.mktmpdir
 
-      command = "git clone '#{git_url}' '#{@dir}'"
+      command = %(git clone "#{git_url}" '#{@dir}')
```
joshdholtz: Can we make this one like the rest with double quotes and escaped double quotes? 😊

joshdholtz (changes requested):
Small nit pick 🙃 Also, what does your `git_url` look like to enable environment variables for gitlab? Just curious 😇

skrabacz-michal:
change not needed

PR #12457, opened 2018-05-05
added `team_name` to spaceship client

RishabhTayal:
Currently the spaceship client doesn't contain a function to set the `team_name` from iTC. There is a `select_team` function. However I think it would be better if there is a `team_name` function similar to `team_id` to set the name.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -209,8 +209,6 @@ module Spaceship
           team_name: (team["contentProvider"] || {})["name"]
         }
       end
-      require 'pry'
-      binding.pry
 
       result = available_teams.find do |available_team|
         team_name.to_s == available_team[:team_name].to_s
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -211,7 +211,7 @@ module Spaceship
       end
       require 'pry'
       binding.pry
-      
+
       result = available_teams.find do |available_team|
         team_name.to_s == available_team[:team_name].to_s
       end
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -186,6 +186,56 @@ module Spaceship
       @current_team_id = team_id
     end
 
+    # @return (String) The currently selected Team Name
+    def team_name
+      return @current_team_name if @current_team_name
+
+      if teams.count > 1
+        puts("The current user is in #{teams.count} teams. Pass a team Name or call `select_team` to choose a team. Using the first one for now.")
+      end
+      @current_team_name ||= teams[0]['contentProvider']['name']
+    end
+
+    # Set a new team Name which will be used from now on
+    def team_name=(team_name)
+      # First, we verify the team actually exists, because otherwise iTC would return the
+      # following confusing error message
+      #
+      #     invalid content provider name
+      #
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/client.rb
```diff
@@ -186,6 +186,56 @@ def team_id=(team_id)
...
+      return @current_team_name if @current_team_name
+
+      if teams.count > 1
+        puts("The current user is in #{teams.count} teams. Pass a team Name or call `select_team` to choose a team. Using the first one for now.")
+      end
+      @current_team_name ||= teams[0]['contentProvider']['name']
+    end
+
+    # Set a new team Name which will be used from now on
+    def team_name=(team_name)
```
joshdholtz: I totes see the use case for the `team_name` but do we really need `team_name=`? Selected with a team name _should_ be possible with existing functionality in `select_team(team_id: nil, team_name: nil)`, right?

RishabhTayal: @joshdholtz I don't think we really *need* `team_name=`. It's a nice to have since we have `team_id=`. We could definitely use `select_team()` function to set the `team_name`.

Feel free to reject this PR. 😄 

If you think about it we don't even need `team_id=` as well. If not, I can remove that code. LMK.

joshdholtz: @RishabhTayal We totally need `team_id=` still for backwards compatibility which is why I didn't recommend deleting that 🙃 But if we could 💣 that `team_name=` I think this PR is valid to add a `team_name` attribute/method

spaceship/lib/spaceship/client.rb
```diff
@@ -186,6 +186,54 @@ def team_id=(team_id)
...
 
+    # @return (String) The currently selected Team Name
+    def team_name
+      return @current_team_name if @current_team_name
+
+      if teams.count > 1
+        puts("The current user is in #{teams.count} teams. Pass a team Name or call `select_team` to choose a team. Using the first one for now.")
+      end
+      @current_team_name ||= teams[0]['contentProvider']['name']
+    end
```
joshdholtz: @RishabhTayal Now that I think about this, we should probably 💣 `@current_team_name` and replace it as a method that looks up from `teams` based on `@current_team_id`

joshdholtz (changes requested):

RishabhTayal:
@joshdholtz it's ready to review again.

joshdholtz (changes requested):
Can we add some tests for this? 🙃

RishabhTayal:
Not sure what kind of tests we can add for this. I don't see any tests written for `team_id` as well.

joshdholtz:
@RishabhTayal Something like this 😊 https://github.com/fastlane/fastlane/blob/master/spaceship/spec/portal/portal_client_spec.rb#L32-L67

RishabhTayal:
@joshdholtz I am not sure this is applicable here. The test you pointed out is for `portal_client.rb` class. When trying to write similar tests I am getting `nil` values.

RishabhTayal:
@joshdholtz could you sign the CLA? 😝🤣🤣

RishabhTayal:
This is ready to review cc @taquitos @revolter

joshdholtz (approved):

joshdholtz (approved):
💪 🚀

PR #12466, opened 2018-05-07
[action] get_apple_pay_certificate

### LINKED ISSUE #11809: [match] How can make certificate for ApplePay base on CSR?
How can make certificate for ApplePayfor merchant base on CSR through fastlane?
Please help me

RishabhTayal:
fixes: #11809

Generates Apple pay merchant certificates


RishabhTayal:
@joshdholtz, @taquitos This is ready to review.

kioqq (approved):
looks good

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+          if ex.to_s.include?("You already have a current")
+            UI.message(ex.to_s)
+            UI.user_error!("You already have 2 apple pay certificates for this merchant. You'll need to revoke an old certificate to make room for a new one")
+          else
+            raise ex
+          end
+        end
+
+        x509_certificate = cert.download
+        filename_base = params[:merchant_bundle_id] # PEM.config[:pem_name] || "#{PEM.config[:app_identifier]}"
```
janpio: What does this comment mean here?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.is_supported?(platform)
+        [:ios].include?(platform)
+      end
+
+      def self.description
+        'Creates Apple Pay certificate'
```
janpio:
```suggestion
        'Creates an Apple Pay certificate'
```

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+      end
+
+      def self.details
+        "This action allows you to create Apple Pay merchant certificate."
```
janpio:
```suggestion
        "This action allows you to create an Apple Pay merchant certificate."
```

janpio: What is the difference between an "Apple Pay certificate" and an "Apple Pay merchant certificate"? Are these identical? Are there any other "Apple Pay vertificates" this could clash with?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+                                 verify_block: proc do |value|
+                                   ENV["FASTLANE_TEAM_ID"] = value.to_s
+                                 end),
+          FastlaneCore::ConfigItem.new(key: :p12_password,
+                                     short_option: "-p",
+                                     env_name: "APPLE_PAY_CERT_P12_PASSWORD",
+                                     sensitive: true,
+                                     description: "The password that is used for your p12 file",
+                                     default_value: ""),
+          FastlaneCore::ConfigItem.new(key: :team_name,
```
janpio: shouldn't team_id and team_name be next to each other?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+                                   description: "You merchant bundle identifier (e.g: merchant.com.example)"),
+          FastlaneCore::ConfigItem.new(key: :team_id,
+                                 short_option: "-b",
+                                 env_name: "APPLE_PAY_CERT_TEAM_ID",
+                                 code_gen_sensitive: true,
+                                 default_value: CredentialsManager::AppfileConfig.try_fetch_value(:team_id),
+                                 default_value_dynamic: true,
+                                 description: "The ID of your Developer Portal team if you're in multiple teams",
+                                 optional: true,
+                                 verify_block: proc do |value|
```
janpio: what does this verify block do an why?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+                                       optional: true,
+                                       code_gen_sensitive: true,
+                                       default_value: CredentialsManager::AppfileConfig.try_fetch_value(:team_name),
+                                       default_value_dynamic: true,
+                                       verify_block: proc do |value|
+                                         ENV["FASTLANE_TEAM_NAME"] = value.to_s
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :generate_p12,
+                                     env_name: "APPLE_PAY_CERT_GENERATE_P12_FILE",
+                                     description: "Generate a p12 file additionally to a PEM file",
```
janpio:
```suggestion
                                     description: "Generate a p12 file additionally to the PEM file",
```

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,153 @@
...
+        'Creates Apple Pay certificate'
+      end
+
+      def self.details
+        "This action allows you to create Apple Pay merchant certificate."
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :username,
```
janpio: Is this optional? The first code example doesn't inclue a username

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -206,6 +206,23 @@ def create_certificate_signing_request
...
+          OpenSSL::PKey::EC.send(:alias_method, :private?, :private_key?)
+
+          ec_domain_key = OpenSSL::PKey::EC.new('prime256v1')
+          ec_public = OpenSSL::PKey::EC.new('prime256v1')
+          ec_domain_key.generate_key
+          ec_public.public_key = ec_domain_key.public_key
+
+          csr = OpenSSL::X509::Request.new
+          csr.subject = OpenSSL::X509::Name.new([
+                                                  ['CN', 'PEM', OpenSSL::ASN1::UTF8STRING]
```
janpio: Strange formatting here.

janpio: What does the `CN` stand for here?

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -292,17 +309,26 @@ def create!(csr: nil, bundle_id: nil)
           mac = MAC_CERTIFICATE_TYPE_IDS.include?(type)
 
           # look up the app_id by the bundle_id
-          if bundle_id
+          if bundle_id && !bundle_id.start_with?('merchant.')
```
janpio: maybe better add an additional inner `if` instead of this strange combination of conditions? I think currently this is buggy if there actually is no `bundle_id`

Something like this maybe?
```
if bundle_id
   if !bundle_id.start_with?('merchant.')
      ...
   else
      ...
   end
end
```

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -292,17 +309,26 @@ def create!(csr: nil, bundle_id: nil)
...
+            raise "Could not find merchant with bundle id '#{bundle_id}" unless merchant
+            merchant_id = merchant.merchant_id
           end
 
           # ensure csr is a OpenSSL::X509::Request
           csr = OpenSSL::X509::Request.new(csr) if csr.kind_of?(String)
 
           # if this succeeds, we need to save the .cer and the private key in keychain access or wherever they go in linux
-          response = client.create_certificate!(type, csr.to_pem, app_id, mac)
+          if type == '4APLUP237T'
```
janpio: What is this magic number? Extract to a variable so it can be documented via the var name maybe?

spaceship/lib/spaceship/portal/portal_client.rb
```diff
@@ -586,6 +585,18 @@ def create_certificate!(type, csr, app_id = nil, mac = false)
       parse_response(r, 'certRequest')
     end
 
+    def create_certificate_apple_pay!(type, csr, merchant_id = nil, mac = false)
```
janpio: strange method name

fastlane/lib/fastlane/actions/apple_pay_cert.rb
janpio: Does this actually refer to this process?
> Create a payment processing certificate
https://help.apple.com/developer-account/#/devb2e62b839?sub=devf31990e3f

tmaly1980:
@RishabhTayal Seriously dude, can you approve the changes, since someone else made some modifications? This has been sitting around for a month, it's a pretty important feature for anyone using apple pay.

RishabhTayal:
@tmaly1980 I am not one who can approve the pull requests.

tmaly1980:
> @tmaly1980 I am not one who can approve the pull requests.

Not the PR. But something mentioned above:

```
The bad news is that it appears that one or more commits were authored or co-authored by someone other than the pull request submitter. We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that here in the pull request.
```

tmaly1980:
@googlebot Who else is needed to approve the changes?

RishabhTayal:
@janpio thanks for jumping in. I will address the comments later tomorrow.

RishabhTayal:
@tmaly1980 if you want to use action right now you can point to my fork and branch.

janpio:
There was another comment on this PR here:
https://github.com/fastlane/fastlane/issues/11809#issuecomment-438319150

RishabhTayal:
@janpio I think that something we can add later. That would be an enhancement on this feature.

tmaly1980:
@RishabhTayal I don't understand the line about creating a CSR. In my case (and I'd imagine many others), I have to provide an EXISTING CSR from my Stripe account in order to make Apple Pay work at all. How exactly does your code work? Wouldn't the Apple Pay calls require that the merchant (ie Stripe) has authority (via the CSR) to make them?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
tmaly1980: https://github.com/fastlane/fastlane/pull/12466/files#diff-a0e4b204f34e72aedc94d3610c0af05dR20

Would I be able to provide the CSR from my Stripe account to this? Apple Pay isn't going to work in my case unless I use Stripe's CSR. I don't believe I have the option to upload the CSR to Stripe - I have to use theirs I think.

janpio:
(I don't know anything about Apple Play and those signing certificates - and I suspect other reviewers will have the same problem. One of you might want to summarize the process that the user has to go through and where he gets what CSRs etc.)

tmaly1980:
@janpio @RishabhTayal I've done this a few times and Stripe generates the CSR for you via their dashboard. But I'd imagine the process varies for every merchant provider. 

https://stripe.com/docs/apple-pay/apps

RishabhTayal:
@tmaly1980 I can create a separate pr for that feature. But I don't want to mix that with this pr. This pr works as intended. We can add an option later to pass the csr file.

janpio:
> I think it's an essential feature and not worth waiting for.

You can create a PR based on @RishabhTayal's branch and create a PR that adds the functionality into his PR. Just fork his repository and branch, do the work, then open a PR _onto his branch_.  
If you are done before we get this one here merged, it will just be integrated into this PR and included in the merge when we do merge this one. If not, you can still change the merge base to master of fastlane itself without losing any work.

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -206,6 +206,23 @@ def create_certificate_signing_request
           return [csr, key]
         end
 
+        def create_apple_pay_certificate_signing_request
```
janpio: Is this different from the other CSR logic we already have to have around somewhere in spaceship?

RishabhTayal: @janpio  I don't think so. It's the same logic.

janpio: Wouldn't it make sense to use that logic here then as well?

spaceship/lib/spaceship/portal/certificate.rb
janpio: "Common Name"

janpio:
Resolved the merge conflict.

Could you take another look at the open comments here @RishabhTayal? Thanks.

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: @janpio actually, looks like the CSR logic for apple pay certificate is different than the one we currently have.

Current logic: https://github.com/RishabhTayal/fastlane/blob/2637eb2c673bb9fb7951bd9e3d8a5c1d4d58d21c/spaceship/lib/spaceship/portal/certificate.rb#L202-L202

Apple Pay logic: https://github.com/RishabhTayal/fastlane/blob/2637eb2c673bb9fb7951bd9e3d8a5c1d4d58d21c/spaceship/lib/spaceship/portal/certificate.rb#L214-L214

spaceship/lib/spaceship/portal/portal_client.rb
RishabhTayal: @janpio Do you have any other preferred name? This seems pretty self-explanatory to me.

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: @janpio This magic number is defined by Apple. It's also used in other places in the `fastlane` repo. https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/certificate.rb#L165

fastlane/lib/fastlane/actions/apple_pay_cert.rb
RishabhTayal: removed the comment. It was irrelevant.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
RishabhTayal: This is not optional. I have updated the first code example to include the username. Good catch. 👍

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: made the change.

RishabhTayal:
@janpio All the comments are addressed. This is ready for review again.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,154 @@
...
+
+      def self.author
+        'rishabhtayal'
+      end
+
+      def self.example_code
+        [
+          'apple_pay_cert(
+      			username: "test@example.com",
+            merchant_bundle_id: "merchant.com.rtayal.app",
```
janpio: fix indentation:
```suggestion
     			merchant_bundle_id: "merchant.com.rtayal.app",
```

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -211,6 +211,23 @@ def create_certificate_signing_request
...
+          OpenSSL::PKey::EC.send(:alias_method, :private?, :private_key?)
+
+          ec_domain_key = OpenSSL::PKey::EC.new('prime256v1')
+          ec_public = OpenSSL::PKey::EC.new('prime256v1')
+          ec_domain_key.generate_key
+          ec_public.public_key = ec_domain_key.public_key
+
+          csr = OpenSSL::X509::Request.new
+          csr.subject = OpenSSL::X509::Name.new([
+                                                  ['CN', 'PEM', OpenSSL::ASN1::UTF8STRING]
```
janpio: Is "Common Name" really `PEM` here? I think this is properly not surfaced anywhere, but this should probably be `spaceship` or something similar here, right?

RishabhTayal: We have a similar logic for `create_certificate_signing_request` method as well.

janpio: Yes, that is used by `PEM` and most probably a historical artifact. You are creating new code, so it should make sense _now_.

spaceship/lib/spaceship/portal/certificate.rb
janpio: So the next question is: Why does this have to be different? Is the certificate signing request different with newer requirements? Or doesn't this really matter and this is just the same thing, done differently?

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -298,16 +315,26 @@ def create!(csr: nil, bundle_id: nil)
 
           # look up the app_id by the bundle_id
           if bundle_id
-            app = portal_type.set_client(client).find(bundle_id)
-            raise "Could not find app with bundle id '#{bundle_id}'" unless app
-            app_id = app.app_id
+            if bundle_id.start_with?('merchant.')
+              merchant = Spaceship::Portal.merchant.find(bundle_id)
+              raise "Could not find merchant with bundle id '#{bundle_id}" unless merchant
```
janpio: This this really also called `bundle id`, although it refers to a merchant? Or is this just how our internal variable is called?

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: Make sense. I have changed it for the legacy code as well.

spaceship/lib/spaceship/portal/certificate.rb
janpio: Let's make this a bit more explicit here:
```suggestion
          if type == '4APLUP237T' # = ApplePay
```

Better would be if we could just read this from a constant by key `ApplePay` here of course, but we don't seem to have this in the code base in that direction, only from the magic number _to_ Apple Pay.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
janpio: still unanswered. 

I think this block this way is usually used to just copy the value to the ENV var, but I am unsure if this is actually what we want to do here as `APPLE_PAY_CERT_TEAM_ID != FASTLANE_TEAM_ID`

fastlane/lib/fastlane/actions/apple_pay_cert.rb
janpio: still unanswered - as this defines the order the params are displayed in the docs, this is important

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,154 @@
...
+                                     default_value: ""),
+          FastlaneCore::ConfigItem.new(key: :team_name,
+                                       short_option: "-l",
+                                       env_name: "APPLE_PAY_CERT_TEAM_NAME",
+                                       description: "The name of your Developer Portal team if you're in multiple teams",
+                                       optional: true,
+                                       code_gen_sensitive: true,
+                                       default_value: CredentialsManager::AppfileConfig.try_fetch_value(:team_name),
+                                       default_value_dynamic: true,
+                                       verify_block: proc do |value|
```
janpio: same as https://github.com/fastlane/fastlane/pull/12466/files#r265621743

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: It's called `bundle_id`. Refer to this 
Screenshot 2019-03-14 10 19 01

`bundle_id` is `merchant.com.krausefx.app.production` which is what we are using in here.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
```diff
@@ -0,0 +1,154 @@
...
+                                         ENV["FASTLANE_TEAM_NAME"] = value.to_s
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :generate_p12,
+                                     env_name: "APPLE_PAY_CERT_GENERATE_P12_FILE",
+                                     description: "Generate a p12 file additionally to the PEM file",
```
janpio: What does `PEM file` refer to here?

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -141,7 +141,7 @@ class VoipPush < PushCertificate; end
       class Passbook < Certificate; end
 
       # ApplePay certificate
-      class ApplePay < Certificate; end
```
janpio: 🚨 

This is not used anywhere else?
Could this be a breaking change for existing users of spaceship?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
RishabhTayal: I think `APPLE_PAY_CERT_TEAM_ID` is the `FASTLANE_TEAM_ID`. Why do you think it's not?

fastlane/lib/fastlane/actions/apple_pay_cert.rb
RishabhTayal: `FASTLANE_TEAM_NAME` is equal to `APPLE_PAY_CERT_TEAM_NAME`

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: Should I change it back to `ApplePay`? I don't remember why I changed it in the first place.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
RishabhTayal: 🤷‍♂️

spaceship/lib/spaceship/portal/certificate.rb
RishabhTayal: I am not exactly sure why it's different. Looks like Apple Pay signing certificate requires few more parameters to be sent for CSR. I don't want to change the existing `create_certificate_signing_request` method so I had to create a new one.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
RishabhTayal: No problem. Moved it.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
janpio: Do we want to set the globally applied ENV var via an action that "only" creates the apple pay cert?

(ping @mollyIV: Another usage of ENV vars that is a bit strange...)

spaceship/lib/spaceship/portal/certificate.rb
janpio: I am not really sure. You probably changed it as it is a better description of what it actually is, right? Can we add backwards compat via an alias?

RishabhTayal: added.

fastlane/lib/fastlane/actions/apple_pay_cert.rb
janpio: unanswered

fastlane/lib/fastlane/actions/apple_pay_cert.rb
janpio: unanswered

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -204,13 +207,30 @@ def create_certificate_signing_request
           csr = OpenSSL::X509::Request.new
           csr.version = 0
           csr.subject = OpenSSL::X509::Name.new([
-                                                  ['CN', 'PEM', OpenSSL::ASN1::UTF8STRING]
+                                                  ['CN', 'spaceship', OpenSSL::ASN1::UTF8STRING]
```
janpio: Do we want to or can we change this unrelated method here in this PR? Probably not.

spaceship/lib/spaceship/portal/certificate.rb
janpio: Hm, that is still a bit unsatisfactory, isn't it?

janpio:
@googlebot I consent.

spaceship/lib/spaceship/portal/certificate.rb
janpio: Ok, makes sense here - this was not correct before.

janpio (approved):
This works as advertised.

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -312,16 +335,27 @@ def create!(csr: nil, bundle_id: nil)
...
+              app_id = app.app_id
+            end
           end
 
           # ensure csr is a OpenSSL::X509::Request
           csr = OpenSSL::X509::Request.new(csr) if csr.kind_of?(String)
 
           # if this succeeds, we need to save the .cer and the private key in keychain access or wherever they go in linux
-          response = client.create_certificate!(type, csr.to_pem, app_id, mac)
+          if type == '4APLUP237T' # = ApplePay
```
mollyIV: Maybe instead of adding a comment, we could re-use previously declared key:

```
if type == IOS_CERTIFICATE_TYPE_IDS.key('ApplePayCertificate')
```

⚠️ ☝️ Most probably requires some tweaking, didn't test it out in the code.

mollyIV (changes requested):
Overall looks 💯 

Left a few comments here and there 🙇

fastlane/lib/fastlane/actions/get_apple_pay_certificate.rb
```diff
@@ -0,0 +1,154 @@
+require 'spaceship'
+
+module Fastlane
+  module Actions
+    module SharedValues
+    end
```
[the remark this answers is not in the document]
janpio: I would say we keep those, to set a standard of stuff that can be used in actions. Isn't it that way in other actions? (not at a computer right now, sry)

fastlane/lib/fastlane/actions/get_apple_pay_certificate.rb
```diff
@@ -0,0 +1,154 @@
...
+      def self.certificate
+        Spaceship.certificate.apple_pay_certificate
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.is_supported?(platform)
+        [:ios].include?(platform)
```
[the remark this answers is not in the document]
janpio: Hm, good question. Can Apply Pay certificates be used somewhere else? Do they also work on Macs? Or even web (which is not a valid `platform` yet...)?

fastlane/lib/fastlane/actions/get_apple_pay_certificate.rb
```diff
@@ -0,0 +1,154 @@
...
+      end
+
+      def self.login(params)
+        UI.message("Starting login with user '#{params[:username]}'")
+        Spaceship.login(params[:username], nil)
+        Spaceship.client.select_team(team_id: params[:team_id], team_name: params[:team_name])
+        UI.message("Successfully logged in")
+      end
+
+      def self.create_certificate(params)
```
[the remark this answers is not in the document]
janpio: Hm, that might be a valid point - both the the action naming and the method naming here.

The action name is derived from `get_certificates` (main alias of `cert`). The action was called `apple_pay_cert` which felt wrong to me, so I chose to rename to `get_apple_pay_certificate`. If there is no actual "getting" in the code (not sure right now) we should probably rename to `create_apple_pay_certificate`.

The method naming I think is fair, it is just an internal method that does exactly that - create the certificate (which in the context of an `apple pay certificate` action of course is an apply pay one. But I can totally live with renaming that to `create_apple_pay_certificate` as well 🤷‍♀

fastlane/lib/fastlane/actions/get_apple_pay_certificate.rb
```diff
@@ -0,0 +1,154 @@
...
+                                     description: "Generate a p12 file additionally to the .pem file",
+                                     is_string: false,
+                                     default_value: true),
+          FastlaneCore::ConfigItem.new(key: :save_private_key,
+                                    short_option: "-s",
+                                    env_name: "APPLE_PAY_CERT_SAVE_PRIVATEKEY",
+                                    description: "Set to save the private RSA key",
+                                    is_string: false,
+                                    default_value: true)
+        ]
```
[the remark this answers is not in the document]
janpio: As rubocop and any other linters are happy with it how it is, nothing is wrong really. If we have other requirements here, we should add a rule that makes these explicit.

(That being said, yep that is a big ugly.)

spaceship/lib/spaceship/portal/certificate.rb
janpio: If that works, that would definitely be nice. Can you test it and report back?

gholias:
Do we have plans to merge this PR soon?

snatchev:
@gholias thank you for bump this.

While this PR does have some outstanding comments, and could be cleaned up, perhaps "perfect" is the enemy of "done" and we should move ahead.

pinging @mollyIV @joshdholtz for thoughts if this contains any hard blockers.

joshdholtz:
@gholias @snatchev I will test it today/tomorrow and see what we can do to push it through 😊 Just want to make sure those Apple APIs are still working as we expect 💪

gholias:
thanks @joshdholtz

fastlane/lib/fastlane/actions/get_apple_pay_certificate.rb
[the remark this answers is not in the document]
mollyIV: > Isn't it that way in other actions?

It depends, although I am quite sure there's a mix - some actions define an empty `SharedValues` module, some do not. So it's totally fine to leave an empty `SharedModule` here 👌

fastlane/lib/fastlane/actions/get_apple_pay_certificate.rb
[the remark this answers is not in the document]
mollyIV: Very well, thanks @janpio for the context, let's leave the naming as it is then 👍

spaceship/lib/spaceship/portal/certificate.rb
mollyIV: @RishabhTayal don't you mind testing the suggestion out? 🙏

mollyIV (approved):
Thanks a lot for making these changes 💯 

Please note that this piece of code is using `Spaceship` and recently a lot of changes have been done in that area due to broken API. It would be great to pull the newest changes and re-test the feature manually before merging to double check everything is still working fine 🙇 

Approved with minor comments 😊

gholias:
Any updates on this PR?

yhkaplan:
@joshdholtz @RishabhTayal Any updates on this? I can help if needed!

joshdholtz:
@yhkaplan Hey! A feature like this will probably end up getting slightly rewritten due to the new App Store Connect API endpoints. I'm planning to start the module that includes this sometime next week probably!

ryechus:
@joshdholtz any update? Do you need help?

joshdholtz:
@ryechus Wooooowwww... things slipped away from me didn’t they 😅 I... actually don’t know if there are new API endpoints for this yet so I just revisit this PR and probably merge it 😊 Have you been able to test this PR to see if it works for you?

whalemare:
Still needed

bartpowers:
Please see my PR for a simple addition that makes this all possible

PR #12472, opened 2018-05-08
[slack] add pretext to show text above attachment block

### LINKED ISSUE #12468: SlackAction does not show message when attachment text exists.
It seems that slack API changed and message was renamed into text.

Try:

[Slack API](https://api.slack.com/docs/messages/builder)

```json
{
    "text": "I am a test message http://slack.com",
    "attachments": [
        {
            "text": "And here’s an attachment!"
        }
    ]
}
```

I send use it as:

```ruby
  def growl_to_team(options)
    slack(
      message: options[:message],
      default_payloads: [],
      attachment_properties: {
          text: options[:text]
      },
      use_webhook_configured_username_and_icon: true,
      success: options[:success] || true
      )
  end
```

There is no message, only attachment_properties[:text] in slack channel.

joshdholtz:
Fixes #12468

## Wut
- `:message` is put into the "attachments" text so no way to put text above (outside) attachment block

## Fix
- Added in `:pretext` option which puts text above (outside) of attachment block

### With pretext
screen shot 2018-05-08 at 7 29 04 am

### Wihout pretext
screen shot 2018-05-08 at 7 43 11 am

### DIFF
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -71,6 +71,10 @@ module Fastlane
                                        env_name: "FL_SLACK_MESSAGE",
                                        description: "The message that should be displayed on Slack. This supports the standard Slack markup language",
                                        optional: true),
+          FastlaneCore::ConfigItem.new(key: :pretext,
+                                       env_name: "FL_SLACK_PRETEXT",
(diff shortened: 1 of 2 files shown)


taquitos (approved):
:balloon::goat:

PR #12499, opened 2018-05-11
fixed match does not recreate provisioning profile

wade0317:
bundle exec rspec

### Checklist

### Motivation and Context
 -  Why is this change required? What problem does it solve?

 	When the provisioning profile is invalid, it might be caused by deleting the provisioning profile on the Dev Portal.
 	fastlane does not work! fastlane match do not recreate provisioning when i set readonly is false , meanwhile, it do not  recreate provisioning when the provisioning file is invalid.

 -  If it fixes an open issue, please link to the issue here. 

    Not found.

-  Please describe in detail how you tested your changes. 

   I delete adhoc provisioning file in apple developer website, then i run **fastlane match adhoc**,
   It work!

### Description

  If  the provisioning file is invalid , it should not raise a exception right now. May be we need recreate the   the provisioning file.

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -71,6 +71,21 @@ module FastlaneCore
         destination
       end
 
+      # Installs a provisioning profile for Xcode to use
+      def uninstall(path, keychain_path = nil)
+        UI.message("Delete provisioning profile...")
+        unless path.nil?
+          profile_filename = uuid(path, keychain_path) + ".mobileprovision"
+          destination = File.join(profiles_path, profile_filename)
+          if File.exist?(destination)
+            File.delete(destination)
+          end
+          if File.exist?(path)
+            File.delete(path)
+          end
+        end
+      end
+
       private
 
       def decode(path, keychain_path = nil)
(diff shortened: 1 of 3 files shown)


joshdholtz:
@wade0317 Hey, quick question that I thought of while reviewing that might help me review this a bit better and make some suggestions... Without introducing this change, does setting `force: true` somewhat accomplish the same thing for you? 

I'm not saying `force: true` is the correct thing to do here but just curious if that is another solution to this problem 😊

joshdholtz:
@wade0317 While looking at this I wanted to try out a different approach to solving this problem. I created a separate PR for this if you don't mind taking a look 😊 #12539

wade0317:
@joshdholtz  Yes , even set  force as true , the provisioning profile will not be created. It seem that we both solve a same problem,  but i think it may be some more for adding a option with recreate_deleted_profiles, since it already has  readonly options. if readonly is false, it mean match need recreate something which is invalid.

Meanwhile ,  it should check if the provisioning profile if valid, then install it to the local directory. if match find a valid provisioning profile, it's bette delete it from the local directory, so i exchange some logic in

joshdholtz:
@wade0317 Weird, setting `force: true` should have definitely solved it for you 🤔 I was able to recreate your situation and setting `force: true` worked for me. Can you try pulling down PR #12539 and seeing if my new option (`recreate_deleted_profiles: true`) works for your situation?

I think we would like to add a new option to solve this as an opt-in approach reduces confusion in "new behavior" and will reduce of the risk regressions but changing simply adding a new path 🙃

wade0317:
@joshdholtz ok, I accept your PR ~  ~

joshdholtz:
Closing this in favor of https://github.com/fastlane/fastlane/pull/12539

PR #12513, opened 2018-05-12
Make sure the user is logged in RubyGems when releasing a new version

### LINKED ISSUE #12292: Verify that user is logged in to ruby gems in `verify_env_variables` private lane
`fastlane release` will fail if not logged into ruby gems (which happens at very end). This should get checked in the beginning in `verify_env_variables`

Idea from @revolter 😊

revolter:
Closes #12292.


joshdholtz (approved):
YOU ARE MY HERO ❤️ 💪

PR #12518, opened 2018-05-14
added loading indicator for frameit

RishabhTayal:
### Motivation and Context

Currently, the action `frameit` takes some time to complete. There is no loading indicator when the screenshots are being framed. It would be nice to show indicator while the action is doing its magic.

### Description

Before:
kcnq6ozwea

After:
dhwrdxylri

### DIFF
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -59,6 +59,7 @@ module Frameit
       output_path = screenshot.path.gsub('.png', '_framed.png').gsub('.PNG', '_framed.png')
       image.format("png")
       image.write(output_path)
+      Helper.hide_loading_indicator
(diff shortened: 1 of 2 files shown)


revolter:
Could it end in a finished state like the initial one ends with a checkmark?

RishabhTayal:
@revolter Not sure how that's possible. The initial load spinner code is different than the one used in rest of the project.

revolter:
I looked at the gif again and noticed that the dots remain, then a message is printed, then the checkmark is printed. So that message is being displayed too soon

RishabhTayal:
@revolter I have updated the PR. Now the loading indicator is showing inline with the message.

joshdholtz (approved):
Looks good 👍

PR #12524, opened 2018-05-15
Add attr_reader for :prefix in AccountManager

KrauseFx:
Just to make it publicly accessible, we don't need it now, but might be needed for http://fastlane.ci in the future


joshdholtz (approved):
I would have liked to see tests but... I guess I can let this slide 😉

PR #12539, opened 2018-05-17
[match] recreate profiles that were deleted from dev portal

joshdholtz:
## Description
- `match` will error out if provisioning profile was deleted from the dev portal but remained in the match repo
- This can be a bit cumbersome for users as they either need to:
  1. delete the profile manually from the repo
  2. run nuke
- This proposed feature will automatically recreate the deleted profile
  - The deleted profile essentially behaves like an "invalid" profile 

## Additional Thoughts?
- I currently set the new option to default to `false`  as to not change existing behavior under current users **BUT** this could be set to default to `true` as it _could_ prevent some confusion for new users

## Usage
```rb
lane :something do
  match(
    recreate_deleted_profiles: true
  )
end
```

### To use this branch in your project using a Gemfile...
```rb
source "https://rubygems.org"
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "joshdholtz-match-recreated-deleted-profiles"
```


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

revolter:
Is there a `match init`? If yes, it could set it to `true` for new inits. Maybe by simply adding it here: https://github.com/fastlane/fastlane/blob/master/match/lib/assets/MatchfileTemplate.

RishabhTayal:
Any update on this PR?

getaaron (changes requested):
@joshdholtz I think a default value of `true` would be better since it's more helpful, even if it might confuse a few people, it won't hurt anything

joshdholtz:
@janpio That is correct! It errors out if not remote copy which always felt a little bit weird 😖 This will allow _match_ to recreate a new profile if it isn't found remotely (like expected).

janpio:
Would the now deleted profile be of any use of _another development machine_? So if my colleague still has it, do I harm him in any way when my own match creates a new one? Or is this broken for him anyway already? 

What happens if my colleague still has the old one on his machine and runs match after it recreated the profile when I ran it - will this recognize that the remote one is a new one and not the one they still have on their machine?

match/lib/match/runner.rb
```diff
@@ -197,7 +197,7 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
       parsed = FastlaneCore::ProvisioningProfile.parse(profile, keychain_path)
       uuid = parsed["UUID"]
 
-      if spaceship && !spaceship.profile_exists(username: params[:username], uuid: uuid)
+      if spaceship && !spaceship.profile_exists(username: params[:username], uuid: uuid, return_if_missing: params[:recreate_deleted_profiles])
```
KrauseFx: The "renaming" of the parameter here from `recreate_deleted_profiles` to `return_if_missing` adds unnecessary complexity, can we keep the name?

match/lib/match/spaceship_ensure.rb
```diff
@@ -53,14 +53,17 @@ def certificate_exists(username: nil, certificate_id: nil)
...
+    def profile_exists(username: nil, uuid: nil, return_if_missing: false)
       found = Spaceship.provisioning_profile.all.find do |profile|
         profile.uuid == uuid
       end
 
       unless found
         UI.error("Provisioning profile '#{uuid}' is not available on the Developer Portal")
         UI.error("for the user #{username}")
+
+        return false if return_if_missing
```
KrauseFx: this would be a chance to tell the user about the parameter they can set, right?

KrauseFx (approved):
Overall this is a good change, and I like that it's not on by default yet. I ran into this problem myself in the past. I added a few comments.

KrauseFx:
I'd also be ok setting this on by default, and making sure to monitor GH issues after shipping

getaaron:
this should be on by default (can’t think of a good reason to have it off — I almost think it shouldn’t even be an option)

@janpio no, once deleted the profile is of no use on another dev machine. And the new profile would replace the old one 👍🏼

match/lib/match/options.rb
```diff
@@ -111,6 +111,11 @@ def self.available_options
                                      description: "Renew the provisioning profiles every time you run match",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :recreate_deleted_profiles,
+                                     env_name: "MATCH_CREATE_MISSING_PROFILES",
+                                     description: "Automatically recreate provisioning profiles that were deleted from the Dev Portal",
+                                     is_string: false,
+                                     default_value: false),
```
getaaron: Should be true IMO

getaaron (changes requested):

KrauseFx:
Yah I'm ok with what @getaaron said, let's make it the default, and not even an option

getaaron (approved):

getaaron:
🛳

joshdholtz:
@getaaron WOOAAAAHH DDUUUDDEEEEE, that was quick!

PR #12545, opened 2018-05-17
Update docs on gym's :export_options action parameter

aleh:
The :export_options cannot be a path to a plist anymore when running gym as a Fastlane action.
It was OK before when this config item was consumed only to prepare a file for xcodebuild
where it would be read from disk in case it was not a hash. However now it is used in
DetectValues.set_additional_default_values which does not expect it to be a non-hash.

### Checklist

### DIFF
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -111,7 +111,7 @@ module Gym
                                      end),
         FastlaneCore::ConfigItem.new(key: :export_options,
                                      env_name: "GYM_EXPORT_OPTIONS",
-                                     description: "Path to an export options plist or a hash with export options. Use 'xcodebuild -help' to print the full set of available options.",
+                                     description: "Path to an export options plist or a hash with export options. Use 'xcodebuild -help' to print the full set of available options",
                                      is_string: false,
                                      optional: true,
                                      type: Hash,
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -111,7 +111,7 @@ module Gym
                                      end),
         FastlaneCore::ConfigItem.new(key: :export_options,
                                      env_name: "GYM_EXPORT_OPTIONS",
-                                     description: "A path to an export options plist when running gym from the command line; a hash with export options when using gym as a Fastlane action. Use 'xcodebuild -help' to print the full set of available options.",
+                                     description: "Path to an export options plist or a hash with export options. Use 'xcodebuild -help' to print the full set of available options.",
                                      is_string: false,
                                      optional: true,
                                      type: Hash,
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -111,7 +111,7 @@ module Gym
                                      end),
         FastlaneCore::ConfigItem.new(key: :export_options,
                                      env_name: "GYM_EXPORT_OPTIONS",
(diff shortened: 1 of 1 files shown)


aleh:
Signed as dzenisiuk@gmail.com

gym/lib/gym/options.rb
```diff
@@ -111,7 +111,7 @@ def self.plain_options
                                      end),
         FastlaneCore::ConfigItem.new(key: :export_options,
                                      env_name: "GYM_EXPORT_OPTIONS",
-                                     description: "Specifies path to export options plist. Use 'xcodebuild -help' to print the full set of available options",
+                                     description: "A path to an export options plist when running gym from the command line; a hash with export options when using gym as a Fastlane action. Use 'xcodebuild -help' to print the full set of available options.",
```
joshdholtz: @aleh Can we change this to...
```
Path to an export options plist or a hash with export options. Use 'xcodebuild -help' to print the full set of available options
```

joshdholtz (changes requested):
Using a plist path does actually work when using a `Fastfile` - I just verified that it worked locally 😊 

But I think this description change is still needed 👌

gym/lib/gym/options.rb
aleh: Thanks for the review!

Which version or branch were you using in your test? Did you use `gym` action directly or `build_ios_app` instead (the latter at least checks `if values[:export_options].kind_of?(Hash)`).

Passing a path certainly does not work with gym action in 2.95.0 when exporting for the App Store. It did work about a half year ago indeed when the export options were not used in the action itself, but now they are used in `detect_selected_provisioning_profiles` without any checking for type even before `:skip_profile_detection` is checked for (so there is no way to work this around).

```
    def self.detect_selected_provisioning_profiles
      Gym.config[:export_options] ||= {}
      hash_to_use = (Gym.config[:export_options][:provisioningProfiles] || {}).dup || {} # dup so we can show the original values in `verbose` mode

      unless Gym.config[:skip_profile_detection]
```

joshdholtz: @aleh I used latest master (2.95.0-ish) with this in my `Fastfile`...

```rb
gym(
      scheme: 'Test',
      export_method: 'app-store',
      export_options: 'fastlane/ExampleExport.plist',
)
```

And that code example you provided doesn't "work" with a string path getting passed in but there is a `rescue` in that method that catches things which does allow `gym` to continue on with its process.

You can see in the below screenshot how my plist is getting used 🙃 

screen shot 2018-05-18 at 5 13 11 am

gym/lib/gym/options.rb
```diff
@@ -111,7 +111,7 @@ def self.plain_options
                                      end),
         FastlaneCore::ConfigItem.new(key: :export_options,
                                      env_name: "GYM_EXPORT_OPTIONS",
-                                     description: "A path to an export options plist when running gym from the command line; a hash with export options when using gym as a Fastlane action. Use 'xcodebuild -help' to print the full set of available options.",
+                                     description: "Path to an export options plist or a hash with export options. Use 'xcodebuild -help' to print the full set of available options.",
```
joshdholtz: Tests are failing because of the "." at the end of this description 🙃

joshdholtz (changes requested):

aleh:
@joshdholtz Updated.

joshdholtz (approved):
Woot! Thank you 😊

aleh:
Thanks!

Issue #12577, opened 2018-05-21
[!] The request could not be completed because: {"authType"=>"sa"}


danivegamx:
### Question Checklist

- [ ✅ ] Updated fastlane to the latest version
- [ ✅ ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ✅ ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ ✅ ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
Running fastlane for an adhoc distribution is not working. Error below.

#### Question Description

When I run via `bundle exec fastlane beta` I got this:

[!] The request could not be completed because: {"authType"=>"sa"} aa=5A52AF019EF22EE5C2D8BBA8AADA6E4F; Domain=idmsa.apple.com; Path=/; Secure; HttpOnly, dslang=US-EN; Domain=apple.com; Path=/; Secure; HttpOnly, site=USA; Domain=apple.com; Path=/; Secure; HttpOnly, acn01=57NJeJgkumOF6rAbV4AvInJhJaP4TxVrExywB7p+W1pyLjWuY0gBrxkAECrx0bG3+w==; Max-Age=31536000; Expires=Tue, 21-May-2019 22:41:58 GMT; Domain=apple.com; Path=/; Secure; HttpOnly

This is my configuraiton file:

```

default_platform(:ios)

lane :beta do
    get_certificates
    get_provisioning_profile(
        username: ENV['APPLE_TEAM_DEV_ID'],
        adhoc: true,
        filename: "AdHoc_com.company.mobileprovision",
        app_identifier: "com.company"
    )
    get_info_plist_path(
        build_configuration_name: "staging",
        target: "staging"
    )
    increment_version_number_in_plist(
        target: "staging",
        bump_type: "patch"
    )
    gym(
        scheme: "staging",
        export_method: "ad-hoc",
        export_xcargs: "-allowProvisioningUpdates"
    )
    aws_s3(
        access_key: ENV['S3_ACCESS_KEY'],
        secret_access_key: ENV['S3_SECRET_ACCESS_KEY'],
        bucket: "bucket-name",
        region: "region",
        ipa: "appname.ipa",
        app_directory: "beta"
    )
end
```

Any idea?

AaronSofaer:
[...]

PR #12578, opened 2018-05-22
Fix for handling Apple's new Apple ID and Privacy statement

joshdholtz:
Fixes #12577

## Problem
- Handles status code `412` (precondition error) with body of `{"authType": "sa"}` and `{"authType": "hsa"}`

PR #12590, opened 2018-05-22
[spaceship] display apple id and privacy statement instructions to non-sa accounts

joshdholtz:
Follow up on #12578 
Fixes https://github.com/fastlane/fastlane/issues/12577#issuecomment-390980837


taquitos (approved):
:saxophone::goat:

PR #12605, opened 2018-05-23
[spaceship] Updated Tunes.client so select_team returns team_id when executed just like Portal.client does

kumichou:
### Checklist

### Motivation and Context
When using Spaceship directly, and selecting a team based on the name, it was convenient to get the team_id back in Portal to use in any other calls. However, Tunes didn't work that way and just returned nil so you needed to make an extra call to Tunes.client.team_id to get the value.

### DIFF
--- a/spaceship/spec/tunes/tunes_client_spec.rb
+++ b/spaceship/spec/tunes/tunes_client_spec.rb
@@ -119,14 +119,9 @@ describe Spaceship::TunesClient do
       end
 
       it "returns team_id from environment variable" do
-        ENV["FASTLANE_ITC_TEAM_NAME"] = "Harry"
+        stub_const('ENV', { 'FASTLANE_ITC_TEAM_NAME' => 'Harry' })
         allow(subject).to receive(:teams).and_return(associated_teams)
         expect(subject.select_team).to eq('5678')
-      end
-
-      after do
-        ENV.delete("FASTLANE_ITC_TEAM_ID")
-        ENV.delete("FASTLANE_ITC_TEAM_NAME")
       end
     end
   end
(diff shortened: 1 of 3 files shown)


kumichou:
I signed it!

spaceship/spec/tunes/tunes_client_spec.rb
```diff
@@ -104,6 +104,31 @@
...
+        expect(subject.team_id).to eq('1234')
+      end
+
+      it "returns team_id from legitimate team_name parameter" do
+        allow(subject).to receive(:teams).and_return(associated_teams)
+        expect(subject.select_team(team_name: 'Harry')).to eq('5678')
+      end
+
+      it "returns team_id from environment variable" do
+        ENV["FASTLANE_ITC_TEAM_NAME"] = "Harry"
```
joshdholtz: You can probably use `stub_const` to do this 😊 
```rb
stub_const('ENV', { 'FASTLANE_ITC_TEAM_NAME' => 'Harry })
```

spaceship/spec/tunes/tunes_client_spec.rb
```diff
@@ -104,6 +104,31 @@
...
+        allow(subject).to receive(:teams).and_return(associated_teams)
+        expect(subject.select_team).to eq('5678')
+      end
+
+      after do
+        ENV.delete("FASTLANE_ITC_TEAM_ID")
+        ENV.delete("FASTLANE_ITC_TEAM_NAME")
```
joshdholtz: You can delete these if you use `stub_const`

joshdholtz (changes requested):
One small change request 😊

kumichou:
Thanks for the feedback. I didn't realize I could do that so updated the code in the RSpec test.

kumichou:
@joshdholtz Didn't know if I needed to tag you or not that changes you requested have been implemented.

joshdholtz (approved):
Thanks for the contribution! 😊

PR #12614, opened 2018-05-24
Login is the default keychain name

dgyesbreghs:
### Checklist

### Motivation and Context
Because macOS default KeyChain is called 'login', we can define a default value to the unlock_keychain action.
I've tested this with our own Jenkins instance.

### Description
Provide a default path value 'login.keychain' to the unlock_keychain action and make it optional.


fastlane/lib/fastlane/actions/unlock_keychain.rb
```diff
@@ -72,7 +72,8 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :path,
                                        env_name: "FL_UNLOCK_KEYCHAIN_PATH",
                                        description: "Path to the keychain file",
-                                       optional: false),
+                                       default_value: "login.keychain",
+                                       optional: true),
```
joshdholtz: Can we change this to `optional: false`? We can still have a `default_value` but we don't want this to ever be `nil` 😊

joshdholtz (changes requested):
One small change! Can we also add some tests to `unlock_keychain_spec.rb`?

fastlane/lib/fastlane/actions/unlock_keychain.rb
dgyesbreghs: Fixed it 🙂

dgyesbreghs:
@joshdholtz I've created a new test to make sure everything is working.

joshdholtz (changes requested):
Can you delete the changes to `fastlane/spec/fixtures/xcodeproj/automatic_code_signing.xcodeproj/project.pbxproj`? Its on my list to prevent this change from happening 😉 Its an issue with our tests

joshdholtz:
@dgyesbreghs Also, one quick question about this... Won't most login keychains still require the user to enter a password for the keychain to unlock it?

joshdholtz (approved):
Looks good! 👍

PR #12618, opened 2018-05-25
spaceship: allow to remove tester(s) from testflight

lacostej:
When you import the wrong list of testers, you have to clean up your mess.

### DIFF
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -202,6 +202,17 @@ module Spaceship
         handle_response(response)
       end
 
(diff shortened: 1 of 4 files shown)


joshdholtz (changes requested):
Can we get a test added to `test_flight/client_spec.rb`to mock and test the `providers/#{team_id}/apps/#{app_id}/deleteTesters` route? 😊 Similar to how `delete_tester_from_app` is tested

lacostej:
@joshdholtz changes applied! Thanks for the feedback

joshdholtz (approved):
Tested this out and it works 💯 Thanks for doing this! 💪

PR #12632, opened 2018-05-28
abort on interrupt to set exit code to 1

### LINKED ISSUE #12629: SIGINT shouldn't return status code =  0
### New Issue Checklist

### Issue Description

If fastlane is running a lane and user presses `ctrl+c` (sends SIGINT) fastlane stops the lane but the `$?` (last command status) is `0`. This is definitely wrong as the process didn't complete, it was interrupted. This bug is causing unexpected issues for me when using fastlane with sidekiq.

In my sidekiq worker class, I build ipa file (using a custom lane in fastfile), if the command status code is 0 then I enqueue another job to upload ipa file. When my custom lane is running, and if I stop sidekiq by pressing `ctrl+c` then build ipa file is stopped, as the status code is `0`; upload job is enqueued and then the sidekiq shutdowns.

joshdholtz:
Fixes #12629

## Wut
- The interrupt error caught was capturing the error which causing the exit code to be `0` (which is shouldn't be)
- The `puts` that was giving nice message to the user is not an `abort` (an exit code of `1` that displays same message)

### DIFF
--- a/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
+++ b/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
@@ -90,7 +90,7 @@ module Commander
           raise e
         else
           action_completed(@program[:name], status: FastlaneCore::ActionCompletionStatus::INTERRUPTED, exception: e)
-          puts("\nCancelled... use --verbose to show the stack trace")
+          abort("\nCancelled... use --verbose to show the stack trace")
         end
       rescue \
         OptionParser::InvalidOption,


fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb
```diff
@@ -90,7 +90,7 @@ def run!
           raise e
         else
           action_completed(@program[:name], status: FastlaneCore::ActionCompletionStatus::INTERRUPTED, exception: e)
-          puts("\nCancelled... use --verbose to show the stack trace")
+          abort("\nCancelled... use --verbose to show the stack trace")
```
taquitos: I think this is the right thing to do, will this break anybody's setups? I don't think so, worst case is it stops a lane from running due to exit code being 1, which is probably the right behavior.

joshdholtz: @taquitos This _shouldn't_ break anyone setup 😊 This only get's called if a user presses CONTROL+C so the user knows things are going to stop. This fixes an issue when a CI/worker/queue kills a job running fastlane where it actually returns an error like it should. This _may_ have some benefit forward fastlane.ci? Unsure since I don't know how those jobs are started up yet and what happens if manually cancelled 🙃

taquitos (approved):
👍

Issue #12604, opened 2018-05-23
Allow two instances of FastlaneRunner running at same time in Swift


josesm82:
#### Motivation Behind Feature
I'd like to have the possibility of run several instances of FastlaneRunner al same time to allow Jenkins to run concurrent builds. Right now, when I have one instance of fastlane running, the next one gets this error: 
fastlane failed to receive a connection from the FastlaneRunner binary after 5 seconds, shutting down
After dig a little bit, I found this error is being generated because it tries to open a socket in port 2000 which is being used by the other instance

#### Feature Description
Fastlane is listening to FastlaneRunner on port 2000, on the FastlaneSwiftRunner code you can change this port easily but on the ruby side there is no option to set this port as parameter. This value is set on socket_server.rb when the server starts. Id could be great to have the option to set this port as parameter so we can choose a different port to listen and avoid this issue. 

#### Alternatives or Workarounds
I'm not allowing Jenkins to have concurrent buils 

fastlane env output:

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.13.4                                                        |
| Ruby                        | 2.3.3                                                          |
| Bundler?                    | true                                                           |
| Git                         | git version 2.15.1 (Apple Git-101)                             |
| Installation Source         | /Library/Ruby/Gems/2.3.0/bin/fastlane                          |
| Host                        | Mac OS X 10.13.4 (17E202)                                      |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib |
| OpenSSL Version             | LibreSSL 2.2.7                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version               | 9.3.1                                                          |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE |             |   |

### fastlane files:

<details><summary>`./fastlane/Fastfile.swift`</summary>

```ruby
// This file contains the fastlane.tools configuration
// You can find the documentation at https://docs.fastlane.tools
//
// For a list of all available actions, check out
//
[...]

PR #12634, opened 2018-05-29
Allow parameterize port

josesm82:
### Checklist

### Motivation and Context

This change allows to run several instances of fastlane with FastlaneRunner. Until now, there were using always the port 2000 to communicate, so I was unable to run more that one instance in the same server because the port was in use. This PR allow to set the port which will be used. This param is optional and it could be sent only if you need it. The port could be set using same way than the other params: 
bundle exec fastlane myLane port:2001
If port is set, both codes use it instead of the default one (2000)

This fixes this issue: 
https://github.com/fastlane/fastlane/issues/12604

### Description

Both fastlane and FastlaneSwiftRunner checks if param port is set. If so, they use this port instead of default to communicate.

### DIFF
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -6,7 +6,6 @@ module Fastlane
     def self.handle(args, options)
       lane_parameters = {} # the parameters we'll pass to the lane
       platform_lane_info = [] # the part that's responsible for the lane/platform definition
-      port = 2000
       args.each do |current|
         if current.include?(":") # that's a key/value which we want to pass to the lane
           key, value = current.split(":", 2)
@@ -14,9 +13,6 @@ module Fastlane
           value = convert_value(value)
           UI.verbose("Using #{key}: #{value}")
           lane_parameters[key.to_sym] = value
-          if key.include?("port")
-            port = value
-          end
         else
           platform_lane_info << current
         end
@@ -34,8 +30,8 @@ module Fastlane
 
       if FastlaneCore::FastlaneFolder.swift?
         disable_runner_upgrades = options.disable_runner_upgrades || false
-
-        Fastlane::SwiftLaneManager.cruise_lane(lane, port, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades)
+        swift_server_port = options.swift_server_port || 2000
+        Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, swift_server_port, disable_runner_upgrades: disable_runner_upgrades)
       else
         Fastlane::LaneManager.cruise_lane(platform, lane, lane_parameters, dot_env)
       end
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -31,11 +31,7 @@ module Fastlane
       if FastlaneCore::FastlaneFolder.swift?
         disable_runner_upgrades = options.disable_runner_upgrades || false
         swift_server_port = options.swift_server_port
-        if swift_server_port
-          Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades, swift_server_port: swift_server_port)
-        else
-          Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades)
-        end
+        Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades, swift_server_port: swift_server_port)
       else
         Fastlane::LaneManager.cruise_lane(platform, lane, lane_parameters, dot_env)
       end
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -30,8 +30,12 @@ module Fastlane
 
       if FastlaneCore::FastlaneFolder.swift?
         disable_runner_upgrades = options.disable_runner_upgrades || false
-        swift_server_port = options.swift_server_port || 2000
-        Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades, swift_server_port: swift_server_port)
+        swift_server_port = options.swift_server_port
+        if swift_server_port
+          Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades, swift_server_port: swift_server_port)
+        else
+          Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades)
+        end
       else
         Fastlane::LaneManager.cruise_lane(platform, lane, lane_parameters, dot_env)
       end
--- a/fastlane/lib/fastlane/command_line_handler.rb
+++ b/fastlane/lib/fastlane/command_line_handler.rb
@@ -6,6 +6,7 @@ module Fastlane
     def self.handle(args, options)
       lane_parameters = {} # the parameters we'll pass to the lane
       platform_lane_info = [] # the part that's responsible for the lane/platform definition
+      port = 2000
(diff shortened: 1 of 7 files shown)


josesm82:
I signed it!

2018-05-29 10:10 GMT+02:00 googlebot <notifications@github.com>:

joshdholtz (changes requested):
Just a small thought on how port should be set 😊

josesm82:
Hi @joshdholtz , thanks for the comments. I committed a few changes following your suggestions:
https://github.com/fastlane/fastlane/pull/12634/commits/113e9e1a745ca2d908b8b566c45374f36d01d009

fastlane/lib/fastlane/swift_lane_manager.rb
```diff
@@ -6,7 +6,7 @@ class SwiftLaneManager < LaneManagerBase
     # @param lane_name The name of the lane to execute
     # @param parameters [Hash] The parameters passed from the command line to the lane
     # @param env Dot Env Information
-    def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false)
+    def self.cruise_lane(lane, parameters = nil, env = nil, swift_server_port, disable_runner_upgrades: false)
```
joshdholtz: Can we change this to?
```rb
def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false, swift_server_port: nil)
```

Main reason being for backwards compatibility if someone is calling this manually in some implementation. Move to end to make sure `disable_runner_upgrades` stays in the same position and add a default of `nil` so this doesn't "need" be passed in since its being default to `2000` later on.

fastlane/lib/fastlane/server/socket_server.rb
```diff
@@ -32,12 +32,12 @@ def initialize(
     end
 
     # this is the public API, don't call anything else
-    def start
-      listen
+    def start(port: 2000)
```
joshdholtz: Can we pass the port in through the initializer instead?

fastlane/lib/fastlane/swift_lane_manager.rb
```diff
@@ -216,14 +218,14 @@ def self.link_user_configs_to_project(updated_message: nil)
       return project_modified
     end
 
-    def self.start_socket_thread
+    def self.start_socket_thread(port: 2000)
```
joshdholtz: We can probably remove this default of `2000` since its being default in `socket_server.rb`? I would just set `port: nil`

fastlane/swift/Runner.swift
```diff
@@ -60,12 +60,12 @@ class Runner {
 
 // Handle threading stuff
 extension Runner {
-    func startSocketThread() {
+    func startSocketThread(port: UInt32 = 2000) {
```
joshdholtz: Is there a way to remove this default to `2000` if its being defaulted in `ArgumentProcessor.swift` already?

joshdholtz (changes requested):
Looking good! Just a few small nit picks on how port is getting set and passed around to make future maintenance easier 😊

fastlane/swift/Runner.swift
josesm82: Yes! You're right.

fastlane/lib/fastlane/swift_lane_manager.rb
josesm82: Done!

josesm82:
Hi @joshdholtz  I really appreciate your comments. This is the first time I'm working on a Ruby project and It is a little bit hard for me to understand some things. I committed a new patch according to your comments. And again, thanks a lot for helping me :)

josesm82 (commented):
Thanks for your comments

fastlane/lib/fastlane/server/socket_server.rb
```diff
@@ -134,8 +136,8 @@ def send_response(json)
       return COMMAND_EXECUTION_STATE[:ready]
     end
 
-    def listen
-      @server = TCPServer.open('localhost', 2000) # Socket to listen on port 2000
+    def listen(port)
```
joshdholtz: We can probably remove the `@port` as a parameter here since its already an instance variable? 🙃

fastlane/lib/fastlane/server/socket_server.rb
```diff
@@ -29,15 +30,16 @@ def initialize(
...
 
     # this is the public API, don't call anything else
     def start
-      listen
+      listen(@port)
 
       while @stay_alive
         UI.important("stay_alive is set to true, restarting server")
-        listen
+        listen(@port)
```
joshdholtz: Don't need to pass in `@port` here if we remove as a parameter below

fastlane/lib/fastlane/server/socket_server.rb
```diff
@@ -29,15 +30,16 @@ def initialize(
       @return_value_processor = return_value_processor
       @connection_timeout = connection_timeout.to_i
       @stay_alive = stay_alive
+      @port = port.to_i
     end
 
     # this is the public API, don't call anything else
     def start
-      listen
+      listen(@port)
```
joshdholtz: Don't need to pass in `@port` here if we remove as a parameter below

fastlane/lib/fastlane/command_line_handler.rb
```diff
@@ -30,8 +30,8 @@ def self.handle(args, options)
 
       if FastlaneCore::FastlaneFolder.swift?
         disable_runner_upgrades = options.disable_runner_upgrades || false
-
-        Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades)
+        swift_server_port = options.swift_server_port || 2000
```
joshdholtz: I think we should also remove this `|| 2000` and put it into as the default in the `cruise_lane` method

joshdholtz (changes requested):
This is 💯 Just a few nit picks and then we will be good to merge 😊

fastlane/lib/fastlane/swift_lane_manager.rb
```diff
@@ -6,7 +6,7 @@ class SwiftLaneManager < LaneManagerBase
     # @param lane_name The name of the lane to execute
     # @param parameters [Hash] The parameters passed from the command line to the lane
     # @param env Dot Env Information
-    def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false)
+    def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false, swift_server_port: 2000)
```
joshdholtz: Ah, my bad! For some reason I thought if a `nil` was passed in that it would still default to `2000`. Let's change this too...

```rb
def self.cruise_lane(lane, parameters = nil, env = nil, disable_runner_upgrades: false, swift_server_port: nil)
  swift_server_port ||= 2000
```

This will clean up `command_line_handler.rb` (see other comment)

fastlane/lib/fastlane/command_line_handler.rb
```diff
@@ -30,8 +30,12 @@ def self.handle(args, options)
...
       if FastlaneCore::FastlaneFolder.swift?
         disable_runner_upgrades = options.disable_runner_upgrades || false
-
-        Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades)
+        swift_server_port = options.swift_server_port
+        if swift_server_port
+          Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades, swift_server_port: swift_server_port)
+        else
+          Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades)
+        end
```
joshdholtz: We can get rid of the `if swift_server_port` and just make this... (see notes from other comment)

```rb
swift_server_port = options.swift_server_port
Fastlane::SwiftLaneManager.cruise_lane(lane, lane_parameters, dot_env, disable_runner_upgrades: disable_runner_upgrades, swift_server_port: swift_server_port)
```

joshdholtz (changes requested):
Two small things changes and then just need to run rubocop and we should be 💯

josesm82:
Hi @joshdholtz , thanks a lot for all your comments. I'm learning a lot!

joshdholtz (approved):
Thanks for making all of my requested changes ❤️ This is 💯

PR #12640, opened 2018-05-29
[action] add mute to say action (with rework)

### LINKED ISSUE #11750: Mute Siri say command
### New Issue Checklist

### Issue Description
I use Siri say command to notify about possible errors or mistakes that I've made in release distribution ( oh, CI ).

However, I would like to define local_say function which could check fast lane option "mute_siri" or something to mute Siri.

Is it possible to do via fast lane tools?

Thanks!

My possible solution:

```ruby
def say_or_write(message)
    if ENV["SIRI_ALLOWED"]
       say message
    else
       puts message
    end
end
```

joshdholtz:
Fixes #11750

## Wut
- Reworked how `say` method gets overridden
  - Gets called as an action now to allow `params` in `actions/say.rb` to actually be config items 
- Added option for `mute` which ends up just calling `UI.message`
  - `Fastfile` option, command line, environment variable
- Added more tests

### DIFF
--- a/fastlane/lib/fastlane/actions/say.rb
+++ b/fastlane/lib/fastlane/actions/say.rb
@@ -2,16 +2,36 @@ module Fastlane
   module Actions
     class SayAction < Action
       def self.run(params)
-        text = params.join(' ') if params.kind_of?(Array) # that's usually the case
-        text = params if params.kind_of?(String)
-        UI.user_error!("You can't call the `say` action as OneOff") unless text
+        text = params[:text]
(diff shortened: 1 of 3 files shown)


taquitos (approved):
Looks backwards compatible too, that's cool 👍 
🎈🐐

lolgear:
@joshdholtz 
It is not backward-compatible.

Fastlane::Actions::SayAction.run now accepts hash instead of plain text, array or ( even! ) variadic parameters.

Also, documentation is not up-to-date.

PR #12668, opened 2018-06-01
Raise error when submission fails on export compliance

luikore:
### Checklist

### Motivation and Context

When I submit an app for the first time with spaceship, it returns success but store status is still "prepare for submission". After debugging into fastlane I found that there was in fact an error in exportCompliance, though ignored.

### Description
Make the client also check the `exportCompliance` section for error keys so it will raise error when submit fails.

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -1019,11 +1019,14 @@ module Spaceship
       handle_itc_response(r.body)
 
       # iTunes Connect still returns a success status code even the submission
-      # was failed because of Ad ID info. This checks for any section error
-      # keys in returned adIdInfo and prints them out.
+      # was failed because of Ad ID Info / Export Complicance. This checks for any section error
+      # keys in returned adIdInfo / exportCompliance and prints them out.
       ad_id_error_keys = r.body.fetch('data').fetch('adIdInfo').fetch('sectionErrorKeys')
+      export_error_keys = r.body.fetch('data').fetch('exportCompliance').fetch('sectionErrorKeys')
       if ad_id_error_keys.any?
         raise "Something wrong with your Ad ID information: #{ad_id_error_keys}."
+      elsif export_error_keys.any?
+        raise "Something wrong with your Export Complicance: #{export_error_keys}"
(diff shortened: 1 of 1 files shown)


joshdholtz:
@luikore This is a great find 🙌 

Do you happen to know if its possible to get the actual questions to the things that were missed? This is the response I got while testing this out...
```sh
[14:46:42]: Something wrong with your Export Complicance: ["You must provide an answer for this question.", "You must provide an answer for this question."]
```

luikore:
@joshdholtz 

For my case (no encryption). After some comparison with correct requests / responses, I found setting all these options work:

```ruby
  submission.export_compliance_available_on_french_store = ""
  submission.export_compliance_contains_proprietary_cryptography = ""
  submission.export_compliance_contains_third_party_cryptography = ""
  submission.export_compliance_uses_encryption = "false"
  submission.export_compliance_is_exempt = ""

  if submission.export_compliance_compliance_required
    submission.export_compliance_encryption_updated = "false"
  else
     submission.export_compliance_encryption_updated = nil
  end
```

joshdholtz:
@luikore Do you mean those options work to provide a nicer error message? Or those options work to stop the error message from happening? 🙃

luikore:
@joshdholtz

I mean, if I don't set these submission options, it fails in Appstore for the first submission of an app. But fastlane doesn't report the failure.

With the change in this PR, fastlane can report this failure of export compliance.

With the change in this PR and submission options in my previous comment, Appstore succeeds and fastlane reports success.

joshdholtz:
@luikore Got it got it, that is what I was able to see as well with this PR. My only concern was when I tested this and got the failure of export compliance, the error message didn't actually say what the errors were for. It just said `"You must provide an answer for this question."` two times which is better than what we currently have but it didn't say **what** question we needed to provide an answer for 🙃

luikore:
@joshdholtz 

I think there can be different invalid combinations, but Apple doesn't show more helpful info in the response. The only thing we can get is: whether there is error, which section is invalid. That's what I added in this PR.

joshdholtz (approved):
I think this should work for now! Would be nice to have a more helpful error message from Apple but at least this will fail properly 😊

PR #12684, opened 2018-06-04
Make Carthage action support dependencies for bootstrap command

msanders:
### Checklist

### Motivation and Context

With Carthage you can specify which dependencies should be built, e.g. you can run `carthage bootstrap Alamofire` to only build Alamofire and it's dependencies. Currently the Carthage action ignores the dependencies parameter when for the `bootstrap` command. This PR adds support for dependencies for this command.

I used the same tests as the update and build commands, and just changed it to use bootstrap instead of build/update.


msanders:
I signed it!

joshdholtz (approved):
Thanks for the first time contribution ❤️  This looks 💯

PR #12716, opened 2018-06-07
Prevent automatic_code_signing spec from changing test fixture

joshdholtz:
Spec for `automatic_code_signing` was leaving some changes to a test fixture which sometimes could easily get accidentally committed 🙃 

So this fixes that 😉


KrauseFx (approved):
🐇

PR #12720, opened 2018-06-08
[mailgun] Add ability to pass custom placeholders

lisional:
### Checklist

### Motivation and Context
Presently the mailgun action uses an erb template to send as message body. Placeholders in this erb template cannot be extended; they are based on parameters such as **:message**, **:app_link**, **git author of the last commit**, etc. Mailgun put them in an internal **Hash** to use them (see [mailgun.rb:134](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/mailgun.rb#L134))
I would like to be able to add other placeholders in my erb template and have mailgun action populate them.

I have tested this locally and was able to completely modify the erb template and populating it properly.
I also tried to use the mailgun action without specifying this new parameter and everything worked fine thanks to the default empty Hash value.

### Description
I have added a new Hash parameter, **custom_placeholders** with an empty Hash default value.
This parameter is then merged with the internal hash used by the mailgun action before rendering the final html file.
I have looked at the **erb** action and took [this parameter](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/erb.rb#L42) as an example

### DIFF
--- a/fastlane/lib/fastlane/actions/mailgun.rb
+++ b/fastlane/lib/fastlane/actions/mailgun.rb
@@ -91,8 +91,14 @@ module Fastlane
                                       env_name: "MAILGUN_ATTACHMENT",
                                       description: "Mail Attachment filenames, either an array or just one string",
                                       optional: true,
-                                      is_string: false)
-
+                                      is_string: false),
+          FastlaneCore::ConfigItem.new(key: :custom_placeholders,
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Tested this out and works great! Thanks for adding this 😊 ❤️

Issue #12750, opened 2018-06-14
Question - How to add an Internal Tester to app?


RishabhTayal:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
action: `pilot`

#### Question Description
This is a follow up question from existing issue #7115. 

In the comment https://github.com/fastlane/fastlane/issues/7115#issuecomment-262048703 it mentions to use `tester.add_to_app!(app.apple_id)` to assign an existing internal tester to the app. However I don't see this method in the latest version of `fastlane` (2.97.0). 

Was this method removed? If yes, what is the alternative way to assign internal tester to an app?

**PS:** I also don't see any `Internal` class as mentioned in the same comment
`tester = Spaceship::Tunes::Tester::Internal.find(tester_email)`. Hence I am using `tester = Spaceship::TestFlight::Tester.search(app_id: app.apple_id, text: tester_email).first` to search for a tester.

RishabhTayal:
After discussing with @joshdholtz, seems like we need a new method to get the internal users from App Store Connect. After the new method is available we can use the following code to get the app store connect users and add them to the internal testers group.

```ruby
  client = Spaceship::Tunes.login(admin_email, password)
  client.select_team(team_name: team_name)

  app = Spaceship::Tunes::Application.find(app_identifier)

  client = Spaceship::TestFlight::Base.client
  user = client.internal_users(app.apple_id).select { |user|
 	user['email'] == 'tester_email'
  }.first

  tester = Spaceship::TestFlight::Tester.new
  tester.first_name = user['firstName']
  tester.last_name = user['lastName']
  tester.email = user['email']

  groups = Spaceship::TestFlight::Group.add_tester_to_groups!(tester: tester, app: app, groups: ['ITC.apps.tf.labels.internalTesters'])
```

I am sure there is a better way to write the above code. Not a ruby expert. Would like to hear how we can improve the above code snippet.

However the idea is to use the `client.internal_users` method to find the users and then `Group.add_tester_to_groups!` method to add the tester to the group. 👍

PR #12752, opened 2018-06-14
testflight - added method to find internal users

RishabhTayal:
### Checklist

### Description
This is followup pull request after discussing with @joshdholtz. Fixes issue https://github.com/fastlane/fastlane/issues/12750.

Added new method `internal_users` to get the internal users from App Store Connect.

### DIFF
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -171,7 +171,7 @@ module Spaceship
         url = "providers/#{team_id}/apps/#{app_id}/internalUsers"
         r = request(:get, url)
 
-        internal_users = parse_response(r, 'data')
+        parse_response(r, 'data')
       end
 
       ##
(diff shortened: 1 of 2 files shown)


joshdholtz (changes requested):
Can we get some tests written for this new method? 😊

RishabhTayal:
@joshdholtz could you please help me look into the test case. Why it's failing?

heartofagoof:
@RishabhTayal You need to set a response json with "data" key in it. That is what parse_response expects. So changing your line 166 - where you set the stubbing to 

```ruby
 MockAPI::TestFlightServer.get('/testflight/v2/providers/fake-team-id/apps/some-app-id/internalUsers') { '{"data" : ""}' }
```

instead of 

```ruby
 MockAPI::TestFlightServer.get('/testflight/v2/providers/fake-team-id/apps/some-app-id/internalUsers') { }
```
would do the trick.

However, I do see that you are not parsing this data into any object - if you just need the itunesconnect users for a particular team may be you can use the `all` method of `spaceship/lib/spaceship/tunes/members.rb` class?

RishabhTayal:
@heartofagoof Awesome. That does it. Now the test case are passing. 

Ready for review now. ✅

joshdholtz (approved):
Looks good! 👍

PR #12754, opened 2018-06-14
Scan formatter backticks patch

wag-miles:
### Checklist

### Description

Fixed the 'formatter' argument for scan using backticks to quote the value. Couldn't run scan with a custom formatter.

### DIFF
--- a/scan/lib/scan/test_command_generator.rb
+++ b/scan/lib/scan/test_command_generator.rb
@@ -1,4 +1,4 @@
-require_relative "xcpretty_reporter_options_generator"
+require_relative 'xcpretty_reporter_options_generator'
 
 module Scan
   # Responsible for building the fully working xcodebuild command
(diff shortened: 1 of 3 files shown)


wag-miles:
Signed

joshdholtz (changes requested):
Can you remove the changes to ` fastlane/spec/fixtures/xcodeproj/automatic_code_signing.xcodeproj/project.pbxproj`?

This change is a side effect of running `rspec` but a fix for this was just merged in so this won't be an issue in the future 😊

wag-miles:
Done

scan/lib/scan/test_command_generator.rb
```diff
@@ -85,9 +85,9 @@ def pipe
 
       formatter = []
       if Scan.config[:formatter]
-        formatter << "-f `#{Scan.config[:formatter]}`"
+        formatter << "-f '#{Scan.config[:formatter]}'"
       elsif FastlaneCore::Env.truthy?("TRAVIS")
-        formatter << "-f `xcpretty-travis-formatter`"
+        formatter << "-f 'xcpretty-travis-formatter'"
```
joshdholtz: The backticks are actually needed if you are loading a custom formatter as a gem (which is what this is doing so this needs to stay as backticks)

scan/lib/scan/test_command_generator.rb
```diff
@@ -85,9 +85,9 @@ def pipe
 
       formatter = []
       if Scan.config[:formatter]
-        formatter << "-f `#{Scan.config[:formatter]}`"
+        formatter << "-f '#{Scan.config[:formatter]}'"
```
joshdholtz: Single quotes works if `:formatter` is a file path but this also needs to be backticks if `:formatter` is a gem 

I _think_ what we need to do is only use single quotes **if** `:formatter` end in `.rb` and if the file exists 🤔

joshdholtz (changes requested):

wag-miles:
I applied the latest suggestions and created an additional test case

joshdholtz (changes requested):
Can you run rubocop on this so it passes the linter? 😊

wag-miles:
Done. Aggressive formatter on the loose

joshdholtz (approved):
Looks good! Thanks for doing the changes 😊

PR #12781, opened 2018-06-21
Re-add metrics

KrauseFx:
_fastlane_ tracks a few key metrics to understand how developers are using the tool and to help us know what areas need improvement. No personal/sensitive information is ever collected. Metrics that are collected include: 
 
* The number of _fastlane_ runs
* A salted hash of the app identifier or package name, which helps us anonymously identify unique usage of _fastlane_
 
You can easily opt-out of metrics collection by adding `opt_out_usage` at the top of your `Fastfile` or by setting the environment variable `FASTLANE_OPT_OUT_USAGE`.

Related https://github.com/fastlane/docs/pull/627

Remaining tasks:

- [x] One bug has to be fixed that sends analytics events per action @powerivq 
- [x] We should remove the `backfill_p_hashes` method
- [x] We have to fix sending the request for short Fastfiles @krausefx
- [x] Update code to not take an `events` array, but just a single `event`
- [x] Update user agent to include fastlane version number

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -220,11 +220,7 @@ module Fastlane
       actions_path = File.join(File.expand_path("..", path), 'actions')
       Fastlane::Actions.load_external_actions(actions_path) if File.directory?(actions_path)
 
-      action_launched('import')
-
       return_value = parse(File.read(path), path)
-
-      action_completed('import', status: FastlaneCore::ActionCompletionStatus::SUCCESS)
 
       return return_value
     end
@@ -238,8 +234,6 @@ module Fastlane
 
       Actions.execute_action('import_from_git') do
         require 'tmpdir'
-
-        action_launched('import_from_git')
 
         # Checkout the repo
         repo_name = url.split("/").last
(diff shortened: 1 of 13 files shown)


KrauseFx (commented):
Works! Ready from my side 🎉

KrauseFx:
Tests need to be fixed though

KrauseFx:
```

FastlaneCore::AnalyticsSession single action execution action launch analytics session: launch - in fastlane_core.spec.analytics.analytics_session_spec
--
 
```

powerivq:
Fixed all test failure. Now pending approval.

taquitos:
`We should remove the backfill_p_hashes method`
Ahh, actually, sometimes we don't have the app-id until the end of the run, so `backfill_p_hashes` was originally used to backfill our data when we finally got the p-hash. 

So, instead of immediately sending metrics, I think we actually need to hold on to the event, and wait until we have a p-hash. At that time, we can send the event back.

powerivq:
@KrauseFx @taquitos Updated to delay the metrics until we get the AppID.

Tested on an iOS and Android project locally to confirm that it's calculating the right package name/bundle id. Verified the GA tracking ID is up-to-date. PTAL?

powerivq:
@joshdholtz PTAL too!

joshdholtz (approved):
Looks good and works good 💯 Let's get some metrics 📈

PR #12783, opened 2018-06-22
Fix fastlane session caching

luikore:
### Checklist

### Motivation and Context

I set `SPACESHIP_COOKIE_PATH` and still saw fastlane login with password. Session caching didn't work. Resulted in a lot of 503 errors for me after a lot of operations.

After some debugging, found the line `return true if fetch_olympus_session.count > 0` expects the return value of `fetch_olympus_session` to be an array. But the change in https://github.com/fastlane/fastlane/commit/b11eaa929ecd0144066cdd92ebfaa8b19b40d489 modified the returning value type without updating usage site. Then `load_session_from_file` never really works because `fetch_olympus_session.count` always raises a NoMethodError.

### Description

Change the returning type of `fetch_olympus_session` to boolean and use simple truth test instead.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -394,7 +394,7 @@ module Spaceship
           # As this will raise an exception if the old session has expired
           # If the old session is still valid, we don't have to do anything else in this method
           # that's why we return true
-          return true if fetch_olympus_session.count > 0
+          return true if fetch_olympus_session
         rescue
           # If the `fetch_olympus_session` method raises an exception
           # we'll land here, and therefore continue doing a full login process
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/client.rb
```diff
@@ -491,8 +491,13 @@ def fetch_olympus_session
...
 
         provider = body["provider"]
-        self.provider = Spaceship::Provider.new(provider_hash: provider) unless provider.nil?
+        if provider
+          self.provider = Spaceship::Provider.new(provider_hash: provider)
+          return true
+        end
       end
+
+      false
```
KrauseFx: Let's use explicit returns 👍

joshdholtz (approved):
Looks good and works good 👍 🚀 Thanks for fixing this!

PR #12788, opened 2018-06-22
providing the ability to update b2b users

paramv17:
### Checklist

### Motivation and Context

This change will provide an ability to add and remove b2b users. The payload required for the "update" also changed a bit. So I am including it in the new method. 

I tried to remove, add, and add and remove b2b users together. I have also added new rspec tests and updated existing ones that test the payload for correct behaviour.
### Description

1. added method `update_b2b_users` in `availability.rb`.
2. removed the accessor for `b2b_users` from availability - this was a bug which only manifested itself in the new implementation. 
3. updated method `from_username` in `b2b_user.rb` to be able to mark a user to be deleted or added.
4. added `rspec` tests for functionality and updated existing tests to reflect the changes to existing methods - while ensuring that the legacy behaviour works fine for anyone that might be using the method.

### DIFF
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -24,6 +24,7 @@ module Spaceship
       # @return (Bool) b2b available for distribution
       attr_accessor :b2b_unavailable
 
+      # @return (Array of Spaceship::Tunes::B2bUser objects) A list of users
       attr_accessor :b2b_users
 
       attr_mapping(
@@ -98,27 +99,33 @@ module Spaceship
         @b2b_app_enabled = true
         # need to set the educational discount to false
         @educational_discount = false
-        self
+        return self
       end
 
       # Adds users for b2b enabled apps
       def add_b2b_users(user_list = [])
         raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
         @b2b_users = user_list.map { |user| B2bUser.from_username(user) }
-        self
+        return self
       end
 
       # Updates users for b2b enabled apps
       def update_b2b_users(user_list = [])
         raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
+
         added_users = b2b_users.map(&:ds_username)
+
+        # Returns if list is unchanged
         return self if (added_users - user_list) == (user_list - added_users)
+
         users_to_add = user_list.reject { |user| added_users.include?(user) }
         users_to_remove = added_users.reject { |user| user_list.include?(user) }
+
         @b2b_users = b2b_users.reject { |user| users_to_remove.include?(user.ds_username) }
         @b2b_users.concat(users_to_add.map { |user| B2bUser.from_username(user) })
         @b2b_users.concat(users_to_remove.map { |user| B2bUser.from_username(user, is_add_type: false) })
-        self
+
+        return self
       end
     end
   end
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -24,10 +24,6 @@ module Spaceship
       # @return (Bool) b2b available for distribution
       attr_accessor :b2b_unavailable
 
-      # @return (Array of Spaceship::Tunes::B2bUser objects) A list of users set by user - if not
-      # then the b2b user list that is currently set
-      attr_accessor :b2b_users
-
       attr_mapping(
         'theWorld' => :include_future_territories,
         'preOrder.clearedForPreOrder.value' => :cleared_for_preorder,
@@ -78,7 +74,7 @@ module Spaceship
       end
 
       def b2b_users
-        @b2b_users ||= raw_data['b2bUsers'].map { |user| B2bUser.new(user) }
+        @b2b_users ? @b2b_users : raw_data['b2bUsers'].map { |user| B2bUser.new(user) }
       end
 
       def b2b_app_enabled
@@ -109,6 +105,15 @@ module Spaceship
         @b2b_users = user_list.map do |user|
           B2bUser.from_username(user)
         end
+        self
+      end
+
+      # Updates users for b2b enabled apps
+      def update_b2b_users(users_to_add: [], users_to_remove: [])
+        raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
+        @b2b_users = b2b_users.reject { |user| users_to_remove.include?(user.ds_username) }
+        @b2b_users.concat(users_to_add.map { |user| B2bUser.from_username(user) })
+        @b2b_users.concat(users_to_remove.map { |user| B2bUser.from_username(user, is_add_type: false) })
         self
       end
     end
--- a/spaceship/lib/spaceship/tunes/availability.rb
+++ b/spaceship/lib/spaceship/tunes/availability.rb
@@ -24,6 +24,8 @@ module Spaceship
       # @return (Bool) b2b available for distribution
       attr_accessor :b2b_unavailable
 
+      attr_accessor :b2b_users
+
       attr_mapping(
         'theWorld' => :include_future_territories,
         'preOrder.clearedForPreOrder.value' => :cleared_for_preorder,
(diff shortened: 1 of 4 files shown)


spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -78,7 +74,7 @@ def territories
       end
 
       def b2b_users
-        @b2b_users ||= raw_data['b2bUsers'].map { |user| B2bUser.new(user) }
+        @b2b_users ? @b2b_users : raw_data['b2bUsers'].map { |user| B2bUser.new(user) }
```
joshdholtz: Can we change this to `@b2b_users || raw_data['b2bUsers'].map { |user| B2bUser.new(user) }`?

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -24,10 +24,6 @@ class Availability < TunesBase
       # @return (Bool) b2b available for distribution
       attr_accessor :b2b_unavailable
 
-      # @return (Array of Spaceship::Tunes::B2bUser objects) A list of users set by user - if not
-      # then the b2b user list that is currently set
-      attr_accessor :b2b_users
```
joshdholtz: Deleting this `attr_accessor` still feels weird if we are still using `@b2b_users` 🤔 It feels like we should keep this since we are still technically using it but keep the `b2b_users` method changes like you have below

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -111,6 +107,15 @@ def add_b2b_users(user_list = [])
         end
         self
       end
+
+      # Updates users for b2b enabled apps
+      def update_b2b_users(users_to_add: [], users_to_remove: [])
+        raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
+        @b2b_users = b2b_users.reject { |user| users_to_remove.include?(user.ds_username) }
+        @b2b_users.concat(users_to_add.map { |user| B2bUser.from_username(user) })
```
joshdholtz: Do we need to make sure the user isn't already added? Or will we let that responsibility fall onto the user?

spaceship/spec/tunes/b2b_user_spec.rb
```diff
@@ -40,6 +40,20 @@ class B2bUserSpec
         !expect(b2b_user_created.delete)
         expect(b2b_user_created.ds_username).to eq('b2b2@def.com')
       end
+
+      it 'creates correct object to add with explicit input' do
+        b2b_user_created = Spaceship::Tunes::B2bUser.from_username("b2b2@def.com", is_add_type: true)
+        expect(b2b_user_created.add)
+        !expect(b2b_user_created.delete)
```
joshdholtz: Can we change this line and all other expects to have either `.to eq(true)` or `.to eq(false)` to be more explicit and consistent with the rest of the code base? 😊

joshdholtz (changes requested):
Looks good! Just left some nit picks 😊

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: well we define our own `b2b_users` method - i can also define a `b2b_users=` method as well. Let me see if i can keep it around though.

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: I ll try to incorporate that check.

spaceship/spec/tunes/b2b_user_spec.rb
heartofagoof: Ah it was my own rubocop settings - not fastlane's that did this.

heartofagoof:
@joshdholtz I changed the method signature. Now `update_b2b_users` just takes a list of users - and figures out which users to add and remove. see if this looks better? i ll update the documentation after you give all clear.

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -106,9 +104,20 @@ def enable_b2b_app!
       # Adds users for b2b enabled apps
       def add_b2b_users(user_list = [])
         raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
-        @b2b_users = user_list.map do |user|
-          B2bUser.from_username(user)
-        end
+        @b2b_users = user_list.map { |user| B2bUser.from_username(user) }
+        self
```
joshdholtz: Can we change this to `return self`? Explicit is a little easier to read/maintain 😊

spaceship/lib/spaceship/tunes/availability.rb
```diff
@@ -106,9 +104,20 @@ def enable_b2b_app!
...
+      def update_b2b_users(user_list = [])
+        raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled
+        added_users = b2b_users.map(&:ds_username)
+        return self if (added_users - user_list) == (user_list - added_users)
+        users_to_add = user_list.reject { |user| added_users.include?(user) }
+        users_to_remove = added_users.reject { |user| user_list.include?(user) }
+        @b2b_users = b2b_users.reject { |user| users_to_remove.include?(user.ds_username) }
+        @b2b_users.concat(users_to_add.map { |user| B2bUser.from_username(user) })
+        @b2b_users.concat(users_to_remove.map { |user| B2bUser.from_username(user, is_add_type: false) })
         self
```
joshdholtz: Same here. Can we change this to `return self`?

spaceship/lib/spaceship/tunes/b2b_user.rb
```diff
@@ -17,8 +17,13 @@ class B2bUser < TunesBase
...
       )
 
-      def self.from_username(username)
-        self.new({ 'value' => { 'add' => true, 'delete' => false, 'dsUsername' => username } })
+      def self.from_username(username, is_add_type: true)
+        self.new({ 'value' => { 'add' => is_add_type, 'delete' => !is_add_type, 'dsUsername' => username } })
+      end
+
+      # equality check for the two objects (for cleaning up some tests etc)
+      def ==(other)
```
joshdholtz: Is this **only** for tests like the comment? If so, this should probably be removed from here and put into the test file.

If not, does this mean a `josh@example.com` could both be "added" and "deleted" and not be seen as the same user? 🤔

spaceship/lib/spaceship/tunes/availability.rb
joshdholtz: Can we also had some space between these lines with some comments so its a bit easier on the eyes? 😇 Something like...

```rb
      # Updates users for b2b enabled apps
      def update_b2b_users(user_list = [])
        raise "Cannot add b2b users if b2b is not enabled" unless b2b_app_enabled

        # Returns if list is unchanged
        added_users = b2b_users.map(&:ds_username)
        return self if (added_users - user_list) == (user_list - added_users)

        users_to_add = user_list.reject { |user| added_users.include?(user) }
        users_to_remove = added_users.reject { |user| user_list.include?(user) }

        @b2b_users = b2b_users.reject { |user| users_to_remove.include?(user.ds_username) }
        @b2b_users.concat(users_to_add.map { |user| B2bUser.from_username(user) })
        @b2b_users.concat(users_to_remove.map { |user| B2bUser.from_username(user, is_add_type: false) })

        return self
end
```

joshdholtz (changes requested):
Few more comments 😊 But looking closer to 💯 ✊

spaceship/lib/spaceship/tunes/b2b_user.rb
heartofagoof: While this method is used only in test, it is still a method for this class. I can see future uses in the availability class but even if that weren't the case, I'd insist on this method being defined here. It would violate the object oriented design principle to not do it this way - what we can do here in one line ( and one more line in test ) would take nine lines in test. 

Well, yes. Since "add" and "delete" are class properties. And that is what we are checking in equality. So, despite them having the same email_id, a user to be added and a user to be removed are different by design.

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: Sure thing. I will add comments elsewhere as well. : )

spaceship/lib/spaceship/tunes/availability.rb
heartofagoof: Sure thing! : )

joshdholtz (approved):
Thanks for handling all my requests! This looks 💯 💪

PR #12811, opened 2018-06-28
added default value for cocoapods option

RishabhTayal:
Currently there is no default value for `try_repo_update_on_error` option in `cocoapods` action.


joshdholtz (approved):
👍

Issue #12807, opened 2018-06-27
pod_push requires array for 'Sources', but Swift implementation only allows String which throws error


deathByLiche:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

The [pod_push](https://docs.fastlane.tools/actions/pod_push/) documentation requires 'sources' to be an array,  but the Swift implementation only allows for passing a string, which ends up erroring out when being called because ruby is expecting an array. 

This seems to be an issue with pod_lib_lint for Swift as well.

```ruby
# If the podspec has a dependency on another private pod, then you will have to supply the sources you want the podspec to lint with for pod_push to succeed. Read more here - https://github.com/CocoaPods/CocoaPods/issues/2543.
pod_push(path: "TMessages.podspec", repo: "MyRepo", sources: ["https://github.com/username/Specs", "https://github.com/CocoaPods/Specs"])
```

```swift
func podPush(useBundleExec: Bool = false,
             path: String? = nil,
             repo: String? = nil,
             allowWarnings: String? = nil,
             useLibraries: String? = nil,
             sources: String? = nil,
             swiftVersion: String? = nil,
             verbose: Bool = false)
```

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>$ fastlane pushPodspec --verbose
[✔] 🚀 
DEBUG [2018-06-27 15:38:39.54]: Checking if there are any plugins that should be loaded...
INFO [2018-06-27 15:38:40.00]: loading manifest: /Users/jhall/.rvm/gems/ruby-2.5.0/gems/fastlane-2.98.0/fastlane/swift/upgrade_manifest.json
DEBUG [2018-06-27 15:38:40.06]: No missing groups found, so we don't need to worry about adding new groups
DEBUG [2018-06-27 15:38:40.06]: Fastlane.swift FastlaneRunnerAPIVersion (bundled/target): 0.9.21/0.9.21
DEBUG [2018-06-27 15:38:40.06]: DeliverfileProtocol.swift FastlaneRunnerAPIVersion (bundled/target): 0.9.3/0.9.3
[...]

PR #12822, opened 2018-06-29
Bugfix for Swift podPush and podLibLint 'sources' parameter

deathByLiche:
### Checklist

### Motivation and Context

This fixes issue [#12807](https://github.com/fastlane/fastlane/issues/12807), where the swift version of pod_push and pod_lib_lint fail if 'sources' is passed as a parameter. The ruby command is expecting an array, and the swift version was sending a string.

### Description

I changed the parameter type of 'sources' on pod_push and pod_lib_lint to be an array of strings. I also removed a piece of the swift ruby command that was flattening arrays of strings; it looks like the ruby commands are already doing this, so flattening them on the swift side is unnecessary and breaks the ruby commands.


deathByLiche:
I signed it!

joshdholtz:
@deathByLiche Heyyy 👋 Thanks for this contribution! I actually just fixed this in #12823 so I'm going to close this 😊 

The solution was to add a `type: Array` into the actions (which will autogenerate new Swift files with the proper types). Thanks for making this PR and happy fastlaning! 🚀

deathByLiche:
Awesome!

PR #12825, opened 2018-06-30
[snapshot] add result_bundle option (same as gym and scan)

### LINKED ISSUE #12797: Add support to snapshot for result_bundle
When running UI tests for `snapshot` it would be helpful to get the test results for analysis with report generating tools, like https://github.com/TitouanVanBelle/XCTestHTMLReport, to better determine failures.

#### Feature Description
Both `gym` and `scan` already support a `result_bundle` boolean option which saves the test results to the output directory. For snapshots it should behave the same, making sure the test results are put in separate locale directories.

I can use `xcargs` to specify `resultBundlePath` for `snapshot`, but this will only work if I have one device and one locale. If multiple runs will happen as a result of calling `snapshot`, subsequent calls to `xcodebuild` will fail because it expects there to not be a directory at `resultBundlePath`. This is the same issue that happened in #12349.

joshdholtz:
Fixes #12797

## What
- Adds `result_bundle` option that outputs result information in the `output_directory`
  - Same functionality that _gym_ and _scan_ provide


snapshot/lib/snapshot/test_command_generator_base.rb
```diff
@@ -80,6 +81,17 @@ def derived_data_path
         Snapshot.cache[:derived_data_path] ||= (Snapshot.config[:derived_data_path] || Dir.mktmpdir("snapshot_derived"))
       end
 
+      def result_bundle_path
+        unless Snapshot.cache[:result_bundle_path]
+          path = File.join(Snapshot.config[:output_directory], Snapshot.config[:scheme]) + ".test_result"
```
sirnacnud: Can we include the current language/locale in the path? The snapshot test scheme will be executed per language and we would like to save each result_bundle, rather than having them overwritten.

sirnacnud: We should also include the device as well, as multiple devices can be specified to snapshot.

joshdholtz: @sirnacnud I can do the current language/locale part but I don't think I can do the devices part 😬 The new xcodebuild commands take multiple devices and one result bundle path so I'm not able to make a separate path for each device. So I'm assuming that there is only one result bundle path since there is only one xcodebuild. Thoughts?

joshdholtz:
@sirnacnud Updated for the locale but not for the device! (see previous comment https://github.com/fastlane/fastlane/pull/12825#discussion_r200794526)

screen shot 2018-07-06 at 8 28 01 pm

Thoughts? 😊

sirnacnud:
@joshdholtz I think that makes sense, thanks for implementing it!

snapshot/lib/snapshot/test_command_generator_base.rb
```diff
@@ -80,6 +81,20 @@ def derived_data_path
         Snapshot.cache[:derived_data_path] ||= (Snapshot.config[:derived_data_path] || Dir.mktmpdir("snapshot_derived"))
       end
 
+      def result_bundle_path(language, locale)
+        Snapshot.cache[:result_bundle_path] = {}
+        language_key = locale || language
+
+        unless Snapshot.cache[:result_bundle_path][language_key]
```
taquitos: This is never going to be true because on line 85 you always reset it.

snapshot/lib/snapshot/test_command_generator_base.rb
```diff
@@ -17,12 +17,13 @@ def project_path_array
...
       end
 
-      def options
+      def options(language, locale)
         config = Snapshot.config
         options = []
         options += project_path_array
         options << "-sdk '#{config[:sdk]}'" if config[:sdk]
         options << "-derivedDataPath '#{derived_data_path}'"
+        options << "-resultBundlePath '#{result_bundle_path(language, locale)}'" if config[:result_bundle]
```
taquitos: nit: I'm not a fan of method invocation during string interpolation.
it's not a big at all, just a note, and feel free to ignore me if you disagree 👍

taquitos (commented):
One comment to look at and one nit

snapshot/lib/snapshot/test_command_generator_base.rb
joshdholtz: Ehhhh, I also feel the same way 😬 It _was_ already a method so it _looked_ okay but then I added parameters and then if got ugly so I 💯 agree with you. I can rework 😉

snapshot/lib/snapshot/test_command_generator_base.rb
joshdholtz: This could totally be true if someone is running multiple `snapshot`s for the same language during the same fastlane execution 😜

taquitos: 😨

taquitos (approved):
🐐it!

sirnacnud:
@joshdholtz I am testing out this functionality and am noticing when we specify multiple devices to `snapshot`, more than one call to `xcodebuild` per locale is happening, which wipes out the `result_bundle`.

If `concurrent_simulators` is `false`,`xcodebuild` will happen once per device. If `concurrent_simulators` is `true`, it appears to be batching up the devices, but not all together. First call to `xcodebuild` specified `iPhone SE`, `iPhone 8`, and `iPhone 8 Plus`. The second call to `xcodebuild` specified `iPhone X`, `iPad Pro (9.7-inch)`, and `iPad Pro (12.9-inch)`. Is there a limitation preventing us from supplying all devices in one `xcodebuild` call?

Are we able to account for these cases to prevent the `result_bundle` from being wiped out? Sorry I didn't catch it earlier.

sirnacnud:
Also I noticed if a test fails and we have specified `number_of_retries` to run them again, it errors out as the `result_bundle` is already present. Maybe we can account for this as well in the `result_bundle` path, something like appending a number to the `test_result` directory?

PR #11946, opened 2018-02-28
Show failed point of lines in configuration file

matsune:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I've updated the documentation if necessary.

### Motivation and Context

implement #11706 #11705

When Fastlane failed to parse configuration file (e.g. Fastfile, Matchfile) it should output lines around error point on console with error message.

### Description

↓Outputs message as below when caught NameError in Fastfile.

PR #12841, opened 2018-07-03
Remove un-needed puts from configuration_file.rb

joshdholtz:
Fixes https://github.com/fastlane/fastlane/pull/11946/files#r192949219

## What
- `puts` was not actually needed so it has been removed now

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
@@ -43,7 +43,7 @@ module FastlaneCore
 
       begin
         # rubocop:disable Security/Eval
-        puts(eval(content)) # this is okay in this case
+        eval(content) # this is okay in this case
         # rubocop:enable Security/Eval
 
         print_resulting_config_values unless skip_printing_values # only on success


fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
```diff
@@ -43,7 +43,7 @@ def initialize(config, path, block_for_missing, skip_printing_values = false)
 
       begin
         # rubocop:disable Security/Eval
-        puts(eval(content)) # this is okay in this case
```
taquitos: Huh, any idea why it was there in the first place?

taquitos (approved):
🎈:goat:

fastlane_core/lib/fastlane_core/configuration/configuration_file.rb
joshdholtz: Something I missed when reviewing #11946 🙃

revolter: @taquitos It was added by https://github.com/matsune/fastlane/commit/7d12cdabc8850225daa44a0a66477e94d1b0dd78

revolter: I'm guessing it was added for debugging purposes and was left there by mistake.

taquitos: Oh, that makes sense. I suspected that to be the case. Thanks for checking!

PR #12846, opened 2018-07-03
Fix option to disable ssl verification

raymondhoagland:
Fix logic to respect false value for params[:secure]

### Checklist

### Motivation and Context
Currently even if params[:secure] is passed as false it is overwritten to true.  This means ssl verification _cannot_ be disabled.

Using the following simple Ruby snippet
```ruby
x = false || true
```
we can see that x is equal to true at the end.

Similarly if we run
```ruby
y = false
x = y || true
```
we can see that x is equal to true again.

### Description
Only coerce secure to be true if it is not specified at all.  In general since the default is `true` anyway, we can probably get away with not touching it at all.

### DIFF
--- a/fastlane/lib/fastlane/actions/github_api.rb
+++ b/fastlane/lib/fastlane/actions/github_api.rb
@@ -17,7 +17,6 @@ module Fastlane
           payload = construct_body(params[:body], params[:raw_body])
           error_handlers = params[:error_handlers] || {}
           secure = params[:secure]
-          secure = true if secure.nil?
 
           response = call_endpoint(
             url,
@@ -141,7 +140,7 @@ module Fastlane
             FastlaneCore::ConfigItem.new(key: :secure,
                                          env_name: "FL_GITHUB_API_SECURE",
                                          description: "Optionally disable secure requests (ssl_verify_peer)",
-                                         is_string: false,
+                                         type: Boolean,
                                          default_value: true,
                                          optional: true)
           ]
--- a/fastlane/lib/fastlane/actions/github_api.rb
+++ b/fastlane/lib/fastlane/actions/github_api.rb
@@ -16,7 +16,8 @@ module Fastlane
           headers = construct_headers(params[:api_token], params[:headers])
           payload = construct_body(params[:body], params[:raw_body])
           error_handlers = params[:error_handlers] || {}
-          secure = params[:secure] || true
+          secure = params[:secure]
+          secure = true if secure.nil?
 
           response = call_endpoint(
             url,
--- a/fastlane/spec/actions_specs/github_api_spec.rb
+++ b/fastlane/spec/actions_specs/github_api_spec.rb
@@ -262,7 +262,7 @@ describe Fastlane do
               end
             end
           end
-          
+
           context 'secure is set' do
             it 'correctly submits without ssl verification' do
               Excon.defaults[:ssl_verify_peer] = true
--- a/fastlane/spec/actions_specs/github_api_spec.rb
+++ b/fastlane/spec/actions_specs/github_api_spec.rb
@@ -262,6 +262,64 @@ describe Fastlane do
               end
             end
           end
+          
+          context 'secure is set' do
+            it 'correctly submits without ssl verification' do
+              Excon.defaults[:ssl_verify_peer] = true
+              result = Fastlane::FastFile.new.parse("
+                lane :test do
+                  github_api(
+                    api_token: '123456789',
+                    http_method: 'PUT',
+                    path: 'repos/fastlane/fastlane/contents/TEST_FILE.md',
+                    secure: false
+                  )
+                end
+              ").runner.execute(:test)
+
+              expect(Excon.defaults[:ssl_verify_peer]).to eq(false)
+              expect(result[:status]).to eq(200)
+              expect(result[:body]).to eq(response_body)
+              expect(result[:json]).to eq(JSON.parse(response_body))
+            end
+
+            it 'correctly submits with ssl verification' do
+              Excon.defaults[:ssl_verify_peer] = false
+              result = Fastlane::FastFile.new.parse("
+                lane :test do
+                  github_api(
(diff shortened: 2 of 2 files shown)


raymondhoagland:
I signed it!

fastlane/lib/fastlane/actions/github_api.rb
```diff
@@ -16,7 +16,8 @@ def run(params)
           headers = construct_headers(params[:api_token], params[:headers])
           payload = construct_body(params[:body], params[:raw_body])
           error_handlers = params[:error_handlers] || {}
-          secure = params[:secure] || true
+          secure = params[:secure]
+          secure = true if secure.nil?
```
joshdholtz: We can remove this line (the one with the `if secure.nil?`) completely 💣 

Can you also add in `type: Boolean` into the `FastlaneCore::ConfigItem` for `:secure`? Like the following?
```rb
FastlaneCore::ConfigItem.new(key: :secure,
  env_name: "FL_GITHUB_API_SECURE",
  description: "Optionally disable secure requests (ssl_verify_peer)",
  is_string: false,
  type: Boolean,
  default_value: true,
  optional: true)
```

This provides a more secure default value and will make this work with `Fastlane.swift` 😊

joshdholtz (changes requested):

joshdholtz (changes requested):
Can you also add a test around this? You will need to stub a non-SSL request like the following:

https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/github_api_spec.rb#L16-L18

And then add a context/it that sets `secure: false`. After that we should be 💯!

raymondhoagland:
So I'm not sure what (if anything) would be different from webmock's view - the secure parameter gets passed to excon which will disable  ssl verification but doesn't appear to change the request webmock receives.

If there's something specific the stub should be checking for then I'd be happy to add a test.

Some sample code I used to check the requests to webmock
```ruby
require 'webmock'
require 'net/http'
require  'excon'

include WebMock::API
WebMock.enable!

# Print info on the request
WebMock.after_request do |req, response|
    puts req.inspect
    puts response.inspect
end

# Stubs
stub_request(:post, "https://www.example.com").
  to_return(status: 200, body: "received https request", headers: {})

uri = URI.parse('https://www.example.com/')
req = Net::HTTP::Post.new(uri.path)
req.body = {}.to_json
req.content_type = "application/json"

# Net/HTTP with SSL verification
res = Net::HTTP.start(
        uri.host, 
        uri.port,
        :use_ssl => true,
        :verify_mode => OpenSSL::SSL::VERIFY_PEER
) do |http|
    http.request(req)
end

# Net/HTTP w/o SSL verification
res = Net::HTTP.start(
        uri.host, 
        uri.port,
        :use_ssl => true,
        :verify_mode => OpenSSL::SSL::VERIFY_NONE
) do |http|
    http.request(req)
end

# Excon w/ SSL verification
Excon.defaults[:ssl_verify_peer] = false
connection = Excon.new('https://www.example.com')
connection.post(:body => {}.to_json )

# Excon w/o SSL verification
Excon.defaults[:ssl_verify_peer] = true
connection = Excon.new('https://www.example.com')
connection.post(:body => {}.to_json )
```

joshdholtz:
@raymondhoagland You can do something like this 😊 

```rb
        context "secure is set" do                                                                                                                                                                                                                                                       
          it "correctly submits using secure false" do                                                                                                                                                                                                                                   
            expect(Excon.defaults[:ssl_verify_peer]).to eq(true)                                                                                                                                                                                                                         
            result = Fastlane::FastFile.new.parse("                                                                                                                                                                                                                                      
                lane :test do                                                                                                                                                                                                                                                            
                  github_api(                                                                                                                                                                                                                                                            
                    api_token: '123456789',                                                                                                                                                                                                                                              
                    http_method: 'PUT',                                                                                                                                                                                                                                                  
                    path: 'repos/fastlane/fastlane/contents/TEST_FILE.md',                                                                                                                                                                                                               
                    secure: false                                                                                                                                                                                                                                                        
                  )                                                                                                                                                                                                                                                                      
                end                                                                                                                                                                                                                                                                      
              ").runner.execute(:test)                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                         
            expect(Excon.defaults[:ssl_verify_peer]).to eq(false)                                                                                                                                                                                                                        
            expect(result[:status]).to eq(200)                                                                                                                                                                                                                                           
            expect(result[:body]).to eq(response_body)                                                                                                                                                                                                                                   
            expect(result[:json]).to eq(JSON.parse(response_body))                                                                                                                                                                                                                       
          end                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                         
          it "correctly submits using secure default" do                                                                                                                                                                                                                                 
            result = Fastlane::FastFile.new.parse("                                                                                                                                                                                                                                      
                lane :test do                                                                                                                                                                                                                                                            
                  github_api(                                                                                                                                                                                                                                                            
                    api_token: '123456789',                                                                                                                                                                                                                                              
                    http_method: 'PUT',                                                                                                                                                                                                                                                  
                    path: 'repos/fastlane/fastlane/contents/TEST_FILE.md',                                                                                                                                                                                                               
                  )                                                                                                                                                                                                                                                                      
                end                                                                                                                                                                                                                                                                      
              ").runner.execute(:test)                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                         
            expect(Excon.defaults[:ssl_verify_peer]).to eq(true)                                                                                                                                                                                                                         
            expect(result[:status]).to eq(200)                                                                                                                                                                                                                                           
            expect(result[:body]).to eq(response_body)                                                                                                                                                                                                                                   
            expect(result[:json]).to eq(JSON.parse(response_body))                                                                                                                                                                                                                       
          end                                                                                                                                                                                                                                                                            
        end  
```

joshdholtz (approved):
Perfect! Thanks for adding those tests 😊

PR #12850, opened 2018-07-04
[upload_symbols_to_crashlytics] Get rid of the limit of the number of threads

chuganzy:
### Checklist

### Motivation and Context

I was using `sysctl -n hw.ncpu` for this option but recently noticed that I cannot do on iMac Pro as it has at least 16 of them (8 * 2).

### Description

Just increasing this value also works, however as computers are getting more and more powerful, I thought it’s hard to maintain this value. Also I think calculating the limit from the number of cores is too much as the way to get it is different on each platforms, so went with getting rid of the limit.


joshdholtz (approved):
Seems legit 😊 💪

PR #12862, opened 2018-07-06
[spaceship] remove defaulting push notifications on in create_app

### LINKED ISSUE #12853: Why is produce always enabling APN when creating a new App ID?
### Question Checklist

### Question Subject
example command: `fastlane produce -u 'development@incapptic.com' -a 'com.example.fastlane' -q 'Fastlane' --verbose`

image

#### Question Description
I would expect that non of these entitlements is enabled. But in my case especially the APN is problematic.

joshdholtz:
Fixes #12853

## What
- _spaceships_'s `create_app!` method would enable the push notification feature/service for any app is created (along with game center and in-app payments)
- Apple enabled (and doesn't allow disabling) of game center and in-app payments but **does not** enable push notifications by default
- This PR removes push notifications from getting enabled by default
 
## Side Effects?
- One side effect I _thought_ could happen would be that `pem` wouldn't be able to create the push certificate if the push notifications feature/service was disable
- After some testing, it does appear that `pem` can still create a push certificate (and enable the push notification feature/service) if it is disabled so this is not an issue 👍 

## Bonus Fix
- And added a bonus fix for some duplicate looking up of `enable_services` when should have been also looking for deprecated `enabled_features`

### DIFF
--- a/produce/lib/produce/developer_center.rb
+++ b/produce/lib/produce/developer_center.rb
@@ -81,8 +81,8 @@ module Produce
       app_service = Spaceship.app_service
       enabled_clean_options = {}
 
-      # "enable_services" was deprecated in favor of "enable_services"
-      config_enabled_services = Produce.config[:enable_services] || Produce.config[:enable_services]
+      # "enabled_features" was deprecated in favor of "enable_services"
+      config_enabled_services = Produce.config[:enable_services] || Produce.config[:enabled_features]
 
       config_enabled_services.each do |k, v|
         if k.to_sym == :data_protection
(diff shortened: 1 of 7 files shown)


produce/lib/produce/developer_center.rb
```diff
@@ -81,8 +81,8 @@ def enable_services
       app_service = Spaceship.app_service
       enabled_clean_options = {}
 
-      # "enable_services" was deprecated in favor of "enable_services"
-      config_enabled_services = Produce.config[:enable_services] || Produce.config[:enable_services]
+      # "enabled_features" was deprecated in favor of "enable_services"
+      config_enabled_services = Produce.config[:enable_services] || Produce.config[:enabled_features]
```
taquitos: Huh, backwards compatibility fixed!

taquitos (approved):
🆒:goat:

PR #12881, opened 2018-07-11
[snapshot] fix bug getting the device width

asam139:
Fix bug: get first window to get the device width.

### Checklist

### Motivation and Context

In some cases and devices the SnapshotHelper.swift fails finding the window using the current way.

I have tested with all simulators for iOS 10 and iOS 11 (iPad an iPhone)

### Description

Change the way which the device width is obtained

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -277,4 +277,4 @@ private extension CGFloat {
 
 // Please don't remove the lines below
 // They are used to detect outdated configuration files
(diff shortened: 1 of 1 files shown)


asam139:
I signed it!

joshdholtz (changes requested):
This looks good! Can you also bump the last line to `// SnapshotHelperVersion [1.11]`? 😊 Then we should be good to go

asam139:
Ready!! Thanks so much 💪🏽

joshdholtz (approved):
Perfect! Thanks for the contribution ❤️

PR #12906, opened 2018-07-15
Add support for release notes from bundles

tmtrademarked:
### Checklist

### Motivation and Context
For projects that use AAB instead of APK, the release notes functionality of fastlane wasn't working. This is because the `upload_changelogs` functionality specifically looks for APK version codes - which obviously don't include AABs. 

Another approach would be to change the `apk_version_codes` client library call to also include bundles. This felt like it would be more disruptive for other use cases.

### Description
This PR is quite simple - it extends the `client` class to be able to list bundle versions in addition to APK versions. It then uses this in `uploader` to upload changelogs for bundles as well.

This was tested locally by using the "dry run" functionality to verify upload of metadata succeeded on a project with bundles. No errors were reported from Play using the new uploads.

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -156,13 +156,22 @@ module Supply
       end
     end
 
-    # Get a list of all apks verion codes - returns the list of version codes
+    # Get a list of all APK version codes - returns the list of version codes
     def apks_version_codes
       ensure_active_edit!
 
       result = call_google_api { android_publisher.list_apks(current_package_name, current_edit.id) }
 
       return Array(result.apks).map(&:version_code)
+    end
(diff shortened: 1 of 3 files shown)


joshdholtz (changes requested):
This looks great! Thanks for finding this 😊 Can you add `list_edit_bundles` to https://github.com/fastlane/fastlane/blob/master/supply/spec/client_spec.rb? Then we should be good 👍

tmtrademarked:
Thanks for the rapid review, @joshdholtz! Added the method to the `client_spec` test - good catch!

joshdholtz (approved):
This is great! Thanks again 😊

PR #12912, opened 2018-07-16
Add new option to "jira" action

Ikloo:
### Checklist

### Motivation and Context
After integrating fastlane into my project, I will need to use action `fastlane jira` to comment ticket at JIRA. But I have self-hosted JIRA server and that is a problem for me to use action with current options.
It resolves a problem for self-hosted JIRA servers via adding `"/jira"` into `context_path`.

### Description

Added `self_hosted_server` option to action. Option is optional, default - `false`
It appends `"/jira"` to context_path variable for self-hosted jira servers and corrects using in 'jira-ruby' gem.
Updated example with the new option.

### DIFF
--- a/fastlane/lib/fastlane/actions/jira.rb
+++ b/fastlane/lib/fastlane/actions/jira.rb
@@ -43,11 +43,11 @@ module Fastlane
                                        verify_block: proc do |value|
                                          UI.user_error!("No url for Jira given, pass using `url: 'url'`") if value.to_s.length == 0
                                        end),
-          FastlaneCore::ConfigItem.new(key: :self_hosted_server,
+          FastlaneCore::ConfigItem.new(key: :context_path,
                                       env_name: "FL_JIRA_CONTEXT_PATH",
-                                      description: "Is JIRA self hosted server and need custom context path",
+                                      description: "Appends to the url (ex: \"/jira\")",
                                       optional: true,
-                                      default_value: "",
+                                      default_value: ""),
           FastlaneCore::ConfigItem.new(key: :username,
                                        env_name: "FL_JIRA_USERNAME",
                                        description: "Username for JIRA instance",
@@ -98,7 +98,7 @@ module Fastlane
           )',
           'jira(
             url: "https://yourserverdomain.com",
-            context_path: "/jira", # append \'/jira\' path to your url
+            context_path: "/jira",
             username: "Your username",
             password: "Your password",
             ticket_id: "Ticket ID, i.e. IOS-123",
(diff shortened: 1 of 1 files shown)


joshdholtz (commented):
@Ikloo Thanks for adding this! Quick question... Would it be more flexible for other JIRA installs to add a `context_path` option instead of `self_hosted_server` option?

I don't know much about self hosted JIRA instances but I'm assuming that there is a chance where the `context _path` might not be specifically `/jira`. It seems like adding an option for `context_path` might be more powerful? Just some thoughts 😊

Ikloo:
@joshdholtz Initially, I did, but then I thought that I did not see other possible `context_path`. Maybe it's because of my little experience.🤔
I will update the PR with this idea. 🚀

joshdholtz:
@Ikloo Thank you! I'm not sure **if** there will be another context path but that _kinda_ seems like what that `context_path` variable was meant for but it was just never finished 🙃

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -43,6 +43,11 @@ def self.available_options
                                        verify_block: proc do |value|
                                          UI.user_error!("No url for Jira given, pass using `url: 'url'`") if value.to_s.length == 0
                                        end),
+          FastlaneCore::ConfigItem.new(key: :self_hosted_server,
```
joshdholtz: Can you change this to `:context_path`?

Ikloo:
@joshdholtz 
I agree with you.
Updated PR 👌

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -43,6 +43,11 @@ def self.available_options
...
+                                      env_name: "FL_JIRA_CONTEXT_PATH",
+                                      description: "Is JIRA self hosted server and need custom context path",
```
joshdholtz: Can you change this to `Appends to the url (ex: "/jira")`

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -90,6 +95,14 @@ def self.example_code
             password: "Your password",
             ticket_id: "Ticket ID, i.e. IOS-123",
             comment_text: "Text to post as a comment"
+          )',
+          'jira(
+            url: "https://yourserverdomain.com",
+            context_path: "/jira", # append \'/jira\' path to your url
```
joshdholtz: Can you remove the `# append \'/jira\' path to your url`? This information should be in the option description 😊

joshdholtz (changes requested):

Ikloo:
@joshdholtz check it, please. All okay!

joshdholtz (approved):
Looks great! Thanks for adding this 😊 💪 🚀

Ikloo:
@joshdholtz Great! Thanks!

Ikloo:
Awesome 🎊  🎉 🚀

PR #12922, opened 2018-07-18
[test] Fix failing tests on master

chuganzy:
As `dsym_path` has the default value which is dynamic, need to stub `Dir` to purely test `dsym_paths` parameter.

https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb#L150

Also updated to clear env and lane_context just in case.


joshdholtz (approved):
I have a solution for this is another PR but I like your solution better 😉

Issue #11585, opened 2018-01-16
frameit not shrinking text for multiline titles


JosephDuffy:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

After an update to my app is re-ran my `create_screenshots` lane, which generated my screenshots and then ran `frameit`. However, the newly generated image do not seem to be reducing the size of the text of the title, causing the text to be clipped. I see this across multiple iOS devices and computers performing `frameit`. Below is an example from a 9.7" iPad, followed by the original screenshot.

![ipad pro 9 7-inch -3actionextension_framed](https://user-images.githubusercontent.com/1325813/35013018-c5e87aa8-fb03-11e7-89d4-279663caf50e.png)

![ipad pro 9 7-inch -3actionextension](https://user-images.githubusercontent.com/1325813/35013000-b7fb9e5c-fb03-11e7-9565-8b1b3fb81cc6.png)

##### Complete output when running fastlane, including the stack trace and command used

```
$ fastlane add_device_frames
[18:23:56]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[18:23:56]: ----------------------------------------
[18:23:56]: --- Step: Verifying fastlane version ---
[18:23:56]: ----------------------------------------
[18:23:56]: Your fastlane version 2.76.0 matches the minimum requirement of 1.48.0  ✅
[18:23:56]: ------------------------------
[18:23:56]: --- Step: default_platform ---
[18:23:56]: ------------------------------
[18:23:56]: Driving the lane 'ios add_device_frames' 🚀
[18:23:56]: ---------------------
[18:23:56]: --- Step: frameit ---
[18:23:56]: ---------------------
[18:23:56]: Framing screenshots at path fastlane/screenshots/ (via frameit)
[18:23:56]: Using device frames version 'latest'
[18:23:56]: Framing screenshot './default/iPad Pro (10.5-inch)-3ActionExtension.png'
[18:23:56]: Unknown device type for size iOS-iPad-10.5 for path './default/iPad Pro (10.5-inch)-3ActionExtension.png'
[...]

PR #11014, opened 2017-11-23
[FrameIt] Multiple layout improvements (including iPhone X, multiline labels)

bvirlet:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

_frameit_ has some limitations to its layout engine. With new devices such as the iPhone X, this sometimes results in awkward results. This PR addresses a couple of these layout issues and adds options to customize the layout. This are all useful changes that I applied for the latest version of our app Genius Scan.

### Description

This addresses multiple layout issues:

#### Multi-lines labels were improperly laid out

A multiline label is a string with a `\n`.

beforeafter

Note that the layout isn't particularly "beautiful" but I have set the padding to 0 to be clear about the consequences. With a large padding, the problem was less visible.

This is fixed by not relying on the font size anymore (it would always return the height of 1 line) but on the actual title height which returns the height for the multiple lines.

#### Add title_min_height param to avoid iPhone X layout issues

This enables to have the device tops being all at the same height even if the title/keyword are taller on a given screenshot. This can happen with multiline titles but also titles were the size was adjusted (see example). **Having devices at different heights is not super nice when the user scrolls horizontally between the screenshots on the App Store.**

Additionally, it's another way to add some breathing room to the screenshots, independently from the `padding` option.

_Before_

_After_, with `"title_min_height" : "20%"`

#### Add cascading filters

You may want to adjust the layout not only based on the screenshot number/name, but also on the device. Currently, only the first filter matching the screenshot path is applied. With this change, the filters are applied in order.

This lets modify the iPad screenshot without modifying the iPhone's:

```
{
  "device_frame_version": "latest",
  "default": {
    "title": {
      "color": "#ffffff"
    },
    "background": "./Frameit-background.png",
    "padding": "3%",
    "show_complete_frame": false,
    "stack_title" : true,
    "title_below_image": false,
    "font_scale_factor": 0.05
  },
   "data": [
    {
      "filter": "scan"
    },
    {
      "filter": "edit"
    },
    {
      "filter": "multipage_pdf"
    },
    {
      "filter": "organize"
    },
    {
      "filter": "export"
    },
    {
      "filter" : "iPad",
[...]

PR #12484, opened 2018-05-09
[frameit] Multiple layout improvements

futuretap:
Rebased [PR 11014 (Multiple layout improvements (including iPhone X, multiline labels))](https://github.com/fastlane/fastlane/pull/11014) to master.

PR #12928, opened 2018-07-19
[frameit] Support multiple cascading filters

bvirlet:
🔑 

### Checklist

### Motivation and Context

My existing #11014 PR didn't get merged – admittedly, it wasn't easy to review – and got stale. However, it mixed different changes and some of them have been already addressed now (I believe) by #11585.

This part hasn't been addressed though, so I would like to resubmit if for review:

#### Add cascading filters

You may want to adjust the layout not only based on the screenshot number/name, but also on the device. Currently, only the first filter matching the screenshot path is applied. With this change, the filters are applied in order.

This lets modify the iPad screenshot without modifying the iPhone's:

```
{
  "device_frame_version": "latest",
  "default": {
    "title": {
      "color": "#ffffff"
    },
    "background": "./Frameit-background.png",
    "padding": "3%",
    "show_complete_frame": false,
    "stack_title" : true,
    "title_below_image": false,
    "font_scale_factor": 0.05
  },
   "data": [
    {
      "filter": "scan"
    },
    {
      "filter": "edit"
    },
    {
      "filter": "multipage_pdf"
    },
    {
      "filter": "organize"
    },
    {
      "filter": "export"
    },
    {
      "filter" : "iPad",
      "show_complete_frame" : true,
      "font_scale_factor": 0.035
    }
  ]
}
```

### Documentation

Regarding the documentation update, here is the proposed change:

Before:
```
This is mandatory to link the individual configuration to the screenshot, based on part of the file name.
Example:
If a screenshot is named iPhone 8-Brainstorming.png you can use value Brainstorming for filter. All other keys from that array element will only be applied on this specific screenshot.
```
After:
```
This is mandatory to link the individual configuration to the screenshot, based on part of the file name.
Example:
If a screenshot is named `iPhone5_Brainstorming.png` the first entry in the `data` array will be used. If there are more than one `filter` matching an entry, they will all be applied in order (which means that the last one has the highest precedence).
```
If I understand correctly, I will need to make the corresponding PR to the fastlane/docs repo?

### DIFF
--- a/frameit/lib/frameit/config_parser.rb
+++ b/frameit/lib/frameit/config_parser.rb
@@ -26,11 +26,14 @@ module Frameit
     # Fetches the finished configuration for a given path. This will try to look for a specific value
     # and fallback to a default value if nothing was found
     def fetch_value(path)
-      specific = @data['data'].find { |a| path.include?(a['filter']) }
+      specifics = @data['data'].select { |a| path.include?(a['filter']) }
 
       default = @data['default']
 
-      values = default.fastlane_deep_merge(specific || {})
+      values = default.clone
+      specifics.each do |specific|
+        values = values.fastlane_deep_merge(specific)
+      end
 
       change_paths_to_absolutes!(values)
       validate_values(values)


joshdholtz:
@bvirlet Heyyy! TBH, I don't have frameit setup to test this PR but I will get something setup to test this 💪

bvirlet:
Hi @joshdholtz ? Do you still you will find some time to review this? I think it can be really useful to others but I'm afraid this might go stale again 😄  Thanks for all what you are doing for fastlane in any case!

janpio:
Hey @bvirlet, could you maybe explain the code change in plain English? Right now I don't really get it :/

> If I understand correctly, I will need to make the corresponding PR to the fastlane/docs repo?

Fortunately not, you can just edit this file in this repository: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/frame_screenshots.md

bvirlet:
hi @janpio. Thanks for taking the time to check this.

Line 29: instead of just finding the first `data` object of the JSON matching the filter, we find _all_ of them.
Line 33-36: before, we were merging the selected `data` object of the JSON into the default data object. Now we merge all the selected ones. We have to clone the `default` object because otherwise we would modify the object we are iterating on (that wasn't a problem before with just a single object).
A good way to see what happens would be to just print the result before and after line 36.

janpio:
Ah ok, I wasn't aware that it was just searching for any match for the `filter` string - I thought it had to be an exact match of the non-device part. This doesn't seem to be the case though.

---

BUT: It seems this PR is already merged via the rebased https://github.com/fastlane/fastlane/pull/12484! Correct @bvirlet? https://docs.fastlane.tools/actions/frameit/#specific-parameters also already has your wording from your original description here, correct?

That being said: We should really improve the documentation and make sure this is all written down somewhere...

janpio (changes requested):
Don't merge: Already part of the code base via #12484

bvirlet:
Ah yes, looks like @futuretap rebased it and got it merged in the meantime!

janpio:
Feel free to create new PRs, next time I will make sure they get merged a "bit" faster 🎂

PR #12952, opened 2018-07-24
[WIP] [snapshot][screengrab] offer option for snapshot reports for generating rows by screenshot

### LINKED ISSUE #12949: [snapshot] Add parameter to generate Snapshot reports with columns and rows reversed
The reason behind this is that during review, most displays are widescreen, and it really helps to have all of the first snapshot in all languages side by side to swipe left and right. Then it’s possible to move down to the next row for the second snapshot, etc. A big display is used for reviews also so it’s mounted in widescreen orientation.

The gain is this would allow more of the same snapshot on the display at once, organized together. This is standard for the review process to do one screen at at time in a group setting.

#### Feature Description
There is a `reports_generator.rb` that is run at the end of snapshot tests, and can also be called on its own. This by default will generate rows, one for each language, and columns, one for each snapshot.

The desire is to provide an argument or capability to reverse this. Each column would be each language, and, each row would be the same snapshot. Row 1 would be all the snapshot 1's, row 2 would be all the snapshot 2's, etc.

The default would of course still be left to the current report generation paradigm so as to not break existing usage.

None that I am aware of other than to hard code an HTML or write a separate script.

joshdholtz:
Fixes #12949

## Tasks
- [x] Add `html_summary_format` to _snapshot_
- [ ] Add `html_summary_format` to _screengrab_
- [ ] Clean up code for _snapshot_
- [ ] Write tests

## Description
- Adds new `html_summary_format` option to `snapshot`
  - Ex: `html_summary_format: "group_by_device"` (this is what original implementation was)
  - Ex: `html_summary_format: "group_by_screenshot"` (this is new implementation)
- `group_by_device`
  - Original implementation
  - Groups rows for each language and columns for each screenshot
- `group_by_screenshot`
  - Groups rows by screenshot and each column by language

screen shot 2018-07-23 at 10 13 22 pm

### DIFF
--- a/snapshot/lib/snapshot/options.rb
+++ b/snapshot/lib/snapshot/options.rb
@@ -86,6 +86,14 @@ module Snapshot
                                      description: "By default, the latest version should be used automatically. If you want to change it, do it here",
                                      short_option: "-i",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :html_summary_format,
+                                     env_name: 'SNAPSHOT_HTML_SUMMARY_FORMAT',
+                                     description: "The format of how the HTML summary should be generated",
+                                     default_value: "group_by_device",
+                                     verify_block: proc do |value|
+                                       valid_values = %w(group_by_device group_by_screenshot)
+                                       UI.user_error!("Unsupported export_method '#{value}', must be: #{valid_values}") unless valid_values.include?(value)
+                                     end),
         FastlaneCore::ConfigItem.new(key: :skip_open_summary,
                                      env_name: 'SNAPSHOT_SKIP_OPEN_SUMMARY',
                                      description: "Don't open the HTML summary after running _snapshot_",
--- a/snapshot/lib/snapshot/page.html.erb
+++ b/snapshot/lib/snapshot/page.html.erb
@@ -11,11 +11,14 @@
       .language {
 
       }
-      .deviceName {
+      .deviceName, .languageName, .screenshotTitle  {
         display: block;
         font-size: 30px;
         padding-bottom: 24px;
         padding-top: 45px;
+      }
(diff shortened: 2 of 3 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

BJap (approved):
This is working well for me and does exactly what I needed for my team.

Source: https://github.com/fastlane/fastlane/issues/12949

BJap:
As a general test step I also ran the report builder after taking one screenshot out of one folder and confirmed it still works, and that language is just skipped for that row. (maybe some languages have a screen some don't so the screenshot might be skipped in some edge cases.

I may suggest testing it with something that makes multiple screenshots in a test run to make sure things are in the correct order but it looks great! I am hoping it grabs all screens by similar names.

joshdholtz:
@BJap Thanks for doing some testing on it and reporting back 😊 I will definitely need to update my _snapshot_ setup so its easier to test/debug this 😉 Flipping this table around wasn't as trivial as I thought due to the way the files are output but I can definitely find a way to get this working 💯. I will try to get this finished up this week for ya

BJap:
Agreed and thanks so much for doing this. I thought it through abstractly and thought it could be more difficult to rotate it, and started making a mental note of tests to run to make sure. I guess that served well to note what could go wrong.

janpio (changes requested):
In the meantime unfortunately another PR was merged, that added a switcher between grouping by language and device - this will make merging this PR almost impossible. It will probably be easier to recreate it from scratch based on current `master`.

PR #12964, opened 2018-07-25
Fix/window is not found

asam139:
Fix bug: get first window to get screenshot

### Checklist

### Motivation and Context

In some cases and devices the SnapshotHelper.swift fails finding the window using the current way.

I have tested with all simulators for iOS 10 and iOS 11 (iPad an iPhone)

### Description

Change the way which the window is obtained.

I saw an old pull request (merged) which this line was modified. I have a couple of apps where the current code does not work. Getting the first window works really fine.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -273,4 +273,4 @@ private extension CGFloat {
 
 // Please don't remove the lines below
 // They are used to detect outdated configuration files
-// SnapshotHelperVersion [1.11]
+// SnapshotHelperVersion [1.12]
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -160,7 +160,7 @@ open class Snapshot: NSObject {
                 return
             }
             
-            let window = window.firstMatch
+            let window = app.windows.firstMatch
             let screenshot = window.screenshot()
             guard let simulator = ProcessInfo().environment["SIMULATOR_DEVICE_NAME"], let screenshotsDir = screenshotsDirectory else { return }
             let path = screenshotsDir.appendingPathComponent("\(simulator)-\(name).png")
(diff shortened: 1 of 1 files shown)


joshdholtz:
I'm down for this change. This actually is how Apple recommends on it - https://developer.apple.com/documentation/xctest/activities_and_attachments/adding_attachments_to_tests_and_activities

screen shot 2018-07-28 at 1 02 10 pm

asam139:
Yes it is. I saw it in the documentation, so how you prefer it. :). 

Using "firstMatch" works really fine for me, the other way alway fails finding the window.

asam139:
I have updated the version to 1.12 in case that pull request is merged 👍 
If not I keep my fork jajaja

Thanks so much.

joshdholtz:
@asam139 I was just gunna ask if you could bump that version 😉 Thanks!

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -160,11 +160,7 @@ open class Snapshot: NSObject {
                 return
             }
             
-            guard let window = app.windows.allElementsBoundByIndex.first(where: { $0.frame.isEmpty == false }) else {
-                print("Couldn't find an element window in XCUIApplication with a non-empty frame.")
-                return
-            }
-
+            let window = window.firstMatch
```
joshdholtz: I _think_ this line should really be `let window = app.windows.firstMatch`?

joshdholtz (changes requested):

snapshot/lib/assets/SnapshotHelper.swift
asam139: Yes it is, sorry, copy and paste is so dangerous jejeje. I have it fixed.

joshdholtz (approved):
Works for me 💪

owurman:
Unfortunately, this actually **reverted** the fix for #12174, so that's now failing in some cases again.

Can we restore that fix, and instead of returning an error, fall back on firstMatch?

PR #12967, opened 2018-07-26
[spaceship] Parametrize view_by in analytics API

pauchan:
### Checklist

### Motivation and Context
When calling iTunes analytics IPA I had the need to get the analytics split by country/source so that 
I added an additional optional parameter that can be passed via command line.

### Description
This PR adds optional `view_by` parameter to `*_interval` methods that allow retrieving iTunes statistics split by country/source.
It also adds `measure_interval(start_t, end_t, measure, filter = nil)` method that will allow passing the measure as a command line parameter.

I don't have much experience in writing Ruby so please be easy on me 🙇

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_analytics.rb
+++ b/spaceship/lib/spaceship/tunes/app_analytics.rb
@@ -72,44 +72,48 @@ module Spaceship
         app_crashes_interval(start_t, end_t)
       end
 
-      def app_units_interval(start_t, end_t)
-        client.time_series_analytics([apple_id], ['units'], start_t, end_t, "DAY")
+      def dimension_interval(start_t, end_t, dimension, filter = nil)
+        client.time_series_analytics([apple_id], [dimension], start_t, end_t, "DAY", filter)
       end
 
-      def app_views_interval(start_t, end_t)
-        client.time_series_analytics([apple_id], ['pageViewCount'], start_t, end_t, "DAY")
+      def app_units_interval(start_t, end_t, filter = nil)
+        client.time_series_analytics([apple_id], ['units'], start_t, end_t, "DAY", filter)
       end
 
-      def app_in_app_purchases_interval(start_t, end_t)
-        client.time_series_analytics([apple_id], ['iap'], start_t, end_t, "DAY")
+      def app_views_interval(start_t, end_t, filter = nil)
+        client.time_series_analytics([apple_id], ['pageViewCount'], start_t, end_t, "DAY", filter)
       end
 
-      def app_sales_interval(start_t, end_t)
-        client.time_series_analytics([apple_id], ['sales'], start_t, end_t, "DAY")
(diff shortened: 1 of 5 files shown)


spaceship/lib/spaceship/tunes/app_analytics.rb
```diff
@@ -72,44 +72,48 @@ def app_crashes
         app_crashes_interval(start_t, end_t)
       end
 
-      def app_units_interval(start_t, end_t)
-        client.time_series_analytics([apple_id], ['units'], start_t, end_t, "DAY")
+      def measure_interval(start_t, end_t, measure, view_by = nil)
```
joshdholtz: Can we rename this one to be `app_measure_interval` to be like the rest? 😊

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -1409,6 +1409,20 @@ def sso_token_for_video
       @sso_token_for_video ||= ref_data.sso_token_for_video
     end
 
+    # generates group hash used in the analytics time_series API
+    def group_for_view_by(view_by, measures)
+      if view_by.nil? || measures.nil?
+        return nil
+      else
+        return {
```
joshdholtz: Can we add a comment here something like `# Using rank=DESCENDING and limit=3 as this is what the App Store Connect analytics dashboard uses`?

joshdholtz (changes requested):
This looks great! Just two small comments 😊 

Also, can you add some new tests to `app_analytics_spec.rb` that uses these new parameters? I can help you out with these if you need. Just lemme know 🙃

pauchan:
@joshdholtz thank you very much for the review!
I've applied the suggested changes.
Let me know if this is good enough for you, especially around specs 🙇

spaceship/spec/tunes/app_analytics_spec.rb
```diff
@@ -74,6 +74,25 @@
...
+    it "grabs live analytics split by view_by" do
+      start_time, end_time = app.analytics.time_last_7_days
+      TunesStubbing.itc_stub_analytics(start_time, end_time)
+      analytics = app.analytics
+      measure_installs_by_source = analytics.app_measure_interval(start_time, end_time, 'installs', 'source')
+      expect(measure_installs_by_source['size']).to eq(5)
+      val = measure_installs_by_source['results'].find do |a|
+        a['adamId'].include?('898536088')
+      end
+      expect(val['meetsThreshold']).to eq(true)
```
pauchan: I've added only a single test case and a single stub for passing `view_by` parameter as they all would be pretty similar. To make it more understandable/distinguishable, I've extracted it to a separate test case.

Let me know if this is enough or do you want me to make stubs using `view_by` for each measure.

joshdholtz (approved):
This looks great! Thanks for the contribution 💪 🚀

PR #12969, opened 2018-07-26
Add ability to specify Carthage execuatable

brentleyjones:
### Checklist

### Motivation and Context

The `swiftlint` action allows setting an executable. We've found this useful because we use [Mint](https://github.com/yonaskolb/Mint) to manage our Swift command line tools. We want to use Mint for Carthage as well, so we needed an `:executable` config item for it as well.

### Description

Added the `:executable` config item to the `carthage` action to allow providing an override for the executable path used to run Carthage.

### DIFF
--- a/fastlane/lib/fastlane/actions/carthage.rb
+++ b/fastlane/lib/fastlane/actions/carthage.rb
@@ -5,7 +5,7 @@ module Fastlane
       def self.run(params)
         validate(params)
 
-        cmd = ["carthage"]
+        cmd = [params[:executable]]
         command_name = params[:command]
         cmd << command_name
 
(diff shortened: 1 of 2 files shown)


brentleyjones:
I signed it!

fastlane/lib/fastlane/actions/carthage.rb
```diff
@@ -167,6 +167,10 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :log_path,
                                        env_name: "FL_CARTHAGE_LOG_PATH",
                                        description: "Path to the xcode build output",
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :executable,
+                                       env_name: "FL_CARTHAGE_EXECUTABLE",
+                                       description: "Path to the `carthage` executable on your machine",
                                        optional: true)
```
joshdholtz: Instead of making this optional can you try making this have `default_value: "carthage"`? This should also clean up the code on line 8 🙃

joshdholtz (changes requested):
Small suggestion 😊

fastlane/lib/fastlane/actions/carthage.rb
brentleyjones: Addressed your feedback @joshdholtz.

joshdholtz (changes requested):
Thanks for making that code change! One more request would be to add a new test to `carthage_spec.rb` and passing in a different `executable` option and validating the command that is getting shelled out. After that we should be :100: to merge 👌

brentleyjones:
@joshdholtz Test added.

joshdholtz (approved):
Thanks for adding that test! This is 💯

PR #12981, opened 2018-07-30
[supply] added new timeout option

### LINKED ISSUE #12943: Being able to set send timeout for supply
🔑

Supply fails because of the 300s send timeout here [https://github.com/fastlane/fastlane/blob/56e8020a942b86999df82e0d681502f13dec70d3/supply/lib/supply/client.rb#L73](https://github.com/fastlane/fastlane/blob/56e8020a942b86999df82e0d681502f13dec70d3/supply/lib/supply/client.rb#L73)

> /usr/lib/ruby/2.3.0/openssl/buffering.rb:322:in `syswrite': [!] execution expired (Google::Apis::TransmissionError)

My upload speed is bad, I need about 900 seconds to upload my 65Mb APK.

#### Feature Description

It would be great if we could set this timeout in the Appfile, as an environment variable, or as a parameter of the supply action. Don't know if being able to set open and read timeouts would be useful too.
I guess it would look like this:
`google_timeout_sec(300, 300, 1000)` in the Appfile
or `supply(apk: "....", send_timeout_sec: 1000)` in a Fastfile

When I change the value directly in the installed fastlane gem (/var/lib/gems/2.3.0/gems/fastlane-2.100.1/supply/lib/supply/client.rb), the upload works. But of course it's not great, as updating fastlane would reset this change. Forking is a better option but my Ruby knowledge is still limited and I'm not sure if it should be implemented as a parameter of supply or not.

joshdholtz:
Fixes #12943

## Problem
The `300` default for `read_timeout_sec`, `open_timeout_sec`, and `send_timeout_sec` wasn't long enough for all internet connections and APK sizes

## Solution
Added new `timeout` option that defaults to `300` but can be set to whatever 😊

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -68,9 +68,9 @@ module Supply
 
       Google::Apis::ClientOptions.default.application_name = "fastlane - supply"
       Google::Apis::ClientOptions.default.application_version = Fastlane::VERSION
-      Google::Apis::ClientOptions.default.read_timeout_sec = 300
-      Google::Apis::ClientOptions.default.open_timeout_sec = 300
-      Google::Apis::ClientOptions.default.send_timeout_sec = 300
+      Google::Apis::ClientOptions.default.read_timeout_sec = Supply.config[:timeout]
+      Google::Apis::ClientOptions.default.open_timeout_sec = Supply.config[:timeout]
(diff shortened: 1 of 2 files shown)


thii (approved):

KrauseFx (approved):
Beautiful

PR #12985, opened 2018-07-30
[Improvement] Filter out non-xcodeproj files from the list returned by `project_paths`.

justMaku:
### Checklist

### Motivation and Context
Xcode workspace files can contain any type of file, not only `xcodeproj` ones. Prior to this PR Fastlane was iterating over a list of all `FileRef` entries found in the workspace and only filtered entries that ended with the `Pods\Pods.xcodeproj`.

In case a workspace contained non-xcodeproj files Fastlane would still try to open them as they were a valid `xcodeproj` file and fail with a message:

```
[Xcodeproj] Unable to open `/some/path/to/a/project/file` because it doesn't exist.
```

### Description
This PR improves the filtering mechanism in a way that only files ending with `.xcodeproj` will be returned from the `project_paths` method. 

I also added the above mentioned edge cases (loose files and `Pods.xcodeproj`) to the workspace fixture being used to test this method.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -456,6 +456,9 @@ module FastlaneCore
         @_project_paths = workspace_data.scan(/\"group:(.*)\"/).collect do |current_match|
           # It's a relative path from the workspace file
           File.join(File.expand_path("..", path), current_match.first)
+        end.select do |current_match|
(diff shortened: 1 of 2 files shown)


justMaku:
I signed it!

joshdholtz (approved):
This is 💯 Thanks for the contribution! ❤️

PR #12998, opened 2018-07-31
Return empty array when aab not set

jrnk:
### Checklist

### Motivation and Context
I was getting an error upon submitting my builds to google play seems like the `concat` call can be made on a nil return. which results in `lib/supply/uploader.rb:28:in `concat': [!] no implicit conversion of nil into Array (TypeError)` :)
I reran the specs without a problem.

### Description
I've made sure the `upload_bundles` method returns an array when `aab` is not set so the `concat` call on the return value works.

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -153,7 +153,7 @@ module Supply
 
     def upload_bundles
       aab_path = Supply.config[:aab]
-      return unless aab_path
+      return [] unless aab_path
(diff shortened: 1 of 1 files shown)


comm1x:
It helps, thanks!

joshdholtz (approved):
Thanks for making this fix!

RishabhTayal:
I had this same issue yesterday. Thanks for the fix. 👍

PR #13000, opened 2018-07-31
Fix itms tests on Windows

janpio:
### Checklist

### Motivation and Context
ITMS tests currently get an empty string as path for Windows so tests can run instead of looking for a non-existing ITMS. But those empty strings as path cause crashes:

```
  1) FastlaneCore FastlaneCore::ItunesTransporter with Xcode 7.x installed by default upload command generation generates a call to java directly
     Failure/Error: expect(transporter.upload('my.app.id', '/tmp')).to eq(java_upload_command)

     Errno::EINVAL:
       Invalid argument @ dir_chdir -
     # ./fastlane_core/lib/fastlane_core/itunes_transporter.rb:286:in `execute'
     # ./fastlane_core/lib/fastlane_core/itunes_transporter.rb:379:in `upload'
     # ./fastlane_core/spec/itunes_transporter_spec.rb:83:in `block (6 levels) in <top (required)>'
```

### Description
Using './' doesn't.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -199,7 +199,7 @@ module FastlaneCore
     # @return the full path to the iTMSTransporter executable
     def self.itms_path
       return ENV["FASTLANE_ITUNES_TRANSPORTER_PATH"] if FastlaneCore::Env.truthy?("FASTLANE_ITUNES_TRANSPORTER_PATH")
-      return '' unless self.mac? # so tests work on Linux and Windows too
+      return './' unless self.mac? # so tests work on Linux and Windows too
 
       [
         "../Applications/Application Loader.app/Contents/MacOS/itms",


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

joshdholtz (changes requested):
Even though this may fix the tests on Linux/Windows, I don't feel like this solution fits the name/function for this method 🤔 

IMO, this should return a `nil` for non-mac since it can't find the path. Returning a path to current directory feels more like a hack than a solution to me 🙃 It seems like this should return `nil` and then that `nil` be handled elsewhere? Open for discussion though

janpio:
Definitely, that would be the better code. 
But better code is the adversary of working code _right now_ ;)

(Will look into where this is used later)

janpio:
Replaced by https://github.com/fastlane/fastlane/pull/13148

PR #13001, opened 2018-07-31
allow fastlane to fetch list of available bundle ids

paramv17:
### Checklist

### Motivation and Context

A way to query the list of bundle ids available in AppStoreConnect to create a new app. 

This solves a problem where creating the appId in developer portal does not propagate 

To test : 

run `bundle exec irb` using the fastlane repo with these changes
```
> Spaceship::Tunes.login(uname,pword)
> Spaceship::Tunes.select_team
> Spaceship::Tunes::Application.available_bundle_ids
```
The last command should return an array of available bundle ids

### Description

Added the method `available_bundle_ids` in the `application` class - it makes sense to put it in there because it is something you d do before creating the application. 
Added client method and test for them.

### DIFF
--- a/spaceship/lib/spaceship/tunes/application.rb
+++ b/spaceship/lib/spaceship/tunes/application.rb
@@ -101,6 +101,10 @@ module Spaceship
                                     platform: platform,
                                     itunes_connect_users: itunes_connect_users)
         end
+
+        def available_bundle_ids(platform: nil)
+          client.get_available_bundle_ids(platform: platform)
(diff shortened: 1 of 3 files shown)


joshdholtz:
Thanks for showing an example of how to use this! That makes my job a lot easier 😊 Also, you can use `bundle console` instead of `bundle exec irb` 😉

joshdholtz (approved):
Looks good! Thanks for adding this 👍

PR #13003, opened 2018-07-31
Add Match manual for handling multiple apps per developer account

janpio:
These are changes from https://github.com/fastlane/docs/pull/542

closes https://github.com/fastlane/docs/pull/542

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
+++ b/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
@@ -146,6 +146,18 @@ end
 
 Then all your team has to do is `fastlane certificates` and keys, certs and profiles for all targets will be synced.
 
+#### Handle multiple apps per developer/distribution certificate
+If you want to use a single developer and/or distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
+
+Matchfile for both App #1 and #2:
+
+```ruby
+git_url("https://github.com/example/example-repo.git")
+git_branch("master")
+```
+
+_match_ will reuse certificates and will create separate provisioning profiles for each app.
+
 #### Passphrase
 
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/docs/sync_code_signing.md
```diff
@@ -146,6 +146,18 @@ end
...
+
+Matchfile for both App #1 and #2:
+
+```ruby
+git_url("https://github.com/example/example-repo.git")
+git_branch("master")
+```
+
+_match_ will reuse certificates and will create separate provisioning profiles for each app.
+
```
getaaron: How about adding:

```markdown
Alternatively, you can use one repo for distribution credentials, with an additional per-app repo for development credentials. An example is shown on [this Stack Overflow answer](https://stackoverflow.com/a/48591262/1445366).
```

janpio: Ping @vmalyi

janpio: That would be another method to handle multiple apps? 
Why only link to SO, not include it here?

getaaron: Including it here would be fine too.

joshdholtz: I'm not sure about adding this additional in here, TBH 🤔 _match_ is fine is almost every use case with having all the apps in one branch so I don't necessarily see a reason to explicitly call out putting them into separate branches per app. I feel like seeing this may cause more confusion and issues then solving them but that is just my hunch 🙃

getaaron: It’s probably because I always wind up at giant megacorps with several app teams but are still limited to 2 distribution certs, but I end up doing this frequently

joshdholtz (approved):

PR #13009, opened 2018-08-01
Version bump to 2.101.1

joshdholtz:
Auto-generated by fastlane 🤖

**Changes since release '2.101.0':**

* Return empty array when aab not set (#12998) via jrnk
* allow fastlane to fetch list of available bundle ids (#13001) via heartofagoof

### DIFF
--- a/fastlane/swift/Fastlane.swift
+++ b/fastlane/swift/Fastlane.swift
@@ -3963,7 +3963,7 @@ func xcov(workspace: String? = nil,
           coverallsServiceJobId: String? = nil,
           coverallsRepoToken: String? = nil,
           xcconfig: String? = nil,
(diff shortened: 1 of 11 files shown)


fastlane/swift/Fastlane.swift
```diff
@@ -3963,7 +3963,7 @@ func xcov(workspace: String? = nil,
           coverallsServiceJobId: String? = nil,
           coverallsRepoToken: String? = nil,
           xcconfig: String? = nil,
-          ideFoundationPath: String = "/Applications/Xcode.app/Contents/Developer/../Frameworks/IDEFoundation.framework/Versions/A/IDEFoundation",
+          ideFoundationPath: String = "/Applications/Xcode-9.2.app/Contents/Developer/../Frameworks/IDEFoundation.framework/Versions/A/IDEFoundation",
```
janpio: Hm, intentional?

joshdholtz: Oh dang! Good catch 🤔

PR #13043, opened 2018-08-06
Document api_token as optional in the action create_pull_request

dcordero:
### Checklist

### Motivation and Context
The example does not reflect the optionality of the parameter `api_token` in the action `create_pull_request`

### Description
The parameter `api_token`  in the action `create_pull_request` defaults to `ENV["GITHUB_API_TOKEN"]` which makes it optional. The example, on the other hand, does not reflect this optionality, and it uses a different env variable making it a bit confusing.


joshdholtz (approved):
Good catch! That is kind of unclear 😊 Thanks for making this!

PR #13064, opened 2018-08-08
[screengrab] replace double escape with simple escape so command also works on Windows

### LINKED ISSUE #13060: [screengrab] Can't figure out external storage path for screenshots on Windows
### New Issue Checklist

### Issue Description

On Windows `screengrab` can't figure out the external storage path (`determine_external_screenshots_path`) as the `adb` command that is being run doesn't work as expected:

```
INFO [2018-08-08 12:26:40.16]: $ C:\Android/platform-tools/adb -s 02e7f7e9215da7f8 shell echo \$EXTERNAL_STORAGE
INFO [2018-08-08 12:26:40.31]: ▸ $EXTERNAL_STORAGE
```

Normally this should output something like this:

```
[15:53:52]: $ adb -s 3004cc574d361200 shell echo \$EXTERNAL_STORAGE
[15:53:52]: ▸ /storage/emulated/legacy
```

Problem is that on Windows those `\` seems to be a problem or work differently:

```
C:\Projects
λ C:\Android/platform-tools/adb -s 02e7f7e9215da7f8 shell echo \\$EXTERNAL_STORAGE
\/sdcard

C:\Projects
λ C:\Android/platform-tools/adb -s 02e7f7e9215da7f8 shell echo \$EXTERNAL_STORAGE
$EXTERNAL_STORAGE

C:\Projects
λ C:\Android/platform-tools/adb -s 02e7f7e9215da7f8 shell echo $EXTERNAL_STORAGE
/sdcard
```

So you get the correct result with either `\\$EXTERNAL_STORAGE` or `$EXTERNAL_STORAGE` - but not `\$EXTERNAL_STORAGE`.

Code for the command is here:
https://github.com/fastlane/fastlane/blob/ce4eac3f45565b2f8b61f7a46732c170da063e2a/screengrab/lib/screengrab/runner.rb#L138-L143

Is this actually the correct escaping here? 
Or should there maybe only be one `\`?

janpio:
`adb` command included double escaped string which doesn't seem to be necessary but caused problems on Windows.

closes #13060

### DIFF
--- a/screengrab/lib/screengrab/runner.rb
+++ b/screengrab/lib/screengrab/runner.rb
@@ -136,7 +136,7 @@ module Screengrab
     end
 
     def determine_external_screenshots_path(device_serial)
-      device_ext_storage = run_adb_command("adb -s #{device_serial} shell echo \\$EXTERNAL_STORAGE",
+      device_ext_storage = run_adb_command("adb -s #{device_serial} shell echo \$EXTERNAL_STORAGE",
                                            print_all: true,
                                            print_command: true)
       File.join(device_ext_storage, @config[:app_package_name], 'screengrab')
--- a/screengrab/lib/screengrab/runner.rb
+++ b/screengrab/lib/screengrab/runner.rb
@@ -136,7 +136,10 @@ module Screengrab
     end
 
     def determine_external_screenshots_path(device_serial)
-      device_ext_storage = run_adb_command("adb -s #{device_serial} shell echo \$EXTERNAL_STORAGE",
+      # macOS evaluates $foo in `echo $foo` before executing the command,
+      # Windows doesn't - hence the double backslash vs. single backslash
+      command = Helper.windows? ? "shell echo \$EXTERNAL_STORAGE " : "shell echo \\$EXTERNAL_STORAGE"
+      device_ext_storage = run_adb_command("adb -s #{device_serial} #{command}",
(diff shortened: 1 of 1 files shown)


joshdholtz (changes requested):
The `\\` is needed for this to work on mac 😬 (I am not sure about Linux though)

I tested this PR on my mac and no screenshot were detected but things worked when switching back over to master

janpio:
Meh, there probably should be a test for that... 

I'll look into this a bit more.

joshdholtz:
That might be hard to test because that is a shell command that is actually getting run on an Android emulator I believe 🙃

janpio:
Hm, probably connected to the shell you are using :/

janpio:
Windows:

```
C:\Projects
λ echo $foo
$foo

C:\Projects
λ echo \$foo
\$foo
```

macOS:

```
MacBook-Pro:FastlaneDemoIonic sujan$ echo $foo

MacBook-Pro:FastlaneDemoIonic sujan$ echo \$foo
$foo
```

janpio:
New code now uses different command for Windows. 
I also added comment that explains, as I am sure this is not common knowledge.

I checked the codebase (searched for `\\$`) for similar cases, but didn't find anything relevant.

joshdholtz (approved):
This works on mac 👍

PR #13078, opened 2018-08-09
Invalidate group list when a group is created

paramv17:
### Checklist

### Motivation and Context

Because of the way caching and creating works for testflight groups - if you try to query the created group during the same session it was created, it won't show up in the list of groups. invalidating the cached groups for that app ids after creating a new group allows the list to be updated properly. 

```
bundle console
require ‘spaceship’
Spaceship::Tunes.login(name, password)
Spaceship::Tunes.select_team
# Create a new group with new name
Spaceship::TestFlight::Group.create!(app_id: app_id ,group_name: some_name)
```

After that - without this change - if you query 

```
Spaceship::TestFlight::Group.all(app_id: app_id)
```

The new group won't be listed in the groups returned. 

It will be with this change. 

### Description

Deleted the entry for the app_id in `cached_groups` when a group associated with the app id is created.

### DIFF
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -114,7 +114,7 @@ module Spaceship
         end
 
         # This is invalid now.
-        @cached_groups.delete(app_id)
+        @cached_groups.delete(app_id) if @cached_groups
 
         handle_response(response)
       end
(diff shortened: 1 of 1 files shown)


getaaron (approved):

getaaron:
Thanks for the PR, @heartofagoof!

PR #13085, opened 2018-08-10
Output supported platforms in action tables

janpio:
Step 1 of https://github.com/fastlane/docs/issues/649 is to show the supported platforms of an action. This PR outputs adds this.

### DIFF
--- a/fastlane/lib/assets/Actions.md.erb
+++ b/fastlane/lib/assets/Actions.md.erb
@@ -32,11 +32,11 @@ For _fastlane_ plugins, check out the [available plugins](https://docs.fastlane.
 <%- @categories.each do |category, actions| %>
 # <%= category %>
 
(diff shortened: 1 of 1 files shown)


getaaron (approved):
👍

joshdholtz (commented):
Can you include a screenshot for doc changes so its easier to see what this change visually looks like at a quick glance? 😊

janpio:
Nope, as I can't create the docs on Windows. (Also honestly no idea what the proper process would be to get these generated changes from `fastlane` over to `docs` locally)

joshdholtz (approved):

janpio:
Thanks for the explanation. (Is this documented somewhere?)

Now I only have to make `mkdocs serve` work on Windows for the next time ;)

PR #13088, opened 2018-08-10
[CI] Make generated docs (`validated_docs`) downloadable as artifact on Circle

### LINKED ISSUE #13086: [CI] Make built docs downloadable as artifact
Being able to download and look at the generated documentation of a PR would be nice.

#### Feature Description
During the "Validate Documentation" step of the CI build the lane `validate_docs` is run. This also creates the following output:

> [16:13:44]: Generated new docs on path /tmp/fl_clone20180810-82-1aayx3g/docs

The content of this folder could be zipped up and stored as a Circle artifact: https://circleci.com/docs/2.0/artifacts/

Run the lane locally and look at the files there.

janpio:
### Checklist

### Motivation and Context
Circle runs the lane `validated_docs` to validate the generated docs. You can [define files to be stored as "artifacts" of a test run](https://circleci.com/docs/2.0/artifacts/) at Circle. Having those generated docs enables us to look at them manually as well.

### Description
Adds a new ENV var that is used in `clone_docs` to clone to a specific location, not just in a temp dir. This param is used in CI to save the generated docs to a known path (vs. the unknown generated tmp dir name) that can be saved as an artifact.

### Result
The result can be seen here: https://circleci.com/gh/fastlane/fastlane/22440#artifacts/containers/0 Click on validated_docs/ -> docs/ -> mkdocs.yml to download the generated file and see that all the action files are listed.

---
closes #13086


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
```diff
@@ -185,7 +185,7 @@ def generate!(target_path: nil)
         FileUtils.cp(current_asset_path, File.join(docs_dir, "img", "actions", File.basename(current_asset_path)))
       end
 
-      UI.success("Generated new docs on path #{target_path}")
+      UI.success("Generated new docs files (actions.md, actions/*.md, mkdocs.yml, assets) on path #{target_path}")
```
getaaron: Maybe we should generate `(actions.md, actions/*.md, mkdocs.yml, assets)` from the actual changes instead of hardcoding them as a string

getaaron (approved):
This looks good. Had one suggestion if you're interested but doesn't need to block this getting merged.

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
janpio: I thought about building that string dynamically, adding an entry with each "step" - but that would just make the code more complicated for almost no benefit. Same for somehow using the actual changed files to generate this string out of it - super complicated for almost no use.

Or did I misunderstand? Is there a simple way I missed?

getaaron: I didn’t look at a way to implement it but I imagined the latter. I just noticed it as a future maintenance point and it doesn’t feel very “fastlaney”. But it’s also not really a problem either since this list isn’t likely to change often. 🤷🏻‍♀️

fastlane/Fastfile
```diff
@@ -406,23 +406,27 @@ lane :update_docs do |options|
   end
 end
 
-def clone_docs(debug: false)
+def clone_docs(debug: false, target_dir: nil)
```
joshdholtz: Instead of making this a parameter that gets passed in, can we make this an environment variable called `FASTLANE_CLONE_PATH`? This removes the hardcoding of `/tmp/validated_docs` on line 319 which won't work on Windows? We can then set this path dynamically on any CI system

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
joshdholtz: I think I agree that this doesn't feel very "fastlaney". I _think_ I would lean towards keeping what was originally here and if someone is doesn't know and is curious what all got generated, they can just go look at that path 🙃

joshdholtz (dismissed):

fastlane/Fastfile
janpio: Would have to be `FASTLANE_DOCS_CLONE_PATH` as this refers to the docs I think. But yep, sounds useful. Will make the change.

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
janpio: Unfortunately not, as this is _inside_ a full clone of the docs where some additional files get added or modified.

Instead of one `UI.success` we could add an individual `UI.message` for each step (which would also make it consistent with the output for the assets).

fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lib/fastlane/documentation/markdown_docs_generator.rb
janpio: Also: fastlane is doing it this way in other places: https://github.com/fastlane/fastlane/blob/665a23e6351702bbd6980fe1f7ec75a6f1f7bfa6/fastlane/Fastfile#L392

.circleci/config.yml
```diff
@@ -125,7 +128,10 @@ jobs:
 
       - run: bundle exec fastlane validate_docs
 
-  "Lint Source Code":
+      - store_artifacts:
+          path: /tmp/validated_docs
```
janpio: Unfortunately CircleCI 2 doesn't let you use env variables here :/
(Correct @DanToml?)

fastlane/Fastfile
```diff
@@ -326,21 +326,26 @@ lane :validate_docs do
 
   # Test if generating the docs is successful
   clone_docs do
+    # Test if generating the docs is successful
     generate_markdown_docs(target_path: ".")
```
joshdholtz: Could we just move this `generate_markdown_docs(target_path: ".")` into the `verify_docs` lane? And then just call the 'verify_docs` instead of doing the clone here?

fastlane/Fastfile
```diff
@@ -326,21 +326,26 @@ lane :validate_docs do
...
-      puts(`sed -i -e "s/activate_bin_path/bin_path/g" $(which bundle)`) # workaround for bundler https://github.com/bundler/bundler/issues/4602#issuecomment-233619696
-      sh("bundle install")
-      sh("bundle exec fastlane test skip_building:true") # skip_building since we don't have a proper python environment set up
-    end
+    run_docs_tests
+  end
+end
+
+desc "Makes sure the tests on https://docs.fastlane.tools still work with the latest version"
+private_lane :run_docs_tests do
```
joshdholtz: Even though this does look good in its own lane I don't think it 100% works since this is required to run within the context of the `clone_docs` block (because it needs to run in a clone docs directory). 

If we move the `generate_markdown_docs(target_path: ".")` call within the `verify_docs` lane (like mentioned above), we can move this code back into `verify_docs` and not have any duplicate code

joshdholtz (changes requested):
A few small comments/change requests 😊

fastlane/Fastfile
janpio: This would change what `verify_docs` does. As it is a public lane, I wasn't sure if someone/something is depending on this behavior and tried _not_ to change it. Because verifying something and generating some more into something is a bit different. Thoughts?

fastlane/Fastfile
janpio: That's why its a private lane that can not be run by itself. Other comments see above.

joshdholtz: I think I'm fine with leaving this as is then but maybe add a comment that saying this needs to run in the directory/scope of the docs directory? My biggest concern is that this can **ALSO** run "successfully" in the scope of this directory because there is also a `test` lane. That is why I would prefer if we could keep this code inside of a `clone_docs` block 🙃

janpio: Can you just make the necessary changes in this PR? I'm totally fine with cleaning up a little here, but I am not 100% I would solve it correctly.

janpio:
ping @joshdholtz

fastlane/Fastfile
janpio: Waiting for you on this one @joshdholtz

janpio:
This is still a useful feature that could help anyone working on docs generation in the future. I think the code is good (enough), I hope you agree when you take another look @joshdholtz. (Even if not, I think merging is better than letting it die here...)

endocrimes (approved):
Circle changes lgtm, will defer to josh on the generation

joshdholtz:
Oh yeah, I like this change! I'll rebase this guy so we can get some fresh tests to make sure everything is still 💯

joshdholtz:
This failure is super weird 🤔 That action totally exists and that option totally exists 😱

janpio:
Errors are strange:

```
ERROR [2019-09-28 23:03:26.54]: Found errors in the documentation, more information above
ERROR [2019-09-28 23:03:26.54]: Error found in code sample:
ensure_env_vars(
  env_vars: ['GITHUB_USER_NAME', 'GITHUB_API_TOKEN']
)
Could not find method or action named 'ensure_env_vars'
ERROR [2019-09-28 23:03:26.54]: Error found in code sample:
onesignal(
  app_id: "Your OneSignal App ID",
  auth_token: "Your OneSignal Auth Token",
  app_name: "New Name for OneSignal App",
  android_token: "Your Android GCM key (optional)",
  android_gcm_sender_id: "Your Android GCM Sender ID (optional)",
  apns_p12: "Path to Apple .p12 file (optional)",
  apns_p12_password: "Password for .p12 file (optional)",
  apns_env: "production/sandbox (defaults to production)"
)
```

- ensure_env_vars:
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/ensure_env_vars.rb totally exists, but https://docs.fastlane.tools/actions/ensure_env_vars/ does not (yet?)

- onesignal:
Current public docs at https://docs.fastlane.tools/actions/onesignal/ indeed do not have `app_id`, but https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/onesignal.rb does - so this is a false positive. 

Is it comparing to older state of the actions somehow?

janpio:
Ok, I think I semi understood it:
- fastlane runs `validate_docs`. That clones the docs, generates the new file and runs `run_docs_tests`. 
- `run_docs_tests` executes the tests of the docs which were cloned. This includes `ensure_code_samples`. As the actions are regenerated, it finds things like `ensure_env_var` and `opensignal.app_id` in there, but https://github.com/fastlane/docs/blob/master/fastlane/actions/test_sample_code.rb somehow can not test for that properly.

---

The best way forward is probably to really refactor the `Fastfile`s of both `fastlane/fastlane` and `fastlane/docs` to clean this mess up and make it much easier to understand what is going on.

PR #13093, opened 2018-08-12
Add comment for incorrect env names in pilot and snapshot

revolter:
### Checklist

### DIFF
--- a/pilot/lib/pilot/options.rb
+++ b/pilot/lib/pilot/options.rb
@@ -22,6 +22,7 @@ module Pilot
                                      description: "The bundle identifier of the app to upload or manage testers (optional)",
                                      optional: true,
                                      code_gen_sensitive: true,
+                                     # This incorrect env name is here for backwards compatibility
                                      default_value: ENV["TESTFLIGHT_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
                                      default_value_dynamic: true),
         FastlaneCore::ConfigItem.new(key: :app_platform,
--- a/snapshot/lib/snapshot/options.rb
+++ b/snapshot/lib/snapshot/options.rb
@@ -122,6 +122,7 @@ module Snapshot
                                      optional: true,
                                      description: "The bundle identifier of the app to uninstall (only needed when enabling reinstall_app)",
                                      code_gen_sensitive: true,
+                                     # This incorrect env name is here for backwards compatibility
                                      default_value: ENV["SNAPSHOT_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
                                      default_value_dynamic: true),
         FastlaneCore::ConfigItem.new(key: :add_photos,


janpio:
`IDENTITIFER`? Haha.

revolter:
Shit happens ¯\_(ツ)_/¯ 😆

pilot/lib/pilot/options.rb
```diff
@@ -22,7 +22,7 @@ def self.available_options
                                      description: "The bundle identifier of the app to upload or manage testers (optional)",
                                      optional: true,
                                      code_gen_sensitive: true,
-                                     default_value: ENV["TESTFLIGHT_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
+                                     default_value: ENV["PILOT_APP_IDENTIFIER"] || ENV["TESTFLIGHT_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
```
joshdholtz: Is this actually needed since `PILOT_APP_IDENTIFIER` is set as the `env_name` for this option? 🤔

snapshot/lib/snapshot/options.rb
```diff
@@ -122,7 +122,7 @@ def self.available_options
                                      optional: true,
                                      description: "The bundle identifier of the app to uninstall (only needed when enabling reinstall_app)",
                                      code_gen_sensitive: true,
-                                     default_value: ENV["SNAPSHOT_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
+                                     default_value: ENV["SNAPSHOT_APP_IDENTIFIER"] || ENV["SNAPSHOT_APP_IDENTITIFER"] || CredentialsManager::AppfileConfig.try_fetch_value(:app_identifier),
```
joshdholtz: Same here. `SNAPSHOT_APP_IDENTIFIER` is already set as the `env_name`.  Is this just duplicate code? 🤔

joshdholtz (changes requested):
Two questions!

pilot/lib/pilot/options.rb
joshdholtz: I don't think `try_fetch_value` gets called if the `PILOT_APP_IDENTIFIER` env is already set because there the `env_name` has a higher "default value" priority than `default_value` 😖 But I will test to verify my understandings are correct

joshdholtz: Docs show the ENV should have higher priority - https://docs.fastlane.tools/advanced/#priorities-of-parameters-and-options

Also, this shows if the `env_name` is defined and if that env actually has a value, that will get loaded first and the `default_value` won't get called - https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/configuration.rb#L219-L227

I _think_ that the only reason the `_IDENTITIFER` env names are in the `default_value` are for backwards compatibility for a mistake that was made 😁

revolter: Oh, so then this PR is invalid, right?

janpio: (Please create a PR that documents via a comment why those ENV vars are there as a default_value then.)

revolter: @janpio, That was my plan too 😛

revolter:
@joshdholtz, @janpio, Updated.

janpio (approved):

joshdholtz (approved):

PR #10528, opened 2017-10-06
Update code samples to pass docs tests

KrauseFx:
Will submit a separate follow-up PR to run the docs test on the fastlane main repo also, I thought we did that

PR #13102, opened 2018-08-14
[match] Fix code sample

janpio:
This code sample is being evaluated as a lane as it is marked as `ruby`. But a `Matchfile` uses methods, that are not available in that content, so those have to be marked in this way.

(This did not trigger in the tests because I think parts of it are broken and inconsistent. While trying to fix stuff, this caused the tests to fail and sent me down the rabbit hole to learn about `ruby-skip-tests`.)

Background see https://github.com/fastlane/fastlane/pull/10528

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
+++ b/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
@@ -91,7 +91,7 @@ username("user@fastlane.tools")
 
 _match_ also supports storing certificates of multiple teams in one repo, by using separate git branches. If you work in multiple teams, make sure to set the `git_branch` parameter to a unique value per team. From there, _match_ will automatically create and use the specified branch for you.
 
-```ruby-skip-tests
+```ruby
 match(git_branch: "team1", username: "user@team1.com")
 match(git_branch: "team2", username: "user@team2.com")
 ```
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/docs/sync_code_signing.md
```diff
@@ -91,7 +91,7 @@ username("user@fastlane.tools")
 
 _match_ also supports storing certificates of multiple teams in one repo, by using separate git branches. If you work in multiple teams, make sure to set the `git_branch` parameter to a unique value per team. From there, _match_ will automatically create and use the specified branch for you.
 
-```ruby
+```ruby-skip-tests
 match(git_branch: "team1", username: "user@team1.com")
```
joshdholtz: >This code sample is being evaluated as a lane as it is marked as ruby. But a Matchfile uses methods, that are not available in that content, so those have to be marked in this way.

I'm confused here 🤔 This is totally syntax that would show up in a `lane` and not in a `MatchFile` so should being marked as `ruby` the correct thing?

janpio: You are totally right, I changed the wrong code sample. Now the `ruby-skip-tests` is for a sample that uses "options methods" outside of a lane, like it would in a `Matchfile`. 

Thanks for catching this.

joshdholtz (approved):

PR #13113, opened 2018-08-15
New action `download_from_play_store` that does as an action what `fastlane supply init` does as a command

janpio:
This is a proof of concept for one of the ideas I suggested in https://github.com/fastlane/docs/issues/671: adding actions for all the `fastlane <toolname> init` commands that do more than creating a config file.

- Similar to `upload_to_play_store` this is a thin wrapper around the code that is already present in `supply`. 
- To make the options list more manageable, I hardcoded a manual list of options this action actually needs and removed all the others.
- As the action doesn't contain its actual code, I had to exclude it from the check in `fastlane/spec/unused_options_spec.rb` as well.

### DIFF
--- a/fastlane/lib/fastlane/actions/download_from_play_store.rb
+++ b/fastlane/lib/fastlane/actions/download_from_play_store.rb
@@ -1,0 +1,61 @@
+module Fastlane
+  module Actions
+    class DownloadFromPlayStoreAction < Action
+      def self.run(params)
+        require 'supply'
+        require 'supply/options'
+
+        Supply.config = params # we already have the finished config
+
+        require 'supply/setup'
+        Supply::Setup.new.perform_download
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Download metadata and binaries from Google Play (via _supply_)"
+      end
+
+      def self.details
+        "More information: https://docs.fastlane.tools/actions/download_from_play_store/"
+      end
+
+      def self.available_options
+        require 'supply'
+        require 'supply/options'
+        options = Supply::Options.available_options.clone
+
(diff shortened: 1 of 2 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

janpio:
Turns out https://docs.fastlane.tools/actions/google_play_track_version_codes/ does something similar by using `supply` to power another action.

joshdholtz:
I am currently torn about this action 🤔 The code itself looks 💯 and the idea of pulling a part of a tool into the action is also 💯 but I feel like if this is going to be done on a large scale (for fastlane 3.0) I feel like there might need to be a bit more thought/planning into how these are going to get organized. If everything is done this way, I feel like list of actions are going to get pretty long and unorganized 🤔

janpio:
This PR was my try to see if the "actionification" could happen without the "let's wait for 3.0". I'm sure this will happen some day, but I see no real reason to wait with things like this until then.

The only thing that gets defined now by merging this, is the action name. I defined those for all existing command line commands (I could find) in [a Google spreadsheet](https://docs.google.com/spreadsheets/d/1HnflxsVyuj1nprHqE5uMnHsuiQ5SQRdZ-0Z6x8B4og4/edit#gid=0). 

The logical next step will be to have additional "grouping" for actions - but that's a thing for 3.0 I guess as it is really something new and not just iterating on what we already have.

joshdholtz (approved):
This looks good and works good! Let's ship it 😊

PR #13134, opened 2018-08-16
[scan] should_zip_build_products sets SCAN_ZIP_BUILD_PRODUCTS_PATH in lane context

### LINKED ISSUE #13108: Add lane context for scan build_for_tests build artifact
### Feature Request
Export a lane context for zipped build_for_test build.

Integration w/ Firebase Testlab plugin.

#### Feature Description

At https://github.com/fastlane/fastlane/blob/master/scan/lib/scan/runner.rb#L106, the scan action would zip a build_for_test into a zip file. It would be nice to have the path of the zip file exported through lane_context so other plugins can use the artifact.

The most suitable place to add the context appears to be https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/run_tests.rb#L35, but we still need a way to either feed back the path from scan lib, or let the run_tests action dictate the path and transmit it into the scan lib.

joshdholtz:
Fixes #13108

## Problem
The path generated by _scan_'s `should_zip_build_products` was not nicely available for anybody to use.

## Solution
Now it is available via `SCAN_ZIP_BUILD_PRODUCTS_PATH` in lane context

### DIFF
--- a/scan/lib/scan/runner.rb
+++ b/scan/lib/scan/runner.rb
@@ -94,23 +94,19 @@ module Scan
       end
     end
 
-    def self.zip_build_products_path
+    def zip_build_products
       return unless Scan.config[:should_zip_build_products]
+
+      # Gets :derived_data_path/Build/Products directory for zipping zip
+      derived_data_path = Scan.config[:derived_data_path]
+      path = File.join(derived_data_path, "Build/Products")
 
       # Gets absolute path of output directory
       output_directory = File.absolute_path(Scan.config[:output_directory])
       output_path = File.join(output_directory, "build_products.zip")
 
-      return output_path
-    end
-
-    def zip_build_products
-      output_path = Scan::Runner.zip_build_products_path
-      return unless output_path
-
-      # Gets :derived_data_path/Build/Products directory for zipping zip
-      derived_data_path = Scan.config[:derived_data_path]
-      path = File.join(derived_data_path, "Build/Products")
+      # Caching path for action to put into lane_context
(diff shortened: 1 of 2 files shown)


janpio:
(I like your PR template)

scan/lib/scan/runner.rb
```diff
@@ -94,17 +94,24 @@ def handle_results(tests_exit_status)
...
       # Gets absolute path of output directory
       output_directory = File.absolute_path(Scan.config[:output_directory])
       output_path = File.join(output_directory, "build_products.zip")
 
+      return output_path
+    end
+
+    def zip_build_products
+      output_path = Scan::Runner.zip_build_products_path
+      return unless output_path
```
janpio: Isn't that a bit more cumbersome and harder to understand than just to have `return unless Scan.config[:should_zip_build_products]` here once again, and only then get the output_path?

joshdholtz: Ehhhh, yes and no. I didn't want `Scan.config[:should_zip_build_products]` to have to instances where it was getting looked up but I can see why this might not be ideal either 🤔

joshdholtz: I will rework

joshdholtz:
@janpio Reworked this to use `Scan.cache` instead which felt a little bit better 🙃

janpio (approved):
So simple.

janpio:
Is this cache thing a pattern we use more often?

PR #13107, opened 2018-08-14
[CI] check "special action docs" code samples

janpio:
Until now the code samples of the "special action docs" (https://github.com/fastlane/fastlane/tree/master/fastlane/lib/fastlane/actions) were not tested at all. This PR adds a new private lane to the `Fastfile` that does this.

This PR includes https://github.com/fastlane/fastlane/pull/13105 (which was first proposed as a standalone PR, but then merged here because of convenience)

PR #13138, opened 2018-08-16
Fix ensure code samples where a block in all_actions was required

joshdholtz:
`all_actions` requires a block (and isn't an array) so doing count this way


janpio (approved):
Haha, exact same fix waiting in another PR: https://github.com/fastlane/fastlane/pull/13107/files#diff-56c5021ef06ac6cea807ae53269b6f4fR508

PR #13148, opened 2018-08-17
iTMSTransporter support on Windows

janpio:
### Checklist

### Motivation and Context
iTMSTransporter is available for Windows, but fastlane's implementation doesn't support it yet.

### Description
These commits add support for running `fastlane pilot upload` / `upload_to_testflight` and `fastlane deliver` / `upload_to_app_store` with binaries on Windows.

If iTMSTransporter is not installed, the tests still pass (as they just generate the command, but don't execute it) but on actual execution of `upload_to_testflight` or `upload_to_app_store` you get this error message:
```
...fastlane/fastlane_core/lib/fastlane_core/ui/interface.rb:141:in `user_error!': [!] Could not find transporter at usual locations. Please use environment variable `FASTLANE_ITUNES_TRANSPORTER_PATH` to specify your installation path. (FastlaneCore::Interface::FastlaneError)
```

Along the way several smaller bugs in the implementation were fixed and tests added (e.g. for special handling of Xcode >= 9).

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -219,7 +219,7 @@ module FastlaneCore
         UI.user_error!("Could not find transporter at usual locations. Please use environment variable `FASTLANE_ITUNES_TRANSPORTER_PATH` to specify your installation path.")
       else
         # not Mac or Windows
-        return '' 
+        return ''
       end
     end
 
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -200,16 +200,25 @@ module FastlaneCore
     # @return the full path to the iTMSTransporter executable
     def self.itms_path
       return ENV["FASTLANE_ITUNES_TRANSPORTER_PATH"] if FastlaneCore::Env.truthy?("FASTLANE_ITUNES_TRANSPORTER_PATH")
-      return '' unless self.mac? # so tests work on Linux and Windows too
+      return '' unless self.mac? || self.windows? # so tests work on Linux too
 
-      [
-        "../Applications/Application Loader.app/Contents/MacOS/itms",
-        "../Applications/Application Loader.app/Contents/itms"
-      ].each do |path|
-        result = File.expand_path(File.join(self.xcode_path, path))
-        return result if File.exist?(result)
+      if self.mac?
+        [
+          "../Applications/Application Loader.app/Contents/MacOS/itms",
+          "../Applications/Application Loader.app/Contents/itms"
+        ].each do |path|
+          result = File.expand_path(File.join(self.xcode_path, path))
+          return result if File.exist?(result)
+        end
+        UI.user_error!("Could not find transporter at #{self.xcode_path}. Please make sure you set the correct path to your Xcode installation.")
+      elsif self.windows?
+        [
+          "C:/Program Files (x86)/itms"
+        ].each do |path|
+          return path if File.exist?(path)
+        end
+        UI.user_error!("Could not find transporter at usual locations. Please use environment variable `FASTLANE_ITUNES_TRANSPORTER_PATH` to specify your installation path.")
       end
-      UI.user_error!("Could not find transporter at #{self.xcode_path}. Please make sure you set the correct path to your Xcode installation.")
     end
 
     # keychain
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -200,7 +200,6 @@ module FastlaneCore
     # @return the full path to the iTMSTransporter executable
     def self.itms_path
       return ENV["FASTLANE_ITUNES_TRANSPORTER_PATH"] if FastlaneCore::Env.truthy?("FASTLANE_ITUNES_TRANSPORTER_PATH")
-      return '' unless self.mac? || self.windows? # so tests work on Linux too
 
       if self.mac?
         [
@@ -218,6 +217,9 @@ module FastlaneCore
           return path if File.exist?(path)
         end
         UI.user_error!("Could not find transporter at usual locations. Please use environment variable `FASTLANE_ITUNES_TRANSPORTER_PATH` to specify your installation path.")
+      else
+        # not Mac or Windows
+        return '' 
       end
     end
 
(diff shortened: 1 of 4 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -193,22 +193,32 @@ def self.transporter_java_path
...
+        end
+        UI.user_error!("Could not find transporter at #{self.xcode_path}. Please make sure you set the correct path to your Xcode installation.")
+      elsif self.windows?
+        [
+          "C:/Program Files (x86)/itms"
+        ].each do |path|
+          return path if File.exist?(path)
+        end
+        UI.user_error!("Could not find transporter at usual locations. Please use environment variable `FASTLANE_ITUNES_TRANSPORTER_PATH` to specify your installation path.")
       end
```
joshdholtz: Can we add an `else` here that does that `return ''`?

I can see a future issue right now here we get that early return on line 203 with `return '' unless self.mac? || self.windows?` because something was changed but then this if statement wasn't change and then this bug is hard to track down 😊

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -61,16 +61,13 @@ def execute(command, hide_output)
...
-            Process.wait(pid)
           end
         end
       rescue => ex
         @errors << ex.to_s
       end
 
-      exit_status = $?.exitstatus
-      unless exit_status.zero?
+      if exit_status != 0
```
joshdholtz: Is there a reason for not keeping the user of `.zero?` here? (and the other spot below)

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -187,13 +185,16 @@ def build_download_command(username, password, apple_id, destination = "/tmp", p
 
     def handle_error(password)
       # rubocop:disable Style/CaseEquality
-      unless password === /^[0-9a-zA-Z\.\$\_]*$/
+      # rubocop:disable Style/YodaCondition
+      unless /^[0-9a-zA-Z\.\$\_\-]*$/ === password
```
joshdholtz: Is there a reason for flopping this comparison around? 🤔

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -419,8 +424,11 @@ def load_password_for_transporter
     def handle_two_step_failure(ex)
       if ENV[TWO_FACTOR_ENV_VARIABLE].to_s.length > 0
         # Password provided, however we already used it
-        UI.error("Application specific password you provided using #{TWO_FACTOR_ENV_VARIABLE}")
-        UI.error("is invalid, please make sure it's correct")
+        UI.error("")
+        UI.error("Application specific password you provided using")
+        UI.error("environment variable #{TWO_FACTOR_ENV_VARIABLE}")
+        UI.error("is invalid, please make sure it's correct.")
```
joshdholtz: Can you remove the `.`? Its fastlane style to not add those 😉

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -433,13 +441,15 @@ def handle_two_step_failure(ex)
         UI.error("Please make sure to follow the instructions")
         a.remove_from_keychain
       end
+      UI.error("")
       UI.error("Your account has 2 step verification enabled")
       UI.error("Please go to https://appleid.apple.com/account/manage")
       UI.error("and generate an application specific password for")
-      UI.error("the iTunes Transporter, which is used to upload builds")
+      UI.error("the iTunes Transporter, which is used to upload builds.")
```
joshdholtz: Same here with removing the `.`

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -433,13 +441,15 @@ def handle_two_step_failure(ex)
...
+      UI.error("")
       UI.error("To set the application specific password on a CI machine using")
       UI.error("an environment variable, you can set the")
-      UI.error("#{TWO_FACTOR_ENV_VARIABLE} variable")
+      UI.error("#{TWO_FACTOR_ENV_VARIABLE} variable.")
```
joshdholtz: Same here with removing the `.`

joshdholtz (dismissed):
Some small questions and changes

fastlane_core/lib/fastlane_core/itunes_transporter.rb
joshdholtz: It is functionally equivalent but its a better way of making sure that we are comparing an number to a number. I would personally keep the `.zero?` 😊

fastlane_core/lib/fastlane_core/itunes_transporter.rb
joshdholtz: Its just the way things have been generally printed out in the rest of fastlane so would probably like to keep things consistent if possible 😊

fastlane_core/lib/fastlane_core/itunes_transporter.rb
joshdholtz: Ah okay. I didn't think there was a difference in what was on left/right side but wasn't sure. So this _could_ be changed back and things would still work the same?

janpio: No. If you change it back, it doesn't work. 
`rubocop` actually wants to write it the other way around, which is probably why this was the other way around and didn't work in the first place - I had to add the `rubocop:disable Style/YodaCondition` to quiet it.

fastlane_core/lib/fastlane_core/itunes_transporter.rb
janpio: ok.

joshdholtz:
@janpio Can we get a rebase on this since I merged one of your PRs that conflicts with a change in here? 😇

janpio:
Conflict could be solved in GitHub web UI fortunately ;) Still passes @joshdholtz

joshdholtz (approved):
Windows support!!! 🙌

janpio:
(One tiny part of it anyway ;) )

PR #13161, opened 2018-08-19
[screengrab] Docs: add instructions how to build APKs via lane

janpio:
No reason why people have to build this manually, the old way. 
We have an action that does the exact same thing.

Related: https://github.com/fastlane/docs/pull/686


joshdholtz (approved):
Niiiicccceeeee! This should hopefully prevent a lot of issue that come in asking this question 😊

joshdholtz (approved):

PR #13163, opened 2018-08-19
[gradle] Docs: Add `task: "clean"` example

janpio:
Cleaning the project to get rid of the output folder and generated APKs is a quite useful and common task, so I aded it as a code example to the `gradle` action.


joshdholtz (approved):
Very clean PR 👍

PR #13166, opened 2018-08-20
measure fastlane startup time via execution time of `--help` command 🏎️

janpio:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
Fastlane can be pretty slow to start. Slow startup is bad.

### Description
This PR adds a new lane `benchmark_help_command` that is also run during `execute_tests`. It installs a (cross platform) node tool via npm called [`gnomon`](https://github.com/paypal/gnomon) that prepends timestamps to lines of out and measures the total execution time. This total is extracted and output in the lane. To make sure we get a good overview, the lane runs `fastlane --help` 3 times and outputs it runtime.

In the future this lane can be used to benchmark `fastlane` and that way track and measure the effect of any changes (like https://github.com/fastlane/fastlane/pull/13167).

PR #13167, opened 2018-08-20
Improve fastlane startup time using bootsnap - behind feature flag (-61% on macOS, -65% on Ubuntu, -7% on Windows!)

janpio:
### Checklist

### Motivation and Context
This PR is another try to integrate bootsnap (see previous PR at https://github.com/fastlane/fastlane/pull/11230) into fastlane.

`fastlane` takes quite long to start up as it requires a lot of files. This can be sped up by using https://github.com/Shopify/bootsnap.

### Description
I added the gem, the necessary code and configuration to `bin/fastlane` and tested the configuration options to find one that works on all platforms. A one line code change in `Fastfile` was necessary, where a `require` had to be replaced with `require_relative`.

---

<strike>This PR currently contains unrelated commits from https://github.com/fastlane/fastlane/pull/13166 that were necessary to be able to measure the startup time. I can rebase this one when the other PR is merged for a cleaner "Files changed" view.</strike>


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

janpio:
Hmmm... the main thing doesn't work and breaks on _all_ platforms.

Here are the other results:
image
compile_cache_yaml + compile_cache_iseq work pretty well on Ubuntu and macOS, improving the execution time by 61% or 41%. But on Windows it adds 7%

janpio:
Ok, after disabling one problematic `require`, we now how these sweet results:

image

-65% on Ubuntu, -43% on macOC, -7% on Windows :D

bin/fastlane
```diff
@@ -4,6 +4,30 @@ if RUBY_VERSION < '2.0.0'
...
 
+def self.windows?
+  # taken from: https://stackoverflow.com/a/171011/1945875
+  (/cygwin|mswin|mingw|bccwin|wince|emx/ =~ RUBY_PLATFORM) != nil
+end
+
+require 'bootsnap'
+require 'tmpdir'
+require 'fileutils'
+# env = ENV['RAILS_ENV'] || "development"
```
joshdholtz: This can probably get 💣 ed?

bin/fastlane
```diff
@@ -4,6 +4,30 @@ if RUBY_VERSION < '2.0.0'
...
+require_relative "../fastlane/lib/fastlane/version.rb"
+cache_dir = "#{Dir.tmpdir}/fastlane/bootsnap-cache/" + Fastlane::VERSION
+unless File.directory?(cache_dir)
+  FileUtils.mkdir_p(cache_dir)
+end
+Bootsnap.setup(
+  cache_dir:            cache_dir, # Path to your cache
+  development_mode:     false, # env == 'development', # Current working environment, e.g. RACK_ENV, RAILS_ENV, etc
+  load_path_cache:      true, # Optimize the LOAD_PATH with a cache
+  autoload_paths_cache: false, # Optimize ActiveSupport autoloads with cache
```
joshdholtz: This comment about "ActiveSupport" doesn't seem relevant as we don't use `ActiveSupport`?

bin/fastlane
```diff
@@ -4,6 +4,30 @@ if RUBY_VERSION < '2.0.0'
...
+require 'fileutils'
+# env = ENV['RAILS_ENV'] || "development"
+require_relative "../fastlane/lib/fastlane/version.rb"
+cache_dir = "#{Dir.tmpdir}/fastlane/bootsnap-cache/" + Fastlane::VERSION
+unless File.directory?(cache_dir)
+  FileUtils.mkdir_p(cache_dir)
+end
+Bootsnap.setup(
+  cache_dir:            cache_dir, # Path to your cache
+  development_mode:     false, # env == 'development', # Current working environment, e.g. RACK_ENV, RAILS_ENV, etc
```
joshdholtz: Can probably remove all of the comments on this line? `RACK_ENV` and `RAILS_ENV` are not relevant at all

joshdholtz (changes requested):
**Thought 1:**
This PR also seems to also include #13166 so should probably do something about that 😊 

**Thought 2:**
I think I have the same concerns about this as the last PR that attempted to do this. Mainly the points in this comment - https://github.com/fastlane/fastlane/pull/11230#issuecomment-352062659.

If this has been thoroughly tested across all those different installation methods I think I would feel better about this.

One thought is to **maybe** wrap this in an environment variable so its opt-in for a while? 🤔

bin/fastlane
janpio: Exactly - which is why this is set to `false`. 
Those comments are the default ones suggested by `bootsnap` documentation.

bin/fastlane
janpio: Yes, see previous comment.

I left the `development_mode` stuff in, as fastlane doesn't seem to have something like this - and this would be one of the cases where it would probably be useful. (Although I didn't have any negative effects myself while developing stuff with `bootsnap` enabled)

janpio:
Re "Thought 1": Yep, see my comment in the description. "Someone" could review #13166 and I'll be happy to rebase this.

Re "Thought 2": Right.

> One thought is to maybe wrap this in an environment variable so its opt-in for a while?

Would this actually help with the "many different install methods" problem? 
Of course wrapping this in a "feature flag" first would be totally fine to me.

joshdholtz:
@janpio Wrapping this in a feature flag would allow us to make sure this runs on all the platforms through all the install methods properly after getting packaged up without the fear of this 100% breaking things for a user. I just a fear that since if there is an issue with this somehow **and** that its at the beginning of the fastlane execution process that it could completely prevent a user from using fastlane. A lot of this fear just comes from not knowing all the ins and outs of how bootsnap works 🙃

janpio:
Sure. Suggestion for a good ENV variable name that has to be set for this to be activated?

janpio:
(rebased on master, now clean changes)

bin/fastlane
joshdholtz: I think its okay to leave the development mode stuff in but remove the ` e.g. RACK_ENV, RAILS_ENV, etc` as these are specific to web frameworks which fastlane is not 😉

bin/fastlane
joshdholtz: I still think we should remove _this_ comment_ because we don't use the `ActiveSupport` gem so this doesn't make sense in our context even though the `bootsnap` documentation has it suggested

bin/fastlane
joshdholtz: Can we get rid of this comment? The `RAILS_ENV` example is not relevant to _fastlane_ at all

joshdholtz (dismissed):
A few more comments/suggestions here! And I think a good environment variable name for this might be something like `FASTLANE_OPTIMIZE_BOOT`??

janpio:
Hm, I just made the necessary changes to add `FASTLANE_OPTIMIZE_BOOT`.
Which of course disables that code path completely. I could add a test that runs `fastlane -v` or `fastlane --help` with that ENV variable set to test that it works generally (Edit: I actually did that in the `benchmark_help_command` lane), but what about the complete test suite? How do we want to handle that? Do we just ignore it for now?

bin/fastlane
```diff
@@ -4,6 +4,31 @@ if RUBY_VERSION < '2.0.0'
   abort("fastlane requires Ruby 2.0.0 or higher")
 end
 
+def self.windows?
+  # taken from: https://stackoverflow.com/a/171011/1945875
+  (/cygwin|mswin|mingw|bccwin|wince|emx/ =~ RUBY_PLATFORM) != nil
+end
+
+if ENV["FASTLANE_OPTIMIZE_BOOT"]
```
KrauseFx: 👍

KrauseFx (approved):
Looks good, let's wait for @joshdholtz to confirm and then we can merge + ship 🚀

joshdholtz (approved):
Woot! This is great! Next week's release is gunna be 🔥

awgeorge:
This may have a negative effect on loading plugins, as enabled the ENV var resulted in this: 

```
[22:23:19]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[22:23:19]: Error loading plugin 'fastlane-plugin-changelog': cannot load such file -- fastlane/plugin/changelog
[22:23:19]: It seems like you wanted to load some plugins, however they couldn't be loaded
[22:23:19]: Please follow the troubleshooting guide: https://docs.fastlane.tools/plugins/plugins-troubleshooting/
+---------------------------+-----------+------------------+
|                       Used plugins                       |
+---------------------------+-----------+------------------+
| Plugin                    | Version   | Action           |
+---------------------------+-----------+------------------+
| fastlane-plugin-changelog | undefined | No actions found |
+---------------------------+-----------+------------------+
```

Disabling the var results in the system working again. Not sure if I'm doing it wrong, but wanted to point it out as it seems `experimental`

janpio:
Thanks @awgeorge - that is an interesting observation. Could could you please create a new issue and mention me in it (also include that this was with fastlane 2.106.0) so we can take a look at this in its own issue. Thanks.

PR #13174, opened 2018-08-21
[deliver, pilot] Look up provider IDs programmatically

mnespor:
🔑

### Checklist

### Motivation and Context
This PR is motivated by issue #10796. When you're on multiple App Store Connect teams, it would be helpful if Pilot and Deliver could detect your iTunes provider short name.

Andrew mentioned four goals in the conversation on #10796:

1. constructing and running this long command from a fastlane action
2. using my existing credentials in the keychain
3. parsing the output into something that can be easily consumed programmatically
4. automatically mapping to developer portal team IDs & iTunes Connect team IDs based on the organization name

This PR implements goals 2, 3, and 4 on that list. I'm not sure about adding the idea of adding a core fastlane action, at least not as part of this PR. However, with the changes to ItunesTransporter described below, it will become possible to make a plugin or custom action that prints out the list of provider names.

I tested this PR by using Pilot to upload an IPA with and without specifying a provider short name.

### Description
This PR modifies three components: Pilot, Deliver, and ItunesTransporter. My goal was to keep Pilot and Deliver completely unchanged for single-team users, but add the power to look up the provider short name programmatically for users who are on more than one App Store Connect team.

#### ItunesTransporter
A new method, `provider_ids`, is added to ItunesTransporter in Fastlane Core. It runs iTMS Transporter with the `-m provider` option and parses the table of provider names returned by Apple. It returns a map of provider short names keyed by team name.

#### Pilot and Deliver
When you're on fewer than two App Store Connect teams, no change.

If you pass an explicit `itc_provider`, no change.

Otherwise, Pilot and Deliver automatically invoke iTMS Transporter with `-m provider`. If one of the provider long names matches the name of the active team in your Tunes session, the corresponding provider short name will be used to upload or download binaries.

Thank you so much for your time!


fastlane_core/spec/itunes_transporter_spec.rb
```diff
@@ -106,22 +141,36 @@ def java_download_command(provider_short_name = nil)
...
+            transporter = FastlaneCore::ItunesTransporter.new('fabric.devtools@gmail.com', "!> p@$s_-+=w'o%rd\"&#*<")
+            expect(transporter.provider_ids).to eq(java_provider_id_command)
+          end
+        end
       end
 
       describe "use_shell_script is true with a itc_provider short name set" do
         describe "upload command generation" do
-          it 'generates a call to java directly' do
+          it 'generates a call to the shell script' do
```
janpio: uh, bug before?

pilot/lib/pilot/options.rb
```diff
@@ -174,7 +174,7 @@ def self.available_options
         # rubocop:disable Metrics/LineLength
         FastlaneCore::ConfigItem.new(key: :itc_provider,
                                      env_name: "PILOT_ITC_PROVIDER",
-                                     description: "The provider short name to be used with the iTMSTransporter to identify your team. To get provider short name run `pathToXcode.app/Contents/Applications/Application\\ Loader.app/Contents/itms/bin/iTMSTransporter -m provider -u 'USERNAME' -p 'PASSWORD' -account_type itunes_connect -v off`. The short names of providers should be listed in the second column",
+                                     description: "The provider short name to be used with the iTMSTransporter to identify your team. This value will override the provider short name automatically detected by Pilot. To get provider short name run `pathToXcode.app/Contents/Applications/Application\\ Loader.app/Contents/itms/bin/iTMSTransporter -m provider -u 'USERNAME' -p 'PASSWORD' -account_type itunes_connect -v off`. The short names of providers should be listed in the second column",
```
janpio: if we add an action, this command does not have to be run manually any more, but we could direct to the new action here, correct?

janpio (commented):
Looks pretty good! Left some comments.

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -99,7 +104,10 @@ def execute(command, hide_output)
         UI.important("Although errors occurred during execution of iTMSTransporter, it returned success status.")
       end
 
-      exit_status.zero?
+      {
+        successful: exit_status.zero?,
+        output: @all_lines
+      }
```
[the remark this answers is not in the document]
joshdholtz: It looks like this is needed in the `provider_ids` function in `ItunesTransporter` to parse out the response to be able to infer the `itc_provider`?

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -99,7 +104,10 @@ def execute(command, hide_output)
         UI.important("Although errors occurred during execution of iTMSTransporter, it returned success status.")
       end
 
-      exit_status.zero?
+      {
```
joshdholtz: @mnespor Instead of changing this return type, could we instead add an optional block that is passed into this `execute` method where you can parse the `@all_lines`? Changin this return type is a bit spooky 😱

pilot/lib/pilot/manager.rb
```diff
@@ -20,8 +20,8 @@ def login
       config[:username] ||= CredentialsManager::AppfileConfig.try_fetch_value(:apple_id)
 
       UI.message("Login to App Store Connect (#{config[:username]})")
-      Spaceship::Tunes.login(config[:username])
-      Spaceship::Tunes.select_team
+      @session = Spaceship::Tunes.login(config[:username])
+      @selected_team_id = Spaceship::Tunes.select_team
```
joshdholtz: Instead of storing off `@session` and `@selected_team_id` here and using these in `build_manager.rb`, you could call `Spaceship::Tunes.client` and `Spaceship::Tunes.client.team_id`

pilot/lib/pilot/build_manager.rb
```diff
@@ -171,6 +171,25 @@ def should_update_app_test_information?(options)
       options[:beta_app_description].to_s.length > 0 || options[:beta_app_feedback_email].to_s.length > 0
     end
 
+    # If itc_provider was explicitly specified, use it.
+    # If there are multiple teams, infer the provider from the selected team name.
+    # If there are fewer than two teams, don't infer the provider.
+    def transporter_for_selected_team(options)
+      generic_transporter = FastlaneCore::ItunesTransporter.new(options[:username], nil, false, options[:itc_provider])
+      return generic_transporter unless options[:itc_provider].nil? && @session.teams.count > 1
```
joshdholtz: Left a note in `pilot/manager.rb` about maybe removing the `@session` and `@selected_team_id` and doing something like this at the beginning of this method to use...
```rb
client = Spaceship::Tunes.client
team_id = client.team_id
```

joshdholtz (dismissed):
Left some comments on things I think could be changed😊 But this is look really good so far 🙌

fastlane_core/lib/fastlane_core/itunes_transporter.rb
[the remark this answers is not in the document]
mnespor: That's accurate—the iTMS Transporter response was previously discarded. I like the idea in Josh's comment further down, where this would send the response out through a block instead of changing the return type.

fastlane_core/spec/itunes_transporter_spec.rb
mnespor: Yes, at least in the test description. The test condition itself wasn't bugged.

fastlane_core/lib/fastlane_core/itunes_transporter.rb
mnespor: Definitely

pilot/lib/pilot/options.rb
mnespor: Yes

mnespor:
Thank you for your time in reviewing this! The spots that had been storing `Spaceship::Tunes` state in ivars have been switched over to `Spaceship::Tunes.client`. The original return type of `TransporterExecutor.execute` has been restored. Finally, I made some small tweaks in the new documentation, replacing "iTunes Transporter" with "deliver" and "pilot."

Once CI passes, I think this is ready for another look.

joshdholtz:
@mnespor Thanks for making those changes! I will spend some time tomorrow and/or Thursday thoroughly testing this since this touches a major part of fastlane 😊

svanzoest:
@mnespor thanks for putting this together. 🚀

janpio (approved):
<3

janpio:
(Failing test on master? What how where why?)

mnespor:
False alarm. spaceship/spec/portal/key_spec.rb turned out to be a Ruby version issue—that test passes under 2.4 on my machine

mnespor:
@joshdholtz, how does the code at the head of mnespor/feature/look-up-provider-ids-programmatically-10796 look? Does it address the concerns you raised?

janpio:
Took care of another merge conflict.

This "just" needs proper testing, right?

mnespor:
Thank you kindly! It’s ready for testing as far as I know—I didn’t have any further functional changes in mind.

PR #13191, opened 2018-08-22
[CI/Appveyor] cleaner way to install iconv on Windows

janpio:
This installs http://gnuwin32.sourceforge.net/packages/libiconv.htm instead of this other, strange package I have been using as it also included `iconv`. Still a download from Sourceforge, but at least a very popular one.


joshdholtz (approved):
Looks good to me 💪

PR #13195, opened 2018-08-22
Use absolute app icon path for deliver upload preview

KrauseFx:
**Note** this is not a good solution, I'm wondering if this works for anyone else? Does the _deliver_ preview show the app icon at all for you? 

Using an absolute path is bad, as it doesn't allow you to zip and share the preview or just move it around

```
[2] pry(#<Deliver::HtmlGenerator>)> @options[:app_icon]
=> "./fastlane/metadata/app_icon.png"
```

### DIFF
--- a/deliver/lib/assets/summary.html.erb
+++ b/deliver/lib/assets/summary.html.erb
@@ -144,13 +144,13 @@
         <% if @options[:app_icon] %>
         <div class="app-icon">
           Large App Icon:<br>
-          <img src="<%= @options[:app_icon] %>">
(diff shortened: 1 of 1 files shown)


KrauseFx:
My assumption is that there is something weird going on, that it behaves differently depending on if you have _fastlane_ setup or not, or if you call _deliver_ via _fastlane_ or directly

KrauseFx:
Whoah maybe related

KrauseFx:
Pls don't merge this

joshdholtz:
@KrauseFx Closing this in favor of #13223 😊

PR #13201, opened 2018-08-23
[deliver] fix double screenshots directory on deliver setup

### LINKED ISSUE #13194: fastlane init (deliver init) seems to create unnecessary folder
screenshot 2018-08-22 06 16 51

So when running `fastlane init` and telling _fastlane_ to manage my app's metadata for a new app that didn't exist before and was created by _fastlane_, it looks like it generates an extra `screenshots` folder inside the actual `screenshots` folder, resulting in the _deliver_ run to fail.

joshdholtz:
Fixes #13194

## Problem
The full screenshots path was getting generated here - https://github.com/fastlane/fastlane/blob/dd0b74f87915dec2666efe938123116b4129cc0c/deliver/lib/deliver/setup.rb#L27

But then `screenshots` was getting appended to the path again here - https://github.com/fastlane/fastlane/blob/dd0b74f87915dec2666efe938123116b4129cc0c/deliver/lib/deliver/setup.rb#L130

## Solution
Removed the second `screenshots` from getting appended in  the `download_screenshots` method

### DIFF
--- a/deliver/lib/deliver/setup.rb
+++ b/deliver/lib/deliver/setup.rb
@@ -126,8 +126,7 @@ module Deliver
       end
     end
 
-    def download_screenshots(deliver_path, options)
-      path = File.join(deliver_path, 'screenshots')
+    def download_screenshots(path, options)
       FileUtils.mkdir_p(path)
       Deliver::DownloadScreenshots.run(options, path)
     end


janpio (approved):

KrauseFx:
This fixes the newly created screenshots folder, however I have the feeling something else is going wrong here, e.g. https://github.com/fastlane/fastlane/pull/13195.

I'm worried something broke recently (e.g. https://github.com/fastlane/fastlane/issues/13176)

joshdholtz:
@KrauseFx Aight, I will look into those too 👌

PR #13202, opened 2018-08-23
[produce] reprompt app name on failed create_app_online setup

### LINKED ISSUE #13192: produce has endless loop when invalid app name is used
screenshot 2018-08-22 06 01 52

This asks me if I want to try again, but then it still auto-fills that username to the invalid one. It should ask for my input again.

joshdholtz:
Fixes #13192

## Problem
If app name was too long or taken, `fastlane init` would fail when creating app on App Store Connect but ask if user wants to try again but without reprompting for app name

## Solution
Move `Produce.config` into loop so the app name prompt happens again

### DIFF
--- a/fastlane/lib/fastlane/setup/setup_ios.rb
+++ b/fastlane/lib/fastlane/setup/setup_ios.rb
@@ -470,14 +470,16 @@ module Fastlane
         produce_options[:skip_devcenter] = true
       end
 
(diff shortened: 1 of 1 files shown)


janpio (approved):

KrauseFx (approved):
Yess, nice fix 👍

PR #13204, opened 2018-08-23
[deliver] fix submission of tvos build on combined ios/tvos apps

tmm1:
### Checklist

### Motivation and Context

I have a combined app on the App Store with both tvos and ios builds under the same bundle. When I try to deliver tvos builds, it fails because the code was only considering ios builds for submission.

### Description

Simply passing through platform so the correct candidate builds are considered.

### DIFF
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -7,7 +7,7 @@ module Deliver
       select_build(options)
 
       UI.message("Submitting the app for review...")
-      submission = app.create_submission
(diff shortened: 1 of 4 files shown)


tmm1:
Ping?

janpio (approved):
This definitely looks good, but I don't know enough to actually merge it. Someone else will have to take a look and best test it locally.

joshdholtz (approved):
Tested this out by adding a tvOS target in my iOS and everything seemed to work 😊 Thanks for making this change! 💯

PR #13211, opened 2018-08-24
[supply] remove unused CredentialsLoader

janpio:
`supply/client` was defining `CredentialsLoader = Google::Auth::CredentialsLoader` which was not used anywhere, so I removed it.


joshdholtz (approved):
Tested this out and seems like everything still works 👍

PR #13214, opened 2018-08-24
[WIP] monkey patch `String.shellescape` and `Array.shelljoin` for Windows support

### LINKED ISSUE #13213: `Shellwords.shellescape` is not cross platform safe
`shellescape` is used ~234 times in the fastlane codebase:
https://github.com/fastlane/fastlane/search?q=shellescape&type=Code

Unfortunately is seems that is not cross platform safe at all :(

https://ruby-doc.org/stdlib-2.5.1/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellescape

> Escapes a string so that it can be safely used in a Bourne shell command line. str can be a non-string object that responds to to_s.
> 

First sentence already tells us that this is for "a Bourne shell", which Windows command lines are certainly not.
Second sentence then tells us that the result should not be quoted - which is necessary on Windows for paths with spaces, not matter it escaped or not.

This is independently confirmed by the great StackOverflow:
https://stackoverflow.com/a/5703599/252627
https://stackoverflow.com/a/50436815/252627

--- 

From previous testing of fastlane on Windows I know that this _mostly_ is not a problem. Most of the strings that `shellescape` is applied on do not contain spaces, so no quotes are necessary on Windows. The rest it does, does not hurt at least.

But there are cases where 

a) there could be spaces in the strings (see https://github.com/fastlane/fastlane/issues/13212) 
or 
b) there are single quotes applied to the resulting string before usage.

https://github.com/fastlane/fastlane/blob/6a658a4cf9c56c3c838aed397166c2770009b2d2/fastlane_core/lib/fastlane_core/itunes_transporter.rb#L213-L214

---

What can we do about that?

janpio:
### Checklist

### Description
This also required adapting some tests that had hardcoded strings as the expected command of any `sh` calls - if they include parameters with values containing spaces or empty strings they had to be adapted.

TODO
- [x] Split `CrossplatformShellwords` into more appropriate place
- [x] Write tests for `CrossplatformShellwords` methods
- [ ] How to also monkey patch `Shellwords`?
- [ ] Write tests that confirm that `Shellwords`, `String.shellescape` and `Array.shelljoin` now use the `CrossplatformShellwords` implementation

Open Questions
- [ ] Is what `configuration_spec.rb` now produces actually what we want?

closes https://github.com/fastlane/fastlane/issues/13213


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lib/fastlane/action.rb
```diff
@@ -187,4 +187,20 @@ def dedent!
 
     self.gsub!(/^#{first_line_indent}/, "")
   end
+
+  def shellescape
+    if FastlaneCore::Helper.windows?
+      # double quotes have to be doubled
+      self.gsub('"', '""')
```
revolter: I think this does nothing, as it returns a copy of the string after the substitutions. You have to use `gsub!` to replace it in-place.

fastlane/lib/fastlane/action.rb
```diff
@@ -187,4 +187,20 @@ def dedent!
...
+      # wrap in double quotes if contains space
+      # then return (and skip Shellwords.escape)
+      if self =~ /\s/
+        '"' + self + '"'
```
revolter: Please use `return` so it's clearer

PR #13216, opened 2018-08-24
[Pilot] Print error when failing to import tester

JaviSoto:
### Checklist

### Motivation and Context
Right now this fails silently. I was hitting errors due to being rate-limited and I had no way of knowing what was happening.
 

### Description
I tested this by running it locally on my machine, where I was getting rate-limit errors.

### DIFF
--- a/pilot/lib/pilot/tester_importer.rb
+++ b/pilot/lib/pilot/tester_importer.rb
@@ -42,8 +42,8 @@ module Pilot
         begin
           tester_manager.add_tester(config)
           imported_tester_count += 1
-        rescue
-          # do nothing, move on to the next row
+        rescue  => exception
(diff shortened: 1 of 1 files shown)


janpio:
Good idea. Can you think of any use cases where having output errors is not beneficial?

Also please take a look at the failing linter: https://circleci.com/gh/fastlane/fastlane/22921

JaviSoto:
I will fix the linter error.

I can't think of anything. The user has run a command that is supposed to import a list of testers. They expect it to work for all of them, or be able to find out when it failed for some of them, instead of failing silently.

joshdholtz (approved):
I like this change! We should rarely ever eat an error and do nothing with it. Thanks for making this! ✈️ 🛩

PR #13230, opened 2018-08-28
[core] Extend Shellwords (`String.shellescape`, `Array.shelljoin`) for Windows support

### LINKED ISSUE #13213: `Shellwords.shellescape` is not cross platform safe

janpio:
### Checklist

### Motivation and Context

[`Shellwords.shellescape` is not cross platform safe](https://github.com/fastlane/fastlane/issues/13213). This causes problems for commands that are to be used on the shell and contain e.g. paths with spaces or other non-[A-Z0-9] characters.

### Description

This PR monkey patches the `String.shellescape` and `Array.shelljoin` methods that `Shellwords` usually supplies. The patch uses a new module `CrossplatformShellwords` that decides if the normal `Shellwords` or a new `WindowsShellwords` should be used to call the methods.

The PR fixes all the tests that started failing on Windows after the monkey patch became active: all the tests that had hardcoded expected command strings that were compared to command strings generated by the actions (mostly via `Actions.sh` internally) via `var.shellescape`. 

The PR also adds tests for the monkey patched methods, both `WindowsShellwords` and the `*.shell*` methods (in a mock Windows and non-Windows environment) are tested.

#### Limitations

`WindowsShellwords` itself is a pretty basic implementation of mainly `shellescape`. I don't expect this version to cover all use cases and edge cases. Those will be discovered over time when fastlane actions will fail because of malformed commands that do not run on Windows as expected - but it is a starting point.

Tests for actions and methods that compare a hardcoded string to a manually built command string could **not** be identified automatically during this PR. I hope there aren't to many, but it will only affect usage on Windows anyway. macOS is not affected.

This PR also could do nothing to actually _test_ the generated commands, as during test execution all the commands are just returned from `Actions.sh` and not executed. Those tests will have to happen in actual usage of the actions.

--- 
closes #13213

### DIFF
--- a/fastlane/lib/fastlane/actions/add_git_tag.rb
+++ b/fastlane/lib/fastlane/actions/add_git_tag.rb
@@ -14,7 +14,7 @@ module Fastlane
         cmd << ["-am #{message.shellescape}"]
         cmd << '--force' if options[:force]
         cmd << '-s' if options[:sign]
-        cmd << "'#{tag}'"
+        cmd << tag.shellescape
         cmd << options[:commit].to_s if options[:commit]
 
         UI.message("Adding git tag '#{tag}' 🎯.")
--- a/fastlane/spec/actions_specs/add_git_tag_spec.rb
+++ b/fastlane/spec/actions_specs/add_git_tag_spec.rb
@@ -14,7 +14,9 @@ describe Fastlane do
           add_git_tag
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am builds/test/1337\\ \\(fastlane\\) \'builds/test/1337\'")
+        message = "builds/test/1337 (fastlane)"
+        tag = "builds/test/1337"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
       end
 
       it "allows you to specify grouping and build number" do
@@ -28,7 +30,9 @@ describe Fastlane do
           })
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am #{grouping}/test/#{specified_build_number}\\ \\(fastlane\\) \'#{grouping}/test/#{specified_build_number}\'")
+        message = "#{grouping}/test/#{specified_build_number} (fastlane)"
+        tag = "#{grouping}/test/#{specified_build_number}"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
       end
 
       it "allows you to specify a prefix" do
@@ -40,7 +44,9 @@ describe Fastlane do
           })
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am builds/test/#{prefix}#{build_number}\\ \\(fastlane\\) \'builds/test/#{prefix}#{build_number}\'")
+        message = "builds/test/#{prefix}#{build_number} (fastlane)"
+        tag = "builds/test/#{prefix}#{build_number}"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
       end
 
       it "allows you to specify a postfix" do
@@ -52,7 +58,9 @@ describe Fastlane do
           })
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am builds/test/#{build_number}#{postfix}\\ \\(fastlane\\) \'builds/test/#{build_number}#{postfix}\'")
+        message = "builds/test/#{build_number}#{postfix} (fastlane)"
+        tag = "builds/test/#{build_number}#{postfix}"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
       end
 
       it "allows you to specify your own tag" do
@@ -64,7 +72,8 @@ describe Fastlane do
           })
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am #{tag}\\ \\(fastlane\\) \'#{tag}\'")
+        message = "#{tag} (fastlane)"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
       end
 
       it "specified tag overrides generate tag" do
@@ -79,7 +88,8 @@ describe Fastlane do
           })
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am #{tag}\\ \\(fastlane\\) \'#{tag}\'")
+        message = "#{tag} (fastlane)"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
       end
 
       it "allows you to specify your own message" do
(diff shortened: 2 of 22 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lib/fastlane/actions/add_git_tag.rb
```diff
@@ -14,7 +14,7 @@ def self.run(options)
         cmd << ["-am #{message.shellescape}"]
         cmd << '--force' if options[:force]
         cmd << '-s' if options[:sign]
-        cmd << "'#{tag}'"
+        cmd << tag.shellescape
```
janpio: this changes the behavior a bit as it doesn't wrap the param in `'` any more, but as the tag itself is now `shellescape`d this should not cause any problem.

fastlane/spec/actions_specs/add_git_tag_spec.rb
```diff
@@ -14,7 +14,9 @@
           add_git_tag
         end").runner.execute(:test)
 
-        expect(result).to eq("git tag -am builds/test/1337\\ \\(fastlane\\) \'builds/test/1337\'")
+        message = "builds/test/1337 (fastlane)"
+        tag = "builds/test/1337"
+        expect(result).to eq("git tag -am #{message.shellescape} #{tag.shellescape}")
```
janpio: matching the previous comment, `'` are gone here as well and replaced with `shellescape`

fastlane_core/spec/core_ext/shellwords_ext_spec.rb
```diff
@@ -0,0 +1,372 @@
+# references
```
joshdholtz: Should this file actually be called `shellwords_spec.rb` to match with the new `shellwords.rb`?

fastlane_core/lib/fastlane_core/core_ext/shellwords.rb
```diff
@@ -0,0 +1,60 @@
+require_relative '../helper'
+require 'shellwords'
+
+# Here be monkey patches
+
+class String
+  # CrossplatformShellwords
+  def shellescape
```
joshdholtz: Does this 100% override the method defined by `shellwords` because this is required after `shellwords`?

fastlane_core/lib/fastlane_core/core_ext/shellwords.rb
```diff
@@ -0,0 +1,60 @@
...
+    CrossplatformShellwords.shellescape(self)
+  end
+end
+
+class Array
+  def shelljoin
```
joshdholtz: Same question here - Does this 100% override the method defined by `shellwords` because this is required after `shellwords`?

joshdholtz (commented):
Few small questions! Otherwise this looks 💯

fastlane_core/spec/core_ext/shellwords_ext_spec.rb
janpio: I chose the name after the other file `\fastlane_core\spec\core_ext\cfpropertylist_ext_spec.rb` which contains the tests for an implementation in `\fastlane_core\lib\fastlane_core\core_ext\cfpropertylist.rb`. But the other one would work as well, of course.

fastlane_core/lib/fastlane_core/core_ext/shellwords.rb
janpio: I am not sure what you are asking, but I mostly did this by how the other stuff was done - so possible I missed something.

My current understanding:
1. `fastlane_core/lib/fastlane_core.rb` requires `fastlane_core/core_ext/shellwords` which loads it from `/lib/fastlane_core/core_ext/shellwords.rb`.
2. There we require 'shellwords, which loads the native implementation.
3. Then the monkey patching happens.

janpio: If your question is if this overwrites the `String.shellescape` defined by `Shellwords` completely, then: Yes. It completely deletes that method and replaces it with this code - which for non-Windows environments does the exact same thing as shellword's `String.shellescape` - handing the string over to `Shellwords.escape`.

joshdholtz: Ah yes, your assumption of what I was asking in your last response was correct 👌 Thanks for the answer! ❤️

joshdholtz (dismissed):
Two small requests about adding assertion that `Shellwords.escape` is actually getting called 🙃

fastlane_core/spec/core_ext/shellwords_ext_spec.rb
```diff
@@ -0,0 +1,372 @@
...
+  end
+
+  describe "on other OSs (macOS, Linux)" do
+    before(:each) do
+      allow(FastlaneCore::Helper).to receive(:windows?).and_return(false)
+    end
+
+    os = 'other'
+    shelljoin_testcases.each do |testcase|
+      it testcase['it'] + ': ' + testcase['it_result'][os] do
```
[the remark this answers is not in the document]
janpio: On non-Windows...

Yes, good idea. Will add this.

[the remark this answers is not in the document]
janpio: Turned out to be a bit of work, especially as I had to learn about `.and_call_original` first 🍇 

But well, not it checks if the correct methods of the correct classes are called.

fastlane_core/spec/core_ext/shellwords_ext_spec.rb
```diff
@@ -0,0 +1,372 @@
...
+  end
+
+  describe "on other OSs (macOS, Linux)" do
+    before(:each) do
+      allow(FastlaneCore::Helper).to receive(:windows?).and_return(false)
+    end
+
+    os = 'other'
+    shellescape_testcases.each do |testcase|
+      it testcase['it'] + ': ' + testcase['it_result'][os] do
```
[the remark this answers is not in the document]
janpio: Done, see new commit.

joshdholtz (approved):
@janpio This looks good 👍 Thanks for making those testing changes and answering my questions ❤️

PR #13250, opened 2018-08-29
[produce] removed log with a tag of iTC

RishabhTayal:
Removed itc

### DIFF
--- a/produce/lib/produce/itunes_connect.rb
+++ b/produce/lib/produce/itunes_connect.rb
@@ -18,7 +18,7 @@ module Produce
     def create_new_app
       application = fetch_application
       if application
-        UI.success("[ASC] App '#{Produce.config[:app_identifier]}' already exists (#{application.apple_id}), nothing to do on App Store Connect")
+        UI.success("App '#{Produce.config[:app_identifier]}' already exists (#{application.apple_id}), nothing to do on App Store Connect")
         # Nothing to do here
       else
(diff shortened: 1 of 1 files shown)


produce/lib/produce/itunes_connect.rb
```diff
@@ -18,7 +18,7 @@ def run
     def create_new_app
       application = fetch_application
       if application
-        UI.success("[iTC] App '#{Produce.config[:app_identifier]}' already exists (#{application.apple_id}), nothing to do on App Store Connect")
+        UI.success("[ASC] App '#{Produce.config[:app_identifier]}' already exists (#{application.apple_id}), nothing to do on App Store Connect")
```
janpio: As `ASC` is not really a common acronym, I think this should just be removed instead of replaced.

RishabhTayal: Maybe it will be soon. 🤷‍♂️

joshdholtz: I agree with @janpio here. I would probably prefer to move it since "asc" is more known for meaning "ascending" which is actually what I thought this PR was about based on the title 🙃

joshdholtz (changes requested):
Remove `[iTC]` (`[ASC]`) since this is actually the only one 🤔

RishabhTayal:
Removed!

joshdholtz (approved):
👌

PR #13255, opened 2018-08-30
[spaceship] also catch and handle "Gateway Timeout"

### LINKED ISSUE #13241: download_dsyms does not retry errors
### New Issue Checklist

### Issue Description

Getting lots of errors during download_dsyms recently.

<details>
  <pre>Spaceship::UnexpectedResponse: [!] <HTML><HEAD>
<TITLE>Gateway Timeout - In read </TITLE>
</HEAD><BODY>
<H1>Gateway Timeout</H1>
The proxy server did not receive a timely response from the upstream server.<P>
Reference&#32;&#35;1&#46;54273017&#46;1535504570&#46;2614afa
</BODY></HTML>
  /Users/fancybox/source/channels/vendor/gems/ruby/2.4.0/gems/fastlane-2.102.0/spaceship/lib/spaceship/client.rb:602:in `parse_response'
  /Users/fancybox/source/channels/vendor/gems/ruby/2.4.0/gems/fastlane-2.102.0/spaceship/lib/spaceship/tunes/tunes_client.rb:256:in `applications'
  /Users/fancybox/source/channels/vendor/gems/ruby/2.4.0/gems/fastlane-2.102.0/spaceship/lib/spaceship/tunes/application.rb:67:in `all'
  /Users/fancybox/source/channels/vendor/gems/ruby/2.4.0/gems/fastlane-2.102.0/spaceship/lib/spaceship/tunes/application.rb:73:in `find'
  /Users/fancybox/source/channels/vendor/gems/ruby/2.4.0/gems/fastlane-2.102.0/fastlane/lib/fastlane/actions/download_dsyms.rb:19:in `run'
</pre>
</details>

tmm1:
### Checklist

### Motivation and Context
fixes #13241

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -50,6 +50,7 @@ module Spaceship
     UnexpectedResponse = Spaceship::UnexpectedResponse
     AppleTimeoutError = Spaceship::AppleTimeoutError
     UnauthorizedAccessError = Spaceship::UnauthorizedAccessError
+    GatewayTimeoutError = Spaceship::GatewayTimeoutError
     InternalServerError = Spaceship::InternalServerError
 
     # Authenticates with Apple's web services. This method has to be called once
@@ -532,6 +533,7 @@ module Spaceship
         Faraday::Error::TimeoutError,
         Faraday::ParsingError, # <h2>Internal Server Error</h2> with content type json
         AppleTimeoutError,
+        GatewayTimeoutError,
         InternalServerError => ex # New Faraday version: Faraday::TimeoutError => ex
       tries -= 1
       unless tries.zero?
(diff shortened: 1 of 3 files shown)


janpio (approved):

joshdholtz (changes requested):
Can we have this actually raise a new error called `GatewayTimeoutError`? 😊 A gateway error and internal service error will get handled the same way but it feels a bit dangerous to have the gateway error map to the `InternalServerError` and have the same error message as the `InternalServerError`. 

I think it would be ideal if we could get this mapped to a new error 💪

tmm1:
I don't understand what's dangerous about it? They're both the same "in read" error.

joshdholtz:
@tmm1 A gateway timeout error (504) is inherently different from an internal server error (500) where an 500 happens at an application level and a 504 happens at some network level.

I would suggest adding a new conditional with something like...
```rb
raise GatewayTimeoutInReadError, "Received an \"Gateway Timeout - In Read\" from App Store Connect / Developer Portal, please try again later"
```
as this error message will get printed out to the user. We do not want to map a "Gateway Timeout" as an "Internal Server Error".

To then account for this in the retry, we would add `GatewayTimeoutInRead` error to this rescue so that it gets retried - https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L530-L535

This should have the same end result as your initial fix but this will present the proper error message to the end user if the retries are not successful

tmm1:
Okay I can make those changes this week. Feel free to commit on this PR directly if you would like.

joshdholtz:
@tmm1 Can you give me permission to push to this branch? There should be some permission option on the right side of the page I believe 🤔 Thanks!

tmm1:
It's enabled already. Are you not able to push to my branch?

tmm1:
I guess maybe it only works for "Maintainers"?

joshdholtz:
@tmm1 Whoops! Was pushing to `apple_errors` instead of `apple-errors` 😝

joshdholtz:
@tmm1 @janpio Pushed some changes here! Added new `GatewayTimeoutError` and also added a bunch of tests around these parts that should have existed before but didn't 😇

tmm1:
👍🏼 LGTN

joshdholtz (approved):

PR #13290, opened 2018-09-05
FL_HOCKEY_TIMEOUT fix for "undefined method `zero?'"

jmihalicza:
### Checklist

### Motivation and Context
There was no type specified for FL_HOCKEY_TIMEOUT, and therefore it was passed to faraday as string, causing: undefined method `zero?'

### Description
I set the config item type to Integer.
I did not actually test it, but given that the original version leads to undefined method, that path of execution can only become better with the change.


fastlane/lib/fastlane/actions/hockey.rb
```diff
@@ -323,7 +323,8 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :timeout,
                                       env_name: "FL_HOCKEY_TIMEOUT",
                                       description: "Request timeout in seconds",
-                                      is_string: false,
+                                      type: Integer,
+                                      default_value: 0,
```
janpio: 0 is zero or endless in this case? Might want to add a comment to specify.

jmihalicza:
I signed it!

PR #13299, opened 2018-09-05
Add an option to overwrite zip file

powerivq:
### Checklist

### Motivation and Context
By default `zip -r something.zip *` appends all the files into if the zip file already exists. This might cause unexpected behavior.

### Description
This PR adds an option to disable appending.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -279,14 +279,20 @@ module FastlaneCore
     end
 
     # Zips directory
-    def self.zip_directory(path, output_path, contents_only: false, print: true)
+    def self.zip_directory(path, output_path, contents_only: false, overwrite: false, print: true)
+      if overwrite
+        overwrite_command = " && rm -f '#{output_path}'"
+      else
+        overwrite_command = ""
+      end
+
       if contents_only
-        command = "cd '#{path}' && zip -r '#{output_path}' *"
+        command = "cd '#{path}'#{overwrite_command} && zip -r '#{output_path}' *"
       else
         containing_path = File.expand_path("..", path)
         contents_path = File.basename(path)
 
-        command = "cd '#{containing_path}' && zip -r '#{output_path}' '#{contents_path}'"
+        command = "cd '#{containing_path}'#{overwrite_command} && zip -r '#{output_path}' '#{contents_path}'"
       end
 
       UI.command(command) unless print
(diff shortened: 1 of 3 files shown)


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -279,14 +279,20 @@ def self.strip_ansi_colors(str)
     end
 
     # Zips directory
-    def self.zip_directory(path, output_path, contents_only: false, print: true)
+    def self.zip_directory(path, output_path, contents_only: false, overwrite: false, print: true)
+      if overwrite
+        overwrite_command = " && rm -f '#{output_path}'"
```
joshdholtz: I don't know if this is more or less efficient/cleaner but the `zip -FS` would also do this same thing I think?
```
       -FS
       --filesync
              Synchronize the contents of an archive with the files on the OS.  Normally when an archive is updated, new files are added and changed files are updated but files that no longer exist on the OS are not deleted from the archive.  This option enables a new mode that checks entries in the archive against the file
              system.  If the file time and file size of the entry matches that of the OS file, the entry is copied from the old archive instead of being read from the file system and compressed.  If the OS file has changed, the entry is read and compressed as usual.  If the entry in the archive does not match a file on the
              OS, the entry is deleted.  Enabling this option should create archives that are the same as new archives, but since existing entries are copied instead of compressed, updating an existing archive with -FS can be much faster than creating a new archive.  Also consider using -u for updating an archive.

              For this option to work, the archive should be updated from the same directory it was created in so the relative paths match.  If few files are being copied from the old archive, it may be faster to create a new archive instead.

              Note that the timezone environment variable TZ should be set according to the local timezone in order for this option to work correctly.  A change in timezone since the original archive was created could result in no times matching and recompression of all files.

              This  option deletes files from the archive.  If you need to preserve the original archive, make a copy of the archive first or use the --out option to output the updated archive to a new file.  Even though it may be slower, creating a new archive with a new archive name is safer, avoids mismatches between ar-
              chive and OS paths, and is preferred.
```

I'm cool with either of these but not sure if one would be preferred over the other 😊 Thoughts?

joshdholtz (commented):
One small question!

fastlane_core/lib/fastlane_core/helper.rb
joshdholtz: Sounds good! Just wanted to put these here because I was also curious why zip didn't haven't an override argument 🙃

powerivq: @joshdholtz cool. can you approve this?

joshdholtz (approved):
Looks good! 💯

PR #13309, opened 2018-09-06
[snapshot] fix default configuration was not set to Release

yuya-oc:
### Checklist

### Motivation and Context

When building the app in snapshot action, "Debug" configuration was used unlike [the documentation](https://docs.fastlane.tools/actions/snapshot/#parameters).
> The configuration to use when building the app. Defaults to 'Release'

This PR sets the default value to "Release".

### Description

Set `default_value: "Release"` instead of `default_value_dynamic: true` in `snapshot/lib/snapshot/options.rb`

I tested this by following way.

1. Create a new iOS project from "Single View App" template.
2. Set up fastlane for snapshot.
3. Add following block into the test case.
```swift
#if DEBUG
    snapshot("01-first-screen-debug")
#else
    snapshot("01-first-screen-release")
#endif
```
4. Execute snapshot action.
5. Check `01-first-screen-release` screenshot is located in the screenshot directory.
6. Check `configuration` can be changed via Snapfile or environment variable, `SNAPSHOT_CONFIGURATION`.


joshdholtz (changes requested):
I am torn about this change 🤔 Even though the description says `Defaults to 'Release'`, I'm not sure if we actually want to hard code `default_value: "Release"`

The reason being not every app will necessarily have a "Release" configuration. So putting release here will break backwards compatibility for users that don't explicitly specify this option. This means that it would break existing implementations on their next run.

I think the proper solution here would be to either update the description to remove `Defaults to 'Release'` or to look for the existence of the `Release` configuration in `scan/lib/scan/detect_values.rb` and default it there (which is probably why `default_value_dynamic: true` was set.

Lemme know your thoughts 😊

yuya-oc:
Thanks for feedback. Yeah, I had a question about that there is `default_value_dynamic: true` without an explicit variable.

Surely "Release" might not be there. However Scan doc also says configuration defaults to Release, so I feel the documentation should be correct. And I think Snapshots' main feature is for testing and App Store submission, so the app should be built for Release configuration in normal use cases.

Because of those reason, I feel that the second option `detect_values.rb` would make sense.

joshdholtz:
@yuya-oc Documentation can sometimes be wrong 😝 Even though it says it will default to "Release" it still shouldn't force a user to have a "Release" scheme for the default to work as the fastlane run will go 💥 if its not there. 

I think the best thing would be rework this PR to inspect the Xcode project for the "Release" configuration and use that if there is one otherwise don't. That way the documentation is still correct 😉

yuya-oc:
@joshdholtz Yeah, I agree. I'll try implementing that.

yuya-oc:
@joshdholtz Updated the branch. Now when the Xcode project has "Release" and the configuration is not specified, snapshot would use "Release" as the default configuration.

janpio:
Great PR @yuya-oc! 

2 thoughts:

- What happens if `Release` is not present? What is run then?
- Would it make sense to update the option documentation to `Defaults to 'Release' if it is present.`?

janpio:
Meaning what exactly? Should we close this PR or update it to point to that option somehow?

yuya-oc:
Sorry, it was not obvious. First, I close this PR because this will make breaking change: https://github.com/fastlane/fastlane/pull/13309#pullrequestreview-154178725. If "Release" is needed for users, they can explicitly specify "Release" in their config.

PR #13322, opened 2018-09-07
FL_HOCKEY_TIMEOUT fix for "undefined method `zero?'"

jmihalicza:
### Checklist

### Motivation and Context
There was no type specified for `FL_HOCKEY_TIMEOUT`, and therefore it was passed to _faraday_ as string, causing: ``undefined method `zero?'``

### Description
I set the config item type to `Integer`.
I did not actually test it, but given that the original version leads to undefined method, that path of execution can only become better with the change.


jmihalicza:
@janpio I recreated https://github.com/fastlane/fastlane/pull/13290 in this new PR.

joshdholtz (approved):
This checks out ✅ Thanks for making this fix! ❤️

Issue #10598, opened 2017-10-14
Is there a way to add xcpretty additional arguments in Scan?


Nmbr-13:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I am trying to attach screenshots to my html-reports. Because I use KIF framework for my UI-tests, so I cant use Snapshot. Thats why I decided to try **`--screenshots`** xcpretty-argument. 
 
But here's the problem:  I can't add this argument via **`xargs`** parameter, as it will fill it between other xcodebuild-argument, though I need it right after **`| xcpretty -r html`**

So, for now I didn't find the way to fix it inside Fastfile or Scanfile, and the only way to run tests with manually modified command, like:
**`set -o pipefail && env NSUnbufferedIO=YES xcodebuild -workspace ./MyApp.xcworkspace -scheme KIFUITests -destination 'platform=iOS Simulator,name=iPhone 7 Plus,OS=10.3.1' -derivedDataPath 'myBuild' -only-testing:KIFUITests/TestSuite01 test-without-building| xcpretty -r html --screenshots -r junit`**.

It works for me this way, but it's inconvenient, obviously, and the idea of Fastlane is being lost. 

I found out that Snapshot has parameter **`xcpretty-args`**. So, maybe there is some analogical, "pretty" way to solve this task?

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.12.6                                     |
| Ruby                        | 2.2.4                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.14.1                          |
| Installation Source         | ~/.fastlane/bin/bundle/bin/fastlane         |
| Host                        | Mac OS X 10.12.6 (16G29)                    |
| Ruby Lib Dir                | ~/.fastlane/bin/bundle/lib                  |
| OpenSSL Version             | OpenSSL 1.0.2g  1 Mar 2016                  |
| Is contained                | true                                        |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
[...]

Issue #1946, opened 2015-10-23
Turn logging off, silent mode?


fastlane-bot-helper:
<table>
      <tr>
        <td>
          <img src='https://avatars0.githubusercontent.com/u/51020?v=3&s=70' width='35'>
        </td>
        <td>
          Original issue by @jshier - Imported from <a href='https://github.com/fastlane/scan/issues/10'>fastlane/scan#10</a>
        </td>
      </tr>
    </table>

I'd like the option to turn logging off entirely, as well as most of the command line output. I don't need to see what files are building or the other information most of the time, just the warnings or errors.

PR #13324, opened 2018-09-07
[scan] add option to pass xcpretty options

jspargo:
### Checklist

### Motivation and Context
With a large number of tests, the `xcpretty` output, although pretty, is very verbose. [gym](https://docs.fastlane.tools/actions/gym/#parameters) has a `silent` option, but `scan` doesn't; it simply pipes the output to `xcpretty`. `xcpretty` has a number of it's [own format options](https://github.com/supermarin/xcpretty#formats) for how it outputs to the console - this PR allows us to pass these directly into the `scan` action. There is already an `output_types` option that interacts with the [reporter option](https://github.com/supermarin/xcpretty#reporters) of `xcpretty` so I've piggybacked off the great work that's already there and tried to follow that pattern as much as possible.
This is also a solution to:
https://github.com/fastlane/fastlane/issues/1946 and possibly https://github.com/fastlane/fastlane/issues/10598 (though I realise screenshots isn’t currently covered by this PR as this is for the console format only. But it could be?)

### Description
Add `xcpretty_args` option to `scan` (optional). ~~This accepts one of:~~
- ~~simple (default)~~
- ~~test (shorter rspec format)~~
- ~~tap (TAP format)~~
- ~~knock (Knock format)~~

This _now_ accepts any string so that command line arguments for `xcpretty` can be passed in directly (e.g. `xcpretty_args: "--tap --no-utf"`)


jspargo:
I'll mark the documentation task as done, as it looks like it's handled automatically in `fastlane/lib/assets/ActionDetails.md.erb` - please let me know if it needs to be updated elsewhere. First PR to fastlane! 😃 

@googlebot signed

scan/lib/scan/options.rb
```diff
@@ -151,6 +151,11 @@ def self.available_options
                                      env_name: "SCAN_FORMATTER",
                                      description: "A custom xcpretty formatter to use",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :xcpretty_format,
```
joshdholtz: The PR description mentions that the `xcpretty_args` option was added but this is called `xcpretty_format`. Can we actually change this to `xcpretty_args`?

And then probably change other things to match this? 😊

scan/lib/scan/options.rb
```diff
@@ -151,6 +151,11 @@ def self.available_options
...
+                                     env_name: "XCPRETTY-FORMAT",
```
joshdholtz: This should use an underscore instead of a dash. But can we also change this to `XCPRETTY_ARGS`?

scan/lib/scan/xcpretty_reporter_options_generator.rb
```diff
@@ -3,25 +3,29 @@
 module Scan
   class XCPrettyReporterOptionsGenerator
     SUPPORTED_REPORT_TYPES = %w(html junit json-compilation-database)
+    SUPPORTED_XCARGS_TYPES = %w(simple test tap knock)
```
joshdholtz: We allow other xcpretty formatters to be used with the `:format` option. Do other formatters allow other potential arg types? I don't want to limit this if other options are also usable 😊

scan/spec/xcpretty_reporter_options_generator_spec.rb
```diff
@@ -224,6 +224,44 @@
         end
       end
 
+      describe "xcpretty format options generation" do
+        it "generates options to use xcpretty's `test` format option" do
+          generator = Scan::XCPrettyReporterOptionsGenerator.new(false, "html", "report.html", "/test_output", false, "test")
+          reporter_options = generator.generate_xcpretty_format_options
+
+          puts(reporter_options)
```
joshdholtz: Delete this `puts`?

joshdholtz (changes requested):
Left a few comments and requests 😊 This is looking good though! 💪

scan/spec/xcpretty_reporter_options_generator_spec.rb
jspargo: 😱 oops. Done!

scan/lib/scan/xcpretty_reporter_options_generator.rb
jspargo: One more thing to add: for the format options, one of `simple`, `test`, `tap`, or `knock` can be used at a time, but these can be used together with `--report`, `--color`, etc. in the same execution.

scan/lib/scan/xcpretty_reporter_options_generator.rb
```diff
@@ -37,6 +41,12 @@ def initialize(open_report, output_types, output_files, output_directory, use_cl
       (@output_types - SUPPORTED_REPORT_TYPES).each do |type|
         UI.error("Couldn't find reporter '#{type}', available #{SUPPORTED_REPORT_TYPES.join(', ')}")
       end
+
+      unless @xcpretty_args.nil?
+        (@xcpretty_args - SUPPORTED_XCARGS_TYPES).each do |type|
```
joshdholtz: I'm not sure if we actually want to validate these options, TBH 🤔 Validating args like adds a lot of extra support/maintenance when `xcpretty` either adds or removes arguments in the future (see my other comments attached to this review)

scan/lib/scan/xcpretty_reporter_options_generator.rb
```diff
@@ -64,6 +74,18 @@ def generate_reporter_options
       return reporter
     end
 
+    def generate_xcpretty_args_options
+      xcargs = []
+      unless @xcpretty_args.nil?
+        valid_args = @xcpretty_args & SUPPORTED_XCARGS_TYPES
+        valid_args.each do |raw_arg|
+          arg = raw_arg.strip
+          xcargs << "--#{arg}"
```
joshdholtz: Going along with my comment of validating arguments, maybe we shouldn't even auto format with `--`. We could just allow `xcpretty_args` to be a string that just gets appended onto the command. This will allow for any future arguments that accept parameters.

Example:
```rb
xcpretty_args: "--tap --no-utf"
```

Thoughts?

joshdholtz (commented):
Had a view more thoughts on this addition. Maybe we open this up and validate less and simply make this a string that gets appended onto the `xcpretty` command which would make less maintenance later on if/when things change on `xcpretty`'s side along with more flexability for our users 🙃

scan/lib/scan/xcpretty_reporter_options_generator.rb
jspargo: Yeh, I can see why that makes sense. I suppose I have two minor concerns with this:
1) that it will be used as a first option in place of `formatter`, or even `output_types` in error, which for obvious reasons have better logic and error handling behind them
2) If invalid args are passed, the output isn't as helpful - you just get a stack trace like below, and scan continues without logging anything to xcpretty:
```
/Users/jackspargo/.rvm/gems/ruby-2.4.1/gems/xcpretty-0.3.0/bin/xcpretty:73:in `block in <top (required)>': invalid option: --invalid (OptionParser::InvalidOption)
```

That being said, I've made these changes in this PR - see what you think. I suppose there is a third option here to simple allow any string to be passed in the array and just remove any `SUPPORTED_XCARGS_TYPES` reference. I'd be happy with either solution though to be honest, though I think I'm leaning towards your suggestion in the comment above (that this PR now matches) 😄

jspargo:
Hi @joshdholtz - thought I'd just bump this as I realised that I only left comments under your PR comments so I don't know whether this will have appeared in your notifications.

joshdholtz:
@jspargo Ah, thanks for the bump 😊

scan/lib/scan/options.rb
```diff
@@ -151,6 +151,11 @@ def self.available_options
                                      env_name: "SCAN_FORMATTER",
                                      description: "A custom xcpretty formatter to use",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :xcpretty_args,
+                                     env_name: "XCPRETTY_ARGS",
```
joshdholtz: This should be `SCAN_XCPRETTY_ARGS` 🙃

scan/lib/scan/runner.rb
```diff
@@ -122,9 +122,10 @@ def test_results
       # We'll have to regenerate from the xcodebuild log, like we did before version 2.34.0.
       UI.message("Generating test results. This may take a while for large projects.")
 
-      reporter_options_generator = XCPrettyReporterOptionsGenerator.new(false, [], [], "", false)
+      reporter_options_generator = XCPrettyReporterOptionsGenerator.new(false, [], [], "", false, "")
```
joshdholtz: Should this be a `nil` instead of an empty string? The tests use a `nil` when not passing anything but this uses any empty string 🤔 Maybe we also add a tests for `""` so we know that works as expect too? I'm okay with either of these solutions 😊

joshdholtz (changes requested):
Two VERY small comments otherwise this works 💯

scan/lib/scan/runner.rb
jspargo: Good point. I've made this change and added one spec for nil, and updated 4 of the existing 10 `XCPrettyReporterOptionsGenerator` specs  to use `""` instead of `nil` to validate that it doesn't impact the other reporter options - I hope that's OK and not considered corner-cutting! Happy to go the long way round if you think it's better. 😊

jspargo:
@joshdholtz thanks! Comments addressed - let me know what you think 😄

joshdholtz (approved):
Looks great!! Thanks for making all those changes I requested 😊 ❤️

PR #13333, opened 2018-09-10
Report fastlane.swift usage vs normal fastlane (ruby) usage

taquitos:
We want to understand how many folks are using fastlane.swift, so we need to report execution of fastlane.swift vs normal fastlane.
Also includes a tiny bit of cleanup.

### Checklist

### DIFF
--- a/fastlane_core/lib/fastlane_core/analytics/action_launch_context.rb
+++ b/fastlane_core/lib/fastlane_core/analytics/action_launch_context.rb
@@ -8,22 +8,22 @@ module FastlaneCore
     attr_accessor :action_name
     attr_accessor :p_hash
     attr_accessor :platform
-    attr_accessor :configuration_language # example: ruby fastfile, swift fastfile
+    attr_accessor :fastlane_client_language # example: ruby fastfile, swift fastfile
 
-    def initialize(action_name: nil, p_hash: UNKNOWN_P_HASH, platform: nil, configuration_language: nil)
+    def initialize(action_name: nil, p_hash: UNKNOWN_P_HASH, platform: nil, fastlane_client_language: nil)
       @action_name = action_name
       @p_hash = p_hash
       @platform = platform
-      @configuration_language = configuration_language
+      @fastlane_client_language = fastlane_client_language
     end
 
-    def self.context_for_action_name(action_name, configuration_language: "ruby", args: nil)
+    def self.context_for_action_name(action_name, fastlane_client_language: :ruby, args: nil)
       app_id_guesser = FastlaneCore::AppIdentifierGuesser.new(args: args)
       return self.new(
         action_name: action_name,
         p_hash: app_id_guesser.p_hash || UNKNOWN_P_HASH,
         platform: app_id_guesser.platform,
-        configuration_language: configuration_language
+        fastlane_client_language: fastlane_client_language
       )
     end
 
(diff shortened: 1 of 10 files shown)


fastlane_core/lib/fastlane_core/analytics/analytics_event_builder.rb
```diff
@@ -2,16 +2,18 @@ module FastlaneCore
...
       @session_id = session_id
       @action_name = action_name
+      @fastlane_client_language = fastlane_client_language
     end
 
     def new_event(action_stage)
       {
         client_id: @p_hash,
-        category: :undefined,
+        category: @fastlane_client_language,
```
taquitos: cc: @powerivq ^

powerivq:
There are very limited dimensions we could use for a Google Analytics event (category, label and value). I wish we could make the values like a map (something like "lang=ruby,another_cat=val") in the event label, so that we can group the users in any conceivable way.

taquitos:
@powerivq right now, we only care about ruby vs swift. After a while, we should probably remove this code anyway. I don't see us adding a bunch of analytics, so I think utilizing a category for this is ok, for now 😄

powerivq (approved):

fastlane_core/lib/fastlane_core/analytics/action_launch_context.rb
```diff
@@ -8,22 +8,22 @@ class ActionLaunchContext
...
-    attr_accessor :configuration_language # example: ruby fastfile, swift fastfile
+    attr_accessor :fastlane_client_language # example: ruby fastfile, swift fastfile
 
-    def initialize(action_name: nil, p_hash: UNKNOWN_P_HASH, platform: nil, configuration_language: nil)
+    def initialize(action_name: nil, p_hash: UNKNOWN_P_HASH, platform: nil, fastlane_client_language: nil)
       @action_name = action_name
       @p_hash = p_hash
       @platform = platform
-      @configuration_language = configuration_language
+      @fastlane_client_language = fastlane_client_language
```
joshdholtz: @taquitos @powerivq Running an action directly (like `fastlane run debug`) actually has different values for `@configuration_language` / `@fastlane_client_language` before and after this change.

Before this change, `@configuration_language` had a value of `ruby` for a Swift project
After this change, `@fastlane_client_language` has a value of `swift` for a Swift project

Both **can** be right but I want to make sure that we are recording the right thing. Are we recording...
1) the client language that is initiating the action? This would be for us recording language stats per action.
or
2) the client language that the whole fastlane project is configured with? This would be for us recording language configured specifically per project and not directly what language is executing that action.

If its **1)**, I think this PR needs a change. `fastlane run debug` doesn't do anything with Swift so this will inflate the Swift analytics as the users isn't actually running anything fastlane with their `Fastfile.swift`
If its **2)**, this PR is 💯.

Lemme know if this doesn't make sense or isn't a problem 🙃

joshdholtz (commented):
Got a question about _what_ we are actually trying to record here 🙃 (see comment in `action_launch_context.rb`)

fastlane_core/lib/fastlane_core/analytics/action_launch_context.rb
taquitos: Great catch @joshdholtz! 
I think **(2)** is OK for what we're trying to do right now. I want to see how many folks have fastlane configured for their project. It would be really helpful to understand this number so we can figure out how to move fastlane.swift forward.
I had the sense that doing **(1)** would be a bit more accurate, but at the same time, a bit more involved. Due to the timing of recording the launch actions, it might mean we would need to rejigger the analytics calls a bit. I didn't want to do that this time around 😄

joshdholtz (approved):
Looks good! 👍 LGTM 💯

PR #13341, opened 2018-09-11
[downloads_dsyms] also check `displayVersionNumber`

bguidolim:
### Checklist

### Motivation and Context
I was facing a problem with `download_dsyms` action, it was caused firstly for a misconfiguration on App Store Connect. Below the steps to reproduce:

1. Register a new app with version 1.0
1. Set your app version (Xcode) to 1.0.0
1. Upload the app to TestFlight

I know that's not a common path, but could happen. The JSON provided by App Store Connect API will return a train with `versionString: "1.0"` (currently in use to check) and `displayVersionString: "1.0.0"`.

For some reason, Apple is using the app store version (1.0) as a key to get the builds but displaying (1.0.0) to the list.

I've tried to change the app store number from 1.0 to 1.0.0, but the JSON still using 1.0 as the identifier.

### Description

In order to fix that, I've added a new check to the action as a rescue check before return the error.

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -59,15 +59,16 @@ module Fastlane
           # - App Store version set as 1.0 and app version set as 1.0.0;
           #
           # In that case, the response from API will be "versionString = 1.0" and "displayVersionString = 1.0.0"
-          UI.verbose("Found train: #{train.version_string}, comparing to supplied version: #{version}")
-          if version && train.version_string == train.display_version_string && version != train.version_string
-            UI.verbose("Version #{version} doesn't match: #{train.version_string}")
-            next
-          else
-            UI.verbose("Failed. Comparing to display_version_string: #{train.display_version_string}")
-            if version && version != train.display_version_string
-              UI.verbose("Version #{version} doesn't match: #{train.display_version_string}")
+          if version
+            if train.version_string == train.display_version_string && version != train.version_string
+              UI.verbose("Version #{version} doesn't match: #{train.version_string}")
               next
+            else
+              UI.verbose("Failed. Comparing to display_version_string: #{train.display_version_string}")
+              if version != train.display_version_string
+                UI.verbose("Version #{version} doesn't match: #{train.display_version_string}")
+                next
+              end
             end
           end
 
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -64,8 +64,7 @@ module Fastlane
             UI.verbose("Version #{version} doesn't match: #{train.version_string}")
             next
           else
-            UI.verbose("Version #{version} doesn't match: #{train.version_string}")
-            UI.verbose("Found train: #{train.display_version_string}, comparing to supplied version: #{train.display_version_string}")
+            UI.verbose("Failed. Comparing to display_version_string: #{train.display_version_string}")
             if version && version != train.display_version_string
               UI.verbose("Version #{version} doesn't match: #{train.display_version_string}")
               next
@@ -73,7 +72,7 @@ module Fastlane
           end
 
           app.tunes_all_builds_for_train(train: train.version_string, platform: platform).each do |build|
-            # This probably is not necessary, but as we have that situation mentioned above, better check for safety.
+            # This is probably not necessary, but as we have the situation mentioned above, better check for safety.
             UI.verbose("Comparing train version #{build.train_version} to supplied version #{version}")
             if version && version != build.train_version
               UI.verbose("Train version #{build.train_version} doesn't match with supplied version: #{version}")
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -52,12 +52,34 @@ module Fastlane
         UI.message(message.join(" "))
 
         app.tunes_all_build_trains(platform: platform).each do |train|
+          # We try to check first with version_string, but also we try to check with display_version_string since
+          # this value could be different.
+          #
+          # Example:
+          # - App Store version set as 1.0 and app version set as 1.0.0;
+          #
+          # In that case, the response from API will be "versionString = 1.0" and "displayVersionString = 1.0.0"
           UI.verbose("Found train: #{train.version_string}, comparing to supplied version: #{version}")
-          if version && version != train.version_string
+          if version && train.version_string == train.display_version_string && version != train.version_string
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -52,12 +52,33 @@ def self.run(params)
...
+          # We try to check first with version_string, but also we try to check with display_version_string since
+          # this value could be different.
+          #
+          # Example:
+          # - App Store version set as 1.0 and app version set as 1.0.0;
+          #
+          # In that case, the response from API will be "versionString = 1.0" and "displayVersionString = 1.0.0"
           UI.verbose("Found train: #{train.version_string}, comparing to supplied version: #{version}")
-          if version && version != train.version_string
+          if version && train.version_string == train.display_version_string && version != train.version_string
```
joshdholtz: I think instead of adding the additional conditions in this `if` and the `else`, we could just change the original from `if version && version != train.version_string` to `if version && version != train.display_version_string`

Both of the changes that were done essentially come down to `version != train.display_version_string` so it doesn't look like there is a reason to have extra comparisons (except for maybe the verbose logging???)

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -52,12 +52,33 @@ def self.run(params)
...
+            if version && version != train.display_version_string
+              UI.verbose("Version #{version} doesn't match: #{train.display_version_string}")
+              next
+            end
           end
+
           app.tunes_all_builds_for_train(train: train.version_string, platform: platform).each do |build|
+            # This is probably not necessary, but as we have the situation mentioned above, better check for safety.
+            UI.verbose("Comparing train version #{build.train_version} to supplied version #{version}")
+            if version && version != build.train_version
```
joshdholtz: I _think_ this still makes sense though as a "just in case" 🧐

joshdholtz (changes requested):
I tried to reproduce his myself but could not get the same thing to happen where `versionString` and `displayVersionString` were different. But maybe we simply just use `displayVersionString` instead???? 🤔

fastlane/lib/fastlane/actions/download_dsyms.rb
bguidolim: I did my first implementation like that, I just wanted to play safe as I don't know if keep only one condition will fix the problem and continue working for all users of Fastlane.

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -52,12 +52,33 @@ def self.run(params)
...
+          # - App Store version set as 1.0 and app version set as 1.0.0;
+          #
+          # In that case, the response from API will be "versionString = 1.0" and "displayVersionString = 1.0.0"
           UI.verbose("Found train: #{train.version_string}, comparing to supplied version: #{version}")
-          if version && version != train.version_string
+          if version && train.version_string == train.display_version_string && version != train.version_string
             UI.verbose("Version #{version} doesn't match: #{train.version_string}")
             next
+          else
+            UI.verbose("Failed. Comparing to display_version_string: #{train.display_version_string}")
```
joshdholtz: @bguidolim I _think_ this `else` needs to get reworked. This `else` will get executed if `version` is `nil` which I don't think we want. I think it almost needs to get reworked something like...

```rb
if version
  if train.version_string == train.display_version_string && version != train.version_string
    UI.verbose("Version #{version} doesn't match: #{train.version_string}")
    next
  else
    UI.verbose("Failed. Comparing to display_version_string: #{train.display_version_string}")
    if version != train.display_version_string
      UI.verbose("Version #{version} doesn't match: #{train.display_version_string}")
      next
    end
  end
end
```

Thoughts?

joshdholtz (changes requested):
Thoughts on this rework? 🤔

fastlane/lib/fastlane/actions/download_dsyms.rb
bguidolim: Yeah! Now I see it. Fixing now.

joshdholtz:
Soooo, I pulled down this branch and this actually breaks my setup 😔 I had a similar issue to yours but where my `display_version_string` was the wrong one so this completely 💥 ed my setup.

I think we actually may need to do something like...
```rb
if version && (version != train.version_string || version != train.display_version_string)
```

This may feel weird because we may let through more than one but your change here (https://github.com/fastlane/fastlane/pull/13341/files#diff-65dfcad8f098c3c96688110dade15ed7R78) I think will make sure that only correct ones _actually_ gets downloads

bguidolim:
@joshdholtz I've simplified the condition and removed unnecessary check since API call already sent the version as parameter.

bguidolim:
ping

joshdholtz:
Thanks for the ping!

joshdholtz (commented):
One quick question!

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -52,11 +52,19 @@ def self.run(params)
...
-            UI.verbose("Version #{version} doesn't match: #{train.version_string}")
+          UI.verbose("Found train: version_string=#{train.version_string} / display_version_string=#{train.display_version_string}, comparing to supplied version: #{version}")
+          # We try to check first with version_string, but also we try to check with display_version_string since
+          # this value could be different.
+          #
+          # Example:
+          # - App Store version set as 1.0 and app version set as 1.0.0;
+          #
+          # In that case, the response from API will be "versionString = 1.0" and "displayVersionString = 1.0.0"
+          if version && version != train.version_string && version != train.display_version_string
```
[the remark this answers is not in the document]
bguidolim: No, this is not correct. This condition ensure that if we have a version set by the user, it should be equal to `version_string` or `display_version_string`.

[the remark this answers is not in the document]
bguidolim: I've tested here, it's working as before and for that specific case that made me open this PR.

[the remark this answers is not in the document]
joshdholtz: OH MAN, I reviewed that way too late at night last night 😝 I didn't do my boolean logic right. You are right you are right 🙃

bguidolim:
ping

joshdholtz (commented):
@bguidolim Sorry for getting back to you so late on this! I was re-reviewing this and thinking about this again. The one issue I can still see happening is that this dSYM that has mismatchd train and display version is 100% undownloadable now (I think) 🤔 Is my thinking correct?

I don't know what the proper solution is here. This may require a rework of how dSYMs get queried possibly. Thoughts?

PR #13351, opened 2018-09-12
[supply] prevent disabling of tracks on promotion

### LINKED ISSUE #13313: Supply seems to incorrectly deactivate lower tracks during rollout
### New Issue Checklist

### Issue Description
Supply appears to deactivate lower tracks when promoting to a new track. From what I can tell, this seems to result in APKs being unavailable to users who are whitelisted for the lower tracks.

Let's take an example here, where user A is whitelisted for the alpha track, but not the beta track. When an APK is promoted via supply from alpha to beta, we'd expect user A to be able to continue to receive the build. This doesn't appear to be the case - when looking at my Play Store console, it appears that the APK was removed from the alpha track during the promotion process.

I'm reasonably confident that this line (https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/uploader.rb#L67) is the source of this bug. It appears that Supply is *deliberately* clearing the original track when promoting, which appears to be causing unintentional behavior. Note that it doesn't matter if the user passes the `check_superseded_tracks` option or not - this happens *outside* of that block of logic.

It's possible that the behavior of the Google APIs have changed since this was authored two years ago - maybe it was necessary at one point? But it certainly seems like this is no longer needed.

I'd be happy to roll a PR adding a flag to bypass this line if we believe that's a reasonable fix. But I figured I would check with the experts here to see if my understanding is at all reasonable before attempting to fix.

tmtrademarked:
### Checklist

### Motivation and Context
This PR fixes an odd bug with Supply’s `promote` functionality. If a user is using Supply to promote APKs between tracks, the current code will deactivate the binary in the lower track. This is not the desired behavior - generally, you want to leave the binary in the lower track untouched. This PR adds an option to bypass this behavior if desired.

This will fix https://github.com/fastlane/fastlane/issues/13313 if accepted.

### Description
Added an option to allow bypassing the "deactivate on promote" behavior. By default, this is set to true to preserve the current behavior. Tested in the wild by using this to promote my most recent build, and observing that the existing binaries are untouched.

### DIFF
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -208,8 +208,13 @@ module Supply
                                      optional: true,
                                      description: "Timeout for read, open, and send (in seconds)",
                                      type: Integer,
-                                     default_value: 300)
-
+                                     default_value: 300),
+        FastlaneCore::ConfigItem.new(key: :deactivate_on_promote,
+                                     env_name: "SUPPLY_DEACTIVATE_ON_PROMOTE",
+                                     optional: true,
+                                     description: "When promoting to a new track, deactivate the binary in the origin track",
+                                     is_string: false,
+                                     default_value: true)
       ]
     end
     # rubocop:enable Metrics/PerceivedComplexity
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -208,13 +208,8 @@ module Supply
                                      optional: true,
                                      description: "Timeout for read, open, and send (in seconds)",
                                      type: Integer,
-                                     default_value: 300),
-        FastlaneCore::ConfigItem.new(key: :deactivate_on_promote,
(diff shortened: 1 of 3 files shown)


janpio:
(See discussion in #13313 before merging)

janpio (approved):

joshdholtz (changes requested):
@tmtrademarked Ahhhh, so **this** is what that deleted line was preventing 🤔 I think you are right. I think we might want to add a new option called `deactivate_on_promote` for the backwards compatability 😬 Do you mind reverting your second commit that removed that? 😇 

screen shot 2018-09-19 at 8 04 53 pm

tmtrademarked:
@joshdholtz - updated as per request! :-)

joshdholtz (approved):
This is great! Thanks for making this change 😊 ❤️

PR #13098, opened 2018-08-14
[action] Add force-with-lease to push_to_git_remote

sprabhuanand:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

I need to have the git option --force-with-lease, so that my changes to git will be pushed only when both local and remote is same. Force option sometimes overwrites the code when the remote is ahead of local

I ran the bundle exec fastlane test command and made sure all actions passed. I have pasted the output below

### Description

+------+---------------------------------------------------------------+-------------+
|                                  fastlane summary                                  |
+------+---------------------------------------------------------------+-------------+
| Step | Action                                                        | Time (in s) |
+------+---------------------------------------------------------------+-------------+
| 1    | Switch to validate_repo lane                                  | 0           |
| 2    | Switch to lint_source lane                                    | 0           |
| 3    | ensure_no_debug_code                                          | 0           |
| 4    | ensure_no_debug_code                                          | 0           |
| 5    | ensure_no_debug_code                                          | 0           |
| 6    | ensure_no_debug_code                                          | 0           |
| 7    | ensure_no_debug_code                                          | 0           |
| 8    | ensure_no_debug_code                                          | 0           |
| 9    | rubocop                                                       | 1           |
| 10   | find -E . -regex ".*\.(sh|bash)" -not -name "Pods-*" -name ". | 0           |
| 11   | Switch to execute_tests lane                                  | 0           |
| 12   | Switch to local_version lane                                  | 0           |
| 13   | PAGER=cat bin/fastlane --help                                 | 0           |
| 14   | bundle check || bundle install                                | 0           |
| 15   | rake install                                                  | 15          |
| 16   | bundle exec rake test_all                                     | 201         |
| 17   | Switch to validate_docs lane                                  | 0           |
| 18   | git clone https://github.com/fastlane/docs --depth=1          | 13          |
| 19   | Switch to verify_docs lane                                    | 0           |
| 20   | git clone https://github.com/fastlane/docs --depth=1          | 20          |
| 21   | bundle install                                                | 0           |
| 22   | bundle exec fastlane test skip_building:true                  | 3           |
[...]

PR #13391, opened 2018-09-19
[docs] Fix test_sample_code to use ConfigItem methods for ensuring valid values

joshdholtz:
Fixes some code sample issues that were created by #13098

## Problem
`test_sample_code` was actually falling with a config item that was marked as `type: Boolean` but not marked as `optional: true`. 

This was ended up falling because of...
```sh
'force' value must be a Fastlane::Boolean! Found TrueClass instead.
```

This was because there were no special test cases for `Fastlane::Boolean`. The only reason this was passing previously was  because it just so happened that every `type: Boolean` was marked as optional 🙃 

## Solution
The solution was to remove the custom testing of values and piggybacking off of the config item's `ensure_boolean_type_passes_validation` and `ensure_generic_type_passes_validation`.

### DIFF
--- a/fastlane/actions/test_sample_code.rb
+++ b/fastlane/actions/test_sample_code.rb
@@ -59,8 +59,14 @@ module Fastlane
             config_item = available_options.find { |a| a.key == current_argument }
             UI.user_error!("Unknown parameter '#{current_argument}' for action '#{method_sym}'") if config_item.nil?
 
-            if config_item.data_type && !value.kind_of?(config_item.data_type) && !config_item.optional && !config_item.skip_type_validation
-              UI.user_error!("'#{current_argument}' value must be a #{config_item.data_type}! Found #{value.class} instead.")
+            if value.nil? && config_item.optional
+              next
+            end
+
+            if config_item.data_type == Fastlane::Boolean
+              config_item.ensure_boolean_type_passes_validation(value)
+            else
+              config_item.ensure_generic_type_passes_validation(value)
             end
           end
         else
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/push_to_git_remote.rb
```diff
@@ -91,9 +91,9 @@ def self.example_code
             remote: "origin",         # optional, default: "origin"
             local_branch: "develop",  # optional, aliased by "branch", default: "master"
             remote_branch: "develop", # optional, default is set to local_branch
-            force: true,              # optional, default: false
```
janpio: why get rid of the `:`?

joshdholtz: Oops, lol...  I didn't mean too 😉

fastlane/actions/test_sample_code.rb
```diff
@@ -59,8 +59,12 @@ def self.method_missing(method_sym, *arguments, &_block)
             config_item = available_options.find { |a| a.key == current_argument }
             UI.user_error!("Unknown parameter '#{current_argument}' for action '#{method_sym}'") if config_item.nil?
 
-            if config_item.data_type && !value.kind_of?(config_item.data_type) && !config_item.optional && !config_item.skip_type_validation
-              UI.user_error!("'#{current_argument}' value must be a #{config_item.data_type}! Found #{value.class} instead.")
+            unless config_item.optional
```
powerivq: skip_type_validation?

Wondering why we won't test it if the arg is optional?

joshdholtz: `skip_type_validation` happens inside of the `ensure_boolean_type_passes_validation` and the `ensure_generic_type_passes_validation`

And I don't SUPER know, TBH  😬 But it was in the original implementation and it was failing without it. There were some that were failing that were optional where sample was specifying an `ENV` which was causing it to fail so this fixed it 🙃

powerivq (approved):

PR #13400, opened 2018-09-20
Comments on files that change with every release

janpio:
2 simple comments in files that are changed on each release when the list of authors is replaced and resorted.


joshdholtz (approved):
I like this 😊

PR #13405, opened 2018-09-20
[match] add support for multiple storage and encryption providers

KrauseFx:
### Open tasks 🎉 🍾 

- [x] Refactor match to support multiple implementations and ensure git based implementation still works
- [x] Clearly separate the storage from the encryption
- [x] Unify creating the `README.md` of the repo (if necessary)
- [x] Resolve all open `# TODO`s
- [x] Ensure the `change_password` command still works as expected for git based storage
- [x] Ensure the `nuke` command still works as expected for all storage providers
- [x] Ensure the `decrypt` command still works as expected for all storage providers
- [x] Ensure we don't use `:workspace` any where any more
- [x] Update tests
- [x] Never call `Storage::GitStorage.new` or `Encryption::OpenSSL.new` directly

## Separate PR
- [ ] Unify creating the `match_version` file
- [ ] When running `fastlane match decrypt`, is it expected that the `/var/...` folder is printed out as the exported repo?
- [ ] Automatically load all available implementations for the `Match::Storage::` and `Match::Encryption::` namespace to allow it to be extensible
- [ ] Offer a migrator between the various storage providers (separate PR)
- [ ] Update docs
	- [ ] Update docs.fastlane.tools guide
	- [ ] Update main _match_ README


KrauseFx:
The Windows test failures don't seem to be related

KrauseFx:
Ready to be merged from my side

janpio:
Windows failures are related, you are using a git action in a specific way for the first time in the codebase/tests that shows that there is a problem - fixing this now.

janpio:
Et voila, passes on Windows. 

(Lines 118 in this file indicates that there is other code that will not work in Windows as well - but this will have to be resolved in another PR)

match/lib/match/nuke.rb
```diff
@@ -15,22 +18,34 @@ class Nuke
...
 
-      params[:workspace] = GitHelper.clone(params[:git_url],
-                                           params[:shallow_clone],
-                                           skip_docs: params[:skip_docs],
-                                           branch: params[:git_branch],
-                                           git_full_name: params[:git_full_name],
-                                           git_user_email: params[:git_user_email],
-                                           clone_branch_directly: params[:clone_branch_directly])
+      self.storage = Storage::Interface.storage_class_for_storage_mode(params[:storage_mode]).new
+      self.storage.configure(git_url: params[:git_url],
```
milch: Hmm, reading more of the code, how about all the sub-commands skim the params they need off the top and then return the rest to the storage engine directly? e.g.

```
storage_mode = params.delete(:storage_mode)
readonly = params.delete(:readonly)
# etc.

self.storage = Storage::Interface.storage_class_for_storage_mode(storage_mode).new
self.storage.configure(**params)

# ... etc.
```

That way the parameters can be changed to make sense for the respective storage engine. For example, say I was writing a "simple sftp storage interface" (SSSInterface for short) then things like `git_url` and  `clone_branch_directly` won't make sense, but instead I may want things like `sss_url` or `only_download_subfolder`

match/lib/match/encryption/interface.rb
```diff
@@ -0,0 +1,29 @@
+module Match
+  module Encryption
+    class Interface
+      # Returns the class to be used for a given `storage_mode`
+      def self.encryption_class_for_storage_mode(storage_mode)
```
milch: This interface seems a bit clunky, did you have some usage in mind that does not always look like this?

```
self.storage = Storage::Interface.storage_class_for_storage_mode(params[:storage_mode]).new
```

Otherwise this could be more elegantly replaced by adding a helper function to the module to something like 

```
self.storage = Storage.for_mode(params[:storage_mode])
```

match/lib/match/change_password.rb
```diff
@@ -1,31 +1,44 @@
...
+      storage.download
+
+      encryption = Encryption::Interface.encryption_class_for_storage_mode(params[:storage_mode]).new(
+        git_url: storage.git_url,
+        working_directory: storage.working_directory
+      )
+      encryption.decrypt_files
+
+      encryption.clear_password
+      encryption.store_password(to)
```
milch: These two aren't defined on the interface. I think it makes sense to have one public call `change_password`. I'm not sure storing a password makes sense for all encryption interfaces, so I would also add a `has_local_password?` method to the interface. Then this action should have an early return for encryption interfaces that don't save the password to the local keychain.

match/lib/match/nuke.rb
```diff
@@ -89,11 +104,11 @@ def prepare_list
         self.profiles += profile_type(prov_type).all
       end
 
-      certs = Dir[File.join(params[:workspace], "**", cert_type.to_s, "*.cer")]
-      keys = Dir[File.join(params[:workspace], "**", cert_type.to_s, "*.p12")]
+      certs = Dir[File.join(self.storage.working_directory, "**", cert_type.to_s, "*.cer")]
+      keys = Dir[File.join(self.storage.working_directory, "**", cert_type.to_s, "*.p12")]
```
milch: I think it'd be more elegant to provide this as a method on the storage interface, e.g.

```
self.storage.find_all(File.join(cert_type.to_s, "*.cer")) 
```

match/lib/match/runner.rb
```diff
@@ -54,22 +68,27 @@ def run(params)
...
         end
       end
 
       # Done
-      if self.files_to_commmit.count > 0 && !params[:readonly]
-        message = GitHelper.generate_commit_message(params)
-        GitHelper.commit_changes(params[:workspace], message, params[:git_url], params[:git_branch], self.files_to_commmit)
+      if self.files_to_commit.count > 0 && !params[:readonly]
+        Dir.chdir(storage.working_directory) do
+          return if `git status`.include?("nothing to commit")
```
milch: Most of this block should be in the storage

match/lib/match/runner.rb
```diff
@@ -85,23 +104,23 @@ def run(params)
...
         UI.important("Couldn't find a valid code signing identity in the git repo for #{cert_type}... creating one for you now")
         UI.crash!("No code signing identity found and can not create a new one because you enabled `readonly`") if params[:readonly]
-        cert_path = Generator.generate_certificate(params, cert_type)
+        cert_path = Generator.generate_certificate(params, cert_type, working_directory)
         private_key_path = cert_path.gsub(".cer", ".p12")
 
-        self.files_to_commmit << cert_path
-        self.files_to_commmit << private_key_path
+        self.files_to_commit << cert_path
+        self.files_to_commit << private_key_path
```
milch: This can be better kept track of in the storage, e.g. replace this with:

```
self.storage.updated_file(cert_path)
self.storage.updated_file(private_key_path)
```

That'll allow you to get rid of `files_to_commit` here completely, which is very git-centric

For some storage interfaces there'd be no concept of committing files, so for them this would be a noop.

match/lib/match/encryption/interface.rb
joshdholtz: Ohhhhh, I think I like that too... I will make that change 😊

match/lib/match/nuke.rb
joshdholtz: @milch Are you thinking we should just pass the whole `params` object in to the configure method?

match/lib/match/module.rb
```diff
@@ -4,12 +4,16 @@ module Match
...
   ROOT = Pathname.new(File.expand_path('../../..', __FILE__))
-  DESCRIPTION = "Easily sync your certificates and profiles across your team using git"
+  DESCRIPTION = "Easily sync your certificates and profiles across your team"
 
   def self.environments
     return %w(appstore adhoc development enterprise)
   end
 
+  def self.storage_modes
+    return %w(git google_cloud)
```
joshdholtz: `google_cloud` isn't implemented yet so we should probably remove this for now 🙃

match/lib/match/encryption/openssl.rb
```diff
@@ -0,0 +1,149 @@
...
+  module Encryption
+    class OpenSSL < Interface
+      attr_accessor :git_url
+
+      attr_accessor :working_directory
+
+      # @param git_url: The Git URL is used for identifiying a specific repo
+      #                 which is used to store the passphrase in the Keychain
+      # @param working_directory: The path to where the certificates are stored
+      def initialize(git_url: nil, working_directory: nil)
```
joshdholtz: I _think_ we should actually rename this from `git_url` to `server_name` or `keychain_key` since this should be storage agnostic 🙃

match/lib/match/change_password.rb
```diff
@@ -1,31 +1,44 @@
...
-                                  manual_password: from,
-                                  skip_docs: params[:skip_docs],
-                                  branch: params[:git_branch],
-                                  git_full_name: params[:git_full_name],
-                                  git_user_email: params[:git_user_email],
-                                  clone_branch_directly: params[:clone_branch_directly])
-      Encrypt.new.clear_password(params[:git_url])
-      Encrypt.new.store_password(params[:git_url], to)
+
+      to ||= ChangePassword.ask_password(message: "New passphrase for Git Repo: ", confirm: true)
```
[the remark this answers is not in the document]
joshdholtz: @milch Can you explain more what you mean here? I think I might be a bit confused 😖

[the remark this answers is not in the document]
milch: Of course, sorry about that. I'm just saying that "git repo" is not accurate anymore, and all references outside of the git storage should rename it to some other consistent value. If you want to be fancy you could add a method to the Storage interface (e.g. `human_readable_name`) and use that when referencing the storage:

```
ChangePassword.ask_password(message: "New passphrase for #{storage.human_readable_name}:")
```

Would then print this if I were using git storage: 

```
New passphrase for Git Repo:
```

Or if I used, say, a theoretical "Simple SFTP Storage" (SSS for short) that called its version of repositories "Canisters" then it would print 

```
New passphrase for SSS Canister:
```

joshdholtz (approved):
I think we are good to go with this! 🙌

PR #13442, opened 2018-09-26
Update `fastlane spaceauth` command to work with new API

KrauseFx:
It seems like either this was broken before, or the cookie names have changed. We need to use `myacinfo` here.


joshdholtz (approved):
This checkouts out ✅ LGTM

PR #13459, opened 2018-09-28
[action] remove UDID format validation from register_device

thii:
### Checklist

### Motivation and Context

In the MDM Protocol Reference, Apple recommends to not hard code assumptions of UDID format as the format may change in the future.

https://developer.apple.com/enterprise/documentation/MDM-Protocol-Reference.pdf

### Description

Remove UDID format validation in `register_deivce` and `register_deivce` actions.

### DIFF
--- a/fastlane/lib/fastlane/actions/register_device.rb
+++ b/fastlane/lib/fastlane/actions/register_device.rb
@@ -3,8 +3,6 @@ require 'credentials_manager'
 module Fastlane
   module Actions
     class RegisterDeviceAction < Action
-      UDID_REGEXP = /^(\h{40}|\h{8}-\h{16})$/
-
       def self.is_supported?(platform)
         platform == :ios
       end
@@ -19,7 +17,6 @@ module Fastlane
         Spaceship.login(credentials.user, credentials.password)
         Spaceship.select_team
 
-        UI.user_error!("Passed invalid UDID: #{udid} for device: #{name}") unless UDID_REGEXP =~ udid
         Spaceship::Device.create!(name: name, udid: udid)
 
         UI.success("Successfully registered new device")
--- a/fastlane/lib/fastlane/actions/register_device.rb
+++ b/fastlane/lib/fastlane/actions/register_device.rb
@@ -20,9 +20,8 @@ module Fastlane
         begin
           Spaceship::Device.create!(name: name, udid: udid)
         rescue => ex
-          UI.error("Failed to register new device")
           UI.error(ex.to_s)
-          return nil
+          UI.crash!("Failed to register new device (name: #{name}, UDID: #{udid})")
         end
 
         UI.success("Successfully registered new device")
--- a/fastlane/lib/fastlane/actions/register_device.rb
+++ b/fastlane/lib/fastlane/actions/register_device.rb
@@ -17,7 +17,13 @@ module Fastlane
         Spaceship.login(credentials.user, credentials.password)
         Spaceship.select_team
 
-        Spaceship::Device.create!(name: name, udid: udid)
+        begin
+          Spaceship::Device.create!(name: name, udid: udid)
+        rescue => ex
+          UI.error("Failed to register new device")
+          UI.error(ex.to_s)
+          return nil
+        end
 
         UI.success("Successfully registered new device")
         return udid
--- a/fastlane/lib/fastlane/actions/register_devices.rb
+++ b/fastlane/lib/fastlane/actions/register_devices.rb
@@ -51,8 +51,8 @@ module Fastlane
       def try_create_device(name, udid, mac)
         Spaceship::Device.create!(name: name, udid: udid, mac: mac)
       rescue => ex
-        UI.error("Failed to register new device (name #{name}, UDID: #{udid})")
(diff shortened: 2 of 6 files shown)


janpio:
What happens if one passes a really invalid ID in there? Do we get API feedback on that? Is the error handling code up to speed and gives some helpful output?

fastlane/lib/fastlane/actions/register_devices.rb
```diff
@@ -28,7 +24,6 @@ def self.run(params)
 
         if devices
           device_objs = devices.map do |k, v|
-            UI.user_error!("Passed invalid UDID: #{v} for device: #{k}") unless udid_regexp =~ v
             next if existing_devices.map(&:udid).include?(v)
             Spaceship::Device.create!(name: k, udid: v, mac: mac)
```
janpio: I think you missed this one.

janpio:
Awesome update @thii. Left you a comment.

fastlane/lib/fastlane/actions/register_device.rb
```diff
@@ -19,8 +17,13 @@ def self.run(params)
         Spaceship.login(credentials.user, credentials.password)
         Spaceship.select_team
 
-        UI.user_error!("Passed invalid UDID: #{udid} for device: #{name}") unless UDID_REGEXP =~ udid
-        Spaceship::Device.create!(name: name, udid: udid)
+        begin
+          Spaceship::Device.create!(name: name, udid: udid)
+        rescue => ex
+          UI.error("Failed to register new device")
```
janpio: What's the difference between `UI.error` and `UI.user_error!` that was used before?

thii:
I think `UI.user_error` is used for user input errors, and `UI.error` is for other kinds of errors (fastlane errors, system errors, etc.). 

I chose the latter since the error returned from App Store Connect may not be an user error (e.g. Apple server hiccups).

fastlane/lib/fastlane/actions/register_device.rb
joshdholtz: - `UI.error` will print out message formatted as errors to the console (typically red)
- `UI.user_error!` will also print out messages formatted in errors but will also raise an error (which is what the `!` means here` as this will halt execution)

janpio: Returning `nil` achieves the same as raising an error?

spaceship/lib/spaceship/portal/portal_client.rb
```diff
@@ -532,7 +532,12 @@ def create_device!(device_name, device_id, mac: false)
         register: 'single'
       })
 
-      parse_response(req, 'devices').first
+      devices = parse_response(req, 'devices')
+      return devices.first unless devices.empty?
+
+      raise parse_response(req, 'validationMessages').map { |message|
```
joshdholtz: Can we make this raise an `UnexpectedResponse` like the following taken from https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/portal_client.rb#L595...

```rb
raise UnexpectedResponse.new, "Couldn't download certificate, got this instead: #{a}"
```

Just want to be consistent with some other error messages 😊

fastlane/lib/fastlane/actions/register_devices.rb
```diff
@@ -54,6 +48,13 @@ def self.run(params)
         return device_objs
       end
 
+      def try_create_device(name, udid, mac)
+        Spaceship::Device.create!(name: name, udid: udid, mac: mac)
+      rescue => ex
+        UI.error("Failed to register new device (name #{name}, UDID: #{udid})")
+        UI.error(ex.to_s)
```
joshdholtz: I think we might want to make this...
```rb
UI.error(ex.to_s)
UI.crash!("Failed to register new device (name #{name}, UDID: #{udid})")
```

We want to make sure that we end execution and that an error return code gets bubbled to the top of this execution

fastlane/lib/fastlane/actions/register_device.rb
```diff
@@ -19,8 +17,13 @@ def self.run(params)
...
+          UI.error(ex.to_s)
+          return nil
```
joshdholtz: Same here...

I think we might want to make this...
```rb
UI.error(ex.to_s)
UI.crash!("Failed to register new device (name #{name}, UDID: #{udid})")
```

We want to make sure that we end execution and that an error return code gets bubbled to the top of this execution

joshdholtz (changes requested):
This looks great! Just some minor feedback/improvements about how to handle the errors 😊

spaceship/lib/spaceship/portal/portal_client.rb
thii: But... this is an expected response... which is why we could get the validation messages from the response. Though I added a case where the `validationMessages` array could be empty.

joshdholtz (changes requested):
One last request! Can you also run rubocop when you are done too to fix the linting errors? 🙃

joshdholtz (approved):
This looks great! Thanks for making these changes 😊 ❤️

Issue #11841, opened 2018-02-14
Showing language for failed deliver metadata (Copy of #6948, #10277)


timothycosta:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
When uploading metadata, the problem language does not appear.  In one of the below issues `--verbose` was requested but never received.  The output below includes it.

It would be neat if precheck could look for invalid characters.  I finally had to open a hex editor to figure out what the issue was with my metadata today (sometimes a space is more than a space!).

Duplicates [here](https://github.com/fastlane/fastlane/issues/6948) and [here](https://github.com/fastlane/fastlane/issues/10277).  

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
timothycosta@Timothys-Mac-Mini ~/D/l/i/fastlane> fastlane --verbose --env main upload metadata:1 skip_metadata_fetch:1
[✔] 🚀
[14:38:36]: fastlane detected a Gemfile in the current directory
[14:38:36]: however it seems like you don't use `bundle exec`
[14:38:36]: to launch fastlane faster, please use
[14:38:36]:
[14:38:36]: $ bundle exec fastlane --verbose --env main upload metadata:1 skip_metadata_fetch:1
[14:38:36]:
[14:38:36]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
DEBUG [2018-02-14 14:38:37.68]: Checking if there are any plugins that should be loaded...
Successfully loaded Appfile at path '/Users/timothycosta/Desktop/lingq/iLingQ/fastlane/Appfile'
- apple_id: ''
- team_id: ''
-------
(eval):4: warning: already initialized constant FastlaneCore::ConfigurationFile::MATCH_SHALLOW_CLONE
(eval):4: warning: previous definition of MATCH_SHALLOW_CLONE was here
DEBUG [2018-02-14 14:38:38.11]: Using metadata: 1
DEBUG [2018-02-14 14:38:38.11]: Using skip_metadata_fetch: 1
[...]

PR #13461, opened 2018-09-28
[deliver] allow metadata items to be uploaded one at a time

timothycosta:
### Checklist

### Motivation and Context

When uploading localized metadata it can be difficult to determine which specific items have issues.  ITC does not return a detailed report of which languages had errors.

Addresses [this issue](https://github.com/fastlane/fastlane/issues/11841), though not via precheck.

### Description

Added an option `individual_metadata_items`.  Items in this array will be uploaded one at a time instead of in bulk.  Rather than using `UI.user_error!` issues are displayed with `UI.error` so that all problems can be identified in one run

I set my metadata to use a known-unavailable `name` and uploaded it, receiving an error.  Also tried with `marketing_url` to ensure that both categories of metadata (version and app) work.

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -257,6 +257,13 @@ module Deliver
                                      default_value: :warn),
 
         # App Metadata
+        FastlaneCore::ConfigItem.new(key: :individual_metadata_items,
+                                     env_name: "DELIVER_INDIVUDAL_METADATA_ITEMS",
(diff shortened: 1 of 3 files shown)


joshdholtz (changes requested):
This is a really cool PR 🙌 Do you think you would be able to add some tests around this in `upload_metadata_spec.rb`? 😇

timothycosta:
@joshdholtz Finally was able to set some time aside here.  Let me know if those tests look ok.

joshdholtz (approved):
This for adding those tests! This PR is 🔥 and I hope this is helps a lot of people out 💪 🚀

PR #13462, opened 2018-09-28
[snapshot] Improve screenshot rotation code

janpio:
Screenshot rotation in `snapshot` was still using `FastlaneCore::FastlanePty.spawn` directly. Changed it to use `FastlaneCore::CommandExecutor.execute` (which then uses `FastlaneCore::FastlanePty.spawn` internally, but also properly handles errors etc.)


KrauseFx (approved):
👍

PR #13471, opened 2018-09-29
Add action of bugtags

wliu6:
add action for submit dsym to (Bugtags)[https://work.bugtags.com] .

Bugtags like apteligent and/or splunk,that has the most of user in Chinese Developers.

I hope it can bring convenience to the developer.  so, hope to adopt.

### DIFF
--- a/fastlane/lib/fastlane/actions/bugtags.rb
+++ b/fastlane/lib/fastlane/actions/bugtags.rb
@@ -1,0 +1,113 @@
+module Fastlane
+  module Actions
+    module SharedValues
+      BUGTAGS_CUSTOM_VALUE = :BUGTAGS_CUSTOM_VALUE
+    end
+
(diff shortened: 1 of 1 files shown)


joshdholtz (changes requested):
@wliu6 Thanks for taking the time to make this contribution! This good looks 💯 but we are currently not accepting the addition of new actions to third-party services. We created the plugin system for actions just like this - https://docs.fastlane.tools/plugins/create-plugin/

There are a few benefits of making actions like this as a plugin:
1) They can be updated independently of _fastlane_ versions
2) Prevents maintainers from having to keep knowledge of third-party services and less code to test/maintain

We do appreciate the time taken to make this PR and are looking forward to the addition of another plugin 😊

wliu6:
I see, so i can create a new plugin for fastlane users to use bugtags.

PR #13478, opened 2018-10-02
[match] reworked requires for new match storage and encryption classes

joshdholtz:
## Problem
- There was a circular dependency between `ChangePassword` and `Encryption::OpenSSL`
- The rest were some files that weren't being required

## Solution
- Remove comments with `# TODO`
- Created `match/lib/match/storage.rb` and `match/lib/match/encryption.rb`
  - Used for grouping all the requires together (less imports in other files)
- Added imports

### DIFF
--- a/match/lib/match/storage/git_storage.rb
+++ b/match/lib/match/storage/git_storage.rb
@@ -1,5 +1,8 @@
 require 'fastlane_core/command_executor'
 
+require_relative '../module'
+require_relative './interface'
+
 module Match
   module Storage
     # Store the code signing identities in a git repo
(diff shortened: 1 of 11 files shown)


match/lib/match/storage/git_storage.rb
```diff
@@ -31,7 +34,7 @@ def configure(type: nil,
         self.git_url = git_url
         self.shallow_clone = shallow_clone
         self.manual_password = manual_password
-        self.skip_docs = skip_docs # TODO: We don't use this yet
+        self.skip_docs = skip_docs # We don't use this yet
```
KrauseFx: Ah I need to look into those, thanks

joshdholtz: Np! I Made this a separate PR so that the todo removals stood out more 🙃

KrauseFx (approved):
Thanks @joshdholtz, let's merge this, I'll do the follow-up changes for the `# TODO` items 👍

PR #13500, opened 2018-10-08
match multiple storage implementation reworks/cleanup

joshdholtz:
## 1. Reworked Storage
`Match::Storage::Interface` was being used as a factory for creating which interface was needing to be used at runtime. Having implementation being referenced in the interface is what caused a some weird circular dependency issues in previous PR. This fixes those fixes and moves the factory creation into `Match::Storage`.

```rb
Storage::Interface.storage_class_for_storage_mode(params[:storage_mode])
```
is now
```rb
Storage.for_mode
```

## 2. Reworked Encryption
Similar to the above with storage but for `Match::Encryption::Interface` and `Match::Encryption`

```rb
Encryption::Interface.encryption_class_for_storage_mode
```
is now
```rb
Encryption.for_storage_mode
```

## 3. Relpalced `git_url` with `keychain_name` in OpenSSL
`Match::Encryption::OpenSSL` had an attribute for `git_url` which was only being used to set the Keychain server name (for storing the password). I renamed `git_url` to `keychain_name` so that this `Match::Encryption::OpenSSL` implementation could be used with other storage implementations in the future (svn, ftp, etc).

## 4. Removed `google_cloud` as option
This is going to get released out into the wild before `google_clould` implementation is completed so this is being removed as an option (for now).

### DIFF
--- a/match/lib/match/change_password.rb
+++ b/match/lib/match/change_password.rb
@@ -24,7 +24,7 @@ module Match
       storage.download
 
       encryption = Encryption.for_storage_mode(params[:storage_mode], {
-        git_url: storage.git_url,
+        git_url: params[:git_url],
         working_directory: storage.working_directory
       })
       encryption.decrypt_files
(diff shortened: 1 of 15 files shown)


KrauseFx (approved):
Beautiful, thanks for doing the clean-ups 👍

PR #13509, opened 2018-10-09
Fix new try_create_device method

jgadbois:
### Checklist

### Motivation and Context

Master is broken when attempting to register devices

### Description

Updated the new method to be a class method with keyword arguments since that's how it's being called

Our ios dev ran a build for a client


KrauseFx (approved):
Oh wow, nice catch 👍 thanks

joshdholtz (approved):
Thanks for finding this! ❤️

PR #13519, opened 2018-10-11
Deliver: skip version check when no icons are provided.

barakwei:
### Checklist

### Motivation and Context
When uploading an IPA with a version that already exists, and without changing metadata, assets or icons (by using `skip_*` flags and not providing icons), the upload fails with the error `Could not find a version to edit for app "MyApp"`.This is because in this case the method `edit_version` in the `Spaceship::Tunes::Application` class return `nil`. The class `UploadAssets` does the `nil` verification regardless if icons were provided.

`UploadScreenshots` has the same check but the `skip_screenshots` flag can be used if screenshop uploading is not needed,  but I don't think another `skip_assets` flag is necessary.

### Description
My fix adds early bail if icons are not provided, thus also skipping precious calls to the server.

### DIFF
--- a/deliver/lib/deliver/upload_assets.rb
+++ b/deliver/lib/deliver/upload_assets.rb
@@ -4,9 +4,9 @@ module Deliver
   class UploadAssets
     SUPPORTED_ICON_EXTENSIONS = [".png", ".jpg", ".jpeg"]
     def upload(options)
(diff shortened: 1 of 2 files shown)


getaaron (approved):
This looks great, thanks for the PR! (And the tests!)

PR #13543, opened 2018-10-15
[supply] adding multiple aabs

### LINKED ISSUE #13536: Support for multiple AABs
### Feature Request

Right now I don't see a way to upload multiple .aab to the Play Store, just like it is with APK. I have different build types that I would like to upload and this is not a problem using apk (using `apk_paths`), but I can't do it with aab. 

#### Feature Description
The feature would the like this:
```
lane :internal_bundle do
        supply(
            package_name: "com.my.package",
            track: "internal",
            aab_paths: [
                "path/to/my/app/appBuild1.aab",
                "path/to/my/app/appBuild2.aab",
            ]
        )
```

The only alternative that I see for now is to don't use aab if you use more than one buildType. If there is some workaround, please let me know!

If this is something that makes sense to be added to Fastlane, I don't mind creating the PR. Please let me know what do you think.

leandroBorgesFerreira:
### Checklist

### Motivation and Context
fixes #13536. Add the possibility to pass multiple aab just like apk_paths. 

### Description
The code has the same logic as apk_path command. 
As a last test, I uploaded an array of aab to Play Store using the `aab_paths` without any problems.

### DIFF
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -92,7 +92,7 @@ module Supply
                                      env_name: "SUPPLY_APK",
                                      description: "Path to the APK file to upload",
                                      short_option: "-b",
-                                     conflicting_options: [:apk_paths, :aab],
+                                     conflicting_options: [:apk_paths, :aab, :aab_paths],
                                      code_gen_sensitive: true,
                                      default_value: Dir["*.apk"].last || Dir[File.join("app", "build", "outputs", "apk", "app-Release.apk")].last,
                                      default_value_dynamic: true,
@@ -103,7 +103,7 @@ module Supply
                                      end),
         FastlaneCore::ConfigItem.new(key: :apk_paths,
                                      env_name: "SUPPLY_APK_PATHS",
-                                     conflicting_options: [:apk, :aab],
+                                     conflicting_options: [:apk, :aab, :aab_paths],
                                      optional: true,
                                      type: Array,
                                      description: "An array of paths to APK files to upload",
@@ -119,7 +119,7 @@ module Supply
                                      env_name: "SUPPLY_AAB",
                                      description: "Path to the AAB file to upload",
                                      short_option: "-f",
-                                     conflicting_options: [:apk_path, :apk_paths],
+                                     conflicting_options: [:apk_path, :apk_paths, :aab_paths],
                                      code_gen_sensitive: true,
                                      default_value: Dir["*.aab"].last || Dir[File.join("app", "build", "outputs", "bundle", "release", "bundle.aab")].last,
                                      default_value_dynamic: true,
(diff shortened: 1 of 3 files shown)


leandroBorgesFerreira:
I signed it!

bdrelling (commented):
Just a few tiny things that might be worth addressing.

supply/lib/supply/options.rb
```diff
@@ -128,6 +128,20 @@ def self.available_options
                                        UI.user_error!("Could not find aab file at path '#{value}'") unless File.exist?(value)
                                        UI.user_error!("aab file is not an aab") unless value.end_with?('.aab')
                                      end),
+        FastlaneCore::ConfigItem.new(key: :aab_paths,
+                                     env_name: "SUPPLY_AAB_PATHS",
+                                     conflicting_options: [:apk_path, :apk_paths, :aab],
```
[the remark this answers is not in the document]
leandroBorgesFerreira: I think we should address it in this PR. Good catch.

leandroBorgesFerreira:
@bdrelling I just addressed your comments. The pipeline didn't run because of some infrastructure problem in Circle CI. would you mind retriggering it?

supply/lib/supply/uploader.rb
```diff
@@ -152,11 +152,17 @@ def upload_mapping(apk_version_codes)
     end
 
     def upload_bundles
-      aab_path = Supply.config[:aab]
-      return [] unless aab_path
+      aab_paths = [Supply.config[:aab]] unless (aab_paths = Supply.config[:aab_paths])
+      # return [] unless aab_paths
```
joshdholtz: @leandroBorgesFerreira Can we remove this commented out line? 🙃

joshdholtz (changes requested):
This looks great! Thanks for making this change 😊 Can you remove that one commented out line and then we can get this merged and released today 💪

supply/lib/supply/uploader.rb
leandroBorgesFerreira: fixed =]

supply/lib/supply/uploader.rb
```diff
@@ -152,11 +152,17 @@ def upload_mapping(apk_version_codes)
     end
 
     def upload_bundles
-      aab_path = Supply.config[:aab]
-      return [] unless aab_path
+      aab_paths = [Supply.config[:aab]] unless (aab_paths = Supply.config[:aab_paths])
```
joshdholtz: Ah, one more thing! There is a very small possibility that `Supply.config[:aab]` might be `nil` so `aab_paths` could be `[nil]`. Can we add a `aab_paths.compact!` after this line? Thhheeennnn I think we should be good 💪

joshdholtz (changes requested):
One more quick change!

supply/lib/supply/uploader.rb
leandroBorgesFerreira: wouldn't be better to use `compact` instead of `compact!`? `compact!` would return nil if no removal were made.

supply/lib/supply/uploader.rb
```diff
@@ -152,11 +152,17 @@ def upload_mapping(apk_version_codes)
...
+      return [] unless aab_paths.compact
```
joshdholtz: Ah sorry, I was looking at how this was implemented in the `upload_apks` - https://github.com/fastlane/fastlane/pull/13543/files#diff-fd047a79b6e303ba583ff34a4c489d78L134. `upload_apks` doesn't early return so doing a `aab_paths.compact!` on its own lined would worked.

I think it will always get passed with `return [] unless aab_paths.compact` (as long as `aab_paths` isn't `nil`). If `aab_paths` is `nil` this looks like it will 💥 .

I think what we actually want is something like...

```suggestion
      return [] unless aab_paths
      aab_paths.compact!
```

This way it will early return with an empty array if `aab_paths` is `nil`, otherwise it will remove all possible `nil` files inside of `aab_paths`

joshdholtz (changes requested):
One small suggestion and then I think we are 💯

leandroBorgesFerreira:
So clicking in "apply suggestion" wasn't a great idea hahaha. Now this PR contains commits that were not done by me... Anyway, I just created another PR, just like this one, but it passes the `cla/google` check. https://github.com/fastlane/fastlane/pull/13639

joshdholtz:
@leandroBorgesFerreira Yeah, that is expected from googlebot - its okay 😊 Its not more of a warning than an error 😉 We can still continue with this PR

joshdholtz (approved):
Thanks for this contribution ❤️ Looking forward to getting this one released 🚀

PR #13585, opened 2018-10-25
Use train_version to detect build on download_dsym action

giginet:
### Checklist

### Motivation and Context

I use `download_dsyms` action with `latest` version.

```ruby
download_dsyms(version: 'latest')
```

It attempts to search builds by a version which specified at AppStore Connect.

However, when it version and build's train_version is mismatched. This action can't find the correct build.

screen shot 2018-10-25 at 17 08 59

### Description

When `latest` is passed as version, use `train_version` to search latest build.

And I added specs for `download_dsym` action.
There were no specs....


fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -98,6 +97,16 @@ def self.download(download_url, bundle_id, train_number, build_version, output_d
         Actions.lane_context[SharedValues::DSYM_PATHS] << File.expand_path(path)
       end
 
+      def self.get_version(latest_version)
+        candidate_build = latest_version.candidate_builds.first
```
giginet: We can't use Safe navigation operator because of TargetRubyVersion 😢 
So this method is needed.

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -98,6 +97,16 @@ def self.download(download_url, bundle_id, train_number, build_version, output_d
...
+        if candidate_build.nil?
+          latest_version.version
```
giginet: If any `candidate_build` are not found, fails back to previous behavior.

joshdholtz (approved):
Ah, I've experienced this before. Thanks for making this change 😊

PR #13589, opened 2018-10-25
Implement Google Cloud Storage match

KrauseFx:
Implement match backend storage using Google Cloud KMS together with Google Cloud Storage

**Open tasks:**
- [x] Decide on how/where to encrypt
- [x] Implement file storage
- [x] How does a user specify the bucket name and the auth file
- [x] `match migrate` should probably tell the user how to migrate the `Matchfile`
- [x] Prefix the bucket with the team_id, while staying backwards compatible
- [x] Ensure `fastlane match nuke` works
- [x] Extract out keys.json
- [x] Ensure all other `fastlane match` commands still work with the new storage
- [x] Check Google API key permission and show nice error message to debug permission issues
- [x] Automatically create bucket for user, and let user provide a custom bucket name
- [x] Add basic error handling
- [x] `fastlane match init` should ask for storage backend to use
- [x] Unit tests
- [x] Make sure to support the `gc_keys.json` file to be inside the `fastlane` folder or outside
- [x] Check if we can download multiple files in parallel to speed up the process
- [x] green `rubocop`
- [x] Fix all `# TODO`s

Moved all follow-up tasks over to https://github.com/fastlane/fastlane/issues/13772, to get this PR merged first


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please provide a changelog summary in the Pull Request description @KrauseFx</td>
    </tr>
  </tbody>
</table>

</p>

milch:
Related: 
https://github.com/fastlane/fastlane/pull/13646

KrauseFx:
Continuing work on this one now

KrauseFx:
Follow-up tasks https://github.com/fastlane/fastlane/issues/13772

match/lib/match/migrate.rb
```diff
@@ -0,0 +1,97 @@
...
+      files_to_commit = []
+      Dir.chdir(git_storage.working_directory) do
+        Dir[File.join("**", "*")].each do |current_file|
+          next if File.directory?(current_file)
+
+          to_path = File.join(google_cloud_storage.working_directory, team_id, current_file)
+          FileUtils.mkdir_p(File.expand_path("..", to_path))
+
+          if File.exist?(to_path)
+            UI.user_error!("Looks like file already exists on Google Cloud Storage at path '#{to_path}', stopping the migration process. Please make sure the bucket is empty, or at least doesn't contain any files related to the same Team ID")
```
snatchev: Do you mean "directory" instead of "bucket" ?

match/lib/match/runner.rb
```diff
@@ -107,16 +120,44 @@ def run(params)
...
+          # method, as the value won't change. This way the warning is only printed once
+          UI.important("Looks like you run `match` in `readonly` mode, and didn't provide a `team_id`. This will still work, however it is recommended to provide a `team_id` in your Appfile or Matchfile")
+          @_folder_prefix = "*"
+        end
+        return File.join(working_directory, @_folder_prefix)
+      end
+    end
+
+    # Be smart about optional values here
+    # Depending on the storage mode, different vlaues are required
```
snatchev: "values"

match/lib/match/runner.rb
```diff
@@ -107,16 +120,44 @@ def run(params)
...
+    def update_optional_values_depending_on_storage_type(params)
+      if params[:storage_mode] != "git"
+        params.option_for_key(:git_url).optional = true
+      end
+    end
+
     def fetch_certificate(params: nil, working_directory: nil)
       cert_type = Match.cert_type_sym(params[:type])
 
-      certs = Dir[File.join(working_directory, "certs", cert_type.to_s, "*.cer")]
```
snatchev: we keep using `prefixed_working_directory(working_directory)` here. Maybe we can Dir.chdir to it? Stick it in a local?

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -0,0 +1,245 @@
...
+        end
+        finished_pushing_message
+      end
+
+      def upload_files(files_to_upload: [], custom_message: nil)
+        # `files_to_upload` is an array of files that need to be uploaded to Google Cloud
+        # Those doesn't mean they're new, it might just be they're changed
+        # Either way, we'll upload them using the same technique
+
+        files_to_upload.map do |current_file|
```
snatchev: this map is not returned by `upload_files` Just raising it up. If that's intentionally than maybe `each` would be more appropriate.

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -0,0 +1,245 @@
...
+      def initialize(type: nil,
+                     platform: nil,
+                     google_cloud_bucket_name: nil,
+                     google_cloud_keys_file: nil)
+        self.type = type if type
+        self.platform = platform if platform
+        self.google_cloud_keys_file = google_cloud_keys_file
+        self.bucket_name = google_cloud_bucket_name
+
+        ensure_keys_file_exists
```
snatchev: avoid functions that have mysterious side-effects (no inputs, no outputs).

I would recommend doing something like:
```ruby
this.google_cloud_keys_file = ensure_keys_file_exists
```

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -0,0 +1,245 @@
...
+            credentials: self.google_cloud_keys_file,
+            project_id: project_id
+          )
+        rescue => ex
+          UI.error(ex)
+          UI.verbose(ex.backtrace.join("\n"))
+          UI.user_error!("Couldn't log into your Google Cloud account using the provided JSON file at path '#{File.expand_path(self.google_cloud_keys_file)}'")
+        end
+
+        ensure_bucket_is_selected
```
snatchev: same as above.

Perhaps can be clearer with `this.bucket_name = ensure_bucket_is_selected`

match/lib/match/migrate.rb
KrauseFx: In this case, we'd actually want an empty bucket 👍 While we do support multiple teams to be in one bucket, the most common setup is probably one bucket per team, as it allows more fine-grained control of permissions

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -0,0 +1,245 @@
...
+      end
+
+      def download
+        # Check if we already have a functional working_directory
+        return self.working_directory if @working_directory
```
snatchev: I am unclear if this is intentional. If we have set `@working_directory`, we do not perform a download? Perhaps there should be a warning to the user that the download was deemed unnecessary?

Also, this function does not return anything if the download is successful, but returns when the download was skipped. Is this return value used anywhere?

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -0,0 +1,245 @@
...
+      attr_accessor :type
+      attr_accessor :platform
+      attr_accessor :bucket_name
+      attr_accessor :google_cloud_keys_file
+
+      # Managed values
+      attr_accessor :gc_storage
+
+      # Append Google Cloud specific options to `Match::Options`
+      Match::Options.append_option(
```
joshdholtz: What is the reason for appending these here? 🤔 Might be difficult for another contributor to find these as other options are all put into `options.rb`. I think these would _probably_ fit better in `options.rb` but I would probably be okay with a comment in `options.rb`

joshdholtz (changes requested):
A few small requests comments on placing of options and other things 🙃

match/lib/match/runner.rb
KrauseFx: In my experience we should avoid changing the directory if possible, as it has caused issues in the past. In particular running `Dir.chdir` within another `Dir.chdir` causes issues, which is something we already do in `fastlane/runner.rb`

match/lib/match/storage/google_cloud_storage.rb
KrauseFx: will update

match/lib/match/storage/google_cloud_storage.rb
KrauseFx: Oh yah, good point on both. I think in the remaining _fastlane_ code base we made use of the `!` operator, as in `ensure_bucket_is_selected!` if it was a potentially blocking call. Agreed that your proposal is nicer. Will updat

match/lib/match/nuke.rb
```diff
@@ -76,6 +80,14 @@ def run(params, type: nil)
       end
     end
 
+    # Be smart about optional values here
+    # Depending on the storage mode, different vlaues are required
+    def update_optional_values_depending_on_storage_type(params)
+      if params[:storage_mode] != "git"
+        params.option_for_key(:git_url).optional = true
```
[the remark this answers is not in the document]
milch: +1 to that. 

I've thought that maybe one way to get around this is to set `available_options` to the common denominator of all storage modes (e.g. `type`, `app_identifier`, and so on) and then have the storage implementation add the options it needs to the list when it is configured. 

I think the thing that's not nice about that is the user doesn't get the options on `--help` in the list, but rather has to consult the docs out of bounds to get that information. On the other hand it might be confusing when to set what parameters depending on the storage you're using, so they'd need to consult the docs anyway. That'd be better still than confusing/overloaded or extra parameter names

match/lib/match/runner.rb
```diff
@@ -32,7 +42,9 @@ def run(params)
         git_user_email: params[:git_user_email],
         clone_branch_directly: params[:clone_branch_directly],
         type: params[:type].to_s,
-        platform: params[:platform].to_s
+        platform: params[:platform].to_s,
+        google_cloud_bucket_name: params[:google_cloud_bucket_name].to_s,
+        google_cloud_keys_file: params[:google_cloud_keys_file].to_s
```
milch: Why not pass in params like `Storage.for_mode(params[:storage_mode], **params)`? 

With this method plugins can't provide their own storage mode that has custom parameters

milch (changes requested):

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -0,0 +1,245 @@
...
+  module Storage
+    # Store the code signing identities in on Google Cloud Storage
+    class GoogleCloudStorage < Interface
+      DEFAULT_KEYS_FILE_NAME = "gc_keys.json"
+
+      # User provided values
+      attr_accessor :type
+      attr_accessor :platform
+      attr_accessor :bucket_name
+      attr_accessor :google_cloud_keys_file
```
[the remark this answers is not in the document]
KrauseFx: Good point, will update

match/lib/match/storage/google_cloud_storage.rb
KrauseFx: Yep was a copy & paste error

match/lib/match/storage/google_cloud_storage.rb
KrauseFx: Updated

match/lib/match/options.rb
```diff
@@ -4,11 +4,17 @@
...
+      self.available_options # to ensure we created the initial `@available_options` array
+      @available_options << option
+    end
+
     def self.available_options
       user = CredentialsManager::AppfileConfig.try_fetch_value(:apple_dev_portal_id)
       user ||= CredentialsManager::AppfileConfig.try_fetch_value(:apple_id)
 
-      [
+      @available_options ||= [
```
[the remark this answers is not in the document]
KrauseFx: oh right, nice catch, thanks

match/lib/match/nuke.rb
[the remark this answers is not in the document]
KrauseFx: I like Manu's idea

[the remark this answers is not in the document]
KrauseFx: We can even use the `storage_mode` verify block to call the implementation's method that appends the options

match/lib/match/migrate.rb
```diff
@@ -0,0 +1,97 @@
...
+  class Migrate
+    def migrate(args, options)
+      params = FastlaneCore::Configuration.create(Match::Options.available_options, options.__hash__)
+      loaded_matchfile = params.load_configuration_file("Matchfile")
+
+      ensure_parameters_are_valid(params)
+
+      # We init the Google storage client before the git client
+      # to ask for all the missing inputs *before* cloning the git repo
+      google_cloud_storage = Storage.for_mode("google_cloud", {
```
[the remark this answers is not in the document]
KrauseFx: There isn't, but for now that's the one and only use-case. We have a bunch of follow-up tasks https://github.com/fastlane/fastlane/issues/13772

match/lib/match/nuke.rb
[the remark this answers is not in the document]
milch: 👍 Sounds good to me. I'd add something like a `base_options` method to match's `options.rb` and then in the `storage_mode` block set the available options to `base_options + GoogleCloudStorage.available_options`

match/lib/match/options.rb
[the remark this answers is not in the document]
KrauseFx: @milch that's kind of what I came up with: in the `commands_generator.rb`, something like
```ruby
            params = FastlaneCore::Configuration.create(Match::Options.available_options, options.__hash__)
            params.load_configuration_file("Matchfile") # this has to be done *before* overwriting the value
            additional_options = Storage.for_mode(params[:storage_mode], {}).additional_options

            params = FastlaneCore::Configuration.create(
              Match::Options.available_options + additional_options, 
              options.__hash__
            )
            params.load_configuration_file("Matchfile") # this has to be done *before* overwriting the value
```
Will think more about it later today/tomorrow

KrauseFx:
Ok, so this change would then cause the following issue, where commander fails before we can even load the `additional_options` based on the implementation...

```
$ be fastlane match --git_url https://github.com/krausefx/certs --verbose
[✔] 🚀
invalid option: --git_url
```

KrauseFx:
Alright, created a new branch for this with https://github.com/fastlane/fastlane/pull/13830, however it will require larger refactors, and maybe even a change in the `commander` gem, so I'll delay this to after merging this PR, as it's already a big one

KrauseFx:
so the last blocker to get this merged is to not cache the `available_options`

match/lib/match/migrate.rb
```diff
@@ -0,0 +1,97 @@
...
+      spaceship = SpaceshipEnsure.new(params[:username], params[:team_id], params[:team_name])
+      team_id = spaceship.team_id
+      UI.message("Detected team ID '#{team_id}' to use for Google Cloud Storage...")
+
+      files_to_commit = []
+      Dir.chdir(git_storage.working_directory) do
+        Dir[File.join("**", "*")].each do |current_file|
+          next if File.directory?(current_file)
+
+          to_path = File.join(google_cloud_storage.working_directory, team_id, current_file)
```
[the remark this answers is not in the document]
milch: Assuming this also gets done in #13772

match/lib/match/encryption.rb
```diff
@@ -11,6 +11,9 @@ def backends
             # to keychain_name for the name of the keychain entry
             params[:keychain_name] = params[:git_url]
             return Encryption::OpenSSL.configure(params)
+          },
+          "google_cloud" => lambda { |params|
```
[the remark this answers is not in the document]
milch: Low priority 👍

milch (approved):

joshdholtz (approved):

PR #13646, opened 2018-11-06
[match] Refactor to storage/encryption interfaces

milch:
- Make it possible to dynamically change backends, e.g. so that plugins can provide their own
- Refactor storage interface so common operations on file upload aren’t repeated

Tested by creating an example backend. 

Note that I can't seem to run the tests, so I'll rely on the CI to do that. Some unrelated tests fail, I think potentially due to the AV installed on the corporate Mac because the AV is going crazy when running the tests.

### DIFF
--- a/match/lib/match/storage.rb
+++ b/match/lib/match/storage.rb
@@ -3,13 +3,35 @@ require_relative 'storage/git_storage'
 
 module Match
   module Storage
-    def self.for_mode(storage_mode, params)
-      if storage_mode == "git"
-        return Storage::GitStorage.configure(params)
-      elsif storage_mode == "google_cloud"
-        # return Storage::GoogleCloudStorage
-      else
-        UI.user_error!("Invalid storage mode '#{storage_mode}'")
+    class << self
+      def backends
+        @backends ||= {
+          "git" => lambda { |params|
+            return Storage::GitStorage.configure(params)
+          }
+        }
+      end
+
+      def register_backend(type: nil, storage_class: nil, &configurator)
+        UI.user_error!("No type specified for storage backend") if type.nil?
+
+        normalized_name = type.to_s
+        UI.message("Replacing Match::Encryption backend for type '#{normalized_name}'") if backends.include?(normalized_name)
+
+        if configurator
+          @backends[normalized_name] = configurator
+        elsif storage_class
+          @backends[normalized_name] = ->(params) { return storage_class.configure(params) }
+        else
+          UI.user_error!("Specify either a `storage_class` or a configuration block when registering a storage backend")
+        end
+      end
+
+      def for_mode(storage_mode, params)
+        configurator = backends[storage_mode.to_s]
+        return configurator.call(params) if configurator
+
+        UI.user_error!("No storage backend for storage mode '#{storage_mode}'")
       end
     end
   end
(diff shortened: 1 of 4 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

milch:
Ready for review 🎉

match/lib/match/storage/git_storage.rb
```diff
@@ -223,6 +185,26 @@ def add_user_config(user_name, user_email)
           end
         end
       end
+
+      private # rubocop:disable Lint/UselessAccessModifier
+
+      def git_push(commands: [], commit_message: nil)
+        commit_message ||= generate_commit_message
```
armcburney: Instead of doing an `||=` here, could you make `generate_commit_message` be the default value for `commit_message` please?

```ruby
def git_push(commands: [], commit_message: generate_commit_message)
  # ...
end
```

match/lib/match/storage.rb
```diff
@@ -3,13 +3,35 @@
...
+        if configurator
+          @backends[normalized_name] = configurator
+        elsif storage_class
+          @backends[normalized_name] = ->(params) { return storage_class.configure(params) }
+        else
+          UI.user_error!("Specify either a `storage_class` or a configuration block when registering a storage backend")
+        end
+      end
+
+      def for_mode(storage_mode, params)
```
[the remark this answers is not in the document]
milch: The names `for_mode` and `for_storage_mode` were taken over from how they were named previously

I thought about refactoring this into a module, but thought it might make what's happening here non-obvious, so going by principle of least surprise decided to violate DRY. If you have a better way or think that having this refactored out would make more sense, I'm open to changing it

match/lib/match/storage/git_storage.rb
milch: That wouldn't work - `commit_message` already has a `nil` default in the methods up the call stack, it would end up being passed with `nil` here, meaning `generate_commit_message` wouldn't be called in your example. In pry:

```
[1] pry(main)> def a(test: nil)
[1] pry(main)*   b(test: test)
[1] pry(main)* end
=> :a
[2] pry(main)> def b(test: c)
[2] pry(main)*   puts test
[2] pry(main)* end
=> :b
[3] pry(main)> def c
[3] pry(main)*   "Hello"
[3] pry(main)* end
=> :c
[4] pry(main)> a

=> nil
```

match/lib/match/storage.rb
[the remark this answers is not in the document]
armcburney: Sounds like a good reason to me, makes sense 👍:)

match/lib/match/storage/git_storage.rb
armcburney: Oh, oops my bad. I didn't think about that. 👍

armcburney (approved):
Nice work @milch! This looks good to me! 🎉

match/lib/match/storage/interface.rb
```diff
@@ -37,7 +41,41 @@ def download
...
+            UI.user_error!("Neither `files_to_commit` nor `files_to_delete` were provided to the `save_changes!` method call")
+          end
+        end
+      end
+
+      def upload_files(files_to_upload: [], custom_message: nil)
+        not_implemented(__method__)
+      end
+
+      def delete_files(files_to_delete: [], custom_message: nil)
```
KrauseFx: Oh this is really nice

KrauseFx (approved):
This is beautiful, thanks for taking the time to build this 👍 💯

PR #13671, opened 2018-11-12
[match] fixes device count in match runner

jasdev:
### Checklist

### Description

Currently (tested with version `2.108.0`), when passing `force_for_new_devices` into `match`, it determines if profiles should be renewed if `Spaceship.device.all.count` differs from the target instance of `Space.Portal.ProvisioningProfile` device count. There is a subtle bug here in that `Spaceship.device.all.count` includes _all_ devices, even those outside the realm of what the user intended via the `platform` parameter.

This commit makes the counting logic a tad smarter, to avoid unnecessary profile renewals 🎉

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -166,7 +166,7 @@ module Match
 
       if params[:force_for_new_devices] && !params[:readonly]
         if prov_type != :appstore
-          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_s) unless params[:force]
+          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_sym) unless params[:force]
         else
           # App Store provisioning profiles don't contain device identifiers and
           # thus shouldn't be renewed if the device count has changed.
@@ -241,11 +241,11 @@ module Match
 
         portal_device_count =
           case platform
-          when :ios.to_s
+          when :ios
             Spaceship.device.all_ios_profile_devices.count
-          when :tvos.to_s
+          when :tvos
             Spaceship.device.all_apple_tvs.count
-          when :mac.to_s
+          when :mac
             Spaceship.device.all_macs.count
           else
             Spaceship.device.all.count
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -238,7 +238,20 @@ module Match
 
       if portal_profile
         profile_device_count = portal_profile.devices.count
-        portal_device_count = Spaceship.device.all.count
+
+        platform = params[:platform].to_s
+        portal_device_count =
+          case platform
+          when :ios.to_s
+            Spaceship.device.all_ios_profile_devices.count
+          when :tvos.to_s
+            Spaceship.device.all_apple_tvs.count
+          when :mac.to_s
+            Spaceship.device.all_macs.count
+          else
+            Spaceship.device.all.count
+          end
+
         return portal_device_count != profile_device_count
       end
       return false
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -166,7 +166,7 @@ module Match
 
       if params[:force_for_new_devices] && !params[:readonly]
         if prov_type != :appstore
-          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path) unless params[:force]
+          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_s) unless params[:force]
         else
           # App Store provisioning profiles don't contain device identifiers and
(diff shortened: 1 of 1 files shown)


jasdev:
CLA signed!

janpio:
Oh, good catch!

Can you explain a bit in exactly what case this was renewing to many profiles until now?
Did you stumble over any tests for this logic that could be improved to add this as well?

jasdev:
@janpio: Thanks 🙏🏽

> Can you explain a bit in exactly what case this was renewing to many profiles until now?

Our CI setup at Peloton has a step where we call `match` with `force_for_new_devices` set to `true` (for a purely iOS project). The bug wasn't an issue until now because we only had iOS/watchOS devices in our profiles, but _recently_ added two tvOS devices. This lead to `device_count_different?` always returning true, since it was over counting by two via `Spaceship.device.all.count`. 

> Did you stumble over any tests for this logic that could be improved to add this as well?

I took a glance at the existing tests and none exercise the `force_for_new_devices` path. Have any suggestions on how to proceed? The core of the test would be making sure [this switch](https://github.com/jasdev/fastlane/blob/5c8584bc0b7b83d25275debfff39656915b62eaf/match/lib/match/runner.rb#L242-L252) is correct and I'm not sure how valuable such a test would be. I'd be glad to add one, but since this helper is in `match`'s runner at the moment, it requires a lot of stubbing/mocking to drill down into the aforementioned switch expression 😌

janpio (approved):
Yeah, that doesn't sound like very useful tests. 

But the code is good ♥️

match/lib/match/runner.rb
```diff
@@ -166,7 +166,7 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
 
       if params[:force_for_new_devices] && !params[:readonly]
         if prov_type != :appstore
-          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path) unless params[:force]
+          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_s) unless params[:force]
```
joshdholtz: Can we remove this `.to_s` here? And all the other `.to_s`s inside of the case below? 🤔 Should be able to case on just the symbols 👌

joshdholtz (changes requested):

match/lib/match/runner.rb
jasdev: @joshdholtz: I'm mostly writing Swift these days—so please keep any and all feedback coming!

Regarding your comment, not sure we can drop the string conversion. I tried the following and all matching squashed down to the `else` clause of the switch. Any suggestions?

```patch
diff --git a/match/lib/match/runner.rb b/match/lib/match/runner.rb
index 5f7c7eacd..e3fdeba19 100644
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -166,7 +166,7 @@ module Match

         if prov_type != :appstore
-          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_s) unless params[:force]
+          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform]) unless params[:force]
         else
           # App Store provisioning profiles don't contain device identifiers and
           # thus shouldn't be renewed if the device count has changed.
@@ -241,11 +241,11 @@ module Match

         portal_device_count =
           case platform
-          when :ios.to_s
+          when :ios
             Spaceship.device.all_ios_profile_devices.count
-          when :tvos.to_s
+          when :tvos
             Spaceship.device.all_apple_tvs.count
-          when :mac.to_s
+          when :mac
             Spaceship.device.all_macs.count
           else
             Spaceship.device.all.count
```

jasdev:
@joshdholtz: Updated to prefer switching on symbols 👌🏽

match/lib/match/runner.rb
```diff
@@ -166,7 +166,7 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
 
       if params[:force_for_new_devices] && !params[:readonly]
         if prov_type != :appstore
-          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_s) unless params[:force]
+          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_sym) unless params[:force]
```
janpio: Ugh, interesting one! So the param is a string and has to be converted to a symbol?

jasdev: Seems like it—wasn't able to get the switch to work without the `to_sym` call (kept hitting the else clause)

joshdholtz (approved):
👍 Thanks for making this change!

KrauseFx:
What is this sound? The 🚂 train leaving for production?

PR #13674, opened 2018-11-13
Testfairy upload timeout

vijaysharm:
### Checklist

### Motivation and Context

While in Australia, I could not upload my app to TestFairy because the request would often timeout. With this parameter, I can upload my apps even if I'm in the middle of no where with a bad connection.

### Description

I've added a timeout parameter to the TestFairy Action. This will allow users who are very far away from TestFairy servers to actually upload their apps before the timeout kicks in.

I used Charles Proxy to throttle my connection so that I could see my timeout value being used.

### DIFF
--- a/fastlane/lib/fastlane/actions/testfairy.rb
+++ b/fastlane/lib/fastlane/actions/testfairy.rb
@@ -5,7 +5,7 @@ module Fastlane
     end
 
     class TestfairyAction < Action
-      def self.upload_build(upload_url, ipa, options)
+      def self.upload_build(upload_url, ipa, options, timeout)
         require 'faraday'
         require 'faraday_middleware'
 
@@ -29,7 +29,7 @@ module Fastlane
 
         begin
           connection.post do |req|
-            req.options.timeout = options.delete(:timeout)
+            req.options.timeout = timeout
             req.url("/api/upload/")
             req.body = options
           end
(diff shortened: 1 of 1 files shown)


vijaysharm:
I signed it!

janpio (approved):

vijaysharm:
CLA signed!

fastlane/lib/fastlane/actions/testfairy.rb
```diff
@@ -29,6 +29,7 @@ def self.upload_build(upload_url, ipa, options)
 
         begin
           connection.post do |req|
+            req.options.timeout = options.delete(:timeout)
```
joshdholtz: Instead of doing `options.delete(:timeout)`, can we just pass `timeout` as a parameter into this method?

fastlane/lib/fastlane/actions/testfairy.rb
```diff
@@ -89,6 +90,8 @@ def self.run(params)
             ['auto-update', value]
           when :notify
             [key, value]
+          when :timeout
+            [key, value]
```
joshdholtz: Instead of mapping `:timeout` here, can we reject `:timeout` like we do with `:upload_url`? `:timeout` isn't a client option so it shouldn't be mapped in `client_options` 🙃

And then below here call `self.upload_build` like...
```rb
self.upload_build(params[:upload_url], path, client_options, params[:timeout])
```

joshdholtz (changes requested):
Requesting a small rework otherwise this is 👍

vijaysharm:
Thanks @joshdholtz! Made the changes you asked for.

joshdholtz (approved):
Looks good! Thanks for making those changes 😊

joshdholtz (dismissed):
Ah, just noticed a test is failing actually! Can you fix? 😊

janpio:
I'll take this, that's code I added. Strange things happening there.

joshdholtz (approved):
@janpio Thanks for fixing!

PR #13692, opened 2018-11-15
Improve AdbHelper

janpio:
### Checklist

### Motivation and Context
While using `adb_devices` on Windows on CI (Azure Pipelines) I encountered several problems: The generated ADB path had both `/` and `\`, the generated command wasn't properly escaped. While debugging this, I also noticed wrong test description.

### Description
- Properly build the adb_path
- Apply .shellescape on parts of the command
- Better describe what the test actually does and expects

### DIFF
--- a/fastlane/lib/fastlane/helper/adb_helper.rb
+++ b/fastlane/lib/fastlane/helper/adb_helper.rb
@@ -18,7 +18,7 @@ module Fastlane
       def initialize(adb_path: nil)
         android_home = ENV['ANDROID_HOME'] || ENV['ANDROID_SDK_ROOT'] || ENV['ANDROID_SDK']
         if (adb_path.nil? || adb_path == "adb") && android_home
-          adb_path = Pathname.new(android_home).join("platform-tools/adb").to_s
+          adb_path = File.join(android_home, "platform-tools", "adb")
         end
         self.adb_path = adb_path
       end
--- a/fastlane/lib/fastlane/helper/adb_helper.rb
+++ b/fastlane/lib/fastlane/helper/adb_helper.rb
@@ -26,7 +26,7 @@ module Fastlane
       # Run a certain action
       def trigger(command: nil, serial: nil)
         android_serial = serial != "" ? "ANDROID_SERIAL=#{serial}" : nil
-        command = [android_serial, adb_path, command].join(" ")
+        command = [android_serial, adb_path.shellescape, command.shellescape].join(" ")
         Action.sh(command)
       end
 
@@ -38,7 +38,7 @@ module Fastlane
       def load_all_devices
         self.devices = []
 
-        command = [adb_path, "devices"].join(" ")
+        command = [adb_path.shellescape, "devices"].join(" ")
         output = Actions.sh(command, log: false)
         output.split("\n").each do |line|
           if (result = line.match(/(.*)\tdevice$/))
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/helper/adb_helper.rb
```diff
@@ -38,7 +38,7 @@ def device_avalaible?(serial)
       def load_all_devices
         self.devices = []
 
-        command = [adb_path, "devices"].join(" ")
+        command = [adb_path.shellescape, "devices"].join(" ")
```
hjanuschka: how about doing the `.shellescape` only once in `initialize`?

hjanuschka (changes requested):
LGTM - left one comment. 🚀

fastlane/lib/fastlane/helper/adb_helper.rb
janpio: I decided against that because one could theoretically use the `adb_path` somewhere outside of the context of a command (e.g. for output, logging etc), and then you would not want it to be shellescaped. Afaik shellescaping is only useful if used in a command (on the shell...), so I thought we should not _always_ do that. Agree?

hjanuschka: good point.

hjanuschka (approved):

tmaly1980:
Sorry, what does the shellescape do to help? I'm suddenly getting all of my adb commands failing, simply printing out the help text for the adb command.... 

```
adb(
   command: "shell settings put global window_animation_scale 0.0"
)
```

Ends up calling "adb shell " (with a space at the end, which is not a valid command, so it returns the help screen):

```/Users/zingfitdev2/Library/Android/sdk/platform-tools/adb shell\ settings\ put\ global\ window_animation_scale\ 0.0```

Which then just does:

```
[22:59:20]: ▸ Android Debug Bridge version 1.0.39
[22:59:20]: ▸ Revision 3db08f2c6889-android
[22:59:20]: ▸ Installed as /Users/zingfitdev2/Library/Android/sdk/platform-tools/adb
[22:59:20]: ▸ global options:
[22:59:20]: ▸ -a         listen on all network interfaces, not just localhost
[22:59:20]: ▸ -d         use USB device (error if multiple devices connected)
[22:59:20]: ▸ -e         use TCP/IP device (error if multiple TCP/IP devices available)
[22:59:20]: ▸ -s SERIAL
[22:59:20]: ▸ use device with given serial number (overrides $ANDROID_SERIAL)
[22:59:20]: ▸ -p PRODUCT
[22:59:20]: ▸ name or path ('angler'/'out/target/product/angler');
[22:59:20]: ▸ default $ANDROID_PRODUCT_OUT
[22:59:20]: ▸ -H         name of adb server host [default=localhost]
[22:59:20]: ▸ -P         port of adb server [default=5037]
[22:59:20]: ▸ -L SOCKET  listen on given socket for adb server [default=tcp:localhost:5037]
[22:59:20]: ▸ general commands:
[22:59:20]: ▸ devices [-l]             list connected devices (-l for long output)
[22:59:20]: ▸ help                     show this help message
[22:59:20]: ▸ version                  show version num
```

What am I doing wrong?

tmaly1980 (commented):
This unfortunately completely breaks the adb command. Any commands that take parameters (such as "adb shell ls", and most other commands) will fail, because it puts a backslash before the space, and thinks the command is "shell\ " rather than "shell"

janpio:
Ugh, thanks for reporting this - I will look into it asap.

I created an issue at https://github.com/fastlane/fastlane/issues/13845 where we can discuss.

PR #13703, opened 2018-11-16
[WIP] Operating System incompatibility of actions

### LINKED ISSUE #11859: [Discuss] Operating System compatibility of tools and actions

janpio:
### Checklist

### Motivation and Context

Some actions and tools just can't work on other operating systems than macOS: Xcode and similar tools only run on macOS, there is no way to get them to work on Windows on Linux. In the tests we cover this by disabling the tests on those platforms and make them "pending", but for the actual execution of actions and commands this doesn't exist yet.

### Description

This PR adds a method to actions that is internally used to find out if an action works on a specific operating system or not. The developer can specify which OSes an action does **not** work on.

#### Content (and TODO)

- [x] Add method stub `is_incompatible?` to action templates (and clean them up along the way)
- [x] Implement `verify_compatible_os` method in fastlane runner that is called before execution of an action and checks compatibility via the action's `is_compatible` method
  - [x] Throw exception (`UI.user_error!`) if incompatible
  - [x] Replace exception with error (`UI.error`) if env var `FASTLANE_IGNORE_OS_INCOMPAT` is set
  - [x] Write tests for `verify_compatible_os`
  - [x] Also disable exceptions if tests are running (to not mess with other, existing tests that don't like an exception) and make sure the actual tests do hide there test-iness
- [x] Output OS compatibility in action documentation
- [ ] Mark all affected actions as incompatible
- [ ] Also handle commands/tools that are executed on the command line directly
- [ ] Add link to exception for documentation on ways to make `Fastfile` cross platform

##### Potential
- [ ] Output an incompatibility reason similar to deprecation information in action docs (exception and docs)

##### Semi related

- [x] Remove leftover/outdated `requires_xcodeproj: true` flags in `get_version_number_action_spec` tests
- [x] Clean up Helper.operating_system (and reorder methods)

---
### Questions

- Did I cover this with tests well enough? I thought about also testing with something like `result = Fastlane::FastFile.new.parse("lane :test do scan end").runner.execute(:test)` to make sure it also actually gets triggered when a lane is run, and not only when the `verify_compatible_os` method is called directly, but didn't really know where to put these test :/ Would they just repeat the same testing as the ones I added?

---
closes #11859

### DIFF
--- a/fastlane/lib/assets/ActionDetails.md.erb
+++ b/fastlane/lib/assets/ActionDetails.md.erb
@@ -20,7 +20,7 @@ To modify it, go to its source at https://github.com/fastlane/fastlane/blob/mast
 <%= action.action_name %> ||
 ---|---
 Supported platforms | <%= [:ios, :android, :mac].find_all { |a| action.is_supported?(a) }.join(", ") %>
-Compatible operating systems | <%= ['MacOS', 'Linux', 'Windows'].find_all { |a| !action.is_incompatible?(a) }.join(", ") %>
+Compatible operating systems | <%= ['MacOS', 'Linux', 'Windows'].delete_if { |a| action.respond_to?(:is_incompatible?) && action.is_incompatible?(a) }.join(", ") %>
 Author | @<%= Array(action.author || action.authors).join(", @") %>
 <%- unless action.return_value.to_s.empty? -%>
 Returns | <%= action.return_value.gsub("\n", "<br>") %>
--- a/fastlane/lib/assets/ActionDetails.md.erb
+++ b/fastlane/lib/assets/ActionDetails.md.erb
@@ -20,6 +20,7 @@ To modify it, go to its source at https://github.com/fastlane/fastlane/blob/mast
 <%= action.action_name %> ||
 ---|---
 Supported platforms | <%= [:ios, :android, :mac].find_all { |a| action.is_supported?(a) }.join(", ") %>
+Compatible operating systems | <%= ['MacOS', 'Linux', 'Windows'].find_all { |a| !action.is_incompatible?(a) }.join(", ") %>
 Author | @<%= Array(action.author || action.authors).join(", @") %>
 <%- unless action.return_value.to_s.empty? -%>
 Returns | <%= action.return_value.gsub("\n", "<br>") %>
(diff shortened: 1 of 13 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

iBotPeaches:
This would be cool to revisit, but its not been touched since 2019 and I'm trying to do a solid cleanup to make this more manageable on the remaining folks instead of looking at hundreds of issues/prs daily. Closing as decayed out.

PR #13724, opened 2018-11-20
Improve Spaceship API documentation

janpio:
Spaceship has API documentation for both "targets", App Store Connect and the Developer Portal. I fixed the headline levels and added table of contents (using https://ecotrust-canada.github.io/markdown-toc/) to make it easier to get an overview.


KrauseFx (approved):
This is great 👍

PR #13727, opened 2018-11-20
[docs] add id to plugin headline so it becomes linkable

janpio:
This adds an ID to the plugin headline so it becomes linkable via an anchor `#pluginname`.

### DIFF
--- a/fastlane/assets/render_plugin.md.erb
+++ b/fastlane/assets/render_plugin.md.erb
@@ -2,7 +2,7 @@
 
 <div class="quality-index" style="color: #<%= @plugin.color_to_use %>"><%= @plugin.data[:overall_score] %></div>
 
-<h3><a href="<%= @plugin.homepage %>" target="_blank"><code><%= @plugin.name.gsub("fastlane-plugin-", "") %></code></a></h3>
+<h3 id="<%= @plugin.name.gsub("fastlane-plugin-", "") %>"><a href="<%= @plugin.homepage %>" target="_blank"><code><%= @plugin.name.gsub("fastlane-plugin-", "") %></code></a></h3>
 
 <p class="via-text">via <%= @plugin.raw_hash["authors"] %></p>
 


fastlane/assets/render_plugin.md.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="quality-index" style="color: #<%= @plugin.color_to_use %>"><%= @plugin.data[:overall_score] %></div>
 
-<h3><a href="<%= @plugin.homepage %>" target="_blank"><code><%= @plugin.name.gsub("fastlane-plugin-", "") %></code></a></h3>
+<h3 id="<%= @plugin.name.gsub("fastlane-plugin-", "") %>"><a href="<%= @plugin.homepage %>" target="_blank"><code><%= @plugin.name.gsub("fastlane-plugin-", "") %></code></a></h3>
```
joshdholtz: Is it preferred to make this a `name` attribute on the anchor? This way it will update the navigation bar to have the named anchor in the url when clicked? 🤔

```suggestion
<h3><a  name="<%= @plugin.name.gsub("fastlane-plugin-", "") %>" href="<%= @plugin.homepage %>" target="_blank"><code><%= @plugin.name.gsub("fastlane-plugin-", "") %></code></a></h3>
```

joshdholtz (commented):
Like this addition! Just one small question

fastlane/assets/render_plugin.md.erb
janpio: Does it do that? I think this only happens if it is an empty anchor with `href="#"` or similar, not for an actual link that open a site.

joshdholtz: Ahhhhhhh, good point! Your implementation is 👍

joshdholtz (approved):
This looks good 👍

fastlane/assets/render_plugin.md.erb
janpio: Hmm, for normal headlines we get his "header anchor" that appears left to the actual headline. This is the code for it:
image
But I couldn't find out where this comes from really - but it would be perfect to be able to hotlink to the individual plugins as well!

janpio: Oh, it's an extension to mkdocs:
https://github.com/fastlane/docs/search?q=headeranchor&unscoped_q=headeranchor

So why isn't this applied to the `h3` headline of the plugins?

PR #13752, opened 2018-11-22
[spaceship] support --verbose for `fastlane spaceauth` 

### LINKED ISSUE #13750: [spaceship] `fastlane spaceauth` is missing support for `--verbose`
```
λ bundle exec fastlane spaceauth --verbose
[✔] 🚀
invalid option: --verbose
```

It might be useful to have `verbose` support here as well to be able to get verbose output.

janpio:
`fastlane spaceauth` didn't support `--verbose`, which is handy to get verbose output from spaceship when stuff is going wrong.

closes #13750


KrauseFx (approved):
Yep, nice catch 👍

PR #13754, opened 2018-11-23
 [precheck] Allow mentioning Google Drive

douglashill:
### Checklist

### Motivation and Context

This fixes a false positive in `precheck`. It is reasonable that apps might integrate with Google Drive and users would want to know about this from the App Store description. Our app PDF Viewer has been approving mentioning Google Drive.

### Description

I added an additional entry to the whitelist for mentioning other platforms. Google is on the blacklist because it is a competitor to Apple, then Google Analytics and Google Drive are whitelisted because iOS app may integrate with these services.

I added tests for `Precheck::OtherPlatformsRule`, just covering a one allowed case and one not allowed case. I don’t really know Ruby or any of the rig here so I copied and modified `curse_words_rule_spec.rb`.

### DIFF
--- a/precheck/spec/rules/other_platforms_rule_spec.rb
+++ b/precheck/spec/rules/other_platforms_rule_spec.rb
@@ -1,0 +1,21 @@
+require 'precheck'
+
+module Precheck
+  describe Precheck do
+    describe Precheck::OtherPlatformsRule do
+      let(:rule) { OtherPlatformsRule.new }
+      let(:allowed_item) { TextItemToCheck.new("We have integration with the Files app so you can open documents stored in your Google Drive.", :description, "description") }
+      let(:forbidden_item) { TextItemToCheck.new("Google is really great.", :description, "description") }
+
+      it "passes for allowed text" do
+        result = rule.check_item(allowed_item)
+        expect(result.status).to eq(VALIDATION_STATES[:passed])
+      end
+
+      it "fails for mentioning competitors" do
+        result = rule.check_item(forbidden_item)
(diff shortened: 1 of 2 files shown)


douglashill:
I signed it!

janpio:
Thanks for these changes!

Semi OT: Is mentioning "Google" actually a problem? "Android" I get, but the company in general?

douglashill:
I don’t know. I'm sceptical about whether most of these other platform checks are useful. Most of them seem irrelevant these days and unlikely to be used or be a problem. Android seems a good warning as that'a definitely caused some people trouble in the past. Google seems a bad warning to me. We already have two examples, and there are lots more Google services an iOS might mention reasonably. Or the app might be one made by Google itself :)

So if it might be accepted, I'd say +1 to just removing google from the blacklist.

precheck/spec/rules/other_platforms_rule_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'precheck'
+
+module Precheck
+  describe Precheck do
+    describe Precheck::OtherPlatformsRule do
+      let(:rule) { OtherPlatformsRule.new }
+      let(:allowed_item) { TextItemToCheck.new("We have integration with the Files app so you can open documents stored in your Google Drive.", :description, "description") }
+      let(:forbidden_item) { TextItemToCheck.new("Google is really great.", :description, "description") }
```
KrauseFx: 💯

KrauseFx (approved):
Perfect. I don't think Apple actually specified the exact rules around Google related products, so we can only add rules based on experiences by the community 👍 Let's go with this

PR #13765, opened 2018-11-26
Allow for the host of the Appetize API to be configured

steprescott:
### Checklist

### Motivation and Context
Appetize offer a hosted solution if you require dedicated instances. This solution then changes the host of the API from `api.appetize.io` to `<CUSTOM_API>.appetize.io`. This change will allow you to set a custom host that will be used when uploading the app.

### Description
A new `ConfigItem` has been added named `api_host`. This has a default value of `api.appetize.io` and this will cater for most cases and therefore results in a **none breaking change**. Now with this new item the host can be changed to support any hosted Appetize solution.

A new test case has been added that provides test coverage for a custom or default api host.

### DIFF
--- a/fastlane/lib/fastlane/actions/appetize.rb
+++ b/fastlane/lib/fastlane/actions/appetize.rb
@@ -172,7 +172,7 @@ module Fastlane
         [
           'appetize(
             path: "./MyApp.zip",
-            api_host: "company.appetize.io" # only needed for enterprise hosted solution
+            api_host: "company.appetize.io", # only needed for enterprise hosted solution
             api_token: "yourapitoken", # get it from https://appetize.io/docs#request-api-token
             public_key: "your_public_key" # get it from https://appetize.io/dashboard
           )'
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/appetize.rb
```diff
@@ -161,6 +172,7 @@ def self.example_code
         [
           'appetize(
             path: "./MyApp.zip",
+            api_host: "company.appetize.io", # only needed for enterprise hosted solution
```
janpio: Can we move this to a second example that uses it so the first one stays the default case? Thanks.

steprescott: Done :-)

janpio (approved):

janpio (approved):
(visual review)

joshdholtz (approved):
Looks good 👍 Thanks for the contribution!

PR #13766, opened 2018-11-26
Use `crash!` instead of `user_error!` on configuration access mis-use

KrauseFx:
Since those methods are mostly used internally and for plugins, this will make debugging easier, as the crash will show the full stack trace, instead of just the error message


joshdholtz (approved):
💪

Issue #13740, opened 2018-11-21
[resign.sh] Resign fails to generate valid entitlements


lyndsey-ferguson:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

We want to use the resign tool to resign customized versions of our app for our clients. We provide `resign` an unsigned ipa that has been prepared for resign in the following way:
1. We customize the `Info.plist's` `CFBundleIdentifier`, and `AppIdentifierPrefix`, and `CFBundleVersion`, and `CFBundleDisplayName`.
1. We modify the `App.entitlements` file in the following way:
    1. `aps-environment` is removed _if_ Push is to be turned off.
    1. `com.apple.developer.icloud-container-identifiers`, `com.apple.developer.ubiquity-container-identifiers`, `com.apple.developer.icloud-services` are removed _if_ iCloud is to be turned off.
    1. We replace the app's bundle IDs that are values for the above keys with the client's bundle ID (if those entries still remain in the entitlements file after no. 1 and no. 2).
1. We then call the `resign` action with the client's bundle ID.

The problem is that the resigned app is rejected by Apple for the following reasons:
1. `value 'Development' for key 'com.apple.developer.icloud-container-environment' in 'Payload/App.app/App' is not supported."`
1. `key 'com.apple.developer.icloud-container-development-container-identifiers' in 'Payload/App.app/App' is not supported."`
1. `value '*' for key 'com.apple.developer.icloud-services' in 'Payload/App.app/App' is not supported.`
1. `the key 'com.apple.developer.ubiquity-kvstore-identifier' in 'Payload/App.app/App' is not supported.`

Our app does not have these entries in the app binary, and has been successfully uploaded to the App Store.

Analyzing the `resign.sh` script, the code path goes through the [default conditional](https://github.com/fastlane/fastlane/blob/master/sigh/lib/assets/resign.sh#L798) to embed the entitlements from the provisioning profile into the application binary. The problem with that seems to be that the provisioning profile's entitlements are a "template" for what could be, not what "is" in the current app. For example:

```xml
<key>Entitlements</key>
<dict>
    <key>keychain-access-groups</key>
    <array>
[...]

PR #13767, opened 2018-11-26
[resign.sh] Fix icloud-container-environment value

ryanprobus:
### Checklist

### Motivation and Context

The `com.apple.developer.icloud-container-environment` entitlement was being incorrectly set to `development` when using app entitlements if passing in a SHA1 as the `CERTIFICATE` parameter. This behavior appears to have been introduced by this [pull request](https://github.com/fastlane/fastlane/pull/4898/files) which added the ability to pass the SHA1 instead of the certificate name.

Partially fixes this [issue](https://github.com/fastlane/fastlane/issues/13740) pertaining to entitlements when resigning.

### Description

When iterating over the entitlement transfer rules, there is a specific check for `com.apple.developer.icloud-container-environment`, within this check, I added a regex to check if `CERTIFICATE` matches a SHA1. If it does, I get the name of the identity by grepping `security find-identity`.

Used the below script to test various types of inputs including non-SHA1 inputs. Additionally, I resigned our IPAs using the fastlane `resign` action and checked the embedded entitlements. The entitlement was correctly set to `production` if the certificate's name contained `Distribution:`.

```
CERTIFICATE="CERTIFICATE_VALUE"
sha1_pattern='[0-9A-F]{40}'
if [[ "$CERTIFICATE" =~ $sha1_pattern ]]; then
  certificate_matches="$(security find-identity -v -p codesigning | grep -m 1 "$CERTIFICATE")"
  if [ -n "$certificate_matches" ]; then
    certificate_name="$( sed -E s/[^\"]+\"\([^\"]+\)\".*/\\1/ <<< $certificate_matches )"
    echo "$certificate_name"
  fi
fi
```

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -63,6 +63,9 @@
 # new features November 2018
 # 1. only create the archived-expanded-entitlements.xcent file if the version of Xcode < 9.3 as Xcode 10 does not create it.
 #
+# new features January 2019
+# 1. fixed bug where the com.apple.icloud-container-environment entitlement was being assigned an incorrect value
+#
 
 # Logging functions
 
@@ -743,7 +746,19 @@ function resign {
                 # This value is set by Xcode during export (manually selected for Development and AdHoc, automatically set to Production for Store)
                 # Would need an additional dedicated option to specify the iCloud environment to be used (Development or Production)
                 # For now, we assume Production is to be used when signing with a Distribution certificate, Development if not
-                if [[ "$CERTIFICATE" =~ "Distribution:" ]]; then
+                local certificate_name=$CERTIFICATE
+                local sha1_pattern='[0-9A-F]{40}'
+
+                if [[ "$CERTIFICATE" =~ $sha1_pattern ]]; then
+                    log "Certificate $CERTIFICATE matches a SHA1 pattern"
+                    local certificate_matches="$( security find-identity -v -p codesigning | grep -m 1 "$CERTIFICATE" )"
+                    if [ -n "$certificate_matches" ]; then
+                        certificate_name="$( sed -E s/[^\"]+\"\([^\"]+\)\".*/\\1/ <<< $certificate_matches )"
+                        log "Certificate name: $certificate_name"
+                    fi
+                fi
+
+                if [[ "$certificate_name" =~ "Distribution:" ]]; then
                     ICLOUD_ENV="Production"
                 else
                     ICLOUD_ENV="Development"


sigh/lib/assets/resign.sh
```diff
@@ -743,7 +743,19 @@ function resign {
                 # This value is set by Xcode during export (manually selected for Development and AdHoc, automatically set to Production for Store)
                 # Would need an additional dedicated option to specify the iCloud environment to be used (Development or Production)
                 # For now, we assume Production is to be used when signing with a Distribution certificate, Development if not
-                if [[ "$CERTIFICATE" =~ "Distribution:" ]]; then
+                local certificate_name=$CERTIFICATE
+                local sha1_pattern='[0-9A-F]{5,40}'
```
lyndsey-ferguson: this should be a straight 40

lyndsey-ferguson (changes requested):
one change requested

sigh/lib/assets/resign.sh
```diff
@@ -743,7 +743,19 @@ function resign {
...
+
+                if [[ "$CERTIFICATE" =~ $sha1_pattern ]]; then
+                    log "Certificate $CERTIFICATE matches a SHA1 pattern"
+                    local certificate_matches="$( security find-identity -v -p codesigning | grep "$CERTIFICATE" )"
```
kpink224: Let's do `grep -m 1 "$CERTIFICATE"` to ensure we only get one option. It will cause problems if we get multiple.

lyndsey-ferguson (approved):
LGTM

sigh/lib/assets/resign.sh
```diff
@@ -743,7 +743,19 @@ function resign {
...
                 # For now, we assume Production is to be used when signing with a Distribution certificate, Development if not
-                if [[ "$CERTIFICATE" =~ "Distribution:" ]]; then
+                local certificate_name=$CERTIFICATE
+                local sha1_pattern='[0-9A-F]{40}'
+
+                if [[ "$CERTIFICATE" =~ $sha1_pattern ]]; then
+                    log "Certificate $CERTIFICATE matches a SHA1 pattern"
+                    local certificate_matches="$( security find-identity -v -p codesigning | grep "$CERTIFICATE" )"
+                    if [ -n "$certificate_matches" ]; then
+                        certificate_name="$( sed -E s/[^\"]+\"\([^\"]+\)\".*/\\1/g <<< $certificate_matches )"
```
kpink224: you don't need `g` at the end of sed

kpink224 (commented):
LG just some minor changes

kpink224 (approved):

joshdholtz (changes requested):
@ryanprobus Sorry about just getting to this! I will admin that I am not a `resign.sh` expert but this looks like a solid change so 🙌🙌 to you for making this 😊

One request I have is if you could add to the change log at the top of the file since that is kind of what has been happening - https://github.com/fastlane/fastlane/blob/b188b9d9c9a3a739e875312a58fb2c008b23cff6/sigh/lib/assets/resign.sh#L42-L66

Thanks!

ryanprobus:
@joshdholtz No problem, thanks for taking a look at this. A lot of people are afraid of `resigh.sh`, but its definitely something could use some loving. I added a note to the change log and squashed the commit.

joshdholtz (approved):
@ryanprobus Thanks for adding that note and giving `resign.sh` the love that it needs 😊 You are a true hero 🙌

PR #13769, opened 2018-11-27
Improves parse error reporting for non-conventional setups

bricker:
### Checklist

### Motivation and Context

In some unconventional setups, the stringified error object doesn't match the format expected by the code and a useless `undefined method '[]' for nil:NilClass` error is thrown instead of the actual syntax error.

For example, in my Fastlane setup, I have multiple util files that I require into my `Fastfile`. This causes the `.match()` line to throw the nil error.

### Description

I've left in the original functionality for conventional setups. However, in the case where the stringified error doesn't match the given regular expression, the code will still throw a useful error.


bricker:
I signed it!

KrauseFx (approved):
This looks like a great change, thanks so much for taking the time working on this 💯

PR #13778, opened 2018-11-29
[spaceship] Use default sort order for listing TestFlight testers

tbodt:
### Checklist

### Motivation and Context
Apple's API is broken with email sort when dealing with thousands of anonymous (public link) testers. All anonymous testers have the same first name, last name, and email, so the sort order needs to be something other than those, and default seems to correspond to sorting by UUID.

With email sort order, it would return a page of results, then the "next" link would contain an offset starting with a comma (%2C). If you request that link, or change the offset in that link to anything else and try requesting that, it would return the first 40 testers sorted in reverseorder of UUID. I suspect this can be reproduced with any app with over 40 anonymous testers.

### Description
Some playing around in the spaceship REPL found that using default sort order fixed the problem. I tested it by exporting almost 10,000 anonymous testers from my app.

### DIFF
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -187,8 +187,8 @@ module Spaceship
         assert_required_params(__method__, binding)
         page_size = 40 # that's enforced by the iTC servers
         resulting_array = []
-        # Sort order is set to `default` instead of `email`, because any other sort order breaks
-        # pagination when dealing with lots of anonymous (public link) testers.
+        # Sort order is set to `default` instead of `email`, because any other sort order breaks pagination
+        # when dealing with lots of anonymous (public link) testers: https://github.com/fastlane/fastlane/pull/13778
         initial_url = "providers/#{team_id}/apps/#{app_id}/testers?limit=#{page_size}&sort=default&order=asc"
         response = request(:get, initial_url)
         link_from_response = proc do |r|
--- a/spaceship/lib/spaceship/test_flight/client.rb
+++ b/spaceship/lib/spaceship/test_flight/client.rb
@@ -187,7 +187,9 @@ module Spaceship
         assert_required_params(__method__, binding)
         page_size = 40 # that's enforced by the iTC servers
         resulting_array = []
-        initial_url = "providers/#{team_id}/apps/#{app_id}/testers?limit=#{page_size}&sort=email&order=asc"
+        # Sort order is set to `default` instead of `email`, because any other sort order breaks
+        # pagination when dealing with lots of anonymous (public link) testers.
+        initial_url = "providers/#{team_id}/apps/#{app_id}/testers?limit=#{page_size}&sort=default&order=asc"
         response = request(:get, initial_url)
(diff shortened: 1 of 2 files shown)


janpio:
Maybe add a comment why this sort order was chosen and how one could reproduce the error to see if it is still happening or the workaround can be removed?

tbodt:
@janpio Added a comment, let me know if I should explain more.

tbodt:
@janpio ping

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -187,7 +187,19 @@ def testers_for_app(app_id: nil)
         assert_required_params(__method__, binding)
         page_size = 40 # that's enforced by the iTC servers
         resulting_array = []
-        initial_url = "providers/#{team_id}/apps/#{app_id}/testers?limit=#{page_size}&sort=email&order=asc"
+        # Sort order is set to default, because any other sort order breaks
```
janpio:
```suggestion
        # Sort order is set to `default` instead of `email`, because any other sort order breaks
```

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -187,7 +187,19 @@ def testers_for_app(app_id: nil)
...
+        # pagination when dealing with lots of anonymous (public link) testers.
+        # All anonymous testers have the same first name, last name, and email,
```
janpio: Everything in this comment from this line can be in the PR description, and a link to the PR here for future reference. Better than this big comment here, sorry for the confusion I caused.

janpio:
Left some comments, but won't be able to really review this as I know nothing about Testflight testers. Someone else from the team will have to chip in.

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -187,7 +187,9 @@ def testers_for_app(app_id: nil)
         assert_required_params(__method__, binding)
         page_size = 40 # that's enforced by the iTC servers
         resulting_array = []
-        initial_url = "providers/#{team_id}/apps/#{app_id}/testers?limit=#{page_size}&sort=email&order=asc"
+        # Sort order is set to `default` instead of `email`, because any other sort order breaks
```
janpio:
```suggestion
        # Sort order is set to `default` instead of `email`, because any other sort order breaks pagination
```

spaceship/lib/spaceship/test_flight/client.rb
```diff
@@ -187,7 +187,9 @@ def testers_for_app(app_id: nil)
...
+        # pagination when dealing with lots of anonymous (public link) testers.
```
janpio:
```suggestion
        # when dealing with lots of anonymous (public link) testers: https://github.com/fastlane/fastlane/pull/13778
```

janpio (approved):
Looks good to me, someone who knows more about this API should double check though.
(And best someone who can replicate the actual issue as well!)

joshdholtz (approved):
Great find! This looks good to me 💪

PR #13782, opened 2018-11-29
Fix set Github release to accept an array for the uploadAssets folder

thedavidharris:
### Checklist

### Motivation and Context

Could not use the `setGithubRelease` function from Fastlane.swift due to bad translation from Swift to Ruby

### Description

Adds the `Array` type to the Ruby function parameter

Tested these changes on a current project attempting to use the this function in Fastlane.swift and it works correctly

### DIFF
--- a/fastlane/swift/Fastlane.swift
+++ b/fastlane/swift/Fastlane.swift
@@ -2962,7 +2962,7 @@ func setChangelog(appIdentifier: String,
                                          description: String? = nil,
                                          isDraft: Bool = false,
                                          isPrerelease: Bool = false,
-                                         uploadAssets: [String]? = nil) -> [String : String] {
+                                         uploadAssets: String? = nil) -> [String : String] {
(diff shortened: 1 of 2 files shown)


thedavidharris:
I signed it!

fastlane/swift/Fastlane.swift
```diff
@@ -2962,7 +2962,7 @@ func setChangelog(appIdentifier: String,
                                          description: String? = nil,
                                          isDraft: Bool = false,
                                          isPrerelease: Bool = false,
-                                         uploadAssets: String? = nil) -> [String : String] {
+                                         uploadAssets: [String]? = nil) -> [String : String] {
```
taquitos: This change will automatically happen at the next release when the release script regenerates the Swift implementation.

taquitos (changes requested):
Undo that `Fastlane.swift` stuff then it looks great 👍

fastlane/swift/Fastlane.swift
dharr419-ford: Ahh perfect

taquitos (approved):
🎈:goat: thanks!

Issue #13783, opened 2018-11-29
Scan improvement to use SlackAction


gligorkot:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
What tool/action do you have a question about?
**Scan**

#### Question Description
Hi fastlane team,

I've been looking at improving (making the scan/run_tests slack message more consistent) with the slack action output, so I've implemented a version that uses the `SlackAction.run(options)` inside `SlackPoster` of the `scan` action. However, I am not 100% proficient with ruby, so not sure how to do this right without causing a circular dependency. Can anyone please help me with what I need to `require` in `slack_poster.rb` to call `SlackAction.run(options)`?

I have a version that I've tested it works, but `bundle exec fastlane test` fails at the `rubocop` step with `1041 files inspected, 1 offense detected`

You can find my change here -> [https://github.com/gligorkot/fastlane/tree/improve_scan_slack_action](https://github.com/gligorkot/fastlane/tree/improve_scan_slack_action)

gligorkot:
Thanks @janpio, I've raised #13791 with the changes.

gligorkot:
Fixed in #13791

PR #13791, opened 2018-12-01
[scan] Use slack action to post scan test results

gligorkot:
### Checklist

### Motivation and Context
Hi team, this is the PR that addresses my question in #13783 

I've implemented the change as I'd like to have more consistent Slack messages when running test builds between iOS and Android, so using the Slack Action keeps them the same across the board.

### Description
Excuse the `rubocop` fails, I'm not very proficient with ruby, this is my first go at it. I think there's a cyclic dependency cause I call `Firebase` from within the `SlackPoster` class

### DIFF
--- a/scan/lib/scan/slack_poster.rb
+++ b/scan/lib/scan/slack_poster.rb
@@ -1,3 +1,6 @@
+require 'fastlane/action'
+require 'fastlane/actions/slack'
+
 require_relative 'module'
 
 module Scan
--- a/scan/lib/scan/slack_poster.rb
+++ b/scan/lib/scan/slack_poster.rb
@@ -12,56 +12,44 @@ module Scan
         channel = ('#' + channel) unless ['#', '@'].include?(channel[0]) # send message to channel by default
       end
 
-      require 'slack-notifier'
-      notifier = Slack::Notifier.new(Scan.config[:slack_url]) do
-        defaults(channel: channel,
-                username: 'fastlane')
-      end
+      fields = []
 
-      attachments = []
-
-      if Scan.config[:slack_message]
-        attachments << {
-          text: Scan.config[:slack_message].to_s,
-          color: "good"
+      if results[:build_errors]
+        fields << {
+          title: 'Build Errors',
+          value: results[:build_errors].to_s,
+          short: true
         }
       end
 
-      attachments << {
-        text: "Build Errors: #{results[:build_errors] || 0}",
-        color: results[:build_errors].to_i > 0 ? "danger" : "good",
-        short: true
-      }
-
       if results[:failures]
-        attachments << {
-          text: "Test Failures: #{results[:failures]}",
-          color: results[:failures].to_i > 0 ? "danger" : "good",
+        fields << {
+          title: 'Test Failures',
+          value: results[:failures].to_s,
           short: true
         }
       end
 
       if results[:tests] && results[:failures]
-        attachments << {
-          text: "Successful Tests: #{results[:tests] - results[:failures]}",
-          color: "good",
+        fields << {
+          title: 'Successful Tests',
+          value: (results[:tests] - results[:failures]).to_s,
           short: true
         }
       end
 
-      result = notifier.ping("#{Scan.project.app_name} Tests:",
-                             icon_url: 'https://s3-eu-west-1.amazonaws.com/fastlane.tools/fastlane.png',
-                             attachments: attachments)
-      result = result.first
-
-      if result.code.to_i == 200
-        UI.success('Successfully sent Slack notification')
-      elsif result.code.to_i == 404
-        UI.error("The Slack URL you provided could not be reached (404)")
-      else
-        UI.error("The Slack notification could not be sent:")
-        UI.error(result.to_s)
-      end
+      Fastlane::Actions::SlackAction.run({
+        message: "#{Scan.project.app_name} Tests:\n#{Scan.config[:slack_message]}",
+        channel: channel,
+        slack_url: Scan.config[:slack_url].to_s,
+        success: results[:build_errors].to_i == 0 && results[:failures].to_i == 0,
+        username: 'fastlane',
+        icon_url: 'https://s3-eu-west-1.amazonaws.com/fastlane.tools/fastlane.png',
+        payload: {},
+        attachment_properties: {
+          fields: fields
+        }
+      })
     end
   end
 end


janpio:
```
scan/lib/scan/slack_poster.rb:41:7: C: Require/MissingRequireStatement: Fastlane::Actions::SlackAction not found, you're probably missing a require statement or there is a cycle in your dependencies.
      Fastlane::Actions::SlackAction.run({
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Find another part of the codebase that also uses an internal action and see how that one is required to not trigger this error.

gligorkot:
Hi @janpio, thanks for getting back to me so quick. I'm looking for a similar use case, but haven't found one so far. All I can find using internal actions is using `require` but these are all in the `fastlane/actions` subdirectory. I haven't found an example of using an internal action from outside of this directory. Is what I'm trying to do a bad practice? Is there appetite from people for the change I'm proposing at all?

Thanks!

gligorkot:
Hi @janpio, thanks again for your comments and guidance on this one. I've managed to fix the issue I was facing, so now it's a question if whether or not there's appetite for my change. I made this change as I personally like having more consistent slack messages between my builds, so this made sense.

In addition to this, I was hoping to add another PR to enable people to change the name of the Slack message poster as well as the icon, similar to how this is done using the Slack Action, and this change makes it easy to do the above.

Looking forward to hearing your thoughts?

scan/lib/scan/slack_poster.rb
```diff
@@ -12,56 +15,44 @@ def run(results)
...
-      attachments << {
-        text: "Build Errors: #{results[:build_errors] || 0}",
-        color: results[:build_errors].to_i > 0 ? "danger" : "good",
-        short: true
-      }
-
       if results[:failures]
-        attachments << {
-          text: "Test Failures: #{results[:failures]}",
-          color: results[:failures].to_i > 0 ? "danger" : "good",
```
janpio: Is this represented in the rewrite?

scan/lib/scan/slack_poster.rb
```diff
@@ -12,56 +15,44 @@ def run(results)
...
+          title: 'Test Failures',
+          value: results[:failures].to_s,
           short: true
         }
       end
 
       if results[:tests] && results[:failures]
-        attachments << {
-          text: "Successful Tests: #{results[:tests] - results[:failures]}",
-          color: "good",
```
janpio: Is this represented in the rewrite?

scan/lib/scan/slack_poster.rb
```diff
@@ -12,56 +15,44 @@ def run(results)
...
-                username: 'fastlane')
-      end
-
-      attachments = []
+      fields = []
 
-      if Scan.config[:slack_message]
-        attachments << {
-          text: Scan.config[:slack_message].to_s,
-          color: "good"
```
janpio: Is this represented in the rewrite?

scan/lib/scan/slack_poster.rb
```diff
@@ -12,56 +15,44 @@ def run(results)
...
+        fields << {
+          title: 'Build Errors',
+          value: results[:build_errors].to_s,
+          short: true
         }
       end
 
-      attachments << {
-        text: "Build Errors: #{results[:build_errors] || 0}",
-        color: results[:build_errors].to_i > 0 ? "danger" : "good",
```
janpio: Is this represented in the rewrite?

janpio (commented):
Could you share before/after screenshots? I can't really test this, and it is a bit hard to visualize from the code alone ;)

Your follow up idea makes total sense. Customizing this sounds like a great idea so it becomes usable for more companies and teams.

gligorkot:
Hey @janpio, thanks so much for taking this into consideration. I do think it's a worthwhile change as it standardises the slack message output across the `scan` and `slack` actions. I have also seen some previous issues raised around git commits and branches being added as part of the slack custom message of the `scan` action, so this kind of extends that by default instead of users having to do it using the custom message parameter `slack_message`.

Here's a screenshot of the output before the change:
screen shot 2018-12-19 at 8 16 04 am

Here it is, after the change:
screen shot 2018-12-19 at 8 17 10 am

joshdholtz (changes requested):
This is great! Just one small change on the removing the hardcoding of `username` and then we can ship this 🚀

scan/lib/scan/slack_poster.rb
```diff
@@ -12,56 +15,44 @@ def run(results)
...
-      else
-        UI.error("The Slack notification could not be sent:")
-        UI.error(result.to_s)
-      end
+      Fastlane::Actions::SlackAction.run({
+        message: "#{Scan.project.app_name} Tests:\n#{Scan.config[:slack_message]}",
+        channel: channel,
+        slack_url: Scan.config[:slack_url].to_s,
+        success: results[:build_errors].to_i == 0 && results[:failures].to_i == 0,
+        username: 'fastlane',
```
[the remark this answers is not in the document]
gligorkot: Hi @joshdholtz thanks for the review! This used to be up the top (line 18) in the original code. I can remove if needed, let me know.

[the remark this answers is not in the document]
gligorkot: Actually thinking about it, the slack action has `username = 'fastlane'` as the default value, so this can be removed. Doing it now.

gligorkot:
@joshdholtz, I just removed the hardcoded username and the message in slack now comes from `incoming-webhook`, not the most descriptive name. As mentioned above in the review, this used to be hardcoded in line 18 of the original `slack_poster.rb` so I kept that intact, but can remove. Let me know how you'd prefer it and I'll update it if a change is needed.

joshdholtz:
@gligorkot Ahhhhh, good point good point 🤔 Removing that would actually be a breaking change but I just don't like it being hardcoded like that 😱 But I guessssssss we can leave this like how it is. We could introduce another parameter if somebody has an issue with it being "fastlane" sooooo... I think we are good here 👍

joshdholtz (approved):
This is a super solid change 🙌 Thanks for contributing ❤️

gligorkot:
> @gligorkot Ahhhhh, good point good point 🤔 Removing that would actually be a breaking change but I just don't like it being hardcoded like that 😱 But I guessssssss we can leave this like how it is. We could introduce another parameter if somebody has an issue with it being "fastlane" sooooo... I think we are good here 👍

Yeah, I agree. I was hoping to add another PR to enable people to change the name of the Slack message poster as well as the icon, similar to how this is done using the Slack Action, and this change makes it easy to do that! I'll get that done in the next couple of days.

Issue #13626, opened 2018-11-02
Organize screenshots.html to compare all localizations for a screen?


claesjacobsson:
Wouldn't it be useful to be able to generate a version of the screenshots.html that organizes the screenshots so that we can can see all localizations of a screen next to each other?

#### Motivation Behind Feature
This would provide a great tool to verify translations and that they work on different device types.

Is this interesting? Is there already a way to do this? If not, I might give it a try.

janpio:
Sounds useful. Any idea how to best implement this?

PR #13795, opened 2018-12-02
[snapshot] Sort screenshots.html by screen name

claesjacobsson:
### Checklist

### Motivation and Context

The screenshots.html page currently sorts the screenshots by language and device type. 
Sometimes it's useful to compare all localizations of a screen side by side. This PR makes that possible by providing two sorting options: _By Language_ and _By Screen_.

Fixes issue #13626.

### Description

_reports_generator.rb_ now produces two data objects (_data_by_language_ and _data_by_screen_) for _page.html.erb_ to use when creating _screenshots.html_.

_screenshots.html_ has been updated with a top menu bar to jump between the two sorting options.

As far as I understand, the current version only tests the device names being used in _reports_generator_. I have not made any changes and didn't see a reason to add any tests at this time.


janpio (approved):
Works as described, looks like this:

image

The only thing that one could nitpick here is that it only works with Javascript enabled. In my testing, with Javascript disabled in Firefox, I couldn't see any screenshots at all. 

Maybe the tabs could both be visible by default, and only set to `display:none` via Javascript? (Add another headline above the "tab content" so it is clear what is actually shown below, and then also hide that new headline if you want to) What do you think?

claesjacobsson:
Didn't consider that. Your idea makes sense. Would you like me to make a separate PR for that or add to this one?

janpio:
If you can, just commit it to this PR's branch. Then we can merge it in one swoop.

claesjacobsson:
Done!

snapshot/lib/snapshot/page.html.erb
```diff
@@ -64,23 +90,48 @@
...
-        </td><% end %>
-      </tr><% end %>
-    </table><% end %>
+  <body>
+    <div id="sortMenu">
+      <button id="defaultTab" class="tabLink" onclick="openTab(event, 'byLanguage')">By Language</button>
+      <button class="tabLink" onclick="openTab(event, 'byScreen')">By Screen</button>
+    </div>
+    <div id="byLanguage" class="tabContent"><h1 class="tabTitle">By Language:</h1><% image_counter = 0 %><% @data_by_language.each do |language, content| %>
+      <h1 id="<%= language %>"><%= language %></h1>
```
janpio: I know this doesn't really matter here, but could we bump those down to h2 so the semantics are more correct? Thanks.

claesjacobsson: For the language, you mean?

janpio: yep. but actually, I'm fine either way.

claesjacobsson: I can fix that. Makes sense.

janpio (approved):
Looks perfect now, even without JS:

image

And with JS:

image
and
image

janpio:
Thanks for applying my suggestions to this PR @claesjacobsson, the result is awesome! 🚀 💯

PR #13798, opened 2018-12-03
[resign.sh] Add healthkit access entitlement to the blacklist and transfer list

Travolter:
### Checklist

### Motivation and Context
`com.apple.developer.healthkit` is an entitlement that should also be taken from the App. 
Before resigning:   
```
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
        <key>application-identifier</key>
        <string>application.identifier.healthkitswift</string>
        <key>aps-environment</key>
        <string>production</string>
        <key>beta-reports-active</key>
        <true/>
        <key>com.apple.developer.healthkit</key>
        <true/>
        <string>TEAMID</string>
        <key>get-task-allow</key>
        <false/>
</dict>
</plist>.  
``` 
After running `resign.sh HealthKit_sample.ipa  XXXXX  -p "Healthkitswift_test_distribution.mobileprovision" --use-app-entitlements fastlane_resigned.healthkit.ipa`:
```
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
        <key>application-identifier</key>
        <string>application.identifier.healthkitswift</string>
        <key>aps-environment</key>
        <string>production</string>
        <key>beta-reports-active</key>
        <true/>
        <key>com.apple.developer.healthkit</key>
        <true/>
        <key>com.apple.developer.healthkit.access</key>
        <array>
                <string>health-records</string>
        </array>
        <string>TEAMID</string>
        <key>get-task-allow</key>
        <false/>
</dict>
</plist>
```
This is obviously not correct, with this patch it's back to the original entitlements.

### Description

Added the `"com.apple.developer.healthkit.access"` entitlements to the list of entitlements that should not be taken from the provisioning profile but from the app.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -648,6 +648,8 @@ function resign {
             "com.apple.developer.homekit" \
             # If actually used by the App, this value will be set in its entitlements
             "com.apple.developer.healthkit" \
+            # If actually used by the App, this value will be set in its entitlements
+            "com.apple.developer.healthkit.access" \
             # PP list identifiers inconsistent with app-defined ones, must use App entitlements value
             "com.apple.developer.in-app-payments" \
             # If actually used by the App, this value will be set in its entitlements
@@ -706,6 +708,7 @@ function resign {
             "com.apple.developer.associated-domains" \
             "com.apple.developer.default-data-protection" \
             "com.apple.developer.healthkit" \
+            "com.apple.developer.healthkit.access" \
             "com.apple.developer.homekit" \
             "com.apple.developer.icloud-container-environment" \
             "com.apple.developer.icloud-container-identifiers" \


Travolter:
I signed it!

lyndsey-ferguson:
`com.apple.developer.healthkit.access` seems to be added by Apple as I saw comments in the Apple Dev forums. Does this code remove this entry if it does not appear in the App, but leaves it in if so. Not familiar with how the blacklist works

Travolter:
Yes, when you use the `--use-app-entitlements` flag (idk how this translate to the `fastlane sigh` command, it's slightly different there) then it will trigger this code path. 

The blacklist will prevent the script from copying over the entitlement from the provisioning profile (which would by default contain the `<string>health-records</string>` entry). And the `ENTITLEMENTS_TRANSFER_RULES` will then be used to copy over from the app's entitlements instead. If it's not in the original app's entitlements, there is nothing to copy over ofcourse.

lyndsey-ferguson (approved):
Looks good to me.

joshdholtz (approved):
This looks good to me! 👍

Issue #13815, opened 2018-12-05
[2.109.0] Could not upload with fastlane deliver: source sequence is illegal / malformed utf-8 (JSON :: GeneratorError)


kazuhidet:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [X] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [X] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

`source sequence is illegal / malformed utf-8 (JSON :: GeneratorError)` error occurs when uploading ipa to iTunes connect with fastlane deliver.

The problem is in the following part modified by fastlane - 2.109.0, and to_json is in error because there is a character string which is not normally escaped in body.
(The difference is meaningless because it is an ad hoc modification that I added.)

<pre>
--- client.rb	2018-12-05 10:44:31.000000000 +0900
+++ /opt/local/lib/ruby2.3/gems/2.3.0/gems/fastlane-2.109.0/spaceship/lib/spaceship/client.rb	2018-12-05 10:37:32.000000000 +0900
@@ -780,7 +780,11 @@
       params_to_log = params_to_log.collect do |key, value|
         "{#{key}: #{value}}"
       end
+      begin
       logger.info(">> #{method.upcase} #{url}: #{body.to_json} #{params_to_log.join(', ')}")
+      rescue Exception => e
+        logger.info(">> #{method.upcase} #{url}: [Broken body!] #{params_to_log.join(', ')}")
+      end
     end

     def log_response(method, url, response, headers = nil, &block)
</pre>

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>[08:55:49]: Successfully loaded '/Users/Shared/Develop/AutoBuild/workspace/**********/fastlane/Deliverfile' 📄

+-----------------------+---------------------------+
|   Detected Values from './fastlane/Deliverfile'   |
+-----------------------+---------------------------+
| app_identifier        | ********** |
| username              | ********** |
+-----------------------+---------------------------+

[08:55:49]: Login to App Store Connect (**********)
[08:55:50]: Login successful

+--------------------------------------+--------------------------------------+
|                           deliver 2.109.0 Summary                           |
+--------------------------------------+--------------------------------------+
| username                             | ********** |
| dev_portal_team_id                   | ********** |
[...]

PR #13822, opened 2018-12-05
Could not upload ipa by fastlane deliver. #13815

kazuhidet:
### Checklist

### Motivation and Context

'source sequence is illegal / malformed utf-8 (JSON :: GeneratorError)' error occurs when uploading ipa to iTunes connect with fastlane deliver.

As a result of analysis of the code, the 'body' of 'block' passed to this function 'log_request' contains information that is not JSON, such as uploaded PNG file. so if you try to convert it to JSON unconditionally it will cause problems.
[#13815](https://github.com/fastlane/fastlane/issues/13815)

### Description

In the case where the result of analyzing body is not JSON, logic that does not output body to log was added.


kazuhidet:
¥

PR #13823, opened 2018-12-05
[spaceship] Do not attempt to log non-JSON bodies

kazuhidet:
### Checklist

### Motivation and Context

'source sequence is illegal / malformed utf-8 (JSON :: GeneratorError)' error occurs when uploading ipa to iTunes connect with fastlane deliver.

As a result of analysis of the code, the 'body' of 'block' passed to this function 'log_request' contains information that is not JSON, such as uploaded PNG file. so if you try to convert it to JSON unconditionally it will cause problems.
#13815

### Description

In the case where the result of analyzing body is not JSON, logic that does not output body to log was added.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -766,10 +766,12 @@ module Spaceship
     def log_request(method, url, params, headers = nil, &block)
       url ||= extract_key_from_block('url', &block)
       body = extract_key_from_block('body', &block)
+      log_body = false
       if body
         begin
           body = JSON.parse(body)
           body['password'] = '***' if body.kind_of?(Hash) && body.key?("password")
+          log_body = true
         rescue JSON::ParserError
           # no json, no password
         end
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/client.rb
```diff
@@ -780,7 +782,9 @@ def log_request(method, url, params, headers = nil, &block)
       params_to_log = params_to_log.collect do |key, value|
         "{#{key}: #{value}}"
       end
-      logger.info(">> #{method.upcase} #{url}: #{body.to_json} #{params_to_log.join(', ')}")
+      if log_body
+        logger.info(">> #{method.upcase} #{url}: #{body.to_json} #{params_to_log.join(', ')}")
```
janpio: We definitely need an `else` here that still logs the request, but without the body.

Alternatively we could rewrite this so a new `body_to_log` variable is introduced with value `""` before the `if body`, then set to `body.to_json` if we could parse it to JSON, and to `"[body could not be encoded for logging]"` in the `rescue JSON::ParserError`... then the actual `logger.info` stays simple.

janpio: I created my suggestion as an individual PR at https://github.com/fastlane/fastlane/pull/13825

janpio (changes requested):
We have to handle the else case as well.
More general solution in https://github.com/fastlane/fastlane/pull/13825

PR #13831, opened 2018-12-05
[spaceship] copy cookie to clipboard

### LINKED ISSUE #13742: `fastlane spaceauth` could ask if you want to copy the cookie to the clipboard
... as `fastlane env` already does.

https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/fastlane/lib/fastlane/environment_printer.rb#L13-L16

https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/fastlane/lib/fastlane/environment_printer.rb#L292-L296

https://github.com/fastlane/fastlane/blob/f0fc7f8f8b3790dee0dd06b0d1d928576a7b430c/spaceship/lib/spaceship/spaceauth_runner.rb#L52-L60

RishabhTayal:
fixes https://github.com/fastlane/fastlane/issues/13742

### DIFF
--- a/spaceship/lib/spaceship/spaceauth_runner.rb
+++ b/spaceship/lib/spaceship/spaceauth_runner.rb
@@ -49,7 +49,7 @@ module Spaceship
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
-      if FastlaneCore::Helper.mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 should fastlane put cookie into your clipboard, so you can easily paste it? (y/n)", true)
+      if Spaceship::Client::UserInterface.interactive? && agree("🙄 should fastlane put cookie into your clipboard, so you can easily paste it? (y/n)", true)
         require 'open3'
         Open3.popen3('pbcopy') { |input, _, _| input << yaml }
         puts("Successfully copied text into your clipboard 🎨".green)
--- a/spaceship/lib/spaceship/spaceauth_runner.rb
+++ b/spaceship/lib/spaceship/spaceauth_runner.rb
@@ -49,11 +49,10 @@ module Spaceship
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
-
       if FastlaneCore::Helper.mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 should fastlane put cookie into your clipboard, so you can easily paste it? (y/n)", true)
         require 'open3'
         Open3.popen3('pbcopy') { |input, _, _| input << yaml }
-        puts("Successfully copied text into your clipboard 🎨".green) 
+        puts("Successfully copied text into your clipboard 🎨".green)
       end
 
       puts("---")
--- a/spaceship/lib/spaceship/spaceauth_runner.rb
+++ b/spaceship/lib/spaceship/spaceauth_runner.rb
@@ -49,7 +49,7 @@ module Spaceship
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
-      if Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
+      if mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
         require 'open3'
         Open3.popen3('pbcopy') { |input, _, _| input << yaml }
         puts("Successfully copied text into your clipboard 🎨".green)
@@ -63,6 +63,10 @@ module Spaceship
       puts("")
       puts("Example:")
       puts("export FASTLANE_SESSION='#{yaml}'".cyan.underline)
+    end
+
+    def mac?
+      (/darwin/ =~ RUBY_PLATFORM) != nil
     end
   end
 end
--- a/spaceship/lib/spaceship/spaceauth_runner.rb
+++ b/spaceship/lib/spaceship/spaceauth_runner.rb
@@ -49,6 +49,11 @@ module Spaceship
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,6 +49,11 @@ def run
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
+      if FastlaneCore::Helper.mac? && UI.interactive? && UI.confirm("🙄  Wow, that's a lot of markdown text... should fastlane put it into your clipboard, so you can easily paste it on GitHub?")
```
janpio: This is not markdown here.

RishabhTayal: @janpio I have updated the copy.

janpio:
See https://circleci.com/gh/fastlane/fastlane/26902 - spaceship doesn't like `UI` I think.

RishabhTayal:
How can I fix it?

janpio:
Find out if this is really the problem, if it is find another way to achieve what you want.

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,6 +49,11 @@ def run
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
+      if FastlaneCore::Helper.mac? && UI.interactive? && UI.confirm("🙄  Wow, that's a lot of text... should fastlane put it into your clipboard, so you can easily paste it?") 
+        copy_to_clipboard(yaml)
```
KrauseFx: This method is defined in `environment_printer`, does this work even?

KrauseFx: Oh yah, so `spaceship` probably shouldn't just access a `fastlane` class, it's probably fair to copy that method over here, as it basically just shells out?

RishabhTayal: > Find out if this is really the problem, if it is find another way to achieve what you want.

@janpio `UI` is not accessible here. What is the right way to use it?

RishabhTayal: You are right. I have addressed it.

RishabhTayal: I just tried and it seems to be working. Am I missing anything?

RishabhTayal:
@janpio and @KrauseFx this is ready for review.

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,6 +49,12 @@ def run
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
+      if Spaceship::Client::UserInterface.interactive? && agree("🙄 should fastlane put cookie into your clipboard, so you can easily paste it? (y/n)", true)
```
janpio:
```suggestion
      if Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
```

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,6 +49,12 @@ def run
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
+      if Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
+        require 'open3'
+        Open3.popen3('pbcopy') { |input, _, _| input << yaml }
```
joshdholtz: `pbcopy` is only a mac thing so this won't work if linux or windows 🤔 Are we able to get a cross platform solution? Or if not can we add a check specifically for macos?

joshdholtz (changes requested):
Works good on mac but just quick thoughts on other OSes 😇

spaceship/lib/spaceship/spaceauth_runner.rb
janpio: I noticed this as well and created an issue some time ago: https://github.com/fastlane/fastlane/issues/13897 Although for this PR it is probably out of scope.

But: I think the other "copy to clipboard" functionalities are hidden behind the `mac?` helper, which is a great idea to not suggest this can work on Linux or Windows.

RishabhTayal: @joshdholtz @janpio So what's the solution here?

janpio: Exactly what I wrote in my previous comment - make sure this code is only run when on Mac.

RishabhTayal: @janpio I have added a check for `mac?`. Please review again.

janpio: I confirmed that this doesn't get output on Windows. Someone else will have to confirm it does get output on Mac.

(Having tests for `spaceauth` would be neat...)

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,6 +49,12 @@ def run
 
       yaml = cookies.to_yaml.gsub("\n", "\\n")
 
+      if mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
```
janpio: Shouldn't this happen after the output of the cookie? I have to be able to look at it first if it makes sense, don't I?

RishabhTayal: I think it makes sense to move it after the output. I will move it.

janpio (commented):
✔️ does not appear on Windows

getaaron (approved):

joshdholtz (approved):
Thanks for adding this!! Copying this was always touch 😝 ❤️

PR #13836, opened 2018-12-06
[zip] Handle symbolic links

Hanstad:
With symlinks option it is now possible to create a zip file that stores symbolic links to the archive

### Checklist

### Motivation and Context
Fastlane comes with a Zip action that enable users to zip files and folders. But it does not handle symbolic links as links, but instead it include all linked files. This is not wanted when the folder to zip already contains the linked files.

### Description
I have improved the already existing Zip action to control whether symbloic links should be treated as symbolic links or not. 
By default the behaviour of the action hasn't changed. By setting symlinks: true the user will get a zip file that can contain symbolic links.

I tested this on my build system by creating a copy of the zip action and included it in my repo. It works as expected :)

### DIFF
--- a/fastlane/lib/fastlane/actions/zip.rb
+++ b/fastlane/lib/fastlane/actions/zip.rb
@@ -81,8 +81,13 @@ module Fastlane
           'zip(
             path: "MyApp.app",
             output_path: "Latest.app.zip",
+            verbose: false
+          )',
+          'zip(
+            path: "MyApp.app",
+            output_path: "Latest.app.zip",
             verbose: false,
-            symlinks: false
+            symlinks: true
           )'
         ]
       end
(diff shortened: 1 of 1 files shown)


Hanstad:
I signed it!

Hanstad:
I signed it!

Hanstad:
Why is this PR blocked? I need this functionality and the code work and will help others.

janpio:
Where does it say it is blocked? Until now only nobody had the time to review, test and merge it.

fastlane/lib/fastlane/actions/zip.rb
```diff
@@ -74,7 +81,8 @@ def self.example_code
           'zip(
             path: "MyApp.app",
             output_path: "Latest.app.zip",
-            verbose: false
+            verbose: false,
+            symlinks: false
```
janpio: It doesn't make much sense to add the param with its default value to an existing example. Better create a new, third one, that shows it changing the default behavior.

janpio (approved):
Besides my comment regarding the example this looks good.

KrauseFx (approved):
Beautiful, thanks for your contribution @Hanstad 👍

PR #13846, opened 2018-12-07
Fix AdbHelper

### LINKED ISSUE #13845: adb is broken
From the comments to https://github.com/fastlane/fastlane/pull/13692:

@tmaly1980 writes:

> Sorry, @janpio - what does the shellescape do to help? I'm suddenly getting all of my adb commands failing, simply printing out the help text for the adb command....
> 
> ```
> adb(
>    command: "shell settings put global window_animation_scale 0.0"
> )
> ```
> 
> Ends up calling "adb shell\ " (with a space at the end, which is not a valid command, so it returns the help screen):
> 
> `/Users/zingfitdev2/Library/Android/sdk/platform-tools/adb shell\ settings\ put\ global\ window_animation_scale\ 0.0`
> 
> Which then just does:
> 
> ```
> [22:59:20]: ▸ Android Debug Bridge version 1.0.39
> [22:59:20]: ▸ Revision 3db08f2c6889-android
> [22:59:20]: ▸ Installed as /Users/zingfitdev2/Library/Android/sdk/platform-tools/adb
> [22:59:20]: ▸ global options:
> [22:59:20]: ▸ -a         listen on all network interfaces, not just localhost
> [22:59:20]: ▸ -d         use USB device (error if multiple devices connected)
> [22:59:20]: ▸ -e         use TCP/IP device (error if multiple TCP/IP devices available)
> [22:59:20]: ▸ -s SERIAL
> [22:59:20]: ▸ use device with given serial number (overrides $ANDROID_SERIAL)
> [22:59:20]: ▸ -p PRODUCT
> [22:59:20]: ▸ name or path ('angler'/'out/target/product/angler');
> [22:59:20]: ▸ default $ANDROID_PRODUCT_OUT
> [22:59:20]: ▸ -H         name of adb server host [default=localhost]
> [22:59:20]: ▸ -P         port of adb server [default=5037]
> [22:59:20]: ▸ -L SOCKET  listen on given socket for adb server [default=tcp:localhost:5037]
> [22:59:20]: ▸ general commands:
> [22:59:20]: ▸ devices [-l]             list connected devices (-l for long output)
> [22:59:20]: ▸ help                     show this help message
> [22:59:20]: ▸ version                  show version num
> ```
> 
> What am I doing wrong?

and

> This unfortunately completely breaks the adb command. Any commands that take parameters (such as "adb shell ls", and most other commands) will fail, because it puts a backslash before the space, and thinks the command is "shell\ " rather than "shell"

That would a regression :(

janpio:
### Checklist

### Motivation and Context
https://github.com/fastlane/fastlane/pull/13692 added one `shellescape` too much and broke the execution of `adb` commands with several parts: `foo` works, `foo bar` is escaped to `foo\ bar` and does not work any more.

### Description
- Removes the `.shellescape`
- Adds tests for the multi part command case and the other case where shellescape of the path is actually needed (code added in #13692)
- Adds `.strip!` to the command, adapts the tests to not expect commands to be prefixed with an additional space

---

closes #13845

### DIFF
--- a/fastlane/lib/fastlane/helper/adb_helper.rb
+++ b/fastlane/lib/fastlane/helper/adb_helper.rb
@@ -26,7 +26,7 @@ module Fastlane
       # Run a certain action
       def trigger(command: nil, serial: nil)
         android_serial = serial != "" ? "ANDROID_SERIAL=#{serial}" : nil
-        command = [android_serial, adb_path.shellescape, command.shellescape].join(" ")
+        command = [android_serial, adb_path.shellescape, command].join(" ")
         Action.sh(command)
       end
 
--- a/fastlane/lib/fastlane/helper/adb_helper.rb
+++ b/fastlane/lib/fastlane/helper/adb_helper.rb
@@ -26,7 +26,7 @@ module Fastlane
       # Run a certain action
       def trigger(command: nil, serial: nil)
         android_serial = serial != "" ? "ANDROID_SERIAL=#{serial}" : nil
-        command = [android_serial, adb_path.shellescape, command].join(" ")
+        command = [android_serial, adb_path.shellescape, command].join(" ").strip!
         Action.sh(command)
       end
 
--- a/fastlane/spec/actions_specs/adb_spec.rb
+++ b/fastlane/spec/actions_specs/adb_spec.rb
@@ -6,7 +6,15 @@ describe Fastlane do
           adb(command: 'test', adb_path: './fastlane/README.md')
         end").runner.execute(:test)
 
-        expect(result).to eq(" ./fastlane/README.md test")
+        expect(result).to eq("./fastlane/README.md test")
+      end
+
+      it "generates a valid command for commands with multiple parts" do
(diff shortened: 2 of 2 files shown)


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

tmaly1980:
@janpio Those tests appear inadequate (which is why the bug got through in the first place). How about you add some multi-word commands to adb, such as "adb shell ls" and the like?

janpio:
Isn't this exactly what I did in https://github.com/fastlane/fastlane/pull/13846/files#diff-c6e12996480160bc02efa839f076b436R12? The tests don't actually run the shell commands, but return the commands to be run that are then compared with the expected command.

tmaly1980:
@janpio ok I see, you're right.

joshdholtz (approved):
🙌

PR #13854, opened 2018-12-07
[action] Fix Crashlytics deprecated message for submit binary path

rkrim:
### Checklist

### Motivation and Context
A deprecated message is always printed if the path of crashlytics submit binary is located under 'crashlytics.framework'.
While the deprecated message make sense for cocoapods install of crashlytics, it is not the case for other installation methods, like carthage.
Link to open issue: [13574](https://github.com/fastlane/fastlane/issues/13574)
All test passed.
All codestyle inspection passed

### Description
Added a test to check that the path also contains "pods" before displaying the deprecated message.
Tested path manipulation in/out to check that the deprecated message is printed only if "pods" is contained in the path.

### DIFF
--- a/fastlane/lib/fastlane/helper/crashlytics_helper.rb
+++ b/fastlane/lib/fastlane/helper/crashlytics_helper.rb
@@ -24,7 +24,8 @@ module Fastlane
           path ||= Dir["./Pods/iOS/Crashlytics/Crashlytics.framework/submit"].last
           path ||= Dir["./**/Crashlytics.framework/submit"].last
 
-          if path && path.downcase.include?("crashlytics.framework")
+          downcase_path = path ? path.downcase : nil
+          if downcase_path && downcase_path.include?("pods") && downcase_path.include?("crashlytics.framework")
             UI.deprecated("Crashlytics has moved the submit binary outside of Crashlytics.framework directory as of 3.4.1. Please change :crashlytics_path to `<PODS_ROOT>/Crashlytics/submit`")
           end
 


rkrim:
I signed it!

fastlane/lib/fastlane/helper/crashlytics_helper.rb
```diff
@@ -24,7 +24,7 @@ def discover_crashlytics_path(params)
           path ||= Dir["./Pods/iOS/Crashlytics/Crashlytics.framework/submit"].last
           path ||= Dir["./**/Crashlytics.framework/submit"].last
 
-          if path && path.downcase.include?("crashlytics.framework")
+          if path && (path_downcase = path.downcase).include?("pods") && path_downcase.include?("crashlytics.framework")
```
janpio: What does the `(path_downcase = path.downcase)` do?

rkrim: As there is two conditions using `path.downcase`, the result is stored in `path_downcase` variable before evaluating the first test, to be used in this first test then be reused in the second test and avoid another call to downcase function.

janpio: Ah, I overlooked the difference between `_` and `.`. Maybe I should get some glasses :/

rkrim: As it happened, means there's a readability issue.
What do you think about changing `path_downcase` to `downcase_path` to better notice the difference? `(downcase_path = path.downcase)`

janpio: Very good idea. Or just do it in an extra line - I wouldn't have been to sure that this would actually work (but I also don't really know ruby, so there's that...) 🙊

KrauseFx:
Agreed with @janpio, let's just use multiple lines instead, as it will be easier to read for people :)

fastlane/lib/fastlane/helper/crashlytics_helper.rb
rkrim: Done in multi-line

janpio (approved):

KrauseFx (approved):
Great, thanks for your contribution 👍

PR #13871, opened 2018-12-11
[spaceship] Retry request when App Store Connect raises Bad Gateway

giginet:
### Checklist

### Motivation and Context

Sometimes, App Store Connect raises Bad Gateway.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <style>
        body {
            font-family: "Helvetica Neue", "HelveticaNeue", Helvetica, Arial, sans-serif;
            font-size: 15px;
            font-weight: 200;
            line-height: 20px;
            color: #4c4c4c;
            text-align: center;
        }

        .section {
            margin-top: 50px;
        }
    </style>
</head>
<body>
<div class="section">
    <h1>&#63743;</h1>

    <h3>Bad Gateway</h3>
    <p>Correlation Key: XXXXXXXXXXXXXXXXXXXXX </p>
</div>
</body>
</html>
```

When fastlane/deliver is waiting a processing phase, lane execution is stoped due to this error.

We want to ignore this error and retry requests.

### Description

- Defined BadGateway error
- If BadGateway is raised, fastlane attempt to retry sending.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -53,6 +53,7 @@ module Spaceship
     UnauthorizedAccessError = Spaceship::UnauthorizedAccessError
     GatewayTimeoutError = Spaceship::GatewayTimeoutError
     InternalServerError = Spaceship::InternalServerError
+    BadGatewayError = Spaceship::BadGatewayError
 
     def self.hostname
       raise "You must implement self.hostname"
@@ -650,7 +651,7 @@ module Spaceship
       @csrf_tokens || {}
     end
 
-    def request(method, url_or_path = nil, params = nil, headers = {}, auto_paginate = false, &block)
+    def request(method, url_or_path = nil, params = nil, headers = {}, auto_paginate = false, retry_on_bad_gateway = true, &block)
       headers.merge!(csrf_tokens)
       headers['User-Agent'] = USER_AGENT
 
@@ -663,13 +664,20 @@ module Spaceship
         params, headers = encode_params(params, headers)
       end
 
-      response = if auto_paginate
-                   send_request_auto_paginate(method, url_or_path, params, headers, &block)
-                 else
-                   send_request(method, url_or_path, params, headers, &block)
-                 end
-
-      return response
+      # Retry requests when Apple returns BadGateway
+      max_retry_count = retry_on_bad_gateway ? 3 : 1
+      max_retry_count.times do
+        begin
+          response = if auto_paginate
+                       send_request_auto_paginate(method, url_or_path, params, headers, &block)
+                     else
+                       send_request(method, url_or_path, params, headers, &block)
+                     end
+          return response
+        rescue BadGatewayError
+          next
+        end
+      end
     end
 
     def parse_response(response, expected_key = nil)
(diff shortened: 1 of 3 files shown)


spaceship/lib/spaceship/client.rb
```diff
@@ -603,6 +604,7 @@ def with_retry(tries = 5, &_block)
     rescue \
         Faraday::Error::ConnectionFailed,
         Faraday::Error::TimeoutError, # New Faraday version: Faraday::TimeoutError => ex
+        BadGatewayError,
```
janpio: Is BadGatewayError really a "timeout" error?

giginet: Maybe no. But it seems to occur temporarily.
It should be ignored or retried.

spaceship/lib/spaceship/client.rb
```diff
@@ -829,6 +831,10 @@ def send_request(method, url_or_path, params, headers, &block)
           raise AppleTimeoutError.new, "Apple 302 detected - this might be temporary server error, check https://developer.apple.com/system-status/ to see if there is a known downtime"
         end
 
+        if response.body.to_s.include?("<h3>Bad Gateway</h3>")
```
janpio: This is only returned with status code 502? Then we should maybe check for that as well (or have an additional `if` for that)?

giginet: I have no idea. How about checking status codes satisfy 5XX?

```
&& response.status >= 500 && response.status <= 599
```

janpio:
Do we have a concrete issue this is trying to solve? I think I remember some, but can't find them right now.

janpio:
If you are using 2.109+ already you can check the spaceship logs at `/tmp/spaceship-*.log` - those should contain the response body and header code now. 

You could also send it to me at piotrowski@gmail.com, then I can take a deeper look to understand what seems to be going on and which calls exactly are erroring (and if you include the output of fastlane at the same time, even correlate to what exactly fastlane was doing).

spaceship/spec/client_spec.rb
```diff
@@ -126,6 +127,43 @@ def stub_client_retry_auth(status_error, times, status_ok, body)
...
+    <p>Correlation Key: XXXXXXXXXXXXXXXXXXXX</p>
+</div>
+</body>
+</html>
+BODY
+      stub_client_retry_auth(502, 1, 200, body)
+
+      expect do
+        subject.req_home
+      end.to raise_error(Spaceship::Client::BadGatewayError)
```
KrauseFx: Nice test

KrauseFx:
Thanks for the PR @giginet 👍

giginet:
We are still on 2.108.0 so logs are disappeared. 😢 
I'll wait for the next error to further investigation.

janpio:
(2.108.0 also has logs of course, but they problably don't contain the neccessary information to debug this properly. But have a look anyway if you can - maybe I am wrong)

spaceship/lib/spaceship/client.rb
snatchev: Let's keep the `BadGatewayError` only for 502 responses.

snatchev (approved):
Looks good to me :)

joshdholtz (approved):
I like this 🙌

giginet:
We were waiting for this problem to investigate further error responses.
However, it might not be caused by the latest version.

Thanks!!

PR #13878, opened 2018-12-11
[spaceship] Improve TunesClient request retry

janpio:
- Fix bug with wrong number of retries being output
- Extract wait times into variable

### DIFF
--- a/spaceship/lib/spaceship/tunes/tunes_client.rb
+++ b/spaceship/lib/spaceship/tunes/tunes_client.rb
@@ -1435,20 +1435,22 @@ module Spaceship
     def with_tunes_retry(tries = 5, potential_server_error_tries = 3, &_block)
       return yield
     rescue Spaceship::TunesClient::ITunesConnectTemporaryError => ex
+      seconds_to_sleep = 60
       unless (tries -= 1).zero?
-        msg = "App Store Connect temporary error received: '#{ex.message}'. Retrying after 60 seconds (remaining: #{tries})..."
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Beautiful 👍

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -16,7 +16,7 @@ class TunesClient < Spaceship::Client
     # Legacy support
     ITunesConnectError = Tunes::Error
     ITunesConnectTemporaryError = Tunes::TemporaryError
-    ITunesConnectPotentialServerError = Tunes::PotentialServerError
+    ITunesConnectFlakyCallPotentialServerError = Tunes::PotentialServerError
```
snatchev: What is the purpose of this rename? Also, This may break people's code if they rescue `ITunesConnectPotentialServerError` and now it is undefined.

janpio: The thinking behind the rename was that it describes the exception better - this can only be raised by "flaky calls" and doesn't really describe a server error, to be honest not even a potential one. 

The actual problem is probably a level deeper, in the upload of assets that are failing (and not being handled) - which then triggers the "flaky call" to return an error. If I figure out what _really_ is the other problem, and it is the sole reason for this here being triggered, I was thinking about just removing this.

janpio: But I agree, the rename is problematic - and doesn't really help anyone here. I will remove it.

janpio: Removed that commit and force pushed. Thanks for checking me on this! 👍

PR #13816, opened 2018-12-05
[WIP] Adding rubocop linter for FastlaneCore::ConfigItem `is_string`

joshdholtz:
## Description
`is_string: true|false` is legacy and should be replaced with `type: Integer|Float|String|Boolean|Array|Symbol`

- [x] Add rubocop check
- [ ] Remove from **ALL** the places

### Example
```sh
screengrab/lib/screengrab/options.rb:46:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false),
                                     ^^^^^^^^^^^^^^^^
screengrab/lib/screengrab/options.rb:126:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false),
                                     ^^^^^^^^^^^^^^^^
screengrab/lib/screengrab/options.rb:131:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false)
                                     ^^^^^^^^^^^^^^^^
sigh/lib/sigh/options.rb:15:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false,
                                     ^^^^^^^^^^^^^^^^
sigh/lib/sigh/options.rb:24:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false,
                                     ^^^^^^^^^^^^^^^^
sigh/lib/sigh/options.rb:33:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false,
                                     ^^^^^^^^^^^^^^^^
sigh/lib/sigh/options.rb:42:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false,
                                     ^^^^^^^^^^^^^^^^
sigh/lib/sigh/options.rb:47:38: W: is_string key in used in FastlaneCore::ConfigItem. Replace with type: <Integer|Float|String|Boolean|Array|Symbol>
                                     is_string: false,
                                     ^^^^^^^^^^^^^^^^
[...]

PR #13880, opened 2018-12-11
[Swift] Fix frameit's Boolean options are generated as String? instead of Bool?

jhandguy:
### Checklist

### Motivation and Context

See issue: https://github.com/fastlane/fastlane/issues/13879

When the `frameScreenshot` action is generated from Ruby to Swift, its implicit `Optional Boolean` parameters are interpreted as `String?` (default fallback for non-explicit Ruby types).

See issue: https://github.com/fastlane/fastlane/issues/13879

### Description

Giving the `Options Config type` explicitly allows the `generate_swift_api` action to detect the type.

### Discussion

Why are these Options `Optional` at the first place?
Could we simply have them required with a `default_value` of `false`?

### DIFF
--- a/frameit/lib/frameit/options.rb
+++ b/frameit/lib/frameit/options.rb
@@ -8,22 +8,25 @@ module Frameit
       @options ||= [
 
         FastlaneCore::ConfigItem.new(key: :white,
-                                       env_name: "FRAMEIT_WHITE_FRAME",
-                                       description: "Use white device frames",
-                                       optional: true,
-                                       is_string: false),
+                                     env_name: "FRAMEIT_WHITE_FRAME",
+                                     description: "Use white device frames",
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :silver,
+                                     env_name: "FRAMEIT_SILVER_FRAME",
                                      description: "Use white device frames. Alias for :white",
-                                     optional: true,
-                                     is_string: false),
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :rose_gold,
+                                     env_name: "FRAMEIT_ROSE_GOLD_FRAME",
                                      description: "Use rose gold device frames. Alias for :rose_gold",
-                                     optional: true,
-                                     is_string: false),
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :gold,
+                                     env_name: "FRAMEIT_GOLD_FRAME",
                                      description: "Use gold device frames. Alias for :gold",
-                                     optional: true,
-                                     is_string: false),
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :force_device_type,
                                      env_name: "FRAMEIT_FORCE_DEVICE_TYPE",
                                      description: "Forces a given device type, useful for Mac screenshots, as their sizes vary",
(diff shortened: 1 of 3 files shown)


frameit/lib/frameit/options.rb
```diff
@@ -8,22 +8,29 @@ def self.available_options
       @options ||= [
 
         FastlaneCore::ConfigItem.new(key: :white,
-                                       env_name: "FRAMEIT_WHITE_FRAME",
-                                       description: "Use white device frames",
-                                       optional: true,
-                                       is_string: false),
+                                     env_name: "FRAMEIT_WHITE_FRAME",
+                                     description: "Use white device frames",
+                                     is_string: false,
```
joshdholtz: You can remove this `is_string` 😊

frameit/lib/frameit/options.rb
```diff
@@ -8,22 +8,29 @@ def self.available_options
...
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :silver,
+                                     env_name: "FRAMEIT_SILVER_FRAME",
                                      description: "Use white device frames. Alias for :white",
-                                     optional: true,
-                                     is_string: false),
+                                     is_string: false,
```
joshdholtz: And this one

frameit/lib/frameit/options.rb
```diff
@@ -8,22 +8,29 @@ def self.available_options
...
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :rose_gold,
+                                     env_name: "FRAMEIT_ROSE_GOLD_FRAME",
                                      description: "Use rose gold device frames. Alias for :rose_gold",
-                                     optional: true,
-                                     is_string: false),
+                                     is_string: false,
```
joshdholtz: And this one

frameit/lib/frameit/options.rb
```diff
@@ -8,22 +8,29 @@ def self.available_options
...
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :gold,
+                                     env_name: "FRAMEIT_GOLD_FRAME",
                                      description: "Use gold device frames. Alias for :gold",
-                                     optional: true,
-                                     is_string: false),
+                                     is_string: false,
```
joshdholtz: And this one

frameit/lib/frameit/options.rb
```diff
@@ -35,29 +42,31 @@ def self.available_options
                                        end
                                      end),
         FastlaneCore::ConfigItem.new(key: :use_legacy_iphone5s,
-                           env_name: "FRAMEIT_USE_LEGACY_IPHONE_5_S",
-                           is_string: false,
-                           description: "Use iPhone 5s instead of iPhone SE frames",
-                           default_value: false),
+                                     env_name: "FRAMEIT_USE_LEGACY_IPHONE_5_S",
+                                     is_string: false,
```
joshdholtz: And this one

frameit/lib/frameit/options.rb
```diff
@@ -35,29 +42,31 @@ def self.available_options
...
+                                     description: "Use iPhone 5s instead of iPhone SE frames",
+                                     default_value: false,
+                                     type: Boolean),
         FastlaneCore::ConfigItem.new(key: :use_legacy_iphone6s,
-                           env_name: "FRAMEIT_USE_LEGACY_IPHONE_6_S",
-                           is_string: false,
-                           description: "Use iPhone 6s frames instead of iPhone 7 frames",
-                           default_value: false),
+                                     env_name: "FRAMEIT_USE_LEGACY_IPHONE_6_S",
+                                     is_string: false,
```
joshdholtz: And this one

joshdholtz (changes requested):
These changes will work but if you could also remove the `is_string` that would be 💯  as that's not needed!

I actually have a PR that I'm working on that is removing the `is_string` from all options completely so this shouldn't be an issue for any tool/action in the future - https://github.com/fastlane/fastlane/pull/13816

jhandguy:
Done @joshdholtz ✅

How about having `default_value: false` instead of `optional: true`?
Would it break something?

joshdholtz:
@jhandguy That one is a little more difficult to answer 😇 Setting a `default: false` will never instead of `optional: true` _should_ be fine in _most_ cases but there are some instances where the code will actually check for a `nil` which won’t over happen if that `optional: true` is set. I don’t know if this is the case for anything in _frameit_ but it requires a deeper inspection of the underlying code 😬

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jhandguy:
@joshdholtz got ya, let's leave it as it is then! 😊

jhandguy:
@joshdholtz did you have time to take a look?

joshdholtz (approved):
🙌

Issue #13772, opened 2018-11-28
Google Cloud Storage follow-up tasks


KrauseFx:
- [ ] Add more fine-grained error handling (e.g. readonly access to bucket)
- [ ] More unit tests
	- [ ] In particular, cover the migration
- [ ] Make sure to support the `gc_keys.json` file to be inside the `fastlane` folder or outside
- [x] `fastlane match init` should ask for storage backend to use
- [ ] Unify or remove the `update_optional_values_depending_on_storage_type` method
- [x] `fastlane match nuke` show the bucket that will be deleted
- [x] `fastlane match nuke` should also delete the folders from the Google Cloud bucket
- [x] Update fastlane docs to mention Google Cloud, and provide more information on how to migrate
- [ ] Move `prefixed_working_directory` method into sub-classes so it can actually be extended
- [ ] How does the new separation of `ConfigItems` affect the fastlane/docs page
- [ ] Finish up https://github.com/fastlane/fastlane/pull/13830
- [ ] Address https://github.com/fastlane/fastlane/pull/13589#discussion_r238396182

**Optional**
- [ ] Make `fastlane match` migrate from any storage to any storage?
- [ ] Check if we can download multiple files in parallel to speed up the process

KrauseFx:
Don't tell me what to do @fastlane-bot, I'm your creator

KrauseFx:
Regarding `fastlane match nuke should also delete the folders from the Google Cloud bucket`, that actually isn't easily possible, as Google Cloud Storage isn't designed like this. It's good to keep the folders around for now.

PR #13887, opened 2018-12-11
Add support for `gc_keys.json` file in fastlane subfolder

KrauseFx:
Related to https://github.com/fastlane/fastlane/issues/13772

### DIFF
--- a/match/lib/match/storage/google_cloud_storage.rb
+++ b/match/lib/match/storage/google_cloud_storage.rb
@@ -154,8 +154,8 @@ module Match
 
         return DEFAULT_KEYS_FILE_NAME if File.exist?(DEFAULT_KEYS_FILE_NAME)
 
-        fastlane_subfolder = File.join(FastlaneCore::FastlaneFolder.path, DEFAULT_KEYS_FILE_NAME)
-        return fastlane_subfolder if File.exist?(fastlane_subfolder)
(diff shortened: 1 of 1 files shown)


match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -152,9 +152,10 @@ def ensure_keys_file_exists(google_cloud_keys_file)
           return google_cloud_keys_file
         end
 
-        if File.exist?(DEFAULT_KEYS_FILE_NAME)
-          return DEFAULT_KEYS_FILE_NAME
-        end
+        return DEFAULT_KEYS_FILE_NAME if File.exist?(DEFAULT_KEYS_FILE_NAME)
+
+        fastlane_subfolder = File.join(FastlaneCore::FastlaneFolder.path, DEFAULT_KEYS_FILE_NAME)
```
armcburney: Should this be called `fastlane_subfolder` if it's a file path? How about `fastlane_folder_gc_keys_path` or `gc_keys_path`? It's a bit more verbose, but I think it makes more sense semantically.

armcburney (approved):
LGTM! Just the one comment ✅

match/lib/match/storage/google_cloud_storage.rb
KrauseFx: Sounds good

PR #13889, opened 2018-12-12
[action] sigh resign: Fix `keychain_path` parameter and output bug

MiB3:
🔑

### Checklist

### Motivation and Context

Currently the `sigh resign` keychain_path parameter is simply passed to `/usr/bin/codesign`. This means the passed keychain needs to already be added to the keychain list (`security list-keychains`), as otherwise the signing_identity will not be found by resign. This pull request does the adding (and afterwards removing) of the passed keychain automatically.

This pull request also fixes the issue that if the signing identity cannot be found, resign outputs `Couldn't find signing identity ''.` instead of `Couldn't find signing identity 'name of the needed signing identity'.`. This is because if the signing identity cannot be found, it is overwritten by the empty string. 

### Description

The keychain is added and removed with `security list-keychains -s #{new|old_keychains}`.

The `signing_identiy` is saved in two instead of one variable so the not-found-one can be still accessed to display it to the user.


MiB3:
I signed it!

sigh/lib/sigh/resign.rb
```diff
@@ -60,8 +70,11 @@ def resign(ipa, signing_identity, provisioning_profiles, entitlements, version,
 
       puts(command.magenta)
       puts(`#{command}`)
+      command_return_value = $?.to_i
```
janpio: why extract this into a variable and move it before the `security` call?

sigh/lib/sigh/resign.rb
```diff
@@ -60,8 +70,11 @@ def resign(ipa, signing_identity, provisioning_profiles, entitlements, version,
...
+
+      `security list-keychains -s #{current_keychains}` if current_keychains
```
janpio: The use of `current` is confusing here. Do you think it makes sense to rename this to `old`, `previous` or `before` or similar?

sigh/lib/sigh/resign.rb
janpio: lol of course. Didn't think properly 💥

sigh/lib/sigh/resign.rb
MiB3: Yes also totally valid in my opinion. I though `current` would be best, but `previous` is also fine.

janpio: (`current` confused me, as at the time of the second usage it is not the current set of keychains any more, as we changed it...)

MiB3: Yes you are right. Should I change all occurrences of `current_keychains` to `previous_keychains`?

janpio: Sounds like a plan 👍

sigh/lib/sigh/resign.rb
```diff
@@ -72,7 +72,7 @@ def resign(ipa, signing_identity, provisioning_profiles, entitlements, version,
       puts(`#{command}`)
       command_return_value = $?.to_i
 
-      `security list-keychains -s #{current_keychains}` if current_keychains
+      `security list-keychains -s #{previous_keychains}` if previous_keychains
```
janpio: Just noticed that this will _always_ be true if `keychain_path` exists, true?  
A bit less stuff executed if the if block in lines 25+ was rewritten to only add the new keychain (and backup the previous ones) if it is not already in there. Does this make sense or did I misudnerstand your code?

MiB3: Yes you are correct.
So the if block could be written like this. This would really only execute each step if it is truly needed, right?
```
if keychain_path
    keychain_path = File.expand_path(keychain_path)

    current_keychains = `security list-keychains`
    current_keychains.delete!("\n")

    unless current_keychains.include?(keychain_path)
        previous_keychains = current_keychains 
        `security list-keychains -s #{current_keychains} '#{keychain_path}'`
    end
end
```

sigh/lib/sigh/resign.rb
```diff
@@ -21,6 +21,16 @@ def self.resign(ipa, signing_identity, provisioning_profiles, entitlements, vers
 
     def resign(ipa, signing_identity, provisioning_profiles, entitlements, version, display_name, short_version, bundle_version, new_bundle_id, use_app_entitlements, keychain_path)
       resign_path = find_resign_path
+
+      if keychain_path
+        keychain_path = File.expand_path(keychain_path)
```
janpio: expanding the keychain path will have no negative influence on the already present uses of that variable further down? (e.g. line 42(old)/52(new))

sigh/lib/sigh/resign.rb
janpio: That looks like it would work to me. (But I am much better at pointing out stuff that might not work than to actually write working ruby code myself, so... 🍶 )

MiB3: Done

janpio (approved):
This look good! Thanks for working on this with me @MiB3 🥇

sigh/lib/sigh/resign.rb
```diff
@@ -60,8 +73,11 @@ def resign(ipa, signing_identity, provisioning_profiles, entitlements, version,
 
       puts(command.magenta)
       puts(`#{command}`)
+      command_return_value = $?.to_i
+
+      `security list-keychains -s #{previous_keychains}` if previous_keychains
```
joshdholtz: Can this get executed inside of an `ensure` at the bottom of this method? Things _look_ save enough where an error/exception will never get thrown but it seems like this is a pretty important step that we **always** want to get executed no matter what. I think? 🤔

joshdholtz (changes requested):
@MiB3 @janpio  One small comment/request on executing the setting of the previous keychain list in an `ensure` block. Lemme know your thoughts 🙃

sigh/lib/sigh/resign.rb
MiB3: Yes, you are right. It does make sense to put this into an ensure block (Although its usually not to much of a problem for the user if an extra keychain added). See my newest commit. It now has the ensure block added.

joshdholtz (approved):
Thanks for adding this! This will go out in today's release 💪

PR #13926, opened 2018-12-17
[snapshot] Create screenshot directory on demand if needed

ghost:
When running snapshot tests locally, my screenshots folder didn't exist yet which led to the screenshots never being saved, these changes fixed that

### Checklist

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -168,8 +168,15 @@ open class Snapshot: NSObject {
             let screenshot = window.screenshot()
             guard let simulator = ProcessInfo().environment["SIMULATOR_DEVICE_NAME"], let screenshotsDir = screenshotsDirectory else { return }
             let path = screenshotsDir.appendingPathComponent("\(simulator)-\(name).png")
+        
             do {
+                var isDirectory: ObjCBool = false
+                if !FileManager.default.fileExists(atPath: screenshotsDir.path, isDirectory: &isDirectory) || !isDirectory.boolValue {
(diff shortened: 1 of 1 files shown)


joshdholtz:
@CalQL8ed-K-OS Heyyy 👋 Just trying to understand how this happened. When you say...
>When running snapshot tests locally...

Do you mean that you were running your snapshot tests directly through Xcode and not dirctly with _snapshot_? The _snapshot_ tool/action should always create this folder for you 🤔

CalQL8ed-K-OS:
@joshdholtz Sorry for the delay, but I think so yes, it was a normal run of `fastlane snapshot`.

joshdholtz:
@CalQL8ed-K-OS No worries on the delay 😊 Do you think would be able to post output of `fastlane snapshot --verbose` here? I'm curious as to what cause the error of the screenshots directory not being there in the first place. This should be all handled within the Ruby part of `snapshot` so even though this fixes it, there could be a bigger underlying problem within the Ruby part 🤔

CalQL8ed-K-OS:
Sorry to say but I don't have access to the repo where I could reproduce the issue any longer, nor did I have any capture of the output from the time of reporting the issue :(

janpio:
@CalQL8ed-K-OS To recreate the underlying problem it should be enough to just create a `snapshot` setup for a new project and run it - and it should fail because of a missing folder. Then with your changes, it should automatically create the required folders and not fail any more - correct?

CalQL8ed-K-OS:
@janpio I'm unclear if there were any other circumstances complicating the issue at the time, but yes, in theory it sounds testable via that method.  If it's not reproducible via that method, I think it's totally appropriate to close the issue as unreproducible.

Issue #13936, opened 2018-12-18
[sigh] recreates the new AdHoc profile in Apple Developer with different names (like "cn.xxxx AdHoc 1544776407"), when we updated Fastlane to 2.111.0


UZheng:
- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

It is a new issue from [#13910](https://github.com/fastlane/fastlane/issues/13910).

Everything works fine before this error report for "Can't find class 'adhoc'([#13905](https://github.com/fastlane/fastlane/issues/13905)).Then I solved the problem according to the instructions to upgrade the version to 2.111.0, but this situation has occurred. 
I replace some important information with some characters, ignore it please.

##### Complete output when running fastlane, including the stack trace and command used

<details>

  <pre>
+ fastlane for_pgyAdhoc  'des: some infomation' isDebug:true

[19:24:14]: Your Fastfile has had smart quotes sanitised. To avoid issues in the future, you should not use TextEdit for editing it. If you are not using TextEdit, you should turn off smart quotes in your editor of choice.
[19:24:14]: ------------------------------
[19:24:14]: --- Step: default_platform ---
[19:24:14]: ------------------------------
[19:24:14]: Driving the lane 'ios start' 🚀
[19:24:14]: ---------------------------
[19:24:14]: --- Step: git branch -l ---
[19:24:14]: ---------------------------
[19:24:14]: $ git branch -l
[19:24:14]: ▸ * (HEAD detached at d2b11c733)
[19:24:14]: ------------------------
[19:24:14]: --- Step: git_branch ---
[19:24:14]: ------------------------
[19:24:14]: -----------------------------
[19:24:14]: --- Step: last_git_commit ---
[19:24:14]: -----------------------------
[19:24:14]: $ 开始执行--> build 
[19:24:14]: --------------------------------------
[19:24:14]: --- Step: Switch to ios build lane ---
[19:24:14]: --------------------------------------
[19:24:14]: Cruising over to lane 'ios build' 🚖
[19:24:14]: build_model: true--->环境为测试环境
[19:24:14]: -----------------
[19:24:14]: --- Step: pwd ---
[19:24:14]: -----------------
[...]

Issue #13916, opened 2018-12-14
[spaceship] Latest version is unable to download ad-hoc provision profiles.


fwigorcferreira:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Fastlane, on version 2.111.0 is unable to download ad-hoc provision profiles. The command fails with `No existing profiles found, that match the certificates you have installed locally!` even if `-z, --skip_certificate_verification` is passed. I double & triple checked, and the provision profile exists in the developer portal. Tried it on a machine that had the valid certificate installed, and the same error was reported

##### Complete output when running fastlane, including the stack trace and command used

<details>
<pre>
Command:

fastlane sigh -z -w -u _username_ -b _team_id_ -a _package_ -n "_provision_profile_name_" -q profile.mobileprovision --ignore_profiles_with_different_name --skip_install --adhoc

Result:

[✔] 🚀
[17:02:14]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile

+-------------------------------------+-------------------------------------+
|                         Summary for sigh 2.111.0                          |
+-------------------------------------+-------------------------------------+
| skip_certificate_verification       | true                                |
| skip_fetch_profiles                 | true                                |
| username                            | _username_                        |
| team_id                             | _team_id_                         |
| app_identifier                      | _package_                         |
| provisioning_name                   | _provision_profile_name_          |
| filename                            | profile.mobileprovision             |
| ignore_profiles_with_different_name | true                                |
| skip_install                        | true                                |
| adhoc                               | true                                |
| developer_id                        | false                               |
| development                         | false                               |
| force                               | false                               |
| platform                            | ios                                 |
| readonly                            | false                               |
+-------------------------------------+-------------------------------------+

[17:02:14]: Starting login with user '_username_'
[17:02:16]: Successfully logged in
[17:02:16]: No existing profiles found, that match the certificates you have installed locally! Creating a new provisioning profile for you
[17:02:17]: Found more than one code signing identity. Choosing the first one. Check out `fastlane sigh --help` to see all available options.
[17:02:17]: Available Code Signing Identities for current filters:
[...]

PR #13907, opened 2018-12-14
[match] fix match adhoc RuntimeError error

machty:
Closes #13905

It seems something changed on the API end that
is suddenly causing "adhoc" to show up as a
distribution method, yet the case statement
wasn't previously handling it.

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
Closes #13905

### Description
Adds a missing "adhoc" case statement to handle new values coming from the API.

PR #13942, opened 2018-12-18
[spaceship] Remove workaround where Adhoc profiles where filtered as appstore

csknns:
After #13907 we can remove the workaround where Adhoc profiles where filtered as appstore. Fixes fastlane/fastlane#13936

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Fixes #13936 and #13916

### Description

After #13907 we can remove the workaround where Adhoc profiles where filtered as appstore.

PR #13945, opened 2018-12-19
[spaceship] Fix Spaceship not returning AdHoc profile

rkrim:
### Checklist

### Motivation and Context
Since something chaned API side that was causing `Can't find class 'adhoc'` and the attempt to fix in #13907 by handling the correct return type for AdHoc profiles, Spaceship was no more returning AdHoc profiles as they where still treated as AppStore profiles.

Link to open issue: #13936 #13916
All test passed.
All codestyle inspection passed

### Description
Remove workaround that treat Adhoc profiles as AppStore profile as it's no more needed
Remove the ProvisioningProfile `is_adhoc?` as it's no more required (known by the type)
Remove sigh filter_profiles_for_adhoc_or_app_store as it's no more needed
Remove ProvisioningProfile is_adhoc related tests
Update other tests
Update Documentation
Update AdHoc ProvisioningProfiles with `adhoc` as distributionMethod

Tested on my own projects with multiple profiles AppStore and Adhoc, all tests where done for both kind of profiles
- With profiles already created manually, detection is OK
- With no profiles, detection of no profile is OK > automatic creation of the right profiles
- With profiles previously created by sigh, detection is OK.

### DIFF
--- a/sigh/lib/sigh/runner.rb
+++ b/sigh/lib/sigh/runner.rb
@@ -81,13 +81,6 @@ module Sigh
 
       # Take the provisioning profile name into account
       results = filter_profiles_by_name(results) if Sigh.config[:provisioning_name].to_s.length > 0
-
-      # Since September 20, 2016 spaceship doesn't distinguish between AdHoc and AppStore profiles
-      # any more, since it requires an additional request
-      # Instead we only call is_adhoc? on the matching profiles to speed up spaceship
-
-      results = filter_profiles_for_adhoc_or_app_store(results)
-
       return results if Sigh.config[:skip_certificate_verification]
 
       UI.message("Verifying certificates...")
@@ -156,18 +149,6 @@ module Sigh
         profiles = filtered
       end
       profiles
-    end
-
-    def filter_profiles_for_adhoc_or_app_store(profiles)
-      profiles.find_all do |current_profile|
-        if profile_type == Spaceship.provisioning_profile.ad_hoc
-          current_profile.is_adhoc?
-        elsif profile_type == Spaceship.provisioning_profile.app_store
-          !current_profile.is_adhoc?
-        else
-          true
-        end
-      end
     end
 
     def certificates_for_profile_and_platform
--- a/spaceship/lib/spaceship/portal/provisioning_profile.rb
+++ b/spaceship/lib/spaceship/portal/provisioning_profile.rb
@@ -328,22 +328,7 @@ module Spaceship
           end
 
           return profiles if self == ProvisioningProfile
-
-          # To distinguish between AppStore and AdHoc profiles, we need to send
-          # a details request (see `profile_details`). This is an expensive operation
-          # which we can't do for every single provisioning profile
-          # Instead we'll treat App Store profiles the same way as Ad Hoc profiles
-          # Spaceship::Portal::ProvisioningProfile::AdHoc.all will return the same array as
-          # Spaceship::Portal::ProvisioningProfile::AppStore.all, containing only AppStore
-          # profiles. To determine if it's an Ad Hoc profile, you can use the
-          # is_adhoc? method on the profile.
-          klass = self
-          klass = AppStore if self == AdHoc
-
-          # only return the profiles that match the class
-          return profiles.select do |profile|
-            profile.class == klass
-          end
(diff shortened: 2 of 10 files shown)


spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -329,16 +329,9 @@ def all(mac: false, xcode: false)
...
-          # a details request (see `profile_details`). This is an expensive operation
-          # which we can't do for every single provisioning profile
-          # Instead we'll treat App Store profiles the same way as Ad Hoc profiles
-          # Spaceship::Portal::ProvisioningProfile::AdHoc.all will return the same array as
-          # Spaceship::Portal::ProvisioningProfile::AppStore.all, containing only AppStore
-          # profiles. To determine if it's an Ad Hoc profile, you can use the
-          # is_adhoc? method on the profile.
+          # Set ProvisioningProfile Type [Spaceship::Portal::ProvisioningProfile::Type]
+          # check factory() for possible values
           klass = self
```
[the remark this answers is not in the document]
rkrim: Right, I was keeping it while debugging.
Nice, thanks.

janpio:
Thanks @rkrim, looks great at first look. Could you double check with #13942, an existing PR that tries to solve the same problem, if this PR could replace that one or if there is something interesting you could also use? Thanks.

rkrim:
Hi @janpio, well both PR are fixing the main problem which is this line
`klass = AppStore if self == AdHoc`
One major difference is about keeping or not `is_adhoc?` in ProvisioningProfile.
In PR #13942, `is_adhoc?` is kept.
While in this one, I decided to remove it as the information is now directly available in the type.
So keeping it does not make sense for me.
By removing `is_adhoc?`, I'm also removing unnecessary ProvisioningProfiles processing in sigh to sync this information as its no more needed.

Well I would say that PR #13942 is included in this one.

sigh/lib/sigh/runner.rb
```diff
@@ -81,13 +81,6 @@ def fetch_profiles
 
       # Take the provisioning profile name into account
       results = filter_profiles_by_name(results) if Sigh.config[:provisioning_name].to_s.length > 0
-
-      # Since September 20, 2016 spaceship doesn't distinguish between AdHoc and AppStore profiles
-      # any more, since it requires an additional request
-      # Instead we only call is_adhoc? on the matching profiles to speed up spaceship
-
-      results = filter_profiles_for_adhoc_or_app_store(results)
```
csknns: Are we sure we need to remove this method?
If I call sigh to download adhoc provision profile with a specific name, and in the dev portal I only have an appstore provision profile with that name, the filtering will prevent this method from returning the profile I did not requested, I think?

rkrim: Hi @csknns,
Previously, yes the method `filter_profiles_for_adhoc_or_app_store()` had the responsibility of returning the right profile between AdHoc and AppStore.
Because AdHoc was treated as AppStore (adhoc profiles objects were instances of the class AppStore same as for appstore profiles) and we needed to read `is_adhoc?` (does the profile have devices in it or not) to make the distinction between them.

Now that is no more the case, `all()` method in `Spaceship::Portal::ProvisioningProfile` is doing the job correctly as it relies directly on the type.

csknns: @rkrim Yep you are right, now filtering by class type works as expected 👍 . This PR seems to better address the issue so I am closing my pull request.

joshdholtz (approved):
This looks great and works great! Thanks for making this fix ❤️

benjarwar:
@rkrim @joshdholtz cross-posting [my comment](https://github.com/fastlane/fastlane/issues/13936#issuecomment-449166674), any advice?

Fastlane is generating `adhoc` provisioning profiles with numbers at the end of their profile names. We've already upgraded to `2.112.0` and tried re-running `match`, but code signing still fails, because the profile that's downloaded for our app ID is like `match AdHoc com.foo.app-name 1545256864` but the cert is expecting `match AdHoc com.foo.app-name`.

@samodom suggested running `fastlane match nuke distribution` and recreating the provisioning profiles. But that's a bit drastic for us since it will invalidate currently used Ad Hoc builds. Do you know of any other solutions?

PR #13946, opened 2018-12-19
Fix support for specifying emulator serial

raymondhoagland:
### Checklist

### Description
The `strip!` function  that was added in #13846 returns nil if the string was not altered. [Ruby docs](https://ruby-doc.org/core-2.2.0/String.html#method-i-strip-21).  The adb action now fails whenever a serial _is_ specified with `no implicit conversion of nil into String` on line 42 (the Actions.sh call).

Since command is being set with the return value, `strip` should be used instead which will always return the new value, regardless of whether it was changed.

Can be easily replicated in irb
```
android_serial = 'emulator-5554'
adb_path = '/test'

command = 'devices'
command = [android_serial, adb_path, command].join(" ").strip!
puts command
# => nil
command = 'devices'
command = [android_serial, adb_path, command].join(" ").strip
puts command
# => emulator-5554 /test devices
```

### DIFF
--- a/fastlane/lib/fastlane/helper/adb_helper.rb
+++ b/fastlane/lib/fastlane/helper/adb_helper.rb
@@ -26,7 +26,7 @@ module Fastlane
       # Run a certain action
       def trigger(command: nil, serial: nil)
         android_serial = serial != "" ? "ANDROID_SERIAL=#{serial}" : nil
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
Looks good! Thanks for making this fix and writing new tests 😊

janpio:
Thanks fir fixing my mess @raymondhoagland 🦆

PR #13919, opened 2018-12-15
[spaceship] replace post_for_testflight_review with new public API

joshdholtz:
Fixes #13903

## Description 
The private API used in `Spaceship::TestFlight`'s `post_for_testflight_review` was taken down by Apple and replaced with the new public official App Store Connect API. This PR is a fix for that.

### Thing to note
- This current solution is probably **NOT** a permanent one (the way that it is currently implemented)
- This needed to get fixed and this works 😊
- I put this in its own new `Spaceship::API` module so that it was easy to replace later on 
- I replaced the internals of `Spacehip::TestFlight::Build` to use the new `Spaceship::API` implementation so that there were less changes in the PR and that we can just get things working again

PR #13918, opened 2018-12-15
[core] remove leftover launch and completion analytics from UpdateChecker

janpio:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
`UpdateChecker` still contains some Fabric analytics code that is used on launch and on completion. Besides the data not being used, it is also causing problems because it is run in a thread and one of its methods uses `Dir.chdir`, which then can lead to race conditions.

### Description
Remove the analytics code from `UpdateChecker`, move `AppIdentifierGuesser` tests into their own file.

PR #13909, opened 2018-12-14
[spaceship] Improve Tests for Spaceship::Portal::ProvisioningProfile#factory

TheMetalCode:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Leftovers from https://github.com/fastlane/fastlane/pull/13907 and https://github.com/fastlane/fastlane/issues/13905 - the author had asked for help with tests and I obliged. This didn't make it into the fix that merged though, which is fine given the severity. Now that that's done, I figured I'd PR this for consideration.

### Description

- Adds tests for each accepted distribution method, plus a test for the expected raised error when the distribution method is not recognized.
- Corrects comment to reflect all the possible distribution methods.

PR #13955, opened 2018-12-19
Version bump to 2.112.0

joshdholtz:
Auto-generated by fastlane 🤖

**Changes since release '2.111.0':**

* Fix Spaceship not returning AdHoc profile (#13945) via Riad Krim
* [docs] sigh: fix typo in options (#13931) via Ivan Smetanin
* [spaceship] replace post_for_testflight_review with new public API (#13919) via Josh Holtz
* [deliver] default html generator to current dir (#13920) via Josh Holtz
* [core] remove leftover launch and completion analytics from UpdateChecker (#13918) via Jan Piotrowski
* [spaceship] Improve Tests for Spaceship::Portal::ProvisioningProfile#factory (#13909) via Jason Hagglund
* New action ensure_bundle_exec to enforce use of bundler (#13866) via Rishabh Tayal


fastlane/swift/Fastlane.swift
```diff
@@ -4061,7 +4065,7 @@ func xcov(workspace: String? = nil,
           coverallsServiceJobId: String? = nil,
           coverallsRepoToken: String? = nil,
           xcconfig: String? = nil,
-          ideFoundationPath: String = "/Applications/Xcode-9.0.1.app/Contents/Developer/../Frameworks/IDEFoundation.framework/Versions/A/IDEFoundation",
+          ideFoundationPath: String = "/Applications/Xcode-10.1.app/Contents/Developer/../Frameworks/IDEFoundation.framework/Versions/A/IDEFoundation",
```
janpio: Huh, what caused this change?

janpio (approved):
See comment, but if this is intentional then LGTM.

fastlane/swift/Fastlane.swift
joshdholtz: Ah, this just happens based on the bumpers xcode version selected. It doesn't really matter at all since this gets regenerated on the client side when they use it 😊

PR #13966, opened 2018-12-20
Add more granular control over shell command output

akashivskyy:
### Checklist

- [x] I've run `bundle exec rspec` from the root directory
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a`

### Motivation and Context

Currently, both `CommandExecutor.execute` and `Actions.sh` re-print the script output in red when it exits with a non-zero status. While this is acceptable with short output _(well... is it?)_, it ruins readability of the build log with long ones. Especially with `gym` and `scan` which, in large projects, generate hundreds (or even thousands when using CocoaPods) of the same lines **twice**.

This pull request solves that issue by allowing more granular control over what's printed and re-printed in various places, while maintaining backwards-compatibility and existing behavior (mostly). 

### Description

I'll divide the description of changes introduced in this pull request by the place they were introduced. 

Note: When I mention "printing error output", I mean the re-printing output of the whole command if it exits with a non-zero status.

#### Actions.sh

This pull request changes the interpretation of `log` argument, allowing users to specify what exactly they want to be printed.

This change is backwards-compatible and it doesn't change default behavior.

| print... | if... |
|:--|:--|
| command | `log == true`<br>`\|\| log.include?(:command)` |
| real-time output | `log == true`<br>`\|\| log.include?(:output)` |
| error output | `log == true`<br>`\|\| log.include?(:error)` |

As a bonus, this enables new possibilities, e.g. to only print command and error output if it exits with a non-zero status (by using `[:comand, :error]`). Currently, it's not possible without providing the error callback and manually raising the error.

📣 **Discussion:** Do you think the existing behavior should be changed so that it's `[:command, :output]` by default? It would mean that the command and its output is still printed but only once. It would also affect all steps using `Actions.sh`.

#### CommandExecutor.execute

A new argument has been added: `suppress_error_output`, allowing users to specify they want to suppress it exclusively.

This change is backwards-compatible and it doesn't change default behavior.

| print... | if... |
|:--|:--|
| command | `print_command == true` |
| loading text | `print_all == true` |
| real-time output | `print_all == true`<br>`&& suppress_output == false` |
| error output | `suppress_output == false`<br>`&& suppress_error_output == false` |

📣 **Discussion:** Do you think the existing behavior should be changed so that it's suppressing only error output? It would mean that the command and its output is still printed but only once. It would also affect all steps using `CommandExecutor.execute`.

#### scan

A new config item has been added to scan: `enabled_xcode_output` which is an array of values `:command`, `:output` and `:error`.

This change is backwards compatible but it changes the behavior. It no longer prints error output if the tests fail, unless `:error` is included in `enabled_xcode_output`.

| print... | if... |
|:--|:--|
| command | `enabled_xcode_output.include?(:command)` |
| real-time output | `suppress_xcode_output == false`<br>`&& enabled_xcode_output.include?(:output)` |
| error output | `suppress_xcode_output == false`<br>`&& enabled_xcode_output.include?(:error)` |

📣 **Discussion:** Do I need to re-generate Swift API for scan as well?

#### gym

A new config item has been added to gym: `enabled_xcode_output` which is an array of values `:command`, `:output` and `:error`.

This change is backwards compatible but it changes the behavior. It no longer prints error output if the tests fail, unless `:error` is included in `enabled_xcode_output`.

| print... | if... |
|:--|:--|
| command | `silent == false`<br>`&& enabled_xcode_output.include?(:command)` |
| real-time output | `silent == false`<br>`&& suppress_xcode_output == false`<br>`&& enabled_xcode_output.include?(:output)` |
| error output | `silent == false`<br>`&& suppress_xcode_output == false`<br>`&& enabled_xcode_output.include?(:error)` |

📣 **Discussion:** Do I need to re-generate Swift API for gym as well?

### Discussion

In the above description I mentioned a couple of discussion items, marked with 📣. I'd like to trigger a conversation on those topics before proceeding with this pull request.

**What do you think?** Does the user really need to see the raw error output if they see it above? Was it done mostly for the ability to copy without the `[hh:mm:ss] >` prefix?

---

P.S. Gosh, that's a long description for `+88 -22`. 😅


akashivskyy:
@googlebot I signed it!

janpio:
Thanks for tackling this, indeed an area where things can be optimized and improved. But also a lot to process 📚 💥

akashivskyy:
@jdee Hi! :wave: Did you have a chance to look at this PR?

joshdholtz:
@akashivskyy This is an amazing change 🙌 I do love the granularity and cleanup that you did here ❤️ 

> 📣 Discussion: Do you think the existing behavior should be changed so that it's [:command, :output] by default? It would mean that the command and its output is still printed but only once. It would also affect all steps using Actions.sh.

- I think the existing behavior here for `Actions.sh` should be `[:command, :output, :error]`? 🤔 Anybody using `Actions.sh` in their `Fastfile` would definitely need to see both the command being executed along with results and any error output that occurs.  

> 📣 Discussion: Do you think the existing behavior should be changed so that it's suppressing only error output? It would mean that the command and its output is still printed but only once. It would also affect all steps using CommandExecutor.execute.

- I think its always good to show error output? In my experience, sometimes not showing error output can lead to more confusion to users. Do you think there are some good reasons to suppress error output? 

> 📣 Discussion: Do I need to re-generate Swift API for scan as well?

You do not 😊 These generate in our build/release process so it will be done automatically next release

> 📣 Discussion: Do I need to re-generate Swift API for gym as well?

You do not 😊 These generate in our build/release process so it will be done automatically next release

akashivskyy:
@joshdholtz To answer your 2 answers regarding the error output, I feel we must use the same glossary here. I wrote in the description of this PR that:

>Note: When I mention "printing error output", I mean the re-printing output of the whole command if it exits with a non-zero status.

Let's redefine that:

- **all output** — Any messages printed by a command to stdout or stderr.
- **error output** — Any error messages printed by a command (subset of above).
- **re-printing output** — The behavior of Fastlane re-printing **all output** when a command fails.

With the above definitions, `:output` controls **all output** and `:error` controls **re-printing output**. So, in other words, if `Actions.sh` is given `:output`, the error messages are still printed as part of real-time output.

I prepared a table to visualise what's happening:



The current default behavior of `Actions.sh` and `CommandExecutor.execute` is the bottom one, which means users will see the same output **twice** if a command fails. That's why I'm suggesting changing that to the top one so that they only see it once.

joshdholtz (changes requested):
@akashivskyy Thanks for explanation! Sorry that I missed that earlier 😬 I don't think I'm personally tied to either approach here. If I get an error I usually scroll up anyway but in the case where a user's configuration shows the fastlane what's new text they would be forced to scroll up to see there error (see below image)

Hopefully user's update to latest right away or hide the output if they don't want to see it but its something that might cause a pretty big opinion about the error being gone 🤔 

**Need to scroll**
screen shot 2019-02-11 at 8 10 00 pm

**No need to scroll cause hidden update text**
screen shot 2019-02-11 at 8 15 29 pm

I think I see the value in hiding it if there isn't any large changelog text but there is value in seeing it again if there is a large changelog.

**TL;DR**
I think the use case for it being hidden may cause more confusion/work for majority of users which would be forced to scroll up for an error than the benefit it would provide preventing it from showing a second time... even though I personally wouldn't mind it being hidden 😉 


Also also, I forgot to submit my line comments/change requests from last time 🙃

scan/lib/scan/options.rb
```diff
@@ -193,9 +194,18 @@ def self.available_options
                                      optional: true),
         FastlaneCore::ConfigItem.new(key: :suppress_xcode_output,
                                      env_name: "SCAN_SUPPRESS_XCODE_OUTPUT",
-                                     description: "Suppress the output of xcodebuild to stdout. Output is still saved in buildlog_path",
+                                     description: "Suppress the output of xcodebuild to stdout. Takes precedence over `enabled_xcode_output`. Output is still saved in buildlog_path",
                                      optional: true,
                                      is_string: false),
```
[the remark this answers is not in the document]
akashivskyy: Yeah, I noticed some info about the deprecation of `is_string` but thought that acting on it is not in the scope of this PR. Should it be?

gym/lib/gym/runner.rb
```diff
@@ -83,16 +83,24 @@ def fix_generic_archive
 
     def mark_archive_as_built_by_gym(archive_path)
       escaped_archive_path = archive_path.shellescape
-      system("xattr -w info.fastlane.generated_by_gym 1 #{escaped_archive_path}")
+      system("xattr -w info.fastlane.generated_by_gym 1 #{escaped_archive_path} > /dev/null 2>&1")
```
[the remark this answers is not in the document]
akashivskyy: To be honest, I don't remember why I did that. Seems unnecessary to me now. Will revert.

akashivskyy:
@joshdholtz Hi! Sorry for the late reply. I completely understand your rationale that changing the default behavior now would bring more confusion than benefits. I felt that as well and you reassured me.

In addition, I refined the way _gym_ executes commands — now, it always uses `CommandExecutor` instead of `Helper.backticks`. I think this justifies a re-request for review. 😊

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -2,30 +2,64 @@
...
   module Actions
-    # Execute a shell command
-    # This method will output the string and execute it
-    # Just an alias for sh_no_action
-    # When running this in tests, it will return the actual command instead of executing it
-    # @param log [Boolean] should fastlane print out the executed command
-    # @param error_callback [Block] a callback invoked with the command output if there is a non-zero exit status
+    # Execute a shell command. This method will output the string and execute it.
+    #
+    # The `log` argument accepts either a boolean or an aray of symbols
```
janpio:
```suggestion
    # The `log` argument accepts either a boolean or an array of symbols
```

fastlane/lib/fastlane/helper/sh_helper.rb
```diff
@@ -2,30 +2,64 @@
...
+        print_command_error: parse_log_argument.call(:error),
+        error_callback: error_callback,
+        &b
+      )
     end
 
+    # Just an alias for `sh`.
     def self.sh_no_action(*command, log: true, error_callback: nil, &b)
-      sh_control_output(*command, print_command: log, print_command_output: log, error_callback: error_callback, &b)
+      sh(*command, log: log, error_callback: error_callback, &b)
```
janpio: why the change here?

scan/lib/scan/options.rb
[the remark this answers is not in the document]
janpio: No, there is a separate PR that will take care of this. When that one is merged, this will also trigger rubocop to complain - so better get this merged before the other one ;)

fastlane/lib/fastlane/helper/sh_helper.rb
akashivskyy: As `sh_no_action` is supposed to be an alias for `sh` (they had exactly the same implementations), I didn't want to duplicate the code in `sh` and `sh_no_action`. So I made `sh_no_action` call `sh`.

If you feel that this added level of indirection is bad, I can copy&paste the changed implementation from `sh` here as well.

janpio: Nah, I was just wondering why the change is needed now. Wasn't this the case before as well?

janpio: got it.

akashivskyy:
I lost interest in pursuing this — closing for now so it doesn't create clutter. Maybe I'll reopen this some day. 😉

Issue #6866, opened 2016-11-04
Keychain prompts for permission to use private key regardless of access control settings on sierra


godrei:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Issue Description

Fastlan gym hangs on `xcodebuild archive` commands.

Keychain prompts for permission to use private key regardless of access control settings on sierra this cases a hang on build servers when tools calling `codesign` tool, like signing an archive, or during copying a framework or embedding pods.

Related [openradar issue](https://openradar.appspot.com/28524119).

The mentioned solution seems to be working, our implementation: https://github.com/bitrise-io/steps-certificate-and-profile-installer/blob/master/main.go#L548

# Log 1:

```
+----------------------+-------------------------+
|             [32mSummary for gym 1.11.3[0m             |
+----------------------+-------------------------+
| scheme               | [REDACTED]                 |
| configuration        | Alpha                   |
| use_legacy_build_api | true                    |
| export_method        | development             |
| workspace            | ./[REDACTED].xcworkspace   |
| destination          | generic/platform=iOS    |
| output_name          | [REDACTED]                 |
| clean                | false                   |
| output_directory     | .                       |
| silent               | false                   |
| buildlog_path        | ~/Library/Logs/gym      |
| xcode_path           | /Applications/Xcode.app |
+----------------------+-------------------------+

[07:58:19]: [4m[36m$ set -o pipefail && xcodebuild -workspace ./[REDACTED].xcworkspace -scheme [REDACTED] -configuration 'Alpha' -destination 'generic/platform=iOS' -archivePath /Users/vagrant/Library/Developer/Xcode/Archives/2016-11-03/[REDACTED]\ 2016-11-03\ 07.58.19.xcarchive archive | tee /Users/vagrant/Library/Logs/gym/[REDACTED]-[REDACTED].log | xcpretty[0m[0m
[07:58:24]: [35m[33m▸[0m [39;1mBuilding[0m [REDACTED]/[REDACTED]-notification [(Release)][0m
[07:58:24]: [35m[33m▸[0m [39;1mCheck Dependencies[0m[0m
[07:58:24]: [35m[33m▸[0m [39;1mProcessing[0m Info.plist[0m
[07:58:24]: [35m[33m▸[0m [39;1mCompiling[0m NotificationService.swift[0m
[...]

PR #13967, opened 2018-12-20
check exit status of security command

snatchev:
### Checklist

### Motivation and Context
During a keychain import, the `security set-key-partition-list` would fail silently if the keychain password was not supplied correctly.

This would cause `CodeSign` to have a gui popup requesting permissions to access the keychain.

This command failed silently because `Helper.backticks` never considers the exit status of the command.

Related Issue: #6866

### Description
This PR removes the usage of `Helper.backticks` to use `popen3` which gives better control over stdout, stderr and the exit status.

Previously this command would fail silently but now it will raise an error:
```
Could not configure key to bypass permission popup:
security: SecKeychainItemSetAccessWithPassword: The user name or passphrase you entered is not correct
```

If you see this error, you may need to provide the `MATCH_KEYCHAIN_PASSWORD` environment variable. If the keychain is `login.keychain` then the password would be the user's login password. If you cannot or do not want to use this password, please consider adding a new keychain.

### DIFF
--- a/fastlane/spec/actions_specs/import_certificate_spec.rb
+++ b/fastlane/spec/actions_specs/import_certificate_spec.rb
@@ -14,13 +14,13 @@ describe Fastlane do
         expected_command = "security import #{cert_name} -k '#{keychain_path}' -P #{password} -T /usr/bin/codesign -T /usr/bin/security &> /dev/null"
 
         # this command is also sent on macOS Sierra and we need to allow it or else the test will fail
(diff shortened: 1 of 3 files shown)


nakhbari (approved):
LGTM

joshdholtz:
@snatchev This is actually a breaking change for a lot of users that rely on `login.keychain` 😬  (This has broken all of my _match_ setups that I have. All of my imports are failing since I don't specify a `MATCH_KEYCHAIN_PASSWORD` and I don't think I should have to if I'm using a login keychain 🤔

PR #13974, opened 2018-12-21
[core] added loading indicator for importing keys command

RishabhTayal:
Sometimes it takes a lot of time to run command `security set-key-partition-list` in https://github.com/fastlane/fastlane/blob/f9db58aa8059cf2a47cb00939e9566efeffe69cc/fastlane_core/lib/fastlane_core/keychain_importer.rb#L20-L20

Adding a loading indicator to give feedback to the user that the keys are being imported.

### DIFF
--- a/fastlane_core/lib/fastlane_core/keychain_importer.rb
+++ b/fastlane_core/lib/fastlane_core/keychain_importer.rb
@@ -24,8 +24,10 @@ module FastlaneCore
 
         Open3.popen3(command) do |stdin, stdout, stderr, thrd|
           if output
+            Helper.show_loading_indicator("Importing keys...")
             UI.command(command)
             UI.command_output(stdout.read)
+            Helper.hide_loading_indicator
           end
 
           unless thrd.value.success?


janpio:
What does that look like? What are the reasons for this command needing so long?

RishabhTayal:
@snatchev has a [branch](https://github.com/fastlane/fastlane/tree/speed-up-keychain-importer) which reduces the time so this PR might not be relevant. @snatchev can make the decision.

RishabhTayal:
Actually, I test @snatchev's branch and there is still some processing time for the command. It's significantly improved but there is still some duration. I think it's better to add a loading indicator.

joshdholtz (changes requested):
@RishabhTayal Can you rebase this so that this is ontop of the commits with the improved import times? 🙃 My import isn't instant so I'd like to see what this is like with that. How long does this take for you?

RishabhTayal:
Done.

RishabhTayal:
@Joshdholtz its ready to review again.

joshdholtz (approved):

PR #14005, opened 2018-12-31
[action] Fix version regex in increment_version_number

chuganzy:
### Checklist

### Motivation and Context

Noticed that regexes to validate version format between https://github.com/fastlane/fastlane/pull/14005/files#diff-5a9635727c1c2c153502b19faacb2320L29 and https://github.com/fastlane/fastlane/pull/14005/files#diff-5a9635727c1c2c153502b19faacb2320L37 are different and both are not strict enough.

### Description

This PR fixes the regex for version format check and also write tests for the case.

### DIFF
--- a/fastlane/lib/fastlane/actions/increment_version_number.rb
+++ b/fastlane/lib/fastlane/actions/increment_version_number.rb
@@ -23,11 +23,14 @@ module Fastlane
           '&&'
         ].join(' ')
 
-        current_version = Actions.sh(
-          "#{command_prefix} agvtool what-marketing-version -terse1",
-          log: FastlaneCore::Globals.verbose?,
-          error_callback: ->(result) { current_version = '' }
-        ).split("\n").last
+        begin
+          current_version = Actions
+                            .sh("#{command_prefix} agvtool what-marketing-version -terse1", log: FastlaneCore::Globals.verbose?)
+                            .split("\n")
+                            .last
+        rescue
+          current_version = ''
+        end
 
         version_regex = /^\d+\.\d+\.\d+$/
         if params[:version_number]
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/increment_version_number.rb
```diff
@@ -23,20 +23,24 @@ def self.run(params)
           '&&'
         ].join(' ')
 
-        current_version = `#{command_prefix} agvtool what-marketing-version -terse1`.split("\n").last || ''
+        begin
+          current_version = Actions
+                            .sh("#{command_prefix} agvtool what-marketing-version -terse1", log: FastlaneCore::Globals.verbose?)
+                            .split("\n")
+                            .last
```
joshdholtz: We _may_ also need to call `.strip` here with the addition of `^` and `$` to the regex. Thoughs? 🤔

fastlane/lib/fastlane/actions/increment_version_number.rb
```diff
@@ -23,20 +23,24 @@ def self.run(params)
...
 
+        version_regex = /^\d+\.\d+\.\d+$/
         if params[:version_number]
-          UI.verbose("Your current version (#{current_version}) does not respect the format A.B.C") unless current_version =~ /\d.\d.\d/
+          UI.verbose("Your current version (#{current_version}) does not respect the format A.B.C") unless current_version =~ version_regex
 
           # Specific version
           next_version_number = params[:version_number]
         else
-          if Helper.test?
```
joshdholtz: LOVE the removal of this `Helper.test?` part ❤️

joshdholtz (changes requested):
Perhaps also add a `.strip` because shell output is unpredictable when parsing 😊

fastlane/lib/fastlane/actions/increment_version_number.rb
chuganzy: So far `agvtool` does not put any space around but it's definitely safer to put that! Will do.

chuganzy: Done!

joshdholtz (approved):
Looks good! Thanks for making this change ❤️

Issue #12566, opened 2018-05-21
Crashlytics lane fails on first run


vincentbrison:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When using [crashlytics action](https://docs.fastlane.tools/actions/crashlytics/) to upload an Android build to crashlytics beta, the first run always crash.

Given the stacktrace, it looks like a temp file does not exist when `com.crashlytics.tools` jar is running, causing the crash.

We use the following configuration:
```
lane :beta do
    gradle(task: 'assemble', build_type: 'beta')
    crashlytics(
        api_token: 'an_api_token',
        build_secret: 'a_build_secret',
        groups: 'a_group_of_testers',
        notes: 'some changelog'
    )
end
```

The issue only occurs when at the first time we run `fastlane beta`, when `notes` parameter is set. On the second run, the temp file is here, and the task complete successfully.

#### Our understanding
From the sources of [crashlytics.rb](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/crashlytics.rb), when using `notes` parameter, a temp file is used to store the notes, before passing the path of this file to crashlytics dev tools which itself do the upload to beta store.
From [crashlytics_helper.rb](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/helper/crashlytics_helper.rb), this temp file is generated using [ruby `tempfile` class](https://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html).

So it may looks like the first time we use `fastlane beta`, something is cleaning the temp folder before the upload by crashlytics dev tools. The only difference between the first run of `fastlane beta` and subsequent runs is that [crashlytics_helper.rb](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/helper/crashlytics_helper.rb) will call `download_android_tools` to download crashlytics dev tools jar.
[...]

PR #14037, opened 2019-01-08
[action] explicitly unlink to keep references to temp files for crashlytics

chuganzy:
### Checklist

### Motivation and Context

Try to solve https://github.com/fastlane/fastlane/issues/12566

### Description

Apparently sometimes GC runs before running actual sh and crashlytics action fails randomly because temp files are deleted..

### DIFF
--- a/fastlane/lib/fastlane/helper/crashlytics_helper.rb
+++ b/fastlane/lib/fastlane/helper/crashlytics_helper.rb
@@ -56,7 +56,7 @@ module Fastlane
           return command
         end
 
-        def generate_android_command(params)
+        def generate_android_command(params, android_manifest_path)
           params[:crashlytics_path] = download_android_tools unless params[:crashlytics_path]
 
           UI.user_error!("The `crashlytics_path` must be a jar file for Android") unless params[:crashlytics_path].end_with?(".jar") || Helper.test?
@@ -71,6 +71,7 @@ module Fastlane
           command << "-apiKey #{params[:api_token]}"
           command << "-apiSecret #{params[:build_secret]}"
           command << "-uploadDist #{File.expand_path(params[:apk_path]).shellescape}"
+          command << "-androidManifest #{File.expand_path(android_manifest_path).shellescape}"
 
           # Optional
           command << "-betaDistributionEmails #{params[:emails].shellescape}" if params[:emails]
@@ -115,6 +116,11 @@ module Fastlane
           end
 
           return jar_path
+        end
+
+        def generate_android_manifest_tempfile
+          # We have to generate an empty XML file to make the crashlytics CLI happy :)
+          write_to_tempfile('<?xml version="1.0" encoding="utf-8"?><manifest></manifest>', 'xml')
         end
 
         def write_to_tempfile(value, tempfilename)
--- a/fastlane/lib/fastlane/helper/crashlytics_helper.rb
+++ b/fastlane/lib/fastlane/helper/crashlytics_helper.rb
@@ -57,12 +57,6 @@ module Fastlane
         end
 
         def generate_android_command(params)
-          # We have to generate an empty XML file to make the crashlytics CLI happy :)
-          require 'tempfile'
-          xml = Tempfile.new('xml')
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/helper/crashlytics_helper.rb
```diff
@@ -57,12 +57,6 @@ def generate_ios_command(params)
         end
 
         def generate_android_command(params)
-          # We have to generate an empty XML file to make the crashlytics CLI happy :)
```
janpio: unrelated deletion?

chuganzy: Moved to https://github.com/fastlane/fastlane/pull/14037/files#diff-23c7195bf7a4d9b35a0a2076e247b544R30

fastlane/lib/fastlane/helper/crashlytics_helper.rb
```diff
@@ -57,12 +57,6 @@ def generate_ios_command(params)
...
-          require 'tempfile'
-          xml = Tempfile.new('xml')
-          xml.write('<?xml version="1.0" encoding="utf-8"?><manifest></manifest>')
-          xml.close
-
```
joshdholtz: @chuganzy Can we find a way to keep this inside of `generate_android_command`? It feels disconnect to pull this code out of this method. It would be great if there was a way for us to not have to do that 😇

fastlane/lib/fastlane/helper/crashlytics_helper.rb
```diff
@@ -77,7 +71,6 @@ def generate_android_command(params)
           command << "-apiKey #{params[:api_token]}"
           command << "-apiSecret #{params[:build_secret]}"
           command << "-uploadDist #{File.expand_path(params[:apk_path]).shellescape}"
-          command << "-androidManifest #{xml.path.shellescape}"
```
joshdholtz: Same here. Keeping this here with the other parts of this commands is definitely preferred 👌

joshdholtz (changes requested):
Good find on this! I think the solution here could be tweaked a little bit though to make future maintenance a little be easier 🙃 See my comments on some of the lines but extracting out the XML tempfile creation from `generate_android_command` separates those bits of code/knowledge which feels a bit dangerous. If we could put that back together that would be great.

There are a few different implementations we could do to fix that. We could add an array paramter onto `write_to_tempfile` which takes an array which the tempfile gets appended to. Another option is to add a block onto these method which returns the tempfile.  Either of this might be a more elegant solution so that the`generate_android_command` commands can all stay together 😊

Let me know if you have any other thoughts!

fastlane/lib/fastlane/helper/crashlytics_helper.rb
chuganzy: Hmm, there are many ways to solve that.
1: Make `generate_xxx_command` to return hash or some class that has `command` and `tempfiles`.
2: Make `generate_xxx_command` to receive block so that these functions can determine the scope.
3: Add new `android_manifest_path` parameter to this action and generate empty manifest in action as we do for the release notes if not specified.

No.3 might be the best. What do you think?

chuganzy:
> write_to_tempfile which takes an array which the tempfile gets appended to

so you mean something like this?:

```ruby
tmps = []
params[:note_path] = write_to_tempfile(reelase_note, 'notes', tmps).path
command = generate_xxx_command(params, tmps)
sh(command)
tmps.each(&:unlink)
```

> add a block onto these method which returns the tempfile
I thought that way first but it makes the nest deeper..
```ruby
def generate_xxx_command(params)
	tmp = ...
	command = ...
	yield command
	tmp.unlink
end

generate_xxx_command(params) do |command|
	sh(command)
end
other_tmps.each(&:unlink)
```

joshdholtz:
@chuganzy I _thhiiinnnkkkkk_ passing in a tempfile path might be the best bet 🤔 That seems to be the cleanest approach to me! Let's give that a shot 😊

joshdholtz (approved):
This is 💯 Thanks for working with me on those changes ❤️

PR #14054, opened 2019-01-13
[gym] Detect true build configuration for archive action

akashivskyy:
### Checklist

- [x] I've run `bundle exec rspec` from the root directory.
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a`.
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md).

### Motivation and Context

>If no build configuration is specified and -scheme is not passed then _[whatever is set as default configuration for command-line builds]_ is used.
>
>— [Technical Note TN2339: Building from the Command Line with Xcode FAQ, Listing 5](https://developer.apple.com/library/archive/technotes/tn2339/_index.html#//apple_ref/doc/uid/DTS40014588-CH1-SOURCECODE8)

Currently, Gym doesn't work like that when detecting the provisioning profile mapping. It actually goes straight to the default configuration for command-line builds **even if a scheme is given**.

This is unexpected behavior for projects which specify different bundle identifiers and code signing settings per scheme – as Gym, unlike Xcode, fails to find the proper provisioning profile mapping.

This pull request fixes that issue and makes Gym behave in the same way as Xcode.

### Description

In short, Gym now checks what configuration is set on the scheme used to archive the app, instead of just falling back directly to the "default for command-line builds" or, what's worse, to the first one it finds.

It does so by loading the scheme file using `Xcodeproj::XCScheme` and extracting the values from it. If no scheme was given or if there's a problem loading it, it falls back to existing behavior.

A new test case has been added to make sure the new behavior works as expected.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -55,7 +55,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = Dir["*/lib"]
 
   spec.add_dependency('slack-notifier', '>= 2.0.0', '< 3.0.0') # Slack notifications
-  spec.add_dependency('xcodeproj', '>= 1.6.0', '< 2.0.0') # Modify Xcode projects
+  spec.add_dependency('xcodeproj', '>= 1.8.1', '< 2.0.0') # Modify Xcode projects
   spec.add_dependency('xcpretty', '~> 0.3.0') # prettify xcodebuild output
   spec.add_dependency('terminal-notifier', '>= 1.6.2', '< 2.0.0') # macOS notifications
   spec.add_dependency('terminal-table', '>= 1.4.5', '< 2.0.0') # Actions documentation
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -103,8 +103,8 @@ module FastlaneCore
     # returns the Xcodeproj::Workspace or nil if it is a project
     def workspace
       return nil unless workspace?
+
       @workspace ||= Xcodeproj::Workspace.new_from_xcworkspace(path)
-      @workspace.load_schemes(path)
       @workspace
     end
 
--- a/gym/lib/gym/code_signing_mapping.rb
+++ b/gym/lib/gym/code_signing_mapping.rb
@@ -91,9 +91,30 @@ module Gym
       return destination_sdkroot.include?(sdkroot)
     end
 
+    def detect_configuration_for_archive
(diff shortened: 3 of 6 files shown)


akashivskyy:
I rebased and force-pushed to trigger a rebuild of macOS job, since it's fixed now.

joshdholtz:
@akashivskyy Closing and opening the PR _should_ also trigger a rebuild 😊

joshdholtz:
@akashivskyy This PR looks 💯 I was planning on cutting a new release tonight but I want to create an example project and play with this change a little more so I know how this could effect current users. But just wanted to let you know I saw this 👀

joshdholtz:
@akashivskyy Finally got around to finish reviewing this PR! I came across a bug though 🙃 `xcodeproj` was actually returning the wrong path for the xcode project's for my schemes.

I created a PR over on `xcodeproj` to fix this though 💪  - https://github.com/CocoaPods/Xcodeproj/pull/656

This also requires one small change on our side to remove https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/project.rb#L107.

Are you cool if I push to this PR with my change and my `xcodeproj` version bump once my fix gets through and a new release is made? Thanks! 🚀

joshdholtz:
And sorry about this taking so long! This scheme bug I actually "fixed" in 4 different places until I found the root cause of the problem 🙃

akashivskyy:
@joshdholtz No problem! I know you guys have a massive load of issues and pull requests to manage here, I'm glad you found time to handle mine here. 😊

I don't have any problem with you pushing the fixes, how would you like to proceed? Should I change the base of this PR to some other branch so that you can then push to it?

joshdholtz:
@akashivskyy I can push a commit directly to your branch but just waned to ask permission anyway 😉 And thanks for understanding! ❤️

joshdholtz:
@akashivskyy Heyyyy 👋  We got that new release from `xcodeproj` for my fix in https://github.com/CocoaPods/Xcodeproj/pull/656 🚀  Do you mind reviewing / testing this out to make sure this works like you intended? 😊 Then we can ship this!

akashivskyy:
@joshdholtz I confirm that everything works as expected. :shipit:

joshdholtz (approved):
Thanks for this fix and waiting patiently for our dependencies to update ❤️  This will get shipped out early this week!

Issue #14039, opened 2019-01-09
[Regression] Breaking change when importing a new cert in 2.113.0, worked in 2.112.0


nemui:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
There is an error message when importing a new certificate:
```
Could not configure imported keychain item (certificate) to prevent UI permission popup when code signing
Check if you supplied the correct `keychain_password` for keychain: `/Users/dmitry/Library/Keychains/fastlane.keychain-db`

[17:59:19]: 
[17:59:19]: Please look at the following docs to see how to set a keychain password:
[17:59:19]:  - https://docs.fastlane.tools/actions/sync_code_signing
[17:59:19]:  - https://docs.fastlane.tools/actions/get_certificates
```

This leads to build failures in CI because of the UI permission popup.
Keychain password in supplied in the Match file, as per details below.

##### Command executed

```
fastlane certificates
```
^ this installs certificates for various build types
##### Complete output when running fastlane, including the stack trace and command used
<details>
  <pre>:fastlane dmitry$ fastlane certificates
[09:51:41]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[09:51:43]: ----------------------------------------
[09:51:43]: --- Step: Verifying fastlane version ---
[09:51:43]: ----------------------------------------
[09:51:43]: Your fastlane version 2.113.0 matches the minimum requirement of 2.44.1  ✅
[09:51:43]: ------------------------------
[09:51:43]: --- Step: default_platform ---
[09:51:43]: ------------------------------
[09:51:43]: Driving the lane 'certificates' 🚀
[09:51:43]: -----------------------------
[09:51:43]: --- Step: delete_keychain ---
[09:51:43]: -----------------------------
[09:51:43]: -----------------------------
[09:51:43]: --- Step: create_keychain ---
[09:51:43]: -----------------------------
[09:51:43]: $ security list-keychains -d user
[09:51:43]: ▸ "/Users/dmitry/Library/Keychains/login.keychain-db"
[09:51:43]: -------------------
[09:51:43]: --- Step: match ---
[...]

PR #14056, opened 2019-01-14
[match] fix/improve import keys/certificates now showing UI permission popup

joshdholtz:
🤞 #14039 🤞 

## Fixes
1. Reroutes only stdout to `/dev/null` but not stderr since we need to show error messages
2. Removed `-l 'Imported Private Key'` as it seemed this was too specific and still caused the UI permission popup when code signing
3. Only showing specific error message about code signing when security error is `SecKeychainItemSetAccessWithPassword`

### DIFF
--- a/fastlane_core/lib/fastlane_core/keychain_importer.rb
+++ b/fastlane_core/lib/fastlane_core/keychain_importer.rb
@@ -10,7 +10,7 @@ module FastlaneCore
       command << " -P #{certificate_password.shellescape}"
       command << " -T /usr/bin/codesign" # to not be asked for permission when running a tool like `gym` (before Sierra)
       command << " -T /usr/bin/security"
-      command << " &> /dev/null" unless output
+      command << " 1> /dev/null" unless output
 
       Helper.backticks(command, print: output)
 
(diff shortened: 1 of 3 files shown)


fastlane_core/lib/fastlane_core/keychain_importer.rb
```diff
@@ -19,31 +19,37 @@ def self.import_file(path, keychain_path, keychain_password: "", certificate_pas
...
         command = "security set-key-partition-list"
         command << " -S apple-tool:,apple:"
-        command << " -l 'Imported Private Key'"
         command << " -k #{keychain_password.to_s.shellescape}"
         command << " #{keychain_path.shellescape}"
-        command << " &> /dev/null" # always disable stdout. This can be very verbose, and leak potentially sensitive info
+        command << " 1> /dev/null" # always disable stdout. This can be very verbose, and leak potentially sensitive info
+
+        # Showing loading indicator as this can take some time if a lot of keys installed
+        Helper.show_loading_indicator("Setting key partition list... (this can take a minute if there are a lot of keys installed)")
```
KrauseFx: Niceee

KrauseFx (approved):

PR #14060, opened 2019-01-14
Update docs of each action to mention CLI call option

KrauseFx:
This will make it easier to know that you can use call every fastlane action from the CLI also

---

screen shot 2019-01-14 at 11 29 06 am

### DIFF
--- a/fastlane/lib/assets/ActionDetails.md.erb
+++ b/fastlane/lib/assets/ActionDetails.md.erb
@@ -74,7 +74,9 @@ To pass parameters, make use of the `:` symbol, for example
 fastlane run <%= @action.action_name %> parameter1:"value1" parameter2:"value2"
 ```
 
-It's important to note that the CLI doesn't support setting all kinds of parameters (e.g. Arrays are not supported). It is recommended to add all _fastlane_ actions you use to your `Fastfile`.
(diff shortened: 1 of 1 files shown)


fastlane/lib/assets/ActionDetails.md.erb
```diff
@@ -50,12 +50,37 @@ Key | Description | Default
...
+fastlane run <%= @action.action_name %>
+```
+
+To pass parameters, make use of the `:` symbol, for example
+
+```no-highlight
+fastlane run <%= @action.action_name %> parameter1:"value1" parameter2:"value2"
+```
+
+It's important to note that the CLI doesn't support setting all kinds of parameters (e.g. Arrays are not supported). It is recommended to add all _fastlane_ actions you use to your `Fastfile`.
```
joshdholtz: @KrauseFx Arrays are actually supported if they are defined with `type: Array` through a comma delimited string (ie: `value:"1,2,3"` will get parsed as `["1","2","3"]`). `type: Hash` is currently not supported via CLI though.

joshdholtz (changes requested):

fastlane/lib/assets/ActionDetails.md.erb
KrauseFx: Can you update the PR directly so that the phrasing is correct?

Issue #14094, opened 2019-01-18
[frameit] Automate measuring of "slot" offsets and width


janpio:
Currently we have to manually measure the offsets and width of the "slot" where the screenshots will be placed. We should automate that as part of the frame update process.

Manual process: E.g. use paint.net to open the file, then use the magic wand tool on the slot area to get both the offsets and dimensions, write that data to the file `offsets.json`.

The file used to update the frames is this: https://github.com/fastlane/fastlane/blob/master/frameit/frames_generator/Rakefile `MiniMagick` is already used to trim the whitespace around the device frames (which has its own problems, see #14093), maybe there is a way to "measure" an area similar to the "magic wand technique" described above as well.

janpio:
iPhone X, XR, XS, XS Max have rounded corners (and that silly notch) which luckily didn't worry paint.net at all - but might complicate a solution via code a bit. Just something to keep in mind.

janpio:
This could be a starting point: http://www.fmwconcepts.com/imagemagick/magicwand/index.php (especially see the commands behind the scenes at the bottom)

joshdholtz:
@janpio I looovvvveeee this idea so much! I had to do the same manual process awhile back and it was not fun and didn't scale very well 😉

mollyIV:
👋 after some docs tickets, I guess it's about time to do some coding 😅 

@janpio  could you please give me a bit more context about `frames_generator` steps? 
As far as I understood, you modify `offsets.json` file manually once you download the images from [Facebook frameset](https://facebook.design/devices) by running `rake` command? 🤔 

The goal of this task is to put the correct values into `offsets.json` file automatically. It's not about getting rid of the json file, right? I can imagine we are going to add some code (`MiniMagick` magic wand) before writing the offsets to the json file:

```ruby
...

cp("offsets.json", File.join(output, "offsets.json"))
  puts("Copied over offsets.json")
```

🙏 🙇

janpio:
I have actually never generated frames myself, so @joshdholtz has to fill you in on that, he did that just recently. But the rest of what you wrote follow my understanding - when we update the frames, the process should automatically update `offsets.json` for you to confirm manually. `MiniMagick` is already used in some other places of frameit, so it makes sense to also use it here, and has that magicwand thing that looked similar to what I used manually in paint.net.

mollyIV:
[...]

PR #14095, opened 2019-01-18
[frameit] Fix offsets.json for frames

janpio:
Fix order of devices and offsets+width by manually measuring it with paint.net.

We should not need to do that by hand, so I created #14094


joshdholtz (approved):
👍

joshdholtz (approved):
I like this 👍 I will create a new PR over in https://github.com/fastlane/frameit-frames with these offsets 💪

PR #14099, opened 2019-01-18
[docs] Improve documentation for get_version_number

lordcodes:
### Checklist

### Motivation and Context

When I first used the get_version_number action on a project, there was an issue with it getting the wrong version number due to a stray number in the configuration of one of the project's frameworks. I also felt some extra detail on the options you can provide are helpful.

### Description

Add clearer documentation of options that can be specified.
Add explanation of an issue that can occur if there are multiple targets that have different version numbers specified. When this happens, the wrong version number can sometimes be returned.

### DIFF
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -106,14 +106,17 @@ module Fastlane
       end
 
       def self.details
-        "This action will return the current version number set on your project."
+        [
+          "This action will return the current version number set on your project.",
+          "Make sure that any frameworks or test targets don't have different version numbers from the main project, as it can cause the wrong version number to be returned. Note that some references to the version may not show within the Xcode UI, so worth checking the project file XML and Info.plist files to be sure too."
+        ].join("\n")
       end
 
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :xcodeproj,
                              env_name: "FL_VERSION_NUMBER_PROJECT",
-                             description: "optional, you must specify the path to your main Xcode project if it is not in the project root directory",
+                             description: "Path to the main Xcode project to read version number from, optional. By default will use the first Xcode project found within the project root directory",
                              optional: true,
                              verify_block: proc do |value|
                                UI.user_error!("Please pass the path to the project, not the workspace") if value.end_with?(".xcworkspace")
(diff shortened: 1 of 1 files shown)


joshdholtz (commented):
Thanks for making these doc changes ❤️ I made push a commit that updated a bit of the wordings but otherwise these look like they will be helpful 🙌 

I had one quick question about the following though:
>Make sure that any frameworks or test targets don't have different version numbers from the main project to prevent the wrong version number to be returned.

Question 1:
We do filter out test targets already in the code. Was your setup getting test targets pulled into it? If so, that is also probably a bug we should fix.

Question 2: 
What kind of frameworks did you have in your app that this was detecting? We should be able to do something to filter those out. Were these cocoapods or carthage frameworks? Or were these some other internal frameworks that your project has? 

Asking these questions because even though I love thorough doc, I also would love a smarter `get_version_number` that users don't have to worry about this 😉 

Thanks again! ❤️

lordcodes:
Thank you! It was an internal framework that may not have been correctly
configured. Will have a look when I'm at my laptop and try and reproduce
what caused it.

Maybe I could reword slightly. It was meant to be a reminder to help people
when if / when it happens as it was quite confusing. However, you are right
may be able to get the action to avoid it happening. Will investigate and
then comment when I've reminded myself what caused it.

Thanks.

joshdholtz:
@lordcodes That would be great! I will await for the response 😊

lordcodes:
@joshdholtz It was a while since the issue occurred and it turned out the issue actually only occurred with increment_version_number. It seems get_version_number asks for the target, however, increment_version_number just selects one automatically.

I will update the docs for get_version_number in this PR to remove the mention of the issue.

joshdholtz:
@lordcodes I actually rewrote `get_version_number` a while back to migration from using `avgtool` to `xcodeproj` because I _think_ if I rewrote `increment_version_number` it would prevent that weird behavior selecting the wrong types of targets... I should probably get on this 😉

fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -113,19 +113,19 @@ def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :xcodeproj,
                              env_name: "FL_VERSION_NUMBER_PROJECT",
-                             description: "optional, you must specify the path to your main Xcode project if it is not in the project root directory",
+                             description: "Path to the main Xcode project to read version number from, optional. By default will use the first Xcode project found within the project root directory",
```
joshdholtz: I actually don't like that this action selects the first project 😱 But I like the documentation that tells about this behavior. I think I'm going to create a follow up issue about fixing that to prompt for a project to select 🤔

joshdholtz: Created https://github.com/fastlane/fastlane/issues/14123 as a follow up to this

joshdholtz (approved):
Thanks for looking into my questions and adjusting the documentation ❤️ This will get pushed out in tomorrow's release 💪

fastlane/lib/fastlane/actions/get_version_number.rb
lordcodes: Completely agree, thanks.

lordcodes:
> @lordcodes I actually rewrote `get_version_number` a while back to migration from using `avgtool` to `xcodeproj` because I _think_ if I rewrote `increment_version_number` it would prevent that weird behavior selecting the wrong types of targets... I should probably get on this 😉

@joshdholtz I actually created #14115 to ask why there was this inconsistency. Thanks for clearing that up.

Issue #14070, opened 2019-01-16
Rate limit has been exceeded for: itunes-connect|nonauth_url_spamControl|<my_ip_address>


scorpion35:
I am also getting this error - `Message: Rate limit has been exceeded for: itunes-connect|nonauth_url_spamControl|<my_ip_address>`

I am trying to get a list of all users and their permissions using Spaceship
```ruby
all_users = Spaceship::Tunes::Members.all
user = all_users[0] # for example
account_permissions = {}
account_permissions[:name] = teamname
account_permissions[:islive] = !user.not_accepted_invitation
account_permissions[:permissions] = user.roles
account_permissions[:apps] = user.has_all_apps ? ["all"] : user.selected_apps.map { |a| a.name if !a.nil? }
account_permissions[:lastupdated] = Time.now.strftime("%Y-%m-%d")
```

My code gets the details for first 15 or so users, and fails after with the above error message. Was able to find this information - https://stackoverflow.com/a/41377650/1946418; but not sure how true that is. The number seems too low?

Let me know if you need more information, please!

This is my Fastlane's version - `fastlane 2.113.0`

janpio:
Can you check the spaceship logfiles in `/tmp/spaceship-*.log` and post them here? This should show what requests are being generated and which one is giving you the exact error you are seeing.

Do the requests being generated by spaceship match the requests sent by the web UI when doing the same task manually?

scorpion35:
This is the last line - 
`DEBUG [14:14:21]: << GET ra/apps/manageyourapps/summary/v2: 429 Rate limit has been exceeded for: itunes-connect|nonauth_url_spamControl|<my_ip_address>`

Is this what you are looking for?

(Sorry, I cannot post the full file as it seems to contain sensitive information!)

janpio:
Can you filter it down to show all the requests (`...: >> ...`)? I am trying to understand what kind of request profile your code produces.

scorpion35:
@janpio Sorry for the delay! Please find the attached file.

[spaceship_requests_spamcontrol.log](https://github.com/fastlane/fastlane/files/2768732/spaceship_requests_spamcontrol.log)

This line is fetching apps each user has access to. Could it be this line is causing the issue? 
`account_permissions[:apps] = user.has_all_apps ? ["all"] : user.selected_apps.map { |a| a.name if !a.nil? }`

scorpion35:
Confirmed. If I remove the above line, I am not running into this error anymore. 

[...]

PR #14104, opened 2019-01-20
Users optimization

scorpion35:
By calling `Application.all` once, the GET request is only executed once per user, when `selected_apps` is called. Thanks @janpio for the tip!

Branch: "master" | Commit Type: "Bug fix" | Changes tested: "yes"

---

### Checklist

### Motivation and Context
**Context:**
- I am trying to get all users in a account and fetch the app names each user has access to. 

**Problem:**
- If a user has access to 10 apps (as an example), `10` GET requests were sent to Apple.
- And if there are 100 users/account, `100 * 10 = 1000` GET requests would be sent to Apple.

**If it fixes an open issue, please link to the issue here.**
- https://github.com/fastlane/fastlane/issues/14070
- This doesn't fix the issue completely, but optimizes **the number of GET requests to only one per user** (which is a huge win I think).

### Description
- Get all `Applications` before the loop. `all_apps = Application.all`
- Run `select` on the local `all_apps` variable.

**Please describe in detail how you tested your changes.**
- I have about over 600 users in all of our accounts combined. When I ran my script for all users, I have only seen ~600 GET requests.
- I have also noticed it's very fast now, even if the user has access to over 250 apps.

### DIFF
--- a/spaceship/lib/spaceship/tunes/member.rb
+++ b/spaceship/lib/spaceship/tunes/member.rb
@@ -58,7 +58,7 @@ module Spaceship
         parsed_apps = []
         all_apps = Application.all
         raw_data["userSoftwares"]["value"]["grantedSoftwareAdamIds"].each do |app_id|
-          parsed_apps << all_apps.select { |app| !app.apple_id.nil? && app.apple_id == app_id }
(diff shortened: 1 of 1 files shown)


scorpion35:
I have signed.

scorpion35:
I have signed.

Issue #14033, opened 2019-01-08
Unable to distribute to testflight builds


charles-won:
### Question Checklist

- [X] Updated fastlane to the latest version
- [X] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [X] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [X] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
Error when distributing testflight builds via pilot. 

#### Question Description
We started seeing these errors in our CI pipe line a couple weeks ago that says we are unable to update the change log. But seems like there is something else going on here. 
As we haven't made any changes to our fastlane scripts, anyone else having issues like this starting a couple weeks ago?

Here's are verbose log:

```
ERROR [2019-01-07 17:11:12.56]: fastlane finished with errors
Traceback (most recent call last):
	24: from /Users/chwon/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `<main>'
	23: from /Users/chwon/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `eval'
	22: from /Users/chwon/.rvm/gems/ruby-2.5.1/bin/fastlane:23:in `<main>'
	21: from /Users/chwon/.rvm/gems/ruby-2.5.1/bin/fastlane:23:in `load'
	20: from /Users/chwon/.rvm/gems/ruby-2.5.1@global/gems/fastlane-2.113.0/bin/fastlane:23:in `<top (required)>'
	19: from /Users/chwon/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/gems/fastlane-2.113.0/fastlane/lib/fastlane/cli_tools_distributor.rb:108:in `take_off'
	18: from /Users/chwon/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/gems/fastlane-2.113.0/fastlane/lib/fastlane/commands_generator.rb:41:in `start'
	17: from /Users/chwon/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/gems/fastlane-2.113.0/fastlane/lib/fastlane/commands_generator.rb:333:in `run'
	16: from /Users/chwon/.rvm/gems/ruby-2.5.1/gems/commander-fastlane-4.4.6/lib/commander/delegates.rb:15:in `run!'
[...]

PR #14137, opened 2019-01-25
[download_dsyms] wait for build to process

tmm1:
When specifying a version/build_number to download dsyms for, wait to make sure the build has finished processing and dsyms are available first.

This makes it much easier to deal with multiple builds, since pilot can be used to upload them back-to-back, and the slow part of waiting for them to process can be deferred allowing all builds to be processed in parallel.


janpio:
Hm, how does that change existing behavior? 
How many HTTP calls does it add via spaceship, could you please check via a before/after comparison of a spaceship log (`/tmp/spaceship*.log`) please?

tmm1:
The existing behavior is to fail when it attempts to download dSYMs for a build which has not yet been processed (i.e. dSYMs have not been generated yet).

The spaceship log shows requests to `providers/x/apps/y/platforms/z/trains/a/builds`

The number of requests depends on how long Apple takes to process the build.

janpio:
Yeah, I meant the current default use case of downloading a dsym for an app that _is_ processed. Does this add the call for the trains-builds there as well? Can we avoid that?

tmm1:
I can put the wait call behind an option if you prefer. Otherwise there's no way to know if the build is processed or not without making atleast one call to the train.

janpio:
No changes required, I am just trying to understand if the solution is appropriate.

What happens if one tries to download a dsym of an app that is still processing?

tmm1:
There is some sort of error. Likely saying version not found or something similar.

An alternative that would be more flexible would be to add a new action that just waits for a build to process. That way it would be up to the user to compose it as they like.. so they could call it before they call download_dsyms, or use it elsewhere in their pipeline. Right now the only way to wait for processing is as part of `pilot`, which makes it hard to upload multiple builds together to allow them to be processed in parallel.

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -44,6 +44,8 @@ def self.run(params)
             version = latest_candidate_build.train_version
             build_number = latest_candidate_build.build_version
           end
+        else
+          FastlaneCore::BuildWatcher.wait_for_build_processing_to_be_complete(app_id: app.apple_id, platform: platform, train_version: version, build_version: build_number, poll_interval: 30, strict_build_watch: true)
```
joshdholtz: Thoughts on adding a new option around this? Similar to the `pilot` one that is called `skip_waiting_for_build_processing`? Even though I think this is 💯 myself, I don't want to have to _force_ users to wait for processing if they don't want to wait for whatever reason 🙃 

I think I'm more concerned with breaking/changing any existing behavior/implementation around how users are currently expecting this work. I'm okay with making this new option default to `true` but just at least give the user an option to opt out 🤔

PoissonBallon:
Any news about it ? 
I often have `No dSYM files found on App Store Connect - this usually happens when no recompiling has happened yet` when I tried to download dsyms of my last upload :/

tmm1:
I've been having some issues using this locally when I have multiple apps with the same version/build number. Seems like a similar bug to #14033

fastlane/lib/fastlane/actions/download_dsyms.rb
tmm1: I think that makes sense. I ran into some cases where it would watch indefinitely and never find the build due to other bugs.

janpio:
(The Ubuntu CI failure is not your fault: https://github.com/fastlane/fastlane/issues/14481)

fastlane/lib/fastlane/actions/download_dsyms.rb
gregkerzhner: I guess if you didn't want to introduce breaking changes, the best thing to do might be to have it default to not waiting (old behavior), and introduce a new option that turns your feature on only if the user opts in.  I can see how other people might get confused when this feature hits about why their CI is hanging on some task all of a sudden.

joshdholtz:
@tmm1 If you would allow the maintainers access to make changes to your branch that would be 💯 I would be happy to fix this Ubuntu test error and make the changes needed to get this merged 🙃

https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

tmm1:
Looks enabled by default already. Would be great if you can take this over!

PatrickDotStar:
We have the same issues for our project. We recently enabled bitcode and now when refreshing the dsym after uploading the app to appstore connect it may be the case that the dsym are not available atm even when the build finished processing. We pass the `wait_for_uploaded_build: true` parameter for the `testflight` command. 
I noticed, that the donwload dsym button in appstore connect is not showing immediately after the build finished processing. It can take some time until they are available for download and if this is not the case the `download_dsyms` script will fail saying that no dSYM URL is available and we then have to run the script again at a later moment

missingDsyms

tmm1:
> I noticed, that the download dsym button in appstore connect is not showing immediately after the build finished processing. It can take some time until they are available for download and if this is not the case the `download_dsyms` script will fail saying that no dSYM URL is available and we then have to run the script again at a later moment

I'm noticing this now too. The whole point of this PR was to wait for processing before downloading, but if that is now not enough I'm not sure what to do. We would need to keep polling for the dSYM url I suppose?

bguidolim:
As a workaround, I simply put a `sleep(60)` before calling my refresh dsyms lane.

PatrickDotStar:
> As a workaround, I simply put a `sleep(60)` before calling my refresh dsyms lane.

Haha we also do the same but we sleep for 3 minutes just to be sure😅
60 sec was not enough for us.

tmm1:
> Haha we also do the same but we sleep for 3 minutes just to be sure😅
> 60 sec was not enough for us.

Yikes.

@joshdholtz It seems like the best approach here is to add a loop around the dSYMs URL fetch instead of giving up after the first failure.

tmm1:
@joshdholtz do you have any time this week to take a look here?

janpio (approved):
Code itself looks good, not sure of the internal logic though - I would defer that to @joshdholtz to evaluate before merging.

notahat:
I've been experimenting around the same problem, and `BuildWatcher.wait_for_build_processing_to_be_complete` is not enough to guarantee that the dSYMs will be available for download. My experience has been that they're usually available within 30 seconds of that.

I'm considering a PR that adds an option to make `download_dsyms` poll until they're ready (with an eventual timeout.)

tmm1:
> I'm considering a PR that adds an option to make `download_dsyms` poll until they're ready (with an eventual timeout.)

Please do this!

janpio:
(Thanks for your continued work on this issue - it's a really hard nut to crack and I hope we can find a good solution with all your help.)

Issue #13270, opened 2018-09-03
setup_travis fails when run twice


JosephDuffy:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I have `setup_travis` at the top of my Fastfile, but use Fastlane to install dependencies, download certs via `match`, etc. as part of a `before_install` step in my `.travis.yml`. Fastlane is then used to run the tests in the `script` section.

This means that `bundle exec fastlane NAME_OF_LANE` is ran multiple times, with different lanes. This causes `setup_travis` to be ran multiple times, which appears to be failing because it is unable to create another temporary keychain with the same name.

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>bundle exec fastlane bootstrap

+-------------------------------------------------+---------+---------------------------------+

|                                        [32mUsed plugins[0m                                         |

+-------------------------------------------------+---------+---------------------------------+

| Plugin                                          | Version | Action                          |

+-------------------------------------------------+---------+---------------------------------+

| fastlane-plugin-latest_hockeyapp_version_number | 1.0.0   | latest_hockeyapp_version_number |

| fastlane-plugin-badge                           | 1.1.0   | add_badge                       |

+-------------------------------------------------+---------+---------------------------------+

[14:52:16]: Sending anonymous analytics information

[14:52:16]: Learn more at https://docs.fastlane.tools/#metrics

[14:52:16]: No personal or sensitive data is sent.

[14:52:16]: You can disable this by adding `opt_out_usage` at the top of your Fastfile

[14:52:16]: [32m------------------------------[0m

[14:52:16]: [32m--- Step: default_platform ---[0m

[14:52:16]: [32m------------------------------[0m

[14:52:16]: [32m--------------------------[0m

[14:52:16]: [32m--- Step: setup_travis ---[0m

[14:52:16]: [32m--------------------------[0m

[14:52:16]: Creating temporary keychain: "fastlane_tmp_keychain".

[14:52:16]: Enabling readonly mode for Travis

[14:52:16]: [32mDriving the lane 'ios bootstrap' 🚀[0m

[14:52:16]: [32m------------------------------------------------------[0m

[...]

PR #14139, opened 2019-01-25
[action] CreateKeychainAction patch

ojizero:
### Checklist

### Motivation and Context

When running `setup_travis` or `setup_circle_ci` to perform setup actions on their respective CI environments creates a Keychain DB using the `create_keychain` action. And the `create_keychain` can only be run once per system (per CI job).

This causes error in the situation where you run multiple lanes separately in the same job, the error is due to the `create_keychain` action failing to be rerun since it already created the Keychain DB.

This is unideal as the correct placement for such action is in the `before_all` hook, and placing it inside specific lanes create unneeded coupling and dependence between different lanes.

More details are in the discussion in #13270.

### Description

The change introduced allows `create_keychain` to be called multiple times without failing, by checking if the requested DB was created before and skipping the `security create-keychain` command needed to create it.

It also provides an option `require_create` which is a flag (defaulting to false) and if it is set to true it would cause the action to fail if the Keychain DB already exists, which could be very helpful in certain situations when you want to create a clean Keychain DB.

The implementation was done in the `create_keychain` action itself to avoid have to duplicate the logic whenever a new consumer wants this behaviour.

### DIFF
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -26,7 +26,7 @@ module Fastlane
         if !exists?(keychain_path)
           commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{keychain_path}", log: false)
         elsif params[:require_create]
-          UI.user_error!("`require_create` option passed, but found keychain '#{keychain_path}', failing create_keychain action")
+          UI.abort_with_message!("`require_create` option passed, but found keychain '#{keychain_path}', failing create_keychain action")
         else
           UI.important("Found keychain '#{keychain_path}', creation skipped")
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -52,6 +60,14 @@ def self.run(params)
         commands
       end
 
+      def self.exists?(keychain_path)
+        keychain_path = File.expand_path(keychain_path)
+
+        # Creating Keycahins using the security
```
janpio:
```suggestion
        # Creating Keychains using the security
```

joshdholtz (approved):
Thanks for the contribution! I love this change 😊 I actually ran into this same issue on CircleCI on Friday and was going to make a change for it but you beat me to the punch 😉

Issue #14088, opened 2019-01-17
[frameit] Split advanced functionality from basic framing


janpio:
`frameit` or `frame_screenshots` is one action, that does 2 different things: 

1. Framing a simple screenshot in a device frame
2. Positioning a framed screenshot on a background and places some (potentially localized) text besides it

1 is documented as "Basic Usage", and 2 as "Advanced Usage (optional)". In the code they are both separate, but still intermingled - the most irritating thing is that because of reasons [step 2 needs to produce an image of the same dimensions as the original screenshot, see footnote 1] the code is not actually written in the same step 1, then step 2 way but different.

As there tend to pop up small bugs from time to time, it would be beneficial to split the basic device framing from the other more advanced stuff.

Current suggested (code) naming convention: frame (steps 1) + wrap (step 2)

Footnote 1:
This results in the following sizes when using `frameit`:
```
Basic mode:
Screenshot + Frame = Framed Screenshot
100        +   5   = 105

Advanced mode:
Screenshot + Frame = Framed screenshot + wrapping = Wrapped and framed screenshot
100        +   5   = 105 =resized=> 100 +   10    = 110 =resized=> 100     
```

---

TODO along the way:

- [ ] Read `Framefile.json` only once
- [ ] Better output for advanced mode
- [ ] Fix trimbox output ` Trim box for key "title" is adjusted to align top: #<Frameit::Trimbox:0x000000000781c480>`
-

PR #14142, opened 2019-01-25
[frameit] debug mode

janpio:
### Checklist

### Motivation and Context
Working with frameit can be confusing and frustrating, as the code is a bit of a black box. I started to work on fixing this in #14088, but quickly learned that there is missing something: a way to visually debug what frameit is doing.

### Description
This PR adds a `debug_mode` parameter to the action that enables some additional functionality when frameit is used. In this first iteration it adds the filename of the frame that is being used to frame a screenshot to the resulting picture. This makes it possible to quickly understand what frame was used.

Here are some examples of what it looks like used with a screenshot of a full pink app that shows the device or simulator name:

ipad 5th generation -01startscreen_framed
ipad pro 12 9-inch -01startscreen_framed
iphone 6s plus-01startscreen_framed
iphone xs max-01startscreen_framed
iphone se-01startscreen_framed
iphone 5s-01startscreen_framed

As you see, there are some mismatches - which means there is some further work on frameit needed.

### DIFF
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -12,20 +12,23 @@ module Frameit
   # Currently the class is 2 lines too long. Reevaluate refactoring when it's length changes significantly
   class Editor # rubocop:disable Metrics/ClassLength
     attr_accessor :screenshot # reference to the screenshot object to fetch the path, title, etc.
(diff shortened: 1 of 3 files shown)


joshdholtz (approved):
I dig it 😍

PR #14156, opened 2019-01-29
add ability to update a specific target's development team

addbrick:
### Checklist

### Motivation and Context

We build multiple apps from a single project. They all have their own target, but share extension targets. The `DEVELOPMENT_TEAM` need to be updated on the extension targets when building each app target. We don't want to change it across the project, just on the specific targets. Much like `update_app_identity`.

### Description

Sometimes you want to update the `DEVELOPMENT_TEAM` of just one target, instead of the entire project. This adds `target` as an option to `update_project_team`. When supplied the `DEVELOPMENT TEAM` of that one target will be updated instead of the entire project.

I tested this by integrating it into my company's build process. Running it multiple times while testing builds.

### DIFF
--- a/fastlane/lib/fastlane/actions/update_project_team.rb
+++ b/fastlane/lib/fastlane/actions/update_project_team.rb
@@ -6,7 +6,7 @@ module Fastlane
     class UpdateProjectTeamAction < Action
       def self.run(params)
         project_path = params[:path]
-        target_name = params[:target]
+        selected_targets = params[:targets]
 
         UI.user_error!("Could not find path to xcodeproj '#{project_path}'. Pass the path to your project (not workspace)!") unless File.exist?(project_path)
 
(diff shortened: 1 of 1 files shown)


joshdholtz (changes requested):
Can we actually make a change in this action to **only** use `xcodeproj`? 😊 That would be 💯  so we only have to support one method of doing things

addbrick:
I thought about that. I like the convenience of using a regex to replace it everywhere. Doing the samething in `xcodeproj` would be more complicated.

fastlane/lib/fastlane/actions/update_project_team.rb
```diff
@@ -5,18 +5,20 @@ module SharedValues
...
-          # Load .xcodeproj
-          project = Xcodeproj::Project.open(project_path)
+        # Load .xcodeproj
+        project = Xcodeproj::Project.open(project_path)
 
-          # Fetch target
-          target = project.native_targets.find { |native_target| native_target.name == target_name }
+        # Fetch target
+        targets = project.native_targets
+        if params[:target]
```
addbrick: Since we are defaulting to all targets, maybe we should just turn this into an array `targets`?

addbrick:
@joshdholtz that change looks great. I was making it more complicated in my head, thinking about workspaces rather than projects.

I added one comment suggesting an update

fastlane/lib/fastlane/actions/update_project_team.rb
joshdholtz: Good call good call! I like that idea... I will make that change quick 😉

joshdholtz:
@addbrick Here you go! Replaced `target` with `targets` 💪 Mind taking a look again?!

addbrick:
@joshdholtz that looks great! Thanks for the assist!

joshdholtz (approved):

joshdholtz:
@addbrick Thanks for doing all the hard work here! ❤️

Issue #7000, opened 2016-11-15
Ability to return information about a Membership and Contracts


barnaclebarnes:
### Issue Description

We manage over 150 different client apps on our platform and we often find that client let their Apple Developer memberships lapse and fail to agree to new developer agreements. We would like to automate a check on the developer and iTunes Connect portal to:

* Return Membership information (i,.e. Program Type, Entity Type, Phone, Address, Expiration Date, Device Reset Date, Team Agent)
* Return an indicator if there are any agreements that need to be signed from the Developer Portal
* Return an indicator if there are any agreements that need to be signed from iTunesConnect

If we could do this then we could proactively reach out to clients so their apps do not get removed from the app store.

TKBurner:
@barnaclebarnes Are you interested in submitting a PR for this? I'll mark it as a feature request for now 👍

PR #14159, opened 2019-01-29
[spaceship] add a spaceship method to check Program License Agreement warnings

### LINKED ISSUE #13619: Updated License Agreement
### Question Checklist

### Question Subject
 Apple Developer Program License Agreement

#### Question Description
Is there a way to tell if an account that I'm a team member on, needs to have the agent logged in to accept a new Apple license agreement? Is anything like this even exposed by the Apple api or can this only be done by screen scraping? 

Basically, I need to be able to tell if this message is displayed when I login to the developer portal or app store connect. "The Apple Developer Program License Agreement has been updated."

keehun:
### Checklist

### Motivation and Context

Fixes #13619 and #7000

I work with a client who develops, manages, and releases countless applications on behalf of their own clients. We needed a quick way to keep tabs on which of our clients' Apple Developer Accounts need updated Program License Agreement signatures.

### Description

Add a new method named `fetch_program_license_agreement_messages` in `Spaceship::Portal.client` to be able to access any messages on Olympus for our client accounts in our `Fastfile`.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -600,7 +600,7 @@ module Spaceship
       all_messages = []
 
       messages_request = request(:get, "https://olympus.itunes.apple.com/v1/contractMessages")
-      body = messagesRequest.body
+      body = messages_request.body
       if body
         body = JSON.parse(body) if body.kind_of?(String)
         body.map do |messages|
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -596,11 +596,10 @@ module Spaceship
     end
 
     # Get contract messages from App Store Connect's "olympus" endpoint
-    def fetch_program_license_agreement_messages()
-
+    def fetch_program_license_agreement_messages
       all_messages = []
 
-      messagesRequest = request(:get, "https://olympus.itunes.apple.com/v1/contractMessages")
+      messages_request = request(:get, "https://olympus.itunes.apple.com/v1/contractMessages")
       body = messagesRequest.body
       if body
         body = JSON.parse(body) if body.kind_of?(String)
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -595,6 +595,23 @@ module Spaceship
       ENV["FASTLANE_SESSION"] || ENV["SPACESHIP_SESSION"]
     end
 
+    # Get contract messages from App Store Connect's "olympus" endpoint
+    def fetch_program_license_agreement_messages()
+
+      all_messages = []
+
+      messagesRequest = request(:get, "https://olympus.itunes.apple.com/v1/contractMessages")
+      body = messagesRequest.body
+      if body
+        body = JSON.parse(body) if body.kind_of?(String)
+        body.map do |messages|
+          all_messages.push(messages["message"])
+        end
+      end
+
+      return all_messages
+    end
+
     #####################################################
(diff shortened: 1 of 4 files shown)


joshdholtz (changes requested):
This looks like a solid addition! Would you also be able to add a fixture with an example body of what this returns along a test? A similar example can be seen in this PR - https://github.com/fastlane/fastlane/pull/12967/files?utf8=%E2%9C%93&diff=unified

If you want me to add the tests I totally can but I would just need a sample response 😊

spaceship/lib/spaceship/client.rb
```diff
@@ -595,6 +595,22 @@ def self.spaceship_session_env
       ENV["FASTLANE_SESSION"] || ENV["SPACESHIP_SESSION"]
     end
 
+    # Get contract messages from App Store Connect's "olympus" endpoint
+    def fetch_program_license_agreement_messages
+      all_messages = []
+
+      messages_request = request(:get, "https://olympus.itunes.apple.com/v1/contractMessages")
+      body = messages_request.body
+      if body
```
KrauseFx: Small thing, instead of having an `if` block here, you can do

```
return all_messages unless body

...
```

KrauseFx (commented):
Love it

spaceship/lib/spaceship/client.rb
keehun: @KrauseFx, thanks for the suggestion! I am personally against the `unless` operator in Ruby, but as the maintainer of this project, if you feel you'd prefer `unless`, I will change it. Just let me know!

keehun:
@joshdholtz, I added a test for it! Took a while to understand how to structure the mock response/write test expectations, but once I understood it, it was pretty clear! I've ran `bundle exec rspec spaceship/spec/portal/portal_client_spec.rb` which returned `45 examples, 0 failures`. 🎉

keehun:
While I'm here with the `fastlane` experts, I wanted to ask: what is _the best way_ to access this method in my `Fastfile`?

The best I could figure out was (I'm mostly concerned about the first 2 lines):

```ruby
spaceship = Spaceship::Portal.login("EMAIL", "PASSWORD")
all_pla_warnings = spaceship.fetch_program_license_agreement_messages
if all_pla_warnings.count > 0
  UI.error all_pla_warnings
  UI.user_error! "Please log into the App Store Connect account and sign any necessary Program License Agreement updates!"
end
```

I imagine that there's _already_ an instance of `Spaceship::Portal.client`... I just can't seem to understand how to use it.

joshdholtz:
@keehun There _may_ be a shared client out there that is already logged in (depending on what other tools you have used) - `Spaceship::Portal.client`. This is what I usually use anyway. My normal setups would look something like this...

```rb
Spaceship::Portal.login("EMAIL", "PASSWORD")
Spaceship::Portal.client.select_team(team_name: "TEAM_NAME")
Spaceship::Portal.client.fetch_program_license_agreement_messages
```

But doing it the way you do it also works too 😊

spaceship/lib/spaceship/client.rb
joshdholtz: You could also do...
```rb
return all_messages if body.nil?
```

Either works but I think what @KrauseFx was going for was an early return if no body which makes the code a bit flatter and easier to read 🙃 

But what you have here matches with the other implementations already in this file so what you got is 👌

joshdholtz (approved):
Thanks for adding this API and those tests! Nothing makes me happier than more tests 😊 ❤️

PR #14164, opened 2019-01-30
[frameit] Add "Screenshot orientation" section to `frameit` documentation

### LINKED ISSUE #14076: [frameit] Document option `force_orientation_block`
#11161 added a `force_orientation_block` parameter to enable better customization of the screenshot orientation (and to fix #10928). Unfortunately this is pretty much undocumented right now.

Some default behaviour of frameit is defined via the default value of that parameter:
https://github.com/fastlane/fastlane/blob/c4def9f78622851b7c51bdfbc55080659b38b8fe/frameit/lib/frameit/options.rb#L60-L67

The PR has 2 examples of blocks that can be used:
https://github.com/fastlane/fastlane/pull/11161#issuecomment-353684428
https://github.com/fastlane/fastlane/pull/11161#issuecomment-353728307

The file to be edited: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/frame_screenshots.md

A good headline could maybe be "Screenshot orientation" or similar, then an explanation of the default behavior and an introduction of the param, how it works and a few simple examples.

mollyIV:
### Checklist

### Motivation and Context

closes https://github.com/fastlane/fastlane/issues/14076 
The usage of `force_orientation_block` is not very well documented. 

### Description

We are introducing a brand new "Screenshot orientation" section. At the beginning, we are explaining the default behaviour of taking the screenshots. Later we are introducing `force_orientation_block ` parameter and wrapping it up with two simple examples from https://github.com/fastlane/fastlane/pull/11161. 

Thanks for reviewing it! 💪🏅

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
@@ -223,7 +223,7 @@ One way to override the default behavior is editing the file name by adding `for
 
 ### `force_orientation_block`
 
-If the default behavior doesn't fit your needs and you don't want or can't rename your screenshots, you can customize _frameit_'s orientation behavior by setting a `force_orientation_block` parameter. The valid values are: `landscape_left` (home button on the left side), `:landscape_right` (home button on the right side), `:portrait` (home button on the bottom), `nil` (home button on the right side).
+If the default behavior doesn't fit your needs and you don't want or can't rename your screenshots, you can customize _frameit_'s orientation behavior by setting a `force_orientation_block` parameter. The valid values are: `:landscape_left` (home button on the left side), `:landscape_right` (home button on the right side), `:portrait` (home button on the bottom), `nil` (home button on the right side).
 
 ### Examples
 
--- a/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
@@ -235,7 +235,7 @@ frameit(
       when "iPad Pro (12.9-inch)-01LoginScreen" 
         :landscape_right
       when "iPhone 6 Plus-01LoginScreen"
-        : portrait
+        :portrait
       # and so on
     end
   end
--- a/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
 - These `.strings` files **MUST** be utf-8 (UTF-8) or utf-16 encoded (UTF-16 BE with BOM). They also must begin with an empty line. If you are having trouble see [issue #1740](https://github.com/fastlane/fastlane/issues/1740)
 - You **MUST** provide a background if you want titles. _frameit_ will not add the tiles if a background is not specified.
 
+### Screenshot orientation
+
+By default `frameit` adds a frame to your screenshot based on an orientation you took it. For a portrait (vertical orientation) it is going to add portrait frame and for a landscape (horizontal orientation) - landscape left.
+
+If you'd like to get a frame in a landscape right for your horizontal screenshots, you need to add `force_landscaperight` at the end of your screenshot's filename. 
+
+### `force_orientation_block`
+
+In a case you don't want to rename your screenshots or default behaviour does not meet your requirements, there's a way to customise it. In order to do so, you need to set `force_orientation_block` parameter.
+
+### Examples
+
+```ruby
+frameit(
+  path: "./fastlane/screenshots",
+  force_orientation_block: proc do |filename|
+    case filename
+      when "1Login" 
+        :landscape_left
+      when "2Profile"
+        :landscape_right
+      # and so on
+    end
+  end
+)
+```
+
+```ruby
+frameit(
+  silver: true,
+  path: "./fastlane/screenshots",
+  force_orientation_block: proc do |filename|
+    f = filename.downcase
+    if f.include?("landscape")
+      :landscape_right
+    end
+  end
+)
+```
+
 # Mac
 
 With _frameit_ it's possible to also frame macOS Application screenshots. You have to provide the following:
--- a/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
@@ -217,13 +217,13 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
 
 ### Screenshot orientation
 
-By default _frameit_ adds a frame to your screenshot based on an orientation you took it. For a portrait (vertical orientation) it is going to add portrait frame and for a landscape (horizontal orientation) - landscape left.
+By default _frameit_ adds a frame to your screenshot based on an orientation you took it. For a portrait (vertical orientation) it is going to add portrait frame and for a landscape (horizontal orientation) - landscape left (= [Home button on the left side](https://developer.apple.com/documentation/uikit/uiinterfaceorientation/landscapeleft)).
 
-If you'd like to get a frame in a landscape right for your horizontal screenshots, you need to add `force_landscaperight` at the end of your screenshot's filename. 
+One way to override the default behavior is editing the file name by adding `force_landscaperight` to the end.
 
 ### `force_orientation_block`
 
-In a case you don't want to rename your screenshots or default behaviour does not meet your requirements, there's a way to customise it. In order to do so, you need to set `force_orientation_block` parameter.
+If the default behavior doesn't fit your needs and you don't want or can't rename your screenshots, you can customize _frameit_'s orientation behavior by setting a `force_orientation_block` parameter. The valid values are: `landscape_left` (home button on the left side), `:landscape_right` (home button on the right side), `:portrait` (home button on the bottom), `nil` (home button on the right side).
 
 ### Examples
 
@@ -232,10 +232,10 @@ frameit(
   path: "./fastlane/screenshots",
   force_orientation_block: proc do |filename|
     case filename
-      when "1Login" 
-        :landscape_left
-      when "2Profile"
+      when "iPad Pro (12.9-inch)-01LoginScreen" 
         :landscape_right
+      when "iPhone 6 Plus-01LoginScreen"
+        : portrait
       # and so on
     end
   end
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
...
+In a case you don't want to rename your screenshots or default behaviour does not meet your requirements, there's a way to customise it. In order to do so, you need to set `force_orientation_block` parameter.
+
+### Examples
+
+```ruby
+frameit(
+  path: "./fastlane/screenshots",
+  force_orientation_block: proc do |filename|
+    case filename
+      when "1Login"
```
joshdholtz: This may be nitpicky but we may want to rework this example as I don't think it will actually work if copy pasted 😇 `filename` returns something like "iPad Pro (12.9-inch)-01LoginScreen" so this example may be confusing if copy pasted 🤔

joshdholtz (changes requested):
I actually didn't know _frameit_ had these capabilities 😇 A few small nitpicks otherwise 💯

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
mollyIV: Actually it depends on the way the users generate the screenshots. The way I was testing the parameter was just running a sample application, hit `cmd + s` and move the screenshot to the proper catalog. Then I renamed it as I wanted. However I can imagine that most common scenario to generate screenshots is `snapshot` tool. In that case, indeed would be great provide better context to our example 🙂 

```ruby
frameit(
  path: "./fastlane/screenshots",
  force_orientation_block: proc do |filename|
    case filename
      when "iPad Pro (12.9-inch)-01LoginScreen" 
        :landscape_right
      when "iPhone 6 Plus-01LoginScreen"
        : portrait
      # and so on
    end
  end
)
```
WDYT @joshdholtz ? 🤔

janpio:
FYI: I added `closes ` before the issue link, so that the issue will automatically get closed when we merge this PR.

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
 - These `.strings` files **MUST** be utf-8 (UTF-8) or utf-16 encoded (UTF-16 BE with BOM). They also must begin with an empty line. If you are having trouble see [issue #1740](https://github.com/fastlane/fastlane/issues/1740)
 - You **MUST** provide a background if you want titles. _frameit_ will not add the tiles if a background is not specified.
 
+### Screenshot orientation
+
+By default _frameit_ adds a frame to your screenshot based on an orientation you took it. For a portrait (vertical orientation) it is going to add portrait frame and for a landscape (horizontal orientation) - landscape left.
```
janpio: What does "landscape left" actually mean here?

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
...
+
+If you'd like to get a frame in a landscape right for your horizontal screenshots, you need to add `force_landscaperight` at the end of your screenshot's filename.
```
janpio: Maybe start this paragraph with "One way to override the default behavior ..." and then describe that one can edit the file name in this specific way to get that specific result.

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
...
 
+### Screenshot orientation
+
+By default _frameit_ adds a frame to your screenshot based on an orientation you took it. For a portrait (vertical orientation) it is going to add portrait frame and for a landscape (horizontal orientation) - landscape left.
+
+If you'd like to get a frame in a landscape right for your horizontal screenshots, you need to add `force_landscaperight` at the end of your screenshot's filename. 
+
+### `force_orientation_block`
+
+In a case you don't want to rename your screenshots or default behaviour does not meet your requirements, there's a way to customise it. In order to do so, you need to set `force_orientation_block` parameter.
```
janpio:
```suggestion
If the default behavior doesn't fit your needs and you don't want or can't rename your screenshots, you can customize _frameit_'s orientation behavior by setting a `force_orientation_block` parameter.
```

janpio: Maybe add some additional information how this block works (code executed when the action is run) and what possible return values there are, and what the effect of each are.

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
janpio: Better. Maybe we can also add an example that matches only part of the filename to decide?

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
...
+        :landscape_left
+      when "2Profile"
+        :landscape_right
+      # and so on
+    end
+  end
+)
+```
+
+```ruby
```
janpio: Some comment on what this does in plain English maybe?

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
...
+
+If you'd like to get a frame in a landscape right for your horizontal screenshots, you need to add `force_landscaperight` at the end of your screenshot's filename. 
+
+### `force_orientation_block`
+
+In a case you don't want to rename your screenshots or default behaviour does not meet your requirements, there's a way to customise it. In order to do so, you need to set `force_orientation_block` parameter.
+
+### Examples
+
+```ruby
```
janpio: Some comment on what this does in plain English maybe?

janpio:
Awesome start @mollyIV - I also added a few smaller comment to maybe improve the text. This param is _really_ complicated with all its dynamic power it brings. No wonder nobody bothered to document it yet 🐒

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
mollyIV: According to Apple's [documentation](https://developer.apple.com/documentation/uikit/uiinterfaceorientation/landscapeleft):

> The device is in landscape mode, with the device held upright and the Home button on the left side.

So maybe it's worth to add this sentence? 

```
... for a landscape (horizontal orientation) - landscape left. The device is in landscape mode, with the device held upright and the Home button on the left side.
```

Not sure if it's okay to use the same sentence from Apple's documentations without including the source, so maybe better you can insert a link to the docs: `... - [landscape left](https://developer.apple.com/documentation/uikit/uiinterfaceorientation/landscapeleft)` ? 🤔

janpio: Simple:
```suggestion
By default _frameit_ adds a frame to your screenshot based on an orientation you took it. For a portrait (vertical orientation) it is going to add portrait frame and for a landscape (horizontal orientation) - landscape left (= [Home button on the left side](https://developer.apple.com/documentation/uikit/uiinterfaceorientation/landscapeleft)).
```

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
janpio: You can accept this change here in the GitHub UI and it will automatically be added. (Caution: You have to fetch if you have a local checkout before committing again, otherwise you might get a merge conflict)

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
janpio: Oh right - ignore that and just keep these two examples.

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
mollyIV: Hmm, not sure if an additional comment will add more value. 

```
# It frames the screenshots in landscape right whenever the filename contains `landscape` word
...
```

Code is pretty simple anyway so I'd not add an extra noise 🤷‍♂️ WDYT?

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
```diff
@@ -215,6 +215,46 @@ The `keyword.strings` and `title.strings` are standard `.strings` file you alrea
...
+frameit(
+  path: "./fastlane/screenshots",
+  force_orientation_block: proc do |filename|
+    case filename
+      when "iPad Pro (12.9-inch)-01LoginScreen" 
+        :landscape_right
+      when "iPhone 6 Plus-01LoginScreen"
+        : portrait
```
janpio:
```suggestion
        :portrait
```

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
janpio: Yep, that sounds good. Keep in mind that typical fastlane and frameit users usually do not know ruby! They just use the "syntax" of the `Fastfile` they see somewhere which makes intuitive sense to all developers. The blocks... not so much. (e.g. missing `return` etc)

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
janpio: You switched the both comments, but at the correct example those look good :)

fastlane/lib/fastlane/actions/docs/frame_screenshots.md
mollyIV: That makes sense 🙇 Actually I was pretty sure most of the fastlane users know ruby 😅

janpio: **I** don't really know ruby 🙊

joshdholtz: ^ Lies lol

joshdholtz (approved):
Thanks for adding this doc! I pushed one commit to add a colon to one on the orientation symobls that was missing one but other than that this was 💯 Thanks for the contribution! ❤️

PR #14168, opened 2019-01-31
[match] check certificate validity date

claireduf:
### Checklist

### Motivation and Context
I propose to check validity of certificate stored in match. If you encrypted manually a certificate (which was valid at the time you encrypted it) and stored it in a match repo, you may encounter this: you decrypt the certificate after its end date, and no error is risen
The next step (gym for example) will fail with absolutely no explanation

This method just check that start date of certificate is in the past and end date is in the future.

### Description
* I get the certificate
* I get start and end datetime
* I compare them to present time
* if start < present < end, i display a message "its valid"
* otherwise, display an error


janpio:
Looks useful!

Do you have insight why this wasn't needed before, especially with certificates created by match? I am confused why we didn't have this yet...

match/lib/match/utils.rb
```diff
@@ -58,6 +58,16 @@ def self.get_cert_info(cer_certificate_path)
       return {}
     end
 
+    def self.check_cert_validity(cer_certificate_path)
+      cert = OpenSSL::X509::Certificate.new(File.binread(cer_certificate_path))
+      now = Time.now.utc
+      if (cert.not_before <=> now) * (now <=> cert.not_after) < 1
```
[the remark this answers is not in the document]
claireduf: yes, no problem :)

line 62: i get the certificate from the provided path (i use OpenSSL lib which do this perfectly)
line 63: i get the present time, in UTC format because start and end time of certificate are in UTC format (Apple's choice)
line 64: the tricky part :D 
_date1 <=> date2  = -1_ if date1 < date2
_date1 <=> date2  = 0_ if date1 = date2
_date1 <=> date2  = 1_ if date1 > date2

so if if (cert.not_before <=> now) * (now <=> cert.not_after) < 1
it means it is equal to 0 or -1

if 0, it means start OR end date is now, it means actually end date is now because start date is the date of creation of certificate. so certificate is invalid

if -1, it means
Option A: start <=> now = -1 AND now <=> end = 1 which means start < now AND end < now which means certificate is out of date
Option B: start <=> now = 1 AND now <=> end = -1 which means now < start AND now < end it is impossible because start < end

claireduf:
I think it wasn't needed before because if user uses spaceship we check if certificate is present on Apple console
if certificate is outdated, it is not present and fastlane raises an error

this case occurs when you store on match a certificate but do not have access to Apple dev console

match/lib/match/utils.rb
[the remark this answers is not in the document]
janpio: Ohh, one of those. I can never remember these. Thanks!

match/lib/match/utils.rb
```diff
@@ -58,6 +58,16 @@ def self.get_cert_info(cer_certificate_path)
...
+        UI.error("Your certificate '#{File.basename(cer_certificate_path)}' is not valid, please check start and end datetime")
+      else
+        UI.message("Your certificate '#{File.basename(cer_certificate_path)}' is valid")
```
joshdholtz:
```suggestion
        UI.verbose("Your certificate '#{File.basename(cer_certificate_path)}' is valid")
```

I don't think its 100% necessary to put this output - even though it is nice to know 😊 I think it can be implied if it doesn't error that it is valid so I think `verbose` fits best here 👌

match/lib/match/utils.rb
```diff
@@ -58,6 +58,16 @@ def self.get_cert_info(cer_certificate_path)
       return {}
     end
 
+    def self.check_cert_validity(cer_certificate_path)
```
joshdholtz: I'm kind thinking that this method shouldn't have any `UI.error` or `UI.verbose` (or `UI.message` as it currently is implemented in it). Since this is a utils class, this method should be one that can be used in other parts of the code without influencing any UI. 

I think I would suggest renaming this to `is_cert_valid?` (as that is what this is checking **and then** doing then either do a `UI.user_error!` (since I think the goal of this PR was to exit out early) or a `UI.verbose` if cert is good. Thoughts on that?

joshdholtz (changes requested):
I think this is a great additional check that could speed up some failures in match 👌 I left a few comments for you thought about how I think this could get reworked to better fit existing code styles 😊  Appreciate the time spent on this so far ❤️

match/lib/match/utils.rb
claireduf: very useful comment, totally agree on this
i made some change to take your advice into account
+ i only check end date of validity because start date is - for apple certificate- necessary its creation date, so it is always in the past

claireduf: Using user_error makes the test "Match Match::Runner uses existing certificates and profiles if they exist" fail, since this certificate seems outdated. 
Is there a way to renew it?

match/lib/match/utils.rb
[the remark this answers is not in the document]
KrauseFx: Jumping in here as another Ruby noob, can we update the code to use some less cool Ruby here (and therefore probably use more lines of code), so that it's easier for contributors in the future to read and update the code? :)

match/lib/match/utils.rb
```diff
@@ -58,6 +58,12 @@ def self.get_cert_info(cer_certificate_path)
       return {}
     end
 
+    def self.is_cert_valid?(cer_certificate_path)
+      cert = OpenSSL::X509::Certificate.new(File.binread(cer_certificate_path))
+      now = Time.now.utc
+      (now <=> cert.not_after) == -1
```
KrauseFx: Same here, also would like to see an explicit `return` assuming we're using the return value

claireduf: I added an explicit return for readability
i have simplified the test to test only end date, is it better like this? (more readable, respectful of the coding style ?)

joshdholtz (changes requested):
@claireduf Your changes look good! Thanks for making them 😊 It looks like there is a merge conflict though 😇 Do you want to rebase this or would you like me to rebase this? I'm cool with either!

joshdholtz (approved):
This looks 💯 now! This will get released here within the next few days! Thanks for the contribution ❤️

PR #14184, opened 2019-02-05
[frameit] Automate measuring of "slot" offsets and width

### LINKED ISSUE #14094: [frameit] Automate measuring of "slot" offsets and width

mollyIV:
### Checklist

### Motivation and Context
closes https://github.com/fastlane/fastlane/issues/14094

> Currently we have to manually measure the offsets and width of the "slot" where the screenshots will be placed. We should automate that as part of the frame update process.

### Description
In order to detect a "slot" on a device image, we are looking for transparent areas.  The detection is based on [this]( https://stackoverflow.com/questions/26244191/imagemagick-detect-coordinates-of-transparent-areas) solution. We are analysing the output command and parse the dimensions of the slot. Below you can find an example of raw command output for iPhone 5c image:

```bash
$ convert Apple\ iPhone\ 5c\ Blue.png -alpha extract -negate           \
   -define connected-components:verbose=true       \
   -define connected-components:area-threshold=100 \
   -connected-components 8 -auto-level  result.png
Objects (id: bounding-box centroid area mean-color):
  494: 640x1136+67+239 386.5,806.5 727040 srgb(255,255,255)
  4: 767x1605+0+0 385.4,801.2 478199 srgb(0,0,0)
  522: 115x1004+0+601 12.1,1244.6 10023 srgb(255,255,255)
  0: 517x232+0+0 125.3,36.3 8303 srgb(255,255,255)
  8: 131x115+636+0 731.1,26.2 3908 srgb(255,255,255)
  561: 108x108+659+1497 740.4,1578.0 2954 srgb(255,255,255)
  500: 7x48+0+301 3.0,323.7 325 srgb(255,255,255)
  511: 7x42+0+454 3.0,473.7 283 srgb(255,255,255)
```
By parsing the first line (the largest transparent area)  of the verbose log`494: 640x1136+67+239 386.5,806.5 727040 srgb(255,255,255)`, we get the dimensions: x = 67, y = 239, width = 640, height = 1136.

I tested it by running `rake` command and comparing generated `offsets.json` file to the original `offests.json` file from the repository. The final results are the same as the current `offests.json` values except that we are adding `Macbook` now. 

I've added a few comments to start  discussion and double check those part of the code are okay, or maybe we could improve them. 

Thanks 🙇 🙇

### DIFF
--- a/frameit/frames_generator/Rakefile
+++ b/frameit/frames_generator/Rakefile
@@ -20,7 +20,7 @@ DEVICE_COLORS = [
   "White",
   "Yellow",
   "Gray",
-  "Coral",
+  "Coral"
 ]
 
 task(:generate_device_frames) do
@@ -45,7 +45,7 @@ task(:generate_device_frames) do
   mkdir_p(output) unless File.directory?(output)
 
   image_assets = []
-  offsets = Hash.new
+  offsets = {}
   (Dir["raw/Phones/Apple*/Device/*.png"] + Dir["raw/Computers/Apple*/Device/*.png"] + Dir["raw/Tablets/Apple*/Device/*.png"]).each do |current_raw|
     basename = File.basename(current_raw)
     next if DEVICES_TO_SKIP.include?(basename)
@@ -58,9 +58,9 @@ task(:generate_device_frames) do
     convert_image_resource(sanitaized_file_path)
     image_assets << sanitaized_basename
 
-    x, y, width, height = measure_slot(sanitaized_file_path)
+    x, y, width = measure_slot(sanitaized_file_path)
     device_name = sanitize_device_name(sanitaized_basename)
-    offsets["#{device_name}"] = {'offset' => "+#{x}+#{y}", 'width' => width}
+    offsets[device_name.to_s] = { 'offset' => "+#{x}+#{y}", 'width' => width }
   end
 
   raise "Could not find any images, make sure to copy all extracted files into ./raw" if image_assets.count < 10
@@ -71,7 +71,7 @@ task(:generate_device_frames) do
   File.write(File.join(output, "version.txt"), current_time)
   puts("Wrote the current time stamp to version.txt")
 
-  File.write(File.join(output, "offsets.json"), JSON.pretty_generate({'portrait': offsets}))
+  File.write(File.join(output, "offsets.json"), JSON.pretty_generate({ 'portrait' => offsets }))
   puts("Wrote the slot offsets to offsets.json")
 
   cp_r(output, "output/latest")
@@ -114,7 +114,7 @@ def measure_slot(path)
   command_output = convert.call
 
   sh("rm #{tmp_output}", verbose: false)
-  
+
   raw_dimensions = command_output.split("\n")[1].strip
   dimensions = /(\d+)x(\d+)\+(\d+)\+(\d+)/.match(raw_dimensions)
   x = dimensions[3]
@@ -128,9 +128,8 @@ def sanitize_device_name(filename)
   basename = File.basename(filename, File.extname(filename))
   basename = basename.gsub("Apple", "")
   basename = basename.gsub("-", " ")
-  DEVICE_COLORS.each { |color| basename.slice! color }
+  DEVICE_COLORS.each { |color| basename.slice!(color) }
   basename.strip
 end
 
 task(default: [:generate_device_frames])
-
--- a/frameit/frames_generator/Rakefile
+++ b/frameit/frames_generator/Rakefile
@@ -45,7 +45,7 @@ task(:generate_device_frames) do
   mkdir_p(output) unless File.directory?(output)
 
   image_assets = []
-  offsets = {}
+  devices = Hash.new { |h, k| h[k] = [] }
   (Dir["raw/Phones/Apple*/Device/*.png"] + Dir["raw/Computers/Apple*/Device/*.png"] + Dir["raw/Tablets/Apple*/Device/*.png"]).each do |current_raw|
     basename = File.basename(current_raw)
     next if DEVICES_TO_SKIP.include?(basename)
@@ -60,7 +60,7 @@ task(:generate_device_frames) do
 
     x, y, width = measure_slot(sanitized_file_path)
     device_name = sanitize_device_name(sanitized_basename)
-    offsets[device_name.to_s] = { 'offset' => "+#{x}+#{y}", 'width' => width }
+    devices[device_name] << { filename: sanitized_basename, x: x, y: y, width: width }
   end
 
   raise "Could not find any images, make sure to copy all extracted files into ./raw" if image_assets.count < 10
@@ -71,8 +71,21 @@ task(:generate_device_frames) do
   File.write(File.join(output, "version.txt"), current_time)
   puts("Wrote the current time stamp to version.txt")
 
+  # Create `offests.json` file
+  offsets = {}
+  devices.each do |device, models|
+    offsets[device] = { offset: "+#{models[0][:x]}+#{models[0][:y]}", width: models[0][:width] }
+  end
   File.write(File.join(output, "offsets.json"), JSON.pretty_generate({ 'portrait' => offsets }))
   puts("Wrote the slot offsets to offsets.json")
+
+  # Check if the offsets for a device are identical for all colors
+  devices.each do |device, models|
+    next if models.all? { |m| m[:x] == (models[0][:x]) && m[:y] == (models[0][:y]) && m[:width] == (models[0][:width]) }
+    puts("The offsets for #{device} are not identical for all colors:".yellow)
+    models.each { |m| puts("#{m[:filename]} +#{m[:x]}+#{m[:y]}, #{m[:width]}") }
+    puts("You can either keep the value in `offset.json` or replace it with a more appropriate value from above.")
+  end
 
   cp_r(output, "output/latest")
 
@@ -142,7 +155,7 @@ def sanitize_device_name(filename)
   basename = basename.gsub("Apple", "")
   basename = basename.gsub("-", " ")
   DEVICE_COLORS.each { |color| basename.slice!(color) }
-  basename.strip
+  basename.strip.to_s
 end
 
 task(default: [:generate_device_frames])
--- a/frameit/frames_generator/Rakefile
+++ b/frameit/frames_generator/Rakefile
@@ -52,14 +52,14 @@ task(:generate_device_frames) do
 
     cp(current_raw, output)
     output_file_path = File.join(output, basename)
-    sanitaized_basename = sanitize_filename(basename)
-    sanitaized_file_path = File.join(output, sanitaized_basename)
-    mv(output_file_path, sanitaized_file_path) unless output_file_path == sanitaized_file_path
-    convert_image_resource(sanitaized_file_path)
-    image_assets << sanitaized_basename
+    sanitized_basename = sanitize_filename(basename)
+    sanitized_file_path = File.join(output, sanitized_basename)
+    mv(output_file_path, sanitized_file_path) unless output_file_path == sanitized_file_path
+    convert_image_resource(sanitized_file_path)
+    image_assets << sanitized_basename
 
-    x, y, width = measure_slot(sanitaized_file_path)
-    device_name = sanitize_device_name(sanitaized_basename)
+    x, y, width = measure_slot(sanitized_file_path)
+    device_name = sanitize_device_name(sanitized_basename)
     offsets[device_name.to_s] = { 'offset' => "+#{x}+#{y}", 'width' => width }
   end
 
--- a/frameit/frames_generator/Rakefile
+++ b/frameit/frames_generator/Rakefile
@@ -118,16 +118,22 @@ end
 def measure_slot(path)
   tmp_output = Time.now.to_i.to_s + ".png"
 
-  convert = MiniMagick::Tool::Convert.new
-  convert << path
-  convert.negate # it makes white, black and black, white, adjusting all the colors to match
-  # connected-component analysis uniquely labels connected components in an image
-  convert.define("connected-components:verbose=true") # it enables a verbose output to the shell console
-  convert.define("connected-components:area-threshold=100") # it eliminates small objects by merging them with their larger neighbors
-  convert.connected_components("8") # it visits 8 neighbors
-  convert.auto_level
-  convert << tmp_output
-  command_output = convert.call
+  begin
+    command_output = MiniMagick::Tool::Magick.new do |magick|
+      magick << path
+      magick.negate # it makes white, black and black, white, adjusting all the colors to match
+      # connected-component analysis uniquely labels connected components in an image
+      magick.define("connected-components:verbose=true") # it enables a verbose output to the shell console
+      magick.define("connected-components:area-threshold=100") # it eliminates small objects by merging them with their larger neighbors
+      magick.connected_components("8") # it visits 8 neighbors
+      magick.auto_level
+      magick << tmp_output
+    end
+  rescue StandardError => e
+    puts(e)
+    puts("The functionality of slot measurements does not work with ImageMagick older than 7.x".red)
+    exit!
+  end
 
   sh("rm #{tmp_output}", verbose: false)
 
--- a/frameit/frames_generator/Rakefile
+++ b/frameit/frames_generator/Rakefile
@@ -82,9 +82,11 @@ task(:generate_device_frames) do
   # Check if the offsets for a device are identical for all colors
   devices.each do |device, models|
     next if models.all? { |m| m[:x] == (models[0][:x]) && m[:y] == (models[0][:y]) && m[:width] == (models[0][:width]) }
-    puts("The offsets for #{device} are not identical for all colors:".yellow)
+    puts
+    puts("The offsets for #{device} are not identical for all colors:".red)
     models.each { |m| puts("#{m[:filename]} +#{m[:x]}+#{m[:y]}, #{m[:width]}") }
     puts("You can either keep the value in `offset.json` or replace it with a more appropriate value from above.")
+    puts
   end
 
   cp_r(output, "output/latest")
--- a/frameit/frames_generator/Rakefile
+++ b/frameit/frames_generator/Rakefile
@@ -105,16 +105,29 @@ def measure_slot(path)
 
   convert = MiniMagick::Tool::Convert.new
   convert << path
-  convert.negate
-  convert.define("connected-components:verbose=true")
-  convert.define("connected-components:area-threshold=100")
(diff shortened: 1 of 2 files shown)


frameit/frames_generator/Rakefile
```diff
@@ -79,4 +100,36 @@ def convert_image_resource(path)
...
+  raw_dimensions = command_output.split("\n")[1].strip
+  dimensions = /(\d+)x(\d+)\+(\d+)\+(\d+)/.match(raw_dimensions)
+  x = dimensions[3]
+  y = dimensions[4]
+  width = dimensions[1]
+  height = dimensions[2]
+  return x.to_i, y.to_i, width.to_i, height.to_i
+end
+
+def sanitize_device_name(filename)
```
mollyIV: This method is quite tricky, because we need to sanitize a device name based on the filename to be compliant with the current conventions , i.e. `Apple iPhone 7 Plus.png` => `iPhone 7 Plus`. If that bothers us, maybe it's worth to considering if we could put the full device name to `offsets.json` file? 🤔 Although don't have much context what we are doing with `offsets.json` next.

frameit/frames_generator/Rakefile
```diff
@@ -79,4 +100,36 @@ def convert_image_resource(path)
...
+  convert.define("connected-components:area-threshold=100")
+  convert.connected_components("8")
+  convert.auto_level
+  convert << tmp_output
+  command_output = convert.call
+
+  sh("rm #{tmp_output}", verbose: false)
+
+  raw_dimensions = command_output.split("\n")[1].strip
+  dimensions = /(\d+)x(\d+)\+(\d+)\+(\d+)/.match(raw_dimensions)
```
mollyIV: I assumed that `x` and `y` offsets are always non-negative. Do you think we should support both: non-negative and negative?

frameit/frames_generator/Rakefile
```diff
@@ -40,6 +57,10 @@ task(:generate_device_frames) do
     mv(output_file_path, sanitaized_file_path) unless output_file_path == sanitaized_file_path
     convert_image_resource(sanitaized_file_path)
     image_assets << sanitaized_basename
+
+    x, y, width = measure_slot(sanitaized_file_path)
+    device_name = sanitize_device_name(sanitaized_basename)
```
janpio: typo on `sanitaized` from existing code

frameit/frames_generator/Rakefile
```diff
@@ -79,4 +100,36 @@ def convert_image_resource(path)
   image.write(path)
 end
 
+def measure_slot(path)
+  tmp_output = Time.now.to_i.to_s + ".png"
+
+  convert = MiniMagick::Tool::Convert.new
```
janpio: Maybe add a few comment on the individual parts of the command what they do (similar to other PR)?

frameit/frames_generator/Rakefile
janpio: Not until this explodes - then we can handle it.

frameit/frames_generator/Rakefile
```diff
@@ -79,4 +100,36 @@ def convert_image_resource(path)
...
+  convert.define("connected-components:verbose=true")
+  convert.define("connected-components:area-threshold=100")
+  convert.connected_components("8")
+  convert.auto_level
+  convert << tmp_output
+  command_output = convert.call
+
+  sh("rm #{tmp_output}", verbose: false)
+
+  raw_dimensions = command_output.split("\n")[1].strip
```
janpio: add comment that shows what `command_output` could look like

frameit/frames_generator/Rakefile
janpio: Yeah, we don't want touch that part if it currently works ;)

frameit/frames_generator/Rakefile
```diff
@@ -79,4 +100,36 @@ def convert_image_resource(path)
...
+  width = dimensions[1]
+  height = dimensions[2]
+  return x.to_i, y.to_i, width.to_i, height.to_i
+end
+
+def sanitize_device_name(filename)
+  basename = File.basename(filename, File.extname(filename))
+  basename = basename.gsub("Apple", "")
+  basename = basename.gsub("-", " ")
+  DEVICE_COLORS.each { |color| basename.slice!(color) }
```
janpio: Hm, does this check somehow that all the dimensions for all colors are identical?

frameit/frames_generator/Rakefile
mollyIV: Fair enough. Let's leave it as it is then 👍

frameit/frames_generator/Rakefile
mollyIV: @janpio  You mean we should fix `sanitized_file_path` and `sanitized__basename`? If yes, I will change that 🙂

frameit/frames_generator/Rakefile
mollyIV: No, it does not. If there were multiple images of the same device type with different dimensions, we would set the dimensions from the last processed image. 

We can add the check you are talking about, however how we should handle it? 🤔 
Shall we print the warning on the console (yellow text)? 
```bash
$ The dimensions for "iPhone 5c " are inconsistent: 640x1136+67+239, 640x1137+66+237.  
```
Or maybe we should even fail the script in that case? WDYT @janpio ?

frameit/frames_generator/Rakefile
mollyIV: Done.

frameit/frames_generator/Rakefile
janpio: Definitely - no need to keep that misspelling around.

frameit/frames_generator/Rakefile
janpio: Hm, this is a script only a contributor would use, so no need to be super descriptive. Failing could mean that a new release of frames becomes impossible, which is not what we want. 

So I think outputting a big fat warning that list the different affected files and there offsets would be great, then the "user" can decide if they want to keep the value written into the file or overwrite it with another one (e.g. the one that applies for black+white, or the most common one).

mollyIV: Cool, I see 👍 please let me know if that looks good to you:

```
Wrote the slot offsets to offsets.json
⚠️There are some affected files which have the "slot" offsets inconsistency: # written in yellow
Apple iPhone 8 Gold.png +61+219, 750
Apple iPhone 8 Silver.png +61+221, 750
Apple iPhone 8 Space Gray +61+219, 750
...
💡You can keep the offsets written into the "offsets.json" file or overwrite it with other values. 
```
From a technical point of view, I guess in a case of inconsistency we should print the whole family of devices under the same model, i.e. iPhone 8 Gold, Silver, Space Gray, etc.

WDYT? 🤔

janpio: Perfect.

Let's just change the text a bit:

`The offsets for #{name_of_device} are not identical for all colors:`

`You can either keep the value in `offset.json` or replace it with a more appropriate value from above.

mollyIV: @janpio I have made the changes 🙏 👀 

image

By the way, there's so many logs on the console while generating the frames:

```
...
rm -rf output
mkdir -p output/1549606566
Updating and trimming image file 'output/1549606566/Apple iPad Air 2 Silver.png'
cp raw/Tablets/Apple iPad Mini 4/Device/Apple iPad Mini 4 Gold.png output/1549606566
Updating and trimming image file 'output/1549606566/Apple iPad Mini 4 Gold.png'
cp raw/Tablets/Apple iPad Mini 4/Device/Apple iPad Mini 4 Silver.png output/1549606566
Updating and trimming image file 'output/1549606566/Apple iPad Mini 4 Silver.png'
cp raw/Tablets/Apple iPad Mini 4/Device/Apple iPad Mini 4 Space Gray.png output/1549606566
Updating and trimming image file 'output/1549606566/Apple iPad Mini 4 Space Gray.png'
Wrote names of available files to files.json
Wrote the current time stamp to version.txt
Wrote the slot offsets to offsets.json
```

Maybe it's worth to check if we could simplify them somehow (more informative messages, getting rid of some) in a separate issue? 🤔 WDYT?

janpio: Not worth the effort, as the frames are updated only every few months when Apple releases new devices.

But: You can add an empty line before and after this important message so it pops more and it is clear that something can/should be done. Can we maybe even show it in red as an error?

mollyIV: screenshot 2019-02-08 at 19 39 19

Done @janpio 💪

joshdholtz (changes requested):
@mollyIV This is another super cool change! 💪  I did however get this error when running it 😔 Any idea what could have caused this?

```sh
➜  frames_generator git:(mollyIV-frameit-measure-slot-automatically) rake generate_device_frames
Download and extract the latest devices from Facebook
  https://facebook.design/devices
Press any key when you're ready to continue
Move the content of the zip file into './raw'
Press any key when you're ready to continue
rm -rf output
mkdir -p output/1549828828
cp raw/Phones/Apple iPhone 6s Plus/Device/Apple iPhone 6s Plus Space Gray.png output/1549828828
Updating and trimming image file 'output/1549828828/Apple iPhone 6s Plus Space Gray.png'
rake aborted!
MiniMagick::Error: `convert output/1549828828/Apple iPhone 6s Plus Space Gray.png -negate -define connected-components:verbose=true -define connected-components:area-threshold=100 -connected-components 8 -auto-level 1549828829.png` failed with error:
convert: too many objects `output/1549828828/Apple iPhone 6s Plus Space Gray.png' @ error/vision.c/ConnectedComponentsImage/438.
/Users/josh/.gem/ruby/2.3.7/gems/mini_magick-4.5.1/lib/mini_magick/shell.rb:18:in `run'
/Users/josh/.gem/ruby/2.3.7/gems/mini_magick-4.5.1/lib/mini_magick/tool.rb:92:in `call'
/Users/josh/Projects/fastlane/fastlane/frameit/frames_generator/Rakefile:130:in `measure_slot'
/Users/josh/Projects/fastlane/fastlane/frameit/frames_generator/Rakefile:61:in `block (2 levels) in <top (required)>'
/Users/josh/Projects/fastlane/fastlane/frameit/frames_generator/Rakefile:49:in `each'
/Users/josh/Projects/fastlane/fastlane/frameit/frames_generator/Rakefile:49:in `block in <top (required)>'
/Users/josh/.gem/ruby/2.3.7/gems/rake-12.3.1/exe/rake:27:in `<top (required)>'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/cli/exec.rb:74:in `load'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/cli/exec.rb:74:in `kernel_load'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/cli/exec.rb:28:in `run'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/cli.rb:463:in `exec'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/cli.rb:27:in `dispatch'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/cli.rb:18:in `start'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/exe/bundle:30:in `block in <top (required)>'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/lib/bundler/friendly_errors.rb:124:in `with_friendly_errors'
/Users/josh/.gem/ruby/2.3.7/gems/bundler-1.17.3/exe/bundle:22:in `<top (required)>'
/Users/josh/.gem/ruby/2.3.7/bin/bundle:22:in `load'
/Users/josh/.gem/ruby/2.3.7/bin/bundle:22:in `<main>'
Tasks: TOP => generate_device_frames
(See full trace by running task with --trace)
```

mollyIV:
Hey @joshdholtz , thanks for testing the changes 👍 Unfortunately I was not able to reproduce the issue. According to `minimagick` [documentation](https://github.com/minimagick/minimagick) :

> ImageMagick or GraphicsMagick command-line tool has to be installed.

Could you please check what version you use by running `convert -version` command? 🤔 

Here's the output from my local machine:

```bash
❯ convert -version
Version: ImageMagick 7.0.8-11 Q16 x86_64 2018-09-17 https://www.imagemagick.org
Copyright: © 1999-2018 ImageMagick Studio LLC
License: https://www.imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules
Delegates (built-in): bzlib freetype jng jpeg ltdl lzma png tiff xml zlib
```

🙇

joshdholtz:
@mollyIV lol, not that one 😉 

```sh
➜  fastlane git:(master) convert -version
Version: ImageMagick 6.9.9-47 Q16 x86_64 2018-05-22 https://www.imagemagick.org
Copyright: © 1999-2018 ImageMagick Studio LLC
License: https://www.imagemagick.org/script/license.php
Features: Cipher DPC Modules
Delegates (built-in): bzlib freetype jng jpeg ltdl lzma png tiff xml zlib
```

So... thoughts on how to best handle this? 🤔 I'm totally fine upgrading but we should probably add a version check somewhere so that this gets executed the same every time? 🙃

janpio:
Can you confirm the problem is actually gone when upgrading to 7.x @joshdholtz? If so, @mollyIV should find out if this actually a new feature and if we can "guard" its usage somehow.

mollyIV:
Hey @janpio @joshdholtz 👋 
I did a bit more research on the issue related to `convert` tool version. 

I was reading the release notes of [imagemagick](https://imagemagick.org/script/changelog.php) and [minimagick](https://github.com/minimagick/minimagick) to find the more info about input params of `convert` [command](https://imagemagick.org/script/convert.php). Unfortunately no luck there 😞 

However during my research, I found interesting info on `minimagick` docs, [Metal](https://github.com/minimagick/minimagick#metal) section:

> If you're on ImageMagick 7, you should probably use MiniMagick::Tool::Magick, though the legacy MiniMagick::Tool::Convert and friends will work too. On ImageMagick 6 MiniMagick::Tool::Magick won't be available, so you should instead use MiniMagick::Tool::Convert and friends.

I did the change in the code locally (used magick instead of convert) and indeed, the command behaves as expected - `offsets.json` file looks the same ✨ 

Maybe we could use `MiniMagick::Tool::Magick` as recommended and restrict the version of `minimagick` to `gem 'mini_magick', '~> 4.9.0'`? `4.9.0` is the version where they introduced [ImageMagick7](https://github.com/minimagick/minimagick/releases/tag/v4.9.0) support.

Still wondering how we can specify the requirement about `imagemagick7` version. The tool is available on [homebrew](https://formulae.brew.sh/formula/imagemagick). 

One solution is to add "Requirements"  section to README.md file in `frames_generator` project.

WDYT? 🤔

janpio:
(I have in the back of my head a feeling that I actually increased the minimagick requirement because of a Windows bug, and had tested most of the stuff for ImageMagick7. I really don't think we should limit our use to outdated software but find a better solution.)

mollyIV:
🤔 not sure I understood what you are suggesting @janpio 

> I really don't think we should limit our use to outdated software but find a better solution.

That's why I suggested to bump the `mini_magick` version and use `MiniMagick::Tool::Magick` 🙂The problem is how will we explicitly require the `ImageMagick7` version. 

Do you have other ideas how we could solve the issue?

janpio:
Oh, you want to _require_ at least the new version - I misread that you wanted to pin on an older version ;)

Then: Just catch that error and output that this functionality doesn't work with ImageMagick older than 7.x and output that as an error or message to the user. As this is a maintainer facing operation, no need to get fancy here. They can upgrade to get the full functionality.

mollyIV:
Actually I think we are not able to catch the specific exception about using an older version of `convert` tool. What we can do however is to catch any exception, print it to the console and add our custom output:

```bash
❯ rake
...

Updating and trimming image file 'output/1549999413/Apple iPhone 6s Plus Space Gray.png'
`magick output/1549999413/Apple iPhone 6s Plus Space Gray.png -negate -define connected-components:verbose=true -define connected-components:area-threshold=100 -connected-components 8 -auto-level -error-sssfds 1549999413.png` failed with error:
magick: unrecognized option `-force-exception-throwing' at CLI arg 10 @ fatal/magick-cli.c/ProcessCommandOptions/428.
The functionality of slot measurements does not work with ImageMagick older than 7.x
```

From a technical point of view it'd look like this:

```ruby
begin
    command_output = MiniMagick::Tool::Magick.new do |magick|
      magick << path
      magick.negate # it makes white, black and black, white, adjusting all the colors to match
      # connected-component analysis uniquely labels connected components in an image
      magick.define("connected-components:verbose=true") # it enables a verbose output to the shell console
      magick.define("connected-components:area-threshold=100") # it eliminates small objects by merging them with their larger neighbors
      magick.connected_components("8") # it visits 8 neighbors
      magick.auto_level
      magick << tmp_output
    end
  rescue Exception => e
    puts e
    puts("The functionality of slot measurements does not work with ImageMagick older than 7.x".red)
    exit!
  end
```

I am not very experienced with ruby, so could anyone confirm it is a good approach? 🙏 
Looks good for you guys? 🤔

mollyIV:
@janpio @joshdholtz I just pushed the newest changes, please have a  👀 🙏

janpio (commented):
Looks good to me.

joshdholtz (approved):
This is 🔥 🔥 🔥 Thanks for doing this!! ❤️

janpio:
Now we can update the frames all the time without an additional manual labour. So awesome!

PR #14208, opened 2019-02-07
Speed up tests by moving IAP stubbing to the relevant specs

snatchev:
This PR improves test runs for spaceship by around 15% on my machine. From 77 seconds, to around 65 seconds.

The stubbing method `itc_stub_iap` has 21 calls to `stub_request` which would get executed on every single example. These stubs were only needed in the specs relevant to IAP. I am sure it can be scoped further, but it's probably not such an easy win.


janpio (approved):
:car: 💨

janpio:
Couldn't this also be worth to be done to the other stubbings?

snatchev:
Yes. From my preliminary benchmarks, these were the most bang for the buck.
Some stubbings like login, etc are used a lot more.

PR #14210, opened 2019-02-07
[scan] Always zip build products

sa-spag:
### Checklist

### Motivation and Context
`should_zip_build_products` has no effect if tests fail or if the test execution failed. As build products can be handy to debug these cases, I suggest to always zip them accordingly.

### Description
Moved `zip_build_products` call up so it's called before any premature exit. No need to update `should_zip_build_products` in my opinion.

### DIFF
--- a/scan/lib/scan/runner.rb
+++ b/scan/lib/scan/runner.rb
@@ -93,6 +93,7 @@ module Scan
       puts("")
 
       copy_simulator_logs
+      zip_build_products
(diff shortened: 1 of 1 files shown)


sa-spag:
I just signed the CLA.

joshdholtz (approved):
Ahhh, this makes so much more sense 😊 Thanks for the contribution!

PR #14281, opened 2019-02-20
[import_certificate] Improve documentation

janpio:
resort options, add text about private keys and add another example


joshdholtz (approved):
👍

Issue #14283, opened 2019-02-20
[Spaceship] Unexpected response: There was an error generating the provisioning profile


brentdur:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When trying to update a provisioning profile using Spaceship, I get the referenced error. Taking a look at fastlane's request vs the request that the Apple Web UI sends, I believe this is due to the use of `template` in the request body. Spaceship is using the `template_name` ~ `template.purpose_display_name`, which results in the request data resembling `template: "Company (Development)"`. Apple Developer Center users the equivalent of `template.purpose_name` which results in the request data being `template: "company-development"`. I think the fix here is to adjust which parameter of `template` is being used by the provisioning profile class.

##### Command executed

```
> fastlane spaceship
dev = Spaceship::Portal.provisioning_profile.all
dev = dev.first
dev.update!
```
(I realize here I've made no changes to the profile, but this is the shortest path to reproduce)

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> 
dev = Spaceship::Portal.provisioning_profile.all
dev = dev.first
dev.update!
spaceship::UnexpectedResponse: {"responseId"=>"XXX", "resultCode"=>8530, "resultString"=>"There was an error generating the provisioning profile 'XXXXXX'.  Please try again.  If the problem persists, please contact Apple Developer Program Support. https://developer.apple.com/support", "userString"=>"There was an error generating the provisioning profile 'XXXX'.  Please try again.  If the problem persists, please contact Apple Developer Program Support. https://developer.apple.com/support", "creationTimestamp"=>"2019-02-20T...", "protocolVersion"=>"XXXX", "userLocale"=>"en_US", "requestUrl"=>"https://developer.apple.com/services-account/XXXX/account/ios/profile/regenProvisioningProfile.action", "httpCode"=>200}
from ~/.fastlane/bin/bundle/lib/ruby/gems/2.2.0/gems/fastlane-2.116.1/spaceship/lib/spaceship/client.rb:708:in `parse_response' </pre>
</details>

### Environment

<details>
<summary>✅ fastlane environment ✅</summary>

### Stack

[...]

PR #14284, opened 2019-02-20
[spaceship] Change provisioning profile update request to match what Apple Developer passes

brentdur:
### Checklist

### Motivation and Context

Issue: https://github.com/fastlane/fastlane/issues/14283

When trying to update a provisioning profile using Spaceship, I've been getting the `Unexpected response: There was an error generating the provisioning profile 'XXXXXX'`. Taking a look at fastlane's request vs the request that the Apple Web UI sends, there was a difference in the `template` field in the request body. Spaceship was using `template.purpose_display_name`, which results in the request data resembling `template: "Company (Development)"`. Apple Developer Center users the equivalent of `template.purpose_name` which results in the request data being `template: "company-development"`. This change resolve the issue.

### Description

I've altered the data getting passed from `provisioning_profile.rb` to use the `template.purpose_name` instead of the `template.purpose_display_name`. Passing this request through Charles and comparing it to the Apple Developer Center request shows that these fields on the requests now match. And I'm no longer getting the ambiguous error.


brentdur:
I signed it!

janpio:
Similar call is found at https://github.com/fastlane/fastlane/blob/5f61fd44a6ba3c6b5e1536af580c86148f438451/spaceship/lib/spaceship/portal/provisioning_profile.rb#L304 - should this be changed as well?

brentdur:
I don’t think so. That call is for the create and given that the template check doesn’t exist, I think it should be left as the passed in value.

brentdur:
@janpio :D

BalestraPatrick:
Hey @janpio! Would it be possible to get this changed merged? We currently this bug in automating our provisioning profile management that may be fixed with this PR.

spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -457,7 +457,7 @@ def update!
             devices.map(&:id),
             mac: mac?,
             sub_platform: tvos? ? 'tvOS' : nil,
-            template_name: template_name
+            template_name: is_template_profile ? template.purpose_name : nil
```
janpio: Did I understand correctly that this additional `is_template_profile` conditional is to make this equal to the previous `template_name`, which is actually a call to a method, that includes this conditional?
Can't we remove that method then - or change it to using the correct field instead of replacing the call here? The other `template_name` usages in this file seem to be for a method parameter, and not that method.

(Don't know this piece of code very well)

janpio:
Fair enough, took a look at the code and left a comment. 

(ping @max-ott)

max-ott:
Looking good from my end. 👍

spaceship/lib/spaceship/portal/provisioning_profile.rb
brentdur: Seems fair. Initially I thought this function was being used elsewhere but I'll drop the method

spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -548,12 +548,6 @@ def template
 
         @template ||= ProvisioningProfileTemplate.factory(profile_details['template'])
       end
-
-      # @return (String) The name of the template (as displayed in Dev Portal)
-      #   or nil if provisioning profile doesn't have a template
-      def template_name
-        is_template_profile ? template.purpose_display_name : nil
-      end
```
niw: So... now I am facing this issue on my environment and found this patch.

However, since on the latest master, this method is used from 2 places (`create_provisioning_profile!` and `repair_provisioning_profile!`,) so, to address the issue, I think instead of removing this method,
rewrite this method behavior or add new method that returns `purpose_name` instead.

janpio:
@niw Looking forward to your PR.
Sorry @brentdur we didn't have the time to review and merge this.

PR #14293, opened 2019-02-21
[spaceship] Docs: Fix vendor_id description

max-ott:
Previous description text was incorrect. SKU is not automatically generated by Apple, but rather something internally created by the developer. 

> A unique ID for your bundle. You can use letters, numbers, hyphens, periods, and underscore. The SKU cannot start with a hyphen, period, or underscore.

### Checklist

### Description
Previous description was incorrect. Fixing that with this commit.

### DIFF
--- a/spaceship/lib/spaceship/tunes/application.rb
+++ b/spaceship/lib/spaceship/tunes/application.rb
@@ -25,7 +25,7 @@ module Spaceship
       #   "Spaceship App"
       attr_accessor :name
 
-      # @return (String) The Vendor ID provided by App Store Connect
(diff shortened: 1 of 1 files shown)


janpio:
Does it make sense to keep the accessor name identical anyway?
Any source to Apple documentation on this value?

max-ott:
Official docs here: https://help.apple.com/app-store-connect/#/dev219b53a88

Reference > My Apps > App Information > Property > SKU

joshdholtz (approved):
🙌 🔥 🚀

PR #13556, opened 2018-10-17
iOS SnapshotHelper.swift Set locale only when not empty

river2202:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I've updated the documentation if necessary.

### Motivation and Context

Few of our CI machines were missing ~/Library/Caches/tools.fastlane/locale.txt and SnapshotHelper set the app locale to empty string and cause date time format returning wrong string like: expecting "01 Jan 2018" but got "01 M01 2018"

https://github.com/fastlane/fastlane/issues/13555

### Description

setLocale will skip -AppleLocale parameter if locale.isEmpty as setLanguage did when deviceLanguage is empty

Issue #13555, opened 2018-10-17
Date time format wrong when ~/Library/Caches/tools.fastlane/locale.txt file missing


river2202:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
SnapshotHelper.swift set empty locale when ~/Library/Caches/tools.fastlane/locale.txt file missing and cause the Date time format wrong. 

PR open for this issue.

### Environment

<details>
  <pre><details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14                                       |
| Ruby                        | 2.4.1                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.17.1 (Apple Git-112)          |
| Installation Source         | ~/.rvm/gems/ruby-2.4.1/bin/fastlane         |
| Host                        | Mac OS X 10.14 (18A389)                     |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.4.1/lib                |
| OpenSSL Version             | OpenSSL 1.1.0f  25 May 2017                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 10.0                                        |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_AU.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
default_platform :ios
platform :ios do

  before_all do |lane, options|
    xcode_select "/Applications/Xcode.app"
    clear_derived_data
    cocoapods(
      use_bundle_exec: true
    )
    ENV['CI_BUILD_NUMBER'] = options[:build_number].to_s
    ENV['CI_JOB_ID'] = options[:job_id].to_s
[...]

PR #14306, opened 2019-02-24
[snapshot] Set locale only when identifier is valid

jszumski:
### Checklist
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md) 🔑

### Motivation and Context
Similar to #13555 and https://github.com/fastlane/fastlane/pull/13556, a CI machine was missing `~/Library/Caches/tools.fastlane/locale.txt` and `~/Library/Caches/tools.fastlane/language.txt` which ended up breaking date formatted strings where "January" was replaced with "M01", "February" with "M02", etc.

### Description
Fixed issue where a bad locale could be set if neither a locale nor language string could be detected.

If language.txt and locale.txt are both missing, the default value of `deviceLanguage` (the empty string) was incorrectly used to create the locale passed via `-AppleLocale`.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -115,7 +115,7 @@ open class Snapshot: NSObject {
             print("Couldn't detect/set locale...")
         }
         
-        if locale.isEmpty {
+        if locale.isEmpty && !deviceLanguage.isEmpty {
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
This looks good!! Thanks for catching this and thanks for the contribution ❤️

Issue #14310, opened 2019-02-25
Issues with Keychains with spaces (" ") and shellescaping


datajockey:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Keychains with spaces in the path are not being escaped correctly and causing unexpected issues. I discovered this while importing signing credentials, which failed silently, but later failed when the signing credentials could not be found. The user had a space in his home directory path and thus his `login` keychain path.

I created the following Fastfile with two targets to test this: 

```
desc "Test Control"  # no spaces in path
lane :test_control do
  create_keychain ({
    path: '/tmp/test_control.keychain',
    password: 'testpassword',
    require_create: true
  })
  cert ({
    development: true,
    keychain_path: '/tmp/test_control.keychain',
    keychain_password: 'testpassword'
  })
end

desc "Test Spaces"
lane :test_spaces do
  create_keychain ({
    path: '/tmp/test spaces.keychain',
    password: 'testpassword',
    require_create: true
  })
  cert ({
    development: true,
    keychain_path: '/tmp/test spaces.keychain',
    keychain_password: 'testpassword'
  })
end
```
The `test_control` target works as expected and the `test_spaces` target does not create a valid keychain or import certificate or private key. This screenshot shows the bad keychain:

![screen shot 2019-02-25 at 11 06 28 am](https://user-images.githubusercontent.com/707113/53366878-a90e6c00-3912-11e9-9557-2c868fa3a33e.png)

##### Command executed
```
bundle exec fastlane test_spaces
```

**Note:** The 'test spaces` keychain can be pain to clean up between test runs. I've found that you have to create a keychain that actually matches the keychain listed in the search paths before deleting it:

```
# Clean up 'test\ spaces'
security create-keychain -p testpassword '/tmp/test\ spaces.keychain'
security delete-keychain '/tmp/test\ spaces.keychain'
```

##### Complete output when running fastlane, including the stack trace and command used

As you can see from the following output Fastlane indicates success, however inspecting the keychain shows otherwise. 

<details>
  <pre>
[✔] 🚀
[15:30:35]: Driving the lane 'test_spaces' 🚀
[...]

PR #14311, opened 2019-02-25
[fastlane] fix keychain shellescaping

datajockey:
### Checklist

### Motivation and Context

There are issues with keychains being escaped and/or used in an inconsistent way. 
See issue #14310 for a description and reproduction steps.

### Description

1. The first commit ensures that keychain creation security commands use the escaped keychain path and avoids double escaping via `.shelljoin` 

   From the shelljoin docs:
   > All elements are joined into a single string with fields separated by a space, where each element is escaped for the Bourne shell and stringified using to_s.

2. The second commit removes the quotes around the escaped keychain path while importing.

   From the shellescape docs:

3. ~~The third commit constrains the certificate check to the provided keychain. 
   Through my test Fastfile (see #14310), I discovered that if my credentials were in any keychain in the search path, they would not be installed in the provided/target keychain. 
   I thought this was odd so I added this commit even though it is not directly related to the underlying issues.~~ 
   **Upate:** This commit was moved to a new branch to be evaluated a separate PR.

Testing consisted of manually running a Fastfile (see #14310) with control and test scenarios and comparison with command line executions of the security command. 

The unit tests do not actually execute the commands, but compare the command that would be issued to an expected value. I updated the unit tests to reflect my changes and identify things I missed.

**Note:** I doubt it is a complete solution but a step in the right direction. Please let me know of any changes or extensions. I hope this helps!

### DIFF
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -11,20 +11,20 @@ module Fastlane
         escaped_password = params[:password].shellescape
 
         if params[:name]
-          escaped_name = params[:name].shellescape
-          keychain_path = "~/Library/Keychains/#{escaped_name}"
+          keychain_path = "~/Library/Keychains/#{params[:name]}"
         else
-          keychain_path = params[:path].shellescape
+          keychain_path = params[:path]
         end
+        escaped_keychain_path = keychain_path.shellescape
 
-        if keychain_path.nil?
+        if escaped_keychain_path.nil?
           UI.user_error!("You either have to set :name or :path")
         end
 
         commands = []
 
-        if !exists?(keychain_path)
-          commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{keychain_path}", log: false)
+        if !exists?(escaped_keychain_path)
+          commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{escaped_keychain_path}", log: false)
         elsif params[:require_create]
           UI.abort_with_message!("`require_create` option passed, but found keychain '#{keychain_path}', failing create_keychain action")
         else
@@ -38,21 +38,22 @@ module Fastlane
             Actions.lane_context[Actions::SharedValues::ORIGINAL_DEFAULT_KEYCHAIN] = Fastlane::Actions.sh("security default-keychain", log: false).strip
           rescue
           end
-          commands << Fastlane::Actions.sh("security default-keychain -s #{keychain_path}", log: false)
+          commands << Fastlane::Actions.sh("security default-keychain -s #{escaped_keychain_path}", log: false)
         end
 
-        commands << Fastlane::Actions.sh("security unlock-keychain -p #{escaped_password} #{keychain_path}", log: false) if params[:unlock]
+        commands << Fastlane::Actions.sh("security unlock-keychain -p #{escaped_password} #{escaped_keychain_path}", log: false) if params[:unlock]
 
         command = "security set-keychain-settings"
         command << " -t #{params[:timeout]}" if params[:timeout]
         command << " -l" if params[:lock_when_sleeps]
         command << " -u" if params[:lock_after_timeout]
-        command << " #{keychain_path}"
+        command << " #{escaped_keychain_path}"
 
         commands << Fastlane::Actions.sh(command, log: false)
 
         if params[:add_to_search_list]
           keychains = Action.sh("security list-keychains -d user").shellsplit
+          # Do not use 'escaped_keychain_path' here because .shelljoin will cause double escaping
           keychains << File.expand_path(keychain_path)
           commands << Fastlane::Actions.sh("security list-keychains -s #{keychains.shelljoin}", log: false)
         end
@@ -60,12 +61,12 @@ module Fastlane
         commands
       end
 
-      def self.exists?(keychain_path)
-        keychain_path = File.expand_path(keychain_path)
+      def self.exists?(escaped_keychain_path)
+        escaped_keychain_path = File.expand_path(escaped_keychain_path)
 
         # Creating Keychains using the security
         # CLI appends `-db` to the file name.
-        File.exist?("#{keychain_path}-db") || File.exist?(keychain_path)
+        File.exist?("#{escaped_keychain_path}-db") || File.exist?(escaped_keychain_path)
       end
 
       def self.description
--- a/fastlane/spec/actions_specs/create_keychain_spec.rb
+++ b/fastlane/spec/actions_specs/create_keychain_spec.rb
@@ -20,7 +20,7 @@ describe Fastlane do
         expect(result[1]).to_not(include("-u"))
         expect(result[1]).to include(keychain_path.to_s)
         expect(result[2]).to start_with("security list-keychains -s")
-        expect(result[2]).to end_with(File.expand_path(keychain_path.to_s).to_s)
+        expect(result[2]).to end_with(File.expand_path(keychain_path.to_s).shellescape.to_s)
       end
 
       it "works with name and password that contain spaces or `\"`" do
--- a/fastlane/spec/actions_specs/create_keychain_spec.rb
+++ b/fastlane/spec/actions_specs/create_keychain_spec.rb
@@ -2,43 +2,50 @@ describe Fastlane do
   describe Fastlane::FastFile do
     describe "Create keychain Integration" do
       it "works with name and password" do
+        keychain_password = "testpassword"
+        keychain_name = "test.keychain"
+        keychain_path = "~/Library/Keychains/#{keychain_name}"
         result = Fastlane::FastFile.new.parse("lane :test do
           create_keychain ({
-            name: 'test.keychain',
-            password: 'testpassword',
+            name: '#{keychain_name}',
+            password: '#{keychain_password}',
           })
         end").runner.execute(:test)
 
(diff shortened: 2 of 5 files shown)


datajockey:
I signed it!

datajockey:
The third commit to check if the certs were installed in the target keychain is problematic. 
I can drop this commit or with some direction, I can fix it up. 
Please advise.

janpio:
Without looking at it in detail, best move the third commit to its own PR so we can look at this different thing in isolation. (Thanks for the great separation into commits etc - makes it easy to follow what is going on here.)

fastlane/spec/actions_specs/import_certificate_spec.rb
```diff
@@ -11,7 +11,7 @@
         password = 'testpassword'
 
         keychain_path = File.expand_path(File.join('~', 'Library', 'Keychains', keychain))
-        expected_command = "security import #{cert_name} -k '#{keychain_path}' -P #{password} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
+        expected_command = "security import #{cert_name} -k #{keychain_path} -P #{password} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
```
janpio: isn't this missing `.shellescape` then?

fastlane/spec/actions_specs/import_certificate_spec.rb
```diff
@@ -66,7 +66,7 @@
         password = 'testpassword'
 
         keychain_path = File.expand_path(File.join('~', 'Library', 'Keychains', keychain))
-        expected_command = "security import #{cert_name} -k '#{keychain_path}' -P #{password} -T /usr/bin/codesign -T /usr/bin/security"
+        expected_command = "security import #{cert_name} -k #{keychain_path} -P #{password} -T /usr/bin/codesign -T /usr/bin/security"
```
janpio: Same here:
isn't this missing `.shellescape` then?

fastlane/spec/actions_specs/import_certificate_spec.rb
datajockey: Escaping is a no-op here because there are no special characters in the test parameters. 
Good catch and the other parameters should be escaped too.

fastlane/spec/actions_specs/import_certificate_spec.rb
datajockey: Same here:
> Escaping is a no-op here because there are no special characters in the test parameters. 
> Good catch and the other parameters should be escaped too.

datajockey:
I moved the third commit to a new branch as recommended and reset HEAD to second commit.  
I addressed the missing `.shellescape` comments in the amended second commit and force pushed the branch.

janpio:
Hm yeah, the noop in the specs are strange to decide - you don't need it, but it makes it more correct :/ Thanks for changing.

match/spec/utils_spec.rb
```diff
@@ -6,7 +6,7 @@
 
     describe 'import' do
       it 'finds a normal keychain name relative to ~/Library/Keychains' do
-        expected_command = "security import item.path -k '#{Dir.home}/Library/Keychains/login.keychain' -P #{''.shellescape} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
+        expected_command = "security import item.path -k #{Dir.home}/Library/Keychains/login.keychain -P #{''.shellescape} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
```
janpio: Isn't this missing a `.shellescaoe` for the `-k` value here? (relevant e.g. when `Dir.home` has a space)

match/spec/utils_spec.rb
```diff
@@ -25,7 +25,7 @@
       it 'treats a keychain name it cannot find in ~/Library/Keychains as the full keychain path' do
         tmp_path = Dir.mktmpdir
         keychain = "#{tmp_path}/my/special.keychain"
-        expected_command = "security import item.path -k '#{keychain}' -P #{''.shellescape} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
+        expected_command = "security import item.path -k #{keychain} -P #{''.shellescape} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
```
janpio: same as above

match/spec/utils_spec.rb
```diff
@@ -50,7 +50,7 @@
       end
 
       it "tries to find the macOS Sierra keychain too" do
-        expected_command = "security import item.path -k '#{Dir.home}/Library/Keychains/login.keychain-db' -P #{''.shellescape} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
+        expected_command = "security import item.path -k #{Dir.home}/Library/Keychains/login.keychain-db -P #{''.shellescape} -T /usr/bin/codesign -T /usr/bin/security 1> /dev/null"
```
janpio: and again

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -38,34 +38,35 @@ def self.run(params)
...
         command << " -l" if params[:lock_when_sleeps]
         command << " -u" if params[:lock_after_timeout]
-        command << " #{keychain_path}"
+        command << " #{escaped_keychain_path}"
 
         commands << Fastlane::Actions.sh(command, log: false)
 
         if params[:add_to_search_list]
           keychains = Action.sh("security list-keychains -d user").shellsplit
+          # Do not use 'escaped_keychain_path' here because .shelljoin will cause double escaping
```
janpio: all the changes in this file are just so this is possible, correct?

match/spec/utils_spec.rb
datajockey: @janpio, Yes, I noticed other cases where the `spec` files needed to be updated with `.shellescape` after I pushed my initial update. 

I have not been unable to get back to them, but I will address these today.

fastlane/lib/fastlane/actions/create_keychain.rb
datajockey: I did this to be more explicit with what is being used throughout the file.  I.e. `escaped_keychain_path` vs sprinkling `.shellescape` in all of these cases except for the `.shelljoin` case where we want to use the plain `keychain_path`.

datajockey: I did this to be more explicit with what is being used throughout the file.  I.e. `escaped_keychain_path` vs sprinkling `.shellescape` in all of these cases except for the `.shelljoin` case where we want to use the plain `keychain_path`.

match/spec/utils_spec.rb
datajockey: I pushed an update to these.

match/spec/utils_spec.rb
datajockey: Same here.

match/spec/utils_spec.rb
datajockey: Same here.

fastlane/spec/actions_specs/create_keychain_spec.rb
```diff
@@ -2,106 +2,122 @@
...
-        expect(result[1]).to include('~/Library/Keychains/test.keychain')
+        expect(result[0]).to eq("security create-keychain -p #{keychain_password.shellescape} #{keychain_path.shellescape}")
+        expect(result[1]).to start_with("security set-keychain-settings")
+        expect(result[1]).to include("-t 300")
+        expect(result[1]).to_not(include("-l"))
+        expect(result[1]).to_not(include("-u"))
+        expect(result[1]).to include(keychain_path.to_s)
         expect(result[2]).to start_with("security list-keychains -s")
-        expect(result[2]).to end_with(File.expand_path('~/Library/Keychains/test.keychain').to_s)
+        expect(result[2]).to end_with(File.expand_path(keychain_path.to_s).to_s)
```
datajockey: @janpio Note that I did not change this to `keychain_path.shellescape` because of the link in the link causes the keychain path to be expanded:
https://github.com/fastlane/fastlane/pull/14311/files#diff-0231a345cb1ba3f328959032b555024cR57

So `~/Library/Keychains/...` becomes `/Users/foo/Library/Keychains/...` in this test case. It seems a little odd so if you have any thoughts, let me know.

datajockey: @janpio Nevermind the note above, I realized my mistake and resolved the use of `.shellescape` in this test case.

joshdholtz (approved):
@datajockey This PR looks good! Thanks for contribution here 👍 And thanks for all the reviewing, @janpio ❤️

datajockey:
@joshdholtz Glad I can help out! 
@janpio Thank you for your help and guidance.

Following up on the problematic third commit which @janpio suggested that I break out into a separate PR. It was meant to address the discovery that signing credentials were not being installed in the target keychain if they were already installed in another keychain on the search path. This non-deterministic behavior seemed odd to me, but maybe a desired behavior so I wanted to get your thoughts. 

My initial commit for this issue passed all local rspec tests, but failed on the Ubuntu and Appveyor checks. It was unclear to me what conditions were being tested that lead to the failures, so there is work to do to isolate and ready a PR.

janpio:
Feel free to create a `[WIP]` PR @datajockey, then we can maybe help debug this further. (And also to avoid this getting forgotten and lost)

datajockey:
Will do @janpio 
👍

PR #14322, opened 2019-02-27
[supply] Download existing screenshots with `supply init`

scorpion35:
### Checklist
  - Yes. No errors reported!
  ```
    Finished in 3 minutes 56.2 seconds (files took 3.71 seconds to load)
    5563 examples, 0 failures
  ```
  - Yes. Submitted changes done by rubocop.
  - Not yet. Will take a look soon.

### Motivation and Context
**Why is this change required? What problem does it solve?**
  - Downloads original screenshots from Google.

### Description
 **Describe your changes in detail**
  - It seems images are stored in `webp` format on Google's servers, and they return compressed images when requested. 
  - Because it's `webp` format, we can request for resize and then download the image!
  - Then download the image into correct folder.
  - Also removes the `alpha` from the image if selected.

**Please describe in detail how you tested your changes.**
  - I am not sure how to test with the `fastlane supply`. Instead I tested by importing `supply` library 
 and calling the method directly, like below,
```ruby
      Supply.config = { :metadata_path => "#{android_folder_path}", :json_key => "<my_service_json>", :package_name => "com.my.awesomeapp" }
      Supply::Setup.new.perform_download()
```

P.s: I made these changes from my work computer, so there will be '2' authors. But both are me :D I confirm it is me who made these changes, and would like to merge these changes into Fastlane. I have signed the CLA using my both accounts!


scorpion35:
Here is the log file and some screenshots for downloaded images - 

[fastlane-supply-images.log](https://github.com/fastlane/fastlane/files/2911449/fastlane-supply-images.log)

fastlane-supply-sample1

fastlane-supply-sample2

scorpion35:
Also it seems `spaceship/lib/spaceship/tunes/member.rb` sneakily made it into this PR. Those changes have already been merged last month, please ignore this file!

janpio:
I removed the changes of `member.rb`.

scorpion35:
Can you give me some pointers for updating the documentation, please? Not sure how to edit this website - https://docs.fastlane.tools/actions/supply/

Let me know if I need to do anything as well, please. Would like to wrap this up soon! :)

janpio:
The content of that docs page can be found at https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/upload_to_play_store.md - you can just add those changes to this PR for best merge-ability ;)

Someone will actually have to test this PR if this works - but I guess I will do that later.

fastlane/lib/fastlane/actions/docs/upload_to_play_store.md
```diff
@@ -1,5 +1,5 @@
 <p align="center">
-  <img src="/img/actions/supply.png" width="250">
+  <img src="assets/supply.png" width="250">
```
janpio: This will not work in the rendered state of the article - the image currently works on the docs site, so no need to change anything here.

fastlane/lib/fastlane/actions/docs/upload_to_play_store.md
```diff
@@ -31,7 +31,7 @@ _supply_ uploads app metadata, screenshots, binaries, and app bundles to Google
 
 ## Setup
 
-Setup consists of setting up your Google Developers Service Account
+Setup consists of setting up your Google Developers Service Account - https://docs.fastlane.tools/actions/supply/#setup
```
janpio: accident?

scorpion35: oh I get it now. This doc is somehow imported and shown on the website. I am looking at the preview in vscode and was thinking it's broken, lol. Will undo these changes!

scorpion35:
I think the doc looks good, no changes required I think. Let me know if you need me to do anything else!

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
 
-          url = client.fetch_images(image_type: image_type, language: listing.language).last
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next unless urls
+
+          if urls.length > 0
+            image_counter = 1
+            urls.each do |url|
+              url = "#{url}=w6000-h6000" # blowing up the sizes intentionally. So, Google would return the original image sizes as they are uploaded.
```
janpio: 6000 is a magic number here?

```suggestion
              url = "#{url}=w6000-h6000" # using huge dimentions intentionally so Google returns the images in their original size they were uploaded
```

scorpion35: Just picked some random number! Too bad for someone who is actually using image sizes over this. They would get 'compressed' images then :D

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+              image = MiniMagick::Image.open(path)
+              image_details = JSON.parse(image.details.to_json, symbolize_names: true)
 
-          path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, "#{image_type}.png")
-          File.write(path, Net::HTTP.get(URI.parse(url)))
+              is_alpha_present = image_details.keys.any? { |k| k == :Alpha }
+              image.alpha('remove') if is_alpha_present && remove_alpha
+
+              image.format(image_format) # Properly formatting the final image. Sometimes "jpg" or "webp" or "png" formats are returned by Google.
+              image.write(path) # Properly formatted file is written to `final_path`
```
janpio: Is this path different to the previous path? 
Do you edit the files in place?

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+          else
+            UI.message("\tNo images found, nothing to download.")
+          end # if(urls.length > 0)
         rescue => ex
           UI.error(ex.to_s)
           UI.error("Error downloading '#{image_type}' for #{listing.language}...")
         end
+      end # allowed_imagetypes.each do |image_type|
+
+      3.times do
```
janpio: ???

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
 
-          path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, "#{image_type}.png")
-          File.write(path, Net::HTTP.get(URI.parse(url)))
+              is_alpha_present = image_details.keys.any? { |k| k == :Alpha }
+              image.alpha('remove') if is_alpha_present && remove_alpha
+
+              image.format(image_format) # Properly formatting the final image. Sometimes "jpg" or "webp" or "png" formats are returned by Google.
+              image.write(path) # Properly formatted file is written to `final_path`
+              UI.message("\tDownloaded  - #{path}")
+              UI.message("\t\tAlpha removed") if is_alpha_present && remove_alpha
```
janpio: Why is this information relevant?

supply/lib/supply/setup.rb
janpio: I guess Google offers no param to just get the original files?

supply/lib/supply/setup.rb
scorpion35: Was different, but then I am overwriting the downloaded file with properly formatted file. So they are same now. Updated the comment - https://github.com/scorpion35/fastlane/commit/dba543192721140a4dfbc642ae2ba73f8388a9c5

supply/lib/supply/setup.rb
scorpion35: Had that during my testing, and would like to see the path when the feature is still new. If you don't think this is relevant, we can remove it. No issues from my side!

supply/lib/supply/setup.rb
scorpion35: This formats the output nicely. Can remove if you don't this it's needed!

supply/lib/supply/setup.rb
scorpion35: I think so. Because the file are in `webp` format, we can request any size. Figured if I use a big dimension, they would only return what they have.

supply/lib/supply.rb
```diff
@@ -1,3 +1,5 @@
+require 'mini_magick'
```
joshdholtz: I don't think we can safely add `mini_magick` as a dependency inside of one of the core `supply` functions. This will require everybody who currently runs `supply` to install imagemagick which we probably don't want as a requirement for _fastlane_ 😇

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
           UI.error(ex.to_s)
           UI.error("Error downloading '#{image_type}' for #{listing.language}...")
         end
+      end # allowed_imagetypes.each do |image_type|
+
+      3.times do
+        UI.message("")
       end
-    end
+    end # def download_images(listing)
```
joshdholtz:
```suggestion
    end
```

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+              image_counter += 1
+            end # urls.each do |url|
+          else
+            UI.message("\tNo images found, nothing to download.")
+          end # if(urls.length > 0)
         rescue => ex
           UI.error(ex.to_s)
           UI.error("Error downloading '#{image_type}' for #{listing.language}...")
         end
+      end # allowed_imagetypes.each do |image_type|
```
joshdholtz:
```suggestion
      end
```

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+              image.format(image_format) # Properly formatting the final image. Sometimes "jpg" or "webp" or "png" formats are returned by Google.
+              image.write(path) # Properly formatted file is written to disk
+              UI.message("\tDownloaded  - #{path}")
+              UI.message("\t\tAlpha removed") if is_alpha_present && remove_alpha
+
+              image_counter += 1
+            end # urls.each do |url|
+          else
+            UI.message("\tNo images found, nothing to download.")
+          end # if(urls.length > 0)
```
joshdholtz:
```suggestion
          end
```

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+              is_alpha_present = image_details.keys.any? { |k| k == :Alpha }
+              image.alpha('remove') if is_alpha_present && remove_alpha
+
+              image.format(image_format) # Properly formatting the final image. Sometimes "jpg" or "webp" or "png" formats are returned by Google.
+              image.write(path) # Properly formatted file is written to disk
+              UI.message("\tDownloaded  - #{path}")
+              UI.message("\t\tAlpha removed") if is_alpha_present && remove_alpha
+
+              image_counter += 1
+            end # urls.each do |url|
```
joshdholtz:
```suggestion
            end 
```

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+      allowed_imagetypes = [Supply::IMAGES_TYPES, Supply::SCREENSHOT_TYPES].flatten
 
+      allowed_imagetypes.each do |image_type|
         begin
           UI.message("Downloading `#{image_type}` for #{listing.language}...")
 
-          url = client.fetch_images(image_type: image_type, language: listing.language).last
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next unless urls
```
joshdholtz:
```suggestion
          next if urls.nil? || urls.empty?
```
We can do this and remove the extra if statement below

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
...
+
+          if urls.length > 0
+            image_counter = 1
```
joshdholtz: Why doesn't this start at `0`? 🤔

supply/lib/supply/setup.rb
```diff
@@ -40,34 +39,58 @@ def store_metadata(listing)
       end
     end
 
-    def download_images(listing)
+    def download_images(listing, image_format = "png", remove_alpha = "false")
```
joshdholtz: Any reasons why `remove_alpha` is a string and not a boolean here?

joshdholtz (dismissed):
Thanks for working on adding this feature! Left some minor comments about some of the code and one major comment about requiring the user to install `imagemagick` which we don't want to do 😬

supply/lib/supply.rb
scorpion35: Yes, was wondering about this. I had this change tested on my colleague's computer (who doesn't have `mini_magick` installed, and it errored out. I am not sure how to make this work. Please advise :)

supply/lib/supply/setup.rb
scorpion35: `image_counter` is used to prefix the final images. Please see this screenshot - https://user-images.githubusercontent.com/18679225/53509689-51930c00-3a8a-11e9-8b81-4d4a7c3bbe11.png. Let me know if you'd like to have a different naming convention, please!

supply/lib/supply/setup.rb
scorpion35: Good catch! My bad, was using `remove_alpha` as a `boolean` (https://github.com/fastlane/fastlane/pull/14322/files#diff-674b785c2e89b32d433060b131c0cf28R72), but passing it as a string. Fixed it here - https://github.com/fastlane/fastlane/pull/14322/commits/fd93d52e0d23e70b304525b6bd27e0c87ed6a6cd

supply/lib/supply/setup.rb
joshdholtz: Let's remove this these empty messages! We usually prefer to not add empty lines in the output 😇

supply/lib/supply/setup.rb
joshdholtz: Can we maybe just add a comment above this line on why this number is used?

supply/lib/supply/setup.rb
```diff
@@ -40,33 +39,57 @@ def store_metadata(listing)
...
+      allowed_imagetypes.each do |image_type|
         begin
           UI.message("Downloading `#{image_type}` for #{listing.language}...")
 
-          url = client.fetch_images(image_type: image_type, language: listing.language).last
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next if urls.nil? || urls.empty?
+
+          if urls.length > 0
```
joshdholtz: We can remove this line now too

supply/lib/supply.rb
joshdholtz: Is the removal of alpha 100% needed? 🤔 We can remove this if its not, right?

joshdholtz (commented):
Responded to a few more things 😊

supply/lib/supply.rb
janpio: We already have `mini_magick` in the dependencies: https://github.com/fastlane/fastlane/blob/f47f5268f899c2994b253601f1eef401b0c36615/fastlane.gemspec#L83 It's also used in frameit for example.

janpio:
So @scorpion35, before I go in actually testing this: What about this alpha/minimagick/ImageMagick stuff? Can we remove this to simplify here? What does this do exactly?

scorpion35 (commented):
> So @scorpion35, before I go in actually testing this: What about this alpha/minimagick/ImageMagick stuff? Can we remove this to simplify here? What does this do exactly?

Thanks for taking a look janpio :) These are my observations/notes - 
- Google Play randomly sends back `png` or `jpg` or `webp` format. (Not sure how to make it send only `png` or `jpg`)
- If it's `png` or `jpg`, we are safe. But we still need to figure out which type is sent back, so we can suffix it with the proper file extension (`.png` or `.jpg`).
  - We can simply suffix both types with `.png` (i.e. without any smart checks). The user would be able to simply double click and open the file. But, it would be 'technically' wrong if the returned data if of `jpg` type, and has a file extension of `png`.
- Now, the **main** issue is when they return the data in `webp` format. Adding `.png` or `.jpg` corrupts the written file, and cannot be opened. I think only way to see `.webp` file is to drag-drop it into Chrome browser.
  - I originally didn't had the dependency on `minimagick`, but found about the `webp` format when I was testing on my computer. Not sure what to do, I start using `minimagick` library :)
  - Also noticed returned data is mostly `png`/`jpg`, but `webp` is also present.

Because I wasn't sure when they return a `webp` formatted data, I am using https://github.com/minimagick/minimagick to properly convert the returned data into `png` and safely writing it to a file, ensuring consitency as well. Yes, I also would like to remove the external dependancy as well, but not sure how to use the same library as used in `frameit`. I think `frameit` users should have to install `minimagick` as an external dependency as well, but I could be totally wrong (hadn't used it so far).

`alpha` stuff is optional for now. I think it only matters when uploading to Google Play, not when downloading it. You can remove this if you don't think it's needed, but I would leave it as it might come in handy when uploading images. Google Play suggets to remove alpha when uploading - https://support.google.com/googleplay/android-developer/answer/1078870?hl=en -> **Screenshots**

If you have a way to properly write correct file type to disk, I am open to using your method :)

Please let me know if you need more information!

supply/lib/supply/setup.rb
```diff
@@ -1,6 +1,6 @@
 module Supply
   class Setup
-    def perform_download
+    def perform_download(image_format = "png", remove_alpha = false)
```
janpio: Why these params here? 
Unless I missed something, these are not really used but the default values are applied?

supply/lib/supply/setup.rb
```diff
@@ -12,8 +12,7 @@ def perform_download
 
       client.listings.each do |listing|
         store_metadata(listing)
-        create_screenshots_folder(listing)
```
janpio: As this method is not called any more, the method itself can also be removed.

janpio: What happens if there are no screenshots available at Google Play - are the folders created correctly anyway now with those changes?

scorpion35: https://github.com/fastlane/fastlane/pull/14322/commits/379316c2ce3f1663a1ba626b1682d270760e005f

supply/lib/supply/setup.rb
```diff
@@ -40,28 +39,44 @@ def store_metadata(listing)
...
+
+            p = Pathname.new(path)
+            FileUtils.mkdir_p(p.dirname.to_s)
+
+            File.write(path, Net::HTTP.get(URI.parse(url))) # Write to tmp `path`
+
+            image = MiniMagick::Image.open(path)
+            image_details = JSON.parse(image.details.to_json, symbolize_names: true)
+
+            is_alpha_present = image_details.keys.any? { |k| k == :Alpha }
```
janpio: clarify code:

```suggestion
            # remove alpha
            is_alpha_present = image_details.keys.any? { |k| k == :Alpha }
```

supply/lib/supply/setup.rb
```diff
@@ -40,28 +39,44 @@ def store_metadata(listing)
...
+            image.alpha('remove') if is_alpha_present && remove_alpha
+
+            image.format(image_format) # Properly formatting the final image. Sometimes "jpg" or "webp" or "png" formats are returned by Google.
```
janpio: clarify code:
```suggestion
            # convert image
            image.format(image_format) # Properly formatting the final image. Sometimes "jpg" or "webp" or "png" formats are returned by Google.
```

supply/lib/supply/setup.rb
```diff
@@ -40,28 +39,44 @@ def store_metadata(listing)
...
+            image.write(path) # Properly formatted file is written to disk
+            UI.message("\tDownloaded  - #{path}")
```
janpio: clarify code:
```suggestion

            UI.message("\tDownloaded  - #{path}")
```

supply/lib/supply/setup.rb
scorpion35: Yes :)

> What happens if there are no screenshots available at Google Play - are the folders created correctly anyway now with those changes?

Yes :) This line ensures folders are created before hand - https://github.com/fastlane/fastlane/blob/379316c2ce3f1663a1ba626b1682d270760e005f/supply/lib/supply/setup.rb#L63

supply/lib/supply/setup.rb
scorpion35: Is it some trick question? :D What changed here?

supply/lib/supply/setup.rb
```diff
@@ -40,28 +39,44 @@ def store_metadata(listing)
...
 
-          url = client.fetch_images(image_type: image_type, language: listing.language).last
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next if urls.nil? || urls.empty?
+
+          image_counter = 1 # Used to prefix the downloaded files, so order is preserved.
+          urls.each do |url|
+            url = "#{url}=w6000-h6000" # using huge dimentions intentionally so Google returns the images in their original size they were uploaded
+            path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, image_type, "#{image_counter}_#{listing.language}.#{image_format}")
```
janpio: Can we make these two `path` values a bit clearer that one is for screenshots, and the other for `%w(featureGraphic icon promoGraphic tvBanner)`? Would be clearer if this was a normal `if/elseif` with a comment to clarify.

supply/lib/supply/setup.rb
janpio: Added an empty line so the logic is separated from the UI output ;)

supply/lib/supply/setup.rb
janpio: Doesn't this get skipped when there are no screenshots via https://github.com/fastlane/fastlane/pull/14322/files#diff-674b785c2e89b32d433060b131c0cf28R54?

janpio:
Thanks for the explanation @scorpion35 regarding the file format and alpha, this is now much clearer to me - those are two separate tasks here.

Does the alpha removal actually make sense for _downloading_ the screenshots? I don't really get how this could be relevant if we only should upload screenshots without alpha anyway, right?

I could not find a screenshot delivered as webp for any of my own apps. Can you share a URL where this happens? (Nothing in the docs of the underlying API as well of course: https://developers.google.com/android-publisher/api-ref/edits/images/list#response)

supply/lib/supply/setup.rb
scorpion35: Yes! If there are no images returned (for example, `tvBanner` in most cases) doesn't have images. So there's no need to create its folder. 

I think if the user uploads images for `tvBanner`, then next time someone tries to download the images, `tvBanner` will be created as well!

supply/lib/supply/setup.rb
scorpion35: Made the chagne - https://github.com/fastlane/fastlane/pull/14322/commits/7a03d2ef3becf665d52bd14b4b301e48c6c30c71. 

Please let me know if you'd like this in some other way :)

supply/lib/supply/setup.rb
scorpion35: I use `Spaceship` directly most of the time (don't have much experience with fastfiles). Having these optional params for users like myself would be extremely useful, as I can simply change them. 

And, being default params, they default way will be applied for users not using them :)

janpio:
I found some "documentation" on the image links:
https://sites.google.com/site/picasaresources/Home/Picasa-FAQ/google-photos-1/how-to/how-to-get-a-direct-link-to-an-image

- Adding `=s0` to the URL (instead of what is currently behind `=`) might give you the original dimensions.
- There is also something about file endings, but this doesn't seem to affect the actual image format being returned.

https://developers.google.com/photos/library/guides/access-media-items#image-base-urls is some official documentation, but only for parameters that are not relevant for us.

scorpion35:
> Thanks for the explanation @scorpion35 regarding the file format and alpha, this is now much clearer to me - those are two separate tasks here.

> Does the alpha removal actually make sense for _downloading_ the screenshots? I don't really get how this could be relevant if we only should upload screenshots without alpha anyway, right?

You are right! It doesn't make sense when downloading. Removed the lines related - https://github.com/fastlane/fastlane/pull/14322/commits/224ba805933363ce10260f8259394fabf88d30d2. Can we keep the optional parameter as is? I would like to override this feature from my scripts as it allows me to store clean images in our Github repos.

 
> I could not find a screenshot delivered as webp for any of my own apps. Can you share a URL where this happens? (Nothing in the docs of the underlying API as well of course: https://developers.google.com/android-publisher/api-ref/edits/images/list#response)

This is very tricky. It did bite me a couple of times when testing on my computer, but don't have good repro steps.

janpio:
Seems I found a way to _force_ webp:

- Normal image: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC
- Original size: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC=s0
- Original size and webp: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC=s0-rw
- Just webp: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC=rw

Can you confirm my observations?

scorpion35:
> Seems I found a way to _force_ webp:
> 
> * Normal image: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC
> * Original size: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC=s0
> * Original size and webp: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC=s0-rw
> * Just webp: https://lh3.googleusercontent.com/GTb4PMz4SRwuGIUMgaKqcXZuxLn17hYMP-iIYKlpDW2dEPTjjyeRWznsyKZULNfS1XDC=rw
> 
> Can you confirm my observations?

Love it! Just came across this article - https://www.howtogeek.com/325864/how-to-save-googles-webp-images-as-jpeg-or-png/. It seems simply removing `-rw` does the trick. I am going to test the other flags you specified above (`=s0`, etc.) and come back here with results. Looks promising, think we can remove the `minimagick` dependency!

janpio:
Oh, so you maybe got back image URLs from the API _with_ the `=rw` bit, thus making them webp? Then removing this part of the URL might indeed be enough to make sure we get a png/jpg back. (Although we still have the problem of the file extension then - but I am pretty sure there are simpler ways than imagemagick for that if we look for it)

supply/lib/supply/setup.rb
```diff
@@ -40,48 +39,50 @@ def store_metadata(listing)
...
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next if urls.nil? || urls.empty?
+
+          image_counter = 1 # Used to prefix the downloaded files, so order is preserved.
+          urls.each do |url|
+            url = "#{url}=w6000-h6000" # using huge dimentions intentionally so Google returns the images in their original size they were uploaded
+            
+            if IMAGES_TYPES.include?(image_type) # IMAGE_TYPES are stored in locale/images location
+              path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, "#{image_type}.#{image_format}")
```
janpio: rubocop doesn't like the space at the end:
```suggestion
              path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, "#{image_type}.#{image_format}")
```

janpio:
And here we go with params that convert to .jpg and .png:

> rj — forces the resulting image to be JPG
> rp — forces the resulting image to be PNG
> rw — forces the resulting image to be WebP

https://github.com/albertcht/python-gcs-image#image-format

scorpion35:
> Oh, so you maybe got back image URLs from the API _with_ the `=rw` bit, thus making them webp? Then removing this part of the URL might indeed be enough to make sure we get a png/jpg back. (Although we still have the problem of the file extension then - but I am pretty sure there are simpler ways than imagemagick for that if we look for it)

Yes, exactly! :) 

One way I found is to read the file as a binary and look for either `png` or  `jpeg` in them (they are usually at the very beginning of the string). We can use this to determine file extension, and write the file accordingly. The problem then would be if Google somehow decides to return another type of new extension, we will be in trouble :D 

Are you ok with just having support for `png` or `jpg` for now? Maybe we can look at other file types when Google changes them.

janpio:
I am absolutely fine with that, as you can actually only upload...

> JPEG or 24-bit PNG (no alpha)

Source: https://support.google.com/googleplay/android-developer/answer/1078870

`bool: nw  ==> NoWebp` might be another nice option to use _if_ you actually got a webp back for URLs without `rw` before.

> One way I found is to read the file as a binary and look for either png or jpeg in them (they are usually at the very beginning of the string). We can use this to determine file extension, and write the file accordingly. 

Fastlane actually uses another gem that should be able to do this:
https://github.com/fastlane/fastlane/blob/master/fastlane.gemspec#L79
https://github.com/sdsykes/fastimage#examples

That should solve your problem in a nice way.

> The problem then would be if Google somehow decides to return another type of new extension, we will be in trouble :D

That's what this issue tracker is for 🙈

supply/lib/supply/setup.rb
```diff
@@ -40,48 +39,50 @@ def store_metadata(listing)
...
 
-          url = client.fetch_images(image_type: image_type, language: listing.language).last
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next if urls.nil? || urls.empty?
+
+          image_counter = 1 # Used to prefix the downloaded files, so order is preserved.
+          urls.each do |url|
+            url = "#{url}=w6000-h6000" # using huge dimentions intentionally so Google returns the images in their original size they were uploaded
+
```
janpio:
```suggestion

```

scorpion35:
This should be it - https://github.com/fastlane/fastlane/pull/14322/commits/f47fe92770ffae3235b2a1a27c5f6c5e8589e54c :) 

No more arbitrary dimensions, and no more `imagemagick` required. Didn't knew about `FastImage`. I have uninstall `imagemagick` from my computer and tested the screenshots download. New changes chooses the correct file extension as well! Please let me know if you have any feedback!

scorpion35:
**bundle exec rspec**
```
Finished in 4 minutes 40.5 seconds (files took 4.38 seconds to load)
5563 examples, 0 failures
```

**bundle exec rubocop -a** - https://github.com/fastlane/fastlane/pull/14322/commits/4c57e9af66ed588e07a3eec2e0267992b88cad1e

supply/lib/supply/setup.rb
```diff
@@ -55,23 +55,22 @@ def download_images(listing, image_format = "png", remove_alpha = false)
 
           image_counter = 1 # Used to prefix the downloaded files, so order is preserved.
           urls.each do |url|
-            url = "#{url}=w6000-h6000" # using huge dimentions intentionally so Google returns the images in their original size they were uploaded
+            url = url.gsub("-rw", "") # Remove "-rw" (if present) so the url will become either 'jpg' or 'png' (https://github.com/fastlane/fastlane/pull/14322#issuecomment-473017130)
+            url = "#{url}=s0" # '=s0' param ensures full image size is returned (https://github.com/fastlane/fastlane/pull/14322#issuecomment-473012462)
```
janpio: Are we sure we can just add `=` here? If the URL already has a `=` in it, we have to add `-s0` instead :/

scorpion35: Good point! This should fix it - https://github.com/fastlane/fastlane/pull/14322/commits/39fca173c86ba56b155b1acaa5fe7fc775771f41

supply/lib/supply/setup.rb
janpio: Yes, for the images that's correct - but not for the screenshots. There the folder always has to exist (at least that is how I red the old code).

supply/lib/supply/setup.rb
```diff
@@ -43,45 +42,46 @@ def store_metadata(listing)
...
 
-          url = client.fetch_images(image_type: image_type, language: listing.language).last
-          next unless url
+          urls = client.fetch_images(image_type: image_type, language: listing.language)
+          next if urls.nil? || urls.empty?
+
+          image_counter = 1 # Used to prefix the downloaded files, so order is preserved.
+          urls.each do |url|
+            url = url.gsub(url.match("=.*").to_s, "") # Remove everything after '=' (if present)
+            url = url.gsub("-rw", "") # Remove "-rw" (if present) so the url will become either 'jpg' or 'png' (https://github.com/fastlane/fastlane/pull/14322#issuecomment-473017130)
```
janpio: We don't need this any more if we removed everything after and including `=`, right?

supply/lib/supply/setup.rb
```diff
@@ -43,45 +42,46 @@ def store_metadata(listing)
...
+            url = "#{url}=s0" # '=s0' param ensures full image size is returned (https://github.com/fastlane/fastlane/pull/14322#issuecomment-473012462)
```
janpio: Can we add a verbose log output here that mentions that we messed with `url` here (if we actually did) so if Google every decides to break these URLs we have a way for users to actually find out that our messing with the `url` is the culprit here?

supply/lib/supply.rb
```diff
@@ -1,12 +1,12 @@
-require 'json'
```
janpio: Uh, I don't think we want to change `'` to `"` here and in all the other places.

supply/lib/supply/setup.rb
scorpion35: https://github.com/fastlane/fastlane/pull/14322/commits/f720f66544def11d165bd8d02fbef66a8f73a429

supply/lib/supply/setup.rb
scorpion35: Added some verbose logging. Let me know if you'd like to change it to something else, please!

https://github.com/fastlane/fastlane/pull/14322/commits/f720f66544def11d165bd8d02fbef66a8f73a429

janpio: logging of the removal should probably only happen if something was removed, and also include the before/after of the url. that we do add `=s0` doesn't have to be logged I guess if we see the end result.

supply/lib/supply.rb
scorpion35: It's `rufo` extension in my vs code. Reverted the changes - https://github.com/fastlane/fastlane/pull/14322/commits/cc4c4efe2d1f40950ab998d00c31206c3b4b136f

supply/lib/supply/setup.rb
scorpion35: Made these changes - https://github.com/fastlane/fastlane/pull/14322/commits/5e8c57f62c3f87c71002cc0b7a49a93351e20d7c

scorpion35:
Wondering what you think about this line - 
https://github.com/fastlane/fastlane/blob/978448df5da0e7fa221293617d508529ba9fdb01/supply/lib/supply/setup.rb#L59

I think it should be fine, only came across these types of url' - https://lh3.ggpht.com/xCsL4NrBzeH676lsyEmP-8Z5Hpw3rXC38R2-dUuXTc7jedwipLm3iMjyJ_Y0vFKHZQI**=w720-h310-rw**, for example.

Do you suggest any changes to `if !url_params.nil? && url_params.length == 1`? Let me know please.

supply/lib/supply/setup.rb
```diff
@@ -43,45 +42,52 @@ def store_metadata(listing)
...
+              path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, image_type.to_s)
+            else # All other screenshot types goes under their respective folders.
+              path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, image_type, "#{image_counter}_#{listing.language}")
+            end
+
+            p = Pathname.new(path)
+            FileUtils.mkdir_p(p.dirname.to_s)
+
+            path = "#{path}.#{FastImage.type(url)}"
+            File.write(path, Net::HTTP.get(URI.parse(url)))
```
janpio: Currently this doesn't work on Windows at all, just produces black images. With this change it works fine:

```suggestion
            File.binwrite(path, Net::HTTP.get(URI.parse(url)))
```

janpio:
Finally had the time to manually test your branch:

- It did _not_ create folders for screenshot classes that had no screenshot uploaded (sevenInchScreenshots, tenInchScreenshots, tvScreenshots, wearScreenshots in my case), the old version did. This should be fixed.
- It did download a `featureGraphic.png` and the 5 `phoneScreenshots` that exist for the app I tested with 🚀

scorpion35:
> Finally had the time to manually test your branch:
> 
> * It did _not_ create folders for screenshot classes that had no screenshot uploaded (sevenInchScreenshots, tenInchScreenshots, tvScreenshots, wearScreenshots in my case), the old version did. This should be fixed.
> * It did download a `featureGraphic.png` and the 5 `phoneScreenshots` that exist for the app I tested with 🚀

Thank you very much! I will look into this soon.

scorpion35:
> Finally had the time to manually test your branch:
> 
> * It did _not_ create folders for screenshot classes that had no screenshot uploaded (sevenInchScreenshots, tenInchScreenshots, tvScreenshots, wearScreenshots in my case), the old version did. This should be fixed.
> * It did download a `featureGraphic.png` and the 5 `phoneScreenshots` that exist for the app I tested with 🚀

Submitted the fix - https://github.com/fastlane/fastlane/pull/14322/commits/42e3768b7d59364229c9022add8a0b0809e0b8a1. It's creating folders even if the no images are present!

supply/lib/supply/setup.rb
scorpion35: https://github.com/fastlane/fastlane/pull/14322/commits/42e3768b7d59364229c9022add8a0b0809e0b8a1

supply/lib/supply/setup.rb
```diff
@@ -48,32 +48,32 @@ def download_images(listing)
 
       allowed_imagetypes.each do |image_type|
         begin
+          image_counter = 1 # Used to prefix the downloaded files, so order is preserved.
+          if IMAGES_TYPES.include?(image_type) # IMAGE_TYPES are stored in locale/images location
+            path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, image_type.to_s)
+          else # All other screenshot types go under their respective folders.
+            path = File.join(metadata_path, listing.language, IMAGES_FOLDER_NAME, image_type, "#{image_counter}_#{listing.language}")
```
janpio: this won't work outside of the `urls.each` loop though, right?

scorpion35: I moved it to the beginning of `allowed_imagetypes` loop, so it creates folders for all image types. If `urls` is `null`, nothing is downloaded (but the folder is still created). 

Hope this is what you are looking for. Let me know if you are looking for something else, please :)

janpio: `image_counter` in `path` will always be `1` now, as it is only incremented in line 82 (inside the loop).

I think you have to separate the folder and filename parts of the path - the folder can be defined outside, but the filename has to be inside of the loop.

scorpion35: Ah, I see what you mean :)

If there are no images to download, do we need to increment the counter? (And if there are images, it would be incremented.).`FileUtils.mkdir_p(p.dirname.to_s)` will create the parent directory.

janpio: > (And if there are images, it would be incremented.)

Sure?

The path is defined only once now, but we are saving many images. The variable will be incremented, but the path will not be updated. (Or did I miss something?)

scorpion35: ah I see, my bad 🤦‍♂️ fixing it now :)

scorpion35 (commented):
This should be it - https://github.com/fastlane/fastlane/pull/14322/commits/fc8ff1ee739eb4f24060e8abc41e5afa60adbe27

Tested it thoroughly this time :)

scorpion35:
Any update, janpio? :)

scorpion35:
Any update @janpio ? Let me know if you are expecting anything else from my side, please! 👌

janpio:
No, I will have to find the time to test the whole flow with multiple cases again.

janpio:
Output for an "empty" app:

```
λ bundle exec fastlane supply init --json_key "..." --package_name ...
 [✔] �
[18:44:03]: �  Downloading metadata, images, screenshots...
[18:44:05]: �  Downloading metadata (de-DE)
[18:44:05]: Writing to fastlane/metadata/android/de-DE/title.txt...
[18:44:05]: Writing to fastlane/metadata/android/de-DE/short_description.txt...
[18:44:05]: Writing to fastlane/metadata/android/de-DE/full_description.txt...
[18:44:05]: Writing to fastlane/metadata/android/de-DE/video.txt...
[18:44:05]: �️  Downloading images (de-DE)
[18:44:05]: Downloading `featureGraphic` for de-DE...
[18:44:05]: Downloading `icon` for de-DE...
[18:44:06]: Downloading `promoGraphic` for de-DE...
[18:44:06]: Downloading `tvBanner` for de-DE...
[18:44:06]: Downloading `phoneScreenshots` for de-DE...
[18:44:06]: Downloading `sevenInchScreenshots` for de-DE...
[18:44:07]: Downloading `tenInchScreenshots` for de-DE...
[18:44:07]: Downloading `tvScreenshots` for de-DE...
[18:44:07]: Downloading `wearScreenshots` for de-DE...
[18:44:08]: ✅  Successfully stored metadata in 'fastlane/metadata/android'
```

Created this file structure:

```
.../metadata
└───android
    └───de-DE
        │   full_description.txt
        │   short_description.txt
        │   title.txt
        │   video.txt
        │
        └───images
            ├───phoneScreenshots
            ├───sevenInchScreenshots
            ├───tenInchScreenshots
            ├───tvScreenshots
            └───wearScreenshots
```

---

Output for an app with actual metadata available:

```
λ bundle exec fastlane supply init --json_key "..." --package_name ...
 [✔] �
[18:49:07]: �  Downloading metadata, images, screenshots...
[18:49:10]: �  Downloading metadata (de-DE)
[18:49:10]: Writing to fastlane/metadata/android/de-DE/title.txt...
[18:49:10]: Writing to fastlane/metadata/android/de-DE/short_description.txt...
[18:49:10]: Writing to fastlane/metadata/android/de-DE/full_description.txt...
[18:49:10]: Writing to fastlane/metadata/android/de-DE/video.txt...
[18:49:10]: �️  Downloading images (de-DE)
[18:49:10]: Downloading `featureGraphic` for de-DE...
[18:49:12]:     Downloaded - fastlane/metadata/android/de-DE/images/featureGraphic
[18:49:12]: Downloading `icon` for de-DE...
[18:49:13]:     Downloaded - fastlane/metadata/android/de-DE/images/icon
[18:49:13]: Downloading `promoGraphic` for de-DE...
[18:49:13]: Downloading `tvBanner` for de-DE...
[18:49:14]: Downloading `phoneScreenshots` for de-DE...
[18:49:15]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:15]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:15]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:16]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:16]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:16]: Downloading `sevenInchScreenshots` for de-DE...
[18:49:17]: Downloading `tenInchScreenshots` for de-DE...
[18:49:17]: Downloading `tvScreenshots` for de-DE...
[18:49:17]: Downloading `wearScreenshots` for de-DE...
[18:49:19]: �  Downloading changelogs (de-DE, 3)
[18:49:19]: Writing to fastlane/metadata/android/de-DE/changelogs/3.txt...
[18:49:21]: ✅  Successfully stored metadata in 'fastlane/metadata/android'
```

And the created file structure:

```
.../metadata
└───android
    └───de-DE
        │   full_description.txt
        │   short_description.txt
        │   title.txt
        │   video.txt
        │
        ├───changelogs
        │       3.txt
        │
        └───images
            │   featureGraphic.png
            │   icon.png
            │
            ├───phoneScreenshots
            │       1_de-DE.png
            │       2_de-DE.png
            │       3_de-DE.png
            │       4_de-DE.png
            │       5_de-DE.png
            │
            ├───sevenInchScreenshots
            ├───tenInchScreenshots
            ├───tvScreenshots
            └───wearScreenshots
```

So that looks good!

---

Only thing that should be fixed before this is really ready to merge is this output:

```
[18:49:10]: Downloading `featureGraphic` for de-DE...
[18:49:12]:     Downloaded - fastlane/metadata/android/de-DE/images/featureGraphic
[18:49:12]: Downloading `icon` for de-DE...
[18:49:13]:     Downloaded - fastlane/metadata/android/de-DE/images/icon
...
[18:49:14]: Downloading `phoneScreenshots` for de-DE...
[18:49:15]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:15]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:15]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:16]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
[18:49:16]:     Downloaded - fastlane/metadata/android/de-DE/images/phoneScreenshots
```

For images it is missing the file extension, and for screenshot it is missing the actual file names and just outputting the folder name. 

I will suggest a trivial fix here as a code review that should fix that.

supply/lib/supply/setup.rb
```diff
@@ -43,45 +42,57 @@ def store_metadata(listing)
...
+
+            if IMAGES_TYPES.include?(image_type) # IMAGE_TYPES are stored in locale/images location
+              file_path = "#{path}.#{FastImage.type(url)}"
+            else # SCREENSHOT_TYPES go under their respective folders.
+              file_path = "#{path}/#{image_counter}_#{listing.language}.#{FastImage.type(url)}"
+            end
+
+            File.binwrite(file_path, Net::HTTP.get(URI.parse(url)))
+
+            UI.message("\tDownloaded - #{path}")
```
janpio:
```suggestion
            UI.message("\tDownloaded - #{file_path}")
```

janpio:
Did you ever get additional output for `--verbose`? My image URLs never seem to include any parameters.

scorpion35:
I test my changes using `spaceship` library, not sure how to test pass the `--verbose` flag. Can you tell me how to pass this flag, please?

(Sorry for the delay, installing new flooring took a lot of time. Back in business now!)

scorpion35:
@janpio From this code - https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb#L82, it seems `verbose` is printed as `debug`? Not sure if I am looking in the right place or not :)

Let me know if you need anything else from me. Any ETA on when this can be merged?

janpio:
Yes, it would be those "Removed params ('#{url_params}') from the URL" messages you added when it has to manipulate the URLs.

No ETA, I currently don't have the setup and time to review and merge any PRs.

scorpion35:
> Yes, it would be those "Removed params ('#{url_params}') from the URL" messages you added when it has to manipulate the URLs.
> 
> No ETA, I currently don't have the setup and time to review and merge any PRs. But I think you are good here and just have to wait for someone to come around and do that.
> 
> Update: One last rebase on `master` would be nice, so that CI can run again (Circle CI removed the image we were using before)

Merged latest from fastlane:master. All tests passed. Let me know if you need anything else, please!

scorpion35:
@joshdholtz @janpio Sorry for the ping! Can you merge this PR, please? That would be a big help for me because I don't have to maintain separate code. Everyone else would benefit from these changes as well! :)

ISDr:
Any ideas when this will be in place?

scorpion35:
I am in wondering the same thing too!

janpio:
No change on my side, not the setup and time to verify and test this. Hope this will change soon-ish, but no promises. Sorry.

ISDr:
Thanks for the update :)

janpio:
No need to close this @scorpion35 - sooner or later we or someone will get to it ;)

scorpion35:
My bad janpio, should've added a comment before closing. Apologies!

I kinda snuck this in with AndroidPublisher V3 migration 😄 - https://github.com/fastlane/fastlane/commit/ce8d0bdad21a7aada08999e09bb09742f9a8db2d#diff-78d66f4d98357e17f51c049ba9980b41R447. It seems joshholtz took a look and his fix is much more elegant than what I came up with :)

janpio:
Ahhhh, that of course is even better 🚀 Thanks!

PR #14346, opened 2019-03-05
[actions] extract shared code from setup_travis and setup_circle_ci actions

### LINKED ISSUE #14279: setup_ci
We have `setup_travis` and `setup_circle` which are doing almost the same thing, so we should probably just have a `setup_ci` and the other two can become aliases to that.

mollyIV:
### Checklist

### Motivation and Context

`setup_travis` and `setup_circle` are doing almost the same thing so in order to make maintenance and future code changes easier we shall extract and reuse the common logic. 

closes https://github.com/fastlane/fastlane/issues/14279

### Description

The initial idea was creating `setup_ci` lane however it turned out there are some differences in the logic behind `setup_travis` and `setup_circle_ci` actions. I decided to give a go to the composition instead of subclassing. In that way we are able to reuse some logic in different CIs setup implementation, i.e. `setup_jenkins` and still maintain high readability of the code. 

I started a draft pull request because it's a good starting point for a discussion and to challenge the implementation idea. Below you can find other ideas:

1. New action `setup_ci` which we would run from `setup_travis` run method and `setup_circle_ci` run method. 

2. `setup_ci` action where we would pass a parameter: `travis`, `circle_ci`, ...

```ruby
# setup_ci.rb

def self.run(params)
    unless should_run?(force: params[:force])
          UI.message("Currently not running on CI system, skipping ci setup")
          return
    end

    case params[:provider]
    when "travis"
      # run `travis` specific tasks
    when "circle_ci"
      # run `circle_ci` specific tasks
    else
      # error
    end
end
```

Please share your thoughts if the chosen solution is okay for you or if we shall implement it differently  🙏 

I tested my changes by running unit tests + adding a few new unit tests. 

🙇 🙇


janpio:
This looks perfectly fine to me.

The only difference between the both is the additional `setup_output_paths`, right? Does it make sense to just copy that over as well?

mollyIV:
That's correct @janpio - the only difference is `setup_output_paths`. Not sure we shall move it to `ci.rb` though. Maybe we can do that later when we will need it due to other ci migrations, like `setup_jenkins`?

janpio:
Do you want to merge this first and then adapt `setup_jenkins` and/or create the new `setup_ci` later or do it all now?

mollyIV:
I'd prefer to split this work into smaller pieces and merge this one first 👍 

Do we still need `setup_ci ` though? 🤔

janpio:
Of course, there is no general action to "setup fastlane for any CI provider". I can call `setup_circle` or `setup_travis` of course, but that just feels wrong when e.g. running on Azure Pipelines or Cirrus CI.

mollyIV:
Cool, so the next steps might be: 

1. Review and merge this pull request if all good
2. Create pull request related to `setup_jenkins` changes (we need to analyse what's different there and reuse the logic from `ci.rb` class)
3. Create pull request where we will introduce `setup_ci` lane with `provider` parameter + support for the current actions (`setup_travis`, `setup_circleci`, `setup_jenkins` - most probably aliases or calling `setup_ci` lane inside their appropriate `run` method)

👍 👍

joshdholtz (changes requested):
I really like this change! This makes things a ton cleaner here 🙌  Left a comment about moving this into the `helpers` directory and some questions about some tests

fastlane/spec/actions_specs/setup_circle_ci_spec.rb
```diff
@@ -21,65 +21,57 @@
...
-    describe "#setup_keychain" do
-      context "when MATCH_KEYCHAIN_NAME is set" do
-        it "skips the setup process" do
-          stub_const("ENV", { "MATCH_KEYCHAIN_NAME" => "anything" })
-          expect(Fastlane::UI).to receive(:message).with("Skipping Keychain setup as a keychain was already specified")
-          described_class.setup_keychain
-        end
+        expect(ENV["MATCH_KEYCHAIN_NAME"]).to eq(tmp_keychain_name)
+        expect(ENV["MATCH_KEYCHAIN_PASSWORD"]).to eq("")
+        expect(ENV["MATCH_READONLY"]).to eq("true")
```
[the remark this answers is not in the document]
mollyIV: I think we don't need to do that - the proof that it works is checking the ENV (we set all the ENV variables to an empty map at the beginning of the test `stub_const("ENV", {})`). If it hadn't worked, we would have not gotten the ENV variables set. That was the original unit test in `setup_travis_spec.rb` class and I was inspired by it 🙂 

```
expect(ENV["MATCH_KEYCHAIN_NAME"]).to eq(tmp_keychain_name)
expect(ENV["MATCH_KEYCHAIN_PASSWORD"]).to eq("")
expect(ENV["MATCH_READONLY"]).to eq("true")
```

WDYT @joshdholtz ? 🤔

fastlane/spec/actions_specs/setup_circle_ci_spec.rb
```diff
@@ -21,65 +21,57 @@
...
+            lock_when_sleeps: true,
+            password: ""
+          }
+        )
 
-        it "sets the MATCH_KEYCHAIN_NAME env var" do
-          described_class.setup_keychain
-          expect(ENV["MATCH_KEYCHAIN_NAME"]).to eql("fastlane_tmp_keychain")
-        end
+        stub_const("ENV", { "FL_SETUP_CIRCLECI_FORCE" => "true" })
```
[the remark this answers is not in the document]
mollyIV: I think the explanation is like the above one 🙂 Although I am open for the code suggestion 👍

mollyIV:
Hey @joshdholtz 👋 I have implemented the requested changes and answered to your comments, could you please have a 👀 ? Thanks! 🙇

fastlane/lib/fastlane/helper/ci_helper.rb
```diff
@@ -0,0 +1,33 @@
+module Fastlane
+  module Helper
+    class CI
+      def self.should_run?(force: false)
```
dantoml: This doesn't belong in `Helper::CI` IMO, or at least not with `should_run?` as the name

mollyIV: Hey @dantoml 👋 hmm, why is that? 🤔 Could you please suggest a new name for this method if in your opinion the current one is wrong? 🙏 Thanks!

dantoml: 👋, So with `CI.should_run?` feels like inverting the thing we want to actually do. We're not asking if a `CI` thing should run, we have actions that should only run in CI or if forced.

Personally, I'd keep this specific function in the actions, to avoid accidental unintended consequences of changing this method.

mollyIV: Keeping this specific function in the actions would be copy-paste all over the place - imagine making the change there one day. 

What do you think about adding `should_run_on_ci?(force: false)` to `actions_helper.rb` file:

```
# actions_helper.rb

module Fastlane
  module Actions
      def self.should_run_on_ci?(force: false)
        Helper.ci? || force
      end
  end
end
```
? 🤔

dantoml: The fastlane model doesn't really give us a nice way to add these types of filters generically, because ideally this would be some type of trait or action filter, but that's not the reality we have today.

Extracting a single `||` into a helper isn't really providing us any value IMO, and I don't think we can give it a name that is actually what we're asking.

joshdholtz: This helper is more than just `Helper.ci? || force`. It also has the `setup_keychain` method. But... I guess maybe this doesn't belong in helpers and should end up being a standalone action??? 🤔 

And then this `setup_ci_keychain` action is called by `setup_circle_ci` and `setup_travis_ci`? Thoughts?

fastlane/spec/actions_specs/setup_circle_ci_spec.rb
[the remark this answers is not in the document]
joshdholtz: Ahhhh, that makes sense to me then 👍

joshdholtz (commented):
Added some thoughts about the helper!

fastlane/lib/fastlane/helper/ci_helper.rb
mollyIV: Sounds good to me 👍 Does this mean we are going to expose `setup_ci_keychain` to fastlane user (if so I guess having `ci_helper.rb` would prevent that)? 🤔 

Also let's keep in mind the future improvements pointed out in [this comment](https://github.com/fastlane/fastlane/pull/14346#issuecomment-470657627). We might end up with having more small CI related actions - is that okay for us? 🤔

joshdholtz: @mollyIV Maybe we just 💣 this helper and create the new aciton with the name of `setup_ci` instead? 🤔 That way it can flex into doing other things in the future if needed?

@dantoml Thoughts? 😇

mollyIV: @joshdholtz `setup_ci` was our last step in this plan: https://github.com/fastlane/fastlane/pull/14346#issuecomment-470657627. 

However I guess you are right and why not introduce it right away? 🤔 

To briefly sum up in a code snippets that everyone'd one the same page:

```ruby
# setup_travis.rb file

def self.run(params)
    Fastlane::Actions.setup_ci(provider: "travis", force: params[:force])
end
```

```ruby
# setup_circle_ci.rb file

def self.run(params)
    Fastlane::Actions.setup_ci(provider: "circleci", force: params[:force])
end
```

```ruby
# setup_ci.rb file

def self.run(params)
    force = params[:force]
    provider = params[:provider]

    unless should_run?(force: force)

    case provider
    when 'travis'
      # Do travis specific stuff: setup_keychain
    when 'circleci'
      # Do circleci specific stuff: setup_keychain, setup_output_paths
    else
      # Shall we throw an exception here and fail the action? 
end
```

@joshdholtz @dantoml please share your thoughts 🙏

dantoml: If we really want to introduce another layer of abstraction to this (I'm still not sure this is worth DRYing), the most ruby like abstraction would be to have a `ci_setup` base class, that implements `setup_keychain` and `should_run`, with `setup_circle_ci` and `setup_travis` subclasses, that can then override `self.run` with

```ruby
def self.run(params)
  unless should_run?(force: params[:force])
    ...
  end

  setup_keychain(params)
  setup_output_paths
end
```

```ruby
def self.run(params)
  unless should_run?(force: params[:force])
    ...
  end

  setup_keychain(params)
end
```

Arguably the superclass run implementation could default to should_run... setup_keychain, but that would be overly abstract probably.

janpio: I like what @mollyIV posted above - simple, clear and obvious. (Only exception: Take the `setup_keychain` out of the `case` if it is valid for all cases)

mollyIV:
👋 @janpio @joshdholtz @dantoml 

I pushed the newest changes and introduced `setup_ci` action, could you please have a 👀 🙏 🙇

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -0,0 +1,121 @@
...
+        ENV["FL_BUILDLOG_PATH"] = (root + "buildlogs").to_s
+        ENV["SCAN_INCLUDE_SIMULATOR_LOGS"] = true.to_s
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Setup the keychain and match to work with CircleCI"
```
janpio: This should be generalized.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -0,0 +1,121 @@
...
+      def self.details
+        list = <<-LIST.markdown_list(true)
+            Creates a new temporary keychain for use with match
+            Switches match to `readonly` mode to not create new profiles/cert on CI
+            Sets up log and test result paths to be easily collectible
+          LIST
+
+        [
+          list,
+          "This action helps with CircleCI integration. Add this to the top of your Fastfile if you use CircleCI."
```
janpio: This should be generalized.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -0,0 +1,121 @@
...
+        ].join("\n")
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :force,
+                                       env_name: "FL_SETUP_CI_FORCE",
+                                       description: "Force setup, even if not executed by CircleCI",
```
janpio: This should be generalized.

fastlane/lib/fastlane/actions/setup_ci.rb
mollyIV: Fixed 💪

fastlane/lib/fastlane/actions/setup_ci.rb
mollyIV: Fixed 💪

fastlane/lib/fastlane/actions/setup_ci.rb
mollyIV: Fixed 💪

joshdholtz (approved):
I like this direction this PR took 👍 Thanks for the another contribution ❤️

PR #14360, opened 2019-03-07
[gym] Fix provisioning profile detection for projects based on inherited xcconfigs

jmbrocal:
### Checklist

### Motivation and Context
Fix detection of provisioning profile mapping for projects based on inherited xcconfigs for resolve build settings.

It solves the provisioning profile mapping generated for projects with more than one target which uses inheritance between the configuration and the target xcconfigs for resolve the build settings.

[Here](https://github.com/jmbrocal/ProjectExampleWithInheritedXcconfig) you can find an example project which the current implementation fails to detect the provisioning profile mapping.

### Description
The change consists in passing the target of the build configuration for resolve the PRODUCT_BUNDLE_IDENTIFIER, PROVISIONING_PROFILE_SPECIFIER and PROVISIONING_PROFILE with a complete recursion between project and targets.

I added an unit test covering the changes.


jmbrocal:
> I signed it!

0xpablo (approved):

jmbrocal:
Hi!

It's been a few days since I opened this PR and I have not had comments. The repo has activity recently so I am worried if my PR is wrong or it does not follow the rules correctly.

We need this changes for a project that I am working on because reduce the number of xcconfigs from 44 to 16 files.

How can I unlock this PR?

0xpablo (approved):

0xpablo:
@jmbrocal Sorry, I don't have permissions to move this forward. Maybe @janpio can help you.

janpio:
Did I get that right that your "only" real change is adding the `target` param to some method calls? Can you talk me through what exactly this does (in terms a non iOS developer might understand) exactly?

The change looks good and safe enough, but as it is in `gym` which is one of the most used actions I of course want to make sure we don't break anything.

jmbrocal:
> Did I get that right that your "only" real change is adding the target param to some method calls?

Yes, concretely the `target` param is passed to `#resolve_build_setting(key, root_target = nil) ⇒ String` method of the `XCBuildConfiguration` object. Passing this param the function will does a complete recursion between the project and the target. 

You can read the documentation of `resolve_build_setting` method [here](https://www.rubydoc.info/gems/xcodeproj/Xcodeproj/Project/Object/XCBuildConfiguration#resolve_build_setting-instance_method)

> Can you talk me through what exactly this does (in terms a non iOS developer might understand) exactly?


Sure, It allows to `resolve_build_setting` method to resolves the value of a build setting resolving first the variable assigned to it.

An example is

For the debug configuration we have this
```
// Debug.xcconfig
PRODUCTION_BUNDLE_IDENTIFIER = $(MY_VARIABLE)
```
For the target configuration we have this
```
// AppTarget.xcconfig
MY_VARIABLE = "A value"
```
If we do not pass the `target` param to `resolve_build_setting` method, it cannot resolve the value of 
the build setting `PRODUCTION_BUNDLE_IDENTIFIER`, because the value of that variable is assigned in other level.

janpio:
What does "to resolve complete recursion between project and targets" actually mean? Does this change the values return? Does it improve the values returned?

jmbrocal:
Improve the value returned because it will work like Xcode does.

Without the `target` param the resolution of a build setting does not consider the `target` so the resolved value is incomplete.

Adding the `target` param, we are resolving the build setting like Xcode does so `gym` will be more precise.

jmbrocal:
Hi!

Can I help in any way to unlock this PR? I really need this fix but I don't want to fork the repo.

I can provide more tests, other examples or a better explanation about the changes for unlocking the PR.

Thanks!!

joshdholtz (approved):
This looks good! Thanks for adding this 😊❤️

PR #11834, opened 2018-02-13
Improve SnapshotHelper.swift to support snapshotting for multiple apps.

genkernel:
Support multiple apps testing flows with native XCUIApplication(bundleIdentifier:).

### Checklist
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)

### Motivation and Context
Subsequent call to XCUIApplication() would rise an exception if UITest was previously setup with XCUIApplication(bundleIdentifier:).
Replacing XCUIApplication() with Snapshot.app is safe because 'app' variable was previous initialized via setupSnapshot() function invocation.
Reusing an existing 'Snapshot.app' variable will enable multiple apps testing by leveraging  XCUIApplication(bundleIdentifier:) initialization.

### Description
Support multiple apps testing flows with native XCUIApplication(bundleIdentifier:).

PR #14367, opened 2019-03-08
[snapshot] Improve SnapshotHelper.swift to support snapshotting for multiple apps.

genkernel:
#11834 duplicate.

### Checklist

### Motivation and Context
Subsequent call to `XCUIApplication()` would rise an exception if UITest was previously setup with `XCUIApplication(bundleIdentifier:)`.
Replacing `XCUIApplication()` with `Snapshot.app` is safe because `app` variable was previously initialized via `setupSnapshot()` function invocation.
Reusing an existing `Snapshot.app` variable will enable multiple apps testing by leveraging `XCUIApplication(bundleIdentifier:)` from user tests.

### Description
Support multiple apps testing flows with native XCUIApplication(bundleIdentifier:).

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -156,7 +156,7 @@ open class Snapshot: NSObject {
         sleep(1) // Waiting for the animation to be finished (kind of)
 
         #if os(OSX)
-            XCUIApplication().typeKey(XCUIKeyboardKeySecondaryFn, modifierFlags: [])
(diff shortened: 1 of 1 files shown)


joshdholtz:
@genkernel Heyyy 👋  Good catch here! We definitely don’t want to keep making a new `XCUIApplication` instance. Your change didn’t fully work because `Snapshot.app` is optional so I had to add some `guards` around things and I also bumped the version up so I added some commits for those 💪

janpio (approved):
Man, Swift is a strange language...

but looks like this should work.

joshdholtz (approved):

PR #14387, opened 2019-03-12
[spaceship] Handle "Apple ID is locked" and other authentication errors better

max-ott:
Apple id returning different status and messages when the Apple ID is not "alright". We should better handle the cases, so spaceship does not break.

### Checklist

# Case: wrong password
```# Logfile created on 2019-03-12 17:52:12 +0100 by logger.rb/61378
INFO  [17:52:12]: >> GET https://olympus.itunes.apple.com/v1/session: [undefined body] 
DEBUG [17:52:13]: << GET https://olympus.itunes.apple.com/v1/session: 401 Unauthenticated

Request ID: LTG6ZX6EN55MAMQMQAYAKJN6.0.0

WARN  [17:52:13]: Auth lost
INFO  [17:52:13]: >> POST https://idmsa.apple.com/appleauth/auth/signin: {"accountName":"wrongpassword@acmecorp.com","password":"***","rememberMe":true} 
DEBUG [17:52:14]: << POST https://idmsa.apple.com/appleauth/auth/signin: 401 {"serviceErrors"=>[{"code"=>"-20101", "message"=>"Your Apple ID or password was incorrect."}]}
WARN  [17:52:14]: Auth lost
```

# Case: account locked by Apple
```# Logfile created on 2019-03-12 17:52:21 +0100 by logger.rb/61378
INFO  [17:52:21]: >> POST https://idmsa.apple.com/appleauth/auth/signin: {"accountName":"locked@acmecorp.com","password":"***","rememberMe":true} 
DEBUG [17:52:22]: << POST https://idmsa.apple.com/appleauth/auth/signin: 403 {"serviceErrors"=>[{"code"=>"-20209", "message"=>"This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."}]}
```

# Case: Apple ID messed up by Apple (currently in process with dev support)
```# Logfile created on 2019-03-12 17:53:58 +0100 by logger.rb/61378
INFO  [17:53:58]: >> POST https://idmsa.apple.com/appleauth/auth/signin: {"accountName":"messedup@acmecorp.com","password":"***","rememberMe":true} 
DEBUG [17:53:59]: << POST https://idmsa.apple.com/appleauth/auth/signin: 403 {"serviceErrors"=>[{"code"=>"-20751", "message"=>"Your Apple ID or password was incorrect."}]}
```

# Need help with:
It's probably possible to have the error handling in a nicer way. With my quick fix I get the following messages: 
```Spaceship::Tunes::Error ({"serviceErrors"=>[{"code"=>"-20751", "message"=>"Your Apple ID or password was incorrect."}]})
aa=432B7D3AE74EF5684B3E447669F46985; Domain=idmsa.apple.com; Path=/; Secure; HttpOnly, dslang=US-EN; Domain=apple.com; Path=/; Secure; HttpOnly, site=USA; Domain=apple.com; Path=/; Secure; HttpOnly
```

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -484,8 +484,6 @@ module Spaceship
       # Now we know if the login is successful or if we need to do 2 factor
 
       case response.status
-      when 403
-        raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
       when 200
         fetch_olympus_session
         return response
@@ -496,7 +494,11 @@ module Spaceship
         fetch_olympus_session
         return true
       else
-        if (response.body || "").include?('invalid="true"')
+        if (response.status == 401 && (response.body).include?("locked"))
+          raise InvalidUserCredentialsError.new, "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
+        elsif (response.status == 401)
+          raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
+        elsif (response.body || "").include?('invalid="true"')
           # User Credentials are wrong
           raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
         elsif response.status == 412 && AUTH_TYPES.include?(response.body["authType"])
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -487,6 +487,17 @@ module Spaceship
       when 200
         fetch_olympus_session
         return response
+      when 401
+        # Unauthorized
+        raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
+      when 403
+        # Forbidden
+        if response.body["serviceErrors"][0]["code"] == "-20209"
+          # Apple ID got locked by Apple for security reasons
+          raise InvalidUserCredentialsError.new, "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
+        else
+          raise InvalidUserCredentialsError.new, "Login worked, but Apple reported that something was wrong anyway. Historically this has happened when an Apple ID got lost on Apple's end. See https://github.com/fastlane/fastlane/issues/14398"
+        end
       when 409
         # 2 step/factor is enabled for this account, first handle that
         handle_two_step_or_factor(response)
@@ -494,11 +505,7 @@ module Spaceship
         fetch_olympus_session
         return true
       else
-        if (response.status == 401 && (response.body).include?("locked"))
-          raise InvalidUserCredentialsError.new, "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
-        elsif (response.status == 401)
-          raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
-        elsif (response.body || "").include?('invalid="true"')
+        if (response.body || "").include?('invalid="true"')
           # User Credentials are wrong
           raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
         elsif response.status == 412 && AUTH_TYPES.include?(response.body["authType"])
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -491,13 +491,17 @@ module Spaceship
         # Unauthorized
         raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
       when 403
+        error_string = "Login worked, but access was forbidden. Try logging in on https://appstoreconnect.apple.com and see if any issues are raised to you there."
         # Forbidden
-        if response.body["serviceErrors"][0]["code"] == "-20209"
-          # Apple ID got locked by Apple for security reasons
-          raise InvalidUserCredentialsError.new, "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
-        else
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/client.rb
```diff
@@ -484,8 +484,6 @@ def send_shared_login_request(user, password)
       # Now we know if the login is successful or if we need to do 2 factor
 
       case response.status
-      when 403
```
janpio: 403 not handled any more at all?

spaceship/lib/spaceship/client.rb
```diff
@@ -496,7 +494,11 @@ def send_shared_login_request(user, password)
         fetch_olympus_session
         return true
       else
-        if (response.body || "").include?('invalid="true"')
+        if (response.status == 401 && (response.body).include?("locked"))
+          raise InvalidUserCredentialsError.new, "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
+        elsif (response.status == 401)
+          raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
```
janpio: how is this different from the one below?

janpio:
> With my quick fix I get the following messages:

What does this refer to? The code you posted looks like part of a cookie!?

spaceship/lib/spaceship/client.rb
max-ott: https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses The "old" 403 case handles 401 as well, but is not as precise, because it can mean different things. If you try it out in the browser or check the spaceship logs, you will see that there are actual more helpful responses from Apple. I moved the logic down to lines 497+ to further split this into different and more helpful response messages

janpio: Well, important is what Apple returns: if they return a 403 in some cases, before that was explicitly handled with these lines. Now it is not handled by the status code any more.

spaceship/lib/spaceship/client.rb
```diff
@@ -484,11 +484,20 @@ def send_shared_login_request(user, password)
...
-        raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
       when 200
         fetch_olympus_session
         return response
+      when 401
+        # Unauthorized
+        raise InvalidUserCredentialsError.new, "Invalid username and password combination. Used '#{user}' as the username."
+      when 403
+        # Forbidden
+        if response.body["serviceErrors"][0]["code"] == "-20209"
```
getaaron: I like these smarter errors 💪

Better errors are something that make a great product and better user experience, so thanks for the PR!

A few suggestions

- For 403 we should use `UnauthorizedAccessError` instead of `InvalidUserCredentialsError`

- I think we should switch on the code since we know a few others from https://github.com/fastlane/fastlane/issues/14398

- If we really don't know what's wrong, we should provide the iForgot url and also the url to contact dev relations

for example (this is pseudo-ruby, I didn't try running it or check for typos)

```ruby
error_string = "Login worked, but access was forbidden. Try logging in on developer.apple.com and see if any issues are raised to you there."

case response.body["serviceErrors"][0]["code"]
when "-20209"
  error_string = "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
when "-20751"
  error_string = "Login worked, but something went wrong anyway. This has happened when an Apple ID got lost on Apple's end. See https://github.com/fastlane/fastlane/issues/14398; try resetting your account on https://iforgot.apple.com; or contact developer relations https://developer.apple.com/contact/"
end

raise UnauthorizedAccessError.new, error_string

```

getaaron (changes requested):

getaaron:
Looks like there’s a linter error

max-ott:
Not sure how to fix this, because it never got touched:

```Offenses:

spaceship/lib/spaceship/client.rb:400:5: C: Metrics/PerceivedComplexity: Perceived complexity for send_shared_login_request is too high. [20/18]
    def send_shared_login_request(user, password)
    ^^^

1049 files inspected, 1 offense detected```

File: spaceship/lib/spaceship/client.rb [hunk 4]
```diff
@@ -491,13 +491,17 @@ def send_shared_login_request(user, password)
...
       when 403
+        error_string = "Login worked, but access was forbidden. Try logging in on https://appstoreconnect.apple.com and see if any issues are raised to you there."
         # Forbidden
-        if response.body["serviceErrors"][0]["code"] == "-20209"
-          # Apple ID got locked by Apple for security reasons
-          raise InvalidUserCredentialsError.new, "This Apple ID has been locked for security reasons. Visit iForgot to reset your account (https://iforgot.apple.com)."
-        else
-          raise InvalidUserCredentialsError.new, "Login worked, but Apple reported that something was wrong anyway. Historically this has happened when an Apple ID got lost on Apple's end. See https://github.com/fastlane/fastlane/issues/14398"
+        case response.body["serviceErrors"][0]["code"]
+        when "-20209"
```
janpio: Can we add a comment with the full response body here (and for all other error cases) for reference? 
(Or as fixtures for a corresponding test?)

getaaron:
@max-ott _ Perceived complexity for send_shared_login_request is too high_ means (roughly) that there's too many conditionals in 1 method, and you can usually fix it by extracting some logic to a separate method

max-ott:
No idea how to do that. Any help appreciated.

janpio:
This are the docs of the cop: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Metrics/PerceivedComplexity
As I understand it, moving parts of the method into their own method would probably solve the problem.
Question is: Which parts, why and how.

rogerluan:
@max-ott @janpio @getaaron any idea if this PR (and its underlying issue) is still relevant? It seems like its branch got deleted, not sure if we want to keep this PR open and move push it forward 👀

Issue #12734, opened 2018-06-11
"Malformed version number" when iOS12 simulators are installed


Bersaelor:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Two weeks ago `fastlane` was working fine on my iMac. Last week I installed Xcode10-Beta with the iOS12 simulators.
Now when I run `fastlane screenshot` I get the following error:
`Malformed version number string com.apple.CoreSimulator.SimRuntime.iOS-12-0`

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>[konrad@iMac Sporttotal]$ fastlane screenshots --capture_output
[✔] 🚀 
DEBUG [2018-06-11 19:00:59.12]: Checking if there are any plugins that should be loaded...
INFO [2018-06-11 19:00:59.22]: ------------------------------
INFO [2018-06-11 19:00:59.22]: --- Step: default_platform ---
INFO [2018-06-11 19:00:59.22]: ------------------------------
INFO [2018-06-11 19:00:59.22]: Driving the lane 'ios screenshots' 🚀
INFO [2018-06-11 19:00:59.22]: ---------------------------------
INFO [2018-06-11 19:00:59.22]: --- Step: capture_screenshots ---
INFO [2018-06-11 19:00:59.22]: ---------------------------------
INFO [2018-06-11 19:00:59.72]: Successfully loaded '/Users/konrad/Documents/Sporttotal/fastlane/Snapfile' 📄
DEBUG [2018-06-11 19:00:59.72]: Fetching available simulator devices
INFO [2018-06-11 19:00:59.92]: $ xcodebuild -showBuildSettings -scheme Snapshot -project Sporttotal.xcodeproj
DEBUG [2018-06-11 19:01:01.53]: Fetching available simulator devices
DEBUG [2018-06-11 19:01:01.71]: Found 23 simulators.
DEBUG [2018-06-11 19:01:01.71]: Fetching available simulator devices
DEBUG [2018-06-11 19:01:01.89]: Checking that helper files contain SnapshotHelperVersion [1.10]
INFO [2018-06-11 19:01:01.93]: Building and running project - this might take some time...
[...]

PR #14388, opened 2019-03-12
[scan] Skip unavailable os_versions from simclt output

polac24:
* Stop processing simctl output immediately after first "Unavailable" OS as simctl aggregates them at the end of the output

### Checklist 🔑

### Motivation and Context
If your machine had already booted simulators that are not available in current's Xcode toolchain, these booted simulators get parsed as usually, but with some invalid os_version (like `com.apple.CoreSimulator.SimRuntime.iOS-12-1`). This breaks `scan` command which tries to transform "com.apple.CoreSimulator.SimRuntime.iOS-12-0" into `Gem::Version`. 

Fixes:
* https://github.com/fastlane/fastlane/issues/12734

### Description
Since `simctl` always returns all the unavailable OSes below available ones, once we recognize first unavailable OS (using `^-- Unavailable` regexp) we don't have to parse `simctl` output anymore.

Test spec adds a custom `simctl` output with already booted unavailable device that previously was considered as a valid simulator even it should not.
I was also to reproduce #12734 by given configuration which now succeeds:

```
xcversion(version: "10.1")
run_tests(device: "iPhone XR")
xcversion(version: "10.0")
run_tests(device: "iPhone XR")
```

### DIFF
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -40,12 +40,12 @@ module FastlaneCore
         end
 
         output.split(/\n/).each do |line|
-          break if line =~ /^-- Unavailable/
           next if line =~ /unavailable/
           next if line =~ /^== /
           if line =~ /^-- /
             (os_type, os_version) = line.gsub(/-- (.*) --/, '\1').split
           else
+            next if os_type =~ /^Unavailable/
 
             # "    iPad (5th generation) (852A5796-63C3-4641-9825-65EBDC5C4259) (Shutdown)"
             # This line will turn the above string into
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -40,6 +40,7 @@ module FastlaneCore
         end
 
         output.split(/\n/).each do |line|
+          break if line =~ /^-- Unavailable/
           next if line =~ /unavailable/
           next if line =~ /^== /
           if line =~ /^-- /
--- a/fastlane_core/spec/device_manager_spec.rb
+++ b/fastlane_core/spec/device_manager_spec.rb
@@ -248,7 +248,7 @@ describe FastlaneCore do
       allow(Open3).to receive(:popen3).with("xcrun simctl list runtimes").and_yield(nil, thing, nil, nil)
 
       devices = FastlaneCore::DeviceManager.simulators
-      expect(devices.count).to eq(2)
+      expect(devices.count).to eq(3)
 
       expect(devices[0]).to have_attributes(
         name: "iPhone 5s", os_type: "iOS", os_version: "12.0",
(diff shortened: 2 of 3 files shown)


polac24:
I signed it!

fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -40,6 +40,7 @@ def simulators(requested_os_type = "")
         end
 
         output.split(/\n/).each do |line|
+          break if line =~ /^-- Unavailable/
           next if line =~ /unavailable/
```
janpio: This wasn't enough, as it only skipped to the next line, correct?

polac24: `unavailable` (case insensitive) is meant to skip one simulator, not the entire group of simulators for some OS version. 
Because single simulator may be unavailable even os_version is available (e.g. see Xcode7 test spec) that line filtered such weird cases.

For Xcode 12 I get the output as below so even skipping one line (starting with`-- Unavailable`) would lead to an incorrect simulator `iPhone XR (8350A403-0E68-4F48-8ABA-C392F9CB8D7C) iOS 12.0` since in an `each` loop we initialize a simulator with a least recently observed valid iOS version.
```
-- iOS 12.0 --
    iPhone 5s (238C6D64-8720-4BFF-9DE9-FFBB9A1375D4) (Shutdown) 
    iPhone 6 (C68031AE-E525-4065-9DB6-0D4450326BDA) (Shutdown) 
-- Unavailable: com.apple.CoreSimulator.SimRuntime.iOS-12-1 --
    iPhone XR (8350A403-0E68-4F48-8ABA-C392F9CB8D7C) (Booted) 
    iPhone 5s (2721F3DC-A96B-4CA4-A9BD-B7CE3595BC82) (Shutdown) (unavailable, runtime profile not found) 
```



Optionally, I could just never initialize any simulator in a loop unless previously observed os_version was valid.

janpio (approved):
Looks good to me 👍

fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -40,6 +40,7 @@ def simulators(requested_os_type = "")
         end
 
         output.split(/\n/).each do |line|
+          break if line =~ /^-- Unavailable/
```
joshdholtz: This is assuming that `-- Unavailable` will always be in the bottom of the output, correct? If for some reason that `-- Unavailable` is not at the bottom of the output, this will break out of this loop completely and skip over some devices, correct? I think this is what your comment below meant of...

>Optionally, I could just never initialize any simulator in a loop unless previously observed os_version was valid.

I think if we could make this a bit safer and not break out that would be ideal. I can see this leading to some hard to debug/discover issues in the future if that `--Unavailable` wasn't always at the bottom.

Thoughts?

joshdholtz (dismissed):

fastlane_core/lib/fastlane_core/device_manager.rb
polac24: I updated PR as suggested

getaaron (approved):

getaaron:
Thanks so much @polac24, this is great work

PR #14423, opened 2019-03-18
Add the ability to specify Step Name for all actions

AdrieanKhisbe:
## Checklist
- [ ] I've updated the documentation if necessary. (🚧  _incoming_ )

### Motivation and Context
Some actions are meant to be runned several time with different configuration.

One of this exemple is  the [yarn plugin](https://github.com/joshrlesch/fastlane-plugin-yarn) that you may invoke different time with different script. Each of this invocation will appear with the same step_name, as `self.step_text` does not allow override on a per invocation basis.

**This PR tries to add a way to customize this (_each step name_)**

### Description

This PR add a mecanism so that the `step_name` of each action can be customized.
This is done by a facultative `:step_name` param that gets them removed from the properties passed to `action.run(params)`

#### Usage Example: 
```ruby
  rubocop(step_name: 'policekeeping_the_code_with_rubocop')
  yarn(command: 'fetch:version', step_name: 'fetch_version_code')
  yarn(command: 'pull:certificate', step_name: 'pull_certificates')
```

### DIFF
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -228,7 +228,11 @@ module Fastlane
           # If another action is calling this action, we shouldn't show it in the summary
           # (see https://github.com/fastlane/fastlane/issues/4546)
 
-          action_name = from_action ? nil : class_ref.step_text
+          if arguments.kind_of? Array and arguments.first.is_a?(Hash)
+            step_name_override = arguments.first[:step_name]
+            arguments.first.delete(:step_name)
+          end
+          action_name = from_action ? nil : (step_name_override || class_ref.step_text)
           Actions.execute_action(action_name) do
             # arguments is an array by default, containing an hash with the actual parameters
             # Since we usually just need the passed hash, we'll just use the first object if there is only one
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -228,11 +228,11 @@ module Fastlane
           # If another action is calling this action, we shouldn't show it in the summary
           # (see https://github.com/fastlane/fastlane/issues/4546)
 
-          if arguments.kind_of?(Array) && arguments.first.kind_of?(Hash)
-            step_name_override = arguments.first[:step_name]
-            arguments.first.delete(:step_name)
+          unless from_action # see https://github.com/fastlane/fastlane/issues/4546
+            args = arguments.kind_of?(Array) && arguments.first.kind_of?(Hash) ? arguments.first : {}
+            action_name = args[:step_name] || class_ref.step_text
+            args.delete(:step_name)
           end
-          action_name = from_action ? nil : (step_name_override || class_ref.step_text)
           Actions.execute_action(action_name) do
             # arguments is an array by default, containing an hash with the actual parameters
             # Since we usually just need the passed hash, we'll just use the first object if there is only one
--- a/fastlane/lib/fastlane/runner.rb
+++ b/fastlane/lib/fastlane/runner.rb
@@ -226,9 +226,8 @@ module Fastlane
       begin
         Dir.chdir(custom_dir) do # go up from the fastlane folder, to the project folder
           # If another action is calling this action, we shouldn't show it in the summary
-          # (see https://github.com/fastlane/fastlane/issues/4546)
 
-          unless from_action # see https://github.com/fastlane/fastlane/issues/4546
+          unless from_action
             args = arguments.kind_of?(Array) && arguments.first.kind_of?(Hash) ? arguments.first : {}
             action_name = args[:step_name] || class_ref.step_text
             args.delete(:step_name)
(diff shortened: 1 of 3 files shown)


AdrieanKhisbe:
Just signed the CLA 😉

fastlane/lib/fastlane/runner.rb
```diff
@@ -228,7 +228,11 @@ def execute_action(method_sym, class_ref, arguments, custom_dir: nil, from_actio
           # If another action is calling this action, we shouldn't show it in the summary
           # (see https://github.com/fastlane/fastlane/issues/4546)
 
-          action_name = from_action ? nil : class_ref.step_text
+          if arguments.kind_of?(Array) && arguments.first.kind_of?(Hash)
+            step_name_override = arguments.first[:step_name]
+            arguments.first.delete(:step_name)
```
getaaron: Why delete it?

fastlane/lib/fastlane/runner.rb
```diff
@@ -228,7 +228,11 @@ def execute_action(method_sym, class_ref, arguments, custom_dir: nil, from_actio
...
+          end
+          action_name = from_action ? nil : (step_name_override || class_ref.step_text)
```
getaaron: I find this confusing to read (not your fault, part of it's the original implementation)

What about something like this:

```ruby
unless from_action # see https://github.com/fastlane/fastlane/issues/4546
  action_name = arguments.first[:step_name] || class_ref.step_text
end
```

I think you can remove `if arguments.kind_of?(Array) && arguments.first.kind_of?(Hash)` by either

- moving the `arguments = [{}] if arguments.empty?` on line 137 out of its conditional
- moving the `if arguments.count == 0` stuff below to above

fastlane/lib/fastlane/runner.rb
AdrieanKhisbe: If I don't do this, it will crash because it"s not a supported option of the action

fastlane/lib/fastlane/runner.rb
AdrieanKhisbe: I do agree :)

fastlane/lib/fastlane/runner.rb
AdrieanKhisbe: though that could make sense to pass it though to the actoin. WDYT?

fastlane/lib/fastlane/runner.rb
AdrieanKhisbe: I did not manage to get ride of the kind_of check by trying the two alternative.
the first alternative break many test assumptions

fastlane/lib/fastlane/runner.rb
```diff
@@ -228,7 +228,11 @@ def execute_action(method_sym, class_ref, arguments, custom_dir: nil, from_actio
           # If another action is calling this action, we shouldn't show it in the summary
           # (see https://github.com/fastlane/fastlane/issues/4546)
```
getaaron: We can delete this comment now

fastlane/lib/fastlane/runner.rb
```diff
@@ -228,7 +228,11 @@ def execute_action(method_sym, class_ref, arguments, custom_dir: nil, from_actio
...
 
-          action_name = from_action ? nil : class_ref.step_text
+          unless from_action # see https://github.com/fastlane/fastlane/issues/4546
+            args = arguments.kind_of?(Array) && arguments.first.kind_of?(Hash) ? arguments.first : {}
```
getaaron: Nice 👍

getaaron:
LGTM 💗

Thoughts on a unit test?

fastlane/lib/fastlane/runner.rb
AdrieanKhisbe: will do

AdrieanKhisbe:
> LGTM 💗
> 
> Thoughts on a unit test?

@getaaron well, if you ask for 😁

janpio:
Whoa, so little code for such a nice improvement. I like where this is going!

AdrieanKhisbe:
@janpio @getaaron  And here it is, the unit tests just landed ✈️

jaleksynas:
This will be amazing for lanes with dozens of `sh` action calls as well.  There is some special handling the the `sh` action to support `log: false` case that may warrant a test case all it's own ;)

janpio (approved):

janpio:
I think this can be merged when some documentation was added (probably via a PR to fastlane/docs, somwhere in the "Advanced" sections). 

A rebase would also be nice, just to make 100% sure that nothing broke on Ubuntu as well (although I don't expect there to be any problems - our Ubuntu CI was just broken when this PR was created).

AdrieanKhisbe:
@janpio I rebased as you desired =)

janpio:
Beautiful these 6 green checks :)

Should we wait with the merge until a docs PR is done or do you maybe want to create an issue over at https://github.com/fastlane/docs/issues and create one at a later time? I'm good with both.

AdrieanKhisbe:
@janpio If this can speed up the process 👉 https://github.com/fastlane/docs/pull/811 :)

AdrieanKhisbe:
Thanks @fastlane-bot No hurry, I can wait a week for next release 😉

PR #14436, opened 2019-03-19
[spaceship] Automate phone number selection for "request code via SMS" in 2FA with `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` env var

### LINKED ISSUE #13980: [spaceship] Skip 2FA pushed token entry (to trusted phone selection), select correct phone no automatically
#13487 from @jarrodparkes contained a feature request:

>  I manage Apple developer accounts for ~100 different customers, some organization accounts, some individual accounts. For the individual accounts, I'm forced to access the Apple developer portal using their credentials which means 2FA in most cases. Thankfully, by coordinating with the customer, I can add my phone as a trusted device so that I don't have to contact them each time I need to access their account or generate a session...
> 
> But, the problem arises when I try to generate a web session using fastlane spaceauth -u [apple_id_email]. Since most of these Apple accounts have devices other than my phone registered, fastlane never prompts me to choose my phone for sending the code. Looking at the documentation and this closed issue #10790, there is no way to force fastlane to send the code to a certain number. One option would be to remove all the registered devices from the account, leaving only the desired phone number, but this is problematic because many of the customers I work with are using their devices with Apple throughout the day and logging them out would be BAD.

After #13800 is fixed, @jarrodparkes won't have to use the fallback any more but will be able to exit from the "pushed code entry" to a "select phone number" -> "enter code from text message" flow somehow, but this leaves the need for him to 1. exit the "pushed code entry" and 2. selection of his phone number.

---

We can probably implement an ENV variable to just skip 1. and request the phone numbers.

Then 2. can maybe be skipped with another ENV variable, depends on how much information we get on the phone number if they are very similar (by default Apple return the last 2 digits of a number. But what if you have 2 numbers with identical 3 digits at the end?).

janpio:
### Checklist

### Motivation and Context

2FA with token entry is a hassle to fastlane users. Some use cases also force people to use trusted phone numbers instead of the usual pushed tokens, so we can streamline their workflow by allowing them to skip the pushed token and remove the necessary interaction to get a token sent to a specific phone number.

### Description

The new environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` allows you to set a phone number, that is automatically used as the recipient of the 2FA token. You do not have to "escape" the normal input via inputting `sms` and then selecting your phone number from a dropdown any more.

#### Testing

I wrote a few tests for the specific methods I introduced that helped me to make sure they actually do what I expected them to do. Then I manually tested via `bundle exec fastlane spaceauth` as it basically just does the spaceship login - which is what this code impacts.

#### Unrelated changes 

This PR also includes a little comment in the cert runner and changes the way `fastlane spaceauth` handles exceptions - before they were just swallowed, but as my new code raises one if it can't find the correct phone number, that was not very good for testing. Now they are bubbled up to the user.

closes #13980

---

### Review Notes

- I am not really happy with the code of `phone_id_from_number`, the way I find the `phone_id` for the value of the ENV variable via the regex pattern, and especially the way I build it looks really crap and complicated. Is there a better way to do this?
- Does my spec file make sense? It worked for me when writing the initial code, but are those actually good tests? Should I do more here? The whole file was completely untested before. Now at least those two methods I touched are tested a tiny bit.

### DIFF
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -206,7 +206,7 @@ module Spaceship
         number_with_dialcode_regex = /^#{number_with_dialcode_regex_part}$/
         # => /^\+49([0-9]{8})85$/ or /^\+1([0-9]{7,8})66$/
 
-        return phone['id'] if phone_number.match?(number_with_dialcode_regex)
+        return phone['id'] if phone_number =~ number_with_dialcode_regex
         # +491621234585 matches /^\+49([0-9]{8})85$/
       end
 
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -122,8 +122,14 @@ module Spaceship
       # },
       code_length = security_code["length"]
 
-      unless(ENV["SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER"])
-        puts("")
+      puts("")
+      if ENV["SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER"]
+        puts("Environment variable `SPACESHIP_2FA_DEFAULT_PHONE_NUMBER` is set, automatically requesting 2FA token via SMS to that number")
+        puts("SPACESHIP_2FA_DEFAULT_PHONE_NUMBER = #{ENV['SPACESHIP_2FA_DEFAULT_PHONE_NUMBER']}")
+        phone_number = ENV["SPACESHIP_2FA_DEFAULT_PHONE_NUMBER"]
+        phone_id = phone_id_from_number(response.body["trustedPhoneNumbers"], phone_number)
+        body = request_two_factor_code_from_phone(phone_id, phone_number, code_length)
+      else
         puts("(Input `sms` to escape this prompt and select a trusted phone number to send the code as a text message)")
         code_type = 'trusteddevice'
         code = ask("Please enter the #{code_length} digit code:")
@@ -134,13 +140,6 @@ module Spaceship
           code_type = 'phone'
           body = request_two_factor_code_from_phone_choose(response.body["trustedPhoneNumbers"], code_length)
         end
-      else
-        puts("")
-        puts("Environment variable `SPACESHIP_2FA_DEFAULT_PHONE_NUMBER` is set, automatically requesting 2FA token via SMS to that number")
-        puts("SPACESHIP_2FA_DEFAULT_PHONE_NUMBER = #{ENV['SPACESHIP_2FA_DEFAULT_PHONE_NUMBER']}")
-        phone_number = ENV["SPACESHIP_2FA_DEFAULT_PHONE_NUMBER"]
-        phone_id = phone_id_from_number(response.body["trustedPhoneNumbers"], phone_number)
-        body = request_two_factor_code_from_phone(phone_id, phone_number, code_length)
       end
 
       puts("Requesting session...")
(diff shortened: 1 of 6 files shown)


max-ott:
Thanks for your great work @janpio 🙏 Will test this and give feedback. Is only running `bundle exec fastlane spaceauth` required to test this with having the right ENV set?

janpio:
That should execute that code, yes. But testing in other circumstances (normal spaceship usage) is of course also welcome.

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -172,23 +183,68 @@ def handle_two_factor(response, depth = 0)
...
+        number_with_dialcode_regex = /^#{number_with_dialcode_regex_part}$/
+        # => /^\+49([0-9]{8})85$/ or /^\+1([0-9]{7,8})66$/
+
+        return phone['id'] if phone_number =~ number_with_dialcode_regex
+        # +491621234585 matches /^\+49([0-9]{8})85$/
       end
+
+      # Handle case of phone_number not existing in phone_numbers because ENV var is wrong or matcher is broken
+      raise Tunes::Error.new, %(
+Could not find a matching phone number to #{phone_number} in #{phone_numbers}.
```
joshdholtz: We may want to mask the `phone_number` here 🤔 The `phone_numbers` also prints out as a Ruby object. Do we want to make that a little more CLI friendly? 😇
```sh
Could not find a matching phone number to <plain text phone number> in [{"numberWithDialCode"=>"+1 (•••) •••-••98", "pushMode"=>"sms", "obfuscatedNumber"=>"(•••) •••-••98", "id"=>1}].
```

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -123,14 +123,25 @@ def handle_two_factor(response, depth = 0)
...
+        puts("Environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` is set, automatically requesting 2FA token via SMS to that number")
+        puts("SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER = #{ENV['SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER']}")
+        phone_number = ENV["SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER"]
+        phone_id = phone_id_from_number(response.body["trustedPhoneNumbers"], phone_number)
         code_type = 'phone'
-        body = request_two_factor_code_from_phone(response.body["trustedPhoneNumbers"], code_length)
+        body = request_two_factor_code_from_phone(phone_id, phone_number, code_length)
+      else
+        puts("(Input `sms` to escape this prompt and select a trusted phone number to send the code as a text message)")
+        puts("(You can also set the environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` to automate this)")
```
joshdholtz: Could we put an example of the format here? It actually took me a few tries to get this working 😬 I didn't realize I had to add a `+1` to the start of my phone number. I tried adding a `1` to the start and that still didn't work 😝  So I think it might be good to put an example here for us people that don't regularly use the `+1` 😇

joshdholtz (changes requested):
@janpio I left a few small suggestions about masking numbers and changing some of the CLI output but this is so 🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥

spaceship/lib/spaceship/two_step_or_factor_client.rb
janpio: Problem is that we actually need the phone number here to be able to debug the problem. We could add something like "If you are uncomfortable sharing your phone number, please switch a few digits or replace them with `0`."

spaceship/lib/spaceship/two_step_or_factor_client.rb
janpio: This should probably be changed to a link to the documentation, that then will contain this information. (Basically: Phone number with country code (`+1`, `+49` or similar)) Having the instructions here will create a bit much output, don't you think?

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -123,14 +123,25 @@ def handle_two_factor(response, depth = 0)
       code_length = security_code["length"]
 
       puts("")
-      puts("(Input `sms` to escape this prompt and select a trusted phone number to send the code as a text message)")
-      code_type = 'trusteddevice'
-      code = ask("Please enter the #{code_length} digit code:")
-      body = { "securityCode" => { "code" => code.to_s } }.to_json
-
-      if code == 'sms'
+      if ENV["SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER"]
```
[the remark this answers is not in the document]
janpio: > Do we also want to make sure its not an empty string? 🤔

I chose _not_ to do that, because as soon as `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` is set the user wants this code to execute. The output should hopefully give them enough indication to discover that the variable is empty.

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -123,14 +123,25 @@ def handle_two_factor(response, depth = 0)
...
+        puts("Environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` is set, automatically requesting 2FA token via SMS to that number")
+        puts("SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER = #{ENV['SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER']}")
```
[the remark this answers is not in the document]
janpio: Interesting point.

Depends on how you view the system. I though that the user set the environment variable themselves, so no reason to not show him any data - if they want to paste the output they can still mask it manually (as many other thing from fastlane output).

Is there actually a way where the users could _not_ be aware of the phone number used? Not really, as they have to have access to that phone anyway, right?

spaceship/lib/spaceship/two_step_or_factor_client.rb
joshdholtz: @janpio Yeah, I definitely don't want to make this output too much longer. Linking to some documentation would be 💯

spaceship/lib/spaceship/two_step_or_factor_client.rb
joshdholtz: @janpio Could we only not mask if verbose is on? 🤔

spaceship/lib/spaceship/two_step_or_factor_client.rb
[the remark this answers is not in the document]
joshdholtz: @janpio Yeah yeah, might be good to say something if not empty! Empty string things are usually hard to debug otherwise 🙃

spaceship/lib/spaceship/two_step_or_factor_client.rb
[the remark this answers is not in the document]
joshdholtz: The only spot I could maybe see this happening is when sharing output/logs in an issue/PR. We don't want their phone numbers to be shared publicly on the internet like that 😬

spaceship/lib/spaceship/two_step_or_factor_client.rb
janpio: We could, but then we would have to add some information that the users needs to run `--verbose` before reporting the bug.

janpio:
Documentation part of this PR will be mainly taken care of via https://github.com/fastlane/docs/pull/804 which will create https://docs.fastlane.tools/advanced/authentication

spaceship/lib/spaceship/two_step_or_factor_client.rb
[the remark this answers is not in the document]
janpio: I would handle that similar to how we handle bundle IDs, team IDs etc - the user has take care of that themselves.

spaceship/lib/spaceship/two_step_or_factor_client.rb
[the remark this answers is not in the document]
janpio: Done.

spaceship/lib/spaceship/two_step_or_factor_client.rb
janpio: I prefer not masking here and leave the masking to users as we do with team ID and name etc.

spaceship/lib/spaceship/two_step_or_factor_client.rb
janpio: Added an output below that links to the content of this PR: https://github.com/fastlane/fastlane/pull/14524

joshdholtz (approved):
This is so goooooood! We are shipping this tomorrow 💪

PR #14439, opened 2019-03-20
Avoid Search for the Crashlytics API token if parameter gsp_path is set

oklimberg:
### Checklist

### Motivation and Context

This change reduces the confusion mentionend in Issue 14176, tho make clear, the missing symbols in Firebase after an upload are not related to the errors in the log, created, when searching for the API token

### Description

When the parameter gsp_path is set, there is no need to search for the API token
and we can skip the search, like it is already done in function find_gsp_path
if parameter api_token is set.


snatchev (approved):
Looks good to me.

Issue #14404, opened 2019-03-15
Add pattern matching to changelog_from_git_commits


rukano:
### Feature Request

In the action `changelog_from_git_commits` there should be an option for pattern matching the commit texts.

#### Motivation Behind Feature
In some projects, might be needed to create release notes and Changelog entries from the commits, but only merge commits into a specific branch (e.g. `develop`). This feature will also  provide more flexibility to the users to choose their own commit naming schema to bypass or filter irrelevant commits for the release notes.

#### Feature Description
The action `changelog_from_git_commits` provides some useful parameters already, but a `git log` paramter `--grep=""` is not implemented. Providing a parameter to pass to this function (e.g. `mathching_pattern`) will allow the users to have more flexibility when generating their changelogs.

Code examples:

In both functions `Actions.git_log_between` and `Actions.git_log_last_commits` an extra parameter will be added:

```ruby
def self.git_log_between(pretty_format, from, to, matching_pattern, merge_commit_filtering, date_format = nil, ancestry_path)
def self.git_log_last_commits(pretty_format, commit_count, matching_pattern, merge_commit_filtering, date_format = nil, ancestry_path)
```

In the function body, if the parameter exists, will be added as a `--grep=""` parameter to `git log`. Example:

```ruby
      command << "--grep=\"#{matching_pattern}\"" if matching_pattern
```

Since the parameter will be optional, if won't break the existing functionality.

#### Alternatives or Workarounds

My current workaround, before I decided to request this feature looks as follows:

```ruby
lane :make_release_notes do 
  command = ['git log $(git describe --tags --abbrev=0)..HEAD']
  command << "--grep=\"Merge branch .* into 'develop'\""
  command << "--pretty=\"\%b\""
  msg = Actions.sh(command.compact.join(' '), log: false).chomp
  UI.message msg # <- here I use the contents later for the Changelog.
end
```

Which practically mimicks the `git_log_last_commits` and the `git_log_between` without the parameters I haven't been using.

Adding the extra parameter to `changelog_from_git_commits` would make the function more capable, and using the `git log` filters instead of filtering by hand on the lane or ruby itself.

I have implemented this in my fastlane fork, but since it is my first issue/contribution, I'm not familiar with the testing and build process, but I might be able to make a Pull Request if desired. :)

EDIT: fastlane env as requested by the bot
[fastlane_env.txt](https://github.com/fastlane/fastlane/files/2974265/fastlane_env.txt)

janpio:
[...]

PR #14440, opened 2019-03-20
[action] Add optional argument to filter git commits for changelog

rukano:
Added an optional argument `matching_pattern` to the `git_log_between` and `git_log_last_commits` actions. It injects the pattern to the `git log` command for the `grep` argument.

Added documentation and sample code.

### Checklist

### Motivation and Context
In some projects, might be needed to create release notes and Changelog entries from the commits, but only merge commits into a specific branch (e.g. `develop`). This feature will also  provide more flexibility to the users to choose their own commit naming schema to bypass or filter irrelevant commits for the release notes.

#14404 

### Description
The action `changelog_from_git_commits` provides some useful parameters already, but a `git log` parameter `--grep=""` is not implemented. Providing a parameter to pass to this function (e.g. `matching_pattern`) will allow the users to have more flexibility when generating their Changelogs.

Code examples:

In both functions `Actions.git_log_between` and `Actions.git_log_last_commits` an extra parameter will be added:

```ruby
def self.git_log_between(pretty_format, from, to, matching_pattern, merge_commit_filtering, date_format = nil, ancestry_path)
def self.git_log_last_commits(pretty_format, commit_count, matching_pattern, merge_commit_filtering, date_format = nil, ancestry_path)
```

In the function body, if the parameter exists, will be added as a `--grep=""` parameter to `git log`. Example:

```ruby
      command << "--grep=\"#{matching_pattern}\"" if matching_pattern
```

Since the parameter will be optional, if won't break the existing functionality.

I've been using the workaround for a while, and tested it on my local fork, from where I am creating this PR.

### DIFF
--- a/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
+++ b/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
@@ -37,9 +37,9 @@ module Fastlane
 
         Dir.chdir(params[:path]) do
           if params[:commits_count]
-            changelog = Actions.git_log_last_commits(params[:pretty], params[:commits_count], merge_commit_filtering, params[:date_format], params[:ancestry_path])
+            changelog = Actions.git_log_last_commits(params[:pretty], params[:commits_count], params[:matching_pattern], merge_commit_filtering, params[:date_format], params[:ancestry_path])
           else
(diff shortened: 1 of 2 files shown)


rukano:
Wrong signing identity for the commit. Will open a new PR

PR #14441, opened 2019-03-20
[changelog_from_git_commits] Add optional argument to filter commits.

rukano:
Added an optional argument `matching_pattern` to the `git_log_between` and `git_log_last_commits` actions. It injects the pattern to the `git log` command for the `grep` argument.

Added documentation and sample code.

### Checklist

### Motivation and Context
In some projects, might be needed to create release notes and Changelog entries from the commits, but only merge commits into a specific branch (e.g. `develop`). This feature will also  provide more flexibility to the users to choose their own commit naming schema to bypass or filter irrelevant commits for the release notes.

#14404 

### Description
The action `changelog_from_git_commits` provides some useful parameters already, but a `git log` parameter `--grep=""` is not implemented. Providing a parameter to pass to this function (e.g. `matching_pattern`) will allow the users to have more flexibility when generating their Changelogs.

Code examples:

In both functions `Actions.git_log_between` and `Actions.git_log_last_commits` an extra parameter will be added:

```ruby
def self.git_log_between(pretty_format, from, to, matching_pattern, merge_commit_filtering, date_format = nil, ancestry_path)
def self.git_log_last_commits(pretty_format, commit_count, matching_pattern, merge_commit_filtering, date_format = nil, ancestry_path)
```

In the function body, if the parameter exists, will be added as a `--grep=""` parameter to `git log`. Example:

```ruby
      command << "--grep=\"#{matching_pattern}\"" if matching_pattern
```

Since the parameter will be optional, if won't break the existing functionality.

I've been using the workaround for a while, and tested it on my local fork, from where I am creating this PR.

### DIFF
--- a/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
+++ b/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
@@ -37,9 +37,9 @@ module Fastlane
 
         Dir.chdir(params[:path]) do
           if params[:commits_count]
-            changelog = Actions.git_log_last_commits(params[:pretty], params[:commits_count], merge_commit_filtering, params[:date_format], params[:ancestry_path])
+            changelog = Actions.git_log_last_commits(params[:pretty], params[:commits_count], params[:matching_pattern], merge_commit_filtering, params[:date_format], params[:ancestry_path])
           else
-            changelog = Actions.git_log_between(params[:pretty], from, to, merge_commit_filtering, params[:date_format], params[:ancestry_path])
+            changelog = Actions.git_log_between(params[:pretty], from, to, params[:matching_pattern], merge_commit_filtering, params[:date_format], params[:ancestry_path])
           end
 
           changelog = changelog.gsub("\n\n", "\n") if changelog # as there are duplicate newlines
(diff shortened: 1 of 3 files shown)


joshdholtz (dismissed):
@rukano Thanks for this contribution! I would be 🔥 if we could add some tests here before we merge this. I tried to push a commit to your PR with a test but it doesn't look like I was given the ability to do that. So, can you add this test for this in `fastlane/spec/actions_specs/changelog_from_git_commits_spec.rb`...

```rb
       it "Uses grep matching pattern if requested" do
         matching_pattern = 'pizza'
           changelog_from_git_commits(matching_pattern: '#{matching_pattern}')
         end").runner.execute(:test)

         inner_command = "git describe --tags `git rev-list --tags --max-count=1`"
         pseudocommand = "git log --pretty=\"%B\" --grep=\"#{matching_pattern}\" #{inner_command.shellescape}...HEAD"
         expect(result).to eq(pseudocommand)
       end
```

rukano:
@joshdholtz thanks, you are right. I pushed the test case you suggested. Thanks for reviewing and let me know if you want access to the branch, or if I should add something else. Cheers!

joshdholtz:
@rukano Ahhh, it looks like we have a merge conflict here! If you want to give me permission (and other contributors write access to your branch that would be 💯) and I can fix this for you if you don't want to 🙃

https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

rukano:
Hi @joshdholtz 
I have mantainer edits activated actually, you should be able to make the changes. If it doesn't work, I can try to solve the conflicts and push again, if necessary. Or is there another way to give you access?
Screenshot 2019-04-01 at 13 22 40

ahti:
Is there anything preventing this from being merged? This functionality would come in very handy.

lukebrandonfarrell:
Why didn't this get merged?

pzmudzinski:
Any chance for a merge? That we could extract commits including ticket numbers from JIRA instead all of them.

iBotPeaches (approved):
tested locally. Works

Issue #14449, opened 2019-03-21
verify_block not invoked when using environment variables


jaleksynas:
### New Issue Checklist
- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Similar to #11017 I can confirm that values set via environment variables are not validated by the verify_block. This allows invalid values to get into the `run` of an action where the validation should be assumed.

##### Sample code
https://github.com/jaleksynas/VerifyBlockBroken

##### Command executed
`fastlane test`
`fastlane test_env`

janpio:
Difficult from #11017 to retrace if this once worked differently or if there were just related changes that might have fixed (or not fixed) the issue back then.

This is where it seems to get the values from the option or ENV var alternatively:
https://github.com/fastlane/fastlane/blob/4ad2e58e41b11c8197f517d902bd0fbdb90f8e71/fastlane_core/lib/fastlane_core/configuration/configuration.rb#L219-L227

The verification seems to happen here:
https://github.com/fastlane/fastlane/blob/4ad2e58e41b11c8197f517d902bd0fbdb90f8e71/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L207-L228

Which the first one connects to via `set` (`valid?` is wrapped in `verify!`):
https://github.com/fastlane/fastlane/blob/4ad2e58e41b11c8197f517d902bd0fbdb90f8e71/fastlane_core/lib/fastlane_core/configuration/configuration.rb#L265-L277

I don't really understand where the readinf of the var and the verification come together - but maybe that _is_ the problem.

jaleksynas:
@janpio That _is_ the problem!  Setting the value in other ways `verify!`'s  correctly.  It looks like the value from the env is read during fetch (but without the verification).

Adding `&& option.verify!(ENV[option.env_name])` on line 221 of configuration.rb does the trick.

janpio:
That is a great first step!

Is this maybe where all the "normal" parameters are validated? 
https://github.com/fastlane/fastlane/blob/4ad2e58e41b11c8197f517d902bd0fbdb90f8e71/fastlane_core/lib/fastlane_core/configuration/configuration.rb#L80-L89 
[...]

PR #14450, opened 2019-03-22
[fastlane_core] Verify environment variable values for options

jaleksynas:
### Checklist

### Motivation and Context
The change ensures that value for configuration options are validated by the verify_block of the option as expected. #14449  

### Description
Convert and verify the value from the environment before allowing it to be fetched.

Tested using the sample project from the issue as well as running all the rspec tests.

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -219,7 +219,8 @@ module FastlaneCore
       value = if @values.key?(key) && !@values[key].nil?
                 @values[key]
               elsif option.env_name && !ENV[option.env_name].nil?
-                ENV[option.env_name].dup
+                # Convert and verify! before using it (https://github.com/fastlane/fastlane/issues/14449)
+                ENV[option.env_name].dup if option.verify!(option.auto_convert_value(ENV[option.env_name]))
               elsif self.config_file_options.key?(key)
                 self.config_file_options[key]
               else
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -219,7 +219,7 @@ module FastlaneCore
       value = if @values.key?(key) && !@values[key].nil?
                 @values[key]
               elsif option.env_name && !ENV[option.env_name].nil?
-                # Convert and verify! before using it (https://github.com/fastlane/fastlane/issues/14449)
+                # verify! before using (see https://github.com/fastlane/fastlane/issues/14449)
                 ENV[option.env_name].dup if option.verify!(option.auto_convert_value(ENV[option.env_name]))
               elsif self.config_file_options.key?(key)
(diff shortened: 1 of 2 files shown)


getaaron:
Looks like a nice improvement 👍👍👍

I have 2 questions

- Would it be more readable with the `verify!` in the existing conditional on line 221?
- Can we add a test to cover the new expected behavior?

CC @janpio

jaleksynas:
@getaaron I most certainly considered it, but with the nested call to auto_convert_value it was reading a bit long on that line.
This placement also allowed for a comment to clarify the extra logic. 

I defer to the wider group on this as it is purely stylistic.

I'll look to add a specific test, There are certainly tests (like [this test](https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/say_spec.rb#L56) ) covering this -- which is why it was necessary to `auto_convert_values` before `verify!`.

janpio:
(Feel free to clean up even more if you want to @jaleksynas - this all looks pretty "organically grown" and this might be a chance to untangle that a bit if you have the time and motivation. Would 😻 you for it)

jaleksynas:
@getaaron I added relevant tests.

@janpio I would like to see this particular issue get resolved sooner than I can get a more complete refactor completed.  

I do plan to continue to dig into the flow and try to untangle it, though I have found _any_ changes here to have wide reaching impacts within the tests, so I am treading lightly.

fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -219,7 +219,8 @@ def fetch(key, ask: true)
       value = if @values.key?(key) && !@values[key].nil?
                 @values[key]
               elsif option.env_name && !ENV[option.env_name].nil?
-                ENV[option.env_name].dup
+                # Convert and verify! before using it (https://github.com/fastlane/fastlane/issues/14449)
```
janpio: Does the conversion actually get used in the end or is it only for the `verify!` call? Then this might be formulated confusingly.

getaaron (approved):
Looks like a strong improvement. Thanks for catching this, I’m honestly surprised fastlane doesn’t already do this

fastlane_core/lib/fastlane_core/configuration/configuration.rb
jaleksynas: It is only for the `verify!` call due to part of the validation being the boolean check within the method.  [This test](https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/say_spec.rb#L56) fails when verifying before.  

I tried moving the verify around in a variety of ways, most of which resulted in a number of failed tests. This one seemed to preserve the most identical functionality while addressing the verify issue directly.

janpio:
```suggestion
                # verify! before using (see https://github.com/fastlane/fastlane/issues/14449)
```

joshdholtz (approved):
@jaleksynas This is a great find! Thanks for the contribution here ❤️

PR #14455, opened 2019-03-22
[spaceship] Clean up README

janpio:
Remove outdated information:

- "Bypass trusted device and use SMS for verification" is not needed any more with `sms` escape implemented
- "transporter" section is not related to spaceship and better described in real docs


joshdholtz (approved):
💯

PR #14456, opened 2019-03-22
[pilot] Do not use spaceship if both skip_waiting_for_build_processing and apple_id are set 

### LINKED ISSUE #14258: Upload to AppStore without FASTLANE_SESSION
### Question Checklist

### Question Subject
Uploading to AppStore without 2FA

#### Question Description
As asked in #14244 I just want to upload my app to AppStore. 
I have set `FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD ` and I still get prompted with 2FA. 

Is there a way to avoid this or do I need to update a session every month?

My lane is here: 

``` ruby
  lane :beta do
    build_app(scheme: "MyApp",
              include_bitcode: true)
    upload_to_testflight(skip_waiting_for_build_processing: true)
  end
```

janpio:
### Checklist

### Motivation and Context
With the new 2FA rules by Apple, more and more users are looking for a way to use app specific passwords to upload to the App Store. 

### Description
As iTunes Transporter does accept app specific passwords, and is used by `pilot` to upload the binary, we can achieve this with a few tiny changes in the program logic (do not log in via Spaceship if there is no reason to. This is the case if both `skip_waiting_for_build_processing` and `apple_id` are set).

closes #14258
closes #14723

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -51,7 +51,7 @@ module Pilot
       end
 
       UI.message("If you want to skip waiting for the processing to be finished, use the `skip_waiting_for_build_processing` option")
-      latest_build = wait_for_build_processing_to_be_complete
+      latest_build = wait_for_build_processing_to_be_complete(options)
       distribute(options, build: latest_build)
     end
 
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -10,12 +10,13 @@ require_relative 'manager'
 module Pilot
   class BuildManager < Manager
     def upload(options)
-      start(options) unless options[:skip_waiting_for_build_processing] && options[:apple_id]
-      @config = options if options[:skip_waiting_for_build_processing] && options[:apple_id] # ugly as hell, we want to do that differently
+      # Only need to login before upload if no apple_id was given
+      should_login = options[:apple_id].nil?
+      start(options, should_login: should_login)
 
       options[:changelog] = self.class.sanitize_changelog(options[:changelog]) if options[:changelog]
 
-      UI.user_error!("No ipa file given") unless options[:ipa]
+      UI.user_error!("No ipa file given") unless config[:ipa]
 
       if options[:changelog].nil? && options[:distribute_external] == true
         if UI.interactive?
@@ -25,18 +26,18 @@ module Pilot
         end
       end
 
-      UI.success("Ready to upload new build to TestFlight (App: #{fetch_only_apple_id})...")
+      UI.success("Ready to upload new build to TestFlight (App: #{fetch_apple_id})...")
 
       dir = Dir.mktmpdir
 
       platform = fetch_app_platform
-      package_path = FastlaneCore::IpaUploadPackageBuilder.new.generate(app_id: fetch_only_apple_id,
+      package_path = FastlaneCore::IpaUploadPackageBuilder.new.generate(app_id: fetch_apple_id,
                                                                       ipa_path: options[:ipa],
                                                                   package_path: dir,
                                                                       platform: platform)
 
       transporter = transporter_for_selected_team(options)
-      result = transporter.upload(fetch_only_apple_id, package_path)
+      result = transporter.upload(fetch_apple_id, package_path)
 
       unless result
         UI.user_error!("Error uploading ipa file, for more information see above")
@@ -50,12 +51,14 @@ module Pilot
         return
       end
 
+      login_if_not_logged_in
+
       UI.message("If you want to skip waiting for the processing to be finished, use the `skip_waiting_for_build_processing` option")
-      latest_build = wait_for_build_processing_to_be_complete(options)
+      latest_build = wait_for_build_processing_to_be_complete
       distribute(options, build: latest_build)
     end
 
-    def wait_for_build_processing_to_be_complete(options)
+    def wait_for_build_processing_to_be_complete
       platform = fetch_app_platform
       app_version = FastlaneCore::IpaFileAnalyser.fetch_app_version(config[:ipa])
       app_build = FastlaneCore::IpaFileAnalyser.fetch_app_build(config[:ipa])
@@ -188,6 +191,10 @@ module Pilot
 
     private
 
+    def login_if_not_logged_in
+      login unless Spaceship::Tunes.client
+    end
+
     def describe_build(build)
       row = [build.train_version,
              build.build_version,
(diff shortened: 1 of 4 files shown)


pilot/lib/pilot/build_manager.rb
```diff
@@ -10,11 +10,12 @@
 module Pilot
   class BuildManager < Manager
     def upload(options)
-      start(options)
+      start(options) unless options[:skip_waiting_for_build_processing] && options[:apple_id]
+      @config = options if options[:skip_waiting_for_build_processing] && options[:apple_id] # ugly as hell, we want to do that differently
```
janpio: This is ugly as hell, but the only way to have `@config` be filled as it would be in `start`. I want to avoid calling `start()`, as it triggers the spaceship login.

janpio: Actually, thinking about it, doing this here manually might _break_ further calls to `start` because it will prevent the spaceship login to happen later if the presence of `@config` is used as an indicator if a login has happened before or not.

Not a problem right now as we don't have any further calls of `start` or anything that then needs the spaceship login in this case, it is still ugly.

What would a better way to acheive the same thing be? Add a `spaceship_login_happened` bool to use instead of abusing `@config`?

joshdholtz (commented):
@janpio This is a cool PR 🙌 I rebased and pushed some changes on top here to clean up that start/login logic that you commented on that needed improvement. Let me know your thoughts!

pilot/lib/pilot/build_manager.rb
```diff
@@ -50,15 +51,17 @@ def upload(options)
         return
       end
 
+      login_if_not_logged_in
```
janpio: That's smart.

pilot/lib/pilot/build_manager.rb
```diff
@@ -218,7 +225,8 @@ def should_update_localized_build_information?(options)
     # If there are fewer than two teams, don't infer the provider.
     def transporter_for_selected_team(options)
       generic_transporter = FastlaneCore::ItunesTransporter.new(options[:username], nil, false, options[:itc_provider])
-      return generic_transporter unless options[:itc_provider].nil? && Spaceship::Tunes.client.teams.count > 1
+      return generic_transporter if options[:itc_provider] || Spaceship::Tunes.client.nil?
```
janpio: What's this change about?

joshdholtz: @janpio `Spaceship::Tunes.client.teams.count > 1` would 💣 if not logged in so checking that on the following line separately 😊

janpio:
Ok, your changes look good to me @joshdholtz. From my side this is good to merge.

pilot/lib/pilot/build_manager.rb
```diff
@@ -49,6 +51,8 @@ def upload(options)
         return
       end
 
+      login_if_not_logged_in
```
snatchev: Code style nitpick: methods with no input and no outputs give me pause. Why does this exist if it's only called once (and it doesn't appear to be used in mocking or anything like that) ?

pilot/spec/build_manager_spec.rb
```diff
@@ -221,4 +223,61 @@
...
+        fake_itunestransporter = double
+        allow(fake_itunestransporter).to receive(:upload).and_return(true)
+        allow(FastlaneCore::ItunesTransporter).to receive(:new).and_return(fake_itunestransporter)
+      end
+
+      it "is not executed when skip_waiting_for_build_processing and apple_id are set" do
+        fake_build_manager.upload(upload_options)
+      end
+
+      it "is executed when skip_waiting_for_build_processing and apple_id are not set" do
```
snatchev: There are no expectations in this spec. Perhaps change some of the `allow`s to `expect` ?

pilot/spec/build_manager_spec.rb
```diff
@@ -221,4 +223,61 @@
...
+        allow(fake_ipauploadpackagebuilder).to receive(:generate).and_return(true)
+        allow(FastlaneCore::IpaUploadPackageBuilder).to receive(:new).and_return(fake_ipauploadpackagebuilder)
+
+        fake_itunestransporter = double
+        allow(fake_itunestransporter).to receive(:upload).and_return(true)
+        allow(FastlaneCore::ItunesTransporter).to receive(:new).and_return(fake_itunestransporter)
+      end
+
+      it "is not executed when skip_waiting_for_build_processing and apple_id are set" do
+        fake_build_manager.upload(upload_options)
```
snatchev: I am unclear on how we assert that `upload` is not executed here.

pilot/spec/build_manager_spec.rb
janpio: Silly question: What's the difference?

pilot/spec/build_manager_spec.rb
janpio: This test is: When using the upload, no login is executed, under special circumstances.
Although this is only verified by all the other stuff from line 254 and following not being present.

pilot/lib/pilot/build_manager.rb
janpio: @joshdholtz, your code.

pilot/spec/build_manager_spec.rb
snatchev: `expect` is similar to an `allow`, but it will fail if that method does _not_ get called by the end of the spec. Just by having `allow`, the spec will either throw an error or pass, but it cannot fail.

janpio: Ah ok, then this really makes sense to change here. 
I built this test up from nothing until it let the code execute enough to pass.

pilot/lib/pilot/build_manager.rb
janpio: ping @joshdholtz

snatchev:
Heyo, who's the point person on this? @janpio or @joshdholtz? Couldnt quite tell from the history.

janpio:
I had the idea and implemented a first version, with some pretty complicated code. Then @joshdholtz fortunately jumped in and refactored it to something a lot more readable via https://github.com/fastlane/fastlane/pull/14456/commits/68eddca23178e22c768152de96487e726e6005cf

I have on my TODO to rewrite the tests to take care of the things your commented on, and I @ mentioned @joshdholtz in your comment that applied to something that he wrote.

pilot/spec/build_manager_spec.rb
janpio: I added an appropriate `expect` to each of the two specs: Once that the `Manager.login` method (which triggers a spaceship login) is _NOT_ called, and once that it is called at least once.

pilot/spec/build_manager_spec.rb
janpio: Taken care of with the `expect.not to.`

janpio:
Rebased so CI can actually run (Circle removed our old Xcode 9.0.0 we were using).

Also reworked the tests to address valid and great feedback from @snatchev. Now uses `expect` to make sure the login method that triggers the spaceship login is called or NOT called correctly (which this PR is all about).

I also created https://github.com/fastlane/fastlane/pull/14723 with _just_ the tests of this, to show that the current codebase always calls the spaceship login and thus fails the test (in the expected way).

joshdholtz (approved):
This is 💯 to me!! LGTM

PR #14462, opened 2019-03-23
[sign] fix sigh not using custom keychain path

### LINKED ISSUE #14453: Cert doesn't use custom keychain to find certificate
### New Issue Checklist

### Issue Description

Cert checker supports custom keychain via `in_keychain` parameter:
https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/cert_checker.rb#L9

This is however not used from cert:

https://github.com/fastlane/fastlane/blob/master/cert/lib/cert/runner.rb#L15
https://github.com/fastlane/fastlane/blob/master/cert/lib/cert/runner.rb#L95

The result of this is that if you pass custom keychain to cert and already have certificate and key imported, cert is not able to find them and it either:
- fails because of certificate limit
- creates new certificate

jcavar:
### Checklist

### Motivation and Context

This fixes https://github.com/fastlane/fastlane/issues/14453

### DIFF
--- a/cert/spec/runner_spec.rb
+++ b/cert/spec/runner_spec.rb
@@ -16,14 +16,20 @@ describe Cert do
       allow(Spaceship.client).to receive(:in_house?).and_return(false)
       allow(Spaceship.certificate.production).to receive(:all).and_return([certificate])
 
-      allow(FastlaneCore::CertChecker).to receive(:installed?).and_return(true)
+      certificate_path = "#{Dir.pwd}/cert_id.cer"
+      keychain_path = Dir.pwd.to_s
+      expect(FastlaneCore::CertChecker).to receive(:installed?)
+        .with(certificate_path, in_keychain: keychain_path)
+        .twice
+        .and_return(true)
 
       options = { keychain_path: "." }
       Cert.config = FastlaneCore::Configuration.create(Cert::Options.available_options, options)
 
       Cert::Runner.new.launch
       expect(ENV["CER_CERTIFICATE_ID"]).to eq("cert_id")
-      expect(ENV["CER_FILE_PATH"]).to eq("#{Dir.pwd}/cert_id.cer")
+      expect(ENV["CER_FILE_PATH"]).to eq(certificate_path)
+      expect(ENV["CER_KEYCHAIN_PATH"]).to eq(keychain_path)
       File.delete(ENV["CER_FILE_PATH"])
     end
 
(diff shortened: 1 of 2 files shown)


cert/spec/runner_spec.rb
```diff
@@ -95,7 +101,7 @@
       end
 
       let(:generate) do
-        options = { output_path: temp, filename: filename }
+        options = { output_path: temp, filename: filename, keychain_path: "." }
```
jcavar: This is workaround for tests failing on Ubuntu machine. Default argument value for `keychain_path` (`~/Library/Keychains/login.keychain`) is not valid on Ubuntu.

I am not sure if and how cert works on Ubuntu and workarounds keychain? I think concept of macOS keychain is not available on Ubuntu (`security` command doesn't work)?

joshdholtz: We probably don't want this as a work around just for Ubuntu tests because this means that if anybody tries to run `cert` in Ubuntu without passing a `keychain_path` of `"."`, it will probably fail in real life 😬

We might want to do a check for macOS somewhere instead? I think @janpio might have a better suggestion here since he has more of the non-macOS experience 😇

joshdholtz (changes requested):
Might want to find a proper solution to fixing the tests in Ubuntu 🤔

cert/spec/runner_spec.rb
jcavar: I am open for suggestions. However, `keychain_path` has default value and is not optional argument. So it seems to me that `cert` hasn't been working properly on Ubuntu anyway.

It is true that this change pushed `Cert.config[:keychain_path]` up in code path which triggered exception earlier. So it might break some of existing code.

Is there a way to do `Cert.config[:keychain_path]` so that it doesn't raise error if `keychain_path` is not valid? Or check if it is valid somehow?

janpio: `cert` is probably broken on Ubuntu outside of the tests if there is no special handling for "keychain doesn't actually exist on non-macOS" yet. So I am good with working with `.` here, although a comment might not be a bad idea. One day we might want to get it to work there, and this might be a valuable pointer.

getaaron: What about skipping the tests `unless FastlaneCore::Helper.mac?`?

janpio: The tests are actually passing, so no need to skip. The test is not _really_ testing all the functionality. The real solution here is to go in deeper and make sure all the keychain stuff is limited to macOS - but that is quite a lot of work and outside the scope of this PR.

joshdholtz (approved):
Thanks for adding that comment! This is good for me then 😊

Issue #13227, opened 2018-08-27
[Actions.sh] Windows doesn't support command substitution


janpio:
http://www.gnu.org/software/bash/manual/bashref.html#Command-Substitution

> Command substitution allows the output of a command to replace the command itself. Command substitution occurs when a command is enclosed as follows:
> 
> $(command)
> 
> or
> 
> \`command\`

Fastlane uses this in some places:
https://github.com/fastlane/fastlane/blob/daaf326ea8492b8e07a0315812c53219e046ebba/fastlane/lib/fastlane/helper/git_helper.rb#L32-L35

Unfortunately Windows does not support this: https://stackoverflow.com/a/11061413/252627

---

During tests this doesn't pop up as `Actions.sh` just returns a command string instead of executing it, which of course works fine. The problem would only pop up when the command was executed, not doing what it should.

---

There might be hacky alternatives:
https://superuser.com/questions/289344/is-there-something-like-command-substitution-in-windows-cli
https://stackoverflow.com/questions/2768608/batch-equivalent-of-bash-backticks
http://cplusplus.bordoon.com/cmd_exe_variables.html

---

Probably better solution:
- Execute the wrapped command seperately, use the result in the outer command
- In test scenario, return the test command as it would have been executed on Mac so tests pass

janpio:
See https://github.com/fastlane/fastlane/pull/14471#issuecomment-476998114

PR #14471, opened 2019-03-25
[fastlane] git: more responsible, Windows-compatible shellout

dotdoom:
### Checklist

### Motivation and Context

Some git actions (ex.: `changelog_from_git_commits`) were not working on Windows. This have been noticed when our projects got student contributors with Windows workstations. While fixing that, I also fixed shell escaping in `git_helper.rb`.

### Description

* some things are `#shellescape`'d, the others aren't. Make it uniform and let `Actions.sh` handle escaping properly;
* additionally, do not use bquot (`` ` ``). It is not compatible with Windows cmd.exe;
* tested on a real project by running on Windows in AppVeyor with `changelog_from_git_commits`, `number_of_commits` and `last_git_tag`.

### DIFF
--- a/fastlane/lib/fastlane/helper/git_helper.rb
+++ b/fastlane/lib/fastlane/helper/git_helper.rb
@@ -3,36 +3,38 @@ module Fastlane
     GIT_MERGE_COMMIT_FILTERING_OPTIONS = [:include_merges, :exclude_merges, :only_include_merges].freeze
 
     def self.git_log_between(pretty_format, from, to, merge_commit_filtering, date_format = nil, ancestry_path)
-      command = ['git log']
-      command << "--pretty=\"#{pretty_format}\""
-      command << "--date=\"#{date_format}\"" if date_format
+      command = %w(git log)
+      command << "--pretty=#{pretty_format}"
+      command << "--date=#{date_format}" if date_format
       command << '--ancestry-path' if ancestry_path
-      command << "#{from.shellescape}...#{to.shellescape}"
+      command << "#{from}...#{to}"
       command << git_log_merge_commit_filtering_option(merge_commit_filtering)
-      Actions.sh(command.compact.join(' '), log: false).chomp
+      Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
     end
 
     def self.git_log_last_commits(pretty_format, commit_count, merge_commit_filtering, date_format = nil, ancestry_path)
-      command = ['git log']
-      command << "--pretty=\"#{pretty_format}\""
-      command << "--date=\"#{date_format}\"" if date_format
+      command = %w(git log)
+      command << "--pretty=#{pretty_format}"
+      command << "--date=#{date_format}" if date_format
       command << '--ancestry-path' if ancestry_path
-      command << "-n #{commit_count}"
+      command << '-n' << commit_count.to_s
       command << git_log_merge_commit_filtering_option(merge_commit_filtering)
-      Actions.sh(command.compact.join(' '), log: false).chomp
+      Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
     end
 
     def self.last_git_tag_name(match_lightweight = true, tag_match_pattern = nil)
-      tag_pattern_param = tag_match_pattern ? "=#{tag_match_pattern.shellescape}" : ''
+      tag_pattern_param = tag_match_pattern ? "=#{tag_match_pattern}" : ''
+      tag_name = Actions.sh('git', 'rev-list', "--tags#{tag_pattern_param}",
+                            '--max-count=1').chomp
 
-      command = ['git describe']
+      command = %w(git describe)
       command << '--tags' if match_lightweight
-      command << "`git rev-list --tags#{tag_pattern_param} --max-count=1`"
-      Actions.sh(command.compact.join(' '), log: false).chomp
+      command << tag_name
+      Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
     end
@@ -52,10 +54,10 @@ module Fastlane
     # Gets the last git commit information formatted into a String by the provided
     # pretty format String. See the git-log documentation for valid format placeholders
     def self.last_git_commit_formatted_with(pretty_format, date_format = nil)
-      command = ['git log -1']
-      command << "--pretty=\"#{pretty_format}\""
-      command << "--date=\"#{date_format}\"" if date_format
-      Actions.sh(command.compact.join(' '), log: false).chomp
+      command = %w(git log -1)
+      command << "--pretty=#{pretty_format}"
+      command << "--date=#{date_format}" if date_format
+      Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
     end
--- a/fastlane/lib/fastlane/helper/git_helper.rb
+++ b/fastlane/lib/fastlane/helper/git_helper.rb
@@ -9,6 +9,8 @@ module Fastlane
       command << '--ancestry-path' if ancestry_path
       command << "#{from}...#{to}"
       command << git_log_merge_commit_filtering_option(merge_commit_filtering)
+      # "*command" syntax expands "command" array into variable arguments, which
+      # will then be individually shell-escaped by Actions.sh.
       Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
--- a/fastlane/lib/fastlane/helper/git_helper.rb
+++ b/fastlane/lib/fastlane/helper/git_helper.rb
@@ -26,14 +26,23 @@ module Fastlane
       nil
     end
 
-    def self.last_git_tag_name(match_lightweight = true, tag_match_pattern = nil)
+    def self.last_git_tag_hash(tag_match_pattern = nil)
       tag_pattern_param = tag_match_pattern ? "=#{tag_match_pattern}" : ''
-      tag_name = Actions.sh('git', 'rev-list', "--tags#{tag_pattern_param}",
-                            '--max-count=1').chomp
+      Actions.sh('git', 'rev-list', "--tags#{tag_pattern_param}", '--max-count=1').chomp
+    rescue
+      nil
+    end
+
+    def self.last_git_tag_name(match_lightweight = true, tag_match_pattern = nil)
+      hash = last_git_tag_hash(tag_match_pattern)
+      # If hash is nil (command fails), "git describe" command below will still
+      # run and provide some output, although it's definitely not going to be
+      # anything reasonably expected. Bail out early.
+      return unless hash
 
       command = %w(git describe)
       command << '--tags' if match_lightweight
-      command << tag_name
+      command << hash
       Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
--- a/fastlane/spec/actions_specs/changelog_from_git_commits_spec.rb
+++ b/fastlane/spec/actions_specs/changelog_from_git_commits_spec.rb
@@ -6,10 +6,12 @@ describe Fastlane do
           changelog_from_git_commits
         end").runner.execute(:test)
 
-        inner_command = "git describe --tags `git rev-list --tags --max-count=1`"
-        # this is not really the command that would have been executed, but a "fabricated" representation for tests (by Actions.sh) that includes both command that would have been run
-        pseudocommand = "git log --pretty=\"%B\" #{inner_command.shellescape}...HEAD"
-        expect(result).to eq(pseudocommand)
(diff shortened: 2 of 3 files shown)


fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -3,36 +3,38 @@ module Actions
     GIT_MERGE_COMMIT_FILTERING_OPTIONS = [:include_merges, :exclude_merges, :only_include_merges].freeze
 
     def self.git_log_between(pretty_format, from, to, merge_commit_filtering, date_format = nil, ancestry_path)
-      command = ['git log']
-      command << "--pretty=\"#{pretty_format}\""
-      command << "--date=\"#{date_format}\"" if date_format
+      command = %w(git log)
```
janpio: This is equivalent to `['git', 'log']`, correct?

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -3,36 +3,38 @@ module Actions
...
+      command << "--pretty=#{pretty_format}"
+      command << "--date=#{date_format}" if date_format
       command << '--ancestry-path' if ancestry_path
-      command << "#{from.shellescape}...#{to.shellescape}"
+      command << "#{from}...#{to}"
       command << git_log_merge_commit_filtering_option(merge_commit_filtering)
-      Actions.sh(command.compact.join(' '), log: false).chomp
+      Actions.sh(*command.compact, log: false).chomp
```
janpio: What does the `*` indicate here? (Yep, my ruby is rudimentary)

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -3,36 +3,38 @@ module Actions
     GIT_MERGE_COMMIT_FILTERING_OPTIONS = [:include_merges, :exclude_merges, :only_include_merges].freeze
 
     def self.git_log_between(pretty_format, from, to, merge_commit_filtering, date_format = nil, ancestry_path)
-      command = ['git log']
-      command << "--pretty=\"#{pretty_format}\""
-      command << "--date=\"#{date_format}\"" if date_format
+      command = %w(git log)
+      command << "--pretty=#{pretty_format}"
```
janpio: Are we sure `pretty_format` and `date_format` don't have to be quoted? Always?

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -3,36 +3,38 @@ module Actions
...
-      Actions.sh(command.compact.join(' '), log: false).chomp
+      Actions.sh(*command.compact, log: false).chomp
     rescue
       nil
     end
 
     def self.last_git_tag_name(match_lightweight = true, tag_match_pattern = nil)
-      tag_pattern_param = tag_match_pattern ? "=#{tag_match_pattern.shellescape}" : ''
+      tag_pattern_param = tag_match_pattern ? "=#{tag_match_pattern}" : ''
+      tag_name = Actions.sh('git', 'rev-list', "--tags#{tag_pattern_param}",
```
janpio: Should this maybe be extracted to another method for clarity? (Then the command can also be constructed step by step, similar to this current method here.

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -3,36 +3,38 @@ module Actions
...
 
     def self.git_log_between(pretty_format, from, to, merge_commit_filtering, date_format = nil, ancestry_path)
-      command = ['git log']
-      command << "--pretty=\"#{pretty_format}\""
-      command << "--date=\"#{date_format}\"" if date_format
+      command = %w(git log)
+      command << "--pretty=#{pretty_format}"
+      command << "--date=#{date_format}" if date_format
       command << '--ancestry-path' if ancestry_path
-      command << "#{from.shellescape}...#{to.shellescape}"
```
janpio: Where did this `shellescape`ing move to?

janpio:
Thanks @dotdoom for taking the initiative and fixing this! I left some (probably pretty dumb) comments with questions, would be neat if you could through them.

You probably noticed that fastlane's own tests "only" check for the generated command, not that the command would work cross platform on Linux and Windows as well :/ This is a problem, but we don't have real end to end tests (yet) unfortunately.
Could you share the `Fastfile` you used to test this via AppVeyor?

fastlane/lib/fastlane/helper/git_helper.rb
dotdoom: Done. Do you think it needs a separate test?

dotdoom (commented):
Here's a PR to another project that starts executing fastlane git actions on Windows:
https://github.com/dasfoo/delern/pull/988/files (look for "if Gem.win_platform?" which is now commented out).

The output of Git commands is used in "flutter build" command, specifically `--build-number` and `--build-name` flags, greppable in [Travis (OSX)](https://travis-ci.org/dasfoo/delern/builds/511600064) and [AppVeyor (Windows)](https://ci.appveyor.com/project/dotdoom/delern/builds/23373876) build logs.

Regarding tests, since `git` commands are rather hermetic, I believe they may be tested on a real fixture repository.  I might be able to lend a hand here, but it may need a major change to `Actions.sh` framework, e.g. adding a special parameter to this method that says that a command is "test-ready", i.e. should be executed even in test mode. I won't do this without previously discussing it with fastlane team!

dotdoom:
I believe the latest test failure on Ubuntu is unrelated to my changes:

```plain
W: Failed to fetch http://deb.debian.org/debian/dists/jessie-updates/main/binary-amd64/Packages  404  Not Found
```

janpio:
Correct, that seems to be a networking or CDN problem with the Debian mirrors :/ Right now I hope this will fix itself soon.

fastlane/lib/fastlane/helper/git_helper.rb
janpio: If you know about it, yeah ;) Ruby sometimes is not the most discoverable language - but I am of course fine with it.

The reason for switching from `git log` to `git, log` is consistency I guess?

fastlane/lib/fastlane/helper/git_helper.rb
janpio: Ok, so before we used a joined string as a param, not the "vararged" array.

janpio: Is this common enough that most ruby devs will know about this? Otherwise maybe a simple comment above that line?

fastlane/lib/fastlane/helper/git_helper.rb
janpio: Ah, https://github.com/fastlane/fastlane/blob/30be4096f3c9a84c9fee3cf4a73c71a7508313f4/fastlane/lib/fastlane/helper/sh_helper.rb#L121-L129 always felt a bit "magic" to me but now I think I get it: Array as param, String as param, many params (that then can be shelljoined).

fastlane/lib/fastlane/helper/git_helper.rb
janpio: Never hurts - if you ask 👍

janpio:
> Regarding tests, since `git` commands are rather hermetic, I believe they may be tested on a real fixture repository. I might be able to lend a hand here, but it may need a major change to `Actions.sh` framework, e.g. adding a special parameter to this method that says that a command is "test-ready", i.e. should be executed even in test mode. I won't do this without previously discussing it with fastlane team!

That sounds awesome. The `git` actions might not be the most complex ones we have, but they would provide a great template to learn from and apply to other areas. And if we catch some more Windows/Linux bugs with it, that would be ace.

Best create a super small issue (you can remove any issue template of course) and describe how you would go at it, then other members of the team can look at it and comment as well. But I think we never downvoted a improvement in our tests 🙈

fastlane/lib/fastlane/helper/git_helper.rb
dotdoom: Great. If you prefer to rather not have these cryptic constucts, I can also go with plain arrays, and teach rubocop to complain if they are ever used.

Consistency indeed is one of the reasons, but in this case `git log` won't work anymore: as we pass each argument separately, `shelljoin` will consider each array element a **single** argument. This is the power of `shelljoin`, but it also means that `git log` will be shell-escaped and the OS will look for a binary called `git log` (with a space inside).

fastlane/lib/fastlane/helper/git_helper.rb
dotdoom: Exactly.

It's been a while since I've been paid for writing Ruby code, so I can't tell how common it is nowadays. Note that it's also used in `Actions.sh` itself (both argument list and a method call). I added a comment nevertheless, better safe than sorry.

https://github.com/fastlane/fastlane/blob/30be4096f3c9a84c9fee3cf4a73c71a7508313f4/fastlane/lib/fastlane/helper/sh_helper.rb#L11-L13

fastlane/lib/fastlane/helper/git_helper.rb
dotdoom: Right, I think it's trying to be compatible with [IO#popen](https://www.rubydoc.info/stdlib/core/IO.popen).

dotdoom:
Regarding the test for the new method: I think my suggestion was a little hasteful. Since it's an internal interface to Actions and not an actual, er, "action", it's not possible to test it with a conventional test (`Fastfile.parse`). Besides, that code is already covered by existing tests, and a new test will become a "change watchdog" which merely triggers on harmless changes to the code which do not actually affect the external interface.

While looking for similar existing issues I found #13227, which this PR fixes. Also filed #14479 for `Actions.sh` testing.

Looking further into the test failure: apparently [Debian "jessie" has been removed from the mirrors network](https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html). This will need a fix to the [image](https://github.com/fastlane/fastlane/blob/30be4096f3c9a84c9fee3cf4a73c71a7508313f4/.circleci/config.yml#L69). Filed https://github.com/milch/fastlane-docker/issues/2 for that.

fastlane/lib/fastlane/helper/git_helper.rb
janpio: I am the worst benchmark here, as I am really dabbling in a language I haven't really mastered. And `%w(` is google-able enough that this should not be a problem for anyone really.

> but in this case git log won't work anymore:

A good point, I didn't see that.

fastlane/lib/fastlane/helper/git_helper.rb
janpio: Thanks.

`sh_helper.rb` has been written by @jdee whose ruby is pretty advanced sometimes (at least compare to what I produce) ;) But I think we are good here as well now.

janpio:
Thanks for all that.

> Looking further into the test failure: apparently [Debian "jessie" has been removed from the mirrors network](https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html). This will need a fix to the [image](https://github.com/fastlane/fastlane/blob/30be4096f3c9a84c9fee3cf4a73c71a7508313f4/.circleci/config.yml#L69). Filed [milch/fastlane-docker#2](https://github.com/milch/fastlane-docker/issues/2) for that.

❤️ ❤️ ❤️ 

We can merge with this test failing if need be.

ohayon (approved):
This seems legit to me!

dotdoom:
Thank you for your friendliness and a very fast review turnaround!

dotdoom:
I had to force-push to rebase on the latest `master`.  Is there anything else I can do for this PR?
Also just in case GitHub buried my comment about live test too deep to be discoverable:

> Here's a PR to another project that starts executing fastlane git actions on Windows:
> https://github.com/dasfoo/delern/pull/988/files (look for "if Gem.win_platform?" which is now commented out).
> 
> The output of Git commands is used in "flutter build" command, specifically `--build-number` and `--build-name` flags, greppable in [Travis (OSX)](https://travis-ci.org/dasfoo/delern/builds/511600064) and [AppVeyor (Windows)](https://ci.appveyor.com/project/dotdoom/delern/builds/23373876) build logs.

janpio:
Nope, you are good - now we're just waiting for one of the committers to be bold and sure enough to hit the "Merge" button (After adding an "Approve" review). (Won't be me this time as the ruby syntax is too new for me)

dotdoom:
Sounds good.  No pressure, please take your time.  I wanted to be sure it's not blocked on me.

joshdholtz (approved):
This looks really good! The rework is 🔥 Thanks for the contribution!

PR #14478, opened 2019-03-27
[spaceauth] Add new --output option

ened:
### Checklist

### Motivation and Context
For improved compatibility with non-interactive environments, it would be best if the session token (created by fastlane spaceauth -u ...) could be written to a file.

### Description
Added a new option -o/--output FILE to the spaceauth runner.
Using this option will effectively turn the spaceauth session non-interactive.

I could not find existing spec or documentation for the spaceauth command, please advise where to add or document this?


janpio:
Normal usage of `fastlane spaceauth` usually requires you to enter a 2FA token, which obviously doesn't work in a non-interactive session. What is your use case here?

---

`spaceauth` doesn't really have documentation in the common sense right now, it is only mentioned for solving specific problems:
https://docs.fastlane.tools/best-practices/continuous-integration/#spaceauth
https://github.com/fastlane/fastlane/tree/master/spaceship#web-sessions
Suggestion on where we could put proper documentation?

janpio:
(The Ubuntu CI failure is not your fault: https://github.com/fastlane/fastlane/issues/14481)

ened:
My use case is to:
 - Run spaceauth from within a `schedule_beta` lane, which appropriately creates the token and stores it in a tmp file
 - Next part of the lane picks it up and stores the token in a secured value on my CI service
 - The secured value is updated in the ci config file
 - committed
 - uploaded, CI server will then later use the token from the variable to publish to App Store

So it's really about the upper part which is made so much easier if the output is written to a file.

By interactive I also meant the part where it asks whether things should be copied to the clipboard.
There is a phone number selection but I hope that #14436 lands soon so that part can be cleaned up.
Only missing thing then is the PIN code entry, which is hopefully not blocked by the fastlane script (non-interactive etc).

---
For documentation, it would be great if spaceauth lists all options, including the global ones.
There is a README in the spaceship folder which can (or already does) cover the spaceauth command as you mentioned.

I may have missed the `spec` for spaceauth.. should be extended if this PR moves forward.

janpio:
There probably is no spec as well for spaceauth itself. All the spaceship login stuff in general is greatly undertested.

> Run spaceauth from within a schedule_beta lane, which appropriately creates the token and stores it in a tmp file

So you don't expect any prompts (2FA) when running this?

ened:
> > Run spaceauth from within a schedule_beta lane, which appropriately creates the token and stores it in a tmp file
> 
> So you don't expect any prompts (2FA) when running this?

Yes, there would be a prompt and I have not put too much thought in how that is being handled in my fastlane script. I think fastlane will wait properly when run in my interactive shell on the laptop, where this whole spaceauth thing is being executed.

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,19 +51,26 @@ def run
...
+          Open3.popen3('pbcopy') { |input, _, _| input << yaml }
+          puts("Successfully copied text into your clipboard 🎨".green)
+        end
+      end
 
-      if mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
-        require 'open3'
-        Open3.popen3('pbcopy') { |input, _, _| input << yaml }
-        puts("Successfully copied text into your clipboard 🎨".green)
+      unless @output.nil?
```
janpio: lets just make this an `else` of the `if @output.nil?` then

janpio:
Ok, so your plan is to automate the "get a session manually and put it on my CI service" part, right?

spaceship/lib/spaceship/spaceauth_runner.rb
ened: OK

ened:
> Ok, so your plan is to automate the "get a session manually and put it on my CI service" part, right?

Yes.

In other words, it used to be:
 - get spaceauth session
 - if expired
  - press 1 (#14436)
  - possibly wait for SMS, enter PIN code in spaceauth
 - press y (`pbcopy`)
 - open browser @ ci settings page, paste code in "encrypt variable" field, press "encrypt", copy encrypted hash 
 - open ci config file, replace encrypted variable
 - save new ci config file
 - continue deployment (and stop pulling out hair)

And it should be:
 - get spaceauth session
 - if expired
  - press 1 (#14436)
  - enter PIN
 - ... that's it! fastlane script can now do everything itself


I should note that the second flow would also work with the current version, but it requires a LOT of trickery rgd. the 'clipboard' prompt and then another sed/grep script to extract the output from spaceauth, so I thought it would be a LOT easier if spaceauth could simply write the session somewhere.

Also, `stdout`, `stderr` is not really used in this fastlane component. I'd actually prefer if anything interactive is on stderr and stdout would just contain the token. Plus, the exit code for the command itself should be set so any shell script could use it.

janpio:
Your last two paragraphs don't make too much sense to me, but I'm tired and that is ok. Just open new issues (or PRs) if you want to change anything here - as you can see we are open for any improvements.

Adding a way to output the session value into a file directly definitely is a valid improvement, as it would bring us closer to a more streamlined process as you describe it (which is also why I implemented the PR you link to regarding the sms stuff).

(If was just a bit confused by your wording in the beginning, but we worked through that together)

By the way: The "change the environment variable on my CI provider" stuff would probably make a great plugin, or at least custom action. So if that works, please do think about creating one or "publishing" it here via an issue.

ened:
> By the way: The "change the environment variable on my CI provider" stuff would probably make a great plugin, or at least custom action. So if that works, please do think about creating one or "publishing" it here via an issue.

Good idea! I'm using Cirrus - and there seems to be no actions or integrations available, yet. Will come back to it at some point.

janpio:
Still WIP or should we do a proper review and possible merge now?

ened:
@janpio I have not managed to write the unit tests for this, yet. However, can confirm the code is running well for the past couple days (in which I had multiple fast lane deployments)

janpio (approved):

PR #14480, opened 2019-03-27
[action] Post line break characters as actual interpreted line break through slack integration

albertbass86:
### Checklist

### Motivation and Context
Fixes open issue https://github.com/fastlane/fastlane/issues/14141
The issue is when posting to a slack channel with a line break in the message.

### Description
The problem seems to happen because when a message is sent to the slack action, the \n line break character is being parsed as two characters slash and n instead of a line break character. In ruby, this happens when \n is declared with single quotes. The fix simply uses gsub method to replace the '\n' from the input with the interpreted line break character \n, by using double quotes to declare it.

The test has an input with a message that contains a non-interpreted line break '\n' and checks that after the message is sent through the integration the line break is substituted by an interpreted one "\n".

### DIFF
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -15,7 +15,9 @@ module Fastlane
         # We want the last 7000 characters, instead of the first 7000, as the error is at the bottom
         start_index = [message.length - 7000, 0].max
         message = message[start_index..-1]
-        message
+        # We want line breaks to be shown on slack output so we replace
+        # input non-interpreted line break with interpreted line break
+        message.gsub('\n', "\n")
       end
 
       def self.run(options)
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -26,7 +26,7 @@ module Fastlane
         options[:message] = self.trim_message(options[:message].to_s || '')
         options[:message] = Slack::Notifier::Util::LinkFormatter.format(options[:message])
 
-        options[:pretext] = (options[:pretext].gsub('\n', "\n")) unless options[:pretext].nil?
+        options[:pretext] = options[:pretext].gsub('\n', "\n") unless options[:pretext].nil?
 
         if options[:channel].to_s.length > 0
           channel = options[:channel]
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -26,6 +26,8 @@ module Fastlane
         options[:message] = self.trim_message(options[:message].to_s || '')
         options[:message] = Slack::Notifier::Util::LinkFormatter.format(options[:message])
 
+        options[:pretext] = (options[:pretext].gsub('\n', "\n")) unless options[:pretext].nil?
+
         if options[:channel].to_s.length > 0
           channel = options[:channel]
           channel = ('#' + options[:channel]) unless ['#', '@'].include?(channel[0]) # send message to channel by default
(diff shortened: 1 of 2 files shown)


janpio:
What a simple solution!

Have you tested manually as well?

albertbass86:
Yes, I have tested on one slack instance I have, here is a screenshot, sorry I did not include it before.

Screenshot 2019-03-27 at 16 14 21

On a side note, but a bit related...

I have been looking into why the Ubuntu tests had failed. I have seen now you have opened an issue. I was trying to test by using node 8 on Fastfile as you pointed out that was the line to blame. `sh('curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -')`

However I am a bit confused with the system used and I am still working on investigating more with the circleci cli. I digress, I guess I am trying to say if you need help investigating that I would be happy to help.

janpio:
Sorry, should have added to the [Ubuntu CI issue](https://github.com/fastlane/fastlane/issues/14481) that we are aware what is broken and how it probably can be fixed. (If you want to learn more: The third link in the issue has a nice description [super short version: Ubuntu/Debian changed where some files are stored, our Docker image doesn't know that yet, so fails `apt-get update`]).

janpio (approved):
Simple change, added test that seems to be testing what was changed, and even a screenshot of successful manual testing - what more could we wish for? 🍰 

Thanks!

adoankim:
hey here!, I didn't noticed that there was already a PR for this fix and [created](https://github.com/fastlane/fastlane/pull/14499) one, I just closed it 😅.

Just wanted to point out that the issue happens on `pretext` argument too.

joshdholtz:
@albertbass86 Heyyy 👋 This looks great! Do you mind rebasing from the latest master? 🙃 This will fix the failing Ubuntu test 👌

albertbass86:
Hi I did the rebase but I have an issue with macOS tests. CircleCI says I have to select a plan but I cannot do that, no plans appear.

Screenshot 2019-04-02 at 08 07 18

janpio:
Ugh, normally this should have been run on _our_ fastlane Circle CI account where there is a plan configured. I have never seen this happen as here, where somehow the rebase or force push broke the connection to our account. Hmm...

janpio:
While we are looking into this, could you @albertbass86 confirm that everything from https://github.com/fastlane/fastlane/pull/14480#issuecomment-478047205 is also taken care of in your PR? Thanks.

janpio:
@joshdholtz said this might help

albertbass86:
I opened a support ticket with CircleCI. Also added part fix of pretext courtesy of @adoankim.

I did not reuse the gsub method because I thought better to keep the tests for pretext and message separate in case in the future someone has to modify the way pretext is being parsed, but I could do it more like @adoankim 's fix if you want.

joshdholtz:
I have reached out to circle about this as well! 😊 I want to know what's going on.

joshdholtz:
@albertbass86  @janpio 

```
This happens when the contributor is following the fork under their own plan, because we only build a commit once, so if the commit is being built under albertbass86/fastlane, it will not be built again under fastlane/fastlane. Could you ask albertbass86 to unfollow the albertbass86/fastlane project?
```

^ CircleCI

I'm asking them to not do this but if you unfollow this _should_ do it

albertbass86:
Thank you, I am not following anything though!



I clicked unfollow all nevertheless 

--
Albert Donaire

janpio:
(Ok Circle, that doesn't make to much sense if you want people to actually use your service...)

@albertbass86 You might have to push something (or forcepush nothing) to the repo to trigger a rebuild, my closing and re-opening didn't do anything.

albertbass86:
Closing, no way to make circleCI work without closing and opening a new one apparently.

PR #14485, opened 2019-03-27
[pilot] changelog bytesize truncate tests

albertbass86:
### Checklist

### Motivation and Context
Solves issue #14443

### Description
Changes include modifying truncation for using bytesize instead of string length.
Test with fixture added for testing correct bytesize truncation with kanji.
I ran the automated test to ensure old behaviour was not broken but could not manually test because I do not have an Apple developer account with enough permission to do so.
Checked documentation for description of maximum character length, as stated on issue #14443 
Added comment on hardcoded character limit to link to Apple statement on their documentation


pilot/spec/build_manager_spec.rb
```diff
@@ -1,10 +1,18 @@
+require 'active_support'
```
janpio: what is this used for here?

pilot/spec/build_manager_spec.rb
```diff
@@ -1,10 +1,18 @@
...
 describe "Build Manager" do
   describe ".truncate_changelog" do
-    it "Truncates Changelog" do
+    it "Truncates Changelog if it exceeds character size" do
       changelog = File.read("./pilot/spec/fixtures/build_manager/changelog_long")
       changelog = Pilot::BuildManager.truncate_changelog(changelog)
       expect(changelog).to eq(File.read("./pilot/spec/fixtures/build_manager/changelog_long_truncated"))
     end
+    it "Truncates Changelog if it exceeds byte size" do
+      changelog = File.read("./pilot/spec/fixtures/build_manager/changelog_bytes_long")
```
janpio: those might have to be `File.binread` for Windows compat - but not 100% sure

pilot/spec/build_manager_spec.rb
albertbass86: This surprised me but if I had:

`require 'active_support/core_ext'`

instead of

```
require 'active_support'
require 'active_support/core_ext'
```

I got this error:

`NameError: uninitialized constant ActiveSupport::Autoload`

By looking into it I found a couple of related stack overflow posts:

https://stackoverflow.com/questions/24303856/rails-uninitialised-constant-activesupportautoload-nameerror#24304046

https://stackoverflow.com/questions/4429068/getting-an-uninitialized-constant-error-with-rspec-have-no-idea-whats-causing

And I was not sure which solution to use to force autoload. Now after you pointed it out maybe this one makes more sense: https://stackoverflow.com/a/32271976

But I am not a ruby expert and I am all for knowing what you all think is the best workaround. There's probably a simpler way.

janpio:
Huh, where did Circle CI go for this PR? Strange, completely missing.

pilot/spec/build_manager_spec.rb
```diff
@@ -1,10 +1,18 @@
+require 'active_support'
+require 'active_support/core_ext'
```
janpio: What extensions do we actually use and need of these?
Why only in the tests?

albertbass86: Oops! You are absolutely right, these should not go in the tests. The reason why it is included is for the mb_chars method that is used in build_manager.rb (not in the spec). It is used to truncate the changelog string by bytes and not characters.

So I should actually move the require to build_manager.rb. It was a mistake adding it to the spec but it worked because I launched the test and thus the require was loaded from the spec when using the actual method.

pilot/lib/pilot/build_manager.rb
```diff
@@ -159,12 +159,12 @@ def update_beta_app_meta(options, build)
...
-      max_changelog_length = 4000
-      if changelog && changelog.length > max_changelog_length
-        original_length = changelog.length
+      max_changelog_bytesize = 4000
+      if changelog && changelog.bytesize > max_changelog_bytesize
+        original_bytesize = changelog.bytesize
         bottom_message = "..."
-        changelog = "#{changelog[0...max_changelog_length - bottom_message.length]}#{bottom_message}"
-        UI.important("Changelog has been truncated since it exceeds Apple's #{max_changelog_length} character limit. It currently contains #{original_length} characters.")
+        changelog = "#{changelog.mb_chars.limit(max_changelog_bytesize - bottom_message.bytesize)}#{bottom_message}"
```
janpio: Ah, here it is, `mb_chars`.

If in any way possible we would want to avoid having to add a dependency for this.

(As soon as encodings and similar are involved, I am totally useless - so I am no help here unfortunately)

albertbass86: I understand. Perfect, will continue from here. Thank you for the review!

pilot/spec/build_manager_spec.rb
albertbass86: Looking into how to substitute active support dependency I stumbled accross their docs. It explains the reasoning behind the (I don't know the name of this, so I will say), lazy requiring.

https://edgeguides.rubyonrails.org/active_support_core_extensions.html

pilot/spec/build_manager_spec.rb
albertbass86: Can someone else confirm if this is the case? I do not have a Windows machine to test and all resources I have found speak about reading images, not text. I understand the text has special characters though so it might be needed, any further info?

janpio: I was just going on the failing AppVeyor test, which is running on a Windows machine:

```
  1) Build Manager .truncate_changelog Truncates Changelog if it exceeds byte size
     Failure/Error: expect(changelog).to eq(File.read("./pilot/spec/fixtures/build_manager/changelog_bytes_long_truncated"))
       expected: "\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6...\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91\xE6\x88\x91..."
            got: "\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u62...\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211\u6211..."
       (compared using ==)
     # ./pilot/spec/build_manager_spec.rb:14:in `block (3 levels) in <top (required)>'
Finished in 2 minutes 31.9 seconds (files took 5.61 seconds to load)
```

So there is definitely something fishy going on there. In the past `binread` was sometimes the solution (although you are right, mostly for binary files) _or_ setting the encoding when reading/writing files (probably to utf-8 in this case) as on Windows ruby's defaults are sometimes different. (So maybe this is an alternative solution that might fix this as well.)

albertbass86: Thank you, that is actually also helpful for the solution I am working on, since I have to enforce encoding to UTF-8 so I will have to explore side effects of that enforcing.

pilot/lib/pilot/build_manager.rb
```diff
@@ -159,12 +159,18 @@ def update_beta_app_meta(options, build)
...
+        original_bytesize = changelog.bytesize
         bottom_message = "..."
-        changelog = "#{changelog[0...max_changelog_length - bottom_message.length]}#{bottom_message}"
-        UI.important("Changelog has been truncated since it exceeds Apple's #{max_changelog_length} character limit. It currently contains #{original_length} characters.")
+        truncated_bytes = changelog.bytes[0...max_changelog_bytesize - bottom_message.length]
+        # We convert the bytes back to a string, which can contained invalid bytes that
+        # do not form a character
+        unscrubbed_truncated_string = truncated_bytes.pack("C*").force_encoding("UTF-8")
+        changelog = "#{unscrubbed_truncated_string.scrub('')}#{bottom_message}"
+        UI.important("Changelog has been truncated since it exceeds Apple's #{max_changelog_bytesize} character limit. It currently contains #{original_bytesize} characters.")
```
joshdholtz: Should we change `character limit` and `characters` to `byte limit` and `bytes?
```suggestion
        UI.important("Changelog has been truncated since it exceeds Apple's #{max_changelog_bytesize} byte limit. It currently contains #{original_bytesize} bytes.")
```

pilot/lib/pilot/build_manager.rb
```diff
@@ -159,12 +159,18 @@ def update_beta_app_meta(options, build)
     end
 
     def self.truncate_changelog(changelog)
-      max_changelog_length = 4000
-      if changelog && changelog.length > max_changelog_length
-        original_length = changelog.length
+      # Apple documentation on "What's New in this Version" field, the one used for changelog
+      # https://help.apple.com/app-store-connect/?lang=en#/devf29afbb74
```
joshdholtz: The documentation actually does say ` Limited to 4000 characters.`. Do we want to make a note there that the documentation is wrong and it should really be bytes?

joshdholtz (changes requested):
This looks really good! Just a few small question on wordings 🙃

pilot/lib/pilot/build_manager.rb
albertbass86: Hi, My first commit before actually finding the documentation had changed character to byte because I felt it was more correct, given that I considered a kanji to be a character. However after reading Apple docs and investigating Unicode I saw the correct term is actually "multibyte character" so I put it back.

I believe your comment is about my original commit, not the latest.

joshdholtz: I was actually talking about the message that gets displayed to the user on your latest commit. If someone has 4000 kanji and it gets truncated someone might be confused by since its actually checking against 4000 bytes and not 4000 kanji characters. But the odds of that happening are probably pretty low so I'm okay with this 😊

joshdholtz (commented):
I like this change! If we can get the tests passing and that commented added (https://github.com/fastlane/fastlane/pull/14485/files#r270671081) I'd be ready to merge this!

pilot/lib/pilot/build_manager.rb
albertbass86: Ah sorry, I had confused the suggested change with the change I had already done.

janpio:
Hm, this is now also using only your Circle CI account. Strange!?

albertbass86:
> Hm, this is now also using only your Circle CI account. Strange!?

Yes, I had created an account. Have you opened a support request with CircleCI about this or should I?

janpio:
Please go ahead and see if they can help.

janpio:
The code was updated to use the correct length in the meantime, but not tests were added - so taking them from this PR was a good solution.

Thanks @albertbass86!

PR #14278, opened 2019-02-20
[pilot] Add support for using App Store Connect API keys for iTunes Transporter

UnsafePointer:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Fixes https://github.com/fastlane/fastlane/issues/14244

### Description

This PR has the following changes:

* Adds api_issuer and api_key to the Appfile configuration.
* Adds api_issuer and api_key options to both Deliver and Pilot.
* Updates `FastlaneCore#ItunesTransporter` to support usage of the App Store Connect API keys.

PR #14486, opened 2019-03-28
[pilot] Add support for using App Store Connect API keys for iTunes Transporter

### LINKED ISSUE #14244: Support API keys for iTMSTransporter
### Feature Request

As a result of [this conversation](https://github.com/fastlane/fastlane/issues/14239#issuecomment-463801382), it was brought up that using API keys for the transporter part of fastlane may be useful.

Required 2FA by apple means that users have to once a month generate a session token to use for their CI. Fastlane provides a wrapper around transporter to upload builds to testflight, so you _should_ be able to use api keys with transporter to avoid that problem for this feature of fastlane.

janpio:
Recreation of #14278 by @Ruenzuo:

> ### Motivation and Context
> 
> Fixes #14244
> ### Description
> 
> This PR has the following changes:
> 
> * Adds api_issuer and api_key to the Appfile configuration.
>  * Adds api_issuer and api_key options to both Deliver and Pilot.
>  * Updates `FastlaneCore#ItunesTransporter` to support usage of the App Store Connect API keys.

Also merge conflicts resolved and some tiny changes (comments, formatting)

Original author is credited in commit via `Co-authored-by: Ruenzuo <Ruenzuo@users.noreply.github.com>`


Ruenzuo:
I'm okay with this.

janpio:
Haha yeah, @googlebot is a bit skeptical and hard to convince - we know you agree in your original PR and that would have been fine, but thanks for repeating this here. 

Just waiting for a review from someone of the team now to be able to merge this.

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -353,7 +353,7 @@ def self.hide_transporter_output?
     #                            see: https://github.com/fastlane/fastlane/issues/1524#issuecomment-196370628
     #                            for more information about how to use the iTMSTransporter to list your provider
     #                            short names
-    def initialize(user = nil, password = nil, use_shell_script = false, provider_short_name = nil)
+    def initialize(user = nil, password = nil, use_shell_script = false, provider_short_name = nil, using_api_key = false)
```
snatchev: Overall, it looks pretty good. I am not a huge fan on adding another parameter to the initializer tho. Perhaps we can add a new class method, something like `ItunesTransporter.using_api_key(key: '', issuer: '')` that would call the original initializer and set the `using_api_key` attribute.

pilot/lib/pilot/build_manager.rb
```diff
@@ -215,16 +215,27 @@ def should_update_localized_build_information?(options)
     # If itc_provider was explicitly specified, use it.
     # If there are multiple teams, infer the provider from the selected team name.
     # If there are fewer than two teams, don't infer the provider.
+    # duplicated in deliver\lib\deliver\runner.rb
```
snatchev: Why is this duplicated? Can it be extracted into a static method somewhere?

snatchev (commented):
Just a few questions.

fastlane_core/lib/fastlane_core/itunes_transporter.rb
janpio: Code is from @Ruenzuo, maybe they can say something about that?

Looking at the code myself, I think this is an easier option of adding this to the existing code. `user` and `pass` are repurposed, so nothing else really has to be changed in between.

pilot/lib/pilot/build_manager.rb
janpio: Was that way in the code, but as I only reopened a PR for changes that @Ruenzuo did I only added this note so this relationship becomes a bit more obvious. One could of course create an issue so someone else can fix this in the future. (Meaning: I think out of scope for this PR)

readefries:
Any update on this issue? This would solve my troubles with the current implementation to authenticate with a username, password and session file, as the CI's run in another region as I live and I cannot use session based solutions.

janpio:
Ping @UnsafePointer

UnsafePointer:
@janpio nothing more to add to my previous comment on this: https://github.com/fastlane/fastlane/pull/14486#issuecomment-477669173, as for the review, I can't take ownership of this as I don't own any Apple device anymore.

Dinsen:
@janpio I fixed the merge conflict in a fork. Should I create a pull request into this branch or how can we go on with this feature?

PR #14490, opened 2019-03-28
[deliver] Do not set ratings_resets to null if not specified

carlosefonseca:
### Checklist

### Motivation and Context and Description
I use Deliver as a library and, if I don't specify `options[:reset_ratings] = true|false` when running `Deliver::UploadMetadata.new.upload(options)`, it changes the value to `null` and then fails to upload because `null` is not valid.
Most parameters in the `upload` method are only set if they exist in the `options` hash, but this one missed the check and it kept breaking my stuff.

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -130,7 +130,7 @@ module Deliver
       set_trade_representative_contact_information(v, options)
       set_review_information(v, options)
       set_app_rating(v, options)
-      v.ratings_reset = options[:reset_ratings]
+      v.ratings_reset = options[:reset_ratings] unless options[:reset_ratings].nil?
 
       Helper.show_loading_indicator("Uploading metadata to App Store Connect")
       v.save!


janpio:
Interesting, in the option definition this has a `default: false` which is why this probably was not checked before: https://github.com/fastlane/fastlane/blob/3fe3a7e91527f79512bf095519604a1132bd88b3/deliver/lib/deliver/options.rb#L169-L173

I am curious though: Can you explain your usecase of Deliver a bit more? Always interesting to see how people use fastlane (or parts of it).

janpio (approved):

carlosefonseca:
@janpio hum… actually, in this instance, I'm not going through the `FastlaneCore::Configuration.create`, like I am in other places. Maybe that would have helped fix the issue. Thanks for pointing that out.

We publish template apps and have a centralised back office for each app store data and have ruby scripts to generate and publish the customised apps and we invoke several Fastlane tools within using the `FastlaneCore::Configuration.create`. I think I skipped that for Deliver because it was better to just give it the data to upload instead of telling it to fetch from a file. I wasn't aware there were configs for it. Maybe that would have helped fix the issue. Thanks for pointing that out.

joshdholtz (approved):
Seems good to me 😊 Thanks for the contribution!

PR #14492, opened 2019-03-28
[scan] Add env_name values for commonly used options

jaleksynas:
### Checklist

### Motivation and Context
I often have to manually create environment variables and copy the values when configuring these options. This PR makes these three options behave like standard options by supporting an env_name.
Example....
```
scan(
       skip_slack: ENV['SCAN_SKIP_SLACK'],
       code_coverage: ENV['SCAN_CODE_COVERAGE'],
       slack_only_on_failure: ENV['SCAN_SLACK_ONLY_ON_FAILURE'],
)
```

### Description
* Add three `env_name` settings
* Correct spacing slightly to match other options.
* Ran all the automated tests


heidiproske (approved):

scan/lib/scan/options.rb
```diff
@@ -130,6 +130,7 @@ def self.available_options
                                      is_string: false,
                                      default_value: false),
         FastlaneCore::ConfigItem.new(key: :code_coverage,
+                                     env_name: "SCAN_GENERATE_CODE_COVERAGE",
```
janpio: If there is no deeper reason for the naming I am missing, I suspect this could/should be `SCAN_CODE_COVERAGE` as the `env_name` usually seems to be toolname + key name.

jaleksynas: Heh...  I created it by reading the description instead of the key.  I'll make the change.

joshdholtz (approved):
Thanks for adding these ❤️ Not sure why these weren't here before 🙃

janpio:
I think in the past those might have been added only selectively to the options where they seem to be needed - not on all as it is mostly done recently.

(Interesting point for our ShareValues/ENV project thingie @mollyIV)

PR #14498, opened 2019-03-29
[snapshot] Make the `sleep` waiting for animations optional

xavierLowmiller:
### Checklist

### Motivation and Context

We recently started running our snapshot tests without animations (`UIView.setAnimationsEnabled(false)`).

In the `SnapshotHelper` there is a `sleep` that waits for animations:

```swift
sleep(1) // Waiting for the animation to be finished (kind of)
```

When we remove this line, our screenshot UI tests speed up by **75%**

### Description

This request adds the option `animationsEnabled` to the `setupSnapshot` function.
If set to `false`, the snapshot helper will not sleep.

Since it has a default argument that make the helper behave as before, this is not a breaking change.

### Open Questions

- [ ] Should I document this new option somewhere?
- [ ] Is there a way to derive this value from `UIView.areAnimationsEnabled` automatically?

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/capture_ios_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/capture_ios_screenshots.md
@@ -349,6 +349,12 @@ if ProcessInfo().arguments.contains("SKIP_ANIMATIONS") {
 }
 ```
 
+This requires you to pass the launch argument like so:
+
+```ruby
+snapshot(launch_arguments: ["SKIP_ANIMATIONS"])
+```
+
 By default, _snapshot_ will wait for a short time for the animations to finish.
 If you're skipping the animations, this is wait time is unnecessary and can be skipped:
 
--- a/fastlane/lib/fastlane/actions/docs/capture_ios_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/capture_ios_screenshots.md
@@ -336,3 +336,22 @@ When the app dies directly after the application is launched there might be 2 pr
 ## Determine language
 
 To detect the currently used localization in your tests, access the `deviceLanguage` variable from `SnapshotHelper.swift`.
+
+## Speed up snapshots
+
+A lot of time in UI tests is spent waiting for animations.
+
+You can disable `UIView` animations in your app to make the tests faster:
+
+```swift
+if ProcessInfo().arguments.contains("SKIP_ANIMATIONS") {
+    UIView.setAnimationsEnabled(false)
+}
+```
+
+By default, `snapshot` will wait for a short time for the animations to finish.
+If you're skipping the animations, this is wait time is unnecessary and can be skipped:
+
+```swift
+setupSnapshot(app, waitForAnimations: false)
+```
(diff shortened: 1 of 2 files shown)


xavierLowmiller:
I signed it!

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -18,8 +18,8 @@ import XCTest
 var deviceLanguage = ""
 var locale = ""
 
-func setupSnapshot(_ app: XCUIApplication) {
-    Snapshot.setupSnapshot(app)
+func setupSnapshot(_ app: XCUIApplication, animationsEnabled: Bool = true) {
```
joshdholtz: Can we rename this to something like `waitForAnimations`? The current name seems like it will enable/disable all animations in the app 🙃

joshdholtz (changes requested):
One small request on the naming! And if you want to also add some documentation to the tips section of this page that would be 💯 - https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/capture_ios_screenshots.md#tips

snapshot/lib/assets/SnapshotHelper.swift
xavierLowmiller: Sure 👌

joshdholtz:
@xavierLowmiller Thanks for making that change! And version bumping this file 😉 I will test this out now 💪

xavierLowmiller:
Thanks for reviewing and pointing me to the documentation!

fastlane/lib/fastlane/actions/docs/capture_ios_screenshots.md
```diff
@@ -336,3 +336,22 @@ When the app dies directly after the application is launched there might be 2 pr
...
 To detect the currently used localization in your tests, access the `deviceLanguage` variable from `SnapshotHelper.swift`.
+
+## Speed up snapshots
+
+A lot of time in UI tests is spent waiting for animations.
+
+You can disable `UIView` animations in your app to make the tests faster:
+
+```swift
+if ProcessInfo().arguments.contains("SKIP_ANIMATIONS") {
```
joshdholtz: @xavierLowmiller Do we want to mention about putting this value in `launch_arguments`? Otherwise users might assume this is built in by default 🤔

xavierLowmiller: Yes, good point.

The launch arguments can be passed
* via ruby: `snapshot(launch_arguments: ["SKIP_ANIMATIONS"])`
* via Swift: `app.launchArguments = ["SKIP_ANIMATIONS")`

Which should I use in the documentation?

joshdholtz: Yeah yeah, if you want to put the ruby one in the doc that would be 🔥 And then we can merge this!

joshdholtz (approved):
This is 💯 Thanks for the contribution ❤️

PR #14501, opened 2019-03-30
[action] Continue git_commit action execution if there is nothing to commit

### LINKED ISSUE #14437: git_commit - continue execution if there is nothing to commit
### Question Checklist

### Question Subject

**git_commit** action

#### Question Description

I use the [git_commit action][1]:

> git_commit(path: "*", message: "message")

This action fails if there is nothing to commit. How can I continue executing my Fastlane script even if there is nothing to commit?

What I'm looking for could be accomplished with the any of the two options:

1. Add some parameter to the `git_commit` action so it would not crash the script if there is nothing to commit (but rather, continue)
2. Check if there are any changes and only then invoke `git_commit`.

I didn't find any way to get this done using method 1, so **I'm looking for an implementation with the second approach.**

Also, is it possible to add any flag to continue execution, or use a similar action:

```Bash
git_commit_if_needed.....
```

  [1]: https://docs.fastlane.tools/actions/git_commit/

mollyIV:
### Checklist

### Motivation and Context

[`git_commit`](https://docs.fastlane.tools/actions/git_commit/) action fails if there's nothing to commit. 

git_commit_nothing_to_commit

closes https://github.com/fastlane/fastlane/issues/14437

### Description

Added a new parameter `allow_nothing_to_commit`. 

Tested the changes by adding a new unit test and doing manual testing. 

Screenshot 2019-03-30 at 19 27 40

Screenshot 2019-03-30 at 19 27 03

### DIFF
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -10,17 +10,15 @@ module Fastlane
 
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
-        command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
-        nothing_to_commit = false
-
         if params[:allow_nothing_to_commit]
           repo_clean = Actions.sh("git status --porcelain").empty?
           UI.success("Nothing to commit, working tree clean ✅.") if repo_clean
           return if repo_clean
         end
 
+        command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
         result = Actions.sh(command)
-        UI.success("Successfully committed \"#{params[:path]}\" 💾.") unless nothing_to_commit
+        UI.success("Successfully committed \"#{params[:path]}\" 💾.")
         return result
       end
 
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -10,8 +10,23 @@ module Fastlane
 
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
-        result = Actions.sh("git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip)
-        UI.success("Successfully committed \"#{params[:path]}\" 💾.")
+        command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
+        nothing_to_commit = false
+
+        if params[:allow_nothing_to_commit]
+          result = Actions.sh(
+            command,
+            log: FastlaneCore::Globals.verbose?,
+            error_callback: lambda { |_|
+              nothing_to_commit = true
+              UI.success("Nothing to commit, working tree clean ✅.")
+            }
+          )
+        else
+          result = Actions.sh(command)
+        end
+
+        UI.success("Successfully committed \"#{params[:path]}\" 💾.") unless nothing_to_commit
         return result
       end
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -10,8 +10,23 @@ def self.run(params)
...
-        result = Actions.sh("git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip)
-        UI.success("Successfully committed \"#{params[:path]}\" 💾.")
+        command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
+        nothing_to_commit = false
+
+        if params[:allow_nothing_to_commit]
+          result = Actions.sh(
+            command,
+            log: FastlaneCore::Globals.verbose?,
+            error_callback: lambda { |_|
```
janpio: Is "nothing to commit" the only reason this can error?

joshdholtz: This catching of error callback might be a bit to aggressive. It _will_ allow nothing to get committed but it will also eat all/any other errors as well 🤔 Might need to look into another solution 🙃

joshdholtz (changes requested):

mollyIV:
👋 @janpio and @joshdholtz 

Thanks for your questions / suggestions. I updated the implementation - now we're checking if there's something to commit via `git status --porcelain`. If the git index is clean, there's no point of committing changes, so we just early return. 

Added one more unit tests 💪

Let me know what you think 🙇

fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -10,8 +10,17 @@ def self.run(params)
 
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
-        result = Actions.sh("git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip)
-        UI.success("Successfully committed \"#{params[:path]}\" 💾.")
+        command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
+        nothing_to_commit = false
```
joshdholtz: I don't think we need this. The if statement is early returning out of the method don't need this 😊

fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -10,8 +10,17 @@ def self.run(params)
...
+
+        if params[:allow_nothing_to_commit]
+          repo_clean = Actions.sh("git status --porcelain").empty?
+          UI.success("Nothing to commit, working tree clean ✅.") if repo_clean
+          return if repo_clean
+        end
+
+        result = Actions.sh(command)
+        UI.success("Successfully committed \"#{params[:path]}\" 💾.") unless nothing_to_commit
```
joshdholtz: Can remove the `unless nothing_to_commit` since the clean repo is early returning 💣

fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -10,8 +10,17 @@ def self.run(params)
 
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
-        result = Actions.sh("git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip)
-        UI.success("Successfully committed \"#{params[:path]}\" 💾.")
+        command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
```
joshdholtz: We could probably move this line after the `if params[:allow_nothing_to_commit]` block since this line isn't needed until after that 😊

joshdholtz (changes requested):
This implementation looks much better! Just a few minor code nitpicks 😊

mollyIV:
👋 @joshdholtz 

PR is ready for another round, thanks for your comments! 💪

joshdholtz (approved):
This is 🔥 Thanks again for another contribution! ❤️

PR #14502, opened 2019-03-31
DO NOT MERGE: Demo that some tests never run

dotdoom:
This is a demo for #14503. Please do not merge.


dotdoom:
Okay, the linter complains but `rspec` passes :smiley_cat:

PR #14503, opened 2019-03-31
spec_helper refactoring and fixing Scan tests

dotdoom:
### Checklist

### Motivation and Context

* some methods (like `with_env_values`) are defined in more than one `spec_helper.rb`, with the same signatures and functionality
* they also have different behavior. For example, the method below does not accept any block (contrary to what the comment says), which means that the block passed in is ignored (**sic!**), and any test code inside the block does not get executed. See #14502 for a demo.

https://github.com/fastlane/fastlane/blob/875e3482906044a4f9806dc869c8813d3a2e9a57/fastlane_core/spec/spec_helper.rb#L36-L43

* `stubbing.rb` does not seem to add any value, so I simply moved its content into `spec_helper.rb` to have less entities. If the team feels strongly about having it as a separate file, I will delete that commit from this PR
* from several implementations of `with_...` methods (some more permissive than others) I picked / reimplemented with the following principles:
  - always take a block (this is a common convention for `with_...` methods in rspec. Ruby developers would expect it to take a block). Despite `stub_const` looks very pretty to use, it does not allow to limit scope (the scope always equals that of the whole `it` block) and requires a different naming convention (ex.: `stub_env_values`)
  - complain if the method is called without block (dropped unused feature to "permanently change environment variable" because it's not useful - `ENV['MYVAR']` does the same).


dotdoom:
Uh-oh, I've no idea what happened on AppVeyor. Time to dust off my virtualbox and [download some Windows images](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/).

dotdoom:
The tests failing are those which were not running before the fix. See #14502 which "breaks" them but they keep passing.. Looks like they are severely out of date and can be removed. WDYT?

fastlane.gemspec
```diff
@@ -117,4 +117,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency('fakefs', '~> 0.8.1')
   spec.add_development_dependency('sinatra', '~> 1.4.8')
   spec.add_development_dependency('xcov', '~> 1.4.1') # Used for xcov's parameters generation: https://github.com/fastlane/fastlane/pull/12416
+  spec.add_development_dependency('climate_control', '~> 0.2.0')
```
janpio: development dependency means that this in no way influences what we deliver to end users, right?

fastlane/spec/spec_helper_spec.rb
```diff
@@ -23,16 +23,6 @@
       end
       expect(ARGV).to eq(current_ARGV)
     end
-
-    it "sets ARGV for good if no block is given" do
```
janpio: Is this another feature that was dropped and not mentioned in the PR description?

janpio:
Ugh, I had looked at that `with_env_values` a bit ago and wondered how it works - but assumed I must be missing some magic here. How did you discover this?

Left some comments.

dotdoom (commented):
Hey, glad this sorts things out! I was actually fixing a Windows-specific issue in documentation generator and while writing a test, I needed to temporarily change environment.

I think the tests were written for `Slack::Notifier`, which no longer exists, and instead has `SlackAction`. Here's the PR that would break those tests, should they have run by that time: #13791.

janpio:
Oh, pretty recent that test even. Well, I think nobody will miss those tests then. Bye 👋

dotdoom:
After sitting for a while staring at the code over the weekend, I decided that it's worth fixing these tests. PTAL?

dotdoom:
Is there anything else I can do for this PR?

janpio:
No, I think we are just waiting for someone with more ruby chops to come along from the team and leave an "Approve" review. I don't fully trust myself here.

dotdoom:
Hey there friendly Fastlane team, may I humbly request some feedback for this PR? It's blocking some of the further contributions I have in queue.

janpio (approved):
Look good from my side, but needs some more eyes from someone with more ruby.

(Pinged the appropriate people in the background @dotdoom)

joshdholtz (approved):
This is soooooo 💯 Thanks for taking the time to clean up our tests ❤️ ❤️ ❤️

Issue #13616, opened 2018-10-31
Missing changelog for supply Android bundle to internal track


multiholle:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Fastlane supply won't upload the changelog for a Android bundle (aab) to the internal track.

**Before upload, no releases, no binaries**
<img width="400" alt="screen shot 2018-10-31 at 20 07 03" src="https://user-images.githubusercontent.com/5654539/47813128-66fd6100-dd4a-11e8-8090-3ddf78a1bc92.png">
<img width="400" alt="screen shot 2018-10-31 at 20 07 26" src="https://user-images.githubusercontent.com/5654539/47813130-66fd6100-dd4a-11e8-8023-70f3c43634ce.png">

**Changelog file in fastlane folder**
<img width="400" alt="screen shot 2018-10-31 at 20 19 44" src="https://user-images.githubusercontent.com/5654539/47813131-6795f780-dd4a-11e8-9eb9-091198f88044.png">

**Created release after upload, missing changelog**
<img width="400" alt="screen shot 2018-10-31 at 20 25 30" src="https://user-images.githubusercontent.com/5654539/47813352-22be9080-dd4b-11e8-8410-d6c48627a5df.png">

<details><summary>fastfile</summary>
  <pre>
default_platform(:android)

platform :android do
  desc "Submit a new Beta Build to Crashlytics Beta"
  lane :internal do
    keystore_password = prompt(text: "Keystore Password: ", secure_text: true)

    gradle(
      task: "clean bundle",
      build_type: "release",
      properties: {
        "keystorePassword": keystore_password 
      }
    )

    supply(
      track: 'internal'
    )
  end
end
</pre>
</details>

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>$ fastlane internal
[✔] 🚀 
[20:09:54]: fastlane detected a Gemfile in the current directory
[20:09:54]: however it seems like you don't use `bundle exec`
[20:09:54]: to launch fastlane faster, please use
[20:09:54]: 
[20:09:54]: $ bundle exec fastlane internal
[20:09:54]: 
[...]

PR #14511, opened 2019-04-02
[supply] Also upload changelog when uploading bundle

bed42:
### Checklist

### Motivation and Context
This PR addresses the issue documented at https://github.com/fastlane/fastlane/issues/13616

New android bundle updates would not upload the changenotes for the new version.

### Description
After uploading an apk, the existing code checks for new changenote metadata from the uploaded version and immediately uploads it. This was not occurring when uploading app bundles. This fix does the same thing for bundles, as it is doing for apks in the upload_binary_data(apk_path) function.

changes have been tested on our android code base, by utilising my local modified version of fastlane as documented at https://github.com/fastlane/fastlane/blob/master/Testing.md#test-your-local-fastlane-code-base-with-your-setup

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -162,7 +162,7 @@ module Supply
 
       aab_paths.each do |aab_path|
         UI.message("Preparing aab at path '#{aab_path}' for upload...")
-        bundle_version_code = client.upload_bundle(aab_path) 
+        bundle_version_code = client.upload_bundle(aab_path)
 
         if metadata_path
           all_languages.each do |language|
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -162,7 +162,16 @@ module Supply
 
       aab_paths.each do |aab_path|
         UI.message("Preparing aab at path '#{aab_path}' for upload...")
-        aab_version_codes.push(client.upload_bundle(aab_path))
+        bundle_version_code = client.upload_bundle(aab_path) 
+
+        if metadata_path
+          all_languages.each do |language|
+            next if language.start_with?('.') # e.g. . or .. or hidden folders
+            upload_changelog(language, bundle_version_code)
+          end
+        end
+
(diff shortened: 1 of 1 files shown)


bed42:
I signed it!

janpio:
Awesome catch!

Would it make sense to extract the shared logic to a `upload_changelog_for_all_languages` (or similarly named) method? Or is that overkill?

bed42:
I did consider that, but was weary of getting into confusing naming conflicts related the changelogs.. there's already def upload_changelogs(language) - which doesn't take an explicit version number, so didn't want to cause confusion (especially as this is my first tinkering in this code base (and in ruby in general). Seemed like a rabbit hole for an otherwise quick fix atm ;)

janpio:
Valid point.

What is the quickest way to create one of these aab files / bundles for testing this? (Windows, Android Studio installed)

bed42:
The quickest might be to grab this example code https://www.raywenderlich.com/9043-android-app-bundles-getting-started - though of course you'll need to change the bundle ID and create a dummy app in your play store account. 

You could otherwise try just building a bundle of any existing code directly out of Android Studio ie https://mobikul.com/android-app-bundles/ - if that works then the `gradle(task: "bundleRelease")` fastlane command should also work.

janpio:
They sure don't make it easy to just test this with an existing app:

> Allowing Google Play to manage your app signing key is a requirement to use the Android App Bundle, and to benefit from Google Play’s Dynamic Delivery.

bed42:
anything I can do to help this along?

AndreasBoehm:
Could we get some help from you on that one @KrauseFx? Thanks

KrauseFx:
@joshdholtz

janpio (approved):
I dislike how `upload_bundles` and `upload_apks` (which internally uses `upload_binary_data` to do the actual upload and also metadata) are different, but this is not really in scope for this PR.

So this looks good to merge! Thanks @bed42 🚀

Issue #13096, opened 2018-08-13
Automatically upload DSYM to Firebase/Crashlytics


taglia3:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I want to automatically upload the .DSYM file to Firebase/Crashlytics.

My GoogleService-Info.plist is located at _"${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.app/GoogleService-Info.plist"_

I've a RunScript Phase with:
`"${PODS_ROOT}/Fabric/run"`

But it doesn't upload the .DSYM file to the Firebase portal. Why?
Can I resolve by using the _upload_symbols_to_crashlytics_ lane?
How can I configure the gsp_path?

taglia3:
Fixed by simply using _upload_symbols_to_crashlytics(gsp_path: gsp_path)_ with _gsp_path_ correctly configured.

MauriceArikoglu:
@taglia3 would you be so kind as to show an example of a configured gsp_path? Or point to the direction where one could find details on configuration? Thanks

mmdock:
I am also struggling to get this to work.  I have a lane that looks like the following:
```
  desc "Fetch and upload dSYM files to Firebase Crashlytics"
  lane :refresh_dsyms_firebase do |options|
    download_dsyms(version: options[:version])        
    upload_symbols_to_crashlytics(gsp_path: "./App/GoogleService-Info.plist") 
    clean_build_artifacts
  end
```


After that I see:

[17:22:47]: invalid byte sequence in UTF-8
[17:22:47]: invalid byte sequence in UTF-8
[17:22:47]: invalid byte sequence in UTF-8

and one of these for every dSYM file:
[17:22:48]: Shell command exited with exit status  instead of 0.
[17:22:48]: Uploading '70DBE65E-227E-3754-89F2-EEFA6B8EEC2F.dSYM'...

leonirqranio:
What is the solution?

MauriceArikoglu:
@mmdock I cannot help you. Struggling with this myself. I'll update in case I find a solution

eli-symbolab:
Also experiencing this problem.
[...]

PR #14512, opened 2019-04-02
[action] prioritize GSP path over fabric token

mmdock:
### Checklist

### Motivation and Context

GSP path should be prioritized over the Fabric token.  It is required because users who migrated from Fabric to Firebase Crashlytics will still have the Fabric token setup as well as the GSP path, and currently it will not upload to Firebase Crashlytics if Fabric is also setup.  Should prioritize Firebase for smoother transition for when Fabric is fully shut down in a few months for users with dual configs atm. TL;DR: if both are setup, dSYMS get sent to deprecated service. 

Not and open issue, but one that was mistakenly closed and there have been active conversation on: https://github.com/fastlane/fastlane/issues/13096 ... the issue under discussion, despite it being closed last year was only resolved 11 days ago, but not entirely.
### Description

Search for the gsp path _first_ before fabric token.  Also, don't set the `-a` flag if gsp is already set.  

I have a product with both a gsp path and fabric token.  Tested on latest Fastlane first to show that the `-a` flag was being used for the `upload-symbols` binary, despite the existence of a GoogleService-Info.plist in my project.  With the changes, the -gsp flag was successfully used instead.

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -87,8 +87,11 @@ module Fastlane
         UI.message("Uploading '#{path}'...")
         command = []
         command << File.expand_path(params[:binary_path]).shellescape
-        command << "-gsp #{params[:gsp_path].shellescape}" if params[:gsp_path]
-        command << "-a #{params[:api_token]}" if params[:api_token] && !params[:gsp_path]
+        if params[:gsp_path]
+          command << "-gsp #{params[:gsp_path].shellescape}"
+        elsif params[:api_token]
+          command << "-a #{params[:api_token]}"
+        end
         command << "-p #{params[:platform] == 'appletvos' ? 'tvos' : params[:platform]}"
         command << File.expand_path(path).shellescape
         begin
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -9,7 +9,7 @@ module Fastlane
         find_binary_path(params)
         find_gsp_path(params)
         find_api_token(params)
-        
+
         if !params[:api_token] && !params[:gsp_path]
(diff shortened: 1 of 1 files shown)


mmdock:
I signed it!

vinhlk:
hi @mmdock does your PR been accept? what version can we use your solution?

mmdock:
@vijaysharm this PR is still open.  It has not been merged as of yet.

However, if you are interested in giving this a test, please try setting up your Gemfile to use: 
`gem 'fastlane', git: 'https://github.com/mmdock/fastlane.git', branch: 'patch-1'
`
instead of just `gem 'fastlane'`

fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
```diff
@@ -87,8 +87,8 @@ def self.upload_dsym(params, path)
         UI.message("Uploading '#{path}'...")
         command = []
         command << File.expand_path(params[:binary_path]).shellescape
-        command << "-a #{params[:api_token]}" if params[:api_token]
         command << "-gsp #{params[:gsp_path].shellescape}" if params[:gsp_path]
+        command << "-a #{params[:api_token]}" if params[:api_token] && !params[:gsp_path]
```
joshdholtz: Can we actually make this an `if/else`? I think that would make this logic a bit clearer 🙃 Something like...
```rb
if params[:gsp_path]
  command << "-gsp #{params[:gsp_path].shellescape}"
elsif params[:api_token]
  command << "-a #{params[:api_token]}"
end
```

joshdholtz (changes requested):
One small change request!

mmdock:
@joshdholtz just saw this comment.  Will make change later today.

fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
mmdock: Made the change.  Sorry i didn't check this PR for so long.

mmdock:
@joshdholtz made the recommended change

joshdholtz (approved):
This is good! Thanks for making that change ❤️

Issue #13577, opened 2018-10-22
Deliver States Waiting App Store Connect Processing Even though Build has been Processed Already.


pparmaei:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When using deliver to upload an app to App Store Connect with version 2.02.30(2.2.30) with the following parameters:
![image](https://user-images.githubusercontent.com/23205547/47324082-49761c00-d613-11e8-9698-fd60497e808a.png)

After the upload, I see that the console:

```
INFO [2018-10-23 00:48:30.07]: You can tweet: "App Store Connect #iosprocessingtime 51 minutes"
INFO [2018-10-23 00:49:07.19]: Waiting App Store Connect processing for build 2.2.30 (2.2.30)... this might take a while...
```

However, the build has already been processed. On the Activity section on App Store Connect, I see the version as **2.02.30** and the build as **2.2.30**. So it looks like Deliver command is truncating the leading zeros from the **CFBundleShortVersionString**. However, according to the Apple documentation the leading zeros are only truncated for [CFBundleVersion](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364) and not [CFBundleShortVersionString](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-111349).

Nov1kov:
what fastlane build do you use?
probably the [same issue](https://github.com/fastlane/fastlane/issues/13372)

janpio:
@pparmaei can you share the fastlane version you are using? Did similar version numbers cause problems in the past already?
The issues @Nov1kov linked to and created is very similar, and points to this being a regression of a change made in fastlane.

pparmaei:
Hello @janpio 
Using version 2.107.0

I have come across the same issue with multiple different apps that have leading zeros in their release version.

Issue #13372, opened 2018-09-17
[Regression] waiting app store connect processing build infinity time


Nov1kov:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
after update from 2.103.1 to 2.104
when I upload build via _deliver_ with --submit_for_review, I receive an infinity "waiting processing build".

Probably, it is because of https://github.com/fastlane/fastlane/pull/13315
Same problem is when I use _pilot_ https://github.com/fastlane/fastlane/issues/11999

I guess it is because of my build number and version name ending: "00" (110.30.00)
Spaceship or fastlane are parsing it like 110.30.0

##### Complete output when running fastlane, including the stack trace and command used
I use python for execute fastlane
```
fastlane deliver -f --overwrite_screenshots --submit_for_review
```
running it from directory with DeliverFile

<details>
  <pre>
09:58:11 - fastlane deliver                    : INFO   
09:58:11 - fastlane deliver                    : INFO   [09:58:11]: Making sure the latest version on App Store Connect matches '110.30.00' from the ipa file...
09:58:12 - fastlane deliver                    : INFO   [09:58:12]: '110.30.00' is the latest version on App Store Connect
09:58:13 - fastlane deliver                    : INFO   [09:58:13]: Loading './metadata/ru/description.txt'...
09:58:15 - fastlane deliver                    : INFO   [09:58:15]: Activating language ru...
09:58:17 - fastlane deliver                    : INFO   [09:58:17]: Setting the app's age rating...
09:58:17 - fastlane deliver                    : INFO   [09:58:17]: Uploading metadata to App Store Connect
09:58:19 - fastlane deliver                    : INFO   [09:58:19]: Successfully uploaded set of metadata to App Store Connect
09:58:20 - fastlane deliver                    : INFO   [09:58:20]: Starting with the upload of screenshots...
09:58:20 - fastlane deliver                    : INFO   [09:58:20]: Removing all previously uploaded screenshots...
09:58:20 - fastlane deliver                    : INFO   [09:58:20]: Activating language ru...
09:58:22 - fastlane deliver                    : INFO   [09:58:22]: Uploading 5 screenshots for language ru
[...]

PR #14513, opened 2019-04-02
[deliver] Fix builds whose versions have components with leading zeros infinitely wait for submission

justindhill:
### Checklist

### Motivation and Context
This PR fixes an issue in Deliver's `SubmitForReview` which causes builds containing zero-padding in their version components to fail to submit. It is caused by an Apple issue which causes the App Store Connect API to temporarily return a `train_version` whose components are stripped of their zero padding. For example, "2019.04.01" would be returned as "2019.4.1". During the processing, eventually the API starts returning the correct form of the version again, i.e. "2019.04.01". 

The above causes Deliver to infinitely wait for the build to be processed because it notes the most recently uploaded build as soon as Transporter finishes and uses it to compare the version of builds in future calls to `candidate_builds` to prevent accidentally selecting the wrong build for submission.

Below is some trimmed json demonstrating the relevant changes to the data throughout processing:

Immediately after upload:
```json
{
	"buildVersion": "82",
	"trainVersion": "2019.4.1",
	"uploadDate": 1554160879000,
	"processing": true,
	"processingState": "processing",
	"processingError": null,
}
```

After the first phase of processing; note the trainVersion format has changed:
```json
{
	"buildVersion": "82",
	"trainVersion": "2019.04.01",
	"uploadDate": 1554161105000,
	"processing": true,
	"processingState": "processing",
	"processingError": null,
}
```

Processing complete:
```json
{
	"buildVersion": "82",
	"trainVersion": "2019.04.01",
	"uploadDate": 1554161105000,
	"processing": false,
	"processingState": null,
	"processingError": null,
}
```

Should solve fastlane/fastlane#13372 and fastlane/fastlane#13577

### Solution

The solution is to use the `app_version` attached to the options hash for comparing to the `candidate_builds` instead of the `train_version` from the previously found build. This ensures that a correctly formatted version is always used for comparison. In cases where an `app_version` is not supplied in the options hash, such as when calling `SubmitForReview` without an ipa and also not providing an `app_version` explicitly, we fall back to the behavior of using the one on the first build we find and hoping for the best.

I added a test that confirms that `wait_for_build` properly returns a build with the supplied app version even when `candidate_builds` returns a build with the zeros stripped. I also added a test that confirms a build is returned when nil is passed as the app version to `wait_for_build`.

### DIFF
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -74,18 +74,16 @@ module Deliver
         end
 
         latest_build = find_build(candidate_builds)
-        latest_build_matches = (latest_build&.train_version == app_version && latest_build&.build_version == build&.build_version)
+
+        # if the app version isn't present in the hash (could happen if we are waiting for submission, but didn't provide
+        # it explicitly and no ipa was passed to grab it from), then fall back to the best guess, which is the train_version
+        # of the most recently uploaded build
+        app_version = app_version || latest_build.train_version
+
         # Sometimes latest build will disappear and a different build would get selected
         # Only set build if no latest build found or if same build versions as previously fetched build
         # Issue: https://github.com/fastlane/fastlane/issues/10945
-        if build.nil? || latest_build_matches
-          # if the app version isn't present in the hash (could happen if we are waiting for submission, but didn't provide
-          # it explicitly and no ipa was passed to grab it from), then fall back to the best guess, which is the train_version
-          # of the most recently uploaded build
-          if app_version.nil?
-            app_version = latest_build.train_version
-          end
-
+        if build.nil? || (latest_build && latest_build.build_version == build.build_version && latest_build.train_version == app_version)
           build = latest_build
         end
 
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -74,10 +74,11 @@ module Deliver
         end
 
         latest_build = find_build(candidate_builds)
+        latest_build_matches = (latest_build && latest_build.train_version == app_version && latest_build.build_version == build.build_version)
         # Sometimes latest build will disappear and a different build would get selected
         # Only set build if no latest build found or if same build versions as previously fetched build
         # Issue: https://github.com/fastlane/fastlane/issues/10945
-        if build.nil? || (latest_build && latest_build.train_version == app_version && latest_build.build_version == build.build_version)
+        if build.nil? || latest_build_matches
           # if the app version isn't present in the hash (could happen if we are waiting for submission, but didn't provide
           # it explicitly and no ipa was passed to grab it from, then fall back to the best guess, which is the train_version
           # of the most recently uploaded build)
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -80,7 +80,7 @@ module Deliver
         # Issue: https://github.com/fastlane/fastlane/issues/10945
         if build.nil? || latest_build_matches
           # if the app version isn't present in the hash (could happen if we are waiting for submission, but didn't provide
-          # it explicitly and no ipa was passed to grab it from, then fall back to the best guess, which is the train_version
+          # it explicitly and no ipa was passed to grab it from), then fall back to the best guess, which is the train_version
           # of the most recently uploaded build)
           if app_version.nil?
             app_version = latest_build.train_version
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -29,7 +29,7 @@ module Deliver
       UI.success("Successfully submitted the app for review!")
     end
 
-    def select_build(options)
+    private def select_build(options)
       app = options[:app]
       app_version = options[:app_version]
       v = app.edit_version(platform: options[:platform])
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -74,7 +74,7 @@ module Deliver
         end
 
         latest_build = find_build(candidate_builds)
-        latest_build_matches = (latest_build && latest_build.train_version == app_version && latest_build.build_version == build.build_version)
+        latest_build_matches = (latest_build&.train_version == app_version && latest_build&.build_version == build&.build_version)
         # Sometimes latest build will disappear and a different build would get selected
         # Only set build if no latest build found or if same build versions as previously fetched build
         # Issue: https://github.com/fastlane/fastlane/issues/10945
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -31,6 +31,7 @@ module Deliver
 
     def select_build(options)
       app = options[:app]
+      app_version = options[:app_version]
       v = app.edit_version(platform: options[:platform])
 
       if options[:build_number] && options[:build_number] != "latest"
@@ -41,9 +42,9 @@ module Deliver
         end
       else
         UI.message("Selecting the latest build...")
-        build = wait_for_build(app)
+        build = wait_for_build(app, app_version)
       end
-      UI.message("Selecting build #{build.train_version} (#{build.build_version})...")
+      UI.message("Selecting build #{app_version} (#{build.build_version})...")
 
       v.select_build(build)
       v.save!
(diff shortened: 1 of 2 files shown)


justindhill:
I signed it.

justindhill:
Ok, signed with corporate email as well.

justindhill:
Added my corporate email to GitHub.

deliver/spec/submit_for_review_spec.rb
```diff
@@ -49,28 +49,98 @@ def make_fake_version
     describe :wait_for_build do
       context 'no candidates' do
         let(:fake_app) { make_fake_app }
-        let(:fake_version) { make_fake_version }
         let(:time_now) { Time.now }
+        let(:fake_version) do
+          fake_version = make_fake_version
+          fake_version[:app_version] = "1.2.3"
```
janpio: Should this maybe moved to `make_fake_version`?

janpio:
Oh wow, bad Apple - and all developers can pretty much guess what is going on there 🙈 

Fix looks good on a superficial review ✔️

deliver/spec/submit_for_review_spec.rb
justindhill: Sure, that makes sense. I'll change it right now.

justindhill: Done

deliver/spec/submit_for_review_spec.rb
```diff
@@ -17,7 +17,11 @@ def make_fake_app
   end
 
   def make_fake_version
-    return OpenStruct.new({})
+    fake_version = OpenStruct.new({
```
giginet: You can use rspec's `double` instead.

deliver/spec/submit_for_review_spec.rb
```diff
@@ -66,11 +71,74 @@ def make_fake_version
...
-          expect(review_submitter.wait_for_build(fake_app)).to eq(only_build)
+          expect(review_submitter.wait_for_build(fake_app, "1.2.3")).to eq(only_build)
+        end
+      end
+
+      context 'has candidates and one build, but no app version' do
+        let(:fake_app) { make_fake_app }
+        let(:fake_builds) do
+          builds = make_fake_builds(1)
+          builds[0][:train_version] = "1.2.3"
```
giginet: You should use `double` and `allow`.

```ruby
build = double('fake_build')
allow(build).to receive(:[]).with(:train_version).and_return("1.2.3")
allow(build).to receive(:[]).with(:build_version).and_return("1")
allow(build).to receive(:[]).with(:processing).and_return(false)
[build]
```

deliver/lib/deliver/submit_for_review.rb
```diff
@@ -73,17 +74,25 @@ def wait_for_build(app)
         end
 
         latest_build = find_build(candidate_builds)
+        latest_build_matches = (latest_build && latest_build.train_version == app_version && latest_build.build_version == build.build_version)
```
giginet: You can use `&.`.

```suggestion
        latest_build_matches = (latest_build&.train_version == app_version && latest_build&.build_version == build&.build_version)
```

deliver/spec/submit_for_review_spec.rb
justindhill: I don't really know Ruby, so I'm totally willing to defer to your expertise, but would you mind explaining the advantage of stubbing with rspec as opposed to just setting the value? To my untrained eye, it just looks more verbose while accomplishing the same thing.

deliver/lib/deliver/submit_for_review.rb
justindhill: This line just extracts the expression from the line below, but what you proposed is a great syntax I didn't know about!

deliver/spec/submit_for_review_spec.rb
giginet: According to RSpec documentation.
https://relishapp.com/rspec/rspec-mocks/docs/basics/test-doubles

`any message you have not allowed or expected will trigger an error`

I think that is the main benefit using Test Doubles.
Using test doubles, we can notice unexpected method callings.

deliver/lib/deliver/submit_for_review.rb
giginet: I'm sorry, this syntax is not allowed on fastlane project 😢 
`&.` was introduced since Ruby 2.3, but unfortunately, TargetRubyVersion is still 2.0.
https://github.com/fastlane/fastlane/blob/master/.rubocop.yml#L244

This syntax is correct, but we can't use this.
https://circleci.com/gh/fastlane/fastlane/30932?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

giginet: `build` seems to be optional, so we have to do nil-cheking for `build`.

`latest_build_matches = (latest_build && build && latest_build.train_version == app_version && latest_build.build_version == build.build_version)`

justindhill:
@giginet @janpio I've made some changes based on your feedback and they're ready for review.

1. I made wait_for_build read a bit more nicely in addition to addressing the nil check @giginet pointed out.
2. I reworked the tests to use doubles instead of OpenStructs.

deliver/lib/deliver/submit_for_review.rb
```diff
@@ -41,17 +42,17 @@ def select_build(options)
...
+        build = wait_for_build(app, app_version)
       end
-      UI.message("Selecting build #{build.train_version} (#{build.build_version})...")
+      UI.message("Selecting build #{app_version} (#{build.build_version})...")
 
       v.select_build(build)
       v.save!
 
       UI.success("Successfully selected build")
     end
```
giginet: you can add `private`

deliver/spec/submit_for_review_spec.rb
```diff
@@ -17,7 +17,9 @@ def make_fake_app
   end
 
   def make_fake_version
-    return OpenStruct.new({})
+    fake_version = double('fake_version')
+    allow(fake_version).to receive(:[]).with(:app_version).and_return("1.2.3")
+    return fake_version
```
giginet: You can drop `return`.

deliver/spec/submit_for_review_spec.rb
```diff
@@ -66,11 +69,77 @@ def make_fake_version
...
+
+      context 'has candidates and one build, but no app version' do
+        let(:fake_app) { make_fake_app }
+        let(:fake_builds) do
+          build = double('fake_build')
+          allow(build).to receive(:train_version).and_return("1.2.3")
+          allow(build).to receive(:build_version).and_return(1)
+          allow(build).to receive(:upload_date).and_return(1_554_754_590)
+          expect(build).to receive(:processing).and_return(true)
+          return [build]
```
giginet: You can drop `return` in ruby way.

giginet (approved):
Great job 🌸

deliver/lib/deliver/submit_for_review.rb
justindhill: Sure, done.

deliver/spec/submit_for_review_spec.rb
justindhill: Done.

deliver/spec/submit_for_review_spec.rb
justindhill: Done.

giginet:
@justindhill Please agree with CLA before merging

https://github.com/fastlane/fastlane/pull/14513#issuecomment-478844478

justindhill:
> @justindhill Please agree with CLA before merging
> 
> [#14513 (comment)](https://github.com/fastlane/fastlane/pull/14513#issuecomment-478844478)

@giginet I actually did agree to the CLA. The PR was labeled with cla:yes before I accepted your suggestion a few days ago, at which point @googlebot removed the label. According to the comment it left, you have to manually re-add the tag after confirming consent from all developers. Since you are the other developer, I think it's ok. :)

janpio:
Yep, @googlebot is not smart enough for Github suggestions that are accepted, but the `cla:no` will be no problem here when someone decides this is good to merge - so nothing to do in that regard.

deliver/lib/deliver/submit_for_review.rb
```diff
@@ -41,17 +42,17 @@ def select_build(options)
         end
       else
         UI.message("Selecting the latest build...")
-        build = wait_for_build(app)
+        build = wait_for_build(app, app_version)
       end
-      UI.message("Selecting build #{build.train_version} (#{build.build_version})...")
+      UI.message("Selecting build #{app_version} (#{build.build_version})...")
```
vincentisambart: That is incorrect, the `app_version` passed to `deliver` is `build_version` not `train_version`.

Issue #11788, opened 2018-02-06
Importing provisioning profiles from file


akashivskyy:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Hi! I have a need for an action that would import a `.mobileprovision` file from a folder in my repository to `~/Library/MobileDevice/Provisioning Profiles`.

Is there a fastlane action that does that (I cound't find such options in any of the code signing actions) or should I do this manually using `cp`?

akashivskyy:
@fastlane-bot Nope and nope! 😅

taquitos:
I think the `copy_artifacts` action will work for you
Check out the example code:
```
'copy_artifacts(
            target_path: "artifacts",
            artifacts: ["*.cer", "*.mobileprovision", "*.ipa", "*.dSYM.zip", "path/to/file.txt", "another/path/*.extension"]
          )
```

You can provide paths to the files, and a target directory.

akashivskyy:
@taquitos That did the trick, but it's semantically incorrect (these are not artifats). I believe there should be an action for importing `.mobileprovision` files to `~/Library/MobileDevice/Provisioning Profiles` just like `import_certificate` imports `.p12` to Keychain.

taquitos:
Yeah, I agree, while it works, it could lead to confusion.

akashivskyy:
I used Joshua's solution for now which is satisfactory enough. Maybe I'll take advantage of Fastlane's amazing contribution culture and open a PR with an action that meets the need described in this issue. I'm closing the issue for now.

janpio:
Hey @akashivskyy, I just searched for a similar action and found this "old" issue. I would love if such an action would exist, even if it would "only" be a pretty shallow one.

For documentation, this is what I ended up with:
```
  require 'fileutils'
  path = File.expand_path("~") + "/Library/MobileDevice/Provisioning Profiles/"
  unless File.directory?(path)
    FileUtils.mkdir_p(path)
  end
  FileUtils.cp('../foo/367f12b1-4765-4d71-86be-e69ac68885b1.mobileprovision', path)
```
[...]

Issue #14255, opened 2019-02-16
ENV vs. lane_context vs. SharedValues?


mollyIV:
Few days ago the issue related to `lane_context` and ENV has been created https://github.com/fastlane/fastlane/issues/14182. The investigation opened up a broader discussion about the usage of `SharedValues` in general. I talked to @janpio and we came up with a few questions / ideas and decided to open a new issue for the further discussion. 

**When should a user use the values from `lane_context[SharedValues::XYZ]` and when should a user use the values from ENV?** 

`SharedValues` and ENV seem to be very close with each other. Could you please provide a bit more context what was the reason behind introducing `SharedValues` and not just using ENV variables in order to pass the values between lanes? 

I couldn't find the docs saying about the relationship between the values available as `SharedValues` and the values available as environment variables (ENV). Currently there's no much documentation about [lane_context](https://docs.fastlane.tools/actions/lane_context/):

* https://docs.fastlane.tools/advanced/lanes/#lane-context

> The different actions can communicate with each other using a shared hash.

The documentation basically says you can use the shared hash between the lanes but there's missing info which lanes expose which values which points us to the question number two 👇 

**It is not possible to print out the ENV and `SharedValues` values which are set by a particular lane.** 

Couldn't find any info how to print out the values set by a lane. What we can do though is to print out  ENV variables (`fastlane action [action_name]`) that we can set as input parameters instead of passing input parameters directly to the lane. We could add a new functionality of printing out the summary table and it would be very helpful to see what `SharedValues` / ENV action/lane sets. In that way a user could easily spot what `SharedValues`/ENV will be set by a particular lane/action. 

**How to decide whether a user should set `SharedValues` or ENV in his plugin or new lane?**

Since there's no clear rules to follow whether to use `SharedValues` or ENV variables it's hard to decide what to support in new lanes / custom plugins. 

We'd love to hear your thoughts on this subject and come up with the ideas how we could clean up the usage of `SharedValue`, ENV and `lane_context` and make them easier to understand 🚀  

Thank you! 🙇 🙇

janpio:
I think the question is a bit broader even than what you suggested:

> How should fastlane actions, plugins and user lanes work with and manipulate `ENV` variables, `lane_context` and `SharedValues`?

Different problem areas are:

- Usage Documentation
- Development Documentation
- Development Practices

---

My current knowledge:

### ENV variables 

[...]

PR #14516, opened 2019-04-02
[action] New action to install provisioning profile from file

SofteqDG:
### Checklist

### Motivation and Context
This action can be very useful if user doesn't use `match` action for automatic provisioning profiles management. It can be used inside of Fastfile and also can be run from shell.
This PR also fixes open issue #11788 

### Description
This action allow user to install provisioning profile from local file. This is just a wrapper around the `FastlaneCore::ProvisioningProfile` class and doesn't provide any additional functionality.

### DIFF
--- a/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
@@ -4,16 +4,16 @@ module Fastlane
   module Actions
     class InstallProvisioningProfileAction < Action
       def self.run(params)
-        absolute_path = File.expand_path(params[:provisioning_profile_path])
+        absolute_path = File.expand_path(params[:path])
         FastlaneCore::ProvisioningProfile.install(absolute_path)
       end
 
       def self.description
-        "Install provisioning profile from inputfile"
+        "Install provisioning profile from path"
       end
 
       def self.details
-        "Install provisioning profile from inputfile for current user"
+        "Install provisioning profile from path for current user"
       end
 
       def self.authors
@@ -30,7 +30,7 @@ module Fastlane
 
       def self.available_options
         [
-          FastlaneCore::ConfigItem.new(key: :provisioning_profile_path,
+          FastlaneCore::ConfigItem.new(key: :path,
                                   env_name: "FL_INSTALL_PROVISIONING_PROFILE_PATH",
                                description: "Path to provisioning profile",
                                   optional: false,
@@ -50,7 +50,7 @@ module Fastlane
 
       def self.example_code
         [
-          'install_provisioning_profile(provisioning_profile_path: "profiles/profile.mobileprovision")'
+          'install_provisioning_profile(path: "profiles/profile.mobileprovision")'
         ]
       end
     end
--- a/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
@@ -1,0 +1,52 @@
+require 'shellwords'
+
+module Fastlane
+  module Actions
+    class InstallProvisioningProfileAction < Action
+      def self.run(params)
+        provisioning_profile_path = params[:provisioning_profile_path]
+        FastlaneCore::ProvisioningProfile.install(provisioning_profile_path)
+      end
+
+      def self.description
+        "Install provisioning profile from inputfile"
+      end
+
+      def self.details
+        "Install provisioning profile from inputfile for current user"
+      end
+
+      def self.authors
+        ["SofteqDG"]
+      end
+
+      def self.category
+        :code_signing
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :provisioning_profile_path,
+                                  env_name: "FL_INSTALL_PROVISIONING_PROFILE_PATH",
+                               description: "Path to provisioning profile",
+                                  optional: false,
+                                      type: String)
+        ]
+      end
+
+      def self.return_value
+        "The absolute path to the installed provisioning profile"
+      end
+
+      def self.example_code
+        [
+          'install_provisioning_profile(provisioning_profile_path: "profiles/profile.mobileprovision")'
+        ]
+      end
+    end
+  end
+end
--- a/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
@@ -4,8 +4,13 @@ module Fastlane
   module Actions
     class InstallProvisioningProfileAction < Action
       def self.run(params)
-        provisioning_profile_path = params[:provisioning_profile_path]
-        FastlaneCore::ProvisioningProfile.install(provisioning_profile_path)
+        profile_path = params[:provisioning_profile_path]
+        absolute_profile_path = File.expand_path(profile_path)
+        if File.exist?(absolute_profile_path)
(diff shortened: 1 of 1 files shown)


SofteqDG:
I signed it!

SofteqDG:
I signed it!

fastlane/lib/fastlane/actions/install_provisioning_profile.rb
```diff
@@ -0,0 +1,52 @@
...
+        :code_signing
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :provisioning_profile_path,
```
janpio: Should we maybe check if the path exists here? There is a [check if the copying actually worked](https://github.com/fastlane/fastlane/blob/2545fd0225b85c1204b30061f01a30ebd21d45d3/fastlane_core/lib/fastlane_core/provisioning_profile.rb#L78-L80), but might be nice to catch this even earlier. Some other actions might give ideas how to implement that I guess.

zeDDer: Nice catch! I'll add a check for path

zeDDer: @janpio 
I've added a check inside of `run` method with additional logic to get absolute path cos i don't know if `FastlaneCore::ProvisioningProfile.install` method can work with relative paths.

janpio: That additional logic is interesting.

It might be possible to check the path in the options definition already - but I am not 100% sure.

zeDDer: Yep, it is possible. So what should i do? I can check the path inside of the options definition and only expand path to absolute inside of `run` method without additional checks.

janpio: I guess that would be the optimal way to do this for an action.

zeDDer: Done

fastlane/lib/fastlane/actions/install_provisioning_profile.rb
```diff
@@ -0,0 +1,58 @@
...
+                                  env_name: "FL_INSTALL_PROVISIONING_PROFILE_PATH",
+                               description: "Path to provisioning profile",
+                                  optional: false,
+                                      type: String,
+                              verify_block: proc do |value|
+                                absolute_path = File.expand_path(value)
+                                unless File.exist?(absolute_path)
+                                  UI.user_error!("Failed installation of provisioning profile from file at path: '#{value}'")
```
janpio: This should probably just say Could not find provisioning profile at path: ..., as it doesn't try to install this yet.

SofteqDG: Message changed

SofteqDG:
@janpio 
Yep, i can try to add tests for this action. Could you please tell me where i can see an example of how to add tests?

There is also one improvement in my head. We can expose some parameters of the provisioning profile through the lane context. I think that the most useful parameters are profile name, profile uuid, team name and team identifier. They can be used for other actions. What do you think about it?

janpio:
I am not sure how one would exactly test this, but you can look at all the other files with `spec` in it to get an idea how testing works in general.

> There is also one improvement in my head. We can expose some parameters of the provisioning profile through the lane context. I think that the most useful parameters are profile name, profile uuid, team name and team identifier. They can be used for other actions. What do you think about it?

I don't know the usecase, but could be useful. But best create a new PR for this that is separate from this (You can create a new branch in your current commit, then create your further commits there - and create a PR from that. When this is merged, the first few commits will automatically be left out of your second PR or your can rebase on master)

SofteqDG:
> I am not sure how one would exactly test this, but you can look at all the other files with `spec` in it to get an idea how testing works in general.

Thanks. Will take a look on how testing works.

> I don't know the usecase, but could be useful.

It can be used in conjunction with manual code signing. Here is an example of Fastfile:
```ruby
  lane :build_production do

    # Install provisioning profile
    install_provisioning_profile(
      provisioning_profile_path: "profiles/profile.mobileprovision"
    )

    # Read provisioning profile parameters exposed by 'install_provisioning_profile' action
    profile_name = Actions.lane_context[SharedValues::PROVISIONING_PROFILE_NAME]
    prodile_uuid = Actions.lane_context[SharedValues::PROVISIONING_PROFILE_UUID]
    profile_team_id = Actions.lane_context[SharedValues::PROVISIONING_PROFILE_TEAM_ID]

    # Use manual code signing
    automatic_code_signing(
      use_automatic_signing: false,
      path: "MyApp.xcodeproj",
      team_id: profile_team_id,
      profile_name: profile_name,
      profile_uuid: profile_name
    )

    # Build ipa
    ...

  end
```



> But best create a new PR for this that is separate from this (You can create a new branch in your current commit, then create your further commits there - and create a PR from that. When this is merged, the first few commits will automatically be left out of your second PR or your can rebase on master)

Ok

janpio:
@mollyIV That looks like what we talked about for `SharedValues`, correct?

mollyIV:
That's correct @janpio 👍 

According to our discussion [here](https://github.com/fastlane/fastlane/issues/14255#issuecomment-471986433), `SharedValues` / `lane_context` is a preferred way of passing a value generated by one action to another (as the action's input). 

Taking into account @SofteqDG [comment](https://github.com/fastlane/fastlane/pull/14516#issuecomment-483391739), in this PR we would have:

```ruby
module Fastlane
  module Actions
    module SharedValues
      PROVISIONING_PROFILE_NAME = : PROVISIONING_PROFILE_NAME
      PROVISIONING_PROFILE_UUID =: PROVISIONING_PROFILE_UUID
      PROVISIONING_PROFILE_TEAM_ID =: PROVISIONING_PROFILE_TEAM_ID
    end

    class InstallProvisioningProfileAction < Action
      ...
      
      def self.output
        [
          ['PROVISIONING_PROFILE_NAME', 'Description to be delivered'],
          ['PROVISIONING_PROFILE_UUID', 'Description to be delivered'],
          ['PROVISIONING_PROFILE_TEAM_ID', 'Description to be delivered']
        ]
      end
      
      ...
```

To get a bit more details about `SharedValues` and `output` method, please have a look at [this thread](https://github.com/fastlane/fastlane/issues/14517) 🙇 

Cheers!

fastlane/lib/fastlane/actions/install_provisioning_profile.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :provisioning_profile_path,
+                                  env_name: "FL_INSTALL_PROVISIONING_PROFILE_PATH",
+                               description: "Path to provisioning profile",
+                                  optional: false,
+                                      type: String,
+                              verify_block: proc do |value|
```
mollyIV: What if the file path contains the spaces? Shouldn't we `shellescape` it? 🤔 

Adding a unit test for it would be great 👍

joshdholtz: @mollyIV No need for a `shellescape` here! This path is only used in Ruby stuff so it doesn't actually get pushed down to a shell command anywhere on our end 🙃

joshdholtz (approved):
This looks good! I renamed the option to `:path` and updated the description but this was 💯 THanks for adding this 😊

SofteqDG:
Hi, @joshdholtz 
Thanks for merging this and for your improvements!

Issue #14116, opened 2019-01-22
[deliver] Cannot upload iPad 3rd generation screenshots


dayjer:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Despite using snapshot to generate screenshots "iPad Pro (12.9-inch) (3rd generation)-01_screenshot_X.png", and after battling with the ongoing sizing issues, such as https://github.com/fastlane/fastlane/issues/13698 and https://github.com/fastlane/fastlane/issues/13610 
It still uploads the images for "iPad Pro (2nd Generation) 12.9-inch Display" -- the problem with this is that all my localisations then go and use English (e.g. https://cl.ly/db4d1a265bbb) 

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>[08:19:18]: Uploading 30 screenshots for language zh-Hant
[08:19:18]: Uploading './screenshots/zh-Hant/iPad Pro (12.9-inch) (3rd generation)-01_screenshot_1.png'...
[08:19:43]: Uploading './screenshots/zh-Hant/iPad Pro (12.9-inch) (3rd generation)-02_screenshot_2.png'...
[08:20:08]: Uploading './screenshots/zh-Hant/iPad Pro (12.9-inch) (3rd generation)-03_screenshot_3.png'...
[08:20:35]: Uploading './screenshots/zh-Hant/iPad Pro (12.9-inch) (3rd generation)-04_screenshot_4.png'...
[08:20:52]: Uploading './screenshots/zh-Hant/iPad Pro (12.9-inch) (3rd generation)-05_screenshot_5.png'...
...
[✔] Saving changes 
[08:23:45]: Successfully uploaded screenshots to App Store Connect</pre>
</details>

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.14.2                                                        |
| Ruby                        | 2.3.7                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.17.2 (Apple Git-113)                             |
[...]

PR #14533, opened 2019-04-07
[deliver] upload_screenshots_spec to test screenshot collection

liamnichols:
### Checklist

### Motivation and Context

The iPad Pro 12.9" (3rd Generation) has the same resolution as the 2nd and 1st Generation versions however in App Store Connect, they have different slots (and `device_type` values) for screenshots and this breaks a fundamental assumption that `deliver` has made when uploading screenshots.  

#14116 is tracking the actual issue itself and its pretty certain that `app_screenshot.rb` and `upload_screenshots.rb` will need to be updated to account for this however while I was investigating, I struggled undemanding all of the complicated things that these modules are handling so decided to try and write some tests for the existing behaviour.

### Description

These changes simply introduce an additional spec to test the collection of screenshots by `Deliver::UploadScreenshots`. The following cases are tested: 

- An empty directory (no screenshots)
- The most basic case of a screenshot
- Excluding device_type's unsupported by App Store Connect (the iPhone XR)
- Localisation
- Excluding original screenshots if _framed variants exist
  - But preserving Apple Watch screenshots without frames
- The magic appleTV directory
- iMessage screenshots

Since this module relies heavily on the file system, I copied the approach used in loader_spec.rb and used FakeFS to simulate things. Not sure if it's the best approach but it seems to have worked!

### DIFF
--- a/deliver/lib/deliver/upload_screenshots.rb
+++ b/deliver/lib/deliver/upload_screenshots.rb
@@ -88,7 +88,7 @@ module Deliver
       screenshots = []
       extensions = '{png,jpg,jpeg}'
 
-      available_languages = Spaceship::Tunes.client.available_languages.each_with_object({}) do |lang, lang_hash|
+      available_languages = UploadScreenshots.available_languages.each_with_object({}) do |lang, lang_hash|
         lang_hash[lang.downcase] = lang
       end
 
@@ -145,6 +145,15 @@ module Deliver
       end
 
       return screenshots
+    end
+
+    # helper method so Spaceship::Tunes.client.available_languages is easier to test
+    def self.available_languages
+      if Helper.test?
+        FastlaneCore::Languages::ALL_LANGUAGES
+      else
+        Spaceship::Tunes.client.available_languages
+      end
     end
   end
 end
--- a/deliver/spec/upload_screenshots_spec.rb
+++ b/deliver/spec/upload_screenshots_spec.rb
@@ -1,0 +1,93 @@
+require 'deliver/upload_screenshots'
+require 'fakefs/spec_helpers'
+
+describe Deliver::UploadScreenshots do
+  describe "#collect_screenshots_for_languages (screenshot collection)" do
+    include(FakeFS::SpecHelpers)
+
+    def add_screenshot(file)
+      FileUtils.mkdir_p(File.dirname(file))
+      File.open(file, 'w') { |f| f << 'touch' }
+    end
+
+    def collect_screenshots_from_dir(dir)
+      Deliver::UploadScreenshots.new.collect_screenshots_for_languages(dir, false)
+    end
+
+    before do
+      allow(FastImage).to receive(:size) do |path|
+        path.match(/{([0-9]+)x([0-9]+)}/).captures.map(&:to_i)
+      end
+    end
+
+    it "should not find any screenshots when the directory is empty" do
+      screenshots = collect_screenshots_from_dir("/")
+      expect(screenshots.count).to eq(0)
+    end
+
+    it "should find screenshot when present in the directory" do
+      add_screenshot("/en-GB/iPhone8-01First{750x1334}.jpg")
+      screenshots = collect_screenshots_from_dir("/")
+      expect(screenshots.count).to eq(1)
+      expect(screenshots.first.screen_size).to eq(Deliver::AppScreenshot::ScreenSize::IOS_47)
+    end
+
+    it "should not collect iPhone XR screenshots" do
+      add_screenshot("/en-GB/iPhoneXR-01First{828x1792}.jpg")
+      screenshots = collect_screenshots_from_dir("/")
+      expect(screenshots.count).to eq(0)
+    end
+
+    it "should find different languages" do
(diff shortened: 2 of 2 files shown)


deliver/lib/deliver/upload_screenshots.rb
```diff
@@ -146,5 +146,13 @@ def collect_screenshots_for_languages(path, ignore_validation)
 
       return screenshots
     end
+
+    def self.available_languages
```
janpio: Interesting. Why was this change necessary?

liamnichols: I copied the same approach from loader.rb 

https://github.com/liamnichols/fastlane/blob/2fabb8ec9f84a19ebb00af47ac6875ca4fdb576f/deliver/lib/deliver/loader.rb#L27-L31

Since this code path will run through the tests, I didn’t want it to try and fetch this via the ITC api...

Although now you point it out, I guess I could just mock that spaceship method in the spec instead? Would it be better to do that? 🤔

deliver/spec/upload_screenshots_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'deliver/upload_screenshots'
+require 'fakefs/spec_helpers'
+
+describe Deliver::UploadScreenshots do
+  describe "screenshot collection" do
+    include(FakeFS::SpecHelpers)
+
+    def add_screenshot(file)
+      FileUtils.mkdir_p(File.dirname(file))
+      File.open(file, 'w') { |f| f << 'touch' }
```
giginet: Should this be binary?

deliver/lib/deliver/upload_screenshots.rb
giginet: It can be `private`

giginet (approved):
🚀

deliver/lib/deliver/upload_screenshots.rb
liamnichols: I think I prefer the idea of mocking it now that you mention it and since that is what I’m doing for the file size it probably makes even more sense... I’ll update the branch to do this before merging 👍🏼

deliver/spec/upload_screenshots_spec.rb
liamnichols: I'm not sure.. I copied this approach from `loader_spec.rb#L14` so I hope its fine

liamnichols:
Whoops ☝️ I tried making the changes on an iPad without actually testing the code 🙈 ... will fix it tomorrow

deliver/lib/deliver/upload_screenshots.rb
liamnichols: So for some reason `Spaceship::Tunes.client` is `nil` during the tests so it involves more than just stubbing the `available_languages` property on the class as I have to also initialise it in the tests and I'm not really sure how to do that 😅 

I've just reverted it back to how it was and used the `Helper.test?` check again

janpio: That's fine, maybe create an issue for what you tried to achieve so we don't forget about it and can maybe take another shot later?

liamnichols:
@giginet .. I tried to apply the `private` attribute but got the following error:

> private (on line 150) does not make singleton methods private. Use private_class_method or private inside a class << self block instead.
Introduce upload_screenshots_spec to test screenshot collection

I used `private_class_method` instead and it caused deployment tests to fail 🙈 
I'm not sure if there was a better way to do this but my Ruby abilities are holding me back, do you still want me to make the changes or are you happy to merge? Thanks 🙇

deliver/lib/deliver/upload_screenshots.rb
janpio:
```suggestion
    # helper method so Spaceship::Tunes.client.available_languages is easier to test
    def self.available_languages
```
maybe?

deliver/spec/upload_screenshots_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'deliver/upload_screenshots'
+require 'fakefs/spec_helpers'
+
+describe Deliver::UploadScreenshots do
+  describe "screenshot collection" do
```
janpio:
```suggestion
  describe "#collect_screenshots_for_languages (screenshot collection)" do
```
Is this correct ruby convention?

janpio (approved):
Looks good to me. 

Not sure if this covers 100% of the method, but we can figure that out later - this is much better than no tests at all.

giginet (approved):
> I used private_class_method instead and it caused deployment tests to fail 🙈

I got.
You can also test for private methods using `Object#send`, but this approach is tricky and complex. you not need to make `private`.

Thank you 👍

janpio:
(No need to clean up your commits and force push @liamnichols btw - we always "Squash and merge" so that a PR always gets merged as one clean commit)

janpio:
❤️ 😍 💟 New tests!  😻 💓 🚀

liamnichols:
> (No need to clean up your commits and force push @liamnichols btw - we always "Squash and merge" so that a PR always gets merged as one clean commit)

This is good to know 😬 

Thanks for all the help with things!

PR #14538, opened 2019-04-09
[rspec] add an e2e test for Fastlane doc generator pipeline

dotdoom:
### Checklist

### Motivation and Context
There's no end-to-end test right now, and I suspect there might be some issues on Windows platform (depending on Git SCM settings).

### Description
Add an e2e test, with diff checking using git. While at it, touch a couple surrounding tests to use proper Tempfile.


dotdoom:
Ha! Looks like the test doesn't like the documentation autogenerated by fastlane.

dotdoom:
@janpio I pinged the tests, but there's a failure on macOS which I think is unrelated:

```
 ```

janpio:
Yes. What is confusing me is that I fixed this on master, so it should not really happen any more 🤷

PR #14566, opened 2019-04-11
[spaceship/pilot] New Spaceship::ConnectAPI module and removal of Spaceship::TestFlight from Pilot and BuildWatcher

joshdholtz:
### Motivation and Context
`Spaceship::TestFlight` can't retrofit to work with new App Store Connect API so building out `Spaceship::ConnectAPI`

### Description
There were three modules that were added/modified in the making of this PR:
- Spaceship::Connect::API
- FastlaneCore::BuildWatcher
- Pilot

Below is a summary 👇  

#### Spaceship::Connect::API
The first goal was to create a stable release of `Spaceship::ConnectAPI` that allowed for calling of the API, paging, and mapping the response to usable and extendable models with as little cookie cutter and copy/pasting as possible without it being too magical 🙃 **There will be a follow up PR that adds documentation and migration strategies from `Spaceship::TestFlight`**

- `Spaceship::ConnectAPI::Client`
  - The core of the App Store Connect API which makes all the network calls using `Faraday`
  - API methods return `Spaceship::ConnectAPI::Response` objects
- `Spaceship::ConnectAPI::Response`
  - Contains raw `body` and `status`
  - Automatically maps the `body` to an array of `Spaceship::ConnectAPI::Model`
  - Has logic to get next page and all pages
- `Spaceship::ConnectAPI::Model`
  - A module that gives the model classes all their power
  - Registers the models `type` when the module is included so that the API responses can be automatically mapped to model objects
  - Some models wrap `Spaceship::ConnectAPI::Client` to allow for easier use of calling API
  - Available models
    - `Spaceship::ConnectAPI::App`
    - `Spaceship::ConnectAPI::BetaAppLocalization`
    - `Spaceship::ConnectAPI::BetaAppReviewDetail`
    - `Spaceship::ConnectAPI::BeatAppReviewSubmission`
    - `Spaceship::ConnectAPI::BetaBuildLocalization`
    - `Spaceship::ConnectAPI::BetaBuildMetric`
    - `Spaceship::ConnectAPI::BetaGroup`
    - `Spaceship::ConnectAPI::BetaTester`
    - `Spaceship::ConnectAPI::BetaTesterMetric`
    - `Spaceship::ConnectAPI::Build`
    - `Spaceship::ConnectAPI::BuildBetaDetail`
    - `Spaceship::ConnectAPI::BuildDelivery`
    - `Spaceship::ConnectAPI::PreReleaseVersion`
    - `Spaceship::ConnectAPI::User`
- `Spaceship::ConnectAPI::Models`
  - Handles the JSON API spec parsing using the registered types from `Spaceship::ConnectAPI::Model`
  - Uses the `included` and `relationships` to map models to attributes in the model already being mapped

##### Example
```rb
Spaceship::Tunes.login('email@email.com')
Spaceship::Tunes.select_team(team_name: 'Team Name')

app = Spaceship::ConnectAPI::App.find("com.joshholtz.FastlaneTest") 
builds = app.builds
puts "Number of builds: #{builds.size}"
  
builds.each do |build|
  puts "\tVersion: #{build.version}, Uploaded: #{build.uploaded_date}"
end
```

#### FastlaneCore::BuildWatcher
The next step after creating `Spaceship::ConnectAPI` was to start removing references of `Spaceship::TestFlight` from _Pilot_. _Pilot_ makes use of `FastlaneCore::BuildWatcher` to wait for the builds to finish processing before all of the options can be applied.

`FastlaneCore::BuildWatcher.wait_for_build_processing_to_be_complete` received a temporary new parameter called `return_spaceship_testflight_build`. This was added because we needed to make `BuildWatcher` backwards compatible by optionally returning a `Spaceship::TestFlight::Build` object as there are other places (and users) that call this module directly. By passing `true` to `return_spaceship_testflight_build` (which is its default value), the method will still return a `Spaceship::TestFlight::Build` object. _Pilot_ will pass in a `false` to this command which will allow to return a new `Spaceship::ConnectAPI::Build` object.

⚠️ The `Spaceship::TestFlight::Build` object that gets returns is a generated object from the `Spaceship::ConnectAPI::Build` object. Its not a fully qualified `Spaceship::TestFlight::Build` but _should_ keep thing working for most things that will use it.

`BuildWatcher` calls two `Spaceship::ConnectAPI::Client` methods to determine when a build is done processing:
- `get_build_deliveries`
  - When a build is initially submitting in a the "processing" state, it will appear in this response
- `get_builds`
  - When a a build is done processing (either successfully or non successfully), it will then show up in this response

#### Pilot
Once `FastlaneCore::BuildWatcher` changes were complete in returning the new `Spaceship::ConnectAPI::Build` object, _Pilot_ could start passing this new build object around to perform all of the configurations that it needed to do. There weren't _too_ many big changes that had to occur for _Pilot_ to use these new models. _PIlot_ was already using most of the new App Store Connect APIs but was just doing so directly with raw JSON requests and responses so these changes were mainly to make the code cleaner and use the models.


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

spaceship/lib/spaceship/connect_api/models/model.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      def self.inflate_model_relationships(type_instance, model_data, included)
+        # Relationship attributes to set
+        attributes = {}
+
+        # 1. Iterate over relations
+        # 2. Find id and type
+        # 3. Find matching id and type in included
+        # 4. Inflate matching data and set in attributes
+        relationships = model_data["relationships"] || []
```
snatchev: it looks like `model_data` is only needed to get at the `relationships`. Lets change this method so that it accepts `relationships` instead of model data and promote that lookup outside of this method.

spaceship/lib/spaceship/connect_api/models/model.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+
+      def self.parse(json)
+        data = json["data"]
+        raise "No data" unless data
+
+        included = json["included"] || []
+
+        if data.is_a?(Hash)
+          inflate_model(model_data, included)
```
snatchev: typo? `model_data` -> `data`?

snit-tile:
@joshdholtz shouldn't this: https://github.com/fastlane/fastlane/pull/14566/files#diff-d4092cd80ac929af2fa18075a4704f21R18
be using `https://api.appstoreconnect.apple.com/v1`?

snit-tile:
`req.options.params_encoder = Faraday::NestedParamsEncoder` is giving me the following error: 
```
client.rb:323:in `block in get_beta_groups': undefined method `options' for #<Object:0x00007f9c3e9ead18 @headers={}> (NoMethodError)
```

joshdholtz:
@snit-tile `https://api.appstoreconnect.apple.com/v1` only works if using the JWT authentication method - it doesn't work if using the web session like _fastlane_ currently uses so we cannot make a full switch over to that. We will need to chose dynamically which API endpoint to use depending on the authentication method being used 😬

snatchev (approved):
@joshdholtz let's get this PR merged in. It's such a big change to spaceship and pilot that the more time and review that goes into it, the more it will diverge from master.

I think the next step will be to get the necessary community testing, perhaps as an alpha release? I am open to community input here for how we should proceed on the releases.

PR #14574, opened 2019-04-11
[spaceship] Make Spaceship::Tunes::DisplayFamily main data source for screenshot information

liamnichols:
### Checklist

### Motivation and Context
As part of work to address the issue raised in #14116, I am attempting to consolidate all of the information around screenshot device types into a single place so that some core functionality in deliver can be fixed. Currently, deliver is maintaining its own definitions of common definitions (screenshot resolutions) that are also used internally inside the spaceship module. Since these values all come from App Store Connect, it makes sense to define them centrally within Spaceship. 

I've made similar changes in an experimental branch aimed at understanding the work required to actually address the bug, you can see those changes here: https://github.com/liamnichols/fastlane/compare/master...experimental-fix. Note that they're a bit of a mess, that's why this PR is isolating a specific change and hopefully making it easy to understand.

### Description
I noticed that spaceship already uses JSON assets to store some mappings/definitions so I decided to follow the same approach. Since there is quite a bit of information for each device type, it was hard to initialise DeviceType instances in a hash in code so it seems to make a bit more sense to extract it out into an easily updatable JSON file. 

However I'm not sure how nice the implementation is for accessing each device type definition so am open to suggestions. 

To be sure that I didn't break anything, I added some tests that ensure the return values for all existing functions remain the same. They don't really make much sense to keep going forward but I will remove them once a core contributor is happy that I'm not going to be introducing any regressions. I hope this is an ok approach. 

The `name` and `category` properties have been added to help in the future to provide more user friendly naming of things since the internal device_types can be very confusing. The `name` value is taken from the App Store Connect web pages and the `category` is just something I populated myself... Open to suggestions on how this should be


spaceship/lib/spaceship/du/du_client.rb
```diff
@@ -99,61 +100,15 @@ def upload_file(app_version: nil, upload_file: nil, path: nil, content_provider_
...
-        iphone6:      "MZPFT.SortedN61ScreenShot",
-        iphone6Plus:  "MZPFT.SortedN56ScreenShot",
-        iphone58:     "MZPFT.SortedD22ScreenShot",
-        iphone65:     "MZPFT.SortedD33ScreenShot",
-        watch:        "MZPFT.SortedN27ScreenShot",
-        watchSeries4: "MZPFT.SortedN131ScreenShot",
-        appleTV:      "MZPFT.SortedATVScreenShot",
-        desktop:      "MZPFT.SortedDesktopScreenShot"
-      }
+      Spaceship::Tunes::DeviceType.all.map { |k, v| [k, v.picture_type] }.to_h
```
janpio: Maybe we should keep a few of the old lines here as a comment to understand what is going to be returned from these methods? (this one and below)

janpio:
Awesome, that definitely makes sense.

Is "DeviceTypes" the correct/best name for this? The current naming is probably deduced from some App Store Connect parameter or API param, which could cause confusion (by reusing the same file that did specific things in the past, by other things).

The .json file is "bad" in such that it doesn't allow us to add comments on where the data comes from. Some of the methods had a small comment in front of them, how one would retrieve those values. (On the other hand, this is probably better suited for some documentation `.md` anyway, isn't it?)

liamnichols:
> Is "DeviceTypes" the correct/best name for this? The current naming is probably deduced from some App Store Connect parameter or API param, which could cause confusion (by reusing the same file that did specific things in the past, by other things).


I did wonder this... The other name idea that I had would have been `ScreenshotDevice` but I avoided it since we could potentially add app preview (video) information to this at a later date and I haven't seen any other names floating about that would make sense.

In terms of where this data lives in the user facing side of App Store Connect, it's all within the "MediaManager" so `MediaDevice` could also be another option maybe?

If you think it makes sense though, I can move this into it's own class with whatever name we choose. 

> The .json file is "bad" in such that it doesn't allow us to add comments on where the data comes from. Some of the methods had a small comment in front of them, how one would retrieve those values. (On the other hand, this is probably better suited for some documentation .md anyway, isn't it?)

Good point... I can add some comments into the associated class (to account for the fact that the json isn't documented) and then a bit more detail in a new .md file in the docs directory if that helps

spaceship/lib/assets/deviceTypes.json
```diff
@@ -0,0 +1,177 @@
...
+    "name": "iPad Pro (2nd Generation) 12.9-inch Display",
+    "category": "iPad",
+    "screenshot_resolutions": [
+      [2048, 2732],
+      [2732, 2048]
+    ],
+    "picture_type": "MZPFT.SortedJ99ScreenShot",
+    "messages_picture_type": "MZPFT.SortedJ99MessagesScreenShot"
+  },
+  {
```
janpio:
```suggestion

  {
```

spaceship/lib/assets/deviceTypes.json
```diff
@@ -0,0 +1,177 @@
...
+    "name": "iPad Pro (3rd Generation) 12.9-inch Display",
+    "category": "iPad",
+    "screenshot_resolutions": [
+      [2048, 2732],
+      [2732, 2048]
+    ],
+    "picture_type": "MZPFT.SortedJ320ScreenShot",
+    "messages_picture_type": "MZPFT.SortedJ320MessagesScreenShot"
+  },
+  {
```
janpio:
```suggestion

  {
```

spaceship/lib/assets/deviceTypes.json
```diff
@@ -0,0 +1,177 @@
...
+    "name": "6.5-inch Display",
+    "category": "iPhone",
+    "screenshot_resolutions": [
+      [1242, 2688],
+      [2688, 1242]
+    ],
+    "picture_type": "MZPFT.SortedD33ScreenShot",
+    "messages_picture_type": "MZPFT.SortedD33MessagesScreenShot"
+  },
+  {
```
janpio:
```suggestion

  {
```

spaceship/lib/assets/deviceTypes.json
```diff
@@ -0,0 +1,177 @@
...
+    "identifier": "watchSeries4",
+    "name": "Series 4",
+    "category": "Apple Watch",
+    "screenshot_resolutions": [
+      [368, 448]
+    ],
+    "picture_type": "MZPFT.SortedN131ScreenShot",
+    "messages_picture_type": null
+  },
+  {
```
janpio:
```suggestion

  {
```

spaceship/lib/assets/deviceTypes.json
```diff
@@ -0,0 +1,177 @@
...
+    "name": "Apple TV",
+    "category": "Apple TV",
+    "screenshot_resolutions": [
+      [1920, 1080],
+      [3840, 2160]
+    ],
+    "picture_type": "MZPFT.SortedATVScreenShot",
+    "messages_picture_type": null
+  },
+  {
```
janpio:
```suggestion

  {
```

janpio:
`DeviceMediaDefinitions`?

spaceship/lib/spaceship/du/du_client.rb
giginet: `Array#to_h` is introduced since Ruby 2.1. Unfortunately, our TargetRubyVersion in rubocop is still 2.0 😭 
Is this OK? 

https://ruby-doc.org/core-2.1.0/Array.html#method-i-to_h

giginet (approved):

spaceship/lib/spaceship/du/du_client.rb
janpio: Hm, good question - if CI doesn't complain we should be good, right?

giginet: Currently, we runs tests only above Ruby 2.3. That is why CI doesn't complain.
To be exact, we should write in Ruby 2.0 compatible code because rubocop currently supports Ruby 2.0.
However, that is tough. So we don't care this 😛

janpio: I expected rubocop to complain on CI when we use stuff we shouldn't. Not?

giginet: rubocop is static analyzer so they can't detect whether a method is defined or not.
This is why rubocop said nothing.
But this code doesn't work on old Ruby at runtime.
There is an only way to notice incompatible is running CI also on old Ruby versions.

In my opinion, currently, there are no CI environment for Ruby 2.0, so we should not care about old Ruby and we should update rubocop ASAP.

janpio: Ok, understood. Best create an issue so we can discuss this and see if there are reasons against it.

liamnichols:
@janpio: I went with `DeviceMediaDefinition` and have added some documentation inline in the new class file and I _think_ that this will be enough to avoid having to add anything in the docs dir? 

Please ♻️ 

With the tests that I added, I planned to just remove them since they won't serve much value going forward but keen to know if you think that there are any other tests that I can replace them with? 🤔

spaceship/lib/spaceship/tunes/device_media_definition.rb
```diff
@@ -0,0 +1,56 @@
+require_relative "../module"
+
+module Spaceship
+  module Tunes
+    class DeviceMediaDefinition
+      # An identifier that identifies unique device types that can
+      # be supplied as app screenshots in App Store Connect.
+      attr_accessor :device_type
```
janpio: These values come from some API param or list, can we describe that somehow?

spaceship/lib/spaceship/tunes/device_media_definition.rb
```diff
@@ -0,0 +1,56 @@
...
+      # Manager in App Store Connect.
+      attr_accessor :name
+
+      # The user friendly category for this defintion (i.e iPhone,
+      # Apple TV or Desktop).
+      attr_accessor :category_name
+
+      # An array of supported screen resolutions (in pixels) that are
+      # supported for the associated device.
+      # https://help.apple.com/app-store-connect/#/devd274dd925
```
janpio:
```suggestion
      # Source: https://help.apple.com/app-store-connect/#/devd274dd925
```

spaceship/lib/spaceship/tunes/device_media_definition.rb
```diff
@@ -0,0 +1,56 @@
...
+      attr_accessor :screenshot_resolutions
+
+      # An internal identifier for the same device definition used
+      # in the DUClient. You can find this by uploading an image in
```
janpio:
```suggestion
      # in the DUClient. 
      # Source: You can find this by uploading an image in
```

spaceship/lib/spaceship/tunes/device_media_definition.rb
```diff
@@ -0,0 +1,56 @@
...
+      # App Store Connect using your browser and then look for the
+      # X-Apple-Upload-Validation-RuleSets value in the uploads
+      # request headers.
+      attr_accessor :picture_type
+
+      # Similar to `picture_type`, but for iMessage screnshots.
```
janpio:
```suggestion
      # Similar to `picture_type`, but for iMessage screenshots.
```

spaceship/spec/du/du_client_spec.rb
```diff
@@ -60,4 +60,63 @@
   describe "#upload_screenshot" do
     # we voluntary skip tests for this method, it's mostly covered by other functions
   end
+
+  describe "mapping migrations" do
```
janpio:
```suggestion
  # these tests were created when we migrated the mappings from several files
  # to `spaceship/lib/assets/deviceMediaDefinitions.json`. they make sure our logic
  # to parse that file works as expected
  describe "mapping migrations" do
```

janpio:
That starts to look and feel pretty good @liamnichols! 

Left some more comments.

Another thought I had: One info bit that got lost were the comments `# iPhone XS` or `# iPhone XS Max`. What do you think about adding a hash of `devices` to the file? I think that might turn out useful.

liamnichols:
> Another thought I had: One info bit that got lost were the comments # iPhone XS or # iPhone XS Max. What do you think about adding a hash of devices to the file? I think that might turn out useful.

Oh hmm interesting... Potentially we could have a hash of simulator names that map to a definition, do you think that would work? (We'd then be able to use this so that Snapshot can put the screenshots in the correct directories for deliver if we go with that approach)

Something like this: 

```ruby
def self.simulator_name_map
  {
    "iPhone 4s" => for_device_type(:iphone35),
    "iPhone 5" => for_device_type(:iphone4),
    "iPhone 5s" => for_device_type(:iphone4),
    "iPhone 6" => for_device_type(:iphone6),
    "iPhone 6 Plus" => for_device_type(:iphone6Plus),
    "iPhone 6s" => for_device_type(:iphone6),
    "iPhone 6s Plus" => for_device_type(:iphone6Plus),
    "iPhone 7" => for_device_type(:iphone6),
    "iPhone 7 Plus" => for_device_type(:iphone6Plus),
    "iPhone 8" => for_device_type(:iphone6),
    "iPhone 8 Plus" => for_device_type(:iphone6Plus),
    "iPhone SE" => for_device_type(:iphone4),
    "iPhone X" => for_device_type(:iphone58),
    "iPhone Xs" => for_device_type(:iphone58),
    "iPhone Xs Max" => for_device_type(:iphone64),
    "iPhone Xʀ" => nil, # Probably just exclude this?
    "iPad mini (5th generation)" => for_device_type(:ipad),
    "iPad Air (3rd generation)" => for_device_type(:ipad105),
    "iPad 2" => for_device_type(:ipad),
    "iPad Retina" => for_device_type(:ipad),
    "iPad Air" => for_device_type(:ipad),
    "iPad mini 2" => for_device_type(:ipad),
    "iPad mini 3" => for_device_type(:ipad),
    "iPad mini 4" => for_device_type(:ipad),
    "iPad Air 2" => for_device_type(:ipad),
    "iPad (5th generation)" => for_device_type(:ipad),
    "iPad Pro (9.7-inch)" => for_device_type(:ipad),
    "iPad Pro (12.9-inch)" => for_device_type(:ipadPro),
    "iPad Pro (12.9-inch) (2nd generation)" => for_device_type(:ipadPro),
    "iPad Pro (10.5-inch)" => for_device_type(:ipad105),
    "iPad (6th generation)" => for_device_type(:ipad),
    "iPad Pro (11-inch)" => for_device_type(:ipad11),
    "iPad Pro (12.9-inch) (3rd generation)" => for_device_type(:ipadPro129),
    "Apple TV" => for_device_type(:appleTV),
    "Apple TV 4K" => for_device_type(:appleTV),
    "Apple TV 4K (at 1080p)" => for_device_type(:appleTV),
    "Apple Watch - 38mm" => for_device_type(:watch),
    "Apple Watch - 42mm" => for_device_type(:watch),
    "Apple Watch Series 2 - 38mm" => for_device_type(:watch),
    "Apple Watch Series 2 - 42mm" => for_device_type(:watch),
    "Apple Watch Series 3 - 38mm" => for_device_type(:watch),
    "Apple Watch Series 3 - 42mm" => for_device_type(:watch),
    "Apple Watch Series 4 - 40mm" => for_device_type(:watchSeries4),
    "Apple Watch Series 4 - 44mm" => for_device_type(:watchSeries4)
  }
end
```

(I built this list from the "Device Type" part of the output from `xcrun simctl list` using Xcode 10.2)

Maybe the simulator names could be an array in the JSON and then a `from_simulator_name` function could just do a find on the hash of definitions? 

As I'm typing this, I'm starting to wonder if definitions of simulator names should live somewhere else as they don't all map to device_type's (i.e iPhone XR) and device_types don't all map to simulators (i.e Mac) 🤔🤔🤔

liamnichols:
I've just spotted that `device_type` is actually the name of a "display family" 

https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/tunes/app_version.rb#L746

I wonder if I should rename things again 🙈 

- `DeviceMediaDefinition` -> `DisplayFamily`
- `device_type` (in the json and the attribute name) -> `name`
- `name` -> `friendly_name`
- `category_name` -> `friendly_category_name`

But at the same time, `Spaceship::Tunes::AppScreenshot` is mapping this into `device_type` so maybe its simpler just to keep it how it is? 🤔

janpio:
Bingo, exactly where my thoughts went.

Side question: Why is the iPhone XR not represented? (I really stopped caring after my iPhone 8...)

janpio:
> I've just spotted that `device_type` is actually the name of a "display family"

Interesting, but is this more than an internal variable?
https://github.com/fastlane/fastlane/blob/195ca870a0456709191af94796c8c5d504fb1f2e/spaceship/lib/spaceship/tunes/app_version.rb#L744

liamnichols:
> Side question: Why is the iPhone XR not represented? (I really stopped caring after my iPhone 8...)

So the iPhone XR physically has a 6.1 inch display (828 x 1792 pixels) ([source](https://www.paintcodeapp.com/news/ultimate-guide-to-iphone-resolutions)) but doesn't have its own "display family" in App Store Connect. Instead the help document lists it together with the iPhone XS Max but it doesn't support screenshots at that size (lol).

Screenshot 2019-04-13 at 16 01 08

> Interesting, but is this more than an internal variable?

Seems like its the field name in the app_version API response

https://github.com/fastlane/fastlane/blob/195ca870a0456709191af94796c8c5d504fb1f2e/spaceship/spec/tunes/fixtures/app_version.json#L305-L618

janpio:
Thanks for the explanation, XR really is a strange beast: https://en.wikipedia.org/wiki/List_of_iOS_devices#Display (Unfortunately iPad Pro 12.9 3rd gen is missing there)

Seems `displayFamilies.json` would be a good name after all. Anything that would indicate we should _not_ standardize on that?

liamnichols:
> Seems displayFamilies.json would be a good name after all. Anything that would indicate we should not standardize on that?

I don't think so.. The "family" part makes it a bit easier to understand that the focus is on the physical display setup and not just a specific size or device so I think that this will be good going forward. 

It certainly makes a bit more sense as to why the values in `device_type` have no consistency

I'll go ahead and rename thing again 😅

liamnichols:
Ok I think we're in a good place now 😅 please ♻️ @janpio 

Going back to `simulator_name_map`.. I think it will be best to leave it for now in this PR and maybe add something either in Deliver::AppSreenshot or somewhere in Snapshot? Maybe I'll open up an issue with some refactoring tasks around this area and add it as an item on the list?

janpio:
> Maybe I'll open up an issue with some refactoring tasks around this area and add it as an item on the list?

That sounds like a plan, you can also just link to these comments for context.

Will do something else now for a bit and then come back to this PR and check if it all still makes sense.

spaceship/lib/spaceship/tunes/display_family.rb
```diff
@@ -0,0 +1,66 @@
+require_relative "../module"
+
+module Spaceship
+  module Tunes
+    # A definition of differnet styled displays used by devices
```
janpio:
```suggestion
    # A definition of different styled displays used by devices
```

spaceship/lib/spaceship/du/du_client.rb
```diff
@@ -1,4 +1,5 @@
 require_relative '../client'
+require_relative '../tunes/device_type'
```
janpio: should this import display_family directly maybe?

spaceship/lib/spaceship/tunes/device_type.rb
```diff
@@ -1,39 +1,15 @@
+require_relative "display_family"
+
 module Spaceship
   module Tunes
     # identifiers of devices that App Store Connect accepts screenshots for
     class DeviceType
```
janpio: In general we want to deprecate and not use this class any more, correct? 
If so, we could maybe add some output or a comment?
Does fastlane itself still use it or do we only leave it for external users (via `spaceship`)?

spaceship/lib/spaceship/tunes/display_family.rb
```diff
@@ -0,0 +1,66 @@
...
+    # that App Store Connect supports storing screenshots.
+    # Display styles often only vary based on screen resolution
+    # however other aspects of a displays physical apperance are
+    # also factored (i.e if the home indicator is provided via
+    # a hardware button or a software interface).
+    class DisplayFamily
+      # The display family name from the App Store Connect API that
+      # is used to assign `Spaceship::Tunes::AppScreenshot` objects
+      # to the correct display families via its `device_type` property.
```
janpio: `device_type` of `Spaceship::Tunes::AppScreenshot`? This is currently formulated a bit confusing.

spaceship/lib/spaceship/tunes/display_family.rb
```diff
@@ -0,0 +1,66 @@
...
+      attr_accessor :name
+
+      # The user friendly name of this defintion as seen in the Media
+      # Manager in App Store Connect.
```
janpio:
```suggestion
      #
      # Source: Media Manager in App Store Connect.
```

line 17 also has to be edited

spaceship/lib/spaceship/tunes/display_family.rb
liamnichols: `DeviceType#exists?` is used twice in Spaceship and once in Deliver and `DeviceType#types` appears to no longer be used (other than for `exists?`) 

I could introduce something like the following to `DisplayFamily` to replace the current usages:

```diff
- exists = Spaceship::Tunes::DeviceType.exists?(screenshot.device_type)
+ exists = !Spaceship::Tunes::DisplayFamily.for_name(screenshot.device_type.to_sym).nil?
```

Open to suggestions to make that look nicer if you have any 👍

spaceship/lib/spaceship/tunes/device_type.rb
liamnichols: `DeviceType#exists?` is used twice in Spaceship and once in Deliver and `DeviceType#types` appears to no longer be used (other than for `exists?`) 

I could introduce something like the following to `DisplayFamily` to replace the current usages:

```diff
- exists = Spaceship::Tunes::DeviceType.exists?(screenshot.device_type)
+ exists = !Spaceship::Tunes::DisplayFamily.for_name(screenshot.device_type.to_sym).nil?
```

Open to suggestions to make that look nicer if you have any 👍

spaceship/lib/spaceship/tunes/display_family.rb
liamnichols: Hmmm yeah.. How about this: 

> The display family name from the App Store Connect API that is used when uploading or listing screenshots. This value is then assigned to the `Spaceship::Tunes::AppScreenshot#device_type` attribute.

Maybe it doesn't need the second part but I thought it would be useful to explain where it comes from in the rest of the project?

spaceship/lib/spaceship/tunes/device_type.rb
liamnichols: > do we only leave it for external users (via spaceship)?

So I have no way to answer this.. I just had a quick search of the project and it looks like we have a `UI.deprecated` so I guess that I could introduce something like the following:

```ruby
def self.types
  UI.deprecated("Spaceship::Tunes::DeviceType has been deprecated. Use the new Spaceship::Tunes::DisplayFamily class instead")
  return DisplayFamily.all.map(&:name)
end
```

I'm just slightly refining the methods on `DisplayFamily` so that it fits in a bit better with the way they are used in the project.. Will push it shortly

janpio: I think we don't use `UI` in spaceship, as it should also be usable outside of the context of fastlane as a standalone library.

So probably the easier way will be to just add a comment that explains that we don't use this any more (if we actually do and you find a good way to replace its usage what you posted in https://github.com/fastlane/fastlane/pull/14574#discussion_r275161079 looks fine to me).

spaceship/lib/spaceship/tunes/display_family.rb
janpio: Yeah, context is important here - ever pointer helps :) Your suggestion is much better.

spaceship/lib/spaceship/tunes/device_type.rb
liamnichols: So yeah I just realised that 🙈 

After some googling, I saw a suggestion to use `warn` to print a message and I just committed that but I could just change it to a comment if you still think that it's better?

I've gone through the uses in the main project and replaced them with DisplayFamily so in theory we could just delete this file if we wanted as well

deliver/lib/deliver/upload_screenshots.rb
```diff
@@ -134,7 +133,7 @@ def collect_screenshots_for_languages(path, ignore_validation)
       # to have it in there for frameit support
       unaccepted_device_shown = false
       screenshots.select! do |screenshot|
-        exists = Spaceship::Tunes::DeviceType.exists?(screenshot.device_type)
+        exists = !screenshot.device_type.nil?
```
liamnichols: I've changed this to just be a simple nil check since `AppScreenshot` has the responsibility of ensuring that the return value of this attribute is valid, otherwise it will just return `nil`

spaceship/lib/spaceship/tunes/device_type.rb
janpio: > I've gone through the uses in the main project and replaced them with DisplayFamily so in theory we could just delete this file if we wanted as well

Yeah, the problem is that people might use spaceship in isolation as a library, and depend on this file and its methods - that's why I think we should keep it around - otherwise we have a breaking change.

liamnichols: Ah yeah this makes sense... Deprecating seems like a good approach for the time being then. Are you happy with the `warn` message or should I change it to a comment?

janpio: I am perfectly happy. The question is if ruby and other people knowing more about proper ruby etiquette are ;) Will try to get some other people form the team to look at it.

spaceship/lib/spaceship/tunes/device_type.rb
```diff
@@ -1,39 +1,16 @@
...
-        # Apple TV
-        'appleTV',
-
-        # Mac
-        'desktop'
-      ]
-      class << self
-        attr_accessor :types
+      def self.types
+        warn("Spaceship::Tunes::DeviceType has been deprecated. Use the new Spaceship::Tunes::DisplayFamily class instead.")
```
snatchev: I dont think we have a precedent for `warn` in spaceship yet. I think it's a good use of the method.
Could you add either a comment with a URL to this PR, or add the URL to the warn message?

spaceship/lib/spaceship/tunes/display_family.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      # All DisplayFamily types currently supported by App Store Connect.
+      def self.all
+        lookup.values
+      end
+
+      private_class_method
+
+      def self.lookup
+        @lookup ||= JSON.parse(File.read(File.join(Spaceship::ROOT, "lib", "assets", "displayFamilies.json"))).map { |data| [data["name"].to_sym, DisplayFamily.new(data)] }.to_h
```
snatchev: small style nitpick: very long single-line method (did rubocop not catch it?). Could you break this up a little?
If memoization (`||=`) was a motivator for making this a single line, you could try something like this:
```ruby
def self.lookup
  return @lookup if defined?(@lookup)
  @lookup = . . . 
end
```

snatchev (approved):
👍 Looks good to me, just a few style comments :)

spaceship/lib/spaceship/tunes/display_family.rb
liamnichols: Sure! I originally just copied what was in langauge_converter.rb but then added the map to it so thats probably why it got a little lengthy 🙈 

Will break it up a little 🙇

spaceship/lib/spaceship/tunes/device_type.rb
liamnichols: Sure, I updated the message to link to the PR 🙂

liamnichols: Sure, I updated the message to link to the PR 🙂

snatchev:
@liamnichols thanks!

Going to defer to @joshdholtz for merging this. I'd like him to weigh in on how a deprecation should be handled.

janpio (approved):
I absolutely love how this turned out 🥇 Awesome improvement, not only in maintainability but also understandability of the code in general.

(Same re the deprecation via `warn`, lets see what @joshdholtz says)

spaceship/lib/spaceship/tunes/app_version.rb
```diff
@@ -692,7 +692,7 @@ def trailer_data_for_language_and_device(language, device)
       end
 
       def container_data_for_language_and_device(data_field, language, device)
-        raise "#{device} isn't a valid device name" unless DeviceType.exists?(device)
+        raise "#{device} isn't a valid device name" if DisplayFamily.find(device.to_sym).nil?
```
getaaron: I think `unless DisplayFamily.find()` would better match our general style

spaceship/lib/spaceship/tunes/display_family.rb
```diff
@@ -0,0 +1,83 @@
...
+      # Identifies if the device definition supports iMessage screenshots.
+      def messages_supported?
+        true unless messages_picture_type.nil?
+      end
+
+      # Finds a DisplayFamily definition.
+      #
+      # @param name (Symbol) The name of the display family being searched for.
+      # @return DisplayFamily object matching the given name.
+      def self.find(name)
```
getaaron: Should we convert name to a symbol if passed in as a string?

getaaron (approved):
# THIS IS AWESOME

getaaron:
The deprecation message looks fine to me, what’s the question for josh?

spaceship/lib/spaceship/tunes/display_family.rb
liamnichols: I did think about that, but I didn't know how to approach it or if its a common thing to do in Ruby so instead I left it 😅

Do you have a suggestion as to what the implementation should look like? 

`name = name.to_sym if name.is_a?(String)`?

Is there a specific structure for the @param documentation also?

`@param name (Symbol, String) The name...`?

snatchev:
@getaaron the question is more in terms of versioning, releasing, release notes etc. I want to get his perspective on releasing a deprecation.

liamnichols:
FYI @getaaron I've applied your suggestions in https://github.com/fastlane/fastlane/pull/14574/commits/ac13a4c00ec0eecfd510d36450b6f916e2fcf444, thanks for the tips 👍

joshdholtz (approved):
This is 🔥 The deprecation seems good to me! We will ship this out later today 😊 Thanks for the contribution! ❤️

Issue #14474, opened 2019-03-26
Deliver uploading 3rd gen iPad screenshots into 2nd gen iPad slot


scorpion35:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
3rd gen iPad screenshots are uploaded into 2nd gen iPad's slot. Think Apple's API is getting confused (because 3rd gen iPad and 2nd gen iPad screenshot dimensions are same) and sending them to 2nd gen iPad slot, eventually causing doubles in 2nd gen iPad slots.

##### Command executed
`fastlane deliver run -f -u <myusername> --skip_screenshots false -z TBD`

##### Complete output when running fastlane, including the stack trace and command, used

[10:10:50]: Uploading 25 screenshots for language en-US
**[10:10:50]: Uploading './screenshots/en-US/1_12.9inchiPad_EN_app_2732x2048_01.jpg'...**
...
**[10:11:03]: Uploading './screenshots/en-US/2_12.9inchiPad_EN_app_2732x2048_02.jpg'...**
...
**[10:11:13]: Uploading './screenshots/en-US/3_12.9inchiPad_EN_app_2732x2048_03.jpg'...**
...
**[10:11:26]: Uploading './screenshots/en-US/4_12.9inchiPad_EN_app_2732x2048_04.jpg'...**
...
**[10:11:49]: Uploading './screenshots/en-US/5_12.9inchiPad_EN_app_2732x2048_05.jpg'...**
...
[✔] Saving changes 

![output_result](https://user-images.githubusercontent.com/18679225/55007197-61ccd700-4fb5-11e9-9905-16f8a2ab5de4.png)

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14.3                                     |
| Ruby                        | 2.5.1                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.16.2                          |
| Installation Source         | ~/.rbenv/versions/2.5.1/bin/fastlane        |
| Host                        | Mac OS X 10.14.3 (18D109)                   |
[...]

PR #14576, opened 2019-04-12
[deliver] iPad Pro (12.9-inch) (3rd generation) ambiguity resolution

viktorasl:
### Checklist

### Motivation and Context
As iTC now requires iPad Pro 12.9" 3rd generation screenshots (https://help.apple.com/app-store-connect/#/devd274dd925) deliver's inability to upload them has become quite a blocker.
It solves #14474 & #14116.

### Description
Approach is quite simple: if screenshot size matches iPad Pro's then ambiguity is resolved taking into account screenshot name: 3rd gen screenshots has to have `(3rd generation)` part in the name (naming is taken from default simulator names).
Changes have been tested in production, successfully uploaded them! 🚀 Furthermore, HTML preview is generated correctly out-of-the-box. Hopefully solution approach is good enough 🙏 

If changes are approved please describe where ambiguity resolution should be documented 🙇

### DIFF
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -32,7 +32,7 @@ module Deliver
       # iPad Pro
       IOS_IPAD_PRO = "iOS-iPad-Pro"
       # iPad Pro (12.9-inch) (3rd generation)
-      IOS_IPAD_PRO129_3RD = "iOS-iPad-Pro-12.9-3rd-gen"
+      IOS_IPAD_PRO_12_9 = "iOS-iPad-Pro-12.9-3rd-gen"
 
       # iPhone 5 iMessage
       IOS_40_MESSAGES = "iOS-4-in-messages"
@@ -56,7 +56,7 @@ module Deliver
       # iPad Pro iMessage
       IOS_IPAD_PRO_MESSAGES = "iOS-iPad-Pro-messages"
       # iPad Pro (12.9-inch) (3rd generation) iMessage
-      IOS_IPAD_PRO129_3RD_MESSAGES = "iOS-iPad-Pro-12.9-3rd-gen-messages"
+      IOS_IPAD_PRO_12_9_MESSAGES = "iOS-iPad-Pro-12.9-3rd-gen-messages"
 
       # Apple Watch
       IOS_APPLE_WATCH = "iOS-Apple-Watch"
@@ -105,7 +105,7 @@ module Deliver
         ScreenSize::IOS_IPAD_10_5 => "ipad105",
         ScreenSize::IOS_IPAD_11 => "ipadPro11",
         ScreenSize::IOS_IPAD_PRO => "ipadPro",
-        ScreenSize::IOS_IPAD_PRO129_3RD => "ipadPro129",
+        ScreenSize::IOS_IPAD_PRO_12_9 => "ipadPro129",
         ScreenSize::IOS_40_MESSAGES => "iphone4",
         ScreenSize::IOS_47_MESSAGES => "iphone6", # also 7 & 8
         ScreenSize::IOS_55_MESSAGES => "iphone6Plus", # also 7 Plus & 8 Plus
@@ -113,7 +113,7 @@ module Deliver
         ScreenSize::IOS_65_MESSAGES => "iphone65",
         ScreenSize::IOS_IPAD_MESSAGES => "ipad",
         ScreenSize::IOS_IPAD_PRO_MESSAGES => "ipadPro",
-        ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES => "ipadPro129",
+        ScreenSize::IOS_IPAD_PRO_12_9_MESSAGES => "ipadPro129",
         ScreenSize::IOS_IPAD_10_5_MESSAGES => "ipad105",
         ScreenSize::IOS_IPAD_11_MESSAGES => "ipadPro11",
         ScreenSize::MAC => "desktop",
@@ -138,7 +138,7 @@ module Deliver
         ScreenSize::IOS_IPAD_10_5 => "iPad 10.5",
         ScreenSize::IOS_IPAD_11 => "iPad 11",
         ScreenSize::IOS_IPAD_PRO => "iPad Pro",
-        ScreenSize::IOS_IPAD_PRO129_3RD => "iPad Pro (12.9-inch) (3rd generation)",
+        ScreenSize::IOS_IPAD_PRO_12_9 => "iPad Pro (12.9-inch) (3rd generation)",
         ScreenSize::IOS_40_MESSAGES => "iPhone 5 (iMessage)",
         ScreenSize::IOS_47_MESSAGES => "iPhone 6 (iMessage)", # also 7 & 8
         ScreenSize::IOS_55_MESSAGES => "iPhone 6 Plus (iMessage)", # also 7 Plus & 8 Plus
@@ -147,7 +147,7 @@ module Deliver
         ScreenSize::IOS_65_MESSAGES => "iPhone XS Max (iMessage)",
         ScreenSize::IOS_IPAD_MESSAGES => "iPad (iMessage)",
         ScreenSize::IOS_IPAD_PRO_MESSAGES => "iPad Pro (iMessage)",
-        ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES => "iPad Pro (12.9-inch) (3rd generation) (iMessage)",
+        ScreenSize::IOS_IPAD_PRO_12_9_MESSAGES => "iPad Pro (12.9-inch) (3rd generation) (iMessage)",
         ScreenSize::IOS_IPAD_10_5_MESSAGES => "iPad 10.5 (iMessage)",
         ScreenSize::IOS_IPAD_11_MESSAGES => "iPad 11 (iMessage)",
         ScreenSize::MAC => "Mac",
@@ -174,7 +174,7 @@ module Deliver
         ScreenSize::IOS_65_MESSAGES,
         ScreenSize::IOS_IPAD_MESSAGES,
         ScreenSize::IOS_IPAD_PRO_MESSAGES,
-        ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES,
+        ScreenSize::IOS_IPAD_PRO_12_9_MESSAGES,
         ScreenSize::IOS_IPAD_10_5_MESSAGES,
         ScreenSize::IOS_IPAD_11_MESSAGES
       ].include?(self.screen_size)
@@ -306,9 +306,9 @@ module Deliver
       is_3rd_gen = filename.include?("(3rd generation)")
       if is_3rd_gen
         if device_type == ScreenSize::IOS_IPAD_PRO
-          return ScreenSize::IOS_IPAD_PRO129_3RD
+          return ScreenSize::IOS_IPAD_PRO_12_9
         elsif device_type == ScreenSize::IOS_IPAD_PRO_MESSAGES
-          return ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES
+          return ScreenSize::IOS_IPAD_PRO_12_9_MESSAGES
         end
       end
       device_type
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -302,8 +302,8 @@ module Deliver
       }
     end
 
-    def self.resolve_ipadpro_conflict_if_needed(device_type, file_name)
-      is_3rd_gen = file_name.include?("(3rd generation)")
+    def self.resolve_ipadpro_conflict_if_needed(device_type, filename)
+      is_3rd_gen = filename.include?("(3rd generation)")
       if is_3rd_gen
         if device_type == ScreenSize::IOS_IPAD_PRO
           return ScreenSize::IOS_IPAD_PRO129_3RD
@@ -311,7 +311,7 @@ module Deliver
           return ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES
         end
       end
-      return device_type
+      device_type
     end
 
     def self.calculate_screen_size(path)
@@ -322,7 +322,7 @@ module Deliver
       # Walk up two directories and test if we need to handle a platform that doesn't support landscape
       path_filenames = Pathname.new(path).each_filename.to_a
       path_component = path_filenames[-3]
-      file_name = path_filenames.to_a[-1]
+      filename = path_filenames.to_a[-1]
       if path_component.eql?("appleTV")
         skip_landscape = true
       end
@@ -334,12 +334,12 @@ module Deliver
         array.each do |resolution|
           if skip_landscape
             if size[0] == (resolution[0]) && size[1] == (resolution[1]) # portrait
-              return resolve_ipadpro_conflict_if_needed(device_type, file_name)
+              return resolve_ipadpro_conflict_if_needed(device_type, filename)
             end
           else
             if (size[0] == (resolution[0]) && size[1] == (resolution[1])) || # portrait
                (size[1] == (resolution[0]) && size[0] == (resolution[1])) # landscape
-              return resolve_ipadpro_conflict_if_needed(device_type, file_name)
+              return resolve_ipadpro_conflict_if_needed(device_type, filename)
             end
           end
         end
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -31,6 +31,8 @@ module Deliver
       IOS_IPAD_11 = "iOS-iPad-11"
       # iPad Pro
       IOS_IPAD_PRO = "iOS-iPad-Pro"
+      # iPad Pro (12.9-inch) (3rd generation)
+      IOS_IPAD_PRO129_3RD = "iOS-iPad-Pro-12.9-3rd-gen"
 
       # iPhone 5 iMessage
       IOS_40_MESSAGES = "iOS-4-in-messages"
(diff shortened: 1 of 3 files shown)


deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
...
+    def self.resolve_ipadpro_conflict_if_needed(device_type, file_name)
+      is_3rd_gen = file_name.include?("(3rd generation)")
+      if is_3rd_gen
+        if device_type == ScreenSize::IOS_IPAD_PRO
+          return ScreenSize::IOS_IPAD_PRO129_3RD
+        elsif device_type == ScreenSize::IOS_IPAD_PRO_MESSAGES
+          return ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES
+        end
+      end
+      return device_type
```
giginet: You can drop return in Ruby way

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
       }
     end
 
+    def self.resolve_ipadpro_conflict_if_needed(device_type, file_name)
```
giginet: `filename`?

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
...
     def self.calculate_screen_size(path)
       size = FastImage.size(path)
 
       UI.user_error!("Could not find or parse file at path '#{path}'") if size.nil? || size.count == 0
 
       # Walk up two directories and test if we need to handle a platform that doesn't support landscape
-      path_component = Pathname.new(path).each_filename.to_a[-3]
+      path_filenames = Pathname.new(path).each_filename.to_a
+      path_component = path_filenames[-3]
+      file_name = path_filenames.to_a[-1]
```
giginet: `filename`

deliver/spec/app_screenshot_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe Deliver::AppScreenshot do
+  before do
+    allow(FastImage).to receive(:size).and_return([2732, 2048])
+  end
+
+  describe "should resolve iPad Pro (12.9-inch) ambiguity if" do
+    it "has path not including distinction" do
+      screenshot = Deliver::AppScreenshot.new("path/to/screenshot/Screen-Name-iPad Pro (12.9-inch).png", "de-DE")
+      expect(screenshot.screen_size).to eq(Deliver::AppScreenshot::ScreenSize::IOS_IPAD_PRO)
```
giginet: 👍

deliver/spec/app_screenshot_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'deliver/app_screenshot'
+require 'deliver/setup'
+
+describe Deliver::AppScreenshot do
+  before do
+    allow(FastImage).to receive(:size).and_return([2732, 2048])
+  end
+
+  describe "should resolve iPad Pro (12.9-inch) ambiguity if" do
```
giginet: According to better RSpec, you can write specs in this way.

http://www.betterspecs.org/

```ruby
describe "#initialize" do
  context "when filename doesn't contain '(3rd generation)'" do
     it "returns iPad Pro(12.9-inch) 3rd generation" do
     end
  end

  context "when filename contains '(3rd generation)" do
     it "returns iPad Pro(12.9-inch)" do
     end
  end
end
```

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
       }
     end
 
+    def self.resolve_ipadpro_conflict_if_needed(device_type, file_name)
+      is_3rd_gen = file_name.include?("(3rd generation)")
```
giginet: This behavior is tricky!

Could you update documentation about this?
https://github.com/fastlane/docs/blob/master/docs/actions/upload_to_app_store.md

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
...
+        if device_type == ScreenSize::IOS_IPAD_PRO
+          return ScreenSize::IOS_IPAD_PRO129_3RD
+        elsif device_type == ScreenSize::IOS_IPAD_PRO_MESSAGES
+          return ScreenSize::IOS_IPAD_PRO129_3RD_MESSAGES
+        end
+      end
+      return device_type
+    end
+
     def self.calculate_screen_size(path)
```
giginet: I think this method should return all available `device_types`(`[ScreenSize]`).

For example, image which is 2732x2048 is passed, this method will return `[ScreenSize::IOS_IPAD_PRO, IOS_IPAD_PRO129_3RD]`.
And filter this on out of this method.

Current implementation is difficult to scale for same situations in the future.

giginet: should be `private`

giginet (commented):
Great job!
This issue annoys my team! I'm glad to fix this 😄

deliver/lib/deliver/app_screenshot.rb
viktorasl: It will introduce a lot of refactoring of this class (i.e `is_valid?` method, `initialize` now takes a single screen size as an argument). Are we sure we want to change that?

liamnichols: FYI I'm currently trying to work out a new approach that will hopefully see this function removed in the future so maybe it doesn't matter for now? 🤔

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -136,6 +138,7 @@ def formatted_name
         ScreenSize::IOS_IPAD_10_5 => "iPad 10.5",
         ScreenSize::IOS_IPAD_11 => "iPad 11",
         ScreenSize::IOS_IPAD_PRO => "iPad Pro",
+        ScreenSize::IOS_IPAD_PRO129_3RD => "iPad Pro (12.9-inch) (3rd generation)",
```
janpio: `ScreenSize::IOS_IPAD_PRO => "iPad Pro",` above is second generation, correct?
If so, wouldn't it make more sense to leave the `_3RD`?

viktorasl: Nope, the naming is adhoc here.
I try to follow Xcode's simulator naming:
Screenshot 2019-04-13 at 18 21 47
I guess `Pro` here is quite redundant so I suggest renaming (taking into account https://help.apple.com/app-store-connect/#/devd274dd925):
`ScreenSize::IOS_IPAD_PRO` to `ScreenSize::IOS_IPAD_12_9_2ND` and
`ScreenSize::IOS_IPAD_PRO129_3RD` to `ScreenSize::IOS_IPAD_12_9_3RD`
Would it be ok?

janpio: No, these _are_ an iPad Pros - which are a different family of devices.

I guess `IOS_IPAD_PRO_12_9` could be defined as being the 3rd gen (vs. `IOS_IPAD_PRO` being 1st and 2nd gen). Inconsistent, but in a consistent way ;)

viktorasl:
Fixed according to @giginet & @janpio comments. 
I will make PR in `docs` repo now.

deliver/lib/deliver/app_screenshot.rb
viktorasl: Frameit uses `calculate_screen_size`

giginet (approved):
Looks good to me.
Great 🌸

janpio:
You can include docs changes in this PR, the deliver/`upload_to_app_store` docs are generated from this file: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/upload_to_app_store.md.erb

cwagner22:
Any idea when this fix will be released? Thanks.

viktorasl:
Documented ambiguity resolution 15c80c2 hopefully it's clear.

janpio:
> @cwagner22
> Any idea when this fix will be released? Thanks.

Not sure if this will _ever_ be released, this is a very complicated topic. We will have to evaluate if this is the "correct" solution for this problem, and then see if this PR ticks all the boxes.

GandjaFuzz:
cannot release app, because apple required screenshots for iPad Pro 3rd Generation. When this fix will be released ?

pedro380085:
Tested fix locally and works. Seems like a viable solution for now. Potential low risk future conflict include other devices named `3rd generation`.

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
...
     def self.calculate_screen_size(path)
       size = FastImage.size(path)
 
       UI.user_error!("Could not find or parse file at path '#{path}'") if size.nil? || size.count == 0
 
       # Walk up two directories and test if we need to handle a platform that doesn't support landscape
-      path_component = Pathname.new(path).each_filename.to_a[-3]
+      path_filenames = Pathname.new(path).each_filename.to_a
+      path_component = path_filenames[-3]
+      filename = path_filenames.to_a[-1]
```
janpio: Lets replace this strangeness by https://ruby-doc.org/stdlib-2.1.3/libdoc/pathname/rdoc/Pathname.html#method-i-basename, possibly before the `devices.each` in line 333 (and replace this here with what it was before)

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -293,13 +302,27 @@ def self.devices
       }
     end
 
+    def self.resolve_ipadpro_conflict_if_needed(device_type, filename)
+      is_3rd_gen = filename.include?("(3rd generation)")
```
janpio: `ipadPro129` should also be checked here, that's what downloading the screenshots with deliver will produce for iPad Pro 3rd gen screenshots, then those would also be covered.

janpio: I though about asking you to actually check for `iPad Pro (12.9-inch) (3rd generation)` instead of just `(3rd generation)`, but as the `device_type` is also checked it doesn't really matter, does it?

fastlane/lib/fastlane/actions/docs/upload_to_app_store.md.erb
```diff
@@ -704,7 +704,11 @@ You can do the same with folders
 
 In this case, default values for keywords, urls, name and release notes are used in all localizations, but each language has a fully localized description
 
+## iPad Pro 12.9-inch 2nd and 3rd generation ambiguity resolution
```
janpio:
```suggestion
## Uploading screenshots for "iPad Pro (12.9-inch) (3rd generation)"
```

fastlane/lib/fastlane/actions/docs/upload_to_app_store.md.erb
```diff
@@ -704,7 +704,11 @@ You can do the same with folders
...
 
+From March 20, 2019 Apple App Store [requires](https://developer.apple.com/news/?id=03202019a) 2nd and 3rd generation iPad [screenshots](https://help.apple.com/app-store-connect/#/devd274dd925).
```
janpio:
```suggestion
[Starting March 20, 2019 Apple's App Store](https://developer.apple.com/news/?id=03202019a) requires 12.9-inch iPad Pro (3rd generation) screenshots additionally to the iPad Pro 2nd generation [screenshots](https://help.apple.com/app-store-connect/#/devd274dd925). As fastlane historically uses the screenshot dimensions to determine the "display family" of a screenshot, this poses a problem as both use the same dimensions and are recognized as the same device family.
```

fastlane/lib/fastlane/actions/docs/upload_to_app_store.md.erb
```diff
@@ -704,7 +704,11 @@ You can do the same with folders
...
+
+Ambiguity arise as Fastlane checks for screenshot size to identify the device. To solve this issue 3rd generation screenshots must contain `(3rd generation)` part in the name of screenshot.
```
janpio:
```suggestion
To solve this a screenshot of a 12.9-inch iPad Pro (3rd generation) must contain either the string `(3rd generation)` or `ipadPro129` (Apple's internal naming of the display family for the 3rd generation iPad Pro) in its filename to be assigned the correct display family and to be uploaded to the correct screenshot slot in your app's metadata.
```

deliver/spec/app_screenshot_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    end
+
+    context "when filename contains '(3rd generation)" do
+      it "returns iPad Pro(12.9-inch) 3rd generation" do
+        screenshot = Deliver::AppScreenshot.new("path/to/screenshot/Screen-Name-iPad Pro (12.9-inch) (3rd generation).png", "de-DE")
+        expect(screenshot.screen_size).to eq(Deliver::AppScreenshot::ScreenSize::IOS_IPAD_PRO_12_9)
+      end
+    end
+  end
```
janpio: also add test for string `ipadPro129`

janpio:
@liamnichols Could you please take a look at this and see if you can poke any holes in it? Does it miss any case or is it a bad idea because of something I don't see right now? Thanks.

deliver/lib/deliver/app_screenshot.rb
viktorasl: neat!

viktorasl: Yes, filename is checked only if screen size matches iPad Pro's

viktorasl:
Okey, merged `master`, resolved conflicts & fixed based on MR comments 🙂

deliver/spec/app_screenshot_spec.rb
```diff
@@ -13,6 +14,29 @@ def screen_size_from(path)
...
+      end
+    end
+
+    context "when filename contains 'ipadPro129'" do
+      it "returns iPad Pro(12.9-inch) 3rd generation" do
+        screenshot = Deliver::AppScreenshot.new("path/to/screenshot/ipadPro129-AAABBBCCCDDD{2732x2048}.png", "de-DE")
+        expect(screenshot.screen_size).to eq(ScreenSize::IOS_IPAD_PRO_12_9)
+      end
+    end
+  end
```
liamnichols: Just a suggestion for the specs so that it matches the format of the other examples...

`when filename doesn't contain '(3rd generation)'` is already covered on line 58, in `should calculate all 12.9 inch iPad resolutions`.. I wonder if it would be good to add just a single example below that one like the following:

```ruby
it "should calculate all 12.9 inch iPad resolutions (3rd generation)" do
  expect_screen_size_from_file("iPad Pro (12.9-inch) (3rd generation)-01Name{2048x2732}.jpg").to eq(ScreenSize::IOS_IPAD_PRO_12_9)
  expect_screen_size_from_file("iPad Pro (12.9-inch) (3rd generation)-01Name-Landscape{2732x2048}.jpg").to eq(ScreenSize::IOS_IPAD_PRO_12_9)
  expect_screen_size_from_file("1_ipadPro129_1.Filename{2048x2732}.jpg").to eq(ScreenSize::IOS_IPAD_PRO_12_9)
  expect_screen_size_from_file("1_ipadPro129_1.Filename-Landscape{2732x2048}.jpg").to eq(ScreenSize::IOS_IPAD_PRO_12_9)
end
```

It'll also be good to add a similar example for the iMessage screenshots also (below line 130)

liamnichols:
@janpio, the only thing that I am thinking within the context of #14598 is that maybe we should lock down the filename format a bit more as per point 2 in https://github.com/fastlane/fastlane/issues/14598#issuecomment-484864619.

Snapshot writes the filename like `{deviceName}-{screenshotName}.png` and Deliver saves the dowloaded files as `{order}_{displayFamily}_{order}.{originalFilename}`. 

When we come to implementing a longer-term solution, do we want the flexibility for the specifier keywords to be positioned anywhere within the filename or should we require that a specific format is followed? 

I feel like an `include?` check might be a little risky when we go for the less hacky implementation since a display family such as 'watch' or 'desktop' could realistically be part of the user's screenshot name resulting in false-positives (i.e the screenshot `iPhone 8-01WatchList.png` includes both the device name `iPhone 8` and the display family `watch`)

Maybe instead it should be something like:

```ruby
```

viktorasl:
Well, IMHO for now (and probably in a longer run) collision is only in this iPad Pro case. I think it's an overkill to prepare for cases like these.

deliver/spec/app_screenshot_spec.rb
viktorasl: I can change that

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -31,6 +31,8 @@ module ScreenSize
       IOS_IPAD_11 = "iOS-iPad-11"
       # iPad Pro
       IOS_IPAD_PRO = "iOS-iPad-Pro"
+      # iPad Pro (12.9-inch) (3rd generation)
+      IOS_IPAD_PRO_12_9 = "iOS-iPad-Pro-12.9-3rd-gen"
```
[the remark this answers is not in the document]
viktorasl: sure

viktorasl:
On the other hand having a bit stricter protocol might turn our well

joshdholtz (approved):
This is great 🙌  Thanks for adding this! This will go out later tonight in a release 🚀

pedro380085:
Hi guys, just wanted to thank you for your work and contribution, this solves our problem.

janpio:
Please note that we will make this a bit more strict in a [follow up PR](https://github.com/fastlane/fastlane/pull/14653) and release that will require you to add `iPad Pro (12.9-inch) (3rd generation)` instead of just `(3rd generation)` to the file name of those screenshots. So if you have to rename some files now, best choose that string already so you don't have to change it later. Sorry we missed that before this got merged and released.

deliver/spec/app_screenshot_spec.rb
janpio: This unfortunately did not make it into the PR. Could one of your create a follow up PR to improve these tests please? I would really appreciate that. Thanks!

deliver/lib/deliver/app_screenshot.rb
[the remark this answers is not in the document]
janpio: This got taken care of in the follow up PR at https://github.com/fastlane/fastlane/pull/14653

PR #14580, opened 2019-04-12
[action] artifactory: Add a read timeout option

raymondhoagland:
The Artifactory gem supports a configurable option for the read timeout.  The Fastlane action can be extended to pass the argument through to the client.

The [default](https://github.com/chef/artifactory-client/blob/master/lib/artifactory/defaults.rb#L149) read timeout for the client is 120 so I see no reason to set a different default.

### Checklist

### Motivation and Context

Allow configuration of a timeout in case there is a slow connection or other blocker.

### Description

This change just adds a new option to set the read timeout for the Artifactory client.  This is already a configurable in the underlying gem and will just be passed when initializing the client.

### DIFF
--- a/fastlane/lib/fastlane/actions/artifactory.rb
+++ b/fastlane/lib/fastlane/actions/artifactory.rb
@@ -148,7 +148,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :read_timeout,
                                        env_name: "FL_ARTIFACTORY_READ_TIMEOUT",
                                        description: "Read timeout",
-                                       default_value: 120,
+                                       default_value: nil,
                                        optional: true)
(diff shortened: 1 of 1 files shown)


janpio:
Couldn't we not define a default to use the gem default?

raymondhoagland:
Based on [https://github.com/chef/artifactory-client/blob/master/lib/artifactory/client.rb#L71-L75]?  Yeah that looks right.

joshdholtz (approved):
@raymondhoagland This looks good to me! @janpio thanks again for the review while I was away 😍

zlobotron:
Hi
it seems you have missed `is_string: false` in ConfigItem.new

PR #14583, opened 2019-04-13
[snapshot] Introduce option to select screen screenshot method

### LINKED ISSUE #14555: Snapshot screenshoting method to make device screenshot
I'd like to produce screenshots w/ notch (and other elements that are now excluded with current method)

#### Feature Description
Now screenshot is done by making a screenshot of first window, though it captures only the content without notch, extended margins (if layout does not support iPad Pro 3rd generation), etc.

https://github.com/fastlane/fastlane/blob/875e3482906044a4f9806dc869c8813d3a2e9a57/snapshot/lib/assets/SnapshotHelper.swift#L177
has to be replaced with
```swift
let screenshot = XCUIScreen.main.screenshot()
```

But I assume the change should be backwards compatible so this feature should be hidden under some option passed to `capture_ios_screenshots` lane.
I've already prepared a solution so if you'll approve I'll make PR.

Currently I have `SnapshotHelper.swift` modified locally.

viktorasl:
### Checklist

### Motivation and Context
Fixed #14555.

### Description
I didn't find a better way to pass configuration so chose to create `configuration.json` file currently containing only a single param (maybe in the future it could be used to pass any predefined or/and custom params)

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -60,6 +60,11 @@ enum SnapshotError: Error, CustomDebugStringConvertible {
     }
 }
 
+enum ScreenshotMethod {
+    case window
+    case screen
+}
+
 @objcMembers
 open class Snapshot: NSObject {
     static var app: XCUIApplication?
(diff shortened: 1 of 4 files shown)


joshdholtz:
@viktorasl I can't seem to run this to get any different results 🤔 Would you be able to post some sample screenshots of what I should be looking for? 🙃

viktorasl:
The difference is shown in lined issue. Did you try running it on `iPhone X`. The main difference is the notch indicator.

viktorasl:
@joshdholtz any thoughts about this? I can close it now as we started supporting iPad's with notches and we don't need this fix anymore but the PR is still hanging and I would like to determine the future of this.

getaaron:
I'm going to close this since it seems it's not needed and it has several conflicts. Feel free to reopen a PR and tag me if you'd like us to take another look 👍

PR #12235, opened 2018-04-04
[match][spaceship] Fixed matchfile overriding fastfile team_id with bonus spaceship select_team additions

joshdholtz:
Fixes #12146
Starting implementation of #12234 (see this issue for more details and examples)

## Issue
Using action options `verify_block` can have some unintended side effects that will break the rules of [Priorities of parameters and options](https://docs.fastlane.tools/advanced/#priorities-of-parameters-and-options)
1. CLI parameter (e.g. gym --scheme Example) or Fastfile (e.g. gym(scheme: 'Example'))
1. Environment variable (e.g. GYM_SCHEME)
1. Tool specific config file (e.g. Gymfile containing scheme 'Example')
1. Default value (which might be taken from the Appfile, e.g. app_identifier from the Appfile)
1. If this value is required, you'll be asked for it (e.g. you have multiple schemes, you'll be asked for it)

### Example of how this happens
This is mainly a problem when `load_configuration_file` is called outside of standard action loading process - https://github.com/fastlane/fastlane/blob/055f325725ac58d006ba46f0d85c6bb74e98adbc/fastlane/lib/fastlane/actions/sync_code_signing.rb#L7-L16

In `match`'s case above, it will load...
1. Load the `Fastfile` parameter (this is correct)
2. Attempt to load the `Matchfile` parameter but doesn't because `Fastfile` parameter set
3. During `match`'s run, this extra `load_configuration_file` will attempt to load files from `Matchfile` again (and won't override the `Fastfile` value) **but** the `verify_block` method gets called and sets `ENV["FASTLANE_TEAM_ID"]` to a potentially different team id (if the `Matchfile` value is different than the `Fastfile` value) which will cause a different value to be shown in `match`'s summary than what actually get's used by `spaceship`

## Solution
1. `verify_block` should not be used to set environment variables `verify_block` can actually get called multiple times and should really only be used for verifying a value (not setting it)
2. We should stop setting environment variables for `FASTLANE_TEAM_ID`, `FASTLANE_TEAM_NAME`, `FASTLANE_ITC_TEAM_ID`, and `FASTLANE_ITC_TEAM_NAME` for to use with the `spaceship`'s `select_team` method
3. `select_team` should take a parameter for `team_id` and `team_name` when explicitly looking for a team instead of setting a global environment variable that may potentially get overridden

PR #14604, opened 2019-04-16
[action] app_store_build_number: Pass teamID when selecting itunes team

lexorus:
### Checklist

### Motivation and Context

We are using the `LatestTestflightBuildNumberAction` for incrementing our project build number. We are providing all the necessary params and credentials for signing and it was working smooth. Lately, I guess after unification of AppStoreConnect and DeveloperPortal teams, fastlane is asking to specify the team id even if we are passing it within the params.
```
Multiple App Store Connect teams found, please enter the number of the team you want to use:
Note: to automatically choose the team, provide either the App Store Connect Team ID, or the Team Name in your fastlane/Appfile:
Alternatively you can pass the team name or team ID using the `FASTLANE_ITC_TEAM_ID` or `FASTLANE_ITC_TEAM_NAME` environment variable
{...}
```
If we pass the environment variable before launching this fastlane action, everything works. But we are passing it through the parameters, and it should also work. I started looking where the problem is and found that when `AppStoreBuildNumberAction` is preparing `Spaceship::Tunes`, it was not passing the specified `team_id`.
```ruby
Spaceship::Tunes.login(params[:username])
Spaceship::Tunes.select_team #no team_id passed
```
Adding the `team_id` parameter solved the issue.

### Description

There is a pretty small change and it is described above in detail. About how did I test it, I tested within our own project, the next use cases:
1. faslane is automatically `Login to App Store Connect` when a `FASTLANE_ITC_TEAM_ID` env var is set
2. faslane is automatically `Login to App Store Connect` when `team_id` argument is passed to the  
`LatestTestflightBuildNumberAction`
3. If neither of above is done, fastlane is asking for the teamID within the dialog.

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -21,7 +21,7 @@ module Fastlane
       def self.get_build_number(params)
         UI.message("Login to App Store Connect (#{params[:username]})")
         Spaceship::Tunes.login(params[:username])
-        Spaceship::Tunes.select_team(team_id: params[:team_id])
+        Spaceship::Tunes.select_team(team_id: params[:team_id], team_name: params[:team_name])
         UI.message("Login successful")
 
         app = Spaceship::Tunes::Application.find(params[:app_identifier])
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -21,7 +21,7 @@ module Fastlane
       def self.get_build_number(params)
         UI.message("Login to App Store Connect (#{params[:username]})")
         Spaceship::Tunes.login(params[:username])
-        Spaceship::Tunes.select_team
+        Spaceship::Tunes.select_team(team_id: params[:team_id])
         UI.message("Login successful")
 
         app = Spaceship::Tunes::Application.find(params[:app_identifier])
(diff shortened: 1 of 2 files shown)


lexorus:
I signed it!

janpio:
Context for someone reviewing this: latest_testflight_build_number` action internally uses `app_store_build_number` action.

fastlane/lib/fastlane/actions/app_store_build_number.rb
```diff
@@ -21,7 +21,7 @@ def self.run(params)
       def self.get_build_number(params)
         UI.message("Login to App Store Connect (#{params[:username]})")
         Spaceship::Tunes.login(params[:username])
-        Spaceship::Tunes.select_team
+        Spaceship::Tunes.select_team(team_id: params[:team_id])
```
joshdholtz: Can we also add `team_name` in here as well?
```suggestion
        Spaceship::Tunes.select_team(team_id: params[:team_id], team_name: params[:team_name])
```

joshdholtz (changes requested):
While we are in here... we should remove the setting of the `ENV`s from the `verify_block` like we did [here](https://github.com/fastlane/fastlane/pull/12235/files) with that are in the `team_id` and `team_name` options since that is not what it is meant for 🙃

fastlane/lib/fastlane/actions/app_store_build_number.rb
lexorus: Updated with both requested changes.

joshdholtz (approved):
Thanks for making those changes! This is 🔥

PR #14610, opened 2019-04-17
[rubocop] Add a new rule to prevent missing keys on the shared area provided by an action

### LINKED ISSUE #14517: Add a new `rubocop` rule to prevent missing keys on the shared area provided by an action
### Issue Description

It's confusing to figure out what are `SharedValues` keys stored in [`lane_context`](https://docs.fastlane.tools/advanced/lanes/#lane-context) when given fastlane action completes. To get more details about this issue, please have a look at [this thread](https://github.com/fastlane/fastlane/issues/14255). 

`fastlane action action_name` prints `SharedValues` declared in `output` action's method, however not every action declares `SharedValues` in its `output` scope.

To improve it, we could introduce a new `rubocop` rule. The rule would rise an exception whenever the provided `SharedValues` are not declared in `output` method in a given action's code.

**Incorrect declaration:**

```
module Fastlane
    module Actions 
        module SharedValues
            DSYM_PATHS = :DSYM_PATHS
        end
        
        class DownloadDsymsAction < Action
             ...

            def self.output
              [] # missing values here ⚠️
            end

            ....
```

**Correct declaration:**

```
module Fastlane
    module Actions 
        module SharedValues
            DSYM_PATHS = :DSYM_PATHS
        end
        
        class DownloadDsymsAction < Action
             ...

            def self.output
              [
                ['DSYM_PATHS', 'An array to all the zipped dSYM files']
              ]
            end

            ....
```

mollyIV:
### Checklist

### Motivation and Context

> It's confusing to figure out what are `SharedValues` keys stored in [`lane_context`](https://docs.fastlane.tools/advanced/lanes/#lane-context) when given fastlane action completes. To get more details about this issue, please have a look at [this thread](https://github.com/fastlane/fastlane/issues/14255). 

> fastlane action action_name prints SharedValues declared in output action's method, however not every action declares SharedValues in its output scope.

closes https://github.com/fastlane/fastlane/issues/14517

### Description

Added a new `rubocop` rule called `missing_keys_on_shared_area`. 

I wish I could add unit tests according to the [`rubocop` documentation](https://github.com/rubocop-hq/rubocop-rspec), however couldn't get that working. 

```bash
rspec NameError: uninitialized constant ...
```

Would be great, if someone could help me on that ☝️🙏 I really would like to add unit tests.

With the new `missing_keys_on_shared_area` rule, we detected `26 offenses` (fixed in this PR).

```bash
$ 1053 files inspected, 26 offenses detected
```

**Ignored files**

We do ignore the rule for two actions: 
* ~`git_branch` due to array const assignment~ 
* ~`xcode_server_get_assets` due to multiple classes~
* `get_provisioning_profile` due to deprecated `SIGH_UDID`

**Manual testing**

I performed a couple of manual testing: 

✅- the rule **does not** trigger a warning 
🥊- the rules triggers a warning 

+ No `SharedValues` module && no `output` method ✅
+ Empty `SharedValues` module  && no `output` method ✅
+ Single const defined in `SharedValues` && no `output` method 🥊
+ Single const defined in `SharedValues` && empty `output` method 🥊
+ Single const defined in `SharedValues` && non-array content in `output` method 🥊
+ Single const defined in `SharedValues` && empty array in `output` method 🥊
+ Single const defined in `SharedValues` && incorrect key in array in `output` method 🥊
+ Single const defined in `SharedValues` && correct key in array in `output` method ✅
+ Two consts defined in `SharedValues` && one correct key in array in `output` method 🥊
+ Two consts defined in `SharedValues` && two correct keys in array in `output` method ✅
+ Missing consts defined in `SharedValues` && one key in array in `output` method 🥊
+ Missing consts defined in `SharedValues` && two keys in array in `output` method 🥊
+ Missing `SharedValues` module && one key in array in `output` method 🥊

🙇 🙇

### DIFF
--- a/fastlane/lib/fastlane/actions/build_ios_app.rb
+++ b/fastlane/lib/fastlane/actions/build_ios_app.rb
@@ -95,7 +95,7 @@ module Fastlane
       def self.output
         [
           ['IPA_OUTPUT_PATH', 'The path to the newly generated ipa file'],
-          ['DSYM_OUTPUT_PATH', 'Te path to the dSYM files']
+          ['DSYM_OUTPUT_PATH', 'The path to the dSYM files']
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/build_ios_app.rb
+++ b/fastlane/lib/fastlane/actions/build_ios_app.rb
@@ -92,6 +92,13 @@ module Fastlane
         "More information: https://fastlane.tools/gym"
       end
 
+      def self.output
+        [
+          ['IPA_OUTPUT_PATH', 'The path to the newly generated ipa file'],
+          ['DSYM_OUTPUT_PATH', 'Te path to the dSYM files']
+        ]
+      end
+
       def self.return_value
         "The absolute path to the generated ipa file"
       end
--- a/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
+++ b/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
@@ -69,7 +69,7 @@ module Fastlane
 
       def self.output
         [
-          ['FL_CHANGELOG', 'The changelog String generated from the collected Git commit messages']
+          ['FL_CHANGELOG', 'The changelog string generated from the collected git commit messages']
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
+++ b/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
@@ -68,7 +68,9 @@ module Fastlane
       end
 
       def self.output
-        ['FL_CHANGELOG', 'The changelog String generated from the collected Git commit messages']
+        [
+          ['FL_CHANGELOG', 'The changelog String generated from the collected Git commit messages']
+        ]
       end
 
       def self.available_options
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -74,7 +74,7 @@ module Fastlane
 
       def self.output
         [
-          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the new system default keychain']
+          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the default keychain']
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -72,6 +72,12 @@ module Fastlane
         "Create a new Keychain"
       end
 
+      def self.output
+        [
+          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the new system default keychain']
+        ]
+      end
+
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :name,
--- a/fastlane/lib/fastlane/actions/create_pull_request.rb
+++ b/fastlane/lib/fastlane/actions/create_pull_request.rb
@@ -68,6 +68,12 @@ module Fastlane
         "This will create a new pull request on GitHub"
       end
 
+      def self.output
+        [
+          ['CREATE_PULL_REQUEST_HTML_URL', 'The HTML URL to the created pull request']
+        ]
+      end
+
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :api_token,
--- a/fastlane/lib/fastlane/actions/ensure_xcode_version.rb
+++ b/fastlane/lib/fastlane/actions/ensure_xcode_version.rb
@@ -1,9 +1,5 @@
 module Fastlane
   module Actions
-    module SharedValues
-      ENSURE_XCODE_VERSION_CUSTOM_VALUE = :ENSURE_XCODE_VERSION_CUSTOM_VALUE
-    end
-
     class EnsureXcodeVersionAction < Action
       def self.run(params)
         Actions.verify_gem!('xcode-install')
--- a/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
@@ -3,6 +3,7 @@ module Fastlane
     module SharedValues
       SIGH_PROFILE_PATH = :SIGH_PROFILE_PATH
       SIGH_PROFILE_PATHS = :SIGH_PROFILE_PATHS
+      SIGH_UDID = :SIGH_UDID # deprecated
       SIGH_UUID = :SIGH_UUID
       SIGH_NAME = :SIGH_NAME
       SIGH_PROFILE_TYPE = :SIGH_PROFILE_TYPE
@@ -21,7 +22,6 @@ module Fastlane
         Actions.lane_context[SharedValues::SIGH_PROFILE_PATHS] ||= []
         Actions.lane_context[SharedValues::SIGH_PROFILE_PATHS] << path
 
-        # SharedValues::SIGH_UDID is deprecated
         uuid = ENV["SIGH_UUID"] || ENV["SIGH_UDID"] # the UUID of the profile
         name = ENV["SIGH_NAME"] # the name of the profile
         Actions.lane_context[SharedValues::SIGH_UUID] = Actions.lane_context[SharedValues::SIGH_UDID] = uuid if uuid
@@ -52,6 +52,7 @@ module Fastlane
         "KrauseFx"
       end
 
+      # rubocop:disable Lint/MissingKeysOnSharedArea
       def self.output
         [
           ['SIGH_PROFILE_PATH', 'A path in which export certificates, key and profile'],
--- a/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
@@ -57,7 +57,7 @@ module Fastlane
         [
           ['SIGH_PROFILE_PATH', 'A path in which certificates, key and profile are exported'],
           ['SIGH_PROFILE_PATHS', 'Paths in which certificates, key and profile are exported'],
-          ['SIGH_UUID', 'UUID (Universally Unique IDentifier) of an application on a device'],
+          ['SIGH_UUID', 'UUID (Universally Unique IDentifier) of a provisioning profile'],
           ['SIGH_NAME', 'The name of the profile'],
           ['SIGH_PROFILE_TYPE', 'The profile type, can be appstore, adhoc, development, enterprise']
         ]
--- a/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
@@ -55,9 +55,9 @@ module Fastlane
       # rubocop:disable Lint/MissingKeysOnSharedArea
       def self.output
         [
-          ['SIGH_PROFILE_PATH', 'A path in which export certificates, key and profile'],
-          ['SIGH_PROFILE_PATHS', 'Paths in which export certificates, key and profile'],
-          ['SIGH_UUID', 'UUID (Universally Unique IDentifier)'],
+          ['SIGH_PROFILE_PATH', 'A path in which certificates, key and profile are exported'],
+          ['SIGH_PROFILE_PATHS', 'Paths in which certificates, key and profile are exported'],
+          ['SIGH_UUID', 'UUID (Universally Unique IDentifier) of an application on a device'],
           ['SIGH_NAME', 'The name of the profile'],
           ['SIGH_PROFILE_TYPE', 'The profile type, can be appstore, adhoc, development, enterprise']
         ]
--- a/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
@@ -3,7 +3,6 @@ module Fastlane
     module SharedValues
       SIGH_PROFILE_PATH = :SIGH_PROFILE_PATH
       SIGH_PROFILE_PATHS = :SIGH_PROFILE_PATHS
-      SIGH_UDID = :SIGH_UDID # deprecated
       SIGH_UUID = :SIGH_UUID
       SIGH_NAME = :SIGH_NAME
       SIGH_PROFILE_TYPE = :SIGH_PROFILE_TYPE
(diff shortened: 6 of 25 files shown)


rubocop/missing_keys_on_shared_area.rb
```diff
@@ -0,0 +1,82 @@
...
+    class MissingKeysOnSharedArea < RuboCop::Cop::Cop
+      MISSING_KEYS_MSG = "There are missing keys on the shared area. Check constants provided in 'SharedValues' or keys provided in 'output' method in the action's code".freeze
+      MISSING_OUTPUT_METHOD_MSG = "There are declared keys on the shared area 'SharedValues', but 'output' method has not been found".freeze
+
+      attr_writer :shared_values_constants
+      def shared_values_constants
+        @shared_values_constants ||= []
+      end
+
+      def on_module(node)
```
mollyIV: I believe we could simplify the logic of getting constants defined in `SharedModule` module by using `def on_const(node)` however I could get a module name for the constants there. Does anyone have an idea how to do that? 🤔

fastlane/lib/fastlane/actions/ensure_xcode_version.rb
```diff
@@ -1,9 +1,5 @@
 module Fastlane
   module Actions
-    module SharedValues
-      ENSURE_XCODE_VERSION_CUSTOM_VALUE = :ENSURE_XCODE_VERSION_CUSTOM_VALUE
```
janpio: unused?

fastlane/lib/fastlane/actions/xcode_server_get_assets.rb
```diff
@@ -1,5 +1,6 @@
 module Fastlane
   module Actions
+    # rubocop:disable Lint/MissingKeysOnSharedArea
```
joshdholtz: Why are we disabling this here? Can we put a comment in here if we do really want to do this? 😊

fastlane/lib/fastlane/actions/git_branch.rb
```diff
@@ -26,8 +26,11 @@ def self.available_options
         []
       end
 
+      # rubocop:disable Lint/MissingKeysOnSharedArea
```
joshdholtz: Is there a purpose for disabling it here?

rubocop/missing_keys_on_shared_area.rb
joshdholtz: We may need to call upon @milch for help on that? 🙃 🙏

joshdholtz (commented):
This is a cool change! Left come comments about some comments otherwise it seems good to me 🙃

rubocop/missing_keys_on_shared_area.rb
milch: If you want to find all constant assignments (which seems to be what you are doing looking at the `casgn` check) under the module node you should be able to define a [node searcher](https://www.rubydoc.info/gems/rubocop/RuboCop/NodePattern/Macros#def_node_search-instance_method) similar to this: 

```ruby
def_node_search :extract_const_assignment, <<-PATTERN
  (casgn nil? $_ ...)
PATTERN
 ```

It’s been a while since I’ve written a RuboCop though so might be a bit rusty here with the syntax.

File: rubocop/missing_keys_on_shared_area.rb [hunk 5]
```diff
@@ -0,0 +1,82 @@
+require 'rubocop'
+
+module RuboCop
+  module Lint
+    class MissingKeysOnSharedArea < RuboCop::Cop::Cop
+      MISSING_KEYS_MSG = "There are missing keys on the shared area. Check constants provided in 'SharedValues' or keys provided in 'output' method in the action's code".freeze
```
milch: It’d be nice to point to which keys are specifically missing, e.g. output `Found %<key>s In SharedValues but not in output method. Keys in the output method: %<list>s`

rubocop/missing_keys_on_shared_area.rb
```diff
@@ -0,0 +1,82 @@
...
+            _scope, const_name, _value = *child
+            self.shared_values_constants << const_name.to_s if const_name
+          end
+
+          _scope, const_name, _value = *body unless const_name
+          self.shared_values_constants << const_name.to_s if const_name
+        end
+      end
+
+      def on_defs(node)
```
milch: Similarly you could define a node searcher which gives you the output method directly:

```
def_node_searcher :find_output_method, '(def :output ...)'
```

rubocop/missing_keys_on_shared_area.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      def on_class(node)
+        name, _superclass, body = *node
+        add_offense(node, :expression, MISSING_OUTPUT_METHOD_MSG) if body.nil? && self.shared_values_constants.any?
+        return if body.nil?
+
+        check(name, body)
+      end
+
+      def check(_name, node)
```
milch: A more descriptive name would be helpful here, i.e. what is this checking? It looks like this will just check to see if an `output` method is defined. With the above suggestion to use a node searcher for finding that method you may be able to get rid of most of this

milch (commented):
Left some notes regarding node searchers. You can also read up more on the pattern syntax rubocop is using [here](https://docs.rubocop.org/en/latest/node_pattern/)

rubocop/missing_keys_on_shared_area.rb
mollyIV: Thank you @milch 😍that works very well, I changed the code!

fastlane/lib/fastlane/actions/ensure_xcode_version.rb
mollyIV: Indeed, couldn't find any usage of it:

image

fastlane/lib/fastlane/actions/xcode_server_get_assets.rb
mollyIV: The issue was related to nested classes - we were checking for the `output` method in both classes. 

I improved the rubocop rule and now we are only checking for an `output` method in a class which inheritances from `Action` class 💪 

```ruby
def on_class(node)
        name, superclass, body = *node
        return unless superclass
        puts superclass.loc.name.source
        return unless superclass.loc.name.source == 'Action'

       ...
```

fastlane/lib/fastlane/actions/git_branch.rb
mollyIV: With @milch suggestion, I was able to remove it 🙌

fastlane/lib/fastlane/actions/ensure_xcode_version.rb
janpio: Ok, so these are leftovers. Did you investigate when or how those got orphaned? Always seems strange when something like this is in the codebase.

rubocop/missing_keys_on_shared_area.rb
mollyIV: Hey @milch 🙂 

Thanks for your suggestions 🙇 Unfortunately this one does not work. Please note we are trying to detect a static `output` method. I was playing with:

```ruby
def_node_matcher :find_output_method, <<-PATTERN
        (defs (self) :output ...)
      PATTERN
```

Although 

```ruby
def on_class(node)
  output = find_output_method(node)
```

gave me nothing ☝️ 

Any idea? 🤔 🙏

rubocop/missing_keys_on_shared_area.rb
mollyIV: Hey @milch 🙂 

Thanks for your suggestions 🙇 Unfortunately this one does not work. Please note we are trying to detect a static `output` method. I was playing with:

```ruby
def_node_matcher :find_output_method, <<-PATTERN
        (defs (self) :output ...)
      PATTERN
```

Although 

```ruby
def on_class(node)
  output = find_output_method(node)
```

gave me nothing ☝️ 

Any idea? 🤔 🙏

rubocop/missing_keys_on_shared_area.rb
mollyIV: I think searching for a node this way is the same as:

```
def on_defs(node) 
  ...
  
  _definee, method_name, _args, body = *node
  return unless method_name.to_s == 'output'
```

We still need to check if the body of the method is not empty and if the body type of the method is an array. 

So I would leave it as it is, WDYT? 🤔

rubocop/missing_keys_on_shared_area.rb
mollyIV: Good idea 💪 Fixed with the newest commit 🙇

mollyIV:
👋 @joshdholtz @janpio @milch 

Thanks for all your help and suggestions! I implemented the requested changes, could you please have a second 👀 🙇 🙇

fastlane/lib/fastlane/actions/build_ios_app.rb
```diff
@@ -92,6 +92,13 @@ def self.details
         "More information: https://fastlane.tools/gym"
       end
 
+      def self.output
+        [
+          ['IPA_OUTPUT_PATH', 'The path to the newly generated ipa file'],
+          ['DSYM_OUTPUT_PATH', 'Te path to the dSYM files']
```
janpio:
```suggestion
          ['DSYM_OUTPUT_PATH', 'The path to the dSYM files']
```

fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
```diff
@@ -68,7 +68,9 @@ def self.details
       end
 
       def self.output
-        ['FL_CHANGELOG', 'The changelog String generated from the collected Git commit messages']
+        [
+          ['FL_CHANGELOG', 'The changelog String generated from the collected Git commit messages']
```
janpio:
```suggestion
          ['FL_CHANGELOG', 'The changelog string generated from the collected git commit messages']
```

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -72,6 +72,12 @@ def self.description
         "Create a new Keychain"
       end
 
+      def self.output
+        [
+          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the new system default keychain']
```
janpio: `ORIGINAL` vs. `new`? Is this correct?

fastlane/lib/fastlane/actions/create_pull_request.rb
```diff
@@ -68,6 +68,12 @@ def self.description
         "This will create a new pull request on GitHub"
       end
 
+      def self.output
+        [
+          ['CREATE_PULL_REQUEST_HTML_URL', 'The HTML URL to the created pull request']
```
janpio: What is a HTML URL?

fastlane/lib/fastlane/actions/get_provisioning_profile.rb
```diff
@@ -52,6 +52,17 @@ def self.author
         "KrauseFx"
       end
 
+      # rubocop:disable Lint/MissingKeysOnSharedArea
+      def self.output
+        [
+          ['SIGH_PROFILE_PATH', 'A path in which export certificates, key and profile'],
```
janpio: grammar
"are exported" maybe?

fastlane/lib/fastlane/actions/get_provisioning_profile.rb
```diff
@@ -52,6 +52,17 @@ def self.author
...
+          ['SIGH_PROFILE_PATHS', 'Paths in which export certificates, key and profile'],
```
janpio: see above

fastlane/lib/fastlane/actions/get_provisioning_profile.rb
```diff
@@ -52,6 +52,17 @@ def self.author
...
+          ['SIGH_UUID', 'UUID (Universally Unique IDentifier)'],
```
janpio: of what?

fastlane/lib/fastlane/actions/xcode_install.rb
```diff
@@ -71,7 +71,7 @@ def self.available_options
 
       def self.output
         [
-          ['XCODE_INSTALL_CUSTOM_VALUE', 'A description of what this value contains']
+          ['XCODE_INSTALL_XCODE_PATH', 'The path to the installed Xcode']
```
janpio:
```suggestion
          ['XCODE_INSTALL_XCODE_PATH', 'The path to the newly installed Xcode']
```

janpio (approved):
Besides the comments I left, this looks great to me.

fastlane/lib/fastlane/actions/create_keychain.rb
mollyIV: Nice catch, according to the comment in the code:

> # if there is no default keychain - setting the original will fail - silent this error

we should change the description to:

```
['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the default keychain']
```

fastlane/lib/fastlane/actions/create_pull_request.rb
mollyIV: According to [GitHub documentation](https://developer.github.com/v3/pulls/#create-a-pull-request) `html_url` is an URL to the new newly created pull request.

fastlane/lib/fastlane/actions/get_provisioning_profile.rb
```diff
@@ -55,9 +55,9 @@ def self.author
       # rubocop:disable Lint/MissingKeysOnSharedArea
       def self.output
         [
-          ['SIGH_PROFILE_PATH', 'A path in which export certificates, key and profile'],
-          ['SIGH_PROFILE_PATHS', 'Paths in which export certificates, key and profile'],
-          ['SIGH_UUID', 'UUID (Universally Unique IDentifier)'],
+          ['SIGH_PROFILE_PATH', 'A path in which certificates, key and profile are exported'],
+          ['SIGH_PROFILE_PATHS', 'Paths in which certificates, key and profile are exported'],
+          ['SIGH_UUID', 'UUID (Universally Unique IDentifier) of an application on a device'],
```
janpio: Does this make sense in the context of sigh/get_provisioning_profile?

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -74,7 +74,7 @@ def self.description
 
       def self.output
         [
-          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the new system default keychain']
+          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the default keychain']
```
janpio: Out of scope for this PR, but:
Why is this added to the output here?

joshdholtz: I think its in scope because its a shared value that wasn't included - https://github.com/fastlane/fastlane/pull/14610/files#diff-0231a345cb1ba3f328959032b555024cR6

fastlane/lib/fastlane/actions/get_provisioning_profile.rb
joshdholtz: Maybe the following makes more sense? 👇 
```suggestion
          ['SIGH_UUID', 'UUID (Universally Unique IDentifier) of a provisioning profile'],
```

joshdholtz (changes requested):
One small change request otherwise 👍

joshdholtz (approved):
This looks good to me! 💪

fastlane/lib/fastlane/actions/create_keychain.rb
janpio: Nah, I meant the question is out of scope: Why does it make sense to output the original default keychain when the action should create a new one?

PR #14620, opened 2019-04-18
[cert] Prevent from crash when keychain_path param wasn't passed

lexorus:
### Checklist

### Motivation and Context

This is probably a regression in `fastlane 2.120.0` that was added in #14462. On `2.119.0` everything works as expected.
When running `CertAction` fastlane was crashing with the next stacktrace:
```
02:01:14 [02:01:14]: An error has occured: no implicit conversion of nil into String
02:01:14 [02:01:14]: /.../fastlane-2.120.0/cert/lib/cert/runner.rb:97:in `expand_path'
02:01:14 [02:01:14]: /.../fastlane-2.120.0/cert/lib/cert/runner.rb:97:in `block in find_existing_cert'
02:01:14 [02:01:14]: /.../fastlane-2.120.0/cert/lib/cert/runner.rb:87:in `each'
02:01:14 [02:01:14]: /.../fastlane-2.120.0/cert/lib/cert/runner.rb:87:in `find_existing_cert'
02:01:14 [02:01:14]: /.../fastlane-2.120.0/cert/lib/cert/runner.rb:38:in `run'
02:01:14 [02:01:14]: /.../fastlane-2.120.0/cert/lib/cert/runner.rb:13:in `launch'
02:01:14 [02:01:14]: /.../fastlane-2.120.0/fastlane/lib/fastlane/actions/get_certificates.rb:17:in `run'
```
This is because we are setting the keychain path using an env var (`CERT_KEYCHAIN_PATH`) rather than passing it as a parameter (`keychain_path`), as it specified in `fastlane cert --help`.

### Description

I just added a check it the param was passed before actually expending the path and passing it to the `CertChecker`.
@jcavar I'm not really sure about why this change was implemented in the first place, so it would be nice if you will also take a look at this PR. As another solution I can check the env var too, and if it was set, proceed with it.

### DIFF
--- a/cert/lib/cert/runner.rb
+++ b/cert/lib/cert/runner.rb
@@ -94,7 +94,7 @@ module Cert
 
         # As keychain is specific to macOS, this will likely fail on non macOS systems.
         # See also: https://github.com/fastlane/fastlane/pull/14462
-        keychain = File.expand_path(Cert.config[:keychain_path])
+        keychain = File.expand_path(Cert.config[:keychain_path]) unless Cert.config[:keychain_path].nil?
         if FastlaneCore::CertChecker.installed?(path, in_keychain: keychain)
           # This certificate is installed on the local machine
           ENV["CER_CERTIFICATE_ID"] = certificate.id


jcavar:
Where are you running this? Is it `macOS` system? 
`keychain_path` is required parameter that has default value, so not sure how that happens.

lexorus:
Yes, on `macOS`

jcavar:
Interesting, how are you invoking `cert`?

jcavar:
What is the value of environment variable that you are passing for keychain path?

lexorus:
Nothing special, just a path to the keychain: `/Users/…/distribution.keychain-db`. Note that passing the same value as a property fixes the problem on `2.120.0`, but it worked on `2.119.0` just with env var, that's why I said that this is a regression. The env var we are using is the one specified when running `fastlane cert --help` -> `-k, --keychain_path STRING Path to a custom keychain (CERT_KEYCHAIN_PATH)`.

jcavar:
Interesting, I am not sure how environment variable to property mapping works? Could there be an issue?

I think that code worked before just because cert was always using default value (which was fixed in that commit).

lexorus:
@jcavar As I understand, previously, the [line that is causing the crash](https://github.com/fastlane/fastlane/pull/14462/files#diff-b4b9b8ba735c52a8fe2212a8f0981c4dL104) was called only if [this condition](https://github.com/fastlane/fastlane/pull/14462/files#diff-b4b9b8ba735c52a8fe2212a8f0981c4dL95) wasn't met. Which seems to be logic, if we have the certificate at a specified path, there is no necessity to find/unlock keychain at all.
I guess this is the real regression rather than "evn var" <-> "param" mapping thing. I guess I mislead you in the PR description.

jcavar:
That is correct but it was working by luck because `if FastlaneCore::CertChecker.installed?(path)` was not using custom keychain that you pass, but default one.

lexorus:
@jcavar I found some time to debug and see how it was working in `2.119.0`. Here are the steps:
1. From `CertAction.run` we got in `Cert::Runner.new.launch`, then `run` and `find_existing_cert` functions of `Cert::Runner`
2. For finding the certificates is used `FastlaneCore::CertChecker`. So `FastlaneCore::CertChecker.installed?(path)` was called, with no keychain passed (line:96)
3. Inside `CertChecker.installed?(path)`, is used `installed_identies` function, where the `in_keychain` property is also optional and nil by default.
4. All the identities are listed using this command `security find-identity -v -p codesigning`, and if there was no keychain passed, it does not affect the identities in any way. Check the function below:
```ruby
    def self.list_available_identities(in_keychain: nil)
      commands = ['security find-identity -v -p codesigning']
      commands << in_keychain if in_keychain
      `#{commands.join(' ')}`
    end
```
5. So, if the found identities include the finger print, the function
```ruby
def self.installed?(path, in_keychain: nil)
  UI.user_error!("Could not find file '#{path}'") unless File.exist?(path)

  ids = installed_identies(in_keychain: in_keychain)
  finger_print = sha1_fingerprint(path)

  return ids.include?(finger_print)
end
```
will return `true`. Which means the keychain is not necessary for `find_existing_cert` function and `CertAction`.

**This is how it was working in `2.119.0`.**
After the update to `2.120.0`, practically nothing changed in the way the certificate is found. The fastlane is just crashing if no `keychain_path` was passed, even if there is no needed certificate.

joshdholtz (approved):
Thanks for making this fix! This will get shipped out today ❤️

jcavar:
@lexorus all you said is correct but the problem was that custom keychain that you pass was ignored. That is why keychain was not necessary.

As you are passing `keychain_path` in your example, it still seems to me that there is some other problem here.

PR #14623, opened 2019-04-18
update LICENSE

snatchev:
This PR was requested by Google's Open Source License reviewers.

While the _fastlane_ brand name is stylized as all lower-case, in this case, it was requested be titlized as such.


joshdholtz (approved):
This seems legit 👍

PR #14643, opened 2019-04-23
[action] Added `format` option to `swiftlint` action

bguidolim:
### Checklist

### Motivation and Context

This PR adds the option `format` to `swiftlint` action.

### Description

The option is only available for mode `autocorrect`, it formats the code besides the changes regarding the rules. If the mode is set to `lint`, then the option has no effect.

### DIFF
--- a/fastlane/lib/fastlane/actions/swiftlint.rb
+++ b/fastlane/lib/fastlane/actions/swiftlint.rb
@@ -92,6 +92,7 @@ module Fastlane
                                        description: 'Fail on warnings? (true/false)',
                                        default_value: false,
                                        is_string: false,
+                                       type: Boolean,
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :files,
                                        description: 'List of files to process',
@@ -102,6 +103,7 @@ module Fastlane
                                                     don't fail the build (true/false)",
                                        default_value: false,
                                        is_string: false,
+                                       type: Boolean,
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :reporter,
                                        description: 'Choose output reporter',
@@ -111,16 +113,18 @@ module Fastlane
                                        description: "Don't print status logs like 'Linting <file>' & 'Done linting'",
                                        default_value: false,
                                        is_string: false,
+                                       type: Boolean,
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :executable,
                                        description: "Path to the `swiftlint` executable on your machine",
                                        is_string: true,
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :format,
-                                        description: "Format code when mode is :autocorrect",
-                                        default_value: false,
-                                        is_string: false,
-                                        optional: true)
+                                       description: "Format code when mode is :autocorrect",
+                                       default_value: false,
+                                       is_string: false,
+                                       type: Boolean,
+                                       optional: true)
         ]
       end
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/swiftlint.rb
```diff
@@ -114,7 +115,12 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :executable,
                                        description: "Path to the `swiftlint` executable on your machine",
                                        is_string: true,
-                                       optional: true)
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :format,
+                                        description: "Format code when mode is :autocorrect",
```
giginet: `indent` 👀

fastlane/lib/fastlane/actions/swiftlint.rb
```diff
@@ -114,7 +115,12 @@ def self.available_options
...
+                                        default_value: false,
```
giginet: You can specify `type: Boolean,`.

fastlane has `Boolean` type.

fastlane/spec/actions_specs/swiftlint_spec.rb
```diff
@@ -1,7 +1,7 @@
 describe Fastlane do
   describe Fastlane::FastFile do
     describe "SwiftLint" do
-      let(:swiftlint_gem_version) { Gem::Version.new('0.9.2') }
+      let(:swiftlint_gem_version) { Gem::Version.new('0.11.0') }
```
giginet: Should you change this value?
This change may affect to exist tests.

I think it is better to override this value only for new test cases.

fastlane/spec/actions_specs/swiftlint_spec.rb
```diff
@@ -289,6 +289,56 @@
...
+            swiftlint(
+              mode: :lint,
+              format: true
+            )
+          end").runner.execute(:test)
+
+          expect(result).to eq("swiftlint lint")
+        end
+
+        it "omits the switch if swiftlint version is too low" do
```
giginet: too earlier? (I'm not a native speaker)

fastlane/spec/actions_specs/swiftlint_spec.rb
```diff
@@ -289,6 +289,56 @@
...
+          allow(Fastlane::Actions::SwiftlintAction).to receive(:swiftlint_version).and_return(Gem::Version.new('0.10.0'))
+
+          result = Fastlane::FastFile.new.parse("lane :test do
+            swiftlint(
+              mode: :autocorrect,
+              format: true
+            )
+          end").runner.execute(:test)
+
```
giginet: You're better to test whether ` UI.important` is called with the message.

giginet (commented):
Good job!

I pointed a few. Please check it out

fastlane/spec/actions_specs/swiftlint_spec.rb
bguidolim: I think so, by default I believe we should use the most recent version, but for tests, I set the latest version comparing with the other options. If I don't change it here, then I need to set the version for every test I made.

fastlane/spec/actions_specs/swiftlint_spec.rb
bguidolim: It's not right, but I basically copied and pasted. I'm gonna change this.

fastlane/spec/actions_specs/swiftlint_spec.rb
bguidolim: I didn't get your point.

joshdholtz (approved):
@bguidolim This is 💯! Thanks for the contribution ❤️

Issue #14626, opened 2019-04-19
[Frameit] Frameit is very slow (15s+ to frame a single image) 


bvirlet:
:key:

### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Frameit is very slow and takes more than 15s to frame a single screenshot.

I have benchmarked the code and found that the main culprit is the command `identify` run on an image to evaluate the trim area.

The regression was apparently introduced here https://github.com/fastlane/fastlane/blame/ca2e2d8f5d479edc4206e3e90610d2f083bdd396/frameit/lib/frameit/editor.rb#L372. 

Potential fixes:
- Avoid this `identify` command
- Apply identify to a smaller image (the issue isn't necessarily identify but that we apply it to a giant image). I think it would even be faster to downscale the image first, calculate the trim box and upscale the trim box _in code_.

PR #14644, opened 2019-04-23
[frameit] Add ability to resume frameit with --resume flag

bvirlet:
:key:

### Checklist

### Motivation and Context

`frameit` can be quite slow (#14626), taking around 10-20 seconds per screenshot sometimes. Sometimes, we just want to re-frame only a subset of the screenshots.

### Description

This PR adds a `--resume` flag to Frameit. When this flag is turned on, screenshots will be reframed only if the framed file doesn't exist yet _or_ if the framed file exists and is older than the screenshot.

This change is backward compatible because the `--resume` flag is false by default.

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
@@ -237,7 +237,7 @@ frameit(
   path: "./fastlane/screenshots",
   force_orientation_block: proc do |filename|
     case filename
-      when "iPad Pro (12.9-inch)-01LoginScreen" 
+      when "iPad Pro (12.9-inch)-01LoginScreen"
         :landscape_right
       when "iPhone 6 Plus-01LoginScreen"
         :portrait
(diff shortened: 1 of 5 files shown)


janpio:
That's a great feature.

Can you add some documentation please? https://docs.fastlane.tools/actions/frameit/#tips, via https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/frame_screenshots.md, would probably be the simplest location to add am explanation about this feature and its benefits.

bvirlet:
@janpio you will find the updated documentation here: https://github.com/fastlane/docs/pull/819

janpio:
Thanks, and sorry for not being clear enough: That change has to be done in this repo here by editing https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/frame_screenshots.md - the file you edited is generated from this one here.

bvirlet:
My bad, I read your previous comment too quickly!

mollyIV (approved):
Love this feature, thanks a bunch @bvirlet <3 

The description in the documentation is very clear 💪 
👏 👏 for an additional code cleanup.

The very minor thing is that relying on file system modification date might be not that accurate because there are antivirus scanners and cloud storage sync tools that continually modify files - in our use case not a big deal 🙂 

I am sure this feature will save much time for the users 🚀

janpio (approved):
Thanks @mollyIV for checking this.
And super thanks to @bvirlet for implementing it in the first place! 🚀

PR #14646, opened 2019-04-23
[pilot] Fix typo in message

pjaspers:
comlpiance → compliance

I have not tested this, I've just changed the typo in a string. I was hoping CI would pick up if this broke something.


pjaspers:
'I signed it!'

janpio (approved):
Good catch!

Issue #14570, opened 2019-04-11
modify_services - impossible to enable data_protection


martinkasa:
### Issue Description
 ```
   modify_services(
      app_identifier: MY_APP_ID,
      team_id: MY_TEAM,
      services: {
        data_protection: "complete"
      }
    )
```

always disable data_protection. There is no way how to enable it via modify_services lane.

##### Complete output when running fastlane, including the stack trace and command used
```
[15:27:43]: -----------------------------
[15:27:43]: --- Step: modify_services ---
[15:27:43]: -----------------------------
[15:27:43]: Starting login with user '*****@icloud.com'
[15:27:44]: Successfully logged in
[15:27:45]: [DevCenter] App found '*****'
[15:27:45]: Disabling services
[15:27:45]:     Data Protection
[15:27:46]: Done! Disabled 1 services.
```

### Environment

<details><summary><b>Loaded gems</b></summary>

| Gem                                             | Version      |
| ----------------------------------------------- | ------------ |
| did_you_mean                                    | 1.3.0        |
| slack-notifier                                  | 2.3.2        |
| atomos                                          | 0.1.3        |
| CFPropertyList                                  | 3.0.0        |
| claide                                          | 1.0.2        |
| colored2                                        | 3.1.2        |
| nanaimo                                         | 0.2.6        |
| xcodeproj                                       | 1.8.1        |
| rouge                                           | 2.0.7        |
| xcpretty                                        | 0.3.0        |
| terminal-notifier                               | 2.0.0        |
| unicode-display_width                           | 1.4.1        |
| terminal-table                                  | 1.8.0        |
| plist                                           | 3.5.0        |
| public_suffix                                   | 2.0.5        |
| addressable                                     | 2.6.0        |
| multipart-post                                  | 2.0.0        |
| word_wrap                                       | 1.0.0        |
| tty-screen                                      | 0.6.5        |
| tty-cursor                                      | 0.6.0        |
| tty-spinner                                     | 0.9.0        |
| babosa                                          | 1.0.2        |
| colored                                         | 1.2          |
| highline                                        | 1.7.10       |
| commander-fastlane                              | 4.4.6        |
| excon                                           | 0.62.0       |
| faraday                                         | 0.15.4       |
| unf_ext                                         | 0.0.7.5      |
| unf                                             | 0.1.4        |
| domain_name                                     | 0.5.20180417 |
| http-cookie                                     | 1.0.3        |
[...]

PR #14571, opened 2019-04-11
[action] allow modify_services to enable data_protection 

martinkasa:
Fixes enabling of services with string values other than "on". Example:
```
    modify_services(
      app_identifier: MY_APP_ID,
      team_id: MY_TEAM,
      services: {
        data_protection: "complete"
      }
    )
```

Problem was that this lane disabled service instead of enable. This commit disables all services with value "off" or false and enables other ones.

fixes #14570

PR #14648, opened 2019-04-23
[action] modify_services: fix "data_protection": "complete"

martinkasa:
There is a bug, when 'complete' value was wrongly considered as value to enable AND value to disable service at the same time. When it is excexuted, it turned on the feature, but immediately turned if off.

This fixes issue #14570 which already has PR #14571 merged, but I found this bug right after the merge was done...

This is still partial solution, because it seems that update_service_for_app method in spaceship/portal/portal_client.rb is not working as expected and it does not actually enable the service in case of data_protection: "complete". Simple boolean values work as expected.

I will check opened issues and if it is not reported yet, I will create a new one.


janpio:
Hm, we should probably improve the test coverage and existing tests of this if it broke even after we reviewed :/ Any idea how to do that?

martinkasa:
I am not sure how to improve it. I am just beginner with ruby. 

But from what I see, problem with current test is, that it compares some object with expected result. BUT services are being enabled/disabled during costruction of this object, so there can be situations when some services are enabled/disabled and it is not reflected in the object which is tested.

mollyIV:
👋 @janpio @martinkasa 

You can find the unit tests in [`modify_services_spec.rb`](https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/modify_services_spec.rb) file. 

Please correct me if I am wrong, but the problem is even though we set `data_protection: "complete"`, the service is enabled, but it gets disabled immediately after it. 

We can test it ☝️ in unit tests like so:

```Ruby
expect(Produce::Service).to receive(:disable) do |options, args|
    expect(options.associated_domains).to eq('off')
    expect(options.apple_pay).to eq('off')
    expect(options.multipath).to eq('off')
    expect(options.data_protection).to be_nil    // we do not expect data_protection being disabled
end
```

I ran the above test without your changes and the test failed:

```Bash
Failures:

  1) Fastlane Fastlane::FastFile modify_services sends enable and disable with strings, symbols, and booleans
     Failure/Error: expect(options.data_protection).to be_nil
     
       expected: nil
            got: "off"
```

Your changes fixes this 👏 👏 

```Bash
Fastlane
  Fastlane::FastFile
    modify_services
      sends enable and disable with strings, symbols, and booleans

Finished in 0.03534 seconds (files took 2.02 seconds to load)
1 example, 0 failures
```

Could you please modify unit tests according to the above suggestion? 🙏 

🙇

fastlane/lib/fastlane/actions/modify_services.rb
```diff
@@ -11,7 +11,7 @@ def self.run(params)
           services = params[:services]
 
           enabled_services = services.select { |_k, v| v == true || v.to_s != 'off' }.map { |k, v| [k, v == true || v.to_s == 'on' ? 'on' : v] }.to_h
-          disabled_services = services.reject { |_k, v| v == true || v.to_s == 'on' }.map { |k, v| [k, 'off'] }.to_h
+          disabled_services = services.select { |_k, v| v == false || v.to_s == 'off' }.map { |k, v| [k, 'off'] }.to_h
```
mollyIV: 👋 @janpio @martinkasa 

You can find the unit tests in [`modify_services_spec.rb`](https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/modify_services_spec.rb) file. 

Please correct me if I am wrong, but the problem is even though we set `data_protection: "complete"`, the service is enabled, but it gets disabled immediately after it. 

We can test it ☝️ in unit tests like so:

```Ruby
expect(Produce::Service).to receive(:disable) do |options, args|
    expect(options.associated_domains).to eq('off')
    expect(options.apple_pay).to eq('off')
    expect(options.multipath).to eq('off')
    expect(options.data_protection).to be_nil    // we do not expect data_protection being disabled
end
```

I ran the above test without your changes and the test failed:

```Bash
Failures:

  1) Fastlane Fastlane::FastFile modify_services sends enable and disable with strings, symbols, and booleans
     Failure/Error: expect(options.data_protection).to be_nil
     
       expected: nil
            got: "off"
```

Your changes fixes this 👏 👏 

```Bash
Fastlane
  Fastlane::FastFile
    modify_services
      sends enable and disable with strings, symbols, and booleans

Finished in 0.03534 seconds (files took 2.02 seconds to load)
1 example, 0 failures
```

Could you please modify unit tests according to the above suggestion? 🙏 

🙇

mollyIV (changes requested):

martinkasa:
I signed it!

martinkasa:
As I wrote in my previous comment. Even when I fixed modify_services action, data_protection is not set to specified value on the Apple portal. Probably HTTP request which is made is not valid anymore.

mollyIV:
Tests on macOS did fail due to:

```
Build-agent version 1.0.11727-b0960fc9 (2019-05-23T02:12:54+0000)
Creating a dedicated VM with xcode:9.0.0 image
failed to create host: Image xcode:9.0.0 is not supported

failed to create host: Image xcode:9.0.0 is not supported
```

Could you please try to rebase `master` and it will run the tests again? 👍

martinkasa:
hmm, I messed up something..

mollyIV:
Please sync a fork first: https://help.github.com/en/articles/syncing-a-fork. 
After that, please rebase master: https://git-scm.com/book/en/v2/Git-Branching-Rebasing

Let me know if you have any questions / problems with that and I will try to provide more info / help ☝️

mollyIV:
Nice, all check passed 🎉 

> Probably HTTP request which is made is not valid anymore.

I think we could open another issue to tackle this problem. 

For now, @janpio I think this PR is ready for the final review 🙂 Looks good to me 👍

fastlane/spec/actions_specs/modify_services_spec.rb
```diff
@@ -15,6 +15,7 @@
           expect(options.associated_domains).to eq('off')
           expect(options.apple_pay).to eq('off')
           expect(options.multipath).to eq('off')
+          expect(options.data_protection).to be_nil
```
janpio: What does this check for `nil` do here? We already have `expect(options.data_protection).to eq('complete')` on the `:enable` expect. Is this the other side of the same thing?

martinkasa: Original implementation is unfortunately written in the way that it is easy possible that services will occur in both lists. After adding tests you have suggested I found another case when it was happening. In following commit I added test cases and fixed the issue.

mollyIV: > What does this check for nil do here?

The bug was about that the `data_protection` was enabled, but immediately after that was disabled. In tests there was a false positive - it passed because we expected `data_protection` to be enabled, however we didn't expect (and didn't check) if it was not disabled. This check prevents us from this scenario. 

>  Should all the other options also be nil here and vice versa?

I think that would be a nice improvement 💪

janpio: Ok, think I got it now. Hard to wrap your around that by just looking at code.

mollyIV: Agree, so maybe to be symmetric in the code, we can add the nil check there and vice versa 👍

martinkasa: I already added nil checks to all cases in my last commit. Maybe you overlooked it.

mollyIV (approved):

janpio (approved):
Thanks for moving this forward @mollyIV and of course huge thanks to @martinkasa for fixing it in the first place.

Issue #14661, opened 2019-04-24
uninitialized constant Spaceship::Pathname (NameError) after upgrading from 2.120.0 to 2.121.1


pparmaei:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject

 After upgrading Fastlane from version 2.120.0 to 2.121.1 using the command
```gem install fastlane -NV```, I am now seeing the error:
```.../.rvm/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/fastlane-2.121.1/spaceship/lib/spaceship/module.rb:2:in `<module:Spaceship>': uninitialized constant Spaceship::Pathname (NameError)```

#### Question Description

We have been using simple ruby scripts to use Spaceship for a while now, the template of such a script is as below: 
```
require "spaceship"
Spaceship::Tunes.login(username, password)
```

Today after upgrading from version 2.120.0 to 2.121.1, all our ruby scripts that require spaceship are failing with the following error:
```
/users/user/.rvm/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/fastlane-2.121.1/spaceship/lib/spaceship/module.rb:2:in `<module:Spaceship>': uninitialized constant Spaceship::Pathname (NameError)
	from /users/user/.rvm/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/fastlane-2.121.1/spaceship/lib/spaceship/module.rb:1:in `<top (required)>'
	from /users/user/.rvm/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/fastlane-2.121.1/spaceship/lib/spaceship/tunes/display_family.rb:1:in `require_relative'
	from /users/user/.rvm/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/fastlane-2.121.1/spaceship/lib/spaceship/tunes/display_family.rb:1:in `<top (required)>'
	from /users/user/.rvm/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/fastlane-2.121.1/spaceship/lib/spaceship/du/du_client.rb:2:in `require_relative'
[...]

PR #14668, opened 2019-04-24
[tests/CI] Test module import

viktorasl:
### Checklist

### Motivation and Context
In the future it would be best to avoid cases like (https://github.com/fastlane/fastlane/issues/14661).

### Description
The idea is to execute `test_modules/run_tests.sh` file which execute all dummy ruby import tests scripts. I'm quite out of ideas now how to produce "test cases" from shell script.

Any input is welcomed! Or I can just close it and open an issue instead.

Still WIP. Made PR just to be able to test CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,25 +128,7 @@ jobs:
   "Lint Source Code":
     docker:
       - image: circleci/ruby:2.3
-    steps:
-      - checkout
-
-      - restore_cache:
-          key: v1-ubuntu-gems-{{ checksum "Gemfile" }}
-      - run:
-          name: Setup Build
-          command: |
-            bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
-      - save_cache:
-          key: v1-ubuntu-gems-{{ checksum "Gemfile" }}
-          paths:
-            - .bundle
-
-      - run: bundle exec fastlane lint_source
-
-  "Run module import tests":
-    docker:
-      - image: circleci/ruby:2.3
+    shell: /bin/bash --login -eo pipefail
     steps:
       - checkout
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,6 @@ jobs:
   "Lint Source Code":
     docker:
       - image: circleci/ruby:2.3
-    shell: /bin/bash --login -eo pipefail
     steps:
       - checkout
 
@@ -143,6 +142,21 @@ jobs:
           paths:
             - .bundle
 
+      - run: bundle exec fastlane lint_source
+
+  "Execute modules load up tests":
+    environment:
+      LC_ALL: C.UTF-8
+      LANG: C.UTF-8
+    docker:
+      - image: circleci/ruby:2.3
+    shell: /bin/bash --login -eo pipefail
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: v1-macos-gems-{{ checksum "Gemfile" }}-{{ checksum "fastlane.gemspec" }}
+
       - run: /bin/bash ./test_modules/run_tests.sh
 
 workflows:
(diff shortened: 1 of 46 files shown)


viktorasl:
Managed to produce failing CI step:
https://circleci.com/gh/fastlane/fastlane/31759?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

janpio:
Awesome!

Can you add this as its own build step in the Circle config? 
What would be an appropriate name for these checks? 
What other checks should this be doing?

viktorasl:
Created a CI step named `Execute module load up tests` without used cache (probably someone should elaborate on that).
So this PR now has MVP structure of how I imagine these tests.
Btw I'm not a ruby guy (nor bash scripting guy) so I would appreciate if someone with ruby experience would take a look if this solution is viable.

.circleci/config.yml
```diff
@@ -144,6 +144,23 @@ jobs:
...
+    docker:
+      - image: circleci/ruby:2.3
+    shell: /bin/bash --login -eo pipefail
+    steps:
+      - checkout
+
+      - run:
+          name: Setup Build
+          command: |
+            gem install $(echo $(gem build fastlane.gemspec) | sed 's/.*File: //')
```
janpio: What does this do?

test_modules/run_tests.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# running tests
+TEST_FILES=(
+  "test_modules/test_spaceship.rb"
+  "test_modules/test_scan.rb"
+)
+EXIT_CODE=0
+for TEST_FILE in ${TEST_FILES[*]} 
+do
+  ruby $TEST_FILE
```
janpio: Per https://stackoverflow.com/a/8795650/252627 you should be able to just do `ruby -e "require 'spaceship'"` here to achieve the same result - and avoid needing those files, but that doesn't work for me locally :/

viktorasl: omg! I searched for something like this and didn't manage to find 😫

.circleci/config.yml
viktorasl: It builds gem from gemspec and installs it right away

janpio: The wonders of shell/bash stuff...

(Might want to add `Build gem from fastlane.gemspec and install locally` or similar as a comment above)

viktorasl:
Tweaked a bit 👍

janpio:
Interesting that `ruby -e` actually works for you and on CI, as it didn't locally on my machine. But whatever 🤷‍♀ 

So this proof of concept definitely works. Now the question is what else should be included here. Maybe all the toplevel modules? (basically the folders in the root of the repo)

Hey Ruby people, what do you think?

snatchev:
@viktorasl looks like you're test works and caught the failure :)

Could you rebase off of master to get this to pass.

I don't quite understand the changes in .circleci.yaml but otherwise LGTM.

janpio:
> I don't quite understand the changes in .circleci.yaml but otherwise LGTM

I can approve those changes, it adds another "job" that builds and installs the local gem and then runs the script - so the circle part looks good.

viktorasl:
Yep, it worked as expected after merge of master (I did not rebase on master, but unless you require I can do that).
If you approve I'll add outstanding modules in the test list.

janpio:
Go for it!

viktorasl:
Done. Works well, but only tests top level module load up (not testing i.e `pem/manager` and similar)

janpio:
It does. 

Let's get this merged then, so please remove the `WIP` from the title @viktorasl.

janpio (approved):

viktorasl:
Ready to be merged. I wanted to style (add colors in terminal for success / failure) tests log but I'll leave it for the future. I have to gain more experience in shell scripting.

Issue #12713, opened 2018-06-07
Add Support for the new App Store Connect API


fabb:
In a few weeks, Apple will release a new API for managing apps and users in the Appstore. This can be used instead of a few current workarounds in Spaceship.

In the first release there will be API for managing provisioning profiles, users, distribute betas and receive analytics. In further releases, metadata and submission should be possible. 

For more information see „Automating App Store Connect“: https://developer.apple.com/wwdc18/303

joshdholtz:
@fabb This is definitely on our radar to do! I was at WWDC this week with a few other core contributors and we ended up attending the App Store Connect API session and talked to some of the Apple engineers. This API will get released near the end of summer and will contain a subset of APIs that we use.

How these APIs will get used within _fastlane_ is still up for discussion. The authentication and request/response format are completely different so we will need to be careful how these new APIs get used within existing and next to current API implementations. We definitely want to use officially supported APIs but we also want to make sure we aren't making the _fastlane_ codebase more complicated to maintain and aren't forcing our users to include two forms of authentication (which would be both the current email/password and the new private key).

So this is 100% on our radar to use but how/when/where it will get used is still TBD 😊

scollings:
Hey fastlane-bot, I suspect you don't want to close this one ;)  Does anyone know if there any other Fastlane feature request that is looking after this implementation that this one can be linked to?

awgeorge:
Curiously, Apple still haven't released it yet 🤔

manimct:
does anyone know whether Apple will continue to authenticate using username and password ?

csknns:
It seems they released the API https://developer.apple.com/app-store-connect/api/

KrauseFx:
Hey @fastlane-bot, not nice :)

fabb:
The App Store Connect API is now available:
https://developer.apple.com/news/?id=11022018

KrauseFx:
It looks the API isn't publicly available yet, but you need to request access. Or am I missing something?

grEvenX:
Seems like it is open? You just need to follow the guides to create an API
key. One of the example resources are documented here:
https://developer.apple.com/documentation/appstoreconnectapi/list_builds

fre. 9. nov. 2018 kl. 11:37 skrev Felix Krause <notifications@github.com>:

rlam3:
Is this still an issue?
When I'm trying to upload first build to app store connect. It tells me 
```
[...]

PR #14690, opened 2019-04-26
[spaceship] class to handle JWT token generation for AppStore connect

snit-tile:
### Checklist

### Motivation and Context

Adds support for JWT authentication required by AppStore Connect API
It will help addressing #12713

### Description
Created a `ConnectAPI::Token` class that manages generation of an JWT access token to authenticate with AppStore Connect API

I've validated it by using the class to generate a token using our combination of Private Key, Issuer ID, and Key ID, then run the curl command described at: https://developer.apple.com/documentation/appstoreconnectapi/generating_tokens_for_api_requests. The request succeeded.

One way I image the token could be used is that `ConnectAPI::Client` could have a `token` method that looks somewhat like this:
```ruby
def token
  if @token.nil? || @token.expired? 
    @token = Token.new(
      key_id: 'some-key-id-provided-by-user',
      issuer_id: 'some-issuer-id-provided-by-user',
      key: OpenSSL::PKey::EC.new(File.read('private-key-file-provided-by-user.pem'))
    ))
  end

  @token
end
```

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -91,6 +91,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('faraday', '~> 0.9') # Used for deploygate, hockey and testfairy actions
   spec.add_dependency('faraday_middleware', '~> 0.9') # same as faraday
   spec.add_dependency('simctl', '~> 1.6.3') # Used for querying and interacting with iOS simulators
+  spec.add_dependency('jwt', '~> 2.1.0') # Used for generating authentication tokens for AppStore connect api
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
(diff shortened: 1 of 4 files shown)


snit-tile:
@snatchev please take a look at this. That generates the token in the required format.
the authorization header to be sent to AppStore connect APIs would look like: "Authorization: Bearer #{token.text}"

joshdholtz (approved):
@snit-tile This is great! Thanks for finding my WIP branch and adding this 😉 I'm gunna click the merge button on this 💪 Hoping we can get this added in to fastlane master here soon enough 🤞

snit-tile:
@joshdholtz, @snatchev, please LMK if there's something else I can help with. Maintaining our automation infra is one of my responsibilities, and helping fastlane is probably the best way to achieve that.

PR #14691, opened 2019-04-27
[pilot] Trim leading zeros in app_version in wait_for_build_processing_to_be_complete

edwardlai3582:
### Checklist

### Motivation and Context
Testflight truncates leading zeros in CFBundleShortVersionString and after wait_for_uploaded_build was deprecated, we can no longer get the original CFBundleShortVersionString.
To solve https://github.com/fastlane/fastlane/issues/13372#issuecomment-486729226

### Description
Compare app_version with leading zeros trimmed

### DIFF
--- a/fastlane_core/lib/fastlane_core/build_watcher.rb
+++ b/fastlane_core/lib/fastlane_core/build_watcher.rb
@@ -27,16 +27,23 @@ module FastlaneCore
 
       private
 
+      # Remove leading zeros ( https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364 )
+      def remove_version_leading_zeros(version: nil)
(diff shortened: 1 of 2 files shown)


illl48:
> I signed it!

I signed it!

janpio:
Ugh, this sounds like a nasty edge case based on an unfortunate implementation by Apple. We should probably add some tests for this, just that it doesn't get lost accidentally in the future.

joshdholtz (changes requested):
@illl48 Heyyy! This is an interesting find 🤔 Do you think you can move this change though to https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/build_watcher.rb and possibly add a test or two? 😊 Moving this to `BuildWatcher` will allow this change to work for other parts of _fastlane_ that might also being experiencing this same issue ❤️

illl48:
@janpio @joshdholtz I moved the change and add a test case. If there are any problems just let me know, thanks

joshdholtz (approved):
@illl48 This is a great fix! Didn't know they allowed leading zeroes and didn't know they removed them 😇 Thanks for the fix!

PR #14702, opened 2019-05-02
Fix failing tests

joshdholtz:
Fixes a failing tests that happened after the merge of https://github.com/fastlane/fastlane/pull/14503 (which had successful tests so 🙃)


janpio:
Damn, one time we don't rebase properly and then there was a PR in the meantime using that removed method again 🤷‍♀

PR #14705, opened 2019-05-02
[WIP] [spaceship] Add authorization header to AppStore Connect api requests

snit-tile:
### Checklist


snit-tile:
@joshdholtz I'm not sure changing the `hostname` here is OK, but it seems like the right thing to do.

One thing I would like some guidance about is the user-input. User has to provide the following info for us to authenticate:
* private key
* key id
* issuer id

I don't know what's the recommendation on how to read/store that info, so for now I'm reading those form environment variables.

snit-tile:
the following snippet has worked for me (making the proper substitutions) using the new hostname

```ruby
require_relative 'spaceship/lib/spaceship/connect_api/client'

ENV["SPACESHIP_CONNECT_KEY_ID"] = 'some value here'
ENV["SPACESHIP_CONNECT_ISSUER_ID"] = 'some value hre'
ENV["SPACESHIP_CONNECT_KEY_FILE"] = '/path/to/key.pem'

client = Spaceship::ConnectAPI::Client.new(cookie: nil, current_team_id: nil)
response = client.get_beta_groups()
p(response)
```

but I had to comment `req.options.params_encoder = Faraday::NestedParamsEncoder`. It's giving me an error for some reason

janpio:
(Rebasing your PR on `master` should get rid of the current test failures)

spaceship/lib/spaceship/connect_api/client.rb
```diff
@@ -347,8 +348,25 @@ def add_beta_groups_to_build(build_id: nil, beta_group_ids: [])
...
       end
 
+      def request(method, url_or_path = nil, params = nil, headers = {}, auto_paginate = false, &block)
+        headers['Authorization'] = "Bearer #{token.text}"
+        super
+      end
+
       protected
 
+      def token
```
joshdholtz: This method also needs a way to pass in `key_id`, `issuer_id`, and `key`. Using environment variable won't be the only way that people will want to use this 🙃

spaceship/lib/spaceship/connect_api/client.rb
```diff
@@ -347,8 +348,25 @@ def add_beta_groups_to_build(build_id: nil, beta_group_ids: [])
         handle_response(response)
       end
 
+      def request(method, url_or_path = nil, params = nil, headers = {}, auto_paginate = false, &block)
+        headers['Authorization'] = "Bearer #{token.text}"
```
joshdholtz: We probably need to add an `if token` around this line because there won't always be a `token` if we are using the web cookie sessions like _fastlane_ uses now 🤔

joshdholtz (changes requested):
This definitely needs to get added to my WIP branch but just a few things to change before we do that 🙃

spaceship/lib/spaceship/connect_api/client.rb
```diff
@@ -15,7 +16,7 @@ class Client < Spaceship::Client
       # Each request method should make only one request. For more high-level logic, put code in the data models.
 
       def self.hostname
-        'https://appstoreconnect.apple.com/iris/v1/'
+        'https://api.appstoreconnect.apple.com/v1/'
```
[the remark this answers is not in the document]
snit-tile: I assumed this version would always use the JWT token and the public and official API. I can make changes to accommodate the different sign-in strategies.

The `get_beta_groups` endpoint while using the iris version was not returning a JSON, it rather gave me an html response (maybe was authentication issue).

spaceship/lib/spaceship/connect_api/client.rb
snit-tile: In this case, what do you think about changing `login` method to accept either user/pwd or key_id/key/issuer_id?

I think the user should provide this info once and the client manages refreshing the token as necessary, that's why I didn't expose the `token` method.

It seems to align with the way to programmatically use spaceship.
1. require
2. login
3. call your methods.

What do you think?

joshdholtz:
@snit-tile Thanks for taking the time to make this PR! Even though I'd love to have merge these changes in for you I'm going to have to close this due to a big recent rewrite of how `Spaceship::ConnectAPI` was written (which does now included the JWS headers). Thanks again, thought! ❤️

snit-tile:
@joshdholtz sorry for not following up on this. I had a surgery, then went on vacations and end up not working for a long time. I’m happy this got solved

Issue #14717, opened 2019-05-06
[scan] Race condition between simulator reset and force starting


jaleksynas:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
When using the `reset_simulator` option and the hidden `FASTLANE_EXPLICIT_OPEN_SIMULATOR` option, one simulator is started _before_ the reset and therefore is (a) not reset and (b) causes reset to throw an error since it is already booted.

##### Command executed
`export FASTLANE_EXPLICIT_OPEN_SIMULATOR=2 fastlane scan`

##### Abbreviated output when running fastlane
```
15:28:11 Killing all running simulators
15:28:13 Resetting iPhone 8
15:28:14 An error was encountered processing the command (domain=com.apple.CoreSimulator.SimError, code=164):
15:28:14 Unable to erase contents and settings in current state: Booted
```

`Killing all running simulators` - https://github.com/fastlane/fastlane/blob/2326f410e26f37632cef0858305f4e62e47dbd84/scan/lib/scan/runner.rb#L31

Reset & error -
https://github.com/fastlane/fastlane/blob/2326f410e26f37632cef0858305f4e62e47dbd84/scan/lib/scan/runner.rb#L43-L46

#####  Fix Ideas.
1. Make the `reinstall_app` and `reset_simulator` conflicting options, add `force_quit_simulators` option.  And the first three steps in runner are:
```
`killall Simulator &> /dev/null` if Scan.config['force_quit_simulators']

Scan.devices.each do |device|
    FastlaneCore::Simulator.reset(udid: device.udid)
end if Scan.config[:reset_simulator]

open_simulator_for_device(Scan.devices.first) if Scan.devices
```

2. Move the if/elsif block for  `reset_simulator`/`reinstall_app` above the open_simulator_for_device call.

janpio:
The two remaining uses of `FASTLANE_EXPLICIT_OPEN_SIMULATOR` look a bit like artifacts that were forgotten to be removed. Should `FASTLANE_EXPLICIT_OPEN_SIMULATOR` be documented? What do you use it for?

jaleksynas:
Resolved by PR #14718

PR #14718, opened 2019-05-06
[scan] reorganize order of operations for simulator management in scan action.

jaleksynas:
### Checklist

### Motivation and Context
Makes simulator management options behave as expected when used together. Suggested fix for: #14717 

### Description
- Move all simulator management operations to the start. 
- Add option and method to force quit simulators
- Add option conflicts so they reflect the `elsif` usage

Ran all automated tests. validated manually that the messaging from the issue did not occur.

### DIFF
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -75,20 +75,12 @@ module Scan
                                      env_name: 'SCAN_FORCE_QUIT_SIMULATOR',
                                      description: "Enabling this option will automatically killall Simulator processes before the run",
                                      default_value: false,
-                                     type: Boolean,
-                                     conflicting_options: [:reinstall_app],
-                                     conflict_block: proc do |value|
-                                       UI.user_error!("You can't use 'force_quit_simulator' and 'reinstall_app' options in one run, the simulator will not be running to uninstall the app")
-                                     end),
+                                     type: Boolean),
         FastlaneCore::ConfigItem.new(key: :reset_simulator,
                                      env_name: 'SCAN_RESET_SIMULATOR',
                                      description: "Enabling this option will automatically erase the simulator before running the application",
                                      default_value: false,
-                                     type: Boolean,
-                                     conflicting_options: [:reinstall_app],
-                                     conflict_block: proc do |value|
-                                       UI.user_error!("You can't use 'reinstall_app' and 'reset_simulator' options in one run, the reset will uninstall the app too")
-                                     end),
+                                     type: Boolean),
         FastlaneCore::ConfigItem.new(key: :prelaunch_simulator,
                                      env_name: 'SCAN_PRELAUNCH_SIMULATOR',
                                      description: "Enabling this option will launch the first simulator prior to calling any xcodebuild command",
@@ -99,11 +91,7 @@ module Scan
                                      env_name: 'SCAN_REINSTALL_APP',
                                      description: "Enabling this option will automatically uninstall the application before running it",
                                      default_value: false,
-                                     type: Boolean,
-                                     conflicting_options: [:reset_simulator],
-                                     conflict_block: proc do |value|
-                                       UI.user_error!("You can't use 'reinstall_app' and 'reset_simulator' options in one run, the reset will uninstall the app too")
-                                     end),
+                                     type: Boolean),
         FastlaneCore::ConfigItem.new(key: :app_identifier,
                                      env_name: 'SCAN_APP_IDENTIFIER',
                                      optional: true,
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -2,8 +2,8 @@ require 'fastlane_core/configuration/config_item'
 require 'credentials_manager/appfile_config'
 require_relative 'module'
 
+# rubocop:disable Metrics/ClassLength
 module Scan
-  # rubocop:disable Metrics/ClassLength
   class Options
     def self.verify_type(item_name, acceptable_types, value)
       type_ok = [Array, String].any? { |type| value.kind_of?(type) }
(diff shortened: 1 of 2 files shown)


scan/lib/scan/options.rb
```diff
@@ -70,19 +70,30 @@ def self.available_options
...
+                                     description: "Enabling this option will automatically killall Simulator processes before the run",
+                                     default_value: false,
+                                     type: Boolean),
         FastlaneCore::ConfigItem.new(key: :reset_simulator,
                                      env_name: 'SCAN_RESET_SIMULATOR',
                                      description: "Enabling this option will automatically erase the simulator before running the application",
                                      default_value: false,
-                                     type: Boolean),
-
-        # reinstall app
```
janpio: this structuring comment is still valid, right? please keep this.

scan/lib/scan/options.rb
```diff
@@ -70,19 +70,30 @@ def self.available_options
                                      type: Boolean,
                                      optional: true),
 
-        # reset simulator
```
janpio: this structuring comment is not valid any more as `force_quit_simulator` was added and "belongs" to `reset_simulator` thematically, correct? If so, what would be a new description of what these to do? "before simulator usage" maybe?

janpio:
Good catch re the conflicting options! 

Left two very nitpicky comments re the comment.

scan/lib/scan/options.rb
jaleksynas: I think it is a group of options (rather than simply restating the name of a single option).
I added a ` simulator management` structural comment for the related three options.

janpio: Yes, before there were two different "group" comments, one just repeating the obvious as I couldn't find a better label. What you did now think is fine and matches the intention behind those options just fine.

scan/lib/scan/runner.rb
```diff
@@ -23,12 +23,28 @@ def run
     end
 
     def test_app
+      force_quit_simulator_processes if Scan.config[:force_quit_simulator]
+
+      if Scan.config[:reset_simulator]
```
jaleksynas: Should simulators be forced stopped here to avoid spurious errors?

scan/lib/scan/runner.rb
```diff
@@ -163,10 +166,15 @@ def copy_simulator_logs
...
-      UI.message("Killing all running simulators")
-      `killall Simulator &> /dev/null`
+      # For backward compatibility force the reset again without regard for the option.
+      force_quit_simulator_processes
 
       FastlaneCore::Simulator.launch(device)
     end
+
+    def force_quit_simulator_processes
+      UI.message("Killing all running simulators")
```
jaleksynas: Should this change to `other_action.sh({log:false])`  Then there is no log impacts unless debug mode is enabled.

jaleksynas:
I included some of the details mentioned in #14721 as well.

janpio:
(No need to force push and squash commits, we squash on merge anyway so it is easier to follow PRs with incremental commits)

raymondhoagland:
From the changes/description the behavior it looks like we're going to is

## Options
| Name                 | Behavior                                   | Conflicts                            | Workflow                                        |
|----------------------|--------------------------------------------|--------------------------------------|-------------------------------------------------|
| force_quit_simulator | Kill all running simulators                | reinstall_app                        | This happens before any other mgmt              |
| reset_simulator      | Reset a set of emulators (set by :devices) | reinstall_app                        | After killing any running simulators (optional) |
| prelaunch_simulator  | Launch a simulator (legacy behavior?)      | none                                 | After killing/resetting simulators (optional)   |
| reinstall_app        | Reinstall the app on a running simulator   | force_quit_simulator | After prelaunch (optional)                      |

Based on that chart, behavior I'd expect to see falls into a couple categories as follows:


1) Emulator was already running

| Option               | Value | Behavior                                        |
|----------------------|-------|-------------------------------------------------|
| force_quit_simulator | True  | Kill the running simulator                      |
| force_quit_simulator | False | no-op                                           |
| reset_simulator      | True  | Error (rescued) unless force_quit first, otherwise reset  |
| reset_simulator      | False | no-op                                           |
| prelaunch_simulator  | True  | Error unless force_quit first, otherwise launch |
| prelaunch_simulator  | False | no-op                                           |
| reinstall_app        | True  | Error if force_quit used (and not prelaunch), otherwise reinstall   |
| reinstall_app        | False | no-op                                           |

2) Emulator not running

| Option               | Value | Behavior                                         |
|----------------------|-------|--------------------------------------------------|
| force_quit_simulator | True  | no-op (kill)                                     |
| force_quit_simulator | False | no-op                                            |
| reset_simulator      | True  | Reset the simulators                             |
| reset_simulator      | False | no-op                                            |
| prelaunch_simulator  | True  | Launch the simulator                             |
| prelaunch_simulator  | False | no-op                                            |
| reinstall_app        | True  | Error unless prelaunch used, otherwise reinstall |
| reinstall_app        | False | no-op                                            |

The conflict between force_quit_simulator and reinstall_app is true _unless_ prelaunch is also used to start the simulator.  For example, if the user specifies all four options, the simulator will be quit, reset, launched, and then the app will be reinstalled, which should still work -- so there isn't necessarily a conflict.  Do we err on the side of expecting people to follow a "preferred" flow of just calling reinstall with an _existing_ simulator?  Or do we sidestep this and allow builds to fail if the user follows an invalid workflow e.g. force_quit & reinstall_app only?

jaleksynas:
@raymondhoagland I believe that less opinionated is better, I'd be glad to remove the conflict blocks since there are combinations that allow for proper functioning.  

@joshdholtz @janpio  Any thoughts or direction would be appreciated so this PR can progress.

janpio (approved):
To be honest this is a bit over my head. What @raymondhoagland summarized looks good to me, and you obviously spent a lot more time thinking this through than I ever could while reviewing. Do you see any further need to change things here, removing the conflict stuff again?

raymondhoagland:
That last change addresses my concerns :) lgtm!

joshdholtz (approved):
This is super clean 💯 Thanks for putting the time in to add this and thanks to @janpio for all the reviews! ❤️

PR #14738, opened 2019-05-10
[spaceship] add a check in app_version.setup_screenshots to also check if there is an app_preview already uploaded

xpy:
### Checklist

### Motivation and Context
When trying to do any update, fastlane will first update the `scaled` flags of the display_families to `true` **unless** they have a screenshot. This doesn't take in account that a trailer (app preview) might be present instead. As a result, the `scaled` flag is set to `true` for said display_family and upon saving, the server complains that *Your app preview can't be uploaded because you have selected to use an app preview from a larger display size*.

This happens when a user tries to do a deliver, even with the `skip_screenshots` flags set to `true`. And that is because Fastlane will send the whole configuration again in order to update any part of it. The prerequisites are that there should be **only video previews** in a display_family that can be set as scalable. It might also happen if the `overwrite_screenshots` flag is also set to `true`.

### Description
There is one more check added to check if the `trailers` key is present and the `value` key of it has contents.

### DIFF
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -708,7 +708,7 @@ module Spaceship
       end
 
       def setup_screenshots
-        # Enable Scaling for all screen sizes that don't have at least one screenshot
+        # Enable Scaling for all screen sizes that don't have at least one screenshot or at least one trailer (app_preview)
         # We automatically disable scaling once we upload at least one screenshot
         language_details = raw_data_details.each do |current_language|
           language_details = (current_language["displayFamilies"] || {})["value"]
--- a/spaceship/lib/spaceship/tunes/app_version.rb
+++ b/spaceship/lib/spaceship/tunes/app_version.rb
@@ -713,8 +713,13 @@ module Spaceship
         language_details = raw_data_details.each do |current_language|
           language_details = (current_language["displayFamilies"] || {})["value"]
           (language_details || []).each do |device_language_details|
+            # Do not enable scaling if a screenshot already exists
             next if device_language_details["screenshots"].nil?
             next if device_language_details["screenshots"]["value"].count > 0
+
+            # Do not enable scaling if a trailer already exists
+            next if device_language_details["trailers"].nil?
+            next if device_language_details["trailers"]["value"].count > 0
 
             # The current row includes screenshots for all device types
             # so we need to enable scaling for both iOS and watchOS apps


xpy:
I signed it!

janpio:
The comment at the beginning of the method might need an update as well.

Can you also explain the bigger picture here? What would a user do to trigger this code?

max-ott:
Hey @xpy, looks like a great PR. What’s the fastest way to test this? Can I directly test it with spaceship?

xpy:
@max-ott Hello! I have to admit that I only tested around my case and I didn't dive into other cases. What I do is `fastlane deliver --skip_binary_upload true --run_precheck_before_submit false --force true --overwrite_screenshots true` I also tried with `--skip_screenshots` and `--skip_metadata` and still have the same problem. And on the App Store Connect I have only one app preview video on the iPhone 5.5'', but I suspect it would also trigger for any display size that can be checked as "scalable".

itchingpixels:
hi! we're encountering the same issue, is there any chance this will be merged? thank you!

max-ott:
It needs more tests. We can’t simply merge each PR without enough tests and reviews from maintainers. As there were no other ways presented to easily test and debug this, its still waiting for more input from the author.

xpy:
@max-ott I already mentioned that I tested for the very case that the PR is about. I also described the way to reproduce, if it seems difficult to reproduce, it probably is because it is indeed a very specific case that is difficult to reproduce. I am not sure what else you expect from me to do to validate this PR.

max-ott:
@xpy as this changes `spaceship` behaviour it would be useful to have steps to reproduce via `spacecship`. PRs require tests from other people than the initial contributor and if no one else can test it, it won't get merged.

Easiest way would be to provide steps for the `spaceship` playground / IRB. (https://github.com/fastlane/fastlane/blob/master/spaceship/docs/AppStoreConnect.md)

max-ott:
I've just had some time to review the code. I'm not sure if it's the right place to put the code there and it solves all issues. The issue at the moment with the app preview upload in general seems to be related to the `scaled` property.

We need to disabled the `scaled` property for the `display_family`, if:
- There are screenshots present (probably the case right now) 
- There are app previews present
- If you are trying to upload an app preview
- If you are trying to upload a screenshot (probably the case right now)

mackoj:
This fix, fixes our CI has anyone have any issue with it ?

rabbitinspace:
fixed #15036 for me as well

mackoj:
Hi @joshdholtz,
Could you review this PR please ?

joshdholtz:
Ah yes yes! I will do some review on this this week!  Thanks for friendly ping ❤️

mackoj:
How could we help with this PR ? Did you get time to review it @joshdholtz ?

joshdholtz:
@mackoj Thanks for ping on this! Got lost in the shuffle 😔 Bumping this to top of list for rest of this week

joshdholtz (approved):
This looks great! Sorry it took so long to get through 😔 I'm really trying to improve my personal processes to keep conversations and merges going. Sorry to everyone that needed this but really appreciate the contribution! ❤️

PR #14766, opened 2019-05-15
[supply] Change default track from production to internal

awgeorge:
Change default track value to `internal` to mitigate accidentally releasing an APK to production.

Google presently doesn't allow the cancellation of a `Full Rollout` meaning that the default track `production` can not be cancelled, if the script was run in error.

I believe it would be best to require the developer to force the track to production, therefore explicating giving acknowledgement that this behaviour will occur.

### Checklist

### Motivation and Context

While trying to debug a problem with Supply, I accidentally released an app to production. I think it would be better to mitigating this problem by setting the default value to `internal`

### Description

I've changed the default from `Production` to `Internal`

### DIFF
--- a/supply/lib/supply/options.rb
+++ b/supply/lib/supply/options.rb
@@ -18,7 +18,7 @@ module Supply
                                      short_option: "-a",
                                      env_name: "SUPPLY_TRACK",
                                      description: "The track of the application to use. The default available tracks are: #{default_tracks.join(', ')}",
-                                     default_value: 'production'),
+                                     default_value: 'internal'),
         FastlaneCore::ConfigItem.new(key: :rollout,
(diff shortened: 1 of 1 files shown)


awgeorge:
I think the docs are automated, so going to check the box above. Please correct me if I'm wrong.

joshdholtz (changes requested):
@awgeorge Even thought this default value goes with my personal way of doing things, this is actually a breaking change that could possibly change a lot of _fastlane_ implementations that are expecting `production` but then would automatically change to using `internal`. That would required warnings and migration docs that we could provide but not all users see read the updates or see the updates.

So even though I totally get this I don't think we can accept this change at this time 😔 I appreciate the time though!

PR #14778, opened 2019-05-17
Spaceship::ConnectAPI models for BuildWatcher

joshdholtz:
### Motivation and Context
- Remove all API calls from `Spaceship::TestFlight` out of `BuildWatcher`
- This is needed before `pilot` can fully remove `Spaceship::TestFlight` usages

#### Usage with models
```rb
app = Spaceship::ConnectAPI::App.find("com.joshholtz.FastlaneTest")
puts "App: #{app.name}"

builds = app.builds # builds(filter: {}, includes: nil, limit: nil, sort: nil)
puts "Number of builds: #{builds.size}"

builds.each do |build|
  puts "\tVersion: #{build.version}, Uploaded: #{build.uploaded_date}"
end
```

#### Usage with client (which the models use internally)
```rb
# returns Spaceship::ConnectAPI::Response
# has `.next_page` and `all_pages` methods
client = Spaceship::ConnectAPI::Base.client
resp = client.get_builds(filter: {app: app_id}, includes: "", limit: 50)

# next_page will return nil if no more
next_resp = resp.next_page

# array of initial page and all other pages
all_resp = resp.all_pages

# Spaceship::ConnectAPI::Response also has a `to_models` method
# this map the response's data to the models
# the model that gets created is based on the type in the data
# here, models is an Array of Spaceship::ConnectAPI::Build
models = resps.map(&:to_models).flatten
```


spaceship/lib/spaceship/connect_api/response.rb
```diff
@@ -0,0 +1,50 @@
...
+        return links["next"]
+      end
+
+      def next_page
+        url = next_url
+        return nil if url.nil?
+        return client.get(url)
+      end
+
+      def all_pages
```
snatchev: This eager loading of the results feels a little heavy-handed. Do we have a page_size in the response? We could make this lazy by extending `Enumerable` and implementing `#each` which would yield a single data hash.

The one thing that makes me hesitant about this is cognitive overhead of using such an API. Would it be clear to the consumer of whats happening?

spaceship/lib/spaceship/connect_api/models/build.rb
```diff
@@ -35,6 +36,56 @@ class Build
...
+      # API
+      #
+
+      def self.all(app_id: nil, version: nil, build_number: nil, includes: nil)
+        resps = client.get_builds(
+          filter: { app: app_id, "preReleaseVersion.version" => version, version: build_number },
+          includes: includes,
+          limit: 30
+        ).all_pages
+        return resps.map(&:models).flatten
```
snatchev: Is the `flatten` called to turn the pages-of-lists to a 1D list?

snatchev (commented):
Over all, looks really good. Would just like to maybe refine the iteration over pages mechanism.

spaceship/lib/spaceship/connect_api/response.rb
joshdholtz: Ahh yup yup! I was planning on adding a "get X number of pages" or "X number of results" to eventually 🙃 All pages is def not great so I can add that

spaceship/lib/spaceship/connect_api/models/build.rb
joshdholtz: Yup yup, that is correct 👌

fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Issue #14309, opened 2019-02-25
[screengrab] Add an option to disable suffixes to screenshots


carstenhag:
### Feature Request
Add an option to disable suffixes to screengrab screenshots.

Has already been mentioned on https://github.com/fastlane/fastlane/issues/5988 but that was closed due to inactivity (I really hate activitybot!)

#### Motivation Behind Feature
Currently, a timestamp suffix is added to all screenshots, and this suffix is not configurable. I don't want there to be a suffix, because it increases git diffs.

I include all my app resources into one folder (that's the https://github.com/Triple-T/gradle-play-publisher/ concept) and yeah, having change diffs instead of remove/add ones would be nicer.

```
phone-screenshots/1-select-category_1551116523886.png
phone-screenshots/2-select-level_1551116524717.png
phone-screenshots/3-tutorial-level_1551116525847.png
phone-screenshots/4-tutorial-experienced_1551116528575.png
```

#### Feature Description
Add a Parameter to disable suffixes on the screenshots.

#### Alternatives or Workarounds
See https://github.com/fastlane/fastlane/issues/5988

carstenhag:
See https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/screengrab/screengrab-lib/src/main/java/tools.fastlane.screengrab/FileWritingScreenshotCallback.java#L60-L63 for the relevant code

janpio:
What is the parameter you suggest to be implemented to control this behavior?

carstenhag:
What do you think about `disable_timestamp_suffix`?

janpio:
Maybe invert it and do it via a parameter that is `true` by default and can be set `false` to not suffix?

carstenhag:
Or that, sure. Seems good to me!

carstenhag:
Yes, this issue still exists. It's a feature request.

carstenhag:
Yes, this issue still exists. It's a feature request. I really dislike @fastlane-bot.

carstenhag:
Still exists.

carstenhag:
Still open.

PR #14781, opened 2019-05-19
[screengrab] Add config to allow disabling timestamp suffix to filename 

Hiroto-N:
Implementing feature request in issue #14309

### Checklist

### Motivation and Context

See linked issue above - Adding a configuration so that user can opt out of appending timestamp to sceenshots produced by screengrab
### Description

Tested on example project in screengrab codebase by running bundle exec fastlane screengrab with appropriate parameters.

### DIFF
--- a/screengrab/lib/screengrab/options.rb
+++ b/screengrab/lib/screengrab/options.rb
@@ -132,6 +132,11 @@ module Screengrab
                                      env_name: 'SCREENGRAB_REINSTALL_APP',
                                      description: "Enabling this option will automatically uninstall the application before running it",
                                      default_value: false,
+                                     type: Boolean),
+        FastlaneCore::ConfigItem.new(key: :use_timestamp_suffix,
+                                     env_name: 'SCREENGRAB_USE_TIMESTAMP_SUFFIX',
+                                     description: "Add timestamp suffix to screenshot filename",
+                                     default_value: true,
                                      type: Boolean)
       ]
     end
--- a/screengrab/lib/screengrab/runner.rb
+++ b/screengrab/lib/screengrab/runner.rb
@@ -264,6 +264,7 @@ module Screengrab
       instrument_command = ["adb -s #{device_serial} shell am instrument --no-window-animation -w",
                             "-e testLocale #{locale.tr('-', '_')}",
                             "-e endingLocale #{@config[:ending_locale].tr('-', '_')}"]
+      instrument_command << "-e appendTimestamp #{@config[:use_timestamp_suffix]}"
       instrument_command << "-e class #{test_classes_to_use.join(',')}" if test_classes_to_use
       instrument_command << "-e package #{test_packages_to_use.join(',')}" if test_packages_to_use
       instrument_command << launch_arguments.map { |item| '-e ' + item }.join(' ') if launch_arguments
--- a/screengrab/lib/screengrab/runner.rb
+++ b/screengrab/lib/screengrab/runner.rb
@@ -264,7 +264,7 @@ module Screengrab
       instrument_command = ["adb -s #{device_serial} shell am instrument --no-window-animation -w",
                             "-e testLocale #{locale.tr('-', '_')}",
                             "-e endingLocale #{@config[:ending_locale].tr('-', '_')}"]
-      instrument_command << "-e appendTimestamp #{@config[:use_timestamp_suffix] == true}"
+      instrument_command << "-e appendTimestamp #{@config[:use_timestamp_suffix]}"
       instrument_command << "-e class #{test_classes_to_use.join(',')}" if test_classes_to_use
       instrument_command << "-e package #{test_packages_to_use.join(',')}" if test_packages_to_use
       instrument_command << launch_arguments.map { |item| '-e ' + item }.join(' ') if launch_arguments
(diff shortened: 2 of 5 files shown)


Hiroto-N:
I signed it!

screengrab/spec/runner_spec.rb
```diff
@@ -36,7 +36,7 @@ def mock_adb_response(mock_response)
       end
       it 'sets custom launch_arguments' do
         expect(mock_executor).to receive(:execute)
-          .with(hash_including(command: "adb -s device_serial shell am instrument --no-window-animation -w \\\n-e testLocale en_US \\\n-e endingLocale en_US \\\n-e username hjanuschka -e build_type x500 \\\n/"))
+          .with(hash_including(command: "adb -s device_serial shell am instrument --no-window-animation -w \\\n-e testLocale en_US \\\n-e endingLocale en_US \\\n-e appendTimestamp false \\\n-e username hjanuschka -e build_type x500 \\\n/"))
```
mollyIV: Hey @Hiroto-N 👋 
Thanks for making this PR 💪 

Just wondering why we are expecting `appendTimestamp false` in a command. According to the code:

https://github.com/fastlane/fastlane/blob/48ac12f577e9f4c82033adb8395442eda5dcfdad/screengrab/spec/runner_spec.rb#L31-L42

we do not set the value of `use_timestamp_suffix ` (default `true`) so we would expect getting `appendTimestamp true`, wouldn't we? 

Am I missing something? 🤔

screengrab/spec/runner_spec.rb
```diff
@@ -36,7 +36,7 @@ def mock_adb_response(mock_response)
       end
       it 'sets custom launch_arguments' do
```
mollyIV: Could we add more unit tests to test `use_timestamp_suffix`? 🙏 

* check if the default value is `true`
* set `use_timestamp_suffix` to `false`

🙇 🙇

mollyIV (changes requested):

screengrab/spec/runner_spec.rb
Hiroto-N: Ah right - When I was testing running the `fastlane screengrab` command it looked like it was picking up the correct default value even when it wasn't specified in the screengrabfile, but in the tests config[:use_timestamp_suffix] was returning nil (which resolves to false). I wasn't sure how exactly that part worked, but I'll dig a bit deeper and add tests around this flag as well.

Thanks for the feedback!

screengrab/lib/screengrab/options.rb
```diff
@@ -128,6 +128,11 @@ def self.available_options
                                      env_name: 'SCREENGRAB_REINSTALL_APP',
                                      description: "Enabling this option will automatically uninstall the application before running it",
                                      default_value: false,
+                                     is_string: false),
+        FastlaneCore::ConfigItem.new(key: :use_timestamp_suffix,
+                                     env_name: 'SCREENGRAB_USE_TIMESTAMP_SUFFIX',
+                                     description: "Add timestamp suffix to screenshot filename",
+                                     default_value: true,
```
joshdholtz: Can we add a`type: Boolean` in here? 😊

screengrab/lib/screengrab/runner.rb
```diff
@@ -255,6 +255,7 @@ def run_tests_for_locale(locale, device_serial, test_classes_to_use, test_packag
       instrument_command = ["adb -s #{device_serial} shell am instrument --no-window-animation -w",
                             "-e testLocale #{locale.tr('-', '_')}",
                             "-e endingLocale #{@config[:ending_locale].tr('-', '_')}"]
+      instrument_command << "-e appendTimestamp #{@config[:use_timestamp_suffix] == true}"
```
joshdholtz: This `== true` can be removed with the addition of the `type: Boolean` mentioned above

joshdholtz (changes requested):
I like this addition! Just a few small nitpicks to clean some things up 😊

screengrab/spec/runner_spec.rb
Hiroto-N: Hey, I've updated the PR - it looks like the config is usually configured through the Configurations in `Screengrab::Options.available_options`, so in a realistic scenario its never nil (which is what was making this false before).

I've set `config[:use_timestamp_suffix] = true` for this test and added tests below that populate the config through `FastlaneCore::Configuration` and passing values there.

Hope this is ok 😃

screengrab/spec/runner_spec.rb
Hiroto-N: added 👍

screengrab/lib/screengrab/runner.rb
Hiroto-N: fixed 👍

screengrab/lib/screengrab/options.rb
Hiroto-N: fixed 👍

mollyIV (approved):
Looks good ✨Thanks for the contribution @Hiroto-N  :bow:  🚀

felipe-silvestre-morais (approved):

Hiroto-N:
hey @joshdholtz sorry to dig up an old PR, is it possible to get this in? I believe I've addressed the changes suggested - thanks 👍

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This change is 🔥 Thanks for adding this! Appreciate the contribution ❤️

Issue #14713, opened 2019-05-03
iPad Pro 3rd Generation - Framing for app store


Alarson93:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

(Well technically I'm using 2.121.1 - but I suspect the minor version change hasn't addressed this based on the frame files I see on Github.)

### Issue Description

I cannot frame iPad Pro 3rd Gen screenshots. Framing other devices (XS Max, 8 Plus, etc) works just fine.

I assume that this is a result of Facebook not having created iPad Pro (3rd gen) device frames. If this is the case - our designer has created frames that you could perhaps use. 

Screenshots work fine - I can even upload unframed screenshots of the Pro 3rd Gen - this is strictly a framing issue. 

##### Command executed

```
  lane :take_screenshots do
    # Flavor 
    setup_for_external_demo

    # Snapshot
    devices = ["iPad Pro (12.9-inch) (3rd generation)"] # Only do what is required by the store ### ["iPhone 8 Plus", "iPhone Xs Max", "iPad Pro (12.9-inch)", "iPad Pro (12.9-inch) (3rd generation)"]
    capture_screenshots(scheme:"Foobar", stop_after_first_error: true, reinstall_app: true, number_of_retries: 0, skip_helper_version_check: true, devices: devices, output_directory: ENV["SNAPSHOT_DIRECTORY"])

    # Frame
    makeBackgroundTileCommand = "convert -size 1x1 xc:#{ENV['PRIMARY_COLOR']} #{ENV['WORKSPACE']}/#{ENV['SNAPSHOT_DIRECTORY']}/Background.jpg"
    puts("Calling: ")
    puts(makeBackgroundTileCommand)

    makeBackgroundTileCommandOutput = `#{makeBackgroundTileCommand}`
    puts(makeBackgroundTileCommandOutput)

    frameit(path: ENV["SNAPSHOT_DIRECTORY"])

    # Upload
    upload_to_app_store(force: true, overwrite_screenshots: true)
  end
```
##### Complete output when running fastlane, including the stack trace and command used

I'm only including a relevant sample.
<details>
  <pre>[16:34:34]: ---------------------
[...]

PR #14653, opened 2019-04-23
[deliver] fix ipad 3rd gen naming to prevent future collisions

joshdholtz:
### Description
- Finishes up some changes needed from https://github.com/fastlane/fastlane/pull/14576
- Removed `-3rd-gen` from the name
- More exact match on simulator name

PR #14787, opened 2019-05-21
[frameit] Add missing case for iPad Pro 12.9 (3rd Generation)

mikelrob:
### Checklist

### Motivation and Context
`fastlane` cannot frame screenshots with an asset named for the iPad Pro 3rd Generation.
Although some great ground work has already been completed in #14653 and  #14576, the `frameit` action was not able to use the disambiguation code to determine the device type. Instead I saw the error output:
`[12:42:35]: Unknown device type for size iOS-iPad-Pro-12.9 for path './en-US/iPad Pro (12.9-inch) (3rd generation)-2Deals.png'`

There is a related issue open #14713 

### Description

Adds the missing case for the newly added IOS_IPAD_PRO_12_9

After implementing the change, I ran the test suite.
After the test suit passed, I pointed our project to my local `fastlane` instance with the changes and ran our `fastlane` lane which generates the screen shots for our four selected devices and subsequently frames the screenshots and places them in our background. This completed the lane with the desired results including two iPad framed assets, one with the 2nd generation frame and the other with the 3rd generation frame which I have added to my `~/.fastlane/frameit/` 

Available from https://github.com/mikelrob/frameit-frames

### DIFF
--- a/frameit/lib/frameit/screenshot.rb
+++ b/frameit/lib/frameit/screenshot.rb
@@ -47,6 +47,8 @@ module Frameit
         return 'iPad Air 2'
       when sizes::IOS_IPAD_PRO
         return 'iPad Pro'
+      when sizes::IOS_IPAD_PRO_12_9
+        return 'iPad Pro 3rd Generation'
       when sizes::MAC
         return 'MacBook'
       else
--- a/frameit/lib/frameit/screenshot.rb
+++ b/frameit/lib/frameit/screenshot.rb
@@ -48,7 +48,7 @@ module Frameit
       when sizes::IOS_IPAD_PRO
         return 'iPad Pro'
       when sizes::IOS_IPAD_PRO_12_9
-        return 'iPad Pro 3rd Generation'
+        return 'iPad Pro (12.9-inch) (3rd generation)'
       when sizes::MAC
         return 'MacBook'
       else


janpio:
Thanks for your continued work on this.

frameit/lib/frameit/screenshot.rb
```diff
@@ -47,6 +47,8 @@ def device_name
         return 'iPad Air 2'
       when sizes::IOS_IPAD_PRO
         return 'iPad Pro'
+      when sizes::IOS_IPAD_PRO_12_9
+        return 'iPad Pro 3rd Generation'
```
janpio: Can we change this "filename pattern"?
```suggestion
        return 'iPad Pro (12.9-inch) (3rd generation)'
```
This is what we require the screenshots to contain, so not introducing another string here would be nice. (This is of course still messy and inconsistent, but that's Apple's fault 🤷‍♀ )

mikelrob: Thank you.
Yea, I'm cool with that.
I am using this string to locate the correct frame file. Those files do not currently have this same naming convention (with screen size and generation information).
I am happy to move toward this naming convention if it is appropriate.
What are your thoughts? @janpio

janpio: Yeah, this is trying to be similar to https://github.com/fastlane/fastlane/blob/0de811cec8bc50e266b2724d96a3a3852c845e6e/deliver/lib/deliver/app_screenshot.rb#L140-L141

mikelrob: Great 👍
I've tested the proposed change and it works, in tandem with the changes in the `frameit` repo of course.
I'll push both changes now.

janpio:
Your thoughts on this @liamnichols? 

Gives our users a way to manually put something into their  frameit frames folder before we figure out a way how to solve this over there.

MattMossIAG (approved):

getaaron (approved):

getaaron:
What about the 10.5" iPad Pro?

mikelrob:
> What about the 10.5" iPad Pro?

Hey @getaaron, the App Store connect will accept the iPad 12.9 2nd generation screen shot and down scale it to the iPad Pro 10.5 resolution. I decided to implement the support for the minimum required by Apple.

PR #14793, opened 2019-05-22
[download_dsyms] show platform name during download

tmm1:
### Checklist

### Motivation and Context

Easier debugging when looking at the logs. I have apps with the same bundle id, same version, same build number but one is ios and the other is appletvos.

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -53,7 +53,7 @@ module Fastlane
 
         # Write a nice message
         message = []
-        message << "Looking for dSYM files for #{platform} #{params[:app_identifier]}"
+        message << "Looking for dSYM files for `#{params[:app_identifier]}` (platform: #{platform})"
         message << "v#{version}" if version
         message << "(#{build_number})" if build_number
         UI.message(message.join(" "))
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -53,7 +53,7 @@ def self.run(params)
 
         # Write a nice message
         message = []
-        message << "Looking for dSYM files for #{params[:app_identifier]}"
+        message << "Looking for dSYM files for #{platform} #{params[:app_identifier]}"
```
janpio: To make clearer that the platform name is not part of the app identifier:
```suggestion
        message << "Looking for dSYM files for `#{params[:app_identifier]}` (platform: #{platform})"
```
Is `platform` always set or could it be empty?

tmm1: Platform is required afaik

janpio (approved):

janpio:
(@googlebot can be ignored here, it can't handle the "Suggest Changes" feature of GitHub unfortunately. We will handle this Pull Request as if it had the `cla: yes` label and merge as usual.)

joshdholtz (approved):
Looks good! Simple enough 😊

PR #14825, opened 2019-05-29
[supply] Allow track rollout value to be 1.0

phatmann:
### Checklist

### Motivation and Context

The default rollout value when updating a track via supply is 0.1 instead of 1.0. This was done because the value 1.0 was being rejected by the Google API. However, it is being rejected because the track's `status` property was not properly set to "completed". 

This is the correct fix for [this issue](https://github.com/fastlane/fastlane/issues/6501).

### Description

* Sets the track status to "completed" if rollout value is 1.0.
* Restores the default rollout value to 1.0 since it will not work as expected.

In order to test these changes, we need to promote a build from a non-production track to the "rollout" track with a rollout value of 1.0. So either we get someone to try this change who needs to roll out an app soon, or the next time I roll out my app to production (likely within next 1-2 weeks) I can test this change myself.

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -311,15 +311,12 @@ module Supply
       ensure_active_edit!
 
       track_version_codes = apk_version_code.kind_of?(Array) ? apk_version_code : [apk_version_code]
-
-      # This change happend on 2018-04-24
-      # rollout cannot be sent on any other track besides "rollout"
-      # https://github.com/fastlane/fastlane/issues/12372
-      rollout = nil unless track == "rollout"
+      staged_release = (track == "rollout" && rollout < 1.0)
 
       track_body = Androidpublisher::Track.new({
         track: track,
-        user_fraction: rollout,
+        user_fraction: staged_release ? rollout : nil,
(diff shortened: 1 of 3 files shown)


phatmann:
I signed it!

phatmann:
My change does not work because we are using V2 of the Android Publisher API, and we need to update to V3 in order to use the needed "status" property. I will take care of this unless this update has already been planned.

joshdholtz:
@phatmann Thanks for getting back to us the the results of this! We do actually need to upgrade from v2 to v3 since v2 is getting end of lifed in December of 2019 🙃 I can work on getting to that change this week probably so that we can make this work 💪

pavlospt:
@joshdholtz Is there an update or PR on Google Play Developer API v3? Any chance we could help on this?

joshdholtz:
Heyy 👋 I'm hoping to have this V3 migration out next week 🙃

prabhjyotsingh:
Any update on this `Google Play developer API` ?

phatmann:
Now that we are on v3 of API, can this PR be rebased and merged? Or is it no longer needed? @joshdholtz

PR #14841, opened 2019-06-02
[snapshot] Take screenshot of entire device screen instead of just the app window

GP89:
### Checklist

### Motivation and Context

Currently screenshots are taken of the app windows, which isn't necessarily the same size as the app screen which means they can't be used to upload to the app store

https://github.com/fastlane/fastlane/issues/13610

### Description

Take screenshot of the screen instead of the window

Run snapshot on all devices and verified the output screenshot resolutions match that expected in the appstore


GP89:
I signed it!

GP89:
I signed it!

janpio:
> Currently screenshots are taken of the app windows, which isn't necessarily the same size as the app screen

Do we know why and how this happens?

---

Note, I think there is another PR doing identical or similar changes. Before this is merged, these should both be looked at to make sure this all works together.

GP89:
I'm not sure why views don't fill the screen on some devices, sometimes there's just a few pixels at the bottom of the screen which are left blank giving the discrepancy.

Here on a iPhone 6s Plus the view stops just 2 pixels from the bottom of the screen, giving the discrepancy between `window.screenshot()` and `XCUIScreen.main.screenshot()`

iPhone-6s-Plus-Photos

Could possibly make this an optional thing, in case there's some reason you would want to take a screenshot of the window instead of the screen. But seeing as the purpose of these screenshots is for uploading to the app store, taking a screenshot of the screen instead of the window should always output the desired resolution

janpio:
> But seeing as the purpose of these screenshots is for uploading to the app store, taking a screenshot of the screen instead of the window should always output the desired resolution

Probably. 

I am a bit unsure because of how that code is not even that old and there was some previous complicated logic: https://github.com/fastlane/fastlane/commit/9031c665823250064008986d3ecfe58325644e1b But going back in history for that line, it seems to have been present for quite some time, and I don't really understand what was there before: https://github.com/fastlane/fastlane/commit/f410f2a6d9d20ae8e6747db978cae7142a03d781#diff-1a1173200436690d61ef63b6702ae6d6R131

> Here on a iPhone 6s Plus the view stops just 1 pixel from the bottom of the screen and 1 from the top, giving the discrepancy between `window.screenshot()` and `XCUIScreen.main.screenshot()`

Thanks for that illustration, the problem makes a lot more sense with that.

janpio:
Does device vs. app window maybe have to do something with statusbars?

Here is some code that also uses `app.windows.firstMatch` that is left in the file: https://github.com/fastlane/fastlane/blob/7e26ef5d16658a5eef364ff8be8e6e0578d8ca3f/snapshot/lib/assets/SnapshotHelper.swift#L277

GP89:
> Does device vs. app window maybe have to do something with statusbars?
> 

I don't believe so, The view will happily go under the status bar so not sure that's affecting this

> Here is some code that also uses `app.windows.firstMatch` that is left in the file:
> 
> https://github.com/fastlane/fastlane/blob/7e26ef5d16658a5eef364ff8be8e6e0578d8ca3f/snapshot/lib/assets/SnapshotHelper.swift#L277

That will probably have the same issue. Most discrepancies are with the height so window and screen width will be the same, but at least on iPad 3rd gen for my app the window width is less than the screen/device width.

janpio:
https://github.com/fastlane/fastlane/issues/13610#issuecomment-506463242 and following have some confirmations from users that this change improves the result.

joshdholtz (approved):
@GP89 Thanks for changing this! I added a commit to bump the version and change the `let` back to a `var` (it was causing issues because `simulator` gets reassigned a few lines below). But I tested this with a lane of mine that actually compares the expected sizes to the actual sizes and seems good 👌 I wasn't able to replicate the error that everyone was seeing but it still works for me 😊 Thanks!

GP89:
That let / var line wasn't to do with this pull request. it was from some other changes when I pulled in master, probably should have put that fix elsewhere but guess it doesn't matter now!

Thanks all!

PR #14892, opened 2019-06-13
[fastlane] Improve Lane Variables documentation by detecting `SharedValues` key assignment in functions

### LINKED ISSUE #14891: Improve Lane Variables documentation by detecting `SharedValues` key assignment in functions
### New Issue Checklist

### Issue Description

Recently we have been working on adding a [new `rubocop` rule](https://github.com/fastlane/fastlane/issues/14517) to prevent missing keys on the shared area provided by an action. @janpio noticed that there are `SharedValues` key assignment in functions, but they are **not** added to the `SharedValues` constants declaration. These keys are not detect by the `rubocop` rule so can be easily missed in the documentation process. 

The `rubocop` rule called `MissingKeysOnSharedArea` should be improved and check if there's `Actions.lane_context[SharedValues::KEY_NAME] = ...` assignment in a code but not declared in `SharedValues` module.

mollyIV:
### Checklist

### Motivation and Context

The rubocop rule called `MissingKeysOnSharedArea` should be improved and check if there's `Actions.lane_context[SharedValues::KEY_NAME] = ...` assignment in a code but not declared in `SharedValues` module.

https://github.com/fastlane/fastlane/blob/457c5d647c77f0e078dafa5129da616914e002c5/fastlane/lib/fastlane/actions/sync_code_signing.rb#L1-L38

This PR is an attempt to fix it 💪 

closes https://github.com/fastlane/fastlane/issues/14891

The whole thread: https://github.com/fastlane/fastlane/issues/14255. 

### Description

Modified `MissingKeysOnSharedArea` rule, added missing `SharedValues` keys and `output` values. 

🙇 🙇

### DIFF
--- a/fastlane/lib/fastlane/actions/build_ios_app.rb
+++ b/fastlane/lib/fastlane/actions/build_ios_app.rb
@@ -3,6 +3,7 @@ module Fastlane
     module SharedValues
       IPA_OUTPUT_PATH = :IPA_OUTPUT_PATH
       DSYM_OUTPUT_PATH = :DSYM_OUTPUT_PATH
+      XCODEBUILD_ARCHIVE ||= :XCODEBUILD_ARCHIVE
     end
 
     class BuildIosAppAction < Action
@@ -95,7 +96,8 @@ module Fastlane
       def self.output
         [
           ['IPA_OUTPUT_PATH', 'The path to the newly generated ipa file'],
-          ['DSYM_OUTPUT_PATH', 'The path to the dSYM files']
+          ['DSYM_OUTPUT_PATH', 'The path to the dSYM files'],
+          ['XCODEBUILD_ARCHIVE', 'The path to the xcodebuild archive']
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/build_ios_app.rb
+++ b/fastlane/lib/fastlane/actions/build_ios_app.rb
@@ -3,7 +3,7 @@ module Fastlane
     module SharedValues
       IPA_OUTPUT_PATH = :IPA_OUTPUT_PATH
       DSYM_OUTPUT_PATH = :DSYM_OUTPUT_PATH
-      XCODEBUILD_ARCHIVE ||= :XCODEBUILD_ARCHIVE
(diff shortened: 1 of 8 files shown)


janpio:
> The rubocop rule called `MissingKeysOnSharedArea` should be improved and check if there's `Actions.lane_context[SharedValues::KEY_NAME] = ...` assignment in a code but not declared in `SharedValues` module.

These are mainly "foreign" SharedValues being used here then, that usually "belong" to another action, correct? If so, it would probably be good to add a comment behind the docs definition to make this obvious.

mollyIV:
> These are mainly "foreign" SharedValues being used here then, that usually "belong" to another action, correct?

That is correct 👍 

> it would probably be good to add a comment behind the docs definition to make this obvious.

Any suggestions how the comment could look like @janpio ? 🤔

janpio:
Not really, just something that differentiates them from the "non foreign" ones. Maybe something along the lines of "# originally defined in ...Action"

mollyIV:
@janpio I implemented the suggested changes, could you please 👀 ? 🙏

joshdholtz (commented):
@mollyIV Quick question on this! Instead of redefining with `||=`, is it possible to add some sort of rubocop exception? It seems like it might be best to fix this with rubocup insetad of with `||=` 🤔

mollyIV:
Hey @joshdholtz 👋 

We assume that all `SharedValues` values must be defined in `module SharedValues` module and added to `self.output` action's method. I agree that redefining with `||=` is not the best solution, however I believe it is quite good compromise 🙂 

We added additional comment `# originally defined in ...Action` to highlight the existing const definition. 

The cases we are handling here are rather edge cases than a common scenario. 

WDYT @joshdholtz? 🤔

joshdholtz (approved):
@mollyIV I do agree that its a pretty good compromise 😊 I think the only thing that is still a _bit_ weird with me is that adding the `||=` is needed not because of functionality of code but because a rubocop rule 😇 But... I might just be nitpicky so this is probably good 👌

janpio (approved):
👍

mollyIV:
Great, thank you guys for the review 🙇

janpio:
First own PR merged? Yay! 🎆 🍰

PR #14904, opened 2019-06-14
[spaceship] fix Spaceship::Portal.provisioning_profile.all not reporting expired profiles

radex:
### Checklist

### Description

There is a recent regression in `Spaceship::Portal.provisioning_profile.all` (actually, seems to be a behavior change on Apple's API part), in that this API no longer reports expired profiles — which makes it difficult/impossible to actually update them not to be expired :) This change brings back previous behavior — without any adverse effects I can see.

### DIFF
--- a/spaceship/lib/spaceship/portal/portal_client.rb
+++ b/spaceship/lib/spaceship/portal/portal_client.rb
@@ -648,6 +648,7 @@ module Spaceship
         r.params = {
           teamId: team_id,
           includeInactiveProfiles: true,
+          includeExpiredProfiles: true,
           onlyCountLists: true
         }
       end
--- a/spaceship/spec/portal/portal_stubbing.rb
+++ b/spaceship/spec/portal/portal_stubbing.rb
@@ -40,7 +40,7 @@ class PortalStubbing
       stub_request(:post, "https://developer.apple.com/services-account/QH65B2/account/ios/profile/listProvisioningProfiles.action").
         to_return(status: 200, body: adp_read_fixture_file('listProvisioningProfiles.action.json'), headers: { 'Content-Type' => 'application/json' })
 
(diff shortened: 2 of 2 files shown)


janpio:
Can you please take care of the CLA stuff @radex?
You probably also have to update some of the stubbed or expected HTTP requests to this new parameter so the tests don't fail any more. Thanks.

radex:
@googlebot I signed it!

radex:
@janpio Sorry about this, fixed now!

janpio:
Awesome.

I think this closes a few issues we have that requested this functionality. Now the question is if this should be by default, as you implemented it, or if we should "hide" it behind an option somehow.

radex:
@janpio As far as I can tell, this fixes a regression, and this call always worked the way it works after the fix. I think Apple changed their API. My custom provisioning automation scripts (working directly with Spaceship) broke after WWDC 2019 (without me `bundle install`-ing) because of this, and after this change, they work like they used to :)

And so IMO it would be better to keep it the way it is, and if it becomes a problem for some people (e.g. because of performance), a flag could be added to _hide_ expired profiles.

this would also make sense because the API i` provisioning_profile.all`, and not `provisioning_profile.valid_profiles`

max-ott:
@janpio What's currently blocking this PR from getting merged? Need additional testing? From what @radex said it just brings back old functionality. If you give me the repro/test steps for `irb`, I'm more than happy to test.

janpio:
> What's currently blocking this PR from getting merged? Need additional testing?

As usual someone from the needs to be comfortable enough to merge this into the codebase and potentially live with breaking thousands of workflows out there.

I am not here, as I have no idea what the previous behavior really was. As I wrote, I think there were some issues, but we would have to hunt those down and confirm this PR really fixes them.

> If you give me the repro/test steps for `irb`, I'm more than happy to test.

I have not the slightest idea 🦐

max-ott:
> If you give me the repro/test steps for irb, I'm more than happy to test.

Part was  for @radex

joshdholtz (approved):
@radex This looks good and works good for me! Thanks for noticing this and making this change ❤️

PR #14911, opened 2019-06-16
[actions] reset simulator content tvOS & watchOS

dgyesbreghs:
### Checklist

### Motivation and Context
We need to be able to reset the simulator content for both tvOS and watchOS simulators. Currently only iOS simulators are supported.

### Description
I've renamed the current iOS parameter to os_versions, also modified the supported_platform method.
I've created a couple of example apps, installed them on a couple of simulators. Afterwards i cleared the content with the `reset_simulator_contents` action.

### DIFF
--- a/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
+++ b/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
@@ -2,12 +2,12 @@ module Fastlane
   module Actions
     class ResetSimulatorContentsAction < Action
       def self.run(params)
-        ios_versions = params[:ios]
+        os_versions = params[:version]
 
         if Helper.xcode_at_least?("9")
-          reset_xcode9_and_higher(ios_versions)
+          reset_xcode9_and_higher(os_versions)
         else
-          reset_up_to_xcode8(ios_versions)
+          reset_up_to_xcode8(os_versions)
         end
       end
 
--- a/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
+++ b/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
@@ -14,7 +14,7 @@ module Fastlane
           os_versions.each do |os_version|
             reset_all_by_version(os_version)
           end
-        else 
+        else
           reset_all
         end
 
--- a/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
+++ b/fastlane/lib/fastlane/actions/reset_simulator_contents.rb
@@ -46,9 +46,7 @@ module Fastlane
                                        description: "Which OS versions of Simulators you want to reset content and settings, this does not remove/recreate the simulators",
                                        is_string: false,
                                        optional: true,
-                                       type: Array,
-                                       default_value: ENV["FASTLANE_RESET_SIMULATOR_OS_VERSIONS"],
-                                       default_value_dynamic: true),
+                                       type: Array),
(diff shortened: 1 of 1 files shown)


janpio:
> I've renamed the current iOS parameter to os_versions

This would be a breaking change, which we _really_ try to avoid. Can you create back the `ios` option so this can still be used the old way and maybe add a deprecation warning when used? Then users can move over, and when we do a major release we can get rid of it.

janpio:
Example for a deprecated option: https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/fastlane/lib/fastlane/actions/update_project_code_signing.rb#L37-L42

dgyesbreghs:
@janpio I have made the necessary changes. So nothings going to break.

fastlane/lib/fastlane/actions/reset_simulator_contents.rb
```diff
@@ -42,9 +40,19 @@ def self.description
...
           FastlaneCore::ConfigItem.new(key: :ios,
+                                       deprecated: "Use `:os_versions` instead",
                                        short_option: "-i",
                                        env_name: "FASTLANE_RESET_SIMULATOR_VERSIONS",
-                                       description: "Which versions of Simulators you want to reset content and settings, this does not remove/recreate the simulators",
+                                       description: "Which OS versions of Simulators you want to reset content and settings, this does not remove/recreate the simulators",
+                                       is_string: false,
+                                       optional: true,
+                                       type: Array,
+                                       default_value: ENV["FASTLANE_RESET_SIMULATOR_OS_VERSIONS"],
```
janpio: Do we need this?

janpio:
I noticed you also got rid of the <=Xcode8 vs. <=Xcode9 thing and chose the code of the Xcode8 one as a template - can you explain that decision to me please?

fastlane/lib/fastlane/actions/reset_simulator_contents.rb
joshdholtz: I don't think we need this since this option is getting deprecated. If somebody has `FASTLANE_RESET_SIMULATOR_OS_VERSIONS` set, they are probably trying to use `os_versions` and not `ios`so I think we can 💣 this

joshdholtz: I don't think we need this since this option is getting deprecated. If somebody has FASTLANE_RESET_SIMULATOR_OS_VERSIONS set, they are probably trying to use `os_versions` and not `ios` I think we can 💣 this

fastlane/lib/fastlane/actions/reset_simulator_contents.rb
```diff
@@ -42,9 +40,19 @@ def self.description
...
+                                       default_value_dynamic: true),
```
joshdholtz: We can also 💣 this `default_value_dynamic`

joshdholtz (changes requested):
This looks really good! Thanks for making these changes. I'm going to add a commit to remove the defaulting to the different ENV so that we can get this merged in tonight and released 💪

joshdholtz:
@janpio Xcode 8 is no longer supported on Mojave so its kinnnnddddaaa end of lifed. I think its probably save to remove that support from this action 👌

joshdholtz (approved):
Thanks again for adding this! ❤️

PR #14944, opened 2019-06-24
[fastlane] use IO.popen instead of sh for release because input needed for 2FA

joshdholtz:
### Motivation and Context
Enabling 2FA on a rubygems account requires entering of otp when pushing

### Description
`sh` does not allow for inputting of executed commands. Needed to switch to using `IO.popen`

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -264,8 +264,14 @@ lane :release do
     )
 
     # Actual release of the gem
+    # Using IO.popen because `gem push` will prompt for 2FA if enabled
     #
-    sh("gem push ../pkg/fastlane-#{version}.gem")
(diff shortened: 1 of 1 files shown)


snatchev (commented):
How does this change read in from IO when there is a prompt?

joshdholtz:
@snatchev `sh` skips over the prompt completely and then things go 💥 `sh` is _really_ only for output 😬

snatchev:
@joshdholtz ah, okay so the change is intended not to error on input. It is not the scope of this PR to handle any input, is that right?

Wouldn't `gem push` fail anyway if 2FA is required and we cannot supply it?

snatchev (approved):

PR #14945, opened 2019-06-24
[fastlane/action] allow sh helper to take keyword arguments/hash to match sh action options

### LINKED ISSUE #14930: Using sh to run a command
### Question Checklist

### Question Subject

[sh](https://docs.fastlane.tools/actions/sh/)

#### Question Description

I am trying to implement a shell command in my lane. Running the command as such works fine:

```bash
sh('ls')
# prints directory
```

However I want to disable logs, as the [sh docs for parameters](https://docs.fastlane.tools/actions/sh/#parameters) it says you can do the following but it errors.

```bash
sh(command: "ls", log: false)
# [12:58:32]: unknown keyword: command
```

joshdholtz:
Fixes #14930

### Motivation and Context
`sh` "action" is usually called without any named arguments. However, if somebody was to look at the docs and compare `sh` to action actions, it is very confusing that it doesn't take named arguments. This change adds additional support for `sh(command: "cmd", log:true)` where before it was `sh("cmd", log:true)` (which doesn't feel like any other action)

### Description
`Fastlane::Actions.sh` now accepts only `(*args, &b)`. If `args` is only a single hash, the `:command` key is then pulled out of it and passed into `sh_no_action` like it would have otherwise.

This is backwards compatible with all other ways `sh` could be called

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -182,8 +182,29 @@ module Fastlane
     end
 
     # Execute shell command
-    def sh(*command, log: true, error_callback: nil, &b)
-      FastFile.sh(*command, log: log, error_callback: error_callback, &b)
+    # Accepts arguments with with and without the command named keyword so that sh
+    # behaves like other actions with named keywords
+    # https://github.com/fastlane/fastlane/issues/14930
+    #
+    # Example:
+    #  sh("ls")
+    #  sh("ls", log: false)
+    #  sh(command: "ls")
+    #  sh(command: "ls", log: false)
+    def sh(*args, &b)
+      # First accepts hash (or named keywords) like other actions
+      # Otherwise uses sh method that doesn't have an interface like an action
+      if args.count == 1 && args.first.kind_of?(Hash)
+        hash = args.first
+        command = hash.delete(:command)
+
+        raise ArgumentError, "sh requires :command keyword in argument" if command.nil?
+
+        new_args = [*command, hash]
+        FastFile.sh(*new_args, &b)
+      else
+        FastFile.sh(*args, &b)
+      end
     end
 
     def self.sh(*command, log: true, error_callback: nil, &b)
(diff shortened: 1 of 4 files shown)


mollyIV (commented):
Hey @joshdholtz 👋 , 

I was testing your changes locally, here are the results:

1. `Fastlane::Actions.sh`

```
lane :custom_lane do
    Fastlane::Actions.sh(command: "ls -1 | wc -l", log: true)
end
```
image

🎊 👍 

2. `sh` action

```
lane :custom_lane do
    sh(command: "ls -1 | wc -l", log: true)
end
```
image

⚠️  Is there anything that I am missing here or using a command in a wrong way? 🤔

joshdholtz:
@mollyIV That was a great catch! Thanks for testing this out ❤️  I actually moved the logic from `sh_helper.rb` to `fast_file.rb` (I forgot that `sh` technically gets called here first). So this change now doesn't touch `sh_helper` itself but instead gives the `Fastfile` the flexible way of interpreting the `sh` method/action

mollyIV (approved):
Hey Josh 👋 

Thank you for making changes 💪 

I just re-tested your PR:

```
// Fastfile

lane :custom_lane do
    sh(command: "ls -1 | wc -l", log: true)
end
```

```
$ bundle exec fastlane custom_lane                                                                                                         19:42  
[✔] 🚀 
[19:42:18]: ------------------------------
[19:42:18]: --- Step: default_platform ---
[19:42:18]: ------------------------------
[19:42:18]: Driving the lane 'ios custom_lane' 🚀
[19:42:18]: ---------------------------
[19:42:18]: --- Step: ls -1 | wc -l ---
[19:42:18]: ---------------------------
[19:42:18]: $ ls -1 | wc -l
[19:42:18]: ▸ 7

+------+------------------+-------------+
|           fastlane summary            |
+------+------------------+-------------+
| Step | Action           | Time (in s) |
+------+------------------+-------------+
| 1    | default_platform | 0           |
| 2    | ls -1 | wc -l    | 0           |
+------+------------------+-------------+
$ [19:42:18]: fastlane.tools finished successfully 🎉
```
It is working as expected  ✨ 🎊 

Thank you!

PR #14948, opened 2019-06-25
[screengrab] add `adb_host` option

CedricGatay:
Allow to specify the host to use when invoking `screengrab` :

 * either through CLI `adb_host`
 * environment variable: `SCREENGRAB_ADB_HOST`
 * configuration parameter: `:adb_host`

This allows to use a device farm to run screen grabs instead of having
an attached device to the build machine.

It is useful in the context of CI

### Checklist

### Motivation and Context

When using CI, it is cumbersome and slow to provision emulators (and we can't always as KVM is not always available). To solve this we can use a remote adb daemon and issue commands to it, but there is not any environment variables available to do this easily in adb.

This way I can run a CI build step on the proper network that can talk to remote adb to take screengrabs of app. 

### Description

Added an option to `screengrab` to allow specifying the adb host to use, if one is provided, adb command is updated to include the `-H $adb_host`elements.
Added a unit test to cover the case where `adb_host` is set and updated the method `run_adb_command` to remove the need for passing "adb" but focus on the command to be run: 

 * this way it is easier  to add the required `-H ...` flag
 * it makes no sense to repeat "adb" in the `run_adb_command` method call.

### DIFF
--- a/screengrab/lib/screengrab/options.rb
+++ b/screengrab/lib/screengrab/options.rb
@@ -137,7 +137,11 @@ module Screengrab
                                      env_name: 'SCREENGRAB_USE_TIMESTAMP_SUFFIX',
                                      description: "Add timestamp suffix to screenshot filename",
                                      default_value: true,
-                                     type: Boolean)
+                                     type: Boolean),
+        FastlaneCore::ConfigItem.new(key: :adb_host,
+                                     env_name: 'SCREENGRAB_ADB_HOST',
+                                     description: "Configure the host used by adb to connect, allows running on remote devices farm",
+                                     default_value: "localhost")
       ]
     end
   end
--- a/screengrab/lib/screengrab/options.rb
+++ b/screengrab/lib/screengrab/options.rb
@@ -141,7 +141,7 @@ module Screengrab
         FastlaneCore::ConfigItem.new(key: :adb_host,
                                      env_name: 'SCREENGRAB_ADB_HOST',
                                      description: "Configure the host used by adb to connect, allows running on remote devices farm",
-                                     default_value: "localhost")
+                                     optional: true)
       ]
     end
   end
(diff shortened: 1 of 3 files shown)


CedricGatay:
I signed it !


      Cedric GATAY
      06 69 50 20 69
     ______

      MAME
      Cité de la création et du numérique
      49 bd Preuilly - 37 000 Tours - Fr
      Code-Troopers.com

janpio:
Awesome feature @CedricGatay, thanks for implementing this. Are there any providers that offer such emulators for testing I could use to try this? Or how would I set one up locally?

CedricGatay:
Thank you @janpio !

As far as I know there is not any provider allowing this easily but you can test it rather quickly if you have two machines on the same network (or a VM in a host network mode).

On one machine (the device farm), attach an USB device or start an emulator. Then start `adb` telling it to listen on all network interfaces (that's the tricky part) with this command `adb -a nodaemon server start`.

On the other machine, the "build one", export the IP of the device farm under the environment variable, the CLI parameter or the build script, the way you prefer, it should work.
One way to diagnose if `adb` is properly listening on remote interfaces is to issue a `adb -H $SCREENGRAB_ADB_HOST devices` on the build machine, it has to work.

If I sum up, on device farm:
 * attach a device
 * start adb with `adb -a nodaemon server start`

On build machine:
 * `export SCREENGRAB_ADB_HOST=xxx.yyy.zzz.ttt`
 * `adb -H $SCREENGRAB_ADB_HOST devices`
 * `fastlane screengrab`

CedricGatay:
Hey @janpio can I do something to help with this PR ?

janpio:
Nope, someone from us will have to take another look and actually try it. We will try to get to that soon.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for adding this!! Really appreciate the contribution ❤️

Issue #14951, opened 2019-06-25
Wrong screen sizes on recent iphones


ajgr:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

This bug seems to affect only most recent iPhones starting with iPhone X all the way through iPhone XS Max. The images are fine, but all sizes are totally off which of course means that submission to the app store is a no-go. 

Fastlane vs simulator/required image sizes:
iPhone X:  1125 × 1996  - (1125 × 2436)
iPhone Xʀ:  828 × 1469 - (828 × 1792)
iPhone Xs Max:  1242 × 2204 - (1242 × 2688)

Additionally fastlane has difficulties with the simulator names. Instead of separating rows by devices, as is typically the case, it generates one gigantic row for anything that includes "iPhone X" (4 different devices). 

##### Complete output when running fastlane, including the stack trace and command used

<details>
fastlane screenshots
</details>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14.5                                     |
| Ruby                        | 2.5.1                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.20.1 (Apple Git-117)          |
| Installation Source         | /usr/local/bin/fastlane                     |
| Host                        | Mac OS X 10.14.5 (18F132)                   |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/2.5.1/lib            |
| OpenSSL Version             | OpenSSL 1.0.2o  27 Mar 2018                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 10.2.1                                      |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE |             |   |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
# This file contains the fastlane.tools configuration
[...]

PR #14963, opened 2019-06-28
[snapshot] Fix matching simulator and screenshot file names

claesjacobsson:
### Checklist

### Motivation and Context
Apple seems to have changed the casing of an Xcode simulator name between iOS 12.1 and 12.2: `iPhone XS Max` -> `iPhone Xs Max`. When the casing of the simulator name doesn't match the casing of the screenshot file name, the screenshots.html report doesn't group screenshots under the right device name.

Making the matching case insensitive seems to solve this problem.

Fixes parts of issue #14951

### Description
In `snapshot/lib/snapshot/reports_generator.rb`, I replaced `next unless file_name.include?(key_name)` with `next unless file_name.downcase.include?(key_name.downcase)`.
Adding the `downcase` makes the matching of screenshot file names and `available_devices` work even if Apple changes the casing of an Xcode simulator.

I tested this change by running a Snapfile including all iPhone and iPad devices.

I addition, I added a new 12.0 Simulator called `iPhone Xs Max`. Now, I had two simulators for the same device and iOS version but with different casing of the name. This resulted in an error: `The requested device could not be found because multiple devices matched the request`. Which makes sense. Removing the duplicate simulator made it work again.


janpio:
Could you maybe add a comment that explains why the `downcase` is necessary? This might not be obvious in a year or two and be removed and then causing problems. 

Are there maybe even tests for this where we could add this somehow?

claesjacobsson:
Not sure of there are any tests covering this.

joshdholtz (changes requested):
It looks like this PR only accounts for the differences between the `iPhone XS (Max)` and `iPhone Xs (Max)`. The `iPhone XR` also had changes to `iPhone Xʀ` so is this also something we need to take into consideration?

| Xcode 10 | Xcode 10.2 |
|---|---|
|Screen Shot 2019-07-03 at 9 02 24 AM|Screen Shot 2019-07-03 at 9 03 00 AM|

claesjacobsson:
I'll look into it.

claesjacobsson:
What version of Ruby should we support?

JulesMoorhouse:
I think the iPhone Xr uses a different character for the "r". Still an issue.

JulesMoorhouse:
@joshdholtz oh it looks like @claesjacobsson has fixed this now!!

getaaron:
I'm going to close this since it's no longer relevant for modern Xcode. Thank you for opening the PR, and feel free to open a new one and tag me if you disagree 👍

PR #15023, opened 2019-07-12
[match] import certs and p12 files into match repo

joshdholtz:
### Motivation and Context
It is almost impossible to use _match_ right now if you aren't able to create a new certificate. There are two main use cases where importing certs into _match_ is the ideal way:
1) Enterprise with distribution certs already created
    - You do not want to revoke certs just to have _match_ recreate certs since that will prevent all apps using that cert from working that are installed
2) When only given "read access" to your apple account on another team
   - Here you want put in the `.cer` and `.p12` that that team has given you so that the profiles that you can download (with your read access) will be valid when building

### Description
1. Pass in / prompt for the `.cer` and `.p12` path
2. Clone and decrypt the match repo
3. Find the cert ID of the cert passed in in step 1.
   - This is used for naming the `.cer` and `.p12` in the match repo
   - To find the cert ID, we query all certs and compare the `certificate_content` from the API response against the `.cer` content that will be imported
   - If matched, we will continue with import and name the `.cer` and `.p12` with the ID
4. Copy `.cer` and `.p12` into the correct directory in the repo with names of `<ID>.cer` and `<ID>.p12`
5. Encrypt the files
6. Commit the repo back up

#### Additional fixes
- _match_ was always forcing _sigh_ to true
  - This appears to be a bug when the _match_ repo is empty since _sigh_ should try to pull down provisioning profiles that already exist (this is extremely important for users that only have write access on developer portal)
- `Match::Runner` was doing some storage specific logic with `prefixed_working_directory` so this has been moved into the storage implementations (git and google cloud) to clean up the runner and make that core more reusable 

#### Remaining tasks
- [ ] Probably some doc

### Demo

fastlane_match_import_cert_and_p12

### How to test
Updated `Gemfile` and run `bundle install` or `bundle update fastlane`
```rb
source "https://rubygems.org"
gem 'fastlane', :git => 'https://github.com/fastlane/fastlane.git', :branch => 'joshdholtz-import-into-match'
```

#### Example command

```sh
fastlane match import \
    --username email@email.com \
    --git_url https://github.com/your_org/match_repo \
    --app_identifier com.your.app \
    --team_name "Your Team" \
    --type appstore
```


match/lib/match/importer.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      output_dir = File.join(storage.prefixed_working_directory, "certs", cert_type.to_s)
+
+      # Need to get the cert id by comparing base64 encoded cert content with certificate content from the API responses
+      Spaceship::Portal.login(params[:username])
+      Spaceship::Portal.select_team(team_id: params[:team_id], team_name: params[:team_name])
+      certs = Spaceship::ConnectAPI::Certificate.all(filter: { certificateType: certificate_type })
+
+      matching_cert = certs.find do |cert|
+        cert.certificate_content == cert_contents_base_64
```
snatchev: How do you feel about moving this to a class method on `Spaceship::ConnectAPI::Certificate` ?

match/lib/match/runner.rb
```diff
@@ -124,25 +116,8 @@ def run(params)
...
-          # method, as the value won't change. This way the warning is only printed once
-          UI.important("Looks like you run `match` in `readonly` mode, and didn't provide a `team_id`. This will still work, however it is recommended to provide a `team_id` in your Appfile or Matchfile")
-          @_folder_prefix = "*"
-        end
-        return File.join(working_directory, @_folder_prefix)
-      else
-        UI.crash!("No implementation for `prefixed_working_directory`")
-      end
+    def prefixed_working_directory
+      return self.storage.prefixed_working_directory
```
snatchev: nice refactor.

match/lib/match/runner.rb
```diff
@@ -221,18 +196,19 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
...
       end
 
       # Install the provisioning profiles
       profile = profiles.last
+      force = params[:force]
 
       if params[:force_for_new_devices] && !params[:readonly]
         if prov_type != :appstore
-          params[:force] = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_sym) unless params[:force]
+          force = device_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_sym) unless params[:force]
```
snatchev: Long line + a trailing negative conditional makes it hard to understand. Lets set `force` in an `if/else` block.

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -81,6 +99,32 @@ def initialize(type: nil,
...
+        # We fall back to "*", which means certificates and profiles
+        # from all teams that use this bucket would be installed. This is not ideal, but
+        # unless the user provides a `team_id`, we can't know which one to use
+        # This only happens if `readonly` is activated, and no `team_id` was provided
+        @_folder_prefix ||= currently_used_team_id
+        if @_folder_prefix.nil?
+          # We use a `@_folder_prefix` variable, to keep state between multiple calls of this
+          # method, as the value won't change. This way the warning is only printed once
+          UI.important("Looks like you run `match` in `readonly` mode, and didn't provide a `team_id`. This will still work, however it is recommended to provide a `team_id` in your Appfile or Matchfile")
+          @_folder_prefix = "*"
```
snatchev: So, if there is no team_id, we will prefix the paths with `*` character? What is the reason for this character? It feels over-loaded with bash directory globbing. Could we use something else that doesn't have incidental meaning?

match/spec/importer_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        type: "appstore",
+        git_url: git_url,
+        shallow_clone: true,
+        username: "flapple@something.com"
+      }
+
+      config = FastlaneCore::Configuration.create(Match::Options.available_options, values)
+      repo_dir = Dir.mktmpdir
+      keychain_path = FastlaneCore::Helper.keychain_path("login.keychain") # can be .keychain or .keychain-db
+      destination = File.expand_path("~/Library/MobileDevice/Provisioning Profiles/98264c6b-5151-4349-8d0f-66691e48ae35.mobileprovision")
```
snatchev: This doesn't seem to be used in the spec.

snatchev (commented):
Just a few comments.

match/lib/match/runner.rb
joshdholtz: Oh god, I didn't even see that railing negative conditional, TBH... refactor happening 💪

match/lib/match/storage/google_cloud_storage.rb
joshdholtz: I'm actually not sure 😕  I just moved this code over but it would probalby be good to understand this cause the `*` does seem a bit dangerous if that is what its used for

match/lib/match/importer.rb
```diff
@@ -0,0 +1,92 @@
...
+        certificate_type = Spaceship::ConnectAPI::Certificate::CertificateType::IOS_DEVELOPMENT
+      when :distribution, :enterprise
+        certificate_type = Spaceship::ConnectAPI::Certificate::CertificateType::IOS_DISTRIBUTION
+      else
+        UI.user_error!("Cert type '#{cert_type}' is not supported")
+      end
+
+      output_dir = File.join(storage.prefixed_working_directory, "certs", cert_type.to_s)
+
+      # Need to get the cert id by comparing base64 encoded cert content with certificate content from the API responses
```
janpio: Uff, this is not available anywhere else?

match/lib/match/importer.rb
```diff
@@ -0,0 +1,92 @@
...
+        cert.certificate_content == cert_contents_base_64
+      end
+
+      UI.error!("This certificate cannot be imported - the certificate contents did not match with any available on the Developer Portal") if matching_cert.nil?
+
+      # Make dir if doesn't exist
+      FileUtils.mkdir_p(output_dir)
+      dest_cert_path = File.join(output_dir, "#{matching_cert.id}.cer")
+      dest_p12_path = File.join(output_dir, "#{matching_cert.id}.p12")
+      files_to_commit = [dest_cert_path, dest_p12_path]
```
janpio: this should probably be done after they were actually copied there

janpio:
Nice feature!

A note (probably for future iterations of this):
Many cloud build services take the certificate and private key in a combined `p12` that is protected with a password (which you get when exporting the certificate (which contains the private key) from keychain access manually). One can pretty simply differentiate a private key in a `.p12` by the textual content of the file from a cert+key combination (which has to be "decrypted" (opened with password) first to get the actual certificate and key). I have all of this code in a custom thing of mine.

fastlane/lib/fastlane/actions/docs/sync_code_signing.md
```diff
@@ -407,6 +407,16 @@ fastlane match change_password
 
 You'll be asked for the new password on all your machines on the next run.
 
+### Import
+
+To import and encrypt a certificate (`.cer`) and the private key (`.p12`) into the _match_repo run:
```
janpio:
```suggestion
To import and encrypt a certificate (`.cer`) and the private key (`.p12`) into the _match_ repo run:
```

joshdholtz: I read through this like 10000 times and I didn't catch this 😝 I'm so bad at doc

snatchev (approved):

zcramos:
@snatchev @joshdholtz This PR (or possibly something else in 2.128.0, but this seems likely) broke our match config in our Bitrise CI pipeline. Match now fails with `No value found for 'username'`. Reverting to 2.127.2 resolves the issue. We set `username(ENV['APPLE_DEVELOPER_ID'].to_s)` in Matchfile, but that config doesn't appear to be used - something is overriding the value. 

I can open an issue later - don't have time to deal with the template right now but if needed I can make time for it later.

janpio:
I created https://github.com/fastlane/fastlane/issues/15072 to ring the correct bells, feel free to comment with more details later @zcramos - logs etc are most important of course.

joshdholtz:
@zcramos I think I found the issue 😱 Are you using `readonly: true` on your CI?

zcramos:
Yep @joshdholtz.

joshdholtz:
@zcramos I got a fix coming then. It should be out in a few hours 😊 Thanks for the report!

joshdholtz:
@zcramos `2.128.1` should be out with a fix for this for you now!

tedgonzalez:
Hi nice work on this! 
The description says "Imports certificates and profiles into the encrypted repository", can you show how we can use this to import provisioning profiles?

Thanks!

PR #15045, opened 2019-07-18
[snapshot] Fix warning about unused variable in SnapshotHelper.swift

MortenGregersen:
### Checklist
(The first two is intentionally kept unchecked as this change has nothing to do with Ruby)

### Motivation and Context
Xcode will throw a warning when running a target containing the current version of `SnapshotHelper.swift`. The `app` variable isn't used after the guard block.

### Description
This fix will remove the warning in Xcode when running/debugging a UI test for Snapshot.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -168,7 +168,7 @@ open class Snapshot: NSObject {
             app.typeKey(XCUIKeyboardKeySecondaryFn, modifierFlags: [])
         #else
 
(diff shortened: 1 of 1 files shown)


janpio:
Thanks!

Please also increase the version in the last line, so this can be taken care of via the merge as well. Thank you.

MortenGregersen:
> Please also increase the version in the last line, so this can be taken care of via the merge as well. Thank you.

Fixed 👍

joshdholtz (approved):
This looks good! Thanks for fix this 😊

PR #15058, opened 2019-07-21
[gradle] update gradle docs with example for task "bundle"

### LINKED ISSUE #15053: Upload android app bundle instead of APK to the playstore
I have this configuration to upload `.apk` to the Playstore which is working fine.

    lane :playstore do
      gradle(task: "clean")
      gradle(
        task: 'assemble',
        build_type: 'Release',
        properties: {
          "android.injected.signing.store.file" => "Keystore.jks",
          "android.injected.signing.store.password" => "abc",
          "android.injected.signing.key.alias" => "abc",
          "android.injected.signing.key.password" => "abc"
        }
      )
      upload_to_play_store(track: 'alpha')
    end

But, instead of `apk` file I want to upload `.aab` app bundle. How can I do it?
I checked in `app/build/outputs` directory and there is no `aab` directory generated.

FinnGu:
### Checklist

### Motivation and Context

Resolves #15053 by adding missing documentation to gradle action. It specifically adds an example for building an .aab file (Android Application Bundle) and clarifies that the existing example builds an .apk file.

### Description

No code changes were made, only comments and documentation was updated.
All tests were run successfully afterwards.

### DIFF
--- a/fastlane/lib/fastlane/actions/gradle.rb
+++ b/fastlane/lib/fastlane/actions/gradle.rb
@@ -102,7 +102,7 @@ module Fastlane
         [
           FastlaneCore::ConfigItem.new(key: :task,
                                        env_name: 'FL_GRADLE_TASK',
-                                       description: 'The gradle task you want to execute, e.g. `assemble` or `test`. For tasks such as `assembleMyFlavorRelease` you should use gradle(task: \'assemble\', flavor: \'Myflavor\', build_type: \'Release\')',
+                                       description: 'The gradle task you want to execute, e.g. `assemble`, `bundle` or `test`. For tasks such as `assembleMyFlavorRelease` you should use gradle(task: \'assemble\', flavor: \'Myflavor\', build_type: \'Release\')',
                                        optional: false,
                                        is_string: true),
           FastlaneCore::ConfigItem.new(key: :flavor,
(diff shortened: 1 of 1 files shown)


janpio:
Via https://github.com/fastlane/fastlane/issues/15053#issuecomment-513557059

> However I am not so sure about moving the texts outside of the code sections. The docs look better that way, but this kind of works against having `self.example_code` as an array that automatically wraps each example as formatted ruby code. Also I could not find any other action where there was plain text above the code examples. And lastly, using this method it is not possible to have plain text before the first example.
> 
> I think we should stick to having text inside code examples as comments or otherwise the existing structure of `self.example_code` should be reworked so that it is the same for all actions.

Valid points - I just assumed our logic there was smarter as there were a few of those. But I think we can keep this how you did it here, and open an issue to solve this in a better way somehow.

joshdholtz (commented):
I think that only thing that scares me about this approach is that this assume the `example_code` is markdown which it is not what it technically is that the moment 🤔 It is used in the docs which is markdown but could get used someone else (like in the CLI) which would make things look a bit weird.

This probably _does_ look better than that comments but it does have some assumptions associated with it. Maybe instead we go back to the comments and add a follow up issue/PR for improving how example code works?

janpio:
As there were some of these before, I don't think we should put the work on @FinnGu here but merge this and fix later.

joshdholtz (approved):
Thanks for fixing these examples for us! ❤️

FinnGu:
Happy to be able to make this fantastic project a little better :)

Issue #15084, opened 2019-07-24
fastlane "download_all" doesn't include mac profiles :(  


nancy-gomez:
### New Issue Checklist

- [ ] Updated fastlane to the latest version 
No, but basing this off master commit source code
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
We have a scheduled job to download all the provisioning profiles from the Apple Developer Portal and we are just now noticing none of the Mac profiles have been downloaded. We are running as follows:

##### Command executed
`fastlane sigh download_all --readonly --skip_install --output_path /tmp/profiles/`

##### Complete output when running fastlane, including the stack trace and command used

The output of the command is not necessary as it is just a huge list of our profiles and this would mean nothing to you, but looking through the output I am searching for a profile that I generated for macOS and it is nowhere to be found. After looking at the source code: https://github.com/fastlane/fastlane/blob/master/sigh/lib/sigh/download_all.rb I can see why! I have been playing around with Spaceship and I know that in order to download mac specific stuff you need to pass `mac: true` as such:

`Spaceship.provisioning_profile.all(mac: true)` but this is for all of the mac profiles, this doesn't include the ios ones. So the source code would have to be updated to combine the two collections of profiles

### Environment

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14.5                                     |
| Ruby                        | 2.2.3                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.20.1 (Apple Git-117)          |
| Installation Source         | ~/.rbenv/versions/2.2.3/bin/fastlane        |
| Host                        | Mac OS X 10.14.5 (18F132)                   |
| Ruby Lib Dir                | ~/.rbenv/versions/2.2.3/lib                 |
| OpenSSL Version             | OpenSSL 1.0.2p  14 Aug 2018                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 10.2.1                                      |

### System Locale

[...]

PR #15091, opened 2019-07-24
[sigh] Make download_all include mac profiles

nancy-gomez:
### Checklist

### Motivation and Context

When I saw the "download_all" option I assumed this meant all profiles -- including the macos ones. After seeing that some of our machines were missing some profiles, I dug deeper and saw the only profiles missing were out mac ones. As a sanity check I looked at the documentation, assumed I could probably just pass the "platform" flag but multiple attempts showed me this was not the case. Looking at the source code it became apparent that this functionality was missing! And just calling sigh with the platform flag is forcing me to specify a bundle identifier but I want all of the mac profiles!

https://github.com/fastlane/fastlane/issues/15084

### Description

All I did was combine the collection of ios profiles (with or without the xcode managed ones) as well as the mac profiles.

First, I checked via spaceship if what I did is syntactically and logically correct.
Then, I executed rspec.
Finally, I used my fastlane clone to test and see if the behavior is as expected. At the very bottom of the output I saw all of the mac profiles that weren't being downloaded before!

I made sure I was running my local version by running
`bundle show fastlane`
Then ran the same exact command as before
`bundle exec fastlane sigh download_all --readonly --skip_install --output_path /tmp/profiles/`

### DIFF
--- a/sigh/lib/sigh/download_all.rb
+++ b/sigh/lib/sigh/download_all.rb
@@ -12,7 +12,9 @@ module Sigh
       Spaceship.select_team
       UI.message("Successfully logged in")
 
-      Spaceship.provisioning_profile.all(xcode: download_xcode_profiles).each do |profile|
+      # all ios profiles (including or excluding Xcode Managed Profiles) and mac profiles
+      profiles = Spaceship.provisioning_profile.all(xcode: download_xcode_profiles) + Spaceship.provisioning_profile.all(mac: true)
+      profiles.each do |profile|
         if profile.valid?
           UI.message("Downloading profile '#{profile.name}'...")
           download_profile(profile)


janpio:
I am not 100% sure that changing the behavior of the existing option is the right way to go here. As obvious as it might be to make the connection "all = including Mac", this has been this way for quite some time and people might rely on the current behavior. 

Might it make sense to "hide" this behavior change behind an additional option?

nancy-gomez:
So I just noticed also in the Spaceship source code that if we truly wanted all profiles it would be something as such:

```
  ios_profiles = Spaceship.provisioning_profile.all(xcode: download_xcode_profiles)
  mac_profiles = Spaceship.provisioning_profile.all(mac: true)
  tvos_profiles = Spaceship.provisioning_profile.all_tvos
```

This is quite unfortunate and I understand that download_all has been working this way for some time. It just seems a bit strange for download_all to not include all.

Honestly I think it would make the most sense for download_all's default behavior to download everything, and then we can exclude xcode profiles, mac profiles, tvos profiles, watchOs profiles. 

I'm open to adjusting this with an option but would like to know what makes the most sense to name it. I also realized I have to fix the appended extension depending on platform as well. 

I'm all 👂s :)

nancy-gomez:
What if we support the platform flag actually? 🤔  This is how I first thought it would work and I would just have to do separate executions depending on the platform, which is annoying but I mean better than nothing. Or what if we pass all the platforms we want?

Such as 
```
fastlane sigh download_all --platform ios,macos,tvos,watchos
```

Or something 🤷‍♀ Idk how the exact syntax should look

janpio:
Do we have flags as additional input mechanism for the CLI commands somewhere else?

nancy-gomez:
Well @janpio there's this example: `fastlane sigh download_all --download_xcode_profiles`
So we can perhaps make a flag like `--download_mac_profiles` ? Would that be reasonable?

janpio:
I didn't call your suggestion unreasonable ;) But yes, maybe that would follow existing samples more. But I don't want to make this suggestion and make you change your PR based on this feeling - there should be some more investigation (which I unfortunately can't provide right now).

nancy-gomez:
Hm more investigation :o What do you mean? Who should we loop in? I definitely would love some agreement regarding this feature request, I would hate to make a bigger PR and then get a "no" :'(

If a flag seems best, I can definitely get around to changing this implementation!

joshdholtz:
@nancy-gomez Hey 👋 I was out on vacation so sorry about not seeing this sooner! I think making use of the platforms flag would _probably_ make the most sense to me (since that is what the platform flag is intended for 😇). However, I think the platforms flag only allow one platform at a time so you would still need to run the command twice. Thoughts?

nancy-gomez:
Yeah I guessed that would be the case. Alrighty then so if I can manage to implement the platform flag we can get this feature added in? :)

nancy-gomez:
Closed this one to open this one: https://github.com/fastlane/fastlane/pull/15175

PR #15095, opened 2019-07-25
Allow Crashlytics devtools.jar to be auto updated

jeremydurham:
### Checklist

### Motivation and Context
Allow the Crashlytics devtool jar to be auto updated

### Description
When running a fastlane command that uses the crashlytics android devtools jar, we will perform a HEAD request to verify that the artifact on disk is up to date with the artifact on S3. Use the ETag from S3 and verify it against the MD5 of the file on disk. If the file on disk does not exist or the hash doesn't match, download the artifact from S3 and unzip it to replace the existing files.


snatchev (approved):
LGTM! Thanks Jeremy!

PR #15105, opened 2019-07-29
Update Fastlane.swift

cdoky:
[swift] fixed changelogFromGitCommits between

### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/issues/15104#issue-473914348

### Description

```swift
changelog = changelogFromGitCommits(
	between: ["\(commit_id)", "HEAD"]
)
```

### DIFF
--- a/fastlane/swift/Fastlane.swift
+++ b/fastlane/swift/Fastlane.swift
@@ -913,7 +913,7 @@ func cert(development: Bool = false,
                                                                                       RubyCommand.Argument(name: "platform", value: platform)])
   _ = runner.executeCommand(command)
 }
-@discardableResult func changelogFromGitCommits(between: String? = nil,
+@discardableResult func changelogFromGitCommits(between: [String]? = nil,
                                                 commitsCount: Int? = nil,
                                                 path: String = "./",
                                                 pretty: String = "%B",


janpio (changes requested):
As far as I know this file is generated from the ruby code - so we would have to change the ruby side so that the generator knows how to generate this properly.

philCc:
ok

joshdholtz (changes requested):
@philCc The Swift interface is generated from options defined in the action (so [here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb#L78-L92)) so the properly place for this change would be to change the `type` of the option for how the Swift interface is generated.

This _does_ have an `is_string` type of `false` so (https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb#L82) so IMO the Swift interface should actually be `between: Any? = nil` 🤔 I'm not sure why its being generated as a `String?` in this case.

The reason this needs to be `Any?` is for backwards comparability since you can see [here](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/changelog_from_git_commits.rb#L84-L91) there is logic for taking a string or an array

I'm going to look into why this is generating as a `String?` and not `Any?`

philCc:
May be?
Between configuration used commit_count: line:83
Commit_count used between line:98

joshdholtz:
Closing this in favor of #15165 (which fixes `is_string: false`) for a bunch of actions. 

@philCc Allowing support for plugins is 100% in the plans in the near future 😊

PR #15130, opened 2019-08-02
[spaceship/cert] add Apple Development and Apple Distribution certificate type

niw:
### Checklist

### Motivation and Context

Xcode 11 is creating new type of development certificate labeled with "Apple Development".
This is only type of certificate that Xcode 11 will create and it will no longer create "iOS Development" or such per platform certificate.

### Description

This patch adds known "Apple Development" certificate ID in the Spaceship certificate ID list, so that we can fetch such certificate from the developer portal.

### DIFF
--- a/spaceship/lib/spaceship/portal/certificate.rb
+++ b/spaceship/lib/spaceship/portal/certificate.rb
@@ -89,6 +89,9 @@ module Spaceship
       # Certs are not associated with apps
       #####################################################
 
+      # An Apple development code signing certificate used for development environment
+      class AppleDevelopment < Certificate; end
+
       # A development code signing certificate used for development environment
       class Development < Certificate; end
 
@@ -152,6 +155,10 @@ module Spaceship
       # A Mac push notification certificate for production environment
       class MacProductionPush < PushCertificate; end
 
+      APPLE_CERTIFICATE_TYPE_IDS = {
+        "83Q87W3TGH" => AppleDevelopment
+      }
+
       IOS_CERTIFICATE_TYPE_IDS = {
         "5QPB9NHCEI" => Development,
         "R58UK2EWSO" => Production,
@@ -187,7 +194,9 @@ module Spaceship
         "DIVN2GW3XT" => DeveloperIdApplication
       }
 
-      CERTIFICATE_TYPE_IDS = IOS_CERTIFICATE_TYPE_IDS.merge(MAC_CERTIFICATE_TYPE_IDS)
+      CERTIFICATE_TYPE_IDS = APPLE_CERTIFICATE_TYPE_IDS
+                             .merge(IOS_CERTIFICATE_TYPE_IDS)
+                             .merge(MAC_CERTIFICATE_TYPE_IDS)
 
       # Class methods
       class << self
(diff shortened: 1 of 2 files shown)


niw:
@googlebot I signed it!

spaceship/lib/spaceship/portal/certificate.rb
```diff
@@ -258,7 +267,7 @@ def factory(attrs)
         #  only include certificates matching the current type.
         def all(mac: false)
           if self == Certificate # are we the base-class?
-            type_ids = mac ? MAC_CERTIFICATE_TYPE_IDS : IOS_CERTIFICATE_TYPE_IDS
+            type_ids = mac ? MAC_CERTIFICATE_TYPE_IDS : APPLE_CERTIFICATE_TYPE_IDS.merge(IOS_CERTIFICATE_TYPE_IDS)
```
niw: Probably merge `APPLE_CERTIFICATE_TYPE_IDS` for both?

janpio:
Thanks for adding this, I had shortly looked into this before and thought this wasn't actually supported by the new API - seems I just didn't look in the right places for the right things ❤️ 

Ping @max-ott for review.

max-ott:
Hi @niw looks good to me. Could you provide steps to use this via the spaceship `playground` so we can check if it's working?

niw:
Here is the reproducible steps to confirm the behavior.

**Prerequisite**
1. Install Xcode 11 beta, and launch it.
1. Open Preferences window, then open Accounts tab.
1. Add new Apple ID that is used for Apple developer account, then login.
1. Select one of teams that the Apple developer account is associated, then click Manage Certificates.
1. Add new Apple Development to create a new Apple Development certificate then click Done.

**Playground usage**
1. Launch spaceship playground.
    ```
    $ bundle exec fastlane spaceship
    ```
1. Type Apple ID for `Username:` that is used to create the Apple Development certificate previously, and type password to login.
1. Execute next command to confirm that the created Apple Development certificate are listed in the result.
    ```
    [1] pry(#<Spaceship::Playground>)> Certificate::AppleDevelopment.all 
    ```
1. Execute next command to confirm that the one is also listed in the result.
    ```
    [1] pry(#<Spaceship::Playground>)> Certificate.all 
    ```

janpio:
(Prerequisite should also be possible via the online interface from Apple probably)

niw:
@janpio That’s true, let me revise comment.

spaceship/lib/spaceship/portal/certificate.rb
joshdholtz: I just added a commit that fixed this ^

joshdholtz (approved):
@niw Thanks for making this change! I just added a commit onto here that also adds `AppleDistribution`. Would you be able to review this when you get a chance? 😊

spaceship/lib/spaceship/portal/certificate.rb
niw: Thank you, I was wondering what is expected behavior in case `mac` is `true` here.
Probably this is expected, but prefer to defer to the decision to someone who knows expected behavior here.

niw:
@joshdholtz Probably you already noticed though, you need to follow what @googlebot said, or this patch will not be mergeable.

joshdholtz:
@niw Yup yup, I can override that on the merge 😊💪

niw:
Ooh! I see, I was misunderstanding it. Thank you so much for adding this!

PR #15131, opened 2019-08-03
[core] Less hacks in command_executor_spec

dotdoom:
### Checklist

### Motivation and Context

Improving spec to be less of a change trigger, and having less hacks.

### Description

- get rid of `allow_message_expectations_on_nil`
- set mocks on the right call (not implementation detail of calling `String#strip`, but rather yielding from `each`)

### DIFF
--- a/fastlane_core/spec/command_executor_spec.rb
+++ b/fastlane_core/spec/command_executor_spec.rb
@@ -12,23 +12,21 @@ describe FastlaneCore do
       end
 
       it 'handles reading which throws a EIO exception', requires_pty: true do
-        explodes_on_strip = 'danger! lasers!'
-        fake_std_in = ['a_filename', explodes_on_strip]
-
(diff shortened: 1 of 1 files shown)


janpio:
I am scared of all this command executor code.

How confident are you of these changes?

dotdoom:
These are just specs, right?

I think it's less scary to have a realistic test rather than `expect($?).to_receive` which is essentially `expect(nil).to_receive`, plus set mocks on the correct method (which according to the comment was intended, but for some reason considered not easier).

I agree the other PR I have could look scary, so I'll add more tests to it; but the spec update here I am confident about.

janpio:
Yep, specs that I had to touch some time ago and never really understood - so I am still scared no matter how unreasonable that is ;)

janpio (approved):

joshdholtz (approved):

PR #15139, opened 2019-08-05
[action] fix for Xcode 11 when using `$(MARKETING_VERSION)` in `get_version_number`

florentmorin:
### Checklist

### Motivation and Context

When using Xcode 11, `CFBundleShortVersionString` use `$(MARKETING_VERSION)` instead of version number like `xxx.yyy.zzz` in `Info.plist`. So, `get_version_number` returns `$(MARKETING_VERSION)`.

### Description

`$(MARKETING_VERSION)` comes from Build settings of Xcode project.

So, if version number from `Info.plist` looks like a build setting, the good value is extracted from build settings and returned by `get_version_number`.

### DIFF
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -16,7 +16,11 @@ module Fastlane
         project = get_project!(folder)
         target = get_target!(project, target_name)
         plist_file = get_plist!(folder, target, configuration)
-        version_number = get_version_number!(plist_file)
+        version_number = get_version_number_from_plist!(plist_file)
+        
+        if version_number =~ /\$\(([\w\-]+)\)/
+          version_number = get_version_number_from_build_settings!(target, $1, configuration)
+        end
 
         # Store the number in the shared hash
         Actions.lane_context[SharedValues::VERSION_NUMBER] = version_number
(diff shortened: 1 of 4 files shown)


florentmorin:
@googlebot I signed it!

janpio:
Thanks for this! 

I just created #15142 which adds CI for Xcode 11 - in the best case this will fail some tests because of exactly this reason, and your PR will make it pass again. In the worst case we have loads of unrelated failures as well, but we'll have to deal with that then 🙈

florentmorin:
@janpio OK. I'll try to do that for the next week. 😄

florentmorin:
@janpio : Test integrated 😄

janpio:
Awesome. If I copy those over to https://github.com/fastlane/fastlane/pull/15142 it should fail, until we also add the other code from this PR?

florentmorin:
@janpio This code is also compatible with Xcode 10. So you can merge in current release.
But you need to report this code for Xcode 11, because Xcode 11 automatically use `$(MARKETING_VERSION)` value.

joshdholtz (approved):
This is great! I added one commit change for a small nitpick but otherwise this is 🔥 Thanks for the contribution!

grioja:
This broke variable interpolation in version numbers

`Foo${BAR}` no longer works as a result

@florentmorin

janpio:
Can you please open an issue @grioja and provide more information and best a reproduction? Thanks.

PR #15142, opened 2019-08-05
[CI] Run tests with Xcode 11 (and ruby 2.5) on Circle CI

janpio:
### Checklist

### Motivation and Context
Xcode 11 changes some bits around, so it makes sense to run CI with it as well.

### Description
This duplicates our macOS build job with Xcode 11 and Ruby 2.5 (as 2.3 is not available on that image any more). One test had to be adapted to the new paths of Xcode 11 as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,7 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
+            echo "2.5" > .ruby-version
             gem update --system
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,8 @@ jobs:
           name: Post Test Results to GitHub
           command: bundle exec danger || echo "danger failed"
           when: always # Run this even when tests fail
-"Execute tests on macOS (Xcode 11.0.0)":
+
+  "Execute tests on macOS (Xcode 11.0.0)":
     macos:
       xcode: "11.0.0"
     environment:
@@ -117,6 +118,7 @@ jobs:
           name: Post Test Results to GitHub
           command: bundle exec danger || echo "danger failed"
           when: always # Run this even when tests fail
+
   "Execute tests on Ubuntu":
     environment:
       CIRCLE_TEST_REPORTS: ~/test-reports
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
-            echo "2.3" > .ruby-version
+            echo "2.3.7" > .ruby-version
             gem update --system
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,6 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
-            echo "2.3.7" > .ruby-version
             gem update --system
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2
 
 jobs:
-  "Execute tests on macOS (Xcode 9.0.1)":
+  "Execute tests on macOS (Xcode 9.0.1, Ruby 2.3)":
     macos:
       xcode: "9.0.1"
     environment:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -59,6 +59,66 @@ jobs:
...
+    shell: /bin/bash --login -eo pipefail
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: v1-macos-gems-{{ checksum "Gemfile" }}-{{ checksum "fastlane.gemspec" }}
+      - restore_cache:
+          keys:
+            - v3-homebrew-{{ epoch }}
+            - v3-homebrew
```
mollyIV: Are the cached artifacts the same for builds with Xcode 11.0.0 and Xcode 9.0.1? 🤔 It looks like we are using the same key for the cache.

.circleci/config.yml
```diff
@@ -59,6 +59,66 @@ jobs:
...
+      - store_test_results:
+          path: ~/test-reports
+      - store_artifacts:
+          path: ~/test-reports/rspec
+          destination: test-reports
+
+      - run:
+          name: Post Test Results to GitHub
+          command: bundle exec danger || echo "danger failed"
+          when: always # Run this even when tests fail
```
mollyIV: The only difference between `Execute tests on macOS (Xcode 9.0.1, Ruby 2.3)` and `Execute tests on macOS (Xcode 11.0.0, Ruby 2.5)` jobs is the `macos` parameter. It may be a bit hard to maintain these jobs, because we will need to make the changes in both (copy/paste most probably). Is it a way on circle CI to make some parts of the job reusable? Maybe [Reusable Commands](https://circleci.com/docs/2.0/reusing-config/#invoking-reusable-commands) could help here? 🤔

mollyIV (commented):
Overall looks good, I suggested a potential improvement to avoid a job duplication. It requires some research, so I am okay to do that in a separate ticket 😉

.circleci/config.yml
janpio: Doe these matter regarding homebrew? I kinda assumed the Xcode version doesn't.

.circleci/config.yml
janpio: The Ruby version is also different - but that's it. If there is a way to simplify the job definition, go for it. There probably is some way to extract jobs to files maybe?

mollyIV: > If there is a way to simplify the job definition, go for it.

Maybe we could leverage the [YAML's anchors](https://docs.gitlab.com/ee/ci/yaml/README.html#anchors) feature:

> YAML has a handy feature called ‘anchors’, which lets you easily duplicate content across your document. Anchors can be used to duplicate/inherit properties, ....

janpio: If Circle supports this, indeed this could work. I did so for Apache Cordova taht uses Travis: https://github.com/apache/cordova-paramedic/blob/master/.travis.yml The syntax is a bit ... unintuitive though. Best check the Circle CI docs again if they have something about structuring your jobs etc.

joshdholtz:
I wish Circle made it easier to handle things like this 😔 So what did we decide on this? 🙃 I'm not familiar with anchors at all so my approach would have been to just copy/paste. Do anchors provide any other value? Or is copying/pasting this once just as "clean"?

janpio:
There are two problem I guess:
1. My copy/paste I did re-uses the same cache keys. Not sure that is a problem across Xcode (and other tooling that is also different) versions
2. My copy/paste duplicates some stuff, that maybe could be avoided if there was a way to to that on Circle. Anchors could be one, Circle itself could also provide something - @mollyIV wanted to look into that a bit, correct?

joshdholtz:
That cache keys _shouldn't_ be a problem across multiple Xcodes and Ruby versions. It would probably only really make a difference if we were using different system images/versions of some sort but we should be good to reuse those cache keys, IMO 👌

mollyIV:
> My copy/paste duplicates some stuff, that maybe could be avoided if there was a way to to that on Circle. Anchors could be one, Circle itself could also provide something - @mollyIV wanted to look into that a bit, correct?

Actually that was just a suggestion and I though that you @janpio would do the investigation on that as the scope of this PR 😅 

I did a quick research and found this blog post from CircleCI: https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/

They support YAML's anchors and aliases, although I am totally okay with copy/paste guys 👍 Most probably we will not be changing this piece of code a lot and I do hope we will leverage GitHub Actions in near future anyway 😉 😉

janpio:
Different Xcode means different OS version and tooling across the board:
9.3.1 https://circle-macos-docs.s3.amazonaws.com/image-manifest/build-954/index.html
vs.
11.0.0 https://circle-macos-docs.s3.amazonaws.com/image-manifest/build-419/index.html

janpio:
I'm fine with keeping the copy/paste as well - just CI config.

joshdholtz (approved):
LFTM 💪❤️

PR #15152, opened 2019-08-07
[Fastlane Swift] Fix suppress_xcode_output parameter type

jhandguy:
### Checklist

### Motivation and Context

Generated `Fastlane.swift` includes the wrong type for the argument `suppressXcodeOutput`, which is `String?` instead of `Bool?`.

### Description

Giving implicit `Boolean` type to `suppress_xcode_output` in Ruby solves the issue.

### DIFF
--- a/gym/lib/gym/options.rb
+++ b/gym/lib/gym/options.rb
@@ -210,7 +210,7 @@ module Gym
                                      env_name: "SUPPRESS_OUTPUT",
                                      description: "Suppress the output of xcodebuild to stdout. Output is still saved in buildlog_path",
                                      optional: true,
-                                     is_string: false),
+                                     type: Boolean),
         FastlaneCore::ConfigItem.new(key: :disable_xcpretty,
                                      env_name: "DISABLE_XCPRETTY",
                                      description: "Disable xcpretty formatting of build output",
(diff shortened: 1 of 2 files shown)


janpio:
Ping @mollyIV ;) 🙊

mollyIV (approved):
Hey @jhandguy 👋 

Looks good, thank you for fixing this 🙌  

I am wondering how many more places like that we have in our codebase? 🤔Maybe we could create a new issue and do the investigation? 

CC: @janpio

janpio:
Probably some, going through all of these is kind of busy work though - so we depended on the issues created by users like @jhandguy until now. But if you want to do a sweep of all action options anyway for some reason, go for it!

jhandguy:
> Hey @jhandguy 👋
> 
> Looks good, thank you for fixing this 🙌
> 
> I am wondering how many more places like that we have in our codebase? 🤔Maybe we could create a new issue and do the investigation?
> 
> CC: @janpio

Sounds like a good idea 👍 
Let's create an issue for it and make another PR for this, maybe?
I might take a look later! 👌

joshdholtz:
@jhandguy @janpio There is already a WIP PR for this that I started a while back (but I bit off more than I could chew) - https://github.com/fastlane/fastlane/pull/13816

It ended up changing too many things to make it a safe merge. I talked with @mollyIV about this and I think we are going to revive this but do it in smaller portions 😊

joshdholtz (approved):
Thanks for fixing this! ❤️

PR #15165, opened 2019-08-12
[Fastlane.swift] options of `is_string: false` should be Any and not String

joshdholtz:
### Motivation and Context
Inspired by #15105

### Description
Parameters in `Fastlane.swift` functions were being set to `String` (or `String?`) if there was not `type` associated with it. This is acceptable except for when `is_string` is set to to `false` on the option. If `is_string` is set to `false` and not `type` is given or can be inferred, then the parameter type should be set as `Any` (or `Any?`).

This will change the generated Swift API a significant amount. In the near future, we should be removing `is_string` completely and start using `type` everywhere as this provides better type checking and documentation.

### Example of a change
This diff shows that more values types will be allowed for this option
Screen Shot 2019-08-12 at 7 05 51 AM
Screen Shot 2019-08-12 at 7 05 29 AM

### DIFF
--- a/fastlane/lib/fastlane/swift_fastlane_function.rb
+++ b/fastlane/lib/fastlane/swift_fastlane_function.rb
@@ -7,15 +7,17 @@ module Fastlane
     attr_accessor :param_default_values
     attr_accessor :param_optionality_values
     attr_accessor :param_type_overrides
+    attr_accessor :param_is_strings
     attr_accessor :reserved_words
     attr_accessor :default_values_to_ignore
 
-    def initialize(action_name: nil, keys: nil, key_descriptions: nil, key_default_values: nil, key_optionality_values: nil, key_type_overrides: nil, return_type: nil)
+    def initialize(action_name: nil, keys: nil, key_descriptions: nil, key_default_values: nil, key_optionality_values: nil, key_type_overrides: nil, key_is_strings: nil, return_type: nil)
       @function_name = action_name
       @param_names = keys
       @param_descriptions = key_descriptions
       @param_default_values = key_default_values
       @param_optionality_values = key_optionality_values
+      @param_is_strings = key_is_strings
       @return_type = return_type
       @param_type_overrides = key_type_overrides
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -91,11 +93,13 @@ def override_default_value_if_not_correct_type(param_name: nil, param_type: nil,
...
     end
 
-    def get_type(param: nil, default_value: nil, optional: nil, param_type_override: nil)
+    def get_type(param: nil, default_value: nil, optional: nil, param_type_override: nil, is_string: true)
       unless param_type_override.nil?
         type = determine_type_from_override(type_override: param_type_override)
       end
-      type ||= "String"
+
+      # defaulting type to Any if is_string is false so users area llowed to input all allowed types
```
mollyIV:
```suggestion
      # defaulting type to Any if is_string is false so users are allowed to input all allowed types
```

fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -7,15 +7,17 @@ class SwiftFunction
     attr_accessor :param_default_values
     attr_accessor :param_optionality_values
     attr_accessor :param_type_overrides
+    attr_accessor :param_is_strings
     attr_accessor :reserved_words
     attr_accessor :default_values_to_ignore
 
-    def initialize(action_name: nil, keys: nil, key_descriptions: nil, key_default_values: nil, key_optionality_values: nil, key_type_overrides: nil, return_type: nil)
+    def initialize(action_name: nil, keys: nil, key_descriptions: nil, key_default_values: nil, key_optionality_values: nil, key_type_overrides: nil, key_is_strings: nil, return_type: nil)
```
mollyIV: There's plenty of logic behind this action. Wish we had some unit tests 😅 Unfortunately dunno if that's possible, not very familiar with the way `fastlane` generates `Swift` files.

mollyIV (approved):
Looks good @joshdholtz 💪 

> In the near future, we should be removing is_string completely and start using type everywhere as this provides better type checking and documentation.

I think we kind of started this discussion there: https://github.com/fastlane/fastlane/pull/15152#pullrequestreview-271872898

I can imagine having a simple regex + grep or `rubocop` rule to detect setting `is_string = false` and doing fixes in small chunks would do. If that sounds about right, will be happy to create a separate issue for the investigation 😉 

Thoughts? 🤔

joshdholtz:
@mollyIV I totes have a PR started for that rubocop rule here 😊 - https://github.com/fastlane/fastlane/pull/13816

I definitely bit off more than I could chew by trying to do all of it at once but I would be down for reworking that to do a few tools / actions at a time 😇

PR #15210, opened 2019-08-20
[CI] Github Action: Workflow that runs `bundle exec rake`

janpio:
Hello Github Actions!


rogerluan:
By the looks of this PR, not sure if it's something we'd like to pursue and if we would, it's something we can easily open a new PR with more recently-updated APIs (e.g. checkout v4, newer GHA syntax, etc 😇)

I'm closing this PR then :)

PR #15223, opened 2019-08-25
[match] add option to  skip provisioning profiles

addbrick:
### Checklist

### Motivation and Context

We are managing the provisioning profiles separate from our certificates. This options allows for a smoother, instead of having to catch errors and ignore them if they are about provisioning profiles.

This is a personal issue. In the future we plan on moving the profiles in match as well, but thought this would be useful to others.

### Description

Adds the option `skip_provisioning_profile: Bool (default: false)` to `match`

I have run this in test scripts where I am migrating all our Certificates and Private Keys to Match, and updating our build to use Match to manage Certs and Sigh to manage Profiles.

I also added a new spec.

### Testing Steps

Setup a match branch where there are only certs and no profiles. Run it in read only mode and you will get an error saying profiles have to be generated. Run it again in read only and pass in `skip_provisioning_profile: true` and it will run successfully, no error.


joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This looks good! Thanks for adding those tests 🙏 I added one (two) commits for changing the name to `skip_provisioning_profiles` and to fix the env name to be prefixed with `MATCH_`. Other than that this was 💯 Thanks for the contribution!

PR #15227, opened 2019-08-26
[fastlane_core] added support for Github Actions CI

RishabhTayal:
Description

Currently, `is_ci` action doesn't support Github Actions CI. This PR adds support for the same.


joshdholtz (approved):
I like it! Thanks for adding this ❤️

PR #15228, opened 2019-08-27
[spaceship] Change provisioning profile update request to match what Apple Developer passes

niw:
### Checklist

### Motivation and Context

Issue: #14283 
Original pull request: #14284

When trying to update a provisioning profile using Spaceship, We’ve been getting the unexpected response such as `There was an error generating the provisioning profile ...`.

@brentdur took a look at Spaceship’s request vs the request that the Apple Developer Portal sends, there was a difference in the `template` field in the request body. Spaceship was using `template.purpose_display_name`, which results in the request data resembling template `Company (Development)`. Apple Developer Portal users the equivalent of `template.purpose_name`, which results in the request data being template `company-development`.

### Description

Based on original #14284, apply same change to use `template.purpose_name`, however keep public method `template_name` for unnecessary backward compatibility break.

### DIFF
--- a/spaceship/lib/spaceship/portal/provisioning_profile.rb
+++ b/spaceship/lib/spaceship/portal/provisioning_profile.rb
@@ -457,7 +457,7 @@ module Spaceship
             devices.map(&:id),
             mac: mac?,
             sub_platform: tvos? ? 'tvOS' : nil,
-            template_name: template_name
+            template_name: is_template_profile ? template.purpose_name : nil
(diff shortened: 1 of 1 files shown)


giginet (approved):
Looks good to me.

By the way, there are no tests about Provisioning Profiles. 🤕

giginet:
@niw I merged!!!

niw:
@giginet Thank you!

PR #15235, opened 2019-08-28
[download_dsyms] Remove leading zeros from version string

werminghoff:
### Checklist

### Motivation and Context
The `download_dsyms` action fails when an app's version has a leading zero in one of its parts, such as a version "4.07". AppStore reports this version as "4.7", so the action fails to find its dSYM files, since "4.07" != "4.7".

### Description
When it finds the version, split the string using the dot symbol, then convert each part to integer and join everything back using the same dot symbol.

"4.07" -> ["4", "07"] -> [4, 7] -> "4.7"

### Testing Steps
1. Increment a app's version on its Info.plist, adding a leading zero to a part of its version string.
2. Build, archive and upload it to AppStore connect
3. Run `fastlane run download_dsyms` or a lane that has this action set up
4. The dSYM zip file should be downloaded successfully

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -56,6 +56,11 @@ module Fastlane
           version = live_version.version
           build_number = live_version.build_version
         end
+        
+        # Remove leading zeros from version string (eg. 1.02 -> 1.2)
+        if version
+          version = version.split(".").map{ |p| p.to_i }.join(".")
+        end
 
         # Make sure output_directory has a slash on the end
         if output_directory && !output_directory.end_with?('/')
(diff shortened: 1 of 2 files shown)


janpio:
Thanks, interesting problem. Is there also an issue for this that will be solved by this PR?

We should probably add a test to https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/download_dsyms_spec.rb which confirms this behavior and makes sure we don't accidentally remove it in the future.

werminghoff:
Hi @janpio, sorry for the late response. I don't think there's an issue for that, at least I didn't find one. I'll see how to implement the tests (as I have very little experience with Ruby) and will update the PR as soon as I can.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This for adding in the fix for this! I rebased this and added some tests to make sure this doesn't get removed 😇 This will get released tomorrow! Appreciate the contribution ❤️

PR #15284, opened 2019-09-07
[screengrab] Let users set clean status bar

rickclephas:
### Checklist

### Motivation and Context
The goal is to automatically enable and disable the clean status bar so that you don't have to do this manually with for example the QuickDemo app.

### Description
I have added the options `clean_status_bar` and `clean_status_bar_config` to `screengrab`.
With `clean_status_bar` the clean status bar can be enabled. 
Use the `clean_status_bar_config` option to customise the appearance of the clean status bar.

The configuration options are based on the commands described here:
https://android.googlesource.com/platform/frameworks/base/+/master/packages/SystemUI/docs/demo_mode.md

### ToDos
- We probably need some documentation for the possible configuration keys and their values
- It seems that the configuration for the mobile network icons are reset when screengrab changes the local to anything other then the device local. Maybe we should move this functionality to the `screengrab-lib` so that it can set the clean status bar after every local change?

@janpio any thoughts?

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md
@@ -171,7 +171,39 @@ If you're having trouble getting your device unlocked and the screen activated t
 
 ## Clean Status Bar
 
-You can use [QuickDemo](https://github.com/PSPDFKit-labs/QuickDemo) to clean up the status bar for your screenshots.
+Screengrab can clean your status bar to make your screenshots even more beautiful.  
+Note: the clean status bar feature is only supported on devices with *API level >= 23*.
+
+To use the clean status bar feature add the following lines to your src/debug/AndroidManifest.xml
+```xml
+<!-- Indicates the use of the clean status bar feature -->
+<uses-feature android:name="tools.fastlane.screengrab.cleanstatusbar"/>
+<!-- Allows for changing the status bar -->
+<uses-permission android:name="android.permission.DUMP"/>
+```
+
+After that you can enable and disable the clean status bar at any moment during your tests.  
+In most cases you probably want to do this in the @BeforeClass and @AfterClass methods.
+```java
+@BeforeClass
+public static void beforeAll() {
+    CleanStatusBar.enableWithDefaults();
+}
+
+@AfterClass
+public static void afterAll() {
+    CleanStatusBar.disable();
+}
+```
+
+Have a look at the methods of the `CleanStatusBar` class to customize the status bar even more.  
+You could for example show the Bluetooth icon and the LTE text.
+```java
+new CleanStatusBar()
+    .setBluetoothState(BluetoothState.DISCONNECTED)
+    .setMobileNetworkDataType(MobileDataType.LTE)
+    .enable();
+```
 
 # Advanced _screengrab_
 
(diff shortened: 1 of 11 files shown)


janpio:
Thanks for tackling this @rickclephas!

If you don't want to put the options description into the `description` of the option, we should probably put it into the full documentation that can be edited via https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md Maybe a headline _before_ https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md#tips?

> * It seems that the configuration for the mobile network icons are reset when screengrab changes the local to anything other then the device local. Maybe we should move this functionality to the `screengrab-lib` so that it can set the clean status bar after every local change?

Sounds valid, and I am sure you know more about this right now - so I trust your judgement to decide how to do this.

rickclephas:
> we should probably put it into the full documentation

Yeah that is probably best since there are 25 configuration options.

> Sounds valid, and I am sure you know more about this right now - so I trust your judgement to decide how to do this.

Alright I will see if moving it to the `screengrab-lib` works better.
If it does we could allow the customisation of the status bar in the test classes.
Would that be preferable over configuring it in the `Screengrabfile`?

janpio:
I honestly can't answer that as I don't know enough about that functionality. 

If I understood everything correctly: In general it feels like configuring the screenshots should maybe not be in the tests, but in the screenshot taking configuration.

rickclephas:
So I have now moved all the configuration to the `screengrab-lib`.

To use the feature user will need to add the following lines to their `debug/AndroidManifest.xml`
```xml
<uses-feature android:name="tools.fastlane.screengrab.cleanstatusbar"/>
<uses-permission android:name="android.permission.DUMP" tools:ignore="ProtectedPermissions" />
```

The `android.permission.DUMP` permission is required to allow the app to configure the status bar.
Screengrab will need to grant this permission before running the test.
With the `tools.fastlane.screengrab.cleanstatusbar` feature we can detect if the users wants to use the clean status bar option and grant the permission.
**NOTE:** We could also detect the permission but for the clean status bar to work we also set the `sysui_demo_allowed` global setting and I am not sure if the DUMP permission is used by any other tools.

Then in the test class you can enable and disable the clean status bar at any moment.
You probably want to put this in the `@BeforeClass` and `@AfterClass` methods.

```java
@BeforeClass
public static void beforeAll() {
    Screengrab.setDefaultScreenshotStrategy(new UiAutomatorScreenshotStrategy());

    CleanStatusBar.enableWithDefaults();
}

@AfterClass
public static void afterAll() {
    CleanStatusBar.disable();
}
```

You can also use a custom configuration.
For example to show the LTE text and Bluetooth icon:
```java
new CleanStatusBar()
    .setMobileNetworkDataType(MobileDataType.LTE)
    .setBluetoothState(BluetoothState.DISCONNECTED)
    .enable();
```

With the configuration in the test class it would also be possible to have a different status bar for different screenshots.
That might be useful to change the status bar from `transparent` to `semi-transparent` based on different activities so that it matches the actual behaviour of the app.

If we move the configuration to the lane/actions options we have to pass all the configurations to the test `am instrument` command. With 25 configuration options that seems a little ugly in the logs.

rickclephas:
Documentation has been updated.
Let me know if there is anything else that needs to be done.

joshdholtz (approved):
This is sooooo 💯 Thanks so much for all the time you spent on this 😊 I rebased from master and fixed it to add support for AndroidX. Really appreciate the contribution ❤️

PR #15285, opened 2019-09-08
[action] Fix `get_version_number` curly braces

florentmorin:
### Checklist

### Motivation and Context

Fix issue #15194 when `get_version_number` should read build setting variable.

### Description

On issue #15194, when `get_version_number` read build setting variable with curly braces, it didn't work.
Now, it does.

### Testing Steps

Test code was included in `get_version_number_action_spec.rb` and is called `gets the correct version number for 'TargetVariableCurlyBraces'`.
So, It seems to be OK.


fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -21,6 +21,8 @@ def self.run(params)
         # Get from build settings if needed (ex: $(MARKETING_VERSION) is default in Xcode 11)
         if version_number =~ /\$\(([\w\-]+)\)/
           version_number = get_version_number_from_build_settings!(target, $1, configuration)
+        elsif version_number =~ /\$\{([\w\-]+)\}/
```
janpio: Maybe we can combine these two regexes?
Something like `/\$(\(|\{)([\w\-]+)(\)|\})/` could work maybe.

florentmorin: Yes. But no. This regex [allows](https://rubular.com/r/0xhimXyOFsgu4j) `${MARKETING_VERSION)`

The good regex is `/\$\(([\w\-]+)\)|\{([\w\-]+)\}/` but :

- it's harder to understand _(and to maintain)_
- it's requires double check to extract value from matching groups (`$1` and `$2`), so more code is finally required.

This is why I prefer an easier code than an optimized code.

janpio: Ha, good point - you are totally right.
Maybe add a comment then for the new line a la "${MARKETING_VERSION} also works" or something like that.

florentmorin: OK. Comment just append.

joshdholtz (approved):
Looks good! Thanks for the contribution 😊❤️

PR #15289, opened 2019-09-10
[latest_testflight_build_number] Support :platform parameter for non-live builds

krin-san:
### Checklist

### Motivation and Context
Support `platform` filter for [latest_testflight_build_number](https://docs.fastlane.tools/actions/latest_testflight_build_number/) action to get only tvOS or only iOS app build number when both are available and might be deployed separately from each other

### Description
- Mapped `platform` to ITC platform when it's not `nil`
- Set it as `preReleaseVersion.platform` filter in [`app/{id}/builds`](https://developer.apple.com/documentation/appstoreconnectapi/list_all_builds_of_an_app) request

⚠️This actually changes the currentent behavior of `latest_testflight_build_number`: before it was taking the latest TestFlight build for ANY platform and now _(when `platform` is not specifically set)_ it will return only build for iOS platform. That might be fixed by removing default value from `platform` field but I guess it might lead to other issues I'm not aware about yet.

### Testing Steps
Any tests which can be written are either integration tests or `available_options` tests.

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -36,6 +36,7 @@ module Fastlane
           return build_nr
         else
           version_number = params[:version]
+          platform = params[:platform]
 
           # Create filter for get_builds with optional version number
           filter = { app: app.apple_id }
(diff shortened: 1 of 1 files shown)


janpio:
Hm, how would one actually get the `any platform` case if the option has a default `ios`?

krin-san:
@janpio I think, there's no way to get it. `platform` parameter is optional and putting `any platform` is a good idea, but when nothing or `platform: nil` is passed, `default_value: "ios"` gets picked.

joshdholtz:
Yeah, I don't think its really possible to get to the `any platform` at the moment but it seems fine to keep in here for now IMO 🙃

joshdholtz (approved):
Thanks for adding this ❤️  Not having this filter on platform may have been an oversight when replacing with the `Spaceship::Connect` logic 🙃

PR #15303, opened 2019-09-13
[snapshot] Added Snapshot support for dark mode in iOS 13

webventil:
### Checklist

### Motivation and Context
Dark mode is introduced with iOS 13. To take screenshots in dark mode, the simulator has to be configured for this. To automatically do this I added an option to `Snapshot` to create a plist file before launching the UI tests.

### Description
I created a new environment variable `dark_mode` that can be used to configure a simulator to be in dark mode.

### Testing Steps
To use that option just enter `dark_mode true` in a Snapfile. This will launch the iOS simulator for all tests in dark mode.


letatas:
If we want both dark and light mode, can we do something like `darkmode([true,false])`?
Will it run all the configurations in both modes?

snapshot/lib/snapshot/simulator_launchers/simulator_launcher_base.rb
```diff
@@ -124,6 +127,18 @@ def localize_simulator(device_type, language, locale)
       end
     end
 
+    def inteface_style(device_type, dark_mode)
```
letatas:
```suggestion
    def interface_style(device_type, dark_mode)
```

snapshot/lib/snapshot/simulator_launchers/simulator_launcher_base.rb
```diff
@@ -55,11 +55,14 @@ def prepare_simulators_for_launch(device_types, language: nil, locale: nil)
...
 
       device_types.each do |type|
-        if launcher_config.erase_simulator || launcher_config.localize_simulator
+        if launcher_config.erase_simulator || launcher_config.localize_simulator || launcher_config.dark_mode
           erase_simulator(type)
           if launcher_config.localize_simulator
             localize_simulator(type, language, locale)
           end
+          if launcher_config.dark_mode
+            inteface_style(type, launcher_config.dark_mode)
```
letatas:
```suggestion
            interface_style(type, launcher_config.dark_mode)
```

webventil:
> If we want both dark and light mode, can we do something like `darkmode([true,false])`?
> Will it run all the configurations in both modes?

I could introduce an enum to set interface style via an array to `dark` and `light`. But I'm not sure how to do this sufficient with the multi simulator setup.

webventil:
Maybe a better solution would be to link the interface style to the simulator. But this would actually mean, that snapshot creates a clone of the simulator for each test run, to be able to run two configurations in parallel.

```
devices([
  "iPhone Xs" : "dark",
  "iPhone Xs" : "light",
  "iPhone Xs Max" : "light",
])
```

letatas:
That could be a good idea indeed.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for adding this feature! I added one commit to make `dark_mode` an optional boolean. There was an issue where I couldn't go back to light mode because if was looking at `dark_mode` as an if statement and putting `dark_mode: false` would never allow the plist to get a 1 written to it. I made this an optional and we are now checking for `nil?`

So it will only write that plist if that option is explicitly set to either `true` or `false`

roelvanderkraan:
Hi I've just tried this feature, I was surprised to see when I set `dark_mode` to `true` the simulators are also removed? Is that expected behavior? When I remove the settings the simulators are no longer removed. 
It's not really convenient behavior for me, I just want to rerun the screenshot test with the existing app in dark mode to prevent having to go through the whole process of setting up the test data.

PR #15309, opened 2019-09-13
[download_dsyms] Added 'live' version parameter to only download live_version dSYMs

### LINKED ISSUE #15308: `download_dsyms` action is missing a fetching 'live' version option
We have setup a Jenkins job that executes every 3 days to download the dSYMs from App Store Connect and uploads it to New Relic.

While doing that, we noticed that `download_dsyms` only has the option of `version: 'lastest'` or passing specific build version.

But since we are disabling BITCODE_ENABLED for our tests builds (faster waiting for processing time), we would only need to upload dSYM files for our latest released App Store live version.

As our apps are always in heavy development, 'lastest' option doesn't return our desired version.

Example app versions in App Store Connect:

* **3.9.1** (live_version, released App Store version)
* **4.0** (edit_version, waiting for review)
* **4.1** (test build uploaded to TestFlight only)

`'latest'` would return a build from the **4.1** train.

#### Feature Description

Add another `version: 'live'` option to the action to only fetch the latest released App Store live version.

I'm currently preparing a PR for this (my first one ^^)

Buju77:
### Checklist

### Motivation and Context
This should fix https://github.com/fastlane/fastlane/issues/15308

### Description
Added a new `if version == 'live'` next to `if version == 'latest'` check to only fetch the dSYMs of latest live version.

### Testing Steps
Added a new test in `fastlane/spec/actions_specs/download_dsyms_spec.rb`

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -240,7 +240,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :version,
                                        short_option: "-v",
                                        env_name: "DOWNLOAD_DSYMS_VERSION",
-                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs or pass in 'live' to download only the latest live verion dSYMs",
+                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs or 'live' to download only the live verion dSYMs",
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :build_number,
                                        short_option: "-b",
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -31,7 +31,7 @@ module Fastlane
         # Set version if it is latest
         if version == 'latest'
           # Try to grab the edit version first, else fallback to live version
-          UI.message("Looking for latest or live version...")
+          UI.message("Looking for latest version...")
           latest_version = app.edit_version(platform: platform) || app.live_version(platform: platform)
 
           UI.user_error!("Could not find latest version for your app, please try setting a specific version") if latest_version.version.nil?
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -240,7 +240,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :version,
                                        short_option: "-v",
                                        env_name: "DOWNLOAD_DSYMS_VERSION",
-                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs. Pass in 'live' to download only the latest live build's dSYMs.",
+                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs or pass in 'live' to download only the latest live verion dSYMs",
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :build_number,
                                        short_option: "-b",
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -31,7 +31,7 @@ module Fastlane
         # Set version if it is latest
         if version == 'latest'
           # Try to grab the edit version first, else fallback to live version
-          UI.message("Looking for latest version...")
+          UI.message("Looking for latest or live version...")
           latest_version = app.edit_version(platform: platform) || app.live_version(platform: platform)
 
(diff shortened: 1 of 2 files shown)


fastlane/spec/actions_specs/download_dsyms_spec.rb
```diff
@@ -74,6 +75,36 @@
         end
       end
 
+      context 'when version is live' do
+        before do
+          # live
+          allow(app).to receive(:live_version).and_return(live_version)
```
Buju77: full disclosure: I was not sure how to adapt the tests with a live_version. I hope this is correct 😅

Buju77:
@googlebot I signed it!

Buju77:
@googlebot I fixed it.

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -231,7 +240,7 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :version,
                                        short_option: "-v",
                                        env_name: "DOWNLOAD_DSYMS_VERSION",
-                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs",
+                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs or pass in 'live' to download only the latest live verion dSYMs",
```
janpio:
```suggestion
                                       description: "The app version for dSYMs you wish to download, pass in 'latest' to download only the latest build's dSYMs or 'live' to download only the live version dSYMs",
```

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -45,6 +45,15 @@ def self.run(params)
             version = latest_candidate_build.train_version
             build_number = latest_candidate_build.build_version
           end
+        elsif version == 'live'
+          UI.message("Looking for live version...")
+          latest_live_version = app.live_version(platform: platform)
+
+          UI.user_error!("Could not find live version for your app, please try setting 'latest' or a specific version") if latest_live_version.version.nil?
```
joshdholtz: Should this be `if latest_live_version.nil?` I _think_ if there is no live version then `latest_live_version` will be nil 🤔

```suggestion
          UI.user_error!("Could not find live version for your app, please try setting 'latest' or a specific version") if latest_live_version.nil?
```

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -45,6 +45,15 @@ def self.run(params)
             version = latest_candidate_build.train_version
             build_number = latest_candidate_build.build_version
           end
+        elsif version == 'live'
+          UI.message("Looking for live version...")
+          latest_live_version = app.live_version(platform: platform)
```
joshdholtz: Can we rename this to just `live_version`?

fastlane/spec/actions_specs/download_dsyms_spec.rb
```diff
@@ -74,6 +75,36 @@
...
+          allow(live_version).to receive(:version).and_return('1.0.0')
+          allow(live_version).to receive(:build_version).and_return('42')
+          allow(live_version).to receive(:candidate_builds).and_return([build])
+          allow(build).to receive(:build_version).and_return('42')
+          allow(build).to receive(:upload_date).and_return(1_547_196_482_000)
+          # latest
```
joshdholtz: I think the `allows` from here and below can be deleted since we really only care about `app.live_version`, right?

joshdholtz (changes requested):
This looks good! Just a few small questions/comments and then this should be good to merge after that 😊

fastlane/lib/fastlane/actions/download_dsyms.rb
Buju77: ah true, good catch thx 😅

fastlane/spec/actions_specs/download_dsyms_spec.rb
Buju77: I'm new to this test setup, so I wasn't sure if I should leave it or not. Because my PR was to fix getting the live version, if there is already an edit_version *and* even a newer build train, which was the initial issue we discovered. should I remove it or keep it?

joshdholtz:
@googlebot I consent.

fastlane/spec/actions_specs/download_dsyms_spec.rb
joshdholtz: I added a commit to remove this for you! It can be removed because the latest code is being executed in a different if statement during the test so nothing needs to get mocked up for that 😊 After the tests run I will get this merged in. Thanks for contribution ❤️

joshdholtz (approved):
Thanks for the contribution ❤️

fastlane/spec/actions_specs/download_dsyms_spec.rb
Buju77: cool thx! 🎉

Issue #15296, opened 2019-09-12
fastlane match migrate leaves user stuck at picking role for service account


tpaik18:
Reopening https://github.com/fastlane/fastlane/issues/13971 reported by @janpio 

Using fastlane 2.130.0.

`match migrate` tells me to:

```
[21:16:30]: Now create a new JSON auth file by clicking on
[21:16:30]:
[21:16:30]:              1. Create credentials
[21:16:30]:              2. Service account key
[21:16:30]:              3. App Engine default service account
...
```
But "App Engine default service account" is not an option, only "New service account".
There I have a "Role" dropdown, but nothing that fully matches "App Engine default service account". I tried with the "App Engine" entries there, but then I get this error later after I have downloaded and copied the JSON to gc_keys.json:
```
/Library/Ruby/Gems/2.3.0/gems/google-cloud-storage-1.16.0/lib/google/cloud/storage/service.rb:570:in 
`rescue in execute': [!] forbidden: xxxxxx-signing@xxxxxxxx.iam.gserviceaccount.com 
does not have storage.buckets.list access to project xxxxxxxxxxx. 
(Google::Cloud::PermissionDeniedError)
```

Can you please update the instructions as to which Role in that big nested dropdown we should pick to have the proper access level?

rickclephas:
The documentation is referring to the already existing "App Engine default service account" service account.
The service accounts email ends with `@appspot.gserviceaccount.com`.
Under "Actions" (the three dots on the right) there is an option "Create key".

Using that key should work as this service account has the "Editor" role.

@janpio IMHO the documentation shouldn't encourage users to use this service account.
This service account has **[1945](https://console.cloud.google.com/iam-admin/roles/details/roles<editor)** permissions!
From my understanding `match` only needs access to a Google Cloud Storage bucket so if you ask me we should have documentation on creating a service account with the required permissions only.

rickclephas:
Alright it seems newly created projects don't have the App Engine default service account.
The service account is created after you [create](https://console.cloud.google.com/appengine/start) an App Engine application.

Creating a new service account with the "Editor" role has the same effect.

janpio:
Thanks @rickclephas, I unfortunately could these instructions to work because of 

> Alright it seems newly created projects don't have the App Engine default service account.
> The service account is created after you [create](https://console.cloud.google.com/appengine/start) an App Engine application.

... and I had created multiple issues about this in the past.

Can you summarize your suggestions or create a Pull Request to the code and docs?

Issue #13971, opened 2018-12-21
[match] Can't create JSON auth file for Google Cloud Storage (for `match migrate`)


janpio:
`match migrate` tells me to:

```
[21:16:30]: Now create a new JSON auth file by clicking on
[21:16:30]:
[21:16:30]:              1. Create credentials
[21:16:30]:              2. Service account key
[21:16:30]:              3. App Engine default service account
...
```

But "App Engine default service account" is not an option, only "New service account".
There I have a "Role" dropdown, but nothing that fully matches "App Engine default service account". I tried with the "App Engine" entries there, but then I got an error later in the process that the key doesn't have the correct permissions.

What am I doing wrong?

janpio:
Still unsolved :/

janpio:
Still unsolved.

PR #15310, opened 2019-09-14
[match] Improve Google Cloud Storage wizard

rickclephas:
### Checklist

### Motivation and Context

As described in #13971 and #15296 the current wizard makes some assumptions about the Google Cloud project.
With this PR I have tried to improve the wizard for newly created Google Cloud projects.

### Description

The new wizard doesn't assume the "App Engine default service account" service account exists.
Instead it instructs users to create a new service account.

The permissions have also been reduced. The "App Engine default service account" service account has the role "Editor" which can do almost anything with a users project.
The new wizard instructs users to give the service account the "Storage Admin" role on the bucket only. This greatly reduces any risks to other data in the Google project.

Unfortunately the bucket selection required the "storage.buckets.list" permission.
I couldn't find a standard role with that permissions without having too many other permissions so I changed the selection to a manual input.

I have also added a permission/existence check so we can show the "create a bucket" instructions or the "set permissions" instructions.

The changes @janpio made in #14457 have also been included.


match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -234,74 +235,120 @@ def ensure_keys_file_exists(google_cloud_keys_file, google_cloud_project_id)
...
-            UI.input("Once you're finished, please confirm with enter")
+            UI.input("Press [Enter] once you created a bucket")
+          end
+          bucket_name = UI.input("Enter the name of your bucket: ")
+
+          # Verify if the bucket exists
+          begin
+            bucket_exists = !self.gc_storage.bucket(bucket_name).nil?
+          rescue Google::Cloud::PermissionDeniedError
+            bucket_exists = true
```
joshdholtz: I'm a little unsure what behavior we actually want to do here 🤔 I tried running this to see what would happen when I _didn't_ create a bucket and I got a `Google::Cloud::PermissionDeniedError` so the wizard continued on with the permission check. Getting a permission denied makes sense because Google don't want to give out if a bucket exists or not to anyone so I'm just not sure if doing this check or handling this check like this is the correct approach.

I think would either need to:
1) Remove the exists check all together because its just going to give a permission failure if it doesn't
2) Keep the check but throw a warning saying something like "We couldn't find the bucket. Did you enter it correctly? If not, enter the correct name. If so, check the permissions."

Maybe something like that?

joshdholtz (changes requested):
This looks really good! Thanks for making the time to test out the wizard and improve it 😊 My one change request would be how we are handling the bucket exists check since its not really checking that. Once that is addressed this will be good to merge 💪

match/lib/match/storage/google_cloud_storage.rb
rickclephas: It is actually possible to know if a bucket exists or not. 
I tested this with a non existing bucket name like `rickclephas-fastlane-match-test-test`.
This will only check if a bucket exists in Google Cloud, it doesn't have to be in the users project.

rickclephas: So for clarification the bucket names are publicly visible and unique across GCS.
> Bucket names reside in a single Cloud Storage namespace. This mean that:
> - Every bucket name must be unique.
> - Bucket names are publicly visible.

https://cloud.google.com/storage/docs/naming

The bucket method will return `nil` if the bucket doesn't exists and throw a `PermissionDeniedError` if it does exists but the service account doesn't have access.

So users who enter a bucket name that does exists but is not theirs will pass the exists test but will fail the permission check.

When testing this make sure to use a name that you are sure doesn't exists. 
E.g. try with a domain you own (but doesn't have a bucket).

joshdholtz: Thanks for the clarification! I entered a weird bucket name to test this out but I guess I may have entered one that already existed and I just got really unlucky 😇 I will test again!

joshdholtz (approved):
Ran though tests again and looks good! Thanks so much for cleaning this up ❤️

Issue #15322, opened 2019-09-17
get_version_number can't find MARKETING_VERSION when target value is inherited from project settings


ilendemli:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
get_version_number can't find MARKETING_VERSION when target value is inherited from project settings and fails with the following error:
`[!] Unable to find Xcode build setting: MARKETING_VERSION`

##### Command executed
bundle exec fastlane run get_version_number

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                                    |
| --------------------------- | -------------------------------------------------------- |
| OS                          | 10.14.6                                                  |
| Ruby                        | 2.4.5                                                    |
| Bundler?                    | true                                                     |
| Git                         | git version 2.21.0 (Apple Git-122)                       |
| Installation Source         | ~/.rvm/gems/ruby-2.4.5/bin/fastlane                      |
| Host                        | Mac OS X 10.14.6 (18G95)                                 |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.4.5/lib                             |
| OpenSSL Version             | OpenSSL 1.1.1b  26 Feb 2019                              |
| Is contained                | false                                                    |
| Is homebrew                 | false                                                    |
| Is installed via Fabric.app | false                                                    |
| Xcode Path                  | /Applications/Xcode-11.GM.Seed.2.app/Contents/Developer/ |
| Xcode Version               | 11.0                                                     |

### System Locale

| Error                       |
| --------------------------- |
| No Locale with UTF8 found 🚫 |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
# This file contains the fastlane.tools configuration
# You can find the documentation at https://docs.fastlane.tools
#
# For a list of all available actions, check out
#
#     https://docs.fastlane.tools/actions
#
# For a list of all available plugins, check out
#
#     https://docs.fastlane.tools/plugins/available-plugins
#

# Uncomment the line if you want fastlane to automatically update itself
# update_fastlane

default_platform(:ios)

defaults = { :projectFile => "App.xcodeproj" }

private_lane :basic do
[...]

PR #15331, opened 2019-09-18
[action] Improve get_version_number: fallback to project settings if no MARKETING_VERSION is set for target

ilendemli:
### Checklist

### Motivation and Context
https://github.com/fastlane/fastlane/issues/15322

### Description
`get_version_number_from_build_settings` now returns `nil` if no version number is found instead of throwing an error
the `run` method will throw the error if version_number is not set
it will look into the project settings for `variable` (in this case `MARKETING_VERSION`) if there is no `version_number` set after checking the target

### Testing Steps
`bundle exec fastlane store` resulted with the correct value after these changes

### DIFF
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -18,19 +18,13 @@ module Fastlane
         plist_file = get_plist!(folder, target, configuration)
         version_number = get_version_number_from_plist!(plist_file)
 
-        # Get from build settings if needed (ex: $(MARKETING_VERSION) is default in Xcode 11)
+        # Get from build settings (or project settings) if needed (ex: $(MARKETING_VERSION) is default in Xcode 11)
         if version_number =~ /\$\(([\w\-]+)\)/
-          version_number = get_version_number_from_build_settings!(target, $1, configuration)
-
-          # Read $(MARKETING_VERSION) from project settings if version_number is not set in target settings
-          version_number ||= get_version_number_from_build_settings!(project, $1, configuration)
+          version_number = get_version_number_from_build_settings!(target, $1, configuration) || get_version_number_from_build_settings!(project, $1, configuration)
 
         # ${MARKETING_VERSION} also works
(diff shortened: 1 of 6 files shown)


ilendemli:
@googlebot I signed it!

michael-patzer:
This is an important improvement for us, and a use-case that I can see being common is setting the MARKETING_VERSION on the project and having all of the targets inherit that value with $(MARKETING_VERSION) to keep versions aligned across multiple related targets (app, watch, extensions). Thank you for this addition.

fastlane/lib/fastlane/actions/get_version_number.rb
```diff
@@ -26,6 +26,16 @@ def self.run(params)
           version_number = get_version_number_from_build_settings!(target, $1, configuration)
         end
 
+        # Check if version_number is not set and try retrieving value from project settings
+        if version_number.nil?
+          version_number = get_version_number_from_build_settings!(project, $1, configuration)
```
joshdholtz: This usage of $1 here seems to be a bit confusing 🤔 It isn't really tied to any specific regex here. I think this fallback almost needs to go inside the if statements above. Thoughts?

joshdholtz (changes requested):
I like the change but I think things just need to be moved around 😊

fastlane/lib/fastlane/actions/get_version_number.rb
ilendemli: you are absolutely correct, the downside would be having the same line twice, one in each if check

michael-patzer:
@ilendemli Looks like one of the CI tests failed (Xcode 9.0.1, Ruby 2.3). Bringing to your attention in case you hadn't seen it.

ilendemli:
@michael-patzer I have seen it fail, I am just not sure if it my fault that it is failing or if that was just an error that just happened. @joshdholtz Could you check this out? I can't rerun tests.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for making those requested changes 😊 I added some tests for them and rebased from master. Appreciate that contribution ❤️

PR #15243, opened 2019-08-29
[gym] Provide bcsymbolmap path to dsymutil for each architecture

leo150:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Currently `gym` doesn't map bcsymbolmap files with debug symbols for the project frameworks. The problem appears in the log as a warning: 

> [09:44:25]: Mapping dSYM(s) using generated BCSymbolMaps
> [09:44:25]: dsymutil --symbol-map /Users/user/Library/Developer/Xcode/Archives/2019-08-29/MyApp-1.1.0-2\ 2019-08-29\ 09.41.05.xcarchive/BCSymbolMaps /Users/user/Library/Developer/Xcode/Archives/2019-08-29/MyApp-1.0.0-1\ 2019-08-29\ 09.41.05.xcarchive/dSYMs/Alamofire.framework.dSYM
> [09:44:25]: warning: /Users/user/Library/Developer/Xcode/Archives/2019-08-29/MyApp-1.0.0-1 2019-08-29 09.41.05.xcarchive/BCSymbolMaps/Alamofire-armv7.bcsymbolmap: No such file or directory. Not unobfuscating.
> [09:44:25]: warning: /Users/user/Library/Developer/Xcode/Archives/2019-08-29/MyApp-1.0.0-1 2019-08-29 09.41.05.xcarchive/BCSymbolMaps/Alamofire-arm64.bcsymbolmap: No such file or directory. Not unobfuscating.

It can be observed in the logs of at least [16 issues](https://github.com/fastlane/fastlane/issues?utf8=%E2%9C%93&q=No+such+file+or+directory.+Not+unobfuscating). 
These two address the issue directly: #10902 and #14197

As result, dSYM files don't unobfuscate stacktrace completely:

> #0. Crashed: com.apple.main-thread
> 0  libswiftCore.dylib     0x21dee4ce0 specialized \_assertionFailure(_:_:file) + 260
> 1  libswiftCore.dylib      0x21bb071bc \_assertionFailure(_:_:file) + 42
> 2  Alamofire                   0x104e17414 (Missing) 
[...]

Issue #14197, opened 2019-02-06
[gym] "warning: /path/app-armv7.bcsymbolmap: No such file or directory. Not unobfuscating"


GandjaFuzz:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
What tool/action do you have a question about?
gym
Is this a question about documentation?
no
Is this a question about a third party plugin? (If so, please go to the plugin repository first)
no

#### Question Description

when archiving app with bitcode give errors. bitcode enable on all targets in app and in project.
```
			gym(
				scheme: FASTLANE_SCHEME,
				workspace: APP_WORKSPACE_PATH
			)
```
```
warning: /path/app-armv7.bcsymbolmap: No such file or directory. Not unobfuscating.
```
But symbols generated but names not readable, is view UUID. 
and give me many warnings, it is normal?
App Store release success this app. 

if i set for gym xcargs: "ENABLE_BITCODE=1", all warnings hide, but bitcode not generate. 
if use option include_bitcode it is not help.

PR #15346, opened 2019-09-20
[gym] let dsymutil figure out which .bcsymbolmap files to use.

pdrtrifork:
### Checklist

### Motivation and Context
Sometimes we end up with partly symbolicated crash reports after using the dSYM.zip file generated by fastlane. It turns out that fastlane sometimes fails to combine DWARF files and .bcsymbolmap files.

### Description
Alongside a DWARF file, there may be an associated .plist file containing the key `DBGOriginalUUID` with an UUID as the value. This UUID points to the .bcsymbolmap file to be used for converting _hidden symbols.

The `dsymutil` utility used to combine a DWARF and a.bcsymbolmap allows a directory containing .bcsymbolmap files to be specified, instead of specifying a single .bcsymbolmap file. This lets `dsymutil` figure out which .bcsymbolmap file to use, simplifying the fastlane code.

### DIFF
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -146,7 +146,6 @@ module Gym
       containing_directory = File.expand_path("..", PackageCommandGenerator.dsym_path)
       bcsymbolmaps_directory = File.expand_path("../../BCSymbolMaps", PackageCommandGenerator.dsym_path)
       available_dsyms = Dir.glob("#{containing_directory}/*.dSYM")
-      uuid_regex = /[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}/
 
       if Dir.exist?(bcsymbolmaps_directory)
         UI.message("Mapping dSYM(s) using generated BCSymbolMaps") unless Gym.config[:silent]
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -146,6 +146,7 @@ module Gym
       containing_directory = File.expand_path("..", PackageCommandGenerator.dsym_path)
       bcsymbolmaps_directory = File.expand_path("../../BCSymbolMaps", PackageCommandGenerator.dsym_path)
       available_dsyms = Dir.glob("#{containing_directory}/*.dSYM")
+      uuid_regex = /[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}/
 
       if Dir.exist?(bcsymbolmaps_directory)
         UI.message("Mapping dSYM(s) using generated BCSymbolMaps") unless Gym.config[:silent]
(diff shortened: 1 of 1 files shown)


joshdholtz (commented):
This looks like a revert of #15243 🤔 It seemed like that PR added more symbols that were getting skipped. Was that not the case?

janpio:
Note that this PR only was merged a few days ago and released with the last release. Is this a super new problem you are observing @pdrtrifork?

pdrtrifork:
Yes, it is a very new problem. We have been using fastlane for quite some time with no symbolication issues since bcsymbolmap-support was added.

I wanted to check that our symbolication subsystem is still working after installing the official iOS 13 release, and needed to build a new release of our internal test-app.

Got a new Mac a short while ago, and decided to upgrade our test-app to the latest fastlane release, and immediately ran in to this problem. I initially thought we had messed up something at our end, but the fastlane-generated DWARF files contain _hidden# symbols. This pull request fixed my problem, but I will have a look at #15243.

joshdholtz:
@pdrtrifork Hey! Just wanted to check in if you were able to look at the latest master version of fastlane to test if #15243 fixed things for you 😊

bamx23:
I face the same problem as @pdrtrifork. As said before:

> Alongside a DWARF file, there may be an associated .plist file containing the key DBGOriginalUUID with an UUID as the value. This UUID points to the .bcsymbolmap file to be used for converting _hidden symbols.

`dwarfdump` shows only UUID of the DWARF inside dSYM. So for dSYMs after export(with recompilation) we should use UUID from `DBGOriginalUUID` in Info.plist. 

Another solution is to parse the output of `dsymutil` and seach for `No such file or directory. Not unobfuscating.`. If found we should call `dsymutil` with direcory or parse Info.plist and use `DBGOriginalUUID`.

joshdholtz:
@bamx23 Thanks for the explanation! So it looks like (to me) that this PR is a full revert of #15243, correct? So this would undo the fixes that that PR made for https://github.com/fastlane/fastlane/issues/10902 and https://github.com/fastlane/fastlane/issues/14197? So it seems like this PR needs a bit more added to it to fully support those?

I appreciate your expertise here 😊🙌

pdrtrifork:
> @pdrtrifork Hey! Just wanted to check in if you were able to look at the latest master version of fastlane to test if #15243 fixed things for you 😊

No, I believe #15243 is the change that messed things up for us :-)

I will update this PR, adding a solution to the issue fixed in #15243 as well. Stay tuned :-)

pdrtrifork:
The latest commit uses a different approach:
- if an <uuid>.plist exists next to the DWARF file, the uuid of the bcsymbolmap file is extracted, and used, if the file exists
- else, <uuid>.bcsymbolmap is used, if the file exists
- else, let dsymutil figure it out.

As we have had no issues in our builds getting dsymutil to behave correctly, I am not able to test if this improves the situation for those having issues.

Also, I don't really speak Ruby, so any comments regarding coding style are appreciated :-)

pdrtrifork:
@googlebot I consent.

joshdholtz:
@pdrtrifork It looks like a rebase is needed here! I tried to rebase and push up to the branch to fix this but it doesn't look like I have access to push to this branch for some reason even though the GitHub UI says I should 🤔 Would you be able to manually add me?

pdrtrifork:
@joshdholtz I have added you as a collaborator.

bhstrifork:
Hi! I've fixed the PR so that it only contains the commits made by @pdrtrifork. Not sure why CircleCI is still complaining about "Execute modules load up tests".

Could you give it a look @joshdholtz ?

janpio:
```
ERROR:  Error installing fastlane-2.133.0.gem:
	The last version of signet (~> 0.9) to support your Ruby & RubyGems was 0.11.0. Try installing it with `gem install signet -v 0.11.0` and then running the current command again
	signet requires Ruby version >= 2.4.0. The current ruby version is 2.3.8.459.
```
so probably https://github.com/fastlane/fastlane/issues/15522

bamx23:
Tests have passed. 
This PR is important for dSYM upload automations so I hope it is merged soon.

gym/lib/gym/runner.rb
```diff
@@ -155,31 +155,45 @@ def compress_and_move_dsym
...
+            if File.exist?(dsym_plist_file_path)
+              dsym_plist = Plist.parse_xml(dsym_plist_file_path)
+              original_uuid = dsym_plist['DBGOriginalUUID']
+              possible_symbol_map_path = "#{bcsymbolmaps_directory}/#{original_uuid}.bcsymbolmap"
+              if File.exist?(possible_symbol_map_path)
+                symbol_map_path = possible_symbol_map_path.shellescape
+              end
+            end
+            if symbol_map_path.nil?
+              possible_symbol_map_path = File.join(bcsymbolmaps_directory, "#{uuid}.plist")
```
darrenbrooke: Should this be looking for `uuid.bcsymbolmap` rather than`uuid.plist` ?

pdrtrifork: Yes, you are absolutely right. Thank you for noticing this bug.

pdrtrifork:
We have fixed the issue noted above, and hope that this pull request can be merged soon :-)

kringelbach:
@joshdholtz Will you be able to review this PR, or can anyone else do it? Hope to hear from you soon :-)

joshdholtz:
Ah yes yes!! I can take a look at this one over the weekend 😊

bamx23:
Any news here?

kringelbach:
@joshdholtz Happy New Year 🎉😊
Any chance you will have time to take a look at this soon?

joshdholtz (approved):
I think this looks good to me 😊 Thanks for fixing this! 💪 ❤️

PR #15349, opened 2019-09-21
[snapshot] Specify custom path for erb

### LINKED ISSUE #12892: Customize HTML generated by Fastlane Snapshot

It would be very handy to customise the HTML generated by Fastlane Snapshot. This would effectively allow us to generate customised reports, apply custom CSS, etc.

#### Feature Description
Specify a custom `page.html.erb` as an option in the `Snapfile`.

Only drawback I can think of is that this can break fastlane snapshot in an update if e.g. the bindings change, however, this can be specified in the documentation.

nderkach:
### Checklist

### Motivation and Context
Resolves #12892

This PR addresses the issue outlined in https://github.com/fastlane/fastlane/issues/12892 and allows to specify a custom `.irb` file for the snapshot HTML template.

### Description
I added a configuration option for `Snapfile`: `html_template` and provided a custom `.irb` file with a template. Then I ran `fastlane snapshot` and generated a `screenshots.html` file using my custom template.


janpio:
In code the param seems to be called `page_html_path`, not `html_template`, correct?
This should probably also be added to https://github.com/fastlane/fastlane/blob/master/snapshot/lib/snapshot/options.rb so it can be specified as action param and be discovered in the documentation.

snapshot/lib/snapshot/options.rb
```diff
@@ -137,6 +137,12 @@ def self.available_options
                                      description: "A list of videos that should be added to the simulator before running the application",
                                      type: Array,
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :html_template,
+                                     env_name: 'SNAPSHOT_HTML_TEMPLATE',
+                                     short_option: "-e",
+                                     description: "A path to screenshots.html template",
+                                     default_value: false,
```
janpio: Would it maybe make sense to set the default to `File.join(Snapshot::ROOT, "lib", "snapshot/page.html.erb")`? Then you could simplify the code a bit and also clean up the naming of the variables a bit more.

janpio (approved):

Issue #15211, opened 2019-08-21
require generates warning for stdlib libraries


lsegal:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Reproduction steps:

1. Place the following in a Fastfile:

```ruby
require 'yaml'

lane :test do
end
```

You will see the following warning:

```
[18:23:43]: You have required a gem, if this is a third party gem, please use `fastlane_require 'yaml'` to ensure the gem is installed locally.
```

This warning should not display on Ruby core or Ruby stdlib files.

2. Run `bundle exec fastlane test`

##### Command executed

```sh
bundle exec fastlane test
```

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>bundle exec fastlane test
 [✔] �
[18:23:43]: You have required a gem, if this is a third party gem, please use `fastlane_require 'yaml'` to ensure the gem is installed locally.
[18:23:43]: ------------------------------
[18:23:43]: --- Step: default_platform ---
[18:23:43]: ------------------------------
[18:23:43]: Driving the lane 'test' �

+------+------------------+-------------+
|           fastlane summary            |
+------+------------------+-------------+
| Step | Action           | Time (in s) |
+------+------------------+-------------+
| 1    | default_platform | 0           |
+------+------------------+-------------+

[18:23:43]: fastlane.tools finished successfully �
</pre>
</details>

### Environment

<details><summary>≡ƒÜ½ fastlane environment ≡ƒÜ½</summary>

### Stack

| Key                         | Value                        |
| --------------------------- | ---------------------------- |
| OS                          | UNKNOWN                      |
| Ruby                        | 2.6.1                        |
| Bundler?                    | true                         |
| Git                         | git version 2.19.1.windows.1 |
| Installation Source         | C:/tools/ruby26/bin/fastlane |
| Host                        | Unknown  ()                  |
| Ruby Lib Dir                | C:/tools/ruby26/lib          |
| OpenSSL Version             | OpenSSL 1.1.1a  20 Nov 2018  |
| Is contained                | false                        |
| Is homebrew                 | false                        |
| Is installed via Fabric.app | false                        |

### System Locale

[...]

PR #15354, opened 2019-09-22
[fastlane] attempt to require lib in Fastfile before warning

lsegal:
### Checklist

### Motivation and Context

See #15211 for issue details

### Description

Prior to warning the user about a missing gem (with installation guide), attempt to require the library. This avoids issues where the library being required is not actually a gem but a local Ruby file or gem provided by Ruby's stdlib.

Tested with RSpec with new tests & fixtures added.

### Testing Steps

1. `bundle exec rspec`

You can also test manually by using something like `require 'base64'` inside of a Fastfile and expecting no warning. You should still see a warning with any gem that is not installed on your system (i.e. anything that will lead to a LoadError).

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -26,10 +26,15 @@ module Fastlane
                 'you should turn off smart quotes in your editor of choice.')
       end
 
-      content.scan(/^\s*require (.*)/).each do |current|
+      content.scan(/^\s*require ["'](.*?)["']/).each do |current|
         gem_name = current.last
         next if gem_name.include?(".") # these are local gems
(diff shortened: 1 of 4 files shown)


lsegal:
FYI: a cleaner way to do this would be to handle LoadErrors from the eval in `parse` and bubble them up but I avoided this because it can change behavior in a subtle way. It would be a much more consistent experience, though, since there are current;y a number of ways to accidentally avoid this warning, like using any number of different require syntaxes.

joshdholtz (approved):
I actually really like this PR! I had to swap between current master and this PR to test out the changes and this fixes a lot of things I actually never noticed 😇 Thanks for doing this!

PR #15407, opened 2019-09-29
Revert "pin faraday to < 0.16.0 (#15399)"

janpio:
This reverts commit 596a388da317f847100587668002ce80f25f1ed0.

Now that `faraday` released a new version that undoes the problems, we can go back to using their newest version.


joshdholtz:
@janpio Do we need to do this revert? 🤔

janpio:
We don't need to, we can keep using the old version forever. But we probably do not want to do that to get updates and fixes of course.

This will only be a valid PR when 0.16.3 is released though, it should fix the failing tests again.

joshdholtz (approved):
Looks good to me! Thanks for fixing 😍

PR #15421, opened 2019-10-01
[action] remove nightly from update_fastlane

joshdholtz:
### Motivation and Context
Remove nightly builds because the bot that uploaded them was not able to have 2FA enabled which made the bot susceptible for take over 😱

### Description
- Deprecated option for `nightly`
- Also removed deprecated `tools` action as super no longer relevant

### DIFF
--- a/fastlane/lib/fastlane/actions/update_fastlane.rb
+++ b/fastlane/lib/fastlane/actions/update_fastlane.rb
@@ -10,18 +10,10 @@ module Fastlane
       def self.run(options)
         return if options[:no_update] # this is used to update itself
 
-        tools_to_update = options[:tools].split(',') unless options[:tools].nil?
-        tools_to_update ||= all_installed_tools
-
-        if tools_to_update.count == 0
-          UI.error("No tools specified or couldn't find any installed fastlane.tools")
-          return
-        end
-
+        tools_to_update = ALL_TOOLS
         UI.message("Looking for updates for #{tools_to_update.join(', ')}...")
 
         updater = Gem::CommandManager.instance[:update]
-        updater.options[:prerelease] = true if options[:nightly]
         cleaner = Gem::CommandManager.instance[:cleanup]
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/update_fastlane.rb
```diff
@@ -134,21 +98,17 @@ def self.details
...
-                                       env_name: "FL_UPDATE_FASTLANE_NIGHTLY",
-                                       description: "Opt-in to install and use nightly fastlane builds",
-                                       is_string: false,
-                                       default_value: false),
           FastlaneCore::ConfigItem.new(key: :no_update,
                                        env_name: "FL_NO_UPDATE",
                                        description: "Don't update during this run. This is used internally",
                                        is_string: false,
                                        default_value: false),
-          FastlaneCore::ConfigItem.new(key: :tools,
```
janpio: Was this intentional that this is also removed in this PR?

fastlane/lib/fastlane/actions/update_fastlane.rb
```diff
@@ -134,21 +98,17 @@ def self.details
...
-                                       env_name: "FL_TOOLS_TO_UPDATE",
-                                       description: "Comma separated list of fastlane tools to update (e.g. `fastlane,deliver,sigh`)",
-                                       deprecated: true,
-                                       optional: true)
```
janpio: Was this intentionally removed in this PR?

joshdholtz: Yup yup! It is sooooo old. This is when _fastlane_ had multiple gems (one for each tool). I was triggered when I saw it so I wanted it gone too while I was in here 🙃

janpio: Bye bye then

janpio (approved):

PR #15471, opened 2019-10-11
Make Minimum Ruby Version 2.4

getaaron:
### Checklist

### Motivation and Context

Currently we support Ruby 2.0+. But Ruby 2.3 and earlier is end-of-life with no security updates. Nobody should be using it for anything as it's now considered insecure.

Additionally we recently added the `signet` Rubygem as a _fastlane_ dependency, which requires Ruby version >= 2.4.0.

We should probably make a major version bump for this (although we probably should've when adding signet).

### Description
- Update `.gemspec` file with 2.4 minimum
- Update Travis and CircleCI with 2.4 minimum
- Delete comment that applied to Ruby <2.2
- Misc. other stuff

### Testing Steps
- rspec and rubocop both pass


getaaron:
@fastlane/team could use some help resolving the `Execute tests on Ubuntu` failure, not sure where that Docker `Gemfile` issue is coming from

milch:
FYI this is already creating problems in other places where a major version bump was not done when upgrading to 2.4: https://github.com/googleapis/google-api-ruby-client/issues/833

I would **strongly** recommend not to merge this without also doing a major version bump.

endocrimes:
@milch If dependencies we didn't pin aggressively enough are already causing issues, then I don't think there's much harm in doing this, because the requirement is already there transitively, and this would leave a more productive error message for most users (who should already be running >=2.4 on most supported configurations of fastlane).

That being said, CircleCI config changes lgtm, but we'll also need to build a new CI image with the new ruby version and bump that.

endocrimes:
@getaaron I can honestly say, I don't remember - @janpio might tho

fastlane.gemspec
```diff
@@ -45,7 +45,7 @@ Gem::Specification.new do |spec|
     "docs_url" => "https://docs.fastlane.tools"
   }
 
-  spec.required_ruby_version = '>= 2.0.0'
+  spec.required_ruby_version = '>= 2.4.0'
```
joshdholtz: IMO, we can't do this without a major version bump. This could potentially break a lot of setups where Ruby 2.4 isn't installed by default and a bunch of CIs. If/when we are going to do this, we should really add a warning for a while first saying that this is going to happen.

joshdholtz (commented):
I would personally be fine with a 2.4 requirement bump but this is a breaking change for users that can't be done without a major _fastlane_ version bump and/or a warning

joshdholtz:
@getaaron Also, the docker file that is used is here - https://github.com/fastlane/docker 😊 This is used by both https://github.com/fastlane/fastlane and https://github.com/fastlane/docs

janpio:
> Additionally we recently added the `signet` Rubygem as a _fastlane_ dependency, which requires Ruby version >= 2.4.0.

Where and how did that happen? I couldn't find `signet` in the codebase.

> We should probably make a major version bump for this (although we probably should've when adding signet).

Wouldn't it be possible to pin `signet` to an earlier version that still works with Ruby 2.3?

joshdholtz:
@janpio @getaaron I have a fix for this that doesn't require the Ruby version bump here - https://github.com/fastlane/fastlane/pull/15483

milch:
> Wouldn't it be possible to pin signet to an earlier version that still works with Ruby 2.3?

+1 to that. Note that signet is consumed transitively through `google-api-client`.

According to the response in googleapis/google-api-ruby-client#833 they will stay on a 0 major version with breaking changes between minor versions. I’d say it would almost make sense to pin to specific versions in that case, otherwise you’ll run into implicit version dependencies where some versions of fastlane only work with some other versions of the google api client in a relationship that is not expressed in the gemspec.

> If dependencies we didn't pin aggressively enough are already causing issues, then I don't think there's much harm in doing this, because the requirement is already there transitively, and this would leave a more productive error message for most users

I disagree, there’s still room to fix. As time goes on you’ll have more and more people updating their dependencies, and this will cause issues for them. One way to fix would be by releasing another version pinning an earlier version of the api client, then doing a major version bump with the new ruby requirement.

>  (who should already be running >=2.4 on most supported configurations of fastlane).

I’m not so sure about that, 2.3 is still the default on Mojave. With Catalina people will probably install the latest, but not everyone upgrades to Catalina immediately

fastlane.gemspec
getaaron: I agree with this being a major version bump @joshdholtz. Although I’m not sure we need an advance warning, since nobody is forcing them to upgrade and the `update_fastlane` action will ignore major version bumps.

getaaron:
Closing in favor of https://github.com/fastlane/fastlane/pull/15483

PR #15475, opened 2019-10-11
[fastlane_core] add clearer error message when incorrectly defining tests that need xcodebuild

revolter:
### Checklist

### Motivation and Context

This is to prevent issues like: https://circleci.com/gh/fastlane/fastlane/36544

### Description

image

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -163,7 +163,12 @@ module FastlaneCore
 
     # @return true if Xcode version is higher than 8.3
     def self.xcode_at_least?(version)
-      FastlaneCore::UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
+      if self.ci?
+        error_message = 'Forgot to set `requires_xcodebuild` to `true` for the test: `it \'does someting\', requires_xcodebuild: true do ... end`'
+      else
+        error_message = 'Unable to locate Xcode. Please make sure to have Xcode installed on your machine'
+      end
+      FastlaneCore::UI.user_error!(error_message) if xcode_version.nil?
       v = xcode_version
       Gem::Version.new(v) >= Gem::Version.new(version)
     end
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -164,7 +164,7 @@ module FastlaneCore
     # @return true if Xcode version is higher than 8.3
     def self.xcode_at_least?(version)
       if self.ci?
-        error_message = 'Forgot to set `requires_xcodebuild` to `true` for the test: `it \'does someting\', requires_xcodebuild: true do ... end`'
+        error_message = 'Forgot to set `requires_xcodebuild` to `true` for the test (e.g. `it \'does someting\', requires_xcodebuild: true do ... end`)'
       else
         error_message = 'Unable to locate Xcode. Please make sure to have Xcode installed on your machine'
       end


revolter:
But, I now noticed that we have tests running on macOS with Xcode installed, so, why this parameter?

fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

janpio (changes requested):
`requires_xcodebuild: true` is to being able to disable test on Windows and Linux that require Xcode to be installed, often via the Xcode version check you also changed. 

If you get these on macOS, something else is broken.

revolter:
Hmm, I didn't pay much attention to the OS, and maybe I incorrectly thought that they failed on macOS.

janpio:
The GH run you linked to was Ubuntu, so it would make sense.

Did I understand correctly that you added additional tests, and did not set `requires_xcodebuild` thus giving you this error message? Then you changes may be useful after all - although in theory this would have to be output at _all_ locations where we use Xcode/`xcodebuild`. Not optimal.

Btw: I think `require_xcode` and `require_xcodebuild` might be different things, you have to check the codebase.

revolter:
> The GH run you linked to was Ubuntu, so it would make sense.

Exactly. So https://github.com/fastlane/fastlane/pull/15475#issuecomment-541015559 came out of confusion, so it should be ignored.

> Did I understand correctly that you added additional tests, and did not set `requires_xcodebuild` thus giving you this error message?

Yes, and it took me quite a while to figure it out. Hence, this change.

> although in theory this would have to be output at all locations where we use Xcode/`xcodebuild`. Not optimal.

What do you mean by this? If _x_ tests that indirectly use `xcodebuild`, don't specify that parameter, the error message that I added **will** be output _x_ times.

> Btw: I think `require_xcode` and `require_xcodebuild` might be different things, you have to check the codebase.

Meaning that the PR title is a bit misleading? _(I updated it now ⬆️)_

janpio:
> What do you mean by this? If _x_ tests that indirectly use `xcodebuild`, don't specify that parameter, the error message that I added **will** be output _x_ times.
> 
> >

Only if this exact code is executed. There are x other locations that _also_ use `xcodebuild` or Xcode and which do not output this error message.

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -163,7 +163,12 @@ def self.xcode_version
 
     # @return true if Xcode version is higher than 8.3
     def self.xcode_at_least?(version)
-      FastlaneCore::UI.user_error!("Unable to locate Xcode. Please make sure to have Xcode installed on your machine") if xcode_version.nil?
+      if self.ci?
```
joshdholtz: Do we actually want to make this `self.mac?`? 🤔  Since this would be nice to see on any linux/windows run and not necessarily just on CI?

joshdholtz (commented):
Love this change! Just one quick question on the logic 😇

fastlane_core/lib/fastlane_core/helper.rb
janpio: Not really, this is explicit communication to a developer. As I mentioned above, this is only 1 location out of hundreds that use Xcode (although one of the most "popular" ones), so this is also not complete.

revolter: Why not? Maybe a Linux user clones the branch of someone using macOS that added some tests, but forgot to specify that flag. In which case, for that other person, the tests pas, but for the Linux user, they fail. And with this PR and Josh's change, he will knew exactly what the problem is, when running the tests locally, because the error message is clearer.

And back to your point, if xcodebuild is used from multiple places (which I'm not doubting), how does it work on Linux? And, we could fix it (e.g. make them all pass through a single function) in another PR, right? Meaning, if this is ok, it could be merged, right?

revolter:
I don't think that this is needed anymore, hence, closing it.

PR #15480, opened 2019-10-11
[action] cocoapods - add `--clean-install`

att55:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

`--no-clean` option has removed and deprecated in fastlane for some time, but `--clean-install` option has introduced from CocoaPods version 1.7.0.
https://github.com/CocoaPods/CocoaPods/releases

So, I add the option to cocoapods action to use same option.

### related PR

I also updated docs.
https://github.com/fastlane/docs/pull/871

### Description

If `clean_install` option is true and  pods version is 1.7 or more, `--clean-install` option is added to pod install command.

### Options check

- CocoaPods ver: 1.6.0

```
$ bundle exec pod install help

Usage:

    $ pod install

      Downloads all dependencies defined in `Podfile` and creates an Xcode Pods
      library project in `./Pods`.

      The Xcode project file should be specified in your `Podfile` like this:

          project 'path/to/XcodeProject.xcodeproj'

      If no project is specified, then a search for an Xcode project will be made. If
      more than one Xcode project is found, the command will raise an error.

      This will configure the project to reference the Pods static library, add a
      build configuration file, and add a post build script to copy Pod resources.

      This may return one of several error codes if it encounters problems. * `1`
      Generic error code * `31` Spec not found (i.e out-of-date source repos, mistyped
      Pod name etc...)

Options:

    --repo-update                       Force running `pod repo update` before install
    --deployment                        Disallow any changes to the Podfile or the
                                        Podfile.lock during installation
    --project-directory=/project/dir/   The path to the root of the project directory
    --silent                            Show nothing
    --verbose                           Show more debugging information
    --no-ansi                           Show output without ANSI codes
    --help                              Show help banner of specified command
```

- CocoaPods ver: 1.7.0 ~

```
$ bundle exec pod install help

Usage:

    $ pod install

      Downloads all dependencies defined in `Podfile` and creates an Xcode Pods
      library project in `./Pods`.

      The Xcode project file should be specified in your `Podfile` like this:

          project 'path/to/XcodeProject.xcodeproj'

      If no project is specified, then a search for an Xcode project will be made. If
[...]

PR #15482, opened 2019-10-12
[action] carthage - add `no-checkout` option

att55:
### Checklist

### Motivation and Context

Latest (ver: 0.33.0) carthage bootstrap command has `--no-checkout` option.
But, carthage action isn't implemented the option.
So, I will add `no_checkout` option to carthage action.

### related PR

I also updated docs.
https://github.com/fastlane/docs/pull/872

### carthage bootstrap option

```
❯ carthage version
0.33.0
❯ carthage help bootstrap
Check out and build the project's dependencies

[--no-checkout]
	skip the checking out of dependencies after updating

[--no-build]
	skip the building of dependencies after updating
	(ignored if --no-checkout option is present)

[--verbose]
	print xcodebuild output inline (ignored if --no-build option is present)

[--log-path (string)]
	path to the xcode build output. A temporary file is used by default

[--new-resolver]
	use the new resolver codeline when calculating dependencies. Default is false

[--configuration (string)]
	the Xcode configuration to build
	(ignored if --no-build option is present)

[--platform (platform)]
	the platforms to build for (one of 'all', 'macOS', 'iOS', 'watchOS', 'tvOS', or comma-separated values of the formers except for 'all')
	(ignored if --no-build option is present)

[--toolchain (string)]
	the toolchain to build with

[--derived-data (string)]
	path to the custom derived data folder

[--cache-builds]
	use cached builds when possible

[--no-use-binaries]
	don't use downloaded binaries when possible

[--use-ssh]
	use SSH for downloading GitHub repositories

[--use-submodules]
	add dependencies as Git submodules

[--color (color)]
	whether to apply color and terminal formatting (one of 'auto', 'always', or 'never')

[--project-directory (string)]
	the directory containing the Carthage project

[<dependency names…>]
	the dependency names to update, checkout and build
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
-            echo "2.3" > .ruby-version
+            echo "2.4" > .ruby-version
             gem update --system
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
@@ -209,7 +209,7 @@ jobs:
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
     docker:
-      - image: circleci/ruby:2.3
+      - image: circleci/ruby:2.4
     shell: /bin/bash --login -eo pipefail
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


janpio:
Thanks for this PR!

Re the CI changes, same as https://github.com/fastlane/fastlane/pull/15480#issuecomment-541336544 please. Thanks.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for the change and the added test 😍

PR #15483, opened 2019-10-12
[fastlane] Lock signet to 0.11 for under Ruby 2.4

### LINKED ISSUE #15484: Build on `master` is failing because of `signet`
https://circleci.com/gh/fastlane/fastlane/36719

```
ERROR:  Error installing fastlane-2.133.0.gem:
	The last version of signet (~> 0.9) to support your Ruby & RubyGems was 0.11.0. Try installing it with `gem install signet -v 0.11.0` and then running the current command again
	signet requires Ruby version >= 2.4.0. The current ruby version is 2.3.8.459.
```

Related discussion: https://github.com/fastlane/fastlane/pull/15471

joshdholtz:
### Motivation and Context
Fixes #15484
Fixes issues fixed by #15471 without minimum Ruby version bump

### Description
The issue raised with `signet` [requiring Ruby version 2.4.0](https://github.com/googleapis/signet/pull/145) does not occur when running `bundler install` or `bundle update` (because bundler respects Ruby version requirements when installing) but was with `gem install` (which does not respect Ruby versions when resolving dependencies).

This fix with manually check if the Ruby version is less than 2.4 and if so it will lock the `signet` dependency to less than `0.11`.

#### bundler with Ruby 2.3
Installed `signet (0.11)`

```sh
bundle update
```

#### bundler with Ruby 2.4
Installed `signet (0.12)`

```sh
bundle update
```

#### gem install with Ruby 2.3
Installed `signet (0.11)`

```sh
gem cleanup fastlane
gem cleanup signet
gem install $(echo $(gem build fastlane.gemspec) | sed 's/.*File: //')
gem list | grep signet
```

#### gem install Ruby 2.4
Installed `signet (0.12)`

```sh
gem cleanup fastlane
gem cleanup signet
gem install $(echo $(gem build fastlane.gemspec) | sed 's/.*File: //')
gem list | grep signet
```

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -92,7 +92,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency('simctl', '~> 1.6.3') # Used for querying and interacting with iOS simulators
   spec.add_dependency('jwt', '~> 2.1.0') # Used for generating authentication tokens for AppStore connect api
 
-  # need to lock 0.11 and under when using less than Ruby 2.4
+  # need to lock 0.11 and under when using less than Ruby 2.4 to prevent install issues when using 'gem install'
+  # 'gem install' does not respect Ruby versions when and would try installing 0.12 on Ruby 2.3 or less
+  # https://github.com/fastlane/fastlane/pull/15483
   if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4')
     spec.add_dependency('signet', '<= 0.11') # Because yeah
   end
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -91,7 +91,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('bundler', '>= 1.12.0', '< 3.0.0') # Used for fastlane plugins
   spec.add_dependency('simctl', '~> 1.6.3') # Used for querying and interacting with iOS simulators
   spec.add_dependency('jwt', '~> 2.1.0') # Used for generating authentication tokens for AppStore connect api
-  #spec.add_dependency('signet', '>= 0.9') # Because yeah
+  spec.add_dependency('signet', '~> 0.9') # Because yeah
   # Space to test something
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -91,6 +91,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('bundler', '>= 1.12.0', '< 3.0.0') # Used for fastlane plugins
   spec.add_dependency('simctl', '~> 1.6.3') # Used for querying and interacting with iOS simulators
   spec.add_dependency('jwt', '~> 2.1.0') # Used for generating authentication tokens for AppStore connect api
+  spec.add_dependency('signet', '>= 0.9') # Because yeah
   # Space to test something
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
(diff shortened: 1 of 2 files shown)


getaaron:
Personally I’m uncomfortable with the level of security risk. Pinning an old version of an oauth library to an EOL version of Ruby, neither of which will receive any security updates could have consequences that are hard to mitigate if a vulnerability is discovered.

Aside from that (e.g. if you disagree with my risk assessment) the fix looks good.

joshdholtz:
@getaaron I totally agree that there is a level of security risk that locking in to Ruby 2.4 solves but it creates a whole new issue for users that can't or will struggle with changing to that version as it is breaking.

I would say that in the install instructions or during runtime we recommend switching to 2.4 or greater for security risks but I really don't think that we can lock to 2.4 at this time 😔

milch:
IMO: 

- Locking the oauth client is less likely to cause an issue, since it could only be exploited in a limited set of circumstances (is there even user input that influences the oauth flow?). It does still increase the attack surface, however
- Using an EOL version of ruby presents the much bigger attack surface 

However, I don’t think fastlane should be the one to choose for the user that they must upgrade their ruby version, unless there is a specific technical requirement for it. Eventually I’d do a major version bump with Ruby 2.4 (or even 2.5 or later) required, which will cause users to upgrade organically as new features are added and bugs they encounter are fixed. I think adding a warning message (even to this PR? `“WARN: Locking to a potentially insecure version of signet because you are using a version of ruby which is marked as End-Of-Life. Please upgrade your ruby installation to 2.4 or later”` with a link to docs on how to do that) is a good way to let users know in advance that they need to do this, and empower them to do it on their own timeline

fastlane.gemspec
```diff
@@ -92,6 +92,13 @@ Gem::Specification.new do |spec|
   spec.add_dependency('simctl', '~> 1.6.3') # Used for querying and interacting with iOS simulators
   spec.add_dependency('jwt', '~> 2.1.0') # Used for generating authentication tokens for AppStore connect api
 
+  # need to lock 0.11 and under when using less than Ruby 2.4 to prevent install issues when using 'gem install'
+  # 'gem install' does not respect Ruby versions when and would try installing 0.12 on Ruby 2.3 or less
+  # https://github.com/fastlane/fastlane/pull/15483
+  if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4')
+    spec.add_dependency('signet', '<= 0.11') # Because yeah
```
milch: See my other comment, I think we could add a log message here, something like 

```
WARN: Locking to a potentially insecure version of signet because you are using a version of ruby which is marked as End-Of-Life. Please upgrade your ruby installation to 2.4 or later
```

I don’t remember whether gem shows that during installation though, it might suppress output. Something worth to test though

joshdholtz: I will test it out 👌

joshdholtz: @milch I have to do an `STDERR.puts` for it to show (which I'm pretty sure is what the other warnings use that rubygems outputs

milch: I think `STDERR` is definitely appropriate for this 👌

joshdholtz: @milch Perfect! New commit added 😊

milch (approved):

fastlane.gemspec
milch: Cool! 

As a side note, thinking about WWFD (what would @KrauseFx do), it’d probably be to figure out how the user installed ruby and then show them the commands they need to run to do the upgrade. I maybe wouldn’t do that right away ‘cause people may not need it, but may be something to consider if users are confused about what to do and start sending in issues about it

janpio (approved):
Very nice, didn't expect it would be possible to change dependencies depending on Ruby version. Ruby <3

getaaron (approved):

getaaron:
@joshdholtz i think this is good to merge which should fix the downstream dependency issues. I can start working on suggested 2.4+ installation instructions.

PR #15524, opened 2019-10-21
[gradle] add multiple tasks option

att55:
### Checklist

### Motivation and Context

If user wants to execute multiple gradle tasks, Fastfile is like below. 

```
gradle(task: 'assembleDebug bundleDebug')
```

In this case, I think the key should be `tasks` and the value type should be `Array`.  (**Or gradle action should not accept multiple tasks.**)

So, I will add `tasks` parameter which type is Array to prevent confusing users. 

### Description

Which one of these two do you prefer? 

- Accept multiple gradle tasks. (The content in this PR)

- Disallow multiple gradle tasks.
If user want to execute multiple tasks, Fastfile is like below.

```
gradle(task: 'assembleDebug')
gradle(task: 'bundleDebug')
```
If disallowing multiple gradle tasks is better, I will change this PR content.


JhonJimenezCastro (approved):

janpio:
The currently behavior probably stems from how the `gradle` CLI command behaves. Why is this change _needed_?

att55:
@janpio Thank you for your review!

> The currently behavior probably stems from how the gradle CLI command behaves.

Yes, you are right. But I think `tasks option(type: array)` is needed to be **more understandable** when users pass multiple tasks to gradle action.

janpio:
No, I also like it - just have to make sure of the motivation behind it and then make sure the new option doesn't lead to confusion.

You moved the `required` check to the code, correcT?

att55:
> I also like it

Sorry, I misunderstood 🙏 

> You moved the required check to the code, correcT?

Yes! I moved the check which `task` is passed or not to the code.
That's because `task` was `optional`.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This looks great! Thanks for adding this 😊 Really appreciate that contribution ❤️

PR #15526, opened 2019-10-21
[gradle] print out task name instead of 'gradle' step

att55:
### Checklist

### Motivation and Context

To prevent printing out multiple 'gradle step' in summary, `step_name` option is prepared.

```
+------+-------------+-------------+
|         fastlane summary         |
+------+-------------+-------------+
| Step | Action      | Time (in s) |
+------+-------------+-------------+
| 1    | gradle      | 10          |
| 2    | gradle      | 8           |
| 3    | gradle      | 15          |
+------+-------------+-------------+
```

But, I think `task name` is used as `step_name` instead of 'gradle' in most cases.
In these case, the code in Fastfile is a little **redundant** as below... 💧 

```
gradle(task: 'assmebleDebug', step_name: 'assembleDebug')
gradle(task: 'bundleDebug', step_name: 'bundleDebug')
```

So, I think `task name` is better than 'gradle' to be printed out as default.

```
+------+----------------+-------------+
|         fastlane summary            |
+------+----------------+-------------+
| Step | Action         | Time (in s) |
+------+----------------+-------------+
| 1    | assmebleDebug  | 10          |
| 2    | bundleDebug    | 8           |
+------+----------------+-------------+
```

### Description

In case gradle action is called, `task name` is printed out instead of 'gradle step'.
If `step_name` is passed, `step_name` is printed out.

### DIFF
--- a/fastlane/lib/fastlane/actions/gradle.rb
+++ b/fastlane/lib/fastlane/actions/gradle.rb
@@ -103,6 +103,9 @@ module Fastlane
         # Give a helpful message in case there were no new apks or aabs. Remember we're only running this code when assembling, in which case we certainly expect there to be an apk or aab
         UI.message('Couldn\'t find any new signed apk files...') if new_apks.empty? && new_aabs.empty?
 
+        # Display a gradle task name in fastlane summary to make it easy to understand the content of this step
+        print_task_name_in_summary(params, gradle_task)
+
         return result
       end
       # rubocop:enable Metrics/PerceivedComplexity
@@ -115,6 +118,12 @@ module Fastlane
         end
 
         gradle_task
+      end
+
+      def self.print_task_name_in_summary(params, gradle_task)
+        if params[:print_task_name_in_summary]
+          Actions.executed_actions << { name: gradle_task }
+        end
       end
 
       #####################################################
@@ -192,6 +201,11 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :print_command_output,
                                        env_name: 'FL_GRADLE_PRINT_COMMAND_OUTPUT',
                                        description: 'Control whether the output produced by given Gradle command is printed while running (true/false)',
+                                       is_string: false,
+                                       default_value: true),
+          FastlaneCore::ConfigItem.new(key: :print_task_name_in_summary,
+                                       env_name: 'FL_GRADLE_PRINT_TASK_NAME_IN_SUMMARY',
+                                       description: 'Display the gradle task name instead of `gradle` in fastlane summary. If you want to display `gradle` or step_name which you set in summary, please set this option to `false`.(true/false)',
                                        is_string: false,
                                        default_value: true)
         ]
--- a/fastlane/lib/fastlane/actions/gradle.rb
+++ b/fastlane/lib/fastlane/actions/gradle.rb
@@ -103,7 +103,7 @@ module Fastlane
         # Give a helpful message in case there were no new apks or aabs. Remember we're only running this code when assembling, in which case we certainly expect there to be an apk or aab
         UI.message('Couldn\'t find any new signed apk files...') if new_apks.empty? && new_aabs.empty?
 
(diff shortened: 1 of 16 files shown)


joshdholtz (changes requested):
I actually _really_ like how this looks but forcing the runner to know about the gradle action is not something we'd like to do 🤔  The runner shouldn't have anything action specific. 

I think it would be cool if there was a way the **any** action could proide this 😇 I don't think know how easy this would be with the current implementation of actions but it would be nice to look into if possible 👌

att55:
@joshdholtz Thank you for your review!

> The runner shouldn't have anything action specific.

You are completely right, I think the runner should be global module too...
So, I will try to find another way, which the change effects only gradle action or any action could provide this (as you written).

att55:
I will add the title to `WIP` 🙏

att55:
@joshdholtz @janpio I added the parameter `print_task_name_in_summary` to gradle action not to effect the runner. How is this implementation? 

- before diff
https://github.com/fastlane/fastlane/pull/15526/commits/d318d1e0e318082ae79b34e7621a3969438ba3f7

att55:
@janpio Please remove `status: wip` label 🙏

att55:
@joshdholtz Please review when you can 🙏

janpio:
Rebase onto master should get rid of the last failing CI.

att55:
@janpio Got rid of the last failure 🙆‍♂

att55:
@joshdholtz Please review this PR when you can 🙏

joshdholtz:
Aye aye! Will review first thing tomorrow AM

att55:
@joshdholtz Thank you for your reply ! That would be fine 🙆‍♂

joshdholtz (commented):
@att55 This is great improvement! This actually inspired me to add a commit to make this a little more powerful and flexible 😊 Commit incoming in a bit! If you are able to review/test this that would be 💯

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
I have some commits in here but this looks 💯 to me 😊

fastlane/lib/fastlane/actions/gradle.rb
```diff
@@ -94,6 +94,20 @@ def self.run(params)
         return result
       end
 
+      def self.step_text(params)
```
snatchev: Is there any way we could the task name (and others) without changing the method signatures?

joshdholtz: Not without a rearchitecture 😬 Actions and their methods are execute as class methods and not as instance methods so there is no shared state that `step_text` can get 😢

snatchev: Understood.

snatchev:
I would recommend we do not introduce a new option for this, but instead we just adapt this as the new behavior for gradle actions.

I don't love the fact that this change forces the `step_text` on the other runners to change their signatures. An alternative would be to check the arity of `step_text` and not pass `params` if it's arity is 0. This seems a bit hacky and magical, but the alternative would require re-architecting these methods to not be class methods which also has some downsides. Any thoughts @att55 @joshdholtz ?

joshdholtz:
@snatchev Checking if it could take a param or not did cross my mind originally when reworking some of this but it seemed less buggy and more straight forward to just adjust those few actions that needed the change. But now that I think about it, that might be needed for any plugins that use `step_text` 🤔 So yeeaaaahhhh... might have to make that change to check for arity.

att55:
@snatchev Thank you for your review 😊
Sorry I'm not good at English, so please clear my head 🙏  

1. Your feedback is adding the code to check the arity of `step_text`, right?
2. Does this `step_text` means the method overrode in Gradle action?

> but the alternative would require re-architecting these methods to not be class methods which also has some downsides.

3. Could you give me more details about this comment 🙇

joshdholtz (approved):
I really love this feature! Thanks for doing this 😊

Issue #15445, opened 2019-10-05
[setup_ci] setup_keychain should be called with all CI providers


tlvenn:
### Problem:

I am running fastlane (2.131.0) with Github Actions and while the CI is properly detected, nothing is happening. When a different provider than `travis` or `circleci` is given, the action results in a noop.

### Expected:

If a CI is detected, setup_keychain should be called in all cases and only after, depending of the provider, more tasks could be executed with a `case/when statement`.

janpio:
Ping @mollyIV

svenmuennich:
I just opened PR #15542 that changes `setup_ci` as proposed above.

mollyIV:
Hello 👋 

We cannot possible provide a support for every CI provider out of the box. We have dedicated lanes for Travis (`setup_travis`), CircleCI (`setup_circle_ci`) and Jenkins (`setup_jenkins`).

> I am running fastlane (2.131.0) with Github Actions and while the CI is properly detected, nothing is happening.

@tlvenn could you please provide more details: 
* what fastlane lanes you are running (including `fastlane env` would be helpful)
* what is the expected result

> If a CI is detected, setup_keychain should be called in all cases

@tlvenn could you please explain why would we want that? 🤔 Are we sure that creating a temp keychain on some CI providers would not break stuff?

tlvenn:
Hi @mollyIV,
 
I am using match and my understanding and experience with it is that for it to work, you definitely need a keychain that simply does not exist at all in most CI.

Right now this action simply does nothing at all if you are not specifying `travis` or `circle_ci` as a provider. I understand it was factored out of the `setup_travis` & `setup_circle_ci` actions but there is no reason it could not be used without it.

> https://github.com/fastlane/fastlane/blob/d27dd65f975c067fa4e91df4d283e68c5ab2a949/fastlane/lib/fastlane/actions/setup_ci.rb#L10-L16

The whole purpose of this action is to setup a temp keychain and force match in readonly which is clearly communicated in the documentation so I believe the expectation are pretty clear.

If for some reason your CI is already providing a keychain, you should simply not use this action.

I really don't see any harm in making sure it calls `setup_keychain` in all cases. Am i missing something here ?

tlvenn:
And as far as I can tell, only `setup_travis` and `setup_circle_ci` are actually calling `setup_ci`
>https://github.com/fastlane/fastlane/search?q=setup_ci&unscoped_q=setup_ci

[...]

PR #15542, opened 2019-10-24
[action] setup_ci – always setup keychain when running on a CI

svenmuennich:
### Checklist

### Motivation and Context
Currently `setup_ci` must be passed a valid `:provider` (either `travis` or `circleci`) that controls which setup is performed for each of them. However, the only difference between the providers is that passing `circleci` also configures the output paths, while the temporary keychain is created for both of them.

As pointed out in #15445, running `setup_ci` in GitHub Actions without passing one of the valid providers currently results in a noop, even though the environment is recognized as being a CI. However, one would expect that the action also works on other CI providers and e.g. creates a temporary keychain.

### Description
Based on the @tlvenn's proposal put forward in #15445 I changed `setup_ci` to always create a temporary keychain, but only after performing any extra tasks necessary for a specific CI provider.

Furthermore the CI provider is now detected automatically, if no `:provider` is passed. I left the validation of a passed `:provider` in place to ensure backwards compatibility, even though passing `travis` does  not change the behavior of the action compared to not passing any `:provider`.

To not having to duplicate parts of the CI detection I added a new method `Helper.is_circle_ci`, that is used both in `Helper.ci?` and `setup_ci`.


svenmuennich:
@googlebot I signed it!

janpio:
> @googlebot I signed it!

That didn't work it seems. Usually this is because your GitHub account uses a different email address or the commits were made using a different email address.

svenmuennich:
> That didn't work it seems. Usually this is because your GitHub account uses a different email address or the commits were made using a different email address.

I think this is due to the company CLA we signed last week. The verification process has not yet completed.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -7,19 +7,22 @@ def self.run(params)
...
         end
+
+        setup_keychain
       end
 
       def self.should_run?(params)
         Helper.ci? || params[:force]
       end
 
+      def self.detect_provider(params)
```
tlvenn: I feel this is beyond the scope of the PR and I feel it's weird that circleci receive a special treatment there. Either we can flirt the CI and then we should do it or we should keep the current behavior.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -7,19 +7,22 @@ def self.run(params)
...
-        case params[:provider]
-        when 'travis'
-          setup_keychain
+        case detect_provider(params)
         when 'circleci'
-          setup_keychain
           setup_output_paths
         end
+
+        setup_keychain
```
tlvenn: `setup_keychain` was always called first, this is now called last which seems wrong to me even if probably it does not matter in the case of the additional `setup_output_paths` for Travis. However i believe there is a test which expect a given order and this is breaking.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -87,18 +90,20 @@ def self.available_options
                                        default_value: false),
           FastlaneCore::ConfigItem.new(key: :provider,
                                        env_name: "FL_SETUP_CI_PROVIDER",
-                                       description: "CI provider",
+                                       description: "CI provider. If none is set, the provider is detected automatically",
```
tlvenn: This is misleading as only circleci would be detected and imho as discussed above, this is outside of the scope of what this PR should do.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -108,7 +113,7 @@ def self.is_supported?(platform)
       def self.example_code
         [
           'setup_ci(
-            provider: "travis"
+            provider: "circleci"
```
tlvenn: Any reason to change the example ?

fastlane/lib/fastlane/actions/setup_ci.rb
tlvenn: Actually as you commented Message is asserted in reverse order so the test would pass but still I wonder if conceptually the action should not do first what it is supposed to do unless there is a very good reason to do something else first.

tlvenn (commented):
Thanks for the PR, I have mixed feeling with this one, I feel you are blending too much circleci stuff that could / should ? be in a separate PR.

fastlane/lib/fastlane/actions/setup_ci.rb
svenmuennich: Since the gist of this PR is that running `setup_ci` always installs the keychain, passing a `provider` param should be optional. That said, it might appear like _CircleCI_ receives special treatment but that is just because it is the only CI provider for which other setup step need to be performed. I didn't see the point in detecting every CI provider but then not acting on the result.

fastlane/lib/fastlane/actions/setup_ci.rb
svenmuennich: I actually just followed the suggestion made in your issue #15445:
> If a CI is detected, setup_keychain should be called in all cases **and only after** [...] **more tasks could be executed** with a case/when statement.

So I really don't mind moving this before the `case/when` statement.

fastlane/lib/fastlane/actions/setup_ci.rb
svenmuennich: Please see my explanation above.

fastlane/lib/fastlane/actions/setup_ci.rb
svenmuennich: I changed this to to prevent false assumptions that passing `travisci` changed the behaviour of this action, which it doesn't.

fastlane/lib/fastlane/actions/setup_ci.rb
tlvenn: I meant to say only after calling `setup_keychain`, other actions should be called with...
Sorry for the confusion, I see how this could be interpreted the other way around.

svenmuennich:
@googlebot I signed it!

joshdholtz (approved):
Thanks for making this change! I think these changes make a lot more sense. Appreciate the contribution ❤️

PR #15567, opened 2019-10-30
[action] Change `import_from_git` to support multiple files.

JoeMatt:
### Checklist

### Motivation and Context

I group my fastlane actions by objective. ie;

Fastfile+Carthage
Fastfile+Fabric
Fastfile+AppStore
Fastfile+Circle

To distribute clusters of actions to other teams, I use a stock `Fastfile` with some global settings, and array of the types of actions they'd like to import from our central repo.

The current `import_from_git` is slow to import multiple files compared to local import, since it shallow clones the target repo for each import statement.

Supporting arrays cuts my import time for 7 files from ~30s each `fastlane` call, to < 5 seconds.

### Description

I'm not a Ruby developer at all so my edits are amateur.

They consist of,

1. changing argument `path: String` to `paths: [String]`
2. calling the git checkout steps in a `each`

#### Issues

1. I didn't know what to do with the return statement. I'm sure ruby has a way to break out of an each block if the value is an error, but not something I know without help here.
2. Normally I'd make two versions of this type of function with overloading and wrap the original version of `path: String` to call this new version with `paths: [path]`. I'm used to Swift/C++, not sure the best practices in ruby for this kind of API change that can be resolved easily with an overload.

I can fix 1 and 2 if someone with more experience in best practices can point me to some clues, rather not waste a ton of time on something basic outside of my normal tech set, that the community can solve in seconds I'm sure. 

### Testing Steps

```ruby
# Have a repo somewhere with various fastlane files under `./Fastlane/Fastfile+{name}`

FASTLANE_IMPORT_LOCAL = false
IMPORTS = %w[Foo Bar]

  if FASTLANE_IMPORT_LOCAL
    IMPORTS.each do |import|
      import("Fastfile+#{import}")
    end
  else
    paths = IMPORTS.map {|import| "Fastlane/Fastfile+#{import}" }
    import_from_git(url: FASTLANE_GIT_URL, branch: FASTLANE_GIT_BRANCH, paths: paths, version: FASTLANE_GIT_VERSION)
  end
```

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -261,9 +261,9 @@ module Fastlane
 
     # @param url [String] The git URL to clone the repository from
     # @param branch [String] The branch to checkout in the repository
-    # @param path [String] The path to the Fastfile
+    # @param paths [String] The path to the Fastfile
     # @param version [String, Array] Version requirement for repo tags
-    def import_from_git(url: nil, branch: 'HEAD', path: 'fastlane/Fastfile', version: nil)
+    def import_from_git(url: nil, branch: 'HEAD', paths: ['fastlane/Fastfile'], version: nil)
       UI.user_error!("Please pass a path to the `import_from_git` action") if url.to_s.length == 0
 
       Actions.execute_action('import_from_git') do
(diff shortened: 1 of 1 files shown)


JoeMatt:
@googlebot I signed it!

janpio:
> I didn't know what to do with the return statement. I'm sure ruby has a way to break out of an each block if the value is an error, but not something I know without help here.

What exactly are you trying to achieve? What is the current return value and how do you want to change it?

> Normally I'd make two versions of this type of function with overloading and wrap the original version of `path: String` to call this new version with `paths: [path]`. I'm used to Swift/C++, not sure the best practices in ruby for this kind of API change that can be resolved easily with an overload.

That sounds like a good idea - best try it out if it works as expected.

joshdholtz:
Thanks for making these changes! We appreciate the time you spent on this ❤️  We are closing this in favor of #15804

PR #15655, opened 2019-11-14
[spaceship] fix live family edit

max-ott:
### Checklist

### Motivation and Context
It seems that there are currently multiple issues when it comes to updating live subscriptions on the App Store. 

If you try to update a `subscription family` for example, `spaceship` will fail with the following error:
> You must have at least one approved localization on this subscription.

Even though the user can simply go to App Store Connect and change the name or any other setting there. After some investigation I found that `spaceship` was missing an ID in the `details` object. Once added one can edit both live & draft subscription families.
 
### Testing Steps
```
bundle exec irb
require "spaceship"
Spaceship::Tunes.login("test@test.com", "•••••••••••••••••")
app = Spaceship::Tunes::Application.find(123456789)
edit = app.in_app_purchases.families.all.first.edit
edit.name = "renamed"
edit.save!
```

The same code above should fail on any existing `fastlane` version with the above mentioned error for any approved & live subscription families/groups.


max-ott:
spec.add_development_dependency('hashdiff', '< 0.4.0') # Locked at less than 0.4.0 until wemock bumped up to 3.6.0 or greater - HashDiff was changed to Hashdiff

joshdholtz:
@googlebot I consent.

joshdholtz:
@max-ott This should be good to go now! Lemme know if this is good to merge from your end 😊

joshdholtz (approved):
This is 🔥 Thanks again for the awesome spaceship work ❤️

max-ott:
@joshdholtz 👍 Looking forward to having this included in the next release.

Issue #15674, opened 2019-11-18
screengrab crashes after upgrade to 2.136.0 from 2.126.0


henningBunk:
### New Regression Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Regression Information
- Breaking version: `2.136.0`
- Last working version: `2.135.2`

### Regression Description
fastlane screengrab is crashing after update to 2.136.0. it happens on macos catalina and mojave. 2.135.2 worked fine.

`/Library/Ruby/Gems/2.3.0/gems/fastlane-2.136.0/fastlane_core/lib/fastlane_core/fastlane_pty.rb:24:in `spawn': [!] No such file or directory - fork failed (FastlaneCore::FastlanePtyError)
`

##### Complete output when running fastlane, including the stack trace and command used
<details>
  <pre>fastlane screengrab                        165ms  Mon Nov 18 17:08:20 2019
[✔] 🚀
[17:08:43]: fastlane detected a Gemfile in the current directory
[17:08:43]: however it seems like you don't use `bundle exec`
[17:08:43]: to launch fastlane faster, please use
[17:08:43]:
[17:08:43]: $ bundle exec fastlane screengrab
[17:08:43]:
[17:08:43]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[17:08:43]: Successfully loaded '/a/path/Screengrabfile' 📄

+-----------------------------+---------------------------------------------------------+
|                        Detected Values from './Screengrabfile'                        |
+-----------------------------+---------------------------------------------------------+
| launch_arguments            | ["isScreengrab true"]                                   |
| build_tools_version         | 27.0.1                                                  |
| reinstall_app               | false                                                   |
| app_apk_path                | a/path/an.apk                                           |
| tests_apk_path              | another/path/aTest.apk                                  |
| app_package_name            | my.company                                              |
| tests_package_name          | my.company.test                                         |
| use_tests_in_packages       | ["my.company.app.tests"]                                |
| test_instrumentation_runner | my.company.RxTestRunner                                 |
| output_directory            | fastlane/screenshots                                    |
| clear_previous_screenshots  | false                                                   |
[...]

PR #15684, opened 2019-11-20
[screengrab] Skip clean status bar if aapt not found

Hiroto-N:
### Checklist

### Motivation and Context

Addresses issue #15674 - Like the validate_apk method, enable_clean_status_bar should also check for aapt before trying to execute `aapt dump badging ...` so it doesn't fail, but just runs screengrab as it would have done in previous versions.

### Description

Just added an if check to verify that `@android_env.aapt_path` is not empty. Tested by running `bundle exec fastlane screengrab` on example screengrab project.


joshdholtz (approved):
Ah, thanks for adding this! Good catch 😊

henningBunk:
thanks for fixing this

Issue #15636, opened 2019-11-12
[supply] Add support for Playstore's Internal App Sharing


gneokleous:
#### Motivation Behind Feature
Allow distribution of APKs and AABs without manipulating version codes.

#### Feature Description
Google Developers API has added support for the recent feature `internal app sharing`. 

**Internal App Sharing**
Allows uploading an artifact without having to change version codes. The most important thing is that enables App Bundles to be distributed without rolling a unique artifact in any of the release tracks.

Explanation: https://support.google.com/googleplay/android-developer/answer/9303479?hl=en-GB

Google Developers API v3 Support Docs: https://developers.google.com/android-publisher/api-ref/internalappsharingartifacts.

There was another issue for this feature request: https://github.com/fastlane/fastlane/issues/14924 and closed due to inactivity. Also it requires changes that have been made as part of #15269.  

#### Alternatives or Workarounds
Current solution is manually uploading the binary files at:  [Uploading Link](https://play.google.com/apps/publish/internalappsharing/)

joshdholtz:
@gneokleous I like this! I'm currently working on some backlogged `screengrab` PRs and issues but I can look into adding this once I'm done with these other things 🙃

PaulWoitaschek:
That would be great! This would really simplify the CI->Testers workflow

ktae13-npnf:
Hi, I needed this feature, and made it work on my local machine. I am preparing for PR for the first time, but hoping I could create PR within couple days. It is going to use `Google API Client 0.34.1` for the sharing. Please let me know if it could be a trouble.

derrikammons:
Please create your PR and get it approved @ktae13-npnf.  We need this, too!

ktae13-npnf:
PR created!!!
https://github.com/fastlane/fastlane/pull/15696

Luckygirlllll:
@ktae13-npnf  I'm interested in this feature as well, what should I do to be able to use it in the fastlane? how exactly lane should look like?

ktae13-npnf:
@Luckygirlllll, as you know, it's in PR right now, and @joshdholtz will make changes for better structure. Let's wait a few more days until his change applies.

charlesdurham:
Just adding another "This is an awesome feature" comment 😄

PR #15696, opened 2019-11-22
Playstore internal app sharing

ktae13-npnf:
### Checklist

### Motivation and Context

There are requests to add functionality uploading android apk / aab to internal app sharing, and this change will allow users to upload apk and aab to internal app sharing by using `upload_apk_internal_app_sharing` and `upload_aab_internal_app_sharing` options.

https://github.com/fastlane/fastlane/issues/15636

### Description

1. Added boolean options `upload_apk_internal_app_sharing` and `upload_aab_internal_app_sharing`.
2. When either options are set to true, `skip_upload_apk` or `skip_upload_aab` will be ignored, and will only upload to internal app sharing.
3. Once uploading is done, it'll print out the link the build can be downloaded.
`download link : https://play.google.com/apps/test/XXXXXXXX/XXXXXXXXXXXXXXXX`

I have build my apps, and successfully upload/download apk/abb.

### Testing Steps

1. build the android app (either apk or aab)
2. if aab:
```upload_to_play_store(aab:'aab path', upload_aab_internal_app_sharing:true)```
if apk:
```upload_to_play_store(apk:'apk path, upload_apk_internal_app_sharing:true)```

### CODE REVIEW

**Comment by ktae13-npnf** [comment]
@googlebot I signed it!

File: supply/lib/supply/client.rb [hunk 1]
```diff
@@ -310,6 +310,20 @@ def upload_apk(path_to_apk)
       return result_upload.version_code
     end
 
+    def upload_apk_interanl_app_sharing(path_to_apk)
```
gneokleous: There is a typo here at `interanl`

supply/lib/supply/client.rb
```diff
@@ -340,6 +354,20 @@ def upload_bundle(path_to_aab)
       return result_upload.version_code
     end
 
+    def upload_bundle_interanl_app_sharing(path_to_aab)
```
gneokleous: Another typo, I am guessing copy-pasting error :)

supply/lib/supply/client.rb
ktae13-npnf: Fixed.

supply/lib/supply/client.rb
ktae13-npnf: Fixed. :)

joshdholtz (commented):
@ktae13-npnf Thanks for adding this! I was curious what your thoughts were on moving this from `supply` to its own own action? This looks like it doesn't leverage anything else that `supply` has so putting it into its own action would make `supply` less complicated and prone to regression issues in the future. I'm thinking we move it over to an action called `upload_to_playstore_internal_app_sharing`? I can do this quickly if you think that's okay 😊

ktae13-npnf:
@joshdholtz Thank you for reviewing, and suggestion. I don't have strong opinion about where it live. You can modify the code whatever it makes more sense to you. It is more important to me that more people needs this functionality, and hoping it would be in fastlane asap. :)

PR #15701, opened 2019-11-24
[supply] Support default changelog

janicduplessis:
### Checklist

### Motivation and Context
Currently on Android using supply it is needed to create a release notes file for every build. Most release end up including the same release notes so having a way to specify default notes if none are provided can save a lot of time and prevent release without notes.

### Description
This adds support for default release notes when a release note file for the current build does not exist.

It works by looking for a file named `default.txt` if a file with the build number does not exist.

### Testing Steps

Tested in my project by patching fastlane locally and adding a `default.txt` file with some text under the `changelogs` directory. Then I ran my deployment script and checked that the text included in `default.txt` was set as release notes in the google play console.

### DIFF
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -197,7 +197,7 @@ module Supply
       else
         default_changelog_path = File.join(Supply.config[:metadata_path], language, Supply::CHANGELOGS_FOLDER_NAME, "default.txt")
         if File.exist?(default_changelog_path)
-          UI.message("Updating default changelog for '#{version_code}' and language '#{language}'...")
+          UI.message("Updating changelog for '#{version_code}' and language '#{language}' to default changelog...")
           changelog_text = File.read(default_changelog_path, encoding: 'UTF-8')
         else
           UI.message("Could not find changelog for '#{version_code}' and language '#{language}' at path #{path}...")
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -195,7 +195,13 @@ module Supply
         UI.message("Updating changelog for '#{version_code}' and language '#{language}'...")
         changelog_text = File.read(path, encoding: 'UTF-8')
       else
-        UI.message("Could not find changelog for '#{version_code}' and language '#{language}' at path #{path}...")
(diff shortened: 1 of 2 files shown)


supply/lib/supply/uploader.rb
```diff
@@ -195,7 +195,13 @@ def upload_changelog(language, version_code)
         UI.message("Updating changelog for '#{version_code}' and language '#{language}'...")
         changelog_text = File.read(path, encoding: 'UTF-8')
       else
-        UI.message("Could not find changelog for '#{version_code}' and language '#{language}' at path #{path}...")
+        path = File.join(Supply.config[:metadata_path], language, Supply::CHANGELOGS_FOLDER_NAME, "default.txt")
+        if File.exist?(path)
+          UI.message("Updating default changelog for '#{version_code}' and language '#{language}'...")
```
janpio:
```suggestion
          UI.message("Updating changelog for '#{version_code}' and language '#{language}' to default changelog...")
```

supply/lib/supply/uploader.rb
```diff
@@ -195,7 +195,13 @@ def upload_changelog(language, version_code)
         UI.message("Updating changelog for '#{version_code}' and language '#{language}'...")
         changelog_text = File.read(path, encoding: 'UTF-8')
       else
-        UI.message("Could not find changelog for '#{version_code}' and language '#{language}' at path #{path}...")
+        path = File.join(Supply.config[:metadata_path], language, Supply::CHANGELOGS_FOLDER_NAME, "default.txt")
```
janpio: rename to `default_changelog_path` maybe

janpio:
Thanks, this looks great @janicduplessis and is a great idea. Left 2 comments with nitpicks.

supply/lib/supply/uploader.rb
janicduplessis: I wanted I slightly different message so you can know whether it used the default changelog or not, what do you think about keeping that?

janpio: Yeah, but what happens here is that the "changelog" is updated. In this new case it is "update to the default changelog text" - hence my suggestion.

janicduplessis: Oh my bad I didn’t notice the text added at the end of the line! Looks good.

joshdholtz (approved):
This is amazing! Thanks for taking the time to add this 😊 Really appreciate the contribution ❤️

Issue #15720, opened 2019-11-27
[Fastlane.swift] Type string_callback is mapped to Any? instead of (String) -> Void


jhandguy:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

In the generated Swift API, the callback arguments have a type of `Any?` instead of `(String) -> Any`. 

##### Command executed

```
sh(command: "exit 1", errorCallback: { println(message: $0) })
```

The callback closure does not get called.

PR #15721, opened 2019-11-27
[Fastlane.swift] Fix string_callback type override

jhandguy:
### Checklist

### Motivation and Context

Up until now, `:string_callback` was mapped to `Any?` instead of `(String) -> Void` when generating the Swift API.
Thus, there was no way to get a hook to callback functions when using Fastlane Swift, i.e. the `sh` action which takes an `error_callback` parameter.

Open issue: https://github.com/fastlane/fastlane/issues/15720

### Description

The issue lied in the way the `ConfigItem` `data_type` was built, returning nil when `data_type` is of kind `Symbol`.
Fixing the above issue, as well as adding the Swift `@escaping` solved the issue and I was able to test using:

```
sh(command: "exit 1", errorCallback: { fatalError($0) })
```

### Testing Steps

```
sh(command: "exit 1", errorCallback: { fatalError($0) })
```

### Known Issues

```
sh(command: "exit 1", errorCallback: { println(message: $0) })
```

Will make the runner get stuck.

### DIFF
--- a/fastlane/lib/fastlane/swift_fastlane_function.rb
+++ b/fastlane/lib/fastlane/swift_fastlane_function.rb
@@ -96,7 +96,7 @@ module Fastlane
       elsif type_override == Boolean
         return "Bool"
       elsif type_override == :string_callback
-        return "((String) -> Void)"
+        return "@escaping (String) -> Void"
       else
         return default_type
       end
@@ -104,7 +104,7 @@ module Fastlane
 
     def override_default_value_if_not_correct_type(param_name: nil, param_type: nil, default_value: nil)
       return "[]" if param_type == "[String]" && default_value == ""
-      return "{_ in }" if param_type == "((String) -> Void)"
+      return "{_ in }" if param_type == "@escaping (String) -> Void"
 
       return default_value
     end
@@ -119,7 +119,7 @@ module Fastlane
 
       optional_specifier = ""
       # if we are optional and don't have a default value, we'll need to use ?
-      optional_specifier = "?" if (optional && default_value.nil?) && type != "((String) -> Void)"
+      optional_specifier = "?" if (optional && default_value.nil?) && type != "@escaping (String) -> Void"
 
       # If we have a default value of true or false, we can infer it is a Bool
       if default_value.class == FalseClass
@@ -149,7 +149,7 @@ module Fastlane
             # we can't handle default values for Hashes, yet
             # see method swift_default_implementations for similar behavior
             default_value = "[:]"
-          elsif type != "Bool" && type != "[String]" && type != "Int" && type != "((String) -> Void)"
+          elsif type != "Bool" && type != "[String]" && type != "Int" && type != "@escaping (String) -> Void"
             default_value = "\"#{default_value}\""
           end
         end
--- a/fastlane/swift/RubyCommand.swift
+++ b/fastlane/swift/RubyCommand.swift
@@ -49,7 +49,7 @@ struct RubyCommand: RubyCommandable {
                     let typeJson: String
                     if let type = type {
                         typeJson = ", \"value_type\" : \"\(type.typeString)\""
-                    }else {
+                    } else {
                         typeJson = ""
                     }
 
--- a/fastlane_core/lib/fastlane_core/configuration/config_item.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/config_item.rb
@@ -255,9 +255,7 @@ module FastlaneCore
 
     # Determines the defined data type of this ConfigItem
     def data_type
-      if @data_type.kind_of?(Symbol)
-        nil
-      elsif @data_type
+      if @data_type
         @data_type
       else
         (@is_string ? String : nil)


jhandguy:
If you have any idea what might cause the SwiftRunner to get stuck when calling an action from the errorCallback like so:

```
sh(command: "exit 1", errorCallback: { println(message: $0) })
```

Please let me know, I'd be happy to fix it as well! 👍

taquitos:
No laptop for me to debug for this week, but a thought: if println is being calling and then the runner is stuck, I think the ruby fastlane process died and never sent a signal that it did. It’s been a while since I built it, but maybe the pipe broke and the runner doesn’t know it so the next command is just sitting there waiting for a response that will never come. Fatal error works because it kills the swift runner. So I think maybe sending a command over a broken connection needs work. Exit 1 might be a corner case.

jhandguy:
> No laptop for me to debug for this week, but a thought: if println is being calling and then the runner is stuck, I think the ruby fastlane process died and never sent a signal that it did. It’s been a while since I built it, but maybe the pipe broke and the runner doesn’t know it so the next command is just sitting there waiting for a response that will never come. Fatal error works because it kills the swift runner. So I think maybe sending a command over a broken connection needs work. Exit 1 might be a corner case.

Tried to do some more testing, here are the verbose logs:
```
DEBUG [2019-12-07 14:57:16.42]: received command:#<Fastlane::ActionCommand:0x00007fd3c2a01188 @method_name="sh", @class_name=nil, @command_id="", @args=[named argument: command, value: ls -2, type: , named argument: log, value: true, type: , named argument: error_callback, value: ignored_for_closure, type: string_closure]>
INFO [2019-12-07 14:57:16.42]: -------------------
INFO [2019-12-07 14:57:16.42]: --- Step: ls -2 ---
INFO [2019-12-07 14:57:16.42]: -------------------
INFO [2019-12-07 14:57:16.42]: $ ls -2
INFO [2019-12-07 14:57:16.43]: ▸ ls: illegal option -- 2
INFO [2019-12-07 14:57:16.43]: ▸ usage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]
ERROR [2019-12-07 14:57:16.43]: Exit status of command 'ls -2' was 1 instead of 0.
ls: illegal option -- 2
usage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]

DEBUG [2019-12-07 14:57:16.43]: sending {"payload":{"status":"ready_for_next","return_object":"ls: illegal option -- 2\nusage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]\n","closure_argument_value":"ls: illegal option -- 2\nusage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]\n"}}
DEBUG [2019-12-07 14:57:16.43]: received command:#<Fastlane::ActionCommand:0x00007fd3b455a3e0 @method_name="println", @class_name=nil, @command_id="", @args=[named argument: message, value: Hello World!, type: ]>
INFO [2019-12-07 14:57:16.43]: Hello World!
DEBUG [2019-12-07 14:57:16.43]: return_value_type is nil value:
DEBUG [2019-12-07 14:57:16.43]: sending {"payload":{"status":"ready_for_next","return_object":"","closure_argument_value":""}}
```

When executing action:
```
sh(command: "ls -2", errorCallback: { _ in
    println(message: "Hello World!")
})
```

So it seems the runner does get the response but just doesn't know when to exit.
Any idea where should I look into @taquitos?
I couldn't find a solution myself.

taquitos:
I'm running 2.137.0 and seeing some changes I don't recognize. It's bee a while, so maybe red herring, but last I remember, these error callbacks should be typed `errorCallback: ((String) -> Void) = { _ in },` in the `Fastlane.swift`. Instead, I'm seeing `errorCallback: Any? = nil` and that's causing me problems because the argument that is getting passed to the json parser is a `Function` instead of a value and that's blowing up.
```
[FastlaneRunner.RubyCommand.Argument(name: "command", value: Optional("sh"), type: nil), FastlaneRunner.RubyCommand.Argument(name: "log", value: Optional(true), type: nil), FastlaneRunner.RubyCommand.Argument(name: "error_callback", value: Optional((Function)), type: nil)]
```
Note the `value: Optional((Function))`

Which is triggering:
```
FastlaneRunner[75084:5434033] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid type in JSON write (__SwiftValue)'
```

taquitos:
That is to say, I think there is a problem with the `Fastlane.swift` function generator.

fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -96,15 +96,15 @@ def determine_type_from_override(type_override: nil, default_type: nil)
...
+        return "@escaping (String) -> Void"
       else
         return default_type
       end
     end
 
     def override_default_value_if_not_correct_type(param_name: nil, param_type: nil, default_value: nil)
       return "[]" if param_type == "[String]" && default_value == ""
-      return "{_ in }" if param_type == "((String) -> Void)"
+      return "{_ in }" if param_type == "@escaping (String) -> Void"
```
taquitos: I think I remember doing `((String) -> Void)` because later on, it might be made optional by adding `?` for `((String) -> Void)?`

So *_I think_* if you change this to `@escaping (String) -> Void` there might be a point where it turns into `@escaping (String) -> Void?` which isn't what you intend.

joshdholtz:
@taquitos Did you run the `fastlane generate_swift_api` lane before creating/updating your sample project? The `Fastlane.swift` file as the correct method signature for me 🙃 

```swift
/**
 Runs a shell command

 - parameters:
   - command: Shell command to be executed
   - log: Determines whether fastlane should print out the executed command itself and output of the executed command. If command line option --troubleshoot is used, then it overrides this option to true
   - errorCallback: A callback invoked with the command output if there is a non-zero exit status

 - returns: Outputs the string and executes it. When running in tests, it returns the actual command instead of executing it

 Allows running an arbitrary shell command.
 Be aware of a specific behavior of `sh` action with regard to the working directory. For details, refer to [Advanced](https://docs.fastlane.tools/advanced/#directory-behavior).
*/
@discardableResult func sh(command: String,
                           log: Bool = true,
                           errorCallback: @escaping (String) -> Void = {_ in }) -> String {
  let command = RubyCommand(commandID: "", methodName: "sh", className: nil, args: [RubyCommand.Argument(name: "command", value: command),
                                                                                    RubyCommand.Argument(name: "log", value: log),
                                                                                    RubyCommand.Argument(name: "error_callback", value: errorCallback, type: .stringClosure)])
  return runner.executeCommand(command)
}
```

---

## The issue

@taquitos @jhandguy The issue with the hanging starts happening here 👇 
https://github.com/fastlane/fastlane/blob/master/fastlane/swift/Runner.swift#L120

This block is being execute on the same thread (???) which is okay if just standard Swift things are being called but becomes weird when another socket call is made for another action. What ends up happening is that we hang in this function 👇 
https://github.com/fastlane/fastlane/blob/master/fastlane/swift/SocketClient.swift#L149-L159

It will eventually time out after the 1 hour default 👉  https://github.com/fastlane/fastlane/blob/master/fastlane/swift/SocketClient.swift#L36

## So what are we going to do?!

It seems like we need to do some magical changes that allow for nesting of action calls. I'm have no great ideas of the top of my head. I assume that I would probably start experimenting with some sort of operation queue or another blocking process to send off this nested command on another thread or something. @taquitos any big 💡 ideas?

jhandguy:
Wow, sounds rather complex @joshdholtz! 😮

My initial intention was to fix the bug with the `sh` action specifically, which does not fail when running using Fastlane Swift.
To try it out simply run:
```
sh(command: "exit 1")
```
and the exit status will still be 0 (success) instead of 1.
It probably has to do with the bridging and the Ruby action thinks there is an error callback while there isn't.

Maybe something to investigate first? 🤔

jhandguy:
A short update on this: I don't have enough time yet to look deeper into it, but meanwhile, I've made another PR for fixing the `sh` action never failing in Fastlane.swift (https://github.com/fastlane/fastlane/pull/15951). 

Feel free to take a look! 👍

joshdholtz:
This shouldn’t have been closed!

minuscorp:
After having debugged a lot...
```
[1591470738.879617]: [FastlaneRunner.RunnerArgument(name: "lane", value: "customLane")]
Enter
Enter
Enter
Leave
Leave
Leave
[1591470738.911611]: connected!
[1591470738.9116821]: Running lane: customLane
Enter
[1591470742.177505]: sending: {"commandType" : "action", "command" : {"commandID" : "","methodName" : "sh","args" : [{"value":"ls -2","name":"command"},{"name":"log","value":true},{"name" : "error_callback", "value" : "ignored_for_closure", "value_type" : "string_closure"}]}}
Enter
[1591470746.6402822]: response is: {"payload":{"status":"ready_for_next","return_object":"ls: illegal option -- 2\nusage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]\n","closure_argument_value":"ls: illegal option -- 2\nusage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]\n"}}
[1591470746.640304]: command executed
Performing callback with: ls: illegal option -- 2
usage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]

Enter
[1591470759.092904]: sending: {"commandType" : "action", "command" : {"commandID" : "","methodName" : "println","args" : [{"value":"Hello World!","name":"message"}]}}
Enter
```
Enter: 7
Leave:  3

This is a dispatch group unbalanced entering-exiting, even if they're waiting, there're unbalanced calls to fix here...

minuscorp:
@jhandguy I need you to sign with: @googlebot I consent. In #16577 16577

Issue #15676, opened 2019-11-18
[!] Your current version ($(MARKETING_VERSION)) does not respect the format A or A.B or A.B.C


SaundersB:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues) <-- It was closed automatically.

### Issue Description
Attempting to update XCode project version. I.E. A minor bump version fail such as 1.44 to 1.45.
I am able to successfully update the build number like build 345 to 346 but I cannot update the version number. There was a related issue that was closed due to inactivity to this problem.

##### Complete output when running fastlane, including the stack trace and command used
<details>
  <pre>$ fastlane bump
[✔] 🚀 
+----------------------------+---------+--------------------------------------------------------------------+
|                                               Used plugins                                                |
+----------------------------+---------+--------------------------------------------------------------------+
| Plugin                     | Version | Action                                                             |
+----------------------------+---------+--------------------------------------------------------------------+
| fastlane-plugin-versioning | 0.4.0   | get_version_number_from_git_branch                                 |
|                            |         | increment_build_number_in_xcodeproj get_version_number_from_plist  |
|                            |         | get_build_number_from_plist ci_build_number                        |
|                            |         | get_version_number_from_xcodeproj                                  |
|                            |         | increment_version_number_in_xcodeproj                              |
|                            |         | increment_version_number_in_plist get_app_store_version_number     |
|                            |         | get_info_plist_path get_build_number_from_xcodeproj                |
|                            |         | increment_build_number_in_plist                                    |
+----------------------------+---------+--------------------------------------------------------------------+

[13:51:15]: ------------------------------
[13:51:15]: --- Step: default_platform ---
[13:51:15]: ------------------------------
[13:51:15]: Driving the lane 'ios bump' 🚀
[13:51:15]: --------------------------------------
[13:51:15]: --- Step: increment_version_number ---
[13:51:15]: --------------------------------------
[13:51:15]: Before being able to increment and read the version number from your Xcode project, you first need to setup your project properly. Please follow the guide at https://developer.apple.com/library/content/qa/qa1827/_index.html
+------------------+----------+
|        Lane Context         |
+------------------+----------+
[...]

PR #15737, opened 2019-11-29
[action] increment_version_number: New default value for bump_type

adamstener:
### Checklist

### Motivation and Context
This change was requested because the default 'bump version' call would fail with a less than help error, if your current version did not include all three semantic fields (major minor and patch).  The default parameter assumes the user is trying to specifically bump the patch version, yet the error explains that the current version is not able to be incremented because it doesn't have the correct format of A or A.B or A.B.C

https://github.com/fastlane/fastlane/issues/15676

### Description
A new default parameter was added called 'bump'.  This 4th case (along with the existing major minor and patch) exists as the default way to bump a correctly formatted version, of the least significant digit. i.e. the default will now bump 1 -> 2 and 1.1 -> 1.2 and 1.1.1 -> 1.1.2

The tests that were added test the cases listed above.  Give that the method is already determining through a proven regex that the version number is in a valid format, each of the three types of semantic versions were tested to ensure the correct output.

### Testing Steps
bundle exec rspec ./fastlane/spec/actions_specs/increment_version_number_action_spec.rb

### DIFF
--- a/fastlane/lib/fastlane/actions/increment_version_number.rb
+++ b/fastlane/lib/fastlane/actions/increment_version_number.rb
@@ -43,6 +43,9 @@ module Fastlane
           version_array = current_version.split(".").map(&:to_i)
 
           case params[:bump_type]
+          when "bump"
+            version_array[-1] = version_array[-1] + 1
+            next_version_number = version_array.join(".")
           when "patch"
             UI.user_error!(version_token_error) if version_array.count < 3
             version_array[2] = version_array[2] + 1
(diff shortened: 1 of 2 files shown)


adamstener:
@googlebot I signed it

adamstener:
@googlebot I fixed it.

adamstener:
@googlebot I fixed it.

adamstener:
@googlebot I fixed it

janpio:
Hm, interesting! Does this change the current default behaviour for users of this action?

adamstener:
That depends.  There are 4 different scenarios to consider:
1) The user doesn't have a valid version string, and this does not change the response of that error
2) The user has a version that matches a valid sem-ver patch i.e. 1.2.3, in which case the previous and new functionality have the same result.

The next two cases are what would be changing.
Currently, If the user has 1 or 1.4 for their version, then they currently are met with an unhelpful error that their version doesn't match A or A.B or A.B.C

With the new functionality:
3) User has 1.2 -> this action bumps the version to 1.3 as expected per the repo issue that was filed
4) This also now would correctly work with a version of just 1 -> 2

So the default is a bump to whatever the least significant digit is.  Which technically means that this doesn't change any current behavior because of how it works with patch in the same way as before, it just increases the automatic capability of bumping any valid sem-ver the user has, which seems to naturally fit the idea of a default "version_increment".

joshdholtz (approved):
@adamstener This is 🔥 I actually really love this implementation and your explanation in your description and comment. Really appreciate the contribution ❤️

Issue #15742, opened 2019-11-30
[Scan] - Add support for disabling "KeyboardContinuousPathEnabled"


AlexTrott:
### Feature Request
Add a new config option for Scan which allows for test to run with "Slide to type" disabled.

#### Motivation Behind Feature

With the release of iOS 13, newly recreated/reset simulators all have "Slide to type" enabled by default.
This new functionality now causes an issue when executing XCUITest via `scan` which interact with textfields. As they will see the following screen when interacting with the UITextField for the first time.

<img width="559" alt="Screen Shot 2019-11-30 at 4 42 32 PM" src="https://user-images.githubusercontent.com/595623/69903404-853ff100-1390-11ea-96d8-0f9b546d7627.png">

This will then cause the XCUITest to fail, due to the inability to interact with the keyboard.

#### Feature Description

Offer developers the ability to disable slide to type when using scan, this would look something like this in the fastfile
```
  scan(
    reset_simulator: true,
    disable_slide_to_type: true
  )
```

Which I think would then require an update in scan's runner.rb, which would be something like this (Copied reset_simulators)
```
      if Scan.config[:disable_slide_to_type]
        Scan.devices.each do |device|
          FastlaneCore::Simulator.disable_slide_to_type(udid: device.udid)
        end
      end
```

Which would then require a final change in fastlane_core/device_manager.rb which would do the following:

```
def disable_slide_to_type(uuid)
    # go to the simulator preferences 
    ## /Users/{USERNAME}/Library/Developer/CoreSimulator/Devices/{UUID}/data/Library/Preferences 
   
   # Create a new file called: com.apple.keyboard.ContinuousPath.plist

   # Add the following:
   ## Key: KeyboardContinuousPathEnabled
   ## Type: Boolean
   ## Value: NO
end
```

Example plist file:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>KeyboardContinuousPathEnabled</key>
	<false/>
</dict>
</plist>
```

Which will then execute the test on the newly refreshed simulator with slide to type disabled.

##### Workaround:
The only work around I know, is to reset the simulator before executing scan, and then finding the simulator UUID and manually doing this.

##### Plugins:
I have not found any plugins which do this.

janpio:
[...]

PR #15749, opened 2019-12-02
[scan] Add support for disabling "KeyboardContinuousPathEnabled"

AlexTrott:
### Checklist

### Motivation and Context

With the release of iOS 13, newly recreated/reset simulators all have "Slide to type" enabled by default.
This new functionality now causes an issue when executing XCUITest via `scan` which interact with textfields. As they will see the following screen when interacting with the UITextField for the first time.

Screen Shot 2019-11-30 at 4 42 32 PM

This will then cause the XCUITest to fail, due to the inability to interact with the keyboard.

### Description

This change adds a new config option for scan,`SCAN_DISABLE_SLIDE_TO_TYPE` this is disabled by default, to revert back the keyboard behavior of a clean simulator.

**Resolves:** https://github.com/fastlane/fastlane/issues/15742

### Testing Steps

Testing in the following project: 
https://github.com/AlexTrott/Fastlane-iOS13-TextField-test

Test:
```
  desc "Test: Fastlane sets the default param, so the user has to do nothing"
  lane :test_default_param do
    scan(
      reset_simulator: true,
      disable_slide_to_type: true
    )
  end

  desc "Test: Fastlane responds when the user inputs the same selection as the default param"
  lane :test_setting_param do
    scan(
      reset_simulator: true,
      disable_slide_to_type: true
    )
  end

  desc "Test: Fastlane responds when the user sets the property to false"
  lane :test_which_will_fail do
    # This test will fail, as the 'Slide to Type' prompt is shown
    scan(
      reset_simulator: true,
      disable_slide_to_type: false
    )
  end
```


AlexTrott:
@googlebot I signed it!

jaleksynas (approved):
This looks good to me.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This is 🔥 And thank you sooooo much for providing that sample project. That sped up my testing process so much 😊 I added one commit to add some conditional checking to make sure that this is an iOS simulator running 13.0 or greater and some tests. Thanks so much for this contribution ❤️

PR #15822, opened 2019-12-19
[action] add `upload_to_play_store_internal_app_sharing` action

### LINKED ISSUE #15636: [supply] Add support for Playstore's Internal App Sharing

andrewhavens:
### Checklist

### Motivation and Context

Adds support for uploading Android builds to the Play Store Internal App Sharing by introducing a new action called `upload_to_play_store_internal_app_sharing`. Fixes #15636. This PR also closes #15696.

### Description

This PR adds `upload_apk_to_internal_app_sharing` and `upload_bundle_to_internal_app_sharing` methods to the `Supply::Client` class, which are called from a new method `perform_upload_to_internal_app_sharing` within the `Supply::Uploader` class. The return value is the download URL of the uploaded APK/AAB (or array of URLs if multiple were uploaded).

This PR also bumps the minimum required `google-api-client` version in order to add support for uploading to Internal App Sharing.

I have compared the changes from the previous maximum version (0.23.9) to the new minimum version (0.29.2) and can confirm that they are only minor/insignificant changes. https://github.com/googleapis/google-api-ruby-client/compare/0.23.9...0.29.2 I have also compared to the latest release (0.36.1) and have confirmed the same. https://github.com/googleapis/google-api-ruby-client/compare/0.29.2...0.36.1 So I have set the maximum version to `< 0.37.0`.

### Testing Steps

I have manually tested that this code works by uploading my app to the Play Store Internal App Sharing.

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -310,6 +310,19 @@ module Supply
       return result_upload.version_code
     end
 
+    def upload_apk_to_internal_app_sharing(package_name, path_to_apk)
+      # NOTE: This Google API is a little different. It doesn't require an active edit.
+      result_upload = call_google_api do
+        client.uploadapk_internalappsharingartifact(
+          package_name,
+          upload_source: path_to_apk,
+          content_type: "application/octet-stream"
+        )
+      end
+
+      return result_upload.download_url
+    end
+
     def upload_mapping(path_to_mapping, apk_version_code)
       ensure_active_edit!
 
@@ -338,6 +351,19 @@ module Supply
       end
 
       return result_upload.version_code
+    end
+
+    def upload_bundle_to_internal_app_sharing(package_name, path_to_aab)
+      # NOTE: This Google API is a little different. It doesn't require an active edit.
+      result_upload = call_google_api do
+        client.uploadbundle_internalappsharingartifact(
+          package_name,
+          upload_source: path_to_aab,
+          content_type: "application/octet-stream"
+        )
+      end
+
+      return result_upload.download_url
     end
 
     # Get a list of all tracks - returns the list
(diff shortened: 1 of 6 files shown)


andrewhavens:
@googlebot I signed it!

supply/lib/supply/client.rb
```diff
@@ -340,6 +353,19 @@ def upload_bundle(path_to_aab)
       return result_upload.version_code
     end
 
+    def upload_bundle_to_internal_app_sharing(package_name, path_to_aab)
+      # NOTE: This Google API is a little different. It doesn't require an active edit.
+      result_upload = call_google_api do
+        client.uploadbundle_internalappsharingartifact(
+          package_name,
+          upload_source: path_to_apk,
```
charlesdurham: should this be `path_to_aab`?

andrewhavens: Good catch. Sorry about that. Just pushed a fix for that.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for moving this over to an action! 🙏 I added some commits to reuse the options that supply uses so that we don't have duplicate options but other than this this if 🔥 Appreciate the contribution! ❤️

PR #15836, opened 2019-12-27
[GitHub Actions] enable automatic bump of homebrew formula with github action

joshdholtz:
### Motivation and Context
Want to keep automating all of the release things

### Description
Bumps the `url` and `sha` from the tag that was just made

### Testing Steps
Hopefully this works after next release 🤞

### DIFF
--- a/.github/workflows/tags.yml
+++ b/.github/workflows/tags.yml
@@ -1,0 +1,15 @@
+on:
+  push:
+    tags: 'v*'
+
+jobs:
+  homebrew:
+    name: Bump Homebrew formula
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tags.yml
```diff
@@ -0,0 +1,15 @@
+on:
+  push:
+    tags: 'v*'
```
mollyIV: If we'd like to bump it on every release, we could leverage `release` trigger:

```yaml
on: 
  release:
    types: [published]
```

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#release-event-release

mollyIV (approved):

PR #15838, opened 2019-12-27
[gym] using the correct build command when `skip_archive` is used

### LINKED ISSUE #11971: build_ios_app always packages the app, but doesn't support *building* only
### New Issue Checklist

### Description

I would have expected that I could just build an iOS app via `build_ios_app` as the name suggests. But that's just the name for gym, and gym is always attempting to archive and package the scheme 💪 -  even when the option `skip_archive` is set to true and `export_method` is set to `development`. Given the new name, I would have expected that I could just build an app as well, without archiving. Turns out that there is also the `xcodebuild` action, which can be used for this purpose. That has unfortunately a slightly different option syntax and doesn't support custom xcpretty formatters.

Skoti:
### Checklist

### Motivation and Context

Just wanted to PR a quick fix for `gym` command.

Fixes #11971

When running `gym --skip-archive` the generated `xcodebuild` command does not specify any action/"build command" to use because `archive` action is not added (as expected).

The issue is that the `xcodebuild` still works because when no action is specified it uses `build` action by default. Unfortunately, this breaks when `clean` option is added: `gym --skip-archive --clean` as the generated command will now contain `clean` action, and because this is the only action specified the generated command only performs "clean" and no build - this is **unexpected**.

It also silently crashes with error `xattr: No such file: *.xcarchive` as there is no archive to mark at the next step:
https://github.com/fastlane/fastlane/blob/18c78326d0afec17256e98ed401441b8aa7f74fa/gym/lib/gym/runner.rb#L93-L100

This PR fixes both.

### DIFF
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -18,6 +18,9 @@ module Gym
         build_app
       end
       verify_archive unless Gym.config[:skip_archive]
+
+      return nil if Gym.config[:skip_archive]
+
       FileUtils.mkdir_p(File.expand_path(Gym.config[:output_directory]))
 
       if Gym.project.ios? || Gym.project.tvos?
@@ -97,9 +100,11 @@ module Gym
                                                 ErrorHandler.handle_build_error(output)
                                               end)
 
-      mark_archive_as_built_by_gym(BuildCommandGenerator.archive_path)
-      UI.success("Successfully stored the archive. You can find it in the Xcode Organizer.") unless Gym.config[:archive_path].nil?
-      UI.verbose("Stored the archive in: " + BuildCommandGenerator.archive_path)
+      unless Gym.config[:skip_archive]
+        mark_archive_as_built_by_gym(BuildCommandGenerator.archive_path)
+        UI.success("Successfully stored the archive. You can find it in the Xcode Organizer.") unless Gym.config[:archive_path].nil?
+        UI.verbose("Stored the archive in: " + BuildCommandGenerator.archive_path)
+      end
 
       post_build_app
     end
--- a/gym/lib/gym/generators/build_command_generator.rb
+++ b/gym/lib/gym/generators/build_command_generator.rb
@@ -51,6 +51,7 @@ module Gym
 
         buildactions = []
         buildactions << :clean if config[:clean]
+        buildactions << :build if config[:skip_archive]
         buildactions << :archive unless config[:skip_archive]
 
         buildactions
--- a/gym/spec/build_command_generator_spec.rb
+++ b/gym/spec/build_command_generator_spec.rb
@@ -88,6 +88,25 @@ describe Gym do
                            ])
     end
 
+    it "uses the correct build command when `skip_archive` is used", requires_xcodebuild: true do
+      log_path = File.expand_path("#{FastlaneCore::Helper.buildlog_path}/gym/ExampleProductName-Example.log")
+
+      options = { project: "./gym/examples/standard/Example.xcodeproj", scheme: 'Example', skip_archive: true }
+      Gym.config = FastlaneCore::Configuration.create(Gym::Options.available_options, options)
+
+      result = Gym::BuildCommandGenerator.generate
+      expect(result).to eq([
+                             "set -o pipefail &&",
+                             "xcodebuild",
+                             "-scheme Example",
+                             "-project ./gym/examples/standard/Example.xcodeproj",
+                             "-destination 'generic/platform=iOS'",
+                             :build,
+                             "| tee #{log_path.shellescape}",
+                             "| xcpretty"
+                           ])
+    end
+
     describe "Standard Example" do
       before do
         options = { project: "./gym/examples/standard/Example.xcodeproj", scheme: 'Example' }


joshdholtz:
@Skoti Thanks for fixing this! Are you also setting `skip_package_ipa` to `true` when testing this? I'm erring out in the `package_app` method in `Gym::Runner` and I just want to make sure I'm testing this the same way as you use it 😇

gym/lib/gym/runner.rb
```diff
@@ -97,7 +97,7 @@ def build_app
                                                 ErrorHandler.handle_build_error(output)
                                               end)
 
-      mark_archive_as_built_by_gym(BuildCommandGenerator.archive_path)
+      mark_archive_as_built_by_gym(BuildCommandGenerator.archive_path) unless Gym.config[:skip_archive]
       UI.success("Successfully stored the archive. You can find it in the Xcode Organizer.") unless Gym.config[:archive_path].nil?
       UI.verbose("Stored the archive in: " + BuildCommandGenerator.archive_path)
```
joshdholtz: We should probably wrap all these lines if we skip archive since there won't be anything in that path 😇 

```rb
if Gym.config[:skip_archive]  
  mark_archive_as_built_by_gym(BuildCommandGenerator.archive_path)
  UI.success("Successfully stored the archive. You can find it in the Xcode Organizer.") unless Gym.config[:archive_path].nil?
  UI.verbose("Stored the archive in: " + BuildCommandGenerator.archive_path)
end
```

joshdholtz (changes requested):
I think we also need to early return right after `build_app` if `skip_archive` is `true` 🤔 Otherwise gym will try to verify the archive and package the app which it can't because the archive doesn't exist

Besides those things this is a great find! Thanks so much for this and can't wait to get this merged 💪

gym/lib/gym/runner.rb
Skoti: Ok I will wrap these lines although with `if !Gym.config[:skip_archive]`

Skoti:
> I think we also need to early return right after build_app if skip_archive is true 🤔

I'm not so sure. You can still run gym with `skip_build_archive` to just package earlier prepared archive.
`skip_archive` is only for the xcodebuild build step.

joshdholtz:
@Skoti Nevermind 😝 You are right. I must have had something messed up. Don't need to early return at all 👌

gym/lib/gym/runner.rb
joshdholtz: @Skoti I think we still wants something like this ☝️ We don't want any archive path stuff printed out because it didn't make one, correct?

Skoti:
@joshdholtz  I think you were right. Look at the edited comment.
If someone is explicitly `skipping build_archive` then we allow to package as it is intentional, otherwise we ensure `xcodebuild` was executed with the `archive` action.
What do you think?

gym/lib/gym/runner.rb
Skoti: correct, that is why I think your example should use "!" in the if statement 😃

gym/lib/gym/runner.rb
```diff
@@ -97,9 +97,11 @@ def build_app
                                                 ErrorHandler.handle_build_error(output)
                                               end)
 
-      mark_archive_as_built_by_gym(BuildCommandGenerator.archive_path)
-      UI.success("Successfully stored the archive. You can find it in the Xcode Organizer.") unless Gym.config[:archive_path].nil?
-      UI.verbose("Stored the archive in: " + BuildCommandGenerator.archive_path)
+      if !Gym.config[:skip_archive]
```
joshdholtz: This looks good 👌

gym/lib/gym/runner.rb
```diff
@@ -22,7 +22,7 @@ def run
 
       if Gym.project.ios? || Gym.project.tvos?
         fix_generic_archive # See https://github.com/fastlane/fastlane/pull/4325
-        return BuildCommandGenerator.archive_path if Gym.config[:skip_package_ipa]
+        return BuildCommandGenerator.archive_path if Gym.config[:skip_package_ipa] || (!Gym.config[:skip_build_archive] && Gym.config[:skip_archive])
```
joshdholtz: I'm still a little bit unsure about this one 🤔 I think I'm still in favor of...

```rb
return nil if Gym.config[:skip_archive]

```

There isn't an archive path if we are skipping archive so `nil` should be returned. This should be the case fo both iOS and macOS.

I also think someone could specific `skip_build_archive: true` and `skip_archive: true` (for whatever reason) and use `gym`. This would call `verify_archive` which would be okay and then it return nil because it shouldn't go into any building of anything.

This is getting complicated 😇  Thoughts?

Skoti: Right it is simpler that way. Thanks!

joshdholtz: Not a problem! Thanks for making changes 😊 I gotta drive to a holiday party now but I will review this tomorrow and most likely get it merged as well 💪

joshdholtz (approved):
This is great! Thanks for making those changes 😇 Really appreciate the contribution ❤️

gym/lib/gym/runner.rb
```diff
@@ -18,6 +18,9 @@ def run
         build_app
       end
       verify_archive unless Gym.config[:skip_archive]
+
+      return nil if Gym.config[:skip_archive]
```
pyrtsa: FWIW, this single line here seemed to break our use of
```ruby
build_ios_app(skip_build_archive: true, skip_archive: true, export_options: {method: "ad-hoc"}, ...)
```
for exporting a freshly built and archived build under a different provisioning mode (e.g. one to be uploaded to TestFlight and another as Ad Hoc to an internal distribution system).

I don't know if it was ever intended of `gym` to support that use case, so maybe it's not a bug but I should request (or contribute) it as a new action instead. (Curiously, the existing `resign` action doesn't seem to support our use case AFAICT.)

For now, we can just use `xcodebuild -exportArchive -archivePath <xcarchive> -exportOptionsPlist <plist> -exportPath <dir>` with `sh`, but would be nice if it were supported better by Fastlane out of the box.

PR #15898, opened 2020-01-14
[match] update an instruction in match/storage/google_cloud_storage.rb

puls:
This wording would have saved me a couple of hours of headaches this morning.

### Checklist

### Motivation and Context

This is just a small change to an instruction in the UI. If this string had been different, I could have saved myself a few hours this morning.

### Description

This is just a small change to an instruction in the UI.

### DIFF
--- a/match/lib/match/storage/google_cloud_storage.rb
+++ b/match/lib/match/storage/google_cloud_storage.rb
@@ -349,7 +349,7 @@ module Match
           UI.message("\t\t 1. Click on your bucket to open it".cyan)
           UI.message("\t\t 2. Click 'Permissions'".cyan)
           UI.message("\t\t 3. Click 'Add members'".cyan)
-          UI.message("\t\t 4. Enter the email of your service account".cyan)
+          UI.message("\t\t 4. Enter the *email of your service account* (not your own email)".cyan)
           UI.message("\t\t 5. Set the role to 'Storage Admin'".cyan)
           UI.message("\t\t 6. Click 'Save'".cyan)
           UI.message("")
--- a/match/lib/match/storage/google_cloud_storage.rb
+++ b/match/lib/match/storage/google_cloud_storage.rb
@@ -181,7 +181,7 @@ module Match
       end
 
       def generate_matchfile_content
-        return "bucket_name(\"#{self.bucket_name}\")"
+        return "google_cloud_bucket_name(\"#{self.bucket_name}\")"
       end
 
       private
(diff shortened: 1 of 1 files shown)


puls:
@googlebot I signed it!

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -349,7 +349,7 @@ def check_bucket_permissions
           UI.message("\t\t 1. Click on your bucket to open it".cyan)
           UI.message("\t\t 2. Click 'Permissions'".cyan)
           UI.message("\t\t 3. Click 'Add members'".cyan)
-          UI.message("\t\t 4. Enter the email of your service account".cyan)
+          UI.message("\t\t 4. Enter the *email of your service account* (not your own email)".cyan)
```
janpio: Does `*` work to highlight something in `UI.message`?

janpio: Isn't it obvious that service account != your own email?

match/lib/match/storage/google_cloud_storage.rb
```diff
@@ -181,7 +181,7 @@ def skip_docs
       end
 
       def generate_matchfile_content
-        return "bucket_name(\"#{self.bucket_name}\")"
+        return "google_cloud_bucket_name(\"#{self.bucket_name}\")"
```
joshdholtz: Ah shoot! Was this really wrong? 😱

match/lib/match/storage/google_cloud_storage.rb
joshdholtz: I think we need to remove the `*` here as they don't bold anything and nothing like this is used anywhere else in fastlane. If anything would suggest we make this change 👇 
```suggestion
          UI.message("\t\t 4. Enter the service account email address".cyan)
```

joshdholtz (changes requested):
Was that `generate_matchfile_content` really wrong?! And left a comment on the email message 😊

match/lib/match/storage/google_cloud_storage.rb
puls: It would be worth double-checking but I had to change my `Matchfile` to this before it would work.

match/lib/match/storage/google_cloud_storage.rb
puls: I mean, I spent hours trying to figure out what I was doing wrong but maybe I'm just dumb.

match/lib/match/storage/google_cloud_storage.rb
janpio: That this didn't pop up earlier also confused me, but checking the documentation it is indeed `google_cloud_bucket_name`.

joshdholtz (approved):
This for catching this! Really appreciate the contribution ❤️

PR #15906, opened 2020-01-16
[ensure_xcode_version] Implement flexible version check

jhandguy:
### Checklist

### Motivation and Context

Sometimes we want to ignore the `patch` version number (i.e. Bitrise does not keep track of it).
Currently, if we check for Xcode version `11.3` while having the version `11.3.1` installed, the action will fail.

### Description

These changes add more flexibility: 
Instead of matching the whole string `11.3.1`, we can simply split it to an array based on the separator `.` and only care about the version numbers actually specified.

For instance:
Checking for the Xcode version `11.3` while having `11.3.1` will match `11` with `11` and `3` with `3` but discard the last patch version number `1`.

On the other hand, when specifying `11.3.1` while having `11.3` will mismatch with `1` and `nil`.

### DIFF
--- a/fastlane/lib/fastlane/actions/ensure_xcode_version.rb
+++ b/fastlane/lib/fastlane/actions/ensure_xcode_version.rb
@@ -4,6 +4,7 @@ module Fastlane
       def self.run(params)
         Actions.verify_gem!('xcode-install')
         required_version = params[:version]
+        strict = params[:strict]
 
         if required_version.to_s.length == 0
           # The user didn't provide an Xcode version, let's see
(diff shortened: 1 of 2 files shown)


janpio:
Hm, doesn't this kind of change the existing behavior? (Although, does this existing behavior actually make sense?)

jhandguy:
Hi @janpio!

I would say not really, if you provide the full set of version numbers (i.e. 13.3.1) it still is as strict as it is currently.
The only situation that will make it different from the current implementation, is that if you started off just checking for let’s say 13.3, then the day you upgrade to 13.3.1, the check will still pass.
But thinking about it, a patch version upgrade is usually very low risk, so isn’t it better to actually be more tolerant especially with the patch version? And if one wants to still enforce the full version 13.3.1, he can still do so and nothing will change.

In my situation, I saw it can be very annoying to have this check fail on a patch upgrade, which happen a lot lately on Xcode.

Now, if we decide to still allow the stricter way, I can add a flag, I guess that’s no issue, but I honestly think being flexible might just be better fitting with real world case scenarios.

joshdholtz (changes requested):
This is a breaking change and I feel like there are cases where a specific patch version may be required if there is a bug or something in a previous version. Can we make this new behavior as a new option? 😊

jhandguy:
Done @joshdholtz ! ✅ 
Added a `strict` flag with a default value `true` 👍

jhandguy:
Any progress on this @joshdholtz? 😃 
Let me know if you need further testing from my end! 👍

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for adding this! Really like this 😊 I added some tests onto this too. Really appreciate the contribution ❤️

jhandguy:
Thank you for this @joshdholtz, will think about adding more tests as well next time! 👌

PR #15914, opened 2020-01-19
[deliver] docs: document conflicting options

### LINKED ISSUE #817: The parameter of auto_release_date cannot be used with automatic_release, Not documented
In all of the references to `auto_release_date` there are no references to the fact that it cannot be used along with `automatic_release`. The only reference I could find to this restriction is in the deliver options.rb file.

It would be helpful to call this out in the documentation.

https://github.com/fastlane/fastlane/blob/049fe3dd417955c28c4365640c89c0db565ac663/deliver/lib/deliver/options.rb#L162

janpio:
closes https://github.com/fastlane/docs/issues/817

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -149,12 +149,12 @@ module Deliver
 
         # release
         FastlaneCore::ConfigItem.new(key: :automatic_release,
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Looks good to me! ❤️

PR #15918, opened 2020-01-21
[match] add mac and catalyst support for match

joshdholtz:
## Motivation and Context
Everybody should be able to `match`

### Cert
- New `: mac_installer_distribution` option
  - Tells `cert` to create a new "3rd Party Mac Developer Installer" certificate instead of a signing certificate
  - Defaults to `false`

### Match
- New `: mac_installer_distribution` option
  - Tells `match` to additionally create/store/fetch a  "3rd Party Mac Developer Installer" certificate

repo_installer

### fastlane_core
- Updated logic in `FastlaneCore::CertChecker` to check if "3rd Party Mac Developer Installer" certs are/were installed

### Spaceship
- Added a little bit of a hack to find "Mac" apps that are enabled by Catalyst
  - Catalyst apps are 100% iOS apps so Catalyst apps won't appear when filtering for Catalyst apps
  - If looking for a Mac app for a specific bundle id, spaceship will first look for Mac and then look for iOS if its not found (😔 but not really a better way at the moment)


joshdholtz (commented):
REQUESTING CHANGES

match/lib/match/generator.rb
```diff
@@ -3,14 +3,19 @@
...
   # Generate missing resources
   class Generator
-    def self.generate_certificate(params, cert_type, working_directory)
+    def self.generate_certificate(params, cert_type, working_directory, specific_cert_type: nil)
       require 'cert/runner'
       require 'cert/options'
 
       output_path = File.join(working_directory, "certs", cert_type.to_s)
 
+      if cert_type.to_sym == :developer_id_application
```
joshdholtz: Comment on why we need to do this

match/lib/match/generator.rb
```diff
@@ -70,8 +75,14 @@ def self.generate_provisioning_profile(params: nil, prov_type: nil, certificate_
       }
 
       values[:platform] = params[:platform]
-      values[:adhoc] = true if prov_type == :adhoc
-      values[:development] = true if prov_type == :development
+
+      if params[:type] == "developer_id"
```
joshdholtz: Comment on why we need to do this

match/lib/match/nuke.rb
```diff
@@ -112,17 +113,34 @@ def prepare_list
...
 
-      self.certs = certificate_type(cert_type).flat_map(&:all)
       self.profiles = []
       prov_types.each do |prov_type|
-        self.profiles += profile_type(prov_type).all
+        self.profiles += profile_type(prov_type).all(mac: false)
+        self.profiles += profile_type(prov_type).all(mac: true)
+      end
+
+      cert_types += (params[:additional_cert_types] || []).each do |ct|
```
joshdholtz: Explain the logics here in a comment

match/lib/match/options.rb
```diff
@@ -19,18 +19,26 @@ def self.available_options
...
                                      short_option: "-y",
                                      default_value: 'development',
                                      verify_block: proc do |value|
                                        unless Match.environments.include?(value)
                                          UI.user_error!("Unsupported environment #{value}, must be in #{Match.environments.join(', ')}")
                                        end
                                      end),
+        FastlaneCore::ConfigItem.new(key: :additional_cert_types,
+                                     env_name: "CERT_TYPE",
+                                     description: "Create specific certificate type (takes precedence over :development)",
```
joshdholtz: Wrong description

match/lib/match/spaceship_ensure.rb
```diff
@@ -58,11 +58,18 @@ def certificate_exists(username: nil, certificate_id: nil)
...
 
-    def profile_exists(username: nil, uuid: nil)
-      found = Spaceship.provisioning_profile.all.find do |profile|
+    def profile_exists(username: nil, uuid: nil, platform: nil)
+      is_mac = platform == "macos"
+      found = Spaceship.provisioning_profile.all(mac: is_mac).find do |profile|
         profile.uuid == uuid
       end
 
+      if !found && is_mac
```
joshdholtz: Maybe add more comment

spaceship/lib/spaceship/portal/app.rb
```diff
@@ -108,9 +108,18 @@ def create!(bundle_id: nil, name: nil, mac: false, enable_services: {})
         # @return (App) The app you're looking for. This is nil if the app can't be found.
         def find(bundle_id, mac: false)
           raise "`bundle_id` parameter must not be nil" if bundle_id.nil?
-          all(mac: mac).find do |app|
-            return app if app.bundle_id.casecmp(bundle_id) == 0
+          found_app = all(mac: mac).find do |app|
+            app if app.bundle_id.casecmp(bundle_id) == 0
           end
+
+          # Find catalyst enabled mac apps
```
joshdholtz: Maybe add more comment

joshdholtz (commented):
More self review!

cert/lib/cert/runner.rb
```diff
@@ -134,6 +134,19 @@ def certificates
 
     # The kind of certificate we're interested in
     def certificate_type
+      if Cert.config[:type]
```
snatchev: Is it possible for `Cert.config[:type]` to be nil?

match/lib/match/nuke.rb
```diff
@@ -112,17 +113,39 @@ def prepare_list
...
       prov_types.each do |prov_type|
-        self.profiles += profile_type(prov_type).all
+        self.profiles += profile_type(prov_type).all(mac: false)
+        self.profiles += profile_type(prov_type).all(mac: true)
       end
 
-      certs = Dir[File.join(self.storage.working_directory, "**", cert_type.to_s, "*.cer")]
-      keys = Dir[File.join(self.storage.working_directory, "**", cert_type.to_s, "*.p12")]
+      # Gets the main and additional cert types
+      cert_types += (params[:additional_cert_types] || []).each do |ct|
```
snatchev: should this be a map? What happens to the the value returned by `Match.cert_type_sym(ct)`

match/lib/match/nuke.rb
```diff
@@ -240,6 +263,11 @@ def delete_files!
 
     # The kind of certificate we're interested in
     def certificate_type(type)
+      case type.to_sym
+      when :mac_installer_distribution
+        return [Spaceship.certificate.mac_installer_distribution]
+      end
```
snatchev: this method should be just a regular o' `case` instead of this hash-lookup-with-raise.
```ruby
case type.to_sym
when :mac_installer_distribution
  [Spaceship.certificate.mac_installer_distribution]
when :distribution
  [Spaceship.certificate.production, Spaceship.certificate.apple_distribution]
...
else
  raise "Unknown certificate type `#{type}`
end
```

match/lib/match/nuke.rb
```diff
@@ -250,10 +278,11 @@ def certificate_type(type)
...
-        appstore:    Spaceship.provisioning_profile.app_store,
-        development: Spaceship.provisioning_profile.development,
-        enterprise:  Spaceship.provisioning_profile.in_house,
-        adhoc:       Spaceship.provisioning_profile.ad_hoc
+        appstore:     Spaceship.provisioning_profile.app_store,
+        development:  Spaceship.provisioning_profile.development,
+        enterprise:   Spaceship.provisioning_profile.in_house,
+        adhoc:        Spaceship.provisioning_profile.ad_hoc,
+        developer_id: Spaceship.provisioning_profile.direct
       }[prov_type] ||= raise "Unknown provisioning type '#{prov_type}'"
```
snatchev: this should be a `case` please 🙇

match/lib/match/spaceship_ensure.rb
```diff
@@ -45,8 +45,8 @@ def bundle_identifier_exists(username: nil, app_identifier: nil)
       UI.user_error!("Couldn't find bundle identifier '#{app_identifier}' for the user '#{username}'")
     end
 
-    def certificate_exists(username: nil, certificate_id: nil)
-      found = Spaceship.certificate.all.find do |cert|
+    def certificate_exists(username: nil, certificate_id: nil, platform: nil)
+      found = Spaceship.certificate.all(mac: platform == "macos").find do |cert|
```
snatchev: There is a Certificate.find https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/portal/certificate.rb#L290
But that still calls .all.find.

We should accept a `certificate_ids` list and check them all in 1 pass instead of making N requests containing the same list every time.

snatchev (changes requested):

cert/lib/cert/runner.rb
joshdholtz: Yup yup, its optional and takes precedence over the `:development` option which is why we check for it first

match/lib/match/nuke.rb
joshdholtz: OMG yes

match/lib/match/nuke.rb
joshdholtz: Smart 💡

match/lib/match/runner.rb
```diff
@@ -79,13 +80,23 @@ def run(params)
...
       cert_id = fetch_certificate(params: params, working_directory: storage.working_directory)
-      spaceship.certificate_exists(username: params[:username], certificate_id: cert_id) if spaceship
+      # spaceship.certificate_exists(username: params[:username], certificate_id: cert_id, platform: params[:platform]) if spaceship
+
+      # Mac Installer Distribution Certificate
+      additional_cert_types = params[:additional_cert_types] || []
+      cert_ids = additional_cert_types.map do |additional_cert_type|
+        fetch_certificate(params: params, working_directory: storage.working_directory, specific_cert_type: additional_cert_type)
+        # spaceship.certificate_exists(username: params[:username], certificate_id: installer_cert_id, platform: params[:platform]) if spaceship
+      end || []
```
snatchev: optional: `|| []` seems redundant since `additional_cert_types` is already `|| []`'d

match/lib/match/runner.rb
```diff
@@ -79,13 +80,23 @@ def run(params)
...
+
+      cert_ids << cert_id
+      spaceship.certificates_exists(username: params[:username], certificate_ids: cert_ids, platform: params[:platform]) if spaceship
```
snatchev: the `if spaceship` makes me nervous as it appears to be disabling a check without a warning and is spaceship expected to be nil? If it's nil, is that an intentional code-path?

I think it would prefer something like:
```ruby
if spaceship
  spaceship.certificates_exists
else
  UI.warning("cannot verify existence of certificate ids #{cert_ids}, because we do not have a Spaceship client.")
end
```

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -47,12 +48,48 @@ def self.installed_identies(in_keychain: nil)
...
+      ids = []
+      available.split("\n").each do |current|
+        begin
+          (ids << current.match(/SHA-1 hash: ([[:xdigit:]]*)/)[1])
+        rescue
+          # the last line does not match
+        end
+      end
+
+      return ids
```
snatchev: this can be reduced to:
```ruby
return available.scan(/^SHA-1 hash: ([[:xdigit:]]+)$/).flatten
```

match/lib/match/runner.rb
```diff
@@ -79,13 +80,23 @@ def run(params)
...
         app_identifiers.each do |app_identifier|
-          spaceship.bundle_identifier_exists(username: params[:username], app_identifier: app_identifier)
+          spaceship.bundle_identifier_exists(username: params[:username], app_identifier: app_identifier, platform: params[:platform])
         end
       end
 
       # Certificate
       cert_id = fetch_certificate(params: params, working_directory: storage.working_directory)
-      spaceship.certificate_exists(username: params[:username], certificate_id: cert_id) if spaceship
+      # spaceship.certificate_exists(username: params[:username], certificate_id: cert_id, platform: params[:platform]) if spaceship
```
snatchev: remove commented out code plz 🙇

snatchev (changes requested):

match/lib/match/runner.rb
joshdholtz: Ope, you right!

match/lib/match/runner.rb
joshdholtz: This only happens on `readonly: true` which is a user specified setting so no warning is needed

fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane_core/lib/fastlane_core/cert_checker.rb
snatchev: @joshdholtz could you give this a try?

snatchev (approved):
👍 Looks great, thank you!

crewshin:
Is there any kind of tutorial/docs for adding this functionality to an existing fastlane/match setup? I don't really see anything docs related in the changed files.

I'm getting:
"An app with that bundle ID needs to exist in order to create a provisioning profile for it" and "Couldn't find bundle identifier 'maccatalyst.*.*.*.*' for the user '*******'"

Or if I try and manually create a profile from the portal, I get: 
"An attribute in the provided entity has invalid value
An App ID with Identifier 'maccatalyst.*.*.*.*' is not available. Please enter a different string."

This sounds pretty noobish, but I'm telling you that I've been through signing issues quite a bit in the past and this catalyst stuff is a pita. Nothing seems to work.

PR #15928, opened 2020-01-23
[match] Change http_auth_header capitalization

sakul1991:
### Checklist

### Motivation and Context
In bitbucket server the header for HTTP Basic Auth is case sensitive, contrary to http standard specifications, for this reason git cannot authorize against bitbucket server using the git_basic_authorization or git_bearer_authorization parameters. 

### Description
Changed the headers spelling to one more commonly used. Tested and verified against our corporate bitbucket server installation.

### Testing Steps
Unfortunately not possible without a bitbucket server and connecting via http basic auth.

### DIFF
--- a/match/lib/match/storage/git_storage.rb
+++ b/match/lib/match/storage/git_storage.rb
@@ -73,8 +73,8 @@ module Match
         self.working_directory = Dir.mktmpdir
 
         command = "git clone #{self.git_url.shellescape} #{self.working_directory.shellescape}"
-        command << " -c http.extraheader='AUTHORIZATION: basic #{self.git_basic_authorization}'" unless self.git_basic_authorization.nil?
-        command << " -c http.extraheader='AUTHORIZATION: bearer #{self.git_bearer_authorization}'" unless self.git_bearer_authorization.nil?
+        command << " -c http.extraheader='Authorization: Basic #{self.git_basic_authorization}'" unless self.git_basic_authorization.nil?
+        command << " -c http.extraheader='Authorization: Bearer #{self.git_bearer_authorization}'" unless self.git_bearer_authorization.nil?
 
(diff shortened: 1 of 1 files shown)


janpio:
Is there some public documentation for this requirement of case sensitivity for HTTPS headers for Bitbucket? Super strange.

match/lib/match/storage/git_storage.rb
```diff
@@ -73,8 +73,8 @@ def download
         self.working_directory = Dir.mktmpdir
 
         command = "git clone #{self.git_url.shellescape} #{self.working_directory.shellescape}"
-        command << " -c http.extraheader='AUTHORIZATION: basic #{self.git_basic_authorization}'" unless self.git_basic_authorization.nil?
-        command << " -c http.extraheader='AUTHORIZATION: bearer #{self.git_bearer_authorization}'" unless self.git_bearer_authorization.nil?
+        command << " -c http.extraheader='Authorization: Basic #{self.git_basic_authorization}'" unless self.git_basic_authorization.nil?
+        command << " -c http.extraheader='Authorization: Bearer #{self.git_bearer_authorization}'" unless self.git_bearer_authorization.nil?
```
joshdholtz: Can we had comment above here like like this?
```
# HTTP headers are supposed to be be case insensitive but 
# Bitbucket requires `Authorization: Basic` and `Authorization Bearer` to work
```

joshdholtz (changes requested):
I'm not a really a fan of this but if its what we need to do 😬 A comment like the one I suggested would be great so somebody doesn't come back in here and change it

joshdholtz (approved):
Thanks for making this change! Hopefully this will work for everyone now 🙃 🤞 Appreciate the contribution ❤️

PR #15943, opened 2020-01-28
[deliver] Add skip_auto_detecting_app_version option

genius-munou:
### Checklist

### Motivation and Context

Due to various circumstances, AppStore and CFBundleShortVersionString versions do not match. app_version is specified in Deliverfile.

If the train version is different from the app version, `tunes_all_builds_for_train(train: app_version)` is called when acquiring a candidate build with wait_for_build, and `ITC.response.error.OPERATION_FAILED` occurs.

With this fix, the latest uploaded build is always sent for review, even if the app_version is specified.

This problem did not occur when using `fastlane 2.114.0`, but subsequent fixes seem to have caused the problem.

### Description
Added option to use latest_version in `wait_for_build(use_latest_build)`, default value is false.
When use_latest_build is true, the same processing as when app version is not specified can be performed by passing nil to wait_for_build.

We confirmed that no error occurred by actually executing the "deliver" action in an application where the AppStore version and CFBundleShortVersionString were different.

### DIFF
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -42,7 +42,8 @@ module Deliver
         end
       else
         UI.message("Selecting the latest build...")
-        build = wait_for_build(app, app_version)
+        target_app_version = options[:use_latest_build] ? nil : app_version
+        build = wait_for_build(app, target_app_version)
       end
       UI.message("Selecting build #{app_version} (#{build.build_version})...")
 
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -42,8 +42,7 @@ module Deliver
         end
       else
         UI.message("Selecting the latest build...")
-        target_app_version = options[:use_latest_build] ? nil : app_version
-        build = wait_for_build(app, target_app_version)
+        build = wait_for_build(app, app_version)
       end
       UI.message("Selecting build #{app_version} (#{build.build_version})...")
 
--- a/deliver/lib/deliver/detect_values.rb
+++ b/deliver/lib/deliver/detect_values.rb
@@ -59,6 +59,7 @@ module Deliver
 
     def find_version(options)
       return if options[:app_version]
+      return if options[:skip_auto_detecting_app_version]
 
       if options[:ipa]
         options[:app_version] ||= FastlaneCore::IpaFileAnalyser.fetch_app_version(options[:ipa])
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -85,12 +85,6 @@ module Deliver
                                      verify_block: proc do |value|
                                        UI.user_error!("The platform can only be ios, appletvos, or osx") unless %('ios', 'appletvos', 'osx').include?(value)
                                      end),
-        FastlaneCore::ConfigItem.new(key: :use_latest_build,
-                                     optional: true,
-                                     default_value: false,
-                                     env_name: "DELIVER_USE_LATEST_BUILD",
-                                     description: "Force usage of latest build version",
-                                     is_string: false),
 
         # live version
         FastlaneCore::ConfigItem.new(key: :edit_live,
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -85,6 +85,12 @@ module Deliver
                                      verify_block: proc do |value|
                                        UI.user_error!("The platform can only be ios, appletvos, or osx") unless %('ios', 'appletvos', 'osx').include?(value)
                                      end),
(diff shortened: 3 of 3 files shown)


genius-munou:
@googlebot I signed it!

janpio:
Hm, the option naming `use_latest_build` might be a bit too broad - use latest build for what and why? Can this be described better in some way?

genius-munou:
@janpio 
Thank you for reacting.  
First, the latest build is usually used to send reviews to Apple.  

And the existing implementation of `wait_for_build` selects the latest build from the latest version candidates, unless` app_version` is specified as follows:
```ruby
      use_latest_version = app_version.nil?

      loop do
        # Sometimes candidate_builds don't appear immediately after submission
        # Wait for candidate_builds to appear on App Store Connect
        # Issue https://github.com/fastlane/fastlane/issues/10411
        if use_latest_version
          candidate_builds = app.latest_version.candidate_builds
        else
          candidate_builds = app.tunes_all_builds_for_train(train: app_version)
        end
        if (candidate_builds || []).count == 0
          UI.message("Waiting for candidate builds to appear...")
          if (Time.now - start) > (60 * 5)
            UI.user_error!("Could not find any available candidate builds on App Store Connect to submit")
          else
            sleep(30)
            next
          end
        end

        latest_build = find_build(candidate_builds)

```
<br>
This makes a lot of sense, but if you have a build version different from `app_version`, you won't be able to suggest the latest build binaries.  

For example, some of our apps have changed v1.0.0 (AppStore) to v100 (build version).  

In such cases, you will be able to explicitly specify the latest build version when submitting to the AppStore.  
How about the option name in `use_explicit_latest_build_binary`? Please give me advice!

deliver/lib/deliver/submit_for_review.rb
```diff
@@ -42,7 +42,8 @@ def submit!(options)
         end
       else
         UI.message("Selecting the latest build...")
-        build = wait_for_build(app, app_version)
+        target_app_version = options[:use_latest_build] ? nil : app_version
+        build = wait_for_build(app, target_app_version)
```
joshdholtz: This should probably be moved into https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/detect_values.rb#L63-L67

joshdholtz (changes requested):
I did some digging into why `app_version` was actually never `nil` (because IMO it should be if you don't specify it and it turns out its because of https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/detect_values.rb#L63-L67

☝️ `deliver` will always try to auto detect the app version from the IPA or PKG **even** if you explicitly set `nil` 😬

I like my ideal approach here would be to:
1) not add a new option
2) explicitly setting `nil` would stop the detection of IPA or PKG

I'm not sure how possible or clean that is though 🤔 Instead, maybe we...
1) Rename to `skip_auto_detecting_app_version`
2) Move this logic into before https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/detect_values.rb#L63-L67

Thoughts?

genius-munou:
Thank you for the response!
In the case of the pointed out method, will I have to call `deliver` separately to create a new version in AppStore Connect?
https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/runner.rb#L89-L92

I think that my explanation was probably not good, so I'll list what I want to do again.
1. AppStore wants to use general version notation like `1.2.3`. (Set to Deliverfile)
2. However, the IPA (or PKG) version does not match the AppStore version like `123`. (Because it has already been released, you can not submit it with a lower version)
3. I want to use the latest build uploaded to AppStore Connect for submission to Apple.
4. If possible, I want to complete with one `deliver` action call.

Is 4 too greedy?😅

genius-munou:
@joshdholtz 
Thank you for the response!
In the case of the pointed out method, will I have to call `deliver` separately to create a new version in AppStore Connect?
https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/runner.rb#L89-L92

I think that my explanation was probably not good, so I'll list what I want to do again.
1. AppStore wants to use general version notation like `1.2.3`. (Set to Deliverfile)
2. However, the IPA (or PKG) version does not match the AppStore version like `123`. (Because it has already been released, you can not submit it with a lower version)
3. I want to use the latest build uploaded to AppStore Connect for submission to Apple.
4. If possible, I want to complete with one `deliver` action call.

Is 4 too greedy?😅

deliver/lib/deliver/submit_for_review.rb
genius-munou: OK, moved the logic as you pointed out.☺️

genius-munou:
@joshdholtz Sorry to bother you, but please check when you can 🙏

maxammann:
Had the same problem with verson codes like "2020.04.1". Apple automatically creates a version "2020.4.1". Therefore it is not able to find the build.

PR #15946, opened 2020-01-28
[app_store_build_number/latest_testflight_build_number ] Export version in SharedValues

### LINKED ISSUE #15735: latest_testflight_build_number should export version number

The action latest_testflight_build_number can query the latest TestFlight build number for either a specified version or the current live version of an app. It would be helpful if in the latter case the action could also export the actual (short) version number of the live version.

This would make for more comprehensive status information (I want to be able to quickly check what short version and bundle number of my app is currently live in the store).

#### Feature Description

Add a second export parameter `LATEST_TESTFLIGHT_SHORT_VERSION` to the action and fill it with the short version that the build number applies to.

I am not aware of how to get this information otherwise using fastlane.

Johennes:
This adds the version as second export parameter to the mentioned actions. This
is useful when the live version is queried.

Fixes: #15735

### Checklist

### Motivation and Context

See linked issue.

### Description

See commit message.

### Testing Steps

Run the two actions.

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_build_number.rb
+++ b/fastlane/lib/fastlane/actions/app_store_build_number.rb
@@ -1,14 +1,18 @@
+require 'ostruct'
+
 module Fastlane
   module Actions
     module SharedValues
       LATEST_BUILD_NUMBER = :LATEST_BUILD_NUMBER
+      LATEST_VERSION = :LATEST_VERSION
     end
 
     class AppStoreBuildNumberAction < Action
       def self.run(params)
         require 'spaceship'
 
-        build_nr = get_build_number(params)
+        result = get_build_number(params)
+        build_nr = result.build_nr
 
         # Convert build_nr to int (for legacy use) if no "." in string
         if build_nr.kind_of?(String) && !build_nr.include?(".")
@@ -16,6 +20,9 @@ module Fastlane
         end
 
         Actions.lane_context[SharedValues::LATEST_BUILD_NUMBER] = build_nr
+        Actions.lane_context[SharedValues::LATEST_VERSION] = result.build_v
+
+        return build_nr
       end
 
       def self.get_build_number(params)
@@ -33,7 +40,7 @@ module Fastlane
 
           UI.message("Latest upload for live-version #{app.live_version.version} is build: #{build_nr}")
 
-          return build_nr
+          return OpenStruct.new({ build_nr: build_nr, build_v: app.live_version.version })
         else
           version_number = params[:version]
           platform = params[:platform]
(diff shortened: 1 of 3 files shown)


Johennes:
I signed the CLA.

janpio:
@googlebot unfortunately does not agree - or the CLA label would have switched to green and `yes`. Different email address than commits maybe?

janpio:
Uh, does that also change the return value of the actions and thus be a breaking change for existing users?

Johennes:
@janpio, no, the return value of the actions is still the build number. I only changed the return value of an internal (at least I hope it's internal) method to also include the version string.

janpio:
Ah right, missed that line. 
We're using `ostruct` in other places as well, so this should be fine.

fastlane/lib/fastlane/actions/app_store_build_number.rb
```diff
@@ -159,13 +166,14 @@ def self.available_options
...
-          ['LATEST_BUILD_NUMBER', 'The latest build number of either live or testflight version']
+          ['LATEST_BUILD_NUMBER', 'The latest build number of either live or testflight version'],
+          ['LATEST_VERSION', 'The version of the latest build number']
         ]
       end
 
       def self.details
         [
-          "Returns the current build number of either the live or testflight version - it is useful for getting the build_number of the current or ready-for-sale app version, and it also works on non-live testflight version.",
+          "Returns the current build number and version of either the live or testflight version - it is useful for getting the build_number of the current or ready-for-sale app version, and it also works on non-live testflight version.",
```
janpio: This is not correct then. It only returns the build number (and also makes the version available in `SharedValues`).

fastlane/lib/fastlane/actions/latest_testflight_build_number.rb
```diff
@@ -4,20 +4,25 @@ module Fastlane
...
+        return build_nr
       end
 
       #####################################################
       # @!group Documentation
       #####################################################
 
       def self.description
-        "Fetches most recent build number from TestFlight"
+        "Fetches most recent build number and version from TestFlight"
```
janpio: See https://github.com/fastlane/fastlane/pull/15946/files#r373800255

fastlane/lib/fastlane/actions/app_store_build_number.rb
Johennes: Good call. I removed my changes to the description strings in both actions. The main purpose of the actions still is to fetch the build number and the fact that the version string is also exported via the `SharedValues` is already contained in the action details.

joshdholtz (approved):
This is 💯 Thanks for making this change! Really appreciate the contribution ❤️

PR #15951, opened 2020-01-29
[socket_server_action_command_executor] Fix sh error_callback

jhandguy:
### Checklist

### Motivation and Context

In Fastlane.swift, the `sh` action never fails, whether an `error_callback` is given or not.
It was intended for the `error_callback` to be optional, however, the `socket_server_action_command_executor` was setting a `proc` anyway.

### Description

Setting a `proc` as error callback only when an actual `error_callback` is given fixes the issue.


taquitos (approved):
This looks totally reasonable. Thank you for the fix!
🎈🐐

PR #15958, opened 2020-01-31
Allow subfolders in the Actions folder

loremattei:
### Checklist

### Motivation and Context
Sometimes it can useful to be able to organise actions into subfolders. `Fastlane` already supports this, in fact the actions that live in a subfolder of the `actions` folder are found and used. 
The only issue is that they are not shown in the initial dashboard and, if there are no actions in the root `Actions/` folder, a warning is shown at every run.

### Description
This change makes `Fastlane` look for actions also in subfolders of the `action` folder when the plugin is discovered and loaded by considering all the folders that have `/actions` in the path and not only the ones that have it has the last part of the path.

### DIFF
--- a/fastlane/lib/fastlane/plugins/plugin_manager.rb
+++ b/fastlane/lib/fastlane/plugins/plugin_manager.rb
@@ -373,7 +373,7 @@ module Fastlane
       # (a plugin may contain any number of actions)
       version_number = Fastlane::ActionCollector.determine_version(gem_name)
       references = Fastlane.const_get(module_name).all_classes.collect do |path|
-        next unless File.dirname(path).end_with?("/actions") # we only want to match actions
+        next unless File.dirname(path).include?("/actions") # we only want to match actions
 
         File.basename(path).gsub("_action", "").gsub(".rb", "").to_sym # the _action is optional
       end


loremattei:
@googlebot I signed it!

janpio:
What do you refer to by "initial dashboard"?
Any documentation to update?

loremattei:
> What do you refer to by "initial dashboard"?

Oh, sorry I wasn't clear. I mean the message that is shown when `fastlane` starts. 
This one: 
```
+-----------------------------------+---------+----------------------------------------------------------------------------------------+
|                                                             Used plugins                                                             |
+-----------------------------------+---------+----------------------------------------------------------------------------------------+
| Plugin                            | Version | Action                                                                                 |
+-----------------------------------+---------+----------------------------------------------------------------------------------------+
| fastlane-plugin-XXX                          | X.Y.Z   | No actions found                                                                       |
| fastlane-plugin-YYY                          | XX.YY.ZZ   | action_1 action_2 action_n  |
+-----------------------------------+---------+----------------------------------------------------------------------------------------+

[!] No actions were found while loading one or more plugins
    Please use `bundle exec fastlane` with plugins
    More info - https://docs.fastlane.tools/plugins/using-plugins/#run-with-plugins
```

> Any documentation to update?
I don't think. The actions are not shown in that table, but if you then call them they work flawlessly, so this PR doesn't change any actual behaviour.

fastlane/lib/fastlane/plugins/plugin_manager.rb
```diff
@@ -373,7 +373,7 @@ def store_plugin_reference(gem_name)
       # (a plugin may contain any number of actions)
       version_number = Fastlane::ActionCollector.determine_version(gem_name)
       references = Fastlane.const_get(module_name).all_classes.collect do |path|
-        next unless File.dirname(path).end_with?("/actions") # we only want to match actions
+        next unless File.dirname(path).include?("/actions") # we only want to match actions
```
janpio: Could this be to "lax" in some way? Right now it is quite strict looking for `..../actions`. Is there any way that `.../actions...` could lead to problems? What values can `path` have?

loremattei: > What values can path have?

They are the values defined by the plugin in `all_classes`. 
And this is the point: in the plugin you can do something like 
```
def self.all_classes
      Dir[File.expand_path('**/{actions,helper}/**/*.rb', File.dirname(__FILE__))]
end
```
and it will load all the actions in `actions` subfolders. 
But they aren't listed when `fastlane` starts because of the `end_with?` limit.

>  Is there any way that .../actions... could lead to problems?

The only bad case I could think about was that if a plugin has `helpers` with a name that starts with "actions", they will pass the filter and get listed together with the other actions. 

Two thought about this:
- the initial `/` in the filter should make sure that only helpers with a name that _starts_ with "actions" could pass the filter. If "actions" is used anywhere else in the helper's name, they will be filtered out.
- the worst case is that a helper is listed in the initial message. This is probably as bad as not having actions listed like now, but it still shouldn't affect the plugin functionalities.

janpio: I agree, this sounds like a win.

joshdholtz (approved):
Seems good to me 😊

Issue #403, opened 2015-07-10
silent option is not available in ipa block


iCarry:
ipa(
      configuration: "Release",
      scheme: "HelloWorld",
      destination: "fastlane_build",
      archive: true,
      silent: false,
    )

result:

    [15:12:33]: You passed invalid parameters to 'ipa'.
    [15:12:33]: Check out the error below and available options by running `fastlane action ipa`
    [15:12:33]: Variable Dump:
    [15:12:33]: {:DEFAULT_PLATFORM=>:ios, :ENVIRONMENT=>nil, :PLATFORM_NAME=>:ios, :LANE_NAME=>"ios qa", :BUILD_NUMBER=>0}
    [15:12:33]: Could not find available option 'silent' in the list of available options: workspace, project, configuration, scheme, clean, archive, destination, embed, identity, sdk, ipa, xcconfig, xcargs
    [15:12:33]: fastlane finished with errors
    /Library/Ruby/Gems/2.0.0/gems/fastlane_core-0.9.2/lib/fastlane_core/configuration.rb:45:in `block in verify_value_exists': Could not find available option 'silent' in the list of available options: workspace, project, configuration, scheme, clean, archive, destination, embed, identity, sdk, ipa, xcconfig, xcargs (RuntimeError)

KrauseFx:
Sorry, there is no `silent` option. I'll update the README, please just remove this key for now :+1:

KrauseFx:
Where do you have the `silent` from?

iCarry:
Hi,
It's part of the ipa action (version 1.7)
https://github.com/KrauseFx/fastlane/blob/master/lib/fastlane/actions/ipa.rb

However, that option is not available when used in the ipa block.

iCarry:
oh, nvm. 
It's silent,  I spelled it wrong...

KrauseFx:
I removed it from the last version, sorry about the confusion

PR #15970, opened 2020-02-04
Security: Updating rubyzip version to 2.2.0

sylvester-loreto-mypulse:
Updating rubyzip version to 2.2.0 as 1.3.0 has security vulnerabilities.

https://github.com/rubyzip/rubyzip/releases/tag/v2.0.0
https://github.com/rubyzip/rubyzip/releases/tag/v2.2.0

### Checklist


janpio (changes requested):
Fastlane supports older Ruby versions (because those are still used as default for older macos), so we will probably not be able to merge this. Right @joshdholtz?

But the release notes mention another workaround:

> This option defaulted to false in 1.3.0 for backward compatibility, but it now defaults to true. If you are using an older version of ruby and can't yet upgrade to 2.x, you can still use 1.3.0 and set the option to true.

Maybe this can work for fastlane.

sylvester-loreto-mypulse:
This is not about running older ruby versions.
This is about a security vulnerability which has been fixed on `rubyzip` [version 2.0.0](https://github.com/rubyzip/rubyzip/releases/tag/v2.0.0).

> **Security**
> Default the `validate_entry_sizes` option to `true`, so that callers can trust an entry's reported size when using extract #403
> This option defaulted to `false` in `1.3.0` for backward compatibility, but it now defaults to `true`. If you are using an older version of ruby and can't yet upgrade to 2.x, **you can still use 1.3.0** and set the option to `true`.

janpio:
As you can see in the failing tests, newer `rubyzip` versions also require newer Ruby versions:

> `ERROR:  Error installing fastlane-2.141.0.gem:
	There are no versions of rubyzip (~> 2.2.0) compatible with your Ruby & RubyGems. Maybe try installing an older version of the gem you're looking for?
	rubyzip requires Ruby version >= 2.4. The current ruby version is 2.3.8.459.`

sylvester-loreto-mypulse:
Any particular reason to stick to an older version? Given there is a security issue, wouldn't be better to update to a more recent and stable version?
Also, `2.3.8` is no longer maintained.
https://www.ruby-lang.org/en/downloads/

fastlane.gemspec
```diff
@@ -86,7 +86,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('json', '< 3.0.0') # Because sometimes it's just not installed
   spec.add_dependency('mini_magick', '>= 4.9.4', '< 5.0.0') # To open, edit and export PSD files
   spec.add_dependency('multi_xml', '~> 0.5')
-  spec.add_dependency('rubyzip', '>= 1.3.0', '< 2.0.0') # fix swift/ipa in gym
+  spec.add_dependency('rubyzip', '~> 2.2.0') # fix swift/ipa in gym
```
joshdholtz: We can't accept locking `rubyzip` to `2.2.0` and up due to the `rubyzip` requirement of Ruby 2.4. We could possibly accept `< 2.2.0` it it doesn't break anything 🤔

joshdholtz (commented):
When building testing and building _fastlane_ we use newer versions of Ruby but we still need to allow our users to use older versions of Ruby. Some of the machines that users are using already have older versions of Ruby preinstalled and we don't want to prevent them from updating _fastlane_ if they can't or if it isn't easy for them to do.

For now we need to keep supporting back to Ruby 2.1ish

fastlane.gemspec
sylvester-loreto-mypulse: It looks like my comments are not being read nor understood.

What part from my previous comments regarding the security fix on `rubyzip 2.0.0` and `ruby 2.3.8` is no longer maintained?

janpio: It does not matter what is not maintained. Users of this library are using it on these versions, so we support these versions. We can not upgrade to a library version that does not support the versions of Ruby some users are using.

joshdholtz: @janpio is correct. We understand that there older versions have security issues and are not necessarily maintained. Updating to this would be a breaking change for a lot of users that are using older versions of Ruby that **fastlane still needs to support** and **that are still available for download**.

I would like it if could bump up the minimum required version of _fastlane_ but we cannot do it at this time **unless...**  `rubyzip 2.2.0` isn't a breaking change for us. 

If we could `spec.add_dependency('rubyzip', '>= 1.3.0', '< 2.2.0')` without _fastlane_ breaking that would allow us to **support older versions of Ruby** while allowing users to update to **that latest and securest dependencies**. But I don't know if the APIs for `rubyzip` between `1.3.0` and `2.2.0` remained unchanged that wouldn't break any way that we use it.

janpio:
To keep using 1.3.x you have to modify the `Zip` behavior like this (according to the release notes): https://github.com/fastlane/fastlane/pull/16072#issue-378742701

sylvester-loreto-mypulse:
Why was this PR closed?
I made the changes as per @joshdholtz recommendation. There was another issue #16072 trying to address this but both have now been closed and the issue persists.
Have a look at my latest commit https://github.com/fastlane/fastlane/pull/15970/commits/90f258712ea5dc9cf22bf5a8661ecd1f71af5863

joshdholtz:
Ah, this may have been closed by GitHub when a PR that referenced this was merged?! 🤔 I did. It close this on purpose even though it says I did 😬 I actually have no idea. Reopening. Sorry about that!

janpio:
#16072 solved the security problem by setting the param advised by rubyzip to be used if we can not use the newer versions - which we could not back then.

Does this PR add anything on top of that?

turtlemaster19:
@joshdholtz It was closed because at the bottom of the PR, you included: `"Closes #15970"`
I agree that it should be closed.

joshdholtz:
I may have referenced the wrong number? 🤷‍♂️ I don’t know if it should be closed but I think we should change this so rubyzip is greater than 2.0.

I’ll add this to my list of things this week 😊

turtlemaster19:
@janpio Of course, for this particular pull to be merged, instead of a separate one and this being closed, we'll need @sylvester-loreto-mypulse to sign CLAs.

janpio:
Replaced by https://github.com/fastlane/fastlane/pull/16660

Issue #15979, opened 2020-02-05
Getting a deprecation notice for authentication via URL query parameters because of Octokit


eygraber:
I started getting emails from GitHub saying that I am using a deprecated authentication method via OCtokit. Fastlane is the only thing in my project that uses Octokit. They released a new version that fixes this.

[See here for more details](https://github.com/octokit/octokit.rb/issues/1191)

eygraber:
I don't use fastlane locally, and cannot access my CI right now to get that output. This is a pretty cut and dry issue. I don't think you need the output to deal with it.

janpio:
What action are you using? We have quite a few of them, and multiple could talk to Github.

eygraber:
As far as I can tell, we only use `set_github_release` and `get_github_release`

PR #15995, opened 2020-02-08
[get_github_release] Use GITHUB_API_TOKEN for default for consistency

chuganzy:
### Checklist

### Motivation and Context

Other GitHub related actions use `GITHUB_API_TOKEN` but `get_github_release` does not. For consistency I thought it is better to use `GITHUB_API_TOKEN` here too.

### Description

Just use `ENV['GITHUB_API_TOKEN']` as the default value.

### Testing Steps

n/a

### DIFF
--- a/fastlane/lib/fastlane/actions/get_github_release.rb
+++ b/fastlane/lib/fastlane/actions/get_github_release.rb
@@ -129,6 +129,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :api_token,
                                        env_name: "FL_GITHUB_RELEASE_API_TOKEN",
                                        sensitive: true,
+                                       default_value: ENV["GITHUB_API_TOKEN"],
                                        description: "GitHub Personal Token (required for private repositories)",
                                        optional: true)
         ]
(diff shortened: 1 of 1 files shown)


eygraber:
Is this related? https://github.com/fastlane/fastlane/issues/15979

chuganzy:
@eygraber Hi. I looked into the issue but I think it is not related to this PR.

fastlane/lib/fastlane/actions/get_github_release.rb
```diff
@@ -129,6 +129,7 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :api_token,
                                        env_name: "FL_GITHUB_RELEASE_API_TOKEN",
                                        sensitive: true,
+                                       default_value: ENV["GITHUB_API_TOKEN"],
```
joshdholtz:
```suggestion
                                       code_gen_sensitive: true,
                                       default_value: ENV["GITHUB_API_TOKEN"],
                                       default_value_dynamic: true,
```

Good call! Can we also add these too? ^

joshdholtz (changes requested):
Love this change! Just a need to add `code_gen_sensitive ` and `code_gen_sensitive` and then I think we should be good 😊

joshdholtz (approved):
Looks good to me now! Thanks for fixing this 😊  Hope all is well ❤️

chuganzy:
@joshdholtz Thank you! Yeah, doing good. Hope you are doing well too!

PR #16008, opened 2020-02-12
[snapshot] disable Pasteboard sync

xavierLowmiller:
### Checklist

### Motivation and Context
As pointed out by [Peter Steinberger on Twitter](https://twitter.com/steipete/status/1227551552317140992), the Pasteboard sharing of the Simulator can cause issues during UI Tests. `UIResponder.becomeFirstResponder` can trigger the Pasteboard sync, which can lead to stalls.

This PR disables Pasteboard automatic sync on a system level for all launched simulators.

### Description
I chose to implement this in the `Snapshot::Fixes` module for two reasons:
+ Ideally, Pasteboard automatic sharing shouldn't interfere with UI tests, but it does. 
+ The implementation and intent of the code is very similar to the other Fixes

### Testing Steps

1. Turn on Simulator Pasteboard Automatic Sync:
`defaults write com.apple.iphonesimulator PasteboardAutomaticSync -bool true`
The command `defaults read com.apple.iphonesimulator PasteboardAutomaticSync` should return `1`.
2. Run `fastlane snapshot`
3. The command `defaults read com.apple.iphonesimulator PasteboardAutomaticSync` should return `0`.

### DIFF
--- a/snapshot/lib/snapshot/fixes/simulator_shared_pasteboard.rb
+++ b/snapshot/lib/snapshot/fixes/simulator_shared_pasteboard.rb
@@ -1,0 +1,16 @@
+require_relative '../module'
+
+module Snapshot
+  module Fixes
+    # Becoming first responder can trigger Pasteboard sync, which can stall and crash the simulator
+    # See https://twitter.com/steipete/status/1227551552317140992
+
+    class SharedPasteboardFix
+      def self.patch
+        UI.verbose("Patching simulator to disable Pasteboard automatic sync")
+
+        Helper.backticks("defaults write com.apple.iphonesimulator PasteboardAutomaticSync -bool false", print: FastlaneCore::Globals.verbose?)
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
This is perfectly implemented 🥰 Thanks so much for making this PR so quickly!

akmarinov:
@xavierLowmiller is this enabled for scan as well or just snapshot? Should we include it as an option in scan?

xavierLowmiller:
I assumed that the issue only happens in UI Tests.

I didn't test what happens when `becomeFirstResponder` is triggered programmatically in unit tests, so if this is an issue as well, it can be easily be used in `scan`, too.

akmarinov:
It does happen only for UITest, I just saw that it was implemented under snapshot and was wondering whether it only applies when capturing screenshots via snapshot or also works with UITests that are run via scan?

xavierLowmiller:
Oh! Obviously!

Yes, this should be in `scan`.

I'll look into this when I have some time 👍.

Issue #15992, opened 2020-02-07
`load_plugins` does not respect bundler groups


manuyavuz-pointr:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I have a Gemfile with a few fastlane plugins that are separated with bundler groups. Gemfile has the following:

```
group :ios, optional: true do
  gem 'fastlane-plugin-firebase_test_lab'
end

group :android, optional: true do
  gem 'fastlane-plugin-run_tests_firebase_testlab'
end
```

I would expect plugin loader to try to validate only plugins from enabled groups. However, it prints failure messages for the plugin from disabled group `:android`.

##### Command executed

`bundle exec fastlane`

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> (base) ➜  fastlane-load-plugins-sample git:(master) ✗ bundle exec fastlane ios custom_lane
[✔] 🚀
[21:54:01]: Missing gem 'fastlane-plugin-run_tests_firebase_testlab', please add the following to your local Gemfile:
[21:54:01]:
[21:54:01]: ▸ gem "fastlane-plugin-run_tests_firebase_testlab"
[21:54:01]:
[21:54:01]: Error loading plugin 'fastlane-plugin-run_tests_firebase_testlab': Add 'gem "fastlane-plugin-run_tests_firebase_testlab"' to your Gemfile and restart fastlane
+---------------------------+-----------+---------------------------+
|                           Used plugins                            |
+---------------------------+-----------+---------------------------+
| Plugin                    | Version   | Action                    |
+---------------------------+-----------+---------------------------+
| fastlane-plugin-firebase  | 1.0.5     | firebase_test_lab_ios_xc  |
| _test_lab                 |           | test                      |
| fastlane-plugin-run_test  | undefined | No actions found          |
| s_firebase_testlab        |           |                           |
+---------------------------+-----------+---------------------------+

[!] No actions were found while loading one or more plugins
    Please use `bundle exec fastlane` with plugins
    More info - https://docs.fastlane.tools/plugins/using-plugins/#run-with-plugins

[21:54:01]: ------------------------------
[...]

PR #16013, opened 2020-02-13
[PluginManager] Take into account enabled bundler groups

manuyavuz:
### Checklist

### Motivation and Context

Fixes issue #15992.

### Description

Used intersection of `Bundler.definition.specs` and `Bundler.definition.dependencies` to get direct dependencies of `Gemfile` filtered by enabled groups.

Tested on my project which previously was failing due to `plugin not loaded` errors. With the change, fastlane runs successfully.

### Important
Currently `plugin_manager` specs are failing because they directly mock `Bundler::SharedHelpers.default_gemfile`. However, with new change, fastlane directly depends on what bundler has built as definition during startup. I don't know a convenient way to mock this with a `Gemfile`. Any suggestions on this?

cc @janpio


manuyavuz:
@googlebot I signed it!

fastlane/lib/fastlane/plugins/plugin_manager.rb
```diff
@@ -54,8 +54,10 @@ def self.to_gem_name(plugin_name)
     # Returns an array of gems that are added to the Gemfile or Pluginfile
     def available_gems
       return [] unless gemfile_path
-      dsl = Bundler::Dsl.evaluate(gemfile_path, nil, true)
-      return dsl.dependencies.map(&:name)
+      # Bundler's `dependencies` method returns all gems included in Gemfile, ignoring enabled Bundler groups.
+      # To get only gems from enabled groups, we need to combine with `specs`, which does include only gems from enabled groups.
+      definition = Bundler.definition
```
joshdholtz:
```suggestion
      definition = Bundler::Dsl.evaluate(gemfile_path, nil, true)
```
I think this can still use `Bundle::Dsl.evaluate` as this returns a definition. We need to this so that we can pass a `gemfile_path`during tests

joshdholtz (commented):
This change makes complete sense 👌  I'm spent some time looking into what is happening with the intersection and even though it works on your end it seems that intersection might not be the correct solution 🤔  Either that for the unit tests are testing the wrong thing. 

I might have to spend some more time trying to understand what is happening here and what _actually_ needs to happen 😇

fastlane/lib/fastlane/plugins/plugin_manager.rb
manuyavuz: @joshdholtz thanks for review!

It's true that `Dsl.evaluate` returns a definition, but it does not perform actual dependency resolution, so `specs` method returns `nil` on that. Instead, Bundler.definition gives you currently active environment which includes dependency resolution.

Maybe we can force returned `definition` to do resolution again using `resolve_remotely!` or `resolve_with_cache!` methods, but there is no guarantee that it will be the same as currently active bundler environment.


I believe that this implementation behaves correctly, but I just couldn't find an easy way to change Bundler.definition behavior for unit tests.

Maybe we can extract inline `Bundler.definition` call to a method, and stub that method for unit tests to return a resolved bundler definition from specific gemfile we want in unit tests?

joshdholtz: It's been a few weeks since I looked at this but it didn't look like it was returning the correct/expected gems. Either that or _fastlane_ isn't doing the correct things with what is being returned. I will run this again and see if I can provide a better analysis of what's happening and what's expected 🤔

PR #16025, opened 2020-02-16
Use size-label-action to label PRs by size

janpio:
This PR adds the [`size-label-action` GitHub Action](https://github.com/pascalgn/size-label-action) to the repository to label new incoming PRs automatically by size.

I created the labels already:
image

(The label is structure is different from what we use, if this turns out to be useful I would create a PR to the action's repo to make the structure configurable [from `size/` to `size: ` for example])

### DIFF
--- a/.github/workflows/size-label.yml
+++ b/.github/workflows/size-label.yml
@@ -1,0 +1,10 @@
+name: size-label
+on: pull_request
+jobs:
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Love it! ❤️ Looking forward to this 👍

PR #16029, opened 2020-02-17
Update size-label.yml

joshdholtz:
### Motivation and Context
We got a 403 on the run

### Description
We use the `BOT_GITHUB_TOKEN` secret

### DIFF
--- a/.github/workflows/size-label.yml
+++ b/.github/workflows/size-label.yml
@@ -7,4 +7,4 @@ jobs:
       - name: size-label
         uses: "pascalgn/size-label-action@df7ad4303b35cbeb20937dbb12d5a050520e469e"
         env:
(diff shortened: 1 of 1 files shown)


janpio:
Sure this will fix the problem? This was a copy/paste from the Action README.

joshdholtz:
@janpio Yup yup, it did! Look at the label added to this PR 😊

janpio:
Oh yeah that also worked on my PR before when I added it of course!

joshdholtz:
I'm so confused then 😱 It said there was 403 when it was running once it was merged. This is the variable in other workflows though 🤔  Maybe we give this a try?

janpio:
Yeah this will just change from the `github-action` user to `fastlane-bot` - which is fine of course. But I suspect there to be a different problem :/

janpio (approved):

PR #16043, opened 2020-02-18
[scan] Add testPlan option to scan

### LINKED ISSUE #15462: [scan] Support for -testPlan and .xctestplan
Xcode 11 supports test plans which simplify configuration. I don't believe that everything that `scan` supports is also supported by test plans, so I would be interested in the niceties of `scan` being available for running tests configured as an .xctestplan

#### Feature Description
Evolve scan to support the use of .xctestplan files to define (and override) the parts of running tests that are redundant between xctestplan and scan, while continuing to support the xcpretty and any other scan-specific features that xctestplans do NOT support.

```
xcodebuild -workspace Project.xcworkspace -scheme ProjectUITests -destination 'platform=iOS Simulator,OS=13.1,name=iPhone 11 Pro' test -testPlan UITestPlan
```

esteluk:
Adds a configuration option that specifies a testplan to execute Scan
with. As with test parallelisation, using this can cause a few problems
with other integrations (such as xcpretty's ability to parse the test
output). This change isn't intended as a holistic change to address
incompatible combinations of options when using the new test report
formats, just a simple option to pass in a command line option.

### Checklist

### Motivation and Context
Xcode 11 introduced testplans as a new way to specify groups of tests that can be run together independently of the project's schemes. eg. I can specify separate test plans for my pre-merge pipeline and nightly builds. 

Per https://github.com/fastlane/fastlane/issues/15462 there's clearly some desire for Fastlane to support test plans properly. At the moment the any use of them with scan requires passing a custom `xcarg` configuration. 

### Description
This change just adds a simple Scan configuration option that gets passed into `xcodebuild` as a command line option. 

closes #15462

### DIFF
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -124,6 +124,11 @@ module Scan
                                      end),
 
         # other test options
+        FastlaneCore::ConfigItem.new(key: :testPlan,
+                                     env_name: "SCAN_TESTPLAN",
+                                     description: "The testplan associated with the scheme that should be used for testing",
+                                     is_string: true,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :xctestrun,
                                      short_option: "-X",
                                      env_name: "SCAN_XCTESTRUN",
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -124,7 +124,7 @@ module Scan
                                      end),
 
         # other test options
-        FastlaneCore::ConfigItem.new(key: :testPlan,
+        FastlaneCore::ConfigItem.new(key: :testplan,
                                      env_name: "SCAN_TESTPLAN",
                                      description: "The testplan associated with the scheme that should be used for testing",
                                      is_string: true,
--- a/scan/lib/scan/test_command_generator.rb
+++ b/scan/lib/scan/test_command_generator.rb
@@ -45,7 +45,7 @@ module Scan
       options << "-enableCodeCoverage #{config[:code_coverage] ? 'YES' : 'NO'}" unless config[:code_coverage].nil?
       options << "-enableAddressSanitizer #{config[:address_sanitizer] ? 'YES' : 'NO'}" unless config[:address_sanitizer].nil?
       options << "-enableThreadSanitizer #{config[:thread_sanitizer] ? 'YES' : 'NO'}" unless config[:thread_sanitizer].nil?
-      options << "-testPlan #{config[:testPlan]}" if config[:testPlan]
+      options << "-testPlan #{config[:testplan]}" if config[:testplan]
       options << "-xctestrun '#{config[:xctestrun]}'" if config[:xctestrun]
       options << config[:xcargs] if config[:xcargs]
 
--- a/scan/lib/scan/test_command_generator.rb
+++ b/scan/lib/scan/test_command_generator.rb
@@ -45,7 +45,9 @@ module Scan
       options << "-enableCodeCoverage #{config[:code_coverage] ? 'YES' : 'NO'}" unless config[:code_coverage].nil?
       options << "-enableAddressSanitizer #{config[:address_sanitizer] ? 'YES' : 'NO'}" unless config[:address_sanitizer].nil?
       options << "-enableThreadSanitizer #{config[:thread_sanitizer] ? 'YES' : 'NO'}" unless config[:thread_sanitizer].nil?
-      options << "-testPlan #{config[:testplan]}" if config[:testplan]
+      if FastlaneCore::Helper.xcode_at_least?(11)
+        options << "-testPlan #{config[:testplan]}" if config[:testplan]
+      end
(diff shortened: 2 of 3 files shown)


scan/lib/scan/options.rb
```diff
@@ -124,6 +124,11 @@ def self.available_options
                                      end),
 
         # other test options
+        FastlaneCore::ConfigItem.new(key: :testPlan,
```
janpio:
```suggestion
        FastlaneCore::ConfigItem.new(key: :testplan,
```

Any reason for the non common capitalization? For Ruby (how we use it) we would probably expect either `testplan` or `test_plan`.

janpio: Ah, I see the Xcode option is like this. I think we should not use that pattern for fastlane code though.

scan/lib/scan/test_command_generator.rb
```diff
@@ -45,6 +45,7 @@ def options # rubocop:disable Metrics/PerceivedComplexity
       options << "-enableCodeCoverage #{config[:code_coverage] ? 'YES' : 'NO'}" unless config[:code_coverage].nil?
       options << "-enableAddressSanitizer #{config[:address_sanitizer] ? 'YES' : 'NO'}" unless config[:address_sanitizer].nil?
       options << "-enableThreadSanitizer #{config[:thread_sanitizer] ? 'YES' : 'NO'}" unless config[:thread_sanitizer].nil?
+      options << "-testPlan #{config[:testPlan]}" if config[:testPlan]
```
janpio:
```suggestion
      options << "-testPlan #{config[:testPlan]}" if config[:testplan]
```

scan/spec/test_command_generator_spec.rb
```diff
@@ -613,5 +613,19 @@
                                      ])
       end
     end
+
+    describe "Specifying a test plan" do
+      before do
+        options = { project: "./scan/examples/standard/app.xcodeproj", testPlan: "simple" }
```
janpio:
```suggestion
        options = { project: "./scan/examples/standard/app.xcodeproj", testplan: "simple" }
```

scan/lib/scan/options.rb
esteluk: Thanks for the review :)

Yeah I took it initially from the Xcode parameter (even though `testplan` reads a little better to me). Definitely makes sense to prefer consistency in the Fastlane/ruby syntax instead.

janpio (approved):

iainsmith:
CI size-label looks like it's failing due to a missing GITHUB_TOKEN.

janpio:
@iainsmith It's failing because it can't handle forks - we will remove that action again soon. For now here this can be ignored.

iainsmith (approved):

lyndsey-ferguson (approved):

scan/lib/scan/test_command_generator.rb
```diff
@@ -45,6 +45,7 @@ def options # rubocop:disable Metrics/PerceivedComplexity
       options << "-enableCodeCoverage #{config[:code_coverage] ? 'YES' : 'NO'}" unless config[:code_coverage].nil?
       options << "-enableAddressSanitizer #{config[:address_sanitizer] ? 'YES' : 'NO'}" unless config[:address_sanitizer].nil?
       options << "-enableThreadSanitizer #{config[:thread_sanitizer] ? 'YES' : 'NO'}" unless config[:thread_sanitizer].nil?
+      options << "-testPlan #{config[:testplan]}" if config[:testplan]
```
lyndsey-ferguson: Does this work with Xcode 10, or we don't care about older versions?

iainsmith: IMO, this should be wrapped in the `if FastlaneCore::Helper.xcode_at_least?(11)` block.

spadafiva:
Hey all, maybe this isn't the right place to ask, but is there a potential timeline on when this action might. be available in fastlane?  Our team would really love to use test plans instead of rolling our own solution for creating the test plans with ruby scripts.

scan/lib/scan/test_command_generator.rb
esteluk: Makes sense!

iainsmith:
> Our team would really love to use test plans instead of rolling our own solution for creating the test plans with ruby scripts.

@spadafiva you can use ‘xcargs: “-testPlan plan-name”’ today

spadafiva:
@iainsmith Thanks!  I appreciate the help.

iainsmith:
cc @joshdholtz is anything else needed to merge this?

joshdholtz (approved):
My bad! I totally thought I approved and merged this one already 😔  Sorry about that! Really appreciate the contribution ❤️

Sherlouk:
Hey @joshdholtz - noticed a very small issue with this MR which I've fixed in #16205. Hope that's okay!

joshdholtz:
Of course! I’m just waking up but I’ll take a look once I get some coffee in me 😊

Issue #15438, opened 2019-10-04
Why is Xcode 11 resultBundlePath `.xcresult` symlinked?


crsantos:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject

This is mostly a question related with `Scan`

#### Question Description

As of Xcode 11, the `.xcreport` format changed and in order to generate code coverage, we need to run the following :

```bash
xcrun xccov view --report --json Example.xcresult
```

I was doing some tests to generate a `json` file with coverage data but I need the final output path that `Scan` defines as `test_result`, so I was checking if I can use `Scan.cache[:result_bundle_path]` (not even sure if that's reliable/will change, but I found no other way).

It's something like; `./fastlane/test_output/MyLibName.test_result`

The thing is, that cached value is a **symlink**, to the real `./fastlane/test_output/MyLibName.test_result.xcresult`.

**Problem**

It's impossible to run `xcrun xccov view --report` against that symlink, or against something that doesn't have the extension `.xcresult`, the xcov tool triggers an error saying: 

`Error: unrecognized file format`.

**So my 2 questions are:**

- Is Scan that is doing this symlink? I couldn't find a symlink being created inside fast lane Scan src-code ([but maybe I didn't dig a lot](https://github.com/fastlane/fastlane/blob/155c5506b9625c1397433fe6b45b67ca7463a64f/scan/lib/scan/test_command_generator.rb))
    - If it isn't `Scan`,  does anyone have an idea how is that symlinking? Because on DerivedData there's no such symlink. It's a regular `.xcresult` bundle.
- Do I have another way of retrieving the output path of that `.xcresult`, the bundle itself?
     -  If I had the previous dir, maybe I can do some glob pattern match to find the file (like `../*.result`) but I have no guarantee that's the one that contains the last test data for that test run.

Any ideas are welcome.

### `fastlane env` output

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14.6                                     |
| Ruby                        | 2.5.3                                       |
| Bundler?                    | true                                        |
[...]

PR #16044, opened 2020-02-19
[scan] Set an extension of resultBundlePath as ".xcresult" when using Xcode 11

att55:
### Checklist

### Motivation and Context

Since Xcode 11, Result Bundle is implemented.

https://developer.apple.com/documentation/xcode_release_notes/xcode_11_release_notes?language=objc

```
The format of result bundles changed in Xcode 11. 
A result bundle is an asset produced by Xcode 11 with the xcresult file extension that contains information about the build, tests, code coverage, and more. 
```

To open Result Bundle by Xcode 11, the extension must be ".xcresult".
But in fastlane,  the extension is always set ".test_result" when a scan action receives the param `result_bundle:true`.

So, I fixed that the extension is set ".xcresult" when using Xcode11.

### Description

I also discover that a symbolic link file is generated  when an extension of Result Bundle is set other than ".xcresult".

The related issue is below.
https://github.com/fastlane/fastlane/issues/15438

I think in order to resolve like this issue, the extension should be ".xcresult" when using Xcode 11.

### DIFF
--- a/scan/lib/scan/test_command_generator.rb
+++ b/scan/lib/scan/test_command_generator.rb
@@ -149,11 +149,8 @@ module Scan
 
     def result_bundle_path
       unless Scan.cache[:result_bundle_path]
-        if FastlaneCore::Helper.xcode_version.to_i >= 11
-          path = File.join(Scan.config[:output_directory], Scan.config[:scheme]) + ".xcresult"
-        else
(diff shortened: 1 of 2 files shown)


janpio (approved):

joshdholtz (approved):
Oh my goodness! Not sure how this wasn't found sooner 😊 Thanks for fixing!

PR #16072, opened 2020-02-23
Improve rubyzip security behavior to be able to continue using 1.3.0

janpio:
As per https://github.com/rubyzip/rubyzip/releases/tag/v2.0.0 rubyzip 2.x changed the default behavior of `validate_entry_sizes` to `true` to improve security:

image

As we can not upgrade to rubyzip 2.x as we want to continue older Ruby versions that are unfortunately bundled with older macOS versions, I applied the suggested "workaround" from the release notes in all (1) locations where we actually use `Zip`:

> If you are using an older version of ruby and can't yet upgrade to 2.x, you can still use 1.3.0 and set the option to `true`.

closes #15970


sylvester-loreto-mypulse:
I think this is being addressed in this pull request https://github.com/fastlane/fastlane/pull/15970

joshdholtz (approved):
Seems good to me!

Issue #16065, opened 2020-02-21
How to generate only selective screenshots with Snapshot


grzegorzkrukowski:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
I have a question about Snapshot tool.
We have a working setup and we generate screenshots on external servers.
It all works, but with amount of screenshots we need to generate for QA purposes, we are looking for an option to generate just selective tests.

We have:
```
@interface SnapshotUITests : XCTestCase
- (void)testPurchaseFlow;
- (void)testOnboarding;
- (void)testTrainingFlow;
```

From what we found in documentation, snapshot only allows to run all tests from given XCTestCase.

`xcodebuild` itself is allowing to use `-only-testing:TestBundle/TestSuite/TestCase` as parameter to run single test only from all set of tests.

Is it possible to pass and use this parameter via Snapshot ?

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 10.15.3                                                        |
| Ruby                        | 2.6.3                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.21.1 (Apple Git-122.3)                           |
| Installation Source         | /usr/local/bin/fastlane                                        |
| Host                        | Mac OS X 10.15.3 (19D76)                                       |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib |
| OpenSSL Version             | LibreSSL 2.8.3                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version               | 11.3.1                                                         |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE |             |   |

### fastlane gems

| Gem      | Version | Update-Status |
| -------- | ------- | ------------- |
[...]

PR #16074, opened 2020-02-24
Added support for only-testing, skip-testing for Snapshot

grzegorzkrukowski:
### Checklist

### Motivation and Context

It adds support for `only-testing` and `skip-testing` parameters to `Snapshot` similarly to `Scan`.
It allows to selectively run screenshots for selective set of UITests, instead of running all of them each time.

https://github.com/fastlane/fastlane/issues/16065

### Description

I added support for mentioned parameters to `Snapshot`.
Also I moved some common method that are now used by `Snapshot` and `Scan` into `FastlaneCore::Helper` and updated `Scan` code to use new common methods.
All required test are updated. I also added unittests for Helper methods.
`detect_values_spec` was added to `Snapshot` as it was not existing before.

All changes are covered with tests in proper specs. 

### Testing Steps

All covered with unit tests.


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -376,6 +376,16 @@ def self.json_file?(filename)
...
+    # splitting on comma allows us to support comma-separated lists of values
+    # from the command line, even though the ConfigItem is not defined as an
+    # Array type
+    def self.coerce_to_array_of_strings(value)
+      return if value.nil?
+
+      value = value.split(',') unless value.kind_of?(Array)
+      return value.map(&:to_s)
+    end
+
```
joshdholtz: I don't think we need this added here? 🤔 This is already a thing if you define `type: Array` on the config item.

scan/lib/scan/detect_values.rb
```diff
@@ -48,6 +48,10 @@ def self.set_additional_default_values
       return config
     end
 
+    def self.coerce_to_array_of_strings(config_key)
+      Scan.config[config_key] = FastlaneCore::Helper.coerce_to_array_of_strings(Scan.config[config_key])
+    end
+
```
joshdholtz: Should also be able to remove this

fastlane_core/spec/helper_spec.rb
```diff
@@ -159,5 +159,21 @@
...
+
+      it "leaves value as an array", requires_xcodebuild: true do
+        array_value = ["StringValue1", "StringValue2"]
+        expect(FastlaneCore::Helper.coerce_to_array_of_strings(arrayValue)).to eq(array_value)
+      end
+
+      it "handles nil", requires_xcodebuild: true do
+        expect(FastlaneCore::Helper.coerce_to_array_of_strings(nil)).to be(nil)
+      end
+    end
```
joshdholtz: Should also be able to remove this

snapshot/lib/snapshot/options.rb
```diff
@@ -232,7 +237,23 @@ def self.available_options
                                      env_name: "SNAPSHOT_CLONED_SOURCE_PACKAGES_PATH",
                                      description: "Sets a custom path for Swift Package Manager dependencies",
                                      type: String,
-                                     optional: true)
+                                     optional: true),
+        FastlaneCore::ConfigItem.new(key: :only_testing,
+                                    env_name: "SNAPSHOT_ONLY_TESTING",
+                                    description: "Array of strings matching Test Bundle/Test Suite/Test Cases to run",
+                                    optional: true,
+                                    is_string: false,
```
joshdholtz: Change this to `type: Array`

snapshot/lib/snapshot/options.rb
```diff
@@ -232,7 +237,23 @@ def self.available_options
...
+                                    verify_block: proc do |value|
+                                      verify_type('only_testing', [Array, String], value)
+                                    end),
+        FastlaneCore::ConfigItem.new(key: :skip_testing,
+                                    env_name: "SNAPSHOT_SKIP_TESTING",
+                                    description: "Array of strings matching Test Bundle/Test Suite/Test Cases to skip",
```
joshdholtz: Change this to `type: Array`

snapshot/lib/snapshot/options.rb
```diff
@@ -232,7 +237,23 @@ def self.available_options
                                      env_name: "SNAPSHOT_CLONED_SOURCE_PACKAGES_PATH",
                                      description: "Sets a custom path for Swift Package Manager dependencies",
                                      type: String,
-                                     optional: true)
+                                     optional: true),
+        FastlaneCore::ConfigItem.new(key: :only_testing,
```
joshdholtz: We may want to think of a different key besides `only_testing` 🤔 A name like this kind of implies a boolean so we might want to have the key name be a little more explicit for what it actually wants.

snapshot/lib/snapshot/options.rb
```diff
@@ -232,7 +237,23 @@ def self.available_options
...
+                                    env_name: "SNAPSHOT_ONLY_TESTING",
+                                    description: "Array of strings matching Test Bundle/Test Suite/Test Cases to run",
+                                    optional: true,
+                                    is_string: false,
+                                    verify_block: proc do |value|
+                                      verify_type('only_testing', [Array, String], value)
+                                    end),
+        FastlaneCore::ConfigItem.new(key: :skip_testing,
```
joshdholtz: Same here with the name 😊

joshdholtz (changes requested):
Thanks for adding this 😊 Left some requests and thoughts! I didn't run this yet but just did an initial review. Let me know your thoughts on my thoughts ❤️

fastlane_core/lib/fastlane_core/helper.rb
grzegorzkrukowski: The thing here is we would like to support both: `String` and `Array` type here dynamically:

https://github.com/fastlane/fastlane/pull/16074/files#diff-d60df02a10f3258c3bcf73b71e00bd9fR7

https://github.com/fastlane/fastlane/pull/16074/files#diff-d60df02a10f3258c3bcf73b71e00bd9fR16

If you think it's better to always force users to use `Array` of `Strings` here even for single test being passed - then your whole feedback makes sense.

snapshot/lib/snapshot/options.rb
grzegorzkrukowski: Good point - I didn't think of that as original naming comes from Apple directly and `xcodebuild` command:
https://developer.apple.com/library/archive/technotes/tn2339/_index.html

So just to be consistent with naming used by Apple - I used exactly same parameter names.
But I agree that something like `tests_to_run` `tests_to_skip` would be much more self-explanatory compared to their naming. But question is what's better - consistency with parent commands or self-explanatory parameter names. So far I saw fastlane is using 1:1 naming with `xcodebuild` so it's easier to identify them.

joshdholtz: I would probably say consistency with _fastlane_ naming in this case. IMO, most users aren't familiar with the options in the commands that _fastlane_ actions get piped too (especially with xcodebuild since there are a lot of them there 😊 )

fastlane_core/lib/fastlane_core/helper.rb
joshdholtz: Mixed types get weird with our `Fastlane.swift` interfaces which uses types pretty heavily and it makes validation a little bit weird. We are slowly moving towards getting all the options in all the actions/tools to have a defined `type: ` 😊 We have could keep this as `Array` that would be 💯

grzegorzkrukowski: Ok understood ! will turn all this into `Array`s

joshdholtz: Thank you! ❤️

snapshot/lib/snapshot/options.rb
joshdholtz: Much appreciated ❤️

grzegorzkrukowski:
Hey @joshdholtz 

I have applied your feedback - but I went all the way from start.
Old method with mixed types and `coerce_to_array_of_strings` was implemented in `scan` before.

Due to backwards compatibility I didn't want to modify that part.

I just implemented proper methods in `snapshot` instead - and I left `scan` with it's original implementation. Probably some people still using it, so replacing it with different parameter type may cause lot's of problems.

What I could do is to implement `tests_to_run` and `tests_to_skip` for `scan` similarly - next to existing solution - and we could deprecate old implementation that allows to pass either `Array` or `String` soon. What do you think ?

Let me also know if naming is fine for you.

joshdholtz:
Ahhh, it looks like there were two PRs that added the same thing and I ended up merging them in the wrong order 😔 This was added in #16312. My apologies! I really appreciate the time spent on this ❤️

PR #16089, opened 2020-02-27
[swift] Fix `submissionInformation` data type in `DeliverfileProtocol`

Woollim:
### Checklist

### Motivation and Context
Currently, `submissionInformation` data type is `String` in `DeliverfileProtocol`
However, if `submissionInformation` is returned in String, `deliver` action fails in app review submission

### Description
I changed `submissionInformation` data type to `[String: Any?]` which is `submissionInformation` data type in Ruby Fastlane

After changed, `delivery` action completed the application review submission.

Changed submissionInformation

submissionInformation Result

### DIFF
--- a/fastlane/swift/DeliverfileProtocol.swift
+++ b/fastlane/swift/DeliverfileProtocol.swift
@@ -76,7 +76,9 @@ protocol DeliverfileProtocol: class {
   var appRatingConfigPath: String? { get }
 
   /// Extra information for the submission (e.g. compliance specifications, IDFA settings)
-  var submissionInformation: String? { get }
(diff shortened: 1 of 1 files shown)


janpio:
As far as I know (and that is very little for Swift!) this file is generated and would be overwritten on the next release. The actual changes need to happen in the ruby files.

rogerluan:
@janpio is right. This needs to be fixed in the ruby files probably, as these swift files are autogenerated and will be overwritten.

But I'm facing the same issue as you @Woollim , the type should be a dictionary, not a string 😞

janpio (changes requested):
See above

Issue #13356, opened 2018-09-13
the Deliver build_number parameter does not uniquely specify which build to submit for review 


JimHowell20:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

According to the deliver docs, you can:

> Select a previously uploaded build and submit it for review.
> 
> fastlane deliver submit_build --build_number 830
> 

The issue with this is that the build number is only unique per version number.  So if you upload the following builds for example:

Version Number: 1.2.33
Build Number: 300

Version Number: 1.2.32
Build Number: 300

Version Number: 1.2.31
Build Number: 300

and then try to submit a previously uploaded build using:

 fastlane deliver submit_build --build_number 300

Then fastlane will not chose the latest build in this case.

This would be resolved if you could specify the "version_number" (CFBundleShortVersionString)  as well as the "build_number" (CFBundleVersion) in order to uniquely identify the build you're trying to submit for review.

StefanoBiasu:
> This would be resolved if you could specify the "version_number" (CFBundleShortVersionString) as well as the "build_number" (CFBundleVersion) in order to uniquely identify the build you're trying to submit for review.

Apple shouldn't let you upload more than a single version of your app. You define beforehand the version to put to store, then you upload the ipa with the correct version, and the build number is the only discriminator between your uploads.

JimHowell20:
The app store does allow you to upload multiple build versions with the same build number.   The build number only needs to be unique per build version.  There is no requirement to only upload a single build version before release.   The app submission metadata can only have a single version specified if that's what you mean, but that's not tied to the build version.  See the screenshot below:
<img width="811" alt="screen shot 2018-10-04 at 1 30 30 pm" src="https://user-images.githubusercontent.com/7559768/46491946-33c2c300-c7da-11e8-9073-8c3cfaadec5b.png">

syedsmith:
is this issue solved now?

PR #16091, opened 2020-02-27
[deliver] Add fix for app_version and build_number matching

onurpolattimur:
### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/issues/13356
If we use build_number and app_version action in deliver, it selects the build of any version it finds. It does not take the app_version into consideration.

For example given lane:
```ruby
lane :submit_review do
  deliver(
    build_number: '12',
    submit_for_review: true,
    app_version: '2.0'
  )
end
```
If there are many app versions that have build number 12, it will select the build **12** from random app version instead of **2.0.**

### Description

In select_build method, when app_version number is given, it will find the build corresponding to it.
I tested these changes manually with my app.

### DIFF
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -36,7 +36,11 @@ module Deliver
 
       if options[:build_number] && options[:build_number] != "latest"
         UI.message("Selecting existing build-number: #{options[:build_number]}")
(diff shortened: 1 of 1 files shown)


onurpolattimur:
@googlebot I signed it!

onurpolattimur:
@janpio

joshdholtz (approved):
This is great! I’m not sure why it wasn’t like this before 😊 Thanks for contribution ❤️

PR #16096, opened 2020-02-29
[fastlane] load additional .env files from within CLIToolsDistributor

### LINKED ISSUE #16077: --env argument is only accepted when running a lane
### New Issue Checklist

### Issue Description
This isn't really a bug report, since I think that it's potentially implemented as intended however today i noticed that using `fastlane run my_action --env jenkins` results in the following error:

> invalid option: --env

We've just started using dotenv to organise all the environment variables along with overrides that are specific for different environments as described here: https://docs.fastlane.tools/best-practices/keys/#dotenv.

Usually this isn't a problem since we run a lane directly, but we have a custom action for various tasks and sometimes I need to test them locally via the `run` command, this is where I hit the roadblock.

As a temporary work around, I can just wrap the action in a lane however I want to know what the expected behavior is and potentially correct it if somebody agrees.

First time .env files are loaded is in **cli_tools_distributor.rb**, right at the very beginning

https://github.com/fastlane/fastlane/blob/099fd43ecc70f862e5346b61c80d89f8ec61038b/fastlane/lib/fastlane/cli_tools_distributor.rb#L62-L66

The reasoning in the comment makes sense, we'd like to hide non-essential output on the CI so I would like to define `FASTLANE_HIDE_CHANGELOG` in my `.env.jenkins` file, however because we pass `nil` into this method, only variables from `.env` and `.env.default` are loaded at this point.

Its then not until **lane_manager.rb** where we call `load_env_vars` again but this time, we include the value of `--env` passed in via the command line arguments:

https://github.com/fastlane/fastlane/blob/0c6a6d24098b9f6d7b16e4328a2048c6da4fa8df/fastlane/lib/fastlane/lane_manager.rb#L51

The above explains why I'm not able to pass `--env` when using `fastlane run [action]`, but my question is: is this expected?

Its not immediately obvious to me why we don't make `--env` as a `global_option` like `--verbose` is? 

Screenshot 2020-02-25 at 12 46 58

https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/fastlane/lib/fastlane/commands_generator.rb#L102

Sorry for the strange format of this ticket... its more of a question than a bug report... if one of the contributors can help me out here, I'm happy to work on changing the behaviour but I'm guessing that there is more to consider

Thanks fro all the hard work 🙏

liamnichols:
### Checklist

### Motivation and Context

Closes #16077 

As I raised in the linked ticket, currently the additional .env files specified with `--env` are only supported when running a lane directly. This means that you cannot run any specific tool or action and be able to pass the `--env` argument without wrapping it in a lane.

This behaviour is odd, since the **.env** and **.env.default** files however are supported.

I work on a large project that has five different configurations that we can build and release that all use different code signing and bundle identifiers. We use XcodeGen to generate the project and rely heavily on injecting environment variables to define the differing build settings.

To support the above project, we use custom **.env** files to overload the environment variables for the given configuration... For example, `fastlane build --env internal`. This is great, however to share the environment with XcodeGen, we have a custom action and to avoid repetition between our CI setup and local setups, we invoke this locally to generate our project as well. 

Since `fastlane run my_action` doesn't support the `--env` argument however, we're unable to generate projects for different environments locally without wrapping the action in a dummy lane like such:

```ruby
lane :generate do
    my_action
end
```

### Description

The changes in this PR intend of supporting the use of the --env argument universally within the Fastlane cli however this was a bit trickier than I was initially expecting... `CLIToolsDistributor` is a wrapper that basically works out which Commander interface needs to be invoked and as a result, it doesn't have access to any of Commanders nice argument parsing. 

I was able to work around this using the simple `env_cl_param` lambda in the `load_dot_env` method that I've introduced on `CLIToolsDistributor` however I don't think its very elegant and would appreciate some input 🙏 

The general gist however is that we strip out `--env blah` from `ARGV` here before invoking any of the defined tools, main command interface or credentials manager. 

To keep the visibility of this option, I've then gone through every `CommandsGenerator` class and listed `--env` as a global_option, kind of following how `--verbose` seems to be handled. I'm not sure if this is the right approach though, since there is a bit of a disconnect between where we parse `--env` and where its defined on the command interface... Will appreciate some input here also.

~Additionally I've left three TODO's in the code that I need some help answering.~ (answered now, thanks @joshdholtz) 

### Testing Steps

Amongst the changes, I've both removed old tests and introduced new ones... I added a comprehensive suite of tests to `DotenvHelper` since its functionality was previously being tested as part of the *fastlane/spec/commands_generator_spec.rb** and **fastlane/spec/lane_manager.rb** however since we have moved the loading up a level, this is no longer the correct place to test. 

I also added tests to **cli_tools_distributor_spec.rb** however instead of checking that the `ENV` object is loaded as expected, the tests focus on checking that the `DotenvHelper` is invoked correctly.

I realise that these changes have kind of focused on testing individual components now and I haven't really added any kind of test to verify that the entire process works as a whole since I wasn't really sure how to approach it.. I was thinking that maybe it'll be good to test somewhere that invoking `fastlane test_lane` would result in an output that prints `ENV['SOME_VAR']` where the value is expected based on the **.env** setup... I'm not sure if this is a bit much or not so again I need some guidance here 🙇 

Thanks again for the great tooling, it's really helping our team and I'm enjoying pushing it to the limits and getting stuck in trying to improve things for everybody 🚀


joshdholtz:
I think I'm going to like this PR 😊

fastlane/lib/fastlane/lane_manager.rb
```diff
@@ -46,10 +46,9 @@ def self.cruise_lane(platform, lane, parameters = nil, env = nil, fastfile_path
       # and not return to the original working directory
       # https://github.com/CocoaPods/Xcodeproj/issues/426
       # Setting this environment variable causes xcodeproj to work around the problem
+      # TODO: Does this need moving into CLIToolsDistributor to be called **before** loading Dotenv?
       ENV["FORK_XCODE_WRITING"] = "true" unless platform == 'android'
```
liamnichols: I _think_ its safe for us to remove this now..

I traced it back to #6278, which is also discussed in https://github.com/CocoaPods/Xcodeproj/issues/426#issuecomment-251804075 where they talk about it being resolved in 1.4.0 of Xcodeproj...

If that's the case, since the gemspec is currently requires >= 1.13.0 meaning that the bug is no longer a thing... 

https://github.com/fastlane/fastlane/blob/204fab2f681dad3a60314844e8e062b19cc1bf28/fastlane.gemspec#L60

WDYT?

joshdholtz: Seems safe to remove!

fastlane/lib/fastlane/cli_tools_distributor.rb
```diff
@@ -62,8 +62,8 @@ def take_off
         # Loading any .env files before any lanes are called since
         # variables like FASTLANE_HIDE_CHANGELOG and FASTLANE_DISABLE_COLORS
         # need to be set early on in execution
-        require 'fastlane/helper/dotenv_helper'
-        Fastlane::Helper::DotenvHelper.load_dot_env(nil)
+        # TODO: Should load_dot_env be moved to the top of take_off in order to support FASTLANE_DISABLE_ANIMATION preference?
+        load_dot_env
```
liamnichols: `load_dot_env` will internally call `require 'fastlane/helper/dotenv_helper'`

If I were to move this call up to the top of `take_off`, this would mean that we would require `fastlane/helper/dotenv_helper` before the long running `require 'fastlane'` call... Would that cause a problem? If so, I guess I should leave this call where it is.

fastlane/lib/fastlane/cli_tools_distributor.rb
```diff
@@ -122,6 +122,28 @@ def take_off
...
+          index = ARGV.index("--env")
+          return nil if index.nil?
+          ARGV.delete_at(index)
+
+          return nil if ARGV[index].nil?
+          value = ARGV[index]
+          ARGV.delete_at(index)
+
+          value
+        end
```
liamnichols: The aim of `env_cl_param` is something like the following:

```ruby
ARGV = ['foo', 'bar', '--env', 'baz', '--verbose']

param = env_cl_param

param # => baz
ARGV # => ['foo', 'bar', '--verbose']
```

I have no idea how to do it nicer though... I think there _must_ be a better way but I can't figure it out

fastlane/lib/fastlane/cli_tools_distributor.rb
joshdholtz: I _think_ this needs to stay after `require 'fastlane'` 🤔 Trial/error might prove me wrong though 🙃

fastlane/lib/fastlane/cli_tools_distributor.rb
joshdholtz: Are you popping it off because the CLIs the tools provide define `--env` on their own?

liamnichols: @joshdholtz: originally, I **had** to pop it off since not all of the Commander CLIs declared the `global_option("--env")` and would error if I it wasn’t popped from `ARGV`.

Although I’ve since declared the global_option on every CLI, I _could_ leave it in `ARGV`, but the thought was that if a new CLI was added that forgot to declare it, it would cause the “invalid option --env” error despite the fact that that we aren’t expecting the CLI itself to use it.

It’s a bit weird because I pop it, and process it from `CLIToolsDistributor` but have to declare the global_option on each CLI purely for it to show up in `--help`.

Think it will be better to leave it in `ARGV`, or do you have any other ideas?

fastlane/lib/fastlane/cli_tools_distributor.rb
liamnichols: Thanks for the info.. Originally I was worried that I wouldn't be able to tell if it worked or not, but it was pretty obvious in the end 😂 

If I try to move it above `require 'fastlane'` the method completely explodes so thats a pretty good sign that it's a no-go 😅 

I have however opted to move it above the locale check and `print_bundle_exec_warning` so that `LANG`, `LC_ALL` and `SKIP_SLOW_FASTLANE_WARNING` can be defined in external .env files. 

My understanding from the [docs](https://docs.fastlane.tools/getting-started/ios/setup/#set-up-environment-variables) is that UTF8 is required by things like xcodebuild and not fastlane/ruby itself so specifying something like `LANG` in a .env file should be fine? I have to define these variables on my Jenkins CI setup already so it'll be ace if I can move them into my **.env.jenkins** file with this change.

joshdholtz (approved):
This is so good! I reviewed the code pretty hard and tested it out quite a bit too on my local machine 😊 Thanks for letting dotenv power more of fastlane! Really appreciate the contribution ❤️

janpio:
This might have caused a regression: https://github.com/fastlane/fastlane/issues/16461 (not enough information available yet)

PR #16100, opened 2020-02-29
Fix frameit rotation bug

QYL:
### Checklist

### Description

The original rotation degree is wrong.
The home button is on the bottom in portrait frame. Hence, when :landscape_right(home button on the right side) is set the frame should rotate -90 (counterclockwise), and 90(clockwise) degree when :landscape_left is set.

Run the code

### DIFF
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -60,8 +60,8 @@ module Frameit
     end
 
     def rotation_for_device_orientation
-      return 90 if self.screenshot.landscape_right?
-      return -90 if self.screenshot.landscape_left?
+      return -90 if self.screenshot.landscape_right?
+      return 90 if self.screenshot.landscape_left?
       return 0
     end
 


QYL:
@googlebot I signed it!

janpio:
Are there any issue about this behavior?

QYL:
> Are there any issue about this behavior?

The original behavior does the opposite as expected.

janpio:
This code has been in there for ages, exactly this way. If this really was broken all this time, I would expect there to be some GitHub issues referring to the problem. Did maybe something else change?

QYL:
I just found out that there are two type of landscapeleft, one is the property of [UIDeviceOrientation](https://developer.apple.com/documentation/uikit/uideviceorientation/landscapeleft) which means home button on the right side, and another is of [UIInterfaceOrientation](https://developer.apple.com/documentation/uikit/uiinterfaceorientation/landscapeleft) which means the Home button on the left side. While frameit uses UIInterfaceOrientation.landscapeleft, I took screenshots by adopting UIDeviceOrientation.landscapeleft. Maybe that's the cause. I'm closing this pull request, sorry for the troubles caused.

janpio:
No trouble caused at all - if this is indeed the cause for the problem I am happy. If not, feel free to open this PR again of course!

PR #16123, opened 2020-03-05
[core] Suppress output of PlistBuddy when disabling "KeyboardContinuousPathEnabled"

robnadin:
### Checklist

### Motivation and Context

Suppress the error output from PlistBuddy when the `KeyboardContinuousPathEnabled` key already exists in the simulator's preferences. On some CI systems this manifests as an error in the build log:
```
Add: ":KeyboardContinuousPathEnabled" Entry Already Exists
```

### DIFF
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -223,7 +223,7 @@ module FastlaneCore
         plist_buddy_cmd = "-c \"Add :KeyboardContinuousPathEnabled bool false\""
         plist_path = File.expand_path("~/Library/Developer/CoreSimulator/Devices/#{self.udid}/data/Library/Preferences/com.apple.keyboard.ContinuousPath.plist")
 
-        Helper.backticks("#{plist_buddy} #{plist_buddy_cmd} #{plist_path}")
+        Helper.backticks("#{plist_buddy} #{plist_buddy_cmd} #{plist_path} >/dev/null 2>&1")
       end
     end
   end


fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -223,7 +223,7 @@ def disable_slide_to_type
         plist_buddy_cmd = "-c \"Add :KeyboardContinuousPathEnabled bool false\""
         plist_path = File.expand_path("~/Library/Developer/CoreSimulator/Devices/#{self.udid}/data/Library/Preferences/com.apple.keyboard.ContinuousPath.plist")
 
-        Helper.backticks("#{plist_buddy} #{plist_buddy_cmd} #{plist_path}")
+        Helper.backticks("#{plist_buddy} #{plist_buddy_cmd} #{plist_path} >/dev/null 2>&1")
```
joshdholtz:
```suggestion
        Helper.backticks("#{plist_buddy} #{plist_buddy_cmd} #{plist_path}", print: false)
```

This _should_ also work if you want to try this 😊 ☝️

joshdholtz (changes requested):
Would you be able to add some more in the description of this PR on the need for it? Was the output too verbose? I don't remember what happens 🙃 

And maybe we could change to using `print: false` on `backticks` way to all hide output? Or is it helpful to see the command that is being executed? `print: false` would hide both the command and the output.

fastlane_core/lib/fastlane_core/device_manager.rb
robnadin: Unfortunately, after trying this, because the output is an error it is still printed 😠

robnadin:
> Would you be able to add some more in the description of this PR on the need for it? Was the output too verbose? I don't remember what happens 🙃
> 
> And maybe we could change to using `print: false` on `backticks` way to all hide output? Or is it helpful to see the command that is being executed? `print: false` would hide both the command and the output.

Updated description with some context.

joshdholtz (approved):
Looks good! Thanks for the contribution ❤️ Really appreciate it!

Issue #16131, opened 2020-03-06
Move AppStoreInfo.plist to build output directory if generated


Ashton-W:
#### Motivation Behind Feature
If the `exportOptions.plist` contains the key `generateAppStoreInformation` set to `true`, xcodebuild will produce a file `AppStoreInfo.plist` in it's output directory. **gym** does not currently move this file to the final output directory, it stays in the temporary directory.
If the user is generating this file, they want it in the output directory.

#### Feature Description
From `xcodebuild -help`:
> 	generateAppStoreInformation : Bool
>
>		For App Store exports, should Xcode generate App Store Information for uploading with iTMSTransporter? Defaults to NO.
It's used for uploading using iTMSTransporter on Windows and Linux. On macOS it can be generated on demand during upload.

The file `AppStoreInfo.plist` is generated in the xcodebuild output directory if the `exportOptions.plist` contains the key `generateAppStoreInformation` set to `true`. **gym** should move this file to the gym output directory, like it does a few other metadata files like App Thinning Report, if present.

#### Alternatives or Workarounds
You can instead use `sh` to run the `swinfo` tool outside of xcodebuild to analyze an already built IPA file. If doing this you should set `generateAppStoreInformation` to `false` because the file takes time to generate. There is no easy way to find the temporary directory gym uses to collect xcodebuild output in your script, but the file is in there as expected.

```ruby
# these options appear to match xcodebuild behaviour, but the file isn't identical because it includes env
sh "xcrun swinfo -f App.ipa -o AppStoreInfo.plist -prettyprint true --plistFormat binary"
```

janpio:
> It's used for uploading using iTMSTransporter on Windows and Linux.

Unrelated: Do you have a link to get more information about iTMSTransporter for Windows and Linux?

Ashton-W:
> Unrelated: Do you have a link to get more information about iTMSTransporter for Windows and Linux?

https://help.apple.com/itc/transporteruserguide/en.lproj/static.html

PR #16133, opened 2020-03-06
[gym] Move AppStoreInfo.plist to build output directory if generated …

Ashton-W:
…(#16131)

### Checklist
- [x] I've updated the documentation if necessary. (I don't think necessary since this is expected behaviour and other similar files aren't documented)

### Motivation and Context

#16131 has all details.

### Description

In the same place we move other files generated when an IPA file is generated, we now also check for `AppStoreInfo.plist` and move it to gyms output directory. I followed how it was done for other similar optional generated files that are enabled by `exportOptions.plist` entries.

I built an app with the following `exportOptions.plist` with `app-store` export method and `generateAppStoreInformation` set to `true`:
```xml
	<key>method</key>
	<string>app-store</string>
	<key>generateAppStoreInformation</key>
	<true/>
```

### Testing Steps

Use gym to export an IPA with an `exportOptions.plist` with the above mentioned keys. In the xcodebuild output directory you'll find an `AppStoreInfo.plist` file, if this works, you'll find it in the gym output directory too. If the `exportOptions.plist` has `generateAppStoreInformation` set to `false` instead, the file won't be generated, so should not exist in the gym output directory.

### DIFF
--- a/gym/lib/gym/generators/package_command_generator.rb
+++ b/gym/lib/gym/generators/package_command_generator.rb
@@ -53,6 +53,10 @@ module Gym
         generator.asset_packs_path
       end
 
+      def appstore_info_path
+        generator.appstore_info_path
+      end
+
       # The generator we need to use for the currently used Xcode version
       # Since we dropped Xcode 6 support, it's just this class, but maybe we'll have
       # new classes in the future
--- a/gym/lib/gym/generators/package_command_generator.rb
+++ b/gym/lib/gym/generators/package_command_generator.rb
@@ -49,10 +49,6 @@ module Gym
         generator.apps_path
       end
 
-      def appstore_info_path
-        generator.appstore_info_path
-      end
-
       def asset_packs_path
         generator.asset_packs_path
       end
--- a/gym/lib/gym/generators/package_command_generator.rb
+++ b/gym/lib/gym/generators/package_command_generator.rb
@@ -49,6 +49,10 @@ module Gym
         generator.apps_path
       end
 
+      def appstore_info_path
+        generator.appstore_info_path
+      end
+
       def asset_packs_path
         generator.asset_packs_path
       end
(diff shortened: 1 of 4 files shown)


Ashton-W:
@googlebot I signed it!

janpio:
Do you think this might be a breaking change for anyone? Not really, right?

Ashton-W:
It could only be a breaking change if you had the option enabled in your `exportOptions.plist` but weren’t using the output, and now that the output is there some scripting breaks, like it relies on that file not being there. In that case, turn the option off and get faster exports if you don’t want the file, or remove whatever workaround you put in place to get the file generated there. 
It seems unlikely, this seems to be a rarely used feature, not many people talking about it online.

janpio (approved):

janpio:
```
Offenses:

gym/lib/gym/runner.rb:341:1: C: Layout/TrailingWhitespace: Trailing whitespace detected.
gym/lib/gym/generators/package_command_generator.rb:55:1: C: Layout/TrailingWhitespace: Trailing whitespace detected.

1145 files inspected, 2 offenses detected

**Comment by Ashton-W** [comment]
Saw that, I don’t have access to a Mac that can access public GitHub right now

**Review by joshdholtz — APPROVED** [review]
This is great! I had no idea this was a feature 😛 I added a commit to add Mac support but other than that this was 💯 Thanks for adding this! Really appreciate the contribution ❤️

**Comment by joshdholtz** [comment]
@googlebot I consent.

**Comment by Ashton-W** [comment]
The CI failure seems unrelated
```
```

Ashton-W:
Thanks @joshdholtz you fixed the rubocop issues too :)

AppStoreInfo.plist is only for iOS and Mac Installer Packages https://help.apple.com/xcode/mac/current/#/deva1f2ab5a2
It looks like that is how you have implemented it 👍

joshdholtz:
@Ashton-W Not a problem! Thanks for doing all the hard work 😊 I had a rebase but this will get merged after that 💪

Issue #14219, opened 2019-02-11
[upload_to_app_store] #<ArgumentError: different prefix: "/" and "fastlane">


nrbrook:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Same issue as https://github.com/fastlane/fastlane/issues/13853
Fastlane configuration is kept in a subdirectory `fastlane` of App root.
Only occurs when using capture_screenshots

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>Didn't work??
INFO [2019-02-11 14:58:34.43]: Loading './fastlane/metadata/review_information/notes.txt'...
ERROR [2019-02-11 14:58:35.63]: #<ArgumentError: different prefix: "/" and "fastlane">
ERROR [2019-02-11 14:58:35.63]: /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/pathname.rb:520:in `relative_path_from'
/Library/Ruby/Gems/2.3.0/gems/fastlane-2.116.0/deliver/lib/deliver/html_generator.rb:39:in `render_relative_path'
(erb):240:in `block (3 levels) in render'
(erb):239:in `each'
(erb):239:in `each_with_index'
(erb):239:in `block (2 levels) in render'
(erb):233:in `each'
(erb):233:in `block in render'
(erb):157:in `each'
(erb):157:in `render'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/erb.rb:864:in `eval'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/erb.rb:864:in `result'
/Library/Ruby/Gems/2.3.0/gems/fastlane-2.116.0/deliver/lib/deliver/html_generator.rb:57:in `render'
/Library/Ruby/Gems/2.3.0/gems/fastlane-2.116.0/deliver/lib/deliver/html_generator.rb:14:in `run'
/Library/Ruby/Gems/2.3.0/gems/fastlane-2.116.0/deliver/lib/deliver/runner.rb:180:in `validate_html'
[...]

Issue #13853, opened 2018-12-07
[upload_to_app_store] #<ArgumentError: different prefix: "/" and "fastlane">


janpio:
When running `deliver` this user got an exception because he was using an absolute path where a relative one was expected:
https://github.com/fastlane/fastlane/issues/12789#issuecomment-445095698
https://github.com/fastlane/fastlane/issues/12789#issuecomment-445357628

PR #16146, opened 2020-03-09
[deliver] Fix crash when metadata_path is outside of the fastlane path

bvirlet:
:key:

When calling deliver when `metadata_path` points outside of `fastlane/` folder, deliver fails with the following error: `#<ArgumentError: different prefix: "/" and "fastlane">`

The issue is that in [html_generator.rb](https://github.com/fastlane/fastlane/blob/5e23f949f4c943f484d48249a39f2662b56f466d/deliver/lib/deliver/html_generator.rb#L39), `path` and `export_path` must be relative to each other. If one of the path is an absolute path (eg. `export_path` is `/var/tmp/something`) and the other one relative (eg. `path` is always `./fastlane`), then `PathName#relative_path_from` fails.

Note: the spec for HtmlGenerator actually reveal the issue as soon as you include `app_icon` in the options: in the specs, the export path is set to a temporary directory, which reproduces the exact above issue.

Likely fixes old auto-closed bugs: #14219, #13853

### Checklist

### DIFF
--- a/deliver/lib/deliver/html_generator.rb
+++ b/deliver/lib/deliver/html_generator.rb
@@ -35,8 +35,8 @@ module Deliver
     # Returns a path relative to FastlaneFolder.path
     # This is needed as the Preview.html file is located inside FastlaneFolder.path
     def render_relative_path(export_path, path)
-      export_path = Pathname.new(export_path)
-      path = Pathname.new(path).relative_path_from(export_path)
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
This is great! Thanks for fixing this 😊 Really appreciate the contribution ❤️

PR #16132, opened 2020-03-06
[swiftlint] add support for the --no-cache flag on autocorrect and lint

kronenthaler:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

This PR allows to pass the --no-cache flag to the autocorrect and lint commands. This helps to solve some weird behavior when formatting the code and the cache becomes stale causing lost of chunks of code.

### Description

Added an option (like --format) to allow passing --no-cache flag.

### Testing Steps

Updated the unit test to append the --no-cache flag if present to lint and autocorrect commands.

PR #16153, opened 2020-03-10
[action] fix a type error in the Swift setupCi action.

IngmarStein:
`provider` should be an optional String, not a Bool. This fixes the following error when using the action:

```
[!] 'provider' value must be a String! Found FalseClass instead.
```

### Checklist

### DIFF
--- a/fastlane/lib/fastlane/actions/setup_ci.rb
+++ b/fastlane/lib/fastlane/actions/setup_ci.rb
@@ -93,7 +93,7 @@ module Fastlane
                                        env_name: "FL_SETUP_CI_PROVIDER",
                                        description: "CI provider. If none is set, the provider is detected automatically",
                                        is_string: true,
-                                       default_value: false,
+                                       optional: true,
                                        verify_block: proc do |value|
                                          value = value.to_s
                                          # Validate both 'travis' and 'circleci' for backwards compatibility, even


janpio (changes requested):
As far as I know `Fastlane.swift` is auto generated - to fix this the underlying ruby action has to be fixed.

IngmarStein:
I missed that `Fastlane.swift` is autogenerated, so I fixed the underlying Ruby action. Looks like the generated file picked up some unrelated changes.

janpio:
Thanks, that looks good now.

Unfortunately I do not know enough about the Swift part to be able to confirm if the unrelated changes are a problem - we'll have to wait for someone else to approve here.

IngmarStein:
Regarding the other changes: noCache was only added in #16132 and it was probably forgotten to generate the Swift file after making the change.
The metadataPath argument doesn't seem to have a default value: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/download_from_play_store.rb#L32

joshdholtz:
@IngmarStein The swift file gets generated automatically so I'm not going to include those changes in this PR but I will look into those discrepancies that you found 🤔

IngmarStein:
@joshdholtz this PR also includes the necessary changes to the Ruby action (`setup_ci.rb`), not just the generated Swift API. Am I missing something?

joshdholtz:
@IngmarStein Yup yup, we'll be keeping the changes in `setup_ci.rb` 😊 We will just remove the changes to the `.swift` files since those are autogenerated

IngmarStein:
@joshdholtz Ah, sure thing. In that case, I've just removed those changes from this PR to make it mergeable.

joshdholtz (approved):
This makes so much sense 😊 Thanks for catching this! Appreciate the contribution ❤️

PR #16162, opened 2020-03-11
[spaceship] Fix 2FA to handle sms fallback for accounts with no trusted devices

jesiegel1:
### Checklist

### Motivation and Context
Issue: #16108 

If a user has 2FA enabled but is not logged into an Apple device with their Apple ID (no trusted devices), auth will default to SMS for verification. For an account with no trusted devices, if a user only has one trusted phone number associated with their account, an SMS verification code will automatically be sent to that number. If they have more than one trusted phone number, an SMS won’t be sent until it’s explicitly requested.

Trusted phone numbers have a separate endpoint, but Spaceship currently posts 2FA codes to the trusted devices endpoint for all cases (except when the default number env var is set). For accounts with no trusted devices, if a user with one trusted number enters the SMS code they automatically received, Spaceship exits with an Unauthorized Error. If the account has more than one trusted number, the user won’t have received a 2FA code at all and will have nothing to enter into the prompt.

The current workaround is to enter `sms` when prompted by Fastlane for the 2FA code which will force Fastlane to have the user select a phone number, request a new SMS code, and post the code to the correct endpoint.

### Description
Checks for `"noTrustedDevices"` field to determine SMS fallback, adds logic to handle the different possible cases, one number/multiple numbers, default number env var set/not set.

Also added new fixtures and a bunch of tests for handling the different verification paths, hitting the right endpoints, and just filling out coverage for the 2FA flow in general.

### Testing Steps
`bundle exec rspec ./spaceship/spec/two_step_or_factor_client_spec.rb` or run `Spaceship::Portal.login("<apple_id>")` from a rake task to manually test the behavior.

### Review Notes
* The Unauthorized Error people are seeing is actually a side effect - `Spaceship::TunesClient#handle_itc_response` doesn’t catch the invalid code error when hitting the wrong endpoint (same for wrong code but correct endpoint), so the rescue block on `line 143` doesn’t execute, `store_session` gets called, returns a 401, and exits with Unauthorized Error - I can open a separate PR for this, have an easy fix saved locally.
* I think the 2FA code can really use a refactor - the tight coupling between methods makes accounting for these new cases (and testing) a bit hacky, and can cause some things to break like the incorrect code recursive retry (which currently doesn't work bec of the broken invalid code error handling^) - I have a partial refactor saved locally.
* The `does not prompt user to choose number` test is brittle because the test will pass if `choose_phone_number` is ever removed or its name changes, but wasn’t sure how else to test it.
* Note that the `"noTrustedDevices"` field seems to only be present in the initial response for `GET /appleauth/auth` and not subsequent requests. Two other fields persist, `"mode"` and `"type"`, but seem to only be present once an sms has already been sent (e.g. initial response when one trusted number bec auto sent; after explicitly requesting an sms when two numbers). 
* I also have a bunch of Charles recordings for different cases which could be useful to save for the future, but would have to find a way to redact them properly.

### DIFF
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -206,10 +206,16 @@ module Spaceship
       return true
     end
 
+    # For reference in case auth behavior changes:
+    # The "noTrustedDevices" field is only present
+    # in the response for `GET /appleauth/auth`
+
+    # Account is not signed into any devices that can display a verification code
     def sms_fallback(response)
       response.body["noTrustedDevices"]
     end
 
+    # see `sms_fallback` + account has only one trusted number for receiving an sms
     def sms_automatically_sent(response)
       (response.body["trustedPhoneNumbers"] || []).count == 1 && sms_fallback(response)
     end
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -131,18 +131,32 @@ module Spaceship
         puts("Environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` is set, automatically requesting 2FA token via SMS to that number")
         puts("SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER = #{env_2fa_sms_default_phone_number}")
         puts("")
+
         phone_number = env_2fa_sms_default_phone_number
         phone_id = phone_id_from_number(response.body["trustedPhoneNumbers"], phone_number)
+        should_request_code = !sms_automatically_sent(response) 
+
         code_type = 'phone'
-        body = request_two_factor_code_from_phone(phone_id, phone_number, code_length)
+        body = request_two_factor_code_from_phone(phone_id, phone_number, code_length, should_request_code)
+      elsif sms_automatically_sent(response) # sms fallback, code was automatically sent
+        fallback_number = response.body["trustedPhoneNumbers"].first
+        phone_number = fallback_number["numberWithDialCode"]
+        phone_id = fallback_number["id"]
+
+        code_type = 'phone'
+        body = request_two_factor_code_from_phone(phone_id, phone_number, code_length, false)
+      elsif sms_fallback(response) # sms fallback but code wasn't sent bec > 1 phone number
+        code_type = 'phone'
+        body = request_two_factor_code_from_phone_choose(response.body["trustedPhoneNumbers"], code_length)
       else
         puts("(Input `sms` to escape this prompt and select a trusted phone number to send the code as a text message)")
         puts("")
         puts("(You can also set the environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` to automate this)")
         puts("(Read more at: https://github.com/fastlane/fastlane/blob/master/spaceship/docs/Authentication.md#auto-select-sms-via-spaceship-2fa-sms-default-phone-number)")
         puts("")
-        code_type = 'trusteddevice'
+        
         code = ask_for_2fa_code("Please enter the #{code_length} digit code:")
+        code_type = 'trusteddevice'
         body = { "securityCode" => { "code" => code.to_s } }.to_json
 
         # User exited by entering `sms` and wants to choose phone number for SMS
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -191,6 +191,14 @@ module Spaceship
       return true
     end
 
+    def sms_fallback(response)
+      response.body["noTrustedDevices"]
+    end
+
+    def sms_automatically_sent(response)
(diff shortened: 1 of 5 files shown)


jesiegel1:
@googlebot I signed it!

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -191,11 +206,24 @@ def handle_two_factor(response, depth = 0)
       return true
     end
 
+    def sms_fallback(response)
+      response.body["noTrustedDevices"]
+    end
+
+    def sms_automatically_sent(response)
+      (response.body["trustedPhoneNumbers"] || []).count == 1 && sms_fallback(response)
```
janpio: maybe add a comment what this line means exactly?

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -191,11 +206,24 @@ def handle_two_factor(response, depth = 0)
       return true
     end
 
+    def sms_fallback(response)
+      response.body["noTrustedDevices"]
```
janpio: maybe add a comment what this line means exactly?

janpio: My guess: No device connected to account that could get pushed auth code

spaceship/lib/spaceship/two_step_or_factor_client.rb
janpio: My guess: Exactly 1 phone number connected to account + ... (see above)

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -247,27 +275,29 @@ def request_two_factor_code_from_phone_choose(phone_numbers, code_length)
...
     # this is used in two places: after choosing a phone number and when a phone number is set via ENV var
-    def request_two_factor_code_from_phone(phone_id, phone_number, code_length)
-      # Request code
-      r = request(:put) do |req|
-        req.url("https://idmsa.apple.com/appleauth/auth/verify/phone")
-        req.headers['Content-Type'] = 'application/json'
-        req.body = { "phoneNumber" => { "id" => phone_id }, "mode" => "sms" }.to_json
-        update_request_headers(req)
-      end
+    def request_two_factor_code_from_phone(phone_id, phone_number, code_length, should_request_code = true)
```
janpio: should this maybe be renamed and the `should_request_code` part be called explicitly before calling this method in the places where `should_request_code = true`? might make it clearer how the code "flows"

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -131,18 +131,33 @@ def handle_two_factor(response, depth = 0)
         puts("Environment variable `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` is set, automatically requesting 2FA token via SMS to that number")
```
janpio: This whole `if/elsif/else` block feels ready for a refactoring now that we have 4 cases: ENV var, sms already sent, sms to be sent after choosing phone, and pushed code. Probably a future PR though to not make this one more complex than it already is?

janpio (commented):
I like this! Seems like I totally ignored the SMS cases when I wrote this. Still don't understand how this now finally can fail and explode... Apple must have cleaned up something in how they handle their stuff.

spaceship/lib/spaceship/two_step_or_factor_client.rb
jesiegel1: I gave the `should_request_code` a default value to keep the behavior of this method the same as prior to these changes unless `should_request_code` was explicitly provided, but given that this method isn't called anywhere else I agree we should make it more clear.

spaceship/lib/spaceship/two_step_or_factor_client.rb
jesiegel1: 100%, and I have a partial refactor saved locally to better handle the expanded cases and loosen the coupling between methods (e.g. splitting up `request_two_factor_code_from_phone` which tightly couples requesting a new code with prompting the user for a 2fa code and building the payload). Definitely better to include as a separate PR though to minimize the potential for regressions.

spaceship/lib/spaceship/two_step_or_factor_client.rb
jesiegel1: @janpio Sorry, I may have misunderstood, can you clarify if you mean explicitly passing a value for `should_request_code` rather than having a default value, or if you mean separating out the call to request a code into its own method?

janpio: I meant renaming this method to something without "request" and have the request part in an explicit method call before. But doesn't matter too much.

jesiegel1: I think this should wait for a refactor in a future PR, breaking up this method would affect other areas of the code and make the big conditional block even more unwieldy.

joshdholtz (approved):
This change looks sooooooo good 😊 Appreciate the contribution ❤️ I'll let @janpio merge this one when he thinks its good since he did work on this originally 💪

janpio:
(Thanks @joshdholtz)

PR #16179, opened 2020-03-16
[deliver] Provide 'env_name' to each config option that didn't previously have one*

liamnichols:
### Checklist
  - These get auto generated, right? 

### Motivation and Context

I've recently been having a good cleanup of our Fastlane scripts and configurations to start relying more on **.env** files as our main source of configuration. We prefer this approach since we utilise many custom actions that don't offer support for other config files (such as Deliverfile and so on) and as a result, all of our default action configuration exists in one place (and is overloaded for different environments).

As a result, our Fastfiles become very clear and easy to follow since we're passing very minimal arguments into each action in most cases. The downside is that not all actions provide the `env_name` for all of their `ConfigItem` definitions.

### Description
In this change, I've updated Deliver's options to update any `ConfigItem` instances that are currently missing `env_name` where the type can be passed via a string (in this case, it's any types other than `Hash`). 

I've used the naming convention `DELIVER_{key}` and in some cases, where the value is documented as a file path, and the key doesn't contain the suffix `_PATH`, I've appended it also (to match `:ipa` and `:pkg`).

Please let me know if there was a specific reason as to why these shouldn't be provided? Hopefully this is a relatively small and harmless change 🙏 

### Testing Steps

Run deliver using the new environment variables to pass in the parameters.


joshdholtz (approved):
Ah, I noticed a few days ago that these were missing! Thanks for adding them 😊 Appreciate the contribution ❤️

Issue #15810, opened 2019-12-16
Spaceship::Testflight::Group is using the wrong API


racer1988:
### New Issue Checklist

- [X] Updated fastlane to the latest version
- [X] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [X] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [X] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Using TestFlight::Group module always gives a NOT FOUND error from Fastlane

Seems to me because module declares: https://appstoreconnect.apple.com/testflight/v2/ in https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/test_flight/client.rb#L18

while [Docs](https://developer.apple.com/documentation/appstoreconnectapi/list_beta_groups) from apple declare to use `https://api.appstoreconnect.apple.com/v1/betaGroups`

##### Command executed
eg: `Spaceship::TestFlight::Group.internal_group(app_id: app.apple_id)`

##### Complete output when running fastlane, including the stack trace and command used

```
Spaceship::UnexpectedResponse: <!DOCTYPE html><html lang="en">
<head>
<title></title>
<link rel="stylesheet" href="/styles/main.css">
<style type='text/css'>
    /*include ../../dist/styles/main.css*/
</style>
</head>

<body><div id="wrapper" class="j42">
<h2>Not Found</h2></div>

<div class="footer">
    <p>Copyright &copy; <span id='year'>2015</span> Apple Inc. All rights reserved.</p>
    <ul><li><a href="https://www.apple.com/legal/internet-services/itunes/">Terms of Service</a></li><li><a href="https://www.apple.com/legal/privacy">Privacy Policy</a></li></ul>
</div>


<script type='text/javascript'>(function() {
    var isIos = navigator.userAgent.match(/(iPod|iPhone|iPad)/);
    document.body.setAttribute( 'class', (!isIos) ? 'otherBrowser' : 'isMobileSafari' )
    var yearEl = document.getElementById('year');
    if (yearEl) { yearEl.textContent = new Date().getFullYear(); }
})();
</script>

</body>
</html>
from /Users/x/.gem/ruby/2.6.5/gems/fastlane-2.137.0/spaceship/lib/spaceship/test_flight/client.rb:321:in `handle_response'
```

### Environment
```
| Key                         | Value                                       |
[...]

PR #16180, opened 2020-03-17
[spaceship] Add ability to create a beta group to Spaceship

chrisjwindisch:
### Checklist

### Motivation and Context

The Spaceship::TestFlight::Group endpoint is no longer being maintained since the conversion to the ConnectAPI. As a result there's no way to create beta groups for test flight using Spaceship.

https://github.com/fastlane/fastlane/issues/15810
Though I couldn't find documentation, the method `Spaceship::ConnectAPI::App.find("YOUR_BUNDLE_ID").get_beta_groups` works to list the groups, but you can't create a new group, which is what this commit enables.

### Description

I've created a method that creates a beta test group for TestFlight.

To do this, I've added a create_beta_group() method to the Spaceship::ConnectAPI::App object which calls another new method called Spaceship::ConnectAPI::create_beta_group(), which makes a post request to https://api.appstoreconnect.apple.com/v1/betaGroups
(https://developer.apple.com/documentation/appstoreconnectapi/create_a_beta_group#http-body)

The group_name must be provided, publicLinkEnabled defaults to false, publicLinkLimit defaults to the max of 10_000, publicLinkLimitEnabled defaults to false. I'm not sure these are the right defaults, please give me your feedback if I should change them.

Finally, I've added a test case and fixture to stubb the api call.

Note: I did not update any documentation. If there's a place to show examples for Spaceship please point me to it and I'll update it.

I've added a test case that can be run with 
```bash
bundle exec rspec spaceship
```

The easiest way to test it on an actual app is from the spaceship playground. From your app's project directory run:
```bash
bundle exec fastlane spaceship
```
From the resulting prompt run:
```ruby
Spaceship::ConnectAPI::App.find("YOUR_BUNDLE_ID").create_beta_group(group_name: "New Test Group")
```
You can verify the group was created from the Apple Developer portal by:
1. Navigate to https://appstoreconnect.apple.com/
2. Click My Apps
3. Click the app corresponding to YOUR_BUNDLE_ID
4. Click "TestFlight" from the navigation menu
5. Verify "New Test Group" was created as seen under "Testers & Groups" in the left sidebar.

You can also check that the group was created from the spaceship playground:
```ruby
Spaceship::ConnectAPI::App.find("YOUR_BUNDLE_ID").get_beta_groups
```

### DIFF
--- a/spaceship/lib/spaceship/connect_api/models/app.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app.rb
@@ -104,9 +104,15 @@ module Spaceship
         return resps.flat_map(&:to_models)
       end
 
-      def create_beta_group(group_name: nil, publicLinkEnabled: false, publicLinkLimit: 10_000, publicLinkLimitEnabled: false)
-        resps = Spaceship::ConnectAPI.create_beta_group(app_id: id, group_name: group_name, publicLinkEnabled: publicLinkEnabled, publicLinkLimit: publicLinkLimit, publicLinkLimitEnabled: publicLinkLimitEnabled).all_pages
-        return resps.flat_map(&:to_models)
+      def create_beta_group(group_name: nil, public_link_enabled: false, public_link_limit: 10_000, public_link_limit_enabled: false)
+        resps = Spaceship::ConnectAPI.create_beta_group(
+          app_id: id,
+          group_name: group_name,
+          public_link_enabled: public_link_enabled,
+          public_link_limit: public_link_limit,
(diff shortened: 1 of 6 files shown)


chrisjwindisch:
@googlebot I signed it!

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for doing this! I added a commit to change the name of the arguments to be named with `_`. Appreciate the contribution ❤️

PR #16181, opened 2020-03-17
[gym] Gym::DetectValues test date fix

### LINKED ISSUE #15987: Time-sensitive `Gym::DetectValues` tests make the CI fail
The tests on my PR #15986 just failed because there are some tests which fail when started at the end of the day but end at the beginning of the next day, due to a check of the day in the tests. See these two tests run on Circle CI which failed although there were clearly no changes related to it in my PR:
https://app.circleci.com/jobs/github/fastlane/fastlane/40527
https://app.circleci.com/jobs/github/fastlane/fastlane/40530

Here's an excerpt (just in case the links don't work later in the future):
```
Gym Gym::DetectValues Xcode config handling fetches the custom build path from the Xcode config
     Failure/Error: expect(path).to eq("/test/path/#{day}")

       expected: "/test/path/2020-02-06"
            got: "/test/path/2020-02-07"

       (compared using ==)
     # ./gym/spec/detect_values_spec.rb:14:in `block (4 levels) in <top (required)>'


```

macblazer:
### Checklist

### Motivation and Context

Improving test reliability is the motivation to fix this.

On my computer the `Gym::DetectValues` tests run in about 16 seconds.  The three tests related to the `Gym.config[:build_path]` take a few seconds each to run.

Currently in master the code pulls the current date into a string variable at the beginning of the `Gym::DetectValues` run and then uses it in three different tests.  The real code also uses the current day at the time it is called.  Because those three tests take some time, if the tests are started close to midnight any of the three calling the real code may run the next day causing the tests to fail.  The specific window of opportunity is if the `Gym::DetectValues` tests start roughly between 3 to 15 seconds before midnight (on my computer).  This is maybe a ten second window overall.  Slower computers would have a larger window because the tests would run longer (I have an iMac with a 3.6Ghz 8-core i9).

Closes #15987 

### Description

I added a `next_day` cached date string in addition to the existing `day` string.  I then had the three affected tests accept either the `day` or `next_day` as possibilities.  I also moved the cached date strings one `describe` level closer to the three tests.

In order to test this:

1. I had a Terminal window open with the test command pre-typed.
2. I turned off my system date/time synchronization with internet time in System Preferences.
3. I set the time to 23:59:45, hit Save, and switched over to Terminal.
4. At 23:59:55 I started the test run.

This caused the tests to run over midnight, and was possible to reproduce as many times as I desired.  With the code in `master` the second two `build_path` related tests failed.  With the code in this PR, they always passed.

### Testing Steps

`bundle exec rspec ./gym/spec/detect_values_spec.rb`


macblazer:
The **Lint Source Code** step in circleci found a bunch of problems with the frameit library.  I didn't touch that code, so that must be a problem in master as well I guess.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for making this PR! I ended up changing it up a bit to mock `Time.now` at the start of the tests so that it will stay that way for the rest of the tests. I appreciate the contribution ❤️

PR #16230, opened 2020-03-30
[action] Allow `get_version_number` to resolve version numbers in xcconfig files

esteluk:
At the moment this block only allows the MARKETING_VERSION to be resolved if it's
directly specified in the xcodeproj file itself. This change adopts `resolve_build_setting`
to resolve a marketing version specified in nested xcconfig files if none is present
in the project file.

### Checklist

### Motivation and Context
We found an issue where this action was no longer working in our project after migrating the `MARKETING_VERSION` from the project file and into an xcconfig file. If the version is being picked up correctly in xcode, it seems reasonable for this action to also pick up the same version number. 

### Description

Rather than using the `build_settings` subscript (which only accesses those settings defined in directly in the project file), this change uses `resolve_build_setting` (which is also provided by xcodeproj to grab the true build setting in the same way xcode itself would.

Added a test to the existing test suite that uses a target configured in this way.


joshdholtz (approved):
This looks great! Thanks for making this change ❤️

PR #16254, opened 2020-04-04
[action] setup_jenkins with match

michallaskowski:
### Checklist

### Motivation and Context

`setup_ci` action already handles Match, though by reading `ENV["MATCH_KEYCHAIN_NAME"]`
In a setup where default keychain is not `login.keychain`, and Jenkins is used as CI system, Match needs to be set up separately, because it will try to still use `login.keychain`. This leads to this piece of code being present in projects:

```
before_all do
    setup_jenkins(
      keychain_path: ENV["BUILD_KEYCHAIN"],
      keychain_password: ENV["BUILD_KEYCHAIN_PASSWORD"],
      set_code_signing_identity: false)

    # setup fastlane match with same keychain and password as just unlocked with setup_jenkins
    ENV["MATCH_KEYCHAIN_NAME"] = File.basename(ENV["BUILD_KEYCHAIN"]) if ENV["BUILD_KEYCHAIN"]
    ENV["MATCH_KEYCHAIN_PASSWORD"] = ENV["BUILD_KEYCHAIN_PASSWORD"] if ENV["BUILD_KEYCHAIN_PASSWORD"]
end
```
because otherwise `setup_jenkins` would unlock a BUILD_KEYCHAIN, but match would by default try to to install certificates and keys into `login.keychain`.

### Description
With that change, setup_jenkins will make sure the unlocked keychain will be used also by Match.
Environment variables used by Fastlane Match are set if a keychain is unlocked by `setup_jenkins`

### Testing Steps

Can be tested by creating a test keychain: `bundle exec fastlane run create_keychain name:test.keychain password:test unlock:true`
Then trying to use before_all pasted earlier, and running a lane that uses Match, example:
`FL_SETUP_JENKINS_FORCE=true BUILD_KEYCHAIN=~/Library/Keychains/test.keychain BUILD_KEYCHAIN_PASSWORD=test bundle exec fastlane build`

### DIFF
--- a/fastlane/lib/fastlane/actions/setup_jenkins.rb
+++ b/fastlane/lib/fastlane/actions/setup_jenkins.rb
@@ -106,8 +106,8 @@ module Fastlane
         [
           list,
           "This action helps with Jenkins integration. Creates own derived data for each job. All build results like IPA files and archives will be stored in the `./output` directory.",
-          "The action also works with [Keychains and Provisioning Profiles Plugin](https://wiki.jenkins-ci.org/display/JENKINS/Keychains+and+Provisioning+Profiles+Plugin), the selected keychain will be automatically unlocked and the selected code signing identity will be used." 
-          "[Match](https://docs.fastlane.tools/actions/match/) will be also set up to use the unlocked keychain, and set in read-only mode."
+          "The action also works with [Keychains and Provisioning Profiles Plugin](https://wiki.jenkins-ci.org/display/JENKINS/Keychains+and+Provisioning+Profiles+Plugin), the selected keychain will be automatically unlocked and the selected code signing identity will be used.",
+          "[Match](https://docs.fastlane.tools/actions/match/) will be also set up to use the unlocked keychain, and set in read-only mode.",
           "By default this action will only work when _fastlane_ is executed on a CI system."
         ].join("\n")
       end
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/setup_jenkins.rb
```diff
@@ -40,6 +43,9 @@ def self.run(params)
             add_to_search_list: params[:add_keychain_to_search_list],
             set_default: params[:set_default_keychain]
           )
+          ENV['MATCH_KEYCHAIN_NAME'] = keychain_path
+          ENV['MATCH_KEYCHAIN_PASSWORD'] = params[:keychain_password]
+          ENV["MATCH_READONLY"] = true.to_s
```
joshdholtz, line 48:
```suggestion
          ENV['MATCH_KEYCHAIN_NAME'] ||= keychain_path
          ENV['MATCH_KEYCHAIN_PASSWORD'] ||= params[:keychain_password]
          ENV["MATCH_READONLY"] ||= true.to_s
```

I think we may went to only set these if they haven’t been set already by the user 🤔 Thoughts?

joshdholtz (changes requested):
I like this addition! Just one quick thought on setting the match environment variables 😊

fastlane/lib/fastlane/actions/setup_jenkins.rb
michallaskowski: Sure, will change it.

michallaskowski: argh, I am not up to date with those "newer" GitHub features, day to day I use BitBucket. I just noticed I could have just applied it. Anyway, did the same except, I left ENV["MATCH_READONLY"] ||= "true". Let me know, I can change it.

joshdholtz (approved):
Thanks for making that change! One test is failing but a rebase from master will fix that. I will rebase and merge in the morning. Thanks for the contribution ❤️

Issue #14157, opened 2019-01-29
Match creates new provisioning profiles every time, appending timestamp to profile name


JanBrinker:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues) (I found some, but none solved my problem)

### Issue Description
Ever since sometime November or December last year my team is having issues with match. When we run our lane so create new provisioning profiles we get the following warnings: 

```
[13:41:55]: No existing profiles found, that match the certificates you have installed locally! Creating a new provisioning profile for you
[13:41:57]: The name 'match AdHoc com.acme.myapp.prod.adhoc' is already taken, using another one.
[13:41:57]: Creating new provisioning profile for 'com.acme.myapp.Prod.adhoc' with name 'match AdHoc com.acme.myapp.Prod.adhoc 1548765717' for 'ios' platform
```

For some reason our setup worked before and then started to behave like this after we had to use a new Apple ID for our team as a code signing identity. Sometimes match was able to create the profile with the correct name, sometimes it used a timestamp at the end. I was not able to find a pattern here so far.

I was able to get rid of this issue by using `force: true` in my match call for all other bundle IDs (we use different bundle IDs for dev, stage and prod and prod_adhoc [for testers who need access to live data]). However for this single bundle ID the issues remain. Any ideas what's the issue?

##### Command executed
    match(
        type: "adhoc",
        app_identifier: "com.acme.myapp.Prod.adhoc",
        force_for_new_devices: true,
        force: force,
        keychain_password: keychain_password
    )

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
wookieebook:MyApp jan$ bundle exec fastlane register_devices_generate_certificates
[✔] 🚀 
+---------------------------+---------+------------------+
|                      Used plugins                      |
+---------------------------+---------+------------------+
| Plugin                    | Version | Action           |
+---------------------------+---------+------------------+
| fastlane-plugin-appcenter | 0.1.7   | appcenter_upload |
| fastlane-plugin-badge     | 1.0.0   | add_badge        |
+---------------------------+---------+------------------+

[14:28:24]: ----------------------------------------
[...]

Issue #11072, opened 2017-11-30
can we stop the execution at point when "the name xyz is already taken, creating new one"


kumarkalluri:
- Sigh with config.
          
Issue is : 
       Actually, provision profile(xyz) is taking from the config file. So, in the above case if that named provision profile(xyz) already created and associated with other certificates. Sigh is looking the for the provisioning profile(xyz) and if it exists, creating the profile with name appending(xyz 1150345) with the timestamp, but my scheme is looking for the xyz profile and then build fails.
  So, I want to stop the execution when the provisioning profile is already taken and throw the customized message to user.

joshdholtz:
@kumarkalluri Thanks for the feature suggestion! We have a long list of things to work on so if you are willing to create a PR for this that would be great 😊

PR #16281, opened 2020-04-08
[sigh][match] Add :fail_on_name_taken in sigh, use in match

michallaskowski:
### Checklist

### Motivation and Context
As discussed in #11072 and #14157 it may happen when using match, that a profile should be recreated, but instead during creation we got a message that the name is taken and a duplicate, with timestamp appended to the name, is created and uploaded to the repo/s3.
The name change may easily cause codesigning to break, when profile is set in pbxproj statically like in the [the docs](https://docs.fastlane.tools/actions/match/#to-build-from-xcode-manually).

I was not able to find a reason of this issue, it happened a few times during force_for_new_devices run. Because we create multiple development profiles at once, it is easy to miss it, because it doesn't raise an exception.
This change, while not fixing the root issue, will make match (by sigh) to raise an exception when this happens. 

### Description
Sigh takes a new parameter, `fail_on_name_taken`, by default false. When fail_on_name_taken: true and profile name is for some reason taken, sigh will raise an exception explaining the situation.
Match will always set `fail_on_name_taken` to be true when using sigh.

### Testing Steps
Tests added to match and sigh. It is harder to replicate this via http proxy, but is possible by stubbing delete operation on a profile.


sigh/lib/sigh/options.rb
```diff
@@ -153,7 +153,13 @@ def self.available_options
                                      env_name: "SIGH_PROVISIONING_PROFILE_TEMPLATE_NAME",
                                      description: "The name of provisioning profile template. If the developer account has provisioning profile templates (aka: custom entitlements), the template name can be found by inspecting the Entitlements drop-down while creating/editing a provisioning profile (e.g. \"Apple Pay Pass Suppression Development\")",
                                      optional: true,
-                                     default_value: nil)
+                                     default_value: nil),
+        FastlaneCore::ConfigItem.new(key: :fail_on_name_taken,
+                                     env_name: "SIGH_PROVISIONING_PROFILE_FAIL_ON_NAME_TAKEN",
```
joshdholtz, line 158:
```suggestion
                                     env_name: "SIGH_FAIL_ON_NAME_TAKEN",
```

sigh/lib/sigh/runner.rb
```diff
@@ -126,6 +126,7 @@ def create_profile!
 
       unless Sigh.config[:skip_fetch_profiles]
         if Spaceship.provisioning_profile.all.find { |p| p.name == name }
+          raise "The name '#{name}' is already taken, and fail_on_name_taken is true" if Sigh.config[:fail_on_name_taken]
```
joshdholtz, line 129:
```suggestion
          UI.user_error!("The name '#{name}' is already taken, and fail_on_name_taken is true") if Sigh.config[:fail_on_name_taken]
```

joshdholtz (changes requested):
I really like this feature! I’ve run into this and didn’t think of making a solution like this 😊 A few small suggestions but should be good to merge after that!

match/lib/match/generator.rb
```diff
@@ -72,7 +72,8 @@ def self.generate_provisioning_profile(params: nil, prov_type: nil, certificate_
         ignore_profiles_with_different_name: true,
         team_id: params[:team_id],
         team_name: params[:team_name],
-        template_name: params[:template_name]
+        template_name: params[:template_name],
+        fail_on_name_taken: true
```
joshdholtz, line 76: Now that I think about it... I’m a little bit iffy on this change since its breaking. I’m going to add a commit into this PR to make this an option on _match_ as well 😊

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
I love this as a new option! Thanks for doing all the hard work on this ❤️ Really appreciate it!

Issue #16168, opened 2020-03-13
upload_symbols_to_crashlytics not working with Firebase/Crashlytics


larryonoff:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I'm trying to automate dSYM upload into Firebase/Crashlytics, but it doesn't work. The same fastlane script works perfectly with Fabric Crashlytics.

##### Command executed
```ruby
  lane :upload_symbols do |options|
    download_dsyms(
      app_identifier: app_identifier,
      version: options[:version]
    )

    upload_symbols_to_crashlytics(
      gsp_path: "./inpreview/GoogleService-Info.plist"
    )

    clean_build_artifacts()
  end
```

```
fastlane upload_symbols version:"2.3.1"
```

##### Complete output when running fastlane, including the stack trace and command used
<details>
  <pre>
[17:24:33]: -------------------------------------------
[17:24:33]: --- Step: upload_symbols_to_crashlytics ---
[17:24:33]: -------------------------------------------
[17:24:33]: Extracting '/Users/larryonoff/Develop/Projects/work/in/inpreview-copy/instapreview-2.3.1-279.dSYM.zip'...
[17:24:33]: $ unzip -qo /Users/larryonoff/Develop/Projects/work/in/inpreview-copy/instapreview-2.3.1-279.dSYM.zip
[17:24:33]: Uploading 'f99ab8b0-2478-3868-829f-aa671ed5cdec.dSYM'...
[17:24:33]: Shell command exited with exit status 13 instead of 0.
[17:24:33]: Uploading '8a998e4d-bc5f-34b3-9ee8-2c657fedb00b.dSYM'...
[17:24:33]: Shell command exited with exit status 13 instead of 0.
[17:24:33]: Uploading '42b93d12-42b4-31e9-a3fe-52759c173b77.dSYM'...
[17:24:33]: Shell command exited with exit status 13 instead of 0.
[17:24:33]: Uploading '5aecd2f7-9240-3076-93f8-620fad49d99a.dSYM'...
[17:24:33]: Shell command exited with exit status 13 instead of 0.
[...]

PR #16296, opened 2020-04-12
[upload_symbols_to_crashlytics] Prefer Firebase Crashlytics's upload-symbols to Fabric

dlackty:
### Checklist

### Motivation and Context
image

According to the Firebase documentation, if user has upgraded to Firebase Crashlytics, we should use `FirebaseCrashlytics`'s `upload-symbols` cli instead of `Fabric`'s one.

Currently, Fastlane action `upload_symbols_to_crashlytics` finds the cli automatically but if both presented, it will use the legacy Fabric one.

Though it's not common to have both Fabric & FirebaseCrashlytics installed, for those affected by this issue, it's quite hard to find the root cause.

Possible related issue: #16168.
### Description

Instead of finding binary using `Dir["./Pods/**/upload-symbols"]`, I expand it to `Dir["./Pods/Fabric/upload-symbols"] + Dir["./Pods/FirebaseCrashlytics/upload-symbols"]` and take the last found one.

Tested on a sample repo with both Pod installed.

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -130,7 +130,7 @@ module Fastlane
       end
 
       def self.find_binary_path(params)
-        params[:binary_path] ||= (Dir["/Applications/Fabric.app/**/upload-symbols"] + Dir["./Pods/**/upload-symbols"]).last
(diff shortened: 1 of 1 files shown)


janpio:
There are not other possible paths `**` could have covered?

joshdholtz (approved):
This makes sense to me! I actually don’t like the `**` that was there 😛 I feel like that could have accidentally could have picked up `upload-symbols` from another pod 🤷‍♂️ 

Thanks for the contribution! ❤️

Blackjacx:
Would it be thinkable to also include the carthage users :-D ? Since carthage currenlty doesn't package this correctly the `upload-symbols` is not in the Carthage subfolder. Rather we put it in a custom tools folder. Wdyt?

PR #14458, opened 2019-03-22
[frameit] Split framing and wrapping

janpio:
Some time ago I spent too many hours with frameit. End result was that I took it apart and rewrote it to handle framing screenshots in device frames, and then wrapping it onto a background with text separately.

- The few existing tests were not considered when rewriting
- Performance seems to be really bad now

### TODO

- Rebase when https://github.com/fastlane/fastlane/pull/14460 and https://github.com/fastlane/fastlane/pull/14459 are merged

PR #16302, opened 2020-04-14
[frameit] Speedup framing (~2x) + regression tests

### LINKED ISSUE #14626: [Frameit] Frameit is very slow (15s+ to frame a single image)

bvirlet:
## Description

As stated here #14626, frameit is very slow particularly because of a regression. This PR aims at fixing this with commit 0d38f52. Other commits do some required cleanup and most importantly add regression tests.

The involved code was responsible for creating the title/keywords of the screenshots. In summary, to create this text, this piece of code creates a very large empty image, draws the title in it, and then trims the resulting image to fit closely around the text by using `identify -format ‘%@‘`. This command is actually quite slow (6s on my Macbook Pro 2019).

To avoid using this command, we rely on a slightly different image magick tool that lets us directly create a text image properly sized for the text. That command is very fast (~ 25 ms). We cannot directly do that with MiniMagick high-level methods (this doesn’t seem to be supported), so we dig a bit deeper to use ImageMagick’s convert directly.

Basically, this enables us to reduce the framing time by around 6s on my machine, which leads to a 2x+ speedup.

The advantage of this regression tests compared to more internal unit tests would be that they would allow heavier refactoring of the code base such as what @janpio had started here: #14458. It would be a good idea to extend this regression test base over time, particularly when someone submits a PR affecting a feature of frameit that isn't tested yet by these regression tests.

Fix #14626

## Caveats

- The regression tests will slightly slow down the fastlane specs as frameit is a slow tool. However, I  believe the regression tests are really worth it given how fragile the code base of frameit is.


bvirlet:
Not sure why the tests fail on CI whereas they succeed locally. Any advice?

bvirlet:
@janpio it looks like my tests fail on CircleCI because imagemagick is not installed on the Mac machines. Is it reasonable to add it to the `brew install` command in `.circleci/config.yml`? Should I consider running the regression tests on a Linux machine only (but that'd be a pity as `frameit` is mainly used on Mac)?

janpio:
> @janpio it looks like my tests fail on CircleCI because imagemagick is not installed on the Mac machines. Is it reasonable to add it to the `brew install` command in `.circleci/config.yml`? Should I consider running the regression tests on a Linux machine only (but that'd be a pity as `frameit` is mainly used on Mac)?

Any idea why this only pops up now? Wasn't the code executed before? How is this handled on user's machines?

But yeah in general installing it on the CI machine manually sounds like a reasonable thing to do here.

bvirlet:
> Any idea why this only pops up now? Wasn't the code executed before? How is this handled on user's machines?
> 

I believe no code was relying on ImageMagick before the new regression tests included in this PR. All existing specs were only unit tests.

In my last commit, I have installed it on the Mac machines, but it looks like this causes other issues (it seems it could be related to an OpenSSL dependency installed by ImageMagick that later causes problems for bundler) but I'm not too familiar with that.

janpio:
Looks to me that Lint, Windows/AppVeyor, Ubuntu and `(Execute tests on macOS (Xcode 11.3.0, Ruby 2.6))` show proper test failures, and the other Mac stuff is confusing :/

joshdholtz:
I will take a look and see whats’s up!

bvirlet:
> Looks to me that Lint, Windows/AppVeyor, Ubuntu and `(Execute tests on macOS (Xcode 11.3.0, Ruby 2.6))` show proper test failures, and the other Mac stuff is confusing :/

Yes, there are definitely some "proper" failures: 
- rubocop issues I need to fix
- actual failures in the regression tests. It seems that depending on the machine and/or the ImageMagick version, the compositing of images doesn't result in exactly the same images despite the same input being used. I need to add some tolerance for the image comparison or maybe only run the tests on a select architecture. Maybe also lock the imagemagick version. Not entirely sure of the best route to go.

janpio:
> * It seems that depending on the machine and/or the ImageMagick version, the compositing of images doesn't result in exactly the same images despite the same input being used. I need to add some tolerance for the image comparison or maybe only run the tests on a select architecture. Maybe also lock the imagemagick version. Not entirely sure of the best route to go.

Maybe you can allow a slight mismatch in the images? Did you actually see the difference this is returning in the files? Sub or single pixel difference in the composition can make quite a difference in sum.

I'll try to look into the unrelated CI problems.

janpio:
@bvirlet Can you please give maintainers/committers here permission to update the PR branch? Should be possible via a checkbox on the right - then I can merge master into the branch and see if my recent CI fixes also help here. Thanks"

bvirlet:
image

@janpio I cannot see this checkbox.

janpio:
Strange, this is the documentation: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

bvirlet:
I wonder if it's not because the initial PR is owned by an organization (my company) and not a "user-owned repo" as the documentation mentions. Do you want me to merge master to this branch? Alternatively, feel free to "copy" the branch if that's easier.

janpio:
Oh that is indeed possible. You merging `master` (or rebasing on master)  would be nice - then we would get a new CI run and can see where we are at here.

bvirlet:
Done!

bvirlet:
Well… it's not much better, there are still a bunch of issues related to Homebrew and openssl.

janpio:
Yes, I am a bit helpless with some of them. I did this to output the ruby version before the failing "setup build" step: https://github.com/fastlane/fastlane/commit/86ba1a5081fb034c3cea3a200a7b528f93bfaaa9 Maybe that can tell us a bit if this indeed fails (what the error message indicates)

bvirlet:
I'm closing this as I don't have the time to maintain, and we are switching to a different way to frame screenshots internally.

Issue #16289, opened 2020-04-10
screengrab not considering frameit folder location


wellavelino:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I'm happy to open a pull request on frameit, but first I need to understand if this was made on purpose :) 

When using screengrab it pulls screenshots to:
`found this piece of code on: screengrab/runner.rb:353`
```
 # We then replace the last segment of the screenshots directory path with the device_type
 # specific name, as expected by supply
 #
 # (Moved to: fastlane/metadata/android/en-US/images/phoneScreenshots)`
```
I understand that has to be in that locally for supply to be able to upload those. But frameit requires the screenshots to be in the same folder as title.strings/keyword.strings.

My actual folder structure is

This is one specific folder for one language 
`fastlane/screenshots/language/
```
|____title.xml
|____images
| |____phoneScreenshots
| | |____screenShot3_Macros_1586444424911.png
| | |____screenShot1_TicketList_1586444420940_framed.png
| | |____screenShot3_Macros_1586444424911_framed.png
| | |____screenShot4_Notifications_1586444420200.png
| | |____screenShot2_TicketDetails_1586444424182.png
| | |____screenShot4_Notifications_1586444420200_framed.png
| | |____screenShot2_TicketDetails_1586444424182_framed.png
| | |____screenShot1_TicketList_1586444420940.png
|____title.strings
|____mock.xml

```
##### Command executed
note that language is generated using an array of languages.
```
 LANGUAGES = %w[ko ja en-US ar-SA fr-FR pl cs sv nl-NL zh-TW no en-GB bg fi da
               zh-CN pt-BR tr it ru de-DE es-ES ro th id el hi hu vi he]  
   desc "Creates localized mocks, takes screenshots"
   lane :capture_screenshots do
    #update_screenshot_mock_localizations
    # build_for_screengrab
     # screengrab
     LANGUAGES.each do |language|
       frameit(use_platform: 'ANDROID', path: "./fastlane/screenshots/#{language}/images/phoneScreenshots")
     end
   end
```

##### Complete output when running fastlane, including the stack trace and command, used

<details>
[...]

PR #16306, opened 2020-04-14
[frameit] Change framefile.json load path to support Screengrab and Supply

wellavelino:
Issue https://github.com/fastlane/fastlane/issues/16289
When using screengrab and frameit, frameit doesnt load properly
the config file because it matches the current iOS setup,
to be able to load the android properly we need to fully extend the
path.

### Checklist

### Motivation and Context
This PR fixes this issue: https://github.com/fastlane/fastlane/issues/16289
I noticed that frameit expects the Framefile.json to be in a specific folder location, but when using it with screengrab and supply it doesn't support the folder structure, I saw the in the PR `ea4d8a2`
the method `create_config` was created but isn't considering the actual support to supply/screengrab, the screenshots have added the frame but it doesn't add the background.

### Description
I've just extended the path to be 4 levels deep.
Since I couldn't see much value in add a test for this specific method I tested loading from different paths, but if you guys think that might be cool I can add it.

### Testing Steps
Just created a new test in the config_parser_spec to simulate it to be and tested in a local project.
I can add the test if it's ok.

### DIFF
--- a/frameit/lib/frameit/runner.rb
+++ b/frameit/lib/frameit/runner.rb
@@ -82,8 +82,12 @@ module Frameit
     # Loads the config (colors, background, texts, etc.)
     # Don't use this method to access the actual text and use `fetch_texts` instead
     def create_config(screenshot_path)
+      # Screengrab pulls screenshots to a different folder location
+      # frameit only handles two levels of folders, to not break
+      # compatibility with Supply we look into a different path for Android
+      # Issue https://github.com/fastlane/fastlane/issues/16289
       config_path = File.join(File.expand_path("..", screenshot_path), "Framefile.json")
-      config_path = File.join(File.expand_path("../..", screenshot_path), "Framefile.json") unless File.exist?(config_path)
+      config_path = File.join(File.expand_path("../../../..", screenshot_path), "Framefile.json") unless File.exist?(config_path)
       file = ConfigParser.new.load(config_path)
       return {} unless file # no config file at all
       file.fetch_value(screenshot_path)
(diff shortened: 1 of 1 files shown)


wellavelino:
@googlebot I signed it!

frameit/lib/frameit/runner.rb
```diff
@@ -82,8 +82,12 @@ def editor(screenshot, config)
     # Loads the config (colors, background, texts, etc.)
     # Don't use this method to access the actual text and use `fetch_texts` instead
     def create_config(screenshot_path)
+      # Screengrab pulls screenshots to a different folder location
+      # frameit only handles two levels of folders, to not break
+      # compatibility with Supply we look into a different path for Android
+      # Issue https://github.com/fastlane/fastlane/issues/16289
       config_path = File.join(File.expand_path("..", screenshot_path), "Framefile.json")
-      config_path = File.join(File.expand_path("../..", screenshot_path), "Framefile.json") unless File.exist?(config_path)
+      config_path = File.join(File.expand_path("../../../..", screenshot_path), "Framefile.json") unless File.exist?(config_path)
```
joshdholtz, line 90: Do we want to keep the `../..`? I don’t know if removing that is actually breaking or not but it might be safer to keep it in 🤔 Thoughts?

joshdholtz (changes requested):
One small thought on keeping the removed line 😇  And if there is an existing spot to add another test for this that would be great! If not don’t worry about it 🙃

frameit/lib/frameit/runner.rb
wellavelino: Hey @joshdholtz, I was waiting for someone to reply in here, I guess it makes sense so we don't break compatibility with an older path.
I'll put it back.  thanks for replying.

wellavelino:
hey @joshdholtz can you have a look when you have some time?
Cheers

joshdholtz (approved):
Thanks for fixing this! Appreciate the contribution ❤️

wellavelino:
@joshdholtz no worries, looking forward to open my second PR screengrab related

Issue #16294, opened 2020-04-12
Spaceship::UnexpectedResponse: [!] Missing required information to submit for external testing - Missing requied data.


liftyourgame:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Expect fastlane to upload app to Apple TestFlight
Instead get error:
 Spaceship::UnexpectedResponse: [!] Missing required information to submit for external testing - Missing requied data.

Misspelling of 'requied' is as per the actual error.

Was working until a few days ago when I updated Fastlane.

##### Command executed

fastlane beta

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
[14:44:20]: ------------------------------
180 [14:44:20]: --- Step: default_platform ---
181 [14:44:20]: ------------------------------
182 [14:44:20]: Driving the lane 'ios beta' 🚀
183 [14:44:20]: ------------------------------------
184 [14:44:20]: --- Step: increment_build_number ---
185 [14:44:20]: ------------------------------------
186 Current version of project myapp is: 
187     32
188 /Users/greg/builds/ykHv483k/0/myapp/myapp/ios
189 [14:44:22]: $ cd /Users/greg/builds/ykHv483k/0/myapp/myapp/ios && agvtool next-version -all && cd -
190 [14:44:23]: ▸ Setting version of project myapp to:
191 [14:44:23]: ▸ 33.
192 [14:44:23]: ▸ Also setting CFBundleVersion key (assuming it exists)
193 [14:44:23]: ▸ Updating CFBundleVersion in Info.plist(s)...
194 [14:44:23]: ▸ Updated CFBundleVersion in "myapp.xcodeproj/../myapp-tvOS/Info.plist" to 33
195 [14:44:23]: ▸ Updated CFBundleVersion in "myapp.xcodeproj/../myapp-tvOSTests/Info.plist" to 33
196 [14:44:23]: ▸ Updated CFBundleVersion in "myapp.xcodeproj/../myapp/Info.plist" to 33
197 [14:44:23]: ▸ Updated CFBundleVersion in "myapp.xcodeproj/../myappTests/Info.plist" to 33
[...]

PR #16307, opened 2020-04-15
[pilot] mark Hash options with is_string:false

tmm1:
In #16294 we discovered that `PILOT_BETA_APP_REVIEW_INFO` cannot be set via the pilot distribute cli. The options are listed as `STRING` in `pilot distribute --help` incorrectly.

### DIFF
--- a/pilot/lib/pilot/options.rb
+++ b/pilot/lib/pilot/options.rb
@@ -75,6 +75,7 @@ module Pilot
                                      default_value: false),
         FastlaneCore::ConfigItem.new(key: :beta_app_review_info,
                                      type: Hash,
+                                     is_string: false,
                                      env_name: "PILOT_BETA_APP_REVIEW_INFO",
                                      description: "Beta app review information for contact info and demo account",
                                      optional: true,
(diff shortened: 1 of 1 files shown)


snatchev:
I am surprised this had the desired effect. AFAIK, `is_string` is deprecated and should be avoided. @joshdholtz does this ring a bell?

joshdholtz:
Ah, so `is_string`is starting to slowly get removed from all tools and I believe its removed in `pilot` already. The `beta_app_review_info` is a `Hash` because it supports localized keys so I don’t remember exactly how or if it works with CLI. 

It might be a String via CLI because CLI can’t really natively support a Hash. I _think_ it might handle the input as a JSON object thought? 

I could be wrong though and that might need to be added. 😬

joshdholtz:
Totally did not realize this was a PR and thought this was an issue 🙃 I think the actual solution there should not be setting `is_string: false` but by changing the logic in the fastlane option / CLI generation logic to handle `Hash` properly 🤔

tmm1:
Thanks for explanation. Closing as this is not the right approach.

Issue #15908, opened 2020-01-17
Provide latest build number for CLI "fastlane pilot distribute"


Blackjacx:
### Question Subject
```shell
fastlane pilot distribute -q <itc_team_id> --distribute_external "true" --notify_external_testers "true" --demo_account_required "true" -a <bundle_id> -g <tester_groups> -w <changelog>
```

#### Question Description
I use the command above to notify testers about already uploaded builds.

The command figures out the latest version - build number - combination and adds the specified tester groups to this build.

BUT figuring out the build number takes very long and I already have it locally. Can I provide it myself to the command? This would bring us a massive speed boost. I sadly didn't find any documentation related to this... :(

Blackjacx:
I think its not necessary since it only a question related to ne most up to date fastlane.

bguidolim:
@Blackjacx you need to set the build number before build with [increment_build_number](https://docs.fastlane.tools/actions/increment_build_number/#increment_build_number) action.

Blackjacx:
> @Blackjacx you need to set the build number before build with [increment_build_number](https://docs.fastlane.tools/actions/increment_build_number/#increment_build_number) action.

Hey there this we do of course. The question is about providing the latest build number when distributing the build to testers, not about the upload. The build is already uploaded (and processed by apple) when I run the command above.

I reformulated my question and added the complete command we use to add testers to the latest build.

Blackjacx:
Is there anybody who can help me on this topic?

Blackjacx:
This is still an issue and it would be nice I could get an answer. I could tray to implement it myself and file a PR... Would that be an option?

tmm1:
You can add `--build_number`

https://github.com/fastlane/fastlane/pull/16221

Blackjacx:
This is amazing! Thanks man. This saves sooo much time for us, since we batch-distribute 15 apps and counting 👍

PR #16314, opened 2020-04-15
[fastlane_core] allow passing in Hash options via cli/env as json

tmm1:
### Checklist

### Motivation and Context

Some fastlane options are defined as `Hash`es, and cannot currently be specified when using the fastlane cli or environment variables.

cc #16294 #16307 

### Description

Hash options can be set via the cli/env using JSON syntax.

### Testing Steps

Ran `fastlane pilot distribute` with `PILOT_BETA_APP_REVIEW_INFO='{"contact_first_name":"...",...}'` set.

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/config_item.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/config_item.rb
@@ -2,6 +2,7 @@ require_relative '../ui/ui'
 require_relative '../ui/errors/fastlane_error'
 require_relative '../helper'
 require_relative '../module'
+require 'json'
 
 module FastlaneCore
   class ConfigItem
@@ -240,6 +241,12 @@ module FastlaneCore
       elsif allow_shell_conversion
         return value.shelljoin if value.kind_of?(Array)
         return value.map { |k, v| "#{k.to_s.shellescape}=#{v.shellescape}" }.join(' ') if value.kind_of?(Hash)
+      elsif data_type == Hash && value.kind_of?(String)
+        begin
+          parsed = JSON.parse(value)
+          return parsed if parsed.kind_of?(Hash)
+        rescue JSON::ParserError
+        end
       elsif data_type != String
         # Special treatment if the user specified true, false or YES, NO
         # There is no boolean type, so we just do it here
--- a/fastlane_core/lib/fastlane_core/configuration/config_item.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/config_item.rb
@@ -228,6 +228,7 @@ module FastlaneCore
       true
     end
 
+    # rubocop:disable Metrics/PerceivedComplexity
     # Returns an updated value type (if necessary)
     def auto_convert_value(value)
       return nil if value.nil?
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -240,6 +241,12 @@ def auto_convert_value(value)
       elsif allow_shell_conversion
         return value.shelljoin if value.kind_of?(Array)
         return value.map { |k, v| "#{k.to_s.shellescape}=#{v.shellescape}" }.join(' ') if value.kind_of?(Hash)
+      elsif data_type == Hash && value.kind_of?(String)
+        begin
+          parsed = JSON.parse(value)
+          return parsed if parsed.kind_of?(Hash)
+        rescue JSON::ParserError
+        end
```
snatchev, line 249: I would make sure we return something in the case of a `ParserError`

`retun value` seems like a non-breaking change.

snatchev (changes requested):

fastlane_core/lib/fastlane_core/configuration/config_item.rb
tmm1: Hmm. I wanted to preserve the existing behavior, which basically throws an error saying expecting Hash but got String

I could let the JSON::ParserError surface up if that's preferable?

snatchev: Ah yes! It does `return value` at the end of the function, so it's effectively the same. This looks good.

snatchev (approved):

joshdholtz (approved):
This looks good to me! Thanks for making this change 😊 Appreciate the contribution ❤️

joshdholtz:
@googlebot I consent.

jmichas:
How can I use this in Azure Devops pipeline task?
In the App Store Release task how do I put this in the fastlaneArguments?
I tried `--beta_app_review_info $(beta_app_review_info_json)` where the variable is:
```
beta_app_review_info_json: '{
       "contact_email": "me@mine.io",
        "contact_first_name": "Test",
        "contact_last_name": "Guy",
        "contact_phone": "12344567",
        "demo_account_name": "sometestuser",
        "demo_account_password": "password!"
    }'
```
but I get the `[!] 'beta_app_review_info' value must be a Hash! Found String instead.` error.

1Siger1:
Can someone answer @jmichas question? I am also using Azure DevOps and we were left without answer.

yajiex:
@1Siger1 you don't need to pass in argument, just define `PILOT_BETA_APP_REVIEW_INFO` in Variables. I followed this, upgraded fastlane to latest version, and problem solved! https://github.com/fastlane/fastlane/issues/16294#issuecomment-618406901

Blackjacx:
Wow this issue made me really headaches the last weeks. Thank you so much this fix 👍🎉

Blackjacx:
To all using the command line version of `pilot distribute`, could you please post the command you use? I have a different issue open on fastlane regarding this without many reactions 🤔. Maybe you could help me there? 

https://github.com/fastlane/fastlane/issues/15908

PR #16334, opened 2020-04-18
[pilot] Remove redundant update_review_detail call

ykay:
🔑

### Checklist

### Motivation and Context
Pilot has recently started failing in my automated build script when distributing the TestFlight to external testers. Other users have reported similar symptoms (in the issue linked below), including a "Test Information" section being wiped on App Store Connect. I suspect this is happening due to an unnecessary call being made to Apple's `/betaAppReviewDetails` endpoint even when we don't pass the `beta_app_review_info` or `demo_account_required` parameters to `pilot`. The intended logic would seem that we only want to call it if the parameter was explicitly passed, since the logic that follows right after is a conditional check for `should_update_beta_app_review_info(options)`.

The current workaround is to explicitly pass `beta_app_review_info` (i.e. contact information) to `pilot`. This prevents the contact information from being wiped from the TestFlight settings, however this is not ideal.

Fixes issue: https://github.com/fastlane/fastlane/issues/16294

### Description
There are two consecutive calls to `update_review_detail()` inside `update_beta_app_meta()`. The first call is not required, because `demo_account_required` should be one of the fields for `beta_app_review_info`. 

This fix checks when `demo_account_required` is passed to pilot and merges it into the `beta_app_review_info` hash.

A test case was removed because the test input doesn't contain `demo_account_required`, so the expectation should be that it doesn't call `update_review_detail()` anymore.

### Testing Steps
Call `fastlane pilot` to upload an app, and pass an external tester group to trigger the beta app review, but do not pass `beta_app_review_info` as a parameter.

Expected Result: Pilot should return successfully without failing during distribution.

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -215,8 +215,11 @@ module Pilot
     end
 
     def update_beta_app_meta(options, build)
-      # Setting account required wth AppStore Connect API
-      update_review_detail(build, { demo_account_required: options[:demo_account_required] })
+      # If demo_account_required is a parameter, it should added into beta_app_review_info
+      unless options[:demo_account_required].nil?
+        options[:beta_app_review_info] = {} if options[:beta_app_review_info].nil?
+        options[:beta_app_review_info][:demo_account_required] = options[:demo_account_required]
+      end
 
       if should_update_beta_app_review_info(options)
         update_review_detail(build, options[:beta_app_review_info])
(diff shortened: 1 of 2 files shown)


ykay:
@googlebot I signed it!

joshdholtz (approved):
This looks great! Thanks for improving this 😊 Appreciate the contribution! ❤️

ykay:
> This looks great! Thanks for improving this 😊 Appreciate the contribution! ❤️

Thanks! Happy to be able to contribute.

Issue #16216, opened 2020-03-27
match nuke fails with deleting from S3


theirix:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I am running `fastlane match nuke distribution`.
Fastlane deletes certificates from Apple but leaves them in S3 because it fails here.
Access to S3 is read/write and checked by previous runs of `fastlane match appstore`.

##### Complete output when running fastlane, including the stack trace and command used
<details>
  <pre>
irix@tau ~/projects/REDPROJECT/ios [∓ ci +] % fastlane match nuke distribution
[✔] 🚀
[13:46:44]: fastlane detected a Gemfile in the current directory
[13:46:44]: However, it seems like you didn't use `bundle exec`
[13:46:44]: To launch fastlane faster, please use
[13:46:44]:
[13:46:44]: $ bundle exec fastlane match nuke distribution
[13:46:44]:
[13:46:44]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[13:46:46]: Successfully loaded '/Users/irix/projects/REDPROJECT/ios/fastlane/Matchfile' 📄

+----------------------+------------------------------------------+
|           Detected Values from './fastlane/Matchfile'           |
+----------------------+------------------------------------------+
| storage_mode         | s3                                       |
| s3_bucket            | REDBUCKET                                |
| s3_region            | eu-west-1                                |
| s3_access_key        | RED                                      |
| s3_secret_access_key | RED                                      |
| type                 | appstore                                 |
+----------------------+------------------------------------------+

[13:46:50]: 🔓  Successfully decrypted certificates repo

+----------------------------+------------------------------------------+
|                    Summary for match nuke 2.144.0                     |
+----------------------------+------------------------------------------+
| type                       | appstore                                 |
| readonly                   | false                                    |
| generate_apple_certs       | true                                     |
| skip_provisioning_profiles | false                                    |
| username                   | RED                                      |
| team_id                    | RED                                      |
| storage_mode               | s3                                       |
| git_branch                 | master                                   |
| shallow_clone              | false                                    |
[...]

Issue #16210, opened 2020-03-26
match always creates new certificates for appstore, fails on readonly [XCode 11]


warrify:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

We set up match in our project. Initially we ran `bundle exec fastlane match nuke` which deleted all certificates and profiles. After that I set up an S3 Bucket for storage and created a Matchfile. Running `bundle exec fastlane match appstore` created the corresponding certificates, keys and profiles which showed up both in S3 and my keystore. I am then able to run gym locally and my app gets signed with the correct key. However whenever I run match again it first downloads all certificates and keys and then tries to create a new code signing identity. When I add the readonly flag it fails with the following output.

##### Command executed

```bash
bundle exec fastlane ios certificates
```
##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> 
bundle exec fastlane ios certificates --verbose                               
[✔] 🚀 
INFO [2020-03-26 17:05:22.60]: Successfully loaded custom action '/Users/user/Documents/app-name/app/fastlane/actions/commit_timestamp.rb'.
DEBUG [2020-03-26 17:05:22.60]: Checking if there are any plugins that should be loaded...
DEBUG [2020-03-26 17:05:22.60]: Loading 'fastlane-plugin-versioning_android' plugin
INFO [2020-03-26 17:05:22.60]: gem 'fastlane-plugin-versioning_android' is already installed
+------------------------------------+---------+----------------------------------------------------+
|                                           Used plugins                                            |
+------------------------------------+---------+----------------------------------------------------+
| Plugin                             | Version | Action                                             |
+------------------------------------+---------+----------------------------------------------------+
| fastlane-plugin-versioning_android | 0.1.0   | android_get_version_name android_set_version_code  |
|                                    |         | android_set_version_name android_get_version_code  |
+------------------------------------+---------+----------------------------------------------------+

Successfully loaded Appfile at path '/Users/user/Documents/app-name/app/fastlane/Appfile'
- app_identifier: '...'
- apple_id: '...'
- itc_team_id: '...'
- team_id: '...'
- package_name: '...'
-------
[...]

PR #16335, opened 2020-04-18
[match] Fix S3 errors for match nuke and sync

youngern:
### Checklist

### Motivation and Context

* When using S3 as storage for `match`, error occurs when calling command `match nuke <env>`

* When downloading files, the file path is split using delimiter string `s3.amazonaws.com` which fails when the url specifies a region i.e. `s3.us-east-1.amazonaws.com`
As a result, match thinks there are no files in s3 and tries to recreate certificates. This fails when `readonly: true`, and it also creates Profile-Cert mismatch when `readonly: false`

#### Github Issues:
[match nuke fails with deleting from S3](https://github.com/fastlane/fastlane/issues/16216)
[match always creates new certificates for appstore, fails on readonly [XCode 11]](https://github.com/fastlane/fastlane/issues/16210)
### Description

Uses correct `bucket_name` variable in `helper/s3_client_helper.rb`
Use `object.key` instead of parsing `public_url` in `S3Storage`
For `delete_files` in `S3Storage`, passes prefixed working directory

Adds `spec/helper/s3_client_helper_spec.rb`
Adds `match/spec/storage/s3_storage_spec.rb`
Added tests for methods `#find_bucket!` and `#delete_file` for integration and behavior
Since `delete_file` is called per-file, opted to use `bucket.object(key)` vs `bucket.objects[key]`

### Testing Steps

* `bundle exec rspec`
* Tested locally on machine against local fastlane setup - was able to successfully run `match nuke development`
* Tested locally on machine against local fastlane setup - was able to successfully run `match development` and `match development --readonly true`

### DIFF
--- a/fastlane/lib/fastlane/helper/s3_client_helper.rb
+++ b/fastlane/lib/fastlane/helper/s3_client_helper.rb
@@ -7,16 +7,14 @@ module Fastlane
 
       def initialize(access_key: nil, secret_access_key: nil, region: nil)
         creds = Aws::Credentials.new(access_key, secret_access_key)
-        Aws.config.update({
+        Aws.config.update(
           region: region,
           credentials: creds
-        })
-
-        @client = Aws::S3::Client.new
+        )
       end
 
       def list_buckets
-        return @client.list_buckets
+        return client.list_buckets
       end
 
       def upload_file(bucket_name, file_name, file_data, acl)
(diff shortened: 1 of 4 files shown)


youngern:
@googlebot I signed it!

joshdholtz (approved):
This looks amazing! Thanks so much for fix all of this and adding tests ❤️ I appreciate the contribution!

AustinFuller:
@youngern Thanks so much for this contribution! I was hitting these errors and it was driving me crazy. Very excited to use this to get my deployment scripts working again 👍

PR #16345, opened 2020-04-20
[docs][gym] Add info on passing plist into export_options on gym

### LINKED ISSUE #16327: gym uses provisioning profiles from match automatically
### Question Checklist

I'm using match and gym back to back, with more than 1 app ids:
```
  match(
    type: "adhoc",
    app_identifier: [
      id_one,
      id_two,
      id_three
    ],
    readonly: true
  )

  gym(
    project: "correct_project_path",
    scheme: "correct_scheme",
    configuration: "Debug",
    export_method: "ad-hoc",
    export_options: {
                      provisioningProfiles: { "id_one" => "match AdHoc id_one" }
                    }
  )
```

gym is disregarding what I pass in `export_options.provisioningProfiles` and uses what's coming from match: all 3 id's have mapping.

Is it possible to override what match passes, so id_two and id_three doesn't have matchings?

b-onc:
### Checklist

### Motivation and Context
Fixes #16327 

### Description
There's this small part in the source code of `build_app`:
```ruby
if Actions.lane_context[SharedValues::MATCH_PROVISIONING_PROFILE_MAPPING]
          # Since Xcode 9 you need to explicitly provide the provisioning profile per app target
          # If the user is smart and uses match and gym together with fastlane, we can do all
          # the heavy lifting for them
          values[:export_options] ||= {}
          # It's not always a hash, because the user might have passed a string path to a ready plist file
          # If that's the case, we won't set the provisioning profiles
          # see https://github.com/fastlane/fastlane/issues/9490
          if values[:export_options].kind_of?(Hash)
```

Well, using a plist correctly overrides the options match passes, but this should be more apparent in documentation. For example, in https://docs.fastlane.tools/actions/gym/ there's the sentence:
> Note: If you use fastlane with match you don't need to provide those values manually.

This should be changed to:
> Note: If you use fastlane with match you don't need to provide those values manually, unless you pass a plist file into `export_options`


b-onc:
@googlebot I signed it!

janpio (approved):

joshdholtz (approved):
This makes sense to me! Thanks for contribution ❤️

PR #16375, opened 2020-04-24
[scan] Add disable_xcpretty option that skips it post-xcodebuild too

### LINKED ISSUE #16372: xcpretty is ran at the end of scan when output_style is raw (which is meant to disable xcpretty)
### New Issue Checklist

### Issue Description
If I specify `output_style: 'raw'` on `scan` then xcpretty is ripped out from the xcodebuild pipe, as expected (per docs, it disables xcpretty). Because I'm not interested in xcpretty's functionality, I use `trainer` to generate my junit report. However, when the tests finish, I see this printed:
```
Generating test results. This may take a while for large projects.
```
Looking at [the code that prints that message](https://github.com/fastlane/fastlane/blob/0e0c881eefdc54105a1a0062460c0d0e5daf9396/scan/lib/scan/runner.rb#L150-L156), the allegedly expensive operation (per _"may take a while"_) running under the hood is xcpretty, something I've opted out and the reason why this `temp_junit_report` wasn't found.

Furthermore, the printed test results table says 0 tests due to an [xcpretty issue](https://github.com/xcpretty/xcpretty/issues/295) so this final run isn't beneficial anyway.
```
+--------------------+---+
|      Test Results      |
+--------------------+---+
| Number of tests    | 0 |
| Number of failures | 0 |
+--------------------+---+
```

### Possible solution
For my case specifically, I don't need xcpretty to run at all because trainer will create the junit (faster) and report what I need, so skipping that `test_results` method when `output_style='raw'` works for me. Apparently there are people who "disable xcpretty" (output_style=raw) but still want it to run at the end in order to get the test results table as [there's a unit test verifying just that](https://github.com/fastlane/fastlane/blob/0e0c881eefdc54105a1a0062460c0d0e5daf9396/scan/spec/runner_spec.rb#L94-L95). So we can make this a parameter; I actually prefer it to be because it would also allow for clearer documentation.

As of today, gym uses `disable_xcpretty` for the same purpose that scan uses `output_style='raw'`. Maybe scan could bring that option in, then force output_style to `raw` if `disable_xcpretty`. People who still want xcpretty to run at the end would still use `output_style='raw'`, and people like me who don't, would migrate to `disable_xcpretty`

Thank you for reading 🙏

gobetti:
### Checklist

### Description
Closes #16372 - see issue for motivation and context

### DIFF
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -180,6 +180,11 @@ module Scan
                                      description: "Should the HTML report be opened when tests are completed?",
                                      is_string: false,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :disable_xcpretty,
+                                     env_name: "SCAN_DISABLE_XCPRETTY",
+                                     description: "Disable xcpretty formatting of build, similar to `output_style='raw'` but this will also skip the test results table",
+                                     type: Boolean,
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :output_directory,
                                      short_option: "-o",
                                      env_name: "SCAN_OUTPUT_DIRECTORY",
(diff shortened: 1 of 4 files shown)


gobetti:
@googlebot I signed it!

gobetti:
the failing check doesn't seem related to this PR, I'm seeing in the log `Homebrew must be run under Ruby 2.6! You're running 2.3.3.`

janpio (approved):

PR #16386, opened 2020-04-27
[match] profile name parameter to specify your own provisioning profile name

josherbushel:
### Checklist

### Motivation and Context

The default provisioning profile names can sometimes exceed Apple's allowed length (50 characters). While this works, transferring an app to a different account can cause long profile names to become hidden and thus unable to be deleted. It'd be nice to specify your own custom profile name as you/your company could create your own naming format and/or trim long names to Apple's allowed length. 

### Description

I added a profile_name parameter to the Match tool. If this is specified, it will use that instead of the default provisioning profile. I added the option in options.rb and modified the generator.rb. I tested it by running fastlane match --profile_name "Your Profile Name". Below is a video of how it works. First command is just the default. 
https://www.dropbox.com/s/7m7ordts0d9pczo/FastlaneProfileName.mov?dl=0

### Testing Steps

Run fastlane match --profile_name "Your own profile name" in the terminal, or add profile_name parameter in the fastfile.


josherbushel:
@googlebot I signed it!

craigomac:
Thanks for this PR—it'll solve an issue we're experiencing where we have a "hidden" provisioning profile with the name Match wants to use. Apple won't remove it (we've tried asking).

janpio (approved):

PR #16401, opened 2020-04-29
[upload_symbols_to_crashlytics] Include Path Suggested by Carthage Instructions

Blackjacx:
### Checklist

### Motivation 
I use Carthage and the upload_symbols command is not found since it is not packaged with the framework for [some reasons](https://github.com/firebase/firebase-ios-sdk/issues/4720). So I think it is a good idea to add the path suggested in the [docs](https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md) to the search paths. This way fastlane can automatically find the executable.

### Description
Includes the [FirebaseCrashlytics upload_symbols directory suggested by the fastlane docs regarding integration with Carthage](https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md). This one is preferred over the Fabric installation.

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -130,7 +130,7 @@ module Fastlane
       end
 
       def self.find_binary_path(params)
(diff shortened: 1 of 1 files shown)


Blackjacx:
@googlebot I signed it!

Blackjacx:
Maybe it is a good idea to put the scripts in a subdirectory like `./scripts/firebase-crashlytics/` to provide some grouping-context? If you agree I would also like to update the [docs](https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md).

joshdholtz (approved):
This seems great to me! Thanks for the contribution ❤

PR #16410, opened 2020-04-30
[notarize] Add parameter to select if compressed package should be removed

rafcabezas:
### Checklist

### Motivation and Context
Was not able to distribute my zipped up app after notarization, as notarize removes the zip file it generates, which is being notarized as well.

### Description
The problem is that notarize creates a zip file that is sent to the notarization service. The zip file is then also notarized.
```
{
  "logFormatVersion": 1,
  "jobId": "48e1bb69-e2aa-47ec-8c82-adc340aaa88a",
  "status": "Accepted",
  "statusSummary": "Ready for distribution",
  "statusCode": 0,
  "archiveFilename": "AppName.app.zip",
  "uploadDate": "2020-04-30T18:23:09Z",
  "sha256": "1d88ea6a1af6ad82fa7c23b49e81da4cfaaace1637a2df26819f9b91158b6d68",
```

Then, notarize removes the zip file. If a new zip file is then created to distribute the app, it fails notarization (at least on macOS 10.15.4)

This PR provides a new configuration parameter that can be used to prevent the zip file from being removed. This notarized zip file can then be used for distribution.

I tried setting the flag to true and to false, and it behaves as expected.
Also, I set the default to "true" so notarize will continue to work the same way unless the parameter is set to false.

Regarding documentation, I documented the parameter in the code by adding it to the available options. Not sure what other documentation might be required for this small change.

### Testing Steps
- Set the parameter `remove_compressed_package` to `false`
- Run notarize 
- Verify that the generated zip file is not removed.

### DIFF
--- a/fastlane/lib/fastlane/actions/notarize.rb
+++ b/fastlane/lib/fastlane/actions/notarize.rb
@@ -7,6 +7,7 @@ module Fastlane
         try_early_stapling = params[:try_early_stapling]
         print_log = params[:print_log]
         verbose = params[:verbose]
+        remove_compressed_package = params[:remove_compressed_package]
 
         # Compress and read bundle identifier only for .app bundle.
         compressed_package_path = nil
@@ -44,7 +45,7 @@ module Fastlane
           log: verbose
         )
 
-        FileUtils.rm_rf(compressed_package_path) if compressed_package_path
+        FileUtils.rm_rf(compressed_package_path) if compressed_package_path && remove_compressed_package
 
         notarization_upload_plist = Plist.parse_xml(notarization_upload_response)
         notarization_request_id = notarization_upload_plist['notarization-upload']['RequestUUID']
@@ -167,6 +168,12 @@ module Fastlane
                                        description: 'Whether to log requests',
                                        optional: true,
                                        default_value: false,
+                                       type: Boolean),
+          FastlaneCore::ConfigItem.new(key: :remove_compressed_package,
+                                       env_name: 'FL_REMOVE_COMPRESSED_PACKAGE',
+                                       description: 'Whether to remove generated compressed package',
+                                       optional: true,
+                                       default_value: true,
                                        type: Boolean)
         ]
       end


rafcabezas:
@googlebot I signed it!

fastlane/lib/fastlane/actions/notarize.rb
```diff
@@ -44,7 +45,7 @@ def self.run(params)
           log: verbose
         )
 
-        FileUtils.rm_rf(compressed_package_path) if compressed_package_path
+        FileUtils.rm_rf(compressed_package_path) if compressed_package_path && remove_compressed_package
```
joshdholtz, line 48: Question! Instead of introduction a new package `remove_compressed_package `, could we change this to `package_path != compressed_package_path `?

This why the user doesn’t need to set any options or anything 😇 Thoughts?

joshdholtz (commented):
Just a quick question if we even need a new option 🙃

fastlane/lib/fastlane/actions/notarize.rb
rafcabezas: But `compressed_package_path` is `package_path`+`.zip`. So this comparison would always be true.

rafcabezas: We could add a parameter to specify the `compressed_package_path` and then do something like that?

joshdholtz: Ohhh, one more question which might help me out 🙃 What is the purpose of uploading the `.zip` file? After notarization, a ticket gets stapled to the `.app` so I _think_ you don’t even need the `.zip`... unless there is something I’m missing which is totally possible 😛

rafcabezas: Sure! So notarization is for distributing apps outside of the AppStore, and you can choose to distribute the app via zip, if that's the case, the zip that was notarized is the one that needs to be distributed. Otherwise it won't run.
I use app center for distribution, and the appcenter plugin needs a zip.
I invoke it like this after notarization:
```fastfile
appcenter_upload(
      api_token: "aaabbbcccdddeeefff",
      owner_name: "username",
      owner_type: "user",
      app_name: "AppName",
      file: "artifacts/AppName.app.zip"
    )
```

joshdholtz: Ah, weird! I’ve notarized a macOS app before and the ticket was getting stapled to my `.app`.  You can see below that the `package_path` is the one getting stapled with the ticket so you shouldn’t need the zip 🤔 

https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/notarize.rb#L111

rafcabezas: Yea I don't know why it happens. I can give it another try now and tell you the exact error I'm seeing.

rafcabezas: Well, it worked this time 😓 haha. Thanks for making me check! 
I zipped up the file for distribution after notarizing and uploaded it. I don't know if it's because I used ditto to zip up like notarize does instead of zip. But it's working now. I'm gonna go ahead and close this PR. Thanks for your time!

Issue #16352, opened 2020-04-21
Add support for user refresh tokens to authenticate supply


moly:
### Feature Request
Add support for user refresh tokens to authenticate supply. User refresh tokens are described here: https://developers.google.com/android-publisher/authorization

#### Motivation Behind Feature
In some corporate setups the play store page is not owned/managed internally and it is therefore difficult/impossible to get access to the service account. Adding support for alternative authentication methods would make fastlane more useful in these situations.

#### Feature Description
It would be useful if supply was able to accept a json file containing refresh token details and use this to authenticate with the google api. The code change required would be very small, this would need to be added to `initialize` in `AbstractGoogleServiceClient`:

```ruby
if is_refresh_token
    auth_client = Google::Auth::UserRefreshCredentials.make_creds(json_key_io: service_account_json, scope: self.class::SCOPE)
else
    auth_client = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: service_account_json, scope: self.class::SCOPE)
end
```

`is_refresh_token` could be set by a command line switch, or determined automatically by inspecting the provided json.

#### Alternatives or Workarounds
None that I know of.

PR #16414, opened 2020-05-01
[supply] added support for refresh tokens as an authentication method

moly:
### Checklist

### Motivation and Context
Resolves issue #16352. In some corporate setups the play store page is not owned/managed internally, which can make it difficult to get access to the service account. Adding support for alternative authentication methods would make fastlane more useful in these situations.

### Description
Adds two new parameters to the supply action, `json_token` and `json_token_data`. The former allows the user to provide a path to a json file containing the refresh token data, while the latter allows specifying the raw refresh token string directly. If either of these parameters are present, the supply client will use `Google::Auth::UserRefreshCredentials.make_creds()` to authenticate with the play store api.

### Testing Steps
Generate a refresh token using the steps here: https://developers.google.com/android-publisher/authorization#generating_a_refresh_token

Create a json file with the following format:
```
{
  "grant_type" : "refresh_token",
  "client_id" : <the client ID token created in the APIs Console>,
  "client_secret" : <the client secret corresponding to the client ID>,
  "refresh_token" : <the refresh token from the previous step>
}
```
Run: `bundle exec fastlane supply --json-token /path/to/json`

You can also specify the parameter using the `SUPPLY_JSON_TOKEN` environment variable or by setting `json_token_file` in your AppFile.


moly:
@googlebot I signed it!

janpio:
Nice.

Is `json_token` the right name for this?

moly:
I named it `json_token` to keep in line with the existing `json_key` parameter. I'm happy to change it if you have any suggestions?

supply/lib/supply/options.rb
```diff
@@ -116,6 +117,35 @@ def self.available_options
                                          UI.user_error!("Could not parse service account json  JSON::ParseError")
                                        end
                                      end),
+        FastlaneCore::ConfigItem.new(key: :json_token,
```
joshdholtz, line 120: I think it make sense to call this `refresh_token`. `json_token` is a bit to similar to `json_key` and I think it would cause some confusion 😊

supply/lib/supply/options.rb
```diff
@@ -116,6 +117,35 @@ def self.available_options
...
+                                     optional: true,
+                                     description: "The path to a file containing refresh token JSON, used to authenticate with Google",
+                                     code_gen_sensitive: true,
+                                     default_value: CredentialsManager::AppfileConfig.try_fetch_value(:json_token_file),
+                                     default_value_dynamic: true,
+                                     verify_block: proc do |value|
+                                       UI.user_error!("Could not find refresh token json file at path '#{File.expand_path(value)}'") unless File.exist?(File.expand_path(value))
+                                       UI.user_error!("'#{value}' doesn't seem to be a JSON file") unless FastlaneCore::Helper.json_file?(File.expand_path(value))
+                                     end),
+        FastlaneCore::ConfigItem.new(key: :json_token_data,
```
joshdholtz, line 133: And this be `refresh_token_data`

supply/lib/supply/options.rb
```diff
@@ -116,6 +117,35 @@ def self.available_options
...
                                        end
                                      end),
+        FastlaneCore::ConfigItem.new(key: :json_token,
+                                     env_name: "SUPPLY_JSON_TOKEN",
+                                     short_option: "-o",
+                                     conflicting_options: [:issuer, :key, :json_key, :json_key_data, :json_token_data],
+                                     optional: true,
+                                     description: "The path to a file containing refresh token JSON, used to authenticate with Google",
+                                     code_gen_sensitive: true,
+                                     default_value: CredentialsManager::AppfileConfig.try_fetch_value(:json_token_file),
```
joshdholtz, line 127: This should look for same name as the option so `refresh_token`

supply/lib/supply/options.rb
```diff
@@ -116,6 +117,35 @@ def self.available_options
...
+                                       UI.user_error!("'#{value}' doesn't seem to be a JSON file") unless FastlaneCore::Helper.json_file?(File.expand_path(value))
+                                     end),
+        FastlaneCore::ConfigItem.new(key: :json_token_data,
+                                     env_name: "SUPPLY_JSON_TOKEN_DATA",
+                                     short_option: "-g",
+                                     conflicting_options: [:issuer, :key, :json_key, :json_key_data, :json_token],
+                                     optional: true,
+                                     description: "The raw refresh token JSON data used to authenticate with Google",
+                                     code_gen_sensitive: true,
+                                     default_value: CredentialsManager::AppfileConfig.try_fetch_value(:json_token_data_raw),
```
joshdholtz, line 140: And this `refresh_token_data`

credentials_manager/lib/credentials_manager/appfile_config.rb
```diff
@@ -132,6 +132,14 @@ def json_key_data_raw(*args, &block)
       setter(:json_key_data_raw, *args, &block)
     end
 
+    def json_token_file(*args, &block)
```
joshdholtz, line 135: `refresh_token`

credentials_manager/lib/credentials_manager/appfile_config.rb
```diff
@@ -132,6 +132,14 @@ def json_key_data_raw(*args, &block)
...
+      setter(:json_token_file, *args, &block)
+    end
+
+    def json_token_data_raw(*args, &block)
```
joshdholtz, line 139: `refresh_token_data`

supply/lib/supply/client.rb
```diff
@@ -45,7 +45,11 @@ def self.service_account_authentication(params: nil)
     # Initializes the service and its auth_client using the specified information
     # @param service_account_json: The raw service account Json data
     def initialize(service_account_json: nil, params: nil)
-      auth_client = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: service_account_json, scope: self.class::SCOPE)
+      if params[:json_token] || params[:json_token_data]
+        auth_client = Google::Auth::UserRefreshCredentials.make_creds(json_key_io: service_account_json, scope: self.class::SCOPE)
+      else
+        auth_client = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: service_account_json, scope: self.class::SCOPE)
```
joshdholtz, line 51: This method is now failing for with `NoMethodError: [!] undefined method `gsub' for nil:NilClass `. Did this break for you?

joshdholtz (changes requested):
This is super cool! I wasn’t able to get this to work tonight. I got an error of `[!] Google Api Error: forbidden: The caller does not have permission - The caller does not have permission`. So that is probably on me somehow but I’m not sure exactly what yet. But I did leave some questions and requested some changes 😊 

Thanks for all the work on this!

supply/lib/supply/client.rb
moly: Good catch, I'd accidentally caused `service_account_authentication` to return `nil` when using a json key. Should be fixed now.

moly:
Renamed all instances of `json_token` and `json_token_data` to `refresh_token` and refresh_token_data` as requested.

joshdholtz:
@moly Thanks for making those changes! Do you have any idea what could be causing my permission issue I mentioned above? 😇 I’m stumped lol

moly:
@joshdholtz does your Google account have the right permissions for the android project you're using? It might need to have the "release manager" role or higher. I'll check my setup tomorrow and provide better step by step instructions.

moly:
@joshdholtz I set up a new app today and couldn't reproduce your permission issue. Based on the error message, the only things I could think of is that either the google account you're using doesn't have the right permissions for the play store app you're targeting, or you didn't allow access to the api project on the oauth consent screen correctly. Here's the steps I used, my google account is the owner of the play store app.

1) Go to https://play.google.com/apps/publish/#ApiAccessPlace. Click "Create OAuthClient"
2) If you haven't before you may need to create an oauth consent screen. You can give it a name but leave everything else as it is and click save.
3) https://play.google.com/apps/publish/#ApiAccessPlace should now show "Google Play Android Developer" under linked projects and a new client under "OAuth Clients". Under "OAuth Clients" click "view in Google Developers Console", then click "Google Play Android Developer" under "OAuth 2.0 Client IDs". You should be able to see the client ID and client secret.
4) Go to the following url, replacing YOUR_CLIENT_ID with the client ID from the last step: https://accounts.google.com/o/oauth2/auth?scope=https://www.googleapis.com/auth/androidpublisher&response_type=code&access_type=offline&redirect_uri=urn:ietf:wg:oauth:2.0:oob&client_id=YOUR_CLIENT_ID
5) Click to allow the Google Play Android Developer project permission to manage your android apps.
6) You should get redirected to a page with an authorisation code. Run the following command:
`curl -s -d "grant_type=authorization_code&code=YOUR_AUTHORISATION_CODE&client_id=YOUR_CLIENT_ID&client_secret=YOUR_CLIENT_SECRET&redirect_uri=urn:ietf:wg:oauth:2.0:oob" -X POST https://accounts.google.com/o/oauth2/token`
7) The response should contain your refresh token. Create a json file with the following format:
`
{
  "grant_type" : "refresh_token",
  "client_id" : "YOUR_CLIENT_ID",
  "client_secret" : "YOUR_CLIENT_SECRET",
  "refresh_token" : "YOUR_REFRESH_TOKEN"
}
`
8) Save the json somewhere in the project directory, then run `fastlane supply init --refresh-token path/to/refresh_token.json`.

moly:
Refactored a little to fix some failing tests

iBotPeaches:
Looks like we handle refresh, authorized, adc creds and more via #29982 now. Closing this out. Sorry it never got through the finish line. It positively looks solved now.

PR #16421, opened 2020-05-04
[action] set_pod_key: remove shellescape of params

arnauddorgans:
### Checklist

### Motivation and Context
I use cocoapods-keys to store API keys but using `set_pod_key` action automatically escapes the provided parameters.
So each time I provide an API key that contains escapable characters, the API key becomes malformed.

### Description
Remove shellescape

### Testing Steps
Before:
```ruby
set_pod_key(key: "PublicKey", value: "XXXX|XXXX", project: "Y")
set_pod_key(key: "ApiKey", value: "XX/nXX==-XX", project: "Y")
```
Capture d’écran 2020-05-04 à 14 44 37_censored

After:
```ruby
set_pod_key(key: "PublicKey", value: "XXXX|XXXX", project: "Y")
set_pod_key(key: "ApiKey", value: "XX/nXX==-XX", project: "Y")
```
Capture d’écran 2020-05-04 à 14 43 45_censored

### DIFF
--- a/fastlane/lib/fastlane/actions/set_pod_key.rb
+++ b/fastlane/lib/fastlane/actions/set_pod_key.rb
@@ -8,9 +8,9 @@ module Fastlane
         cmd << ['bundle exec'] if params[:use_bundle_exec] && shell_out_should_use_bundle_exec?
         cmd << ['pod keys set']
 
-        cmd << ["\"#{params[:key].shellescape}\""]
-        cmd << ["\"#{params[:value].shellescape}\""]
-        cmd << ["\"#{params[:project].shellescape}\""] if params[:project]
+        cmd << ["\"#{params[:key]}\""]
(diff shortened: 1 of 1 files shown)


arnauddorgans:
@googlebot I signed it!

janpio:
Are you sure the current behavior is not just a bug that should be fixed? (I am missing the context what this is usually used for)

arnauddorgans:
You are right, there is no point to shellescape parameters, event if we directly call the action from terminal we have to escape our values first overwise the command will fail.

Test:
fail
success

I did update the PR with the shellescape deletion

joshdholtz (approved):
This makes sense to me! The shellescape was definitely not needed since it was already wrapped in quotes 🤷‍♂️ Thanks for fixing! ❤

PR #16444, opened 2020-05-10
[frameit] add iPad Pro 12.9 4th generation frame

bouk:
As the title says. Currently the old ipad pro frame is used when framing ipad pro 12.9 4th gen screenshots


joshdholtz (approved):
Thanks for adding this! ❤️

Issue #15747, opened 2019-12-02
Scan tool works on local machine with "xctestrun" parameter, but not working with teamcity


elidir:
### Question Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Question Subject
Scan with "xctestrun" parameter works on local machine but fail with teamcity

#### Question Description
I'm trying to run tests from teamcity using fastlane scan action. Before running tests I use the "xcodebuild" action with the "build_for_testing" parameter to get the "xctestrun" file. 
Then I save a path to xctesrun file to .env file.

In my lane unitTesting I use the "xctestrun" parameter ("fastlane action scan" says that it exists) and pass full path to the file using .env

xctestrun_path = ENV['XCTestRun_Path']

lane :unitTesting do
    parameters = {
      scheme: SHEME_NAME,
      # test_without_building: true, (no effect if enabled)
      only_testing: 'MyAppUnitTests',
      output_files: 'unittests/report.junit',
      output_types: 'junit',
      destination: 'platform=iOS Simulator,name=iPhone Xʀ',
      # derived_data_path: "DerivedData", (no effect if enabled)
      xctestrun: xctestrun_path
    }
    puts("Unit tests - parameters")
    puts(parameters)
    scan(parameters)
  end

This lane works on my local machine or on my build agent machine form terminal:
"bundle exec fastlane ios unitTesting"
But the same line is not working form temcity.

I added puts(parameters) to my Fistfile to see them in teamcity build logs and there is no "xctestrun" value:

Unit tests - parameters
[12:47:36][Step 12/18] [16:47:36]: {:scheme=>"MyAppShared", :only_testing=>"MyAppUnitTests", :output_files=>"unittests/report.junit", :output_types=>"junit", :destination=>"platform=iOS Simulator,name=iPhone Xʀ", :xctestrun=>""}

And this is full xcodebuild command (with empty "xctesrun" parameter):

[...]

PR #16446, opened 2020-05-11
[GH Actions] Enable lock action

mollyIV:
### Description

We are working on a bot's code migration to GitHub Actions 👉 fastlane/github-actions#3.
This PR is about enabling locking closed, inactive issues and pull requests workflow.

Enabling this workflow will not conflict with an issue-bot processing flow. 

### Testing Steps

The workflow file used for testing the action 👇

```yml
name: Locking issues and pull requests
on:
  schedule:
  - cron: "0 0 * * *"

jobs:
  lock:
    runs-on: ubuntu-latest
    steps:
    - uses: fastlane/github-actions/lock@latest
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
        days-before-lock: 1
```

image

image

> BOT_GITHUB_TOKEN is set to my personal access token, that's why you see my personal avatar and "Thanks for your contribution."  comment in a pull request comes from a different GH Action.


lock.yml
```diff
@@ -0,0 +1,14 @@
...
+  - cron: "0 0 * * *"
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: fastlane/github-actions/lock@latest
+      with:
+        repo-token: ${{ secrets.BOT_GITHUB_TOKEN }}
+        days-before-lock: 60
```
mollyIV, line 14: The time before locking is taken from an original bot's implementation:

https://github.com/fastlane/issue-bot/blob/457348717d99e5ffde34ca1619e7253ed51ec172/bot.rb#L286-L292

joshdholtz:
@mollyIV Heyyy 😊 I'm slightly torn about this one, TBH 🤔 I've seen quite a bit of backlash about locking issues and PR lately. Closing seems totally okay but I'm not sure if we should auto lock since it prohibits communication and than can frustrate some members of the community. Do you have any thoughts or experienced with locked issues in other communities?

mollyIV:
Hello Josh 👋 Actually the current implementation of our active bot is locking the issues:

https://github.com/fastlane/issue-bot/blob/457348717d99e5ffde34ca1619e7253ed51ec172/bot.rb#L139

Thought that we want to keep that functionality - that's why we've created a [lock GH Action](https://github.com/fastlane/github-actions/tree/master/lock) 😊  

> Do you have any thoughts or experienced with locked issues in other communities?

TBH I am okay with locking the closed issues after a bit longer time, like 2 months or something. Opening new issues might be more appreciated than digging deep into old, already closed conversation, wdyt? 🤔

janpio (commented):
Moving more logic into the GH Actions makes sense 👍

janpio (approved):
Final confirmation before merging @mollyIV: This just recreates already existing behavior?

mollyIV:
According to [the source code of an existing bot](https://github.com/fastlane/issue-bot/blob/457348717d99e5ffde34ca1619e7253ed51ec172/bot.rb#L287-L304) - yes, it does. Here you can find an example of locked issue via the existing bot:
* https://github.com/fastlane/fastlane/issues/15747

PR #16448, opened 2020-05-11
[match] prompt for keychain password to set partition list for certificates to disable xcode prompt for password on sign

### LINKED ISSUE #14196: Fastlane match isn't prompting me for my keychain password
I'm using fastlane 2.115.0.

When I run `fastlane match appstore` and enter the various details, eventually I get:

```
[13:26:19]: Could not configure imported keychain item (certificate) to prevent UI permission popup when code signing
Check if you supplied the correct `keychain_password` for keychain: `/Users/karolis/Library/Keychains/login.keychain-db`
security: SecKeychainItemSetAccessWithPassword: The user name or passphrase you entered is not correct.
[13:26:19]:
[13:26:19]: Please look at the following docs to see how to set a keychain password:
[13:26:19]:  - https://docs.fastlane.tools/actions/sync_code_signing
[13:26:19]:  - https://docs.fastlane.tools/actions/get_certificates
```

Am I doing something wrong?

It never asked me for my keychain password. And _I do not want to_ put the password into the `-p` flag in the terminal, I shouldn't need to be doing that.

Note, if I do put my password into `-p`, it all works. But this seems like a bug / regression? It should prompt me for password in a MacOS dialogue?

joshdholtz:
### Motivation and Context
Fixes #14196
Entering the keychain password used for certificate signing (usually login) required storing the keychain password in an environment variable or passing it in via options which was pretty insure. This will allow for a prompt of the password and store the password for the keychain in the keychain to use later.

### Description
1. Uses password if given
2. Uses keychain password from login keychain if found
3. Prompts user for keychain password and stores it in login keychain for user later

Also fixed broken match tests that were using `allow` instead of `expect` for some `Open3` commands. The `Open3` command weren’t getting mocked properly so the inside was never getting executed which means not all of the things were getting tested properly 🤷‍♂️ 

### Testing Steps
1. Nuke all existing certs
2. Run match
  - Will prompt for keychain password
3. Look in Keychain Access for `fastlane_keychain_<keychain_name>` entry
4. Nuke all certs again
5. Run match again
  - There won’t be any prompt

### DIFF
--- a/fastlane_core/lib/fastlane_core/keychain_importer.rb
+++ b/fastlane_core/lib/fastlane_core/keychain_importer.rb
@@ -1,9 +1,10 @@
 require_relative 'helper'
 require 'open3'
+require 'security'
 
 module FastlaneCore
   class KeychainImporter
-    def self.import_file(path, keychain_path, keychain_password: "", certificate_password: "", output: FastlaneCore::Globals.verbose?)
+    def self.import_file(path, keychain_path, keychain_password: nil, certificate_password: "", output: FastlaneCore::Globals.verbose?)
       UI.user_error!("Could not find file '#{path}'") unless File.exist?(path)
 
       command = "security import #{path.shellescape} -k '#{keychain_path.shellescape}'"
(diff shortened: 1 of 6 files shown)


fastlane_core/lib/fastlane_core/keychain_importer.rb
```diff
@@ -32,7 +33,38 @@ def self.import_file(path, keychain_path, keychain_password: "", certificate_pas
       end
     end
 
-    def self.set_partition_list(path, keychain_path, keychain_password: "", output: FastlaneCore::Globals.verbose?)
+    def self.set_partition_list(path, keychain_path, keychain_password: nil, output: FastlaneCore::Globals.verbose?)
+      keychain_name = File.basename(keychain_path, ".*")
+      server = server_name(keychain_name)
+
+      # https://github.com/fastlane/fastlane/issues/14196
```
snatchev, line 40: Could this lookup of the keychain_password be moved into a function and called where `set_partition_list` gets called? i.e. decouple the lookup from this method (since it already accepts keychain password as a value)

fastlane_core/lib/fastlane_core/keychain_importer.rb
```diff
@@ -53,6 +85,8 @@ def self.set_partition_list(path, keychain_path, keychain_password: "", output:
 
             # Inform user when no/wrong password was used as its needed to prevent UI permission popup from Xcode when signing
             if err.include?("SecKeychainItemSetAccessWithPassword")
+              Security::InternetPassword.delete(server: server)
```
snatchev, line 88: good catch!

snatchev (changes requested):

fastlane_core/lib/fastlane_core/keychain_importer.rb
joshdholtz: Yup yup, can do!

snatchev (approved):
LGTM!

PR #16452, opened 2020-05-12
[supply] Added `in_app_update_priority` option to set the InAppUpdatePriority …

### LINKED ISSUE #16279: Add inAppUpdatePriority option to the upload_to_play_store (supply) action

Google Play Developer API allows specifying the update priority of the android application when creating a new release (see inAppUpdatePriority field in Edits.tracks.releases). 'inAppUpdatePriority' can take int values from 0 (lowest priority) to 5 (highest priority). This priority can be used in the android app for handling/delivering non-critical or critical updates to the users (https://developer.android.com/guide/playcore/in-app-updates). If the priority is set correctly android application can check for available updates + their priorities using client library and then a standard user-friendly system dialog can be shown to the user with the suggestion to update the application. 

#### Feature Description
As a result of specifying inAppUpdatePriority and implementing checking this priority in the android app end-users can see the following standard screen (image is taken from the official android documentation):

screen

Actually I'm not ruby developer so the code example is in Java:
```java
AndroidPublisher publisher = new AndroidPublisher(httpTransport, factory, credentials);

// ... uploading APK here and getting version code
// ------

TrackRelease release = new TrackRelease()
    .setVersionCodes(versionCodes)
    .setInAppUpdatePriority(/* value from 0 to 5 here */); 

Track track = new Track().setReleases(singletonList(release));

publisher.edits().tracks()
    .update(APP_PACKAGE, editId, TRACK_PRODUCTION, track)
    .execute();
```
This feature would not break anything for existing users.

It seems there is no way to do this with existing Fastlane actions & plugins (at least I couldn't find how to do this in the Fastlane documentation and also there is no such option in [sources](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb)).

jomisj:
…for a releaseTrack in the supply tool

### Checklist

### Motivation and Context
Resolves #16279. 

Google Play Developer API allows specifying the update priority of the android application when creating a new release (see inAppUpdatePriority field in Edits.tracks.releases). 'inAppUpdatePriority' can take int values from 0 (lowest priority) to 5 (highest priority). This priority can be used in the android app for handling/delivering non-critical or critical updates to the users (https://developer.android.com/guide/playcore/in-app-updates). If the priority is set correctly android application can check for available updates + their priorities using client library and then a standard user-friendly system dialog can be shown to the user with the suggestion to update the application.

This MR adds a new option `in_app_update_priority` for the supply tool (upload_to_play_store )

### Description
I have added a new option `in_app_update_priority` for the `supply` action by including it in options.rb. I have also added the validation for this new option as it has to be an integer value from 0 to 5. And in uploader.rb, I am using this new option (if set) to set the `AndroidPublisher::TrackRelease.in_app_update_priority` property.
I also had to increment the version on `google-api-client` in fastlane.gemspec file in order to have access to the property `in_app_update_priority` in `AndroidPublisher::TrackRelease`.

`upload_to_play_store.md` have been updated to add documentation for this new option and Fastlane.swift file to have been updated.

I have tested it by uploading new bundles (aab file) to Alpha track in Google PlayStore with different update priority values

### Testing Steps
The below line should upload an app at path `path/app.apk` to track `beta` with an in_app_update_priority of `3`
```
fastlane supply --apk path/app.apk --track beta --in_app_update_priority 3
```

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -96,7 +96,7 @@ Gem::Specification.new do |spec|
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
   # If you upgrade this gem, make sure to upgrade the users of it as well.
-  spec.add_dependency('google-api-client', '>= 0.29.2', '< 0.38.0') # Google API Client to access Play Publishing API
+  spec.add_dependency('google-api-client', '>= 0.37.0', '< 0.39.0') # Google API Client to access Play Publishing API
   spec.add_dependency('google-cloud-storage', '>= 1.15.0', '< 2.0.0') # Access Google Cloud Storage for match
 
   spec.add_dependency('emoji_regex', '>= 0.1', '< 2.0') # Used to scan for Emoji in the changelog
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -96,7 +96,7 @@ Gem::Specification.new do |spec|
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
   # If you upgrade this gem, make sure to upgrade the users of it as well.
-  spec.add_dependency('google-api-client', '>= 0.29.2', '< 0.37.0') # Google API Client to access Play Publishing API
+  spec.add_dependency('google-api-client', '>= 0.29.2', '< 0.38.0') # Google API Client to access Play Publishing API
   spec.add_dependency('google-cloud-storage', '>= 1.15.0', '< 2.0.0') # Access Google Cloud Storage for match
 
   spec.add_dependency('emoji_regex', '>= 0.1', '< 2.0') # Used to scan for Emoji in the changelog
--- a/fastlane/lib/fastlane/actions/docs/upload_to_play_store.md
+++ b/fastlane/lib/fastlane/actions/docs/upload_to_play_store.md
@@ -78,6 +78,12 @@ To gradually roll out a new build use
 fastlane supply --apk path/app.apk --track beta --rollout 0.5
 ```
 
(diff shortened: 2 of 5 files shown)


jomisj:
@googlebot I signed it!

fastlane.gemspec
```diff
@@ -96,7 +96,7 @@ Gem::Specification.new do |spec|
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
   # If you upgrade this gem, make sure to upgrade the users of it as well.
-  spec.add_dependency('google-api-client', '>= 0.29.2', '< 0.37.0') # Google API Client to access Play Publishing API
+  spec.add_dependency('google-api-client', '>= 0.29.2', '< 0.38.0') # Google API Client to access Play Publishing API
```
dlackty, line 99: I wonder if we shall bump the minimal version requirement to `0.37.0`, which contains the support of `in_app_update_priority`

jomisj: Yes. That makes sense. In fact I can change the min to `0.37.0` and max to `< 0.39.0` (0.39.2 is the latest but is only 12 hours old. Whereas the release before 0.39.0 is 0.38.0 and is about one month old)

jomisj: @dlackty I have pushed the change. Thank you for the suggestion.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This is great! Thanks for adding this 😊 I added an `ENV_NAME` to the new option and also removed the changes to `Fastlane.swift` since that is a generated file. Appreciate the contribution ❤️

PR #16470, opened 2020-05-18
[fastlane] fix overly strong substitution of “action” for plugin/actions with the phrase “action” in it

joshdholtz:
### Motivation and Context
An action named `GithubActionAction` would be seen just as `github` when calling `fastlane action` or `fastlane action <action_name>`

### Description
Use `gsub` for replacing `_action` and `Action` at the end of the action/plugin name


nafu (approved):
Looks good to me :octocat:

joshdholtz:
@nafu Thanks for the review! ❤

Issue #16494, opened 2020-05-22
xcov: 'minimum_coverage_percentage' value must be a Float! Found Integer instead


xlsmearlx:
### New Issue Checklist

- [ X] Updated Fastlane to the latest version
- [ X] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ X] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ X] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

**Command**
`bundle exec fastlane qualityCheckLane`

**Error:**
`[16:29:25]: 'minimum_coverage_percentage' value must be a Float! Found Integer instead.`

**Fastfile**
<details>
import Foundation

class Fastfile: LaneFile {
    func qualityCheckLane() {
        desc("This is a lane")
        schemes.forEach(test)
    }
    private func test(scheme: String) {
        scan(
            scheme: scheme,
            clean: true,
            codeCoverage: true,
            outputDirectory: "fastlane/TestData/\(scheme)/scan/",
            outputTypes: "html",
            outputFiles: "scan.html"
        )
        xcov(
            scheme: scheme,
            outputDirectory: "fastlane/TestData/\(scheme)/xcov/",
            minimumCoveragePercentage: 1.1
        )
    }
}
</details>

**Issue in file:**
https://github.com/fastlane/fastlane/blob/34c0b5193f55b209ac2ee363c20e5b4ba7c9502d/fastlane/swift/Fastlane.swift#L8745

### Environment
<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.15.4                                     |
| Ruby                        | 2.7.0                                       |
| Bundler?                    | true                                        |
| Git                         | git version 2.24.3 (Apple Git-128)          |
| Installation Source         | ~/.rvm/gems/ruby-2.7.0/bin/fastlane         |
| Host                        | Mac OS X 10.15.4 (19E287)                   |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.7.0/lib                |
| OpenSSL Version             | OpenSSL 1.1.1b  26 Feb 2019                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
[...]

PR #16496, opened 2020-05-22
Fastlane.swift : Fixes issue with XCOV params 

xlsmearlx:
### Checklist

### Motivation and Context

FastlaneSwift fails using XCOV:

**Issue 1: Hardcoded Xcode path**
Fails with error
```
[16:56:16]: ▸ Loading... 
[16:56:16]: ▸ IDEFoundation could not be loaded IDEFoundation could not be loaded 
[16:56:16]: Exit status: 66
```

**Issue 2: Minimum coverage should be Double/Float as same as the ruby function**
Fails with error
`[16:29:25]: 'minimum_coverage_percentage' value must be a Float! Found Integer instead.`

### Description

Fix: Change `minimumCoveragePercentage: Int` to `minimumCoveragePercentage: Double`

Fix: Change  Xcode path from:
`"/Applications/Xcode-11.4.app/Contents/Developer/../Frameworks/IDEFoundation.framework/Versions/A/IDEFoundation"` 

to 

`ideFoundationPath: String? = nil`

### FIXES ISSUES

- https://github.com/fastlane/fastlane/issues/16495
- https://github.com/fastlane/fastlane/issues/16494


xlsmearlx:
@googlebot I signed it!

janpio (changes requested):
Unfortunately the `Fastlane.swift` file is auto generated and should not be edited manually: https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/fastlane/swift/Fastfile.swift#L1-L4

xlsmearlx:
@janpio is `Fastlane.swift` also autogenerated? the file does not have that message like `Fastfile.swift`. If so, where is the correct place to make this modification to the autogenerated file?

janpio:
You are right, we are missing a helpful message in the file - but it _is_ generated on release/publish (see the history of the file). I think the things you touched come from https://github.com/fastlane/fastlane/blob/1aa8cdd2df87378a487b421607a1532a84f7974a/fastlane/lib/fastlane/actions/xcov.rb#L26-L45 - although some more investigation might be necessary.

janpio:
Let's continue this discussion in the issue 👍

PR #16508, opened 2020-05-24
Rubocop: Enable linter for FastlaneCore::ConfigItem `is_string` in all files

janpio:
We have a linter that checks for the deprecated `is_string` in options, but it is only enabled for some of our tools and actions.

This PR enables it for all our code.

Starting point is:
> 1151 files inspected, 628 offenses detected

## TODO

- [ ] Adapt changes from https://github.com/fastlane/fastlane/pull/13816
- [ ] Fix all the rest

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -294,11 +294,7 @@ CrossPlatform/ForkUsage:
 
 Lint/IsStringUsage:
   Include:
-    - 'cert/**/*'
-    - 'gym/**/*'
-    - 'match/**/*'
-    - 'screengrab/**/*'
-    - 'supply/**/*'
+    - '**/*'
 
 # ( ) for method calls
 Style/MethodCallWithArgsParentheses:


minuscorp:
Does autocorrect fix it? 😅

janpio:
Well, if someone writes a rule and auto-fix for it somehow... I fear this is a bit more manual here though. Maybe @mollyIV or @joshdholtz can describe the "algorithm"? They did this before.

minuscorp:
De we want to just remove it or to replace it for something else?

janpio:
Usually it should be replaced with a `type`, but sometimes we also need to skip the validation it seems: https://github.com/fastlane/fastlane/pull/13816/files

minuscorp:
I have a working regex... https://regex101.com/r/SXIbmK/1 that finds the group, later on we can add the logic

minuscorp:
It seems easy to replace whose the value is `true`, even though automatically we don't know if those are dynamic values... For the `false` vaulues, is a lottery

joshdholtz:
Yeah yeah, this isn’t just simple replace 🤷‍♂️ I’ve been enabling this tools when I have large changes in them. I got a current branch that adds `sigh` into this. I was a little scared to do them all at once since its such a large change. I think it might be good to split this up for individual tools/directories for easier reviews and easier reverting (if we need to incase of a regression).

Thoughts?

janpio:
We can totally do that, this PR currently just tracks and collects the error output for the complete codebase.

minuscorp:
Actually, we could make a small PR for each `Action` and see if that produces regressions, step by step. Thoughts? Anyone to make the list and assign pieces for each one?

Issue #16532, opened 2020-05-29
Match nuke deletes wrong files when working with multiple teams


castus:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Match nuke deletes wrong files when working with multiple teams. I'm experiencing it using S3 storage. When I have multiple teams there, when nuking particular team it should nuke only that folder. Unfortunately it nukes completely different folders.
I found, that the path to files (certs, keys and provisiones) doesn't have team_id in it, so `Dir` will match always the very last files on the list. That's why it nukes completely different files.

I think everything begins here and this piece of code:
https://github.com/fastlane/fastlane/blob/efd68a67a1dab647257983e965cbde19a0500c7c/match/lib/match/nuke.rb#L140-L153

should be

```ruby
      # Finds all the .cer and .p12 files in the file storage
      certs = []
      keys = []
      cert_types.each do |ct|
        certs += Dir[File.join(self.storage.working_directory, params[:team_id].to_s,  "**", ct.to_s, "*.cer")]
        keys += Dir[File.join(self.storage.working_directory, params[:team_id].to_s, "**", ct.to_s, "*.p12")]
      end

      # Finds all the iOS and macOS profofiles in the file storage
      profiles = []
      prov_types.each do |prov_type|
        profiles += Dir[File.join(self.storage.working_directory, params[:team_id].to_s, "**", prov_type.to_s, "*.mobileprovision")]
        profiles += Dir[File.join(self.storage.working_directory, params[:team_id].to_s, "**", prov_type.to_s, "*.provisionprofile")]
      end

```

##### Command executed
`bundle exec fastlane match nuke distribution --team_id 'XXX' --username me@me.com --storage_mode 's3' --s3_region 'eu-central-1' --s3_access_key XXX --s3_secret_access_key XXX --s3_bucket 'bucket' --skip_confirmation true --skip_docs true --keychain_name login.keychain`

<details>
  <pre><details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.15.4                                     |
[...]

PR #16541, opened 2020-05-31
Fix problem when match nuke deletes to much files when working with ultiple teams on S3

castus:
### Checklist

### Motivation and Context
It fixes a bug https://github.com/fastlane/fastlane/issues/16532

### Description
According to documentation S3 and Google Cloud storage saves the files starting with `team_id`. Git storage uses branches to distinguish between teams. The fix was to correct the way each storage list files to delete. I've tested it on S3 and Git using multiple teams and it works.

### Testing Steps
* To test it you need to have an account with multiple teams.
* Then you need to use `match` to generate profiles and certificates. You don't need to create each type, one type is enough to test the fix (for ex. development).
* You need to `match` for Git and S3 (or Google Cloud)
* Then you need to `nuke` it. Previously when you nuke S3 account, it will delete all files no matter what team you give. 
* Expected behaviour is to nuke only files in the team you provided. This pull request fixes it.

### DIFF
--- a/match/lib/match/nuke.rb
+++ b/match/lib/match/nuke.rb
@@ -41,7 +41,8 @@ module Match
         s3_region: params[:s3_region].to_s,
         s3_access_key: params[:s3_access_key].to_s,
         s3_secret_access_key: params[:s3_secret_access_key].to_s,
-        s3_bucket: params[:s3_bucket].to_s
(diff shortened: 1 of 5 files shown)


castus:
@googlebot I fixed it.

castus:
@googlebot I fixed it.

PR #16542, opened 2020-05-31
[match] delete the correct files when working with multiple teams on S3

castus:
### Checklist

### Motivation and Context
It fixes the bug https://github.com/fastlane/fastlane/issues/16532

### Description
According to documentation S3 and Google Cloud storage saves the files starting with `team_id`. Git storage uses branches to distinguish between teams. The fix was to correct the way each storage list files to delete. I've tested it on S3 and Git using multiple teams and it works.

### Testing Steps
* To test it you need to have an account with multiple teams.
* Then you need to use `match` to generate profiles and certificates. You don't need to create each type, one type is enough to test the fix (for ex. development).
* You need to `match` for Git and S3 (or Google Cloud)
* Then you need to `nuke` it. Previously when you nuke S3 account, it will delete all files no matter what team you give. 
* Expected behaviour is to nuke only files in the team you provided. This pull request fixes it.

### DIFF
--- a/match/lib/match/nuke.rb
+++ b/match/lib/match/nuke.rb
@@ -27,6 +27,8 @@ module Match
 
       update_optional_values_depending_on_storage_type(params)
 
+      spaceship_login
+
       self.storage = Storage.for_mode(params[:storage_mode], {
         git_url: params[:git_url],
         shallow_clone: params[:shallow_clone],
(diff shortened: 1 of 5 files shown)


joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for making this fix! Great catch 😊 I added one commit to get the `team_id` from the logged in Spaceship session but other than that this was 💯

castus:
Thanks for accepting @joshdholtz It was a real fun contributing to fastlane, I should definitely do it more often ;) Could you tell me when you would like to release that fix?

joshdholtz:
@castus We'd be happy to have more contributions! And its release now 😊 Thanks again! ❤️

PR #16609, opened 2020-06-15
[gym] iOS Export Options "Destination" Handling

DomenicBianchi01:
### Checklist

### Motivation and Context
`gym` supports passing in archive export options through the `export_options` parameter. When `gym` runs `fastlane/gym/lib/assets/wrap_xcodebuild/xcbuild-safe.sh -exportArchive ...`, it expects that a `.ipa` file (amongst other files) will be created afterwards. 

However, if within `export_options` you specify `destination: "upload"`, no external files are generated (including the `.ipa`) since `destination: "upload` tells `xcodebuild` to upload to iTunes Connect immediately.

As a result, if you specify `destination: "upload"`, the archive will successfully upload to iTunes Connect (assuming you have configured everything else correctly), but you will get the following error:

```
ERROR [2020-06-15 09:29:18.37]: The generated archive is invalid, this can have various reasons:
ERROR [2020-06-15 09:29:18.37]: Usually it's caused by the `Skip Install` option in Xcode, set it to `NO`
ERROR [2020-06-15 09:29:18.37]: For more information visit https://developer.apple.com/library/ios/technotes/tn2215/_index.html
ERROR [2020-06-15 09:29:18.37]: Also, make sure to have a valid code signing identity and provisioning profile installed
ERROR [2020-06-15 09:29:18.37]: Follow this guide to setup code signing https://docs.fastlane.tools/codesigning/GettingStarted/
ERROR [2020-06-15 09:29:18.37]: If your intention was only to export an ipa be sure to provide a valid archive at the archive path.
ERROR [2020-06-15 09:29:18.37]: This error might also happen if your workspace/project file is not in the root directory of your project.
ERROR [2020-06-15 09:29:18.37]: To workaround that issue, you can wrap your calls to gym with
ERROR [2020-06-15 09:29:18.37]: `Dir.chdir('../path/to/dir/containing/proj') do`
ERROR [2020-06-15 09:29:18.37]: For an example you can check out
ERROR [2020-06-15 09:29:18.37]: https://github.com/artsy/emission-nebula/commit/44fe51a7fea8f7d52f0f77d6c3084827fe5dd59e
```

The purpose of this PR is to prevent Fastlane from displaying this misleading error.

### Description

The changes in this PR are fairly straight forward.

I simply check if the export options plist contains the `destination` key with a value of `upload`. If it does, the code no longer executes the "move" related functions (since there are no files to move in the first place).

### Testing Steps

Run `gym` using the following export options (depending on how your specific Xcode project is configured, you might need to include other options):

```
{
    "destination": "upload",
    "method": "app-store",
    "signingStyle": "automatic"
}
```

Also, specify `export_team_id` (an existing gym parameter)


DomenicBianchi01:
@googlebot I signed it!

DomenicBianchi01:
@googlebot I fixed it.

joshdholtz:
@DomenicBianchi01 I learned so much from this! 🤯 Hopefully I learned the right thing though lol... So if I’m understanding this correctly, you use `gym` to upload to App Store Connect by setting `destination` as `upload`? I had absolutely no idea this was a thing! 

This code change totally makes sense and looks good from a review standpoint. I’m going to try to get something setup locally to test this because I really want to see this working on my own machine too 😇

joshdholtz:
Oh my goodness I did it! I uploaded an IPA through `gym`. This is 🔥

DomenicBianchi01:
> @DomenicBianchi01 I learned so much from this! 🤯 Hopefully I learned the right thing though lol... So if I’m understanding this correctly, you use `gym` to upload to App Store Connect by setting `destination` as `upload`? I had absolutely no idea this was a thing!
> 
> This code change totally makes sense and looks good from a review standpoint. I’m going to try to get something setup locally to test this because I really want to see this working on my own machine too 😇

Correct! At the company I work at, we have multiple targets, each which need to be signed using a different team id/provisioning profiles/certificates/etc.

By setting the `destination` field, Xcode is actually smart enough to figure out exactly which iTMSTransporter team id needs to be used (sorry I don't know the actual name for this identifier, but it is NOT the team id shown in within the Signing & Configurations menu in Xcode) for a specific export and is therefore able to upload to App Store Connect in one shot.

Without these changes, I need to manually provide the iTMSTransporter team id for `upload_to_app_store`. Long story short, because we manage many different targets with different team ids, having to provide this extra ID is not ideal since setting the `destination` field means I don't need to worry about the ID at all.

joshdholtz:
@googlebot I consent.

joshdholtz:
@DomenicBianchi01 Thanks again for adding support for this! I pushed a commit on to this to also add this for mac as well. I will get this merged in tomorrow morning and out in the next release 💪

joshdholtz (approved):
This is 🔥 Thanks so much for making this change! I’m going to be adding this to some of my projects 😊 Really appreciate the contribution!

Issue #16475, opened 2020-05-19
[match] Fails to checkout master branch if it is not the default branch of the repository


mbogh:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I changed the default branch of my certificates repository to something different than `master` and now **match** fails to sync any certificate and/or profile on the `master` branch.

If ` unless self.branch == "master"` is removed from [git_storage.rb](https://github.com/fastlane/fastlane/blob/master/match/lib/match/storage/git_storage.rb#L116) it works on both repositories with `master` as a default branch and not.

So the solution could either be to always checkout the specified branch or check the working directory if there is a branch mismatch.

##### Command executed

```
sync_code_signing(
      git_url: "git@github.com:ORG/certificates.git",
      git_branch: 'master',
      app_identifier: ['com.abc.xyz'],
      type: 'appstore',
      team_id: "XXX",
      readonly: true
    )
```

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> 
[✔] 🚀 
[08:52:13]: ------------------------------
[08:52:13]: --- Step: default_platform ---
[08:52:13]: ------------------------------
[08:52:13]: Driving the lane 'ios app' 🚀
[08:52:13]: -------------------------------
[08:52:13]: --- Step: sync_code_signing ---
[08:52:13]: -------------------------------

+----------------------------+--------------------------------------------+
|                        Summary for match 2.148.1                        |
+----------------------------+--------------------------------------------+
| git_url                    | git@github.com:ORG/certificates.git        |
| git_branch                 | master                                     |
| app_identifier             | ["com.abc.xyz"]                            |
| type                       | appstore                                   |
| team_id                    | xxx                                        |
| readonly                   | true                                       |
| git_private_key            | ********                                   |
| generate_apple_certs       | true                                       |
| skip_provisioning_profiles | false                                      |
| storage_mode               | git                                        |
| shallow_clone              | false                                      |
| clone_branch_directly      | false                                      |
| keychain_name              | login.keychain                             |
| force                      | false                                      |
| force_for_new_devices      | false                                      |
| skip_confirmation          | false                                      |
[...]

PR #16622, opened 2020-06-17
[match] Always checkout the specified branch as master may not be the default branch

mbogh:
… branch.

### Checklist

### Motivation and Context

As described in #16475 `master` is not guaranteed to be the default branch and then `match` fails to work.

### Description

I have changed the `git_storage.rb` such that the branch specified is always checked out.


joshdholtz (approved):
Thanks for fixing this! Can confirmed this work for me on a master and non-master branch. I appreciate the contribution ❤️

Issue #16631, opened 2020-06-19
Failure information is limited in socket server


rhdeck:
### Feature Request

The socket server presently only remits a `failure_information` array with stack traces organized by cause. However, human-usable information, such as that raised by `FastlaneError` are not remitted or displayed on stdout or stderr, making it hard to help a user diagnose their problem. 

#### Feature Description
Basically, we should at a minimum add the `class` and `message` method results of the `StandardError` that is the basis for sending the `failure` payload in the socket server. These two in combination would allow a client to make the kinds of decisions that the `UI` class makes for remitting information to the user. 

The only canonical socket interface for Fastlane is Swift, which uses `payload` as a dictionary. As long as we _extend_ the shape rather than changing already-used keys, this will not break anything. 

#### Alternatives or Workarounds
Since this is a matter of information not coming over the transom, there is no workaround - can't hack what you don't have. 

I am putting together a PR to solve this issue.

Siler177:
On the TCP level the tuple (source ip, source port, destination ip, destination port) must be unique for each simultaneous connection. That means a single client cannot open more than 65535 simultaneous connections to a server. But a server can (theoretically) serve 65535 simultaneous connections per client.

PR #16632, opened 2020-06-19
[fastlane] Extend socket failure payload

rhdeck:
### Checklist
- [x] I've updated the documentation if necessary. *(Socket server is mostly undocumented)*

### Motivation and Context
Fixing issue #16631.

When accessing fastlane over the socket server, only a little error information comes back, and user-based error messages (e.g. like the intended UX of `FastlaneError`) do not get remitted at all. The `failureinformation` array does not include the message, and is frankly just a mess of backtraces that have limited utility unless one is diving back through the ruby code. 

### Solution: 

Extend the failure payload include the class as `failure_class` and message as`failure_message` of the raised error in the payload to complement the backtrace in `failure_information`. The socket client can then decide how to better inform the user of what happened. 

### Description
Edited socket.server.rb to add the returns of `message` and `class` methods on the error object to the payload. Since both are part of `StandardError` this should work every time. 

I ran this with a customized Gemfile to confirm it generates the results that help the user. 

Overall this is a simple (four lines of new code, one line with an added comma) change that makes the socket server much more useful!


minuscorp:
This should be in sync with the Swift client in order to read about that issues and raises them to the user somehow.

rhdeck:
Id say go for it - make a PR to extend the swift client to be smarter about Fastlane error types and messages. For the purposes of this PR:

1) swift isn’t my use case, so client side extension was not in my scope
2) this PR enables such an extension by exposing the data on the socket server, and 3) because it extends the failure payload shape rather than changing a key, the PR doesn’t prevent the swift client from either using what it had before or using this new info if a client is so inclined.

So no downside to the PR and a potential layup for someone who wants to update the swift extension with intelligence for responding to FL error types and messages.

minuscorp:
I don't see the point of integrating this on the Ruby side and expect to someone else to fix the Swift part 🤔 I tend to think that this PR is incomplete.

rhdeck:
My use case is a OSS TS/JS-based client for Fastlane I am writing: https://github.com/rhdeck/fastlane-js. I am using the socket server to enable the cross-language support, much as FS-swift does. I submitted a PR for setting the port which was merged without incident. I appreciate the client isn't part of FS core, but it's really useful - especially for React Native. 

I made sure that the PR would address the problem without causing more issues in swift-land. I am not a user of the swift client and don't feel qualified to submit a PR refactoring its error handling - especially to get four lines of code. 

Maybe the issue is that I see the socket server and the swift client as separate concerns, where we can improve one and get value from it without having to improve the other? It's my hope that we can be expansive about functionality and do what's easy to add value to more people. 

FWIW it was not I who added the swift label l to this or to issue #16631, which related to the ruby socket _server_ not the swift _client_.

Anyhow, it would help my consumers' experience a lot to get the messages, especially for FastlaneError situations where the message is key for the user to make a smart decision. 

Let me know if me doing something additional would help.

rhdeck:
I took a look at the code in the swift client, and the crux of error handling is in SocketResponse and SocketClient, which basically dumps failure data to a log and shuts down the socket. (This use case explains the lack of data from the server - its not used for any flow control except hard stop.) 

That said, we could surface the information to that log with the following inserted, starting on line 41 in `SocketResponse.swift`:
```swift
            } else if status == "failure" {
                guard var failureInformation = statusDictionary["failure_information"] as? [String]  else {
                    self = .parseFailure(failureInformation: ["Ruby server indicated failure but Swift couldn't receive it"])
                    return
                }
                if let failureClass = statusDictionary["failure_class"] failureInformation.append(failureClass)
                if let failureMessage = statusDictionary["failure_message"] failureInformation.append(failureMessage)
                self = .failure(failureInformation: failureInformation)
                return
            }
```
I'm glad to commit this change so that the client can profit from the additional data, but I'd appreciate the read of whether this would be a helpful addition, since it's not my use case. 

How can I be most helpful?

janpio:
Hey @rhdeck, sorry I misunderstood what you were doing there - for me the socket server was/is directly connected to Swift - hence the label. 

But your project looks super interesting! Any docs on how (you will be able) to use it in JS?

janpio (approved):
As this PR only surfaces additional data, I think this is safe to merge as is.

If someone wants, please feel free to open a PR against the Swift client of this to utilize this new power and use the error class and message to improve any error messages in case of error.

minuscorp (approved):
LGTM 🚀

joshdholtz (approved):
This looks good to me! And https://github.com/rhdeck/fastlane-js looks dope 😊 Excited to try it out!

rhdeck:
Working on it! This error payload will help a lot. 

Your point about docs was well-taken...

PR #16636, opened 2020-06-20
[fastlane_core] fix: paths with spaces fail

rigor789:
### Checklist

### Motivation and Context

If the path has a space in it (for example `C:/Users/User Name/AppData/Local/Temp/....`), the command would ignore anything after the space, thus exiting with
```Can't open C:/Users/User for reading, No such file or directory```

### Description

The only change is wrapping the path in quotes.

### Testing Steps

I personally didn't test - because it requires creating a new user on Windows that has a home directory with a space in it, however this came up on the NativeScript discord by someone having trouble with it (screenshot submitted by a community member)

image

### DIFF
--- a/fastlane_core/lib/fastlane_core/provisioning_profile.rb
+++ b/fastlane_core/lib/fastlane_core/provisioning_profile.rb
@@ -113,7 +113,7 @@ module FastlaneCore
           else
             # `security` only works on Mac, fallback to `openssl`
             # via https://stackoverflow.com/a/14379814/252627
-            decoded = `openssl smime -inform der -verify -noverify -in "#{path}" 2> #{err}`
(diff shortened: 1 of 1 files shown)


rigor789:
@googlebot I signed it!

janpio (changes requested):
The correct solution is to shellescape the parameter - searching for `shellescape` in the codebase should give some examples on how to use this.

janpio (approved):
Awesome, thank you @rigor789!

rigor789:
No worries - I wasn't sure if just the quotes would be enough, but being safe with `shellescape` is definitely better!

janpio:
If I remember correctly it depends on the OS - some like `"`, some one `'` etc - which is why we have a modified version of `shellescape` that does the right thing on Windows.

PR #16640, opened 2020-06-21
[deliver][produce][spaceship] Update to use new App Store Connect API endpoints

joshdholtz:
Fixes #16639
Fixes https://github.com/fastlane/fastlane/issues/16649
Fixes https://github.com/fastlane/fastlane/issues/16653

## Fixes
- [x] deliver completely broken
  - Issue: #16621 
  - PR: #16626
- [x] deliver build processing shouldn't require app version
  - Issue: https://github.com/fastlane/fastlane/pull/16640#discussion_r443233736
  - Issue: https://github.com/fastlane/fastlane/issues/16663
  - Issue: https://github.com/fastlane/fastlane/issues/16648
  - PR: #16713
- [x] Completely new app gets: You must provide a value for the attribute 'usesNonExemptEncryption' with this request 
  - Issue: #16645
  - PR: #16646
- [x] `usesIdfa` - is there a way around?
  - Issue: https://github.com/fastlane/fastlane/pull/16626#issuecomment-647214671
  - Fixes: https://github.com/fastlane/fastlane/pull/16706
- [x] make sure primary category isn't getting cleared out if not being set
  - Issue: https://github.com/fastlane/fastlane/issues/16621#issuecomment-647143129
  - PR: https://github.com/fastlane/fastlane/pull/16652
- [x] produce broken
  - Issue: https://github.com/fastlane/fastlane/issues/16649
- [x] You must provide a value for the attribute 'honorsLimitedAdTracking'
  - Issue: https://github.com/fastlane/fastlane/issues/16621#issuecomment-647241384
  - PR: #16646
- [x] 504 gateway timeout on uploading screenshot (extend timeout???)
  - Issue: https://github.com/fastlane/fastlane/pull/16626#issuecomment-647127031
  - PR: #16667
  - PR: #16676
- [x] produce map previous languages
  - Issue: https://github.com/fastlane/fastlane/pull/16626#issuecomment-647378487
  - PR: #16656
- [x] add `:default` to languages for deliver
  - Issue: https://github.com/fastlane/fastlane/issues/16639#issuecomment-647366971
  - PR: #16657
- [x] category mapping validation (tell user thing was mapped and show link to new doc)
  - Issue: NONE
  - PR: #16652 
- [x] app rating (tell user thing was mapped and show link to new doc)
  - Issue: NONE
  - PR: #16655
- [x] fix `set_changelog`
  - Issue: https://github.com/fastlane/fastlane/issues/16639#issuecomment-647661174
  - PR: #16658
[...]

PR #16656, opened 2020-06-23
[deliver][produce] map language names and available countries on app

joshdholtz:
### Motivation and Context
Fixes https://github.com/fastlane/fastlane/pull/16626#issuecomment-647378487
Fixes https://github.com/fastlane/fastlane/pull/16640#issuecomment-648155512

### Description
- produce language was expecting language name like "English" but now it wants "en-US". Added mapping for backwards compatibility
- uploading pricing in deliver will add all countries if its the first price


nzaghini-phorest (commented):
Tested in real life and works perfectly defaulting price to 0 and opening up for all regions.

Didn't test the language name mapping

PR #16664, opened 2020-06-24
[fastlane] Explain multi_xml dependency

aried3r:
### Checklist

### Motivation and Context

`faraday_middleware` is being used to parse XML responses in the spaceship client. Indirectly, this requires `multi_xml` as a dependency.

https://github.com/lostisland/faraday_middleware#dependencies
https://github.com/fastlane/fastlane/blob/2.150.0.rc1/spaceship/lib/spaceship/client.rb#L208

### Description

Explain `multi_xml` dependency

### Testing Steps

Ran the tests and they ran through. Let's hope the same is true on CI.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -83,7 +83,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('highline', '>= 1.7.2', '< 2.0.0') # user inputs (e.g. passwords)
   spec.add_dependency('json', '< 3.0.0') # Because sometimes it's just not installed
   spec.add_dependency('mini_magick', '>= 4.9.4', '< 5.0.0') # To open, edit and export PSD files
-  spec.add_dependency('multi_xml', '~> 0.5')
+  spec.add_dependency('multi_xml', '~> 0.5') # Used indirectly by spaceship client via faraday_middleware
   spec.add_dependency('rubyzip', '>= 1.3.0', '< 2.0.0') # fix swift/ipa in gym
   spec.add_dependency('security', '= 0.1.3') # macOS Keychain manager, a dead project, no updates expected
   spec.add_dependency('xcpretty-travis-formatter', '>= 0.0.3')


aried3r:
Aaah, guess I was wrong about this one. I'll check why I missed this locally and leave a comment in the `.gemspec` file instead.

janpio:
Why is this comment important? Is this the only location where this is used? If it is indirectly, can we possibly get rid of it? (Our gemspec is not super groomed...)

aried3r:
Hey! Thanks for the questions!

> Why is this comment important?

So, while working on the other PRs I was going through dependencies and `bundle outdated` and found `multi_xml` with not comment explaining what it does. There's also no direct usage in this repo. The first commit it appears in was when fastlane was converted into a monorepo, so I tried removing it and thought that worked. However I was looking at the wrong test-run and when opening this PR I saw tests failed.

So, in short, because this dependency has no direct usage and this dependency block is sorted alphabetically, it doesn't appear directly under the faraday dependencies, I thought about adding a comment (like almost all other dependencies have).

> Is this the only location where this is used? If it is indirectly, can we possibly get rid of it? 

Long story short, `multi_xml` is an optional dependency, or rather, not so optional when parsing XML responses. [See](https://github.com/lostisland/faraday_middleware/blob/v1.0.0/lib/faraday_middleware/response/parse_xml.rb) the [faraday_middleware](https://github.com/lostisland/faraday_middleware/tree/v1.0.0#as-of-v0160-faraday-and-faraday_middleware-no-longer-officially-support-jruby-or-rubinius) repo.

In fastlane, I believe it's used at least in two places:

https://github.com/fastlane/fastlane/blob/2.150.0.rc4/spaceship/lib/spaceship/connect_api/client.rb#L31
https://github.com/fastlane/fastlane/blob/2.150.0.rc4/spaceship/lib/spaceship/client.rb#L209

I have no knowledge about what the API returns and if there are indeed XML responses. If not, we can potentially remove it and thus `multi_xml`.

janpio (approved):

joshdholtz:
Closing due to #16697 ❤️

PR #16669, opened 2020-06-24
[sigh] Resign OnDemandResources

### LINKED ISSUE #16668: Resign should also resign OnDemandResources

We have team's generate IPA's with their development profiles/certs, and then we resign centrally.  We are trying to use ODR's though and when we try to resign their IPA via *sigh*, the ODR's are not signed/resigned.

#### Feature Description

If an IPA is resigned and the IPA contains an OnDemandResources folder within the Payload folder, resign the assetpack within the ODR as well.

At the moment, we have to write a plugin that simulates much of what resign.sh does ( unzip, get the keychain, sign, etc ).

sjmh:
Allows sigh to resign OnDemandResource assetpacks

fixes #16668

### Checklist

### Motivation and Context

We have teams that create/sign IPAs with development provisioning profiles and then we centrally sign with fastlane before deploying to the store.  They would like to start using ODR's, however, when we resign, these ODR's are not being (re)signed.

### Description

Detects if an ODR folder is within the Payload folder, if so, removes any codesignature's from each pack and resigns each pack.

### Testing Steps

Modified a local installation of fastlane with these changes and ran an IPA with ODR through it.  IPA came out successfully signed.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -69,6 +69,9 @@
 # new features March 2019
 # 1. two more fixes for only creating the archived-expanded-entitlements.xcent file if the version of Xcode < 9.3 as Xcode 10 does not create it.
 #
+# new features June 2020
+# 1. enable (re)signing of OnDemandResources when ipa has been built for the appstore
+#
 
 # Logging functions
 
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -528,6 +528,21 @@ function resign {
         PlistBuddy -c "Set :CFBundleVersion $BUNDLE_VERSION" "$APP_PATH/Info.plist"
     fi
 
+    # Check for and resign OnDemandResource folders
+    ODR_DIR="$(dirname $APP_PATH)/OnDemandResources"
+    if [ -d "$ODR_DIR" ]; then
+        for assetpack in "$ODR_DIR"/*
+        do
+            if [[ "$assetpack" == *.assetpack ]]; then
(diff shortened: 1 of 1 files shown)


sjmh:
@googlebot I signed it!

janpio (approved):
Looks safe enough to approve without actually understanding what exactly it does ;)

sjmh:
Thanks @janpio - I probably should have explained better.

Apple allows you to push certain assets to OnDemandResources ( https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/On_Demand_Resources_Guide/index.html ).

Normally during your dev cycles, you can either host these extra resources on your own local web servers, or embed them directly in the app.  When you embed, it puts them in the Payload/<app name>/OnDemandResources, but that's really only for debug purposes..  In those instances though, fastlane signs them because it finds them in the app directory.

However, when you deploy to the AppStore, the AppStore must host the resources, but when you build the IPA for the AppStore, it actually creates the OnDemandResources folder in Payload, rather than the Payload/<app name>.   *sigh* currently doesn't detect that however, so the resources don't get signed/resigned.

joshdholtz (changes requested):
@sjmh Thanks for adding this! Can you add a changelog to the top of the file? There should be a format you can follow up there 😇

joshdholtz (changes requested):
@sjmh Thanks for adding this! Can you add a changelog to the top of the file? There should be a format you can follow up there 😇

sjmh:
Oops, of course!

joshdholtz (approved):
This seems good to me! Thanks for adding this ❤️

PR #16682, opened 2020-06-25
[match] Add `s3_object_prefix` option to match's S3Storage.

atreat:
### Checklist

### Motivation and Context

Wanted to use the Match S3 storage option but realized that it would take over the bucket I had planned to use. With a prefix, I'd be able to keep android/ios codesign assets in the same bucket but partitioned off correctly.

### Description

This adds the `s3_object_prefix` (`MATCH_S3_OBJECT_PREFIX` envvar) as an available option. This can be used to help partition match in an existing bucket.

### Testing Steps

I've added RSpec examples for the areas that were updated. Also updated what seemed to be an odd stub mechanism in the runner spec as I did not expect tests to fail because this new optional option was not present.

I also switched over to use this updated option locally and was able to import my existing certificates/profiles into the expected s3 structure.

### DIFF
--- a/match/lib/match/storage/s3_storage.rb
+++ b/match/lib/match/storage/s3_storage.rb
@@ -167,7 +167,7 @@ module Match
       end
 
       def strip_s3_object_prefix(object_path)
-        object_path.gsub(s3_object_prefix, "")
+        object_path.gsub(/^#{s3_object_prefix}/, "")
       end
 
       def sanitize_file_name(file_name)
--- a/match/lib/match/storage/s3_storage.rb
+++ b/match/lib/match/storage/s3_storage.rb
@@ -14,6 +14,7 @@ module Match
       attr_reader :s3_bucket
       attr_reader :s3_region
       attr_reader :s3_client
+      attr_reader :s3_object_prefix
       attr_reader :readonly
       attr_reader :username
       attr_reader :team_id
@@ -24,6 +25,7 @@ module Match
         s3_access_key = params[:s3_access_key]
         s3_secret_access_key = params[:s3_secret_access_key]
         s3_bucket = params[:s3_bucket]
+        s3_object_prefix = params[:s3_object_prefix]
 
         if params[:git_url].to_s.length > 0
           UI.important("Looks like you still define a `git_url` somewhere, even though")
@@ -37,6 +39,7 @@ module Match
           s3_access_key: s3_access_key,
           s3_secret_access_key: s3_secret_access_key,
           s3_bucket: s3_bucket,
+          s3_object_prefix: s3_object_prefix,
           readonly: params[:readonly],
           username: params[:username],
           team_id: params[:team_id],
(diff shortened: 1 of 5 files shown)


atreat:
Not sure how best to address the failing build
```
```

The step beforehand prints out:
```
ruby -v
ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-darwin17]
```

If anyone has any thoughts on the best path forward let me know.

janpio:
I usually just restart the tests - this error message does not make any sense. I fear it is some random machine on Circle CI that just lies to us for some reason. IF you have an idea how to fix it, pleeeeease let me know.

janpio:
Uh hey @googlebot, how about the CLA label thing!?

match/lib/match/storage/s3_storage.rb
```diff
@@ -152,6 +159,17 @@ def generate_matchfile_content(template: nil)
...
 
+      def s3_object_path(file_name)
+        santized = sanitize_file_name(file_name)
+        return santized if santized.start_with?(s3_object_prefix)
+
+        s3_object_prefix + santized
+      end
+
+      def strip_s3_object_prefix(object_path)
+        object_path.gsub(s3_object_prefix, "")
```
atreat, line 170: This should be anchored to the front of the object_path just in case the s3_object_prefix is found multiple times in the path. Want to prevent this behavior
```
prefix = hello
object_path = 'hello_there_hello'
stripped = '_there_'
```

```suggestion
        object_path.gsub(/^#{s3_object_prefix}/, "")
```

match/spec/runner_spec.rb
```diff
@@ -38,31 +38,7 @@
...
-            s3_region: "",
-            s3_access_key: "",
-            s3_secret_access_key: "",
-            s3_bucket: "",
-            readonly: false,
-            username: values[:username],
-            team_id: nil,
-            team_name: nil
-          ).and_return(fake_storage)
+          expect(Match::Storage).to receive(:for_mode).and_return(fake_storage)
```
joshdholtz, line 41: Can we ad the `.with(` back in? Its actually caught a few issues in the past. 😇

match/spec/runner_spec.rb
```diff
@@ -119,31 +95,7 @@
...
```
joshdholtz, line 98: Same here ☝️

match/spec/runner_spec.rb
```diff
@@ -199,31 +151,7 @@
...
```
joshdholtz, line 154: Same here ☝️

match/spec/runner_spec.rb
```diff
@@ -265,31 +193,7 @@
...
```
joshdholtz, line 196: Same here ☝️

joshdholtz (changes requested):
This looks great! I like this addition. Just a few small thoughts on adding back in some checking on the tests 😊

joshdholtz (approved):
Just tested this out and it works great for me! Thanks for adding this in! ❤️

PR #16687, opened 2020-06-26
[fastlane] Fix S3ClientHelper side effects

atreat:
### Checklist

### Motivation and Context

When testing the changes in https://github.com/fastlane/fastlane/pull/16682 I found that my authentication was not working because the S3ClientHelper was overwriting the Aws config I expected. This change will remove that side-effect and allow the caller to dictate what AWS credentials should be used.

### Description

Updates the S3ClientHelper to remove the access_key/secret initializer parameters. Removes the call to `Aws.config.update`. Instead, passes the remaining `region` parameter to the S3 client when initializing that.

Also cleans up some confusing code in the file. The `client` method was defined as private outside of the S3ClientHelper class. The S3ClientHelper then had a public `attr_reader :client` that allowed the underlying client to leak out. This has been updated such that there is a private `client` method that is not accessible outside the helper class as I believe the intention is to always use the interface defined by the helper.

### Testing Steps

Updated the necessary call sites in match's S3Storage and specs. Specs for the S3ClientHelper continue to pass.

### DIFF
--- a/fastlane/lib/fastlane/helper/s3_client_helper.rb
+++ b/fastlane/lib/fastlane/helper/s3_client_helper.rb
@@ -3,7 +3,6 @@ require 'aws-sdk-s3'
 module Fastlane
   module Helper
     class S3ClientHelper
-
       attr_reader :region
 
       def initialize(region: ENV['AWS_REGION'])
@@ -53,7 +52,6 @@ module Fastlane
       def client
         @client ||= Aws::S3::Client.new(region: region)
       end
-
     end
   end
 end
--- a/fastlane/lib/fastlane/helper/s3_client_helper.rb
+++ b/fastlane/lib/fastlane/helper/s3_client_helper.rb
@@ -5,8 +5,9 @@ module Fastlane
     class S3ClientHelper
       attr_reader :region
 
-      def initialize(region: ENV['AWS_REGION'])
+      def initialize(region: ENV['AWS_REGION'], s3_client: nil)
         @region = region
+        @client = s3_client
       end
 
       def list_buckets
--- a/fastlane/lib/fastlane/helper/s3_client_helper.rb
+++ b/fastlane/lib/fastlane/helper/s3_client_helper.rb
@@ -54,6 +54,8 @@ module Fastlane
 
       private
 
+      attr_reader :secret_access_key
+
       def client
         @client ||= Aws::S3::Client.new(
           {
@@ -64,11 +66,11 @@ module Fastlane
       end
 
       def create_credentials
-        return nil if access_key.nil? || @secret_access_key.nil?
+        return nil if access_key.nil? || secret_access_key.nil?
 
         Aws::Credentials.new(
           access_key,
-          @secret_access_key
+          secret_access_key
         )
       end
     end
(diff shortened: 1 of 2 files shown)


janpio:
Could this be a breaking change to anyone or is this deep enough in the code that we do not think anyone uses this directly?

atreat:
I couldn't find any other callers of the S3ClientHelper. It is possible that people call this directly though. I can add back in the original initializer and print out a deprecation notice if that is best.

atreat:
Realized I could pass the access_key/secret to the underlying S3 client if they exist. This allows for the existing interface to be used and removes the need to call `Aws.config`

atreat:
@janpio let me know if you have any feedback on the updates

fastlane/lib/fastlane/helper/s3_client_helper.rb
```diff
@@ -50,12 +51,26 @@ def find_bucket!(bucket_name)
...
+        @client ||= Aws::S3::Client.new(
+          {
+            region: region,
+            credentials: create_credentials
+          }.compact
+        )
+      end
+
+      def create_credentials
+        return nil if access_key.nil? || @secret_access_key.nil?
```
joshdholtz, line 67: Is there reason why `secret_access_key` has the `@` in front of it and `access_key` doesn't? 🤔 Just asking so we can have consistency or a comment mentioning why 😇

fastlane/lib/fastlane/helper/s3_client_helper.rb
```diff
@@ -50,12 +51,26 @@ def find_bucket!(bucket_name)
...
+
+        Aws::Credentials.new(
+          access_key,
+          @secret_access_key
```
joshdholtz, line 71: Same here ☝️

match/lib/match/storage/s3_storage.rb
```diff
@@ -45,16 +41,14 @@ def self.configure(params)
...
-                     s3_access_key: nil,
-                     s3_secret_access_key: nil,
                      s3_bucket: nil,
                      readonly: nil,
                      username: nil,
                      team_id: nil,
                      team_name: nil)
         @s3_bucket = s3_bucket
         @s3_region = s3_region
-        @s3_client = Fastlane::Helper::S3ClientHelper.new(access_key: s3_access_key, secret_access_key: s3_secret_access_key, region: s3_region)
```
joshdholtz, line 57: What is the reason for removing the access_key and secret_key from here? Isn't that needed for authenticating? I _think_ that access key and secret is needed for match to authenticate the user Into S3?

joshdholtz (changes requested):
I like this change! Just a few small changes requested / questions 😊

fastlane/lib/fastlane/helper/s3_client_helper.rb
atreat: This is a small attempt at keeping the secret_access_key from being accessed using an attr_reader. I can make that private or if its not a concern I can just open it up alongside the access_key.

match/lib/match/storage/s3_storage.rb
atreat: I think I just missed revisiting this after adding back in the access_key/secret. Will update.

fastlane/lib/fastlane/helper/s3_client_helper.rb
joshdholtz: I don't think its too much of a concern! I would go for consistency here if possible 😊

joshdholtz (approved):
Thanks for making those changes! This looks 🔥 Will get this released in tonight's release 💪

Issue #16677, opened 2020-06-25
[Fastlane.Swift] Runner hanging on failure 


jhandguy:
### New Regression Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Regression Information

- Breaking version: `2.150.0.rc1`
- Last working version: `2.149.0`

### Regression Description

This behaviour happens for any lane (scan, etc..) which do not have an `error_callback`, and sometimes also the lanes that have an `error_callback` but only for certain kind of errors (i.e. `sh` with a command not found error).

Code Samples:
```
class Fastfile: LaneFile {
    func testLane() {
        sh(command: "echoo") { error in
            sh(command: "echo OK")
        }
    }
}
```

```
class Fastfile: LaneFile {
    func testLane() {
        sh(command: "ls stuff")
    }
}
```

joshdholtz:
@minuscorp 👋 Hey! Do you mind taking a look at this whenever you get some time? 😊

jhandguy:
There seems to be 2 cases where the new logic fails:
1. Any lane without an `error_callback` specified will hang if it fails
2. For lanes that have an `error_callback` specified, it will still hang for certain kind of errors (i.e. command not found)

My guess is that the logic for handling failure is the faulty part, and in the case of an `error_callback` specified, the server socket won't actually return an error but a readyForNext. That would explain why it works sometimes for lanes with an `error_callback`.

minuscorp:
Yeah, the Swift part goes ahead with the execution if it receives a `ready_for_next` from the Ruby Socket, so the Ruby part for some reason is not getting the correct status code from the execution of each of this parts. As we left the Ruby part unchanged with the last PR, could you get a latter version of fastlane and test if this happens in the same way? (with no closures ofc). Until the weekend I'm quite out of time 😅

minuscorp:
Thinking of it, I'm quite sure why is happening, but I'll need more time to fix it 🗡️

jhandguy:
@minuscorp I was running it on the last version of Fastlane, I've put the breaking version as `2.149.0` because it was not happening before then, but my findings are based from `master` 👍 
[...]

PR #16691, opened 2020-06-27
[Fastlane.swift] Fix method hanging and propose an unified lint method

minuscorp:
### Checklist

### Motivation and Context

As raised in #16677 the new version of Fastlane.swift hang when certain error clauses converged at the same time between the Ruby and the Swift part. This PR aims to address it.

Also, I've seen that the Swift codebase doesn't follow any lint rules nor common syntax rules, so the first step is to integrate Swiftformat to ensure that every API or external change gets linted properly into the codebase. This not makes conflict to propose changes or cleanup code as proposed by another members.


fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -31,7 +31,7 @@ def initialize(action_name: nil, action_description: nil, action_details: nil, k
 
       # rubocop:disable LineLength
       # class instance?
-      @reserved_words = %w[associativity break case catch class continue convenience default deinit didSet do else enum extension fallthrough false final for func get guard if in infix init inout internal lazy let mutating nil operator override postfix precedence prefix private public repeat required return self set static struct subscript super switch throws true try var weak where while willSet].to_set
+      @reserved_words = %w[associativity break case catch class continue convenience default deinit didSet do else enum extension fallthrough false final for func guard if in infix init inout internal lazy let mutating nil operator override precedence private public repeat required return self static struct subscript super switch throws true try var weak where while willSet].to_set
```
janpio, line 34: What is this about?

fastlane/swift/ArgumentProcessor.swift
```diff
@@ -19,68 +19,68 @@ struct ArgumentProcessor {
...
         let fastlaneArgs = stride(from: 1, to: args.count - 1, by: 2).map {
-            RunnerArgument(name: args[$0], value: args[$0+1])
+            RunnerArgument(name: args[$0], value: args[$0 + 1])
         }
         self.args = fastlaneArgs
-        
+
         let fastlaneArgsMinusLanes = fastlaneArgs.filter { arg in
-            return arg.name.lowercased() != "lane"
+            arg.name.lowercased() != "lane"
```
janpio, line 31: no functional change here?
(I don't know any swift)

janpio (changes requested):
Can we possibly separate the fix and the link thigns into two PRs?

I could not really find the method fix as there was so much unrelated formatting and code style going on.
(Alternatively a link to the location of the actual change would be nice as well.)

fastlane/lib/fastlane/swift_fastlane_function.rb
minuscorp: This are not required reserved words in Swift 4.0. There was a time that they were in Swift 3 where you had to backtick them, but not in the current project version

fastlane/swift/ArgumentProcessor.swift
minuscorp: Let me double-check, but the return is unneeded.

fastlane/swift/SocketClient.swift
```diff
@@ -27,121 +27,121 @@ public enum SocketClientResponse: Error {
...
+
+        socketStatus = .ready
+        socketDelegate?.connectionsOpened()
     }
-    
+
     public func send(rubyCommand: RubyCommandable) {
         verbose(message: "sending: \(rubyCommand.json)")
         send(string: rubyCommand.json)
+        writeSemaphore.signal()
```
minuscorp, line 120: This is the line that was causing the hang

fastlane/swift/ArgumentProcessor.swift
minuscorp: Yeah, no `return` keyword is needed in one-line methods like `map`, `flatMap`, `filter`, etc.

fastlane/swift/SocketClient.swift
minuscorp: @janpio You're right I should've splitted the work in two tasks, but the fix was only one line, and the linting stuff was something it was in my head for a while now. I hope this serves. The explanation is more or less simple, there was a point were when the Ruby socket sent a failure and the Swift part tried both to write concurrently, causing a deadlock. This frees the write channel when that stack of functions is called.

janpio: Wouldn't it be easier to quickly create a second PR for that 1 line change then? This PR will then be just about the linting after the other one is merged and this one rebased.

PR #16697, opened 2020-06-29
[spaceship] Remove multi_xml dependency

aried3r:
### Checklist

### Motivation and Context

I believe `multi_xml` is not needed. This PR is what #16664 should have been. Thanks to the feedback by @janpio I took another look.

### Description

This was an optional dependency for `faraday_middleware` when parsing XML responses. As far as I can tell, and from what is being tested with [fixtures](https://github.com/fastlane/fastlane/tree/master/spaceship/spec/portal/fixtures), that used to be the case but is not any longer, using the App Store Connect API.

There is a custom plist middleware, but that one is in fastlane itself and does not make use of `multi_xml`. I believe it's used by `#provisioning_profiles_via_xcode_api`.

https://github.com/fastlane/fastlane/blob/2.150.0.rc4/spaceship/lib/spaceship/portal/portal_client.rb#L653-L674
https://github.com/fastlane/fastlane/blob/2.150.0.rc4/spaceship/lib/spaceship/helper/plist_middleware.rb

The client was introduced in https://github.com/fastlane/fastlane/pull/14888.

### Testing Steps

Ran the tests, worked. :)

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -83,7 +83,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency('highline', '>= 1.7.2', '< 2.0.0') # user inputs (e.g. passwords)
   spec.add_dependency('json', '< 3.0.0') # Because sometimes it's just not installed
   spec.add_dependency('mini_magick', '>= 4.9.4', '< 5.0.0') # To open, edit and export PSD files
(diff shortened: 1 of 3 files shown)


joshdholtz (approved):
This is great! I ran across this the other day and wasn't actually sure if it was used anymore but it sure doesn't 😇 Thanks for taking the time to do the research!

PR #16733, opened 2020-07-02
run setup_keychain on macOS environments only

seanreinhardtapps:
### Checklist

### Motivation and Context
`setup_travis` fails on the Linux CI agents in the setup_keychain action, throwing an error and failing the task.
To run Travis CI lanes on Mac and Linux agents, `setup_travis` must be conditionally run.

Since that action only applies to a macOS environment, it could be updated to skip the setup_keychain step on non macOS agents.
🔑
### Description
Exit early out of `setup_keychain` in the `setup_ci` action on non mac-OS envoronments, logging a message that it was skipped.

### Testing Steps
- Run `setup_ci` on a macOS agent without any change in behavior.
- Run `setup_ci` on a linux agent and see that it no longer fails, and logs "Skipping Keychain setup on non-macOS CI Agent"

### DIFF
--- a/fastlane/lib/fastlane/actions/setup_ci.rb
+++ b/fastlane/lib/fastlane/actions/setup_ci.rb
@@ -29,6 +29,11 @@ module Fastlane
           return
         end
 
+        unless Helper.operating_system == 'macOS'
+          UI.message("Skipping Keychain setup on non-macOS CI Agent")
+          return
+        end
+
         keychain_name = "fastlane_tmp_keychain"
         ENV["MATCH_KEYCHAIN_NAME"] = keychain_name
         ENV["MATCH_KEYCHAIN_PASSWORD"] = ""
--- a/fastlane/lib/fastlane/actions/setup_ci.rb
+++ b/fastlane/lib/fastlane/actions/setup_ci.rb
@@ -24,13 +24,13 @@ module Fastlane
       end
 
       def self.setup_keychain
-        unless ENV["MATCH_KEYCHAIN_NAME"].nil?
-          UI.message("Skipping Keychain setup as a keychain was already specified")
+        unless Helper.mac?
+          UI.message("Skipping Keychain setup on non-macOS CI Agent")
           return
         end
 
-        unless Helper.operating_system == 'macOS'
-          UI.message("Skipping Keychain setup on non-macOS CI Agent")
+        unless ENV["MATCH_KEYCHAIN_NAME"].nil?
+          UI.message("Skipping Keychain setup as a keychain was already specified")
           return
         end
 
--- a/fastlane/spec/actions_specs/setup_ci_spec.rb
+++ b/fastlane/spec/actions_specs/setup_ci_spec.rb
@@ -1,6 +1,5 @@
 describe Fastlane do
   describe Fastlane::Actions::SetupCiAction do
-    is_macos_env = RUBY_PLATFORM.downcase.include?("darwin")
     describe "#run" do
       context "when it should run" do
         before do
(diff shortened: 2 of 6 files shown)


seanreinhardtapps:
@googlebot I signed it!

janpio:
Seems you will need some special handling in the tests as well, as they are run on Linux and Windows as well.

mollyIV:
Hey @seanreinhardtapps , @janpio  👋 

The `setup_ci` lane has been extracted as a common piece of logic for specific CI providers. It makes maintenance and future code changes easier. [Here](https://github.com/fastlane/fastlane/pull/14346) is the full context.

According to the source code of a `setup_travis` lane, the action is intended to be ran only on `ios` and `mac` platforms:

https://github.com/fastlane/fastlane/blob/5734f464c121427e5505a632fe28fc944a141540/fastlane/lib/fastlane/actions/setup_travis.rb#L42-L44 

I had a look at `setup_ci` code and besides setting up the keychain and setting up the output paths only for a CircleCI provider, nothing is happening. 

https://github.com/fastlane/fastlane/blob/5734f464c121427e5505a632fe28fc944a141540/fastlane/lib/fastlane/actions/setup_ci.rb#L4-L16

Taking that into account, I am wondering if `setup_travis` should be supported on systems other than macOS right now 🤔 Maybe we could do it just in time (when adding new functionality to the `setup_travis` action that could be supported on the macOS, Linux and Windows), not just in case. 

@seanreinhardtapps could you please provide a bit more context, ideally a use-case, why and when you need to run `setup_travis` on an environment different than macOS?

seanreinhardtapps:
@mollyIV, @janpio thanks for the reviews on this idea.
It's not my intent to actually run `setup_travis` on a linux agent.  My thought was that the `setup_travis` action could skip running on a linux agent, like it already does on a non-CI environment.  

When I looked into the implementation, I thought it would make more sense to skip the `setup_keychain` portion since thats what fails outside of a Mac, and that expands it beyond just Travis CI.

I understand it's not usual to use a linux agent for iOS tests.  I'm running a multi-stage Travis config where the last stage runs a docker container on the agent and starts XCUITests on a device farm.  Thought it could be just a ruby script, it's scripted into a lane along with all of our other tasks.  We had `setup_travis` at the top of our Fastfile before the need for the linux agent.  Now its run conditionally which is fine, but I thought it would be an improvement if it did so automatically.

fastlane/lib/fastlane/actions/setup_ci.rb
```diff
@@ -29,6 +29,11 @@ def self.setup_keychain
           return
         end
 
+        unless Helper.operating_system == 'macOS'
```
joshdholtz, line 32:
```suggestion
        unless Helper.mac?
```

fastlane/spec/actions_specs/setup_ci_spec.rb
```diff
@@ -1,24 +1,33 @@
 describe Fastlane do
   describe Fastlane::Actions::SetupCiAction do
+    is_macos_env = RUBY_PLATFORM.downcase.include?("darwin")
```
joshdholtz, line 3: I think we could probably rework this so the tests aren't dependent on the platform they are being run on 😇 

The following tests should be reworked like 👇 
```rb
describe "calls setup_keychain after setup_output_paths if :provider is set to circleci on Mac Agents" do
  it "on Mac" do
    allow(FastlaneCore::Helper).to receive(:mac?).and_return(true)

    # Message is asserted in reverse order, hence output of setup_output_paths is expected last
    expect(Fastlane::UI).to receive(:message).with("Enabling match readonly mode.")
    expect(Fastlane::UI).to receive(:message).with("Creating temporary keychain: \"fastlane_tmp_keychain\".")
    expect(Fastlane::UI).to receive(:message).with("Skipping Log Path setup as FL_OUTPUT_DIR is unset")
  end

  it "on linux" do
    allow(FastlaneCore::Helper).to receive(:mac?).and_return(false)

    expect(Fastlane::UI).to receive(:message).with("Skipping Log Path setup as FL_OUTPUT_DIR is unset")
    expect(Fastlane::UI).to receive(:message).with("Skipping Keychain setup on non-macOS CI Agent")
  end
end
```

fastlane/spec/actions_specs/setup_circle_ci_spec.rb
```diff
@@ -2,6 +2,21 @@
   describe Fastlane::Actions::SetupCircleCiAction do
     describe "Setup CircleCi Integration" do
       let(:tmp_keychain_name) { "fastlane_tmp_keychain" }
+      is_macos_env = RUBY_PLATFORM.downcase.include?("darwin")
```
joshdholtz, line 5: See above comment for making tests not dependent on platform being executed on ☝️

fastlane/spec/actions_specs/setup_travis_spec.rb
```diff
@@ -2,6 +2,21 @@
   describe Fastlane::FastFile do
     describe "Setup Travis Integration" do
       let(:tmp_keychain_name) { "fastlane_tmp_keychain" }
+      is_macos_env = RUBY_PLATFORM.downcase.include?("darwin")
```
joshdholtz, line 5: See above comment for making tests not dependent on platform being executed on ☝️

fastlane/lib/fastlane/actions/setup_ci.rb
joshdholtz: I also think this should get moved up before the checking of `ENV["MATCH_KEYCHAIN_NAME"]`

joshdholtz (changes requested):
I think the addition of platform checking before setting up keychain makes sense! I have a few suggestions on the code in `SetupCiAction` and on the tests 😊 But once those are fixed I think this should be 👍

mollyIV:
> I think the addition of platform checking before setting up keychain makes sense! 

👍 👍

Let's also remember to change a supported platforms for a `setup_travis` action 😊 

https://github.com/fastlane/fastlane/blob/5734f464c121427e5505a632fe28fc944a141540/fastlane/lib/fastlane/actions/setup_travis.rb#L42-L44

joshdholtz:
@seanreinhardtapps BTW, let me know if you want any help making changes to those tests! I'd be happy to hop in and help 😊

seanreinhardtapps:
Thanks for the reviews!  PR should be ready for another look.

joshdholtz (approved):
This is great! Thank you for making those changes 😊 I will get this out in a release later today for you 💪

PR #16767, opened 2020-07-06
[deliver] set type of auto release date to integer

joshdholtz:
### Motivation and Context
Fixes https://github.com/fastlane/fastlane/pull/16640#issuecomment-654282307

### Description
Fixes issue like 👇 
```rb
upload_metadata.rb:327:in convert_ms_to_iso8601': undefined method /' for "1596146400000":String
```

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -165,7 +165,7 @@ module Deliver
         FastlaneCore::ConfigItem.new(key: :auto_release_date,
                                      env_name: "DELIVER_AUTO_RELEASE_DATE",
                                      description: "Date in milliseconds for automatically releasing on pending approval (Can not be used together with `automatic_release`)",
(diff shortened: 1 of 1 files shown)


myanime (approved):
looks good 👍 I spent nearly half a day trying to figure this out ;). 
EDIT: Was just thinking about this, maybe it might be a good idea to somehow validate if this Integer is at least 13 numbers long, ie milliseconds. Its pretty easy to make a mistake and put seconds in here. Not for this PR obviously, but maybe something to think about down the line.

joshdholtz:
@myanime I like that validation actually 🤔 I will add it to this PR now 😊

joshdholtz:
@myanime 👋 I updated this with a validation to make sure `auto_release_date` is greater than `Time.now.to_i * 1000`. This should make sure its in ms and in the future.

Issue #16749, opened 2020-07-04
[Transporter Error Output]: An exception has occurred: Broken pipe (Write failed)


benasher44:
🔑
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Hi there! We've been following along with all of the recent updates, and everything was up and running great with 2.150.0 and 2.150.1. Then sometime on Friday, we started seeing these connection reset / broken pipe issues, which are still ongoing. It's unclear to me if it's an App Store issue, a fastlane issue, or maybe a unique combination :).

Any advice is appreciated — thanks in advance!

##### Command executed
upload_to_app_store(
    ipa: './PlanGrid.ipa',
    force: true,
    skip_screenshots: true,
    skip_app_version_update: false,
    run_precheck_before_submit: false
)

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>[23:20:00]: ---------------------------------
[23:20:00]: --- Step: upload_to_app_store ---
[23:20:00]: ---------------------------------
[23:20:00]: Successfully loaded '/private/var/lib/jenkins/plangrid-adsk/fastlane/Deliverfile' 📄

+------------------+----------------------------+
| Detected Values from './fastlane/Deliverfile' |
+------------------+----------------------------+
| app_identifier   | com.loupe.PlanGrid         |
| username         | <scrubbed> |
| team_name        | PlanGrid, Inc.             |
+------------------+----------------------------+

[23:20:00]: Login to App Store Connect (<scrubbed>)
[23:20:02]: Login successful

+--------------------------------------+----------------------------+
|                      deliver 2.150.1 Summary                      |
+--------------------------------------+----------------------------+
| ipa                                  | ./PlanGrid.ipa             |
| force                                | true                       |
| skip_screenshots                     | true                       |
| skip_app_version_update              | false                      |
| run_precheck_before_submit           | false                      |
| screenshots_path                     | ./fastlane/screenshots     |
| metadata_path                        | ./fastlane/metadata        |
| app_version                          | 5.86.1                     |
| username                             | <scrubbed> |
| app_identifier                       | com.loupe.PlanGrid         |
| platform                             | ios                        |
[...]

PR #16774, opened 2020-07-06
[deliver] set default ITMSTransporter transport flag to blank #16749

castus:
### Checklist

### Motivation and Context
It solves issue https://github.com/fastlane/fastlane/issues/16749 

### Description
According to [Apple's Transporter User Guide 2.0 documentation](https://help.apple.com/itc/transporteruserguide/#/apdATD1E1288-D1E1A1303-D1E1288A1126) Apple recommends to not  specify transport option and let the tool to determine the best transport mode for packages. It is also visible in the comments, in the mentioned issue that setting this parameter to empty string, fix the issue.

### Testing Steps
Try to use deliver or pilot to send build to TestFlight without `DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS` ENV variable being set.

### DIFF
--- a/fastlane_core/lib/fastlane_core/itunes_transporter.rb
+++ b/fastlane_core/lib/fastlane_core/itunes_transporter.rb
@@ -149,14 +149,22 @@ module FastlaneCore
     end
 
     def additional_upload_parameters
-      # Workaround because the traditional transporter broke on 1st March 2018
-      # More information https://github.com/fastlane/fastlane/issues/11958
-      # As there was no communication from Apple, we don't know if this is a temporary
-      # server outage, or something they changed without giving a heads-up
-      if ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"].to_s.length == 0
-        ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"] = "-t DAV,Signiant"
+      # As Apple recommends in Transporter User Guide we shouldn't specify the -t transport parameter
+      # and instead allow Transporter to use automatic transport discovery
+      # to determine the best transport mode for packages.
+      # It became crucial after WWDC 2020 as it leaded to "Broken pipe (Write failed)" exception
+      # More information https://github.com/fastlane/fastlane/issues/16749
+      env_deliver_additional_params = ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"]
+      if env_deliver_additional_params.nil?
+        return nil
       end
-      return ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"]
+
+      deliver_additional_params = env_deliver_additional_params.to_s.strip
+      if !deliver_additional_params.include?("-t ")
+        UI.user_error!("Invalid transport parameter")
+      else
+        return deliver_additional_params
+      end
     end
   end
 
--- a/fastlane_core/lib/fastlane_core/itunes_transporter.rb
+++ b/fastlane_core/lib/fastlane_core/itunes_transporter.rb
@@ -155,7 +155,7 @@ module FastlaneCore
       # It became crucial after WWDC 2020 as it leaded to "Broken pipe (Write failed)" exception
       # More information https://github.com/fastlane/fastlane/issues/16749
       env_deliver_additional_params = ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"]
-      if env_deliver_additional_params.nil?
+      if env_deliver_additional_params.to_s.empty?
         return nil
       end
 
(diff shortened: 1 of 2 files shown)


fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -149,14 +149,17 @@ def parse_line(line, hide_output)
     end
 
     def additional_upload_parameters
-      # Workaround because the traditional transporter broke on 1st March 2018
-      # More information https://github.com/fastlane/fastlane/issues/11958
-      # As there was no communication from Apple, we don't know if this is a temporary
-      # server outage, or something they changed without giving a heads-up
-      if ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"].to_s.length == 0
-        ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"] = "-t DAV,Signiant"
+      env_deliver_additional_params = ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"]
```
joshdholtz, line 152: Can you add a comment above this with the new issue that this is fixing? 😊 I know that info will come in handy down the road

joshdholtz (changes requested):
This is great! Thanks for fixing 😊 Just one request on adding a comment with the link to the GitHub issue that this is solving 💪 After that we will get this merged and released today!

fastlane_core/lib/fastlane_core/itunes_transporter.rb
castus: Sure, will do within 2 hours

castus: Done, please take a look ;)

joshdholtz: This is perfect! Thank you ❤

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -149,14 +149,22 @@ def parse_line(line, hide_output)
...
-      # server outage, or something they changed without giving a heads-up
-      if ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"].to_s.length == 0
-        ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"] = "-t DAV,Signiant"
+      # As Apple recommends in Transporter User Guide we shouldn't specify the -t transport parameter
+      # and instead allow Transporter to use automatic transport discovery
+      # to determine the best transport mode for packages.
+      # It became crucial after WWDC 2020 as it leaded to "Broken pipe (Write failed)" exception
+      # More information https://github.com/fastlane/fastlane/issues/16749
+      env_deliver_additional_params = ENV["DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS"]
+      if env_deliver_additional_params.nil?
```
joshdholtz, line 158:
```suggestion
      if env_deliver_additional_params.to_s.empty?
```

One last thought! Should we change it to ☝️ so it checks either `nil` or `””`(empty string)? Not checking for empty string might break for users that are explicitly setting it to empty string 🤔 

Sorry about not catching this sooner!

joshdholtz (changes requested):
Caught one more thing! Thoughts appreciated 😊

joshdholtz:
@googlebot I consent.

fastlane_core/lib/fastlane_core/itunes_transporter.rb
joshdholtz: I just added a commit with this change!

castus: Thanks, this is a good idea. I will also add test for blank space.

joshdholtz (approved):
This is ready! Thanks again for making this PR ❤️ Really appreciate it!

fastlane_core/lib/fastlane_core/itunes_transporter.rb
castus: @joshdholtz I have a question. Currently, for users who set this to one space, the script will raise an error, because it isn't a valid transport parameter. I'm wondering should we live it as it is, or treat one space as if the ENV variable isn't set? Doing that the script won't raise and error for current users. What do you think?

joshdholtz: Ah! Good point. I’ll create a follow PR to strip it before checking if empty 🙂

castus: If you'd like I can do it together with tests that validate it.

joshdholtz: Ah, don’t worry about it! I can handle it but I appreciate the offer ❤

PR #16778, opened 2020-07-07
[spaceship] Loosen jwt dependency

aried3r:
### Checklist

### Motivation and Context

I want to be able to use newer versions of the `jwt` gem. Right now, there's no feature that fastlane could use in versions > 2.1.x, but it should also keep us up to date so huge changes and complicated updates don't pile up.

### Description

Allow newer 2.x versions. The gem follows semantic versioning so this should be fine.

https://github.com/jwt/ruby-jwt#development-and-tests


joshdholtz (approved):
This makes sense to me! Thanks for making this change 😊 I appreciate the contribution ❤️

PR #16399, opened 2020-04-29
[Ruby 2.7] Allow Faraday 1.0.0 and Faraday-middleware 1.0.0 with Ruby 2.7 fixes, fix base Ruby 2.3 tests

Kaspik:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I've updated the documentation if necessary.

### Motivation and Context
- partially fixes https://github.com/fastlane/fastlane/issues/16342
- partially fixes https://github.com/fastlane/fastlane/issues/15856
- fixes https://github.com/fastlane/fastlane/issues/16339

### Description
Allow Faraday 1.0.0 as first final stable release with Ruby 2.7 fixes included. We still allow older version for users with older Ruby.

### Testing
- update `Execute modules load up tests` to Ruby 2.4 (should be higher?)
- remove tests for Ruby 2.3 - there should be support for Ruby 2.0 for example but there are no tests for this. All PRs are failing on this and I think we should drop version 2.3 completely and keep supporting only Ruby maintained versions that are not at EOL - https://www.ruby-lang.org/en/downloads/branches/

PR #16793, opened 2020-07-08
[actions] update `danger` to 8, lock faraday to <1

aried3r:
### Checklist

### Motivation and Context

The reason to update was that it was holding other dependencies back, like `kramdown` or `faraday`. Also it was behind several major versions.

### Description

`danger` 8 supports Ruby 2.4+, like fastlane. The range could of course include `danger` 4 up to 8, but since it's a development dependency, it can be as high as we want it.

Talking about `faraday`, ironically it had to be locked to less than 1 because of two failing test cases with `faraday` 1.x. Specifically these two:
`rspec ./spaceship/spec/portal/enterprise_spec.rb:21`
`rspec ./spaceship/spec/portal/portal_client_spec.rb:286`

I believe at least one of them is caused by `webmock` in combination with `faraday` and empty arrays as parameters. See also the changelog of `webmock` for version 3.6, which supposedly fixes this. However I haven't tried since this would entail updating `webmock`.

The `faraday` dependency was initially loosened in a PR, however to no `Gemfile.lock` being committed, it probably remained largely unnoticed that another dependency (`danger`) was holding it back.

https://github.com/danger/danger/blob/v8.0.2/CHANGELOG.md#800
https://github.com/bblimke/webmock/blob/v3.8.3/CHANGELOG.md#360
https://github.com/fastlane/fastlane/pull/16399
https://github.com/fastlane/fastlane/pull/8135

Further reading about committing a `Gemfile.lock` file:
https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html

### Testing Steps

I didn't really see breaking changes in `danger`'s changelog, so I guess we'll see how the danger bot reacts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,8 @@ source("https://rubygems.org")
 
 gem "xcode-install", ">= 2.2.1" # needed for running xcode-install related tests
 
-gem "danger", ">= 4.2.1", "< 5.0.0"
-gem "danger-junit", ">= 0.7.3", "< 1.0.0"
+gem "danger", "~> 8.0"
+gem "danger-junit", "~> 1.0"
 
 gemspec(path: ".")
 
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -76,7 +76,11 @@ Gem::Specification.new do |spec|
   spec.add_dependency('commander-fastlane', '>= 4.4.6', '< 5.0.0') # CLI parser
   spec.add_dependency('excon', '>= 0.71.0', '< 1.0.0') # Great HTTP Client
   spec.add_dependency('faraday-cookie_jar', '~> 0.0.6')
-  spec.add_dependency('faraday', '>= 0.17', '< 2.0') # Used for deploygate, hockey and testfairy actions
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -2,9 +2,6 @@ source("https://rubygems.org")
 
 gem "xcode-install", ">= 2.2.1" # needed for running xcode-install related tests
 
-gem "danger", ">= 4.2.1", "< 5.0.0"
-gem "danger-junit", ">= 0.7.3", "< 1.0.0"
```
joshdholtz, line 6: Can we keep the `danger` and `danger-junit` in the `Gemfile`? They are used in CI and not for development so I _think_ this is still the correct place for them? 😇

joshdholtz (changes requested):
One request on removal of `danger` from the `Gemfile` 🙃

Gemfile
aried3r: You got it!

joshdholtz (approved):
This seems good to me! Thanks for making these changes ❤️

PR #16798, opened 2020-07-08
[fastlane] fixes issue in S3ClientHelper Aws::Credentials were being initialized with blank access_key/secret

atreat:
### Checklist

### Motivation and Context

S3 requests were failing because invalid credentials were being created if you did not explicitly pass an AWS access_key/secret to match.

Caused because nil.to_s == ""
Also adds defensive check for `blank?` values instead of `nil?`

Introduced in https://github.com/fastlane/fastlane/pull/16687

### Description

Removes the `.to_s` cast from parameters that are now optional. `nil.to_s` returns an empty String. Also adds more defensive `.blank?` check when deciding if the AWS credentials need to be created. This will better prevent invalid parameters being passed in.

### Testing Steps

Updated specs to pass in nil parameters. Adds specs around client initialization logic in S3ClientHelper. Tested patch locally and was able to make S3 requests.


joshdholtz (approved):
This makes sense to me! Thanks for the contribution ❤️

PR #16819, opened 2020-07-10
[fastlane_core] add codesign partition ID to keychain when importing

jakubgs:
From my research I found that adding the `codesign` partition ID to the keychain partition list helps with errors like:
```
errSecInternalComponent
Command /usr/bin/codesign failed with exit code
```
or
```
error: The specified item could not be found in the keychain.
```

### DIFF
--- a/fastlane_core/lib/fastlane_core/keychain_importer.rb
+++ b/fastlane_core/lib/fastlane_core/keychain_importer.rb
@@ -39,7 +39,7 @@ module FastlaneCore
       # See https://openradar.appspot.com/28524119
       if Helper.backticks('security -h | grep set-key-partition-list', print: false).length > 0
         command = "security set-key-partition-list"
-        command << " -S apple-tool:,apple:"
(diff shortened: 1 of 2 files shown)


joshdholtz:
@jakubgs Hey 👋  Do you mind signing the CLA when you get a chance? Can't merge without it 🤷‍♂️

joshdholtz (changes requested):
I think this makes sense to me! I'll do some local testing on this too. But it does look like we got a few more failing tests 🤔

jakubgs:
I'm not really interested in reading or accepting any CLA. Maybe I should just open an issue instead?

joshdholtz:
@jakubgs If you want to open an issue that would be great! Sorry about the CLA 😔  Not something we are able to get around at the moment.

jakubgs:
Sure thing, there you go: https://github.com/fastlane/fastlane/discussions/16903

joshdholtz:
@jakubgs Thank you! ❤

Issue #16701, opened 2020-06-29
--submission_information does not seem to be optional for new apps


markmuniz:
### New Release Candidate Issue Checklist

- [x] Updated fastlane to the latest release candidate version (2.150.0.rc4)
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Information
- Breaking version: [`2.150.0.rc1`]
- Last working version: [`2.149.1`]
- State of app: [`Prepare for Submission`]
- First version of app?: [`Yes`]

### Issue Description
Running Deliver with `--submit_for_review true` and no `--submission_options` requires IDFA information to submit.

##### Complete output when running fastlane, including the stack trace and command used
`fastlane deliver --team_id 123456 --username test@test.com --app_identifier com.org.test --skip_metadata true --skip_screenshots true --skip_binary_upload true --submit_for_review true --run_precheck_before_submit false --verbose`
<details>
  <pre>+--------------------------------------+------------------------------------------+
|                           deliver 2.150.0.rc4 Summary                           |
+--------------------------------------+------------------------------------------+
| team_id                              | 123456                                   |
| username                             | test@test.com                            |
| app_identifier                       | com.org.test                             |
| skip_metadata                        | true                                     |
| skip_screenshots                     | true                                     |
| skip_binary_upload                   | true                                     |
| submit_for_review                    | true                                     |
| run_precheck_before_submit           | false                                    |
| screenshots_path                     | ./.fastlane/screenshots                  |
| metadata_path                        | ./.fastlane/metadata                     |
| platform                             | ios                                      |
| edit_live                            | false                                    |
| use_live_version                     | false                                    |
| skip_app_version_update              | false                                    |
| force                                | false                                    |
| overwrite_screenshots                | false                                    |
| reject_if_possible                   | false                                    |
| automatic_release                    | false                                    |
| phased_release                       | false                                    |
| reset_ratings                        | false                                    |
| precheck_default_rule_level          | warn                                     |
| individual_metadata_items            | []                                       |
| ignore_language_directory_validation | false                                    |
| precheck_include_in_app_purchases    | true                                     |
+--------------------------------------+------------------------------------------+

INFO [2020-06-29 11:36:45.51]: Selecting the latest build...
[...]

PR #16837, opened 2020-07-12
[deliver][submit][regression] Fix issue: Deliver failed to select IDFA agreement, if app uses IDFA

crazymanish:
### Checklist

### Motivation and Context
- Earlier (before 2.150.x version), Fastlane(spaceship) always **automatically** select the IDFA agreement, if any app uses IDFA.
- More details: in the below-attached screenshot.

### Description
- After 2.150.x release (new AppStore API migration release), Fastlane(deliver) needs **DEPRECATED** parameter `add_id_info_limits_tracking: true` inside the submission information to select the IDFA agreement. 🤷🏻‍♂️
- If application uses IDFA then Fastlane(deliver) `must select` the mandatory agreement checkbox automatically. 😈

### Testing Steps
Run `deliver` **without** passing `add_id_info_limits_tracking` submission_information.
```ruby
deliver(
  username: "xxxx@xxxx.nl",
  app_identifier: "xxxx-app_identifier",
  app_version: 1.0,
  skip_screenshots: true,
  force: true,
  submit_for_review: true,
  submission_information: {
      add_id_info_uses_idfa: true,
      add_id_info_serves_ads: false,
      add_id_info_tracks_action: true,
      add_id_info_tracks_install: true
  }
)
```

#### More details (about this issue)
Found this issue, when my release pipeline was failed to submit for Apple-review, last week :(
Because I am not sending `deprecated` add_id_info_limits_tracking parameter inside submission_information details.
.
.
Screenshot 2020-07-10 at 17 47 26 copy

#### Pre 2.150.x implementation
spaceship-document

### DIFF
--- a/deliver/lib/deliver/submit_for_review.rb
+++ b/deliver/lib/deliver/submit_for_review.rb
@@ -125,6 +125,7 @@ module Deliver
           "    add_id_info_tracks_action: true,",
           "    add_id_info_limits_tracking: true",
           "  }",
+          "Note: If application uses IDFA, deliver automatically set the mandatory \"add_id_info_limits_tracking\" value to \"true\" always (agreement)",
           "  Example CLI:",
           "    --submission_information \"{\\\"add_id_info_uses_idfa\\\": false}\""
         ].join("\n")
@@ -141,7 +142,8 @@ module Deliver
         end
       elsif uses_idfa == true
         attributes = {
-          honorsLimitedAdTracking: !!submission_information[:add_id_info_limits_tracking],
+          # Application uses IDFA, before sending for submission limitsTracking key in the request JSON must be set to true (agreement).
+          honorsLimitedAdTracking: true,
           servesAds: !!submission_information[:add_id_info_serves_ads],
           attributesAppInstallationToPreviousAd: !!submission_information[:add_id_info_tracks_install],
           attributesActionWithPreviousAd: !!submission_information[:add_id_info_tracks_action]


crazymanish:
cc: @joshdholtz 🚀

crazymanish:
Whoops! one test job failed because of the ruby-version mismatch in CircleCI, is it a known issue? 🤔

rogerluan (changes requested):
I think this change is unwanted, because it sort of makes _fastlane_ legally liable for checking a checkbox that should be explicitly checked by the maintainer of your app. It's controversial, I know, because fastlane operates using an authentication you provide to _fastlane_ 🤷 but I read somewhere that changes like these were unwanted. Ref.: https://github.com/fastlane/fastlane/issues/16701#issuecomment-651260747

Instead, I add this to `submissionInformation` dictionary (aka `submission_information` hash): `"add_id_info_limits_tracking" : true,`, like you mentioned that was deprecated. This solves the issue :) maybe that key shouldn't have been marked as deprecated in the first place 🤔 ?

Anyways, let's also hear from @joshdholtz :)

crazymanish:
> I think this change is unwanted, because it sort of makes _fastlane_ legally liable for checking a checkbox that should be explicitly checked by the maintainer of your app. It's controversial, I know, because fastlane operates using an authentication you provide to _fastlane_ 🤷 but I read somewhere that changes like these were unwanted. Ref.: [#16701 (comment)](https://github.com/fastlane/fastlane/issues/16701#issuecomment-651260747)
> 
> Instead, I add this to `submissionInformation` dictionary (aka `submission_information` hash): `"add_id_info_limits_tracking" : true,`, like you mentioned that was deprecated. This solves the issue :) maybe that key shouldn't have been marked as deprecated in the first place 🤔 ?
> 
> Anyways, let's also hear from @joshdholtz :)

Hi @rogerluan, 
Fastlane already doing it for us for a long time, see below spaceship code, which stops working after 2.150.x release... 🤔
.
spaceship

crazymanish:
Also @rogerluan , my point is very simple
- If the application uses IDFA then Fastlane(deliver) `must select` the mandatory agreement checkbox automatically. 😈

Basically, to me we don't need to send two params always, `add_id_info_uses_idfa: true` should be enough to select the mandatory agreement checkbox automatically.:)

Option1:
```ruby
add_id_info_uses_idfa: true,
add_id_info_limits_tracking: true
```
vs 

Option2:
```ruby
add_id_info_uses_idfa: true
```

I will go with option2 always, less code in my release pipeline, less maintenance because I know Fastlane is super smart and select the `add_id_info_limits_tracking: true` if I am sending `add_id_info_uses_idfa: true` :)

rogerluan:
I agree! Let's wait for legal clearance then. The technical aspect of your solution looks good to me 👍

crazymanish:
> I agree! Let's wait for legal clearance then. The technical aspect of your solution looks good to me 👍

Glad, you liked the solution! 🙇🏻‍♂️
I think, there should not be any legal issue, we are just making Fastlane backward compatible, right? 🤔 
(we are telling Fastlane that Hey my App uses IFDA)

joshdholtz:
@crazymanish Thanks for the PR! I don't think we can actually accept this PR due to legal things 😔 Hardcoding this to `true` isn't an assumption we can make for our users.

Not every app uses IDFA and the App Store is now being more transparent about what kind of tracking apps all use. We need to allow the option for our users to actually set `true` or `false`. 

I think @rogerluan is correct in which that depreciation in the previous code was wrong or not applicable (especially since things changed a lot of the new App Store Connect API).

Thanks again for this PR and I do hope this isn't your last! PRs that touch legal things are always difficult ones. This does make things easier on most users but we need to take responsibility off _fastlane_ and allow flexibility on the users 🤷‍♂️ 

---

Side note 👇 

What I _am_ planning to do is reworking the whole `submission_information` option in `deliver`. It's not very easy to use or document at the moment and its also pretty verbose. I'm thinking of splitting that up and making the verbiage more aligned to what its shown on the App Store Connect website.

crazymanish:
Thank you @joshdholtz @rogerluan 🙌🏻 Closing the PR...:)

joshdholtz:
@crazymanish Thanks for understanding! Always a little heart breaking closing PRs 💔 Hope to still see some more from you in the future!

rogerluan:
@joshdholtz refactoring `submission_information` to split up the hash into their own separate fields is 💯 🤩 looking forward to it!

Issue #16566, opened 2020-06-04
[Fastlane.swift] Swift Package Manager support


UInt2048:
#### Motivation Behind Feature
There is a `swift` subdirectory in Swift Fastlane products exactly mirroring this directory: https://github.com/fastlane/fastlane/tree/master/fastlane/swift

This sounds like the perfect opportunity for a Swift package which could simply be imported in the Xcode project.

#### Feature Description
I'm considering opening a PR, but I wanted to get feedback first since a `Package.swift` would need to be created at the top-level directory of this repository. This file would make it so that these files would instead be imported from this repository in a package, which has the side benefit of preventing users from editing these files.

However, doing this would require users to `import Fastlane` in the Swift files they create using it, such as `Fastfile.swift`. To compensate for this, this feature shouldn't be made mandatory, but optional. Eventually, this could be the default behaviour for new users.

This won't break anything for existing users that still have the existing directory, but documentation would need to be updated.

#### Alternatives or Workarounds
The alternative is, of course, the existing behaviour where over 30 files are added to the user's repository that they can't edit or remove.

turtlemaster19:
@joshdholtz Good to know. We probably would have to not include those in the package, and keep just those two the same way they are now.

Unless, of course, you have any better ideas? I don't really know how that process works, but I would prefer two files to thirty!

joshdholtz:
Ha, agreed... Thirty is a lot of files!

One other thing to think about is that Swift Package would need need to stay in sync with the fastlane version. When an action/tool has updated options, the Swift interface gets updated. We would need to make sure that there is a proper error message when the Swift package is outdated for the version of fastlane being used. I forget how that all works but that is one other thing that would need to get looked at 😊 

I’m totally down for reviewing and testing a PR if you want to give this is a shot. There is a chance that it might not work out if you are okay with that but I’m down for trying if you are down 🙃

UInt2048:
@fastlane-bot Still relevant.

minuscorp:
@UInt2048 you can check out the PR using my branch and tell me your experience! 😁

joshdholtz:
Reviewing this now!!!!! 😊

rogerluan:
This can be closed @joshdholtz @minuscorp since #16844 was released 🚀

PR #16844, opened 2020-07-13
[Fastlane.swift] Swift Package Manager support

minuscorp:
### Checklist

### Motivation and Context
Motivated by: #16566; Fastlane.swift is now a bundled solution to use just from your fastlane copy, but with the raise of Swift Package Manager more and more use cases can be integrated into third-party solutions.

### Description
This PR adds a preliminary support to SPM, this is: the legacy code remains unchanged and the new one keeps awaiting for a significant number of testers to make sure that is stable enough to be merged into the main code.

### Testing Steps

1. Create an executable Swift Package Manager project with `swift package init --type executable`. 

2. Add an entry point or a main.swift file. 

3. `import Fastlane`

4. Create a class with: `class FastFile: LaneFile {}`

5. Add lanes to the class.

6. Modify the target of your executable to have executable arguments `lane myLane` or add them in the call after making `swift build`.


fastlane/swift/main.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  main.swift
+//  mainn.swift
```
rogerluan, line 2:
```suggestion
//  main.swift
```

fastlane/swift/MainProcess.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+//  Created by Joshua Liebowitz on 8/26/17.
+//
+
+//
+//  ** NOTE **
+//  This file is provided by fastlane and WILL be overwritten in future updates
+//  If you want to add extra functionality to this project, create a new file in a
+//  new group so that it won't be marked for upgrade
+//
```
rogerluan, line 13: This header and the 3 lines at the bottom make it look like this file is part of the Swift code generator, although I don't think they were actually auto generated. Maybe cleaning up those files and adding some doc comments would be great to instruct future readers/maintainers how to handle this file 👍

rogerluan (changes requested):
It's incredible the minimum code change that it took to enable fastlane to work with SPM! 😮 

One thing I noticed is that you committed all the code generated by the Swift code generator, and according to the docs ([Testing.md](https://github.com/fastlane/fastlane/blob/master/Testing.md#fastlaneswiftrunner)) this shouldn't be committed, as these changes are part of the release process (and also because this PR currently bumps the swift file versions by a few versions). So maybe the files that were auto-generated should be reverted. The only files I'm on the fence about are `fastlane/swift/main.swift`, `fastlane/swift/MainProcess.swift` and `fastlane/swift/LaneFileProtocol.swift` as they don't look like they were auto-generated (at least not this time) 🤔 

Thanks for this PR! This will be very welcome change to those teams that have been using _fastlane_ Swift! (me included 😉 )

minuscorp:
Thank you for your comments! This is a preliminary version. Why do I commit the autogenerated files? Because there files which generate those files changed and then the result change. When this gets merged, all this files goes to master but when the new release is done they'll be like in the PR, just in case they are needed for the transition or the review process, they are committed on purpose but I take note of that! Also I'll review all the comments tomorrow after work and will do another iteration, don't take my mistakes too seriously 😛. I'll work on getting this working for the community ASAP! 🙌

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -52,7 +48,11 @@ public class LaneFile: NSObject, LaneFileProtocol {
     private static var laneFunctionNames: [String] {
         var lanes: [String] = []
         var methodCount: UInt32 = 0
-        let methodList = class_copyMethodList(self, &methodCount)
+        #if !SWIFT_PACKAGE
+            let methodList = class_copyMethodList(self, &methodCount)
+        #else
+            let methodList = class_copyMethodList(type(of: fastfileInstance!), &methodCount)
+        #endif
```
joshdholtz, line 55: @minuscorp Can we put a comment above this on why the packaged version is different? 😇

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -92,16 +92,25 @@ public class LaneFile: NSObject, LaneFileProtocol {
...
-
-        guard let fastfileInstance: Fastfile = self.fastfileInstance else {
-            let message = "Unable to instantiate class named: \(className())"
-            log(message: message)
-            fatalError(message)
-        }
-
+        #if !SWIFT_PACKAGE
+            loadFastfile()
+        #endif
```
joshdholtz, line 99: Can we add some comments here too on why this isn’t needed on Swift package version? 😇 I just want to know incase I need to come here and debug in the future ❤️

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -92,16 +92,25 @@ public class LaneFile: NSObject, LaneFileProtocol {
...
+                let message = "Unable to instantiate class named: \(className())"
+                log(message: message)
+                fatalError(message)
+            }
+        #else
+            guard let fastfileInstance: LaneFile = fastfile else {
+                log(message: "Found nil instance of fastfile")
+                preconditionFailure()
+            }
+        #endif
```
joshdholtz, line 112: Maybe a comment here too if the comment from above doesn’t also apply here?

fastlane/swift/MainProcess.swift
```diff
@@ -0,0 +1,58 @@
...
+
+import Foundation
+
+let argumentProcessor = ArgumentProcessor(args: CommandLine.arguments)
+let timeout = argumentProcessor.commandTimeout
+
+class MainProcess {
+    var doneRunningLane = false
```
joshdholtz, line 17: Question there! How is this different than `fastlane/swift/main.swift`? It looks similar but I’m not able to tell how this gets used differently than the other one 😇

joshdholtz (changes requested):
Left a few comments about adding comments in places and also one question! Besides that this is 🔥 Thanks so much for doing this!

fastlane/swift/MainProcess.swift
minuscorp: So the issue is that Xcode generates an executable when there's a main.swift, which is considered entry-point for executables, but SPM does the same, uses main.swift as the executable entry point, which conflicts with the fact that Fastlane is not distributed as a executable but as a library. The main.swift lives in the package which declares itself as a executable. That's the reason of the duplication.

UInt2048:
Thank you for doing this! I haven't tested it or anything, but I am hoping it will work and be added.

minuscorp:
Testers would be much appreciated @UInt2048 you can install my branch in a directory and then initialize a `swift package init --type executable` to test it

rogerluan:
> Testers would be much appreciated @UInt2048 you can install my branch in a directory and then initialize a `swift package init --type executable` to test it

This needs to be built using Xcode 12 though, right @minuscorp ?

UInt2048:
@rogerluan It shouldn't. That would be a problem if you did need it.

minuscorp:
Only need Swift 4+. Using Xcode is discouraged.

minuscorp:
Updated to latest `master` @joshdholtz, resolved also all the comments from the first review. Should launch the socket process and print it before letting the fastlane code run.

rogerluan (approved):

minuscorp:
Ping @joshdholtz ❤️

UInt2048:
@minuscorp You might want to resolve the requested changes if you can...

minuscorp:
There are no changes to be resolved. Which aren't they have a reason. Josh's ones are resolved since weeks now

joshdholtz (approved):
@minuscorp This is so great! I think this is getting really close to being shippable 😊 I think one of the last remaining things is that we need to put this in the docs somewhere yet, correct? 

But that is another repo so this actually might be good to merge 🤔 Unless there is anything else you think needs to be done still!

minuscorp:
Yeah, I want to update the docs once this is merged. At first maybe some bugs come up but it is pretty delicate code. We can talk about the docs and I'll try to have a PR for review this weekend.

PR #16853, opened 2020-07-14
Update webmock, faraday

aried3r:
### Checklist

### Motivation and Context

Keep dependencies up to date. Using a newer webmock, we can also use a newer faraday version.

### Description

Because of https://github.com/lostisland/faraday/pull/1084 the stubs had to be modified. Alternatively, we could filter out blank parameters so they don't show up in the resulting URL. Or the tests could stop calling it with an empty array and instead pass `nil`, then we could call `.compact` or let faraday handle it.

https://github.com/fastlane/fastlane/blob/4342b0049890ea9eb3f088cb01f87ee129fb615e/spaceship/lib/spaceship/portal/portal_client.rb#L689-L696

https://github.com/fastlane/fastlane/blob/4342b0049890ea9eb3f088cb01f87ee129fb615e/spaceship/spec/portal/portal_client_spec.rb#L288


joshdholtz (approved):
This looks great to me! Thanks for updating these 😊 Appreciate the contribution ❤️

Issue #16440, opened 2020-05-08
Snapshot simulator location not showing when on "external display"/headless


Steffion:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
For my project I would like to automate the screenshot process with snapshot. My app uses a map view to show the user's current location, which I would like to include in one of the screenshots. My problem is the following:
When using a simulator location with snapshot, the 'current location' is not showing up on the map view when the simulator is in headless/external display mode.
A workaround is to open the Simulator app manually after the script kills the simulators, but this way, I cannot combine snapshot with my CI. (after `Patching simulators '/Users/username/Library/Preferences/com.apple.iphonesimulator.plist' to scale to 100%`)

What I tried so far:
- Setting the simulator location in the project scheme.
- Using set-simulator-location in the pre-action test script.
- Reinstalling the app when snapshot starts.
- Resetting the simulator when snapshot starts.

I am aware of the other issues around regarding this problem, but it didn't help me find a working fix or workaround. (for example: #1678, #10381)

What I would like:
If possible, it would be nice if there is a fix for this when the simulator is not showing on the user's screen. Otherwise: make an option to run the simulator on the screen?

If any more info is required, let me know.

##### Command executed

fastlane snapshot

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>[⠦] 🚀 /Users/username/project_x_folder/vendor/bundle/ruby/2.7.0/gems/faraday_middleware-0.13.1/lib/faraday_middleware/response_middleware.rb:14: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
[✔] 🚀 
[10:20:27]: Successfully loaded '/Users/username/project_x_folder/fastlane/Snapfile' 📄

+----------------------------+-----------------------+
|     Detected Values from './fastlane/Snapfile'     |
+----------------------------+-----------------------+
| devices                    | ["iPhone 11 Pro Max"] |
| languages                  | ["nl-NL"]             |
| scheme                     | Project X             |
| clear_previous_screenshots | true                  |
+----------------------------+-----------------------+

[10:20:27]: Resolving Swift Package Manager dependencies...
[...]

Issue #12682, opened 2018-06-04
[snapshot] Add parameter to not use headless simulator


mmysliwiec:
#### Motivation Behind Feature
Xcode 9 introduced headless mode for UI testing. It's used by default when no simulator is running. But Snapshot is killing simulator by default to fix some issues (https://github.com/fastlane/fastlane/issues/2533). Unfortunately, this will always use headless simulator while taking screenshots. It's an issue in rare circumstances when headless simulator behaves differently than UI simulator. One blocker for me is when screenshot relly on geolocation. The headless simulator ignores settings for location (Debug->Location->Custom Location). This is a huge problem because all screenshots with maps show my country silhouette instead of my simulated location.

#### Feature Description
New Snapshot configuration parameter should be added (e.g. "headless") with the default value of true. Default parameter would kill simulator as it is now so no changes for current scenarios. When set to false it should reopen simulator with 'open "/Applications/Xcode.app/Contents/Developer/Applications/Simulator.app"' equivalent after applying all fixes.

#### Alternatives or Workarounds
One workaround is to manually launch simulator after Snapshot is killing it. Then the headless mode is off and Snapshot is launching UI test on the opened simulator. The geolocation is configured properly then. The problem is when we have more than one device inside devices list. The simulator is killed before each device is picked to make screenshots. So I could manually monitor the situation and relaunch the simulator each time Snapshot kills it but it's useless for automating the process.

Leeds-eBooks:
This is still an issue. Is it impossible to fix, or just not high priority?

mmysliwiec:
It's still an issue for me and anyone who want to make screenshots with specific geo-location showing user position on the map. I guess it's not high priority enough 😢.

Leeds-eBooks:
My use case is different, but it's the same cause – I'm using React Native, and it appears that nothing renders in a headless simulator, so all of my "tests" (which snapshot uses to take screenshots) fail.

Leeds-eBooks:
A **very** hacky workaround I've just used is to set a trigger in iTerm2 that alerts me to open the simulator manually whenever the fastlane output shows the test is about to run...!

m-ruhl:
We have the same issue: Our UI-Tests and Snapshot-Tests require a geo-location to function properly..

These workarounds would only work local, not on a CI environment

janpio:
[...]

Issue #10381, opened 2017-09-20
Make simulator visible when snapshot runs


vmalyi:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I'm currently facing with an issue when one of my UI tests works well when started directly in Xcode but has some issues when I use it in **snapshot** for generating screenshots.

I would like to visually debug the issue but the simulator is not shown when running snapshot in Xcode 9. 

Could you please add an option to make simulator visible when snapshot is running?

Thanks in advance!

KrauseFx:
Hey @vmalyi, that's a great idea, yeah. We were thinking, you could debug the issue by using Xcode to run the UI Tests, would that work for you?

vmalyi:
@KrauseFx thanks for the reply! 

I already follow this approach. The problem is that the **tests work when launched in Xcode** and **don't work when launched from snapshot**. 

The problem I'm facing is 100% related to the app under test and tests itself. I just wanted to see what exactly happens when snapshot runs, since I can't reproduce it in Xcode.

PS: I went through options of _xcodebuild_ command and didn't find anything related to enabling the simulator windows.

I will appretiate a solution from you if this feasible from your point of view!

haydus:
I have a similar issue and it seems that when running tests even a warning causes a SIGABORT, where running normally just outputs a warning in the console.
Mine was the new UI thread checker introduced in iOS 11 saying "Checker: UI API called on a background thread".
As a workaround i use snapshot in iOS 10.2

carlosefonseca:
@haydus Disable Main Thread Checker on your scheme. ![](https://dsc.cloud/carlosefonseca/Screen-Shot-2017-09-26-11-33-41-copy-NGFF5.png)

haydus:
Thanks!
@carlosefonseca will that setting affect the way snapshot is running?

carlosefonseca:
@haydus I don't know. I'm still trying to get snapshot running again since upgrading to Xcode9… But that check box was a problem running the tests on Xcode itself.

carlosefonseca:
[...]

PR #16863, opened 2020-07-15
[snapshot] add headless option

ChaosCoder:
### Checklist

### Motivation and Context

When snapshotting simulators we are currently not showing the simulator windows. This has UI side effects, one of the more common problems is that a simulated user location (via Simulators `Features -> Location -> *`) will not be shown. This is especially cumbersome for apps that display maps, as the "blue dot" never appears. #16440 #10381 #12682 #1678

This PR introduces a new `headless` option. If `headless` is disabled, snapshot will take care of opening the simulator after it has been killed by snapshot and applying some simulator settings.

### Description

Introducing a new `headless` parameter and documenting it. The `headless` parameter is `true` by default, leaving the old behavior as default. After killing the simulator and applying the settings (like localization, appearance, slide to type), the simulator will now be opened again, if `headless` is `false`.

### Testing Steps

Specify the `headless` parameter when invoking snapshot and set it to `false`, or leave it off completely to get the old (default) behavior. Set it to `true` and the Simulator.app will be launched before the snapshot tests are invoked.


snapshot/lib/snapshot/simulator_launchers/simulator_launcher_base.rb
```diff
@@ -76,6 +76,11 @@ def prepare_simulators_for_launch(device_types, language: nil, locale: nil)
           disable_slide_to_type(type)
         end
       end
+
+      unless launcher_config.headless
+        simulator_path = File.join(Helper.xcode_path, 'Applications', 'Simulator.app')
+        Helper.backticks("open -a #{simulator_path} -g", print: FastlaneCore::Globals.verbose?)
+      end
```
janpio, line 83: That easy? Nice.

ChaosCoder: Yeah, that easy. As I said, snapshot kills the Simulator.app process and never starts it again. The workaround was to manually open the Simulator.app after snapshot closed it, so now we just do that automatically. 😅

janpio (approved):

joshdholtz (commented):
@ChaosCoder Hey 👋 Code looks great! Just a question about behavior here 😊

I pulled this down and ran it with both `headless: true` and `headless: false`. The simulators opened for me both times. Is that not what I should be expecting? Or is headless a different meaning here? 🤔

ChaosCoder:
> @ChaosCoder Hey 👋 Code looks great! Just a question about behavior here 😊
> 
> I pulled this down and ran it with both `headless: true` and `headless: false`. The simulators opened for me both times. Is that not what I should be expecting? Or is headless a different meaning here? 🤔

Specifying `headless: false` should not open the simulators. Can you run it with verbose enabled? Was the Simulator.app process killed by snapshot (e.g. is the Simulators.app running)? It might be, that it was never killed because snapshot did not apply any special settings to the simulators (just a guess).

ChaosCoder:
@googlebot I consent.

joshdholtz:
@googlebot I consent.

joshdholtz:
> > @ChaosCoder Hey 👋 Code looks great! Just a question about behavior here 😊
> > I pulled this down and ran it with both `headless: true` and `headless: false`. The simulators opened for me both times. Is that not what I should be expecting? Or is headless a different meaning here? 🤔
> 
> Specifying `headless: false` should not open the simulators. Can you run it with verbose enabled? Was the Simulator.app process killed by snapshot (e.g. is the Simulators.app running)? It might be, that it was never killed because snapshot did not apply any special settings to the simulators (just a guess).

Yeah yeah yeah, this is all good now! I had to attempt to force quit simulators a few times 🤷‍♂️

joshdholtz (approved):
Got this working! Thanks for adding this and waiting for me to figure out how to test it properly ❤️ Really appreciate the contribution!

hactar:
@ChaosCoder @joshdholtz Could we have this for scan too? We have the same issue when running our tests - User Location simulation doesn't work reliably in headless mode...

PR #16870, opened 2020-07-16
[frameit] add font_size parameter

ChaosCoder:
### Checklist

### Motivation and Context

_frameit_ scales the `keyword` and `title` automatically per screenshot to fit the available size above or below the device. This is a good approach for simple screenshots, but lacks a **consistent** and **unified appearance** across the screenshots.

Specifying a fixed `font_size` for keyword and title (separately) gives users of _frameit_ the ability to fix the font size and keep it consistent across screenshots.

### Description

This PR introdocudes a new parameter `font_size` for the `keyword` and `title` parameters. Instead of calculating the `actual_font_size` from the available space everywhere, it now checks for the `font_size` and returns it if specified and  `font_size  > 0`. This way, we keep the old behavior intact.

### Testing Steps

Specifiy a `font_size` in either the `default` config in your `Framefile.json` or for a single screenshot config for the `title` or `keyword`. Frameit should use this font size when generating the framed screenshots.

### DIFF
--- a/frameit/lib/frameit/editor.rb
+++ b/frameit/lib/frameit/editor.rb
@@ -298,7 +298,7 @@ module Frameit
       resize_text(keyword)
 
       vertical_padding = vertical_frame_padding # assign padding to variable
-      spacing_between_title_and_keyword = (actual_font_size / 2)
+      spacing_between_title_and_keyword = (actual_font_size('keyword') / 2)
       title_left_space = (background.width / 2.0 - title.width / 2.0).round
       keyword_left_space = (background.width / 2.0 - keyword.width / 2.0).round
 
@@ -359,7 +359,7 @@ module Frameit
       vertical_padding = vertical_frame_padding # assign padding to variable
       left_space = (background.width / 2.0 - sum_width / 2.0).round
 
-      self.space_to_device += actual_font_size + vertical_padding
+      self.space_to_device += actual_font_size('title') + vertical_padding
 
       if title_below_image
         title_top = background.height - effective_text_height / 2 - title.height / 2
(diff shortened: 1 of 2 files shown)


janpio (approved):
Great addition, although a bit scary. Sure this will not change existing behavior?

ChaosCoder:
> Great addition, although a bit scary. Sure this will not change existing behavior?

The code now checks (see below), if a `font_size` is configured for either keyword or title and only if it does, it returns that. If not, the old implementation continues. So yes, I'm sure this does not change existing behavior.

```
       font_size = @config[key.to_s]['font_size']
       return font_size if !font_size.nil? && font_size > 0
```

frameit/lib/frameit/editor.rb
```diff
@@ -385,15 +385,18 @@ def put_title_into_background(background, stack_title)
       background
     end
 
-    def actual_font_size
+    def actual_font_size(key)
+      font_size = @config[key.to_s]['font_size']
```
joshdholtz, line 389:
```suggestion
      font_size = (@config[key.to_s] ||  {})['font_size']
```

Can we add an extra safety check for this? 👆 I don’t know if that key will also be there so just want to make sure there is an empty hash incase its not 😇

joshdholtz (changes requested):
One small safety check!

frameit/lib/frameit/editor.rb
ChaosCoder: Okay, I will add that. I'm no ruby dev, but looking at all other `@config[xxx]` subscripts in the file we never check for that (even in the cases we use the same key `@config[key.to_s]`). Why should we treat this one differently?

joshdholtz: @ChaosCoder Ah, really?! The kind of scares me a little bit but 🤷‍♂️ I’m not as familiar with the `frameit` codebase so I wasn’t aware that is how the other areas accessed this values. Its probably fine then to keep the code what you had 😇

joshdholtz (approved):
This looks good to me! Thank you so much for adding this and responding to my review! Really appreciate the contribute ❤️

PR #16920, opened 2020-07-23
[create_keychain] Populate KEYCHAIN_NAME to lane variables upon creation of keychain

alex-benoit:
### Checklist

### Motivation and Context
This change populates the share lane variable `KEYCHAIN_NAME` so you can use it in subsequent actions such as `import_certificate`. Closes https://github.com/fastlane/fastlane/discussions/16906

### Description
I am not sure if we want to support the use-case where a user passes `foo/bar.keychain` as keychain name. Currently do `keychain_path.split('/').last` to get the keychain name and support both `:name` or `:path` options, but not tied to this implementation.

### DIFF
--- a/fastlane/lib/fastlane/actions/create_keychain.rb
+++ b/fastlane/lib/fastlane/actions/create_keychain.rb
@@ -26,13 +26,14 @@ module Fastlane
 
         if !exists?(keychain_path)
           commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{keychain_path}", log: false)
-          Actions.lane_context[Actions::SharedValues::KEYCHAIN_PATH] = keychain_path
         elsif params[:require_create]
           UI.abort_with_message!("`require_create` option passed, but found keychain '#{keychain_path}', failing create_keychain action")
         else
           UI.important("Found keychain '#{keychain_path}', creation skipped")
           UI.important("If creating a new Keychain DB is required please set the `require_create` option true to cause the action to fail")
         end
+
+        Actions.lane_context[Actions::SharedValues::KEYCHAIN_PATH] = keychain_path
 
         if params[:default_keychain]
           # if there is no default keychain - setting the original will fail - silent this error
@@ -94,7 +95,7 @@ module Fastlane
       def self.output
         [
           ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the default keychain'],
-          ['KEYCHAIN_PATH', 'The path of the newly created keychain']
+          ['KEYCHAIN_PATH', 'The path of the keychain']
         ]
       end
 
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -51,6 +52,8 @@ def self.run(params)
 
         commands << Fastlane::Actions.sh(command, log: false)
 
+        Actions.lane_context[Actions::SharedValues::KEYCHAIN_NAME] = keychain_path.split('/').last
```
alex-benoit, line 55: not tied to this implementation - but did this to cater for both `:name` or `:path` options

alex-benoit: not sure if we want to test for the `name: 'my/key/chain'` case

alex-benoit:
I changed my mind on the implementation. 1st version was using the `KEYCHAIN_NAME` variable but led to some weird magic to set the correct name if pass only the `:path` option to create the keychain https://github.com/fastlane/fastlane/pull/16920/commits/9eda209cf8424ec76f6b93d19df2f06953bd41e8#diff-0231a345cb1ba3f328959032b555024cR55

In this case we only populate the `KEYCHAIN_PATH` variable, that also works with https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/helper.rb#L241

Also cleaned up specs

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -25,6 +26,7 @@ def self.run(params)
 
         if !exists?(keychain_path)
           commands << Fastlane::Actions.sh("security create-keychain -p #{escaped_password} #{keychain_path}", log: false)
+          Actions.lane_context[Actions::SharedValues::KEYCHAIN_PATH] = keychain_path
```
joshdholtz, line 29: Do we want to bump this outside and above of this if statement? I think it would be good to set the lane context even if the keychain already exists 😊 Thoughts?

fastlane/lib/fastlane/actions/create_keychain.rb
```diff
@@ -91,7 +93,8 @@ def self.description
 
       def self.output
         [
-          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the default keychain']
+          ['ORIGINAL_DEFAULT_KEYCHAIN', 'The path to the default keychain'],
+          ['KEYCHAIN_PATH', 'The path of the newly created keychain']
```
joshdholtz, line 97:
```suggestion
          ['KEYCHAIN_PATH', 'The path of the keychain']
```

joshdholtz (changes requested):
This looks great! Thanks for making this change 😊 Just a small suggestion on placement of setting of the lane context

fastlane/lib/fastlane/actions/create_keychain.rb
alex-benoit: yes of course!

alex-benoit:
@joshdholtz Implemented suggested changes 👌

joshdholtz:
Thank you! I will review and get this merged later tonight / tomorrow morning 💪

mollyIV (approved):
LGTM 👍 

Thanks for the contribution @alex-benoit 🎉

joshdholtz (approved):
This is 🔥 Thank you so much! Really appreciate the contribution ❤️

PR #16936, opened 2020-07-25
[fastlane_core] Update the simulator device state

lyndsey-ferguson:
When booting, or shutting down, we save the new state.

### Checklist

### Motivation and Context

When interacting with the simulator devices, the `:state` value can be left in an old state. Code that relies on a device being booted, for example, may fail if the `shutdown` method was called and the state was not updated.

### Description

Change the state for the device when booting or shutting down.

### Testing Steps

I tested locally to ensure that when a device is booted from shutdown state, it boots, otherwise, it does not boot. The same way around for shutting down a device.

### DIFF
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -206,7 +206,7 @@ module FastlaneCore
 
         UI.message("Booting #{self}")
 
-        `xcrun simctl boot #{self.udid}`
+        `xcrun simctl boot #{self.udid} 2>/dev/null`
         self.state = 'Booted'
       end
 
@@ -216,7 +216,7 @@ module FastlaneCore
         return if self.state != 'Booted'
 
         UI.message("Shutting down #{self.udid}")
-        `xcrun simctl shutdown #{self.udid}`
+        `xcrun simctl shutdown #{self.udid} 2>/dev/null`
         self.state = 'Shutdown'
       end
 
--- a/fastlane_core/lib/fastlane_core/device_manager.rb
+++ b/fastlane_core/lib/fastlane_core/device_manager.rb
@@ -202,17 +202,22 @@ module FastlaneCore
       def boot
         return unless is_simulator
         return unless os_type == "iOS"
+        return if self.state == 'Booted'
 
         UI.message("Booting #{self}")
-        `xcrun simctl boot #{self.udid}` unless self.state == "Booted"
+
+        `xcrun simctl boot #{self.udid}`
+        self.state = 'Booted'
       end
 
       def shutdown
         return unless is_simulator
         return unless os_type == "iOS"
+        return if self.state != 'Booted'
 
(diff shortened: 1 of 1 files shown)


fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -202,17 +202,22 @@ def to_s
       def boot
         return unless is_simulator
         return unless os_type == "iOS"
+        return if self.state == 'Booted'
 
         UI.message("Booting #{self}")
-        `xcrun simctl boot #{self.udid}` unless self.state == "Booted"
+
+        `xcrun simctl boot #{self.udid} 2>/dev/null`
+        self.state = 'Booted'
```
janpio, line 210: How was this updated before after booting?

lyndsey-ferguson: Hi @janpio, are you asking how was this updated before I added the `boot` method? This method didn't exist, so simulators were booted using the [`Simulator.launch` method](https://github.com/fastlane/fastlane/pull/16833/files#diff-b6a3ba6dff446741cb76248e99156be7L47) before I added the `boot` method.

If you're asking, "why do we care what the state is", when I call the method that collects the logs for example, I may want to check, "is the device booted", so that I can avoid getting the error that occurs when getting logs from a shutdown simulator.

If you're asking, "how is this state not matching the actual device's state?", that's a different story. The `FastlaneCore::DeviceManager.simulators('iOS')` method can be used to create a _new_ `FastlaneCore::DeviceManager::Device` and that instance's shutdown method can be called to make the other state invalid. So, the new Device's state is shutdown, but the first Device's state is booted. I could have added code for the state to make it a method and _always_ query `simctl`, but that seemed potentially expensive, and nothing terrible seems to happen if I ignore the error, so that's why I put the `2>/dev/null` to the command.

Do any of those answer the question?

minuscorp, line 210: Maybe out of scope but I'd rather see this as symbols than strings, I think they stick better but I think that would imply a wider refactor.

lyndsey-ferguson: That would affect quite a bit. You'll see that when the [device instance is created](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/device_manager.rb#L54), it is using the grepped string from the `xcrun simctl` command to get the state.

janpio: I was not aware you added the `boot` method. So it this a follow up to an earlier PR?

lyndsey-ferguson: Yes, this a follow up change. I saw a problem when this code was being exercised and wanted to clean it up.

# 🧼

janpio (approved):

joshdholtz (approved):
This seems straight forward enough to me! 👍 Thanks again for the contribution!

PR #16948, opened 2020-07-27
[match] Updated match documentation to with hashing algorithm option

nxtstep:
🔑

Updated match encryption documentation to prevent further confusion when trying to debug/decrypt the _match_ managed and encrypted certificates/profiles.

As stated in the `match/lib/match/encryption/openssl.rb:110/138` the default behavior of openssl has changed with regards to the password hashing algorithm. The openssl default is now `sha256` where _match_ is still encrypting with a `MD5` password hashing algorithm.

### Checklist
- [x] I've updated the documentation if necessary. <--

### Motivation and Context
The fact that openssl changed their default password hashing algorithm from `md5` to the more cryptographically secure `sha256` has led to numerous hours of debugging match source-code and confusion as to why I couldn't successfully decrypt the _match_ managed certificates and profiles. To mitigate this for my future self and possibly others I'd like to add this hint to the documentation.

### Description
Added a hint to the match Manual Decrypt section: `fastlane/lib/fastlane/actions/docs/sync_code_signing.md`

### Testing Steps
Just try to decrypt a _match_ managed certificate/profile with the right password and right/wrong hashing algorithm

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
+++ b/fastlane/lib/fastlane/actions/docs/sync_code_signing.md
@@ -488,14 +488,14 @@ _match_ stores the certificate (`.cer`) and the private key (`.p12`) files separ
 Decrypt your cert found in `certs/<type>/<unique-id>.cer` as a pem file:
 
 ```no-highlight
-openssl aes-256-cbc -k "<password>" -in "certs/<type>/<unique-id>.cer" -out "cert.der" -a -d
+openssl aes-256-cbc -k "<password>" -in "certs/<type>/<unique-id>.cer" -out "cert.der" -a -d -md [md5|sha256]
 openssl x509 -inform der -in cert.der -out cert.pem
 ```
 
 Decrypt your private key found in `certs/<type>/<unique-id>.p12` as a pem file:
 
 ```no-highlight
-openssl aes-256-cbc -k "<password>" -in "certs/distribution/<unique-id>.p12" -out "key.pem" -a -d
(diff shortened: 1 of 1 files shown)


janpio:
Should match be changed? Can it even?

nxtstep:
> Should match be changed? 

it would be good to eradicate `md5` password hashing 👍 

> Can it even?

Sure, we have the source-code 💻  🤓

janpio:
> > Can it even?
> 
> Sure, we have the source-code 💻 🤓

Touché - should have defined that my question is about backwards compatibility of course.

nxtstep:
> > > Can it even?
> > 
> > 
> > Sure, we have the source-code 💻 🤓
> 
> Touché - should have defined that my question is about backwards compatibility of course.

Well, forward compatibility has already been foreseen apparently: `match/lib/match/encryption/openssl.rb:150~153`
So backwards compatibility is already in place, we only need to flip the default to `sha256` and test it I suppose 🤞

joshdholtz:
_match_ was changed years ago to handle this "non-breaking" version of openssl. It was a great surprise at the time 😛 This doc change makes sense! 

>Well, forward compatibility has already been foreseen apparently: match/lib/match/encryption/openssl.rb:150~153
So backwards compatibility is already in place, we only need to flip the default to sha256 and test it I suppose 🤞

This _does_ feel like something that could be done and it _doesn't_ look like it would break anything but it still scares me to do a little bit since things are working right now 😇 That's a public method and some users I do think call directly so changing that default value might lead to a breaking change for them 😬

joshdholtz (approved):
This looks great to me! Thanks for adding this ❤️ Really appreciate it!

PR #16953, opened 2020-07-28
[spaceship] fix infinite retry of with_asc_retry

valerio-castelli:
### Checklist

### Motivation and Context

The previous implementation of `with_asc_retry` decremented the number of attempts only after a response from the yield block. However, if the yield block raised an unhandled exception, the execution jumped straight to the rescue block without decrementing the number of retry attempts. This led to an infinite loop of retries.

### Description

The fix moves the `tries -= 1` instruction at the beginning of the rescue block. The block is always entered upon a 500 or 504 response from the yield block thanks to the manually raised exception, so the previous behaviour is preserved; however, the number of retries will now be properly decremented also in case of unhandled exceptions raised within the yield block.

### Testing Steps

We've verified the fix by deploying the change in production on our business infrastructure, and it has indeed solved the infinite retry issue we were experiencing. We've also run `bundle exec rspec` to verify that existing tests are still valid.

### DIFF
--- a/spaceship/lib/spaceship/connect_api/client.rb
+++ b/spaceship/lib/spaceship/connect_api/client.rb
@@ -129,7 +129,6 @@ module Spaceship
         tries = 1 if Object.const_defined?("SpecHelper")
         response = yield
 
-        tries -= 1
         status = response.status if response
 
         if [500, 504].include?(status)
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
This makes a lot of sense! Thank you ❤️ Really appreciate the contribution!

PR #16982, opened 2020-08-02
[snapshot] Add `suppress_xcode_output` option

### LINKED ISSUE #16792: SNAPSHOT - SUPPRESS OUTPUT LOGS WARNINGS during SCREENSHOTS capture.
**FASTLANE SNAPSHOT LANE - SUPPRESS OUTPUT LOGS**

#### Feature Description

is there any way to suppress the logs warnings when running the SNAPSHOT lane during screenshots. similar to Xcode suppress_xcode_output: true

rogerluan:
🔑 

### Checklist

### Motivation and Context

Thanks to the `You can do this` label, this PR resolves #16792 🚀 

### Description

This change is quite simple actually, it simply adds a `suppress_xcode_output` to `snapshot` action, similar to the analogous argument in `scan` and `gym`.

### Testing Steps

`bundle exec fastlane snapshot --suppress_xcode_output=true`

### DIFF
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -34,7 +34,7 @@ module Snapshot
         xcpretty = "xcpretty #{Snapshot.config[:xcpretty_args]}"
         xcpretty << "--no-color" if Helper.colors_disabled?
         pipe << "| #{xcpretty}"
-        pipe << " > /dev/null" if Snapshot.config[:suppress_xcode_output]
+        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
         return pipe
       end
 
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -34,7 +34,7 @@ module Snapshot
         xcpretty = "xcpretty #{Snapshot.config[:xcpretty_args]}"
         xcpretty << "--no-color" if Helper.colors_disabled?
         pipe << "| #{xcpretty}"
-        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
+        pipe << " > /dev/null" if Snapshot.config[:suppress_xcode_output]
         return pipe
       end
 
--- a/snapshot/lib/snapshot/test_command_generator.rb
+++ b/snapshot/lib/snapshot/test_command_generator.rb
@@ -33,8 +33,9 @@ module Snapshot
 
         xcpretty = "xcpretty #{Snapshot.config[:xcpretty_args]}"
         xcpretty << "--no-color" if Helper.colors_disabled?
-
-        return pipe << "| #{xcpretty}"
+        pipe << "| #{xcpretty}"
+        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
+        return pipe
       end
 
       def destination(devices)
--- a/snapshot/lib/snapshot/test_command_generator_xcode_8.rb
+++ b/snapshot/lib/snapshot/test_command_generator_xcode_8.rb
@@ -24,7 +24,10 @@ module Snapshot
 
       def pipe(device_type, language, locale)
         log_path = xcodebuild_log_path(device_type: device_type, language: language, locale: locale)
-        return ["| tee #{log_path.shellescape} | xcpretty #{Snapshot.config[:xcpretty_args]}"]
+        pipe = "| tee #{log_path.shellescape} "
+        pipe << "| xcpretty #{Snapshot.config[:xcpretty_args]}"
+        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
+        return [pipe]
       end
 
       def destination(device_name)
--- a/snapshot/lib/snapshot/test_command_generator_xcode_8.rb
+++ b/snapshot/lib/snapshot/test_command_generator_xcode_8.rb
@@ -26,7 +26,7 @@ module Snapshot
         log_path = xcodebuild_log_path(device_type: device_type, language: language, locale: locale)
         pipe = "| tee #{log_path.shellescape} "
         pipe << "| xcpretty #{Snapshot.config[:xcpretty_args]}"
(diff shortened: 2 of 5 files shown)


rogerluan:
I can't see what failed there ☝️ the tests pass locally, I may need some help 😬

snapshot/lib/snapshot/test_command_generator_xcode_8.rb
```diff
@@ -24,7 +24,10 @@ def generate(device_type: nil, language: nil, locale: nil)
 
       def pipe(device_type, language, locale)
         log_path = xcodebuild_log_path(device_type: device_type, language: language, locale: locale)
-        return ["| tee #{log_path.shellescape} | xcpretty #{Snapshot.config[:xcpretty_args]}"]
+        pipe = "| tee #{log_path.shellescape} "
+        pipe << "| xcpretty #{Snapshot.config[:xcpretty_args]}"
+        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
```
joshdholtz, line 29: Do we need a space before the `>`?

snapshot/lib/snapshot/test_command_generator.rb
```diff
@@ -33,8 +33,9 @@ def pipe(log_path: nil)
 
         xcpretty = "xcpretty #{Snapshot.config[:xcpretty_args]}"
         xcpretty << "--no-color" if Helper.colors_disabled?
-
-        return pipe << "| #{xcpretty}"
+        pipe << "| #{xcpretty}"
+        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
```
joshdholtz, line 37: Do we need a space before the `>`?

joshdholtz (changes requested):
This looks great! Thanks for tackling this 😊 Just a question on if a space is needed before the `>` 😇

snapshot/lib/snapshot/test_command_generator_xcode_8.rb
rogerluan: I had initially written it with a space, and while testing I noticed it'd print out e.g. `…| xcpretty  > /dev/null` (double spaces), so I removed the leading space here.
But now, revisiting this, there's also the `xcpretty_args` case, which needs a space.

Do we care about the command style if it contains double spaces? It's just a mere styling issue cuz obv it works with multiple extra whitespaces :) If we do care about making the command look pretty, I gotta check how to write this neatly in ruby 😇

rogerluan:
```suggestion
        pipe << "| xcpretty #{Snapshot.config[:xcpretty_args]}"
        pipe << " > /dev/null" if Snapshot.config[:suppress_xcode_output]
```

snapshot/lib/snapshot/test_command_generator.rb
rogerluan:
```suggestion
        pipe << "| #{xcpretty}"
        pipe << " > /dev/null" if Snapshot.config[:suppress_xcode_output]
```

rogerluan:
Thanks for bringing that up @joshdholtz ! Check my comment above on the rationale 🙏 For now, I added the extra space in both occurrences that you raised. If you agree, it's good to be merged 💪

snapshot/lib/snapshot/test_command_generator.rb
```diff
@@ -33,8 +33,9 @@ def pipe(log_path: nil)
 
         xcpretty = "xcpretty #{Snapshot.config[:xcpretty_args]}"
         xcpretty << "--no-color" if Helper.colors_disabled?
-
-        return pipe << "| #{xcpretty}"
+        pipe << "| #{xcpretty}"
+        pipe << " > /dev/null" if Snapshot.config[:suppress_xcode_output]
```
joshdholtz, line 37:
```suggestion
        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
```

😱 My bad! I didn't realize this pipe was going into an array. This array later on gets a `.join(' ')` which adds the space. Sorry about that!

snapshot/lib/snapshot/test_command_generator_xcode_8.rb
```diff
@@ -24,7 +24,10 @@ def generate(device_type: nil, language: nil, locale: nil)
 
       def pipe(device_type, language, locale)
         log_path = xcodebuild_log_path(device_type: device_type, language: language, locale: locale)
-        return ["| tee #{log_path.shellescape} | xcpretty #{Snapshot.config[:xcpretty_args]}"]
+        pipe = "| tee #{log_path.shellescape} "
+        pipe << "| xcpretty #{Snapshot.config[:xcpretty_args]}"
+        pipe << " > /dev/null" if Snapshot.config[:suppress_xcode_output]
+        return [pipe]
```
joshdholtz, line 30:
```suggestion
        pipe = ["| tee #{log_path.shellescape}"]
        pipe << "| xcpretty #{Snapshot.config[:xcpretty_args]}"
        pipe << "> /dev/null" if Snapshot.config[:suppress_xcode_output]
        return pipe
```

This one I think is what got me thinking weird about the other pipe 🤷‍♂️ This suggested changed makes the initial pipe an array instead of a string. The space in between will happen in the `.join(' ')` later on 😊

joshdholtz (changes requested):
Hey, sorry! I think I got some things confused on last review. I added two more suggestions should make this all gucci with spaces 😇

snapshot/lib/snapshot/test_command_generator_xcode_8.rb
rogerluan: Oooh sweet. This makes more sense! I tried keeping the original logic as much as possible but I see the difference now 😇

snapshot/lib/snapshot/test_command_generator.rb
rogerluan: Oh, I missed that. Thanks for catching it!

rogerluan:
Looks g2g now 😄

joshdholtz (approved):
This looks great!! Thank you so much for another awesome contribution ❤️ Really appreciate it!

Issue #16558, opened 2020-06-03
[scan] derivedDataPath getting added to xcodebuild command twice, causing error


TheMetalCode:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Once I updated to the latest release (`2.149.0`), I have noticed that `fastlane scan` is appending `-derivedDataPath` to the `xcodebuild` command twice, which causes it to fail.

This is the lane I use in my Fastfile to run scan:

```ruby
  lane :build_for_ci do
    scan(
      clean: false,
      scheme: $scheme,
      workspace: $workspace,
      configuration: "UnitTest",
      build_for_testing: true,
      buildlog_path: "build_logs",
      skip_build: true,
      devices: $devices
    )
  end
```
This only appears to be affecting `scan` - my `gym` actions are operating as expected.

##### Command executed

`bundle exec fastlane build_for_ci` (see Issue Description, this is a custom lane in my Fastfile that wraps the `scan` action)

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
    +--------------------------------------------+---------+-------------------------------------------------+
|                                              [32mUsed plugins[0m                                              |
+--------------------------------------------+---------+-------------------------------------------------+
| Plugin                                     | Version | Action                                          |
+--------------------------------------------+---------+-------------------------------------------------+
| fastlane-plugin-update_project_codesigning | 0.2.0   | update_project_codesigning                      |
| fastlane-plugin-trainer                    | 0.4.1   | trainer                                         |
| fastlane-plugin-firebase_app_distribution  | 0.1.4   | firebase_app_distribution                       |
| fastlane-plugin-bugsnag                    | 1.4.2   | upload_symbols_to_bugsnag send_build_to_bugsnag |
+--------------------------------------------+---------+-------------------------------------------------+

[12:00:28]: [32m----------------------------------------[0m
[12:00:28]: [32m--- Step: Verifying fastlane version ---[0m
[12:00:28]: [32m----------------------------------------[0m
[12:00:28]: Your fastlane version 2.149.0 matches the minimum requirement of 2.121.1  ✅
[12:00:28]: [32m------------------------------[0m
[...]

PR #16986, opened 2020-08-03
[gym][scan] xcodebuild -resolvePackageDependencies missing -derivedDataPath

nxtstep:
🔑 

Added `-derivedDataPath` parameter to xcodebuild when `derived_data_path` is set in options and or `SCAN_DERIVED_DATA_PATH` or `GYM_DERIVED_DATA_PATH` env are set.

### Checklist

### Motivation and Context
When adding SPM dependencies to your Xcode project the checkout repositories are cached within the `DerivedData` folder of your current project. However when building/testing your project with `fastlane gym/scan` the derived data path setting is not propagated to the `xcodebuild -resolvePackageDependencies` step. This can lead to confusion and bugs when adding/removing/updating said dependencies.

### Description
Added the `derivedDataPath` step to `Project` and skip setting/overwriting this parameter in `Scan` and `Gym`.

### Testing Steps
Just run a `fastlane scan [derived_data_path:your_path]` and/or `fastlane gym [derived_data_path:your_path]` on your project that **has** SPM dependencies (managed by Xcodeproj) and see that the `xcodebuild -resolvePackageDependencies` step has propagated the derivedDataPath when either `derived_data_path` is passed to gym/scan as an argument. Or respectively `SCAN_DERIVED_DATA_PATH` or `GYM_DERIVED_DATA_PATH` env are set.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -320,7 +320,7 @@ module FastlaneCore
       proj << "-scheme #{options[:scheme].shellescape}" if options[:scheme]
       proj << "-project #{options[:project].shellescape}" if options[:project]
       proj << "-configuration #{options[:configuration].shellescape}" if options[:configuration]
-      proj << "-derivedDataPath '#{options[:derived_data_path]}'" if options[:derived_data_path]
+      proj << "-derivedDataPath #{options[:derived_data_path].shellescape}" if options[:derived_data_path]
       proj << "-xcconfig #{options[:xcconfig].shellescape}" if options[:xcconfig]
 
       if FastlaneCore::Helper.xcode_at_least?('11.0') && options[:cloned_source_packages_path]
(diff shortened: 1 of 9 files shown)


joshdholtz (changes requested):
👋 Thanks for making this change! We had a change similar to this that had to get reverted because it ended up breaking some things so I want to make sure we covered all cases before merging this one 😇 

For reference:
- Initial PR - https://github.com/fastlane/fastlane/pull/16534/files
- Issue - https://github.com/fastlane/fastlane/issues/16558
- Revert PR - https://github.com/fastlane/fastlane/pull/16560

I think this PR is also missing a change to `snapshot` to prevent double `-derivedDataPath` 🤔 Can we get that changed added in too? 🙃

nxtstep:
@joshdholtz updated the PR as requested :bowtie:

joshdholtz (approved):
This looks good to me! Thanks for making those changes ❤️ I really appreciate the contribution!

bguidolim:
@nxtstep @joshdholtz I know that I arrived here a little bit late, but it is not the same as the parameter `cloned_source_packages_path`?

nxtstep:
No it’s not the same. However, using the cloned_source_packages_path parameter c/would solve [e.g. work around] the described problem as well I suppose. Nevertheless, I would still expect the `derived_data_path` to be considered in all `xcodebuild` commands when available.

Issue #17047, opened 2020-08-12
upload_to_app_store failing when uploading no screenshots


pyrtsa:
### New Regression Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Regression Information

- Breaking version: [e.g. `2.156.0`]
- Last working version: [e.g. `2.149.1`]

### Regression Description
In our setup we never had any screenshots to upload by Fastlane and `upload_to_app_store` worked fine. Having today updated to the latest Fastlane, uploads to App Store started failing with this error:

> Failed verification of all screenshots uploaded... 0 incomplete screenshot(s) still exist

This error is unexpected because we don't expect any screenshots to be uploaded so there's really nothing to verify about them either. (OTOH we never specified `skip_screenshots`.)

**As a workaround**, setting the environment variable `DELIVER_SKIP_SCREENSHOTS=true` made the upload pass.

The above error message was introduced by https://github.com/fastlane/fastlane/pull/16972, so I assume this to be a new regression in [`2.156.0`](https://github.com/fastlane/fastlane/releases/tag/2.156.0).

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
[13:55:26]: Will begin uploading snapshots for '3.2.1' on App Store Connect
[13:55:26]: Starting with the upload of screenshots...
[✔] Waiting for all the screenshots processed... 
[13:55:29]: Failed to upload all screenshots... Tries remaining: 4
[✔] Waiting for all the screenshots processed... 
[13:55:30]: Failed to upload all screenshots... Tries remaining: 3
[✔] Waiting for all the screenshots processed... 
[13:55:32]: Failed to upload all screenshots... Tries remaining: 2
[✔] Waiting for all the screenshots processed... 
[13:55:34]: Failed to upload all screenshots... Tries remaining: 1
[✔] Waiting for all the screenshots processed... 
...
[13:55:35]: Failed verification of all screenshots uploaded... 0 incomplete screenshot(s) still exist

+------+---------------------------------------------------------------+-------------+
|                                  fastlane summary                                  |
+------+---------------------------------------------------------------+-------------+
| Step | Action                                                        | Time (in s) |
[...]

PR #16927, opened 2020-07-24
Version bump to 2.153.0

joshdholtz:
Auto-generated by fastlane 🤖

**Changes since release '2.152.0':**

* [deliver] verify all screenshot deleted succesfully (#16842) via Josh Holtz
* [frameit] Add font_size parameter (#16870) via Andreas Ganske
* [action] download_dsyms - add option to pass `build_number` as an integer (#16900) via Alex Benoit
* [spaceship] add endpoint availableTerritories (#16877) via Bolo Michelin
* [fastlane] fix other_action to run in root project directory (#16889) via Jorge
* [scan] fix empty simulator logs (#16833) via Lyndsey Ferguson
* [spaceship] feat: support appStoreVersionReleaseRequests (#16765) via Valerio Castelli
* [action] erb uses ErbTemplateHelper and adds trim mode (#16882) via Josh Holtz
* [deliver] reimplementing tests from 2.150.0 (#16880) via Josh Holtz
* [tests] add tests for submit_for_review spec (#16876) via Josh Holtz
* [gems] update webmock, faraday (#16853) via Anton Rieder
* [deliver] fetches app with Spaceship::ConnectAPI (#16857) via Josh Holtz
* [gems] Update rspec, rspec_junit_formatter (#16841) via Anton Rieder

PR #17051, opened 2020-08-12
[deliver] fix upload_screenshots when not using upload_to_app_store for screenshots

ainame:
### Checklist

### Motivation and Context

Since this PR #16972, `deliver` has updated its logic to handle screenshot uploading and deleting. Some users reported a regression that it fails in case they don't use `deliver` for screenshot uploading but don't set `skip_screenshots: false`.  

https://github.com/fastlane/fastlane/issues/17047

I actually didn't take account of this case while working on #16927.

### Description

The error happens because of the logic to verify the completion of uploading by checking whether the number of screenshots to be uploaded matches the number of screenshots correctly uploaded on App Store Connect. 

In the reported case, the number of screenshots to be uploaded is always **0** but the number of screenshots correctly uploaded on App Store Connect can be as many as they have on it. So that comparison will never succeed.

Fortunately, we've got a given parameter `screenshots_per_language` to check screenshots to be uploaded. I could simply add `!screenshots_per_language.empty?`  to check if it's missing screenshot to upload.

https://github.com/fastlane/fastlane/blob/6413734cbbde7ba2d0d1961ed11fb27475b7bf6a/deliver/lib/deliver/upload_screenshots.rb#L179-L183

### Testing Steps

Added a unit test case. 

To reproduce the problem quickly, at first you need to set up a project that has no screenshots directory to be uploaded. (You can just delete the existing directory.) And then call `upload_to_app_store` with following options.

```ruby
upload_to_app_store(
  overwrite_screenshots: false,
  skip_binary_upload: true,
  skip_metadata: true,
)
```

### DIFF
--- a/deliver/lib/deliver/upload_screenshots.rb
+++ b/deliver/lib/deliver/upload_screenshots.rb
@@ -179,7 +179,8 @@ module Deliver
     # Verify all screenshots states on App Store Connect are okay
     def retry_upload_screenshots_if_needed(iterator, states, number_of_screenshots, tries, localizations, screenshots_per_language)
       is_failure = states.fetch("FAILED", 0) > 0
-      is_missing_screenshot = states.reduce(0) { |sum, (k, v)| sum + v } != number_of_screenshots
+      is_missing_screenshot = states.reduce(0) { |sum, (k, v)| sum + v } != number_of_screenshots && !screenshots_per_language.empty?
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
This makes sense to me! Thanks for fixing this 💪 So much ❤️

PR #17115, opened 2020-08-31
[action] add file ignore param for ensure_git_status_clean

devzhr:
### Checklist

### DIFF
--- a/fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
+++ b/fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
@@ -76,7 +76,6 @@ module Fastlane
                                        env_name: "FL_ENSURE_GIT_STATUS_CLEAN_IGNORED_FILE",
                                        description: "The flag whether to ignore file the git status if the repo is dirty",
                                        optional: true,
-                                       default_value: false,
                                        is_string: true)
         ]
       end
--- a/fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
+++ b/fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
@@ -77,7 +77,7 @@ module Fastlane
                                        description: "The flag whether to ignore file the git status if the repo is dirty",
                                        optional: true,
                                        default_value: false,
-                                       is_string: false)
+                                       is_string: true)
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
+++ b/fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
@@ -7,13 +7,13 @@ module Fastlane
     # Raises an exception and stop the lane execution if the repo is not in a clean state
     class EnsureGitStatusCleanAction < Action
       def self.run(params)
-        if params[:ignore_file] 
-          file_for_ignore = params[:ignore_file]
-          repo_status = Actions.sh("git status --porcelain --ignored #{file_for_ignore}")
+        if params[:ignored]
+          ignored_file = params[:ignored]
+          repo_status = Actions.sh("git status --porcelain --ignored #{ignored_file}")
         else
           repo_status = Actions.sh("git status --porcelain")
-        end  
-        
+        end
+
         repo_clean = repo_status.empty?
 
         if repo_clean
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
```diff
@@ -7,7 +7,13 @@ module SharedValues
     # Raises an exception and stop the lane execution if the repo is not in a clean state
     class EnsureGitStatusCleanAction < Action
       def self.run(params)
-        repo_status = Actions.sh("git status --porcelain")
+        if params[:ignore_file] 
+          file_for_ignore = params[:ignore_file]
```
joshdholtz, line 11: The key you are fetching from `params` needs to be defined in the `options` method below 😇 

Can you add the option in but also rename this option to `:ignored` so that it matches the CLI name? 😊

joshdholtz (changes requested):

fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
devzhr: I modified `ignore_file` to `ignored` and I added `params` from `options`

devzhr:
@googlebot I signed it!

fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
```diff
@@ -65,6 +71,12 @@ def self.available_options
                                        description: "The flag whether to show the git diff if the repo is dirty",
                                        optional: true,
                                        default_value: false,
+                                       is_string: false),
+          FastlaneCore::ConfigItem.new(key: :ignored,
+                                       env_name: "FL_ENSURE_GIT_STATUS_CLEAN_IGNORED_FILE",
+                                       description: "The flag whether to ignore file the git status if the repo is dirty",
+                                       optional: true,
+                                       default_value: false,
                                        is_string: false)
```
joshdholtz, line 80: `is_string` should be true, right? `--ignored` looks like its 3 strings values and not a boolean?

```sh
       --ignored[=<mode>]
           Show ignored files as well.
           The mode parameter is used to specify the handling of ignored files. It is optional: it defaults to traditional.
           The possible options are:
           o   traditional - Shows ignored files and directories, unless --untracked-files=all is specified, in which case individual files in ignored
               directories are displayed.
           o   no - Show no ignored files.
           o   matching - Shows ignored files and directories matching an ignore pattern.
           When matching mode is specified, paths that explicitly match an ignored pattern are shown. If a directory matches an ignore pattern, then it is
           shown, but not paths contained in the ignored directory. If a directory does not match an ignore pattern, but all contents are ignored, then the
           directory is not shown, but all contents are shown.
```

joshdholtz (changes requested):
This looks great! Just one small questions on the value `--ignored` takes

fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
devzhr: Yes `is_string` to `true`

fastlane/lib/fastlane/actions/ensure_git_status_clean.rb
```diff
@@ -65,7 +71,13 @@ def self.available_options
                                        description: "The flag whether to show the git diff if the repo is dirty",
                                        optional: true,
                                        default_value: false,
-                                       is_string: false)
+                                       is_string: false),
+          FastlaneCore::ConfigItem.new(key: :ignored,
+                                       env_name: "FL_ENSURE_GIT_STATUS_CLEAN_IGNORED_FILE",
+                                       description: "The flag whether to ignore file the git status if the repo is dirty",
+                                       optional: true,
+                                       default_value: false,
```
joshdholtz, line 79: Can we remove this `default_value`? I don’t think this needed since `optional: true` is set!

Once this is fixed we can get this merged! Thanks 😊

joshdholtz (changes requested):

joshdholtz (approved):
This looks good to me now! Thanks for making the requested changes ❤️  Really appreciate the contribution!

PR #16537, opened 2020-05-30
ci: Fix Homebrew + Ruby version

janpio:
With this, Homebrew should really use the correct Ruby version.

Fix #16536

PR #17142, opened 2020-09-03
[scan] fix default device list when using `skip_detect_devices`

johndelong:
This change fixes a bug where the `devices` list is not used when `skip_detect_devices` is `true`.

### Checklist

### Motivation and Context

For our UI tests, we want our tests to run on specific simulators (e.g. '["iPhone 6s", "iPad Air"]'). However, if Scan cannot find these simulators because they are not installed on the machine running the tests, we want the tests to fail as opposed to automatically finding another installed simulator. In the Scan docs, there is a configuration value called `skip_detect_devices` which specifically states:

> Should skip auto detecting of devices if none were specified

Therefore, setting this configuration to `true` should remove the ability to automatically detect another simulator and force Scan to use one of the simulators we specified in the `devices` list.

However, as it is currently written, if you set `skip_detect_devices` and also have a list of simulators to test with using the `devices` array, those simulators you specified will not be used and the execution of Scan will error with the following message:

```
xcodebuild: error: Failed to build workspace [Workspace] with scheme [Scheme].
	Reason: A build only device cannot be used to run this target.
	Recovery suggestion: No supported devices are available running a compatible version of iOS. Connect a device with a newer version of iOS in order to run your application (or choose a simulated device as the destination).
```

### Description

This change makes it so that `matches` are not thrown out if the `default` option is `null` by assigning `default` to an empty array when `skip_detect_devices` is `true`. I tested my proposed fix by running `scan` with the following command and successfully ran the tests with the following simulators:

```
  desc "Runs all the tests"
  lane :lane_test do
    scan(
      workspace: ENV["XCWORKSPACE"],
      scheme: ENV["SCHEME"],
      configuration: "Local",
      skip_build: true,
      skip_detect_devices: true,
      devices: ["iPhone 6s", "iPad Air"]
    )
  end
```

### Testing Steps

```
bundle exec fastlane lane_test
```

### DIFF
--- a/scan/lib/scan/detect_values.rb
+++ b/scan/lib/scan/detect_values.rb
@@ -187,15 +187,12 @@ module Scan
         end
       end
 
+      # Convert array to lazy enumerable (evaluate map only when needed)
       # grab the first unempty evaluated array
-      if default
-        Scan.devices = [matches, default].lazy.map { |x|
-          arr = x.call
-          arr unless arr.empty?
-        }.reject(&:nil?).first
-      else
-        Scan.devices = []
-      end
+      Scan.devices = [matches, default].lazy.reject(&:nil?).map { |x|
+        arr = x.call
+        arr unless arr.empty?
+      }.reject(&:nil?).first
     end
 
     def self.min_xcode8?


johndelong:
@googlebot I signed it!

johndelong:
@googlebot I fixed it.

johndelong:
Looks like the build error might be related to this?
https://github.com/fastlane/fastlane/pull/16537

scan/lib/scan/detect_values.rb
```diff
@@ -187,15 +187,16 @@ def self.detect_simulator(devices, requested_os_type, deployment_target_key, def
...
-      else
-        Scan.devices = []
+      # default to an empty array if `skip_detect_devices` is `true`
+      unless default
+        default = []
       end
+
+      # grab the first unempty evaluated array
+      Scan.devices = [matches, default].lazy.map { |x|
+        arr = x.call
```
joshdholtz, line 197: I tried running with this an unknown simulator name and `skip_detect_devices` set to `true` and got 👇 

```sh
[!] undefined method `call' for []:Array
```

🤔 Did you see this at all?

joshdholtz (commented):
@johndelong Thanks for looking into this! This looks like the right solution but I got an error when running it (I think I set things up correctly)... See my other comment for the error

scan/lib/scan/detect_values.rb
johndelong: Interesting, I did not think to try to run this with an empty list for `matches` AND an empty `default` array. Seems like `call` isn't something that can be invoked on an empty array.

I'll admit, I don't know much about Ruby, but after some googling I have some ideas to try. I'll fiddle with it tonight and see if I can come up with a better solution while testing with your additional scenarios.

Thanks for testing!

johndelong: Ok, this issue should be addressed now. If there is a more eloquent way of writing my solution, please let me know and i'll make the necessary updates. Thank you!

joshdholtz: Is an invalid name now what you were testing it with? 😇 What simulator name(s) were you using?

joshdholtz:
```rb
# Convert array to lazy enumerable (evaluate map only when needed) 
# Grabs the first unempty evaluated array
Scan.devices = [matches, default].lazy.reject(&:nil?).map { |x|
  arr = x.call
  arr unless arr.empty?
}.reject(&:nil?).first
```

@johndelong I took me some time to understand this code part of the code base (I haven't done anything in here before surprisingly 😛 ). But I _think_ this is the solution we want here ☝️ We will just get rid of the if statement from the original code and add `reject(&:nil?)` before `map`. Thoughts? 😊

joshdholtz (changes requested):

scan/lib/scan/detect_values.rb
johndelong: @joshdholtz Ahhh, I _knew_ there had to be a better way to write this. Very clean! ✨

I tested with an invalid simulator (`iPhone 6sz`), with a valid simulator (`iPhone 6s`), and with `skip_detect_devices` enabled and disabled. This solution checks out ✅ Thank you very much for your review and help. I appreciate it!

TIL about Ruby `.reject(&:nil?)` 😉

joshdholtz: Glad it works! This was fun to work on 😊 Excited to get this released for you!

joshdholtz (approved):
OMG I thought this got merged in 😱 Sorry about that! Thanks again for this ❤️

PR #17169, opened 2020-09-05
[deliver] fix UploadMetadata taking too long to run

pNre:
### Checklist

### Motivation and Context
Updating our lanes to use App Store Connect API, deliver went from running in seconds to over 10 minutes.

I've tracked down the issue to the fact that we have >300 releases and `get_app_store_versions` trying to fetch all of them in pages of 2.

#### Issue
To check whether a release is the first one for an app, `UploadMetadata` calls `Spaceship ::ConnectAPI::App.get_app_store_versions` with `limit: 2`.
[get_app_store_versions](https://github.com/pNre/fastlane/blob/2e122c59d8abd9374c92c2a4e3640fe406ea4bff/spaceship/lib/spaceship/connect_api/models/app.rb#L243), in turn takes the result of `Spaceship::ConnectAPI.get_app_store_versions` and then calls `all_pages` on it.
But `all_pages` paginates the response until there are no more pages. In our case, 300 releases and limit 2 means 150 pages/requests, which leads to deliver taking forever to complete.

#### Solution
I went head and tried to implement a workaround that, instead of first requesting all pages then counting, uses the `meta.paging.total` value to enumerate the versions an app has.

Please let me know if 

### Testing Steps
Run deliver to update metadata for an app with hundreds of releases.

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -111,7 +111,7 @@ module Deliver
       end
 
       # Needed for to filter out release notes from being sent up
-      number_of_versions = app.get_app_store_versions(filter: { platform: platform }, limit: 2).size
+      number_of_versions = app.get_app_store_versions_count(filter: { platform: platform }, limit: 2)
       is_first_version = number_of_versions == 1
       UI.verbose("Version '#{version.version_string}' is the first version on App Store Connect") if is_first_version
 
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -114,7 +114,6 @@ module Deliver
       number_of_versions = Spaceship::ConnectAPI.get_app_store_versions(
         app_id: app.id,
         filter: { platform: platform },
-        includes: Spaceship::ConnectAPI::AppStoreVersion::ESSENTIAL_INCLUDES,
         limit: 2
       ).count
       is_first_version = number_of_versions == 1
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -111,7 +111,12 @@ module Deliver
       end
 
       # Needed for to filter out release notes from being sent up
-      number_of_versions = app.get_app_store_versions_count(filter: { platform: platform }, limit: 2)
+      number_of_versions = Spaceship::ConnectAPI.get_app_store_versions(
+        app_id: app.id,
+        filter: { platform: platform },
+        includes: Spaceship::ConnectAPI::AppStoreVersion::ESSENTIAL_INCLUDES,
+        limit: 2
+      ).count
       is_first_version = number_of_versions == 1
       UI.verbose("Version '#{version.version_string}' is the first version on App Store Connect") if is_first_version
 
--- a/spaceship/lib/spaceship/connect_api/models/app.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app.rb
@@ -244,6 +244,9 @@ module Spaceship
         return resps.flat_map(&:to_models)
       end
 
+      def get_app_store_versions_count(filter: {}, includes: Spaceship::ConnectAPI::AppStoreVersion::ESSENTIAL_INCLUDES, limit: nil, sort: nil)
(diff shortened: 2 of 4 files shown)


rogerluan (approved):
💯 💯 💯  Good stuff!

spaceship/lib/spaceship/connect_api/models/app.rb
```diff
@@ -244,6 +244,14 @@ def get_app_store_versions(filter: {}, includes: Spaceship::ConnectAPI::AppStore
...
       end
 
+      def get_app_store_versions_count(filter: {}, limit: nil)
+        return Spaceship::ConnectAPI.get_app_store_versions(
+          app_id: id,
+          filter: filter,
+          includes: Spaceship::ConnectAPI::AppStoreVersion::ESSENTIAL_INCLUDES,
+          limit: limit
+        ).paged_resources_count
+      end
```
joshdholtz, line 254: @pNre This is really great find! One quick question. Instead of creating a method for this, can we instead just not call `.all_pages` if `limit` isn't `nil` in `get_app_store_versions`? I think that might make things a little cleaner and make that method behave as expect with limit.

joshdholtz (changes requested):
This is such a great find! Just a small thought/request on how to change existing method 😇

spaceship/lib/spaceship/connect_api/models/app.rb
pNre: I think it depends on how limit works and how long pages are.
For example, if `limit = 30` and pages have at most `20` items, don't I still need to paginate to get the last 10 items?

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -111,7 +111,7 @@ def upload(options)
       end
 
       # Needed for to filter out release notes from being sent up
-      number_of_versions = app.get_app_store_versions(filter: { platform: platform }, limit: 2).size
+      number_of_versions = app.get_app_store_versions_count(filter: { platform: platform }, limit: 2)
```
joshdholtz, line 114:
```suggestion
      number_of_versions  = Spaceship::ConnectAPI.get_app_store_versions(app_id: app.id, filter: { platform: platform }, limit: 2).count
```

What if we change to this? 😇

spaceship/lib/spaceship/connect_api/models/app.rb
joshdholtz: Hmm... good point! What about my next suggestion above? 😇  This seems pretty specific for `deliver` so I think I'd like avoid adding a method for count in `Spaceship::ConnectAPI::App` 🤔

deliver/lib/deliver/upload_metadata.rb
pNre: 👍

spaceship/lib/spaceship/connect_api/models/app.rb
pNre: @joshdholtz change implemented

joshdholtz: Thank you so much! I'll get this merged and out the door later today or tomorrow ❤️

rogerluan (commented):
The refactor of the last commit looks good, but IMO having it in Spasceship as a utility made more sense 😬 not necessarily the only use case will be in deliver. Right now it may be, but there may be other Spaceship consumers that could make good use of this performant utility. But it's more of a design decision 😇 @joshdholtz

joshdholtz:
@googlebot I consent.

joshdholtz:
@rogerluan I totally appreciate the thought 😊 I _doooooooo_ agree that this might get used by other somewhere else 🤔 I'm also rethinking how the method should call `all_pages` by default. I think that was a bad design decision by me back in the day. A method that use `all_pages` shouldn't necessarily allow `limit` since it doesn't really do what its supposed to be doing. I think I almost want to rethink some of that from the ground up 😭

rogerluan:
@joshdholtz Sounds like a problem for future Josh 😂🙈

joshdholtz:
I know 😩😩😩😩😩😩😩

joshdholtz (approved):
This is 🔥 Thank you!! ❤️  Really appreciate the changes!

rogerluan:
Hey, I just realized how faster my submissions got!! Thanks again for this PR ❤️❤️

Issue #15476, opened 2019-10-11
Frameit produces a portrait image for a landscape screenshot


mrgrauel:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
* Frameit produces a portrait screenshot for a landscape image
  * Framed: [iPad Pro (12 9-inch) (2nd generation)-1_stream_framed](https://user-images.githubusercontent.com/148758/66648497-fa623600-ec2b-11e9-97f9-2d1cb4bf47a5.png)
  * Screenshot: [iPad Pro (12 9-inch) (2nd generation)-1_stream](https://user-images.githubusercontent.com/148758/66648539-0b12ac00-ec2c-11e9-9ba8-24867cfbd747.png)

##### Command executed
`frameit(path: lane_context[SharedValues::SNAPSHOT_SCREENSHOTS_PATH], white: true)`

No error message just the wrong result

### Environment

<details>
  <pre>
 <details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.15                                       |
| Ruby                        | 2.6.5                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.23.0                          |
| Installation Source         | /usr/local/lib/ruby/gems/2.6.0/bin/fastlane |
| Host                        | Mac OS X 10.15 (19A583)                     |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/2.6.5/lib            |
| OpenSSL Version             | OpenSSL 1.1.1d  10 Sep 2019                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 11.1                                        |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   | en_US.UTF-8 | ✅ |
| LANGUAGE | en_US.UTF-8 | ✅ |

### fastlane files:

[...]

PR #17180, opened 2020-09-07
[snapshot] fix snapshot landscape orientation issue

jvk75:
### Checklist

### Motivation and Context
Snapshot images do not respect landscape orientation set by the UI test with ```XCUIDevice.shared.orientation```. This caused images to be always portrait even if landscape was set. Issue arises for real when used with frameit. There is frameit issue #15476 that explains more (auto closed but still an issue).
Also snapshot preview page has been different in safari and chrome. Safari reads orientation from png metadata and displays correct but chrome skips it and landscape is shown in portrait form.

### Description
This fix will use the UIImage of screenshot and does redrawing of the image in landscape case. Redraw resets the images orientation to be correct in landscape case. With portrait orientation the screenshot image is used as is (metadata and image orientation are the same a.k.a no issue).

### Testing Steps
Change ```XCUIDevice.shared.orientation``` between landscape(left/right) and portrait in your UITest to see that the generated images are respecting the correct orientation. Also the generated webpage now works same in safari and chrome.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -165,6 +165,8 @@ open class Snapshot: NSObject {
             }
 
             let screenshot = XCUIScreen.main.screenshot()
+            let image = XCUIDevice.shared.orientation.isLandscape ?  fixLandscapeOrientation(image: screenshot.image) : screenshot.image
+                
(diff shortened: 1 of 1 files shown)


jvk75:
@googlebot I signed it!

joshdholtz (approved):
This is great! Thanks so much for fixing this 😊 Really appreciate the contribution!

Issue #17003, opened 2020-08-06
can't match appstore provisioning profile from match import


Tobrek:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
An AppStore MobileProvision profile, imported with fastlane match import can't be found by fastlane match, because it is stored as Distribution_com.company.mobileprovision and search as AppStore_com.company.

[12:50:50]: No matching provisioning profiles found for 'AppStore_com.company'

##### Command executed
fastlane match import --git_branch company -u xxx -a com.company -b XXX --type appstore

execute as action
match(
		app_identifier: com.company,
		type: "appstore",
		username: xxx
		team_id: XXX
		git_branch: company
	)

##### Complete output when running fastlane, including the stack trace and command used
[me]$ fastlane match import --git_branch company -u xxx -a com.company -b XXX --type appstore
[✔] 🚀 
[11:15:31]: Successfully loaded '/.../myproject/fastlane/Matchfile' 📄                                                 

+-----------------------+-----------------------------------------------------------+
|                    Detected Values from './fastlane/Matchfile'                    |
+-----------------------+-----------------------------------------------------------+
| git_url               | git@.../ios-certificates.git                                       |
| git_branch            | baseCompany
| type                  | enterprise                                                |
| app_identifier        | ["com.baseComanpy.one", "com.baaseComany.two"] |
| readonly              | true                                                      |
| clone_branch_directly | true                                                      |
+-----------------------+-----------------------------------------------------------+

[11:15:31]: Certificate (.cer) path:
/.../ios_distribution.cer                                        
[11:15:43]: Private key (.p12) path:
/.../ios.p12                                        
[11:15:48]: Provisioning profile (.mobileprovision or .provisionprofile) path or leave empty to skip this file:
/.../ios.mobileprovision                                              
[11:15:53]: Cloning remote git repo...
[11:15:55]: Checking out branch company...
[11:15:55]: 🔓  Successfully decrypted certificates repo
[11:15:55]: Repo is at: '/var/folders/7x/1clhqpgj54vcbg9s6k65ysmm0000gn/T/d20200806-42450-wrp2my'
[...]

PR #17213, opened 2020-09-10
[match] fix match import wrong profile names

DunbarXX:
### Checklist

### Motivation and Context

We were trying to import our existing adhoc and development profiles using **match import** command and I noticed that the file and folder formats were different in some cases.

There is a bug ticket here: https://github.com/fastlane/fastlane/issues/17003

### Description

I changed the **importer.rb** file to use the correct method for retrieving the provisioning profile prefix.

### Testing Steps

Tested with **development, appstore, adhoc and inhouse** profiles for **iOS**. We do not have **developer_id** profiles and we couldn't test that.

As noted above in the checklist I ran all the tests and rubocop. Tests failed for the importing appstore provisioning profiles and I saw that they were wrong and corrected the tests.

Please note that I'm not a ruby programmer and I don't know if the changes that I made are OK.

### DIFF
--- a/match/lib/match/importer.rb
+++ b/match/lib/match/importer.rb
@@ -2,6 +2,7 @@ require_relative 'spaceship_ensure'
 require_relative 'encryption'
 require_relative 'storage'
 require_relative 'module'
+require_relative 'generator'
 require 'fastlane_core/provisioning_profile'
 require 'fileutils'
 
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
Ahhhhh, great find! Really appreciate the fix 😊  Thanks for the contribution! ❤️

PR #17218, opened 2020-09-10
Modify `Net::HTTPGenericRequest` monkeypatch so that only Apple domains are affected

atyndall:
### Checklist

### Motivation and Context
Commit eb8d59959fe80d237bbe548ae413eb5803b08665 introduced a monkeypatch of `Net::HTTPGenericRequest#supply_default_content_type` in `lib/spaceship/helper/net_http_generic_request.rb` that resolved an issue whereby a Content-Type header was being automatically added to a HTTP request that deliberately lacked one, causing the request to fail. The patch modifies `Net::HTTP`'s behaviour so that this automatic header is not added.

This patch however is very broad, and affects `Net::HTTP`'s behaviour globally. In our case, we added fastlane to an internal project as a gem to utilise it programatically, and noted that unrelated areas of the project began to inexplicably fail as fastlane's monkeypatch changed this behaviour for all HTTP requests in the project.

### Description
Based on the description of what the patch is intending to fix, it seems that its scope can be significantly limited, thus making the patch play nice with other code that relies on the default behaviour.

This modification ensures that the skipping of the Content-Type header only occurs for Apple subdomains, which covers the patch's intended targets, while preventing projects that include fastlane from suffering from strange unrelated failures when they include the gem.

### Testing Steps
Automated testing suite has run. I have also verified that POST requests using the `Spaceship::Client` methods trigger the monkeypatch's conditions. Internal tests in our codebase verify the monkeypatch no longer affects the methods in the project that were broken by it.

Looking at the hostnames the commit affects, it seems to be "developer.apple.com", "idmsa.apple.com" and "developerservices2.apple.com", all of which are covered by the condition.

### DIFF
--- a/spaceship/lib/spaceship/helper/net_http_generic_request.rb
+++ b/spaceship/lib/spaceship/helper/net_http_generic_request.rb
@@ -5,10 +5,16 @@ require 'net/http'
 # Certain apple endpoints return 415 responses if a Content-Type is supplied.
 # Net::HTTP will default a content-type if none is provided by faraday
 # This monkey-patch allows us to leave out the content-type if we do not specify one.
-module Net
-  class HTTPGenericRequest
-    def supply_default_content_type
-      return if content_type
-    end
+module NetHTTPGenericRequestMonkeypatch
+  def supply_default_content_type
+    # Return no content type if we communicating with an apple.com domain
+    return if !self['host'].nil? && self['host'].end_with?('.apple.com')
+
+    # Otherwise use the default implementation
+    super
   end
 end
+
+# We prepend the monkeypatch so the patch has access to the original implementation
+# using `super`.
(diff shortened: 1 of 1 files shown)


atyndall:
@googlebot I signed it!

atyndall:
@googlebot I signed it!

atyndall:
Looks like it's going to take a few days for our corporate CLA to get approved by Google, but it should hopefully it sorted soon :) The PR is ready for review

snatchev (approved):
Thank you for this fix! I completely forgot about this, and 5 years ago I would have never imagined that this monkeypatch would end up in a library, included in other people's code 😅 

LGTM

atyndall:
@googlebot I signed it!

atyndall:
@snatchev Thanks for your approval, our CLA has now been processed so this should be good to go now

PR #17227, opened 2020-09-13
Dysm download date filter

mwoollard:
### Checklist
Note that on my system the keychain tests fail, but this occurs before PR:
```
Fastlane
  Fastlane::FastFile
    Create keychain Integration
      with name and password options
        works when keychain doesn't exist
        work when keychain already exist
        works with a password that contain spaces or `"` (FAILED - 1)
        works with keychain-settings (FAILED - 2)
        works with default_keychain (FAILED - 3)
        works with unlock (FAILED - 4)
        works with all params (FAILED - 5)
        sets the correct keychain path
      with path and password options
        successfully creates the keychain
        sets the correct keychain path
```

Note that there is one warning:

```
fastlane/lib/fastlane/actions/download_dsyms.rb:8:7: C: Perceived complexity for run is too high. [27/18]
      def self.run(params)
      ^^^
```
This was present prior to this PR

### Motivation and Context
Currently download_dsyms can fetch all dsym files, or specific files based on app version, build, live or latest. For some situations when an app may have live version, and multiple versions in development / TestFlight, these options don't provide a means to ensure all dsyms are downloaded without requesting all files every time the command is run.

By adding an option to download files newer than a given uploaded_date, and providing the most recent date of files downloaded during a pass, a controlling script can track the date and request only newer files each time.

### Description
The PR adds a new command option `after_uploaded_date` which takes a date / time string in any format supported by the ruby date parser. When provided only downloads dsym files for builds with an uploaded date more recent than the option.

The most recent uploaded date of the builds that symbols were downloaded for is added to the SharedValues hash as `DSYM_LATEST_UPLOADED_DATE`. 

Additionally the `DSYM_LATEST_UPLOADED_DATE` `SharedValue` is cleared by `clean_build_artifacts`. Although this is not a build artifact it is logical that it would be reset along with `DSYM_FILES`.

### Testing Steps
Have added additional unit test to download_dsyms_specs to validate the correct download calls are made:

```
$ bundle exec rspec --pattern "**{,/*/**}/download_dsyms_spec.rb"
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
Gem::Specification#rubyforge_project= called from /Users/markw/.gem/specifications/open4-1.3.4.gemspec:16.
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
Changing stdout to /var/folders/6w/4yzk82317zbc8jwwdr9cn2xw0000gp/T/fastlane_tests, set `DEBUG` environment variable to print to stdout (e.g. when using `pry`)

Fastlane
  Fastlane::FastFile
    download_dsyms
      with no special options
        downloads all dsyms of all builds in all trains
      with version with leading zero
        downloads all dsyms of all builds in train 1.07.0
      when build_number is an integer
        downloads the correct dsyms
      when version is latest
        downloads only dsyms of latest build in latest train
      when version is live
        downloads only dsyms of live build
      when min_version is set
        downloads only dsyms of trains newer than or equal min_version
      with after_uploaded_date
        downloads dsyms with more recent uploaded_date

Finished in 0.12012 seconds (files took 3.53 seconds to load)
7 examples, 0 failures
```

Have run on our developer accounts and applications to check it correctly selects the expected dsyms when the new option is provided, and that the `DSYM_LATEST_UPLOADED_DATE` is correct.

Finally, this is my first dive into ruby + fastlane codebase so all feedback appreciated !

### DIFF
--- a/fastlane/lib/fastlane/actions/clean_build_artifacts.rb
+++ b/fastlane/lib/fastlane/actions/clean_build_artifacts.rb
@@ -23,6 +23,7 @@ module Fastlane
 
         Actions.lane_context[Actions::SharedValues::SIGH_PROFILE_PATHS] = nil
         Actions.lane_context[Actions::SharedValues::DSYM_PATHS] = nil
+        Actions.lane_context[Actions::SharedValues::DSYM_LATEST_UPLOADED_DATE] = nil
 
         UI.success('Cleaned up build artifacts 🐙')
       end
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -2,13 +2,14 @@ module Fastlane
   module Actions
     module SharedValues
       DSYM_PATHS = :DSYM_PATHS
+      DSYM_LATEST_UPLOADED_DATE = :DSYM_LATEST_UPLOADED_DATE
     end
     class DownloadDsymsAction < Action
-      # rubocop:disable Metrics/PerceivedComplexity
       def self.run(params)
         require 'openssl'
         require 'spaceship'
         require 'net/http'
+        require 'date'
 
         # Team selection passed though FASTLANE_ITC_TEAM_ID and FASTLANE_ITC_TEAM_NAME environment variables
         # Prompts select team if multiple teams and none specified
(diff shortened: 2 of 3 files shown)


mwoollard:
@googlebot I signed it!

mwoollard:
@googlebot I signed it!

mwoollard:
Need to re-submit from different GitHub user

PR #17280, opened 2020-09-20
Add and delete multiple testers API at a time

csc-EricWu:
Add API  delete_beta_testers_from_app


csc-EricWu:
@googlebot I signed it!

PR #17302, opened 2020-09-23
[deliver] transform string key to symbol, since hash args in options from CLI keyed by string

LancerTZ:
transform string key to symbol, since hash args in options from CLI keyed by string

### Checklist

### Motivation and Context
`app_review_information` passed into deliver from CLI is parsed to hash keyed by string but not symbol. This makes `set_review_information` failed.

### Description
I'm not very familiar with Ruby and I haven't much time, so my solution looks a bit ugly.
Better way to fix this problem is to make sure hash args in options keyed by symbol when generate options, hope maintainers can help me.

### Testing Steps
I use command below to test
```
DELIVER_SUBMISSION_INFO='{"add_id_info_limits_tracking":true,"add_id_info_serves_ads":false,"add_id_info_tracks_action":false,"add_id_info_tracks_install":true,"add_id_info_uses_idfa":true}'

DELIVER_APP_REVIEW_INFO='{"demo_user":"xxxxx","demo_password":"xxxxx"}'

fastlane deliver submit_build --username 'email' --team_id 000000 --app_identifier 'com.xxxx.xxxx' --platform 'ios' --app_version $app_version --build_number $build_number --skip_binary_upload true --skip_screenshots true --skip_app_version_update true --automatic_release false --phased_release true --reset_ratings false --submission_information "$DELIVER_SUBMISSION_INFO" --app_review_information "$DELIVER_APP_REVIEW_INFO" --reject_if_possible true --force true
```

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -583,7 +583,7 @@ module Deliver
 
     def set_review_information(version, options)
       return unless options[:app_review_information]
-      info = options[:app_review_information]
+      info = options[:app_review_information].transform_keys(&:to_sym)
       UI.user_error!("`app_review_information` must be a hash", show_github_issues: true) unless info.kind_of?(Hash)
 
       attributes = {}
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -583,7 +583,8 @@ module Deliver
 
     def set_review_information(version, options)
       return unless options[:app_review_information]
-      info = options[:app_review_information].transform_keys(&:to_sym)
+      info = options[:app_review_information]
+      info = info.collect { |k, v| [k.to_sym, v] }.to_h
       UI.user_error!("`app_review_information` must be a hash", show_github_issues: true) unless info.kind_of?(Hash)
 
       attributes = {}


TragedyStar:
> @googlebot I signed it!

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -583,7 +583,7 @@ def normalize_language_keys(options)
 
     def set_review_information(version, options)
       return unless options[:app_review_information]
-      info = options[:app_review_information]
+      info = options[:app_review_information].transform_keys(&:to_sym)
```
joshdholtz, line 586: `transform_keys` is only available in Ruby 2.5 and up so we need to do something a little bit more manual here 😬 

```suggestion
      info = options[:app_review_information]
      info = info.collect { |k, v| [k.to_sym, v] }.to_h
```

joshdholtz (changes requested):
Good call on doing this! Can't use `transform_keys` thought because only available in Ruby 2.5 🤷‍♂️

TragedyStar:
> Good call on doing this! Can't use `transform_keys` thought because only available in Ruby 2.5 🤷‍♂️

Thanks for changes.
As I said above:
```
I'm not very familiar with Ruby and I haven't much time, so my solution looks a bit ugly.
Better way to fix this problem is to make sure hash args in options keyed by symbol when generate options, hope maintainers can help me.
```
Since options[:app_review_information] is still keyed by string and some other places using it may still meet this bug, could someone help to fix this completely and more elegantly?

deliver/lib/deliver/upload_metadata.rb
TragedyStar: Thanks for changes!

joshdholtz (approved):
Thanks for making the requested changes! This looks great. I really appreciate the contribution ❤

PR #17304, opened 2020-09-23
[spaceship] add ability to delete multiple beta testers at a time

csc-EricWu:
### Checklist

### DIFF
--- a/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
+++ b/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
@@ -277,6 +277,19 @@ module Spaceship
           test_flight_request_client.delete("betaTesters/#{beta_tester_id}/relationships/betaGroups", nil, body)
         end
 
+        def delete_beta_testers_from_app(beta_tester_ids: [], app_id: nil)
+          body = {
+            data: beta_tester_ids.map do |id|
+              {
+                type: "betaTesters",
+                id: id
+              }
+            end
+          }
+
+          test_flight_request_client.delete("apps/#{app_id}/relationships/betaTesters", nil, body)
+        end
+
         #
         # betaTesterMetrics
         #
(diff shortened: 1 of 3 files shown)


csc-EricWu:
Please provide exact steps to test this. So far I can only see how to delete testers. Also we cannot add PRs that are not matching our code style.

@max-ott I changed to the same style as you. I provide an API for deleting multiple testers at one time. You should know that you need an account to have a case.

csc-EricWu:
@joshdholtz  Can you have a look? I need this API please

joshdholtz:
I have some family things I need to do this weekend but I will look on Monday!

csc-EricWu:
ok, Thank you for your reply.

spaceship/lib/spaceship/connect_api/testflight/testflight.rb
```diff
@@ -277,6 +277,19 @@ def delete_beta_tester_from_beta_groups(beta_tester_id: nil, beta_group_ids: [])
           test_flight_request_client.delete("betaTesters/#{beta_tester_id}/relationships/betaGroups", nil, body)
         end
 
+        def delete_beta_testers_from_app(beta_tester_ids: [], app_id: nil)
```
joshdholtz, line 280: The title of the PR is "add and delete" but this is named `delete_beta_testers_from_app` 🤔 Should this method actually be called something like `update_beta_testers_from_app`?

joshdholtz (changes requested):
Thanks for making this PR! Just one change request on the name of the new method (I think) 😇 Let me know your thoughts!

spaceship/lib/spaceship/connect_api/testflight/testflight.rb
csc-EricWu: It's just that you can delete more than one tester at a time, which is very useful to me. My English is not very good. I'm sorry.

csc-EricWu: Delete multiple testers, which greatly improves efficiency.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks so much for adding this! This should get released later this week. I really appreciate the contribution ❤️

PR #17315, opened 2020-09-24
[spaceship] support for adding & removing users

max-ott:
### Checklist

### Motivation and Context
Own branch with new model: ref: https://github.com/fastlane/fastlane/pull/17305


max-ott:
Added the model. Not sure how to do the `attributes: []` stuff. I'm not finding it very approachable. This way one can directly discover, what to use and how to implement it.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
@max-ott Tested this out and added a few commits! This looks great though! Thanks for adding this ❤️ 

I doooooo agree with putting attribute names in the method so I might start migrating that on other calls at some point 😊

PR #17316, opened 2020-09-24
Clarify Sigh Arg Conflicts Error Messages

mylesbarros:
### Checklist

### Description
Sigh propagates an error if the user passes in two arguments that sigh
perceives as being mutually exclusive. For example, if the user passes
both an `:adhoc` and a `:development` arg, it will fail.

The error message that it currently propagates however states that "You can't enable both
:development and :adhoc".

This is problematic because the word "enable" suggests that the two
boolean arguments can't both be `true` when in reality sigh will still
error if given values of `false` and `false` or `false` and `true`.

This commit thus removes the word "enable" and clarifies that the problem is
caused by mutually exclusive arguments.

### Testing Steps
`bundle exec fastlane run sigh adhoc:true development:false`

### DIFF
--- a/sigh/lib/sigh/options.rb
+++ b/sigh/lib/sigh/options.rb
@@ -16,7 +16,7 @@ module Sigh
                                      default_value: false,
                                      conflicting_options: [:developer_id, :development],
                                      conflict_block: proc do |option|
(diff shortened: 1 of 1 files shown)


mylesbarros:
@googlebot I signed it!

mylesbarros:
@googlebot I fixed it

PR #17336, opened 2020-09-29
[Fastlane.Swift] fix: beforeAll method not called

jhandguy:
### Checklist

### Motivation and Context

The method `beforeAll` is not called since the last update (`2.161.0`).

### Description

Calling both `beforeAll` and `afterAll` methods via selector solved the issue.

### Testing Steps

Tested via Xcode Project, all works as expected now.

### DIFF
--- a/fastlane/swift/LaneFileProtocol.swift
+++ b/fastlane/swift/LaneFileProtocol.swift
@@ -12,18 +12,18 @@ import Foundation
 
 public protocol LaneFileProtocol: class {
     var fastlaneVersion: String { get }
-    static func runLane(from fastfile: LaneFile?, named: String, parameters: [String: String]) -> Bool
+    static func runLane(from fastfile: LaneFile?, named lane: String, with parameters: [String: String]) -> Bool
 
     func recordLaneDescriptions()
-    func beforeAll(currentLane _: String)
-    func afterAll(currentLane: String)
+    func beforeAll(with lane: String)
+    func afterAll(with lane: String)
     func onError(currentLane: String, errorInfo: String)
 }
 
 public extension LaneFileProtocol {
     var fastlaneVersion: String { return "" } // Defaults to "" because that means any is fine
-    func beforeAll(currentLane _: String) {} // No-op by default
-    func afterAll(currentLane _: String) {} // No-op by default
+    func beforeAll(with lane: String) {} // No-op by default
+    func afterAll(with lane: String) {} // No-op by default
     func onError(currentLane _: String, errorInfo _: String) {} // No-op by default
     func recordLaneDescriptions() {} // No-op by default
 }
@@ -90,8 +90,8 @@ open class LaneFile: NSObject, LaneFileProtocol {
         }
     }
 
-    public static func runLane(from fastfile: LaneFile?, named: String, parameters: [String: String]) -> Bool {
-        log(message: "Running lane: \(named)")
+    public static func runLane(from fastfile: LaneFile?, named lane: String, with parameters: [String: String]) -> Bool {
+        log(message: "Running lane: \(lane)")
         #if !SWIFT_PACKAGE
             // When not in SPM environment, we load the Fastfile from its `className()`.
             loadFastfile()
@@ -111,7 +111,7 @@ open class LaneFile: NSObject, LaneFileProtocol {
             self.fastfileInstance = fastfileInstance
         #endif
         let currentLanes = lanes
-        let lowerCasedLaneRequested = named.lowercased()
+        let lowerCasedLaneRequested = lane.lowercased()
 
         guard let laneMethod = currentLanes[lowerCasedLaneRequested] else {
             let laneNames = laneFunctionNames.map { laneFuctionName in
@@ -122,7 +122,7 @@ open class LaneFile: NSObject, LaneFileProtocol {
                 }
             }.joined(separator: ", ")
 
-            let message = "[!] Could not find lane '\(named)'. Available lanes: \(laneNames)"
+            let message = "[!] Could not find lane '\(lane)'. Available lanes: \(laneNames)"
             log(message: message)
 
             let shutdownCommand = ControlCommand(commandType: .cancel(cancelReason: .clientError), message: message)
@@ -131,15 +131,15 @@ open class LaneFile: NSObject, LaneFileProtocol {
         }
 
         // Call all methods that need to be called before we start calling lanes.
-        fastfileInstance.beforeAll(currentLane: named)
+        fastfileInstance.beforeAll(with: lane)
 
         // We need to catch all possible errors here and display a nice message.
         _ = fastfileInstance.perform(NSSelectorFromString(laneMethod), with: parameters)
 
         // Call only on success.
-        fastfileInstance.afterAll(currentLane: named)
+        fastfileInstance.afterAll(with: lane)
 
-        log(message: "Done running lane: \(named) 🚀")
+        log(message: "Done running lane: \(lane) 🚀")
         return true
     }
 }
--- a/fastlane/swift/LaneFileProtocol.swift
+++ b/fastlane/swift/LaneFileProtocol.swift
@@ -15,14 +15,14 @@ public protocol LaneFileProtocol: class {
     static func runLane(from fastfile: LaneFile?, named: String, parameters: [String: String]) -> Bool
 
     func recordLaneDescriptions()
-    func beforeAll()
+    func beforeAll(currentLane _: String)
     func afterAll(currentLane: String)
     func onError(currentLane: String, errorInfo: String)
 }
 
 public extension LaneFileProtocol {
     var fastlaneVersion: String { return "" } // Defaults to "" because that means any is fine
-    func beforeAll() {} // No-op by default
+    func beforeAll(currentLane _: String) {} // No-op by default
     func afterAll(currentLane _: String) {} // No-op by default
     func onError(currentLane _: String, errorInfo _: String) {} // No-op by default
     func recordLaneDescriptions() {} // No-op by default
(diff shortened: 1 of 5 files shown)


minuscorp:
Is this SPM compliant and xcodeproj as well? It seems weird to me that it does not get called, it should crash if the Lanefile is nil...🤔 could you check that the latest release which fixes some Swift issues doesn't fix this as well?
Thank you!🤗

jhandguy:
@minuscorp I find it weird too, but I have no clue what else it could be. 🤔 
This is happening on the latest version `2.161.0` when using the Xcode project. Nothing is crashing, only the `beforeAll` is not getting called. In this fix, I just made it consistent in the way `afterAll` is called: no need to force unwrap the instance variable `fastfileInstance` when this one is available in the local scope as non-optional.

Now, I have trouble testing this since I have only Xcode 12 installed on my machine and I don't seem to be able to build the `FastlaneSwiftRunner` properly, getting a Swift 5.3 error, can you maybe help me test it? We just need to find out whether the `beforeAll` method gets called or not and why. 👍

jhandguy:
Alright @minuscorp I managed to fix the Swift 5.3 issue by updating `SwiftShell` to `5.1.0` 👍 
Now I could test the change and it did not solve the issue 😱 

Any idea what the problem could be?

jhandguy:
@minuscorp I was able to fix the issue by calling the `beforeAll` and `afterAll` methods via selector! 🎉 
It is all working as expected now! 🤗 

I also took the opportunity to add a `currentLane _: String` parameter in the `beforeAll` method to be consistent with the `afterAll` one. 🚀

jhandguy:
Added a `responds(to: selector)` check before running the `perform(selector)`, as both `beforeAll` and `afterAll` are "optional" and have a default implementation in the `LaneFileProtocol` extension. ✅

minuscorp:
It seems very weird to me... And doesn't makes sense that responds to selector works and calling the method directly doesn't.
I have a feeling that it has something to be with the no-op implementation. Maybe we could try to see if providing a mock imp to the beforeAll method that is at the top and see if it gets called. If it is, the issue comes from other place.
I'm unfortunately unable to test it by myself due to sickness this week, hope this is not a rush and an older version is usable for you.

Thank you very much🤗

jhandguy:
@minuscorp since this issue was introduced as part of https://github.com/fastlane/fastlane/pull/16844 and released in version `2.159.0`, this means the last working version is `2.158.0`. Is it too much of a problem to merge this and figure out an alternative later?

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -136,13 +131,19 @@ open class LaneFile: NSObject, LaneFileProtocol {
         }
 
         // Call all methods that need to be called before we start calling lanes.
-        fastfileInstance.setUpAllTheThings()
+        let beforeAllSelector = NSSelectorFromString("beforeAllWithCurrentLane:")
+        if fastfileInstance.responds(to: beforeAllSelector) {
+            _ = fastfileInstance.perform(beforeAllSelector, with: laneMethod)
+        }
```
joshdholtz, line 137: Quick question! What is the difference between calling the method directly and calling it via `NSSelectorFromString`? 🤔  TBH, calling methods from a string always scares me since there isn't any compile time check 🤷‍♂️

joshdholtz (commented):
@jhandguy Quick question because I don't know stuff 🙃

fastlane/swift/LaneFileProtocol.swift
jhandguy: @joshdholtz I am also very much unaware about the way this was initially built, but my guess is that somehow changing the type of the `fastfileInstance` from `Fastfile` to `Lanefile` made calling the method directly to be executing the default implementation of the `LaneFileProtocol`'s extension, instead of the actual implementation declared in the `Fastfile`. 🤔 

image 

I also don't like this solution very much and I would love to hear suggestions, as long as it allows `Fastlane.Swift` users to be able to update from `2.158.0` to `2.161.0` without risking to get more regression issues. 🚀 

Unfortunately my time is also limited, though I'll try to unlock some more and play around with the type of `fastfileInstance` to see if my assumption was correct. 👍

minuscorp: Yeah, that change was to make spm work properly, but I clearly messed up. It could be what you say, that's why I recommended to change the no-op to something else to see the caller. Also maybe call that method has to be deferent for SPM or Xcode project.

jhandguy: @minuscorp we could also try type-casting `fastfileInstance` to `Fastfile` when using an Xcode project, maybe? 🤔

minuscorp:
```swift
#if !SWIFT_PACKAGE
(fastfileInstance as? Fastfile)?.beforeAll()
#endif
```

jhandguy: And in the case of SPM? Did you test if it was working?

minuscorp: As I said, in currently resting from two weeks at the hospital, I cannot use a laptop yet sorry. In SPM it should work because the Lanefile comes from the parent dependency, but if an issue is there I'll check it out ASAP I'm healthier 😅

jhandguy:
@joshdholtz @minuscorp that was it, the `fastfileInstance` was explicitly unwrapped and casted to `LaneFile` instead of `Fastfile` when running on Xcode project. 👍 
Force-pushed to tidy up the commits. ✅

fastlane/swift/LaneFileProtocol.swift
jhandguy: @minuscorp no worries, that wasn't my question actually: I was asking if you tested the `beforeAll` & `afterAll` use case at the time you made the PR, wasn't expecting you to test that now. 👌
Get well soon! 🤒

minuscorp: IIRC correctly it was working correctly on from SPM, I will revisit it though, although there user base is much smaller. In a future maybe we only have spm, who knows.

jhandguy: Alright, I might try testing it on SPM as well later 👍

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -136,13 +131,14 @@ open class LaneFile: NSObject, LaneFileProtocol {
         }
 
         // Call all methods that need to be called before we start calling lanes.
-        fastfileInstance.setUpAllTheThings()
+        fastfileInstance.beforeAll(currentLane: named)
```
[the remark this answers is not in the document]
jhandguy: I added it to be consistent with the `afterAll` lifecycle method. If you think it doesn't belong there I would argue it does not belong in the `afterAll` method either. My guess is that one might want to add shared logic in between lanes that should happen either before or after:

```
fun beforeAll(currentLane: String) {
   switch currentLane {
      case "myLane1", "myLane2":
         doSomethingBeforeLane1And2()
  }
}

...


fun afterAll(currentLane: String) {
   switch currentLane {
      case "myLane1", "myLane2":
         doSomethingAfterLane1And2()
  }
}
```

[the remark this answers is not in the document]
minuscorp: Afaik, beforeAll and afrerAll are called just once, so there logic adding the first and last Lane executed as parameter is confused and inconsistent with the Ruby behavior doesn't it?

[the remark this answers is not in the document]
jhandguy: Yes absolutely, `beforeAll` and `afterAll` are called only once, before and after a lane is executed from the command line.
Now, I'm not sure what you are arguing about, as my point here had nothing to do with the Ruby behaviour, but the current Swift behaviour and about making both `beforeAll` and `afterAll` lifecycle methods consistent with each other.

However, if you think `beforeAll` and `afterAll` do not need a parameter, I'm happy to remove it in both places, I'm only aiming for consistency here 👍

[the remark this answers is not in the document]
minuscorp: Maybe I didn't explained well. Fastlane Swift aims to mimic Ruby behaviour, so the best solution here would be to use arguments if the method in Ruby accept any argument and remove them if in Ruby's fastlane has no arguments. 🤔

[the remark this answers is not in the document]
jhandguy: Ah ok, I didn't understand that: yes, the Ruby methods `before_all` & `before_each` and `after_all` & `after_each` are receiving the `lane` as an argument. 👍 

Also, the Swift `beforeAll` and Ruby `before_all` seem to be consistent with one another: they are executed only once, compared to the Ruby `before_each` which is executed multiple times.

See: https://docs.fastlane.tools/advanced/lanes/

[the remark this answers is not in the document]
jhandguy: @minuscorp would you maybe prefer renaming the argument from `currentLane` to simply `lane` then?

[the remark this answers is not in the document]
minuscorp: It seems that both has the lane as argument: https://docs.fastlane.tools/advanced/lanes/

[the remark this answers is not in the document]
minuscorp: I'd use a more Swift Benning like `beforeAll(with lane: )` and same way with afterAll. Do you think it fits?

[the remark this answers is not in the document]
jhandguy: Sure! 👍

jhandguy:
@minuscorp renamed the `beforeAll` and `afterAll` parameter. ✅

minuscorp (approved):
LGTM

joshdholtz (approved):
This looks much better! Thanks for the great work, team ❤️

Issue #17243, opened 2020-09-15
deliver issue: "Waiting for screenshot to appear before uploading..."


exavi:
### New Regression Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Regression Information

- Breaking version: [e.g. `2.x.x`]
- Last working version: [e.g. `2.x.x`]

I can't quite specify when it last worked, but I know for sure, based on the uploads that I have done through Xcode's "Organizer" that it first failed for me on the 21st August, and it last worked on the 14th, so somewhere between that it broke.

I have tested it with recent versions and it seems to keep happening.

### Regression Description

When running deliver, it gets stuck uploading images, and it keeps repeating:
"Waiting for screenshot to appear before uploading..."

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>[20:42:25]: Successfully exported and compressed dSYM file
[20:42:25]: Successfully exported and signed the ipa file:
[20:42:25]: /Users/exavi/git/sw/app/src/ios/app.ipa
[20:42:26]: ---------------------
[20:42:26]: --- Step: deliver ---
[20:42:26]: ---------------------
[20:42:26]: Successfully loaded '/Users/exavi/git/sw/app/src/ios/fastlane/Deliverfile' 📄

+-----------------------+-----------------------+
| Detected Values from './fastlane/Deliverfile' |
+-----------------------+-----------------------+
| app_identifier        | com.my.app  |
| username              | my@email.com     |
+-----------------------+-----------------------+

[20:42:26]: Login to App Store Connect (my@email.com)
Requesting session...
[20:43:51]: Login successful

+--------------------------------------+-----------------------------------------------+
|                               deliver 2.159.0 Summary                                |
+--------------------------------------+-----------------------------------------------+
| force                                | true                                          |
| ipa                                  | /Users/exavi/git/sw/app/src/ios/app.ipa |
| screenshots_path                     | ./fastlane/screenshots                        |
| metadata_path                        | ./fastlane/metadata                           |
| app_version                          | 1.7.1                                         |
| username                             | byexavi@gmail.com                             |
| app_identifier                       | app                          |
| platform                             | ios                                           |
[...]

PR #17346, opened 2020-09-30
[deliver] fix screenshot uploads issue when overwrite_screenshot option is set to false

ainame:
### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/issues/17243
In this issue, it's reported that `deliver` got stuck with the message "Waiting for screenshot to appear before uploading..." and didn't recover from it shortly. That message comes from this bit.

https://github.com/fastlane/fastlane/blob/53ce1fab1f616b981fc80d3abd97320f65187ba7/spaceship/lib/spaceship/connect_api/models/app_screenshot.rb#L100-L115

After some investigations, I managed to replicate the similar errors locally and concluded that it was a regression caused in #16972.  

https://github.com/fastlane/fastlane/issues/17243#issuecomment-700347130 

Whilst #16972 gives significant performance improvements, it seemed to have a slight behaviour change when `overiwrite_screenshots` option is set to `false`. 

### Description

(This is just copied from  https://github.com/fastlane/fastlane/issues/17243#issuecomment-700347130 )

https://github.com/fastlane/fastlane/blob/53ce1fab1f616b981fc80d3abd97320f65187ba7/deliver/lib/deliver/upload_screenshots.rb#L131-L148

The above check code "index >= 10" still works without "overwrite_screenshots: true" **if you append images to only the end of the set**. But in the scenario below, "index" will kind of make a contradiction... 

Let's say we have already uploaded 6 screenshots for 6.5-inch screen like this. 

1. "6.5_1.jpg" (uploaded)
1. "6.5_2.jpg" (uploaded)
1. "6.5_3.jpg" (uploaded)
1. "6.5_4.jpg" (uploaded)
1. "6.5_5.jpg" (uploaded)
1. "6.5_6.jpg" (uploaded)

And then you happen to copy another lang's 6.5inch screenshots and paste to the directory ^ those screenshots are stored. It will be like this if sorted. (This would sound weird but was the easiest way to replicate this actually.)

1. "6.5_1 copy.jpg" (local)
1. "6.5_1.jpg" (uploaded)
1. "6.5_2 copy.jpg" (local)
1. "6.5_2.jpg" (uploaded)
1. "6.5_3 copy.jpg" (local)
1. "6.5_3.jpg" (uploaded)
1. "6.5_4 copy.jpg" (local)
1. "6.5_4.jpg" (uploaded)
1. "6.5_5 copy.jpg" (local)
1. "6.5_5.jpg" (uploaded)
1. "6.5_6 copy.jpg" (local)
1. "6.5_6.jpg" (uploaded)

With the current code, it's going to pick up the first 10 images from the above list to decide "skip" or "upload", which means -

* 5 local images; "6.5_1 copy.jpg", "6.5_2 copy.jpg", "6.5_3 copy.jpg", "6.5_4 copy.jpg" and "6.5_5 copy.jpg", are going to be newly uploaded and "6.5_1.jpg", "6.5_2.jpg" , "6.5_3.jpg" , "6.5_4.jpg" , "6.5_5.jpg" are skipped
* "6.5_6 copy.jpg" and "6.5_6.jpg" will be just ignored
* Since 6 images "6.5_[1-6].jpg" have been uploaded already, at least one of those uploads must fail (it's concurrently working so can't tell which one fails)

------------------------

This behaviour was covered by this initialisation of the count of existing uploaded screenshots.

https://github.com/fastlane/fastlane/blob/8c1d2c9cd894d76b29abf1fe40ef3c830e23a166/deliver/lib/deliver/upload_screenshots.rb#L140-L144

But with the refactoring of extractions of iterators, it was changed to rely only on index worked out from local data.
https://github.com/fastlane/fastlane/blob/53ce1fab1f616b981fc80d3abd97320f65187ba7/deliver/lib/deliver/app_screenshot_iterator.rb#L87-L91

This PR obsoleted that `index` params returned from the iterator and then restore counting logic with initialising `app_screenshot_set.app_screenshots.count` like before.

### Testing Steps

You can test this with the scenario in the description. I also added a unit test case to cover the scenario.

### DIFF
--- a/spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
@@ -111,7 +111,9 @@ module Spaceship
           timeout_minutes = (ENV["SPACESHIP_SCREENSHOT_UPLOAD_TIMEOUT"] || 20).to_i
 
           loop do
-            puts("Waiting for screenshot to appear before uploading...")
+            # This error handling needs to be resived since any error occured can reach here.
+            # It should handle errors based on what status code is.
+            puts("Waiting for screenshots to appear before uploading. This is unlikely to be recovered unless it's 503 error. error=\"#{error}\"")
             sleep(30)
 
             screenshots = Spaceship::ConnectAPI::AppScreenshotSet
--- a/spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
@@ -111,7 +111,7 @@ module Spaceship
           timeout_minutes = (ENV["SPACESHIP_SCREENSHOT_UPLOAD_TIMEOUT"] || 20).to_i
 
           loop do
-            # This error handling needs to be resived since any error occured can reach here.
+            # This error handling needs to be revised since any error occured can reach here.
             # It should handle errors based on what status code is.
             puts("Waiting for screenshots to appear before uploading. This is unlikely to be recovered unless it's 503 error. error=\"#{error}\"")
             sleep(30)
--- a/deliver/lib/deliver/app_screenshot_iterator.rb
+++ b/deliver/lib/deliver/app_screenshot_iterator.rb
@@ -52,10 +52,10 @@ module Deliver
       end
     end
 
-    # Iterate given local app_screenshot over localizations and app_screenshot_sets with index within each app_screenshot_set
+    # Iterate given local app_screenshot over localizations and app_screenshot_sets
     #
     # @param screenshots_per_language [Hash<String, Array<Deliver::AppScreenshot>]
-    # @yield [localization, app_screenshot_set, app_screenshot, index]
+    # @yield [localization, app_screenshot_set, app_screenshot]
     # @yieldparam [optional, Spaceship::ConnectAPI::AppStoreVersionLocalization] localization
     # @yieldparam [optional, Spaceship::ConnectAPI::AppStoreScreenshotSet] app_screenshot_set
     # @yieldparam [optional, Deliver::AppScreenshot] screenshot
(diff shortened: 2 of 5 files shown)


spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
```diff
@@ -111,7 +111,9 @@ def self.create(app_screenshot_set_id: nil, path: nil, wait_for_processing: true
           timeout_minutes = (ENV["SPACESHIP_SCREENSHOT_UPLOAD_TIMEOUT"] || 20).to_i
 
           loop do
-            puts("Waiting for screenshot to appear before uploading...")
+            # This error handling needs to be resived since any error occured can reach here.
+            # It should handle errors based on what status code is.
+            puts("Waiting for screenshots to appear before uploading. This is unlikely to be recovered unless it's 503 error. error=\"#{error}\"")
```
ainame, line 116: This is to give us more details when fastlane-users report issues. I feel like we should revise this error handling to check error response's status code to be accurate (or maybe we should get rid of this workaround somehow...?) 

See https://github.com/fastlane/fastlane/issues/17243#issuecomment-700262947

spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
```diff
@@ -111,7 +111,9 @@ def self.create(app_screenshot_set_id: nil, path: nil, wait_for_processing: true
           timeout_minutes = (ENV["SPACESHIP_SCREENSHOT_UPLOAD_TIMEOUT"] || 20).to_i
 
           loop do
-            puts("Waiting for screenshot to appear before uploading...")
+            # This error handling needs to be resived since any error occured can reach here.
```
rogerluan, line 114:
```suggestion
            # This error handling needs to be revised since any error occured can reach here.
```

😇

rogerluan (changes requested):
This is not a full review, just a typo 🙈 
I don't know enough to review this properly for the time being 🙇 but as far as my knowledge goes, logic looks good. This will need more testing from kind volunteers 😇

spaceship/lib/spaceship/connect_api/models/app_screenshot.rb
ainame: Nice catch! Thanks

joshdholtz (approved):
Thank you so much for jumping on this so quickly! I really appreciate it ❤️ I tested this out and it seems like it all works now 😊 You are the best!

PR #17354, opened 2020-10-01
[sigh] fix resigning of OnDemandResources when app has space in name

silvervest:
### Checklist

### Motivation and Context
If the app path has a space in it, the call to dirname here results in the following error, which is silently ignored
```
usage: dirname path
```

The following line of code then checks if `/OnDemandResources` is a directory, which never will be, and this code is never executed.

### Description
Just wrapping the $APP_PATH variable in double quotes to ensure it will work for paths with spaces in them.

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -535,7 +535,7 @@ function resign {
     fi
 
     # Check for and resign OnDemandResource folders
-    ODR_DIR="$(dirname $APP_PATH)/OnDemandResources"
+    ODR_DIR="$(dirname "$APP_PATH")/OnDemandResources"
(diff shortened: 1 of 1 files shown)


joshdholtz:
@lyndsey-ferguson This is a simple change but I think you are still more of an expert here than me 😛 Can I borrow your 👀?

lyndsey-ferguson (approved):
Looks good to me. Also double-checked on my Bash console 🖥

joshdholtz:
@lyndsey-ferguson Thank you so much!!

joshdholtz (approved):
Looks great! Thank you so much for the contribution ❤

silvervest:
Thanks! 👍

PR #17360, opened 2020-10-02
[snapshot] update SnapshotHelper.swift to work with tvOS

### LINKED ISSUE #17359: SnapshotHelper.swift fails on tvOS
### New Issue Checklist

### Issue Description

When I run my UITests on Apple TV, I get the following error with [SnapshotHelper 1.24](https://github.com/fastlane/fastlane/commit/8ec8fcd1e963a456fe5e0a916a63d733ca2acf94#diff-1a1173200436690d61ef63b6702ae6d6)

```
SnapshotHelper.swift: value of type 'XCUIDevice' has no member 'orientation'
```

The reason is that [UIDeviceOrientation](https://developer.apple.com/documentation/uikit/uideviceorientation) is not supported on tvOS.

The fix is simple, wrap the failing line in `#if os(iOS)` like this:

```
#if os(iOS)
    let image = XCUIDevice.shared.orientation.isLandscape ?  fixLandscapeOrientation(image: screenshot.image) : screenshot.image
#else
    let image = screenshot.image
#endif
```

felginep:
### Checklist

### Motivation and Context

Fix #17359

### Description

Avoid using `UIDeviceOrientation` on tvOS.

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -298,4 +298,4 @@ private extension CGFloat {
 
 // Please don't remove the lines below
 // They are used to detect outdated configuration files
-// SnapshotHelperVersion [1.25]
+// SnapshotHelperVersion [1.24]
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -165,7 +165,11 @@ open class Snapshot: NSObject {
             }
 
             let screenshot = XCUIScreen.main.screenshot()
+            #if os(iOS)
             let image = XCUIDevice.shared.orientation.isLandscape ?  fixLandscapeOrientation(image: screenshot.image) : screenshot.image
+            #else
+            let image = screenshot.image
+            #endif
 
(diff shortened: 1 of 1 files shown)


felginep:
@googlebot I signed it!

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -294,4 +298,4 @@ private extension CGFloat {
 
 // Please don't remove the lines below
 // They are used to detect outdated configuration files
-// SnapshotHelperVersion [1.24]
+// SnapshotHelperVersion [1.25]
```
rogerluan, line 301: This can be reverted as it's auto-increased as part of the release process 🙏

rogerluan (approved):
Neat fix 🙌  
Nice catch, and thanks for your contribution! ❤️

snapshot/lib/assets/SnapshotHelper.swift
felginep: Done

joshdholtz (approved):
This is a good catch! Thank you for fixing this ❤️

PR #17387, opened 2020-10-06
Fix comment about using `force` option

frozenzia:
Using `force` option in `upload_to_app_store` is designed to skip the HTML verification step.

Not 100% sure if there is or is NOT some sort of PDF report which is being referenced in line 29:
"If you don't want a PDF report for App Store builds, use the `:force` option." So I am not including any sort of change regarding that.

### Checklist

### Motivation and Context

Just a simple documentation fix. Helps reduce confusion.

### Description

Changed single reference to "PDF" to "HTML"

Documentation change, no testing needed.

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -153,7 +153,7 @@ module Deliver
         FastlaneCore::ConfigItem.new(key: :force,
                                      short_option: "-f",
                                      env_name: "DELIVER_FORCE",
-                                     description: "Skip the HTML report file verification",
+                                     description: "Skip verification of HTML preview file",
                                      is_string: false,
                                      default_value: false),
         FastlaneCore::ConfigItem.new(key: :overwrite_screenshots,
--- a/fastlane/lib/fastlane/actions/upload_to_app_store.rb
+++ b/fastlane/lib/fastlane/actions/upload_to_app_store.rb
@@ -50,7 +50,7 @@ module Fastlane
       def self.example_code
         [
           'upload_to_app_store(
-            force: true, # Set to true to skip PDF verification
+            force: true, # Set to true to skip HTML verification
             itc_provider: "abcde12345" # pass a specific value to the iTMSTransporter -itc_provider option
           )',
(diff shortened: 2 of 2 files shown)


frozenzia:
@googlebot I signed it!

rogerluan (changes requested):
I checked that the `upload_to_app_store` action uses _deliver_, and _deliver_ has this:

https://github.com/fastlane/fastlane/blob/5ee9f2e1f1d602e2af129e94eec2313091448abf/deliver/lib/deliver/options.rb#L153-L158

So in fact the code you're referencing sound like it should print HTML instead of PDF.

You just missed another reference to PDF here:

https://github.com/fastlane/fastlane/blob/5ee9f2e1f1d602e2af129e94eec2313091448abf/fastlane/lib/fastlane/actions/upload_to_app_store.rb#L29

Can you also update it?

frozenzia:
On Tue, 6 Oct 2020, 16:26 Roger Oba, <notifications@github.com> wrote:

> You just missed another reference to PDF here:
>
>
> https://github.com/fastlane/fastlane/blob/5ee9f2e1f1d602e2af129e94eec2313091448abf/fastlane/lib/fastlane/actions/upload_to_app_store.rb#L29
>
> Can you also update it?
>

If u check the text in my PR, this is exactly the bit that I was unsure
about. If u are confirming this is talking about the same HTML summary,
then sure, I can do it.

-paul

>

rogerluan:
Yes, looks like it is :) Probably in the past we had some sort of PDF but then it got updated to a HTML page and these docs were never updated 😅

frozenzia:
Alright, changes made, actually tweaked it even further so it is more in line with what actually happens -- you get an HTML PREVIEW of the changes to be made, and are asked to verify that unless you use the `force` option. Also tweaked the text in 'deliver'.

rogerluan (approved):
Nice! Thanks for the contribution @frozenzia 🚀

joshdholtz (approved):
I think this change! Really appreciate the contribution ❤️

Issue #12984, opened 2018-07-30
Assign Uploaded Build to Test Group


awgeorge:
### Feature Request
Ability to assign an uploaded build to another testing group #11249

#### Motivation Behind Feature

We have many groups configured in our organisation to control the flow of our releases. For example the build first gets uploaded for developers, then progresses to QA, continues to PMs and so forth.

Our CI pipeline can handle this, however we get an error message when trying to assign a build to a new group.

#### Feature Description

Allow `upload_to_testflight` to `skip_binary_upload` on subsequent calls, to enable assigning the already uploaded build to a new group.

Currently trying to upload the same IPA again, but to a different group results in an error:

> ERROR ITMS-90189: "Redundant Binary Upload. You've already uploaded a build with build number '160775' for version number '1.0.4'. Make sure you increment the build string before you upload your app to App Store Connect. Learn more in Xcode Help (http://help.apple.com/xcode/mac/current/#/devba7f53ad4)."
Return status of iTunes Transporter was 1: ERROR ITMS-90189: "Redundant Binary Upload. You've already uploaded a build with build number '160775' for version number '1.0.4'. Make sure you increment the build string before you upload your app to App Store Connect. Learn more in Xcode Help (http://help.apple.com/xcode/mac/current/#/devba7f53ad4)."
The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure.

First call to upload the binary and distribute externally (as today):

```
upload_to_testflight(
      ipa: options[:ipa],
      username: ENV["APP_STORE_USERNAME"],
      groups: ["1) DeveloperTesting"],
      distribute_external: true,
      changelog: changelog
 )
```

Second call with a change of groups would succeed either by modifing the command to `skip_binary_upload` or by detecting the response error `ITMS-90189` and handle it accordingly.

```
upload_to_testflight(
      ipa: options[:ipa],
      username: ENV["APP_STORE_USERNAME"],
      groups: ["2) QA"],
      skip_binary_upload: true
)
```

 One would assume a different error would be returned from the service if you were trying to assign a build to a group that already have that build - but would require investigation. Otherwise I do not see a reason why this would be a negative change.

#### Alternatives or Workarounds

Maybe using spaceship you could code something, but modifying the `upload_to_testflight` is likely the sensible and easiest approach.

awgeorge:
@fastlane-bot - As I marked this as a feature request, there was no mentioning in the ticket that I needed to include my environment.

[...]

PR #17408, opened 2020-10-08
[pilot] added a pilot command to add individual existing testers to a build.

dg76:
### Checklist
There are errors in create_keychain_spec.rb but I did not change the file and I don't think it was caused by my changes. It was probably already there.
Should I update the documentation of "fastlane pilot"? Or don't you want the new command in that place anyway?

### Motivation and Context
It adds support for adding individual testers to a build, which is possible in Apple's connect API but was not possible with fastlane so far. Please see here: https://github.com/fastlane/fastlane/discussions/17404#discussioncomment-92965

### Description
Using e.g.

`fastlane pilot add_to_build --email tester@tester.com -a com.bundle.test -u you@yourdomain.com --build-number 1.0.5
`

it is now possible to add the existing tester "tester@tester.com" to the build number "1.0.5".

As you can see it adds a new command to "pilot". This is exactly what I wanted to have but I am not sure if you maybe want to rename the command or make it a parameter of another command. What do you think? 🙂


dg76:
@googlebot I signed it!

pilot/lib/pilot/tester_manager.rb
```diff
@@ -27,6 +27,32 @@ def add_tester(options)
...
+    def add_tester_to_build(options)
+      start(options)
+      app = find_app(apple_id: config[:apple_id], app_identifier: config[:app_identifier])
+      UI.user_error!("You must provide either a Apple ID for the app (with the `:apple_id` option) or app identifier (with the `:app_identifier` option)") unless app
+
+      email_param = config[:email]
+      UI.user_error!("You must provide 1 or more emails (with the `:email` option)") unless email_param
+
+      if config[:build_number].nil?
+        build = app.get_builds[0]
```
rogerluan, line 39: This is more ruby 💎 
```suggestion
        build = app.get_builds.first
```

pilot/lib/pilot/tester_manager.rb
```diff
@@ -27,6 +27,32 @@ def add_tester(options)
...
+      else
+        builds = app.get_builds.select do |b|
+          config[:build_number] == b.version
+        end
+        UI.user_error!("Build #{config[:build_number]} not found.") unless builds.count == 1
+        build = builds[0]
```
rogerluan, line 45: Same here 😇 
```suggestion
        build = builds.first
```

pilot/lib/pilot/tester_manager.rb
```diff
@@ -27,6 +27,32 @@ def add_tester(options)
...
+      end
+      UI.success("Build: #{build.version} id: #{build.id}")
```
rogerluan, line 47: I think this doesn't need to be displayed prominently as a success message. Maybe `UI.message` or `UI.verbose` would be more appropriate, what do you think? 😊

pilot/lib/pilot/tester_manager.rb
```diff
@@ -27,6 +27,32 @@ def add_tester(options)
...
+
+      app.get_beta_testers.select do |tester|
+        next unless email_param.include?(tester.email)
```
rogerluan, line 50: Should we check for an exact (case insensitive) match instead of using `include`? This would return false positives in case `email_param` is `jpmorgan@email.com` and `tester.email` is `morgan@email.com`, for instance 😅

rogerluan: Oh, I get it now... You used `include` to cover cases where the `:email` argument contains multiple emails, is that right?
Although this might work, it's not very intuitive imo, first because email is in singular but also because of the false positive I mentioned above 🤔 
What if this API allowed only single emails to be used, for now? Either that, or use a `:emails` (plural) field, that is a list of strings that can be properly parsed as so. Then, in this condition, we do an exact match 😃 

Thoughts @dg76 ?

rogerluan (commented):
First of all, welcome to _fastlane_! I'm so excited that you decided to create a PR after our discussion [here](https://github.com/fastlane/fastlane/discussions/17404#discussioncomment-92965) 🤩 

I left some comments regarding the usage of UI messages and arguments, please take a look 🙂 
Architecture-wise, the way you implemented makes sense to me, but I'm not the most experienced one to judge this, so I'll leave that to @joshdholtz 🤓 

Documentation-wise, the commands_generator might generate documentation automatically for you, but I'm actually not sure. Maybe Josh can point us in the right direction here as well 🙇 

Once again, thanks for your contribution @dg76 ❤️

pilot/lib/pilot/tester_manager.rb
dg76: Yes, that's better, I didn't know these functions yet. Maybe like this?

```suggestion
      UI.verbose("Build: #{build.version} id: #{build.id}")
```

pilot/lib/pilot/tester_manager.rb
dg76: Oh, you are right, I didn't see that there is only one email. I used some code for searching the correct groups from the add_tester function as an example. Yes, then let's better use an exact match. So maybe like this?

```suggestion
        next unless email_param == tester.email
```

dg76:
> First of all, welcome to _fastlane_! I'm so excited that you decided to create a PR after our discussion [here](https://github.com/fastlane/fastlane/discussions/17404#discussioncomment-92965) 🤩
> 
> Once again, thanks for your contribution @dg76 ❤️

Thank you very much, I am glad that you like my contribution!! 😊

Let's see what the others say regarding the architecture and the documentation. 🙂

pilot/lib/pilot/tester_manager.rb
rogerluan: Yup, sounds good!

pilot/lib/pilot/tester_manager.rb
rogerluan: That's fine by me, but note that that implies the email argument should take just one email, so the docs and comments will need to be updated 😊

dg76: I have added docs and updated the comments now. Please see below, I have added the changes to the PR. 😊

joshdholtz (commented):
@dg76 I really like what this PR is accomplishing! 😊 I'm a _little_ bit torn though if this should be a pilot command or if this should be a new separate action 🤔 If this is a separate action, it could get run with like `fastlane run add_tester_to_build email:"email@email.com"`

One reason for my thought that this _could_ be an action is that it can then also get run from within a `Fastfile` if somebody wanted to use it. We have been trending from not adding new commands because they aren't easily callable from a `Fastfile`

But that being said... this command _does_ fit very well into `pilot` 🤔 

I'm tend to put everything into `Fastfile`'s myself so I would personally lean towards an action approach for this but I'd be happy to hear your thoughts! 😊

dg76:
@joshdholtz Thank you very much, I am glad that you like my PR! 😊 I needed it to send test builds to single users so that I can examine a problem with them without annoying all the other beta testers with too many updates.

I think it would be good if it could also be used from within a `Fastfile`. However as a user of fastlane I would search such a  function in `fastlane pilot`. So I can understand both, too. 

You know fastlane better than I do. Isn't it possible to directly call functions from within the `Fastfile` as written by you on 31 Jul 2018 here:

https://github.com/fastlane/fastlane/issues/12984

? So wouldn't it be possible to keep it as a command in "fastlane pilot" but to also be able to call it from the `Fastfile` by directly calling the function e.g. from the connect API of fastlane? 🤔 Or would it be necessary to add a function in another file so that it can be called from Fastfile?

dg76:
I have found a problem. My function successfully adds an individual tester to a build using the official function Apple provides: https://developer.apple.com/documentation/appstoreconnectapi/assign_individual_testers_to_a_build

However the build stays in "ready to submit" state and thus the testers cannot install it. 🙁

I also need to call

`Spaceship::ConnectAPI.post_beta_app_review_submissions(build_id: build.id)
`

This works fine for the first tester but causes an error if the build has already been posted for review submission. I have searched a while now for a way to check if the build has already been postet but e.g. `build.betaAppReviewSubmission` did not work. Can you tell me how I can check if the build has already been postet?

dg76:
I have found and implemented a solution now that will only post the beta version for review if it has not already been postet.

dg76:
I think there is a bug in Apple's API 🙁. When using my new function everything looks fine in App Store Connect afterward. The tester has been added to the build and the build is displayed with a green icon and the label "Testing". The tester also receives a notification from TestFlight that a new build is available with the correct build number. However when the tester opens TestFlight using the notification, TestFlight shows that it cannot find the build any more. And despite refreshing TestFlight multiple times, only the old builds are visible. The only way to fix it is to manually remove and re-add the tester on the App Store Connect website, which makes this whole new function useless. 😖

So it is obviously a bug in Apple's API, because it looks correct on the App Store Connect website (tester added, green "Testing" status) but the tester can not see or download the build. Do you know someone at Apple who can fix this?

rogerluan:
It does sound like an issue on Apple's end 😞 You should probably report it to Apple @dg76 .

@joshdholtz do you usually use the Feedback Assistant, or do you have a closer connection to someone there? Or is there a dedicated section in the new Apple forums? 😇 I'm also curious as to the usual process here, as I see that you report bugs to them quite often 😁

joshdholtz:
I'll take a look! I'll try to debug what's going on but if not I'll file another radar 😀

dg76:
@joshdholtz Hi! Did you have time to take a look at the problem, why one cannot download the TestFlight version after it has been released using post_beta_app_review_submissions? 🤔 Were you able to reproduce the problem?

joshdholtz:
@dg76 Hello, sorry! Got a little bit side tracked 😔 I'll _shhhouullddd_ be looking into this this afternoon hopefully!

dg76:
@joshdholtz Did you find something out? 😬

rogerluan:
Ping @joshdholtz 🤗

joshdholtz:
Will look at this tomorrow!

joshdholtz:
@dg76 I have no been able to replicate the issue you are seeing yet 😔 I will test it later tonight with my wife’s phone and Apple ID when she gets home from work tonight. I will also be adding a commit to this PR if that’s cool for adding a tester when their email isn’t already in the system 😇

dg76:
Hi @joshdholtz ! I have tried it again now, with my wife's phone and her Apple ID but it still doesn't work. I have uploaded a new version using `fastlane beta`. Afterward I ran `fastlane pilot add_to_build --email mywife@gmail.com -a my.package -u myuser`. The Apple ID was already registered as a tester. The build appears in green and "Testing" afterward in AppStore Connect and she has been added as an individual tester but she cannot see the build in her TestFlight app. 🙁

joshdholtz:
@dg76 Hey! Did she get an email for the invite and did she accept it? After accepting it she should see the app in her TestFlight app? I _believe_ that is how it works 🤷‍♂️

dg76:
@joshdholtz She is already an external tester in TestFlight (as I said, "The Apple ID was already registered as a tester."). So she does not get an email invitation. She just gets a notification from TestFlight on her iPhone. It says that a new build is available. When she opens that notification, TestFlight displays an error that the build cannot be found. She can still see the old build but despite refresing the view or restarting TestFlight using the task manager, she cannot see the new build.

dg76:
To mention it again: In AppStore Connect everything looks correct, the build is green, "Testing" and my wife is listed as tester. Nevertheless she does not see the build. But if I enter her via the AppStore Connect website instead of via fastlane or the Connect API, it looks exactly the same, but then it works and she sees the build.

dg76:
@joshdholtz Have you tested it with an existing user as described in my previous post? Does that work for you?

dg76:
@joshdholtz If it works fine for you maybe you can just merge it and then we can see if others have the same problem. Maybe it is just a problem of my Apple account.

dg76:
Hello? I have not heard anything for four weeks now. I fear that this PR might become forgotten. 😬

joshdholtz:
@dg76 Looking back into this!! Sorry for doing dark 😢

dg76:
Great! 😊 I was unsure if it was forgotten forever. 😬

dg76:
The conflict that is shown here now does not really seem to be a conflict. Instead a new function seems to have been added that looks very similar. Just keep both functions.

rogerluan:
@dg76 could you merge master into your branch and solve the merge conflicts? 🙏

dg76:
@rogerluan I have merged it now.

PR #17505, opened 2020-10-27
[snapshot] pass FASTLANE_LANGUAGE env variable to xcode for use in run scripts

### LINKED ISSUE #16337: [snapshot] localize status bar time, since Apple's status_bar tool doesn't
if I run `xcrun simctl status_bar <device> override --time $(date -r 1568122860 +%FT%T%z)` in the Test pre-actions, as suggested in https://github.com/fastlane/fastlane/issues/15124, I get a nice status bar. However, the time is not localized, even if I have localized the app and launched it with a language override (FB7668656). I want to fix this in my app by generating a localized date string myself, since I found that I can pass a string like 上午9:41 to `status_bar override --time` and it will accept it.

However, I can't find a reference to the language that the current test is running in. Is there an env variable or fastlane Ruby command I can run to figure out the language currently being overridden? I tried digging into the `localize_simulator` command in fastlane to see if I could read the language out of the plist that it write, but it looks like it's modified too late for my script to catch it.

#### Feature Description
In lieu of a proper `clean_status_bar` feature/plugin (which could come later) I'd like the ability to look up the language that is being tested by the current `snapshot` session, either via an env var or by other means, so I can use it in my build script.

I could probably write a script that loops over the languages one by one and runs snapshot for each of them, but I'd rather not tear up my whole snapfile, since it's already set up the way I want it. I just need access to this one piece of information via an env variable or similar.

joshdholtz:
### Motivation and Context
Fixes #16337

### Description
- Pass `FASTLANE_LANGUAGE` as environment variable into xcodebuild for use in build phase run scripts

### Testing Steps

```rb
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "joshdholtz-snapshot-pass-language-as-env-to-xcode"
```


ZevEisenberg (approved):
I can't vouch for the ruby code, but I've confirmed that this passes the env var correctly, so it's available in a script where I need it. 🚀

PR #15035, opened 2019-07-17
[action] Consolidate git_branch action logic into git_helper.rb to ensure consistent results

arri-cc:
Fixes: #15034 

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

As reported in #15034, in certain scenarios the `git_branch()` Action would return different results than the `git_helper.rb` method with the same name `Actions.git_branch`.  This was discovered while trouble shooting issues with the `ensure_git_branch()` action.  Instead of maintaining 2 copies of the logic to retrieve the current branch, we should consolidate it into the appropriate helper class.

### Description

I simply copied the shared values and logic from `fastlane/lib/fastlane/actions/git_branch.rb` into `fastlane/lib/fastlane/helper/git_helper.rb` and updated the origin to call the updated internal method.

Issue #15034, opened 2019-07-17
Update git_branch action to support all known environment variables


arri-cc:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

When running the following lane in bitrise.io in PR mode, I get different responses from the `git_branch()` action and the internal `Actions.git_branch` action used by the `ensure_git_branch()` action.  
```ruby
platform :android do

  desc "Test both git_branch actions"
  lane :test_git_branch do |options|
    UI.message "Call with Actions.git_branch #{Actions.git_branch}"
    UI.message "Call with git_branch() #{git_branch()}"
  end

end
```

As previously mentioned in the #13346 the `Actions.git_branch` internal action `/fastlane/lib/fastlane/helper/git_helper.rb` should be updated to account for the full range of known CI environment variables to ensure consistency.  Given it's an overloaded method name, it's not immediately clear there are 2 different implementations, as the documentation only references the public action.  The public `git_branch()`action implemented here `/fastlane/lib/fastlane/actions/git_branch.rb` should directly call the internal version, once updated, to keep this logic DRY and provide convenience for those using this action.

##### Command executed

`bundle exec fastlane test_git_branch`
##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> [04:28:07]: [32m------------------------------[0m
[04:28:07]: [32m--- Step: default_platform ---[0m
[04:28:07]: [32m------------------------------[0m
[04:28:07]: [32mDriving the lane 'android test_git_branch' 🚀[0m
[04:28:07]: Call with Actions.git_branch HEAD
[04:28:07]: [32m------------------------[0m
[04:28:07]: [32m--- Step: git_branch ---[0m
[04:28:07]: [32m------------------------[0m
[04:28:07]: Call with git_branch() chore/my-test-branch

+------+------------------+-------------+
|           [32mfastlane summary[0m            |
+------+------------------+-------------+
| Step | Action           | Time (in s) |
+------+------------------+-------------+
| 1    | default_platform | 0           |
| 2    | git_branch       | 0           |
+------+------------------+-------------+

[04:28:07]: [32mfastlane.tools finished successfully 🎉[0m

[...]

PR #17527, opened 2020-10-30
[action] slack - Fix branch being set to HEAD when running from CI

revolter:
### Checklist

### Motivation and Context

When running from a CI, the branch is always set to `HEAD` instead of the actual branch name. This is because the action, as opposed to the helper, checks for CI presence. I left the helper call in the `if` check, because that one returns `nil` if no branch is checked out. Otherwise, it would have always returns `true` because the action returns an empty string in this case.

Related to #15034 and #15035.


revolter:
@joshdholtz, @janpio, is this something that would get accepted?

rogerluan:
@googlebot I consent.

Issue #17498, opened 2020-10-27
FASTLANE_DISABLE_COLORS doesn't have any effect when executing a plugin action


Cyberbeni:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I'm using the [test_center](https://github.com/lyndsey-ferguson/fastlane-plugin-test_center) plugin to run tests (UI tests sometimes fail because failed to capture view hierarcy and this plugin retries failed tests) and I would like to disable colors on CI but `export FASTLANE_DISABLE_COLORS=1` doesn't seem to have any effect when running an action from the plugin.

##### Command executed
```
export FASTLANE_DISABLE_COLORS=1
bundle exec fastlane multi_scan # has colors
bundle exec fastlane lane_that_only_calls_multi_scan_with_predefined_parameters # has colors
bundle exec fastlane scan # no colors
```

Cyberbeni:
Colors are also not disabled for `bundle exec fastlane add_plugin ...` and `bundle exec fastlane incorrect_lane_name`

Cyberbeni:
From our logs, this used to work with 2.114.0, which was before this change:
https://github.com/fastlane/fastlane/blob/404ca994f8ecd1a8cfec1f1deca445605327fdc9/fastlane/lib/fastlane/cli_tools_distributor.rb#L130
Might not be related. @liamnichols any idea what would cause this?

liamnichols:
Hmmm, `load_dot_env` should only relate to anything in your Dotfiles... 

Does it work if you create the **.env** file in the root of your repo and add `export FASTLANE_DISABLE_COLORS=1` inside it?

Cyberbeni:
Having the `.env` file produces the same results as exporting it manually before running fastlane. I had no .env or .env.default file before.

liamnichols:
Hmmm this is odd... I created a new directory, added the gem and gave it a go and could not reproduce:

<img width="850" alt="Screenshot 2020-11-02 at 18 11 32" src="https://user-images.githubusercontent.com/482871/97897701-28cf1980-1d37-11eb-9f02-a8d80edbd49b.png">

[...]

PR #17547, opened 2020-11-03
[fastlane_core] disable colors for colorize gem too

Cyberbeni:
### Checklist

### Motivation and Context

Some plugins might use the colorize gem for coloring the texts. If we can import that, we should also disable the colors for that too. It would also override the String coloring methods that we disabled, making some of the output colored again.

### Testing Steps

Only fastlane in gem/pluginfile initially
```
export FASTLANE_DISABLE_COLORS=1
bundle install
bundle exec fastlane not_existing_lane
bundle exec fastlane add_plugin test_center
bundle install
bundle exec fastlane not_existing_lane
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -95,6 +95,7 @@ Require/MissingRequireStatement:
     - '**/spec/**/*.rb'
     - '**/spec_helper.rb'
     - 'spaceship/lib/spaceship/babosa_fix.rb'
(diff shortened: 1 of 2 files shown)


Cyberbeni:
Fixes #17498

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Cool fix! Really appreciate the contribution ❤️ Thank you!!

PR #17560, opened 2020-11-05
[pilot] update to new emoji regex

joshdholtz:
### Motivation and Context
The `emoji_regex` gem that is used in `pilot` has a new regex that is preferred and the other is deprecated - https://github.com/ticky/ruby-emoji-regex/releases/tag/v3.2.0

### Description
- Check if new regex constant exists and use it otherwise use the other

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -281,7 +281,7 @@ module Pilot
 
     def self.strip_emoji(changelog)
       if changelog && changelog =~ emoji_regex
-        changelog.gsub!(EmojiRegex::Regex, "")
+        changelog.gsub!(emoji_regex, "")
         UI.important("Emoji symbols have been removed from the changelog, since they're not allowed by Apple.")
       end
       changelog
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -273,8 +273,14 @@ module Pilot
       changelog
     end
 
+    def self.emoji_regex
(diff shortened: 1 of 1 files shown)


rogerluan (approved):
Cool! 😎 

Edit: oops 🙈 👇

pilot/lib/pilot/build_manager.rb
```diff
@@ -273,8 +273,14 @@ def self.truncate_changelog(changelog)
...
+    def self.emoji_regex
+      # EmojiRegex::RGIEmoji is now preferred over EmojiRegex::Regex which is deprecated as of 3.2.0
+      # https://github.com/ticky/ruby-emoji-regex/releases/tag/v3.2.0
+      return defined?(EmojiRegex::RGIEmoji) ? EmojiRegex::RGIEmoji : EmojiRegex::Regex
+    end
+
     def self.strip_emoji(changelog)
-      if changelog && changelog =~ EmojiRegex::Regex
+      if changelog && changelog =~ emoji_regex
         changelog.gsub!(EmojiRegex::Regex, "")
```
rogerluan, line 284: What about this one? Shouldn't we...

```suggestion
        changelog.gsub!(emoji_regex, "")
```

rogerluan (changes requested):

pilot/lib/pilot/build_manager.rb
joshdholtz: lol, you are right

rogerluan (approved):
🚀🚀

PR #17599, opened 2020-11-12
[Fastlane.Swift][update_fastlane] upgrade Fastlane.swift files

jhandguy:
### Checklist

### Motivation and Context

It would be nice to be able to upgrade Fastlane.swift files automatically when running the `update_fastlane` action.

### Description

Upgrade Fastlane.swift files automatically upon general Fastlane update.

### DIFF
--- a/fastlane/lib/fastlane/actions/update_fastlane.rb
+++ b/fastlane/lib/fastlane/actions/update_fastlane.rb
@@ -71,6 +71,9 @@ module Fastlane
           cleaner.execute
         end
 
+        upgrader = SwiftRunnerUpgrader.new
+        upgrader.upgrade_if_needed!
+
         UI.message("fastlane.tools successfully updated! I will now restart myself... 😴")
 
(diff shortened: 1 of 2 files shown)


rogerluan (approved):
I like this! 🎉 (code review only, I didn't test it)

joshdholtz (changes requested):
This change makes a lot of sense! But it looks like there are some tests failing 😔 Do you mind fixing these? 😊 

Also... what is the behavior like if its a pure Ruby implementation? Or do we need another check for that somewhere?

jhandguy:
@rogerluan @joshdholtz it should be good now, found the helper to check wether Fastlane is being ran with Swift or not 👍 
Also tested with the Fastlane project itself, by removing the `swift` folder, worked like a charm 👌

joshdholtz (approved):
Looks perfect! Thanks for making that change ❤️ ❤️ ❤️

PR #17661, opened 2020-11-24
[deliver] prepare for adding new validations for screenshots (refactoring)

ainame:
### Checklist

### Motivation and Context

This is a preparation for this idea https://github.com/fastlane/fastlane/discussions/17191 to implement. 

It's basically about making `deliver` more proactive in screenshot uploading. When a file extension doesn't match its content, `fastlane` should notify that to users. I have seen this mistake a few times at my company. Thus I thought it's worth implementing a simple validation for image format in `deliver`. I believe this will help people to identify issues and give them more confidence to use fastlane.

In order to achieve that, I'd like to start with some refactoring to avoid massive PR but the overall plan is below

1. Refactor `UploadScreenshots.collect_screenshots` to make it easy to add another rule apart from screen resolution checks (this PR)
2. Add image format checks with using FastImage, which fastlane already depends on
3. Resolve the issue where wrong screenshot sizes can be detected but fastlane just crashes before pointing all the issues out (see details here https://github.com/fastlane/fastlane/discussions/17191#discussioncomment-119758)

3 is actually to fix an existing issue where validation doesn't perform well right now. Once all the above steps are done, screenshots validation will be that -

* Error reports include all the possible reasons for each image (as many checks as implemented)
* Error reports include all the wrong screenshots (deliver won't crash before completing all checks)

I think that's the user-friendly behaviour when one has many issues. You can find possible issues at once and don't have to repeat running `deliver` and fixing an issue found again and again. 

(This is the current working branch for 2 and 3 on my fork repo https://github.com/ainame/fastlane/compare/prepare-for-adding-new-validation...fix-app-screenshot-init-crash)

### Description

(Sorry for a big PR but actual code change besides tests are 141 lines additions and some test cases are moved to another file🙇 )

<details>

```
% git diff --stat fastlane/master -- deliver/lib
 deliver/lib/deliver/loader.rb             | 154 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------
 deliver/lib/deliver/upload_metadata.rb    |  14 ++++----------
 deliver/lib/deliver/upload_screenshots.rb |  65 +----------------------------------------------------------------
 3 files changed, 141 insertions(+), 92 deletions(-)
```

</details> 

First of all, I refactored `UploadScreenshots#collect_screenshots` at this time. I think that logic can belong to `Deliver::Loader` as the name suggests. So I extracted it and moved it to `Deliver::Loader`. 

And then I also introduced `Deliver::Loader::LanguageFolder` which abstracts a language folder/directory that is used for both metadata and screenshots. It makes collection logic focus on what matters. For example, `deliver` repeatedly uses `File.basname` to work out what language is represented by a directory path but that piece of code shouldn't be repeated as it contains certain knowledge that can be encapsulated. Another example is that once a `Deliver::Loader::LanguageFolder` is initialised it guarantees that it is a directory path and not a file path. That is safer than doing manual `File.directory?` checks where needed.

There was a recursive call of `collect_screenshots_for_languages` to collect special nested folders for Apple TV or iMessage screenshots. I got rid of it and re-implement it within `Deliver::Loader.language_folders` with option argument. So that the logic using it will be simplified.

https://github.com/fastlane/fastlane/blob/2624a24407dee94f67d7e39b5651f9fdc41dcb00/deliver/lib/deliver/upload_screenshots.rb#L257-L272

----

**Folder vs Directory**

I was wondering which term we would prefer to use in this project. In `deliver`, both words are quite mixed in variable names😇 I tried to use `folder` in my PR as much as possible. I believe we should use consistent naming. But that refactoring is out of the scope of this PR.  

e.g.

https://github.com/fastlane/fastlane/blob/2624a24407dee94f67d7e39b5651f9fdc41dcb00/deliver/lib/deliver/loader.rb#L45-L50

### Testing Steps

There shouldn't be a behaviour change and this passed existing test scenarios. 

With this in Gemfile,
 
```ruby
gem 'fastlane', git: 'https://github.com/ainame/fastlane.git', branch: 'prepare-for-adding-new-validation'
```

you can run `deliver` like this for checking if both metadata and screenshot upload s work.

```ruby
upload_to_app_store(
  skip_binary_upload: true,
  overwrite_screenshots: true,
)
```

You can also test if existing validation still works by making images wrong size.

```
% cd fastlane/screenshots/en-US
% mogrify -resize 388x392! 5.5_1.jpg 5.5_2.jpg 5.5_3.jpg 5.5_4.jpg 5.5_5.jpg 5.5_6.jpg
``

### CODE REVIEW

File: deliver/lib/deliver/upload_screenshots.rb [hunk 1]
```diff
@@ -251,70 +251,7 @@ def sort_screenshots(localizations)
...
-          UI.important("Unaccepted device screenshots are detected! 🚫 Screenshot file will be skipped. 🏃") unless unaccepted_device_shown
-          unaccepted_device_shown = true
-
-          UI.important("🏃 Skipping screenshot file: #{screenshot.path} - Not an accepted App Store Connect device...")
-        end
-        exists
-      end
-
-      return screenshots
+      return Loader.load_app_screenshots(options[:screenshots_path], options[:ignore_language_directory_validation])
```
joshdholtz, line 254: This is the best change 🤣

joshdholtz (approved):
This is fantastic 🤯 Read through code a few times and ran locally and everything look 💯 Thanks so much for another amazing contribution ❤️

Issue #17423, opened 2020-10-11
[Fastlane.Swift] SnapshotHelper.swift has Swift Compiler Error


xta:
### New Regression Checklist

- [X] Updated fastlane to the latest version
- [X] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [X] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [X] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Regression Information
I'm using fastlane (2.162.0) & Xcode Version 12.0.1 (12A7300).

### Regression Description
I'm unable to run `fastlane snapshot` after following the docs.

With the [current helper file](https://github.com/fastlane/fastlane/blob/master/snapshot/lib/assets/SnapshotHelper.swift#L179), it contains `try image.pngData()?.write(to: path, options: .atomic)`. Xcode cannot run since this throws the compiler error, `'pngData()' has been renamed to 'UIImagePNGRepresentation(_:)'`.

I was able to get fastlane snapshot working with this replacement line, `try UIImagePNGRepresentation(image)?.write(to: path, options: .atomic)`. Obviously your team can figure out what replacement line is preferred.

Thanks

minuscorp:
Thank you for reporting this, I'll work on this ASAP.

JimHowell20:
I'm getting the same error as well.

JimHowell20:
Updating the swift version from Swift 4.1 to Swift 5 in the xCode build settings for the build target fixed this error for me:  <img width="635" alt="Screen Shot 2020-10-13 at 2 48 27 PM" src="https://user-images.githubusercontent.com/7559768/95902886-6c9bb800-0d63-11eb-983b-acbfb24bcb43.png">

minuscorp:
Do you have an older code version (I.e. 11.X) where you can select version 4 and check if it fails? Thank you!

rogerluan:
I don't understand, `image.pngData()` should be the Swift version of `UIImagePNGRepresentation(image)`, which is in Objective-C. If you see [pngData docs](https://developer.apple.com/documentation/uikit/uiimage/1624096-pngdata) you will see it's not deprecated. @minuscorp does this still make sense to you? I initially thought it'd just be a matter of replacing one with the other, but `pngData` looks right too AFAICT 🤷‍♂️ 

[...]

PR #17687, opened 2020-11-27
Conditional compiler directive

minuscorp:
### Checklist

### Motivation and Context

Solves  #17423

### DIFF
--- a/snapshot/lib/assets/SnapshotHelper.swift
+++ b/snapshot/lib/assets/SnapshotHelper.swift
@@ -180,7 +180,7 @@ open class Snapshot: NSObject {
                 simulator = regex.stringByReplacingMatches(in: simulator, range: range, withTemplate: "")
 
                 let path = screenshotsDir.appendingPathComponent("\(simulator)-\(name).png")
-                #if Swift(<5.0)
+                #if swift(<5.0)
                     UIImagePNGRepresentation(image)?.write(to: path, options: .atomic)
                 #else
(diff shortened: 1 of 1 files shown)


joshdholtz (commented):
This looks great! Does this just need some testing for it to be ready to ship? 😊

minuscorp:
Yeah, i wrote this from the ipad because i was sure of what was the issue, but I haven't been able of testing it. Also, I wasn't totally sure of this being a real issue, but as we do support Swift 4, this directive had to be change, which is against of the apple docs (irony: _weird doesn't it?_) so tomorrow I'll compile in swift 4 mode and see if it works just fine!

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -180,7 +180,11 @@ open class Snapshot: NSObject {
                 simulator = regex.stringByReplacingMatches(in: simulator, range: range, withTemplate: "")
 
                 let path = screenshotsDir.appendingPathComponent("\(simulator)-\(name).png")
-                try image.pngData()?.write(to: path, options: .atomic)
+                #if Swift(<5.0)
```
rogerluan, line 183: >Unexpected platform condition (expected 'os', 'arch', or 'swift')

```suggestion
                #if swift(<5.0)
```

🎉

rogerluan (changes requested):

rogerluan (approved):
LGTM! 🚀

PR #17693, opened 2020-11-30
[docs] improve GitHub Pull Request template by unifying two similar items of the checklist

rogerluan:
### Checklist

### Motivation and Context
Having those two similar items seemed wrong to me, since we have `bundle exec fastlane test` that performs both actions.

### Description

I think the those two items can be unified since there's an action that would check both items of the checklist at once. This makes opening PRs 1.3 seconds faster, which would have saved us a total of 135 minutes in the lifetime of this repository since it was created 😇 😂 

### Testing Steps

Nothing, really. Just updating the PR template :)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,7 @@
 <!-- Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes by putting an x in the [ ] (don't: [x ], [ x], do: [x]) -->
 
 ### Checklist
-- [ ] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
-- [ ] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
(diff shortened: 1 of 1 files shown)


joshdholtz (changes requested):
I think I'd like to actually keep these separate if that's okay! Even though `bundle exec fastlane test` does do all the things, its terrible slow (especially if you are dealing with lint issues. I think its good to inform that users that `bundle exec rspec` and `bundle exec rubocop -a` are things they can run. Maybe we can find a way to incorporate both of those still?

rogerluan:
Interesting, I didn't notice it was too slow (probably because I just run `bundle exec fastlane test` and not the others 🙈)

[Testing.md](https://github.com/fastlane/fastlane/blob/master/Testing.md) is clear enough about all the options available (lint, test, test file, test group, etc), so during development you could use the faster ones and to open a PR run `bundle exec fastlane test` - it's not like people would run it multiple times a day 🤔

But you have a very valid point, I don't mind keeping them separate - I just thought they were all equivalent with no cons 😊 

Closing :) thanks for reviewing this!

PR #17703, opened 2020-12-01
[spaceship] ConnectAPI::App - actually limit get_app_store_versions

joshdholtz:
### Motivation and Context
Fix issues where passing a limit would still fetch all pages

### Description
Previously, using `limit` would fetch all pages but with the limited page size. This fix doesn't call `all_pages` if a limit is passed.


max-ott (approved):
Looking good to me!

PR #17364, opened 2020-10-02
[deliver] fix iPad 12.9" 3rd gen screenshot identifier for App Store Connect API

rogerluan:
🔑 

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Resolves #16975 
Resolves #18045
More context in this comment of mine: https://github.com/fastlane/fastlane/issues/16975#issuecomment-699516304

### Description

It appears that since the App Store Connect API, the name of screenshots downloaded from App Store Connect of the iPad 12.9" 3rd gen was changed to `%d_APP_IPAD_PRO_3GEN_129_%d`.
Since there's no logic to handle that identifier out of the box (i.e. if you download the screenshots using _deliver_ and try to re-upload them, it won't work unless you rename the iPad 12.9" 3rd gen screenshots according to [the docs](https://docs.fastlane.tools/actions/deliver/#uploading-screenshots-for-ipad-pro-129-inch-3rd-generation)), this PR fixes that, maintaining the old name identifiers, but also adding support to `IPAD_PRO_3GEN_129`.
This PR also improves some copy and documentation in general, that helps people better understand this caveat of the iPad screenshots.

### Testing Steps

This issue could be reproduced by downloading your screenshots from App Store Connect using deliver, and try re-uploading them. _fastlane_ doesn't know that it has to upload those screenshots and assign them to the iPad 12.9" 3rd gen, so it will attempt to upload them as regular iPad 12.9" 2nd gen screenshots, which isn't accepted by Apple during review time (it could potentially get metadata rejected - I've witnessed this).

To test this branch, modify your Gemfile as:

```ruby
gem "fastlane", :git => "https://github.com/rogerluan/fastlane.git", :branch => "fix-ipad-screenshot-naming"
```

PR #17708, opened 2020-12-02
[deliver] improve screenshot validation

ainame:
### Checklist

### Motivation and Context

Resolves https://github.com/fastlane/fastlane/discussions/17191

Prior to this PR, in https://github.com/fastlane/fastlane/pull/17661, I refactored `Deliver::UploadScreenshots.collect_screenshots` to report better error details for screenshot validation.

There are two issues to improve in `deliver`.

1. `Deliver::AppScreenshot.new` currently crashes (UI.user_error!) when the given screenshot has the wrong size
2. `deliver` doesn't check screenshot's image format before uploading and results in getting errors from App Store Connect API

With this PR, error reports for screenshot validation will become like below.

* Error reports include all the possible error details for each image (as many checks as implemented)
* Error reports include all the wrong screenshots (deliver won't crash before completing all checks)
* Error reports include some debug information for users

As a result, `deliver` users will be able to resolve screenshots issues easily and save time.

### Description

I implemented a new validation rule to check if screenshot's file extension matches its file format and ported file extension checks implemented in `Deliver::AppScreenshot.is_valid?`. Then, I extracted the entire validation logic to `Deliver::AppScreenshotValidator`. `Deliver::AppScreenshot.is_valid?` will be deprecated.

And also I changed `DeliverAppScreenshot.new` not to crash when not detecting valid screen size to allow us to report errors for all the images located. I assume that crash behaviour was implemented to make sure that given `screen_size` param at initialiser matches the actual size calculated. However, if we were to not use `screen_size` param, we wouldn't have to worry about it after all. Since nothing in fastlane code uses `screen_size` param apart from tests, I'd like to deprecate it in favour of `Deliver::AppScreenshotValidator`.

Since the unintentional crash will be fixed, `deliver` won't crash with the current `Deliver::Loader.load_app_screenshots` even when detecting invalid screenshots. So I've introduced a new control flow here in `load_app_screenshots` to avoid unwanted uploading. This way `deliver` still offers better error reports but the previous behaviour to prevent uploading screenshots will remain.

https://github.com/fastlane/fastlane/blob/effa740f61bed9b0a634ed0e9fe2f34c7dca6341/deliver/lib/deliver/loader.rb#L103-L117

As a result, you can see the error reports like this. 

Screenshot 2020-12-03 at 02 19 02

To compare with existing behaviour, this is what you only get with 2.168 with the same screenshot sets.

Screenshot 2020-12-03 at 02 35 58

### Testing Steps

With this in Gemfile 

```ruby
gem 'fastlane', git: 'https://github.com/ainame/fastlane.git', branch: 'fix-app-screenshot-init-crash'
```

you can run deliver like this for checking if both metadata and screenshot upload s work.

```ruby
upload_to_app_store(
  skip_binary_upload: true,
  overwrite_screenshots: true,
)
```

You can also test if existing validation still works by making images wrong size.

```
% cd fastlane/screenshots/en-US
% mogrify -resize 388x392! 5.5_*.jpg # to create invalid screen size screenshots
% mogrify -resize 828x1792! 12.9_*.jpg # to create unaccepted device's screenshots
% mv 6.5_1.jpg 6.5_1.png # to create invalid screenshots that mismatch its content format and file extension
```

### DIFF
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -84,7 +84,7 @@ module Deliver
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
     #  will automatically be calculated when you don't set it. (Deprecated)
     def initialize(path, language, screen_size = nil)
-      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deperecated in favor of the default behavior to calculate size automatically. Passed value is no longer validated.') if screen_size
+      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deprecated in favor of the default behavior to calculate size automatically. Passed value is no longer validated.') if screen_size
       self.path = path
       self.language = language
       self.screen_size = screen_size || self.class.calculate_screen_size(path)
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -82,15 +82,12 @@ module Deliver
     # @param path (String) path to the screenshot file
     # @param language (String) Language of this screenshot (e.g. English)
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
-    #  will automatically be calculated when you don't set it.
+    #  will automatically be calculated when you don't set it. (Deperecated)
     def initialize(path, language, screen_size = nil)
+      UI.deprecated('screen_size param for Deliver::AppScreenshot.new is deperecated.') if screen_size
       self.path = path
       self.language = language
-      screen_size ||= self.class.calculate_screen_size(path)
-
-      self.screen_size = screen_size
-
-      UI.error("Looks like the screenshot given (#{path}) does not match the requirements of #{screen_size}") unless self.is_valid?
+      self.screen_size = screen_size || self.class.calculate_screen_size(path)
     end
 
     # The iTC API requires a different notation for the device
@@ -161,6 +158,7 @@ module Deliver
 
     # Validates the given screenshots (size and format)
     def is_valid?
+      UI.deprecated('Deliver::AppScreenshot#is_valid? is depereated in favor of Deliver::AppScreenshotValidator')
       return false unless ["png", "PNG", "jpg", "JPG", "jpeg", "JPEG"].include?(self.path.split(".").last)
 
       return self.screen_size == self.class.calculate_screen_size(self.path)
@@ -350,7 +348,7 @@ module Deliver
         end
       end
 
-      UI.user_error!("Unsupported screen size #{size} for path '#{path}'")
+      nil
     end
   end
 
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -84,7 +84,7 @@ module Deliver
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
     #  will automatically be calculated when you don't set it. (Deprecated)
     def initialize(path, language, screen_size = nil)
-      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deperecated in favor of the default behavior to calculate size automatically.') if screen_size
+      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deperecated in favor of the default behavior to calculate size automatically. Passed value is no longer validated.') if screen_size
       self.path = path
       self.language = language
       self.screen_size = screen_size || self.class.calculate_screen_size(path)
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -84,7 +84,7 @@ module Deliver
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
     #  will automatically be calculated when you don't set it. (Deprecated)
     def initialize(path, language, screen_size = nil)
-      UI.deprecated('screen_size param for Deliver::AppScreenshot.new is deperecated.') if screen_size
+      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deperecated in favor of the default behavior to calculate size automatically.') if screen_size
       self.path = path
       self.language = language
       self.screen_size = screen_size || self.class.calculate_screen_size(path)
--- a/deliver/lib/deliver/app_screenshot.rb
+++ b/deliver/lib/deliver/app_screenshot.rb
@@ -82,7 +82,7 @@ module Deliver
     # @param path (String) path to the screenshot file
     # @param language (String) Language of this screenshot (e.g. English)
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
-    #  will automatically be calculated when you don't set it. (Deperecated)
+    #  will automatically be calculated when you don't set it. (Deprecated)
     def initialize(path, language, screen_size = nil)
       UI.deprecated('screen_size param for Deliver::AppScreenshot.new is deperecated.') if screen_size
       self.path = path
@@ -158,7 +158,7 @@ module Deliver
 
     # Validates the given screenshots (size and format)
     def is_valid?
-      UI.deprecated('Deliver::AppScreenshot#is_valid? is depereated in favor of Deliver::AppScreenshotValidator')
+      UI.deprecated('Deliver::AppScreenshot#is_valid? is deprecated in favor of Deliver::AppScreenshotValidator')
       return false unless ["png", "PNG", "jpg", "JPG", "jpeg", "JPEG"].include?(self.path.split(".").last)
 
       return self.screen_size == self.class.calculate_screen_size(self.path)
--- a/deliver/lib/deliver/app_screenshot_validator.rb
+++ b/deliver/lib/deliver/app_screenshot_validator.rb
@@ -2,14 +2,14 @@ require 'fastimage'
 
 module Deliver
   class AppScreenshotValidator
-    # A simple structure that holds error information as well as format error message consistently
-    # Set `true` to `to_skip` when just needing to skip uploading rather than causing a crash.
+    # A simple structure that holds error information as well as formatted error messages consistently
+    # Set `to_skip` to `true` when just needing to skip uploading rather than causing a crash.
     class ValidationError
       # Constants that can be given to `type` param
       INVALID_SCREEN_SIZE = 'Invalid screen size'.freeze
       UNACCEPTABLE_DEVICE = 'Not an accepted App Store Connect device'.freeze
       INVALID_FILE_EXTENSION = 'Invalid file extension'.freeze
-      FILE_EXTENSION_MISMATCH = 'File extension mistmaches its image format'.freeze
+      FILE_EXTENSION_MISMATCH = 'File extension mismatches its image format'.freeze
 
       attr_reader :type, :path, :debug_info, :to_skip
 
@@ -32,8 +32,8 @@ module Deliver
     # Access each array by symbol returned from FastImage.type
     ALLOWED_SCREENSHOT_FILE_EXTENSION = { png: ['png', 'PNG'], jpeg: ['jpg', 'JPG', 'jpeg', 'JPEG'] }.freeze
 
-    # Validate a screenshot and inform an error message via `errors` parameters. `errors` is mutated
-    # to append the messages and each message should contain the corresponding path to let users know which file gets the error.
+    # Validate a screenshot and inform an error message via `errors` parameter. `errors` is mutated
+    # to append the messages and each message should contain the corresponding path to let users know which file is throwing the error.
     #
     # @param screenshot [AppScreenshot]
     # @param errors [Array<Deliver::AppScreenshotValidator::ValidationError>] Pass an array object to add validation errors when detecting errors
@@ -83,11 +83,11 @@ module Deliver
       end
 
       format = FastImage.type(screenshot.path)
-      is_extension_macthed = ALLOWED_SCREENSHOT_FILE_EXTENSION[format] &&
+      is_extension_matched = ALLOWED_SCREENSHOT_FILE_EXTENSION[format] &&
                              ALLOWED_SCREENSHOT_FILE_EXTENSION[format].include?(extension)
 
       # This error only appears when file extension is valid
-      if is_valid_extension && !is_extension_macthed
+      if is_valid_extension && !is_extension_matched
         errors_found << ValidationError.new(type: ValidationError::FILE_EXTENSION_MISMATCH,
                                             path: screenshot.path,
                                             debug_info: "Actual format is #{format}")
--- a/deliver/lib/deliver/app_screenshot_validator.rb
+++ b/deliver/lib/deliver/app_screenshot_validator.rb
@@ -32,11 +32,14 @@ module Deliver
     # Access each array by symbol returned from FastImage.type
     ALLOWED_SCREENSHOT_FILE_EXTENSION = { png: ['png', 'PNG'], jpeg: ['jpg', 'JPG', 'jpeg', 'JPEG'] }.freeze
 
+    APP_SCREENSHOT_SPEC_URL = 'https://help.apple.com/app-store-connect/#/devd274dd925'.freeze
+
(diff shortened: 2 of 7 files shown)


deliver/lib/deliver/app_screenshot_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+    def self.validate(screenshot, errors)
+      # Given screenshot will be diagnosed and errors found are accumulated
+      errors_found = []
+
+      validate_screen_size(screenshot, errors_found)
+      validate_device_type(screenshot, errors_found)
+      validate_file_extension_and_format(screenshot, errors_found)
+
+      # Merge errors found into given errors array
+      errors_found.each { |error| errors.push(error) }
```
rogerluan, line 50: To understand this code, I had to read how arguments are passed in ruby (as value, or as reference): https://robertheaton.com/2014/07/22/is-ruby-pass-by-reference-or-pass-by-value/ - just to realize that it's neither value or reference, but something like a mix of them 🤯 Very interesting! haha cool 😁

deliver/lib/deliver/app_screenshot_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+      errors_found.empty?
+    end
+
+    def self.validate_screen_size(screenshot, errors_found)
+      if screenshot.screen_size.nil?
+        errors_found << ValidationError.new(type: ValidationError::INVALID_SCREEN_SIZE,
+                                            path: screenshot.path,
+                                            debug_info: "Actual size is #{get_formatted_size(screenshot)}")
```
rogerluan, line 58: I saw how the output looks like in your PR description, and as a user, I'm told that "the size is wrong" and that the size is currently X, but what size (or sizes)  _should_ it be (or are valid)? Is this possible/inferable? 🤔
As a user, I'd like to know how to fix the issue I'm reading :)

deliver/lib/deliver/app_screenshot_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+      is_valid_extension = valid_file_extensions.include?(extension)
+
+      unless is_valid_extension
+        errors_found << ValidationError.new(type: ValidationError::INVALID_FILE_EXTENSION,
+                                            path: screenshot.path,
+                                            debug_info: "Only #{valid_file_extensions.join(', ')} are allowed")
+      end
+
+      format = FastImage.type(screenshot.path)
+      is_extension_macthed = ALLOWED_SCREENSHOT_FILE_EXTENSION[format] &&
```
rogerluan, line 86:
```suggestion
      is_extension_matched = ALLOWED_SCREENSHOT_FILE_EXTENSION[format] &&
```

deliver/lib/deliver/app_screenshot_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+                             ALLOWED_SCREENSHOT_FILE_EXTENSION[format].include?(extension)
+
+      # This error only appears when file extension is valid
+      if is_valid_extension && !is_extension_macthed
```
rogerluan, line 90:
```suggestion
      if is_valid_extension && !is_extension_matched
```

deliver/lib/deliver/app_screenshot_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+        errors_found << ValidationError.new(type: ValidationError::FILE_EXTENSION_MISMATCH,
+                                            path: screenshot.path,
+                                            debug_info: "Actual format is #{format}")
```
rogerluan, line 93: Likewise, here, as a user, I'd like to see "what it is" and "what it should be". Could we suggest how to fix this issue? For instance if the `image.png` has a jpeg format, they should probably rename the file to `image.jpeg`, right? Should we suggest this?

deliver/lib/deliver/loader.rb
```diff
@@ -100,36 +101,22 @@ def self.load_app_screenshots(root, ignore_validation)
...
-      # Given screenshot will be diagnosed and errors found are accumulated
-      errors_found = []
-
-      # Checking if the device type exists in spaceship
-      # Ex: iPhone 6.1 inch isn't supported in App Store Connect but need
-      # to have it in there for frameit support
-      if screenshot.device_type.nil?
-        errors_found << "🏃 Skipping screenshot file: #{screenshot.path} - Not an accepted App Store Connect device..."
+      unless errors_to_crash.empty?
+        UI.important("🚫 Invalid screenshots are detected! Here's the reasons.")
```
rogerluan, line 114: I think this would read better 😇 

```suggestion
        UI.important("🚫 Invalid screenshots were detected! Here are the reasons:")
```

deliver/lib/deliver/loader.rb
```diff
@@ -100,36 +101,22 @@ def self.load_app_screenshots(root, ignore_validation)
...
+        errors_to_crash.each { |error| UI.error(error) }
+        UI.user_error!("Canceled uploading screenshot. Please check above error messages and correct screenshots.")
```
rogerluan, line 116: I also think this would read better 🤗 
```suggestion
        UI.user_error!("Canceled uploading screenshots. Please check the error messages above and fix the screenshots.")
```

rogerluan (changes requested):
First of all, this feature is amazing! Well done - the PR description and content are 💯 ✨ 
I personally experience problems with invalid screenshots more often than I'm comfortable with, in the company I work for, and this is mostly because we lack an automated way to validate the screenshots we have 😊 

Having said that, I had thought about creating a utility just like the one you created. However, the way I pictured it, it would be a standalone new `action` to be executed on its own (without _deliver_), allowing the user to validate not only the screenshots, but also even check their filenames (because of the [iPad 3rd gen issue](https://github.com/fastlane/fastlane/pull/17364) that we have).

How do you feel about this? If we were to make this available as an action, do you think the action should simply call `Deliver::AppScreenshotValidator……`, or should this validator be implemented outside of deliver (e.g. as an action itself), and `deliver(…)` calls that action? 

Would love to hear your thoughts and other people's 😊 

Nonetheless, the logic of this PR seems solid to me. Most (if not all) of the suggestions I made are light, and copy-related only 😁 Good stuff!! 🚀

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -82,15 +82,12 @@ module ScreenSize
     # @param path (String) path to the screenshot file
     # @param language (String) Language of this screenshot (e.g. English)
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
-    #  will automatically be calculated when you don't set it.
+    #  will automatically be calculated when you don't set it. (Deperecated)
     def initialize(path, language, screen_size = nil)
+      UI.deprecated('screen_size param for Deliver::AppScreenshot.new is deperecated.') if screen_size
```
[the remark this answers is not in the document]
ainame: I just wanted to remove this param so there is no alternative in reality. I can add something like "don't use this"😄 

----

I suppose this param was introduced from [the very beginning](https://github.com/fastlane/fastlane/commit/f278d6f5961e26c461c20194aa210b208ccae577#diff-1e2d19d7725a6fd408e51052baa986851fe9b6decdb0bb03cf8c33bfa36a8e20) before [introducing automatic detection of the size](https://github.com/fastlane/fastlane/commit/bb543d376eef798dfd034137f88ac2dc5af95000#diff-1e2d19d7725a6fd408e51052baa986851fe9b6decdb0bb03cf8c33bfa36a8e20). 

https://github.com/fastlane/fastlane/blob/f278d6f5961e26c461c20194aa210b208ccae577/lib/ios_deploy_kit/app_screenshot.rb#L10-L18

There is no reason to keep it nowadays. It was used for testing purpose but we can mock it in tests with the power of RSpec.

deliver/lib/deliver/app_screenshot_validator.rb
ainame: Ah yes, it's correct. This is a bit tricky. Normally, this should be avoided as it's all about side-effect in a function. But API design-perspective, I wanted to return just a boolean rather than doing throw-catch dance. This allows me to accumulate errors found easily. I believe this is not that bad since this `validate` method is a simple class method. Using a small side effect for this purpose won't harm us that much.

deliver/lib/deliver/app_screenshot_validator.rb
ainame: Will think about a better idea for the debug information. Might end up with a link for the spec page though 😅

ainame:
@rogerluan Thank you for your review☺️ I'll come back soon with updating what you pointed about "suggestions for users".

ainame:
> Having said that, I had thought about creating a utility just like the one you created. However, the way I pictured it, it would be a standalone new action to be executed on its own (without deliver), allowing the user to validate not only the screenshots, but also even check their filenames (because of the iPad 3rd gen issue that we have).

> How do you feel about this? If we were to make this available as an action, do you think the action should simply call Deliver::AppScreenshotValidator……, or should this validator be implemented outside of deliver (e.g. as an action itself), and deliver(…) calls that action?

Regarding this, I have a thought on this. I'll drop it here later.

deliver/lib/deliver/app_screenshot_validator.rb
ainame: Added the link to help page 😄

deliver/lib/deliver/app_screenshot_validator.rb
ainame: Added a suggestion message (and a test case for it)

deliver/lib/deliver/app_screenshot_validator.rb
ainame: Updated document message to mention that the array passed will be mutated.

ainame:
@rogerluan 

> the way I pictured it, it would be a standalone new action to be executed on its own (without deliver), allowing the user to validate not only the screenshots, but also even check their filenames (because of the iPad 3rd gen issue that we have).

This is a good point. I thought about this before. Ideally, devs should notice the problems before they submit apps with `deliver` when this PR's validation is running😄 (Better late than never though). For example, in a very common scenario, `build_app` runs before `upload_to_app_store`  so if there are any issues in screenshots, the time/CI resources taken by `build_app` will be wasted due to fastlane crashing at `upload_to_app_store`. That's not ideal.

In reality, in the common scenario, we can save a bit of time without having to implement a new action for it. This way at least you won't waste about 10 or more mins for `build_app`. 

```ruby
# Build time will be wasted if there are screenshot issues
lane :common_submit do
  build_app
  upload_to_app_store(submit_for_review: true)
end

# Screenshot issues will  be detected in advance so you won't waste time
lane :proactive_submit do
  # Validate screenshots and reflect them if valid
  upload_to_app_store(skip_binary_upload: true, submit_for_review: false, overwrite_screenshots: true)

  # Build and archive app to submit 
  build_app

  # Only upload ipa file and go to submission
  upload_to_app_store(skip_metadata: true, skip_screenshots: true, submit_for_review: true)
end
```

In addition, the thing is that this is documented here. So I'm afraid that adding a new action that can be achieved without it has less value here 😞 

> New Actions
> Please be aware that we don’t accept submissions for new actions at the moment. You can find more information about that here.

https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md

----

In my opinion, the most practical way to address this issue is checking screenshots when you update fastlane's screenshots directory. If you were to commit screenshot images directly to your iOS project's git repo and manage them with pull requests, you can check your updates by `danger` or whatever CI toolings you prefer.

I'm currently using this piece of code in our `Dagerfile` at work. If this PR gets merged, I'm going to replace the logic with `Deliver::AppScreenshotValidator` for ourselves. But that would be another story...

```ruby
require 'deliver'

def review_screenshots
  # To support files that are tracked as "renamed" correctly we have to exclude file names removed from `git.modified_files`
  rename_map = git.renamed_files.map { |x| [x[:before], x[:after]] }.to_h
  modified_files = git.modified_files.map { |x| rename_map[x] || x }
  app_screnshot_paths = (git.added_files + modified_files).grep(%r|^fastlane/screenshots/(.+)/.*|)

  app_screnshot_paths.each do |path|
    language = path.match(%r|^fastlane/screenshots/(.+)/.*|).captures.first
    app_screenshot = Deliver::AppScreenshot.new(path, language)
    failure("#{path} is invalid. File format needs to be PNG or JPEG.") unless app_screenshot.is_valid?
    # to catch this error when image is invalid https://github.com/fastlane/fastlane/blob/6537974e792690dc63fb6fe8209046e8df0a8ac4/deliver/lib/deliver/app_screenshot.rb#L347
  rescue ::FastlaneCore::Interface::FastlaneError => error
    # Deliver::AppScreenshot.new can raise errors because of mismatch of resoluation
    failure(error.to_s)
  end
end
```

deliver/lib/deliver/app_screenshot.rb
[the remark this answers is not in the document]
rogerluan: If removing is possible without breaking backwards compatibility, I'd be in favor!

deliver/lib/deliver/app_screenshot.rb
```diff
@@ -82,9 +82,9 @@ module ScreenSize
     # @param path (String) path to the screenshot file
     # @param language (String) Language of this screenshot (e.g. English)
     # @param screen_size (Deliver::AppScreenshot::ScreenSize) the screen size, which
-    #  will automatically be calculated when you don't set it. (Deperecated)
+    #  will automatically be calculated when you don't set it. (Deprecated)
     def initialize(path, language, screen_size = nil)
-      UI.deprecated('screen_size param for Deliver::AppScreenshot.new is deperecated.') if screen_size
+      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deperecated in favor of the default behavior to calculate size automatically. Passed value is no longer validated.') if screen_size
```
rogerluan, line 87:
```suggestion
      UI.deprecated('`screen_size` for Deliver::AppScreenshot.new is deprecated in favor of the default behavior to calculate size automatically. Passed value is no longer validated.') if screen_size
```

deliver/lib/deliver/app_screenshot.rb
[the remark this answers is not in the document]
ainame: As fastlane is distributed as a gem and follows semantic versioning. I'd like to leave it as is to keep the interface!

rogerluan (changes requested):
Just one typo left 😁

deliver/lib/deliver/app_screenshot.rb
ainame: My emacs still completes words with old typo info 🙈

rogerluan:
@ainame love that `proactive_submit` solution - in fact, no separate action is required! I didn't think of using `upload_to_app_store` just to merely validate screenshots 😁 

Your idea of using Danger to validate the screenshots is great. I will definitely incorporate that into my workflow (although I use Danger in Swift at work 🙊 😂 )

In my particular workflow I also would like to validate the name of the screenshots (the iPad 3rd gen ones must be named after something specific, otherwise they silently fail on upload). Any ideas about that? Probably candidate for another PR? 😬 

Lmk if you need more clarification of what I'm referring to 😊 It's related to the PR I referenced earlier!

rogerluan (approved):
🚀 Good stuff!! Thanks for implementing this 🤩 Can't wait to start using it 🥳

ainame:
> In my particular workflow I also would like to validate the name of the screenshots (the iPad 3rd gen ones must be named after something specific, otherwise they silently fail on upload). Do you have any implementation design in mind to solve this? Probably candidate for another PR? 😬

No, not yet but it will be easier to add the rule as I have extracted validator already😄 I can work on it.

rogerluan:
> > In my particular workflow I also would like to validate the name of the screenshots (the iPad 3rd gen ones must be named after something specific, otherwise they silently fail on upload). Do you have any implementation design in mind to solve this? Probably candidate for another PR? 😬
> 
> No, not yet but it will be easier to add the rule as I have extracted validator already😄 I can work on it.

Absolutely! A lot easier I'd say haha. Now it works something like Precheck, with standalone checks 🤓 
Let me know if you need any help with more details, how-to-reproduce steps, or just testing!

joshdholtz (approved):
This is so fire! Sorry it took me so long to review but new 👶 as you know 😇  Will get this merged and out today!

joshdholtz:
@googlebot I consent.

PR #17740, opened 2020-12-06
[nexus_upload] fail action when the underlying nexus upload command line fails

### LINKED ISSUE #17444: Nexus upload does not fail when not having correct access.
### New Issue Checklist

### Issue Description
When uploading a file to nexus with the [nexus_upload](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/nexus_upload.rb) plugin without having the right access to the repo the fastlane lane does not fail. This is due to the plugin not handling errors. So for example I do get a `403 Forbidden` my nexus lane still finish successfully. 

##### Command executed
```
bundle exec fastlane upload_nexus
```

In my fastlane lane I have the following:
```ruby
platform :ios do
    desc "Upload to nexus"
    lane :upload_nexus do |options|
	nexus_upload(
	    nexus_version: 3,
	    file: "/path/to/file.zip",
	    mount_path: "/nexus",
	    repo_id: "repo",
	    repo_classifier: "my-repo_classifier",
	    repo_group_id: "my-repo_group_id",
	    repo_project_name: "my-repo_project_name",
	    repo_project_version: "1.0.0",
	    endpoint: "https://my-endpoint.com",
	    username: "my-username",
	    password: "my-password",
	    verbose: true
	)
   end
end
```

output:
```bash

+------+------------------+-------------+
|           fastlane summary            |
+------+------------------+-------------+
| Step | Action           | Time (in s) |
+------+------------------+-------------+
| 1    | opt_out_usage    | 0           |
| 2    | default_platform | 0           |
| 4    | nexus_upload     | 0           |
+------+------------------+-------------+
fastlane.tools finished successfully   
```

### Environment

Fastlane version: 2.163.0

## Posible solution
The plugin is using a curl command to upload the file to nexus. It should handle the non 20x statuses. A possible solution for this is adding the `--fail` option to the curl [command](https://curl.haxx.se/docs/). Than it will give a non zero exit code when not having a 200 status. 
An another solution is parsing the output of the curl command and handle all non 200 responses.

rogerluan:
### Checklist

### Motivation and Context
Resolves #17444

### Description
After reviewing #17444 I noticed that the OP suggested a fix that would resolve the issue for them, but they couldn't sign Google's CLA to open a PR to _fastlane_. This PR simply applies the suggested fix :) 

OP @firecodeing suggested that this option could be optional and default to the current implementation to avoid possible regressions, but both him and I agree that the current behavior should be considered a bug, since it's error prone, and should be fixed instead. So I decided to unconditionally add the `--fail` option to the command.

The `--fail` option is provided by is `curl` option: https://curl.se/docs/manpage.html

### Testing Steps

```ruby
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "rogerluan-fail-nexus-upload-on-error"
```

Then run [nexus_upload](https://docs.fastlane.tools/actions/nexus_upload/) in a way that it could fail, to witness the `--fail` option doing its magic and failing the lane 😊 

### Credits

Credits to this solution goes 100% to @firecodeing for doing the research and providing the solution 🙏

### DIFF
--- a/fastlane/lib/fastlane/actions/nexus_upload.rb
+++ b/fastlane/lib/fastlane/actions/nexus_upload.rb
@@ -5,6 +5,7 @@ module Fastlane
         command = []
         command << "curl"
         command << verbose(params)
+        command << "--fail"
         command += ssl_options(params)
         command += proxy_options(params)
         command += upload_options(params)


firecodeing:
@rogerluan Thank you for applying the fix. I can confirm this will fix my issue.

joshdholtz (approved):
This is great! Thank you so much ❤️

PR #17793, opened 2020-12-14
[itunes-transporter] fix transporter `DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS` env var validation

### LINKED ISSUE #17453: Invalid transport parameter
### New Regression Checklist

### Regression Information

- Breaking version: Since #16774 merged

### Regression Description
We specify a custom DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS env parameter to make itmsTransporter to use our proxy to speed up upload. 
We also not specify a -t parameter intentionally to let itmsTranspoter determine best trasport method for us. 
#16774 introduce a check which force DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS contains `-t ` else it will report `Invalid transport parameter`. Which is against what Apple suggested mentioned in #16774. 
Is that a bug?

rogerluan:
### Checklist

### Motivation and Context
Resolves #17453

Resources & Related Tickets:

This issue seems to have originated in this PR: https://github.com/fastlane/fastlane/pull/16774
The error described in #17453 was introduced here, more specifically: https://github.com/fastlane/fastlane/pull/16774/commits/977f8721910a512bafba9be3285016744f22d144#diff-e4392686971e0801abe75321cc8a4c90eb2f2c971734d8cc88b0ac5fef5ee12dR163 which goes against [Apple's Transporter User Guide 2.1](https://help.apple.com/itc/transporteruserguide/#/apdATD1E1288-D1E1A1303-D1E1288A1126).

### Description
Looks like the PR aforementioned added a validation to the `DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS` env var that would throw everytime that var didn't include `-t ` in its contents. However, looks like this option can be used to send other kinds of parameters other than `-t`, and at the same time, Apple suggests that users don't specify `-t` explicitly, but instead allow Transporter to automatically detect the best transport method (which at the moment defaults to `http`).

So removing the forced crash made sense to me, as discussed in #17453. 
Furthermore, I believe that adding a harmless warning/informative logs to the user can never hurt too much, so I added a `UI.important` message when that env var contains a `-t` option.
Lastly, by making those changes, 2 specs started failing. However, I think those specs wrong fundamentally wrong, as they expect that that env var should only be used to inform `-t` additional option, and that doesn't seem to be the case. For this reason, I deleted both of those specs.

### Testing Steps

```ruby
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "rogerluan-transporter-t-option"
```

Prior to this PR, passing `-t` option in a `DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS ` env var would succeed. If you passed a non-empty value in that same env var, without including `-t`, it would throw an error.
In this PR, if you pass `DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS` defining `-t`, it will throw a warning only.

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/upload_to_testflight.md
+++ b/fastlane/lib/fastlane/actions/docs/upload_to_testflight.md
@@ -218,7 +218,7 @@ fastlane pilot upload --verbose
 
 ## Firewall Issues
 
-_pilot_ uses the iTunes Transporter to upload metadata and binaries. In case you are behind a firewall, you can specify a different transporter protocol from the command line using
+_pilot_ uses the iTunes [Transporter](https://help.apple.com/itc/transporteruserguide/#/apdATD1E1288-D1E1A1303-D1E1288A1126) to upload metadata and binaries. In case you are behind a firewall, you can specify a different transporter protocol from the command line using
 
 ```no-highlight
 DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS="-t DAV" pilot ...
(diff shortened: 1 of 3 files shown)


ainopara:
lgtm👍

joshdholtz (approved):
This is so 🔥 Thank you so much for not just fixing this but also updating doc and adding that important information ❤️ Love love love.

PR #17804, opened 2020-12-14
[action] git_commit - skip commit if nothing is staged

felginep:
### Checklist

### Motivation and Context

For now, the action `git commit(allow_nothing_to_commit: true)` skips the command `git commit` only if the git status is clean. 

### Description

This PR improves the effect of `allow_nothing_to_commit`. The `git commit` command is skipped if nothing is staged. Meaning you can have a dirty repository, and still use `allow_nothing_to_commit`.

### DIFF
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -11,9 +11,9 @@ module Fastlane
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
         if params[:allow_nothing_to_commit]
-          repo_clean = Actions.sh("git status --porcelain").empty?
-          UI.success("Nothing to commit, working tree clean ✅.") if repo_clean
-          return if repo_clean
+          nothing_staged = Actions.sh("git --no-pager diff --name-only --staged").empty?
+          UI.success("Nothing staged to commit ✅.") if nothing_staged
+          return if nothing_staged
         end
 
         command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
--- a/fastlane/spec/actions_specs/git_commit_spec.rb
+++ b/fastlane/spec/actions_specs/git_commit_spec.rb
@@ -63,7 +63,7 @@ describe Fastlane do
 
       it "does not generate the git command when configured to allow nothing to commit and there are no changes to commit" do
         allow(Fastlane::Actions).to receive(:sh)
-          .with("git status --porcelain")
+          .with("git --no-pager diff --name-only --staged")
           .and_return("")
         result = Fastlane::FastFile.new.parse("lane :test do
           git_commit(path: './fastlane/README.md', message: 'message', allow_nothing_to_commit: true)


rogerluan (approved):
Nice catch!

joshdholtz (approved):
This makes so much more sense 😊 Thank you for the contribution! Really appreciate it ❤️

ainame:
@felginep Hi, I realised that this causes a behaviour change in the case where `git_commit` action tries to commit file changes without explicitly staging them. In that case, `--staged` option for `git diff` will miss unstaged changes in such a scenario because `gi commit` can commit "unstaged" changes and `git_commit` action was supposed to behave so, I suppose. The meaning of `allow_nothing_to_commit` seemed to be changed to `allow_nothing_staged`, to be precise.

For example, I used to have this kind of script but now I have to add `sh("git add #{ROOT_PATH}")` above the `git_commit` by myself, which feels not good to me since `git_commit` was a complete action for such operations but now I have to call a shell command via `sh` additionally.

```ruby
download_something
git_commit(path: ".", message: "update something", allow_nothing_to_commit: true)
```

Would you mind if I revert this? What are the scenarios you expect for this change?

felginep:
Oh I see, you do not use the `git_add` action before `git_commit`, and I understand this is a regression for you then. Sorry for that.

My personnal issue is the following. If I have this changes: 
```
> git status
modified: fileA
modified: fileB
```

And I try to commit `fileC` with 
```git_commit(path: "fileC", allow_nothing_to_commit: true)```
I want to avoid an error because `fileC` does not exist and I can't skip the `git commit` action because the repository is dirty.

ainame:
Oh I see I didn't know that there is git_add🙈 I'll have a think about this and get you back later!

felginep:
Same, I'll think of a solution and share it here if I find something relevant.

felginep:
@ainame I think we can revert my change and update the `repo_clean` variable to be `repo_clean = Actions.sh("git status #{paths} --porcelain").empty?` (note the new `#{paths}` in the command)

This way, we only verify the paths we pass to the lane are modified.

There are two cases:
- fileA **is** modified (meaning `git status fileA --porcelain` will return a non empty string)
  - we call `git_commit(path: "fileA", allow_nothing_to_commit: true)` -> the commit is created
  - we call `git_commit(path: "fileA", allow_nothing_to_commit: false)` -> the commit is created
- fileA **is not** modified (meaning `git status fileA --porcelain` will return an empty string)
  - we call `git_commit(path: "fileA", allow_nothing_to_commit: true)` -> the commit is skipped (even if `git status --porcelain` does not return an empty string, meaning there are other files modified)
  - we call `git_commit(path: "fileA", allow_nothing_to_commit: false)` -> an error occurs because `fileA` can't be staged

ainame:
@felginep Thank you for the suggestion. Yes, I had something similar in mind but was not sure if it works when I replied last time! This sounds good to me. 

Please go ahead to make a PR and make sure you add unit tests that cover the case I mentioned (commit without staging). I can then review your PR 🙇

felginep:
@ainame You can find the new PR here #17913

ainame:
Will have a look tonight! Thank you very much!

PR #17837, opened 2020-12-21
[xcodebuild] add (partial) support for xcframework generation

### LINKED ISSUE #17836: Support to create xcframeworks
### New Issue Checklist

### Issue Description

There's currently no action in `fastlane` to create XCFrameworks, other than manually crafting the proper parameters calling the `xcodebuild` action.

jgongo:
### Checklist

### Motivation and Context

Resolves #17836 

### Description

I have added a new action mimicking the already existing actions related with `xcodebuild`. I don't know if this is the proper approach for this, as the `xcodebuild` action has a lot of parameters that doesn't make sense when building xcframeworks. This also made a bit complex to add the `-library` and `-headers` parameters, as these parameters have to be coupled and the header option is optional for each library. Due to this I only provide support for the `-framework` parameter.

I have added tests to check that the command is properly built for the allowed parameters.

### Testing Steps

```
bundle exec fastlane test
```

### DIFF
--- a/fastlane/lib/fastlane/actions/xcodebuild.rb
+++ b/fastlane/lib/fastlane/actions/xcodebuild.rb
@@ -670,7 +670,7 @@ module Fastlane
       end
 
       def self.description
-        "Packages multiple build configurations of a given library or framework into a single xcframework"
+        "Packages multiple frameworks into a single xcframework"
       end
 
       def self.available_options
--- a/fastlane/lib/fastlane/actions/xcodebuild.rb
+++ b/fastlane/lib/fastlane/actions/xcodebuild.rb
@@ -659,7 +659,7 @@ module Fastlane
       def self.example_code
         [
           'xccreatexcframework(
-            framework: ["FrameworkA.framework", "FrameworkB.framework"],
+            frameworks: ["FrameworkA.framework", "FrameworkB.framework"],
             output: "UniversalFramework.xcframework"
           )'
         ]
--- a/fastlane/lib/fastlane/actions/xcodebuild.rb
+++ b/fastlane/lib/fastlane/actions/xcodebuild.rb
@@ -41,7 +41,6 @@ module Fastlane
         export_profile: "-exportProvisioningProfile",
         export_signing_identity: "-exportSigningIdentity",
         export_with_original_signing_identity: "-exportWithOriginalSigningIdentity",
-        frameworks: "-framework",
         hide_shell_script_environment: "-hideShellScriptEnvironment",
         jobs: "-jobs",
         output: "-output",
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/xcodebuild.rb
```diff
@@ -636,6 +648,47 @@ def self.author
...
+      def self.run(params)
+        params_hash = { create_xcframework: true }.merge(params)
+
+        XcodebuildAction.run(params_hash)
+      end
+
+      def self.example_code
+        [
+          'xccreatexcframework(
+            framework: ["FrameworkA.framework", "FrameworkB.framework"],
```
rogerluan, line 662: Typo:
```suggestion
            frameworks: ["FrameworkA.framework", "FrameworkB.framework"],
```

rogerluan (approved):
This is a great contribution, @jgongo ! 

I'm unsure whether we want to add this directly to _fastlane_'s `xcodebuild` action, or if a separate action would be more suitable (outside of _fastlane_ repo). I'll leave that up to @joshdholtz , but since we have similar `xcodebuild` actions already, I wouldn't mind leaving it here. 

Overall logic of this PR LGTM 💪

fastlane/lib/fastlane/actions/xcodebuild.rb
jgongo: Gonna try to create an independent action. As the `-create-xcframework` is quite independent from the rest of the `xcodebuild` commands, and most (if not all) of the parameters are exclusive to this action, I think it makes sense to do it. In addition having an independent action will let me easily include the parameters I had to leave unsupported because it would have made the logic to integrate them with the existing parameters too complicated. Let me work on it for a day or two...

jgongo:
> This is a great contribution, @jgongo !

Thanks 😄 

> I'm unsure whether we want to add this directly to _fastlane_'s `xcodebuild` action, or if a separate action would be more suitable (outside of _fastlane_ repo). I'll leave that up to @joshdholtz , but since we have similar `xcodebuild` actions already, I wouldn't mind leaving it here.

Gonna try to create an independent action. As the `-create-xcframework` is quite independent from the rest of the `xcodebuild` commands, and most (if not all) of the parameters are exclusive to this action, I think it makes sense to do it. In addition having an independent action will let me easily include the parameters I had to leave unsupported because it would have made the logic to integrate them with the existing parameters too complicated. Let me work on it for a day or two...

jgongo:
@rogerluan @joshdholtz I have created a new PR at #17844. I wasn't happy with the implementation integrated in the `xcodebuild` action, so I created a separate action which includes all the parameters available in the `-create-xcframework` action. Please feel free to comment and choose just one of both (I prefer the other one)

joshdholtz:
Closing this in favor of #17844 ❤️

PR #17854, opened 2020-12-24
[spaceship] allow 2FA request only in interactive mode

kohtenko:
### Checklist

### Motivation and Context
Under company wide policy it's impossible to create an account without 2FA on Apple developer portal. At the same time, we use `spaceauth` on our CI and once the session expires, the job hangs until it times out instead of failing immediately. As I don't see ways to automate process of 2FA on CI I rather prefer it to fail immediately in non-interactive mode.

### Description
Check for `interactive` on the very first step after receiving `409` and raise an exception to fail the execution.
Due to [the lint rule](https://github.com/fastlane/fastlane/blob/e89cdcfb869ad0640438cc5f6ed261bb50887641/fastlane/Fastfile#L36) to avoid `FastlaneCore::UI` I duplicated the logic to determine interactive mode. If there's no clear reason not to import `fastlane_core` I can simplify this by removing this lint

Tests expect the exception in case non-interactive launch and no exception in interactive.

### Testing Steps
- Run `bundle exec fastlane spaceauth` with an account with 2FA. Make sure interactive console requests 2FA code.
- Run `export CI=1 && bundle exec fastlane spaceauth` with an account with 2FA. Make sure action fails without waiting for interaction.


kohtenko:
@googlebot I signed it!

spaceship/lib/spaceship/two_step_or_factor_client.rb
```diff
@@ -4,6 +4,7 @@
 module Spaceship
   class Client
     def handle_two_step_or_factor(response)
+      raise "2FA can only be performed in interactive mode" unless interactive
```
joshdholtz, line 7:
```suggestion
      raise "2FA can only be performed in interactive mode" if ENV["FASTLANE_IS_INTERACTIVE"]
```

We like to keep `fastlane_core`  out of `spaceship` so this solution is pretty okay! But I'm curious if maybe we do something a little bit different and look for an environment variable called `FASTLANE_IS_INTERACTIVE`

And then maybe we do something like... 
```rb
# Set interactive environment variable for spaceship (which can't require fastlane_core)
ENV["FASTLANE_IS_INTERACTIVE"] = FastlaneCore::UI.interactive?
``

in https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/cli_tools_distributor.rb#L73

Thoughts? 😊

joshdholtz (changes requested):
I think this is a great addition! I got one though/suggestion to remove the duplicate logic 🤷‍♂️

spaceship/lib/spaceship/two_step_or_factor_client.rb
kohtenko: Makes sense to me. As `ENV` can only hold `String` values I compare the value to `false` to allow `spaceship` execution in case it's run outside of Fastlane.

rogerluan:
@googlebot I consent.

joshdholtz (approved):
This looks good to me! Thanks for making those changes 😊  Really appreciate the contribution!

ronaldjeremy:
fastlane-2.172.0 has broken the ability to do the request.  I set SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER and in versions prior to 2.172.0 it works.  Starting with 2.172.0 I always get "2FA can only be performed in interactive mode".

kohtenko:
@ronaldjeremy do you run it on CI in non-interactive mode? If it's non-interactive, how a default SMS number helps? You'll need to interactively provide the code to the console on the next step 🤔

garylai:
We run `fastlane spaceauth` inside a `node` program to refresh Apple session periodically. The 2FA code is written into the child_process using its `stdin`. But because it is used programmatically the stdio are not tty our pipeline is now broken. I think this may be breaking everyone that is using `fastlane spaceauth` programatically. Can you please add a config to turn off this check?

ryechus:
There are instructions in the docs for 2FA use in CI? Are those still accurate? https://docs.fastlane.tools/best-practices/continuous-integration/#application-specific-passwords

kohtenko:
Fixed in [this PR](https://github.com/fastlane/docs/pull/1019)

ryechus:
@kohtenko I agree with the sentiment of this change, but my concern is around changing default behavior that some of us have already designed systems around. As @garylai said, there is a way to send input to an interactive program. Would it be better to revert this and change the implementation so that it helps your specific use case without adding complexity to existing use cases that have been solved?

kohtenko:
@ryechus I now see that it breaks the existing integrations. Although from my perspective those use cases used sort of a loophole, because Spaceship was designed for interactive console initially. I'm not saying those use cases are wrong, in fact I love the idea and it would be great if they could make their way to fastlane plugins or a documentation reference. I didn't have a clue someone uses a separate service to automate this. If it's open source we can set the ENV inside it so that everybody wins 🎉

PR #17861, opened 2020-12-27
[CI] support Ruby 2.7 on Circle CI

ainame:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

It's been a year since Ruby 2.7 was released and then "3.0" was just released on the last Christmas day. As fastlane is a large size Ruby project, it's inevitable to continue following the latest version of Ruby. However, whilst fastlane does seem to be running with Ruby 2.7, it hasn't officially supported a bunch of unit tests running on Circle CI yet, which makes us feel less confident to develop fastlane for new Ruby versions. There is an attempt to make CI work with Ruby 2.7 https://github.com/fastlane/fastlane/pull/16407 but It has never had testing working all successfully yet. 

This PR fixes all the failures in unit testing with Ruby 2.7 and enables CI settings to run tests with it. This PR hasn't addressed warnings or deprecation messages appearing when running fastlane command yet. (See details here https://github.com/fastlane/fastlane/issues/16342)

### Description

You can see what exactly failing with Ruby 2.7 here https://gist.github.com/ainame/4d412e39fc69eb1f8aec153d3cd206c7

There are two kinds of failures due to version bumps of Ruby -

1. Syntax error message change https://github.com/fastlane/fastlane/commit/85f95f8cfcae51c5dd9d11467642fb76a505d0f5
2. Behaviour change in `Shellwords.escape` https://github.com/fastlane/fastlane/commit/c1c2804143b98e40745c29fdfaa5a465617da4b8

**1. Syntax error message change**

There was very minor change in error message that cause failures in unit testing. For example, this broken ruby code can show differences. 
 
```ruby
lane :beta do
  sigh(app_identifier: "hi"
end
```

```
% rbenv local 2.6.6
% ruby -c a.rb
a.rb:3: syntax error, unexpected end, expecting ')'
% rbenv local 2.7.2
% ruby -c a.rb
a.rb:3: syntax error, unexpected `end', expecting ')'
```

"end" in the message is now just quated! 😄 I added this to the regex pattern in testing.

**2. Behaviour change in `Shellwords.escape`**

[...]

Issue #16536, opened 2020-05-30
Setting Ruby version was does not work for Homebrew (in `xecute tests on macOS (Xcode 9.4.1, Ruby 2.4)`)


janpio:
See this CI run: https://app.circleci.com/pipelines/github/fastlane/fastlane/318/workflows/5bc3c6da-b4dc-4938-b987-3297b95978a9/jobs/46397/steps

![image](https://user-images.githubusercontent.com/183673/83335573-1e9f7f80-a2ae-11ea-9012-205d1d49df5e.png)
![image](https://user-images.githubusercontent.com/183673/83335579-2b23d800-a2ae-11ea-9afe-9034e476abca.png)

I do not understand why Homebrew tries to use Ruby 2.3 here.

janpio:
Hmm, maybe some cache that is wrong and does not take the version into account? That would also kind of explain why this happens only _sometimes_ I think.

janpio:
Nope, that's not it: https://app.circleci.com/pipelines/github/fastlane/fastlane/319/workflows/0cc8a650-59f6-4250-bb40-d65a3be169b4/jobs/46405/steps

janpio:
Turns out that "Updating Homebrew..." always installs a Ruby 2.6:

Ruby 2.4:
https://app.circleci.com/pipelines/github/fastlane/fastlane/322/workflows/56a94ed8-a0ee-402d-b877-5909f898261a/jobs/46434/steps
```
Untapping caskroom/versions...
 
Untapped 199 casks (360 files, 55.4MB).
 
Updating Homebrew...
 
==> Downloading https://homebrew.bintray.com/bottles-portable-ruby/portable-ruby-2.6.3.mavericks.bottle.tar.gz
 
######################################################################## 100.0%
 
==> Pouring portable-ruby-2.6.3.mavericks.bottle.tar.gz
 
==> Auto-updated Homebrew!
 
Updated 3 taps (homebrew/core, homebrew/cask and adoptopenjdk/openjdk).
 
```

Ruby 2.5:
https://app.circleci.com/pipelines/github/fastlane/fastlane/322/workflows/56a94ed8-a0ee-402d-b877-5909f898261a/jobs/46430/steps
```
Untapping caskroom/versions...
 
Untapped 161 casks (209 files, 317.4KB).
 
Updating Homebrew...
 
[...]

PR #15532, opened 2019-10-22
[fastlane] warn user during load_plugins when failure to find action

joshdholtz:
Fixes https://github.com/fastlane/fastlane/issues/15506
Fixes https://github.com/fastlane-community/fastlane-plugin-firebase_app_distribution/issues/15

And bonus fix for failing CI with https://github.com/Homebrew/brew/issues/6112#issuecomment-520950707

### Motivation and Context
Give users some more help when plugins fail to load without an actions

### Description
- Add a message with a link to new doc that also explains
- Also added a new optional parameter onto `load_plugins` to prevent the setting of `ENV["FASTLANE_ENV_PRINTER"]` which was causing test failures due to the environment variables never being cleared out

Screen Shot 2019-10-22 at 2 54 43 PM

PR #17863, opened 2020-12-27
[CI]  Fix homebrew issue on CircleCI

ainame:
### Checklist

### Motivation and Context

While working on #17861, I realised that a bug was introduced in `.circleci/config.yml` in #17708 (see below).  The behaviour of homebrew seemed to be changed in Homebrew/brew#8883 so I suppose something was changed in cache compatibility around that as well🤔  The temporary solution made at https://github.com/fastlane/fastlane/pull/17708/commits/287b4f607c26df957a2b0b1470609d8c41b49fdd doesn't work for subsequent builds after all.

This is the error I mentioned.

```
#!/bin/bash --login -eo pipefail
mkdir -p ~/test-reports
git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core fetch --unshallow
git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask fetch --unshallow
brew update
brew untap caskroom/versions || true
brew install shellcheck
bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
fatal: --unshallow on a complete repository does not make sense

Exited with code exit status 128
CircleCI received exit code 128
```

https://app.circleci.com/pipelines/github/fastlane/fastlane/1796/workflows/e077559c-1757-447c-a66f-0fc24413fb80/jobs/58879

### Description

There are 3 changes in my PR. The main change is 1. to fix the error but the other two are good to do together since I investigated all the related issue around that.

1. Remove `brew update` called unnecessarily
2. Stop caching homebrew updates
3. Remove `brew untap caskroom/versions` workaround 

I will explain the details one by one.

**1. Remove `brew update` called unnecessarily**

`brew update` was actually what caused the error. It was added in https://github.com/fastlane/fastlane/pull/16537 but the original issue https://github.com/fastlane/fastlane/issues/16536 seemed to be resolved by dropping "Xcode 9.4.1, Ruby 2.4" test condition.  Since Homebrew normally tries to update itself before running "install" subcommand, there is no reason to do it. Importantly "brew install"  without `brew update` don't get the error previously we got.

(I couldn't find any bug reports related to the issue within homebrew's project, but hope this change is just okay🙏 )

**2. Stop caching homebrew updates** 

The cache for homebrew was introduced here https://github.com/fastlane/fastlane/pull/11489. I suppose it was okay and did speed up builds at that moment but nowadays saving/restoring homebrew repo's cache is not that cheap as the formula repo grew (just like CocoaPods/Specs). It comes with gigabytes of data, which takes time anyway. (See also @rogerluan's attempt in https://github.com/fastlane/fastlane/pull/17870) 

In fact, in the official document, caching homebrew is not mentioned at all. (And instead, they suggest `HOMEBREW_NO_AUTO_UPDATE` when the install versions don't matter) https://circleci.com/docs/2.0/testing-ios/#optimizing-homebrew

So I think removing cache for homebrew is the simplest and pragmatic way to keep sane at this moment. Possibly we can use `HOMEBREW_NO_AUTO_UPDATE` if needed in future. 

----

Caching/Restoring cache takes time for "shallow" clone. Without them, "Setup Build" will just take 2-3 mins.

Screenshot 2020-12-30 at 04 48 51

https://app.circleci.com/pipelines/github/fastlane/fastlane/1816/workflows/4f02d12a-4a32-4b31-850d-db9aaf9e6a40/jobs/59037

This can be even worse with "unshallowed" clone.
https://github.com/fastlane/fastlane/pull/17870

----

**3. Remove `brew untap caskroom/versions` workaround**

`brew untap` is an also workaround added here https://github.com/fastlane/fastlane/pull/15532 but the original issue in homebrew itself was resolved later https://github.com/Homebrew/brew/pull/6703 so there is not really a reason to keep it. 

#### Build time

As a result, the build time is much faster than "unshallow" one without any issue. 

before|after
--|--
Screenshot 2020-12-30 at 04 41 48|Screenshot 2020-12-30 at 04 40 56

### Testing Steps

See if CI passes in all conditions. And see the last empty commit if the other cache is working fine and build time is not terrible.

### Notes

While working on CI stuff, I noticed a few issues existing. I will work on fixing them in another PR.

* Danger stopped working at some point (Already asked Josh)
* Bundler's cache doesn't effectively work since it's cached by CPU arch and not Ruby version...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,6 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
-            brew untap caskroom/versions || true
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
       - *cache_save_homebrew
@@ -172,7 +171,6 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
-            brew untap caskroom/versions
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
       - *cache_save_homebrew
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,12 +31,12 @@ aliases:
     restore_cache:
       name: cache | restore homebrew
       keys:
-        - v3-homebrew-{{ epoch }}
-        - v3-homebrew
+        - v4-homebrew-{{ epoch }}
+        - v4-homebrew
   - &cache_save_homebrew
     save_cache:
       name: cache | store homebrew
-      key: v3-homebrew-{{ epoch }}
+      key: v4-homebrew-{{ epoch }}
       paths:
         - "/usr/local/Homebrew"
 
@@ -87,9 +87,6 @@ jobs:
           name: Setup Build
           command: |
             mkdir -p ~/test-reports
-            git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core fetch --unshallow
-            git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask fetch --unshallow
-            brew update
             brew untap caskroom/versions || true
             brew install shellcheck
             bundle check --path .bundle || bundle install --jobs=4 --retry=3 --path .bundle
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,19 +27,6 @@ aliases:
       paths:
         - ".bundle"
 
-  - &cache_restore_homebrew
-    restore_cache:
-      name: cache | restore homebrew
-      keys:
-        - v4-homebrew-{{ epoch }}
-        - v4-homebrew
-  - &cache_save_homebrew
-    save_cache:
-      name: cache | store homebrew
-      key: v4-homebrew-{{ epoch }}
-      paths:
-        - "/usr/local/Homebrew"
-
   - &cache_restore_rubocop
     restore_cache:
       name: cache | restore rubocop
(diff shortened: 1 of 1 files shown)


ainame:
I will continue this on another day 🙇

.circleci/config.yml
```diff
@@ -31,12 +31,12 @@ aliases:
     restore_cache:
       name: cache | restore homebrew
       keys:
-        - v3-homebrew-{{ epoch }}
-        - v3-homebrew
+        - v4-homebrew-{{ .Branch }}-{{ epoch }}
```
ainame, line 34: This is a temporary key for debugging to not share caches with other else branches.

.circleci/config.yml
```diff
@@ -77,23 +64,21 @@ jobs:
...
-      - *cache_restore_homebrew
       - *set_ruby
       - run:
           name: debug | ruby version
           command: |
             ruby -v
+      - run:
+          name: debug | brew version
+          command: |
+            brew -v
```
ainame, line 75: I added this to track any future issues related to homebrew... There was no info about this in CircleCI's page

https://circle-macos-docs.s3.amazonaws.com/image-manifest/v4250/index.html

rogerluan (approved):
💯 💯  This is amazing! Can't wait to have this in master so we can make all PRs green again 🚀 
Thanks for taking time to fix this!

If (when) this gets approved and merged, I'll close my other similar PR 🙇 🙏

joshdholtz (approved):
This is 🔥 Thank you so much for properly fixing this 😊

PR #17890, opened 2021-01-01
ci: add xcode 12.3 ruby 2.7

chenrui333:
Since now it is [buildable with ruby 2.7](https://github.com/Homebrew/homebrew-core/pull/68104), adding ruby 2.7 into CI.

Also update some other xcode versions as well.

cc @joshdholtz

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,18 +95,24 @@ jobs:
           name: Post Test Results to GitHub
           command: bundle exec danger || echo "danger failed"
           when: always  # Run this even when tests fail
-  'Execute tests on macOS (Xcode 11.0.0, Ruby 2.5)':
+  'Execute tests on macOS (Xcode 11.0, Ruby 2.5)':
     <<: *tests_macos
     macos:
-      xcode: '11.0.0'
+      xcode: '11.0'
     environment:
       _RUBY_VERSION: '2.5'
-  'Execute tests on macOS (Xcode 11.4.0, Ruby 2.6)':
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -95,18 +95,24 @@ jobs:
...
     <<: *tests_macos
     macos:
-      xcode: '11.4.0'
+      xcode: '11.7'
     environment:
       _RUBY_VERSION: '2.6'
+  'Execute tests on macOS (Xcode 12.3, Ruby 2.7)':
+    <<: *tests_macos
+    macos:
+      xcode: '12.3'
```
ainame, line 113: Since this PR failed with 12.3, this might need to be "12.3.0", I suppose...? 
https://discuss.circleci.com/t/xcode-12-3-release/385701

```suggestion
      xcode: '12.3.0'
```

ainame (commented):
In my PR #17861, adding 2.7 support was already done (but not merged yet). We can't enable Ruby 2.7 settings without fixing a few test cases failing with Ruby 2.7. In #17861, I've already fixed failure tests alongside adding Xcode 12.3.0 x Ruby 2.7 pattern. 

So I would say this is duplication PR, but bumping Xcode versions are interesting changes. Do you see any clear benefits on bumping those versions? If so I'd like to apply the same change to my PR.

chenrui333:
`clear benefits` would be more swift related, and I do see sometimes.

chenrui333:
yeah, I can close my PR since it is a duplicate. :)

PR #17892, opened 2021-01-01
[CI] improve bundler's cache on CircleCI

ainame:
### Checklist

### Motivation and Context

Note: There is another my PR #17861 that conflicts to this PR. Either way, I will resolve the conflict after one gets merged🙇 

----

While working on #17861 and #17863, I realised that bundler cache doesn't work. This is because the cache key is just wrong from the beginning to be used across all Ruby environments.

In this PR, I fixed the cache-key to make caching work properly. To do so, I also migrated to `.circleci/config.yml` to the latest version "2.1" to describe parameterised builds.

If the cache works correctly, build time would be shortened by 1-2 mins or so presumably... 

### Description

> v2-gems-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile" }}-{{ checksum "fastlane.gemspec" }}

This is the cache key to store/restore installed gems. This is just wrong in the following reasons. 

* The cache (un)intentionally will be used from all the Ruby version as long as "arch" is the same
    * `bundler` keep installed gems separated by Ruby version, which means the cache works only for one of Ruby versions tested
* The cache will be available only within the same branch's builds
* Appending checksums of "Gemfile" and "gemspec" isn't accurate enough to serve all the required dependencies precisely 
   * Eventually `bundler` resolves dependencies but it's not 100% efficient (cache miss)

(FYI: `arch` matters if we use a gem that has native extensions.)

I'm going to change the key to this. 

> v3-gems-{{ arch }}-<< parameters.ruby_version >>-{{ checksum "Gemfile.lock" }}

This way 

- The cache will be kept by each Ruby version or docker image
- The cache will be shared across all the branches
- Builds can pick up exactly correct cache according to `Gemfile.lock`

To do this, I had to switch CircleCI's config format version to "2.1" as well. It allows you to reuse a "job" definition and run it against different configurations using parameters (a.k.a parameterised builds). See https://circleci.com/docs/2.0/reusing-config/

### Testing Steps

See if "bundle install" step is done within a few seconds across all the Ruby version/environments to install "bundler" gem.

Screenshot 2021-01-03 at 00 17 52

https://app.circleci.com/pipelines/github/fastlane/fastlane/1899/workflows/b9988e82-2674-4ba4-abe1-096a051300a3

In the above build, the following links are the corresponding "bundle install" step's logs.

* Ruby 2.6 https://app.circleci.com/pipelines/github/fastlane/fastlane/1899/workflows/b9988e82-2674-4ba4-abe1-096a051300a3/jobs/59652/parallel-runs/0/steps/0-109
* Ruby 2.5 https://app.circleci.com/pipelines/github/fastlane/fastlane/1899/workflows/b9988e82-2674-4ba4-abe1-096a051300a3/jobs/59651/parallel-runs/0/steps/0-109
* Ruby 2.4 https://app.circleci.com/pipelines/github/fastlane/fastlane/1899/workflows/b9988e82-2674-4ba4-abe1-096a051300a3/jobs/59658/parallel-runs/0/steps/0-109
* Ubuntu https://app.circleci.com/pipelines/github/fastlane/fastlane/1899/workflows/b9988e82-2674-4ba4-abe1-096a051300a3/jobs/59653/parallel-runs/0/steps/0-106


.circleci/config.yml
```diff
@@ -47,17 +49,29 @@ aliases:
         echo "ruby-${_RUBY_VERSION}" > ~/.ruby-version
         echo 'chruby ruby-${_RUBY_VERSION}' >> ~/.bash_profile
 
-version: 2
+  - &bundle_install
+    run:
+      name: bundle install
+      command: |
+        gem install bundler -v $(cat Gemfile.lock | tail -1 | tr -d " ")
+        bundle config set --local path .bundle
```
ainame, line 57: This is to fix this deprecation warning.

> [DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path '.bundle'`, and stop using this flag

.circleci/config.yml
```diff
@@ -47,17 +49,29 @@ aliases:
         echo "ruby-${_RUBY_VERSION}" > ~/.ruby-version
         echo 'chruby ruby-${_RUBY_VERSION}' >> ~/.bash_profile
 
-version: 2
+  - &bundle_install
+    run:
+      name: bundle install
+      command: |
+        gem install bundler -v $(cat Gemfile.lock | tail -1 | tr -d " ")
```
ainame, line 56: This way CI can always use `bundler` that has exactly the same version as we used to generate "Gemfile.lock"

Gemfile.lock
```diff
@@ -0,0 +1,366 @@
+PATH
```
ainame, line 1: ### Check `Gemfile.lock` in

Checking in `Gemfile.lock` is now officially recommended in both apps and gems. It won't affect end-users but benefits for gem developers. 

> As a result, you SHOULD check your Gemfile.lock into version control, in both applications and gems. If you do not, every machine that checks out your repository (including your production server) will resolve all dependencies again, which will result in different versions of third-party code being used if any of the gems in the Gemfile(5) or any of their dependencies have been updated.
>
> When Bundler first shipped, the Gemfile.lock was included in the .gitignore file included with generated gems. Over time, however, it became clear that this practice forces the pain of broken dependencies onto new contributors, while leaving existing contributors potentially unaware of the problem. Since bundle install is usually the first step towards a contribution, the pain of broken dependencies would discourage new contributors from contributing. As a result, we have revised our guidance for gem authors to now recommend checking in the lock for gems.

https://bundler.io/man/bundle-install.1.html#THE-GEMFILE-LOCK

Gemfile.lock
```diff
@@ -0,0 +1,366 @@
...
+      terminal-table
+      xcodeproj
+    xcpretty (0.3.0)
+      rouge (~> 2.0.7)
+    xcpretty-travis-formatter (1.0.1)
+      xcpretty (~> 0.2, >= 0.0.7)
+    yard (0.9.26)
+
+PLATFORMS
+  ruby
```
ainame, line 336: FYI: 

Apparently, you need to explicitly declare this by the below command otherwise `bundle install` will fail if you create `.lock` file on macOS and run `bundle install` on Ubuntu.

```sh
bundle lock --add-platform ruby
```

This is what you get instead if you were to generate `Gemfile.lock` without the above command. This would not make sense to be compatible on Ubuntu obviously.

```
 PLATFORMS
-  ruby
+  x86_64-darwin-19
```

https://bundler.io/v2.2/man/bundle-lock.1.html

Gemfile.lock
```diff
@@ -0,0 +1,366 @@
...
+  rubocop (= 0.49.1)
+  rubocop-require_tools (>= 0.1.2)
+  sinatra (~> 2.0.8)
+  webmock (~> 3.8)
+  xcode-install (>= 2.6.7)
+  xcov (~> 1.4.1)
+  yard (~> 0.9.11)
+
+BUNDLED WITH
+   2.2.4
```
ainame, line 366: This doesn't mean we force end-users to use this version of `bundler`, of course.

rogerluan: Sorry, it's not obvious to me how this doesn't force users to use this version of bundler 😬 could you explain? 😅 
I've had trouble with this in the past where I had a different bundler version than the one specified by a Gemfile.lock and I ended up either changing the Gemfile.lock or changing my bundler version (not sure which way I went, it was a long time ago)

rogerluan (commented):
Such a great and welcome refactor of the .circleci config file! It's much easier to follow and way more readable imo, thank you! 🤩 

Also great job on fixing the caching strategy 🚀

Gemfile.lock
ainame: @rogerluan 

`Gemfile.lock`, in this case, is not something that affects ".gem" package to be installed at all. All needed to make ".gem" package is "gemspec" file. This is for sharing what fastlane contributors install to their own environment. Again, that way we have the consistency of behaviour in unit testing and can be likely to replicate the same bug in others' environment. (Importantly this applies to CI😄)

Regarding your issue, I suppose that the issue you had was due to `bundler`' versioning issue between the major versions; 1 and 2. For example, if you have only `bundler` **1** installed locally, `bundle install` against this lock file will fail. 

<details> 

```
% gem install bundler -v 1.17
Fetching: bundler-1.17.0.gem (100%)
Successfully installed bundler-1.17.0
Parsing documentation for bundler-1.17.0
Installing ri documentation for bundler-1.17.0
Done installing documentation for bundler after 4 seconds
1 gem installed
% bundle install
You must use Bundler 2 or greater with this lockfile.
```

</details>

This is what I saw on my testing. Only the mentioned case above fails. The differences in minor version didn't affect any for `bundle install` (`bundle` will just warn the difference in the console) 

Installed version  \ Locked version| 1.x | 2.x |
--|--|--
1.x | 🆗 | 🙅‍♂️
2.x | 🆗  ( lock file will be automatically migrated to 2.x) | 🆗  

The good thing is that `bundler` 1 and 2 can coexist so if you have both installed, you don't have to remove either one for a project using another one. 

> Version autoswitching means that Bundler will automatically run the correct 1.x or 2.x version based on your lockfile. This means you can have both Bundler 1.17.3 and Bundler 2.0.0 installed at the same time, and each application will use the version saved in your lockfile when you run bundle install.

https://bundler.io/blog/2019/01/03/announcing-bundler-2.html

I would say there are no reasons to start using `bundler` **1** since it's been two years from the release of **2**. Ruby 2.7 has got `bunder` 2.x version by default. 

https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/

--------

Does that make sense...?

rogerluan: Ah gotcha, yeah that makes total sense! 💪

rogerluan (approved):
🚀 🚀

ainame:
I will rebase this PR onto the master having #17861 merged!

ainame:
@joshdholtz This PR would reduce build time significantly from 28-32 mins to 16-18 mins. This might help your work to merge PRs😄

joshdholtz (approved):
This is so 🔥🔥🔥🔥 Thank you for fixing this!

PR #17913, opened 2021-01-07
[action] git_commit - skip commit if git status is clean for paths

felginep:
### Checklist

### Motivation and Context

Reverts the commit ef1ccf17 and fix a regression introduced in #17804.

There is an issue when the `git_commit` lane is executed alone (meaning without `git_add` before).
In this case, it is not possible to run the sh command `git commit` because there is no staged files and `git --no-pager diff --name-only --staged` always returns an empty string.

### Description

The proposed solution is to run `git status #{paths} --porcelain` to check if the git status is clean, **only** for the paths we want to commit. This way, even if there are other files modified, the `git commit` can be skipped.

In practice, there are two cases:

1. *fileA* is modified (meaning `git status fileA --porcelain` will return a non empty string)
1.1 if we call `git_commit(path: "fileA", allow_nothing_to_commit: true)` -> the commit is created
1.2 if we call `git_commit(path: "fileA", allow_nothing_to_commit: false)` -> the commit is created

2. *fileA* is not modified (meaning `git status fileA --porcelain` will return an empty string)
2.1 we call `git_commit(path: "fileA", allow_nothing_to_commit: true)` -> the commit is skipped (even if there are other files modified, i.e. `git status --porcelain` does not return an empty string)
2.2 we call `git_commit(path: "fileA", allow_nothing_to_commit: false)` -> an error occurs because *fileA* can't be staged

### DIFF
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -11,9 +11,9 @@ module Fastlane
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
         if params[:allow_nothing_to_commit]
-          nothing_staged = Actions.sh("git --no-pager diff --name-only --staged").empty?
-          UI.success("Nothing staged to commit ✅.") if nothing_staged
-          return if nothing_staged
+          repo_clean = Actions.sh("git status #{paths} --porcelain").empty?
+          UI.success("Nothing to commit, working tree clean ✅.") if repo_clean
+          return if repo_clean
         end
 
         command = "git commit -m #{params[:message].shellescape} #{paths} #{skip_git_hooks}".strip
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -11,6 +11,10 @@ module Fastlane
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
         if params[:allow_nothing_to_commit]
+          # Here we check if the path passed in parameter contains any modification
+          # and we skip the `git commit` command if there is none.
+          # That means you can have other files modified that are not in the path parameter
+          # and still make use of allow_nothing_to_commit.
           repo_clean = Actions.sh("git status #{paths} --porcelain").empty?
           UI.success("Nothing to commit, working tree clean ✅.") if repo_clean
           return if repo_clean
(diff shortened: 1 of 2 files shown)


fastlane/spec/actions_specs/git_commit_spec.rb
```diff
@@ -54,6 +54,13 @@
       end
 
       it "generates the correct git command when configured to allow nothing to commit and there are changes to commit" do
+        allow(Fastlane::Actions).to receive(:sh)
+          .with("git status ./fastlane/README.md --porcelain")
+          .and_return("M  ./fastlane/README.md")
+        allow(Fastlane::Actions).to receive(:sh)
```
ainame, line 60:
```suggestion
        expect(Fastlane::Actions).to receive(:sh)
```

It feels like this must be called in the given scenario so `expect` makes sure future change will also behave the same. Though it is checked by below assertion, this would produce a slightly better error message when failing.

fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -11,9 +11,9 @@ def self.run(params)
         skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
 
         if params[:allow_nothing_to_commit]
-          nothing_staged = Actions.sh("git --no-pager diff --name-only --staged").empty?
-          UI.success("Nothing staged to commit ✅.") if nothing_staged
-          return if nothing_staged
+          repo_clean = Actions.sh("git status #{paths} --porcelain").empty?
```
ainame, line 14: Can you please add a comment about this behaviour here, and at the description for "allow_nothing_to_commit" param, which "unstaged" changes are taken account and this way allows you to keep the repo dirty without failing?

We missed the behaviour in the last PR so it's definitely worth documenting. Ideally, we should cover it in unit tests but the way we test doesn't allow us to describe the exact situation for it due to `sh` being mocked or the difficulty to prepare a complete git repo just for a minor action. I can only come up leaving a comment as a way to avoid regressions for now.

ainame (changes requested):
Thank you very much for doing this.
I'd like to request you to add a way to avoid future regressions. Just a code comment + description updates for `allow_nothing_to_commit` to clarify the behaviours is fine🙇 

Otherwise LGTM 🚀

fastlane/spec/actions_specs/git_commit_spec.rb
felginep: You are right, I made the change

ainame (approved):
Well done🎉

@joshdholtz will double-check and merge this PR when he can!

joshdholtz (approved):
I like the tests and comment here! Thanks for fixing! Teamwork ❤️

Issue #17677, opened 2020-11-26
Download DSYMs action broken


gasigo:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I have a Bitrise workflow to download DSYM files from the App Store Connect and upload them to Firebase. The step to download the DSYM files is not working anymore since last week.

##### Command executed
`Fastlane::Actions::DownloadDsymsAction.run username:ENV["ASC_APPLE_ID"], app_identifier:ENV["BUNDLE_ID"], team_id:ENV["TEAM_ID"], version:"latest"`

##### Complete output when running fastlane, including the stack trace and command used
```
[09:43:25]: Login to App Store Connect ([REDACTED])
[09:43:28]: Login successful
/Users/vagrant/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/fastlane-2.168.0/spaceship/lib/spaceship/connect_api.rb:79:in `map': undefined method `to_sym' for nil:NilClass (NoMethodError)
	from /Users/vagrant/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/fastlane-2.168.0/fastlane/lib/fastlane/actions/download_dsyms.rb:37:in `run'
	from /var/folders/6q/wgy6jtp12w5gzgm9lzcglpqw0000gn/T/bitrise928764210/step_src/ruby_content_file.rb:4:in `<main>'
--- finished ---
```

rogerluan:
Hey @gasigo 👋 

You're not sending `platform` to your `download_dsyms` action. You shouldn't need to invoke "run" explicitly - let `download_dsyms` action build its parameters on its own by simply calling `download_dsyms(<your params here>)`.

`platform` has a default value, but since you're not invoking the action the designed way, it's not initializing the default parameters 🤷 

Three ways to fix this problem:
a. call `download_dsyms(< your params here>)`
b. pass `platform:"ios"` parameter
c. set `DOWNLOAD_DSYMS_PLATFORM` env var to "ios"

[...]

PR #17148, opened 2020-09-03
[actions] fix download_dsyms, app_store_build_number, and test_flight_build_number

joshdholtz:
### Motivation and Context
Fixes #17140
Fixes #17132

### Description
- Replace deprecated `Spaceship::Tunes` calls with `Spaceship::ConnectAPI`
  - `app.live_version` to `app.get_live_app_store_version`
  - `app.edit_version` to `get_edit_app_store_version`

### Testing Steps

Update `Gemfile` and run `bundle update fastlane`, `bundle update`, or `bundle install`

```rb
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "joshdholtz-fix-broken-live-version-edit-version-actions"
```

PR #17936, opened 2021-01-09
[download_dsyms][docs] change `platform` option optionality (to nonnull) since passing nil would already crash

rogerluan:
### Motivation and Context

This PR https://github.com/fastlane/fastlane/pull/17148 made `download_dsyms` 'platform' parameter required, but the option was still marked as optional.

Discussion: https://github.com/fastlane/fastlane/issues/17677#issuecomment-757344123

### Description
With the changes introduced in #17148, if platform is passed as `nil`, it will actually crash here:

https://github.com/fastlane/fastlane/blob/e2ff4990a44f28fa7b25d745666007dc8e315577/fastlane/lib/fastlane/actions/download_dsyms.rb#L37

Because `Spaceship::ConnectAPI::Platform.map` doesn't actually accept a `nil` value.

If users omit the platform, the default value should be picked up. For this reason, this is a rare corner case since people would rarely **actively** pass a "nil" value, but if anyone did, they'd get a cryptic crash log instead of a nicer one 😊 

### Testing Steps

```ruby
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "rogerluan-update-download-dsyms-platform-optionality"
```

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -285,7 +285,6 @@ module Fastlane
                                        short_option: "-p",
                                        env_name: "DOWNLOAD_DSYMS_PLATFORM",
                                        description: "The app platform for dSYMs you wish to download (ios, appletvos)",
-                                       optional: true,
                                        default_value: :ios),
           FastlaneCore::ConfigItem.new(key: :version,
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Ahhhhh, good catch!! 😊 Thank you thank you!

Issue #17879, opened 2020-12-29
App Store Connect create function fails


Adam0101:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
When calling Spaceship::ConnectAPI::App.create it crashes with an unexpected response.

##### Command executed
Spaceship::Tunes::Application.create!(name: title,
                                  primary_language: "English",
                                               sku: sku,
                                         bundle_id: bundleID)

##### Complete output when running fastlane, including the stack trace and command used
Spaceship::Tunes::Application.create!(name: title,
                                  primary_language: "English",
                                               sku: sku,
                                         bundle_id: bundleID)

Traceback (most recent call last):
	3: from ./createAppContainer.rb:46:in `<main>'
	2: from /Library/Ruby/Gems/2.6.0/gems/fastlane-2.170.0/spaceship/lib/spaceship/tunes/application.rb:96:in `create!'
	1: from /Library/Ruby/Gems/2.6.0/gems/fastlane-2.170.0/spaceship/lib/spaceship/tunes/tunes_client.rb:296:in `create_application!'
/Library/Ruby/Gems/2.6.0/gems/fastlane-2.170.0/spaceship/lib/spaceship/client.rb:732:in `parse_response': {"data"=>nil, "messages"=>{"warn"=>nil, "error"=>["Not Found"], "info"=>nil}, "statusCode"=>"ERROR"} (Spaceship::UnexpectedResponse)

<details>
  <pre>  </pre>
</details>

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 11.1                                                           |
| Ruby                        | 2.6.3                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.24.3 (Apple Git-128)                             |
| Installation Source         | /usr/local/bin/fastlane                                        |
| Host                        | macOS 11.1 (20C69)                                             |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib |
[...]

PR #17938, opened 2021-01-10
[spaceship][docs] update spaceship app store connect docs

rogerluan:

### Motivation and Context

This was exposed as an issue here: https://github.com/fastlane/fastlane/issues/17879#issuecomment-754101484

### Description

Simple docs updating. Removed a comment that said that something was not working (but it is), and update param name.

### DIFF
--- a/spaceship/docs/AppStoreConnect.md
+++ b/spaceship/docs/AppStoreConnect.md
@@ -60,9 +60,8 @@ Spaceship::ConnectAPI::App.all.collect do |app|
 end
 
 # Create a new app
-// Not working yet
 app = Spaceship::ConnectAPI::App.create(name: "App Name",
-                                        primary_language: "English",
+                                        primary_locale: "English",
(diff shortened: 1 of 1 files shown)


rogerluan:
Actually I'm not sure whether the "not working" part should be removed. Looks like it doesn't work with ASC, but the same function does work with regular authentication credentials. 🤔

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
🔥

spaceship/docs/AppStoreConnect.md
```diff
@@ -60,11 +60,11 @@ Spaceship::ConnectAPI::App.all.collect do |app|
 end
 
 # Create a new app
-// Not working yet
+# Currently only works with Apple ID login (not API Key)
```
rogerluan, line 63: Cheers! 🎉 much more clear now!

PR #17981, opened 2021-01-16
[fastlane_core][Fastlane.swift][sigh][pem][upload_app_privacy_details_to_app_store] consider only truthy values when evaluating conflicting options

### LINKED ISSUE #17343: [Fastlane.Swift] get_provisioning_profile error: You can't enable both :developer_id and :adhoc
### New Issue Checklist

### Issue Description
Using sample lane
```
    func fetcherLane() {
        getProvisioningProfile(
            appIdentifier: "XXX",
            username: "XXX",
            teamId: "XXX",
            teamName: "XXX",
            provisioningName: "XXX",
            ignoreProfilesWithDifferentName: true,
            certOwnerName: nil,
            skipFetchProfiles: true,
            skipCertificateVerification: true,
            readonly: true
        )
    }
```
You'll get error
```
You can't enable both :developer_id and :adhoc
```
similar to [May 2018 Issue 12506](https://github.com/fastlane/fastlane/issues/12506)

##### Command executed
```
bundle exec fastlane fetcher
```

<details>
</details>

### Environment

<details>
  <pre> <details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                            |
| --------------------------- | ------------------------------------------------ |
| OS                          | 10.15.6                                          |
| Ruby                        | 2.6.6                                            |
| Bundler?                    | true                                             |
| Git                         | git version 2.24.3 (Apple Git-128)               |
| Installation Source         | ~/.rvm/gems/ruby-2.6.6/bin/fastlane              |
| Host                        | Mac OS X 10.15.6 (19G2021)                       |
| Ruby Lib Dir                | ~/.rvm/rubies/ruby-2.6.6/lib                     |
| OpenSSL Version             | OpenSSL 1.1.1g  21 Apr 2020                      |
| Is contained                | false                                            |
| Is homebrew                 | false                                            |
| Is installed via Fabric.app | false                                            |
| Xcode Path                  | /Applications/Xcode-12.0.app/Contents/Developer/ |
| Xcode Version               | 12.0.1                                           |

### System Locale

### fastlane files:

<details><summary>`./fastlane/Fastfile.swift`</summary>

```ruby
// This file contains the fastlane.tools configuration
// You can find the documentation at https://docs.fastlane.tools
//
// For a list of all available actions, check out
//
//     https://docs.fastlane.tools/actions
//

import Foundation

class Fastfile: LaneFile {
    func fetcherLane() {
        getProvisioningProfile(
            appIdentifier: "XXX",
            username: "XXX", //environmentVariable(get: "APPLE_USERNAME"),
            teamId: "XXX",
            teamName: "XXX",
            provisioningName: "XXX",
            ignoreProfilesWithDifferentName: true,
            certOwnerName: nil,
            skipFetchProfiles: true,
            skipCertificateVerification: true,
            readonly: true
        )
    }
    func fetchAndResignLane(withOptions options:[String: String]?) {
        desc("Fetches the ipa file from `repository`, increments build number and resigns with `signing_identity`. Usage: `fastlane fetchAndResign app:befaring`")
        switch options?["app"] {
        case "befaring":
            xcversion(version: environmentVariable(get: "FL_XCODE_VERSION"))
            //let repository = "ssh://git@git.tech-01.net/nmb/ios_resigning.git"
            //sh(command: "rm -rf ../build && git clone --depth 1 \(repository) ../build")
            getProvisioningProfile(
                appIdentifier: "no.dnb.befaringspres",
                username: "theodore.partoza.gonzalez@dnb.no", //environmentVariable(get: "APPLE_USERNAME"),
                teamId: "YVFWMVJ9Y9",
                teamName: "DnB ASA",
                provisioningName: "BefaringProd",
                ignoreProfilesWithDifferentName: true,
                certOwnerName: nil,
                skipFetchProfiles: true,
                skipCertificateVerification: true,
                readonly: true
            )
            resign(
                ipa: "../build/Befaring.ipa",
                signingIdentity: "iPhone Distribution: DnB ASA (119504671)",
                provisioningProfile: "BefaringProd",
                version: "1" //environmentVariable(get: "BUILD_NUMBER")
            )

        default:
            echo(message: "Please specify app")
        }
	}
}

```
</details>

**No Appfile found**

### fastlane gems

| fastlane | 2.161.0 | ✅ Up-To-Date  |

### Loaded fastlane plugins:

**No plugins Loaded**

| Gem                       | Version      |
| ------------------------- | ------------ |
| did_you_mean              | 1.3.0        |
| executable-hooks          | 1.6.0        |
| bundler-unload            | 1.0.2        |
| rubygems-bundler          | 1.4.5        |
| bundler                   | 2.1.4        |
| rake                      | 13.0.1       |
| CFPropertyList            | 3.0.2        |
| public_suffix             | 4.0.6        |
| addressable               | 2.7.0        |
| atomos                    | 0.1.3        |
| aws-eventstream           | 1.1.0        |
| aws-partitions            | 1.376.0      |
| aws-sigv4                 | 1.2.2        |
| jmespath                  | 1.4.0        |
| aws-sdk-core              | 3.108.0      |
| aws-sdk-kms               | 1.38.0       |
| aws-sdk-s3                | 1.81.1       |
| babosa                    | 1.0.3        |
| claide                    | 1.0.3        |
| colored                   | 1.2          |
| colored2                  | 3.1.2        |
| highline                  | 1.7.10       |
| commander-fastlane        | 4.4.6        |
| declarative               | 0.0.20       |
| declarative-option        | 0.1.0        |
| digest-crc                | 0.6.1        |
| unf_ext                   | 0.0.7.7      |
| unf                       | 0.1.4        |
| domain_name               | 0.5.20190701 |
| dotenv                    | 2.7.6        |
| emoji_regex               | 3.0.0        |
| excon                     | 0.76.0       |
| multipart-post            | 2.0.0        |
| faraday                   | 1.0.1        |
| http-cookie               | 1.0.3        |
| faraday-cookie_jar        | 0.0.7        |
| faraday_middleware        | 1.0.0        |
| fastimage                 | 2.2.0        |
| gh_inspector              | 1.1.3        |
| jwt                       | 2.2.2        |
| memoist                   | 0.16.2       |
| multi_json                | 1.15.0       |
| os                        | 1.1.1        |
| signet                    | 0.14.0       |
| googleauth                | 0.13.1       |
| httpclient                | 2.8.3        |
| mini_mime                 | 1.0.2        |
| uber                      | 0.1.0        |
| representable             | 3.0.4        |
| retriable                 | 3.1.2        |
| google-api-client         | 0.38.0       |
| google-cloud-env          | 1.3.3        |
| google-cloud-errors       | 1.0.1        |
| google-cloud-core         | 1.5.0        |
| google-cloud-storage      | 1.29.0       |
| json                      | 2.3.1        |
| mini_magick               | 4.10.1       |
| plist                     | 3.5.0        |
| rubyzip                   | 2.3.0        |
| security                  | 0.1.3        |
| naturally                 | 2.2.0        |
| simctl                    | 1.6.8        |
| slack-notifier            | 2.3.2        |
| terminal-notifier         | 2.0.0        |
| unicode-display_width     | 1.7.0        |
| terminal-table            | 1.8.0        |
| tty-screen                | 0.8.1        |
| tty-cursor                | 0.7.1        |
| tty-spinner               | 0.9.3        |
| word_wrap                 | 1.0.0        |
| nanaimo                   | 0.3.0        |
| xcodeproj                 | 1.18.0       |
| rouge                     | 2.0.7        |
| xcpretty                  | 0.3.0        |
| xcpretty-travis-formatter | 1.0.0        |
| xcode-install             | 2.6.6        |
</details>

*generated on:* **2020-09-29**
</details> </pre>
</details>

rogerluan:
### Checklist

### Motivation and Context
Resolves #17343

### Description

A comment of mine, copied from the issue itself:

After analyzing this:

https://github.com/fastlane/fastlane/blob/4753b8dea25c9f9d9b7fc49101b4ee1e4590cb80/sigh/lib/sigh/options.rb#L12-L38

Turns out that in Swift, by omitting the arguments, since they're nonnull Bools, it forwards to ruby as non-optional `false`, so ruby detects as if the user was providing all the options.

### Possible Solutions

1. IMO a reasonable solution would be to add `next if option.value == false` all those `conflict_block`s. However, `option` doesn't seem to have the runtime "value" of the option - does it? 🤔

2. Another solution would be to make those 3 booleans optional. I'm not a fan of optional booleans 🙃 but we use them in other places where reasonable, not sure if there's a strong consensus about good practices around this topic.

3. Simply remove the validation if we agree it's nonsense and is just hurting us 🙈 

#### ❗ [New]❗

4. Similar to `1.`, but this solution is more generic. Instead of consider that condition only for this action/lane, we'd consider that all falsey values would be treated the same as if they were nil, passing the conflicting options validation.

This PR implements solution number `4`, although [it was discussed we could go forward with solution `2`](https://github.com/fastlane/fastlane/issues/17343#issuecomment-757188443).

----

This PR is a fix proposal. If we decide that this global/generic solution would hurt us more than help, I can revert this and apply solution number 2, assuming we only want to fix this issue here. 
However, after using a regex to search for similar issues in the codebase, I found out that there're actually 2 other actions where this issue can be reproduced:

1. `pem` (test by running `pem(appIdentifier: "a", username: "a", p12Password: "a")` in a Swift project)

image

2. `uploadAppPrivacyDetailsToAppStore` (test by running `uploadAppPrivacyDetailsToAppStore(username: "a", appIdentifier: "a")` in a Swift project)

image

So if we decide to move forward with solution `2`, keep in mind the fix will have to be individually applied in all those 3 actions ☝️ 

#### Solution Safety

Due to its nature, this solution affects all fastlane users. However, its behavior makes lanes/actions more permissive than they used to, instead of more restricting, hence why I believe it's safe. As a plus, I believe it could 

### Testing Steps

```ruby
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "rogerluan-fix-conflicting-options-swift"
```

To test and see the error (in master), you could run either of these lanes:

```swift
func crashScenario1Lane() {
    getProvisioningProfile(appIdentifier: "a", username: "a")
}

func crashScenario2Lane() {
    pem(appIdentifier: "a", username: "a", p12Password: "a")
}

func crashScenario3Lane() {
    uploadAppPrivacyDetailsToAppStore(username: "a", appIdentifier: "a")
}
```

(yeah you can leave in the fake values, it doesn't matter).

If you run the same swift lane in this branch, it will continue without that error, as expected (it will ask for "password for account 'a'" 😄 )

### DIFF
--- a/fastlane/spec/fixtures/plist/Info.plist
+++ b/fastlane/spec/fixtures/plist/Info.plist
@@ -40,7 +40,7 @@
 	<key>CFBundleVersion</key>
 	<string>0.9.14</string>
 	<key>LSApplicationCategoryType</key>
-	<string></string>
+	<string/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSLocationWhenInUseUsageDescription</key>
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -120,8 +120,8 @@ module FastlaneCore
           index = @available_options.find_index { |item| item.key == conflicting_option_key }
           conflicting_option = @available_options[index]
 
-          # ignore conflicts because value of conflict option is nil
-          next if @values[conflicting_option.key].nil?
+          # ignore conflicts because value of conflict option is falsey (nil, false, etc)
+          next unless @values[conflicting_option.key]
 
           if current.conflict_block
             begin
--- a/fastlane_core/spec/configuration_spec.rb
+++ b/fastlane_core/spec/configuration_spec.rb
@@ -29,8 +29,8 @@ describe FastlaneCore do
         expect do
           FastlaneCore::Configuration.create([FastlaneCore::ConfigItem.new(
             key: :cert_name,
-       env_name: "asdf",
-    description: "Set the profile name."
+            env_name: "asdf",
+            description: "Set the profile name."
           )], {})
         end.to raise_error("Do not let descriptions end with a '.', since it's used for user inputs as well for key :cert_name")
       end
@@ -40,12 +40,12 @@ describe FastlaneCore do
           expect do
             FastlaneCore::Configuration.create([FastlaneCore::ConfigItem.new(
               key: :cert_name,
-         env_name: "asdf"
+              env_name: "asdf"
             ),
-                                                FastlaneCore::ConfigItem.new(
-                                                  key: :cert_name,
-                                             env_name: "asdf"
-                                                )], {})
+            FastlaneCore::ConfigItem.new(
+              key: :cert_name,
+              env_name: "asdf"
+            )], {})
           end.to raise_error("Multiple entries for configuration key 'cert_name' found!")
         end
 
--- a/fastlane_core/spec/configuration_spec.rb
+++ b/fastlane_core/spec/configuration_spec.rb
@@ -64,7 +64,7 @@ describe FastlaneCore do
           end.to raise_error("Multiple entries for short_option '-f' found!")
         end
 
-        it "raises an error for unresolved conflict between options" do
+        it "raises an error for unresolved conflict between options with truthy values" do
           conflicting_options = [
             FastlaneCore::ConfigItem.new(key: :foo,
                                          conflicting_options: [:bar, :oof]),
@@ -80,6 +80,29 @@ describe FastlaneCore do
           expect do
             FastlaneCore::Configuration.create(conflicting_options, values)
           end.to raise_error("Unresolved conflict between options: 'foo' and 'bar'")
+        end
+
+        it "doesn't raise for conflicting options that have falsey values" do
+          conflicting_options = [
+            FastlaneCore::ConfigItem.new(key: :foo,
+                                         conflicting_options: [:bar, :oof, :meh]),
+            FastlaneCore::ConfigItem.new(key: :bar,
+                                         optional: true),
+            FastlaneCore::ConfigItem.new(key: :oof,
+                                         is_string: false),
+            FastlaneCore::ConfigItem.new(key: :meh,
+                                         optional: true)
+          ]
+
+          values = {
+              foo: "",
+              bar: nil,
+              oof: false
+          }
+
+          expect do
+            FastlaneCore::Configuration.create(conflicting_options, values)
+          end.not_to raise_error
         end
 
         it "calls custom conflict handler when conflict happens between two options" do
--- a/fastlane_core/spec/configuration_spec.rb
+++ b/fastlane_core/spec/configuration_spec.rb
@@ -38,14 +38,16 @@ describe FastlaneCore do
       describe "config conflicts" do
         it "raises an error if a key was used twice" do
           expect do
-            FastlaneCore::Configuration.create([FastlaneCore::ConfigItem.new(
-              key: :cert_name,
-              env_name: "asdf"
-            ),
-            FastlaneCore::ConfigItem.new(
-              key: :cert_name,
-              env_name: "asdf"
-            )], {})
+            FastlaneCore::Configuration.create([
+                                                 FastlaneCore::ConfigItem.new(
+                                                   key: :cert_name,
+                                                   env_name: "asdf"
+                                                 ),
+                                                 FastlaneCore::ConfigItem.new(
+                                                   key: :cert_name,
+                                                   env_name: "asdf"
+                                                 )
+                                               ], {})
           end.to raise_error("Multiple entries for configuration key 'cert_name' found!")
         end
 


minuscorp:
I don't see this as a bad solution at all, keep it simple as it is said! Maybe the other options are more tedious to implement, test and affect more the codebase.

minuscorp (approved):
LGTM! 🚀

minuscorp:
Friendly ping for @joshdholtz  to this stale PR 🙏🙇🏻‍♂️

joshdholtz (changes requested):
@rogerluan @minuscorp 

I am slightly terrified of this change, TBH 😬 I think this is making an assumption that any actions/tools only _do_ something with a `true` value. I’m pretty sure that _most_ actions do only do things with a `true` value but 🤷‍♂️ If that is the case, the conflicting options would be letting through some `false`s that shouldn’t go through.

I think another path we could take to solve this issue is to make the config items `optional: true` instead of `default_value: false`. TBH... I don’t like these options anyway 😛 It should really be one option that is an enum

Do either of you have any thoughts on my thoughts? 😇

minuscorp:
Yeah, I totally agree that this within isn't optimal, as we discussed but, we didn't wanted to change the current `ConfigItem` behavior. So when a Swift value detects a conflicting options you think it should take a value from an enum that reflects the status of being optional and with a default value at the same time. But maybe the solution could be changing the body of the action using Swift `preconditionFailure` when an option has any conflicting value that it is already being set (i.e. not `nil`) and skipping the default value (`false` in this example) or making an enum like:
```swift
enum OptionalConfigItem<Value> {
    case `default`(Value)
    case userDefined(Value)
}
```

Where we skip validation when the item is in a `default` state and is ignored when sending it as a `RubyArgument` and only being sent when is in `.userDefined` state, when it is possible to raise errors because they are being malformed into the `RubyArgument` and `RubyCommand`.

What do you think?

rogerluan:
A simpler and time-efficient solution is the `3` in my `possible solutions` 😄 From the PR description:
>Simply remove the validation if we agree it's nonsense and is just hurting us 🙈

Is the validation block really providing that much value? In the end the validation block just acts as a user friendlier way to tell users when they're doing something wrong, but if it's causing so many problems, as a stop gap it would virtually cause zero loss if we removed it immediately while we at least work on another solution. However, in this case particularly, I don't even think it's worth it to implement another solution. For future actions/config items we could just make the booleans optional from day zero to avoid facing this problem later on, or be more careful with validation blocks used on booleans.

The solution currently implemented is the most generic one that would fix the problems we have today as well as prevent them from revisiting us in the future (ofc, assuming we take the risk of the slight behavior change). 
The enum solution is probably the most elegant solution but also probably a bit over engineered. Not to mention it's not backwards compatible if I understood the proposal correctly 🙇 so we'd have to deprecate the existing broken config items and create new ones that use the new syntax 🙈 

In favor of valuing the limited time our contributors have that is put towards this project, I think we could go with a simpler and time efficient solution 😊

minuscorp:
The while enum proposal had to be thought about a little more, probably making it `ExpressibleBySomethingLiteral` in different `Value` cases so it'd be backwards compatible (when is set, is set to `userDefined`) of course I agree that is not the most straightforward solution neither the simpler, I was thinking of getting rid of this kind of issues at once. Anyway this is what you all decide is the best solution, I just brainstorm a bit 😅

rogerluan:
@minuscorp ah yes, that'd improve usability and backwards compatibility, it's a good point! 

Let's see what @joshdholtz thinks of my last message :)

Just a heads up that this issue manifested in the `xcov` gem as well. Issue linked above 😅

minuscorp:
Yeah, I saw... Issues spread 🤯

joshdholtz:
@rogerluan @minuscorp 

I’m so torn on this 😱 I went into the code and now my thought is do we even give a default value in Swift 🤦‍♂️ 

The default value gets returned dynamically when the key is being fetched from the config (which is why this isn’t a problem in Ruby). It’s happening during runtime. This is a problem when using Swift because the `default_value` (which was meant as a runtime thing) is now being used as a compile time thing. This compile time default value and conflicting options are what is causing this error. So my thought is now...

Do we just get rid of any parameter defaulting in Swift and instead make those parameters optional so... if they are passed as `nil`, the `default_value` will still get picked up dynamically at runtime when being fetched in the Ruby part 😬 

Let me know if I’m way off here in my thought and/or explaining this 😅

rogerluan:
@joshdholtz do you mean remove default values for every function and argument in Swift files? 😱 That would drastically change the UX of the Swift APIs making it significantly worse imo. We'd go from:

```swift
deliver()
```

To:

```swift
deliver(apiKeyPath: nil, apiKey: nil, username: "foo", appIdentifier: nil, appVersion: nil, ipa: nil, pkg: nil, buildNumber: nil, platform: "ios", editLive: true, useLiveVersion: false, metadataPath: nil, screenshotsPath: nil, skipBinaryUpload: true, skipScreenshots: false, skipMetadata: true, skipAppVersionUpdate: false, force: false, overwriteScreenshots: true, submitForReview: true, rejectIfPossible: true, automaticRelease: nil, autoReleaseDate: 1, phasedRelease: false, resetRatings: false, priceTier: nil, appRatingConfigPath: nil, submissionInformation: nil, teamId: nil, teamName: nil, devPortalTeamId: nil, devPortalTeamName: nil, itcProvider: nil, runPrecheckBeforeSubmit: false, precheckDefaultRuleLevel: "warning", individualMetadataItems: nil, appIcon: nil, appleWatchAppIcon: nil, copyright: nil, primaryCategory: nil, secondaryCategory: nil, primaryFirstSubCategory: nil, primarySecondSubCategory: nil, secondaryFirstSubCategory: nil, secondarySecondSubCategory: nil, tradeRepresentativeContactInformation: nil, appReviewInformation: nil, appReviewAttachmentFile: nil, description: nil, name: nil, subtitle: nil, keywords: nil, promotionalText: nil, releaseNotes: nil, privacyUrl: nil, appleTvPrivacyPolicy: nil, supportUrl: nil, marketingUrl: nil, languages: nil, ignoreLanguageDirectoryValidation: false, precheckIncludeInAppPurchases: true, app: "")
```

😅😅😅

But even if you meant removing the default value of only the options that have conflicting blocks declared, or the for optional booleans, that'd still be kinda weird and would change the signature (thus be a breaking change).

----

Following your train of thought that the conceptual difference is that ruby's is analyzed during runtime and Swift's during compile time, another perhaps more elegant solution would be for those values to become closures in Swift (or `@autoclosure`? Not sure how Ruby would interprete those haha), so we can evaluate them during runtime only. But if you ask me, that'd still be over engineering 🤔 

----

Do you have thoughts on the other aforementioned solutions? Namely `2` and `3` of the  PR description (assuming the `4`, which's the one implemented in this PR currently was already discarded 😬 )

joshdholtz:
@rogerluan Would it change the signature thought? 🤔 Since the values have a default value right now they can be left out. If we change them to optional with a default value of `nil` they can still be left out? Swift wouldn’t know anything about default values just like how Ruby doesn’t know anything of defaults values right away. It would all come down to being the runtime Ruby logic.

The only reason the default values are nice in Swift is for a documentation purpose right now when autocomplete. The default values in Swift don’t actually serve a purpose (besides allowing the user to not have to specify them but making them optional and defaulting them to `nil` would yield that same thing).

<hr/>

I think option 2 or option 3 would be fine but it wouldn’t really solve the problem at hand for all actions and would maybe lead to future problems if not caught right away 😔 

This probably is manifested on the Swift side of things (this wouldn’t be an issue if the Swift part didn’t exist) so I think this problem needs to be solved there 😬

minuscorp:
The enum approach will be always there, just if it is needed to be used and, even if it is a bit overingeneered I could do it without much issue I think. Also it is not a huge urgent thing to solve (for now), so let me know which one you all prefer😄

rogerluan:
@joshdholtz re: changing the signature, I think we're talking about diff things haha I'm not sure if you mean changing the default value of arguments for all options in Swift, or just optional booleans, or if it's something else 😅 

>The default values in Swift don’t actually serve a purpose

Actually the most valuable ones to me are those that default to e.g. `deliverfile.someSetting` because then I can define those values in my `Deliverfile.swift`. If it wasn't for that, I'd agree with you that defaulting all of them to nil would be the same as what we have today, in practical terms.

>This probably is manifested on the Swift side of things (this wouldn’t be an issue if the Swift part didn’t exist) so I think this problem needs to be solved there 😬

Ah I see! Looking from that perspective, I agree.

@minuscorp The enum option doesn't sound too bad after all if we can pull that. If we consider Josh's last comment, I see only these 2 solutions: the enum proposal, and perhaps the "default value change" that Josh is proposing, but I'm still uncertain about it yet 😅

minuscorp:
I'm eager to implement the enum one if you decide is worth doing it or, if you come up with another approach I can either implement it or if you feel confident going for it it's also totally fine. Let's wait for the master 👨‍💻

rogerluan:
I'm like the enum solution overall 🤗

minuscorp:
I'm on it, I'll try to have it ASAP❤️

minuscorp:
Created https://github.com/fastlane/fastlane/pull/18544 for the process!

joshdholtz:
Hello 👋  We can close this now because of #18544, right? 😇

rogerluan:
I _think_ so, although the original issue hasn't been fixed by #18544 😬  (ref: https://github.com/fastlane/fastlane/issues/17343#issuecomment-850659756)

But the fix for that should probably come in the format of the new enum, and not in this hacky way I implemented in this PR :) so I think we can close this!

PR #18001, opened 2021-01-19
[actions] download_dsyms - add auth `Connect for Api Token` automate

tastafur:
### Checklist

### Motivation and Context
As of February 27, the Two Factor will be imposed, so my continuous integration pipeline will stop working. So that it doesn't stop working I have changed the authentication to use this app_store_connect_api_key and I have verified that DownloadDsymsAction only uses the login authentication type, so I have added the other authentication type, the authentication by apiToken

### Description
If the APP_STORE_CONNECT_API_KEY or params key comes in the context lane, I use the authentication by api token instead of using the login

### Testing Steps
use the action app_store_connect_api_key in front of download_dsyms
`app_store_connect_api_key
 download_dsyms
`

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -14,9 +14,14 @@ module Fastlane
 
         # Team selection passed though FASTLANE_ITC_TEAM_ID and FASTLANE_ITC_TEAM_NAME environment variables
         # Prompts select team if multiple teams and none specified
-        UI.message("Login to App Store Connect (#{params[:username]})")
-        Spaceship::ConnectAPI.login(params[:username], use_portal: false, use_tunes: true)
-        UI.message("Login successful")
+        if (token = api_token(params))
+          UI.message("Using App Store Connect API token...")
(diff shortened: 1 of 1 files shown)


tastafur:
@googlebot I signed it!

tastafur:
Hello @joshdholtz @ainame, Would the PULL REQUEST be fine? I see that several tests are failing and from what I have seen in the code I think it should not fail, could you take a look at the pull request? Thanks @joshdholtz and @ainame

joshdholtz:
@tastafur I appreciate the time spent on this PR but this isn't going to work yet sadly 😔 The App Store Connect API does not have an endpoint available to download dSYMs with the API Token yet. I'm hoping it comes out sooner rather than later. I know this is going to be rough for everyone until that endpoint is available but its out of our control right now 💔 

Thanks again for the time spent on this!

tastafur:
@joshdholtz will not know other methods to download the dSYM files in a continuous integration system that is not to use login, because as of February 27, if you do not have to use FASTLANE_SESSION?

joshdholtz:
@tastafur Correct. FASTLANE_SESSION will be to be used ☹️

PR #18015, opened 2021-01-20
[fastlane_core] Support `SUPPORTS_UIKITFORMAC` build setting

DavidBrunow:
This build setting was originally used to determine whether an app supports Mac Catalyst and is still a valid flag.

### Checklist

### Motivation and Context

While trying to automate the builds of our Mac Catalyst app I ran into an issue where Fastlane failed to determine that our app supported Mac Catalyst. After some troubleshooting I figured out that it was due to the fact that we were using the original flag to indicate Mac Catalyst support (`SUPPORTS_UIKITFORMAC`) instead of the current flag that is used for that purpose (`SUPPORTS_MACCATALYST`). I think it would be helpful to support the legacy flag since it is still supported by Xcode.

### Description

The change is very simple -- instead of looking at only the `SUPPORTS_MACCATALYST` flag to determine whether an app supports Mac Catalyst the new code now looks at that flag and the `SUPPORTS_UIKITFORMAC` flag and if either are true then it treats the app as supporting Mac Catalyst.

### Testing Steps

Tested manually with our codebase.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -275,7 +275,7 @@ module FastlaneCore
     end
 
     def supports_mac_catalyst?
-      build_settings(key: "SUPPORTS_MACCATALYST") == "YES"
+      build_settings(key: "SUPPORTS_MACCATALYST") == "YES" || build_settings(key: "SUPPORTS_UIKITFORMAC") == "YES"
     end
 
     def command_line_tool?


joshdholtz:
@DavidBrunow Hello! Would you be able to rebase this from the latest master 😊

joshdholtz:
@DavidBrunow Would you be able to rebase this from master as well? 😊

DavidBrunow:
Not sure how I messed this up so badly -- I'll look into it more tomorrow.

joshdholtz (approved):
Looks good to me! Thank you for fixing things 😊

PR #18016, opened 2021-01-20
[gym] Handle situations where the `EXECUTABLE_NAME` is not accurate

DavidBrunow:
### Checklist

### Motivation and Context

When copying the Mac app the `EXECUTABLE_NAME` variable does not always match the `EXECUTABLE_NAME` that will be created. For example, when building an app in a workspace the first project's `EXECUTABLE_NAME` is used instead of the app project's `EXECUTABLE_NAME`.

### Description

This change simply looks for any files with the `.app` file extension and chooses the last one found if there are no files found when looking for the specific executable by name.

### Testing Steps

Tested manually with our code base.

### DIFF
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -274,11 +274,11 @@ module Gym
       exe_name = Gym.project.build_settings(key: "EXECUTABLE_NAME")
       app_path = File.join(BuildCommandGenerator.archive_path, "Products/Applications/#{exe_name}.app")
 
-      if File.exist?(app_path) == false
+      unless File.exist?(app_path)
         # Apparently the `EXECUTABLE_NAME` is not correct. This can happen when building a workspace which has a project
         # earlier in the build order that has a different `EXECUTABLE_NAME` than the app. Try to find the last `.app` as
         # a fallback for this situation.
-        app_path = Dir[File.join(BuildCommandGenerator.archive_path, "Products/Applications/*.app")].last
+        app_path = Dir[File.join(BuildCommandGenerator.archive_path, "Products", "Applications", "*.app")].last
       end
 
       UI.crash!("Couldn't find application in '#{BuildCommandGenerator.archive_path}'") unless File.exist?(app_path)
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -271,8 +271,7 @@ module Gym
 
     # Copies the .app from the archive into the output directory
     def copy_mac_app
-      exe_name = Gym.project.build_settings(key: "EXECUTABLE_NAME")
-      app_path = File.join(BuildCommandGenerator.archive_path, "Products/Applications/#{exe_name}.app")
+      app_path = Dir[File.join(BuildCommandGenerator.archive_path, "Products/Applications/*.app")].last
 
       UI.crash!("Couldn't find application in '#{BuildCommandGenerator.archive_path}'") unless File.exist?(app_path)
 
--- a/gym/lib/gym/runner.rb
+++ b/gym/lib/gym/runner.rb
@@ -271,7 +271,12 @@ module Gym
 
     # Copies the .app from the archive into the output directory
     def copy_mac_app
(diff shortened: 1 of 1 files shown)


joshdholtz:
@DavidBrunow And rebase this from master as too? 😇  I tried but I'm not allowed to add a commit to your fork since its an organization account 🤷‍♂️

DavidBrunow:
Not sure how I messed this up so badly -- I'll look into it more tomorrow.

gym/lib/gym/runner.rb
```diff
@@ -271,8 +271,7 @@ def copy_files_from_path(path)
 
     # Copies the .app from the archive into the output directory
     def copy_mac_app
-      exe_name = Gym.project.build_settings(key: "EXECUTABLE_NAME")
-      app_path = File.join(BuildCommandGenerator.archive_path, "Products/Applications/#{exe_name}.app")
+      app_path = Dir[File.join(BuildCommandGenerator.archive_path, "Products/Applications/*.app")].last
```
joshdholtz, line 274: Do you know if there is any chance that there could be multiple `.app`s that are created or there could be a previous version that existed with a different name that this could be picking up? 🤔 

Maybe this needs to pick up the newest `.app` file from the directory? 🤷‍♂️  Thoughts?

joshdholtz (changes requested):
Sorry about slow review! Starting to catch back up from being on family leave 😊 This looks great but just one thought on making sure we pick up the correct `.app`! Lemme know your thoughts 🙃

gym/lib/gym/runner.rb
DavidBrunow: I'll do some research on how Xcode handles multiple apps in the same workspace/project to see if it is possible to have more than one.

Another option is that Fastlane could fallback to the last `.app` found only if an app cannot be found with the `EXECUTABLE_NAME`.

joshdholtz: Ohhhh, I would feel a lot better if we used the last `.app` as a fallback 😇 Do you mind changing it up to do that?

DavidBrunow: @joshdholtz This is ready for another look when you get a chance.

DavidBrunow: @joshdholtz Don't want to be a bother but bumping this in case you missed the last message.

gym/lib/gym/runner.rb
```diff
@@ -274,6 +274,13 @@ def copy_mac_app
       exe_name = Gym.project.build_settings(key: "EXECUTABLE_NAME")
       app_path = File.join(BuildCommandGenerator.archive_path, "Products/Applications/#{exe_name}.app")
 
+      if File.exist?(app_path) == false
```
minuscorp, line 277: `unless File.exist?(app_path)`

minuscorp (commented):
One minor comment and looks fine.

gym/lib/gym/runner.rb
```diff
@@ -274,6 +274,13 @@ def copy_mac_app
...
+        # Apparently the `EXECUTABLE_NAME` is not correct. This can happen when building a workspace which has a project
+        # earlier in the build order that has a different `EXECUTABLE_NAME` than the app. Try to find the last `.app` as
+        # a fallback for this situation.
+        app_path = Dir[File.join(BuildCommandGenerator.archive_path, "Products/Applications/*.app")].last
```
minuscorp, line 281: I'd separate each path component to let `join` do its work.

gym/lib/gym/runner.rb
DavidBrunow: Thanks, that is much better!

minuscorp (approved):
LGTM!

DavidBrunow:
@joshdholtz another friendly ping on this pull request in case you have a chance to look at it again.

joshdholtz:
Thank you for friendly ping! Will look at today 😁

joshdholtz (approved):
Looks good! Thanks for making this change ❤️

PR #18031, opened 2021-01-22
[docs] update git_commit description

Lewis-Bright:
### Checklist

### Motivation and Context
There is nothing in the documentation on how to commit all files, as opposed to just one. This will make it easier to see this common use case without trial and error.

### Description
Simply updated the description to explain how to add all files.

### Testing Steps
No functionality change so nothing to test there. It's best just to check the wording and grammar is appropriate.

### DIFF
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -37,7 +37,7 @@ module Fastlane
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :path,
-                                       description: "The file(s) or directory(ies) you want to commit. You can pass an array of multiple file-paths or fileglobs \"*.txt\" to commit all matching files. The files already staged but not specified and untracked files won't be committed.",
+                                       description: "The file(s) or directory(ies) you want to commit. You can pass an array of multiple file-paths or fileglobs \"*.txt\" to commit all matching files. The files already staged but not specified and untracked files won't be committed",
                                        is_string: false),
           FastlaneCore::ConfigItem.new(key: :message,
                                        description: "The commit message that should be used"),
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -37,7 +37,7 @@ module Fastlane
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :path,
-                                       description: "The file(s) you want to commit, use . to match all paths",
+                                       description: "The file(s) or directory(ies) you want to commit. You can pass an array of multiple file-paths or fileglobs \"*.txt\" to commit all matching files. The files already staged but not specified and untracked files won't be committed.",
                                        is_string: false),
           FastlaneCore::ConfigItem.new(key: :message,
                                        description: "The commit message that should be used"),
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -37,7 +37,7 @@ module Fastlane
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :path,
-                                       description: "The file you want to commit",
+                                       description: "The file(s) you want to commit, use . to match all paths",
                                        is_string: false),
           FastlaneCore::ConfigItem.new(key: :message,
                                        description: "The commit message that should be used"),


fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -37,7 +37,7 @@ def self.description
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :path,
-                                       description: "The file you want to commit",
+                                       description: "The file(s) you want to commit, use . to match all paths",
```
ainame, line 40: Hi, thank you for taking the time. I'm happy to join you to improve this document. This is a good point that parameter name is singular "path" which doesn't sound like you can commit multiple files. So I like adding "(s)"🚀 

However, I believe that generally "." is a bad practice that you may involve unwanted changes and occurs incidents. (Though I use it. I know this is very handy🙈) The thing is that if you know how this could be bad, you would be fine but given fastlane are used a lot of users including beginners, I think it's not good that spreading a bad practice in its document. For example, this case would be safe because the top `ensure_git_status_clean` prevents you from committing files made between the first and last line. You must be aware of what you are going to commit.

```ruby
ensure_git_status_clean
sh('touch tmp/.gitignore')
git_commit(path: ".")
```

I couldn't come up with a good example but if you were to commit files that are generated by their party tool, it can be dangerous in future as you may not know what file updates will be coming after some version bumps of if.

```ruby
sh('some_third_pary_cli_tool_to_update_some_files_in_project')
git_commit(path: ".")
```

-------------------------

I come to think that instead, we should mention you can use `Array` to specify multiple file paths or fileglobs, which are missing here. Additionally, it's worth mentioning here about `git commit <file path>`'s behaviour in untracked and staged files. 

```suggestion
                                       description: "The file(s) or directory(ies) you want to commit. You can pass an array of multiple file-paths or fileglobls \"*.txt\" to commit all matching files. The files already staged but not specified in this or untracked files won't be committed.",
```

This is what I can think of to suggest better practice to beginners instead.😅 What do you think?

ainame (commented):
Welcome to the fastlane community and congrats your first-time contribution! 🚀  I left my thought on your change. I hope it makes sense to you🤞 

And also, if you want this to get merged, you need to follow the steps that google-cla bot says above eventually. Please follow that 🙇

Lewis-Bright:
> @googlebot I signed it!

fastlane/lib/fastlane/actions/git_commit.rb
Lewis-Bright: Hi, thanks for your reply and feedback, it's much appreciated. I understand your point on not wanting to commit added files, if you haven't specified them. My main motivator for this PR was actually the following error message, which occurs if you specify an empty string for the path parameter:
```
C:/tools/ruby27/lib/ruby/gems/2.7.0/gems/fastlane-2.172.0/fastlane_core/lib/fastlane_core/ui/interface.rb:153:in `shell_error!': \e[31m[!] Exit status of command 'git commit -m "Version bump and release" ""' was 128 instead of 0. (FastlaneCore::Interface::FastlaneShellError)
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
\e[0m
```
This sort of ends up recommending that you use "." anyway (although I understand it's just passing up an error from the shell). 

I'll commit your suggestion. I may open a second PR to catch the empty path parameter earlier and raise a custom error (perhaps suggesting that you can use "." but recommending you not to, using your reason above?). What do you think about that idea?

Thanks again! :)

fastlane/lib/fastlane/actions/git_commit.rb
```diff
@@ -37,7 +37,7 @@ def self.description
       def self.available_options
         [
           FastlaneCore::ConfigItem.new(key: :path,
-                                       description: "The file you want to commit",
+                                       description: "The file(s) or directory(ies) you want to commit. You can pass an array of multiple file-paths or fileglobs \"*.txt\" to commit all matching files. The files already staged but not specified and untracked files won't be committed.",
```
ainame, line 40:
```suggestion
                                       description: "The file(s) or directory(ies) you want to commit. You can pass an array of multiple file-paths or fileglobs \"*.txt\" to commit all matching files. The files already staged but not specified and untracked files won't be committed",
```

Oh sorry apparently a description can't end with "." 😅

fastlane/lib/fastlane/actions/git_commit.rb
ainame: I understood how you came across "."😓 and the plan sounds good to me! 

> perhaps suggesting that you can use "." but recommending you not to, using your reason above?

Maybe we should still suggest passing file paths *first* and then the above can follow it. An empty string for `path` can be a program error if it's given from some logic.

ainame (approved):
Thank you for pointing this out 🚀   `git_commit` will be more understandable for users 💪 
And congrats your first-time contribution to fastlane 🎉 

@joshdholtz will then double-check and merge this when he can 🙇

joshdholtz (approved):
Love this change! Thank you for updating this ❤️

PR #18042, opened 2021-01-23
[action] fix download dsyms with zero prefixed versions

diskshima:
### Checklist

### Motivation and Context

The the `preReleaseVersion.version` filter in the [App Store Connect API to list builds](https://developer.apple.com/documentation/appstoreconnectapi/list_builds) seems to be comparing major/minor/subminor versions as integers not as verbatim strings and fails to retrieve builds when the version includes leading zeros (e.g. `21.01.1`).

Also the version number in the build info returned by the App Store Connect API includes the version as strings (with leading zeros) but as the existing `download_dsyms` code is stripping out leading zeros prior to comparing the version specified and the versions returned from the API, it **fails to download the corresponding dSYMs** thinking that none of the versions match.

### Description

This change  deals with the two issues above by 1) stripping out leading zeros before setting it to the `preReleaseVersion.version` filter, 2) removes the leading zero stripping when comparing the versions.

I have tested by uploading and releasing an iPhone app with a leading zero in its minor version (e.g. `21.01.1`), running both
`download_dsyms(version: "live")` and `download_dsyms(version: "latest")` and confirmed that both have successfully downloaded the dSYM.

### Testing Steps

Please see above Description for testing steps.

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -163,7 +163,8 @@ module Fastlane
 
       def self.get_latest_build!(app_id: nil, version: nil, platform: nil)
         filter = { app: app_id }
-        filter["preReleaseVersion.version"] = version
+        # The version filter seems to ignore zero prefixes in versions (e.g. 21.01)
+        filter["preReleaseVersion.version"] = version.split(".").map(&:to_i).join(".")
         filter["preReleaseVersion.platform"] = platform
         latest_build = Spaceship::ConnectAPI.get_builds(filter: filter, sort: "-uploadedDate", includes: "preReleaseVersion").first
 
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -59,11 +59,6 @@ module Fastlane
           build_number = live_version.build.version
         end
 
-        # Remove leading zeros from version string (eg. 1.02 -> 1.2)
-        if version
-          version = version.split(".").map(&:to_i).join(".")
-        end
-
         # Make sure output_directory has a slash on the end
         if output_directory && !output_directory.end_with?('/')
           output_directory += '/'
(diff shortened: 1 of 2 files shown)


diskshima:
@googlebot I signed it!

fastlane/lib/fastlane/actions/download_dsyms.rb
```diff
@@ -59,11 +59,6 @@ def self.run(params)
           build_number = live_version.build.version
         end
 
-        # Remove leading zeros from version string (eg. 1.02 -> 1.2)
-        if version
-          version = version.split(".").map(&:to_i).join(".")
-        end
```
ainame, line 65: @diskshima Any insights/thoughts on why this was done like this? If not, don't worry. I will dig into this later to review 😅

ainame, line 65: Note: This was added here https://github.com/fastlane/fastlane/pull/15235

> The download_dsyms action fails when an app's version has a leading zero in one of its parts, such as a version "4.07". AppStore reports this version as "4.7", so the action fails to find its dSYM files, since "4.07" != "4.7".

According to the previous PR, App Store Connect (not the public API) seemed to return the version without leading zeros previously. So it makes sense to change this to follow the latest spec of APIs.

ainame (approved):
Welcome to fastlane community👋 and congrats the first-time contribution! 🎉  Looks good to me!

@joshdholtz will then double-check and merge this when he can 🙇

fastlane/lib/fastlane/actions/download_dsyms.rb
diskshima: Thank you for the review @ainame!

Just to add, I couldn't find any official Apple API docs on how the leading zeros are being treated so it's merely based on the observed behaviour of the Apple API. It may break (again) in the future but I guess it shouldn't affect too many people as not many people use leading zeros in their versions 😉

joshdholtz:
This seems good to me! Thank you so much for the great description and the contribution ❤ Really appreciate the work!

And thanks for the review, @ainame 🙌

joshdholtz (approved):

PR #18074, opened 2021-01-29
[action] appetize add timeout support

raulriera:
### Checklist

### Motivation and Context
Adding support for the `timeout` option in Appetize, as seen in their documentation https://docs.appetize.io/api/update-existing-app it would be ideal to support more than the default 2mins.

### Description
Expanded the `appetize` and `build_and_upload_to_appetize` actions to include this parameter. Also included the appropriate validations and test to make sure we always submit the values specified by Apptize themselves.

### Testing Steps
Running `bundle exec rspec ./fastlane/spec/actions_specs/appetize_spec.rb ` will result in green tests.

### DIFF
--- a/fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
+++ b/fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
@@ -77,7 +77,14 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :note,
                                        description: "Notes you wish to add to the uploaded app",
                                        is_string: true,
-                                       optional: true)
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :timeout,
+                                       description: "The number of seconds to wait until automatically ending the session due to user inactivity. Must be 30, 60, 90, 120, 180, 300, 600, 1800, 3600 or 7200. Default is 120",
+                                       type: Integer,
+                                       optional: true,
+                                       verify_block: proc do |value|
+                                         UI.user_error!("The value provided doesn't match any of the supported options.") unless [30, 60, 90, 120, 180, 300, 600, 1800, 3600, 7200].include?(value)
+                                       end)
         ]
       end
 
--- a/fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
+++ b/fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
@@ -21,7 +21,8 @@ module Fastlane
         other_action.appetize(path: zipped_bundle,
                               api_token: params[:api_token],
                               public_key: params[:public_key],
-                              note: params[:note])
+                              note: params[:note],
+                              timeout: params[:timeout])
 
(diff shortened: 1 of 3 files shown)


raulriera:
@googlebot I signed it!

fastlane/lib/fastlane/actions/build_and_upload_to_appetize.rb
```diff
@@ -21,7 +21,8 @@ def self.run(params)
         other_action.appetize(path: zipped_bundle,
                               api_token: params[:api_token],
                               public_key: params[:public_key],
-                              note: params[:note])
+                              note: params[:note],
+                              timeout: params[:timeout])
```
raulriera, line 25: Wasn't able to find a test for using this file

fastlane/swift/Fastlane.swift
```diff
@@ -230,15 +230,17 @@ public func appetize(apiHost: String = "api.appetize.io",
                      platform: String = "ios",
                      path: String? = nil,
                      publicKey: String? = nil,
-                     note: String? = nil)
+                     note: String? = nil,
+                     timeout: Integer? = nil)
```
Caiopia, line 234: Can't comment on the exact line, but we could update the docs for this new parameter too. Same with the `buildAndUploadToAppetize` function below.

fastlane/swift/Fastlane.swift
```diff
@@ -882,13 +884,15 @@ public func buildAndUploadToAppetize(xcodebuild: [String: Any] = [:],
                                      scheme: String? = nil,
                                      apiToken: String,
                                      publicKey: String? = nil,
-                                     note: String? = nil)
+                                     note: String? = nil,
+                                     timeout: Integer?)
```
Caiopia, line 888: Any reason this shouldn't have the same default value as above?

Caiopia (commented):
Code looks good to me, just some minor suggestions. Haven't tested this yet though.

minuscorp (changes requested):
Initially it seems good to me. A few notes, don't push Swift changes add they are autogenerated by fastlane, and add to the `ConfigItem` the `:type` key to pass it correctly to the Swift code (I.e. `Integer`)

raulriera:
Hey @minuscorp , what are the request changes? GitHub is not showing any information

raulriera:
Thanks for the review @minuscorp, code is updated. I didn't found anything in the docs about generating the swift code. Is there a step I need to do for this?

minuscorp:
Just try to update it locally and see if the changes that you expect are done, there's a lane called `generate_swift_api` or similar that updates and build the whole interface and SPM package.

raulriera:
Thanks, that did it. Should I include the output of all that generation here? (more files were changed) or is that only a release process?

minuscorp:
That's part of the release process, but that way you can make sure that your changes are good. Useful for contributions 🎉

minuscorp (approved):
LGTM 🎉

joshdholtz (approved):
This is 🔥 Thanks so much for adding this! Really appreciate the contribution ❤️

PR #18080, opened 2021-01-31
[spaceship] added ASC endpoints for bundle id & bundle id capability

dinsen:
### Checklist

### Motivation and Context
I use fastlane every day in my day job and I want to contribute back to the tool that saves me a lot of hours every day.
So I found some endpoints that was not implemented for the ASC API and added them.

### Description
While I wanted to add the bundleIdCapability endpoints, I found out that the creation of an bundle identifier was also missing. So I added it since I was already in the code 😄 

I tested the method in the Rakefile to verify the new methods was working.

### Testing Steps
I used the following in the Rakefile and run it with `bundle exec rake debug --trace`
```ruby
task(:debug) do
  require 'spaceship'

  token = Spaceship::ConnectAPI::Token.from_json_file(File.expand_path("fastlane.json"))
  Spaceship::ConnectAPI.token = token

  puts("Create bundle id")
  bundle_id = Spaceship::ConnectAPI::BundleId.create(name: "fastlane demo", identifier: "dk.fastlane.demo")
  puts(bundle_id.to_yaml)

  # List
  puts("List capabilities")
  bundle_id_capabilities = bundle_id.get_capabilities
  puts(bundle_id_capabilities.to_yaml)

  # Create
  puts("Create capability maps")
  bundle_id.create_capability(Spaceship::ConnectAPI::BundleIdCapability::Type::MAPS)

  # Sleep so I can refresh the browser to verify MAPS was added
  sleep(8)

  # List
  puts("List capabilities")
  bundle_id = Spaceship::ConnectAPI::BundleId.get(bundle_id_id: bundle_id.id)
  bundle_id_capabilities = bundle_id.get_capabilities
  puts(bundle_id_capabilities.to_yaml)

  # Delete
  puts("Delete capability maps")
  bundle_id_capabilities.each do |capability|
    if capability.capability_type == Spaceship::ConnectAPI::BundleIdCapability::Type::MAPS
      capability.delete!
    end
  end

  # List
  puts("List capabilities")
  bundle_id = Spaceship::ConnectAPI::BundleId.get(bundle_id_id: bundle_id.id)
  bundle_id_capabilities = bundle_id.get_capabilities
  puts(bundle_id_capabilities.to_yaml)
end
```

### DIFF
--- a/spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
+++ b/spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
@@ -41,6 +41,21 @@ module Spaceship
         MARZIPAN = "MARZIPAN" # Catalyst
       end
 
+      module Settings
+        ICLOUD_VERSION = "ICLOUD_VERSION"
+        DATA_PROTECTION_PERMISSION_LEVEL = "DATA_PROTECTION_PERMISSION_LEVEL"
+        APPLE_ID_AUTH_APP_CONSENT = "APPLE_ID_AUTH_APP_CONSENT"
+      end
+
+      module Options
+        XCODE_5 = "XCODE_5"
+        XCODE_6 = "XCODE_6"
+        COMPLETE_PROTECTION = "COMPLETE_PROTECTION"
+        PROTECTED_UNLESS_OPEN = "PROTECTED_UNLESS_OPEN"
+        PROTECTED_UNTIL_FIRST_USER_AUTH = "PROTECTED_UNTIL_FIRST_USER_AUTH"
+        PRIMARY_APP_CONSENT = "PRIMARY_APP_CONSENT"
+      end
+
       def self.type
         return "bundleIdCapabilities"
       end
@@ -59,14 +74,26 @@ module Spaceship
       # API
       #
 
-      def self.all(client: nil, filter: {}, includes: nil, limit: nil, sort: nil)
+      def self.all(client: nil, bundle_id_id: nil, filter: {}, includes: nil, limit: nil, sort: nil)
+        raise "bundle_id_id is required " if bundle_id_id.nil?
+
         client ||= Spaceship::ConnectAPI
-        return client.get_users(filter: filter, includes: includes)
+        resp = client.get_bundle_id_capabilities(bundle_id_id: bundle_id_id, filter: filter, includes: includes).all_pages
+        return resp.flat_map(&:to_models)
       end
 
-      def self.find(client: nil, email: nil, includes: nil)
+      def self.create(client: nil, bundle_id_id: nil, capability_type: nil, settings: [])
+        raise "bundle_id_id is required " if bundle_id_id.nil?
+        raise "capability_type is required " if capability_type.nil?
+
         client ||= Spaceship::ConnectAPI
-        return all(client: client, filter: { email: email }, includes: includes)
+        resp = client.post_bundle_id_capability(bundle_id_id: bundle_id_id, capability_type: capability_type, settings: settings)
+        return resp.to_models.first
+      end
+
+      def delete!(client: nil, filter: {}, includes: nil, limit: nil, sort: nil)
+        client ||= Spaceship::ConnectAPI
+        client.delete_bundle_id_capability(bundle_id_capability_id: id)
       end
     end
   end
--- a/spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
+++ b/spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
@@ -74,18 +74,13 @@ module Spaceship
       # API
       #
 
-      def self.all(client: nil, bundle_id_id: nil, filter: {}, includes: nil, limit: nil, sort: nil)
-        raise "bundle_id_id is required " if bundle_id_id.nil?
-
+      def self.all(client: nil, bundle_id_id:, limit: nil)
         client ||= Spaceship::ConnectAPI
-        resp = client.get_bundle_id_capabilities(bundle_id_id: bundle_id_id, filter: filter, includes: includes).all_pages
+        resp = client.get_bundle_id_capabilities(bundle_id_id: bundle_id_id, limit: limit).all_pages
         return resp.flat_map(&:to_models)
       end
 
-      def self.create(client: nil, bundle_id_id: nil, capability_type: nil, settings: [])
-        raise "bundle_id_id is required " if bundle_id_id.nil?
-        raise "capability_type is required " if capability_type.nil?
-
+      def self.create(client: nil, bundle_id_id:, capability_type:, settings: [])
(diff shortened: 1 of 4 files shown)


spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
```diff
@@ -59,14 +74,26 @@ def is_type?(type)
       # API
       #
 
-      def self.all(client: nil, filter: {}, includes: nil, limit: nil, sort: nil)
+      def self.all(client: nil, bundle_id_id: nil, filter: {}, includes: nil, limit: nil, sort: nil)
+        raise "bundle_id_id is required " if bundle_id_id.nil?
```
ainame, line 78: FYI: You can use `ArgumentError` for this kind of validation in Ruby in general.


```suggestion
        raise ArgumentError, 'bundle_id_id is required' if bundle_id_id.nil?
```

(You don't have to check if `bundle_id_id` is given manually since you can remove default value)

spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
```diff
@@ -59,14 +74,26 @@ def is_type?(type)
...
+
+        client ||= Spaceship::ConnectAPI
+        resp = client.get_bundle_id_capabilities(bundle_id_id: bundle_id_id, filter: filter, includes: includes).all_pages
+        return resp.flat_map(&:to_models)
+      end
+
+      def self.create(client: nil, bundle_id_id: nil, capability_type: nil, settings: [])
+        raise "bundle_id_id is required " if bundle_id_id.nil?
+        raise "capability_type is required " if capability_type.nil?
```
ainame, line 87:
```suggestion
      def self.create(client: nil, bundle_id_id:, capability_type:, settings: [])
```

We can now remove the default value of keyword argument since we support Ruby from v2.4😄

spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
ainame, line 78: I understood the previous implementation is like a copy-and-paste from `user.rb`😅  

https://developer.apple.com/documentation/appstoreconnectapi/list_all_capabilities_for_a_bundle_id
Looking at the document page, there are only a few options supported. So you can shorten the method signature like this?

```suggestion
      def self.all(client: nil, bundle_id_id:, limit: nil)
```

spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
```diff
@@ -41,6 +41,21 @@ module Type
...
+      end
+
+      module Options
+        XCODE_5 = "XCODE_5"
+        XCODE_6 = "XCODE_6"
+        COMPLETE_PROTECTION = "COMPLETE_PROTECTION"
+        PROTECTED_UNLESS_OPEN = "PROTECTED_UNLESS_OPEN"
+        PROTECTED_UNTIL_FIRST_USER_AUTH = "PROTECTED_UNTIL_FIRST_USER_AUTH"
+        PRIMARY_APP_CONSENT = "PRIMARY_APP_CONSENT"
+      end
```
ainame, line 57: I read the document but it looks like `CapabilitySetting` is a quite complicated structure.
https://developer.apple.com/documentation/appstoreconnectapi/capabilitysetting

How did you compose that? Do we need to provide exactly all the properties described in the page?
If so, maybe we can help users to compose it by providing a plain Ruby object for it?

```ruby
class CapacitySetting
  attr_reader :key, :options # and etc..
  def initialize(key:, options:, ....)
end
```

spaceship/lib/spaceship/connect_api/provisioning/provisioning.rb
```diff
@@ -27,6 +27,61 @@ def get_bundle_id(bundle_id_id: {}, includes: nil)
...
+                }
+              }
+            }
+          }
+
+          provisioning_request_client.post("bundleIdCapabilities", body)
+        end
+
+        def delete_bundle_id_capability(bundle_id_capability_id: nil)
+          raise "Capability id is nil" if bundle_id_capability_id.nil?
```
ainame, line 80:
```suggestion
        def delete_bundle_id_capability(bundle_id_capability_id:)
```

ainame (changes requested):
Thank you for the first time contribution to the codebase🚀 I've seen you helping the community on issues already😄 

I've pointed a question regarding how to compose a request and suggestions for method signature.

spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
Dinsen: Hi @ainame 
Thank you for taking the time to explain all the stuff in your other comments!

Of all the endpoints I choose to implement, I found out that I choose to implement the most complicated one 😅

I have never used one of the capabilities that require configuration before, so I have no clue what all the parameters should be or which one is required.
I just added the possibility to add the settings, so that users that knows how to set it up, actually can do it.

But I will be happy to add an object for the `CapabilitySetting` if anyone can provide some information about thies parameters 😄

ainame (commented):
> But I will be happy to add an object for the CapabilitySetting if anyone can provide some information about thies parameters 😄

This is great! Would you mind if you work out what is needed for it at least? If you do so, we can note it in the code. Implementing this in another PR is fine. Otherwise, this might end up in not really working and we get error tickets without no clues😅 Does that make sense?

spaceship/lib/spaceship/connect_api/models/bundle_id_capability.rb
joshdholtz: I love this idea of adding a `CapabilitySetting` building 😊 I don’t think it has to be added on this PR but I will be making an update to `produce` soon that needs this so I’ll add this builder to that PR 💪 

I think leaving it out for now is perfectly fine for this PR but I do love the idea 😍

joshdholtz:
@ainame @Dinsen I create an issue for the `produce` rewrite that will contain the new `CabailitySetting` builder - https://github.com/fastlane/fastlane/issues/18157 

I’ll be starting on this rewrite next week 😊

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This is so 🔥  Thank you for adding this! The example you added to the PR description made this so so so nice to test 🙌 

I really appreciate your contribution to _fastlane_ ❤️  Hope to see you back soon!

ainame (approved):
> @ainame @Dinsen I create an issue for the produce rewrite that will contain the new CabailitySetting builder - #18157

If @joshdholtz is going to be responsible for improving `CapacitySetting` thing, I'm now happy with this PR😄  

Thank you for your contribution! 🚀

Dinsen:
Yay! Thanks @ainame & @joshdholtz 🙌

PR #18088, opened 2021-02-02
[fastlane_core] improve config item validation conversion and use UI.error instead of puts to report validation exception

mokagio:
### Checklist

### Motivation and Context

At [Automattic](https://automattic.com), we use Fastlane to automate the release process for our mobile app 💪 ✨ We built [a suite of custom actions](github.com/wordpress-mobile/release-toolkit) to help us with that.

While running one of our actions, I passed a string to the interactive prompt for a parameter that was supposed to only ever be an integer. I was happy to see that Fastlane allows to specify the type of the parameters 😍 I went and changed the type from `String` to `Integer` but when I tried to verify my fix, Fastlane kept telling me that my `123` input wasn't an `Integer`.

To recap, if you build a custom action and configure a parameter like:

```ruby
FastlaneCore::ConfigItem.new(
  key: :int_param,
  env_name: "INTEGER_PARAM_DEMO_YOUR_OPTION",
  description: "A non optional Integer parameter",
  optional: false,
  type: Integer
)
```

When you run your action without passing the value at the call-site (i.e. `bundle exec fastlane run my_action` instead of `bundle exec fastlane run my_action int_param:123`) and Fastlane prompts you for the value, even if you pass a correct value (in the case of `type: Integer`, you type only digits), Fastlane will consider the input a `String` and ask you to pass it again.

```
➜ bf run integer_param_demo
[✔] 🚀
+------------------------------------+---------+--------------------+
|                           Used plugins                            |
+------------------------------------+---------+--------------------+
| Plugin                             | Version | Action             |
+------------------------------------+---------+--------------------+
| fastlane-plugin-integer_param_demo | 0.1.0   | integer_param_demo |
+------------------------------------+---------+--------------------+

[15:08:09]: --------------------------------
[15:08:09]: --- Step: integer_param_demo ---
[15:08:09]: --------------------------------
[15:08:09]: To not be asked about this value, you can specify it using 'int_param'
[15:08:09]: A non optional Integer parameter: 123
'int_param' value must be a Integer! Found String instead.
[15:08:15]: To not be asked about this value, you can specify it using 'int_param'
[15:08:15]: A non optional Integer parameter:
```

I reproduced the issue in [this isolated repository](https://github.com/mokagio/fastlane_plugin_test/). There are detailed instructions on how to install and reproduce in the README and a PR that points Fastlane to this fix to show its effectiveness.

### Description

To ensure that parameters provided via UI input are converted to their desired `type`, I added a call to `auto_convert_value` after reading the user input. More details in the inline comments.

### Testing Steps

Please refer to in the demo repo. I also added a unit test for this.

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -243,6 +243,12 @@ module FastlaneCore
       while value.nil?
         UI.important("To not be asked about this value, you can specify it using '#{option.key}'") if ENV["FASTLANE_ONBOARDING_IN_PROCESS"].to_s.length == 0
         value = option.sensitive ? UI.password("#{option.description}: ") : UI.input("#{option.description}: ")
+
+        # ConfigItem allows to specify a type for the item but UI.password and
+        # UI.input return String values. Try to convert the String input to
+        # the option's type before passing it along.
+        value = option.auto_convert_value(value)
+
         # Also store this value to use it from now on
         begin
           set(key, value)
--- a/fastlane_core/lib/fastlane_core/configuration/configuration.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/configuration.rb
@@ -247,7 +247,7 @@ module FastlaneCore
         begin
           set(key, value)
         rescue => ex
-          puts(ex)
+          UI.error(ex)
           value = nil
         end
       end
--- a/fastlane_core/spec/configuration_spec.rb
+++ b/fastlane_core/spec/configuration_spec.rb
@@ -718,14 +718,24 @@ describe FastlaneCore do
               expect(config[:item]).to eq(123)
             end
 
-            # It would be good to have a test for the scenario where the user
-            # inserts a value that doesn't match the expected type to verify
-            # we show an error. Unfortunately, because after showing the error
-            # we ask for the value again, we'd be stuck into an infinite loop
-            # in the tests.
-            #
-            # Note that using `.to receive(:input).once` like above is not an
-            # option because that would fail the test in this case.
+            it "shows an error after user inputs value that doesn't match type (the first time) and works the second time" do
+              config_item = FastlaneCore::ConfigItem.new(key: :item,
+                                                         short_option: "-i",
+                                                         env_name: "ITEM_ENV_VAR",
+                                                         description: "a description",
+                                                         is_string: false,
+                                                         type: Integer,
+                                                         # false is the default, but let's be explicit
+                                                         skip_type_validation: false)
+              config = FastlaneCore::Configuration.create([config_item], {})
+
+              config.set(:item, nil)
+              expect(FastlaneCore::UI).to receive(:input).once.and_return("123abc")
+              expect(FastlaneCore::UI).to receive(:input).once.and_return("123")
+              expect(FastlaneCore::UI).to(receive(:error)).once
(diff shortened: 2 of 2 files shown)


fastlane_core/lib/fastlane_core/configuration/configuration.rb
```diff
@@ -243,11 +243,17 @@ def fetch(key, ask: true)
...
+        # UI.input return String values. Try to convert the String input to
+        # the option's type before passing it along.
+        value = option.auto_convert_value(value)
+
         # Also store this value to use it from now on
         begin
           set(key, value)
         rescue => ex
-          puts(ex)
+          UI.error(ex)
```
mokagio, line 256: I didn't go for `user_error!` because it would make Fastlane exit, right? Whereas here we just want to show an error and give the user another chance to insert the correct value.

fastlane_core/spec/configuration_spec.rb
```diff
@@ -693,28 +693,60 @@
...
+
+            it "doesn't show an error when skipping type validation" do
+              # Taken from match/options.rb
+              config_item = FastlaneCore::ConfigItem.new(key: :app_identifiers,
+                                                         short_option: "-a",
+                                                         env_name: "MATCH_APP_IDENTIFIER",
+                                                         description: "The bundle identifier(s) of your app (comma-separated)",
+                                                         is_string: false,
+                                                         type: Array, # we actually allow String and Array here
+                                                         skip_type_validation: true,
```
mokagio, line 738: This is the reason that, despite having a test for the auto conversion, the issue was never caught: this test input skips type validation.

fastlane_core/spec/configuration_spec.rb
```diff
@@ -693,28 +693,60 @@
...
+            end
+
+            # It would be good to have a test for the scenario where the user
+            # inserts a value that doesn't match the expected type to verify
+            # we show an error. Unfortunately, because after showing the error
+            # we ask for the value again, we'd be stuck into an infinite loop
+            # in the tests.
+            #
+            # Note that using `.to receive(:input).once` like above is not an
+            # option because that would fail the test in this case.
```
mokagio, line 728: Does this make sense?

I thought about using something like:

```ruby
call_count = 0
expect(FastlaneCore::UI).to receive(:input) {
  if call_count == 0
    call_count++
    return "not an integer"
  else
    # bail out
  end
}
```

Unfortunately, I couldn't find a way to stop the "ask user for input - validation failed - ask user for input ..." loop that didn't make the test fail.

Vyazovoy:
Really nice UX improvement! 👏 I hope it will be merged soon. (@joshdholtz seems like a pretty lightweight change to merge)

mokagio:
@Vyazovoy thank you for your kind words 😄

fastlane_core/spec/configuration_spec.rb
joshdholtz: @mokagio Pushed a commit to add a test for this! Thank you for mentioned a test like this was needed 💪

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This is such a good change 🔥 Thank you so much for the great comments and added tests! Nothing puts a bigger smile on face than seeing those when opening up the “Files Changed” 😁 

Really appreciate the contribution ❤️

mokagio:
Thanks @joshdholtz for the review and for your kind words. The test you added is brilliant. 👏  TIL you can call `.once` more than once.

PR #18153, opened 2021-02-08
Fix "undefined method 'each' for NilClass" error with PushToGitRemoteAction

neilco:
When calling `PushToGitRemoteAction.run(…)` from another action without the `push_options` parameter, it will result in the following exception:

```ruby
NoMethodError: [!] undefined method `each' for nil:NilClass
```

This pull request fixes that problem by only enumerating `params[:push_options]` if it has a value.

### Checklist

### Motivation and Context

When calling `PushToGitRemoteAction` from another action without the `:push_options` parameter, `params[:push_options]` evaluates to `nil`. This results in `params[:push_options].each` failing with the following exception:

```ruby
NoMethodError: [!] undefined method `each' for nil:NilClass
```

### Description

The fix supplied in this pull request checks that `params[:push_options]` holds value before enumerating it. 

This was tested by creating an action that calls `PushToGitRemoteAction` without the `push_options` argument. 

### Testing Steps

- Create a new action that calls `PushGitToRemoteAction`. I used the following action:

    ```ruby
    module Fastlane
      module Actions
        class NeilcoGitBugAction < Action
          def self.run(params)
              require 'time'
    
              files = ["test_file.txt"]
    
              files.each do |filename|
                unless File.exist?(filename)
                  FileUtils.touch(filename)
                end
    
                File.open(filename, 'w') do |f|
                  f.write("Testing at #{Time.now()}")
                end
              end
    
              GitAddAction.run(path: files)
              GitCommitAction.run(path: files, message: "#{params[:message]} [ci skip]")
              PushToGitRemoteAction.run(tags: true, remote: "origin")
          end
    
          def self.available_options
            [
              FastlaneCore::ConfigItem.new(key: :message,
                                           env_name: "FL_NEILCO_GIT_BUG_MESSAGE", # The name of the environment variable
                                           description: "Git commit message", # a short description of this parameter
                                           is_string: true,
                                           optional: false
                                           )
            ]
          end
        end
      end
    end
    ```
- Create a lane that calls this new action:

    ```ruby
    lane :git_bug do
        neilco_git_bug(message: "Testing, testing")
    end
    ```

- Run `bundle exec fastlane git_bug`

### DIFF
--- a/fastlane/lib/fastlane/actions/push_to_git_remote.rb
+++ b/fastlane/lib/fastlane/actions/push_to_git_remote.rb
@@ -33,7 +33,7 @@ module Fastlane
         command << '--set-upstream' if params[:set_upstream]
 
         # optionally add the --push_options components
-        params[:push_options].each { |push_option| command << "--push-option=#{push_option}" }
+        params[:push_options].each { |push_option| command << "--push-option=#{push_option}" } if params[:push_options]
 
         # execute our command
         Actions.sh('pwd')


neilco:
@googlebot I signed it!

minuscorp (approved):
Nice catch! LGTM🚀

joshdholtz (approved):
I like this safety here! Thank you for the contribution ❤️ 

One fun tip... is that you should call this like 👇 

```rb
other_action.push_to_get_remote(your_option)
```

Docs here - http://docs.fastlane.tools/advanced/actions/#calling-other-actions

This adds the config item handling login with defaults and everything which is why this wasn’t caught before 🙃

Issue #17910, opened 2021-01-06
Undefined method `status' for nil:NilClass (NoMethodError)


cheasley-dvs:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I am using a fastlane API Key JSON to upload our builds to TestFlight as a way around 2FA that is being required by Apple by Feb 2021. 
The build successfully uploads, then the build step will fail afterwards with this error: [!] undefined method `status' for nil:NilClass (NoMethodError)

##### Command executed
fastlane pilot upload --api_key_path /5FF9D7A7M3.json

API JSON looks like:
`{
  "key_id": "Our key id",
  "issuer_id": "Our issuer ID",
  "key": "-----BEGIN PRIVATE KEY-----\nOur key\n-----END PRIVATE KEY-----"
}`

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> 
[21:12:58]The build is removed from the queue to be prepared for the start
[21:12:58]Collecting changes in 1 VCS root (1s)
[21:12:59]Starting the build on the agent "MAC AGENT"
[21:12:59]Updating tools for build
[21:12:59]Clearing temporary directory: /Applications/buildAgent/temp/buildTmp
[21:12:59]Publishing internal artifacts (3s)
[21:12:59]Full checkout enforced. Reason: [Checkout directory is empty or doesn't exist]
[21:12:59]Will perform clean checkout. Reason: Checkout directory is empty or doesn't exist
[21:12:59]Checkout directory: /Applications/buildAgent/work/23b883b82c08ec30
[21:12:59]Updating sources: auto checkout (on agent) (11s)
[21:13:11]Step 1/8: Restore Nuget Packages (Command Line) (12s)
[21:13:24]Step 2/8: Update Fastlane (Command Line) (5s)
[21:13:30]Step 3/8: Change plist to Staging Test (Command Line)
[21:13:30]Step 4/8: Change entitlements to staging test (Command Line)
[21:13:30]Step 5/8: Update Version Number Staging Test (Command Line)
[21:13:30]Step 6/8: Update Build Number Staging Test (Command Line)
[...]

PR #18158, opened 2021-02-09
[pilot] fix default value of `demo_account_required` to nil instead of false

acasademont:
### Checklist

### Motivation and Context
As discussed in #17910 pilot seems to be unable to work with an api_key with just the Developer permissions, needing App Manager permissions instead. However, it should work with the Developer permissions if no distribution of the uploaded binary is performed. This is the first in a series of patches.

### Description
The `demo_account_required` has a default value of `false` but then on the code that chacks for its presence https://github.com/fastlane/fastlane/blob/832e3e4a19d9cff5d5a14a61e9614b5659327427/pilot/lib/pilot/build_manager.rb#L217 it performs a null check that always evaluates to true (`false` is not `nil`) and that causes an API call that needs the said App Manager permissions. 

With this patch, I propose that option be made optional (and thus nil if not set) so that only when an explicit value is set, the API call is performed, avoiding the need for App Manager permissions.

### Testing Steps
Before: If the option was not set, the `false` default would also trigger an API call
Now: If the option is not set, the `nil` value will prevent an API call.

### DIFF
--- a/pilot/lib/pilot/options.rb
+++ b/pilot/lib/pilot/options.rb
@@ -88,7 +88,7 @@ module Pilot
                                      type: Boolean,
                                      env_name: "DEMO_ACCOUNT_REQUIRED",
                                      description: "Do you need a demo account when Apple does review?",
-                                     default_value: false),
+                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :beta_app_review_info,
                                      type: Hash,
                                      env_name: "PILOT_BETA_APP_REVIEW_INFO",
--- a/pilot/spec/build_manager_spec.rb
+++ b/pilot/spec/build_manager_spec.rb
@@ -425,6 +425,40 @@ describe "Build Manager" do
     end
   end
 
+  describe "#update_beta_app_meta" do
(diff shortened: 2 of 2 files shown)


minuscorp:
I'm wondering if this kind of cases should be handled as exceptions instead of silently pass. If this parameter is set to `true` or `false` but you have no permissions, what happens in the submission?

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for catching this! I added a commit to add some tests around this but this chagne is 🔥 Really appreciate the contribution! ❤️

acasademont:
Thanks to you @joshdholtz, the idea is to keep working on small patches to allow uploading testflight versions with Developer api key permissions instead of App Manager ones. More to come :)

acasademont:
> I'm wondering if this kind of cases should be handled as exceptions instead of silently pass. If this parameter is set to `true` or `false` but you have no permissions, what happens in the submission?

It will throw an exception if there are no permissions on the api key for that api operation, it's what was happening to me, hence the fix!

PR #18173, opened 2021-02-10
[fastlane_core] env_names - new config item option

joshdholtz:
### Motivation and Context
This is the first PR for #18145 for consolidating the App Store Connect API Key across multiple tools/actions.
Currently, each tool/action has its own environment variable. This will allow one variable to be set an used by all.

There will be a follow up PR that will actually modify all the tools/actions with this new option

### Description
- Adds optional `env_names` option to `FastlaneCore::ConfigItem` 
- Moves fetching of `ENV` value from `FastlaneCore::Configuration` to `FastlaneCore::ConfigItem`
  - New `fetch_env_value` method first looks for `env_name` and then looks for `env_names`
- Updated action list to show all environment variables
- Updated command help to show all environment variables

#### Action List
AF35996B-F14B-43F8-96B1-98014DDE5C3D

#### Command Help
23EC166A-8484-46DA-8DE0-9A8FBAB1FEB9

### DIFF
--- a/fastlane/swift/formatting/Brewfile.lock.json
+++ b/fastlane/swift/formatting/Brewfile.lock.json
@@ -31,9 +31,9 @@
   "system": {
     "macos": {
       "catalina": {
-        "HOMEBREW_VERSION": "3.0.0-57-g59dd425",
+        "HOMEBREW_VERSION": "3.0.0-59-g4695220",
         "HOMEBREW_PREFIX": "/usr/local",
-        "Homebrew/homebrew-core": "637196b33ecd885dc057c16fc5cb652a4561f632",
+        "Homebrew/homebrew-core": "7fd600e6644b774e47e12b13ba1ccf2f2ede6c67",
         "CLT": "11.0.33.12",
         "Xcode": "12.2",
         "macOS": "10.15.7"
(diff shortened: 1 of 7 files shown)


fastlane_core/lib/fastlane_core/configuration/config_item.rb
```diff
@@ -138,6 +149,7 @@ def initialize(key: nil,
 
       @key = key
       @env_name = env_name
+      @env_names = [env_name].compact + (env_names || [])
```
lacostej, line 152: maybe use `([env_name] + (env_names || [])).compact` in case `env_names` contain `nil`.

`([env_name] + env_names).compact` would also work by using the side effect of `compact`. Maybe too smart.

fastlane_core/spec/configuration_spec.rb
```diff
@@ -804,14 +804,44 @@
             ENV.delete("abc")
           end
 
-          it "prioritizes config file values after ENV" do
+          it "prioritizes env_names after CLI" do
+            ENV["abc"] = "val env"
```
lacostej, line 808: you might want to use `with_env_values` to avoid having to set / reset the environment. 

maybe for a future cleanup. I see  it uses this pattern elsewhere already.

fastlane/swift/formatting/Brewfile.lock.json
```diff
@@ -31,9 +31,9 @@
   "system": {
     "macos": {
       "catalina": {
-        "HOMEBREW_VERSION": "3.0.0-57-g59dd425",
+        "HOMEBREW_VERSION": "3.0.0-59-g4695220",
```
lacostej, line 34: this change doesn't seem related.

lacostej (commented):
looks good.

PR #18186, opened 2021-02-12
[app store connect] cleanup token creation code across entire project

joshdholtz:
### Motivation and Context
Adds on to #18173
Partly resolves #18145

### Description
- Commonizes the Spaceship::ConnectAPI::Token creation across entire projects
- Changed `app_store_connect_api_key` action
  - Removed `SharedValues::APP_STORE_CONNECT_API_KEY`
  - Replaced with setting `Spaceship::ConnectAPI.token`


fastlane/lib/fastlane/actions/app_store_connect_api_key.rb
```diff
@@ -2,10 +2,6 @@
 
 module Fastlane
   module Actions
-    module SharedValues
-      APP_STORE_CONNECT_API_KEY = :APP_STORE_CONNECT_API_KEY
```
lacostej, line 6: do we really want to break API for users who might depend on this?

match/lib/match/importer.rb
```diff
@@ -149,9 +149,9 @@ def ensure_valid_file_path(file_path, file_description, file_extension, optional
     end
 
     def api_token(params)
-      @api_token ||= Spaceship::ConnectAPI::Token.create(params[:api_key]) if params[:api_key]
-      @api_token ||= Spaceship::ConnectAPI::Token.from_json_file(params[:api_key_path]) if params[:api_key_path]
-      return @api_token
+      api_token = Spaceship::ConnectAPI::Token.from(hash: params[:api_key], filepath: params[:api_key_path])
```
lacostej, line 152: in this class (and a few below) we are not consistent with the way we log about wether we reuse or create a token.

For example the implementation in `fastlane/lib/fastlane/actions/set_changelog.rb` treats them differently

fastlane/lib/fastlane/actions/app_store_connect_api_key.rb
joshdholtz: I was unsure about this. I didn’t super want to keep it but I also didn’t want to break 😬 I can put it back in!

match/lib/match/importer.rb
joshdholtz: `match` has a bunch of tools that are only called from the CLI as sub commands. So this one should never really have token to reuse already 🤷‍♂️  But I can make these look and reuse a token if there is an existing one!

fastlane/lib/fastlane/actions/check_app_store_metadata.rb
```diff
@@ -7,7 +7,6 @@ class CheckAppStoreMetadataAction < Action
       def self.run(config)
         require 'precheck'
         Precheck.config = config
-        Precheck.config[:api_key] ||= Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
```
joshdholtz, line 10: Not needed anymore because `Precheck` will look for existing `Spaceship::ConnectAPI.client` now and use that

fastlane/lib/fastlane/actions/get_certificates.rb
```diff
@@ -13,7 +13,6 @@ def self.run(params)
 
         begin
           Cert.config = params # we alread have the finished config
-          Cert.config[:api_key] ||= Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
```
joshdholtz, line 16: Not needed anymore because `Cert` will look for existing `Spaceship::ConnectAPI.client` now and use that

fastlane/lib/fastlane/actions/get_provisioning_profile.rb
```diff
@@ -15,7 +15,6 @@ def self.run(values)
         require 'credentials_manager/appfile_config'
 
         Sigh.config = values # we already have the finished config
-        Sigh.config[:api_key] ||= Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
```
joshdholtz, line 18: Not needed anymore because `Sigh` will look for existing `Spaceship::ConnectAPI.client` now and use that

fastlane/lib/fastlane/actions/sync_code_signing.rb
```diff
@@ -10,7 +10,6 @@ def self.run(params)
         require 'match'
 
         params.load_configuration_file("Matchfile")
-        params[:api_key] ||= Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
```
joshdholtz, line 13: Not needed anymore because `Match` will look for existing `Spaceship::ConnectAPI.client` now and use that

fastlane/lib/fastlane/actions/upload_to_testflight.rb
```diff
@@ -15,10 +15,6 @@ def self.run(values)
           values[:ipa] = File.expand_path(values[:ipa]) if values[:ipa]
         end
 
-        if values[:api_key_path].nil?
-          values[:api_key] ||= Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
-        end
```
joshdholtz, line 20: Not needed anymore because `Pilot` will look for existing `Spaceship::ConnectAPI.client` now and use that

match/lib/match/importer.rb
joshdholtz: lol, I read your original comment wrong 😛  I will make this change!

joshdholtz:
@lacostej I think this is ready for a re-review (finally 😅)

Issue #17931, opened 2021-01-09
📣 Plan for Ruby 3.0 support (Released!)


ainame:
👋 Hello, fastlane community!

Ruby 3.0 support was released at v2.183.0. Please give it a try and give us feedback on this issue.
https://github.com/fastlane/fastlane/releases/tag/2.183.0

The plan is now hidden under "Details".
<details>

**Please react with :+1:  emoji on this issue if you are waiting for Ruby 3.0 support!**

This is @ainame who recently became a core contributor 🚀  I'm pleased to announce that I will be working on getting Ruby 3.0 support in fastlane. So this ticket is to track the progress and let fastlane users be aware of the work being addressed.

## TL; DR

* Use Ruby 2.4-2.7 until 3.0 migration is done
* Don't send us a bug report about installation failure with Ruby 3.0 as it will cause duplication a lot
* Help us if we need to test any for the migration in future :pray:

## Ruby 3.0 and currently supported versions

[Ruby 3.0 was released on 25th December 2020](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/). It's a major version update but its source code compatibility with 2.x should be good. There are performance improvements and interesting new features implemented.

However, fastlane hasn't officially supported Ruby 3.0 yet. Please read the below section for details. The currently supported versions are Ruby 2.4-2.7. I highly recommend that you upgrade Ruby version to 2.6 or 2.7. Ruby 2.4 is now in EOL period, which means there will be no security fix for it. Ruby 2.5 will be the same soon.

https://www.ruby-lang.org/en/downloads/branches/ 

If you try to install fastlane against Ruby 3.0, it actually succeeds with the version "2.54.1" BUT it is too old and `deliver` won't work for the latest App Store Connect APIs. Please don't use it.

### Ruby 2.7's warnings

Before going to Ruby 3.0, we should make sure the last version in 2.x series "2.7" is supported correctly. It's very important because 2.7 warns us if there are chances of breaking with some deprecation messages to prepare Ruby 3.0 migration. https://github.com/fastlane/fastlane/issues/15856

It's been recognised, in fact, fastlane can run on Ruby 2.7, which is the latest version in 2.x series. However, despite the fact, fastlane project hasn't guaranteed that all the testing specs pass with Ruby 2.7 on CI yet. This will be resolved on https://github.com/fastlane/fastlane/pull/17861

### Blocker to install fastlane to Ruby 3.0

[...]

PR #18189, opened 2021-02-12
[docs] Remove "sudo" from instructions in code

ainame:
### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/issues/17931

Prior to this PR, I've updated https://docs.fastlane.tools/getting-started/ios/setup/ page to not recommend "sudo gem install" habit officially. 
https://github.com/fastlane/docs/pull/1022

This PR removes hardcoded instructions displaying "sudo gem install" in order to match the document. Although I feel like some of them need to be revised due to the fact we now recommend "Bundler" as the first option, I guess there would be no harm to get rid of "sudo" from the code.

I might work on improving how the instructions are later on. For example,  we might need to guide users to use `bundle update xxx` instead of `gem update xxx`.

### DIFF
--- a/fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md
@@ -21,7 +21,7 @@ _screengrab_ generates localized screenshots of your Android app for different d
 Install the gem
 
 ```no-highlight
-sudo gem install fastlane
+gem install fastlane
 ```
 
 ##### Gradle dependency
--- a/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
+++ b/fastlane/lib/fastlane/actions/docs/frame_screenshots.md
@@ -345,5 +345,5 @@ brew install imagemagick
 
 ## Uninstall
 
-- `sudo gem uninstall fastlane`
+- `gem uninstall fastlane`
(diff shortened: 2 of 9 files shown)


minuscorp (commented):
This assumes that you have some gem manager (rvm, rbenv, bundler, etc) maybe we should explain that more concisely before running in lots of issues of no sufficient permissions when people tries to install it through system Ruby. Thoughts?

ainame:
@minuscorp 

> This assumes that you have some gem manager (rvm, rbenv, bundler, etc)

This is a half-truth. You could still install gems with system Ruby without having to use those managers by setting some own directory to `GEM_HOME` env. (Actually, we should add this to the document as an option 🙂) Given I've updated to recommend new users set up the manager it won't be harmful to them.

https://github.com/fastlane/docs/pull/1022

> maybe we should explain that more concisely before running in lots of issues of no sufficient permissions when people tries to install it through system Ruby. Thoughts?

That's the plan actually! I wanted instructions in code to match the document first as a concise explanation might need some time to prepare for me.

>> I might work on improving how the instructions are later on. For example, we might need to guide users to use bundle update xxx instead of gem update xxx.

minuscorp (approved):

joshdholtz (approved):
Yesssssssssssssss 🔥  Bye sudo 👋

PR #18235, opened 2021-02-18
[spaceship] allow extra fields in API Key JSON file

Katsz:
### Checklist

### Motivation and Context
Currently `Spaceship::ConnectAPI::Token.from_json_file` passes all json fields to `Spaceship::ConnectAPI::Token.create`. When managing multiple teams API key json is convenient place to store some extra information like `dev_team_id`, `itc_team_id` or similar. But in case if json contains extra fields `.create` will fail with `ArgumentError (unknown keywords: dev_team_id, itc_team_id)`

### Description
By adding vararg `**` to arguments of `.create` we can ignore any extra fields that are present. Worth noting that similar result may be achieved by filtering json output before passing to `.create`. 

### Testing Steps
Extra fields in API key json are not causing `ArgumentError`.

### DIFF
--- a/spaceship/lib/spaceship/connect_api/token.rb
+++ b/spaceship/lib/spaceship/connect_api/token.rb
@@ -40,7 +40,7 @@ module Spaceship
         self.create(json)
       end
 
-      def self.create(key_id: nil, issuer_id: nil, filepath: nil, key: nil, is_key_content_base64: false, duration: nil, in_house: nil)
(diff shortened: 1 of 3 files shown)


Katsz:
@googlebot I signed it!

joshdholtz (approved):
I like this change! I added a test for this to make sure things keep working in the future. Thanks for the contribution ❤️

PR #18239, opened 2021-02-19
[spaceship][spaceauth] fix SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER forcing 'sms' push mode

Hiroto-N:
### Checklist

### Motivation and Context
When running fastlane spaceauth with SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER set, because the `push_mode_from_masked_number` compares the full phone number with the masked phone number from the response from apple, it doesn't find a matching trusted phone number and defauts to pushMode = "sms".

### Description
The code handling the case of the environment variable being set was correctly handling the masked vs unmasked phone numbers, so this logic was refactored out to a method `match_phone_to_masked_phone` so that it could be shared for both the `phone_id_from_number` and the new `push_mode_from_number` methods.
### Testing Steps

Set up a phone on the apple account with the "Phone Call" option, then run
```
SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER="phone with voice" bundle exec fastlane spaceauth -u email@email.com
```

### DIFF
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -235,9 +235,7 @@ module Spaceship
 
     def phone_id_from_number(phone_numbers, phone_number)
       phone_numbers.each do |phone|
-
         return phone['id'] if match_phone_to_masked_phone(phone_number, phone['numberWithDialCode'])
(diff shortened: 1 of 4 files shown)


joshdholtz (approved):
This is so good! I can’t believe this wasn’t caught sooner 😱  Thank you so much for the contribution! Really appreciate it ❤️

PR #18251, opened 2021-02-22
[precheck] update error message for `Precheck cannot check In-app purchases with the App Store Connect API Key`

### LINKED ISSUE #18250: Precheck cannot check In-app purchases with the App Store Connect API Key
### New Issue Checklist

### Issue Description

When releasing through Fastlane, i'd run into an error: `fastlane_core/ui/interface.rb:141:in 'user_error!': [!] Precheck cannot check In-app purchases with the App Store Connect API Key (yet). Exclude In-app purchases from precheck or use Apple ID login (FastlaneCore::Interface::FastlaneError)`
I immediately jumped into my AppStoreConnect console however was confused to find the app didn't support any In App Purchases (IAP).

I dug a little deeper and saw mention of `in_app_purchases` being checked in the precheck stage. Since the app itself it quite simple, I don't run any of these. 

I was at a loss. But then I dug further and noticed the failing step was not pre-check, it was instead `upload_to_app_store`; 
image

I then dug deeper and found out that by default the build step `upload_to_app_store` runs precheck for me, and by default checks for IAP. 

This was causing my build to fail no matter what, regardless of if I supported it or not. 
By parsing the option `precheck_include_in_app_purchases: false` to my upload step like so: 

```
upload_to_app_store(
	precheck_include_in_app_purchases: false
)
```

I could now successfully release to the app store.

##### Command executed

`bundle exec fastlane ios release --verbose`

<details>
</details>

### Environment

<details>
</details>

builtbyproxy:
### Checklist
image

image

### Motivation and Context

Resolves #18250

### Description

I believe in providing solutions and not problems. I don't doubt we have this as `true` for a reason however with the recent additions to code signing with Apple's 2FA requirements, more people are moving to AppStoreConnect API Keys. This includes me and like me, many of us aren't expecting these checks. 

I'd like to change the default value (as a shot in the dark) or ideally update the error to highlight you may be calling `precheck` without even being aware of it.

### DIFF
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -450,7 +450,7 @@ module Deliver
                                      description: "Should precheck check in-app purchases?",
                                      is_string: false,
                                      optional: true,
-                                     default_value: true),
+                                     default_value: false),
 
         # internal
         FastlaneCore::ConfigItem.new(key: :app,
--- a/deliver/lib/deliver/options.rb
+++ b/deliver/lib/deliver/options.rb
@@ -450,7 +450,7 @@ module Deliver
                                      description: "Should precheck check in-app purchases?",
                                      is_string: false,
                                      optional: true,
-                                     default_value: false),
+                                     default_value: true),
 
         # internal
         FastlaneCore::ConfigItem.new(key: :app,
--- a/deliver/lib/deliver/runner.rb
+++ b/deliver/lib/deliver/runner.rb
@@ -83,7 +83,7 @@ module Deliver
 
       if options[:api_key] || options[:api_key_path]
         if options[:precheck_include_in_app_purchases]
-          UI.user_error!("Precheck cannot check In-app purchases with the App Store Connect API Key (yet). Exclude In-app purchases from precheck or use Apple ID login")
+          UI.user_error!("Precheck cannot check In-app purchases with the App Store Connect API Key (yet). Exclude In-app purchases from precheck, disable the precheck step in your build step, or use Apple ID login")
         end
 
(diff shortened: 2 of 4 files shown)


joshdholtz (changes requested):
Heyyy! I fought myself so hard on what this default value should be 😱  The reason I set it set it to `true` was so that it wouldn’t be a breaking change for anybody using Apple ID auth and that it would be a small additional change for people migrating to App Store Connect API Key.

Even though I doooooo agree with this, this will be a breaking change for people still using Apple ID 😔 

One thing I _could_ think of doing is just automatically skipping pre-check if there is a `Spaceship::ConnectAPI.token` available. One reason I went against this is this would then introduce a breaking change when deleting that auto bypass when the App Store Connect API does introduce IAP stuff.

So I _still_ think that even though this requires an option to be changed when migrating to the App Store Connect API Key, what we already have is the best approach for introducing  breaking changes now and in the future but 🤷‍♂️ 

Let me know your thoughts on my thoughts 😊

builtbyproxy:
Thanks for the awesome thoughts! @joshdholtz 

Alas I was expecting something around breaking changes to be the cause of this. That's great to hear it's been so deeply considered. 

What is Apple ID Auth VS App Store Connect API Keys? Is that just traditional username and password? 

I definitely don't love the idea of a temporary fix with a check around `Spaceship::ConnectAPI.token`. Seems like a lot more pain for the future. 

If you're happy with it, I would be very happy to just see the error message extended to highlight you may inadvertently be calling the pre-check without realising.

joshdholtz (approved):
I like this change! Thank you for updating after my review ❤️  Really appreciate the contribution

PR #18260, opened 2021-02-22
[jazzy] add module_version option

WezSieTato:
### Checklist

### Motivation and Context
Currently, if we want to use jazzy from fastlane we need to keep every parameter in the config file. With `module_version` option is possible to set this parameter dynamically during lane without changing the file.

### Description
Add `module_version` optional parameter to jazzy action. This is a version string to use as part of the default docs title and inside the docset.

### Testing Steps
Proper test was added to `jazzy_spec.rb`, you can check if the runned script is compatible with jazzy (https://github.com/realm/jazzy).


joshdholtz (approved):
Looks good to me! Thanks so much for adding this and updating the tests ❤️

PR #18279, opened 2021-02-25
[spaceship] resolve issue with current team_id being incorrectly reported

atyndall:
### Checklist

### Motivation and Context
The team_id method currently uses `teams[0]['contentProvider']['contentProviderId']` to determine the team ID that is in use.
`teams` ultimately derives from the `user_details_data` session information, but critically, it is manually sorted by name before being returned.
This means that the first element in `teams` does not have a semantic meaning, and may only by chance reflect the currently selected team.

`git blame` suggests this bug may have been present since the initial implementation of this code in 3ffaaf2d0. It makes sense that it may not have been detected, as I doubt many users use the team switching functionality, and you still have a 50:50 chance of it selecting the right one out of the box if you only have two teams, as it just depends on their alphabetical sorting.

### Description
This fix uses the `user_details_data['sessionToken']['contentProviderId']` directly to determine the current team, as this attribute always reflects the ID of the current team.
The `user_details_data` is typically cached for performance, but this data changes when switching team.
To prevent potential confusion and stale info this cache is cleared on manual team switch with `team_id=`.

### Testing Steps
You will need an Apple account with two teams joined, where the alphabetical team names don't happen to line up with the account selected by default.

Updated specs fail with old code and succeed with new one. Because they were directly mocking `teams`, they weren't testing how the sorting impacts the selection of the first element, thus masking this bug, which actually causes the example data already present to fail, as "Harry" sorts before "Tom".

Basic testing on our system indicates this fixes the issue, but I will be deploying this patch to our prod to verify the fix works more extensively in the coming days.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -134,7 +134,7 @@ module Spaceship
       if teams.count > 1
         puts("The current user is in #{teams.count} teams. Pass a team ID or call `select_team` to choose a team. Using the first one for now.")
       end
-      @current_team_id ||= teams[0]['contentProvider']['contentProviderId']
+      @current_team_id ||= user_details_data['sessionToken']['contentProviderId']
     end
 
     # Set a new team ID which will be used from now on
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
Oh... my goodness 😱 I cannot believe this worked like this for so long! I’ve tested this out on my account and this is 💯 

Thank you so much for finding this and fixing it ❤️ Really appreciate it!

PR #18285, opened 2021-02-25
[pilot] change itmsp saved name to allow multiple upload.

MaatheusGois:
### Checklist

### Motivation and Context
When you try send two apps to TestFlight at same moment, one upload gets broken because icmsp.token have the same name. 

### Description
- `ItunesTransporter.upload` takes an optional full path to the itms location (using `package_path` named argument)
  - Still is backwards compatible with `upload(app_id, dir)` that was previously used if anybody is using this directly
- Fixed `IpaUploadPackageBuilder` and `PkgUploadPackageBuilder` to return the full path used for ☝️ 
  -  Previously they returned the same path (only the directory) that was being passed in which didn’t really make sense

### Testing Steps
Proper test was added to helper_spec.rb, you can check if the runned script.


fastlane_core/spec/helper_spec.rb
```diff
@@ -147,6 +147,11 @@
         expect(FastlaneCore::Helper.transporter_path).to match(%r{/Applications/Xcode.*.app/Contents/Applications/Application Loader.app/Contents/itms/bin/iTMSTransporter|/Applications/Xcode.*.app/Contents/SharedFrameworks/ContentDeliveryServices.framework/Versions/A/itms/bin/iTMSTransporter})
       end
 
+      it "#path_name_itmsp", requires_xcode: true do
+        value = FastlaneCore::Helper.path_name_itmsp
+        expect(FastlaneCore::Helper.path_name_itmsp).to match(value)
```
minuscorp, line 152: This expectation is kinda redundant isn't it? You're checking the value with itself, which will be always true.

MaatheusGois: Hello @minuscorp thank you so much your comment!

Is expertly this, because when this value is called in the first time it generate one value and when it called again, this value must need de same, because the name path can't change 😁

When the _path_name_itmsp_ is set in _value_ is expect that is the same in the match.

Again, thank you so much for your comment!

minuscorp (approved):
LGTM! 🚀

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -207,6 +207,13 @@ def self.user_defined_itms_path?
       return FastlaneCore::Env.truthy?("FASTLANE_ITUNES_TRANSPORTER_PATH")
     end
 
+    def self.path_name_itmsp
+      unless @name_itms
+        @name_itms = ENV["FASTLANE_ITUNES_TRANSPORTER_PATH_ITMSP"] || "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx".gsub("x") { rand(16).to_s(16) }
+      end
```
joshdholtz, line 213:
```suggestiont
       @name_itms ||= ENV["FASTLANE_ITUNES_TRANSPORTER_PATH_ITMSP"] || SecureRandom.uuid
```

Thoughts on this? 😊

joshdholtz (changes requested):
One small nit pick!

But one other question... so with this change, the path is stored in `@name_itms` for an entire fastlane run and will only get set once. I’ve had some fastlane setups where I upload multiple apps during the same run. Previously it would place this in a different spot because of `app_id` but now with this change it will give them the same `itms` path 😬  Is that something of concern and going to be a breaking change for users that do something similar?

fastlane_core/lib/fastlane_core/helper.rb
MaatheusGois: Thank you so much for your comment @joshdholtz, it's really amazing!

I include in the path the **app_id** again,  this resolve your comment 😃?

joshdholtz:
@MaatheusGois Are you cool if I change this up a bit? 😇  

I want to change the `ItunesTranporter.upload` to also take a `path` (instead of just an `app_id`). I’m still a little uncomfortable with `@name_itms` being cached for an entire _fastlane_ run. The `generate` methods of `IpaUploadPackageBuilder` and `PkgUploadPackageBuilder` return the itms path and this should be passed directly into `upload` instead of the path being generated both places 🤷‍♂️

MaatheusGois:
Of course you can! @joshdholtz 

I'm a kid in Ruby programming, uahuahuaha, I'm here to learn from you! 😁

joshdholtz:
@MaatheusGois It doesn’t look like I have access to push to this repo since its not your personal repo 😱  Would you mind adding me as a collab so I can push to your branch? 😊

MaatheusGois:
@joshdholtz It's an honor for me!! 😁

joshdholtz:
@googlebot I consent.

joshdholtz:
@MaatheusGois @minuscorp I’ve updated this PR if you both want to take a look! 😊 

- `ItunesTransporter.upload` takes an optional full path to the itms location (using `package_path` named argument)
  - Still is backwards compatible with `upload(app_id, dir)` that was previously used if anybody is using this directly
- Fixed `IpaUploadPackageBuilder` and `PkgUploadPackageBuilder` to return the full path used for ☝️ 
  -  Previously they returned the same path (only the directory) that was being passed in which didn’t really make sense

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -471,11 +471,15 @@ def download(app_id, dir = nil)
...
     # @return (Bool) True if everything worked fine
     # @raise [Deliver::TransporterTransferError] when something went wrong
     #   when transferring
-    def upload(app_id, dir)
-      actual_dir = File.join(dir, "#{app_id}.itmsp")
+    def upload(app_id = nil, dir = nil, package_path: nil)
+      raise "app_id and dir are required or package_path is required" if (app_id.nil? || dir.nil?) && package_path.nil?
+
+      actual_dir = package_path
+      actual_dir ||= File.join(dir, "#{app_id}.itmsp")
```
minuscorp, line 482: `actual_dir = package_path || File.join(dir, "#{app_id}.itmsp")` ?

fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -471,11 +471,15 @@ def download(app_id, dir = nil)
...
+    # @param package_path [String] the path to the package file (used instead of app_id and dir)
     # @return (Bool) True if everything worked fine
     # @raise [Deliver::TransporterTransferError] when something went wrong
     #   when transferring
-    def upload(app_id, dir)
-      actual_dir = File.join(dir, "#{app_id}.itmsp")
+    def upload(app_id = nil, dir = nil, package_path: nil)
+      raise "app_id and dir are required or package_path is required" if (app_id.nil? || dir.nil?) && package_path.nil?
+
+      actual_dir = package_path
```
minuscorp, line 481: So this can go away

minuscorp (approved):
It looks great to me, didn't thought about all scenarios that could've been involved in the process 😄

fastlane_core/lib/fastlane_core/itunes_transporter.rb
joshdholtz: Yeha yeah yeah, I can change that up!

joshdholtz (approved):
@MaatheusGois Thanks again for starting this PR and reviewing my changes! I will get this out in a new version today 💪

PR #18289, opened 2021-02-26
[pilot] do not update app beta details if not needed

acasademont:
### Checklist

### Motivation and Context

As discussed in #17910 pilot seems to be unable to work with an api_key with just the Developer permissions, needing App Manager permissions instead. However, it should work with the Developer permissions if no distribution of the uploaded binary is performed. This is the second patch after #18158

### Description
If the `notify_external_testers` option is false there's no need to make the update build beta details API call. It's an API call that requires App Manager permissions. Following this patch, pilot can be used with just Developer permissions and have it wait for the build to finish uploading & processing using the following config:

`upload_to_testflight(skip_submission: true)`

### Testing Steps
Before the patch: the API call is always performed, despite not having any effects if `notify_external_testers` is false
After the patch: the API call is only performed when `notify_external_testers` is true (which is the default behaviour)

### DIFF
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -253,9 +253,11 @@ module Pilot
         end
       end
 
-      update_build_beta_details(build, {
-        auto_notify_enabled: options[:notify_external_testers]
-      })
+      if options[:notify_external_testers]
+        update_build_beta_details(build, {
+          auto_notify_enabled: options[:notify_external_testers]
+        })
+      end
     end
 
     def self.truncate_changelog(changelog)
--- a/pilot/lib/pilot/build_manager.rb
+++ b/pilot/lib/pilot/build_manager.rb
@@ -253,7 +253,9 @@ module Pilot
         end
       end
 
-      if options[:notify_external_testers]
+      if options[:notify_external_testers].nil?
+        UI.important("Using App Store Connect's default for notifying external testers (which is true) - set `notify_external_testers` for full control")
+      else
         update_build_beta_details(build, {
           auto_notify_enabled: options[:notify_external_testers]
         })
--- a/pilot/lib/pilot/options.rb
+++ b/pilot/lib/pilot/options.rb
@@ -182,7 +182,7 @@ module Pilot
         FastlaneCore::ConfigItem.new(key: :notify_external_testers,
                                     is_string: false,
                                     env_name: "PILOT_NOTIFY_EXTERNAL_TESTERS",
-                                    description: "Should notify external testers?",
+                                    description: "Should notify external testers? (Not setting a value will use App Store Connect's default which is to notify)",
                                     optional: true),
         FastlaneCore::ConfigItem.new(key: :app_version,
                                      env_name: "PILOT_APP_VERSION",
--- a/pilot/lib/pilot/options.rb
+++ b/pilot/lib/pilot/options.rb
@@ -183,7 +183,7 @@ module Pilot
                                     is_string: false,
                                     env_name: "PILOT_NOTIFY_EXTERNAL_TESTERS",
                                     description: "Should notify external testers?",
-                                    default_value: true),
+                                    optional: true),
         FastlaneCore::ConfigItem.new(key: :app_version,
                                      env_name: "PILOT_APP_VERSION",
                                      description: "The version number of the application build to distribute. If the version number is not specified, then the most recent build uploaded to TestFlight will be distributed. If specified, the most recent build for the version number will be distributed",
--- a/pilot/spec/build_manager_spec.rb
+++ b/pilot/spec/build_manager_spec.rb
@@ -441,6 +441,15 @@ describe "Build Manager" do
     it "does not attempt to set demo account required" do
       options = {}
 
+      expect(fake_build_manager).not_to receive(:update_review_detail)
+      expect(fake_build_manager).not_to receive(:update_build_beta_details)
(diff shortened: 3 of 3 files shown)


pilot/lib/pilot/build_manager.rb
```diff
@@ -336,6 +338,10 @@ def should_update_localized_build_information?(options)
       !options[:localized_build_info].nil?
     end
 
+    def should_update_build_beta_details?(options)
+      options[:notify_external_testers]
+    end
```
joshdholtz, line 343: Is this method going to get reused in the future? If not... I think I’d prefer this logic to just be moved into the if statement on line 256 since this is pretty specific to that call of `update_build_beta_details` 😊

joshdholtz (changes requested):

pilot/lib/pilot/build_manager.rb
acasademont: @joshdholtz sure! I was just following what I thought to be the convention, writing these should_* methods. I'll update the patch asap, thx!

acasademont:
@joshdholtz done!

joshdholtz:
@googlebot I consent.

pilot/lib/pilot/build_manager.rb
```diff
@@ -253,9 +253,11 @@ def update_beta_app_meta(options, build)
...
       end
 
-      update_build_beta_details(build, {
-        auto_notify_enabled: options[:notify_external_testers]
-      })
+      if options[:notify_external_testers]
+        update_build_beta_details(build, {
+          auto_notify_enabled: options[:notify_external_testers]
+        })
+      end
```
joshdholtz, line 260: @acasademont It turns out that `auto_notify_enabled` is defaulted to `true` so this logic will actually always notify even when `notify_external_testers` is set to `false`

One thing we _could_ do is 👇  but this makes a pretty big assumption about how App Store Connect API works and will break if we change it 🤔 

_Maybe_ we end up making `:notify_external_testers` optional and only run this request if its explicitly set? This would work because it would apply App Store Connect APIs default but `true` and `false` would do exactly what its supposed to.

Thoughts?

```
      unless options[:notify_external_testers]
        update_build_beta_details(build, {
          auto_notify_enabled: options[:notify_external_testers]
        })
      end
```

joshdholtz (changes requested):
This doesn’t work as expected but might have a solution/suggestions that works? 🤷‍♂️

pilot/lib/pilot/build_manager.rb
acasademont: Hi @joshdholtz , I'm afraid I don't really understand this:

> @acasademont It turns out that `auto_notify_enabled` is defaulted to `true` so this logic will actually always notify even when `notify_external_testers` is set to `false`

If `notify_external_testers` is explicitly set false the API call  won't be made (i've used it quite a lot already in my dev environment)-
> 
> One thing we _could_ do is 👇 but this makes a pretty big assumption about how App Store Connect API works and will break if we change it 🤔
> 
> _Maybe_ we end up making `:notify_external_testers` optional and only run this request if its explicitly set? This would work because it would apply App Store Connect APIs default but `true` and `false` would do exactly what its supposed to.
> 
> Thoughts?
> 
> ```
>       unless options[:notify_external_testers]
>         update_build_beta_details(build, {
>           auto_notify_enabled: options[:notify_external_testers]
>         })
>       end
> ```

This is precisely what I wanted to avoid using the current default value, it needs to be explicitly set to `false`.

joshdholtz: @acasademont  Sorry, I should have been more clear!

```rb
 if options[:notify_external_testers]
```

The current code in the PR has this ☝️ `options[:notify_external_testers]` is a boolean (so either `true` or `false`). So this means that `update_build_beta_details` will only get called if `options[:notify_external_testers]` is set to `true`

This is problematic because App Store Connect (not _fastlane_) defaults the “notifying of external testers” to `true`. So...

If `options[:notify_external_testers]` is set to `false`, this if statement won’t get executed and App Store Connect will still notify external testers because it itself will keep it default to `true`.

My thoughts on a solution would be to:
1) Set the `notify_external_testers` to optional
2) Change the if statement to `unless options[:notify_external_testers].nil?`

So with this, `update_build_beta_details` will only get called if explicitly set and would retain backwards compatibility because App Store Connect defaults notifying external testers to true 😊

acasademont: aaah, gotcha, sorry for that!! Then, yes, definititely your suggestion is the way to go. And although it's a BC-break in terms of the fastlane API, it won't have any effect in terms of the actual notification to external testers if the App Store default is to notify the external testers.

Should I work on those changes?

joshdholtz: @acasademont I don’t think its a _full_ break since the option defaults to `true` right now and App Store Connect currently also defaults it to `true` 🤷‍♂️  But it _could_ break if App Store Connect changes it but at least it can be explicitly set with ease 😊 

If you want to make this change that would be great! I’d be happy to test it out again ❤️

acasademont: @joshdholtz done! thank you :D

joshdholtz (approved):
I think this is good to go now! Thank you for working through this with me 😊 I’m glad this worked out! Really appreciate your contribution ❤️

acasademont:
Thanks to you @joshdholtz ! It's now finally possible to upload to testflight with the Developer permissions only with:

`upload_to_testflight(skip_submission: true)`



`

PR #18294, opened 2021-02-27
[spaceship] make ensure_version pass the client to get_edit_app_store_version

### LINKED ISSUE #18288: ensure_version fails with nil exception
I think there's a bug here - it needs to pass the given client to the call to get_edit_app_store_version.

https://github.com/fastlane/fastlane/blob/8dcb94a95a3661c6d2945563890ce628670db3bc/spaceship/lib/spaceship/connect_api/models/app.rb#L206

BoaNeo:
Without this you will get a nil exception if you supply a client instead of relying on Spaceship::ConnectAPI

### Checklist

### Motivation and Context

Because you get a nil exception when supplying a client otherwise

### Description

Pass the given client to get_edit_app_store_version instead of the default value

I ran my existing fastlane project

### DIFF
--- a/spaceship/lib/spaceship/connect_api/models/app.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app.rb
@@ -203,7 +203,7 @@ module Spaceship
       # @return (Bool) Was something changed?
       def ensure_version!(version_string, platform: nil, client: nil)
         client ||= Spaceship::ConnectAPI
-        app_store_version = get_edit_app_store_version(platform: platform)
+        app_store_version = get_edit_app_store_version(client:client,platform: platform)
 
         if app_store_version
           if version_string != app_store_version.version_string
--- a/spaceship/lib/spaceship/connect_api/models/app.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app.rb
@@ -203,7 +203,7 @@ module Spaceship
       # @return (Bool) Was something changed?
       def ensure_version!(version_string, platform: nil, client: nil)
         client ||= Spaceship::ConnectAPI
-        app_store_version = get_edit_app_store_version(client:client,platform: platform)
+        app_store_version = get_edit_app_store_version(client: client, platform: platform)
 
         if app_store_version
           if version_string != app_store_version.version_string


BoaNeo:
@googlebot I signed it!

spaceship/lib/spaceship/connect_api/models/app.rb
```diff
@@ -203,7 +203,7 @@ def reject_version_if_possible!(client: nil, platform: nil)
       # @return (Bool) Was something changed?
       def ensure_version!(version_string, platform: nil, client: nil)
         client ||= Spaceship::ConnectAPI
-        app_store_version = get_edit_app_store_version(platform: platform)
+        app_store_version = get_edit_app_store_version(client: client, platform: platform)
```
joshdholtz, line 206: @BoaNeo Do you happen to have a stacktrace of the nil exception? We _shouldn’t_ need to pass in `client` here because ‘ get_edit_app_store_version’ will set `client` if its not passed in the same way this method does 🤔

joshdholtz (changes requested):

spaceship/lib/spaceship/connect_api/models/app.rb
BoaNeo: I'm no Ruby expert, I don't even like Ruby :), but unless there's some language-level magic going on that I don't understand, I think you are reading the code wrong.

If the client is not passed, you are correct that it will default to the same value in both functions, but that is not the use case that is broken. 

The case that doesn't work is when I do pass a client. In this case the code will use that client only in this function but then pass nil to get_edit_app_store_version (which will then use the default, which is probably nil if I'm passing in a client of my own, and in any case probably isn't the one I want it to use).

Does that make sense?

joshdholtz: OMG I DID READ IT WRONG 😛  I most likely didn’t have enough coffee in me when I reviewed this 🤦‍♂️ 

You right you right, this is 💯 Thanks for explaining this to me!

joshdholtz (approved):
This is 💯 Thanks for fixing this! I appreciate the contribution ❤️

PR #18299, opened 2021-02-27
[spaceship] Better error handling of 403 status code when using API Key

### LINKED ISSUE #18193: [spaceship] 403 error clash with Portal and ConnectAPI
### New Issue Checklist

### Issue Description

When calling `Spaceship::ConnectAPI::App.find("dk.dinsen.secret.app")` using the API Key on an account that have license agreements waiting to be accepted.
I get this error: `undefined method 'status' for nil:NilClass`

I dived into the code and found that with `Spaceship::Portal` the error code 403 is handled here:
https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L896-L899

But because of this line: `raise AccessForbiddenError.new, msg` it goes into a loop where it makes 5x5 retries.
And after that it throw this error: `undefined method 'status' for nil:NilClass`

If I comment out this line `raise AccessForbiddenError.new, msg` then I'm able to add this into the `APIClient.handle_erros` and catch the error I'm looking for:
```ruby
if response.body['errors'].first['code'].eql?("FORBIDDEN.REQUIRED_AGREEMENTS_MISSING_OR_EXPIRED")
  raise ProgramLicenseAgreementUpdated, handle_errors(response) if response && (response.body || {})['errors']
else
  raise AccessForbiddenError, handle_errors(response) if response && (response.body || {})['errors']
end
```

I have not been able to catch the 403 using `Spaceship::Portal` since the Developer Portal is showing the `AppleIDAndPrivacyAcknowledgementNeeded` when trying to login.
So I'm not able to find a solution where we can handle this 403 status for both `Portal` and `ConnectAPI`

dinsen:
### Checklist

### Motivation and Context

Resolves #18193

When using the API Key for App Store Connect API and the account have new program license agreements that needs to be signed, I was getting this error:
`undefined method 'status' for nil:NilClass`

Motivation for this PR is to fix this error and show a proper error message when there is new agreements needed to be signed.

### Description

In the client class we already handle the 403 status code, but the raise of the error class is causing the `status for nil:NilClass` error, when using the API Key.
Adding a check if client is kind of `PortalClient`, will prevent the status nil error. 

I then added handling of 403 status error to get a proper error message when new agreements needs to be signed.

### Testing Steps

I have not been able to trigger the current error handling of the 403 status code. 

To test the new error handling, you will need an API Key to an account where license agreements needs to be signed, so you are able to get this error response:
```json
{
    "errors": [
        {
            "id": "6LVKXRE2VZG23UUYXXXXXXXXXX",
            "status": "403",
            "code": "FORBIDDEN.REQUIRED_AGREEMENTS_MISSING_OR_EXPIRED",
            "title": "A required agreement is missing or has expired.",
            "detail": "This request requires an in-effect agreement that has not been signed or has expired.",
            "links": {
                "see": "/agreements"
            }
        }
    ]
}
```

Run with: `bundle exec rake debug --trace`
```ruby
task(:debug) do
  require 'spaceship'

  token = Spaceship::ConnectAPI::Token.from_json_file(File.expand_path("./authkey.json"))
  Spaceship::ConnectAPI.token = token

  begin
    app = Spaceship::ConnectAPI::App.find("dk.dinsen.cool.app")
    puts(app.to_yaml)
  rescue Spaceship::ProgramLicenseAgreementUpdated => ex
    puts(ex)
  end
end
```


spaceship/lib/spaceship/connect_api/api_client.rb
```diff
@@ -199,8 +201,17 @@ def handle_response(response)
...
       end
 
-      def handle_401(response)
-        raise UnauthorizedAccessError, handle_errors(response) if response && (response.body || {})['errors']
+      def handle_4xx_errors(response)
+        case response.status.to_i
+        when 401
+          raise UnauthorizedAccessError, handle_errors(response) if response && (response.body || {})['errors']
+        when 403
+          if response.body['errors'].first['code'].eql?("FORBIDDEN.REQUIRED_AGREEMENTS_MISSING_OR_EXPIRED")
```
Dinsen, line 209: I really don't like this line! Any suggestions for doing this better is more than welcome 🙏

joshdholtz:
@googlebot I consent.

spaceship/lib/spaceship/client.rb
```diff
@@ -880,10 +880,7 @@ def send_request(method, url_or_path, params, headers, &block)
         response = @client.send(method, url_or_path, params, headers, &block)
         log_response(method, url_or_path, response, headers, &block)
 
-        resp_hash = response.to_hash
-        if resp_hash[:status] == 401
-          handle_401(response)
-        end
+        handle_error(response)
```
joshdholtz, line 883: Moved All error handling into `handle_error` method which can be overridden by other clients (like the connect client)

spaceship/lib/spaceship/connect_api/api_client.rb
```diff
@@ -159,15 +165,15 @@ def with_asc_retry(tries = 5, &_block)
 
         if [500, 504].include?(status)
           msg = "Timeout received! Retrying after 3 seconds (remaining: #{tries})..."
-          raise msg
+          raise TimeoutError, msg
```
joshdholtz, line 168: Created new error class so this can be specifically caught and handled in the retry logic

ainame: Although it's deprecated there is exactly the same error class in the standard library in Ruby. So it would be great to prefix here with the namespace to not confuse Ruby or rename the own error class🙂 

https://rubyapi.org/3.0/o/object/timeouterror

joshdholtz: Oh yikes! I’ll just rename it lol

joshdholtz: @ainame Updated it! Renamed it to `TimeoutRetryError` since... that’s kind of what it is 🤷‍♂️

joshdholtz:
@Dinsen Heyyy! Thanks for making this PR 😊 I ended up making some changes so all of the logic was contained in `Spaceship::ConnectAPI::APIClient` and not `Spaceship::Client`... and I added a bunch of tests because I wanted tests around these 🤷‍♂️ Would you be able to take a review of my changes when you get a chance?

Dinsen:
Hey @joshdholtz! 
This is way better 🙌 
Much easier to understand the code now and will be easier to extend it with more status codes in the future!

joshdholtz (approved):
Thanks for starting this and reviewing my changes! ❤️

PR #18325, opened 2021-03-03
[WIP] Adding additional options for AppAnalytics

max-ott:
### Checklist

### Motivation and Context
WIP


spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -567,10 +567,10 @@ def update_member_roles!(member, roles: [], apps: [])
     # @!group AppAnalytics
     #####################################################
 
-    def time_series_analytics(app_ids, measures, start_time, end_time, frequency, view_by)
+    def time_series_analytics(app_ids, dimensions, measures, start_time, end_time, frequency, view_by)
       data = {
         adamId: app_ids,
-        dimensionFilters: [],
+        dimensionFilters: dimensions,
```
max-ott, line 573: @joshdholtz The `dimensionFilters` actually allows the user to filter by country/storefront. The value look something like this: `[{"dimensionKey": "storefront", "optionKeys": [143441]}]` (143441 is US storefront). Just as a reference, even though Apple uses arrays, only one value is allowed 🤭

Question: How can we make sure that users can filter by country, but existing workflows don't break, because of too few attributes?

joshdholtz:
@max-ott I added `[WIP]` to the title for now 🤷‍♂️ Feel free to ping me when this is done!

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -567,10 +567,10 @@ def update_member_roles!(member, roles: [], apps: [])
     # @!group AppAnalytics
     #####################################################
 
-    def time_series_analytics(app_ids, measures, start_time, end_time, frequency, view_by)
+    def time_series_analytics(app_ids, dimensions, measures, start_time, end_time, frequency, view_by)
```
joshdholtz, line 570:
```suggestion
    def time_series_analytics(app_ids, measures, start_time, end_time, frequency, view_by, dimensions = [])
```

Move `dimensions` to the end and default it to an empty array

I would do something like this to allow backwards compatibility ☝️

joshdholtz (changes requested):

PR #18339, opened 2021-03-05
[ci] bump CI jobs to use Ruby 2.5 since some dependencies now require 2.5

joshdholtz:
### Motivation and Context
CI broke in #18328 because `signet` updated 0.15.0 to require Ruby 2.5+

### Description
Updating CI jobs to use Ruby 2.5

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -97,6 +97,7 @@ Gem::Specification.new do |spec|
   # If you upgrade this gem, make sure to upgrade the users of it as well.
   spec.add_dependency('google-api-client', '>= 0.37.0', '< 0.39.0') # Google API Client to access Play Publishing API
   spec.add_dependency('google-cloud-storage', '>= 1.15.0', '< 2.0.0') # Access Google Cloud Storage for match
+  spec.add_dependency('signet', '>= 0.14.0') # Signet 0.14.0 is needed for Ruby 2.4
 
   spec.add_dependency('emoji_regex', '>= 0.1', '< 4.0') # Used to scan for Emoji in the changelog
 
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -58,14 +58,13 @@ Gem::Specification.new do |spec|
 
   # need to lock under 0.15 using less than Ruby 2.5 to prevent install issues when using 'gem install'
   # 'gem install' does not respect Ruby versions and would try installing 0.15 on Ruby 2.4 or less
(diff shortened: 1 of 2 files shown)


fastlane.gemspec
```diff
@@ -56,6 +56,16 @@ Gem::Specification.new do |spec|
...
 
+  # need to lock under 0.15 using less than Ruby 2.5 to prevent install issues when using 'gem install'
+  # 'gem install' does not respect Ruby versions and would try installing 0.15 on Ruby 2.4 or less
+  # signet - https://github.com/googleapis/signet/commit/bd6fe87948f8fc7702720dae651e82f4fd348b5d
+  # googleauth - https://github.com/googleapis/google-auth-library-ruby/commit/6644806ab47cea6d08e1901c2ed808e53a579bc3
+  if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.5')
+    spec.add_dependency('signet', '<= 0.14.1')
+    spec.add_dependency('googleauth', '<= 0.15.1')
+    STDERR.puts("WARNING: Locking to a potentially insecure version of 'signet' and 'googleauth' because you are using a version of Ruby which is marked as End-Of-Life. Please upgrade your Ruby installation to 2.5 or later")
+  end
```
ainame, line 67: This is one of the known issues in RubyGem or rubygem.org but this branching based on `RUBY_VERSION` doesn't effectively work. Even worse this can cause unexpected result when releasing a new version of the gem.

* https://github.com/kamui/retriable/issues/79
* https://github.com/rubocop/ruby-style-guide/issues/763

So I would suggest doing followings

* Don't lock them in `gemspec` and let users choose their dependencies
* Move this logic in `Gemfile` in order to make sure fastlane developers/contributors can install proper dependencies
* Announce `gem install fastlane` or `gem "fastlane"` in `Gemfile` starts installing new dependencies that don't support Ruby 2.4 so you need to either
   * Lock `signet` and `googleauth` in your `Gemfile`  with `gem 'signet',  '<= 0.14.1'` and `gem 'googleauth', '<= 0.15.1'`
   *  Install old dependencies with `gem install signet -v 0.14.1` and `gem install googleauth -v 0.14` before running `gem install fastlane` or `gem update fastlane`

We can fix CI in the above way (installing old ones first) but I think it's fine just to bump the used Ruby version for that checks as Ruby 2.4. will be deprecated near future anyway. If we want to make sure everyone's installation works we should prepare a different kind of checks for CI.

ainame (changes requested):
👋 Nice catch the issue, however, this solution doesn't work as you expected. So I suggested a different way. (CI just passes because it uses Ruby 2.4 to build the gem file to be installed presumably.)

I'm happy to help get this resolved tonight (GMT) if you don't have enough time or need time to figure this out! 🤝

PR #18345, opened 2021-03-06
[action] create_xcframework - delete .xcframework if already exists

### LINKED ISSUE #18344: [create_xcframework] Creating fails if .xcframework already exists
### New Issue Checklist

### Issue Description

create_xcframework fails if .xcframework already exists in output folder.
Error:
`“<framework_name>.framework” couldn’t be copied to “ios-arm64” because an item with the same name already exists.`
##### Command executed

create_xcframework(
      frameworks: ["<path_to_framework>", "<path_to_second_framework>"],
      output: "<path_to_xcframework>"
    )

<details>
</details>

### Environment

<details>
</details>

Kukurijek:
### Checklist

### Motivation and Context

Resolves #18344
### Description
`create_xcframework` now checks if xcframework already exists in output folder, so you do not need to delete it manually.
If there is, delete it and display message "Deleting existing: <path_to_framework>"

Please add new label "xcframework" to github repo. 🙏

### Testing Steps

Try to create xcframework with command `create_xcframework` twice, without deleting the old one (with identical name).


minuscorp (changes requested):
One little comment about the failing tests, also, I'm worried about this maybe breaking an intentional behavior, so maybe another approach would be to add a `ConfigItem` to set the new behavior. Thoughts?

fastlane/lib/fastlane/actions/create_xcframework.rb
```diff
@@ -14,9 +14,15 @@ def self.run(params)
...
 
+        if File.directory?(params[:output])
+          UI.message("Deleting existing: #{params[:output]}")
+          FileUtils.remove_dir(params[:output])
+        end
+
         Actions.lane_context[SharedValues::XCFRAMEWORK_PATH] = params[:output]
 
         sh(create_command)
+        UI.message("Successfully created: #{params[:output]}")
```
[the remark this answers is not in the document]
Kukurijek: That is right. I seem to have run tests on some other branch 🤦🤦🤦
I could just move this line up, but it is unnecessary, because the action prints anyway the absolute path where it was created 😅
So I removed this line.
Sorry and thanks 👍 👍

minuscorp (commented):
Can you add tests regarding the behavior that is going to be introduced? 🙏

fastlane/spec/actions_specs/create_xcframework_spec.rb
```diff
@@ -118,6 +118,31 @@
...
+        it "should fail due to the deleted xcframework" do
+          begin
+            Fastlane::FastFile.new.parse("lane :test do
+              create_xcframework(
+                frameworks: ['FrameworkA.framework', 'FrameworkB.framework'],
+                output: 'UniversalFramework.xcframework'
+              )
+            end").runner.execute(:test)
+          rescue => e
+            expect(e.to_s).to eq("No such file or directory @ apply2files - UniversalFramework.xcframework").or(eq("No such file or directory @ unlink_internal - UniversalFramework.xcframework"))
```
minuscorp, line 139: Why is this being raised? It didn't supposed to delete the previous xcframework silently? Maybe I'm missing something here.

minuscorp (commented):
Just a comment, but seems good to me

minuscorp (approved):
LGTM🚀

fastlane/spec/actions_specs/create_xcframework_spec.rb
joshdholtz: @minuscorp @Kukurijek I add a commit to expect `FileUtils.remove_dir` to be called that removes the testing for failure here. I think this change should work 🤷‍♂️ 

I also updated some existing tests while I was in here 😛 The testing of failures now use an rspec way to catching and testing errors.

joshdholtz:
@googlebot I consent.

PR #18358, opened 2021-03-08
[xcov] Removed wrong default values.

ddeveloperuch:
### Checklist

### Motivation and Context

Wrong default parameters affects **xcov** functionality:

1. `onlyProjectTargets` is one of the parameters (only_project_targets , exclude_targets, include_targets), the simultaneous presence of which contradicts each other by design (e.g. values of `include_targets` conflict with `exclude_targets` and `only_project_targets` values). 
Current default value (`false`) implicitly limits using only `only_project_targets` and blocks to use `exclude_targets, include_targets` parameters.
2. Default value of `ideFoundationPath` contains a hardcoded Xcode version (12.4) and limits using only Xcode v12.4. Apart from that, native functionality of **xcov** perfectly detects the absolute path to the IDEFoundation.framework binary (based on `/usr/bin/xcode-select -p`). 
3. Default value of `minimumCoveragePercentage` doesn't make sense, native functionality of **xcov** has the same default value (_0.0_). Apart from that, due to an issue with parsing float value by RubyCommand. Argument struct (I haven't found proper info about that):
`'minimum_coverage_percentage' value must be a Float! Found Integer instead.`
it leads to the necessity to set the `minimum_coverage_percentage` to a value greater than zero explicitly.


minuscorp:
Fastlane.Swift values should not be changed from its definition, instead are the `ConfigItem` for the action which should be changed instead.

minuscorp:
We have to revisit the `ConfigItem`s of the Ruby implementation of the xcov action, because the Swift values are extracted from the Ruby implementation and then converted into Swift code, that's why changing the Swift code doesn't matter, it gets generated automatically. So any change in the source code won't get into a fix.

griganskiy:
> ...So any change in the source code won't get into a fix.

Now I see your point, thanks. 
Looks like `xcov.rb` has default action implementation and needs to be updated by adding a proper mapper of native parameters.

minuscorp:
That's it, you can change them in Ruby and reset the Swift changes, and generate the new interface using the lane `generate_swift_interface` to get how the Ruby changes are reflected into the Swift implementation.

griganskiy:
> ... and generate the new interface using the lane `generate_swift_interface` to get how the Ruby changes are reflected into the Swift implementation.

I got it, thanks. I will update PR later.

minuscorp:
Any news on this?

griganskiy:
> Any news on this?

Hello @minuscorp, give me a few days.

rogerluan:
@dgrigansky just a heads up, the file you're looking to change is this one https://github.com/fastlane-community/xcov/blob/master/lib/xcov/options.rb so it's outside of _fastlane_ itself 😬 that gem's `options` get parsed by _fastlane_ which then converts into the Swift API you see.

1. The issue with `onlyProjectTargets` conflicting with other options might get fixed in this PR: https://github.com/fastlane/fastlane/pull/17981. If not, I think the solution would be to remove the `default_value` in `xcov` `options.rb`
2. `ideFoundationPath`: not sure how we should fix this. Probably by also removing the `default_value` of the option, and reading it dynamically (and using the right path 😄 ) during runtime, instead of when the API is generated.
3. @minuscorp made a typo up there, he didn't mean `0,0`, but actually `0.0`. Looks like passing `0.0` explicitly _does_ fix the issue 👍 I've created a new Issue to track this problem alone https://github.com/fastlane/fastlane/issues/18433

Hope this shed some light into fixing `xcov` gem @dgrigansky 🙏 Looking forward to seeing your contributions 🤗

rogerluan:
Since this PR won't be able to accomplish what it initially proposed (since the changes should actually be done in xcov repo), I'm closing it.

Let us know if you still have questions or need any help 🙏 Feel free to tag me to review your next PR 💪

dgrigansky:
@rogerluan Thank you for the update, good to know that my PR led to creation of the additional PRs. 👍 
> ...the file you're looking to change is this one https://github.com/fastlane-community/xcov/blob/master/lib/xcov/options.rb so it's outside of _fastlane_ itself

I was working on `fastlane/lib/fastlane/actions/xcov.rb` 😆.

> ...
> 2. `ideFoundationPath`: not sure how we should fix this. Probably by also removing the `default_value` of the option ...

Yes, definitely by removing the `default_value` of the option.

I noticed some weird things (with current `fastlane` source (without any my changes)), I use `generate_swift_api` to get how the Ruby changes are reflected into the Swift implementation. I have success result:

```
+------+----------------------+-------------+
|             fastlane summary              |
+------+----------------------+-------------+
| Step | Action               | Time (in s) |
+------+----------------------+-------------+
| 1    | rake                 | 3           |
| 2    | gym                  | 20          |
| 3    | swift build          | 17          |
| 4    | rm -r ../.build      | 0           |
| 5    | rm ../FastlaneRunner | 0           |
+------+----------------------+-------------+

DEBUG [2021-03-22 12:15:46.56]: All plugins are up to date
INFO [2021-03-22 12:15:46.56]: fastlane.tools finished successfully 🎉
DEBUG [2021-03-22 12:15:46.56]: All plugins are up to date
```
but `Fastlane.swift` contains base implementation of `xcov`:

```
/**
 Nice code coverage reports without hassle

 Create nice code coverage reports and post coverage summaries on Slack *(xcov gem is required)*.
 More information: [https://github.com/nakiostudio/xcov](https://github.com/nakiostudio/xcov).
 */
public func xcov() {
    let command = RubyCommand(commandID: "", methodName: "xcov", className: nil, args: [])
    _ = runner.executeCommand(command)
}
```

Correct me if I'm wrong.

rogerluan:
@dgrigansky glad I was able to point you to the right direction before any further damage was done to your time 😆 🙈 

>but `Fastlane.swift` contains base implementation of `xcov`…

This doesn't happen here, when I run `fastlane generate_swift_api` in fastlane's master branch it generates the swift files correctly (including xcode with all of its parameters). Do you have `xcov` installed (either in your Gemfile, or globally in your system)?

dgrigansky:
> ... Do you have xcov installed (either in your Gemfile, or globally in your system)?

@rogerluan You was faster and found issue on my side, yes, I missed it and forgot to install on my second laptop.

Thanks for the answers I'm looking forward for the changes with `xcov` action.

PR #18373, opened 2021-03-10
[fastlane_core] fix package upload after two-step rescue.

lhunath:
### Checklist

### Motivation and Context
After two-step login fails and is rescued with an app-specific password,
the upload is not correctly restarted when it was initially initiated
with a package_path.

This results in a failure:
app_id and dir are required or package_path is required

### DIFF
--- a/fastlane_core/lib/fastlane_core/itunes_transporter.rb
+++ b/fastlane_core/lib/fastlane_core/itunes_transporter.rb
@@ -492,7 +492,7 @@ module FastlaneCore
         result = @transporter_executor.execute(command, ItunesTransporter.hide_transporter_output?)
       rescue TransporterRequiresApplicationSpecificPasswordError => ex
         handle_two_step_failure(ex)
-        return upload(app_id, dir)
+        return upload(app_id, dir, package_path: package_path)
       end
 
       if result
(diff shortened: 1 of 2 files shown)


fastlane_core/lib/fastlane_core/itunes_transporter.rb
```diff
@@ -492,7 +492,7 @@ def upload(app_id = nil, dir = nil, package_path: nil)
         result = @transporter_executor.execute(command, ItunesTransporter.hide_transporter_output?)
       rescue TransporterRequiresApplicationSpecificPasswordError => ex
         handle_two_step_failure(ex)
-        return upload(app_id, dir)
+        return upload(app_id, dir, package_path)
```
joshdholtz, line 495:
```suggestion
        return upload(app_id, dir, package_path: package_path)
```

joshdholtz (changes requested):
Ah, my apologies! Looks like I missed this 😱 The method needs a keyword argument when calling it so I made a suggestion for that but other than that this looks 💯

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks so much for fixing this! Really appreciate the contribution ❤️

PR #18387, opened 2021-03-12
[action][match] match_nuke action - Nuke your certificate and provisioning profiles (via match)

crazymanish:
### Checklist

#### Motivation
Allow user to use `api_key` / `api_key_path` to `match nuke` to use App Store Connect JWT Auth

#### Context
Problem1: Nuke certificate and provisioning profiles using match nuke CLI command is crazy 
```ruby 
fastlane match nuke development --passing here AppStoreConnect API key details 
```
Problem2: match nuke CLI command with `app_store_connect_api_key` (@joshdholtz's lovely action) is almost impossible 

### Description
- Introduce a new action `match_nuke` to easily nuke certificate and provisioning profiles (via _match_)

### Testing Steps

```ruby
lane :renew_certificate do
   ENV["MATCH_TYPE"] = "development"
   
  app_store_connect_api_key(
    key_id: "D383SF739",
    issuer_id: "6053b7fe-68a8-4acb-89be-165aa6465141",
    key_filepath: "./AuthKey_D383SF739.p8",
    in_house: false
  )

  # Automatically loads Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
  match_nuke
  match
end
```

Screenshot 2021-03-12 at 23 14 22

### DIFF
--- a/fastlane/lib/fastlane/actions/match_nuke.rb
+++ b/fastlane/lib/fastlane/actions/match_nuke.rb
@@ -1,6 +1,5 @@
 module Fastlane
   module Actions
-
     class MatchNukeAction < Action
       def self.run(params)
         require 'match'
--- a/fastlane/lib/fastlane/actions/match_nuke.rb
+++ b/fastlane/lib/fastlane/actions/match_nuke.rb
@@ -1,0 +1,60 @@
+module Fastlane
+  module Actions
+
+    class MatchNukeAction < Action
+      def self.run(params)
+        require 'match'
+
+        params.load_configuration_file("Matchfile")
+        params[:api_key] ||= Actions.lane_context[SharedValues::APP_STORE_CONNECT_API_KEY]
+
+        cert_type = Match.cert_type_sym(params[:type])
+        UI.important("Going to revoke your '#{cert_type}' certificate type and provisioning profiles")
+
+        Match::Nuke.new.run(params, type: cert_type)
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Easily nuke your certificate and provisioning profiles (via _match_)"
+      end
+
+      def self.details
+        [
+          "Use the match_nuke action to revoke your certificates and provisioning profiles.",
+          "Don't worry, apps that are already available in the App Store / TestFlight will still work.",
+          "Builds distributed via Ad Hoc or Enterprise will be disabled after nuking your account, so you'll have to re-upload a new build.",
+          "After clearing your account you'll start from a clean state, and you can run match to generate your certificates and profiles again.",
+          "More information: https://docs.fastlane.tools/actions/match/"
+        ].join("\n")
+      end
+
+      def self.available_options
+        require 'match'
+        Match::Options.available_options
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
+      end
+
+      def self.example_code
+        [
+          'match_nuke(type: "development")', # See all other options https://github.com/fastlane/fastlane/blob/master/match/lib/match/module.rb#L23
+          'match_nuke(type: "development", api_key: app_store_connect_api_key)'
+        ]
+      end
+
+      def self.authors
+        ["crazymanish"]
+      end
+
+      def self.category
+        :code_signing
+      end
+    end
+  end
+end
--- a/fastlane/spec/unused_options_spec.rb
+++ b/fastlane/spec/unused_options_spec.rb
@@ -71,6 +71,7 @@ describe Fastlane do
           validate_play_store_json_key
           update_fastlane
           s3
+          match_nuke
         )
       end
 
--- a/match/lib/match/module.rb
+++ b/match/lib/match/module.rb
@@ -21,6 +21,7 @@ module Match
   end
 
   def self.cert_type_sym(type)
+    type = type.to_s
     return :mac_installer_distribution if type == "mac_installer_distribution"
     return :developer_id_installer if type == "developer_id_installer"
     return :developer_id_application if type == "developer_id"


joshdholtz (commented):
I’m so torn about this 😱 I didn’t think I was going to like this as an action but... I kind of like this as an action 🤔 

I’m very conflicted on a few fronts...
1) The action sub commands (like `match nuke`) should be automatically made as actions somehow because that would make mike it easier on everyone so they don’t need to be created like this
2) The naming might need to also be aliased because of `match` and `sync_signing`. I mean... I think `match_nuke` makes enough sense but 🤷‍♂️ 
3) I think that `fastlane match nuke` _should_ be able to be easily called with the `--api_key_path` option. With this option you just need to point to a JSON file which contains the same info that the `app_store_connect_api_key` action takes - https://docs.fastlane.tools/app-store-connect-api/#using-fastlane-api-key-json-file

I’m do still like this PR and I just want to make sure that everything is thought about before merging 😊 

So if you have any thoughts/concerns on my thoughts/concerns that would be great to hear!

crazymanish:
Thank you very much for reviewing this PR! 🙇‍♂️ @joshdholtz 

> I’m so torn about this 😱 I didn’t think I was going to like this as an action but... I kind of like this as an action 🤔
> 
> I’m very conflicted on a few fronts...
> 
> 1. The action sub commands (like `match nuke`) should be automatically made as actions somehow because that would make mike it easier on everyone so they don’t need to be created like this

That will be really really great and perfectly make sense! but I am guessing it might require lots of testing and lots of effort. because of many sub-commands. maybe we can make a goal for 2021 or something in the Fastlane roadmap.

> 2. The naming might need to also be aliased because of `match` and `sync_signing`. I mean... I think `match_nuke` makes enough sense but 🤷‍♂️

I agree! I name it _match_nuke_ to be consistent with `fastlane match nuke`  sub-command and easily recognizable/readable by name what this action will do. 

> 3. I think that `fastlane match nuke` _should_ be able to be easily called with the `--api_key_path` option. With this option you just need to point to a JSON file which contains the same info that the `app_store_connect_api_key` action takes - https://docs.fastlane.tools/app-store-connect-api/#using-fastlane-api-key-json-file

The main problem is **Security team** does not allow JSON file under Git commit and most of the CI (i.e CircleCI, GitHub action) just support to store secrets in ENV variables ONLY. 
Of course, there is always a way, i.e Create the JSON file first using CI ENV variables and run the nuke sub-command but that will defeat the purpose of this action `Easily nuke certificate and provisioning profiles (via _match_)`


> 
> I’m do still like this PR and I just want to make sure that everything is thought about before merging 😊

Agree! Thank you very much for taking your time, really appreciate it! Also if this action does not make sense then don't worry I will close the PR 😉 [The best way is to support all subcommands as action, but I don't know yet how to do that 😬]

> 
> So if you have any thoughts/concerns on my thoughts/concerns that would be great to hear!

Please see some input inline and let me know if this action can merge or not. 😇
Thank you again!

joshdholtz:
@crazymanish 

>The main problem is Security team does not allow JSON file under Git commit and most of the CI (i.e CircleCI, GitHub action) just support to store secrets in ENV variables ONLY.

Ahhhh, this makes sense on why it’s almost impossible mentioned in your description 😊 Thanks for explaining that!

<hr/>

I’m having such a internal struggle with this lol 😛 I had two other thoughts that you _could_ put this in the `fastlane/actions` directory of your own project but... this is probably a thing that others would use. I then thought that you could make this a plugin but... its such a core _fastlane_ thing that it should just be made an action (which is where we are at right now 🙃 )

Making the sub commands all actions would be ideal but would be **a lot** of work sooooooo...

I think that last thing that is holding me back is making this almost **too easy** for people to delete all of their certificates when they really didn’t mean to. I use `match nuke` pretty often myself but its my own team where things really don’t matter. **Do you have any thoughts if making this too easy would cause any problems?**

ainame:
> The main problem is Security team does not allow JSON file under Git commit and most of the CI (i.e CircleCI, GitHub action) just support to store secrets in ENV variables ONLY.
Of course, there is always a way, i.e Create the JSON file first using CI ENV variables and run the nuke sub-command but that will defeat the purpose of this action Easily nuke certificate and provisioning profiles (via _match_)

I do understand this concern. However, it does support environment variables to inject App Store Connect API key settings. You can inject `.p8` file as a string so don't need to store/keep a file somewhere insecure🙂 (The document page hasn't got enough explanations about it.)

```
export APP_STORE_CONNECT_API_KEY_KEY_ID=
export APP_STORE_CONNECT_API_KEY_ISSUER_ID=
# You need to combine multiple lines of .p8 file's content with "\n"
export APP_STORE_CONNECT_API_KEY_KEY= 
```

https://github.com/fastlane/fastlane/blob/e6fc532712772f574636492bd3a9e209344854c8/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb#L49-L68

crazymanish:
> > The main problem is Security team does not allow JSON file under Git commit and most of the CI (i.e CircleCI, GitHub action) just support to store secrets in ENV variables ONLY.
> > Of course, there is always a way, i.e Create the JSON file first using CI ENV variables and run the nuke sub-command but that will defeat the purpose of this action Easily nuke certificate and provisioning profiles (via _match_)
> 
> I do understand this concern. However, it does support environment variables to inject App Store Connect API key settings. You can inject `.p8` file as a string so don't need to store/keep a file somewhere insecure🙂 (The document page hasn't got enough explanations about it.)
> 
> ```
> export APP_STORE_CONNECT_API_KEY_KEY_ID=
> export APP_STORE_CONNECT_API_KEY_ISSUER_ID=
> # You need to combine multiple lines of .p8 file's content with "\n"
> export APP_STORE_CONNECT_API_KEY_KEY= 
> ```
> 
> https://github.com/fastlane/fastlane/blob/e6fc532712772f574636492bd3a9e209344854c8/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb#L49-L68

@ainame 
I know, we can store .p8 file's content with "\n" as String in the ENV variable but passing those ENV variable values to `fastlane match nuke` sub-command is really crazy and this is exactly what I mentioned in the PR description 😉 


Screenshot 2021-03-18 at 19 14 29

crazymanish:
> I think that last thing that is holding me back is making this almost **too easy** for people to delete all of their certificates when they really didn’t mean to. I use `match nuke` pretty often myself but its my own team where things really don’t matter. **Do you have any thoughts if making this too easy would cause any problems?**

@joshdholtz 
Personally this what I believe, people might be using `match nuke` command once in a year, but code-signing is real pain from Day1 and Fastlane is a lifesaver from Day1. 

AppStore API connect KEY is the future and pretty sure people will use the API_KEY more & more.
`match_nuke` action is for that day when suddenly people realize their certificates are expired and what to do next...how to renew using API_KEY, that moment people will proud of Fastlane again because _match_nuke_ will be lifesaver again on that day. 😇

ainame:
@crazymanish 

> I know, we can store .p8 file's content with "\n" as String in the ENV variable but passing those ENV variable values to fastlane match nuke sub-command is really crazy and this is exactly what I mentioned in the PR description 😉

Oh, I got what you mean. This is actually a good point✅ I personally use only custom lanes on my own Fastfiles so didn't notice it.

Regarding how to pass API key via CLI, another option I can suggest is adding all of the options that `app_store_connect_api_key` action has to other ASC API kRaey compatible actions. That way it enables not only `match nuke`  command to be able to pass ASC API key settings via environment variables (with what I explained) but also other main modules of fastlane (deliver, sigh. pilot, ...) and some actions using it. Or it might be good that we can make it possible to inject JSON content with a single option like `api_key_json` (`APP_STORE_CONNECT_API_KEY_JSON`) rather than giving key_id, issuers_id, etc.. individually.

https://github.com/fastlane/fastlane/blob/e6fc532712772f574636492bd3a9e209344854c8/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb#L49-L87

If we make some helper that returns the set of options for ASC API key, it's not something that we have to worry about too much about duplication or misaligned.

@joshdholtz What do you think about ^ option?

---

Either way, `match_nuke` action itself sounds good to me. If the concern remains, we can prompt the "y/n" question by default (and prepare "force" option for CI)?

crazymanish:
> Either way, match_nuke action itself sounds good to me.

Thank you @ainame 🙇‍♂️

> If the concern remains, we can prompt the "y/n" question by default (and prepare "force" option for CI)?

**match_nuke** action is using match options, so we don't need to introduce anything. 😇

```ruby
# for local: It will ask for confimation with "y/n" question before nuke
# see for more details: https://github.com/fastlane/fastlane/blob/master/match/lib/match/nuke.rb#L85
match_nuke(
    type: "development",
    api_key: app_store_connect_api_key,
    read_only: false,
    force: true
)
```

**For CI**, we can pass additional `skip_confirmation` param
```ruby
match_nuke(
    type: "development",
    api_key: app_store_connect_api_key,
    read_only: false,
    force: true,
    skip_confirmation: true
)
```

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
I think... this is good to go! 😊 Thanks for adding this in! Really appreciate the contribution ❤️ 

And thanks for taking the time to review this as well, @ainame 🙌

PR #18021, opened 2021-01-21
[Ruby 3.0] Fix Ruby 2.7's deprecation warnings

ainame:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

As per plan #17931, I've been working on Ruby  3.0 migration in fastlane project. Prior to this PR, I have enabled CI testing for Ruby 2.7 #17861. We can now run unit testing in Ruby 2.7 in order to see if there are deprecation warnings by leveraging the new option "-W:deprecated" in Ruby 2.7.2.  

> Deprecation warnings are off by default on 2.7.2 and later. You can turn on deprecation warnings by specifying the -w or -W:deprecated option at the command-line.
https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/

There is a minor breaking change in Ruby 3.0 which is called "Separation of positional and keyword arguments". In short, the below code won't run on Ruby 3.0. Previously you could pass a Hash object as keyword arguments to a method, but now you need to either use `**` to turn Hash into keywords arguments or simply use keywords to pass parameters. This is the most common case we need to fix. 

```ruby
def say(string:)
  puts string
end
params = { string: "Hello" }

# Bad in Ruby 3.0
say(params)
say({ string: "Hello" })

# Good in Ruby 3.0
say(string: "Hello")
say(**params)
say(**{ string: "Hello" })
```

See details below. 
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

Thanks to the warnings, I could fix them as far as I got. There might be more within this project potentially but it's impossible to review every single line of code. So I expect the community to report something not working on Ruby 3.0 if they find once the whole migration is done in future🙇  

Note that this PR mostly covers the codebase within this project. There are also the same warnings from dependencies. I will work on it in another PR. 

### Description

Other than adding `**` or giving parameters to keyword arguments, I did the followings.

* Update Rakefile and .circleci/config.yml to prevent regressions by monitoring deprecation warnings
* Clean up unused attributes in `FastlaneCore::Project`
* Using keywords arguments in ERB.new
[...]

PR #18395, opened 2021-03-14
[Ruby 3.0] Replace Kernel.open overridden by open-uri with URI.open

ainame:
### Checklist

### Motivation and Context

Prior to this PR, I've fixed most issues about Ruby 2.7 deprecation warnings at least within fastlane/fastlane's codebase #18021 as per plan #17931. There are some left in its dependencies. While working on this PR https://github.com/fastlane/fastlane/pull/18278 that randomises the order of execution of test cases, I found that there are more issues that I missed last time due to the code path in unit testing which is always the same unless using `--order random` like #18278. (In other words, #18278 will be quite handy to reveal undiscovered issues😄)

https://app.circleci.com/pipelines/github/fastlane/fastlane/2446/workflows/604cf165-0e95-4227-8088-732c3957d62c/jobs/64390/parallel-runs/0/steps/0-115

### Description

`Kernel#open` which is overridden by `require 'open-uri'` is deprecated in Ruby 2.7 and removed in Ruby 3.0. For some reasons, it wasn't detected in fastlane's codebase. So this PR replace `Kernel#open` that appears as `open(url)` with `URI.open` which remains available in Ruby 3.0. 
https://github.com/ruby/ruby/commit/a73b5cc556bd131fe924ed6bb02b3c5bdf1593e8

### DIFF
--- a/deliver/lib/deliver/setup.rb
+++ b/deliver/lib/deliver/setup.rb
@@ -1,4 +1,3 @@
-require 'open-uri'
 require 'spaceship/tunes/tunes'
 
 require_relative 'module'
--- a/deliver/lib/deliver/download_screenshots.rb
+++ b/deliver/lib/deliver/download_screenshots.rb
@@ -68,7 +68,7 @@ module Deliver
           end
 
           path = File.join(containing_folder, file_name)
-          File.binwrite(path, open(url).read)
+          File.binwrite(path, URI.open(url).read)
         end
       end
     end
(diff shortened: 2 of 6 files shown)


deliver/lib/deliver/setup.rb
```diff
@@ -1,4 +1,3 @@
-require 'open-uri'
```
ainame, line 1: This is not in use.

rogerluan (approved):
Seems simple enough! LGTM 🎉

twocountriestrus:
Gracias

joshdholtz (approved):
Oh goodness, I had no idea `Kernel.open` was being called anywhere 😱  Good catch on this! ❤️

ainame:
This causes failures in master branch's builds for Ruby 2.4😅 I'm not sure why this PR did pass CI checks 😂

PR #18402, opened 2021-03-15
[action] app_store_connect_api_key - fixed issue with ApiKey in home directory

bimusiek:
### Checklist

* I cannot run the rspec and rubocop on M1 somehow... 

### Motivation and Context
If you want to specify where the api key is stored, you are not able to use home directory. Ruby by default does not expand path of `~/` to `/Users/username/`

CleanShot 2021-03-15 at 19 49 19@2x

### Description
When checking if the file exists, we need to first expand the path.
Not sure if this is correct solution and I don't have an idea how to add a testcase for this scenario (due to zero knowledge about CI runners fastlane uses) so would love to get some feedback.

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb
+++ b/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb
@@ -29,7 +29,7 @@ module Fastlane
         key = {
           key_id: key_id,
           issuer_id: issuer_id,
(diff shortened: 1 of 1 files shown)


bimusiek:
@googlebot I signed it!

joshdholtz (approved):
Ahhh, this is a great! Thank you so much for fixing this ❤️  Really appreciate the contribution

PR #18429, opened 2021-03-19
[fastlane_core] fix 2FA in CI environments by correctly evaluating falsey values of CI identifier env vars

Katsz:
### Checklist

### Motivation and Context
With mandatory 2fa sometimes it's necessary to generate and save `FASTLANE_SESSION` on remote machine. However it can't be done if `CI` variable is defined in current environments, regardless of it's value.

### Description
When defining `CI` var as untruthy (i.e. false or 0) and running `fastlane spaceauth` fastlane will ask for username and password but will crash with `2FA can only be performed in interactive mode` error when required to enter 2fa code.
```bash
export CI=1 # somewhere in .bash_profile
CI=0 fastlane spaceauth # crashes with "2FA can only be performed in interactive mode" error
unset CI && fastlane spaceauth # works fine
```


Katsz:
On a sidenote, empty string should be considered untruthy, but it's not, only `["no", "false", "off", "0"]` are:
https://github.com/fastlane/fastlane/blob/48151291f2c4949c3b1b9919ba2cc81a7cc33293/fastlane_core/lib/fastlane_core/env.rb#L3-L7
But I'm afraid it could  break someone’s flow.

rogerluan (approved):
This LGTM! 💪  Very good catch, and the fix seems to be 🎯 !

joshdholtz (approved):
This looks great to me! Thank for improving this ❤️

PR #18450, opened 2021-03-24
[Fastlane.Swift] Add some missing return types

### LINKED ISSUE #18393: [Fastlane.Swift] Runner.executeCommand(_:) returns empty String for commands with other result types
### New Issue Checklist

### Issue Description

There are a lot of actions, plugins that are not usable in Fastlane Swift as generated functions are Void. For example `gitTagExists(tag:remote:remoteName:)`. Even if we will try to write our own implementations using `RubyCommand` we will not be able to do that for actions that have not `String` return type as `Runner.executeCommand(_:)` uses `String` as return type.
The issue is that `return_object` from `payload` can have any type and during casting we get `nil` for such cases.

Let see how it looks in code:
We have next code that's cast execution result to `String` in `SocketResponse.ResponseType.init(statusDictionary:)`:
```swift
let returnedObject = statusDictionary["return_object"] as? String
```
It means that for types like `Bool` we will get `nil` and `Runner.executeCommand(_:)` will return empty `String` for this case.

The easiest way to have `String`  result type is to replace
```swift
let returnedObject = statusDictionary["return_object"] as? String
```
with
```swift
let returnedObject = statusDictionary["return_object"].map { "\($0)" }
``` 

But I don't think that it is expected behaviour.
I think we should have something like `Runner.executeCommand<V>(_:) -> V` instead, to be able to handle different types.

To achieve this  we can do next:
 - in `SocketResponse. ResponseType` change `returnedObject ` type to `Any`  and do not cast value from payload
```swift
struct SocketResponse {
    enum ResponseType {
        ...
        case readyForNext(returnedObject: Any?, closureArgumentValue: String?) //updated
        ...
        init(statusDictionary: [String: Any]) {
            ...
            if status == "ready_for_next" {
                ...
                let returnedObject = statusDictionary["return_object"] //updated
               ...
                return

            }
            ...
        }
    }
    ...
}
```
 - change `returnedObject ` type to `Any` in`SocketClientResponse` 
 ```swift
 public enum SocketClientResponse: Error {
    ...
    case success(returnedObject: Any?, closureArgumentValue: String?) // updated
}
 ```
  -  and finally update `Runner` to make `Runner.executeCommand(_:)` generic `
  ```swift
 protocol OptionalProtocol {}

extension Optional: OptionalProtocol {}

class Runner {
    ...
    private var returnValue: Any? //updated
    ...

    @discardableResult
    func executeCommand(_ command: RubyCommandable) -> String {
        execute(command).map { "\($0)"} ?? ""
    }

    func executeCommand<Value>(_ command: RubyCommandable) -> Value {
        guard let commandResult = execute(command) else {
            fatalError("Command id:\(command.id) returned 'nil'")
        }
        guard let value: Value = commandResult as? Value else {
            fatalError("Unable to cast \(commandResult) to \(Value.self)")
        }
        return value
    }

    func executeCommand<Value: OptionalProtocol>(_ command: RubyCommandable) -> Value? {
        execute(command).flatMap({ $0 as? Value })
    }

    private func execute(_ command: RubyCommandable) -> Any? {
        dispatchGroup.enter()
        currentlyExecutingCommand = command
        socketClient.send(rubyCommand: command)

        let secondsToWait = DispatchTimeInterval.seconds(SocketClient.defaultCommandTimeoutSeconds)
        // swiftformat:disable:next redundantSelf
        let timeoutResult = Self.waitWithPolling(self.executeNext[command.id], toEventually: { $0 == true }, timeout: SocketClient.defaultCommandTimeoutSeconds)
        executeNext.removeValue(forKey: command.id)
        let failureMessage = "command didn't execute in: \(SocketClient.defaultCommandTimeoutSeconds) seconds"
        let success = testDispatchTimeoutResult(timeoutResult, failureMessage: failureMessage, timeToWait: secondsToWait)
        guard success else {
            log(message: "command timeout")
            preconditionFailure()
        }
        return returnValue
    }
 ...
}
 ```

Then functions like existing `gitTagExists(tag:remote:remoteName:)` can be rewritten next way:
```swift
    func gitTagExists(
        tag: String,
        remote: Bool = false,
        remoteName: String = "origin"
    ) -> Bool {
        let command = RubyCommand(
            commandID: "",
            methodName: "git_tag_exists",
            className: nil,
            args: [
                RubyCommand.Argument(name: "tag", value: tag),
                RubyCommand.Argument(name: "remote", value: remote),
                RubyCommand.Argument(name: "remote_name", value: remoteName)
            ]
        )
        return runner.executeCommand(command)
    }
```

Not sure how code for plugins is generated and how return type is determined, but I believe the next step will be to update those scripts and generate proper return types there to make Fastlane Swift fans happy :).

minuscorp:
### Checklist

### Motivation and Context
Resolves #18393
Have checked most of the actions but maybe I missed some, so a review is appreciated 😄.

### Description
Adds the `action` class method `return_type` for Fastlane.Swift to return a proper result value.
 
### Testing Steps
1. Install from `https://github.com/minuscorp/fastlane/tree/swift-add-missing-return-types`
2. Create your Fastfile.swift with your favorite returning type action.
3. See the proper return type from it.

### DIFF
--- a/fastlane/lib/fastlane/actions/build_app.rb
+++ b/fastlane/lib/fastlane/actions/build_app.rb
@@ -120,6 +120,10 @@ module Fastlane
         "The absolute path to the generated ipa file"
       end
 
+      def self.return_type
+        :string
+      end
+
       def self.author
         "KrauseFx"
       end
--- a/fastlane/lib/fastlane/actions/check_app_store_metadata.rb
+++ b/fastlane/lib/fastlane/actions/check_app_store_metadata.rb
@@ -28,6 +28,10 @@ module Fastlane
         return "true if precheck passes, else, false"
       end
 
+      def self.return_type
+        :bool
+      end
+
       def self.authors
         ["taquitos"]
       end
--- a/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/get_provisioning_profile.rb
@@ -68,6 +68,10 @@ module Fastlane
         "The UUID of the profile sigh just fetched/generated"
       end
 
+      def self.return_type
+        :string
+      end
+
       def self.details
         "**Note**: It is recommended to use [match](https://docs.fastlane.tools/actions/match/) according to the [codesigning.guide](https://codesigning.guide) for generating and maintaining your provisioning profiles. Use _sigh_ directly only if you want full control over what's going on and know more about codesigning."
       end
--- a/fastlane/lib/fastlane/actions/git_tag_exists.rb
+++ b/fastlane/lib/fastlane/actions/git_tag_exists.rb
@@ -45,6 +45,10 @@ module Fastlane
         "Boolean value whether the tag exists or not"
       end
 
+      def self.return_type
+        :bool
+      end
+
       def self.output
         [
         ]
--- a/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
+++ b/fastlane/lib/fastlane/actions/install_provisioning_profile.rb
@@ -48,6 +48,10 @@ module Fastlane
         "The absolute path to the installed provisioning profile"
       end
 
+      def self.return_type
+        :string
+      end
+
       def self.example_code
         [
           'install_provisioning_profile(path: "profiles/profile.mobileprovision")'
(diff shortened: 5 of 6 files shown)


rogerluan (approved):
Nice find! 💪 🎉 

Tested, and generated swift files correctly:

image

joshdholtz (approved):
This is perfect! Great to have you back ❤️ Thanks again for this and all of your contributions!

PR #18452, opened 2021-03-25
[Fastlane.Swift] Fix onError not being called

### LINKED ISSUE #17482: [Fastlane.Swift] `onError` never called.
### New Issue Checklist

### Issue Description

Currently, the method `onError` isn't implemented in `LaneFileProtocol` so that we can't handle errors, for example during `deliver`.
If any lane fails, there is no way to handle the failure.

##### Command executed

Any lane, for example:

`fastlane deliver`

Related: https://github.com/fastlane/fastlane/issues/12381

minuscorp:
### Checklist

### Motivation and Context
Resolves #17482

### Description
This fix an issue that never was initially implemented, the `onError` method call was never implemented, now it does. 

### Testing Steps
1. Checkout and install this fastlane version: `https://github.com/minuscorp/fastlane/tree/swift-on-error-call-fix`
2. Create a failing lane like for example:
```swift
func testLane() {
  sh(command: "exit -1")
}
```
3. Define and override the `onError` method and **don't forget to call the `super`** _(if you don't call the `super`, `afterAll` will be called)_:
image
4. Check that `onError` is being called but `afterAll` doesn't (because an error occurred).

### DIFF
--- a/fastlane/swift/LaneFileProtocol.swift
+++ b/fastlane/swift/LaneFileProtocol.swift
@@ -17,20 +17,20 @@ public protocol LaneFileProtocol: class {
     func recordLaneDescriptions()
     func beforeAll(with lane: String)
     func afterAll(with lane: String)
-    func onError(currentLane: String, errorInfo: String)
+    func onError(currentLane: String, errorInfo: String, errorClass: String?, errorMessage: String?)
 }
 
 public extension LaneFileProtocol {
     var fastlaneVersion: String { return "" } // Defaults to "" because that means any is fine
     func beforeAll(with _: String) {} // No-op by default
     func afterAll(with _: String) {} // No-op by default
-    func onError(currentLane _: String, errorInfo _: String) {} // No-op by default
     func recordLaneDescriptions() {} // No-op by default
 }
 
 @objcMembers
 open class LaneFile: NSObject, LaneFileProtocol {
     private(set) static var fastfileInstance: LaneFile?
+    private static var onErrorCalled: [String: Bool] = [:]
 
     private static func trimLaneFromName(laneName: String) -> String {
         return String(laneName.prefix(laneName.count - 4))
@@ -38,6 +38,10 @@ open class LaneFile: NSObject, LaneFileProtocol {
 
     private static func trimLaneWithOptionsFromName(laneName: String) -> String {
         return String(laneName.prefix(laneName.count - 12))
+    }
+  
+    public func onError(currentLane: String, errorInfo: String, errorClass: String?, errorMessage: String?) {
+      LaneFile.onErrorCalled[currentLane] = true
     }
 
     private static var laneFunctionNames: [String] {
@@ -137,7 +141,9 @@ open class LaneFile: NSObject, LaneFileProtocol {
         _ = fastfileInstance.perform(NSSelectorFromString(laneMethod), with: parameters)
 
         // Call only on success.
-        fastfileInstance.afterAll(with: lane)
+        if !(Self.onErrorCalled[lane] ?? false) {
+          fastfileInstance.afterAll(with: lane)
+        }
 
         log(message: "Done running lane: \(lane) 🚀")
         return true
--- a/fastlane/swift/LaneFileProtocol.swift
+++ b/fastlane/swift/LaneFileProtocol.swift
@@ -30,7 +30,7 @@ public extension LaneFileProtocol {
 @objcMembers
 open class LaneFile: NSObject, LaneFileProtocol {
     private(set) static var fastfileInstance: LaneFile?
-    private static var onErrorCalled: [String: Bool] = [:]
+    private static var onErrorCalled = Set<String>()
 
     private static func trimLaneFromName(laneName: String) -> String {
         return String(laneName.prefix(laneName.count - 4))
@@ -39,9 +39,9 @@ open class LaneFile: NSObject, LaneFileProtocol {
     private static func trimLaneWithOptionsFromName(laneName: String) -> String {
         return String(laneName.prefix(laneName.count - 12))
     }
-  
-    public func onError(currentLane: String, errorInfo: String, errorClass: String?, errorMessage: String?) {
-      LaneFile.onErrorCalled[currentLane] = true
+
+    public func onError(currentLane: String, errorInfo _: String, errorClass _: String?, errorMessage _: String?) {
+        LaneFile.onErrorCalled.insert(currentLane)
     }
 
     private static var laneFunctionNames: [String] {
@@ -141,8 +141,8 @@ open class LaneFile: NSObject, LaneFileProtocol {
         _ = fastfileInstance.perform(NSSelectorFromString(laneMethod), with: parameters)
 
         // Call only on success.
-        if !(Self.onErrorCalled[lane] ?? false) {
-          fastfileInstance.afterAll(with: lane)
+        if !LaneFile.onErrorCalled.contains(lane) {
+            fastfileInstance.afterAll(with: lane)
         }
 
         log(message: "Done running lane: \(lane) 🚀")
--- a/fastlane/swift/SocketClient.swift
+++ b/fastlane/swift/SocketClient.swift
@@ -302,8 +302,9 @@ extension SocketClient: StreamDelegate {
                 self.closeSession(sendAbort: false)
             }
 
-        case let .failure(failureInformation):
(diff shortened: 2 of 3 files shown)


fastlane/swift/SocketResponse.swift
```diff
@@ -35,12 +35,16 @@ struct SocketResponse {
                 return
 
             } else if status == "failure" {
-                guard let failureInformation = statusDictionary["failure_information"] as? [String] else {
+                guard
+                  let failureInformation = statusDictionary["failure_information"] as? [String],
+                  let failureClass = statusDictionary["failure_class"] as? String?,
+                  let failureMessage = statusDictionary["failure_message"] as? String?
```
joshdholtz, line 41: `failureClass` and `failureMessage` are optional in the num. Do these really need to be in the `guard`? 🤔

joshdholtz (changes requested):
One small question!

fastlane/swift/SocketResponse.swift
minuscorp: Is the most straightforward way to assign them into a optional variable to use them afterwards instead of searching them into the dictionary in the following lines and make the cast there, just a way of making the information more readable, but the other approach is as valid as the current one. Just a matter of readability for short. But I can change it if you want 😄

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -17,20 +17,20 @@ public protocol LaneFileProtocol: class {
...
     func beforeAll(with _: String) {} // No-op by default
     func afterAll(with _: String) {} // No-op by default
-    func onError(currentLane _: String, errorInfo _: String) {} // No-op by default
     func recordLaneDescriptions() {} // No-op by default
 }
 
 @objcMembers
 open class LaneFile: NSObject, LaneFileProtocol {
     private(set) static var fastfileInstance: LaneFile?
+    private static var onErrorCalled: [String: Bool] = [:]
```
andemengo, line 33: Just to understand better, why we use a `[String: Bool]`?
I notice that we associate only `true` and never `false`. Maybe a `Set<String>` or an `Array<String>` could be a more indicated structure?

```suggestion
    private static var onErrorCalled: Set<String> = []
```

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -39,6 +39,10 @@ open class LaneFile: NSObject, LaneFileProtocol {
     private static func trimLaneWithOptionsFromName(laneName: String) -> String {
         return String(laneName.prefix(laneName.count - 12))
     }
+  
+    public func onError(currentLane: String, errorInfo: String, errorClass: String?, errorMessage: String?) {
+      LaneFile.onErrorCalled[currentLane] = true
```
andemengo, line 44: If we decided to use a `Set` for example this will become:

```suggestion
      LaneFile.onErrorCalled.insert(currentLane)
```

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -137,7 +141,9 @@ open class LaneFile: NSObject, LaneFileProtocol {
         _ = fastfileInstance.perform(NSSelectorFromString(laneMethod), with: parameters)
 
         // Call only on success.
-        fastfileInstance.afterAll(with: lane)
+        if !(Self.onErrorCalled[lane] ?? false) {
```
andemengo, line 144: If we decided to use a `Set` for example this will become:

```suggestion
        if !(Self.onErrorCalled.contains(lane)) {
```

fastlane/swift/LaneFileProtocol.swift
```diff
@@ -17,20 +17,20 @@ public protocol LaneFileProtocol: class {
...
     func afterAll(with lane: String)
-    func onError(currentLane: String, errorInfo: String)
+    func onError(currentLane: String, errorInfo: String, errorClass: String?, errorMessage: String?)
 }
 
 public extension LaneFileProtocol {
     var fastlaneVersion: String { return "" } // Defaults to "" because that means any is fine
     func beforeAll(with _: String) {} // No-op by default
     func afterAll(with _: String) {} // No-op by default
-    func onError(currentLane _: String, errorInfo _: String) {} // No-op by default
```
andemengo, line 27: Why this default public implementation was deleted?

andemengo (commented):
Thanks for working on the issue that I've opened a few month ago! 👏 
I just want to understand better some motivation behind the data structure used. 🤔

fastlane/swift/LaneFileProtocol.swift
minuscorp: It is now living in the `LaneFile` superclass which it will record the called lanes to avoid `afterAll` being called unnecessarily. Also, for some reason even if the Fastfile overloaded the protocol's implementation, it gets called anyway and ignoring the new implementation (couldn't figure out why exactly).

I'll address the changes you mention once I test that they're working properly, thanks for the feedback! ❤️❤️

fastlane/swift/LaneFileProtocol.swift
minuscorp: The approach for a dictionary is mostly wise in terms of optimization, as reading in a dictionary takes O(1) whereas in an array or set it takes O(N). Not that the performance will be an issue here, but it was just pure instinct.

andemengo: I totally agree with the optimization point; array isn't the best solution, so that I suggested `Set`.
Also `Set` has `O(1)` reading / writing complexity and since in the dictionary we only add `true` values for lanes, it could make sense to semplificate the structure and the future validation operations and checks. ;)

minuscorp: You're completely right! I thought array and set shared same complexity, so your approach is much better than the dictionary for the reasons you mentioned. I'll address this once I get energy to pick the laptop in the weekend😂.

minuscorp:
cc: @andemengo Just finished pushing the review comments 😄

fastlane/swift/SocketClient.swift
```diff
@@ -302,8 +302,9 @@ extension SocketClient: StreamDelegate {
                 self.closeSession(sendAbort: false)
             }
 
-        case let .failure(failureInformation):
-            socketDelegate?.commandExecuted(serverResponse: .serverError) {
+        case let .failure(failureInformation, failureClass, failureMessage):
+          LaneFile.fastfileInstance?.onError(currentLane: ArgumentProcessor(args: CommandLine.arguments).currentLane, errorInfo: failureInformation.joined(), errorClass: failureClass, errorMessage: failureMessage)
+          socketDelegate?.commandExecuted(serverResponse: .serverError) {
```
rogerluan, line 307: Looks like the indentation here is wrong (for Swift)? 🙈 👀

rogerluan (approved):
Good stuff!! So glad to see this implemented after so long 🤩

fastlane/swift/SocketClient.swift
minuscorp: 🤔 maybe, anyway, there's a linter once this is merged so I don't think it is much trouble. In fact I think this is the xcode indentation that is slightly different than Swiftformat's one. Should I change it?

joshdholtz (changes requested):
@minuscorp Thanks so much for adding support for this! I had no idea this wasn’t implemented 😱 I did run into one error when I tried to run the `puts` action from within the `onError`. It ended up hanging after printing “Ho” 😬 Is this a user error on my side? Or is this something that would still need to get fixed 🤔 

```swift

import Foundation

class Fastfile: LaneFile {
        func testLane() {
          desc("Description of what the lane does")
    puts(message: "Hey")
    sh(command: "exit -1")
        }

  override public func onError(currentLane: String, errorInfo: String, errorClass: String?, errorMessage: String?) {
    super.onError(currentLane: currentLane, errorInfo: errorInfo, errorClass: errorClass, errorMessage: errorMessage)

    puts(message: "Ho")
  }
}
```

minuscorp:
At this point of the execution, the socket is already closed due to an error (whichever error it is) so adding more Ruby actions won't work at that point, just Swift code instead, this is how the socket is implemented underneath so I'm not quite sure that I can recover the connection after an error in the execution TBH. 😕

joshdholtz:
@minuscorp Ahhhhhhh, that makes sense! Thanks for explaining that 😊

joshdholtz (approved):
This looks great to me! Thanks for adding this in 🙌

PR #18364, opened 2021-03-09
[action] notarize: add support for AppStore Connect API Key

niezbop:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Following Apple's incentives towards using AppStore Connect API keys instead of passwords, my team is trying to move to this new system. `notarize`, which is a critical component in our Mac builds pipeline does not seem to support it however, which is what this pull request aims to solve.

### Description

The biggest change in the interface of `notarize` this PR suggests is the addition of a new parameter, `api_key_path`, which would be the path to an API key file.

On a purely black box perspective, the only change there is the support for this new authentication method. I made this backwards compatibility in mind, so this change should not change anything for users that do not use API keys.

On a more technical level, I've done a few things:
- make notarize calls authentication abstract of which mean of authentication is used
- select authentication method based on the absence or presence of api_key_path
- support API key parameters
- I took the liberty of making `self.run` ignore rubocop Perceived Complexity metric, because addressing it would have me change more than what I was comfortable with for this specific PR

One important specificity of notarize (which I believe is not present in other apple tools) is that in order to be able to use the key, it must be present on the file system in a specific location, which means that I had to actually write the content of the token on the file system.

PR #18492, opened 2021-04-01
Don't fallback to user password if app-specific password is not available in notarize

mokagio:
### Checklist

### Motivation and Context
While working on automating the build and notarization for the beta version of [Simplenote for macOS](https://github.com/Automattic/simplenote-macos/), which we distribute using [Sparkle](https://sparkle-project.org/), my system ended up in a state such that the `altool` upload would hang. I did a bit of investigation and I thought it was due to the password value not being set by `notarize`. 

That lead me through more investigation and I discovered that ASC now rejects notarization attempts that are not done with an app-specific password.

```
Sign in with the app-specific password you generated. If you forgot the app-specific password or need to create a new one, go to appleid.apple.com
```

### Description

This PR:

- Removes the fallback to the user's account password if no `FL_APP_SPECIFIC_PASSWORD` is set
- Makes the action fail if early if no `FL_APP_SPECIFIC_PASSWORD` is set

### Testing Steps
Unfortunately, there were no tests for `notarize` in which I could hook up into.

And idea I have would be to extract the logic that generates the `altool` commands in a dedicated object that can be tested in isolation. @joshdholtz let me know what you think of this approach. Although, I do wonder if it's out of scope for a PR like this.

`altool` does support using the ASC API key. From `man altool`:

```
xcrun altool --notarize-app  -f  file  --primary-bundle-id  bundle_id  {-u
       username  [-p  password] | --apiKey api_key --apiIssuer issuer_id} [--asc-
       provider provider_shortname | --team-id wwdr_team_id]
```

I would prefer to use that approach in the long run for my app, so I might be able to work on it. That might be a good occasion to make the refactor to enable more testability happen.

I manually tested it as part of my work on Simplenote, although I don't have a PR for it yet.

### DIFF
--- a/fastlane/lib/fastlane/actions/notarize.rb
+++ b/fastlane/lib/fastlane/actions/notarize.rb
@@ -8,6 +8,10 @@ module Fastlane
         print_log = params[:print_log]
         verbose = params[:verbose]
 
+        # Add password as a temporary environment variable for altool.
+        # Use app specific password if specified.
+        ENV['FL_NOTARIZE_PASSWORD'] = ENV['FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD'] || apple_id_account.password
+
         # Compress and read bundle identifier only for .app bundle.
         compressed_package_path = nil
         if File.extname(package_path) == '.app'
@@ -29,10 +33,6 @@ module Fastlane
         UI.user_error!('Could not read bundle identifier, provide as a parameter') unless bundle_id
 
         apple_id_account = CredentialsManager::AccountManager.new(user: params[:username])
-
-        # Add password as a temporary environment variable for altool.
-        # Use app specific password if specified.
-        ENV['FL_NOTARIZE_PASSWORD'] = ENV['FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD'] || apple_id_account.password
 
         UI.message('Uploading package to notarization service, might take a while')
 
--- a/fastlane/lib/fastlane/actions/notarize.rb
+++ b/fastlane/lib/fastlane/actions/notarize.rb
@@ -10,7 +10,17 @@ module Fastlane
 
         # Add password as a temporary environment variable for altool.
         # Use app specific password if specified.
-        ENV['FL_NOTARIZE_PASSWORD'] = ENV['FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD'] || apple_id_account.password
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/notarize.rb
```diff
@@ -1,13 +1,29 @@
 module Fastlane
   module Actions
     class NotarizeAction < Action
+      PASSWORD_ENV_KEY = 'FL_NOTARIZE_PASSWORD'
```
mokagio, line 4: Extracted in constant to DRY

fastlane/lib/fastlane/actions/notarize.rb
```diff
@@ -1,13 +1,29 @@
...
 
+        # Add the app-specific password as a temporary environment variable for
+        # altool.
+        app_specific_password_key = 'FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD'
+        password = ENV[app_specific_password_key]
+
+        message = %{Please specify an app specific password to access App Store Connect for the notarization process.
+          You can do so via the #{app_specific_password_key} environment variable.
+          More information at https://docs.fastlane.tools/best-practices/continuous-integration/#application-specific-passwords.
+        }
```
mokagio, line 21: I want to fix the indentation of this one so it doesn't have extra spaces when printed but I haven't figured out to do it in a Rubocop friendly way yet. I suppose I just gotta look around the codebase.

fastlane/lib/fastlane/actions/notarize.rb
```diff
@@ -1,13 +1,29 @@
...
+
+        UI.user_error!(message) if password.nil? || password.empty?
```
mokagio, line 23: The password empty is a bit of an extra, but I think it's a reasonable check to make.

mokagio:
The changes in https://github.com/fastlane/fastlane/pull/18364 make some of this work redundant.

I'll rebase and keep only the relevant bits once I have a chance...

dede760 (commented):
https://www.facebook.com/deniez.jeloo

Williams241:
I don't know if this is "the right way" but it worked, so here is what I did – maybe There is no place within my personal ADC account or the Team account an additional property to pass to notarize : the "ProviderShortname".

Keith-25:
Sign, notarize, and package macOS CLI tools and applications written in any language. no-frills tool for signing and notarizing your CLI binaries for macOS. gon is available as a CLI This will default to the AC_USERNAME environment variable if not set. password ( string ) - The password for the associated Apple ID.

PR #18504, opened 2021-04-03
[action][jira] API response improvement

crazymanish:
### Checklist

### Motivation and Context
There were two problems in JIRA action.
**Problem1: OnSuccess:** After adding a JIRA comment, it was not returning the JIRA API JSON response.
**Problem2: OnError:** After getting an Error, it was not handling the exception, resulting in the entire Fastlane failed with errors.

### Description
- To solve the problem1, now this action will return the exact JSON returned from the JIRA API
- To solve the problem2, now this action will not fail the entire Fastlane, added begin-rescue block
- Please see attached image for more details. **Earlier v/s Now**

### Testing Steps
Nothing change here.
```ruby
jira(
            url: "https://bugs.yourdomain.com",
            ticket_id: "IOS-123",
            comment_text: "Text to post as a comment"
          )
```

Screenshot 2021-04-03 at 10 25 27

.

Screenshot 2021-04-03 at 10 09 34

### DIFF
--- a/fastlane/lib/fastlane/actions/jira.rb
+++ b/fastlane/lib/fastlane/actions/jira.rb
@@ -43,7 +43,12 @@ module Fastlane
             return json_response
           end
         rescue => exception
-          UI.error("Exception: #{exception}")
+          message = "Received exception when adding a JIRA comment: #{exception}"
+          if params[:fail_on_error]
+            UI.user_error!(message)
+          else
+            UI.error(message)
+          end
         end
       end
 
@@ -92,13 +97,19 @@ module Fastlane
                                        description: "Text to add to the ticket as a comment",
                                        verify_block: proc do |value|
                                          UI.user_error!("No comment specified") if value.to_s.length == 0
-                                       end)
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :fail_on_error,
+                                       env_name: "FL_JIRA_FAIL_ON_ERROR",
+                                       description: "Should an error adding the JIRA comment cause a failure?",
+                                       type: Boolean,
+                                       optional: true,
+                                       default_value: true) # Default value is true for 'Backward compatibility'
         ]
       end
 
       def self.output
         [
-          ['JIRA_JSON', 'The whole JIRA api JSON object']
+          ['JIRA_JSON', 'The whole JIRA API JSON object']
         ]
       end
 
@@ -137,6 +148,11 @@ module Fastlane
             password: "Your password or API token",
             ticket_id: "IOS-123",
             comment_text: "Text to post as a comment"
+          )',
+          'jira(
+            ticket_id: "IOS-123",
+            comment_text: "Text to post as a comment",
+            fail_on_error: false
           )'
         ]
       end
--- a/fastlane/lib/fastlane/actions/jira.rb
+++ b/fastlane/lib/fastlane/actions/jira.rb
@@ -1,5 +1,9 @@
 module Fastlane
   module Actions
+    module SharedValues
+      JIRA_JSON = :JIRA_JSON
+    end
+
     class JiraAction < Action
       def self.run(params)
         Actions.verify_gem!('jira-ruby')
@@ -21,10 +25,27 @@ module Fastlane
                     password: password
                   }
 
-        client = JIRA::Client.new(options)
-        issue = client.Issue.find(ticket_id)
-        comment = issue.comments.build
-        comment.save({ 'body' => comment_text })
+        begin
+          client = JIRA::Client.new(options)
+          issue = client.Issue.find(ticket_id)
+          comment = issue.comments.build
+          comment.save({ 'body' => comment_text })
+
+          # An exact representation of the JSON returned from the JIRA API
+          # https://github.com/sumoheavy/jira-ruby/blob/master/lib/jira/base.rb#L67
+          json_response = comment.attrs
+        rescue => exception
+          UI.error("Exception: #{exception}")
+        ensure
+          if json_response.nil?
+            UI.error('Failed to add a comment on JIRA ticket')
+            return nil
+          else
+            Actions.lane_context[SharedValues::JIRA_JSON] = json_response
+            UI.success('Successfully added a comment on JIRA ticket')
+            return json_response
+          end
+        end
       end
 
       #####################################################
@@ -32,7 +53,7 @@ module Fastlane
       #####################################################
 
       def self.description
-        "Leave a comment on JIRA tickets"
+        "Leave a comment on JIRA ticket"
       end
 
       def self.available_options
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -21,18 +25,34 @@ def self.run(params)
...
+            UI.error('Failed to add a comment on JIRA ticket')
+            return nil
+          else
+            Actions.lane_context[SharedValues::JIRA_JSON] = json_response
+            UI.success('Successfully added a comment on JIRA ticket')
+            return json_response
+          end
+        rescue => exception
+          UI.error("Exception: #{exception}")
+        end
```
joshdholtz, line 47: I don’t think we’d want to wrap this in a begin/rescue here 🤔 This is a breaking change for anybody that is actually expecting this to fail and halt the _fastlane_ process.

You _could_ instead update your `Fastfile` to look like...

```ruby
lane :do_thing do
  begin
    jira
  rescue
    UI.error(“Failed to add a comment on JIRA ticket”)
  end
end
```

I think this _might_ be a better approach as its more flexible and non-breaking 😇 I would like to hear your thoughts on this though!

joshdholtz (changes requested):
I’m thinking the begin/rescue would need to be removed as it might break existing implementations and removes some flexability 🤷‍♂️

fastlane/lib/fastlane/actions/jira.rb
crazymanish: Thank you very much for your time to review the PR and provide me this valuable suggestion. 🙇‍♂️
I agree we should not break the existing flow. Added one new option `fail_on_error` with **default value** true

- [x] Now, for existing users, On error, it will halt the fastlane process. 'Backward compatibility'
- [x] Also, now users can use **FL_JIRA_FAIL_ON_ERROR** in the DotEnv file to control the flow. 😇

Requesting you for re-review. Thank you very much again! 🙇‍♂️

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -21,18 +25,39 @@ def self.run(params)
...
       end
 
       #####################################################
       # @!group Documentation
       #####################################################
 
       def self.description
-        "Leave a comment on JIRA tickets"
+        "Leave a comment on JIRA ticket"
```
joshdholtz, line 60:
```suggestion
        "Leave a comment on a Jira ticket"
```

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -72,15 +97,35 @@ def self.available_options
                                        description: "Text to add to the ticket as a comment",
                                        verify_block: proc do |value|
                                          UI.user_error!("No comment specified") if value.to_s.length == 0
-                                       end)
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :fail_on_error,
+                                       env_name: "FL_JIRA_FAIL_ON_ERROR",
+                                       description: "Should an error adding the JIRA comment cause a failure?",
```
joshdholtz, line 103:
```suggestion
                                       description: "Should an error adding the Jira comment cause a failure?",
```

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -72,15 +97,35 @@ def self.available_options
...
+                                       type: Boolean,
+                                       optional: true,
+                                       default_value: true) # Default value is true for 'Backward compatibility'
+        ]
+      end
+
+      def self.output
+        [
+          ['JIRA_JSON', 'The whole JIRA API JSON object']
```
joshdholtz, line 112:
```suggestion
          ['JIRA_JSON', 'The whole Jira API JSON object']
```

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -72,15 +97,35 @@ def self.available_options
...
         ]
       end
 
       def self.return_value
+        [
+          "A hash containing all relevant information of the JIRA comment",
```
joshdholtz, line 118:
```suggestion
          "A hash containing all relevant information of the Jira comment",
```

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -72,15 +97,35 @@ def self.available_options
...
+          "Access things like JIRA comment 'id', 'author', 'body'"
```
joshdholtz, line 119:
```suggestion
          "Access Jira comment 'id', 'author', 'body', and more"
```

fastlane/lib/fastlane/actions/jira.rb
```diff
@@ -21,18 +25,39 @@ def self.run(params)
...
+          # https://github.com/sumoheavy/jira-ruby/blob/master/lib/jira/base.rb#L67
+          json_response = comment.attrs
+          if json_response.nil?
+            UI.error('Failed to add a comment on JIRA ticket')
+            return nil
+          else
+            Actions.lane_context[SharedValues::JIRA_JSON] = json_response
+            UI.success('Successfully added a comment on JIRA ticket')
+            return json_response
+          end
```
joshdholtz, line 44: I think we could raise an error here if response is nil so that the `fail_on_error` option is respected 😇 Thoughts?

```suggestion
          raise 'Failed to add a comment on JIRA ticket' if json_response.nil?

          Actions.lane_context[SharedValues::JIRA_JSON] = json_response
          UI.success('Successfully added a comment on JIRA ticket')
          return json_response
```

joshdholtz (changes requested):
A few more small suggestions!

crazymanish:
Thank you very much for the review and providing all the suggestions. 🙇‍♂️
Requesting for re-review again! 🙏

joshdholtz:
@crazymanish Thanks for making those changes! I don’t have a Jira project setup to test this but just wanted to verify that you tested my suggestions to make sure they actually worked 😇

crazymanish:
> @crazymanish Thanks for making those changes! I don’t have a Jira project setup to test this but just wanted to verify that you tested my suggestions to make sure they actually worked 😇

Hi, @joshdholtz 
No issue. I am testing different scenarios (success + failure) both right away and will add screenshots here 
Question: Do you have any specific scenarios to test? please let me. you can wait for my screenshots first as well (maybe I will cover all scenarios) 😉

joshdholtz:
@crazymanish Not a ton of specific scenarios in mind! Really just want to make sure it still can post successfully, fail with error, and continue with error 😊

crazymanish:
> @crazymanish Not a ton of specific scenarios in mind! Really just want to make sure it still can post successfully, fail with error, and continue with error 😊

Hello @joshdholtz 
Everything looks good to me, Please see attached screenshots
Screenshot 2021-04-07 at 15 58 11
Screenshot 2021-04-07 at 15 59 15
Screenshot 2021-04-07 at 16 00 18

joshdholtz (approved):
This is great! Thanks for applying my suggestions and doing those tests ❤️ Really appreciate the contribution 🚀

PR #18512, opened 2021-04-06
[Ruby 3.0] improve specs for Slack action 

ainame:
### Checklist

### Motivation and Context

As per the plan #17931 for Ruby 3.0, I've been working on fixing Ruby 2.7's deprecation warnings which will crash on Ruby 3.0. There is an issue where one of the fastlane's dependencies `slack-notifier`, [which is not maintained any more](https://github.com/stevenosloan/slack-notifier/pull/119), keeps producing that warning.

There would be some options in solving that issue - 

1. Ping the maintainers and ask them to release a new version
2. Fork the repo and publish it as a different gem, like `fastlane-slack-notifier`
3. Replace the gem with our own one

In my opinion, 1 is too little hope and 2 would be too much. I would go with 3 as I've got this rewritten one tailored for fastlane's use-case within 60 LoC, which I think is fair to have its own implementation. https://github.com/fastlane/fastlane/pull/18537

So, in order to migrate the Slack action, I'd like to prepare for it by improving the testability of it. Prior to this PR, `slack_spec.rb` uses a hack that makes it return specific objects to verify instead of performing its process entirely. That pulls out `Slack::Notifier` object from the gem to unit tests. Meaning the unit tests will be completely broken when removing the dependency. This PR makes the test cases independent from the gem.

### Description

There are two parts to this PR. The first one is to extract logic in `Fastlane::SlackAction.run` into a runner class that allows me to mock a network request to Slack by `expect(..).to receive(...).with(...)` in `rspec-mocks`.  It is a simpler approach than using the `Helper.test?` hack.  The second one is to use [argument matcher](https://github.com/rspec/rspec-mocks#argument-matchers) to describe what parameters are given to the method handling network requests without having to depend on `slack-notifier` gem.

```diff
-        notifier, attachments = Fastlane::Actions::SlackAction.run(options)
-
-        expect(notifier.config.defaults[:username]).to eq('fastlane')
-        expect(notifier.config.defaults[:channel]).to eq(channel)
-
-        expect(attachments[:color]).to eq('danger')
-        expect(attachments[:text]).to eq(message)
-        expect(attachments[:pretext]).to eq(nil)
-
-        fields = attachments[:fields]
-        expect(fields[1][:title]).to eq('Built by')
-        expect(fields[1][:value]).to eq('Jenkins')
-
-        expect(fields[2][:title]).to eq('Lane')
-        expect(fields[2][:value]).to eq(lane_name)
-
-        expect(fields[3][:title]).to eq('Result')
-        expect(fields[3][:value]).to eq('Error')
+        expected_args = {
+          attachments: [
+            hash_including(
+              color: 'danger',
+              pretext: nil,
+              text: message,
+              fields: array_including(
+                { title: 'Built by', value: 'Jenkins', short: false },
+                { title: 'Lane', value: lane_name, short: true },
+                { title: 'Result', value: 'Error', short: true },
+              )
+            )
+          ],
+          link_names: false,
+          icon_url: 'https://fastlane.tools/assets/img/fastlane_icon.png',
+          fail_on_error: true,
+        }
+        expect(subject).to receive(:post_message).with(expected_args)
+        subject.run(options)
```

Argument Matcher is much more descriptive than using many assertions to verify a big `Hash` object🙁 

### Testing Steps

Add a lane like below to `fastlane/Fastfile` in this repo's clone.

```ruby
lane :slack_test do
  slack(
    slack_url: 'https://hooks.slack.com/services/xxxxxxxxxxx',
    pretext: 'pretext test <@xxxxxx> #general',
    username: 'ainame-bot',
    channel: '#general',
    message: 'test <@xxxxxx> #general',
    link_names: true,
  )
end
```

Screenshot 2021-04-07 at 02 41 16

### DIFF
--- a/fastlane/lib/fastlane/actions/slack.rb
+++ b/fastlane/lib/fastlane/actions/slack.rb
@@ -4,49 +4,40 @@
 module Fastlane
   module Actions
     class SlackAction < Action
-      def self.is_supported?(platform)
-        true
-      end
-
-      # As there is a text limit in the notifications, we are
-      # usually interested in the last part of the message
-      # e.g. for tests
-      def self.trim_message(message)
-        # We want the last 7000 characters, instead of the first 7000, as the error is at the bottom
-        start_index = [message.length - 7000, 0].max
-        message = message[start_index..-1]
-        # We want line breaks to be shown on slack output so we replace
-        # input non-interpreted line break with interpreted line break
-        message.gsub('\n', "\n")
-      end
-
-      def self.run(options)
-        require 'slack-notifier'
-
-        options[:message] = self.trim_message(options[:message].to_s || '')
-        options[:message] = Slack::Notifier::Util::LinkFormatter.format(options[:message])
-
-        options[:pretext] = options[:pretext].gsub('\n', "\n") unless options[:pretext].nil?
-
-        if options[:channel].to_s.length > 0
-          channel = options[:channel]
-          channel = ('#' + options[:channel]) unless ['#', '@'].include?(channel[0]) # send message to channel by default
+      class Runner
+        def initialize(slack_url)
+          @notifier = Slack::Notifier.new(slack_url)
         end
 
-        username = options[:use_webhook_configured_username_and_icon] ? nil : options[:username]
+        def run(options)
+          options[:message] = self.trim_message(options[:message].to_s || '')
+          options[:message] = Slack::Notifier::Util::LinkFormatter.format(options[:message])
 
-        notifier = Slack::Notifier.new(options[:slack_url], channel: channel, username: username)
+          options[:pretext] = options[:pretext].gsub('\n', "\n") unless options[:pretext].nil?
 
-        link_names = options[:link_names]
+          if options[:channel].to_s.length > 0
+            channel = options[:channel]
+            channel = ('#' + options[:channel]) unless ['#', '@'].include?(channel[0]) # send message to channel by default
+          end
 
-        icon_url = options[:use_webhook_configured_username_and_icon] ? nil : options[:icon_url]
+          username = options[:use_webhook_configured_username_and_icon] ? nil : options[:username]
 
-        slack_attachment = generate_slack_attachments(options)
+          slack_attachment = generate_slack_attachments(options)
+          link_names = options[:link_names]
+          icon_url = options[:use_webhook_configured_username_and_icon] ? nil : options[:icon_url]
 
-        return [notifier, slack_attachment] if Helper.test? # tests will verify the slack attachments and other properties
+          post_message(
+            channel: channel,
+            username: username,
+            attachments: [slack_attachment],
+            link_names: link_names,
+            icon_url: icon_url,
+            fail_on_error: options[:fail_on_error]
+          )
+        end
 
-        begin
-          results = notifier.ping('', link_names: link_names, icon_url: icon_url, attachments: [slack_attachment])
+        def post_message(channel:, username:, attachments:, link_names:, icon_url:, fail_on_error:)
+          results = @notifier.ping('', channel: channel, username: username, link_names: link_names, icon_url: icon_url, attachments: attachments)
         rescue => exception
           UI.error("Exception: #{exception}")
         ensure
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -185,99 +292,13 @@ def self.details
...
-      def self.deep_merge(a, b)
-        merger = proc do |key, v1, v2|
-          Hash === v1 && Hash === v2 ?
-                 v1.merge(v2, &merger) : Array === v1 && Array === v2 ?
-                   v1 | v2 : [:undefined, nil, :nil].include?(v2) ? v1 : v2
-        end
-        a.merge(b, &merger)
+      def self.generate_slack_attachments(options)
+        UI.deprecation('`Fastlane::Actions::Slack.generate_slack_attachments` is subject to be removed as Slack recommends migrating `attachments` to Block Kit. fastlane will also the same direction.')
+        Runner.generate_slack_attachments(options)
```
ainame, line 301: I've got feedback to keep this method working here. 
https://github.com/fastlane/fastlane/pull/18537#discussion_r611031719

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -185,99 +292,13 @@ def self.details
...
-      # Adapted from https://stackoverflow.com/a/30225093/158525
-      def self.deep_merge(a, b)
-        merger = proc do |key, v1, v2|
-          Hash === v1 && Hash === v2 ?
-                 v1.merge(v2, &merger) : Array === v1 && Array === v2 ?
-                   v1 | v2 : [:undefined, nil, :nil].include?(v2) ? v1 : v2
-        end
-        a.merge(b, &merger)
+      def self.generate_slack_attachments(options)
+        UI.deprecation('`Fastlane::Actions::Slack.generate_slack_attachments` is subject to be removed as Slack recommends migrating `attachments` to Block Kit. fastlane will also the same direction.')
```
crazymanish, line 300:
```suggestion
        UI.deprecated('`Fastlane::Actions::Slack.generate_slack_attachments` is subject to be removed as Slack recommends migrating `attachments` to Block Kit. fastlane will also the same direction.')
```
`UI.deprecation` does not exist and **halt fastlane**
Screenshot 2021-04-11 at 09 41 41

------

Another thought, 
**For now,** can we just use here `UI.important` for this warning because Slack is not deprecating attachments anytime soon. 
Once we will have Block-Kit implementation then we can make it `UI.deprecated` later.


Screenshot 2021-04-11 at 08 30 31

ainame: > UI.deprecation does not exist and halt fastlane

Nice catch! Thanks🙈 

> For now, can we just use here UI.important for this warning because Slack is not deprecating attachments anytime soon.

Yes, It's not deprecated yet on _Slack_. I meant the use of this method is deprecated in fastlane or plugins as I feel like fastlane has nothing to guarantee this interface or use of `attachments` in JSON for the future. If we postpone deprecating this until we get Block Kit implementation, we need to end up leaving the old implementation even though there's no internal code to use it. That's unacceptable, I think.

rogerluan (approved):
Just a UX comment, other than that this PR LGTM 💪

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -185,99 +292,13 @@ def self.details
...
-      # Adapted from https://stackoverflow.com/a/30225093/158525
-      def self.deep_merge(a, b)
-        merger = proc do |key, v1, v2|
-          Hash === v1 && Hash === v2 ?
-                 v1.merge(v2, &merger) : Array === v1 && Array === v2 ?
-                   v1 | v2 : [:undefined, nil, :nil].include?(v2) ? v1 : v2
-        end
-        a.merge(b, &merger)
+      def self.generate_slack_attachments(options)
+        UI.deprecated('`Fastlane::Actions::Slack.generate_slack_attachments` is subject to be removed as Slack recommends migrating `attachments` to Block Kit. fastlane will also follow the same direction.')
```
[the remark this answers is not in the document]
ainame: What do you think about this?
https://github.com/fastlane/fastlane/pull/18512#discussion_r611319174

I guess `UI.deprecated` is not necessarily meant to provide alternatives and can be used to announce future removal 🤔

joshdholtz (approved):
This is such a high quality code change! I’m learning so much about better code architecture from you 😍

PR #18526, opened 2021-04-09
[supply] upload native symbols for crash symbolication

capyvara:
### Checklist

### Motivation and Context
Fixes #17517, this allows uploading native symbols zip file in order for the Google Play Console to symbolicate crashes.

Google Documentation: https://support.google.com/googleplay/android-developer/answer/9848633#

### Description
It's a simple change, at the upload_mapping it looks for the file extensions and decide to use either the parameter `"proguard"` or `"nativeCode"` when it's a `.zip`.

It only can be uploaded together in the same edit as the `.apk` or `.abb` upload. 

I've tested to upload it alone after and  it says `"The Deobfuscation File can't be modified"`, however you can still do it manually via web later, but beware that any crash types that happened before won't be symbolicated.

### Testing Steps
- Generate your project zipped symbols following the format mentioned on Google's documentation.
- `bundler exec fastlane supply -f path/to/yourapp.aab -d path/to/symbols.zip --package_name "yourpackage.name" --skip_upload_metadata --skip_upload_images --skip_upload_screenshots`
- It should print something like `Preparing mapping at path...` after the app binary upload.
- You can check later at the Google Play Console if the symbols file are there:
image

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -327,12 +327,14 @@ module Supply
     def upload_mapping(path_to_mapping, apk_version_code)
       ensure_active_edit!
 
+      extension = File.extname(path_to_mapping).downcase
+
       call_google_api do
         client.upload_edit_deobfuscationfile(
           current_package_name,
           current_edit.id,
           apk_version_code,
-          "proguard",
+          extension == ".zip" ? "nativeCode" : "proguard",
           upload_source: path_to_mapping,
           content_type: "application/octet-stream"
         )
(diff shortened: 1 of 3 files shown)


aberaud:
Does it allow to upload a proguard and a native debug file in the same upload?

capyvara:
> Does it allow to upload a proguard and a native debug file in the same upload?

Yes, but you need to use --mapping_paths, which accepts an array, instead of just --mapping

joshdholtz (approved):
This looks great! Thanks for adding support for this 😊

PR #18537, opened 2021-04-10
[Ruby 3.0] replace slack-notifier gem with internal library

ainame:
### Checklist

### Motivation and Context

As per the plan #17931 for Ruby 3.0, I've been working on fixing Ruby 2.7's deprecation warnings which will crash on Ruby 3.0. There is an issue where one of the fastlane's dependencies slack-notifier, which is not maintained any more, keeps producing that warning. As I explained in https://github.com/fastlane/fastlane/pull/18512 I decided to go with having our own implementation to post messages to Slack in fastlane. That way we get full control of logic. It is not only about Ruby version support but also Slack's webhook transition.

Right now there are two translations ongoing 

* Legacy incoming webhook https://api.slack.com/legacy/custom-integrations/messaging/webhooks vs The latest incoming webhook using Slack app https://api.slack.com/messaging/webhooks 
* Legacy attachments in formatting https://api.slack.com/reference/messaging/attachments vs BlockKit https://api.slack.com/messaging/composing/layouts#sending-messages

To be fair, `slack-notifier` does support those but we don't know what changes will be coming from Slack in future.

### Description

Prior to this PR, I've updated `fastlane/spec/fastlane/actions_specs/slack_spec.rb` so it's now independent from the gem. This PR added `Fastlane::Notification::Slack` class to own simple alternative of `slack-notifier`. (I suppose it might be interesting to support Microsoft teams or whatever other popular chat tools in fastlane as it's remote working era🙂 That's why it's got `Fastlane::Notification` namespace.) 

After digging through the gem, I realised most of [the extra features](https://github.com/stevenosloan/slack-notifier/tree/master/lib/slack-notifier/payload_middleware) in `slack-notifier` was not used by fastlane. Only `Slack::Notifier::Util::LinkFormatter` was needed to be re-implemented.

### Testing Steps

Get an incoming webook url from here
https://www.slack.com/apps/manage/custom-integrations

Add this snippet to `fastlane/Fastfile` in the repo

```ruby
lane :slack_test do
  slack(
    slack_url: 'https://hooks.slack.com/services/xxxxxxxxxxx',
    pretext: 'pretext test <@xxxxxx> #general',
    username: 'ainame-bot',
    channel: '#general',
    message: 'test <@xxxxxx> #general  <a href="https://fastlane.tools">fastlane</a> [GitHub](https://github.com/fastlane)',
    link_names: true,
  )
end
```

Run `bundle exec fastlane slack_test`

Legacy incoming webhook | Slack App's incoming webhook
--|--
Screenshot 2021-04-10 at 08 20 29 | Screenshot 2021-04-10 at 08 30 08

As you can see Slack app's incoming webhook can't overwrite username and icon. Instead, you need to customise them on its settings page.


fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -1,70 +1,178 @@
...
-      # e.g. for tests
-      def self.trim_message(message)
-        # We want the last 7000 characters, instead of the first 7000, as the error is at the bottom
-        start_index = [message.length - 7000, 0].max
-        message = message[start_index..-1]
-        # We want line breaks to be shown on slack output so we replace
-        # input non-interpreted line break with interpreted line break
-        message.gsub('\n', "\n")
-      end
+        def run(options)
```
crazymanish, line 14: First of all, Good to see, finally we are working on Slack notification improvement.

But I am afraid if this is the right approach or not. This is really breaking change. 
Now we have wrapped everything inside this Runner class and removed all helper methods. 😢

For example:
- **Earlier** SlackAction has helper method `def self.generate_slack_attachments(options)` and anyone can call it directly to generate beautiful slack attachements i.e https://github.com/crazymanish/fastlane-plugin-slack_bot/blob/master/lib/fastlane/plugin/slack_bot/actions/post_to_slack.rb#L23
- **Now** SlackAction removed above and introduced `def generate_slack_attachments(options)` and to call this method we have to init runner first `Runner.new(options[:slack_url]).generate_slack_attachments(options)`

This is a breaking change, why SlackAction need Runner with slack_url to generate slack attachements
We have combined two things into one:
- Notifier just required to post a slack message, that's all.
- Generate slack attachments should not be dependent on the notifier

I am requesting you, please think about it! 🙇‍♂️😇

Otherwise, the User has to do some hack to access the `def generate_slack_attachments(options)` method i.e
```ruby
def self.generate_slack_attachments(options)
   fake_url = "https://this.is.fake.url.to.generate.attachments.com"
   Runner.new(fake_url).generate_slack_attachments(options)
 end
```

Screenshot 2021-04-10 at 12 47 18

crazymanish (changes requested):

fastlane/lib/fastlane/actions/slack.rb
```diff
@@ -1,70 +1,178 @@
...
+          UI.error("Exception: #{error}")
+          message = "Error pushing Slack message, maybe the integration has no permission to post on this channel? Try removing the channel parameter in your Fastfile, this is usually caused by a misspelled or changed group/channel name or an expired SLACK_URL"
+          if fail_on_error
+            UI.user_error!(message)
+          else
+            UI.error(message)
+          end
+        end
 
-        notifier = Slack::Notifier.new(options[:slack_url], channel: channel, username: username)
```
crazymanish, line 38: we could have kept the old implementation (all helper methods) and just removed `require 'slack-notifier'` with this one line to below Fastlane to minimize the breaking changes
`notifier = Fastlane::Notification::Slack.new(options[:slack_url])`

fastlane/lib/fastlane/actions/slack.rb
ainame: Hey! Thank you for your feedback and sorry for making you confused but that breaking change was made in https://github.com/fastlane/fastlane/pull/18512 and this PR is subsequent PR of it. I couldn't open a PR targeting #18512's branch as both sent from my forked repo😅 

> This is really breaking change.
> Now we have wrapped everything inside this Runner class and removed all helper methods. 😢

Regarding the breaking you pointed, I can keep it working for now, in short. The purpose of the extraction is to make it testable without hacky way as well as making existing testing specs independent from `slack-notifier` gem. I didn't intend such disruption for any users🙂 

However, in _my personal_ opinion, fastlane focuses on providing DSL for mobile apps CI/CD rather than being a Ruby library for writing tools. Especially, classes sitting under `Fastlane::Actions` namespace, it's never meant to be libraries directly accessed from other gems, I guess. So I don't recommend you reuse methods implemented in actions. That said, as far as I'm aware there's no such policy explicitly declared and in fact it was not a private method so nothing prevents you from reusing them on your own. 

Another reason that I don't recommend is that as I explained in the PR description, Slack's attachment is now _almost_ deprecated https://api.slack.com/messaging/composing/layouts#when-to-use-attachments. While it will be working for a while it's good to migrate the attachment format to the new Block Kit format. I am thinking of doing the migration near future for fastlane too.

What do you think about keeping the helpers working with deprecation message and remove them when migrating to Block Kit? If you still want exactly the same behaviour in your gem, it's up to you to copy existing logic in fastlane to your gem.

ainame: I'll update #18512 to keep `generate_slack_attachments` or other helpers working.

ainame: @crazymanish Can you please have a look at this and leave any feedback there?🙂 

https://github.com/fastlane/fastlane/pull/18512/files#diff-e17259d099a0447ae1c18920e3251ec219f6069913276bf742f6687d9ea8ff80R299-R301

crazymanish: > What do you think about keeping the helpers working with deprecation message and remove them when migrating to Block Kit?

Thank you very much @ainame I love this approach. 🙇 🙇‍♂️ ❤️ 
Now, instead of introducing breaking change directly, now first we are introducing deprecation message and later breaking changes, which is really cool. This deprecation message will give time to users to perform the migration. 
Thank you very much again! 👏🙌

> However, in my personal opinion, fastlane focuses on providing DSL for mobile apps CI/CD rather than being a Ruby library for writing tools. Especially, classes sitting under Fastlane::Actions namespace, it's never meant to be libraries directly accessed from other gems, I guess. 

Yes, I agree!
[fastlane-plugin-slack_bot](https://github.com/crazymanish/fastlane-plugin-slack_bot) gem is fastlane-plugin (because Fastlane don't accept lots of new actions anymore) so Idea was to reuse Fastlane logic as much as possible and keep things consistent without duplicate code 😊

fastlane/lib/fastlane/notification/slack.rb
```diff
@@ -0,0 +1,56 @@
...
+            channel: channel,
+            username: username,
+            icon_url: icon_url,
+            attachments: attachments,
+            link_names: link_names
+          }.to_json
+        end
+      end
+
+      # This is a substitue of this LinkFormatter in slack-notifier
```
rogerluan, line 27:
```suggestion
      # This class was inspired by `LinkFormatter` in `slack-notifier` gem
```

fastlane/lib/fastlane/notification/slack.rb
```diff
@@ -0,0 +1,56 @@
...
+    class Slack
+      def initialize(webhook_url)
+        @webhook_url = webhook_url
+        @client = Faraday.new do |conn|
+          conn.use(Faraday::Response::RaiseError)
+        end
+      end
+
+      # Overriding channel, icon_url and username is only supported in legacy incoming webhook.
+      # And use of attachments is now deprecated.
```
rogerluan, line 12:
```suggestion
      # Also note that the use of attachments has been deprecated by Slack, in favor of Block Kit.
```

rogerluan (approved):
🚀 Really good stuff!!

fastlane/lib/fastlane/notification/slack.rb
```diff
@@ -0,0 +1,56 @@
+module Fastlane
+  module Notification
```
rogerluan, line 2: I almost forgot to mention that I like this namespace a lot 😊

joshdholtz:
@ainame This is ready for a rebase whenever you got time! I tried to do it on my own but it looked liked a large rebase than I felt comfortable doing as I didn’t want to mess up your PR 😉

fastlane/lib/fastlane/notification/slack.rb
ainame:
```suggestion
      # Also note that the use of attachments has been discouraged by Slack, in favor of Block Kit.
```

I rephrased it a little bit to be accurate😅

ainame:
FYI: We can rebase this branch without having to resolve conflicts like below🙂 

```
$ git rebase --onto origin/master dc2cd566
```

dc2cd566 is the last commit in the previous branch. This way commits made for this PR will be applied to the master branch that has previous PR's changes squashed.

ainame:
@joshdholtz Thanks! This PR is ready to review!

joshdholtz:
> FYI: We can rebase this branch without having to resolve conflicts like below🙂
> 
> ```
> $ git rebase --onto origin/master dc2cd566
> ```
> 
> [dc2cd56](https://github.com/fastlane/fastlane/commit/dc2cd566b144999c058a4145e591fe93de0c99ec) is the last commit in the previous branch. This way commits made for this PR will be applied to the master branch that has previous PR's changes squashed.

OMG... I had no idea about `--onto` 🤯  Thanks!

crazymanish (approved):
This looks great to me, thank you 👏

ainame:
@googlebot I consent.

joshdholtz (approved):
This is beautiful 😍 Thanks again for all your work. This 3.0 dream is getting reeaaaalllll ❤️

PR #18538, opened 2021-04-10
[spaceauth] add `--copy_to_clipboard` option to `spaceauth`

rogerluan:
### Checklist

### Motivation and Context
Resolves https://github.com/fastlane/fastlane/discussions/17843

### Description

After investigating this issue, I tried to accomplish what both users in https://github.com/fastlane/fastlane/discussions/17843 were seeking (`eval $(fastlane spaceauth --eval)`). However, that wasn't possible exactly as is because it would require _fastlane_ to not print anything else to stdout except the export command, and _fastlane_ doesn't have a global `--silent` argument or anything similar - it may print out session-related logs not directly related to spaceauth, or even the initial `[✔] 🚀`.
To work around this issue, the solution I came up with was to export the session to the clipboard without prompting the user for it, specially in a CI environment.

### Documentation

I've documented this new option in `spaceship/README.md` (in this PR) and also here https://github.com/fastlane/docs/pull/1049

### Testing Steps

```ruby
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "origin/rogerluan/spaceauth-export-to-clipboard"
```

Then run `bundle exec fastlane spaceauth -u user@example.org --exports_to_clipboard && eval $(pbpaste)` and `echo $FASTLANE_SESSION` to see if the session was correctly exported.

cc @return-main @Semmu


spaceship/README.md
```diff
@@ -123,6 +123,14 @@ This will authenticate you and provide a string that can be transferred to your
 export FASTLANE_SESSION='---\n- !ruby/object:HTTP::Cookie\n  name: DES5c148586dfd451e55afbaaa5f62418f91\n  value: HSARMTKNSRVTWFla1+yO4gVPowH17VaaaxPFnUdMUegQZxqy1Ie1c2v6bM1vSOzIbuOmrl/FNenlScsd/NbF7/Lw4cpnL15jsyg0TOJwP32tC/NguPiyOaaaU+jrj4tf4uKdIywVaaaFSRVT\n  domain: idmsa.apple.com\n  for_domain: true\n  path: "/"\n  secure: true\n  httponly: true\n  expires: 2016-04-27 23:55:56.000000000 Z\n  max_age: \n  created_at: 2016-03-28 16:55:57.032086000 -07:00\n  accessed_at: 2016-03-28 19:11:17.828141000 -07:00\n'
 ```
 
+Alternatively, pass in the `--exports_to_clipboard` option so that the export code is automatically copied to clipboard, which you can then execute by running `eval $(pbpaste)` right after running `spaceauth`, for instance:
```
return-main, line 126: You may have to explain that `pbcopy`/`pbpaste` is MacOS exclusive.

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -58,9 +60,12 @@ def run
       puts("")
       puts("")
       puts("Example:")
-      puts("export FASTLANE_SESSION='#{yaml}'".cyan.underline)
+      puts(export_command.cyan.underline)
 
-      if mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
+      if @exports_to_clipboard
```
return-main, line 65: I like that you are not checking for the platform here, that way we can set:
```bash
alias pbcopy='xsel --clipboard --input'
alias pbpaste='xsel --clipboard --output'
```
OR
```bash
alias pbcopy='xclip -selection clipboard'
alias pbpaste='xclip -selection clipboard -o'
```
for use on Linux machines.

MarcPerezPro (approved):
Looks good to me, and it's useful for our use-case.

```ruby
  desc 'Updates the FASTLANE_SESSION in AWS Secrets Manager.'
  desc 'We will have to run this lane once a month.'
  desc "It's interactive, so no running on Bitrise."
  desc "You'll need to input the SMS code manually."
  lane :update_fastlane_session do
    fastlane_session_key = 'FASTLANE_SESSION'
    if ENV.has_key?(fastlane_session_key)
      fastlane_session = ENV[fastlane_session_key]
      UI.important "Loaded #{fastlane_session_key} from Envrionment"
    else
      Spaceship::SpaceauthRunner.new(exports_to_clipboard: true).run
      fastlane_session = `pbpaste`.match(/export FASTLANE_SESSION=\'(.*)\'/).captures[0]
    end

    client = Aws::SecretsManager::Client.new()
    resp = client.get_secret_value({
      secret_id: "arn:aws:secretsmanager:████:████:secret:IOSEnvironment"
    })
    secrets = JSON.parse(resp.secret_string)
    secrets[fastlane_session_key] = fastlane_session
    client.put_secret_value({
      secret_id: resp.arn,
      secret_string:  JSON.generate(secrets)
    })
    UI.important "Successfully set the #{fastlane_session_key} in AWS Secrets Manager."
  end
```

spaceship/lib/spaceship/spaceauth_runner.rb
rogerluan: Very interesting point @return-main ! Once I saw the failing specs I thought of either one of the two:
a) Make it work on all OS;
b) Make it work only on macOS.

The first would be a bit tricky. I could add a new dependency such as https://github.com/janlelis/clipboard or implement it on my own, but it'd likely be wonky. 
Making it work only on macOS would be kinda limited. Once I saw this I refactored my code so that it should work on linux if you have an alias set up.

Thanks! 🙌

rogerluan:
@return-main thank you for your review!

I've addressed your points and would be happy if you could review this PR again 🙏 
I have a design question though:

>fastlane_session = `pbpaste`.match(/export FASTLANE_SESSION=\'(.*)\'/).captures[0]

Do you think it would make more sense to have the `export_to_clipboard` option export only the session escaped string (i.e. removing `export FASTLANE_SESSION=` prefix), or perhaps it's good as is? Which one you think it's gonna be more flexible, or attend to users' expectations more often? 😁

spaceship/README.md
rogerluan: Good point! Done! 🎉

return-main:
> @return-main thank you for your review!
> 
> I've addressed your points and would be happy if you could review this PR again 🙏
> I have a design question though:
> 
> > fastlane_session = `pbpaste`.match(/export FASTLANE_SESSION='(.*)'/).captures[0]
> 
> Do you think it would make more sense to have the `export_to_clipboard` option export only the session escaped string (i.e. removing `export FASTLANE_SESSION=` prefix), or perhaps it's good as is? Which one you think it's gonna be more flexible, or attend to users' expectations more often? 😁 also @Semmu maybe you could share how you're using this as well 😇

For our use-case, separating the cookie filtering logic into a function would be perfect. It would allow us to programmatically get the `FASTLANE_SESSION` without any side-effects on the clipboard. If that's too complicated, then just returning the `FASTLANE_SESSION` without prefix at the end should be enough.

For `export_to_clipboard`, I think that keeping the `export FASTLANE_SESSION=` prefix should be easier for the majority of Fastlane users. But at the same time, this behaviour differs from the interactive clipboard copying logic that was previously in place:
```ruby
if mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
```

fastlane_core/lib/fastlane_core/clipboard.rb
```diff
@@ -0,0 +1,20 @@
+require 'fastlane_core'
+require 'open3'
+
+module FastlaneCore
+  class Clipboard
+    def self.copy(content: nil)
+      return UI.crash!("'pbcopy' or 'pbpaste' command not found.") unless is_supported?
+      Open3.popen3('pbcopy') { |input, _, _| input << content }
```
minuscorp, line 8: I wonder if it is worth to make a macOS and Linux version of this, taking advantage of the refactor, just as a suggestion ofc

Semmu:
Hey @rogerluan ,

Thanks for picking up this task! It will be certainly useful for us in my company!

I have read through the code and the discussion here and these are my thoughts:
- Some kind of `--silent` or an equivalent would be the best, where only the `export ...=...` line (or even better, just the session ID) is printed - but as you mentioned, it would required a huge refactoring and would affect most of the codebase, so it is definitely out of scope.
- The clipboard as a workaround seems good, but now as you started writing the core utility for that, it would be the best if fastlane would support all/most platforms, not only macOS. (Though I'm not sure about fastlane usage on other platforms 🤷 )
- Even though with this approach, I would only copy the resulting session ID to the clipboard (omitting the `export` keyword and such), so people can use the session ID whatever way they want without cluttering the environment, for example inline, such as:
  ```bash
  FASTLANE_SESSION=$(pbpaste) fastlane ...
  ```

What do you think?

fastlane_core/lib/fastlane_core.rb
```diff
@@ -3,39 +3,40 @@
...
-require_relative 'fastlane_core/analytics/analytics_event_builder'
-require_relative 'fastlane_core/analytics/analytics_ingester_client'
-require_relative 'fastlane_core/analytics/analytics_session'
 require_relative 'fastlane_core/tag_version'
-require_relative 'fastlane_core/fastlane_pty'
-require_relative 'fastlane_core/queue_worker'
+require_relative 'fastlane_core/test_parser'
+require_relative 'fastlane_core/ui/errors'
+require_relative 'fastlane_core/ui/ui'
+require_relative 'fastlane_core/update_checker/update_checker'
```
return-main, line 39: I like reorganising dependencies alphabetically, but it may have some unintended side-effects in Ruby.
This PR works on my machine, but let's wait for CircleCI's unit tests to complete before merging.

MarcPerezPro (approved):
LGTM

MarcPerezPro (approved):

fastlane_core/lib/fastlane_core.rb
rogerluan: As long as tests pass, is this safe? I have no clue 😮 what are/were the risks? maybe cc @ainame 🙏 😁

rogerluan:
@return-main yeah the inconsistency bothered me when implementing, but I designed it to be as close to what you guys were looking forward to use 😄but I think we can achieve both results, with just a little bit more typing on either end, depending on which way we go.

@Semmu I think the current state for cross-platform support for this feature is decent. We already have a clipboard action that never had built-in support for Linux or Windows, so I think it's fine as is. I wouldn't want to re-implement the `Clipboard` gem, but we might want to decide to add it into fastlane in the future (separate from this PR) if we see necessity - in general we avoid adding more and more dependencies to fastlane unless it's really needed. So there is a significant amount of users facing problems, we can reconsider 😊 

---

Guys, I addressed your latest comments! I changed the clipboard content to contain only the cookie, and updated the docs ([here](https://github.com/fastlane/docs/pull/1049/files)), those were all great and juicy feedbacks! Thanks all!

Lemme know your thoughts on this latest iteration 😊 The only thing I would reconsider is perhaps the name of the option `exports_to_clipboard` - I was thinking `copy_to_clipboard`, thoughts?

fastlane_core/lib/fastlane_core.rb
return-main: CircleCI passed, so it should be safe ✅

As you saw in [fastlane_core.rb](https://github.com/fastlane/fastlane/blob/b2b214f4a6d2450bdefff4dd36d50df79471b51a/fastlane_core/lib/fastlane_core.rb), some requires have to be ordered before:
```ruby
# Ruby monkey-patches - should be before almost all else
require_relative 'fastlane_core/core_ext/string'
require_relative 'fastlane_core/core_ext/shellwords'
```
And some after:
```ruby
require 'commander'

# after commander import
require_relative 'fastlane_core/ui/fastlane_runner' # monkey patch
```

They are usually extensions of Ruby classes, without those we'll probably get `NoMethodError`s in CircleCI.

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -51,25 +51,28 @@ def run
...
-        puts("Successfully copied export command into your clipboard 🎨".green)
-      elsif mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
-        FastlaneCore::Clipboard.copy(content: yaml)
-        puts("Successfully copied text into your clipboard 🎨".green)
+      if @exports_to_clipboard || (mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true))
+        FastlaneCore::Clipboard.copy(content: @yaml)
+        puts("Successfully copied the session string into your clipboard 🎨".green)
       end
+
+      return self
```
return-main, line 70: Thank you for that, it simplified my logic to:
```ruby
fastlane_session = Spaceship::SpaceauthRunner.new().run.instance_variable_get(:@yaml)
```

MarcPerezPro (approved):
LGTM 👍

spaceship/lib/spaceship/spaceauth_runner.rb
return-main: Oh I just noticed the `session_string` method.
So the proper way is:
```ruby
fastlane_session = Spaceship::SpaceauthRunner.new().run.session_string
```

Semmu:
Hey @rogerluan,

The latest changes look really nice! I agree with your point on clipboard support, this solution is sufficient enough.

And also I agree on your naming concerns, maybe `copy_to_clipboard` would make more sense now with the latest changes 😄

fastlane_core/lib/fastlane_core.rb
ainame: As long as test cases cover those files, it would be fine🙂 Even in the monkey patch scenario mentioned, it doesn't matter as long as Ruby doesn't evaluate such patched methods directly. 

For example,  let's say we need to load two files in `fastlane_core.rb` and `file A` is like below, it is safe. The order doesn't matter. `monkey_patch_method_return_true` will be evaluated after everything loaded. 

```ruby
# fastlane_core.rb
require 'file A'
require 'monkey_patch'
```

```ruby
# file A' 
class A 
  def hello
    puts("hello") if "some_string".monkey_patch_method_return_true
  end
end
```

However, `file A'` causes an error because we need to `require` file 'moneky_patch' that implements `monkey_patch_method_return_true` to String before `file A` is loaded. 

```ruby
# file A
if "some_string".monkey_patch_method_return_true
  class A 
    def hello; end
  end
end
```

`commander` scenario surely depends on the gem to define the classes first and then monkey patching follows.

You can test whether that file loading works by `ruby fasltane_core/lib/fastlane_core.rb`. Even so, there is a small chance order updates create logic-level discrepancy but unit tests should cover that 🤞 

---

It's off-topic but `fastlane`'s unit testing runs test cases exactly the same order every time right now so combining existing `require`  order and unit test cases `order` may create a happy code path that still leaves some edge-cases technically😬  

I've been trying to improve that in a PR... 
https://github.com/fastlane/fastlane/pull/18278

joshdholtz: This slightly terrifies me but I do prefer it this way 😬 I don’t _think_ this should causes any issues but still a bit spooky 🤷‍♂️

spaceship/lib/spaceship/commands_generator.rb
```diff
@@ -39,9 +39,10 @@ def run
       command :spaceauth do |c|
         c.syntax = 'fastlane spaceship spaceauth'
         c.description = 'Authentication helper for spaceship/fastlane to work with Apple 2-Step/2FA'
+        c.option('--exports_to_clipboard', 'Whether the export method should be copied to clipboard. When using this option, the session can be exported to an env var by calling `eval $(pbpaste)`')
```
joshdholtz, line 42: I think I saw somebody else suggest this but we can possibly rename this to `copy_to_clipboard` 😇 “Copy to clipboard” tends to be what the standard phrase is for this. Sorry if that is a lot of work 😔

joshdholtz (changes requested):
Small suggestions to rename to `copy_to_clipboard` 😬

spaceship/lib/spaceship/commands_generator.rb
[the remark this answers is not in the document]
rogerluan: 😅 it sure was! Nice catch 💪 I'll update it

fastlane/lib/fastlane/actions/clipboard.rb
```diff
@@ -30,11 +27,11 @@ def self.available_options
       end
 
       def self.authors
-        ["KrauseFx", "joshdholtz"]
+        ["KrauseFx", "joshdholtz", "rogerluan"]
       end
 
       def self.is_supported?(platform)
-        true
+        return FastlaneCore::Clipboard.is_supported?
```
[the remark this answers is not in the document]
rogerluan: Should this have been caught by linter? Or should we configure it to do so? 🤔 
I've changed this specific instance 👍 but what do you think?

spaceship/lib/spaceship/commands_generator.rb
rogerluan: Yup! Np 😊

rogerluan:
Addressed comments and updated https://github.com/fastlane/docs/pull/1049 once again 😊

fastlane/lib/fastlane/actions/clipboard.rb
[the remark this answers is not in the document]
ainame: There is a rule for this but the current Rubocop version fastlane uses don't support it 😢  
https://docs.rubocop.org/rubocop/cops_style.html

(I just stared working on the bumping version lately but there's lot of to do 😂 )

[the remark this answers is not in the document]
rogerluan: ahh got it!
Yeah I think we should really work on bumping its version then 😮 but bumping linter version (if there's no autocorrect) is soooo troublesome 🙈 
Lmk if there's a open PR or branch I can contribute to!

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,22 +51,28 @@ def run
...
       puts("")
       puts("Example:")
-      puts("export FASTLANE_SESSION='#{yaml}'".cyan.underline)
+      puts("export FASTLANE_SESSION='#{@yaml}'".cyan.underline)
 
-      if mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true)
-        require 'open3'
-        Open3.popen3('pbcopy') { |input, _, _| input << yaml }
-        puts("Successfully copied text into your clipboard 🎨".green)
+      if @copy_to_clipboard || (mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true))
```
return-main, line 65: Could you change
```ruby
      if @copy_to_clipboard || (mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true))
        FastlaneCore::Clipboard.copy(content: @yaml)
        puts("Successfully copied the session string into your clipboard 🎨".green)
      end
```
to
```ruby
      if @copy_to_clipboard == false
        puts("Skipped asking to copy the cookie into your clipboard ⏭️".green)
      elsif @copy_to_clipboard || (mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true))
        FastlaneCore::Clipboard.copy(content: @yaml)
        puts("Successfully copied the session string into your clipboard 🎨".green)
      end
```
That way if you explicitly set copy_to_clipboard to false, the lane won't ask you for your input to copy the cookie into your clipboard. It's useful when creating non-interactive scripts that run on interactive MacOS terminals.
Example:
```ruby
fastlane_session = Spaceship::SpaceauthRunner.new(copy_to_clipboard: false).run.session_string
```

rogerluan: Good suggestion! I didn't think anyone would face such scenario. I'll change this 👍

rogerluan: Done! 💪

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -1,15 +1,17 @@
...
 require 'credentials_manager/appfile_config'
 require 'yaml'
+require 'fastlane_core'
 
 require_relative 'tunes/tunes_client'
 
 module Spaceship
   class SpaceauthRunner
-    def initialize(username: nil)
+    def initialize(username: nil, copy_to_clipboard: nil)
```
joshdholtz, line 10:
```suggestion
    def initialize(username: nil, copy_to_clipboard: false)
```

spaceship/README.md
```diff
@@ -111,21 +111,11 @@ When your Apple account has 2 factor verification enabled, you'll automatically
...
-```
+See [Continuous Integration > Authenticating with Apple services > Method 2: Two-step or two-factor authentication > Storing a manually verified session using spaceauth](https://docs.fastlane.tools/best-practices/continuous-integration/#storing-a-manually-verified-session-using-spaceauth)
 
 #### Transporter
 
-See [Continuous Integration > Authentication with Apple services > Application specific passwords](https://docs.fastlane.tools/best-practices/continuous-integration/#application-specific-passwords)
+See [Continuous Integration > Authenticating with Apple services > Method 3: Application-specific passwords](https://docs.fastlane.tools/best-practices/continuous-integration/#method-3-application-specific-passwords)
```
joshdholtz, line 118: Will this URL be dead because of https://github.com/fastlane/docs/pull/1052 ? 🤔

spaceship/lib/spaceship/spaceauth_runner.rb
```diff
@@ -49,22 +51,30 @@ def run
...
-        Open3.popen3('pbcopy') { |input, _, _| input << yaml }
-        puts("Successfully copied text into your clipboard 🎨".green)
+      if @copy_to_clipboard == false
+        puts("Skipped asking to copy the session string into your clipboard ⏭️".green)
+      elsif @copy_to_clipboard || (mac? && Spaceship::Client::UserInterface.interactive? && agree("🙄 Should fastlane copy the cookie into your clipboard, so you can easily paste it? (y/n)", true))
+        FastlaneCore::Clipboard.copy(content: @yaml)
+        puts("Successfully copied the session string into your clipboard 🎨".green)
       end
+
+      return self
```
joshdholtz, line 72: Can this just return `@yaml`? 😇 This is already running on an instance so I don’t think returning `self` is as useful as returning the actual value? 🤔 

This also gets rid of the need for the `session_string` addition below 🤷‍♂️ 

cc: @rogerluan @return-main

joshdholtz (changes requested):
Few more small questions/suggestions 😇 Sorry!

MarcPerezPro (approved):

spaceship/lib/spaceship/spaceauth_runner.rb
return-main: That would change the default behaviour...

spaceship/lib/spaceship/spaceauth_runner.rb
return-main: I'd be okay with that, but by returning `self` we can future-proof the lane in case we create new methods.

spaceship/lib/spaceship/spaceauth_runner.rb
rogerluan: ☝️ Yeah, here the absence of the param behaves differently than an explicit `false` value

spaceship/README.md
rogerluan: Sure will, but it's already dead 😆 
I'll update it here 👍

spaceship/lib/spaceship/spaceauth_runner.rb
rogerluan: I share the same thoughts as @return-main  ☝️ 
Not sure how/if this class plans to be changed much in the future, but it's a possibility. It would make it possible to create new methods similar to how `session_string` works, without worrying about retrocompatibility.

spaceship/README.md
rogerluan:
```suggestion
See [Getting Started > iOS > Authentication > Method 3: Application-specific passwords](https://docs.fastlane.tools/getting-started/ios/authentication/#method-3-application-specific-passwords)
```

spaceship/lib/spaceship/spaceauth_runner.rb
joshdholtz: I’m fine with that 😇  We should be good to go then

joshdholtz (approved):
This is 🔥 Thanks so much for adding this and handling all the reviews ❤️ You my hero!

PR #18544, opened 2021-04-13
[Fastlane.Swift] Swift's `ConfigItem` for conflicting values into Ruby's end.

minuscorp:
### Checklist

### Motivation and Context
This is a fix for many Swift related issues with `nil` and conflicting values being sent into the Ruby side, like: #17343

### Description
This introduces a new type into the _fastlane_ Swift's system called `ConfigItem`, the ability of this `enum` to conform a whole family of `DescriptibleBy*Literal` makes the transition from the current code to the new one just transparent. With a bit of boilerplate that is not going to change into future breaking APIs anytime soon.

#### Example of new implementation:

```swift
public func setupCi(force: Bool = false,
                    provider: ConfigItem<String?> = .fastlaneDefault(nil))
{
    let forceArg = RubyCommand.Argument(name: "force", value: force, type: nil)
    let providerArg = provider.asRubyArgument(name: "provider", type: nil)
    let args = [forceArg,
                providerArg]
        .compactMap { $0 }
    let command = RubyCommand(commandID: "", methodName: "setup_ci", className: nil, args: args)
    _ = runner.executeCommand(command)
}
```

### Testing Steps

1. Install:
```ruby
gem "fastlane", :git => "https://github.com/minuscorp/fastlane.git", :branch => "swift-configitem"
```
2. Introduce two non-`nil` conflicting options -> see that it does break (as it doesn't pass the `conflict_block`).
3. Introduce any other Ruby's valid set of options into the Swift end and see that it works properly.

### DIFF
--- a/fastlane/lib/fastlane/swift_fastlane_function.rb
+++ b/fastlane/lib/fastlane/swift_fastlane_function.rb
@@ -97,6 +97,8 @@ module Fastlane
         return "Bool"
       elsif type_override == Float
         return "Float"
+      elsif type_override == String
+        return "String"
       elsif type_override == :string_callback
         # David Hart:
         # It doesn't make sense to add escaping annotations to optional closures because they aren't function types:
@@ -190,7 +192,13 @@ module Fastlane
         if default_value.nil?
           "#{param}: #{type}"
         else
-          "#{param}: #{type} = #{default_value}"
+          if type == "((String) -> Void)?"
+            "#{param}: #{type} = nil"
+          elsif optional
+            "#{param}: ConfigItem<#{type}> = .current(#{default_value})"
+          else
+            "#{param}: #{type} = #{default_value}"
+          end
         end
       end
 
@@ -271,12 +279,20 @@ module Fastlane
         return "[]" # return empty list for argument
       end
 
-      argument_object_strings = @param_names.zip(param_type_overrides).map do |name, type_override|
+      argument_object_strings = @param_names.zip(param_type_overrides, param_default_values, param_optionality_values).map do |name, type_override, default_value, is_optional|
         sanitized_name = camel_case_lower(string: name)
         sanitized_name = sanitize_reserved_word(word: sanitized_name)
         type_string = type_override == :string_callback ? ", type: .stringClosure" : nil
+        require 'pry-byebug'
+        # binding.pry if function_name == "deliver"
+        method_call = unless type_override == :string_callback || !is_optional 
+                        ".get()"
+                      else
+                        nil
+                      end 
 
-        "RubyCommand.Argument(name: \"#{name}\", value: #{sanitized_name}#{type_string})"
+
+        "RubyCommand.Argument(name: \"#{name}\", value: #{sanitized_name}#{method_call}#{type_string})"
       end
       return argument_object_strings
     end
@@ -414,7 +430,13 @@ module Fastlane
         param = sanitize_reserved_word(word: param)
         static_var_for_parameter_name = param
 
-        "#{param}: #{type} = #{self.class_name.downcase}.#{static_var_for_parameter_name}"
+        if type == "((String) -> Void)?"
+          "#{param}: #{type} = nil"
+        elsif optional
+          "#{param}: ConfigItem<#{type}> = .current(#{self.class_name.downcase}.#{static_var_for_parameter_name})"
+        else
+          "#{param}: #{type} = #{self.class_name.downcase}.#{static_var_for_parameter_name}"
+        end
       end
 
       return param_names_and_types
--- a/fastlane/lib/fastlane/swift_fastlane_function.rb
+++ b/fastlane/lib/fastlane/swift_fastlane_function.rb
@@ -283,9 +283,13 @@ module Fastlane
       argument_object_strings = @param_names.zip(param_type_overrides, param_default_values, param_optionality_values).map do |name, type_override, default_value, is_optional|
         sanitized_name = camel_case_lower(string: name)
         sanitized_name = sanitize_reserved_word(word: sanitized_name)
-        type_string = type_override == :string_callback ? ", type: .stringClosure" : nil
+        type_string = type_override == :string_callback ? ".stringClosure" : "nil"
 
-        "RubyCommand.Argument(name: \"#{name}\", value: #{sanitized_name}#{type_string})"
+        if !(type_override == :string_callback || !is_optional)
+          { name: "#{sanitized_name.gsub('`', '')}Arg", arg: "let #{sanitized_name.gsub('`', '')}Arg = #{sanitized_name}.asRubyArgument(name: \"#{name}\", type: #{type_string})" }
+        else
+          { name: "#{sanitized_name.gsub('`', '')}Arg", arg: "let #{sanitized_name.gsub('`', '')}Arg = RubyCommand.Argument(name: \"#{name}\", value: #{sanitized_name}, type: #{type_string})" }
+        end
       end
       return argument_object_strings
     end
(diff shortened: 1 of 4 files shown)


fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -271,12 +279,19 @@ def build_argument_list
...
         sanitized_name = camel_case_lower(string: name)
         sanitized_name = sanitize_reserved_word(word: sanitized_name)
         type_string = type_override == :string_callback ? ", type: .stringClosure" : nil
 
-        "RubyCommand.Argument(name: \"#{name}\", value: #{sanitized_name}#{type_string})"
+        method_call = unless type_override == :string_callback || !is_optional 
+                        ".get()"
+                      else
+                        nil
+                      end
```
rogerluan, line 291: I think these double negatives are hard to follow 🙈 Perhaps inverting them will make it more easily readable, such as:

```suggestion
        method_call = if type_override != :string_callback || is_optional
                        ".get()"
                      else
                        nil
                      end 
```

note: I'm not rubyist so not sure if my logic is correct (e.g. idk if negating `==` in this case is the equivalent of `!=`)

rogerluan (commented):
Mostly formatting suggestions 😁 but overall this is looking really good!

minuscorp:
Yeah, don't take it so serious at the moment, still a work in progress in terms of code generation and so on, needs tons of code cleanup as you have commented (just had to change lots of autogenerated code by the compiler you know😪) about the naming, it indicates if the value of an item correspond to a `current` value or a user defined one (aka `new`) but any other naming is ok for me", like `default` `userDefined` and `nil` for example. I'm open to suggestions! ❤️

fastlane/lib/fastlane/swift_fastlane_function.rb
minuscorp: Yeah, got rubocop offenses for my double negation logic, but needed for fast testing the possible states. I'll try yours!

fastlane/swift/ConfigItem.swift
```diff
@@ -0,0 +1,130 @@
...
+    case `nil`
+
+    public func get() -> T {
+        switch self {
+        case let .current(value):
+            return value
+        case let .new(value):
+            return value
+        case .nil:
+            fatalError("\(#function) getting called in a nil value")
```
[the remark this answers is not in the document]
minuscorp: I'll apply a filter onto current and nil values to the array of Ruby arguments, so it is really a dead path in theory, that's why I didn't wasted much time on the message but yours is much descriptive

rogerluan:
Ahh got it!
I think `rubyDefaultValue` or `fastlaneDefault` would be easier to understand. And then `userDefined` and `nil` 💯

ainame:
I'm keen to review Fastlane.swift related PRs but I've never really tried it🙂 Would you mind if you guide me on how Fastlane.swift works roughly? 🙏

minuscorp:
You can checkout the docs for a quick start. Roughly this is a Swift runloop that attaches to a socket that it's opened by the Ruby side where the commands (which action to run and which arguments to be run with) are sent by the Swift internal interface so users just use autogenerated code that creates the messages for the socket. Ping me in slack if you need more info about it!❤️

ainame:
Thanks a lot! I thought it is using something like dynamic callable😂 Will learn fastlane.swift gradually 🚶

minuscorp:
It uses a bit of dark magic like getting the lane names from the method naming in the Fastfile class using `NSObject` and objc call table to run the correct method from the argument lane name.

fastlane/swift/ConfigItem.swift
```diff
@@ -0,0 +1,129 @@
...
+//  If you want to add extra functionality to this project, create a new file in a
+//  new group so that it won't be marked for upgrade
+//
+
+import Foundation
+
+public enum ConfigItem<T> {
+    case fastlaneDefault(T)
+    case userDefined(T)
+    case `nil`
```
ainame, line 16: I don't fully understand this yet but could it be possible that we use only two cases; `fastlaneDefault` and `userDefined`?

fastlane/lib/fastlane/swift_fastlane_function.rb
```diff
@@ -190,7 +193,13 @@ def parameters
         if default_value.nil?
           "#{param}: #{type}"
         else
-          "#{param}: #{type} = #{default_value}"
+          if type == "((String) -> Void)?"
+            "#{param}: #{type} = nil"
+          elsif optional
+            "#{param}: ConfigItem<#{type}> = .fastlaneDefault(#{default_value})"
+          else
+            "#{param}: #{type} = #{default_value}"
```
ainame, line 201: Why some params are treated as ConfigItem others are not? Wouldn't it be natural that we always use ConfigItem for every single param for actions?

fastlane/swift/ConfigItem.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    func asRubyArgument(name: String, type: RubyCommand.Argument.ArgType? = nil) -> RubyCommand.Argument? {
+        if case let .userDefined(value) = self {
+            return RubyCommand.Argument(name: name, value: value, type: type)
+        }
```
ainame, line 21: Why `.fastlaneDefault(T)` is ignored here? Is that for document purpose?

ainame (commented):
I had basic questions to help me understand this PR. Feel free to answer/ignore them😄 At least I understand that this can solve the issue that occurss due to default values being `nil` are given to the Ruby side.

fastlane/swift/ConfigItem.swift
minuscorp: `case nil` serves the purpose of filling the case where the user defines as `nil` on-hand but that shouldn't be sent to the Ruby side

fastlane/lib/fastlane/swift_fastlane_function.rb
minuscorp: `ConfigItem` is just designed for optional Ruby values

fastlane/swift/ConfigItem.swift
minuscorp: We don't generate an argument when the case is not `.userDefined`, so `nil` will be returned and cleaned up in the `compactMap { $0 }` call before being sent into the socket.

fastlane/swift/ConfigItem.swift
minuscorp: And as the extensions are being set over concrete non-optional values, we have to set at some point the `ExpressibleByNilLiteral` to be set to that extra `nil` case.

fastlane/swift/ConfigItem.swift
```diff
@@ -0,0 +1,129 @@
...
+//
+//  ** NOTE **
+//  This file is provided by fastlane and WILL be overwritten in future updates
+//  If you want to add extra functionality to this project, create a new file in a
+//  new group so that it won't be marked for upgrade
+//
+
+import Foundation
+
+public enum ConfigItem<T> {
```
joshdholtz, line 13: A small nit pick maybe but thoughts on renaming this to `ConfigValue`? 😇 A config item is something a little more descriptive on the Ruby side where this is specifically just the value 🤷‍♂️

joshdholtz (changes requested):
@minuscorp I tried running this with a command with `adb`. It has an `ConfigItem<String?>` and a `ConfigItem<String>` in it and I got some compile errors when trying to simply use Strings 👇 

Screen Shot 2021-05-04 at 4 10 39 PM

I solved these by adding the `.userDefined` but I don't think that is what the intended solution is, right? 🤷‍♂️ 

Screen Shot 2021-05-04 at 4 12 09 PM

It seems like the issue is with `ConfigItem<String>` as there is not a string literal (???) or something for it to auto convert/cast?

minuscorp:
The issue is that for some reason the parameter is declared as optional but the real parameter later on is not `String?` so that might be causing the issue, I'll check that ASAP

fastlane/swift/ConfigItem.swift
ainame: I agree with renaming. This is a bit confusing to me. Since it handles some sort of optionality on Swift side, `OptionalConfigValue` might make sense slightly more? 😆  
https://github.com/fastlane/fastlane/pull/18544#discussion_r625198720

minuscorp: I ended up renaming as @ainame suggested if that's fine with y'all 😄

minuscorp:
I think I fixed those edge cases with the interpretation of what is an optional value and has a default value 🎉

joshdholtz (approved):
I think this is good to go now! I tested this out a bunch of times and reviewed some of the diff in the generated file 😊 I have no idea how you thought of this and made this because its so meta but 🏆

PR #18555, opened 2021-04-14
[scan] Keep initial xcresult

cnotethegr8:
### Checklist

### Motivation and Context

Scan has a `number_of_retries` parameter which upon retry, will delete an existing `xcresult` file and create a new one. The issue with this is the new file will only contain the previous tests that failed. The reason this is an issue is because the `xcresult` file is used for code coverage.

### Description

This PR is building on top of the `number_of_retries` feature. Instead of removing the `xcresult` file for each retry, it will create a new file with an appending number. The first attempt will not append a number to the `xcresult` file. 

### Testing Steps

This involves enabling `result_bundle` along with setting `number_of_retries` greater than 0. When performing scan make sure one of your tests is failing. You will see inside the `output_directory` path that the first `xcresult` has the whole test suite and is not removed upon retry attempts.

### DIFF
--- a/scan/lib/scan/runner.rb
+++ b/scan/lib/scan/runner.rb
@@ -55,6 +55,8 @@ module Scan
     end
 
     def execute(retries: 0)
+      Scan.cache[:retry_attempt] = Scan.config[:number_of_retries] - retries
+
       command = @test_command_generator.generate
 
       prefix_hash = [
--- a/scan/lib/scan/test_command_generator.rb
+++ b/scan/lib/scan/test_command_generator.rb
@@ -164,8 +164,9 @@ module Scan
 
     # The path to the result bundle
     def result_bundle_path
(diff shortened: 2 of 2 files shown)


joshdholtz (approved):
This is great! I’m so sorry if this broke things for you 😩 This is a very clean way of doing it 💯 Thank you so much for your first contribution!

tengfoung:
@cnotethegr8 Do we have the same mechanism for junit and it's html?

cnotethegr8:
@tengfoung I don't believe so. 
FYI, @joshdholtz

kitskomina:
@joshdholtz 
This issue is also reproducible for junit and html reports format.
We use junit with our GitLab CI and retry functionality breaks tests statistic (it shows only last retried test as successful test and total amount of tests - 1)
Could we have this fixed, please ? Should I create a bug report for this?
Screenshot 2021-05-18 at 17 47 33

PR #17412, opened 2020-10-08
[action] import_from_git - Add caching support

revolter:
### Checklist
- [ ] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [ ] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Even if you use `import_from_git` with a specific branch or tag, the respective repository is cloned every time you run it. This means that a lane that uses this actions is delayed some time, which increases indirectly proportional with the quality of the current Internet connection.

It's even worse when you call it from an Xcode build script! Because it means that every build of your app is delayed.

### Description

If one the `branch` or `version` parameters is specified, along with the newly added `cache_path`, the repository at the specified `url` parameter is cloned into the specified static path, instead of being cloned into a temporary directory that gets deleted after the action finishes running. After this, on subsequent calls, the specified `branch` or `version` is searched in the locally cached repository, and only if it's not found, an update from remote is performed (`git pull`), which it's also faster than a full download (`git clone`).

I tested using manual repositories and `Fastfile`s, but then I moved them entirely in the spec file.

PR #18568, opened 2021-04-17
[action] import_from_git - Add caching support for branch

pixyzehn:
### Checklist

> I've run bundle exec rspec from the root directory to see all new and existing tests pass

I have seen so many errors on this, but I have run `bundle exec rspec ./fastlane/spec/actions_specs/import_from_git_spec.rb` to make sure the tests for the change pass.

### Motivation and Context

First of all, I referred to [[action] import_from_git - Add caching support](https://github.com/fastlane/fastlane/pull/17412) to better understand the context.

With this in mind, the current caching support for `import_from_git` is eligible only if the version is specified, but adding caching support for a branch could be useful when you don't manage versions in a remote git repository, but need to point to the latest commit.

As additional benefits,
- This also prevents us from cloning the repository every time on subsequent calls. For example, on CI, it's likely that you call multiple different lanes as a different step, which could lead to the fact that it clones the repo again and again.
- This makes the usage of `import_from_git` straightforward in that caching support becomes eligible when `cache_path` is specified. Currently, we had to specify `version` to make use of caching support.

### Description

- Update the repo if it's eligible for caching but the version isn't specified.
- Removed `This is ignored if version is not specified.` from the description.
- Added punctuation marks to example code of `import_from_git`.
- Tested this local fastlane code base with my sample project.

### DIFF
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -342,7 +342,7 @@ module Fastlane
             # Update the repo if it's eligible for caching but the version isn't specified
             UI.message("Fetching remote git branches and updating git repo...")
             Helper.with_env_values('GIT_TERMINAL_PROMPT' => '0') do
-              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all -q && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git merge")
+              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all --quiet && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git merge")
             end
           else
             Actions.sh("cd #{clone_folder.shellescape} && git checkout #{checkout_param.shellescape} #{checkout_path}")
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -284,7 +284,7 @@ module Fastlane
 
         action_launched('import_from_git')
 
-        is_eligible_for_caching = !version.nil? && !cache_path.nil?
+        is_eligible_for_caching = !cache_path.nil?
 
         UI.message("Eligible for caching") if is_eligible_for_caching
 
@@ -338,7 +338,15 @@ module Fastlane
             UI.user_error!("No tag found matching #{version.inspect}") if checkout_param.nil?
           end
 
-          Actions.sh("cd #{clone_folder.shellescape} && git checkout #{checkout_param.shellescape} #{checkout_path}")
+          if is_eligible_for_caching && version.nil?
+            # Update the repo if it's eligible for caching but the version isn't specified
+            UI.message("Fetching remote git branches and updating git repo...")
+            Helper.with_env_values('GIT_TERMINAL_PROMPT' => '0') do
+              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all -q && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git merge")
+            end
+          else
+            Actions.sh("cd #{clone_folder.shellescape} && git checkout #{checkout_param.shellescape} #{checkout_path}")
+          end
 
           # Knowing that we check out all the files and directories when the
           # current call is eligible for caching, we don't need to also
--- a/fastlane/lib/fastlane/fast_file.rb
+++ b/fastlane/lib/fastlane/fast_file.rb
@@ -342,7 +342,7 @@ module Fastlane
             # Update the repo if it's eligible for caching but the version isn't specified
             UI.message("Fetching remote git branches and updating git repo...")
             Helper.with_env_values('GIT_TERMINAL_PROMPT' => '0') do
-              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all --quiet && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git merge")
+              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all --quiet && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git rebase")
             end
           else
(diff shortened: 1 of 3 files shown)


pixyzehn:
@googlebot I signed it!

fastlane/lib/fastlane/fast_file.rb
```diff
@@ -338,7 +338,15 @@ def import_from_git(url: nil, branch: 'HEAD', path: 'fastlane/Fastfile', version
             UI.user_error!("No tag found matching #{version.inspect}") if checkout_param.nil?
           end
 
-          Actions.sh("cd #{clone_folder.shellescape} && git checkout #{checkout_param.shellescape} #{checkout_path}")
+          if is_eligible_for_caching && version.nil?
+            # Update the repo if it's eligible for caching but the version isn't specified
+            UI.message("Fetching remote git branches and updating git repo...")
+            Helper.with_env_values('GIT_TERMINAL_PROMPT' => '0') do
+              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all -q && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git merge")
```
revolter, line 345:
```suggestion
              Actions.sh("cd #{clone_folder.shellescape} && git fetch --all --quiet && git checkout #{checkout_param.shellescape} #{checkout_path} && git reset --hard && git merge")
```

I personally prefer to use the long versions of shell arguments, because they're clearer even for people that are new to the respective command.

fastlane/spec/actions_specs/import_from_git_spec.rb
```diff
@@ -208,26 +209,53 @@ def self.is_supported?(platform)
...
-            import_from_git(url: '#{source_directory_path}', branch: '3', cache_path: '#{cache_directory_path}')
+            import_from_git(url: '#{source_directory_path}', branch: 'version-2.1', cache_path: '#{cache_directory_path}')
 
             works
           end").runner.execute(:test)
         end
 
-        it "ignores it when version is not specified" do
-          expect(UI).not_to receive(:message).with(caching_message)
+        # Respects the specified version even if the branch is specified
```
revolter, line 237: Nit:
```suggestion
        # Respects the specified version even if the branch is specified.
```

fastlane/spec/actions_specs/import_from_git_spec.rb
```diff
@@ -208,26 +209,53 @@ def self.is_supported?(platform)
...
+            `git tag "6"`
+          end
+
+          allow(UI).to receive(:message)
+          expect(UI).to receive(:message).with(caching_message)
+          expect(UI).to receive(:important).with('Works until v6')
 
           Fastlane::FastFile.new.parse("lane :test do
-            import_from_git(url: '#{source_directory_path}', cache_path: '#{cache_directory_path}')
+            import_from_git(url: '#{source_directory_path}', branch: 'version-2.1', version: '6', cache_path: '#{cache_directory_path}')
```
revolter, line 256: So the `branch` is essentially ignored in this case?

fastlane/lib/fastlane/fast_file.rb
revolter, line 345: I feel like a `rebase` would make more sense than a `merge`. What do you say?

fastlane/spec/actions_specs/import_from_git_spec.rb
```diff
@@ -208,26 +209,53 @@ def self.is_supported?(platform)
...
 
-        # Because we don't know if the specified value represents a branch or a
-        # tag. And knowing that it doesn't make sense to cache a branch, that
-        # points to a different commit when one is added, as opposed to a tag,
-        # we decided to enable caching only when used with the `version` param.
-        it "ignores it when tag is specified via branch param" do
-          expect(UI).not_to receive(:message).with(caching_message)
-          expect(UI).to receive(:important).with('Works since v1')
+        # Meaning, `git fetch` and `git merge` are performed when caching is eligible
+        # and the version isn't specified in the local clone.
```
revolter, line 213: What do you mean by "in the local clone"?

revolter (changes requested):
I'm trying to understand some things, but it does look good!

fastlane/lib/fastlane/fast_file.rb
pixyzehn: Totally agree with this. Will update it.

fastlane/spec/actions_specs/import_from_git_spec.rb
pixyzehn: I meant the repo that you imported in the cache path, but I think the version should be specified in `import_git_from` action so it can be removed.

fastlane/lib/fastlane/fast_file.rb
pixyzehn: Sounds like a good idea.

To be clear, the only difference between `git merge` and `git rebase` is whether to create a merge commit or not if there are additional commits that a remote branch doesn’t have. Even though it’s not a preferable thing (e.g. adding commits directly from the repo in the cache path), but I think not creating a merge commit would be better in this case.

fastlane/spec/actions_specs/import_from_git_spec.rb
pixyzehn: Yes, that's correct. This is important because
- make sure that it won't break the current cache system with this change.
- specifying `version` should be prioritized over `branch` given that it's an optional parameter and optimistic match operator or multiple conditions can be used.

pixyzehn:
@revolter Thank you for the review! I have updated the PR based on your comments. Since it's my first PR for fastlane, any additional suggestions are appreciated😊

fastlane/lib/fastlane/fast_file.rb
revolter: Exactly 👍🏻

fastlane/spec/actions_specs/import_from_git_spec.rb
revolter: I see, thanks!

revolter (approved):
For a first PR, I can say that it's a great one. Good job, and thanks for it!

PR #18573, opened 2021-04-19
[spaceship] fix case sensitive email issue for Spaceship::Members.find

max-ott:
### Checklist

### Motivation and Context
Apple treats emails case sensitive. Sample: `Spaceship::Members.all` => `"emil.dvorjak@acme.com", "Stacy.Hammilton@acme.com", "samual.Ethan@acme.com"` 

If you search for `Spaceship::Members.find("stacy.hammilton@acme.com")`, the call will fail, because there is no match. This PR tries to fix the issue.

#edgecase

### DIFF
--- a/spaceship/lib/spaceship/tunes/members.rb
+++ b/spaceship/lib/spaceship/tunes/members.rb
@@ -15,7 +15,7 @@ module Spaceship
 
         def find(email)
           all.each do |member|
-            if member.email_address == email
+            if member.email_address.downcase == email
               return member
             end
           end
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/tunes/members.rb
```diff
@@ -15,7 +15,7 @@ def all
 
         def find(email)
           all.each do |member|
-            if member.email_address == email
+            if member.email_address.downcase == email
```
joshdholtz, line 18:
```suggestion
            if member.email_address casecmp?(email)
```

Can we do this instead? ☝️ `email` might not get passed in as all lowercase so we want to make sure its fully caseless compare

joshdholtz (changes requested):

joshdholtz (approved):
Looks good to me! 😊 Thanks for all your contributions ❤️

PR #18585, opened 2021-04-21
[GH Actions] Move 'lock.yml' to '.github/workflows' to make it work

pixyzehn:
### Checklist

### Motivation and Context

I stumbled on the inactive `lock.yml` github action just because it's in the root directory. To make it work, it should be in `.github/workflows`. You can see that it hasn't been active in [the actions tab](https://github.com/fastlane/fastlane/actions).

I assumed this is unexpected based on https://github.com/fastlane/fastlane/pull/16446, but please let me know if we'd like to remove it for some reason.

### Description

Just moved `lock.yml` to `.github/workflows` to make it work again.

### Testing Steps

Tested the action is shown in the actions tab in my private repo with this change.

Screen Shot 2021-04-22 at 8 20 06


joshdholtz (approved):
Huh 😅  This makes a lot of sense now! Thanks for finding and fixing this ❤️

ainame:
This change seems to cause failures with lock workflow on forked repos every midnight as forked repos don't have access to GitHub's access token😅 

Screenshot 2021-04-25 at 03 53 46

pixyzehn:
Perhaps there is a better way to handle this, but as far as I know, you could delete all the workflow for `Locking issues and pull requests` in the forked repository. Unfortunately, you need to do so one by one.

Screen Shot 2021-04-25 at 18 09 50

Once you delete them, you won't see it anymore. Sorry for the inconvenience!

Screen Shot 2021-04-25 at 18 16 05

PR #18600, opened 2021-04-25
[fastlane_core] `xcode_at_least` method improvements

crazymanish:
### Checklist

### Motivation and Context

- I was learning Fastlane-Core's helper class and notice that the `xcode_at_least` method was slightly off/incorrect (in the documentation) 😇

### Description
- Updated the method documentation
- Replaced `FastlaneCore::UI.user_error!("xxxx")` with `UI.user_error!("xxxx")` message
- Improved naming and calling `xcode_version` only once now, earlier it was getting called 2 times (2times calling doesn't really matter in terms of performance, but still why call two times if not required 😊)

### Testing Steps
- Nothing changed in functionality.
- All existing unit tests are passing.

### DIFF
--- a/fastlane_core/lib/fastlane_core/helper.rb
+++ b/fastlane_core/lib/fastlane_core/helper.rb
@@ -167,11 +167,11 @@ module FastlaneCore
       @xcode_version
     end
 
-    # @return true if Xcode version is higher than 8.3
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
OMG that docs 😆 Thank you so much for fixing/cleaning this up ❤️

PR #18607, opened 2021-04-27
[action][upload_symbols_to_crashlytics] Add a check for app_id before searching for gsp and api_token

cristi-lupu:
### Checklist

### Motivation and Context

`upload_symbols_to_crashlytics` needs only one parameter of these: `app_id`, `gsp_path` and `api_token`. If only `app_id` was given no need to find the rest of those.
A issue what I have which requires this patch is that the fastlane searches for `api_token` in every `Info.plist` it can find. But not every `Info.plist` is a `utf-8` file 😃  (the derived data has a lot of `Info.plist` for storyboard files which are binary files). And when it tries to read that non-utf8 `Info.plist` it gives a lot of errors: `invalid byte sequence in UTF-8`. I don't see any value for fastlane to search `api_token` in `Info.plist` files when the `app_id` was already given.

### Description

I've added a check for `app_id` before searching for `gsp_path` and `api_token`. If `app_id` was not given, it will try to find these, otherwise will not.

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
+++ b/fastlane/lib/fastlane/actions/upload_symbols_to_crashlytics.rb
@@ -5,8 +5,10 @@ module Fastlane
         require 'tmpdir'
 
         find_binary_path(params)
-        find_gsp_path(params)
-        find_api_token(params)
+        unless params[:app_id]
+          find_gsp_path(params)
+          find_api_token(params)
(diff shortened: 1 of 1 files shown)


cristi-lupu:
@googlebot I signed it!

joshdholtz (approved):
This checks out! This is a good addition 💪  Thank you so much for the contribution!

PR #18617, opened 2021-04-28
[deliver] remove all instances of `is_string` in options and use `type`

joshdholtz:
### Motivation and Context
`is_string` is slowly being replaced by `type` (and `skip_type_validation` if needed) to make the docs clearer, options safter, and Swift generation more correct.

This PR does this for only `deliver`

### Description
- Most `is_string` that a a `default_value` of a `true`/`false` were replacd with `type: Boolean`
- Support for type `Symbol` was added
  - Will show string in CLI options
  - String values will get converted with `to_sym`
- `skip_type_validation` was added to `team_id` since it can take strings and numbers
- Option `:app` is no longer being used as a cache and replaced with `Deliver.cache` 
  - This option was never set with anything being passed in (only used as a global cache for `Spaceship::ConnectAPI::App`)
  - Added new `Deliver.cache` where app is stored instead


minuscorp:
Does `type`-less ConfigItem get declared as `String` in Swift or into `Any`?

PR #18656, opened 2021-05-07
[Ruby 3.0] Migrate google-api-client gem and drop Ruby 2.4

ainame:
### Checklist

### Motivation and Context

As per plan #17931, I've been working on Ruby 3.0 support. Prior to this PR, I have done the followings -

- Updated CI environment/checks to prevent regressions   
- Removed deprecation warnings
- Bumped Rubocop's version to the latest  
- Replaced unmaintained gems with alternatives

This PR finally unblocks Ruby 3.0 installation by migrating `google-api-client` gem to new ones that support Ruby 3.0. I also enabled CI checks with Ruby 3.0 on CircleCI, which requires us to install it manually as they haven't released a new macOS image with Xcode 12.5 that is supposed to bundle it. 

After this PR gets merged, we are going to be ready to release the new version that officially supports Ruby 3.0 and drop Ruby 2.4 support at the same time.

### Description

https://github.com/googleapis/google-api-ruby-client/blob/master/google-api-client/OVERVIEW.md

`google-api-client` is now split by service. (Just like AWS did for aws-sdk-ruby v3!). We can install only what we use. For fastlane we only use `google-apis-playcustomapp_v1` and `google-apis-androidpublisher_v3`. I put them in gemspec. `google-cloud-storage` is more like a higher abstracted library to interact with Cloud Storage. It depends on `google-apis-iamcredentials_v1` and `google-apis-storage_v1` internally.

As far as I read changelog files, there's no breaking change in those four services `playcustomapp_v1`, `androidpublisher_v3`, `google-apis-iamcredentials_v1` and `google-apis-storage_v1` (since `google-api-client 0.37.0`) and `google-cloud-storage` (sicne `1.15`).

Links
https://github.com/googleapis/google-api-ruby-client/blob/master/google-api-client/CHANGELOG.md
https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-storage/CHANGELOG.md#1310--2021-03-10

That change only broke a test case that stub API requests. The gem changed the hostname of API once apparently.

### Testing Steps

Try supply. I personally have never used supply or Android related actions. I appreciate if people try to use this branch for testing 😂


fastlane.gemspec
```diff
@@ -103,8 +103,9 @@ Gem::Specification.new do |spec|
 
   # The Google API Client gem is *not* API stable between minor versions - hence the specific version locking here.
   # If you upgrade this gem, make sure to upgrade the users of it as well.
```
chris-araman, line 105: Does this comment still apply?

ainame: But there's no reason to keep them as is. Will revise it.

chris-araman: I don't know whether the new gems are intended to use semantic versioning. I only noticed that these are no longer version locked.

ainame: That's right! Thanks.

mikehardy: Obviously these are different ecosystems, but if the versions line up at all, at least for javascript the google-apis follows semantic versioning, but it's essentially nothing but major versions because of it - there is always some API changing incompatibly: https://github.com/googleapis/google-api-nodejs-client/blob/master/CHANGELOG.md

joshdholtz:
@ainame The title of this PR says “drop Ruby 2.5” but its supposed to say “drop Ruby 2.4”, right? 😇

joshdholtz (approved):
This is the one!!! 🎊  I pulled this down and ran some tests on `supply` to make sure we are good and it seems like everything still works as it should 🤷‍♂️ Let's ship this!

PR #18646, opened 2021-05-05
[Ruby 3.0] Replace obsolete URI.escape with alternatives

ainame:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

As per #17931, I've been working on Ruby 3.0 support. Prior to this, I've fixed most deprecation warnings from Ruby 2.7 #18021. It now prevents PRs from making new warnings by checking them in CircleCI.

https://github.com/fastlane/fastlane/blob/72ff4fc82c0cb57c63204eaa4c6d2b9ca33a1bb5/.circleci/config.yml#L106-L111

However, there is still another type of deprecation warning introduced in Ruby 1.9; `warning: URI.escape is obsolete`. This message bypassed the above check😂 I recognised this warning before but didn't know `URI.escape` was finally removed in Ruby 3.0 until I run fastlane on Ruby 3.0, which became possible after #18599.

This PR replace `URI.escape` with some alternatives. Some of them were fixed in this PR previously https://github.com/fastlane/fastlane/pull/16409 but it had some defects that the PR changed results. This PR also restores previous results.

You can read some history in blog posts; for example 
https://docs.knapsackpro.com/2020/uri-escape-is-obsolete-percent-encoding-your-query-string

### Description

There's no 100% equivalent to `URI.escape`. So we need to use alternatives case by case.

1. `URI.encode_www_form_component` - can be used where you encode a part of URI query parameter 
2. `URI.encode_www_form` - can be used with query parameters in Array on Hash
3. `CGI.escape` - similar to `URI.encode_www_form_component` but it follows CGI's spec
4. `Addressable::URI.encode` - is 3rd party dependency (we've already adopted) but is the closest option to `URI.escape`

I applied 1 for places where we process query parameters and 3 for the rest of URLs and file paths. 

Note that there is different behaviour between `URI.escape` and `URI.encode_www_form_component` especially for whitespaces. While `URI.escape` and `Addressable::URI.escape` can work well for file paths including whitespaces,  `URI.encode_www_form_component` is only effective for "query parameters" really.

These are examples of differences among those APIs (except for `CGI.escape`). 
<details> 

```
[...]

PR #18689, opened 2021-05-14
[Ruby 3.0] Migrate google-api-client gem and drop Ruby 2.4 (#18656)

joshdholtz:
I merged these in the wrong order 😱 

This contains #18656 and #18646


ainame:
@googlebot I consent.

ainame (approved):
The order was correct but I just needed to rebase the latter branch after merging the former one 😄

PR #18696, opened 2021-05-16
[fastlane_core][configuration] suggestion message improvement when fetching value with non-symbol key

crazymanish:
### Checklist

### Motivation and Context
- When fetching value with non-symbol key Fastlane provides us the suggestion to use symbol key, which was slightly outdated and was using `:app_id` in the example always.

### Description
- Fix the suggestion message, now Fastlane will provide us symbol key suggestion always instead of hardcoded `:app_id`

### Testing Steps
- Updated unit tests
- Nothing changed in functionality


joshdholtz (approved):
I can’t believe it was every doing this 😛 Thank you for improving this! Really appreciate it ❤️

Issue #18698, opened 2021-05-17
upload_to_play_store fails with error wrong number of arguments (given 1, expected 0)


edouardruiz:
### New Regression Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Regression Information

- Breaking version: 2.183.0
- Last working version: 2.182.0

### Regression Description

The below command (using `upload_to_play_store`) fails with error message `error wrong number of arguments (given 1, expected 0)` since upgrading fastlane from 2.182.0 to 2.183.0 and ruby from 2.7.2 to 3.0.1.

```
bundle exec fastlane run upload_to_play_store version_name:2021.5.17 metadata_path:fastlane/metadata/android skip_upload_images:true skip_upload_screenshots:true
```

If I rollback to the previous version (fastlane 2.182.0 / ruby 2.7.2) the command succeeds.

I have a `json_key.json` file with Google API credentials in the current folder validated by the command `validate_play_store_json_key`.

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
### Captured Output

Command Used: `run upload_to_play_store --capture_output version_name:2021.5.17 metadata_path:fastlane/metadata/android skip_upload_images:true skip_upload_screenshots:true`
<details><summary>Output/Log</summary>

```

[✔] 🚀 
+------------------------------+---------+------------------------------+
|                             Used plugins                              |
+------------------------------+---------+------------------------------+
| Plugin                       | Version | Action                       |
+------------------------------+---------+------------------------------+
| fastlane-plugin-firebase_ap  | 0.2.9   | firebase_app_distribution    |
| p_distribution               |         | firebase_app_distribution_g  |
|                              |         | et_udids                     |
|                              |         | firebase_app_distribution_l  |
|                              |         | ogin                         |
| fastlane-plugin-badge        | 1.5.0   | add_badge                    |
+------------------------------+---------+------------------------------+

[14:21:20]: ----------------------------------
[14:21:20]: --- Step: upload_to_play_store ---
[14:21:20]: ----------------------------------

+---------------------------------+---------------------------+
|                 Summary for supply 2.183.0                  |
+---------------------------------+---------------------------+
[...]

PR #18703, opened 2021-05-17
[supply][Ruby 3] fix Ruby 3 keyword args issues

ainame:
### Checklist

### Motivation and Context

https://github.com/fastlane/fastlane/issues/18698

### Description

Due to the lack of unit tests, we missed some Ruby 3 keyword args issues. I fixed another issue in `supply/lib/supply/uploader.rb`. 
I checked with `git grep "Google::Apis"` and `git grep "AndroidPublisher::"` but the rest looks okay.

### DIFF
--- a/supply/lib/supply/client.rb
+++ b/supply/lib/supply/client.rb
@@ -279,13 +279,13 @@ module Supply
     def update_listing_for_language(language: nil, title: nil, short_description: nil, full_description: nil, video: nil)
       ensure_active_edit!
 
-      listing = AndroidPublisher::Listing.new(**{
+      listing = AndroidPublisher::Listing.new(
         language: language,
         title: title,
         full_description: full_description,
         short_description: short_description,
         video: video
-      })
+      )
 
       call_google_api do
         client.update_edit_listing(
--- a/supply/lib/supply/uploader.rb
+++ b/supply/lib/supply/uploader.rb
@@ -233,10 +233,10 @@ module Supply
         end
       end
 
-      AndroidPublisher::LocalizedText.new({
+      AndroidPublisher::LocalizedText.new(
         language: language,
         text: changelog_text
-      })
+      )
     end
 
     def upload_changelogs(release_notes, release, track)


supply/lib/supply/client.rb
```diff
@@ -279,13 +279,13 @@ def latest_version(track)
     def update_listing_for_language(language: nil, title: nil, short_description: nil, full_description: nil, video: nil)
       ensure_active_edit!
 
-      listing = AndroidPublisher::Listing.new(**{
```
ainame, line 282: Since other places already use keyword-ish style, I made this consistent with others.  `**` would make sense where `Hash` object is given from somewhere else whereas this is just a hash literal that we can just omit `{}`.

```
[1] pry(main)> def hello(**args)
[1] pry(main)*   p args
[1] pry(main)* end
=> :hello
[2] pry(main)> hello(a: 1, b: 2)
{:a=>1, :b=>2}
=> {:a=>1, :b=>2}
[3] pry(main)> hello(**{a: 1, b: 2})
{:a=>1, :b=>2}
=> {:a=>1, :b=>2}
[4] pry(main)> hello({a: 1, b: 2})
ArgumentError: wrong number of arguments (given 1, expected 0)
from (pry):1:in `hello'
```

Both result in the same

supply/lib/supply/uploader.rb
```diff
@@ -233,10 +233,10 @@ def upload_changelog(language, version_code)
         end
       end
 
-      AndroidPublisher::LocalizedText.new({
+      AndroidPublisher::LocalizedText.new(
```
ainame, line 236: https://github.com/fastlane/fastlane/issues/18698#issuecomment-842625709

joshdholtz (approved):
This looks good to me!

supply/lib/supply/client.rb
ainame: https://github.com/googleapis/google-api-ruby-client/blob/master/google-api-client/generated/google/apis/androidpublisher_v3/classes.rb#L1917-L1919

PR #18732, opened 2021-05-21
[deliver][spaceship] migrate AgeRatingDeclaration from AppStoreVersion to AppInfo for App Store Connect API 1.3 update

joshdholtz:
### Motivation and Context
App Store Connect API released version 1.3 today
- `AgeRatingDelcaration` deprecated on `AppStoreVersion`
- `AgeRatingDeclaration` added as relationship on `AppInfo`
- `gamblingAndContests` split into
  - `gambling` (boolean value)
  - `contests` (string value)
- Added new `seventeenPlus` rating (boolean value)

### Description
- Moved API and helper methods from `AppStoreVersion` to `AppInfo`
- Added a new deprecation method on `AgeRatingDelcaration` to map values to new format if possible
  - Returns the changed Hash
  - Returns deprecation messages to be printed out
  - Returns error messages to be raised

### DIFF
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -342,7 +342,7 @@ module Deliver
 
       set_review_information(version, options)
       set_review_attachment_file(version, options)
-      set_app_rating(version, options)
+      set_app_rating(app_info, options)
     end
 
     # rubocop:enable Metrics/PerceivedComplexity
@@ -642,7 +642,7 @@ module Deliver
       end
     end
 
-    def set_app_rating(version, options)
+    def set_app_rating(app_info, options)
       return unless options[:app_rating_config_path]
 
       require 'json'
@@ -675,9 +675,23 @@ module Deliver
         has_mapped_values = true
         UI.deprecated("Age rating '#{k}' from iTunesConnect has been deprecated. Please replace with '#{v}'")
       end
-      UI.deprecated("You can find more info at https://docs.fastlane.tools/actions/deliver/#reference") if has_mapped_values
 
-      age_rating_declaration = version.fetch_age_rating_declaration
+      # Handle App Store Connect deprecation/migrations of keys/values if possible
+      attributes, deprecation_messages, errors = Spaceship::ConnectAPI::AgeRatingDeclaration.map_deprecations_if_possible!(attributes)
+      deprecation_messages.each do |message|
+        UI.deprecated(message)
+      end
+
+      unless errors.empty?
+        errors.each do |error|
+          UI.error(error)
+          UI.user_error!("There are Age Rating deprecation errors that cannot be solved automatically... Please apply any fixes and try again")
+        end
+      end
+
+      UI.deprecated("You can find more info at https://docs.fastlane.tools/actions/deliver/#reference") if has_mapped_values || !deprecation_messages.empty?
+
+      age_rating_declaration = app_info.fetch_age_rating_declaration
       age_rating_declaration.update(attributes: attributes)
     end
   end
--- a/deliver/lib/deliver/upload_metadata.rb
+++ b/deliver/lib/deliver/upload_metadata.rb
@@ -685,8 +685,8 @@ module Deliver
       unless errors.empty?
         errors.each do |error|
           UI.error(error)
-          UI.user_error!("There are Age Rating deprecation errors that cannot be solved automatically... Please apply any fixes and try again")
         end
+        UI.user_error!("There are Age Rating deprecation errors that cannot be solved automatically... Please apply any fixes and try again")
       end
 
       UI.deprecated("You can find more info at https://docs.fastlane.tools/actions/deliver/#reference") if has_mapped_values || !deprecation_messages.empty?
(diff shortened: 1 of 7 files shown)


spaceship/lib/spaceship/connect_api/models/age_rating_declaration.rb
```diff
@@ -83,6 +93,32 @@ def self.type
...
+            hash['contests'] = value # setting to unchanged value
+          else
+            hash['contests'] = 'NONE'
+          end
+
+          messages << "Age Rating 'gamblingAndContests' has been deprecated and split into 'gambling' and 'contests'"
+        end
+
+        return hash, messages, errors
+      end
```
ainame, line 120:
```suggestion
      def map_deprecation_if_possible(attributes)
        attributes = attributes.dup
        messages = []
        errors = []

        value = attributes.delete('gamblingAndContests')
        return attributes, messages, errors if value.nil?

        messages << "Age Rating 'gamblingAndContests' has been deprecated and split into 'gambling' and 'contests'"

        if value == true
          errors << "'gamblingAndContests' could not be mapped to 'contests' - 'contests' requires a value of 'NONE', 'INFREQUENT_OR_MILD', or 'FREQUENT_OR_INTENSE'"
          attributes['gambling'] = value
          attributes['contests'] = value
        else
          attributes['gambling'] = value
          attributes['contests'] = 'NONE'
        end

        return attributes, messages, errors
      end
```

This is an example, but I think if you copy a given hash (with `Object#dup`), you don't need to name the method with `!`. `!` for a mapping method sounds a bit strange to me. And you can reduce nesting by using the guard idiom?

```
irb(main):001:0> a = { a: 1, b: 2, c: 3}
=> {:a=>1, :b=>2, :c=>3}
irb(main):002:0> b = a.dup
=> {:a=>1, :b=>2, :c=>3}
irb(main):003:0> b.delete(:a)
=> 1
irb(main):004:0> a
=> {:a=>1, :b=>2, :c=>3}
```

deliver/lib/deliver/upload_metadata.rb
```diff
@@ -675,9 +675,23 @@ def set_app_rating(version, options)
...
+      deprecation_messages.each do |message|
+        UI.deprecated(message)
+      end
+
+      unless errors.empty?
+        errors.each do |error|
+          UI.error(error)
+          UI.user_error!("There are Age Rating deprecation errors that cannot be solved automatically... Please apply any fixes and try again")
+        end
+      end
```
ainame, line 690:
```suggestion
      unless errors.empty?
        errors.each do |error|
          UI.error(error)
        end
        UI.user_error!("There are Age Rating deprecation errors that cannot be solved automatically... Please apply any fixes and try again")
      end
```

I suppose you wanted to let it crash after iterating through all of the possible error messages?

spaceship/lib/spaceship/connect_api/tunes/tunes.rb
```diff
@@ -17,9 +17,13 @@ def tunes_request_client
...
-        def get_age_rating_declaration(app_store_version_id: nil)
-          params = tunes_request_client.build_params(filter: nil, includes: nil, limit: nil, sort: nil)
-          tunes_request_client.get("appStoreVersions/#{app_store_version_id}/ageRatingDeclaration", params)
+        def get_age_rating_declaration(app_info_id: nil, app_store_version_id: nil)
+          raise "Keyword 'app_store_version_id' is deprecated and 'app_info_id' is required" if app_store_version_id
+
+          if app_info_id
+            params = tunes_request_client.build_params(filter: nil, includes: nil, limit: nil, sort: nil)
+            tunes_request_client.get("appInfos/#{app_info_id}/ageRatingDeclaration", params)
+          end
```
ainame, line 26:
```suggestion
          raise "Keyword 'app_store_version_id' is deprecated and 'app_info_id' is required" if app_store_version_id || app_info_if.nil?

          params = tunes_request_client.build_params(filter: nil, includes: nil, limit: nil, sort: nil)
          tunes_request_client.get("appInfos/#{app_info_id}/ageRatingDeclaration", params)
```

This handles the case where both `app_store_version_id` and `app_info_id` are `nil` slightly better. The current will do nothing in that situation whereas my suggestion gives you the reason and then causes a crash.

ainame (changes requested):
It's so quick! Amazing to see you've done this update from day 1😍 
Changes make sense to me. What I suggested is mostly improving minor details.

deliver/lib/deliver/upload_metadata.rb
joshdholtz: @ainame We don't have access to `Fastlane::UI` inside of `spaceship` otherwise I would have done this 😔 I was either going to:
1) raise an error from the method which is why I had a `!`
2) pass in a block that would return the errors so the caller could call `UI`
3) Or just return the errors like I did here

I went with return the errors and the deprecation messages to be displayed in the calling method that has access to `UI`

joshdholtz: Oh whoops, this was totally my intention 😛

ainame (approved):
🚀

Saravia83c:
@

PR #18736, opened 2021-05-22
[fastlane_core] ConfigItem - `auto_convert_value` improvements for Boolean type

crazymanish:
### Checklist

### Motivation and Context
- Fastlane_Core `ENV.truthy?` check for `["no", "false", "off", "0"]` [see here](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/env.rb#L5) but
- Fastlane_Core ConfigItem auto-convert value was using only `NO no false FALSE`

### Description
- Added a couple of more options to ConfigItem for Boolean type auto-convert i.e `off, OFF` for false and `on, ON` for true. 

### Testing Steps
- No existing functionality changed, also added unit tests,

### DIFF
--- a/fastlane/lib/fastlane/actions/appledoc.rb
+++ b/fastlane/lib/fastlane/actions/appledoc.rb
@@ -180,8 +180,8 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :warnings, env_name: "FL_APPLEDOC_WARNINGS", description: "Documentation generation warnings", is_string: true, optional: true),
 
           # MISCELLANEOUS
-          FastlaneCore::ConfigItem.new(key: :logformat, env_name: "FL_APPLEDOC_LOGFORMAT", description: "Log format [0-3]", is_string: false, optional: true),
-          FastlaneCore::ConfigItem.new(key: :verbose, env_name: "FL_APPLEDOC_VERBOSE", description: "Log verbosity level [0-6,xcode]", is_string: false, optional: true)
+          FastlaneCore::ConfigItem.new(key: :logformat, env_name: "FL_APPLEDOC_LOGFORMAT", description: "Log format [0-3]", type: Integer, optional: true),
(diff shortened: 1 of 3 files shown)


fastlane/lib/fastlane/actions/appledoc.rb
```diff
@@ -180,8 +180,8 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :warnings, env_name: "FL_APPLEDOC_WARNINGS", description: "Documentation generation warnings", is_string: true, optional: true),
 
           # MISCELLANEOUS
-          FastlaneCore::ConfigItem.new(key: :logformat, env_name: "FL_APPLEDOC_LOGFORMAT", description: "Log format [0-3]", is_string: false, optional: true),
-          FastlaneCore::ConfigItem.new(key: :verbose, env_name: "FL_APPLEDOC_VERBOSE", description: "Log verbosity level [0-6,xcode]", is_string: false, optional: true)
+          FastlaneCore::ConfigItem.new(key: :logformat, env_name: "FL_APPLEDOC_LOGFORMAT", description: "Log format [0-3]", type: Integer, optional: true),
```
minuscorp, line 183: If this change is needed because it breaks compatibility with the previous parsing, plugins will be affected as well and then will break them.

minuscorp (commented):
A question about the change🙏

fastlane/lib/fastlane/actions/appledoc.rb
crazymanish: Thank you for your review and reverted these changes to ensure that this PR will not break the previous compatibility!
Requesting you to do a re-review! 😇

minuscorp (approved):
Now it LGTM 🚀

joshdholtz (approved):
Seems good to me 😊 Thank you!

PR #18738, opened 2021-05-22
[fastlane][runner] Executing deprecated action message improvements

crazymanish:
### Checklist

### Motivation and Context
- Notice that, Runner class was using old-style `put` with the .deprecated method.

### Description
- Now, using Fastlane's `UI.deprecated` which do the same thing with much better results 😉 (i.e consistency, unit-testing)

### Testing Steps
- No functionality changed, Updated the unit tests!


joshdholtz:
I think we want to keep the `puts` here 😇  `UI.deprecated` adds a timestamp onto each line which makes it a little bit harder to read where `puts` with the `.deprecated` has no timestamp and just colors it blue.

E938A108-ACCF-4D65-B84F-C0C5FA8359F0

crazymanish:
Okay, thank you! I agree! Closing this PR now! :)

PR #18766, opened 2021-05-28
[spaceship] parse response body as json if not already

ilendemli:
### Checklist

### Motivation and Context
`spaceship` failed to parse an api-responses such as expired membership when running `fastlane match`. This MR checks if the respone body is an object of type `String` and parses the body as a JSON object if that is the case (better handling for #18431 and similar failures).

### Description
updates `handle_error` and `format_errors` in `Spaceshipt::ConnectAPIT::APIClient` and checks wether `response.body` is empty or not. uses an empty object if String is empty. Then parses result as JSON if the result is tyoe of string

### Testing Steps
`rspec` and `rubocop` went through as-well as `bundle exec fastlane test`

### DIFF
--- a/spaceship/lib/spaceship/connect_api/api_client.rb
+++ b/spaceship/lib/spaceship/connect_api/api_client.rb
@@ -212,14 +212,14 @@ module Spaceship
 
         case response.status.to_i
         when 401
-          raise UnauthorizedAccessError, format_errors(response) if response && body['errors']
+          raise UnauthorizedAccessError, format_errors(response)
         when 403
           error = (body['errors'] || []).first || {}
           error_code = error['code']
           if error_code == "FORBIDDEN.REQUIRED_AGREEMENTS_MISSING_OR_EXPIRED"
-            raise ProgramLicenseAgreementUpdated, format_errors(response) if response && body['errors']
+            raise ProgramLicenseAgreementUpdated, format_errors(response)
           else
-            raise AccessForbiddenError, format_errors(response) if response && body['errors']
+            raise AccessForbiddenError, format_errors(response)
           end
         end
       end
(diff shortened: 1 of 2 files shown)


max-ott:
Hi @ilendemli 👋 ,
Looks like a great PR! Do you have any tests steps for us to test this via the `spaceship playground` or similar? What kind of "errors" / test cases did you use besides expired membership?

spaceship/lib/spaceship/connect_api/api_client.rb
```diff
@@ -221,6 +222,14 @@ def handle_error(response)
         end
       end
 
+      def response_body_as_json(body)
+        if body.kind_of?(Hash)
+          return body
+        end
+
+        return JSON.parse(body)
```
joshdholtz, line 230: We might want to rescue an error here if it can't parse and display something _somewhat_ user friendly 🤔 Thoughts?

joshdholtz (changes requested):

ilendemli:
> Hi @ilendemli 👋 ,
> Looks like a great PR! Do you have any tests steps for us to test this via the `spaceship playground` or similar? What kind of "errors" / test cases did you use besides expired membership?

Hey @max-ott!

I traced the call:
`Spaceship::ConnectAPI::Provisioning::Client` sends a request by calling the request method from its superclass `Spaceship::ConnectAPI::APIClient` which is a subclass of `Spaceship::Client`. `Spaceship::ConnectAPI::APIClient` overrides the `handle_error` method but the passed object to `Spaceship::Client` is a String and not a JSON/Hash.

I am now doing the same as in `fetch_olympus_session` in [Spaceship::Client#L542](https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L542).

Not sure if I have missed something else.

spaceship/lib/spaceship/connect_api/api_client.rb
ilendemli: I have changed this part of the code, but I am doing the same as in other places.

dede760 (commented):
.

dede760:
https://www.facebook.com/deniez.jeloo

dede760:
https://www.facebook.com/deniez.jeloo

joshdholtz:
@google-admin I consent.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
Thanks for this PR! I added a little bit to it and some tests. I really appreciate the contribution ❤️

PR #18782, opened 2021-05-31
[action][add_git_tag] remove all instances of `is_string` in options and use `type`

crazymanish:
### Checklist

### Motivation and Context
- This PR does this for only `add_git_tag` action.

### Description
- Remove `is_string` from Boolean params and added `type: Boolean`
- Remove `is_string` from build_number param ~and added `type: Integer`~
- Added missing unit-tests for `includes_lane` param

### Testing Steps

### DIFF
--- a/fastlane/lib/fastlane/actions/add_git_tag.rb
+++ b/fastlane/lib/fastlane/actions/add_git_tag.rb
@@ -63,7 +63,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :includes_lane,
                                        env_name: "FL_GIT_TAG_INCLUDES_LANE",
                                        description: "Whether the current lane should be included in the tag and message composition, e.g. '<grouping>/<lane>/<prefix><build_number><postfix>'",
-                                       is_string: false,
+                                       type: Boolean,
                                        default_value: true),
           FastlaneCore::ConfigItem.new(key: :prefix,
                                        env_name: "FL_GIT_TAG_PREFIX",
@@ -78,7 +78,7 @@ module Fastlane
                                        description: "The build number. Defaults to the result of increment_build_number if you\'re using it",
                                        default_value: Actions.lane_context[Actions::SharedValues::BUILD_NUMBER],
                                        default_value_dynamic: true,
-                                       is_string: false,
+                                       type: Integer,
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :message,
                                        env_name: "FL_GIT_TAG_MESSAGE",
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/add_git_tag.rb
```diff
@@ -78,7 +78,7 @@ def self.available_options
                                        description: "The build number. Defaults to the result of increment_build_number if you\'re using it",
                                        default_value: Actions.lane_context[Actions::SharedValues::BUILD_NUMBER],
                                        default_value_dynamic: true,
-                                       is_string: false,
+                                       type: Integer,
```
crazymanish, line 81: Earlier: `is_string: false` was supporting both type inputs `Integer + String`
Now: `type: Integer` will support only `Integer` input

Q: is it okay? or should we auto-convert value for `type: Integer` from string inside FastlaneCore::ConfigItem ?

lucgrabowski: It looks like `default_value`is populated in `increment_build_number` as a String
https://github.com/fastlane/fastlane/blob/3456bdebc4a3089a0fd147f501c81557f74a5150/fastlane/lib/fastlane/actions/increment_build_number.rb#L51-L54

minuscorp: I was thinking on adding an `Any` type for this cases which is just a placeholder for a parameter that can be an integer, a string, a double, a float or a symbol.

crazymanish: > I was thinking on adding an `Any` type for this cases which is just a placeholder for a parameter that can be an integer, a string, a double, a float or a symbol.

Agree! this makes sense to me! 💯 

> So maybe this can wait a bit until this is implemented if no rush involved.

Agree! shall I keep using `is_string` **only for this param** for now? we can revisit again for such tricky params once we will have `Any` type support! what do you say @minuscorp?

joshdholtz: This can kind of be done with `skip_type_validation: true` 🤷‍♂️

crazymanish: > This can kind of be done with `skip_type_validation: true` 🤷‍♂️

Ohh really, this is 💡 I will try that and keep you all posted in this thread! Thank you 🙇

crazymanish: I can confirm, `skip_type_validation: true` works! YaY... 🎉

fastlane/lib/fastlane/actions/add_git_tag.rb
```diff
@@ -78,7 +78,7 @@ def self.available_options
                                        description: "The build number. Defaults to the result of increment_build_number if you\'re using it",
                                        default_value: Actions.lane_context[Actions::SharedValues::BUILD_NUMBER],
                                        default_value_dynamic: true,
-                                       is_string: false,
+                                       skip_type_validation: true, # skipping validation because we allow integer and string both
```
minuscorp, line 81:
```suggestion
                                       skip_type_validation: true, # skipping validation because we both allow integer and string
```

minuscorp (approved):
🚀

joshdholtz (approved):
🔥 🔥 🔥

Issue #18722, opened 2021-05-20
Snapshot is not saving screenshots


alexnot95:
### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [ ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Snapshot is not saving screenshots. I've updated snapshot to 1.25 version and fastlane version is 2.183.1
I'm running my localization test suite, tests passed successfuly but folders with screenshots are empty :( 
Running on xcode 12.5 + ios 14.5 simulator
![Screenshot 2021-05-20 at 14 39 23](https://user-images.githubusercontent.com/37154209/118972395-2f427d00-b979-11eb-9e2d-9c9c95fdbc20.png)

My lane:
```
desc 'Run localization UI tests'
lane :native_localization_ui_test do |options|
    
  fastlane_require 'fastlane-plugin-test_center'

  clean_up_and_install_dependencies
  scheme = ENV["UI_TESTS_SCHEME"] || "Acceptance"
  testplan = ENV["UI_TEST_PLAN"]
  workspace = "#{ENV["MAIN_TARGET"]}.xcworkspace"
  derived_data_path = "~/Library/Developer/Xcode/DerivedData"
  cloned_source_packages_path = "~/Library/Caches/SPMCloneCaches"
  devices = ENV["UI_TESTS_DEVICE_TYPE"]
  output_dir = "./fastlane/test_output"
  languages = ENV["UI_TESTS_LANGUAGE"]

  scan(
    workspace: workspace,
    devices: devices,
    scheme: scheme,
    testplan: testplan,
    skip_slack: true,
    derived_data_path: derived_data_path,
    cloned_source_packages_path: cloned_source_packages_path,
    build_for_testing: true
  )

  snapshot(
    workspace: workspace,
    devices: devices,
    scheme: scheme,
    testplan: testplan,
    languages: languages,
    localize_simulator: true,
    output_directory: output_dir,
    clear_previous_screenshots: true,
    output_simulator_logs: true,
    erase_simulator: true,
    result_bundle: true,
    number_of_retries: 0,
    skip_open_summary: true,
    test_without_building: true,
    derived_data_path: derived_data_path,
[...]

PR #18797, opened 2021-06-03
[snapshot][examples] Fixed broken example project for trying out fastlane snapshot feature

bdolewski:
### Checklist

### Motivation and Context
Recently I decided to use `fastlane snapshot` for the first time. However I encounter the same issue as here: https://github.com/fastlane/fastlane/issues/18722

I tried to use minimal environment to see if it's working with bare minimum and decided to use this documentation:
https://docs.fastlane.tools/getting-started/ios/screenshots/ and clone the test code:

```
git clone https://github.com/fastlane/fastlane  # Clone the fastlane repo
cd fastlane/snapshot/example                    # Navigate to the example project
fastlane snapshot                               # Generate screenshots for the sample app
```

However it doesn't even build because the sample code is simply outdated.

### Description
- Updated Xcode project files to recommended settings
- Updated Xcode project to use Swift 5
- to be faithful with documentation then I removed the `snappy` lane and stick with `Snapfile`
- updated test simulators to be iPhone 12 Pro Max and SE (2nd generation) instead of iPhones 6/7

### Testing Steps
Just running `fastlane snapshot` like in documentation to see if I can get screenshots right.


minuscorp:
We keep the old configuration settings to ensure compatibility with all further versions. This project must be unchanged she if something fails to compile I think we rather fix the issue than changing the whole project.

joshdholtz:
@minuscorp This update is needed 😀👇  The project itself was super old and needed an upgrade. This seems like a good update so the project actually does run 😇 

Screen Shot 2021-06-05 at 3 25 49 PM

joshdholtz (approved):
One thing that feels a little bit weird about this is the `SnapshotHelper.swift` file being committed into the project because it could become out of date. But at the same time... the project works perfectly by just running `fastlane snapshot` and it's super easy to update `SnapshotHelper.swift` when running the example project again.

The upgrade to Swift 5 is also totally needed! This is 🔥 Thanks so much for fixing this!

bdolewski:
Thanks @joshdholtz @minuscorp for feedback, really appreciate it!

The deployment target is set to iOS 12 because that's what Xcode recommended. I can change it to some lower version, like iOS 9. However anything below iOS 9 is just... ridiculous 🙃

I'd not touch the Objective-C code since there is so much point of rewriting that part other than just getting rid of Objective-C for the sake of it.

I'd like to keep the changes to `Fastline` and `Snapfile` as proposed since that would be perfectly aligned to documentation.

minuscorp:
Just for the record, in the fastlane Swift project we support down to Swift 4 settings, that's why I was afraid of bumping so much from Swift 3 (that we all know is deprecated), to avoid misalignment from the _main_ project. Just as the SPM support stablishes as well, but we'd need to downgrade Xcode to a Swift 4 release and make the same changes that have been merged here.

crazymanish:
Hi,
This PR got merged into master without running Unit-tests on CI 😉  (maybe because this was @bdolewski first PR...)
Fixed the broken CI in this PR: https://github.com/fastlane/fastlane/pull/18811

Thank you!

bdolewski:
> Hi,
> This PR got merged into master without running Unit-tests on CI 😉 (maybe because this was @bdolewski first PR...)
> Fixed the broken CI in this PR: #18811
> 
> Thank you!

Thanks, yes - that was my 1st PR :) Did I miss something then?

crazymanish:
> > Hi,
> > This PR got merged into master without running Unit-tests on CI 😉 (maybe because this was @bdolewski first PR...)
> > Fixed the broken CI in this PR: #18811
> > Thank you!
> 
> Thanks, yes - that was my 1st PR :) Did I miss something then?

Ahhh...no you did not miss anything but CI did not run unit-tests for this PR, CircleCI don't run unit-tests for 1st PR (maybe because of security issue 🤷‍♂️)

Fastlane CI is running unit-tests with the old Xcode version too i.e `Xcode 10.2.1`, which does not support `iPhone 12 Pro Max` simulator added in this PR example snapfile 😉 
btw, Nothing to worry about...fixed the broken CI in this PR: #18811  😇
Enjoy your weekend!

PR #17075, opened 2020-08-19
[resign] add full sed path for resign.sh

keith:
If users have gnu-sed from brew in their PATH, these invocations fail
since they used BSD sed specific syntax.

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] ~I've updated the documentation if necessary.~

PR #18804, opened 2021-06-04
[resign] fix app entitlements extraction

kapfab:
* Fix the way app entitlements are extracted (see https://github.com/fastlane/fastlane/pull/18713)
* Homogenise sed invocations (following https://github.com/fastlane/fastlane/pull/17075)

### Checklist
- [ ] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass → N/A
- [ ] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid → N/A
- [ ] I've updated the documentation if necessary. → N/A

### Motivation and Context
https://github.com/fastlane/fastlane/pull/18713 allowed `resign --use-app-entitlements` to handle entitlements that were ignored before.
Due to the very long time this PR (and its ancestors) took before being accepted, one of the initial changes seems to have been lost in the multiple merge attempts.
This leads to an issue when using `--use-app-entitlements` to resign an app having iCloud capability.

This fix resolves this remaining issue related in #18713.

### Description
The way `resign.sh` extracts app entitlements have always been faulty (headers are not removed as expected due to `sed` limitations), but this went unnoticed as `plutil` is permissive enough to ignore them subsequently.
However, these unexpected headers break a new entitlement patching method introduced by #18713.

As `sed` cannot handle the single-line-mode regular expression used to extract entitlements, this change removes newlines before calling it (entitlements being made of basic XML elements, newlines can be safely removed).

Also, this change homogenizes `sed` calls and follows #17075 by adding the full path of the system `sed` command to all `sed` invocations.

`resign` is the ugly duckling of Fastlane, it is based on a decade-old shell script and comes with no automated tests.
Though, resigning with these changes have been tested by myself and @csc-EricWu (see https://github.com/fastlane/fastlane/pull/18713#issuecomment-854966645).

### DIFF
--- a/sigh/lib/assets/resign.sh
+++ b/sigh/lib/assets/resign.sh
@@ -78,6 +78,8 @@
 # new features May 2021
 # 1. fix entitlements merging when changing team
 #
+# new features June 2021
+# 1. fix the way app entitlements are extracted
 
 # Logging functions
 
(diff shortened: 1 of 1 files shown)


kapfab:
@joshdholtz, this very simple PR fixes an issue introduced by #18713, may you have a quick look at it?

joshdholtz (approved):
Seems good to me 😊 Thank you for fixing this! And resign is indeed the ugly duckling of fastlane 😱 I wish it wasn't so I appreciate people like you and PRs like this that are able to fix and improve it ❤️

kapfab:
You’re welcome. Thanks a lot for this express validation!

PR #18815, opened 2021-06-07
[fastlane_core] add pods project schemes environment variable

tejassharma96:
### Checklist

### Motivation and Context

At my workplace for one of our projects, we run a test spec from one of our internal cocoapods as a smoke test before running lengthier UI tests that depend on that pod. I was able to use the cocoapods `post_install` block to make the scheme public when installing on CI, but we're still unable to run those tests on fastlane since fastlane filters out pods schemes.

### Description

I added an environment variable to optionally disable the filtering out of environment variables.
I tested this out by verifying that the scheme could be found with the environment set to a truthy value, and it could not be found when the environment variable is unset or set to a falsy value.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -123,9 +123,13 @@ module FastlaneCore
     # Get all available schemes in an array
     def schemes
       @schemes ||= if workspace?
-                     workspace.schemes.reject do |k, v|
-                       v.include?("Pods/Pods.xcodeproj")
-                     end.keys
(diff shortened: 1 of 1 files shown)


tejassharma96:
@googlebot I signed it!

tejassharma96:
whoops looks like I forgot to commit the linting autocorrections, will do so now.

joshdholtz (approved):
This makes sense to me! Sorry if this caused any issues but thank you for fixing 😊

tejassharma96:
Thanks for the quick merge!

PR #18833, opened 2021-06-08
[action][badge] remove all instances of `is_string` in options and use `type`

crazymanish:
### Checklist

### Motivation and Context
- This PR does this for only `badge` action.

### Description
- Remove `is_string: true` from options
- Added `type: Boolean` 

### Testing Steps

### DIFF
--- a/fastlane/lib/fastlane/actions/badge.rb
+++ b/fastlane/lib/fastlane/actions/badge.rb
@@ -65,7 +65,7 @@ module Fastlane
                                        env_name: "FL_BADGE_DARK",
                                        description: "Adds a dark flavored badge ontop of your icon",
                                        optional: true,
-                                       is_string: false,
+                                       type: Boolean,
                                        verify_block: proc do |value|
                                          UI.user_error!("dark is only a flag and should always be true") unless value == true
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/badge.rb
```diff
@@ -110,33 +108,31 @@ def self.available_options
                                        env_name: "FL_BADGE_SHIELD_IO_TIMEOUT",
                                        description: "Set custom duration for the timeout of the shields.io request in seconds",
                                        optional: true,
-                                       is_string: false,
+                                       skip_type_validation: true, # allow integers, strings both
```
joshdholtz, line 111: We can actually use `type: Integer` here! A string will be converted to int with `.to_i`. We only want to use `skip_type_validation` if an int and string are both allowed and the action doesn't care about the type (which isn't very common)

```suggestion
                                       type: Integer, # allow integers, strings both
```

joshdholtz (changes requested):
One small change otherwise 🔥

joshdholtz (approved):
🔥 🔥 🔥

PR #18837, opened 2021-06-08
[action][changelog_from_git_commits] remove all instances of `is_string` in options and use `type`

crazymanish:
### Checklist

### Motivation and Context
- This PR does this for only `changelog_from_git_commits` action.

### Description
- Remove `is_string: true` from options
- Added `type: Boolean` 

### Testing Steps


fastlane/lib/fastlane/actions/changelog_from_git_commits.rb
```diff
@@ -79,7 +79,7 @@ def self.available_options
                                        env_name: 'FL_CHANGELOG_FROM_GIT_COMMITS_BETWEEN',
                                        description: 'Array containing two Git revision values between which to collect messages, you mustn\'t use it with :commits_count key at the same time',
                                        optional: true,
-                                       is_string: false,
+                                       skip_type_validation: true, # allow Array, String both
```
joshdholtz, line 82: I think making `type: Array` actually allows both a String and an Array of Strings so I think this is what we want here 😊 

```suggestion
                                       type: Array, # allow Array, String both
```

joshdholtz (changes requested):
One small suggestion on using `type: Array`

joshdholtz (approved):
Looks great! Thank you thank you 😊

PR #18845, opened 2021-06-09
[action]Adds action for obtaining remote repo default branch

seanmcneil:
### Checklist

### Motivation and Context
This introduces a new action to obtain the default branch name from a remote git repository. Doing this can support other actions such as creating merge requests by ensuring that the proper branch name is provided as a target. With the shift away from "master", it is beneficial to have this kind of flexibility.

### Description

A new action has been created in the source control category that enables retrieving a remote git repository's default branch.

It was tested under the following conditions:
- In a folder with no available git repo (an empty string was returned)
- In a folder with a git repo, but no available remote (an empty string was returned)
- In a folder with a valid git repo and the local branch on the default remote (correct branch name returned)
- In a folder with a valid git repo and the local branch NOT on the default remote (correct branch name returned)

### Testing Steps

- Scenario 1, test in a folder with no git repo. Run `bundle exec fastlane run git_default_remote_branch`
- Scenario 2, test in a folder only a local git repo.
  - Init a git repo, commit a file
  - Run `bundle exec fastlane run git_default_remote_branch` 
- Scenario 3, test in a folder with a valid remote repo
  - Clone a repo, ensure you are on the default branch
  - Run `bundle exec fastlane run git_default_remote_branch` . Default name should be seen 
- Scenario 4, test in a folder with a valid remote repo while not on default branch
  - Clone a repo, switch to a non-default branch
  - Run `bundle exec fastlane run git_default_remote_branch` . Default name should be seen


fastlane/lib/fastlane/actions/git_default_remote_branch.rb
```diff
@@ -0,0 +1,51 @@
+module Fastlane
+  module Actions
+    class GitDefaultRemoteBranchAction < Action
+      def self.run(params)
+        Actions.git_default_remote_branch_name || ""
```
crazymanish, line 5: Hi, one small **humble request**...Please add unit tests for this action 🙇‍♂️
One of the reasons the Fastlane team doesn't allow new action because it adds lots of maintenance and unit tests really help to reduce the maintenance workload and any enhancement work.

You can add unit tests for all 4 scenarios, you mentioned in the PR description 😇
Thank you!

naveentnj:
The mails are so anonymous

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
seanmcneil: No worries. I have added tests for verifying the command and the first two test cases. However, I have not been able to get the remote working just yet so it may take some time.

seanmcneil: I added tests for all the conditions. However, this is the first time I have written anything like this in ruby so I probably got something wrong. If so, I'd be happy to sort it out if someone can point me in the right direction.

fastlane/spec/actions_specs/git_default_remote_branch.rb
```diff
@@ -0,0 +1,101 @@
+describe Fastlane do
+  describe Fastlane::FastFile do
+    directory = "fl_spec_default_remote_branch"
+    repository = "https://github.com/seanmcneil/Empty.git"
```
crazymanish, line 4: Naah...not this way, if this repository will delete for some reason, Fastlane CI will break in `git clone` step.
I will recommend unit tests using mocks with fake/stub responses. 

This is a nice spec where you can see how to mock git commands using fake/stub responses
https://github.com/fastlane/fastlane/blob/master/fastlane/spec/actions_specs/push_to_git_remote_spec.rb

Please let me know if need more help! Happy to help :innocent:

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
```diff
@@ -0,0 +1,49 @@
+module Fastlane
+  module Actions
+    class GitDefaultRemoteBranchAction < Action
+      def self.run(params)
+        Actions.git_default_remote_branch_name || "No remote default available"
```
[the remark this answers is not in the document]
joshdholtz: I agree here! I would either return nil or raise an error 🤷‍♂️  I think returning nil makes the most sense though!

fastlane/lib/fastlane/helper/git_helper.rb
```diff
@@ -139,6 +139,17 @@ def self.git_branch_name_using_HEAD
       end
     end
 
+    # Returns the default git remote branch name
+    def self.git_default_remote_branch_name
```
joshdholtz, line 143: Can we add an option on here to choose the remote it looks it? 😊 Some of my repos have multiple remotes

And then also add this as an option on the action?

joshdholtz (changes requested):
This looks great! I totes see a need for this. Just got two small requests 💪

fastlane/lib/fastlane/helper/git_helper.rb
seanmcneil: Sure, I can look into that.

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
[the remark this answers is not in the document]
seanmcneil: Ditto. @crazymanish explanation made the use case very clear and I have updated to return nil now - thanks!

fastlane/spec/actions_specs/git_default_remote_branch.rb
seanmcneil: I completely agree and had this specific item in mind with my "I'm not sure I am doing this right comment". I am still working on this item and it may be a bit before I can resume work and correct it.

fastlane/lib/fastlane/helper/git_helper.rb
seanmcneil: Done & includes a unit test and example code.

fastlane/spec/actions_specs/git_default_remote_branch.rb
seanmcneil: I updated the tests based on the push spec you linked. My use case was a little different, but I hope I implemented it correctly. Please let me know if there is anything I missed.

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
```diff
@@ -0,0 +1,57 @@
+module Fastlane
+  module Actions
+    class GitDefaultRemoteBranchAction < Action
```
crazymanish, line 3: I think, we can `omit` Default from this action name now...because now this action accepts custom `remote_name` input param, thoughts? 🤔

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
```diff
@@ -0,0 +1,57 @@
...
+      def self.run(params)
+        Actions.git_default_remote_branch_name(params[:remote_name])
```
crazymanish, line 5: ditto ⬆️
I think, this can become just `Actions.git_remote_branch_name(params[:remote_name])`

crazymanish (approved):
This looks good to me, one minor suggestion if you agree! 🚀

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
seanmcneil: Done 👍

fastlane/lib/fastlane/actions/git_default_remote_branch.rb
seanmcneil: Works for me & already implemented in most recent commit 👍

seanmcneil:
@joshdholtz Can you please check on your change request?

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This is great! Thank you so much for adding that `remote_name` option 🙌  I added one commit to clean up some syntax but other than that this is 💯  Really appreciate this ❤️

PR #18937, opened 2021-06-22
Fix docs of available options of `increment_version_number`

jklausa:
This is a simple docs fix, to align them with what the actual options this action takes is.


jklausa:
I'm sorry, but I'm not signing a CLA for this, feel free to cherry-pick this patch, I hereby release it into public domain.

PR #18971, opened 2021-06-25
[action][retrieve_devices] New Action | Retrieve and log all devices for an Apple Certificate

builtbyproxy:
### Checklist
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid | **~I can't get this to pass. If I run `new_action` it will generate it in a directory other than `lib/fastlane/actions`, but then rubocop will fail. It's also not liking my naming of the file?~** (seems to build in the PR)

### Motivation and Context
When we register a device, we do so using the [register_devices action](https://docs.fastlane.tools/actions/register_devices/). However, often at times it feels like these devices are being registered to a box. At times we have the liberty of checking our [developer account](https://developer.apple.com/). However, with so much happening in the CLI, it's odd to jump out and into a UI. 

There's a bigger option here to parse the result to `register_devices` but I would like some general input before moving closer to that goal. 

This is very much just a check right now, hence why we only map the `name` and `udid` to our `existing_devices` sharedValue.

### Description
This action takes what exists in the [register_devices action](https://docs.fastlane.tools/actions/register_devices/) and moves it to a QOL action called `retrieve_devices`. The intention is to allow us to quickly loop through what we are changing, ensure the changes exist, and move on. 

This isn't a crazy change, its simply intended to give more visibility over an already cloudy process.

### Testing Steps

Run `fastlane action retrieve_devices` and `fastlane run retrieve_devices` and ensure the output is to your satisfaction for both the project AND the documented use case. Thank you

### DIFF
--- a/fastlane/actions/retrieve_devices.rb
+++ b/fastlane/actions/retrieve_devices.rb
@@ -1,0 +1,104 @@
+module Fastlane
+  module Actions
+    module SharedValues
+      DEVICES_FOR_APPLE_CERTIFICATE = :DEVICES_FOR_APPLE_CERTIFICATE
+    end
+
+    class RetrieveDevicesAction < Action
+      def self.run(params)
+        require 'spaceship'
+        if (api_token = Spaceship::ConnectAPI::Token.from(hash: params[:api_key], filepath: params[:api_key_path]))
+          UI.message("Creating authorization token for App Store Connect API")
+          Spaceship::ConnectAPI.token = api_token
+        elsif !Spaceship::ConnectAPI.token.nil?
+          UI.message("Using existing authorization token for App Store Connect API")
+        else
+          UI.message("Login to App Store Connect (#{params[:username]})")
+          credentials = CredentialsManager::AccountManager.new(user: params[:username])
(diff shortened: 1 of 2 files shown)


builtbyproxy:
@joshdholtz When you're busy would you mind inputting on this please? 
I've filled out the PR but can't seem to get the tests to pass, they are failing for things I have no control over: 
- Name of the action (seems fine?)
- Location of the action ( I generated with `fastlane new_action` )

Thank you!

joshdholtz:
@builtbyproxy Heyyyy 👋 This action does look pretty useful 😊 We are currently trying to limit the number of actions we are accepting. This seems like something that might be a good fit for a plugin? http://docs.fastlane.tools/plugins/create-plugin/#create-your-own-fastlane-plugin

This exact code can be placed directly in the plugin 😁

builtbyproxy:
Thanks for your patience @joshdholtz, i've created the plugin here: https://github.com/builtbyproxy/fastlane-plugin-retrieve-devices

How do I get it added to the list of available plugins [here](https://docs.fastlane.tools/plugins/available-plugins/)

builtbyproxy:
Bump on this if possible? Thanks!

PR #18003, opened 2021-01-19
[screengrab] general improvements, removed deprecated and un-needed code

penn5:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Currently, screengrab does not work on many devices, including Android 11 (due to scoped storage). Plus, many dependencies are outdated, and internal API hacks are (ab)used.
This PR fixes all the issues I could find.

### Description

1. Updated docs - removed unneeded permissions, fixed grammar, added simpler idioms
2. Removed use of ActivityTestRule which is deprecated
3. Updated gradle wrapper and removed bintray-release plugin which is outdated
4. Changed target sdk to 30
5. Updated dependencies
6. Attempted (but failed) to fix issue where UiAutomatorScreengrabStrategy randomly captures too early/late. The only workaround I know is to put a Thread.sleep(1000) before each screenshot
7. Removed need for READ/WRITE_EXTERNAL_STORAGE permission on API > 18
8. Removed dependency on AAPT (it was only used to validate APKs, and the error is clear from stack traces generated by ADB if the user made a mistake anyway)
9. Deprecated use_adb_root as it should be useless now (screenshots should always be readable by ADB)
10. Use $ANDROID_DATA and $EXTERNAL_STORAGE when fetching screenshots (the path changes)
11. Don't error when a permission grant fails (just warn)
12. Kill app between tests to make sure the clean status bar is properly displayed
13. Disable hidden API checks during tests, as LocaleUtils violates them
14. Add support for restoring the previous locale (for people who don't actually use en-US as their language)
15. Deprecated and ignored ending_locale, since it is automatically set to what the tests started at.
16. Added theoretical suppport for testing a mix of locales at once (e.g. fr-FR, fallback to fr-CA). I didn't implement any way to actually use this, but it is used when restoring the original locale.
17. Add shellescape the correct number of times on all variables passed to ADB

I wrote unit tests and tested on my app, as well as the sample app

### Testing Steps

cd screengrab/example && bundler exec fastlane build_and_screengrab

PR #19019, opened 2021-06-29
[screengrab] fix typo

penn5:
### Checklist

### Motivation and Context

Fixes bug from #18003

### Description

Otherwise the run_as state is ignored because the array is _too_ flat

### Testing Steps

Run against Android R


penn5:
@joshdholtz my last patch broke android R (not that it worked before, but hey) so here's the fix

joshdholtz (approved):
This makes sense to me! Thank you again for your work on this 😊

PR #19034, opened 2021-06-30
[scan] add support for SPM packages with new :package_path option

### LINKED ISSUE #17734: Scan: Can't test Swift Package Manager packages using scheme only
### New Issue Checklist

### Issue Description
As described here: https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#depending-on-apple-modules you can test packages easily by just running `xcodebuild` using `-scheme PACKAGE_NAME`. However, running Scan without a project or workspace automatically fails.

##### Command executed
Scan using a `scheme` only.

There's not really much to tell, other than this is what you'll get:

image

<details>
</details>

### Environment
This is the case with any Fastlane version, I'm pretty sure the environment doesn't impact this.

# Expected behavior
I would love to see Scan accepting a certain `package_testing` property to skip scheme settings. I've tried looking into contributing myself but there are a lot of features dependent on having a project reference.

I feel like this impacts quite some structural changes to make this work. However, it will add support for Package testing within Fastlane, which would be very welcome.

joshdholtz:
### Motivation and Context
Fixes #17734

### Description

Adds support for testing SPM packages with `scan`

- New `:package_path` option
  - Conflicts with `:project`

### Testing Steps

```rb
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "joshdholtz-scan-spm-packages"
```

### DIFF
--- a/scan/examples/package/.gitignore
+++ b/scan/examples/package/.gitignore
@@ -1,0 +1,7 @@
+.DS_Store
+/.build
+/Packages
+/*.xcodeproj
+xcuserdata/
+DerivedData/
+.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
--- a/scan/examples/package/.swiftpm/xcode/package.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
+++ b/scan/examples/package/.swiftpm/xcode/package.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
--- a/scan/examples/package/Package.swift
+++ b/scan/examples/package/Package.swift
@@ -1,0 +1,28 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "package",
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "package",
+            targets: ["package"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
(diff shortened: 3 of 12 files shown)


AvdLee (approved):
Tried this out locally and it's working great!

Dim0N22:
Tested it locally, works for me as well. Thank you!

There is a scenario where `scan` still doesn't work though: having both `.xcodeproj` and `Package.swift` in the same package.
Xcode distinguishes between them depending on which path is opened:
- `open package.xcodeproj` to open a package as an old Xcode project
- `open Package.swift` to open a package as an SPM package
 
Same logic is applied to using `xcodebuild` from command line. To run tests for an SPM package that also declares an `.xcodeproj` we can use
```
xcodebuild -scheme PackageScheme -workspace "." -destination 'platform=iOS Simulator,name=iPhone 12' build test
```
Note `-workspace "."`. It prevents `xcodebuild` from using `.xcodeproj` and it uses `Package.swift` instead.

To generate the same command using `scan` I configured it like this:
```
scan(
  package_path: package_path,
  workspace: '.', # ignore .xcodeproj if it exists
  destination: 'platform=iOS Simulator,name=iPhone 12'
  scheme: "PackageScheme",
  configuration: "Debug",
  code_coverage: true
)
```
but it causes 2 problems:
1. Scan [checks](https://github.com/fastlane/fastlane/blob/0aeca562ccbb71bcc171a7db0ba46fcf17746b4f/scan/lib/scan/options.rb#L27) that the `workspace` path contains `.xcworkspace`, but in this case it's not true;
2. With this check disabled, the workspace is still not getting passed to the resulting `xcodebuild` command generated by `scan`: 
```
set -o pipefail && cd <package_path> && env NSUnbufferedIO=YES xcodebuild -scheme PackageScheme -destination 'platform=iOS Simulator,name=iPhone 12' -enableCodeCoverage YES build test | tee ...
```
Note the missing `-workspace "."`.

So what can be fixed to improve support for SPM/Xcode projects:
1. Allow `.` as workspace path (alternatively: remove the workspace path check);
2. Pass the workspace path to `xcodebuild` if it is set in `scan`.

BTW, [xcodebuild](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/xcodebuild.rb) works great with passing `workspace: '.'`.

milch (approved):
LGTM, ship it 🚀

joshdholtz:
@Dim0N22 Thanks for the detailed test scenario! I added a commit to fix this and wrote some tests around it. Do you mind testing this out again whenever you get time? 😇

Dim0N22:
@joshdholtz thanks, will test it today.

In the meantime I wanted to ask about the new `package_path` parameter. Given that I call `scan` from a `Fastfile`, it looks like the `package_path` will always be `.` (or `package_path: '.'`) for me (meaning we just want to build the current package). 

Is it the intended behavior? I guess it will be different only for running `scan` from the command line?

Issue #17899, opened 2021-01-04
Seems pilot uses non-exist param value in Apple Connect API


myme5261314:
```shell
fastlane pilot list --api_key_path XX.json --app_identifier "XX" --verbose
```
<details><summary>Traceback Output</summary>

```shell

Traceback (most recent call last):
	18: from /usr/local/bin/fastlane:23:in '<main>'
	17: from /usr/local/bin/fastlane:23:in 'load'
	16: from /Users/ben/.gems/gems/fastlane-2.171.0/bin/fastlane:23:in '<top (required)>'
	15: from /Users/ben/.gems/gems/fastlane-2.171.0/fastlane/lib/fastlane/cli_tools_distributor.rb:111:in 'take_off'
	14: from /Users/ben/.gems/gems/fastlane-2.171.0/pilot/lib/pilot/commands_generator.rb:18:in 'start'
	13: from /Users/ben/.gems/gems/fastlane-2.171.0/pilot/lib/pilot/commands_generator.rb:166:in 'run'
	12: from /Library/Ruby/Gems/2.6.0/gems/commander-fastlane-4.4.6/lib/commander/delegates.rb:15:in 'run!'
	11: from /Users/ben/.gems/gems/fastlane-2.171.0/fastlane_core/lib/fastlane_core/ui/fastlane_runner.rb:76:in 'run!'
	10: from /Library/Ruby/Gems/2.6.0/gems/commander-fastlane-4.4.6/lib/commander/runner.rb:476:in 'run_active_command'
	 9: from /Library/Ruby/Gems/2.6.0/gems/commander-fastlane-4.4.6/lib/commander/command.rb:153:in 'run'
	 8: from /Library/Ruby/Gems/2.6.0/gems/commander-fastlane-4.4.6/lib/commander/command.rb:178:in 'call'
	 7: from /Users/ben/.gems/gems/fastlane-2.171.0/pilot/lib/pilot/commands_generator.rb:114:in 'block (2 levels) in run'
	 6: from /Users/ben/.gems/gems/fastlane-2.171.0/pilot/lib/pilot/tester_manager.rb:76:in 'list_testers'
	 5: from /Users/ben/.gems/gems/fastlane-2.171.0/pilot/lib/pilot/tester_manager.rb:111:in 'list_testers_by_app'
	 4: from /Users/ben/.gems/gems/fastlane-2.171.0/spaceship/lib/spaceship/connect_api/models/app.rb:344:in 'get_beta_testers'
[...]

PR #19039, opened 2021-07-01
[spaceship] Adding three missing beta testers/groups methods

gsavit:
### Checklist

### Motivation and Context

Implementing the following missing TestFlight endpoints for beta testers/groups related functionality:

[Add testers to build](https://developer.apple.com/documentation/appstoreconnectapi/assign_individual_testers_to_a_build)
[Remove testers from build](https://developer.apple.com/documentation/appstoreconnectapi/remove_individual_testers_from_a_build)
[Remove groups from build](https://developer.apple.com/documentation/appstoreconnectapi/remove_access_for_beta_groups_to_a_build)

### DIFF
--- a/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
+++ b/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
@@ -366,7 +366,7 @@ module Spaceship
           test_flight_request_client.post("builds/#{build_id}/relationships/individualTesters", body)
         end
 
-        def delete_beta_testers_from_build(beta_tester_ids: [], build_id: nil)
+        def delete_beta_testers_from_build(build_id: nil, beta_tester_ids: [])
           body = {
             data: beta_tester_ids.map do |id|
               {
--- a/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
+++ b/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
@@ -191,7 +191,7 @@ module Spaceship
           test_flight_request_client.post("builds/#{build_id}/relationships/betaGroups", body)
         end
 
-        def delete_beta_groups_from_build(beta_group_ids: [], build_id: nil)
+        def delete_beta_groups_from_build(build_id: nil, beta_group_ids: [])
           body = {
             data: beta_group_ids.map do |id|
               {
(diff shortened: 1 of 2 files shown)


spaceship/lib/spaceship/connect_api/testflight/testflight.rb
```diff
@@ -191,6 +191,19 @@ def add_beta_groups_to_build(build_id: nil, beta_group_ids: [])
           test_flight_request_client.post("builds/#{build_id}/relationships/betaGroups", body)
         end
 
+        def delete_beta_groups_from_build(beta_group_ids: [], build_id: nil)
```
joshdholtz, line 194:
```suggestion
        def delete_beta_groups_from_build(build_id: nil, beta_group_ids: [])
```

spaceship/lib/spaceship/connect_api/testflight/testflight.rb
```diff
@@ -340,6 +353,32 @@ def add_beta_tester_to_builds(beta_tester_id: nil, build_ids: [])
...
+                type: "betaTesters",
+                id: id
+              }
+            end
+          }
+
+          test_flight_request_client.post("builds/#{build_id}/relationships/individualTesters", body)
+        end
+
+        def delete_beta_testers_from_build(beta_tester_ids: [], build_id: nil)
```
joshdholtz, line 369:
```suggestion
        def delete_beta_testers_from_build(build_id: nil, beta_tester_ids: [])
```

joshdholtz (changes requested):
This is great! Thanks for adding these 😊 Just a few suggestions about the order of method parameters so we can keep things consistent

gsavit:
> This is great! Thanks for adding these 😊 Just a few suggestions about the order of method parameters so we can keep things consistent

For sure, thanks for the review Josh! I wasn't too sure what to do with the ordering as it seemed to already vary within the file 😅

kerrmarin-lvmh:
This is awesome! And I believe it will fix #17899 ?

gsavit:
> This is awesome! And I believe it's a first step towards fixing #17899 ?

Thanks! I don't quite think this has anything to do with that other issue, unless I'm missing something? (Incidentally, separately, I've been poking around looking for `betaTesterMetrics` as well and noticing it's not actually available in ASC API, maybe related to that issue...)

joshdholtz (approved):
This looks good to me! Thank you so much for this PR. Really appreciate it 🥰

PR #19053, opened 2021-07-04
[spaceship] Add visible_apps relationship to invite users with app permissions and fetch user's app permissions

lucgrabowski:
### Checklist

### Motivation and Context

`User` and `UserInvitation` models should have a relationship to `visible_apps` to manage app permissions.

### Description

Changes:
It's possible to assign app permissions when inviting user with `post_user_invitation`.
It's possible to fetch app permissions (`visible_apps`) for both `User` and `UserInvitation`

### Testing Steps

Testing in `irb`:

```ruby
require 'spaceship'
Spaceship::ConnectAPI.login

users = Spaceship::ConnectAPI::User.all(includes: Spaceship::ConnectAPI::User::ESSENTIAL_INCLUDES)
# or
users = Spaceship::ConnectAPI::User.all(includes: "visibleApps")
# or
users = Spaceship::ConnectAPI::User.all

user = Spaceship::ConnectAPI::User.find(email:"USER_EMAIL").first
user.get_visible_apps 
user.get_visible_apps(limit: 1)

user = Spaceship::ConnectAPI::User.find(email:"USER_EMAIL", includes: "visibleApps")


invitations = Spaceship::ConnectAPI::UserInvitation.all(includes: Spaceship::ConnectAPI::UserInvitation::ESSENTIAL_INCLUDES)
# or
invitations = Spaceship::ConnectAPI::UserInvitation.all(includes: "visibleApps")
# or 
invitations = Spaceship::ConnectAPI::UserInvitation.all

Spaceship::ConnectAPI::UserInvitation.create(email: "YOUR_EMAIL", first_name: "YOUR_FIRST_NAME", last_name: "YOUR_LAST_NAME", roles: ["DEVELOPER"], provisioning_allowed: true, all_apps_visible: false, visible_app_ids: ["APP_ID"])
user_invitation = Spaceship::ConnectAPI::UserInvitation.find("YOUR_EMAIL")
user_invitation.get_visible_apps
user_invitation.get_visible_apps(limit: 1)
user_invitation.first.delete!

# Testing of invalid combinations of `all_apps_visible` and `visible_app_ids`:

Spaceship::ConnectAPI::UserInvitation.create(email: "YOUR_EMAIL", first_name: "YOUR_FIRST_NAME", last_name: "YOUR_LAST_NAME", roles: ["DEVELOPER"], provisioning_allowed: true, all_apps_visible: true, visible_app_ids: ["APP_ID"])
# It will fail with "The provided entity includes a relationship with an invalid value - If you set allAppsVisible to true, you must not provide values for the visibleApps relationship. - /data/relationships/visibleApps"

Spaceship::ConnectAPI::UserInvitation.create(email: "YOUR_EMAIL", first_name: "YOUR_FIRST_NAME", last_name: "YOUR_LAST_NAME", roles: ["DEVELOPER"], provisioning_allowed: true, all_apps_visible: false)
# It will fail with "The provided visible app information is invalid. - If you set allAppsVisible to false, you must provide at least one value for the visibleApps relationship. - /data/attributes/allAppsVisible
```

### DIFF
--- a/spaceship/lib/spaceship/connect_api/models/user.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user.rb
@@ -57,10 +57,10 @@ module Spaceship
         return all(client: client, filter: { email: email }, includes: includes)
       end
 
-      def fetch_visible_apps(client: nil)
+      def fetch_visible_apps(client: nil, limit: nil)
         client ||= Spaceship::ConnectAPI
-        resp = client.get_user_visible_apps(user_id: id).all_pages
-        return resp.flat_map(&:to_models)
+        resp = client.get_user_visible_apps(user_id: id, limit: limit)
+        return resp.to_models
       end
     end
   end
--- a/spaceship/lib/spaceship/connect_api/models/user.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user.rb
@@ -57,7 +57,7 @@ module Spaceship
         return all(client: client, filter: { email: email }, includes: includes)
       end
 
-      def fetch_visible_apps(client: nil, limit: nil)
+      def get_visible_apps(client: nil, limit: nil)
         client ||= Spaceship::ConnectAPI
         resp = client.get_user_visible_apps(user_id: id, limit: limit)
         return resp.to_models
--- a/spaceship/lib/spaceship/connect_api/models/user.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user.rb
@@ -16,6 +16,8 @@ module Spaceship
       attr_accessor :email_vetting_required
       attr_accessor :notifications
 
+      attr_accessor :visible_apps
+
       attr_mapping({
         "username" => "username",
         "firstName" => "first_name",
@@ -27,8 +29,14 @@ module Spaceship
         "allAppsVisible" => "all_apps_visible",
         "provisioningAllowed" => "provisioning_allowed",
         "emailVettingRequired" => "email_vetting_required",
-        "notifications" => "notifications"
+        "notifications" => "notifications",
+
+        "visibleApps" => "visible_apps"
       })
+
+      ESSENTIAL_INCLUDES = [
+        "visibleApps"
+      ].join(",")
 
       def self.type
         return "users"
--- a/spaceship/lib/spaceship/connect_api/models/user.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user.rb
@@ -56,6 +56,12 @@ module Spaceship
         client ||= Spaceship::ConnectAPI
         return all(client: client, filter: { email: email }, includes: includes)
       end
+
+      def fetch_visible_apps(client: nil)
+        client ||= Spaceship::ConnectAPI
+        resp = client.get_user_visible_apps(user_id: id).all_pages
+        return resp.flat_map(&:to_models)
+      end
     end
   end
 end
--- a/spaceship/lib/spaceship/connect_api/models/user.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user.rb
@@ -46,13 +46,13 @@ module Spaceship
       # API
       #
 
-      def self.all(client: nil, filter: {}, includes: nil, limit: nil, sort: nil)
+      def self.all(client: nil, filter: {}, includes: ESSENTIAL_INCLUDES, limit: nil, sort: nil)
         client ||= Spaceship::ConnectAPI
         resps = client.get_users(filter: filter, includes: includes).all_pages
         return resps.flat_map(&:to_models)
       end
 
-      def self.find(client: nil, email: nil, includes: nil)
+      def self.find(client: nil, email: nil, includes: ESSENTIAL_INCLUDES)
         client ||= Spaceship::ConnectAPI
         return all(client: client, filter: { email: email }, includes: includes)
       end
--- a/spaceship/lib/spaceship/connect_api/models/user_invitation.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user_invitation.rb
@@ -86,10 +86,10 @@ module Spaceship
       end
 
       # Get visible apps for invited user
-      def fetch_visible_apps(client: nil)
+      def fetch_visible_apps(client: nil, limit: nil)
         client ||= Spaceship::ConnectAPI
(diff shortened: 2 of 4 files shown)


spaceship/lib/spaceship/connect_api/models/user.rb
```diff
@@ -27,9 +29,15 @@ class User
...
         "emailVettingRequired" => "email_vetting_required",
-        "notifications" => "notifications"
+        "notifications" => "notifications",
+
+        "visibleApps" => "visible_apps"
       })
 
+      ESSENTIAL_INCLUDES = [
+        "visibleApps"
+      ].join(",")
```
ainame, line 39: This isn't automatically used at all. You need to pass this to the parent API calls.

For example, you need to pass it to the corresponding methods as the default value of `includes` argument. This way App Store Connect API will include "visibleApps" objects within the response and spaceship client will automatically map it to `visible_apps` property. That's what `ESSENTIAL_INCLUDES` convention is supposed to work.

```ruby
def self.all(filter: {}, includes: Spaceship::ConnectAPI::User::ESSENTIAL_INCLUDES, limit: nil, sort: nil)
```


* https://github.com/fastlane/fastlane/pull/19053/files#diff-75baf27580ebd9af6509c2bbdb9fca7cd67b716bf144964ffc8e208c397d57bcR49
* https://github.com/fastlane/fastlane/pull/19053/files#diff-75baf27580ebd9af6509c2bbdb9fca7cd67b716bf144964ffc8e208c397d57bcR55
* https://github.com/fastlane/fastlane/blob/c0b94f1d5b38700b3b1b57031e9ec2931331ac5e/spaceship/lib/spaceship/connect_api/models/user.rb#L41-L45

ainame: Otherwise `visible_apps` property ends up returning `nil` depending on how you get the `User` object.

spaceship/lib/spaceship/connect_api/models/user_invitation.rb
```diff
@@ -11,15 +11,23 @@ class UserInvitation
...
         "email" => "email",
         "roles" => "roles",
         "allAppsVisible" => "all_apps_visible",
-        "provisioningAllowed" => "provisioning_allowed"
+        "provisioningAllowed" => "provisioning_allowed",
+
+        "visibleApps" => "visible_apps"
       })
 
+      ESSENTIAL_INCLUDES = [
```
ainame, line 27: The same applies to this. You need to pass this to `self.all` and `self.find`.

spaceship/lib/spaceship/connect_api/models/user.rb
```diff
@@ -48,6 +56,12 @@ def self.find(client: nil, email: nil, includes: nil)
         client ||= Spaceship::ConnectAPI
         return all(client: client, filter: { email: email }, includes: includes)
       end
+
+      def fetch_visible_apps(client: nil, limit: nil)
```
ainame, line 60: @joshdholtz Any thought on naming this kind of method with `get` vs `fetch`? 🤔

ainame (changes requested):
Thank you for your contribution! I wanted to point how `ESSENTIAL_INCLUDES` convention is supposed to work🙂 otherwise this looks great change 😍

spaceship/lib/spaceship/connect_api/models/user.rb
lucgrabowski: Thanks for the explanatory comment @ainame. 
I compared how `ESSENTIAL_INCLUDES` are used in `App`, `AppStoreVersion` models and I thought for a while if user `visible_apps` should be fetched by default and I wasn't sure if I should use them as default `inlcudes` argument. 
I added the definition though to allow to use it as in example `testing steps`, e.g.
```ruby
users = Spaceship::ConnectAPI::User.all(includes: Spaceship::ConnectAPI::User::ESSENTIAL_INCLUDES)
```
Otherwise those that would like to fetch it would have to investigate what value should be used in `includes`.

Should `visible_apps` be fetched by default in `self.find` and `self.all` methods?

spaceship/lib/spaceship/connect_api/models/user.rb
lucgrabowski: @ainame This also made me think for a while and compare code in other models and I wasn't sure if the choice is correct. 🤔

ainame: Me too honestly. Some uses "get" others do "fetch" 😅

spaceship/lib/spaceship/connect_api/models/user.rb
ainame: My point is that if it's not "by default", it shouldn't be named "ESSENTIAL". It sounds more like `OPTIONAL_INCLUDES` to me.
https://developer.apple.com/documentation/appstoreconnectapi/list_users

> Should visible_apps be fetched by default in self.find and self.all methods?

To me, it can be an "essential" info when it comes to handling users for ASC. Any advice to determine if it's essential? @joshdholtz

lucgrabowski: If the advice will be to go with optional, I'll rename it to `OPTIONAL_INCLUDES`. Otherwise, I'll add essential as default values for includes parameters in `self.all` and `self.find`.

spaceship/lib/spaceship/connect_api/models/user.rb
joshdholtz: UGGHHHH... I hate naming 😅 I honestly don't know which it should be but I do want it to be consistent at some point 🤷‍♂️

lucgrabowski: I've renamed it to `get` and it's used more in code.

spaceship/lib/spaceship/connect_api/models/user.rb
joshdholtz: Wow, I missed this conversation completely 😱 Sorry! I had my notifications under control for a while but lost control this past month 😬 But...

I think essential is fine here!

joshdholtz (approved):
Tested this out and its working good 😊 Thanks for adding this! Really appreciate the contribution ❤️

PR #19070, opened 2021-07-10
[deliver] introduce a new synchronization logic for screenshots

ainame:
### Checklist

### Motivation and Context

`deliver` doesn't delete existing screenshots on App Store Connect unless you use `overwrite_screenshots` option enabled. This behaviour makes it difficult to use. For example,

* A user who doesn't know the behaviour may upload duplicate old and new screenshots and release it without checking the result as `fastlane` tends to be run on CI and they don't check the console after all
* Even for a mature user, `overwrite_screenshots` is a bit too much as it takes longer time and consuming API calls than synchronising only what is changed

I've been observing the former situation not a few times at my company or on the Internet and suffering the latter situation due to my project at work having a large number of screenshots.

To resolve this, I'd like to introduce a newly written logic to `deliver` in order to sync local screenshots with App Store Connect intuitively and efficiently. I know this can be a breaking logic to existing users so this PR aims to introduce the logic as a _beta_ feature/behaviour in `deliver` and hopefully makes it default behaviour somehow in the next major version of fastlane; maybe it still needs to be "opt-in" to keep the backwards compatibility but `fastlane init` generates `Deliverfile` enabling it by default.

#### Pseudocode

I'm going to show you pseudocode of the uploading logic here so that you can see the differences.

**Current** - It deletes existing screenshots on App Store Connect only when `overwrite_screenshots` option is enabled. The thing is that this won't recognise what changes a fastlane user intend. It's going to only upload when a new screenshot is found so that it could produce unwanted result that mixes both old and new ones. With `overwrite_screenshots` enabled, this is going to consume so many API calls even for just a single screenshot replacement.

```ruby
if optons[:overwrite_screenshots]
  delete_all()
end

for local_screenshot in local_screenshots 
  next if screenshot_exists_on_app_store_connect?(local_screenshot)
  upload(local_screenshot) 
end 
```

**New** - In contrary, this always tries to figure out what is missing on App Store Connect and local so that it can keep App Store Connect in sync with local screenshots. In other words, local screenshots directory is the source of truth. This can save the number of API calls a lot when the update is partial; i.e. only updating screenshots of a single language.

```ruby
# Exlude screenshots from the collection of `Spaceship::ConnectAPI::AppScreenshots` objects, which don't exist on local
screenshots_to_delete = remote_screenshots - local_screenshots

for remote_screenshot in screenshots_to_upload
  delete(remote_screenshot) 
end

# Exclud screenshots from the collection of `Deliver::AppScreenshot` objects, which don't exist on App Store Connect 
screenshots_to_upload = local_screenshots - remote_screenshots

for local_screenshot in screenshots_to_upload
  upload(local_screenshot) 
end
```

#### How comparison works

In pseudocode and actual code, I used `Array#-` to work out `screenshots_to_delete` and `screenshots_to_upload`. But how does that work? `local_screenshots` is an array of `Deliver::AppScreenshot` and `remote_screenshots` is an array of `Spaceship::ConnectAPI::AppScreenshot` so those are different classes that can't be compared directly.

The secret is that I implemented an adaptor/wrapper class that makes it possible to compare those different types via `Array#-`. According to the document, `Array#-` refers to `hash` and `eql?` methods to compare elements. 

> ary - other_ary → new_ary
> Array Difference
> Returns a new array that is a copy of the original array, removing any items that also appear in other_ary. The order is preserved from the original array.
> It compares elements using their hash and eql? methods for efficiency.
>
> [ 1, 1, 2, 2, 3, 3, 4, 5 ] - [ 1, 2, 4 ]  #=>  [ 3, 3, 5 ]

https://docs.ruby-lang.org/en/2.5.0/Array.html#method-i-2D

So I defined `ScreenshotComparable` class and a key that is unique across all of the screenshots no matter which side they are on. It consists of `locale`,  `filename`, and `checksum`. We can get those values from both App Store Connect and local screenshots in the fastlane directory. If a key matches between the one on App Store Connect and the other on local, both are identical and no need to delete or upload.

#### Beta feature/behaviour

Since the new logic changes the existing behaviour, it can break existing users' workflow. I added a new option to `deliver` called `sync_screenshots`. To make it work actually, you also need to set `true` to `FASTLANE_DELIVER_BETA_SYNC_SCREENSHOTS` environment variable in order to make the use of beta features "opt-in" by `FastlaneCore::Feature`'s feature toggle. This should protect users from unintentional usage of a beta feature.

The aim is that after merging this we can then release the new version, call beta testers of this logic from fastlane users on a "Discussion" page and hopefully get feedback there to improve.

#### API Rate Limit

API throttling is a thing with App Store Connect API key. I see 3600 reqs/hour on HTTP headers so far. This means that `deliver` can easily hit the limit with `overwrite_screenshots` option.

* 4 screen sizes (6.5 iPhone, 5.5 iPhone, 12.9 iPad, 12.9 Pro (2nd gen), 12.9 iPad Pro (3rd gen or later)) * 22 languages * 6 images = 528 images!
* A screenshot consumes two API calls to be uploaded; `post https://api.appstoreconnect.apple.com/v1/appScreenshots` and `patch https://api.appstoreconnect.apple.com/v1/appScreenshots/:id`
* There are other types of requests to each `app_screenshot_sets` to sort screenshots or get uploading statuses...

With the situation like above but with 510 images on my own, roughly it takes 1318 calls to complete with the new logic. This won't be different even with the previous logic when it comes to uploading everything but it will be significantly different with partial updates.

#### Rooms to improve

There are still some areas that I have in mind to improve further.

* Deletion of screenshots can be optimised with using `Spaceship::ConnectAPI::AppScreenshotSet#delete!` when applicable
* No matter what API rate limit is in place,`deliver` should be crash-safe when used correctly but somehow it crashes when hitting API rate limit (I need to investigate why this happens) 
* We may want something like `clear_all_screenshots` option for CLI just to clear data and not to upload for making tests easier? 😂 

### Testing Steps

```ruby
gem 'fastlane', git: 'https://github.com/ainame/fastlane.git', branch: 'ainame:ai/better-sync-logic'
```

```ruby
ENV['FASTLANE_ENABLE_BETA_DELIVER_SYNC_SCREENSHOTS'] = true

upload_to_app_store(
  skip_upload_binary: true,
  skip_metadata: true,
  sync_screenshots: true
)
```

You can try the above with various screenshot update scenarios.


KrauseFx:
This is such a great idea. I remember the reason the `overwrite_screenshots` option was added and set to be `false` by default to prevent any existing setups/flows to break. This is something to keep in mind today as well: how we can smoothly migrate existing users, without breaking their screenshot setup? e.g. what if they don't have the `overwrite_screenshots` option set? 

I agree with you, the new approach you built here makes way more sense, and is something that would improve the overall experience by a lot. I'm just thinking if there are any cases we didn't think about yet, and where we could at least offer a heads-up on the change coming, asking the user to explicitly set or unset a flag, or alike. I don't have GH notifications turned on anywhere, but if you'd like to get my input on anything around that, DM me on Twitter any time 💯

ainame:
@KrauseFx Thank you for your early feedback!! This is still WIP and I don't intend to break existing users's workflow at all🙂 

> I'm just thinking if there are any cases we didn't think about yet, where we could at least offer a heads-up on the change coming, asking the user to explicitly set or unset a flag, or alike. 

This is exactly what I have in mind but haven't done anything just yet. A minimal approach I thought in this PR is that introducing a flag to enable the new approach, which is prefixed with "beta" (or alpha?) that should make beginners think this is not a solid feature yet. And then I will call testers to get feedback from them in "Discussions". We can then iterate improvements of this over based on the feedback from the testers. And I agree with you too. This change needs a bit more inline guides that other fastlane features do; e.g. proper messaging at beginning, y/n prompt, and so on...  

In future, when we reach the point where users are happy with new approach in most cases, we might want to make it the default behaviour in the next major version of the fastlane with friendly migration guides in the way fastlane users don't lose their data or break their workflow hopefully.

KrauseFx:
Yess, beautiful, opt-in for this new approach, and maybe having it on by default for new `fastlane init` setups sounds like a great solution 👍 And then potentially printing about the new option as part of the output, educating folks about it

ainame:
Unit tests were added. I will update the PR description with a plan to roll this out and add some user-friendly logging, prompt or something according to the above feedback.

ainame:
I realised that we can use `FastlaneCore::Feature` to roll out a beta feature so renamed `beta_sync_screenshots` to `sync_screenshots` without `beta` prefix and make the new logic require `FASTLANE_ENABLE_BETA_SYNC_SCREENSHOTS`.

joshdholtz:
@googlebot I consent.

joshdholtz (changes requested):
This looks really good! The one issue I found out was that all of my iPad screenshots moved into the 2nd gen 😬 

It seems like we may need to update this conflict resolution method 👇 
https://github.com/fastlane/fastlane/blob/e874a47c6e2e0e61590a03d3b71e75e5a505d1ce/deliver/lib/deliver/app_screenshot.rb#L323-L337

Screen Shot 2021-07-30 at 3 57 27 PM
Screen Shot 2021-07-30 at 3 57 33 PM

I think we need to add `3GEN` to the list? 🤔 I think this is a me issue because I ran `fastlane deliver download_screenshots` and that is the name it downloaded it as 🤷‍♂️ 

Thoughts?

ainame:
@joshdholtz Thank you for adding the change! After reading your comment, your change makes sense to me

joshdholtz (approved):
This PR is straight 🔥 Thank you for adding this! 

I have a radar open asking for the App Store Connect API to either keep the source file or source file MD5 but it hasn't gotten any attention 😔 I believe if that was addressed this PR wouldn't be needed but I here we are 😉 

But this new synchronization experience is soooo nice! 

I'm thinking we should dedicate either a new section or new page in the docs for this. Would you like me to do that? 😊

joshdholtz:
@ainame Also, feel free to smash that merge button when you are ready! This is good to merge from me 💯

ainame:
Thanks @joshdholtz! I think I will write up a discussion ticket to call testers. Within that I'm going to describe how this works differently compared to the current behaviour. And once we get ready to roll this out to public users, we should update the docs! I'll work on that part soon🙂

ainame:
Created this ticket. I'm going to merge this 🚀  https://github.com/fastlane/fastlane/discussions/19204

Issue #19140, opened 2021-07-23
[spaceship] Creation of in-app purchases returns NotFound error


philippmachs:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Creation of in-app purchases via Spaceship::Tunes client returns Not Found error even after the fix introduced in 2.188.0

##### Command executed

  <pre>
APPLE_USER=*** APPLE_PASS=*** APPLE_ID=*** VERBOSE=true bundle exec irb 
require 'spaceship'
Spaceship::Tunes::login(ENV['APPLE_USER'],ENV['APPLE_PASS'])
Spaceship::Tunes::Application.find(ENV['APPLE_ID']).in_app_purchases.create!({"type":"nonConsumable","versions":{"en-US":{"name":"test name","description":"test desctiption"}},"reference_name":"test refname","product_id":"1230987","review_notes":"","pricing_intervals":[{"country":"WW","tier":550}]})
</pre>

##### Complete output when running fastlane, including the stack trace and command used

<details>

  <pre> 
# INFO  [12:39:32]: >> GET ra/apps/manageyourapps/summary/v2: [undefined body]
# DEBUG [12:39:33]: << GET ra/apps/manageyourapps/summary/v2: 200 {...}
# INFO  [12:39:40]: >> GET ra/apps/***/iaps/nonConsumable/template: [undefined body]
# DEBUG [12:39:41]: << GET ra/apps/***/iaps/nonConsumable/template: 200 {...}
# INFO  [12:39:41]: >> POST ra/apps/***/iaps:
# DEBUG [12:39:41]: << POST ra/apps/***/iaps: 404 {"data"=>nil, "messages"=>{"warn"=>nil, "error"=>["Not Found"], "info"=>nil}, "statusCode"=>"ERROR"}
# Traceback (most recent call last):
#        16: from .../ruby-2.7.0/lib/ruby/2.7.0/bundler/vendor/thor/lib/thor/base.rb:476:in `start'
#        15: from .../ruby-2.7.0/lib/ruby/2.7.0/bundler/cli.rb:30:in `dispatch'
[...]

PR #19145, opened 2021-07-24
[spaceship] Add x-csrf-itc header to requests

petersonbill64:
### Checklist

### Motivation and Context

Fixes issues #19124 #19140

### Description

Adds a CSRF header to all requests in the tunes client. I discovered that this header was necessary by encountering the issue from #19140 (even on 2.188.0), doing a similar request via the App Store Connect UI, then diffing the requests, and binary searching through the headers to find the one that was missing for the request to succeed.

I verified that requests to create families, create iaps, update families, and update iaps fail without the header but succeed with it.

### DIFF
--- a/spaceship/lib/spaceship/client.rb
+++ b/spaceship/lib/spaceship/client.rb
@@ -41,6 +41,7 @@ module Spaceship
     attr_accessor :logger
 
     attr_accessor :csrf_tokens
+    attr_accessor :additional_headers
 
     attr_accessor :provider
 
@@ -717,8 +718,13 @@ module Spaceship
       @csrf_tokens || {}
     end
 
+    def additional_headers
+      @additional_headers || {}
+    end
+
     def request(method, url_or_path = nil, params = nil, headers = {}, auto_paginate = false, &block)
       headers.merge!(csrf_tokens)
+      headers.merge!(additional_headers)
       headers['User-Agent'] = USER_AGENT
 
       # Before encoding the parameters, log them
--- a/spaceship/lib/spaceship/connect_api/testflight/client.rb
+++ b/spaceship/lib/spaceship/connect_api/testflight/client.rb
@@ -11,6 +11,9 @@ module Spaceship
 
           super(cookie: cookie, current_team_id: current_team_id, token: token, another_client: another_client)
 
+          # Used by most iris requests starting in July 2021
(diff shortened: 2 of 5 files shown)


torgefaehrlich (approved):
Had the same problem, applied the changes locally, solved it!

Mezmeraid (approved):
I also had this problem with iap edit, fixed with this ! thanks

DagrutCr:
From my tests, it seems that this header is also required for `get_shared_secret` and `generate_shared_secret` in this same file, it doesn't work without it.

joshdholtz (changes requested):
Since it looks like almost every request needs this, thoughts on moving this into 👉  https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L720-L722

Seems like this might make things a little bit cleaner and easier to maintain 🤷‍♂️

joshdholtz:
@petersonbill64 Are you cool if I add a commit to make this change? 😇

hziKen:
I would like to ask whether the master has updated this code?

joshdholtz:
@googlebot I consent.

joshdholtz:
@petersonbill64 Added a commit to remove some of the duplication needed for this new header! I also looked and saw there was one similar for App Store Connect endpoints using Apple ID auth 🤷‍♂️ 

Do you mind taking a look at this to see if this still works for you? I did some testing and I'm pretty sure it does but just want to make sure 😇

petersonbill64:
Thanks @joshdholtz, I tested this with my use case and it's still working correctly. I'm waiting on legal to get back on the CLA; if there's some urgency and you want to open a different PR as the owner to get it merged quickly I'm fine with that, especially since you ended up rewriting it anyway.

petersonbill64:
@googlebot I signed it!

petersonbill64:
@joshdholtz Should be good to go once you dismiss the requested changes. Thanks!

hziKen:
[fastlane/fastlane] [spaceship] Creation of in-app purchases returns NotFound error (#19140) this question not solve man

hziKen:
[fastlane/fastlane] [spaceship] Creation of in-app purchases returns NotFound error (#19140) this question not solve man

joshdholtz (approved):
Thanks so much for find this fix! Really appreciate the contribution ❤️

PR #19149, opened 2021-07-26
[zip] Add `:include` and `:exclude` config options

liamnichols:
### Checklist

### Motivation and Context

I work with a project where various aspects of the project workspace are actually generated via a bootstrap command (i.e the .xcodeproj with XcodeGen) and as a result, its not possible to clone the repo and simply open a project to build and run without having a series of dependencies configured. While this is fine, I recently decided that i would be good to attach the generated workspace for each release as an artefact so that we can inspect the exact source code and even build and run it should we need to. To enable this, I need to use the `zip` action however I don't want to include some directories (`.git`, `derivedData/` etc) and I noticed that the `zip` action does not yet support this. 

Instead of running the `sh` action directory, I thought it might be beneficial to add support to the action here 🙂 

### Description

In this series of commits, I do a few things:

1. Define a `shell_command` matcher - ffcaf53
    - This helps with the next change since the action will now invoke `sh` with an array of unescaped arguments instead of the escaped string.
    - It's functionally the same but I didn't want to cause a messy diff in **zip_spec.rb** which might have let some regressions slip by.
2. Restructure `Fastlane::Actions::ZipAction` - 05c5540
    - Define an internal `Runner` class to break the logic out a bit better
    - Build the shell command up as an array rather than a single string
3. Define our own fixtures - e5e2821
    - Previously we relied on a single file that was defined for other tests.
    - Future tests will need a directory and a couple of files so its easier to define these in a fixture dir called **zip**
4. Add a test for archiving a directory - e7273f6
    - It wasn't already included in the tests so now we have the new fixture directory, it makes sense to add one
5. Add `:include` and `:exclude` config options - 2cc4721
    - This was relatively easy to add after the refactoring above
    - The array of patterns/files are mapped and applied with the `-i` and `-x` flags (see `man zip`)
6. Fixed issue with using an un-expanded path for `:path` - 98787f6
    - `File.exist?` needs the expanded path so my test using `~/Desktop` didn't work
7. Remove the custom `shell_command` matcher - 4c2383f 
    - This matcher only really added value during refactoring so i've updated the specs again to match against the array of arguments directly

I'll add some comments inline, but i recommend filtering through the commits for easier to process diffs 🙏 

### Testing Steps

The specs cover most things here, but if you want to test then you can setup a test lane or run the action directly to compare the results


joshdholtz (approved):
This looks sooooo good! Thank you for making these changes and cleaning up this code 🙌  Really appreciate it ❤️

liamnichols:
No, thank **you**! Have a good day 🚀

PR #19175, opened 2021-07-29
[action][app_store_connect_api_key] Change default token duration

DavidBrunow:
### Checklist

### Motivation and Context

This change fixes a problem detailed in this discussion: https://github.com/fastlane/fastlane/discussions/18100

To summarize, folks are running into issues uploading their builds to App Store Connect using the API key because their machine's time is not the same as the time on Apple servers. Since the default value for this action is the same as the maximum duration that Apple's API will allow, this causes uploads to fail with cryptic messages that do not clearly point folks to the root cause. 

### Description

I've lowered the default value so that folks with times that are not the same as Apple's servers should not run into this issue when using the default settings. This should reduce the number of people that run into this issue but will not eliminate the issue. If we wanted to do that one possible way to do so would be to check the system time against a standard time source to ensure it is within some acceptable range.

### Testing Steps

Ran tests.

### DIFF
--- a/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb
+++ b/fastlane/lib/fastlane/actions/app_store_connect_api_key.rb
@@ -80,7 +80,7 @@ module Fastlane
                                        env_name: "APP_STORE_CONNECT_API_KEY_DURATION",
                                        description: "The token session duration",
                                        optional: true,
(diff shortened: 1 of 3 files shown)


joshdholtz (approved):
This makes sense to me! I had no idea setting at max was going to cause errors 😢 Really appreciate the contribution! ❤️

PR #19212, opened 2021-08-06
[spaceship] Adding missing method to fetch build IDs for beta group

gsavit:
### Checklist

### Motivation and Context

Implementing the following missing TestFlight/Beta Groups endpoint to fetch build IDs for a beta group: https://developer.apple.com/documentation/appstoreconnectapi/get_all_build_ids_in_a_beta_group

### DIFF
--- a/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
+++ b/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
@@ -251,9 +251,7 @@ module Spaceship
           test_flight_request_client.get("betaGroups/#{group_id}/builds")
         end
 
-        def get_build_ids_for_beta_group(group_id: nil)
-          raise "group_id is nil" if group_id.nil?
-
+        def get_build_ids_for_beta_group(group_id:)
           test_flight_request_client.get("betaGroups/#{group_id}/relationships/builds", {})
         end
 
--- a/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
+++ b/spaceship/lib/spaceship/connect_api/testflight/testflight.rb
@@ -251,6 +251,12 @@ module Spaceship
           test_flight_request_client.get("betaGroups/#{group_id}/builds")
         end
 
+        def get_build_ids_for_beta_group(group_id: nil)
+          raise "group_id is nil" if group_id.nil?
+
+          test_flight_request_client.get("betaGroups/#{group_id}/relationships/builds", {})
+        end
+
         #
         # betaTesters
         #
--- a/spaceship/spec/connect_api/testflight/testflight_client_spec.rb
+++ b/spaceship/spec/connect_api/testflight/testflight_client_spec.rb
@@ -421,6 +421,20 @@ describe Spaceship::ConnectAPI::TestFlight::Client do
         end
       end
 
+      context 'get_build_ids_for_beta_group' do
+        let(:path) { "betaGroups" }
+        let(:beta_group_id) { "123" }
+
+        it 'succeeds' do
+          url = "#{path}/#{beta_group_id}/relationships/builds"
+          params = {}
+          req_mock = test_request_params(url, params)
+
+          expect(client).to receive(:request).with(:get).and_yield(req_mock).and_return(req_mock)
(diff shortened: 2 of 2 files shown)


spaceship/lib/spaceship/connect_api/testflight/testflight.rb
```diff
@@ -251,6 +251,12 @@ def get_builds_for_beta_group(group_id: nil)
           test_flight_request_client.get("betaGroups/#{group_id}/builds")
         end
 
+        def get_build_ids_for_beta_group(group_id: nil)
+          raise "group_id is nil" if group_id.nil?
```
joshdholtz, line 255:
```suggestion
        def get_builds_for_beta_group(group_id:)
```

spaceship/spec/connect_api/testflight/testflight_client_spec.rb
```diff
@@ -421,6 +421,20 @@ def test_request_body(url, body)
         end
       end
 
+      context 'get_build_ids_for_beta_group' do
```
joshdholtz, line 424:
```suggestion
      context 'get_builds_for_beta_group' do
```

joshdholtz (changes requested):
This looks good! Just wanted to rename this from `build_ids` to `builds` and make the `group_id:` argument required 😇 

Other than that this is 💯

spaceship/lib/spaceship/connect_api/testflight/testflight.rb
gsavit: hmm the thing is there's a separate method that fetches builds on beta groups. this one is specifically fetching just build ids. see the method just about this one 😁

gsavit: but, the required arg makes sense!!

joshdholtz: Ha! To be fair... my eyes were dilated when I reviewed this PR so I was 100% not seeing clearly 😛

joshdholtz (commented):
@gsavit I'm slightly torn about this PR 🤔 

Is there any reason why the existing   `get_builds_for_beta_group(group_id: group_is).map(&:id)` couldn't be used instead?

The issue I'm seeing here is that `betaGroups/#{group_id}/relationships/builds` returns something like...

```json
{
  "data" : [ {
    "type" : "builds",
    "id" : "6ac6a092-94ba-4097-bfb6-67d7b38f3236"
  }, {
    "type" : "builds",
    "id" : "be17adad-40b9-4bff-856c-ec2fd9c843f1"
  }, {
    "type" : "builds",
    "id" : "66242589-d099-42c9-9994-4a5ff7f20059"
  }, {
    "type" : "builds",
    "id" : "0a324af5-0486-4e7e-86b4-e1234687e75f"
  }, {
    "type" : "builds",
    "id" : "c4613b60-28fd-4291-9b33-a6e1eacfc9e5"
```

which will end up getting transformed into a `Spaceship::ConnectAPI::Build` model and some a lot of those attributes on it will be empty which might get confusing 😬 

So its not actually returning an array of build ids but partial models 🤷‍♂️ 

Thoughts on my thoughts? 😇

gsavit:
> Is there any reason why the existing `get_builds_for_beta_group(group_id: group_is).map(&:id)` couldn't be used instead?

So, we were using that call initially and experienced pretty significant performance/slowness issues (very heavy TestFlight presence). We tried calling this other endpoint instead, and it was noticeably quicker. No formal benchmarking, happy to do more and share here if you prefer?

> So its not actually returning an array of build ids but partial models

This didn't even occur to me. I suppose the naming becomes a bit imprecise 🙈 Hmm...

joshdholtz:
@gsavit Ahhh, yeah yeah, it makes sense that doing the other way would be slower! I'm not _really_ sure this new method belongs in the way that it's currently implemented. 

I think we either need to think of a new way to do this so it doesn't create an array of partial models or... maybe you can continue just hitting this endpoint directly in your implementation? 🤷‍♂️

gsavit:
@joshdholtz that's fair enough! We'll try hitting the endpoint directly for now, will close this and revisit later on if needed. Thanks for the consideration!

PR #19232, opened 2021-08-13
[fastlane] optimize bash and zsh completion scripts

mcornella:
### Checklist

I had issues with `bundle exec rspec`, probably due to a broken installation since it's the first time I've had to use it.

### Motivation and Context

The `_fastlane_complete` function for bash and zsh can be improved:

- It makes 3 calls to get lane names, which can be grouped into a single `sed` call.
- It leaks the `file` variable.
- It continues execution even if no `Fastfile` file has been found.

### Description

Changes:

- Group the 3 calls to `grep` and `awk` into a single `sed` call with [-E for extended regex (POSIX-compatible argument)](https://www.gnu.org/software/sed/manual/sed.html#index-_002dE).
- Makes `$file` variable local.
- Exit the function when no Fastfile file was found.

### Testing Steps

1. Add `compctl -K _fastlane_complete fastlane` to `completion.zsh` file.
2. Add `complete -F _fastlane_complete fastlane` to `completion.bash` file.
3. cd into `fastlane` repository root directory.
4. `source path/to/completion.{bash,zsh}` file depending on shell.
5. Try to complete `fastlane` command.

The same options should appear with and without these changes, while only making a single `sed` call.

### Others

Explanation of `sed` call:

- `-E`: enables extended regexp
- `-n`: disables output unless specifically enabled with `/p`.
- `s/^\s*lane\s+:(\S+).*$/\1/p`: matches whole lines that have a `lane :<lane_name> ...` format and prints only `<lane_name>` ([explanation of RegExp strings](https://www.gnu.org/software/sed/manual/sed.html#ERE-syntax)).

### DIFF
--- a/fastlane/lib/assets/completions/completion.bash
+++ b/fastlane/lib/assets/completions/completion.bash
@@ -18,7 +18,7 @@ _fastlane_complete() {
   fi
 
   # parse 'beta' out of 'lane :beta do', etc
-  completions="$(sed -En 's/^[ 	]*lane +:([^ ]+).*$/\1/p' "$file")"
+  completions="$(sed -En 's/^[ 	]*lane +:([^ 	]+).*$/\1/p' "$file")"
   completions="$completions update_fastlane"
 
   COMPREPLY=( $(compgen -W "$completions" -- "$word") )
--- a/fastlane/lib/assets/completions/completion.bash
+++ b/fastlane/lib/assets/completions/completion.bash
@@ -18,7 +18,7 @@ _fastlane_complete() {
   fi
 
   # parse 'beta' out of 'lane :beta do', etc
-  completions="$(sed -En 's/^\s*lane\s+:(\S+).*$/\1/p' "$file")"
+  completions="$(sed -En 's/^[ 	]*lane +:([^ ]+).*$/\1/p' "$file")"
   completions="$completions update_fastlane"
 
   COMPREPLY=( $(compgen -W "$completions" -- "$word") )
--- a/fastlane/lib/assets/completions/completion.bash
+++ b/fastlane/lib/assets/completions/completion.bash
@@ -4,6 +4,7 @@ _fastlane_complete() {
   COMPREPLY=()
   local word="${COMP_WORDS[COMP_CWORD]}"
   local completions=""
+  local file
 
   # look for Fastfile either in this directory or fastlane/ then grab the lane names
   if [[ -e "Fastfile" ]]; then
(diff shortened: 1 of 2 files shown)


mcornella:
@googlebot I signed it!

joshdholtz (commented):
@mcornella This looks like a great change! I've actually never used this completions before (🤦‍♂️) but I'm getting an issue where its only coming up with `update_fastlane` 😬 Do you have any idea why this would be?

mcornella:
> I'm getting an issue where its only coming up with `update_fastlane` 😬 Do you have any idea why this would be? 

Hmm it has to be an issue with the sed command using some syntax that's not truly portable. What platform/version are you running, and what are the Fastfile contents?

joshdholtz:
> > I'm getting an issue where its only coming up with `update_fastlane` 😬 Do you have any idea why this would be?
> 
> Hmm it has to be an issue with the sed command using some syntax that's not truly portable. What platform/version are you running, and what are the Fastfile contents?

I'm running on macOS (tried on both Big Sur and Monterey).  Here is a sample `Fastfile` 👇 

```ruby
lane :test_ios do
  scan(
    scheme: "TestAllPlatforms (iOS)",
    devices: ["iPhone 12"]
  )
end

lane :test_tvos do
  scan(
    scheme: "TestAllPlatformsTV",
    devices: ["Apple TV"]
  )
end

lane :test_both do
  scan(
    scheme: "TestAllPlatformsTV",
    devices: ["iPhone 12", "Apple TV"]
  )
end
```

Is there a good way for me to debug this?

mcornella:
A fast way to iterate on the sed command is to run it directly. Does this one work?
```sh
sed -n 's/^\s*lane\s\+\:\(\S\+\).*$/\1/p' Fastfile
```

joshdholtz:
> A fast way to iterate on the sed command is to run it directly. Does this one work?
> 
> ```shell
> sed -n 's/^\s*lane\s\+\:\(\S\+\).*$/\1/p' Fastfile
> ```

Nope nope, that doesn't output anything for me 😱

mcornella:
Huh, it looks like its a lack of support for extended character groups, this seems to work for me:
```
sed -En 's/^ *lane +:([^ ]+).*$/\1/p' Fastfile
```

joshdholtz:
> Huh, it looks like its a lack of support for extended character groups, this seems to work for me:
> 
> ```shell
> sed -En 's/^ *lane +:([^ ]+).*$/\1/p' Fastfile
> ```
> 
> EDIT: we should also support tab-indentation, so this is better:
> 
> ```shell
> sed -En 's/^[ 	]*lane +:([^ ]+).*$/\1/p' Fastfile
> ```

Looks like that command works for me now 🙌 

```sh
➜  TestAllPlatforms git:(main) ✗ sed -En 's/^[  ]*lane +:([^ ]+).*$/\1/p' fastlane/Fastfile 
test_ios
test_tvos
test_both
```

mcornella:
This is ready to merge I think.

joshdholtz (approved):
This looks great to me! Thank you so much for fixing this ❤️

PR #19285, opened 2021-08-25
[match] Don't add key if cert already present

### LINKED ISSUE #19268: Bug in `security import` adding random private keys to keychain
### New Issue Checklist

- [ ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I have an automated loop that runs `Match::Runner.new.run` to update the certs at a fixed interval. There is a case where a cert is already present (added in the previous loop) and in that case, Fastlane detects the cert is already present but it tries to add the private key anyway. Code: https://github.com/fastlane/fastlane/blob/master/match/lib/match/runner.rb#L199

Internally Fastlane runs the `security import` command and that errors out with `SecKeychainItemImport: The specified item already exists in the keychain` but for some reason, there is a random 16 char key added to my keychain, exactly like this: https://github.com/fastlane/fastlane/issues/12291

My hunch is, this is a bug with the `security import` command and the random private key we see is a placeholder that the import command fails to delete after error. As my loop keys trying, the keychain ends up with thousands of such random private keys and eventually starts hanging.

 Would like to know why we are trying to add a key in the current code when we know the cert is already present. I think we should not try to import the key when the cert is already present or at least check if the key is already present before importing it.

Will try to raise a PR with these changes.

##### Command executed
`Match::Runner.new.run`

SagarGaniga:
### Checklist

### Motivation and Context
Resolves: https://github.com/fastlane/fastlane/issues/19268

### Description
Avoid adding a private key to the keychain if the cert is already present while doing `fetch_certificate`. The issue is with `security import` which adds a random private key if this key is already present.

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -191,12 +191,12 @@ module Match
             UI.verbose("Certificate '#{File.basename(cert_path)}' is already installed on this machine")
           else
             Utils.import(cert_path, params[:keychain_name], password: params[:keychain_password])
-          end
 
-          # Import the private key
-          # there seems to be no good way to check if it's already installed - so just install it
-          # Key will only be added to the partition list if it isn't already installed
-          Utils.import(keys.last, params[:keychain_name], password: params[:keychain_password])
+            # Import the private key
+            # there seems to be no good way to check if it's already installed - so just install it
+            # Key will only be added to the partition list if it isn't already installed
+            Utils.import(keys.last, params[:keychain_name], password: params[:keychain_password])
+          end
         else
           UI.message("Skipping installation of certificate as it would not work on this operating system.")
         end


SagarGaniga:
@googlebot I signed it!

joshdholtz (commented):
@SagarGaniga Really appreciate you looking into this! I think this makes sense... 🤨

The reason that this is separate is the cert and private key are two different things. So there are situations where the cert may be installed by the private key may not and the private key is really the important thing for code signing.

But... I've experience a lot of these random private keys and I would REALLY prefer that they don't get added 😅

I'm going to do a bit of testing on this myself right now but I _THHHIIINNNNKKKK_ this will be good to merge 👍

joshdholtz (approved):
Oh yeah, this is good! I don't know why this wasn't done sooner 😅 Thank you!

PR #19296, opened 2021-08-26
[pilot][upload_to_testflight] add Mac support

joshdholtz:
### Motivation and Context
Apple opened up TestFlight to Mac apps on August 24, 2021

### Description
- Pilot
  - Added new option for `pkg`
  - Removed `ios` as default option on `app_platform` option
- BuildWatcher
  - Turns out that builds have multiple processing states (1 on builds and 2 on buildBetaDetails)
  - The processings are independent of each other but **should** finish at the same time
  - Previously we only wait for the build processing but Mac builds are slower on the buildBetaDetails processing currently
  - Added an option (that pilot uses) to wait for all 3 processing status
- Spaceship
  - Added new helpder function on buildBetaDetails called `processing?` that will return true when internal and external states are done processing

### Testing Steps

```rb
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "joshdholtz-pilot-for-mac"
```

### DIFF
--- a/fastlane/lib/fastlane/actions/upload_to_testflight.rb
+++ b/fastlane/lib/fastlane/actions/upload_to_testflight.rb
@@ -13,6 +13,8 @@ module Fastlane
         unless distribute_only
           values[:ipa] ||= Actions.lane_context[SharedValues::IPA_OUTPUT_PATH]
           values[:ipa] = File.expand_path(values[:ipa]) if values[:ipa]
+          values[:pkg] ||= Actions.lane_context[SharedValues::PKG_OUTPUT_PATH]
+          values[:pkg] = File.expand_path(values[:pkg]) if values[:pkg]
         end
 
         # Only set :api_key from SharedValues if :api_key_path isn't set (conflicting options)
@@ -117,7 +119,7 @@ module Fastlane
       end
 
       def self.is_supported?(platform)
-        [:ios].include?(platform)
+        [:ios, :mac].include?(platform)
       end
     end
   end
--- a/fastlane/lib/fastlane/actions/upload_to_testflight.rb
+++ b/fastlane/lib/fastlane/actions/upload_to_testflight.rb
@@ -119,7 +119,7 @@ module Fastlane
       end
 
       def self.is_supported?(platform)
-        [:ios, :mac].include?(platform)
+        [:ios, :mac, :tvos].include?(platform)
       end
     end
   end
--- a/fastlane_core/lib/fastlane_core/build_watcher.rb
+++ b/fastlane_core/lib/fastlane_core/build_watcher.rb
@@ -11,7 +11,7 @@ module FastlaneCore
 
     class << self
       # @return The build we waited for. This method will always return a build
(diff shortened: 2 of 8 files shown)


crazymanish (approved):
This looks very good to me 🚀 also those pkg unit tests in BuildManager ❤️ 🔥

lucgrabowski (approved):

idoodler:
Hi, thanks for that feature!

I am trying to implement this, the upload works 👍 

However I am unable to select the uploaded build in AppStore Connect.

I watched the WWDC session regarding the TestFlight and macOS. They mention, that we must manually create a special provisioning profile if we don't use automatic codesigning.

I am building an Electron app using electron-builder, thus I use manual codesigning. I am unable to find any information on what provisioning profile is necessary.

HI also asked for help in the [Apple Developer Forums](https://developer.apple.com/forums/thread/690045)

Maybe someone here can point me in the right direction.

thx

uglide:
Hi @idoodler 
Have you found a solution? I faced the same problem and cannot figure out how to fix "Not Available for Testing" status in AppStore Connect.

idoodler:
> Hi @idoodler 
> 
> Have you found a solution? I faced the same problem and cannot figure out how to fix "Not Available for Testing" status in AppStore Connect.

No, not yet:/

tremblay:
Hi, I'm also having trouble with this feature. It seems the `build_app` gives us a `.app` as the output, but `upload_to_testflight` says that the .app "doesn't seem to be a pkg file". Could you give any advice about what settings I may have mixed up?

For reference, here are my calls to `build_app` and `upload_to_testflight` (I replaced my actual app name with `Name`):

```
def build_mac(configuration, scheme)
  build_app(
    configuration: configuration,
    export_method: "mac-application",
    scheme: scheme,
    include_symbols: true,
    include_bitcode: true,
    export_options: {
      provisioningProfiles: {
        "app.Name" => "Mac AppStore",
        "app.Name.AutoFill" => "Mac AutoFill",
        "app.Name.Safari" => "Mac Safari"
      }
    }
  )
end
```

```
upload_to_testflight(
  pkg: "./Name.app",
  app_platform: "osx",
  distribute_external: true,
  groups: ['External Testers'],
  changelog: options[:changelog]
)
```

uglide:
@idoodler I'm packaging Qt app. I found a solution how to successfully upload macOS App to TestFlight. The following attributes should be added to `Entitlements.plist`:

```xml
        <key>com.apple.application-identifier</key>
	<string>XXXXXXX.com.sample.app</string>
	<string>XXXXXXX</string>
```

In addition, you should supply the app with `embedded.provisionprofile` which is Mac App Store Distribution Profile.

idoodler:
@tremblay I am not using `build_app` but `electron-builder` instead

idoodler:
@uglide Thanks, I will try that with todays alpha build

idoodler:
@tremblay I just checked the [documentation](https://docs.fastlane.tools/actions/build_mac_app/) of fastlane. You may want to use a different `export_method`, something like `app_store` or `package`

idoodler:
@uglide No, this did not the trick for me. It still states that the build is not available for TestFlight testing.

idoodler:
@uglide I was just checking the `embedded.provisionprofile`. The entitlements entry has the keys and values you suggest. Did you recreate such a profile, or did you simply use your already existing one?

uglide:
@idoodler new one generated after Testflight release.

PR #19303, opened 2021-08-27
[match] Adds identifiers supplied to match to the git commit message

mikelrob:
### Checklist

### Motivation and Context
We have multiple teams developing multiple apps. We all use `fastlane` to automate our deployment, it's so great to have a reliable automated process. Recently we had to renew our iOS Distribution certificate which resulted in all teams renewing their respective Distribution provisioning profiles. This was super easy using `fastlane match` however the git repo we use for storage does not have a readable commit log. I wanted to include the identifiers supplied to `match` from the CLI or Matchfile in the commit message to give us better visibility at a glance of our log.

### Description
First I moved the `app_identifiers` block of code in the runner above the storage selection. I want to pass this into the storage object. 
I fixed up the `runner_spec` according to expect these extra params.
Next I added the `identifiers` to the `git_storage` and used them in the `generate_commit_message` func.
I updated the `git_storage_spec` with context block testing for single and multiple identifiers.
Ensuring it works with multiple identifers was important to us as we have an app with linked extensions.
After the changes were made I ran the test suite and rubocop analysis.

### Testing Steps
I ran:
```
bundle exec rspec
bundle exec rubocop -a
```

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -29,6 +29,16 @@ module Match
 
       update_optional_values_depending_on_storage_type(params)
 
+      if params[:app_identifier].kind_of?(Array)
+        app_identifiers = params[:app_identifier]
+      else
+        app_identifiers = params[:app_identifier].to_s.split(/\s*,\s*/).uniq
+      end
+
+      # sometimes we get an array with arrays, this is a bug. To unblock people using match, I suggest we flatten!
+      # then in the future address the root cause of https://github.com/fastlane/fastlane/issues/11324
+      app_identifiers.flatten!
+
       # Choose the right storage and encryption implementations
       self.storage = Storage.for_mode(params[:storage_mode], {
         git_url: params[:git_url],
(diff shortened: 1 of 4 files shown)


joshdholtz (changes requested):
I think this is a great addition! However, I think this could be reworked to use the existing `custom_message` argument on  `def save_changes!(files_to_commit: nil, files_to_delete: nil, custom_message: nil)`. This way you don't need to pass the identifiers into the storage object 😇 

Let me know your thoughts on this!

mikelrob:
> I think this is a great addition! However, I think this could be reworked to use the existing `custom_message` argument on `def save_changes!(files_to_commit: nil, files_to_delete: nil, custom_message: nil)`. This way you don't need to pass the identifiers into the storage object 😇
> 
> Let me know your thoughts on this!

Thanks! Yea, I like your idea of using the existing mechanism. I guess I just need to create an option on the match command to accept the custom message and pass it through?

CaseyCw1:
This works really well for me, thank you! Facing same issue here. Help is appreciated.

getaaron:
@mikelrob hey are you still working on this? we'd love to get it fixed up and merged in 😍

PR #19312, opened 2021-08-31
[gradle] add output context with "tasks" options

davy-tw:
### Checklist

### Motivation and Context

Support filling apk/aab/etc output contexts with `tasks` options in `gradle`/`build_android_app` actions

### Description

Before this change, `gradle`/`build_android_app` detects if `task` option contains `assemble` or `bundle` wording and decides to fill output files in context or not.

With `tasks` option (and without `task`) will cause `gradle`/`build_android_app` stop before searching output files because of empty `task` option.

`gradle` action pass tasks to real gradle with calculated `gradle_task` variable which choose `task` option or string-concatenated-with-spaces `tasks` option, so we can detect this variable instead of `task` option directly.

### Testing Steps

Since original behavior doesn't have spec, we have to test manually.
Simple POC is calling `gradle(tasks: %W(assembleRelease assembleReleaseAndroidTest))` and check if paths of output files in listed in context (e.g. `lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS]`)

### DIFF
--- a/fastlane/lib/fastlane/actions/gradle.rb
+++ b/fastlane/lib/fastlane/actions/gradle.rb
@@ -64,7 +64,7 @@ module Fastlane
                                 print_command_output: params[:print_command_output])
 
         # If we didn't build, then we return now, as it makes no sense to search for apk's in a non-`assemble` or non-`build` scenario
-        return result unless gradle_task =~ /\b(assemble)/ || task =~ /\b(bundle)/
+        return result unless gradle_task =~ /\b(assemble)/ || gradle_task =~ /\b(bundle)/
 
         apk_search_path = File.join(project_dir, '**', 'build', 'outputs', 'apk', '**', '*.apk')
         aab_search_path = File.join(project_dir, '**', 'build', 'outputs', 'bundle', '**', '*.aab')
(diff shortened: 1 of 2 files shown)


david50407:
@googlebot I signed it!

joshdholtz:
@googlebot I consent.

joshdholtz:
@david50407 This is great! I did add one commit to add tests and also fix the some logic so that `bundle` also was using `gradle_task`?

david50407:
@joshdholtz you're right. Thank you :)

joshdholtz (approved):
Thanks for finding this and fixing it! Didn't realize it was like this 🤦‍♂️ Really appreciate the contribution ❤️

flipace:
We just ran into an issue because we depended on `supply` not having `apk_paths` automatically set from the `GRADLE_ALL_APK_OUTPUT_PATHS` (because we use the `tasks` option, in which case this env variable was not set before) -> now that this is working as (likely) intended, we are required to pass the `skip_upload_apk` option to prevent the `Cannot provide both apk(s) and aab - use `skip_upload_apk`, `skip_upload_aab`, or  make sure to remove any existing .apk or .aab files that are no longer needed` error.

Just thought I'd mention it here since this came as a patch release, but actually caused a regression in our use-case.

PR #19340, opened 2021-09-07
[match] Add ability to include all development certificates to the development provisioning profile

nekrich:
### Checklist

### Motivation and Context
We want to automate adding devices and development certificates to the development provisioning profiles.
Our limitations:
* No one (except CI) has access to the match repo.
* We don't want to share the single Apple Development certificate across developers.

So, we become with a solution:
* All devices are listed in the file stored in the project repo. With is perfectly handled by the register_devices.
* Each run we add missing devices to the development provisioning profile.

The thing we were missing is how to automate adding new Apple Development Certificates to the provisioning profiles in the same manner.

### Description
New `include_all_certificates` option in `sigh` & `match`, which will allow us to include all certificates (only for `development` provisionong profile type).
Additionally, introduce a new option to `match` `force_for_new_certificates`, that will work exactly in the same way as the existent option`force_for_new_devices` but for certificates: if the count of development certificates differs (account vs profile) - update the profile.

It still uses `the same account and have one code signing identity` for distribution and access to the App Store Connect and Developer portal as described in the [codesigning.guide](https://codesigning.guide/), but additionally allows the use of multiple identities/certificates for development provisioning profiles (w/o including them in the repo) as an option.

### Testing Steps
```
    sync_code_signing(
      app_identifier: [
        'com.your.bundle.id',
      ],
      type: 'development',
      readonly: false,
      include_all_certificates: true,
      force_for_new_certificates: true,
    )
```


match/lib/match/runner.rb
```diff
@@ -254,6 +254,24 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
         end
       end
 
+      if params[:include_all_certificates]
+        certificate_id = nil
```
joshdholtz, line 258: Can we add a comment above this explaining why its being set to nil? I thought this was initializing something and was confused why it wasn't used down further 😅

```suggestion
        # Clearing specified certificate id which will prevent a profile being created with only one certificate
        certificate_id = nil
```

joshdholtz (changes requested):
This looks great! Just one little suggestion about adding a comment line 😇

nekrich:
> This looks great! Just one little suggestion about adding a comment line 😇
👍
Thanks. Done.

match/lib/match/runner.rb
```diff
@@ -254,6 +254,25 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
...
       end
 
+      if params[:include_all_certificates]
+        # Clearing specified certificate id which will prevent a profile being created with only one certificate
+        certificate_id = nil
+
+        if params[:force_for_new_certificates] && !params[:readonly]
+          if prov_type == :development && !params[:force]
+            force = certificate_count_different?(profile: profile, keychain_path: keychain_path, platform: params[:platform].to_sym)
+          else
```
joshdholtz, line 264: One more question! If `prov_type` is `:development` and `force` is set to `true` already, won't the `"Warning: `force_for_new_certificates` is set but is ignored for non-'development' provisioning profiles."` get printed out?

I think that check for `force` needs to happen somewhere else, correct?

joshdholtz (changes requested):
One more question after testing and reviewing again! After this I think this should be good to merge 😊

match/lib/match/runner.rb
nekrich: Good catch. I think I'll need to wrap it in the `unless force`

nekrich: Fixed. Better to review [w/o space changes](https://github.com/fastlane/fastlane/pull/19340/files?diff=unified&w=1)

nekrich: Moved all checks into separate methods to reduce complexity.

nekrich:
@joshdholtz , appveyor failed to DL some dependency, and there is a strange new CLA request after rebasing. idk 🤷

nekrich:
> Will take a look when I can! Got LASIK a few days ago so eyes are a bit sensitive to light and computer screens still 🤷‍♂️

Get well soon ❤️‍🩹.

joshdholtz:
@googlebot I consent.

techinpark:
@nekrich i need so much this option!!  great work 👍

joshdholtz (approved):
This is so good! Gave this one last test and review today 💯 Thanks for the work on this! This will go out in today's release

Issue #15633, opened 2019-11-11
Gradle properties.versionCode not being used


alexcroox:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I was running into `Google Api Error: apkNotificationMessageKeyUpgradeVersionConflict: APK specifies a version code that has already been used.` and I couldn't work out why as I was using the following:

```
gradle(
      task: "assemble",
      build_type: "Release",
      project_dir: "./android",
      properties: {
        "versionName" => "2.0.0",
        "versionCode" => 200,
        "android.injected.signing.store.file" => "keystore.jks",
        "android.injected.signing.store.password" => ENV["FASTLANE_KEYSTORE_PASSWORD"],
        "android.injected.signing.key.alias" => "upload",
        "android.injected.signing.key.password" => ENV["FASTLANE_KEY_PASSWORD"]
      }
    )
```

and I could see in the output it was being passed in the command:

```
--------------------
[23:22:18]: --- Step: gradle ---
[23:22:18]: --------------------
[23:22:18]: $ /Users/alex/Projects/sml/android/gradlew assembleRelease -p ./android -PversionName=2.0.0 -PversionCode=200 ...
```
However this seems to be completely ignored, and the only value gradle would actually use for the apk sits inside `app/build.gradle`

```
android {
    compileSdkVersion 28
    defaultConfig {
        ...
        versionCode 1 << This will be used no matter what, if I delete this line I get a gradle error
        versionName "1.0"
        ...
    }
    buildTypes {
        release {
            minifyEnabled false
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
        }
    }
}
```

<details>
  <pre>***</pre>
</details>

### Environment

<details>
  <pre><details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14.6                                     |
| Ruby                        | 2.6.3                                       |
| Bundler?                    | true                                        |
| Git                         | git version 2.21.0 (Apple Git-122.2)        |
[...]

PR #19370, opened 2021-09-21
[docs] Improve gradle action

eliaslecomte:
By explaining android.injected.version.code and name can be used to change the apps versionCode and versionName.

### Checklist

### Motivation and Context
A typical use case while using Fastlane is to change the version code / name of your app. While you can do this by passing a gradle propertie and adding code in your build.gradle, its so much easier via the `android.injected.version` notation and doens't require code in build.gradle.
By adding this example to the docs we can help devs to use that.

### Description
The documentation shows an example how you can pass versionName and versionCode as a gradle property while building. To make this work you also need to add code in your build.gradle that checks if the variable is set or not and passes that to androids versionName and versionCode. A much simpler approach is actually to pass the properties `android.injected.version.name` and `android.injected.version.code` as this will modify version code and name without additional code.

There was also this closed issue: https://github.com/fastlane/fastlane/issues/15633 and I've discovered the varaibles first here: https://stackoverflow.com/questions/36213977/is-it-possible-to-set-android-version-code-and-name-via-a-gradle-task/36221899# and then here: https://www.javadoc.io/static/com.android.tools.build/builder-model/2.5.0-alpha-preview-02/constant-values.html

### Testing Steps
I've tested this in a fastlane lane manually and in a ci environment.

### DIFF
--- a/fastlane/lib/fastlane/actions/gradle.rb
+++ b/fastlane/lib/fastlane/actions/gradle.rb
@@ -258,8 +258,21 @@ module Fastlane
             # ...
 
             properties: {
-              "versionCode" => 100,
-              "versionName" => "1.0.0",
+              "exampleNumber" => 100,
+              "exampleString" => "1.0.0",
+              # ...
+            }
+          )
+          ```
+
+          You can use this to change the version code and name of your app:
+          ```ruby
+          gradle(
+            # ...
+
+            properties: {
+              "android.injected.version.code" => 100,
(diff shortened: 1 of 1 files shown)


eliaslecomte:
@googlebot I signed it!

joshdholtz (approved):
I had no idea you could do this! This is an amazing addition 🥳 Thank you ❤️

eliaslecomte:
I was equally happily surprised :-D!

PR #19419, opened 2021-10-01
Fix and ammend gemspec metadata

aried3r:
### Checklist

### Motivation and Context

Was browsing rubygems.org and wanted to see the official docs but got linked to rubydoc.info instead.

### Description

The `docs_url` key is called `documentation_uri`. Right now, because of the wrong key [rubygems.org](https://rubygems.org/gems/fastlane) links to [rubydoc.info](https://www.rubydoc.info/gems/fastlane/2.195.0) instead of the URL previously set in `docs_url`.

The other keys were set to show what could be done to also show up on RubyGems.org. While there is no "mailing list", I think GitHub discussions are the closest thing.

Happy to add, change, remove any attributes that are unwanted. See the [`metadata` docs](https://guides.rubygems.org/specification-reference/#metadata) for the possible keys.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -58,7 +58,6 @@ Gem::Specification.new do |spec|
     "changelog_uri" => "https://github.com/fastlane/fastlane/releases",
     "documentation_uri" => "https://docs.fastlane.tools/",
     "homepage_uri" => spec.homepage,
-    "mailing_list_uri" => "https://github.com/fastlane/fastlane/discussions",
     "source_code_uri" => "https://github.com/fastlane/fastlane"
   }
 
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -54,7 +54,12 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://fastlane.tools"
   spec.license       = "MIT"
   spec.metadata      = {
-    "docs_url" => "https://docs.fastlane.tools"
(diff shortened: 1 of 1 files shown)


fastlane.gemspec
```diff
@@ -54,7 +54,12 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://fastlane.tools"
   spec.license       = "MIT"
   spec.metadata      = {
-    "docs_url" => "https://docs.fastlane.tools"
+    "bug_tracker_uri" => "https://github.com/fastlane/fastlane/issues",
+    "changelog_uri" => "https://github.com/fastlane/fastlane/releases",
+    "documentation_uri" => "https://docs.fastlane.tools/",
+    "homepage_uri" => spec.homepage,
+    "mailing_list_uri" => "https://github.com/fastlane/fastlane/discussions",
```
joshdholtz, line 61: I'd love to have a link to discussions but mailing list seems a little off 🤷‍♂️ I think we are good to ship after removing this one!
```suggestion
```

joshdholtz (changes requested):
This is great! Didn't know this was broken and also all of these exist 🙃 One small request to remove that and then this is good to ship 👍

aried3r:
Good point about the mailing list. I applied the changes. :)

joshdholtz (approved):
This looks great! Thank you ❤️

PR #19424, opened 2021-10-02
[fastlane_core] symbolize keys in options of type Hash

keeshux:
### Checklist

### Motivation and Context

When Hash options are provided via environment, they are parsed as JSON and auto-converted to a hash. However, Ruby treats string and symbol keys differently, and fastlane actions normally expect keys as symbols.

### Description

Converts keys to symbols when parsing an option from a JSON string in the environment.

### Testing Steps

The `APP_STORE_CONNECT_API_KEY` environment variable is a common example exposing the issue. Any key inside the JSON would be parsed as nil, because accessing e.g. `"key_id"` (parsed) is different than `:key_id` (expected).

### DIFF
--- a/fastlane_core/lib/fastlane_core/configuration/config_item.rb
+++ b/fastlane_core/lib/fastlane_core/configuration/config_item.rb
@@ -282,7 +282,7 @@ module FastlaneCore
         return value.map { |k, v| "#{k.to_s.shellescape}=#{v.shellescape}" }.join(' ') if value.kind_of?(Hash)
       elsif data_type == Hash && value.kind_of?(String)
         begin
-          parsed = JSON.parse(value)
+          parsed = JSON.parse(value, :symbolize_names => true)
           return parsed if parsed.kind_of?(Hash)
         rescue JSON::ParserError
         end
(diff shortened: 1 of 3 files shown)


joshdholtz (changes requested):
I like this theory! But it looks like this breaks some existing tests 😔 I think the correct solution for the issue here might be to fix the string/symbol lookup in the App Store Connect API Key area if that is where its happening 🤷‍♂️  Thoughts?!

keeshux:
> I like this theory! But it looks like this breaks some existing tests 😔 I think the correct solution for the issue here might be to fix the string/symbol lookup in the App Store Connect API Key area if that is where its happening 🤷‍♂️ Thoughts?!

Yeah! I saw that l10n tests failed in CI, probably for the opposite reason. In fact such a change might have subtle and undesired consequences.

Therefore, I reverted the global change and restricted the fix to the `app_store_connect_api_key` action, which is where the bug is most apparent to my experience.

joshdholtz:
@googlebot I consent.

joshdholtz (approved):
This is great! Thank you so much for fixing this 🥰

PR #19459, opened 2021-10-12
[pem] fix issue OpenSSL empty password

fermoya:
### Checklist

### Motivation and Context

Resolves [this issue](https://github.com/fastlane/fastlane/issues/2808). I'm having similar problems: passing an empty password to `pem` will make the generated `p12` file impossible to import to the Keychain.

### Description

`OpenSSL` seems to have some kind of issue if the `passphrase` used to generate a `PKCS12` file is `""` (empty string). Thus, I'm replacing that with a `nil` value and also marking the parameter as `optional` (not doing so makes the user to introduce it manually).

After the fix in this pull-request, it looks like this:
https://user-images.githubusercontent.com/11335612/136920514-b6730a44-5804-4a08-9655-9264b1923993.mov

### Testing Steps

* Use `pem` lane and enable `generate_p12` option. Don't pass any `p12_password`.
```
# inside custom lane...
pem(
  development: options[:development],
  generate_p12: true,
  force: true,
  app_identifier: app_identifier,
  username: username,
  team_id: team_id,
  team_name: team_name,
  pem_name: app_name
)
```
* Try to import to Keychain (double-click) and introduce an empty `passphrase` when asked.
https://user-images.githubusercontent.com/11335612/136920349-7aba2681-499e-48ce-9d92-8caed2eb7b3b.mov

### DIFF
--- a/pem/lib/pem/manager.rb
+++ b/pem/lib/pem/manager.rb
@@ -73,7 +73,8 @@ module PEM
 
         if PEM.config[:generate_p12]
           p12_cert_path = File.join(output_path, "#{filename_base}.p12")
-          p12 = OpenSSL::PKCS12.create(PEM.config[:p12_password], certificate_type, pkey, x509_certificate)
+          p12_password = PEM.config[:p12_password] == "" ? nil : PEM.config[:p12_password]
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
This fix is 🔥 Thank you so much for fixing this!! ❤️ Really appreciate it!

PR #19468, opened 2021-10-13
[scan] add `Success rate` row to results

mullerfrederik:
### Checklist

### Motivation and Context
This change adds `Success rate` row to results so user gains better idea how many (percentage) tests are passing/failing.

### Description
Add new row when puts results.


joshdholtz (changes requested):
I think we need to change the calculation here 🤷‍♂️

PR #19507, opened 2021-10-21
[gym] measure build time in gym and share data to lane_context

ainame:
### Checklist

### Motivation and Context

In my project, we are measuring build time by wrapping `build_ios_app` with Benchmark.measure`, like below. This is enough to monitor the trend of changes in our app's build time. 

```ruby
tms = Benchmark.measure do 
  build_ios_app
end
UI.message("Build Time: #{tms.real} secs")
```

However, this is not only inaccurate to see how much time "xcodebuild archive" spends but also not looking great on `Fastfile` as it brings "Ruby" code to fastlane's beautiful DSLs.

This PR adds a small tweak to expose the elapsed time for building/archiving app in `gym` to `Actions.lane_context`, which then we can use it for logging.

### Description

I considered using `Benchmark` module initially as I used it in our project but realised `require 'benchmark'` in fastlane is not worth doing because `Benchmark` is useful when it comes to measuring "CPU time" that is executing code but `xcodebuild archive` normally takes time for IO for copying files so CPU time isn't suitable to see build time. 

Instead I simply used `Time.now` to get start time and finish time and worked out the elapsed time!

There are various tools to check build time but this PR's one benefits by having no dependency other than fastlane itself. 

https://github.com/MobileNativeFoundation/XCLogParser

### Testing Steps

Add this to Gemfile and run `bundle install`

```ruby
gem 'fastlane', git: 'https://github.com/ainame/fastlane.git', branch: 'measure-build-time'
```

You can test this with something like this lane.

```ruby
lane :test do 
  match
  build_ios_app
  UI.message("Build Time = #{Actions.lane_context[SharedValues::GYM_BUILD_TIME]} secs")
end

### DIFF
--- a/fastlane/lib/fastlane/actions/build_app.rb
+++ b/fastlane/lib/fastlane/actions/build_app.rb
@@ -5,6 +5,7 @@ module Fastlane
       PKG_OUTPUT_PATH ||= :PKG_OUTPUT_PATH
       DSYM_OUTPUT_PATH ||= :DSYM_OUTPUT_PATH
       XCODEBUILD_ARCHIVE ||= :XCODEBUILD_ARCHIVE # originally defined in XcodebuildAction
+      GYM_BUILD_TIME = :GYM_BUILD_TIME
     end
 
     class BuildAppAction < Action
@@ -65,7 +66,8 @@ module Fastlane
           end
         end
 
-        gym_output_path = Gym::Manager.new.work(values)
+        manager = Gym::Manager.new
+        gym_output_path = manager.work(values)
         if gym_output_path.nil?
           UI.important("No output path received from gym")
           return nil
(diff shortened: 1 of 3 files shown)


joshdholtz (changes requested):
I like this! Is there any benefit to also adding an `GYM_ARCHIVE_TIME` as well? 😇

ainame:
> I like this! Is there any benefit to also adding an GYM_ARCHIVE_TIME as well? 😇

@joshdholtz 

Thanks! Does that mean you want to distinguish build and archive? As far as I understand that `gym` switches the command to `build` or `archive`  based on `skip_archive` option here. 

https://github.com/fastlane/fastlane/blob/d68d2aac2f1496776aab168c94342ab109bd086d/gym/lib/gym/generators/build_command_generator.rb#L56-L57

I can differentiate `GYM_BUILD_TIME` and `GYM_ARCHIVE_TIME` but it all depends on just one flag to `*skip*_archive`, which isn't clear on what xocdebuild's subcommand is used. In my opinion, it might be a bit confusing to choose an appropriate key for the lane context by that option. A single key to get elapsed time is easier than that and users would be able to figure out what spends for the build time from console log, or their code and intention. 

I think the name `BUILD_TIME` here means in a broad sense. it may not be too clear about whether build or archive but it's not wrong for archiving as well? What do you think?

AdalbertoMadden:
Thank you so much to sharing this post.

Kivettt:
Thank you for sharing your valuable solution, it's work properly.

AdalbertoMadden:
Thank you for sharing your valuable solution, it's work properly.

Keon241:
I'm likewise searching for an answer, Because I am additionally getting an equivalent issue. Kindly assistance! What do we need to do?

PR #19533, opened 2021-10-26
[env] add architecture type info to the fastlane env

Kukurijek:
### Checklist

### Motivation and Context

Resolves #19532
### Description

New variable `Architecture` for architecture type in the `fastlane env` (macOS only).
For more information please see #19532

### Testing Steps

run `fastlane env` and check the result - you should be able to find information about your architecture below the swift version (macOS ONLY).


fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -105,6 +105,12 @@ def self.mac?
       (/darwin/ =~ RUBY_PLATFORM) != nil
     end
 
+    # @return CPU type
+    def self.cpu_type
+      return "" unless self.mac?
+      return `sysctl -n machdep.cpu.brand_string`.split("\n").first
```
joshdholtz, line 111: Instead of CPU type, do we actually (or also) want the architecture type? This might be good to see if its running in Rosetta? This can be checked with `uname -p` or `arch`. Thoughts? 😇

joshdholtz (changes requested):
@Kukurijek This is a great idea! Thoughts on either changing or adding if its running in arm or i386 (Rosetta)?

fastlane_core/lib/fastlane_core/helper.rb
Kukurijek: Hi @joshdholtz !
You are completely right. Information about the type of cpu is completely unnecessary if we know what the architecture is. On the other hand, maybe some users may find this information (CPU Info) sensitive, and since we don't need it, I replace it architecture type.

`uname -p` returns a lot of information that is not needed (at least at this point 😄 ), so I changed the command to `arch` which returns only type eg. `i386`

Thanks! 🚀

Kukurijek: Actually `uname -p` returns the same thing as `arch`. I used some other commando before 🤣. Anyway it works and PR is updated.

iBotPeaches:
Going to close this. Something looks odd with history and full of unrelated changes. Sorry we never got this merged in time.

Kukurijek:
> Going to close this. Something looks odd with history and full of unrelated changes. Sorry we never got this merged in time.

Thanks for taking the time to go through all the old PRs!

In this case, the branch history got polluted when the old remote branch was merged into a newer local branch, which pulled in a lot of unrelated commits. This is no longer a hot topic, but the change itself could still be useful. It would just need to be recreated on a fresh branch.

Have fun working on fastlane 🚀

iBotPeaches:
Yep it was probably this commit in isolation is all we need - https://github.com/fastlane/fastlane/pull/19533/changes/bd51ac0eaf2a9fe7fee1ce50cbab18696a415de0

Saying that though, with Apple taking Intel out of equation shortly. I'm less worried about knowing architecture when I'd only really want to know a few years ago when it was heavily mixed with intel vs apple silicon.

PR #19589, opened 2021-11-09
[action][get_version_number] - search for MARKETING_VERSION in build settings if target has no INFO_PLIST

nekrich:
### Checklist

### Motivation and Context
I have a project with static lib target, and it has no INFO_PLIST in build settings. And get_version_number fails to get MARKETING_VERSION from build settings

### Description
Despite missing INFO_PLISToption, there still is a MARKETING_VERSION 😅.
So I added an option to skip looking for the value, or used variable in the Info.plist file if there is none. As the default value for looking in the build settings, I use MARKETING_VERSION, which is default from Xcode 11.

### Testing Steps
I've updated specs for get_version_number to cover this case.

Manual steps:
1. Create target w/o INFO_PLIST build setting (static lib, aggregate).
2. Set MARKETING_VERSION build setting.
3. Run get_version_number on that target.

### DIFF
--- a/fastlane/lib/fastlane/actions/get_version_number.rb
+++ b/fastlane/lib/fastlane/actions/get_version_number.rb
@@ -109,8 +109,10 @@ module Fastlane
           options = plist_files.keys
           selected = UI.select("What build configuration would you like to use?", options)
           plist_file = plist_files[selected]
-        else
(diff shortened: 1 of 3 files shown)


joshdholtz (approved):
This looks great! Thank you so much for making this change 🙌

PR #19604, opened 2021-11-12
[match][cert] allow developer_id creation if using apple id

joshdholtz:
### Motivation and Context
Disovered (when testing another match PR) that I was having issues running match with a developer id

### Description
- cert
  - opens creation of developer id certificates if authoring with Apple ID. This works because the unofficial API allows this 
- match
  - when used with ASC API, listing all certs won't return developer id certs
  - this changes adds a little line of description in the error that mentions using apple id 

### Testing Steps

```rb
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "joshdholtz-fix-developer_id-certs-not-being-listed"
```


lucgrabowski (approved):
LGTM :)

PR #19614, opened 2021-11-16
[gym] Add xcodebuild_command option

nekrich:
### Checklist

### Motivation and Context
We have not the best CI runner in the world (Azure Pipelines), and we trying to move to self-hosted runners with M1 CPUs. Unfortunately, Azure Pipelines agent runs everything under Rozetta (https://github.com/microsoft/azure-pipelines-agent/issues/3184), and we want to put `arch -arm64` before `xcodebuild` while archiving our targets.

Additionally, we used the `xcodebuild_command` option in `scan`, and were surprised that `gym` still misses it despite the fact scan uses [`GYM_XCODE_BUILD_COMMAND ` ENV var](https://github.com/fastlane/fastlane/blob/5dc9f74769c0b9e3688e8befe24f3ef6de4c6187/scan/lib/scan/options.rb#L465) for this option. If you need me to change the ENV var name in `scan` ping me, and I'll do it.

### Description
Copy-pasted everything related to `xcodebuild_command` from `scan` 😅 and covered it with a test.

Also, I removed `env NSUnbufferedIO=YES` prefix for the default `xcodebuild_command` value in the `gym` action, cause so far it worked well without it.

### Testing Steps
```
scan(xcodebuild_command: `echo 'Hello, World!'; xcodebuild`)
```


joshdholtz (approved):
OMG, I did not remember that `scan` had this and also that it was prefixed with `GYM` 😬 That feels like a bad copy paste bug we will just leave that there. This looks great though! Thanks for adding this ❤️

PR #19658, opened 2021-11-29
[spaceship] spaceauth - add new functionality to check if a user's session is still valid

justAnotherDev:
### Checklist

### Motivation and Context

For CI testing it would be useful to quickly/succinctly tell if a `spaceship` session for a specific user is still valid. This would allow a CI build to immediately fail a task that requires authentication if the session is expired/missing.

### Description

This PR adds a new parameter, `--check_session`, to expose existing code in the `spaceauth` command. Using the new parameter allows users to check if a session is still valid. If the session is valid the command will exit with a status code of 0, if the session is invalid it will exit with a status code of 1.

Example Usage:

```
fastlane spaceauth --check_session
[✔] 🚀 
Username: some-valid-user
Valid session found (some-valid-user). Exiting.
```

```
fastlane spaceauth --check_session -u some-expired-user
[✔] 🚀 
No valid session found (some-expired-user). Exiting.
```

### Testing Steps

Testing via the CLI should be pretty straightforward, recommend testing with and without a user set in the `Appfile` and testing with real and fake usernames.

```
fastlane spaceauth --check_session
fastlane spaceauth --check_session -u real-username
fastlane spaceauth --check_session -u fake-username
```

There are also 3 new tests that can be run locally:

```
bundle exec rspec spaceship/spec/spaceauth_spec.rb

Spaceship::SpaceauthRunner
  uses all required cookies for fastlane session
  copy_to_clipboard option
    when true, it should copy the session to clipboard
    when false, it should not copy the session to clipboard
  check_session option
    when using the default user, it should return a message saying the session is logged in with an exit code of 0
    when passed a known user, it should return a message saying the session is logged in with an exit code of 0
    when passed an unknown user, it should return a message saying no valid session found with an exit code of 1
  #session_string
    should return the session when called after run
    should throw when called before run

Finished in 0.83703 seconds (files took 3.54 seconds to load)
8 examples, 0 failures
```

### DIFF
--- a/spaceship/docs/Authentication.md
+++ b/spaceship/docs/Authentication.md
@@ -26,4 +26,6 @@ Note that Apple does not always allow in all situations. For example for individ
 
 If you _always_ want your security sent via SMS to a specific trusted phone number you can set the `SPACESHIP_2FA_SMS_DEFAULT_PHONE_NUMBER` environment variable to that phone number. The phone number should be specified in the same format as it is displayed in your [Apple ID console](https://appleid.apple.com/) under `TRUSTED PHONE NUMBERS`, e.g. `+49 123 4567890`, `+1-555-123-4567` or similar. Do not leave off the country code or add or remove any numbers, otherwise fastlane will not be able to match the masked value from Apple's API and select the correct number.
 
+## Checking if a session is valid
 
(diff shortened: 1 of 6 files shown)


justAnotherDev:
The CLA has been signed, please rescan.

joshdholtz (approved):
This is 🔥 Thank you so much for adding this and tests! Really appreciate the contribution 🥰

PR #19670, opened 2021-12-03
[action][download_dsyms] use filter argument to /builds to find specific build

### LINKED ISSUE #19669: download_dsyms loops over all builds, taking several minutes
I have an app with a ton of builds (several per day for the last few years). When I run `download_dsyms`, it takes a really long time even though I'm trying to download the dsyms for ONE specific build using the `app_identifier/platform/version/build_number` arguments.

```
$ ls -alh /tmp/spaceship1638471392_17246_70299597086940.log
-rw-r--r-- 1 fancybox wheel 6.2M Dec  2 13:58 /tmp/spaceship1638471392_17246_70299597086940.log

$ grep ">>" /tmp/spaceship1638471392_17246_70299597086940.log | grep /builds | wc -l
      98

$  grep ">>" /tmp/spaceship1638471392_17246_70299597086940.log | grep /builds | head -1
INFO  [13:56:37]: >> GET https://api.appstoreconnect.apple.com/v1/builds?cursor=Cg.GY4EMw&include=preReleaseVersion%2CbuildBundles&limit=10&sort=-uploadedDate&filter%5BpreReleaseVersion.platform%5D=TV_OS&filter%5Bapp%5D=1439824821: [undefined body]

$ grep ">>" /tmp/spaceship1638471392_17246_70299597086940.log | grep /builds | tail -1
INFO  [13:58:16]: >> GET https://api.appstoreconnect.apple.com/v1/builds?cursor=A9Q.HbAEVA&include=preReleaseVersion%2CbuildBundles&limit=10&sort=-uploadedDate&filter%5BpreReleaseVersion.platform%5D=TV_OS&filter%5Bapp%5D=1439824821: [undefined body]
```

This happens regardless of `after_uploaded_date` usage, because the builds are all fetched here:

https://github.com/fastlane/fastlane/blob/dac31640214dc976da675e863d0efbd4a99cb526/fastlane/lib/fastlane/actions/download_dsyms.rb#L78-L81

cc @joshdholtz

tmm1:
fixes #19669 
cc @joshdholtz

### DIFF
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -76,6 +76,7 @@ module Fastlane
         filter = { app: app.id }
         filter["preReleaseVersion.platform"] = platform
         filter["preReleaseVersion.version"] = version if version
+        filter["version"] = build_number if build_number
         build_resp = Spaceship::ConnectAPI.get_builds(filter: filter, sort: "-uploadedDate", includes: "preReleaseVersion,buildBundles")
 
         build_resp.all_pages_each do |build|
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -75,6 +75,7 @@ module Fastlane
 
         filter = { app: app.id }
         filter["preReleaseVersion.platform"] = platform
+        filter["preReleaseVersion.version"] = version if version
         build_resps = Spaceship::ConnectAPI.get_builds(filter: filter, sort: "-uploadedDate", includes: "preReleaseVersion,buildBundles").all_pages
         builds = build_resps.flat_map(&:to_models)
 
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -76,10 +76,9 @@ module Fastlane
         filter = { app: app.id }
         filter["preReleaseVersion.platform"] = platform
         filter["preReleaseVersion.version"] = version if version
-        build_resps = Spaceship::ConnectAPI.get_builds(filter: filter, sort: "-uploadedDate", includes: "preReleaseVersion,buildBundles").all_pages
-        builds = build_resps.flat_map(&:to_models)
+        build_resp = Spaceship::ConnectAPI.get_builds(filter: filter, sort: "-uploadedDate", includes: "preReleaseVersion,buildBundles")
 
-        builds.each do |build|
+        build_resp.all_pages_each do |build|
           asc_app_version = build.app_version
           asc_build_number = build.version
           uploaded_date = DateTime.parse(build.uploaded_date)
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -100,7 +100,7 @@ module Fastlane
 
           if after_uploaded_date && after_uploaded_date >= uploaded_date
             UI.verbose("Upload date #{after_uploaded_date} not reached: #{uploaded_date}")
-            next
+            break
           end
 
           message = []
--- a/fastlane/lib/fastlane/actions/download_dsyms.rb
+++ b/fastlane/lib/fastlane/actions/download_dsyms.rb
@@ -101,7 +101,7 @@ module Fastlane
 
           if after_uploaded_date && after_uploaded_date >= uploaded_date
             UI.verbose("Upload date #{after_uploaded_date} not reached: #{uploaded_date}")
-            break
+            next
           end
 
           message = []
--- a/fastlane/spec/actions_specs/download_dsyms_spec.rb
+++ b/fastlane/spec/actions_specs/download_dsyms_spec.rb
@@ -74,7 +74,12 @@ describe Fastlane do
 
       context 'with no special options' do
         it 'downloads all dsyms of all builds in all trains' do
-          expect(build_resp).to receive(:to_models).and_return([build1, build2, build3, build4, build5, build6])
+          expect(build_resp).to receive(:all_pages_each).and_yield(build1)
+                                                        .and_yield(build2)
+                                                        .and_yield(build3)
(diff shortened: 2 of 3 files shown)


tmm1:
> The former is not specific enough to be useful in filtering.

In my case I have 500 builds under v2020 and another 500 under v2021. So using `filter["preReleaseVersion.version"] = "2021"` does help in that it cuts down the number of requests and time by half. But its still almost a minute to download all 500 builds just to notice that the one we want isn't in there yet.

Anyway I updated this PR so its functional now, but would be nice to find further performance improvements here. If the API let us also specify `preReleaseVersion.buildNumber` or similar that would be ideal. But I didn't see anything like that on https://developer.apple.com/documentation/appstoreconnectapi/list_builds

tmm1:
Alright I managed to get this working. Now the same `download_dsyms` invocation takes less than one second for me, vs almost two minutes before.

joshdholtz:
@tmm1 I _think_ what you want is `filter[version]` 🤔  `build.version` is the build number 

`build.appStoreVersion` and `build. preReleaseVersion` are the version numbers

Example of using `filter[version]` 👇 

Screen Shot 2021-12-02 at 8 41 10 PM

Sooo... you could filter by `preReleaseVersion.version` (for the version) and `version` (for the build number) to really narrow it down 🤷‍♂️ Thoughts?

tmm1:
> Sooo... you could filter by `preReleaseVersion.version` (for the version) and `version` (for the build number) to really narrow it down 🤷‍♂️ Thoughts?

You're right, that totally worked and is super fast. Updated this PR accordingly!

tmm1:
Before:

```
+------+-----------------------+-------------+
|              fastlane summary              |
+------+-----------------------+-------------+
| Step | Action                | Time (in s) |
+------+-----------------------+-------------+
| 1    | opt_out_usage         | 0           |
| 2    | gym                   | 184         |
| 3    | gym                   | 205         |
| 4    | pilot                 | 94          |
| 5    | pilot                 | 73          |
| 6    | download_dsyms        | 108         |
| 7    | download_dsyms        | 99          |
| 8    | download_dsyms        | 97          |
| 9    | download_dsyms        | 103         |
| 10   | download_dsyms        | 103         |
| 11   | download_dsyms        | 102         |
| 12   | download_dsyms        | 99          |
| 13   | download_dsyms        | 109         |
| 14   | sentry_upload_dif     | 90          |
| 15   | download_dsyms        | 98          |
| 16   | download_dsyms        | 94          |
| 17   | download_dsyms        | 103         |
| 18   | sentry_upload_dif     | 16          |
| 19   | clean_build_artifacts | 0           |
+------+-----------------------+-------------+
```

After:
```
+------+-----------------------+-------------+
|              fastlane summary              |
+------+-----------------------+-------------+
| Step | Action                | Time (in s) |
+------+-----------------------+-------------+
| 1    | opt_out_usage         | 0           |
| 2    | gym                   | 178         |
| 3    | gym                   | 186         |
| 4    | pilot                 | 87          |
| 5    | pilot                 | 78          |
| 6    | download_dsyms        | 5           |
| 7    | download_dsyms        | 4           |
| 8    | download_dsyms        | 3           |
| 9    | download_dsyms        | 5           |
| 10   | download_dsyms        | 5           |
| 11   | download_dsyms        | 4           |
| 12   | download_dsyms        | 3           |
| 13   | download_dsyms        | 3           |
| 14   | download_dsyms        | 3           |
| 15   | download_dsyms        | 4           |
| 16   | download_dsyms        | 3           |
| 17   | download_dsyms        | 2           |
| 18   | download_dsyms        | 2           |
| 19   | download_dsyms        | 4           |
| 20   | download_dsyms        | 3           |
| 21   | download_dsyms        | 6           |
| 22   | download_dsyms        | 5           |
| 23   | download_dsyms        | 3           |
| 24   | download_dsyms        | 4           |
| 25   | download_dsyms        | 4           |
| 26   | download_dsyms        | 5           |
| 27   | download_dsyms        | 3           |
| 28   | download_dsyms        | 3           |
| 29   | download_dsyms        | 3           |
| 30   | download_dsyms        | 3           |
| 31   | download_dsyms        | 5           |
| 32   | sentry_upload_dif     | 91          |
| 33   | download_dsyms        | 4           |
| 34   | download_dsyms        | 4           |
| 35   | download_dsyms        | 4           |
| 36   | download_dsyms        | 3           |
| 37   | download_dsyms        | 4           |
| 38   | download_dsyms        | 3           |
| 39   | download_dsyms        | 4           |
| 40   | download_dsyms        | 3           |
| 41   | download_dsyms        | 4           |
| 42   | download_dsyms        | 3           |
| 43   | download_dsyms        | 4           |
| 44   | sentry_upload_dif     | 14          |
| 45   | clean_build_artifacts | 0           |
+------+-----------------------+-------------+

**Review by joshdholtz — APPROVED** [review]
This looks great! I added one commit to fix the tests but this is 🔥 Thanks you so much ❤️

File: fastlane/lib/fastlane/actions/download_dsyms.rb [hunk 1]
```diff
@@ -101,7 +101,7 @@ def self.run(params)
 
           if after_uploaded_date && after_uploaded_date >= uploaded_date
             UI.verbose("Upload date #{after_uploaded_date} not reached: #{uploaded_date}")
-            break
+            next
```
tmm1: Using next here doesn't really make sense to me. I tried to reason about it in my commit message in dac7592c98e0bb84028b5b694558182c8ae49dbd

Since we know the results are sorted by uploadedDate, I think it should be safe to break out of the loop when we're past the expected date already.

joshdholtz: @tmm1  Yeah yeah, that does seem okay but I don't _fully_ trust they will be returned in the right order 😬 I trust the official ASC API but we also use the proxy endpoint for Apple ID auth which is the one I don't really trust. So I _think_ I'd feel safer here doing a next just to make sure all the correct ones are found. The slowest part was the API request so this _shouldn't_ increase things too much 🤞 

I hope my insane uneasiness makes sense 😛

tmm1: Fair enough :boat:

PR #19678, opened 2021-12-07
[action] [update_code_signing_settings] add entitlements file path ca…

abcdev:
### Checklist

### Motivation and Context
Fastlane already has a couple of actions to change/add entries in an apps entitlements file but there is no action that considers that this file is not added to the build settings. Leaving no other option than manually adding this file by opening the project in Xcode and clicking through UI.

We'd like to integrate fastlane into CI/CD of one of our projects where these entitlements files are generated on a per client basis. Hence the need for automation at this point.

### Description
Extended the existing action `update_code_signing_settings` to also handle the build setting `CODE_SIGN_ENTITLEMENTS`.

### Testing Steps
Added test case to existing `update_code_signing_settings_spec.rb` to add the new `CODE_SIGN_ENTITLEMENTS` entry in a Xcode project file where it is missing and to modify a project with an already existing one.


joshdholtz (approved):
This looks great! I've had to do this for a project and I did it a manually way so I'm wishing I had this then 😛 Thanks so much for the contribution 😍

Issue #19692, opened 2021-12-10
invalid byte sequence in UTF-8: upload_to_testflight can't process ipa files larger than 4GB


chenshifei:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Recently I got this `invalid byte sequence in UTF-8` error when uploading my ipa file to testflight. When I traced down the error and examine the output from `unzip -Z -1` I found there's some file name containing characters like `ö` or `å`.

These files have been in the app for quite a long time, and before they can be processed without any problems. I noticed that the last successful build was much smaller (about 2.3G) while now the ipa has grown massively into 6G. Could this be the problem?

##### Command executed

`bundle exec fastlane send_to_testflight "scheme:com.test.app" --verbose`

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
[✔] 🚀
DEBUG [2021-12-10 15:13:29.33]: Checking if there are any plugins that should be loaded...
Successfully loaded Appfile at path '/path/to/Appfile'
- apple_id: 'apple-dev@company.com'
-------
DEBUG [2021-12-10 15:13:29.56]: Using scheme: com.test.app
INFO [2021-12-10 15:13:29.57]: ---------------------------
INFO [2021-12-10 15:13:29.57]: --- Step: opt_out_usage ---
INFO [2021-12-10 15:13:29.57]: ---------------------------
INFO [2021-12-10 15:13:29.57]: Disabled upload of used actions
INFO [2021-12-10 15:13:29.57]: ------------------------------
INFO [2021-12-10 15:13:29.57]: --- Step: default_platform ---
INFO [2021-12-10 15:13:29.57]: ------------------------------
INFO [2021-12-10 15:13:29.57]: Driving the lane 'ios send_to_testflight' 🚀
INFO [2021-12-10 15:13:29.58]: -----------------------------
INFO [2021-12-10 15:13:29.58]: --- Step: setup_circle_ci ---
INFO [2021-12-10 15:13:29.58]: -----------------------------
INFO [2021-12-10 15:13:29.58]: Not running on CI, skipping CI setup
[...]

PR #19697, opened 2021-12-14
[fastlane_core] fix non-UTF-8 char issues when analysing ipa

chenshifei:
### Checklist

### Motivation and Context

Fixed an issue (#19692) where `ipa_file_analyser.rb` can't handle non-UTF-8 chars inside file names if the IPA file exceeds 4G.

### Description

Instead of getting the full file list and then looking for the app's `Info.plist` file, I change the code to look for the `Info.plist` file directly in the zip file.

Tested with my own app (around 6.6G) which has files with `Ö` in the file name.

### DIFF
--- a/fastlane_core/lib/fastlane_core/ipa_file_analyser.rb
+++ b/fastlane_core/lib/fastlane_core/ipa_file_analyser.rb
@@ -73,7 +73,7 @@ module FastlaneCore
     end
 
     def self.fetch_info_plist_with_unzip(path)
-      entry, error, = Open3.capture3("unzip", "-Z", "-1", path, "*/Payload/*.app/Info.plist")
+      entry, error, = Open3.capture3("unzip", "-Z", "-1", path, "*Payload/*.app/Info.plist")
 
       # unzip can return multiple Info.plist files if is an embedded app bundle (a WatchKit app)
       #   - ContainsWatchApp/Payload/Sample.app/Watch/Sample WatchKit App.app/Info.plist
--- a/fastlane_core/lib/fastlane_core/ipa_file_analyser.rb
+++ b/fastlane_core/lib/fastlane_core/ipa_file_analyser.rb
@@ -73,12 +73,9 @@ module FastlaneCore
     end
 
     def self.fetch_info_plist_with_unzip(path)
-      list, error, = Open3.capture3("unzip", "-Z", "-1", path)
+      entry, error, = Open3.capture3("unzip", "-Z", "-1", path, "*.app/Info.plist")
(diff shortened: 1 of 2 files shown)


chenshifei:
I signed the CLA now so it should be fine. Could you restart the check?

fastlane_core/lib/fastlane_core/ipa_file_analyser.rb
```diff
@@ -73,13 +73,20 @@ def self.fetch_info_plist_with_rubyzip(path)
     end
 
     def self.fetch_info_plist_with_unzip(path)
-      list, error, = Open3.capture3("unzip", "-Z", "-1", path)
+      entry, error, = Open3.capture3("unzip", "-Z", "-1", path, "*/Payload/*.app/Info.plist")
```
joshdholtz, line 76: There is a slight problem with using matching with `unzip` 😔 Matching on `*.app/Info.plist` does not handle the case with multiple apps in a `.app`. Like in the case of this WatchKit app 👇 

```
ContainsWatchApp/Payload/Sample.app/Watch/Sample WatchKit App.app/Info.plist                                                                    
ContainsWatchApp/Payload/Sample.app/Info.plist
```

I also tried to change it to `*/Payload/*.app/Info.plist` thinking that this might get the exact match of the main app Info.plist but `*.app` seems to behave like `**/*.app` 😬 

HOWEVER... I have a commit that think would fix that that I added 😊 If you would be able to review and test this for me that would be 🔥

joshdholtz (commented):
Thank you for making this fix! 

I did add a commit that should fix an error with an embedded up and also added some tests. Would you be able to review and test this whenever you get a chance? 😊

fastlane_core/lib/fastlane_core/ipa_file_analyser.rb
chenshifei: Hi Josh! Thank you for fixing this potential bug. I did a simple test on my side but the new code fails, will look into this later and come back to you!

chenshifei: For some reason, `*/Payload/*.app/Info.plist` can't cover my case although theoretically it should also capture `Payload/xxx.app/Info.plist`.

chenshifei: The latest code works for me now, so I assume we're clear for merge?

joshdholtz (approved):
Thank you so much for finding and debugging this! Excited to get this out 🥳

PR #19735, opened 2021-12-21
[pilot] support internal groups for testflight

horvathtamasattila:
### Checklist

### Motivation and Context

Resolves #19682
Apple added the concept of Internal Beta Groups in late August, with version 1.7 of App Store Connect. (Previously there were only External Beta Groups). The main difference between internal and external groups is that for external groups you need to send the app for a review, for internal groups that is not necessary. 

In the current form, `pilot` can successfully send the build to internal groups, but due to the implementation, it will always send the the build for a review, which is unnecessary for internal testing.

### Description

It seems that we just need to change one line in the code the achieve this behaviour, and only send the app for review if `distribute_external` is `true`, which makes sense because we don't want to send the app to review unless it's for external testing.

I tested `pilot` with: 
```
pilot(
  ...
  distribute_external: true,
  groups: ["...", "..."]
) 
```
**Build was distributed to groups and sent to review**
```
pilot(
  ...
  groups: ["...", "..."]
) 
```
**Build was distributed to groups but not sent to review**

### Testing Steps

`bundle exec rspec`


pilot/lib/pilot/build_manager.rb
```diff
@@ -410,7 +410,7 @@ def distribute_build(uploaded_build, options)
       # This is where we could add a check to see if encryption is required and has been updated
       uploaded_build = set_export_compliance_if_needed(uploaded_build, options)
 
-      if options[:groups] || options[:distribute_external]
+      if options[:distribute_external]
```
joshdholtz, line 413: This _totally_ makes sense but this would break existing users 😬 `: distribute_external` defaults to `false` which means if any users only set `:groups` then their existing behavior would break 😔

 I'm not a huge fan of this existing behavior but I don't think we can change that...

I wonder if we need to look at the groups that were specified and **not submit** for review if any of them are internal 🤔 Thoughts?

joshdholtz (changes requested):
Love this change but I think this sadly breaks existing behavior/usage 😔 Left thoughts on what we might need to do. I can help with this if its too much! Just let me know 😊

pilot/lib/pilot/build_manager.rb
horvathtamasattila: I missed this point, you are totally right 🙌
However, not submit for review if any of them are internal might not be the best solution. Users can still provide for example two internal groups and one external, and in that case they would want to send the app for a review. I see two possibilities here:

 1. We can check the groups and only **not submit** the app for review if every group they provided is internal
 2. We add a new parameter to pilot, something like `submit_beta_review` which defaults to `true`, so we keep the existing behaviour. In case the user sets it explicitly to false, the app would not be sent to reiview.

Personally I prefer the second solution, as it gives more control to the users. There might be use cases when they still want to send the app for a review even though all the groups they provided are internal ones. WDYT?

joshdholtz: @horvathtamasattila I like the `submit_beta_review` and default to `true` 😊 Let's go that approach!

horvathtamasattila: @joshdholtz I updated the PR and my manual tests succeed, but [one test](https://app.circleci.com/pipelines/github/fastlane/fastlane/4122/workflows/29cfe7af-0dd9-4f3e-a00a-1fdd252b08b2/jobs/79211) is not passing, and I cannot figure out why. Could you help me on that?

lucgrabowski: To fix this test you could add `submit_beta_review: true,` in `distribute_options_non_localized` in `pilot/spec/build_manager_spec.rb`, but I suppose that it would be also good to add some tests for another set of options with `submit_beta_review: false`. What do you think @joshdholtz ?

horvathtamasattila: @joshdholtz can we merge this? Or should I add something else?

joshdholtz (approved):
Love this! Super excited to get this out tonight 😊

joshdholtz:
@TacticCoder Are you trying to distribute to internal? If so, try...

```rb
pilot(
  skip_waiting_for_build_processing: true,
  api_key_path: "fastlane/api_key.json",
  groups: ['dev-testers'],
  submit_beta_review: false,
)
```

TacticCoder:
> @TacticCoder Are you trying to distribute to internal? If so, try...
> 
> ```ruby
> pilot(
>   skip_waiting_for_build_processing: true,
>   api_key_path: "fastlane/api_key.json",
>   groups: ['dev-testers'],
>   submit_beta_review: false,
> )
> ```

Still no luck😑

joshdholtz:
@TacticCoder Huh, weird! What exactly are you trying to do so that I can try to replicate? 

Is the goal to upload a TeastFlight build to an internal team only?

TacticCoder:
> @TacticCoder Huh, weird! What exactly are you trying to do so that I can try to replicate?
> 
> Is the goal to upload a TeastFlight build to an internal team only?
Thank you for the fast response!
Yes, but only for a predefined group, so if I have: dev-testers, uat-testers , for some builds i may want to allow only for dev-testers group to use.

QuirijnGB:
```ruby
    upload_to_testflight(
      skip_waiting_for_build_processing: true,
      ipa: '../ipa.ipa',
      api_key: app_store_connect_api_key(),
      groups: ['testing'],
      submit_beta_review: false,
    )
```
Am i doing something wrong here? 
It uploads but it distributes it to all the Internal Testing groups.

horvathtamasattila:
@QuirijnGB while creating your internal groups, make sure you have this box unchecked. If you have automatic build distributions enabled for a group, no matter what group you specify, those groups are going to receive every build. I couldn't find a way to disable this option once a group is created, the best way I think is to delete the group and recreate it with this checkbox disabled

Screenshot 2021-12-21 at 15 08 56

PR #11739, opened 2018-01-30
#10187 Fix xcodebuild inherited command not found

rkrim:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
There is actually a bug when executing gym with **analyze_build_time** parameter
Link to open issue: https://github.com/fastlane/fastlane/issues/10187
I updated the test spec according to changes.
All test passed.
Tested from command line.

### Description
When passing "analyze_build_time" parameter to "build_ios_app" aka gym, wrong xcodebuild flag "$(inherited)" is generated.
$(inherited) is an xcode environment variable where xcode checks the same variable value from the project level and replace it before executing the build step.
This is not intended to be used from command line.

PR #19737, opened 2021-12-21
[gym] correctly inherit OTHER_SWIFT_FLAGS when using option 'analyze_build_time'

icecrystal23:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass (well, sort of. Tests unrelated to my change were failing, but the test for the code I changed was passing when run in isolation)

### Motivation and Context

We had an issue with the Apollo SDK when using CocoaPods. Turns out the root cause of that bug was the OTHER_SWIFT_FLAGS not being successfully inherited here. See details of that bug here: https://github.com/apollographql/apollo-ios/issues/2047 

### Description

Added an additional escape character to the $ on "$(inherited)" OTHER_SWIFT_FLAGS so that it successfully gets passed to xcodebuild rather than being evaluated too early.

### Testing Steps

I tested `gym` on an example project with a custom OTHER_SWIFT_FLAGS setting and ensured that the actual `CompileSwiftSources` command in the build log had the custom flag in addition to the build time analysis ones.

```
-D MY_SWIFT_FLAG -Xfrontend -debug-time-function-bodies
```

### DIFF
--- a/gym/lib/gym/generators/build_command_generator.rb
+++ b/gym/lib/gym/generators/build_command_generator.rb
@@ -43,7 +43,7 @@ module Gym
           options << "-scmProvider system"
         end
         options << config[:xcargs] if config[:xcargs]
-        options << "OTHER_SWIFT_FLAGS=\"-Xfrontend -debug-time-function-bodies\"" if config[:analyze_build_time]
+        options << "OTHER_SWIFT_FLAGS=\"\\\$(value) -Xfrontend -debug-time-function-bodies\"" if config[:analyze_build_time]
 
         options
       end
--- a/gym/spec/build_command_generator_spec.rb
+++ b/gym/spec/build_command_generator_spec.rb
@@ -328,7 +328,7 @@ describe Gym do
                                "-project ./gym/examples/standard/Example.xcodeproj",
                                "-destination 'generic/platform=iOS'",
                                "-archivePath #{Gym::BuildCommandGenerator.archive_path.shellescape}",
-                               "OTHER_SWIFT_FLAGS=\"-Xfrontend -debug-time-function-bodies\"",
+                               "OTHER_SWIFT_FLAGS=\"\\$(value) -Xfrontend -debug-time-function-bodies\"",
                                :archive,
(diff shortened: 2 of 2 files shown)


icecrystal23:
^ that build failure seems unrelated to my change. I'm also not sure what the conflicts are since I just forked this morning for this bug and branched off the latest `master`

icecrystal23:
Ok, figured out that I had a REALLY old fork of fastlane and updated my PR.
I noticed the attempt to inherit was removed in #11739 but this correctly adds the flag back (it wasn't correctly included before which I could see leading to issues).

gym/lib/gym/generators/build_command_generator.rb
```diff
@@ -43,7 +43,7 @@ def options
           options << "-scmProvider system"
         end
         options << config[:xcargs] if config[:xcargs]
-        options << "OTHER_SWIFT_FLAGS=\"-Xfrontend -debug-time-function-bodies\"" if config[:analyze_build_time]
+        options << "OTHER_SWIFT_FLAGS=\"\\\$(value) -Xfrontend -debug-time-function-bodies\"" if config[:analyze_build_time]
```
joshdholtz, line 46: I'm not too familiar with this but is `$(value)` the same thing as `$(inherited)`? 🙃

joshdholtz (commented):
One small question otherwise this looks good 😊

gym/lib/gym/generators/build_command_generator.rb
icecrystal23: I don't know where or if there is official documentation for it, but `$(value)` expands to the current build setting value. So it's sort of the command-line equivalent of `$(inherited)`

joshdholtz: Ohhhhhh, cool cool! I see what this is doing now 🤯 Thank you for the explanation 🥰

joshdholtz (approved):
Thank you so much for debugging this and making this change! Really appreciate it 🙌

PR #19749, opened 2021-12-25
[scan] Fix: test retries succeeding when they shouldn't

### LINKED ISSUE #18587: [scan] retry feature did not work as expected
@xavierLowmiller On my side, the retried build fails (see logs below) : "and logs" string seems to cause the issue. Maybe because my app has multiple test targets or has coverage enabled ?
```
Failing tests:
Test session results, code coverage, and logs:
	/Users/admin/.gitlab-runner/builds/ios/xxxx/0/app/fastlane/test_output/App.xcresult
	FunctionalTests:
		TabbarTests.testTabbar()
		MenuTests.testMenu()
** TEST FAILED **
[18:18:36]: Exit status: 65
[18:18:36]: Retrying tests: and logs//Users/admin//gitlab-runner/builds/ios/xxxx/fastlane/test_output/App/xcresult
[18:18:36]: Number of retries remaining: 1
...
[18:19:02]: ▸ xcodebuild: error: Failed to build project Project with scheme App.
[18:19:02]: ▸ 	Reason: Tests in the target “and logs” can’t be run because “and logs” isn’t a member of the specified scheme.
```

_Originally posted by @shuguenot in https://github.com/fastlane/fastlane/issues/17765#issuecomment-823423124_

EricAtomic:
### Checklist

### Motivation and Context

The action `Scan` has a parameter called `number_of_retries` that allows tests to be re-run if some of them fail. However, currently the parsing code provides the wrong name of failed tests to `xcodebuild`'s `-only-testing` option if these tests are Objective-C codes (it works well with Swift codes).
For example, such a failed test could be printed in the log as: 
```
Failing tests:
MyFastlaneTests:
    -[MyObjectiveCTestClass testCoinToss]
```
Which results in the command line as 
`xcodebuild -only-testing:MyFastlaneTests/-[MyObjectiveCTestClass testCoinToss]`. 
Whereas the correct output should be: 
`xcodebuild -only-testing:MyFastlaneTests/MyObjectiveCTestClass/testCoinToss`.

This PR fixes such an issue by adding more parsing logic to the function responsible for generating retry tests.

Resolves #18587

### Description

We change the parsing code in the method `retryable_tests`, which adds the detection of `-[` and `]` when erasing them and the detection of whitespace when replacing them for `/`.

### Testing Steps

Added an objective c version sample method in the file `scan/spec/runner_spec.rb` in the test named `retry a fail test` and the next two tests.

### DIFF
--- a/scan/lib/scan/runner.rb
+++ b/scan/lib/scan/runner.rb
@@ -126,7 +126,7 @@ module Scan
 
         test_cases = suite.split(":\n").fetch(1, []).split("\n").each
                           .select { |line| line.match?(/^\s+/) }
-                          .map { |line| line.strip.gsub(".", "/").gsub("()", "") }
+                          .map { |line| line.strip.gsub(/[\s\.]/, "/").gsub(/[(\-\[)\](\(\))]/, "") }
                           .map { |line| suite_name + "/" + line }
 
         retryable_tests += test_cases
--- a/scan/lib/scan/runner.rb
+++ b/scan/lib/scan/runner.rb
@@ -126,7 +126,7 @@ module Scan
 
         test_cases = suite.split(":\n").fetch(1, []).split("\n").each
                           .select { |line| line.match?(/^\s+/) }
-                          .map { |line| line.strip.gsub(/[\s\.]/, "/").gsub(/[(\-\[)\](\(\))]/, "") }
+                          .map { |line| line.strip.gsub(/[\s\.]/, "/").gsub(/[\-\[\]\(\)]/, "") }
                           .map { |line| suite_name + "/" + line }
 
         retryable_tests += test_cases
--- a/scan/spec/runner_spec.rb
+++ b/scan/spec/runner_spec.rb
@@ -102,9 +102,10 @@ describe Scan do
 Failing tests:
   FastlaneAppTests:
           FastlaneAppTests.testCoinToss()
+          -[FastlaneAppTestsOC testCoinTossOC]
           ERROR_OUTPUT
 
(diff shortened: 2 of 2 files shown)


scan/lib/scan/runner.rb
```diff
@@ -126,7 +126,7 @@ def retryable_tests(input)
 
         test_cases = suite.split(":\n").fetch(1, []).split("\n").each
                           .select { |line| line.match?(/^\s+/) }
-                          .map { |line| line.strip.gsub(".", "/").gsub("()", "") }
+                          .map { |line| line.strip.gsub(/[\s\.]/, "/").gsub(/[\-\[\]\(\)]/, "") }
```
EricAtomic, line 129: The logic of the first regexp: if there exists `.` or white spaces, then replace them with `/`.
The logic of the second regexp: if there exists `-[]()`, then erase them. It is impossible for a single legal `-`, `[`, `(` or `)` to appear in the test name so we simply use a pattern that matches every single occurrence of them, rather than matching the sequence of `-[` and `()`.

EricAtomic, line 129: The logic of the first regexp: if there exists any `.` or white space, replace them with `/`.
The logic of the second regexp: if there exists any `-[]()`, erase them. 

It is impossible for a single legal `-`, `[`, `(` or `)` to appear in the test name so we simply use a pattern that matches every single occurrence of them, rather than using `(?:\-\[|\(\)|\])` to match the sequence of `-[` and `()`, which is three times slower than the former according to benchmark.

joshdholtz (approved):
This looks great! Thank you for fixing this 🙌

PR #19755, opened 2021-12-30
[Fastlane.Swift] Swift fastlane does not run on Apple Silicon 

kikeenrique:
* [Swift] Add test with fixture to check fastlane swift upgrades.

### Checklist

### Motivation and Context

Adds swift fastlane tests that:
* will check problems on Fastlane swift project upgrades.
* will check Fastlane swift project compilation and runtime problems. It will fail in mac M1 until #18502 is merged.

### Description

* Add a fastlane swift project fixture, based on fastlane v2.198. The project also implements a simple "custom" lane.
* Add a extra step to generate_swift_api
* Add `--enable_runner_upgrades` parameter to allow auto accept changes on fastlaneRunner upgrades.

### Testing Steps

Executing `bundle exec fastlane generate_swift_api`


minuscorp:
We shouldn't add the code duplicated, as a fixture, this should be closed and use the other PR better.

kikeenrique:
How would you test then swift project upgrades?
My intention would not be to upgrade fixtures, they will be use to test project upgrades from each version.
Every modification in fastlane runner xcode project should be updated in fastlane runner project, and you can't foresee when a user will update versions.

minuscorp:
Maintenance of Swift is made when there're requirements for it, it makes no sense to use a fixture because it wouldn't be the real code when somebody makes a change if you are using a static version of all the runner and the project itself. The flag and the other changes are good to add in the other PR.

Issue #19632, opened 2021-11-19
scan build_for_testing build directory is outside of derived_data_path when Xcode defaults are changed


Cyberbeni:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I have set Xcode to use project relative derived data and build directories but this makes scan output the build products in that location even when derived_data_path is specified.

##### Command executed
```
defaults write com.apple.dt.Xcode IDECustomDerivedDataLocation .DerivedData.noindex
defaults write com.apple.dt.Xcode IDEBuildLocationStyle Custom
defaults write com.apple.dt.Xcode IDECustomBuildLocationType RelativeToDerivedData
defaults write com.apple.dt.Xcode IDECustomBuildProductsPath Build/Products
defaults write com.apple.dt.Xcode IDECustomBuildIntermediatesPath Build/Intermediates
killall Xcode
fastlane run scan \
	workspace:"$XC_WORKSPACE" \
	scheme:"$SCHEME" \
	derived_data_path:"$OUT_DIR" \
	build_for_testing:true \
	--verbose
```

-> Build products will be at `$XC_WORKSPACE/../.DerivedData.noindex/Build/Products` instead of `$OUT_DIR/Build/Products` as expected

### Environment

<details><summary>🚫 fastlane environment 🚫</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 12.0.1                                                         |
| Ruby                        | 2.6.8                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.30.1 (Apple Git-130)                             |
| Installation Source         | ~/.gem/ruby/2.6.0/bin/fastlane                                 |
| Host                        | macOS 12.0.1 (21A559)                                          |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib |
| OpenSSL Version             | LibreSSL 2.8.3                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
[...]

PR #19756, opened 2021-12-30
[rspec] update defaults before testing on macOS

Cyberbeni:
### Checklist

### Motivation and Context
#19632 

### Description
Some of the tests (on macOS) should fail due to these changes.

### DIFF
--- a/spec_helper.rb
+++ b/spec_helper.rb
@@ -18,6 +18,11 @@ unless ENV["DEBUG"]
 end
 
 if FastlaneCore::Helper.mac?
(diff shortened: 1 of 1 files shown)


joshdholtz (changes requested):
I'm so torn about this 😱 Updating system defaults when the tests run is a little dangerous and these side effects aren't easily noticeable by anybody executing them. 

I'm thinking maybe that if these settings are messing with tests passing that should pass that we need probably adjust those tests 🤔 Thoughts?

Cyberbeni:
Wait, all the tests are passing with the modified settings. :(

joshdholtz:
Computers are a mystery 🤷‍♂️

PR #19828, opened 2022-01-22
[trainer] Make new options used by scan public (to fix crash)

### LINKED ISSUE #19826: trainer fails with "Could not find option 'output_filename'"
### New Issue Checklist

### Issue Description
I'm migrating to fastlane 2.201.0 which is adding `trainer` and thus requiring not to install it as a Plugin anymore.
However it looks like the trainer facade exposed via Fastlane is different, because I'm getting this error when I call it:
```
Could not find option 'output_filename' in the list of available options: path, extension, output_directory, fail_build, xcpretty_naming, silent
```
and I'm not specifying that `output_filename` option at all, however I do see mentions to it in #19629

##### Command executed
in my Fastfile:
```
    scan(
      output_types: '',
      output_directory: output_directory,
      result_bundle: true,
      fail_build: false
    )
    trainer(
      path: output_directory,
      output_directory: output_directory,
      fail_build: false
    )
```

It doesn't seem relevant here, just regular output from `scan` then the already mentioned error from trainer.

### Environment
It doesn't seem relevant here, any environment should be able to reproduce this, but to make this easier I created a repository https://github.com/gobetti/fastlane-201-issues where you can use the branch `trainer-output-filename`. Or this commit https://github.com/gobetti/fastlane-201-issues/commit/fd890253c4412f44d6e865c8e7b078d80c04b0c5
That repository contains a vanilla Xcode 13 project, my only change to it was to add a failing unit test.

joshdholtz:
### Motivation and Context
Fixes #19826

### Description
- Added `output_filename` and `output_remove_retry_attempts` to `trainer`'s options
  - These were already used by `scan` but were causing a crash when `trainer` was used by itself


gobetti (approved):
works! thank you!!!

PR #19844, opened 2022-01-25
[scan] fix off by one for scan's number_of_retries with xcode 13

### LINKED ISSUE #19842: [scan] number_of_retries on 2.201.1 + Xcode 13 behaves differently from documentation
### New Issue Checklist

### Issue Description
Per [docs](https://docs.fastlane.tools/actions/scan/) `number_of_retries` is
> The number of times a test can fail before scan should stop retrying

i.e. 0 (the default value) means no retries, only 1 iteration is allowed.
With Xcode 13, [`number_of_retries` is being input to the new `-test-iterations` setting in xcodebuild](https://github.com/fastlane/fastlane/blob/bdf2336949388cfcd9b13a718bca9c12bc6ab7df/scan/lib/scan/test_command_generator.rb#L75-L78) so it's retrying 1 time less than before.
Another example: now `number_of_retries: 1` actually means "don't retry".

##### Example
This should be reproducible using the `number-of-iterations-equals-retries` branch from https://github.com/gobetti/fastlane-201-issues , or this commit https://github.com/gobetti/fastlane-201-issues/commit/14b4c6ff2515769975edbf11a2c988ad74b1a01f

joshdholtz:
### Motivation and Context
Fixes #19842

### Description
- Adds 1 more test iteration to `:number_of_retries` is using Xcode 13's retries (test iterations)
- Updates `:number_of_retries` description to explain difference in behavior between Xcode 12 and under and Xcode 13 and up

### DIFF
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -515,7 +515,7 @@ module Scan
                                     default_value: false),
         FastlaneCore::ConfigItem.new(key: :number_of_retries,
                                     env_name: 'SCAN_NUMBER_OF_RETRIES',
-                                    description: "The number of times a test can fail before scan should stop retrying",
+                                    description: "In Xcode 13 and up, individual tests will retry on failure (see '-test-iterations' under `xcodebuild --help`). In Xcode 12 and under, the number of times a test can fail before scan should stop retrying",
                                     type: Integer,
                                     default_value: 0)
 
--- a/scan/lib/scan/options.rb
+++ b/scan/lib/scan/options.rb
@@ -515,7 +515,7 @@ module Scan
                                     default_value: false),
         FastlaneCore::ConfigItem.new(key: :number_of_retries,
                                     env_name: 'SCAN_NUMBER_OF_RETRIES',
-                                    description: "In Xcode 13 and up, individual tests will retry on failure (see '-test-iterations' under `xcodebuild --help`). In Xcode 12 and under, the number of times a test can fail before scan should stop retrying",
+                                    description: "The number of times a test can fail",
                                     type: Integer,
                                     default_value: 0)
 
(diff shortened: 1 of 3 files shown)


scan/lib/scan/options.rb
```diff
@@ -515,7 +515,7 @@ def self.available_options
                                     default_value: false),
         FastlaneCore::ConfigItem.new(key: :number_of_retries,
                                     env_name: 'SCAN_NUMBER_OF_RETRIES',
-                                    description: "The number of times a test can fail before scan should stop retrying",
+                                    description: "In Xcode 13 and up, individual tests will retry on failure (see '-test-iterations' under `xcodebuild --help`). In Xcode 12 and under, the number of times a test can fail before scan should stop retrying",
```
gobetti, line 518: this documentation sounds better to me before the change to add 1, given it looks like it might be saying that number of retries is test iterations on Xcode 13. If we're adding 1 though, then I can think we can say that this is the "the number of times a test can fail" regardless of the Xcode version. I'd still lean towards the approach of adding 1 given the name "number of retries" sounds unambiguous and self-explanatory that it's the number of times a test can be retried. What do you think?

gobetti: another (possibly parallel) possibility that occurs to me is to deprecate this config and add a new one borrowing from Xcode the term "iterations" so fastlane is more aligned with the native tool naming, and older versions would be the special case to handle (by retrying iterations-1 times)

joshdholtz: I think you are right! This is way too much 😅 We'll go with "Number of time a test can fail". Users don't need to _really_ know about the inner workings 🤷‍♂️

gobetti (approved):

PR #19860, opened 2022-01-26
[scan] Return results when running scan

tejassharma96:
### Checklist

### Motivation and Context

I wanted an easy way of getting the results from scan with `fail_build: false`. With this update, it's easy to view the results of scan simply by storing the value when you call it. I started a discussion [here](https://github.com/fastlane/fastlane/discussions/19859) as well but figured I'd just make a PR when I realized how straightforward it'd be.

### Description

I just added `results` to the end of `handle_results`, which will mean it gets returned from `run`.

### Testing Steps

Run scan in a new project and store the value. Print it to see the results.

### DIFF
--- a/fastlane/lib/fastlane/actions/run_tests.rb
+++ b/fastlane/lib/fastlane/actions/run_tests.rb
@@ -13,14 +13,13 @@ module Fastlane
         require 'scan'
         manager = Scan::Manager.new
 
-        results = nil
-
         begin
           results = manager.work(values)
 
           zip_build_products_path = Scan.cache[:zip_build_products_path]
           Actions.lane_context[SharedValues::SCAN_ZIP_BUILD_PRODUCTS_PATH] = zip_build_products_path if zip_build_products_path
 
+          return results
         rescue FastlaneCore::Interface::FastlaneBuildFailure => ex
           # Specifically catching FastlaneBuildFailure to prevent build/compile errors from being
           # silenced when :fail_build is set to false
@@ -47,8 +46,6 @@ module Fastlane
             Actions.lane_context[SharedValues::SCAN_GENERATED_PLIST_FILE] = all_test_summaries.last
           end
         end
-
-        return results
       end
 
       def self.description
--- a/fastlane/lib/fastlane/actions/run_tests.rb
+++ b/fastlane/lib/fastlane/actions/run_tests.rb
@@ -13,13 +13,14 @@ module Fastlane
         require 'scan'
         manager = Scan::Manager.new
 
+        results = nil
+
         begin
-          manager.work(values)
+          results = manager.work(values)
 
           zip_build_products_path = Scan.cache[:zip_build_products_path]
           Actions.lane_context[SharedValues::SCAN_ZIP_BUILD_PRODUCTS_PATH] = zip_build_products_path if zip_build_products_path
 
-          return true
         rescue FastlaneCore::Interface::FastlaneBuildFailure => ex
           # Specifically catching FastlaneBuildFailure to prevent build/compile errors from being
           # silenced when :fail_build is set to false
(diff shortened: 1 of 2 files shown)


tahirmt:
This is really useful. Would love to be able to post a custom slack message with the number of tests and failures.

tejassharma96:
I'm not sure what documentation might need to be updated here which is why I haven't checked that box yet, would love some feedback on whether documentation needs to be updated, and if so, where that would go.

tejassharma96:
just pre-empting the new commit here, but currently the hash will contain `:number_of_retries, :number_of_tests, :number_of_failures`. I do however think that it makes sense to return them including the retry counts so I can make that update tomorrow before this gets merged

joshdholtz:
@tejassharma96 Ah, I thought those were carried over too! I'll make that change right now 😊

joshdholtz:
@tejassharma96 Got this updated!

tejassharma96:
awesome, thanks for the fixes😅

tejassharma96:
also while you're here, just wanted to say thanks for the trainer+xcbeautify updates, really appreciating those 😄

joshdholtz:
@tejassharma96 My pleasure! 😊 My apologies on this taking so long to do 😛

tahirmt:
After this was merged, how do we now get count from trainer. I used to run trainer action separately but looks like this is only for scan

joshdholtz:
@tahirmt Hey! Which could are you looking for exactly? You _should_ be able to do something like...

```rb
results = scan(
  # your options here
)
```

It will return a hash like...

```rb
{:number_of_tests=>5,
 :number_of_failures=>0,
 :number_of_retries=>0,
 :number_of_skipped=>1,
 :number_of_tests_excluding_retries=>5,
 :number_of_failures_excluding_retries=>0}
```

Hope that helps! 😊

tahirmt:
Thanks @joshdholtz that’s what I missed in the docs! It works! I’m curious about one more thing though it’s a bit unrelated do I need to run `trainer` separately after scan if I want a junit output from trainer? Or can I get that out of scan itself?

tahirmt:
Oh actually I found it! https://github.com/fastlane/fastlane/blob/156f2114d19f7430a6a1d4c01b1c1e958d064830/scan/lib/scan/runner.rb#L242

tahirmt:
I've been noticing on my runs that when the tests fail then it doesn't seem to return a dictionary at all. I'm not sure why that is.

kamvoick:
Hello, that doesn't work for me either, on any failure there is no output. Possibly got stuck here
https://github.com/fastlane/fastlane/blob/156f2114d19f7430a6a1d4c01b1c1e958d064830/scan/lib/scan/runner.rb#L308

Although, as im seeing right now, trainer itself is also producing same kind of results and that works.

kamvoick:
I meant that scan doesn't return any tests results in case of failure, but trainer actually does
```
results = trainer(
path: lane_context[SharedValues::SCAN_GENERATED_XCRESULT_PATH],
output_directory: "./",
fail_build: false
)
```
then you will get
```
{ 
'*.xcresult' => { 
to_path: '*.xml', 
successful: false, 
number_of_tests: 3, 
number_of_failures: 2, 
number_of_tests_excluding_retries: 3, 
number_of_failures_excluding_retries: 2, 
number_of_retries: 0, 
number_of_skipped: 0 
} }
```

tejassharma96:
Do you have `fail_build` set to false when running scan? The default is true, and if it's set to true you won't receive any results from scan - it'll just exit early.

tahirmt:
@tejassharma96 I have it set to false through `fail_build(false)` in my `ScanFile`

kamvoick:
Yup, did that.

There is a condition on `handle_results`
https://github.com/fastlane/fastlane/blob/156f2114d19f7430a6a1d4c01b1c1e958d064830/scan/lib/scan/runner.rb#L305

that making it unable to pass `UI.test_failure!` because I assume that
`Methods ending in an exclamation point (!) terminate execution of the current lane and report an error`

tahirmt:
@joshdholtz you know more about how fastlane works. Is that the case? Should we create a ticket to fix it?

Issue #18663, opened 2021-05-09
Snapshot with Catalyst fails at `latest_simulator_version_for_device`


edwinveger:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I have an iOS app which supports macOS via Catalyst. I want to take snapshots of the Catalyst build.
I have a lane which executes a specific test case/suite. Executing the lane fails with an error inside snapshot's `simulator_launcher.rb`.

<pre>
fastlane_core/lib/fastlane_core/device_manager.rb:134:in `latest_simulator_version_for_device': \e[31m[!] undefined method `os_version' for nil:NilClass\e[0m (NoMethodError)
</pre>

TBH I'm not sure if this is actually a bug or just missing functionality. I noticed `scan` explicitly supports Catalyst via the parameter `catalyst_platform`. I tried to reverse engineer that but I'm inexperienced with Ruby and the Fastlane internals.

##### Command executed

`bundle exec fastlane stageoverviewmacos `

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> 
[✔] 🚀 
[12:18:24]: ----------------------------------------
[12:18:24]: --- Step: Verifying fastlane version ---
[12:18:24]: ----------------------------------------
[12:18:24]: Your fastlane version 2.182.0 matches the minimum requirement of 2.1.2  ✅
[12:18:24]: ------------------------------
[12:18:24]: --- Step: default_platform ---
[12:18:24]: ------------------------------
[12:18:24]: Driving the lane 'ios stageoverviewmacos' 🚀
[12:18:24]: ----------------------
[12:18:24]: --- Step: snapshot ---
[12:18:24]: ----------------------
[12:18:24]: Successfully loaded '/Users/edwin/Developer/VacationTime/fastlane/Snapfile' 📄

+----------------------------+----------------------------------------------------------------+
|                         Detected Values from './fastlane/Snapfile'                          |
+----------------------------+----------------------------------------------------------------+
| devices                    | ["iPhone 12 Pro Max", "iPad Pro (12.9-inch) (5th generation)"] |
| scheme                     | snapshot                                                       |
[...]

PR #19864, opened 2022-01-26
[snapshot] Fix compatibility with macOS

ptrkstr:
### Checklist

#### Why is this change required? What problem does it solve?
- Snapshot targets that target macOS would fail to build when including `SnapshotHelper.swift` because it contained symbols not available on macOS. This PR solves that problem by ensuring that macOS only sees symbols available to it.
- Additionally screenshots were being stored in an incorrect home directory.
- Enhancement added that allows specifying what `XCUIScreenshotProviding` object should be screenshotted. This is useful when taking screenshots for macOS and you'd prefer to screenshot the window and not the screen. This does not break backwards compatibility.

For conversations around Snapshot on macOS, please see:
- https://github.com/fastlane/fastlane/issues/11092
- https://github.com/fastlane/fastlane/issues/18663

### Description
- Added `screenshotProvider: XCUIScreenshotProviding = XCUIScreen.main` to `open class func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20)`.
- Inside `if os(OSX)` condition of the above mentioned function, removed call of `app.typeKey(XCUIKeyboardKeySecondaryFn, modifierFlags: [])` and added screenshot saving functionality.
- Wrapped `class func fixLandscapeOrientation(image: UIImage) -> UIImage` with `#if os(iOS) || os(watchOS) || os(tvOS)` as `UIImage` isn't available on macOS. 
- Change macOS cache directory to `URL(fileURLWithPath: "/Users/\(NSUserName())")`
- Updated `screenshot.mac?` to check against `Apple MacBook` instead of `MacBook`.
- Removed `self.top_space_above_device = offset['titleHeight'] # needed for centering the title` from `put_device_into_background` as `titleHeight` isn't available for Mac.
- Updated references to `@config` from `fetch_config`.

### Testing Steps
- Followed the fastlane steps on adding Snapshot to an Xcode project (in this case the Xcode project was for macOS).
- Added `concurrent_simulators(false)` to the snapfile based off the suggestion here: https://github.com/fastlane/fastlane/issues/18663#issuecomment-865484653.
- Observed generated html containing screenshots.


ptrkstr:
This is still in progress. We'll get round to it very soon.

joshdholtz:
@ptrkstr Hey! I love this PR and just wanted to see if there was anything that you needed help with 😊

ptrkstr:
Hey @joshdholtz thanks for reaching out! I think it should be good to go for review. Maybe one additional change could be that it doesn't require a user to add 'concurrent_simulators(false)' to their snapfile and instead that's ignored if it's snapshotting for mac?

ptrkstr:
Hey @joshdholtz, bumping this 😌

MortenGregersen:
Shouldn't we get this merged in, @joshdholtz ?

snapshot/lib/assets/SnapshotHelper.swift
```diff
@@ -32,9 +32,10 @@ func snapshot(_ name: String, waitForLoadingIndicator: Bool) {
 /// - Parameters:
 ///   - name: The name of the snapshot
 ///   - timeout: Amount of seconds to wait until the network loading indicator disappears. Pass `0` if you don't want to wait.
+///   - screenshotProvider: The provider to screenshot. Otherwise `XCUIScreen.main` is screenshotted. This is useful when taking screenshots for macOS and you'd prefer to screenshot the window and not the screen.
 @MainActor
-func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20) {
-    Snapshot.snapshot(name, timeWaitingForIdle: timeout)
+func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20, screenshotProvider: XCUIScreenshotProviding = XCUIScreen.main) {
+    Snapshot.snapshot(name, timeWaitingForIdle: timeout, screenshotProvider: screenshotProvider)
```
megabitsenmzq, line 38: The XCUIScreen.main gives me a Swift6 bug. You might need to look into it.

"Main actor-isolated class property 'main' can not be referenced from a nonisolated context; this is an error in the Swift 6 language mode"

megabitsenmzq:
I really hope this can be merged. I haven't tested the macOS compatibility part, but it fixes the bug that causes an error saying there is no such folder "Caches/tools.fastlane".

aehlke:
@joshdholtz is this still desired to merge or would any updates needed be unwelcome? very much want macos screenshots... thanks

PR #19938, opened 2022-02-09
[match] update :force_for_new_certificates option description

Lutzifer:
I'm not sure if this is the right documentation now, but it looks like it was a copy and paste error before.

### Checklist

### Motivation and Context
The description of the parameter matches another one so closely, that it does not seem right.

### Description
Changed the description to better match the parameter name.

### Testing Steps
run

```
fastlane actions match
```


joshdholtz (approved):
Gah! Good catch. Thank you for the fix 😊

PR #19953, opened 2022-02-11
[Fastlane.Swift] readPodspec: return map of [String: Any]

Hais:
🔑 
### Checklist

### Motivation and Context
Fixes a force-cast crash when using `readPodspec`. 
This is because a podspec can't typically be expressed as just a dictionary of `[String: String]`.   

### Description
Changing the return type uses an overload of `parseDictionary` which doesn't force-cast the return value to `[String: String]`, which causes a crash. 

This is technically a breaking API change to anyone using `readPodspec` but I struggle to see how any non-trivial Podspec would have successfully parsed as `[String: String]`.

### Testing Steps
Using `readPodspec` shouldn't crash, and return a successful `[String: Any]`.

### DIFF
--- a/fastlane/lib/fastlane/actions/read_podspec.rb
+++ b/fastlane/lib/fastlane/actions/read_podspec.rb
@@ -79,7 +79,7 @@ module Fastlane
       end
 
       def self.return_type
-        :hash_of_strings
+        :hash
       end
 
       def self.category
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
Good catch here! The Swift code is actually auto generated based on Ruby so I added a commit to fix that. I ran the task that generates the Swift and here is the output 👇 

```sh
  This can be useful when basing your release process on the version string only stored in one place - in the podspec.
  As one of the first steps you'd read the podspec and its version and the rest of the workflow can use that version string (when e.g. creating a new git tag or a GitHub Release).
  */
-@discardableResult public func readPodspec(path: String) -> [String: String] {
+@discardableResult public func readPodspec(path: String) -> [String: Any] {
     let pathArg = RubyCommand.Argument(name: "path", value: path, type: nil)
     let array: [RubyCommand.Argument?] = [pathArg]
     let args: [RubyCommand.Argument] = array
@@ -13430,4 +13430,4 @@ public let snapshotfile: Snapshotfile = .init()
```

Hais:
@joshdholtz Thanks for explaining! I had no idea it was automatically generated. I'll keep this in mind on any future Fastlane Swift contributions.

PR #19978, opened 2022-02-16
[action][git_commit action] use safer way of calling `sh`

AliSoftware:
By letting `sh` do all the escaping and argument handling for us

### Checklist

- [x] ~I've updated the documentation if necessary.~

### Motivation and Context

Even though the current implementation likely already worked for all cases (including spaces in paths etc), it is usually better and safer to let `sh` do the escaping and shell arguments handling for us rather than pass a full string built by us (and have the shell interpret it and split it back in individual arguments)

### Description

Used `sh("git", "command", "args")` instead of `sh("git command args")` as the former is considered more safe against injection and shell param interpretation

PS: There might be a couple more actions that also use `Actions.sh` that way instead of passing arguments as separate parameters; I only updated that action because I happened to using it stumble upon this while taking a look at its source code, but feel free to apply a similar change to other actions using `Actions.sh` as well.

(Optionally, we could consider migrating to the `Git` rubygem to do those git-related tasks in git-related fastlane actions… 🤔 )

### Testing Steps

Run the action with a couple of paths containing spaces and special characters

### DIFF
--- a/fastlane/lib/fastlane/actions/git_commit.rb
+++ b/fastlane/lib/fastlane/actions/git_commit.rb
@@ -2,22 +2,20 @@ module Fastlane
   module Actions
     class GitCommitAction < Action
       def self.run(params)
-        paths = params[:path].map(&:shellescape).join(' ')
-
-        skip_git_hooks = params[:skip_git_hooks] ? '--no-verify' : ''
+        paths = params[:path]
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
Ohhh, this looks good! Thanks for this ❤️ 

And hello 👋 😁

PR #19979, opened 2022-02-16
[Spaceship] Add support for unresolved_issues submission state

VrsajkovIvan33:
### Checklist

### Motivation and Context
Apple does not allow the creation of new review submissions in case there is already one that is considered to be "in progress" in the new submission flow. Consequently, [there is a check for in progress submissions](https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/submit_for_review.rb#L34) when submitting for review in `Deliver`. 

The `get_in_progress_review_submission()` method currently considers the "waiting for review" and "in review" states as the _in progress_ ones.

However, in case Apple detects an issue with a review submission, it will go into a new state which is `"UNRESOLVED_ISSUES"`. Submissions in this state also appear in the "In progress" section in the App Store Connect interface and the creation of a new submission is not allowed until the issues have been resolved or the existing submission has been cancelled.

Fastlane already won't allow you to proceed with the creation of a new submission because Apple will return an error stating:
```
"Resource state invalid - There is another reviewSubmissions with id 1234-abc-4567-def-890aaa123 still in progress."
```

Still, given that Fastlane is already checking for existing submissions that are in progress, we could fail at the same place as if the submission had an "in review" or "waiting for review" status.

### Description

Two things are introduced:
1. `UNRESOLVED_ISSUES` is added to the `ReviewSubmissionState` module
2. The `get_in_progress_review_submission()` method is updated to also add the newly added state to the filter

### Testing Steps

None, as there are no tests for the `get_in_progress_review_submission()` method. Let me know if you wish me to add something. 😃


joshdholtz (approved):
This looks good! Thank you for fixing this 😊

PR #20016, opened 2022-02-25
[trainer] Return Test Plan Configuration in JUnit output

sebskuse:
### Checklist

### Motivation and Context

We've started looking at using `Test Plan` configurations so that we can run our whole UI test suite against different dynamic font sizes. However, there is an issue when using `fastlane` + `trainer` to generate `JUnit`: There's no way to determine which Test Plan configuration the failure originated from - you just get duplicated `<testsuite>`s. We'd like to be able to raise the configuration that failed, and the information simply isn't there right now. For reference, this is how it appears within Xcode when you open the xcresult:

Screenshot 2022-02-25 at 16 22 56

And how it appears in the `JUnit` file after this change:

Screenshot 2022-02-25 at 17 06 37

### Description

The `Test Plan` configuration name is already available to us (via `ActionTestableSummary`), it's just not being used anywhere yet. This PR makes a couple of changes:

- Retrieve the name from the `ActionTestableSummary`, and set it on the returned test results `row` as `configuration_name`.
- Looking at the `JUnit` spec, there's obviously nothing particularly specific to this case, so I settled on using a `property` on the `testsuite`, as defined in the spec called `Configuration`.

### Testing Steps
The new field is now available on the `xcresult` based tests in `test_parser_spec`, so I have updated those, and I've added a new test within `junit_generator_spec` that verifies the output (also some extra coverage as currently there are no tests for `xcresult` -> `junit`).


joshdholtz (approved):
This looks great! Thank you so much for the contribution 🙌

Issue #20078, opened 2022-03-17
The request could not be completed because: Unauthorized Access


assembleMHN:
When using fastlane to try to upload a app update it fails with the below response:

The request could not be completed because:	Unauthorized Access

Fastlane can communicate with the Apple Developer Portal with no issues, but every-time i need to do work in App Store Connect i get the above error message.

I have tried to delete the  /Users/username/.fastlane/spaceship/{email}/cookie that did not do anything, i also tried to change my appleID password and sign-out of all devices, that also did not do anything, which is the solution stated in the past issues i have been able to locate.

Below is the output of Fastlane when it fails:

[14:05:24]: Running lane `ios create_appstore_all`. Next time you can do this by directly typing `fastlane ios create_appstore_all` 🚀.
[14:05:24]: Driving the lane 'ios create_appstore_all' 🚀
[14:05:24]: --------------------------------------
[14:05:24]: --- Disse er envfiler bliver kørt: ---
[14:05:24]: --------------------------------------
[14:05:24]: ▸ xxxxxxxxxxxxxx
[14:05:24]: ---------------------------------------------
[14:05:24]: --- Step: ruby ./assets/GenerateManual.rb ---
[14:05:24]: ---------------------------------------------
[14:05:24]: $ ruby ./assets/GenerateManual.rb
[14:05:24]: ▸ Opdaterer manual
[14:05:24]: ----------------------------------------------------------
[14:05:24]: --- Step: Switch to ios set_environment_variables lane ---
[14:05:24]: ----------------------------------------------------------
[14:05:24]: Cruising over to lane 'ios set_environment_variables' 🚖
[14:05:24]: Cruising back to lane 'ios create_appstore_all' 🚘
[14:05:24]: ------------------------------------------------
[14:05:24]: --- Step: Switch to ios create_appstore lane ---
[14:05:24]: ------------------------------------------------
[14:05:24]: Cruising over to lane 'ios create_appstore' 🚖
[14:05:24]: xxxxxxxxxxxxxxxxxxxxxxx
[14:05:24]: xxxxxxxxxxxxxxxxxxxxxxx
[14:05:24]: xxxxxxxxxxxxxxxxxxxxxxx
[14:05:24]: ---------------------
[14:05:24]: --- Step: produce ---
[14:05:24]: ---------------------

+----------------+------------------------------------------------------------+
|                         Summary for produce 2.204.3                         |
+----------------+------------------------------------------------------------+
[...]

Issue #20072, opened 2022-03-17
Unauthorized Access (Spaceship::UnauthorizedAccessError)


ronpelayo:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
Unauthorized Access on select_team

##### Command executed
```
Spaceship::Tunes.client.teams.each do |team|
teamId = team['contentProvider']['contentProviderId']
selected_team = Spaceship::Tunes.select_team(team_id:teamId.to_s)
```

##### Complete output when running fastlane, including the stack trace and command used

> /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/client.rb:941:in `handle_error': Unauthorized Access (Spaceship::UnauthorizedAccessError)
> 	from /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/client.rb:922:in `block in send_request'
> 	from /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/client.rb:677:in `with_retry'
> 	from /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/client.rb:918:in `send_request'
> 	from /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/client.rb:762:in `request'
> 	from /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/client.rb:165:in `team_id='
> 	from /Users/user/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/fastlane-2.204.3/spaceship/lib/spaceship/tunes/tunes_client.rb:87:in `select_team'
[...]

Issue #20071, opened 2022-03-16
Unauthorized access (upload_to_testflight)


my-name-is-nheo:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
`  desc "Build demo app"
  lane :demo do
    build_app(scheme: "work - Demo", workspace: "./ios/work.xcworkspace", include_bitcode: true)
    upload_to_testflight
  end`

bundle install runs great, asks for my apple id, when logging in with my email, erroring out with 'unauthorized access' error. I have account manager access to the appstoreconnect; tried changing password (signed back into xcode, and appstore) and still no luck. Anyone else having this issue?

##### Complete output when running fastlane, including the stack trace and command used

bundle install && bundle exec fastlane demo

[19:26:06]: Unauthorized Access

+------+----------------------+-------------+
|             fastlane summary              |
+------+----------------------+-------------+
| Step | Action               | Time (in s) |
+------+----------------------+-------------+
| 1    | default_platform     | 0           |
| 2    | build_app            | 155         |
| 💥   | upload_to_testflight | 33          |
+------+----------------------+-------------+

[19:26:06]: fastlane finished with errors

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                          |
| --------------------------- | -------------------------------------------------------------- |
| OS                          | 12.0                                                           |
| Ruby                        | 2.6.3                                                          |
| Bundler?                    | false                                                          |
| Git                         | git version 2.30.1 (Apple Git-130)                             |
| Installation Source         | /usr/local/bin/fastlane                                        |
| Host                        | macOS 12.0 (21A344)                                            |
| Ruby Lib Dir                | /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib |
| OpenSSL Version             | LibreSSL 2.8.3                                                 |
| Is contained                | false                                                          |
| Is homebrew                 | false                                                          |
| Is installed via Fabric.app | false                                                          |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/                    |
| Xcode Version               | 13.1                                                           |
| Swift Version               | 5.5.1                                                          |

[...]

PR #20081, opened 2022-03-17
[spaceship] Replace apple endpoints used for teams with olympus session

VrsajkovIvan33:
### Checklist

### Motivation and Context
Apple started returning "Unauthorized Access" errors when trying to do a login. Namely, we noticed that it was failing at the step where the `team_id` is either retrieved or set in the Spaceship client. Given the widespread reports that reverting back to Fastlane version 2.176.0 solved the issue, we assumed it had to be somehow connected with the [changes to the way the team id is managed in 2.177.0](https://github.com/fastlane/fastlane/pull/18279). 

Further investigation led us to observe that, when switching between teams/providers in App Store Connect, the following two endpoints are not used:
- `/WebObjects/iTunesConnect.woa/ra/user/detail`
- `/olympus/v1/providerSwitchRequests`

Instead, a GET and a POST request are used on `https://appstoreconnect.apple.com/olympus/v1/session`, respectively.

The providers themselves can now be found in the response to the GET request, under the `availableProviders` field. 

Furthermore, while examining the payload sent to Apple when switching between two providers, we determined that the only data that refers to the new provider in the payload is stored inside `"provider": {"providerId": ...}`, using the provider ID and not the public provider ID.

Taking this into account, we tried making a POST request to `https://appstoreconnect.apple.com/olympus/v1/session` while providing just the provider ID in the payload and the call has seemingly worked. Crucially, the `X-Requested-With` header needed to be provided with the value `olympus-ui` in order to make the call work. 

Potentially resolves:
- Unauthorized access (upload_to_testflight) [#20071](https://github.com/fastlane/fastlane/issues/20071)
- Unauthorized Access (Spaceship::UnauthorizedAccessError) [#20072](https://github.com/fastlane/fastlane/issues/20072)
- The request could not be completed because: Unauthorized Access [#20078](https://github.com/fastlane/fastlane/issues/20078)

And the other similar ones being reported.

### Description
**This is just a work in progress!** The main purpose of this PR is to help the Fastlane maintainers understand what is causing the current issue, after we have already come to these conclusions. The implementation is opinionated and any suggestion is very appreciated! 😃 

This PR modifies the way the teams are set + the way the user details data is fetched (and, by extension, the teams too) by changing the methods to target Apple's `https://appstoreconnect.apple.com/olympus/v1/session` endpoint instead.

The mapping is updated so that the current variables in the code can preserve their current role.

As mentioned above, the call to set the team is modified so that only the `provider` -> `providerId` is specified because this is the minimal solution we were able to find. We have observed that the call works even if we put the whole result from the GET request in the body of the POST.

### Testing Steps
Local checks performed show that no `Unauthorized Access` checks are being returned when trying to fetch or modify the team.

The tests in Fastlane have not yet been updated due to this being just a WIP.


snatchev:
This looks promising @VrsajkovIvan33. I am taking a look locally as well.

joshdholtz:
@VrsajkovIvan33 Thanks so much for this! I'm going to test this out now.

Quick question... do you mind giving me access to your fork or remaking this PR form a personal account? I can't add commits to this if its fork on an org 🙃

VrsajkovIvan33:
@joshdholtz We'd need to consult our legal team in order to give you access to our internal fork and that would take time. On the other hand, I can't re-create the PR with my personal account because I do not have CLA set up to contribute. 😞 

But! Since time is of the essence, what about you just copying the changes from this PR as is and using it in any way you seem fit in your official fix, while closing this PR?

We don't mind giving up the contribution if it means that you will be able to prepare a fix ASAP, which is in everyone's interest. 😃

joshdholtz:
@VrsajkovIvan33 You are hero! I can do that but I will make sure you get full credit ❤️ 

I will also talk to GitHub to see if we can get this org fork thing changed 😊

joshdholtz:
@VrsajkovIvan33 @valerio-castelli Thank you both so much! Do either of you have Twitter handles I can mention you in when a release gets made?

joshdholtz:
@VrsajkovIvan33 I've created #20083 with your commit in it! Do you mind commenting in there saying something like "you approve of this"? 🙃 Just for the record 🤷‍♂️

VrsajkovIvan33:
@joshdholtz Left the comment on the new PR as asked, my Twitter handle should be `VrsajkovIvan33`. Happy to have been of help, big fan of your work! 💪

PR #20083, opened 2022-03-17
[spaceship] replace teams endpoints

joshdholtz:
Duplicated from #20081 (with @VrsajkovIvan33 permission) so that I could add commits

---

👇 This was from the description in #20081

### Motivation and Context
Apple started returning "Unauthorized Access" errors when trying to do a login. Namely, we noticed that it was failing at the step where the `team_id` is either retrieved or set in the Spaceship client. Given the widespread reports that reverting back to Fastlane version 2.176.0 solved the issue, we assumed it had to be somehow connected with the [changes to the way the team id is managed in 2.177.0](https://github.com/fastlane/fastlane/pull/18279). 

Further investigation led us to observe that, when switching between teams/providers in App Store Connect, the following two endpoints are not used:
- `/WebObjects/iTunesConnect.woa/ra/user/detail`
- `/olympus/v1/providerSwitchRequests`

Instead, a GET and a POST request are used on `https://appstoreconnect.apple.com/olympus/v1/session`, respectively.

The providers themselves can now be found in the response to the GET request, under the `availableProviders` field. 

Furthermore, while examining the payload sent to Apple when switching between two providers, we determined that the only data that refers to the new provider in the payload is stored inside `"provider": {"providerId": ...}`, using the provider ID and not the public provider ID.

Taking this into account, we tried making a POST request to `https://appstoreconnect.apple.com/olympus/v1/session` while providing just the provider ID in the payload and the call has seemingly worked. Crucially, the `X-Requested-With` header needed to be provided with the value `olympus-ui` in order to make the call work. 

Potentially resolves:
- Unauthorized access (upload_to_testflight) [#20071](https://github.com/fastlane/fastlane/issues/20071)
- Unauthorized Access (Spaceship::UnauthorizedAccessError) [#20072](https://github.com/fastlane/fastlane/issues/20072)
- The request could not be completed because: Unauthorized Access [#20078](https://github.com/fastlane/fastlane/issues/20078)

And the other similar ones being reported.

### Description
**This is just a work in progress!** The main purpose of this PR is to help the Fastlane maintainers understand what is causing the current issue, after we have already come to these conclusions. The implementation is opinionated and any suggestion is very appreciated! 😃 

This PR modifies the way the teams are set + the way the user details data is fetched (and, by extension, the teams too) by changing the methods to target Apple's `https://appstoreconnect.apple.com/olympus/v1/session` endpoint instead.

The mapping is updated so that the current variables in the code can preserve their current role.

As mentioned above, the call to set the team is modified so that only the `provider` -> `providerId` is specified because this is the minimal solution we were able to find. We have observed that the call works even if we put the whole result from the GET request in the body of the POST.

### Testing Steps
Local checks performed show that no `Unauthorized Access` checks are being returned when trying to fetch or modify the team.

The tests in Fastlane have not yet been updated due to this being just a WIP.


fastlane-bot-helper:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

isaac-tribal:
It worked for me !!

image

image

spaceship/spec/tunes/fixtures/user_detail.json
```diff
@@ -14,12 +14,10 @@
...
     "userName": "foo@bar.com",
-    "associatedAccounts": [{
-            "contentProvider": {
-                "contentProviderId": 1234,
-                "name": "Joshua Liebowitz",
-                "contentProviderTypes": ["Purple Software"]
-            },
+    "availableProviders": [{
+            "providerId": 1234,
+            "name": "Joshua Liebowitz",
```
taquitos, line 19: 😄

taquitos:
Nice job y'all.

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -950,13 +949,6 @@ def ref_data
       Spaceship::Tunes::AppVersionRef.factory(data)
     end
 
-    # Fetches the User Detail information from ITC. This gets called often and almost never changes
-    # so we cache it
-    # @return [UserDetail] the response
-    def user_detail_data
```
taquitos, line 956: Oh, this isn't even really used. Smart.

taquitos (approved):
Makes sense.

spaceship/lib/spaceship/client.rb
```diff
@@ -70,10 +70,10 @@ def self.hostname
 
     # @return (Array) A list of all available teams
     def teams
-      user_details_data['associatedAccounts'].sort_by do |team|
```
snatchev, line 73: does this just not exist anymore?

spaceship/lib/spaceship/client.rb
```diff
@@ -124,8 +124,8 @@ def teams
     #  "userName"=>"detlef@krausefx.com"}
     def user_details_data
       return @_cached_user_details if @_cached_user_details
-      r = request(:get, '/WebObjects/iTunesConnect.woa/ra/user/detail')
-      @_cached_user_details = parse_response(r, 'data')
+      r = request(:get, "https://appstoreconnect.apple.com/olympus/v1/session")
```
snatchev, line 127: We already make this request here https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L562

Maybe have that use `user_details_data`?

spaceship/lib/spaceship/tunes/tunes_client.rb
```diff
@@ -1582,9 +1574,12 @@ def clear_user_cached_data
       @sso_token_for_video = nil
     end
 
-    # the contentProviderIr found in the UserDetail instance
+    # the contentProviderId found in the user details data
     def content_provider_id
-      @content_provider_id ||= user_detail_data.content_provider_id
+      provider = user_details_data["provider"]["providerId"]
+      if provider
+        @content_provider_id ||= provider.to_s
```
snatchev, line 1581: AFAICT, this method doesn't use the memoized value.

snatchev (approved):
💪 Strong work as usual. Just a few comments.

spaceship/lib/spaceship/client.rb
joshdholtz: Nope! It got renamed 🤷‍♂️

spaceship/lib/spaceship/tunes/tunes_client.rb
joshdholtz: Do you mean like "it doesn't get used as the return value"? 🤔 

If so, the last line in the if statement becomes the return value in the if statement which will implicitly be the return value of the method but...

I also don't love how this method looks. I'll clean it up!

spaceship/lib/spaceship/client.rb
joshdholtz: Ohhh, we do call this already! `user_details_data` caches responses and the `fetch_olympus_session` does not so I think we would need to do mesh them together some how 🤔 

I would like to combine them but I might punt on that to a future PR because I'm scared of making that change tonight 😅

georgelee114:
just wanted to say thank you for this fix, when I cloned and tried it with my local project, it worked on my end, please get this fix merged in asap!

joshdholtz:
@georgelee114 I'm working on it 😊 I'm doing a bunch more local testing to make sure 100% that everything is covered

spaceship/lib/spaceship/client.rb
valerio-castelli: Indeed, we noticed that `fetch_olympus_session` also called this endpoint, but the authentication logic of fastlane is quite complex and we didn’t feel comfortable doing that change, at least not at this stage, as the implications of dealing with it and with caching weren’t immediately clear 🙏

onimitch:
Thank you everyone for the fixes!

galahador:
Thank you all!

PR #20090, opened 2022-03-18
Added case-sensitive .DS_Store to gitignore

philprime:
### Checklist

### Motivation and Context

I am using a case-sensitive file-system for development and therefore macOS is creating `.DS_Store` files, which should not be included in git.

### Description

Created an APFS container with case-sensitive file system. Ran `git status` to see many `.DS_Store` files. Changed `.gitignore` to include the right casing, ran `git status` again and the changes are ignored.

Currently only `.DS_STORE` files are ignored, which are not sufficient on a case-sensitive system.

### Testing Steps

See the description above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ coverage/
 rspec_logs.json
 test_output/
 .DS_STORE
+.DS_Store
 .keys
 *.swp
 


Kukurijek:
MacOS is not a case sensitive file system by default, but `DS_STORE` does not really exist, so I think it should be just renamed from `DS_STORE` to `DS_Store`.
Good point 👍

joshdholtz (approved):
I had no idea our `.gitignore` had a all caps version of this 🙈  This seems good and better to me! Thank you 😊

PR #20105, opened 2022-03-23
[trainer] Fix missing test failures and improve junit format

tahirmt:
### Checklist

### Motivation and Context

Bringing changes from https://github.com/fastlane-community/trainer/pull/40 by @robnadin and improving the `junit` format.

This is from the original PR:

------------
* Fix xcpretty_naming for xcresult
* Add test result object hierarchy
* Refactored group and name parsing function

These fixes came about due to the way the test result format has changed between Xcode 10 and 11, and how the logic to parse xcresult bundles in Trainer was not finding any of our failing tests.

In our use case, we were finding that test failures were not being parsed correctly because we were grouping our tests from a collection of other test suites like so:

```
class AllTests: XCTestCase {
    
    override class var testCaseClasses: [XCTestCase.Type] {
        return [
            TestsA.self,
            TestsB.self,
            TestsC.self,
        ]
    }
    
    override class var defaultTestSuite: XCTestSuite {
        let testSuite = XCTestSuite(forTestCaseClass: self)
        let testCases = testCaseClasses.flatMap { testCaseClass in
            testCaseClass.testInvocations.map { testCaseClass.init(invocation: $0) }
        }
        testCases.forEach(testSuite.addTest)
        return testSuite
    }
}

...

class TestsA: XCTestCase {
    
    func testFoo() {
        XCTAssertTrue(false)
    }
}

```
This results in find_failure incorrectly comparing AllTests/testFoo to TestsA/testFoo and returning false due to the identifier and sanitized_test_case_name values not matching in the test metadata.

Instead we should call xcresulttool get --format json and pass it the id we get from the summary reference in the metadata to get an accurate representation of the test failure summary. The benefit of this approach is that we also get explicit file name and line number information that we are able to format in exactly the same way as Trainer has been doing with the older test_result bundles in Xcode 10 and earlier.

----------
### Description

In addition to the changes from the PR described above, I updated the junit template because the current template produces a file name within the failure message which, isn't using the correct convention of how a file path and line number should be defined in junit format and therefore breaks our reporting on GitHub CI to report on the correct line number. 

### Testing Steps
Run trainer against an `xcresult` file with failures and expect the correct file and line numbers reported. These numbers should match what is seen in Xcode if you open the `xcresult` in Xcode directly as shown below for the test file.

Screen Shot 2022-03-23 at 12 42 47 AM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -28,6 +28,7 @@ PATH
       multipart-post (~> 2.0.0)
       naturally (~> 2.2)
       optparse (~> 0.1.1)
+      parallel (>= 1.17.0)
       plist (>= 3.1.0, < 4.0.0)
       rubyzip (>= 2.0.0, < 3.0.0)
       security (= 0.1.3)
@@ -391,4 +392,4 @@ DEPENDENCIES
   yard (~> 0.9.11)
 
 BUNDLED WITH
-   2.2.27
+   2.3.8
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -113,6 +113,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('google-cloud-storage', '~> 1.31') # Access Google Cloud Storage for match
   spec.add_dependency('emoji_regex', '>= 0.1', '< 4.0') # Used to scan for Emoji in the changelog
   spec.add_dependency('aws-sdk-s3', '~> 1.0') # Used for S3 storage in fastlane match
+  spec.add_dependency('parallel', '>= 1.17.0') # Used for trainer
 
   # Development only
   spec.add_development_dependency('rake')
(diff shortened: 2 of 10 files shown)


joshdholtz (approved):
This looks so good! Thank you for bringing this PR over 🥰

joshdholtz:
I actually need to revert this. I tested this on another project after merging and it appears that this doesn't expand all test suites.

joshdholtz:
@tahirmt Hey! Do you mind reopening this PR? I found some issues after I merged so I reverted but I can't reopen 😔 

Sorry about this! I don't exactly know where the issues were but I'll try to diagnose and add some commits to fix it!

tahirmt:
> @tahirmt Hey! Do you mind reopening this PR? I found some issues after I merged so I reverted but I can't reopen 😔
> 
> Sorry about this! I don't exactly know where the issues were but I'll try to diagnose and add some commits to fix it!

@joshdholtz of course I can reopen it in a bit. If you can let me know of the issues maybe I can help there as well.

tahirmt:
@joshdholtz opened #20160

PR #20145, opened 2022-04-04
[match] Add support for Developer ID certificates from G2 Sub-CA

triplef:
### Checklist

### Motivation and Context

Developer ID certificates can now be generated using one of two different intermediary CAs, and Fastlane currently doesn’t support the newer G2 Sub-CA. Following is the description from the [certification creation page](https://developer.apple.com/account/resources/certificates/add) on the Apple Developer portal when selecting "Developer ID Application":

> Certificates issued from the Developer ID G2 intermediary certificate authority (Sub-CA) are supported by Xcode 11.4.1 or later. If you need to sign software for distribution on an earlier release, you can create a certificate associated to the previous Sub-CA and turn off Automatic Signing in Xcode. Any certificates created after Feb 01, 2022 and associated to the previous Sub-CA will expire on Feb 01, 2027.

### Description

- Adds a new `DEVELOPER_ID_APPLICATION_G2` certificate type
- Updates `match` and `sigh` to find Developer ID certificates using this new type
- Updates `cert` to generate Developer ID certificates using this new type

I used `match` to generate and find a new Developer ID certificate.


CaseyCw1:
Thank you for posting that it could be just the thing to give inspiration to someone who needs it! Keep up the great work!

triplef:
Thanks! Btw. the CLA is signed.

joshdholtz (changes requested):
Thanks for making this! Did you happen to test this to see how this behaves with App Store Connect API?

The official docs don't show this as a value (yet) - https://developer.apple.com/documentation/appstoreconnectapi/certificatetype

I don't _think_ you can create a Developer ID while using the API Key but wanted to check anyway 🙃

triplef:
Thanks for your feedback @joshdholtz!

Could you give me a hint how to test this with the ASC API? It’s a bit tricky to test this as you can only have one Developer ID cert, and I think you can’t even revoke it. Any guidance how to deal with this would be appreciated.

joshdholtz (approved):
This looks great! I tested this with both Apple ID auth and API Key and works as expected 💪 I also added a commit so that it will still list cert of both the new and old type but will only create the new one.

Really appreciate the contribution 🥰

Issue #20165, opened 2022-04-12
match+gym not working as expected


tejassharma96:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

I'm trying to set up my pipeline to use fastlane match+gym to build my app, but consistently receive an error stating 
```
error: "<TARGET NAME REDACTED>" requires a provisioning profile with the App Groups feature. Select a provisioning profile in the Signing & Capabilities editor. (in target '<TARGET NAME REDACTED>' from project '<PROJECT NAME REDACTED>')
```
I was expecting match to automatically set all of that up. When I go into my xcode project and manually set the provisioning profile for my targets to the profiles installed by match, and then run gym, it works as expected and successfully builds and archives the app. If I don't manually do that, it fails with the message above.

Based on http://docs.fastlane.tools/codesigning/xcode-project/, it seems like this might be intended behaviour. Would you be open to me creating a pull request to go through the provisioning profiles from match and update the targets via fastlane before running gym? I think it would be a much cleaner experience to be able to just say 
```
lane :appstore-build do
  match(type: 'appstore')
  gym(export_method: 'app-store')
end
lane :enterprise-build do
  match(type: 'enterprise')
  gym(export_method: 'enterprise')
end
```
and have that work without needing to manually update provisioning profiles in xcode between running the two lanes.

I think I have pretty much everything I need to get that working and create a PR, except for the ability to get the certificate name from match. Would appreciate any support on figuring out how to get that value from match, but if we can do that, then I would be happy to update the actions and create a PR to set that all up automatically without needing any intervention.

tejassharma96:
I know that tagging you is a bit rude, so apologies in advance, but @joshdholtz you've been very helpful with my PRs in the past so would love to hear if this is something you would be interested in getting added to fastlane, and if you could help me with getting the certificate name. Thanks!

jenne045:
I am also facing exact same issue. Is your issue resloved? any recommendation how to solve this issue.? Can any one please help?

jenne045:
This works really well for us, thank you!	Facing same issue here. Help is appreciated. 
[...]

PR #20187, opened 2022-04-17
[match] extract the certificate name from provisioning profiles

tejassharma96:
### Checklist

### Motivation and Context

Helps with #20165 (would not close it, but it would help make those updates)
I'd like to be able to automatically configure the code sign identity, provisioning profile name, and team id for my targets when using different provisioning profile types (eg enterprise vs appstore).

While we already have ways to get the profile name and team id via match, I've had to write additional code to extract the certificate name. This PR parses the certificate name during the match steps and populates an environment variable to make that name accessible.

### Description

- Added `Match::Util.environment_variable_name_certificate_name` to generate an env variable with the postfix "certificate-name". 
- Updated `Match::Util.get_cert_info` to allow it to parse the certificate name directly out of certificate data rather than just the certificate path
- Update `Match::Runner.fetch_provisioning_profile` to make it use the updated util functions and populate an environment variable with the certificate name associated with each provisioning profile

### Testing Steps

Try running match and installing some profiles, and check that the certificate name associated with each provisioning profile is populated.

### DIFF
--- a/match/spec/runner_spec.rb
+++ b/match/spec/runner_spec.rb
@@ -96,6 +96,7 @@ describe Match do
           expect(spaceship).to receive(:certificates_exists).and_return(true)
           expect(spaceship).to receive(:profile_exists).and_return(true)
           expect(spaceship).to receive(:bundle_identifier_exists).and_return(true)
+          expect(Match::Utils).to receive(:get_cert_info).and_return([["Common Name", "fastlane certificate name"]])
 
           Match::Runner.new.run(config)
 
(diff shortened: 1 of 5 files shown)


match/spec/runner_spec.rb
```diff
@@ -178,6 +181,8 @@
           expect(spaceship).to receive(:certificates_exists).and_return(true)
           expect(spaceship).to receive(:profile_exists).and_return(true)
           expect(spaceship).to receive(:bundle_identifier_exists).and_return(true)
+          expect(Match::Utils).to receive(:get_cert_info)
```
getaaron, line 184: I think you don’t need this line

getaaron (commented):
Just one suggestion otherwise LGTM

match/spec/runner_spec.rb
tejassharma96: Uhh been a while since I created this PR so I'm don't remember 100%, but I think I do need it - iirc this flow calls `get_cert_info` twice. plus wouldn't the tests be failing if there were an extraneous `expect`?

getaaron: yea, you're right, for some reason i was reading it like it was `allow` and not `expect` 👍

getaaron (approved):

PR #20194, opened 2022-04-19
[match] Check validity of provisioning profile certificates when `force_for_new_certificates` option is provided

nekrich:
### Checklist

### Motivation and Context
When a certificate expires (not revoked) provisioning profile stays valid.
And if we regenerate certificate count will not differ:
* For portal certificates, we filter out the expired one but includes a new certificate;
* Profile still contains an expired certificate and is valid.
Thus, we need to check the validity of profile certificates too.

It's my bad that I wasn't counting that the provisioning profile stays valid when certificates expire.

### Description
Simplty add `.select(&:valid?)` 😅

### Testing Steps
* Get the certificate contained in the provisioning profile expired 🤷
* `match(force_for_new_certificates: true)`

### DIFF
--- a/match/lib/match/runner.rb
+++ b/match/lib/match/runner.rb
@@ -419,7 +419,12 @@ module Match
 
       return false unless portal_profile
 
-      profile_certs_count = portal_profile.fetch_all_certificates.count
+      # When a certificate expires (not revoked) provisioning profile stays valid.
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Ohhhhhh, this is a great catch! Thanks for fixing this 🙌

PR #20211, opened 2022-04-23
[aciton][setup_ci] added timeout param into setup_ci

javigines:
### Checklist

### Motivation and Context

I have been facing a problem in my ci-cd system where with a long Fastlane run the keychain created with the action setup_ci after and hour start giving problems on the execution of the command codesigning. It seems to happen because the timeout of the keychain created on the setup_ci action is hardcoded to 3600, it seems logic and useful to me to have an option to pass a parameter to the action with which we could increase/decrease or set as no timeout to this timeout option. So I have done my best to add an optional parameter to setup_ci with is `timeout` that have the same behavior than the timeout from the create_keychain action that is used under the hood.

### Description

Add new timeout optional parameter to setup_ci action to customize the timeout of the keychain that is created under the hood. By default this parameter sets as right now is behaving with the 3600 seconds.

### Testing Steps

Execute the `setup_ci` action with a timeout parameter value.

### DIFF
--- a/fastlane/lib/fastlane/actions/setup_ci.rb
+++ b/fastlane/lib/fastlane/actions/setup_ci.rb
@@ -12,7 +12,7 @@ module Fastlane
           setup_output_paths
         end
 
-        setup_keychain
+        setup_keychain(params)
       end
 
       def self.should_run?(params)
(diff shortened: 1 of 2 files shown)


joshdholtz (approved):
This is great! I actually assumed this was already an option but I guess it wasn't 😱 

Thank you for adding!

PR #20214, opened 2022-04-25
[infra] changelog will now show github usernames which makes contributors stand out more in releases

joshdholtz:
### Motivation and Context
GitHub releases will now contributors in a special section in a release when their username's are tagged in the release notes

### Description
- Replaced the git command that was previously generating release notes
- New process:
  - Gets commits, message, and name (similar to previous implementation) from GitHub compare API
  - Searches for commit SHA in fastlane repo that was merged into master to generate changelog message from::
    - PR title
    - PR author username
    - PR number

This goal of this new process is to:
1. Use PR title as changelog instead of commit message
2. Also mention the username

## Example

I updated the last release with the changeling generated by this PR

Screen Shot 2022-04-24 at 9 54 03 PM

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -343,12 +343,47 @@ private_lane :send_mac_app_ci_reminder do
   end
 end
 
+desc "Generate changelog from GitHub compare and PR data for mentioning GitHub usernames in release notes"
+private_lane :github_changelog do |options|
+  old_version = options[:old_version] || current_version
+  path = "/repos/fastlane/fastlane/compare/#{old_version}...HEAD"
+
+  # Get all commits from previous version (tag) to HEAD
+  resp = github_api(path: path)
+  body = JSON.parse(resp[:body])
+  commits = body["commits"].reverse
+
+  formatted = commits.map do |commit|
+    # Default to commit message info
+    message = commit["commit"]["message"].lines.first.strip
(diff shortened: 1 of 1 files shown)


lucgrabowski:
Looks great 💪
I understand that we do not expect `[WIP]` in PR messages, which were filtered out previously.

fastlane/Fastfile
```diff
@@ -343,12 +343,47 @@ private_lane :send_mac_app_ci_reminder do
...
+
+  formatted
+end
+
 desc "Print out the changelog since the last tagged release and open the GitHub page with the changes"
 lane :show_changelog do |options|
   old_version = options[:old_version] || current_version
 
-  changes = sh("git log --pretty='* %s via %aN' #{old_version}...HEAD --no-merges ..", log: $verbose).gsub("\n\n", "\n")
-  changes.gsub!("[WIP]", "") # sometimes a [WIP] is merged
```
lucgrabowski, line 351: I understand that we do not expect [WIP] in PR messages anymore

lucgrabowski (approved):
Looks great on releases page 💪

fastlane/Fastfile
joshdholtz: Nope nope! And if they are there, I'll just manually change them or delete them in the changeling 😛

Issue #14500, opened 2019-03-29
match with google cloud: allow disabling the application default credentials prompt


jfaust:
### Feature Request
Right now, using application default credentials with GCP and match, every time I run a match operation it asks:
```
[11:07:58]: Do you want to use this system's Google Cloud application default keys? (y/n)
```

This is unnecessary, in what is otherwise a command that requires no user input.

#### Motivation Behind Feature
We run `match` in various developer scripts, and are attempting to switch from git to GCP - this effectively blocks the transition.

#### Feature Description
Add an option to `match`, something like `google_cloud_prompt_for_application_credentials(false)`.

Alternatively, just remove the confirmation prompt entirely, as I'm not sure what benefit it provides.

#### Alternatives or Workarounds
`yes | fastlane match ...`

jfaust:
@mattmckeon looks like you did the work to enable application credentials - thoughts? I'm happy to get a PR going but am not sure the preferred solution.

jfaust:
@fastlane-bot this is still an issue... I'm waiting on guidance for the preferred solution

PR #20223, opened 2022-04-28
[match] add option to skip google account confirmation

trsneed:
### Checklist

#### Why?
We are using google secrets manager for our certificates, and are adding a CD layer to prevent delivering to the AppStore from developer machines. When running fastlane in `non-interactive mode` on our CI/CD servers we cannot answer the question to use the system's google account, configured with gcloud. By using a service worker with the gcloud `auth` GitHub action we can auth the service user on a server and bypass the need for a gc_keys.json file. 

**Resolves** #14500

### Description
Adds a `match` parameter to skip the optional confirmation step.

#### Testing
- Set my environment to `CI` and saw the issue with the workflow. 
- Added an optional `skip_google_cloud_account_confirmation` match parameter that defaults to false. 
- Set the parameter to `true`, saw no confirmation prompt.
- Set the parameter to `false`, saw the confirmation prompt.


joshdholtz (approved):
This is such a cool use case! I would have never thought about this 😅 But thanks for the contribution! Really appreciate it 🥰

PR #20237, opened 2022-04-30
[scan] fix scan failing to return results when there are test failures

### LINKED ISSUE #20236: [scan] Scan does not return output when there are test failures
### New Issue Checklist

### Issue Description

When running scan with `fail_build` set to `false` the result is not returned as an output. Instead only a nil object is received

##### Command executed

Sample from `FastFile`
```
# running with failing tests
scan_output = scan(fail_build: false)
# expecting `scan_output` to not be nil but received nil
```

#### Details

On [this line](https://github.com/fastlane/fastlane/blob/156f2114d19f7430a6a1d4c01b1c1e958d064830/scan/lib/scan/runner.rb#L308) it looks like `UI.test_failure!` is called which raises an exception and that exception is caught when `fail_build` is false so it doesn't fail the step but because an exception was raised [Line 316](https://github.com/fastlane/fastlane/blob/156f2114d19f7430a6a1d4c01b1c1e958d064830/scan/lib/scan/runner.rb#L316) was never called so no output is returned.

Thank you @kamvoick for [pointing it out](https://github.com/fastlane/fastlane/pull/19860#issuecomment-1112815390).

tahirmt:
When using test_failure! method it causes scan to skip sending the output even if `fail_build` is set to false

### Checklist

### Motivation and Context
Resolves #20236

### Description
* While the exception is caught when `fail_build` option is set to false, it still causes the results to not be sent to consumers.
* To fix this I moved the `fail_build` definition from `run_tests` to `scan` and then used the value of `fail_build` to determine whether to raise the exception or to just post a ui error.

### Testing Steps

Added a unit test to confirm that with `fail_build` set to false it returns the output from the action and that the output contains the correct values.

### DIFF
--- a/fastlane/lib/fastlane/actions/run_tests.rb
+++ b/fastlane/lib/fastlane/actions/run_tests.rb
@@ -71,13 +71,7 @@ module Fastlane
       def self.available_options
         require 'scan'
 
-        FastlaneCore::CommanderGenerator.new.generate(Scan::Options.available_options) + [
-          FastlaneCore::ConfigItem.new(key: :fail_build,
-                                       env_name: "SCAN_FAIL_BUILD",
-                                       description: "Should this step stop the build if the tests fail? Set this to false if you're using trainer",
-                                       type: Boolean,
-                                       default_value: true)
-        ]
+        FastlaneCore::CommanderGenerator.new.generate(Scan::Options.available_options)
       end
 
       def self.output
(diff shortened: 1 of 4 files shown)


joshdholtz (approved):
Thanks so much for catching this! I added one commit to also add a check inside of the exit status code check to not raise an error. I tested this against my sample project and I was able to get the results returned as expected for when `fail_build` was set to both `true` and `false`.

Really appreciate the contribution! ❤️

PR #20247, opened 2022-05-06
[deliver] add `verify_only` option to deliver

polpielladev:
### Checklist

### Motivation and Context
Recently, we have been trying to modernise our CI integration with fastlane and we are looking at using `deliver` (and other similar fastlane actions) instead of `altool` in combination with an App Store Connect key to handle all our binary upload operations.

One thing that we find useful in `altool` that we couldn't find in `deliver` is the `--validate` flag, which allows us to run nighly checks on our codebase and flag issues such as embedding duplicate dependencies, problems with our certificates, etc. We thought it would be awesome if we could have the same kind of flag in the `deliver`, which would not upload anything to ASC but do a validation instead and, following a [discussion](https://github.com/fastlane/fastlane/discussions/20204) with @lucgrabowski, we thought this would be the best approach.

### Description
Adds a new command to the iTunes transporter file which verifies a given binary with App Store Connect. This command is then used by the new `deliver` `verify` function, which is exposed as an _off-by-default_ flag to consumers. 

### Testing Steps
The easiest way to test these changes is to generate an invalid archive (for example using a beta version of Xcode) and then running:
`bundle exec fastlane deliver --verify --ipa <path_to_your_api>` 

This is the expected output:
CleanShot 2022-05-06 at 11 53 47@2x


joshdholtz (approved):
This is 💯 I love this option. And thank you for the tests! I did make one small change to rename the option from `verify` to `verify_only` to match some of the there options we have that do a similar type of thing. I just wanted to call out that it will *only* verify 😊 

Really appreciate the contribution!

pol-piella:
> This is 💯 I love this option. And thank you for the tests! I did make one small change to rename the option from `verify` to `verify_only` to match some of the there options we have that do a similar type of thing. I just wanted to call out that it will _only_ verify 😊
> 
> Really appreciate the contribution!

Thanks so much @joshdholtz and @lucgrabowski! The rename to `verify_only` sounds great to me too! 🙌 

Was very fun to make the contribution and get to know a bit more about how fastlane works internally. Looking forward to doing more contributions in the future 🚀 😄

PR #20251, opened 2022-05-09
[spaceship] update `Spaceship::ConnectAPI::User` model to include `delete!` method

liamnichols:
### Checklist

### Motivation and Context

I've been using fastlane and Spaceship to script some batch operations for managing our App Store Connect user access since the web-ui kind of sucks for anything other than one-off changes 😄 

While doing so, I needed to delete some old accounts but noticed that while `Spaceship::ConnectAPI::Users::API` defines a `delete_user` method, it's not exposed on `Spaceship::ConnectAPI::User model` like it is for `UserInvitation`.

https://github.com/fastlane/fastlane/blob/eb4584e0394c1fecccd70d4313260f4707f2af80/spaceship/lib/spaceship/connect_api/users/users.rb#L26-L29

### Description

In this change, I define a new `delete!` method in an almost identical way to `UserInvitation.delete!`

### Testing Steps

N/A

### DIFF
--- a/spaceship/lib/spaceship/connect_api/models/user.rb
+++ b/spaceship/lib/spaceship/connect_api/models/user.rb
@@ -62,6 +62,11 @@ module Spaceship
         resp = client.get_user_visible_apps(user_id: id, limit: limit)
         return resp.to_models
       end
+
+      def delete!(client: nil)
+        client ||= Spaceship::ConnectAPI
+        client.delete_user(user_id: id)
+      end
     end
   end
 end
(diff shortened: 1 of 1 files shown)


spaceship/lib/spaceship/connect_api/models/user.rb
```diff
@@ -62,6 +62,11 @@ def get_visible_apps(client: nil, limit: nil)
         resp = client.get_user_visible_apps(user_id: id, limit: limit)
         return resp.to_models
       end
+
+      def delete!(client: nil)
```
lucgrabowski, line 66: Could this method be moved above `get_visible_apps`?

lucgrabowski (approved):
Thanks for adding this :) LGTM 🔥

spaceship/lib/spaceship/connect_api/models/user.rb
liamnichols: Done 🙂

joshdholtz (approved):
This looks great! Thank you so much for adding this 😊

PR #20293, opened 2022-05-18
[fastlane] search plugins in bundler transitive dependencies as well

lacostej:
### Checklist

### Motivation and Context
Today we use import_from_git to import a shared Fastfile/actions. This has several drawbacks, one of which being that we must add all the plugins this shared fastfile uses inside all projects that use `import_from_git`.

We wish to convert this shared repository into a Gem, and add all required plugins as gem dependency to the gem. 

Unfortunately right now the plugin search doesn't look at transitive dependencies. So this PR fixes this.

The next step is to add another feature `import_from_gem` as alternative to `import_from_git` (see #20294).

### DIFF
--- a/fastlane/lib/fastlane/plugins/plugin_manager.rb
+++ b/fastlane/lib/fastlane/plugins/plugin_manager.rb
@@ -23,6 +23,12 @@ module Fastlane
       nil
     end
 
+    def lockfile_path
+      Bundler::SharedHelpers.default_lockfile.to_s
+    rescue Bundler::GemfileNotFound
+      nil
+    end
+
     def pluginfile_path
       if FastlaneCore::FastlaneFolder.path
         return File.join(FastlaneCore::FastlaneFolder.path, PLUGINFILE_NAME)
@@ -54,8 +60,9 @@ module Fastlane
     # Returns an array of gems that are added to the Gemfile or Pluginfile
     def available_gems
       return [] unless gemfile_path
-      dsl = Bundler::Dsl.evaluate(gemfile_path, nil, true)
-      return dsl.dependencies.map(&:name)
+      dsl = Bundler::Dsl.evaluate(gemfile_path, lockfile_path, true)
+      dependencies = collect_bundler_runtime_dependencies([], dsl).uniq
+      return dependencies.map(&:name)
     end
 
     # Returns an array of fastlane plugins that are added to the Gemfile or Pluginfile
@@ -393,6 +400,22 @@ module Fastlane
         version_number: version_number,
         actions: references
       }
+    end
+
+    private
+
+    # recursively collect runtime dependencies
+    def collect_bundler_runtime_dependencies(collection, source)
+      runtime_deps = source.dependencies.select { |d| d.type == :runtime }
+      # prevent potential infinite recursion. Should never happen. Better safe than sorry.
+      runtime_deps -= collection
+      collection.concat(runtime_deps)
+      runtime_deps.each do |d|
+        collect_bundler_runtime_dependencies(collection, d.to_spec)
+      rescue Gem::MissingSpecError
+        # ignoring unresolvable dependencies
+      end
+      collection
     end
   end
 end
--- a/fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
+++ b/fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
@@ -6,7 +6,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 Gem::Specification.new do |spec|
   spec.name        = 'dependency_with_plugins'
   spec.version     = '1.0.0'
-  spec.authors     = ["Fastlane team"]
+  spec.authors     = ["fastlane team"]
 
   spec.required_ruby_version = '>= 2.6'
 
--- a/fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
+++ b/fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+lib = File.expand_path('lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+
+Gem::Specification.new do |spec|
+  spec.name        = 'dependency_with_plugins'
+  spec.version     = '1.0.0'
+  spec.authors     = ["Fastlane team"]
+
+  spec.required_ruby_version = '>= 2.5'
+
+  spec.summary = "fake gem for fastlane tests"
+
+  spec.add_dependency('fastlane')
+  spec.add_dependency('fastlane-plugin-appcenter')
+end
--- a/fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
+++ b/fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
@@ -8,7 +8,7 @@ Gem::Specification.new do |spec|
   spec.version     = '1.0.0'
   spec.authors     = ["Fastlane team"]
 
-  spec.required_ruby_version = '>= 2.5'
+  spec.required_ruby_version = '>= 2.6'
 
   spec.summary = "fake gem for fastlane tests"
 
--- a/fastlane/spec/plugins_specs/plugin_manager_spec.rb
+++ b/fastlane/spec/plugins_specs/plugin_manager_spec.rb
@@ -26,6 +26,7 @@ describe Fastlane do
 
       it "returns all fastlane plugins with no fastlane_core" do
         allow(Bundler::SharedHelpers).to receive(:default_gemfile).and_return("./fastlane/spec/fixtures/plugins/Pluginfile1")
+        allow(Bundler::SharedHelpers).to receive(:default_lockfile).and_return(nil)
         expect(plugin_manager.available_gems).to eq(["fastlane-plugin-xcversion", "fastlane_core", "hemal"])
       end
     end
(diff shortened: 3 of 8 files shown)


lacostej:
I don't have the proper set of unit tests to cover this feature yet.

Note that I think some of the plugin manager Unit tests are 'fake'.  In particular they load Gemfiles that reference the "fastlane-plugin-xcversion" plugin and this plugin doesn't exist.

The new code tries to resolve those entries (and their dependencies) but fails. The current work-around is to ignore the failures as I would assume those issues should be resolved by the environment before fastlane kicks in.

Better would be to mock the results (call to `.dependencies` and  `.to_spec`) and add unit tests for the transitive scenario I am adding.

Before I do this, I would appreciate feedback on the intent.

@joshdholtz maybe?

lacostej:
I've added an integration test to show how the feature works. Main drawback is that it requires the gems to be installed for the resolution to find the transitive dependencies and the current. I obviously don't want to leave it like this.

fastlane/lib/fastlane/plugins/plugin_manager.rb
```diff
@@ -394,5 +401,19 @@ def store_plugin_reference(gem_name)
...
       }
     end
+
+    private
+
+    # recursively collect runtime dependencies
+    def collect_bundler_runtime_dependencies(collection, source)
+      runtime_deps = source.dependencies.select { |d| d.type == :runtime }
+      collection.concat(runtime_deps)
+      runtime_deps.each do |d|
```
lacostej, line 411: it's possible to return to the previous non recursive behavior by disabling this loop. In case we want to ship this optionally.

lacostej:
And I've added another full integration test that runs fastlane out of process.

lacostej:
Well this PR was long to land. I hit https://github.com/rubygems/rubygems/issues/5558 along the way.

It's ready for review from my hand. I'll rebase #20294 for which we can now add tests.

.circleci/config.yml
```diff
@@ -56,9 +56,11 @@ aliases:
     run:
       name: bundle install
       command: |
+        sudo gem update --system
         gem install bundler -v $(cat Gemfile.lock | tail -1 | tr -d " ")
         bundle config set --local path .bundle
         bundle check || bundle install --jobs=4 --retry=3
+        bundle env
```
lacostej, line 63: for troubleshooting and reporting potential issues. Can make a different PR.

.circleci/config.yml
```diff
@@ -56,9 +56,11 @@ aliases:
     run:
       name: bundle install
       command: |
+        sudo gem update --system
```
lacostej, line 59: I used this to troubleshoot an issue and to ensure we had up to date environment. Slows down the build and could be risky. Could make part of a different PR.

appveyor.yml
```diff
@@ -24,6 +24,8 @@ init:
 
 install:
   - set PATH=C:\Ruby26-x64\bin;%PATH%
+  - gem update --system
+  - FOR /F "usebackq delims==" %%A IN (`ruby appveyor/bundler_version.rb`) DO gem install bundler -v %%A
```
lacostej, line 28: this to ensure we use the version we bundled in the Gemfile.lock. Matches what we do on other CI environments

spec_helper.rb
```diff
@@ -14,9 +14,20 @@ module SpecHelper
 unless ENV["DEBUG"]
   fastlane_tests_tmpdir = "#{Dir.tmpdir}/fastlane_tests"
   $stdout.puts("Changing stdout to #{fastlane_tests_tmpdir}, set `DEBUG` environment variable to print to stdout (e.g. when using `pry`)")
+  $orig_stdout = $stdout
```
lacostej, line 17: helper code. Could be moved in a different PR, or discarded,

fastlane/lib/fastlane/plugins/plugin_manager.rb
joshdholtz: There _shouldn't_ be a case where we run into an infinite loop here, right? 🤔 

Could we add an if statement here to not run `collect_bundler_runtime_dependencies` again if `d` is already in the collection?

spec_helper.rb
joshdholtz: I can't figure out what this is doing 😅  Does this do something different than what setting `DEBUG` would do? 🤔

fastlane/spec/plugins_specs/plugin_manager_spec.rb
```diff
@@ -38,13 +39,32 @@
...
+        expect(plugin_manager.available_plugins).to eq(["fastlane-plugin-appcenter"])
+      end
+
+      # full integration test
+      it "returns all fastlane plugins found in transitive dependencies" do
+        Bundler.with_unbundled_env do
+          Dir.chdir("fastlane/spec/fixtures/plugins/GemfileWithDeps") do
+            `bundle install`
+            output = `bundle exec fastlane lanes`
+            expect(output.index("fastlane-plugin-appcenter")).not_to eq(nil)
```
joshdholtz, line 58: I think this might also work 👇 

```suggestion
            expect(output.include?("fastlane-plugin-appcenter")).to be(true)
```

.circleci/config.yml
joshdholtz: I like this here! I think this makes sense 😊

.circleci/config.yml
joshdholtz: Doesn't seem like it slows things down _too_ much so probably fine to keep in 😇

joshdholtz (dismissed):
This looks great! This changes makes a lot of sense 💪 Just left a few comments and thoughts!

spec_helper.rb
lacostej: @joshdholtz this was meant to be used together with the helper method below. It allows me to add code to re-activate the output for just one or few tests.

fastlane/lib/fastlane/plugins/plugin_manager.rb
lacostej: I made a quick test, with 2 projects dependending on each other. Code and execution below:

```
$ find .. -type f -exec sh -c 'for i; do echo "\n=== $i\n"; cat $i; done' sh {} +

=== ../1/1.gemspec

# frozen_string_literal: true
Gem::Specification.new do |spec|
  spec.name        = 'a1'
  spec.version     = '1.0.0'
  spec.authors     = ["x"]
  spec.email       = 'x@example.com'

  spec.required_ruby_version = '>= 2.5.0'

  spec.summary     = "1"
  #spec.description = U3d::DESCRIPTION

  spec.add_dependency 'a2'
end

=== ../1/test.rb

require 'bundler'

def gemfile_path
  # This is pretty important, since we don't know what kind of
  # Gemfile the user has (e.g. Gemfile, gems.rb, or custom env variable)
  Bundler::SharedHelpers.default_gemfile.to_s
rescue Bundler::GemfileNotFound
  nil
end

def lockfile_path
  Bundler::SharedHelpers.default_lockfile.to_s
rescue Bundler::GemfileNotFound
  nil
end

def collect_bundler_runtime_dependencies(collection, source)
  runtime_deps = source.dependencies.select { |d| d.type == :runtime }
  collection.concat(runtime_deps)
  runtime_deps.each do |d|
    collect_bundler_runtime_dependencies(collection, d.to_spec)
  rescue Gem::MissingSpecError
  # ignoring unresolvable dependencies
  end
  collection
end

puts gemfile_path
puts lockfile_path
dsl = Bundler::Dsl.evaluate(gemfile_path, lockfile_path, true)
dependencies = collect_bundler_runtime_dependencies([], dsl).uniq
puts dependencies.map(&:name).join(",")

=== ../1/Gemfile

source "https://rubygems.org"


gemspec
gem "a2", path: "../2"

=== ../2/2.gemspec

# frozen_string_literal: true
Gem::Specification.new do |spec|
  spec.name        = 'a2'
  spec.version     = '1.0.0'
  spec.authors     = ["x"]
  spec.email       = 'x@example.com'

  spec.required_ruby_version = '>= 2.5.0'

  spec.summary     = "2"
  #spec.description = U3d::DESCRIPTION

  spec.add_dependency 'a1'
end

=== ../2/Gemfile

source "https://rubygems.org"

gemspec
gem "a1", path: "../1"
```


execution

```
lacostej@MacBook-Pro-10 ~/Code/test/1 (dev2021) bundle install
Resolving dependencies...
Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem 'a1' or gem 'a2' and try again.
lacostej@MacBook-Pro-10 ~/Code/test/1 (dev2021) ruby test.rb 
/Users/lacostej/Code/test/1/Gemfile
/Users/lacostej/Code/test/1/Gemfile.lock
a1,a2

lacostej@MacBook-Pro-10 ~/Code//test/1 (dev2021) cd ../2/
lacostej@MacBook-Pro-10 ~/Code/test/2 (dev2021) ruby ../1/test.rb 
/Users/lacostej/Code/test/2/Gemfile
/Users/lacostej/Code/test/2/Gemfile.lock
a2,a1
lacostej@MacBook-Pro-10 ~/Code/test/2 (dev2021) bundle install
Resolving dependencies...
Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem 'a1' or gem 'a2' and try again.
```

so bundle install doesn't work, no Gemfile.lock gets generated. And the code doesn't loop.

lacostej:
I made some changes. Note: given that I've split this into several independent commits, can I merge without squashing this time? It would ease reverting the independent parts if we change our minds. Also makes code history clearer.

fastlane/lib/fastlane/plugins/plugin_manager.rb
lacostej: I added the code to prevent the potential recursive call, in case.

lacostej: For information the code was 
```
      # prevent potential infinite recursion. Should never happen. Better safe than sorry.
      runtime_deps -= collection
```

spec_helper.rb
lacostej: I'll resolve this as well.

fastlane/spec/fixtures/plugins/GemfileWithDeps/dependency_with_plugins/dependency_with_plugins.gemspec
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+lib = File.expand_path('lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+
+Gem::Specification.new do |spec|
+  spec.name        = 'dependency_with_plugins'
+  spec.version     = '1.0.0'
+  spec.authors     = ["Fastlane team"]
```
rogerluan, line 9:
```suggestion
  spec.authors     = ["fastlane team"]
```
_But Roger, even in fixtures??_  😝

spec_helper.rb
```diff
@@ -14,9 +14,20 @@ module SpecHelper
...
   $stdout = File.open(fastlane_tests_tmpdir, "w")
 end
 
+# for troubleshooting CI issues on specific tests
+def with_orig_stdout(&block)
```
rogerluan, line 22: Maybe add docs on how to use this? It's not clear to me when or how I'd use this

rogerluan (approved):
This is a great DX improvement 🙏 thanks @lacostej !

spec_helper.rb
lacostej: documented. Makes sense?

rogerluan (approved):

rogerluan:
AppVeyor CI is failing @lacostej 🤔  how come the other one passed for all envs? 

image

rogerluan:
Nvm, just read your comment in the other PR lol

vovkasm:
Sorry for noise, but can I help with anything to move this forward?

PR #20294, opened 2022-05-18
[Feature] import from gem action

lacostej:
### Checklist

### Motivation and Context
This builds on top of #20293

We wish to replace `import_from_git` with this new action to use gem files to package the shared Fastfile and its dependencies. Together with the transitive dependencies it simplifies the use of a shared Fastfile.

We modify our repository containing our shared Gemfile by
* replacing our shared repo's Gemfile by a gemspec
* moving all Plugins into the gemspec file
* replacing the `fastlane_version "x.y.z"` in the shared Fastfile with a proper version constraint in the gemspec. For consumer projects, bundler will resolve the right dependency when needed

For our consumer projects this gives the following results:
* the project's gemfile contains a single dependency to the shared Gem (instead of one
```
gem 'name_of_gem', git: 'URL_GIT_REPO'
```
* we remove from the Pluginfile all dependencies now managed by bundler
* we replace our Fastfile import from

```
import_from_git(url: "URL_GIT_REPO", dependencies: [...])
```
to
```
import_from_gem(gem_name: "name_of_gem")
```

Extra benefits: builds are now reproducible. Updating the shared fastfile requires doing an update in the corresponding user projects.
 
I am opening this for discussion and can clean the code, add tests as needed.

Who can provide some feedback on this? @joshdholtz maybe?

### DIFF
--- a/fastlane/lib/fastlane/actions/import_from_gem.rb
+++ b/fastlane/lib/fastlane/actions/import_from_gem.rb
@@ -1,0 +1,54 @@
+module Fastlane
+  module Actions
+    class ImportFromGemAction < Action
+      def self.run(params)
+        # in fast_file.rb
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Import fastfile/actions from a gem, alternative to import_from_git"
+      end
+
+      # def self.details
+      # end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :gem_name,
+                                       optional: false,
+                                       type: String,
+                                       default_value: "",
+                                       description: "The gem to import Fastfile and actions from"),
+          FastlaneCore::ConfigItem.new(key: :paths,
+                                       description: "The path(s) of the Fastfile in the repository",
+                                       default_value: ['fastlane/Fastfile*'],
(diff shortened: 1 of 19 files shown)


vovkasm:
For someone with same requirements, there is a workaround. At work we need to import some common lanes from our internal plugin, currently we do this:
1. Place shared Fastfile into `fastlane/Fastfile` folder of gem sources.
2. At plugin entry point `lib/fastlane/plugin/plugin_name.rb` add this snippet
```ruby
# Hack to provide "import_common_fastfile" keyword to Fastfile
module Fastlane
  class FastFile
    def import_common_fastfile
      action_launched('import_common_fastfile')
      path = File.expand_path('../../../fastlane/Fastfile', __dir__)
      UI.user_error!("Could not find common Fastfile at path '#{path}'") unless File.exist?(path)

      return_value = import(path)
      action_completed('import_common_fastfile', status: FastlaneCore::ActionCompletionStatus::SUCCESS)
      return_value
    end
  end
end
```
3. Use this in project Fastfile
```ruby
# This lines will add common shared lanes to your project
fastlane_require "fastlane-plugin-plugin_name"
import_common_fastfile
```

It's not cleanest solution of course, becuase of this I hope this PR will be merged ;-)

PR #20297, opened 2022-05-18
Fix/import from git spec fail with non master default branch

lacostej:
On my computer, tests are failing as my default git branch is `main`.

### DIFF
--- a/fastlane/spec/actions_specs/import_from_git_spec.rb
+++ b/fastlane/spec/actions_specs/import_from_git_spec.rb
@@ -9,6 +9,16 @@ describe Fastlane do
         end.to raise_error("Please pass a path to the `import_from_git` action")
       end
 
+      # (2.32.0)
+      # (2.32.0).win.1
+      # (3)
+      def git_version
+        output = `git --version 2> /dev/null`
+        m = output.match(/(?<version>\d+(?:\.\d+)*)/)
+        raise "Couldn't parse git version in '#{output}'" unless m
+        m['version']
+      end
+
       describe "with caching" do
         let(:caching_message) { "Eligible for caching" }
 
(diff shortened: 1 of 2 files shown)


lacostej:
Looks like not everything is stable on the other platforms. I'll keep the master `branch` name for now.

fastlane/spec/actions_specs/import_from_git_spec.rb
```diff
@@ -24,7 +24,11 @@
           source_directory_path = Dir.mktmpdir("fl_spec_import_from_git_source")
 
           Dir.chdir(source_directory_path) do
-            `git init -b master`
+            if Gem::Version.new(`git --version`.strip.split(' ').last) >= Gem::Version.new("2.28.0")
```
revolter, line 27: `git --version` outputs `git version 2.32.0 (Apple Git-132)` on my machine, so I think that this check would be incorrect.

Isn't there any other way to retrieve only the version, without needing to parse it?

revolter (changes requested):

fastlane/spec/actions_specs/import_from_git_spec.rb
revolter: [Apparently not](https://git-scm.com/docs/git-version) :(

revolter: Maybe using https://regex101.com/r/fImCOu/1 would be a better idea?

lacostej: @revolter yes, better. I even made most groups optional to parse a potential "3" version.

revolter: I didn't do that so as not to match the `132` in my example, but I guess that `.match()` only returns the first match, so it's all good 👍🏻

revolter (approved):
Thanks for the fix 💪🏻

PR #20311, opened 2022-05-22
[match] add profile_type filtering when checking if profile exists

PaulTaykalo:
### Checklist

### Motivation and Context

When `match` checking if profile exists on AppStore Connect, the check is relatively slow, if performed without filtering.
App Store Connect API doesn't support filtering by the profile `uuid`, but it supports filtering by `profileType`.
Adding filtering by `profileType` significantly speeds up requests, that`match` is performing

https://developer.apple.com/documentation/appstoreconnectapi/list_and_download_profiles#query-parameters

It seems that response time  highly depends on the amount of returned items.
It was around ~20s for 85 profiles, and ~4s for 9 profiles (with filtering).

### Description

Additional filtering is added by `profileType`


sigh/lib/sigh/download_all.rb
```diff
@@ -55,8 +55,6 @@ def download_all(download_xcode_profiles: false)
         ]
       end
 
-      # Filtering on 'profileType' seems to be undocumented as of 2020-07-30
-      # but works on both web session and official API
```
PaulTaykalo, line 59: It is actually mentioned in the documentation at this moment (2022-05-22)
https://developer.apple.com/documentation/appstoreconnectapi/list_and_download_profiles#query-parameters

match/lib/match/spaceship_ensure.rb
```diff
@@ -97,5 +99,43 @@ def profile_exists(username: nil, uuid: nil, platform: nil)
         return nil
       end
     end
+
+    # @return the name of the provisioning profile type
+    # The kind of provisioning profile we're interested in
+    def profile_types(prov_type)
```
lucgrabowski, line 105: Could we extract this method to `match/lib/match/module.rb`? It looks like it has the same implementation as in `match/lib/match/nuke.rb`

lucgrabowski (changes requested):
Thanks for optimizing fetching profiles :) This is great.

match/lib/match/spaceship_ensure.rb
```diff
@@ -74,10 +74,12 @@ def certificates_exists(username: nil, certificate_ids: [])
...
     end
 
-    def profile_exists(username: nil, uuid: nil, platform: nil)
+    def profile_exists(type: nil, username: nil, uuid: nil, platform: nil)
       # App Store Connect API does not allow filter of profile by platform or uuid (as of 2020-07-30)
       # Need to fetch all profiles and search for uuid on client side
-      found = Spaceship::ConnectAPI::Profile.all.find do |profile|
+      # But we can filter provisioning profiles based on their type (this, in general way faster than getting all profiles)
+      filter = { profileType: profile_types(type).join(",") } if type
+      found = Spaceship::ConnectAPI::Profile.all(filter: filter).find do |profile|
```
joshdholtz, line 82: This is so smart! Slightly embarrassed I didn't do this right away 🙈

match/lib/match/spaceship_ensure.rb
joshdholtz: Yeah, this would be great if we could abstract this out!

joshdholtz (changes requested):

match/lib/match/nuke.rb
```diff
@@ -148,9 +148,7 @@ def prepare_list
       # Get all iOS and macOS profile
       self.profiles = []
       prov_types.each do |prov_type|
-        types = profile_types(prov_type)
-        # Filtering on 'profileType' seems to be undocumented as of 2020-07-30
-        # but works on both web session and official API
+        types = Match.profile_types(prov_type)
```
PaulTaykalo, line 151: Same here. this information is actually mentioned in documentation at this moment (2022-05-22)
https://developer.apple.com/documentation/appstoreconnectapi/list_and_download_profiles#query-parameters

lucgrabowski (approved):
LGTM 🔥 Thanks for improving performance :)

PaulTaykalo:
@joshdholtz Is there anything needed to be added?

PaulTaykalo:
bump

joshdholtz (approved):
Thanks for this PR! This is so so so much better 😊

PR #20324, opened 2022-05-26
[spaceship] Allow for removing app from sale by updating territories

lucgrabowski:
### Checklist

### Motivation and Context
There are 3 discussions with questions about removing apps from sale: https://github.com/fastlane/fastlane/discussions/19040, https://github.com/fastlane/fastlane/discussions/17032, https://github.com/fastlane/fastlane/discussions/19375
This change allows for removing an app from sale by updating a list of territories.
```ruby
app = ... # fetch app
# 
app.update(attributes:{ "availableInNewTerritories": false }, territory_ids:[], allow_removing_from_sale: true) 
```
### Description
It was not possible to remove app from all territories as app update method expected either `nil` or an array of `territory_ids` with at least one territory id.

Just for explain use of `availableInNewTerritories`: 
`availableInNewTerritories` attribute is responsible for `New Countries or Regions` checkbox
available in new territories

### Testing Steps

Tested by removing an app from sale using code above in `irb`

### DIFF
--- a/deliver/lib/deliver/upload_price_tier.rb
+++ b/deliver/lib/deliver/upload_price_tier.rb
@@ -12,7 +12,9 @@ module Deliver
       app = Deliver.cache[:app]
 
       attributes = {}
-      territory_ids = []
+
+      # Check App update method to understand how to use territory_ids.
+      territory_ids = nil # nil won't update app's territory_ids, empty array would remove app from sale.
 
       # As of 2020-09-14:
       # Official App Store Connect does not have an endpoint to get app prices for an app
--- a/spaceship/lib/spaceship/connect_api/models/app.rb
+++ b/spaceship/lib/spaceship/connect_api/models/app.rb
@@ -101,6 +101,8 @@ module Spaceship
         return client.get_app(app_id: app_id, includes: includes).first
       end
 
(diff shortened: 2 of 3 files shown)


lacostej:
Code looks good. Maybe add documentation to the method stating that empty array removes the app from sale.

Is the `Tunes.update` that calls `patch_app` used anywhere (even in popular external actions)? Is there a risk of callers accidentally passing empty arrays earlier that would trigger accidental removing from sale?

lucgrabowski:
Thanks for review @lacostej . I'll add documentation. 

> Is the `Tunes.update` that calls `patch_app` used anywhere (even in popular external actions)? Is there a risk of callers accidentally passing empty arrays earlier that would trigger accidental removing from sale?

You're right. `patch_app` isn't called in any other place, but `update` is called in several places and in one of them in `UploadPriceTier` it's called with `territory_ids` and it has to be fixed.

lacostej:
@lucgrabowski that makes me worried that someone has code somewhere that would cause this unintended behavior. Do we have a better way of rolling this without risking this?  Maybe it's acceptable as we can always roll it out... WDYT?

getaaron (approved):

PR #20356, opened 2022-06-05
[action][update_info_plist] improve the plist file-path option validation check

crazymanish:
### Checklist

### Motivation and Context
- Plist file-path Validation block was using range `[-6..-1]` which is a kind of magical number range for non-ruby developer

### Description
- Improve the plist file-path option validation check using `value.end_with?(".plist")` which is more readable, i hope 😍

### Testing Steps
- N/A

### DIFF
--- a/fastlane/lib/fastlane/actions/update_info_plist.rb
+++ b/fastlane/lib/fastlane/actions/update_info_plist.rb
@@ -80,7 +80,7 @@ module Fastlane
                                        description: "Path to info plist",
                                        optional: true,
                                        verify_block: proc do |value|
-                                         UI.user_error!("Invalid plist file") unless value.end_with?(".plist")
+                                         UI.user_error!("Invalid plist file") unless value.downcase.end_with?(".plist")
                                        end),
           FastlaneCore::ConfigItem.new(key: :scheme,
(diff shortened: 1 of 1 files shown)


fastlane/lib/fastlane/actions/update_info_plist.rb
```diff
@@ -80,7 +80,7 @@ def self.available_options
                                        description: "Path to info plist",
                                        optional: true,
                                        verify_block: proc do |value|
-                                         UI.user_error!("Invalid plist file") unless value[-6..-1].casecmp(".plist").zero?
+                                         UI.user_error!("Invalid plist file") unless value.end_with?(".plist")
```
lacostej, line 83: don't you want something like `value.downcase.end_with?(".plist")` to keep the `casecmp`? or it this not necessary anymore?

Alternatively `value =~ /\.plist^/i` is an option I guess

crazymanish: Thank you for your review. applied suggestion. 🙇

joshdholtz (approved):
Looks great to me! This feels much better 😊

PR #20386, opened 2022-06-15
Adding support for GitLab Secure Files as a Match storage mode

darbyfrey:
### Checklist

### Motivation and Context

GitLab recently released a feature called [Project-level Secure Files](https://docs.gitlab.com/ee/ci/secure_files/) which is a way to securely store configuration type files to be used in CI/CD pipelines and would work well for provisioning profiles and signing certificates.

### Description

This change introduces support for GitLab [Project-level Secure Files](https://docs.gitlab.com/ee/ci/secure_files/) as a storage backend for fastlane match. This is enabled by adding the following configuration to the Matchfile:

```
storage_mode("gitlab_secure_files")
gitlab_project('gitlab/project/path')
```

I realize this is a fairly large PR, so please let me know what additional details I can provide to make this easier to review, or of course any questions that come up along the way. Thanks!

### Testing Steps

Testing this feature can be done by setting up a project in GitLab and adding files as described in the instructions for  [Project-level Secure Files](https://docs.gitlab.com/ee/ci/secure_files/). Once the files are uploaded you can run `fastlane match` in a CI pipeline.

`fastlane match` can also be run locally by creating a personal access token in GitLab and setting the `PRIVATE_TOKEN` environment variable locally. 

I've also have a test project using this branch which you can see here https://gitlab.com/gitlab-org/incubation-engineering/devops-for-mobile-apps/gitlab_unfiltered, and the matchfile is here https://gitlab.com/gitlab-org/incubation-engineering/devops-for-mobile-apps/gitlab_unfiltered/-/blob/master/ios/fastlane/Matchfile


getaaron:
@darbyfrey thanks so much for adding this! Are you able to sign the CLA and rebase?

match/lib/match/encryption.rb
```diff
@@ -18,6 +18,9 @@ def backends
           "s3" => lambda { |params|
             params[:keychain_name] = params[:s3_bucket]
             return Encryption::OpenSSL.configure(params)
+          },
+          "gitlab_secure_files" => lambda { |params| 
+            return nil
```
getaaron, line 23: why don’t we want to encrypt with OpenSSL?

match/lib/match/storage/gitlab/client.rb
```diff
@@ -0,0 +1,79 @@
+
+module Match
+  module Storage
+    class GitLab
+      class Client
+        def initialize(api_v4_url:, project_id:, job_token: nil, private_token: nil)
```
getaaron, line 6: Should we raise an error if they try to set both job and private token?

match/spec/storage/gitlab_secure_files_storage_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'pry'
```
getaaron, line 1: We don’t need pry right?

getaaron (commented):
Added a few small Qs, otherwise looks awesome!

match/spec/storage/gitlab_secure_files_storage_spec.rb
darbyfrey: Good call, remove ✅

match/lib/match/storage/gitlab/client.rb
darbyfrey: What do you think about logging a warning instead? I'd prefer not to fail the CI job in this case, but logging a warning to let the user know what is happening could be a good option. In this case we would default to the `JOB_TOKEN`.

match/lib/match/encryption.rb
darbyfrey: Since Secure Files are encrypted within GitLab, I thought it wasn't necessary to encrypt again. This would also simplify the configuration since `MATCH_PASSWORD` wouldn't need to be set in this case.

darbyfrey:
@getaaron, thanks so much for the review! I've made some changes and followed up on your comments, so this change is ready for another review when you have a moment.

getaaron (approved):

getaaron:
Looks great!

PR #20399, opened 2022-06-21
[fastlane-core][scan] added destination param support inside the `xcodebuild` command

### LINKED ISSUE #19579: Scan + Xcode 13.1: WARNING: Using the first of multiple matching destinations
When using scan with Xcode 13.1

```
scan({
    devices: ["iPhone 13"],
    ....
  })
```

there is a warning:
WARNING: Using the first of multiple matching destinations

This makes it harder to have CI logs cleaner to easier find some errors.

image

crazymanish:
### Checklist

### Motivation and Context
- Resolves #19579
- Xcode13+ **xcodebuild command** `'without destination param'` generates annoying warnings

### Description
- Start supporting `'destination param'` for Xcode13+ **xcodebuild command**

### Testing Steps

- Update `Gemfile` to 👇 and run `bundle install`, `bundle update fastlane`, or `bundle update`

```rb
gem "fastlane", :git => "https://github.com/fastlane/fastlane.git", :branch => "crazymanish-xcodebuild-destination-param-fix"
```
- Try to run unit tests using `scan`
```ruby
scan(destination: "name=iPhone 13 Pro Max,OS=15.4") # adjust it based on your Xcode version.
```

## DemoApp for testing
- Open/UnZip [DemoApp.zip](https://github.com/fastlane/fastlane/files/8952930/DemoApp.zip) 
- Goto terminal for DemoApp folder location
- `bundle exec fastlane run_unit_test` # this will show warning similar to original issue
- later
- To test fix: `bundle exec fastlane run_unit_test_with_destination` # should not show warnings

### Demo screenshot
Schermafbeelding 2022-06-21 om 10 39 24 PM

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -356,7 +356,7 @@ module FastlaneCore
       # This xcodebuild bug is fixed in Xcode 8.3 so 'clean' it's not necessary anymore
       # See: https://github.com/fastlane/fastlane/pull/5626
       if FastlaneCore::Helper.xcode_at_least?('8.3')
-        command = "xcodebuild -showBuildSettings #{xcodebuild_parameters.join(' ')}#{build_xcodebuild_destination}"
+        command = "xcodebuild -showBuildSettings #{xcodebuild_parameters.join(' ')}#{xcodebuild_destination_parameter}"
       else
         command = "xcodebuild clean -showBuildSettings #{xcodebuild_parameters.join(' ')}"
       end
@@ -365,24 +365,25 @@ module FastlaneCore
 
     def build_xcodebuild_resolvepackagedependencies_command
       return nil if options[:skip_package_dependencies_resolution]
-      command = "xcodebuild -resolvePackageDependencies #{xcodebuild_parameters.join(' ')}#{build_xcodebuild_destination}"
+      command = "xcodebuild -resolvePackageDependencies #{xcodebuild_parameters.join(' ')}#{xcodebuild_destination_parameter}"
       command
     end
 
-    def build_xcodebuild_destination
-      # Xcode13+ xcodebuild command 'without destination param' generates annoying warnings
+    def xcodebuild_destination_parameter
+      # Xcode13+ xcodebuild command 'without destination parameter' generates annoying warnings
       # See: https://github.com/fastlane/fastlane/issues/19579
-      destination = ""
+      destination_parameter = ""
       xcode_at_least_13 = FastlaneCore::Helper.xcode_at_least?("13")
       if xcode_at_least_13 && options[:destination]
         begin
-          destination = " " + "-destination #{options[:destination].shellescape}"
+          destination_parameter = " " + "-destination #{options[:destination].shellescape}"
         rescue => ex
-          # we really don't care about destination warnings if something goes wrong with shellescape
-          UI.important("Failed to set destination for xcodebuild command: #{ex}")
+          # xcodebuild command can continue without destination parameter, so
+          # we really don't care about this exception if something goes wrong with shellescape
+          UI.important("Failed to set destination parameter for xcodebuild command: #{ex}")
         end
       end
-      destination
+      destination_parameter
     end
 
     # Get the build settings for our project
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -342,13 +342,6 @@ module FastlaneCore
         proj << "-disableAutomaticPackageResolution"
       end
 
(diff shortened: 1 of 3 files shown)


crazymanish:
Thank you very much @AliSoftware for testing. I have improved the logic a little bit to avoid the shell-escape exception.

Also, i added the **DemoApp** project to test this PR, and updated the PR description.

## DemoApp for testing
- Open/UnZip [DemoApp.zip](https://github.com/fastlane/fastlane/files/8952930/DemoApp.zip) 
- Goto terminal for DemoApp folder location
- `bundle exec fastlane run_unit_test` # this will show warning similar to original issue
- later
- To test fix: `bundle exec fastlane run_unit_test_with_destination` # should not show warnings

AliSoftware:
Updating my `Gemfile.lock` (`bundle update fastlane`) to your latest commit and tested without a `destination` for the `scan` call and it now seems to pass (as opposed to the crash I reported above yesterday).

After testing _with_ a `destination` parameter for the `scan` call, and I can confirm that the `xcodebuild: WARNING: Using the first of multiple matching destinations` warnings from the `scan` command are now gone… BUT:

1. It's worth noting that I then get this warning because of passing that value now, which doesn't bring confidence to the user 😅 
```
  | [14:25:24]: It's not recommended to set the `destination` value directly
  | [14:25:24]: Instead use the other options available in `fastlane scan --help`
  | [14:25:24]: Using your value 'name=iPhone Xs Max' for now
  | [14:25:24]: because I trust you know what you're doing...
```

2. I still have the `xcodebuild: WARNING: Using the first of multiple matching destinations:` warnings from my invocations of `capture_ios_screenshots` action (aka `snapshot`) that I run after my invocation of `scan` 😢

crazymanish:
Thank you for the testing and feedback!

> I can confirm that the `xcodebuild: WARNING: Using the first of multiple matching destinations` warnings from the `scan` command are now gone…

Yay! 😍

> It's worth noting that I then get this warning because of passing that value now, which doesn't bring confidence to the user 😅

Indeed, Fastlane does not recommend to use `destination` directly, because CI can fail if failed to find that specific/destination simulator. (but there is no other way to avoid this warning without destination param so 🤷‍♂️...)
https://github.com/fastlane/fastlane/blob/master/scan/lib/scan/detect_values.rb#L212

> I still have the xcodebuild: WARNING: Using the first of multiple matching destinations: warnings from my invocations of capture_ios_screenshots action (aka snapshot) 

I just fixed this warning for `scan` i will debug and will fix it for `snapshot` too...will update this PR soon.

AliSoftware:
> Indeed, Fastlane does not recommend to use destination directly, because CI can fail if failed to find that specific/destination simulator.

Maybe a more intuitive solution for the user would be to have `scan` do the inference of the right `destination` to pass to `xcodebuild` (eg calling `xcodebuild -showdestinations -workspace … -scheme …` and picking the first one from the returned list)?

That way not only would it mean that it's not the responsibility of the caller of scan to know what destination to pass, making it easier to use at call site, it that will avoid 
the risk of using an unsupported destination for the project, and avoid xcodebuild complain with the annoying warning that this PR is trying to suppress…?

---

In other words, if Xcode >13:
 - If a specific destination is passed, use it (and maybe keep the warning about "we trust you know what you're doing" for that case then 🤷‍♂️
 - If no options[:destination] is provided, use `xcodebuild -showdestinations` to pick the first one matching, then use that for the ultimate call to `xcodebuild` made by `scan`. And in that case, no warning (neither from fastlane about not recommending we pass an explicit destination, nor from xcodebuild warning it will infer it)

Wdyt?

getaaron:
@crazymanish I think this is a good initial approach, we can always improve it more later

2 suggested improvements before merging, both should be pretty small

1. can you please add tests to cover the expected behavior? so far the only changes to the test file are updates to ensure the old test cases still function. we should add at least 2 tests:
    - xcode is >13 and destination is provided
    - xcode is >13 and destination is not provided 
1. i think we should suppress the warning when destination is provided for xcode >13, we can just update the conditional here to hide it https://github.com/fastlane/fastlane/blob/f0b9bbe0c391f034d997ae2ce48bac567c29fa7e/scan/lib/scan/detect_values.rb#L211-L218

crazymanish:
Thank you Aaron for your review and suggestions. i will work on your suggestions after tomorrow...and inform you once PR is ready for review again... 🙇 

Tomorrow i m attending the [AppDevcon conference](https://appdevcon.nl/schedule-2022/) 😇

getaaron:
> Tomorrow i m attending the [AppDevcon conference](https://appdevcon.nl/schedule-2022/) 😇

Check out @designatednerd ’s WWDC speedrun talk and tell her I said hi 😅

crazymanish:
Applied review suggestions, PR is ready for review! 😇

getaaron (approved):
Looks awesome 🤩

PR #20413, opened 2022-06-25
[Ruby] Ruby 2.6 is now the minimum (dropping Ruby 2.5)

joshdholtz:
### Motivation and Context

Ruby 2.5 has been end-of-lifed 1 year and 2 months ago on March 31st, 2021.

Some of our dependencies are bumping to a minimum of Ruby 2.6 which causes some rough installations. Specifically, `signet`'s latest release of `0.17` no longer supports Ruby 2.5 and `signet` is used by a bunch of Google gems.

### Description

- Bumped `fastlane.gemspec` to minimum Ruby 2.6
- Bumped `.rubocop.yml` to target of Ruby 2.6
- Updated CircleCI to no longer have a Ruby 2.5
- Updated plugin template to use Ruby 2.6

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ require:
   - ./rubocop/is_string_usage.rb
 
 AllCops:
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.6
   NewCops: enable
   Include:
     - '**/*.rb'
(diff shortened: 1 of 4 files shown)


.rubocop.yml
```diff
@@ -24,6 +24,12 @@ AllCops:
     - '**/spec/fixtures/broken_files/broken_file.rb'
     - '**/*.provisionprofile'
 
+Lint/ErbNewArguments:
+  Enabled: false
+
+Style/SlicingWithRange:
+  Enabled: false
+
```
joshdholtz, line 32: Will disable these later but wanted to make this change PR minimal 🤷‍♂️

FilipMaslovaric:
Shouldn't this kinda change warrant a major version bump, rather than a minor bump?

PR #20429, opened 2022-07-03
[snapshot] use empty string to blank out operator name by default

ZevEisenberg:
### Checklist
(made edits in the GitHub web UI; I'll check out the code and run these checks and update this)

### Motivation and Context

If you don't pass `--operatorName` to `xcrun simctl status_bar`, it defaults to "Carrier". This change makes it blank instead, which matches Apple's marketing screenshots.

Note: on some older simulators, it may be necessary to pass a zero-width space `\u200D` instead of an empty string, but iOS 15.5 (and possibly earlier) seem to support an empty string to blank out the carrier.

### Description

Use an empty string explicitly instead of using the default. Tested locally with `xcrun simctl` to verify the correct parameters.

### Testing Steps

Probably easier to just test with simctl. This will produce a good status bar (just the cellular section; leaving out the other parameters for brevity):

```sh
xcrun simctl status_bar booted override --operatorName "" --cellularMode active
```

Simulator Screen Shot - iPhone 8 - 2022-07-03 at 16 20 55

Whereas this produces "Carrier" in the screenshot:

```sh
xcrun simctl status_bar booted override --cellularMode active
```

Simulator Screen Shot - iPhone 8 - 2022-07-03 at 16 20 51

### DIFF
--- a/snapshot/lib/snapshot/options.rb
+++ b/snapshot/lib/snapshot/options.rb
@@ -133,7 +133,7 @@ module Snapshot
                                      is_string: false),
         FastlaneCore::ConfigItem.new(key: :override_status_bar_arguments,
                                      env_name: 'SNAPSHOT_OVERRIDE_STATUS_BAR_ARGUMENTS',
-                                     description: "Fully customize the status bar by setting each option here. See `xcrun simctl status_bar --help`",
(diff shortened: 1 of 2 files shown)


ZevEisenberg:
Had some build flakiness, but I re-ran them a couple times and everything is passing now.

getaaron (approved):

PR #20430, opened 2022-07-04
[action][changelog_from_git_commits] Fix description for documentation

mbernson:
### Checklist

### Motivation and Context

I was using the changelog_from_git_commits Fastlane action. The documentation told me to use a Ruby symbol for the `merge_commit_filtering` parameter, but turned out to actually expect a string. So I decided to update the doc to make this match.

### Description

The documentation says that valid values for the parameter `merge_commit_filtering` in this action are Ruby symbols. But it actually expects the value to be a string. This change updates the action's description to make this match.


getaaron (approved):
Thanks, looks good! 😍

PR #20436, opened 2022-07-05
Ignoring case sensitive when requesting SMS

harrimaatta:
### Checklist

### Motivation and Context

Even that documentation guides users to use 'sms', I have noticed that some people would like to use 'SMS' for acronyms like this one. Removing case sensitivity doesn't allow other strings, but helps for non-precise people.

### Description

In /fastlane/spaceship/lib/spaceship/two_step_or_factor_client.rb line 167 
From
`if code == 'sms'`
To
`if code.casecmp?("sms")`

Before any changes, 
`bundle exec fastlane spaceauth -u <your_email>`
And check that you are able to order SMS 2FA code.

### Testing Steps

`bundle exec fastlane spaceauth -u <your_email>`
And when reaching
`Please enter the 6 digit code:`
You can type any variant of 'sms', like 'sms',  'smS',  'sMs',  'sMS', 'Sms',  'SmS',  'SMs',  'SMS'
When testing variants, running 
`rm -rf ~/.fastlane/spaceship/<your_email>/cookie`
deletes a cookie and on next run you are triggered again with 2FA query. 

Also, non-valid value, like an 'asdf', still works giving expected error: 
`Exception type: Spaceship::UnauthorizedAccessError`

### DIFF
--- a/spaceship/lib/spaceship/two_step_or_factor_client.rb
+++ b/spaceship/lib/spaceship/two_step_or_factor_client.rb
@@ -164,7 +164,7 @@ module Spaceship
         body = { "securityCode" => { "code" => code.to_s } }.to_json
 
         # User exited by entering `sms` and wants to choose phone number for SMS
-        if code == 'sms'
+        if code.casecmp?("sms")
           code_type = 'phone'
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Looks great 👍

PR #20448, opened 2022-07-08
[match][sigh] prefer default keychain during wwdr cert installation

DuMaM:
### Checklist

### Motivation and Context
When there are multiple keychains WWDR cert is installed in random place.
During search it takes first output from `list-keychains` instead of default one.
https://developer.apple.com/support/certificates/
https://developer.apple.com/support/expiration/

### Description
* expired cert got removed to do not introduce conflicts
* latest cert was added 
* changed priority so default keychain is taken as first

### Testing Steps
1. Create new keychain without any cert
2. Run with `--verbose` lane where `match` install Provision Profile and it's certs. 
3. Take a look on debug log to see if it's using correct keychain for installation WWDR cert.

### DIFF
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -22,7 +22,7 @@ module FastlaneCore
     end
 
     def self.installed_identies(in_keychain: nil)
-      install_wwdr_certificates unless wwdr_certificates_installed?
+      install_wwdr_certificate unless wwdr_certificate_installed?
 
       available = list_available_identities(in_keychain: in_keychain)
       # Match for this text against word boundaries to avoid edge cases around multiples of 10 identities!
@@ -81,7 +81,7 @@ module FastlaneCore
       `#{commands.join(' ')}`
     end
 
-    def self.wwdr_certificates_installed?
+    def self.wwdr_certificate_installed?
       certificate_name = "Apple Worldwide Developer Relations Certification Authority"
       keychain = wwdr_keychain
       response = Helper.backticks("security find-certificate -a -c '#{certificate_name}' #{keychain.shellescape}", print: FastlaneCore::Globals.verbose?)
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -86,7 +86,7 @@ module FastlaneCore
       keychain = wwdr_keychain
       response = Helper.backticks("security find-certificate -a -c '#{certificate_name}' #{keychain.shellescape}", print: FastlaneCore::Globals.verbose?)
       certs = response.split("keychain: \"#{keychain}\"").drop(1)
-      certs.count == 2
+      certs.count == 1
     end
 
     def self.install_wwdr_certificate
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -124,8 +124,8 @@ module FastlaneCore
 
     def self.wwdr_keychain
       priority = [
-        "security list-keychains -d user",
-        "security default-keychain -d user"
+        "security default-keychain -d user",
+        "security list-keychains -d user"
       ]
       priority.each do |command|
         keychains = Helper.backticks(command, print: FastlaneCore::Globals.verbose?).split("\n")
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -86,7 +86,7 @@ module FastlaneCore
       keychain = wwdr_keychain
       response = Helper.backticks("security find-certificate -a -c '#{certificate_name}' #{keychain.shellescape}", print: FastlaneCore::Globals.verbose?)
       certs = response.split("keychain: \"#{keychain}\"").drop(1)
-      certs.count == 1
+      certs.count >= 1
     end
 
     def self.install_wwdr_certificate
--- a/fastlane_core/lib/fastlane_core/cert_checker.rb
+++ b/fastlane_core/lib/fastlane_core/cert_checker.rb
@@ -90,10 +90,7 @@ module FastlaneCore
     end
 
     def self.install_wwdr_certificate
-      install_wwdr_certificate('https://www.apple.com/certificateauthority/AppleWWDRCAG6.cer')
-    end
-
-    def self.install_wwdr_certificate(url)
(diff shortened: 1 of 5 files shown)


fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -90,8 +90,7 @@ def self.wwdr_certificates_installed?
     end
 
     def self.install_wwdr_certificates
-      install_wwdr_certificate('https://developer.apple.com/certificationauthority/AppleWWDRCA.cer')
```
getaaron, line 93: @joshdholtz I think removing this should be OK. According to [Apple Worldwide Developer Relations Intermediate Certificate Expiration](https://developer.apple.com/support/expiration/) nobody should be using this any more and it's set to expire on February 7, 2023 anyway, so not much point in installing it.

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -124,8 +123,8 @@ def self.install_wwdr_certificate(url)
 
     def self.wwdr_keychain
       priority = [
-        "security list-keychains -d user",
-        "security default-keychain -d user"
+        "security default-keychain -d user",
+        "security list-keychains -d user"
```
getaaron, line 127: I'm not sure about this change, according to https://github.com/fastlane-old/fastlane_core/pull/116 we should start with the login keychain - if we want to move default up, maybe we should insert `security login-keychain` at the very top?

fastlane_core/spec/cert_checker_spec.rb
```diff
@@ -20,8 +20,7 @@
 
     describe '#install_wwdr_certificates' do
       it 'should install all the official WWDR certificates' do
```
getaaron, line 22:
```suggestion
      it 'should install only the latest official WWDR certificate' do
```

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -90,8 +90,7 @@ def self.wwdr_certificates_installed?
     end
 
     def self.install_wwdr_certificates
```
getaaron, line 92: since this only installs one now, maybe we should rename the method
```suggestion
    def self.install_wwdr_certificate
```

fastlane_core/lib/fastlane_core/cert_checker.rb
DuMaM: Login keychain is default, most of the time. Also if you have keychain named like aaaa.keychain-db then login one will be not picked here at all. 
From me it has more sense to start default one, because in most cases its used by match and other tools.

fastlane_core/lib/fastlane_core/cert_checker.rb
DuMaM: Ok, agree

fastlane_core/lib/fastlane_core/cert_checker.rb
DuMaM: Yep. Thats the reason i removed it.
This old cert also was source of many resign problems.

fastlane_core/spec/cert_checker_spec.rb
```diff
@@ -2,27 +2,26 @@
...
-      it 'should install all the official WWDR certificates' do
-        expect(FastlaneCore::CertChecker).to receive(:install_wwdr_certificate).with(/AppleWWDRCA/)
-        expect(FastlaneCore::CertChecker).to receive(:install_wwdr_certificate).with(/AppleWWDRCAG3/)
-        FastlaneCore::CertChecker.install_wwdr_certificates
+    describe '#install_wwdr_certificate' do
+      it 'should install only the latest official WWDR certificate' do
+        expect(FastlaneCore::CertChecker).to receive(:install_wwdr_certificate)
+        FastlaneCore::CertChecker.install_wwdr_certificate
       end
     end
```
DuMaM, line 26: I'm not sure it this tests is needed now

DuMaM:
@getaaron could you please review this once more? :)

DuMaM:
It's ready to merge.

fastlane_core/lib/fastlane_core/cert_checker.rb
DuMaM: @getaaron 
This old setup is based on this comment
https://github.com/fastlane-old/fastlane_core/pull/116#issuecomment-190934610
Which is not true. Default-keychain is one set by user from gui or `security default-keychain -s <keychain_name>`.
There are chances that it's not set at all, but then we will pick any from list-keychains.

getaaron:
@DuMaM same as your other PR, can you please connect CircleCI to github and ensure all the builds run?

getaaron (approved):
this looks good to me, approved to merge pending CI, and I'll try to get a second review just in case

DuMaM:
@getaaron CI is fixed here also :)

getaaron:
thanks for the PR and the flaky test fix :)

DuMaM:
Great news 🎉 🦖 
Thanks for merge and have a nice day 😄

DuMaM:
@getaaron I don't know why but ubuntu run failed 😱

DuMaM:
For me it looks like this second ulock_keychain is failing because we got an race condition here.

getaaron:
@DuMaM I tried rerunning the build on the main branch so we can check if it fails consistently https://app.circleci.com/pipelines/github/fastlane/fastlane?branch=master

DuMaM:
@getaaron It another flaky test. I guess they was not designed to run i parallel.
https://github.com/fastlane/fastlane/pull/20448-> i think this will helps

grey442:
I've got a problem with that changes

My case: Two different machines which have installed different wwdr certificate, but have same name "Apple Worldwide Developer Relations Certification Authority". As consuqunce sharing profiles works only on one of two machine.

That code `FastlaneCore::CertChecker.install_wwdr_certificate` checks only name installed certifacates. Why don't add checking expiration date? Because `FastlaneCore::CertChecker.install_wwdr_certificate` install explicit certificate "https://www.apple.com/certificateauthority/AppleWWDRCAG6.cer". 

For example:
`    def self.wwdr_certificate_installed?`
  `      certificate_name = "Apple Worldwide Developer Relations Certification Authority"`
  `      keychain = wwdr_keychain`
  `      date = "Not After : Mar 19 00:00:00 2036 GMT"`
  `      response = Helper.backticks("security find-certificate -a -c '#{certificate_name}' -p #{keychain.shellescape} | openssl x509 -text | grep '#{date}'", print: FastlaneCore::Globals.verbose?)`
  `      certs = response.split("\n")`
  `      certs.count >= 1`
`    end`

getaaron:
Good point, could you open a PR with this change?

PR #20452, opened 2022-07-08
[match] fixes set up bugs when using GitLab Secure Files as a Match storage backend

darbyfrey:
### Checklist

### Motivation and Context
This change fixes an error when attempting to upload files via Match to GitLab Secure Files. The error is caused when `human_readable_description` is called during the logging process.

This change also fixes some bugs in the set up process `fastlane match init`, and adds better error handling for file upload errors.

### DIFF
--- a/match/lib/match/storage/gitlab/client.rb
+++ b/match/lib/match/storage/gitlab/client.rb
@@ -69,7 +69,23 @@ module Match
               "name" => target_file
             )
 
-            execute_request(url, request)
+            response = execute_request(url, request)
+
+            log_upload_error(response, target_file) if response.code != "201"
+          end
+        end
+
+        def log_upload_error(response, target_file)
+          begin
+            response_body = JSON.parse(response.body)
+          rescue JSON::ParserError
+            response_body = response.body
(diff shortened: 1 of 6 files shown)


match/lib/match/storage/gitlab/client.rb
```diff
@@ -69,7 +69,23 @@ def upload_file(current_file, target_file)
...
+        end
+
+        def log_upload_error(response, target_file)
+          begin
+            response_body = JSON.parse(response.body)
+          rescue JSON::ParserError
+            response_body = response.body
+          end
+
+          if response_body["message"]["name"] == ["has already been taken"]
```
getaaron, line 85: Is `message` guaranteed to be in the JSON body? I think this would raise an error otherwise. Also looks like this could be a string if line 82 runs.

darbyfrey: Good call, this section is a bit clunky. I added an additional check to the conditional to ensure `message` is present. I also added more tests around that bit of code.

getaaron (approved):

PR #20498, opened 2022-07-22
[scan] Stop require `project` when providing `xctestrun` path

skarol:
### Checklist

### Motivation and Context

The problem exist when you want to run `scan` with `test_without_building` and `xctestrun`. Fastlane doesn't allow to do so and require `project` parameter. It's inconvinient on CI when previous job runs `scan` with `build_for_testing` and test job doesn't have access to whole project file, it only has derived data and `xctestrun` files.

### Description

Add a check if `xctestrun` parameter is set and in this case omit detecting project path.

### DIFF
--- a/scan/lib/scan/detect_values.rb
+++ b/scan/lib/scan/detect_values.rb
@@ -15,8 +15,8 @@ module Scan
 
       prevalidate
 
-      # Detect the project if not SPM package
-      if Scan.config[:package_path].nil?
+      # Detect the project if not SPM package and not running with: test_without_building and xctestrun
+      if Scan.config[:package_path].nil? && !(Scan.config[:test_without_building] && Scan.config[:xctestrun])
         FastlaneCore::Project.detect_projects(config)
         Scan.project = FastlaneCore::Project.new(config)
 
--- a/scan/spec/detect_values_spec.rb
+++ b/scan/spec/detect_values_spec.rb
@@ -36,6 +36,52 @@ describe Scan do
       end
     end
 
+    describe 'test without building' do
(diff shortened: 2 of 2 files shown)


getaaron:
Thanks so much for the PR! Will it work if `xctestrun` is provided but `test_without_building` and `derived_data` are not provided? Should we check for all 3?

Also could you please add a few tests to cover the various cases? They can be pretty short similar to this one:

https://github.com/fastlane/fastlane/blob/f221c959f595bd88e011454b533cc08788769612/scan/spec/detect_values_spec.rb#L28-L37

skarol:
> Thanks so much for the PR! Will it work if `xctestrun` is provided but `test_without_building` and `derived_data` are not provided? Should we check for all 3?

@getaaron Yes, in the current implementation it'll work. Let me fix that by checking all 3 parameters and adding tests.

skarol:
@getaaron actually giving it a second thought, I'm not sure passing `derived_data` is needed. Doesn't `scan` use a default path?

getaaron:
> @getaaron actually giving it a second thought, I'm not sure passing `derived_data` is needed. Doesn't `scan` use a default path?

Yeah I think we could leave `derived_data` out. But if `xctestrun` is provided but `test_without_building` is not I _think_ we'd still want to detect the project files?

skarol:
Updated with the suggestion from @getaaron

skarol:
what is the process to get it merged? PR is ready for over a month now 🙃

esteluk (approved):
I'd love this 😅

PR #20570, opened 2022-08-17
[gym] Disregard Symlinks When Detecting Xcode Installations in Error Handler

TheMetalCode:
### Checklist

### Motivation and Context

When a `gym` build error occurs and the given machine has multiple versions of xcode, it's very useful to output instructions to ensure the correct version is selected. 

The problem is that when using a tool like `xcode-install` to install Xcode on CI executor machines, what occurs is that the `.app` file with the version name is placed into `/Applications` (eg. `/Applications/Xcode-13.4.1.app`), and then `/Applications/Xcode.app` is created as a symlink to the former.  Because the `gym` error handler at present simply globs the `/Applications` dir and counts the files, we end up with a situation where the machine _in reality_ only has one version of xcode installed, but anytime a `gym` build error occurs, the fastlane output nonetheless outputs xcode path instructions as though there were multiple xcode installations, and often enough in my experience iOS developers having issues with their CI builds assume this is the issue because it's the thing they see most prominently in the fastlane output, and it ends up being one more thing I tell them to politely ignore.

I think it would be better to take the symlinks in `/Applications` out of the equation and output the xcode path instructions only when there are multiple distinct `.app` files in the `/Applications` path and using a different xcode version than intended is more likely to be the cause of the build error.

### Description

The change is fairly simple - I placed a `.reject { |filepath| File.symlink?(filepath) }` at the end of the `Dir.glob(File.join(default_xcode_path, "Xcode*.app"))` statement in `def print_xcode_path_instructions`, so that `xcode_installations_in_default_path.count  > 1` will trigger for non-symlinks only, aka distinct Xcode installations.

I added three specs to `gym/spec/error_handler_spec.rb` that make explicit assertions about which conditions should trigger the `UI` statements in `def print_xcode_path_instructions` to execute, with some fairly basic stubbing to ensure that the tests don't depend on any _actual_ Xcode installations a machine may have.

To make the stubbing in the specs I added more straightforward, I changed `Dir[]` to `Dir.glob`, which should be fine for all versions of ruby within the `fastlane` gems minimum requirements (`spec.required_ruby_version = '>= 2.6'`).

### Testing Steps

See `Desciption` for explanation of rspec tests added.


TheMetalCode:
Update: this change may or may not still be worthwhile, but this isn't an issue I've ever run into since retiring `xcode-install` in favor of `xcodes`. `xcodes` doesn't produce symlinks in the `/Applications` directory.

PR #20676, opened 2022-09-20
[match][sigh] add support for Apple Silicon Macs in iOS/iPadOS provisioning profiles

### LINKED ISSUE #18581: MATCH does not comply with Apple (developer) profiles for iOS Apps running on Apple Silicon (M1)
### New Issue Checklist

### Issue Description
To be able to run iOS Apps on Apple Silicon / M1 (Target: _My Mac (Designed for iPad)_)  you have to enable "_include Mac devices_" and include the desired device in the Apple Developer Profile.
Of course, the **UDID** of the M1-device had to be registered beforehand :)

However, if If I run `fastlane match` with `--force_for_new_devices` or `--force` the previously included device gets **deselected/excluded** from the profile and there is no option to avoid it.
btw: The enabled "_include Mac devices_" switch stays enabled in the profile.

So in short: `fastlane match` with `--force_for_new_devices` or `--force` filters out all macos devices on an iOS profile even though we want them to be included to be able to run iOS Apps on Target: _My Mac (Designed for iPad)_

related to #17690

##### Command executed
`fastlane match` with `--force_for_new_devices` or `--force`

The command executes successfully.

### Environment

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 11.2.3                                      |
| Ruby                        | 2.7.2                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.24.3 (Apple Git-128)          |
| Installation Source         | ~/.rbenv/versions/2.7.2/bin/fastlane        |
| Host                        | macOS 11.2.3 (20D91)                        |
| Ruby Lib Dir                | ~/.rbenv/versions/2.7.2/lib                 |
| OpenSSL Version             | OpenSSL 1.1.1j  16 Feb 2021                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 12.4                                        |

</details>

*generated on:* **2021-04-20**

Screenshot of an Apple developer profile highlighting the _"include Mac devices"_ functionality, that `fastlane match` does not comply with:
apple_dev_provisioning_screenshot

One additional note:
The registered (M1) macos device is not a "normal" macos device. We have to register the **UDID** of the macos device NOT the UUID. The provisioning UDID looks similar to an iOS UUID but looks completely different from a macos UUID.
</details>

PinkidG:
### Checklist

### Motivation and Context

Resolves #18581

Macs with Apple Silicon (AS) are capable of running native iOS/iPadOS apps. During development this only works if the Mac is added to the provisioning profile (`development` or `ad-hoc`).

### Description

This PR introduces a new device-class (`APPLE_SILICON_MAC`) that is used for iOS/iPadOS provisioning profiles to include AS Mac hardware. Including these devices is required to natively run apps on AS during development or using an ad-hoc signing. 

To include AS Macs, a new parameter has been added to match (and sigh): `include_mac_in_profiles`. Default is `false`. If enabled the provisioning profile device count and newly generated profiles will include registered AS Macs. 

If no AS Mac is registered in the Developer Portal, this setting won't have any effect. If run again without the `include_mac_in_profiles` flag, it will regenerate the provisioning profile without the AS Macs. (So it makes sense to include this setting in the `Matchfile` to keep it consistent)

### Testing Steps

1. Make sure to have at least one Apple Silicon Mac registered in the Developer Portal.
2. Create a provisioning profile with `fastlane match development --include_mac_in_profiles`
3. Check in Developer Portal that the newly created profile contains the AS Mac
4. Rerunning `fastlane match` without `include_mac_in_profiles` set, will regenerate the profile without the AS Macs.


match/lib/match/options.rb
```diff
@@ -253,6 +253,11 @@ def self.available_options
                                      description: "Renew the provisioning profiles if the device count on the developer portal has changed. Ignored for profile types 'appstore' and 'developer_id'",
                                      type: Boolean,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :include_mac_in_profiles,
+                                     env_name: "INCLUDE_MAC_IN_PROFILES",
```
rogerluan, line 257:
```suggestion
                                     env_name: "MATCH_INCLUDE_MAC_IN_PROFILES",
```

rogerluan (changes requested):
Thank you for this PR @PinkidG ! 💪 

Could you update the related specs? e.g. `sigh/spec/runner_spec.rb` and `match/spec/runner_spec.rb` (not sure if there are others) 🙏

PinkidG:
@rogerluan Thanks for reviewing it! I added changes to the related specs, though I'm not sure if the `sigh/spec/runner_spec.rb` is correct.

sigh/spec/runner_spec.rb
```diff
@@ -151,6 +151,23 @@
...
+        allow(ios_device).to receive(:device_class).and_return(Spaceship::ConnectAPI::Device::DeviceClass::IPHONE)
+        as_device = "as_device"
+        allow(as_device).to receive(:id).and_return(2)
+        allow(as_device).to receive(:device_class).and_return(Spaceship::ConnectAPI::Device::DeviceClass::APPLE_SILICON_MAC)
+
+        expect(Spaceship::ConnectAPI::Device).to receive(:all).and_return([ios_device, as_device])
+
+        devices = fake_runner.devices_to_use
+        expect(devices.size).to eq(2)
+      end
```
rogerluan, line 169: This works but I kinda have the feeling you're `allow`ing too many statements here 😬 not sure if this really attests the functionality you implemented, what do you think?

juliengdt:
Ahoy, bump for this PR. it's a must have feature !

thanks !!

sigh/spec/runner_spec.rb
rogerluan: Ping @PinkidG , WDYT? :)

PinkidG: Hey @rogerluan Thanks for the feedback, I'm sorry I didn't get back to it. I'm not really a ruby expert tbh. 😅

I'll try to improve it 🙂

rogerluan (approved):
Thanks for this, @PinkidG ! 🤗

joshdholtz (approved):
This is great! I added one commit with a comment about the device class not being officially available in ASC API yet but other than that this is 🔥

zakkhoyt:
Hi @PinkidG.  Thanks for putting this PR together.

Question for you (or anyone else knowledgeable).

I couldn't get my AS macs registered by using this PR, and feel like I'm missing some information. 

From the PR Testing Steps: 
> 1) Make sure to have at least one Apple Silicon Mac registered in the Developer Portal.

Looking through the sigh/match code it appears to me that for this PR to work that`device.device_class` must be `"APPLE_SILICON_MAC"`. Is this correct? If so, how do you accomplish this? 

When adding a new device with the Apple developer portal, you must choose from a drop list. 

zakkbookmax2022-11-16_00 33 11@2x

The fastlane tools [register_devices](https://docs.fastlane.tools/actions/register_devices/) and [register_device](https://docs.fastlane.tools/actions/register_device/) don't allow you to specify either. 

My team has a handful of AS macs registered using fastlane's [register_devices](https://docs.fastlane.tools/actions/register_devices/). They all are configured like this:

zakkbookmax2022-11-16_00 44 22@2x

Am I off track here or overlooking something? These [old tricks](https://stackoverflow.com/a/74417665/803882) still work, but would love to not have to do that anymore. 

Thx for reading.

ArtFeel:
He @PinkidG, thanks for PR, but I have same problem as described by @zakkhoyt

I found that if you register your device as a `macOS` (platform). Then it will not be added to the generated provision profile. It doesn't matter which `UUID` or `UDID` you specify. `UDID`-based AS mac is visible as option in iOS profile, but it remains unchecked.

But if you register your mac `Provisioning UDID` as `iOS, tvOS, watchOS` (platform).  
Then it will be added to generated profile and be visible in devices list as iPod 🤯 :

image
In generated profile:
image

Fastlane params:
```
+-------------------------------------+----------------------------------+
|                        Summary for sigh 2.211.0                        |
+-------------------------------------+----------------------------------+
| app_identifier                      | my.id.app                        |
| username                            | user@apple.com                   |
| force                               | true                             |
| cert_id                             | **********                       |
| provisioning_name                   | match Development my.id.app      |
| ignore_profiles_with_different_name | true                             |
| api_key                             | ********                         |
| team_name                           | ****** ***                       |
| fail_on_name_taken                  | false                            |
| include_all_certificates            | false                            |
| include_mac_in_profiles             | true                             |
| platform                            | ios                              |
| development                         | true                             |
| adhoc                               | false                            |
| developer_id                        | false                            |
| skip_install                        | false                            |
| skip_fetch_profiles                 | false                            |
| skip_certificate_verification       | false                            |
| readonly                            | false                            |
+-------------------------------------+----------------------------------+
```

jaysoffian:
There's apparently a bug, or at least inconsistency in the App Store Connect API as of this comment.

It turns out that when you query the API authorized using an API key, it returns `MAC` as the `deviceClass` for all Macs, whether Intel or Apple Silicon. If you query the API using a username/password, it returns the exact same devices with a `deviceClass` of either `INTEL_MAC` or `APPLE_SILICON_MAC`.

I've worked around this in a custom action by checking the length of the UDID:

```
        Spaceship::ConnectAPI::Device.all.select(&:enabled?).each do |d|
          c = d.device_class
          case
          when ios_device_classes.include?(c) then ios_devices.push(d)
          when tvos_device_classes.include?(c) then tvos_devices.push(d)
          when mac_arm_device_classes.include?(c) then mac_arm_devices.push(d)
          when mac_intel_device_classes.include?(c) then mac_intel_devices.push(d)
          # When accessing the App Store Connect API using an API key, Mac devices are
          # returning using a generic "MAC" device class instead of a more specific class of
          # "APPLE_SILICON_MAC" or "INTEL_MAC" as is the case when accessing the API using a
          # username/password. As a work-around, we use the length of the UDID to discern
          # Intel Macs (UDID of length 36) from Apple Silicon Macs (UDID of length 25).
          when mac_generic_device_classes.include?(c) then
            case d.udid.length
            when 25 then mac_arm_devices.push(d)
            when 36 then mac_intel_devices.push(d)
            else UI.message("Unexpected UDID length: #{d.udid}")
            end
          else UI.message("Unexpected device class: #{c} (#{d.udid})")
          end
        end
```

Where:

```
      def self.ios_device_classes
        [
          Spaceship::ConnectAPI::Device::DeviceClass::APPLE_WATCH,
          Spaceship::ConnectAPI::Device::DeviceClass::IPAD,
          Spaceship::ConnectAPI::Device::DeviceClass::IPHONE,
          Spaceship::ConnectAPI::Device::DeviceClass::IPOD
        ]
      end

      def self.tvos_device_classes
        [Spaceship::ConnectAPI::Device::DeviceClass::APPLE_TV]
      end

      def self.mac_arm_device_classes
        [Spaceship::ConnectAPI::Device::DeviceClass::APPLE_SILICON_MAC]
      end

      def self.mac_intel_device_classes
        ["INTEL_MAC"]
      end

      def self.mac_generic_device_classes
        [Spaceship::ConnectAPI::Device::DeviceClass::MAC]
      end
```

PR #20870, opened 2022-11-17
[dependency] relax `multipart_post` dependency version requirement

edouard:
### Checklist

### Motivation and Context

The gem `multipart_post` is currently set to `~> 2.0.0` on fastlane. Since `2.0.0` is the only version released on the `2.0` series, it effectively locks `multipart_post` to `2.0.0` which is a version that was released in 2013. When users bundle fastlane with other dependencies (for instance with the [web_translate_it](https://github.com/webtranslateit/webtranslateit) gem which requires a newer version of `multipart_post`), running `bundle` fails. See https://github.com/webtranslateit/webtranslateit/issues/212 and https://github.com/webtranslateit/webtranslateit/issues/220

We’ve mitigated the issue on our end by relaxing our dependency to multipart_post to let users use older versions, but this almost 10 years old dependency should be updated on your side as well.

### Description

I've relaxed the dependency on `multipart_post` in the gemspec file. I've checked where multipart_post is used on the code and I haven’t noted any changes required in the code.

### DIFF
--- a/fastlane.gemspec
+++ b/fastlane.gemspec
@@ -78,7 +78,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('plist', '>= 3.1.0', '< 4.0.0') # Needed for set_build_number_repository and get_info_plist_value actions
   spec.add_dependency('CFPropertyList', '>= 2.3', '< 4.0.0') # Needed to be able to read binary plist format
   spec.add_dependency('addressable', '>= 2.8', '< 3.0.0') # Support for URI templates
-  spec.add_dependency('multipart-post', '~> 2.0.0') # Needed for uploading builds to appetize
+  spec.add_dependency('multipart-post', '>= 2.0.0', '< 3.0.0') # Needed for uploading builds to appetize
   spec.add_dependency('word_wrap', '~> 1.0.0') # to add line breaks for tables with long strings
 
   spec.add_dependency('optparse', '~> 0.1.1') # Used to parse options with Commander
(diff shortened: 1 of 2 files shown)


edouard:
Is there any way to restart the CI check on [AppVeyvor](https://ci.appveyor.com/project/KrauseFx/fastlane/builds/45410034)? It failed for some reason but cannot restart it.

fastlane.gemspec
```diff
@@ -78,7 +78,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('plist', '>= 3.1.0', '< 4.0.0') # Needed for set_build_number_repository and get_info_plist_value actions
   spec.add_dependency('CFPropertyList', '>= 2.3', '< 4.0.0') # Needed to be able to read binary plist format
   spec.add_dependency('addressable', '>= 2.8', '< 3.0.0') # Support for URI templates
-  spec.add_dependency('multipart-post', '~> 2.0.0') # Needed for uploading builds to appetize
+  spec.add_dependency('multipart-post', '>= 2.0.0', '< 3.0.0') # Needed for uploading builds to appetize
```
rogerluan, line 81: Initially, I thought `~> 2.0.0` and `>= 2.0.0, < 3.0.0` were equivalent, but I see the difference now. I suppose we could change `~> 2.0.0` to `~> 2`, which is simpler, but that would go against the code style pattern being used nearly-throughout this file so I think this is better :)

rogerluan (approved):
I have no objections here as long as it passes CI 😊 Thanks for fixing this, @edouard !

rogerluan:
@edouard could you rebase this branch, to trigger CI again, please? 🙏

fastlane.gemspec
edouard: Yes, I followed the style I saw in the rest of that `.gemspec` file.

edouard:
> @edouard could you rebase this branch, to trigger CI again, please? 🙏

Done! Thank you for reviewing it!

PR #21018, opened 2023-01-30
[match] Added s3_skip_encryption parameter

mbogh:
Furthermore fixed a bug in parameters added to for_storage_mode

### Checklist

### Motivation and Context

We plan to store all certificates and provisioning profiles in S3 with custom tailored IAM Roles for each team/project/app that can be assumed via OIDC and it would be complicated and not so secure, to have to share a encryption key as well across GitHub, GitLab and other.

### Description

Added a single parameter to disable/skip encryption if you are using S3 storage mode. I have had to add everywhere `Encryption.for_storage_mode(...` is invoked.

In order to test this, I created two S3 buckets on my AWS account and ran:  
`bundle exec fastlane match import --skip_certificate_matching true -y appstore -b XYZ` with `s3_skip_encryption(true)` and with `s3_skip_encryption(false)`.
Went into the AWS Console and browsed the S3 buckets, downloaded the imported files and could verify that the encrypted ones were in fact still encrypted and the unencrypted ones plain .cer, .p12.

Then I tested it from a lane:
```ruby
sync_code_signing(
  storage_mode: "s3",
  readonly: true,
  type: "appstore",
  app_identifier: "com.example.fake",
  s3_region: "my-region",
  s3_access_key: "ACCESSKEY!!!!",
  s3_secret_access_key: "mySecretAccessKey",
  s3_bucket: "fastlane-match-unencrypted-bucket",
  s3_skip_encryption: true
)
```

### Testing Steps

1. Create two S3 buckets
2. Obtain access key for your account
3. Configure `Matchfile` with
```ruby
storage_mode("s3")
s3_region("your-region")
s3_access_key("XYZ")
s3_secret_access_key("xyz")
s3_bucket("your-bucket")
s3_skip_encryption(true)
```
4. `bundle exec fastlane match import --skip_certificate_matching true -y appstore -b XYZ`
5. Toggle `s3_skip_encryption` and change bucket
6. `bundle exec fastlane match import --skip_certificate_matching true -y appstore -b XYZ`
7. Go to https://console.aws.amazon.com and browse your buckets.
8. Download files and examine.


mbogh:
Is there any way bump this 🥹

match/lib/match/options.rb
```diff
@@ -222,6 +222,11 @@ def self.available_options
                                      env_name: "MATCH_S3_OBJECT_PREFIX",
                                      description: "Prefix to be used on all objects uploaded to S3",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :s3_skip_encryption,
+                                     env_name: "MATCH_S3_SKIP_ENCRYPTION",
+                                     description: "Skip encryption of all objects uploaded to S3",
```
getaaron, line 227: I think we should add a warning… maybe something like this?

```suggestion
                                     description: "Skip encryption of all objects uploaded to S3. WARNING: only enable this on S3 buckets with sufficiently restricted permissions and server-side encryption enabled. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingEncryption.html",
```

match/lib/match/nuke.rb
```diff
@@ -55,6 +55,7 @@ def run(params, type: nil)
         s3_secret_access_key: params[:s3_secret_access_key].to_s,
         s3_bucket: params[:s3_bucket].to_s,
         s3_object_prefix: params[:s3_object_prefix].to_s,
+        s3_skip_encryption: params[:s3_skip_encryption],
         gitlab_project: params[:gitlab_project],
         team_id: params[:team_id] || Spaceship::ConnectAPI.client.portal_team_id
       })
```
getaaron, line 61: I kinda dislike how the current design of `for_storage_mode` forces all these unrelated params to be littered across all the call sites... what do you think about refactoring it so we just pass in the whole `params` object plus any overrides and nil replacements? Then this entire block of code could be replace with something like

```ruby
self.storage = Storage
  .from_params(params)
  .replace_if_nil({team_id: Spaceship::ConnectAPI.client.portal_team_id})
```

getaaron:
Added a few questions, also

> Furthermore fixed a bug in parameters added to for_storage_mode

what's the bug? I didn't notice it 😅

match/lib/match/nuke.rb
mbogh: Fine with me, I can give it a go. It was kinda more changes than I thought would be needed since I had to hand pick all params.

Stay tuned :)

mbogh:
> Added a few questions, also
> 
> > Furthermore fixed a bug in parameters added to for_storage_mode
> 
> what's the bug? I didn't notice it 😅

So the `s3_bucket` was never sent in to the Encryption class and therefore it wouldn't be stored in the keychain, at least that is how read the code.

`params[:keychain_name] = params[:s3_bucket]`

match/lib/match/nuke.rb
mbogh: Okay, so the `params` object in this case is a `FastlaneCore::Configuration` and that does not play nicely with the existing tests... I will see if it something I can work around.

mbogh: @getaaron I gave it a swing, found that it made the most sense to cherry pick the params going into the Storage classes themselves since they are more specific.

Please give me your thoughts.

mbogh:
@getaaron any chance of a second look at the changes?

mbogh:
Anything I can do to get this PR rolling?

giovannidegani:
Would be great to get this one in

mbogh:
Possible to get some eyes on this?

getaaron:
@mbogh I merged `master` in to resolve some merge conflicts. I also had to update the importer_spec since it was asserting  `GitStorage` would receive params that it no longer does. Could you take a look and let me know if it's good to merge?

@darbyfrey Could you double-check the GitLab-related changes? For this part, I don't think there should be any behavior change, it's just refactoring to avoid sending unnecessary params.

darbyfrey:
Nice improvements @mbogh! 

The GitLab storage mode changes look good. I've tested them locally and everything works as expected.

getaaron:
@mbogh I think this is good to go except a few tests are still failing

mbogh:
@getaaron all green again

getaaron (approved):

mbogh:
Thanks for your work on Fastlane @getaaron 🤗

MichalAlgor:
Hi, @mbogh @getaaron 
After updating from version 2.214.0 to 2.215.0 match stopped working on Gitlab CI, the error printed is 

> Could not find option 'job_token' in the list of available options: type, additional_cert_types, readonly, generate_apple_certs, skip_provisioning_profiles, app_identifier, api_key_path, api_key, username, team_id, team_name, storage_mode, git_url, git_branch, git_full_name, git_user_email, shallow_clone, clone_branch_directly, git_basic_authorization, git_bearer_authorization, git_private_key, google_cloud_bucket_name, google_cloud_keys_file, google_cloud_project_id, skip_google_cloud_account_confirmation, s3_region, s3_access_key, s3_secret_access_key, s3_bucket, s3_object_prefix, s3_skip_encryption, gitlab_project, gitlab_host, keychain_name, keychain_password, force, force_for_new_devices, include_mac_in_profiles, include_all_certificates, force_for_new_certificates, skip_confirmation, safe_remove_certs, skip_docs, platform, derive_catalyst_app_identifier, template_name, profile_name, fail_on_name_taken, skip_certificate_matching, output_path, skip_set_partition_list, verbose

I believe this has something to do with this change, but I have no Ruby experience to find the problem

getaaron:
@MichalAlgor thanks for the report — @darbyfrey said he can take a look today and we'll do a release shortly after

darbyfrey:
@MichalAlgor apologies for the regression! I've got a fix ready in https://github.com/fastlane/fastlane/pull/21520.

MichalAlgor:
@darbyfrey no worries, glad I could help, thanks for quick fix 🙂

match/lib/match/storage.rb
```diff
@@ -39,7 +89,8 @@ def register_backend(type: nil, storage_class: nil, &configurator)
         end
       end
 
-      def for_mode(storage_mode, params)
+      def from_params(params)
```
rogerluan, line 92: It seems like this change wasn't made backwards compatible 😬😬 
This was brought to my attention in https://github.com/fastlane/docs/pull/1225

mbogh: Is there anything I can do to help?

rogerluan: Yeah fair enough. Not sure if we have any guidelines on what to consider important to keep backwards compatible and what not to, so that's to blame haha
Whenever I personally make changes or review code I try to keep _all_ APIs backwards compatible, but that may be exclusive to me and not really the team's preferences

I'll kick off a discussion around this in Slack so perhaps we can better define this :) 

Thanks for the input guys!

PR #21088, opened 2023-02-28
[ci] Only run all mac jobs on master and version bump branches

joshdholtz:
### Motivation and Context

We ran out of CircleCI credits last month for macOS jobs so trying to lighten our ussage a little bit.

### Description

Only run full macOS suite on `master` and `version-bump-<x.y.x>` branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,12 @@ orbs:
     shellcheck: circleci/shellcheck@2.2.2 # brew install shellcheck stopped working so using this
 
 aliases:
+  - &important-branches
+    filters:
+      branches:
+        only:
+          - main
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -258,16 +264,19 @@ workflows:
           name: 'Execute tests on macOS (Xcode 11.7.0, Ruby 2.6)'
           xcode_version: '11.7.0'
           ruby_version: '2.6'
+          <<: *important-branches
```
ainame, line 267: 👍 
https://circleci.com/docs/configuration-reference/#jobfilters

ainame (approved):
`main` needs to be `master` otherwise LGTM

PR #21128, opened 2023-03-11
[action][spm] add `--very-verbose` option

att55:
### Checklist

### Description

Hi there!

I found SPM has very-verbose option.

```
--very-verbose, --vv    Increase verbosity to include debug output
```

I also added to the option to spm action.


att55:
I didn't add a short option because the option accepts only string length 1.
https://github.com/fastlane/fastlane/blob/286aa9be631a958f90597e548274fa072279984f/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L21-L22

Should I add `--vv` after changed the length of option from 1 to 2 ?

mollyIV:
Hello @att55 👋 

Thanks for the contribution 💯 

> Should I add --vv after changing the length of option from 1 to 2 ?

According to the [`config_item`](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb) codebase, [we cannot pass more than 1 character as a shortcut](https://github.com/fastlane/fastlane/blob/master/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L123-L126) 👇 

```ruby
if short_option
        UI.user_error!("short_option for key :#{key} must of type String") unless short_option.kind_of?(String)
        UI.user_error!("short_option for key :#{key} must be a string of length 1") unless short_option.delete('-').length == 1
end
```

and to be honest, I'd stick to it 😉 

If we'd like to support a short option, maybe we could just use `-V` (upper-cased V). 

`short_option` requires being a `String`, which is case-sensitive 👌 

There are existing actions that leverage the same letter in upper and lower cased form, for example [pilot](https://github.com/fastlane/fastlane/blob/master/pilot/lib/pilot/options.rb)'s `-d` and `-D`.

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -10,6 +10,7 @@ def self.run(params)
         cmd << "--configuration #{params[:configuration]}" if params[:configuration]
         cmd << "--disable-sandbox" if params[:disable_sandbox]
         cmd << "--verbose" if params[:verbose]
+        cmd << "--very-verbose" if params[:very_verbose]
```
mollyIV, line 13: Can we please add a unit test for it?  
You can take an inspiration from an existing tests in `spm_spec` for `--vebose` flag 💪

lacostej (approved):
Looks good to me. Bonus point for the unit test if you add i.t

fastlane/lib/fastlane/actions/spm.rb
att55: Sorry for late reply 🙏🏼 
I will add a unit test!

rogerluan (commented):
This PR also LGTM 😊 just missing the unit tests really @att55 🙇 

Also, there are merge conflicts that need to be solved before we can merge this 🙏 

Thank you!

att55:
@rogerluan @mollyIV 
I'm so sorry for late reply. 😢
I added unit tests! Could you please check this? 🙏🏼

lucgrabowski (approved):
LGTM :)

PR #20623, opened 2022-09-06
[match] AWS Secrets Manager support

stanogn:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I see several green `ci/circleci` builds in the "All checks have passed" section of my PR ([connect CircleCI to GitHub](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-) if not)
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary. (https://github.com/fastlane/docs/pull/1179)

### Motivation and Context
It adds support for AWS Secrets manager as a storage backend.
S3 is as secure if configured properly, but it has a bad reputation.

### Description
Added AWS Secrets manager as a storage backend. It follows the same principles as the S3 backend.
Tested through unit tests and inside a live project.

### Testing Steps
Configure secrets manager as the backend in the Matchfile.

PR #21131, opened 2023-03-13
[match] Flexibility improvements

stanogn:
### Checklist

### Motivation and Context

The main issue is that Match doesn't work that good outside of the "golden-path". Someone might have their account already set up with some existing apps or may already have a 3rd party involved also using fastlane for other apps, etc.. in their account and we cant just nuke it. We also need to support people importing their certs manually regularly without stuff breaking.

This is currently not supported and results in a big mess.

The root cause is that match heavily relies on certain preconditions and does not perform more complex checks like actually checking certificate, provisioning profile contents and matching them according to it.

This leads to problems when people import manually created certificates as all crucial info on which match relies right now such as the Apple Cert ID (contained in the name when created by match using API access, not present at all when created manually) is lost.

### Description

The main changes include using the serial number of certificates/profiles to check for their validity on the dev portal (instead of the file name - apple cert id/provisioning profile name) and matching/selecting provisioning profiles by checking the embedded certificates (instead of relying on a single predetermined workflow and selecting the first one present in storage).

The unit tests were edited to account for the changes. It was also tested live in our dev environment.

### Testing Steps

Since the main goal is to make it more robust in all situations, not being able to "break" it by not following the match docs steps should be counted as a success. 
Concrete test cases include: 
1. not using a clean account
2. not nuking
3. importing manually created certs and mixing it with API access to then automatically setup some more of them


Gemfile.lock
```diff
@@ -391,4 +395,4 @@ DEPENDENCIES
   yard (~> 0.9.11)
 
 BUNDLED WITH
-   2.3.14
+   2.3.11
```
getaaron, line 398: Minor issue, but let's avoid going backwards here

cert/lib/cert/runner.rb
```diff
@@ -114,8 +114,8 @@ def find_existing_cert
           return path
         elsif File.exist?(private_key_path)
           password = Cert.config[:keychain_password]
-          FastlaneCore::KeychainImporter.import_file(private_key_path, keychain, keychain_password: password, skip_set_partition_list: Cert.config[:skip_set_partition_list])
           FastlaneCore::KeychainImporter.import_file(path, keychain, keychain_password: password, skip_set_partition_list: Cert.config[:skip_set_partition_list])
+          FastlaneCore::KeychainImporter.import_file(private_key_path, keychain, keychain_password: password, skip_set_partition_list: Cert.config[:skip_set_partition_list])
```
getaaron, line 118: Why the order change here?

cert/lib/cert/runner.rb
```diff
@@ -226,8 +226,8 @@ def create_certificate
       # Import all the things into the Keychain
       keychain = File.expand_path(Cert.config[:keychain_path])
       password = Cert.config[:keychain_password]
-      FastlaneCore::KeychainImporter.import_file(private_key_path, keychain, keychain_password: password, skip_set_partition_list: Cert.config[:skip_set_partition_list])
       FastlaneCore::KeychainImporter.import_file(cert_path, keychain, keychain_password: password, skip_set_partition_list: Cert.config[:skip_set_partition_list])
+      FastlaneCore::KeychainImporter.import_file(private_key_path, keychain, keychain_password: password, skip_set_partition_list: Cert.config[:skip_set_partition_list])
```
getaaron, line 230: Same here

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -200,5 +200,18 @@ def self.sha1_fingerprint(path)
...
       UI.user_error!("Error parsing certificate '#{path}'")
     end
+
+    def self.certificate_key_match?(cert_path, key_path)
+      cert_md5 = `openssl x509 -inform der -noout -modulus -in '#{cert_path}' | openssl md5`.strip
+      key_md5_rsa = `openssl rsa -noout -modulus -in '#{key_path}' | openssl md5`.strip
+      temp_key_path = File.join(File.dirname(key_path), "-temp-key")
+      `openssl pkcs12 -in #{key_path} -nocerts -passin pass: -nodes -out #{temp_key_path}`
+      key_md5_pkcs12 = `openssl rsa -noout -modulus -in #{temp_key_path} | openssl md5`.strip
+      `rm #{temp_key_path}`
```
getaaron, line 210: Ruby has a built-in `OpenSSL` class, would it be possible to use this instead of shelling out to the command line?

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -200,5 +200,18 @@ def self.sha1_fingerprint(path)
...
+      puts("cert_md5: #{cert_md5}")
+      puts("key_md5_rsa: #{key_md5_rsa}")
+      puts("key_md5_pkcs12: #{key_md5_pkcs12}")
```
getaaron, line 213: Is it leftover from debugging or do we want to show the user this? If we want to show the user, we should use the `UI` class

fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -200,5 +200,18 @@ def self.sha1_fingerprint(path)
...
+      return (cert_md5 == key_md5_rsa) || (cert_md5 == key_md5_pkcs12)
```
getaaron, line 214: Might be a good time to drop support for MD5 or at least print a warning

fastlane_core/lib/fastlane_core/certificate.rb
```diff
@@ -0,0 +1,41 @@
...
+      #  "NotBefore"=>"Feb 21 16:39:53 2023 GMT"}
+      def order_by_expiration(certs, ascending = true)
+        certs.sort_by { |c| c["NotAfter"].to_i }.reverse! unless ascending
+        certs.sort_by { |c| c["NotAfter"].to_i }
+      end
+
+      def parse_from_b64(b64_content)
+        serial_number = `echo '#{b64_content}' | base64 --decode | openssl x509 -inform der -noout -serial`.split("=").last.strip
+        not_before = Time.parse(`echo '#{b64_content}' | base64 --decode | openssl x509 -inform der -noout -startdate`.split("=").last.strip)
+        not_after = Time.parse(`echo '#{b64_content}' | base64 --decode | openssl x509 -inform der -noout -enddate`.split("=").last.strip)
```
getaaron, line 19: We should avoid the command line here and use the `Base64` & `OpenSSL` classes

fastlane_core/lib/fastlane_core/certificate.rb
```diff
@@ -0,0 +1,41 @@
...
+          "NotBefore" => not_before,
+          "NotAfter" => not_after
+        }
+      end
+
+      def parse_from_file(path)
+        serial_number = `openssl x509 -inform der -noout -serial -in '#{path}'`.split("=").last.strip
+        file_name = File.basename(path).gsub(".cer", "").gsub(".p12", "").gsub(".pem", "")
+        not_before = Time.parse(`openssl x509 -inform der -noout -startdate -in '#{path}'`.split("=").last.strip)
+        not_after = Time.parse(`openssl x509 -inform der -noout -enddate -in '#{path}'`.split("=").last.strip)
```
getaaron, line 31: I'm going to stop mentioning it after this, but in general we should go through and remove code that shells out directly to the command line

match/lib/match/runner.rb
```diff
@@ -329,6 +359,23 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
...
 
+    def find_matching_profile(profiles, certificates, spaceship = nil)
+      profiles.filter do |profile_path|
+        # Check if it matched any of the certificates
+        found = false
+        certificates.each do |certificate|
+          UI.verbose("Checking if profile '#{profile_path}' includes certificate '#{certificate['FileName']}'")
+          next unless FastlaneCore::ProvisioningProfile.includes_certificate?(profile_path: profile_path, certificate: certificate)
+          UI.verbose("Checking if profile is valid on dev portal")
+          next unless
```
getaaron, line 370: this seems like it should do something to check if the profile is valid, or did i miss something?

match/spec/fixtures/existing/certs/distribution/E7P4EE896K.p12
```diff
@@ -1,24 +1,27 @@
 -----BEGIN RSA PRIVATE KEY-----
-MIIEogIBAAKCAQEAyTm/v40AZb6I1eXRPVaUF+q683gm+XTRaRC9fjqK3seL117n
```
getaaron, line 2: why does it change?

match/spec/runner_spec.rb
```diff
@@ -192,16 +207,16 @@
...
-                                                            type: "appstore")]).to eql('736590c3-dfe8-4c25-b2eb-2404b8e65fb8')
-          expect(ENV[Match::Utils.environment_variable_name_team_id(app_identifier: "tools.fastlane.app",
-                                                                    type: "appstore")]).to eql('439BBM9367')
-          expect(ENV[Match::Utils.environment_variable_name_profile_name(app_identifier: "tools.fastlane.app",
-                                                                         type: "appstore")]).to eql('match AppStore tools.fastlane.app 1449198835')
-          profile_path = File.expand_path('~/Library/MobileDevice/Provisioning Profiles/736590c3-dfe8-4c25-b2eb-2404b8e65fb8.mobileprovision')
-          expect(ENV[Match::Utils.environment_variable_name_profile_path(app_identifier: "tools.fastlane.app",
+          expect(ENV[Match::Utils.environment_variable_name(app_identifier: "com.jigx.test.ios",
+                                                            type: "appstore")]).to eql('2746d535-7d21-403f-8718-5deee23b2366')
+          expect(ENV[Match::Utils.environment_variable_name_team_id(app_identifier: "com.jigx.test.ios",
```
getaaron, line 212: I don't think we should change these assertions unless there's a good reason (we should keep `tools.fastlane.app` instead of `com.jigx.test.ios`)

sigh/lib/sigh/options.rb
```diff
@@ -130,6 +130,11 @@ def self.available_options
                                      env_name: "SIGH_CERTIFICATE_ID",
                                      description: "The ID of the code signing certificate to use (e.g. 78ADL6LVAA) ",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :cert_serial_number,
+                                    short_option: "-s",
+                                    env_name: "SIGH_CERTIFICATE_SERIAL_NUMBER",
+                                    description: "The Serial Number of the code signing certificate to use (e.g. 06:9f:ba:78:51:58:dc:46:d9:a6:60:3e:af:5a:13:16). Takes precedence over cert_id. For advanced use only",
```
getaaron, line 136: Most of our users are software developers, I think instead of saying `For advanced use only` it would be helpful to explain the circumstances in which they might want to match by serial number instead of cert id. (Why did you want to?)

getaaron (commented):
❣️ Thanks so much for the PR, I'm really excited about the improvements to _match_. It's been a long time coming I added a few initial suggestions.

One thought is you might want to split out the AWS Secret Manager work into a separate PR (See @darbyfrey's https://github.com/fastlane/fastlane/pull/20386 and https://github.com/fastlane/fastlane/pull/20452/files as a reference). I think that may be faster to merge, while the other match improvements may benefit from some more iteration and thought from other team members.

stanogn:
> ❣️ Thanks so much for the PR, I'm really excited about the improvements to _match_. It's been a long time coming I added a few initial suggestions.
> 
> One thought is you might want to split out the AWS Secret Manager work into a separate PR (See @darbyfrey's #20386 and https://github.com/fastlane/fastlane/pull/20452/files as a reference). I think that may be faster to merge, while the other match improvements may benefit from some more iteration and thought from other team members.

Hi, thanks for the feedback. I am going to work on it now.
I already created a PR for Secrets Manager support btw: https://github.com/fastlane/fastlane/pull/20623

match/spec/fixtures/existing/certs/distribution/E7P4EE896K.p12
stanogn: This is because we are actually checking the relation of a private key to a public key. This means we need a real key like the one i added.

match/spec/runner_spec.rb
stanogn: The reason is that i am using real certs/profiles. This is because match now checks the actual contents of the files.
I can probably somehow edit it so it looks nice without destroying the checking stuff.

stanogn: I did manage to remove it from tests but the profile itself (binary file) is required to be valid - i can't remove it from there.

stanogn:
I think i have addressed all your comments for now.

xotahal:
@getaaron Any thoughts about when this could be merged?

xotahal:
@getaaron Sorry to bother you again. We would really appreciate it if this could be moved forward. @stanogn will remove `com.jigx.test.ios` id from this PR.

Please let us know if we can improve something else to make the PR review easier for you.

getaaron:
@joshdholtz @darbyfrey @rogerluan @AliSoftware any of you able to take a second look at it?

darbyfrey:
Thanks for the ping @getaaron, and thanks for all the work on this change @stanogn! This looks like it will be a very nice improvement!

I left a few questions, but overall I think we need to ensure we don't introduce any regressions with a change like this. What do you think about improving the test coverage? Or, would there be a way to break down this PR into several smaller changes that could be delivered over a few iterations? One thought that comes to mind would be to split out `fetch_certificates` as a separate code path while keeping `fetch_certificate` active during the refactoring. Just an idea.

darbyfrey (changes requested):

match/lib/match/importer.rb
```diff
@@ -10,9 +10,9 @@ module Match
   class Importer
     def import_cert(params, cert_path: nil, p12_path: nil, profile_path: nil)
       # Get and verify cert, p12 and profiles path
-      cert_path = ensure_valid_file_path(cert_path, "Certificate", ".cer")
-      p12_path = ensure_valid_file_path(p12_path, "Private key", ".p12")
-      profile_path = ensure_valid_file_path(profile_path, "Provisioning profile", ".mobileprovision or .provisionprofile", optional: true)
+      cert_path = ensure_valid_file_path(cert_path || params[:cert_path], "Certificate", ".cer")
```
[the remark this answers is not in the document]
stanogn: Its part of a different PR. https://github.com/fastlane/fastlane/pull/21019
Should i remove it for this PR?

match/lib/match/runner.rb
```diff
@@ -298,37 +323,56 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
...
-                                                                                platform: params[:platform]),
+                                                                          type: prov_type,
+                                                                          platform: params[:platform]),
                              installed_profile)
 
       return uuid
     end
     # rubocop:enable Metrics/PerceivedComplexity
 
+    def find_matching_profile(profiles, certificates, spaceship = nil)
```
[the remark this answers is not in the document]
stanogn: Should not be there.

fastlane_core/lib/fastlane_core/certificate.rb
```diff
@@ -0,0 +1,37 @@
...
+    class << self
+      # @return (Hash) The hash with the data of the certificate
+      # @example
+      #  {"SerialNumber"=>"069FBA785158DC46D9A6603EAF5A1316"
+      #  "FileName"=>"certificate-13",
+      #  "NotAfter"=>"Feb 21 16:39:53 2024 GMT",
+      #  "NotBefore"=>"Feb 21 16:39:53 2023 GMT"}
+      def order_by_expiration(certs, ascending = true)
+        certs.sort_by { |c| c["NotAfter"].to_i }.reverse! unless ascending
+        certs.sort_by { |c| c["NotAfter"].to_i }
```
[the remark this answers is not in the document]
stanogn: Mistake, a return is missing.

fastlane_core/lib/fastlane_core/certificate.rb
```diff
@@ -0,0 +1,37 @@
...
+        openssl_object = OpenSSL::X509::Certificate.new(Base64.decode64(b64_content))
+        {
+          "SerialNumber" => openssl_object.serial.to_s(16).upcase,
+          "NotBefore" => openssl_object.not_before,
+          "NotAfter" => openssl_object.not_after
+        }
+      end
+
+      def parse_from_file(path)
+        file_name = File.basename(path).gsub(".cer", "").gsub(".p12", "").gsub(".pem", "")
```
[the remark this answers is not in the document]
stanogn: Not sure why but originally in runner.rb (commit 8b2cd4f) this was done and the filename without the extension was passed along to generate_profile etc. so i just kept it like that. I did not name it certificate_id (like it was named before) because its a bit misleading. I could maybe keep the file extension here and remove it just before passing it on in runner.rb. Would that be better?

stanogn:
> Thanks for the ping @getaaron, and thanks for all the work on this change @stanogn! This looks like it will be a very nice improvement!
> 
> I left a few questions, but overall I think we need to ensure we don't introduce any regressions with a change like this. What do you think about improving the test coverage? Or, would there be a way to break down this PR into several smaller changes that could be delivered over a few iterations? One thought that comes to mind would be to split out `fetch_certificates` as a separate code path while keeping `fetch_certificate` active during the refactoring. Just an idea.

Thanks for the review.
I agree, it does feel a bit fragile considering its complexity.
I dont think it would be easy to break it down into multiple iterations without adding even more complexity. I would rather add some more test cases.

match/lib/match/encryption/openssl.rb
```diff
@@ -137,7 +98,7 @@ def encrypt_specific_file(path: nil, password: nil)
 
       # The encryption parameters in this implementations reflect the old behavior which depended on the users' local OpenSSL version
       # 1.0.x OpenSSL and earlier versions use MD5, 1.1.0c and newer uses SHA256, we try both before giving an error
-      def decrypt_specific_file(path: nil, password: nil, hash_algorithm: "MD5")
+      def self.decrypt_specific_file(path: nil, password: nil, hash_algorithm: "MD5")
```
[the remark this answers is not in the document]
darbyfrey: Ah okay, that makes sense. I think if you revert that change all of the changes in `openssl.rb` would go away.

fastlane_core/lib/fastlane_core/certificate.rb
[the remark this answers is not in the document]
darbyfrey: Yea, it's hard to tell why that was done. I think it would be helpful to understand if it is necessary or not. Looking at the code I just see it used in one place to output a log, but it's possible it's coming back from an Apple API that way, so it could be done to align with that?

match/lib/match/importer.rb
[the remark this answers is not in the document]
darbyfrey: Yea, if it's not relevant to this PR, I'd move it to the other PR.

fastlane_core/lib/fastlane_core/certificate.rb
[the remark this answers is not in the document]
stanogn: After some digging i found that its used to check if a cert thats on apple dev portal is available locally. Certs created by Spaceship have the cert_id as their file names (its the only way to actually get the cert_id and was used to identify the certs). Its not even being used anymore since the serial number is the "main" identifier now so i think it can be removed to avoid confusion.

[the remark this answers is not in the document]
stanogn: I think it can be safely removed since the log message doesnt say much anyways and the serial number is the main way to identify certs now.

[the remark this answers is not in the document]
stanogn: Actually nope, it is available as an input parameter to the sigh tool. I dont understand how it would be useful in practice though. The one and only way to get the cert id is when generating the certificate through the apple api (spaceship) which doesnt seem very practical. I guess we can just leave it as it is now.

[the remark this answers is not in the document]
darbyfrey: Okay, thanks for investigating. Would you mind adding a comment in the code with some details on where it's used, etc.? That way we have at least a little context there if we want to come back and refactor it in the future.

stanogn:
> > Thanks for the ping @getaaron, and thanks for all the work on this change @stanogn! This looks like it will be a very nice improvement!
> > I left a few questions, but overall I think we need to ensure we don't introduce any regressions with a change like this. What do you think about improving the test coverage? Or, would there be a way to break down this PR into several smaller changes that could be delivered over a few iterations? One thought that comes to mind would be to split out `fetch_certificates` as a separate code path while keeping `fetch_certificate` active during the refactoring. Just an idea.
> 
> Thanks for the review. I agree, it does feel a bit fragile considering its complexity. I dont think it would be easy to break it down into multiple iterations without adding even more complexity. I would rather add some more test cases.

I added some unit tests to demonstrate that it now pairs the certificates with private keys by inspecting their content and certificates don't need to have the ID in their filename and similiar tests. Basically it should demonstrate robustness. I will try to add some more tests related to the dev portal over the weekend but it takes quite some time 😅

PR #21187, opened 2023-04-06
Remove deprecated attributes from apps requests

giginet:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I see several green `ci/circleci` builds in the "All checks have passed" section of my PR ([connect CircleCI to GitHub](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-) if not)
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

closes #21125

This PR fixes this issue.
[Listing apps fails, due to the recent pricing update on App Store? · Issue #21125 · fastlane/fastlane](https://github.com/fastlane/fastlane/issues/21125)

### Description

Since App Store Connect API 2.3([released on March 2023](https://developer.apple.com/news/releases/?id=03092023b)), `App.prices` has been deprecated. 
https://developer.apple.com/documentation/appstoreconnectapi/app/relationships/prices

After this version, the following 409 error is returned when requested with including`prices` attributes.

```
The resource 'appPrices' cannot be viewed, created or updated. Please view and create 'manualPrices' using the resource 'appPriceSchedules'.
```

However, `prices` are always included as an essential attribute.

This PR removes `prices` from `ESSENTIAL_INCLUDES`.

### Testing Steps

1. Enable scheduled prices for your app on AppStore Connect
2. Execute the following lane

```ruby
lane :app_version do
  app_store_connect_api_key 
  app_store_build_number
end
```

### Discussion

This PR is just a hotfix. We need to support the new price API to fetch prices.

The current response mock is old. So we have to update this to follow the latest API spec.
https://github.com/fastlane/fastlane/blob/0c06ab0d8049b1c4012dc5305b190c7f5f825de1/spaceship/spec/connect_api/fixtures/testflight/apps.json#L113

PR #21213, opened 2023-04-18
Fix accessing first element of nil array in upload_price_tier

### LINKED ISSUE #21199: Fastlane getting "The request could not be completed because: Server error got 500" error for iOS Apps
Fastlane failing with the error message **_The request could not be completed because: Server error got 500_** but verbose shows it fails when uploading IPA file to App Store Connect.

Please what could be the issue and how do I go about solving this ?

I have been experiencing this for 3 days now.

lanpau:
### Checklist

### Motivation and Context

Resolves #21199

### Description

Error "undefined method `first' for nil:NilClass" indicates attempting to treat nil as an array. Adding pre-condition to check if the array is not nil fixes that. Also the startDate attribute of the resource appPrices in spaceship does not exist anymore and must be removed.

### Testing Steps

Calling the deliver action for apps with unspecified prices now works as expected.

### DIFF
--- a/deliver/lib/deliver/upload_price_tier.rb
+++ b/deliver/lib/deliver/upload_price_tier.rb
@@ -21,7 +21,7 @@ module Deliver
       # Need to get prices from the app's relationships
       # Prices from app's relationship doess not have price tier so need to fetch app price with price tier relationship
       app_prices = app.prices
-      if app_prices.first
+      if app_prices && app_prices.first
         app_price = Spaceship::ConnectAPI.get_app_price(app_price_id: app_prices.first.id, includes: "priceTier").first
         old_price = app_price.price_tier.id
       else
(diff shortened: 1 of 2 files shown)


deliver/lib/deliver/upload_price_tier.rb
```diff
@@ -21,7 +21,7 @@ def upload(options)
       # Need to get prices from the app's relationships
       # Prices from app's relationship doess not have price tier so need to fetch app price with price tier relationship
       app_prices = app.prices
-      if app_prices.first
+      if app_prices && app_prices.first
```
paskowski, line 24: Adding this additional check did fix this issue for me. I did not apply the change in `tunes.rb` though.

notlaein (approved):

deliver/lib/deliver/upload_price_tier.rb
vincentisambart, line 24: This can be shortened to the following:
```suggestion
      if app_prices&.first
```
However, I don't think it is the correct fix. `app.prices` being `nil` seems to be due to the change in #21187, but if that's indeed the case, `app_prices` will always be `nil`, so you end up always going in the `else` branch of this `if`, even if App Store Connect does have prices.

PR #21250, opened 2023-05-05
[register_devices] Allow hash with UDID key and duplicated device names

rasberik:
### Checklist

### Motivation and Context

`register_devices` currently takes a Hash with custom name as key and udid as value, and internally converts it to array.

This allows passing identical UDID several times under different names but forbids identical names with different UDID. 
It is quite opposite of what Apple Developer portal allows to do - registration of devices with identical name, as long as UDID's are unique.

Although it is clear that identical names should be avoided as much as possible for various reasons, its unexpected that `fastlane` enforces this custom rule (whilst Apple does not)

In my environment, automation is generating device name from combination of inputs, like device model, user id etc, but it is possible that eventually there will be identical names for different devices from same person or team. In order to avoid fastlane limitation, additional logic is needed, although I would like to proceed to registration with same name.

Vision is to allow device registration without adding unexpected complexities.

### Description

This PR introduces another optional variable `is_devices_inv`, in addition to `devices`. By using `is_devices_inv` boolean flag - inverted hash (key ↔︎ value) version of `devices` will be used, it will allow passing hash with only unique UDID and repetitive values.

IMO, Ideally, devices should be always [UDID:DeviceName] hash, but if we enforce it, all existing usages need to be updated in order to adopt the change. Boolean flag is not ideal and hopefully to be removed after (if) 100% adoption, but allows both cases at the same time, and gives teams time to adopt if needed.

### Testing Steps

Added tests to `register_devices_spec.rb` that verifies `registers devices from hash` with regular `devices` and `is_devices_inv` options. 
Previously, it was lacking `devices` tests, so I added few basic sets of tests.

```
bundle exec rspec fastlane/spec/actions_specs/register_devices_spec.rb
```


fastlane/lib/fastlane/actions/register_devices.rb
```diff
@@ -14,7 +14,9 @@ def self.file_column_headers
       def self.run(params)
         platform = Spaceship::ConnectAPI::BundleIdPlatform.map(params[:platform])
 
-        if params[:devices]
+        if params[:devices_inv]
+          new_devices = params[:devices_inv].to_a
```
rasberik, line 18: for `devices_inv`, simply converting to array as there is no need to invert key-value

fastlane/lib/fastlane/actions/register_devices.rb
```diff
@@ -104,6 +106,11 @@ def self.available_options
                                        description: "A hash of devices, with the name as key and the UDID as value",
                                        type: Hash,
                                        optional: true),
+          FastlaneCore::ConfigItem.new(key: :devices_inv,
+                                       env_name: "FL_REGISTER_DEVICES_DEVICES_INV",
+                                       description: "A hash of devices, with the UDID as key and the name as value",
+                                       type: Hash,
+                                       optional: true),
```
rasberik, line 113: [self-review] similar to devices, with updated description

fastlane/lib/fastlane/actions/register_devices.rb
```diff
@@ -184,12 +191,18 @@ def self.example_code
               "Luka iPhone 6" => "1234567890123456789012345678901234567890",
               "Felix iPad Air 2" => "abcdefghijklmnopqrstvuwxyzabcdefghijklmn"
             }
-          ) # Simply provide a list of devices as a Hash',
+          ) # Simply provide a list of devices as a Hash, with the name as key',
```
rasberik, line 194: [self-review] Explicitly added comment that name is key in case of using `devices`

fastlane/spec/actions_specs/register_devices_spec.rb
```diff
@@ -93,6 +93,58 @@
...
+          devices = {
+            "NAME1" => "A123456789012345678901234567890123456789",
+            "NAME2"	=> "B123456789012345678901234567890123456789"
+          }
+
+          expect(Fastlane::Actions::RegisterDevicesAction).to receive(:try_create_device).with(
+            name: 'NAME2',
+            udid: 'B123456789012345678901234567890123456789',
+            platform: "IOS"
+          )
```
rasberik, line 109: [self-review] only 1 request `try_create_device` is expected as `NAME1` already exists from `fake_devices` in line 98.

fastlane/spec/actions_specs/register_devices_spec.rb
```diff
@@ -93,6 +93,58 @@
...
+            end").runner.execute(:test)
+        end
+
+        it "with udid as a key" do
+          expect(Spaceship::ConnectAPI::Device).to receive(:all).and_return(fake_devices)
+
+          devices = {
+            "B123456789012345678901234567890123456789" => "NAME",
+            "C123456789012345678901234567890123456789" => "NAME"
+          }
```
rasberik, line 125: [self-review] UDID are unique (including existing devices from line 120), but names are same

fastlane/lib/fastlane/actions/register_devices.rb
rasberik: Outdated after updating to Bool

fastlane/lib/fastlane/actions/register_devices.rb
rasberik: Outdated after updating to Bool

rasberik:
@getaaron Sorry, I noticed your review on spaceship related PRs, may be you can have a look - I dont have access to request reviews 🙏

rasberik:
Probably cleaner approach would be to suggest non-optional breaking change, but thats harder to get merged so Ill close this

rasberik:
Those who need solution can use workaround:

```ruby
def reverse_udid_map(udid_map)
  reversed_map = []
  grouped_hash = udid_map.group_by(&:last).transform_values { |value| value.map(&:first) }
  grouped_hash&.each do |group|
    group[1].each_with_index do |element, index|
      if reversed_map[index].nil?
        reversed_map << {}
      end
      reversed_map[index][group[0]] = element
    end
  end
  reversed_map
end

# Use
def register_devices(udid_map)
  reversed_map = reverse_udid_map(udid_map)
  reversed_map.each_with_index do |group, index|
    register_devices(
      devices: group,
      team_id: "YOUR_TEAM_ID",
      api_key: app_store_connect_api_key,
    )
  end
end
```

PR #21394, opened 2023-07-13
Updraft support

Hhajdukovic93:
### Checklist
- [x ] I've updated the documentation if necessary.

### Motivation and Context

Resolves Idea #21392

### Description

Add new action updraft()

### DIFF
--- a/fastlane/lib/fastlane/actions/updraft.rb
+++ b/fastlane/lib/fastlane/actions/updraft.rb
@@ -1,0 +1,163 @@
+module Fastlane
+  module Actions
+    module SharedValues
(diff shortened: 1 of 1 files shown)


KevinHatf:
Thanks that you shared the list with us i think you i will get more idea now from here 

Will share my opinion with you soon. 
[ADP Workforce Now App](http://adpworkforce-now.com/)

iBotPeaches:
Sorry for delay. I think its best this stays as a plugin and not part of the core. We are trying to reduce 3rd party actions directly in the codebase.

PR #21457, opened 2023-08-12
[spaceship] improve performance of `app.get_latest_app_store_version` and `app.get_builds`

rogerluan:
### Checklist

### Motivation and Context

While implementing a service that uses _fastlane_ ([statused.com](https://statused.com)), I stumbled upon performance issues that could only be fixed by making some tweaks to _fastlane_ itself. Some apps with lots of builds and lots versions were taking up to 2 minutes to finish some requests (due to its behavior of eagerly fetching all pages), and after these changes the same task now takes to ~8 seconds 🎉 

### Description

Simple improvements, allowing users to limit the number of elements returned in these paginated APIs.

### Testing Steps

These were the changes I made to my service, you can use them to test these changes. I didn't update unit tests because these APIs have none 😬 

```diff
app.get_latest_app_store_version(
  client: client,
  platform: Spaceship::ConnectAPI::Platform::IOS,
  includes: "build",
+  limit: 1,
)
```

```diff
-latest_build_info = app.get_builds(client: client).sort_by(&:uploaded_date).last
+latest_build_info = app.get_builds(
+  client: client,
+  limit: 1,
+  sort: "-uploadedDate",
+).last # This array has just 1 build at this point
```


spaceship/lib/spaceship/connect_api/models/app.rb
```diff
@@ -293,13 +300,14 @@ def get_pending_release_app_store_version(client: nil, platform: nil, includes:
         return get_app_store_versions(client: client, filter: filter, includes: includes).first
       end
 
-      def get_app_store_versions(client: nil, filter: {}, includes: Spaceship::ConnectAPI::AppStoreVersion::ESSENTIAL_INCLUDES, limit: nil, sort: nil)
+      def get_app_store_versions(client: nil, filter: {}, includes: Spaceship::ConnectAPI::AppStoreVersion::ESSENTIAL_INCLUDES, limit: nil)
```
AliSoftware, line 303: 👍  A quick search in the codebase seem to confirm that this method was never called with a value for the `sort:` parameter before anyway, so removing it shouldn't be a breaking change to remove that parameter 👍

AliSoftware (commented):
Apart from the suggestion left below, it's looking good.

> I didn't update unit tests because these APIs have none 😬

I'd love if we used the occasion of this PR to add some unit tests though. Even if it's only a minimal set just for those methods you modified there for starters (as opposed to covering the whole class) if you don't want to spend too much time on them, but just so that we have something instead of nothing, and so that we get a starting point for next time we modify those models and enrich these tests little by little over time.

spaceship/lib/spaceship/connect_api/models/app.rb
```diff
@@ -357,8 +365,13 @@ def get_builds(client: nil, filter: {}, includes: nil, limit: nil, sort: nil)
...
 
-        resps = client.get_builds(filter: filter, includes: includes, limit: limit, sort: sort).all_pages
-        return resps.flat_map(&:to_models)
+        if limit.nil?
+          resps = client.get_builds(filter: filter, includes: includes, limit: limit, sort: sort).all_pages
+          return resps.flat_map(&:to_models)
+        else
+          resp = client.get_builds(filter: filter, includes: includes, limit: limit, sort: sort)
+          return resp.to_models
+        end
```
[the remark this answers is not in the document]
rogerluan: I didn't think about moving the `all_pages` below :) nice. I think the ternary is too convoluted though, would you mind if I used if/else in this case?

```suggestion
        resp = client.get_builds(filter: filter, includes: includes, limit: limit, sort: sort)
        if limit.nil?
          resp.all_pages.flat_map(&:to_models)
        else 
          resp.to_models
        end
```

spaceship/lib/spaceship/connect_api/models/app.rb
```diff
@@ -223,15 +223,22 @@ def ensure_version!(version_string, platform: nil, client: nil)
...
       end
 
-      def get_latest_app_store_version(client: nil, platform: nil, includes: nil)
+      # This will return the latest version that is available on the App Store.
+      # By default, when limit is nil, it will eagerly fetch all versions, sort them, and then return the last one.
+      # If you set a limit, it will only fetch the latest versions (which is much faster), and then return the last one.
+      # However, if Apple changes the way they sort versions, this might not return the latest versions, hence why
+      # the default behavior is to fetch all versions.
+      # Unfortunately, there's no way to sort by version number on the server side, so we have to do it on the client side.
+      # Pass in a limit knowing the risks of this changing silently in the future.
```
getaaron, line 232: IMO the risk that Apple will start returning App Store versions in a different order with `limit` provided is so infinitesimally small that I propose a default value of `limit = 1`

getaaron: If you want as a safeguard we could add an environment variable to override this behavior if it causes problems

spaceship/lib/spaceship/connect_api/models/app.rb
lacostej: Could that break external users though? `spaceship` is also a module.

We could do one thing.
1. Keep the parameter for a while
2. log with deprecated if the value isn't `nil`
3. delete the parameter in X months.

I haven't seen a proper strategy for managing deprecation / API stability yet.

spaceship/lib/spaceship/connect_api/models/app.rb
lacostej: I don't mind the default value being `limit=1`  as it doesn't change the behavior for today's users and bring the performance improvement to all users.

spaceship/lib/spaceship/connect_api/models/app.rb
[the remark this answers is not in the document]
lacostej: 👍 to this change

spaceship/lib/spaceship/connect_api/models/app.rb
```diff
@@ -301,14 +310,14 @@ def get_pending_release_app_store_version(client: nil, platform: nil, includes:
...
         client ||= Spaceship::ConnectAPI
+        resp = client.get_app_store_versions(app_id: id, filter: filter, includes: includes, limit: limit)
         if limit.nil?
-          resps = client.get_app_store_versions(app_id: id, filter: filter, includes: includes, limit: limit, sort: sort).all_pages
-          return resps.flat_map(&:to_models)
+          resp.all_pages.flat_map(&:to_models)
         else
-          resp = client.get_app_store_versions(app_id: id, filter: filter, includes: includes, limit: limit, sort: sort)
-          return resp.to_models
+          resp.to_models
```
AliSoftware, line 320: I just realized that this means that we don't ask subsequent pages of the response if `limit` is provided… but what if I pass `limit: 30` and the API returns a paginated response of 25 results per page? We'd still need to at least request the next page to get past the first 25 and up to 30 results, right?

spaceship/lib/spaceship/connect_api/models/app.rb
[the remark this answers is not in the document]
AliSoftware: Same note as above about the case of `limit` being provided but with a value > the number of results per page

iBotPeaches:
Have a vested interest in this now, so will spin up and investigate. I was wondering why our code was taking so long, but indeed if you set a limit lower - it just uses that limit for each pagination. It doesn't actually reduce anything - just does the opposite and increases the amount of calls - slowing it down further.

My usage was manually calling `get_builds` to check programmatically the status of a build. That will pull all records regardless if you just wanted 1. I believe this is all connected.

spaceship/lib/spaceship/connect_api/models/app.rb
iBotPeaches: I looked into this. This endpoint does not have a `per_page` parameter, so `limit` = `per_page`. So if you ask for 30 - you get 30. This is allowed up to 200. So not passing a `limit` will return up to 200, then you use `links.next` to iterate based on the cursor.

So I don't think you can hit a situation where you can have a different per page and limit. This improvement does solve the issue that when you asked for a limit of 5 - it just requested 5 (per page), then the `all_pages` would just read `links.next` and ask for another 5 - over n over n over.

AliSoftware: That still means that if for whatever reason you want the first 250 ones you can't just set `limit = 250` because it will only return the first 200. And we don't want `all_pages` in that case either.

So I think the only solution is to use `all_pages` when no `limit` is set, but if a `limit` is set we should `while`-loop manually using the `next` cursor to accumulate results until either the total number of results is `>= limit` or `next` is `nil` (last page)

iBotPeaches: ahh now I see your point. If count is greater than the influenced max of Apple.

AliSoftware: Yep.

I haven't looked in details (I'm on my phone …) but I think either Faraday or our Spaceship client might have helpers for handling paging (more than just the ‘all_pages` helper, I mean) that we can probably rely on instead of reinventing the wheel?

For example https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L320 (though that one seems to use page numbers not cursor… but maybe we have other helpers, or the libraries we use do? After all this seems like a pretty common need)

PR #21495, opened 2023-09-01
[docs] add minor branding guidelines to `CONTRIBUTING.md`

rogerluan:
### Checklist

### Motivation and Context

This is already enforced by our CI checks:

https://github.com/fastlane/fastlane/blob/ab9fb3249adaddf6c05655891b4d280415dfc84e/fastlane/Fastfile#L585-L607

But it's not documented anywhere (and I also I believe that CI check is not very thorough). This PR just formalizes those guidelines, and it's a starting point for us to improve/shape it better in the future (should there be any other additions).

I wasn't sure where to place this section. Idk if we'd rather have yet another markdown file, like `BRANDING.md` or something similar? 🤷‍♂️ 

### Testing Steps

Looks like this formatted:

image

You can preview it here: https://github.com/fastlane/fastlane/blob/rogerluan-branding/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,9 +68,9 @@ Help us keep _fastlane_ open and inclusive. Please read and follow our [Code of 
 
 ## Branding
 
-Although we don't have a formal style guide, we do have a few guidelines for how to refer to _fastlane_ and its tools:
+We have a few guidelines for how to refer to _fastlane_ and its tools:
 
-- _fastlane_ and all of the its actions should be written in all lowercase, even at the beginning of a sentence:
+- _fastlane_ and its actions should be written in all lowercase, even at the beginning of a sentence:
     - ❌ "Use Fastlane to automate your screenshots."
     - ❌ "Use _Fastlane_ to automate your screenshots."
     - ❌ "Fastlane helps you deliver faster."
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,6 +72,7 @@ Although we don't have a formal style guide, we do have a few guidelines for how
 
 - _fastlane_ and all of the its actions should be written in all lowercase, even at the beginning of a sentence:
     - ❌ "Use Fastlane to automate your screenshots."
+    - ❌ "Use _Fastlane_ to automate your screenshots."
     - ❌ "Fastlane helps you deliver faster."
     - ❌ "Match makes code signing management easy."
     - ✅ "Use _fastlane_ to automate your screenshots."
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,6 +66,28 @@ One of the best ways we can keep _fastlane_ an approachable, stable, and dependa
 
 Help us keep _fastlane_ open and inclusive. Please read and follow our [Code of Conduct][code of conduct].
 
+## Branding
+
+Although we don't have a formal style guide, we do have a few guidelines for how to refer to _fastlane_ and its tools:
+
+- _fastlane_ and all of the its actions should be written in all lowercase, even at the beginning of a sentence:
+    - ❌ "Use Fastlane to automate your screenshots."
+    - ❌ "Fastlane helps you deliver faster."
+    - ❌ "Match makes code signing management easy."
+    - ✅ "Use _fastlane_ to automate your screenshots."
+    - ✅ "_fastlane_ helps you deliver faster."
+    - ✅ "_match_ makes code signing management easy."
+- _fastlane_ and all of its actions should be italicized when written in prose:
+    - ❌ "`fastlane` is an all-in-one tool for app automation."
+    - ❌ "**fastlane** is an all-in-one tool for app automation."
+    - ❌ "fastlane is an all-in-one tool for app automation."
+    - ❌ "<ins>fastlane</ins> is an all-in-one tool for app automation."
+    - ❌ "`match` makes code signing management easy."
+    - ✅ "_fastlane_ is an all-in-one tool for app automation."
+    - ✅ "_match_ makes code signing management easy."
+
+Please use these guidelines when writing about _fastlane_ and its tools, be it when contributing to the project or writing about it elsewhere.
+
 ## Above All, Thanks for Your Contributions
 
 Thank you for reading to the end, and for taking the time to contribute to the project! If you include the 🔑 emoji at the top of the body of your issue or pull request, we'll know that you've given this your full attention and are doing your best to help!
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -584,7 +584,7 @@ end
 
 desc "Ensure the correct formatting for the fastlane tools"
 private_lane :ensure_tool_name_formatting do
-  test_tool_name_formatting_ensurance_lane
+  require_relative "../fastlane/helper/tool_name_formatting_helper.rb"
   UI.message("🕗  Verifying tool name formatting...")
   require 'fastlane/tools'
   errors = []
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -591,50 +591,13 @@ private_lane :ensure_tool_name_formatting do
   Dir.chdir("..") do
     Dir["**/*.md"].each do |path|
       content = File.read(path)
-      errors += find_tool_name_formatting_errors(content: content, path: path)
+      helper = Helper::ToolNameFormattingHelper.new(content: content, path: path)
+      errors += helper.find_tool_name_formatting_errors
     end
   end
   errors.each { |a| UI.error(a) }
   UI.user_error!("Invalid formatting of one of the fastlane tools") unless errors.empty?
   UI.success("✅  fastlane tools formatting is correct")
-end
-
-private_lane :test_tool_name_formatting_ensurance_lane do
-  content = File.read("spec/fixtures/fastfiles/tool_name_formatting.txt")
-  errors = find_tool_name_formatting_errors(content: content, path: "CONTRIBUTING.md")
-  errors.select! { |e| !e.end_with?("❌") }
-  errors.each do |e|
-    # Lines that end with "❌" are expected to be errors
-    UI.error(e)
-  end
-  UI.user_error!("Tool name formatting tool failed its unit tests") unless errors.empty?
-  UI.success("✅  Tool name formatting tool passed its unit tests")
-end
-
-def find_tool_name_formatting_errors(content:, path:)
-  errors = []
-  Fastlane::TOOLS.each do |tool|
-    lines = content.split("\n")
-    lines.each_with_index do |line, index|
-      line_number = index + 1
-      # Ignore checks if line starts with "^\s*- ❌" (i.e. if it's a line that's already been marked as incorrect)
-      # and if the file is CONTRIBUTING.md (i.e. if it's the branding guidelines section)
-      next if line =~ /^\s*- ❌/ && path == "CONTRIBUTING.md"
-      errors << "Use _#{tool}_ instead of `#{tool}` to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("`#{tool}`")
-      errors << "Use _#{tool}_ instead of `_#{tool}_` to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("`_#{tool}_`")
-      errors << "Use [_#{tool}_] instead of [#{tool}] to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("[#{tool}]")
-      errors << "Use _#{tool}_ instead of **#{tool}** to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("**#{tool}**")
-      errors << "Use <em>#{tool}<em> instead of <code>#{tool}</code> to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("<code>#{tool}</code>")
-      errors << "Use <em>#{tool}<em> or _#{tool}_ instead of <ins>#{tool}</ins> to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("<ins>#{tool}</ins>")
-
-      line.scan(/([A-Z_]*)([_ `"])(#{Regexp.escape(tool.to_s)})\2([A-Z_]*)/i) do |prefix, delimiter, tool_name, suffix|
-        wrong_case = tool_name != tool.to_s.downcase
-        looks_like_an_env_var = (tool_name == tool_name.upcase) && delimiter == '_' && (!prefix.empty? || !suffix.empty?)
-        errors << "fastlane tools have to be formatted in lowercase: #{tool} in '#{path}:#{line_number}': #{line}" if !looks_like_an_env_var && wrong_case
-      end
-    end
-  end
-  errors
 end
 
 # Run the code samples, yo
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -590,8 +590,7 @@ private_lane :ensure_tool_name_formatting do
   errors = []
   Dir.chdir("..") do
     Dir["**/*.md"].each do |path|
-      content = File.read(path)
-      helper = Helper::ToolNameFormattingHelper.new(content: content, path: path)
+      helper = Helper::ToolNameFormattingHelper.new(path: path, is_documenting_invalid_examples: path == 'CONTRIBUTING.md')
       errors += helper.find_tool_name_formatting_errors
     end
   end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -629,7 +629,7 @@ def find_tool_name_formatting_errors(content:, path:)
 
       line.scan(/([A-Z_]*)([_ `"])(#{Regexp.escape(tool.to_s)})\2([A-Z_]*)/i) do |prefix, delimiter, tool_name, suffix|
         wrong_case = tool_name != tool.to_s.downcase
-        looks_like_an_env_var = (tool_name == tool_name.upcase) && delimiter == '_' && (!prefix.empty? || !suffix.emtpy?)
+        looks_like_an_env_var = (tool_name == tool_name.upcase) && delimiter == '_' && (!prefix.empty? || !suffix.empty?)
         errors << "fastlane tools have to be formatted in lowercase: #{tool} in '#{path}:#{line_number}': #{line}" if !looks_like_an_env_var && wrong_case
       end
     end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -627,9 +627,9 @@ def find_tool_name_formatting_errors(content:, path:)
       errors << "Use <em>#{tool}<em> instead of <code>#{tool}</code> to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("<code>#{tool}</code>")
       errors << "Use <em>#{tool}<em> or _#{tool}_ instead of <ins>#{tool}</ins> to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("<ins>#{tool}</ins>")
 
-      line.scan(/([A-Z_]*)([_ `"])(#{Regexp.escape(tool.to_s)})\2([A-Z_]*)/i) do |prefix, delim, tool_name, suffix|
+      line.scan(/([A-Z_]*)([_ `"])(#{Regexp.escape(tool.to_s)})\2([A-Z_]*)/i) do |prefix, delimiter, tool_name, suffix|
         wrong_case = tool_name != tool.to_s.downcase
-        looks_like_an_env_var = (tool_name == tool_name.upcase) && delim == '_' && (!prefix.empty? || !suffix.emtpy?)
+        looks_like_an_env_var = (tool_name == tool_name.upcase) && delimiter == '_' && (!prefix.empty? || !suffix.emtpy?)
         errors << "fastlane tools have to be formatted in lowercase: #{tool} in '#{path}:#{line_number}': #{line}" if !looks_like_an_env_var && wrong_case
       end
     end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -603,8 +603,13 @@ private_lane :ensure_tool_name_formatting do
           errors << "Use _#{tool}_ instead of **#{tool}** to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("**#{tool}**")
           errors << "Use <em>#{tool}<em> instead of <code>#{tool}</code> to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("<code>#{tool}</code>")
           errors << "Use <em>#{tool}<em> or _#{tool}_ instead of <ins>#{tool}</ins> to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("<ins>#{tool}</ins>")
-          if line.include?("_#{tool.to_s.capitalize}_") || line.include?("`#{tool.to_s.capitalize}`") || line.include?(" #{tool.to_s.capitalize} ")
-            errors << "fastlane tools have to be formatted in lowercase: #{tool} in '#{path}:#{line_number}'"
+
+          # Check if the tool is mentioned in the line, but not in lowercase
+          line.scan(/([_ `"])(#{Regexp.escape(tool.to_s)})\1/i) do |match|
+            matched_tool = match[1]
+            tool_is_written_wrong_case = matched_tool != matched_tool.downcase
+            looks_like_an_env_var = line.match?(/[A-Z_]+_#{matched_tool}_[A-Z_]+/)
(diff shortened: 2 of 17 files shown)


rogerluan:
Heh, CI checks are working!

image

Looks like we're gonna need to add an exception to the CI checks, or force merge this PR 👀 any preference? If implementing the exception, it'd probably be an exception on the entire file... Should we move these guidelines to a `BRANDING.md` or something? 🤷

AliSoftware:
I like that it's in `CONTRIBUTING.md` because:
 - `CONTRIBUTING.md` is a standard in the OSS community, but I don't think `BRANDING.md` is (at least I haven't seen any OSS repo with such a file)?
 - Which also means that it's less likely that people are gonna miss it, because people are more likely to read `CONTRIBUTING.md` but less likely to think about reading `BRANDING.md` which is not a common one
 - Finally, I think nowadays GitHub has some integration with `CONTRIBUTING.md`, like have the UI automatically suggest new contributors to read the file before creating their first pull request, or detect if an already existing contributor returns back to the project and open a PR after the last time the `CONTRIBUTING.md` file was changed (i.e. if that's their first PR since the guidelines changes)

So for all those reasons, I'd suggest to keep it in `CONTRIBUTING.md` like you did there.

---

I wonder if we might be able to be clever about the CI check to exclude just that section from the markdown when checking, instead of having to disable that the check of the whole `CONTRIBUTING.md` file. Or maybe detect if the file is `CONTRIBUTING.md` AND the line being checked starts with `- ❌ `, then ignore it, something like that?

rogerluan:
Fully agree with all your points @AliSoftware 😊 

I just pushed some changes lmk if those are good enough 😃 

Before and after enabling the "skip" for the branding guidelines:

image

fastlane/Fastfile
```diff
@@ -591,12 +591,21 @@ private_lane :ensure_tool_name_formatting do
...
+          next if line =~ /^\s*- ❌/ && path == "CONTRIBUTING.md"
+          errors << "Use _#{tool}_ instead of `#{tool}` to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("`#{tool}`")
+          errors << "Use _#{tool}_ instead of `_#{tool}_` to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("`_#{tool}_`")
+          errors << "Use [_#{tool}_] instead of [#{tool}] to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("[#{tool}]")
+          errors << "Use _#{tool}_ instead of **#{tool}** to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("**#{tool}**")
+          errors << "Use <em>#{tool}<em> instead of <code>#{tool}</code> to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("<code>#{tool}</code>")
+          errors << "Use <em>#{tool}<em> or _#{tool}_ instead of <ins>#{tool}</ins> to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("<ins>#{tool}</ins>")
+          if line.include?("_#{tool.to_s.capitalize}_") || line.include?("`#{tool.to_s.capitalize}`") || line.include?(" #{tool.to_s.capitalize} ")
```
AliSoftware, line 606: Note that this test only checks if the tool name had its first letter capitalized (e.g. `Fastlane`) but not any other case variation. So it wouldn't catch typos like `FastLane`.

It might be better to use `line.scan(/\b#{tool}\b/i)` to capture each occurrences of the tool name used as a word (`\b` = "word boundary" marker in Regexp) in a case-insensitive way, then for each occurrence found by `scan`, check if it's == to `to_s.downcase` or not.

Something like (not tested):
```ruby
line.scan(/\b#{Regexp.escape(tool.to_s)}\b./) do |match|
  errors << "fastlane tools have to be formatted in lowercase: #{tool} in '#{path}:#{line_number}'" unless match == match.downcase
end

File: fastlane/lib/fastlane/actions/docs/capture_android_screenshots.md [hunk 2]
```diff
@@ -32,7 +32,7 @@ androidTestImplementation 'tools.fastlane:screengrab:x.x.x'
 
 The latest version is [ ![Download](https://maven-badges.herokuapp.com/maven-central/tools.fastlane/screengrab/badge.svg)](https://search.maven.org/artifact/tools.fastlane/screengrab)
 
-As of Screengrab version 2.0.0, all Android test dependencies are AndroidX dependencies. This means a device with API 18+, Android 4.3 or greater is required. If you wish to capture screenshots with an older Android OS, then you must use a 1.x.x version.
+As of _screengrab_ version 2.0.0, all Android test dependencies are AndroidX dependencies. This means a device with API 18+, Android 4.3 or greater is required. If you wish to capture screenshots with an older Android OS, then you must use a 1.x.x version.
```
AliSoftware, line 35: Interesting that those were not caught as error with the previous implementation 🤔

fastlane/Fastfile
rogerluan: I tried using `\b` before but it doesn't work, it catches more false positives, e.g. `app = Spaceship::ConnectAPI.get_app(app_id: 1013943394).first` <-- "Spaceship" is caught here. Too troublesome to workaround those 🤷  that's why I kept those silly (yet simple) delimiters hehe

I just tried your suggested approach using delimiters and there are a few false positives:

https://github.com/fastlane/fastlane/blob/ab9fb3249adaddf6c05655891b4d280415dfc84e/RespondingToIssuesAndPullRequests.md?plain=1#L66

https://github.com/fastlane/fastlane/blob/ab9fb3249adaddf6c05655891b4d280415dfc84e/Testing.md?plain=1#L86

https://github.com/fastlane/fastlane/blob/ab9fb3249adaddf6c05655891b4d280415dfc84e/Testing.md?plain=1#L91

What do you think we should do? Renaming those env vars to temporarily get rid of the false positive could be an option, but then no one in the future would be able to name an env var that has "FASTLANE" in it 😬 ?

rogerluan: The algo ended up like this btw:

```ruby
# Check if the tool is mentioned in the line, but not in lowercase
delimiters = ["_", " ", "`", "\""]
delimiters.each do |delimiter|
  line.scan(/#{delimiter}#{Regexp.escape(tool.to_s)}#{delimiter}/i) do |match|
    errors << "fastlane tools have to be formatted in lowercase: #{tool} in '#{path}:#{line_number}'" unless match == match.downcase
  end
end
```

rogerluan: I could also exclude full caps `FASTLANE` somehow lol WDYT @AliSoftware?

AliSoftware: I don't think we should exclude all-capts `FASTLANE`.

For your code above, you could have gone with the list of delimiters directly in the `Regexp`, using:
```
line.scan(/([_ `"])#{Regexp.escape(tool.to_s)}\1/i) do |match|`
```
and thus completely avoid doing the loop on `delimiter`

rogerluan: 😲 Great suggestion! I'll apply that change :) 

How do you suggest we solve those 3 false positives? 👀

AliSoftware: > How do you suggest we solve those 3 false positives?

We can allow `[A-Z_]+FASTLANE` and `FASTLANE_[A-Z_]+` as exceptions then, to allow it only if the name is part of a "upper-snake-case" word (like for an env var name)

rogerluan: image


Hmm I came up with a different regex than yours, lemme check yours (as it's simpler)

rogerluan: Yours works and it's simpler, check out the diff:

```
looks_like_an_env_var = line.match?(/[\`"][\w<_"]+#{matched_tool}[\w>_"]+[\`"]/)
looks_like_an_env_var = line.match?(/[A-Z_]+_#{matched_tool}_[A-Z_]+/)
```

Mine was checking if it was wrapped around backticks, double quotes, etc. That's probably not needed 😛 plus mine was allowing lowercase letters, we don't want those either...

rogerluan: Ready for review again @AliSoftware 😊

AliSoftware (changes requested):

CONTRIBUTING.md
```diff
@@ -66,6 +66,29 @@ One of the best ways we can keep _fastlane_ an approachable, stable, and dependa
 
 Help us keep _fastlane_ open and inclusive. Please read and follow our [Code of Conduct][code of conduct].
 
+## Branding
+
+Although we don't have a formal style guide, we do have a few guidelines for how to refer to _fastlane_ and its tools:
```
getaaron, line 71: IMO this sentence is a little wordy and doesn't add much to mention lack of a style guide

```suggestion
We have a few guidelines for how to refer to _fastlane_ and its tools:
```

CONTRIBUTING.md
```diff
@@ -66,6 +66,29 @@ One of the best ways we can keep _fastlane_ an approachable, stable, and dependa
...
+
+- _fastlane_ and all of the its actions should be written in all lowercase, even at the beginning of a sentence:
```
getaaron, line 73:
```suggestion
- _fastlane_ and all of its actions should be written in all lowercase, even at the beginning of a sentence:
```

getaaron (approved):
a few lil suggestions, take em or leave em. Looks great and I love the tests! 😍

CONTRIBUTING.md
getaaron: could also just be "_fastlane_ and its actions"

rogerluan:
```suggestion
- _fastlane_ and its actions should be written in all lowercase, even at the beginning of a sentence:
```

fastlane/Fastfile
```diff
@@ -591,12 +591,26 @@ private_lane :ensure_tool_name_formatting do
...
+          errors << "Use [_#{tool}_] instead of [#{tool}] to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("[#{tool}]")
+          errors << "Use _#{tool}_ instead of **#{tool}** to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("**#{tool}**")
+          errors << "Use <em>#{tool}<em> instead of <code>#{tool}</code> to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("<code>#{tool}</code>")
+          errors << "Use <em>#{tool}<em> or _#{tool}_ instead of <ins>#{tool}</ins> to mention a tool in the docs in '#{path}:#{line_number}'" if line.include?("<ins>#{tool}</ins>")
+
+          # Check if the tool is mentioned in the line, but not in lowercase
+          line.scan(/([_ `"])(#{Regexp.escape(tool.to_s)})\1/i) do |match|
+            matched_tool = match[1]
+            tool_is_written_wrong_case = matched_tool != matched_tool.downcase
+            looks_like_an_env_var = line.match?(/[A-Z_]+_#{matched_tool}_[A-Z_]+/)
```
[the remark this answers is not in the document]
rogerluan: Amazing work dude! amaze 

Perfect solution, I didn't even know we could use that syntax with the `scan` block (multiple objects, one for each match) 😲 TIL!

rogerluan:
@getaaron your copy suggestions were much better, thanks! Applied both 😊 
@AliSoftware I applied your algo changes and also added some "real" unit tests. I wanted to use rspec and all but I couldn't find a practical way to do it? So I just added a fixture and a basic method in the Fastfile itself 🤷‍♂️ this way we can ensure future iterations won't break the hard work we (you!) did in this PR 😂 lmk if you think there's a better place to place these tests :)

fastlane/Fastfile
```diff
@@ -584,28 +584,59 @@ end
...
+      errors << "Use _#{tool}_ instead of `_#{tool}_` to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("`_#{tool}_`")
+      errors << "Use [_#{tool}_] instead of [#{tool}] to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("[#{tool}]")
+      errors << "Use _#{tool}_ instead of **#{tool}** to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("**#{tool}**")
+      errors << "Use <em>#{tool}<em> instead of <code>#{tool}</code> to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("<code>#{tool}</code>")
+      errors << "Use <em>#{tool}<em> or _#{tool}_ instead of <ins>#{tool}</ins> to mention a tool in the docs in '#{path}:#{line_number}': #{line}" if line.include?("<ins>#{tool}</ins>")
+
+      line.scan(/([A-Z_]*)([_ `"])(#{Regexp.escape(tool.to_s)})\2([A-Z_]*)/i) do |prefix, delim, tool_name, suffix|
+        wrong_case = tool_name != tool.to_s.downcase
+        looks_like_an_env_var = (tool_name == tool_name.upcase) && delim == '_' && (!prefix.empty? || !suffix.emtpy?)
```
AliSoftware, line 632: Typo (my bad)
```suggestion
        looks_like_an_env_var = (tool_name == tool_name.upcase) && delim == '_' && (!prefix.empty? || !suffix.empty?)
```

AliSoftware: The fact that your pseudo-unit-test didn't crash due to this typo means that this case wasn't covered by your fixture btw.

I'd suggest some red-green-refactoring of the fixture (ie first change the fixture to include a case of an env var which _starts_ with a tool name, then run the test and confirm it goes red (well… or crashes, in this case), and _only then_ fix the typo in the code and confirm the test goes green); that way you'll be sure to have covered the use case properly.

AliSoftware (commented):
I'm not a fan of having the test for the lane be implemented that way instead of a proper rspec, but I'm also not sure if there's a better solution, short of implementing the lane as a method in the _fastlane_ gem's `lib/` code itself in order to have a spec for it (and then have the lane in the `Fastfile` call that method)… 🤔

fastlane/Fastfile
rogerluan: Great catch 👀 
I did the red-green-refactoring strategy and fixed it 💪

rogerluan:
>short of implementing the lane as a method in the fastlane gem's lib/ code itself in order to have a spec for it (and then have the lane in the Fastfile call that method)

I thought about this before but it's not ideal since it shouldn't be exposed publicly to fastlane users, it's just an internal tool 😞 do you know of any way to do it like this but not expose it to the lib? 🤔

AliSoftware:
> I thought about this before but it's not ideal since it shouldn't be exposed publicly to fastlane users, it's just an internal tool 😞 do you know of any way to do it like this but not expose it to the lib? 🤔

What about extracting the `find_tool_name_formatting_errors` method into a dedicated `.rb` file under `helper/`, and then then write spec against that? A bit like how it's done for `fastlane/helper/plugin_scores_helper.rb` and `fastlane/spec/helper/plugin_scores_helper_spec.rb`?
And then `require_relative` that file from the `Fastfile` to call the `find_tool_name_formatting_errors` method from the lane?

fastlane/Fastfile
```diff
@@ -584,28 +584,59 @@ end
...
   end
   errors.each { |a| UI.error(a) }
   UI.user_error!("Invalid formatting of one of the fastlane tools") unless errors.empty?
   UI.success("✅  fastlane tools formatting is correct")
 end
 
+private_lane :test_tool_name_formatting_ensurance_lane do
+  content = File.read("spec/fixtures/fastfiles/tool_name_formatting.txt")
+  errors = find_tool_name_formatting_errors(content: content, path: "CONTRIBUTING.md")
+  errors.select! { |e| !e.end_with?("❌") }
```
AliSoftware, line 605: The issue with this is that you're only checking if lines not ending with ❌  are _not_ detected at errors (actually, talk about double-negation… had it wrong the first time I read that code and thought it was doing the opposite at first…), but you're not checking if _all_ the lines ending with ✅  are detected as errors.

If you rewrite this as a spec (as suggested in my top-level comment from a couple of minutes ago), you should `expect(errors).to equal(the_exact_expected_list)`, as opposed to do it similar to what you did it, only expecting that all `errors` end with ✅  but without checking that you didn't miss one of them in the original fixture.

rogerluan: Your suggestion for helpers (similar to plugins) makes sense! It's literally the only file we have in the helper folder, nice 🤓 so this can definitely be moved to rspecs and proper unit tests be written, but can you also help me understand what you meant in this message you sent above? I think the tests are covering all lines, whether they're successful or not 🤔 We do check if the ones with ✅ succeed (and they do), and we also check if the ones with ❌ fail (and they do). What's missing? (just trying to understand here, if we were to keep it, although we won't 😇 )

rogerluan:
Lemme know if what I implemented is what you had in mind @AliSoftware 😃 alternatively we could write lots of `describe / it` pairs, it'd provide more information about the test cases (why it should succeed, why it should fail, with more developer comments), but at the same time it sounds like over engineering haha and it would make adding more cases in the future somewhat cumbersome. lmk if you think this is ok as is :)

fastlane/Fastfile
AliSoftware: Well another way of seeing this is that there are 4 categories of lines:
1. Lines that should be considered errors, and are properly flagged by `find_tool_name_formatting_errors` as such
2. Lines that should be valid, and are properly _not_ flagged by `find_tool_name_formatting_errors` as errors
3. Lines that should be considered errors, but are not properly detected and returned by `find_tool_name_formatting_errors`
4. Lines that should be valid, but are incorrectly flagged as errors by `find_tool_name_formatting_errors`

What we want is to ensure not only that cases 1 and 2 happen (i.e. our implementation work when it's supposed to) but also that cases 3 and 4 don't happen (i.e. that our implementation doesn't create neither false positives or false negatives)

In your code above, you only look at if the return value of `find_tool_name_formatting_errors` (aka `errors`), so you're only considering cases 1 and 4 (i.e. things that have been reported as `errors`). And especially then you're raising a `UI.error` if case 4 (the thing that was raised as errors shouldn't have been) happens.

But you're not checking case 3 at all, i.e. lines that should be considered as errors and should have been returned in `errors` when you called `find_tool_name_formatting_errors`, but weren't reported (in case our implementation of `find_tool_name_formatting_errors` would be incorrect — which is what we want to test here after all)

fastlane/Fastfile
```diff
@@ -584,21 +584,15 @@ end
 
 desc "Ensure the correct formatting for the fastlane tools"
 private_lane :ensure_tool_name_formatting do
+  test_tool_name_formatting_ensurance_lane
```
AliSoftware, line 587: Now that we have a spec, this should be removed.

fastlane/spec/helper/tool_name_formatting_helper_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    @helper = Fastlane::Helper::ToolNameFormattingHelper.new(content: content, path: "CONTRIBUTING.md")
+  end
+
+  describe "when parsing fixture file" do
+    it "should match array of expected errors" do
+      expected_errors = [
+        "fastlane tools have to be formatted in lowercase: fastlane in 'CONTRIBUTING.md:19': _FastLane_ makes code signing management easy. ❌",
+        "fastlane tools have to be formatted in lowercase: fastlane in 'CONTRIBUTING.md:23': A sentence that contains a _FastLane_ keyword, but also an env var: `FASTLANE_SKIP_CHANGELOG`. ❌"
+      ]
+      expect(@helper.find_tool_name_formatting_errors).to match_array(expected_errors)
```
AliSoftware, line 15: Yep exactly what I had in mind. Because this not only ensures that what we expect to be reported as errors are indeed reported as errors, but also checks that what we don't expect to be reported as errors and expect to be valid exceptions… isn't accidentally reported as errors either.

Btw, now that we have this implemented as a `spec`, I think it's worth removing the ❌  and ✅  at the end of each lines of your fixtures now, as they are not a needed hack anymore, and could accidentally mislead the thing being tested (e.g. imagine if our implementation of `find_tool_name_formatting_errors` didn't check if the lines of `CONTRIBUTING.md` _started_  with "- ❌", but instead our implementation tested if `include?("❌")`… we don't want those emojis at end of line to mud the waters here.

fastlane/Fastfile
rogerluan: Good point! 🎯

fastlane/Fastfile
rogerluan: Oops, forgot to commit that part :) it's also missing the require

fastlane/spec/helper/tool_name_formatting_helper_spec.rb
rogerluan: I thought about removing them but ended up leaving them (hence why I reworded the header of the fixture file), because I think it's still useful when we need to update this fixture in the future. I think the header explains all that so there's no "accidentally misleading" I think 🤔 unless devs don't read it. But you have a good point nonetheless, I'll remove them

rogerluan:
Interesting, AppVeyor is the only env where this is crashing:

image

All others, which run on linux and macOS, in diff Ruby versions, they're all passing 🤔 

Not sure if I should add `force_encoding("UTF-8")` 🤔 weird issue to begin with

AliSoftware:
I think I you replace `content = File.read(path)` with `content = File.read(path, mode: 'rb:BOM|UTF-8')` that should solve the issue.

That being said, that made me realize that it would be better in terms of memory-footprint to use `File.readlines`, which reads the file one by one, rather that load the entire file in memory via `File.read` only to `split("\n")` and iterate on each line.

So I think it would be better to:
 - Only pass the `path` in your `initialize` but not the `content` anymore
 - Switch your loops between the one that loops on tools and the only looping on lines
 - Replace the loop on lines with `File.read lines(path, mode: 'rb:BOM|UTF-8').each |line|`

fastlane/helper/tool_name_formatting_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+      def initialize(content:, path:)
+        @content = content
+        @path = path
+      end
+
+      def find_tool_name_formatting_errors
+        errors = []
+        Fastlane::TOOLS.each do |tool|
+          lines = content.split("\n")
+          lines.each_with_index do |line, index|
```
AliSoftware, line 15: Switching your double loop to this:
```suggestion
        File.readlines(@path, mode: 'rb:BOM|UTF-8').each_with_index do |line, index|
          Fastlane::TOOLS.each do |tool|
```
Then stop passing/storing `content` to the helper (nor reading the file's content at call site) should improve the memory footprint of the algorithm, and solve your encoding problem.

fastlane/spec/fixtures/fastfiles/tool_name_formatting.txt
```diff
@@ -0,0 +1,27 @@
...
+_match_ makes code signing management easy.
+- ❌ "`fastlane` is an all-in-one tool for app automation."
+- ❌ "**fastlane** is an all-in-one tool for app automation."
+- ❌ "fastlane is an all-in-one tool for app automation."
+- ❌ "<ins>fastlane</ins> is an all-in-one tool for app automation."
+- ❌ "`match` makes code signing management easy."
+_fastlane_ is an all-in-one tool for app automation.
+_match_ makes code signing management easy.
+FastLane makes code signing management easy. Unfortunately we can't catch this brand misalignment because there's no delimiter around the keyword.
+Fastlane makes code signing management easy. Unfortunately we can't catch this brand misalignment because there's no delimiter around the keyword.
```
AliSoftware, line 18: Wait is this still true? Maybe we could adjust the Regexp to catch this case of "left delimiter = start of line and right delimiter = space" and vice-versa?

Like instead of using `\1` in the Regexp to force delimiters on both sides to be the same, catch them separately then check the values (ie "either are the same or one is space and other is beginning/end of line") in the body of the `scan` block?

Or maybe more simply, run `scan` not on the `line`, but on `" #{line} "` instead?

AliSoftware:
@rogerluan I took the liberty to push 2 commits to implement my suggestions above (using `File.readlines` + using `mode: 'rb:BOM|UTF-8'` to indicate the encoding + running `scan` on `" #{line} "` to be able to detect formatting errors at start of line). Let me know what you think.

AliSoftware (approved):

fastlane/spec/fixtures/fastfiles/tool_name_formatting.txt
AliSoftware: Fixed in 2079206

rogerluan: 😲 great solution! 👏

rogerluan:
I don't mind at all, thanks for applying these @AliSoftware ! Great catch with inverting the loops, it's definitely more efficient this way (specially wrt the `readlines` method). Also your 2nd idea (wrapping the line around spaces) was quite neat! 👏 

I think we can move forward with this PR now! 🤩

rogerluan:
Also thanks for pushing the boundaries here 😉 we're in a much better place now, now just with some formalized branding guidelines, but better tools, test coverage, and we even made our docs more consistent 🚀

AliSoftware:
Teamwork at its best 🙂 

I'll let you the honour of merging the PR if you're all good with it now 😃

rogerluan (commented):
Reviewed again 👍 LGTM!

PR #21498, opened 2023-09-05
[supply] introduce a new synchronization logic for screenshots

AliSoftware:
### Checklist

- [x] ~I've updated the documentation if necessary.~

### Motivation and Context

Similar to the optimization we do with `deliver` and `sync_screenshots: true` for ASC, this change intents to allow optimizing the upload of images and screenshots to avoid re-uploading screenshots that already exist in Play Store Console.

This is especially useful if you call `upload_to_play_store` with `skip_upload_images: false` and/or `skip_upload_screenshots: false` every time you release a new version… but in practice don't change your app screenshots or `featureGraphic` / `icon` images on every release.

In addition to optimizing the screenshots upload and the number of upload requests done to the Google API, this also avoids having a long list of those screenshot updates showing up in the "Publishing Overview" of the PlayStore Console on every release when we didn't change the screenshot images since last time. While up until now, since we did a delete-and-reupload-all-screenshots on every call to `upload_to_play_store`, the PlayStore Console showed them as changes in "Publishing Overview > Changes Ready to Publish" even if the images were the same before/after.

### Description

The new algorithm replaces the call to `clear_screenshots`—which was previously done unconditionally before uploading all the screenshots again—with a _selective_ call to `clear_screenshot(…, image.id)` for images that are present in PSC but not present locally anymore (based on their sha256), or for images that are out of order. All while skipping upload for the ones that already present remotely and still in the right order, then finally only uploading the ones that are new locally, or re-uploading the ones that were out of order at the end.

 - Added a `sync_image_upload` parameter to `Supply::Options` to enable this new feature. Defaults to `false` to keep backwards compatibility.
 - Added an `ImageListing` model to represent the data from [the `Image` type returned by the Google API](https://googleapis.dev/ruby/google-apis-androidpublisher_v3/latest/Google/Apis/AndroidpublisherV3/Image.html).
 - Made `Supply::Client#fetch_images` to now return an `Array<ImageListing>` instead of an `Array<String>` of URLs, so that we also have access to the `sha256` and `id` fields of the existing images in PSC
    - Took the occasion to add some YARD doc, and to remove dead code from the `fetch_images` implementation
    - Update the call sites that were using this method, namely `Supply::Setup#download_images`, to call `map(&:url)` on the result, to keep working as before
 - Added `Supply::Client#clear_screenshot(image_type:, language:, image_id:)` to delete single screenshots (as opposed to `clear_screenshots` which deletes them all at once)

### Testing Steps

 - I've added a couple of unit tests in `supply/spec/uploader_spec.rb` to test tricky cases with `sync_image_upload: true`
 - I've also tested this with our own Android project, calling `upload_to_play_store` with `skip_upload_images: false, skip_upload_screenshots: false, sync_image_upload: true, validate_only: true` on the following situations, with our original screenshots being `[A,B,C,D,E]`:
    - Without making any changes to our screenshots ➡️ all uploads were skipped.
    - After removing screenshots `B` locally ➡️  that screenshot got deleted in the edit; all other screenshots were skipped.
    - After switching the names of screenshots B and C in the Finder (`A,C,B,D,E`) ➡️  screenshots `B`, `D` and `E` got deleted from the edit (leaving only `A` and `C` in PSC), then screenshots `B`, `D` and `E` got re-uploaded.
    - After adding new screenshot `F` locally ➡️  all screenshots `A–E` got skipped, new screenshot `F` got uploaded.


supply/lib/supply/client.rb
```diff
@@ -544,29 +549,14 @@ def fetch_images(image_type: nil, language: nil)
...
-          uri.scheme,
-          uri.userinfo,
-          uri.host,
-          uri.port,
-          uri.path
-        ].join
-
-        UI.verbose("Initial URL received: '#{url}'")
-        UI.verbose("Removed params ('#{uri.query}') from the URL")
-        UI.verbose("URL after removing params: '#{clean_url}'")
```
AliSoftware, line 560: All those lines were dead code even before my change: as you can see, `clean_url` was never used in the returned value after all.

supply/spec/uploader_spec.rb
```diff
@@ -221,5 +221,200 @@ def find_obbs
...
+              id_#{screenshot_type}_image1.png
+              id_#{screenshot_type}_image2.png
+              id_#{screenshot_type}_image3.png
+              new-id_#{screenshot_type}_image4.png
+              new-id_#{screenshot_type}_image5.png
+              new-id_#{screenshot_type}_image6.png
+            ]
+            expect(final_remote_images_ids[screenshot_type]).to eq(expected_final_images_ids)
+          end
+        end
```
AliSoftware, line 416: This unit test is the most interesting case I wanted to get right, as it shows a case where you have the same images locally and remotely, but you swapped the order of some of your screenshots:

 - On the Play Store you initially had the images that correspond to `image0.png`, `image1.png`, `image2.png`, `image4.png`, `image3.png`, `image5.png`, `image6.png`, currently appearing in that order in the Play Store
 - But since then, you decided to swap the order of `image4.png` and `image3.png` and now you want to re-upload the images (which will be uploaded in alphabetical order, as `supply` has always done)
 - This will thus:
    - Keep `image0.png`, `image1.png` and `image2.png` untouched
    - Delete the `image4.png`, which is currently after `image2.png` in PlayStore — because it's now out of order compared to the local files
    - But keep `image3.png` in PlayStore (that came after `image4.png`, but will now come after `image2.png` since `image4.png` was deleted from the PlayStore edit)
    - And then delete `image5.png` and `image6.png`, to finally re-upload `image4.png`, `image5.png` and `image6.png` in the correct order.

This looks like this when running this spec with `DEBUG=1 bundle exec rspec supply/spec/uploader_spec.rb`:
```
…
[22:30:47]: 🔍 Checking phoneScreenshots checksums...
[22:30:47]: 🟰 Skipping upload of screenshot image0.png as remote sha256 matches.
[22:30:47]: 🟰 Skipping upload of screenshot image1.png as remote sha256 matches.
[22:30:47]: 🟰 Skipping upload of screenshot image2.png as remote sha256 matches.
[22:30:47]: 🚮 Deleting pt-BR screenshot id #id_phoneScreenshots_image4.png as it does not exist locally or is out of order...
[22:30:47]: 🟰 Skipping upload of screenshot image3.png as remote sha256 matches.
[22:30:47]: 🚮 Deleting pt-BR screenshot id #id_phoneScreenshots_image5.png as it does not exist locally or is out of order...
[22:30:47]: 🚮 Deleting pt-BR screenshot id #id_phoneScreenshots_image6.png as it does not exist locally or is out of order...
[22:30:47]: ⬆️  Uploading screenshot image4.png...
[22:30:47]: ⬆️  Uploading screenshot image5.png...
[22:30:47]: ⬆️  Uploading screenshot image6.png...
…
```

supply/lib/supply/uploader.rb
```diff
@@ -280,13 +290,30 @@ def upload_images(language)
...
+          existing_images.each do |image|
+            if image.sha256 == first_path_checksum
+              UI.message("🟰 Skipping upload of screenshot #{paths.first} as remote sha256 matches.")
+              paths.shift # Remove first path from the list of paths to be uploaded
+              first_path_checksum = paths.empty? ? nil : Digest::SHA256.file(paths.first).hexdigest
+            else
+              UI.message("🚮 Deleting #{language} screenshot id ##{image.id} as it does not exist locally or is out of order...")
+              client.clear_screenshot(image_type: screenshot_type, language: language, image_id: image.id)
+            end
+          end
```
AliSoftware, line 310: Sadly, there doesn't seem to be an Google API on `AndroidPublisherService` that allows to just reorder the screenshots, hence the need to delete the screenshots that are out of order and re-upload them, so that when new uploaded screenshots go at the end of the list, the final order is still correct in Google Play.

This is why I only compare with the checksum of the _next_ local file (well, in practice, checksum of `paths.first`, while `paths` items are being `shift`'ed and removed from the `paths` array as they get skipped), instead of using something like `all_local_checksums.include?(image.sha256)` to just check if the remote image matches an existing local image, which would not guarantee that we'd end up with the expected order of the screenshots in the end.

supply/lib/supply/options.rb
```diff
@@ -210,6 +210,12 @@ def self.available_options
                                      description: "Whether to skip uploading SCREENSHOTS",
                                      type: Boolean,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :sync_image_upload,
+                                     env_name: "SUPPLY_SYNC_IMAGE_UPLOAD",
+                                     optional: true,
+                                     description: "Weather to use sha256 comparison to skip upload of images and screenshots that are already in Play Store",
```
rogerluan, line 216:
```suggestion
                                     description: "Whether to use sha256 comparison to skip upload of images and screenshots that are already in Play Store",
```

supply/lib/supply/options.rb
```diff
@@ -210,6 +210,12 @@ def self.available_options
                                      description: "Whether to skip uploading SCREENSHOTS",
                                      type: Boolean,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :sync_image_upload,
+                                     env_name: "SUPPLY_SYNC_IMAGE_UPLOAD",
+                                     optional: true,
```
rogerluan, line 215: Does it need to be optional since it's just a boolean that has a default value?

rogerluan (commented):
@AliSoftware more of an operational question... With `deliver`, if you pass `skip_screenshots: false`, it will always delete all screenshots and upload them all again even if they're still the same images as before — right? That's what I observe when testing, at least... So what I've been doing for the past few years was to pas `skip_screenshots: true` when I want to update any screenshot, and false otherwise.

First, is your workflow for `deliver` different than this? Does it behave differently for you? 🤔 
Second, wouldn't it make sense for you to apply the same behavior I just described to your Android workflow, and thus pass `skip_upload_images: true` and/or `skip_upload_screenshots: true` when the images haven't changed, and `false` otherwise?

From your PR description alone it sounds like both `deliver` and `supply` are on the same page in terms of implementation (except ASC allows us to sort the screenshots, but that's besides the point here haha).

WDYT? Could you clarify if I understood something wrong? :)
Thanks!

AliSoftware:
Ah I should have clarified in the PR description sorry: on our end we use [the `sync_screenshots: true` option of _deliver_](https://github.com/fastlane/fastlane/blob/1e45f926121d91b0410e2345d7ce6860622937f6/deliver/lib/deliver/options.rb#L165-L169), which allows to always use `skip_screenshots: false` in our lane, and let _deliver_ [figure out what screenshot, if any, to re-upload, which ones are already there, and which ones to delete](https://github.com/fastlane/fastlane/blob/1e45f926121d91b0410e2345d7ce6860622937f6/deliver/lib/deliver/sync_screenshots.rb#L114-L117).

This is the behavior of _deliver_ that this PR is bringing to _supply_.

Note that in _deliver_ that feature is behind a `FASTLANE_ENABLE_BETA_DELIVER_SYNC_SCREENSHOTS` feature flag, while I chose not to bother making one for _supply_ in this PR because I've made the feature opt-in in the first place already (`default_value: false`)

supply/lib/supply/options.rb
AliSoftware: Good point 👍

joshdholtz (approved):

supply/lib/supply/uploader.rb
```diff
@@ -280,13 +290,30 @@ def upload_images(language)
...
+        else
+          client.clear_screenshots(image_type: screenshot_type, language: language) unless Supply.config[:sync_image_upload]
```
AliSoftware, line 312: Ooops, just realized that this `unless` is unneeded (just a leftover of a previous implementation) now that this line is already within the `else` clause of the `if Supply.config[:sync_image_upload]` condition 🙃 
```suggestion
          client.clear_screenshots(image_type: screenshot_type, language: language)
```

supply/lib/supply/options.rb
```diff
@@ -210,6 +210,11 @@ def self.available_options
                                      description: "Whether to skip uploading SCREENSHOTS",
                                      type: Boolean,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :sync_image_upload,
```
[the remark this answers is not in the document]
AliSoftware: The reason I didn't name it `sync_screenshots` is deliberate: this is because this not only applies to screenshots, but also to Play Store artwork — typically the `featuredGraphic` and `icon` images that are not screenshots but rather the banner and large icon shown in the Google Play's download page for the app.

I thought calling it `sync_screenshots` would make people think that this only applied to screenshots (as in `skip_upload_screenshots: false`) while it also applies to `skip_upload_images: false` too.
Arguably, because the app artwork is referenced by the term `…images…` in the `skip_upload_images` option, one could argue that calling my new parameter `sync_image_upload` would make people think that it applies to that option only (i.e. where `image` term here would be interpreted as "the artwork images uploaded as part of `skip_upload_images: false`), while the intent of the term `…image…` in `sync_image_upload` here is more to say "sync the upload of every kind of [PNG, JPEG, …] image file".

But I couldn't find a term that would encompass both "each image file, be it a screenshot or an artwork image" other than just the term "image", hence settling for this.

---

As for trying to make the `deliver` and `supply` option names in sync, that ship as sadly already sailed: For example, [`deliver` uses `skip_screenshots`](https://github.com/fastlane/fastlane/blob/master/deliver/lib/deliver/options.rb#L137) while [`supply` uses `skip_upload_screenshots`](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb#L207) (and [`skip_upload_images`](https://github.com/fastlane/fastlane/blob/master/supply/lib/supply/options.rb#L207)), with `_upload_` in that option name for `supply` while `deliver` doesn't have it 🤷

supply/lib/supply/uploader.rb
```diff
@@ -270,7 +270,17 @@ def upload_images(language)
         path = Dir.glob(search, File::FNM_CASEFOLD).last
         next unless path
 
-        UI.message("Uploading image file #{path}...")
+        if Supply.config[:sync_image_upload]
+          UI.message("🔍 Checking #{image_type} checksum...")
```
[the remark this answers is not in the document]
AliSoftware: Fair point.

I've already merged the PR though (in the hope that we'd include this PR in the new release that @joshdholtz said he'd do later today to fix [the Google API random errors](https://github.com/fastlane/fastlane/issues/21507) as we talked about in Slack), so it's a bit too late 😅.
Not sure if it's worth opening a subsequent PR just to fix this wording? (but if you feel strongly about it I don't mind us doing so to fix it)

supply/lib/supply/options.rb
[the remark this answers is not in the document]
rogerluan: Good points! 😅 I think you named it well 😊 Thanks for clarifying!

supply/lib/supply/uploader.rb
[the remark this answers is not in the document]
rogerluan: Either way works for me! Not strongly opinionated, just think it'd be more befitting for what's actually happening :) I'll leave this up to you! If you decide to open the PR, tag me to review and I'll approve and merge it asap 😊

PR #21528, opened 2023-09-22
[spaceship] allow to specify fetched fields

nekrich:
### Checklist

### Motivation and Context

Currently, I'm working on `match` optimization. I've figured out that `match` fetches certificates with all fields, including enormous `certificateContent` and `csrContent`, and match doesn't need them.
A reduced amount of fetched fields leads to faster response generation and faster transmission to the client.

### Description

Changes are really small, I just added a `fields` parameter to the api_client, certificate, device, bundle id and profile.

### Testing Steps

`Spaceship::ConnectAPI::Device.all(fields: 'platform,status')`

### DIFF
--- a/spaceship/lib/spaceship/connect_api/api_client.rb
+++ b/spaceship/lib/spaceship/connect_api/api_client.rb
@@ -87,13 +87,14 @@ module Spaceship
         return @token.nil?
       end
 
-      def build_params(filter: nil, includes: nil, limit: nil, sort: nil, cursor: nil)
+      def build_params(filter: nil, includes: nil, fields: nil, limit: nil, sort: nil, cursor: nil)
         params = {}
 
         filter = filter.delete_if { |k, v| v.nil? } if filter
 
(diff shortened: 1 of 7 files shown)


Chupaka:
Hey @nekrich, sorry for hijacking your PR, did you look at 'limit' parameter for API requests? Looks like it defaults to 20 responses per page, and max limit is 200 (? I believe I saw it on some Apple website...), but I can't find a way to set it manually... That would speed up things for accounts with many items 🤞

nekrich:
@Chupaka thanks for suggestion, I'll look into it too. Reducing requests count is awesome. 

But even with bigger limit, there is an issue with certificates which include excessive data match and sigh don't really need. And it spreads to the profiles when we fetch them with certificates. Plus I don't think (didn't test it in practice) that certificateContent and csrContent will have a good compression ratio while transferring.

getaaron (approved):

getaaron:
Thanks for the PR, @nekrich. Looks great!

PR #21432, opened 2023-07-31
[trainer] fix issues where number of failures would always be zero

tahirmt:
### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I see several green `ci/circleci` builds in the "All checks have passed" section of my PR ([connect CircleCI to GitHub](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-) if not)
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context

Closes #21565

### Description
If the test case name has spaces or any special characters, the reported `number_of_failures` will always be 0. This will also not include any failed tests inside the junit report.

The reason this was happening was because during the comparison to check the test failure there was a sanitization step that's not applied to the identifier itself so it never matched.

This is from the log output of such a test.
```
** TEST FAILED **
[15:11:03]: Exit status: 65
+--------------------+---+
|      Test Results      |
+--------------------+---+
| Number of tests    | 3 |
| Number of failures | 0 |
+--------------------+---+

[15:11:04]: Test execution failed. Exit status: 65
```

### Testing Steps

The easiest way to reproduce it was to use https://github.com/Quick/quick because Quick 7 produces test case names with spaces if the spec names had spaces in it. And when you run `scan` the output will always say the number of failures is 0.

I have attached a sample project.

1. Download project
2. Run `bundle install`
3. Run `bundle exec fastlane test` 

When it runs the output will show number of failures as 0.

Then you can uncomment the branch line in the `Gemfile` and run bundle install and switch to this branch. Then you will see that the number of failure is 1.

[SpaceTests.zip](https://github.com/fastlane/fastlane/files/12221320/SpaceTests.zip)

PR #21565, opened 2023-10-09
[trainer] sanitize test case name and test identifier consistently when looking up test failures

TheMetalCode:
### Checklist

### Motivation and Context

I recently had an issue in CI builds where failing tests were not being reported as failures once trainer converted the xcresult to junit xml. After digging in and debugging, I found that the existing logic in `Trainer::XCResult::ActionTestMetadata#find_failure` was expecting the test identifier to have a particular format that didn't end up matching. This led me to the insight that instead of having this expectation of the identifier, we could sanitize both it and the test case name consistently, then compare them for a potential match.

Using xcresult artifacts from the failing CI runs, this fix was effective. Unfortunately, since those xcresult artifacts contain proprietary data, I've yet to figure out how to unit test this safely. 🙈


trainer/lib/trainer/xcresult.rb
```diff
@@ -184,16 +184,12 @@ def find_failure(failures)
...
           found_failure = failures.find do |failure|
-            # Clean test_case_name to match identifier format
-            # Sanitize for Swift by replacing "." for "/"
-            # Sanitize for Objective-C by removing "-", "[", "]", and replacing " " for ?/
-            sanitized_test_case_name = failure.test_case_name
-                                              .tr(".", "/")
-                                              .tr("-", "")
-                                              .tr("[", "")
-                                              .tr("]", "")
-                                              .tr(" ", "/")
```
TheMetalCode, line 195: In my case, this ended up being the unmet expectation - not all spaces were replaced by `/` in the `identifier`. So the comparison ended up being [somewhat redacted due to proprietary info]:

```
[1] pry(#<Trainer::XCResult::ActionTestMetadata>)> self.identifier
=> "MyPrivateSpec/my test condition, should return an error()"
[2] pry(#<Trainer::XCResult::ActionTestMetadata>)> sanitized_test_case_name
=> "MyPrivateSpec/my/test/condition,/should/return/an/error()"
```

Anyways, the bet I'm making with this PR is: rather than have any expectation of how `self.identifier` is formatted, sanitize both it and `failure.test_case_name` in the same fashion then compare.

rogerluan (changes requested):
Hi @TheMetalCode 👋 

Thanks for this PR! Your rationale makes sense to me, and I think this is a good solution moving forward.

Do you mind adding unit tests to this PR, hopefully with real-life strings for both test case name and identifiers? Ideally we should cover the happy path (scenarios where the previous implementation used to work) as well as the edge case you were seeing (so that it fails, when run using the previous comparison algorithm, but succeeds now using the new algorithm). That would be super!

Let me know if you need any support adding these tests 🤗 thanks again for your hard work! 🙏 

PS: If you could also add steps to reproduce this issue locally so we can verify your results, that would be great. The simpler these steps are, the better (extra points if you can provide a demo project where the issue is reproducible) 🙇

rogerluan:
Heads up: no need to implement the unit tests in this PR. I decided to merge part of this PR's solution into the PR https://github.com/fastlane/fastlane/pull/21432 where we already have unit tests and a sample project, so once we merge that PR, this one can be closed 🙇 

Thanks for the contribution in this PR! 🙌

TheMetalCode:
> Heads up: no need to implement the unit tests in this PR. I decided to merge part of this PR's solution into the PR #21432 where we already have unit tests and a sample project, so once we merge that PR, this one can be closed 🙇
> 
> Thanks for the contribution in this PR! 🙌

That's great, thank you!

My struggle with unit tests on this one is that I had them locally but they were relying on sample xcresult bundles from a private project and I didn't know of a way to edit xcresult files to make them non-proprietary. This private project is using Quick/Nimble though, didn't realize that in and of itself would create the error condition.

Anyways, happy to have influenced #21432, looking forward to the fix!

Issue #20960, opened 2023-01-02
Could not install WWDR certificate


dev-loic:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

Hello everyone 👋🏻

We are using circle ci with fastlane 2.211.0. 
Randomly when we do the matching certificates, the CI fails with the following error `Could not install WWDR certificate`.
I say randomly because it does not happen each time and the logs are not relevant both ways. 

Thanks in advance for your insights

##### Command executed

`match`

##### Complete output when running fastlane, including the stack trace and command used

Here is the output **when the CI fails**
<details>
  <pre> 
chruby: unknown Ruby: ruby-2.6.8
Your RubyGems version (3.0.3.1) has a bug that prevents `required_ruby_version` from working for Bundler. Any scripts that use `gem install bundler` will break as soon as Bundler drops support for your Ruby version. Please upgrade RubyGems to avoid future breakage and silence this warning by running `gem update --system 3.2.3`
[⠋] 🚀 [⠙] 🚀 [⠹] 🚀 [⠸] 🚀 [⠼] 🚀 [⠴] 🚀 [⠦] 🚀 [✔] 🚀 
+------------------------------------------+---------+--------------------------+
|                                 Used plugins                                  |
+------------------------------------------+---------+--------------------------+
| Plugin                                   | Version | Action                   |
+------------------------------------------+---------+--------------------------+
| fastlane-plugin-diawi                    | 2.1.0   | diawi                    |
| fastlane-plugin-clean_testflight_testers | 0.3.0   | clean_testflight_testers |
+------------------------------------------+---------+--------------------------+

[15:43:52]: Sending anonymous analytics information
[15:43:52]: Learn more at https://docs.fastlane.tools/#metrics
[15:43:52]: No personal or sensitive data is sent.
[15:43:52]: You can disable this by adding `opt_out_usage` at the top of your Fastfile
[15:43:53]: ----------------------------------------
[15:43:53]: --- Step: Verifying fastlane version ---
[15:43:53]: ----------------------------------------
[15:43:53]: Your fastlane version 2.211.0 matches the minimum requirement of 2.114.0  ✅
[...]

PR #21578, opened 2023-10-13
[match] propagate keychain when installing wwdr certificates

rabbitinspace:
### Checklist

### Motivation and Context

After updating from fastlane 2.202.0 to 2.216.0 `fastlane match -s $KEYCHAIN -p $PASSWORD` started to fail for me with the following error:

```shell
[03:54:46]: 🔓  Successfully decrypted certificates repo
[03:54:46]: Installing certificate...
[03:54:46]: $ security find-certificate -a -c 'Apple Worldwide Developer Relations' -p /Users/ec2-user/Library/Keychains/login.keychain-db
[!] Could not install WWDR certificate
🚨 Error: The command exited with status 1
``` 

I import wwdr certs manually after creating a keychain, so something is broken here. The weird thing I found is `security find-certificate` is invoked for the login keychain despite me specifying a separate one. After digging a bit, I figured the `install_missing_wwdr_certificates` method doesn't take into account the `-s` option.

### Description

This PR makes the `-s` option to be propagated to the `install_missing_wwdr_certificates` method instead of unconditionally using `wwdr_keychain`. I found [this issue](https://github.com/fastlane/fastlane/issues/20960) which supposed to be fixed, but I guess this edge case isn't handled.

Also, I hope I fixed the tests the right way.

### Testing Steps

```shell
KEYCHAIN_NAME=certs
KEYCHAIN_PATH="${HOME}/Library/Keychains/${KEYCHAIN_NAME}.keychain-db"
KEYCHAIN_PASS=12345HACKER

# create a new keychain first

if [[ -f "${KEYCHAIN_PATH}" ]]; then
  fastlane run delete_keychain keychain_path:"${KEYCHAIN_PATH}"
fi
fastlane run create_keychain \
  password:"${KEYCHAIN_PASS}" \
  path:"${KEYCHAIN_PATH}" \
  unlock:true \
  timeout:1800

# import all wwdr certs

certs=( AppleWWDRCA.cer AppleWWDRCAG2.cer AppleWWDRCAG3.cer AppleWWDRCAG4.cer AppleWWDRCAG5.cer AppleWWDRCAG6.cer )
for cert in "${certs[@]}"; do
  fastlane run import_certificate \
    certificate_path:"${cert}" \
    keychain_password:"${KEYCHAIN_PASS}" \
    keychain_path:"${KEYCHAIN_PATH}"
done

# run fastlane match to import certs and profiles

fastlane match development -s "${KEYCHAIN_NAME}" -p "${KEYCHAIN_PASS}"
```


fastlane_core/lib/fastlane_core/cert_checker.rb
```diff
@@ -111,12 +112,12 @@ def self.list_available_developer_id_installer(in_keychain: nil)
       `#{commands.join(' ')}`
     end
 
-    def self.installed_wwdr_certificates
+    def self.installed_wwdr_certificates(keychain)
```
lacostej, line 115: from an API compatibility perspective, I wonder if we should change this to

```
def self.installed_wwdr_certificates(keychain: nil)
   keychain ||= wwdr_keychain
```

Same for other methods.

WDYT?

rabbitinspace: do you mean compatibility with other fastlane code which uses this class or users who might use some of its methods?
the first one should already be sorted as (iirc) I didn't change signatures for used methods
as for the second one, there are no clear public/private separation to me so I assumed those methods are implementation detail as they doesn't look like a public api to me
but happy to use your suggestion if you think it would be the right thing to do

lacostej: @rabbitinspace I meant compatibility with potential external code. 

We unfortunately have a grey way of defining API compatibility in fastlane. I know some users are sometimes using parts of the tools programmatically and anything that isn't made `private` can be used. 

So maybe

```ruby
keychain ||= wwdr_keychain # backwards compatibility
```

would make sense.

I'll ask the team if we can agree on what the clarify what the project public API is.

rabbitinspace: thanks! addressed in 6ad8b71bd

rabbitinspace:
conflicts resolved, the CI failure is possibly due to https://github.com/fastlane/fastlane/issues/21663

lacostej (approved):
LGTM. Anyone else?

rogerluan (approved):
💪 LGTM! Simple and safe. Thanks for your contribution @rabbitinspace ! 🚀

DuMaM:
Yay no more changing default keychain :D thanks man

Issue #21109, opened 2023-03-06
Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired.


tutudeloitte:
### New Issue Checklist

- [ x ] Updated fastlane to the latest version. **Version 2.212.1**
- [ x ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ x ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ x ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues) **Checked issues #18100, #19072 and #20270**
- [ x ] I changed key durations to value of 100, 500, 1150, 1200

### Issue Description
I'm trying to use Fastlane in Jenkins to automate uploading .ipa file onto Test Flight.
Developers built .ipa file, which I uploaded to Bitbucket. Commands in Jenkins below.
I have created several App Store API keys with Developer and App Manager permissions to verify that it's not permission or expiration time issues, but all have produced the error below

```
INFO [2023-03-06 22:07:23.86]: Creating authorization token for App Store Connect API
DEBUG [2023-03-06 22:07:23.90]: App identifier (com.transamerica.wfg365)
Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens
Token has expired or has been revoked! Trying to refresh...
```
**Expected behavior**: IPA file uploaded to Test Flight
**Actual behavior**: Authentication error message like above
**Folder structure**

##### Command executed

I have both Fastfile and used fastlane CLI

Commands for fastlane CLI
```
unzip IPA.zip
fastlane pilot upload --api_key_path ./fastlane/key.json --verbose
```
Content of key.json
```
{
        "key_id": "[KEY_ID]",
        "issuer_id": "[ISSUER_ID]",
        "key_content": "-----BEGIN PRIVATE KEY-----\nLINE1\nLINE2\nLINE3\n-----END PRIVATE KEY-----",
        "duration": 100
}
```

Fastfile content

[2023-03-06T22:07:39.704Z] ruby
[2023-03-06T22:07:39.704Z] fastlane_version "2.24.0"
[2023-03-06T22:07:39.704Z] default_platform :ios
[2023-03-06T22:07:39.704Z] 
[...]

Issue #20516, opened 2022-07-29
Authentication credentials are missing or invalid


mahgo:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
I'm trying to set up uploading to AppStore/TestFlight on a new project. I have set up a key in App Store Connect with the App Manager role, and have set up the following auth_key_json:
```
{
    "key_id":"{key id}",
    "is_key_content_base64":true,
    "issuer_id":"{issuer id}",   
    "key":"{base64 key, from calling base64 AuthKey.p8 -b 0 | pbcopy }",
    "in_house":false,
    "duration":500
}
```
When I run the below command, I get the below output including the error.

##### Command executed
`fastlane pilot upload --api_key_path auth_key.json -i app.ipa -a {my bundle id}`

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre>
  [✔] 🚀 
[14:43:02]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[14:43:04]: Creating authorization token for App Store Connect API

Looking for related GitHub issues on fastlane/fastlane...

[!] The request could not be completed because:
        Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens
  </pre>
</details>

### Environment

<details>
  <pre>
  <details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                                   |
| --------------------------- | ------------------------------------------------------- |
| OS                          | 10.15.7                                                 |
| Ruby                        | 3.1.2                                                   |
| Bundler?                    | false                                                   |
| Git                         | git version 2.24.3 (Apple Git-128)                      |
| Installation Source         | /usr/local/Cellar/fastlane/2.208.0/libexec/bin/fastlane |
| Host                        | Mac OS X 10.15.7 (19H2)                                 |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/3.1.2/lib                        |
[...]

PR #21583, opened 2023-10-19
[spaceship] decrease App Store Connect API token `issued-at-time` to prevent server rejection

xinsight:
### Checklist

### Motivation and Context

I have observed sporatic failures with delivering builds to testflight that result in an error message like this:

> Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens

After much debugging and with <a href="https://github.com/fastlane/fastlane/issues/20516#issuecomment-1552609002">this tip</a> I confirmed that my machine time was slightly ahead of Apple's servers which caused the AppStore Connect token to be rejected as the issued-at-time was in the future.

Similar issues have been reported in #20516 and #21109.

Syncing servers to Apple's servers (using ntp.apple.com) is one solution to this problem, however, not everyone is able or willing to sync their system's clocks to Apple's (and Apple's servers might be out-of-sync!) so a more robust and permanent solution is needed.

### Description

By reducing the issued-at-time (iat) in the jwt token, the token will only be rejected if a machine is over a minute in the future. (This value could of course, be increased, but one minute seems like a good starting point.)

Additionally, this fix also addresses a minor race condition in the unit tests where the issued time was being compared to the current time - which would fail if the clock second changed from the time when the token was generated to when the test was run.

### Testing Steps

To reproduce this issue, add 60 seconds to the `iat` value in the payload in: https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/connect_api/token.rb#L103 Attempting to perform any action that requires a token with spaceship will fail.

For example:

```
api_key = app_store_connect_api_key(
  key_id: "xxx",
  issuer_id: "xxx",
  key_filepath: "xxx.p8",
  duration: 1200, # optional (maximum 1200)
  in_house: false # optional but may be required if using match/sigh
)

latest_testflight_build_number(api_key: api_key, app_identifier: "xxx")
```

Repeat the command by removing 60 seconds from iat and it should pass.

### DIFF
--- a/spaceship/lib/spaceship/connect_api/token.rb
+++ b/spaceship/lib/spaceship/connect_api/token.rb
@@ -100,7 +100,8 @@ module Spaceship
 
         payload = {
           iss: issuer_id,
-          iat: now.to_i,
+          # reduce the issued-at-time in case our time is slighly ahead of apple's servers, which causes the token to be rejected
+          iat: now.to_i - 60,
           exp: @expiration.to_i,
           aud: 'appstoreconnect-v1'
         }
--- a/spaceship/lib/spaceship/connect_api/token.rb
+++ b/spaceship/lib/spaceship/connect_api/token.rb
@@ -100,7 +100,7 @@ module Spaceship
 
         payload = {
           iss: issuer_id,
-          # reduce the issued-at-time in case our time is slighly ahead of apple's servers, which causes the token to be rejected
+          # Reduce the issued-at-time in case our time is slighly ahead of Apple's servers, which causes the token to be rejected.
           iat: now.to_i - 60,
           exp: @expiration.to_i,
           aud: 'appstoreconnect-v1'
--- a/spaceship/lib/spaceship/connect_api/api_client.rb
+++ b/spaceship/lib/spaceship/connect_api/api_client.rb
@@ -184,7 +184,7 @@ module Spaceship
         if tries.zero?
           raise error
         else
-          msg = "Token has expired or has been revoked! Trying to refresh..."
+          msg = "Token has expired, issued-at-time is in the future, or has been revoked! Trying to refresh..."
           puts(msg) if Spaceship::Globals.verbose?
           @token.refresh!
           retry
--- a/spaceship/spec/connect_api/token_spec.rb
+++ b/spaceship/spec/connect_api/token_spec.rb
@@ -205,7 +205,7 @@ describe Spaceship::ConnectAPI::Token do
       payload, header = JWT.decode(token.text, key, true, { algorithm: 'ES256' })
 
       expect(payload['iss']).to eq(issuer_id)
-      expect(payload['iat']).to eq(Time.now.to_i)
+      expect(payload['iat']).to be < Time.now.to_i
       expect(payload['aud']).to eq('appstoreconnect-v1')
       expect(payload['exp']).to be > Time.now.to_i
 


artem-alexandrov:
It would be great to have this merged. I'm receiving this error very often sadly...

yolovenka (approved):

b-bondarenko (approved):

jarivers082461 (approved):

orschaef:
Same here... please lets get that merged!

artazar:
For some reason this started to happen more and more.
It would be great to complete his PR to confirm this is exactly the cause or if it requires a completely separate investigation.
Thanks in advance!

PS. In our case macos build agents are provided by CI and cannot be managed from our side, so it's not possible to sync/change time. Just restart the build on another agent, where it then works fine.

mletynski:
Guys can we just merge this PR please ? It's critical for our builds.

yasaswick94 (approved):

shichen85:
One more approval review let's go!

max-ott (approved):
Looking good! Let's get this merge with the next release!

bebecap (approved):

KasperAndersson:
Hi @max-ott , do you know when the next release is planned? Thanks in advance

bjornoleh:
I am currently testing by pointing Gemfile to the branch for this PR:

`gem 'fastlane', git: 'https://github.com/xinsight/fastlane', branch: 'decrease-token-issued-at-time-to-prevent-rejection'`

I am comparing the success rate by running the same workflow on schedule in two different repositories. One with the proposed changes, the other without. I can verify that the PR helps, but unfortunately, I am also seeing failures due to the same error when testing the PR. It seems like there can be more than 60 second time offsets between the GitHub macos-13 runner and the Apple servers. Perhaps the limit should be extended. There ismention of >200 seconds time difference in https://github.com/actions/runner/issues/2996, in case that is the same underlying problem.

The error I get is this:
`Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens`

Results:
Without this PR:
4 failures in 16 runs (25%)

With this PR:
2 failures in 14 runs (14%)

Initial testing of adding the 'Sync clock' step as mentioned by @b099l3 looks promising though.

kpriemchenko:
This is a good solution to an annoying problem! Let's merge this PR, please 👍

bjornoleh:
> This is a good solution to an annoying problem! Let's merge this PR, please 👍

The root cause with GithHub runner machine clocks should already have been sorted: https://github.com/actions/runner/issues/2996#issuecomment-1854895869. 

Are you still seeing this type of failures?

spaceship/lib/spaceship/connect_api/token.rb
```diff
@@ -100,7 +100,8 @@ def refresh!
 
         payload = {
           iss: issuer_id,
-          iat: now.to_i,
+          # reduce the issued-at-time in case our time is slighly ahead of apple's servers, which causes the token to be rejected
```
rogerluan, line 103:
```suggestion
          # Reduce the issued-at-time in case our time is slighly ahead of Apple's servers, which causes the token to be rejected.
```

rogerluan (approved):
I was on the fence in the beginning but this solution seems simple, efficient, and safe, IMO. LGTM 💪

Thanks for fixing this, @xinsight! I see a lot of people are facing this issue 🙏

bjornoleh:
> I was on the fence in the beginning but this solution seems simple, efficient, and safe, IMO. LGTM 💪
> 
> Thanks for fixing this, @xinsight! I see a lot of people are facing this issue 🙏

I think this change is redundant by now. Please see https://github.com/actions/runner/issues/2996#issuecomment-1854895869

rogerluan:
Thanks for pointing that out @bjornoleh ! 

That seems to solve this issue for everyone using fastlane with GitHub Actions, but might not be the case for other CI providers? I'm ok with closing this PR if everyone facing this issue were GHA users though 😮 

From a quick search, it seems like https://github.com/fastlane/fastlane/issues/21109's OP was experiencing this in Jenkins 😬

lacostej (approved):
let's add this.

PR #21591, opened 2023-10-20
[Rspecs] Make error message expectation more flexible in `runner_spec.rb` to support Ruby 2.6 error message format

AliSoftware:
### Checklist

### Motivation and Context

We recently merged https://github.com/fastlane/fastlane/pull/21587 into `trunk`, but while the CI was all green when it ran on the PR, it didn't run on Ruby 2.6 on macOS (because [we intentionally disabled those to avoid running out of CI credits](https://github.com/fastlane/fastlane/pull/21088)) and thus failed to detect a small glitch/difference in how Ruby 2.6 formats the error message we were testing in our Unit Tests (specs).

As a result, CI [started to fail on `trunk` after the PR landed](https://app.circleci.com/pipelines/github/fastlane/fastlane/5669/workflows/ff97e7a5-2191-4d6b-a57e-b90bca8ad599/jobs/91855)

### Description

That PR makes the `expect` of the offending tests more flexible in the way they check the expected `ArgumentError` message, so that the tests can pass on every version of Ruby and be flexible enough to accommodate the different formats used by different Ruby versions for that error message.

### Testing Steps

 - Check that CI is green.
 - If you still have an Intel mac around and thus can still install Ruby 2.6 on your machine, run `rbenv local 2.6.6` (or make sure you use a Ruby version 2.6.x system-wide) then run the tests using `bundle exec rspec`.

Note that CI [won't run on { macOS + Ruby 2.6 }](https://github.com/fastlane/fastlane/pull/21088) in the PR, so it won't be immediately obvious if we fixed that configuration. That being said, We still have some CI jobs that run 2.6 (especially on Linux and Windows) so those could be a way to check the behavior on Ruby 2.6, even if not on macOS.

### DIFF
--- a/fastlane/spec/runner_spec.rb
+++ b/fastlane/spec/runner_spec.rb
@@ -69,31 +69,28 @@ describe Fastlane do
       end
 
       context 'when a lane expects its parameters as keywords' do
-        def keywords_list(list)
-          # The way keyword lists appear in error messages is different in Windows & Linux vs macOS
-          if FastlaneCore::Helper.windows? || FastlaneCore::Helper.linux?
-            list.map(&:to_s).join(', ') # On Windows and Linux, keyword names don't show the `:` prefix in error messages
-          else
-            list.map(&:inspect).join(', ') # On other platforms, they do have `:` in front or keyword names
-          end
+        def keywords_error_message(error, *kwlist)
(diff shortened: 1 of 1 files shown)


rogerluan (approved):
💯 Great improvement!

rogerluan (approved):
I tested this locally on my machine on Ruby 2.6.6 and it works 💪

Issue #20578, opened 2022-08-19
Notarize action using notarytool can fail without a proper error message


WKD:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
When using the `notarize` action, sometimes the `notarytool` tool fails but doesn’t return any useful message in `statusMessage` so this is reported by fastlane as a very unhelpful error message `Could not notarize package with message ‘’` which makes debugging tricky.

It seems to especially happen when it encounters errors signing a framework included in the built app.

Using the `notarytool log` command yields useful information in such cases, so it could be a good idea to integrate its output whenever the `statusMessage` is empty.

NB: `notarytool log` returns a JSON similar to the snippet below which clearly states that in our case the error was signing a framework included into our own framework which is part of the app:

`{"logFormatVersion"=>1, "jobId"=>"[redacted]", "status"=>"Invalid", "statusSummary"=>"Archive contains critical validation errors", "statusCode"=>4000, "archiveFilename"=>"MySampleApp.zip", "uploadDate"=>"2022-08-19T17:31:32.205Z", "sha256"=>"[redacted]", "ticketContents"=>nil, "issues"=>[{"severity"=>"error", "code"=>nil, "path"=>"MySampleApp.app.zip/MySampleApp.app/Contents/Frameworks/MyFramework.framework/Versions/A/Frameworks/Atomics_301F3DE065572157_PackageProduct.framework/Versions/A/Atomics_301F3DE065572157_PackageProduct", "message"=>"The binary is not signed with a valid Developer ID certificate.", "docUrl"=>nil, "architecture"=>"x86_64"}, {"severity"=>"error", "code"=>nil, "path"=>"MySampleApp.app.zip/MySampleApp.app/Contents/Frameworks/MyFramework.framework/Versions/A/Frameworks/Atomics_301F3DE065572157_PackageProduct.framework/Versions/A/Atomics_301F3DE065572157_PackageProduct", "message"=>"The binary is not signed with a valid Developer ID certificate.", "docUrl"=>nil, "architecture"=>"arm64"}]}`

##### Command executed
- `fastlane mac notarize_build`

[...]

PR #21609, opened 2023-11-01
Notarization - Printing log on error

### LINKED ISSUE #21608: Notarize action - print_log parameter is not used for notarytool
**Issue Description**
When using the notarize action, sometimes the notarytool tool fails but doesn’t return any useful message in statusMessage so this is reported by fastlane as a very unhelpful error message : 
```
Could not notarize package with message ‘’
```
which makes finding the root cause almost impossible.

Even if the both `verbose` and `print_log` arguments are set to `true`, the message is still the same : 
```
Could not notarize package with message ‘’
```

Checking the implementation, we can see that `print_log` is not used for `notarytool` and that the `print_log` argument is only used for the old `altool` : 
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/notarize.rb#L48

Implementing the handling of the `print_log` argument in `notarytool` would fix this issue.
The full command for getting the actual errors would look like this : 
```
xcrun notarytool log "<submission-uuid>" <the auth parameters (eg.: --keychain <keychainProfileName> which profile was generated using `xcrun notarytool store-credentials`)>
```

szigetics:
### Checklist

### Motivation and Context

When the notarization is failing via `altool` there is no useful information printed.
Resolves #21608 and #20578 .

### Description

Using `xcrun altool log` to request additional info about the error.

### Testing Steps

-

### DIFF
--- a/fastlane/lib/fastlane/actions/notarize.rb
+++ b/fastlane/lib/fastlane/actions/notarize.rb
@@ -115,6 +115,7 @@ module Fastlane
             UI.user_error!("Could not notarize package with message '#{log_request_response}'")
           else
             UI.user_error!("Could not notarize package. To see the error, please set 'print_log' to true.")
+          end
         else
           UI.crash!("Could not notarize package with status '#{notarization_info['status']}'")
         end
--- a/fastlane/lib/fastlane/actions/notarize.rb
+++ b/fastlane/lib/fastlane/actions/notarize.rb
@@ -100,7 +100,7 @@ module Fastlane
             UI.success("Successfully notarized and stapled package")
           end
         when 'Invalid'
-          if submission_id && print_log:
+          if submission_id && print_log
             log_request_parts = [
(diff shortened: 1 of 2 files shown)


lacostej:
> @szigetics thanks for the PR! the contribution looks good
> 
> Looks like you checked:
> 
> > I see several green ci/circleci builds in the "All checks have passed" section of my PR (connect CircleCI to GitHub if not)
> 
> however I don't see the CircleCI builds — could you please doublecheck this?

Aren't they [here](https://github.com/fastlane/fastlane/runs/18263583035)?

szigetics:
> but the information didn't propagate to the PR checks

@getaaron : Do you know what could cause this? Should we just simply try to re-trigger the builds?
Other than this : does the PR look good? What do you think, when could this change be released? We would love to put our hands on a new version with this change... 🙂

getaaron (approved):

PR #21629, opened 2023-11-09
allow --env CLI flag to use relative env file paths  

shamilovtim:
### Checklist

### Motivation and Context

In multiplatform projects env files could be kept in a root directory especially when multiple platforms share the same envs to stay DRY. For example:
`../env/.env.development` might be for both ios and android in a flutter, react native, or kotlin multiplatform project and same with  `../env/.env.release`. However, the `fastlane foo --env "../../.env.development"` currently doesn't work due to the .env approach being hardcoded to only pass a simple string rather than a potential relative path to the env file.  

I tried alternatives like Dotenv.load(:options[myoption]) but the problem is that these don't work in Matchfile and Appfile which introduces a bunch of unnecessary hurdles and wasted time for engineers. It shouldn't be this hard to load an env file from a different directory if we want.

### Description

I decided that rather than introduce a breaking change I could just check if someone is trying to pass a relative path I could just expand the relative path, and otherwise keep the old approach. 

Personally I think a breaking change would be better because the flexibility of loading an env file from anywhere is much better than the current approach which was presumably intending to save the user from typing only 4 characters `.env.`. 

### Testing Steps

To make sure the old way works pass an env with `--env foo` to reach `.env.foo` file in the fastlane directory.
To make sure the new way works pass an env with `--env ../../.env.foo` to reach `.env.foo` file in a different directory.

Please note that the current local testing env has a bunch of flakes.

### DIFF
--- a/fastlane/lib/fastlane/helper/dotenv_helper.rb
+++ b/fastlane/lib/fastlane/helper/dotenv_helper.rb
@@ -40,7 +40,16 @@ module Fastlane
 
         # Loads .env file for the environment(s) passed in through options
         envs.each do |env|
-          env_file = File.join(base_path, ".env.#{env}")
+          # Determine if `env` is a relative path and construct `env_file` accordingly
+          if env.include?('..') || env.start_with?('./') || env.start_with?('.\\')
+            # `env` appears to be a relative path
+            # Expand it from the current working directory
+            env_file = File.expand_path(env)
+          else
+            # `env` does not appear to be a relative path
+            # Use `File.join` to construct `env_file` with `base_path`
+            env_file = File.join(base_path, ".env.#{env}")
+          end
           UI.success("Loading from '#{env_file}'")
           Dotenv.overload(env_file)
         end


rogerluan (commented):
I agree with @shamilovtim 's points ([shared here](https://github.com/fastlane/fastlane/pull/21631#issuecomment-1806370704)), conceptually. Just not sure whether the assumptions made during the implementation are okay to be made (namely: assumptions of what is a relative path and what's not) — the implementation ended up making a lot of assumptions and restricting how the user should and should not use the API. 

It's complex because I think OP tried to keep the current implementation fully working, which is understandable and expected, but this new API/UX might not work well alongside the previous one if we're keeping it backwards compatible... 

Perhaps we should create a new argument name? Instead of `--env`, name it something different like `--dotenv`, `--envbutbetter`, etc (I'm jk, just illustrating lol) — this way we'd be able to keep the previous implementation intact and also implement a flexible new implementation that makes no assumptions about the string the user sends us.

Thoughts?

shamilovtim:
@rogerluan 
Can modify to `--envPath` and push a breaking change to deprecate `--env`. is semver being followed?

rogerluan:
There are no plans to push breaking changes (i.e. major bump) as of right now. We can plan for it, though, there's a Discussion in this repo about fastlane 3.0, I think it'd be a good topic to bring up :)

rogerluan (changes requested):
Could you add unit tests? 🙏

PR #21631, opened 2023-11-10
[action] New `dotenv` action for overloading and loading dotenv files from default or custom directory

joshdholtz:
### Motivation and Context

I've been manually loading dotenv files for years and this should really be easier and built into _fastlane_

### Description

New `dotenv` action with:
- `name`
- `path`
- `overload`
- `fail_if_missing`

#### Example

```rb
before_all do
  dotenv(name: 'secret')
  # or
  dotenv(name: 'secret', fail_if_missing: true)
  # or
  dotenv(name: 'release', path: './platforms/ios')
end

lane :release do 
  # Stuff
end
```

### Testing Steps

N/A

### DIFF
--- a/fastlane/lib/fastlane/actions/dotenv.rb
+++ b/fastlane/lib/fastlane/actions/dotenv.rb
@@ -9,15 +9,10 @@ module Fastlane
         overload = params[:overload]
         fail_if_missing = params[:fail_if_missing]
 
-        if path
-          env_file = File.join(path, env_file)
-        else
-          base_path = Fastlane::Helper::DotenvHelper.find_dotenv_directory
-          env_file = File.join(base_path, env_file)
-        end
+        env_path = File.join(path, env_file)
 
         if !File.exist?(env_file)
-          error = "Cannot find request dotenv file at '#{env_file}'"
+          error = "Cannot find dotenv file at '#{env_path}'"
 
           if fail_if_missing
             UI.user_error!(error)
@@ -25,11 +20,11 @@ module Fastlane
 
           UI.error(error)
         elsif overload
-          UI.success("Overloading environment variables from '#{env_file}'")
-          Dotenv.overload(env_file)
+          UI.success("Overloading environment variables from '#{env_path}'")
+          Dotenv.overload(env_path)
         else
-          UI.success("Loading environment variables from '#{env_file}'")
-          Dotenv.load(env_file)
+          UI.success("Loading environment variables from '#{env_path}'")
+          Dotenv.load(env_path)
         end
       end
 
@@ -41,11 +36,12 @@ module Fastlane
         [
           FastlaneCore::ConfigItem.new(key: :name,
                                        env_name: 'FL_DOTENV_NAME',
-                                       description: 'Load an dotenv file by name. Ex: ios, android, secret'),
+                                       description: 'Load an dotenv file by name. E.g.: ios, android, secret'),
           FastlaneCore::ConfigItem.new(key: :path,
                                        env_name: 'FL_DOTENV_PATH',
                                        description: 'Directory where the dotenv file is located. Defaults to working directory',
-                                       optional: true),
+                                       default_value: Fastlane::Helper::DotenvHelper.find_dotenv_directory,
+                                       default_value_dynamic: true),
           FastlaneCore::ConfigItem.new(key: :overload,
                                        env_name: 'FL_DOTENV_OVERLOAD',
                                        description: 'Whether environment variables should be overloaded',
@@ -74,7 +70,7 @@ module Fastlane
       def self.example_code
         [
           'dotenv(
-            name: "secret" # looking for .env.secret
+            name: "secret" # Looks for .env.secret
           )',
           'dotenv(
             name: "release",
--- a/fastlane/lib/fastlane/actions/dotenv.rb
+++ b/fastlane/lib/fastlane/actions/dotenv.rb
@@ -11,7 +11,7 @@ module Fastlane
 
         env_path = File.join(path, env_file)
 
-        if !File.exist?(env_file)
+        if !File.exist?(env_path)
           error = "Cannot find dotenv file at '#{env_path}'"
 
           if fail_if_missing
--- a/fastlane/lib/fastlane/actions/dotenv.rb
+++ b/fastlane/lib/fastlane/actions/dotenv.rb
@@ -1,0 +1,75 @@
+require 'dotenv'
+
+module Fastlane
+  module Actions
+    class DotenvAction < Action
+      def self.run(params)
+        env_file = ".env.#{params[:name]}"
+        path = params[:path]
+        overload = params[:overload]
+        fail_if_missing = params[:fail_if_missing]
+
+        if path
+          env_file = File.join(path, env_file)
+        else
+          base_path = Fastlane::Helper::DotenvHelper.find_dotenv_directory
+          env_file = File.join(base_path, env_file)
+        end
+
+        if !File.exist?(env_file)
+          error = "Cannot find request dotenv file at '#{env_file}'"
+
+          if fail_if_missing
+            UI.user_error!(error)
+          end
+
+          UI.error(error)
+        elsif overload
+          UI.success("Overloading environment variables from '#{env_file}'")
+          Dotenv.overload(env_file)
+        else
+          UI.success("Loading environment variables from '#{env_file}'")
+          Dotenv.load(env_file)
+        end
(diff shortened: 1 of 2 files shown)


fastlane/lib/fastlane/actions/dotenv.rb
```diff
@@ -0,0 +1,91 @@
...
+      def self.run(params)
+        env_file = ".env.#{params[:name]}"
+        path = params[:path]
+        overload = params[:overload]
+        fail_if_missing = params[:fail_if_missing]
+
+        if path
+          env_file = File.join(path, env_file)
+        else
+          base_path = Fastlane::Helper::DotenvHelper.find_dotenv_directory
```
rogerluan, line 15: Perhaps move this to the path's default value in the `available_options`, WDYT? Then I think we can make it non-null too (users would still be able to omit it and it'd default to this dynamic value)

fastlane/lib/fastlane/actions/dotenv.rb
```diff
@@ -0,0 +1,91 @@
...
+          env_file = File.join(base_path, env_file)
+        end
+
+        if !File.exist?(env_file)
+          error = "Cannot find request dotenv file at '#{env_file}'"
```
rogerluan, line 20:
```suggestion
          error = "Cannot find dotenv file at '#{env_file}'"
```

fastlane/lib/fastlane/actions/dotenv.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      def self.author
+        "joshdholtz"
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :name,
+                                       env_name: 'FL_DOTENV_NAME',
+                                       description: 'Load an dotenv file by name. Ex: ios, android, secret'),
```
rogerluan, line 44: Nitpick 😄 

```suggestion
                                       description: 'Load an dotenv file by name. E.g.: ios, android, secret'),
```

fastlane/lib/fastlane/actions/dotenv.rb
```diff
@@ -0,0 +1,91 @@
...
+      end
+
+      def self.is_supported?(platform)
+        true
+      end
+
+      def self.example_code
+        [
+          'dotenv(
+            name: "secret" # looking for .env.secret
```
rogerluan, line 77: Nit:

```suggestion
            name: "secret" # Looks for .env.secret
```

fastlane/spec/actions_specs/dotenv_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          end
+        end
+      end
+
+      describe "when dotenv doesn't exist" do
+        it "fails silently with custom directory" do
+          allow(File).to receive(:exist?).with(anything).and_call_original
+          expect(File).to receive(:exist?).with('platforms/ios/.env.release').and_return(false).times.once
+
+          expect(UI).to receive(:error).with("Cannot find request dotenv file at 'platforms/ios/.env.release'")
```
rogerluan, line 62:
```suggestion
          expect(UI).to receive(:error).with("Cannot find dotenv file at 'platforms/ios/.env.release'")
```

fastlane/spec/actions_specs/dotenv_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        it "fails silently with custom directory" do
+          allow(File).to receive(:exist?).with(anything).and_call_original
+          expect(File).to receive(:exist?).with('platforms/ios/.env.release').and_return(false).times.once
+
+          expect do
+            Fastlane::FastFile.new.parse("lane :test do
+                dotenv(name: 'release', path: 'platforms/ios', fail_if_missing: true)
+              end").runner.execute(:test)
+          end.to raise_error("Cannot find request dotenv file at 'platforms/ios/.env.release'")
```
rogerluan, line 77:
```suggestion
          end.to raise_error("Cannot find dotenv file at 'platforms/ios/.env.release'")
```

rogerluan (changes requested):
Yeeeeees 🚀  

It's been awhile since I last touched dotenv in my projects, but I do remember having trouble with dotenv files, and not understanding why they're not autoloaded... When are they autoloaded and when are they not, and why? 🤔  We even have DotenvHelper and it seems to load and overload dotenv files for us, I just don't remember/know why they don't always work 😅 

Anyway I've also been loading them manually, this is a great addition 🤗  just left some minor comments! 💪

shamilovtim:
I've opened a relevant pull request here: #21629. My proposal is for dotenv to support being passed with a CLI flag, rather than requiring a special action. The underlying problem with Fastlane's implementation of environment variables is that the --env option is too inflexible. It does not allow for importing .env files from arbitrary paths or using different naming conventions. For example, why shouldn't someone be able to use production.env instead of .env.production, or why not  ../../foo.env? There appears to be no substantial reason for this limitation in the code; it seems to be a result of hardcoded constraints in the CLI flag.

Creating a DotEnv action in the Fastfile strikes me as just an alias for DotEnv.someMethod(), and it doesn't address the issue that .env files are still necessary in the Appfile, Matchfile, and other custom static Fastlane constructs. Since Fastlane actions cannot be invoked within Appfile and Matchfile, commands like fastlane match import will remain nonfunctional if the dotenv issue is resolved only in the Fastfile. However, if #21629 is merged, environment variables could be utilized in ad hoc Fastlane actions. This would allow for commands like fastlane match import --env ../../env/.env.production, enhancing flexibility and usability.

fastlane/lib/fastlane/actions/dotenv.rb
joshdholtz: Oh yeah lol... that makes A LOT more sense 😅

joshdholtz:
> I've opened a relevant pull request here: #21629. My proposal is to enhance the CLI flag, rather than requiring a special action. The underlying problem with Fastlane's implementation of environment variables is that the --env option is too inflexible. It does not allow for importing .env files from arbitrary paths or using different naming conventions. For example, why shouldn't someone be able to use production.env instead of .env.production, or why not ../../foo.env? There appears to be no substantial reason for this limitation in the code; it seems to be a result of hardcoded constraints in the CLI flag.
> 
> Creating a DotEnv action in the Fastfile strikes me as just an alias for DotEnv.someMethod(), and it doesn't address the issue that .env files are still necessary in the Appfile, Matchfile, and other custom static Fastlane constructs. Since Fastlane actions cannot be invoked within Appfile and Matchfile, commands like fastlane match import will remain nonfunctional if the dotenv issue is resolved only in the Fastfile. However, if #21629 is merged, environment variables could be utilized in ad hoc Fastlane actions. This would allow for commands like fastlane match import --env ../../env/.env.production, enhancing flexibility and usability.

@shamilovtim Hey! Thanks so much for replying to this 🙏 I started reviewing your PR and I totally see the need for it 👍 It's been a while since I've been in this part of the code so this PR is...
1) My warmup to getting back into this part of the code
2) Something that I've been wanting to add as an action for while

I have a dotenv test project that I'm going to use this and your PR on to make sure that all use cases are accounted for 💪  I should have commented on your PR to let you know all of this but I'm traveling from Amsterdam to Chicago so didn't get as much done before my flight wifi cut me off 😇

I'll be getting back to your PR soon!

joshdholtz:
> Yeeeeees 🚀
> 
> It's been awhile since I last touched dotenv in my projects, but I do remember having trouble with dotenv files, and not understanding why they're not autoloaded... When are they autoloaded and when are they not, and why? 🤔 We even have DotenvHelper and it seems to load and overload dotenv files for us, I just don't remember/know why they don't always work 😅
> 
> Anyway I've also been loading them manually, this is a great addition 🤗 just left some minor comments! 💪

@rogerluan Thanks for the review! So _fastlane_ will load `.env` and `.env.default` by default 😊 I like to store my secrets in a `.env.secret` and don't want to use the `--env` option all the time. So my main use case for this is to `dotenv(name: 'secrete'). But I'll also be calling this in the before for of my ios and macos platforms to load those specific things 🚀

rogerluan:
Gotcha, thanks for explaining @joshdholtz ! I remember now that the use case I needed in the past was loading other flavors too like `.env.somethingsomething`, not just `.env` 😄 

I agree with @shamilovtim 's points too, conceptually. Just not sure whether the assumptions made during the implementation are okay to be made. I shared more details here: https://github.com/fastlane/fastlane/pull/21629#pullrequestreview-1725938483

PR #21635, opened 2023-11-14
detect binding.irb in `ensure_no_debug_code` during build.

mataku:
### Checklist

### Motivation and Context

Since binding.irb is a debugging tool available by default in Ruby supported by the current fastlane, we can use the REPL environment easily by writing `binding.irb` without the need for `gem 'irb'` etc.

Therefore, how about adding binding.irb as a target for debugging code checks if you want to ensure that no debugging codes?

### Description
Add binding.irb by ensure_no_debug_code action

### Testing Steps

bundle exec fastlane test

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -27,7 +27,7 @@ lane :lint_source do
   # Verifying that no debug code is in the code base
   #
   exclude_dirs = ["\.bundle", "*/vendor/bundle"]
-  ensure_no_debug_code(text: "binding.pry", extension: ".rb", exclude: "playground.rb", exclude_dirs: exclude_dirs) # debugging code
+  ensure_no_debug_code(text: "binding\.(irb|pry)", extension: ".rb", exclude: "playground.rb", exclude_dirs: exclude_dirs) # debugging code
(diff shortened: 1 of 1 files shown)


lacostej (approved):
makes sense to me.

fastlane/Fastfile
```diff
@@ -28,6 +28,7 @@ lane :lint_source do
   #
   exclude_dirs = ["\.bundle", "*/vendor/bundle"]
   ensure_no_debug_code(text: "binding.pry", extension: ".rb", exclude: "playground.rb", exclude_dirs: exclude_dirs) # debugging code
```
lacostej, line 30: Now that I think about it, we could write

```
ensure_no_debug_code(text: "binding\.(irb|pry)", extension: ".rb", exclude: "playground.rb", exclude_dirs: exclude_dirs)
```
that way we will run a single call. No?
instead

mataku: @lacostej I thought it would be easier to understand if explicitly separated, but the context is the same and no problem to describe them together, so fixed

lacostej:
Merged. Thanks!

PR #21665, opened 2023-11-24
[action][spm] add `parallel` option

bramschulting:
~~I think something is wrong with my local setup, because if I run `bundle exec rspec` on a clean clone of the fastlane repo on the `master` branch, I still get some failures:~~ This is resolved. I was using ruby 3.2.2 and with that the tests fail. But I noticed the CI uses 3.1 so I tried again locally with 3.1.2 and now they pass.

### Checklist

### Motivation and Context

`swift test` allows parallel testing by passing the `--parallel` option. This reduces the time it takes to run the tests in a project. (In my case the time drops from 822 seconds to 133 seconds)

### Description

A `parallel` option has been added to the `SpmAction` action. If this options is set to `true` _and_ the `command` is `"test"`, the `--parallel` flag will be added to the command.

This functionality has been tested by using my [local fastlane setup](https://github.com/fastlane/fastlane/blob/master/Testing.md#test-your-local-fastlane-code-base-with-your-setup) in a real project. 

There is one small issue: when the tests are run in parallel, there seems to be no output to fastlane while the tests are running. When the tests complete (either successfully or as a failure), the output is shown. I don't know how to resolve this, so if anyone has an idea, feel free to let me know! But otherwise I think it's a tradeoff users of the `parallel` flag can make. I suspect the issue is that when running the tests with the `--parallel` option, output is not 'appended', but there's only one line that's constantly being removed and rewritten. You can see this happening when running `swift test --parallel` directly.

### Testing Steps

Create a lane that uses the `spm(command: 'test', parallel: true)` action.

### DIFF
--- a/fastlane/lib/fastlane/actions/spm.rb
+++ b/fastlane/lib/fastlane/actions/spm.rb
@@ -12,6 +12,7 @@ module Fastlane
         cmd << "--verbose" if params[:verbose]
         cmd << params[:command] if package_commands.include?(params[:command])
         cmd << "--enable-code-coverage" if params[:enable_code_coverage] && (params[:command] == 'generate-xcodeproj' || params[:command] == 'test')
+        cmd << "--parallel" if params[:parallel] && params[:command] == 'test'
         if params[:xcconfig]
           cmd << "--xcconfig-overrides #{params[:xcconfig]}"
         end
@@ -48,6 +49,11 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :enable_code_coverage,
                                        env_name: "FL_SPM_ENABLE_CODE_COVERAGE",
                                        description: "Enables code coverage for the generated Xcode project when using the 'generate-xcodeproj' and the 'test' command",
+                                       type: Boolean,
+                                       optional: true),
+          FastlaneCore::ConfigItem.new(key: :parallel,
+                                       env_name: "FL_SPM_PARALLEL",
+                                       description: "Enables running tests in parallel when using the 'test' command",
                                        type: Boolean,
                                        optional: true),
           FastlaneCore::ConfigItem.new(key: :build_path,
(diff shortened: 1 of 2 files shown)


lacostej:
Support for ruby 3.2 was just added in CI. It should be resolved now.

Thanks for the contribution!

lacostej (approved):
Looks good to me

bramschulting:
@lacostej Happy to contribute after using _fastlane_ for a long time! Is there anything left for me to do to get this merged?

lacostej:
@bramschulting one last idea: maybe we could add 
```
spm(command: 'test', parallel: true)
``` 

in the example section.

bramschulting:
> @bramschulting one last idea: maybe we could add
> 
> ```
> spm(command: 'test', parallel: true)
> ```
> 
> in the example section.

Good one, done!

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -50,6 +51,11 @@ def self.available_options
                                        description: "Enables code coverage for the generated Xcode project when using the 'generate-xcodeproj' and the 'test' command",
                                        type: Boolean,
                                        optional: true),
+          FastlaneCore::ConfigItem.new(key: :parallel,
+                                       env_name: "FL_SPM_PARALLEL",
+                                       description: "Enables running tests in parallel when using the 'test' command",
+                                       type: Boolean,
+                                       optional: true),
```
rogerluan, line 58: Since this is a boolean, I'd suggest making it non-optional and defaulting it to false (to keep the previous behavior; in favor of backwards compatibility):

```suggestion
                                       type: Boolean,
                                       default_value: false),
```

rogerluan (changes requested):
Perrrfect test cases!! Thank you! ❤️ 

I just made a minor suggestion regarding the action options 😊 I'll go ahead and apply them, and if the tests pass, this should be good to go!

Great contribution! 🙌

rogerluan (approved):
🚀 

(Just waiting for tests to pass)

bramschulting:
Thanks @lacostej and @rogerluan!

PR #21668, opened 2023-11-24
[tests][ci] clean up tests output on CI

lacostej:
### Checklist

### Description

Still trying to get the test output to only display . or X in minified mode.

Remains 
* plenty of xcodebuild errors like this one
```
2023-11-24 19:18:08.713 xcodebuild[3270:11980] Requested but did not find extension point with identifier Xcode.IDEKit.ExtensionPointIdentifierToBundleIdentifier for extension Xcode.DebuggerFoundation.AppExtensionToBundleIdentifierMap.watchOS of plug-in com.apple.dt.IDEWatchSupportCore
```

### DIFF
--- a/fastlane/lib/fastlane/actions/verify_build.rb
+++ b/fastlane/lib/fastlane/actions/verify_build.rb
@@ -67,7 +67,7 @@ module Fastlane
 
       def self.update_with_profile_info(app_path, values)
         provision_profile_path = "#{app_path}/embedded.mobileprovision"
-        UI.user_error!("Unable to extract profile") unless File.exist?(provision_profile_path)
+        UI.user_error!("Unable to find embedded profile in #{provision_profile_path}") unless File.exist?(provision_profile_path)
 
         profile = `cat #{provision_profile_path.shellescape} | security cms -D`
         UI.user_error!("Unable to extract profile") unless $? == 0
--- a/fastlane/lib/fastlane/actions/verify_build.rb
+++ b/fastlane/lib/fastlane/actions/verify_build.rb
@@ -24,7 +24,7 @@ module Fastlane
 
         case File.extname(build_path)
         when ".ipa", ".zip"
-          `unzip #{build_path.shellescape} -d #{dir.shellescape} -x '__MACOSX/*' '*.DS_Store'`
+          `unzip #{build_path.shellescape} -d #{dir.shellescape} -x '__MACOSX/*' '*.DS_Store' 2>&1`
           UI.user_error!("Unable to unzip ipa") unless $? == 0
           # Adding extra ** for edge-case ipas where Payload directory is nested.
           app_path = Dir["#{dir}/**/Payload/*.app"].first
--- a/fastlane/lib/fastlane/actions/verify_build.rb
+++ b/fastlane/lib/fastlane/actions/verify_build.rb
@@ -66,7 +66,10 @@ module Fastlane
       end
 
       def self.update_with_profile_info(app_path, values)
-        profile = `cat #{app_path.shellescape}/embedded.mobileprovision | security cms -D`
+        provision_profile_path = "#{app_path}/embedded.mobileprovision"
+        UI.user_error!("Unable to extract profile") unless File.exist?(provision_profile_path)
+
+        profile = `cat #{provision_profile_path.shellescape} | security cms -D`
         UI.user_error!("Unable to extract profile") unless $? == 0
 
         plist = Plist.parse_xml(profile)
--- a/fastlane/spec/cli_tools_distributor_spec.rb
+++ b/fastlane/spec/cli_tools_distributor_spec.rb
@@ -33,6 +33,11 @@ describe Fastlane::CLIToolsDistributor do
   end
 
   describe "update checking" do
+    before(:each) do
+      # prevent the update checker to run and clutter the output
+      ENV["FASTLANE_SKIP_UPDATE_CHECK"] = "a_value"
+    end
+
     it "checks for updates when running a lane" do
       FastlaneSpec::Env.with_ARGV(["sigh"]) do
         require 'fastlane/commands_generator'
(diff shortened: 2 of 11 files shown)


lacostej:
Ready for review

Almost all errors are hidden on CI. Remains one issue, but it comes from [Security::InternetPassword](https://github.com/fastlane-community/security/issues/11).

image

fastlane/lib/fastlane/actions/verify_build.rb
```diff
@@ -66,7 +66,10 @@ def self.gather_cert_info(app_path)
       end
 
       def self.update_with_profile_info(app_path, values)
-        profile = `cat #{app_path.shellescape}/embedded.mobileprovision | security cms -D`
+        provision_profile_path = "#{app_path}/embedded.mobileprovision"
+        UI.user_error!("Unable to extract profile") unless File.exist?(provision_profile_path)
```
AliSoftware, line 70:
```suggestion
        UI.user_error!("Unable to find embedded profile in #{provision_profile_path}") unless File.exist?(provision_profile_path)
```

fastlane/spec/cli_tools_distributor_spec.rb
```diff
@@ -33,6 +33,11 @@
   end
 
   describe "update checking" do
+    before(:each) do
+      # prevent the update checker to run and clutter the output
+      ENV["FASTLANE_SKIP_UPDATE_CHECK"] = "a_value"
```
AliSoftware, line 38:
```suggestion
      ENV["FASTLANE_SKIP_UPDATE_CHECK"] = "true"
```

fastlane_core/lib/fastlane_core/helper.rb
```diff
@@ -96,6 +96,11 @@ def self.windows?
       (/cygwin|mswin|mingw|bccwin|wince|emx/ =~ RUBY_PLATFORM) != nil
     end
 
+    def self.command?(name)
+      `which #{name} 2> /dev/null`
```
AliSoftware, line 100: I think `command -v #{name.shell_escape} 2>/dev/null` is more portable / recommended to check if a command exists rather than `which`? (I don't remember where I read that, but I did 😄 )

fastlane_core/lib/fastlane_core/project.rb
```diff
@@ -418,6 +418,8 @@ def build_settings(key: nil, optional: true)
 
         command = build_xcodebuild_showbuildsettings_command
 
+        command = "#{command} 2>&1" # xcodebuild produces errors on stderr #21672
+
```
AliSoftware, line 422: I wonder if it wouldn't make more sense to bake this `2>&1` within the implementation of `build_xcodebuild_showbuildsettings_command` instead? 🤔

fastlane/spec/cli_tools_distributor_spec.rb
lacostej: this was on purpose, to show that the value really doesn't matter. WDYT?

fastlane/lib/fastlane/actions/verify_build.rb
lacostej: I guess I will have to readjust the tests for that.

fastlane/spec/cli_tools_distributor_spec.rb
AliSoftware: Ah, gotcha. I didn't realize that the test and spec you introduced that change on was the one relevant for testing this.

But then I'm confused: that means that before your change, the skip env var was not set, so I'm guessing that means that all the test examples under `describe "update checking"` were actually running the fastlane update check… but if you set the env var value now, won't that disable the fastlane update check that those tests were supposed to run in order to test the `update checking` family of tests?

Or are you saying that a value of `"a_value"` is supposed to be the same as `"0"` or `"false"`, not `"true"`?

I haven't checked how the value of this env var is checked in the implementation to be honest; I guess it depends if it's checked with our `is_truthy?` helper—which only consider values like `"1"`, `"YES"`, `"true"`… as true—or if it's just checked like `ENV["…"].nil? || ENV["…"].empty?`—in which case any non-empty value would be considered as `true`, including `"a_value"` 🤔 💭

AliSoftware: Ok as soon as I wrote this comment I did a quick grep and saw that `FASTLANE_SKIP_UPDATE_CHECK` was always checked using `FastlaneCore::Env.truthy?` after all. (except in `spec_helper.rb:6`… but that's a test anyway)

So that means that `"a_value"` would be equivalent to `false`, not `true` as I initially thought.

Better make the intention more clear with a proper dummy value still:

```suggestion
      ENV["FASTLANE_SKIP_UPDATE_CHECK"] = "a_falsy_value"
```

fastlane_core/lib/fastlane_core/helper.rb
lacostej: I wonder if that will work in WIndows. I can give it a try.

lacostej: done. We'll see what happens.

rogerluan (approved):
LGTM! Great changes to improve a bit of QoL for us 😁

Thank you!

fastlane/spec/cli_tools_distributor_spec.rb
```diff
@@ -33,6 +33,12 @@
   end
 
   describe "update checking" do
+    before(:each) do
+      # prevent the update checker to run and clutter the output
+      # (FastlaneCore::UpdateChecker.start_looking_for_update() will return)
+      ENV["FASTLANE_SKIP_UPDATE_CHECK"] = "a_truthy_value"
```
AliSoftware, line 39: I'm wondering if this doesn't risk modifying the `ENV` of the user that is running the test suite?
Maybe better to `allow(ENV).to receive(:[]).with('FASTLANE_SKIP_UPDATE_CHECK').and_return('a_truthy_value')` instead?

AliSoftware (approved):

fastlane/spec/cli_tools_distributor_spec.rb
lacostej: Thanks. That was indeed helpful. I've added 8840cf664 which restores the context by using our `ClimateControl` wrapper. Using allow is harder as you have to take into account whether ENV is accessed many times.

PR #21669, opened 2023-11-27
[sigh] add `keychain_path` option

igorgn:
### Checklist

### Motivation and Context

Sometimes (right now for us) there's a need to use 'non-default' keychain for downloading apple provisioning profiles. 
This PR adds a possibility to pass it using '**--keychain_path**' or '**-k**' flag.

### Description

The code in _sigh_ action was ready for passing external flag 'keychain_path', but there's was no way to actually pass it in CLI. 

### Testing Steps
Tested in CI and local environment.


lacostej (approved):
looks good to me

sigh/lib/sigh/options.rb
```diff
@@ -114,6 +114,16 @@ def self.available_options
                                      env_name: "SIGH_PROVISIONING_PROFILE_NAME",
                                      description: "The name of the profile that is used on the Apple Developer Portal",
                                      optional: true),
+        FastlaneCore::ConfigItem.new(key: :keychain_path,
+                                     short_option: "-k",
+                                     env_name: "SIGH_KEYCHAIN_PATH",
+                                     description: "Path to a custom keychain",
```
rogerluan, line 120: Can this be a bit more specific, perhaps to indicate when the custom keychain would be used, or what for? It may be clear to advanced users, but maybe not the average user, not sure 😅

rogerluan (changes requested):
Thanks for this contribution @igorgn! 🙏 

I have a few questions/suggestions though! From an implementation perspective, this PR seems solid. However, I lack some more context/knowledge here, perhaps you can help me?

1. Is passing just the path enough? What about the keychain password, if one exists?
2. [match](https://docs.fastlane.tools/actions/match/) has a similar approach of allowing custom keychains, but instead it takes a "keychain name" and password, would this approach make sense to be replicated in _sigh_? (Instead of a keychain path)

Last but not least, could you help us by adding unit tests to this new feature you're adding? 🙇 

Thanks in advance for the clarifications! 🙏  Looking forward to getting this merged 😊

igorgn:
Hey @rogerluan , thank you for your review. 
Those are indeed nice points, so regarding
1. I indeed didn't take into consideration passing of password, but now thinking about it, looks like in most cases keychain should already be unlocked before executing _sigh_ action. I assume this is by design and additionally implementing _unlocking of keychain_ should probably be a separate feature.
2. Honestly I didn't had a chance to use _match_ as we're using _sigh_ for years now without any issues.  I think this is due to our product being built solely in CI environment and because of no real issues with certificates and profiles _match_ was not considered. Just now we're starting migration to a more strict environment.

AStrickland-proteor:
@rogerluan or @igorgn 
Is it possible to review this again? It would make CI processes much nicer if we could get this in.

Issue #21672, opened 2023-11-29
xcodebuild -showBuildSettings produces weird errors on stderr


lacostej:
xcodebuild produces weird errors on stderr.

This isn't a fastlane issue per se, and maybe just a project configuration issue / xcode install. It's main consequence as I can see is that it clutters the test output on CI. We should aim to get rid of it or hide it.

I haven't been able to find the reason behind these errors, but they don't cause the build to fail.

### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

with xcode 14.2 we see errors like
```
2023-11-29 08:43:59.948 xcodebuild[986:78741081] Requested but did not find extension point with identifier Xcode.InterfaceBuilderBuildSupport.PlatformDefinition
```

with xcode 14.3 we see a second error 

```
2023-11-29 08:43:59.637 xcodebuild[986:78741056] DVTCoreDeviceEnabledState: DVTCoreDeviceEnabledState_Disabled set via user default (DVTEnableCoreDevice=disabled)
2023-11-29 08:43:59.948 xcodebuild[986:78741081] Requested but did not find extension point with identifier Xcode.InterfaceBuilderBuildSupport.PlatformDefinition
```

##### Command executed

To reproduce do:
```
xcodebuild -showBuildSettings -scheme Example -project ./gym/examples/multiplatform/Example.xcodeproj -destination generic/platform\=iOS > log
```

PR #21673, opened 2023-11-29
Hide xcodebuild stderr to solve #21672

lacostej:
See #21672 

### Checklist

### Description

I haven't found a way to get rid of those issues, so hiding them by merging them with stdout. It doesn't affect the parsing of the settings as no stderr line contains a `=`.

### DIFF
--- a/fastlane_core/lib/fastlane_core/project.rb
+++ b/fastlane_core/lib/fastlane_core/project.rb
@@ -418,6 +418,8 @@ module FastlaneCore
 
         command = build_xcodebuild_showbuildsettings_command
 
+        command = "#{command} 2>&1" # xcodebuild produces errors on stderr #21672
+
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):
This hides **all** of stderr logs though, without an option to see them. And iinm, while we `tee` the stdout to a file especially to allow people to still browse the full log even when they pipe thru `xcpretty` / `xcbeautify`, we don't have an equivalent for being able to browse `stderr` messages optionally if they're masked in the streamed output.

---

[EDIT] I just realized that that patch is only applied to the command that generates `-showBuildSettings`, so I think that's ok; as I don't expect much _valuable_ info from stderr on that specific command. Just wanted to make sure we don't swallow `stderr` on the regular `xcodebuild` build commands like we'd do for `showBuildSettings`, but that seems to be good here 👍

PR #21677, opened 2023-11-29
[scan] Filter simulators with version greater than SDK version of active Xcode installation when choosing default

wuaar1003:
### Checklist

### Motivation and Context

Xcode 15 introduced a new Simulator subsystem in macOS that is shared across all Xcode versions installed on a machine. Previously, each Xcode had its own separate system service and set of devices for which it could run. With Xcode 15 installed, each Xcode now has access to many of the devices and runtimes for each Xcode version installed on a machine.

This is an issue for fastlane, as it currently picks the latest simulator version for a given device, unless the user explicitly specifies otherwise.

For example, suppose an environment has installations of both Xcode14 and Xcode15, with Xcode14 set as active. For the device `'iPhone 14'`, scan will select iOS simulator 17.0 instead of 16.4. One must use `'iPhone 14 (16.4)'` to run on the default Xcode14 simulator.

Under the proposed changes, `'iPhone 14'` would result in selection of the expected 16.4 simulator. If users wish to run on an older or newer simulator version, the behavior is as before i.e. `'iPhone14 (16.1)'` for the older 16.1 simulator or `'iPhone 14 (17.0)'` for the newer 17.0 simulator.

### Description
The changes affect `Scan::DetectValues`. A function named `detect_sdk_version` is added to determine default simulator SDK versions of the active Xcode installation. We use this information as an upper bound when filtering the set of available simulators if users do not specify a version. The new filtering heuristic is contained within `compatible_with_sdk`.

### Testing Steps
Updated unit tests to exercise new functionality. Tested in local envrionment and CI.


scan/lib/scan/detect_values.rb
```diff
@@ -137,26 +172,23 @@ def self.detect_simulator(devices, requested_os_type, deployment_target_key, def
...
-
           display_device = "'#{device_string}'"
 
           set + (
             if pieces.count == 0
               [] # empty array
             elsif pieces.count == 1
               simulators
-                .select(&selector)
+                .select { |sim| sim.name == pieces.first && compatible_with_sdk(sim) }
```
lacostej, line 188: would it make sense to introduce a `highest_compatible_simulator(name)` method and reuse it both here and line 220? It would do a .last instead of pop(1) here

We could reuse it like 

```
[ highest_compatible_simulator(pieces.first) ]
```

here and 

```
[ highest_compatible_simulator(default_device_name) || simulators.first ] 
```
below.

lacostej (commented):
Thanks for the contribution. A couple of comments.

Related to that feature, I see that we have lots of weird errors when setting up our simulators environments in CI and wonder if we could use some similar logic to avoid them (and speed up the setup) and thus the builds.

image

I believe this happens [here](https://github.com/fastlane/fastlane/blob/master/snapshot/lib/snapshot/reset_simulators.rb)

wuaar1003:
@lacostej addressed the later two comments. 

>  I see that we have lots of weird errors when setting up our simulators environments in CI and wonder if we could use some similar logic to avoid them

As far as I can tell, the logic would be different and would require more substantial changes to that file. For that particular error, what seems to be going on is `xcrun simctl list devicetypes` includes devices not supported by any installed runtime. On the other hand, `xcrun simctl list runtimes --json` includes a list of `supportedDeviceTypes` for each runtime.

scan/lib/scan/detect_values.rb
```diff
@@ -100,6 +108,41 @@ def self.filter_simulators(simulators, operator = :greater_than_or_equal, deploy
...
+         version = /#{Regexp.quote(simulator_name)}(?<val>.*)\.sdk/.match(sdk_path.basename.to_s)
+         UI.crash!("Could not determine SDK version from #{sdk_path}") unless version && version[:val] != ''
+
+         begin
+           Gem::Version.new(version[:val])
+         rescue ArgumentError => e
+           UI.crash!("Could not parse SDK version: #{e.message}")
+         end
+       end
+    end
```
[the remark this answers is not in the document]
wuaar1003: Sounds good, certainly feels more robust to use CLI output rather than relying on directory structure. Will work on a revision to use this approach instead. Thanks!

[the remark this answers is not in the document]
wuaar1003: Thanks for the feedback and the sample code! I've rewritten the implementation to invoke the suggested commands.

scan/lib/scan/detect_values.rb
```diff
@@ -100,6 +109,63 @@ def self.filter_simulators(simulators, operator = :greater_than_or_equal, deploy
...
+
+         # list SDK version for currently running Xcode
+         sdks_output, = Open3.capture3('xcodebuild -showsdks -json')
+         sdk_version = begin
+           JSON.parse(sdks_output).find { |e| e['platform'] == platform[:simulator] }['sdkVersion']
+         rescue StandardError => e
+           UI.error(e)
+           UI.error("xcodebuild CLI broken, please run `xcodebuild` and make sure it works")
+           UI.user_error!("xcodebuild not working")
+         end
```
AliSoftware, line 126: Could be nice to capture the status as well, and check if `status.success?` in case the `xcodebuild` command exited with non-zero status code.

```suggestion
         sdk_version = begin
           sdks_output, _, status = Open3.capture3('xcodebuild -showsdks -json')
           raise status unless status.success?
           JSON.parse(sdks_output).find { |e| e['platform'] == platform[:simulator] }['sdkVersion']
         rescue StandardError => e
           UI.error(e)
           UI.error("xcodebuild CLI broken, please run `xcodebuild` and make sure it works")
           UI.user_error!("xcodebuild not working")
         end
```

scan/lib/scan/detect_values.rb
```diff
@@ -100,6 +109,63 @@ def self.filter_simulators(simulators, operator = :greater_than_or_equal, deploy
...
+
+         # Get runtime build from SDK version
+         runtime_output, = Open3.capture3('xcrun simctl runtime match list -j')
+         runtime_build = begin
+           JSON.parse(runtime_output).values.find { |elem| elem['platform'] == platform[:name] && elem['sdkVersion'] == sdk_version }['chosenRuntimeBuild']
+         rescue StandardError => e
```
AliSoftware, line 132: Same idea here
```suggestion
         runtime_build = begin
           runtime_output, _, status = Open3.capture3('xcrun simctl runtime match list -j')
           raise status unless status.success?
           JSON.parse(runtime_output).values.find { |elem| elem['platform'] == platform[:name] && elem['sdkVersion'] == sdk_version }['chosenRuntimeBuild']
         rescue StandardError => e
```

AliSoftware: This one is especially important because I think `match list` is only available starting Xcode 15, so your `simctl` wont have it if you don't have Xcodde 15 installed on your machine (and thus the latest `simctl` that comes with it installed in your system).

Sure, you check for that in `self.compatibility_constraint` before calling `default_os_version`, but `default_os_version` is a public method, so nothing ensures that this method won't be called by another method in the future (like someone creating a PR months from now to add some feature to fastlane and ending up using it without realizing it'd only work with Xcode 15).

I would instead recommend to move the check for the existence of the `match list` command from `self.compatibility_constraint` to this method instead—making it return `nil` if `match list` command does not exist—so that whoever calls it (even future PRs) gets the check.

Then in `self.compatibility_constaint`, always call `default = default_os_version(sim.os_type)` and decide which `->(name, sim)` lambda to use based on if `default` is `nil` or not. Or even simpler, now that `self.compatibility_constaint` method would be a bit trivial, just remove it and move its now-simplistic logic in `self.highest_compatible_simulator` directly, like this:
```ruby
    def self.highest_compatible_simulator(simulators, device_name)
      latest_os = default_os_version(sim.os_type)
      simulators
        .select { |sim| sim.name == name && (latest_os.nil? || Gem::Version.new(sim.os_version) <= latest_os) }
        .reverse
        .sort_by! { |sim| Gem::Version.new(sim.os_version) }
        .last
    end
```

scan/lib/scan/detect_values.rb
```diff
@@ -100,6 +109,66 @@ def self.filter_simulators(simulators, operator = :greater_than_or_equal, deploy
...
+      sim.name == device_name && (latest_os.nil? || Gem::Version.new(sim.os_version) <= latest_os)
+    end
+
+    def self.highest_compatible_simulator(simulators, device_name)
+      simulators
+        .select { |sim| compatibility_constraint(sim, device_name) }
+        .reverse
+        .sort_by! { |sim| Gem::Version.new(sim.os_version) }
+        .last
+    end
```
wuaar1003, line 170: I tried inlining `self.compatibility_constraint`, but it wound up looking like
```
def self.highest_compatible_simulator(simulators, device_name)
  simulators
    .select { |sim| sim.name = device_name && (default_os_version(sim.os_type).nil? || Gem::Version.new(sim.os_version) <= default_os_version(sim.os_type)) }
    .reverse
    .sort_by! { |sim| Gem::Version.new(sim.os_version) }
    .last
end 
```
I found this a bit difficult to read, so I ended up keeping the constraint as a separate function.

fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -7,13 +7,25 @@
...
+        begin
+          raise status unless status.success?
+          json = JSON.parse(output)
+          @runtime_build_os_versions = json['runtimes'].map { |h| [h['buildversion'], h['version']] }.to_h
+        rescue StandardError => e
+          UI.error(e)
+          UI.error('xcrun simctl CLI broken; cun `xcrun simctl list runtimes` and make sure it works')
+          UI.user_error!('xcrun simctl not working')
+        end
+
```
AliSoftware, line 28: Now that everything is cleaned up I'm wondering if here is the best place to fill this hash, especially since it's not used by the DeviceManager class itself. So that means that having that filled only in `simulators(…)` means it's dependent on the caller / user of DeviceManager to call things in a specific order.

What do you think about moving this initialization code to a memoized getter instead?

```ruby
def runtime_build_os_version
  @runtime_build_os_version ||= begin
    … logic from these lines above
  end
end
```

AliSoftware (commented):
Logic looks good now, except maybe the place in the code where `@runtime_build_os_version` is populated depending on the caller calling things in order.

fastlane_core/lib/fastlane_core/device_manager.rb
wuaar1003: This also simplified many unit tests :)

fastlane_core/lib/fastlane_core/device_manager.rb
```diff
@@ -11,6 +11,19 @@ def all(requested_os_type = "")
...
+        @runtime_build_os_versions ||= begin
+            output, status = Open3.capture2('xcrun simctl list runtimes -j')
+            raise status unless status.success?
+            json = JSON.parse(output)
+            json['runtimes'].map { |h| [h['buildversion'], h['version']] }.to_h
+          rescue StandardError => e
+            UI.error(e)
+            UI.error('xcrun simctl CLI broken; cun `xcrun simctl list runtimes` and make sure it works')
+            UI.user_error!('xcrun simctl not working')
+          end
```
AliSoftware, line 24: Nitpick (that our rubocop should warn about): inconsistent indentation there, should be two spaces less indented

AliSoftware (approved):

fastlane_core/lib/fastlane_core/device_manager.rb
wuaar1003: Odd, my Rubocop didn't pick up on this. Went through the changes again to fix indentation. Should be good to go now.

lacostej:
This change seems to have added the following message in output on some platforms.

```nil versions are discouraged and will be deprecated in Rubygems 4```

See https://app.circleci.com/pipelines/github/fastlane/fastlane/5943/workflows/b4bb5777-bafd-4817-bfc1-0087dc024778/jobs/94514

Probably caused by `Gem::Version.new()`

PR #21691, opened 2023-12-04
[match][sigh] add option to automatically renew expired certificates (defaults to enabled)

nekrich:
### Checklist

### Motivation and Context
Motivation: Automation 😅, and we had one of our certs expired.

Current flow on expired certificate:
1. Nuke or remove certs and provisioning profiles manually from the storage. Which is scary.
2. Remove all provisioning profiles from the portal.

New flow: do nothing, match/sync_code_signing will do all the things for you 🎉.

### Description

Check the certificate expiration date for renewable certificates. Remove expired certs from storage.
This behavior is optional, and there is a new parameter `renew_expired_certs` to turn it on.

#### Extra profile validation
Provisioning profiles are not invalidated automatically on the dev portal when the certificate expires. They become Invalid only when opened directly in the portal 🤷. We need to do an extra check on the expiration date to ensure the profile is valid.

#### Git commit of removed files
Commit deletion of only listed deleted files. Encryption modifies all files in git.

#### Profile installation
I moved the profile installation logic a little bit lower. Now, only valid profiles are installed 🧹.

The `working_directory` parameter was not used in the `fetch_certificate`, so I removed it.

### Testing Steps

Tests provided: run match with an expired certificate and active provisioning profiles.


match/lib/match/storage/git_storage.rb
```diff
@@ -135,9 +135,14 @@ def human_readable_description
...
 
       def delete_files(files_to_delete: [], custom_message: nil)
-        # No specific list given, e.g. this happens on `fastlane match nuke`
-        # We just want to run `git add -A` to commit everything
-        git_push(commands: ["git add -A"], commit_message: custom_message)
+        if files_to_delete.count > 0
+          commands = files_to_delete.map { |filename|  "git rm #{filename}" }
+          git_push(commands: commands, commit_message: custom_message)
+        else
+          # No specific list given, e.g. this happens on `fastlane match nuke`
```
nekrich, line 142: Self-review:
idk if I should leave the old logic. All other storages expect a non-empty array of files to delete.

spaceship/lib/spaceship/portal/provisioning_profile.rb
```diff
@@ -484,7 +484,14 @@ def certificate_valid?
       # @return (Bool) Is the current provisioning profile valid?
       #                To also verify the certificate call certificate_valid?
       def valid?
-        return status == 'Active'
+        # Provisioning profiles are not invalidated automatically on the dev portal when the certificate expires.
```
lacostej, line 487: Does this mean that this PR also fixes the fact that we were not detecting expired certificates properly before?

match/lib/match/runner.rb
```diff
@@ -277,6 +311,10 @@ def fetch_provisioning_profile(params: nil, certificate_id: nil, app_identifier:
         return nil
       end
 
+      if Helper.mac?
```
lacostej, line 314: does the fact that we move this after the previous block means that we stop installing expired profiles into the keychain?

match/lib/match/storage/git_storage.rb
```diff
@@ -135,9 +135,14 @@ def human_readable_description
       end
 
       def delete_files(files_to_delete: [], custom_message: nil)
-        # No specific list given, e.g. this happens on `fastlane match nuke`
-        # We just want to run `git add -A` to commit everything
-        git_push(commands: ["git add -A"], commit_message: custom_message)
+        if files_to_delete.count > 0
+          commands = files_to_delete.map { |filename|  "git rm #{filename}" }
```
lacostej, line 139: is there any risk of the filename needing escaping?

match/lib/match/options.rb
```diff
@@ -292,6 +292,11 @@ def self.available_options
                                      description: "Renew the provisioning profiles if the certificate count on the developer portal has changed. Works only for the 'development' provisioning profile type. Requires 'include_all_certificates' option to be 'true'",
                                      type: Boolean,
                                      default_value: false),
+        FastlaneCore::ConfigItem.new(key: :renew_expired_certs,
+                                     env_name: "MATCH_RENEW_EXPIRED_CERTS",
+                                     description: "Automatically renew expired certificates. Notice: to renew developer_id and developer_id_installer certificates you must login with Account Holder account",
```
lacostej, line 297: Do we have now the ability to check if we have the right permission level, e.g. with spaceship? I don't remember... How does the code handle the failure if we don't have the right permissions level? Just a failure down the line when doing the operations on the dev portal? Any adverse side effect?

match/spec/runner_spec.rb
```diff
@@ -223,6 +224,103 @@
           end.to raise_error("Your certificate 'E7P4EE896K.cer' is not valid, please check end date and renew it if necessary")
         end
 
+        it "renews an outdated certificate", requires_security: true do
```
lacostej, line 227: that's a huge test. I wonder how stable it will be. :)

I liked when you used GIVEN/WHEN/THEN in the other PR. If there was one method where I would add it, it's definitively this one.

match/lib/match/runner.rb
```diff
@@ -133,13 +136,47 @@ def update_optional_values_depending_on_storage_type(params)
       end
     end
 
-    def fetch_certificate(params: nil, working_directory: nil, specific_cert_type: nil)
```
lacostej, line 136: so working_directory was ignored here since 9d199278d3963413b916434b347cccc42c436b05. Nice finding/cleanup.

match/lib/match/storage/interface.rb
```diff
@@ -55,11 +55,14 @@ def save_changes!(files_to_commit: nil, files_to_delete: nil, custom_message: ni
...
         files_to_delete ||= []
+        files_to_delete -= files_to_commit # Make sure we are not removing added files.
+
+        if files_to_commit.count == 0 && files_to_delete.count == 0
+          UI.user_error!("Neither `files_to_commit` nor `files_to_delete` were provided to the `save_changes!` method call")
+        end
 
         Dir.chdir(File.expand_path(self.working_directory)) do
           if files_to_commit.count > 0 # everything that isn't `match nuke`
-            UI.user_error!("You can't provide both `files_to_delete` and `files_to_commit` right now") if files_to_delete.count > 0
```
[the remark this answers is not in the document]
lacostej: Looking at the code, it was added in 60fc2f5bbb and I think the `right now` just means that the method only handled one of the cases at the time, so it was more a protection for the caller than a limit in the overall implementation.

With your change, it's now supported by the implementation. So if there are no limits in the storage we should be good.

match/lib/match/storage/git_storage.rb
lacostej: not sure what you mean here.

lacostej (commented):
that's a nice feature. Seems to fix a few corner cases issues as well.

match/lib/match/runner.rb
```diff
@@ -133,13 +136,47 @@ def update_optional_values_depending_on_storage_type(params)
...
+      is_authenticated_with_login = Spaceship::ConnectAPI.token.nil?
+      is_cert_renewable_via_api = RENEWABLE_CERT_TYPES.include?(cert_type)
+      is_cert_renewable = is_authenticated_with_login || is_cert_renewable_via_api
+
+      # Validate existing certificate first.
+      if renew_expired_certs && is_cert_renewable && storage_has_certs
+        cert_path = select_cert_or_key(paths: certs)
+
+        unless Utils.is_cert_valid?(cert_path)
+          UI.important("Removing invalid certificate '#{File.basename(cert_path)}'")
```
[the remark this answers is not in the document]
nekrich: Jerome, I see your point.
The only destructive operation for me is deleting developer_id* certs. It will probably be a good thing to exclude them from auto-renewal.
Deleting Development and Distribution certs for AppStore are not a big problem because they can be regenerated anytime.

spaceship/lib/spaceship/portal/provisioning_profile.rb
nekrich: Yes 😅. I don't know why Apple does not revoke profiles or make them invalid.
But even in this case, we cannot sign an app since the certificate has expired.

The date is Dec 4, and the expiration date is Dec 1.
The only Expired cert is that I opened it to see why it's not expired, and it expired immediately after opening it.
image

match/lib/match/runner.rb
nekrich: Yes, I see no point in installing an expired profile, generating a new one, and installing a new one after.

match/spec/runner_spec.rb
nekrich: That's a tough one 😅. It's just an adapted copy-paste version from other tests. I'll try to make it look better.

match/lib/match/runner.rb
nekrich: Maybe [RuboCop::Cop::Lint::UnusedMethodArgument](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/UnusedMethodArgument) rule will help a little with such things?

match/lib/match/storage/git_storage.rb
nekrich: I took another look and decided that it would be better if git storage behaved like other storages and did not add unexpected additional changes to git if files_to_delete was empty.

Done in a474f7c5401de16000af7d2e8ed689d2ccd4e162.

match/lib/match/options.rb
nekrich: Check: Using Tunes, we can match username with member email and get roles. But I never saw an endpoint to find a role assigned to the App Store Connect API Key.

Handling: just failure. Find out that the certificate is expired, delete in temp local decrypted folder, fail to generate one, and never reach the `save_changes!` line.

And I never saw fastlane checking user roles before doing some fetches or creating a certificate 😅.

match/lib/match/runner.rb
nekrich: jfy, I did the same in my other PR: https://github.com/fastlane/fastlane/pull/21694/files#diff-8cd09d7ab3f4398e3160f603839ac2891e3cdf60e7f09930a1288ce33ab8fc1fL279

match/lib/match/runner.rb
lacostej: image
Yup. I will have to check if we need to refresh our rubocop rules.

match/lib/match/options.rb
lacostej: thanks for that explanation.

match/lib/match/runner.rb
```diff
@@ -133,13 +136,48 @@ def update_optional_values_depending_on_storage_type(params)
       end
     end
 
-    def fetch_certificate(params: nil, working_directory: nil, specific_cert_type: nil)
+    RENEWABLE_CERT_TYPES = [:mac_installer_distribution, :development, :distribution, :enterprise]
```
lacostej, line 139: maybe we should rename this to `RENEWABLE_CERT_TYPES_VIA_API`?

nekrich: Yeah, it's for sure will improve readability. Done in eb0db7b98482fa99ab595ebd6c3f7d0d9c8c390c.

nekrich:
Rebased and updated the test to include new parameters for the `certificates_exists` stub.

lacostej (approved):

match/lib/match/options.rb
rogerluan, line 297: I reviewed this and the implementation and it seems users will need to authenticate via username + password + 2FA if they wish to renew those 2 certificates. Perhaps we can make this more clear? Because via App Store Connect API Key the highest role is Admin only 😬 which is what most people are using nowadays probably, due to so many challenges around regular auth…

How about…

```suggestion
                                     description: "Automatically renew expired certificates. Note: to renew `developer_id` and `developer_id_installer` certificates you must log in with the Account Holder account by using username and password; App Store Connect API key doesn't work in this case",
```

match/lib/match/options.rb
```diff
@@ -292,6 +292,11 @@ def self.available_options
...
+                                     type: Boolean,
+                                     default_value: false),
```
rogerluan, line 299: I understand this was probably used to keep the previous behavior the same, but are there risks or any reason why someone would not want this feature enabled by default? I think it's also our responsibility (as _fastlane_ developers/contributors) to provide sensible defaults and this seems to be one of those cases :) this feature is too good to be left up for developers to discover it on their own and then have to add +1 LOC to enable it 😅

match/spec/spec_helper.rb
```diff
@@ -8,3 +8,87 @@ def before_each_match
...
+def default_provisioning_type
+  "appstore"
+end
+
+def default_git_url
+  "https://github.com/fastlane/fastlane/tree/master/certificates"
+end
+
+def default_username
+  "flapple@something.com"
```
rogerluan, line 54: I read "f apple" at first 😆

match/spec/spec_helper.rb
```diff
@@ -8,3 +8,87 @@ def before_each_match
   ENV["DELIVER_USER"] = "flapple@krausefx.com"
   ENV["DELIVER_PASSWORD"] = "so_secret"
 end
+
+def create_fake_storage(match_config:, repo_dir:)
```
rogerluan, line 12: Great set of utilities 💪

spaceship/lib/spaceship/connect_api/models/profile.rb
```diff
@@ -64,7 +64,14 @@ def self.type
       end
 
       def valid?
-        return profile_state == ProfileState::ACTIVE
+        # Provisioning profiles are not invalidated automatically on the dev portal when the certificate expires.
+        # They become Invalid only when opened directly in the portal 🤷.
```
rogerluan, line 68: Great doc 👍

spaceship/lib/spaceship/portal/provisioning_profile.rb
rogerluan: They didn't bother setting up like a cron job that goes and expire everything like once a day 😂

rogerluan (approved):
This is a great DX improvement over what we have currently 🎉  thanks for the hard work on this PR @nekrich ! 🙌 

Just left some suggestions and comments, nothing blocking 😊

rogerluan:
It seems this PR developed conflicts @nekrich, if you could resolve those and my comments above, this PR should be good to go then 💪

match/lib/match/options.rb
nekrich: Thanks, updated the description.

match/lib/match/options.rb
nekrich: Great suggestion. The default value is `true` now.

rogerluan:
Thanks for addressing those changes @nekrich 🙏 I see CI is failing now, could you take a look? 🙇

nekrich:
Yeah, I'm looking into it. It looks like default renew certs affect all the tests.

rogerluan (approved):
LGTM 💪  Thanks @nekrich ! 🚀 

@lacostej any final thoughts on the recent changes before we merge this in?

lacostej:
LGTM

PR #21707, opened 2023-12-09
[action][spm] add simulator flag for swift compiler

gharary:
### Checklist

### Motivation and Context

This PR adds extra flag to the `spm` action, to be able to pass the simulator which `swift build` is going to compile on that.

### Description

Currently, using `spm` action, will run the `swift` compiler on the first available simulator. Which is an issue if running on CI server, as some SPM packages only supports iOS platform and running the compiler on a MacOS platform would fail the build. This PR will add an extra flag to the `spm` action, so the user can pass the simulator on which the `swift` compiler should run on. 
The way it works is that, currently `swift` compiler doesn't support a simulator flag, therefore we are passing it as a `Xswiftc` flag. First specifying which SDK the compiler is going to use, by passing the `sdk` flag. Then using `xcrun` to locate tools within the Xcode environment which helps find the path to the specified SDK. Then adding the `target` flag which is used to set the target architecture for the Swift compiler and finally passing the target platform and architecture.
Now user can sets the `simulator` and/or `architercure` of the SDK that is going to be used for compiler.

### Testing Steps
Adding four unit tests to test the commands


lacostej (commented):
See comment

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -10,6 +10,20 @@ def self.run(params)
         cmd << "--configuration #{params[:configuration]}" if params[:configuration]
         cmd << "--disable-sandbox" if params[:disable_sandbox]
         cmd << "--verbose" if params[:verbose]
+        if params[:simulator]
+          # Check if the simulator syntax is correct using a regular expression
+          unless params[:simulator] =~ /^(iphone|ipad|macos)simulator(\d+\.\d+)?$/
```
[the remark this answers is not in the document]
gharary: But this `simulator` flag is optional, so if user doesn't provide it, it wouldn't make sense to check it at the config section?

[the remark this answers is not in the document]
lacostej: The `verify_block` is only ran for the given config items.

https://github.com/fastlane/fastlane/blob/76862ca1f76b5f24494dd65ad880420b08b1172c/fastlane_core/lib/fastlane_core/configuration/config_item.rb#L245-L251

We use it to check individual config items validity.

There are other optional options with verify_block in the config

https://github.com/fastlane/fastlane/blob/76862ca1f76b5f24494dd65ad880420b08b1172c/fastlane/lib/fastlane/actions/spm.rb#L65-L72

lacostej (approved):

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -93,7 +102,15 @@ def self.available_options
...
                                        type: Boolean,
-                                       default_value: false)
+                                       default_value: false),
+          FastlaneCore::ConfigItem.new(key: :simulator,
+                                       env_name: "FL_SPM_SIMULATOR",
+                                       description: "Specifies the simulator to pass for Swift Compiler",
+                                       type: String,
+                                       optional: true,
+                                       verify_block: proc do |value|
+                                         UI.user_error!("Please pass a valid simulator: (iphonesimulator|macossimulator)") unless valid_simulators.include?(value)
```
lacostej, line 112: nitpick: we could reuse the list here

```
UI.user_error!("Please pass a valid simulator: (#{self.valid_simulators.join('|')})") unless valid_simulators.include?(value)
```

Looks good otherwise.

gharary: Done

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -10,6 +10,15 @@ def self.run(params)
         cmd << "--configuration #{params[:configuration]}" if params[:configuration]
         cmd << "--disable-sandbox" if params[:disable_sandbox]
         cmd << "--verbose" if params[:verbose]
+        if params[:simulator]
+          simulator_flags = [
+            "-Xswiftc", "-sdk",
+            "-Xswiftc", "$(xcrun --sdk #{params[:simulator]} --show-sdk-path)",
+            "-Xswiftc", "-target",
+            "-Xswiftc", "x86_64-apple-ios$(xcrun --sdk #{params[:simulator]} --show-sdk-version | cut -d '.' -f 1)-simulator"
```
rogerluan, line 18: Won't this build only for x86 arch? What about arm64? 👀

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -93,7 +102,15 @@ def self.available_options
                                        env_name: "FL_SPM_VERBOSE",
                                        description: "Increase verbosity of informational output",
                                        type: Boolean,
-                                       default_value: false)
+                                       default_value: false),
+          FastlaneCore::ConfigItem.new(key: :simulator,
+                                       env_name: "FL_SPM_SIMULATOR",
+                                       description: "Specifies the simulator to pass for Swift Compiler",
```
rogerluan, line 108: We usually add the available options in the description so they are printed in the documentation and users know them before they need to hit a failure in `verify_block` step 😊 See e.g.:

https://github.com/fastlane/fastlane/blob/7da098f32aa12edd9ec0c97b0b106b7006708282/fastlane/lib/fastlane/actions/spm.rb#L41-L47

```suggestion
                                       description: "Specifies the simulator to pass for Swift Compiler (one of: #{valid_simulators.join(', ')})",
```

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -93,7 +102,15 @@ def self.available_options
...
+                                       type: String,
+                                       optional: true,
+                                       verify_block: proc do |value|
+                                         UI.user_error!("Please pass a valid simulator: (#{valid_simulators.join('|')})") unless valid_simulators.include?(value)
```
rogerluan, line 112: Nit: we could keep the messaging consistent here too (with e.g. the `:command` option of this action):

```suggestion
                                         UI.user_error!("Please pass a valid simulator. Use one of the following: #{valid_simulators.join(', ')}") unless valid_simulators.include?(value)
```

rogerluan (changes requested):
Congrats on your first PR in the _fastlane_ project @gharary!

I left a question and a couple suggestions. Aside from those, would you mind adding unit tests to this PR? `fastlane/spec/actions_specs/spm_spec.rb` contains extensive tests already, in case you need sample code to look at when implementing these tests 😊 Let me know if you need any help!

Looking forward to getting this PR merged; the lack of simulator flag is really annoying! Thanks for working on this! 💪

fastlane/lib/fastlane/actions/spm.rb
gharary: Good point, however I don't have anything in mind now. What do you suggest? building the package twice once for each simulator ? or adding an option that user can also define the arch?

rogerluan: I'd probably allow the user to decide their own adventure, but at the same time provide sensible defaults, which IMO should be arm64 at this point. Perhaps `x86_64`, `arm64` and `both` could be the options? Or maybe not `both` and if they want they can invoke the action twice? 

Open to other thoughts :)

fastlane/spec/actions_specs/spm_spec.rb
```diff
@@ -339,6 +339,43 @@
...
+
+        it "raises an error if simulator syntax is invalid" do
+          expect do
+            Fastlane::FastFile.new.parse("lane :test do
+              spm(
+                command: 'build',
+                simulator: 'invalid_simulator'
+              )
+            end").runner.execute(:test)
+          end.to raise_error("Invalid simulator syntax. Please use 'iphonesimulator', 'ipadsimulator', or 'macossimulator'.")
```
rogerluan, line 363: `ipadsimulator` is not part of `self.valid_simulators` 👀  should it be?

fastlane/lib/fastlane/actions/spm.rb
lacostej: instead of adding `both` make it that the target option is a list of comma separated values, and iterate over the `split(',')` of it?

rogerluan: I was leaning towards not allowing multiple archs to be built at once, but this sounds better ☝️

fastlane/spec/actions_specs/spm_spec.rb
gharary: Good catch. Done.

fastlane/spec/actions_specs/spm_spec.rb
```diff
@@ -360,20 +360,7 @@
...
-        it "raises an error" do
-          expect do
-            Fastlane::FastFile.new.parse("lane :test do
-              spm(
-                command: 'clean',
-                simulator: 'iphonesimulator'
-              )
-            end").runner.execute(:test)
-          end.to raise_error("Simulator is only applicable to package-related commands.")
+          end.to raise_error("Invalid simulator syntax. Please use 'iphonesimulator', or 'macossimulator'.")
```
rogerluan, line 363: tests are failing, perhaps this will fix it:

```suggestion
          end.to raise_error("Please pass a valid simulator. Use one of the following: iphonesimulator, macossimulator"))
```

gharary: Thanks for the help.

fastlane/spec/actions_specs/spm_spec.rb
```diff
@@ -339,6 +339,30 @@
...
+
+        it "raises an error if simulator syntax is invalid" do
+          expect do
+            Fastlane::FastFile.new.parse("lane :test do
+              spm(
+                command: 'build',
+                simulator: 'invalid_simulator'
+              )
+            end").runner.execute(:test)
+          end.to raise_error("Please pass a valid simulator. Use one of the following: iphonesimulator, macossimulator"))
```
rogerluan, line 363:
```suggestion
          end.to raise_error("Please pass a valid simulator. Use one of the following: iphonesimulator, macossimulator")
```

fastlane/lib/fastlane/actions/spm.rb
gharary: Good point guys. Thanks. I will try to implement this idea.

gharary: Hey guys, Please feel free to take a look for new changes adding support to simulator. I was taking a look at swift build script here [github](https://github.com/apple/swift/blob/main/utils/build-script-impl) and found that the currect architectures are `arm64`, `arm64e` and `x86_64` which the first one is only for `macosx` therefor I didn't add it and only the other two.
Also I wasn't sure how to manage the support of multiple archs at once, so for now I added support only one arch at a time.
I'm open to any suggestion/ comment/ changes.

fastlane/spec/actions_specs/spm_spec.rb
rogerluan, line 1: Since it looks like the macOS simulator command wasn't working (? can't really tell for sure ATM), is there any way we can add a unit test that makes sure not only the output command is the expected string, but that the command actually works?

rogerluan (changes requested):
Just a few more changes @gharary 😊 looking good!

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -158,7 +166,11 @@ def self.xcpretty_output_types
       end
 
       def self.valid_simulators
-        %w(iphonesimulator macossimulator)
+        %w(iphonesimulator macosx)
```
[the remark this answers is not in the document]
gharary: Yes it was my mystake that didn't properly testing the macossimulator in terminal and only the iphonesimulator. But now I also tested it, and it also reflects the host list in Swift's [repo](https://github.com/apple/swift/blob/main/utils/build-script-impl)
But to be able to write unit testi for it to be sure that the command is correct, I have no idea ATM.

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -15,7 +15,7 @@ def self.run(params)
             "-Xswiftc", "-sdk",
             "-Xswiftc", "$(xcrun --sdk #{params[:simulator]} --show-sdk-path)",
             "-Xswiftc", "-target",
-            "-Xswiftc", "x86_64-apple-ios$(xcrun --sdk #{params[:simulator]} --show-sdk-version | cut -d '.' -f 1)-simulator"
+            "-Xswiftc", "#{params[:simulator_arch] || "arm64"}-apple-#{params[:simulator] == "iphonesimulator" ? 'ios' : 'macosx'}$(xcrun --sdk #{params[:simulator]} --show-sdk-version | cut -d '.' -f 1)#{"-simulator" if params[:simulator] == "iphonesimulator"}"
```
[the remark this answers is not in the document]
gharary: Thanks for the tips, was helpful. I modified the code with these tips.

fastlane/spec/actions_specs/spm_spec.rb
gharary: I wasn't able to find a way to actually test the code in unit tests. I was thinking of using `open3` to  catch the result of executing command, but then testing the `swift build` command needs an actual SPM. So wasn't making sense to me. Any suggestion?

rogerluan: Uhm, if we wanted, we could create a sample SPM just for testing purposes. But, on a second thought, I think we actually might not need this, as long as we test it manually and make sure it works with all configurations, this should be fine. The reason is that we'd be validating whether "swift build" works, and that's not really the intent of what fastlane's unit tests should be :) 

So this is good as is 🙏  thank you!

fastlane/lib/fastlane/actions/spm.rb
```diff
@@ -11,6 +11,18 @@ def self.run(params)
...
+        if params[:simulator]
+          _simulator_platform = simulator_platform(simulator: params[:simulator], simulator_arch: params[:simulator_arch])
+          _simulator_sdk = simulator_sdk(simulator: params[:simulator])
+          _simulator_sdk_suffix = simulator_sdk_suffix(simulator: params[:simulator])
+          simulator_flags = [
+            "-Xswiftc", "-sdk",
+            "-Xswiftc", "$(xcrun --sdk #{params[:simulator]} --show-sdk-path)",
+            "-Xswiftc", "-target",
+            "-Xswiftc", "#{_simulator_platform}#{_simulator_sdk}#{_simulator_sdk_suffix}"
+          ]
```
rogerluan, line 23:
```suggestion
          simulator_platform = simulator_platform(simulator: params[:simulator], simulator_arch: params[:simulator_arch])
          simulator_sdk = simulator_sdk(simulator: params[:simulator])
          simulator_sdk_suffix = simulator_sdk_suffix(simulator: params[:simulator])
          simulator_flags = [
            "-Xswiftc", "-sdk",
            "-Xswiftc", "$(xcrun --sdk #{params[:simulator]} --show-sdk-path)",
            "-Xswiftc", "-target",
            "-Xswiftc", "#{simulator_platform}#{simulator_sdk}#{simulator_sdk_suffix}"
          ]
```

Shadowing should be okay in Ruby, and this is cleaner, did you observe any issues when attempting this format? Otherwise, I think it's preferred :)

rogerluan (commented):
Just two minor nitpicks left related to code style, otherwise this PR looks good to go @gbrhaz ! Thanks all the hard work here! 💪 🎉

gharary:
I did test the shadowing and it works fine and also tests are passing so I think all good.

rogerluan (approved):
🚀 awesome work! Thank you for the patience and for this contribution @gharary ! 🙏

PR #21716, opened 2023-12-12
[match] git storage: allow simultaneous usage of clone_branch_directly and shallow_clone

nekrich:
### Checklist

### Motivation and Context

I noticed a slow checkout of our match repo 😅.

#### How it worked before
You have to choose `shallow_clone` or `clone_branch_directly`. The first option clones all branches with depth 1, and another clones branch with full history.
Specifying both options (which I've done) led to the shallow cloning of all branches because shallow_clone option took precedence. 

#### How it works in this PR
1. Specifying options separately has no changes in current behavior.
2. Specifying options together leads to a more obvious result: direct cloning of a single branch without any history. Which is the fastest option available 🚀.

### Description

It was one if/elsif statement. I split it into to independent ones, allowing simultaneous usage of both options.
Tests are provided.

### Testing Steps

Run match specifying both options.

### DIFF
--- a/match/spec/storage/git_storage_spec.rb
+++ b/match/spec/storage/git_storage_spec.rb
@@ -1,5 +1,16 @@
+require_relative 'git_storage_spec_helper'
+
 describe Match do
   describe Match::Storage::GitStorage do
+
+    let(:git_url) { "https://github.com/fastlane/fastlane/tree/master/certificates" }
+    let(:git_branch) { "test" }
+
+    before(:each) do
+      @path = Dir.mktmpdir # to have access to the actual path
+      allow(Dir).to receive(:mktmpdir).and_return(@path)
+    end
+
     describe "#generate_commit_message" do
       it "works" do
         storage = Match::Storage::GitStorage.new(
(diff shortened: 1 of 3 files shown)


match/spec/storage/git_storage_spec.rb
```diff
@@ -43,6 +43,46 @@
         expect(File.exist?(File.join(storage.working_directory, 'README.md'))).to eq(false) # because the README is being added when committing the changes
       end
 
+      it "directly clones a single branch wtih shallow clone" do
+        # GIVEN
```
lacostej, line 47: what about being a bit more DRY and move those initializations in one place, either using a `before` or an `around` under the same `describe` ? I mention [around](https://github.com/fastlane/fastlane/blob/master/fastlane/spec/cli_tools_distributor_spec.rb#L4-L10) in case the garbage collector kicks in. 

```
describe "#download" do
  before(:each) do
    @path = Dir.mktmpdir # to have access to the actual path
    allow(Dir).to receive(:mktmpdir).and_return(@path)
  end
```

lacostej (approved):
Looks good.

A potential improvement in or test setup noted.

match/spec/storage/git_storage_spec.rb
```diff
@@ -43,6 +43,46 @@
         expect(File.exist?(File.join(storage.working_directory, 'README.md'))).to eq(false) # because the README is being added when committing the changes
       end
 
+      it "directly clones a single branch wtih shallow clone" do
```
rogerluan, line 46: Typo:

```suggestion
      it "directly clones a single branch with shallow clone" do
```

Despite the typo, this rspec structure is a bit weird to me. I'd normally expect something like

```ruby
describe "#download"
  # Insert some set up code like lacostej mentioned

  describe "when using shallow_clone and clone_branch_directly" do
    it "clones a single branch" do
      …
    end
  end
end
```

What do you think?

rogerluan (commented):
Thanks for another contribution @nekrich ! 🙏 

Just left a minor comment, let me know what you think 😊

match/spec/storage/git_storage_spec.rb
rogerluan: This is not critical though, the rest of the file is structured like this already so I guess it's at least consistent 😅 if you don't feel strongly about fixing this it's ok, we can simply fix the typo then

match/spec/storage/git_storage_spec.rb
nekrich: Refactored tests in 85c8aceea91f83bbe3f42d5eb955522307fc526c.
Combined your and @rogerluan comments.

Git spec helper:
1. branch checkout commands were repeated all over the spec.
2. Command execution check is too long, imo. Added a handful wrapper that accepts a single command string or array of commands.

Spec:
1. Default git url and branch.
2. Default temp dir stub.
3. Shorter test cases.

match/spec/storage/git_storage_spec.rb
```diff
@@ -12,286 +23,226 @@
...
-          }).and_return("")
-        end
-
-        storage = Match::Storage::GitStorage.new(
-          git_url: git_url,
-          shallow_clone: shallow_clone
-        )
-        storage.download
+      describe "when no branch is specified" do
+        it("checkouts the master branch ") do
```
rogerluan, line 27: Nit: Extra space, and it's generally a practice to avoid the parenthesis when declaring e.g. `describe`, `it`, etc.:

```suggestion
        it "checkouts the master branch" do
```

This should've been caught by rubocop but we're not enforcing this particular style yet 😅 could you change it throughout this file?

rogerluan (changes requested):
The new refactor looks great!! Just requesting a minor style change and this looks good to go 🤗

rogerluan (approved):
🚀

PR #21729, opened 2023-12-15
[scan] fix regressions in unit tests

### LINKED ISSUE #21728: Weird error in test output since #21677
The #21677 change seems to have added the following message in output on some platforms.

```nil versions are discouraged and will be deprecated in Rubygems 4```

See https://app.circleci.com/pipelines/github/fastlane/fastlane/5943/workflows/b4bb5777-bafd-4817-bfc1-0087dc024778/jobs/94514

Probably caused by `Gem::Version.new()`

### New Issue Checklist

### Issue Description
Regression in clean test output when running test.

### PR DESCRIPTION
**Pull request by wuaar1003**
### Checklist

### Motivation and Context

Resolves #21728 

### Description

Return version `0` instead of `nil` for correct memoization, which also resolves the issue of passing `nil` to `Gem::Version`

### Testing Steps

New unit test to exercise code path.

### CODE REVIEW

**Review by rogerluan — COMMENTED** [review]
@wuaar1003 thanks for this follow-up PR! 🙏 

It seems to me that, conceptually/semantically, a "nil" value would make more sense than a "Version of 0", no? Could we work around the issue observed while keep using nil? Of course, this would mean not attempting to instantiate a `Gem::Version.new(nil)` but instead guarding against that scenario from happening completely

**Comment by AliSoftware** [comment]
> @wuaar1003 thanks for this follow-up PR! 🙏
> 
> It seems to me that, conceptually/semantically, a "nil" value would make more sense than a "Version of 0", no? Could we work around the issue observed while keep using nil? Of course, this would mean not attempting to instantiate a `Gem::Version.new(nil)` but instead guarding against that scenario from happening completely

Was reviewing this PR and was about to say the same before seeing Roger's comment already mentioning it 😄  I agree that still returning `nil` would feel more idiomatic to Ruby's standards, i.e. `os_version.nil? ? nil : Gem::Version.new(os_version)` or `Gem::Version.new(os_version) unless os_version.nil?` (tip: in Ruby, that last form automatically evaluates to `nil` if the `if` condition is false)

**Comment by lacostej** [comment]
@wuaar1003 I've made a PR to address the code review. https://github.com/wuaar1003/fastlane/pull/1

**Comment by rogerluan** [comment]
Thanks @lacostej I've reviewed your PR there :)

**Comment by wuaar1003** [comment]
Thanks for the helpful feedback and PR. It turns out that I did experiment with using `nil` version instead of `0` -- see [this diff](https://github.com/fastlane/fastlane/compare/e10f528aa4f70337350f231e77e788560b110d48..60cd431949580037e904ebccd9d40462c7c135e7). 

The reason why tests are timing out is because caching does not capture `nil`:
```
    def self.default_os_version(os_type)
      @os_versions ||= {}
      @os_versions[os_type] ||= begin
       ...
          gem_version = Gem::Version.new(os_version) if os_version
        end
        gem_version
      end
    end
```
Thus, we will be doing output parsing every time for OS types which do not have a default. IIRC, this was my rationale for using version `0` as a stand-in.

Apologies for not staying on top of this. I don't foresee having the bandwidth to continue tracking this actively so will close  for now.

PR #21734, opened 2023-12-16
[ci] automate fastlane deployments to RubyGems when a new release is published

rogerluan:
### Motivation and Context

After the exciting news about [RubyGems announcing Trusted Publishing](https://blog.rubygems.org/2023/12/14/trusted-publishing.html), I'm thrilled to implement it in _fastlane_!

This means we won't have to do a manual deployment to RubyGems when there's a new release 🎉 

This takes us one HUGE step closer from implementing automated periodic releases! 🚀

### Description

I simply followed the sample code in https://blog.rubygems.org/2023/12/14/trusted-publishing.html but with the same hook used in `.github/workflows/announce_release.yml` (when there's a new "published release", deploy).

### Testing Steps

Not sure, actually. I guess we'll have to review, merge, enable Trusted Publishing in RubyGems itself, and then see if it works next time we release 😄

### DIFF
--- a/.github/workflows/deploy_to_rubygems.yml
+++ b/.github/workflows/deploy_to_rubygems.yml
@@ -1,7 +1,9 @@
+---
 name: Deploy to RubyGems
 on:
   release:
-    types: [published]
+    types:
+      - published
 
 jobs:
   deploy-to-rubygems:
@@ -11,12 +13,10 @@ jobs:
       contents: write
       id-token: write
     steps:
-      # Set up
       - uses: actions/checkout@v4
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-          ruby-version: "3.2"
-      # Release
+          ruby-version: '3.2'
       - uses: rubygems/release-gem@v1
--- a/.github/workflows/deploy_to_rubygems.yml
+++ b/.github/workflows/deploy_to_rubygems.yml
@@ -1,0 +1,22 @@
+name: Deploy to RubyGems
+on:
+  release:
+    types: [published]
+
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_to_rubygems.yml
```diff
@@ -0,0 +1,22 @@
...
+      contents: write
+      id-token: write
+    steps:
+      # Set up
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: "3.2"
```
lacostej, line 20: does it matter which version of ruby we use here?

lacostej (approved):
LGTM.

.github/workflows/deploy_to_rubygems.yml
rogerluan: I don't think so? At best, we should use the same version between this "validation" workflow and the actual workflow used for the "release" of the documentation itself, but I think we can change that only if we find issues with it? WDYT?

So I just picked one of the newest and most stable versions

rogerluan: I don't think so, so I just picked one of the newest and most stable versions

lacostej:
Let's merge this.

joshdholtz:
@rogerluan @lacostej Working on hooking this all up right now! This requires a few more changes to fully automate which is making a new `bump` GitHub Workflow that will create the bump PR that will include the autogenerated changelog and new `Fastlane.swift` files.

This Trusted Publishing is super neat and can’t wait for this to actually work!

PR #16033, opened 2020-02-17
[ci] update github actions

tenhishadow:
fix yaml;
add test with rake on macos/ubuntu with ruby matrix

### Checklist
- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.

### Motivation and Context
ugly yaml with wrong indentation

### Description
fix yaml;
add test with rake on macos/ubuntu with ruby matrix

### Testing Steps
yamllint; github actions

PR #21738, opened 2023-12-18
[ci] review and lint GitHub Actions workflow files

rogerluan:
### Checklist

### Motivation and Context

This PR supersedes https://github.com/fastlane/fastlane/pull/16033 — the part where it lints the GitHub Actions yaml files.

Closes https://github.com/fastlane/fastlane/pull/16033

### Description

I saw the PR https://github.com/fastlane/fastlane/pull/16033 whose branch got deleted, and thought that the part where the author was linting the yaml files was still valuable for us to add, so I decided to lint the existing files, review them all, adding comments where it's not clear, review their wording/workflow names/etc, and open this PR.

The part of that PR that adds a GHA to run rake on ubuntu/macOS isn't needed as we're already running tests on CircleCI.

### Testing Steps

No testing steps I think; we just need to make sure that the changes I'm making to the workflow identifiers and names won't affect anything, which I believe they won't. If we had GHA checks that would be run on pull requests, where the repo settings would require some of those check to be successful, then those are identified by name, and would be affected, but since we don't have any GHA running on PRs, then I think we're safe 😃

### DIFF
--- a/.github/workflows/tags.yml
+++ b/.github/workflows/tags.yml
@@ -1,5 +1,5 @@
 ---
-name: Deploying to Homebrew
+name: Deploy to Homebrew
 on:
   push:
     tags: 'v*' # Push events to tags matching v*
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: mislav/bump-homebrew-formula-action@v3.1
-        if: "!contains(github.ref, '-')" # skip prereleases
+        if: "!contains(github.ref, '-')" # Skip prereleases
         with:
           formula-name: fastlane
         env:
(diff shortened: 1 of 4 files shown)


lacostej (approved):

rogerluan (commented):
Self-review completed

.github/workflows/tags.yml
```diff
@@ -1,15 +1,17 @@
+---
+name: Deploy to Homebrew
 on:
   push:
-    tags: 'v*'
+    tags: 'v*' # Push events to tags matching v*
```
janbrasna, line 5: Honestly, there are no "tags matching `v*`" for v2.221.1 but the formula is still bumped FWIW;] the brew bumps are being done elsewhere:
https://github.com/fastlane/fastlane/blob/5aa842986cbaf5298623e340241623d73901f35f/fastlane/Fastfile#L435
so this may be completely superfluous…(?)

PR #21751, opened 2023-12-22
[ci] add GitHub Action workflow to label PRs based on tools they touch

rogerluan:
### Checklist

### Motivation and Context

We keep adding these labels manually, but only when a core contributor opens/reviews a PR. A handy tool that does this automatically for us, even before we review the PR, would be helpful!

### Description

This workflow will automatically label PRs with e.g. `tool: fastlane_core`, `type: ci`, `tool: deliver`, etc. based on the directories that are being modified in the PR.

I'm wondering if I forgot any obvious rule we should be adding? What are other files associated to the `type: ci` label? Are there other labels we could automate based on file patterns?

> [!IMPORTANT]
> This automation will only kick in when PRs are open/reopened or a new commit is pushed to it. In other words, it won't update existing open PRs. We'd need to do this in a one-off basis somehow. I intend to rebase all open PRs so that should help adding labels to all PRs that don't have merge conflicts currently, otherwise, we could merge a GHA workflow that will be trigger on cron job and then remove it once it's run once. Lmk your thoughts on this 😇 

### Testing Steps

Check if it adds and removes the labels as promised, in this PR. Then, switch `pull_request` trigger to `pull_request_target`, so it works with forked PRs. Read this (https://github.com/actions/labeler#notes-regarding-pull_request_target-event) to understand why this switching is needed to test.

#### Screenshots of tests observed in this PR:

Labels successfully added by the bot once everything was configured correctly:

image
image

Labels successfully removed when the code changed in those directories was removed:

image

Not sure if we should allow the bot to remove labels, see this comment of mine: https://github.com/fastlane/fastlane/pull/21751#discussion_r1435292843

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,59 +1,70 @@
 # Configuration file for https://github.com/actions/labeler
 
 "tool: cert":
-  - cert/**/*
-
-"tool: credentials_manager":
-  - credentials_manager/**/*
+  - changed-files:
+    - cert/*
 
 "tool: deliver":
-  - deliver/**/*
+  - changed-files:
+    - deliver/*
 
 "tool: fastlane_core":
-  - fastlane_core/**/*
+  - changed-files:
+    - fastlane_core/*
 
 "tool: frameit":
-  - frameit/**/*
+  - changed-files:
+    - frameit/*
 
 "tool: gym":
-  - gym/**/*
+  - changed-files:
+    - gym/*
 
 "tool: match":
-  - match/**/*
+  - changed-files:
+    - match/*
 
 "tool: pem":
-  - pem/**/*
+  - changed-files:
+    - pem/*
 
 "tool: pilot":
-  - pilot/**/*
+  - changed-files:
+    - pilot/*
 
 "tool: precheck":
-  - precheck/**/*
+  - changed-files:
+    - precheck/*
 
 "tool: produce":
-  - produce/**/*
+  - changed-files:
+    - produce/*
 
 "tool: scan":
-  - scan/**/*
+  - changed-files:
+    - scan/*
 
 "tool: screengrab":
-  - screengrab/**/*
+  - changed-files:
+    - screengrab/*
 
 "tool: sigh":
-  - sigh/**/*
+  - changed-files:
+    - sigh/*
 
 "tool: snapshot":
-  - snapshot/**/*
+  - changed-files:
+    - snapshot/*
 
 "tool: spaceship":
-  - spaceship/**/*
+  - changed-files:
+    - spaceship/*
 
 "tool: supply":
-  - supply/**/*
-
-"tool: trainer":
-  - trainer/**/*
+  - changed-files:
+    - supply/*
 
 "type: ci":
-  - .github/**
-  - .circleci/**
+  - changed-files:
+    - .github/*
+    - .circleci/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -53,3 +53,7 @@
 
 "tool: trainer":
   - trainer/**/*
+
+"type: ci":
+  - .github/**
+  - .circleci/**
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -100,3 +100,4 @@
     - any-glob-to-any-file:
       - .github/**/*
       - .circleci/**/*
+      - appveyor.yml
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -7,78 +7,78 @@
 
 "tool: deliver":
   - changed-files:
-      - any-glob-to-any-file:
-          - deliver/*
+    - any-glob-to-any-file:
+      - deliver/*
 
 "tool: fastlane_core":
   - changed-files:
-      - any-glob-to-any-file:
-          - fastlane_core/*
+    - any-glob-to-any-file:
+      - fastlane_core/*
 
 "tool: frameit":
   - changed-files:
-      - any-glob-to-any-file:
-          - frameit/*
+    - any-glob-to-any-file:
+      - frameit/*
 
 "tool: gym":
   - changed-files:
-      - any-glob-to-any-file:
-          - gym/*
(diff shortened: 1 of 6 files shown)


.github/workflows/pull_request_labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/labeler@v5
+        with:
+          sync-labels: true
```
rogerluan, line 19: This is `false` by default, and causes the bot to automatically remove the labels when the PR doesn't change the files anymore (files that were modified e.g. in `deliver/*` were reverted and no longer touch it, so label is removed).

I'm on the fence whether we should set this to true or not. I think it might get on the way, e.g. when we manually add a label because we see fit (e.g. a maintainer add the CI label when the PR is changing the Fastfile/Rakefile or some file we didn't foresee that was related to CI, and then it will keep removing the label on every new commit 🤔

.github/workflows/pull_request_labeler.yml
```diff
@@ -0,0 +1,19 @@
+name: "Pull Request Labeler"
+on:
+  pull_request: # Temporary, for testing while PR is still open
+  # pull_request_target
```
rogerluan, line 4: Note to reviewers: please don't merge this PR with this code in. Before merging, we need to delete `pull_request` and uncomment `pull_request_target`.

See https://github.com/actions/labeler#notes-regarding-pull_request_target-event for more details

.github/workflows/pull_request_labeler.yml
lacostej: I think the devs should have the final decision on what labels should be set. A change that goes accross the whole codebase, e.g. touches all READMEs will have all labels. Not really meaningful.
Also the tool should only touch the labels related to tools. Is there a way to restrict the labels that the tool can add/remove? Or to prevent to touch the labels if e.g. a particular label was added?

I wouldn't mind a tool that helps to propose better PR titles :)

rogerluan: So what you're saying is you're against this whole PR? That's ok too :) 

The issue with letting "devs" add labels is that the only devs allowed to do that are the core contributors (i.e. people with write/management access to this repo); PR authors who aren't members of _fastlane_ can't add labels, reviewers, maybe not even assignees. 😅

So the problem I'm trying to solve is just classifying the PRs without us, core contributors, necessarily opening that PR, just to triage it and add the label. E.g. I might want to wake up and "review all _supply_ PRs because I'm an Android specialist" but it'd take me an effort to find those PRs efficiently without an auto labeler 😅

What do you think?

----

On the other subject, it would be good to have better PR titles, but:

1. idk if bots can modify them on behalf of users, maybe? If it's just a suggestion, it's not as effective
2. It enters the area of "natural language model" and "interpreting" the PR content (files/changes/current title/description), which I don't really wanna plug an LLM in here right now 😅
3. It could arguably just add e.g. [deliver][supply] etc. but the remaining of the title would remain the same and that's usually where the issue in the PR title lies (lack of conciseness, clarity, etc).

I think that we, maintainers, are better skilled to fix the PR titles problem than an automatic tool 😅 but I'm open to hearing other suggestions! How do you see it being improved?

rogerluan: OH, and to the most important part, my take is that you would make the bot not remove labels. Cool, yeah I'm probably leaning towards that too :)

lacostej: The main use case is for labeling tools, and I think that if we mean to use these tools labels, let's automate them.

So remains different strategies:
1. only create the labels when the PR is created
2. adjust the labels automatically (sync) as the PR is updated

Strategy 2 might be useful if a PR is submitted really early, or if reworked enough to warrant a tool removal But at the same time, It might remove labels that we have added, which is bad.

I would be more comfortable to remove the labels if the tool was able to only remove the labels that it is responsible of, e.g. the `tools` related ones. I don't think there's a feature in the labeller action that allows to restrict which labels can be removed though. If that feature existed I think we could rely on it.

So I am not 100% sure activating syncing will be the best. It might help. It might cause issues. It remains to be seen in practice what benefits/drawbacks we would have.

I see 2 options then:
1. we do not activate sync, and we only create the labels at PR creation time
2. we activate sync and actively modify the labels when the PR is updated. But we should probably review this strategy later to see if it causes issues.

Both strategies are fine to me.

rogerluan: We can keep the bot adding labels even after PR creation time (e.g. on every commit, re-evaluate) at the same time that it doesn't remove labels (i.e. `auto-sync: false`). Based on your preferences and mine, I think that's the best for us, for the goals we want to achieve for the project maintainability :) 

I'll make this change! 🙌 thanks for the input @lacostej !

And Merry Christmas to you and your family! 🥰

rogerluan:
```suggestion
```

.github/workflows/pull_request_labeler.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/labeler@v5
+        with:
```
rogerluan, line 18: Committing from GitHub iOS app :P 

```suggestion
```

.github/labeler.yml
```diff
@@ -0,0 +1,102 @@
...
+"tool: trainer":
+  - changed-files:
+    - any-glob-to-any-file:
+      - trainer/**/*
+
+"type: ci":
+  - changed-files:
+    - any-glob-to-any-file:
+      - .github/**/*
+      - .circleci/**/*
```
lacostej, line 102:
```suggestion
      - .circleci/**/*
      - appveyor.yml
```

lacostej (approved):

PR #21777, opened 2024-01-04
[fastlane][ci] Lock google-cloud-env < 2.0.0 for Ruby 2.6 and fix CI to use specified ruby versions

joshdholtz:
### Motivation and Context

- CI wasn't using the specified version of Ruby (always using system)
- `google-cloud-env` dropped Ruby 2.6 support in version `2.0.0` so locking us into `< 2.0.0` until we drop Ruby 2.6 (which will be soon-ish but not doing it because its default on some macOS versions still)

### Description

- Lock `fastlane.gemspec` to use `< 2.0.0` for `google-cloud-env`
- Using macOS orb on CircleCI to switch Ruby versions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -285,7 +285,7 @@ workflows:
       - tests_macos:
           name: 'Execute tests on macOS (Xcode 14.3.1, Ruby 3.1)'
           xcode_version: '14.3.1'
-          ruby_version: '3.1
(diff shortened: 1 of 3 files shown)


ksuther:
Just leaving a note that this change breaks fastlane when installing as a gem and updating. Doing the following:

```
gem install fastlane
gem update
fastlane
```

Will fail with the error in #21788.

PR #21790, opened 2024-01-10
[match] improve encryption internals, solving flaky test (#21663)

### LINKED ISSUE #21663: match encryption is outdated and doesn't guarantee it detects invalid passwords. Causes flaky test.
I've seen this fail twice already on master.

E.g. https://app.circleci.com/pipelines/github/fastlane/fastlane/5764/workflows/bec22718-cc50-4a05-9952-1615d1d76389/jobs/92676

```
 1) Match Match::Encryption::OpenSSL raises an exception if invalid password is passed
     Failure/Error:
       expect do
         @e.decrypt_files
       end.to raise_error("Invalid password passed via 'MATCH_PASSWORD'")

       expected Exception with "Invalid password passed via 'MATCH_PASSWORD'" but nothing was raised
     # ./match/spec/encryption/openssl_spec.rb:33:in `block (3 levels) in <top (required)>'
     # ./.bundle/ruby/2.6.0/gems/webmock-3.12.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 2 minutes 5.9 seconds (files took 3.34 seconds to load)
7267 examples, 1 failure, 329 pending

Failed examples:

rspec ./match/spec/encryption/openssl_spec.rb:27 # Match Match::Encryption::OpenSSL raises an exception if invalid password is passed
```

Investigating...

lacostej:
### Checklist

### Motivation and Context

Resolves #21663 

### Description

See discussion in above issue.


match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+        auth_data = keyIv[44..-1]
+        cipher.key = key
+        cipher.iv = iv
+        cipher.auth_data = auth_data
+      end
+    end
+
+    class MatchDataEncryption
+      V1_PREFIX = "Salted__"
+      V2_PREFIX = "2_Salted__"
```
lacostej, line 89: this one is ugly. We don't really need to keep "Salted__" in the name as we are not compatible with `openssl enc` anymore.

lacostej:
Well apart from the new #21792 flaky test, the proposed implementation won't work with all openssl installations

```
     # ./match/spec/importer_spec.rb:122:in `block (3 levels) in <top (required)>'
     # ./.bundle/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # RuntimeError:
     #   unsupported cipher algorithm (AES-256-GCM)
     #   ./match/lib/match/encryption/encryption.rb:49:in `initialize'
```

fastlane/lib/fastlane/actions/docs/sync_code_signing.md
```diff
@@ -509,7 +509,15 @@ Please be careful when using this option and ensure the certificates and profile
...
-If you want to manually decrypt a file you can.
+If you want to manually decrypt or encrypt, you can use the companion script `match_enc`:
+
+```no-highlight
+match_enc encrypt "<fileYouWantToEncryptPath>" "<password>" "<encryptedFilePath>"
+
+match_enc decrypt "<fileYouWantToDecryptPath>" "<password>" "<decryptedFilePath>"
+```
+
+In versions of fastlane priori to 2.219.0, the encryption was compatible with openssl aes-256-cbc cipher.
```
lacostej, line 520: I could also get rid of the whole "old way of decrypting things".

`match_enc` should also work.

bin/match_enc
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env ruby
+require 'match'
+
+# CLI to encrypt/decrypt files using fastlane match encryption layer
+
+def usage
+  puts("USAGE: [encrypt|decrypt] input_path password [output_path]")
```
milch, line 7: Hmm, I wouldn't put the password in the shell history. Maybe `read -s` it?

bin/match_enc
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env ruby
```
milch, line 1: Personally I'm not a huge fan of abbreviations in the names of tools, maybe let's call it `match_encryption`? `match_file` could also work, it would read as `match_file encrypt`/`match_file decrypt`, which would be fairly intuitive.

fastlane/lib/fastlane/actions/docs/sync_code_signing.md
milch: Yeah, I would just get rid of it.

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+
+module Match
+  module Encryption
+    # This is to keep backwards compatibility with the old fastlane version which used the local openssl installation.
+    # The encryption parameters in this implementation reflect the old behavior which was the most common default value in those versions.
+    # As for decryption, 1.0.x OpenSSL and earlier versions use MD5, 1.1.0c and newer uses SHA256, we try both before giving an error
+    class EncryptionV1
+      ALGORITHM = 'aes-256-cbc'
+
+      def encrypt(data, password, salt, hash_algorithm = "MD5")
```
milch, line 13: Hmm, do we even still need to be able to do v1 encryption or could we just get rid of it. New files should just get v2 and old files can continue to use v1, then they will naturally expire over time and get replaced with v2

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+        keyivgen(cipher, password, salt, hash_algorithm)
+
+        data = cipher.update(encrypted_data)
+        data << cipher.final
+      end
+
+      private
+
+      def keyivgen(cipher, password, salt, hash_algorithm)
+        cipher.pkcs5_keyivgen(password, salt, 1, hash_algorithm)
```
lacostej, line 35: I considered increasing this very low iteration value, but changing the value requires changing the encryption format anyway, so let's just change the cipher etc.

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+    end
+
+    # The newer encryption mechanism
+    # A more secure key and IV generation
+    # IV is randomly generated and provided unencrypted
+    # salt should be randomly generated and provided unencrypted (like in the current implementation)
+    # key is generated with OpenSSL::KDF::pbkdf2_hmac with properly chosen parameters
+    # Short explanation about salt and IV: https://stackoverflow.com/a/1950674/6324550
+    class EncryptionV2
```
milch, line 45: Just as a suggestion but nothing blocking, I think this could be refactored so that the IV gen happens in one class per algorithm, and then there's a single encrypt/decrypt function that takes in the IV gen class. I think overall it would just cut down on the duplication a little bit.

match/lib/match/encryption/encryption.rb
milch: Might as well make it something like `match_encrypted_v2_` or something, so as not to suggest that there might be some compatibility there.

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+
+      def encrypt(data, password, version = 2)
+        salt = SecureRandom.random_bytes(8)
+        if version == 2
+          e = EncryptionV2.new
+          encryption = e.encrypt(data, password, salt)
+          encrypted_data = V2_PREFIX + salt + encryption[:auth_tag] + encryption[:encrypted_data]
+        else
+          e = EncryptionV1.new
+          encrypted_data = V1_PREFIX + salt + e.encrypt(data, password, salt)
```
milch, line 99: Shaping the output and extracting the fields necessary for decryption back out from the string might be a good thing to put directly into the `v1`/`v2` classes, i.e. make it closer to the strategy pattern

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+        Base64.encode64(encrypted_data)
+      end
+
+      # expected_salt is only used when testing
```
milch, line 104:
```suggestion
```

I don't think this is there anymore

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,148 @@
...
+        else
+          salt = stored_data[8..15]
+          data_to_decrypt = stored_data[16..-1]
+          e = EncryptionV1.new
+          begin
+            e.decrypt(data_to_decrypt, password, salt)
+          rescue => _ex
+            # Note that we are not guaranteed to catch the decryption errors here if the password is wrong
+            # as there's no integrity checks.
+            # With a wrong password, there's a 0.4% chance it will decrypt garbage and not fail
```
milch, line 122: Would be good to have the link to that discussion on the OpenSSL issue tracker here too in case anyone wonders about this in the future

bin/match_enc
lacostej: maybe. If I have to start making this tool a bit more complex, I am wondering if I should rewrite it completely, maybe as a fastlane action. The name was originally to be close to `openssl enc`.

bin/match_enc
lacostej: The name was originally to be close to `openssl enc`.  Your comment makes sense though.

match/lib/match/encryption/encryption.rb
lacostej: Well, I use it in the tests :) I figured it didn't cause too much problem to keep it there as it brings some sort of symmetry to the code. We could also tag the class / methods as deprecated, but then I would have to update the tests as well.

match/lib/match/encryption/encryption.rb
lacostej: You mean as passing the ivgen function as a "strategy" to the same code that is shared between V1 and V2?

I thought about it, but decided against for now because
* I preferred to keep V2 on its own, completely unrelated to common code with V1, to make it more readable.
* There used to be a few lines, and now there are already 3 classes. I felt it was maybe enough abstractions

match/lib/match/encryption/encryption.rb
```diff
@@ -0,0 +1,150 @@
...
+            fallback_hash_algorithm = "SHA256"
+            e.decrypt(data_to_decrypt, password, salt, fallback_hash_algorithm)
+          end
+        end
+      end
+    end
+
+    # by default, encrypt or decrypt in place
+    class MatchFileEncryption
+      def encrypt(file_path, password, output_path = nil)
```
lacostej, line 134: Another question I have: do we want to enforce keywords parameters instead?

I think it usually makes for better API management and backwards compatibility.

Should I update this file and use keywords args instead?

bin/match_file
```diff
@@ -0,0 +1,57 @@
...
+  else
+    return nil
+  end
+end
+
+# read the password
+password = nil
+loop do
+  password = ask_password_twice
+  break unless password.nil?
```
rogerluan, line 49: Won't people get stuck in this loop without being able to quit until they enter 2 identical strings?

I'm having issues with my env currently, I couldn't test this 😢 but I remember some of fastlane's built-in "ask for X" prompts doesn't allow `ctrl+C` to quit 😬

rogerluan (changes requested):
Just some nitpick comments about docs and formatting, as I don't quite grasp the whole encryption details 😅 but the overall logic LGTM and I liked your approach of keeping this backwards compatible by trying both V1 and V2 👌

fastlane/lib/fastlane/actions/docs/sync_code_signing.md
```diff
@@ -509,12 +509,16 @@ Please be careful when using this option and ensure the certificates and profile
...
 ### Manual Decrypt
 
-If you want to manually decrypt a file you can.
+If you want to manually decrypt or encrypt a file, you can use the companion script `match_file`:
 
 ```no-highlight
-openssl aes-256-cbc -k "<password>" -in "<fileYouWantToDecryptPath>" -out "<decryptedFilePath>" -a -d -md [md5|sha256]
+match_file encrypt "<fileYouWantToEncryptPath>" ["<encryptedFilePath>]"
+
+match_file decrypt "<fileYouWantToDecryptPath>" ["<decryptedFilePath>]"
```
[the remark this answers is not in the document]
lacostej: I think that once you install the gem, everything under `bin` will be in your CLASSPATH. SO `match_file` should be enough.

The [] were because the second parameter is optional and match_file decrypts / encrypts in place when omitted.

bin/match_file
lacostej: yes. Let's try to fix the lack of ctrl+C support first then ;)

lacostej: added. But just in match-file. Maybe overkill?

match/lib/match/encryption/encryption.rb
lacostej: I added some but not in all classes. Makes sense or inconsistent?

bin/match_file
lacostej: in the end I removed the ctrl-C support as it doesn't seem to be possible to do it on older rubies.

bin/match_file
```diff
@@ -0,0 +1,60 @@
...
+def ask_password(msg)
+  ask(msg) do |q|
+    q.whitespace = :chomp
+    q.echo = "*"
+  end
+end
+
+def ask_password_twice
+  password = ask_password("Enter the password: ")
+  return "" if password.empty? || password == "\u0003" # CTRL-C char
```
lacostej, line 37: if you press enter on the first password, or CTRL-C and enter, then the loop exits.

bin/match_file
rogerluan: Nice!

rogerluan (approved):
LGTM! 🚀 thanks for addressing the CTRL+C exit 😊

arc-v2:
Hi @lacostej I'm experiencing some issue with decryption since 2.220.0 was released. Could you confirm the following query:
Does this mean that match repo encrypted with fastlane version 2.220.0 will not work (decrypt) with fastlane version 2.219.0 but vice-versa should keep working?

andre-alves:
Just experienced the same issue. I saved provisioning profiles using fastlane 2.220.0, and fastlane 1.219.0 or older is no longer able to decrypt them.

stodirascu:
I have the same issue as above. If locally somebody is encrypting a profile or a certificate using 2.220, and the CI is decrypting it with 2.219, it will fail. So it's on a file-by-file basis.

This must be mentioned as a breaking change in the release notes at least.

Issue #21765, opened 2024-01-03
No device found with name ... using 2.218.0


michalsrutek:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

The command executes successfully on 2.217.0, but fails on 2.218.0 with
```
No simulators found that are equal to the version of specifier (17.0.1) and greater than or equal to the version of deployment target (0)
```

and

```
No device found with name 'iPhone 15 Pro Max' with version 17.0.1
```

##### Command executed

```ruby
    run_tests(
      workspace: "XYZ.xcworkspace",
      scheme: "XYZ",
      prelaunch_simulator: true,
      disable_package_automatic_updates: false,
      devices: ['iPhone 15 Pro Max (17.0.1)'],
      ensure_devices_found: true
    )
```

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> [REPLACE THIS WITH YOUR INFORMATION] </pre>
</details>

### Environment

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                              |
| --------------------------- | -------------------------------------------------- |
| OS                          | 13.5.1                                             |
| Ruby                        | 3.2.2                                              |
| Bundler?                    | true                                               |
| Git                         | git version 2.39.3 (Apple Git-145)                 |
| Installation Source         | /usr/local/lib/ruby/gems/3.2.0/bin/fastlane        |
| Host                        | macOS 13.5.1 (22G90)                               |
| Ruby Lib Dir                | /usr/local/Cellar/ruby/3.2.2_1/lib                 |
| OpenSSL Version             | OpenSSL 3.1.1 30 May 2023                          |
| Is contained                | false                                              |
| Is homebrew                 | false                                              |
| Is installed via Fabric.app | false                                              |
| Xcode Path                  | /Applications/Xcode-15.0.1.app/Contents/Developer/ |
| Xcode Version               | 15.0.1                                             |
| Swift Version               | 5.9                                                |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_US.UTF-8 | ✅ |
| LC_ALL   |             |   |
[...]

PR #21800, opened 2024-01-12
Add an interactive console for users as well

lacostej:
### Checklist

### Motivation and Context

Sometimes I wish to send code for users to test directly into their setup. This would allow to do it.

### Description

I've used the similar code in our u3d project here for years https://github.com/DragonBox/u3d/blob/8c156f401f2b8443936e3873b8566a965c783165/lib/u3d/commands.rb#L68-L80

### DIFF
--- a/fastlane/lib/fastlane/commands_generator.rb
+++ b/fastlane/lib/fastlane/commands_generator.rb
@@ -249,6 +249,15 @@ module Fastlane
         end
       end
 
+      command :console do |c|
+        c.syntax = 'fastlane console'
+        c.description = 'Opens an interactive developer console'
+        c.action do |args, options|
+          require 'fastlane/console'
+          Fastlane::Console.execute(args, options)
+        end
+      end
+
       command :enable_auto_complete do |c|
         c.syntax = 'fastlane enable_auto_complete'
         c.description = 'Enable tab auto completion'
--- a/fastlane/lib/fastlane/console.rb
+++ b/fastlane/lib/fastlane/console.rb
@@ -1,0 +1,24 @@
+require 'irb'
+
+module Fastlane
+  # Opens an interactive developer console
+  class Console
+    def self.execute(args, options)
+      ARGV.clear
+      IRB.setup(nil)
+      @irb = IRB::Irb.new(nil)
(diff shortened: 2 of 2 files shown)


AliSoftware (approved):
I'm wondering if we should name this command `repl` instead. Because pedantically speaking the term console does not imply interactivity but more reading output (think macOS's Console.app), while things like `pry` and `irb` are called REPL (Read-Execute-Print Loop).

That being said since you added `bin/console` with such a name too we should keep the names consistent.

lacostej:
> I'm wondering if we should name this command `repl` instead. Because pedantically speaking the term console does not imply interactivity but more reading output (think macOS's Console.app), while things like `pry` and `irb` are called REPL (Read-Execute-Print Loop).
> 
> That being said since you added `bin/console` with such a name too we should keep the names consistent.

I wouldn't change `bin/console` as it follows the bundler convention. Console there probably refer to the original [system consoles](https://en.wikipedia.org/wiki/System_console). The consistency is I think also a strong argument to keep the command identical.

Note that Mac's Console is probably one of the few IT system named console that doesn't do input. I'll blame Apple for the confusion.

rogerluan (approved):
Nice! 🌟 this is awesome and quite intuitive IMO.

I think this deserves to be documented somewhere, for discoverability! Idk where exactly, though. It might be in this repo, or in https://github.com/fastlane/docs, what do you think? Where should this live?

lacostej:
> Nice! 🌟 this is awesome and quite intuitive IMO.
> 
> I think this deserves to be documented somewhere, for discoverability! Idk where exactly, though. It might be in this repo, or in https://github.com/fastlane/docs, what do you think? Where should this live?

yes, in this repo and in the docs would be nice, together with a nice screenshot or even an animated gif.

Example:

image

I hope to use it to help troubleshoot issues like #21765 :)

rogerluan:
@lacostej you probably need to reopen this manually 👀

lacostej:
yes, but I don't seem to be able in the same PR. Something went wrong.

lacostej:
now it lets me

lacostej:
nope, it is stuck. I'll reopen a new one.

PR #21807, opened 2024-01-15
Fix WARNINGs when building fastlane gemspec

lacostej:
Annoying warnings when building the gemspec. Clutters the logs and the mind.

```
$ gem build fastlane.gemspec
WARNING:  description and summary are identical
WARNING:  open-ended dependency on colored (>= 0) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on optparse (>= 0.1.1) is not recommended
  if optparse is semantically versioned, use:
    add_runtime_dependency 'optparse', '~> 0.1', '>= 0.1.1'
WARNING:  open-ended dependency on xcpretty-travis-formatter (>= 0.0.3) is not recommended
  if xcpretty-travis-formatter is semantically versioned, use:
    add_runtime_dependency 'xcpretty-travis-formatter', '~> 0.0', '>= 0.0.3'
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: fastlane
  Version: 2.219.0
  File: fastlane-2.219.0.gem
```

### Checklist

- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I see several green `ci/circleci` builds in the "All checks have passed" section of my PR ([connect CircleCI to GitHub](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-) if not)
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.
- [x] I've added or updated relevant unit tests.

### Description

* tightened some dependencies
* added a gem summary

PR #21808, opened 2024-01-15
Fix gemspec warnings (update `Gemfile.lock`)

lacostej:
Was missing in #21807


rogerluan (approved):
👍 

This is just annoying locally right for us? Because this doesn't affect real users AFAIK

PR #21809, opened 2024-01-16
[match] select cert with the most furthest expiration date in the future

nekrich:
### Checklist

- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I see several green `ci/circleci` builds in the "All checks have passed" section of my PR ([connect CircleCI to GitHub](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-) if not)
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.
- [x] I've added or updated relevant unit tests.

### Motivation and Context

In most cases, there is only one cert/key pair in the match storage.
But when dealing with developer_id certificates, there could be two or even three. The reason for this is that developer ID certs have a 5-year lifespan. And it looks like an app signed with developer_id cert stops working when the developer id cert used to sign it is expired. Every couple of years, we issue a new developer id cert and use it to sign new app versions. In this case, we are sure that the latest release will work for a couple more years, and users will have plenty of time to update the app.

Since the developer id cert is not expired we use manual import. After that, we have two cert/key pairs in the match storage.

The problem I want to address is that the match uses a cert with the biggest id (filename). And this is wrong. We need a match to select a cert with the furthest expiration date in the future.

### Description

1. select_cert_or_key is now select_cert. The method returns the cert with the furthest expiration date.
2. New runner_unit_spec not to mix unit tests with looks like integration tests in the runner_spec.
3. There could be several cert/key pairs in the storage. Refactored cert deletion logic to address this case. There is no value in expired certs at all.

#### Cert renewal updates
5. cert_paths contain only valid certificates. It looks better and requires fewer checks in the code afterward.
6. There is no need for the renew_expired_certs option 😅.

#### Other improvements
7. There is no need in `keys` variable. We need only certs with corresponding p12 keys.
8. Updates some gsubs to use regex, indicating the correct file extension and the end of a string.

PS. Looks like failing tests on Ubuntu is not related to my changes.

### Testing Steps

Run the match with storage having 2 or more certs of the same type.
Run the match with storage having 2 or more certs that are expired.

PR #21812, opened 2024-01-16
[match][hotfix] remove the `renew_expired_certs` option introduced in #21691 and revert the default behavior while we address issues with it

nekrich:
### Checklist

### Motivation and Context

Removes `renew_expired_certs` #21691, but behavior stays the same.
Doing this in favor of #21809, that doesn't require this option any more


match/lib/match/runner.rb
```diff
@@ -84,12 +84,12 @@ def run(params)
       end
 
       # Certificate
-      cert_id = fetch_certificate(params: params, renew_expired_certs: params[:renew_expired_certs])
+      cert_id = fetch_certificate(params: params, renew_expired_certs: true)
```
rogerluan, line 87: If we wanted to keep the previous behavior, we'd pass `false` here, no?

match/lib/match/runner.rb
```diff
@@ -84,12 +84,12 @@ def run(params)
...
 
       # Mac Installer Distribution Certificate
       additional_cert_types = params[:additional_cert_types] || []
       cert_ids = additional_cert_types.map do |additional_cert_type|
-        fetch_certificate(params: params, renew_expired_certs: params[:renew_expired_certs], specific_cert_type: additional_cert_type)
+        fetch_certificate(params: params, renew_expired_certs: true, specific_cert_type: additional_cert_type)
```
rogerluan, line 92: Same?

rogerluan (changes requested):
Thanks for opening this PR, we need to carefully review whether we want to revert the feature (aka merge this PR) or merge your other one (that fixes it) 😬 

Can you review my question below? 🙏

nekrich:
@rogerluan , sorry for the delay. Changed all to false and removed auto-renewal test.

match/lib/match/runner.rb
```diff
@@ -172,7 +172,7 @@ def fetch_certificate(params: nil, renew_expired_certs: false, specific_cert_typ
       is_cert_renewable = is_authenticated_with_login || is_cert_renewable_via_api
 
       # Validate existing certificate first.
-      if renew_expired_certs && is_cert_renewable && storage_has_certs
+      if renew_expired_certs && is_cert_renewable && storage_has_certs && !params[:readonly]
```
rogerluan, line 175: Can you remind me why we need this change? 👀  @nekrich

rogerluan (approved):
LGTM, just left a question. @lacostej I think this is good to go in case we need to deploy 😬 do you think we should merge it in and then revert it + fix the error in the other related PR?

match/lib/match/runner.rb
lacostej: @rogerluan well, we shouldn't be deleting things when we are in readonly mode :)

lacostej (approved):

Issue #21792, opened 2024-01-10
fastlane_core/spec/command_executor_spec.rb:12 is flaky


lacostej:
```
  1) FastlaneCore FastlaneCore::CommandExecutor execute executes a simple command successfully
     Failure/Error: expect(result).to eq('foo')
     
       expected: "foo"
            got: ""
     
       (compared using ==)
     # ./fastlane_core/spec/command_executor_spec.rb:12:in `block (5 levels) in <top (required)>'
     # ./.bundle/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  2) FastlaneCore FastlaneCore::CommandExecutor execute executes a simple command successfully
     Failure/Error: expect(result).to eq('foo')
     
       expected: "foo"
            got: ""
     
       (compared using ==)
     # ./fastlane_core/spec/command_executor_spec.rb:12:in `block (5 levels) in <top (required)>'
     # ./.bundle/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  3) FastlaneCore FastlaneCore::CommandExecutor execute executes a simple command successfully
     Failure/Error: expect(result).to eq('foo')
     
       expected: "foo"
            got: ""
     
       (compared using ==)
     # ./fastlane_core/spec/command_executor_spec.rb:12:in `block (5 levels) in <top (required)>'
     # ./.bundle/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 2 minutes 53.7 seconds (files took 2.1 seconds to load)
10000 examples, 3 failures
```

### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [ ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

##### Complete output when running fastlane, including the stack trace and command used

<details>
  <pre> [REPLACE THIS WITH YOUR INFORMATION] </pre>
</details>

### Environment

<details>
  <pre> [REPLACE THIS WITH YOUR INFORMATION] </pre>
</details>

lacostej:
[...]

PR #21830, opened 2024-01-24
[ci] increase the max number of files available

lacostej:
Increase the amount of files available on mac by default.

This is in handy when running tests that involve stress testing a bit the system (such as #21792)

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,7 @@ jobs:
       - run:
           name: bundle exec fastlane execute_tests
           command: |
+            ulimit -n 65536
             bundle exec fastlane snapshot reset_simulators --force
             RUBYOPT=<< parameters.ruby_opt >> bundle exec fastlane execute_tests
       - *cache_save_rubocop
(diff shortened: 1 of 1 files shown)


lacostej:
not really sure why appveyor failed here. Another flaky test?

lacostej:
Force pushed to see if appveyor fails building again.

AliSoftware (approved):
If that's the reasons those tests are flaky, this is a bit scary 😅  Even if we're stress-testing the filesystem (why should we, though? Not sure that's the place of _fastlane_ itself to do that, would feel more like an Xcode problem?) I'm surprised we need that many opened files during testing in the first place. Might be worth investigating the root cause of that at some point.

lacostej:
> if that's the reasons those tests are flaky, this is a bit scary 😅

@AliSoftware no, that's not the reason. By default we don't need that many files.

The command executor flaky test fails regularly. Last HEAD commit on master failed in one of the executor. But to reproduce the failure more consistently, I am running the same tests thousands of times, and this fails early if we don't increase the file limit.

I thought we'd better increase it right away so that we can run heavy stress testing workloads whenever we feel like.

The only drawback I can see is that if we enable it by default on the project, we wouldn't detect an internal issue in fastlane that would cause too many files to be opened during our tests. Is that an issue? That's a bit unlikely.

Issue #21860, opened 2024-02-09
[match] visionOS devices are not included in provisioning profiles


peter-gyarmati:
### New Issue Checklist

- [x] Updated fastlane to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [x] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description
visionOS devices are not included in provisioning profiles. 

With visionOS coming out Apple didn't bother to add a new platform to their signing infrastructure and you need to add your VisionPro device to the iOS development provisioning profiles as per their portal (picture attached). 

`match` currently doesn't add these devices to the profile. In case of manually adding them to the appropriate profile the next run of `match` will override the profile and remove the VisionPro device from the list of devices. Override is expected as the profile has been manually modified, however there is no other way to include them at the moment.

<img width="813" alt="image" src="https://github.com/fastlane/fastlane/assets/92735687/b3942faa-a02d-45f1-8d8d-3027ec02b92f">

Probably https://github.com/fastlane/fastlane/blob/26b695f8ed697fa21ca30d57db97d060ce21e631/match/lib/match/profile_includes.rb#L46 should be modified (and the used helper `Match::Portal::Fetcher` to allow the platform) to add Vision Pro devices as well to the profiles when new devices are added.

##### Command executed
`fastlane match` with `force_for_new_devices:true`.

##### Complete output when running fastlane, including the stack trace and command used
N/A

### Environment
N/A

BookerE1:
Hello, peter-gyarmati

As per your post, I give you the best answer please see below and follow the step-by-step information...

Match is a tool that helps you sync your certificates and profiles across your team using a git repo, Google Cloud, or Amazon S3. VisionOS is a new operating system for Apple’s upcoming Vision Pro headset, which is a mixed-reality device that combines augmented reality and virtual reality.

According to your issue description, you are unable to include visionOS devices in your provisioning profiles using Match. This seems to be a bug or a missing feature in the match, as it does not recognize visionOS as a valid platform for code signing. You have also identified the possible source of the problem in the match code, which is the devices_differ. method that compares the devices in the portal profile and the cached devices.


To resolve this issue, you have a few options:

[...]

PR #21841, opened 2024-01-30
[pilot] add xros as valid option for pilot upload

demonnic:
### Checklist

- [x] I've run `bundle exec rspec` from the root directory to see all new and existing tests pass
- [x] I've followed the _fastlane_ code style and run `bundle exec rubocop -a` to ensure the code style is valid
- [x] I see several green `ci/circleci` builds in the "All checks have passed" section of my PR ([connect CircleCI to GitHub](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-) if not)
- [x] I've read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [x] I've updated the documentation if necessary.
- [x] I've added or updated relevant unit tests.

### Motivation and Context

The Apple Vision Pro headset officially releases this Friday, and currently the process I use for uploading apps to Apple is blocked because xros is not considered a supported platform by _fastlane pilot_

### Description

This change adds xros as a valid option everywhere appletvos is mentioned for _pilot_, and also adjusts the specs. Prompted by the discussion https://github.com/fastlane/fastlane/discussions/21802

I updated our resigning service to use my updated local copy of _fastlane_ and was able to `fastlane pilot upload` our vision pro ipa with these changes.

Once this PR is ironed out I'm likely to take a look at adding it to the `deliver` in the hopes of maybe getting a release out ahead of or in line with the vision pro launch if at all possible.

### Testing Steps

The command ultimately used to test the upload was `bundle exec fastlane pilot upload --api-key-path <REDACTED> --skip_waiting_for_build_processing --team_id <REDACTED> --ipa <PATH/TO/IPA>`

PR #21871, opened 2024-02-14
[match] Include visionOS devices in provisioning profiles 

peter-gyarmati:
### Checklist

### Motivation and Context
See https://github.com/fastlane/fastlane/issues/21860 for more context.

### Description
Added `APPLE_VISION_PRO` to the `DeviceClass` enum in order to include Vision Pro devices in iOS provisioning profiles.
Also added `VISION_OS` to SpaceShip to support the platform in AppStore connect.

### Testing Steps
Generated new provisioning profiles and validated that the devices are present in the iOS provisioning profiles.

Did it with the following steps:
```
profile=$(security cms -D -i <PATH-TO_PROFILE>)
echo $profile | grep "<DEVICE_UDID>"
```
This should return successfully if the device is present in the profile.

### DIFF
--- a/spaceship/lib/spaceship/connect_api.rb
+++ b/spaceship/lib/spaceship/connect_api.rb
@@ -89,7 +89,6 @@ module Spaceship
       VISION_OS = "VISION_OS"
       WATCH_OS = "WATCH_OS"
 
-
       ALL = [IOS, MAC_OS, TV_OS, VISION_OS, WATCH_OS]
 
       def self.map(platform)
--- a/spaceship/lib/spaceship/connect_api.rb
+++ b/spaceship/lib/spaceship/connect_api.rb
@@ -86,9 +86,11 @@ module Spaceship
       IOS = "IOS"
       MAC_OS = "MAC_OS"
       TV_OS = "TV_OS"
+      VISION_OS = "VISION_OS"
       WATCH_OS = "WATCH_OS"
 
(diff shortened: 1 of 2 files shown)


peter-gyarmati:
@rogerluan @lacostej Could you please take a look at this PR? I'd like to get it merged before the next release. (When is the next release?)

lacostej (approved):
LGTM.

One thing I wonder is that if I understand correctly, Apple has renamed xros into visionos at the last minute. 

We just introduced the public parameter for pilot in #21841, and I wonder if we should have named it `visionos` instead.

peter-gyarmati:
> LGTM.
> 
> One thing I wonder is that if I understand correctly, Apple has renamed xros into visionos at the last minute.
> 
> We just introduced the public parameter for pilot in #21841, and I wonder if we should have named it `visionos` instead.

I think they didn't decide on the naming yet, that's why I went with both options. Actually if you unpack a provisioning profile, the platform part looks like this:
```
<key>Platform</key>
	<array>
		<string>iOS</string>
		<string>xrOS</string>
		<string>visionOS</string>
	</array> 
```

IMHO the most sustainable way would be supporting both until Apple decides on one or the other. (However my hunch is that both will stay in different circumstances)

peter-gyarmati:
Could someone please merge this? @lacostej @rogerluan

PR #21885, opened 2024-02-21
[produce] fix produce passing incorrect teamId when modifying capabilities

Hiroto-N:
### Checklist

### Motivation and Context
Implementation of updating bundleIdCapabilities was incorrectly assuming that seedId was always equals to the teamId of the team that the bundle identifiers is created under. Unsure why this happens, but we've observed a number of bundle identifiers that show this behaviour. 

This functionality to modify capabilities is now described in the public Appstore Connect APIs [here](https://developer.apple.com/documentation/appstoreconnectapi/bundle_id_capabilities) which uses the `DELETE` and `POST` HTTP methods, but since the current implementation utilising the private "PATCH" method is still working, I've opted to fix the current requests rather than implement based on the appstore connect APIs since that looks to requires a pretty hefty rewrite and testing of the bundle id related calls such as `produce` and `modify_services`.

### Description
Fixed the request body sent in `patch_bundle_id_capability` to pass the correct values for `teamId` and `seedId`, as well as adding `edit` and `delete` permissions to fix a BadRequest error when sending the request from the spaceship CLI (matches patch calls in the dev portal). 

Tested from the CLI using web login.

### Testing Steps
```ruby
# inside `fastlane spaceship` context

# Find Bundle ID with seedId that is different to current teamId. Not sure what triggers this scenario
bundleId = Spaceship::ConnectAPI::BundleId.all.filter { |id| id.seed_id != <current team id> }.first

# Enable/Disable Bundle ID capability
bundleId.update_capability "APP_ATTEST", enabled: true|false
```


lacostej (approved):
LGTM

Thanks for the contribution!

joshdholtz (approved):
Looks good to me! Thank you for fixing this 💪

Issue #21877, opened 2024-02-17
Apple changed AppStatus 'Ready for Sales' to 'Ready for Distribution'


naoigcat:
### New Issue Checklist

- [ ] Updated fastlane to the latest version
- [ ] I read the [Contribution Guidelines](https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md)
- [ ] I read [docs.fastlane.tools](https://docs.fastlane.tools)
- [ ] I searched for [existing GitHub issues](https://github.com/fastlane/fastlane/issues)

### Issue Description

An error occurs when `fastlane deliver download_metadata` is executed for an app whose status on AppStoreConnect is `Ready for Distribution`.

I think this is happening because [Apple changed the wording of AppStatus](https://developer.apple.com/news/releases/?id=02082024e).

> Ready for Sale has been renamed to Ready for Distribution, and now indicates that your app can be published. To view if your app is available on the App Store, visit Pricing and Availability.

##### Command executed

```sh
fastlane deliver download_metadata --app_identifier xxxxxx --username xxxxxx --team_id xxxxxx --force
```

##### Complete output when running fastlane, including the stack trace and command used

<details>
<pre>
[✔] 🚀 
[16:33:17]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[16:33:18]: Login to App Store Connect (xxxxxx)
[16:33:19]: Login successful

+----------------------------------------------------------------+
|                    deliver 2.219.0 Summary                     |
+--------------------------------------+-------------------------+
| app_identifier                       | xxxxxx                  |
| username                             | xxxxxx                  |
| team_id                              | xxxxxx                  |
| metadata_path                        | ./metadata              |
| force                                | true                    |
| screenshots_path                     | ./screenshots           |
| platform                             | ios                     |
| edit_live                            | false                   |
| use_live_version                     | false                   |
| skip_binary_upload                   | false                   |
| skip_screenshots                     | false                   |
| skip_metadata                        | false                   |
| skip_app_version_update              | false                   |
| overwrite_screenshots                | false                   |
| sync_screenshots                     | false                   |
| submit_for_review                    | false                   |
| verify_only                          | false                   |
| reject_if_possible                   | false                   |
| phased_release                       | false                   |
| reset_ratings                        | false                   |
| run_precheck_before_submit           | true                    |
| precheck_default_rule_level          | warn                    |
[...]

PR #21890, opened 2024-02-23
[spaceship] handle new app status

lucgrabowski:
### Checklist

### Motivation and Context

There are changes in app statuses on Appstore Connect and in OpenAPI spec (https://developer.apple.com/documentation/appstoreconnectapi) as mentioned in https://developer.apple.com/news/releases/?id=02082024e and https://github.com/fastlane/fastlane/issues/21877. 
There is a new enum `AppVersionState` for `AppStoreVersion`  and there is also a new enum `State` for `AppInfo`.
`AppStoreVersion` was also enhanced with `reviewType` property.

### Description

After new Appstore Connect API release `AppStoreVersion` has two properties related to state: `AppVersionState` and `AppStoreState`. Similarly `AppInfo` has `State` and `AppStoreState`.
As written in release notes https://developer.apple.com/documentation/appstoreconnectapi/app_store_connect_api_release_notes/app_store_connect_api_3_3_release_notes 
```
appStoreState attribute is now deprecated and replaced with appVersionState in the object [AppStoreVersion.Attributes]
...
appStoreState attribute is now deprecated and replaced with state in the object [AppInfo.Attributes]
```

There is a change in meaning of statuses https://developer.apple.com/news/releases/?id=02082024e:
```
Ready for Sale has been renamed to Ready for Distribution, and now indicates that your app can be published. To view if your app is available on the App Store, visit Pricing and Availability.
Developer Removed from Sale, Removed from Sale, and Pending Agreement are no longer app statuses.
```

#### Some observations:
1. When app is removed from sale by developer, instead of `DEVELOPER_REMOVED_FROM_SALE`, we will get an AppStoreVersion with attributes:
```
"appStoreState" : "READY_FOR_SALE",
"appVersionState" : "READY_FOR_DISTRIBUTION",
```
and the only way to see if the app is live or removed is to check available territories using Pricing and Availability list or use  `URL: https://appstoreconnect.apple.com/iris/v2/appAvailabilities/<XXXXXXX>?include=territoryAvailabilities&limit[territoryAvailabilities]=200
` request (currently not implemented in fastlane)

Screenshot 2024-02-24 at 00 14 58

2. When developer program membership is expired, instead of `PENDING_CONTRACT` status for released version, we will get an AppStoreVersion with attributes:
```
"appStoreState" : "READY_FOR_SALE",
"appVersionState" : "READY_FOR_DISTRIBUTION",
```
and when we check in Pricing and Availability, territories are available but with status CANNOT_SELL as in example:
```
"type" : "territoryAvailabilities",
    "id" : "...",
    "attributes" : {
      "available" : true,
      ...
      "contentStatuses" : [ "CANNOT_SELL" ]
    },
```

Screenshot 2024-02-24 at 00 12 31

### Testing Steps

N/A


lacostej (approved):
LGTM. 

Do we need to update any tools to use these statuses?

Some of the observations you made makes me wonder if we should provide high level functions to determinate when an app is remove from sale or the developer membership is expired.

lucgrabowski:
@lacostej Thanks for review. 
There are several different things that I would like to mention, so I'll list them with numbers to be able to discuss it further, some of those things should be discussed outside of this pull request.

1. I think that new statuses could be applied in filtering methods in `App` model and in `Deliver::SubmitForReview`. 
It looks like the change could be applied 1 to 1, of course along with renaming property used `filter` in requests.
That would require testing `deliver`, uploading and downloading metadata, screenshots, generating html summary,  downloading dsyms, actions (`SetChangelogAction`, `AppStoreBuildNumberAction`), `Precheck::RuleProcessor`

2. Those high level methods that you propose could be useful. If we want to check if app is removed from sale we would have to implement a new request for checking app availability territories
https://developer.apple.com/documentation/appstoreconnectapi/read_app_availablity_territories.
Implementing new endpoint (v2) will require changes to client, as it has v1 hardcoded in hostname.
https://github.com/fastlane/fastlane/blob/385a15f74cc617d2a483790575de17cba6d66665/spaceship/lib/spaceship/connect_api/tunes/client.rb#L22
I tried getting app availability data using deprecated v1 api, but without luck.

3. I think that we should skip checking app availability to understand if it's PENDING_CONTRACT.
There is a lot of content statuses possible https://developer.apple.com/documentation/appstoreconnectapi/territoryavailability/attributes. I noticed that there is the same content status (CANNOT_SELL) for both cases - when account membership is expired or app is removed from sale, .
A better way of checking membership expiration would be checking portal teams `status` (active or expired).
https://github.com/fastlane/fastlane/blob/385a15f74cc617d2a483790575de17cba6d66665/spaceship/lib/spaceship/portal/portal_client.rb#L42-L51

4. And one more thing which I think is important, even though not related to this PR.
Fastlane currently updates territory availabilities when uploading metadata (`Deliver::UploadPriceTier`) using a deprecated method, by modifying App 
https://developer.apple.com/documentation/appstoreconnectapi/modify_an_app
https://developer.apple.com/documentation/appstoreconnectapi/appupdaterequest/data/relationships. 
Price tier is also updated in this request and also it is a deprecated method.

PR #21990, opened 2024-04-24
Fix trainer failure result handling with objective C tests

jaysoffian:
### Checklist

- [X] I've added or updated relevant unit tests.

### Motivation and Context

When trainer is parsing an xcresult bundle which contains test failures, it has an `ActionTestMetadata` object that is identified one way (by "identifier") and it's trying to find the corresponding `TestFailureIssueSummary` object that's identified another way (by `testCaseName`). These values are not the same, so in order to locate the correct `TestFailureIssueSummary` object it needs to transform the `testCaseName` in the same way that Xcode does when generating the `ActionTestMetadata` identifier. Exactly how to do this is undocumented, so folks have attempted to figure it out empirically.

Most recently, dda4007c7a ([trainer] fix issues where number of failures would always be zero (#21432), 2024-01-10) changed this transformation code in a way that fixed test names that contain spaces, but broke Objective-C test names.

### Description

This commit fixes the regression caused by dda4007c7a with ObjC tests as follows:

1) It reverts the transformation behavior to how the code worked previously. Instead of "sanitizing" both the identifier and the test case name, it now keeps the identifier as is and "normalizes" the test case name. The test case name normalization has been moved into its own method which passes unit tests for ObjC, Swift, and tests with spaces.

2) It adds an ObjC xcresult bundle and unit test to prevent regressing parsing ObjC test names.

3) It refactors summaries_to_data so that if the corresponding test failure message cannot be located (but hopefully that won't happen anymore), we at least properly report that the test failed, but use the generic "unknown failure message" as the failure message.

See discussion on:

- https://github.com/fastlane/fastlane/pull/21565
- https://github.com/fastlane/fastlane/pull/21432

### Testing Steps

I ran rubocop and rspec on the trainer code. It passes the existing tests as well as a new test I added to identify this issue and prevent regressions.


jaysoffian:
Not sure why the "[Execute tests on macOS (Xcode 15.3, Ruby 3.3)](https://app.circleci.com/pipelines/github/fastlane/fastlane/6379/workflows/8f126780-f121-4717-abee-5204b0fe152b/jobs/98975?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) job fails, but it doesn't seem to have anything to do with this PR.

triplef:
@joshdholtz any chance to get this or #22134 (an alternative fix for the same issue) merged, or who could help with this? 🙏  Using fastlane with Objective-C tests will currently report false positives (tests reported as passing when they fail) because of this issue.

PR #22039, opened 2024-05-19
Added nil check to avoid exception when production track is not present

stivlo:
With the original code if production track is not present, we get the following exception:

/fastlane-2.220.0/supply/lib/supply/client.rb:303:in `block in latest_version': undefined method `name' for nil (NoMethodError)

      latest_version = tracks.select { |t| t.track == Supply::Tracks::DEFAULT }.map(&:releases).flatten.reject { |r| r.name.nil? }.max_by(&:name)
                                                                                                                      ^^^^^
 
with the modified code we get the correct error message:

[!] Unable to find latest version information from "production" track. Please specify track information by using the '--track' option.

### Checklist


ndhunju:
The suggested error message, "[!] Unable to find latest version information from "production" track. Please specify track information by using the '--track' option." implies that providing --track information like "--track internal" would fix the issue but seems like it still gives the same error.

